PR #1186, opened 2017-05-15
Fix Documentation Wording and Links for Clarity

### LINKED ISSUE #1170: Minor documentation issue - Middleware chapter of tutorial
Middleware chapter of tutorial contains a link back to Service chapter in a reference to when the concept of testing a service without a server was introduced, but this concept is actually introduced in the DSL chapter.

ChristopherDavenport:
Kleisli is mentioned in service, and dsl is where we discuss testing without a server. This could lead to confusion so this is hopefully more clear.

Fixes #1170

### DIFF
--- a/docs/src/main/tut/middleware.md
+++ b/docs/src/main/tut/middleware.md
@@ -48,7 +48,7 @@ add the header if the response is a success, and then pass the response on. We c
 just as easily modify the request before we passed it to the service.
 
 Now, let's create a simple service. As mentioned between [service] and [dsl], because `Service`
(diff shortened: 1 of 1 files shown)


ChristopherDavenport:
StreamApp Spec has been failing intermittently. Noticed it in a couple of my local builds an it showed up here as well.

```scala 
requestShutdown Shuts Down a Server From A Separate Thread
```

ChristopherDavenport:
Added a reference to Kleisli documentation as well. A good paper has lots of citations :smile:

Issue #866, opened 2017-01-18
Port examples to fs2-cats


rossabaker:
A lot of it is commented.  Part of the work here is figuring out what we can uncomment as progress has been made in other areas.

The current compilation failure is:

```
[error] http4s/examples/src/main/scala/com/example/http4s/ExampleService.scala:65: value valueOr is not a member of org.http4s.ParseResult[org.http4s.Uri]
[error]       TemporaryRedirect(uri("/http4s/"))
```

This is probably as simple as changing the URI macro definition.

rossabaker:
#879 fixes the compilation error.  There is still an opportunity for people to uncomment and test more of the examples.

aeons:
Fixed by #1384.

PR #1188, opened 2017-05-16
Fix Most Science Experiments

ChristopherDavenport:
Most are completed, some are being tricky, and a few are just painful to port.

Round 1

Refs #866

### DIFF
--- a/examples/src/main/scala/com/example/http4s/ScienceExperiments.scala
+++ b/examples/src/main/scala/com/example/http4s/ScienceExperiments.scala
@@ -71,12 +71,12 @@ object ScienceExperiments {
       def newBodyP(h: Handle[Task, String]): Pull[Task, String, String] = {
         h.await1Option.flatMap{
           case Some((s, h)) =>
-            if (s.startsWith("go")) {
+            if (!s.startsWith("go")) {
               Pull.outputs(notGo) >> Pull.done
             } else {
               Pull.output1(s) >> newBodyP(h)
             }
-          case None => Pull.done
+          case None => Pull.outputs(notGo) >> Pull.done
         }
       }
       Ok(body.pull(newBodyP))
--- a/examples/src/main/scala/com/example/http4s/ScienceExperiments.scala
+++ b/examples/src/main/scala/com/example/http4s/ScienceExperiments.scala
@@ -1,5 +1,4 @@
-// TODO fs2 port
-/*
+
 package com.example.http4s
 
 import java.time.Instant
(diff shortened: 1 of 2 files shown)


examples/src/main/scala/com/example/http4s/ScienceExperiments.scala
```diff
@@ -65,30 +64,30 @@ object ScienceExperiments {
...
-      val parser = await1[String] map {
-        case chunk if chunk.startsWith("Go") =>
-          Task.now(Response(body = emit(chunk) ++ req.bodyAsText |> utf8Encode))
-        case chunk if chunk.startsWith("NoGo") =>
-          BadRequest("Booo!")
-        case _ =>
-          BadRequest("no data")
-      }
-      (req.bodyAsText |> parser).runLastOr(InternalServerError()).run
+//    case req@POST -> Root / "challenge1" =>
```
rossabaker: If you use a block comment here, you get a smaller diff.

examples/src/main/scala/com/example/http4s/ScienceExperiments.scala
```diff
@@ -103,21 +102,28 @@ object ScienceExperiments {
...
 
     case req @ GET -> Root / "broken-body" =>
-      Ok(Process(Task{"Hello "}) ++ Process(Task{sys.error("Boom!")}) ++ Process(Task{"world!"}))
+      Ok(Stream.eval(Task{"Hello "}) ++ Stream.eval(Task{sys.error("Boom!")}) ++ Stream.eval(Task{"world!"}))
 
-    case req @ GET -> Root / "slow-body" =>
-      val resp = "Hello world!".map(_.toString())
-      val body = awakeEvery(2.seconds).zipWith(Process.emitAll(resp))((_, c) => c)
-      Ok(body)
+//    Missing Strategy in Scope Which Is Clearly Not the Case
```
rossabaker: It can't find the `Async[Task]`?  Yeah, what's up with that?  I've had to make that explicit a few times and am not understanding why.

examples/src/main/scala/com/example/http4s/ScienceExperiments.scala
```diff
@@ -103,21 +102,28 @@ object ScienceExperiments {
...
+//      val resp = "Hello world!".map(_.toString())
+//      val body = time.awakeEvery(2.seconds).zipWith(Stream.emits(resp))((_, c) => c)
+//      Ok(body)
 
     case req @ POST -> Root / "ill-advised-echo" =>
       // Reads concurrently from the input.  Don't do this at home.
-      implicit val byteVectorMonoidInstance: Monoid[ByteVector] = Monoid.instance(_ ++ _, ByteVector.empty)
-      val tasks = (1 to Runtime.getRuntime.availableProcessors).map(_ => req.body.foldMonoid.runLastOr(ByteVector.empty))
-      val result = Task.reduceUnordered(tasks)(Reducer.identityReducer)
+      implicit val byteVectorMonoidInstance: Monoid[ByteVector] = new Monoid[ByteVector]{
```
rossabaker: We had this monoid in 0.16, and removed it when ByteVector was temporarily removed from the classpath in 0.17.  It might be worth bringing back.

ChristopherDavenport: We Technically Arn't Using `ByteVector` anymore, just `Stream[Task, Byte]` so I'm not sure its necessary as much as I was bringing the effective solution back to the experiment.

examples/src/main/scala/com/example/http4s/ScienceExperiments.scala
ChristopherDavenport: Yeah, I'll go find the call to explicitly push this through.

ChristopherDavenport:
Seems to be working, don't like the `Pull -> Task` implementation but using Pull.eval puts the value in the use state not the output state. Not sure how to fix that. It becomes a more verbose test as a result.

rossabaker:
:+1:

ChristopherDavenport:
Flipped  a boolean and one return pull was incorrect on Challenge1. Should be good to go now.

ChristopherDavenport:
One should never look at their code during CI, every time I see some little fix that should be applied that I didn't notice. Very frustrating.

PR #1191, opened 2017-05-17
Logging Middleware

ChristopherDavenport:
Most version of observe seem to consume to body and creates errors in writing length, these two combinations seem to avoid this.

Not Sure How Valid These Tests are if at all. Recreating the failures of consuming the body is hard. However I found it a couple times in unexpected places.

### DIFF
--- a/server/src/main/scala/org/http4s/server/middleware/RequestLogger.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/RequestLogger.scala
@@ -1,0 +1,29 @@
+package org.http4s
+package server
+package middleware
+
+import fs2._
+import org.log4s._
+import cats.implicits._
+import fs2.interop.cats._
+
+/**
+  * Simple Middleware for Logging Requests As They Are Processed
+  */
+object RequestLogger {
+  private[this] val logger = getLogger
+
+  def apply(service: HttpService)(implicit strategy: Strategy): HttpService = Service.lift{ req =>
+    Stream(req)
+      .observe(requestLogSink)
+      .evalMap[Task, Task, MaybeResponse]{req => service(req)}
+      .runFoldMap(identity)
+  }
+
+  def requestLogSink: Sink[Task, Request] = stream => {
+    stream
+      .flatMap(_.body)
+      .through(fs2.text.utf8Decode)
+      .evalMap[Task, Task, Unit]{ str => Task.delay(logger.info(str)) }
+  }
+}
--- a/server/src/main/scala/org/http4s/server/middleware/RequestLogger.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/RequestLogger.scala
@@ -6,6 +6,7 @@ import fs2._
 import org.log4s._
 import cats.implicits._
 import fs2.interop.cats._
+import scodec.bits._
 
 /**
   * Simple Middleware for Logging Requests As They Are Processed
@@ -13,17 +14,10 @@ import fs2.interop.cats._
 object RequestLogger {
   private[this] val logger = getLogger
 
-  def apply(service: HttpService)(implicit strategy: Strategy): HttpService = Service.lift{ req =>
+  def apply(logHeaders: Boolean, logBody: Boolean)(service: HttpService)(implicit strategy: Strategy): HttpService = Service.lift{ req =>
     Stream(req)
-      .observe(requestLogSink)
+      .through(Logger.messageLogPipe[Request](logHeaders, logBody)(logger))
       .evalMap[Task, Task, MaybeResponse]{req => service(req)}
       .runFoldMap(identity)
-  }
-
-  def requestLogSink: Sink[Task, Request] = stream => {
-    stream
-      .flatMap(_.body)
-      .through(fs2.text.utf8Decode)
-      .evalMap[Task, Task, Unit]{ str => Task.delay(logger.info(str)) }
   }
 }
--- a/server/src/main/scala/org/http4s/server/middleware/ResponseLogger.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/ResponseLogger.scala
@@ -1,0 +1,29 @@
+package org.http4s
+package server
+package middleware
+
+import fs2._
+import fs2.interop.cats._
+import cats.implicits._
+import org.log4s.getLogger
+
+/**
+  * Simple Middleware for Logging Responses As They Are Processed
+  */
+object ResponseLogger {
+  private[this] val logger = getLogger
+
+  def apply(service: HttpService): HttpService = Service.lift { req =>
+    service(req).flatMap {
+      case resp: Response =>
+        resp
+          .body
+          .through(fs2.text.utf8Decode)
+          .runFoldMap(identity)
+          .map(logger.info(_))
+          .map(_ => resp)
+      case Pass =>
+        Task.delay(Pass)
+    }
+  }
+}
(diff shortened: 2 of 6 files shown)


server/src/main/scala/org/http4s/server/middleware/ResponseLogger.scala
```diff
@@ -0,0 +1,29 @@
...
+  private[this] val logger = getLogger
+
+  def apply(service: HttpService): HttpService = Service.lift { req =>
+    service(req).flatMap {
+      case resp: Response =>
+        resp
+          .body
+          .through(fs2.text.utf8Decode)
+          .runFoldMap(identity)
+          .map(logger.info(_))
```
cquiroz: I wonder if the level could be made customizable by the users

server/src/main/scala/org/http4s/server/middleware/RequestLogger.scala
```diff
@@ -0,0 +1,29 @@
...
+  * Simple Middleware for Logging Requests As They Are Processed
+  */
+object RequestLogger {
+  private[this] val logger = getLogger
+
+  def apply(service: HttpService)(implicit strategy: Strategy): HttpService = Service.lift{ req =>
+    Stream(req)
+      .observe(requestLogSink)
+      .evalMap[Task, Task, MaybeResponse]{req => service(req)}
+      .runFoldMap(identity)
```
rossabaker: Interesting approach. I imagined `req.copy(body = req.body.through(requestLogSink)`, where `requestLogSink` was a `Sink[Task, EntityBody]`.

server/src/main/scala/org/http4s/server/middleware/RequestLogger.scala
```diff
@@ -0,0 +1,29 @@
...
+
+import fs2._
+import org.log4s._
+import cats.implicits._
+import fs2.interop.cats._
+
+/**
+  * Simple Middleware for Logging Requests As They Are Processed
+  */
+object RequestLogger {
```
rossabaker: So this is only logging the body.  Should it be called `RequestBodyLogger`?  How would we go about correlating with headers when we log those?

server/src/main/scala/org/http4s/server/middleware/RequestLogger.scala
```diff
@@ -0,0 +1,29 @@
...
+      .observe(requestLogSink)
+      .evalMap[Task, Task, MaybeResponse]{req => service(req)}
+      .runFoldMap(identity)
+  }
+
+  def requestLogSink: Sink[Task, Request] = stream => {
+    stream
+      .flatMap(_.body)
+      .through(fs2.text.utf8Decode)
+      .evalMap[Task, Task, Unit]{ str => Task.delay(logger.info(str)) }
```
rossabaker: Every chunk is going to log its own line.  Without a request ID, when requests are arriving concurrently, it's going to be exceedingly difficult to reassemble individual requests.

My gut feeling is to log requests in a single line, but that raises a problem of unbounded buffering.

server/src/main/scala/org/http4s/server/middleware/RequestLogger.scala
```diff
@@ -0,0 +1,29 @@
...
+    Stream(req)
+      .observe(requestLogSink)
+      .evalMap[Task, Task, MaybeResponse]{req => service(req)}
+      .runFoldMap(identity)
+  }
+
+  def requestLogSink: Sink[Task, Request] = stream => {
+    stream
+      .flatMap(_.body)
+      .through(fs2.text.utf8Decode)
```
rossabaker: What if the request has a declared charset of ISO-8859-1?  We could respect that.

What if it's binary?  We could sniff that out and render, oh, I don't know what.

What if it doesn't tell us what it is, but is not UTF-8?  If the sink fails, the stream fails.  Is there any sort of `attemptThrough`?  I don't think so.  Hmm.

server/src/test/scala/org/http4s/server/middleware/LoggerSpec.scala
```diff
@@ -0,0 +1,64 @@
+package org.http4s
+package server
+package middleware
+
+import org.http4s.dsl._
+
+/**
+  * Common Tests for Logger, RequestLogger, and ResponseLogger
+  */
+class LoggerSpec extends Http4sSpec {
```
rossabaker: The log statements could be tested if an [OutputStreamAppender](https://logback.qos.ch/apidocs/ch/qos/logback/core/OutputStreamAppender.html) is configured, though it might not be a great deal of fun.

server/src/main/scala/org/http4s/server/middleware/RequestLogger.scala
ChristopherDavenport: We could have a logger, that we enable different types of Logger internally. However this could get out of hand very quickly, however I agree this is not an extensible enough solution.

server/src/main/scala/org/http4s/server/middleware/RequestLogger.scala
ChristopherDavenport: I tried that approach, although through is `Pipe` and Blind Alley of Death is `Sink` and I never got a rendered body. It would display the correct headers, but the body didn't come through. Again this was a couple hours work so it's probably doable. This works, surprisingly. Every version where I attempted a copy I ended up failing to render the body.

server/src/main/scala/org/http4s/server/middleware/RequestLogger.scala
ChristopherDavenport: Pertaining to failure, that is why this happens in a side sink. This observes the request but I believe won't fail the request it is observing.(Not an expert and needs tests to confirm)

As for charset validation. Thats another area for expanded work. We could build charset pipes for rendering similar to this one although they might be better contributed to fs2. However as the next shift for fs2 involves `cats-effect` and some brand new structures would probably be better internally and then shifted to theirs after 1.0

server/src/test/scala/org/http4s/server/middleware/LoggerSpec.scala
```diff
@@ -0,0 +1,64 @@
...
+  "RequestLogger" should {
+    val requestLoggerService = RequestLogger(true, true)(testService)
+
+    "not effect a Get" in {
+      val req = Request(uri = uri("/request"))
+      requestLoggerService.orNotFound(req) must returnStatus(Status.Ok)
+    }
+
+    "not effect a Post" in {
+      val req = Request(uri = uri("/post"), method = POST).withBody(urlForm)
```
rossabaker: Have we confirmed that this leaves the body for the service when the body is not replayable?

ChristopherDavenport: Not yet, needs a test for that. Working on that one. The fact that it now buffers it and then advances it though makes me more optimistic towards it's success in this regard.

ChristopherDavenport:
Update: Have had no time to get back around to this. Basically no full tests but I feel this needs to take a backseat to `Multipart`. I'd welcome either merging and making better from here, or I'd accept a PR against this branch, or it can hold until I finish my highest priority with Multipart.

Would love ideas.

rossabaker:
I'm okay merging it if we know that a body can still be read using it.

ChristopherDavenport:
How is 80%? Because I'm definitely not positive, but I'm fairly confident. Would love for someone else to check me on it.

rossabaker:
I don't know, but I'll try to PR it this weekend with another test and then we'll all know.

rossabaker:
Build failure is because master was in bad state when this ran.

PR #1192, opened 2017-05-18
Vanity UrlForm Apply

ChristopherDavenport:
Simple Change To Allow Lists to be Entered As Well. 

To be honest I just got sick of writing `:_*` and thought others might like it as well.

### DIFF
--- a/core/src/main/scala/org/http4s/UrlForm.scala
+++ b/core/src/main/scala/org/http4s/UrlForm.scala
@@ -81,6 +81,9 @@ object UrlForm {
 
   def apply(values: (String, String)*): UrlForm =
     values.foldLeft(empty)(_ + _)
+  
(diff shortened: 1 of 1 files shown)


ChristopherDavenport:
Might be better targeted at `0.16`...

core/src/main/scala/org/http4s/UrlForm.scala
```diff
@@ -81,6 +81,9 @@ object UrlForm {
 
   def apply(values: (String, String)*): UrlForm =
     values.foldLeft(empty)(_ + _)
+  
+  def apply(values: List[(String, String)]): UrlForm =
```
rossabaker: I'm not usually a fan of `Seq`, but may be good here to support `Vector`.

ChristopherDavenport:
Check.

PR #1194, opened 2017-05-20
Multipart [WIP]

ChristopherDavenport:
Here is an initial implementation that works for Encoder and Decoder and has tests working again. Here are the steps I see that I'd like to outline

- [x] MultipartEncoder
- [x] MultipartDecoder
- [x] Tests
- [ ] Parser
    - [ ] Implement Pulls with Handle Passing
    - [ ] Parse Headers
    - [ ] Parse Body(including Binary, etc)


core/src/main/scala/org/http4s/multipart/MultipartEncoder.scala
```diff
@@ -1,66 +1,74 @@
...
-      } << Boundary.CRLF).toByteVector) ++ p.body
-
-    val parts = mp.parts
-    val body = parts.tail.foldLeft(renderPart(_start, parts.head)) { (acc, part) =>
-      acc ++ renderPart(_encapsulation, part)
-    } ++ Process.emit(_end)
-
-    Task.now(Entity(body, None))
-  }
+  def toEntity(mp: Multipart): Task[Entity] = Task.now(Entity(renderParts(mp.boundary)(mp.parts), None))
```
rossabaker: I am glad to see this odd formatting getting cleaned up.

core/src/main/scala/org/http4s/multipart/MultipartParser.scala
```diff
@@ -1,236 +1,169 @@
...
 
-  private val CRLF = ByteVector('\r', '\n')
-  private val DASHDASH = ByteVector('-', '-')
+  private val CRLF = "\r\n"
+  private val DASHDASH = "--"
+  private val startLine: Boundary => String = boundary => s"$DASHDASH${boundary.value}"
+  private val endLine: Boundary => String = boundary => s"${startLine(boundary)}$DASHDASH"
+  private val expected: Boundary => String = boundary => s"$CRLF${startLine(boundary)}"
+
+  private val CRLFBytes = ByteVector('\r','\n')
```
rossabaker: How do we end up with ByteVector and String versions of these?

core/src/main/scala/org/http4s/multipart/MultipartParser.scala
```diff
@@ -1,236 +1,169 @@
...
+
+    val streamString = s.chunks
+      .through(text.utf8DecodeC)
+      .covary[Task]
+      .through(text.lines)
+      .dropWhile(_ != startLine(boundary)) // Drop Prelude
+      .drop(1)                             // Drop Start Line
+      .takeWhile(_ != endLine(boundary))   // Take Until EndLine
+
+      val headers =
```
rossabaker: Indentation is weird here.

ChristopherDavenport:
Update: I think fastest way to get this up is to buffer and then parse. I have quite a solid async parsing program that I like but it make take a bit. Pushing into queues, and then consuming them within the method. 

Would doing a large buffer to get this operational faster be an ok middle implementation, and then working on a more lasting streaming version as a second wave.

rossabaker:
:+1: I'd rather have a suboptimal implementation than no implementation.

rossabaker:
Is this superseded by #1238 ?

ChristopherDavenport:
I would say so, the ideas were good here, but buffered will get us and running and this PR, does not aid that anymore.

PR #1197, opened 2017-05-24
WIP: Parameterize effect type and use fs2 0.10 / cats-effect

aeons:
The work in this PR was started by @rossabaker some time ago when he started parameterizing the effect type.

Since fs2 0.10 will use type classes from cats-effect instead of fs2's own, so doing both of these changes in one go makes sense.

This is a work in progress.


core/src/main/scala/org/http4s/util/ByteVectorChunk.scala
```diff
@@ -1,14 +1,13 @@
...
+import fs2.{Chunk, MonomorphicChunk}
 import scodec.bits.ByteVector
 
-// This will exist in fs2-1.0.  That version is more optimal, because
-// it has a smarter concatAll.  This will do for now.
-private[http4s] class ByteVectorChunk private (val toByteVector: ByteVector)
-    extends Chunk[Byte] {
+import scala.reflect.{ClassTag, classTag}
+
+final class ByteVectorChunk private (val toByteVector: ByteVector)
```
rossabaker: Why not use the one that's in this version of fs2?

aeons: Because it's not actually in the published jar.

The whole `fs2.interop.scodec` package is not included.

aeons: It's in `fs2-scodec`. Added it and removed this class :)

circe/src/test/scala/org/http4s/circe/CirceSpec.scala
```diff
@@ -83,28 +84,27 @@ class CirceSpec extends JawnDecodeSupportSpec[Json] {
...
+      val body = req.decode { json: Json => Response(Ok).withBody(json.asNumber.flatMap(_.toLong).getOrElse(0L).toString)}.unsafeRunSync.body
       new String(getBody(body), StandardCharsets.UTF_8) must_== "157"
     }
   }
 
   "jsonOf" should {
     "decode JSON from a Circe decoder" in {
-      val result = jsonOf[Foo].decode(Request().withBody(Json.obj("bar" -> Json.fromDoubleOrNull(42))).unsafeRun, strict = true)
-      result.value.unsafeRun must_== Right(Foo(42))
+      val result = jsonOf[IO, Foo].decode(Request[IO]().withBody(Json.obj("bar" -> Json.fromDoubleOrNull(42))).unsafeRunSync, strict = true)
```
rossabaker: This is an unfortunate consequence of the parameterization: `jsonOf[A]` was a lot prettier than `jsonOf[IO, A]`.  I don't have an easy answer for what to do about it.

dsl/src/test/scala/org/http4s/dsl/ResponseGeneratorSpec.scala
```diff
@@ -1,34 +1,42 @@
...
 import org.http4s.headers.{Accept, `Content-Length`, `Content-Type`}
 
 class ResponseGeneratorSpec extends Http4sSpec {
 
   "Add the EntityEncoder headers along with a content-length header" in {
-    val body = "foo"
-    val resultheaders = Ok(body)(EntityEncoder.stringEncoder).unsafeRun.headers
-    EntityEncoder.stringEncoder.headers.foldLeft(ok) { (old, h) =>
+    val body          = "foo"
+    val resultheaders = Ok(body)(Applicative[IO], FlatMap[IO], EntityEncoder.stringEncoder[IO]).unsafeRunSync.headers
```
rossabaker: Do any of these implicits need to be explicit?

aeons: No, not in this case. However, if they aren't there it has to be `Ok[IO, String](body)`.

In the other test cases in that file an explicit encoder is used so the type class instances have to be supplied manually.

Would be nice with multiple implicit parameter lists here.

aeons:
There is a bit more helping the compiler with types, but overall I feel it is worth it due to how clean you can write the non-http4s parts of your code.

core/src/main/scala/org/http4s/EntityEncoder.scala
```diff
@@ -148,23 +146,23 @@ trait EntityEncoderInstances extends EntityEncoderInstances0 {
...
   // TODO if Header moves to Entity, can add a Content-Disposition with the filename
-  implicit def filePathEncoder[F[_]: Suspendable]: EntityEncoder[F, Path] =
+  implicit def filePathEncoder[F[_]: Sync]: EntityEncoder[F, Path] =
     fileEncoder[F].contramap(_.toFile)
 
   // TODO parameterize chunk size
-  implicit def inputStreamEncoder[F[_]: Suspendable]: EntityEncoder[F, F[InputStream]] =
-    streamEncoder[F, Byte].contramap { in: F[InputStream] =>
-      readInputStream[F](in, DefaultChunkSize)
+  implicit def inputStreamEncoder[F[_]: Sync, IS <: InputStream]: EntityEncoder[F, F[IS]] =
```
aeons: I did this to make the tests work, without adding more specific types at the call site when called with a subclass of `InputStream`.

I have no clue if there would be any issues with this, and I don't understand why it's suddenly needed when it worked before parameterising.

tests/src/test/scala/org/http4s/EntityEncoderSpec.scala
```diff
@@ -1,64 +1,67 @@
...
-      implicitly[EntityEncoder[String]].toEntity("pong").map(_.length) must returnValue(Some(4))
+      EntityEncoder[IO, String].toEntity("pong").map(_.length) must returnValue(Some(4))
     }
 
     "render byte arrays" in {
       val hello = "hello"
       writeToString(hello.getBytes(StandardCharsets.UTF_8)) must_== hello
     }
 
+    /* TODO: fs2-0.10 futureEncoder is tricky
```
aeons: The current implementation of futureEncoder uses Task.fromFuture. There is an equivalent in cats-effect in IO.fromFuture, but we don't have access to that when the effect type is unknown.

I still think it makes sense to have an included EntityEncoder[IO, Future[A]] somewhere, since that seems like a useful instance, but I'm not sure where to put it.

circe/src/test/scala/org/http4s/circe/CirceSpec.scala
aeons: ¯\\\_(ツ)\_/¯

aeons:
In the `blaze-core` module there are quite a few `.unsafeRun` and `Task.fromFuture` calls.

Is it ok to use `cats.effect.IO` here? Since an `IO` can be lifted to any `F[_]` with a `cats.effect.Effect` instance.

tests/src/test/scala/org/http4s/EntityEncoderSpec.scala
rossabaker: I'd leave it here, with an `Async` constraint:

```scala
Async[F].async { cb =>
  f.value onComplete {
    case Failure(e) => cb(Left(e))
    case Success(a) => cb(Right(a))
  }
}
```

rossabaker:
`Task.fromFuture` can be replaced by something like the comment above.  I started a discussion in Gitter about possibly having an `Async.fromFuture` operation to avoid a conversion through `IO`.

The `.unsafeRun` requires some thinking.  If a service is in `F: Effect`, we can `runAsync` those to get to an `IO[Unit]`, and then either `unsafeRunAsync` or `unsafeRunSync` those to perform the IO.  I expect in general we shouldn't need to be sync.

circe/src/test/scala/org/http4s/circe/CirceSpec.scala
ChristopherDavenport: Explicit Side Effect is Explicit

aeons:
I've run into a rather annoying issue with type inference.

This works and `IO` is inferred from the type parameter to `HttpService` (I guess):
```scala
HttpService[IO] { case GET -> Root => Ok("Hello") }
```
This doesn't:
```scala
HttpService[IO] { 
  case GET -> Root => Ok("Hello").withContentType(Some(`Content-Type`(MediaType.`text/plain`))) 
}
```
As soon as you add types to the response created by `Ok` in some way, it works again.

```scala
val resp: IO[Response[IO]] = Ok("Hello"
// or
Ok[IO, String]("Hello")
```

rossabaker:
Yes, I think that case requires specializing the DSL. I think an abstract class that contains the DSL, where the package object extends it as IO may be a good approach.

aeons:
I have tried parameterising the dsl in 914a8dc.

I have split the `implicit class _Syntax` classes into a `final class _Ops[F[_]]` class and an `implicit def _Syntax` method in a trait that takes a type parameter.

The dsl package object then extends this trait with the type parameter set to `IO`. To use another effect type one would have to do `package object myDsl extends  Http4sDsl[monix.Task]`.

aeons:
I am starting to hit some variance issues, specifically in regards to `Response[F]` vs `MaybeResponse[F]`.

Since the type parameter of `F[_]` is invariant, an `F[Pass[F]]` is not an `F[MaybeResponse[F]]`, and neither is an `F[Response[F]]`.

(Co)variance is not something I'm well versed in, but it doesn't feel right to clutter the codebase with `.widen`s.

`fs2.Task` (and `cats.effect.IO` and I guess every other type that would be used as the effect type) are covariant in their type parameter. Would it make sense, throughout the project to require that the effect type is covariant in its type parameter? Because as I understand, it can't be a requirement just in those functions that need the covariance.

rossabaker:
What you did with the DSL is pretty much exactly what I had in mind.

There are some effect types that are not covariant that people might consider using.  [TraceT](https://github.com/ccadllc/cedi-dtrace/blob/master/core/src/main/scala/com/ccadllc/cedi/dtrace/TraceT.scala) comes to mind.  Do you have some examples of how bad the `.widen`ing would be?

aeons:
From `master`:
```scala
private def serveWebjarAsset(config: Config, request: Request)
                            (webjarAsset: WebjarAsset): Task[MaybeResponse] =
  StaticFile
    .fromResource(webjarAsset.pathInJar, Some(request))
    .fold(Pass.now)(config.cacheStrategy.cache(request.pathInfo, _))
```
What I would like it to look:
```scala
private def serveWebjarAsset[F[_]: Sync](config: Config[F], request: Request[F])
                                        (webjarAsset: WebjarAsset): F[MaybeResponse[F]] =
  StaticFile
    .fromResource(webjarAsset.pathInJar, Some(request))
    .fold(Pass[F]().pure[F])(config.cacheStrategy.cache(request.pathInfo, _))
```
(maybe with a helper method on `Pass` companion object to do `Pass.pure[F]` corresponding to the current `Pass.now` helper)

This is what I have that compiles right now:
```scala
private def serveWebjarAsset[F[_]](config: Config[F], request: Request[F])
                                  (webjarAsset: WebjarAsset)
                                  (implicit F: Sync[F]): F[MaybeResponse[F]] =
  StaticFile
    .fromResource(webjarAsset.pathInJar, Some(request))
    .fold[F[MaybeResponse[F]]](F.pure(Pass[F]()))(
      config.cacheStrategy.cache(request.pathInfo, _).widen[MaybeResponse[F]]
    )
```
This can be reduced to
```scala
private def serveWebjarAsset[F[_]: Sync](config: Config[F], request: Request[F])
                                  (webjarAsset: WebjarAsset): F[MaybeResponse[F]] =
  StaticFile
    .fromResource(webjarAsset.pathInJar, Some(request))
    .fold(Pass.pure[F])(
      config.cacheStrategy.cache(request.pathInfo, _).widen[MaybeResponse[F]]
    )
```
by moving the type coercion of `Pass` to the aforementioned helper method:
```scala
object Pass {
  def pure[F[_]](implicit F: Applicative[F]): F[MaybeResponse[F]] = F.pure(Pass[F]())
}
```

rossabaker:
I see.  And if `serveWebjarAsset` took a `F[+_]` constraint, it would compile without the widen call, correct?

This is a tricky one.  That `.widen` is irritating for authors, but better for users.  If it's just a few, I'm inclined to think we should keep the widen.  If it's widespread and really cluttering the code, then I think maybe we switch to covariant constraints.  I don't have a strong feeling yet which is right.

aeons:
`server` compile now.

There were a few `widen`, but I think we should keep `F` invariant.

There are still errors in `GzipSpec` and `StreamAppSpec` and intermittent errors in `TimeoutSpec`.

I don't know enough about what's going on in `Gzip` to really look at it.

`StreamApp` and `Timeout` required quite big changes, so that's probably me breaking them, but I wanted it committed, and since this is a WIP, I felt it was okay push broken tests.

464f611 is a nasty, huge commit, sorry.

I also merged `master` into the branch again, adding to the clutter even more.

rossabaker:
I appreciate the effort on this.  It's going to be cluttered, because master is changing too.

`StreamApp` has been having intermittent issues for a while.  I think #1203 might be a solution to that.  Or is it failing all the time for you?

rossabaker:
We're feeling a lot of pain on the master branch from intermittent `StreamApp` failures.  I wonder whether c4827ff has an answer there, while this branch continues to mature.

client/src/main/scala/org/http4s/client/package.scala
```diff
@@ -1,46 +1,58 @@
...
     val s = decoder.consumes.toList
-    EntityDecoder.decodeBy(s.head, s.tail:_*)(resp => decoder.decode(resp, strict = true).map(t => (resp.headers,t)))
+    EntityDecoder.decodeBy(s.head, s.tail: _*)(resp => decoder.decode(resp, strict = true).map(t => (resp.headers, t)))
   }
+}
 
+object Http4sClientDsl {
+
+  /** Syntax classes to generate a request directly from a [[Method]] */
+  implicit class WithBodyOps[F[_]](val method: Method with PermitsBody) extends AnyVal with EntityRequestGenerator[F]
```
rossabaker: Is this going to have similar problems to what we had in dsl with bad inference?  Are we going to want the abstract DSL trick here?

aeons: Yes, this is part of the abstract dsl thing. 

So `package object client extends Http4sClientDsl[IO]`.

blaze-core/src/main/scala/org/http4s/blaze/Http1Stage.scala
```diff
@@ -5,28 +5,29 @@ import java.nio.ByteBuffer
...
 
 /** Utility bits for dealing with the HTTP 1.x protocol */
-trait Http1Stage { self: TailStage[ByteBuffer] =>
+trait Http1Stage[F[_]] { self: TailStage[ByteBuffer] =>
   /** ExecutionContext to be used for all Future continuations
     * '''WARNING:''' The ExecutionContext should trampoline or risk possibly unhandled stack overflows */
   protected implicit def ec: ExecutionContext
 
-  private implicit def strategy: Strategy = Strategy.fromExecutionContext(ec)
+  protected implicit def F: Effect[F]
```
aeons: I added the constraint like this, instead of turning the trait into an abstract class. Not sure if this is the right way to go.

blaze-core/src/main/scala/org/http4s/blaze/util/BodylessWriter.scala
```diff
@@ -4,42 +4,53 @@ package util
...
+    F.async { cb =>
+      val callback = cb
+        .compose((t: Attempt[Unit]) => t.map(_ => close))
+        .andThen(_ => IO.unit)
+
+      pipe.channelWrite(headers).onComplete {
+        case Success(_) =>
+          p.open.close.run.runAsync(callback).unsafeRunAsync(_ => ())
+        case Failure(t) =>
+          p.pull(_ => Pull.fail(t)).run.runAsync(callback).unsafeRunAsync(_ => ())
```
aeons: I feel like this whole block can be done more elegantly.

blaze-core/src/main/scala/org/http4s/blaze/websocket/Http4sWSStage.scala
```diff
@@ -32,15 +36,15 @@ class Http4sWSStage(ws: ws4s.Websocket)(implicit val strategy: Strategy) extends
...
+  def inputstream: Stream[F, WebSocketFrame] = {
+    val t = F.async[WebSocketFrame] { cb =>
       def go(): Unit = channelRead().onComplete {
         case Success(ws)         => ws match {
             case Close(_)    =>
               for {
                 t <- deadSignal.map(_.set(true))
               } yield {
-                t.unsafeRun()
+                t.runAsync(_ => IO.unit).unsafeRunSync()
```
aeons: I can't really see a good way to get around this `unsafeRunSync`.

blaze-core/src/test/scala/org/http4s/blaze/util/EntityBodyWriterSpec.scala
```diff
@@ -232,34 +228,30 @@ class EntityBodyWriterSpec extends Http4sSpec {
 
     "write a deflated resource" in {
       val p = resource through deflate()
-      p.runLog.map(_.toArray) must returnValue(DumpingWriter.dump(p))
+      p.runLog.map(_.toArray) must returnValue(DumpingWriter.dump(resource through deflate()))
```
aeons: Running a stream that is piped through a `deflate` step gives an exception with "Deflater closed", for every way I tried to do it. I checked in fs2 0.9.7 as well, and that gave the same result.

I'm not sure how this test doesn't fail on master.

blaze-core/src/test/scala/org/http4s/blaze/util/EntityBodyWriterSpec.scala
```diff
@@ -232,34 +228,30 @@ class EntityBodyWriterSpec extends Http4sSpec {
...
     }
 
     "must be stack safe" in {
-      val p = repeatEval(Task.async[Byte]{ _(Right(0.toByte))}(Strategy.sequential)).take(300000)
```
aeons: Not sure what to replace `Strategy.sequential` with, nor whether it matters.

aeons:
Re: the StreamApp failures, c4827ff does not do a lot differently. There's a little bit of change in how the halted signal is handled, and of course, the `unsafeRun` calls are turned into monstrosities.

blaze-core/src/main/scala/org/http4s/blaze/Http1Stage.scala
rossabaker: I think this is fine.

blaze-core/src/test/scala/org/http4s/blaze/util/EntityBodyWriterSpec.scala
rossabaker: I'm confused.  You changed `dump(p)` to `dump(resource through deflate())` on master, and it passed?

blaze-core/src/test/scala/org/http4s/blaze/util/EntityBodyWriterSpec.scala
rossabaker: Sequential was used here to stress the stack safety.  A direct ExecutionContext (no trampoline, no thread pool) would be the equivalent.

blaze-core/src/test/scala/org/http4s/blaze/util/EntityBodyWriterSpec.scala
aeons: No, I didn't try changing anything on master. But no matter what version I try or how I arrange things, the second time that stream is run, it fails in the deflate step.

Since it's both run directly in the test and by the `DumpingWriter`, I don't see how this test worked to begin with.

See https://scastie.scala-lang.org/aeons/igTcf7D9SOmXv94CzXZRUA/1

blaze-server/src/main/scala/org/http4s/server/blaze/Http2NodeStage.scala
```diff
@@ -196,19 +195,19 @@ private class Http2NodeStage(streamId: Int,
...
-      }
+      async.unsafeRunAsync {
+        service(req).attempt.flatMap {
+          case Right(resp) => renderResponse(req, resp)
+          case Left(t) =>
+            Response[F](InternalServerError)
+              .withBody(s"500 Internal Service Error\n${t.getMessage}")
+              .flatMap(resp => renderResponse(req, resp))
+        }
+      }(_ => IO.unit)
```
aeons: I think this is correct, but if someone could take a look it would be great.

aeons:
There's some tests failing in `blaze-server`wrt trailer headers. I tried for a bit to see if I could fix it, but I didn't really get anywhere.

aeons:
Updating to `fs2-0.10.0-M2` added the new Stream API. This among other things adds a `Segment` type, which maybe should replace `Chunk` in some places.

rossabaker:
I think we can [get rid of this hack](https://github.com/http4s/http4s/blob/a8dff22dc3da8e7bc539342009551f8a2e5a490b/core/src/main/scala/org/http4s/StaticFile.scala#L48-L51) as of fs2-0.10.0-M2.

aeons:
@rossabaker the hack is gone and the test that was added still passes.

Everything compiles now, with locally published version of `fs2-reactive-streams` (zainab-ali/fs2-reactive-streams#25) and `jawn-fs2` (rossabaker/jawn-fs2#8).

There are some tests that fail (among others gzip middleware and trailer headers stuff). I haven't looked too deeply into them.

A couple of things that should be looked at in the future:
- Many places where we use `Chunk` should probably be `Segment` instead
- The context bounds on the effect type makes the middleware objects clunky to work with.
  What was before `BasicAuth(store, realm)(service)` is now `BasicAuth(store, realm)(implicit instance)(service)`. I'm sure this can be handled in a more elegant way, but I didn't look too much at it.

aeons:
GZip works now.

The issue with trailing headers (which will also be an issue with web sockets) is that the `AttributeKey` for the trailing header is dependent on the type of effect used. It used to be an `AttributeKey[Task[Header]]` `val`.

Since `AttributeKey` only gives reference equality (per comments in the source file), it's not possible to lookup a value by only name and type. @wedens linked to a paste in gitter with the Sigma type stuff as proposed for headers and suggested that it might be the solution here as well.

aeons:
The other errors are in `org.http4s.client.blaze.PooledHttp1Client`.

Requests made with this client times out.

rossabaker:
Would it be helpful to get this merged to that branch now, knowing that the tuts fail, and handle that and subsequent work as a separate PR?  I've followed along and this is looking good, but it has also gotten way to unwieldy for anybody else to review.

aeons:
Yes, I think so. There are several things more I want to do, but having them as separate PRs would be preferable.

Tut is giving me a weird error: 
```
error: package cats contains object and package with same name: implicits
```

Which I've been unable to fix. I've seen it before in another project where it just went away by itself.

rossabaker:
Are there two versions of cats on the classpath?  Circe and fs2 fighting?

aeons:
It sounds like it. It only happens for tut, and I can't find any conflicts in any classpath I could find.

PR #1200, opened 2017-05-26
WebjarService should pass on unmatched methods

rossabaker:
Previously, it threw a MatchError. *spit*

/cc @ZizhengTai


rossabaker:
Ugh, this was supposed to target 0.16.x.

PR #1202, opened 2017-05-27
Copy buffers when reading a URL from an InputStream

### LINKED ISSUE #1201: Corrupt output from readInputStream
When we read a chunk from an InputStream and do a second read without flushing, the content of the unflushed chunk gets overwritten by the content of the second read.

See [gist 1](https://gist.github.com/jarreds/4dde5468fb11800b48d64eb28e8b2128), [gist 2](https://gist.github.com/jarreds/c7d90b89843360c12711e017d1b42043).  This can be reproduced in [pure fs2](https://gist.github.com/rossabaker/392b0dd237f17a268d28c2b6ef874fac) because it doesn't [copy the recycled array](https://github.com/functional-streams-for-scala/fs2/blob/series/0.9/io/src/main/scala/fs2/io/JavaInputOutputStream.scala#L19).  This fs2 behavior surprises me.

We can put a hack into `StaticFile` to work around this, but I don't have a generally good solution.  Let's discuss with the fs2 experts.

/cc @jarreds

rossabaker:
`readInputStream` recycles the buffer that backs the chunk.  If we do any storage or concurrent processing down stream, the chunk mutates out from underneath us.

When reading an inputStream, we'll upgrade to `ByteVectorChunk`, which is immutable and safe for downstream consumption.

Fixes #1201 /cc @jarreds


jarreds:
Unfortunate to have to copy the bytes, but sounds like there is no other option at the moment.

+1

rossabaker:
We could also solve this on the writer side by copying only when we buffer. But then we'd be copying for all sources, not just those derived from an InputStream.  Furthermore, anything else that consumes from readInputStream in any complicated way could reintroduce the bug. I thought it was best to cut of the mutability at the source.

I think changes to fs2 could make this more efficient.

jarreds:
I'm with you on stopping near source. Thank you.

PR #1203, opened 2017-05-28
Remove laziness from arbitraries

rossabaker:
I turned on monitor watching in yourkit.  On my laptop, we blocked threads on `genStandardMethod` hundreds of times.  All of these lazy values could be locking test suites.

If we go this route, we have to be careful about the ordering of the values.  In a trait, scalac doesn't flag questionable forward references.

### DIFF
--- a/testing/src/main/scala/org/http4s/testing/ArbitraryInstances.scala
+++ b/testing/src/main/scala/org/http4s/testing/ArbitraryInstances.scala
@@ -35,18 +35,21 @@ trait ArbitraryInstances {
       list <- arbitrary[List[A]]
     } yield NonEmptyList.nel(a, list) }
 
-  lazy val genTchar: Gen[Char] = oneOf {
+  val genTchar: Gen[Char] = oneOf {
     Seq('!', '#', '$', '%', '&', '\'', '*', '+', '-', '.', '^', '_', '`', '|', '~') ++
       ('0' to '9') ++ ('A' to 'Z') ++ ('a' to 'z')
   }
 
-  lazy val genToken: Gen[String] =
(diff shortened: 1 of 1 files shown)


rossabaker:
It's passed several times in a row on 0.16.0.  We did get one [strange build](https://travis-ci.org/http4s/http4s/jobs/236814190).  A few threads are blocked on line 81 of ArbitraryInstances, and I'm not clear why.

rossabaker:
Every val needs to be initialized before anything that uses it, or we will get an NPE.  This is why a couple generators had to be moved down the file.  That's something we'll usually catch ourselves.

A user who overrides one of these generators will have a bad time.  This is an unlikely use case, but one that worked with lazy.  I guess we could make all the vals final.

```
scala> class Oops extends testing.ArbitraryInstances { override val genTchar = oneOf(Seq('a', 'b', 'c')) }
defined class Oops

scala> (new Oops).genToken.sample
java.lang.NullPointerException
  at org.scalacheck.Gen$.$anonfun$sequence$2(Gen.scala:429)
```

ChristopherDavenport:
Barring outrage, this is bound to lead to more hassles if people do try to override. 

I'm in favor of finalizing.

rossabaker:
Devil's advocate:

- If someone does it wrong, it's going to fail in tests.  Production code won't be using it.
- If someone has a particular itch, we don't prevent them from scratching it.  Early initializers still work:

```scala
scala> class Oops extends { override val genTchar = oneOf(Seq('a')) } with testing.ArbitraryInstances
res0: Option[String] = Some(aa)
```

PR #1205, opened 2017-06-02
CORS: AllowedOrigins as HOF + Access-Control-Expose-Headers

haripriyamurthy:


docs/src/main/tut/cors.md
```diff
@@ -95,12 +95,12 @@ corsMethodSvc(duckPost).unsafeRun
...
 As you can see, the CORS headers were only added to the `GET` and `POST` requests.
 Next, we'll create a configuration that limits the origins to "yahoo.com" and
-"duckduckgo.com".
+"duckduckgo.com". allowedOrigins can use any expression that resolves into a boolean.
 
 ```tut:book
 val originConfig = CORSConfig(
   anyOrigin = false,
-  allowedOrigins = Some(Set("yahoo.com", "duckduckgo.com")),
+  allowedOrigins = origin => if(Set("yahoo.com", "duckduckgo.com").contains(origin)) true else false,
```
rossabaker: `if (x) true else false` can be simplified to `x`.

`Set[String]` is a `String => Boolean`, so it could even just be `Set(...)`, but this might be less obvious.

server/src/test/scala/org/http4s/server/middleware/CORSSpec.scala
```diff
@@ -21,7 +21,12 @@ class CORSSpec extends Http4sSpec {
     anyOrigin = false,
     allowCredentials = false,
     maxAge = 0,
-    allowedOrigins = Some(Set("http://allowed.com/"))))
+    allowedOrigins = origin => if(origin == "http://allowed.com/") true else false,
```
rossabaker: Another unnecessary `if` expression

server/src/main/scala/org/http4s/server/middleware/CORS.scala
```diff
@@ -41,13 +41,13 @@ object CORS {
...
   def apply(service: HttpService, config: CORSConfig = DefaultCORSConfig): HttpService = Service.lift { req =>
-    
+
     // In the case of an options request we want to return a simple response with the correct Headers set.
     def createOptionsResponse(origin: Header, acrm: Header): Response = corsHeaders(origin.value, acrm.value)(Response())
 
     def corsHeaders(origin: String, acrm: String)(resp: Response): Response =
       config.allowedHeaders.map(_.mkString("", ", ", "")).fold(resp) { hs =>
-        resp.putHeaders(Header("Access-Control-Allow-Headers", hs))
+        resp.putHeaders(Header("Access-Control-Allow-Headers", hs), Header("Access-Control-Expose-Headers", hs))
```
rossabaker: I think `Access-Control-Allow-Headers` is supposed to go on the preflight request, whereas `Access-Control-Expose-Headers` is supposed to go on the "simple cross origin request" or "actual request"

https://www.w3.org/TR/cors/#simple-cross-origin-request

I think maybe `corsHeaders` is being shared in two cases when it needs to be split.  Am I reading the spec right?

haripriyamurthy: @rossabaker Looks like the browser looks for these headers based on which call is made. If it is get or post, it doesn't even make an options call and excepts these header in the actual call. But in case of PUT, both the pre-flight call and the actual call need these headers. Explained here - https://www.html5rocks.com/en/tutorials/cors/#toc-adding-cors-support-to-the-server

haripriyamurthy:
@rossabaker updated the PR separating allowed headers and expose headers. Allow-headers is only returned in preflight calls and expose-headers is returned in the actual call. Also returning Forbidden if origin is passed and doesn't match allowedOrigin.

server/src/test/scala/org/http4s/server/middleware/CORSSpec.scala
```diff
@@ -71,9 +78,8 @@ class CORSSpec extends Http4sSpec {
 
     "Respond with 403 when origin is not valid" in {
       val req = buildRequest("/bar").replaceAllHeaders(Header("Origin", "http://blah.com/"))
-      cors2.orNotFound(req).map((resp: Response) => resp.status.code == 403)).unsafeRun
+      cors2.orNotFound(req).map((resp: Response) => resp.status.code == 403).unsafeRun
```
rossabaker: I can't find this behavior in the spec, but as @haripriyamurthy and I discussed, this seems like the right thing to do and exists in other popular implementations (e.g., Tomcat's CORSFilter).

server/src/test/scala/org/http4s/server/middleware/CORSSpec.scala
```diff
@@ -57,6 +76,10 @@ class CORSSpec extends Http4sSpec {
       cors2.orNotFound(req).map(_.headers must contain(headerCheck _)).unsafeRun
     }
 
+    "Respond with 403 when origin is not valid" in {
+      val req = buildRequest("/bar").replaceAllHeaders(Header("Origin", "http://blah.com/"))
+      cors2.orNotFound(req).map((resp: Response) => resp.status.code == 403).unsafeRun
```
rossabaker: This behavior does is not defined in the CORS spec, but as @haripriyamurthy and I discussed, seems like the right thing to do and has precedent in other implementations (e.g., Tomcat).

PR #1206, opened 2017-06-02
Close parens in Request.toString

rexim:


rexim:
I don't know why `org.http4s.util.StreamAppSpec` fails and how it is related to my changes. :( I tried to run this specific test on my machine but the entire sbt process fails with OOM of Metaspace on ~sbt test-only org.http4s.util.StreamAppSpec~. Any clues from anybody are appreciated.

ChristopherDavenport:
That failure is unrelated although it being related to Metaspace means we can start to narrow it down.

rexim:
Yes, it's just a missed paren. :)

rossabaker:
Thanks!

rossabaker:
This has been broken a long time, so I cherry picked it back to the 0.15 series, and it'll find its way to master from there.

PR #1207, opened 2017-06-07
Use Travis Build Pipelines

ChristopherDavenport:
I believe I have taken the mystery out of the build scripts while increasing the complexity of the travis file. 

We have 1 special build that does coverage, if any of the tests fail then the build stops. 

We `source / .` for pulling into the scope and having functions terminate in the calling scripts. 

The setup script sets up the SBT Options, and functions, then very simple scripts are called. I believe this seperation should allow us to get Continuous delivery.

The special script is the build that runs mimaCompatibility and coverageTests. Although I'm still not sure why those don't seem to run. 

cc @bfritz Since you have worked so much on the CI.


ChristopherDavenport:
Here is a [Result of a Build that had an Error](https://travis-ci.org/ChristopherDavenport/http4s/builds/240577306)

.travis.yml
```diff
@@ -36,3 +24,84 @@ cache:
   - "$HOME/.coursier/cache"
   - "$HOME/.ivy2/cache"
   - "$HOME/.sbt/boot"
+
+jobs:
+  include:
+    - &tests
```
rossabaker: I don't understand this syntax or what invokes it.  Is there a reference so I can learn?

.travis.yml
```diff
@@ -36,3 +24,84 @@ cache:
...
+    - <<: *tests
+      scala: 2.10.6
+      env: SCALAZ_VERSION=7.2.13
+      script: $TRAVIS_BUILD_DIR/bin/test.sh
+    - <<: *tests
+      scala: 2.10.6
+      env: SCALAZ_VERSION=7.1.13
+      script: $TRAVIS_BUILD_DIR/bin/test.sh
+
+    - &publishSite
```
rossabaker: Don't we only publish the site from one build?  Why do we still need to specify all six?

bin/publish
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 for SCALAZ_VERSION in $(sed -ne 's/.*SCALAZ_VERSION=\(.*\)/\1/p' .travis.yml); do
-    for SCALA_VERSION in 2.12.1 2.11.8 2.10.6; do
+    for SCALA_VERSION in 2.12.2-bin-typelevel-4 2.11.11-bin-typelevel-4 2.10.6; do
```
rossabaker: Okay, this I can get behind.

.travis.yml
ChristopherDavenport: If we do then that means I can reduce it.

.travis.yml
ChristopherDavenport: It is a [yaml alias](https://github.com/cyklo/Bukkit-OtherBlocks/wiki/Aliases-(advanced-YAML-usage)). The reason I used it was to use [different environment with different builds](https://docs.travis-ci.com/user/build-stages/using-yaml-aliases/), as with build stages they didn't really get matrix's down yet. The tests could be done as a matrix, but then it would need to double a test to use.

Although this messes up the travis-ci file, it removes the entire need to worry about versions and environment variables in the scripts.

ChristopherDavenport:
Not really sure if people are going to like the Build Stages, however it makes it clear that we will not publish unless tests pass, and puts us in a great position to use sbt-rig for full deploys even across branches.

.travis.yml
ChristopherDavenport: Found that is was only that unique [primary](https://github.com/http4s/http4s/blob/release-0.16.x/bin/travis#L77)

.travis.yml
```diff
@@ -36,3 +24,63 @@ cache:
...
+      scala: 2.10.6
+      env: SCALAZ_VERSION=7.1.13
+      script: $TRAVIS_BUILD_DIR/bin/test.sh
+
+    - stage: Publish Site
+      scala: 2.12.2-bin-typelevel-4
+      env: SCALAZ_VERSION=7.2.13
+      script: $TRAVIS_BUILD_DIR/bin/publishSite.sh
+
+    - &publishSonatype
```
rossabaker: All of these cases are identical.  Is it possible to specify this once, for all six builds?

ChristopherDavenport: No because the environment changes with each case. 

I expect that matrices improve moving forward, but this is what they seem to have at this time. I could have missed something, but I'm not sure I have. 

Personally the mess is in .travis.yml  and not in the script code now. Which I am a fan of.

rossabaker: I'm a little wary because this added 60 lines of code with repeated version strings to accomplish the same result.  But when the matrix gets simplified in 0.17, I suppose the pendulum swings a bit?

ChristopherDavenport:
Update: I'm still in favor of a shift like this which moves the code into `.travis.yml` and out of a very hard to follow divergent scripts.

rossabaker:
I'm still not overly fond of the repetition, but I think I'll like the 0.17 version of it better.  Let's do it.

PR #1208, opened 2017-06-08
addCompilerPlugin for @JsonCodec

reactormonk:


reactormonk:
Looks like emacs also cleaned up some whitespace along the way.

rossabaker:
macro-paradise is required for `@JsonCodec`?  A comment would go a long way, because I'm just guessing why this was added.

reactormonk:
circe has `semiauto`, which might be what you're looking for. `auto` is a bit too far-reaching. https://gitter.im/http4s/http4s?at=5939028b6462d8493cf020d5

reactormonk:
It's required for using the `@JsonCodec` annotation. We're not using it in this tut though, I should add an example.

PR #1210, opened 2017-06-09
Method cleanups

rossabaker:
Starting a consistency pass through the model.  Some principles, which are more interesting than this PR itself:

- Partial constructors of a model from a `String` will be called `fromString`.
- An `unsafeFromString` will be provided for people who just need a value, exceptions be damned.
- Compiletime parsing of literals will be done through string interpolators.  This depends on adding contextual as a dependency, which I'm not sure I'm comfortable with.
- Registries will no longer be mutable.


core/src/main/scala/org/http4s/Method.scala
```diff
@@ -1,104 +1,115 @@
...
-      def isIdempotent: Boolean = true
-      def isSafe: Boolean = true
-    }
-  }
-
   // Type tags for a method allowing a body or not
   sealed trait PermitsBody extends Method
   sealed trait NoBody extends Method
 
+  private[http4s] class MethodImpl(name: String, val isIdempotent: Boolean = false, val isSafe: Boolean = false, val isCacheable: Boolean = false)
```
ChristopherDavenport: Perhaps remove defaults. While convenient, if we declare the state of these explicitly it makes it easier for those looking for a method to know it's behavior.

core/src/main/scala/org/http4s/Method.scala
```diff
@@ -1,104 +1,115 @@
...
-      def isSafe: Boolean = false
-    }
-    trait Safe extends Semantics {
-      def isIdempotent: Boolean = true
-      def isSafe: Boolean = true
-    }
-  }
-
   // Type tags for a method allowing a body or not
   sealed trait PermitsBody extends Method
```
ChristopherDavenport: If we are switching Method Semantics, why are `PermitsBody/NoBody` being excluded from this transition to case classes? 

From an outsider to this section of the codebase we could act on a case class boolean as easy as a sealed trait extension.

core/src/main/scala/org/http4s/Method.scala
```diff
@@ -1,104 +1,115 @@
...
 
-  import Semantics._
+  /** Resolves a method from a string. If the name is invalid, throws a
+    * [[ParseFailure]]. */
+  def unsafeFromString(s: String): Method =
+    fromString(s).fold(throw _, identity)
 
   // Lookups will usually be on fromString, so we store it wrapped in a \/-
-  private val registry = TrieMap[String, \/-[Method]]()
+  private var registry =
```
ChristopherDavenport: As this is private, as is the register method. Do we expect this to be overridden and updated somewhere or could this instead be an Immutable Map?

core/src/main/scala/org/http4s/Method.scala
rossabaker: Because those marker types are used to drive what syntax is available in the client.  Which is a detail I don't like making part of core, but then we have to do something horrid like we did for Status syntax in DSL.  I'd love to hear a better idea here.

core/src/main/scala/org/http4s/Method.scala
rossabaker: It became an immutable map.  It's a var so we can build the map as we declare the constants.

core/src/main/scala/org/http4s/Method.scala
rossabaker: It's all private, but it doesn't hurt to be explicit.

rossabaker:
I'm thinking about the syntax with `PermitsBody` and `NoBody`.

- It's awkward that the syntax is in client.  We create requests in tests, too.  Core makes some sense.  But being similar to status codes makes sense.  For status codes, it's banished to dsl, which pulls in server and shouldn't be in core because it interferes with rho.
- The build is broken now because we ascribe a `Method` type and discard the subtype.  
- The approach with status codes is to use a singleton type, as opposed to these marker traits.  That probably ought to be consistent one way or another.
- A third way might be to make each method and status code an object, which gives it a clear type.

rossabaker:
Stale, and many ideas in #3440.

Issue #1183, opened 2017-05-10
Changing Uri when copying a Request should clear pathInfo


frejucafre:
When copying a request object and supplying a new uri, e.g. `req.copy(uri = uri("/new"))`, pathInfo does not currently get cleared. This results in confusing behavior as the previous pathInfo isn't necessarily valid anymore in the new context. Thus, the more intuitive behavior would be to clear the `pathInfo` when uri is changed so that a new, "fresh" copy of the request object is obtained.

rossabaker:
I'm afraid of what this might break if we change it in 0.15, but it seems like a good thing to do for 0.16.

sportanova:
I'd like to take a crack at this - would the best way be to create a custom copy method?

reactormonk:
Just start writing one. The `uri = uri("/new")` syntax looks for the corresponding argument name in the method. Be prepared for some boilerplate.

rossabaker:
Thanks, @sportanova.

You can supply a `copy` method to hide the auto-generated one.  Just create a method called `copy` with each member of the `Uri` case class as a parameter that defaults to the current value.

rossabaker:
Fixed by #1213

PR #1213, opened 2017-06-13
update pathinfo when changing uri via Request.copy

sportanova:
https://github.com/http4s/http4s/issues/1183

What would you guys think about changing the lazy vals of scriptName + pathInfo to functions like this:

```
val caret = attributes.get(Request.Keys.PathInfoCaret).getOrElse(0)

def scriptName = {
  uri.path.splitAt(caret)._1
}

def pathInfo = {
  uri.path.splitAt(caret)._2
}
```

That way we wouldn't need to overwrite the copy method and we could avoid the boilerplate

### DIFF
--- a/core/src/main/scala/org/http4s/Message.scala
+++ b/core/src/main/scala/org/http4s/Message.scala
@@ -117,7 +117,7 @@ object Message {
   * @param body scalaz.stream.Process[Task,Chunk] defining the body of the request
   * @param attributes Immutable Map used for carrying additional information in a type safe fashion
   */
-final case class Request(
+sealed abstract case class Request(
   method: Method = Method.GET,
   uri: Uri = Uri(path = "/"),
   httpVersion: HttpVersion = HttpVersion.`HTTP/1.1`,
@@ -129,8 +129,50 @@ final case class Request(
 
   type Self = Request
 
+  private def requestCopy(
+      method: Method = this.method,
+      uri: Uri = this.uri,
+      httpVersion: HttpVersion = this.httpVersion,
+      headers: Headers = this.headers,
+      body: EntityBody = this.body,
+      attributes: AttributeMap = this.attributes
+    ): Request =
+    Request(
+      method = method,
+      uri = uri,
+      httpVersion = httpVersion,
+      headers = headers,
+      body = body,
+      attributes = attributes
+    )
+
+  @deprecated(message = "Copy method is unsafe for setting path info. Use with... methods instead", "0.17.0-M3")
+  def copy(
+    method: Method = this.method,
+    uri: Uri = this.uri,
+    httpVersion: HttpVersion = this.httpVersion,
+    headers: Headers = this.headers,
+    body: EntityBody = this.body,
+    attributes: AttributeMap = this.attributes
+  ): Request =
+    requestCopy(
+      method = method,
+      uri = uri,
+      httpVersion = httpVersion,
+      headers = headers,
+      body = body,
+      attributes = attributes
+    )
+
+  def withMethod(method: Method) = requestCopy(method = method)
+  def withUri(uri: Uri) = requestCopy(uri = uri, attributes = attributes -- Request.Keys.PathInfoCaret)
+  def withHttpVersion(httpVersion: HttpVersion) = requestCopy(httpVersion = httpVersion)
+  def withHeaders(headers: Headers) = requestCopy(headers = headers)
+  def withBody(body: EntityBody) = requestCopy(body = body)
+  def withAttributes(attributes: AttributeMap) = requestCopy(attributes = attributes)
+
   override protected def change(body: EntityBody, headers: Headers, attributes: AttributeMap): Self =
-    copy(body = body, headers = headers, attributes = attributes)
+    withBody(body).withHeaders(headers).withAttributes(attributes)
 
   lazy val authType: Option[AuthScheme] = headers.get(Authorization).map(_.credentials.authScheme)
 
@@ -140,7 +182,7 @@ final case class Request(
   }
 
   def withPathInfo(pi: String): Request =
-    copy(uri = uri.withPath(scriptName + pi))
+    withUri(uri.withPath(scriptName + pi))
 
   lazy val pathTranslated: Option[File] = attributes.get(Keys.PathTranslated)
 
@@ -222,6 +264,24 @@ final case class Request(
 }
 
 object Request {
+
+  def apply(
+    method: Method = Method.GET,
+    uri: Uri = Uri(path = "/"),
+    httpVersion: HttpVersion = HttpVersion.`HTTP/1.1`,
+    headers: Headers = Headers.empty,
+    body: EntityBody = EmptyBody,
+    attributes: AttributeMap = AttributeMap.empty
+  ) =
+    new Request(
+      method = method,
+      uri = uri,
+      httpVersion = httpVersion,
+      headers = headers,
+      body = body,
+      attributes = attributes
+    ) {}
+
 
   final case class Connection(local: InetSocketAddress, remote: InetSocketAddress, secure: Boolean)
 
--- a/tests/src/test/scala/org/http4s/MessageSpec.scala
+++ b/tests/src/test/scala/org/http4s/MessageSpec.scala
@@ -62,6 +62,28 @@ class MessageSpec extends Http4sSpec {
         Request(Method.GET).addCookie("token", "value").headers.get("Cookie".ci).map(_.value) must beSome("token=value")
       }
     }
+
+    "Request.with..." should {
+      val path1 = "/path1"
+      val path2 = "/somethingelse"
+      val attributes = AttributeMap(Seq(AttributeEntry(Request.Keys.PathInfoCaret, 3)))
+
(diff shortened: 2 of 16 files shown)


sportanova:
failed, but I don't think this is related to pathInfo... @rossabaker ?

```
01:06:32.659 [specs2.fixed.env1722864111-4] ERROR org.http4s.blaze.pipeline.Stage - Error running route: Request(method=GET, uri=/bar, headers=Headers(Transfer-Encoding: chunked))
java.lang.IllegalStateException: Attempted to collect trailers before the body was complete.
	at org.http4s.server.blaze.Http1ServerParser$$anonfun$1.apply(Http1ServerParser.scala:43)
	at org.http4s.server.blaze.Http1ServerParser$$anonfun$1.apply(Http1ServerParser.scala:42)
	at fs2.Task$$anonfun$suspend$1$$anonfun$2.apply(Task.scala:199)
[info] ClientTimeoutSpec
[info] 
[info] Http1ClientStage responses should
[info]   + Timeout immediately with an idle timeout of 0 seconds (11 seconds, 856 ms)
[info]   + Timeout immediately with a request timeout of 0 seconds (11 seconds, 191 ms)
[info]   + Idle timeout on slow response (11 seconds, 101 ms)
[error]   x Request timeout on slow response (20 seconds, 352 ms)
```

tests/src/test/scala/org/http4s/MessageSpec.scala
```diff
@@ -62,6 +62,15 @@ class MessageSpec extends Http4sSpec {
         Request(Method.GET).addCookie("token", "value").headers.get("Cookie".ci).map(_.value) must beSome("token=value")
       }
     }
+
+    "Request.copy" should {
+      "update pathInfo after Request uri is changed" in {
```
rossabaker: Does this work if the caret is present and not zero?  I expect not.

sportanova: Not changing it to def makes sense.

Thanks, I'll modify it to work with the caret

Is this what you mean by null initialization in ```Header.parses``` ?

https://github.com/http4s/http4s/blob/588690d0574aa33671660a19fc78b60dab1feab5/core/src/main/scala/org/http4s/Header.scala#L86

sportanova:
wait - i think this is totally unnecessary. ```pathInfo``` is a lazy val, and the automatically derived version of ```copy``` calls new internally, so the "newed" version is going to reinstantiate ```pathInfo``` with the new ```uri``` value. I'm running the test cases, and the new ```pathInfo``` is reflected whether or not I have a custom ```copy```.

If we want, we can still merge this in for the tests, or create a separate PR

rossabaker:
I think it works because your test case has an empty script name and no path info.  Create a request where the URI is split (i.e., the caret attribute is > 0), change the URI, and I think the problem becomes apparent.

sportanova:
I updated with a test for the caret, but it still passes - I'll test more a little later

tests/src/test/scala/org/http4s/MessageSpec.scala
```diff
@@ -62,6 +62,34 @@ class MessageSpec extends Http4sSpec {
...
+        val attributes = AttributeMap(Seq(AttributeEntry(Request.Keys.PathInfoCaret, 3)))
+
+        val originalReq = Request(uri = Uri(path = path1), attributes = attributes)
+        val updatedReq = originalReq.copy(uri = Uri(path = path2))
+
+        val caret = attributes.get(Request.Keys.PathInfoCaret).getOrElse(0)
+
+        originalReq.pathInfo mustNotEqual updatedReq.pathInfo
+        originalReq.pathInfo mustEqual originalReq.uri.path.splitAt(caret)._2
+        updatedReq.pathInfo mustEqual updatedReq.uri.path.splitAt(caret)._2
```
rossabaker: This is testing that it still splits according to the caret, but says nothing about it resetting the caret.

I think the assertions here should be:

```scala
updatedReq.scriptName must_== ""
updatedReq.pathInfo must_== "/somethingelse"
```

sportanova:
ah ok - i understand. I added a delete (--) method to ```AttributeMap```, and in ```Request.copy``` if the uri changed, I delete ```PathInfoCaret``` from the ```AttributeMap```

The problem with that is what happens when someone does ```req.copy(uri = someOtherUri, attributes = req.attributes ++ Seq(AttributeEntry(Request.Keys.PathInfoCaret, someCaret)))``` expecting to also set a new caret, but the new caret is still going to be deleted.

I could do a check to see if the caret is the same as before, and only delete then, but it's also possible that someone could provide a new uri along with a new caret that just so happens to be the same as the old caret

Maybe check for reference quality or compare against the hash on the caret? I'd have to look into those more

rossabaker:
Ooh, that's a really interesting point on providing an explicit caret.  In practice, I find it hard to imagine that many people are explicitly setting it that way, but it would be very surprising behavior.  Reference equality makes it seem even more likely to cause unexpected behavior.  We also break `request.withPathInfo` with this, which is implemented in terms of `copy`.  This is much more complicated that I expected it to be.

What if `Request` were a [sealed abstract case class](https://gist.github.com/tpolecat/a5cb0dc9adeacc93f846835ed21c92d2) and attributes moved to a second parameter list, where they don't affect equality?

`sealed abstract case class Request(
   method: Method = Method.GET,
   uri: Uri = Uri(path = "/"),
   httpVersion: HttpVersion = HttpVersion.`HTTP/1.1`,
   headers: Headers = Headers.empty,
   body: EntityBody = EmptyBody)(
   attributes: AttributeMap = AttributeMap.empty
)`

We'd have to provide our own `apply` and `copy` in the `Request` constructor.  But this corner case would go away.  It's a little weird, but I think we're going to have problems as long as the caret that we're clearing is part of the main parameter list.

sportanova:
I'm not familiar with multi-paremeter lists in case classes - how would that help (equality)?

If we go that route we'll have to change the ```Message``` trait to not include ```attributes```, because ```Request``` complains about not implementing ```attributes``` when you move it to a second parameter list.

The main problem is that we have two pieces of state (```Path``` and ```Attributes```) that need to be in sync, but their not bound together in any way.

We could avoid the boilerplate of rewriting ```copy``` and ```apply``` if instead we unite the path + caret data by changing the ```path``` field in ```Uri``` from a ```Path``` / ```String``` to a 
```
type Caret = Int

case class PathData(path: (Path, Caret)) {
  lazy val (scriptName, pathInfo) = {
    path._1.splitAt(path._2)
  }
}
```

That way ```Path``` and ```Caret``` are linked, and ```scriptName``` + ```pathInfo``` will always reflect the correct values. But we'd also have to change a bunch of things to be something like ```path.path._1``` or provide an implicit for PathData => String

rossabaker:
Moving this into a structure is good for enforcing consistency.  It means that the same string split at a different point results in unequal `Uri`, which is surprising in the context of RFC 3986 Section 6.  We could solve that by overriding equals, which is a surprising thing to do on a case class.  We could also ignore the caret in just the `Equal[Uri]` instance, which is a surprising when that instance is inconsistent with `equals`.

I experimented with moving `org.http4s.dsl.Path` into `Uri` as the path, and turned `Root` into a case class with the script name.  It works, but it changes every `Root` to `Root(_)` in every DSL match.  That's a really, really big breaking change.

This concept of `scriptName` and `pathInfo` is only useful in request routing, so it would be great to banish it further to the edges.  I haven't thought of a good way to mark what part of the path has been consumed without storing it on the request (or its URI).

I have only unsatisfying ideas so far. :(

rossabaker:
The trouble begins with the copy method. What if we just suppressed it and provided withMethod, withUri, etc.?  Then in withUri, we zero the caret. Everything else leaves it alone.

sportanova:
so we could make it a sealed abstract case class to suppress copy, then provide each of the ```with...``` methods to return a new ```Request``` with the updated value? I could get started on that

rossabaker:
@sportanova Yes, that's the only solution I can think of.  We can't fix the copy method, so we take it away.  (We could be nice and provide it and deprecate it.)  When the Request becomes sealed abstract, then an apply method on the companion will also need to be written.

sportanova:
@rossabaker I deprecated the copy method, but now I can't run the tests (error on deprecated method). I've tried adding ```-deprecated``` to the scalacoptions in build.sbt, but no luck. Any ideas?

rossabaker:
1. Rename `copy` and make it private and undeprecated.  That way, _we_ can call it safely in all those functions like `withMethod`.
2. Change all the references to `copy` to your renamed, private copy.  This will get rid of the deprecation errors.
3. Restore a deprecated `copy` that just calls your renamed, private copy.  This will help people migrate code.

core/src/main/scala/org/http4s/Message.scala
```diff
@@ -129,6 +129,30 @@ final case class Request(
 
   type Self = Request
 
+  @deprecated(message = "Copy method is unsafe for setting path info. Use with... methods instead", "")
```
rossabaker: The second argument to `@deprecated` is the version in which it's deprecated.

sportanova:
Sounds good, just updated

PR #1214, opened 2017-06-16
Fix for multithreaded xml parsing

### LINKED ISSUE #1209: Parallel parsing of xml responses produces a null
A client that parses the response using XML fails when run in parallel. It produces some null xml responses. 

For example this expression:

```
httpClient.expect[Elem](target).map { xml =>
```

Will work fine on a single request but fail when several are done in parallel. producing a `null` xml

cquiroz:
This PR is a fix for #1209 it includes tests and a fix.

The fix is up for discussion as it changes the way an implicit decoder is built. I can see the argument both way of keeping the parser injectable or create one by default, but the approach on this PR gives a saner default and will fix #1209 

If this is accepted I'll port it to the 0.17.x branch

Fixes #1209

### DIFF
--- a/scala-xml/src/main/scala/scalaxml/ElemInstances.scala
+++ b/scala-xml/src/main/scala/scalaxml/ElemInstances.scala
@@ -8,7 +8,11 @@ import scala.util.control.NonFatal
 import scala.xml._
 import scalaz.concurrent.Task
 
+import javax.xml.parsers.SAXParserFactory
+
 trait ElemInstances {
+  private val spf = SAXParserFactory.newInstance()
+
   implicit def xmlEnocder(implicit charset: Charset = DefaultCharset): EntityEncoder[Elem] =
     EntityEncoder.stringEncoder(charset)
       .contramap[Elem](xml => xml.buildString(false))
@@ -22,12 +26,13 @@ trait ElemInstances {
    * @param parser the SAX parser to use to parse the XML
    * @return an XML element
    */
-  implicit def xml(implicit parser: SAXParser = XML.parser): EntityDecoder[Elem] = {
+  implicit val xml: EntityDecoder[Elem] = {
     import EntityDecoder._
     decodeBy(MediaType.`text/xml`, MediaType.`text/html`, MediaType.`application/xml`){ msg =>
       collectBinary(msg).flatMap[Elem] { arr =>
         val source = new InputSource(new StringReader(new String(arr.toArray, msg.charset.getOrElse(Charset.`US-ASCII`).nioCharset)))
-        try DecodeResult.success(Task.now(XML.loadXML(source, parser)))
+        val saxParser = spf.newSAXParser()
+        try DecodeResult.success(Task.now(XML.loadXML(source, saxParser)))
         catch {
           case e: SAXParseException =>
             DecodeResult.failure(MalformedMessageBodyFailure("Invalid XML", Some(e)))
--- a/scala-xml/src/main/scala/scalaxml/ElemInstances.scala
+++ b/scala-xml/src/main/scala/scalaxml/ElemInstances.scala
@@ -11,7 +11,8 @@ import scalaz.concurrent.Task
 import javax.xml.parsers.SAXParserFactory
 
 trait ElemInstances {
-  private val spf = SAXParserFactory.newInstance()
+  /* Initialize a sax parser factory, users can override this if needed */
+  protected def saxFactory: SAXParserFactory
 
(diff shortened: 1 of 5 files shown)


scala-xml/src/main/scala/scalaxml/ElemInstances.scala
```diff
@@ -8,7 +8,11 @@ import scala.util.control.NonFatal
 import scala.xml._
 import scalaz.concurrent.Task
 
+import javax.xml.parsers.SAXParserFactory
+
 trait ElemInstances {
+  private val spf = SAXParserFactory.newInstance()
```
ChristopherDavenport: Is this stateful?

scala-xml/src/main/scala/scalaxml/ElemInstances.scala
```diff
@@ -22,12 +26,13 @@ trait ElemInstances {
...
    * @return an XML element
    */
-  implicit def xml(implicit parser: SAXParser = XML.parser): EntityDecoder[Elem] = {
+  implicit val xml: EntityDecoder[Elem] = {
     import EntityDecoder._
     decodeBy(MediaType.`text/xml`, MediaType.`text/html`, MediaType.`application/xml`){ msg =>
       collectBinary(msg).flatMap[Elem] { arr =>
         val source = new InputSource(new StringReader(new String(arr.toArray, msg.charset.getOrElse(Charset.`US-ASCII`).nioCharset)))
-        try DecodeResult.success(Task.now(XML.loadXML(source, parser)))
+        val saxParser = spf.newSAXParser()
```
ChristopherDavenport: Same, does this manipulate state in a way we should be considerring.

scala-xml/src/test/scala/scalaxml/ScalaXmlSpec.scala
```diff
@@ -29,6 +31,14 @@ class ScalaXmlSpec extends Http4sSpec {
       getBody(resp.body) must_== ("html".getBytes)
     }
 
+    "parse XML in parallel" in {
+      // https://github.com/http4s/http4s/issues/1209
+      val pool = Executors.newFixedThreadPool(5)
```
ChristopherDavenport: Should this use the testong thread pool from Http4sSpec?

scala-xml/src/main/scala/scalaxml/ElemInstances.scala
cquiroz: I think not. This creates a new parser in each request instead of relying on the external parser which is not thread safe

scala-xml/src/main/scala/scalaxml/ElemInstances.scala
cquiroz: That's a good question. Creating the sax parser factory actually reads the configuration to decide what parser to use. I guess it is not strictly pure though the configuration probably never changes after you start the JVM

Any ideas on how to isolate this?

scala-xml/src/test/scala/scalaxml/ScalaXmlSpec.scala
cquiroz: Right, what's the pool provided by `Http4Spec`?

rossabaker: It's in the `Http4sSpec` companion object.

scala-xml/src/main/scala/scalaxml/ElemInstances.scala
rossabaker: Make this an abstract def, and implement it as `SAXParserFactory.newInstance()` in the package object?  Then anyone needing customization can extend `ElemInstances` and import that instead of the package object.  Similar to how pretty printing is made configurable in JSON.

cquiroz:
Changed as requested

scala-xml/src/test/scala/scalaxml/ScalaXmlSpec.scala
ChristopherDavenport: [org.http4s.Http4sSpec.TestPool](https://github.com/http4s/http4s/blob/release-0.16.x/testing/src/test/scala/org/http4s/Http4sSpec.scala#L145) is the implementation I believe.

cquiroz:
Rebased and ready to merge

ChristopherDavenport:
Did I miss something? We talked about switching to TestPool, but then where the comment was made never was addressed, although it was addressed in one spot, and then it got approved and merged.

So what happened? Feels like the review process went a bit off course here.

cquiroz:
Sorry my bad, I fixed it in one spot and forgot the other

PR #1215, opened 2017-06-17
Use Verizon's Scalaz plugin

rossabaker:
Maintain it there so we don't have to here.


project/Http4sPlugin.scala
```diff
@@ -23,6 +24,13 @@ object Http4sPlugin extends AutoPlugin {
   override lazy val projectSettings: Seq[Setting[_]] = Seq(
     scalazVersionRewriter := scalazVersionRewriters.scalazStream_0_8,
 
+    // This is set in ScalazPlugin's projectSettings. Why isn't it effective for
+    // us?
+    unmanagedSourceDirectories in Compile +=
+      (sourceDirectory in Compile).value / scalazSourceDirectory(scalazVersion.value),
+    unmanagedSourceDirectories in Test +=
+      (sourceDirectory in Test).value / scalazSourceDirectory(scalazVersion.value),
```
rossabaker: I don't know why this was necessary, because the plugin does this, and we got other values from the plugin.  🤷‍♂️

ChristopherDavenport: What about other project that use this. It makes me a little worried to be setting variables that should be set by default in the upstream plugins within our own plugin.

rossabaker: Yeah, let's sit on this one until we figure out what's going on.  It works in other projects.

rossabaker:
It works fine when we have a copy of the plugin.  When we used the published one, unmanagedSourceDirectories rots: Verizon/sbt-scalaz#1

rossabaker:
With Scalaz in maintenance mode, I don't think figuring this out is worth our time.

PR #1216, opened 2017-06-17
Backport fs2-0.10 changes to StreamApp

rossabaker:
I'm not sure they're necessary, but let's see if this makes StreamAppSpec more
reliable.


rossabaker:
More complicated and didn't help.

ChristopherDavenport:
Yeah, read that and it seemed right, but was almost too messy to be maintained.

PR #1217, opened 2017-06-17
Use testPool for ScalaXMLSpec parrallel test

ChristopherDavenport:
Just the test improvement we discussed in #1214


cquiroz:
Oh, I missed that, LGTM

PR #1218, opened 2017-06-17
Fix for multithreaded xml parsing for fs2

### LINKED ISSUE #1209: Parallel parsing of xml responses produces a null

cquiroz:
This is a replay of #1214 for master, basically porting scalaz `Task` usage to fs2 `Task`. 

Fixes #1209

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -97,7 +97,7 @@ lazy val client = libraryProject("client")
     description := "Base library for building http4s clients",
     libraryDependencies += jettyServlet % "test"
   )
-  .dependsOn(core, testing % "test->test", server % "test->compile", theDsl % "test->compile")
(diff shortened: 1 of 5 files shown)


cquiroz:
Travis doesn't seem happy about this one

cquiroz:
I'm not sure what you mean by the merge train. The code is not the same though due to `Task` usage.

If there is a better way to move changes across release branches just let me know

rossabaker:
We usually merge from release-0.16.x into master.  If it's trivial, we fix it so it compiles.  If it's not trivial, we turn those merges into a PR.

Doing this as a PR shows us how to make it work on master.  We'll just want to keep this version of these files when the merge conflict from release-0.16.x hits.

PR #1219, opened 2017-06-17
Bump StreamAppSpec timeout to 10 seconds

rossabaker:
A bunch of tests are taking more than 5 seconds to complete.  This raises the technical debt ceiling while we improve the tests.

https://www.youtube.com/watch?v=v5igKuNF1rI


rossabaker:
I don't want to leave that GC debris turned on permanently.  I was hoping to see an obvious reason for the slowdown.

rossabaker:
It's a bad solution that didn't even work.

PR #1220, opened 2017-06-19
XML parsing improved safety

cquiroz:
The default XML parsing settings allows certain classes of attacks like XXE. I propose that by default the parser be set in safer mode with the settings below

if user need the features of parameter entities they can override the `saxParserFactory`

See
https://blog.compass-security.com/2012/08/secure-xml-parser-configuration/
https://github.com/scala/scala-xml/issues/17

Do you think this should target `master`?


rossabaker:
This is targeting the right place.  It will get merged to master.

PR #1222, opened 2017-06-22
CI triage pull request playground

rossabaker:
I'm just experimenting.  Commits will be force-pushed here.  Don't get attached.


rossabaker:
Fixed as described in #1224.

PR #1223, opened 2017-06-22
Merge train pipeline

ChristopherDavenport:
Merge Pipeline fixes forward from `release-0.16.x`. When we do away with watchdog values most can be moved directly into `.travis.yml`


ChristopherDavenport:
Fails correctly, all seems good. may need to semantically put the exit code check back in 0.16

rossabaker:
Merging this, as the outstanding problem seems to be fixed in #1224.

PR #1224, opened 2017-06-26
Switch to GCE build environment

rossabaker:
tl;dr some tests that took 30 seconds are subsecond

The only thing I can come up with in profiling is synchronization on class loading (`ClasspathUtilities$$anon$1`, which is a `URLClassLoader`), but it's not a major problem in local builds on a powerful MBP, a Chromebook, or @bfritz' [vagrant setup](https://github.com/http4s/http4s/compare/master...rossabaker:gce?expand=1#1035).  Brad and I theorized that if I/O is terrible in Travis, it would explain our woes.

@timperrett suggested a move back to the GCE build environment.  We [switched to containers](https://github.com/http4s/http4s/commit/79db49a673d42202a5118b7b41138886e07b109b) in order to cache dependencies. Subsequently, Travis added [caching on GCE](https://docs.travis-ci.com/user/caching/), which makes this a lot more appealing.

In the troublesome `StreamAppSpec`, the `requestShutdown` test needs to run within 5 seconds.

[Recent failure on container build](https://travis-ci.org/http4s/http4s/jobs/245000789#L2793): 

```
[info] StreamApp should
[info]   + Terminate Server on a Process Failure (32 seconds, 93 ms)
[info]   + Terminate Server on a Valid Process (30 seconds, 310 ms)
[error]   x requestShutdown Shuts Down a Server From A Separate Thread (25 seconds, 347 ms)
[error]    an exception was thrown null (StreamAppSpec.scala:53)
[error] org.http4s.util.StreamAppSpec.$anonfun$new$12(StreamAppSpec.scala:53)
```

[Success on GCE build](https://travis-ci.org/http4s/http4s/jobs/246959488#L3943-L3946):

```
[info] StreamApp should
[info]   + Terminate Server on a Process Failure (271 ms)
[info]   + Terminate Server on a Valid Process (241 ms)
[info]   + requestShutdown Shuts Down a Server From A Separate Thread (65 ms)
```

[fingers crossed]

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ notifications:
     on_success: change
     on_failure: always
     on_start: never
-sudo: false
+sudo: required
 cache:
   directories:
   - "$HOME/.coursier/cache"


.travis.yml
```diff
@@ -26,7 +26,7 @@ notifications:
     on_success: change
     on_failure: always
     on_start: never
-sudo: false
+sudo: required
```
ChristopherDavenport: Wow, if that's all it took to fix this, we've spend a lot of work dealing with travis problems...

ChristopherDavenport:
Any change of combining with #1223 ?

rossabaker:
Cherry-picked all the way back to release-0.15.x and merged back up through master.  I think this is better for us across the board.

bfritz:
Glad to hear you found a solution to the Travis woes, @rossabaker .  Would have been nice to find the root cause, but I'm content to believe sbt classloading hacks + poor I/O performance in Travis docker infrastructure is the culprit.

rossabaker:
I'm not even certain sbt is in the picture here.  It was synchronizing on an sbt implementation of a classloader, but sbt doesn't do the synchronization itself.  I think the culprit is simply slow classloading.

bfritz:
Naively, I thought sbt classloader hacks (to keep sbt deps separate from project deps in single JVM) were causing classes to be read from disk more often than is typical.  But I don't have any evidence to support that so I'm probably full of it. 🙂

Either way, it will be great to stop worrying about it.

PR #1225, opened 2017-06-27
Example of a site with automatic http to https redirects

cquiroz:
I had to augment my main project to automatically redirect http to https and I thought it would be a good example for http4s

Do you think this could be of interest in general? Should I add some documentation?

### DIFF
--- a/examples/src/main/scala/com/example/http4s/ssl/SslExampleWithRedirect.scala
+++ b/examples/src/main/scala/com/example/http4s/ssl/SslExampleWithRedirect.scala
@@ -1,0 +1,48 @@
+package com.example.http4s
+package ssl
+
+import java.nio.file.Paths
+
+import fs2._
+import org.http4s.server._
+import org.http4s.dsl._
+import org.http4s.{HttpService, Uri}
+import org.http4s.headers.Host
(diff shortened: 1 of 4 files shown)


examples/src/main/scala/com/example/http4s/ssl/SslExampleWithRedirect.scala
```diff
@@ -0,0 +1,48 @@
...
+
+  // TODO: Reference server.jks from something other than one child down.
+  val keypath = Paths.get("../server.jks").toAbsolutePath().toString()
+
+  def builder: ServerBuilder with SSLKeyStoreSupport
+
+  val redirectService = HttpService {
+    case request =>
+      request.headers.get(Host) match {
+        case Some(Host(host, _)) =>
```
rossabaker: This won't work with a request that passes an absolute URI.  Maybe `baseUri = Uri.copy(scheme = "https".ci, port = securePort)`?

cquiroz:
How would it be a middleware? That could be the case if you are in the same port and you get an http request instead of https?

rossabaker:
Oops, you're right.  It wouldn't call a wrapped service unless you were abusing ports.

cquiroz:
Updated as suggested and rebased

PR #1226, opened 2017-06-27
Unbreak release-0.15.x build

rossabaker:
I put on my cowboy hat on the GCE change and broke the 0.15 build  This is not a particularly focused commit, but it gets us back in business.


.travis.yml
```diff
@@ -31,3 +27,7 @@ cache:
   - "$HOME/.coursier/cache"
   - "$HOME/.ivy2/cache"
   - "$HOME/.sbt/boot"
+before_cache:
```
rossabaker: Taken thoughtlessly from the sbt documentation.

bin/travis
```diff
@@ -15,17 +15,15 @@ Detail: https://travis-ci.org/$TRAVIS_REPO_SLUG/builds/$TRAVIS_BUILD_ID
...
   SBT_COMMAND=";coverage ;clean ;test ;makeSite ;mimaReportBinaryIssues ;coverageReport ;coverageOff"
 else
   SBT_COMMAND=";test"
 fi
 
 if [[ $SBT_COMMAND = *";makeSite"* ]]; then
-  curl -s -L https://github.com/spf13/hugo/releases/download/v${HUGO_VERSION}/hugo_${HUGO_VERSION}_Linux-64bit.tar.gz | tar xzf -
-  mv hugo_${HUGO_VERSION}_linux_amd64/hugo_${HUGO_VERSION}_linux_amd64 $HOME/bin/hugo
+  echo "Installing hugo v${HUGO_VERSION}"
+  wget -q https://github.com/gohugoio/hugo/releases/download/v${HUGO_VERSION}/hugo_${HUGO_VERSION}_Linux-64bit.tar.gz -O - | tar xzf - -C $HOME/bin hugo
```
rossabaker: I think this might not have been necessary until I accidentally removed `mkdir -p $HOME/bin`, but I like it better.

bin/travis
```diff
@@ -44,28 +42,14 @@ if [[ $TRAVIS_BRANCH = "master" || $TRAVIS_BRANCH = "release-"* || $TRAVIS_BRANC
...
-    echo "https://github.com/http4s/http4s/issues/774#issuecomment-273981456 ."
-
-    echo "-Djava.security.egd=file:/dev/./urandom" >> "$JVM_OPTS_FILE"
-    echo
-    echo "JVM_OPTS:"
-    for opt in $(grep -v -e '^#' -e '^$' "$JVM_OPTS_FILE"); do
-        echo "  $opt"
-    done
-    echo
+  export SBT_OPTS="$SBT_OPTS -Djava.security.egd=file:/dev/./urandom"
```
rossabaker: `/etc/sbt/jvmopts` disappeared somewhere along the way, but this builds nicely on whatever is already there.

project/plugins.sbt
```diff
@@ -9,7 +9,7 @@ addSbtPlugin("org.scalastyle" %% "scalastyle-sbt-plugin" % "0.8.0")
 addSbtPlugin("org.scoverage" % "sbt-scoverage" % "1.5.0")
 addSbtPlugin("com.typesafe" % "sbt-mima-plugin" % "0.1.14")
 addSbtPlugin("com.typesafe.sbt" % "sbt-ghpages" % "0.5.4")
-addSbtPlugin("com.typesafe.sbt" % "sbt-site" % "1.2.0")
+addSbtPlugin("com.typesafe.sbt" % "sbt-site" % "1.2.1")
```
rossabaker: 1.2.0 doesn't handle Hugos >= 0.20 because of a bad regex.

bin/travis
bfritz: You can probably just remove the `-Djava.security.egd=file:/dev/./urandom"` bit.  I added that thinking we might have been blocking on lack of entropy, but I really doubt that's the case after spending more time with the build.  Too little entropy used to be a problem on some of the Travis infrastructure, but I think that's been fixed for a long time now.

PR #1228, opened 2017-06-27
Use "message" instead of "request" in failure text

andyscott:
Updates the message text in InvalidMessageBodyFailure and MalformedMessageBodyFailure to use the word "message" instead of "request" in order to avoid confusion when debugging.


rossabaker:
Cherry-picked to 0.15 and merged to 0.16 and master.  Thanks!

PR #1230, opened 2017-06-28
Fix Elif Logic

### LINKED ISSUE #1229: CheckPublishable If Statement Incorrect
[Build Example](https://travis-ci.org/http4s/http4s/builds/247745626)

[This If Statement](https://github.com/http4s/http4s/blob/release-0.16.x/bin/setup.sh#L15) isn't picking up the release branch correctly.

I'm sure it's something obvious, but the check is eluding me.

ChristopherDavenport:
Never would publish due to logic error

Fixes #1229

### DIFF
--- a/bin/setup.sh
+++ b/bin/setup.sh
@@ -12,12 +12,12 @@ function checkPublishable() {
         echo ""
         echo "Builds in Repositories other than http4s/http4s are not published to $publishLocation"
         exit 0
-    elif [[ $TRAVIS_BRANCH == "master" || $TRAVIS_BRANCH == "release-"* ]]; then
-        echo "This Build Will Be Published to $publishLocation"
-    else
+    elif [[ $TRAVIS_BRANCH != "master" && $TRAVIS_BRANCH != "release-"* ]]; then
         echo ""
         echo "As set in bin/setup this is not a publishing branch to $publishLocation"
         exit 0
+    else
+        echo "This Build Will Be Published to $publishLocation"
     fi
 }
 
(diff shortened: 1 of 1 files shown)


bin/setup.sh
```diff
@@ -12,12 +12,12 @@ function checkPublishable() {
         echo ""
         echo "Builds in Repositories other than http4s/http4s are not published to $publishLocation"
         exit 0
-    elif [[ $TRAVIS_BRANCH != "master" || $TRAVIS_BRANCH != "release-"* ]]; then
+    elif [[ $TRAVIS_BRANCH == "master" || $TRAVIS_BRANCH == "release-"* ]]; then
```
rossabaker: Couldn't this have been fixed by changing the `||` to `&&` and keeping all the failures above the fallthrough success?

ChristopherDavenport:
Hold off till tomorrow. Had an idea for a comprehensive easy fix.

ChristopherDavenport:
Yeah, So Ross wins at the Simplest Solution Game.

ChristopherDavenport:
What is the world is that build failure....

bfritz:
"curl | bash" strikes again.  My guess is the `curl` in `bash <(curl -s https://codecov.io/bash)` returned some random XML instead of a shell script.

Random security aside...
Would not surprise me if anyone with access to that script...or with the ability to mitm our dns...can steal the gpg key used to publish http4s artifacts to sonatype.  Might be wise to unset any sensitive environment variables before the "curl | bash". /cc @rossabaker

rossabaker:
I looked at the script a while back and found nothing evil but you're correct about the risks.  And the truth is I haven't looked again.  I would imagine codecov is _very_ careful about that script for the reputation of their own business, but shit happens.

PR #1234, opened 2017-07-01
hide sensitive environment variables from the Codecov "curl | bash"

bfritz:
`env -u` unsets the specified environment variables for the bash command that's running https://codecov.io/bash .  Codecov is probably very careful, but I imagine they're also a juicy target for attackers.

### DIFF
--- a/bin/travis
+++ b/bin/travis
@@ -46,5 +46,12 @@ sbt 'set scalazVersion in ThisBuild := System.getenv("SCALAZ_VERSION")' ++$TRAVI
 
 echo "Uploading codecov"
 if [[ $SBT_COMMAND = *";coverage"* ]]; then
-   bash <(curl -s https://codecov.io/bash)
+    # `env -u` to unset sensitive environment variables the Codecov script does not need to see
+    env \
+        -u encrypted_8735ae5b3321_iv \
+        -u encrypted_8735ae5b3321_key \
+        -u GITTER_WEBHOOK_URL \
+        -u SONATYPE_PASSWORD \
+        -u SONATYPE_USERNAME \
+        bash <(curl -s https://codecov.io/bash)
 fi


bfritz:
Follow-up to [my security comment](https://github.com/http4s/http4s/pull/1230#issuecomment-312108030) on #1230 .

bfritz:
@rossabaker Didn't realize you were working on something similar.  And your whitelist approach is more robust.

Either way, `ssh-add -D` makes sense.  I stopped short of that because  I wasn't sure the publishing tasks ran _before_ Codecov.  Unlike my `env` hack that only affects a subshell, `ssh-add -D` would affect any subsequent processing.  (Suspect we're adding the ssh keys right before they are used, but I have not followed recent build changes very closely and was to lazy/rushed to look across all release branches.)

Probably best to close this in favor of your more thorough changes when they are done.

bin/travis
```diff
@@ -46,5 +46,12 @@ sbt 'set scalazVersion in ThisBuild := System.getenv("SCALAZ_VERSION")' ++$TRAVI
 
 echo "Uploading codecov"
 if [[ $SBT_COMMAND = *";coverage"* ]]; then
-   bash <(curl -s https://codecov.io/bash)
+    # `env -u` to unset sensitive environment variables the Codecov script does not need to see
+    env \
```
ChristopherDavenport: While this works, I generally prefer to use `unset`

bfritz: I used `env -u` instead of `unset` because it only affects the bash child process running codecov.  The environment variables remain available for later use within `bin/travis`.  I was nervous `unset` would break publishing to sonatype or gh-pages, especially if this rolls into `release-0.16.x` and `master` where the Travis setup has changed.

But it sounds like Ross has a better version in the works anyway.  I'll close this one.

PR #1235, opened 2017-07-02
Implementation of the Age header

cquiroz:
Simple implementation of the header `Age`

### DIFF
--- a/core/src/main/scala/org/http4s/headers/Age.scala
+++ b/core/src/main/scala/org/http4s/headers/Age.scala
@@ -9,6 +9,16 @@ import scala.concurrent.duration.FiniteDuration
 import scala.concurrent.duration._
 
 object Age extends HeaderKey.Internal[Age] with HeaderKey.Singleton {
+  def apply(age: FiniteDuration): ParseResult[Age] =
+    if (age >= 0.seconds) {
+      ParseResult.success(new Age(age) {})
+    } else {
+      ParseResult.fail("Invalid age value", s"Age param $age must be more or equal to 0 seconds")
+    }
+
+  def unsafeFromDuration(age: FiniteDuration): Age =
+    apply(age).fold(throw _, identity)
+
   override def parse(s: String): ParseResult[Age] =
     HttpHeaderParser.AGE(s)
 }
@@ -20,8 +30,7 @@ object Age extends HeaderKey.Internal[Age] with HeaderKey.Singleton {
   *
   * @param age age of the response
   */
-final case class Age(age: FiniteDuration) extends Header.Parsed {
-  require(age >= 0.seconds)
+sealed abstract case class Age(age: FiniteDuration) extends Header.Parsed {
   import Age._
 
   val key = Age
--- a/core/src/main/scala/org/http4s/headers/Age.scala
+++ b/core/src/main/scala/org/http4s/headers/Age.scala
@@ -1,5 +1,31 @@
 package org.http4s
 package headers
 
-object Age extends HeaderKey.Default
+import org.http4s.util.{Renderer, Writer}
+import org.http4s.util.Renderable._
+import org.http4s.parser.HttpHeaderParser
+
+import scala.concurrent.duration.FiniteDuration
+import scala.concurrent.duration._
+
+object Age extends HeaderKey.Internal[Age] with HeaderKey.Singleton {
+  override def parse(s: String): ParseResult[Age] =
+    HttpHeaderParser.AGE(s)
+}
+
+/**
+  * Constructs an Age header.
+  *
+  * The value of this field is a positive number of seconds (in decimal) with an estimate of the amount of time since the response
+  *
+  * @param age age of the response
+  */
+final case class Age(age: FiniteDuration) extends Header.Parsed {
+  require(age >= 0.seconds)
+  import Age._
+
+  val key = Age
+  override val value = Renderer.renderString(age)
+  override def renderValue(writer: Writer): writer.type = writer.append(value)
+}
 
--- a/core/src/main/scala/org/http4s/headers/Age.scala
+++ b/core/src/main/scala/org/http4s/headers/Age.scala
@@ -7,20 +7,30 @@ import org.http4s.parser.HttpHeaderParser
 
 import scala.concurrent.duration.FiniteDuration
 import scala.concurrent.duration._
+import scala.util.Try
 
 object Age extends HeaderKey.Internal[Age] with HeaderKey.Singleton {
-  def apply(age: FiniteDuration): ParseResult[Age] =
-    if (age >= 0.seconds) {
+  def apply(age: Long): ParseResult[Age] =
+    if (age >= 0) {
       ParseResult.success(new Age(age) {})
     } else {
       ParseResult.fail("Invalid age value", s"Age param $age must be more or equal to 0 seconds")
     }
 
   def unsafeFromDuration(age: FiniteDuration): Age =
+    apply(age.toSeconds).fold(throw _, identity)
+
+  def unsafeFromLong(age: Long): Age =
     apply(age).fold(throw _, identity)
 
   override def parse(s: String): ParseResult[Age] =
     HttpHeaderParser.AGE(s)
+
+  // Render a finite duration in seconds
+  implicit val longRenderer: Renderer[Long] = new Renderer[Long] {
+    override def render(writer: Writer, d: Long): writer.type =
+      writer << d.toString
+  }
 }
 
 /**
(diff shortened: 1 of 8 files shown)


core/src/main/scala/org/http4s/headers/Age.scala
```diff
@@ -1,5 +1,31 @@
...
+
+/**
+  * Constructs an Age header.
+  *
+  * The value of this field is a positive number of seconds (in decimal) with an estimate of the amount of time since the response
+  *
+  * @param age age of the response
+  */
+final case class Age(age: FiniteDuration) extends Header.Parsed {
+  require(age >= 0.seconds)
```
cquiroz: This doesn't make me too happy but I've seen other Headers do the same

rossabaker: I'm trying to steer toward a standard for this throughout the model.  How about:

- Make it sealed abstract to suppress default `apply` and `copy` 
- `apply` returns a `ParseResult[Age]`
- `unsafeFromDuration` returns an `Age`

ChristopherDavenport: I like this standard, despite the fact I expect code to be littered with `unsafeFromDuration` everywhere in a lot of production code-bases.

cquiroz: Would `unsafeFromDuration` be an extension method on `Duration`, and fail if the requirement doesn't hold?

rossabaker: No, it would be on the `Age` companion object, as an alternative to `apply`.

I think the standard could look something like this:
- all constructors are total unless prefixed with `unsafe`
- because that's a pain, `unsafe` constructors are provided for people who don't mind throwing
- macro versions that support literals without a `ParseResult` wrapper if we can validate them at compile time.  Since there aren't literal durations, the macro is a little harder to provide here.

Another option, to avoid `unsafe` when 99% of people will do it right, is to treat all negative values as duration zero.  But I don't think I support that.

cquiroz: In the `SimpleHeadersParser` would it be ok to use `unsafe`?

```
      Digits ~ EOL ~> ((t: String) => Age.unsafe(FiniteDuration(t.toLong, SECONDS)))
```

We know from the parser that t must be positive

Or is there a better solution considering that `Age.apply` will return `ParseResult`?

core/src/main/scala/org/http4s/headers/Age.scala
```diff
@@ -1,5 +1,40 @@
...
+object Age extends HeaderKey.Internal[Age] with HeaderKey.Singleton {
+  def apply(age: FiniteDuration): ParseResult[Age] =
+    if (age >= 0.seconds) {
+      ParseResult.success(new Age(age) {})
+    } else {
+      ParseResult.fail("Invalid age value", s"Age param $age must be more or equal to 0 seconds")
+    }
+
+  def unsafeFromDuration(age: FiniteDuration): Age =
+    apply(age).right.get
```
cquiroz: I'm not that happy about this, Does this look right to you?

core/src/main/scala/org/http4s/parser/SimpleHeaders.scala
```diff
@@ -94,6 +94,12 @@ private[parser] trait SimpleHeaders {
     }
   }.parse
 
+  def AGE(value: String): ParseResult[Age] = new Http4sHeaderParser[Age](value) {
+    def entry = rule {
+      Digits ~ EOL ~> ((t: String) => Age.unsafeFromDuration(FiniteDuration(t.toLong, SECONDS)))
```
cquiroz: I'm not sure about this usage either. We know it is a positive so it should be ok, but still

core/src/main/scala/org/http4s/headers/Age.scala
rossabaker: `apply(age).fold(throw _, identity)` ?

core/src/main/scala/org/http4s/parser/SimpleHeaders.scala
rossabaker: I guess it could overflow the long, but I bet we've got that problem in a lot of places.

cquiroz:
Updated according to the last comments

rossabaker:
Some more thoughts on this:

- Nothing in the spec puts a maximum on the number of seconds.  Modeling it as anything other than a `BigInt` rejects some valid `Age` headers.  However, a `FiniteDuration` is capable of representing 570 billion years more than the age of the universe, so I think we can defer that conversation to v2. :wink:

- A `FiniteDuration` representation loses precision: `Age(1.millisecond).map(_.age)` will not return `Right(1.millisecond)`.  This is a property that exists in most of our primary constructors.  Maybe the value should be a `Long`, and the header should have a `duration` method that turns it into seconds?

ChristopherDavenport:
The second of those concerns slightly worries me. How is the precision lost? Not a blocker, just curious.

rossabaker:
Oops, my example is wrong.  It's lost when you render it and then parse it.  Rendering truncates to seconds.

cquiroz:
on RFC 2616 we have this:
```
If a cache receives a value larger than the largest positive
integer it can represent, or if any of its age calculations
overflows, it MUST transmit an Age header with a value of
2147483648 (2^31). 
```

I understand this like the `Age` value cannot be more than `MaxInt`, perhaps a simple `Int` for the field is better and `unsafeFromDuration` fails with values larger than that

cquiroz:
Or a `Long` an store it as milliseconds for greater precision but still with the upper limit?

rossabaker:
The latest RFC refers to [this definition of delta-seconds](https://tools.ietf.org/html/rfc7234#section-1.2.1).  A Long exceeds the standard, exceeds any reasonable use, and is "the greatest positive integer [we] can conveniently represent."

I wouldn't store it as millseconds, because we can only render an integer.  The advantage to accepting and converting to a duration is that the user doesn't need to remember whether that integer is seconds or milliseconds.

cquiroz:
I kind of got lost in the discussion. Do you think there is something else to change?

rossabaker:
I think it's desirable that when we create a header through the primary constructor, it round trips through rendering and parsing.

```
forAll { fd: FiniteDuration =>
  Age(fd).map(_.render).flatMap(Age.parse) == Age(fd)
}
```

When we `render` from the duration representation, we lose fractional seconds, and the property will not hold.

Furthermore, the maximum value of a finite duration is not 2^31 seconds, but 2^31 nanoseconds.  That's 292 years, but falls short of what the spec demands.

If it's `case class Age(seconds: Long)`, we could preserve the round trip property:

```
forAll { l: Long => (l > 0) ==>
  Age(l).map(_.render).flatMap(Age.parse) == Age(l)
}}
```

Then `Age` might have some helpers for converting to a duration:

```
// Because we represent a wider range than `FiniteDuration` does
def duration: Option[FiniteDuration] = Try(seconds.seconds).toOption
// Because who the hell is doing ages greater than 292 years in practice
def unsafeDuration: FiniteDuration = seconds.seconds
```

cquiroz:
Updated a suggested

core/src/main/scala/org/http4s/headers/Age.scala
```diff
@@ -7,20 +7,30 @@ import org.http4s.parser.HttpHeaderParser
...
+    apply(age.toSeconds).fold(throw _, identity)
+
+  def unsafeFromLong(age: Long): Age =
     apply(age).fold(throw _, identity)
 
   override def parse(s: String): ParseResult[Age] =
     HttpHeaderParser.AGE(s)
+
+  // Render a finite duration in seconds
+  implicit val longRenderer: Renderer[Long] = new Renderer[Long] {
```
rossabaker: One last thing: couldn't this go into renderer?  This has nothing to do with age.

cquiroz: Sure, I already moved the FiniteDuration renderer to `Renderer`

cquiroz:
Should other headers that have `require` on their constructors receive the same treatement?

core/src/main/scala/org/http4s/headers/Age.scala
cquiroz: Done!

rossabaker:
I think they should.  I think it's good for our model that anything that can throw is marked `unsafe`.  But that can be a different PR.

cquiroz:
I agree, i'll try to take those others up to speed

PR #1237, opened 2017-07-03
FS2 0.10: Standardise on ExecutionContext

aeons:
I've changed all the remaining places that ExecutorService was used to be ExecutionContext.

There are still some ExecutorServices around to create some pools for different things.

There are some comments on no using a default pool for servers and clients, but I haven't changed anything wrt. those.


rossabaker:
How hard would it be to do this on master?  It would have to be converted to a `Strategy` in some places, but steer people toward doing the right thing now.

aeons:
I'll take a look

aeons:
Merged in #1243

ChristopherDavenport:
:tada:

PR #1238, opened 2017-07-04
Multipart Buffer Possible Solution

ChristopherDavenport:
This is failing somehow in the decoder/encoder, as it appears that the parser is behaving as intended. I'm hoping another set of eyes can determine what I am missing. Because with the parser working I'm not sure where I am going wrong.

This uses a buffer however it seems to work almost across the board. The places I am worried is in the encoder and decoder now more than the parse but let me know. 

I'm hoping someone can see what I'm missing because I've been looking over this code and lost sight of where I have gone wrong. With my 2 last failing tests. I know this needs a ton more documentation.


rossabaker:
Are streams being compared directly here, rather than the bytes to which they run?  The test failures are showing a comparison between streams with some opaque effects, and I can't tell if they're equal, or if they're not, what's wrong.

ChristopherDavenport:
You see the problem. 

I can see if I can boil this down to something much more easily testable for everything beside the binary. For text based content we should be able to use `decodeAscii` or `decodeUtf8`. 

I was very lost last night, but I'm starting to feel like I might understand how to make testing a bit easier to interpret. 

Although I'm still not sure where the decoder/encoder may be going wrong. Especially because the parser seems to be working.

rossabaker:
I guess what I'm saying is I can't even tell if the decoder is going wrong.  Stream equality is not going to work.  I would not be shocked if running the streams to bytes reveals that it's working fine, and that it's a bad test.

If not, we'll have a much bigger clue. :)

ChristopherDavenport:
So looks like it is having trouble with Headers in the Second Parts, but works correctly for the first part. So, a little more, and we can switch back to a stream after this is debugged as a Simple Task.

core/src/main/scala/org/http4s/Header.scala
```diff
@@ -121,4 +122,7 @@ object Header {
       writer
     }
   }
+
+  implicit val headerEq : Eq[Header] = Eq.by(_.value)
```
rossabaker: Should this be name and value?  This seems a little scary.

core/src/main/scala/org/http4s/multipart/Part.scala
```diff
@@ -0,0 +1,53 @@
...
+      readInputStream(Task.delay(in), ChunkSize).runLog.map(ByteVector(_)))
+  }
+
+  implicit def partEq : Eq[Part] = Eq.instance[Part] { case (a, b) =>
+    a.headers === b.headers &&
+      {
+        for {
+          abv <- a.body
+          bbv <- b.body
+        } yield abv === bbv
```
rossabaker: Where did we get this `Eq` on the bodies?  This one seems fine for tests but shady for production.

ChristopherDavenport: Agreed, will make the shift of this and Multipart Eq into tests.

core/src/main/scala/org/http4s/Header.scala
ChristopherDavenport: Yikes, nice catch.

ChristopherDavenport:
CI passed. :tada:

tests/src/test/scala/org/http4s/multipart/MultipartSpec.scala
```diff
@@ -1,77 +1,85 @@
...
-    val result     = decoded.run.run
-    
-    result must beRight.like { case mp => mp must beTypedEqualTo(multipart, Equal[Multipart].equal) }
+    val result     = decoded.value.unsafeRun
+
+    result must beRight.like { case mp =>
+      mp === multipart
+//            mp.headers must_=== multipart.headers
+//            mp.parts.map(_.headers) must_=== multipart.parts.map(_.headers)
+//      mp.parts.map(_.body.map(_.decodeUtf8).unsafeRun) must_=== multipart.parts.map(_.body.map(_.decodeUtf8).unsafeRun)
```
rossabaker: Do we need this zombie code?

tests/src/test/scala/org/http4s/multipart/MultipartParserSpec.scala
```diff
@@ -260,14 +286,58 @@ object MultipartParserSpec extends Specification {
...
+      val results: Stream[Task, Either[Headers,Byte]] = unspool(input).through(MultipartParser.parse(boundary))
+
+      val (headers, bv) = results.runLog.unsafeRun().foldLeft((Headers.empty, ByteVector.empty)) {
+        case ((hsAcc, bvAcc), Right(byte)) => (hsAcc, bvAcc ++ ByteVector.fromByte(byte))
+        case ((hsAcc, bvAcc), Left(hs)) => (hsAcc ++ hs, bvAcc)
+      }
+
+      // I have Changed This Test.
+      // It previously only returned bar.
+      // What spec needs to be met that drops the first section?
```
rossabaker: This is kind of an odd test.  It's smushing together the parts that the parser is probably separting pretty well?

core/src/main/scala/org/http4s/multipart/Boundary.scala
```diff
@@ -0,0 +1,38 @@
...
+  // the boundary in the media type, which causes some implementations
+  // pain.
+  private val OTHER = """'()+_,-./:=""".toSeq
+  private val CHARS = (DIGIT ++ ALPHA ++ OTHER).toList
+  private val nchars = CHARS.length
+  private val rand = new Random()
+
+  private def nextChar = CHARS(rand.nextInt(nchars - 1))
+  private def stream: scala.Stream[Char] = scala.Stream continually (nextChar)
+  //Don't use filterNot it works for 2.11.4 and nothing else, it will hang.
```
rossabaker: Not new, and not a blocker for this, but what in the heck is this about?

tests/src/test/scala/org/http4s/multipart/MultipartSpec.scala
ChristopherDavenport: Not really, it survived as it let me test parts separate, now that Eq instances are in scope this could be axed.

ChristopherDavenport: Not really, it survived as it let me test parts separate, now that Eq instances are in scope this could be axed.

tests/src/test/scala/org/http4s/multipart/MultipartParserSpec.scala
ChristopherDavenport: It is, however I wasnt trying to get Parts. It was more of a challenge for someone to tell me why it only returned "bar" before.

Its also a really simple test.

core/src/main/scala/org/http4s/multipart/Boundary.scala
ChristopherDavenport: I honestly have no idea. Our implementation of random, with a ton of boilerplate?

tests/src/test/scala/org/http4s/multipart/MultipartParserSpec.scala
rossabaker: Because the previous test reset the accumulator when it received a header:

```scala
  case ((hsAcc, bvAcc), -\/(hs)) => (hsAcc ++ hs, ByteVector.empty)
```

ChristopherDavenport:
Any idea why this is now failing on totally unrelated code?

PR #1239, opened 2017-07-04
Standardise on ExecutionContext

aeons:
Change uses of `ExecutorService` (left over from scalaz days) to `ExecutionContext` (used in cats-effect) to ease the transition in the future.


blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientConfig.scala
```diff
@@ -47,7 +47,7 @@ final case class BlazeClientConfig(// HTTP properties
 
                                    // pipeline management
                                    bufferSize: Int,
-                                   customExecutor: Option[ExecutorService],
+                                   customExecutionContext: Option[ExecutionContext],
```
rossabaker: `customExecutionContext` was so we knew if it was a default, and that we have to shut it down ourselves.  What if we eliminate the option and just use `ExecutionContext.global` as the default?

aeons: Sounds good to me.

tomcat/src/main/scala/org/http4s/server/tomcat/TomcatBuilder.scala
```diff
@@ -180,7 +177,7 @@ object TomcatBuilder extends TomcatBuilder(
   socketAddress = ServerBuilder.DefaultSocketAddress,
   // TODO fs2 port
   // This is garbage how do we shut this down I just want it to compile argh
```
ChristopherDavenport: Can probably remove this now that we are switching.

blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientConfig.scala
ChristopherDavenport: :+1:

blaze-client/src/main/scala/org/http4s/client/blaze/bits.scala
```diff
@@ -21,12 +21,13 @@ private[blaze] object bits {
...
-    case Some(exec) => (exec, Task.now(()))
-    case None =>
-      val exec = threads.newDaemonPool("http4s-blaze-client")
-      (exec, Task.delay(exec.shutdown()))
-  }
+  def getExecutionContext(config: BlazeClientConfig): (ExecutionContext, Task[Unit]) =
+    config.customExecutionContext match {
+      case Some(ec) => (ec, Task.now(()))
+      case None =>
+        val exec = threads.newDaemonPool("http4s-blaze-client")
```
ChristopherDavenport: It seems like this should use `org.http4s.util.threads.DefaultExecutionContext` as it is in core so should be available.

aeons:
So, should the default be `ExecutionContext.global` or `org.http4s.util.threads.DefaultExecutionContext`?

aeons:
Also, does this mean that all the stuff in `blaze-client` that takes both a config and an `ExecutionContext` could be changed to just take an `ExecutionContext`?

blaze-client/src/main/scala/org/http4s/client/blaze/bits.scala
ChristopherDavenport: So, if I understand correctly. It sounds like standardizing on `ExecutionContext.global` where it is used and getting rid of the option at the edges. 

Then like everything on this branch, get functional first and then we will see where performance improvements can be made.

ChristopherDavenport:
As best I can see, this looks good. Not seeing an obvious hole anywhere. 

:+1:

jetty/src/main/scala/org/http4s/server/jetty/JettyBuilder.scala
```diff
@@ -66,15 +65,14 @@ sealed class JettyBuilder private(
   override def bindSocketAddress(socketAddress: InetSocketAddress): JettyBuilder =
     copy(socketAddress = socketAddress)
 
-  override def withServiceExecutor(serviceExecutor: ExecutorService): JettyBuilder =
-    copy(serviceExecutor = serviceExecutor)
+  override def withExecutionContext(executionContext: ExecutionContext): Self =
```
rossabaker: Actually, one more thing: we could preserve `withServiceExecutor` as a deprecated method, and construct an EC from it and delegate to `withExecutionContext`.  That will help people transition.

PR #1242, opened 2017-07-06
Tolerate `[` and `]` in queries parsing URIs

### LINKED ISSUE #1240: Uri parsing bug with brackets in query param name
> I'm trying to parse `Uri.fromString("http://localhost:8080/index?filter[state]=public")` with no success: `Left(org.http4s.ParseFailure: Invalid URI: Invalid input '[', expected '/', 'EOI', '#', '?' or Pchar (line 1, column 35)`.

Fix should target 0.15.  I ran out of time tonight, but I'll try to fix this tomorrow.

rossabaker:
If we get a `[` or `]` in a query string, parse it by encoding it.  These are the only two reserved characters in RFC 3986 that are neither explicitly permitted in a query string nor preserve special semantics (`#` delimits the fragment).  Parsing them permits real world usage, but encoding them keeps our renderings in strict compliance with the spec.

Fixes #1240 

/cc @TimothyKlim


rossabaker:
Hmm.  For consistency, we should probably permit the same for fragments, for the same reasons.

PR #1244, opened 2017-07-07
Improvement to Content-Length header

cquiroz:
After the implementation of the `Age` header it was suggested to review other cases where headers could throw an exception on construction. The other header doing that was `Content-Length`. Given how much more use `Content-Length` has this PR is quite extensive on changes to support this new style

Unfortunately this change breaks the API as the `apply` method now returns a `ParseResult[Content-Length]` value

### DIFF
--- a/core/src/main/scala/org/http4s/Message.scala
+++ b/core/src/main/scala/org/http4s/Message.scala
@@ -66,7 +66,7 @@ sealed trait Message extends MessageOps { self =>
   def withBody[T](b: T)(implicit w: EntityEncoder[T]): Task[Self] = {
     w.toEntity(b).map { entity =>
       val hs = entity.length match {
-        case Some(l) => `Content-Length`(l)::w.headers.toList
+        case Some(l) => `Content-Length`(l).fold(_ => w.headers.toList, _ :: w.headers.toList)
         case None    => w.headers
       }
       change(body = entity.body, headers = headers ++ hs)
--- a/core/src/main/scala/org/http4s/Message.scala
+++ b/core/src/main/scala/org/http4s/Message.scala
@@ -67,7 +67,7 @@ sealed trait Message extends MessageOps { self =>
   def withBody[T](b: T)(implicit w: EntityEncoder[T]): Task[Self] = {
     w.toEntity(b).flatMap { entity =>
       val hs = entity.length match {
-        case Some(l) => `Content-Length`(l).fold(_ =>
+        case Some(l) => `Content-Length`.fromLong(l).fold(_ =>
           Task.now {
             Message.logger.warn(s"Attempt to provide a negative content length of $l")
             w.headers.toList
--- a/core/src/main/scala/org/http4s/Message.scala
+++ b/core/src/main/scala/org/http4s/Message.scala
@@ -11,6 +11,7 @@ import fs2.text._
 import org.http4s.headers._
 import org.http4s.util.nonEmptyList._
 import org.http4s.server.ServerSoftware
+import org.log4s.getLogger
 
 /**
  * Represents a HTTP Message. The interesting subclasses are Request and Response
@@ -64,12 +65,17 @@ sealed trait Message extends MessageOps { self =>
     * @return a new message with the new body
     */
   def withBody[T](b: T)(implicit w: EntityEncoder[T]): Task[Self] = {
-    w.toEntity(b).map { entity =>
+    w.toEntity(b).flatMap { entity =>
       val hs = entity.length match {
-        case Some(l) => `Content-Length`(l).fold(_ => w.headers.toList, _ :: w.headers.toList)
-        case None    => w.headers
+        case Some(l) => `Content-Length`(l).fold(_ =>
+          Task.now {
+            Message.logger.warn(s"Attempt to provide a negative content length of $l")
+            w.headers.toList
+          },
+          cl => Task.now(cl :: w.headers.toList))
+        case None    => Task.now(w.headers)
       }
-      change(body = entity.body, headers = headers ++ hs)
+      hs.map(newHeaders => change(body = entity.body, headers = headers ++ newHeaders))
     }
   }
 
@@ -100,6 +106,7 @@ sealed trait Message extends MessageOps { self =>
 }
 
 object Message {
+  private[http4s] val logger = getLogger
   object Keys {
     val TrailerHeaders = AttributeKey.http4s[Task[Headers]]("trailer-headers")
   }
--- a/core/src/main/scala/org/http4s/headers/Content-Length.scala
+++ b/core/src/main/scala/org/http4s/headers/Content-Length.scala
@@ -12,17 +12,23 @@ import org.http4s.util.Writer
   *
   * @param length the length; throws an `IllegalArgumentException` if negative
   */
-final case class `Content-Length`(length: Long) extends Header.Parsed {
-  require(length >= 0L)
-
+sealed abstract case class `Content-Length`(length: Long) extends Header.Parsed {
   override def key: `Content-Length`.type = `Content-Length`
   override def renderValue(writer: Writer): writer.type = writer.append(length)
+  def modify(f: Long => Long): `Content-Length` = new `Content-Length`(f(length)) {}
 }
 
 object `Content-Length` extends HeaderKey.Internal[`Content-Length`] with HeaderKey.Singleton {
+  val zero = new `Content-Length`(0) {}
+
+  def apply(length: Long): ParseResult[`Content-Length`] = fromLong(length)
+
   def fromLong(length: Long): ParseResult[`Content-Length`] =
-    if (length >= 0L) ParseResult.success(`Content-Length`(length))
+    if (length >= 0L) ParseResult.success(new `Content-Length`(length) {})
     else ParseResult.fail("Invalid Content-Length", length.toString)
+
+  def unsafeFromLong(length: Long): `Content-Length` =
+    fromLong(length).fold(throw _, identity)
 
   def parse(s: String): ParseResult[`Content-Length`] =
     HttpHeaderParser.CONTENT_LENGTH(s)
--- a/core/src/main/scala/org/http4s/headers/Content-Length.scala
+++ b/core/src/main/scala/org/http4s/headers/Content-Length.scala
@@ -15,15 +15,13 @@ import org.http4s.util.Writer
 sealed abstract case class `Content-Length`(length: Long) extends Header.Parsed {
   override def key: `Content-Length`.type = `Content-Length`
   override def renderValue(writer: Writer): writer.type = writer.append(length)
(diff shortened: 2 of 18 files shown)


core/src/main/scala/org/http4s/headers/Content-Length.scala
```diff
@@ -12,18 +12,24 @@ import org.http4s.util.Writer
...
+sealed abstract case class `Content-Length`(length: Long) extends Header.Parsed {
   override def key: `Content-Length`.type = `Content-Length`
   override def renderValue(writer: Writer): writer.type = writer.append(length)
+  def modify(f: Long => Long): `Content-Length` = new `Content-Length`(f(length)) {}
 }
 
 object `Content-Length` extends HeaderKey.Internal[`Content-Length`] with HeaderKey.Singleton {
+  val zero = new `Content-Length`(0) {}
+
+  def apply(length: Long): ParseResult[`Content-Length`] = fromLong(length)
```
cquiroz: `apply` is now exactly like `fromLength` we could delete it as well. What do you think?

core/src/main/scala/org/http4s/parser/SimpleHeaders.scala
```diff
@@ -57,7 +57,7 @@ private[parser] trait SimpleHeaders {
   }
 
   def CONTENT_LENGTH(value: String): ParseResult[`Content-Length`] = new Http4sHeaderParser[`Content-Length`](value) {
-    def entry = rule { Digits ~ EOL ~> {s: String => `Content-Length`(s.toLong)} }
+    def entry = rule { Digits ~ EOL ~> {s: String => `Content-Length`.unsafeFromLong(s.toLong)} }
```
cquiroz: This usage makes me a bit nervous but on the other hand we know that s is positive from the parsing rules

core/src/main/scala/org/http4s/headers/Content-Length.scala
ChristopherDavenport: I'm rather ambivalent.

I'm trying to decide how to help folks transition, as this is a large break to the API.

rossabaker:
Yeah, this one is unfortunate fallout from that standard.  I hope people aren't creating this one directly very often.

`apply` could treat everything less than 0 as 0 and avoid any breakage, but that idea makes me sad.

We could add an unsigned long type.  It's not totally ridiculous: we have a case insensitive string.  Then the long version could be a deprecated overload.  It becomes more justifiable the more of these we have in the spec.  (Come to think of it, shouldn't Age be unsigned?)

🤷‍♂️

cquiroz:
Yes, Age should be unsigned, that's why we have the constructor checking for less than 0. The idea of an `UnsignedLong` or `PositiveLong` seems fine to me.

core/src/main/scala/org/http4s/headers/Content-Length.scala
```diff
@@ -12,18 +12,24 @@ import org.http4s.util.Writer
   *
   * @param length the length; throws an `IllegalArgumentException` if negative
   */
-final case class `Content-Length`(length: Long) extends Header.Parsed {
-  require(length >= 0L)
-
+sealed abstract case class `Content-Length`(length: Long) extends Header.Parsed {
   override def key: `Content-Length`.type = `Content-Length`
   override def renderValue(writer: Writer): writer.type = writer.append(length)
+  def modify(f: Long => Long): `Content-Length` = new `Content-Length`(f(length)) {}
```
cquiroz: Now that I think about this, this a bogus method. It would let you build a negative content length

core/src/main/scala/org/http4s/Message.scala
```diff
@@ -66,7 +66,7 @@ sealed trait Message extends MessageOps { self =>
   def withBody[T](b: T)(implicit w: EntityEncoder[T]): Task[Self] = {
     w.toEntity(b).map { entity =>
       val hs = entity.length match {
-        case Some(l) => `Content-Length`(l)::w.headers.toList
+        case Some(l) => `Content-Length`(l).fold(_ => w.headers.toList, _ :: w.headers.toList)
```
bryce-anderson: I'd probably log a warning or something if the entity has an invalid length.

core/src/main/scala/org/http4s/headers/Content-Length.scala
```diff
@@ -12,18 +12,24 @@ import org.http4s.util.Writer
   *
   * @param length the length; throws an `IllegalArgumentException` if negative
   */
-final case class `Content-Length`(length: Long) extends Header.Parsed {
-  require(length >= 0L)
-
+sealed abstract case class `Content-Length`(length: Long) extends Header.Parsed {
```
bryce-anderson: I know I'm a bit late to this party, but I'm not a huge fan of using a sealed abstract case class with a bunch of anonymous instances: `zero` is a unique class, the result of `modify` is a unique class, etc. This isn't a problem if you either funnel all calls through one constructor or, IMO better, make a 
```scala
private class ContentLenthImpl(l: Length) extends `Content-Length`(l)`
```
in the object to use.

core/src/main/scala/org/http4s/parser/SimpleHeaders.scala
bryce-anderson: Yeah, I wouldn't worry about it: worst case is numeric overflow on the `.toLong` and that won't yield a negative Long.

core/src/main/scala/org/http4s/headers/Content-Length.scala
```diff
@@ -12,18 +12,24 @@ import org.http4s.util.Writer
...
   override def key: `Content-Length`.type = `Content-Length`
   override def renderValue(writer: Writer): writer.type = writer.append(length)
+  def modify(f: Long => Long): `Content-Length` = new `Content-Length`(f(length)) {}
 }
 
 object `Content-Length` extends HeaderKey.Internal[`Content-Length`] with HeaderKey.Singleton {
+  val zero = new `Content-Length`(0) {}
```
bryce-anderson: How about
```scala
val zero: `Content-Length` = unsafeFromLong(0)
```

core/src/main/scala/org/http4s/Message.scala
rossabaker: Agreed.

core/src/main/scala/org/http4s/headers/Content-Length.scala
rossabaker: I liked the `sealed abstract case class` because it's the only way to reclaim `apply` and suppress `copy`.  But yeah, I've used the `*Impl` name as the lone implementation in a few other places and generally like it.

rossabaker:
That's a good point on the domain of an unsigned long.  There are very few places we wouldn't be unsafely constructing our `ULong` from a `Long`, so it's inconvenience without any ultimate extra safety.  he only place it would be a win is if we could use `ULong` literals, which is rare.  And even these could be validated by a macro.

cquiroz:
Yes, you would move the checks from the header to the construction of the unsigned integer/long. About literals, beside 0 most/all the other usages are on tests and I don't mind too much using `unsafeFromLong` on a test

core/src/main/scala/org/http4s/headers/Content-Length.scala
cquiroz: That's a good idea, i'll switch to that construct instead of the anonymous classes

core/src/main/scala/org/http4s/headers/Content-Length.scala
cquiroz: 👍

core/src/main/scala/org/http4s/Message.scala
cquiroz: This is maybe a more general question. I've seen in the codebase log calls in certain places but you could think that they should be wrapped on e.g. `Task` as side effects. I'm ok doing it either way but just want to be consistent

cquiroz:
Updated according to the comments.  The part i'm unsure is about logging in the `Message` class. I don't know a good way to get a handle of a `logger` at that level. Any suggestions?

core/src/main/scala/org/http4s/Message.scala
rossabaker: A disabled log statement is negligible, whereas a Task of a disabled log statement will never be elided.  So one has to be careful logging in a task in a hotspot.  That said, it's only going to happen in rare cases here, so I'd probably do it right and do it in a Task.

rossabaker:
The log4s docs have a nice discussion of [instance or static](https://github.com/Log4s/log4s#instance-or-static) for getting a reference to the logger in highly instantiated classes.  This might be a good case for a logger in a `Message` companion.

cquiroz:
Rebased and added logging as suggested

core/src/main/scala/org/http4s/headers/Content-Length.scala
```diff
@@ -9,21 +9,29 @@ import org.http4s.util.Writer
...
   override def renderValue(writer: Writer): writer.type = writer.append(length)
+  def modify(f: Long => Long): Option[`Content-Length`] = `Content-Length`.apply(f(length)).right.toOption
 }
 
 object `Content-Length` extends HeaderKey.Internal[`Content-Length`] with HeaderKey.Singleton {
+  private class ContentLengthImpl(length: Long) extends `Content-Length`(length)
+
+  val zero: `Content-Length` = new ContentLengthImpl(0) {}
+
+  def apply(length: Long): ParseResult[`Content-Length`] = fromLong(length)
```
rossabaker: I wonder, is `ParseResult` really right when we're not parsing from a String?  Maybe we could call this `fromLong`, and then the old `apply` signature could remain for a deprecation cycle.

I think I could argue that `apply` should be reserved for constructors that are total.

cquiroz:
Thanks, updated as suggested

cquiroz:
Thanks for al the feeback to make this one a lot better. I'm going to merge it now

core/src/main/scala/org/http4s/headers/Content-Length.scala
```diff
@@ -9,21 +9,27 @@ import org.http4s.util.Writer
...
+sealed abstract case class `Content-Length`(length: Long) extends Header.Parsed {
   override def key: `Content-Length`.type = `Content-Length`
   override def renderValue(writer: Writer): writer.type = writer.append(length)
+  def modify(f: Long => Long): Option[`Content-Length`] = `Content-Length`.fromLong(f(length)).right.toOption
 }
 
 object `Content-Length` extends HeaderKey.Internal[`Content-Length`] with HeaderKey.Singleton {
+  private class ContentLengthImpl(length: Long) extends `Content-Length`(length)
+
+  val zero: `Content-Length` = new ContentLengthImpl(0) {}
```
bryce-anderson: FYI, the `{}` made a new anonymous class that extends `ContentLengthImpl`.

cquiroz:
Oops, that slipped through, I'll fix it asap

PR #1246, opened 2017-07-07
Merge Conflict Made It Upstream

ChristopherDavenport:
Removing My Egregious Error.


ChristopherDavenport:
I accidentally pushed to upstream with this branch, which is why there is the failure on push. Only pay attention to PR which comes from my Repo where it was meant to be. 

Interesting that it somehow triggered the travis-ci push build.

ChristopherDavenport:
We'll see how it goes. At this point it can't get much more broken.

PR #1248, opened 2017-07-11
Add missing StringBuilding trait

rossabaker:
This fixes the breakage from merging #1242.  It has to live in a separate source directory from the rest of parboiled2, which is why it wasn't included in our initial internalization of parboiled2.

### DIFF
--- a/core/src/main/scala/org/http4s/internal/parboiled2/StringBuilding.scala
+++ b/core/src/main/scala/org/http4s/internal/parboiled2/StringBuilding.scala
@@ -1,0 +1,72 @@
+/*
+ * Copyright (C) 2009-2013 Mathias Doenitz, Alexander Myltsev
+ *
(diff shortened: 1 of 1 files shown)


rossabaker:
It's a tradeoff, but here's my defense of the current approach:

- parboiled2 is a slow-moving dependency.  It was pretty much internalize-it-and-forget-it.  The reason that this comes up now is because this is one of three types that depended on a macro and had to be in a separate compilation pass, and we got to punt on that until now.
- parboiled2 is an implementation detail.  Using parboiled2 makes nice, fast parsers, but doesn't make http4s play nicer with anything else.  So the dependency is much more liability than interoperability, unlike cats or fs2.
- the shapeless dependency we get from parboiled2 has caused massive pain on large projects.  I like shapeless, but finding a stack of libraries that all use a compatible shapeless has cost many hours of time at work and impedes upgrades.  We remove ourselves from those constraints by this.

PR #1250, opened 2017-07-12
Let logger middleware pass through body

rossabaker:


server/src/test/scala/org/http4s/server/middleware/LoggerSpec.scala
```diff
@@ -27,10 +37,10 @@ class LoggerSpec extends Http4sSpec {
     }
 
     "not effect a Post" in {
-      val req = Request(uri = uri("/post"), method = POST).withBody(urlForm)
-      val res = req.flatMap(responseLoggerService.orNotFound)
+      val req = Request(uri = uri("/post"), method = POST).withBody(body)
```
rossabaker: I'm getting sleepy and not sure why this is a `Request` instead of a `Task[Request]` like it was with `urlForm`.

rossabaker:
What are we going to do with large bodies?  The existing implementation buffers indefinitely.  An alternate approach could be to flush bodies out at a certain size.  In that case, the request no longer fits in a single log line, and we get into the territory of wanting a request ID to correlate log messages.

ChristopherDavenport:
I agree, however that may be a case of premature optimization. 

I am of the belief that, while a streaming implementation would be better, an implementation is superior to none.

PR #1251, opened 2017-07-12
Test that toHttpService returns a live response body

rossabaker:
This is what they call foreshadowing.  It's broken in master, but let's put up the safety net here.


rossabaker:
I'll do that one, with fix, as a PR when I get back to a computer.

PR #1252, opened 2017-07-12
Return live response body from Client.toHttpService

rossabaker:
This is tested in #1251, but broken in master.  This contains the fix.  We need to attach `dispose` to the finalizer of the body, not dispose and then return the body.

### DIFF
--- a/client/src/main/scala/org/http4s/client/Client.scala
+++ b/client/src/main/scala/org/http4s/client/Client.scala
@@ -77,12 +77,12 @@ final case class Client(open: Service[Request, DisposableResponse], shutdown: Ta
     * signatures guarantee disposal of the HTTP connection.
     */
   def toHttpService: HttpService = {
-    open.flatMapF{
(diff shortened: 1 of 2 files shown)


client/src/main/scala/org/http4s/client/Client.scala
```diff
@@ -77,12 +77,12 @@ final case class Client(open: Service[Request, DisposableResponse], shutdown: Ta
...
   def toHttpService: HttpService = {
-    open.flatMapF{
-      case DisposableResponse(response, dispose) => dispose.flatMap(_ => Task.now(response))
+    open.map { case DisposableResponse(response, dispose) =>
+      response.copy(body = response.body.onFinalize(dispose))
     }
   }
 
-
+n
```
ChristopherDavenport: This is gonna break it...

ChristopherDavenport: This is gonna break it.

rossabaker: Gack, I saw it and I got distracted and I pushed it.

PR #1253, opened 2017-07-12
Re-enables last event id spec

aspic:
I dove into the tests to see if I could help out on the fs2-cats port. Discovered this disabled code and figured I could "fix" it. Kind of refs. https://github.com/http4s/http4s/issues/866


rossabaker:
Thank you!

PR #1254, opened 2017-07-13
Bring Age header up to standard

cquiroz:
After the changes on `Content-Length` on #1244, I'm updating the `Age` header in the same way, removing the partial `apply` constructor and the anonymous class

This is an API break but given that `Age` was add very recently I'd expect the impact to be minimal


cquiroz:
The criteria used were:

* No anonymous classes, instead we make a private class
* No partial `apply` constructors. Instead we have e.g. `fromLong` which is partial
* No `require` calls on constructors though I guess the general rule is no exceptions in constructors

It should be added to some kind of coding standards, do we have one?

rossabaker:
We don't have a coding standards guide, as they've been emerging ad hoc on these various related tickets.  But it would be good to lay this out.

PR #1255, opened 2017-07-14
Coding standards

rossabaker:
We've been talking a lot about making the model more consistent.  Here's a start on explaining it.

### DIFF
--- a/docs/src/hugo/content/standards.md
+++ b/docs/src/hugo/content/standards.md
@@ -37,9 +37,6 @@ Many parts of the HTTP spec require case-insensitive semantics. Use
 `org.http4s.util.CaseInsensitiveString` to represent these. This is important to
 get correct matching semantics when using case class extractors.
 
-## Method names
-
-
 ## Case classes
 
 ### `apply`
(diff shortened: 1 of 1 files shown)


docs/src/hugo/content/standards.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Types
+
+### Effects
+
+Prefer a parameterized effect type over specializing on a task.
+
+```scala
+// Good
+def apply[F[_]](service: HttpService[F])(implicit F: Monad[F]): HttpService[F]
```
ChristopherDavenport: Doesnt this apply to 0.18 or after effect is more presently, we are hard-coded to `Task` .

aeons:
Would formatting fit here as well? It's kinda all over the place currently.

[scalafmt](http://scalafmt.org) with a sane config might be interesting. We use that all the time.

docs/src/hugo/content/standards.md
rossabaker: True, this section is forward looking.  I'll add something about that.

rossabaker:
I would like to start using scalafmt, but I'm scared to with the branches in flux.  I think we should strongly consider introducing scalafmt when the fs2-0.10 is the main line of development.

rossabaker:
`parse` or `fromString`?  We're all over the place here, too.

aeons:
`fromString` is more in line with the other methods imo.

PR #1257, opened 2017-07-14
FS2-0.10: Merge Logger, Content-Length and other changes from master

aeons:


ChristopherDavenport:
That is what I was referring to. I agree entirely, just want @aeons to be aware.

aeons:
Yeah, I want to keep up with the changes, there's still a bit to do every time, though.

PR #1259, opened 2017-07-14
FS2 0.10: Make *Builder.apply work with custom arguments

aeons:
This was broken with the addition of implicit arguments to the builders with default arguments.

Should be merged after #1258


aeons:
This doesn't work at all, I guess I tested a different branch before pushing.

aeons:
So, my motivation for this was that with parameterization, the following won't work in the current incarnation, due to the implicit type class:
```scala
BlazeBuilder[IO](executionContext = customEC).serve
```
(due to the whole reason this stuff was changed to begin with; one of the default arguments has a type parameter of `F`)

It would work as
```scala
BlazeBuilder[IO].withExecutionContext(customEC).server
```

It is a change to the user facing API, but having purely a builder syntax seems reasonable to me.

PR #1260, opened 2017-07-14
Update SpecialTest for False Positives

ChristopherDavenport:
Needs the exit code to be cached on exit to avoid false positives.


ChristopherDavenport:
I'm going to opt to merge this to make sure we have a working implementation as I'm afraid `/usr/bin/env bash -e` might break something in one of the publication scripts we might not expect. The same dilemma will face us when we want to merge those,  so I can be persuaded to instead add that here. 

I just believe that this will get all branches testing again properly which would be a nice change of pace. So I'm being overly cautious worrying of breaking something again.

rossabaker:
:+1:

Issue #920, opened 2017-02-13
Fix Pending test in blaze-client for fs2-cats


ChristopherDavenport:
Test currently fails as it receives a length of 0 and elevates protocol to Http1.1 which causes the Host header to be appended.

[Pending Test](https://github.com/http4s/http4s/blob/cats/blaze-client/src/test/scala/org/http4s/client/blaze/Http1ClientStageSpec.scala#L225)

PR #1261, opened 2017-07-14
Http1.0 if body is non empty, but content-length header not provided

aspic:
I'm not well versed in the internals of the HTTP spec (or Http4s implementation of it for that sake), but this seemed like one logical fix (based on the history of the conditional). Please correct me if I'm wrong :-)

Refs. https://github.com/http4s/http4s/issues/920


rossabaker:
This solves the problem in the common case, but reference equality checks make me nervous.  Setting the body to some other empty stream should be legal in HTTP/1.0, but we'd still force the upgrade here.  If the body is effectful, there isn't much we can do about it.

I think about the only way we can reduce false positives beyond this is if and when we figure out a port of `isHalt` on an fs2 stream.  I couldn't do it on fs2-0.9, but the new algebra may permit that.

aspic:
Yeah, good point :-)

PR #1262, opened 2017-07-15
Remove the 'name: String' param from AttributeKey

bryce-anderson:
The name parameter gives the false hope that the key
is based on something other than referential equality.
Right now, it's only purpose is to override the
`toString` method, and even that is dangerous because
`toString` doesn't allude that it's an `AttributeKey`.

### DIFF
--- a/core/src/main/scala/org/http4s/AttributeMap.scala
+++ b/core/src/main/scala/org/http4s/AttributeMap.scala
@@ -19,6 +19,9 @@ object AttributeKey {
 
   /** Construct a new [[AttributeKey]] */
   def apply[T]: AttributeKey[T] = new AttributeKey()
+
+  @deprecated("Removed because `name` suggests equality between keys with the same name", "0.17")
+  def apply[T](name: String): AttributeKey[T] = apply
 }
 
 /** An immutable map where an [[AttributeKey]]  for a fixed type `T` can only be associated with values of type `T`.
--- a/core/src/main/scala/org/http4s/AttributeMap.scala
+++ b/core/src/main/scala/org/http4s/AttributeMap.scala
@@ -11,22 +11,14 @@ package org.http4s
 /** A key in an [[AttributeMap]] that constrains its associated value to be of type `T`.
   * The key is uniquely defined by its reference: there are no duplicate keys, even
   * those with the same name and type. */
-final class AttributeKey[T] private (val name: String) {
+final class AttributeKey[T] private {
   def apply(value: T): AttributeEntry[T] = AttributeEntry(this, value)
-
-  override def toString: String = name
 }
 
 object AttributeKey {
 
-  /** Construct an [[AttributeKey]] */
-  def apply[T](name: String): AttributeKey[T] = new AttributeKey(name)
-
-  /**
-   * Encourage greater consistency in internal keys by imposing a universal prefix.
-   */
-  private[http4s] def http4s[T](name: String): AttributeKey[T] =
-    apply(s"org.http4s.$name")
+  /** Construct a new [[AttributeKey]] */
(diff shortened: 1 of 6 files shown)


core/src/main/scala/org/http4s/AttributeMap.scala
```diff
@@ -11,22 +11,14 @@ package org.http4s
...
-  /** Construct an [[AttributeKey]] */
-  def apply[T](name: String): AttributeKey[T] = new AttributeKey(name)
-
-  /**
-   * Encourage greater consistency in internal keys by imposing a universal prefix.
-   */
-  private[http4s] def http4s[T](name: String): AttributeKey[T] =
-    apply(s"org.http4s.$name")
+  /** Construct a new [[AttributeKey]] */
+  def apply[T]: AttributeKey[T] = new AttributeKey()
```
rossabaker: Would it make sense to leave a deprecated version of the `apply(String)` version to explain why it's ignored and where it went for people who are migrating?

core/src/main/scala/org/http4s/AttributeMap.scala
```diff
@@ -85,6 +77,6 @@ object AttributeMap
 // type inference required less generality
 /** A map entry where `key` is constrained to only be associated with a fixed value of type `T`. */
 final case class AttributeEntry[T](key: AttributeKey[T], value: T) {
-  override def toString: String = key.name + ": " + value
+  override def toString: String = s"$key: $value"
```
rossabaker: Hmm.  `key` is no longer going to have a useful `toString`.

Does anybody know offhand how sbt's `project` macro works?  It gets the name of the value to which it's assigned.  If it's stupid simple, that might improve debugging.  If it's non-trivial, never mind.

aeons:
While we're looking at these, does any of you have suggestion on how to use these in the parameterized version?

We can't very well create and save a ` AttributeKey[F[Vector[PushResponse]]]` if we don't know the concrete type of `F`.

rossabaker:
We know the `F` on the `Response` to which it's attached.  Can we hide the key so the only way to get and set that attribute is with the `Response` it's attached to, thus making it safe to cast?

core/src/main/scala/org/http4s/AttributeMap.scala
bryce-anderson: I don't know: how the toString will provide the name and its hashcode, which is arguably better than it was before. I agree though: it would be nice if we could get the name from the value it gets assigned to.
```scala
scala> AttributeKey[Int]
res0: org.http4s.AttributeKey[Int] = org.http4s.AttributeKey@3c7d90c5
```

bryce-anderson:
Why must the type of `F` of any type inserted into the `AttributeMap` be the same `F` associated with the `Response`?

core/src/main/scala/org/http4s/AttributeMap.scala
rossabaker: Yeah, I think this is a step forward, and if debugging gets annoying or someone comes along and says it's easy, we can do the Stupid Macro Trick that sbt's `project` uses.

rossabaker:
In the case of a push response, I would always expect it to be the same `F`.  I would be interested to hear a use case for an attribute that is neither in the effect type of its containing message nor hardcoded.  Perhaps a transformer of `F`, but no example springs to mind.

aeons:
There are currently three places where this is an issue:
- `Message.Keys.TrailerHeaders`, this can be moved into `Message` (has type `AttributeKey[F[Headers]]`
- `pushResponsesKey`, not sure how to handle this, since at most we have access to a `F[Vector[PushResponse[F]]]`
- `websocketKey`, not sure here either, since we don't really have anywhere to put it

PR #1266, opened 2017-07-19
FS2 0.10: Refactor StreamApp

aeons:
This is a suggestion for changes to `StreamApp`.

Currently it has quite a few `unsafeRun*`s. This changes the API quite a bit to avoid those (since getting a value out of an `F[_]: Effect` is nigh on impossible, and a horrible thing to do according to the cats-effect gitter).

### DIFF
--- a/core/src/main/scala/org/http4s/util/StreamApp.scala
+++ b/core/src/main/scala/org/http4s/util/StreamApp.scala
@@ -2,64 +2,46 @@ package org.http4s.util
 
 import cats.effect._
 import cats.effect.implicits._
+import fs2.Stream._
 import fs2._
 import fs2.async._
-import fs2.async.mutable.Signal
 import org.log4s.getLogger
-
-import scala.concurrent.SyncVar
 
 abstract class StreamApp[F[_]](implicit F: Effect[F]) {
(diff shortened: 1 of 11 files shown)


aeons:
The Scala compiler will mark anything following a `Stream[F, Nothing]` in a for-comprehension as dead code, so the code in `mainStream` has a couple of casts to get around that.

core/src/main/scala/org/http4s/util/StreamApp.scala
```diff
@@ -2,64 +2,46 @@ package org.http4s.util
...
+      _ <- eval(F.delay(sys.addShutdownHook {
+        unsafeRunAsync(shutdownRequested.set(true))(_ => IO.unit)
+        halted.discrete.takeWhile(_ == false).run.runAsync(_ => IO.unit).unsafeRunSync()
+      }))
+      res <- stream(args.toList, shutdownRequested.set(true))
+        .interruptWhen(shutdownRequested)
+        .onFinalize(halted.set(true))
+        .as(())
+    } yield res
+    s.drain
```
aeons: The Scala compiler will mark anything following a Stream[F, Nothing] in a for-comprehension as dead code, hence the `.as(())` and `.drain`.

Issue #1264, opened 2017-07-19
POST timeouts only using AHC client backend


ktonga:
It is kind of an special case, the request is a POST without body, `Content-Length:0` must be specified, it also has not response body, useful data is provided in the `Location` response header.

http4s version is `0.15.7a`

Here is the request performed with HTTPie:

```
http --verbose  POST 'https://va.convep.liveperson.net/api/account/1234/chat/request.json?v=1&appKey=abcd' Content-Length:0
```

Which results in following response

```
POST /api/account/1234/chat/request.json?v=1&appKey=abcd HTTP/1.1
Accept: */*
Accept-Encoding: gzip, deflate
Connection: keep-alive
Content-Length: 0
Host: va.convep.liveperson.net
User-Agent: HTTPie/0.9.9



HTTP/1.1 201 Created
Content-Length: 0
Content-Type: application/json
Date: Tue, 18 Jul 2017 01:15:22 GMT
Location: https://va.convep.liveperson.net/api/account/1234/chat/H516121171561454067-1e6b301c702c4099ad5f98e6e6ef8d92K8391827
P3P: CP="NON BUS INT NAV COM ADM CON CUR IVA IVD OTP PSA PSD TEL SAM"
Server: WS
```

Here is a code snippet to reproduce the error using http4s

```scala
object Blah extends scala.App {

  import org.http4s.Uri.uri

  val http = AsyncHttpClient()

  val uri_ = uri("https://va.convep.liveperson.net/api/account/1234/chat/request.json")
  val req = Request(POST, uri_ +? ("v", "1") +? ("appKey", "abcd"))
    .putHeaders(`Content-Length`(0))
  println(req)
  val locUri = http.fetch[Uri](req) {
    case Successful(resp) =>
      println(resp)
      resp.headers
        .collectFirst { case Location(loc) => loc.parsed.uri }
        .fold[Task[Uri]](Task.fail(Error.messageError("No URI provided for bridge chat")))(Task.now)
    case failedResponse =>
      println(failedResponse)
      Task.fail(Error.messageError(s"Unexpected status creating bridge chat: ${failedResponse.status}"))
  }

  println(locUri.unsafePerformSync)
  http.shutdownNow()
}
```

What results in a timeout with the following exception:

```
Request(method=POST, uri=https://va.convep.liveperson.net/api/account/1234/chat/request.json?v=1&appKey=abcd, headers=Headers(Content-Length: 0)
[...]

PR #1268, opened 2017-07-19
Pass content length to InputStreamBodyGenerator in AHC

rossabaker:
Without a content-length, `InputStreamBodyGenerator` assumes chunked transfer encoding.  We pass the content length (or -1 if we can establish chunked transfer encoding).

I did not have a failing test before this change.  I can only reproduce the problem as described against #1264.  This fixes the issue described in #1264.

As stated in #1264, a better solution would be a non-blocking implementation of the body generator based on reactive streams.  Since scalaz-stream is dead and a proper implementation is non-trivial, I propose that this is Good Enough™.

/cc @ktonga

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ env:
   - SCALAZ_VERSION=7.2.12
   - SCALAZ_VERSION=7.1.13
 before_script:
-- mkdir $HOME/bin
+- mkdir -p $HOME/bin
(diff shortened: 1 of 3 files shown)


ktonga:
Cool. And thanks for the awesome turnaround on the bugfix time!

rossabaker:
I'll release this tonight as 0.15.15 if there are no objections.

PR #1272, opened 2017-07-20
HSTS Support

cquiroz:
This PR adds support for HSTS in the form of the `Strict-Transport-Security` header and a middleware to activate it

As a reference 
https://www.owasp.org/index.php/HTTP_Strict_Transport_Security_Cheat_Sheet

I was tempted to add it to Blaze as it probably makes more sense as a server feature than a `Middleware` but got lost on the details. Let me know if you think that's an avenue to pursuit

### DIFF
--- a/core/src/main/scala/org/http4s/headers/Strict-Transport-Security.scala
+++ b/core/src/main/scala/org/http4s/headers/Strict-Transport-Security.scala
@@ -9,7 +9,8 @@ import org.http4s.util.Writer
  * Defined by http://tools.ietf.org/html/rfc6797
  */
 object `Strict-Transport-Security` extends HeaderKey.Internal[`Strict-Transport-Security`] with HeaderKey.Singleton {
-  private class StrictTransportSecurityImpl(maxAge: Long, includeSubDomains: Boolean, preload: Boolean)
+  private[headers] class StrictTransportSecurityImpl(maxAge: Long, includeSubDomains: Boolean, preload: Boolean)
+    extends `Strict-Transport-Security`(maxAge, includeSubDomains, preload)
 
   def fromLong(maxAge: Long, includeSubDomains: Boolean = true, preload: Boolean = false): ParseResult[`Strict-Transport-Security`] =
     if (maxAge >= 0) {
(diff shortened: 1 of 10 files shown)


core/src/main/scala/org/http4s/headers/Strict-Transport-Security.scala
```diff
@@ -1,7 +1,25 @@
...
 /**
  * Defined by http://tools.ietf.org/html/rfc6797
  */
-object `Strict-Transport-Security` extends HeaderKey.Default
+object `Strict-Transport-Security` extends HeaderKey.Internal[`Strict-Transport-Security`] with HeaderKey.Singleton {
+  def parse(s: String): ParseResult[`Strict-Transport-Security`] =
+    HttpHeaderParser.STRICT_TRANSPORT_SECURITY(s)
+}
+
+case class `Strict-Transport-Security`(maxAge: FiniteDuration, includeSubDomains: Boolean = true, preload: Boolean = false) extends Header.Parsed {
```
rossabaker: This is another case where passing a `Duration` is more intuitive and useful, but fails a round trip if something tighter than seconds is passed.  It's much like the `Age` header in that regard.

ChristopherDavenport: Yes, very unfortunate that `Duration` misbehaves. Wonder if there is a clean proposal we could come up with for scalac?

rossabaker: I don't think it's a flaw in duration.  It's a flaw in trying to use the `Duration` type, which is capable of nanosecond resolution, to represent seconds.  It's like using a double for something that is always rendered as an int.

cquiroz:
In this case maxAge durations of seconds or much longer than a year probably make little sense so duration is a good fit. However, I appreciate the need to be consistent so let's change it

I'll also try to add an entry on the documentation along `CORS` and `GZip` middleware

rossabaker:
My concern is not the max value, but the fact that duration silently loses precision.  If I were to design a new type for it, it would look like `FiniteDuration`, but with an underlying representation of seconds instead of nanoseconds.

cquiroz:
Changed it to `Long` as the other similar headers

server/src/main/scala/org/http4s/server/middleware/HSTS.scala
```diff
@@ -0,0 +1,27 @@
...
+import scala.concurrent.duration._
+
+/** [[Middleware]] to add HTTP Strict Transport Security (HSTS) support adding
+  * the Strict Transport Security headers
+  */
+object HSTS {
+
+  def apply(service: HttpService, maxAge: FiniteDuration = 365.days, includeSubDomains: Boolean = true, preload: Boolean = false): HttpService = Service.lift { req =>
+
+    val header = `Strict-Transport-Security`.unsafeFromDuration(maxAge, includeSubDomains, preload)
```
cquiroz: This isn't very nice. Changing the param to `Long` only changes this slightly as we'd still be using `unsafeFromLong`. Any suggestions? Perhaps making `HSTS.apply` return a partial too?

ChristopherDavenport:
Going off the grid for several days(only be on mobile), so I don't want to be the blocker. Not sure what the best resolution to the elevation of parse difficulties to Middleware out of Headers. Seems perhaps apply could be removed/total and instead be unsafeFromDuration applied to the middleware level.

server/src/main/scala/org/http4s/server/middleware/HSTS.scala
cquiroz: I'm thinking to define several `apply` methods to pass directly the header and one with a default age of 1y. Both cases would make `apply` total and then convert the current `apply` to `unsafeFromDuration`

WDYT?

rossabaker: Is `delta-seconds` common enough in the spec that we just need a duration type with a granularity of seconds?  I don't like introducing another utility type, but this generic use of Long isn't ideal either.

I think I like your proposed solution here regardless.

cquiroz: I think anything less than days is probably not used in this case. Basically you want to tell the browser not to use http for the next n days/months/years.

cquiroz:
Updated the HSTS middleware as discussed and added documentation

PR #1273, opened 2017-07-21
Add withFragment to remove redundant copy calls

ChristopherDavenport:
Follows current convention. However would we prefer an explicit `Fragment` then wrapped in `Some`?


ChristopherDavenport:
Decided that I preferred the `Fragment/String` directly and it being Total via using `Option` rather than `Some` to handle null Strings.

PR #1280, opened 2017-07-26
Instances for QueryParamDecoder/Encoder

tpolecat:
This adds some obvious instances for `QueryParam` things and deprecates some now-redundant factory methods.

### DIFF
--- a/core/src/main/scala/org/http4s/QueryParam.scala
+++ b/core/src/main/scala/org/http4s/QueryParam.scala
@@ -2,6 +2,7 @@ package org.http4s
 
 import cats._
 import cats.data._
+import cats.functor.Contravariant
 import cats.implicits._
 
 final case class QueryParameterKey(value: String) extends AnyVal
@@ -41,8 +42,16 @@ object QueryParamKeyLike {
  * Type class defining how to encode a `T` as a [[QueryParameterValue]]s
  * @see QueryParamCodecLaws
  */
-trait QueryParamEncoder[T] {
+trait QueryParamEncoder[T] { outer =>
   def encode(value: T): QueryParameterValue
+
+  /** QueryParamEncoder is a contravariant functor. */
+  def contramap[U](f: U => T): QueryParamEncoder[U] =
+    new QueryParamEncoder[U] {
+      override def encode(value: U) =
+        outer.encode(f(value))
+    }
+
 }
 
 object QueryParamEncoder {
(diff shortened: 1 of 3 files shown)


core/src/main/scala/org/http4s/QueryParam.scala
```diff
@@ -41,43 +42,82 @@ object QueryParamKeyLike {
...
-      QueryParamEncoder[U].encode(f(value))
-  }
+  /** QueryParamEncoder is a contravariant functor. */
+  implicit val ContravariantQueryParamEncoder: Contravariant[QueryParamEncoder] =
+    new Contravariant[QueryParamEncoder] {
+      override def contramap[A, B](fa: QueryParamEncoder[A])(f: B => A) =
+        fa.contramap(f)
+    }
+
+  @deprecated("Use QueryParamEncoder[U].contramap(f)", "0.17")
```
rossabaker: If we target this at release-0.16 branch, it'll be more useful for you.  And then update these deprecations accordingly.

core/src/main/scala/org/http4s/QueryParam.scala
```diff
@@ -91,33 +131,56 @@ object QueryParamDecoder {
...
+    qpd.map(f)
+
+  /** A decoder that always fails. */
+  implicit lazy val nothingQueryParamDecoder: QueryParamDecoder[Nothing] =
+    new QueryParamDecoder[Nothing] {
+      override def decode(value: QueryParameterValue) =
+        Validated.invalidNel(ParseFailure("Failed.", "Nothing decoder always fails."))
+    }
+
+  implicit lazy val booleanQueryParamDecoder: QueryParamDecoder[Boolean] =
```
rossabaker: What are we getting out of laziness on these?  laziness and some CI lockups have given me a twitch.

tpolecat: I ran into an NPE here and figured making them lazy is safer than trying to re-order.

tpolecat:
I don't have strong feelings about 0.16 vs 0.17 … I have these instances in a package object locally so I'm not in a hurry. If you prefer to see these changes in 0.16 it's not a problem to retarget. The differences are mechanical.

rossabaker:
I might like them for as long as I'm on 0.16.  0.17 will, of course, have to look like this.  Good thing it's in git.

tpolecat:
Ok I'll rejigger for 0.16 and then forward-port it to 0.17

Issue #1281, opened 2017-07-26
make org.http4s.util.threads.DefaultPool lazy or a method


NicolasRouquette:
In 0.17.0-M3, this variable is defined like this:

```scala
  private[http4s] def newDaemonPool(name: String, min: Int = 4, cpuFactor: Double = 3.0, timeout: Boolean = false) = {
    val cpus = Runtime.getRuntime.availableProcessors
    val exec = new ThreadPoolExecutor(
      math.max(min, cpus), math.max(min, (cpus * cpuFactor).ceil.toInt),
      10L, TimeUnit.SECONDS,
      new LinkedBlockingQueue[Runnable],
      threadFactory(i => s"${name}-$i", daemon = true))
    exec.allowCoreThreadTimeOut(timeout)
    exec
  }

  /** Servers and clients shouldn't run on a daemon.  This will go away. */
  private[http4s] val DefaultPool =
    newDaemonPool("http4s-pool")
```

In http4s, it is used as a default method parameter value in several places.
Since it is not lazy, it is initialized regardless of whether it is actually used or not.
When it is not used, an application nonetheless has an extra thread pool configured with the max number of cpus available that does nothing logically useful.

I suggest refactoring this to make the variable lazy so that the default thread pool would be created only if necessary.

rossabaker:
Fixed by #1282

PR #1282, opened 2017-07-26
Fix for #1281

NicolasRouquette:
In `org.http4s.util.threads`,`DefaultExecutionContext` depends on `DefaultPool` so both must be made lazy.

In `org.http4s.servlet.syntax`, changed `ServletContextOps.mountService` to include an `ExecutionContext` which defaults to the lazy `DefaultExecutionContext`.


rossabaker:
Now that everything is based on ExecutionContext, should we even provide our own default, or should we just use the Scala global by default? 🤔

NicolasRouquette:
I think it would be better to reuse the Scala `ExecutionContext.global` instead of creating a separate one. This issue is caused by the fact that `org.http4s.util.threads.DefaultExecutionContext` is an eager variable; which causes that context to be created even if it is not used. Note that Scala's `ExecutionContext.global` is a lazy def; so there is no penalty in using it as the default value of a function parameter.

rossabaker:
I think I agree with the default execution context, but I'll argue the other side for completeness:
* It's a fork-join pool, and this is not fork-join optimized work.  Whether we can measure a difference in throughput here vs. a fixed thread pool, I'm not particularly sure.
* It's a daemon pool, and I'd like to discourage people from using daemon pools for serving requests.  A daemon pool abruptly stops instead of blocking shutdown until the executor is cleanly shutdown.  I don't think this is good for production.  This problem already exists in the current implementation, and I don't know how to solve it without starting a "special" pool that shuts down with the server/client.  We've done this before, and it was ugly.

NicolasRouquette:
Here's what I ended up doing using this PR to get precise control of how pools, threads, etc... used:

```scala
  def apply(name: String, nbWorkers: Int, idleTimeout: Duration, maxTotalConnections: Int)
  : (Client, fs2.Strategy, Scheduler) = {
    val execService = Executors.newScheduledThreadPool(nbWorkers, Strategy.daemonThreadFactory(name))
    implicit val S = fs2.Strategy.fromExecutor(execService)
    implicit val sched: Scheduler = Scheduler.fromScheduledExecutorService(execService)
    val prov: AsynchronousChannelProvider = AsynchronousChannelProvider.provider()
    val gr = prov.openAsynchronousChannelGroup(execService, nbWorkers)

    val config =
      BlazeClientConfig
        .insecure
        .copy(idleTimeout, executionContext = ExecutionContext.fromExecutor(execService), group = Option(gr))
    val httpClient: Client = PooledHttp1Client(nbWorkers, config).copy(shutdown = Task.delay(execService.shutdown()))
    (httpClient, S, sched)
  }
```

I confirmed with the debugger that there are no default pools of any other kind created; which is what I really needed. I found that there is a subtle "default" behavior in that if we do not provide an `AsynchronousChannelGroup`, then one is created using the default thread pool!

rossabaker:
I am leaning toward eliminating our default pool and moving toward defaulting to the global execution context.  Any strenuous objections?

ChristopherDavenport:
I think that is a better solution overall.

NicolasRouquette:
I've deleted the default pools & merged w/ master. 
Is this what you intended @rossabaker & @ChristopherDavenport?

ChristopherDavenport:
Slightly worried about that merged with master commit. Doing stuff like that opens the door to regressions.

NicolasRouquette:
Mea culpa. I should have merged upstream. 
Since I began this PR, there was another merge to master.

NicolasRouquette:
@ChristopherDavenport Is this what you mean? 
https://github.com/http4s/http4s.g8/pull/20

As far as this PR is concerned, is it ready to be closed?

PR #1283, opened 2017-07-27
Update roadmap

rossabaker:


cquiroz:
Will 0.16 continue after 1.0 is released?

ChristopherDavenport:
My understanding was that Ross is stuck with 0.16 for a while for work. But that it would be in Bugfix/Manual backport. Not primary development.

rossabaker:
Correct.  I'll bugfix it for myself indefinitely, and if someone wants to take over when I'm done with it, cool.  But new stuff would happen on 0.18 with cats-effect and friends.

PR #1285, opened 2017-07-28
Add status and successful headers to client

rossabaker:
These are convenience methods I've implemented elsewhere.


rossabaker:
For consistency, we might want a version that takes a Uri and a String, but it's a drag to have four copies of everything.  Perhaps it's time to deprecate those?

rossabaker:
Build failure due to #1284.

client/src/main/scala/org/http4s/client/Client.scala
```diff
@@ -180,6 +189,15 @@ final case class Client(open: Service[Request, DisposableResponse], shutdown: Ta
   def expect[A](req: Task[Request])(implicit d: EntityDecoder[A]): Task[A] =
     req.flatMap(expect(_)(d))
 
+  /** Submits a request and returns the response status */
+  def status(req: Task[Request]): Task[Status] =
```
ChristopherDavenport: Why are the overloads not located near the non-task equivalents?

rossabaker: I think they were sorted by argument type instead of by name, but that's silly.  I'll shuffle it around.

rossabaker:
Build is still going to fail until parser is fixed.

Added `streaming[Task[Request]]` and undeprecated `fetchAs[Task[Request]]`, which made sense while I was doing the grouping requested by @ChristopherDavenport.

PR #1286, opened 2017-07-28
Fix broken IP parsing

rossabaker:
* Old IPv4 parser allowed invalid octets
* IPv6 parser allowed invalid addresses, threw exceptions

Not sure why #1284 didn't get a PR build to uncover this, but I think this fixes release-0.16.x.

### DIFF
--- a/core/src/main/scala/org/http4s/parser/Rfc2616BasicRules.scala
+++ b/core/src/main/scala/org/http4s/parser/Rfc2616BasicRules.scala
@@ -17,8 +17,6 @@
  */
 package org.http4s.parser
 
-import scala.reflect.ClassTag
-import scalaz.Validation
 import org.http4s.{ParseFailure, ParseResult}
 import org.http4s.internal.parboiled2._
 
--- a/core/src/main/scala/org/http4s/parser/Rfc2616BasicRules.scala
+++ b/core/src/main/scala/org/http4s/parser/Rfc2616BasicRules.scala
@@ -21,7 +21,6 @@ import scala.reflect.ClassTag
 import scalaz.Validation
 import org.http4s.{ParseFailure, ParseResult}
 import org.http4s.internal.parboiled2._
-import org.http4s.internal.parboiled2.CharPredicate.{ HexDigit => HEXDIG }
 
 // direct implementation of http://www.w3.org/Protocols/rfc2616/rfc2616-sec2.html#sec2
 private[http4s] trait Rfc2616BasicRules extends Parser {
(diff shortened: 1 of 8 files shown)


core/src/main/scala/org/http4s/parser/Rfc2616BasicRules.scala
```diff
@@ -74,11 +77,63 @@ private[http4s] trait Rfc2616BasicRules extends Parser {
...
   def LASTCHAR: Rule1[Char] = rule { push(input.charAt(cursor - 1)) }
 
-  // we don't match scoped IPv6 addresses
-  def IPv6Address = rule { oneOrMore(Hex | anyOf(":.")) }
+  private val DIGIT = CharPredicate.Digit
+  private val DIGIT04 = CharPredicate('0' to '4')
+  private val DIGIT05 = CharPredicate('0' to '5')
+  private val DIGIT19 = CharPredicate.Digit19
+
+  def IPNumber = rule {
```
ChristopherDavenport: Breaking this down. Perhaps an easy addition of documentation.

```
2[0-4][0-9] : 200-249
25[0-5]     : 250-255
1[0-9][0-9] : 100-199
[1-9][0-9]  : 10-99
[0-9]       : 0-9
```

bryce-anderson:
How do these rules differ from those in `Rfc3986Parser` [IpV6Address](https://github.com/http4s/http4s/blob/master/core/src/main/scala/org/http4s/parser/Rfc3986Parser.scala#L67) and [IpV4Address](https://github.com/http4s/http4s/blob/master/core/src/main/scala/org/http4s/parser/Rfc3986Parser.scala#L83)? I think we had some duplication going on.

rossabaker:
Ugh.  I thought we had a nicer parser than the ones I updated.  I just overlooked it.  Yeah, there's no reason we can't consolidate on those and get rid of this copy'n'paste.

core/src/main/scala/org/http4s/parser/SimpleHeaders.scala
```diff
@@ -95,12 +96,15 @@ private[parser] trait SimpleHeaders {
...
 //  // https://bugzilla.mozilla.org/show_bug.cgi?id=464162 (FIXED in mozilla)
-  def HOST(value: String): ParseResult[Host] = new Http4sHeaderParser[Host](value) {
-    def entry = rule {
-      (Token | IPv6Reference) ~ OptWS ~
-        optional(":" ~ capture(oneOrMore(Digit)) ~> (_.toInt)) ~ EOL ~> (Host(_:String, _:Option[Int]))
-    }
-  }.parse
+  def HOST(value: String): ParseResult[Host] =
+    new Http4sHeaderParser[Host](value) with Rfc3986Parser {
+      def charset = StandardCharsets.UTF_8
```
rossabaker: This seems like a bad hardcoding, but, geez, is anybody sending these that aren't UTF-8?

core/src/main/scala/org/http4s/parser/Rfc2616BasicRules.scala
```diff
@@ -17,6 +17,8 @@
  */
 package org.http4s.parser
 
+import scala.reflect.ClassTag
```
bryce-anderson: What are these two new imports for? I suspect they are a mistake.

core/src/main/scala/org/http4s/parser/SimpleHeaders.scala
```diff
@@ -162,15 +166,16 @@ private[parser] trait SimpleHeaders {
...
-      oneOrMore((Ip ~> (Some(_))) |
-                (capture(IPv6Address) ~> { s: String => Some(InetAddress.getByName(s)) }) |
-                ("unknown" ~ push(None))).separatedBy(ListSep) ~
-        EOL ~> { xs: Seq[Option[InetAddress]] =>
-        `X-Forwarded-For`(xs.head, xs.tail: _*)
+  def X_FORWARDED_FOR(value: String): ParseResult[`X-Forwarded-For`] =
+    new Http4sHeaderParser[`X-Forwarded-For`](value) with IpParser {
+      def entry = rule {
+        oneOrMore(
+          (capture(IpV4Address) ~> { s: String => Some(InetAddress.getByName(s)) }) |
```
bryce-anderson: Could these two rules be combined into
```scala
(capture(IpV4Address | IpV6Address) ~> { s: String => Some(InetAddress.getByName(s)) })
```

core/src/main/scala/org/http4s/parser/Rfc2616BasicRules.scala
rossabaker: I, uh, don't recall adding them. 😖

PR #1287, opened 2017-07-29
Replace Instant with HttpDate

rossabaker:
* Granularity is seconds, so no implicit loss of precision when rendering.
* Constrained to the range of dates that HTTP supports


rossabaker:
Tagging @cquiroz because he's done the most work with dates and times, but as always, everyone's feedback encouraged.

rossabaker:
This is what I dislike about the status quo:

```scala
scala> val h1 = headers.Date(now)
h1: org.http4s.headers.Date = Date: Sat, 29 Jul 2017 03:37:37 GMT

scala> val h2 = headers.Date(now.plusMillis(1))
h2: org.http4s.headers.Date = Date: Sat, 29 Jul 2017 03:37:37 GMT

scala> h1.renderString == h2.renderString
res8: Boolean = true

scala> h1 == h2
res9: Boolean = false
```

If headers are equal on the wire, they should be equal in the model.

An alternative to creating a custom http4s type would be to truncate them on creation, but that's a surprise unexpressed by the type system.  It also doesn't solve the (probably unimportant) problem of rendering dates outside the four-digit year range.

cquiroz:
This seems solid. I'd be happy to use it. Is the timezone in `HttpDate` always `GMT`?

rossabaker:
Yes, it's always GMT.

core/src/main/scala/org/http4s/HttpDate.scala
```diff
@@ -0,0 +1,105 @@
...
+    unsafeFromInstant(Instant.now)
+
+  /** The `HttpDate` equal to `Thu, Jan 01 1970 00:00:00 GMT` */
+  val Epoch: HttpDate =
+    unsafeFromEpochSecond(0)
+
+  private val MinEpochSecond = -62135596800L
+
+  /** The earliest value reprsentable by RFC1123, `Mon, 01 Jan 0001 00:00:00 GMT`. */
+  val MinValue = HttpDate.unsafeFromEpochSecond(MinEpochSecond)
```
rossabaker: [RFC5322] offers a minimum of 1900, which avoids any unpleasant conversations about Gregorian vs. Julian calendars.

core/src/main/scala/org/http4s/HttpDate.scala
```diff
@@ -0,0 +1,105 @@
...
+  def fromString(s: String): ParseResult[HttpDate] =
+    AdditionalRules.httpDate(s)
+
+  /** Like `fromString`, but throws on invalid input */
+  def unsafeFromString(s: String): HttpDate =
+    fromString(s).fold(throw _, identity)
+
+  /** Constructs a date from the seconds since the [[Epoch]]. If out of range,
+   *  returns a ParseFailure. */
+  def fromEpochSecond(epochSecond: Long): ParseResult[HttpDate] =
```
rossabaker: Coding standard note: I'm not sure we're on the right track returning a `ParseResult` when we're constructing rather than parsing.  It gets the error conditions into a consistent type, but it calls into question the name.

core/src/main/scala/org/http4s/StaticFile.scala
```diff
@@ -98,9 +96,8 @@ object StaticFile {
         }
       }
 
-      val hs = `Last-Modified`(lastModified) ::
-               `Content-Length`(contentLength) ::
-               contentType.toList
+      var hs: List[Header] = `Content-Length`(contentLength) :: contentType.toList
```
rossabaker: I'm not proud.

rossabaker:
I guess proceeding with this sort of implies a `DeltaSeconds` to handle the duration side of this mess?

core/src/main/scala/org/http4s/HttpDate.scala
cquiroz: I agree that the name gets missleading. Couldn't we just remove the alias and just do  `Either`?

core/src/main/scala/org/http4s/HttpDate.scala
cquiroz: 👍

core/src/main/scala/org/http4s/StaticFile.scala
cquiroz: Could you just pattern match?

cquiroz:
I agree that it would make sense to have a `DeltaSeconds` type too

core/src/main/scala/org/http4s/HttpDate.scala
rossabaker: Yeah, but then it's an `Either[ParseFailure, A]`.  It still says "parse".

bryce-anderson:
I don't really like having yet another date type, but the equality thing is a bummer. I have the feeling most manipulations will end up going through the java types if they need to do any sort of manipulation. I'd personally go with truncation: thats pretty much what we're doing with the new `HttpDate` anyway.

rossabaker:
Truncation doesn't address the domain problem.  If we create a date out of range (which is either a bug or a troublemaker), how do we render it?  If we want to hold the line on totality, we'd need the `unsafeFromInstant` constructors on every header, instead of centralizing it on this type.

We could add some basic manipulation functions: `addSeconds`, `addDays`, etc., which would cover most use cases along with `parse` and `now`.  But any arithmetic ops we add raises overflow questions.

Principles can be inconvenient.

rossabaker:
Thought about this more on my run.  Still annoyed by all the solutions.

I think `Instant` is to `HttpDate` as `Double` is to `QValue`: the domain of the standard types is both wider and higher resolution than HTTP supports.  If we use an `Instant` here, why not a `Double` for `QValue`?  `QValue` doesn't itch because it's really easy to create one.  If `HttpDate` itches, is it because the type is a bad idea or because we've given it poor ergonomics?

ritschwumm:
i think HttpDate is the right thing to do here:

making invalid states unrepresentable and make it crystal clear you ar not dealing with any instant, but specifically a date in the HTTP context - with all the rules that apply there - is something that helps correctness in the long run which imho is much more important than short-time benefits like "not having to know another type".

core/src/main/scala/org/http4s/HttpDate.scala
```diff
@@ -0,0 +1,105 @@
...
+    unsafeFromEpochSecond(0)
+
+  private val MinEpochSecond = -62135596800L
+
+  /** The earliest value reprsentable by RFC1123, `Mon, 01 Jan 0001 00:00:00 GMT`. */
+  val MinValue = HttpDate.unsafeFromEpochSecond(MinEpochSecond)
+
+  private val MaxEpochSecond = 253402300799L
+
+  /** The latest value reprsentable by RFC1123, `Fri, 31 Dec 9999 23:59:59 GMT`. */
```
ChristopherDavenport: Nit: Let's add links to RFC's to allow developers working from our Docs to find the information they need without going through a separate source

core/src/main/scala/org/http4s/HttpDate.scala
```diff
@@ -0,0 +1,105 @@
...
+  val MaxValue = HttpDate.unsafeFromEpochSecond(MaxEpochSecond)
+
+  /** Parses a date according to RFC7321, Section 7.1.1.1 */
```
ChristopherDavenport: Same Nit as Above

core/src/main/scala/org/http4s/HttpDate.scala
ChristopherDavenport: `parseFromEpochSecond : ParseResult[HttpDate]` with `fromEpochSecond` convenience method? We can always deprecate the convenience method if we decide we don't like it and have a better naming convention for parsed types.

core/src/main/scala/org/http4s/StaticFile.scala
ChristopherDavenport:
```scala
val lastModifiedHeader : List[Header] = lm match {
     case None => List.empty[Header]
     case Some(httpDate) => List(`Last-Modified(httpDate)`)
}

val hs = `Content-Length`(contentLength) :: lastModifiedHeader :::  contentType.toList
```

Use property of List of ZeroOrMoreValues, or it can be handled within a single match. However should be nicer than effectful code.

core/src/main/scala/org/http4s/HttpDate.scala
rossabaker: The word `parse` still implies coming from a String to me.  `parseFromEpochSecond` would line up nicely with the return type, but not so nicely with the argument type.

core/src/main/scala/org/http4s/StaticFile.scala
rossabaker: This one got cleaned up in the merge.

rossabaker:
I think this is ready for a final review.  I'm punting on the parse naming, or we'll be delaying the releases even longer.

ChristopherDavenport:
Now that the `foreach` is gone I think I'm :+1: on green.

rossabaker:
Oops, I missed the RFC links.

core/src/main/scala/org/http4s/parser/HttpHeaderParser.scala
```diff
@@ -72,7 +73,16 @@ object HttpHeaderParser extends SimpleHeaders
...
     allParsers.get(header.name) match {
       case null => ParseResult.success(header) // if we don't have a rule for the header we leave it unparsed
-      case parser => parser(header.value)
+      case parser =>
+        try parser(header.value)
+        catch {
+          // We need a way to bail on invalid dates without throwing.  There should be a better way.
+          case e: ParseFailure =>
+            ParseResult.success(header)
+          case e: InvocationTargetException if e.getCause.isInstanceOf[ParseFailure] =>
```
rossabaker: I am really not proud of this section.  We are throwing an exception out of the date parser when it's well formed, but not a date.  And then because of the reflection in the parser, it can be wrapped.

I think we should ban throwing exceptions from parsers altogether, but I don't know how to express this in parboiled2.

ChristopherDavenport:
Looks good. Not sure about the Exceptions.

rossabaker:
I think the exception handling in parsers is something to be cleaned up later, but that debt is not new.

PR #1288, opened 2017-07-29
Add option to serve pre-gzipped resources

ZizhengTai:
Changes:

- Add `preferGzipped: Boolean` parameter to `StaticFile.fromResource` to enable serving pre-gzipped files
- Add `preferGzipped: Boolean` field to `ResourceService.Config`
- Change media type for extension ".gz" from `application/x-compressed` to `application/gzip`
- Add a bunch of related tests


core/src/main/scala/org/http4s/MediaType.scala
```diff
@@ -199,6 +199,7 @@ object MediaType extends Registry {
   val `application/excel`                                                         = app("excel", uncompressible, binary, "xl", "xla", "xlb", "xlc", "xld", "xlk", "xll", "xlm", "xls", "xlt", "xlv", "xlw")
   val `application/font-woff`                                                     = app("font-woff", uncompressible, binary, "woff")
   val `application/gnutar`                                                        = app("gnutar", uncompressible, binary, "tgz")
+  val `application/gzip`                                                          = app("gzip", uncompressible, binary, "gz")
```
rossabaker: We should do a resynchronizing with the IANA registry at some point, but :+1: to this first step.

core/src/main/scala/org/http4s/StaticFile.scala
```diff
@@ -126,5 +126,11 @@ object StaticFile {
     readAll[Task](f.toPath, DefaultBufferSize)
   }
 
+  private def nameToContentType(name: String): Option[`Content-Type`] =
+    name.lastIndexOf('.') match {
+      case -1 => None
+      case i => MediaType.forExtension(name.substring(i + 1)).map(`Content-Type`(_))
```
rossabaker: What happens if name is just `.`?  I can't remember whether substring out of bounds returns an empty string or throws.

ZizhengTai:
```scala
scala> ".".substring(1)
res0: String = ""
```
I think we are good.

rossabaker:
I'll merge this by cherry-picking back to 0.16.

rossabaker:
There are a lot of things touching StaticFile right now, and this one is a much harder cherry-pick than I anticipated.  I'm withdrawing my plan to backport this one to 0.16.x.  If another scalaz-streamer really wants it, PR welcome.

I'll merge this tomorrow barring objections.

Issue #972, opened 2017-03-01
Clean out hacks to preserve old docs with new sbt-ghpages


rossabaker:
sbt-ghpages-0.6.0 introduces [filters to protect part of your site](https://github.com/sbt/sbt-ghpages#protecting-existing-files).  I think this feature could get rid of many of the hacks in our project/Http4sGhPages.scala file.

bfritz:
Might be worth experimenting with [the gh-pages branch as a submodule](https://github.com/http4s/http4s/pull/1067#issuecomment-290293769) when working on this one.

rossabaker:
It's still janky, but we're now on a gh-pages version that helps a lot.

PR #1289, opened 2017-07-29
Omnibus dependency upgrade commit

rossabaker:
Holding back specs2 because we're on the latest version that works with both scalazes.  I'd like to bump that one in 0.17.


rossabaker:
I would also like to upgrade rig and get scalatest off our classpath, but that means tackling #972, and that's no way to spend a beautiful Saturday afternoon.

PR #1290, opened 2017-07-30
Backport #1213: update pathinfo when changing uri

rossabaker:


rossabaker:
A consequence of this I didn't realize: we now have overloads of `withBody`, one of which returns a `Message` and one that returns a `Task[Message]`.  That's ... not great.  If anyone has a clever name or way out of that, it would be great to hear and we could fix it here and bubble it back up to master.

rossabaker:
I added a `withBodyStream` as the low-level method that doesn't deal with payload headers.  The functionality is needed, but I am not fond of the name.  I do strongly think this should not be an overload, as the functionality is quite different.

core/src/main/scala/org/http4s/Headers.scala
```diff
@@ -95,6 +96,14 @@ final class Headers private (headers: List[Header])
       this.toList.equals(otherheaders.toList)
     case _ => false
   }
+
+  /** Removes the `Content-Length`, `Content-Range`, `Trailer`, and
```
ChristopherDavenport: This PR contains both two indentation and single indentation function annotation style. Should probably decide which style we prefer, or atleast stay consistent within a PR.

rossabaker: Let's default to the scaladoc style.  I anticipate we'll adopt scalafmt, and that would be one less default to change.

PR #1296, opened 2017-07-31
Fix corner case of bad hash code in CaseInsensitiveString

rossabaker:
This is a latent bug finally uncovered by property testing in [this build](https://travis-ci.org/http4s/http4s/jobs/259076280#L3246).


rossabaker:
Consistent with which other methods?

ChristopherDavenport:
The previous implementation. Where it was `toLowerCase` on the outer layer rather than the inner. I really don't think it matters. If we want to switch to a final upper case representation that is fine by me.

rossabaker:
Oh.  Yeah, I guess we could.  The order was arbitrary.

One thing I haven't done is loop through the entire Unicode space and make sure that this two-step fold doesn't fall short of some three-step monstrosity.

PR #1298, opened 2017-08-01
Add Content-Length header to empty response generators

ZizhengTai:
Added the `Content-Length: 0` header for responses with an obviously empty body.

I didn't touch `EmptyResponseGenerator`, because I don't know if it's intended to be used with status codes forbidden to have a body.


rossabaker:
Going to merge this and make a followup issue.

PR #1300, opened 2017-08-01
Stop Logger Middleware Eating Body

ChristopherDavenport:
Less Gobble Gobble, more Enjoying the Logging.

Continuation Reference of #1250

### DIFF
--- a/server/src/main/scala/org/http4s/server/middleware/Logger.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/Logger.scala
@@ -19,32 +19,38 @@ object Logger {
       )
     )
 
-  def getCharsetFromContentType(m: Message): (Boolean, Option[Charset]) = {
-    val init = m.charset
-    val binary = m.contentType.map(_.mediaType.binary).getOrElse(false)
-    (binary, init)
-  }
 
-  /**
-    * Log Messages As They Pass
-    */
-  def messageLogPipe[A <: Message](logHeaders: Boolean, logBody: Boolean)
-                                  (logger: SLogger): Pipe[Task, A, A] = stream => {
-    stream.flatMap { req =>
-      val (binary, charset) = Logger.getCharsetFromContentType(req)
-      val headers = if (logHeaders) req.headers.toList.mkString("Headers(", ", ", ")") else ""
-      val bodyT: Task[String] = {
-        if (logBody && !binary) {
-          req.bodyAsText(charset.getOrElse(Charset.`UTF-8`))
-        } else if (logBody) {
-          req.body.map(ByteVector.fromByte).map(_.toHex)
-        } else {
-          Stream.empty[Task, String]
-        }
-      }.runFoldMap(identity)
-      Stream.eval(bodyT.flatMap(body =>
-        Task.delay(logger.info(s"$headers + Body: $body"))
-      )) >> Stream.emit(req)
+  def logMessage[A <: Message](message: A)
+                              (logHeaders: Boolean, logBody: Boolean)
+                              (logger: SLogger)
+                              (implicit strategy: Strategy): Task[Unit] = {
+
+    val charset = message.charset
+    val binary = message.contentType.exists(_.mediaType.binary)
+
+    val headers = if (logHeaders) message.headers.toList.mkString("Headers(", ", ", ")") else ""
+
+    val bodyStream = if (logBody && !binary) {
+      message.bodyAsText(charset.getOrElse(Charset.`UTF-8`))
+    } else if (logBody) {
+      message.body.map(ByteVector.fromByte).map(_.toHex)
+    } else {
+      Stream.empty[Task, String]
+    }
+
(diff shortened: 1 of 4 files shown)


server/src/main/scala/org/http4s/server/middleware/Logger.scala
```diff
@@ -19,32 +19,38 @@ object Logger {
...
-      )) >> Stream.emit(req)
+  def logMessage[A <: Message](message: A)
+                              (logHeaders: Boolean, logBody: Boolean)
+                              (logger: SLogger)
+                              (implicit strategy: Strategy): Task[Unit] = {
+
+    val charset = message.charset
+    val binary = message.contentType.exists(_.mediaType.binary)
+
+    val headers = if (logHeaders) message.headers.toList.mkString("Headers(", ", ", ")") else ""
```
rossabaker: Given that the body is (typically) going to force us into multi-line logging anyway, I wonder if this is more readable if it looks like the request: delimit the headers with '\r\n', and then a final one before the unadorned body.

server/src/main/scala/org/http4s/server/middleware/Logger.scala
```diff
@@ -19,32 +19,38 @@ object Logger {
...
+
+    val bodyStream = if (logBody && !binary) {
+      message.bodyAsText(charset.getOrElse(Charset.`UTF-8`))
+    } else if (logBody) {
+      message.body.map(ByteVector.fromByte).map(_.toHex)
```
rossabaker: I feel like we should flag that we're doing binary output, but I don't have a good thought how.

server/src/main/scala/org/http4s/server/middleware/RequestLogger.scala
```diff
@@ -14,10 +14,25 @@ import scodec.bits._
...
-      .through(Logger.messageLogPipe[Request](logHeaders, logBody)(logger))
-      .evalMap[Task, Task, MaybeResponse]{req => service(req)}
-      .runFoldMap(identity)
+  def apply(logHeaders: Boolean, logBody: Boolean)(service: HttpService)(implicit strategy: Strategy): HttpService =
+
+    Service.lift{ req =>
+      if (!logBody) {
+        Logger.logMessage(req)(logHeaders, logBody)(logger)(strategy) >> service(req)
+      } else {
+        async.unboundedQueue[Task, Byte].flatMap { queue =>
```
rossabaker: Seems like this buffering could be DRYed up with the response: pass the `Message` and the `Logger`, and everything is the same?

server/src/main/scala/org/http4s/server/middleware/RequestLogger.scala
```diff
@@ -14,10 +14,25 @@ import scodec.bits._
...
+
+          val newBody = Stream.eval(queue.size.get)
+            .flatMap(size => queue.dequeue.take(size.toLong))
```
rossabaker: What are we getting out of buffering into a queue instead of buffering to a StringBuilder?  It's all going to memory anyway.  Wouldn't a StringBuilder be faster and save the folding that happens just before logging?

server/src/main/scala/org/http4s/server/middleware/RequestLogger.scala
ChristopherDavenport: So rather turn it back into a `Pipe[Task, A, A]` as opposed to a `A => Task[Unit]`

server/src/main/scala/org/http4s/server/middleware/Logger.scala
ChristopherDavenport: The only problem I saw is that this would be very verbose assuming you were only logging Headers, which is why I opted the direction I did. I was attempting to make the logging close to idiomatic scala output, but could be easily convinced to change.

server/src/main/scala/org/http4s/server/middleware/RequestLogger.scala
ChristopherDavenport: I honestly went with `Queue` because it seemed as though it was meant for it. However since technically the loading and unloading happens concurrently just on another thread I think we would be safe with a StringBuilder.

server/src/main/scala/org/http4s/server/middleware/Logger.scala
rossabaker: That's fair.  I'd like to be able to split these in a log analyzer, and sadly the only reliable delimiter would be a newline, as even unquoted spaces are allowed.

I think the right answer is eventually to make it customizable with a sensible default.  The most sensible default is what eludes me.

server/src/main/scala/org/http4s/server/middleware/RequestLogger.scala
rossabaker: I'm getting to the stupid part of the day, but I don't understand where the `A => Task[Unit]` is?

PR #1301, opened 2017-08-02
Instances for QueryParam codec

tpolecat:
This adds instances for `QueryParamEncoder` and `QueryParamDecoder`. Instances are constructed lazily to avoid initialization issues.


reactormonk:
Apparently `widen` isn't part of `scalaz.Functor` in scalaz 7.1

tpolecat:
Hell.

PR #1304, opened 2017-08-02
Tut CI Integration

ChristopherDavenport:
This should work, I have also integrated the `set -e` positioning. This fails due to the tut changes from #1301

### DIFF
--- a/bin/publishCentral.sh
+++ b/bin/publishCentral.sh
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+set -e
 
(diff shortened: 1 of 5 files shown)


ChristopherDavenport:
Actually passed due to merge of #1303

ChristopherDavenport:
`checkPublishable` cleanly exits the script early through `exit 0` effectively a `return` the other option is to return 0 or 1 and include in an if statement in the calling script. Which would be better functional programming.

PR #1305, opened 2017-08-02
Forward port QueryParam codec instances from 0.16 master

tpolecat:
- instances and tests from #1301 
- doc fix from #1303 
- remove method deprecated in #1301

### DIFF
--- a/core/src/main/scala/org/http4s/QueryParam.scala
+++ b/core/src/main/scala/org/http4s/QueryParam.scala
@@ -66,16 +66,6 @@ object QueryParamEncoder {
         fa.contramap(f)
     }
 
-  @deprecated("Use QueryParamEncoder[U].contramap(f)", "0.16")
-  def encodeBy[T, U](f: T => U)(
-    implicit qpe: QueryParamEncoder[U]
-  ): QueryParamEncoder[T] =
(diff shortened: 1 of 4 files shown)


tpolecat:
I have no sense of your merge strategy for any of this. There are a lot of versions flying around. This isn't a ton of work so I can do it again later if needed. 

If this is the grand unification branch I'm not clear on what 0.18.x is … do I need to pull these commits forward?

rossabaker:
release-0.16.x -> master -> release-0.18.x.  The second of those will be easy, because it's all cats.  We usually just silently do that, and PR the tricky ones.

I think we have the pieces we need here.  I normally wait for a second sign-off, but this was pretty uncontroversial on 0.16.  I'm just going to do it.

tpolecat:
Muy bien.

rossabaker:
Oh, well, one comment: I don't like to be too aggressive about removing deprecation, because I'm continually surprised how old people upgrade from, and because they're effectively typesafe migration guides.  I tend to purge deprecation when it becomes hard to drag around, and not after some arbitrary number of releases.  I'd argue for reverting 66acc4a.

tpolecat:
As you wish.

PR #1306, opened 2017-08-03
[WIP] 0.18: Merge master and port multipart

aeons:
This is a merge of master and an attempt at porting multipart to fs2-0.10.

It has the nasty workaround for `AttributeKey`s.

Currently, `MultipartSpec` doesn't terminate. I haven't been able to find out why. Specifically, `unsafeRunSync` on decoded in `encodeAndDecodeMultipartWithBinaryFormData` never returns.


aeons:
It's basically only the last commit that's interesting.

aeons:
The last two. I forgot to actually push the `AttributeKey` changes.

There are also some intermittent failures in `AsyncHttpClientSpec` that I haven't been able to figure out.

rossabaker:
I think the unspeakable things done to `AttributeKey` are about the best we can do.

The intermittent failures in `AsyncHttpClientSpec` send a chill down my spine.  Do you have an example of how they fail?

aeons:
Not on that computer now, but either "POST a normal body" times out, "POST a chunked body" fails with an empty body or both of them fails.

rossabaker:
I suspect something in the transition from Strategy to ExecutionContext, but I'm not yet sure what.

aeons:
`MultipartSpec` still hangs, and `AHC` still has intermittent issues.
It seems that these issues stem from the `fs2-reactive-streams` upstream, since I've seen something like that other places as well.

rossabaker:
I was just sitting down to finish that.  Nice work!

rossabaker:
MultipartSpec goes into an infinite loop in the decoder run on line 122.

rossabaker:
It's not infinite.  It just takes 6-7 minutes, on either Travis or my laptop. 😖

rossabaker:
Confirmed that async-http-client is flaky with or without that slow multipart test.

rossabaker:
The failing tests are exactly the tests that use the ReactiveStreamsBodyGenerator, which is fine in 0.17.  So yeah: I suspect this is a problem with fs2-reactive-0.2.

rossabaker:
It's getting stuck [here](https://github.com/zainab-ali/fs2-reactive-streams/blob/217919a7f6d2759ea6c3b0e8b7c18bbc6ec7ab5f/core/src/main/scala/fs2/interop/reactivestreams/StreamSubscription.scala#L120).  I don't know why `rap` is winning the race just to tell us we don't have more requests queued up.

rossabaker:
zainab-ali/fs2-reactive-streams#27 fixes the async-http-client.

rossabaker:
For the sake of our build and the community build, we should consider a smaller multipart binary until we have time to optimize that.

aeons:
I updated `fs2-reactive-streams` to the new version so the AHC stuff works now.

I updated some code that didn't work with the new version of `jawn` brought in by the update to `jawn-fs2` and updated `jawn-json4s`.

However, `circe` still depends on the older version of `jawn`, so the `circe` module doesn't compile currently.

`circe` master is on both `cats-1.0-MF` and `jawn-0.11`, so a new release of that should fix those issues.

How do we feel about merging this in so I can start working on getting the newer updates from master merged in?

rossabaker:
If we hold back 9da3db2, does the build pass?  This is already huge, but I'm reluctant to merge red over green.  I'd suggest we temporarily disable circe, but that has fallout on examples and tuts.

aeons:
No, that commit only fixes stuff introduced by the bump in jawn from updating jawn-fs2 in https://github.com/http4s/http4s/pull/1306/commits/737cb69fae2528112a85cd03bb1135ad731d6c69.

New release of circe seems to be very close though.

rossabaker:
We're going to break the community build again if we merge this in its current state.

Other than this PR getting even bigger, is there any reason we can't do the master merge to here while we wait on the circe release?

aeons:
That's fine by me. I'll do the work in another branch and PR to the origin of this PR, if that is ok.
Might be a bit busy over the weekend though, so don't expect too much :)

Once 0.18 is up to speed with master, I'd like to start looking at getting scalafmt in.

rossabaker:
:+1:

ChristopherDavenport:
Non Termination Issue... Tests looked good, but no ouput afterwards meant something didn't end cleanly.

PR #1307, opened 2017-08-04
Provide a way to retrieve all cookies from a response

### LINKED ISSUE #1299: `org.http4s.headers.Set-Cookie` should be allowed to return more than one header
A server is allowed to return multiple Set-Cookie headers, but is not allowed to fold them into a single field: 
> https://tools.ietf.org/html/rfc6265#section-3
> Origin servers SHOULD NOT fold multiple Set-Cookie header fields into
>   a single header field.  The usual mechanism for folding HTTP headers
> fields (i.e., as defined in [RFC2616]) might change the semantics of
>   the Set-Cookie header field because the %x2C (",") character is used
>   by Set-Cookie in a way that conflicts with such folding.

`Set-Cookie` extends `HeaderKey.Singleton` which yields an optional result.

`HeaderKey.Recurring` signals that recurring headers of the same kind are allowed to be folded into one header-row so it is not usable for Set-Cookie. There needs to be a third way to conform to the spec:

> https://tools.ietf.org/html/rfc6265#section-3.1
> As shown in the next example, the server can store multiple cookies
>   at the user agent.  For example, the server can store a session
>   identifier as well as the user's preferred language by returning two
>   Set-Cookie header fields.  Notice that the server uses the Secure and
>   HttpOnly attributes to provide additional security protections for
>   the more sensitive session identifier (see Section 4.1.2.)
> 
> == Server -> User Agent ==
> 
> Set-Cookie: SID=31d4d96e407aad42; Path=/; Secure; HttpOnly
> Set-Cookie: lang=en-US; Path=/; Domain=example.com

Related issue: #1143

rossabaker:
Set-Cookie is such an odd, out-of-spec header, I propose that it ceases to be `Extractable` at all.  We instead provide an accessor for the cookies on the Response that returns the entire list.

Fixes #1299.

/cc @megri

### DIFF
--- a/core/src/main/scala/org/http4s/Message.scala
+++ b/core/src/main/scala/org/http4s/Message.scala
@@ -304,6 +304,12 @@ final case class Response(
 
   override def toString: String =
     s"""Response(status=${status.code}, headers=$headers)"""
+
+  /** Returns a list of cookies from the [[org.http4s.headers.Set-Cookie]]
+    * headers. Includes expired cookies, such as those that represent cookie
+    * deletion. */
+  def cookies: List[Cookie] =
+    `Set-Cookie`.from(headers).map(_.cookie)
 }
 
 object Response {
--- a/core/src/main/scala/org/http4s/Headers.scala
+++ b/core/src/main/scala/org/http4s/Headers.scala
@@ -37,6 +37,10 @@ final class Headers private (headers: List[Header])
     */
   def get(key: HeaderKey.Extractable): Option[key.HeaderT] = key.from(this)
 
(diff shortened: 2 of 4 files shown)


core/src/main/scala/org/http4s/Message.scala
```diff
@@ -304,6 +304,12 @@ final case class Response(
 
   override def toString: String =
     s"""Response(status=${status.code}, headers=$headers)"""
+
+  /** Returns a list of cookies from the [[org.http4s.headers.Set-Cookie]]
+    * headers. Includes expired cookies, such as those that represent cookie
+    * deletion. */
+  def cookies: List[Cookie] =
```
bryce-anderson: Would this make sense as a `lazy val`?

rossabaker: Laziness has tended to bite us.  I think if we did go that route, we'd more likely want to go the null init pattern we've used elsewhere (e.g., CI hash code, parsed header value).

PR #1310, opened 2017-08-07
WIP: force headers to be valid

rossabaker:


rossabaker:
This is what it starts to look like if we force headers to be valid .  The domain of a String is wider than the domain of either a field name or a field value.

This is roughly the same logic as, but a more extreme version of, `HttpDate`.

I like the FieldName (it's more descriptive and hardly less convenient than CaseInsensitiveString), but am not enjoying the ergonomics FieldValue.  Though we'd want both to use types to enforce correctness.  How can we make this better?

rossabaker:
Stale and, in current form, cumbersome.

PR #1311, opened 2017-08-08
cats 1.0.0-MF, fs2 0.10.0-M5

tpolecat:
As it says on the tin. Note that circe hasn't been updated to 1.0.0-MF yet … the test pass but it's probably flaky in real life. Options include turning that module off for the milestone or saying  `¯\_(ツ)_/¯` which is what I would probably do.


rossabaker:
This should be sufficient for a snapshot.  Going to merge this, and cut a milestone tomorrow morning unless @aeons screams in pain from this.

tpolecat:
Muy bien.

rossabaker:
It would be really nice to understand what's wrong with #1306, since this is a little behind master and trying to catch it up led to problems.  I just haven't had time to diagnose it.

PR #1313, opened 2017-08-13
Provide content-length header when serving from a URL

### LINKED ISSUE #1295: Provide Content-Length header when serving static files from URL
Currently, `StaticFile.fromURL` doesn't provide a `Content-Length` header. It uses `Transfer-Encoding: chunked` instead. We can easily get the content length of a resource with `url.openConnection.getContentLengthLong`.

rossabaker:
A bunch of file I/O operations are unguarded by task, so sweep all of `StaticFile` into an `OptionT`.

Then, once safely tucked away, we can call `getContentLengthLong` on the `URLConnection`.

Fixes #1295.

/cc @ZizhengTai

### DIFF
--- a/tests/src/test/scala/org/http4s/StaticFileSpec.scala
+++ b/tests/src/test/scala/org/http4s/StaticFileSpec.scala
@@ -109,5 +109,12 @@ class StaticFileSpec extends Http4sSpec {
       // things like CachingChunkWriter that buffer the chunks.
       new String(Chunk.concat(s.chunks.runLog.unsafeRun).toArray, "utf-8") must_== expected
     }
+
+    "Set content-length header from a URL" in {
+      val url = getClass.getResource("/lorem-ipsum.txt")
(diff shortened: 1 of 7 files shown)


tests/src/test/scala/org/http4s/StaticFileSpec.scala
```diff
@@ -100,12 +101,20 @@ class StaticFileSpec extends Http4sSpec {
...
         .fold[EntityBody](sys.error("Couldn't find resource"))(_.body)
       // Expose problem with readInputStream recycling buffer.  chunks.runLog
       // saves chunks, which are mutated by naive usage of readInputStream.
       // This ensures that we're making a defensive copy of the bytes for
       // things like CachingChunkWriter that buffer the chunks.
       new String(Chunk.concat(s.chunks.runLog.unsafeRun).toArray, "utf-8") must_== expected
     }
+
+    "Set content-length header from a URL" in {
+      val url = getClass.getResource("/lorem-ipsum.txt")
```
ChristopherDavenport: Not used if following test.

PR #1314, opened 2017-08-14
Stop Leaking Secrets in Request and Response toString

### LINKED ISSUE #1312: Logging leaks private user data when a request fails
If a request fails the full request object is written to logs by calling `toString` on the request object. This leads to leaking data such as Authentication headers with sensitive credentials being written to logs. 

At the moment there is no easy way to hook in to `Http1ServerStage.runRequest` which calls `internalServerError` resulting in the log message.

It's useful to dump the request for debugging purposes but this also imposes potential security issues. It would be great if some plugable error handler could be used so a user can choose to log or not and the contents of the log message.

Based on this any logging elsewhere in HTTP4s which calls Request.toString also should be refactored to allow overriding the error handling to prevent data leaks in to logging systems

ChristopherDavenport:
Fixes #1312

### DIFF
--- a/core/src/main/scala/org/http4s/Headers.scala
+++ b/core/src/main/scala/org/http4s/Headers.scala
@@ -108,6 +108,12 @@ final class Headers private (headers: List[Header])
     */
   def removePayloadHeaders: Headers =
     filterNot(h => Headers.PayloadHeaderKeys(h.name))
+
+  def redactSensitive(sensitiveHeaderNames: Set[CaseInsensitiveString] = Headers.SensitiveHeaders): Headers =
+    headers.map {
+      case h if sensitiveHeaderNames.contains(h.name) => Header.Raw(h.name, "<REDACTED>")
+      case h => h
+    }
 }
 
 object Headers {
--- a/core/src/main/scala/org/http4s/Headers.scala
+++ b/core/src/main/scala/org/http4s/Headers.scala
@@ -135,4 +135,10 @@ object Headers {
     "Trailer".ci,
     "Transfer-Encoding".ci
   )
+
+  val SensitiveHeaders = Set(
+    "Authorization".ci,
+    "Cookie".ci,
+    "Set-Cookie".ci,
+  )
 }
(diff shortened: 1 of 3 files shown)


core/src/main/scala/org/http4s/Message.scala
```diff
@@ -211,8 +211,10 @@ final case class Request(
   def decodeWith[A](decoder: EntityDecoder[A], strict: Boolean)(f: A => Task[Response]): Task[Response] =
     decoder.decode(this, strict = strict).fold(_.toHttpResponse(httpVersion), f).join
 
-  override def toString: String =
-    s"""Request(method=$method, uri=$uri, headers=${headers})"""
+  override def toString: String = {
+    val newHeaders = headers.filter(_.is(Authorization)).filter(_.is(org.http4s.headers.Cookie))
```
rossabaker: One thing I've done in some other clients is to replace the sensitive header value with `<REDACTED>`.  Then there's evidence of it being there, but it still doesn't leak.  Then it's a map to a raw header, and pass through non-matching ones as-is.

We might also want to use the `CaseInsenstiveString` getter, which will handle invalid values, instead of the `HeaderKey`.  That way we don't leak a credential just because someone forgot a quotation mark.

rossabaker:
This seems relevant to following redirects too.

core/src/main/scala/org/http4s/Message.scala
```diff
@@ -105,6 +108,11 @@ object Message {
   object Keys {
     val TrailerHeaders = AttributeKey[Task[Headers]]
   }
+
+  def redactHeaders(headers: Headers): Headers = headers.map {
```
rossabaker: I put the payload headers as a `Headers` itself.  Perhaps that's a better home for this?

ChristopherDavenport: I agree, didn't really like it.

core/src/main/scala/org/http4s/Headers.scala
```diff
@@ -100,6 +100,12 @@ final class Headers private (headers: List[Header])
       this.toList.equals(otherheaders.toList)
     case _ => false
   }
+
+  def redactSensitive(sensitiveHeaderNames: Set[CaseInsensitiveString] = Headers.SensitiveHeaders): Headers =
```
rossabaker: This is a good idea.  Could generalize to `CaseInsensitiveString => Boolean`, which is what that Set is.  We did something like this in CORS.

rossabaker:
Looks at build.  Cries.  The 2.10 pain will be over soon.

PR #1315, opened 2017-08-15
Fixed Part name returning Content-Disposition to return form-data name parameter.

jmcardon:


ChristopherDavenport:
Also, thank you for contributing!

rossabaker:
Yes, thank you.

No thanks to me, who said this would be cherry-picked, and then I merged to the wrong branch. 😩

PR #1317, opened 2017-08-16
Return index.html when a directory is specified for static content

jedesah:
- Change default behavior of `staticContent` to return the index.html file (when present) when the client asks for a directory
- Make it easier to extend or change the default behavior by removing the `private` modifier so it can be called from external implementations

### DIFF
--- a/server/src/main/scala/org/http4s/server/staticcontent/FileService.scala
+++ b/server/src/main/scala/org/http4s/server/staticcontent/FileService.scala
@@ -26,7 +26,7 @@ object FileService {
     */
   final case class Config(systemPath: String,
                           pathPrefix: String = "",
-                          pathCollector: (File, Config, Request) => Task[Option[Response]] = filesOnly,
+                          pathCollector: (File, Config, Request) => Task[Option[Response]] = defaultFileCollector,
                           bufferSize: Int = 50*1024,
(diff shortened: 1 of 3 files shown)


server/src/main/scala/org/http4s/server/staticcontent/FileService.scala
```diff
@@ -47,8 +47,8 @@ object FileService {
   /* Returns responses for static files.
    * Directories are forbidden.
    */
-  private def filesOnly(file: File, config: Config, req: Request): Task[Option[Response]] = Task.now {
-    if (file.isDirectory()) Some(Response(Status.Unauthorized))
+  def defaultFileCollector(file: File, config: Config, req: Request): Task[Option[Response]] = Task.now {
+    if (file.isDirectory) StaticFile.fromFile(new File(file, "index.html"), Some(req))
```
ChristopherDavenport: What is generated if the file does not exist? As this now generates `None` rather than unauthorized. What would be returned? Could we fold our option into a guaranteed response here?

rossabaker: That's an excellent question.  `None` is nice because it can chain into something else, like an automated directory listing.  Although I'd expect 403 (_not_ 401) would be the most commonly desired default here.

PR #1319, opened 2017-08-16
Dependency Upgrades: 0.17 Edition

rossabaker:


ChristopherDavenport:
EntityBodyWriter- deflated resource and GzipSpec in different runs. Seems like compression or scodec are likely sources.

rossabaker:
scodec and fs2 got upgraded.  I think we can't not upgrade fs2.

rossabaker:
Failing tests removed in favor of https://github.com/functional-streams-for-scala/fs2/pull/921.  We weren't testing http4s, we were testing fs2.

PR #1326, opened 2017-08-21
Update Circe to 0.9.0-M1

tpolecat:
No more cats 0.9 dependencies!

I think.


rossabaker:
This is Travissing on #1306.  If that fixes the outstanding issues, we can release 0.18.0-M1.

PR #1328, opened 2017-08-21
Add syntax for `MaybeResponse`

jedesah:


jedesah:
Yes, the use case is testing. `orNotFound` could work or I was thinking a `total` function on `ServiceOps` that returns a `Kleisli[Task, Request, Response]` where `NotFound` is returned in the case of a `Pass`.

jedesah:
But agreed that this might not be the greatest. Although it does mirror `TaskResponseOps`

PR #1329, opened 2017-08-22
Add syntax to make a `HttpService` "total"

jedesah:
I think this is very convenient for testing. If only for migration purposes alone.

### DIFF
--- a/core/src/main/scala/org/http4s/syntax/ServiceSyntax.scala
+++ b/core/src/main/scala/org/http4s/syntax/ServiceSyntax.scala
@@ -1,6 +1,7 @@
 package org.http4s
 package syntax
 
+import scalaz.concurrent.Task
 import scalaz.Kleisli
 
 trait ServiceSyntax {
@@ -9,6 +10,9 @@ trait ServiceSyntax {
 }
 
 final class ServiceOps[A](self: Service[A, MaybeResponse]) {
+  def orNotFound(a: A): Task[Response] =
+    orNotFound.run(a)
+
   def orNotFound: Service[A, Response] =
     Kleisli(a => self.run(a).map(_.orNotFound))
 }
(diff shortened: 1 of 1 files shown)


rossabaker:
That approach should also work in 0.17, because cats' Kleisli has an `apply` method.

Note that I am compiling both versions in my head, and if scalac and I disagree, scalac is right. :)

jedesah:
@rossabaker I think you may be right. I'll ask scalac

rossabaker:
Ugh.  It thinks the application is an explicit evidence parameter, not an `A`.

jedesah:
Oops... I guess I just compiled `core`

rossabaker:
I was more troubled by having two methods that do about the same thing than by the name `total`.  Maybe we could have syntax on a `Service[A, MaybeResponse]` instead of a `Service[A, B]` with an evidence parameter and still simplify it to one signature.  We'd lose the ability to call it on a `Service[A, Response]` or a `Service[A, Pass]`, but I doubt that's a use case.

Or I'm also fine with the original version, if you think that's superior.

jedesah:
I like `Service[A, MaybeResponse]`. If you know you have either a `Service[A, Response]` or a `Service[A, Pass]` I can't think of any good reason why would you want to call `orNotFound` on it.

rossabaker:
The TickWheelExecutor failure is an intermittent failure that I introduced in #1322.

The 2.12 failures are something I've seen before.  I've had to be explicit about calling `apply` or `run`.  I haven't figured out which specs2 implicit is getting in the way yet.

rossabaker:
Leaving the overload preserves compatibility on Scala 2.10, which can't deal with the implicit conversion to Function1.

Beginning on 0.17, we have a Kleisli with an `apply` and drop 2.10, so I think we can drop that overload for two reasons when we merge.

rossabaker:
Looks like you can't request a review from the original submitter. :)  @jedesah, are you happy with my amendment?

jedesah:
@rossabaker Looks good. What's the process for merging into `0.17`?

rossabaker:
0.17 is on master right now.  There are a couple other things to merge as well.  I'll round those all up at once, as I don't expect any of them to be complicated.

PR #1333, opened 2017-08-23
Fixes deprecation message for decodeBy

aspic:


rossabaker:
Thanks!  Cherry-picked onto 0.16, where the error originated.

rossabaker:
BTW, the build failure looks like something I introduced in #1332.

PR #1334, opened 2017-08-23
[0.18] Refactor StreamApp and fix ServerBuilder.serve

aeons:
I changed `ServerBuilder.serve` back to something that makes sense.

There's also yet another attempt at refactoring `StreamApp`. It tests and I have checked it in a small example app as well.

### DIFF
--- a/core/src/main/scala/org/http4s/util/StreamApp.scala
+++ b/core/src/main/scala/org/http4s/util/StreamApp.scala
@@ -7,15 +7,17 @@ import fs2._
 import fs2.async.mutable.Signal
 import org.log4s.getLogger
 
+import scala.concurrent.ExecutionContext
+
 abstract class StreamApp[F[_]](implicit F: Effect[F]) {
   private[this] val logger = getLogger(classOf[StreamApp[F]])
 
   def stream(args: List[String], requestShutdown: F[Unit]): Stream[F, Nothing]
 
-  private implicit val executionContext = TrampolineExecutionContext
+  private implicit val executionContext: ExecutionContext = TrampolineExecutionContext
 
   /** Adds a shutdown hook that interrupts the stream and waits for it to finish */
-  private[util] def addShutdownHook(requestShutdown: Signal[F, Boolean], halted: Signal[IO, Boolean]): F[Unit] =
+  private def addShutdownHook(requestShutdown: Signal[F, Boolean], halted: Signal[IO, Boolean]): F[Unit] =
     F.delay(sys.addShutdownHook {
       val hook = requestShutdown.set(true).runAsync(_ => IO.unit) >> halted.discrete.takeWhile(_ == false).run
       hook.unsafeRunSync()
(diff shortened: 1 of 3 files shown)


core/src/main/scala/org/http4s/util/StreamApp.scala
```diff
@@ -14,34 +14,35 @@ abstract class StreamApp[F[_]](implicit F: Effect[F]) {
...
-      }))
-      res <- stream(args.toList, shutdownRequested.set(true))
-        .interruptWhen(shutdownRequested)
-        .onFinalize(halted.set(true))
-        .as(())
-    } yield res
-    s.drain
-  }
+  /** Adds a shutdown hook that interrupts the stream and waits for it to finish */
+  private[util] def addShutdownHook(requestShutdown: Signal[F, Boolean], halted: Signal[IO, Boolean]): F[Unit] =
```
rossabaker: Who calls this outside of `StreamApp`?

core/src/main/scala/org/http4s/util/StreamApp.scala
```diff
@@ -14,34 +14,35 @@ abstract class StreamApp[F[_]](implicit F: Effect[F]) {
...
+    F.delay(sys.addShutdownHook {
+      val hook = requestShutdown.set(true).runAsync(_ => IO.unit) >> halted.discrete.takeWhile(_ == false).run
+      hook.unsafeRunSync()
```
rossabaker: Why are we doing this?  To block the shutdown hook until halt is confirmed?

server/src/main/scala/org/http4s/server/ServerBuilder.scala
```diff
@@ -50,7 +50,7 @@ trait ServerBuilder[F[_]] {
    * that runs for the rest of the JVM's life.
    */
   final def serve(implicit F: Async[F]): Stream[F, Nothing] =
-    Stream.bracket(start)(_ => Stream.empty, _.shutdown)
+    Stream.bracket(start)((_: Server[F]) => Stream.eval_(F.async[Unit](_ => ())), _.shutdown)
```
aeons: ssshhh

core/src/main/scala/org/http4s/util/StreamApp.scala
aeons: Noone, was just a sloppy copy/paste from the `doMain`.

core/src/main/scala/org/http4s/util/StreamApp.scala
aeons: The whole thing? It was there before I ported it to cats.

I assume the idea is to allow the server stream to finish before the JVM dies, when a shutdown is started from somewhere outside.

server/src/main/scala/org/http4s/server/ServerBuilder.scala
rossabaker: Once it's on a release branch, it's everybody's fault. :)

PR #1336, opened 2017-08-23
Fix intermittent failure related to TickWheelExecutor shutdown

rossabaker:
I think I broke this in #1332.


rossabaker:
Going to merge this one pretty aggressively.  It's just tests, and the flakiness I introduced is breaking other builds.

PR #1337, opened 2017-08-23
Revert to Lightbend compiler

### LINKED ISSUE #1330: Don't depend on Typelevel scala library
org.http4s packages depend on `org.typelevel.scala-library`. SBT fails to reconcile it with `org.scala-lang.scala-library`, leading to two conflicting library dependencies, and non-deterministic problems like https://github.com/tpolecat/doobie/issues/568 . This an SBT issue first, but it would be better to work around it at http4s end too.

Gitter discussion: https://gitter.im/sbt/sbt?at=599b9f2b9acddb2407b69a8b

rossabaker:
I'd prefer to stay on Typelevel Scala in order to help test it and promote it, but we don't need its extended features yet, and the resulting POMs are causing some pain.  I hope to change back after the dust settles, but it's more important to ship our releases now.

Fixes #1330.

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -12,6 +12,7 @@ addSbtPlugin("io.gatling"         %  "gatling-sbt"           % "2.1.5")
 addSbtPlugin("io.get-coursier"    %  "sbt-coursier"          % "1.0.0-RC1")
 addSbtPlugin("io.spray"           %  "sbt-revolver"          % "0.8.0")
 addSbtPlugin("io.verizon.build"   %  "sbt-rig"               % "2.0.29")
+addSbtPlugin("org.lyranthe.sbt"   %  "partial-unification"   % "1.1.0")
 addSbtPlugin("org.tpolecat"       %  "tut-plugin"            % "0.4.8")
 addSbtPlugin("pl.project13.scala" %  "sbt-jmh"               % "0.2.24")
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,19 +28,19 @@ jobs:
   include:
     - &tests
       stage: Test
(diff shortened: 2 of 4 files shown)


project/plugins.sbt
```diff
@@ -12,6 +12,7 @@ addSbtPlugin("io.gatling"         %  "gatling-sbt"           % "2.1.5")
 addSbtPlugin("io.get-coursier"    %  "sbt-coursier"          % "1.0.0-RC1")
 addSbtPlugin("io.spray"           %  "sbt-revolver"          % "0.8.0")
 addSbtPlugin("io.verizon.build"   %  "sbt-rig"               % "2.0.29")
+addSbtPlugin("org.lyranthe.sbt"   %  "partial-unification"   % "1.1.0")
```
ChristopherDavenport: Do we need the plugin in 2.11.11?

aeons: Only for 2.10 I think: https://github.com/milessabin/si2712fix-plugin#usage

rossabaker: No.  I _think_ our build will work without this in master.  It hides a lot of ugliness in 0.16.

ChristopherDavenport: Ok, let's sure it get yanked in the merge-train.

ChristopherDavenport: *yanked out

rossabaker: And it's not needed for 2.10 because master drops 2.10.  We'll be trying to yank this and the `orNotFound` overload in 0.17.

ChristopherDavenport:
I feel like we may have missed a golden opportunity for which PR took this number. Rather tame for 1337

PR #1340, opened 2017-08-23
Upgrade to blaze-0.12.7

rossabaker:
Release notes sloppily intertwined because I hopped branches too fast. 🤷‍♂️


rossabaker:
Build is likely to fail if Maven Central isn't synced first.  Feel free to give it a kick if I'm not around.

PR #1341, opened 2017-08-23
Prepare for 0.17.0-RC2

rossabaker:
Catches up to everything that will be in 0.16.0-RC2.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,20 +28,20 @@ jobs:
   include:
     - &tests
       stage: Test
-      scala: 2.12.3-bin-typelevel-4
+      scala: 2.12.3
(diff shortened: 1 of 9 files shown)


rossabaker:
Yes, I removed the overload.  It had some tolerable fallout when flatmapping a `Task[Request]`.  (I wish Kleisli had a method to apply an `F[A]`.)  I also removed the partial unification plugin, since it Just Works in all supported Scala versions.

rossabaker:
Hmm.  I think the operation I'm thinking of is called `<<=` in Scalaz and not present in Cats.  Anyway, we just flatMap it to `.run`.

PR #1343, opened 2017-08-24
[0.18] Merge master

aeons:


ChristopherDavenport:
Probably wait till we merge in #1344 ?

rossabaker:
I think #1344 can just bubble up on its own.  It should be a minor merge.

aeons:
It's changed manually here as well, since that's how i noticed it.

PR #1346, opened 2017-08-24
BlazeBuilder is now setting the custom execution context passed in.

### LINKED ISSUE #1345: Blaze Server: can't set a custom Execution Context
`BlazeBuilder.withExecutionContext` is ignoring the parameter passed in. See https://github.com/http4s/http4s/blob/master/blaze-server/src/main/scala/org/http4s/server/blaze/BlazeServer.scala#L91

fabiocognigni:
This closes #1345.

### DIFF
--- a/blaze-server/src/test/scala/org/http4s/server/blaze/BlazeServerSpec.scala
+++ b/blaze-server/src/test/scala/org/http4s/server/blaze/BlazeServerSpec.scala
@@ -2,6 +2,10 @@ package org.http4s
 package server
 package blaze
 
+import java.util.concurrent.Executors
+
+import scala.concurrent.ExecutionContext
+
 class BlazeServerSpec extends ServerAddressSpec {
   def builder = BlazeBuilder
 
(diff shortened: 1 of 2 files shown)


blaze-server/src/test/scala/org/http4s/server/blaze/BlazeServerSpec.scala
```diff
@@ -17,6 +21,20 @@ class BlazeServerSpec extends ServerAddressSpec {
...
       true must_== true
     }
+
+    "Startup with a custom ExecutionContext" in {
+      val s = BlazeBuilder
+        .bindAny()
+        .withExecutionContext(
+          ExecutionContext.
+            fromExecutorService(
+              Executors.newFixedThreadPool(5)))
```
rossabaker: We need to shut this down, or this will leak threads on test runs.  Not a big deal for a single Travis run, but this will eventually kill a long-running sbt.

Unless we're confirming that code runs on the provided executor, I think we can eliminate this test.  I think this would have passed even with the bug.

rossabaker: It would be a terribly verbose test for a parameter change, but to go red-to-green, an execution context that simply records that work was done, a request to the server, and then an assertion on that recording would do the trick, I think.

fabiocognigni: Agreed. Couldn't find a way to confirm execution on provided executor in the tests. I'm going to remove this in a separate PR.

fabiocognigni: It has been removed already with: https://github.com/http4s/http4s/pull/1347

ChristopherDavenport: I'm pretty fast on fixes when it's my fault.

PR #1348, opened 2017-08-25
Replace EntityEncoder[Byte] with EntityEncoder[EntityBody]

rossabaker:
The existing `EntityEncoder[Byte]` is inefficient.  This replacement encoder preserves chunking of the provided or contramapped stream.


rossabaker:
ByteVectorChunks would need to be emitted by whatever streams are encoded with `entityBodyEncoder`.  Which I don't think `fs2.io` does.  I have not kept pace with fs2 on whether concatenation of regular byte chunks has been improved, which would mitigate some of the advantage of a ByteVectorChunk in the first place.

PR #1350, opened 2017-08-28
 Flush header while awaiting first chunk

rossabaker:
A message might have a header substantially before the first chunk is calculated.  We should render the header, sans Content-Length or Transfer-Encoding header, as soon as it's available.  We call a `writeHeader` method with the header buffer, which gives each encoder the discretion to flush the header immediately or to hold it to be flushed with the beginning of the entity.

/cc @scottcarey, who mentioned this behavior on Gitter

### DIFF
--- a/blaze-core/src/main/scala/org/http4s/blazecore/util/BodylessWriter.scala
+++ b/blaze-core/src/main/scala/org/http4s/blazecore/util/BodylessWriter.scala
@@ -13,6 +13,7 @@ import fs2.Stream._
 import fs2.interop.cats._
 import fs2.util.Attempt
 import org.http4s.blaze.pipeline._
+import org.http4s.util.StringWriter
 
 /** Discards the body, killing it so as to clean up resources
   *
@@ -20,26 +21,24 @@ import org.http4s.blaze.pipeline._
   * @param pipe the blaze `TailStage`, which takes ByteBuffers which will send the data downstream
   * @param ec an ExecutionContext which will be used to complete operations
   */
-class BodylessWriter(headers: ByteBuffer, pipe: TailStage[ByteBuffer], close: Boolean)
-                    (implicit protected val ec: ExecutionContext) extends EntityBodyWriter {
+class BodylessWriter(pipe: TailStage[ByteBuffer], close: Boolean)
+  (implicit protected val ec: ExecutionContext) extends Http1Writer {
+  private[this] var headers: ByteBuffer = null
 
-  private lazy val doneFuture = Future.successful( () )
+  def writeHeader(headerWriter: StringWriter): Future[Unit] =
+    pipe.channelWrite(Http1Writer.headersToByteBuffer(headerWriter.result))
 
   /** Doesn't write the entity body, just the headers. Kills the stream, if an error if necessary
     *
     * @param p an entity body that will be killed
     * @return the Task which, when run, will send the headers and kill the entity body
     */
-  override def writeEntityBody(p: EntityBody): Task[Boolean] = Task.async { cb =>
-    val callback = cb.compose((t: Attempt[Unit]) => t.map(_ => close))
+  override def writeEntityBody(p: EntityBody): Task[Boolean] =
+    p.drain.run.map(_ => close)
 
-    pipe.channelWrite(headers).onComplete {
-      case Success(_) => p.open.close.run.unsafeRunAsync(callback)
-      case Failure(t) => p.pull(_ => Pull.fail(t)).run.unsafeRunAsync(callback)
-    }
-  }
+  override protected def writeEnd(chunk: Chunk[Byte]): Future[Boolean] =
+    Future.successful(close)
 
-  override protected def writeEnd(chunk: Chunk[Byte]): Future[Boolean] = doneFuture.map(_ => close)
-
-  override protected def writeBodyChunk(chunk: Chunk[Byte], flush: Boolean): Future[Unit] = doneFuture
+  override protected def writeBodyChunk(chunk: Chunk[Byte], flush: Boolean): Future[Unit] =
+    FutureUnit
 }
--- a/blaze-core/src/main/scala/org/http4s/blazecore/util/BodylessWriter.scala
+++ b/blaze-core/src/main/scala/org/http4s/blazecore/util/BodylessWriter.scala
@@ -21,11 +21,11 @@ import org.http4s.util.StringWriter
   * @param pipe the blaze `TailStage`, which takes ByteBuffers which will send the data downstream
   * @param ec an ExecutionContext which will be used to complete operations
   */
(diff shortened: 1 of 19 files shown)


blaze-core/src/main/scala/org/http4s/blazecore/util/BodylessWriter.scala
```diff
@@ -13,33 +13,32 @@ import fs2.Stream._
...
 
 /** Discards the body, killing it so as to clean up resources
   *
   * @param headers ByteBuffer representation of [[Headers]] to send
   * @param pipe the blaze `TailStage`, which takes ByteBuffers which will send the data downstream
   * @param ec an ExecutionContext which will be used to complete operations
   */
-class BodylessWriter(headers: ByteBuffer, pipe: TailStage[ByteBuffer], close: Boolean)
-                    (implicit protected val ec: ExecutionContext) extends EntityBodyWriter {
+class BodylessWriter(pipe: TailStage[ByteBuffer], close: Boolean)
```
rossabaker: Maybe these concrete types should be removed from the public API.

blaze-core/src/main/scala/org/http4s/blazecore/util/BodylessWriter.scala
```diff
@@ -13,33 +13,32 @@ import fs2.Stream._
...
 
   /** Doesn't write the entity body, just the headers. Kills the stream, if an error if necessary
     *
     * @param p an entity body that will be killed
     * @return the Task which, when run, will send the headers and kill the entity body
     */
-  override def writeEntityBody(p: EntityBody): Task[Boolean] = Task.async { cb =>
-    val callback = cb.compose((t: Attempt[Unit]) => t.map(_ => close))
+  override def writeEntityBody(p: EntityBody): Task[Boolean] =
+    p.drain.run.map(_ => close)
```
rossabaker: I think thisi s a safe refactoring, with the work now done in `writeHeader`, but a second set of brains would be appreciated here.

blaze-core/src/main/scala/org/http4s/blazecore/util/package.scala
```diff
@@ -1,12 +1,16 @@
...
 
+import scala.concurrent.Future
 import fs2._
 
 package object util {
   /** Used as a terminator for streams built from repeatEval */
   private[http4s] val End = Right(None)
 
+  private[http4s] val FutureUnit =
+    Future.successful(())
```
rossabaker: I'll bet there are a ton of these in blaze proper.  This is probably a constant that belongs down there, whence the futures come.

aeons: There's `scala.concurrent.Future.unit`, but that might be a 2.12 thing.

rossabaker: It is 2.12 only. :(

blaze-core/src/main/scala/org/http4s/blazecore/util/Http1Writer.scala
```diff
@@ -0,0 +1,28 @@
...
+import fs2._
+import org.http4s.blaze.http.Headers
+import org.http4s.blaze.pipeline.TailStage
+import org.http4s.blaze.http.http20.NodeMsg._
+import org.http4s.util.StringWriter
+import org.http4s.util.chunk._
+
+trait Http1Writer extends EntityBodyWriter {
+  final def write(headerWriter: StringWriter, body: EntityBody): Task[Boolean] =
+    Task.fromFuture(writeHeader(headerWriter)).flatMap(_ => writeEntityBody(body))
```
bryce-anderson: If the write of the headers fails nobody is going to execute the body, but maybe that's okay? It's not crazy to envision an `EntityBody` that is already holding onto resources like a socket connection.

blaze-core/src/main/scala/org/http4s/blazecore/util/Http1Writer.scala
```diff
@@ -0,0 +1,28 @@
...
+
+  /* Writes the header.  It is up to the writer whether to flush immediately or to
+   * buffer the header with a subsequent chunk. */
+  def writeHeader(headerWriter: StringWriter): Future[Unit]
```
bryce-anderson: s/writerHeader/writerHeaders

blaze-core/src/main/scala/org/http4s/blazecore/util/Http1Writer.scala
```diff
@@ -0,0 +1,28 @@
...
+import scala.concurrent._
+
+import fs2._
+import org.http4s.blaze.http.Headers
+import org.http4s.blaze.pipeline.TailStage
+import org.http4s.blaze.http.http20.NodeMsg._
+import org.http4s.util.StringWriter
+import org.http4s.util.chunk._
+
+trait Http1Writer extends EntityBodyWriter {
```
bryce-anderson: I think this concern spans more than just `HTTP/1.x` so perhaps another name is in order.

bryce-anderson:
Oh, and you might consider taking the opportunity to purge pretty much all of this from the public API: I don't see a reason to expose these details.

blaze-core/src/main/scala/org/http4s/blazecore/util/Http1Writer.scala
rossabaker: It's named thus because Http2Writer needs real `Header`s, to create the frame, whereas we have some optimizations passing around the mutable StringWriter.  But you're not wrong.

blaze-core/src/main/scala/org/http4s/blazecore/util/Http1Writer.scala
rossabaker: That's a good point.  I guess we need to drain the body on failure.

rossabaker:
Anybody have any thoughts on the tension between our new client timeout focusing on header completion and our new flush trying to hold back that final header?

bryce-anderson:
@rossabaker, perhaps it could be configurable? I believe I wrote the
caching writer is to guard against a potentially common case of a single
chunk when there aren't any length or encoding specified. With the length
being provided by the entity encoding process, it may not be as common as I
had initially thought so maybe it could be dropped.

rossabaker:
I updated FlushingChunkWriter to be more aggressive, as it was always going to do chunked transfers anyway.  (In fs2-0.9, and maybe 0.10, we can't tell that a chunk is the last).

CachingChunkWriter still has its original behavior.  One could say we've made it "configurable" by the headers: if you don't specify either Content-Length or Transfer-Encoding, you get dynamic behavior.  When you ask for one, you get what you requested.

rossabaker:
Yeah, it was a bigger diff than I wanted to introduce for this, but I think it's also one of our better tested areas.

PR #1351, opened 2017-08-28
Add overloads for `jsonOf` and `as`

aeons:
To allow specifying explicit decoders.

### DIFF
--- a/argonaut/src/main/scala/org/http4s/argonaut/ArgonautInstances.scala
+++ b/argonaut/src/main/scala/org/http4s/argonaut/ArgonautInstances.scala
@@ -9,6 +9,15 @@ import org.http4s.argonaut.Parser.facade
 import org.http4s.headers.`Content-Type`
 
 trait ArgonautInstances {
+
+  implicit def http4sArgonautJsonDecoder[F[_]: Sync, A: DecodeJson]: JsonDecoder[F, A] =
+    new JsonDecoder[F, A] {
+      override def decodeJson(message: Message[F]): F[A] =
+        jsonOf[F, A]
+          .decode(message, strict = false)
+          .fold(throw _, identity)
+    }
+
   implicit def jsonDecoder[F[_]: Sync]: EntityDecoder[F, Json] =
     jawn.jawnDecoder
 
--- a/core/src/main/scala/org/http4s/JsonDecoder.scala
+++ b/core/src/main/scala/org/http4s/JsonDecoder.scala
@@ -1,0 +1,7 @@
+package org.http4s
+
+trait JsonDecoder[F[_], A] {
+
+  def decodeJson(message: Message[F]): F[A]
+
+}
--- a/core/src/main/scala/org/http4s/JsonDecoder.scala
+++ b/core/src/main/scala/org/http4s/JsonDecoder.scala
@@ -5,3 +5,9 @@ trait JsonDecoder[F[_], A] {
   def decodeJson(message: Message[F]): F[A]
 
 }
+
+object JsonDecoder {
+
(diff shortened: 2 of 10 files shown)


aeons:
@OlegYch had some objections to using overloads on Gitter: https://gitter.im/http4s/http4s?at=59a489fc162adb6d2e437871

argonaut/src/main/scala/org/http4s/argonaut/ArgonautInstances.scala
```diff
@@ -13,6 +13,9 @@ trait ArgonautInstances {
     jawn.jawnDecoder
 
   def jsonOf[F[_]: Sync, A](implicit decoder: DecodeJson[A]): EntityDecoder[F, A] =
+    jsonOf(decoder)
+
+  def jsonOf[F[_]: Sync, A](decoder: DecodeJson[A]): EntityDecoder[F, A] =
```
rossabaker: Wow, so these overloads don't clash because of the desugaring of the `Sync` bound?  I had to stare at this for a couple minutes to figure out how it could compile. 😖 

Is this necessary to support an explicit `DecodeJson` as opposed to calling `jsonOf[Foo]`?

aeons: Correct. If it helps I can make the implicit parameter, erm, explicit. :)

aeons:
I tried implementing `JsonDecoder` and `decodeJson`.

https://github.com/aeons/http4s/commit/a82b3b75bf93d235a7dc389a78f0dbb99dac48ce

rossabaker:
Where do we stand on this?  Are we leaning toward a82b3b7?

aeons:
I think it works well, and I don't mind specialcasing JSON, since its so common.

rossabaker:
Okay, can you get that commit into a PR that passes?

aeons:
I'll just force push into this one.

aeons:
Added some tests and instances for argonaut and json4s as well.

core/src/main/scala/org/http4s/JsonDecoder.scala
```diff
@@ -0,0 +1,13 @@
+package org.http4s
+
+trait JsonDecoder[F[_], A] {
```
rossabaker: This trait doesn't have anything to do with JSON.  This is more of a typeclass for decoders without any HTTP knowledge.

core/src/main/scala/org/http4s/Message.scala
```diff
@@ -123,6 +123,9 @@ sealed trait Message[F[_]] extends MessageOps[F] { self =>
       implicit F: FlatMap[F],
       decoder: EntityDecoder[F, T]): DecodeResult[F, T] =
     decoder.decode(this, strict = false)
+
+  override def decodeJson[T](implicit jsonDecoder: JsonDecoder[F, T], F: FlatMap[F]): F[T] =
```
rossabaker: I imagined syntax being provided by the JSON support modules.  For example, on `CirceInstances`

```scala
implicit class MessageSyntax(self: Message[F]) {
  def decodeJson[A](implicit dec: Decoder[A], F: FlatMap[F]): F[A]
}
```

aeons: Yea, that makes a lot more sense. It was also the original idea I had, but I got sidetracked somehow.

This still requires `msg.decodeJson(decoder, implicitly)`, but I guess we need to do that.

aeons: If the constraint on `F` is moved to the class, you can skip the implicitly. Will finish this up and create a new PR.

rossabaker: :+1:

aeons:
See #1396

Issue #1349, opened 2017-08-25
Blaze client: Time to first byte timeout


scottcarey:
A time to first byte (TTFB) timeout in the BlazeClient would be very useful.  In my case, I have requests that are naturally of very long duration, streaming large chunks of data, but the server might be unresponsive.   A very long timeout is needed for the total request time, but a short timeout to first byte is desired.

I don't care if it is time to the first byte of the body, or time to the headers being returned, but I can see how both could be useful to others.

rossabaker:
#1350 is a server change that flushes the header as soon as it's available.

The idle timeout on the client is reset by both reads and writes.  A TTFB timeout would be trivial if writes didn't cancel the pending idle timeout.  We'll need some additional state somewhere to track the first read.

rossabaker:
Should a first-byte timeout ever exceed the idle timeout, or only more tightly constrain it?  This could impact the right design.

scottcarey:
I'm not sure.   When I think about an HTTP request, I come up with 4 phases of the request that could potentially all have independent timeouts (though that would be overkill):

1.  Establishing the TCP connection or obtaining it from a pool.
2.  Issuing the request and waiting for HTTP response headers.
3.  Waiting for the first byte in the body (if the result has a body).
4.  Waiting for the request to complete.

For my purposes, a timeout for either 2 or 3 independent of 4 is what I need. It could encompass #1 as well.   I need to differentiate between a slow network (delays in the body) from a slow server (delays to first byte or headers).  I would probably configure a 5 second TTFB timeout and 30 second data transfer idle timeout.

The most flexible thing is to have the two timeouts be independent, one applying before the first server response packet (headers, #2 above), the other applying after.

This can be viewed as the same idle timeout with different restrictions in different phases, or as two independent timeouts.

One might have a use case where a TTFB timeout is longer than the idle timeout for the duration of the response.  Perhaps someone is working with busy or eagerly queuing servers on a fast network (like a computing cluster), so long initial delays are expected but once established fast transfers are expected.  This is the opposite of my use case -- servers should generally reply quickly unless unhealthy (we return 'too busy' aggressively, so long delays to first byte indicate something is wrong), but the requests can reach across the globe over slow networks.

I don't have a strong opinion on which way to go, especially as I don't know if there is a large difference in implementation complexity.  My gut says that all else being equal, two independent timeouts that do not constrain each other is a bit better.

rossabaker:
[...]

PR #1352, opened 2017-08-29
Implement a response header timeout

rossabaker:
Adds a timeout between the completion of the request and the completion of the response header.

Addresses #1349.  The original request was for time-to-first-byte, but I think this satisfies @scottcarey's use case, and I've seen this in other APIs.

### DIFF
--- a/blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientConfig.scala
+++ b/blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientConfig.scala
@@ -11,8 +11,14 @@ import scala.concurrent.duration.Duration
 
 /** Config object for the blaze clients
   *
-  * @param idleTimeout duration that a connection can wait without traffic before timeout
-  * @param requestTimeout maximum duration for a request to complete before a timeout
+  * @param responseHeaderTimeout duration between the completion of a request
+  * and the completion of the response header.  Does not include time
+  * to acquire the connection or the time to read the response.
+  * @param idleTimeout duration that a connection can wait without
+  * traffic being read or written before timeout
+  * @param requestTimeout maximum duration for a request to complete
+  * before a timeout.  Does not include time to acquire the the
+  * connection, but does include time to read the response
   * @param userAgent optional custom user agent header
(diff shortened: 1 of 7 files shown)


rossabaker:
Time-to-first-byte would sit better with #1350.  Alternatively, #1350 could concede to `Transfer-Encoding: chunked\r\n\r\n` when the header is flushed.  The client and the server are separate, but it would be nice if the timeout setting and the server chunking worked toward the same goal in the same product.

blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientConfig.scala
```diff
@@ -11,8 +11,14 @@ import scala.concurrent.duration.Duration
 
 /** Config object for the blaze clients
   *
-  * @param idleTimeout duration that a connection can wait without traffic before timeout
-  * @param requestTimeout maximum duration for a request to complete before a timeout
+  * @param responseHeaderTimeout duration between the completion of a request
+  * and the completion of the response header.  Does not include time
```
bryce-anderson: Response _prelude_ seems like better wording.

rossabaker: I like prelude, but I think we're the only ones who have ever used that term.

rossabaker:
🤔 You're proposing that the Http1ClientStage schedules the timeout and sends the error, thus removing those messages?  How would we cancel that timeout if another timeout has fired?  That's why I wanted to keep all the timeout state in the timeout stage, though it did make me feel bad.

bryce-anderson:
Not at a computer, but I believe the result of a timeout I'd to send an Error to the head cancelling all ops with that error and shutting things down. It might echo it back up the pipeline, but the initial part should be  idempotent so multiple timeouts should be fine with a little work.

rossabaker:
:+1: I think the solution we have works, and is in private classes.  I say we get the releases shipped with this and clean it up more on master, where we won't have to do it thrice.

bryce-anderson:
👍  agreed.

PR #1354, opened 2017-08-29
Improve DSL

aeons:
With a parameterized effect type, the old dsl was not really useful. As an interim solution, `org.http4s.dsl` was fixed to using `cats.effect.IO` as the effect type.

This PR improves on this situation. It still keeps the `IO`-dsl around, so if you are fine with fixing your effect type to `IO`, you can just `import org.http4s.dsl.io._`

If you want a parametric effect type, you will have two options, either

- extend `Http4sDsl[F[_]]` in your service, or
- create a value of `Http4sDsl[F[_]]` and import all members from it

Some examples:

```scala
import org.http4s._

class MyExtendingService[F[_]: Sync] extends Http4sDsl[F] {
  val service = HttpService[F] {
      Ok(s"Hello, $name")
  }
}
```

```scala
import org.http4s._

class MyImportingService[F[_]: Sync] {
  val dsl = Http4sDsl[F]
  import dsl._

  val service = HttpService[F] {
      Ok(s"Hello, $name")
  }
}
```

### DIFF
--- a/dsl/src/main/scala/org/http4s/dsl/impl/Methods.scala
+++ b/dsl/src/main/scala/org/http4s/dsl/impl/Methods.scala
@@ -1,4 +1,4 @@
-package org.http4s.dsl
+package org.http4s.dsl.impl
 
 import org.http4s.Method
 
--- a/dsl/src/main/scala/org/http4s/dsl/io/package.scala
+++ b/dsl/src/main/scala/org/http4s/dsl/io/package.scala
@@ -1,0 +1,5 @@
+package org.http4s.dsl
+
+import cats.effect.IO
+
+package object io extends Http4sDsl[IO]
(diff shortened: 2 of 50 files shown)


dsl/src/main/scala/org/http4s/dsl/impl/Methods.scala
```diff
@@ -1,4 +1,4 @@
-package org.http4s.dsl
+package org.http4s.dsl.impl
```
rossabaker: Wow, I guess I'd forgotten these were here.

dsl/src/main/scala/org/http4s/dsl/io/package.scala
```diff
@@ -0,0 +1,5 @@
+package org.http4s.dsl
+
+import cats.effect.IO
+
+package object io extends Http4sDsl[IO]
```
rossabaker: Does this need to be a package object or an `object io` in `dsl`?  Package objects are hard to deprecate.  Though I guess it has a package-like feel.

aeons: That wouldn't be a problem. It was a package object before, that's the only reason.

aeons:
That sounds like a good idea.

PR #1356, opened 2017-08-30
Sync parboiled to d7ad77a5077

rossabaker:
Decided to see if it had changed much. It hasn't, but it's easier to stay current.

### DIFF
--- a/parboiled2/src/main/scala/org/http4s/internal/parboiled2/Parser.scala
+++ b/parboiled2/src/main/scala/org/http4s/internal/parboiled2/Parser.scala
@@ -18,6 +18,7 @@ package org.http4s.internal.parboiled2
 
 import scala.annotation.tailrec
 import scala.collection.immutable.VectorBuilder
+import scala.collection.mutable
 import scala.language.experimental.macros
 import scala.util.{ Failure, Success, Try }
 import scala.util.control.{ NonFatal, NoStackTrace }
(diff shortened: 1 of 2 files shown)


rossabaker:
This PR isn't very important, but gives us a chance to discuss bigger matters.

We have problems running publishLocal since we inhaled parboiled2.  We could solve that by publishing this as an actual module, but it looks like another dependency to those who count dependencies by number instead of by burden.  It's a good solution but bad marketing.

Because this almost never changes, we could also consider republishing this as a hard fork in a separate repo.  That would save us recompiling every time.

rossabaker:
Can someone describe the `publishLocal` problem?  I've heard a couple complaints, but not seen it.  I just ran `sbt publishLocal` on master, and it was fine.

aeons:
If you set version to something that's not published anywhere and run `sbt publishLocal`, it will complain that it can't find the `parboiled2` dependency in that version.

rossabaker:
Oh, it worked for me because there's already an 0.18.0-SNAPSHOT?  But that version of parboiled2 shouldn't be published anywhere. 🤔

ChristopherDavenport:
Current Master  
Change version.sbt -> 0.19.0-SNAPSHOT  
`sbt publishLocal`  
switch dependent project to rely on 0.19.0-SNAPSHOT

```
coursier.ResolutionException: Encountered 1 error(s) in dependency resolution:
    org.http4s:http4s-parboiled2_2.12:0.19.0-SNAPSHOT:
```

Don't forget to clear local cache afterwards

rossabaker:
Okay, so publishing works, but depending on it doesn't.  And we rewrite the POM, but Ivy pulls from its own metadata, which is why the dependency appears locally but not when resolved via Maven?

rossabaker:
I'm punting and making parboiled2 a module: #1382.

PR #1357, opened 2017-08-30
Implement a retry policy for client retry middleware

rossabaker:
/cc @dbousamra

### DIFF
--- a/client/src/main/scala/org/http4s/client/middleware/Retry.scala
+++ b/client/src/main/scala/org/http4s/client/middleware/Retry.scala
@@ -19,15 +19,15 @@ object Retry {
     def prepareLoop(req: Request, attempts: Int): Task[DisposableResponse] = {
       client.open(req).attempt flatMap {
         case \/-(dr) =>
-          policy(req, Right(dr.response), attempts) match {
+          policy(req, \/-(dr.response), attempts) match {
             case Some(duration) =>
               logger.info(s"Request ${req} has failed on attempt #${attempts} with reason ${dr.response.status}. Retrying after ${duration}.")
               dr.dispose.flatMap(_ => nextAttempt(req, attempts, duration))
             case None =>
               Task.now(dr)
           }
-        case -\/(e) =>
-          policy(req, Left(e), attempts) match {
+        case left @ -\/(e) =>
+          policy(req, left, attempts) match {
             case Some(duration) =>
               logger.error(e)(s"Request ${req} threw an exception on attempt #${attempts} attempts. Retrying after ${duration}.")
               nextAttempt(req, attempts, duration)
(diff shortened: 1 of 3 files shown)


client/src/main/scala/org/http4s/client/middleware/Retry.scala
```diff
@@ -61,8 +47,47 @@ object Retry {
...
+    * duration after which the request will be retried.  See
+    * `exponentialBackoff` for a useful implementation.
+    * 
+    * @param retriable determines whether the request is retriable
+    * from the request and either the throwable or response that was
+    * returned.  Defaults to `defaultRetriable`.
+    */
+  def apply(
+    backoff: Int => Option[FiniteDuration],
+    retriable: (Request, Either[Throwable, Response]) => Boolean = defaultRetriable
```
rossabaker: I started this on release-0.17.x  This should probably have been a `\/`, even though we'll have to change it right back.

client/src/main/scala/org/http4s/client/middleware/Retry.scala
```diff
@@ -61,8 +47,47 @@ object Retry {
...
+    // TODO Leaving PayloadTooLarge out until we model Retry-After
+    InternalServerError,
+    ServiceUnavailable,
+    BadGateway,
+    GatewayTimeout
+  )
+
+  /** Default logic for whether a request is retriable. */
+  def defaultRetriable(req: Request, result: Either[Throwable, Response]): Boolean = {
+    if (req.isIdempotent)
```
rossabaker: This method doesn't exist in fs2.  The best we can do is check that the request method is idempotent.   We need to decide whether to default to retrying all idempotent methods, or only methods known not to have a body.  We could conceivably check a content length of zero, but even that doesn't mean the body launches the nukes.

client/src/main/scala/org/http4s/client/middleware/Retry.scala
```diff
@@ -15,41 +15,27 @@ object Retry {
...
 
-  private[this] val RetriableStatuses = Set(
-    RequestTimeout,
-    // TODO Leaving PayloadTooLarge out until we model Retry-After
-    InternalServerError,
-    ServiceUnavailable,
-    BadGateway,
-    GatewayTimeout
-  )
+  type Policy = (Request, Either[Throwable, Response], Int) => Option[FiniteDuration]
```
rossabaker: Ugh, this is redundant.

rossabaker:
When we merge this to fs2:

* `\/` becomes `Either`
* There should be two retriable policies: one that checks the method is idempotent, and a second that checks the method is idempotent and forbids a body. The former is more useful in more cases, but only the latter is safe.

PR #1358, opened 2017-08-30
add Request URI to logging middleware …

bpholt:
… and copy logging middleware to client library


server/src/main/scala/org/http4s/server/middleware/Logger.scala
```diff
@@ -17,11 +17,12 @@ object Logger {
...
                            logHeaders: Boolean,
                            logBody: Boolean,
-                           redactHeadersWhen: CaseInsensitiveString => Boolean = Headers.SensitiveHeaders.contains
+                           redactHeadersWhen: CaseInsensitiveString => Boolean = Headers.SensitiveHeaders.contains,
+                           uriHeaderName: String = "uri"
                          )
                          (httpService: HttpService[F]): HttpService[F] =
-    ResponseLogger(logHeaders, logBody)(
-      RequestLogger(logHeaders, logBody)(
+    ResponseLogger(logHeaders, logBody, redactHeadersWhen, uriHeaderName)(
```
bpholt: I think not passing `redactHeadersWhen` here was a bug.

rossabaker: I think you're right.

bpholt:
@rossabaker Doesn't [`requestCopy`](https://github.com/http4s/http4s/blob/master/core/src/main/scala/org/http4s/Message.scala#L153) return a new instance of the `Request` with the specified changes made? So in this case `req.putHeaders(Header("uri", req.uri.toString()))` ends up calling `requestCopy` with a new `Headers` object, resulting in a new `Request` that shares a `body` with the original request. [That's what is passed to `Logger`](https://github.com/http4s/http4s/pull/1358/files#diff-eb10a9a582f0fc653a8a3e2f616c73d3R38); the [`Request` with unchanged headers is passed to `service`](https://github.com/http4s/http4s/pull/1358/files#diff-eb10a9a582f0fc653a8a3e2f616c73d3R42).

In `ResponseLogger`, it's similar, but it just uses the case class `copy` method internally. The [unchanged `Response` is returned from the `flatMap`](https://github.com/http4s/http4s/pull/1358/files#diff-8d48b95cb07b5444f253909f24b34efcR35) and the [modified `Response` with the extra header is passed to `Logger`](https://github.com/http4s/http4s/pull/1358/files#diff-8d48b95cb07b5444f253909f24b34efcR40).

(I just noticed `req.putHeaders(Header("uri", req.uri.toString()))` doesn't use the parameterized header name. I'll update to fix that.)

rossabaker:
You're right.  I made a bad assumption.

It still seems a little odd to me to log this inside `Headers()` when it's not actually a header.  The context of an attribute, or an extra parameter to logMessage, would help you here.  But I agree that it's safe.

bpholt:
FWIW I've encountered other HTTP libraries that log the entire first line of the request as a pseudo-header (so method, URI, and HTTP version as effectively a single header) but I can't think of where that was and it's proving hard to find. It's weirder for responses, I agree.

It sounds like @ChristopherDavenport is going to try to move this into a shared location for client/server to remove the duplication, so I'll wait for that before proceeding further.

rossabaker:
Doing a standard [NCSA request log format](https://en.wikipedia.org/wiki/Common_Log_Format) is on my agenda at some point.  Ultimately, we need a nice way of customizing this, with or without headers and body.

rossabaker:
I'm cleaning up the tags on PRs from my push -f accident.

The common logging format is happening in #1721.  We still don't have the client logging middleware.  Do we want to reopen this so we don't lose it, or kill it and start over when the dust settles?

bpholt:
I'd still like to see client logging middleware eventually, but this PR probably isn't helpful at this point. I can open a new issue later this week if you want to track the client logging separately from #1721.

PR #1359, opened 2017-08-30
 Introduce scalafmt

rossabaker:


rossabaker:
Presented now for discussion.  Doesn't integrate with CI yet.  This is everybody's chance to paint the style bikeshed, because we're only doing this once.  This commit will be force-pushed before merge.

ChristopherDavenport:
Very smooth that .scalafmt.conf is first.

cquiroz:
Will the build automatically format the files?
Do you think you'll force push into master?

rossabaker:
1. We could add `scalafmtTest` to CI to confirm to fail builds on improperly formatted source.  I think this is the most common way of doing this, though it might be discouraging to new contributors.
2. We could set `scalafmtOnCompile in ThisBuild := true` to automatically run it.  Or we could add it to a `scalafmt` to a `validate` alias and encourage people to run that.  I'd rather transition to `validate` (make people run tut, too!), but getting word out to new contributors is again a problem.
3. We could run `scalafmt` on Travis and make a commit into PRs.  That's automated, but icky.

cquiroz:
I'm not sure about 1 or 2 but 3 doesn't seem very good, automated commits could be quite confusing

I guess a check on Travis and the update the documentation would be enough

aeons:
How about 1 and 2.

2 makes it harder to forget formatting, and 1 catches the few times someone does forget.

rossabaker:
This is irritating.  We got a couple dozen of these:

```
Warning: Unknown Scalafmt version 1.2.0; using 1.0 interface
```

rossabaker:
It looks like `scalafmtOnCompile` is smart enough to process only changed files.  This makes me feel better about that solution.

aeons:
BTW, if you guys don't have any objections to longer lines, setting `maxColumn` to something like 120 might make sense.

Scala isn't the most elegant language at 80 columns.

(and based on the current code base, it seems we don't have any issues with long lines ;) )

cquiroz:
I agree on the longer lines

ChristopherDavenport:
I believe someone said something about 117 or something like that being the max that display on github without folding. I can't remember when or who however so don't want to swear by it.

rossabaker:
Heh, @ChristopherDavenport and I had I are appealing to the same authority.  I'm old and grumpy and like 80 characters, but GitHub browsing is the new 80x25 tty, I guess.  On my 15" Mac, Github maxes out at 139 columns in view.  The diff defaults to 95 at full screen, but expands indefinitely, even past 139.

I tend to prefer narrower, especially for block comments, but I'd yield toward 100 or 120 if that's what everybody likes.  I'm going to fight vertical alignment much harder than I'll fight line length. :)

aeons:
https://github.com/http4s/http4s/blob/master/jetty/src/main/scala/org/http4s/server/jetty/JettyBuilder.scala#L53
https://github.com/http4s/http4s/blob/master/circe/src/main/scala/org/http4s/circe/CirceInstances.scala#L75
https://github.com/http4s/http4s/blob/master/core/src/main/scala/org/http4s/HttpVersion.scala#L35

These were just from a very quick glance through random files :)

rossabaker:
I bumped to 100 columns.  Look at 459401d to see the difference.

Never mind the merge conflicts, as a final run is yet to come.

rossabaker:
I would really like to clean up our imports too, but this is out of scope in scalafmt and I can't find an automated tool.  It's a hard problem, because it's not purely lexicographical in Scala.  IntelliJ has the best I've found, but the best we could do is run it once there and be more picky in code reviews.

cquiroz:
Clean them up in what sense? You can certainly make it fail if you have unused imports

aeons:
Adding some of tpolecat's recommended scalac flags could be nice as well I guess. Like unused imports for example.

http://tpolecat.github.io/2017/04/25/scalac-flags.html

rossabaker:
* Unused imports, which the flags would handle.  And which scalafix can do.
* Organized imports, which I don't think anything but IntelliJ does well at this point.

rossabaker:
e7f935d should fail loudly because scalafmt has not been run.

rossabaker:
Why we're formatting *.sbt and not *.scala:

* I can't seem to make it not format our *.sbt files.  The exclude filter in scalafmt is ignored, and `scalafmtOnCompile in Sbt := false` only works from with an *.sbt, and needs to be repeated in each *.sbt.

* To make it format our *.scala files, we would need to put the plugin in project/project/plugins.sbt and configure it in project/plugins.sbt.  That's a rabbit hole I'm not willing to go down.

rossabaker:
After the final merges, I'm going to rebase the first commit and rerun the second.  If you want to tweak anything, speak tomorrow.

LeifW:
We used Scalariform at my work (because I thought it happened to look like my personal style), and set the equivalent of `scalafmtOnCompile in ThisBuild := true`. Sometimes people just run everything from IntelliJ, and that hook doesn't get run. So I added a "run git diff and fail the build if there's unstaged changes" to the CI. But the `git` command wasn't installed on the build boxes. So I wrote an sbt plugin to do that via JGit. Turns out the sbt-scalariform plugin can also export that info for you I think (which files it reformatted).

rossabaker:
The scalafmt plugin has a `test` task that supports the enforcement, like the git workaround for scalariform.

rossabaker:
This is what I intend to merge unless someone complains.

cquiroz:
I'm a bit lost with the comment that we'll loose history. Why's that?

rossabaker:
@cquiroz The history will still be there, but now when running git blame, 7500 lines or so will say it changed because of scalafmt, and we'll need to jump back one commit to see why things changed.

cquiroz:
:+1:

PR #1360, opened 2017-08-30
honor Retry-After header

eklavya:


client/src/main/scala/org/http4s/client/middleware/Retry.scala
```diff
@@ -56,9 +60,15 @@ object Retry {
       }
 
     def nextAttempt(req: Request[F], attempts: Int, duration: FiniteDuration)
-                   (implicit F: Async[F], executionContext: ExecutionContext): F[DisposableResponse[F]] =
-      // TODO honor Retry-After header
-      scheduler.sleep_[F](duration).run >> prepareLoop(req.withEmptyBody, attempts + 1)
+                   (implicit F: Async[F], executionContext: ExecutionContext): F[DisposableResponse[F]] = {
+      val sleepDuration = req.headers.get(`Retry-After`).map { h =>
```
rossabaker: `Retry-After` is a response header.  We'll need to get that up in the case on line 39.

client/src/main/scala/org/http4s/client/middleware/Retry.scala
```diff
@@ -56,9 +60,15 @@ object Retry {
...
+        h.retry match {
+          case Left(d) => Instant.now().until(d.toInstant, ChronoUnit.SECONDS).seconds
+          case Right(secs) => secs.seconds
+        }
```
rossabaker: I think I would take the max of this value and the passed in `duration` here.  I think `Retry-After` means more "wait a minimum of this time" rather than "come back at exactly this time."

client/src/main/scala/org/http4s/client/middleware/Retry.scala
eklavya: OOPS 👅

client/src/main/scala/org/http4s/client/middleware/Retry.scala
eklavya: makes sense

rossabaker:
I will backport this to 0.17, since that middleware is already touched in #1357.

PR #1361, opened 2017-08-31
Retry enhancements

rossabaker:


client/src/main/scala/org/http4s/client/middleware/Retry.scala
```diff
@@ -2,71 +2,141 @@ package org.http4s
...
+    else false
+  }
+
+  /** Returns true if the request method is idempotent and the result is
+    * either a throwable or has one of the `RetriableStatuses`.  This is
+    * the `defaultRetriable` behavior from 0.16.
+    * 
+    * Caution: if the request body is effectful, the effects will be
+    * run twice.  The most common symptom of this will be resubmitting
+    * an empty request body.
```
rossabaker: This is a really horrid thing to do, but some people will want this.  An alternative is to buffer request bodies, which would slow things down.

client/src/main/scala/org/http4s/client/middleware/Retry.scala
```diff
@@ -2,71 +2,141 @@ package org.http4s
...
+  /** Like [[unsafeRetriable]], but returns true even if the request method
+    * is not idempotent.  This is useful if failed requests are assumed to
+    * have not reached their destination, which is a dangerous assumption.
+    * Use at your own risk.
+    * 
+    * Caution: if the request body is effectful, the effects will be
+    * run twice.  The most common symptom of this will be resubmitting
+    * an empty request body.
+    */
+  def recklesslyRetriable(req: Request, result: Either[Throwable, Response]): Boolean =
```
ChristopherDavenport: Best Ever!

client/src/test/scala/org/http4s/client/middleware/RetrySpec.scala
```diff
@@ -21,24 +21,25 @@ class RetrySpec extends Http4sSpec with Tables {
   }
 
   val defaultClient = Client.fromHttpService(route)
+  val failClient = Client(Service.const(Task.fail(new Exception("boom"))), Task.now(()))
```
ChristopherDavenport: Are we using the failClient anywhere that required it's removal?

client/src/main/scala/org/http4s/client/middleware/Retry.scala
```diff
@@ -2,71 +2,141 @@ package org.http4s
...
       }
     }
 
-    def nextAttempt(req: Request, attempts: Int, duration: FiniteDuration): Task[DisposableResponse] = {
-      // TODO honor Retry-After header
-      prepareLoop(req.withEmptyBody, attempts + 1).schedule(duration)
+    def nextAttempt(req: Request, attempts: Int, duration: FiniteDuration, retryHeader: Option[`Retry-After`]): Task[DisposableResponse] = {
+      val headerDuration = retryHeader.map { h =>
+        h.retry match {
+          case Left(d) => Instant.now().until(d.toInstant, ChronoUnit.SECONDS)
```
ChristopherDavenport: This is the decreasing time location?

client/src/test/scala/org/http4s/client/middleware/RetrySpec.scala
rossabaker: I think I started another block for more tests and moved it and then abandoned that block.

client/src/main/scala/org/http4s/client/middleware/Retry.scala
rossabaker: Yeah, if a Retry-After is a Left, we need to calculate the delta from now until then.  The Right is expressed as a wait time in seconds.

PR #1366, opened 2017-09-01
Reorganize imports

rossabaker:
Propose we organize imports alphabetically, in a single group.  This matches the Twitter guidelines, which @bryce-anderson didn't speak ill of on Gitter last night.  It's simple for users of any editor and disinvites unproductive debates like "why is scala.* grouped as a standard library and cats.* isn't?"

Unlike scalafmt, we will _not_ get tool support for maintaining this, though we could conceivably provide an IntelliJ configuration.  This was mostly produced with IntelliJ, with a few hand tweaks to code that it broke.


examples/src/main/scala/com/example/http4s/ExampleService.scala
```diff
@@ -1,19 +1,18 @@
 package com.example.http4s
 
-import _root_.io.circe.Json
 import cats._
 import cats.effect.IO
 import cats.implicits._
 import fs2._
-import org.http4s.MediaType._
+import _root_.io.circe.Json
```
rossabaker: This is annoying.  If we import `fs2._` first, `io.circe` needs a root import.  But now it's no longer lexicographic.  And if we move it to to the top, it's no longer clear why it needs to be a _root_ import.

aeons: Ugh.

ChristopherDavenport: I just import circe as root always. io is common enough that is conflicts too often. Now we have an io package as well.

rossabaker:
I'm on a branch right now, but so I don't forget, @Daenyth suggests [how not to break the console](https://gitter.im/http4s/http4s?at=59b6e2c3cfeed2eb65f836dd).

rossabaker:
Not proud of 4621c2e.  It's what lets the same source run on 2.11 and 2.12 without any unused imports.

Alternatives would be:
- remove unused import checking in 2.12 (would frequently snare us in 2.11 on Travis)
- remove unused import checking altogether (dirty code)
- gratuitously use the import in affected files?
- extend Http4sSpec with the EitherOps since it's mostly tests?  Would that even work?

aeons:
I think option 4 would work and is a nice solution, if it's only tests that are affected.

aeons:
fs2 handles it like this: https://github.com/functional-streams-for-scala/fs2/blob/series/0.10/core/shared/src/main/scala/fs2/fs2.scala#L25-L40

rossabaker:
I like that fs2 workaround, if it doesn't cause conflicting implicits where it's actually needed.

ChristopherDavenport:
It shouldn't direct members are favored over implicit classes, so since it is right biased in 2.12 that shouldn't be an issue.

aeons:
It does cause a conflict with `catsSyntaxEither`. In fs2 they do `import cats.implicits.{catsSyntaxEither => _, _}`.

I'm seeing how that works atm.

aeons:
Using this approach looks like this: https://github.com/aeons/http4s/commit/2280586b3dbd1079662b5ba42cc11f18cb82c397

rossabaker:
I think 2280586 is going to suffer from a similar problem as just disabling unused import checking: misuses won't fail to compile until Travis tries it on 2.11.  But it will flag _other_ unused imports on 2.12, so it's probably still a better solution.

rossabaker:
There's a `===` somewhere in Specs2 that gets shadowed by the `===` from `Eq` syntax.  I fixed the test failures, but I haven't sorted out how to avoid this happening in the future.

PR #1367, opened 2017-09-01
Max connections per RequestKey in client connection pool

eklavya:


blaze-client/src/main/scala/org/http4s/client/blaze/PooledHttp1Client.scala
```diff
@@ -11,14 +11,20 @@ object PooledHttp1Client {
   /** Construct a new PooledHttp1Client
     *
     * @param maxTotalConnections maximum connections the client will have at any specific time
+    * @param maxConnectionsPerRequestKey Map of RequestKey to number of max connections
     * @param config blaze client configuration options
     */
   def apply[F[_]: Effect](
       maxTotalConnections: Int = DefaultMaxTotalConnections,
+      maxConnectionsPerRequestKey: Map[RequestKey, Int] = Map.empty,
```
rossabaker: Do you still prefer this to a `RequestKey => Int`?  I think I'm on Team Function.

client/src/main/scala/org/http4s/client/PoolManager.scala
```diff
@@ -20,13 +21,42 @@ private final class PoolManager[F[_], A <: Connection[F]](
...
-  private val waitQueue = new mutable.Queue[Waiting]
+  private var curTotal = 0
+  private var allocated = mutable.Map.empty[RequestKey, Int]
+  private val idleQueue = mutable.Map.empty[RequestKey, mutable.Queue[A]]
+  private val waitQueue = mutable.Queue.empty[Waiting]
 
   private def stats =
     s"allocated=$allocated idleQueue.size=${idleQueue.size} waitQueue.size=${waitQueue.size}"
 
+  @inline
```
rossabaker: I almost never use this annotation.  The JVM automates it better than I can guess it.  Do you have evidence this is helpful?

client/src/main/scala/org/http4s/client/PoolManager.scala
```diff
@@ -20,13 +21,42 @@ private final class PoolManager[F[_], A <: Connection[F]](
...
+  private def getMaxConnections(key: RequestKey): Int =
+    maxConnectionsPerRequestKey.getOrElse(key, maxTotal)
+
+  @inline
+  private def getConnectionFromQueue(key: RequestKey): Option[A] =
+    try {
+      Some(idleQueue(key).dequeue())
```
rossabaker: I'd probably rather do an empty check than catch exceptions.

client/src/main/scala/org/http4s/client/PoolManager.scala
```diff
@@ -85,25 +116,28 @@ private final class PoolManager[F[_], A <: Connection[F]](
...
                 createConnection(key, callback)
 
-              case None if idleQueue.nonEmpty =>
+              case None if idleQueue.get(key).exists(_.nonEmpty) =>
                 logger.debug(
                   s"No connections available for the desired key. Evicting oldest and creating a new connection: $stats")
-                allocated -= 1
-                idleQueue.dequeue().shutdown()
+                curTotal -= 1
+                decrConnection(key)
```
rossabaker: Oh, I meant for `curTotal` to be part of those helper functions.  Because you're always doing them together.

client/src/main/scala/org/http4s/client/PoolManager.scala
```diff
@@ -85,25 +116,28 @@ private final class PoolManager[F[_], A <: Connection[F]](
...
+                getConnectionFromQueue(key).get.shutdown()
```
rossabaker: Why are we sure this `.get` is safe?

blaze-client/src/main/scala/org/http4s/client/blaze/PooledHttp1Client.scala
eklavya: That will preclude people from being able to supply partial maps, which would be inconvenient in my opinion.

client/src/main/scala/org/http4s/client/PoolManager.scala
eklavya: Just a request for scala compiler to consider inlining, we can leave it on, no?

client/src/main/scala/org/http4s/client/PoolManager.scala
eklavya: Yeah, I think it's a silly optimisation (if it even was one).

client/src/main/scala/org/http4s/client/PoolManager.scala
eklavya: Because we only come to this case if the queue was non empty. The original code also had a ```dequeue``` there.

rossabaker:
This from 93a0327 is disheartening, but I think totally unrelated to this PR:

```
[info] Timeout Middleware should
[error]   x have no effect if the response is not delayed (97 ms)
[error]    the response status '500 Internal Server Error != 200 OK' (TimeoutSpec.scala:31)
[error] org.http4s.server.middleware.TimeoutSpec.$anonfun$new$2(TimeoutSpec.scala:31)
```

rossabaker:
I think this looks great, but we need some sort of testing strategy, even if it's a manual runthrough, to make sure this doesn't break everything.

client/src/main/scala/org/http4s/client/PoolManager.scala
```diff
@@ -20,13 +22,33 @@ private final class PoolManager[F[_], A <: Connection[F]](
...
+  private var curTotal = 0
+  private var allocated = mutable.Map.empty[RequestKey, Int]
+  private val idleQueue = mutable.Map.empty[RequestKey, mutable.Queue[A]]
+  private val waitQueue = mutable.Queue.empty[Waiting]
 
   private def stats =
     s"allocated=$allocated idleQueue.size=${idleQueue.size} waitQueue.size=${waitQueue.size}"
 
+  private def getConnectionFromQueue(key: RequestKey): Option[A] =
+    idleQueue.get(key).flatMap(q => if (q.nonEmpty) Some(q.dequeue()) else None)
```
bryce-anderson: We should also remove the `Queue` from `idleQueue` if it's empty or else we have a memory leak if folks are using lots of different hosts.

client/src/main/scala/org/http4s/client/PoolManager.scala
```diff
@@ -85,30 +107,37 @@ private final class PoolManager[F[_], A <: Connection[F]](
...
 
               case Some(closedConn) =>
                 logger.debug(s"Evicting closed connection: $stats")
-                allocated -= 1
+                decrConnection(key)
                 go()
 
-              case None if allocated < maxTotal =>
+              case None
+                  if curTotal < maxTotal && allocated.getOrElse(key, 0) < maxConnectionsPerRequestKey(
```
bryce-anderson: This is the same check done in `createConnection`: do we need to do it twice?

client/src/main/scala/org/http4s/client/PoolManager.scala
```diff
@@ -85,30 +107,37 @@ private final class PoolManager[F[_], A <: Connection[F]](
...
+                    key) =>
                 logger.debug(s"Active connection not found. Creating new one. $stats")
                 createConnection(key, callback)
 
-              case None if idleQueue.nonEmpty =>
+              case None if idleQueue.get(key).exists(_.nonEmpty) =>
```
bryce-anderson: Why do we prefer to open a new connection to taking one from the `idleQueue`?

client/src/main/scala/org/http4s/client/PoolManager.scala
```diff
@@ -85,30 +107,37 @@ private final class PoolManager[F[_], A <: Connection[F]](
...
                 createConnection(key, callback)
 
               case None => // we're full up. Add to waiting queue.
                 logger.debug(s"No connections available.  Waiting on new connection: $stats")
-                waitQueue.enqueue(Waiting(key, callback))
+                if (waitQueue.length <= maxWaitQueueLimit) {
+                  waitQueue.enqueue(Waiting(key, callback))
+                } else {
+                  logger.error(s"Max wait length reached, not scheduling.")
+                  callback(Left(new Exception("Wait queue is full")))
```
bryce-anderson: Should add the `maxWaitQueueLimit` to the message to aid in debugging.

client/src/main/scala/org/http4s/client/PoolManager.scala
```diff
@@ -150,17 +179,19 @@ private final class PoolManager[F[_], A <: Connection[F]](
...
 
             // returned connection didn't match any pending request: kill it and start a new one for a queued request
             case None =>
               connection.shutdown()
-              allocated -= 1
-              val Waiting(key, callback) = waitQueue.dequeue()
-              createConnection(key, callback)
+              decrConnection(key)
+              val Waiting(k, callback) = waitQueue.dequeue()
+              createConnection(k, callback)
```
bryce-anderson: This could end up breaking the `maxConnectionsPerRequestKey` invariant since a borrow request could have ended up in `waitQueue` due to exceeding the per-key connection limit, not the `maxTotal` limit.

client/src/main/scala/org/http4s/client/PoolManager.scala
eklavya: We do that when the last connection to a host is dropped.

client/src/main/scala/org/http4s/client/PoolManager.scala
eklavya: I think this is to differentiate between case 3 and 5.

client/src/main/scala/org/http4s/client/PoolManager.scala
eklavya: Not sure, kept the same logic as the previous code here.

client/src/main/scala/org/http4s/client/PoolManager.scala
eklavya: Ok

client/src/main/scala/org/http4s/client/PoolManager.scala
eklavya: We check again in createConnection for that, right?

bryce-anderson: Yes, you're right: we won't break the invariant. If the head could be blocked on its `maxConnectionsPerRequestKey` so it will be put back at the rear of the queue while a different key may be able to open a connection.

client/src/main/scala/org/http4s/client/PoolManager.scala
bryce-anderson: Good point. Care to factor the predicate out into a method? That might clean things up a little.

client/src/main/scala/org/http4s/client/PoolManager.scala
bryce-anderson: As is, we never drop keys from `idleQueue`, only `allocated`. It might be a good idea to rename `idleQueue` to `idleQueues` since there are now more than one.

client/src/main/scala/org/http4s/client/PoolManager.scala
```diff
@@ -20,13 +22,33 @@ private final class PoolManager[F[_], A <: Connection[F]](
   private[this] val logger = getLogger(classOf[PoolManager[F, A]])
 
   private var isClosed = false
-  private var allocated = 0
-  private val idleQueue = new mutable.Queue[A]
-  private val waitQueue = new mutable.Queue[Waiting]
+  private var curTotal = 0
+  private var allocated = mutable.Map.empty[RequestKey, Int]
```
bryce-anderson: Could we make `allocated` a `val` and use `.clear()` in shutdown?

client/src/main/scala/org/http4s/client/PoolManager.scala
bryce-anderson: I just double checked the code of `createConnection` and I believe my original comment stands: it will double checks the invariant but as its written its not intended to handle that situation: it just screams very loudly about it and closes the pending dispatch with an error.

eklavya: So are you saying the invariant is broken? I still don't see how :(

bryce-anderson: Lets say `maxConnectionsPerRequestKey` always returns 2.

1. We make one request to service A, which does not complete yet. 
1. We then make three requests to service B, two make new connections and one gets queued in `waitQueue`. The active requests do not complete yet.
1. The request to service A finishes, and finds its way to this branch, and ends up breaking the invariant since it attempts to make a new connection to service B when there are already two outstanding connections.

I believe this problem also affects the non-recycleable branch of the `release` method.

eklavya: > The request to service A finishes, and finds its way to this branch, and ends up breaking the invariant since it attempts to make a third connection to service B when there are already two outstanding connections.

But it will fail, right? CreateConnection will successfully check the condition and complete with failure.

bryce-anderson: Yes, it will fail, but that isn't desired behavior: we should leave the third request for B in the queue until one of the first two request to B finishes. The failure pathway of `createConnection` is best described as a verbose assert: its not intended to be exercised, only to let everyone know that things aren't working as we thought. Thats why the error message is so agressive:
```scala
s"Invariant broken in ${this.getClass.getSimpleName}! Tried to create more connections than allowed: $stats"
```

eklavya: Ok, I thought you were saying a new connection will be made breaking the max connections invariant.

eklavya: I removed that entire case, it wouldn't be reached in any condition.

bryce-anderson: I think that specific problem is still there and we have another interesting problem: it's possible for other keys to be holding idle connections summing to `maxTotal` and now we won't close one of those for a different key. That used to be handled on what was line 102.

eklavya: Which idleQueue do we dequeue from? There is no total ordering anymore. We could randomly select a queue, would that be good enough?

bryce-anderson: Good question. I see a few options.
1. Do what you just suggested and just close a random one
1. Keep all idle connections in the same queue
1. Use a more complex data structures like LinkedHashMap

Hard to say which is ideal, and it likely depends on the circumstances. I'd probably start with 2 myself.

eklavya: LinkedHashMap would not be the same semantically, since it will order by insertion of keys (hostNames) and not the actual connections. 

I was playing with the idea of a Map pointing into a single queue but I thought that would be overkill and "choose random, dequeue" would be a better solution. I thing random dequeue is fine, what do you think?

bryce-anderson: Sorry, `LinkedHashMap` was a knee jerk reaction to assuming that java didn't have a `LinkedHashSet`: basically yet-another-collection to store the order of idleness. I think closing a random idle connection is probably fine until proven otherwise and simpler to implement although tougher to test.

eklavya:
ping @bryce-anderson @rossabaker

client/src/main/scala/org/http4s/client/PoolManager.scala
```diff
@@ -21,41 +23,36 @@ private final class PoolManager[F[_], A <: Connection[F]](
...
   private var curTotal = 0
-  private var allocated = mutable.Map.empty[RequestKey, Int]
-  private val idleQueue = mutable.Map.empty[RequestKey, mutable.Queue[A]]
+  private val allocated = mutable.Map.empty[RequestKey, Int]
+  private val idleQueues = mutable.Map.empty[RequestKey, mutable.Queue[A]]
   private val waitQueue = mutable.Queue.empty[Waiting]
 
   private def stats =
-    s"allocated=$allocated idleQueue.size=${idleQueue.size} waitQueue.size=${waitQueue.size}"
+    s"allocated=$allocated idleQueues.size=${idleQueues.size} waitQueue.size=${waitQueue.size} maxWaitQueueLimit=$maxWaitQueueLimit"
```
rossabaker: Is this going to be an untenable toString, rendering an arbitrarily large map for `allocated`?

client/src/main/scala/org/http4s/client/PoolManager.scala
```diff
@@ -81,11 +77,22 @@ private final class PoolManager[F[_], A <: Connection[F]](
...
-      callback(Left(error))
+      addToWaitQueue(key, callback)
+    }
+
+  private def getNonEmptyKeys: List[RequestKey] =
+    idleQueues.foldLeft(List.empty[RequestKey]) {
+      case (l, (k, q)) =>
+        if (q.nonEmpty) l :+ k
+        else l
+    }
```
rossabaker: Wait, aren't they all non-empty?

client/src/main/scala/org/http4s/client/PoolManager.scala
```diff
@@ -122,26 +129,29 @@ private final class PoolManager[F[_], A <: Connection[F]](
...
                 logger.debug(
                   s"No connections available for the desired key. Evicting oldest and creating a new connection: $stats")
-                curTotal -= 1
-                decrConnection(key)
-                getConnectionFromQueue(key).get.shutdown()
-                createConnection(key, callback)
+                val nonEmptyKeys = getNonEmptyKeys
+                if (nonEmptyKeys.nonEmpty) {
+                  val randKey = nonEmptyKeys.drop(Random.nextInt(nonEmptyKeys.size)).head
+                  idleQueues(randKey).dequeue().shutdown()
```
rossabaker: How is this idleQueue getting cleaned up if it's the last for the key?  When the connection gets shut down and we call invalidate?

I think a List is not ideal for randomly jumping into the middle, even though it's fast to construct.  Maybe a Vector would be a better choice?

client/src/main/scala/org/http4s/client/PoolManager.scala
eklavya: That's what I thought, and I was yelled at by the test which showed a scenario where this might not be the case.

client/src/main/scala/org/http4s/client/PoolManager.scala
eklavya: Yeah, should we make a public method Key -> Stat for users?

client/src/main/scala/org/http4s/client/PoolManager.scala
eklavya: Oh yeah, the keys that we get from the map are not in a vector. Optimizing the nonEmpty key collection I totally forgot that I am making a new collection so that could be anything ! 👅

client/src/main/scala/org/http4s/client/PoolManager.scala
eklavya: Wait, I forgot how this made sense !!

eklavya: Yeah, if none of the connections are idle because of slow requests, we would get a situation where maxTotal has been reached but idleQueues are empty.

client/src/main/scala/org/http4s/client/PoolManager.scala
eklavya: For the cleanup, we are calling decrConnection.

client/src/main/scala/org/http4s/client/PoolManager.scala
rossabaker: Yes, we need to put some thought into client metrics in general.  It has poor observability at the moment.

client/src/main/scala/org/http4s/client/PoolManager.scala
rossabaker: But don't we remove the idleQueue from the map when we dequeue the last idle connection?

eklavya: Yes, but there were no idle connections to begin with, so there was no entry.

client/src/main/scala/org/http4s/client/PoolManager.scala
```diff
@@ -84,30 +120,36 @@ private final class PoolManager[F[_], A <: Connection[F]](
...
-              case None if idleQueue.nonEmpty =>
+              case None if curTotal == maxTotal =>
                 logger.debug(
                   s"No connections available for the desired key. Evicting oldest and creating a new connection: $stats")
-                allocated -= 1
-                idleQueue.dequeue().shutdown()
-                createConnection(key, callback)
+                val keys = idleQueues.keys
+                if (keys.nonEmpty) {
+                  val randKey = keys.drop(Random.nextInt(keys.size)).head
```
bryce-anderson: How about using keys the iterator so we don't create an intermediate collection:
```scala
val randKey = keys.iterator.drop(Random.nextInt(keys.size)).next()
...
```

client/src/main/scala/org/http4s/client/PoolManager.scala
```diff
@@ -84,30 +120,36 @@ private final class PoolManager[F[_], A <: Connection[F]](
...
                 go()
 
-              case None if allocated < maxTotal =>
+              case None if numConnectionsCheckHolds(key) =>
                 logger.debug(s"Active connection not found. Creating new one. $stats")
                 createConnection(key, callback)
 
-              case None if idleQueue.nonEmpty =>
+              case None if curTotal == maxTotal =>
                 logger.debug(
```
bryce-anderson: We should move the log message into the two respective branches since it would be nice to know if we ended up killing a random idle connection or added it to the wait queue.

client/src/main/scala/org/http4s/client/PoolManager.scala
```diff
@@ -84,30 +120,36 @@ private final class PoolManager[F[_], A <: Connection[F]](
...
                   s"No connections available for the desired key. Evicting oldest and creating a new connection: $stats")
-                allocated -= 1
-                idleQueue.dequeue().shutdown()
-                createConnection(key, callback)
+                val keys = idleQueues.keys
+                if (keys.nonEmpty) {
+                  val randKey = keys.drop(Random.nextInt(keys.size)).head
+                  idleQueues(randKey).dequeue().shutdown()
```
bryce-anderson: As I understand this, we can have an empty `idleQueue` if there is a connection outstanding for its key, and we only remove the queue if the last key belonging to it is removed. That would mean that the `.dequeue()` call could throw a `NoSuchElementException` and we should probably go with `getConnectionFromQueue`.

In this case, should we be now randomly looking at a different queue? I feel like this scheme will get pretty ugly without just guaranteeing that there are no empty queues in the `idleQueues` map.

client/src/main/scala/org/http4s/client/PoolManager.scala
```diff
@@ -84,30 +120,36 @@ private final class PoolManager[F[_], A <: Connection[F]](
...
+                decrConnection(key)
                 go()
 
-              case None if allocated < maxTotal =>
+              case None if numConnectionsCheckHolds(key) =>
                 logger.debug(s"Active connection not found. Creating new one. $stats")
                 createConnection(key, callback)
 
-              case None if idleQueue.nonEmpty =>
+              case None if curTotal == maxTotal =>
```
bryce-anderson: I think we need to check that `maxConnectionsPerRequestKey(key)` didn't also fail since in that case it would break our invariant by making yet another one.

client/src/main/scala/org/http4s/client/PoolManager.scala
bryce-anderson: I believe this is still broken: we need to skim the waitQueue for a waiting request that is not waiting on `maxConnectionsPerRequestKey`.

client/src/main/scala/org/http4s/client/PoolManager.scala
```diff
@@ -171,14 +215,16 @@ private final class PoolManager[F[_], A <: Connection[F]](
               s"Connection returned could not be recycled, new connection needed: $stats")
             val Waiting(key, callback) = waitQueue.dequeue()
```
bryce-anderson: Same problem as line 202: we don't know that the first waiter in the queue is only blocked on `maxTotal`.

client/src/main/scala/org/http4s/client/PoolManager.scala
eklavya: New connections are only made in createConnection and we check for constraints there. How will be break the invariant?

client/src/main/scala/org/http4s/client/PoolManager.scala
eklavya: Ok

client/src/main/scala/org/http4s/client/PoolManager.scala
eklavya: Where is the intermediate collection?

client/src/main/scala/org/http4s/client/PoolManager.scala
eklavya: Sorry, I don't see how we can have an empty queue in idleQueues 👅 . Either the entire map would be empty or it will only have non empty queues.

client/src/main/scala/org/http4s/client/PoolManager.scala
eklavya: It will just be queued back if limit allows, right? We can just keep on dequeuing until something succeeds, no?

eklavya:
@bryce-anderson sorry for the never ending changes :)

client/src/main/scala/org/http4s/client/PoolManager.scala
bryce-anderson: I didn't notice that the `createConnection` method changed, sorry. I still think this is not what we want since it will effectively deprioritize a request if it gets put back at the end of the `waitQueue` and doesn't try another waiter. I kind of liked it the way it was, as a verbose assert.

client/src/main/scala/org/http4s/client/PoolManager.scala
bryce-anderson: When we successfully get a live connection with `getConnectionFromQueue`, it removes a connection from the queue but doesn't remove the queue if it's empty.

client/src/main/scala/org/http4s/client/PoolManager.scala
bryce-anderson: It's made by the `drop` call on `keys`.

bryce-anderson:
No worries @eklavya, this is really tricky stuff so the back and forth is good. I think you're really close on this, and its a great addition.

client/src/main/scala/org/http4s/client/PoolManager.scala
eklavya: @bryce-anderson @rossabaker let me know what you think the behaviour should be. Do we put it on waitQueue or what?

client/src/main/scala/org/http4s/client/PoolManager.scala
eklavya: OOPS !!!

client/src/main/scala/org/http4s/client/PoolManager.scala
eklavya: Ok

blaze-client/src/test/scala/org/http4s/client/blaze/MaxConnectionsInPoolSpec.scala
```diff
@@ -26,8 +28,32 @@ class MaxConnectionsInPoolSpec extends Http4sSpec {
     }
   }
 
+  "Blaze Pooled Http1 Client" should {
+    "Behave and not deadlock" in {
```
rossabaker: I love having a test like this, but we have problems with external resources on Travis CI.  What if we created a service that responds after a small, random amount of latency, and then bound that service to five local ports?

client/src/main/scala/org/http4s/client/PoolManager.scala
bryce-anderson: We want to pick a waiter with something like
```scala
waitQueue.dequeueFirst(numConnectionsCheckHolds(_))
```
It's arguable that if we don't find one that we shouldn't have kill the connection, but that makes the `dequeueFirst` predicate different since the maxConn part will fail.

client/src/main/scala/org/http4s/client/PoolManager.scala
rossabaker: I think @bryce-anderson is correct.  I think we want to do something like:

```scala
waitQueue.dequeueFirst { 
  waiter => allocated.getOrElse(waiter.key) < maxConnectionsPerKey(waiter.key)
} match {
  case Some(Waiter(key, cb)) =>
    // This is the first waiter not blocked on the request key limit.
    // Close the undesired connection and wait for another.
    connection.shutdown()
    decrConnection(key)
    createConnection(k, callback)
  case None =>
     // We're blocked not because of too many connections, but
     // because of too many connections per key.  We might be
     // able to reuse this request.
     idleQueues.add(connection.key, connection)
}
```

client/src/main/scala/org/http4s/client/PoolManager.scala
rossabaker: And I think this is like the other discussion, except we don't put the connection back on the idle queue if we can't find a waiter.

client/src/main/scala/org/http4s/client/PoolManager.scala
```diff
@@ -149,17 +205,25 @@ private final class PoolManager[F[_], A <: Connection[F]](
...
 
-            // returned connection didn't match any pending request: kill it and start a new one for a queued request
             case None =>
-              connection.shutdown()
-              allocated -= 1
-              val Waiting(key, callback) = waitQueue.dequeue()
-              createConnection(key, callback)
+              waitQueue.dequeueFirst { waiter =>
+                allocated.getOrElse(waiter.key, 0) < maxConnectionsPerRequestKey(waiter.key)
+              } match {
```
rossabaker: I think this is right now for this case.  I think this function can be extracted and reused for a `dequeueFirst` on line 236.

I think down there, the Some case just creates a connection, and the None case is a no-op.  So maybe it's as simple as:

```
waitQueue.dequeueFirst { f => _}.map { case Waiting(k, cb) => createConnection(k, cb) }
```

I don't understand @bryce-anderson's [comment](https://github.com/http4s/http4s/pull/1367/files#r139319549) about a different predicate.  I think the logic he's talking about on not closing the connection is handled here, and the connection is already closed down there.

client/src/main/scala/org/http4s/client/PoolManager.scala
```diff
@@ -172,6 +172,52 @@ private final class PoolManager[F[_], A <: Connection[F]](
...
+  private def releaseNonRecyclable(key: RequestKey, connection: A): Unit = {
+    decrConnection(key)
+
+    if (!connection.isClosed) {
+      logger.debug(s"Connection returned was busy.  Shutting down: $stats")
+      connection.shutdown()
+    }
+
+    if (waitQueue.nonEmpty) {
+      logger.debug(s"Connection returned could not be recycled, new connection needed: $stats")
```
rossabaker: We say we're opening a new connection here, but we silently do nothing if `findFirstAllowedWaiter` returns None.  I think we could replace `if (waitQueue.nonEmpty)` with `findFirstAllowedWaiter` and match a `Some` and `None`.

client/src/main/scala/org/http4s/client/PoolManager.scala
```diff
@@ -249,6 +257,11 @@ private final class PoolManager[F[_], A <: Connection[F]](
     }
   }
 
+  private def findFirstAllowedWaitor =
```
rossabaker: `waiter`, not `waitor`

client/src/main/scala/org/http4s/client/PoolManager.scala
```diff
@@ -19,12 +23,46 @@ private final class PoolManager[F[_], A <: Connection[F]](
   private[this] val logger = getLogger(classOf[PoolManager[F, A]])
 
   private var isClosed = false
-  private var allocated = 0
-  private val idleQueue = new mutable.Queue[A]
-  private val waitQueue = new mutable.Queue[Waiting]
+  private var curTotal = 0
```
randallalexander: Should this be an AtomicInteger? I don't believe var is multithread safe.

eklavya: Nope, it's only ever accessed in a synchronised block.

randallalexander: If one thread calls release the code tries to decrement and one thread calls barrow the code tries to increment, I don't see where it is protected by a sync block.

randallalexander: Admittedly I could just be missing it.

eklavya: So how synchronized works is that you need to acquire the lock that it's using to proceed. We are locking on the PoolManager object. So any synchronized block needs to wait if any other synchronized block currently holds the lock (not done executing). So in your case the threads would be serialized.

randallalexander: I missed the fact that there is a sync block in the release call.   It seems like all the public methods have a sync block.

rossabaker: It's gotten increasingly difficult to see the synchronization points with all the complexity in managing the data store.  I think we're in a good place for this feature, but there is certainly room for refactoring here.  Especially as we begin to think about other features, like max wait times, or better algorithms for connection reuse.

But I think this is good enough we need to get it into people's hands.  We'll get this into -M2, and do a quick -M3 if we missed something.

Issue #1327, opened 2017-08-21
GZip client side middleware


agourlay:
Currently the GZip middleware is only for server side.

The client could either reuse the server side middleware or just the compression logic added to a client side middleware.

Discussed on Gitter https://gitter.im/http4s/http4s?at=599aef13c101bc4e3a3f7b71

agourlay:
FYI I am currently working on this

rossabaker:
Which version are you targeting?

- If it's 0.18, cool.
- If it's 0.16 or 0.17 and you're only adding signatures, cool.
- If it's 0.16 or 0.17 and it needs to change any signatures, let's plan, because I'm about to cut RCs.

agourlay:
I am gonna go for 0.18 to get complete freedom :bowtie:

rossabaker:
#1373 was accidentally closed.  That's a good starting point if anyone wants to pick this up and finish it.

agourlay:
fixed by https://github.com/http4s/http4s/pull/2673

PR #1373, opened 2017-09-06
Client side gzip

agourlay:
ref: https://github.com/http4s/http4s/issues/1327

This PR is a draft to get some feedback about the direction to take.

I went with creating a `Compression` abstraction in core to hide the gzip details in there and then proceeded with a client `Gzip` middleware.

As I am not familiar with Http4s and Fs2, I would also love to get ideas regarding testing.

The current spec fails with:

`
[error]   ! GZip.middleware decoding == GZIPOutputStream encoding (545 ms)
[error]    java.util.zip.DataFormatException: Exception raised on property evaluation.
[error]    > ARG_0: ""
[error]    > Exception: java.util.zip.DataFormatException: invalid block type (compress.scala:94)
[error] fs2.compress$._inflate_collect(compress.scala:94)
`


core/src/main/scala/org/http4s/Compression.scala
```diff
@@ -0,0 +1,93 @@
+package org.http4s
```
rossabaker: We've generally reserved this package for core HTTP types.  Supporting code usually lives somewhere else.  Maybe `org.http4s.util`?

client/src/main/scala/org/http4s/client/middleware/GZip.scala
```diff
@@ -0,0 +1,29 @@
...
+          val unZipped = unzipResponse(bufferSize, resp)
+          dr.copy(response = unZipped)
+        case dr => dr
+      }
+
+    client.copy(open = Service.lift(openAndUnzip))
+  }
+
+  private def isZipped[F[_]](resp: Response[F]): Boolean =
+    resp.headers.get(`Content-Type`).exists(_.mediaType eq MediaType.`application/octet-stream`)
```
rossabaker: This content-type isn't necessarily gzipped.  I think we want to look for `Content-Encoding` having a `gzip` parameter?

client/src/main/scala/org/http4s/client/middleware/GZip.scala
```diff
@@ -0,0 +1,29 @@
+package org
+package http4s
```
rossabaker: Usually keep `org.http4s` on one line, or else you're importing everything else in `org.*`.

tests/src/test/scala/org/http4s/CompressionSpec.scala
```diff
@@ -0,0 +1,36 @@
...
+import org.scalacheck.Properties
+
+class CompressionSpec extends Http4sSpec {
+
+  "Compression" should {
+    checkAll(
+      "decode GZIP encoding",
+      new Properties("GZip") {
+        property("middleware decoding == GZIPOutputStream encoding") = forAll {
+          body: String =>
```
rossabaker: I think you could do this on binaries instead of Strings.  You're just testing String codecs on top of everything else.

core/src/main/scala/org/http4s/Compression.scala
```diff
@@ -0,0 +1,93 @@
...
+import fs2.Stream.chunk
+import fs2.compress.{deflate, inflate}
+import org.http4s.headers.{`Content-Encoding`, `Content-Length`, `Transfer-Encoding`}
+import org.log4s.getLogger
+
+object Compression {
+
+  private[this] val logger = getLogger
+
+  def zipResponse[F[_]: Functor](bufferSize: Int, level: Int, resp: Response[F]): Response[F] = {
```
rossabaker: I think this could take a `Message[F]` instead of a `Response[F]`.  Couldn't this apply to requests as well as responses?

core/src/main/scala/org/http4s/Compression.scala
```diff
@@ -0,0 +1,93 @@
...
+            bufferSize = bufferSize
+          )) ++
+      chunk(trailerFinish(trailerGen))
+    resp
+      .removeHeader(`Content-Length`)
+      .putHeaders(`Content-Encoding`(ContentCoding.gzip))
+      .copy(body = b)
+  }
+
+  def unzipResponse[F[_]: Functor](bufferSize: Int, resp: Response[F]): Response[F] = {
```
rossabaker: Same as above with respect to Message.

tests/src/test/scala/org/http4s/CompressionSpec.scala
```diff
@@ -0,0 +1,36 @@
...
+            val entity = EntityEncoder[IO, String].toEntity(body).unsafeRunSync().body
```
rossabaker: What happens in the IO stays in the IO until the end.  I'd start a for-comprehension here and get rod of that `unsafeRunSync()`.

core/src/main/scala/org/http4s/Compression.scala
agourlay: If we go for `Message[F]` then we have to work with the function `withBody` instead of `copy` which requires a `Monad` for `F`.

rossabaker:
This is looking better.  So the question is why that property test is failing.  Are you digging into that, or would you like some help?

agourlay:
First off thank you very much for the guidance, I did struggle a bit to get this test to compile.

TBH I think I was too optimistic regarding the implementation of the decompression, I thought it will be as simple as calling the underlying inflate from FS2 but now I am afraid it is actually more involved.

I would not refuse some help as I am rather stuck here 😃

rossabaker:
So the problem here is that `inflate` is based on deflate compression, and we're doing the closely related gzip, which has a header and a trailer.  I crudely dropped the first ten bytes, and it makes the test pass, though gzip in the wild can have a longer header.  I didn't do anything about the trailer, which seems to silently inflate, though I don't know if anything is checking the CRC32 that it represents.

What we really need here is a proper gzip and gunzip pipe, and the best place for that to live is fs2.  See https://github.com/functional-streams-for-scala/fs2/issues/851.  If we had those, then we could delegate the file format magic there, and just do the HTTP logic in http4s.

Furthermore, there is a trailer, and I'm surprised that inflate handles it.

core/src/main/scala/org/http4s/util/Compression.scala
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  def unzipResponse[F[_]: Functor](bufferSize: Int, resp: Response[F]): Response[F] = {
+    logger.trace("GZip middleware decoding content")
+    val b =
+      // TODO We should validate the header. Futhermore, there are
+      // optional extra headers that can extend the length before we
+      // get to the deflate-compressed body.
+      //
+      // We're also not validating the trailer.
```
rossabaker: I wonder what we should do if we start decoding this and deem it to be invalid GZip.  I think we need either an error return type or restore the original stream.

core/src/main/scala/org/http4s/util/Compression.scala
```diff
@@ -0,0 +1,100 @@
...
+      resp.body
+        .drop(header.size.toLong)
+        .through(
+          inflate(
+            nowrap = true,
+            bufferSize = bufferSize
+          ))
+    resp
+      .removeHeader(`Content-Encoding`)
```
rossabaker: It's weird, but there can be other encodings applied.  I think we'd want to just pop off the gzip at the head of the list.

core/src/main/scala/org/http4s/util/Compression.scala
```diff
@@ -0,0 +1,100 @@
...
+      .putHeaders(`Transfer-Encoding`(TransferCoding.gzip))
```
rossabaker: We're gunzipping, so I'm not sure why we'd add this header here.

Transfer-Encoding is interesting.  The server backend handles decoding chunked transfer encodings.  Maybe the server should handle decoding gzipped transfer encodings?  Seems inconsistent on our part.  And then it raises questions about what the client should to if the request specifies that encoding. 🤔

agourlay:
Thanks a lot for digging into this issue!
Do you think the header trick is a blocker for this issue, silently ignoring the header and trailer does not sound

rossabaker:
Reading the header with java.io [looks like this](https://github.com/dmlloyd/openjdk/blob/e87709def542f064a7ab9fa75542230e40876310/jdk/src/share/classes/java/util/zip/GZIPInputStream.java#L160).  Similar could be used in fs2 to validate the compression method and skip the variable-length header.

If we just skip the first ten bytes, we'll be able to read a lot of GZip streams, including the ones we encoded ourselves.  We'll read some things that look like GZip streams but don't have the correct magic at the front.  And we'll not read things that have flags that cause the header to be longer than 10 bytes.  So it's an interim solution that would cover many (most?) use cases, but its shortcomings should probably be documented with scary words until we make time to do it right.

toebbel:
Hi, I had troubles today with my http4s client side to decode gzip responses and then found this issue. I am not an expert with gzip, but if needed I can try to help to merge this PR. Feel free to ping me.

rossabaker:
I haven't investigated enough to see if the drop-ten-bytes solution will work in the majority of cases.  I'd try that, and if it's good, I'm okay shipping this with a big warning about needing to handle the variable length headers.

rossabaker:
I'm cleaning up everything I accidentally closed, but I still think this was a good idea.  Any volunteers to get this one across the finish line?

djspiewak:
I did this in a more general way on top of fs2: https://github.com/slamdata/fs2-gzip/pull/1 I'm going to see about PRing this to fs2 so it doesn't introduce a dependency. There are no gzip-encoding-specific things in the code; everything delegates to the `java.util.zip` implementations.

PR #1375, opened 2017-09-06
Update some docs and comments and clean up

aeons:
- Update docs and comments with changes wrt. `Task` => effect
- Remove some obsolete files
- Update `AuthenticationSpec` to `fs2`


core/src/main/scala/org/http4s/Service.scala
```diff
@@ -36,7 +36,7 @@ object Service {
   def constVal[F[_], A, B](b: => B)(implicit F: Sync[F]): Service[F, A, B] =
     lift(_ => F.delay(b))
 
-  /** Allows Service chainig through a `scalaz.Monoid` instance. */
+  /** Allows Service chainig through a Monoid instance. */
```
ChristopherDavenport: *chaining

core/src/main/scala/org/http4s/util/io/package.scala
```diff
@@ -1,91 +0,0 @@
-/** TODO fs2 port
```
ChristopherDavenport: Why are we dropping this? Not that we shouldn't just checking.

server/src/test/scala/org/http4s/server/middleware/authentication/AuthenticationSpec.scala
```diff
@@ -137,93 +154,107 @@ class AuthenticationSpec extends Http4sSpec {
...
-            case _ => false
-          }) must_== true
-          val res = doDigestAuth2(digestAuthService, challenge, false)._1
-          // We don't check whether res.status is Ok since it may not
-          // be due to the low nonce stale timer.  Instead, we check
-          // that it's found.
-          res.status mustNotEqual (NotFound)
-        }(sched))
-      Task.gatherUnordered(tasks).run
+      implicit val ec = ExecutionContext.fromExecutor(Executors.newFixedThreadPool(4))
```
ChristopherDavenport: Your threads are showing. This isn't you, but should get fixed.

server/src/test/scala/org/http4s/server/middleware/authentication/AuthenticationSpec.scala
```diff
@@ -137,93 +154,107 @@ class AuthenticationSpec extends Http4sSpec {
...
+      async.parallelSequence(results).unsafeRunSync
 
       ok
     }
 
     "Avoid many concurrent replay attacks" in {
       val n = 100
-      val sched = Executors.newFixedThreadPool(4)
-      val digestAuthService = DigestAuth(realm, authStore)(service)
+      implicit val ec = ExecutionContext.fromExecutor(Executors.newFixedThreadPool(4))
```
ChristopherDavenport: Same as above

core/src/main/scala/org/http4s/util/io/package.scala
aeons: This would be https://github.com/functional-streams-for-scala/fs2/blob/series/0.10/io/src/main/scala/fs2/io/io.scala instead, i think.

server/src/test/scala/org/http4s/server/middleware/authentication/AuthenticationSpec.scala
rossabaker: Wow, am I glad to find these.

PR #1377, opened 2017-09-08
Remove the Semigroup[F[MaybeResponse[F]]] constraint from classes

aeons:
by re-introducing the implicit `Monoid[F[MaybeResponse[F]]]` in `MaybeResponseInstances`.

This trait is then mixed into the `Http4sInstances` trait, which puts it into scope in most places where a user would need it, i.e. when importing or mixing in the DSL.

To let `BlazeBuilder` pick the right instance, when instantiated with `IO` as the effect type, `BlazeBuilder` now mixes in the `MaybeResponseInstances` trait as well.

### DIFF
--- a/blaze-server/src/main/scala/org/http4s/server/blaze/BlazeBuilder.scala
+++ b/blaze-server/src/main/scala/org/http4s/server/blaze/BlazeBuilder.scala
@@ -6,10 +6,8 @@ import java.io.FileInputStream
 import java.net.InetSocketAddress
 import java.nio.ByteBuffer
 import java.security.{KeyStore, Security}
-import java.util.concurrent.ExecutorService
 import javax.net.ssl.{KeyManagerFactory, SSLContext, SSLEngine, TrustManagerFactory}
 
-import cats._
 import cats.effect._
 import org.http4s.blaze.channel
 import org.http4s.blaze.channel.SocketConnection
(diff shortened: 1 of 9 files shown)


rossabaker:
Why was this necessary in the first place?  Because the router uses it to concatenate all mounted services, and blaze uses a router instead of something like servlet mappings?

aeons:
Yes, I guess. 

It's needed to use `|+|` on two `HttpService`, which the router does.

rossabaker:
Extending instances directly from the builder seems weird to me.  I guess with the common pattern of services in one file and a server in the other file, it's likely that `org.http4s.Http4s._` isn't in scope where the `BlazeBuilder` is used?  I'm slightly uncomfortable with this, but I think it may be worth it to reduce friction.

aeons:
The problem with it being an import is that when you try to do `BlazeBuilder[IO]`, you get a multiple implicit error, because of [IO's own `ioMonoid`](https://github.com/typelevel/cats-effect/blob/master/core/shared/src/main/scala/cats/effect/IO.scala#L357).

aeons:
I'll take a stab at doing it a different way.

aeons:
So, I guess the import in `Router` made the multiple implicit error go away. Anyway it works without `BlazeBuilder` extending anything :)

PR #1379, opened 2017-09-08
Fix AutoSlash middleware when a service is mounted prefixed

### LINKED ISSUE #1378: AutoSlash not working when prefix is specified in mountService
I tested this in http4s 0.18.0-M1

How to reproduce:
```
package com.example.http4squickstart

import io.circe._
import org.http4s._
import org.http4s.circe._
import org.http4s.server._
import org.http4s.dsl._
import cats.effect.IO

object HelloWorld {
      Ok(Json.obj("message" -> Json.fromString(s"Hello, ${name}")))
  }
}
```

```
package com.example.http4squickstart

import cats.effect.IO
import fs2.Stream

import org.http4s.server.middleware.AutoSlash

object BlazeExample extends StreamApp[IO] {

  val port : Int              = envOrNone("HTTP_PORT") map (_.toInt) getOrElse 8080
  val ip   : String           = "0.0.0.0"

  override def stream(args: List[String], requestShutdown: IO[Unit]): Stream[IO, Nothing] =
    BlazeBuilder[IO]
      .bindHttp(port, ip)
      .mountService(AutoSlash(HelloWorld.service), "/test/")
      .serve

}
```

`curl -v http://localhost:8080/test/hello/alessandro` works.
```
*   Trying ::1...
* TCP_NODELAY set
* Connected to localhost (::1) port 8080 (#0)
> GET /test/hello/alessandro HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.52.1
> Accept: */*
> 
< HTTP/1.1 200 OK
< Content-Type: application/json
< Date: Fri, 08 Sep 2017 17:01:52 GMT
< Content-Length: 31
< 
* Curl_http_done: called premature == 0
* Connection #0 to host localhost left intact
{"message":"Hello, alessandro"}
```

`curl -v http://localhost:8080/test/hello/alessandro/` doesn't.
```
*   Trying ::1...
* TCP_NODELAY set
* Connected to localhost (::1) port 8080 (#0)
> GET /test/hello/alessandro/ HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.52.1
> Accept: */*
> 
< HTTP/1.1 404 Not Found
< Date: Fri, 08 Sep 2017 17:02:33 GMT
< Content-Length: 0
< 
* Curl_http_done: called premature == 0
* Connection #0 to host localhost left intact
```

If I remove the prefix and curl without the prefix AutoSlash works.

aeons:
Fixes #1378

### DIFF
--- a/server/src/main/scala/org/http4s/server/middleware/AutoSlash.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/AutoSlash.scala
@@ -16,13 +16,11 @@ object AutoSlash {
     req =>
       service(req).flatMap {
         case Pass() =>
-          val p = req.uri.path
-          if (p.isEmpty || p.charAt(p.length - 1) != '/')
+          val pi = req.pathInfo
+          if (pi.isEmpty || pi.charAt(pi.length - 1) != '/')
             Pass.pure[F]
           else {
-            val caret = req.attributes(Request.Keys.PathInfoCaret)
-            val withSlash = req.withUri(req.uri.copy(path = p.substring(caret, p.length - 1)))
-            service.apply(withSlash)
+            service.apply(req.withPathInfo(pi.substring(0, pi.length - 1)))
           }
         case resp =>
           F.pure(resp)
--- a/server/src/main/scala/org/http4s/server/middleware/AutoSlash.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/AutoSlash.scala
@@ -4,7 +4,6 @@ package middleware
 
 import cats._
 import cats.implicits._
-import fs2._
 
 /** Removes a trailing slash from [[Request]] path
   *
@@ -16,12 +15,13 @@ object AutoSlash {
   def apply[F[_]](service: HttpService[F])(implicit F: Monad[F]): HttpService[F] = Service.lift {
     req =>
       service(req).flatMap {
-        case _: Pass[F] =>
+        case Pass() =>
           val p = req.uri.path
           if (p.isEmpty || p.charAt(p.length - 1) != '/')
-            F.pure(Pass[F])
+            Pass.pure[F]
           else {
-            val withSlash = req.withUri(req.uri.copy(path = p.substring(0, p.length - 1)))
+            val caret = req.attributes(Request.Keys.PathInfoCaret)
+            val withSlash = req.withUri(req.uri.copy(path = p.substring(caret, p.length - 1)))
             service.apply(withSlash)
           }
         case resp =>


server/src/main/scala/org/http4s/server/middleware/AutoSlash.scala
```diff
@@ -16,12 +15,13 @@ object AutoSlash {
   def apply[F[_]](service: HttpService[F])(implicit F: Monad[F]): HttpService[F] = Service.lift {
     req =>
       service(req).flatMap {
-        case _: Pass[F] =>
+        case Pass() =>
```
rossabaker: This means the same thing but avoids a check against an erased type, right?

server/src/main/scala/org/http4s/server/middleware/AutoSlash.scala
```diff
@@ -16,12 +15,13 @@ object AutoSlash {
...
           val p = req.uri.path
           if (p.isEmpty || p.charAt(p.length - 1) != '/')
-            F.pure(Pass[F])
+            Pass.pure[F]
```
rossabaker: This is stylistic, right?

server/src/main/scala/org/http4s/server/middleware/AutoSlash.scala
```diff
@@ -16,12 +15,13 @@ object AutoSlash {
...
           else {
-            val withSlash = req.withUri(req.uri.copy(path = p.substring(0, p.length - 1)))
+            val caret = req.attributes(Request.Keys.PathInfoCaret)
+            val withSlash = req.withUri(req.uri.copy(path = p.substring(caret, p.length - 1)))
```
rossabaker: Doesn't this lose the scriptName part of the path?  Should this be `req.withPathInfo`?

I'll bet this is also broken on older versions.

server/src/main/scala/org/http4s/server/middleware/AutoSlash.scala
aeons: I think so yes, it's also to be consistent with other places in the code.

server/src/main/scala/org/http4s/server/middleware/AutoSlash.scala
aeons: Yes, and for consistency.

server/src/main/scala/org/http4s/server/middleware/AutoSlash.scala
aeons: Maybe, I don't know how `pathInfo` works.

The problem right now is that the prefixed part is back in the URI when it's passed to a service that doesn't expect it.

I guess it should `get` and `getOrElse` on the attribute as well.

PR #1380, opened 2017-09-10
Publish into http4s.org repo

rossabaker:
This is how release-0.15.x is publishing into the [http4s.org](https://github.com/http4s/http4s.org) repo.  I accidentally cowboyed that one, but we can talk about it here.


bin/publishSite.sh
```diff
@@ -1,5 +1,10 @@
 #!/usr/bin/env bash
 set -e
+shopt -s nullglob
+
+REPO=https://github.com/$TRAVIS_REPO_SLUG.git
+REPO_SSH=git@github.com:$TRAVIS_REPO_SLUG
+VERSION=v0.16
```
rossabaker: This should to be DRYed up from the build.  Ideas better than `sbt version`?

docs/src/hugo/layouts/partials/nav-docs.html
```diff
@@ -1,23 +1,8 @@
...
-    <li class="dropdown-header">v0.15</li>
-    <li><a href="/v0.15/">Tutorial</a></li>
-    <li><a href="/v0.15/api/org/http4s">Scaladoc</a></li>
-  </ul>
-</li>
+<div class="dropdown-menu" aria-labelledby="doc-menu-item"> 
+  <a class="dropdown-item" href="/v0.18/">v0.18 (development)</a>
+  <a class="dropdown-item" href="/v0.18/">v0.17 (stable)</a>
+  <a class="dropdown-item" href="/v0.18/">v0.16 (stable)</a>
+  <a class="dropdown-item" href="/versions/">Help me choose...</a>
```
rossabaker: This is the only part of the theme I deeply care about going back and republishing old versions when it changes.  Maybe we could stuff a JSON blob out in http4s.org, and have this dropdown populate itself from that.  That could refresh EOLed versions without touching them again.

rossabaker:
I like the recent content improvements on http4s/http4s.org, but all told, I think we're shifting complexity rather than reducing it.  Going to rethink this.

PR #1381, opened 2017-09-11
Split docs project into website and docs

rossabaker:
This divides the site into two projects, `docs` and `website`.  Like now, `website` will be published off master and `docs` will be published off the release branches.  A key difference is that each project gets its own static assets, so we can change the `website` project without needing to republish all versions of `docs`.  It also becomes much easier to preview the site from either project.

This starts all the way back on release-0.15.x, as it's the earliest branch published.  This is a one-time merge train so each site gets its own copy of the static assets.

The upgrade of sbt-ghpages should pave the way toward an sbt-1.x upgrade.

I envision the work done in the http4s.org repo being moved here, but only to master.  Trying to keep the changes minimal on the older branches.


build.sbt
```diff
@@ -332,43 +301,64 @@ lazy val docs = http4sProject("docs")
       IO.write(dest, buildData)
       dest
     },
-    copySiteToStage := copySiteToStage.dependsOn(tutQuick).value,
-    makeSite := makeSite.dependsOn(copySiteToStage, exportMetadataForSite).value,
+    makeSite := makeSite.dependsOn(tutQuick, exportMetadataForSite).value,
     baseURL in Hugo := {
-      if (isTravisBuild.value) new URI(s"http://http4s.org")
+      if (isTravisBuild.value) new URI(s"http://http4s.org/v0.15")
```
rossabaker: A setting can't depend on a task, so the v0.15 is hardcoded. 👎

build.sbt
```diff
@@ -332,43 +301,64 @@ lazy val docs = http4sProject("docs")
...
     siteMappings ++= {
       val m = (mappings in (ScalaUnidoc, packageDoc)).value
       val (major, minor) = apiVersion.value
-      for ((f, d) <- m) yield (f, s"v$major.$minor/api/$d")
+      for ((f, d) <- m) yield (f, s"api/$d")
+    },
+    includeFilter in ghpagesCleanSite := {
+      new FileFilter{
+        def accept(f: File) =
+          f.getCanonicalPath.startsWith((ghpagesRepository.value / "v0.15").getCanonicalPath)
```
rossabaker: A setting can't depend on a task, so the v0.15 is hardcoded. 👎

build.sbt
```diff
@@ -332,43 +301,64 @@ lazy val docs = http4sProject("docs")
...
+      }
+    },
+    ghpagesPushSite ~= { old =>
```
rossabaker: Oops, this belongs on the website project.

build.sbt
```diff
@@ -332,43 +301,64 @@ lazy val docs = http4sProject("docs")
...
+  .enablePlugins(HugoPlugin, GhpagesPlugin)
+  .settings(
+    description := "Common area of http4s.org",
+    baseURL in Hugo := {
+      if (isTravisBuild.value) new URI(s"http://http4s.org")
+      else new URI(s"http://127.0.0.1:${previewFixedPort.value.getOrElse(4000)}")
+    },
+    // all .md|markdown files go into `content` dir for hugo processing
+    ghpagesNoJekyll := true,
+    includeFilter in Hugo := (
```
rossabaker: Maybe we could make this a project-level setting, as it's copied in the docs project.

build.sbt
```diff
@@ -332,43 +301,64 @@ lazy val docs = http4sProject("docs")
...
+        "*.html" |
+        "*.png" | "*.jpg" | "*.gif" | "*.ico" | "*.svg" |
+        "*.js" | "*.swf" | "*.json" | "*.md" |
+        "*.css" | "*.woff" | "*.woff2" | "*.ttf" |
+        "CNAME" | "_config.yml"
+    ),
+    excludeFilter in ghpagesCleanSite := {
+      new FileFilter{
+        def accept(f: File) =
+          f.getCanonicalPath.startsWith((ghpagesRepository.value / "v0.*").getCanonicalPath)
```
rossabaker: This is going to bite us when we go 1.0.  I don't know how to cleanly target the `/^v[0-9]+\./` directories.  An alternative would be to put all the tutorials in a `docs/` subdirectory of the published site (as is common on other projects), but that would change all our URIs.

build.sbt
```diff
@@ -332,43 +301,64 @@ lazy val docs = http4sProject("docs")
...
+      }
+    },
+    git.remoteRepo := "git@github.com:http4s/http4s.git"
```
rossabaker: This is sad.  It's already defined, but ghPages complains about an uninitialized reference unless we restate it here.

docs/src/hugo/config.toml
```diff
@@ -1,8 +1,11 @@
-baseurl = "http://http4s.org/"
+baseurl = "http://http4s.org/v0.15"
+contentDir = "../../target/scala-2.12/tut"
 languageCode = "en-us"
 title = "http4s"
+theme = "http4s.org"
+themesDir = "../../../website/src/hugo/themes"
```
rossabaker: This keeps the common branding elements in one place.  It won't automatically refresh old versions of `docs`, but it will keep the current version of `docs` looking about like the `website`.

Note that the static assets herein will be copied into the `vX.Y` branch.  This is critical to letting us continue to evolve the theme on `website`.

docs/src/hugo/config.toml
```diff
@@ -1,8 +1,11 @@
-baseurl = "http://http4s.org/"
+baseurl = "http://http4s.org/v0.15"
```
rossabaker: This is another hardcoding of the version I'd love to get rid of, but it merely shifts from menu.tut, so it's not worse.

docs/src/hugo/data/build.toml
```diff
@@ -0,0 +1,10 @@
+
+[versions]
```
rossabaker: This file is generated.  We shouldn't be generating this into src.

website/src/hugo/content/versions.md
```diff
@@ -206,7 +206,7 @@ title: Versions
       </table>
     </div>
     <div role="tabpanel" class="tab-pane" id="changelog">
-      {{< readfile "CHANGELOG.md" >}} 
+      {{< readfile "content/CHANGELOG.md" >}}
```
rossabaker: This buries the changelog pretty deep in the directory structure, but avoids a hacky copy in the build.  Hugo can't read outside its parent directory, probably because security and stuff.

An alternative would be a symlink, but I think that's still shady on Windows, and we have a few such developers.

website/src/hugo/data/build.toml
```diff
@@ -0,0 +1,10 @@
+
+[versions]
```
rossabaker: Criminy, we got two of these when we wanted zero.

PR #1382, opened 2017-09-11
Make parboiled2 an actual module

rossabaker:
This solves the problem of being unable to publishLocal and then depend on that when testing http4s.

I was reluctant to do this, because it appears to be another dependency, which is used in negative marketing.  All code in this newly published module is internal and was already in core.  Haters gonna hate.  This just simplifies our build and makes a better story for local development.

Alternatively, I tried `ivyXML` in addition to `pomPostProcess`, but I couldn't get it to affect the delivered Ivy file.


rossabaker:
I think this one has irritated @ChristopherDavenport a lot, so I'll wait for his feedback, too.

PR #1384, opened 2017-09-11
Port remaining examples to Cats.

randallalexander:
Ported the 2 remaining examples to Cats
  - This is minor but it _should_ close out #866 unless I am missing something

### DIFF
--- a/examples/blaze/src/main/scala/com/example/http4s/blaze/BlazeHttp2Example.scala
+++ b/examples/blaze/src/main/scala/com/example/http4s/blaze/BlazeHttp2Example.scala
@@ -1,22 +1,23 @@
-//package com.example.http4s
-//package blaze
-//
-//import com.example.http4s.ssl.SslExample
-//import org.http4s.server.blaze.BlazeBuilder
-//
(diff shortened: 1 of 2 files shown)


examples/blaze/src/main/scala/com/example/http4s/blaze/BlazeHttp2Example.scala
```diff
@@ -1,22 +1,23 @@
...
-//}
+package com.example.http4s
+package blaze
+
+import cats.effect._
+import com.example.http4s.ssl.SslExample
+import org.http4s.server.blaze.BlazeBuilder
+
+/** Note that Java 8 is required to run this demo along with
+  * loading the jetty ALPN TSL classes to the boot classpath.
```
rossabaker: When this comment was written, HTTP/2 required Java 8.  Now the entire library does.  I think we could safely get rid of this comment as well.

randallalexander: Sounds good.  I have removed the comment.

PR #1385, opened 2017-09-12
Added convenience method to Request for X-Forward-For

### LINKED ISSUE #681: Helper for X-Forwarded-For header
`Request` has a `remote` method on it for grabbing the IP address. It would be nice if there were a helper (as either a separate method or as middleware) to check the `X-Forwarded-For` header and fall back to the remote IP. This is a common pattern when load balancers are used.

randallalexander:
Resolves #681 
 - Not the biggest fan of the names but I can't think of anything better

### DIFF
--- a/core/src/main/scala/org/http4s/Message.scala
+++ b/core/src/main/scala/org/http4s/Message.scala
@@ -264,10 +264,10 @@ sealed abstract case class Request[F[_]](
   lazy val remote: Option[InetSocketAddress] = connectionInfo.map(_.remote)
 
   /**
-    Convenience method to grab the first X-Forward-For.  This is a convience method.  If
+    Convenience method to grab the first X-Forwarded-For.  This is a convenience method.  If
     all the X-Forwarded-For are needed then you will need to pull it from the headers.
     */
-  lazy val forwardFor: Option[InetAddress] = headers
+  lazy val forwardedFor: Option[InetAddress] = headers
     .get(http4s.headers.`X-Forwarded-For`)
     .map(_.values)
     .flatMap(_.head)
--- a/core/src/main/scala/org/http4s/Message.scala
+++ b/core/src/main/scala/org/http4s/Message.scala
@@ -4,9 +4,11 @@ import java.io.File
 import java.net.{InetAddress, InetSocketAddress}
 
 import cats._
+import cats.data.NonEmptyList
 import cats.implicits._
 import fs2._
 import fs2.text._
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/http4s/Message.scala
```diff
@@ -260,6 +262,24 @@ sealed abstract case class Request[F[_]](
   private lazy val connectionInfo = attributes.get(Keys.ConnectionInfo)
 
   lazy val remote: Option[InetSocketAddress] = connectionInfo.map(_.remote)
+
+  /**
+    Convenience method to grab the first X-Forward-For.  This is a convience method.  If
+    all the X-Forwarded-For are needed then you will need to pull it from the headers.
+    */
+  lazy val forwardFor: Option[InetAddress] = headers
```
aeons: Shouldn't this be `forwardedFor`? To be consistent with the header name.

randallalexander: @aeons yes, you are correct.  I will change that.

core/src/main/scala/org/http4s/Message.scala
```diff
@@ -4,9 +4,11 @@ import java.io.File
 import java.net.{InetAddress, InetSocketAddress}
 
 import cats._
+import cats.data.NonEmptyList
 import cats.implicits._
 import fs2._
 import fs2.text._
+import org.http4s
```
rossabaker: This isn't necessary, is it?

randallalexander:
@rossabaker How I referenced the header is what required the import.  However it should be cleaned up and I just pushed the change.  I am indifferent on keeping `forwardedFor`.  `from` really solves the intended problem.  If you feel I should yank it just let me know and I will remove it.

randallalexander:
@rossabaker If it throws you over the fence one way or another `forwardedFor` is not needed for `from`.  Original implementation did require it but I figured out out a simpler way to impl `from` and left the `forwardedFor` there because it is somewhat related and I had already impled it.

rossabaker:
Okay, yeah, let's remove that then and keep the `from`.

One other thing: @ceedubs just brought up that the lazy vals are probably not necessary, and I'm inclined to agree.  You're just following prior art there, but we've started removing them elsewhere, and I imagine this is fast enough and once-calculated enough that we could get away with a simple def.

randallalexander:
@rossabaker Just submitted a change for both of those items.

PR #1387, opened 2017-09-12
Update examples

aeons:
Update examples to use a parametric effect type in the base and extend that with `IO` in the final object.

I have also moved the `Scheduler` needed in `ScienceExperiments` and `ExampleService` out, so it's a parameter. Currently it's an implicit parameter, which looks nice and simple, but it could just as easily be explicit.

This uses #1386, so that should be merged before this.

Once #1384 is merged, I will update this PR for those examples as well.


aeons:
No updates were needed for #1384, so this is ready for merging now.

PR #1390, opened 2017-09-13
Fix MiMa detection after parboiled2 split

rossabaker:
This broke when we split off the parboiled2 module.  It wasn't uncovered on release-0.16.x because the parsing of the scalaz suffix was broken.  This brings back MiMa on the 0.16.x series and should fix the problem on 0.17.x.


build.sbt
```diff
@@ -34,7 +34,9 @@ lazy val parboiled2 = libraryProject("parboiled2")
     },
     // https://issues.scala-lang.org/browse/SI-9490
     (scalacOptions in Compile) --= Seq("-Ywarn-inaccessible", "-Xlint", "-Xlint:inaccessible"),
-    macroParadiseSetting
+    macroParadiseSetting,
+    // Remove starting in 0.18
+    mimaPreviousArtifacts := Set.empty
```
rossabaker: Laptop is busy so editing directly on GitHub. Wish me luck. 🤞

ChristopherDavenport: :+1:

PR #1392, opened 2017-09-13
Messy Initial TypeClass Sketch

ChristopherDavenport:
Primary Concern for Discussion is Message Design. Lots of things to be hammered out elsewhere, but some very conflicting things even internally in message due to how much inheritance was being used.


rossabaker:
This seems like a good start.  It would be neat if `Part` could be a `Message`, and participate in the decoding.

rossabaker:
This is probably an enormous, unfeasible breaking change, but: can we imagine a world where there are no EffectResponseOps or EffectRequestOps?  How do we get from here to there?

core/src/main/scala/org/http4s/EntityDecoder.scala
```diff
@@ -23,25 +23,28 @@ import scala.util.control.NonFatal
   */
 @implicitNotFound(
   "Cannot decode into a value of type ${T}, because no EntityDecoder[${F}, ${T}] instance could be found.")
-trait EntityDecoder[F[_], T] { self =>
+trait EntityDecoder[F[_], T, M[_[_]]] { self =>
```
rossabaker: Try moving the `M` off here and as a parameter to `decode`.  What shakes loose?

aeons:
Can this be closed, @ChristopherDavenport?

PR #1393, opened 2017-09-14
Deprecate effect syntax

rossabaker:
Early on, we added syntax to blur the lines between `Message[F]` and `F[Message[F]]`.  This magic has tended to confuse new users who don't understand implicits and confuse experienced users who don't understand why they don't need to `.map` their `F`.

The most common case where this is used is `.putHeaders` on an `F[Response[F]]`.  If we add variadic headers to the smart status and method constructors, we preserve a concise means of building these requests and responses without blurring the lines.

Could simplify the work in #1392, but it still needs to be made to compile if we want to be gentle about deprecation warnings, and I think we do.

/cc @SystemFw, who participated in the [Gitter discussion](https://gitter.im/http4s/http4s?at=59b9a49e614889d475f4e49b).


client/src/test/scala/org/http4s/client/middleware/FollowRedirectSpec.scala
```diff
@@ -132,7 +133,7 @@ class FollowRedirectSpec extends Http4sSpec with Tables {
       val statefulService = HttpService[IO] {
         case GET -> Root / "loop" =>
           val body = loopCounter.incrementAndGet.toString
-          MovedPermanently(uri("/loop")).withBody(body)
+          MovedPermanently(uri("/loop")).flatMap(_.withBody(body))
```
rossabaker: This is an example of a response generator that has a required header.  It should probably take the body and perhaps more variadic headers.  Then it would look like:

```scala
MovedPermanently(uri("/loop"), body)
```

core/src/main/scala/org/http4s/syntax/EffectRequestSyntax.scala
```diff
@@ -5,6 +5,7 @@ import cats._
 import cats.implicits._
 
 trait EffectRequestSyntax extends Any {
+  @deprecated("Use map or flatMap directly on the request", "0.18.0-M2")
```
rossabaker: This message could also refer to the new headers arguments.

We could be more precise if we'd move the deprecation to the ops class, but then we'd need to override every method, because it inherits a bunch of (undeprecated) concrete implementations.

rossabaker: By deprecating here instead of deprecating the syntax type itself, users of `Http4sDsl` or `AllSyntax` can emit deprecation warnings on usage without itself extending a deprecated type (a fatal warning).

dsl/src/main/scala/org/http4s/dsl/impl/ResponseGenerator.scala
```diff
@@ -21,13 +21,18 @@ trait ResponseGenerator extends Any {
   * }}}
   */
 trait EmptyResponseGenerator[F[_]] extends Any with ResponseGenerator {
-  def apply()(implicit F: Applicative[F]): F[Response[F]] = F.pure(Response(status))
+  def apply()(implicit F: Applicative[F]): F[Response[F]] =
+    F.pure(Response(status))
+
+  def apply(headers: Header*)(implicit F: Applicative[F]): F[Response[F]] =
```
rossabaker: Now that I look at it, this probably doesn't need to be an overload.

dsl/src/main/scala/org/http4s/dsl/impl/ResponseGenerator.scala
```diff
@@ -37,10 +42,10 @@ trait EntityResponseGenerator[F[_]] extends Any with ResponseGenerator {
   def apply()(implicit F: Applicative[F]): F[Response[F]] =
     F.pure(Response(status, headers = Headers(`Content-Length`.zero)))
 
-  def apply[A](body: A)(implicit F: Monad[F], w: EntityEncoder[F, A]): F[Response[F]] =
-    apply(body, Headers.empty)(F, w)
+  def apply(headers: Header*)(implicit F: Applicative[F]): F[Response[F]] =
```
rossabaker: Also doesn't need to be an overload.

Note: we don't extend `EmptyResponseGenerator` because we can (and should) set `Content-Length: 0` here.  We can't safely do it on the empty responses, because HTTP is a complicated beast.

examples/src/main/scala/com/example/http4s/ExampleService.scala
```diff
@@ -66,10 +65,7 @@ class ExampleService[F[_]](implicit F: Effect[F]) extends Http4sDsl[F] {
 
       case GET -> Root / "content-change" =>
         // EntityEncoder typically deals with appropriate headers, but they can be overridden
-        val resp: F[Response[F]] = Ok("<h2>This will have an html content type!</h2>")
-        val fr: EffectResponseOps[F] = http4sEffectResponseSyntax(resp)
-        val w: F[Response[F]] = fr.withContentType(Some(`Content-Type`(`text/html`)))
-        w
+        Ok("<h2>This will have an html content type!</h2>", `Content-Type`(`text/html`))
```
rossabaker: Whether or not we proceed with this, we should get that other mess out of our example.

server/src/test/scala/org/http4s/server/middleware/ChunkAggregatorSpec.scala
```diff
@@ -27,10 +27,8 @@ class ChunkAggregatorSpec extends Http4sSpec {
     def checkResponse(body: EntityBody[IO], transferCodings: List[TransferCoding])(
         responseCheck: Response[IO] => MatchResult[Any]): MatchResult[Any] = {
       val service: HttpService[IO] = HttpService.lift[IO] { _ =>
-        Ok()
-          .putHeaders(`Transfer-Encoding`(NonEmptyList(TransferCoding.chunked, transferCodings)))
-          .removeHeader(`Content-Length`)
-          .withBody(body)
+        Ok(body, `Transfer-Encoding`(NonEmptyList(TransferCoding.chunked, transferCodings)))
```
rossabaker: Sometimes I think `Message` should have a `.chunked`, which adds this header and removes any `Content-Length` header.  Having both is bad.

server/src/test/scala/org/http4s/server/middleware/DefaultHeadSpec.scala
```diff
@@ -12,10 +12,10 @@ class DefaultHeadSpec extends Http4sSpec {
       Ok("hello")
 
     case GET -> Root / "special" =>
-      Ok().putHeaders(Header("X-Handled-By", "GET"))
+      Ok(Header("X-Handled-By", "GET"))
```
rossabaker: This is a little weird, because the first argument isn't the body.  Maybe we should force `Ok((), Header(..)` in this case?

blaze-server/src/test/scala/org/http4s/server/blaze/ServerTestRoutes.scala
```diff
@@ -100,22 +99,24 @@ object ServerTestRoutes {
...
-      Response[IO](Ok).withBody(eval(IO.shift >> IO("chu")) ++ eval(IO.shift >> IO("nk")))
+      Ok(eval(IO.shift >> IO("chu")) ++ eval(IO.shift >> IO("nk")))
 
     case req if req.method == Method.POST && req.pathInfo == "/post" =>
-      Response(Ok).withBody("post")
+      Ok("post")
 
     case req if req.method == Method.GET && req.pathInfo == "/twocodings" =>
-      Response[IO](Ok).withBody("Foo").putHeaders(`Transfer-Encoding`(TransferCoding.chunked))
+      Ok("Foo", `Transfer-Encoding`(TransferCoding.chunked))
```
rossabaker: Alternate proposal, without variadic headers:

```scala
Ok("Foo", Headers(`Transfer-Encoding`(TransferCoding.chunked)))
```

client/src/test/scala/org/http4s/client/middleware/FollowRedirectSpec.scala
```diff
@@ -16,7 +16,8 @@ class FollowRedirectSpec extends Http4sSpec with Tables {
 
   val service = HttpService[IO] {
     case req @ _ -> Root / "ok" =>
-      Ok(req.body).putHeaders(
+      Ok(
```
rossabaker: Alternate proposal, without variadic headers:

```scala
      Ok(
        req.body,
         Headers(
           Header("X-Original-Method", req.method.toString),
           Header("X-Original-Method", req.method.toString),
           Header(
             "X-Original-Content-Length",
             req.headers.get(`Content-Length`).fold(0L)(_.length).toString),
           Header("X-Original-Authorization", req.headers.get(Authorization.name).fold("")(_.value))
           Header("X-Original-Authorization", req.headers.get(Authorization.name).fold("")(_.value))
       )
     )
)
```

Yeah, this one was ugly anyway.

client/src/test/scala/org/http4s/client/middleware/FollowRedirectSpec.scala
```diff
@@ -152,19 +153,21 @@ class FollowRedirectSpec extends Http4sSpec with Tables {
     }
 
     "Not send sensitive headers when redirecting to a different authority" in {
-      val req = Request[IO](PUT, uri("http://localhost/different-authority"))
-        .withBody("Don't expose mah secrets!")
-        .putHeaders(Header("Authorization", "Bearer s3cr3t"))
+      val req = PUT(
```
rossabaker: Alternate proposal, without variadic headers:

```scala
PUT(uri("http://localhost/different-authority"),
    "Don't expose mah secrets!",
    Headers(Header("Authorization", "Bearer s3cr3t")))
```

examples/src/main/scala/com/example/http4s/ExampleService.scala
aeons: That looks like some debugging I did the other day...

dsl/src/main/scala/org/http4s/dsl/impl/ResponseGenerator.scala
aeons: It will just be an empty list then, right?

server/src/test/scala/org/http4s/server/middleware/ChunkAggregatorSpec.scala
aeons: 👍 from me on `.chunked`

rossabaker:
Related: client has a bunch of overloads for `Request` and `F[Request]`, which is causing confusion in the channel [right now](https://gitter.im/http4s/http4s?at=59baa10e210ac269202f414d).

dsl/src/main/scala/org/http4s/dsl/impl/ResponseGenerator.scala
rossabaker: Yes.

rossabaker:
There are three response generators that are "special":

- LocationResponseGenerator: for various 3xx that SHOULD have a `Location`.  No problem adding variadic headers here, with optional body support.

- WwwAuthenticateResponseGenerator: for 401, which MUST have at least one `Challenge`.  This is already variadic on Challenge, so we're in a bad place here.  We need to compromise enforcing correctness or compromise harmony with the rest of the DSL.

- ProxyAuthenticateResponseGenerator: for 407, which works just like the 401, but for a different header.

rossabaker:
Tentative solution to the above:

- Effectively deprecate LocationResponseGenerator.  For example, `MovedPermanently(uri)` becomes `MovedPermanently(Location(uri))` and looks like every other `EntityResponseGenerator`.
- Make the authentication statuses like any other `EntityResponseGenerator`, except with a required header before the variadic headers.  This enforces correctness, but if called correctly, behaves like everything else.

PR #1397, opened 2017-09-14
 Refresh website

rossabaker:
Brings the changes from the http4s/http4s.org repo, which I'm abandoning.  That split is now managed by docs/ vs. website/


rossabaker:
I had trouble fitting them across the top, because there are a couple more pages than there used to be.  And that makes more to stay in sync between versions, if we keep a top menu on the tuts.

I'm still open to better. The main thrust of this was to improve content and remove GhPages hacks.

aeons:
http://http4s.org/versions/ is missing something

PR #1403, opened 2017-09-16
Fix Accept-Charset matching

rossabaker:
`Accept-Charset.qValue` was ignoring the passed in charset and returning the first atom's q-value.

### DIFF
--- a/core/src/main/scala/org/http4s/CharsetRange.scala
+++ b/core/src/main/scala/org/http4s/CharsetRange.scala
@@ -7,13 +7,26 @@ import scalaz.{Equal, Show}
 sealed abstract class CharsetRange extends HasQValue with Renderable {
   def qValue: QValue
   def withQValue(q: QValue): CharsetRange
+
+  @deprecated("Use `isSatisfiedBy` on the `Accept-Charset` header", "0.16.1")
   def isSatisfiedBy(charset: Charset): Boolean
+
+  /** True if this charset range matches the charset.
+    * 
+    * @since 0.16.1
+    */
+  def matches(charset: Charset): Boolean
 }
 
 object CharsetRange extends CharsetRangeInstances {
   sealed case class `*`(qValue: QValue) extends CharsetRange {
     final override def withQValue(q: QValue): CharsetRange.`*` = copy(qValue = q)
+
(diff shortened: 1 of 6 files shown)


core/src/main/scala/org/http4s/headers/Accept-Charset.scala
```diff
@@ -14,7 +14,7 @@ final case class `Accept-Charset`(values: NonEmptyList[CharsetRange]) extends He
   type Value = CharsetRange
 
   def qValue(charset: Charset): QValue = {
-    def specific = values.collectFirst { case cs: CharsetRange.Atom => cs.qValue }
+    def specific = values.collectFirst { case cs: CharsetRange.Atom if cs.isSatisfiedBy(charset) => cs.qValue }
     def splatted = values.collectFirst { case cs: CharsetRange.`*` => cs.qValue }
```
ChristopherDavenport: Just checking as the PR also had this line modified if this was the intention?

rossabaker: `isSatisfiedBy` was wrong, because it would let zeroes slide through and fall back to a splat.  I introduced `matches` in 178d075.

Splatted was skipped because it matches all charsets.  We could make it explicit, but it would essentially be a method call to `if true`.

core/src/main/scala/org/http4s/CharsetRange.scala
```diff
@@ -7,21 +7,39 @@ import scalaz.{Equal, Show}
...
   def withQValue(q: QValue): CharsetRange
+
+  @deprecated("Use `isSatisfiedBy` on the `Accept-Charset` header", "0.16.1")
   def isSatisfiedBy(charset: Charset): Boolean
+
+  /** True if this charset range matches the charset.
+    * 
+    * @since 0.16.1
+    */
+  def matches(charset: Charset): Boolean
```
rossabaker: I'd accept an argument this shouldn't be made public until 0.18, but our compatibility historically runs in just one direction.

rossabaker:
I think Accept-Encoding and Accept-Language need attention as well.

All the Accept* headers need to have a consistency pass, but that can happen in 0.18.

PR #1405, opened 2017-09-16
Pass redacted headers to request and response logger

rossabaker:
I want to add a unit test, but we haven't yet developed any means to capture the log output.


rossabaker:
Oops, this was supposed to be against 0.17.

PR #1406, opened 2017-09-17
Fix Accept-Encoding and Accept-Language

rossabaker:
* `satisfiedBy` on the values makes little sense and is deprecated
* `satisfiedBy` on the headers did not understand priority
* made more consistent with `Accept-Charset`

`LanguageTag` and `ContentCoding` are still broken in that they are both values (in the `Content-*` headers) and ranges (in the `Accept-*` headers).  These should be split into different types, but there's no reasonable way to do that without breaking binary compatibility.  We'll address that in a later version.

### DIFF
--- a/core/mima.sbt
+++ b/core/mima.sbt
@@ -1,10 +1,5 @@
 import com.typesafe.tools.mima.core._
 
-// This is a new introduction
-mimaBinaryIssueFilters ++= Seq(
-  ProblemFilters.exclude[ReversedMissingMethodProblem]("org.http4s.CharsetRange.matches")
-)
-
 // This is massive because we split out the parboiled2 module in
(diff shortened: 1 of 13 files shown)


rossabaker:
This passes MiMa for me locally, which concerns me.  I expected to have to add exceptions for new signatures.

ChristopherDavenport:
Well your concerns were valid, as it doesn't pass MiMa on travis.

rossabaker:
I'm still not sure why it doesn't complain about the new methods in core.
Probably related to being final.

I'm going to see if I can do all this without MiMa exceptions.

bryce-anderson:
Looks good to me so 👍 if travis is happy.

PR #1407, opened 2017-09-18
Add max wait queue for BlazeClient

### LINKED ISSUE #1043: Blaze client config should have a maximum number of waiters
If your connection pool is too small, and you burst too many requests, you get an indefinitely sized queue of waiters.

randallalexander:
This PR is an attempt to resolve #1043.
One thing to note is I had to break API compatibility for PooledHttp1Client.apply if both parameters are being set without being named, but I don't think there is a way around it.  I did consider adding the value to BlazeClientConfig but the other pool setting is not present there so I did not feel it was a good spot to put the value.

### DIFF
--- a/blaze-client/src/main/scala/org/http4s/client/blaze/PooledHttp1Client.scala
+++ b/blaze-client/src/main/scala/org/http4s/client/blaze/PooledHttp1Client.scala
@@ -7,18 +7,26 @@ import cats.effect._
 /** Create a HTTP1 client which will attempt to recycle connections */
 object PooledHttp1Client {
   private val DefaultMaxTotalConnections = 10
+  //Negative value is treated as unbounded.  Default to unbounded for backwards compatibility.
+  private val DefaultMaxWaitConnections = -1
 
   /** Construct a new PooledHttp1Client
     *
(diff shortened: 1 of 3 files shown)


bryce-anderson:
Thanks for the PR! It looks like you're trying to solve a similar issue as @eklavya in PR #1367. I think that PR is almost ready so you're feedback would be appreciated.

randallalexander:
@bryce-anderson you are correct.  The solutions overlap with the difference being the approach.  I will take a closer look at it tomorrow.

rossabaker:
Aren't these orthogonal? This is about bounding the wait queue, and #1367 is about bounding the open connections per host. They're definitely overlapping in touched code (hello, merge conflict!) but I think both valuable.

randallalexander:
That PR also tries to bound the waitqueue in method addToWaitQueue.

randallalexander:
#1367 overlaps with the solution proposed here.  I am going to close this PR for now.  I will reopen if for some reason #1367 does not go through.

rossabaker:
Yes, I was so focused on the max connections per host I'd overlooked the max waiters.  Anyway, we appreciate the contribution, and hope to see more!

PR #1402, opened 2017-09-16
Upgrade to sbt-1.0

rossabaker:
Won't build yet, because it's based on a snapshot of sbt-rig.

Pulls in sbt-tpolecat, which makes our compiler flags stricter, which cleans up our code more.  That's the meat of this.

Fixes #1401.

PR #1409, opened 2017-09-18
Compiled with stricter options

rossabaker:
These are the code changes from #1402, merged with the fixes though
v0.17.1.  Separating these so we don't bitrot waiting on rig.


rossabaker:
It would be better if this introduced sbt-tpolecat, but it sets scalacOptions in a different scope than sbt-rig.  This just introduces the things it found.  We'll get the fuller scalacOptions in place as part of #1402.

PR #1410, opened 2017-09-19
Test which thread pool requests route and run on

rossabaker:
On 0.16.1:
* "route requests on the service executor" passes.
* "execute the service task on the service executor" passes.

On 0.17.1:
* "route requests on the service executor" fails.
* "execute the service task on the service executor" passes on servlets, fails on blaze.

On 0.18.0-M1:
* "route requests on the service executor" fails.
* "execute the service task on the service executor" passes.

Expectations:
* "execute the service task on the service executor" should always pass, or there is no point in this executor at all.
* "route requests on the service executor" gets CPU-bound work off server's own pool.  If the server is operating on threads for I/O, this is desired.  If the server has provided its own thread pool, this is a useless context shift.  I'm not yet certain what's the right behavior, but it has changed unintentionally.


rossabaker:
If we don't shift the routing onto an executor, there's no way to get it off the `ServerChannelGroup` in blaze.  Opting out of the shift can be done with a direct execution context.  The response effect can always be shifted by other means.  I think the 0.16 behavior is correct for blaze.

bryce-anderson:
I agree with your assessment: the ExecutionContext gives us all the flexibility we need regarding which thread to execute stuff on, but only if we use it.

rossabaker:
Jetty has a [ThreadPool](http://www.eclipse.org/jetty/javadoc/9.4.6.v20170531/org/eclipse/jetty/util/thread/ThreadPool.html), which is a specialized `Executor`.  We can pass an `ExecutionContext` to that, but we'd have to lie in some of the methods, but we'd [be in good company](https://github.com/eclipse/jetty.project/blob/eee41174f24d651b493a2d382df6402847f005be/jetty-util/src/main/java/org/eclipse/jetty/util/thread/ExecutorThreadPool.java).

The [ServerConnector](http://docs.oracle.com/javase/8/docs/api/java/util/concurrent/Executor.html?is-external=true) accepts a plain old `Executor`, but then wastes the server's threads.

I think the right answer here is to provide a `withThreadPool`.  `withExecutorService` could be implemented as a lying wrapper that delegates to `withThreadPool`.  And then we stop forking and run on the server's executor, which is designed for this.

rossabaker:
Thread pools in embedded Tomcat are a dystopia and I'm going to bed. 💤

eklavya:
Why in unsafeRun do we not execute on the calling thread but instead block on a latch and schedule somewhere else?

rossabaker:
@eklavya Are you referring to the `Task.fork` that got removed?  I can't recall why it was done that way.

eklavya:
No, in Task.unsafeRun. I didn't understand why we didn't execute in the current thread right away. We are blocking for it to complete anyway.

rossabaker:
Can you point to the line number?  I'm still not seeing it.  (I do agree that the `Task.fork(...).unsafePerformSync` that we had before was silly.)

eklavya:
I am sorry it should be an fs2 question. In fs2:

```scala
def unsafeRun(): A = self.get.run match {
      case Left(e) => throw e
      case Right(a) => a
    }

  def run: A = this match {
    case Now(a) => a
    case _ => {
      val latch = new java.util.concurrent.CountDownLatch(1)
      @volatile var result: Option[A] = None
      runAsync { a => result = Some(a); latch.countDown }
      latch.await
      result.get
    }
  }

def runAsync(cb: A => Unit): Unit =
    listen(a => Trampoline.done(cb(a)))
```

rossabaker:
If this gets approved, I think we can start a release train.

@eklavya Did that conversation on Gitter today help clear things up?

eklavya:
@rossabaker yes, including here for the record. The latch mechanism is required because the Tasks being run by unsafeRun might include an async task and we need to wait for it. If there are no async tasks, the whole thing runs on the calling thread.

PR #1413, opened 2017-09-19
Thread shift 0.18

rossabaker:
Allows the choice of reuse of the Tomcat and Jetty native thread pools, while still permitting a standard Scala ExecutionContext.


rossabaker:
Took this one too far.  The reuse of Tomcat and Jetty native pools will be through the direct execution context.  Going to start over.

PR #1415, opened 2017-09-21
Move parboiled2 to separate repo

rossabaker:
Depends on http4s/parboiled2#1


project/Http4sPlugin.scala
```diff
@@ -122,6 +122,7 @@ object Http4sPlugin extends AutoPlugin {
   lazy val macroCompat                      = "org.typelevel"          %% "macro-compat"              % "1.1.1"
   lazy val metricsCore                      = "io.dropwizard.metrics"  %  "metrics-core"              % "3.2.4"
   lazy val metricsJson                      = "io.dropwizard.metrics"  %  "metrics-json"              % metricsCore.revision
+  lazy val parboiled                        = "org.http4s"             %% "parboiled"                 % "1.0.0-SNAPSHOT"
```
rossabaker: When this publishes, we can lose the -SNAPSHOT and this should pass.

PR #1417, opened 2017-09-22
Remove implicit strategy from StreamApp

rossabaker:
This caused [confusion](https://gitter.im/scala/scala?at=59c15d57cfeed2eb652b4e18) on Gitter.

/cc @jmcardon

### DIFF
--- a/core/src/main/scala/org/http4s/util/StreamApp.scala
+++ b/core/src/main/scala/org/http4s/util/StreamApp.scala
@@ -10,16 +10,17 @@ trait StreamApp {
 
   def stream(args: List[String]): Stream[Task, Nothing]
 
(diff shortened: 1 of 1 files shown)


jmcardon:
@rossabaker Thank you. I am on 0.18 for when this thing happened, but it's the same gist there too.

I had made the mistake of initializing an executionContext inside of the `stream` method in my main method, and the implicit resolution went wack because there's the implicit `TrampolineExecutionContext`.

rossabaker:
Yes, we'll fix it there when we merge up. Was trying to stop it at the source.

I didn't mean to Gitter stalk, but someone told me about the conversation. Wanted to let you know it was being addressed.

PR #1418, opened 2017-09-22
Turn coverage off before publishing

rossabaker:
Hard to test without merging it, but I think this will get @cranst0n unblocked.


rossabaker:
It's just a snapshot... going to merge now and ask forgiveness later.

PR #1420, opened 2017-09-23
Grudgingly disable scalafmt on compile

rossabaker:
This makes me sad, but it's taking my laptop 25 seconds to figure out there's nothing new to compile, on every compile.

I welcome better suggestions, because this is going to trip up new contributors constantly.

### DIFF
--- a/project/Http4sPlugin.scala
+++ b/project/Http4sPlugin.scala
@@ -46,7 +46,17 @@ object Http4sPlugin extends AutoPlugin {
 
     addCompilerPlugin("org.spire-math" % "kind-projector" % "0.9.4" cross CrossVersion.binary),
 
-    scalafmtVersion := "1.2.0"
+    scalafmtVersion := "1.2.0",
+    scalafmt in Test := {
+      (scalafmt in Compile).value
+      (scalafmt in Test).value
+      ()
+    },
+    test in (Test, scalafmt) := {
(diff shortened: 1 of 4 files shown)


aeons:
Maybe add an sbt task that formats everything and runs tests? Like something to run before `git commit`.

project/Http4sPlugin.scala
```diff
@@ -47,7 +47,16 @@ object Http4sPlugin extends AutoPlugin {
...
+    scalafmt in Test := {
+      (scalafmt in Compile).value
+      (scalafmt in Test).value
+      ()
+    },
+    test in (Test, scalafmt) := {
+      (test in (Compile, scalafmt)).value
+      (test in (Test, scalafmt)).value
+      ()
+    }
```
rossabaker: I'm actually kind of surprised these aren't defaults.  If we're going to start making people run scalafmt manually, it's nice if we can give them just one command.

rossabaker:
@aeons We already have `validate`, which tests the formatting.  That's the thing people should run before a PR now, but it doesn't format.  Maybe it should, or maybe we could give it a more active name?  (`unsafeValidate` :trollface:)

rossabaker:
Alright.  Let's roll with this, and we can always recalibrate if this proves no good.

PR #1424, opened 2017-09-26
Deprecate Service and remove MaybeResponse

aeons:
Following discussion on Gittter, this is an experiment to 

1. deprecate the `Service[F, A, B]` type alias and object of the same name in favour of the raw `Kleisli[F, A, B]` type
2. remove the `MaybeResponse[F]` and `Pass[F]` types and changing the `HttpService[F]` type to be `Kleisli[OptionT[F, ?], Request[F], Response[F]]`


core/src/main/scala/org/http4s/package.scala
```diff
@@ -136,4 +77,13 @@ package object http4s { // scalastyle:ignore
       case Left(l) => f(l)
     }
   }
+
+  implicit def http4sHttpServiceSemigroup[F[_]: Monad]: Semigroup[HttpService[F]] =
```
aeons: This is just here to make stuff compile.

core/src/main/scala/org/http4s/package.scala
```diff
@@ -136,4 +77,13 @@ package object http4s { // scalastyle:ignore
...
+    new Semigroup[HttpService[F]] {
+      override def combine(x: HttpService[F], y: HttpService[F]) =
+        Kleisli(request => x(request).orElse(y(request)))
+    }
+
+  def FToOptionT[F[_]: Functor]: FunctionK[F, OptionT[F, ?]] =
```
aeons: Not sure where this should live. It's come in handy a couple of times.

core/src/main/scala/org/http4s/package.scala
rossabaker: Are we aiming to provide a `Semigroup` or a `SemigroupK`?

core/src/main/scala/org/http4s/package.scala
rossabaker: This is a good question for the Cats room.  If we can't get a satisfactory cats answer, then we look to `org.http4s.util`, I think.

core/src/main/scala/org/http4s/package.scala
aeons: In the specific case of `HttpService`, I guess a `Semigroup` works just as well.

This still needs to be manually in scope to work, I just wanted the rest to compile while we figure out what to do.

aeons:
There are quite a few places where `MaybeResponse.orNotFound` is used. That is a pure empty 404 response. The `Response` object has a `def notFound` which takes a request and renders `s"${request.pathInfo} not found"`.

I suggest renaming that one to something like `notFoundFor(request: Request[F])` and adding a `notFound` that returns a 404 with something like `Not found.`

The empty page on hitting a 404 is not always obvious what went wrong, and having the default 404 defined somewhere ensures consistency.

rossabaker:
That sounds reasonable.

bryce-anderson:
Drive-by comment and haven't looked at the details in depth, but I'm cool with the general idea.

blaze-server/src/main/scala/org/http4s/server/blaze/Http2NodeStage.scala
```diff
@@ -187,12 +187,11 @@ private class Http2NodeStage[F[_]](
 
       async.unsafeRunAsync {
         try service(req)
-          .recoverWith(serviceErrorHandler(req).andThen(_.widen[MaybeResponse[F]]))
-          .handleError { _ =>
-            Response[F](InternalServerError, req.httpVersion)
-          }
-          .map(renderResponse(_))
-        catch serviceErrorHandler(req).andThen(_.widen[MaybeResponse[F]])
+          .getOrElse(Response.notFound)
```
rossabaker: Not apropos of this commit, but this needs to look more like `Http1ServerStage` to thread shift properly. 😟

client/src/main/scala/org/http4s/client/middleware/Retry.scala
```diff
@@ -7,9 +7,12 @@ import cats.implicits._
 import fs2._
 import java.time.Instant
 import java.time.temporal.ChronoUnit
+
+import cats.data.Kleisli
```
rossabaker: These imports didn't sort correctly.

core/src/main/scala/org/http4s/instances/FunctionKInstances.scala
```diff
@@ -0,0 +1,12 @@
+package org.http4s.instances
+
+import cats.Functor
+import cats.arrow.FunctionK
+import cats.data.OptionT
+
+trait FunctionKInstances {
```
rossabaker: I am a little reluctant to provide this when it's 100% cats functionality.  Have we talked to the Cats project about whether this would be a welcome function there?

core/src/main/scala/org/http4s/instances/KleisliInstances.scala
```diff
@@ -0,0 +1,16 @@
+package org.http4s
+package instances
+
+import cats._
+import cats.data._
+import cats.implicits._
+
+trait KleisliInstances {
+
+  implicit def http4sKleisliSemigroupK[F[_]: SemigroupK, A]: SemigroupK[Kleisli[F, A, ?]] =
```
rossabaker: And this is what we can eliminate with typelevel/cats#1928.

server/src/main/scala/org/http4s/server/middleware/PushSupport.scala
```diff
@@ -87,21 +90,20 @@ object PushSupport {
       service: HttpService[F],
       verify: String => Boolean = _ => true): HttpService[F] = {
 
-    def gather(req: Request[F], resp: Response[F]): Response[F] =
+    def gather(req: Request[F])(resp: Response[F]): Response[F] =
```
rossabaker: Smart.

server/src/main/scala/org/http4s/server/package.scala
```diff
@@ -21,28 +22,24 @@ package object server {
...
+  type Middleware[F[_], A, B, C, D] = Kleisli[F, A, B] => Kleisli[F, C, D]
 
   object Middleware {
-    def apply[F[_], A, B, C, D](f: (C, Service[F, A, B]) => F[D]): Middleware[F, A, B, C, D] = {
-      service =>
-        Service.lift { req =>
-          f(req, service)
-        }
-    }
+    def apply[F[_], A, B, C, D](f: (C, Kleisli[F, A, B]) => F[D]): Middleware[F, A, B, C, D] =
```
rossabaker: Does anybody even use these?  They're kind of Finagle-like.  I have to stare at it every time I remember it's here to understand how it works.

aeons: @deprecated?

core/src/main/scala/org/http4s/instances/KleisliInstances.scala
aeons: 👍

server/src/main/scala/org/http4s/server/package.scala
ChristopherDavenport: Please, lets get rid of funky type wrappers unless they are neccesary. Feel Similarly towards EntityBody not that anyone is asking.

core/src/main/scala/org/http4s/instances/FunctionKInstances.scala
aeons: I implemented this with an extension method on `Kleisli` instead, so instead of
```scala
kleisliOfF.transform(FToOptionT)
```
it's now
```scala
kleisliOfF.liftOptionT
```

core/src/main/scala/org/http4s/syntax/KleisliSyntax.scala
```diff
@@ -4,13 +4,22 @@ package syntax
...
 import cats.data.{Kleisli, OptionT}
 
-trait KleisliResponseSyntax {
-  implicit def http4sServiceSyntax[F[_]: Functor, A](
+trait KleisliSyntax {
+  implicit def http4sKleisliResponseSyntax[F[_]: Functor, A](
       service: Kleisli[OptionT[F, ?], A, Response[F]]): KleisliResponseOps[F, A] =
     new KleisliResponseOps[F, A](service)
+
+  implicit def http4sKleisliSYntax[F[_]: Functor, A, B](
```
ChristopherDavenport: Weird Capitalization here. Although I thought we might be removing it.

rossabaker: Yes, I'd like to see whether Cats is interested in providing this.  `KleisliResponseOps` is our jam, but this one has nothing to do with HTTP.

aeons: Capitalization was a typo.

I don't disagree that this should be provided somewhere else, but having it in a released version of cats will not happen overnight.

Inlining the functionality is basically `.mapF(OptionT.liftF(_))`. Would that be an acceptable solution?

blaze-server/src/main/scala/org/http4s/server/blaze/Http2NodeStage.scala
```diff
@@ -185,25 +185,27 @@ private class Http2NodeStage[F[_]](
...
-          }
-          .map(renderResponse(_))
-        catch serviceErrorHandler(req).andThen(_.widen[MaybeResponse[F]])
-      } {
-        case Right(_) =>
-          IO.unit
-        case Left(t) =>
-          IO(logger.error(t)("Error rendering response"))
-      }
+      executionContext.execute(new Runnable {
```
aeons: This bit mirrors how Http1 works and should do the same as #1435 did for 0.17.

aeons:
The last couple of commits inlines the `F ~> OptionT[F, ?]` and deprecates `{Authed,Http}Service.lift`.

core/src/main/scala/org/http4s/syntax/KleisliSyntax.scala
```diff
@@ -0,0 +1,25 @@
...
+
+import cats._
+import cats.data.{Kleisli, OptionT}
+
+trait KleisliSyntax {
+  implicit def http4sKleisliResponseSyntax[F[_]: Functor, A](
+      service: Kleisli[OptionT[F, ?], A, Response[F]]): KleisliResponseOps[F, A] =
+    new KleisliResponseOps[F, A](service)
+
+  implicit def http4sKleisliSYntax[F[_]: Functor, A, B](
```
rossabaker: Still miscapitalized

core/src/main/scala/org/http4s/syntax/KleisliSyntax.scala
```diff
@@ -0,0 +1,25 @@
...
+    new KleisliOps[F, A, B](kleisli)
+}
+
+final class KleisliResponseOps[F[_]: Functor, A](self: Kleisli[OptionT[F, ?], A, Response[F]]) {
+  def orNotFound: Kleisli[F, A, Response[F]] =
+    Kleisli(a => self.run(a).getOrElse(Response.notFound))
+}
+
+final class KleisliOps[F[_]: Functor, A, B](self: Kleisli[F, A, B]) {
+  def liftOptionT: Kleisli[OptionT[F, ?], A, B] =
```
rossabaker: I'm still a little hesitant to expose implicit syntax for types we don't own and aren't in our domain, but I see the use case for it.  Would it be better to make this `private[http4s]` until Cats provides a better solution?

core/src/main/scala/org/http4s/HttpService.scala
```diff
@@ -0,0 +1,31 @@
...
+import org.http4s.syntax.kleisli._
+
+object HttpService extends Serializable {
+
+  /**
+    * Lifts a total function to an `HttpService`. The function is expected to
+    * handle all requests it is given.  If `f` is a `PartialFunction`, use
+    * `apply` instead.
+    */
+  def lift[F[_]: Functor](f: Request[F] => F[Response[F]]): HttpService[F] =
```
rossabaker: This is one I thought could be deprecated.

core/src/main/scala/org/http4s/HttpService.scala
```diff
@@ -0,0 +1,31 @@
...
+    Kleisli(f).liftOptionT
+
+  def liftF[F[_]](f: Request[F] => OptionT[F, Response[F]]): HttpService[F] =
```
rossabaker: We call it `liftF` because it <em>lift</em>s a <em>F</em>unction?  I don't have a better name.  Cats calls this `apply`, but that's already taken.  `lift` is confusing for the existing `lift` and for this, because `Kleisli.lift` takes just the `F[B]`.

Does this help with type inference, or could we even deprecate this one and just use Kleisli for things that are already Kleisli functions and leave our `apply` as the DSL for partiality?

core/src/main/scala/org/http4s/instances/KleisliInstances.scala
rossabaker: This is in that same awkward position as `KleisliOps`, but it replicates the syntax we want people to have once that cats issue moves forward.  I'm sad but totally okay with this one.

core/src/main/scala/org/http4s/AuthedService.scala
```diff
@@ -0,0 +1,37 @@
...
+import org.http4s.syntax.kleisli._
+
+object AuthedService {
+
+  /**
+    * Lifts a total function to an `AuthedService`. The function is expected to
+    * handle all requests it is given.  If `f` is a `PartialFunction`, use
+    * `apply` instead.
+    */
+  def lift[F[_]: Functor, T](f: AuthedRequest[F, T] => F[Response[F]]): AuthedService[F, T] =
```
rossabaker: Another one to deprecate.

blaze-server/src/main/scala/org/http4s/server/blaze/Http2NodeStage.scala
ChristopherDavenport: cant that be done via `F.shift(executionContext) >> currentRunMethodMinusUnsafeRun`

rossabaker: I thought the same, and no, because the routing happens outside the task.

```scala
scala> def foo(i: Int): IO[List[String]] = { val x = Thread.currentThread.getName; IO(List(x, Thread.currentThread.getName)) }
e; IO(List(x, Thread.currentThread.getName)) }Thread.currentThread.getNam
<console>:27: warning: parameter value i in method foo is never used
       def foo(i: Int): IO[List[String]] = { val x = Thread.currentThread.getName; IO(List(x, Thread.currentThread.getName)) }
               ^
foo: (i: Int)cats.effect.IO[List[String]]

scala> (Effect[IO].shift(global) >> foo(42)).unsafeRunSync
(Effect[IO].shift(global) >> foo(42)).unsafeRunSync
res7: List[String] = List(run-main-0, scala-execution-context-global-125)
```

I think it could be done with an `F.shift` if the `service` were called inside `F` and then flattened.

rossabaker:
Alright, I'm happy with this in current form, with the possible exception of the existence of `liftF` alias for `Kleisli.apply`.  I wonder if those just add more confusion, like the old `Service` alias.

aeons:
I would be perfectly fine with deprecating `.liftF` as well and instructing people to use `Kleisli.apply`.

ChristopherDavenport:
I agree with the deprecations. Let's make things explicit as this PR suggests.

aeons:
`liftF` is now removed. Should be good to merge on green.

PR #1426, opened 2017-09-27
Appveyor CI

ChristopherDavenport:
Allows for testing on Windows Infrastructure

Using what works, can probably update SBT still but wanted to gather opinions first.

[Local Builds](https://ci.appveyor.com/project/ChristopherDavenport/http4s)


ChristopherDavenport:
We still haven't fixed our arbitraries and their code coverage drift.

rossabaker:
I can't recall, did we have a good theory on which ones we could target and how to stop the drift?

ChristopherDavenport:
Believe it had to do with secure vs not and timeout occuring or not. But I'd have to jump back in to be certain. Was a long time ago now, like 3 months probably.

PR #1429, opened 2017-09-27
Deprecate Monoid[MaybeResponse]

rossabaker:
The presence of this monoid risks implicit resolution that breaks service chaining by evaluating both effects.  This retains binary compatibility but warns people who fall into this trap.

The long-term solution is #1424.

### DIFF
--- a/core/src/main/scala/org/http4s/Message.scala
+++ b/core/src/main/scala/org/http4s/Message.scala
@@ -348,7 +348,7 @@ object MaybeResponse {
         a orElse b
     }
 
-  @deprecated("""We derive `Monoid[Kleisli[Task, Request, MaybeResponse]]` from `Monoid[Task[MaybeResponse]]`, which needs to have the "orElse"-like semantics of `taskInstance`.  The presence of this monoid makes it easy to locate a monoid that evaluates both effects, which is not what we want when chaining `HttpService`s.  Use `taskInstance` instead.""", "0.16.3")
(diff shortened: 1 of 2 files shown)


server/src/main/scala/org/http4s/server/staticcontent/WebjarService.scala
```diff
@@ -49,15 +49,17 @@ object WebjarService {
...
     // Intercepts the routes that match webjar asset names
     case request if request.method == Method.GET =>
       Option(request.pathInfo)
           .map(sanitize)
           .flatMap(toWebjarAsset)
           .filter(config.filter)
           .map(serveWebjarAsset(config, request))
-          .getOrElse(Pass.now)
+        .getOrElse(Pass.now)
+    case _ =>
```
ChristopherDavenport: Love the total Services

core/src/main/scala/org/http4s/Message.scala
```diff
@@ -337,14 +337,21 @@ sealed trait MaybeResponse {
...
+    */
+  val monoidInstance: Monoid[MaybeResponse] =
     new Monoid[MaybeResponse] {
       def zero =
         Pass
       def append(a: MaybeResponse, b: => MaybeResponse) =
         a orElse b
     }
 
+  @deprecated("""We derive `Monoid[Kleisli[Task, Request, MaybeResponse]]` from `Monoid[Task[MaybeResponse]]`, which needs to have the "orElse"-like semantics of `taskInstance`.  The presence of this monoid makes it easy to locate a monoid that evaluates both effects, which is not what we want when chaining `HttpService`s.  Use `taskInstance` instead.""", "0.16.3")
```
ChristopherDavenport: I like this message. Although I'm not sure it screams I may run two separate effects and only return the first and ignore the second entirely. Also breaking resource safety in the process possibly.

ChristopherDavenport:
I think we can ignore the failure as your build lacks the appveyor file for the testing, due to where you branched from.

core/src/main/scala/org/http4s/Message.scala
rossabaker: I hope people have fatal warnings turned on and _that_ screams at them.  But we could make this more stearn, I suppose.

rossabaker:
Yes, we're going to need to figure out the right place import for 0.18 still, until #1424 takes over.

PR #1432, opened 2017-09-27
Release with rig

rossabaker:
I don't imagine this will work the first time, but let's try it.


.travis.yml
```diff
@@ -1,15 +1,23 @@
...
+- 2.10.6
 env:
+  matrix:
+  - SCALAZ_VERSION=7.2.15
+  - SCALAZ_VERSION=7.1.13
   global:
   - HUGO_VERSION=0.26
   - LOGBACK_ROOT_LEVEL=WARN
   - LOGBACK_EXCEPTION_PATTERN=%xThrowable{3}
+  - secure: PsdgpD9rs4MiVydsyV+dKJtPvD2Gg677tFzGlPbo239XX3e+bfEDbzoKTqnQyg3SAnb5R32tKKDzi12lluihPxiWLhAAX15SS51nxQhDoRX4//JMOpnL5+xAneUNMEXO5z/3Gwo/pITWlErAawrdOtWt1BBdi3vNMqRxSb5QLuc=
```
rossabaker: This is the PGP_PASS

project/Http4sPlugin.scala
```diff
@@ -3,12 +3,15 @@ package org.http4s.build
...
 import com.typesafe.tools.mima.plugin.MimaPlugin, MimaPlugin.autoImport._
+import com.typesafe.sbt.SbtPgp.autoImport._
 import org.http4s.build.ScalazPlugin.autoImport._
 import org.http4s.build.ScalazPlugin.scalazVersionRewriters
 import sbtrelease._
 import sbtrelease.ReleasePlugin.autoImport._
+import sbtrelease.ReleaseStateTransformations._
 import scala.util.Properties.envOrNone
 import verizon.build.RigPlugin, RigPlugin._
+import verizon.build.common._
```
rossabaker: Rig is terribly sloppy about scoping of its own release steps. :(

project/Http4sPlugin.scala
```diff
@@ -53,7 +53,26 @@ object Http4sPlugin extends AutoPlugin {
     mimaFailOnProblem := http4sMimaVersion.value.isDefined,
     mimaPreviousArtifacts := (http4sMimaVersion.value map {
       organization.value % s"${moduleName.value}_${scalaBinaryVersion.value}" % _
-    }).toSet
+    }).toSet,
+
+    // Override rig's default of the Travis build number being the bugfix number
+    releaseVersion := { ver =>
```
rossabaker: The idea here is that we will be releasing whenever we push a tag.  Nothing here ensures that the version number matches the tag.  We could calculate a version number from the tag.

Because we push the tag to trigger a release instead of letting a rig release apply the tag, there is no guarantee that the value in version.sbt on the tag will match the published version. 🤕 

We could alternatively publish off (a) certain branch(es), which is how we use rig at Verizon.  The trick there is what triggers a release build vs. a snapshot build?

ChristopherDavenport: I've never tried to sneak a tag into a PR before. But is that something we should be worried about with this?

build.sbt
```diff
@@ -499,4 +498,11 @@ def initCommands(additionalImports: String*) =
     "org.http4s._"
   ) ++ additionalImports).mkString("import ", ", ", "")
 
+// Run by the primary CI job
+addCommandAlias("ci", ";coverage ;clean ;test ;coverageReport ;coverageOff ;makeSite ;mimaReportBinaryIssues")
+
+// Run by secondary builds
+addCommandAlias("ciLite", ";coverageOff ;test ;mimaReportBinaryIssues")
+
+// Run by developers before PR
```
ChristopherDavenport: Probably should be the first command alias so it is most visible.

project/Http4sPlugin.scala
rossabaker: Travis [builds tags](https://travis-ci.org/http4s/http4s/builds/276646059) by default, and provides a `TRAVIS_TAG` environment variable for them.

scripts/travis
```diff
@@ -0,0 +1,87 @@
...
+
+mkdir -p $HOME/bin
+PATH=$HOME/bin:$PATH
+
+SBT_OPTS+=" ++$TRAVIS_SCALA_VERSION"
+
+export PATH
+export SBT_OPTS
+
+if [[ $TRAVIS_JOB_NUMBER == *.1 ]]; then
```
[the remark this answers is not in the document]
rossabaker: This is the first job in the build matrix.  In this case, 2.12.3/7.2.15.

Alternatively, we could test that Scala version starts with 2.12 and Scalaz version starts with 7.2.  That's how older versions of our script used to work.  Then we need to make sure to change it when we bump foundation versions, which we failed to do before.

We've used the "job number 1 gets all the bells and whistles" convention at work for cross builds, and it has worked out pretty well for us.

rossabaker:
So there's still manual work in preparing a release:

1. Change version.sbt.  rig will trim the `-SNAPSHOT` suffix, so this isn't _strictly_ necessary.
2. Make versions.md match.  We could automate this, but haven't.
3. Update the changelog.  Automating this is hard unless we get a lot crankier about pull request comments.  We could maybe automate timestamping it.  Maybe we could cut our changelog over to GitHub releases, but we've been hesitant to do that because the source of truth for that is not in git.
4. Tag the release, which ought to match version.sbt, at least up to snapshot.  Rig is good at turning the version into a tag, but then we don't have a trigger for which builds get released to sonatype and which ones are just snapshots.

rossabaker:
Does appveyor use a wild default platform encoding?

https://ci.appveyor.com/project/ChristopherDavenport/http4s-m0hqu/build/1.0.13/job/3802rum34ot9xts0#L4059

rossabaker:
A half-baked proposal:

> Make versions.md match. We could automate this, but haven't.

We have a build step that generates toml.  We have git.  We could spit out the latest tag from each minor release, and generate that into versions.md.

> Update the changelog. Automating this is hard unless we get a lot crankier about pull request comments. We could maybe automate timestamping it. Maybe we could cut our changelog over to GitHub releases, but we've been hesitant to do that because the source of truth for that is not in git.

We could go back to a convention of a notes/*.md file per release.  We own our own data.  And we could have a release step that puts this in the GitHub releases.  And then GitHub Releases is our changelog.

> Rig is good at turning the version into a tag, but then we don't have a trigger for which builds get released to sonatype and which ones are just snapshots.

We trigger a release build if version.sbt does not end in -SNAPSHOT _and_ (as a sanity check) we have release notes.

Now cutting a release is as simple as providing notes/X.Y.Z.md and setting versions.sbt.  Travis/rig will take care of updating versions.md, publishing the release notes, publishing the release to sonatype, creating a tag, and bumping the version number on the branch.

🤔

rossabaker:
I solved the versions list.  We would just need to be sure to tag before making the site on a release build.

I tried GitHub Releases and deem them unfit for use.  They don't read the description from an annotated tag's body, and any attempt to change the description through the API updates the date, which destroys the timeline.  I guess our sanity check could be to grep the changelog for release notes.

rossabaker:
Going to take a different approach on this.

PR #1433, opened 2017-09-29
Release with rig

rossabaker:
- Backport #1426 to appease Appveyor
- Moves a bunch of yaml and shell into Scala
- `ci`: a universal build target, even locally
- Generates versions.md from tags
- Publishes signed builds, even for snapshots
- If build is not a snapshot, will publish to Sonatype. [untested]


project/Http4sPlugin.scala
```diff
@@ -53,7 +87,117 @@ object Http4sPlugin extends AutoPlugin {
...
+      val buildData: String =
+        s"""
+           |[versions]
+           |"http4s.api" = "$major.$minor"
+           |"http4s.current" = "${version.value}"
+           |"http4s.doc" = "${docExampleVersion(version.value)}"
+           |scalaz = "${scalazVersion.value}"
+           |circe = "${circeJawn.revision}"
+           |cryptobits = "${cryptobits.revision}"
+           |"argonaut-shapeless_6.2" = "1.2.0-M5"
```
ChristopherDavenport: Why is this hard-coded?

project/Http4sPlugin.scala
```diff
@@ -102,6 +246,27 @@ object Http4sPlugin extends AutoPlugin {
...
+    JGit(file).tags.collect {
+      case ref if ref.getName.startsWith("refs/tags/v") =>
+        Version(ref.getName.substring("refs/tags/v".size))
+    }.foldLeft(Map.empty[(Int, Int), Version]) {
+      case (m, Some(v)) =>
+        def toMinor(v: Version) = (v.major, v.subversions.headOption.getOrElse(0))
+        def patch(v: Version) = v.subversions.drop(1).headOption.getOrElse(0)
+        def milestone(v: Version) = v.qualifier match {
+          case Some(q) if q.startsWith("-M") => q.substring(2).toInt
+          case Some(q) if q.startsWith("-RC") => q.substring(3).toInt + 1000000
```
ChristopherDavenport: I may be missing it, whats with the `+ 1000000`

project/Http4sPlugin.scala
```diff
@@ -53,7 +87,117 @@ object Http4sPlugin extends AutoPlugin {
...
+          checkSnapshotDependencies,
+          inquireVersions,
+          setReleaseVersion,
+          tagRelease
+        )
+      else
+        Seq.empty[ReleaseStep],
+
+      Seq[ReleaseStep](
+        runTestWithCoverage,
```
ChristopherDavenport: You mentioned in gitter this broke coverage, why isn't this running that?

project/Http4sPlugin.scala
```diff
@@ -53,7 +87,117 @@ object Http4sPlugin extends AutoPlugin {
...
+    }
+  ) ++ releaseSettings
+
+  val releaseSettings = Seq(
+    // Reset a couple sbt-release defaults that rig changed
+    releaseVersion := { ver => Version(ver).map(_.withoutQualifier.string).getOrElse(versionFormatError) },
+    releaseTagName := s"v${if (releaseUseGlobalVersion.value) (version in ThisBuild).value else version.value}",
+
+    releaseProcess := Seq(
+      if (!isSnapshot.value)
```
ChristopherDavenport: Can these be broken out into individual functions for greater clarity? Functional programmer in me looks at this as practically imperative.

project/Http4sPlugin.scala
rossabaker: I think the main build doesn't depend on it, but it also doesn't depend on cryptobits.  This is not new, but it is curious.

project/Http4sPlugin.scala
rossabaker: I tried earlier but struggled a bit with names.  I'll give it another go.

project/Http4sPlugin.scala
rossabaker: I wanted to sort RCs after milestones.  I should turn this into a tuple and Do It Right.

project/Http4sPlugin.scala
rossabaker: I've said a lot of things.  I _think_ it is running now?

project/Http4sPlugin.scala
rossabaker: I'm not finding it much clearer when I give them names.

project/Http4sPlugin.scala
rossabaker: This one is not actually compiled into the tuts.  Unless it ever is, adding it as a dependency would just download another piece of the internet.

project/Http4sPlugin.scala
rossabaker: See if you like 0fc93f6 better.

ChristopherDavenport:
Caught again by [timeout spec](https://travis-ci.org/http4s/http4s/jobs/281386589#L3646)

I'd recommend to rerun that build, to see if it can pass. As this is such a crucial PR I'd prefer it pass CI.

rossabaker:
The fixes there haven't been backported from 0.18 yet.  I'll restart.

I don't know why appveyor doesn't like this.

rossabaker:
Cool, I guessed right on Appveyor, too!

PR #1411, opened 2017-09-19
Thread shift 0.17

rossabaker:
Builds on #1410, makes behavior like 0.16.  Specifically:

* runs routing and response task on `executorService` on blaze
* runs routing on `executorService` on servlets

PR #1435, opened 2017-09-29
Thread shift the service function in HTTP/2 server

rossabaker:
Same change we made in #1411, but for HTTP/2.

We don't have an HTTP/2 client handy, so I'm not certain how we test this.


rossabaker:
I recommend Gorecki's "Symphony of Sorrowful Songs" to accompany the last ticket, but I'm going o hear something a little more upbeat tonight.

ChristopherDavenport:
0.17 is without the appveyor file yet. So no worries on that front.

aeons:
I have this change in #1424 for 0.18. Do you want me to put it in a separate PR?

PR #1436, opened 2017-10-01
Merge 0.16.3

rossabaker:
There were several conflicts here, so doing this merge as a PR.  Travis' opinion is probably more useful than humans' on this one, but I'll listen to both.


project/Http4sPlugin.scala
```diff
@@ -75,7 +75,7 @@ object Http4sPlugin extends AutoPlugin {
     http4sMimaVersion := {
       version.value match {
         case VersionNumber(Seq(major, minor, patch), _, _) if patch.toInt > 0 =>
-          Some(s"${major}.${minor}.0")
+          Some(s"${major}.${minor}.${patch.toInt - 1}")
```
rossabaker: 1. Scala 2.11 is stricter than 2.12, which we didn't discover until we ran on all branches.
2. What's done is done.

core/src/main/scala/org/http4s/util/io/package.scala
```diff
@@ -83,7 +83,7 @@ package object io {
     captureOutputStream(osCallbackToWriterCallback(f))
 
   private def osCallbackToWriterCallback(f: JWriter => Unit) = { out: OutputStream =>
-    val w = new OutputStreamWriter(out)
+    val w = new OutputStreamWriter(out, StandardCharsets.UTF_8)
```
ChristopherDavenport: 😆 Fix into commented code.

project/Http4sPlugin.scala
ChristopherDavenport: Not sure I follow on this one. We're only checking version to version not from initial binary for the minor?

core/src/main/scala/org/http4s/util/io/package.scala
rossabaker: Well, if it ever becomes uncommented... :)

project/Http4sPlugin.scala
rossabaker: Compatibility should be transitive.  If we haven't broken compatibility from .0, these should be equivalent.  If we have broken compatibility from .0, intentionally or not, then we have to choose between going back to .0 (another breaking change) or moving on with `${patch - 1}`.

rossabaker:
Yes, most of this code is a rerun.  It was a non-trivial merge, so I decided to make a PR and solicit extra opinions.

PR #1437, opened 2017-10-01
Start building for JDK9 to detect problems

rossabaker:
We're not requiring Java 9 anytime soon, but with increased modularization of the JDK, we should be sure we're not depending on more modules than necessary.  This adds a non-publishing build to the matrix that acts as a canary.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,29 @@
 language: scala
 dist: trusty
-jdk: oraclejdk8
-scala:
-- 2.12.3
-- 2.11.11
-- 2.10.6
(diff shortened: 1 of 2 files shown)


rossabaker:
I tried, and failed, to make a 2x3+1 matrix.  Once we get to cats, it's just 3.  So I surrender.  🏳️

rossabaker:
I think we're going to need to retarget master with this, with sbt-1.0.

rossabaker:
#1462 is more achievable.  Closing for that.

PR #1440, opened 2017-10-02
Release v0.17.3

rossabaker:
Merging this should trigger a release of v0.17.3 with sonatype release, tag, and version on this branch bumped to v0.17.4-SNAPSHOT.

### DIFF
--- a/version.sbt
+++ b/version.sbt
@@ -1,1 +1,1 @@
-version in ThisBuild := "0.17.3-SNAPSHOT"
+version in ThisBuild := "0.17.3"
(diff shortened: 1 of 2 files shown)


rossabaker:
The interesting thing to watch will be whether we set off a "Travis forkbomb" when this lands and it pushes a change.  My expectation and hope is that it will immediately kick off a second build of 0.17.4-SNAPSHOT, and then be done.

rossabaker:
[https://travis-ci.org/http4s/http4s/jobs/282341602#L6154] I thought would be fixed by `    git.remoteRepo := "git@github.com:http4s/http4s.git"`.  I'm not sure what we're missing here.

I was able to manually tag and bump the version.  The release looks otherwise successful.

ChristopherDavenport:
Looks like it was attempting to go via https rather than ssh authentication.

rossabaker:
Yes.  I think we just need to change the upstream URL in a before_script step.

PR #1443, opened 2017-10-02
Release v0.18.0-M2

rossabaker:
Contains some changes to Http4sPlugin to let us release milestones.


rossabaker:
I don't think this depends on Verizon/sbt-rig#20.  Merging and hoping.

PR #1444, opened 2017-10-02
Change StreamApp Type from Nothing to Unit

ChristopherDavenport:


rossabaker:
@hejfelix has a proposal to return an exit code from the stream instead of `Nothing` or `Unit`.  I think in this case we'd `take(1)` from the stream so the first emit is the exit code, and I suppose return 0 if the stream terminates successfully.

I could also envision an `ExitCode` newtype so we're not intly typed, with some constants for common exit codes.

aeons:
I think Felix' suggestion makes sense. We could newtype the exit code, although `sys.exit` just takes a regular `Int`.

I might try this out tomorrow.

aeons:
This can be closed with #1467 merged, right?

ChristopherDavenport:
Roger That

PR #1447, opened 2017-10-03
Don't decrypt unless on http4s/http4s

### LINKED ISSUE #1446: Error Building Local Travis Builds
Current build only checks if PR, which causes failures on personal travis builds. We should also check for `TRAVIS_REPO_SLUG` to make sure we only unlock in http4s/http4s

rossabaker:
Fixes #1446, though I'm now drawing a blank which builds have `$TRAVIS_PULL_REQUEST = false` and _aren't_ on that repo slug.

### DIFF
--- a/scripts/helpers
+++ b/scripts/helpers
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 function decrypt_deploy_key() {
-    if [[ $TRAVIS_PULL_REQUEST = "false" ]]; then
+    if [[ $TRAVIS_PULL_REQUEST = "false" && $TRAVIS_REPO_SLUG = "http4s/http4s" ]]; then
         echo "Decrypting deploy key"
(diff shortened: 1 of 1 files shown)


rossabaker:
And on further thought, the repo slug is more pertinent than the pull request.  Maybe this should be a replacement instead of an `&&`.

cquiroz:
Can't really comment. I have no idea what a travis slug is 😄

rossabaker:
Yes, both these checks were present.  I still don't know which build failed because of this, but definitely that decrypt is not going to work in a different repo.

rossabaker:
Hmm, this is probably true back to 0.16.x.

ChristopherDavenport:
Probably :unamused:

rossabaker:
Yeah, I'm trying not to spend too much time maintaining that old code, but I'm trying to at least get the build automation down there.

PR #1455, opened 2017-10-04
Fix non-blocking servlet IO

### LINKED ISSUE #1441: 400 posting JSON with non-blocking servlet IO
See https://github.com/bigknife/http4sSample.  I can reproduce this misbehavior locally.

/cc @bigknife

rossabaker:
We weren't flattening the chunked stream correctly.  We were discarding all but the first byte.

Fixes #1441.

/cc @bigknife

### DIFF
--- a/server/src/test/scala/org/http4s/server/ServerSpec.scala
+++ b/server/src/test/scala/org/http4s/server/ServerSpec.scala
@@ -1,7 +1,9 @@
 package org.http4s
 package server
 
+import java.net.HttpURLConnection
 import java.net.URL
+import java.nio.charset.StandardCharsets
 import fs2.Task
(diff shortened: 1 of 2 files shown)


rossabaker:
`unchunk` doesn't turn `Stream[F, Chunk[A]]` into `Stream[F, A]`.  It turns it into `Stream[F, Chunk[A]]` where each chunk is a single chunk.  And then because the types didn't line up, we mapped `_(0)`, which just took the first byte.

aeons:
Let's never `git blame` where this bug was introduced.

rossabaker:
😆 I think `git blame` is mostly going to blame scalafmt for quite a while.

PR #1458, opened 2017-10-04
Disable publishSigned in private projects

rossabaker:
Our examples projects shouldn't be publishing.  I think it's biting us now because we started [signing our snapshots](https://travis-ci.org/http4s/http4s/jobs/283139624#L4514).

If this fixes it, I'll add the fix in rig and remove it from our private project plugin.

### DIFF
--- a/project/PrivateProject.scala
+++ b/project/PrivateProject.scala
@@ -1,5 +1,6 @@
 package org.http4s.build
 
+import com.typesafe.sbt.pgp.PgpKeys.{publishLocalSigned, publishSigned}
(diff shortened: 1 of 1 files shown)


rossabaker:
Appveyor picked up a bad merge of `ServletIo` that was subsequently fixed.

rossabaker:
Okay, that [wasn't it](https://travis-ci.org/http4s/http4s/jobs/283254815#L4483).  My understanding is that `overwrite` is supposed to be true when it's a snapshot.  I don't know why this doesn't work.

PR #1460, opened 2017-10-04
Release v0.16.4

rossabaker:
Not a very thrilling release, but a good test of our new automation.


rossabaker:
We're moving too fast for Appveyor today.

PR #1464, opened 2017-10-04
Release v0.18.0-M3

rossabaker:
Contains a bunch of diagnostic release steps so we can see why it's not bumping our version.sbt.


rossabaker:
It probably should, though it's just a merge train.  I think it's the first such.

PR #1465, opened 2017-10-05
Make client DSL more similar to server DSL

rossabaker:
* `Http4sClientDsl[F]` is like `Http4sClientDsl[F]`
* Create `io` object that extends `Http4sClientDsl[IO]`, like http4s-dsl has
* Don't extend `Http4sClientDsl[IO]` from client package, like http4s-dsl doesn't.

This is still not quite harmonious: http4s-dsl imports the status codes, whereas client doesn't.

### DIFF
--- a/client/src/main/scala/org/http4s/client/Http4sClientDsl.scala
+++ b/client/src/main/scala/org/http4s/client/Http4sClientDsl.scala
@@ -1,0 +1,35 @@
+package org.http4s
+package client
+
+import cats._
+import org.http4s.Method.{NoBody, PermitsBody}
+import org.http4s.client.impl.{EmptyRequestGenerator, EntityRequestGenerator}
+
+trait Http4sClientDsl[F[_]] {
+  import Http4sClientDsl._
+
+  implicit def http4sWithBodySyntax(method: Method with PermitsBody): WithBodyOps[F] =
+    new WithBodyOps[F](method)
+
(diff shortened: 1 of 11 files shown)


rossabaker:
As for whether this is even the right package, questions for which I have no answers:

- Are we continuing to assume that everyone who is creating requests is using the client?
- Are we continuing to assume that no second DSL will emerge for clients, as rho has for servers?

Discuss.

aeons:
Should it maybe be moved to `org.https.client.dsl.io` as in the server dsl? (which is not in the server package, I know)

rossabaker:
What's the preferred way to avoid a conflict with the io package? `Http4sDslIO`?

It only hurts if we import `dsl._`, not `dsl.io._`, right?  Is there any reason to import `dsl._` anymore?

ChristopherDavenport:
I don't think you ever need dsl._ anymore. Only a specific import or io._

rossabaker:
Given that, are we okay with the name?

I like `client.dsl._`, which makes me start to wonder about `server.dsl._`. 🤔

rossabaker:
I refamiliarized myself with the build failure.  Last week was a long time ago for my old brain.  Yes, I think @aeons's suggestion on the package name resolves @ChristopherDavenport and Travis' concerns on the circe clash.

ChristopherDavenport:
You appeased the test gods, but not the tut gods

PR #1466, opened 2017-10-05
WIP: a better URI

rossabaker:


core/src/main/scala/org/http4s/Scheme.scala
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  override def toString = s"Scheme($value)"
+
+  override def compareTo(other: Scheme): Int =
+    value.compareToIgnoreCase(other.value)
+}
+
+object Scheme {
+  def apply(s: String): Scheme =
```
rossabaker: Should be a macro that validates literals.

core/src/main/scala/org/http4s/Scheme.scala
```diff
@@ -0,0 +1,46 @@
...
+    new Scheme(s)
+
+  def fromString(s: String): ParseResult[Scheme] =
+    Right(new Scheme(s))
+
+  val http = Scheme("http")
+  val https = Scheme("https")
+
+  implicit val http4sInstancesForScheme: Show[Scheme] with Renderer[Scheme] with Order[Scheme] =
```
rossabaker: I'd like to get rid of `Renderable`.  Typeclasses, not inheritance.

rossabaker: This could also be a legal `Semigroup`, but I don't think very usefully.

core/src/main/scala/org/http4s/Scheme.scala
rossabaker: Actually, we need to decide once and for all whether we like macros or StringContexts for this and standardize.

aeons:
This looks like a good start.

Do you want to get it all done before merging, or taking it one component at a time?

rossabaker:
I think we can do one component at a time.  Stand by.

PR #1467, opened 2017-10-05
Change StreamApp to take Stream[F, ExitCode]

aeons:
This is an extension on #1444 and @hejfelix s ideas.

It introduces an `ExitCode` type that wraps an int, with values for success (0) and a generic error (-1).

Naming and location of the type can certainly be discussed.

This works around the dead code warning/error when using the `.serve` stream in a for-comprehension, allowing code like the following to compile with strict scalac flags:

```scala
import cats.effect._
import fs2._
import org.http4s.util.StreamApp.ExitCode

object TestApp extends Test[IO]

class Test[F[_]: Effect] extends StreamApp[F] {

  case class SomeClient(success: Boolean)

  val mkClient: Stream[F, SomeClient] = Stream.emit(SomeClient(true))

  def mkServer(client: SomeClient): Stream[F, ExitCode] = {
    // do something with client
    val _ = client
    BlazeBuilder[F].bindAny().serve
  }

  override def stream(args: List[String], requestShutdown: F[Unit]): Stream[F, ExitCode] =
    for {
      client <- mkClient
      exitCode <- mkServer(client)
    } yield exitCode
}
```

### DIFF
--- a/core/src/main/scala/org/http4s/util/StreamApp.scala
+++ b/core/src/main/scala/org/http4s/util/StreamApp.scala
@@ -84,8 +84,8 @@ object StreamApp {
   sealed abstract case class ExitCode(code: Int)
 
   object ExitCode {
-    def fromInt(code: Int): ExitCode = new ExitCode(code) {}
-    val success: ExitCode = fromInt(0)
-    val error: ExitCode = fromInt(-1)
+    def apply(code: Int): ExitCode = new ExitCode(code) {}
+    val success: ExitCode = apply(0)
+    val error: ExitCode = apply(1)
   }
 }
--- a/core/src/main/scala/org/http4s/util/StreamApp.scala
+++ b/core/src/main/scala/org/http4s/util/StreamApp.scala
@@ -77,15 +77,15 @@ abstract class StreamApp[F[_]](implicit F: Effect[F]) {
   }
 
   def main(args: Array[String]): Unit =
-    sys.exit(doMain(args.toList).unsafeRunSync.code)
+    sys.exit(doMain(args.toList).unsafeRunSync.code.toInt)
 }
 
 object StreamApp {
-  sealed abstract case class ExitCode(code: Int)
+  final case class ExitCode(code: Byte)
 
   object ExitCode {
-    def apply(code: Int): ExitCode = new ExitCode(code) {}
-    val success: ExitCode = apply(0)
-    val error: ExitCode = apply(1)
+    def fromInt(code: Int): ExitCode = ExitCode(code.toByte)
+    val success: ExitCode = ExitCode(0)
+    val error: ExitCode = ExitCode(1)
   }
 }
--- a/core/src/main/scala/org/http4s/util/StreamApp.scala
+++ b/core/src/main/scala/org/http4s/util/StreamApp.scala
@@ -4,14 +4,17 @@ import cats.effect._
 import cats.effect.implicits._
 import cats.implicits._
 import fs2._
+import fs2.async.Ref
 import fs2.async.mutable.Signal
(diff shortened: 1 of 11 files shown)


core/src/main/scala/org/http4s/util/StreamApp.scala
```diff
@@ -4,56 +4,88 @@ import cats.effect._
...
   def main(args: Array[String]): Unit =
-    sys.exit(doMain(args.toList).unsafeRunSync)
+    sys.exit(doMain(args.toList).unsafeRunSync.code)
+}
+
+object StreamApp {
+  sealed abstract case class ExitCode(code: Int)
+
+  object ExitCode {
+    def fromInt(code: Int): ExitCode = new ExitCode(code) {}
```
rossabaker: I think this can be an `apply`.  It's total.  It could also take `code % 256`, which will make it reflect what _really_ gets returned.

core/src/main/scala/org/http4s/util/StreamApp.scala
```diff
@@ -4,56 +4,88 @@ import cats.effect._
...
+    val success: ExitCode = fromInt(0)
+    val error: ExitCode = fromInt(-1)
```
rossabaker: I think of 1 as the generic error code, not -1 (aka 255).  Though I've seen both.  I suppose it would be nice to distinguish apps that died by exception vs. apps that were asked to do something they can't.  Sort of like failure vs. error in a test framework.  But I can't point to a standard for that. 

2 is traditional for command line usage failure and might be worth a constant.

core/src/main/scala/org/http4s/util/StreamApp.scala
```diff
@@ -4,56 +4,88 @@ import cats.effect._
...
+      exitCodeRef.get
   }
 
   def main(args: Array[String]): Unit =
-    sys.exit(doMain(args.toList).unsafeRunSync)
+    sys.exit(doMain(args.toList).unsafeRunSync.code)
+}
+
+object StreamApp {
+  sealed abstract case class ExitCode(code: Int)
```
rossabaker: This seems like it should have an `isSuccess` property, but I don't know who would call it, so maybe not. 🤷‍♂️

tests/src/test/scala/org/http4s/util/StreamAppSpec.scala
```diff
@@ -17,23 +18,29 @@ class StreamAppSpec extends Http4sSpec {
...
     }
 
     "Terminate Server on a Valid Process" in {
-      val testApp = new TestStreamApp(_ => emit("Valid Process").drain)
-      testApp.doMain(List.empty) should returnValue(0)
-      testApp.cleanedUp.get.unsafeRunSync should beTrue
+      val testApp = new TestStreamApp(_ => emit(ExitCode.success))
+      testApp.doMain(List.empty) should returnValue(ExitCode.success)
+      testApp.cleanedUp.get should returnValue(true)
+    }
```
rossabaker: How about a test for a process that returns nothing?

core/src/main/scala/org/http4s/util/StreamApp.scala
aeons: I found this: http://tldp.org/LDP/abs/html/exitcodes.html

I guess most of them doesn't really make sense. I will switch error to 1 instead.

core/src/main/scala/org/http4s/util/StreamApp.scala
aeons: I changed it to an apply. If we don't add the modulus, it might as well be a regular case class, no?

PR #1469, opened 2017-10-05
Allow overwrite of signed snapshots

rossabaker:
This combats [this change in sbt-1.0](https://github.com/sbt/sbt-pgp/blob/5891a1b8729b9fdd2df7e0593bf5822b51605d40/pgp-plugin/src/main/scala-sbt-1.0/Compat.scala#L72).  I think it will allow us to publish signed snapshots, as we do on the 0.13 builds.

If this doesn't work, I'm ripping out signed snapshots, breaking my laptop, and finally opening that aquarium shop.  I am particularly fond of African cichlids.


ChristopherDavenport:
ClientTimeoutSpec failure on Appveyor is obviously tangent to this.

PR #1473, opened 2017-10-08
Sanitize more than two dots as well

aeons:
`org.http4s.server.staticcontent.sanitize` is supposed to protect against getting arbitrary files in static services by replacing `..` with `.`

However, if you pass it `...` it will replace it with `..`

This PR replaces an arbitrary number of periods with a single.

### DIFF
--- a/server/src/main/scala/org/http4s/server/staticcontent/FileService.scala
+++ b/server/src/main/scala/org/http4s/server/staticcontent/FileService.scala
@@ -80,7 +80,7 @@ object FileService {
 
   // Attempts to sanitize the file location and retrieve the file. Returns None if the file doesn't exist.
   private def getFile(unsafePath: String): Option[File] = {
-    val f = new File(sanitize(unsafePath))
+    val f = new File(PathNormalizer.removeDotSegments(unsafePath))
     if (f.exists()) Some(f)
     else None
   }
--- a/server/src/main/scala/org/http4s/server/staticcontent/PathNormalizer.scala
+++ b/server/src/main/scala/org/http4s/server/staticcontent/PathNormalizer.scala
@@ -1,0 +1,91 @@
+package org.http4s.server.staticcontent
+
+import java.lang.StringBuilder
+
+// Adapted from https://github.com/Norconex/commons-lang/blob/c83fdeac7a60ac99c8602e0b47056ad77b08f570/norconex-commons-lang/src/main/java/com/norconex/commons/lang/url/URLNormalizer.java#L429
+object PathNormalizer {
+
+  private def startsWith(b: StringBuilder, str: String): Boolean =
+    b.indexOf(str) == 0
+  private def equalStrings(b: StringBuilder, str: String): Boolean =
+    b.length == str.length && startsWith(b, str)
+  private def deleteStart(b: StringBuilder, str: String): StringBuilder =
+    b.delete(0, str.length)
+  private def replaceStart(b: StringBuilder, target: String, replacement: String): StringBuilder = {
+    deleteStart(b, target)
+    b.insert(0, replacement)
+  }
+  private def removeLastSegment(b: StringBuilder): Unit =
+    b.lastIndexOf("/") match {
+      case -1 => b.setLength(0)
+      case n  => b.setLength(n)
+    }
+
+  def removeDotSegments(path: String): String = {
+    // (Bulleted comments are from RFC3986, section-5.2.4)
+
+    // 1.  The input buffer is initialized with the now-appended path
+    //     components and the output buffer is initialized to the empty
+    //     string.
+    val in  = new StringBuilder(path)
+    val out = new StringBuilder
+
+    // 2.  While the input buffer is not empty, loop as follows:
+    while (in.length > 0) {
+
+      // A.  If the input buffer begins with a prefix of "../" or "./",
+      //     then remove that prefix from the input buffer; otherwise,
+      if (startsWith(in, "../"))
+        deleteStart(in, "../")
+      else if (startsWith(in, "./"))
+        deleteStart(in, "./")
+
+      // B.  if the input buffer begins with a prefix of "/./" or "/.",
+      //     where "." is a complete path segment, then replace that
+      //     prefix with "/" in the input buffer; otherwise,
+      else if (startsWith(in, "/./"))
+        replaceStart(in, "/./", "/")
+      else if (equalStrings(in, "/."))
+        replaceStart(in, "/.", "/")
+
+      // C.  if the input buffer begins with a prefix of "/../" or "/..",
+      //     where ".." is a complete path segment, then replace that
+      //     prefix with "/" in the input buffer and remove the last
+      //     segment and its preceding "/" (if any) from the output
+      //     buffer; otherwise,
+      else if (startsWith(in, "/../")) {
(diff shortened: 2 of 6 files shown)


rossabaker:
My expectation is that three (or more) dots would be left alone.  Only one or two dots is special.  Prior art:

```scala
scala> Paths.get("/tmp/.").toAbsolutePath.normalize
res0: java.nio.file.Path = /tmp

scala> Paths.get("/tmp/..").toAbsolutePath.normalize
res1: java.nio.file.Path = /

scala> Paths.get("/tmp/...").toAbsolutePath.normalize
res2: java.nio.file.Path = /tmp/...
```

aeons:
Yeah, that makes sense. So making it match on `/../` is the way to go?

But with the old sanitize, this happens:
```scala
scala> Paths.get(sanitize("/tmp/../../etc/passwd")).toAbsolutePath.normalize
res0: java.nio.file.Path = /tmp/etc/passwd

scala> Paths.get(sanitize("/tmp/.../.../etc/passwd")).toAbsolutePath.normalize
res1: java.nio.file.Path = /etc/passwd
```

ChristopherDavenport:
Yikes, that second behavior needs to get fixed ASAP. Is that same code present on `0.16`?

bryce-anderson:
I agree with @ChristopherDavenport, we should fix this sooner than later.

bryce-anderson:
For reference, this is how Tomcat does it: https://github.com/apache/tomcat/blob/5d7a579366d48668dd81934a74478336dab06ac7/java/org/apache/tomcat/util/http/RequestUtil.java#L95
They fail to normalize (by returning `null`) if they find a "/../" that attempts to escape the current path, which seems reasonable to do here since we're returning an `OptionT` anyway.

aeons:
Failing on '..' would work as well, which solution do you guys prefer?

Do browsers do path normalization on relative paths in html?

I changed `sanitize` to replace `/../` and `^../` with a single dot, so a fix to the original solution to this problem.

rossabaker:
[RFC 3986 5.2.4](https://tools.ietf.org/html/rfc3986#section-5.2.4) is the closest I can think of to a standard.  Any attempt to `..` up out of the root should result in `/`.

I guess I'm okay with failing on `..` out of the root instead of starting over at the root, but `..` in the middle should still go up.

aeons:
So, normalize the path and see if the absolute path does not share a prefix with the system path defined in the service?

rossabaker:
The RFC has an algorithm you can use if you want to do the "../ past the root resets you at the root" approach.  I think what you propose works only if we want to fail on these.

I find the RFC behavior less surprising, but the Tomcat behavior may be more appropriately scolding for what's a mistake or malicious behavior.  I could go either way.  I think the fix should target release-0.16, regardless.

aeons:
Pattern matching on strings sucks in scala. This implementation matches the two tests in the RFC, but I think it's probably too slow to be usable.

```scala
def removeDotSegments(input: String): String = {
  val initialDots       = raw"\.\.?/(.*)".r
  val initialDotSegment = raw"/\./(.*)".r
  val onlyDotSegment    = raw"/\.".r
  val dotsSegment       = raw"/\.\./?(.*)".r
  val onlyDots          = raw"\.\.?".r
  val segment           = raw"(/?[^/]+)(.*)".r

  def go(input: String, output: List[String]): String = {
    println(input)
    println(output)
    if (input.isEmpty) output.reverse.mkString
    else
      input match {
        case initialDots(rest)       => go(rest, output)
        case initialDotSegment(rest) => go(s"/$rest", output)
        case onlyDotSegment()        => go(s"/", output)
        case dotsSegment(rest)       => go(s"/$rest", output.tail)
        case onlyDots()              => go("", output)
        case segment(segment, rest)  => go(rest, segment :: output)
      }
  }
  go(input, Nil)
}
```

ChristopherDavenport:
Yikes, I don't expect giant strings to be used here though, which is an advantage, and if matches segment it's 6 logic checks. My only worry would be making sure it isn't called frequently.

rossabaker:
Here's a [Java implementation](https://github.com/Norconex/commons-lang/blob/c83fdeac7a60ac99c8602e0b47056ad77b08f570/norconex-commons-lang/src/main/java/com/norconex/commons/lang/url/URLNormalizer.java#L429).  It's quite ugly after being immersed in Scala, but I suspect it's quite fast.

aeons:
It's not that bad, since all the mutability is inside the `StringBuilder`s. That project is licensed under Apache 2.0 as well. I guess that makes it painless to bring in?

```scala
object PathNormalizer {

  private def equalStrings(b: StringBuilder, str: String): Boolean =
    b.length == str.length && b.indexOf(str) == 0
  private def startsWith(b: StringBuilder, str: String): Boolean =
    b.indexOf(str) == 0
  private def replaceStart(b: StringBuilder, target: String, replacement: String): StringBuilder =
    deleteStart(b, target).insert(0, replacement)
  private def deleteStart(b: StringBuilder, str: String): StringBuilder =
    b.delete(0, str.length)
  private def removeLastSegment(b: StringBuilder): Unit =
    b.lastIndexOf("/") match {
      case -1 => b.setLength(0)
      case n  => b.setLength(n)
    }

  def removeDotSegments(path: String): String = {
    // (Bulleted comments are from RFC3986, section-5.2.4)

    // 1.  The input buffer is initialized with the now-appended path
    //     components and the output buffer is initialized to the empty
    //     string.
    val in  = new StringBuilder(path)
    val out = new StringBuilder

    // 2.  While the input buffer is not empty, loop as follows:
    while (in.nonEmpty) {
      // A.  If the input buffer begins with a prefix of "../" or "./",
      //     then remove that prefix from the input buffer; otherwise,
      if (startsWith(in, "../"))
        deleteStart(in, "../")
      else if (startsWith(in, "./"))
        deleteStart(in, "./")

      // B.  if the input buffer begins with a prefix of "/./" or "/.",
      //     where "." is a complete path segment, then replace that
      //     prefix with "/" in the input buffer; otherwise,
      else if (startsWith(in, "/./")) {
        replaceStart(in, "/../", "/")
        removeLastSegment(out);
      } else if (equalStrings(in, "/..")) {
        replaceStart(in, "/..", "/")
        removeLastSegment(out)
      }

      // D.  if the input buffer consists only of "." or "..", then remove
      //      that from the input buffer; otherwise,
      else if (equalStrings(in, ".."))
        deleteStart(in, "..")
      else if (equalStrings(in, "."))
        deleteStart(in, ".")

      // E.  move the first path segment in the input buffer to the end of
      //     the output buffer, including the initial "/" character (if
      //     any) and any subsequent characters up to, but not including,
      //     the next "/" character or the end of the input buffer.
      else
        in.indexOf("/", 1) match {
          case nextSlashIndex if nextSlashIndex > -1 =>
            out.append(in.substring(0, nextSlashIndex))
            in.delete(0, nextSlashIndex)
          case _ =>
            out.append(in)
            in.setLength(0)
        }
    }

    out.toString
  }

}
```

rossabaker:
I have the same urge for property testing here, but am not sure what properties.  A generator that creates a bunch of paths, including `.`, `..`, and `...`, would be a good start.  But what's the property that doesn't echo the implementation?

aeons:
Ah, will PR those other examples.

I couldn't figure out a good way to generate examples for property testing either.

PR #1474, opened 2017-10-08
Set content type and encoding for default notfound response

aeons:


aeons:
It could be useful, but it will be so rare that you would have a pure stream as the body, that `.copy` is good enough IMO.

PR #1475, opened 2017-10-10
Scheme type

rossabaker:
Less exciting as a type as much as a template for shoring up the HTTP model.

* `HttpCodec` is anything that can be parsed and rendered as a round trip, with a law to enforce it.
* `unsafeFromString` that is polluting everything gets moved to the `HttpCodec` type class.
* Move literal support to `StringContext` instead of methods on the companion.
* Macros get centralized in their own object.
* Gets rid of skull-and-crossbones constructors for macros.
* `toString` is for debugging, not serialization.
* Begins to phase out `CaseInsensitiveString`.

### DIFF
--- a/bench/src/main/scala/org/http4s/bench/SchemeBench.scala
+++ b/bench/src/main/scala/org/http4s/bench/SchemeBench.scala
@@ -1,0 +1,22 @@
+package org.http4s
+package bench
+
+import org.openjdk.jmh.annotations._
+
+@Fork(2)
+@Threads(1)
+@State(Scope.Benchmark)
+@BenchmarkMode(Array(Mode.Throughput))
+class SchemeBench {
+  @Benchmark
+  def parseHttp =
+    Scheme.parse("http")
+
+  @Benchmark
+  def parseHttps =
+    Scheme.parse("https")
+
+  @Benchmark
+  def parseMailto =
+    Scheme.parse("mailto")
+}
--- a/core/src/main/scala/org/http4s/Scheme.scala
+++ b/core/src/main/scala/org/http4s/Scheme.scala
@@ -1,0 +1,64 @@
+package org.http4s
+
+import cats.{Order, Show}
+import org.http4s.internal.parboiled2.CharPredicate.{Alpha, Digit}
+import org.http4s.internal.parboiled2.{Parser => PbParser}
+import org.http4s.parser.Http4sParser
+import org.http4s.util.{Writer, hashLower}
+
+/** Each [[org.http4s.Uri]] begins with a scheme name that refers to a
+  * specification for assigning identifiers within that scheme.
+  *
+  * @see https://www.ietf.org/rfc/rfc3986.txt, Section 3.1
+  */
+final class Scheme private (val value: String) extends Comparable[Scheme] {
+  override def equals(o: Any) = o match {
+    case that: Scheme => this.value.equalsIgnoreCase(that.value)
+    case _ => false
+  }
+
+  private[this] var hash = 0
+  override def hashCode(): Int = {
+    if (hash == 0) {
+      hash = hashLower(value)
+    }
+    hash
+  }
+
+  override def toString = s"Scheme($value)"
+
+  override def compareTo(other: Scheme): Int =
+    value.compareToIgnoreCase(other.value)
+}
+
+object Scheme {
+  val http: Scheme = new Scheme("http")
+  val https: Scheme = new Scheme("https")
+
+  def parse(s: String): ParseResult[Scheme] =
+    new Http4sParser[Scheme](s, "Invalid scheme") with Parser {
+      def main = scheme
+    }.parse
+
+  private[http4s] trait Parser { self: PbParser =>
+    def scheme = rule {
+      "https" ~ push(https) |
+        "http" ~ push(http) |
+        capture(Alpha ~ zeroOrMore(Alpha | Digit | "+" | "-" | ".")) ~> (new Scheme(_))
+    }
+  }
+
+  implicit val http4sInstancesForScheme: Show[Scheme] with HttpCodec[Scheme] with Order[Scheme] =
+    new Show[Scheme] with HttpCodec[Scheme] with Order[Scheme] {
+      def show(s: Scheme): String = s.toString
+
+      def parse(s: String): ParseResult[Scheme] =
+        Scheme.parse(s)
+
+      def render(writer: Writer, scheme: Scheme): writer.type =
+        writer << scheme.value
+
+      def compare(x: Scheme, y: Scheme) =
+        x.compareTo(y)
+    }
+}
--- a/core/src/main/scala/org/http4s/UriTemplate.scala
+++ b/core/src/main/scala/org/http4s/UriTemplate.scala
@@ -3,6 +3,7 @@ package org.http4s
 import java.net.URLEncoder
 import org.http4s.Uri.{apply => _, unapply => _, Fragment => _, Path => _, _}
 import org.http4s.UriTemplate._
+import org.http4s.util.StringWriter
 import scala.collection.mutable
 import scala.collection.mutable.ArrayBuffer
 import scala.util.{Failure, Success, Try}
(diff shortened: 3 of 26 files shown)


core/src/main/scala/org/http4s/Scheme.scala
```diff
@@ -0,0 +1,59 @@
...
+import org.http4s.util.{hashLower, Writer}
+
+final class Scheme private (val value: String) extends Comparable[Scheme] {
+  override def equals(o: Any) = o match {
+    case that: Scheme => this.value.equalsIgnoreCase(that.value)
+    case _ => false
+  }
+
+  private[this] var hash = 0
+  override def hashCode(): Int = {
```
rossabaker: This is going to get repetitive.  We could have a CaseInsensitiveStringLike, but I'm trying to move away from inheritance.  I don't think we can abstract the equals without lugging around a manifest, because erasure.

core/src/main/scala/org/http4s/Scheme.scala
```diff
@@ -0,0 +1,59 @@
...
+    if (hash == 0) {
+      hash = hashLower(value)
+    }
+    hash
+  }
+
+  override def toString = s"Scheme($value)"
```
rossabaker: This may be controversial vs. just `value`, but I don't want to encourage `toString` for serialization.  I'm willing to bend on this.

core/src/main/scala/org/http4s/Scheme.scala
```diff
@@ -0,0 +1,59 @@
...
+object Scheme {
+  val http: Scheme = new Scheme("http")
+  val https: Scheme = new Scheme("https")
+
+  def parse(s: String): ParseResult[Scheme] =
+    new Http4sParser[Scheme](s, "Invalid scheme") with Parser {
+      def main = scheme
+    }.parse
+
+  private[http4s] trait Parser { self: PbParser =>
```
rossabaker: The parser moves to the Scheme companion so we can keep the constructor private.  Once we do that, we don't need ugly `skullAndCrossBonesUseThisOnlyFromMacrosAndParsers` type of methods.  There really is _no_ way to construct an unsound `Scheme`.

This still gets mixed into larger parsers so the rule is available.

core/src/main/scala/org/http4s/Scheme.scala
```diff
@@ -0,0 +1,59 @@
+package org.http4s
+
+import cats.{Order, Show}
+ import org.http4s.internal.parboiled2.CharPredicate.{Alpha, Digit}
+import org.http4s.internal.parboiled2.{Parser => PbParser}
+import org.http4s.parser.Http4sParser
+import org.http4s.util.{hashLower, Writer}
+
+final class Scheme private (val value: String) extends Comparable[Scheme] {
```
rossabaker: Intentionally does not override `Renderable`, because I want to encourage type classes over inheritance.

Intentionally not a case class, because the value it could extract does not have similar equality semantics to the type itself.

core/src/main/scala/org/http4s/Scheme.scala
```diff
@@ -0,0 +1,59 @@
...
+
+  override def compareTo(other: Scheme): Int =
+    value.compareToIgnoreCase(other.value)
+}
+
+object Scheme {
+  val http: Scheme = new Scheme("http")
+  val https: Scheme = new Scheme("https")
+
+  def parse(s: String): ParseResult[Scheme] =
```
rossabaker: I still like `parse` because `ParseResult`, but we've argued for `fromString` here.  I could be persuaded either way.

core/src/main/scala/org/http4s/UriTemplate.scala
```diff
@@ -220,7 +221,8 @@ object UriTemplate {
     case _ => ""
   }
 
-  protected def renderScheme(s: Scheme): String = s + ":"
+  protected def renderScheme(s: Scheme): String =
```
rossabaker: This method seems shady to me anyway: why isn't this taking a `Writer`?

core/src/main/scala/org/http4s/parser/Rfc3986Parser.scala
```diff
@@ -6,21 +6,23 @@ import java.net.URLDecoder
...
   // scalastyle:off public.methods.have.type
-  import CharPredicate.{Alpha, Digit, HexDigit}
 
   def charset: Charset
 
   def Uri: Rule1[org.http4s.Uri] = rule { (AbsoluteUri | RelativeRef) ~ EOI }
 
   def AbsoluteUri = rule {
-    Scheme ~ ":" ~ HierPart ~ optional("?" ~ Query) ~ optional("#" ~ Fragment) ~> {
+    scheme ~ ":" ~ HierPart ~ optional("?" ~ Query) ~ optional("#" ~ Fragment) ~> {
```
rossabaker: To build these more complex parsers, we're going to have to mix in the rules that know how to construct the types.  I think this is okay.

core/src/main/scala/org/http4s/util/package.scala
```diff
@@ -54,4 +54,23 @@ package object util {
     "Moved to org.http4s.execution.trampoline, is now merely a ExecutionContextExecutor.",
     "0.18.0-M2")
   val TrampolineExecutionContext: ExecutionContextExecutor = execution.trampoline
+
+  /* This is nearly identical to the hashCode of java.lang.String, but converting
+   * to lower case on the fly to avoid copying `value`'s character storage.
+   */
+  def hashLower(s: String): Int = {
```
rossabaker: I hope this is still lightning fast since I moved it to a function.  I assume yes, but haven't measured.

tests/src/test/scala/org/http4s/SchemeSpec.scala
```diff
@@ -0,0 +1,61 @@
...
+      prop { (a: Scheme, b: Scheme) =>
+        a.value.compareToIgnoreCase(b.value) must_== a.compareTo(b)
+      }
+    }
+  }
+
+  "hashCode" should {
+    "be consistent with equality" in {
+      prop { (a: Scheme, b: Scheme) =>
+        (a == b) ==> (a.## must_== b.##)
```
ChristopherDavenport: Hash code equality is the behavior we want to see? What do we gain through this check?

rossabaker: Equality implies hash equality.  The most visible symptom when this rule is broken is use in hash maps: equal objects can go into different buckets and break the invariants of the set.  (Hash equality does not imply equality, because a type may have more than `Int.MaxValue` inhabitants.)

rossabaker:
Summoning @cquiroz because he's been deeply involved in some of the coding standards discussions, which are relevant here.

bench/src/main/scala/org/http4s/bench/SchemeBench.scala
```diff
@@ -0,0 +1,22 @@
+package org.http4s
+package bench
+
+import org.openjdk.jmh.annotations._
+
+@Fork(2)
```
cquiroz: This probably won't play nice with scala.js but I guess bench won't be supported regardles

core/src/main/scala/org/http4s/Scheme.scala
```diff
@@ -0,0 +1,59 @@
+package org.http4s
+
+import cats.{Order, Show}
+import org.http4s.internal.parboiled2.CharPredicate.{Alpha, Digit}
+import org.http4s.internal.parboiled2.{Parser => PbParser}
+import org.http4s.parser.Http4sParser
+import org.http4s.util.{Writer, hashLower}
+
+final class Scheme private (val value: String) extends Comparable[Scheme] {
```
cquiroz: Maybe a short scaladoc of what `Scheme` is?

core/src/main/scala/org/http4s/Scheme.scala
cquiroz: nice!

tests/src/test/scala/org/http4s/SchemeSpec.scala
cquiroz: I think this is a good test, many times this is forgotten when overriding `equals` and `hashCode`

bench/src/main/scala/org/http4s/bench/SchemeBench.scala
rossabaker: Right.  We also don't run them as part of any build, just compile them.  I haven't figured out "benchmarking as a free service for open source" yet.

rossabaker:
I rebased this one, because I ended up with an unused import on the merge build.  It's the same as what was reviewed, plus the scaladoc comment.

rossabaker:
I want to ship v0.18.0-M4 before this, because there are related changes coming that I don't want v0.18.0-M4 to wait for.

PR #1476, opened 2017-10-10
EntityCodecLaws

rossabaker:
If we have an `EntityEncoder[A]` and an `EntityDecoder[A]`, they ought to make a round trip.

### DIFF
--- a/core/src/main/scala/org/http4s/EntityDecoder.scala
+++ b/core/src/main/scala/org/http4s/EntityDecoder.scala
@@ -156,11 +156,17 @@ trait EntityDecoderInstances {
   implicit def binary[F[_]: Sync]: EntityDecoder[F, Chunk[Byte]] =
     EntityDecoder.decodeBy(MediaRange.`*/*`)(collectBinary[F])
 
+  implicit def byteArrayDecoder[F[_]: Sync]: EntityDecoder[F, Array[Byte]] =
+    binary.map(_.toArray)
+
   implicit def text[F[_]: Sync](
       implicit defaultCharset: Charset = DefaultCharset): EntityDecoder[F, String] =
     EntityDecoder.decodeBy(MediaRange.`text/*`)(msg =>
       collectBinary(msg).map(bs =>
         new String(bs.toArray, msg.charset.getOrElse(defaultCharset).nioCharset)))
+
+  implicit def charArrayDecoder[F[_]: Sync]: EntityDecoder[F, Array[Char]] =
+    text.map(_.toArray)
 
   // File operations // TODO: rewrite these using NIO non blocking FileChannels, and do these make sense as a 'decoder'?
   def binFile[F[_]: MonadError[?[_], Throwable]](file: File)(
--- a/testing/src/main/scala/org/http4s/testing/EntityCodecTests.scala
+++ b/testing/src/main/scala/org/http4s/testing/EntityCodecTests.scala
@@ -1,0 +1,72 @@
+package org.http4s
+package testing
+
+import cats.Eq
+import cats.implicits._
+import cats.effect._
+import cats.effect.laws.util.TestContext
+import cats.effect.laws.util.TestInstances._
+import cats.laws._
+import cats.laws.discipline._
+import org.scalacheck.{Arbitrary, Prop, Shrink}
+import org.typelevel.discipline.Laws
+
+trait EntityCodecLaws[F[_], A] {
+  implicit def effect: Effect[F]
+  implicit def encoder: EntityEncoder[F, A]
+  implicit def decoder: EntityDecoder[F, A]
+
+  def entityCodecRoundTrip(a: A): IsEq[IO[Either[DecodeFailure, A]]] = {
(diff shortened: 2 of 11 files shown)


testing/src/main/scala/org/http4s/testing/EntityCodecTests.scala
```diff
@@ -0,0 +1,72 @@
...
+    val effect = effectF
+    val encoder = entityEncoderFA
+    val decoder = entityDecoderFA
+  }
+}
+
+trait EntityCodecTests[F[_], A] extends Laws {
+  def laws: EntityCodecLaws[F, A]
+
+  implicit def eqDecodeFailure: Eq[DecodeFailure] =
```
rossabaker: I'd like to make this a real instance, but it raises questions about `Eq[Throwable]` on some of the implementations.

core/src/main/scala/org/http4s/EntityDecoder.scala
```diff
@@ -156,12 +156,18 @@ trait EntityDecoderInstances {
   implicit def binary[F[_]: Sync]: EntityDecoder[F, Chunk[Byte]] =
     EntityDecoder.decodeBy(MediaRange.`*/*`)(collectBinary[F])
 
+  implicit def byteArrayDecoder[F[_]: Sync]: EntityDecoder[F, Array[Byte]] =
```
rossabaker: I think this one was requested before and I was surprised we didn't have it.  I don't like people using byte arrays, but people do.

core/src/main/scala/org/http4s/EntityDecoder.scala
```diff
@@ -156,12 +156,18 @@ trait EntityDecoderInstances {
...
+    binary.map(_.toArray)
+
   implicit def text[F[_]: Sync](
       implicit defaultCharset: Charset = DefaultCharset): EntityDecoder[F, String] =
     EntityDecoder.decodeBy(MediaRange.`text/*`)(msg =>
       collectBinary(msg).map(bs =>
         new String(bs.toArray, msg.charset.getOrElse(defaultCharset).nioCharset)))
 
+  implicit def charArrayDecoder[F[_]: Sync]: EntityDecoder[F, Array[Char]] =
```
rossabaker: Not sure this one is so widely used, but does give us symmetry.

core/src/main/scala/org/http4s/EntityEncoder.scala
```diff
@@ -18,23 +18,20 @@ import scala.concurrent.{ExecutionContext, Future}
...
-    * __NOTE:__ The Content-Length header will be generated from the resulting Entity and thus should not be added.
-    */
-  def headers: Headers
-
+/**
+  * @param headers Headers that may be added to a [[Message]]. Examples of such
+  * headers would be Content-Type. __NOTE:__ The Content-Length header will be
+  * generated from the resulting Entity and thus should not be added.
+  */
+final class EntityEncoder[F[_], A] private (
```
rossabaker: This change is so we can derive an `Eq` instance to test lawfulness.  Now that I've done it, I think I can do it without this change.  I'll probably roll this back unless someone really likes it better this way.

testing/src/main/scala/org/http4s/testing/package.scala
```diff
@@ -0,0 +1,17 @@
+package org.http4s
+
+import cats.effect.IO
+import cats.effect.laws.util.TestContext
+import org.scalacheck.Prop
+import scala.util.Success
+
+package object testing {
+  def ioBooleanToProp(iob: IO[Boolean])(implicit ec: TestContext): Prop = {
```
rossabaker: This should arguably be implicit but I want to prove out its usefulness first.

tests/src/test/scala/org/http4s/EntityEncoderSpec.scala
```diff
@@ -127,4 +120,19 @@ class EntityEncoderSpec extends Http4sSpec {
       EntityEncoder[IO, ModelB] must_== w2
     }
   }
+
+  {
+    import org.scalacheck._, Arbitrary.arbitrary
+    implicit def arbitraryIO[A](implicit A: Arbitrary[A]): Arbitrary[IO[A]] =
```
rossabaker: A better version of this exists in cats-effect, but isn't published.  I'm asking about this.

tests/src/test/scala/org/http4s/EntityEncoderSpec.scala
```diff
@@ -127,4 +120,19 @@ class EntityEncoderSpec extends Http4sSpec {
...
+      Arbitrary(arbitrary[A].map(IO.apply(_)))
+    implicit def entityEq: Eq[IO[Entity[IO]]] =
```
rossabaker: Maybe this could integrate with cats-effect's TestContext?  I dunno.

core/src/main/scala/org/http4s/EntityEncoder.scala
rossabaker: I rolled it back.

PR #1477, opened 2017-10-11
Release v0.16.5

rossabaker:


server/src/main/scala/org/http4s/server/staticcontent/package.scala
```diff
@@ -20,6 +20,9 @@ package object staticcontent {
   /** Make a new [[org.http4s.HttpService]] that serves static files from webjars */
   def webjarService(config: WebjarService.Config): HttpService = WebjarService(config)
 
+  @deprecated("Replaced by PathNormalizer.removeDotSegments", "0.16.5")
+  private[staticcontent] val sanitize = PathNormalizer.removeDotSegments _
```
rossabaker: This can be removed again when it rolls downhill to master.

bryce-anderson: Is this just to satisfy mima? Presumably this is only used from within the `staticcontent` package.

rossabaker: Yes.  I'm trying to get out of the business of MiMa exceptions.  Better safe than sorry, and it'll be decrufted on the forthcoming merge to master.

Issue #956, opened 2017-02-23
Publish client for scala-js


rossabaker:
Not sure that a server makes sense, but several have asked for the client.  All the libraries we need are ScalaJS capable.

cquiroz:
I've a fair amount of experience on scala.js. I could tackle this one.

cquiroz:
Checking the code I think we want to port `client` and `blaze-client` at least right?
At least `blaze-client` uses some low level connections that won't be available on scala.js. They should be rewritten to be compatible

rossabaker:
Definitely client.  I assumed blaze-client, but really, anything that can provide that Client interface is what we're really after.

I'd imagine also one of the JSON libraries.  Probably http4s-circe?

ChristopherDavenport:
Why is client non js compatible currently? We should get core client fully js compatible.

rossabaker:
Nobody has ever tried to build any of our modules for js, as far as I know.  But yeah, I think it starts at core, and bubbles up to client, and then to blaze-client or some new alternate implementation.  Follow the dependencies.

ChristopherDavenport:
I'll see if I can force client to compile to js, real quick. I think this would probably, be ...-client after core compiles to js. Then this would be a new client implementation.

rossabaker:
If it's going to be a "blaze client", then there needs to be a build of blaze on Scala JS.

cquiroz:
You can see `client` depending on `server` and then on `core` and interesting dependencies like

`scalaReflect(scalaVersion.value) % "provided"`
`fs2Io`

It will be interesting to port

ChristopherDavenport:
fs2 core is all we use ther and that is scalajs compatible. The problem is reflect which is used for inhaled parboiled and I'm unsure if that can be done. Without which our parsing has difficulties.

rossabaker:
Client only depends on server (and dsl) for testing.  DSL should be fine, but that server will need to be untangled somehow.

fs2-io is used in core.  Does that not have a JS equivalent?

parboiled2 is ported to scala.js, but our inhalation of it is not.  That's going to be needed to get core going.

cquiroz:
`fs2-io` doesn't have a .js equivalent because it uses `File` and `nio`

rossabaker:
I think in that case we'll need to split out the entity encoders, Static File, etc. that depend on that, via the jvm/js/shared structure seen in other projects.

cquiroz:
[...]

PR #1480, opened 2017-10-12
Don't use a mutable map to record Methods

cquiroz:
As discussed on #956 we should try to refactor methods that use mutable structures or reflection. This PR removes the use of a mutable map to record

### DIFF
--- a/core/src/main/scala/org/http4s/Method.scala
+++ b/core/src/main/scala/org/http4s/Method.scala
@@ -5,7 +5,6 @@ import cats.implicits._
 import org.http4s.parser.Rfc2616BasicRules
 import org.http4s.util.{Renderable, Writer}
 import org.http4s.Method.Semantics
-import scala.collection.concurrent.TrieMap
 
 /**
   * An HTTP method.
@@ -43,7 +42,7 @@ object Method extends MethodInstances {
   sealed trait NoBody extends Method
 
   def fromString(s: String): ParseResult[Method] =
-    registry.getOrElse(
+    allByKey.getOrElse(
       s,
       Rfc2616BasicRules
         .token(s)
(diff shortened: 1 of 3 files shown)


core/src/main/scala/org/http4s/Method.scala
```diff
@@ -54,54 +53,55 @@ object Method extends MethodInstances {
...
+  val MKACTIVITY: Method = new Method("MKACTIVITY") with Idempotent
+  val MKCALENDAR: Method = new Method("MKCALENDAR") with Idempotent
+  val MKCOL: Method = new Method("MKCOL") with Idempotent
+  val MKREDIRECTREF: Method = new Method("MKREDIRECTREF") with Idempotent
+  val MKWORKSPACE: Method = new Method("MKWORKSPACE") with Idempotent
+  val MOVE: Method = new Method("MOVE") with Idempotent
+  val OPTIONS: Method = new Method("OPTIONS") with Safe with PermitsBody
+  val ORDERPATCH: Method = new Method("ORDERPATCH") with Idempotent
+  val PATCH: Method = new Method("PATCH") with Default with PermitsBody
+  val POST: Method = new Method("POST") with Default with PermitsBody
```
cquiroz: On the list of methods there is a `PRI` method not present here. It seems to be an http2 method. Any reason it wasn't included here?

core/src/main/scala/org/http4s/Method.scala
```diff
@@ -54,54 +53,97 @@ object Method extends MethodInstances {
...
+  val TRACE: SafeMethodWithBody = new Method("TRACE") with Safe with PermitsBody
+  val UNBIND: IdempotentMethod = new Method("UNBIND") with Idempotent
+  val UNCHECKOUT: IdempotentMethod = new Method("UNCHECKOUT") with Idempotent
+  val UNLINK: IdempotentMethod = new Method("UNLINK") with Idempotent
+  val UNLOCK: IdempotentMethod = new Method("UNLOCK") with Idempotent
+  val UPDATE: IdempotentMethod = new Method("UPDATE") with Idempotent
+  val UPDATEREDIRECTREF: IdempotentMethod = new Method("UPDATEREDIRECTREF") with Idempotent
+  val `VERSION-CONTROL`: IdempotentMethod = new Method("VERSION-CONTROL") with Idempotent
+
+  val all = List(
```
rossabaker: This is going to be easy to forget to update, but that happens basically never.

cquiroz: Right, that's a limitation. We could do an automated tests that loads the spec on xml and verifies but it seems too much effort

rossabaker: Another automated test could be done through reflection, but that test wouldn't run on scala.js, which is part of the point.

rossabaker:
Hmm, we never heard from appveyor on this one.  I'm not worried about Windows bugs on this one, but, weird.

PR #1482, opened 2017-10-12
Add more tests for PathNormalizer

aeons:
I'm not convinced the property tests makes sense.

### DIFF
--- a/server/src/test/scala/org/http4s/server/staticcontent/PathNormalizerSpec.scala
+++ b/server/src/test/scala/org/http4s/server/staticcontent/PathNormalizerSpec.scala
@@ -48,6 +48,8 @@ class PathNormalizerSpec extends Specification with Tables with ScalaCheck {
       val path      = Paths.get(prefix, processed).normalize
 
       path.startsWith(Paths.get(prefix)) must beTrue
+      processed must not contain "./"
+      processed must not contain "../"
     }.setGen(pathGen)
   }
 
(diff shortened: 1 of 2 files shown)


server/src/test/scala/org/http4s/server/staticcontent/PathNormalizerSpec.scala
```diff
@@ -1,18 +1,69 @@
...
+        "/a/b/c/g/./h" ! "/a/b/c/g/h" |
+        "/a/b/c/g/../h" ! "/a/b/c/h" |
+        "/a/b/c/g;x=1/./y" ! "/a/b/c/g;x=1/y" |
+        "/a/b/c/g;x=1/../y" ! "/a/b/c/y" | { (path, expected) =>
+        PathNormalizer.removeDotSegments(path) must_=== expected
+      }
+    }
 
-      PathNormalizer.removeDotSegments(path1) must_=== "/a/g"
+    "remove dot segments correctly in other examples" >> prop { input: String =>
```
rossabaker: This is rarely going to generate `..` and `.`, or even multiple segments.  I think for this to make sense, we need to have a smarter generator.

And if we do have a property test, ensuring there are no more `./` or `../` would be good assertions.

aeons: The prop is called with `setGen`, with a custom generator. It's not perfect, but it does create multiple segments and dots.

aeons:
What's going on here: https://travis-ci.org/http4s/http4s/jobs/287183028#L773

aeons:
I rebased on top of e7043e9fc97dde6efe9006a8ddcdb6c668de0dbc, which is only in the merge to 0.17 and is probably why the build broke.

server/src/test/scala/org/http4s/server/staticcontent/PathNormalizerSpec.scala
rossabaker: Oh, I missed that the first time. 😬

PR #1483, opened 2017-10-12
Fix warnings in documentation

BenFradet:
Those were introduced by f159f5c7f6352fa65051751ecf4c0c239bceacf8

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -261,6 +261,10 @@ lazy val docs = http4sProject("docs")
         examplesWar,
         loadTest
       ),
+    scalacOptions in Tut ~= {
+      // unused params warnings are disabled due to undefined functions in the doc
+      _.filterNot(_ == "-Ywarn-unused:params") :+ "-Xfatal-warnings"
+    },
     scalacOptions in (Compile,doc) ++= {
       scmInfo.value match {
         case Some(s) =>
--- a/docs/src/main/tut/auth.md
+++ b/docs/src/main/tut/auth.md
@@ -5,8 +5,7 @@ title: Authentication
 ---
 
 A [service] is a `Kleisli[F, Request, Response]`, the composable version of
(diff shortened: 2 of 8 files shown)


BenFradet:
I think `scalacOptions in Tut ++= Seq("-Xfatal-warnings")` would do, what do you think?

rossabaker:
sbt-tpolecat [disables these flags](https://github.com/DavidGregory084/sbt-tpolecat/blob/09ab072dae5fc1f85224c8b7c51bc70cf7215b66/src/main/scala/io/github/davidgregory084/TpolecatPlugin.scala#L94-L101) in the console.  I imagine tut is a similar environment, though without fatal warnings, we might too easily ignore warnings like what you fixed here.  I think your suggestion is worth a try, and we can step back if the fatal errors are spurious.

BenFradet:
Ok :+1:, I'll try to have a go tomorrow

rossabaker:
Great, thanks!  I'll merge this if the build is green.

build.sbt
```diff
@@ -261,6 +261,9 @@ lazy val docs = http4sProject("docs")
         examplesWar,
         loadTest
       ),
+    scalacOptions in Tut ~= {
+      _.filterNot(Set("-Ywarn-unused:params", "-Ywarn-unused:patvars")) :+ "-Xfatal-warnings"
```
BenFradet: `-Ywarn-unused:params` is due to undefined functions like: `def deleteTweet(id: Int): IO[Unit] = ???`.

`-Ywarn-unused:patvars` is due to `case request @ _ ...`, don't know how we want to handle those.

rossabaker: Is it because `request` isn't referenced?  We can drop the `request @` in that case.  It only needs to be captured if it's going to be used.

BenFradet: yes, it is, I'll remove the unnecessary ones :+1:

PR #1485, opened 2017-10-13
Various and sundry URI improvements

rossabaker:
* Typed `Fragment`
* Typed `UserInfo` that supports the deprecated and ill-advised password field
* Typed `Port`, also used in `Host` header
* `Host` type.  Subtypes that represent arcane syntax rules of RFC3986 are hidden (i.e., no more `RegHost`).
* Convenience `with`, `without`, and `withOption` methods.


core/src/main/scala/org/http4s/Fragment.scala
```diff
@@ -0,0 +1,55 @@
...
+import org.http4s.util.{UrlCodingUtils, Writer}
+import scala.math.Ordered
+
+/** The fragment identifier component of a [[org.http4s.Uri]] allows indirect
+  * identification of a secondary resource by reference to a primary resource
+  * and additional identifying information.
+  *
+  * @see https://www.ietf.org/rfc/rfc3986.txt, Section 3.5
+  */
+final class Fragment private (val value: String) extends AnyVal with Ordered[Fragment] {
```
rossabaker: I usually end up regretting `AnyVal`, but it seems alright here?

core/src/main/scala/org/http4s/UriTemplate.scala
```diff
@@ -100,7 +100,7 @@ object UriTemplate {
 
   type Path = List[PathDef]
   type Query = List[QueryDef]
-  type Fragment = List[FragmentDef]
+  type FragmentDefs = List[FragmentDef]
```
rossabaker: I'm not sure I like this type alias at all.  I certainly didn't like the old alias.

core/src/main/scala/org/http4s/UriTemplate.scala
```diff
@@ -287,8 +290,8 @@ object UriTemplate {
       case MultiFragmentExp(_) =>
         throw new IllegalStateException("MultiFragmentExp cannot be converted to a Uri")
     }
-    if (elements.isEmpty) ""
-    else elements.mkString(",")
+    if (elements.isEmpty) Fragment.empty
```
rossabaker: I suppose this logic could be moved into `Fragment.apply` now that we have it?

core/src/main/scala/org/http4s/parser/Rfc3986Rules.scala
```diff
@@ -0,0 +1,37 @@
+package org.http4s
+package parser
+
+import org.http4s.internal.parboiled2._
+
+trait Rfc3986Rules extends StringBuilding { self: Parser =>
```
rossabaker: This is slowly replacing `Rfc3986Parser`.  When this series of PRs is done, that class will probably be gone.

core/src/main/scala/org/http4s/Fragment.scala
rossabaker: I don't know if it's right to flatten out this URI structure into the root package.  I think URI could be its own microlibrary, though the `HttpCodec` type class is useful for testing and macros.

cquiroz: I separate `uri` library would be nice perhaps useful in other contexts

rossabaker:
I don't know where this PR stops yet, but if you see something you like, we can merge and continue somewhere else.

core/src/main/scala/org/http4s/Uri.scala
```diff
@@ -27,25 +30,40 @@ final case class Uri(
...
       else s"$path$encoded"
     copy(path = newPath)
   }
 
+  def withAuthorityOption(authority: Option[Authority]): Uri = copy(authority = authority)
+  def withAuthority(authority: Authority): Uri = withAuthorityOption(Some(authority))
+  def withoutAuthority: Uri = withAuthorityOption(None)
+
   def host: Option[Host] = authority.map(_.host)
+  def withHostOption(host: Option[Host]) =
```
rossabaker: Without looking at the implementation, what does this do...

- when there is an authority
- when there is no authority

core/src/main/scala/org/http4s/Uri.scala
```diff
@@ -27,25 +30,40 @@ final case class Uri(
...
+    withAuthorityOption(host.fold(Option.empty[Authority])(h =>
+      Some(authority.fold(Authority(host = h))(_.withHost(h)))))
+  def withHost(host: Host) = withHostOption(Some(host))
+
   def port: Option[Int] = authority.flatMap(_.port)
+  def withoutPort: Uri = withAuthorityOption(authority.map(_.withoutPort))
```
rossabaker: No `withPort` or `withPortOption` because there might not be an authority to attach it to.

core/src/main/scala/org/http4s/Uri.scala
```diff
@@ -193,6 +193,57 @@ object Uri extends UriFunctions {
       }
   }
 
+  final class Port private (val toInt: Int) extends AnyVal with Ordered[Port] {
```
rossabaker: This one is more useful even outside a `URI` and should maybe move to the top level.  It wanted to be part of our `Host` header, and could be used in various places on the `Request` as well.  And then `ServerBuilder` would also be a natural fit.

core/src/main/scala/org/http4s/Uri.scala
```diff
@@ -319,6 +318,105 @@ object Uri extends UriFunctions {
...
+
+    final override def hashCode(): Int = value.##
+
+    override final def toString = s"Host($value)"
+  }
+
+  object Host {
+    private def fromName(name: String) = new Host {
+      def inetAddress =
+        try Right(InetAddress.getByName(name))
```
rossabaker: This can block a thread, which would be sad.  I don't know how best to discourage calling this.

core/src/main/scala/org/http4s/Uri.scala
```diff
@@ -319,6 +318,105 @@ object Uri extends UriFunctions {
       }
   }
 
+  sealed abstract class Host {
+    def inetAddress: Either[UnknownHostException, InetAddress]
+    def value: String
+    def render(w: Writer): w.type
+
+    final override def equals(o: Any) = o match {
+      case that: Host => this.value.equalsIgnoreCase(that.value)
```
rossabaker: This should probably compare the encoded values, or else one could do stupid tricks with percent encoding.  Alternatively, we could reintroduce private subclasses.

rossabaker: But then we'd need to normalize, too.  Bah.

core/src/main/scala/org/http4s/Uri.scala
```diff
@@ -319,6 +318,105 @@ object Uri extends UriFunctions {
...
+      def render(w: Writer): w.type = w << value
+    }
+
+    private def fromIPv6(address: InetAddress) = new Host {
+      def inetAddress = Right(address)
+      def value = s"[${util.toAddrString(address)}]"
+      def render(w: Writer): w.type = w << value
+    }
+
+    private def fromIPvFuture(s: String) = new Host {
```
rossabaker: I added support for this, but didn't test it. 😫

core/src/main/scala/org/http4s/util/Renderable.scala
```diff
@@ -60,6 +61,16 @@ object Renderer {
         case Right(b) => rb.render(writer, b)
       }
   }
+
+  implicit val inet4AddressRenderer: Renderer[Inet4Address] = new Renderer[Inet4Address] {
```
rossabaker: I don't think we needed these.  And if we do, it should use the method added to util ported from Guava.

server/src/main/scala/org/http4s/server/middleware/VirtualHost.scala
```diff
@@ -29,10 +30,10 @@ object VirtualHost {
   def exact[F[_]](
       service: HttpService[F],
       requestHost: String,
-      port: Option[Int] = None): HostService[F] =
+      port: Option[Uri.Port] = None): HostService[F] =
```
rossabaker: I don't think we can deprecate old versions because after erasure it's the same signature.

server/src/main/scala/org/http4s/server/middleware/VirtualHost.scala
```diff
@@ -69,11 +72,14 @@ object VirtualHost {
...
-              h.copy(port = req.uri.port.orElse(req.isSecure.map(if (_) Uri.Port.https else Uri.Port.http)))
+              h.copy(
+                port =
+                  req.uri.port.orElse(req.isSecure.map(if (_) Uri.Port.https else Uri.Port.http)))
           }
           (first +: rest).toVector
             .collectFirst { case HostService(s, p) if p(host) => s(req) }
-            .getOrElse(OptionT.liftF(Response[F](NotFound).withBody(s"Host '$host' not found.")))
+            .getOrElse(
+              OptionT.liftF(Response[F](NotFound).withBody(s"Host '${host.value}' not found.")))
```
rossabaker: This echoes request information to the response, which we don't do because of reflection attack paranoia.  I'm not sure why this doesn't return None anyway.

core/src/main/scala/org/http4s/internal/Macros.scala
```diff
@@ -10,11 +10,47 @@ private[http4s] object Macros {
...
     val s = literalString(c)
-    Scheme.parse(s) match {
-      case Right(_) => c.Expr(q"""org.http4s.HttpCodec[org.http4s.Scheme].parseOrThrow($s)""")
+    Uri.Fragment.parse(s) match {
+      case Right(_) => c.Expr(q"""org.http4s.HttpCodec[org.http4s.Uri.Fragment].parseOrThrow($s)""")
+      case Left(e) => throw e
+    }
+  }
+
+  def host(c: Context)(): c.Expr[Uri.Host] = {
```
rossabaker: @ChristopherDavenport brings up that these can interpolate.  I think that's true, for some types, if we can determine that the surrounding context is valid.  That would make these macros super powerful, but it's going to require a lot of careful thought.

An example for this one: an arbitrary string is a `RegHost`, but we can't allow it if the host starts with `[` or `]`, which denotes an IPv6 or IPvFuture address.

tests/src/test/scala/org/http4s/UriSpec.scala
```diff
@@ -233,10 +220,10 @@ http://example.org/a file
       Uri(Some(scheme"mailto"), path = "John.Doe@example.com").toString must_== ("mailto:John.Doe@example.com")
     }
 
-    "render an URL with username and password" in {
+    "render a URL with username and password" in {
       Uri(
         Some(Scheme.http),
-        Some(Authority(Some(userInfo"username:password"), host"some.example.com", None)),
+        Some(authority"username:password@some.example.com"),
```
rossabaker: There are more of these, but I got bored porting them.  I think most could use the URI macro, which is what most people will use, and what I really want to test hard.

ChristopherDavenport:
```
[tut] compiling: /home/travis/build/http4s/http4s/docs/src/main/tut/uri.md
[tut] *** Error reported at /home/travis/build/http4s/http4s/docs/src/main/tut/uri.md:44
// <console>:19: error: value RegName is not a member of object org.http4s.Uri
//          authority = Some(Uri.Authority(host = Uri.RegName("http4s.org"))),
//                                                    ^
[tut] *** Error reported at /home/travis/build/http4s/http4s/docs/src/main/tut/uri.md:46
// <console>:7: error: ')' expected but '.' found.
//        template.toUriIfPossible
```

Once again the great god Tut-enpolecat strikes down the non-believer.

rossabaker:
I'm abandoning this effort.  I think several of us were impressed with the URI interpolator in the sttp presentation.  We should be looking to that for URI reform.

PR #1486, opened 2017-10-16
Add a missing import

olliahonen:
I don't think the example will work without this import


rossabaker:
This is a good question.  I don't see how this worked without this import, but the build should have failed.  A good way to examine the tut would be to follow it in a REPL.  I don't have time this morning, but I'll try to get to this later today unless someone beats me to it.

aeons:
I think tut runs on the entire file, and since that import already was in one of the other examples, it didn't fail.

PR #1488, opened 2017-10-17
Add SemigroupK instance for EntityDecoder

jadlr:
I think it would be nice if we could combine decoders using `<+>` from `SemigroupK`. If you also like it I'll update the docs. I left `orElse` for backwards-compatibility.


core/src/main/scala/org/http4s/EntityDecoder.scala
```diff
@@ -52,7 +52,7 @@ trait EntityDecoder[F[_], T] { self =>
     * @param other backup [[EntityDecoder]]
     */
   def orElse[T2 >: T](other: EntityDecoder[F, T2])(implicit F: Functor[F]): EntityDecoder[F, T2] =
```
reactormonk: I'd add a note to use `<+>` in the doc. Or go all the way, and make use of a deprecation annotation.

core/src/main/scala/org/http4s/EntityDecoder.scala
```diff
@@ -74,6 +74,39 @@ object EntityDecoder extends EntityDecoderInstances {
   /** summon an implicit [[EntityEncoder]] */
   def apply[F[_], T](implicit ev: EntityDecoder[F, T]): EntityDecoder[F, T] = ev
 
+  implicit def semigroupKForEntityDecoder[F[_]: Functor]: SemigroupK[EntityDecoder[F, ?]] =
```
reactormonk: Since we're talking lawful instances here, do you think it would be reasonable to add a few property tests for it?

jadlr: Not really sure on how to do that. Could you guide me to an already existing property test, so I can have a look?

reactormonk: There should be a few around: https://github.com/http4s/http4s/search?utf8=%E2%9C%93&q=Laws&type= I'd use this law: https://github.com/typelevel/cats/blob/155f7f534993c30d6e757de990330ac796dad5da/laws/src/main/scala/cats/laws/SemigroupKLaws.scala

jadlr: Thanks! I'll look into it later.

jadlr: I'm really sorry, but I didn't get that to work in my first attempt. I won't have time to look into this in more depth in the couple of days.

ChristopherDavenport: Nothing to apologize for. Would you mind if I took a look in the meantime?

jadlr: I don't mind. Feel free to give it a shot! Thank you @ChristopherDavenport

rossabaker: Is either @jadlr or @ChristopherDavenport still interested in this?

jadlr: I think it would still be nice to have this, but I failed to write a property test for it. It is tested by several other test, though. That said, I guess it's not a very important change :smiley:

rossabaker: I took a poke at this tonight.  I think we need an `Arbitrary[Message[F]]` so we can have an `Arbitrary[EntityDecoder[F, ?]]`.

rossabaker: It needs a `Cogen[Message[F]]`, and a whole bunch more arbitraries and cogens.  Which are all probably valuable.  I'm [getting close](https://github.com/rossabaker/http4s/tree/semigroupk-entity-decoder), but sleepy.

rossabaker: This is getting messy.  I got everything but an `Eq[EntityDecoder[IO, ?]]`.  The best I can think to do there is [cheat like Cats did on Kleisli](https://github.com/typelevel/cats/blame/5861210bec55b000d1e4962f4002edfdea627c0b/tests/src/test/scala/cats/tests/KleisliSuite.scala#L16), but then we need an `Arbitrary[Message[F]]`, which is pretty easy to generate but much less easy to generate legally.

rossabaker:
Well, I accidentally pulled this commit into master while running the merge train.  The good news is, I have a passing test for its laws. :)

PR #1489, opened 2017-10-17
Upgrade fs2 to 0.10.0-M7

aeons:
I have taken the easy way out and just changed `Sync` to `Effect` wherever the compiler complained. I guess we could use the `runSync` variants in some places, but I do not like the chance of a runtime error if there's an async bit in there.

There are a couple of tests that time out, that I haven't been able to figure out.


aeons:
This might have something to do with fs2-reactive-streams not being on M7, there's a pending PR for that now, will see how it works with that merged.

rossabaker:
Well, we're about to find out. :)

rossabaker:
Okay, so we solved the hang.  I should have asked earlier: what happened in fs2 that requires `Effect` where lesser type classes worked before?  (I think this is probably right, but I want to understand it.)

aeons:
They moved the Effect constraint from the operations on the streams to the `.run*` as I understand it.

aeons:
Yes, exactly.

PR #1495, opened 2017-10-24
Move Exit Code to Util

ChristopherDavenport:


ChristopherDavenport:
Just an FYI, I ran `sbt ci` and it was clean. So that error surprises me.

cquiroz:
my 2 cents do do `clean` as part of `sbt ci`

PR #1497, opened 2017-10-24
Add clean as initial step to sbt ci

ChristopherDavenport:
Prevents false negatives locally by cleaning before the operation is run.


rossabaker:
Merged via cherry-pick.

PR #1498, opened 2017-10-26
Add Partial unification for Clearer Documentation

ChristopherDavenport:
People are running into issues as this is not added to their applications. We should be more transparent about the need to add this.

### DIFF
--- a/docs/src/main/tut/service.md
+++ b/docs/src/main/tut/service.md
@@ -22,6 +22,8 @@ libraryDependencies ++= Seq(
   "org.http4s" %% "http4s-blaze-server" % http4sVersion,
   "org.http4s" %% "http4s-blaze-client" % http4sVersion
 )
+
+scalacOptions ++= Seq("-Ypartial-unification")
 ```
 
 This tutorial is compiled as part of the build using [tut].  Each page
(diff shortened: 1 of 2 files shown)


ChristopherDavenport:
I believe that is in progress currently.

docs/src/main/tut/service.md
```diff
@@ -22,6 +22,8 @@ libraryDependencies ++= Seq(
   "org.http4s" %% "http4s-blaze-server" % http4sVersion,
   "org.http4s" %% "http4s-blaze-client" % http4sVersion
 )
+
+scalacOptions ++= Seq("-Ypartial-unification")
```
rossabaker: There's a [plugin](https://github.com/fiadliel/sbt-partial-unification) that covers a few more cases.  Namely, it uses the plugin for versions of Scala that predate the flag.  Would we be better using that?

rossabaker: The plugin also works on 2.10.6, though that only helps a subset of the subset on Scalaz.  Yeah, I'll go along with this.

I might put in a plug for sbt-tpolecat, which does this _and_ more.

ChristopherDavenport: I recommend another PR. That seems outside of the scope of this PR and could get some debate.

PR #1499, opened 2017-10-26
Make TransferCoding immutable

cquiroz:
I'm trying to remove the use of `Registry` as it creates troubles on `scala.js` and it is not very sound anyway.

In this PR I'm removing `Registry` from `TransferCoding`. Any reason why this was mutable? Should end users be able to add their own `TransferCoding` definitions?

I'm also making the conversion from `CaseInsensitiveString` explicit and added the `Eq` instance

### DIFF
--- a/core/src/main/scala/org/http4s/TransferCoding.scala
+++ b/core/src/main/scala/org/http4s/TransferCoding.scala
@@ -20,21 +20,23 @@ package org.http4s
 
 import org.http4s.syntax.string._
 import org.http4s.util._
+import cats.Eq
 
 final case class TransferCoding private (coding: CaseInsensitiveString) extends Renderable {
   override def render(writer: Writer): writer.type = writer.append(coding.toString)
 }
 
-object TransferCoding extends Registry {
-  type Key = CaseInsensitiveString
-  type Value = TransferCoding
+object TransferCoding {
+  implicit class CI2TransferCoding(val ci: CaseInsensitiveString) extends AnyVal {
+    def toTransferCoding: TransferCoding = TransferCoding(ci)
+  }
 
-  implicit def fromKey(k: CaseInsensitiveString): TransferCoding = new TransferCoding(k)
+  implicit val eq: Eq[TransferCoding] = Eq.fromUniversalEquals
 
   // http://www.iana.org/assignments/http-parameters/http-parameters.xml#http-parameters-2
-  val chunked = registerKey("chunked".ci)
-  val compress = registerKey("compress".ci)
-  val deflate = registerKey("deflate".ci)
-  val gzip = registerKey("gzip".ci)
-  val identity = registerKey("identity".ci)
+  val chunked = TransferCoding("chunked".ci)
+  val compress = TransferCoding("compress".ci)
+  val deflate = TransferCoding("deflate".ci)
+  val gzip = TransferCoding("gzip".ci)
+  val identity = TransferCoding("identity".ci)
 }
--- a/core/src/main/scala/org/http4s/TransferCoding.scala
+++ b/core/src/main/scala/org/http4s/TransferCoding.scala
@@ -20,10 +20,16 @@ package org.http4s
 
 import cats.{Order, Show}
 import cats.data.NonEmptyList
-import org.http4s.util._
+import org.http4s.util.{Renderable, Writer}
 import org.http4s.parser.{Http4sParser, Rfc2616BasicRules}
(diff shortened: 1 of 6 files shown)


rossabaker:
I am strongly in favor of removing the registry.

Not every `CaseInsenstiveString` is a valid `TransferCoding`.  It has to be a token, followed by parameters (which we don't currently model).  If we continue to punt on the parameters, the problem looks exactly like `Scheme`.  What would you think about making the type look like `Scheme`?

cquiroz:
will do. Any reason `Scheme` isn't a `case class`?

rossabaker:
I thought this property was strange for a case class, but maybe it's okay?

```
scala> val s0 = scheme"http"; val s1 = scheme"HTTP"
s0: org.http4s.Scheme = Scheme(http)
s1: org.http4s.Scheme = Scheme(HTTP)

scala> (s0 == s1) == (s0.value == s1.value)
res5: Boolean = false
```

If `value` were still a `CaseInsensitiveString`, it would certainly be a case class.  But then we'd be wrapping the underlying String twice.

🤷‍♂️

cquiroz:
Ok I updated as suggested making `TransferCoding` a `sealed abstract case class`

core/src/main/scala/org/http4s/TransferCoding.scala
```diff
@@ -18,23 +18,52 @@
...
+      }
+
+      def main = rule {
+        oneOrMore(codingRule).separatedBy(ListSep) ~> { codes: Seq[TransferCoding] =>
+          NonEmptyList.of(codes.head, codes.tail: _*)
+        }
+      }
+    }.parse
+
+  implicit val http4sInstancesForTransferCoding: Show[TransferCoding] with Order[TransferCoding] =
```
cquiroz: `Scheme` also implements `HttpCodec`. Should I have it in `TransferCoding` as well?

rossabaker: I think so.  That shows that any value's serialization can be parsed to an equal value.  It's a nice property.

core/src/main/scala/org/http4s/TransferCoding.scala
```diff
@@ -18,23 +18,52 @@
...
 package org.http4s
 
-import org.http4s.syntax.string._
+import cats.{Order, Show}
+import cats.data.NonEmptyList
 import org.http4s.util._
+import org.http4s.parser.{Http4sParser, Rfc2616BasicRules}
 
-final case class TransferCoding private (coding: CaseInsensitiveString) extends Renderable {
+sealed abstract case class TransferCoding private (coding: String) extends Renderable {
```
rossabaker: This could also extend `Ordered` for better non-Cats Scala integration.

I've been phasing out `Renderable` in favor of a `Renderer`: ad hoc polymorphism over inheritance.

core/src/main/scala/org/http4s/TransferCoding.scala
```diff
@@ -18,23 +18,52 @@
...
+  val gzip: TransferCoding = new TransferCodingImpl("gzip")
+  val identity: TransferCoding = new TransferCodingImpl("identity")
+
+  /**
+    * Parse a list of Transfer Coding entries
+    */
+  def parseList(s: String): ParseResult[NonEmptyList[TransferCoding]] =
+    new Http4sParser[NonEmptyList[TransferCoding]](s, "Invalid Transfer Coding")
+    with Rfc2616BasicRules {
+      private def codingRule = rule {
```
rossabaker: We've lost the ability to create a custom `TransferCoding`.  Which probably rarely if ever happens in the wild, and isn't understood by any of our server or client implementations, so it's not a big deal.  But it's legal in the spec.

cquiroz: Right, was that the reason to have it as a `Registry`? To support custom ones?

core/src/main/scala/org/http4s/TransferCoding.scala
cquiroz: This is used in a `Header` which leans to `Renderable` classes. However it is a good goal, I’ll try to use `Renderer` in the header

cquiroz:
Do you think I should try to get `Header` to work with `Renderer` rather than `Renderable`. That maybe better in a separate PR

cquiroz:
With 2 approvals I think i'm ready to merge

PR #1508, opened 2017-11-01
cats-1.0-RC1

### LINKED ISSUE #1507: Cats 1.0.0-RC1 Release
We should get a release out when this is available for the Cats Release Candidate dealing with any errors so that we are proactive.

aeons:
This is a WIP that depends on a lot of dependencies getting updated to cats 1.0-RC1.

So far cats-effect is updated and fs2 0.10 snapshot has merged the updates as well.

Will close #1507 when complete.

**Dependencies**
- [x] cats-effect
- [x] fs2
- [x] jawn-fs2
- [x] fs2-reactive-streams
- [x] circe (once https://github.com/circe/circe/pull/773 is merged)


aeons:
This won't build before `circe-0.9.0-M2` is released, but should work once it is, so if someone wants to restart the build at that point, feel free.

agustafson:
Thanks for looking at it @aeons - will be great once it's released. Knew I shouldn't have used the bleeding edge of everything ...

core/src/main/scala/org/http4s/CharsetRange.scala
```diff
@@ -46,8 +46,9 @@ object CharsetRange extends CharsetRangeInstances {
 }
 
 trait CharsetRangeInstances {
-  implicit val CharacterSetEqual: Eq[CharsetRange] =
+  implicit val http4sEqForCharsetRange: Eq[CharsetRange] =
```
rossabaker: Nice.  I wanted to come back and get these up to Cats standards.

rossabaker:
I'm going to go ahead an merge this so people at least have a snapshot.  We've probably got a few hours before the other loose ends get tied up in case anyone wants to stop the presses, so reviews still encouraged.

PR #1509, opened 2017-11-01
Change type parameter order in Auth to trigger partial unification

### LINKED ISSUE #1506: Service composition syntax with <+> doesn't work for AuthedService
Some people have already stumbled upon this in the wild:  `<+>` syntax doesn't work for AuthedService (also referenced in #1501).

The issue boils down to 
```scala
 type AuthedService[F[_], T] = Kleisli[OptionT[F, ?], AuthedRequest[F, T], Response[F]]
````
having `T` on the right, which prevents partial unification from triggering.

Minimisation:

```scala
object Gotcha {
  import cats._, data._, implicits._

  implicit def http4sKleisliSemigroupK[F[_]: SemigroupK, A]: SemigroupK[Kleisli[F, A, ?]] =
    new SemigroupK[Kleisli[F, A, ?]] {
      def combineK[B](x: Kleisli[F, A, B], y: Kleisli[F, A, B]) =
        Kleisli(req => x(req) <+> y(req))
    }

  case class Req[F[_]]()
  case class Res[F[_]]()

  type Service[F[_]] = Kleisli[OptionT[F, ?], Req[F], Res[F]]

  def normal[F[_]: Monad](a: Service[F], b: Service[F]) = a <+> b

  case class AuthedReq[F[_], T]()
  type AuthedService[F[_], T] = Kleisli[OptionT[F, ?], AuthedReq[F, T], Res[F]]
  type AuthedServiceSwapped[T, F[_]] = Kleisli[OptionT[F, ?], AuthedReq[F, T], Res[F]]

  // def doesNotCompile[F[_]: Monad, T](a: AuthedService[F, T], b: AuthedService[F, T]) =
  //   a <+> b

  def compiles[F[_]: Monad, T](a: AuthedServiceSwapped[T, F], b: AuthedServiceSwapped[T, F]) =
    a <+> b
}
```

I'll PR a fix for this asap

SystemFw:
Should fix #1506 

I hope I haven't forgotten anything, and that `sbt ci` on my machine is reliable. We shall see :)

### DIFF
--- a/core/src/main/scala/org/http4s/AuthedService.scala
+++ b/core/src/main/scala/org/http4s/AuthedService.scala
@@ -11,15 +11,15 @@ object AuthedService {
     * `apply` instead.
     */
   @deprecated("Use liftF with an OptionT[F, Response[F]] instead", "0.18")
-  def lift[F[_]: Functor, T](f: AuthedRequest[F, T] => F[Response[F]]): AuthedService[F, T] =
+  def lift[F[_]: Functor, T](f: AuthedRequest[F, T] => F[Response[F]]): AuthedService[T, F] =
     Kleisli(f.andThen(OptionT.liftF(_)))
 
   /** Lifts a partial function to an `AuthedService`.  Responds with
     * [[org.http4s.Response.notFoundFor]], which generates a 404, for any request
     * where `pf` is not defined.
     */
-  def apply[F[_], T](pf: PartialFunction[AuthedRequest[F, T], F[Response[F]]])(
-      implicit F: Applicative[F]): AuthedService[F, T] =
+  def apply[T, F[_]](pf: PartialFunction[AuthedRequest[F, T], F[Response[F]]])(
+      implicit F: Applicative[F]): AuthedService[T, F] =
     Kleisli(req => pf.andThen(OptionT.liftF(_)).applyOrElse(req, Function.const(OptionT.none)))
 
   /**
@@ -28,7 +28,7 @@ object AuthedService {
     * @tparam T - ignored.
     * @return
     */
-  def empty[F[_]: Applicative, T]: AuthedService[F, T] =
+  def empty[T, F[_]: Applicative]: AuthedService[T, F] =
     Kleisli.lift(OptionT.none)
 
 }
--- a/core/src/main/scala/org/http4s/package.scala
+++ b/core/src/main/scala/org/http4s/package.scala
@@ -34,7 +34,11 @@ package object http4s { // scalastyle:ignore
     */
   type HttpService[F[_]] = Kleisli[OptionT[F, ?], Request[F], Response[F]]
 
-  type AuthedService[F[_], T] = Kleisli[OptionT[F, ?], AuthedRequest[F, T], Response[F]]
+  /**
+    * We need to change the order of type parameters to make partial unification
+    * trigger. See https://github.com/http4s/http4s/issues/1506
+    */
+  type AuthedService[T, F[_]] = Kleisli[OptionT[F, ?], AuthedRequest[F, T], Response[F]]
 
   type Callback[A] = Either[Throwable, A] => Unit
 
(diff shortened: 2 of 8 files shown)


SystemFw:
Hold on on this, a small git ****up. Will fix soon, apologies

SystemFw:
Yeah, I had missed `AuthenticationStore`, so more work to do

rossabaker:
The changes thus far look right.  It feels a bit unnatural to have the higher-kinded type last, but I'm for making partial unification work as expected.

SystemFw:
Yeah I'm not really fond the aesthetics of it either, but can't think of any other way to make partial unification work in the current setting (type synonyms)

ChristopherDavenport:
Just to confirm this is expected behavior that we care about the order of our types? I understand we need to, but this feels a little like a lesser authority and introducing oddness because of what I personally understand to be a compiler problem. Should we consider creating an issue on this topic in regards to partial-unification with scalac?

rossabaker:
Good question.  Partial unification is always left-to-right, like a curried function.  I like [this explanation](https://gist.github.com/djspiewak/7a81a395c461fd3a09a6941d4cd040f2) for why it can't go both ways.

SystemFw:
Yep, it definitely can't go both ways or you will have ambiguities, and the left to right order is consistent with the view of type constructors as curried type-level functions (which ofc they aren't in Scala, but it's a nice model to think in).
I think at some point there was talk of an annotation to signal that an `F[A, B]` should be unified right to left (it was also for aesthetic reasons, Bill Venners had a use case in ScalaTest iirc), but that didn't materialise, which I think is ultimately a good idea: you don't want something so fundamental to rely on a definition site annotation, it would be really hard to reason about what's going on in general.
The downside is that some constructs will look ugly: this is arguably one such case.

aeons:
This also raises some questions on what types should be `[F[_], A]` and which must be `[A, F[_]]`, and where to draw the line without confusing more than necessary.

If you fix the `.` in this PR, you still have that `AuthedService.apply[F[_], A]` returns an `AuthedService[A, F]`.

There's also `BasicAuth[F, A]`, `DigestAuth[F, A]` and `AuthenticationStore[F, A]` (and probably more), that I don't think *needs* to be `[A, F]`, but it might be more intuitive to have all the `*Auth*` stuff be `[A, F]`.

rossabaker:
Where to draw the line is ambiguous enough to me I'm going to be deferential to those with more hands-on experience with our auth middlewares.

SystemFw:
@aeons I'm in the process of doing the changes for everything (emacs broken, so kinda manually...) just for the sake of consistency. I think the immediate issue is solved by `AuthedService` alone

SystemFw:
Assuming this builds now (I think it does), I think all that's left is changing `AuthedService` `lift` and `empty`. Anything else you can think of?

rossabaker:
At this point, I think just changing `AuthedService` looks right.  I wanted to be consistent, but we could argue that flipping so many of these is so inconsistent with general expectation that we just have to live with the fact that `AuthedService` is the outlier.  I'd flip that alias, leave a scaladoc on why it's flipped, and call it a day.

I remain highly deferential on this, but there's my tiny stake in the ground.

SystemFw:
@rossabaker I have actually come around to pretty much the same position while in the process of doing this. So unless anyone has a strong counterargument, I will revert and:
- change the `AuthedService` alias (+ scaladoc)
- The `apply`  `AuthedService`
- I'm not sure about `AuthedService#empty` and the deprecated  `AuthedService#lift`
- Don't touch the rest of the `Auth` stuff (although I share your position of being deferential on this)

I also hope you're ok with all the rebasing...

rossabaker:
I think `empty` should follow `apply`.  I'd flip that.

I think `lift` is deprecated so it's best _not_ to touch that.  We want to catch people who used to use it, not make it work better.

I am perfectly fine with all the rebasing. :)

SystemFw:
@rossabaker Will @deprecated on `lift` still trigger, since I will _have_ to change the return type (I can choose not to change the type variables, as you suggested)

rossabaker:
I _think_ it will still trigger, but I'd have to try it.  I doubt it was frequently called, so I would not spin wheels over that.  I feel strongest about `apply` and `empty`.

PR #1513, opened 2017-11-02
Unseal the message failure hierarchy

### LINKED ISSUE #1500: Bring back GenericDecodeFailure, GenericParsingFailure, GenericMessageBodyFailure
These were commented out in the move to 0.18.  They fail to compile now because `MessageFailure` is parameteric on `toHttpResponse`, while the generic failures are parameteric on a constructor argument.

One solution would be to make it an abstract class.  Another solution would be to provide an awkward looking parametric function to the constructor.

/cc @coltfred

rossabaker:
Having a sealed `MessageFailure` hierarchy does not to a great deal of good.  `MessageFailure`s usually end up in a failed effect and statically typed as `Throwable`.  It is rare to get any benefits from exhaustiveness checking.

Having an unsealed `MessageFailure` hierarchy lets users customize the `toHttpResponse`.  This was previously handled by the addition of "Generic" case classes in the hierarchy, but were commented out during the parameterization of the effect.  We want the effect parameter to be on `toHttpResponse`, not the `MessageFailure`.  The clearest way to define a `MessageFailure` with a custom response is to implement an abstract class.

The use case of customizing HttpResponses could theoretically be done by pattern matching against a (sealed) `MessageFailure` hierarchy.  But history shows that we can't anticipate the structure that each user wants to put in their failures.

This uses a lot more subtyping than I'm comfortable with, but I think this is a case where the benefits of unsealing exceed the benefits of sealing.  I reluctantly push "Create pull request."

Fixes #1500.

### DIFF
--- a/core/src/main/scala/org/http4s/MessageFailure.scala
+++ b/core/src/main/scala/org/http4s/MessageFailure.scala
@@ -5,24 +5,22 @@ import cats.implicits._
 import scala.util.control.{NoStackTrace, NonFatal}
 
 /** Indicates a failure to handle an HTTP [[Message]]. */
-sealed abstract class MessageFailure extends RuntimeException {
+trait MessageFailure extends RuntimeException {
 
   /** Provides a message appropriate for logging. */
   def message: String
 
   /* Overridden for sensible logging of the failure */
-  final override def getMessage: String =
-    message
+  final override def getMessage: String = message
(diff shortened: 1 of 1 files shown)


coltfred:
@rossabaker I think this makes sense. Thanks for working on this.

PR #1514, opened 2017-11-02
Move Scheme back into Uri

rossabaker:
`Scheme` was the first URI component to get renovated and moved out of `Uri`.  In #1485, we're moving back to keeping the URI components as members of the `Uri` object.  #1485 didn't get done before cats-1.0.0-RC1, but this moves things back to closer to how they'll be, so we can release today.


ChristopherDavenport:
@rossabaker Once again is struck down by the tut gods.

PR #1516, opened 2017-11-02
TransferCoding is not a case class

rossabaker:
Its equality semantics are not consistent with the equality of its components.  Better to just make this a wrapper class, like `Scheme`.


rossabaker:
Pinged @cquiroz since he did the work, but I think several people were around when we discussed case class expectations.  Anyone's review is good here.

PR #1517, opened 2017-11-02
Content Type Fix

ChristopherDavenport:
First PR in converting to reliable `withX`, `withoutX` and `withXOption` functions.

### DIFF
--- a/core/src/main/scala/org/http4s/MessageOps.scala
+++ b/core/src/main/scala/org/http4s/MessageOps.scala
@@ -39,11 +39,14 @@ trait MessageOps[F[_]] extends Any {
   final def withType(t: MediaType)(implicit F: Functor[F]): Self =
     putHeaders(`Content-Type`(t))
 
-  final def withContentType(contentType: Option[`Content-Type`])(implicit F: Functor[F]): Self =
-    contentType match {
-      case Some(t) => putHeaders(t)
-      case None => filterHeaders(_.is(`Content-Type`))
-    }
+  final def withContentType(contentType: `Content-Type`)(implicit F: Functor[F]): Self =
+    putHeaders(contentType)
+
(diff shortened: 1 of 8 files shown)


core/src/main/scala/org/http4s/headers/Content-Type.scala
```diff
@@ -7,7 +7,7 @@ import org.http4s.util.Writer
 object `Content-Type` extends HeaderKey.Internal[`Content-Type`] with HeaderKey.Singleton {
   def apply(mediaType: MediaType, charset: Charset): `Content-Type` =
     apply(mediaType, Some(charset))
-  implicit def apply(mediaType: MediaType): `Content-Type` = apply(mediaType, None)
+  def apply(mediaType: MediaType): `Content-Type` = apply(mediaType, None)
```
ChristopherDavenport: Removed implicit application

core/src/main/scala/org/http4s/MessageOps.scala
```diff
@@ -39,11 +39,14 @@ trait MessageOps[F[_]] extends Any {
...
-      case None => filterHeaders(_.is(`Content-Type`))
-    }
+  final def withContentType(contentType: `Content-Type`)(implicit F: Functor[F]): Self =
+    putHeaders(contentType)
+
+  final def withoutContentType(implicit F: Functor[F]): Self = filterHeaders(_.is(`Content-Type`))
+
+  final def withContentTypeOption(contentTypeO: Option[`Content-Type`])(
+      implicit F: Functor[F]): Self =
+    contentTypeO.fold(withoutContentType)(withContentType)
```
ChristopherDavenport: This syntax should be easily extended to other options.

rossabaker: There are a bunch on the builders, I think.

rossabaker:
I can't approve on phone, but :+1:

PR #1519, opened 2017-11-02
Log4s classpath incompatibility warning

ChristopherDavenport:
Added warning to releases M2-M5 about classpath  incompatibilities between blaze and http4s


aeons:
To clarify, the only change from 1.3.* to 1.4.0 is that you do `getLogger` without arguments in a class that takes higher kinded type parameters.

PR #1462, opened 2017-10-04
Add Changes for Java 9 Testing

ChristopherDavenport:

PR #1520, opened 2017-11-03
Upgrade dependencies and plugins

rossabaker:
Notable:
- blaze upgrade should resolve log4s warning reported by @coltfred and @tpolecat 
- coursier upgrade is Java 9 compatible, which may unblock #1462.


ChristopherDavenport:
Odd failure on Windows async-http-client.

ChristopherDavenport:
Looks good otherwise though, tut plugin got added, or did I miss the old location in this PR?

rossabaker:
tut was inherited through sbt-rig.  This makes it slightly newer.

PR #1523, opened 2017-11-04
Refactor Blaze client creation and deprecate BlazeSimpleHttp1Client

### LINKED ISSUE #1522: Refactor Client Creation
Client by creation is not an effect, but as soon as anything is run then it needs to be shutdown, so it is masking mutable internal state.

As a result Client creation should be within an Effect Type. Such that you can only create a client within an effect. This allows for safer use at the call site. 

`PooledHttpClient[F]() : F[Client]` Which allows `Stream.bracket(client)(Stream.emit(_), _.shutdown)` to be a stream implementation we could also provide.

While we are beating on client. It is probably time for the death of `SimpleHttpClient`

aeons:
This is just the simplest possible solution to #1522.

Closes #1522


blaze-client/src/main/scala/org/http4s/client/blaze/Http1Client.scala
```diff
@@ -20,15 +21,26 @@ object PooledHttp1Client {
...
+      val pool = ConnectionManager.pool(
+        http1,
+        maxTotalConnections,
+        maxWaitQueueLimit,
+        maxConnectionsPerRequestKey,
+        config.executionContext)
+      BlazeClient(pool, config, pool.shutdown())
+    }
+
+  def stream[F[_]: Effect](
```
rossabaker: Can these pool properties move into the config?  It seems weird to pass this big old case class in after already having to set three properties.

blaze-client/src/main/scala/org/http4s/client/blaze/package.scala
```diff
@@ -2,12 +2,14 @@ package org.http4s
...
+import fs2.Stream
 
 package object blaze {
 
-  /** Default blaze client
-    *
-    * This client will create a new connection for every request.
-    */
-  def defaultClient[F[_]: Effect]: Client[F] = SimpleHttp1Client(BlazeClientConfig.defaultConfig)
+  def defaultClient[F[_]: Effect]: F[Client[F]] = Http1Client()
```
rossabaker: What does this even get used for?

blaze-client/src/main/scala/org/http4s/client/blaze/package.scala
```diff
@@ -2,12 +2,14 @@ package org.http4s
...
+
+  def defaultClientStream[F[_]: Effect]: Stream[F, Client[F]] = Http1Client.stream()
+
+  @deprecated("Use Http1Client instead", "0.18.0-M6")
+  type PooledHttp1Client = Http1Client.type
```
rossabaker: Since the signature of `apply` is changing, does it make more sense to leave a deprecated object instead of case class?

blaze-client/src/main/scala/org/http4s/client/blaze/package.scala
aeons: Dunno, but with only one client it doesn't make much sense.

blaze-client/src/main/scala/org/http4s/client/blaze/package.scala
aeons: Yes.

blaze-client/src/main/scala/org/http4s/client/blaze/package.scala
aeons: Remove or leave with deprecation notice?

blaze-client/src/main/scala/org/http4s/client/blaze/Http1Client.scala
aeons: Yes

blaze-client/src/main/scala/org/http4s/client/blaze/package.scala
rossabaker: I will always support deprecating changes that don't add drag for us to keep around.  Makes migration easier for users.

blaze-client/src/main/scala/org/http4s/client/blaze/package.scala
rossabaker: The parameters to `PooledHttp1Client.apply` are different than the parameters to `Http1Client`.  Anyone who calls `PooledHttp1Client` is not going to compile.  If we deprecated the old one, we could point them to the new one but let them upgrade at their own pace.

blaze-client/src/test/scala/org/http4s/client/blaze/BlazeSimpleHttp1ClientSpec.scala
```diff
@@ -3,6 +3,7 @@ package org.http4s.client.blaze
 import org.http4s.client.ClientRouteTestBattery
 import org.http4s.util.threads.newDaemonPoolExecutionContext
 
+@deprecated("Well, we still need to test it", "0.18.0-M6")
```
rossabaker: Is this to get around a deprecation warning invoking deprecated code?

I don't think we necessarily do need to test it, if this test coverage exists against the new interface.

docs/src/main/tut/client.md
```diff
@@ -50,7 +50,7 @@ speaks HTTP 1.x.
 ```tut:book
 import org.http4s.client.blaze._
 
-val httpClient = PooledHttp1Client[IO]()
+val httpClient = Http1Client[IO]().unsafeRunSync
```
rossabaker: I get what's going on here, but this runs against "unsafeRunSync only at the end of the world."  But putting the entire tut into a comprehension also won't work.  Maybe we need to explain ourselves here?

ChristopherDavenport: This is a hard issue as best practice would be using the streaming iterface and using it only where we need it, probably in multiple blocks calling different clients. Although in an app you generally want 1 client bracketed near the beginning.

rossabaker: Yeah, I think in some of the other ones we can flatmap our way to victory.  This one, since it's a longer narrative, is the chewy one.

ChristopherDavenport: At a minimum we need to add comments on why this was done.

ChristopherDavenport:
ping @aeons 

Do you have any time to update the comment for `client.md`? If not I'll push that up for you as I think we're ready to approve otherwise.

aeons:
Completely forgot this again. I’m completely booked this weekend, so you can just push what you want.

blaze-client/src/test/scala/org/http4s/client/blaze/BlazeSimpleHttp1ClientSpec.scala
aeons: The pooled client is tested, as it was before. We are still providing the simple client, although it is deprecated. This is just to keep the tests for that running.

Should the deprecated simple client instead point to a default config pooled client?

aeons:
Right now, there's an `apply` that gives you back an `F[Http1Client[F]]`, and a `stream` method that uses `bracket` to give you back a `Stream[F, HttpClient[F]]`.

The apply just wraps creation in `F.delay`.

If you guys are ok with those two, we could add an `unsafeCreate`, that returns a raw `Http1Client[F]`, and use that in docs with a note that you should probably use the stream variant?

blaze-client/src/test/scala/org/http4s/client/blaze/BlazeSimpleHttp1ClientSpec.scala
rossabaker: Let's leave it and not change behavior, in case people are relying on it.  We're already scolding the user.

Issue #1085, opened 2017-04-08
How do I handle authorization?


Daenyth:
AuthedService wants a `getUser: Request => Task[T]` so it clearly only deals with Authentication. What about authorization? Is there a recommended pattern?

reactormonk:
I'd say Authorization comes later in the API, so it's not really part of http4s anymore. Any other opinions?

rossabaker:
I think it's possibly worth showing how the user type might include information about roles.  I think an authorization framework would be out of scope, but authentication providing the information to driver authorization decisions could be a useful hint.

cquiroz:
I agree, authorization is a large subject and normally varies a lot depending on the situation. It sounds like something your application should handle on its own

The one thing that could help is perhaps storing more metadata as part of the user. E.g. User's group but again it sounds to be out of http4s' scope

dozed:
To give an example, this is how one can do authorization with `Either` and simple functions:

```scala

import scalaz._, Scalaz._
import doobie.imports._

object AuthorizationModule {

  // role check
  def checkAdmin(client: User): ErrorCode \/ Unit = {
    if (client.isAdmin) ().right
    else ErrorCode.UserIsNotAdmin.left
  }

  // authorize operations
  def authorizeBaz(client: User): ErrorCode \/ Unit = ???

  def authorizeFoo(client: User): ErrorCode \/ Unit = ???

  def authorizeBar(client: User): ErrorCode \/ Unit = ???


  // with optional database IO
  def checkCanBaz(isFoo: Boolean, isBar: ConnectionIO[Boolean]): ConnectionIO[ErrorCode \/ Unit] = {
    isBar.ifM(FC.delay(().right), other)
  }

  // with an operation type
  def checkAuthorized(op: Operation, client: User): ErrorCode \/ Unit = {
    op match {
      case Operation.Edit   if !isBar => ().right
      case Operation.Delete if !isFoo => ().right
      case _                          => ErrorCode.Forbidden.left
    }
  }

}
```

kevinmeredith:
> I'd say Authorization comes later in the API, 

Could you please say more, @reactormonk? I don't understand what "later" means.

reactormonk:
Later in your code flow. First you find out who you're talking to, and then later you find out if that person is actually authorized.

jmcardon:
I discussed with @rossabaker on gitter, and I will be tackling this specific issue for http4s.

[...]

PR #1526, opened 2017-11-08
update authentication docs to point to tsec-http4s

jmcardon:
I wasn't sure whether tsec-auth should be at the top, or at the bottom, so I put it at the bottom by default.

If you guys feel tsec-http4s is good, we can move it accordingly so it's easier to see.

### DIFF
--- a/docs/src/main/tut/auth.md
+++ b/docs/src/main/tut/auth.md
@@ -4,6 +4,8 @@ weight: 120
 title: Authentication
 ---
 
+## Built in
+
 A [service] is a `Kleisli[F, Request, Response]`, the composable version of
 `Request[F] => F[Response[F]]`. A service with authentication also requires some kind of `User`
 object which identifies which user did the request. To store the `User` object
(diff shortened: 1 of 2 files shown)


docs/src/main/tut/auth.md
```diff
@@ -132,6 +134,12 @@ val authUser: Kleisli[IO, Request[IO], Either[String,User]] = Kleisli({ request
 })
 ```
 
+### Using tsec-http4s for Authentication and Authorization
+The [TSec] project provides an authentication and authorization module
+ for the http4s project 0.18-M4+. Docs specific to http4s are located [here].
```
ChristopherDavenport: If your going to use a world like `here` I think the link belongs inline.

jmcardon:
Also, could this pr close #1085 ?

jmcardon:
@ChristopherDavenport I fixed the definition of service in that paragraph in the last commit, but opened #1527 for the remaining, so it can be a bit more specific.

docs/src/main/tut/index.md
```diff
@@ -4,6 +4,10 @@ menu: main
 weight: 1
 ---
 
+**Note**: To run examples, please make sure that the flag `-Ypartial-unification`
```
rossabaker: Should we also note that it's enabled by default by the g8 template?

jmcardon: Sure, I'm alright with this. I'm all for anything that makes it more obvious that it's enabled, as it's pretty much the most common source of errors currently.

jmcardon: updated.

PR #1528, opened 2017-11-08
fix auth response handling

jmcardon:
On gitter, @aeons mentioned how `AuthedService` should consume anything it doesn't handle into a `401` unauthorized, to potentially avoid information leaks.

I actually agree with this statement. Currently, for a non-match, it will simply return `notFound`.

This PR addresses this example:

```scala
import cats.Applicative
import org.http4s._
import org.http4s.dsl.io._
import cats.effect.IO
import org.http4s.implicits._
import org.http4s.server.AuthMiddleware

case class User(id: Int)

val authedService: AuthedService[User, IO] = AuthedService[User, IO] {
  case GET -> Root / "asd" as user =>
    Ok()
}

val middleware: AuthMiddleware[IO, User] =
  AuthMiddleware[IO, User](Kleisli(_ => OptionT.pure(User(123))))

//Following this pr, the following would return `Unauthorized`. Currently, this returns 404.
val service = middleware(authedService).orNotFound(Request[IO]()).unsafeRunSync()
```

### DIFF
--- a/core/src/main/scala/org/http4s/AuthedService.scala
+++ b/core/src/main/scala/org/http4s/AuthedService.scala
@@ -20,8 +20,10 @@ object AuthedService {
     */
   def apply[T, F[_]](pf: PartialFunction[AuthedRequest[F, T], F[Response[F]]])(
       implicit F: Applicative[F]): AuthedService[T, F] =
-    Kleisli(req => pf.andThen(OptionT.liftF(_))
-      .applyOrElse(req, Function.const(OptionT.pure(Response[F](Status.Unauthorized)))))
+    Kleisli(
+      req =>
+        pf.andThen(OptionT.liftF(_))
+          .applyOrElse(req, Function.const(OptionT.pure(Response[F](Status.Unauthorized)))))
 
   /**
     * The empty service (all requests fallthrough).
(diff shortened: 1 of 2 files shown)


aeons:
I am not sure whether I was right here (also it was at 6 in the morning with no caffeine :) ).

I will defer to @rossabaker

rossabaker:
I think I'm missing something.  Authorization passes, but there is no matching route.  Why should this return a 401?

The case that trips most people up is authorization not passing and returning a 401 when there is no matching route.  I think this behavior is correct for the aforementioned reasons of leakage, but syntactically, it's not obvious that `auth(a) <+> b` applies `auth` to `b`.  We need better docs, or perhaps more radically, better syntax, for this case.

aeons:
Yeah, it didn't match up in my head either when I thought about it again. I think this is correct as is. An explanation of the behaviour and why it is how it is in the docs would be nice though.

jmcardon:
@aeons @rossabaker After looking at this again, I think how it currently works, as is in regards to matching is alright. I'll close this PR.

> The case that trips most people up is authorization not passing and returning a 401 when there is no matching route. I think this behavior is correct for the aforementioned reasons of leakage, but syntactically, it's not obvious that auth(a) <+> b applies auth to b. We need better docs, or perhaps more radically, better syntax, for this case.

Actually I'm not sure this is the case. Currently, the default case of constructing `AuthMiddleware` will _not_ 401 requests that don't pass authentication, I believe it will 404 them. I'll open a different pr to show what I mean.

PR #1529, opened 2017-11-08
Remove duplication of QValue parsers

cquiroz:
I noted several instances of parsers for `QValue`. This PR removes that duplication and adds a `QValueParser`

It also adds an instance of `HttpCodec[QValue]`

Let me know what you think. It may make the header's parsers a bit less readable

### DIFF
--- a/core/src/main/scala/org/http4s/QValue.scala
+++ b/core/src/main/scala/org/http4s/QValue.scala
@@ -15,7 +15,7 @@ import scala.reflect.macros.whitebox.Context
   * @param thousandths between 0 (for q=0) and 1000 (for q=1)
   * @see [[http://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html#sec3.9 RFC 2616, Section 3.9]]
   */
-final class QValue private[QValue] (val thousandths: Int) extends AnyVal with Ordered[QValue] {
+final class QValue private (val thousandths: Int) extends AnyVal with Ordered[QValue] {
   def toDouble: Double = 0.001 * thousandths
 
   def isAcceptable: Boolean = thousandths > 0
--- a/core/src/main/scala/org/http4s/QValue.scala
+++ b/core/src/main/scala/org/http4s/QValue.scala
@@ -2,7 +2,9 @@ package org.http4s
 
 import cats._
 import macrocompat.bundle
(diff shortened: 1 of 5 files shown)


tests/src/test/scala/org/http4s/QValueSpec.scala
```diff
@@ -2,11 +2,13 @@ package org.http4s
 
 import cats.implicits._
 import cats.kernel.laws.discipline.OrderTests
+import org.http4s.testing.HttpCodecTests
 
 class QValueSpec extends Http4sSpec {
   import QValue._
 
   checkAll("QValue", OrderTests[QValue].order)
+  checkAll("HttpCodec", HttpCodecTests[QValue].httpCodec)
```
rossabaker: I got this wrong myself, but I think `HttpCodec[QValue]` and `Order[QValue]` are the appropriate first arguments to this.  That's the instances that are being tested.  That's how most cats libraries use the laws.

core/src/main/scala/org/http4s/QValue.scala
```diff
@@ -13,10 +15,7 @@ import scala.reflect.macros.whitebox.Context
   * @param thousandths between 0 (for q=0) and 1000 (for q=1)
   * @see [[http://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html#sec3.9 RFC 2616, Section 3.9]]
   */
-final class QValue private[QValue] (val thousandths: Int)
-    extends AnyVal
-    with Ordered[QValue]
-    with Renderable {
+final class QValue private[QValue] (val thousandths: Int) extends AnyVal with Ordered[QValue] {
```
rossabaker: Does `private[QValue]` mean anything different than `private`?  This is not new, but it looks weird.

cquiroz: I guess plain `private` will do just fine.

cquiroz: Note that i removed `with Renderable` as it doesn't seem to be used as `Renderable` but you may have a different opinion

tests/src/test/scala/org/http4s/QValueSpec.scala
cquiroz: will change that

PR #1530, opened 2017-11-08
authmiddleware to consume the entire request

jmcardon:
This pr aims to address this issue
```scala
import cats._
import cats.data._
import org.http4s._
import org.http4s.dsl.io._
import cats.effect.IO
import cats.implicits._
import org.http4s.implicits._
import org.http4s.server.AuthMiddleware

case class User(id: Int)

val authedService: AuthedService[User, IO] = AuthedService[User, IO] {
  case GET -> Root / "asd" as user =>
    Ok()
}

val middleware: AuthMiddleware[IO, User] =
  AuthMiddleware[IO, User](Kleisli(_ => OptionT.none))

val service1 = middleware(authedService)

val service2: HttpService[IO] = HttpService[IO] {
  case GET -> Root =>
    Ok()
}


(middleware(authedService) <+> service2).orNotFound(Request[IO]()).unsafeRunSync()
```
For security reasons, going through an `AuthMiddleware` should consume an entire request: i.e, it should never fall through to the other service. This _should_ return 404, but it currently returns ok.

This pr adjusts that behavior in the following manner:
* If the user is authenticated, but the route is unmatched, the request will fall through to not found
* if the user is not authenticated, it will be  `Unauthorized`
* service composition with `<+>` for authmiddleware'd service will never reach any service composed 
to the right of the authed service

### DIFF
--- a/server/src/main/scala/org/http4s/server/package.scala
+++ b/server/src/main/scala/org/http4s/server/package.scala
@@ -52,7 +52,9 @@ package object server {
     def apply[F[_]: Monad, T](
         authUser: Kleisli[OptionT[F, ?], Request[F], T]): AuthMiddleware[F, T] =
       service => {
-        service.compose(Kleisli((req: Request[F]) => authUser(req).map(AuthedRequest(_, req))))
+        Kleisli((r: Request[F]) => authUser(r).map(AuthedRequest(_, r)))
+          .andThen(service.mapF(o => OptionT.liftF(o.fold(Response[F](Status.NotFound))(identity))))
+          .mapF(o => OptionT.liftF(o.fold(Response[F](Status.Unauthorized))(identity)))
       }
 
     def apply[F[_], Err, T](
--- a/server/src/test/scala/org/http4s/server/middleware/authentication/AuthMiddlewareSpec.scala
+++ b/server/src/test/scala/org/http4s/server/middleware/authentication/AuthMiddlewareSpec.scala
@@ -84,7 +84,6 @@ class AuthMiddlewareSpec extends Http4sSpec {
       val authUser: Kleisli[OptionT[IO, ?], Request[IO], User] =
         Kleisli.pure(userId)
 
-
       val authedService: AuthedService[User, IO] =
         AuthedService {
           case POST -> Root as _ => Ok()
@@ -101,7 +100,6 @@ class AuthMiddlewareSpec extends Http4sSpec {
     "return 401 for a matched, but unauthenticated route" in {
       val authUser: Kleisli[OptionT[IO, ?], Request[IO], User] =
         Kleisli.lift(OptionT.none)
-
 
       val authedService: AuthedService[User, IO] =
         AuthedService {
@@ -122,7 +120,6 @@ class AuthMiddlewareSpec extends Http4sSpec {
       val authUser: Kleisli[OptionT[IO, ?], Request[IO], User] =
         Kleisli.pure(userId)
 
-
       val authedService1: AuthedService[User, IO] =
         AuthedService {
           case POST -> Root as _ => Ok()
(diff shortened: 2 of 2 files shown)


server/src/main/scala/org/http4s/server/package.scala
```diff
@@ -52,7 +52,9 @@ package object server {
     def apply[F[_]: Monad, T](
         authUser: Kleisli[OptionT[F, ?], Request[F], T]): AuthMiddleware[F, T] =
       service => {
-        service.compose(Kleisli((req: Request[F]) => authUser(req).map(AuthedRequest(_, req))))
+        Kleisli((r: Request[F]) => authUser(r).map(AuthedRequest(_, r)))
+          .andThen(service.mapF(o => OptionT.liftF(o.fold(Response[F](Status.NotFound))(identity))))
+          .mapF(o => OptionT.liftF(o.fold(Response[F](Status.Unauthorized))(identity)))
```
rossabaker: Should these canned responses come with a default message?  I've never had a good feel for that.  It feels like yes, but then people fight about the content type, and it's a bit anglocentric.

server/src/test/scala/org/http4s/server/middleware/authentication/AuthMiddlewareSpec.scala
```diff
@@ -76,6 +77,90 @@ class AuthMiddlewareSpec extends Http4sSpec {
...
+      val authedService: AuthedService[User, IO] =
+        AuthedService {
+          case POST -> Root as _ => Ok()
+        }
+
+      val middleware = AuthMiddleware(authUser)
+
+      val service = middleware(authedService)
+
+      service.orNotFound(Request[IO](method = Method.POST)) must returnStatus(Ok)
```
rossabaker: Is this to test your test?

server/src/test/scala/org/http4s/server/middleware/authentication/AuthMiddlewareSpec.scala
```diff
@@ -76,6 +77,90 @@ class AuthMiddlewareSpec extends Http4sSpec {
...
+        AuthedService {
+          case POST -> Root as _ => Ok()
+        }
+
+      val middleware = AuthMiddleware(authUser)
+
+      val service = middleware(authedService)
+
+      service.orNotFound(Request[IO](method = Method.POST)) must returnStatus(Unauthorized)
+      service.orNotFound(Request[IO](method = Method.GET)) must returnStatus(Unauthorized)
```
rossabaker: This is an unmatched authenticated route.  The condition is right, but maybe that's a separate test.

server/src/main/scala/org/http4s/server/package.scala
jmcardon: I'd rather they not. If people wanted this, they could use the method that deals with the explicit error in `authmiddleware`.
I'm partial to simply the 401, but I'll let others take the wheel.

server/src/test/scala/org/http4s/server/middleware/authentication/AuthMiddlewareSpec.scala
jmcardon: None of the prior tests actually test the default apply in `authmiddleware`. They test the second apply with explicit error.

I figured it doesn't hurt to test whether it passes through anyway.

server/src/test/scala/org/http4s/server/middleware/authentication/AuthMiddlewareSpec.scala
jmcardon: I can move to a separate test for this one, sure

server/src/test/scala/org/http4s/server/middleware/authentication/AuthMiddlewareSpec.scala
jmcardon: I moved this to a separate test, actually.

PR #1533, opened 2017-11-09
CSRF middleware backport - 0.16.x

jmcardon:
Here's a tentative implementation of the backport from `TSec`.

I had to port apache commons `Hex.encodeHex` to avoid bringing in another dependency to be able to get a 
performant Hex encoder, and handwrote `find` on NonEmptyList only because I'm not too familiar with `scalaz`. I figured there maybe a find function already I just can't find it.

It's syntactically Equivalent to the version I currently am PRing fixes onto `tsec` for, and adapted similarly from the `play framework` implementation. The gist of the vulnerability mitigation is to use the middleware to set the token cookie, and send such a token along with a custom header value. Given that an attacker forging a request cannot access the values of a cookie due to same-origin policy, this will guard against CSRF.

To defend against `BREACH`, we effectively randomize the token on each request,  but to avoid recomputing both the signature _and_ the tokenId, we can reuse the raw value from the incoming request.

After some review and seeing what you guys think (if anyone can improve some of the code in there, awesome, I don't use scalaz much), I can port this onto 0.17.x as well.

### DIFF
--- a/server/src/main/scala/org/http4s/server/middleware/CSRFMiddleware.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/CSRFMiddleware.scala
@@ -3,56 +3,57 @@ package org.http4s.server.middleware
 import java.nio.charset.StandardCharsets
 import java.security.{MessageDigest, SecureRandom}
 import java.time.Clock
-import javax.crypto.{KeyGenerator, Mac, SecretKey}
-import javax.crypto.spec.SecretKeySpec
 import java.util.Base64
+import javax.crypto.spec.SecretKeySpec
+import javax.crypto.{KeyGenerator, Mac, SecretKey}
 
-import org.http4s.{MaybeResponse, Request, Response, Status, Cookie}
-import org.http4s.server.Middleware
 import org.http4s.headers.{Cookie => HCookie}
-import org.http4s.util.{CaseInsensitiveString, NonEmptyList}
+import org.http4s.server.Middleware
+import org.http4s.util.{CaseInsensitiveString, encodeHex}
+import org.http4s._
 
-import scalaz.{Kleisli, OptionT}
 import scalaz.concurrent.Task
 import scalaz.syntax.std.boolean._
+import scalaz.{Kleisli, OptionT}
 
-case class CSRFMiddleware(config: CSRFConfiguration,
-  clock: Clock = Clock.systemUTC()) {
+final class CSRFMiddleware private[middleware] (
+    val headerName: String = "X-Csrf-Token",
+    val cookieName: String = "csrf-token",
+    key: SecretKey,
+    clock: Clock = Clock.systemUTC()) {
 
   /** Sign our token using the current time in milliseconds as a nonce
     * Signing and generating a token is potentially a unsafe operation
     * if constructed with a bad key.
     */
-  protected[middleware] def signToken(token: String): Task[String] = {
+  private[middleware] def signToken(token: String): Task[String] = {
     val joined = token + "-" + clock.millis()
-    for {
-      instance <- Task.delay(Mac.getInstance(CSRFMiddleware.SigningAlgo))
-      _ <- Task.delay(instance.init(config.key))
-      out <- Task.delay(instance.doFinal(joined.getBytes("UTF-8")))
-    } yield joined + "-" + Base64.getEncoder.encodeToString(out)
+    Task.delay {
+      val mac = Mac.getInstance(CSRFMiddleware.SigningAlgo)
+      mac.init(key)
+      val out = mac.doFinal(joined.getBytes(StandardCharsets.UTF_8))
+      joined + "-" + Base64.getEncoder.encodeToString(out)
+    }
   }
 
   /** Generate a new token **/
-  protected[middleware] def generateToken: Task[String] =
+  private[middleware] def generateToken: Task[String] =
     signToken(CSRFMiddleware.genTokenString)
 
   /** Decode our CSRF token and extract the original token string to sign
     */
-  protected[middleware] def extractRaw(token: String): OptionT[Task, String] =
+  private[middleware] def extractRaw(token: String): OptionT[Task, String] =
     token.split("-") match {
       case Array(raw, nonce, signed) =>
-        val res = for {
-          instance <- Task.delay(Mac.getInstance(CSRFMiddleware.SigningAlgo))
-          _ <- Task.delay(instance.init(config.key))
-          out <- Task.delay(
-            instance.doFinal(
-              (raw + "-" + nonce).getBytes(StandardCharsets.UTF_8)))
-          r <- Task.point(
-            MessageDigest
-              .isEqual(out, Base64.getDecoder.decode(signed))
-              .option(raw))
-        } yield r
-        OptionT(res)
+        OptionT[Task, String](Task.delay {
+          val mac = Mac.getInstance(CSRFMiddleware.SigningAlgo)
+          mac.init(key)
+          val out =
+            mac.doFinal((raw + "-" + nonce).getBytes(StandardCharsets.UTF_8))
+          MessageDigest
+            .isEqual(out, Base64.getDecoder.decode(signed))
+            .option(raw)
+        })
       case _ =>
         OptionT.none
     }
@@ -70,9 +71,9 @@ case class CSRFMiddleware(config: CSRFConfiguration,
     service =>
       Kleisli[Task, Request, MaybeResponse] { r =>
         (for {
-          c1 <- CSRFMiddleware.cookieFromHeaders(r, config.cookieName)
+          c1 <- CSRFMiddleware.cookieFromHeaders(r, cookieName)
           c2 <- OptionT(
-            Task.point(r.headers.get(CaseInsensitiveString(config.headerName))))
+            Task.point(r.headers.get(CaseInsensitiveString(headerName))))
           raw1 <- extractRaw(c1.content)
           raw2 <- extractRaw(c2.value)
           response <- if (CSRFMiddleware.isEqual(raw1, raw2)) {
@@ -82,9 +83,8 @@ case class CSRFMiddleware(config: CSRFConfiguration,
           }
           newToken <- OptionT[Task, String](signToken(raw1).map(Some(_)))
         } yield
-          response.cata(
-            _.addCookie(Cookie(config.cookieName, newToken)),
-            Response(Status.NotFound)))
+          response.cata(_.addCookie(Cookie(cookieName, newToken)),
+                        Response(Status.NotFound)))
           .getOrElse(Response(Status.Unauthorized))
       }
   }
@@ -93,8 +93,8 @@ case class CSRFMiddleware(config: CSRFConfiguration,
   def embedNew(res: MaybeResponse): Task[Response] =
     generateToken.map(
       content =>
-        res.cata(_.addCookie(Cookie(config.cookieName, content)),
-          Response(Status.NotFound)))
+        res.cata(_.addCookie(Cookie(cookieName, content)),
+                 Response(Status.NotFound)))
 
   /** Middleware to embed a csrf token into routes that do not have one. **/
   def withNewToken: Middleware[Request, MaybeResponse, Request, MaybeResponse] =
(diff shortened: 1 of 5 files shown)


server/src/main/scala/org/http4s/server/middleware/CSRFMiddleware.scala
```diff
@@ -0,0 +1,196 @@
...
+import java.util.Base64
+
+import org.http4s.{MaybeResponse, Request, Response, Status, Cookie}
+import org.http4s.server.Middleware
+import org.http4s.headers.{Cookie => HCookie}
+import org.http4s.util.{CaseInsensitiveString, NonEmptyList}
+
+import scalaz.{Kleisli, OptionT}
+import scalaz.concurrent.Task
+import scalaz.syntax.std.boolean._
```
rossabaker: Starting in 0.18, we've just done an alphabetical sort of the imports.  This isn't enforced on the legacy branches, but it'll fit in nicer if we start down here.

server/src/main/scala/org/http4s/server/middleware/CSRFMiddleware.scala
```diff
@@ -0,0 +1,196 @@
...
+    * Signing and generating a token is potentially a unsafe operation
+    * if constructed with a bad key.
+    */
+  protected[middleware] def signToken(token: String): Task[String] = {
+    val joined = token + "-" + clock.millis()
+    for {
+      instance <- Task.delay(Mac.getInstance(CSRFMiddleware.SigningAlgo))
+      _ <- Task.delay(instance.init(config.key))
+      out <- Task.delay(instance.doFinal(joined.getBytes("UTF-8")))
+    } yield joined + "-" + Base64.getEncoder.encodeToString(out)
```
rossabaker: What does doing this monadically achieve vs. putting it in a single `Task.delay`?

server/src/main/scala/org/http4s/server/middleware/CSRFMiddleware.scala
```diff
@@ -0,0 +1,196 @@
...
+import org.http4s.{MaybeResponse, Request, Response, Status, Cookie}
+import org.http4s.server.Middleware
+import org.http4s.headers.{Cookie => HCookie}
+import org.http4s.util.{CaseInsensitiveString, NonEmptyList}
+
+import scalaz.{Kleisli, OptionT}
+import scalaz.concurrent.Task
+import scalaz.syntax.std.boolean._
+
+case class CSRFMiddleware(config: CSRFConfiguration,
```
rossabaker: I wouldn't think of this as a case class.  Not going to use equality, extraction, etc.

server/src/main/scala/org/http4s/server/middleware/CSRFMiddleware.scala
```diff
@@ -0,0 +1,196 @@
...
+  clock: Clock = Clock.systemUTC()) {
+
+  /** Sign our token using the current time in milliseconds as a nonce
+    * Signing and generating a token is potentially a unsafe operation
+    * if constructed with a bad key.
+    */
+  protected[middleware] def signToken(token: String): Task[String] = {
```
rossabaker: Is this class open to extension?  Why are these protected?  Testing?

server/src/main/scala/org/http4s/server/middleware/CSRFMiddleware.scala
```diff
@@ -0,0 +1,196 @@
...
+    val joined = token + "-" + clock.millis()
+    for {
+      instance <- Task.delay(Mac.getInstance(CSRFMiddleware.SigningAlgo))
+      _ <- Task.delay(instance.init(config.key))
+      out <- Task.delay(instance.doFinal(joined.getBytes("UTF-8")))
```
rossabaker: I'd probably use the `StandardCharsets` constant here.

server/src/main/scala/org/http4s/server/middleware/CSRFMiddleware.scala
```diff
@@ -0,0 +1,196 @@
...
+          instance <- Task.delay(Mac.getInstance(CSRFMiddleware.SigningAlgo))
+          _ <- Task.delay(instance.init(config.key))
+          out <- Task.delay(
+            instance.doFinal(
+              (raw + "-" + nonce).getBytes(StandardCharsets.UTF_8)))
+          r <- Task.point(
+            MessageDigest
+              .isEqual(out, Base64.getDecoder.decode(signed))
+              .option(raw))
+        } yield r
```
rossabaker: Unnecessarily monadic again, I think.

server/src/main/scala/org/http4s/server/middleware/CSRFMiddleware.scala
```diff
@@ -0,0 +1,196 @@
...
+          raw1 <- extractRaw(c1.content)
+          raw2 <- extractRaw(c2.value)
+          response <- if (CSRFMiddleware.isEqual(raw1, raw2)) {
+            OptionT[Task, MaybeResponse](service(r).map(Some(_)))
+          } else {
+            OptionT.none[Task, MaybeResponse]
+          }
+          newToken <- OptionT[Task, String](signToken(raw1).map(Some(_)))
+        } yield
+          response.cata(
```
rossabaker: I'm going to think a bit here on what happens with a `Pass`.  Have you thought about the desired semantics, or is this slinging code until it compiles to get some feedback?

server/src/main/scala/org/http4s/server/middleware/CSRFMiddleware.scala
```diff
@@ -0,0 +1,196 @@
...
+
+}
+
+object CSRFMiddleware {
+
+  /** An instance of SecureRandom to generate
+    * tokens, properly seeded:
+    * https://tersesystems.com/blog/2015/12/17/the-right-way-to-use-securerandom/
+    */
+  private lazy val CachedRandom = {
```
rossabaker: Does this have to be lazy?

server/src/main/scala/org/http4s/server/middleware/CSRFMiddleware.scala
```diff
@@ -0,0 +1,196 @@
...
+
+  val SigningAlgo = "HmacSHA1"
+  val SHA1ByteLen = 20
+  val CSRFTokenLength = 32
+
+  /** Hex encoding digits. Adapted from apache commons Hex.encodeHex **/
+  private val Digits: Array[Char] = Array('0', '1', '2', '3', '4', '5', '6',
+    '7', '8', '9', 'A', 'B', 'C', 'D', 'E', 'F')
+
+  protected def find[A](nel: NonEmptyList[A], p: A => Boolean): Option[A] =
```
rossabaker: We use our own `NonEmptyList` in 0.16, because Scalaz broke theirs across versions.  I think we could put this straight on there.

We have syntax for NonEmptyList in later versions to add the operations that are missing from cats'.  When it comes time to merge up there, I think this belongs there.  I also think we should offer these missing methods to cats.

server/src/main/scala/org/http4s/server/middleware/CSRFMiddleware.scala
```diff
@@ -0,0 +1,196 @@
...
+
+  protected def cookieFromHeaders(request: Request,
+    headerName: String): OptionT[Task, Cookie] =
+    OptionT(
+      Task.point(HCookie
+        .from(request.headers)
+        .flatMap(v => CSRFMiddleware.find[Cookie](v.values, _.name == headerName))))
+
+  /** A Constant-time string equality **/
+  def isEqual(s1: String, s2: String): Boolean =
```
rossabaker: I think I've seen this somewhere on our project, but it might be in @reactormonk's cryptobits.  This is useful for more than CSRF, so it might go into the util package.

server/src/main/scala/org/http4s/server/middleware/CSRFMiddleware.scala
```diff
@@ -0,0 +1,196 @@
...
+    MessageDigest.isEqual(s1.getBytes(StandardCharsets.UTF_8),
+      s2.getBytes(StandardCharsets.UTF_8))
+
+  /** Encode a string to a Hexadecimal string representation
+    * Adapted from apache commons Hex.encodeHex
+    */
+  protected def encodeHex(data: Array[Byte]): Array[Char] = {
```
rossabaker: Good candidate for util package.

server/src/main/scala/org/http4s/server/middleware/CSRFMiddleware.scala
```diff
@@ -0,0 +1,196 @@
...
+  /** Generate an unsigned CSRF token from a `SecureRandom` **/
+  protected def genTokenString: String = {
+    val bytes = new Array[Byte](CSRFTokenLength)
+    CachedRandom.nextBytes(bytes)
+    new String(encodeHex(bytes))
+  }
+
+  /** Generate a signing Key for the CSRF token **/
+  def generateSigningKey(): SecretKey =
+    KeyGenerator.getInstance(SigningAlgo).generateKey()
```
rossabaker: Task?  This is effectful.

server/src/main/scala/org/http4s/server/middleware/CSRFMiddleware.scala
```diff
@@ -0,0 +1,196 @@
...
+      j += 1
+      out(j) = Digits(0x0F & data(i))
+      j += 1
+      i += 1
+    }
+    out
+  }
+
+  /** Generate an unsigned CSRF token from a `SecureRandom` **/
+  protected def genTokenString: String = {
```
rossabaker: Why protected in an object?

This looks effectful, but if it's not public API and unwrapped for efficiency, that's okay.

server/src/main/scala/org/http4s/server/middleware/CSRFMiddleware.scala
```diff
@@ -0,0 +1,196 @@
...
+    new String(encodeHex(bytes))
+  }
+
+  /** Generate a signing Key for the CSRF token **/
+  def generateSigningKey(): SecretKey =
+    KeyGenerator.getInstance(SigningAlgo).generateKey()
+
+  /** Build a new HMACSHA1 Key for our CSRF Middleware
+    * from key bytes. This operation is unsafe, in that
+    * any amount less than 20 bytes will throw an exception when loaded
```
rossabaker: A failed task, not an exception, right?

server/src/main/scala/org/http4s/server/middleware/CSRFMiddleware.scala
```diff
@@ -0,0 +1,196 @@
...
+    * into `Mac`, and any value above will be truncated (not good for entropy).
+    *
+    * Use for loading a key from a config file, after having generated
+    * one safely
+    *
+    */
+  def buildSigningKey(array: Array[Byte]): Task[SecretKey] =
+    Task.delay(new SecretKeySpec(array.slice(0, SHA1ByteLen), SigningAlgo))
+
+  def unsafeBuildSigningKey(array: Array[Byte]): SecretKey =
```
rossabaker: Why this vs. `buildSigningKey`?

server/src/main/scala/org/http4s/server/middleware/CSRFMiddleware.scala
```diff
@@ -0,0 +1,196 @@
...
+    new SecretKeySpec(array, SigningAlgo)
+
+}
+
+/** A basic CSRF configuration class **/
+case class CSRFConfiguration(
```
rossabaker: I've become wary of case classes for configuration, because they don't evolve in a binary compatible way.  It sucks, but I might expand this out into something that looks like what [contraband generates](http://www.scala-sbt.org/contraband/code-generation.html)

server/src/main/scala/org/http4s/server/middleware/CSRFMiddleware.scala
jmcardon: You are right.

server/src/main/scala/org/http4s/server/middleware/CSRFMiddleware.scala
jmcardon: up to you, I can move it if you'd like.

server/src/main/scala/org/http4s/server/middleware/CSRFMiddleware.scala
jmcardon: this `find` is already in cats actually. I can add it to the http4s nel.

server/src/main/scala/org/http4s/server/middleware/CSRFMiddleware.scala
jmcardon: SecureRandom will can block for the first use,while it gathers entropy, but I only want _one_ instance, since creation is expensive.

I don't feel strongly either way since objects themselves are lazy.

server/src/main/scala/org/http4s/server/middleware/CSRFMiddleware.scala
jmcardon: The `cata` is intentional. I _want it_ to 404 normally if it passes the csrf challenge, but return `Unauthorized` otherwise.

server/src/main/scala/org/http4s/server/middleware/CSRFMiddleware.scala
jmcardon: oh my bad, didn't see this one. Yeah, i'm simply not surrounding it in `Task` for efficiency. I can make private.

server/src/main/scala/org/http4s/server/middleware/CSRFMiddleware.scala
rossabaker: Yeah, we might as well.  This could come in handy elsewhere.

server/src/main/scala/org/http4s/server/middleware/CSRFMiddleware.scala
rossabaker: I think it's great to explain why it can fail, but I wouldn't call it throwing.  It made it sound scarier than it is.

rossabaker:
Regarding the compilation errors, the ways to construct a Scalaz Task are:
* `Task.now` (strict) 
* `Task.delay` (lazy) 
* `Task.apply` (shifts to provided ExecutorService; rare)

jmcardon:
Ah, my bad. `point` wasn't throwing errors on local. 

I will fix

rossabaker:
I think it may have been added in a later Scalaz.

If you run into any differences on run vs. unsafePerformSync, there is an `org.http4s.compatibility._` trait for that.  That should not be used past 0.16.

jmcardon:
@rossabaker I'm a bit stumped on this error... 

Any ideas?

jmcardon:
5th time's the charm :tada:

rossabaker:
Actually, one last suggestion: Can we name this `Csrf` instead of `CSRFMiddleware`?  We don't call anything else `Middleware` in the type name, just the package.

rossabaker:
I went ahead and pushed my own suggestion.  If people don't like it, we can revert and merge as is.  This will affect #1538.

PR #1534, opened 2017-11-09
Experimental FS2 Implicit Design Sketch

ChristopherDavenport:
Uses a Custom released version of 

fs2
fs2-reactive-streams
jawn-fs2


rossabaker:
What's going on here?  Is this a laboratory to propose a change to fs2?

ChristopherDavenport:
Correct, plus a start of the PR when whatever PR goes live, so a 2 for 1

PR #1535, opened 2017-11-09
ContentCoding without Registry

cquiroz:
This started as a removal of `Registry` from `ContentCoding` but grew into a bigger refactoring using the current coding standards:

* Removal of `Registry` usage
* Removed usage of `CaseInsensitiveString`
* Implementation of `HttpCodec[ContentCoding]`
* Tightening constructions. Empty strings codings are not allowed
* More testing

### DIFF
--- a/core/src/main/scala/org/http4s/ContentCoding.scala
+++ b/core/src/main/scala/org/http4s/ContentCoding.scala
@@ -22,17 +22,18 @@ import cats.{Order, Show}
 import cats.syntax.eq._
 import cats.instances.tuple._
 import cats.instances.string._
+import org.http4s.QValue.QValueParser
 import org.http4s.internal.parboiled2.{Parser => PbParser, _}
 import org.http4s.parser.Http4sParser
 import org.http4s.util._
-import org.http4s.QValue.QValueParser
+
 import scala.util.hashing.MurmurHash3
 
-class ContentCoding(val coding: String, override val qValue: QValue = QValue.One)
+class ContentCoding private (val coding: String, override val qValue: QValue = QValue.One)
     extends HasQValue
     with Ordered[ContentCoding]
     with Renderable {
-  def withQValue(q: QValue): ContentCoding = ContentCoding(coding, q)
+  def withQValue(q: QValue): ContentCoding = new ContentCoding(coding, q)
 
   @deprecated("Use `Accept-Encoding`.isSatisfiedBy(encoding)", "0.16.1")
   def satisfies(encoding: ContentCoding): Boolean = encoding.satisfiedBy(this)
@@ -46,7 +47,8 @@ class ContentCoding(val coding: String, override val qValue: QValue = QValue.One
     (this === ContentCoding.`*` || this.coding.equalsIgnoreCase(encoding.coding))
 
   override def equals(o: Any) = o match {
-    case that: ContentCoding => this.coding.equalsIgnoreCase(that.coding) && this.qValue === that.qValue
+    case that: ContentCoding =>
+      this.coding.equalsIgnoreCase(that.coding) && this.qValue === that.qValue
     case _ => false
   }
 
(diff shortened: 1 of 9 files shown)


core/src/main/scala/org/http4s/ContentCoding.scala
```diff
@@ -18,54 +18,117 @@
...
+import cats.syntax.eq._
+import cats.instances.tuple._
+import cats.instances.string._
+import org.http4s.QValue.QValueParser
+import org.http4s.internal.parboiled2.{Parser => PbParser, _}
+import org.http4s.parser.Http4sParser
 import org.http4s.util._
 
-final case class ContentCoding(coding: CaseInsensitiveString, qValue: QValue = QValue.One)
+import scala.util.hashing.MurmurHash3
```
rossabaker: Totally pedantic, but our imports are all generally in one block now.  The ordering is correct.

core/src/main/scala/org/http4s/ContentCoding.scala
```diff
@@ -18,54 +18,117 @@
...
+
+      override def parse(s: String): ParseResult[ContentCoding] =
+        ContentCoding.parse(s)
+
+      override def render(writer: Writer, coding: ContentCoding): writer.type =
+        writer << coding.coding.toLowerCase << coding.qValue
+
+    }
+
+  implicit val order: Order[ContentCoding] = Order.by(c => (c.coding.toLowerCase, c.qValue))
```
rossabaker: Any reason order is split out instead of part of the instances?

core/src/main/scala/org/http4s/ContentCoding.scala
```diff
@@ -18,54 +18,117 @@
...
+  val deflate = new ContentCoding("deflate")
+  val exi = new ContentCoding("exi")
+  val gzip = new ContentCoding("gzip")
+  val identity = new ContentCoding("identity")
+  val `pack200-gzip` = new ContentCoding("pack200-gzip")
 
   // Legacy encodings defined by RFC2616 3.5.
-  val `x-compress` = register("x-compress".ci, compress)
-  val `x-gzip` = register("x-gzip".ci, gzip)
+  val `x-compress` = compress
```
rossabaker: I'm surprised these are equivalent to their replacements, but I guess they always have been.

core/src/main/scala/org/http4s/ContentCoding.scala
```diff
@@ -18,54 +18,117 @@
...
+  val `x-gzip` = gzip
+
+  def standard: Map[String, ContentCoding] =
```
rossabaker: This should be a val.  It will get recreated on every parse otherwise.

core/src/main/scala/org/http4s/ContentCoding.scala
```diff
@@ -18,54 +18,117 @@
...
+    List(`*`, compress, deflate, exi, gzip, identity, `pack200-gzip`).map(c => c.coding -> c).toMap
+
+  /**
+    * Parse a Transfer Coding
```
rossabaker: ContentCoding

core/src/main/scala/org/http4s/parser/SimpleHeaders.scala
```diff
@@ -68,7 +68,7 @@ private[parser] trait SimpleHeaders {
     new Http4sHeaderParser[`Content-Encoding`](value) {
       def entry = rule {
         Token ~ EOL ~> { s: String =>
-          `Content-Encoding`(ContentCoding.getOrElseCreate(s.ci))
+          `Content-Encoding`(ContentCoding.standard.getOrElse(s, ContentCoding.unsafeFromString(s)))
```
rossabaker: If this extends the parser in `ContentCoding`, you can reuse the rule.

core/src/main/scala/org/http4s/ContentCoding.scala
cquiroz: That's correct

core/src/main/scala/org/http4s/ContentCoding.scala
cquiroz: I don't know how could I make it part of `http4sInstances` and use `Order.by` at the same time

rossabaker: Oh, I see.  I've probably just been defining it myself elsewhere.  If it stands alone, it should follow the naming pattern.  We want to get all our implicits with a unique name to avoid shadowing.

cquiroz: What's the naming convention? I'm not sure I've seen it

rossabaker: `http4s{nameOfTypeclass}For{nameOfTypeParameter}` is consistent with what Cats does. Or `http4sInstancesFor{nameOfTypeParameter}` for one big compound instance.

PR #1536, opened 2017-11-10
Render a banner and version info on server startup

rossabaker:


server/src/main/scala/org/http4s/server/Server.scala
```diff
@@ -16,6 +19,25 @@ abstract class Server[F[_]](implicit F: Effect[F]) {
 
   def address: InetSocketAddress
 
+  def isSecure: Boolean
+
+  def baseUri: Uri = Uri(
```
rossabaker: This is why our URI type needs reform. ☣️

ChristopherDavenport: Ugh.

server/src/main/scala/org/http4s/server/ServerBuilder.scala
```diff
@@ -62,6 +69,12 @@ object ServerBuilder {
   val DefaultHost = LoopbackAddress
   val DefaultHttpPort = 8080
   val DefaultSocketAddress = InetSocketAddress.createUnresolved(DefaultHost, DefaultHttpPort)
+  val DefaultBanner =
```
ChristopherDavenport: Love it

rossabaker:
This will have some subtle conflicts in the merge train.

PR #1538, opened 2017-11-10
CSRF Middleware backport 0.17.x

jmcardon:
Same as #1533, but for 0.17.x

The tests kinda look ugly because old fs2 task returns `Option`.

Practically copy paste from the other branch, except with different imports.

### DIFF
--- a/core/src/main/scala/org/http4s/util/package.scala
+++ b/core/src/main/scala/org/http4s/util/package.scala
@@ -12,34 +12,61 @@ package object util {
   def decode[F[_]](charset: Charset): Pipe[F, Byte, String] = {
     val decoder = charset.nioCharset.newDecoder
     val maxCharsPerByte = math.ceil(decoder.maxCharsPerByte().toDouble).toInt
-    val avgBytesPerChar = math.ceil(1.0 / decoder.averageCharsPerByte().toDouble).toInt
+    val avgBytesPerChar =
+      math.ceil(1.0 / decoder.averageCharsPerByte().toDouble).toInt
     val charBufferSize = 128
 
     _.repeatPull[String] {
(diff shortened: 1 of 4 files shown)


server/src/test/scala/org/http4s/server/middleware/CSRFSpec.scala
```diff
@@ -0,0 +1,120 @@
...
+import org.http4s.headers.{`Set-Cookie` => HCookie}
+import fs2.Task
+import org.http4s.implicits._
+import cats.implicits._
+
+class CSRFSpec extends Http4sSpec {
+
+  val dummyService: HttpService = HttpService {
+    case GET -> Root =>
+      Thread.sleep(1) //Fix so clock doesn't compute the same nonce, to emulate a real service
```
rossabaker: I didn't catch this the first time but we can use fs2 primtives to sleep without blocking a thread.  Not that blocking a thread for a millisecond particularly matters, but because it's a bad example for others who throw larger constants into this.

core/src/main/scala/org/http4s/util/package.scala
```diff
@@ -12,34 +12,61 @@ package object util {
   def decode[F[_]](charset: Charset): Pipe[F, Byte, String] = {
     val decoder = charset.nioCharset.newDecoder
     val maxCharsPerByte = math.ceil(decoder.maxCharsPerByte().toDouble).toInt
-    val avgBytesPerChar = math.ceil(1.0 / decoder.averageCharsPerByte().toDouble).toInt
+    val avgBytesPerChar =
```
rossabaker: If we're making changes here anyway, I guess we could roll back this whitespace formatting and not buy ourselves a bigger diff.

PR #1539, opened 2017-11-10
Remove reference to `.as()` function call from docs

yannick-cw:
It seems the syntax with `.as(..)` and the function call is not used anymore so it could be dropped from the docs or replaced with something else if there is explanation necessary?


rossabaker:
Build failure is unrelated.  Will open separate issue for that.

PR #1541, opened 2017-11-10
Fix for TransferCodingSpec random error

### LINKED ISSUE #1540: TransferCodingSpec condition is too strict
We can get slain by the Random Number Gods:

```
[info] hashCode should
[error]   x be consistent with equality (55 ms)
[error]    Falsified after 0 passed tests.
[error]    > Gave up after only 96 passed tests. 501 tests were discarded. (TransferCodingSpec.scala:9)
[error] org.http4s.TransferCodingSpec.check(TransferCodingSpec.scala:9)
```

cquiroz:
Fixes #1540


cquiroz:
Yeah the syntax is nice but it leads to sometimes not having enough tests to fulfill the criteria

PR #1543, opened 2017-11-10
Render version info and server address on startup

rossabaker:
Partial backport of #1536.

I normally wouldn't go this far out of my way to add new functionality to 0.16, but I was already on that branch, and then realized my MiMa error, and then I was sad and did it anyway.

Enjoy.


rossabaker:
Force-pushed to get rid of what looks suspiciously like the tail end of an emacs keybinding. 😊

PR #1544, opened 2017-11-10
Update dependencies in 0.16

rossabaker:


project/ScalazPlugin.scala
```diff
@@ -19,7 +19,7 @@ object ScalazPlugin extends AutoPlugin {
   override def trigger = allRequirements
 
   override lazy val projectSettings = Seq(
-    scalazVersion := sys.env.get("SCALAZ_VERSION").getOrElse("7.2.15"),
+    scalazVersion := sys.env.get("SCALAZ_VERSION").getOrElse("7.2.17"),
```
ChristopherDavenport: Non-extant scalaz version... or if it does you are ahead of their release page -  `7.2.16`

rossabaker: Typo.

PR #1545, opened 2017-11-10
Fix for bogus compareTo test

### LINKED ISSUE #1542: Random Error on ContentCodingSpec test
```
[error] ! order.eq antisymmetry (51 ms)
[error]  org.scalacheck.Gen$RetrievalError: Exception raised on property evaluation.
[error]  > ARG_0: ContentCoding(compress, org.http4s.QValue@3e8)
[error]  > ARG_1: ContentCoding(compress, org.http4s.QValue@3e8)
[error]  > ARG_2: <function1>
[error]  > Exception: org.scalacheck.Gen$RetrievalError: couldn't generate value (Gen.scala:57)
```

cquiroz:
The original tests were incorrect, Now it should be better
Fixes #1542

### DIFF
--- a/core/src/main/scala/org/http4s/QValue.scala
+++ b/core/src/main/scala/org/http4s/QValue.scala
@@ -20,6 +20,8 @@ final class QValue private (val thousandths: Int) extends AnyVal with Ordered[QV
 
   def isAcceptable: Boolean = thousandths > 0
 
+  override def toString = s"QValue(${0.001 * thousandths})"
+
   override def compare(that: QValue): Int = thousandths - that.thousandths
 
   def render(writer: Writer): writer.type =
--- a/tests/src/test/scala/org/http4s/ContentCodingSpec.scala
+++ b/tests/src/test/scala/org/http4s/ContentCodingSpec.scala
@@ -16,13 +16,11 @@ class ContentCodingSpec extends Http4sSpec {
   "compare" should {
     "be consistent with coding.compareToIgnoreCase for same quality" in {
       prop { (a: ContentCoding, b: ContentCoding) =>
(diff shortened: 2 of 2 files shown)


core/src/main/scala/org/http4s/QValue.scala
```diff
@@ -20,6 +20,8 @@ final class QValue private (val thousandths: Int) extends AnyVal with Ordered[QV
 
   def isAcceptable: Boolean = thousandths > 0
 
+  override def toString = s"QValue(${0.001 * thousandths})"
```
cquiroz: Added this to simplify debugging

tests/src/test/scala/org/http4s/ContentCodingSpec.scala
```diff
@@ -16,13 +16,11 @@ class ContentCodingSpec extends Http4sSpec {
...
     "be consistent with coding.compareToIgnoreCase for same quality" in {
       prop { (a: ContentCoding, b: ContentCoding) =>
-        (a.qValue == b.qValue) ==> (a.coding.compareToIgnoreCase(b.coding) must_== a.compare(b))
-      }
-    }
-    "be consistent with qValue.compareTo for same coding" in {
-      prop { (a: ContentCoding, b: ContentCoding) =>
-        (a.coding.toLowerCase == b.coding.toLowerCase) must_== (a.qValue.compareTo(b.qValue) == a
-          .compare(b))
+        ((a.coding, a.qValue), (b.coding, b.qValue)) match {
```
cquiroz: Should I add `ContentCoding.unapply`?

rossabaker: I think not, because the equality of what gets unapplied is not the same as the equality of the object.  I think it's similar to the argument against the case class.  Where it behaves like a product, I would say yes.  Here, I don't think so.

PR #1546, opened 2017-11-11
0.18 csrf middleware

jmcardon:
:train: CHO CHOO CSRF merge train cannot stop :train:

### DIFF
--- a/server/src/main/scala/org/http4s/server/middleware/CSRF.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/CSRF.scala
@@ -24,20 +24,24 @@ import org.http4s.util.{CaseInsensitiveString, encodeHex}
   * When a user authenticates, `embedNew` is used to send a random CSRF value as a cookie.  (Alternatively,
   * an authenticating service can be wrapped in `withNewToken`).
   *
-  * For requests that are unsafe (PUT, POST, DELETE, PATCH), services protected by the `validated` method in the
+  * By default, for requests that are unsafe (PUT, POST, DELETE, PATCH), services protected by the `validated` method in the
   * middleware will check that the csrf token is present in both the header `headerName` and the cookie `cookieName`.
   * Due to the Same-Origin policy, an attacker will be unable to reproduce this value in a
   * custom header, resulting in a `401 Unauthorized` response.
   *
-  * Requests with safe methods (such as GET, OPTIONS, HEAD) will have a new token embedded in them if there isn't one,
+  * By default, requests with safe methods (such as GET, OPTIONS, HEAD) will have a new token embedded in them if there isn't one,
   * or will receive a refreshed token based off of the previous token to mitigate the BREACH vulnerability. If a request
   * contains an invalid token, regardless of whether it is a safe method, this middleware will fail it with
   * `401 Unauthorized`. In this situation, your user(s) should clear their cookies for your page, to receive a new
   * token.
   *
+  * The default can be overridden by modifying the `predicate` in `validate`. It will, by default, check if the method is safe.
+  * Thus, you can provide some whitelisting capability for certain kinds of requests.
+  *
   * We'd like to emphasize that you please follow proper design principles in creating endpoints, as to
   * not mutate in what should otherwise be idempotent methods (i.e no dropping your DB in a GET method, or altering
-  * user data). If you choose to not to, this middleware cannot protect you.
+  * user data). Please do not use the CSRF protection from this middleware as a safety net for bad design.
+  * Every time you mutate in a GET, a puppy dies. Think of the puppies. Don't be that guy.
   *
   *
   * @param headerName your CSRF header name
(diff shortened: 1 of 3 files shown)


server/src/main/scala/org/http4s/server/middleware/CSRF.scala
```diff
@@ -0,0 +1,187 @@
...
+import javax.crypto.{KeyGenerator, Mac, SecretKey}
+
+import cats.Applicative
+import cats.data.{Kleisli, OptionT}
+import cats.effect.Sync
+import cats.syntax.all._
+import org.http4s.headers.{Cookie => HCookie}
+import org.http4s.server.HttpMiddleware
+import org.http4s.util.{CaseInsensitiveString, encodeHex}
+import org.http4s._
```
rossabaker: What editor are you using?  See if you can get all the imports to sort for this project.  In IntelliJ, I think you just delete all the groups.  (I'd still like to figure out how to check this in as a style for the IntelliJ folks.)

server/src/main/scala/org/http4s/server/middleware/CSRF.scala
```diff
@@ -0,0 +1,228 @@
...
+  /** Constructs a middleware that will check for the csrf token
+    * presence on both the proper cookie, and header values.
+    *
+    * If it is a valid token, it will then embed a new one,
+    * to effectively randomize the complete token while
+    * avoiding the generation of a new secure random Id, to guard
+    * against [BREACH](http://breachattack.com/)
+    *
+    */
+  def validate: HttpMiddleware[F] = { service =>
```
rossabaker: So what this leaves uncovered are these cases:

```scala
case GET / "stupid" => truncate() *> Gone() // should be covered
case POST / "silly" => Ok("I could have been a GET.") // doesn't need to be covered
```

We can mitigate that if `validate` takes a `Request => Boolean`, that defaults to `_.method.isSafe`.  Or we can say, "we're not going to support the abuse of HTTP methods, and we're not going to let you shoot yourself in the foot whitelisting endpoints."

Perhaps more legitimately, some users might want CSRF checks only if an Authorization or Cookie is present.  This is the [play default](https://www.playframework.com/documentation/2.6.x/ScalaCsrf).  This could also be handled with a predicate.

I'm not strongly for it or against it, but if we think we might want it, we should get it into 0.16.

jmcardon: > "we're not going to support the abuse of HTTP methods, and we're not going to let you shoot yourself in the foot whitelisting endpoints."

This is the stance I pushed with the latest commit. 

> We can mitigate that if validate takes a Request => Boolean, that defaults to _.method.isSafe

this is an interesting idea, but subject to user misuse.

> Perhaps more legitimately, some users might want CSRF checks only if an Authorization or Cookie is present. This is the play default. This could also be handled with a predicate.

We could meet half way? I think i'm not pro-Http method abuse, so we could make the condition
`if (predicate || method.isSafe) embedOrValidate else checkCSRF`.

How do you feel about this midpoint?

jmcardon: Alright, on gitter we decided a `predicate: Request => Boolean = _.method.IsSafe` is a decent default.

Commenting here for the sake of documenting.

server/src/main/scala/org/http4s/server/middleware/CSRF.scala
```diff
@@ -177,20 +186,21 @@ object CSRF {
...
+  val SigningAlgo: String = "HmacSHA1"
+  val SHA1ByteLen: Int = 20
+  val CSRFTokenLength: Int = 32
+
   /** An instance of SecureRandom to generate
     * tokens, properly seeded:
     * https://tersesystems.com/blog/2015/12/17/the-right-way-to-use-securerandom/
     */
-  private val CachedRandom = {
+  private val InitialSeedArraySize: Int = 20
```
jmcardon: I wanted to get rid of the "magic number" warnings.

server/src/main/scala/org/http4s/server/middleware/CSRF.scala
```diff
@@ -24,20 +24,24 @@ import org.http4s.util.{CaseInsensitiveString, encodeHex}
...
   * token.
   *
+  * The default can be overridden by modifying the `predicate` in `validate`. It will, by default, check if the method is safe.
+  * Thus, you can provide some whitelisting capability for certain kinds of requests.
+  *
   * We'd like to emphasize that you please follow proper design principles in creating endpoints, as to
   * not mutate in what should otherwise be idempotent methods (i.e no dropping your DB in a GET method, or altering
-  * user data). If you choose to not to, this middleware cannot protect you.
+  * user data). Please do not use the CSRF protection from this middleware as a safety net for bad design.
+  * Every time you mutate in a GET, a puppy dies. Think of the puppies. Don't be that guy.
```
rossabaker: I chuckled, but in the interest of concision and not horrifying people unfamiliar with the idiom, this is probably a stronger scaladoc minus this line. :)

rossabaker:
Test failure is still #1542.

jmcardon:
Sorry, I added one final test since it's not merged yet for completeness: 

I want to double check that the value for a different token is essentially the raw generated secure random initial value, with a different nonce + signature. This is a test for both efficiency and randomization of the value to mitigate BREACH. Even if none of the code changed and this condition was correct, it's worthwhile to test this in case of any changes to the class later on.

PR #1548, opened 2017-11-11
Fix #1542 Random error on ContentCoding equality

### LINKED ISSUE #1542: Random Error on ContentCodingSpec test

cquiroz:
I haven't been able to reproduce this locally but I suspect this is happening because `QValue` is an `AnyVal`. I have changed compare to use the `thousandths` element for comparisons

### DIFF
--- a/testing/src/main/scala/org/http4s/testing/ArbitraryInstances.scala
+++ b/testing/src/main/scala/org/http4s/testing/ArbitraryInstances.scala
@@ -208,7 +208,7 @@ trait ArbitraryInstances {
   def genContentCodingNoQuality: Gen[ContentCoding] =
     Gen.frequency(
       (10, oneOf(ContentCoding.standard.values.toSeq)),
-      (2, genToken.filter(!_.startsWith("*")).map(ContentCoding.unsafeFromString))
+      (2, genToken.map(ContentCoding.unsafeFromString))
     )
 
   implicit val arbitraryContentCoding: Arbitrary[ContentCoding] =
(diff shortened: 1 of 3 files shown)


cquiroz:
Here's an update. I'm using `genToken` to produce arbitrary `ContentCoding` as it should be according to the RFC. However, I'm filtering out the ones starting with `*` as they break the parser which makes sense to me 

This runs on both scala versions locally but it seems that's not guarantee that it'd work on travis/appveyor

testing/src/main/scala/org/http4s/testing/ArbitraryInstances.scala
```diff
@@ -208,7 +208,7 @@ trait ArbitraryInstances {
   def genContentCodingNoQuality: Gen[ContentCoding] =
     Gen.frequency(
       (10, oneOf(ContentCoding.standard.values.toSeq)),
-      (2, Gen.alphaStr.filter(_.nonEmpty).map(ContentCoding.unsafeFromString))
+      (2, genToken.filter(!_.startsWith("*")).map(ContentCoding.unsafeFromString))
```
rossabaker: Any `filter` in a generator wrapped in an arbitrary is going to be a potential problem until https://github.com/rickynils/scalacheck/pull/316 gets released.

I think `*foo` is a valid token, and thus a valid content coding.  (A bizarre one, but a valid one.)  I think the problem is in the parser, and this filter should be dropped.

cquiroz:
Changed the `ContentCoding` parser to support any token. Thanks for the comments, I think it is much better now

PR #1551, opened 2017-11-13
Update `HttpService[F].apply` docs

yannick-cw:
It seems that the docs for the `apply` is outdated, as there is no `org.http4s.Response.notFoundFor` anymore.
I suggest replacing it with the `OptionT.none` response.

### DIFF
--- a/core/src/main/scala/org/http4s/HttpService.scala
+++ b/core/src/main/scala/org/http4s/HttpService.scala
@@ -14,9 +14,8 @@ object HttpService extends Serializable {
   def lift[F[_]: Functor](f: Request[F] => F[Response[F]]): HttpService[F] =
     Kleisli(f.andThen(OptionT.liftF(_)))
 
-  /** Lifts a partial function to an `HttpService`. Responds with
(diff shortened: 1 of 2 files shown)


rossabaker:
At the level of this method, it's correct and worth mentioning that we return `Option.none`.

`Response.notFoundFor` still [does exist](https://github.com/http4s/http4s/blob/master/core/src/main/scala/org/http4s/Message.scala#L379).  And at the `HttpService` level, if an `Option.none` is returned, a server backend should render a 404.  Maybe that's a better comment for the `HttpService` type itself?  (If so, I would discuss it more in terms of the 404 and not in terms of the `Response.notFoundFor`, which is an implementation detail.)

yannick-cw:
Sorry I agree that `Response.notFoundFor` still exists :) I meant that it does not exist at the `apply` anymore.
Okay so you mean adding more description to this one : https://github.com/http4s/http4s/blob/1a5844460180692fb452980b98bbc41d14befbfa/core/src/main/scala/org/http4s/package.scala#L35

I'll give it a shot

PR #1552, opened 2017-11-13
synchronized clock to fix intermittent test failure

jmcardon:


server/src/test/scala/org/http4s/server/middleware/CSRFSpec.scala
```diff
@@ -1,24 +1,47 @@
...
 class CSRFSpec extends Http4sSpec {
+
+  /** Create a clock that always ticks forward once per millis() call.
+    *
+    * This is to emulate scenarios where we want to mitigate BREACH where, in
+    * a real world service, a huge number of requests wouldn't be processed
+    * before the clock at least traverses a millisecond.
+    */
+  val testClock: Clock = new Clock { self =>
+    private var clockTick = Instant.now()
```
rossabaker: With an AtomicLong, you wouldn't need to synchronize.  Doesn't matter here, but that's how I'd make a general purpose clock like this.

PR #1558, opened 2017-11-21
Add mediatype 'application/x-ndjson'

ashokkumar:
I would like to add the media-type `application/x-ndjson` https://github.com/ndjson/ndjson-spec.
We are stream json data and GZip middleware is not working as the `application/x-ndjson` is not in the list of compressible MediaType.

### DIFF
--- a/core/src/main/scala/org/http4s/MediaType.scala
+++ b/core/src/main/scala/org/http4s/MediaType.scala
@@ -234,6 +234,7 @@ object MediaType extends Registry {
   val `application/java-archive` = app("java-archive", uncompressible, binary, "jar", "war", "ear")
   val `application/javascript` = app("javascript", compressible, notBinary, "js")
   val `application/json` = app("json", compressible, binary, "json") // we treat JSON as binary, since it's encoding is not variable but defined by RFC4627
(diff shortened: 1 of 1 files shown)


lukestephenson:
This PR is breaking the build because there are two MediaTypes with the same `json` extension:
```
Caused by: java.lang.IllegalArgumentException: requirement failed: Extension 'json' clash: media-types 'MediaType(application/json)' and 'MediaType(application/json)'
```

Another alternative to make this more customisable would be to supply a function which evaluates if the response is compressible.  I.e.:
```
  def apply[F[_]: Functor](
      service: HttpService[F],
      bufferSize: Int = 32 * 1024,
      level: Int = Deflater.DEFAULT_COMPRESSION,
      isZippable: Response[F] => Boolean = existingLogicByDefault)
```

rossabaker:
Generally, we provide a constant for everything in the [IANA registry](https://www.iana.org/form/media-types).  This format is not there yet, though they are [working on it](https://github.com/ndjson/ndjson-spec/issues/21).  I imagine the media type would change from the `x-*` before being registered.

@ashokkumar Do you think @lukestephenson's idea would provide what you need?  That seems like a very good idea to me.

ashokkumar:
@rossabaker Yes, What @lukestephenson suggested will solve our problem. Will raise a different PR  with the change.

PR #1563, opened 2017-11-24
Retain Logging Concept With Less Bulk

ChristopherDavenport:
Should preserve semantics and chunkiness rather than pulling a single element at a time from the queue. This is more of a patch for optimizations than a fix of any kind.

### DIFF
--- a/server/src/main/scala/org/http4s/server/middleware/RequestLogger.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/RequestLogger.scala
@@ -27,24 +27,27 @@ object RequestLogger {
         OptionT(
           Logger.logMessage[F, Request[F]](req)(logHeaders, logBody)(logger) *> service(req).value)
       else
-        OptionT.liftF(async.refOf[F, Vector[Byte]](Vector.empty[Byte])).flatMap { vec =>
-          val newBody =
-            Stream
+        OptionT
+          .liftF(async.refOf[F, Vector[Segment[Byte, Unit]]](Vector.empty[Segment[Byte, Unit]]))
+          .flatMap { vec =>
+            val newBody = Stream
               .eval(vec.get)
-              .flatMap(Stream.emits(_).covary[F])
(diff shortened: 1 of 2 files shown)


rossabaker:
What is the difference between this and just doing `dequeueAvailable` with the old way?

ChristopherDavenport:
`dequeueAvailable` blocks forever if nothing is ever put into the queue. To be fair I haven't been able to determine what causes it to hang. But when swapping from the old way to `dequeueAvailable` caused an infinite hang in our tests that went away. I remember running into it the first time which is why we have the implementation we  had. 

I'm happy to go investigate as well. As I don't like my lack of understanding. As we are the producer and consumer internally I don't see how we could run into problems with this implementation and I believe most of it is similar to queue except it expects the possibility of multiple consumers.

ChristopherDavenport:
As Far as performance goes for this implementation seems to be in a good place. 

Though together this [performance analysis app](https://gist.github.com/ChristopherDavenport/fc641b5ee57e597f34f3b2843c25eac0)

```
Ops: 5000
Execution of : Ref of Segments took 91.664 s
Execution of : Ref of Chunks took 260.235 s
Execution of : Ref of Large Vector took 195.108 s
Execution of : Queue DequeueAvailable took 234.127 s
Execution of : Queue Dequeue took 306.132 s
```

server/src/main/scala/org/http4s/server/middleware/RequestLogger.scala
```diff
@@ -27,15 +27,15 @@ object RequestLogger {
         OptionT(
           Logger.logMessage[F, Request[F]](req)(logHeaders, logBody)(logger) *> service(req).value)
       else
-        OptionT.liftF(async.unboundedQueue[F, Byte]).flatMap { queue =>
+        OptionT.liftF(async.refOf[F, Vector[Byte]](Vector.empty[Byte])).flatMap { vec =>
```
ChristopherDavenport: Looks like I missed the request logger. :-1:

PR #1564, opened 2017-11-24
Deprecate Http4s StreamApp

ChristopherDavenport:
To be merged after we have upgraded to a new version of fs2.


aeons:
Aren’t there some docs that need updates as well for this?

rossabaker:
Was handled instead by a deprecated type alias.

PR #1565, opened 2017-11-24
Logger Rework Sketch

ChristopherDavenport:
Does not compile due to `a.Self` nonsense. 

Looking for feedback on the implementation. Due to the nature of the function I think it may be easiest to just split it for `Request[F]` and `Response[F]` since its so messy at presently trying to work with both simultaneously. This offers far more customize-ability. But also gives enough rope that you could end up in major trouble. This puts us so we could easily implement logger middlewares for both client and server with the same set of tools.

I'm sure it can be much better but I feel a little ungrounded at the moment staring at it. The lack of a message typeclass really puts us in a bad position for this.


core/src/main/scala/org/http4s/util/Logging.scala
```diff
@@ -0,0 +1,118 @@
...
+
+object Logging {
+
+
+  def logMessage[F[_], A <: Message[F]](
+                                        f: Kleisli[F, A, String],
+                                        cacheBody: A => Boolean = {_: A => true},
+                                        forceBody: A => Boolean = {_: A => false},
+                                        logLevel: A => LogLevel = {_: A => org.log4s.Trace},
+                                        a: A,
```
rossabaker: I'd put things without defaults ahead of things with defaults.

core/src/main/scala/org/http4s/util/Logging.scala
```diff
@@ -0,0 +1,118 @@
...
+import scala.concurrent.ExecutionContext
+
+object Logging {
+
+
+  def logMessage[F[_], A <: Message[F]](
+                                        f: Kleisli[F, A, String],
+                                        cacheBody: A => Boolean = {_: A => true},
+                                        forceBody: A => Boolean = {_: A => false},
+                                        logLevel: A => LogLevel = {_: A => org.log4s.Trace},
```
rossabaker: If we're already passing in `a`, what do we get out of passing in functions of `A` for things that will usually be constant with respect to A anyway?

core/src/main/scala/org/http4s/util/Logging.scala
ChristopherDavenport: This was originally curried and I put it back in at some point. I think in general it should probably be curried. As that leaves it as a function that works as a logger taking the thing that does the logger action and the thing to log.

core/src/main/scala/org/http4s/util/Logging.scala
ChristopherDavenport: The idea is that those functions allow the user to determine the functionality of the function. However while it may be constant in regard to `A` that `A` may be a `Request[F]` or a `Response[F]`  so as below this means a function that matches on _:NoBody, etc. But allows the user to pick based on criteria specified in the super - eventually typeclass.

ChristopherDavenport:
So what I have discovered is that since the logger happens as an effect on evaluation of the stream, if the entitybody is never evaluated. Like in `GET` requests then the logging never appears, only the Request which has a body is logged. So `forceBody` forces edge evaluation of the logging but destroys streaming. `cacheBody` means that if we are not caching the body we do the logging effect, but if that effect consumes the body... You're in trouble. Hence why I set default to true. If we are caching the body then we are using the side-loading technique to push values out as they stream and on final evaluation logging is done. This is the scenario where we see non-evaluated bodies, or non-extant bodies not logging.

rossabaker:
If this is abandoned, go ahead and remove my label of shame.  If it's not, please reopen.

PR #1568, opened 2017-11-28
Limit the amount of data pulled from the stream

### LINKED ISSUE #1561: Memory leak in GZip
When the GZip middleware handles a request (ie request has headers / response is zippable) the wrapped response holds onto the complete response stream until completed resulting in a memory leak for large responses.

A fairly basic way of investigating is the following spec:
```
    "have no memory leaks for a large response" in {
      val streamLength = 2000000
      val chunkLength = 1000
      val service: HttpService[IO] = HttpService[IO] {
        case GET -> Root => Ok(Stream.range(0, streamLength).map(_ => "abc" * chunkLength).covary[IO])
      }

      val gzipService: HttpService[IO] = GZip(service)

      val req: Request[IO] = Request[IO](Method.GET, Uri.uri("/"))
        .putHeaders(`Accept-Encoding`(ContentCoding.gzip))

      val actualLength: IO[Option[Byte]] =
        gzipService.orNotFound(req).flatMap { response =>
          response.body.runLast
        }

      actualLength.unsafeRunSync() must beSome
    }
```

When running this, the heap shows the response body stream becoming very large.
image

Note that if the spec is updated to remove the `.putHeaders(`Accept-Encoding`(ContentCoding.gzip))` I can't observe any memory leak.  The test is still slow, but that's probably expected.

(Note this was originally observed monitoring actual responses to a http4s service and I've tried stripping it back to a minimal test case to reproduce)

lukestephenson:
Fixes #1561

Prevents OutOfMemory in GZip with streams with large
segments.  Eg:
def bigStream: fs2.Stream[IO, Byte] = fs2.Stream.range(1, 2014922827).map(_.toByte).covary[IO]

### DIFF
--- a/server/src/main/scala/org/http4s/server/middleware/GZip.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/GZip.scala
@@ -60,7 +60,7 @@ object GZip {
     val trailerGen = new TrailerGen()
     val b = chunk(header) ++
       resp.body
-        .through(trailer(trailerGen))
+        .through(trailer(trailerGen, bufferSize.toLong))
         .through(
           deflate(
             level = level,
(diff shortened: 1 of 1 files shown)


rossabaker:
I wonder if this is also a problem on 0.17. 🤔

PR #1569, opened 2017-11-28
add new media type application/vnd.api+json

Doikor:
see https://www.iana.org/assignments/media-types/application/vnd.api+json for its specification

### DIFF
--- a/core/src/main/scala/org/http4s/MediaType.scala
+++ b/core/src/main/scala/org/http4s/MediaType.scala
@@ -260,6 +260,7 @@ object MediaType extends Registry {
   val `application/postscript` = app("postscript", compressible, binary, "ai", "eps", "ps")
   val `application/rss+xml` = app("rss+xml", compressible, notBinary, "rss")
   val `application/soap+xml` = app("soap+xml", compressible, notBinary)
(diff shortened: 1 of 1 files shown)


core/src/main/scala/org/http4s/MediaType.scala
```diff
@@ -260,6 +260,7 @@ object MediaType extends Registry {
   val `application/postscript` = app("postscript", compressible, binary, "ai", "eps", "ps")
   val `application/rss+xml` = app("rss+xml", compressible, notBinary, "rss")
   val `application/soap+xml` = app("soap+xml", compressible, notBinary)
+  val `application/vnd.api+json` = app("vnd.api+json", compressible, binary) // we treat JSON as binary, since it's encoding is not variable but defined by RFC4627
```
ChristopherDavenport: Does this need to be added to the types in logger that actually gets parsed as text despite being binary?

Doikor:
@rossabaker I guess I'll fix the typos now as I'm changing the file.

core/src/main/scala/org/http4s/MediaType.scala
Doikor: Not really sure. Just doing it the same as the other json media types.

rossabaker:
Appveyor failure is spurious.  Merging.

PR #1570, opened 2017-11-28
Improve client timeout and wait queue handling

eklavya:


blaze-client/src/main/scala/org/http4s/client/blaze/PooledHttp1Client.scala
```diff
@@ -8,18 +8,21 @@ import cats.effect._
...
     * @param maxWaitQueueLimit maximum number requests waiting for a connection at any specific time
     * @param maxConnectionsPerRequestKey Map of RequestKey to number of max connections
+    * @param waitExpiryTime timeout duration for request waiting in queue, default value -1, never expire
     * @param config blaze client configuration options
     */
   def apply[F[_]: Effect](
       maxTotalConnections: Int = DefaultMaxTotalConnections,
       maxWaitQueueLimit: Int = DefaultMaxWaitQueueLimit,
       maxConnectionsPerRequestKey: RequestKey => Int = _ => DefaultMaxTotalConnections,
+      waitExpiryTime: RequestKey => Int = _ => DefaultWaitExpiryTime,
```
rossabaker: How about a `Duration` for the type here instead of `Int`?

client/src/main/scala/org/http4s/client/PoolManager.scala
```diff
@@ -15,18 +19,22 @@ private final class PoolManager[F[_], A <: Connection[F]](
...
+      at: Instant)
 
   private[this] val logger = getLogger
 
   private var isClosed = false
   private var curTotal = 0
   private val allocated = mutable.Map.empty[RequestKey, Int]
   private val idleQueues = mutable.Map.empty[RequestKey, mutable.Queue[A]]
-  private val waitQueue = mutable.Queue.empty[Waiting]
+  private var waitQueue = mutable.Queue.empty[Waiting]
```
rossabaker: Why did this need to become a var?

blaze-client/src/test/scala/org/http4s/client/blaze/PooledClientSpec.scala
```diff
@@ -12,14 +13,21 @@ import scala.util.Random
 import org.http4s.client.testroutes.GetRoutes
 import org.http4s.client.JettyScaffold
 
-class MaxConnectionsInPoolSpec extends Http4sSpec {
+import scala.concurrent.Await
```
rossabaker: Can we sort all these imports alphabetically?  That's how it's done in most places in the code.  I think this one got IntelliJ'ed.

client/src/main/scala/org/http4s/client/PoolManager.scala
```diff
@@ -65,6 +73,11 @@ private final class PoolManager[F[_], A <: Connection[F]](
   private def numConnectionsCheckHolds(key: RequestKey): Boolean =
     curTotal < maxTotal && allocated.getOrElse(key, 0) < maxConnectionsPerRequestKey(key)
 
+  private def isExpired(k: RequestKey, t: Instant): Boolean =
+    waitExpiryTime(k) != -1 && t
+      .plus(waitExpiryTime(k).toLong, ChronoUnit.SECONDS)
+      .isBefore(Instant.now())
```
rossabaker: This gets a lot cleaner if the duration suggestion is followed.  The `-1` is `!_.isFinite`, and there's a clear time unit.

blaze-client/src/main/scala/org/http4s/client/blaze/PooledHttp1Client.scala
eklavya: Ok.

blaze-client/src/test/scala/org/http4s/client/blaze/PooledClientSpec.scala
eklavya: I was relying on scalafmt to do that.

client/src/main/scala/org/http4s/client/PoolManager.scala
eklavya: It was a val before and we were dequeuing for all expired one by one. Now we want to use span while taking out expired requests and just assign the new queue. Let me know if you feel this needs to change.

client/src/main/scala/org/http4s/client/PoolManager.scala
eklavya: Ok.

eklavya:
You are right, we do need to wait in that case, although we take care of the next expired ones immediately in "findFirstAllowedWaiter". But I can't think of an easy solution to this, ideas?

rossabaker:
We could have a scheduler proactively clean out things caught in the wait queue.  I don't know what semantics we should have for a request that's in flight.  Imagine a non-idempotent POST that has been submitted but not yet responded.

eklavya:
I thought about a scheduler but I thought it maybe overkill. Do we really want this?

I don’t understand the second point. If the request is in flight that means it’s not in wait queue, right?

rossabaker:
From the client perspective, do we care how long we wait for a connection from the pool, or do we care how long it takes from submission to getting a timeout response?  Which wait are we trying to fix here?

If we're trying to get timeout responses in a timely fashion, then I think we want a scheduler, and need to account for responses whose timeout is partially but incompletely consumed while in the wait queue. (My second point).  I _think_ this is a positive, but it gets ugly when a non-idempotent request is in flight.

client/src/main/scala/org/http4s/client/PoolManager.scala
```diff
@@ -74,10 +74,11 @@ private final class PoolManager[F[_], A <: Connection[F]](
   private def numConnectionsCheckHolds(key: RequestKey): Boolean =
     curTotal < maxTotal && allocated.getOrElse(key, 0) < maxConnectionsPerRequestKey(key)
 
-  private def isExpired(k: RequestKey, t: Instant): Boolean =
-    waitExpiryTime(k).isFinite() && t
-      .plus(waitExpiryTime(k).toSeconds, ChronoUnit.SECONDS)
-      .isBefore(Instant.now())
+  private def isExpired(t: Instant): Boolean = {
+    val elapsed = Instant.now().getEpochSecond - t.getEpochSecond
```
rossabaker: Why is this in seconds instead of milliseconds?  Is there a performance boost for the loss of granularity?

client/src/main/scala/org/http4s/client/PoolManager.scala
```diff
@@ -74,10 +74,11 @@ private final class PoolManager[F[_], A <: Connection[F]](
...
+    (requestTimeout.isFinite() && elapsed >= requestTimeout.toSeconds) || (responseHeaderTimeout
```
rossabaker: I am feeling stupid.  What's the difference between requestTimeout and responseHeaderTimeout?  They both seem to be enforced in the same place.

client/src/main/scala/org/http4s/client/PoolManager.scala
eklavya: Because the timeouts are specified in seconds (aren't they?).

client/src/main/scala/org/http4s/client/PoolManager.scala
eklavya: I am not sure what you mean. Any of them could have timed out, we need to check for both, no?

client/src/main/scala/org/http4s/client/PoolManager.scala
rossabaker: Do they have to be?  I've had subsecond SLAs before.

client/src/main/scala/org/http4s/client/PoolManager.scala
rossabaker: I don't see any difference in how these two properties are enforced in this function, or anywhere else.  What is the semantic difference between them?

eklavya: I may have the wrong interpretation of the two. What happens when requestTimeout is 2 seconds and responseHeaderTimeout is 1 second? responseHeaderTimeout can timeout independent of requestTimeout and vice versa. Is this correct?

rossabaker: I don't understand how they're being enforced such that they're two different properties.  They're both being compared to the same elapsed time.

eklavya: responseHeaderTimeout not set
requestTimeout set

if we check only for responseHeaderTimeout we miss requestTimeout

responseHeaderTimeout set to 1 sec
requestTimeout set to 2 sec

if we only check for requestTimeout and we start after 1 sec and finish before 2 sec we are ok  but we missed responseHeaderTimeout

etc. What am I missing?

rossabaker: There are two properties on the config.  Both of them are enforced right here, based on the same elapsed time.  i.e., the clock on one does not start ticking before the other.  Effectively, we're enforcing a single timeout that is `min(requestTimeout, responseHeaderTimeout)`, no?

eklavya: Yes.

rossabaker: So if the property we're enforcing is just a min of two properties, why two properties?  I would think that the start or end time we're measuring against would be different.

client/src/test/scala/org/http4s/client/testroutes/GetRoutes.scala
```diff
@@ -4,16 +4,19 @@ import cats.effect._
...
   val ChunkedPath = "/chunked"
+  val DelayedPath = "/delayed"
 
   val getPaths: Map[String, Response[IO]] = {
     Map(
       SimplePath -> Response[IO](Ok).withBody("simple path"),
       ChunkedPath -> Response[IO](Ok).withBody(
-        Stream.emits("chunk".toSeq.map(_.toString)).covary[IO])
+        Stream.emits("chunk".toSeq.map(_.toString)).covary[IO]),
+      DelayedPath -> IO(Thread.sleep(1000L)) *> Response[IO](Ok).withBody("delayed path")
```
rossabaker: We can sleep with an fs2 scheduler and not block a thread.  This should make tests more robust.

rossabaker:
I _think_ this is correct.  We should take the nasty disclaimers out of the scaladoc.

I'd love to get an automated test around this, but am not yet sure the best way to do so.

rossabaker:
I removed the limitations from the scaladoc.

Note that our response header timeout is from when the request is submitted via the client, and not when the request goes over the wire.  (This is different from the [cloudflare document](https://blog.cloudflare.com/the-complete-guide-to-golang-net-http-timeouts/#clienttimeouts) we've all been using as a discussion starter.)

client/src/main/scala/org/http4s/client/PoolManager.scala
```diff
@@ -259,10 +279,17 @@ private final class PoolManager[F[_], A <: Connection[F]](
     }
   }
 
-  private def findFirstAllowedWaiter =
+  private def findFirstAllowedWaiter = {
+    val (expired, rest) = waitQueue.span(w => isExpired(w.at))
```
jmcardon: Should this be partition? Do we want the longest prefix of the expired requests, or all of them?

client/src/main/scala/org/http4s/client/PoolManager.scala
```diff
@@ -1,32 +1,41 @@
...
 import scala.concurrent.ExecutionContext
+import scala.concurrent.duration._
 import scala.util.Random
 
 private final class PoolManager[F[_], A <: Connection[F]](
     builder: ConnectionBuilder[F, A],
     maxTotal: Int,
     maxWaitQueueLimit: Int,
     maxConnectionsPerRequestKey: RequestKey => Int,
+    responseHeaderTimeout: Duration,
```
jmcardon: If you are doing an `isFinite` check with this every time, is there a reason this can't be a `FiniteDuration`?

client/src/main/scala/org/http4s/client/PoolManager.scala
```diff
@@ -65,6 +74,12 @@ private final class PoolManager[F[_], A <: Connection[F]](
   private def numConnectionsCheckHolds(key: RequestKey): Boolean =
     curTotal < maxTotal && allocated.getOrElse(key, 0) < maxConnectionsPerRequestKey(key)
 
+  private def isExpired(t: Instant): Boolean = {
+    val elapsed = Instant.now().toEpochMilli - t.toEpochMilli
+    (requestTimeout.isFinite() && elapsed >= requestTimeout.toMillis) || (responseHeaderTimeout
+      .isFinite() && elapsed >= responseHeaderTimeout.toMillis)
```
jmcardon: same re: comment above. isFinite() seems redundant.

client/src/main/scala/org/http4s/client/PoolManager.scala
eklavya: We need to support no timeouts (infinite timeouts).

jmcardon: oh herp derp. `Duration.Inf` is a thing. Forgot. My bad :+1:

blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClient.scala
```diff
@@ -71,6 +71,8 @@ object BlazeClient {
               }
 
             case Left(e) =>
+              println("FUCK ME")
```
rsoeldner: :D Yes, I like it

client/src/main/scala/org/http4s/client/PoolManager.scala
rossabaker: Great question.  The oldest should be at the front of the queue, so I think `span` is a fair optimization that prevents us from iterating the entire queue.

If this were a `takeWhile`, and we `drop`ped its size, we could get rid of the var.

client/src/main/scala/org/http4s/client/PoolManager.scala
```diff
@@ -272,7 +302,18 @@ private final class PoolManager[F[_], A <: Connection[F]](
     * @return An effect of Unit
     */
   override def invalidate(connection: A): F[Unit] =
-    F.delay(disposeConnection(connection.requestKey, Some(connection)))
+    F.delay(synchronized {
+      decrConnection(connection.requestKey)
+      if (!connection.isClosed) connection.shutdown()
+      findFirstAllowedWaiter match {
```
rossabaker: This could be consolidated with some code in releaseNonRecyclable.  Just trying to get this working right now, to refactor more later.

client/src/main/scala/org/http4s/client/PoolManager.scala
```diff
@@ -147,6 +162,9 @@ private final class PoolManager[F[_], A <: Connection[F]](
                 logger.debug(s"Active connection not found. Creating new one. $stats")
                 createConnection(key, callback)
 
+              case None if maxConnectionsPerRequestKey(key) <= 0 =>
```
rossabaker: This is a new enhancement, but it prevents a hang in the spec, so bolted on to this.

rossabaker:
I think the formatting error in Travis is something I fixed in master in 8ed6be97013692f22dfe5d3b08361dc5bd784d20

eklavya:
Shouldn’t we also drain wait queue (fail all) while shutting down pool in PoolManager.scala? On mobile don’t really know if you have done it already.

rossabaker:
I was aiming for a graceful shutdown: accept no new connections, but fulfill everything that was already queued.  I _think_ that's achieved in 54cfe4a.

rossabaker:
Sorry, I accidentally pushed my master to your master on merge, and then I couldn't fix it.  This is all on #1613, minus the discussion. :(

PR #1572, opened 2017-11-29
Message Typeclass lands

ChristopherDavenport:
Core Compiles Message Typeclass. 

Please tell me your thoughts. I will flesh out this design to the auxilary project. I also may have gutted things we care about. But as a major redesign all the deprecations cause much more trouble than they are worth preserving. They can be readded if necessary at the end.


testing/src/main/scala/org/http4s/testing/Http4sMatchers.scala
```diff
@@ -19,28 +20,28 @@ trait Http4sMatchers extends Matchers with IOMatchers {
       r.unsafeRunSync.aka("the returned")
     }
 
-  def haveBody[A: EntityDecoder[IO, ?]](a: ValueCheck[A]): Matcher[Message[IO]] =
-    returnValue(a) ^^ { m: Message[IO] =>
+  def haveBody[M[_[_]], A: EntityDecoder[IO, ?]](a: ValueCheck[A])(implicit M: Message[M, IO]): Matcher[M[IO]] =
```
ChristopherDavenport: Note: these now return the `M[_[_]]` in the return type. Which shows the pattern this allows people to work on safely.

tests/src/test/scala/org/http4s/EntityDecoderSpec.scala
```diff
@@ -188,7 +202,7 @@ class EntityDecoderSpec extends Http4sSpec with PendingUntilFixed {
         decoder1
           .orElse(decoder2)
           .orElse(failDecoder)
-          .decode(Request(), strict = true) must returnLeft(MediaTypeMissing(expectedMediaRanges))
+          .decode(Request[IO](), strict = true) must returnLeft(MediaTypeMissing(expectedMediaRanges))
```
ChristopherDavenport: We had these requests sitting around that were inferred to nothing. `Request[Nothing](...)` cannot infer the implicit functions if these circumstances

tests/src/test/scala/org/http4s/MessageSpec.scala
```diff
@@ -14,27 +15,28 @@ class MessageSpec extends Http4sSpec {
...
       "not contain remote connection info when not present" in {
-        val r = Request()
+        val r = Request[IO]()
         r.server must beNone
         r.remote must beNone
       }
 
       "be utilized to determine the address of server and remote" in {
-        val r = Request().withAttribute(
+        val r = Request[IO]().withAttributeEntry(
```
ChristopherDavenport: `withAttribute` overload needed to be switch out with a specific method so `withAttributeEntry`

core/src/main/scala/org/http4s/syntax/AllSyntax.scala
```diff
@@ -4,8 +4,6 @@ package syntax
 trait AllSyntax
```
jmcardon: Shouldn't `Message.messSyntax` be included here?

server/src/main/scala/org/http4s/server/middleware/Logger.scala
```diff
@@ -23,11 +24,11 @@ object Logger {
       )
     )
 
-  def logMessage[F[_], A <: Message[F]](message: A)(
+  def logMessage[M[_[_]], F[_]](message: M[F])(
```
jmcardon: nothing wrong with this, just wanted to point out that getting rid of the covariant bound here looks great :+1:

core/src/main/scala/org/http4s/EntityDecoder.scala
```diff
@@ -72,70 +75,98 @@ object EntityDecoder extends EntityDecoderInstances {
...
-        if (strict) {
-          msg.headers.get(`Content-Type`) match {
-            case Some(c) if matchesMediaType(c.mediaType) => f(msg)
-            case Some(c) => DecodeResult.failure(MediaTypeMismatch(c.mediaType, consumes))
-            case None if matchesMediaType(UndefinedMediaType) => f(msg)
-            case None => DecodeResult.failure(MediaTypeMissing(consumes))
-          }
-        } else {
-          f(msg)
+  def apply[M[_[_]], F[_], T](implicit ev: EntityDecoder[F, T]): EntityDecoder[F, T] = ev
```
rossabaker: Where is `M` used here?

core/src/main/scala/org/http4s/EntityDecoder.scala
```diff
@@ -148,45 +179,69 @@ trait EntityDecoderInstances {
...
     new EntityDecoder[F, T] {
-      override def decode(msg: Message[F], strict: Boolean): DecodeResult[F, T] =
+      override def decode[M[_[_]]](msg: M[F], strict: Boolean)(implicit M: Message[M, F]): DecodeResult[F, T] =
         DecodeResult(msg.body.run *> F.raiseError(t))
       override def consumes: Set[MediaRange] = Set.empty
     }
 
   implicit def binary[F[_]: Effect]: EntityDecoder[F, Chunk[Byte]] =
-    EntityDecoder.decodeBy(MediaRange.`*/*`)(collectBinary[F])
+    new EntityDecoder[F, Chunk[Byte]] {
```
rossabaker: I have been awake all night on a plane and am feeling stupid.  Why is `decodeBy` no longer viable?

core/src/main/scala/org/http4s/Request.scala
```diff
@@ -0,0 +1,233 @@
...
+import java.io.File
+import java.net.{InetAddress, InetSocketAddress}
+
+import cats.data.NonEmptyList
+import cats.Monad
+import cats.implicits._
+
+import org.http4s.headers.{Authorization, Host, `X-Forwarded-For`}
+import org.http4s.server.ServerSoftware
+import Message.messSyntax._
```
rossabaker: Our imports are usually alphabetical.  It's the simplest rule that works across tools.  (With some grumbling around `_root_.io.circe` that we haven't quite worked out.)

core/src/main/scala/org/http4s/syntax/AllSyntax.scala
rossabaker: :+1:  I don't want to add to the import tax.

core/src/main/scala/org/http4s/syntax/AllSyntax.scala
```diff
@@ -4,8 +4,6 @@ package syntax
...
     extends AnyRef
     with AsyncSyntax
-    with EffectResponseSyntax
-    with EffectRequestSyntax
```
rossabaker: Getting rid of these is going to be rough on migrants from < 0.18.  Would be nice if these still compiled.  Not mandatory, but worth an effort.

core/src/main/scala/org/http4s/Message.scala
```diff
@@ -55,328 +30,157 @@ sealed trait Message[F[_]] extends MessageOps[F] { self =>
...
+  def transformHeaders(m: M[F])(f: Headers => Headers): M[F] =
+    change(m)(body(m), f(headers(m)), attributes(m))
 
-  def withPathInfo(pi: String)(implicit F: Functor[F]): Self =
-    withUri(uri.withPath(scriptName + pi))
+  def replaceAllHeaders(m: M[F])(headers: Headers): M[F] =
+    change(m)(body(m), headers, attributes(m))
 
-  lazy val pathTranslated: Option[File] = attributes.get(Keys.PathTranslated)
+  def replaceAllHeadersWith(m: M[F])(headers: Header*): M[F] =
```
rossabaker: What prompted this name change?  This seemed to ripple hard in the tests.

core/src/main/scala/org/http4s/Message.scala
```diff
@@ -55,328 +30,157 @@ sealed trait Message[F[_]] extends MessageOps[F] { self =>
...
 
-object Response {
-  private[this] val pureNotFound: Response[Pure] =
-    Response(
-      Status.NotFound,
-      body = Stream("Not found").through(text.utf8Encode),
-      headers = Headers(`Content-Type`(MediaType.`text/plain`, Charset.`UTF-8`)))
+  def apply[M[_[_]], F[_]](implicit mess: Message[M, F]): Message[M, F] = mess
+
+  object messSyntax {
```
rossabaker: Why doesn't this live in the syntax package like our other syntax?

core/src/main/scala/org/http4s/EntityDecoder.scala
ChristopherDavenport: It's not thats a leftover from when I had the `EntityDecoder` parameterized on `M`

core/src/main/scala/org/http4s/EntityDecoder.scala
ChristopherDavenport: Higher-Rank types. I ran into the same question and was set straight by @SystemFw. Effectively As the entity decoders functions are themselves parameterized by the `M[_[_]]` which is determined by the caller. It is impossible to construct values in a method. This is equivalent to how you can have an easy value for creating a `Show` but must invoke `new Functor[F[_]]` when creating a functor as map is itself parameterized.

core/src/main/scala/org/http4s/Message.scala
ChristopherDavenport: Name overloading did not work on implicit classes.

core/src/main/scala/org/http4s/Request.scala
ChristopherDavenport: Absolutely. Sketch builds will need to be cleaned up. This was a giant experiment that fought me every step of the way.

core/src/main/scala/org/http4s/syntax/AllSyntax.scala
ChristopherDavenport: The problem is implicit on implicit here. There is no core set of traits to expand anymore. It could be manageable by creating any EffectMessage Syntax. But in the process these all became concrete rather than traits. We have a `Request[F]` and `Response[F]`, that happen to both satisfy `Message[M[_[_]], F[_]]`.

aeons:
This looks great.

I really don't like the `messSyntax` name though :)

PR #1573, opened 2017-12-03
Fixed a typo in Headers.`Content-Length` docstring

amrhassan:


core/src/main/scala/org/http4s/headers/Content-Length.scala
```diff
@@ -9,7 +9,7 @@ import org.http4s.util.Writer
   *
   * The HTTP RFCs do not specify a maximum length.  We have decided that `Long.MaxValue`
   * bytes ought to be good enough for anybody in order to avoid the irritations of `BigInt`.
-  *t
+  * 
   * @param length the length; throws an `IllegalArgumentException` if negative
```
cquiroz: Is this still true? We have been removing exception throwing form the constructors

amrhassan: It doesn't seem true. Only `unsafeFromLong` seems to be throwing.

PR #1575, opened 2017-12-03
More EntityDecoder composition methods

### LINKED ISSUE #1574: Adding more composing methods on EntityDecoder
In addition to the existing `.map[T2](f: T => T2)(implicit F: Functor[F])` method on `EntityDecoder[F[_], T]` that builds a `EntityDecoder[F[_], T2]`, it might be nice to have 

* ~~`recover`~~ `handleError`
* ~~`recoverWith`~~ `handleErrorWith`
* `transform`
* `transformWith`

That do something similar. The first two are there to help you build another `EntityDecoder[F[_], T]` by mapping and flapmapping on the failure cases respectively and the last two one would allow you to build a `EntityDecoder[F[_], T2]` while mapping and flatmapping both failure and success cases.

lloydmeta:
Closes #1574 

Just one way of doing it.

### DIFF
--- a/core/src/main/scala/org/http4s/EntityDecoder.scala
+++ b/core/src/main/scala/org/http4s/EntityDecoder.scala
@@ -45,20 +45,15 @@ trait EntityDecoder[F[_], T] { self =>
       override def consumes: Set[MediaRange] = self.consumes
     }
 
-  def handleError(f: DecodeFailure => T)(implicit F: Functor[F]): EntityDecoder[F, T] =
-    new EntityDecoder[F, T] {
-      override def consumes: Set[MediaRange] = self.consumes
-
-      override def decode(msg: Message[F], strict: Boolean): DecodeResult[F, T] =
-        self.decode(msg, strict).recover(PartialFunction(f))
-    }
+  def handleError(f: DecodeFailure => T)(implicit F: Functor[F]): EntityDecoder[F, T] = transform {
+    case r @ Right(_) => r
+    case Left(e) => Right(f(e))
+  }
 
   def handleErrorWith(f: DecodeFailure => DecodeResult[F, T])(
-      implicit F: Monad[F]): EntityDecoder[F, T] = new EntityDecoder[F, T] {
-    override def consumes: Set[MediaRange] = self.consumes
-
-    override def decode(msg: Message[F], strict: Boolean): DecodeResult[F, T] =
-      self.decode(msg, strict).leftFlatMap(f)
+      implicit F: Monad[F]): EntityDecoder[F, T] = transformWith {
+    case Right(r) => DecodeResult.success(r)
+    case Left(e) => f(e)
   }
 
   def transform[T2](s: T => T2, f: DecodeFailure => DecodeFailure)(
--- a/core/src/main/scala/org/http4s/EntityDecoder.scala
+++ b/core/src/main/scala/org/http4s/EntityDecoder.scala
@@ -46,17 +46,17 @@ trait EntityDecoder[F[_], T] { self =>
     }
 
   def handleError(f: DecodeFailure => T)(implicit F: Functor[F]): EntityDecoder[F, T] = transform {
-    case r @ Right(_) => r
     case Left(e) => Right(f(e))
+    case r @ Right(_) => r
   }
 
   def handleErrorWith(f: DecodeFailure => DecodeResult[F, T])(
       implicit F: Monad[F]): EntityDecoder[F, T] = transformWith {
-    case Right(r) => DecodeResult.success(r)
     case Left(e) => f(e)
+    case Right(r) => DecodeResult.success(r)
   }
 
-  def transform[T2](s: T => T2, f: DecodeFailure => DecodeFailure)(
+  def bimap[T2](f: DecodeFailure => DecodeFailure, s: T => T2)(
       implicit F: Functor[F]): EntityDecoder[F, T2] =
     transform {
       case Left(e) => Left(f(e))
@@ -72,7 +72,7 @@ trait EntityDecoder[F[_], T] { self =>
         self.decode(msg, strict).transform(t)
     }
 
-  def transformWith[T2](s: T => DecodeResult[F, T2], f: DecodeFailure => DecodeResult[F, T2])(
+  def biflatMap[T2](f: DecodeFailure => DecodeResult[F, T2], s: T => DecodeResult[F, T2])(
       implicit F: Monad[F]): EntityDecoder[F, T2] =
     transformWith {
       case Left(e) => f(e)
--- a/core/src/main/scala/org/http4s/EntityDecoder.scala
+++ b/core/src/main/scala/org/http4s/EntityDecoder.scala
@@ -45,6 +45,56 @@ trait EntityDecoder[F[_], T] { self =>
       override def consumes: Set[MediaRange] = self.consumes
     }
 
+  def recover(f: DecodeFailure => T)(implicit F: Functor[F]): EntityDecoder[F, T] =
+    new EntityDecoder[F, T] {
+      override def consumes: Set[MediaRange] = self.consumes
+
+      override def decode(msg: Message[F], strict: Boolean): DecodeResult[F, T] =
+        self.decode(msg, strict).recover(PartialFunction(f))
+    }
+
+  def recoverWith(f: DecodeFailure => DecodeResult[F, T])(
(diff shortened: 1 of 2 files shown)


lloydmeta:
FYI, it looks like there are a couple of sources for non-deterministic test failures:

* scalafmt. I've not touched [this file](https://travis-ci.org/http4s/http4s/jobs/310844100#L4383-L4405) and [in another run, there were no problems](https://travis-ci.org/http4s/http4s/jobs/310831174#L4341-L4389).
* `ClientTimeoutSpec`: [Here it is failing on AppVeyor](https://ci.appveyor.com/project/ChristopherDavenport/http4s-m0hqu/build/1.0.508/job/5lfymw6diishmm53#L4322), here it is [passing on AppVeyor](https://ci.appveyor.com/project/ChristopherDavenport/http4s-m0hqu/build/1.0.510/job/9dchqdww755u4tr8#L4500)

core/src/main/scala/org/http4s/EntityDecoder.scala
```diff
@@ -45,6 +45,56 @@ trait EntityDecoder[F[_], T] { self =>
       override def consumes: Set[MediaRange] = self.consumes
     }
 
+  def recover(f: DecodeFailure => T)(implicit F: Functor[F]): EntityDecoder[F, T] =
```
rossabaker: In cats, [recover](https://github.com/typelevel/cats/blob/c51fe610704aaa482c74b37b5ca2ec396749f5f3/core/src/main/scala/cats/ApplicativeError.scala#L64) takes a `PartialFunction` and [handleError](https://github.com/typelevel/cats/blob/c51fe610704aaa482c74b37b5ca2ec396749f5f3/core/src/main/scala/cats/ApplicativeError.scala#L37) takes a `Function1`.  It would be nice to remain consistent with that terminology, even if we don't have an `ApplicativeError` instance.

(And on that subject, could we? 🤔)

lloydmeta: Good point, I'll rename it to `handleError` if that's ok with you. Reason being, I prefer `Function1` over `PartialFunction`.

FYI, I tried constraining it to `ApplicativeError[F, DecodeFailure]` 

```scala
  def handleError(f: DecodeFailure => T)(
      implicit Ae: ApplicativeError[F, DecodeFailure]): EntityDecoder[F, T] =
    new EntityDecoder[F, T] {
      override def consumes: Set[MediaRange] = self.consumes

      override def decode(msg: Message[F], strict: Boolean): DecodeResult[F, T] =
        DecodeResult(Ae.handleError(self.decode(msg, strict).value)(e => Right(f(e))))
    }
```

and `core` compiled, but then realised that there is no `ApplicativeError[IO, DecodeFailure]` when I tried to compile tests. I ran into ambiguous implicits when I tried to add one via:

```scala
trait EntityDecoderInstances {
   implicit def applicativeError[F[_]](implicit F: Effect[F]): ApplicativeError[F, DecodeFailure] = new ApplicativeError[F, DecodeFailure] {
    override def raiseError[A](e: DecodeFailure): F[A] = F.raiseError(e)

    override def handleErrorWith[A](fa: F[A])(f: DecodeFailure => F[A]): F[A] = F.recoverWith(fa) {
      case decodeFailure: DecodeFailure => f(decodeFailure)
      case other => F.raiseError(other)
    }

    override def pure[A](x: A): F[A] = F.pure(x)

    override def ap[A, B](ff: F[A => B])(fa: F[A]): F[B] = F.ap(ff)(fa)
  }
}
```

Not sure how to approach it it we want to go down this route.

More info on that error:

```
[error]  both method applicativeError in trait lower of type [F[_]](implicit F: cats.effect.Effect[F])cats.ApplicativeError[F,org.http4s.DecodeFailure]
[error]  and value evidence$3 of type cats.effect.Effect[F]
[error]  match expected type cats.Functor[F]
[error]     DecodeResult.success(msg.body.chunks.runFoldMonoid)
[error]                         ^
[error] 
```

core/src/main/scala/org/http4s/EntityDecoder.scala
```diff
@@ -45,6 +45,56 @@ trait EntityDecoder[F[_], T] { self =>
       override def consumes: Set[MediaRange] = self.consumes
     }
 
+  def handleError(f: DecodeFailure => T)(implicit F: Functor[F]): EntityDecoder[F, T] =
```
jmcardon: The following two methods seem strictly less powerful than `def transform[T2](t: Either[DecodeFailure, T] => Either[DecodeFailure, T2]` and `transformWith` in that you could implement it in terms of `transform` by using `identity` over the `Right` case.

core/src/main/scala/org/http4s/EntityDecoder.scala
```diff
@@ -45,6 +45,56 @@ trait EntityDecoder[F[_], T] { self =>
...
+
+  def handleErrorWith(f: DecodeFailure => DecodeResult[F, T])(
+      implicit F: Monad[F]): EntityDecoder[F, T] = new EntityDecoder[F, T] {
+    override def consumes: Set[MediaRange] = self.consumes
+
+    override def decode(msg: Message[F], strict: Boolean): DecodeResult[F, T] =
+      self.decode(msg, strict).leftFlatMap(f)
+  }
+
+  def transform[T2](s: T => T2, f: DecodeFailure => DecodeFailure)(
```
jmcardon: Why the `DecodeFailure => DecodeFailure`? wouldn't you want this to be `DecodeFailure => T` or `Either[E, T]` ?

lloydmeta: This override is so you can have strict transformation of success (`T => T2`) and error cases (`DecodeFailure => DecodeFailure`) separately. 

Mainly inspired by `Future#transform` in the standard lib.

core/src/main/scala/org/http4s/EntityDecoder.scala
lloydmeta: Excellent point. Reimplemented handleError and handleErrorWith using the transform methods.

core/src/main/scala/org/http4s/EntityDecoder.scala
jmcardon: hmm ok @lloydmeta, it just seems a bit inconsistent wrt [EitherT](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/data/EitherT.scala#L86), but this is only my opinion. If you feel this use case is quite common im not specifically against it.

:+1: thank you for the PR btw!

lloydmeta: @jmcardon I completely here you :) My hope is that the [other override for transform](https://github.com/http4s/http4s/pull/1575/files#diff-02708fea78424f8a8fa6eb4324a9daacR66) will make it feel more like that transform on `EitherT`.

jmcardon: I saw that one as well, I just meant the having two methods part of the `transform`.

Either way, this is cool with me :+1:

core/src/main/scala/org/http4s/EntityDecoder.scala
```diff
@@ -45,6 +45,51 @@ trait EntityDecoder[F[_], T] { self =>
...
+    case Left(e) => Right(f(e))
+  }
+
+  def handleErrorWith(f: DecodeFailure => DecodeResult[F, T])(
+      implicit F: Monad[F]): EntityDecoder[F, T] = transformWith {
+    case Right(r) => DecodeResult.success(r)
+    case Left(e) => f(e)
+  }
+
+  def transform[T2](s: T => T2, f: DecodeFailure => DecodeFailure)(
```
rossabaker: Would it be appropriate to call this `bimap` and avoid the overload?  Perhaps that implies a `Bifunctor`, which it's not.  I wish this had a different name from the other `transform`, which seems to fit nicely with what cats uses in `OptionT` and `EitherT`. :thinking:

In any event, I think it makes sense to flip `s` and `t`.  That's how `bimap` works, and that's the closest cousin I can think of for this method in the ecosystem.  It's not a big deal because it doesn't compile if the arguments are flipped, but I think it's nice for consistency.

lloydmeta: I have to laugh because I thought of the exact same thing today :) Renamed those overloads to `bimap` and `biflatMap` respectively and flipped `s` and `f` :)

PR #1576, opened 2017-12-03
Dropped mention of thrown exception from docstring

amrhassan:
It was incorrect.


rossabaker:
I don't know why I had to merge that twice.

PR #1577, opened 2017-12-03
Release v0.16.6

rossabaker:
Sneaks in a Jetty patch upgrade.


rossabaker:
It would help if I would have bumped the version number.

PR #1579, opened 2017-12-03
Specialize byte chunks to use array-backed chunks.

jmcardon:
This gets rid of a lot of `toChunk` calls, which are actually `Chunk.vector(toVector)` in fs2, for Array-backed Chunks for byte chunk instances.


rossabaker:
I have questions here, which I'll elaborate on as soon as I get a chance.

aeons:
Hmm, after closer inspection, this would not work with a byte chunk that has an offset and a length, right?

https://github.com/functional-streams-for-scala/fs2/blob/series/0.10/core/shared/src/main/scala/fs2/Chunk.scala#L316-L326

jmcardon:
@aeons good point. I should use `toArray`, not `.values`.

jmcardon:
I'm going to close this due to the new fs2 milestone, and I'll PR changes there.

PR #1584, opened 2017-12-06
Upgrade fs2 to 0.10.0-M9

aeons:
Everything compiles, but tests needs updated jawn-fs2 and fs2-reactive-streams.

- [x] [jawn-fs2](https://github.com/http4s/jawn-fs2/pull/13)
- [x] [fs2-reactive-streams](https://github.com/zainab-ali/fs2-reactive-streams/pull/40)


lukestephenson:
This looks good to me.  I agree with your comment on gitter:
> there’s a lot of Chunk[Byte] around that are from before segment was introduced
since we rarely do index-based access, I think those should be Segment[Byte, Unit] instead

rossabaker:
jawn-fs2-0.12.0-M4 is released.

blaze-core/src/main/scala/org/http4s/blazecore/Http1Stage.scala
```diff
@@ -75,7 +75,7 @@ trait Http1Stage[F[_]] { self: TailStage[ByteBuffer] =>
       rr: StringWriter,
       minor: Int,
       closeOnFinish: Boolean): Http1Writer[F] = lengthHeader match {
-    case Some(h) if bodyEncoding.map(!_.hasChunked).getOrElse(true) || minor == 0 =>
+    case Some(h) if bodyEncoding.forall(!_.hasChunked) || minor == 0 =>
```
rossabaker: Good use of `forall`!

blaze-core/src/main/scala/org/http4s/blazecore/util/IdentityWriter.scala
```diff
@@ -12,18 +12,18 @@ import org.http4s.util.chunk._
 import org.log4s.getLogger
 import scala.concurrent.{ExecutionContext, Future}
 
-private[http4s] class IdentityWriter[F[_]](size: Long, out: TailStage[ByteBuffer])(
+private[http4s] class IdentityWriter[F[_]](size: Int, out: TailStage[ByteBuffer])(
```
rossabaker: What forced this back to `Int`?

core/src/main/scala/org/http4s/util/Renderable.scala
```diff
@@ -197,7 +197,8 @@ final case class ChunkWriter(
 ) extends Writer {
 
   override def append(s: String): this.type = {
-    toChunk = (toChunk ++ Chunk.bytes(s.getBytes(charset))).toChunk
+    // This smells wrong
```
rossabaker: Is it better for this to be a `SegmentWriter`?

blaze-core/src/main/scala/org/http4s/blazecore/Http1Stage.scala
aeons: That was an autofix suggestion by IntelliJ :)

blaze-core/src/main/scala/org/http4s/blazecore/util/IdentityWriter.scala
aeons: `take` on `Chunk` is Int, while on `Segment` it's Long. I think before it it used the inherited take from Segment.

core/src/main/scala/org/http4s/util/Renderable.scala
aeons: Yes, I think so, but I think that would be part of a larger overhaul to use Segment generally.

PR #1588, opened 2017-12-08
(WIP) Replace `Chunk` For `Segment` where appropriate

jmcardon:
This is to improve on the previous proliferation of chunks everywhere, especially after the fs2 M9 changes.

### DIFF
--- a/blaze-core/src/test/scala/org/http4s/blazecore/util/DumpingWriter.scala
+++ b/blaze-core/src/test/scala/org/http4s/blazecore/util/DumpingWriter.scala
@@ -2,13 +2,9 @@ package org.http4s
 package blazecore
 package util
 
-import cats._
 import cats.effect.{Effect, IO}
-import cats.implicits._
 import fs2._
 import org.http4s.blaze.util.Execution
-import org.http4s.util.chunk._
-import scala.collection.mutable.ListBuffer
 import scala.concurrent.{ExecutionContext, Future}
 
 object DumpingWriter {
@@ -22,19 +18,19 @@ object DumpingWriter {
 class DumpingWriter(implicit protected val F: Effect[IO]) extends EntityBodyWriter[IO] {
   override implicit protected def ec: ExecutionContext = Execution.trampoline
 
-  private val buffers = new ListBuffer[Chunk[Byte]]
+  private var buffer = Segment.empty[Byte]
 
-  def toArray: Array[Byte] = buffers.synchronized {
-    Foldable[List].fold(buffers.toList).toBytes.values
+  def toArray: Array[Byte] = buffer.synchronized {
+    buffer.force.toArray
   }
 
-  override protected def writeEnd(chunk: Chunk[Byte]): Future[Boolean] = buffers.synchronized {
-    buffers += chunk
+  override protected def writeEnd(chunk: Chunk[Byte]): Future[Boolean] = buffer.synchronized {
+    buffer = buffer ++ Segment.chunk(chunk)
     Future.successful(false)
   }
 
   override protected def writeBodyChunk(chunk: Chunk[Byte], flush: Boolean): Future[Unit] = {
-    buffers += chunk
+    buffer = buffer ++ Segment.chunk(chunk)
     FutureUnit
   }
 }
--- a/blaze-core/src/test/scala/org/http4s/blazecore/util/DumpingWriter.scala
+++ b/blaze-core/src/test/scala/org/http4s/blazecore/util/DumpingWriter.scala
@@ -29,8 +29,9 @@ class DumpingWriter(implicit protected val F: Effect[IO]) extends EntityBodyWrit
     Future.successful(false)
   }
 
-  override protected def writeBodyChunk(chunk: Chunk[Byte], flush: Boolean): Future[Unit] = buffer.synchronized{
-    buffer = buffer ++ Segment.chunk(chunk)
-    FutureUnit
-  }
+  override protected def writeBodyChunk(chunk: Chunk[Byte], flush: Boolean): Future[Unit] =
+    buffer.synchronized {
+      buffer = buffer ++ Segment.chunk(chunk)
+      FutureUnit
+    }
 }
--- a/blaze-core/src/test/scala/org/http4s/blazecore/util/DumpingWriter.scala
+++ b/blaze-core/src/test/scala/org/http4s/blazecore/util/DumpingWriter.scala
@@ -29,7 +29,7 @@ class DumpingWriter(implicit protected val F: Effect[IO]) extends EntityBodyWrit
     Future.successful(false)
   }
 
-  override protected def writeBodyChunk(chunk: Chunk[Byte], flush: Boolean): Future[Unit] = {
+  override protected def writeBodyChunk(chunk: Chunk[Byte], flush: Boolean): Future[Unit] = buffer.synchronized{
     buffer = buffer ++ Segment.chunk(chunk)
     FutureUnit
   }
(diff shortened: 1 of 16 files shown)


circe/src/main/scala/org/http4s/circe/CirceInstances.scala
```diff
@@ -22,7 +22,7 @@ trait CirceInstances {
 
   private def jsonDecoderByteBufferImpl[F[_]: Effect](msg: Message[F]): DecodeResult[F, Json] =
     EntityDecoder.collectBinary(msg).flatMap { chunk =>
-      val bb = ByteBuffer.wrap(chunk.toBytes.values)
+      val bb = ByteBuffer.wrap(chunk.force.toArray)
```
jmcardon: the variable should be called segment

core/src/main/scala/org/http4s/EntityDecoder.scala
```diff
@@ -198,17 +198,20 @@ trait EntityDecoderInstances {
       override def consumes: Set[MediaRange] = Set.empty
     }
 
-  implicit def binary[F[_]: Effect]: EntityDecoder[F, Chunk[Byte]] =
+  implicit def binary[F[_]: Effect]: EntityDecoder[F, Segment[Byte, Unit]] =
     EntityDecoder.decodeBy(MediaRange.`*/*`)(collectBinary[F])
 
+  implicit def binaryChunk[F[_]: Effect]: EntityDecoder[F, Chunk[Byte]] =
+    EntityDecoder.decodeBy(MediaRange.`*/*`)(collectBinary[F]).map(_.force.toChunk)
```
jmcardon: this should just map on the previous entity decoder.

core/src/main/scala/org/http4s/EntityEncoder.scala
```diff
@@ -70,7 +70,7 @@ object EntityEncoder extends EntityEncoderInstances {
     *
     * This constructor is a helper for types that can be serialized synchronously, for example a String.
     */
-  def simple[F[_], A](hs: Header*)(toChunk: A => Chunk[Byte])(
+  def simple[F[_], A](hs: Header*)(toChunk: A => Chunk[Byte])( //Fishy but I'm wondering wat to do
```
jmcardon: Unsure about what to do re: A => Chunk.

tests/src/test/scala/org/http4s/EntityCodecSpec.scala
```diff
@@ -16,10 +16,13 @@ class EntityCodecSpec extends Http4sSpec {
...
     Eq.by(_.toVector)
 
+  implicit def eqSegment[A](implicit ev: Eq[Vector[A]]): Eq[Segment[A, Unit]] =
+    Eq.by(_.force.toVector)
+
   checkAll("EntityCodec[IO, String]", EntityCodecTests[IO, String].entityCodec)
   checkAll("EntityCodec[IO, Array[Char]]", EntityCodecTests[IO, Array[Char]].entityCodec)
 
-  checkAll("EntityCodec[IO, Chunk[Byte]]", EntityCodecTests[IO, Chunk[Byte]].entityCodec)
+  checkAll("EntityCodec[IO, Chunk[Byte]]", EntityCodecTests[IO, Segment[Byte, Unit]].entityCodec)
```
jmcardon: Should be Segment

blaze-core/src/test/scala/org/http4s/blazecore/util/DumpingWriter.scala
```diff
@@ -22,19 +18,19 @@ object DumpingWriter {
...
-  override protected def writeEnd(chunk: Chunk[Byte]): Future[Boolean] = buffers.synchronized {
-    buffers += chunk
+  override protected def writeEnd(chunk: Chunk[Byte]): Future[Boolean] = buffer.synchronized {
+    buffer = buffer ++ Segment.chunk(chunk)
     Future.successful(false)
   }
 
   override protected def writeBodyChunk(chunk: Chunk[Byte], flush: Boolean): Future[Unit] = {
-    buffers += chunk
+    buffer = buffer ++ Segment.chunk(chunk)
```
rossabaker: This should be synchronized if the others are.

circe/src/main/scala/org/http4s/circe/CirceInstances.scala
rossabaker: :+1:

core/src/main/scala/org/http4s/EntityDecoder.scala
rossabaker: Does this even need to exist?

core/src/main/scala/org/http4s/EntityEncoder.scala
rossabaker: What is a use case for creating a segment here instead of a chunk?  This encoder is strict.

core/src/main/scala/org/http4s/EntityEncoder.scala
```diff
@@ -148,6 +148,9 @@ trait EntityEncoderInstances extends EntityEncoderInstances0 {
       charset: Charset = DefaultCharset): EntityEncoder[F, Array[Char]] =
     stringEncoder[F].contramap(new String(_))
 
+  implicit def segmentEncoder[F[_]: Applicative]: EntityEncoder[F, Segment[Byte, Unit]] =
+    chunkEncoder[F].contramap[Segment[Byte, Unit]](_.force.toChunk) //TODO: Seems fishy
```
rossabaker: Maybe this is what sucks because of the definition of `simple`.

I don't yet have a strong opinion on whether we need an encoder for both `Segment` and `Chunk`.

core/src/main/scala/org/http4s/util/Renderable.scala
```diff
@@ -188,17 +188,17 @@ object StringWriter {
...
 final case class ChunkWriter(
-    var toChunk: Chunk[Byte] = Chunk.empty,
+    var toByteSegment: Segment[Byte, Unit] = Segment.empty[Byte],
     charset: Charset = StandardCharsets.UTF_8
 ) extends Writer {
 
   override def append(s: String): this.type = {
     // This smells wrong
-    toChunk = (toChunk.toSegment ++ Segment.array(s.getBytes(charset))).force.toChunk
+    toByteSegment = toByteSegment ++ Segment.array(s.getBytes(charset))
```
rossabaker: Does this still smell wrong?

core/src/main/scala/org/http4s/util/chunk.scala
```diff
@@ -6,13 +6,13 @@ import java.nio.ByteBuffer
 
 trait ChunkInstances {
 
-  implicit def http4sMonoidForChunk[A]: Monoid[Chunk[A]] =
-    new Monoid[Chunk[A]] {
-      // This smells
-      override def combine(x: Chunk[A], y: Chunk[A]): Chunk[A] =
-        (x.toSegment ++ y.toSegment).force.toChunk
+  //Maybe this belongs in fs2?
+  implicit def http4sMonoidForChunk[A]: Monoid[Segment[A, Unit]] =
```
rossabaker: The name is now wrong.

If this is still used, it belongs in fs2.  It's not our types.

jmcardon: I think I can PR this onto fs2.

core/src/main/scala/org/http4s/EntityEncoder.scala
jmcardon: hmm, the only reason I was wondering if there was anything I could do, is that toChunk currently makes me do a `segment.force` further down.

core/src/main/scala/org/http4s/util/Renderable.scala
jmcardon: should be fine now.

core/src/main/scala/org/http4s/util/chunk.scala
jmcardon: Also yeah, it's still used, we use `runFoldMonoid` a lot

core/src/main/scala/org/http4s/EntityEncoder.scala
jmcardon: If we define both a decoder and an encoder, we get an extra test for free in the entitycodec spec is all, so I figured this would be okay.

Issue #1578, opened 2017-12-03
Remove try/catch and tighten bound on EntityDecoder


jmcardon:
As discussed on gitter (history link coming soon), we could probably get rid of the `try/catch` stuff happening [Here](https://github.com/http4s/http4s/blob/master/core/src/main/scala/org/http4s/EntityDecoder.scala#L85), after #1572 is merged for a more principled `EntityDecoder`, by tightening the bound of `F[_]` to at least `Sync`, or something akin to this, since:

1. Evaluating the body of the Message to decode (which is the case most of the time for `Json` or `XML` results) evaluates the stream, and will be necessary anyway.
2. Even in the case that it comes from something like a header, to even instantiate a `ServerBuilder` like Blaze or Jetty, you require an `Effect` bound anyway.

jmcardon:
Closed by #1592

PR #1592, opened 2017-12-11
Remove try/catch semantics in otherwise pure code

jmcardon:
Handling #1578 and cleaning up code in general.

### DIFF
--- a/client/src/main/scala/org/http4s/client/Client.scala
+++ b/client/src/main/scala/org/http4s/client/Client.scala
@@ -33,8 +33,7 @@ final case class DisposableResponse[F[_]](response: Response[F], dispose: F[Unit
     val task: F[A] = try f(response)
     catch {
       case e: Throwable =>
-        logger.error(e)(
-          """Handled exception in client callback to prevent a connection leak.
+        logger.error(e)("""Handled exception in client callback to prevent a connection leak.
              |The callback should always return an F. If your callback can fail
              |with an exception you can't handle, call `F.raiseError(exception)`.
           """.stripMargin)
--- a/client/src/main/scala/org/http4s/client/Client.scala
+++ b/client/src/main/scala/org/http4s/client/Client.scala
@@ -35,8 +35,8 @@ final case class DisposableResponse[F[_]](response: Response[F], dispose: F[Unit
       case e: Throwable =>
         logger.error(e)(
           """Handled exception in client callback to prevent a connection leak.
-             |The callback should always return anF. If your callback can fail with an exception you can't handle,
-             |callF.raiseError(exception)`.
+             |The callback should always return an F. If your callback can fail
+             |with an exception you can't handle, call `F.raiseError(exception)`.
           """.stripMargin)
         F.raiseError(e)
     }
--- a/client/src/main/scala/org/http4s/client/Client.scala
+++ b/client/src/main/scala/org/http4s/client/Client.scala
@@ -26,8 +26,7 @@ final case class DisposableResponse[F[_]](response: Response[F], dispose: F[Unit
     * HTTP connection when the task finishes.
     */
   def apply[A](f: Response[F] => F[A])(implicit F: MonadError[F, Throwable]): F[A] = {
-    val task = try f(response)
-    catch { case e: Throwable => F.raiseError[A](e) }
+    val task = f(response)
     for {
       result <- task.attempt
       _ <- dispose
--- a/client/src/main/scala/org/http4s/client/Client.scala
+++ b/client/src/main/scala/org/http4s/client/Client.scala
@@ -33,7 +33,7 @@ final case class DisposableResponse[F[_]](response: Response[F], dispose: F[Unit
     val task: F[A] = try f(response)
     catch {
       case e: Throwable =>
-        logger.info(s"Caught exception in otherwise pure expression $e")
+        logger.error(s"Caught exception in otherwise pure expression $e")
         F.raiseError(e)
     }
 
--- a/client/src/main/scala/org/http4s/client/Client.scala
+++ b/client/src/main/scala/org/http4s/client/Client.scala
@@ -12,6 +12,7 @@ import org.http4s.Status.Successful
 import org.http4s.headers.{Accept, MediaRangeAndQValue}
 import scala.concurrent.SyncVar
 import scala.util.control.NoStackTrace
+import org.log4s.getLogger
 
 /**
   * Contains a [[Response]] that needs to be disposed of to free the underlying
@@ -21,12 +22,21 @@ import scala.util.control.NoStackTrace
   */
 final case class DisposableResponse[F[_]](response: Response[F], dispose: F[Unit]) {
 
+  private[this] val logger = getLogger
+
   /**
     * Returns a task to handle the response, safely disposing of the underlying
     * HTTP connection when the task finishes.
     */
   def apply[A](f: Response[F] => F[A])(implicit F: MonadError[F, Throwable]): F[A] = {
-    val task = f(response)
+    //Catch possible user bugs in pure expression
+    val task: F[A] = try f(response)
+    catch {
+      case e: Throwable =>
+        logger.info(s"Caught exception in otherwise pure expression $e")
+        F.raiseError(e)
+    }
+
     for {
       result <- task.attempt
       _ <- dispose
--- a/client/src/main/scala/org/http4s/client/Client.scala
+++ b/client/src/main/scala/org/http4s/client/Client.scala
@@ -33,7 +33,11 @@ final case class DisposableResponse[F[_]](response: Response[F], dispose: F[Unit
     val task: F[A] = try f(response)
     catch {
       case e: Throwable =>
-        logger.error(s"Caught exception in otherwise pure expression $e")
+        logger.error(e)(
+          """Handled exception in client callback to prevent a connection leak.
+             |The callback should always return anF. If your callback can fail with an exception you can't handle,
+             |callF.raiseError(exception)`.
+          """.stripMargin)
         F.raiseError(e)
     }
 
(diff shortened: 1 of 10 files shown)


client/src/main/scala/org/http4s/client/Client.scala
```diff
@@ -26,8 +26,7 @@ final case class DisposableResponse[F[_]](response: Response[F], dispose: F[Unit
     * HTTP connection when the task finishes.
     */
   def apply[A](f: Response[F] => F[A])(implicit F: MonadError[F, Throwable]): F[A] = {
-    val task = try f(response)
-    catch { case e: Throwable => F.raiseError[A](e) }
+    val task = f(response)
```
jmcardon: This is specifically the change I want to discuss.

I'm unsure whether catching this is.. good. It's a sort of cushion but I do not think this should be necessary if the result is parametrized to begin with.

rossabaker: An exception here will leak a connection, which is a severe response to a newbie mistake.  A try that is uncaught is negligible, to possibly zero, performance overhead.  (I've been told both and not measured myself.)  I think this might be a good case to lecture in the logs but try to keep the system operational.

core/src/main/scala/org/http4s/EntityDecoder.scala
```diff
@@ -122,25 +121,21 @@ object EntityDecoder extends EntityDecoderInstances {
   /** Create a new [[EntityDecoder]]
     *
     * The new [[EntityDecoder]] will attempt to decode messages of type `T`
-    * only if the [[Message]] satisfies the provided [[MediaRange]]s
+    * only if the [[Message]] satisfies the provided [[MediaRange]]
+    * note: Sync bound to note that such an `F[_]` may throw
     */
-  def decodeBy[F[_]: Applicative, T](r1: MediaRange, rs: MediaRange*)(
+  def decodeBy[F[_]: Sync, T](r1: MediaRange, rs: MediaRange*)(
```
rossabaker: The `DecodeResult` still lets us `raiseError`, no?

core/src/main/scala/org/http4s/util/package.scala
```diff
@@ -63,6 +63,7 @@ package object util {
     new AssertionError(
       s"This is a bug. Please report to https://github.com/http4s/http4s/issues: ${message}")
 
+  // Todo: Possibly deprecate? Cats already has this.
```
rossabaker: It's private and totally redundant.  I'd straight up remove this one.

core/src/main/scala/org/http4s/MessageFailure.scala
```diff
@@ -56,6 +56,7 @@ object ParseResult {
   def success[A](a: A): ParseResult[A] =
     Either.right(a)
 
+  //Todo: Possibly deprecate???
```
rossabaker: I think it's a legit construction from unsafe code.  It's not necessary, but I guess preserves the aliasing of `Either` to a `ParseResult`.  I'm inclined to keep this one.

server/src/main/scala/org/http4s/server/middleware/PushSupport.scala
```diff
@@ -52,30 +49,27 @@ object PushSupport {
...
+                      accumulated ++ _ :+ PushResponse(v.location, response))
+                  }
+                  .getOrElse(F.pure(accumulated :+ PushResponse(v.location, response)))
               }
-              .apply(newReq)
-              .getOrElse(Vector.empty[PushResponse[F]])
-          } catch { case t: Throwable => handleException(t); facc }
+            }
+            .apply(newReq)
+            .getOrElse(Vector.empty[PushResponse[F]])
```
rossabaker: I'm trying to think what exception here we were motivated to catch in the first place.

servlet/src/main/scala/org/http4s/servlet/Http4sServlet.scala
```diff
@@ -96,18 +96,11 @@ class Http4sServlet[F[_]](
       request: Request[F],
       bodyWriter: BodyWriter[F]): F[Unit] = {
     ctx.addListener(new AsyncTimeoutHandler(request, bodyWriter))
-    // TODO: This can probably be handled nicer with Effect
+    // Note: We're catching silly user errors in the lift => flatten.
```
rossabaker: This is a case where I'd be happy to lecture about the silliness.

There is a similar construction in the blaze Http1ServerStage.

client/src/main/scala/org/http4s/client/Client.scala
jmcardon: Fair.. I just wanted to discuss it and leave a note there in the case we do not remove it.

core/src/main/scala/org/http4s/EntityDecoder.scala
jmcardon: I guess but `raiseError` on `EitherT` doesn't result in the same as `F.raiseError`, I think.

server/src/main/scala/org/http4s/server/middleware/PushSupport.scala
jmcardon: I Don't think this one is doing anything in particular.. its removal didn't break anything.

core/src/main/scala/org/http4s/EntityDecoder.scala
jmcardon: Should we not tighten this to a sync bound? I'm just not to keen on just `Applicative` possibly denoting something impure (which people can do in the `f`

client/src/main/scala/org/http4s/client/Client.scala
```diff
@@ -21,13 +22,21 @@ import scala.util.control.NoStackTrace
...
     * HTTP connection when the task finishes.
     */
   def apply[A](f: Response[F] => F[A])(implicit F: MonadError[F, Throwable]): F[A] = {
-    val task = try f(response)
-    catch { case e: Throwable => F.raiseError[A](e) }
+    //Catch possible user bugs in pure expression
+    val task: F[A] = try f(response)
+    catch {
+      case e: Throwable =>
+        logger.info(s"Caught exception in otherwise pure expression $e")
```
ChristopherDavenport: I think we can lecture stronger than this.  `logger.error` perhaps.

servlet/src/main/scala/org/http4s/servlet/Http4sServlet.scala
jmcardon: the one in the `Http1ServerStage` is of the form:
```scala
  executionContext.execute(new Runnable {
          def run(): Unit =
            F.runAsync {
                try serviceFn(req)
                  .getOrElse(Response.notFound)
                  .handleErrorWith(serviceErrorHandler(req))
                catch serviceErrorHandler(req)
              } {
                case Right(resp) =>
                  IO(renderResponse(req, resp, cleanup))
                case Left(t) =>
                  IO(internalServerError(s"Error running route: $req", t, req, cleanup))
              }
              .unsafeRunSync()
        })
```

This is similar to writing `fs2.asyncUnsafeRunAsync(F)(callback)` but with the exception that `F.shift(ec)` submits a task onto the execution context.

Maybe im being a paranoid panda but while the ServerStage1 is uglier, it might be more efficient (In a sense, it's not submitting a useless task to evaluate the continuation) and I'm afraid of a tiny performance regression because of a change there.

core/src/main/scala/org/http4s/EntityDecoder.scala
rossabaker: A failure to decode should be a `DecodeResult.failure`, and people can write an `f` that throws regardless of the type bound.  How is that stronger constraint going to be used?

server/src/main/scala/org/http4s/server/middleware/PushSupport.scala
rossabaker: Kill it.

servlet/src/main/scala/org/http4s/servlet/Http4sServlet.scala
rossabaker: That's one of the hotspots in a ping test, which sadly is how people choose libraries.  (Shakes fist at naive reactions to Techempower.)  Okay, let's be conservative there, but the sanctimonious logging could be the same.

client/src/main/scala/org/http4s/client/Client.scala
```diff
@@ -33,7 +33,7 @@ final case class DisposableResponse[F[_]](response: Response[F], dispose: F[Unit
     val task: F[A] = try f(response)
     catch {
       case e: Throwable =>
-        logger.info(s"Caught exception in otherwise pure expression $e")
+        logger.error(s"Caught exception in otherwise pure expression $e")
```
rossabaker: We don't know it was otherwise pure.  And I'm not sure this tells a new FP programmer what they did wrong.

`logger.error(e)("Handled exception in client callback to prevent a connection leak. The callback should always return an `F`. If your callback can fail with an exception you can't handle, call `F.raiseError(exception)`.")

core/src/main/scala/org/http4s/EntityDecoder.scala
```diff
@@ -122,25 +121,21 @@ object EntityDecoder extends EntityDecoderInstances {
   /** Create a new [[EntityDecoder]]
     *
     * The new [[EntityDecoder]] will attempt to decode messages of type `T`
-    * only if the [[Message]] satisfies the provided [[MediaRange]]s
+    * only if the [[Message]] satisfies the provided [[MediaRange]]
+    * note: Sync bound to note that such an `F[_]` may throw
```
aeons: Isn't the bound still `Applicative` or is it referring to something else?

rossabaker: Good catch.  I took a stab at correcting it.

rossabaker:
For expediency, gambling that Travis likes this one since it liked the last one except for scalafmt.

PR #1593, opened 2017-12-14
Tests for Rfc2616 parser

cquiroz:
This PR starts from a change I want to do removing `Registry` from `MediaRange`. To do so I needed to generate `MediaRange` instances for scalacheck, which led me to create generators for various parts of the http spec previously untested.

This PR contains a test and the generators but there is one that is contentious and I'd like to get your comments

### DIFF
--- a/testing/src/main/scala/org/http4s/testing/ArbitraryInstances.scala
+++ b/testing/src/main/scala/org/http4s/testing/ArbitraryInstances.scala
@@ -42,6 +42,36 @@ trait ArbitraryInstances {
       } yield NonEmptyList(a, list)
     }
 
+  val genChar: Gen[Char] = choose('\u0000', '\u007F')
+
+  val ctlChar: List[Char] = ('\u007F' +: ('\u0000' to '\u001F')).toList
+
+  val lws: List[Char] = " \t".toList
+
+  val genCrLf: Gen[String] = const("\r\n")
+
+  val genRightLws: Gen[String] = nonEmptyListOf(oneOf(lws)).map(_.mkString)
+
+  val genLws: Gen[String] =
+    oneOf(sequence[List[String], String](List(genCrLf, genRightLws)).map(_.mkString), genRightLws)
+
+  val octets: List[Char] = ('\u0000' to '\u00FF').toList
+
+  val genOctet: Gen[Char] = oneOf(octets)
+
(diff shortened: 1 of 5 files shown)


testing/src/main/scala/org/http4s/testing/ArbitraryInstances.scala
```diff
@@ -42,6 +42,36 @@ trait ArbitraryInstances {
...
+
+  val octets: List[Char] = ('\u0000' to '\u00FF').toList
+
+  val genOctet: Gen[Char] = oneOf(octets)
+
+  val allowedText: List[Char] = octets.diff(ctlChar)
+
+  val genText: Gen[String] = oneOf(nonEmptyListOf(oneOf(allowedText)).map(_.mkString), genLws)
+
+  val allowedQDText: List[Char] = allowedText.filterNot(c => c == '"' || c == '\\')
```
cquiroz: According to spec the backslash character `\` is allowed but it seems the parser cannot handle one particular case where the Generated QDText ends on `\"` like e.g.

`"_¡1S^¼m¬Á\"`
`"ìÙ\"`

This is certainly a border case and I couldn't figure how to make the parboiled parser handle it properly so I tweaked the generator. WDYT?

rossabaker: Certainly seems like we _should_ be able to fix the parser.  I don't want it to stand in the way of progress, but I'd like to leave a note so we try to fix it.

cquiroz:
I added a comment and a small fix for scala 2.11. It's probably a good idea to review the generators anyway

cquiroz:
Which standard? I could rename them

rossabaker:
[This is how cats names them](https://github.com/typelevel/cats/blob/master/laws/src/main/scala/cats/laws/discipline/Arbitrary.scala).  I've been using `http4s` instead of `catsPackageName` as a unique-enough prefix.

cquiroz:
I see, I will switch the convention. Looking at the arbitraries you can see there have been many hands involved

cquiroz:
Maybe it's better to do another PR with all arbitraries names fixed in on go. Shall I merge this one to get my another change as a PR?

rossabaker:
Yes. :)

PR #1595, opened 2017-12-15
Today I learned how to list plugin updates

rossabaker:


cquiroz:
How did you do it? I know how to find dependency updates but not for plugins

rossabaker:
[Like this](https://github.com/rtimush/sbt-updates#sbt-plugin-updates)

PR #1597, opened 2017-12-16
MediaRange immutable

cquiroz:
This PR removes the use of `Registry` on `MediaRange` and converts it up to current standards.


core/src/main/scala/org/http4s/MediaType.scala
```diff
@@ -71,27 +78,77 @@ sealed class MediaRange private[http4s] (
...
+  val `*/*` = new MediaRange("*")
+  val `application/*` = new MediaRange("application")
+  val `audio/*` = new MediaRange("audio")
+  val `image/*` = new MediaRange("image")
+  val `message/*` = new MediaRange("message")
+  val `multipart/*` = new MediaRange("multipart")
+  val `text/*` = new MediaRange("text")
+  val `video/*` = new MediaRange("video")
+
+  val standard: Map[String, MediaRange] =
```
rossabaker: I would have thought this includes media types without the second wild card, but I guess it didn't before?

cquiroz: This is basically replaces the registry with the same set of registered media types as before. Should be fairly mechanical

rossabaker: Hmm.  These aren't what I expected them to be.  I thought it was keyed by `mainType/subType` instead of just `mainType`.  I wonder why that is.  But you didn't change it.

cquiroz: Well the `MediaRange` case class only has a `mainType` unlike `MediaType`
https://github.com/http4s/http4s/blob/master/core/src/main/scala/org/http4s/MediaType.scala#L97

What surprised me is that the extensions are ignored to set the key

rossabaker: A MediaType is a MediaRange, so I'd expect to be able to look up MediaTypes via the MediaRange companion.

I'm not sure what I think about extensions.

cquiroz: Any ideas to move this forward?

cquiroz:
This is approved so I could merge it but if you still want some changes related to the key of the standard map please let me know

rossabaker:
Let's merge it now because it's forward progress and then discuss more progress.

Am I weird for thinking the standard media ranges should contain media types since media types are media ranges?

cquiroz:
I think it is a valid concern

PR #1598, opened 2017-12-17
Bump scala versions

rossabaker:


rossabaker:
I think I offended appveyor with my bad branch targeting.  Going to try again.

PR #1600, opened 2017-12-17
Prepare version guide for 0.18.0 and beyond

rossabaker:
- 0.16 and 0.17 are EOL.
- 0.18 gets bugfixes and discretionary backports.
- 0.19 has begun.


aeons:
Dunno if this one broke it, but names of versions before 0.16 are gone. It looks like an issue with the `latestInSeries` shortcode.

Issue #1596, opened 2017-12-15
Websocket convention is more intuitive from the point of the Receiver


jmcardon:
Currently websockets on 0.18 are

`read` -> Sends messages to the RealWorld™
`write` -> Receives messages from RealWorld™

This is pretty backwards. Most people writing websocket services think of it from the point of the server.

@ChristopherDavenport Had a good convention for this:

`read` -> `serverWrite` or maybe `serverEmit`
`write` -> `serverRead`

This is a simple rename, but should also be accompanied by it being noted in the scaladoc.

cquiroz:
This has been the convention for quite a while, since at least 0.14. However I agree that it is confusing. I like the idea of prefixing with `server`.

jmcardon:
@cquiroz maybe you disagree with me, but when you're writing a backend service, the most natural thing to do to when writing a websocket is thinking about it from the point of view of the service you're writing, not the sender to the service.

I think this is natural and to any newcomers (especially from play) it can be a point of confusion.

cquiroz:
I recall being confused too, but didn't want to change the convention when we ported it to fs2. Let's remember though that being a socket there is no preferred side to it. Prefixing seems good but I could be convinced of other options

rossabaker:
Can we use the same model when we add support for the client?  If we prefix with "server", that would impede a shared model.

The words that jump out to me in @jmcardon's explanation are `send` and `receive`.  How about those as names?

rsoeldner:
@rossabaker with `send` and `receive` the flow is (at least for me) much clearer, If you want I can PR this change.

rossabaker:
I would approve `send` and `receive`.  I don't know who else is with me.  😄

jmcardon:
:+1: to `send` and `receive`

rossabaker:
Alright.  @rsoeldner, PR would be appreciated. :)

cquiroz:
@rossabaker 👍 for a shared client/server model

PR #1603, opened 2017-12-18
Rename read/write to send/receive

rsoeldner:
.. as discussed #1596

### DIFF
--- a/core/src/main/scala/org/http4s/websocket/Websocket.scala
+++ b/core/src/main/scala/org/http4s/websocket/Websocket.scala
@@ -4,6 +4,6 @@ import fs2._
 import org.http4s.websocket.WebsocketBits.WebSocketFrame
 
 private[http4s] final case class Websocket[F[_]](
-    read: Stream[F, WebSocketFrame],
-    write: Sink[F, WebSocketFrame]
+    send: Stream[F, WebSocketFrame],
+    receive: Sink[F, WebSocketFrame]
 )
--- a/core/src/main/scala/org/http4s/websocket/Websocket.scala
+++ b/core/src/main/scala/org/http4s/websocket/Websocket.scala
@@ -4,8 +4,8 @@ import fs2._
 import org.http4s.websocket.WebsocketBits.WebSocketFrame
 
 private[http4s] final case class Websocket[F[_]](
-    @deprecatedName('read, "0.18-M6") send: Stream[F, WebSocketFrame],
-    @deprecatedName('write, "0.18-M6") receive: Sink[F, WebSocketFrame]
+    @deprecatedName('read, "0.18.0-M7") send: Stream[F, WebSocketFrame],
+    @deprecatedName('write, "0.18.0-M7") receive: Sink[F, WebSocketFrame]
 ) {
   def read: Stream[F, WebSocketFrame] = send
   def write: Sink[F, WebSocketFrame] = receive
--- a/core/src/main/scala/org/http4s/websocket/Websocket.scala
+++ b/core/src/main/scala/org/http4s/websocket/Websocket.scala
@@ -4,6 +4,9 @@ import fs2._
 import org.http4s.websocket.WebsocketBits.WebSocketFrame
 
 private[http4s] final case class Websocket[F[_]](
-    send: Stream[F, WebSocketFrame],
-    receive: Sink[F, WebSocketFrame]
-)
+    @deprecatedName('read) send: Stream[F, WebSocketFrame],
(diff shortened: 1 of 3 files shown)


server/src/main/scala/org/http4s/server/websocket/package.scala
```diff
@@ -40,13 +40,13 @@ package object websocket {
     * @param status The status code to return to a client making a non-websocket HTTP request to this route
     */
   def WS[F[_]](
-      read: Stream[F, WebSocketFrame],
-      write: Sink[F, WebSocketFrame],
+      send: Stream[F, WebSocketFrame],
```
rsoeldner: @aeons I'm unsure how to handle the deprecation here

server/src/main/scala/org/http4s/server/websocket/package.scala
```diff
@@ -17,8 +17,8 @@ package object websocket {
   /**
     * Build a response which will accept an HTTP websocket upgrade request and initiate a websocket connection using the
     * supplied exchange to process and respond to websocket messages.
-    * @param read     The read side of the Exchange represents the stream of messages that should be sent to the client
-    * @param write    The write side of the Exchange is a sink to which the framework will push the websocket messages
+    * @param send     The send side of the Exchange represents the stream of messages that should be sent to the client
+    * @param receive  The receive side of the Exchange is a sink to which the framework will push the websocket messages
```
cquiroz: Maybe the description of `receive` could be clearer on the line of `receiving messages from the client`

cquiroz:
Looks good. I guess existing users will get a compilation error right away forcing them to change the code and it's a simple rename in this case

core/src/main/scala/org/http4s/websocket/Websocket.scala
```diff
@@ -4,6 +4,9 @@ import fs2._
 import org.http4s.websocket.WebsocketBits.WebSocketFrame
 
 private[http4s] final case class Websocket[F[_]](
-    read: Stream[F, WebSocketFrame],
-    write: Sink[F, WebSocketFrame]
-)
+    @deprecatedName('read) send: Stream[F, WebSocketFrame],
```
ChristopherDavenport: Can the deprecation be versioned here?

server/src/main/scala/org/http4s/server/websocket/package.scala
rsoeldner: Maybe change `client` to `consumer` (on the send parameter) and `producer` on the `receive` parameter ?

server/src/main/scala/org/http4s/server/websocket/package.scala
rossabaker: I don't think we can in this case.  It'll break for people using named arguments, but at least the types are unique.  I think it'll be okay.

core/src/main/scala/org/http4s/websocket/Websocket.scala
```diff
@@ -4,6 +4,9 @@ import fs2._
 import org.http4s.websocket.WebsocketBits.WebSocketFrame
 
 private[http4s] final case class Websocket[F[_]](
-    read: Stream[F, WebSocketFrame],
-    write: Sink[F, WebSocketFrame]
-)
+    @deprecatedName('read, "0.18-M6") send: Stream[F, WebSocketFrame],
+    @deprecatedName('write, "0.18-M6") receive: Sink[F, WebSocketFrame]
```
rossabaker: Sorry, one last nitpick: the version that will include this is 0.18.0-M7.

rsoeldner: Whoop, right :+1:  Thank You

server/src/main/scala/org/http4s/server/websocket/package.scala
jmcardon: Something like `The incoming messages to this service` works as well. In a sense, it's even easier to think about incoming and outgoing messages.

You could do that for the client as well. So you can actually say "The send side of the exchange represents the outgoing messages sent from this service."

This mostly following from @rossabaker 's incentive to make this agnostic to where its done, so it works with the client.

server/src/main/scala/org/http4s/server/websocket/package.scala
jmcardon: :+1: to this as well. It should be a simple fix for users.

rsoeldner:
@rossabaker next try :tada:

PR #1607, opened 2017-12-20
Websocket support for handhake headers (such as:Sec-WebSocket-Protoco…

jodagm:
@rossabaker following your guidance at https://github.com/http4s/http4s-websocket/issues/7 
I started with a minimal approach in the spirit of the existing WS API (thinking about current usage of WS), so I just added another overload with the additional headers. I moved WebSocketContext to the core alongside Websocket. 

I could not make the compiler happy with the default responses as ```def```, I would need your help to appease the implicit Gods :)

### DIFF
--- a/blaze-server/src/main/scala/org/http4s/server/blaze/WebSocketSupport.scala
+++ b/blaze-server/src/main/scala/org/http4s/server/blaze/WebSocketSupport.scala
@@ -52,6 +52,11 @@ private[blaze] trait WebSocketSupport[F[_]] extends Http1ServerStage[F] {
             hdrs.foreach {
               case (k, v) => sb.append(k).append(": ").append(v).append('\r').append('\n')
             }
+
+            val wsContext = ws.get
+            wsContext.headers.foreach(hdr =>
+              sb.append(hdr.name).append(": ").append(hdr.value).append('\r').append('\n'))
+
             sb.append('\r').append('\n')
 
             // write the accept headers and reform the pipeline
@@ -59,7 +64,7 @@ private[blaze] trait WebSocketSupport[F[_]] extends Http1ServerStage[F] {
               case Success(_) =>
                 logger.debug("Switching pipeline segments for websocket")
 
-                val segment = LeafBuilder(new Http4sWSStage[F](ws.get))
+                val segment = LeafBuilder(new Http4sWSStage[F](wsContext.webSocket))
                   .prepend(new WSFrameAggregator)
                   .prepend(new WebSocketDecoder(false))
 
--- a/blaze-server/src/main/scala/org/http4s/server/blaze/WebSocketSupport.scala
+++ b/blaze-server/src/main/scala/org/http4s/server/blaze/WebSocketSupport.scala
@@ -25,56 +25,55 @@ private[blaze] trait WebSocketSupport[F[_]] extends Http1ServerStage[F] {
     val ws = resp.attributes.get(org.http4s.server.websocket.websocketKey[F])
     logger.debug(s"Websocket key: $ws\nRequest headers: " + req.headers)
 
-    if (ws.isDefined) {
-      val hdrs = req.headers.map(h => (h.name.toString, h.value))
-      if (WebsocketHandshake.isWebSocketRequest(hdrs)) {
-        WebsocketHandshake.serverHandshake(hdrs) match {
-          case Left((code, msg)) =>
-            logger.info(s"Invalid handshake $code, $msg")
-            async.unsafeRunAsync {
-              Response[F](Status.BadRequest)
-                .withBody(msg)
-                .map(
-                  _.replaceAllHeaders(
-                    Connection("close".ci),
-                    Header.Raw(headers.`Sec-WebSocket-Version`.name, "13")
-                  ))
-            } {
-              case Right(resp) =>
-                IO(super.renderResponse(req, resp, cleanup))
-              case Left(_) =>
-                IO.unit
-            }
+    ws match {
+      case None => super.renderResponse(req, resp, cleanup)
+      case Some(wsContext) =>
+        val hdrs = req.headers.map(h => (h.name.toString, h.value))
+        if (WebsocketHandshake.isWebSocketRequest(hdrs)) {
+          WebsocketHandshake.serverHandshake(hdrs) match {
+            case Left((code, msg)) =>
+              logger.info(s"Invalid handshake $code, $msg")
+              async.unsafeRunAsync {
+                wsContext.failureResponse
+                  .map(
+                    _.replaceAllHeaders(
+                      Connection("close".ci),
+                      Header.Raw(headers.`Sec-WebSocket-Version`.name, "13")
+                    ))
+              } {
+                case Right(resp) =>
+                  IO(super.renderResponse(req, resp, cleanup))
+                case Left(_) =>
+                  IO.unit
+              }
 
-          case Right(hdrs) => // Successful handshake
-            val sb = new StringBuilder
-            sb.append("HTTP/1.1 101 Switching Protocols\r\n")
-            hdrs.foreach {
-              case (k, v) => sb.append(k).append(": ").append(v).append('\r').append('\n')
-            }
+            case Right(hdrs) => // Successful handshake
+              val sb = new StringBuilder
+              sb.append("HTTP/1.1 101 Switching Protocols\r\n")
+              hdrs.foreach {
+                case (k, v) => sb.append(k).append(": ").append(v).append('\r').append('\n')
+              }
 
-            val wsContext = ws.get
-            wsContext.headers.foreach(hdr =>
-              sb.append(hdr.name).append(": ").append(hdr.value).append('\r').append('\n'))
+              wsContext.headers.foreach(hdr =>
+                sb.append(hdr.name).append(": ").append(hdr.value).append('\r').append('\n'))
 
-            sb.append('\r').append('\n')
+              sb.append('\r').append('\n')
 
-            // write the accept headers and reform the pipeline
-            channelWrite(ByteBuffer.wrap(sb.result().getBytes(ISO_8859_1))).onComplete {
-              case Success(_) =>
-                logger.debug("Switching pipeline segments for websocket")
+              // write the accept headers and reform the pipeline
+              channelWrite(ByteBuffer.wrap(sb.result().getBytes(ISO_8859_1))).onComplete {
+                case Success(_) =>
+                  logger.debug("Switching pipeline segments for websocket")
 
-                val segment = LeafBuilder(new Http4sWSStage[F](wsContext.webSocket))
-                  .prepend(new WSFrameAggregator)
-                  .prepend(new WebSocketDecoder(false))
+                  val segment = LeafBuilder(new Http4sWSStage[F](wsContext.webSocket))
+                    .prepend(new WSFrameAggregator)
+                    .prepend(new WebSocketDecoder(false))
 
-                this.replaceInline(segment)
+                  this.replaceInline(segment)
 
-              case Failure(t) => fatalError(t, "Error writing Websocket upgrade response")
-            }(executionContext)
-        }
-
-      } else super.renderResponse(req, resp, cleanup)
-    } else super.renderResponse(req, resp, cleanup)
+                case Failure(t) => fatalError(t, "Error writing Websocket upgrade response")
+              }(executionContext)
+          }
+        } else super.renderResponse(req, resp, cleanup)
+    }
   }
 }
--- a/server/src/main/scala/org/http4s/server/websocket/WebSocketBuilder.scala
+++ b/server/src/main/scala/org/http4s/server/websocket/WebSocketBuilder.scala
@@ -7,9 +7,8 @@ import org.http4s.websocket.WebsocketBits.WebSocketFrame
 import org.http4s.websocket.{WebSocketContext, Websocket}
 import org.http4s.{AttributeEntry, Headers, Response, Status}
 
-case class WebSocketBuilder[F[_]](
-    send: Stream[F, WebSocketFrame],
-    receive: Sink[F, WebSocketFrame])(implicit F: Monad[F]) {
+case class WebSocketBuilder[F[_]](send: Stream[F, WebSocketFrame], receive: Sink[F, WebSocketFrame])(
+    implicit F: Monad[F]) {
 
   private var onNonWebSocketRequest: F[Response[F]] =
     Response[F](Status.NotImplemented).withBody("This is a WebSocket route.")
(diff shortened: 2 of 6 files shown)


blaze-server/src/main/scala/org/http4s/server/blaze/WebSocketSupport.scala
```diff
@@ -52,14 +52,19 @@ private[blaze] trait WebSocketSupport[F[_]] extends Http1ServerStage[F] {
             hdrs.foreach {
               case (k, v) => sb.append(k).append(": ").append(v).append('\r').append('\n')
             }
+
+            val wsContext = ws.get
```
rossabaker: This is not your fault, but it's something we can clean up.  Can we do a match on the `ws` Option instead of `if (ws.isDefined) { ???; ws.get ; ??? } else { ??? }`?

server/src/main/scala/org/http4s/server/websocket/package.scala
```diff
@@ -37,12 +38,23 @@ package object websocket {
     *                 the case of a normal disconnection such as a Close handshake or due to a connection error. There
     *                 are plans to address this limitation in the future.
     * @param status The status code to return to a client making a non-websocket HTTP request to this route
+    * @param headers Websocket Handshake response headers, such as Sec-WebSocket-Protocol
     */
+  def WS[F[_]](
```
rossabaker: `WS` is kind of a strange name.  I think we could deprecate the old `WS` constructors and make the one we really want here.  I still like:

```
def WebSocketResponse(
    send: Stream[F, WebSocketFrame],
    receive: Sink[F, WebSocketFrame],
    headers: Headers = Headers.empty, // Sec-WebSocket-Protocol, Cookies, etc. go here
    onNonWebSocketRequest: F[Response[F]] = DefaultNonWebSocketResponse,
    onHandshakeFailure: F[Response[F]] = DefaultHandshakeFailureResponse)
```

That arranges the default arguments in the order they're most likely to be overridden.  If you want to set aside the custom handshake failure for now, you can just leave that off the end (as well as the context, which you've already done).

blaze-server/src/main/scala/org/http4s/server/blaze/WebSocketSupport.scala
jodagm: sure, I'll change that

server/src/main/scala/org/http4s/server/websocket/package.scala
jodagm: I need your help with the proper definition of DefaultNonWebSocketResponse, tried to do it with def so I can specify the F as its defined on def WS, but I think I got the implicit wrong.

jodagm: I think I should be doing
```
private def DefaultNonWebSocketResponse[F[_]](implicit F: Monad[F]) = Response[F](Status.NotImplemented).withBody("This is a WebSocket route.")
```

rossabaker: That looks right.  You could alternately use the shorthand:

`private def DefaultNonWebSocketResponse[F: Monad] = `

Your syntax gives you access to `F` if you want to invoke methods directly on the type class.  The shorthand just guarantees that `F` has a `Monad`.  I usually use the shorthand unless I need the named version in the definition.

jodagm: my issue then is with the using it as default param value:

```
onNonWebSocketRequest: F[Response[F]] = DefaultNonWebSocketResponse
```

getting 
```
Error:(59, 66) ambiguous implicit values:
 both value catsInstancesForId in package cats of type => cats.Bimonad[cats.package.Id] with cats.CommutativeMonad[cats.package.Id] with cats.Comonad[cats.package.Id] with cats.NonEmptyTraverse[cats.package.Id]
 and method catsStdInstancesForEither in trait EitherInstances of type [A]=> cats.MonadError[[β$0$]scala.util.Either[A,β$0$],A] with cats.Traverse[[β$1$]scala.util.Either[A,β$1$]]
 match expected type cats.Monad[F]
                         onNonWebSocketRequest: F[Response[F]] = DefaultNonWebSocketResponse,
```

jodagm: this is the entire method signature 

```
  def WebSocketResponseWS[F[_]](
                         send: Stream[F, WebSocketFrame],
                         receive: Sink[F, WebSocketFrame],
                         headers: Headers = Headers.empty, // Sec-WebSocket-Protocol, Cookies, etc. go here
                         onNonWebSocketRequest: F[Response[F]] = DefaultNonWebSocketResponse,
                         onHandshakeFailure: F[Response[F]] = DefaultHandshakeFailureResponse)(implicit F: Functor[F]): F[Response[F]]
```

rossabaker: Oh.  Hmm.  This is tricky.  `onNonWebsocketRequest` needs a `Monad[F]` instance.  You can provide one to `WebSocketResponseWS` (you currently have a `Functor[F]`), but that implicit doesn't apply to a defaul argument.

I'm going to sleep, but I started [rambling on Gitter about here](https://gitter.im/http4s/http4s?at=5a3b395f540c78242ded39a3).

aeons: The way this is handled other places is to have the constructor not take default arguments and then having and `apply` method in the companion object that has default parameters, which allows for type class constraints on the defaults.

jodagm: made it work with the builder approach and then saw @aeons comment
I am not getting it right with the companion object approach...
```
import cats._
import cats.implicits._
import fs2._
import org.http4s.websocket.WebsocketBits.WebSocketFrame
import org.http4s.websocket.{WebSocketContext, Websocket}
import org.http4s.{AttributeEntry, Headers, Response, Status}

case class WebSocketBuilder[F[_]](
                                   send: Stream[F, WebSocketFrame],
                                   receive: Sink[F, WebSocketFrame],
                                   headers: Headers,
                                   onNonWebSocketRequest: F[Response[F]],
                                   onHandshakeFailure: F[Response[F]]) {

  def build(send: Stream[F, WebSocketFrame],
            receive: Sink[F, WebSocketFrame],
            onNonWebSocketRequest: F[Response[F]],
            onHandshakeFailure: F[Response[F]])(implicit F: Monad[F]): F[Response[F]] = {
    onNonWebSocketRequest.map(
      _.withAttribute(AttributeEntry(
        websocketKey[F],
        WebSocketContext(Websocket(send, receive), headers, onHandshakeFailure))))
  }
}

object WebSocketBuilder {

  def apply[F[_]](send: Stream[F, WebSocketFrame],
                  receive: Sink[F, WebSocketFrame],
                  headers: Headers = Headers.empty,
                  onNonWebSocketRequest: F[Response[F]] = Response[F](Status.NotImplemented).withBody("This is a WebSocket route."),
                  onHandshakeFailure: F[Response[F]] = Response[F](Status.BadRequest).withBody("WebSocket handshake failed."))
                 (implicit F: Monad[F]): F[Response[F]] =
    WebSocketBuilder(send, receive, headers, onNonWebSocketRequest, onHandshakeFailure)
}
```


```
Error:(33, 102) could not find implicit value for parameter F: cats.Monad[F]
                  onNonWebSocketRequest: F[Response[F]] = Response[F](Status.NotImplemented).withBody("This is a WebSocket route."),
```

SystemFw: This is a minimisation of the problem, let's assume:
```scala
  import cats._, implicits._
  def default[F[_]: Monad]: F[Int] = 1.pure[F]
```
then:
```scala
  case class No[F[_]: Monad](s: String, f: F[Int] = default[F])
```
Does not compile. You can do:
```scala
  case class Yes[F[_]](s: String, f: F[Int])
  object Yes {
    class Builder[F[_]: Monad] {
      def apply(s: String, f: F[Int] = default[F]) = Yes(s, f)
    }

    def apply[F[_]: Monad] = new Builder[F]
  }

  def yes = Yes[Option].apply("yes") // Yes(yes, Some(1))
```
Ideally, we should be able to call `yes` with Yes[Option]("yes"),
but when we do that the compiler thinks we are passing evidence for
Monad explicitly, and complains. This is annoying, and there's a
proposal to fix it in the language (the `explicitly` keyword)

There is a solution if you're willing to crank up the magic though...
```scala
  // you lose apply, which you want to lose since it's overriden
  // you lose copy, which in this case it kinds sucks
  // you still have equals, hash and toString
  sealed abstract case class Magic[F[_]](s: String, f: F[Int])
  object Magic {
    // same builder as before
    class Builder[F[_]: Monad] {
      def apply(s: String, f: F[Int] = default[F]) = new Magic(s, f){}
    }
    // magic
    class Carrier[F[_]]
    object Carrier {
      implicit def conv[F[_]](c: Carrier[F])(implicit F: Monad[F]): Builder[F] =
        new Builder[F]
    }

    def apply[F[_]] = new Carrier[F]
  }

  def magic = Magic[Option]("magic") // Magic(magic, Some(1))
```

aeons: Yeah, I was a bit unclear in my explanation, You can do what @SystemFw suggests or see how it's done for BlazeBuilder here: https://github.com/http4s/http4s/blob/master/blaze-server/src/main/scala/org/http4s/server/blaze/BlazeBuilder.scala#L294

So the apply on the companion object only takes the implicit evidence, and then you use the `.with*` methods on the builder to change the things you want to change from the defaults.

server/src/main/scala/org/http4s/server/websocket/WebSocketBuilder.scala
```diff
@@ -0,0 +1,39 @@
...
+import org.http4s.{AttributeEntry, Headers, Response, Status}
+
+case class WebSocketBuilder[F[_]](send: Stream[F, WebSocketFrame], receive: Sink[F, WebSocketFrame])(
+    implicit F: Monad[F]) {
+
+  private var onNonWebSocketRequest: F[Response[F]] =
+    Response[F](Status.NotImplemented).withBody("This is a WebSocket route.")
+  private var onHandshakeFailure: F[Response[F]] =
+    Response[F](Status.BadRequest).withBody("WebSocket handshake failed.")
+  private var handshakeResponseHeaders: Headers = Headers.empty
```
aeons: This can be done without vars, see my other comment or look at `BlazeBuilder`: https://github.com/http4s/http4s/blob/master/blaze-server/src/main/scala/org/http4s/server/blaze/BlazeBuilder.scala

server/src/main/scala/org/http4s/server/websocket/package.scala
SystemFw: fwiw, I think the best tradeoff between convenience and clarity is method no.2, except you rename `apply` to `build`, and use default parameters rather than multiple `with*` methods.
Method no.3 is interesting enough to be worth experimenting with imho, but it's a bit of a tradeoff (error messages, weird encoding)

rossabaker:
What does everybody think about the builder approach vs. the carrier solution?  The carrier is a cleaner call when done right, but more complicated machinery if the call is made wrong.

rossabaker:
Oops, I had missed the further conversation in my previous comment.

server/src/main/scala/org/http4s/server/websocket/WebSocketBuilder.scala
jodagm: @rossabaker @aeons @SystemFw thank you for your feedback, I am learning a lot from you guys. I used BlazeBuilder as reference. Still wrapping my head around the (dark) Magic example, but as its now 02:00 for me, I will sure have to return to this tomorrow morning. I hope the current state of the code is good enough to be approved, though.

aeons:
I like the `.build` version.

server/src/main/scala/org/http4s/server/websocket/package.scala
aeons: I tried using the partially applied approach on BlazeBuilder, but it fails if you don't set any of the arguments.

`BlazeBuilder.build[IO]` fails, `BlazeBuilder.build[IO]()` fails since it expects the `Effect` evidence there.

With the carrier, you still have to give it empty parens to make it work, which imo is worse ergonomics than what we have currently, which works with `BlazeBuilder[IO].with...`

server/src/main/scala/org/http4s/server/websocket/package.scala
```diff
@@ -38,17 +38,19 @@ package object websocket {
...
-    WebSocketBuilder(
+    WebSocketBuilder[F].build(
       send,
       receive,
       Headers.empty,
       status,
-      Response[F](Status.BadRequest).withBody("WebSocket handshake failed.")).toResponse
+      Response[F](Status.BadRequest).withBody("WebSocket handshake failed."))
 
+  @deprecated("Use WebSocketBuilder", "0.18")
```
rsoeldner: Shouldn't the version be "0.18.0-M7" ?

jodagm: fixing

jodagm: is it failing because BlazeWebSocketExampleApp uses deprecated WS?

server/src/main/scala/org/http4s/server/websocket/package.scala
SystemFw: yeah you need `BlazeBuilder[IO].build()`. The parenthesis are unavoidable due to the way default arguments work in Scala (for some mysterious reason), it probably works best in cases where there's at least one argument that does not have a default

blaze-server/src/main/scala/org/http4s/server/blaze/WebSocketSupport.scala
```diff
@@ -25,51 +25,55 @@ private[blaze] trait WebSocketSupport[F[_]] extends Http1ServerStage[F] {
...
+              async.unsafeRunAsync {
+                wsContext.failureResponse
+                  .map(
+                    _.replaceAllHeaders(
+                      Connection("close".ci),
+                      Header.Raw(headers.`Sec-WebSocket-Version`.name, "13")
+                    ))
+              } {
+                case Right(resp) =>
+                  IO(super.renderResponse(req, resp, cleanup))
```
rsoeldner: This should be `F.delay` and not `IO`

blaze-server/src/main/scala/org/http4s/server/blaze/WebSocketSupport.scala
```diff
@@ -25,51 +25,55 @@ private[blaze] trait WebSocketSupport[F[_]] extends Http1ServerStage[F] {
...
+                case Left(_) =>
+                  IO.unit
```
rsoeldner: same here, `F.unit`

PR #1608, opened 2017-12-21
Upgrade to scalaz-7.2.18, scalaz-7.1.16

rossabaker:


rossabaker:
DNLGTT (Does Not Look Good to Travis)

I'll fix the deprecation tomorrow.

PR #1610, opened 2017-12-22
Move ByteChunkSyntax to the org.http4s.syntax package

aeons:
To be consistent with other syntax enrichments

### DIFF
--- a/core/src/main/scala/org/http4s/syntax/AllSyntax.scala
+++ b/core/src/main/scala/org/http4s/syntax/AllSyntax.scala
@@ -4,6 +4,7 @@ package syntax
 trait AllSyntax
     extends AnyRef
     with AsyncSyntax
+    with ByteChunkSyntax
     with EffectResponseSyntax
     with EffectRequestSyntax
     with KleisliSyntax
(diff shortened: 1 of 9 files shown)


rsoeldner:
I missed this, good catch :+1:

aeons:
Yeah, it's been  thorn in my side for a long time :)

core/src/main/scala/org/http4s/syntax/AllSyntax.scala
```diff
@@ -4,6 +4,7 @@ package syntax
 trait AllSyntax
     extends AnyRef
     with AsyncSyntax
+    with ByteChunkSyntax
```
rossabaker: One reason this was separate was because we don't own either the type class or the instance type.  Every time we've done this, we've said, "this missing instance/operation should be contributed upstream."  I think this should be an internal syntax until we can get rid of it, either by contributing it to fs2-scodec or dropping fs2-scodec.  (I still suspect we can do the latter.)

aeons: Fair enough. Let's scrap the entire PR for now then.

PR #1612, opened 2017-12-23
Metrics fix

### LINKED ISSUE #1606: server-metrics don't handle requests to the `metricsService` correctly.
I tried out the server-metrics package today and its [BlazeServerExample](/http4s/http4s/blob/master/examples/blaze/src/main/scala/com/example/http4s/blaze/BlazeMetricsExample.scala).  Nice work, it looks really promising.  

An oddity that I noticed was that the `active_requests` would increment but not decrement when the "/metrics" endpoint is invoked. It seems that the code in Metrics HttpMiddleware would increment but because the `prefix` was not found, it would never call funciton `generalMetrics`.

ChristopherDavenport:
1) GET /http4s/metrics does not increment counters (it’s not wrapped in the middleware)
2) GET /http4s/metrics does not leave a dangling active request
3) GET /http4s/dfdajkfldajfklddasfadskl does not leave a dangling active request. (This is one where a wrapped service returns None).
4) GET /http4s/ping increments 200 and does not leaving dangling active request

Closes #1606

### DIFF
--- a/server-metrics/src/main/scala/org/http4s/server/metrics/Metrics.scala
+++ b/server-metrics/src/main/scala/org/http4s/server/metrics/Metrics.scala
@@ -11,7 +11,7 @@ import java.util.concurrent.TimeUnit
 
 object Metrics {
 
-  def apply[F[_]](m: MetricRegistry, prefix: String = "org.http4s.server", )(
+  def apply[F[_]](m: MetricRegistry, prefix: String = "org.http4s.server")(
       implicit F: Effect[F]): HttpMiddleware[F] = { service =>
     val generalServiceMetrics = GeneralServiceMetrics(
       active_requests = m.counter(s"${prefix}.active-requests"),
@@ -27,17 +27,17 @@ object Metrics {
       resp5xx = m.timer(s"${prefix}.5xx-responses")
     )
     val requestTimers = RequestTimers(
-      get_req = m.timer(s"${prefix}.get-requests"),
-      post_req = m.timer(s"${prefix}.post-requests"),
-      put_req = m.timer(s"${prefix}.put-requests"),
-      head_req = m.timer(s"${prefix}.head-requests"),
-      move_req = m.timer(s"${prefix}.move-requests"),
-      options_req = m.timer(s"${prefix}.options-requests"),
-      trace_req = m.timer(s"${prefix}.trace-requests"),
-      connect_req = m.timer(s"${prefix}.connect-requests"),
-      delete_req = m.timer(s"${prefix}.delete-requests"),
-      other_req = m.timer(s"${prefix}.other-requests"),
-      total_req = m.timer(s"${prefix}.requests")
+      getReq = m.timer(s"${prefix}.get-requests"),
+      postReq = m.timer(s"${prefix}.post-requests"),
+      putReq = m.timer(s"${prefix}.put-requests"),
+      headReq = m.timer(s"${prefix}.head-requests"),
+      moveReq = m.timer(s"${prefix}.move-requests"),
+      optionsReq = m.timer(s"${prefix}.options-requests"),
+      traceReq = m.timer(s"${prefix}.trace-requests"),
+      connectReq = m.timer(s"${prefix}.connect-requests"),
+      deleteReq = m.timer(s"${prefix}.delete-requests"),
+      otherReq = m.timer(s"${prefix}.other-requests"),
+      totalReq = m.timer(s"${prefix}.requests")
     )
 
     val serviceMetrics = ServiceMetrics(generalServiceMetrics, requestTimers, responseTimers)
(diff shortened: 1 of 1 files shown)


server-metrics/src/main/scala/org/http4s/server/metrics/Metrics.scala
```diff
@@ -2,102 +2,183 @@ package org.http4s
...
+  private def manageResponse[F[_]: Sync](
+                                   m: Method,
+                                   elapsed: Long,
+                                   serviceMetrics: ServiceMetrics
+                                 )(response: Response[F]): Response[F] = {
+    val newBody = response.body
+      .onFinalize(
+        incrementCounts(serviceMetrics.generalMetrics.headers_times, elapsed) *>
+        requestMetrics(serviceMetrics.requestTimers, serviceMetrics.generalMetrics.active_requests)(m, elapsed) *>
+        responseMetrics(serviceMetrics.responseTimers, response.status, elapsed)
```
rossabaker: These aren't measuring all the way to the end of the body now, right?  That's the difference between `headers_times` and the other timers.  I think these two should be measured to when the finalizer triggers.

server-metrics/src/main/scala/org/http4s/server/metrics/Metrics.scala
```diff
@@ -2,102 +2,183 @@ package org.http4s
...
+    response.copy(body = newBody)
+  }
+
+  private def manageServiceErrors[F[_]: Sync](m: Method, elapsed: Long, serviceMetrics: ServiceMetrics): F[Unit] =
+    requestMetrics(serviceMetrics.requestTimers, serviceMetrics.generalMetrics.active_requests)(m, elapsed) *>
+    incrementCounts(serviceMetrics.responseTimers.resp5xx, elapsed) *>
+    incrementCounts(serviceMetrics.generalMetrics.service_errors, elapsed)
+
+
+  private def handleUnmatched[F[_]: Sync](c: Counter): F[Option[Response[F]]] =
```
rossabaker: If we increment 5xx on these, should we increment 4xx on this?  This is the dirty bit about doing metrics as middleware: we don't know what happens outside middleware.  Maybe we're chained to something that recovers.  Maybe the server handles it.  I think this is inconsistent with `manageServiceErrors`, though I don't have a strong feeling which direction to fix it.

ChristopherDavenport: Yeah. I can increment as it without it we don't match requirement 4

ChristopherDavenport: I question whether us projecting a response is the right move as well. Metrics on a service should measure what happens with that service. For metrics matching unmatched They should make the service total

rossabaker: :+1: I think we just count nones and errors and not double count as 4xx and 5xx. Which is a reversal of point 4 above, but I think correct.

ChristopherDavenport:
Fixes #1606

server-metrics/src/main/scala/org/http4s/server/metrics/Metrics.scala
```diff
@@ -2,102 +2,203 @@ package org.http4s
...
+    val method = req.method
+    for {
+      now <- OptionT.liftF[F, Long](Sync[F].delay(System.nanoTime()))
+      _ <- OptionT.liftF[F, Unit](
+        Sync[F].delay(serviceMetrics.generalMetrics.active_requests.inc()))
+      resp <- OptionT {
         service(req).value.attempt
-          .flatMap(onFinish(req.method, now)(_)
-            .fold[F[Option[Response[F]]]](F.raiseError, F.pure))
+          .flatMap(metricsServiceHandler(method, now, serviceMetrics, _))
```
jmcardon: This block seems redundant to `liftF` multiple times, then wrap the last one. 

You can instead make the comprehension only in terms of `F` and wrap the whole thing. It would be better since `liftF` causes a `map(Some(_))`.

server-metrics/src/main/scala/org/http4s/server/metrics/Metrics.scala
```diff
@@ -2,102 +2,203 @@ package org.http4s
...
+      active_requests: Counter
+  )(method: Method, elapsed: Long): F[Unit] = {
+    val timer = requestTimer(rt, method)
+    incrementCounts(timer, elapsed) *>
+      incrementCounts(rt.total_req, elapsed) *>
+      Sync[F].delay(active_requests.dec())
   }
+
+  private case class RequestTimers(
+      get_req: Timer,
```
jmcardon: re: gitter discussion, could we go ahead and maybe get rid of the snake case?

PR #1617, opened 2017-12-27
Remove scodec-bits dependency

rossabaker:


core/src/main/scala/org/http4s/internal/chunk.scala
```diff
@@ -0,0 +1,60 @@
+package org.http4s.internal
```
rossabaker: This whole file can go away if and when https://github.com/functional-streams-for-scala/fs2/pull/1032 gets merged.

blaze-core/src/test/scala/org/http4s/blazecore/ResponseParser.scala
```diff
@@ -23,7 +22,7 @@ class ResponseParser extends Http1ClientParser {
   /** Will not mutate the ByteBuffers in the Seq */
   def parseResponse(buffs: Seq[ByteBuffer]): (Status, Set[Header], String) = {
     val b =
-      ByteBuffer.wrap(buffs.map(b => ByteVectorChunk(ByteVector.view(b)).toArray).toArray.flatten)
+      ByteBuffer.wrap(buffs.map(b => Chunk.byteBuffer(b).toArray).toArray.flatten)
```
jmcardon: two `toArray` calls might be correct, but seems fishy... In particular, because toArray will cause copying, vs just finding a way to append the array and the offsets.

I could be wrong but this line warrants investigation imo.

jmcardon: For the sake of documenting, im adding a note about: https://gitter.im/http4s/http4s?at=5a43d98aedd2230811e3970b

jmcardon: Retracting my previous comment: 

This is in the blaze core test package. Max efficiency here won't exactly be of any use outside of testing.

rossabaker:
Superseded by #1701.

PR #1619, opened 2017-12-29
Upgrade deps

aeons:
Building on top of #1615, this PR updates fs2, circe, jawn-fs2 and fs2-reactive-streams to currently unpublished  versions that build on top of cats 1.0. It will not build until those versions are released, but I have tested it with locally published versions.

The version numbers are guesses, so they might need updates.


aeons:
There's an M11 of fs2, so we'll go to that first. There are PRs for fs2-reactive-streams and jawn-fs2 pending.

PR #1620, opened 2018-01-06
Prep Changelog and Update Version for 0.18.0-M8

ChristopherDavenport:
This should be set to fire off an M8 Release.


rossabaker:
:+1:

PR #1623, opened 2018-01-07
Modify client documentation to use Http1Client.stream

BenFradet:


docs/src/main/tut/client.md
```diff
@@ -84,14 +83,14 @@ import cats._, cats.effect._, cats.implicits._
...
   val target = Uri.uri("http://localhost:8080/hello/") / name
-  httpClient.expect[String](target)
+  httpClient.evalMap(_.expect[String](target))
 }
 
-val people = Vector("Michael", "Jessica", "Ashley", "Christopher")
+val people = fs2.Stream("Michael", "Jessica", "Ashley", "Christopher").covary[IO]
 
-val greetingList = fs2.async.parallelTraverse(people)(hello)
+val greetingList = people.flatMap(hello).compile.toVector
```
BenFradet: we might want to leverage `concurrently` here to preserve the parallel nature of the original example, what do people think?

aeons:
This works, but it is also "wrong" in that it's now how you should be using it. Everytime you `.evalMap` the stream you create a new client.

I guess it might be better than having an `.unsafeRunSync`, but there should be some more mention that you should only create one client for your program.

BenFradet:
Ah, good to know. Then, I think this is more confusing than what was there originally, closing.

PR #1625, opened 2018-01-09
`BlazeBuilder` now properly shuts down its `ExecutorService`

jedesah:
This means shutdown behavior of `http4s` will be as expected when using non-deamon threads

### DIFF
--- a/blaze-server/src/main/scala/org/http4s/server/blaze/BlazeServer.scala
+++ b/blaze-server/src/main/scala/org/http4s/server/blaze/BlazeServer.scala
@@ -192,6 +192,7 @@ class BlazeBuilder(
       override def shutdown: Task[Unit] = Task.delay {
         serverChannel.close()
         factory.closeGroup()
+        serviceExecutor.shutdown()
       }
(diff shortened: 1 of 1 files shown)


aeons:
Can we be certain that the executor service is not used elsewhere?

jedesah:
@aeons No we can't, that's a very good observation. I'm most certainly against closing things that "we" didn't create.

I guess the reason I didn't think of it in that way is that the common scenario is for `BlazeBuilder` to use `DefaultPool`. `DefaultPool` isn't even lazy which means it's created any time `http4s` is on the classpath (right?). I mean that seems a little wasteful even if it didn't create this problem where no nothing is responsible for closing it. Luckily, it's sort of effectively lazy because it appears it doesn't create any threads if no one uses it which means applications would terminate normally even if it were changed to create non-daemon threads (I tested this).

@aeons I think getting rid of `DefaultPool` is the appropriate fix here. Luckily it's private so I think it's an acceptable change within a minor release.

Let me have a look at what that would look like

PR #1631, opened 2018-01-17
Websocket: Update To Unsafe Signal for Close Rather than Pure Signal

ChristopherDavenport:


blaze-core/src/main/scala/org/http4s/blazecore/websocket/Http4sWSStage.scala
```diff
@@ -80,18 +74,20 @@ class Http4sWSStage[F[_]](ws: ws4s.Websocket[F])(implicit F: Effect[F], val ec:
     val onStreamFinalize: F[Unit] =
       for {
         dec <- F.delay(count.decrementAndGet())
-        _ <- deadSignal.map(signal => if (dec == 0) signal.set(true))
+        _ <- if (dec == 0) deadSignal.set(true) else ().pure[F]
```
SystemFw: `().pure[F]` `-->` `F.unit`?

nicer, but utterly unimportant ofc

blaze-core/src/main/scala/org/http4s/blazecore/websocket/Http4sWSStage.scala
```diff
@@ -80,18 +74,20 @@ class Http4sWSStage[F[_]](ws: ws4s.Websocket[F])(implicit F: Effect[F], val ec:
...
-    val wsStream = for {
-      dead <- deadSignal
-      in = inputstream.to(ws.receive).onFinalize(onStreamFinalize)
-      out = ws.send.onFinalize(onStreamFinalize).to(snk).drain
-      merged <- in.mergeHaltR(out).interruptWhen(dead).onFinalize(sendClose).compile.drain
-    } yield merged
+    val wsStream = inputstream
+      .to(ws.receive)
+      .onFinalize(onStreamFinalize)
+      .mergeHaltR(ws.send.onFinalize(onStreamFinalize).to(snk).drain)
```
SystemFw: `mergeHaltR(foo.drain)` can probably be `concurrently(foo)` now

SystemFw:
Left a few style comments, but nothing too important. I'm also assuming this has been tested

blaze-core/src/main/scala/org/http4s/blazecore/websocket/Http4sWSStage.scala
SystemFw: Actually scratch that sorry, you nee the output of `foo`, but want to complete when `bar` is done

SystemFw:
@ChristopherDavenport you were too fast!

SystemFw:
so I was imprecise, `concurrently` replaces `mergeHaltL` not `mergeHaltR`, so if you indeed want to finish the right branch finishes (but still return the output of the right branch), you indeed need what you used to have (sorry :grin:)

ChristopherDavenport:
:(

PR #1634, opened 2018-01-19
Add expectOr methods for custom error handling on failed expects

### LINKED ISSUE #1632: [Feature Request] `Client.expect` gives the option of including body in the `UnexpectedStatus` error
Many API only gives the detailed error information in the body, including the body in the `UnexpectedStatus` would be really helpful. 
Right now I am using my own implementation of `expect` like 
```scala
def expect[A](req: Request[F], onError: Option[(String, Status) => Throwable] = None)
 (implicit d: EntityDecoder[F, A]): F[A] = {
   val r = if (d.consumes.nonEmpty) {
        val m = d.consumes.toList
        req.putHeaders(Accept(MediaRangeAndQValue(m.head), 
      m.tail.map(MediaRangeAndQValue(_)): _*))
      } else req
   http.fetch(r) {
      case Successful(resp) =>
        d.decode(resp, strict = false).fold(throw _, identity)
      case failedResponse =>
        onError.fold(F.raiseError(UnexpectedStatus(failedResponse.status))) { f => 
          EntityDecoder[F, String].decode(failedResponse, false)
            .fold(identity, f(_, failedResponse.status).flatMap(F.raiseError)
       }
    }
}
```
If would be nice to have an idiomatic way to do this.

rossabaker:
I'm already a bit uncomfortable with the number of overloads of `expect`, and this doubles down on the discomfort in the name of consistency.

We might still argue that the default behavior is bad, and `UnexpectedStatus` should capture the response headers and the (first N characters of?) the body.  Not doing so avoids a second layer of error handling if the body can't be decoded to character data, or what a reasonable cutoff is.

Fixes #1632


client/src/main/scala/org/http4s/client/Client.scala
```diff
@@ -128,12 +128,7 @@ final case class Client[F[_]](
   def streaming[A](req: F[Request[F]])(f: Response[F] => Stream[F, A]): Stream[F, A] =
     Stream.eval(req).flatMap(streaming(_)(f))
 
-  /**
-    * Submits a request and decodes the response on success.  On failure, the
-    * status code is returned.  The underlying HTTP connection is closed at the
-    * completion of the decoding.
-    */
-  def expect[A](req: Request[F])(implicit d: EntityDecoder[F, A]): F[A] = {
+  def expectOr[A](req: Request[F])(onError: Response[F] => F[Throwable])(implicit d: EntityDecoder[F, A]): F[A] = {
```
wedens: I think it can be generalized to
```scala
def expectOr[A](req: Request[F])(onError: Response[F] => F[A])(implicit d: EntityDecoder[F, A]): F[A]
```

kailuowang: The downside then is it's a bit more cumbersome to provide a generic onError handler for all `A`s. Especially if we want make it part of the client conf. But I can see benefits of being able to directly recover from response .

jmcardon: I don't think this is a generalization when you might possibly not want to return an `A`.

It's much easier to handle this outside the function (via handleError) than not. I think I prefer the way it is.

wedens: If you don't want to return an `A`, you can just `F.raiseError(whatever)`.

kailuowang: Returing F[A] is more flexible but the API  gives the impression that the intention here is a recovery. User can always do a recovery with F after, so making it raising an error on non-200 status seems a good design to me.

kailuowang:
I am still feeling the pain of not having this on a regular basis...

PR #1637, opened 2018-01-23
Call Async Callback to Correctly Initiate Shutdown

ChristopherDavenport:
Fixes Concurrently Leakage

I don't know why we were even using async here originally but by not calling the provided callback we broke lots of stuff. Lesson for the day, don't do that.


rossabaker:
Does serve still block until interrupted? That was the original intent.  I'm afraid this will make Stream apps terminate immediately.

ChristopherDavenport:
No. It terminates now immediately. Making an issue and trying to work through it this morning.

aeons:
I "fixed" this in the initial transition to cats as well, (stupider though :)) in https://github.com/http4s/http4s/pull/1266

And reverted it in https://github.com/http4s/http4s/pull/1334

PR #1638, opened 2018-01-23
Fix immediately returning serve

ChristopherDavenport:
Not sure about the EC, but once translate bug is fixed this work as expected.

### DIFF
--- a/server/src/main/scala/org/http4s/server/ServerBuilder.scala
+++ b/server/src/main/scala/org/http4s/server/ServerBuilder.scala
@@ -1,6 +1,7 @@
 package org.http4s
 package server
 
+import cats.implicits._
 import cats.effect._
 import fs2.StreamApp.ExitCode
 import fs2._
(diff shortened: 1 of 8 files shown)


server/src/main/scala/org/http4s/server/ServerBuilder.scala
```diff
@@ -53,10 +54,21 @@ trait ServerBuilder[F[_]] {
...
+    * Useful for servers with associated lifetime behaviors.
+    */
+  final def serveWhile(terminateWhenTrue: Signal[F, Boolean])(
+      implicit F: Sync[F]): Stream[F, ExitCode] =
     Stream.bracket(start)(
-      (_: Server[F]) => Stream.eval_(F.async[Unit](cb => cb(Right(())))),
-      _.shutdown)
+      (_: Server[F]) =>
+        Stream.eval(terminateWhenTrue.discrete.takeWhile(_ == false).compile.drain) >>
+          Stream.emit(StreamApp.ExitCode(0)).covary[F],
```
aeons: This should use the `Success` constant.

aeons:
👍 on the exit code ref

PR #1640, opened 2018-01-26
extract filename for multipart

jaspervz:
The filename is something that is often needed when handling a multi part encoded request, having a method to extract this from a part is helpful.

### DIFF
--- a/core/src/main/scala/org/http4s/multipart/Part.scala
+++ b/core/src/main/scala/org/http4s/multipart/Part.scala
@@ -11,7 +11,8 @@ import org.http4s.headers.`Content-Disposition`
 
 final case class Part[F[_]](headers: Headers, body: Stream[F, Byte]) {
   def name: Option[String] = headers.get(`Content-Disposition`).flatMap(_.parameters.get("name"))
-  def filename: Option[String] = headers.get(`Content-Disposition`).flatMap(_.parameters.get("filename"))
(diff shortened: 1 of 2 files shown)


ritschwumm:
shouldn't this parse `filename*`, too?

jcranky:
The RFC also mentions `to be matched case-insensitively`, so maybe something else to take into account?

ritschwumm:
@jcranky looks like it - at a quick glance it seems parameter names are passed literally from the parser, without any case-mangling.

rossabaker:
Is there any reason that modeled shouldn't have CaseInsensitiveStrings as keys then?  (I bet this is true of a lot of the parameter-based headers.)

jaspervz:
RFC 6266 states in the introduction that it doesn't apply to the Content-Disposition header used for multipart/form-data, for which this pull request is.

ritschwumm:
@jaspervz damn, sorry - i knew there are 2 separate RFCs and still got the wrong one (again) :(

jmcardon:
@rossabaker via https://tools.ietf.org/html/rfc2183 section 2, "filename" is case sensitive as stated by the BNF form of content-disposition. I think this one is ok to not be `CaseInsensitiveString`

jcranky:
On face of 2183 this seems ok indeed. It is confusing to have an RFC that "almost" fits the bill :)

PR #1641, opened 2018-01-27
middleware documentation refactoring

reactormonk:


docs/src/main/tut/middleware.md
```diff
@@ -4,35 +4,34 @@ title: Middleware
...
 libraryDependencies ++= Seq(
   "org.http4s" %% "http4s-dsl" % http4sVersion
 )
 ```
-and some imports.
+1. Add the following import statements:
 
 ```tut:silent
-import cats.effect._
+import cats.effect._ // To import the IO type, see cats-effect documentation.
```
rossabaker: What does this mean?  To specialize on the IO type?  We've already imported the IO type here.

docs/src/main/tut/middleware.md
```diff
@@ -102,21 +102,23 @@ newService.orNotFound(goodRequest).unsafeRunSync
...
 ```
 
-It is possible for the wrapped `Service` to have different `Request` and `Response`
-types than the middleware. Authentication is, again, a good example. Authentication
-middleware is an `HttpService` (an alias for `Service[Request, Response]`) that wraps an `
-AuthedService` (an alias for `Service[AuthedRequest[T], Response]`. There is a type
-defined for this in the `http4s.server` package:
+It is possible for the wrapped `Service` to have different `Request` and
+`Response` types than the `Middleware`. `Authentication` is, a good example.
+`Authentication` `Middleware` is an `HttpService` (an alias for `Service[Request,Response]`)
```
rossabaker: Is it worth mentioning that `Request` and `Response` have a type parameter here?  I don't want to confuse the issue in informal descriptions, but think it might be useful here where we're showing actual signatures.

docs/src/main/tut/middleware.md
reactormonk: "This import statement is to import the `IO` type. For more information about `IO`, see cats-effect."

docs/src/main/tut/middleware.md
reactormonk: Does the other documentation mention the type parameters? I'd vote for unification.

aeons:
There are quite a few references to `Service`, but that type alias was removed some time ago. So there's a raw `Kleisli[F, A, B]` and `HttpService[F] = Kleisli[OptionT[F, ?], Request[F], Response[F]]` (iirc).

docs/src/main/tut/middleware.md
jcranky: Why not use the specific import? I guess this is a bit redundant :)

docs/src/main/tut/middleware.md
rossabaker: I'm not sure that all the documentation has been updated for that in the areas that aren't compiled.  I think we should do it here, and then additionally take a pass through the rest.  Unify on telling it like it is.

rossabaker:
Do we still want to target this for 0.18, or with #1693, do we want to wait for that and then finish this?

PR #1642, opened 2018-01-29
Upgrade to async-http-client-2.4

rossabaker:


rossabaker:
[This commit](https://github.com/AsyncHttpClient/async-http-client/commit/391b4d8e0eced22e56f390d4c5ee8a51c1c98d27) broke us:

```
[error]   x POST a normal body (1 second, 356 ms)
[error]    an exception was thrown Null subscription (ClientRouteTestBattery.scala:84)
[error] org.http4s.client.ClientRouteTestBattery.$anonfun$new$24(ClientRouteTestBattery.scala:84)
[error]   x POST a chunked body (1 second, 369 ms)
[error]    an exception was thrown Null subscription (ClientRouteTestBattery.scala:91)
[error] org.http4s.client.ClientRouteTestBattery.$anonfun$new$27(ClientRouteTestBattery.scala:91)
```

I'm feeling dizzy, and that explanation made me feel dizzier.  I'll try again tomorrow unless one of our other Certified Reactive Streams Experts sees what we can do.

rossabaker:
2.4.0 is out now, but has the same problem.  It still appears to be that commit.

rossabaker:
I think https://github.com/AsyncHttpClient/async-http-client/pull/1522 will fix this.

PR #1650, opened 2018-02-08
Make authentication doc code silent

jcranky:
The REPL output is just adding noise and making reading the documentation
harder.

### DIFF
--- a/docs/src/main/tut/auth.md
+++ b/docs/src/main/tut/auth.md
@@ -16,7 +16,7 @@ function, or more likely, a `Request[F] => OptionT[F, User]`, because the `User`
 come from a database. We can convert that into an `AuthMiddleware` and apply it.
 Or in code, using `cats.effect.IO` as the effect type:
 
-```tut:book
+```tut:book:silent
 import cats._, cats.effect._, cats.implicits._, cats.data._
 import org.http4s._
 import org.http4s.dsl.io._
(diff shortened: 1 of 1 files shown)


jcranky:
Their simple existence is noisy :p 

Specially with the imports:

```scala
import cats._, cats.effect._, cats.implicits._, cats.data._
// import cats._
// import cats.effect._
// import cats.implicits._
// import cats.data._
```

And since it also adds empty lines, it makes the code examples three times longer. When I get to the "meat of the matter", I already forgot what I was looking at.

Plus, they are not really adding value: they essentially display the types of the `val`s, which are already explicitly defined anyway.

jcranky:
Plus, even if we don't "show it", `tut` is compiling it anyway, right?

ChristopherDavenport:
Yes, but when tut compiles, it will happily give you errors and bad code responses. So long as it can be compiled. I guess as a result I could be convinced, but I've caught deprecations and other things left in docs through tut output before.

jcranky:
> Yes, but when tut compiles, it will happily give you errors and bad code responses. 
You mean it doesn't break the build?

jcranky:
Or I don't understand which errors you are talking about :)

rossabaker:
I usually try to make imports silent.  For other code, I suppose the insight of the `toString` of the result is what matters.

jmcardon:
The only reason I like non-silent tut at times is to understand the particular return types of certain functions that are not obvious from the pov of a user.

For the Authentication portion, I find this somewhat important, so I'm lukewarm about this change. Showing the inferred type of the compiler can be cool especially when we have some many people on gitter coming in to ask about how to use it since kleisli tends to not be extremely obvious to the non-fp crowd.

jcranky:
The inferred type type is adding nothing: the types are explicit just about everywhere. From a reader perspective, this is what I'm calling noisy:

selection_006

Where is the `toString` adding anything? There is just the pollution, which is not helping - and that is coming from someone that doesn't fully understand what that page is trying to tell.

Anyway, if everyone else thinks otherwise, I'm happy to close this PR.

Issue #1511, opened 2017-11-01
Missing ETag for Static files


pvillega:
[Documentation](http://http4s.org/v0.17/static/) for `Static files` says:

> Http4s provides a few helpers to handle ETags for you, they’re located in StaticFile.

Unfortunately the implementation of [StaticFile](https://github.com/http4s/http4s/blob/ecaa1155307fc16616315237ace7280455291b62/core/src/main/scala/org/http4s/StaticFile.scala) is not using `ETag` but `Last-Modified` headers. The documented functionality is missing.

rossabaker:
Going ahead and calling this one a bug since the docs promise it.

There are different strategies for creating ETags, so I suppose the first question would be how they should be generated.  I imagine some sort of parameterized strategy would be desired, because it's not one-size-fits-all.

jmcardon:
#1652 closed this, no?

ChristopherDavenport:
Closed By #1652

PR #1652, opened 2018-02-08
For #1511 ETag support

eklavya:


core/src/main/scala/org/http4s/StaticFile.scala
```diff
@@ -74,17 +74,28 @@ object StaticFile {
       }
     })
 
-  def fromFile[F[_]: Sync](f: File, req: Option[Request[F]] = None): OptionT[F, Response[F]] =
-    fromFile(f, DefaultBufferSize, req)
+  def calcETag: File => String = _.length().toString
```
aeons: `File.length()` can throw and should probably be wrapped in `F`.

eklavya: It's only ever called from inside F.delay

ChristopherDavenport: Not referentially transparent is not referentially transparent.

eklavya: I didn't get you, all the file calls aren't, they are all inside one F.delay block.

ChristopherDavenport: This function is not referentially transparent. It is a top level public function that does not respect its own function signature. If it was used safely everywhere or private that would not change the fact that the function itself does not respect its own signature.

core/src/main/scala/org/http4s/StaticFile.scala
```diff
@@ -74,59 +74,91 @@ object StaticFile {
...
-          Some(r)
+  def fromFile[F[_]](
+      f: File,
+      start: Long,
+      end: Long,
+      buffsize: Int,
+      req: Option[Request[F]],
+      etagCalculator: File => F[String])(implicit F: Sync[F]): OptionT[F, Response[F]] =
+    OptionT(for {
+      etagCalc <- etagCalculator(f).map(et => ETag(et))
```
rossabaker: Do we need to `f.isFile` check before passing off to the calculator?

core/src/main/scala/org/http4s/StaticFile.scala
```diff
@@ -74,59 +74,91 @@ object StaticFile {
       }
     })
 
+  def calcETag[F[_]: Sync]: File => F[String] = f => Sync[F].delay(f.length().toString)
```
rossabaker: Is this a sensible default?  I tend to think of hashes of the content for accuracy, or maybe a combination of the timestamp and length for speed.

core/src/main/scala/org/http4s/StaticFile.scala
```diff
@@ -74,7 +74,17 @@ object StaticFile {
       }
     })
 
-  def calcETag[F[_]: Sync]: File => F[String] = f => Sync[F].delay(f.length().toString)
+  def calcETag[F[_]: Sync]: File => F[String] =
+    f =>
+      Sync[F]
+        .delay(f.isFile)
+        .flatMap { isFile =>
+          if (isFile) {
```
rossabaker: We could probably micro-optimize this into a single delay rather than a flatMap.  It's all local.

core/src/main/scala/org/http4s/StaticFile.scala
```diff
@@ -74,7 +74,17 @@ object StaticFile {
...
+            Sync[F].delay(s"${f.lastModified()}${f.length()}")
```
rossabaker: nginx uses these two components, so we're in good company here.  They render as hex, which I'm not sure is significant.  They also have a `-` delimiter, which is pedantically safer.  (`lastModified` shouldn't ever vary by an order of magnitude, but if it did, a collision could happen.)

core/src/main/scala/org/http4s/StaticFile.scala
```diff
@@ -74,7 +74,17 @@ object StaticFile {
...
+          } else {
+            Sync[F].pure("")
```
rossabaker: Is this what we want, or do we want to omit the header?  I don't think an empty `Etag` makes any sense.  Maybe the calculator should return an `Option`.

core/src/main/scala/org/http4s/StaticFile.scala
eklavya: Single flatMap as in call unsafeRun on calcETag or not have F on it? It was without F before because it was only ever called from inside a delay block.

core/src/main/scala/org/http4s/StaticFile.scala
eklavya: As I understand, the only think that matters is that there is some change, as long as that is maintained do we need to polish it more?

core/src/main/scala/org/http4s/StaticFile.scala
eklavya: It doesn't matter, if it's not a file the whole this is never called anyway. Nothing is added.

core/src/main/scala/org/http4s/StaticFile.scala
aeons: No. Instead of `.delay(...).flatMap(...)` just do `.delay(...)`

eklavya: oops, wrong place 👅

rossabaker:
Any other objections to this?  Will merge later today if not.

PR #1654, opened 2018-02-13
Expand the Authentication documentation

jcranky:


jcranky:
@jmcardon Agreed, thanks

PR #1660, opened 2018-02-15
Upgrade to blaze 0.14.x

bryce-anderson:
This depends on some unpublished changes in blaze (https://github.com/http4s/blaze/pull/173) and publishing those locally as `0.14.0-M2`.

For some reason I can't get the ALPN boot classes loaded, so I can't test the HTTP2 server. I'll look into it more later, but it would be nice to know if anybody else is having this issue.


blaze-core/src/main/scala/org/http4s/blazecore/websocket/Serializer.scala
```diff
@@ -0,0 +1,150 @@
...
+  ////////////////////////////////////////////////////////////////////////
+
+  private val _serializerWriteLock = new AnyRef
+  private var _serializerWriteQueue = new ArrayBuffer[I]
+  private var _serializerWritePromise: Promise[Unit] = null
+
+  ///  channel writing bits //////////////////////////////////////////////
+  override def channelWrite(data: I): Future[Unit] = _serializerWriteLock.synchronized {
+    if (maxWriteQueue > 0 && _serializerWriteQueue.length > maxWriteQueue) {
+      Future.failed(new Exception(s"$name Stage max write queue exceeded: $maxWriteQueue"))
```
rossabaker: How about an IllegalStateException?  Or even a Blaze-specific exception that extends it?

Do we want the exception to include the length and not just the cap?

blaze-core/src/main/scala/org/http4s/blazecore/websocket/Serializer.scala
```diff
@@ -0,0 +1,150 @@
...
+  def maxWriteQueue: Int = 0
+
+  ////////////////////////////////////////////////////////////////////////
+
+  private val _serializerWriteLock = new AnyRef
+  private var _serializerWriteQueue = new ArrayBuffer[I]
+  private var _serializerWritePromise: Promise[Unit] = null
+
+  ///  channel writing bits //////////////////////////////////////////////
+  override def channelWrite(data: I): Future[Unit] = _serializerWriteLock.synchronized {
```
rossabaker: Is this worth the code duplication vs. calling `channelWrite(data :: Nil)`?

blaze-core/src/main/scala/org/http4s/blazecore/websocket/Serializer.scala
```diff
@@ -0,0 +1,150 @@
...
+        f
+      }
+      else {
+        _serializerWriteQueue ++= data
+        _serializerWritePromise.future
+      }
+    }
+  }
+
+  // Needs to be in a synchronized because it is called from continuations
```
rossabaker: "be in a synchronized block" or "be synchronized"

blaze-core/src/main/scala/org/http4s/blazecore/websocket/Serializer.scala
```diff
@@ -0,0 +1,150 @@
...
+  private def _checkQueue(t: Try[Unit]): Unit = _serializerWriteLock.synchronized {
+    t match {
+      case f@ Failure(_) =>
+        _serializerWriteQueue.clear()
+        val p = _serializerWritePromise
```
rossabaker: Is this running afoul of your "don't complete promises while holding a lock" rule?

blaze-core/src/main/scala/org/http4s/blazecore/websocket/Serializer.scala
```diff
@@ -0,0 +1,150 @@
...
+      case Success(_) =>
+        if (_serializerWriteQueue.isEmpty) {
+          // Nobody has written anything
+          _serializerWritePromise = null
+        } else {
+          // stuff to write
+          val f = {
+            if (_serializerWriteQueue.length > 1) { // multiple messages, just give them the queue
+              val a = _serializerWriteQueue
+              _serializerWriteQueue = new ArrayBuffer[I](a.size + 10)
```
rossabaker: How does this growth interact with `maxWriteQueue`?

blaze-core/src/main/scala/org/http4s/blazecore/websocket/Serializer.scala
```diff
@@ -0,0 +1,150 @@
...
+import java.util.concurrent.atomic.{AtomicInteger, AtomicReference}
+
+import org.http4s.blaze.pipeline.TailStage
+
+import scala.concurrent.duration.Duration
+import scala.concurrent.{Future, Promise}
+import scala.collection.mutable.ArrayBuffer
+import org.http4s.blaze.util.Execution._
+
+import scala.util.{Failure, Success, Try}
```
rossabaker: We're generally sorting the imports alphabetically.  We got the idea from your employer. :laughing:

blaze-server/src/main/scala/org/http4s/server/blaze/WSFrameAggregator.scala
```diff
@@ -0,0 +1,94 @@
...
+
+    msgs.foldLeft(0) { (i, f) =>
+      System.arraycopy(f.data, 0, arr, i, f.data.length)
+      i + f.data.length
+    }
+
+    val msg = msgs.head match {
+      case _: Text => Text(arr)
+      case _: Binary => Binary(arr)
+      case f => sys.error(s"Shouldn't get here. Wrong type: $f")
```
rossabaker: There's an `org.http4s.util.bug` for these kind of assertion failures.

Are we missing an intermediate type in the `WebSocketFrame` hierarchy?

bryce-anderson:
Thanks for the review @rossabaker. I should have noted that this is in the 'proof of concept' stage and needs a fair amount of cleanup. After I get the blaze stuff sorted out and published as M2 I'll come back to this.

Do you have any clues on why I wouldn't be picking up the alpn classes when I run the `blaze-examples`? I use the `run` command, and it does fork, but unlike the blaze build I don't get the classes loaded. The only significant different I can think of (other than alpn-boot versions, which I also checked) is the sbt version.

blaze-core/src/main/scala/org/http4s/blazecore/util/Http2Writer.scala
```diff
@@ -40,7 +41,7 @@ private[http4s] class Http2Writer[F[_]](
         val hs = headers
         headers = null
         tail.channelWrite(
-          HeadersFrame(None, endStream = false, hs)
+          HeadersFrame(Priority.NoPriority, endStream = false, hs)
```
bryce-anderson: Honestly, I'm starting to think this `Priority.NoPriority` name was a huge mistake.

rossabaker: What would you call it?  `Unprioritized`?

rossabaker:
I don't have any theories on the ALPN issue yet.

blaze-core/src/main/scala/org/http4s/blazecore/util/Http2Writer.scala
bryce-anderson: That would be better. I was thinking `None` but maybe that's to close to `Option`?

bryce-anderson:
The ALPN issue was the coursier plugin changing the path of the jar so it no longer contained the magic string "jetty.alpn".

It turns out that the HTTP/2 support is broken on master due to that key `map -> flatMap` change in Http2NodeStage.scala, line 194. I suppose that means we also don't have tests for any of that.

aeons:
Don't have much to add, other than great to see this.
I'm still not very familiar with how blaze works or integrates with http4s, so I'll leave the reviewing to other people :)

bryce-anderson:
Am I supposed to open a new PR, or will this one be reopened at some point?

I think there are a few things still missing from this PR. I removed the aggregation limiting logic since it was unused by http4s, but perhaps that was a mistake since otherwise websockets are a potential denial of service attack vector. Unless someone thinks otherwise, I'll probably bring it back.

rossabaker:
GitHub support is supposed to run something that will try to bring it back.  If it's holding you back for it to be gone, reopen and link.

rossabaker:
I can't bring back the PR, but we definitely want this.  Please reopen, and then we'll kill my Label of Shame.

bryce-anderson:
I'll reopen it this evening. I'll also try to cut a new blaze milestone and get this on that.

PR #1661, opened 2018-02-15
 Bump fs2, fs2-reactive-streams, jawn-fs2, tomcat, kind-projector

rossabaker:


ChristopherDavenport:
This should target 0.18

rossabaker:
Oops. That was my intent.

PR #1662, opened 2018-02-15
Silence some imports in the json documentation

jcranky:


rossabaker:
Looks like a good candidate to retarget to 0.18.x, but :+1: to the change.

jcranky:
@rossabaker how does that work? are we merging 0.18.x into master then?

ChristopherDavenport:
Yes we run a merge 🚋

jcranky:
ah, ok - I'll re submit

PR #1665, opened 2018-02-15
Reintroduce the option for fallthrough for authenticated services

jmcardon:
While I do enjoy our default behavior stopping spidering, I realize that some people may want to consciously opt into not doing this.

This pr adds `withFallThrough` as a method for the old behavior, as well as `noSpider` for the current behavior, with the option of specifying a handler for authentication failures, so they are not pigeonholed into `401`.

On top of this, I added a note in the docs about the default behavior.

This seems to have been a pain point for many of our users. I hope this mitigates that.


docs/src/main/tut/auth.md
```diff
@@ -55,6 +55,13 @@ val middleware: AuthMiddleware[IO, User] =
   AuthMiddleware(authUser)
 ```
 
+Note: In the above, the default apply method of `AuthMiddleware` will consume all requests either unmatched, or
+not authenticated by returning an empty response with status code 401 (Unauthorized). This mitigates
+a kind of reconnaissance called "spidering", useful for white and black hat hackers to enumerate
+your api for possible unprotected points. To allow fallthrough as normal,
```
rossabaker: Most people think falling through is "normal."  I'd rephrase this.

jmcardon: Maybe just as "to allow fallthrough"?

rossabaker: :+1:

jmcardon:
```scala
[error] Failed tests:
[error] 	org.http4s.client.asynchttpclient.AsyncHttpClientSpec
```

oh boy

rossabaker:
I took a crack at the test failure in #1668.

PR #1668, opened 2018-02-17
Simplify renderResponse in the ClientRouteTestBattery

rossabaker:
Maybe it's writing it byte-by-byte that makes that ClientRouteTestBattery flaky.
Counterpoint: it only seems to hurt async-http-client
Countercounterpoint: either way, this is marginally less ugly.

### DIFF
--- a/client/src/test/scala/org/http4s/client/ClientRouteTestBattery.scala
+++ b/client/src/test/scala/org/http4s/client/ClientRouteTestBattery.scala
@@ -6,7 +6,6 @@ import cats.implicits._
 import fs2._
 import fs2.io._
 import java.net.InetSocketAddress
-import javax.servlet.ServletOutputStream
 import javax.servlet.http.{HttpServlet, HttpServletRequest, HttpServletResponse}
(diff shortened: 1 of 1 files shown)


client/src/test/scala/org/http4s/client/ClientRouteTestBattery.scala
```diff
@@ -123,14 +124,12 @@ abstract class ClientRouteTestBattery(name: String, client: Client[IO])
 
     val os: ServletOutputStream = srv.getOutputStream
 
-    val writeBody: IO[Unit] = resp.body
-      .evalMap { byte =>
-        IO(os.write(Array(byte)))
-      }
+    resp.body.through(writeOutputStream[IO](IO.pure(os)))
```
jmcardon: os.flush() will throw an exception I believe since `writeOutputStream` closes the output stream after use.

rossabaker: Nope.  But we shouldn't close what we don't open, and a servlet should flush when _it_ closes.  So I simplified it some more.

PR #1670, opened 2018-02-18
Reintroduce the option for fallthrough for authenticated services

jmcardon:
Reopening for the fallen comerade #1665

### DIFF
--- a/server/src/test/scala/org/http4s/server/middleware/authentication/AuthMiddlewareSpec.scala
+++ b/server/src/test/scala/org/http4s/server/middleware/authentication/AuthMiddlewareSpec.scala
@@ -193,6 +193,36 @@ class AuthMiddlewareSpec extends Http4sSpec {
         Unauthorized)
     }
 
+    "not consume the entire request when using fall through" in {
+
+      val authUser: Kleisli[OptionT[IO, ?], Request[IO], User] =
+        Kleisli.liftF(OptionT.none)
+
+      val authedService: AuthedService[User, IO] =
+        AuthedService {
+          case POST -> Root as _ => Ok()
+        }
+
+      val regularService: HttpService[IO] = HttpService[IO] {
+        case GET -> Root => Ok()
+      }
+
+      val middleware = AuthMiddleware.withFallThrough(authUser)
+
+      val service = middleware(authedService)
+
+      //Unauthenticated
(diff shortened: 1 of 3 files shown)


server/src/test/scala/org/http4s/server/middleware/authentication/AuthMiddlewareSpec.scala
```diff
@@ -193,6 +193,36 @@ class AuthMiddlewareSpec extends Http4sSpec {
...
+      val regularService: HttpService[IO] = HttpService[IO] {
+        case GET -> Root => Ok()
+      }
+
+      val middleware = AuthMiddleware.withFallThrough(authUser)
+
+      val service = middleware(authedService)
+
+      //Unauthenticated
+      (service <+> regularService).orNotFound(Request[IO](method = Method.POST)) must returnStatus(
```
jcranky: I guess I'm the only one that prefers `combineK` instead of `<+>`?

rossabaker: I don't have strong feelings.  Do you similarly prefer `combine` to `|+|`?.

jcranky: Yes, I usually prefer to avoid symbolic operators - it makes the life of beginners easier.

aeons: When you have only two arguments, I guess the non-symbolic names can be better. If you have more than two I definitely prefer the symbolic version. And since I like consistency, I prefer the symbolic version for two arguments as well.

jmcardon: The discussion but this feels sort of like a funny tangent given the scope of this pr (authentication). `<+>` is in more than just my tests and has been for months.

I'm just hoping that the `combineK` vs `<+>` syntax isn't the blocker for a relatively simple PR here.

jcranky: lol - it is totally a tangent, sorry for that, my bad :)

rossabaker: Well, I already approved it. Don't blame me! :laughing:

PR #1672, opened 2018-02-22
Add `Eq` instance for Uri

jcranky:


jmcardon:
+1 to a property test. We seem to occasionally run into URI unexpected behavior.

jcranky:
Will do. Are there some other tests like that that I should use as a base for writing this one?

rossabaker:
`checkAll("Eq[Uri]", EqTests[Uri].eqv)` should do it.

jcranky:
@rossabaker thanks! When I do that though, I get this:

```
could not find implicit value for parameter arbF: org.scalacheck.Arbitrary[org.http4s.Uri => org.http4s.Uri]
[error]       checkAll("Eq[Uri]", EqTests[Uri].eqv)
```

Is there an `Uri => Uri` arbitrary somewhere or should I write it? Looking at the code though, I couldn't find this kind of arbitrary for other `EqTests` cases. I guess this basically means that I don't understand what is going on :dagger:

rossabaker:
You need a Cogen[Uri] to derive the Arbitrary [Uri => Uri].

jcranky:
@rossabaker I added something that compiles and tests are green. I don't fully understand that though, is there some resource that I could read to find more about `Cogen`? thanks!

rossabaker:
I don't know of any good docs, but most look like this:
* For a case class, contramap the tupled `Cogen`
* Follow this example for a [sum type](https://github.com/rickynils/scalacheck/blob/master/src/main/scala/org/scalacheck/Cogen.scala#L105)

I haven't really needed anything else.

testing/src/main/scala/org/http4s/testing/ArbitraryInstances.scala
```diff
@@ -670,6 +671,9 @@ trait ArbitraryInstances {
     } yield Uri(scheme, authority, path, query, fragment)
   }
 
+  implicit val cogenUri: Cogen[Uri] =
+    Cogen[String].contramap(_.renderString)
```
rossabaker: This could dig into products and sums, but I'd like to refactor that structure anyway.  I think this is legit.

rossabaker:
Oops, this isn't backward compatible.  We need to put this on master, or stuff them into an object.  We've done that before with an `ArbitraryInstances` companion.

jcranky:
@rossabaker you mean the companion that is in the end of the file, empty? I can do that, or also move this against master, I don't mind either way.

rossabaker:
Yes, that companion an add things that aren't in the trait without breaking bincompat.  That's the only way we can get this into release-0.18.x.  Otherwise, I'm also fine retargeting master.  Your call.

jcranky:
@rossabaker I'll target master then, new PR coming a few minutes

PR #1673, opened 2018-02-22
Adding examples from the gvolpe/advanced-http4s examples repo.

gvolpe:
I added a new package `demo` in the `blaze` examples and created all the examples from https://github.com/gvolpe/advanced-http4s. Later on, I'd like to add some documentation with simple examples compiling with tut and pointing out to these examples.

### DIFF
--- a/examples/blaze/src/main/scala/com/example/http4s/blaze/demo/client/MultipartClient.scala
+++ b/examples/blaze/src/main/scala/com/example/http4s/blaze/demo/client/MultipartClient.scala
@@ -1,6 +1,9 @@
 package com.example.http4s.blaze.demo.client
 
+import java.net.URL
+
 import cats.effect.{Effect, IO}
+import cats.syntax.flatMap._
 import cats.syntax.functor._
 import com.example.http4s.blaze.demo.StreamUtils
 import fs2.StreamApp.ExitCode
@@ -18,18 +21,20 @@ class MultipartHttpClient[F[_]](implicit F: Effect[F], S: StreamUtils[F])
     extends StreamApp
     with Http4sClientDsl[F] {
 
-  private val rick = getClass.getResource("/beerbottle.png")
+  private val image: F[URL] = F.delay(getClass.getResource("/beerbottle.png"))
 
-  private val multipart = Multipart[F](
+  private def multipart(url: URL) = Multipart[F](
     Vector(
       Part.formData("name", "gvolpe"),
-      Part.fileData("rick", rick, `Content-Type`(MediaType.`image/png`))
+      Part.fileData("rick", url, `Content-Type`(MediaType.`image/png`))
     )
   )
 
   private val request =
-    POST(Uri.uri("http://localhost:8080/v1/multipart"), multipart)
-      .map(_.replaceAllHeaders(multipart.headers))
+    for {
+      body <- image.map(multipart)
+      req <- POST(Uri.uri("http://localhost:8080/v1/multipart"), body)
+    } yield req.replaceAllHeaders(body.headers)
 
   override def stream(args: List[String], requestShutdown: F[Unit]): Stream[F, ExitCode] =
     Scheduler(corePoolSize = 2).flatMap { implicit scheduler =>
--- a/examples/blaze/src/main/scala/com/example/http4s/blaze/demo/client/MultipartClient.scala
+++ b/examples/blaze/src/main/scala/com/example/http4s/blaze/demo/client/MultipartClient.scala
@@ -1,18 +1,20 @@
 package com.example.http4s.blaze.demo.client
 
 import cats.effect.{Effect, IO}
-import cats.syntax.flatMap._
 import cats.syntax.functor._
+import com.example.http4s.blaze.demo.StreamUtils
 import fs2.StreamApp.ExitCode
 import fs2.{Scheduler, Stream, StreamApp}
 import org.http4s.client.blaze.Http1Client
+import org.http4s.client.dsl.Http4sClientDsl
 import org.http4s.headers.`Content-Type`
+import org.http4s.Method._
 import org.http4s.multipart.{Multipart, Part}
-import org.http4s.{MediaType, Method, Request, Uri}
+import org.http4s.{MediaType, Uri}
 
 object MultipartClient extends MultipartHttpClient[IO]
 
-class MultipartHttpClient[F[_]](implicit F: Effect[F]) extends StreamApp {
+class MultipartHttpClient[F[_]](implicit F: Effect[F], S: StreamUtils[F]) extends StreamApp with Http4sClientDsl[F] {
 
   private val rick = getClass.getResource("/beerbottle.png")
 
@@ -24,19 +26,17 @@ class MultipartHttpClient[F[_]](implicit F: Effect[F]) extends StreamApp {
   )
 
   private val request =
-    Request[F](method = Method.POST, uri = Uri.uri("http://localhost:8080/v1/multipart"))
-      .withBody(multipart)
+    POST(Uri.uri("http://localhost:8080/v1/multipart"), multipart)
       .map(_.replaceAllHeaders(multipart.headers))
 
   override def stream(args: List[String], requestShutdown: F[Unit]): Stream[F, ExitCode] = {
     Scheduler(corePoolSize = 2).flatMap { implicit scheduler =>
-      Stream.eval(
-        for {
-          client  <- Http1Client[F]()
-          req     <- request
-          _       <- client.expect[String](req).map(println)
-        } yield ()
-      )
+      for {
+        client <- Http1Client.stream[F]()
+        req    <- Stream.eval(request)
+        value  <- Stream.eval(client.expect[String](req))
+        _      <- S.evalF(println(value))
+      } yield ()
     }.drain
   }
 
--- a/examples/blaze/src/main/scala/com/example/http4s/blaze/demo/client/MultipartClient.scala
+++ b/examples/blaze/src/main/scala/com/example/http4s/blaze/demo/client/MultipartClient.scala
@@ -1,0 +1,43 @@
+package com.example.http4s.blaze.demo.client
+
+import cats.effect.{Effect, IO}
+import cats.syntax.flatMap._
+import cats.syntax.functor._
+import fs2.StreamApp.ExitCode
+import fs2.{Scheduler, Stream, StreamApp}
+import org.http4s.client.blaze.Http1Client
+import org.http4s.headers.`Content-Type`
+import org.http4s.multipart.{Multipart, Part}
+import org.http4s.{MediaType, Method, Request, Uri}
+
+object MultipartClient extends MultipartHttpClient[IO]
+
+class MultipartHttpClient[F[_]](implicit F: Effect[F]) extends StreamApp {
+
+  private val rick = getClass.getResource("/beerbottle.png")
+
+  private val multipart = Multipart[F](
+    Vector(
+      Part.formData("name", "gvolpe"),
+      Part.fileData("rick", rick, `Content-Type`(MediaType.`image/png`))
+    )
+  )
+
+  private val request =
+    Request[F](method = Method.POST, uri = Uri.uri("http://localhost:8080/v1/multipart"))
+      .withBody(multipart)
+      .map(_.replaceAllHeaders(multipart.headers))
+
+  override def stream(args: List[String], requestShutdown: F[Unit]): Stream[F, ExitCode] = {
+    Scheduler(corePoolSize = 2).flatMap { implicit scheduler =>
+      Stream.eval(
+        for {
+          client  <- Http1Client[F]()
+          req     <- request
+          _       <- client.expect[String](req).map(println)
+        } yield ()
+      )
+    }.drain
+  }
+
+}
--- a/examples/blaze/src/main/scala/com/example/http4s/blaze/demo/client/MultipartClient.scala
+++ b/examples/blaze/src/main/scala/com/example/http4s/blaze/demo/client/MultipartClient.scala
@@ -14,7 +14,9 @@ import org.http4s.{MediaType, Uri}
 
 object MultipartClient extends MultipartHttpClient[IO]
 
-class MultipartHttpClient[F[_]](implicit F: Effect[F], S: StreamUtils[F]) extends StreamApp with Http4sClientDsl[F] {
+class MultipartHttpClient[F[_]](implicit F: Effect[F], S: StreamUtils[F])
+    extends StreamApp
+    with Http4sClientDsl[F] {
 
   private val rick = getClass.getResource("/beerbottle.png")
 
@@ -29,15 +31,14 @@ class MultipartHttpClient[F[_]](implicit F: Effect[F], S: StreamUtils[F]) extend
     POST(Uri.uri("http://localhost:8080/v1/multipart"), multipart)
       .map(_.replaceAllHeaders(multipart.headers))
 
-  override def stream(args: List[String], requestShutdown: F[Unit]): Stream[F, ExitCode] = {
+  override def stream(args: List[String], requestShutdown: F[Unit]): Stream[F, ExitCode] =
     Scheduler(corePoolSize = 2).flatMap { implicit scheduler =>
       for {
         client <- Http1Client.stream[F]()
-        req    <- Stream.eval(request)
-        value  <- Stream.eval(client.expect[String](req))
-        _      <- S.evalF(println(value))
+        req <- Stream.eval(request)
+        value <- Stream.eval(client.expect[String](req))
+        _ <- S.evalF(println(value))
       } yield ()
     }.drain
-  }
 
 }
--- a/examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/Module.scala
+++ b/examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/Module.scala
@@ -11,8 +11,8 @@ import org.http4s.client.Client
 import org.http4s.server.HttpMiddleware
 import org.http4s.server.middleware.{AutoSlash, ChunkAggregator, GZip, Timeout}
 
-import scala.concurrent.duration._
 import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.duration._
 
 class Module[F[_]](client: Client[F])(implicit F: Effect[F], S: Scheduler) {
 
--- a/examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/Module.scala
+++ b/examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/Module.scala
@@ -3,7 +3,10 @@ package com.example.http4s.blaze.demo.server
 import cats.effect.Effect
 import cats.syntax.semigroupk._ // For <+>
 import com.example.http4s.blaze.demo.server.endpoints._
-import com.example.http4s.blaze.demo.server.endpoints.auth.{BasicAuthHttpEndpoint, GitHubHttpEndpoint}
+import com.example.http4s.blaze.demo.server.endpoints.auth.{
+  BasicAuthHttpEndpoint,
+  GitHubHttpEndpoint
+}
 import com.example.http4s.blaze.demo.server.service.{FileService, GitHubService}
 import fs2.Scheduler
 import org.http4s.HttpService
@@ -20,9 +23,10 @@ class Module[F[_]](client: Client[F])(implicit F: Effect[F], S: Scheduler) {
 
   private val gitHubService = new GitHubService[F](client)
 
-  def middleware: HttpMiddleware[F] = {
-    {(service: HttpService[F]) => GZip(service)(F)} compose
-      { service => AutoSlash(service)(F) }
+  def middleware: HttpMiddleware[F] = { (service: HttpService[F]) =>
+    GZip(service)(F)
+  }.compose { service =>
+    AutoSlash(service)(F)
   }
 
   val fileHttpEndpoint: HttpService[F] =
@@ -58,8 +62,8 @@ class Module[F[_]](client: Client[F])(implicit F: Effect[F], S: Scheduler) {
   // You'll get 401 (Unauthorized) instead of 404 (Not found). Mount it separately as done in Server.
   val httpServices: HttpService[F] = (
     compressedEndpoints <+> timeoutEndpoints
-    <+> mediaHttpEndpoint <+> multipartHttpEndpoint
-    <+> gitHubHttpEndpoint
+      <+> mediaHttpEndpoint <+> multipartHttpEndpoint
+      <+> gitHubHttpEndpoint
   )
 
 }
--- a/examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/Module.scala
+++ b/examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/Module.scala
@@ -58,8 +58,6 @@ class Module[F[_]](client: Client[F])(implicit F: Effect[F], S: Scheduler) {
   val basicAuthHttpEndpoint: HttpService[F] =
     new BasicAuthHttpEndpoint[F].service
 
-  // NOTE: If you mix services wrapped in `AuthMiddleware[F, ?]` the entire namespace will be protected.
-  // You'll get 401 (Unauthorized) instead of 404 (Not found). Mount it separately as done in Server.
   val httpServices: HttpService[F] = (
     compressedEndpoints <+> timeoutEndpoints
       <+> mediaHttpEndpoint <+> multipartHttpEndpoint
--- a/examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/Module.scala
+++ b/examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/Module.scala
@@ -1,0 +1,65 @@
+package com.example.http4s.blaze.demo.server
+
+import cats.effect.Effect
+import cats.syntax.semigroupk._ // For <+>
+import com.example.http4s.blaze.demo.server.endpoints._
+import com.example.http4s.blaze.demo.server.endpoints.auth.{BasicAuthHttpEndpoint, GitHubHttpEndpoint}
+import com.example.http4s.blaze.demo.server.service.{FileService, GitHubService}
+import fs2.Scheduler
+import org.http4s.HttpService
+import org.http4s.client.Client
+import org.http4s.server.HttpMiddleware
+import org.http4s.server.middleware.{AutoSlash, ChunkAggregator, GZip, Timeout}
+
+import scala.concurrent.duration._
+import scala.concurrent.ExecutionContext.Implicits.global
+
+class Module[F[_]](client: Client[F])(implicit F: Effect[F], S: Scheduler) {
+
+  private val fileService = new FileService[F]
+
+  private val gitHubService = new GitHubService[F](client)
+
+  def middleware: HttpMiddleware[F] = {
+    {(service: HttpService[F]) => GZip(service)(F)} compose
+      { service => AutoSlash(service)(F) }
+  }
+
+  val fileHttpEndpoint: HttpService[F] =
+    new FileHttpEndpoint[F](fileService).service
+
+  val nonStreamFileHttpEndpoint = ChunkAggregator(fileHttpEndpoint)
+
+  private val hexNameHttpEndpoint: HttpService[F] =
+    new HexNameHttpEndpoint[F].service
+
+  private val compressedEndpoints: HttpService[F] =
+    middleware(hexNameHttpEndpoint)
+
+  private val timeoutHttpEndpoint: HttpService[F] =
+    new TimeoutHttpEndpoint[F].service
+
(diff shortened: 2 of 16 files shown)


examples/blaze/src/main/scala/com/example/http4s/blaze/demo/client/MultipartClient.scala
```diff
@@ -0,0 +1,43 @@
...
+
+  private val multipart = Multipart[F](
+    Vector(
+      Part.formData("name", "gvolpe"),
+      Part.fileData("rick", rick, `Content-Type`(MediaType.`image/png`))
+    )
+  )
+
+  private val request =
+    Request[F](method = Method.POST, uri = Uri.uri("http://localhost:8080/v1/multipart"))
```
rossabaker: Do we want to encourage people to use the `POST(uri, multipart)` syntax from `org.http4s.client.dsl`, or is that a failed experiment?

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/client/MultipartClient.scala
```diff
@@ -0,0 +1,43 @@
...
+      .withBody(multipart)
+      .map(_.replaceAllHeaders(multipart.headers))
+
+  override def stream(args: List[String], requestShutdown: F[Unit]): Stream[F, ExitCode] = {
+    Scheduler(corePoolSize = 2).flatMap { implicit scheduler =>
+      Stream.eval(
+        for {
+          client  <- Http1Client[F]()
+          req     <- request
+          _       <- client.expect[String](req).map(println)
```
rossabaker: Mapping a side effect teaches bad habits.  `.flatMap(s => IO(println(s))` is the right way to do this.  I wish it were less verbose.  Something like your StreamUtils.

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/client/MultipartClient.scala
```diff
@@ -0,0 +1,43 @@
...
+  private val request =
+    Request[F](method = Method.POST, uri = Uri.uri("http://localhost:8080/v1/multipart"))
+      .withBody(multipart)
+      .map(_.replaceAllHeaders(multipart.headers))
+
+  override def stream(args: List[String], requestShutdown: F[Unit]): Stream[F, ExitCode] = {
+    Scheduler(corePoolSize = 2).flatMap { implicit scheduler =>
+      Stream.eval(
+        for {
+          client  <- Http1Client[F]()
```
rossabaker: Is this simpler with `Http1Client.stream`?  I'm not sure.

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/Module.scala
```diff
@@ -0,0 +1,65 @@
...
+import com.example.http4s.blaze.demo.server.endpoints.auth.{BasicAuthHttpEndpoint, GitHubHttpEndpoint}
+import com.example.http4s.blaze.demo.server.service.{FileService, GitHubService}
+import fs2.Scheduler
+import org.http4s.HttpService
+import org.http4s.client.Client
+import org.http4s.server.HttpMiddleware
+import org.http4s.server.middleware.{AutoSlash, ChunkAggregator, GZip, Timeout}
+
+import scala.concurrent.duration._
+import scala.concurrent.ExecutionContext.Implicits.global
```
rossabaker: Most of our code is sorting imports lexicographically.

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/Module.scala
```diff
@@ -0,0 +1,65 @@
+package com.example.http4s.blaze.demo.server
+
+import cats.effect.Effect
+import cats.syntax.semigroupk._ // For <+>
```
rossabaker: I kind of like just importing `cats.implicits._` so new people don't stumble.  I haven't found a lot of drawbacks.

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/endpoints/JsonXmlHttpEndpoint.scala
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  def personXmlDecoder: EntityDecoder[F, Person] =
+    org.http4s.scalaxml.xml[F].map(Person.fromXml)
+
+  implicit def jsonXmlDecoder: EntityDecoder[F, Person] = jsonOf[F, Person] orElse personXmlDecoder
+
+  val service: HttpService[F] = HttpService {
+    case GET -> Root / ApiVersion / "media" =>
+      Ok("Send either json or xml via POST method. Eg: \n{\n  \"name\": \"gvolpe\",\n  \"age\": 30\n}\n or \n <person>\n  <name>gvolpe</name>\n  <age>30</age>\n</person>")
```
rossabaker: Oh, to be young again!

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/endpoints/JsonXmlHttpEndpoint.scala
```diff
@@ -0,0 +1,47 @@
...
+
+    case req @ POST -> Root / ApiVersion / "media" =>
+      req.decode[Person] { person =>
```
rossabaker: How many people are doing this vs. `req.as[Person].flatMap { person => ... }`?  There are two ways to do it, and I don't really know what's more popular in the wild.

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/endpoints/TimeoutHttpEndpoint.scala
```diff
@@ -0,0 +1,17 @@
...
+import org.http4s._
+import org.http4s.dsl.Http4sDsl
+
+import scala.util.Random
+
+class TimeoutHttpEndpoint[F[_]](implicit F: Sync[F]) extends Http4sDsl[F] {
+
+  val service: HttpService[F] = HttpService {
+    case GET -> Root / ApiVersion / "timeout" =>
+      F.delay(Thread.sleep(Random.nextInt(3) * 1000L)).flatMap(_ => Ok())
```
rossabaker: I think we should use an fs2 scheduler here.  Teach people Thread.sleep is never the answer.

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/endpoints/auth/GitHubHttpEndpoint.scala
```diff
@@ -0,0 +1,28 @@
...
+import com.example.http4s.blaze.demo.server.endpoints.ApiVersion
+import com.example.http4s.blaze.demo.server.service.GitHubService
+import org.http4s._
+import org.http4s.dsl.Http4sDsl
+
+class GitHubHttpEndpoint[F[_]](gitHubService: GitHubService[F])
+                              (implicit F: Sync[F]) extends Http4sDsl[F] {
+
+  object CodeQuery extends QueryParamDecoderMatcher[String]("code")
+  object StateQuery extends QueryParamDecoderMatcher[String]("state")
```
rossabaker: I have worked on http4s for a long time and almost never dealt with query parameters and I'm reminded every time that we need to make this better.  Did unfiltered have a better solution?  It's such a similar DSL.

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/service/GitHubService.scala
```diff
@@ -0,0 +1,55 @@
...
+import org.http4s.client.Client
+import org.http4s.client.dsl.Http4sClientDsl
+import org.http4s.{Header, Request, Uri}
+
+// See: https://developer.github.com/apps/building-oauth-apps/authorization-options-for-oauth-apps/#web-application-flow
+class GitHubService[F[_]: Sync](client: Client[F]) extends Http4sClientDsl[F] {
+
+  // NEVER make this data public! This is just a demo!
+  private val ClientId     = "959ea01cd3065cad274a"
+  private val ClientSecret = "53901db46451977e6331432faa2616ba24bc2550"
```
rossabaker: Whose data is this?

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/client/MultipartClient.scala
jmcardon: Agreed on the println map. `flatMap(m => F.delay(println(m)))` shows we do our part to account even for the smallest side effects.

you can also use

```
expected <- client.expect[String](req)
_ <- F.delay(println(expected)
```

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/Module.scala
jmcardon: This I can't say I agree 100% with, mostly because `cats.implicits._` slows down editors like intellij bringing in loads of syntax classes.

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/Module.scala
```diff
@@ -0,0 +1,65 @@
...
+  private val multipartHttpEndpoint: HttpService[F] =
+    new MultipartHttpEndpoint[F](fileService).service
+
+  private val gitHubHttpEndpoint: HttpService[F] =
+    new GitHubHttpEndpoint[F](gitHubService).service
+
+  val basicAuthHttpEndpoint: HttpService[F] =
+    new BasicAuthHttpEndpoint[F].service
+
+  // NOTE: If you mix services wrapped in `AuthMiddleware[F, ?]` the entire namespace will be protected.
```
jmcardon: Note: this is changing with #1665. I'm bringing back the possibility of fallthrough so while it's good to note this, 
there's also the fallthrough options now.

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/Server.scala
```diff
@@ -0,0 +1,30 @@
...
+import scala.concurrent.ExecutionContext.Implicits.global
+
+object Server extends HttpServer[IO]
+
+class HttpServer[F[_]](implicit F: Effect[F]) extends StreamApp[F] {
+
+  override def stream(args: List[String], requestShutdown: F[Unit]): Stream[F, ExitCode] =
+    Scheduler(corePoolSize = 2).flatMap { implicit scheduler =>
+      for {
+        client   <- Stream.eval(Http1Client[F]())
```
jmcardon: Http1Client.stream here. It ensures proper resource release.

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/endpoints/JsonXmlHttpEndpoint.scala
jmcardon: 30 is young, says the dinosaur

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/endpoints/MultipartHttpEndpoint.scala
```diff
@@ -0,0 +1,29 @@
...
+                                 (implicit F: Sync[F]) extends Http4sDsl[F] {
+
+  val service: HttpService[F] = HttpService {
+    case GET -> Root / ApiVersion / "multipart" =>
+      Ok("Send a file (image, sound, etc) via POST Method")
+
+    case req @ POST -> Root / ApiVersion / "multipart" =>
+      req.decodeWith(multipart[F], strict = true) { response =>
+        def filterFileTypes(part: Part[F]): Boolean = {
+          part.headers.exists(_.value.contains("filename"))
```
jmcardon: I thought we added a filename helper? if not then we should.

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/endpoints/TimeoutHttpEndpoint.scala
jmcardon: +1 on the scheduler.

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/service/FileService.scala
```diff
@@ -0,0 +1,51 @@
...
+        .intersperse("\n")
+    }
+  }
+
+  def store(part: Part[F]): Stream[F, Unit] = {
+    val os: F[OutputStream] =
+      for {
+        home      <- F.delay(sys.env.getOrElse("HOME", "/tmp"))
+        filename  <- F.delay(part.filename.getOrElse("sample"))
+        output    <- F.delay(new FileOutputStream(s"$home/$filename"))
```
jmcardon: In this case, I think `fs2.io.file` helpers are more brief than encouraging people to use output streams directly.

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/service/GitHubService.scala
```diff
@@ -0,0 +1,55 @@
...
+      .withQueryParam("client_id", ClientId)
+      .withQueryParam("redirect_uri", RedirectUri)
+      .withQueryParam("scopes", "public_repo")
+      .withQueryParam("state", "test_api")
+
+    client.streaming[Byte](Request[F](uri = uri))(_.body)
+  }
+
+  def accessToken(code: String, state: String): F[String] = {
+    val uri = Uri.uri("https://github.com")
```
jmcardon: ugh as much as I hate oauth2 I really think it's starting to possibly become a necessity to add a helper.

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/client/MultipartClient.scala
jmcardon: It should be used I think, because of the `bracket` assurance `.stream` provides

gvolpe:
Going through it now (while on the train), thanks for the reviews!

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/client/MultipartClient.scala
gvolpe: I think I'm used to using it this way but good point, I'll change it in favor of the dsl version.

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/client/MultipartClient.scala
gvolpe: Actually I'm pretty aware of this, don't know what happen here :smile: thanks for catching this up! Changing it soon.

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/Module.scala
gvolpe: I found that the community is a bit divided on this. I think it's great using `cats.implicits._` when you know what you're doing (and as Jose says if your team is not using Idea) but for the sake of the examples I think qualified imports are a good way to teach people thinking about what they're actually bringing into scope.

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/endpoints/JsonXmlHttpEndpoint.scala
gvolpe: As an old Http4s user I'm used to this. I think unification of these two methods would be great to avoid confusion.

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/endpoints/auth/GitHubHttpEndpoint.scala
gvolpe: I have no idea of what you're talking about. Can you expand on "unfiltered"?

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/service/GitHubService.scala
gvolpe: It's mine. An example GitHub app I just created only for this example, so it's okay to make it public.

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/endpoints/TimeoutHttpEndpoint.scala
gvolpe: I tried to make it work with Scheduler before but I remember something didn't work well that's why I used Thread.sleep but I'm aware of the misuse. So here I tried again an this is the result:

This is the shortest version that works for me (any less verbose version?):
```scala
val randomMillis = FiniteDuration(Random.nextInt(3) * 1000L, TimeUnit.MILLISECONDS)
val p = Scheduler(2).flatMap(_.delay(Stream.emit("").covary[F], randomMillis)).map(Ok(_))
p.compile.fold(NoContent())((_, y) => y).flatten
```

And I expected this one to work but it doesn't:

```scala
val randomMillis = FiniteDuration(Random.nextInt(3) * 1000L, TimeUnit.MILLISECONDS)
Ok(Scheduler(2).flatMap(_.delay(Stream.emit("").covary[F], randomMillis)))
```
Is this a bug or is it expected?

When I say it doesn't work I mean it doesn't give a timeout. It seems that the response body is made before passing through the Timeout Middleware.

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/Module.scala
gvolpe: Good to know! I guess we can leave it as is and modify when that PR is merged.

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/Server.scala
gvolpe: Sure, I wasn't aware of the stream bracketed version of the HttpClient. I'll use it everywhere now!

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/endpoints/MultipartHttpEndpoint.scala
gvolpe: Mmm link?

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/service/FileService.scala
gvolpe: Actually I'm using part of the `fs2.io` api below but I'm not sure this code can be replaced by any of the `fs2.io.file`. If you know how to I'd be glad to change this.

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/service/GitHubService.scala
gvolpe: Yeah unfortunately it's required by many companies so I think it'd be nice to have something out of the box like Silhouette provides for Play apps.

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/Module.scala
gvolpe: Doesn't scalafmt organize the imports on compile? I use NeoVim to write code and I need to fix the `organize imports` command that got broken after an update...

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/client/MultipartClient.scala
rossabaker: Any ideas how we can do better here?  The ergonomics of request construction are one of our weak points.

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/Module.scala
rossabaker: It sorts imports within a line, but it does not sort the import lines themselves.  (This operation can, sadly, change semantics in Scala.)

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/endpoints/JsonXmlHttpEndpoint.scala
rossabaker: `flatMap` seems natural.  I'm open to deprecation on the other if nobody can rise to a stirring defense of it.

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/endpoints/TimeoutHttpEndpoint.scala
rossabaker: That's going to start a scheduler on every request.  I think passing in a bracketed scheduler would make for a better example.

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/endpoints/auth/GitHubHttpEndpoint.scala
rossabaker: [Unfiltered](https://github.com/unfiltered/unfiltered) is one of the first Scala HTTP libraries to be built on extractors.

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/endpoints/TimeoutHttpEndpoint.scala
jmcardon: As ross said, that's literally creating a scheduler, which also creates a thread pool every time. 

Service can be redone to:

`def service(scheduler: Scheduler[F]): HttpService[F]`

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/endpoints/JsonXmlHttpEndpoint.scala
gvolpe: Fair enough. I'll change it in favor of `as`.

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/endpoints/TimeoutHttpEndpoint.scala
gvolpe: Good point. I'll rework this example :+1:

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/endpoints/auth/GitHubHttpEndpoint.scala
gvolpe: I'll look into it!

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/endpoints/TimeoutHttpEndpoint.scala
gvolpe: I guess this one is more acceptable?

```scala
val randomMillis = FiniteDuration(Random.nextInt(3) * 1000L, TimeUnit.MILLISECONDS)
S.sleep(randomMillis).compile.drain *> Ok("delayed response")
```

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/service/FileService.scala
gvolpe: Nevermind @jmcardon I found the clean way to do this, thanks for putting pressure on me :D

```scala
for {
      home      <- S.evalF(sys.env.getOrElse("HOME", "/tmp"))
      filename  <- S.evalF(part.filename.getOrElse("sample"))
      path      <- S.evalF(Paths.get(s"$home/$filename"))
      _         <- part.body to fs2.io.file.writeAll(path)
    } yield ()
```

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/Module.scala
rossabaker: This, from the [Cats docs](https://typelevel.org/cats/typeclasses/imports.html), is why I soured on a la carte.

> Note: Beware that if you import a type class instance or its syntax twice, you will receive conflicting implicits with a less than helpful error message. This usually happens when importing different type classes in the same hierarchy or when importing syntax enrichment for all type classes using cats.syntax.all._ or cats.implicits._ together with a more specific import like cats.syntax.option._ or cats.instances.either._. 

Intermediate FP folks start to mix and match and then the compiler gives them no help.  It's still not the first thing I look for when someone asks for help, even though it's been the answer so many times.

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/endpoints/TimeoutHttpEndpoint.scala
rossabaker: Compiles in my head:

```scala
S.effect.delay(Ok("delayed response"), randomMillis)
```

Since it's a `FiniteDuration`, it's misleading to call it `millis`.

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/service/FileService.scala
jmcardon: no problem, good job :+1:

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/Module.scala
gvolpe: I've seen they even mix `cats.implicits._` with `cats.syntax` and that's even worse. On this one I stick with teaching them to understand what they're actually bringing into scope.

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/endpoints/TimeoutHttpEndpoint.scala
gvolpe: Awesome, I just learned something new :)  Didn't know about the `effect` method on Scheduler.

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/Module.scala
SystemFw: >   think it's great using cats.implicits._ when you know what you're doing

fwiw, I find that it's the complete opposite. You can use specific imports if you know where things are. I've lost track how many teams my team had otherwise working code that wouldn't compile because of too many or too few imports.

I agree with 
> On this one I stick with teaching them to understand what they're actually bringing into scope.

But in the context of docs, you might not get such a chance

gvolpe: Okay, so this is the last piece missing of this PR. I don't mind using `cats.implicits._` instead of `cats.syntax` if that's what the majority wants. You already know my point of view. Flip a coin :D

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/endpoints/TimeoutHttpEndpoint.scala
SystemFw:
```scala
F.delay(Random.nextInt(3)).flatMap { d =>
    S.effect.delay(Ok(), d.seconds)
}
```
Compiles in my head.
That being said, using `Stream` the way you did can be interrupted while sleeping (doesn't matter in this case though)

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/Module.scala
jmcardon: It's merged now, so you can use `.withFallthrough` IIRC.

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/Module.scala
jmcardon: I think we're just being pedantic.

IMO it's ok :+1:

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/client/MultipartClient.scala
```diff
@@ -0,0 +1,44 @@
...
+import org.http4s.multipart.{Multipart, Part}
+import org.http4s.{MediaType, Uri}
+
+object MultipartClient extends MultipartHttpClient[IO]
+
+class MultipartHttpClient[F[_]](implicit F: Effect[F], S: StreamUtils[F])
+    extends StreamApp
+    with Http4sClientDsl[F] {
+
+  private val rick = getClass.getResource("/beerbottle.png")
```
jmcardon: I think this can return `null` btw. Also it's a side effect.

PR #1676, opened 2018-02-25
Break up Cookie into Request and Response

kevinmeredith:
This PR adds a `RequestCookie` and `ResponseCookie` per the discussion in https://gitter.im/http4s/http4s?at=5a90d82b0202dc012e79241a.

I opened this PR due to the following motivation:

> The only reason I ask is that, at work, I, having not used Cookie's in a real app, wrongly wrote code that checked on an HTTP Request's Cookie's cookie.isSecure && cookie.isHttpOnly. If an HTTP Request's Cookie only contained name=value pairs, then that would eliminate that type of error

per https://gitter.im/http4s/http4s?at=5a90d82b0202dc012e79241a.

It uses the `abstract class` approach that @jmcardon mentioned in https://gitter.im/http4s/http4s?at=5a90d8cd0202dc012e79266d.

### DIFF
--- a/core/src/main/scala/org/http4s/ResponseCookie.scala
+++ b/core/src/main/scala/org/http4s/ResponseCookie.scala
@@ -42,7 +42,8 @@ object RequestCookieJar {
       def result(): RequestCookieJar = new RequestCookieJar(Vector(coll.toSeq: _*))
     }
 
-  implicit def canBuildFrom: CanBuildFrom[TraversableOnce[RequestCookie], RequestCookie, RequestCookieJar] =
+  implicit def canBuildFrom
+    : CanBuildFrom[TraversableOnce[RequestCookie], RequestCookie, RequestCookieJar] =
     new CanBuildFrom[TraversableOnce[RequestCookie], RequestCookie, RequestCookieJar] {
       def apply(
           from: TraversableOnce[RequestCookie]
@@ -63,9 +64,11 @@ class RequestCookieJar(private val headers: Seq[RequestCookie])
   def get(key: String): Option[RequestCookie] = headers.find(_.name == key)
   def apply(key: String): RequestCookie = get(key).getOrElse(default(key))
   def contains(key: String): Boolean = headers.exists(_.name == key)
-  def getOrElse(key: String, default: => String): RequestCookie = get(key).getOrElse(RequestCookie(key, default))
+  def getOrElse(key: String, default: => String): RequestCookie =
+    get(key).getOrElse(RequestCookie(key, default))
   override def seq: RequestCookieJar = this
-  def default(key: String): RequestCookie = throw new NoSuchElementException("Can't find RequestCookie " + key)
+  def default(key: String): RequestCookie =
+    throw new NoSuchElementException("Can't find RequestCookie " + key)
 
   def keySet: Set[String] = headers.map(_.name).toSet
 
@@ -135,7 +138,7 @@ final case class RequestCookie(name: String, content: String) extends Renderable
   override lazy val renderString: String = super.renderString
 
   override def render(writer: Writer): writer.type = {
-    writer.append (name).append ('=').append (content)
+    writer.append(name).append('=').append(content)
     writer
   }
 }
--- a/core/src/main/scala/org/http4s/ResponseCookie.scala
+++ b/core/src/main/scala/org/http4s/ResponseCookie.scala
@@ -25,14 +25,14 @@ import scala.collection.generic.CanBuildFrom
 object RequestCookieJar {
   def empty: RequestCookieJar = new RequestCookieJar(Nil)
 
-  def apply(cookies: Cookie*): RequestCookieJar = (newBuilder ++= cookies).result()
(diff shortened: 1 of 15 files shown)


core/src/main/scala/org/http4s/ResponseCookie.scala
```diff
@@ -130,17 +130,30 @@ class RequestCookieJar(private val headers: Seq[Cookie])
 }
 
 // see http://tools.ietf.org/html/rfc6265
-final case class Cookie(
-    name: String,
-    content: String,
+sealed abstract class Cookie(val name: String, val content: String) extends Renderable
```
rossabaker: I think the file should still be called `Cookie.scala`.

core/src/main/scala/org/http4s/ResponseCookie.scala
```diff
@@ -63,7 +63,7 @@ class RequestCookieJar(private val headers: Seq[Cookie])
   def get(key: String): Option[Cookie] = headers.find(_.name == key)
   def apply(key: String): Cookie = get(key).getOrElse(default(key))
   def contains(key: String): Boolean = headers.exists(_.name == key)
-  def getOrElse(key: String, default: => String): Cookie = get(key).getOrElse(Cookie(key, default))
+  def getOrElse(key: String, default: => String): Cookie = get(key).getOrElse(RequestCookie(key, default))
```
rossabaker: Doesn't this whole Cookie jar pertain to `RequestCookie`s?

This cookie jar is full of unsafe methods, isn't particularly useful to a client maintaining an active conversation, etc.  I think it's a candidate for deprecation and/or serious improvement.  In any case, I would also move it to a `RequestCookieJar` file.

core/src/main/scala/org/http4s/ResponseCookie.scala
jmcardon: I'm personally more partial to:

```scala
sealed trait Cookie extends Renderable{
   def name: String
   def content: String
   ...
}
```
Mostly being convention I see a lot in the scala world, for what in this sense we're defining to be an interface specifically.

rossabaker: Where in the API would we continue to abstract over `Cookie`?  Do we even need a base at all?

kevinmeredith: I echo the questioning of the need to actually abstract over `Cookie`. Also, per https://www.reddit.com/r/scala/comments/3887az/everything_you_ever_wanted_to_know_about_sealed/crtvpm5/, it seems wrong to me, albeit allowed in Scala, to use an ADT's sub-type in a signature.

kevinmeredith: Please note that I respectfully question the need. I realize that I've made no commits to this code-base. I look forward to learning from maintainers and authors, yielding to them on this PR!

rossabaker: And I appreciate the help!

What I would do is see what doesn't compile if `Cookie` goes away.  Definitely in the `RequestCookieJar`, which would be for `RequestCookie`.  There might be a few more references in the tests.  If everything is obviously related to a `Request` or a `Response`, fix it.  If something serves in both roles, maybe then we have a reason for the base `Cookie` to remain.  But I doubt that this case exists.

jmcardon: To be fair @kevinmeredith, I agree with you, but I say merely for the sake of convention, approaching Cookie like an interaface rather than an `ADT` which we will exhaustively match on, given this is scala, not haskell, wherein the subtypes themselves are types.

If this were Haskell, there would be no inheritance whatsoever, simply two record types, as we don't want to work ever in terms of `Cookie`.

IMHO we should get rid of inheriting from a cookie trait altogether, given that whether it's an abstract class or a trait, we can still exhaustively pattern match on `Cookie` like if it was an ADT if we are ever presented with it, so it's of no real difference to use an abstract class over a trait in this case. 

I'm just being a pedant maybe,  but if we're going for this, I'm thinking we should stop people from creating signatures in terms of `Cookie` in the first place.

kevinmeredith:
Please excuse the `git push -f`. I almost always avoid re-basing on a PR with comments.

Anyway, I was able to run `sbt "; project tests; test` successfully on this branch.

kevinmeredith:
Putting back the `WIP` since I see `sbt ci` had a few failures. Sorry for the premature removal of `WIP`!

kevinmeredith:
Could you please take a look, @rossabaker or @jmcardon? 

Jose - I know that I didn't use the `trait Cookie` approach. Please let me you know if you disagree with the PR. Thanks!

kevinmeredith:
> Thank you very much for your contribution in the project 

Thank you and the maintainers for this excellent library!

kevinmeredith:
Please let me know if you're waiting on me to do anything more before merging! Since this is a breaking change, would you expect it to be part of v19 or 20?

Thanks!

jmcardon:
@kevinmeredith nope! All good, thank you!

Just waiting on someone to hit the green button is all. You got three :+1: from maintainers, so it's all good.

PR #1677, opened 2018-02-26
Add enough arbitraries to test SemigroupK[EntityDecoder[F, ?]]

rossabaker:
Continuation of #1488.  Adds the test for what I prematurely merged.

/cc @jadlr


core/src/main/scala/org/http4s/MediaType.scala
```diff
@@ -145,10 +142,15 @@ object MediaRange {
       override def render(writer: Writer, range: MediaRange): writer.type =
         range.render(writer)
 
-      override def eqv(x: MediaRange, y: MediaRange): Boolean =
-        x.equals(y)
+      override def compare(x: MediaRange, y: MediaRange): Int = {
+        def orderedSubtype(a: MediaRange) = a match {
+          case mt: MediaType => mt.subType
+          case _ => ""
+        }
```
rossabaker: The awkwardness of this flows from the awkwardness of the types.  We should do better here.  (Sealed concrete types? :confused:)

@cquiroz mentioned he's been thinking about this.

core/src/main/scala/org/http4s/MessageFailure.scala
```diff
@@ -70,6 +70,10 @@ object ParseResult {
   * the entity headers or with the entity itself.   */
 trait DecodeFailure extends MessageFailure
 
+object DecodeFailure {
+  implicit val http4sEqForDecodeFailure: Eq[DecodeFailure] = Eq.fromUniversalEquals
```
rossabaker: I don't feel great about this instance.

core/src/main/scala/org/http4s/MessageFailure.scala
```diff
@@ -96,7 +100,7 @@ final case class InvalidMessageBodyFailure(details: String, cause: Option[Throwa
       .withBody(s"The request body was invalid.")(F, EntityEncoder.stringEncoder[F])
 }
 
-/** Indicates that a [[Message]] came with no supported [[MediaType]]. */
+/** Indicates that Message]] came with no supported [[MediaType]]. */
```
rossabaker: Oops.  Will fix.

testing/src/main/scala/org/http4s/testing/ArbitraryInstances.scala
```diff
@@ -267,22 +267,36 @@ trait ArbitraryInstances {
...
-    }
+    Arbitrary(http4sGenMediaRange)
 
   implicit val http4sCogenMediaRange: Cogen[MediaRange] =
-    Cogen[(String, Map[String, String])].contramap(m => (m.mainType, m.extensions))
+    Cogen[(String, String, Map[String, String])].contramap { m =>
+      val effectiveSubtype = m match {
+        case mt: MediaType => mt.subType
+        case _ => "*"
+      }
```
rossabaker: More garbage from the bad structure of `MediaType` and `MediaRange`.

testing/src/main/scala/org/http4s/testing/ArbitraryInstances.scala
```diff
@@ -350,7 +364,7 @@ trait ArbitraryInstances {
   implicit val arbitraryContentLength: Arbitrary[`Content-Length`] =
     Arbitrary {
       for {
-        long <- arbitrary[Long] if long > 0L
+        long <- sized(n => choose(0L, n.toLong))
```
rossabaker: The filter could cause this generator to fail on lists of headers.  This fix should be backported to 0.18.

testing/src/main/scala/org/http4s/testing/ArbitraryInstances.scala
```diff
@@ -715,9 +729,100 @@ trait ArbitraryInstances {
...
+  ) =
+    Arbitrary(for {
+      f <- arbitrary[(Message[F], Boolean) => DecodeResult[F, A]]
+      mrs <- arbitrary[Set[MediaRange]]
+    } yield new EntityDecoder[F, A] {
+      def decode(msg: Message[F], strict: Boolean): DecodeResult[F, A] = f(msg, strict)
+      def consumes = mrs
+    })
+
+  implicit def cogenMessage[F[_]](implicit F: Effect[F], ec: TestContext): Cogen[Message[F]] =
```
rossabaker: Incomplete.  It doesn't account for the differences between requests and responses.

testing/src/main/scala/org/http4s/testing/ArbitraryInstances.scala
```diff
@@ -715,9 +729,100 @@ trait ArbitraryInstances {
...
+    Cogen[(Headers, EntityBody[F])].contramap(m => (m.headers, m.body))
+
+  implicit def cogenHeaders: Cogen[Headers] =
+    Cogen[List[Header]].contramap(_.toList)
+
+  implicit def cogenHeader: Cogen[Header] =
+    Cogen[(CaseInsensitiveString, String)].contramap(h => (h.name, h.value))
+
+  implicit def arbitraryDecodeFailure: Arbitrary[DecodeFailure] =
```
rossabaker: Incomplete.  `DecodeFailure` is an open hierarchy.  Which I dearly wish it weren't.  If we do complete this, then we have a problem in the Cogen instance.

testing/src/main/scala/org/http4s/testing/ArbitraryInstances.scala
```diff
@@ -715,9 +729,100 @@ trait ArbitraryInstances {
...
+        case MediaTypeMismatch(mt, e) => Cogen[(MediaType, Set[MediaRange])].perturb(seed, (mt, e))
+        // TODO What if it's not one of these?
+      }
+    }
+
+  private[http4s] implicit def arbitraryMessage[F[_]]: Arbitrary[Message[F]] =
+    // TODO this is bad because the underlying generators are bad
+    Arbitrary(Gen.oneOf(arbitrary[Request[F]], arbitrary[Response[F]]))
+
+  private[http4s] implicit def arbitraryRequest[F[_]]: Arbitrary[Request[F]] = Arbitrary {
```
rossabaker: This, and response, are something I've wanted for a long time, but they need to get a lot more sophisticated to make sure they're following the HTTP rules.

tests/src/test/scala/org/http4s/EntityDecoderSpec.scala
```diff
@@ -15,6 +21,10 @@ import scala.concurrent.ExecutionContext
 
 class EntityDecoderSpec extends Http4sSpec with PendingUntilFixed {
   implicit val executionContext: ExecutionContext = trampoline
+  implicit val testContext: TestContext = TestContext()
+
+  implicit def entityDecoderEq[A: Eq]: Eq[EntityDecoder[IO, A]] =
```
rossabaker: Maybe we should have an `eq` package in our testing package for things that are "eq by lack of a counterexample," like cats does.

tests/src/test/scala/org/http4s/EntityDecoderSpec.scala
```diff
@@ -439,4 +449,5 @@ class EntityDecoderSpec extends Http4sSpec with PendingUntilFixed {
     EntityEncoder.simple[IO, ErrorJson](`Content-Type`(MediaType.`application/json`))(json =>
       Chunk.bytes(json.value.getBytes()))
 
+  checkAll("SemigroupK[EntityDecoder[IO, ?]]", SemigroupKTests[EntityDecoder[IO, ?]].semigroupK[String])
```
rossabaker: The line that motivated this entire PR.  Sad: it takes 81 seconds on my machine.

tests/src/test/scala/org/http4s/MediaRangeSpec.scala
```diff
@@ -41,6 +42,6 @@ class MediaRangeSpec extends Http4sSpec {
     }
   }
 
-  checkAll("Eq[MediaRange]", EqTests[MediaRange].eqv)
+  checkAll("Order[MediaRange]", OrderTests[MediaRange].order)
```
rossabaker: Was necessary so we could get a Cogen for a set of them.  HTTP doesn't have anything to say about it, and maybe it should be demoted to a test package.

core/src/main/scala/org/http4s/MessageFailure.scala
jmcardon: Me neither, but then again I'm not exactly sure whether we care about matching too much whether two decodefailures match, vs simply knowing something failed in a specific way.

PR #1680, opened 2018-02-27
Prepare release v0.18.1

rossabaker:


jmcardon:
Just opened #1681 as a #1670 backport.

PR #1684, opened 2018-03-06
Prometheus metrics

ChristopherDavenport:
A layout of a prometheus metrics similar to dropwizard metrics, I believe my use of observe may be in error. Any guidance appreciated. 

We're also not following prometheus best practices that state metrics should not be based upon the response at all. With that it would be easy to implement an initial transformation to something that could be completed correctly in each case and be a little less dependent on responses.

### DIFF
--- a/prometheus-server-metrics/src/main/scala/org/http4s/server/prometheus/PrometheusExportService.scala
+++ b/prometheus-server-metrics/src/main/scala/org/http4s/server/prometheus/PrometheusExportService.scala
@@ -10,16 +10,21 @@ import io.prometheus.client.hotspot._
 import org.http4s.HttpService
 import org.http4s.dsl.Http4sDsl
 
+case class PrometheusExportService[F[_]](
+  service: HttpService[F],
+  cr: CollectorRegistry
+)
+
 object PrometheusExportService {
 
-  def setupMetricsService[F[_]](implicit F: Sync[F]): F[(HttpService[F], CollectorRegistry)] = {
+  def apply[F[_]](implicit F: Sync[F]): F[PrometheusExportService] = {
     for {
       cr <- F.delay(new CollectorRegistry())
       _ <- addDefaults(cr)(F)
-    } yield (service(cr), cr)
+    } yield PrometheusExportService(service(cr), cr)
   }
 
-  private def service[F[_]](collectorRegistry: CollectorRegistry)(implicit F: Sync[F]): HttpService[F] = {
+  def service[F[_]](collectorRegistry: CollectorRegistry)(implicit F: Sync[F]): HttpService[F] = {
     object dsl extends Http4sDsl[F]
     import dsl._
 
--- a/prometheus-server-metrics/src/main/scala/org/http4s/server/prometheus/PrometheusExportService.scala
+++ b/prometheus-server-metrics/src/main/scala/org/http4s/server/prometheus/PrometheusExportService.scala
@@ -36,11 +36,14 @@ object PrometheusExportService {
       _ <- addDefaults(cr)
     } yield new PrometheusExportService[F](service(cr), cr)
 
-  def generateResponse[F[_]: Sync](collectorRegistry: CollectorRegistry): F[Response[F]] = Sync[F].delay {
+  def generateResponse[F[_]: Sync](collectorRegistry: CollectorRegistry): F[Response[F]] =
+    Sync[F]
+      .delay {
         val writer = new StringWriter
         TextFormat.write004(writer, collectorRegistry.metricFamilySamples)
         writer.toString
-  }.map(Response[F](Status.Ok).withEntity(_))
+      }
+      .map(Response[F](Status.Ok).withEntity(_))
 
   def service[F[_]: Sync](collectorRegistry: CollectorRegistry): HttpService[F] = {
     object dsl extends Http4sDsl[F]
--- a/prometheus-server-metrics/src/main/scala/org/http4s/server/prometheus/PrometheusExportService.scala
+++ b/prometheus-server-metrics/src/main/scala/org/http4s/server/prometheus/PrometheusExportService.scala
@@ -7,35 +7,51 @@ import cats.implicits._
 import io.prometheus.client.CollectorRegistry
 import io.prometheus.client.exporter.common.TextFormat
 import io.prometheus.client.hotspot._
-import org.http4s.HttpService
+import org.http4s._
 import org.http4s.dsl.Http4sDsl
 
-case class PrometheusExportService[F[_]](
-    service: HttpService[F],
+class PrometheusExportService[F[_]: Sync] private (
+    s: HttpService[F],
     cr: CollectorRegistry
-)
+) {
+  def withCollectorRegistry(cr: CollectorRegistry): PrometheusExportService[F] =
+    new PrometheusExportService[F](
+      PrometheusExportService.service[F](cr),
+      cr
+    )
+
+  def service: HttpService[F] = s
+
+  def collectorRegistry: CollectorRegistry = cr
+}
 
 object PrometheusExportService {
 
-  def apply[F[_]: Sync]: F[PrometheusExportService[F]] =
+  def apply[F[_]: Sync](collectorRegistry: CollectorRegistry): PrometheusExportService[F] =
+    new PrometheusExportService(service(collectorRegistry), collectorRegistry)
+
+  def build[F[_]: Sync]: F[PrometheusExportService[F]] =
     for {
       cr <- new CollectorRegistry().pure[F]
       _ <- addDefaults(cr)
-    } yield PrometheusExportService[F](service(cr), cr)
+    } yield new PrometheusExportService[F](service(cr), cr)
+
+  def generateResponse[F[_]: Sync](collectorRegistry: CollectorRegistry): F[Response[F]] =
+    for {
+      bodyString <- Sync[F].delay {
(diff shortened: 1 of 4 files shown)


ChristopherDavenport:
If we like our current feature set, this now works in decent parity, as we use the same internal mechanisms as prometheus's `startTimer`/`observeDuration` but done manually.

prometheus-server-metrics/src/main/scala/org/http4s/server/prometheus/PrometheusMetrics.scala
```diff
@@ -0,0 +1,240 @@
...
+  private def responseHistogram(responseTimers: ResponseHistograms, status: Status): Histogram =
+    status.code match {
+      case hundreds if hundreds < 200 => responseTimers.resp1xx
+      case twohundreds if twohundreds < 300 => responseTimers.resp2xx
+      case threehundreds if threehundreds < 400 => responseTimers.resp3xx
+      case fourhundreds if fourhundreds < 500 => responseTimers.resp4xx
+      case _ => responseTimers.resp5xx
+    }
+
+  def apply[F[_]: Effect](
```
rossabaker: What needed `Effect` here?

prometheus-server-metrics/src/main/scala/org/http4s/server/prometheus/PrometheusExportService.scala
```diff
@@ -0,0 +1,51 @@
...
+case class PrometheusExportService[F[_]](
+    service: HttpService[F],
+    cr: CollectorRegistry
+)
+
+object PrometheusExportService {
+
+  def apply[F[_]](implicit F: Sync[F]): F[PrometheusExportService[F]] =
+    for {
+      cr <- F.delay(new CollectorRegistry())
```
rossabaker: Nit: I think this could be `pure`.

prometheus-server-metrics/src/main/scala/org/http4s/server/prometheus/PrometheusExportService.scala
```diff
@@ -0,0 +1,51 @@
...
+            val writer = new StringWriter
+            TextFormat.write004(writer, collectorRegistry.metricFamilySamples)
+            writer.toString
+          }
+          .flatMap(Ok(_))
+    }
+  }
+
+  def addDefaults[F[_]](cr: CollectorRegistry)(implicit F: Sync[F]): F[Unit] =
+    for {
```
rossabaker: Not going to be a hotspot, but I'd have probably done this in one big `delay` block.

prometheus-server-metrics/src/main/scala/org/http4s/server/prometheus/PrometheusMetrics.scala
```diff
@@ -0,0 +1,240 @@
...
+      prefix: String = "org_http4s_server",
+      c: CollectorRegistry): F[HttpMiddleware[F]] = Sync[F].delay {
```
rossabaker: How about `HttpService[F] => F[HttpService[F]]`?  Does the whole function need to be in an effect, or is just effectful function?  And then does `Kleisli[F, HttpService[F], HttpService[F]]` improve the ergonomics?  I think this is our first effectful middleware?

fiadliel:
I would agree with the best practices, in particular for histograms (would favour just one histogram for all requests, and counters for everything else).

The same HTTP method may be used in quite diverse code paths, which makes a histogram per-method quite muddy, and limited in usefulness. If you could do a 1-1 mapping between method and implementation, it would be fine, but that's not true in general. (I would be in favour of histograms for particular matches on a `HttpRequest` partial function, but this isn't really something you can accomplish with middleware).

When monitoring, I would also favour a single gauge with labels for sub-options, over having multiple gauges (e.g. `http_server_requests_by_method{method=GET}`, rather than `http_server_get_requests`. There are a number of advantages, including Grafana/etc. being able to automatically retrieve the TS data for all methods without manually calling them out. See https://prometheus.io/docs/practices/naming/#labels

ChristopherDavenport:
How does one add labels? Attempting to figure that out.

fiadliel:
https://github.com/prometheus/client_java#labels

```java
class YourClass {
  static final Counter requests = Counter.build()
     .name("my_library_requests_total").help("Total requests.")
     .labelNames("method").register();
  
  void processGetRequest() {
    requests.labels("get").inc();
    // Your code here.
  }
}
```

The label names are generally known ahead of time (e.g. "method", "code", etc.).

prometheus-server-metrics/src/main/scala/org/http4s/server/prometheus/PrometheusExportService.scala
```diff
@@ -0,0 +1,51 @@
...
+
+import cats.effect._
+import cats.implicits._
+import io.prometheus.client.CollectorRegistry
+import io.prometheus.client.exporter.common.TextFormat
+import io.prometheus.client.hotspot._
+import org.http4s.HttpService
+import org.http4s.dsl.Http4sDsl
+
+case class PrometheusExportService[F[_]](
```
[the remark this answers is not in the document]
ChristopherDavenport: Exporting as a tuple would do the same, how would you export these for best bincompat?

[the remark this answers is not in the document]
rossabaker: I'd build something similar to what [contraband does](https://www.scala-sbt.org/contraband/code-generation.html).  There's not really any need for equals or hashCode.

prometheus-server-metrics/src/main/scala/org/http4s/server/prometheus/PrometheusMetrics.scala
```diff
@@ -9,231 +9,229 @@ import org.http4s._
...
-      resp3xx =
-        Histogram.build().name(prefix + "_" + "3xx_responses").help("3xx Responses.").register(c),
-      resp4xx =
-        Histogram.build().name(prefix + "_" + "4xx_responses").help("4xx Responses.").register(c),
-      resp5xx =
-        Histogram.build().name(prefix + "_" + "5xx_responses").help("5xx Responses.").register(c)
-    )
+      emptyResponseHandler: Option[Status] = Status.NotFound.some,
+      errorResponseHandler: Throwable => Option[Status] = e => Status.InternalServerError.some
+  ): F[HttpMiddleware[F]] = Sync[F].delay {
```
rossabaker: I'm still not sure about `F[Midldeware]` as opposed to a `Kleisli`.

prometheus-server-metrics/src/main/scala/org/http4s/server/prometheus/PrometheusExportService.scala
```diff
@@ -0,0 +1,51 @@
...
+      cr <- F.delay(new CollectorRegistry())
+      _ <- addDefaults(cr)(F)
+    } yield PrometheusExportService[F](service(cr), cr)
+
+  def service[F[_]](collectorRegistry: CollectorRegistry)(implicit F: Sync[F]): HttpService[F] = {
+    object dsl extends Http4sDsl[F]
+    import dsl._
+
+    HttpService[F] {
+      case GET -> Root / "metrics" =>
```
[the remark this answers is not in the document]
fiadliel: It's configurable: https://prometheus.io/docs/prometheus/latest/configuration/configuration/#<scrape_config>

ChristopherDavenport:
Next Round of Improvements: 

1. `Kleisli[F, HttpService[F], HttpService[F]]` 
2. New Function: `CollectorRegistry => F[Response[F]]` which is then used to generate the default service endpoint.
3. `case class => class X private (...)`
4. Investigate into `Content-Type`, I believe you but didn't find anything on a quick look.

ChristopherDavenport:
https://prometheus.io/docs/instrumenting/exposition_formats/

prometheus-server-metrics/src/main/scala/org/http4s/server/prometheus/PrometheusMetrics.scala
```diff
@@ -200,38 +199,37 @@ object PrometheusMetrics {
...
-
-    { service: HttpService[F] =>
+  ): Kleisli[F, HttpService[F], HttpService[F]] = Kleisli { service: HttpService[F] =>
+    Sync[F].delay {
+      val serviceMetrics =
+        ServiceMetrics(
+          requestDuration = Histogram
+            .build()
+            .name(prefix + "_" + "response_duration_seconds")
+            .help("Response Duration")
```
rossabaker: Might want to say "in seconds"

fiadliel:
One the `Content-Type` question, it's preferably `text/plain; version=0.0.4`. `text/plain` is theoretically less safe (latest version), but the last spec update was nearly 4 years ago, so whatever.

Or `application/vnd.google.protobuf; proto=io.prometheus.client.MetricFamily; encoding=delimited` for the Protobuf format.

For the record, I found Protobuf encoding to be significantly faster/did a lot less work, but dragging around a protobuf dependency for that can drag you into dependency hell, and can certainly be added at another time if somebody cares to add it (with content negotiation).

Google are now promising better backwards compatibility with the protobuf runtime libraries than they were doing up to recently, but it was pretty nasty before 3.x.

ChristopherDavenport:
@fiadliel Does this look good to you? Really appreciate all of your input.

fiadliel:
Looks good 👍

PR #1685, opened 2018-03-06
Alternative prometheus impl

ChristopherDavenport:
Alternative Prometheus impl based on starting timers and then terminating them eventually.  Has less features at the cost of following better prometheus practices.

### DIFF
--- a/prometheus-server-metrics/src/main/scala/org/http4s/server/prometheus/PrometheusMetrics.scala
+++ b/prometheus-server-metrics/src/main/scala/org/http4s/server/prometheus/PrometheusMetrics.scala
@@ -9,12 +9,22 @@ import org.http4s._
 import org.http4s.server.HttpMiddleware
 
 object PrometheusMetrics {
+
+  private case class SingleMetricRun(
+    activeRequests: Gauge,
+    totalRequests: Histogram.Timer,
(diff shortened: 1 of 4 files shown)


prometheus-server-metrics/src/main/scala/org/http4s/server/prometheus/PrometheusMetrics.scala
```diff
@@ -0,0 +1,197 @@
...
+      respOpt <- metricsServiceHandler[F](singleMetricRun, responseAtt)
+    } yield respOpt
+  }
+
+  private def metricsServiceHandler[F[_]: Sync](
+      s: SingleMetricRun,
+      e: Either[Throwable, Option[Response[F]]]
+  ): F[Option[Response[F]]] =
+    e.fold(
+      e => onServiceError(s) *> Sync[F].raiseError[Option[Response[F]]](e),
```
jcranky: noob question, unrelated to this PR, sorry... what is `*>`?

aeons: Run both effect, but discard output of the left one. Same outcome as `>>`

ChristopherDavenport:
Closing in favor of #1684

PR #1686, opened 2018-03-06
Make code slightly more readable

jcranky:
Sorry for such PR, but I was having a hard time reading that code... :)


ChristopherDavenport:
Ugh, I  mean I guess scalafmt doesn't like it, so you'll have to play around to make it readable within acceptable scalafmt. Sorry. :(

jcranky:
@ChristopherDavenport :'( I'll have a look if this is easily fixable

rossabaker:
You can turn scalafmt off for a section, but this is a weird place to do it.

Conjecture: if a function requires newlines, maybe it's doing too much.

jcranky:
> Conjecture: if a function requires newlines, maybe it's doing too much.

Good point, I'll consider a different strategy :)

jcranky:
closing in favor of https://github.com/http4s/http4s/pull/1696

Issue #760, opened 2016-11-07
Support Link Headers


justjoheinz:
It would be nice to have better support for `Link` Headers as defined by https://tools.ietf.org/html/rfc5988

For starters a HeaderKey `Link` would be already nice.

cquiroz:
Should these update target master or cats?

rossabaker:
Anything that isn't cats/fs2 specific should target master.  Then it can be merged to cats so it benefits both versions.

Shortly after 0.16, cats->master and master->scalaz-stream, and we'll still target master and merges will flow the other direction.

ioganegambaputifonguser:
@justjoheinz 
https://github.com/http4s/http4s/pull/1687
does it make sense? 
actually it is my first PR here, could you please support me? thanks!

guizmaii:
Maybe we can close this issue now ?

Issue #1651, opened 2018-02-08
Uri JSON roundtrip failure


Unisay:
```scala
import _root_.io.circe.syntax._
import org.http4s.Uri
import org.scalacheck.Prop._
import org.specs2._
import org.http4s.circe._
import org.http4s.testing.ArbitraryInstances.arbitraryUri

class JsonSpec extends Specification with ScalaCheck { def is = s2"""
  Uri roundtrip                                            $testUriRoundtrip
  Uri equality                                              $testUriEquality
  """

  // This test fails:
  def testUriRoundtrip = forAll { uri: Uri => uri.asJson.as[Uri] must beRight(uri) }

  // This test is fine:
  def testUriEquality = forAll { uri: Uri => uri must_=== uri }
}
```

wedens:
From `UriSpec`:
```scala
    "round trip with toString" in forAll { uri: Uri =>
      Uri.fromString(uri.toString) must be(uri.toString)
    }.pendingUntilFixed
```

aeons:
The example URI on gitter `https://d@:1%44` can't be parsed as a `Uri`.

The Apache Commons `UrlValidator` can't validate it either. Another validator I found thinks the port is `1%44`. So maybe there's something wrong with the arbitrary for `Uri`?

Unisay:
If we are talking about `URI` **not** `URL` here, then we shouldn't refer to `URL` validators checking `URI`s, but rather to the [RFC 3986](https://tools.ietf.org/html/rfc3986#page-7).

aeons:
True. Accoding to https://tools.ietf.org/html/rfc3986#page-22, the port component should only consist of digits. I'm still not convinced that `https://d@:1%44` is a valid URI.

Unisay:
I digged down a bit, here is an example:
Arbitrarily generated Uri has no `UserInfo` and no `Host` but has `Port`
Parsed `Uri` took port value `:1` as a `UserInfo`.
![uriroundtrip](https://user-images.githubusercontent.com/1009751/36062957-712a8034-0e76-11e8-9a72-781f12a0ec81.png)

Unisay:
I can already see several problems with respect to the RFC3986 here. In short: ABNF doesn't capture all the constraints, some of them are given additionally as a plain-text description:

> If a URI contains an authority component, then the path component must either be empty or begin with a slash ("/") character. 

In the example above authority isn't empty - it has port - but path doesn't start from `/`

[...]

PR #1687, opened 2018-03-06
Link header

ioganegambaputifonguser:
Enhancemet #760  Add simple link header with URI-reference only for now. Add test with such header.


blaze-server/src/test/scala/org/http4s/server/blaze/Http1ServerStageSpec.scala
```diff
@@ -5,13 +5,15 @@ import cats.effect._
 import cats.implicits._
 import java.nio.ByteBuffer
 import java.nio.charset.StandardCharsets
+
```
aeons: Imports should be sorted alphabetically without any blank lines.

aeons:
This is a good start.

I don't know much about the Link header, but from skimming the [RFC linked to in the issue](https://tools.ietf.org/html/rfc5988#section-5), it seems the parser is a bit too simplistic, and will fail on most of the examples from the RFC.

blaze-server/src/test/scala/org/http4s/server/blaze/Http1ServerStageSpec.scala
ioganegambaputifonguser: got it
btw do you know why build fails with my `Link.scala` with the format issues i guess?
locally it passed with the `test:scalafmt`

aeons: `test:scalafmt` only formats test sources. You need to also run `scalafmt` to format main sources.

You can run `sbt ci` to do the same as what ci runs. It takes a while though.

ioganegambaputifonguser:
sure, I have to put WIP status 
Also I am going to add a few link params within the current PR, using the RFC what you mentioned

```
Link           = "Link" ":" #link-value
  link-value     = "<" URI-Reference ">" *( ";" link-param )
  link-param     = ( ( "rel" "=" relation-types )
                 | ( "anchor" "=" <"> URI-Reference <"> )
                 | ( "rev" "=" relation-types )
                 | ( "hreflang" "=" Language-Tag )
                 | ( "media" "=" ( MediaDesc | ( <"> MediaDesc <"> ) ) )
                 | ( "title" "=" quoted-string )
                 | ( "title*" "=" ext-value )
                 | ( "type" "=" ( media-type | quoted-mt ) )
                 | ( link-extension ) )
```

blaze-server/src/test/scala/org/http4s/server/blaze/Http1ServerStageSpec.scala
```diff
@@ -413,6 +413,20 @@ class Http1ServerStageSpec extends Http4sSpec {
...
+        val service = HttpService[IO] {
+          case req => IO.pure(Response(body = req.body).replaceAllHeaders(linkHeader))
+        }
+
+        val req = "GET /foo HTTP/1.1\r\n\r\n"
+
+        val buf = Await.result(runRequest(Seq(req), service), 5.seconds)
+        val (_, hdrs, _) = ResponseParser.apply(buf)
+        hdrs.find(_.name == Link.name) must beSome(linkHeader)
+      }
```
rossabaker: This header shouldn't really need to be tested in blaze.

If you can add an `Arbitrary[Link]` (I'll help if you can't), then you can follow a model more like the [other header specs](https://github.com/http4s/http4s/blob/master/tests/src/test/scala/org/http4s/headers/ContentLengthSpec.scala).  It'll be a simple test that covers more.

ioganegambaputifonguser:
@rossabaker somehow I missed a notification that there are some updates. So I am here and ready to finish with it. I added Arbitrary[Link] with the uri only support. But some of the tests fails. Could you please support a little with it? thanks!

rossabaker:
Sorry, now it was my turn to miss a notification.

This failure is not obvious to me, but I strongly suspect it's related to the problem described in #1651.  If a URI doesn't round trip through the parser as the same structure, then the header containing them will also have problems.  I think if we fix that, this will also begin to work.

Yet another reason we need to fix our `Uri` type. :(

booders-12:
_Sent from my Galaxy Tab A (2016) using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)

rossabaker:
Wow, this one fell off the radar.  I'm so sorry.

The proper fix depends on fixing our `Uri`, but we've been busy with fs2 and cats-effect.  Is this something that we could mark the tests pending and get the header?

PR #1688, opened 2018-03-06
Add Eq instance for Uri

jcranky:
Now targeting master.


jcranky:
Successor to https://github.com/http4s/http4s/pull/1672

PR #1690, opened 2018-03-07
More polymorphic middlewares

rossabaker:
Some of our middleware is overconstrained.  Many of the transformations are valid regardless of the effect type.  Others depend only on `MonoidK` rather than a specific monad transformer.  These types permit more reasoning about what happens in the middleware and are applicable to more intermediate shapes in the course of building up a service.


rossabaker:
Happening on #1693 instead.

Issue #1692, opened 2018-03-08
API change ideas, per gitter discussion


edmundnoble:
Discussion of these changes begins at https://gitter.im/http4s/http4s?at=5a9ef5196f8b4b9946f7deaa. It extends for a *while*, into the next day, so I'll try to paraphrase, starting with reasons.

### Ideas
- An original goal of http4s was to allow users to enjoy a separation between http4s-dsl and http4s-server, so that custom routing DSLs and code would be easy to use and could add features to http4s.
- I proposed an alternative routing DSL which is less expressive in exchange for being more easy to analyze and more performant.
- There are other routing DSL ideas loosely following Haskell's `servant`.

### Issues
- Routing as a concern is shared between http4s-server, http4s-dsl, and http4s-core. The only way to do routing is really with `HttpService[F]` and `<+>`; that's the only thing middleware supports, that's the only thing `mountService` will take in, and that's what's created by `HttpService.apply`. This means *all* services must have a) effectful routing and b) "unsealed" routing (to be later "capped" by adding 404's for failed matches).
- The above means alternative routing DSLs are unable to create "sealed" routes which *must* succeed and have that encoded in the type, or to use exclusively "non-effectful" routing (with `Option` outside the service effect `F`), and even if they did, middlewares would not be usable on them.
- Routing with `<+>` creates every effectful `F[Response]` in the entire `HttpService`, even when it knows that effect won't be executed because an earlier route was successfully matched. 300 routes, 300 unnecessary object creations.
- More severely, the way those effects are created (if you're using `HttpService.apply`, as most are) includes executing matcher code and then sequencing a resulting `Option`, which means that no matter which route is accessed the routing itself always takes worst-case time.

### Solutions (in decreasing order of controversy)
- `HttpService.apply` should suspend its matching inside an `F[_]: Sync`. It's a bit sneaky, but it's the best way to avoid re-matching on requests. We will still create the unnecessary effect objects, and benchmarking would be nice to establish the degree this helps (and could be made better).
- A new version of `mountService` which works on total services.
- More polymorphic middlewares that can work on different service types.
- Perhaps a rewrite of the old version of `mountService` to use the new version, so that routing is always handled by user code and never `http4s-server` or any backends.

jmcardon:
In regards to solutions:

- 1 should be implemented for now, on master, for the sake of an efficiency fix.
- 2 we need some notion of what services are total. It also seems not independent from numba 3, but both are good solutions.

[...]

PR #1693, opened 2018-03-08
Deprecate HttpService in favor of HttpApp and HttpRoutes

rossabaker:
This is worth considering a Scalafix for migrating `HttpService` to `HttpRoutes`, as this will affect almost every server app.  This is a step toward #1692.


SystemFw:
Oh wow, I had no idea you were so close to a PR (although I've just seen the `don't merge yet` :) )

rossabaker:
It is a feasibility analysis. :)

SystemFw:
`fromPartial` is not the same as the `routes` I was proposing, right?
It still gets you a total function in the end by capping with `404`?

rossabaker:
I introduced `Http.fromPartial` and `HttpPartial.apply`.  I'm finding them both useful, but am worried about confusion.  Suggestions welcome.

SystemFw:
> I introduced Http.fromPartial and HttpPartial.apply. I'm finding them both useful, but am worried about confusion. Suggestions welcome.

Radical suggestion: remove both.
`HttpPartial` has a `routes` method instead of `apply` (which is then the same as `Kleisli.apply`)
`HttpPartial` also has `toHttp` (ok, not so sure about the name, it might just be `orNotFound` rearing its head again) to convert it to an `Http`

rossabaker:
I like `HttpPartial` not being `apply`.  I'm unsure about `routes`, as case statements aren't the only way to create a partial function.  Although I suppose it's all routing.  Yeah, maybe `routes`.

aeons:
This looks good. What is the L in HttpL?

SystemFw:
@aeons the `L` stands for Layer. The reference is to the approach used in cats-mtl, which moves away from MonadTrans (relating an `MT[_[_], _]` and an `F[_]` to FunctorLayer, MonadLayer and so on (relating `F[_]` and `G[_]`).
The rationale was finding a way to link `OptionT[F, ?]` and `F` through cats-mtl. That being said, we probably think that you won't need that, and that a simple natural transformation `F ~> G` ought to be enough in all cases, so I don't know if the name is still correct or not. It's short and nice though.

aeons:
@SystemFw 👍

jcranky:
Wouldn't calling it just `Http` cause a bit of confusion in regards if it is a `http client` or a `http server` kind of thing? Just `Http` feels a bit to broad to me.

SystemFw:
> Http applies well to server and client

That's one of the nicest things about it actually

> Concept tax vs. import tax, I suppose.

Yeah :(

rossabaker:
I just posted a sketch of a `BlazeServer`.  It's a lot like the `BlazeBuilder`, but:

- It takes an `Http` as a an argument. Mounting multiple services is no longer supported: composition happens outside the server.
- No more attempt to extend a tangled OO mess of `*Support` traits to abstract over server builders.
- No more `KeyStoreBits`.  Real world SSL tends to be more complicated than that case class expresses.
- The design is such that we can add fields in the future without breaking binary compatibility.

Arguably the service could be the last argument, so the config happens in advance.

I cleaned out some redundant looking examples.  I don't know the right difference here between DRY and full examples so people can see the whole picture.  There's a lot of room for improvement here.

rossabaker:
Maybe the server refactor is pushing the scope on this PR too far (we'd want similar for the others), but chanigng the way services mount is a thrust of the ticket. Thoughts?

jmcardon:
@rossabaker I think the server refactor is pulling a 2-in-1.

I do this a lot in tsec and it makes maintainers hate me :(

jcranky:
+1 for more, smaller  PRs :D

rossabaker:
Okay, backing out the server refactor.

SystemFw:
Naming scheme suggestion:
```scala
type Http[F[_], G[_]] = Kleisli[F, Request[G], Response[G]]
type HttpService[F[_]] = Http[F, F]
type HttpRoutes[F[_]] = HttpService[OptionT[F, ?], F]
```

Pros:
- Probably the nicest names so far. I'm especially fond of HttpRoutes being much more descriptive than HttpPartial in terms of intent, and that we don't need any one letter abbreviations.
Cons:
- we keep the name HttpService but change its meaning, makes migration harder (possibly significantly harder)

rossabaker:
I think in the third one, you meant:
`type HttpRoutes[F[_]] = Http[OptionT[F, ?], F]`

I agree that these are the nicest names in a green field, but I'm having trouble mitigating or accepting that migration pain.

SystemFw:
Corrected, thanks.
Yes, I'm having trouble too. I'm just posting it so that we have a reference point of how much worse the names we actually pick are going to be.

rossabaker:
```
type Http[F[_]] = Kleisli[F, Request[F], Response[F]]
type HttpApp[F[_]] = Http[F, F]
type HttpOptional[F[_]] = Http[OptionT[F, ?, F]]
```

Rationale:
* No cryptic letter for `Http`
* `HttpApp` joins several of our conuterparts in other languages, including Rack (the most famous?) and WAI (our closest relative) in calling it an app/application.  It also avoids incompatibly redefining `HttpService`.
* I like `HttpRoutes` when thinking of construction from cases, but is something that serves files `HttpRoutes`?

rossabaker:
These are companion objects for the three traits.  Each contains the relevant methods from `Kleisli` companion (the omitted ones are more polymorphic than the aliases), plus a `routes` constructor for `HttpApp` and `HttpPartial`.

```scala
package org.http4s

import cats.Applicative
import cats.effect.Sync
import cats.implicits._

object Http {
  def apply[F[_], G[_]](run: Request[G] => F[Response[G]]): Http[F, G] =
    Kleisli(run)
  def liftF[F[_], G[_]](fr: F[Response[G]]): Http[F, G] =
    Kleisli.liftF(fr)
  def pure[F[_]: Applicative, G[_]](fr: Response[G]): Http[F, G] =
    Kleisli.pure(fr)
  def local[F[_], G[_]](f: Request[G] => Request[G])(fa: Http[F, G]): Http[F, G] =
    Kleisli.local(f)(fa)
}

object HttpApp {
  def apply[F[_]](run: Request[F] => F[Response[F]]): HttpApp[F] =
    Kleisli(run)
  def liftF[F[_]](fr: F[Response[F]]): HttpApp[F] =
    Kleisli.liftF(fr)
  def pure[F[_]: Applicative](fr: Response[F]): HttpApp[F] =
    Kleisli.pure(fr)
  def local[F[_]](f: Request[F] => Request[F])(fa: HttpApp[F]): HttpApp[F] =
    Kleisli.local[F, Response[F], Request[F]](f)(fa)
  def routes[F[_]](pf: PartialFunction[Request[F], F[Response[F]]], default: Response[F] = Response.notFound[F])(implicit F: Sync[F]): HttpApp[F] = {
    val pureDefault = default.pure[F]
    Kleisli(req => pf.applyOrElse(req, Function.const(pureDefault)))
  }
}

object HttpOptional {
  def apply[F[_]](run: Request[F] => OptionT[F, Response[F]]): HttpOptional[F] =
    Kleisli(run)
  def liftF[F[_]](fr: OptionT[F, Response[F]]): HttpOptional[F] =
    Kleisli.liftF(fr)
  def pure[F[_]](fr: Response[F])(implicit FO: Applicative[OptionT[F, ?]]): HttpOptional[F] =
    Kleisli.pure(fr)
  def local[F[_]](f: Request[F] => Request[F])(fa: HttpOptional[F]): HttpOptional[F] =
    Kleisli.local[OptionT[F, ?], Response[F], Request[F]](f)(fa)
  def routes[F[_]](pf: PartialFunction[Request[F], F[Response[F]]])(implicit F: Sync[F]): HttpOptional[F] =
    Kleisli(req => OptionT(F.delay(pf.lift(req).sequence).flatten))
}
```

rossabaker:
Changes:
* `HttpOptional` becomes `HttpRoutes`
* `HttpRoutes.of` to construct from `PartialFunction` and not contradict `Kliesli.apply`
* No more creating an `HttpApp` directly from a `PartialFunction`.

```scala
object Http {
  def apply[F[_], G[_]](run: Request[G] => F[Response[G]]): Http[F, G] =
    Kleisli(run)
  def liftF[F[_], G[_]](fr: F[Response[G]]): Http[F, G] =
    Kleisli.liftF(fr)
  def pure[F[_]: Applicative, G[_]](fr: Response[G]): Http[F, G] =
    Kleisli.pure(fr)
  def local[F[_], G[_]](f: Request[G] => Request[G])(fa: Http[F, G]): Http[F, G] =
    Kleisli.local(f)(fa)
}

object HttpApp {
  def apply[F[_]](run: Request[F] => F[Response[F]]): HttpApp[F] =
    Kleisli(run)
  def liftF[F[_]](fr: F[Response[F]]): HttpApp[F] =
    Kleisli.liftF(fr)
  def pure[F[_]: Applicative](fr: Response[F]): HttpApp[F] =
    Kleisli.pure(fr)
  def local[F[_]](f: Request[F] => Request[F])(fa: HttpApp[F]): HttpApp[F] =
    Kleisli.local[F, Response[F], Request[F]](f)(fa)
}

object HttpRoutes {
  def apply[F[_]](run: Request[F] => OptionT[F, Response[F]]): HttpOptional[F] =
    Kleisli(run)
  def liftF[F[_]](fr: OptionT[F, Response[F]]): HttpOptional[F] =
    Kleisli.liftF(fr)
  def pure[F[_]](fr: Response[F])(implicit FO: Applicative[OptionT[F, ?]]): HttpOptional[F] =
    Kleisli.pure(fr)
  def local[F[_]](f: Request[F] => Request[F])(fa: HttpOptional[F]): HttpOptional[F] =
    Kleisli.local[OptionT[F, ?], Response[F], Request[F]](f)(fa)

  def of[F[_]](pf: PartialFunction[Request[F], F[Response[F]]])(implicit F: Sync[F]): HttpOptional[F] =
    Kleisli(req => OptionT(F.delay(pf.lift(req).sequence).flatten))
}
```

SystemFw:
Looks great to me ;)

rossabaker:
Wiped this branch in favor of the new proposal.  I haven't resolved the deprecations yet, but this is the new types, with scaladoc.

aeons:
Interesting!

How do you go form an HttpOptional to an Http?

The file containing HttpRoutes is called HttpRoute.scala without the s :)

rossabaker:
I fixed the filename last night without pushing, but good eye. :)

routes.orNotFound syntax still works.  Can provide a function that gets there with a liftK, too.

aeons:
Great.

So something along this for a minimal example?
```scala
val routes = HttpRoutes.of {
  case GET -> Root => Ok("http4s")
}
val server = BlazeServer(routes.orNotFound)
```

rossabaker:
Yeah.  I'd like to do the `orNotFound` without requiring syntax. :thinking:

SystemFw:
@rossabaker If you mean being able to call `orNotFound` on an `HttpRoutes` in infix form (`myRoutes.orNotFound`), but _without_ requiring imports, how hacky do you want to get?
Because I know a way to do that.

rossabaker:
Tell us, and I'll opine on whether it's too hacky. :laughing:

SystemFw:
There you go sir.

```scala
object H {
  import cats._, data._, implicits._
  import cats.effect._

  def orNotFoundF[F[_]](app: HttpRoutes[F]): HttpApp[F] = ???

  trait Syntax {
    implicit class ONF[F[_]](routes: HttpRoutes[F]) {
      def orNotFound: HttpApp[F] = orNotFoundF(routes)
    }
  }

  case class Req[F[_]]()
  case class Res[F[_]]()
  object Res extends Syntax

  type Http[F[_], G[_]] = Kleisli[F, Req[G], Res[G]]
  type HttpApp[F[_]] = Http[F, F]
  type HttpRoutes[F[_]] = Http[OptionT[F, ?], F]

  def a: HttpRoutes[IO] = ???
  def b = a.orNotFound //compiles

}
```

rossabaker:
So it's a Syntax trait defined at the same level as the `Http` aliases.  Which would effectively be like moving it out of `org.http4s.syntax.kleisli._` where it is now (and so also `org.http4s.implicits._`), and into `org.http4s`.

SystemFw:
@rossabaker no, you can define that wherever you want, but you need to make the companion of either Request or Response extend it. It will then be in scope for any `Http*`

rossabaker:
Oh, I missed the subtlety of the `Response` companion.  Now I get it.  I'm not sure whether I like it, but now I get it. :)

SystemFw:
Well I did warn you :P

I think it's worth putting it out there, since it's the only trick I know to get what you want, but I can see the arguments against it as well

rossabaker:
I would say we could live with encouraging `org.http4s.implicits._`, but that's one more thing on top of `org.http4s.dsl_` and it's not obvious what comes from where.

rossabaker:
After a long night of benchmarking, we learned:
- The overhead of matching 10 routes to discard them approaches the overall "ping route" request time.  When the routing is complicated and the responses are simple, laziness can make a noticeable difference.
- Suspended `run`s are not measurably worse than strict `run`s for routing, even on the happy path.
- We expect most use of `SemigroupK` to be on `HttpRoutes`, but can't rule it out for other monad stacks.
- Suspension is transparent with a strictly created `Kleisli`.  It's an optimization that prevents the duplicate running of pure function application.

Conclusions:
- `HttpRoutes.of` will suspend its routing.
-  All `Http` aliases will suspend their `run` on `apply` and `local` for consistency.   These now all require a `Sync` instance, which is still weaker than the server or the client. 
- No need to move to a custom `Service` newtype.

rossabaker:
In scope:
- Deprecated `HttpService`
- Changed client to `toHttpApp` and `fromHttpApp`

Out of scope:
- Polymorphic middleware
- Refactoring `Client` into an `Http` (#1772)
- Mounting `HttpApp` or `Http[Stream[F, ?], F]` to servers
- Renaming `AuthService`.  Do we need an `AuthHttp`?
- Fuller refurbishing of tuts and examples until we settle the `HttpApp` vs. `Http[Stream[F, ?], ?]` debate

rossabaker:
The docs still need attention, but the build should pass, and others want to start building on top of this.

nafg:
Middleware docs page still references (Http)Service

PR #1694, opened 2018-03-08
EntityEncoder as a typeclass with a pure operation

jmcardon:
This PR tackles eneity encoders as pure operations.

`toEntity` becomes `Entity[F]` instead of `F[Entity[F]]` 

We could also tentatively consider `EntityEncoder[A]` with `toEntity[F[_]: Sync]: Entity[F]` given that you need at least `Sync` to run a stream. This universal quantification on the bound would also work, and it would remove the necessity of requiring more than one `F[_]` for an entitydecoder.

### DIFF
--- a/core/src/main/scala/org/http4s/Message.scala
+++ b/core/src/main/scala/org/http4s/Message.scala
@@ -55,23 +55,20 @@ sealed trait Message[F[_]] extends MessageOps[F] { self =>
     * @tparam T type of the Body
     * @return a new message with the new body
     */
-  def withBody[T](b: T)(implicit F: Monad[F], w: EntityEncoder[F, T]): F[Self] =
-    w.toEntity(b).flatMap { entity =>
-      val hs = entity.length match {
-        case Some(l) =>
-          `Content-Length`
-            .fromLong(l)
-            .fold(
-              _ =>
-                F.pure {
-                  Message.logger.warn(s"Attempt to provide a negative content length of $l")
-                  w.headers.toList
-              },
-              cl => F.pure(cl :: w.headers.toList))
-        case None => F.pure(w.headers)
-      }
-      hs.map(newHeaders => change(body = entity.body, headers = headers ++ newHeaders))
+  def withBody[T](b: T)(implicit w: EntityEncoder[F, T]): Self = {
+    val entity = w.toEntity(b)
+    val hs = entity.length match {
+      case Some(l) =>
+        `Content-Length`
+          .fromLong(l)
+          .fold(_ => {
+            Message.logger.warn(s"Attempt to provide a negative content length of $l")
+            w.headers.toList
+          }, cl => cl :: w.headers.toList)
+      case None => w.headers
     }
+    change(body = entity.body, headers = headers ++ hs)
+  }
 
   /** Sets the entity body without affecting headers such as `Transfer-Encoding`
     * or `Content-Length`. Most use cases are better served by [[withBody]],
@@ -378,5 +375,5 @@ object Response {
 
   def notFoundFor[F[_]: Monad](request: Request[F])(
       implicit encoder: EntityEncoder[F, String]): F[Response[F]] =
-    Response(Status.NotFound).withBody(s"${request.pathInfo} not found")
+    Monad[F].pure(Response(Status.NotFound).withBody(s"${request.pathInfo} not found"))
 }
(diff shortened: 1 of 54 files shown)


core/src/main/scala/org/http4s/Message.scala
```diff
@@ -55,23 +55,20 @@ sealed trait Message[F[_]] extends MessageOps[F] { self =>
...
-              _ =>
-                F.pure {
-                  Message.logger.warn(s"Attempt to provide a negative content length of $l")
-                  w.headers.toList
-              },
-              cl => F.pure(cl :: w.headers.toList))
-        case None => F.pure(w.headers)
-      }
-      hs.map(newHeaders => change(body = entity.body, headers = headers ++ newHeaders))
+  def withBody[T](b: T)(implicit w: EntityEncoder[F, T]): Self = {
```
rossabaker: This breaks a lot of things.  I wonder about deprecating this and introducing a new name for the pure one.  How about `withEntity`?

core/src/main/scala/org/http4s/multipart/MultipartEncoder.scala
```diff
@@ -10,8 +10,9 @@ private[http4s] class MultipartEncoder[F[_]: Sync] extends EntityEncoder[F, Mult
   //TODO: Refactor encoders to create headers dependent on value.
   def headers: Headers = Headers.empty
 
-  def toEntity(mp: Multipart[F]): F[Entity[F]] =
-    Sync[F].delay(Entity(renderParts(mp.boundary)(mp.parts), None))
+  //Note: Is this actually effecting?
```
rossabaker: I don't think so.

docs/src/main/tut/dsl.md
```diff
@@ -214,25 +214,26 @@ import scala.concurrent.Future
 import scala.concurrent.ExecutionContext.Implicits.global
 ```
 
-You can seamlessly respond with a `Future` of any type that has an
-`EntityEncoder`.
+You can respond with a `Future` of any type that has an
+`EntityEncoder` by lifting it into IO. Note: `Future` is a side effecting and
```
rossabaker: Pedantry: future is pure.  The problems are that it's strict (so creating a Future with an effect is an effect) and that mapping over it is an effect.

docs/src/main/tut/dsl.md
```diff
@@ -214,25 +214,26 @@ import scala.concurrent.Future
...
+})).flatMap(Ok(_))
 io.unsafeRunSync
 ```
 
 As good functional programmers who like to delay our side effects, we
 of course prefer to operate in [F]s:
 
 ```tut
-val io = Ok(IO {
+val io = IO {
```
rossabaker: Would it be feasible to allow the old syntax with deprecation in http4s DSL?  It would require something like:

```
@deprecated("flatMap that shit", "0.19")
def apply[F[_]](a: F[A])(implicit EE: EntityEncoder[F, A])
```

This overload is probably not going to fly with the existing one, but it's worth a shot if we can do compiler-assisted migration.  This is a big breaking change.

Alternatively, it would be neat if we could target and scalafix the ones we ate.

docs/src/main/tut/dsl.md
SystemFw: `s/effect/side effect`.
You are correct, but how many people do you know that use Future with pure code inside? (I know of 0, they introduce Future to deal with side effects)

rossabaker: I know of 0, but I know of non-zero who point this out every time someone calls `Future` impure on Twitter. :)

SystemFw: But anyway you're right, this could probably we reworded like:
```
Note: unlike IO, wrapping a side effect in Future does not suspend it, and the resulting expression would still be side effectful, unless we wrap it in IO
```

core/src/main/scala/org/http4s/Message.scala
jmcardon: I wish I would've had that idea before I changed like 200 lines that depended on it :(

core/src/main/scala/org/http4s/multipart/MultipartEncoder.scala
jmcardon: Cool. That's what I thought, so I can alter the bound

core/src/main/scala/org/http4s/Message.scala
jmcardon: jk IJ to the rescue :D

SystemFw:
I don't want to dwell on this too long, but I think we should _consider_ keeping the name `withBody` + a  scalafix, instead of introducing `withEntity` + deprecation.

As a general position, I like deprecation better. But `withEntity` is a tiny bit obscure to the casual user, and `withBody` isn't. So which name do we want long term?

PR #1697, opened 2018-03-09
Qualify reference to identity in uriLiteral macro

edmundnoble:
Custom predefs mean qualifying `identity` is a better choice here. Might also be okay to use `x => x` but I don't know much about macros.


rossabaker:
Cherry-picked into #1700.  Will be released in 0.18.2.

PR #1698, opened 2018-03-09
Make retry use the correct duration units

edmundnoble:
Used to convert everything to seconds wrong (as in, `5.millis` => `5.seconds`), now just does the entire `max` computation inside `Duration`.


rossabaker:
:+1: I'll cherry pick this onto release-0.18.x. The macro fix can be, too.

rossabaker:
Cherry-picked into #1700.  Will be released in v0.18.2.

PR #1699, opened 2018-03-10
WIP - Add tests docs

kevinmeredith:


docs/src/main/tut/testing.md
```diff
@@ -0,0 +1,89 @@
+---
+menu: main
+weight: 350
+title: Testing
+---
+
+## Introduction
+
+Testing http4s [[org.http4s.HttpService]]'s is straightforward given that it's:
```
rossabaker: Does this `[[ ]]` syntax link to scaladoc, or did you mean backticks?

docs/src/main/tut/testing.md
```diff
@@ -0,0 +1,89 @@
...
+
+```scala
+    type HttpService[F[_]] = Kleisli[OptionT[F, ?], Request[F], Response[F]]
```
rossabaker: This may be getting tweaked soon (see discussion on #1692), but is true for 0.18.x.  This PR might be better targeted on top of that branch.

kevinmeredith: Thanks - will make sure I open the PR off the right branch!

docs/src/main/tut/testing.md
```diff
@@ -0,0 +1,139 @@
...
+```
+
+The documentation explains:
+
+```
+  /**
+    * A [[Kleisli]] that produces an effect to compute an [[OptionT[F]]]] from a
```
jcranky: I'm slightly worried about quoting the scaladoc like that because this will get out-dated quite easily .

docs/src/main/tut/testing.md
```diff
@@ -0,0 +1,139 @@
...
+final case class User(name: String, age: Int)
+object User {
+  implicit val UserEncoder: Encoder[User] = deriveEncoder[User]
+}
+
+trait UserRepo {
+  def find(userId: UUID): IO[Option[User]]
+}
+
+def service(repo: UserRepo): HttpService = HttpService[IO] {
```
jcranky: Should we hardcode `IO` here? I would suggest specifying `IO` only at the `test` part of the example.

jcranky: The same goes for the `UserRepo` above.

jcranky:
Nice :)
Maybe it is worth being more explicit where in the examples we have a sample test, and where it is supposed to be actual test code for the user's test framework of choise?

docs/src/main/tut/testing.md
jmcardon: Heh you're not wrong. #1693 is already changing this.

docs/src/main/tut/testing.md
jmcardon: Hmm, I do like the idea of being polymorphic on `F[_]: Effect` or `F[_]: Sync`.

kevinmeredith: Good suggestion!

kevinmeredith:
Thanks for the feedback, @jcranky! I'm on the fence about whether to actually show example specs in, let's say, ScalaTest and Specs2, or just show a custom `def compare(actual: IO[Response], expectedStatus: Status, expectedBody: Json)`.

If the reader knows how to call that `compare` function, then surely he/she can apply it to a specs2 or ScalaTest test, no?

Please let me know your thoughts.

kevinmeredith:
I'm going to close this PR, then open a new PR against the v 18 branch.

PR #1700, opened 2018-03-10
Release v0.18.2

rossabaker:
Cherry picks #1697 and #1698 and bumps Tomcat.


jmcardon:
#1701 could prove interesting if I can port it without breaking mima.

What do you guys think?

rossabaker:
#1701 is a pretty big change without looking at performance, and patch versions are cheap.  So I think maybe, but let's get 0.18.2 out first.

jmcardon:
yeah, sorry I don't even know why I proposed that at 6am after a binge code trying to get it compiling. 

Disregard that :smile:

PR #1701, opened 2018-03-10
Remove scodec bits + Streaming multipart

jmcardon:
:D


server/src/main/scala/org/http4s/server/middleware/Logger.scala
```diff
@@ -44,7 +43,10 @@ object Logger {
     val bodyStream = if (logBody && isText) {
       message.bodyAsText(charset.getOrElse(Charset.`UTF-8`))
     } else if (logBody) {
-      message.body.map(ByteVector.fromByte).map(_.toHex)
+      message.body
+        .fold(new StringBuilder)((sb, b) =>
+          sb.append(String.format("%02X", new java.lang.Integer(b & 0xff))))
```
jmcardon: This seems like the only thing making travis complain.

core/src/main/scala/org/http4s/multipart/MultipartParser.scala
```diff
@@ -4,153 +4,498 @@ package multipart
...
+      }
+      i += 1
+    }
+    (sti, i)
+  }
 
-    Stream
-      .eval(listT)
-      .flatMap(list => Stream.emits(list))
+  //Todo: This may fail for an edge case
```
ChristopherDavenport: Will likely need more specifics about this.

jmcardon: `stateIndex`  takes in as input a byte sequence (in an array), a chunk (which we can treat like an array, since we iterate through it) and a `state`, being an index to possibly the last place when this function was called. It returns the next index to check of the byte sequence to check, as well as the last place in the chunk where it ended.. A concrete example is, say we have two chunks like this, and the following byte sequence:
```scala
c1 = Chunk('a', 'b', 'c', 'c')
c2 = Chunk('d', 'e', 'f', 'z')

sequence = Array('c', 'd', 'e')
stateIndex(sequence, c1, 0) == (1, 4)
stateIndex(sequence, c2, 1) == (3, 2)
```

This is fine and dandy for sequences that perfectly align, but if we had:
```s
c1 = Chunk('a', 'b', 'c', 'c')
c2 = Chunk('a', 'c', 'd', 'e')

sequence = Array('c', 'd', 'e')
stateIndex(sequence, c1, 0) == (1, 4)
stateIndex(sequence, c2, 1) == (3, 4)
```

The problem here being that `splitAndIgnorePrev`, in the second case:
```scala
else if (i == c.size) {
    (sti + state, Stream.empty)
````

so for our hypothetical chunk, it considers the _entire_ chunk as part of the sequence to ignore, but only the last 3 bytes were part of it

So I'm considering that this function, in a very specific scenario can fail, but I haven't been able to make the perfect storm happen.

I need to replace this function specifically, but otherwise everything else about this is ok.

File: core/src/main/scala/org/http4s/util/package.scala [hunk 3]
```diff
@@ -1,10 +1,11 @@
 package org.http4s
 
+import java.nio.charset.StandardCharsets
```
rossabaker: IntelliJ keeps messing up my meticulously alphabetically sorted imports.

core/src/main/scala/org/http4s/util/package.scala
```diff
@@ -23,17 +24,43 @@ package object util {
...
-          val nextByteVector = ByteVector.view(byteBuffer.slice)
-          val nextStream = stream.consChunk(ByteVectorChunk(nextByteVector))
+          val nextStream = stream.consChunk(Chunk.byteBuffer(byteBuffer.slice()))
           Pull.output1(charBuffer.flip().toString).as(Some(nextStream))
       }
     }
   }
 
+  /** Converts ASCII encoded byte stream to a stream of `String`. */
+  def asciiDecode[F[_]]: Pipe[F, Byte, String] =
```
rossabaker: Can we make these private and then contribute them upstream?  They have nothing to do with us, and I don't want to break compatibility to get rid of them.

server/src/main/scala/org/http4s/server/middleware/Logger.scala
```diff
@@ -44,7 +43,9 @@ object Logger {
     val bodyStream = if (logBody && isText) {
       message.bodyAsText(charset.getOrElse(Charset.`UTF-8`))
     } else if (logBody) {
-      message.body.map(ByteVector.fromByte).map(_.toHex)
+      message.body
+        .fold(new StringBuilder)((sb, b) => sb.append(String.format("%02X", Int.box(b & 0xff))))
```
rossabaker: This `String.format` isn't dreadfully slow?

jmcardon: I think it is. Good catch. I can use a better encoder

jmcardon:
Running into an issue testing this first.

PR #1707, opened 2018-03-14
Dependency upgrades for 0.18.3

rossabaker:


project/Http4sPlugin.scala
```diff
@@ -251,36 +251,36 @@ object Http4sPlugin extends AutoPlugin {
 
   lazy val alpnBoot                         = "org.mortbay.jetty.alpn" %  "alpn-boot"                 % "8.1.12.v20180117"
   lazy val argonaut                         = "io.argonaut"            %% "argonaut"                  % "6.2.1"
-  lazy val asyncHttpClient                  = "org.asynchttpclient"    %  "async-http-client"         % "2.0.38"
+  lazy val asyncHttpClient                  = "org.asynchttpclient"    %  "async-http-client"         % "2.0.39"
```
rossabaker: Not the latest, but the latest binary compatible.

project/Http4sPlugin.scala
```diff
@@ -297,5 +297,5 @@ object Http4sPlugin extends AutoPlugin {
   lazy val specs2Scalacheck                 = "org.specs2"             %% "specs2-scalacheck"         % specs2Core.revision
   lazy val tomcatCatalina                   = "org.apache.tomcat"      %  "tomcat-catalina"           % "9.0.6"
   lazy val tomcatCoyote                     = "org.apache.tomcat"      %  "tomcat-coyote"             % tomcatCatalina.revision
-  lazy val twirlApi                         = "com.typesafe.play"      %% "twirl-api"                 % "1.3.14"
+  lazy val twirlApi                         = "com.typesafe.play"      %% "twirl-api"                 % "1.3.15"
```
rossabaker: Is there a build info published by this project so we could DRY this up with plugins.sbt?  I can't find one.

rossabaker:
Libraries held back:
- scala-xml.  scala/scala-xml#121 changes behavior, and we can let people opt into that by upgrading themselves, as it's bincompat.  We will use this starting in 0.19.
- blaze (#1660)
- servlet-api (Jetty doesn't support 4 yet in a stable version)

This is otherwise The Latest And Greatest

rossabaker:
Eh, let's not sit on progress.

PR #1709, opened 2018-03-15
adds request/response specific properties to logging

### LINKED ISSUE #1708: middleware.Logging should log interesting request-/response-specific fields, such as method and response status

megri:
* Request: method, http-version, uri
* Response: status, http-version

Resolves #1708

### DIFF
--- a/server/src/main/scala/org/http4s/server/middleware/Logger.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/Logger.scala
@@ -38,7 +38,7 @@ object Logger {
 
     def prelude = message match {
       case Request(method, uri, httpVersion, _, _, _) =>
-        s"$method $httpVersion $uri"
+        s"$httpVersion $method $uri"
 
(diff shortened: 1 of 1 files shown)


megri:
Sample output:

Request: `o.h.server.middleware.RequestLogger HTTP/1.1 POST /api/items?foo=bar Headers(Host: localhost:8080, User-Agent: curl/7.54.0, Accept: */*)`
Response: `o.h.server.middleware.ResponseLogger HTTP/1.1 422 Unprocessable Entity Headers(Content-Length: 108, Content-Type: text/plain; charset=UTF-8)`

rossabaker:
I still think it's to our discredit that we don't have a simple way of outputting the [common log format](
https://en.wikipedia.org/wiki/Common_Log_Format), but this seems like a good step in the right direction.

rossabaker:
Cherry-picked.  This will be in 0.18.3.

PR #1714, opened 2018-03-17
Change Streaming to a more idiomatic response

ChristopherDavenport:
Since responses will be automatically disposed with this signature, this allows a more direct experience to a Response which can be worked with, rather than modified inside this function only.

### DIFF
--- a/client/src/main/scala/org/http4s/client/Client.scala
+++ b/client/src/main/scala/org/http4s/client/Client.scala
@@ -116,17 +116,14 @@ final case class Client[F[_]](
       }
       .mapF(OptionT.liftF(_))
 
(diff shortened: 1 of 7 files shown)


rossabaker:
Half-baked thought: what if `Client` were a wrapper around `Request[F] => Stream[F, Response[F]]` and get rid of `DisposableResponse[F]` entirely?

ChristopherDavenport:
I love it! Guarantees termination everywhere. All that changes are how we use the stream semantics. Which makes it extremely easy to explain to newcomers what each function makes available. 

If client is being redesigned overloads die.

rossabaker:
This has been superseded by #2102, but I like the `streaming` signature.

PR #1715, opened 2018-03-18
Add non-deprecated LocationResponseGenerator functions

gheine:


gheine:
@ChristopherDavenport looks like the deprecation was introduced during 0.18 work, when all response generators (except for this one) where updated to accept a `headers: Header*` and rather than passing a required header value (e.g. Uri in case of Location header) to pass the actual Location header.

dsl/src/main/scala/org/http4s/dsl/impl/ResponseGenerator.scala
```diff
@@ -60,6 +60,28 @@ trait LocationResponseGenerator[F[_]] extends Any with EntityResponseGenerator[F
   def apply(location: Uri)(implicit F: Applicative[F]): F[Response[F]] =
     F.pure(
       Response[F](status = status, headers = Headers(`Content-Length`.zero, Location(location))))
+
+  /** A 300, 301, 302, 303, 307 and 308 status SHOULD contain a Location header, which
```
jcranky: Maybe this scaladoc works better at the trait level? This way we avoid the duplication.

gheine: Fair enough, I was following the pattern in WwwAuthenticateResponseGenerator but happy to move doc to trait level.

rossabaker:
This looks good to me.  I think it's binary compatible.  I'm going to try to get it onto 0.18 branch.

rossabaker:
Cherry-picked to release-0.18.x, and will be merged back to master. Thanks for the contribution!

rossabaker:
I was wrong.  It's not binary compatible.  master was the right target after all.

PR #1719, opened 2018-03-20
Upgrade Scala to 2.12.5

aeons:


aeons:
Hmm, the Java 9 build seems to fail.

rossabaker:
Repeatedly? We could also target 0.18 here.

aeons:
I get it locally as well. It seems to break on macros.
```
... http4s/core/src/main/scala/org/http4s/QValue.scala:99:4: macro implementation not found: macroTransform
[error] (the most common reason for that is that you cannot use macro implementations in the same compilation run that defines them)
[error]   @bundle
[error]    ^
```

CremboC:
Unfortunately there's a regression in 2.12.5: https://github.com/scala/scala-dev/issues/480

aeons:
Was just about to comment that :)

We only compile on Java 9 for testing purposes, right?

Skip 2.12.5 or disable java 9 builds for now? As I understand it, running the jars compiled on Java 8 works on 9.

aeons:
Also Java 10 is out.

rossabaker:
We're not really getting anything out of 2.12.5 other than a broken build. I think we should pass and wait for 2.12.6.

A travis on jdk 10 is probably not a bad idea.  It could get absurd if they keep releasing every 6 months, but I guess we can argue about that in six months.

jmcardon:
re: jdk10, if we don't want CI to go crazy about it, we could wait for the next LTS.

jcranky:
> re: jdk10, if we don't want CI to go crazy about it, we could wait for the next LTS.

It shouldn't get that crazy because as we add new versions, some old ones should be removed, right? i.e. we should probably just keep the supported versions.

jmcardon:
> It shouldn't get that crazy because as we add new versions, some old ones should be removed, right? i.e. we should probably just keep the supported versions.

Well, we should keep the latest LTS until the next one (Which is planned for september this year I think? JDK 11 im talking about).

8 and 9 are bound to break stuff in scala. I'm not sure how much we're planning to support what are essentially minor JDK versions (even though it's called jdk9, on a 6 month release cycle it's practically a minor version with 10 already out).

aeons:
Let's just close this one for now.

PR #1721, opened 2018-03-20
adds server-logging functionality to BlazeBuilder

megri:
This PR adds basic support for the Common Log Format as discussed @ #1708

A few things are still missing or unclear:
* A request that fails and gets handled by `ServiceErrorHandler[F]` seems to yield a `Response[F]` without a Content-Length header. Blaze handles this "elsewhere", like in https://github.com/http4s/http4s/blob/d973c013ca0913ed3250d4c8eefbf8e2e7647ece/blaze-core/src/main/scala/org/http4s/blazecore/Http1Stage.scala#L70. At this point the response is disregarded so the logger will output an unhelpful "-" instead of the actual byte-count
* `Request[F]` has a value `.remoteUser` which is is always None. Not sure what this is for.

The initial implementation focuses more on performance than elegance as a logger at this level means a performance hit for every request. It may need to be even more low-level.

Packages, naming and formatting will probably need some clean-up. I will address this as I go.

### DIFF
--- a/server/src/main/scala/org/http4s/server/logging/ServerLogging.scala
+++ b/server/src/main/scala/org/http4s/server/logging/ServerLogging.scala
@@ -1,0 +1,46 @@
+package org.http4s
+package server
+package logging
+
+import java.time.format.DateTimeFormatter
+import java.time.{Clock, OffsetDateTime}
+
+import org.http4s.headers.Authorization
+
+abstract class ServerLogging[F[_]] {
+  def enabled: Boolean
+  def logRequestResponse(request: Request[F], response: Response[F]): Unit
+}
+
+object ServerLogging {
+  def apply[F[_]](log: (Request[F], Response[F]) => Unit): ServerLogging[F] =
+    new ServerLogging[F] {
+      override final val enabled = true
+      override def logRequestResponse(request: Request[F], response: Response[F]): Unit =
+        log(request, response)
+    }
+
+  def disabled[F[_]]: ServerLogging[F] = new ServerLogging[F] {
+    override final val enabled: Boolean = false
+    override def logRequestResponse(request: Request[F], response: Response[F]): Unit = ()
+  }
+
+  def commonLogFormat[F[_]](clock: Clock = Clock.systemDefaultZone()): ServerLogging[F] = {
+    val formatter = DateTimeFormatter.ofPattern("dd/MMM/yyyy:hh:mm:ss Z")
+
+    apply[F] { (req, res) =>
+      val remotehost = req.remoteAddr.getOrElse("-")
+      val rfc931 = "-"
+      val authuser = Authorization.from(req.headers).map(_.credentials) match {
+        case Some(BasicCredentials(credentials)) => credentials.username
+        case _ => "-"
+      }
+      val date = formatter.format(OffsetDateTime.now(clock))
+      val request = s"${req.method} ${req.uri} ${req.httpVersion}"
+      val status = res.status.code
+      val bytes = res.contentLength.map(_.toString).getOrElse("-")
+
+      println(s"""$remotehost $rfc931 $authuser [$date] "$request" $status $bytes""")
+    }
+  }
+}
--- a/server/src/main/scala/org/http4s/server/logging/ServerLogging.scala
+++ b/server/src/main/scala/org/http4s/server/logging/ServerLogging.scala
@@ -23,21 +23,24 @@ object ServerLogging {
   }
 
   def commonLogFormat[F[_]](clock: Clock = Clock.systemDefaultZone()): ServerLogging[F] = {
+    val log = org.log4s.getLogger
     val formatter = DateTimeFormatter.ofPattern("dd/MMM/yyyy:hh:mm:ss Z")
 
     apply[F] { (req, res) =>
-      val remotehost = req.remoteAddr.getOrElse("-")
-      val rfc931 = "-"
-      val authuser = Authorization.from(req.headers).map(_.credentials) match {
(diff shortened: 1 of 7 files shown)


megri:
Just ran scalafmt on everything.

server/src/main/scala/org/http4s/server/logging/ServerLogging.scala
```diff
@@ -0,0 +1,46 @@
...
+package server
+package logging
+
+import java.time.format.DateTimeFormatter
+import java.time.{Clock, OffsetDateTime}
+
+import org.http4s.headers.Authorization
+
+abstract class ServerLogging[F[_]] {
+  def enabled: Boolean
```
rossabaker: Where is enabled called?  I couldn't find it via search.

server/src/main/scala/org/http4s/server/logging/ServerLogging.scala
```diff
@@ -0,0 +1,46 @@
...
+      val authuser = Authorization.from(req.headers).map(_.credentials) match {
+        case Some(BasicCredentials(credentials)) => credentials.username
+        case _ => "-"
+      }
+      val date = formatter.format(OffsetDateTime.now(clock))
+      val request = s"${req.method} ${req.uri} ${req.httpVersion}"
+      val status = res.status.code
+      val bytes = res.contentLength.map(_.toString).getOrElse("-")
+
+      println(s"""$remotehost $rfc931 $authuser [$date] "$request" $status $bytes""")
```
jmcardon: We might definitely want a custom logger here, but definitely not println.

That aside, I kid you not if we're looking @ performance heavy code, we're better off hand-rolling string builders.

server/src/main/scala/org/http4s/server/logging/ServerLogging.scala
```diff
@@ -0,0 +1,46 @@
...
+package logging
+
+import java.time.format.DateTimeFormatter
+import java.time.{Clock, OffsetDateTime}
+
+import org.http4s.headers.Authorization
+
+abstract class ServerLogging[F[_]] {
+  def enabled: Boolean
+  def logRequestResponse(request: Request[F], response: Response[F]): Unit
```
rossabaker: If we pass a byte count here, we could also log the size of chunked transfer encodings.
If we pass an elapsed time here, we could log the elapsed time.
If we throw every ad hoc thing anyone might want to log into this signature, this signature will become unwieldy.

:thinking:

server/src/main/scala/org/http4s/server/logging/ServerLogging.scala
rossabaker: Should probably pass this a logger or a `String => ()`.

There's a [fast string interpolator macro](https://github.com/sizmek/fast-string-interpolator), but I don't want to add a dependency for it.  I hate to make code crappier without a benchmark, but the `StringBuilder` seems like a good way out of if we can demonstrate a problem.

server/src/main/scala/org/http4s/server/logging/ServerLogging.scala
megri: You're right, this can be removed. The intent was to let a caller inspect this property to avoid creating a log string but the signature of logRequestResponse was later changed to be completely side-effecting.

server/src/main/scala/org/http4s/server/logging/ServerLogging.scala
megri: Yeah, I thought about having logRequestResponse take all parameters it needed directly, but this would mean that every instance of ServerLogging would be strictly forced to work with these parameters.

server/src/main/scala/org/http4s/server/logging/ServerLogging.scala
megri: An instance of ServerLogging can be implemented to use a logger. The reason the commonLogFormat-instance is using println is exactly to avoid having a logger attach a bunch of timestamps and other crap that would make it output something other than Common Log Format.

jmcardon: I agree on trying to not output cruft, but I think the logger can be a custom one so it doesn't go to stdout.

megri:
Now using org.log4s.Logger for logging. Users need to configure logging in logback.xml like so:

```xml
    <logger name="org.http4s.server.logging.ServerLogging" additivity="false">
        <appender class="ch.qos.logback.core.ConsoleAppender">
            <encoder>
                <pattern>%msg%n</pattern>
            </encoder>
        </appender>
    </logger>
```

server/src/main/scala/org/http4s/server/logging/ServerLogging.scala
```diff
@@ -23,21 +23,24 @@ object ServerLogging {
   }
 
   def commonLogFormat[F[_]](clock: Clock = Clock.systemDefaultZone()): ServerLogging[F] = {
+    val log = org.log4s.getLogger
```
rossabaker: I suggested either this or a `String => Unit`.  I'm now thinking maybe the latter makes more sense.  Then you could log directly to the console (and not have to do the irritating XML config) or go through log4s (and get nice rolling logs if you're not externally slurping stdout).

megri: commonLogFormat now takes an extra argument to allow custom output methods.

megri:
I'm experimenting with collecting the logging data into an internally mutable datatype as it is parsed. It will take some more effort before it's ready though.

armanbilge:
This PR seems stale and is largely blaze-focused, which is no longer developed here.

PR #1722, opened 2018-03-21
blaze-0.12.12, log4s-1.6.1, jetty-9.4.9

rossabaker:
The main event is getting rid of the log4s compatibility warning.


rossabaker:
MiMa is angry because I erroneously tried to cherry-pick #1715.

PR #1726, opened 2018-03-23
Add expectOr methods for custom error handling on failed expects

rossabaker:
Reopening #1634 so we can get it across the finish line.  It looks like this is binary compatible.  Apologies for this falling off my radar.

Is this helpful as is, or do we want to address the discussion in #1632 to make it useful?

/cc @kailuowang


kailuowang:
:+1: Definitely helpful for my use cases.

rossabaker:
Alright, MiMa liked this.  We can get this into 0.18.4 if it's positively reviewed.

PR #1727, opened 2018-03-23
Port multipart goodness to to 0.18.x branch

jmcardon:
:D


jmcardon:
seems to pass all tests, but trying it on a small client I cannot get it to run properly, so I'm double checking first if there is a problem

jmcardon:
Bugs have been fixed, and tested with postman. See notes on #1728.

jmcardon:
@rossabaker mostly that the #1728 version doesnt deprecate the other stuff for the sake of MiMa, it straight up deletes the old version.

rossabaker:
Yeah, it's going to have to to get rid of that dependency.  I am pro-deprecation until it exerts drag, like pulling in a new dependency.

Okay, merging this one, will roll forward, and will use #1728 as a guideline if the merge is ugly.

PR #1728, opened 2018-03-23
Remove scodec bits + streaming multipart

### LINKED ISSUE #876: Use ByteVectorChunk?
`ByteVectorChunk` is a type of `Chunk[Byte]` which is better optimized for avoiding copies.  Based on our experience with byte copying in scalaz-stream, I have a feeling this will be a big win for us.

But it only exists in fs2-1.0-SNAPSHOT.  We should think about how we want to cut over, or whether we want to kindly ask for an fs2-0.10 so we can start building on that.

### LINKED ISSUE #1531: Multipart Streaming
As a part of the move to fs2 we lost the streaming multipart and implemented a buffered replacement.

It would be nice at some point to clear this out. I am documenting what difficulties were found. What is essentially needed is a Streaming parser implementation.

The specification
Complete Body: `Stream[F, Byte]` where there we have a Preamble, The Multipart, and the Epilogue - Preamble and Epilogue may be discarded. The Multipart is bordered by the Boundary. And an edge boundary with has `--` as well

Multipart is made up of a Stream of Parts - These parts are composed of headers, and then a stream of bytes. The headers and the body is seperated by `/r/n/r/n`.

I will expand in more detail but this is the initial sketch.

jmcardon:
Continued from #1701 

I'm an idiot and deleted my old fork so now I can't push to it. Thus this exists.

I also accidentally created the branch on the main repo. Sorry. Will delete once this is over. I just needed to check out the PR.

Closes #1531 
Closes #876 

Benchmarks on impl:
```
[info] # Run complete. Total time: 00:03:06
[info] Benchmark                                      Mode  Cnt    Score    Error  Units
[info] Bench.MultipartBench.bigNewMultipart          thrpt   10   22.593 ±  1.373  ops/s
[info] Bench.MultipartBench.bigOldMultipart          thrpt   10    1.342 ±  0.154  ops/s
[info] Bench.MultipartBench.bigPartBodyNewMultipart  thrpt   10   26.711 ±  0.654  ops/s
[info] Bench.MultipartBench.bigPartBodyOldMultipart  thrpt   10    1.583 ±  0.043  ops/s
[info] Bench.MultipartBench.mediumNewMultipart       thrpt   10   71.913 ±  2.072  ops/s
[info] Bench.MultipartBench.mediumOldMultipart       thrpt   10    6.138 ±  0.111  ops/s
[info] Bench.MultipartBench.smallNewMultipart        thrpt   10  746.061 ± 30.352  ops/s
[info] Bench.MultipartBench.smallOldMultipart        thrpt   10  739.245 ± 29.035  ops/s
```

Located [here](https://github.com/jmcardon/routing-bench/blob/master/src/main/scala/Bench/MultipartBench.scala)

Also [here](https://github.com/jmcardon/routing-bench/blob/master/src/main/scala/com/example/quickstart/HelloWorldServer.scala#L154) is an example server with this implementation currently running. Attempted on postman to double check stuff works.


rossabaker:
Came from #1727, mastered by #1730.

PR #1730, opened 2018-03-23
MergeTrain 0.18.4 -> 0.19

ChristopherDavenport:
I really just took the shortest path to recompilation here.

Please check my work. I know I needed ConcurrentEffect in the examples which seems needlessly strong, but I was just following types.

### DIFF
--- a/core/src/main/scala/org/http4s/EntityDecoder.scala
+++ b/core/src/main/scala/org/http4s/EntityDecoder.scala
@@ -176,6 +176,7 @@ object EntityDecoder extends EntityDecoderInstances {
     override val consumes: Set[MediaRange] = (r1 +: rs).toSet
   }
 
+  /**
   private class OrDec[F[_]: Functor, T](a: EntityDecoder[F, T], b: EntityDecoder[F, T])
       extends EntityDecoder[F, T] {
     override def decode(msg: Message[F], strict: Boolean): DecodeResult[F, T] =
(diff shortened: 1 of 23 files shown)


core/src/main/scala/org/http4s/EntityDecoder.scala
```diff
@@ -176,7 +176,36 @@ object EntityDecoder extends EntityDecoderInstances {
     override val consumes: Set[MediaRange] = (r1 +: rs).toSet
   }
 
-  /** Helper method which simply gathers the body into a single ByteVector */
+  /**
+  private class OrDec[F[_]: Functor, T](a: EntityDecoder[F, T], b: EntityDecoder[F, T])
```
ChristopherDavenport: So this is unused now...

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/Module.scala
```diff
@@ -14,10 +14,9 @@ import org.http4s.client.Client
 import org.http4s.server.HttpMiddleware
 import org.http4s.server.middleware.{AutoSlash, ChunkAggregator, GZip, Timeout}
 
-import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.duration._
 
-class Module[F[_]](client: Client[F])(implicit F: Effect[F], S: Scheduler) {
+class Module[F[_]](client: Client[F])(implicit F: ConcurrentEffect[F], S: Scheduler, T: Timer[F]) {
```
aeons: What is the scheduler here? Is it the fs2 one? In that case isn't it superseded by the timer?

core/src/main/scala/org/http4s/EntityDecoder.scala
rossabaker: Oh, neat, because `orElse` is a SemigroupK now.  (I wonder what changed in this merge, though.)

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/Module.scala
rossabaker: `Scheduler` is deprecated in fs2-1.0.  If we can replace it in http4s-0.19, that would be ideal.  Anywhere we have both is suspect.

I'm fine doing that in a subsequent PR.

PR #1735, opened 2018-03-24
Tune scalacheck parameters for SemigroupK[EntityDecoder[IO, ?]]

rossabaker:
This test has added about 5 minutes to our builds.  Build less and smaller.


rossabaker:
It still takes about two minutes on Travis. It seemed like it was running about four before.

It's 15 seconds on my laptop.

PR #1736, opened 2018-03-24
Remove some unneeded schedulers

aeons:
`Stream` doesn't have `awakeEvery` directly, until https://github.com/functional-streams-for-scala/fs2/issues/1100

### DIFF
--- a/blaze-client/src/test/scala/org/http4s/client/blaze/PooledClientSpec.scala
+++ b/blaze-client/src/test/scala/org/http4s/client/blaze/PooledClientSpec.scala
@@ -7,7 +7,6 @@ import java.net.InetSocketAddress
 import javax.servlet.ServletOutputStream
 import javax.servlet.http.{HttpServlet, HttpServletRequest, HttpServletResponse}
 import org.http4s._
-import org.http4s.Http4sSpec.TestScheduler.sleep_
 import org.http4s.client.testroutes.GetRoutes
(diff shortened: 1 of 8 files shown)


aeons:
Is the failing test here wrong?

```scala
"clean up resources of the loser" in {
  val clean = new AtomicBoolean(false)
    case _ =>
      for {
        resp <- IO.sleep(2.seconds) *> NoContent()
        _ <- IO(clean.set(true))
      } yield resp
  }
  val timeoutService = Timeout(1.millis)(service)
  checkStatus(timeoutService.orNotFound(Request[IO]()), InternalServerError)
  // Give the losing response enough time to finish
  clean.get must beTrue.eventually
}
```

Using the new cancelability of cats effect, the losing side should be cancelled, right? So any cleanup should be captured in something else, like a stream.

rossabaker:
Yeah, we can do better now by making that response trigger the flag on cancelation and be sure that gets called. That would be true on 0.18, even.

aeons:
I'll retarget 0.18 and take a stab at improving tests.

rossabaker:
Much of this breaks bincompat with different implicit parameters, so the whole thing can't be retargeted to release-0.18.x.  But I think the test can be improved in 0.18.x.  I should have done that when I redid the timeout middleware.

aeons:
I'll close this and reopen in two different PRs.

PR #1738, opened 2018-03-25
Set up /stable and /latest rewrites

rossabaker:
After this, we can set up a banner to render on EOL versions of the site, and always direct to the current versions without having to go back and republish those.

### DIFF
--- a/website/src/hugo/static/_redirects
+++ b/website/src/hugo/static/_redirects
@@ -1,0 +1,2 @@
+/stable/*   /v0.18/:splat   200
+/latest/*   /v0.19/:splat   200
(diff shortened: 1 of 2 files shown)


website/src/hugo/static/_redirects
```diff
@@ -0,0 +1,2 @@
+/stable/*   /v0.18/:splat   200
+/latest/*   /v0.19/:splat   200
```
rossabaker: I _think_ 200 is the right thing to do for SEO, so we have unchanging URIs and can start to accrue page rank.  But that's a dark art.  Advice welcome here.

PR #1739, opened 2018-03-26
Remove internal syntax for NonEmptyList and use cat's instead

### LINKED ISSUE #1510: Remove Http4s NonEmpty List

cquiroz:
Completely removes the internal custom syntax for `NonEmptyList` using cats' instead. Note this removes some deprecated instances and may break existing code.

Is there a better alternative for `collectFirst` in cats?

Fixes #1510

### DIFF
--- a/core/src/main/scala/org/http4s/headers/ETag.scala
+++ b/core/src/main/scala/org/http4s/headers/ETag.scala
@@ -3,6 +3,7 @@ package headers
 
 import org.http4s.parser.HttpHeaderParser
 import org.http4s.util.Writer
+import cats.Show
 
 object ETag extends HeaderKey.Internal[ETag] with HeaderKey.Singleton {
   final case class EntityTag(tag: String, weak: Boolean = false) {
@@ -11,10 +12,14 @@ object ETag extends HeaderKey.Internal[ETag] with HeaderKey.Singleton {
       else "\"" + tag + '"'
   }
 
+  implicit val http4sShowForEntityTag: Show[EntityTag] =
+    Show.fromToString
+
   def apply(tag: String, weak: Boolean = false): ETag = ETag(EntityTag(tag, weak))
 
   override def parse(s: String): ParseResult[ETag] =
     HttpHeaderParser.ETAG(s)
+
 }
 
 final case class ETag(tag: ETag.EntityTag) extends Header.Parsed {
--- a/core/src/main/scala/org/http4s/headers/If-None-Match.scala
+++ b/core/src/main/scala/org/http4s/headers/If-None-Match.scala
@@ -2,8 +2,8 @@ package org.http4s
 package headers
 
 import cats.data.NonEmptyList
+import cats.syntax.foldable._
 import org.http4s.parser.HttpHeaderParser
-import org.http4s.syntax.nonEmptyList._
 import org.http4s.util.Writer
 
 object `If-None-Match` extends HeaderKey.Internal[`If-None-Match`] with HeaderKey.Singleton {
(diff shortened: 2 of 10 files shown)


dsl/src/test/scala/org/http4s/dsl/PathInHttpServiceSpec.scala
```diff
@@ -3,6 +3,7 @@ package dsl
 
 import cats.data.Validated._
 import cats.effect.IO
+import cats.implicits._
```
jmcardon: I think `cats.syntax.foldable._` here is preferrable to `.implicits`. I like `.implicits._` for examples more than I do for our own code (and it makes some editors poop themselves with so many syntax objs)

core/src/main/scala/org/http4s/headers/If-None-Match.scala
```diff
@@ -2,8 +2,8 @@ package org.http4s
 package headers
 
 import cats.data.NonEmptyList
+import cats.implicits._
```
jmcardon: see: other comment re: mkString_ syntax

dsl/src/test/scala/org/http4s/dsl/PathInHttpServiceSpec.scala
rossabaker: I don't like editors that poop themselves, but I'm willing to concede on this.  I still strongly favor it for examples, because new users shouldn't have to pay a huge import tax up front.

cquiroz:
Updated imports as suggested

jmcardon:
```
[error] /home/travis/build/http4s/http4s/core/src/main/scala/org/http4s/syntax/AllSyntax.scala:10:10: trait NonEmptyListSyntax in package syntax is deprecated (since 0.19.0): Use cats.foldable._
[error]     with NonEmptyListSyntax
[error]          ^
[error] one error found
[error] (core / Compile / compileIncremental) Compilation failed
[error] Total time: 79 s, completed Mar 27, 2018 1:30:27 AM
```
>  I think we could deprecate instead of remove, and make it easier for people.

LOL. Oh scalac

rossabaker:
I thought I commented, but I guess not.  I think we need to deprecate the methods instead of the trait, because the trait is still mixed into a desired object.

jmcardon:
I didn't expect scalac to do that. That's pretty funny :smile: 

but yeah, we gotta do the methods.

cquiroz:
I should probably squash all the commits when this is done, WDYT?

cquiroz:
I can do it. Would it be `release-0.18.x`?

rossabaker:
Yes.

cquiroz:
I got this after rebasing:
```
[info] core: found 1 potential binary incompatibilities while checking against org.http4s:http4s-core_2.12:0.18.4
[error]  * the type hierarchy of object org.http4s.headers.ETag#EntityTag is different in current version. Missing types {scala.runtime.AbstractFunction2}
[error]    filter with: ProblemFilters.exclude[MissingTypesProblem]("org.http4s.headers.ETag$EntityTag$")
```

Is this MIMA related? I added a `Show` instance to `ETag.EntityTag`

rossabaker:
Looks like the generated companion object used to extend AbstractFunction2, and no longer does.  You could explicitly extend that and preserve bincompat, or defer the `Show` until master.

cquiroz:
I think i got it moving the `Show` instance. It is needed to make `mkString_` work.

core/src/main/scala/org/http4s/headers/ETag.scala
```diff
@@ -11,10 +12,14 @@ object ETag extends HeaderKey.Internal[ETag] with HeaderKey.Singleton {
       else "\"" + tag + '"'
   }
 
+  implicit val http4sShowForEntityTag: Show[EntityTag] =
```
rossabaker: In the long run, we'll want to get this back into the companion object, so it's not an orphan instance.  But this is okay for now.

cquiroz:
@jmcardon It is going to `release-0.18.x`
@rossabaker I'll move the `Show` instance to the right companion on `master

jmcardon:
Ninja-swap.

cquiroz:
Should I do another PR targeting `master` or are we doing merge trains every now and then?

rossabaker:
I run merge trains when I think about it.  If it's trivial, I merge it.  If it's difficult, I PR it.  I'll run one now, but leave moving the `Show` for you.

PR #1748, opened 2018-03-27
 Fix failure in ContentCodingSpec

### LINKED ISSUE #1669: ContentCodingSpec failure
[Huh?](https://ci.appveyor.com/project/ChristopherDavenport/http4s-m0hqu/build/1.0.810/job/yvi6u8iqrk9ycu6e)

```
[info] compare should
[error]   x be consistent with coding.compareToIgnoreCase for same quality (47 ms)
[error]    Falsified after 11 passed tests.
[error]    > ARG_0: ContentCoding(l, QValue(0.6900000000000001))
[error]    > ARG_1: ContentCoding(l, QValue(0.0))
[error]    > 690 is not less than 0 (ContentCodingSpec.scala:24)
[error] org.http4s.ContentCodingSpec.$anonfun$new$10(ContentCodingSpec.scala:24)
```

rossabaker:
Only in the first case did we validate that the qValues were equal.

Fixes #1669

### DIFF
--- a/tests/src/test/scala/org/http4s/ContentCodingSpec.scala
+++ b/tests/src/test/scala/org/http4s/ContentCodingSpec.scala
@@ -4,7 +4,6 @@ import cats.implicits._
 import cats.kernel.laws.discipline.OrderTests
 import org.http4s.testing.HttpCodecTests
 import org.http4s.util.Renderer
-import scala.math.signum
 
 class ContentCodingSpec extends Http4sSpec {
   "equals" should {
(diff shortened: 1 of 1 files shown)


rossabaker:
I'm now wondering whether the ordering should be based on qValue alone.  If I ever want to sort these, I want to sort by qValue.  I don't think a lexicographical tiebreaker makes a whole lot of sense.  If we sort on that, why not sort, say, methods?

PR #1749, opened 2018-03-28
Place Show[EntityTag] in the companion object

cquiroz:
This is a leftover change from PR #1739 applicable to master

### DIFF
--- a/core/src/main/scala/org/http4s/headers/ETag.scala
+++ b/core/src/main/scala/org/http4s/headers/ETag.scala
@@ -12,8 +12,10 @@ object ETag extends HeaderKey.Internal[ETag] with HeaderKey.Singleton {
       else "\"" + tag + '"'
   }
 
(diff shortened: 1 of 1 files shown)


cquiroz:
This is odd. Travis is failing in 2.12 with this
```
[tut] *** Error reported at /home/travis/build/http4s/http4s/docs/src/main/tut/testing.md:40
// <console>:46: warning: method withBody in trait Message is deprecated (since 0.19): Use withEntity
//              case Some(user) => Response(status = Status.Ok).withBody(user.asJson)
//                                                              ^
// error: No warnings can be incurred under -Xfatal-warnings.
```

Any ideas?

rossabaker:
We renamed it to `withEntity` when it stopped being effectful.  And it's only happening in 2.12 because of the tuts.  And I think it's acting up now because we merged a tut from 0.18.  You didn't break this.

PR #1718, opened 2018-03-20
Add Testing Docs for v0.18

kevinmeredith:
From https://github.com/http4s/http4s/pull/1699

Could you please take a look, @rossabaker, @jcranky, @jmcardon, @ChristopherDavenport? Thanks!

Closes https://github.com/http4s/http4s/issues/1504

PR #1752, opened 2018-03-28
Fix deprecated call in testing.md

rossabaker:
I broke this when I merged #1718 to master without running tut.  My fault.  This is what ails #1749.

Observation: it's hard to use the DSL for a polymorphic service inside a tut.


cquiroz:
👍 On green

PR #1753, opened 2018-03-28
Update tests to take advantage of cats-effect-0.10

aeons:
Test part of #1736

### DIFF
--- a/server/src/test/scala/org/http4s/server/middleware/TimeoutSpec.scala
+++ b/server/src/test/scala/org/http4s/server/middleware/TimeoutSpec.scala
@@ -7,6 +7,7 @@ import cats.implicits._
 import java.util.concurrent.TimeoutException
 import java.util.concurrent.atomic.AtomicBoolean
 import org.http4s.dsl.io._
+import scala.concurrent.CancellationException
 import scala.concurrent.duration._
 
 class TimeoutSpec extends Http4sSpec {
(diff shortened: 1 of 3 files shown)


server/src/test/scala/org/http4s/server/middleware/TimeoutSpec.scala
```diff
@@ -45,22 +46,20 @@ class TimeoutSpec extends Http4sSpec {
       checkStatus(altTimeoutService.orNotFound(neverReq), customTimeout.status)
     }
 
-    "clean up resources of the loser" in {
-      val clean = new AtomicBoolean(false)
+    "cancel the loser" in {
+      val cancelled = new AtomicBoolean(false)
```
rossabaker: English uses both, but cats-effect uses "canceled".

It's not part of the API, so I won't die on this hill.

ChristopherDavenport: Nitpicky much...

server/src/test/scala/org/http4s/server/middleware/TimeoutSpec.scala
```diff
@@ -47,19 +47,19 @@ class TimeoutSpec extends Http4sSpec {
     }
 
     "cancel the loser" in {
-      val cancelled = new AtomicBoolean(false)
+      val canceled = new AtomicBoolean(false)
```
jmcardon: The travis failure is deserve. Queens english ftw.

PR #1754, opened 2018-03-28
Use cats.effect.Timer for Retry middleware

aeons:


client/src/main/scala/org/http4s/client/middleware/Retry.scala
```diff
@@ -54,7 +50,7 @@ object Retry {
         duration: FiniteDuration,
         retryHeader: Option[`Retry-After`])(
         implicit F: Effect[F],
-        executionContext: ExecutionContext): F[DisposableResponse[F]] = {
+        T: Timer[F]): F[DisposableResponse[F]] = {
```
rossabaker: Don't these implicits just shadow the ones in apply?

aeons: Yes, I didn't pay enough attention to the indentation in this file. Thought they were separate things.

PR #1761, opened 2018-03-29
Upgrade to blaze master

bryce-anderson:
This gets us back on blaze master. Some pieces that had to be merged into http4s include the websocket bits.


bryce-anderson:
After we get back on current blaze I'd like to get the http2 client working and see if we can de-dup things from the client and server stages that exist in both raw blaze and http4s. But baby steps...

blaze-core/src/main/scala/org/http4s/blazecore/websocket/Serializer.scala
```diff
@@ -0,0 +1,117 @@
...
+
+  ///  channel reading bits //////////////////////////////////////////////
+
+  override def channelRead(size: Int = -1, timeout: Duration = Duration.Inf): Future[I] = {
+    val p = Promise[I]
+    val pending = serializerReadRef.getAndSet(p.future)
+
+    if (pending == null) serializerDoRead(p, size, timeout)  // no queue, just do a read
+    else {
+      val started = if (timeout.isFinite()) System.currentTimeMillis() else 0
```
rossabaker: Should this be using nanoTime to avoid the non-monotonic shenanigans of currentTimeMillis?  (I think yes).

Should this be using Deadline so you don't have to do this bookkeeping?  (I don't think you can, but maybe it clicks for you.)

blaze-server/src/main/scala/org/http4s/server/blaze/BlazeBuilder.scala
```diff
@@ -8,17 +8,20 @@ import java.net.InetSocketAddress
 import java.nio.ByteBuffer
 import java.security.{KeyStore, Security}
 import javax.net.ssl.{KeyManagerFactory, SSLContext, SSLEngine, TrustManagerFactory}
+
```
rossabaker: So am I just fighting a losing battle here on using Twitter's import style over IntelliJ's?

blaze-server/src/main/scala/org/http4s/server/blaze/Http2NodeStage.scala
```diff
@@ -179,8 +178,9 @@ private class Http2NodeStage[F[_]](
       error += s"Invalid request: missing pseudo headers. Method: $method, Scheme: $scheme, path: $path. "
     }
 
-    if (error.length() > 0) shutdownWithCommand(Cmd.Error(PROTOCOL_ERROR(error, fatal = false)))
-    else {
+    if (error.length() > 0) {
```
rossabaker: Nit: length doesn't have a side effect. No parens.

blaze-server/src/main/scala/org/http4s/server/blaze/WSFrameAggregator.scala
```diff
@@ -0,0 +1,101 @@
...
+import scala.util.{Failure, Success}
+
+import java.net.ProtocolException
+
+private class WSFrameAggregator extends MidStage[WebSocketFrame, WebSocketFrame] {
+
+  def name: String = "WebSocket Frame Aggregator"
+
+  private var queue = new ArrayBuffer[WebSocketFrame]
+  private var size = 0
```
rossabaker: Would this be any easier to maintain if it were a case class that held a buffer and the size, and adding to the buffer added to the size, and clearing it zeroed the size?  It seems like those mutations must happen together.

blaze-server/src/main/scala/org/http4s/server/blaze/BlazeBuilder.scala
bryce-anderson: No, I'm losing a battle with intellij. I'll fix it and maybe invest the time to make the auto formatter work for me... Sorry you need to keep reminding me. 🙁

blaze-core/src/main/scala/org/http4s/blazecore/websocket/Serializer.scala
bryce-anderson: Deadline is much better and that is probably the better interface for blaze pipeline api's as well. Maybe they can make it into the blaze-0.14 release.

blaze-server/src/main/scala/org/http4s/server/blaze/WSFrameAggregator.scala
bryce-anderson: Yes, that sounds like a good change.

blaze-core/src/main/scala/org/http4s/blazecore/websocket/Serializer.scala
bryce-anderson: The bad news about `Deadline` is that there aren't top or values for it. I'll go with making sure the times are monotonic.

bryce-anderson:
This upgrades to a blaze milestone, but not a full blown release. What approach do we want to take here? Keep this open and keep updating it until we can depend on a non-milestone, or just got with the milestone? I'm content either way, though all things being equal, I'd choose the former.

rossabaker:
Our last few releases have been triggered by big dependency releases, and I imagine the same for 0.19.  cats-effect-1.0 is targeted for May 15.  I would guess an fs2 release follows shortly, and then people will be knocking on our door.

There are a lot of ideas in http4s and blaze right now, and I worry that keeping large PRs open is a recipe for conflicts and bitrot.  I'm inclined to merge this and be prepared to call whatever we have in mid-May http4s-0.19 and blaze-0.14.  The symmetric difference of blaze and http4s users is tiny, so it would be great if they could go together.

bryce-anderson:
Lets merge it as is. Most of the changes I anticipate for blaze between now and May 15 are incremental. The next big thing will be the _may_ transition from `Future` to `IO` but that will definitely be a blaze 0.15 thing, if at all.

PR #1772, opened 2018-04-07
Client As Kleisli - Client Redesign Proposal

ChristopherDavenport:
```
type Client[F[_]] = Kleisli[Stream[F, ?], Request[F], Response[F]]
```
With Syntax Class For Operations

### DIFF
--- a/client/src/main/scala/org/http4s/client/Client.scala
+++ b/client/src/main/scala/org/http4s/client/Client.scala
@@ -6,63 +6,21 @@ import cats.data.{Kleisli, OptionT}
 import cats.effect._
 import cats.implicits._
 import fs2._
-import java.io.IOException
-import java.util.concurrent.atomic.AtomicBoolean
 import org.http4s.Status.Successful
 import org.http4s.headers.{Accept, MediaRangeAndQValue}
-import scala.concurrent.SyncVar
 import scala.util.control.NoStackTrace
-import org.log4s.getLogger
 
-/**
-  * Contains a [[Response]] that needs to be disposed of to free the underlying
-  * HTTP connection.
-  * @param response
-  * @param dispose
-  */
-final case class DisposableResponse[F[_]](response: Response[F], dispose: F[Unit]) {
+object Client {
 
-  private[this] val logger = getLogger
+  object Ops extends ClientSyntax
 
-  /**
-    * Returns a task to handle the response, safely disposing of the underlying
-    * HTTP connection when the task finishes.
-    */
-  def apply[A](f: Response[F] => F[A])(implicit F: MonadError[F, Throwable]): F[A] = {
-    //Catch possible user bugs in pure expression
-    val task: F[A] = try f(response)
-    catch {
-      case e: Throwable =>
-        logger.error(e)("""Handled exception in client callback to prevent a connection leak.
-             |The callback should always return an F. If your callback can fail
-             |with an exception you can't handle, call `F.raiseError(exception)`.
-          """.stripMargin)
-        F.raiseError(e)
-    }
+  trait ClientSyntax {
+    implicit class ClientOps[F[_] : Sync](k: Kleisli[Stream[F, ?], Request[F], Response[F]]){
+    
+    def streaming[A](req: Request[F]): Stream[F, Response[F]] =
+      k(req)
 
-    for {
-      result <- task.attempt
-      _ <- dispose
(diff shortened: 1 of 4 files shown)


client/src/main/scala/org/http4s/client/Client.scala
```diff
@@ -108,221 +59,125 @@ final case class Client[F[_]](
     * [[toKleisli]], and [[streaming]] are safer alternatives, as their
     * signatures guarantee disposal of the HTTP connection.
     */
-  def toHttpService: HttpService[F] =
-    open
-      .map {
-        case DisposableResponse(response, dispose) =>
-          response.copy(body = response.body.onFinalize(dispose))
+    def toHttpService: HttpService[F] = 
+      Kleisli{req: Request[F] => OptionT.liftF(fetch(req)(_.pure[F]))}
```
jmcardon: Alright this I find interesting: If reading the body of the request after fetch results in an error, I think `toHttpService` might not be expressible in terms of `fetch`

rossabaker: It's more problematic than that: the response is going to be closed when the `_.pure[F]` runs, so we're returning a dead response body.  We can't go from `Stream` to `F` without closing the resource.

rossabaker:
There has been a lot of talk on Gitter.  Not everybody follows there, so I'll summarize here.

---

# 1. Status quo
- easy `Client.toHttpApp`
- no opportunity to share middleware between client and server
- `FollowRedirect` and `Retry` are awkward
- Incorrect use of `.open` leaks
- presents ad hoc `DisposableResponse` type
- tricky to create `F[Response[F]]` from a resource acquired and released by `(F[A], A => F[Unit])`
- impossible to create `F[Response[F]]` from a resource bracketed in `Stream[F, A]`
- one failure mode in `F`

# 2. `type HttpClient[F] = Http[Disposable[F, ?], F]`
- easy `Client.toHttpApp`
- can share polymorphic middleware between client and server if they don't rerun the client
- `FollowRedirect` and `Retry` are awkward
- Incorrect use of `.run` leaks
- presents ad hoc `Disposable` monad transformer
- easy(?) to create `F[Response[F]]` from a resource acquired and released by `(F[A], A => F[Unit])`.
- impossible to create `F[Response[F]]` from a resource bracketed in `Stream[F, A]`
- one failure mode in `F`

# 3. `type HttpApp[F] = type HttpClient[F] = Http[Stream[F, ?], ?]` 
- easy `Client.toHttpApp`
- impossible to interpret into `Http[F, F]`
- can share polymorphic middleware between client and server if they don't rerun the client
- `FollowRedirect` and `Retry` are elegant
- No leaks
- No ad hoc monads
- easy to create `F[Response[F]]` from a resource acquired and released by `(F[A], A => F[Unit])`.
- easy to create `F[Response[F]]` from a resource bracketed in `Stream[F, A]`
- three failure modes in `Stream[F, ?]`: 0 outputs, >1 output, error
- more overhead for server, @jmcardon is measuring

---

I think 1 is strictly dominated by 2: it echoes the `MaybeResponse` mistake instead of being a monad.

2 vs. 3: discuss.  If still tied, we will thumb wrestle at the next NE Scala Symposium.

jmcardon:
Bench Results [here](https://github.com/jmcardon/http4s_bench_ix/blob/master/README.md)

I put them up on my website for fun.

rossabaker:
[Another quick and dirty benchmark](https://gist.github.com/rossabaker/6e7620239735728906a6ceacb86c45da).  Probably not as interesting as @jmcardon's.

```
[info] RenderBench.httpApp         avgt   60  1191.201 ±  15.814  ns/op
[info] RenderBench.httpStream      avgt   60  8563.606 ±  59.079  ns/op
[info] RenderBench.httpStreamEval  avgt   60  8914.895 ± 242.725  ns/op
```

This is proportionally a whole lot, marginally not a lot at all.

rossabaker:
Having studied gatling a bit more, the 3% comment above is misleading.  We need to rerun the benchmarks.

If we're at a real impasse here, it's possible to have the servers support both modes of operation.  I think this is confusing and would rather not, but there is a compelling case for each.

rossabaker:
I think this is superseded by #2102?

aeons:
Yes, looks like it. Ok to close for me.

rossabaker:
Please reopen if we're wrong.

PR #1777, opened 2018-04-11
Make AsyncHttpClient.apply return an effect

rossabaker:
Includes merge of #1776.

### DIFF
--- a/async-http-client/src/test/scala/org/http4s/client/asynchttpclient/AsyncHttpClientSpec.scala
+++ b/async-http-client/src/test/scala/org/http4s/client/asynchttpclient/AsyncHttpClientSpec.scala
@@ -7,4 +7,4 @@ import cats.effect.{Effect, IO}
 class AsyncHttpClientSpec
     extends ClientRouteTestBattery(
       "AsyncHttpClient",
-      AsyncHttpClient()(Effect[IO], Http4sSpec.TestExecutionContext))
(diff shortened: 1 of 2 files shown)


async-http-client/src/test/scala/org/http4s/client/asynchttpclient/AsyncHttpClientSpec.scala
```diff
@@ -7,4 +7,4 @@ import cats.effect.{Effect, IO}
 class AsyncHttpClientSpec
     extends ClientRouteTestBattery(
       "AsyncHttpClient",
-      AsyncHttpClient()(Effect[IO], Http4sSpec.TestExecutionContext))
+      AsyncHttpClient()(Effect[IO], Http4sSpec.TestExecutionContext).unsafeRunSync)
```
rossabaker: Ick.  It's already this way in blaze.

jmcardon:
Just for record-keeping, [discussion on gitter](https://gitter.im/http4s/http4s?at=5ace49e4080a3850531ffdad) re: wanting to backport

jmcardon:
Should we close this and merge up instead?

jmcardon:
I approved this, since it's definitely better than the status quo, but I wonder whether we're gonna consider the possible `allocate[F]` and `release[F]` or `close[F]` convention?

rossabaker:
Yes, I think we should.  Otherwise, we're just adding a method we'd like to deprecate in blaze.

rossabaker:
Bitrotten.

PR #1778, opened 2018-04-11
Backport Prometheus Server Metrics

ChristopherDavenport:
Would be cool to cut a release with this.

### DIFF
--- a/prometheus-server-metrics/src/main/scala/org/http4s/server/prometheus/PrometheusExportService.scala
+++ b/prometheus-server-metrics/src/main/scala/org/http4s/server/prometheus/PrometheusExportService.scala
@@ -10,20 +10,10 @@ import io.prometheus.client.hotspot._
 import org.http4s._
 import org.http4s.dsl.Http4sDsl
 
-class PrometheusExportService[F[_]: Sync] private (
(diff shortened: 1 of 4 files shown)


prometheus-server-metrics/src/main/scala/org/http4s/server/prometheus/PrometheusExportService.scala
```diff
@@ -0,0 +1,66 @@
...
+import cats.implicits._
+import io.prometheus.client.CollectorRegistry
+import io.prometheus.client.exporter.common.TextFormat
+import io.prometheus.client.hotspot._
+import org.http4s._
+import org.http4s.dsl.Http4sDsl
+
+class PrometheusExportService[F[_]: Sync] private (
+    s: HttpService[F],
+    cr: CollectorRegistry
```
rossabaker: If these are vals, you don't need the defs.

This could be a final class.

ChristopherDavenport: I agree, but unless you want pain on the way forward, seems like that should be done after we have gotten this through.

PR #1412, opened 2017-09-19
WIP Don't merge. Basic Scala.js compilation

cquiroz:
Here is a first stab at getting http4s to run on scala.js. The final objective is to be able to run the http client on js but we are still several steps away from that

This PR updates the build to cross compile some of the subprojects including core, client and some dependencies

Some notes on the changes to the build

* `core` and `tests` have been converted to full cross projects, This is needed to split the parts that contains references to classes not available in scala.js, in particular `File` Thus `EntityEncoder/Decoder` and `Multipart` have been split. This means the sources files have been moved from `<project>/src` to `<project>/shared/src`
* `testing`, `client`, `circe`, `jawn` and `parboiled2` have been converted to Pure cross compiled projects
* A mock of log4s for `scala.js` have been added. It doesn't do anything at the moment but it lets us compile
* Tests for `client` depend on java only libraries so the tests were moved to a project `clientTesting`

TODO:

[ ] Make the tests run on scala.js
[ ] Implement a scala.js based client
[ ] Example of a client on js

I'm posting this for comments but this certainly doesn't do anything useful at the moment. The main challenge is to come up with a client implementation on JS

Issue #1143, opened 2017-04-26
Header reform


rossabaker:
Current header model is bad.  Depends on a global registry, abuses equality, hard to extend.  Another problem to be solved by typeclasses over subtyping?

[insert better description here]

@wedens has published http://lpaste.net/8545691949200310272

rossabaker:
Discussion notes, for those who can't keep up with Gitter:
- That funny looking Sigma typeclass preserves `putHeaders` with typed headers.  It's really smart.
- `Headers` as an `Iterable[Header]` is doomed.  An `Iterable[RawHeader]` is possible, but not necessarily desired.  Maybe we equip it with the important methods, and there's always the underlying `Vector` for full collection use.

bryce-anderson:
I like the typeclass pattern. Probably the primary reason we didn't use it from the start was so that header parsing could be memorized. Has that been abandoned?

rossabaker:
Yes, we'd lose that.  I think the only ones we're likely to refer to multiple times on a regular basis are `Content-Type` and `Content-Length`, which could just as well be memoized on the message that contains them.

ChristopherDavenport:
I'm going to see what can be done. I'm not sure why we have a problem with an iterable at least for RawHeader. I know you guys know this better than me, so I wonder what I'm missing.

As we continue the length, if we get a CRLF that is not followed by LWSP char then we have a header, if we do we can dismiss the CRLF and LWSP and continue the body. The first Null Line  where a null line is `CLRF CRLF`

There may be illegal header interactions but header definitions at their core seem to behave.

Here is a specification from [RFC 822](https://tools.ietf.org/html/rfc822) Page 8 Section 3.2 **Header Field Definitions**

```
field       =  field-name ":" [ field-body ] CRLF

field-name  =  1*<any CHAR, excluding CTLs, SPACE, and ":">

field-body  =  field-body-contents
                    [CRLF LWSP-char field-body]

field-body-contents =
                   <the ASCII characters making up the field-body, as
                    defined in the following sections, and consisting
                    of combinations of atom, quoted-string, and
                    specials tokens, or else consisting of texts>
```

rossabaker:
I think an iterable of `RawHeader` is absolutely possible, and whether one thinks it's a good idea probably correlates pretty well with one's opinion of standard library collections.  I, having a mixed opinion on those, have a mixed opinion here. :)

m-sp:
finally resolved with https://github.com/http4s/http4s/issues/4297

PR #1801, opened 2018-04-24
WIP Don't merge. Scala.js compilation

cquiroz:
I'm reopening this track though we are still a bit far away from getting a version that would link (It compiles though). This effectively reopens #1412 

Some things needed to get it to work:
- [x] Merge #1770 
- [ ] Remove reflection from headers #1143 
- [x] Bring scala-js#3267 to fix parboiled issues
- [x] Implement any missing api on the scala.js side
- [x] Apparently there is a missing method on `java.time`

Once this is complete we can do the actual client. Perhaps [RosHTTP](https://github.com/hmil/RosHTTP) is a good starting point

/cc @oyvindberg @bpholt


oyvindberg:
Yes yes, awesome, we still need to make this happen! :)

AFAIR i had a handful of smaller tasks in my branch which we should still do as prep work for this, including removing reflection, fixing a few failing tests, documenting platform differences, and so on. I'll try to document this somewhere. Do you think we could open a github project for this so we can create some issues for it?

Meanwhile i have been rebasing my branch at https://github.com/oyvindberg/http4s , though it's a month old now again, pff. At least it links, so you can run the tests there.

About the client, it's an interesting question we should discuss! :) First of all, perhaps RosHTTP is perfect. I haven't tried it, but will make sure to have a look a bit later.

The way I have thought about it so far is that I'm not sure that it necessarily makes sense to have exactly the same API for jvm/js, and even frontend/backend js. As long as you end up with a `Client`, we can share the interesting code from there on.

To that end i started to sketch a solution based on `fetch`, where the user provides their own native/polyfilled instance of fetch, and we take it from there. Feel free to have a look at the `fetch-client` folder in my repo, though be advised that it's far from done. For that very reason I'm also not very committed to that approach.

cquiroz:
Could you make PRs to this PR with your bits to start patching the missing bits? In particular, I think you had fixed the `parboiled` issues, or does it require a `scala.js` update?

I'll do the same with #1170 
I want to suggest that whenever we do a patch that needs an update on `master` like for #1170 we squash the commit into one and mark it in the commit message, e.g. `Remove: ..` So we know that commit should be removed at some moment

about the client, I suggested `RosHTTP` as it's the only one I've seen with `scala.js` support but your solution based on `fetch` seems as good.

rossabaker:
What problems is parboiled causing?

cquiroz:
@oyvindberg  I took some of the code in your branch to incorporate it into the PR

cquiroz:
This is now rebased onto master, it compiles fine but the tests (which at some moment ran) cannot link in .js anymore

There is quite a bit of work still for this, being the most critical the use of reflection (in HttpHeaderParser) issue

rossabaker:
Accomplished via #4938

PR #1802, opened 2018-04-24
Use race against the AsyncContext timeout

rossabaker:
Races the service against the async servlet context timeout.  

* Eliminates a verbose and unsafe callback.
* Cancels a losing, potentially expensive response.


rossabaker:
There is more we can do with `Concurrent`.  This is my simple example to spur discussion.  @jmcardon has expressed worry about it due to the inability to implement it in scalaz8 IO.

I think it's a non-starter to forgo the benefits of `Concurrent` altogether, and highly unpalatable to use something like the `Priority` pattern, requiring two paths.  My hope is that all the effects people want to use can be made `Concurrent`, either through clever implementation or negotiation on the laws.  Time is running very short on the latter for cats-effect-1.0.

servlet/src/main/scala/org/http4s/servlet/Http4sServlet.scala
```diff
@@ -86,75 +86,64 @@ class Http4sServlet[F[_]](
...
       optionTSync
         .suspend(serviceFn(request))
         .getOrElse(Response.notFound)
         .recoverWith(serviceErrorHandler(request))
-
     val servletResponse = ctx.getResponse.asInstanceOf[HttpServletResponse]
-    renderResponse(response, servletResponse, bodyWriter)
+    F.race(timeout, response).flatMap {
+      case Left(()) =>
+        // The F.never is so we don't interrupt the rendering of the timeout response
```
jcranky: Which `F.never`?

jcranky: Ok, but from the comment I thought is was already being used here, and I don't see it anywhere :grinning:

jmcardon: We should leave it in IMHO. It _will_ be never very soon.

rossabaker:
Hell.  This one is broken by #1887, too.

Issue #1797, opened 2018-04-20
Release strategy: 0.19 and 1.0


rossabaker:
cats-1.0 is here.  cats-effect-1.0 is imminent.  fs2-1.0 is expected quickly thereafter.  Dependency releases historically prompt http4s releases  Let's talk.

Goals:
- A stable release on cats/cats-effect/fs2-1.0 stack ASAP.
- Minimize disruptive upgrade cycles.
- Don't release our 1.0 with known warts.
- Maintain no more than two branches (production release and master).

Proposal:
- A conservative 0.19.0 on fs2-1.0.  EOL 0.18.x at this time.
- A glorious 1.0 when it's done, with scalafix from 0.19.

Big ticket items under consideration which may or may not be considered for 0.19:
- [x] HttpApp and HttpRoutes (#1693) 
- [ ] scala.js support (#956)
- [ ] header reform (#1143)
- [ ] websockets as a response type instead of embedded in `Response` 
- [ ] message typeclass (#1572)
- [ ] elimination of backticks
- [x] cancellation support in blaze-server and servlet
- [x] cancellation support in client
- [ ] URI reform (see sttp interpolator for inspiration)
- [ ] server and client creation that returns a `Resource` (relates to #1025)
- [ ] client form (#1772)
- [ ] investigate fastparse as FP-friendly alternative to parboiled2
- [ ] proxy support in blaze-client #1182
- [ ] websocket support in client
- [x] scala-2.13 support (if the dependencies are there)

Breaking changes already in master.:
- [x] Add accumulating version of circe EntityDecoder #1647
- [x] Add ETag support to StaticFile #1652
- [x] Reintroduce the option for fallthrough for authenticated services #1670
- [x] Separate Cookie into RequestCookie and ResponseCookie #1676
- [x] Add Eq[Uri] instance #1688
- [ ] Deprecate Message#withBody in favor of Message#withEntity. The latter returns a Message[F] rather than an F[Message[F]]. #1694
- [x] Myriad new Arbitrary and Cogen instances #1677
- [x] Add non-deprecated LocationResponseGenerator functions #1715
- [x] Relax constraint on Router from Sync to Monad #1723
- [x] Drop scodec-bits dependency #1732
- [x] Add Show[ETag] instance #1749
- [x] Replace fs2.Scheduler with cats.effect.Timer in Retry #1754
- [x] Remove Sync constraint from EntityEncoder[Multipart] #1762
- [x] async-http-client-2.4.2
[...]

PR #1805, opened 2018-04-25
Bump version to 1.0.0-SNAPSHOT

rossabaker:
See #1797.


version.sbt
```diff
@@ -1 +1 @@
-version in ThisBuild := "0.19.0-SNAPSHOT"
+version in ThisBuild := "1.0.0-SNAPSHOT"
```
jcranky: Yey, great times are coming :)

PR #1809, opened 2018-04-26
MimeDB generator

cquiroz:
As part of #1770 the `MediaTypes` are automatically generated. This PR brings the generator code to the project. Note that it is expected that the generation is done manually, this is not included in the build per se.

Some notes:
* I tweaked a bit the generation code to get the main types sorted alphabetically, thus I'm submitting an updated `MimeDB.scala`
* Generation is done with the help of [treehugger](https://github.com/eed3si9n/treehugger) hence it is added as a dependency
* I added this a top-level subproject but perhaps it could be considered an example.
* The generated file is not formatted with `scalafmt`, instead it is expected that will be done manually


mimedb-generator/src/main/scala/mimegenerator/MimeGenerator.scala
```diff
@@ -0,0 +1,287 @@
...
+
+/**
+  * MimeLoader is able to generate a scala file with a database of MediaTypes.
+  * The list of MediaTypes is produced from the list published by `mime-db` in json format.
+  * This json file is parsed, converted to a list of MediaTypes grouped by main type and
+  * converted to a Treehugger syntax tree which is later printed as a scala source file
+  */
+object MimeLoader {
+  implicit val MimeDescrDecoder: Decoder[MimeDescr] = deriveDecoder[MimeDescr]
+  val url = "https://cdn.rawgit.com/jshttp/mime-db/master/db.json"
```
rossabaker: Yeah, I hate our Uri type, too.

mimedb-generator/src/main/scala/mimegenerator/MimeGenerator.scala
```diff
@@ -0,0 +1,287 @@
...
+  // Due to the limits on the jvm class size (64k) we cannot put all instances in one object
+  // This particularly affects `application` which needs to be divided in 2
+  val maxSizePerSection = 500
+  val readMimeDB: Stream[IO, List[Mime]] =
+    for {
+      client <- Http1Client.stream[IO]()
+      _ <- Stream.emit(println(s"Downloading mimedb from $url"))
```
rossabaker: `Stream.eval(IO(println("good habit")))`

cquiroz: i'm still learning fs2. What's the difference in this case?

rossabaker: `emit` is strict, so the side effect runs when the stream is constructed.

```scala
scala> val s = Stream.emit(println("Hi")).covary[IO].compile.toList
Hi
s: cats.effect.IO[List[Unit]] = IO$757370544

scala> s.unsafeRunSync()
res5: List[Unit] = List(())
```

 `eval` embeds the effect into the stream, so it's not run until a value is pulled.

```scala
scala> val s = Stream. eval(IO(println("Hi"))).compile.toList
s: cats.effect.IO[List[Unit]] = IO$560520675

scala> s.unsafeRunSync()
Hi
res7: List[Unit] = List(())
```

`eval_` is be even better: instead of emitting a unit, it emits nothing.  This lets you intersperse effects without polluting your output:

```scala
scala> val s = (Stream(1) ++ Stream.eval_(IO(println("Hi"))) ++ Stream(2)).compile.toList
s: cats.effect.IO[List[Int]] = IO$265737355

scala> s.unsafeRunSync()
Hi
res11: List[Int] = List(1, 2)
```

PR #1814, opened 2018-04-27
Add Config Option for Query Encoding Behavior

ChristopherDavenport:
This is a weird contained concept. Not sure if I like it.

We create a companion config with a provided default. Not sure what outr compatibility story is.

### DIFF
--- a/core/src/main/scala/org/http4s/Query.scala
+++ b/core/src/main/scala/org/http4s/Query.scala
@@ -17,7 +17,7 @@ import scala.collection.mutable.ListBuffer
   * When rendered, the resulting `String` will have the pairs separated
   * by '&' while the key is separated from the value with '='
   */
-final class Query private (pairs: Vector[KeyValue])
+final class Query private (pairs: Vector[KeyValue], queryEncodeParams: Query.QueryEncodeParams = Query.QueryEncodeParams.Default)
     extends IndexedSeq[KeyValue]
     with IndexedSeqOptimized[KeyValue, Query]
     with QueryOps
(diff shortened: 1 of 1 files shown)


core/src/main/scala/org/http4s/Query.scala
```diff
@@ -17,7 +17,7 @@ import scala.collection.mutable.ListBuffer
   * When rendered, the resulting `String` will have the pairs separated
   * by '&' while the key is separated from the value with '='
   */
-final class Query private (pairs: Vector[KeyValue])
+final class Query private (pairs: Vector[KeyValue], queryEncodeParams: Query.QueryEncodeParams = Query.QueryEncodeParams.Default)
```
rossabaker: This could be kept bincompat with a secondary constructor, although it's private anyway.

core/src/main/scala/org/http4s/Query.scala
```diff
@@ -106,14 +106,43 @@ object Query {
 
   val empty: Query = new Query(Vector.empty)
 
+  final class QueryEncodeParams private (
```
rossabaker: Why are only these four special?  [This post](https://bugzilla.mozilla.org/show_bug.cgi?id=1152455#c6) demonstrates how widely browsers disagree.

My inclination was that we could take a `Char => Boolean`.  But that would permit all sorts of ill-advised renderings as well.

core/src/main/scala/org/http4s/Query.scala
rossabaker: A ramification of this is configurability is that two queries that are equal can render to unequal values.

PR #1816, opened 2018-04-27
Upgrade to scala-2.12.6, drop macro-compat

### LINKED ISSUE #1803: Can we get by without macro-paradise?
macro-paradise has an uncertain future.  What are the ramfiications if we shed it?

rossabaker:
Macro Paradise is not yet available for 2.12.6 and won't be for 2.13.  It appears we only needed it to bridge the macro gap to Scala 2.10, which is no longer an issue.

Fixes #1803.

### DIFF
--- a/website/src/hugo/content/changelog.md
+++ b/website/src/hugo/content/changelog.md
@@ -9,7 +9,7 @@ ordered chronologically, so each release contains all changes described below
 it.
 
 # v0.18.10-SNAPSHOT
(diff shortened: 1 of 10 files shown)


website/src/hugo/content/changelog.md
```diff
@@ -9,6 +9,7 @@ ordered chronologically, so each release contains all changes described below
 it.
 
 # v0.18.10-SNAPSHOT
+* Eliminate dependnecy on Macro Paradise and macro-compat [#1816](https://github.com/http4s/http4s/pull/1816)
```
aeons: Typo

PR #1818, opened 2018-04-28
Scala 2.12.4 -> Scala 2.12.6

sullis:


rossabaker:
Thanks.  Macro Paradise isn't published for 2.12.6, so this one isn't simple.  We're already working around this in #1816.

Issue #1817, opened 2018-04-28
Reloading like Play


ScalaWilliam:
I see the following related issues:
- https://github.com/http4s/http4s/issues/766
- https://github.com/http4s/http4s/issues/849

However from my experience with sbt-revolver that's not quite what Play offers.

In development mode, Play starts a server for you, before the sources are even compiled. Only once you hit the server via the web browser, does it compile and start up the application. Likewise, a source code change does not trigger a restart. Once sources are changed, AND you reload, that's when the recompilation happens.

I continue to use Play despite having http4s and akka-http available because I don't enjoy the reloading cycle using eager compilation. Why? I save the sources, Alt-Tab to the browser, refresh, and... bam! "Could not connect to the server". So you then have to line up with the compiler in terms of waiting times, until you can finally Refresh the page.

With Play, you save the code, Alt-Tab, Refresh, and just wait for your page to load. Lower cognitive load, much more pleasant to deal with.

So I'd like to keep this issue open, based on the understanding that the others have not answered the question exactly. Play starts a server in background and recompiles on demand.

While I am all for libraries, this one thing that is enforced by the fact that Play is a framework makes it an absolute delight to use, despite its non-uniform non-FP nature.

---

I'm even wondering if it's possible to embed an http4s route in Play. That'd provide an incremental adoption from Play crowd of http4s - instead of having to throw away all your Play code in one big go. Try with one endpoint, then two, and eventually all. 

---

```
--- (Running the application, auto-reloading is enabled) ---

[info] p.c.s.AkkaHttpServer - Listening for HTTP on /0:0:0:0:0:0:0:0:9000

(Server started, use Enter to stop and go back to the console...)

[info] Compiling 1 Scala source to /.../target/scala-2.12/classes ...
[info] Done compiling.
[info] play.api.Play - Application started (Dev)
[info] Compiling 1 Scala source to /.../target/scala-2.12/classes ...
[info] Done compiling.

--- (RELOAD) ---

[info] play.api.Play - Application started (Dev)
```

jmcardon:
I think you have a fair point. That said, this is on the sbt side of things. If someone is willing to pick up the mantle of adapting whatever plugin play is using, that would be cool.

rossabaker:
This idea has some detractors, and we'd want it to be opt-in.  Closing for lack of progress, but if anyone is still passionate about it and intends to work on it, we can revisit.

ScalaWilliam:
[...]

PR #1819, opened 2018-04-29
[WiP] Play route support

ScalaWilliam:
Related: https://github.com/http4s/http4s/issues/1817

**Rationale**: for applications currently written in Play, there is no straightforward pathway to using http4s. Play also has the hot reloading feature that is exclusive to it, among other nice useful features.

**What it may achieve**: Increased adoption of http4s.

Since Play is a framework, we should not wrap it but provide something to it instead. The Play way is a `Route = PartialFunction[RequestHeader, Action]`, where `Action` is an asynchronous handler.

Here, I map http4s's `HttpService` to Play's `Route`. In order to test it, I have a `PlayTestServerBuilder` that actually runs the `PlayRouteBuilder`. In production usage, this route can be easily be put alongside existing routes, thus providing us with both benefits of Play and http4s, depending on the situation.

It's my first time using `cats`, `cats-effect`, `http4s` and `fs2`. Currently, I have two issues with this change-set and would like the following help on two things first:
**1. The unit test is failing with a strange error, which looks like this:**
```
[info] PlayServerSpec
[info] A server should
[error]   x route requests on the service executor (1 second, 629 ms)
[error]    application-akka.actor.default-dispatcher-10 doesn't start with 'http4s-spec-' (ServerSpec.scala:55)
[error] org.http4s.server.ServerSpec.$anonfun$$init$$2(ServerSpec.scala:55)
[error]   x execute the service task on the service executor (1 second, 684 ms)
[error]    application-akka.actor.default-dispatcher-3 doesn't start with 'http4s-spec-' (ServerSpec.scala:59)
[error] org.http4s.server.ServerSpec.$anonfun$$init$$5(ServerSpec.scala:59)
[info]   + be able to echo its input (1 second, 837 ms)
```
The reason why I'm not solving it myself is because I receive this strange result from `org.http4s.Response#body` (add `.map{x => println(new String(Array(x))); x}` right after `request.body` before calling `convertStream`). Maybe someone's seen this before? Why would thread names be sent as part of response body, I wouldn't know... I think I may have missed something up :-|

2. To make it prettier from FP perspective. I am 100% aware that I am using some shortcuts to "just get the job done", but I am not experienced enough in this style of Scala to know better yet.

What is missing:
- [ ] detailed testing
- [ ] documentation and use cases
- [ ] code-level documentation (references to Play docs for example)
- [x] example code (we would need it, right? Play dependency is not small for the example code)
- [ ] clean, tidy code - it's messy :-)
- [ ] mapping Play attributes and other considerations
- [ ] Tidying up the commit history

I will likely use it for my own project [ActionFPS](https://github.com/ScalaWilliam/ActionFPS) which currently doesn't scale that well code-wise (it's not composed well any more)

I'm fine if this doesn't get accepted eventually, so fire away anyway as I like to learn :-) I received plenty of help from @SystemFw so thank you so much.

Run the example with `examples-play/run`. Visit eg http://localhost:9000/science/date

### DIFF
--- a/play-route/src/main/scala/org/http4s/server/play/PlayRouteBuilder.scala
+++ b/play-route/src/main/scala/org/http4s/server/play/PlayRouteBuilder.scala
@@ -8,7 +8,17 @@ import fs2.Chunk
 import fs2.interop.reactivestreams._
 import org.http4s.server.play.PlayRouteBuilder.{PlayAccumulator, PlayRouting, PlayTargetStream}
 import org.http4s.util.CaseInsensitiveString
-import org.http4s.{EmptyBody, Header, Headers, HttpService, Method, Request, Response, Uri}
+import org.http4s.{
+  EmptyBody,
+  EntityBody,
+  Header,
+  Headers,
+  HttpService,
+  Method,
+  Request,
+  Response,
+  Uri
+}
 import play.api.http.HttpEntity.Streamed
 import play.api.libs.streams.Accumulator
 import play.api.mvc._
@@ -38,13 +48,11 @@ class PlayRouteBuilder[F[_]](
       body = EmptyBody
     )
 
-  type ResponseStream = fs2.Stream[F, Byte]
-
-  def convertStream(responseStream: ResponseStream): PlayTargetStream = {
-    val entityBody: fs2.Stream[F, Byte] = responseStream
+  def convertStream(responseStream: EntityBody[F]): PlayTargetStream = {
+    val entityBody: fs2.Stream[F, ByteString] =
+      responseStream.chunks.map(chunk => ByteString(chunk.toArray))
     Source
       .fromPublisher(entityBody.toUnicastPublisher())
-      .map(byte => ByteString(byte))
   }
 
   def effectToFuture[T](eff: F[T]): Future[T] = {
@@ -69,7 +77,7 @@ class PlayRouteBuilder[F[_]](
     */
   def playRequestToPlayResponse(requestHeader: RequestHeader, method: Method): PlayAccumulator = {
     val sink: Sink[ByteString, Future[Result]] = {
-      Sink.asPublisher[ByteString](false).mapMaterializedValue { publisher =>
+      Sink.asPublisher[ByteString](fanout = false).mapMaterializedValue { publisher =>
         val requestBodyStream: fs2.Stream[F, Byte] =
           publisher.toStream().flatMap(bs => fs2.Stream.chunk(Chunk.bytes(bs.toArray)))
 
--- a/play-route/src/main/scala/org/http4s/server/play/PlayRouteBuilder.scala
+++ b/play-route/src/main/scala/org/http4s/server/play/PlayRouteBuilder.scala
@@ -12,6 +12,7 @@ import org.http4s.{EmptyBody, Header, Headers, HttpService, Method, Request, Res
 import play.api.http.HttpEntity.Streamed
 import play.api.libs.streams.Accumulator
 import play.api.mvc._
+import cats.syntax.all._
 
 import scala.concurrent.duration.Duration
 import scala.concurrent.{Await, ExecutionContext, Future, Promise}
@@ -46,6 +47,19 @@ class PlayRouteBuilder[F[_]](
       .map(byte => ByteString(byte))
   }
 
+  def effectToFuture[T](eff: F[T]): Future[T] = {
+    val promise = Promise[T]
+    F.runAsync(eff) {
+        case Left(bad) =>
+          IO(promise.failure(bad))
+        case Right(good) =>
+          IO(promise.success(good))
+      }
+      .unsafeRunSync()
+
+    promise.future
+  }
+
   /**
     * A Play accumulator Sinks HTTP data in, and then pumps out a future of a Result.
     * That Result will have a Source as the response HTTP Entity.
@@ -59,11 +73,11 @@ class PlayRouteBuilder[F[_]](
         val requestBodyStream: fs2.Stream[F, Byte] =
           publisher.toStream().flatMap(bs => fs2.Stream.chunk(Chunk.bytes(bs.toArray)))
 
+        val http4sRequest = requestHeaderToRequest(requestHeader).withBodyStream(requestBodyStream)
+
         /** The .get here is safe because this was already proven in the pattern match of the caller **/
-        val wrappedResponse: F[Response[F]] =
-          F.map(unwrappedRun(
-            requestHeaderToRequest(requestHeader).withBodyStream(requestBodyStream)).value)(_.get)
-        val wrappedResult: F[Result] = F.map(wrappedResponse) { response =>
+        val wrappedResponse: F[Response[F]] = unwrappedRun(http4sRequest).value.map(_.get)
+        val wrappedResult: F[Result] = wrappedResponse.map { response =>
           Result(
             header = convertResponseToHeader(response),
             body = Streamed(
@@ -94,14 +108,11 @@ class PlayRouteBuilder[F[_]](
     ResponseHeader(
       status = response.status.code,
       headers = response.headers.collect {
-        case header
-            if !PlayRouteBuilder.AkkaHttpSetsSeparately.exists(name =>
-              header.name == CaseInsensitiveString(name)) =>
+        case header if !PlayRouteBuilder.AkkaHttpSetsSeparately.contains(header.name) =>
           header.parsed.name.value -> header.parsed.value
       }.toMap
     )
 
-  /** Big big hack :-( Teach me! **/
   def routeMatches(requestHeader: RequestHeader): Boolean = {
     val optionalResponse: OptionT[F, Response[F]] =
       unwrappedRun.apply(requestHeaderToRequest(requestHeader))
@@ -146,6 +157,7 @@ object PlayRouteBuilder {
       Function.unlift(prefixed.lift.andThen(_.flatMap(t.lift)))
     }
 
-  val AkkaHttpSetsSeparately = Set("Content-Type", "Content-Length", "Transfer-Encoding")
+  val AkkaHttpSetsSeparately: Set[CaseInsensitiveString] =
+    Set("Content-Type", "Content-Length", "Transfer-Encoding").map(CaseInsensitiveString.apply)
 
 }
--- a/play-route/src/main/scala/org/http4s/server/play/PlayRouteBuilder.scala
+++ b/play-route/src/main/scala/org/http4s/server/play/PlayRouteBuilder.scala
@@ -24,10 +24,10 @@ class PlayRouteBuilder[F[_]](
   type UnwrappedKleisli = Request[F] => OptionT[F, Response[F]]
   private[this] val unwrappedRun: UnwrappedKleisli = service.run
 
-  def requestHeaderToRequest(requestHeader: RequestHeader): Request[F] =
+  def requestHeaderToRequest(requestHeader: RequestHeader, method: Method): Request[F] =
     Request(
       // todo fix the ???
-      method = Method.fromString(requestHeader.method).getOrElse(???),
+      method = method,
       uri = Uri(path = requestHeader.uri),
       headers = Headers.apply(requestHeader.headers.toMap.toList.flatMap {
         case (headerName, values) =>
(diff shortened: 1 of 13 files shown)


play-route/src/main/scala/org/http4s/server/play/PlayRouteBuilder.scala
```diff
@@ -0,0 +1,149 @@
...
+    service: HttpService[F]
+)(implicit F: Effect[F], executionContext: ExecutionContext) {
+
+  type UnwrappedKleisli = Request[F] => OptionT[F, Response[F]]
+  private[this] val unwrappedRun: UnwrappedKleisli = service.run
+
+  def requestHeaderToRequest(requestHeader: RequestHeader): Request[F] =
+    Request(
+      // todo fix the ???
+      method = Method.fromString(requestHeader.method).getOrElse(???),
```
rossabaker: You can make this whole method return a `ParseResult[Request[F]]` and deal with errors closer to the seem with Play.  That would be the more FP way.  Or you might just throw an exception here, which (I think) is more the Play way.  Since you're fusing idioms, there's no obvious right answer to me.

play-route/src/main/scala/org/http4s/server/play/PlayRouteBuilder.scala
```diff
@@ -0,0 +1,149 @@
...
+      body = EmptyBody
+    )
+
+  type ResponseStream = fs2.Stream[F, Byte]
+
+  def convertStream(responseStream: ResponseStream): PlayTargetStream = {
+    val entityBody: fs2.Stream[F, Byte] = responseStream
+    Source
+      .fromPublisher(entityBody.toUnicastPublisher())
+      .map(byte => ByteString(byte))
```
rossabaker: This is going to be a stream of single -byte `ByteString`s.  You might want to call `stream.chunks.map(chunk => ByteString(chunk.toArray))` or similar to preserve chunkiness from one stream library to the other.

play-route/src/main/scala/org/http4s/server/play/PlayRouteBuilder.scala
```diff
@@ -0,0 +1,149 @@
...
+      headers = Headers.apply(requestHeader.headers.toMap.toList.flatMap {
+        case (headerName, values) =>
+          values.map { value =>
+            Header(headerName, value)
+          }
+      }),
+      body = EmptyBody
+    )
+
+  type ResponseStream = fs2.Stream[F, Byte]
```
rossabaker: This is already called `EntityBody` in http4s-core, if you want to reuse that.

play-route/src/main/scala/org/http4s/server/play/PlayRouteBuilder.scala
```diff
@@ -0,0 +1,149 @@
...
+    * That Result will have a Source as the response HTTP Entity.
+    *
+    * Here we create a unattached sink, map its materialized value into a publisher,
+    * convert that into an FS2 Stream, then pipe the request body into the http4s request.
+    */
+  def playRequestToPlayResponse(requestHeader: RequestHeader): PlayAccumulator = {
+    val sink: Sink[ByteString, Future[Result]] = {
+      Sink.asPublisher[ByteString](false).mapMaterializedValue { publisher =>
+        val requestBodyStream: fs2.Stream[F, Byte] =
+          publisher.toStream().flatMap(bs => fs2.Stream.fromIterator[F, Byte](bs.toIterator))
```
rossabaker: Maybe something like `Stream.chunk(Chunk.bytes(bs.toArray))`?

play-route/src/main/scala/org/http4s/server/play/PlayRouteBuilder.scala
```diff
@@ -0,0 +1,149 @@
...
+
+        val wrappedResponse: F[Response[F]] =
+          F.map(unwrappedRun(
+            requestHeaderToRequest(requestHeader).withBodyStream(requestBodyStream)).value)(_.get)
```
rossabaker: That `(_.get)` is going to throw if there's no matching route.  There's a `Response.notFound` for a canned 404 response.

Also note that in master, we're moving toward an `HttpApp` type, which will eliminate this complication of optionality.  But you're on the right track for 0.18.x.

play-route/src/main/scala/org/http4s/server/play/PlayRouteBuilder.scala
```diff
@@ -0,0 +1,149 @@
...
+            )
+          )
+        }
+
+        val promise = Promise[Result]
+
+        F.runAsync(wrappedResult) {
+            case Left(bad) =>
+              promise.failure(bad)
+              IO.unit
```
rossabaker: Do the effect in the IO: `IO(promise.failure(bad))`.

play-route/src/main/scala/org/http4s/server/play/PlayRouteBuilder.scala
```diff
@@ -0,0 +1,149 @@
...
+    }
+    Accumulator.apply(sink)
+  }
+
+  def convertResponseToHeader(response: Response[F]): ResponseHeader =
+    ResponseHeader(
+      status = response.status.code,
+      headers = response.headers.collect {
+        case header
+            if !PlayRouteBuilder.AkkaHttpSetsSeparately.contains(header.parsed.name.value) =>
```
rossabaker: I don't think you need to call `parsed` here.  That's for lazily parsing into http4s' richer model, but all you want are String pairs, no?

play-route/src/main/scala/org/http4s/server/play/PlayRouteBuilder.scala
```diff
@@ -0,0 +1,149 @@
...
+    )
+
+  /** Big big hack :-( Teach me! **/
+  def routeMatches(requestHeader: RequestHeader): Boolean = {
+    val optionalResponse: OptionT[F, Response[F]] =
+      unwrappedRun.apply(requestHeaderToRequest(requestHeader))
+    val efff: F[Option[Response[F]]] = optionalResponse.value
+    val completion = Promise[Boolean]()
+    F.runAsync(efff) {
+        case Left(f) => completion.failure(f); IO.unit
```
rossabaker: `IO(completion.failure(f))`

play-route/src/main/scala/org/http4s/server/play/PlayRouteBuilder.scala
```diff
@@ -0,0 +1,149 @@
...
+        case Right(s) => completion.success(s.isDefined); IO.unit
+      }
+      .unsafeRunSync()
+    Await.result(completion.future, Duration.Inf)
```
rossabaker: Does this have to return a `Boolean` instead of a `Future[Boolean]`?  This is going to block not just while the routing is done, but also while the response is generated.  If you can tie it into Play as a `Future`, life is going to be a lot beter.

play-route/src/main/scala/org/http4s/server/play/PlayRouteBuilder.scala
ScalaWilliam: Yep I'll have to flatten out the error somewhere.

play-route/src/main/scala/org/http4s/server/play/PlayRouteBuilder.scala
ScalaWilliam: Unfortunately the route-checking in Play is synchronous (route matcher is a `PartialFunction`, so `.isDefinedAt` is called synchronously). I wonder if there's a better way to do this, cause I really don't like it much myself at all.

ScalaWilliam:
Thanks @rossabaker

play-route/src/main/scala/org/http4s/server/play/PlayRouteBuilder.scala
```diff
@@ -0,0 +1,149 @@
...
+    * convert that into an FS2 Stream, then pipe the request body into the http4s request.
+    */
+  def playRequestToPlayResponse(requestHeader: RequestHeader): PlayAccumulator = {
+    val sink: Sink[ByteString, Future[Result]] = {
+      Sink.asPublisher[ByteString](false).mapMaterializedValue { publisher =>
+        val requestBodyStream: fs2.Stream[F, Byte] =
+          publisher.toStream().flatMap(bs => fs2.Stream.fromIterator[F, Byte](bs.toIterator))
+
+        val wrappedResponse: F[Response[F]] =
+          F.map(unwrappedRun(
```
jmcardon: You might make it easier on yourself to use the `cats.syntax.all._`import, so you can write `unwrappedRun.map`...

play-route/src/main/scala/org/http4s/server/play/PlayRouteBuilder.scala
```diff
@@ -0,0 +1,149 @@
...
+      }.toMap
+    )
+
+  /** Big big hack :-( Teach me! **/
+  def routeMatches(requestHeader: RequestHeader): Boolean = {
+    val optionalResponse: OptionT[F, Response[F]] =
+      unwrappedRun.apply(requestHeaderToRequest(requestHeader))
+    val efff: F[Option[Response[F]]] = optionalResponse.value
+    val completion = Promise[Boolean]()
+    F.runAsync(efff) {
```
jmcardon: This really feels like could could just do `F.runsAsync.unsafetoFuture` and just return that,.

play-route/src/main/scala/org/http4s/server/play/PlayRouteBuilder.scala
jmcardon: The cleanest thing you can do, would be to be able to introspect your route (like in [rho](https://github.com/http4s/rho)) to the extract the endpoints into the routes file play wants, and separate each action endpoint into a play action.

The problem is play routes matching is essentially not directly compatible with http4s without a hacky approach you plug into their routes. That is: Checking routes if they match http4s' routes first, then running the rest. This is not going to be fun to do, at all.

You are better off building a sort of structure from the routes (with `Request[F] => F[Response[F]] => Future[PlayResponse]`), then turning each one into a play action, instead of the whole batch into it.

play-route/src/main/scala/org/http4s/server/play/PlayRouteBuilder.scala
ScalaWilliam: `  def runAsync[A](fa: F[A])(cb: Either[Throwable, A] => IO[Unit]): IO[Unit]` so not sure I can.

ScalaWilliam:
Still having a problem with the tests. It's a bit weird.

```
[http4s] λ play-route/test
[info] PlayServerSpec
[info] A server should
[error]   x route requests on the service executor (784 ms)
[error]    application-akka.actor.default-dispatcher-2 doesn't start with 'http4s-spec-' (ServerSpec.scala:55)
[error] org.http4s.server.ServerSpec.$anonfun$$init$$2(ServerSpec.scala:55)
[info]   + execute the service task on the service executor (824 ms)
[info]   + be able to echo its input (841 ms)
[info] Total for specification PlayServerSpec
[info] Finished in 873 ms
[info] 3 examples, 4 expectations, 1 failure, 0 error
[error] Failed: Total 3, Failed 1, Errors 0, Passed 2
[error] Failed tests:
[error]         org.http4s.server.play.PlayServerSpec
[error] (play-route / Test / test) sbt.TestsFailedException: Tests unsuccessful
[error] Total time: 6 s, completed 29 Apr 2018, 19:44:45
[http4s] λ play-route/test
[info] PlayServerSpec
[info] A server should
[error]   x route requests on the service executor (920 ms)
[error]    application-akka.actor.default-dispatcher-3 doesn't start with 'http4s-spec-' (ServerSpec.scala:55)
[error] org.http4s.server.ServerSpec.$anonfun$$init$$2(ServerSpec.scala:55)
[info]   + execute the service task on the service executor (1 second, 67 ms)
[error]   x be able to echo its input (1 second, 77 ms)
[error]     doesn't start with '{ "Hello": "world" }' (ServerSpec.scala:64)
[error] org.http4s.server.ServerSpec.$anonfun$$init$$8(ServerSpec.scala:64)
[info] Total for specification PlayServerSpec
[info] Finished in 1 second, 95 ms
[info] 3 examples, 4 expectations, 2 failures, 0 error
[error] Failed: Total 3, Failed 2, Errors 0, Passed 1
[error] Failed tests:
[error]         org.http4s.server.play.PlayServerSpec
[error] (play-route / Test / test) sbt.TestsFailedException: Tests unsuccessful
[error] Total time: 5 s, completed 29 Apr 2018, 19:44:53
```

ScalaWilliam:
Closing this as there can be no more progress on this.

I had implemented a library to call Play from http4s instead so that the more powerful thing is in charge. I am using it in actionfps.com but the implementation is quite dirty - but definitely possible.

Issue #1817 is still rather relevant nonetheless.

PR #1820, opened 2018-05-01
copy server logging middleware to client module

bpholt:
The copied files are mostly unchanged, except for updating the package name, running `scalafmt`, and changing the interface of `Logger.apply` to work better with `Client[F]`.

(Hopefully this is set up against the right base branch to be merged and released as part of the 0.18 series.)

### DIFF
--- a/client/src/main/scala/org/http4s/client/middleware/Logger.scala
+++ b/client/src/main/scala/org/http4s/client/middleware/Logger.scala
@@ -1,0 +1,78 @@
+package org.http4s
+package client
+package middleware
+
+import cats.effect._
+import fs2._
+import org.http4s.util.CaseInsensitiveString
+import org.log4s.{Logger ⇒ SLogger}
+
+/**
+  * Simple Middleware for Logging All Requests and Responses
+  */
+object Logger {
+  def apply[F[_]: Effect](
+      logHeaders: Boolean,
+      logBody: Boolean,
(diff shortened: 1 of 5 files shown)


client/src/main/scala/org/http4s/client/middleware/Logger.scala
```diff
@@ -0,0 +1,78 @@
...
+/**
+  * Simple Middleware for Logging All Requests and Responses
+  */
+object Logger {
+  def apply[F[_]: Effect](
+      logHeaders: Boolean,
+      logBody: Boolean,
+      redactHeadersWhen: CaseInsensitiveString => Boolean = Headers.SensitiveHeaders.contains
+  )(client: Client[F]): Client[F] =
+    Client.fromHttpService(
```
bpholt: I'm not sure whether this is the best approach here. It seems to work ok for low-volume clients but the to/from HttpService round-trip definitely adds some overhead. I'm also not sure how to implement this on `master`, which has deprecated `HttpService[F]`. I'd love suggestions on how to remove the `HttpService[F]` interface from these client classes and just deal with `Client[F]`.

rossabaker: It seems highly likely that a client will be based on either `Http[Stream[F, ?], F]` or `Http[Resource[F, ?], F]` in the future.  And then the logging middleware can be built on `Http[G, F]` and be unified for servers and clients.

Issue #1821, opened 2018-05-02
Can't safely shut down TickWheelExecutor


rossabaker:
I tried moving a TickWheelExecutor into the blaze-client, so it could be shut down on the client shutdown, and we don't leak a thread.

Hilarity ensues:

```scala
java.lang.RuntimeException: TickWheelExecutor is shutdown
	at scala.sys.package$.error(package.scala:27)
	at org.http4s.blaze.util.TickWheelExecutor.schedule(TickWheelExecutor.scala:124)
	at org.http4s.blaze.util.TickWheelExecutor.schedule(TickWheelExecutor.scala:82)
	at org.http4s.client.blaze.ClientTimeoutStage.resetTimeout(ClientTimeoutStage.scala:164)
	at org.http4s.client.blaze.ClientTimeoutStage.$anonfun$checkTimeout$2(ClientTimeoutStage.scala:130)
```

This is inspired by @lbialy's work getting things running on graal, and having problems with the statically allocated thread.

rossabaker:
My quick sketch looks like https://github.com/rossabaker/http4s/tree/issue-1821.

armanbilge:
@OstapStrashevskii try putting everything inside of the `resource.use { ... ` block.

PR #1822, opened 2018-05-02
Make the blaze-client tick wheel executor lazy

rossabaker:
Having a resource allocated in an object is something we want to address in #1821.  Meanwhile, @lbialy made this monkey patch for his Graal work, and I don't see any downside for regular use.


lbialy:
:+1:

PR #1826, opened 2018-05-06
Boopickle encoders

cquiroz:
We use [boopickle](https://boopickle.suzaku.io/) on our application with the encoders in this PR. I thought this could be generally useful in http4s though I'm not sure how widely used `boopickle` is.

This is also an opportunity, to receive any criticism you may have to my implementation. In particular I'm not sure if there is enough testing

I couldn't find an aggregate on `build.sbt` to include this into the overall build. Is it needed?

### DIFF
--- a/boopickle/src/main/scala/org/http4s/booPickle/BooPickleInstances.scala
+++ b/boopickle/src/main/scala/org/http4s/booPickle/BooPickleInstances.scala
@@ -1,0 +1,53 @@
+package org.http4s
+package booPickle
+
+import boopickle.Default._
+import boopickle.Pickler
+import cats.Applicative
+import cats.effect.Sync
+import fs2.Chunk
+import java.nio.ByteBuffer
+import org.http4s._
+import org.http4s.EntityEncoder.chunkEncoder
+import org.http4s.headers.`Content-Type`
+import scala.util.{Failure, Success}
+
+/**
+  * Generic factories for http4s encoders/decoders for boopickle
+  * Note that the media type is set for application/octet-stream
+  */
+trait BooPickleInstances {
+
+  private def booDecoderByteBuffer[F[_]: Sync, A](msg: Message[F])(
+      implicit pickler: Pickler[A]): DecodeResult[F, A] =
+    EntityDecoder.collectBinary(msg).flatMap { segment =>
+      val bb = ByteBuffer.wrap(segment.force.toArray)
+      if (bb.hasRemaining) {
+        Unpickle[A](pickler).tryFromBytes(bb) match {
+          case Success(bb) =>
+            DecodeResult.success[F, A](bb)
+          case Failure(pf) =>
+            DecodeResult.failure[F, A](MalformedMessageBodyFailure("Invalid binary body", Some(pf)))
+        }
+      } else {
+        DecodeResult.failure[F, A](MalformedMessageBodyFailure("Invalid binary: empty body", None))
+      }
+    }
+
+  /**
+    * Create an `EntityDecoder` for `A` given a `Pickler[A]`
+    */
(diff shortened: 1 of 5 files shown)


boopickle/src/main/scala/org/http4s/booPickle/BooPickleInstances.scala
```diff
@@ -0,0 +1,53 @@
...
+      } else {
+        DecodeResult.failure[F, A](MalformedMessageBodyFailure("Invalid binary: empty body", None))
+      }
+    }
+
+  /**
+    * Create an `EntityDecoder` for `A` given a `Pickler[A]`
+    */
+  def booOf[F[_]: Sync, A: Pickler]: EntityDecoder[F, A] =
+    EntityDecoder.decodeBy(MediaType.`application/octet-stream`)(booDecoderByteBuffer[F, A])
```
rossabaker: This is going to be hard to use, like `jsonOf` became hard to use in 0.18, isn't it?

boopickle/src/test/scala/org/http4s/booPickle/BoopickleSpec.scala
```diff
@@ -0,0 +1,72 @@
...
+
+  "booOf" should {
+    "decode a class from a boopickle decoder" in {
+      val result = booOf[IO, Fruit]
+        .decode(Request[IO]().withBody(Banana(10.0): Fruit).unsafeRunSync, strict = true)
+      result.value.unsafeRunSync must_== Right(Banana(10.0))
+    }
+  }
+
+  checkAll("EntityCodec[IO, Json]", EntityCodecTests[IO, Fruit].entityCodec)
```
rossabaker: Label is wrong on this test

build.sbt
```diff
@@ -175,6 +175,15 @@ lazy val argonaut = libraryProject("argonaut")
...
   .dependsOn(core, testing % "test->test", jawn % "compile;test->test")
 
+lazy val boopickle = libraryProject("boopickle")
+  .settings(
+    description := "Provides Boopickle codecs for http4s",
+    libraryDependencies ++= Seq(
+      Http4sPlugin.boopickle
+    )
+  )
+  .dependsOn(core, testing % "test->test", jawn % "compile;test->test")
```
rossabaker: Why does this module depend on jawn?

boopickle/src/main/scala/org/http4s/booPickle/BooPickleInstances.scala
cquiroz: I’m not sure what you mean, could you elaborate?

cquiroz: I see. I noted `decodeJson` was on `Message` but I assumed it was grandfathered. OTOH I don't think you want to keep adding `decodeXXX` methods. Or perhaps they  can be extension methods in the bundle?

rossabaker:
New modules need to opt out of MiMa checking, because there isn't a prior version to check against.  I think there is an example of this in the Netty PR.

build.sbt
```diff
@@ -176,6 +176,16 @@ lazy val argonaut = libraryProject("argonaut")
   )
   .dependsOn(core, testing % "test->test", jawn % "compile;test->test")
 
+lazy val boopickle = libraryProject("boopickle")
+  .settings(
+    description := "Provides Boopickle codecs for http4s",
+    libraryDependencies ++= Seq(
+      Http4sPlugin.boopickle
+    ),
+    mimaPreviousArtifacts := Set.empty
```
rossabaker: Just remember to remove this after we release 0.18.11.

PR #1828, opened 2018-05-07
Prometheus Route Tagging

ChristopherDavenport:
So I have had requests for route tagging as per the RED method which does per route metrics. 

I implemented what the solution could look like but extremely ugly.

Thoughts appreciated as I hate the current design.

### DIFF
--- a/prometheus-server-metrics/src/main/scala/org/http4s/server/prometheus/PrometheusMetrics.scala
+++ b/prometheus-server-metrics/src/main/scala/org/http4s/server/prometheus/PrometheusMetrics.scala
@@ -103,15 +103,15 @@ object PrometheusMetrics {
       _ <- emptyResponseHandler.traverse_(status =>
         Sync[F].delay {
           serviceMetrics.requestDuration
-            .labels(reportMethod(m), ServingPhase.report(ServingPhase.HeaderPhase), serviceMetrics.tags.values.toSeq : _*)
+            .labels(Seq(reportMethod(m), ServingPhase.report(ServingPhase.HeaderPhase)) ++ serviceMetrics.tags.values.toSeq : _*)
             .observe(SimpleTimer.elapsedSecondsFromNanos(start, headerTime))
         
           serviceMetrics.requestDuration
-            .labels(reportMethod(m), ServingPhase.report(ServingPhase.BodyPhase), serviceMetrics.tags.values.toSeq : _*)
+            .labels(Seq(reportMethod(m), ServingPhase.report(ServingPhase.BodyPhase)) ++ serviceMetrics.tags.values.toSeq : _*)
             .observe(SimpleTimer.elapsedSecondsFromNanos(start, now))
 
           serviceMetrics.requestCounter
-            .labels(reportMethod(m), reportStatus(status), serviceMetrics.tags.values.toSeq : _*)
+            .labels(Seq(reportMethod(m), reportStatus(status)) ++ serviceMetrics.tags.values.toSeq : _*)
             .inc()
           
       })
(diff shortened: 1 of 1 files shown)


prometheus-server-metrics/src/main/scala/org/http4s/server/prometheus/PrometheusMetrics.scala
```diff
@@ -213,22 +239,24 @@ object PrometheusMetrics {
   def apply[F[_]: Sync](
       c: CollectorRegistry,
       prefix: String = "org_http4s_server",
+      tag: Option[String] = Option.empty[String],
```
rossabaker: Why just one tag?  Why must it be called `"tag"`?  Can you take  `Map[String, String]`?

prometheus-server-metrics/src/main/scala/org/http4s/server/prometheus/PrometheusMetrics.scala
```diff
@@ -101,19 +102,44 @@ object PrometheusMetrics {
...
       _ <- emptyResponseHandler.traverse_(status =>
         Sync[F].delay {
-          serviceMetrics.requestDuration
+          serviceMetrics.tag.fold(
+            serviceMetrics.requestDuration
             .labels(reportMethod(m), ServingPhase.report(ServingPhase.HeaderPhase))
             .observe(SimpleTimer.elapsedSecondsFromNanos(start, headerTime))
+          )(tag => 
+            serviceMetrics.requestDuration
+            .labels(reportMethod(m), ServingPhase.report(ServingPhase.HeaderPhase), tag)
```
rossabaker: If `tags` is a `Map` (see suggestion below), can you just call it `tags.values: _*` here and get rid of some of the duplication?

Even as one, can you *gulp* use that horrid option2Iterable implicit and get the same result?

fiadliel:
As I understand it, this is per-service tagging, rather than per-route tagging?

I have done per-route tagging before, but it required manually tagging each response with the "path" used, which is more manual labour, and perhaps more invasive to the API than you'd want?

(I could imagine a macro existing that can figure out an appropriate tag for a limited selection of services, but the general case requires a reasonable amount of work by the user).

rossabaker:
My understanding of Prometheus is that you want to be very careful with cardinality: more unique sets of tags adds more time series for the server to track.  This is what scared me off per-route tagging in the past, even though I've wanted it.

It's worth thinking about how we'd let people do that, without forcing them to do that, because it's both useful and expensive.

rossabaker:
Interseting.  Our routes are more opaque, so I don't know how we could match Play there.  But that's something to keep in mind for people who are thinking about alternate routing DSLs.

prometheus-server-metrics/src/main/scala/org/http4s/server/prometheus/PrometheusMetrics.scala
```diff
@@ -102,18 +103,24 @@ object PrometheusMetrics {
       _ <- emptyResponseHandler.traverse_(status =>
         Sync[F].delay {
```
barambani: I very minor thing, would it help to use an implicit parameter here like

```scala
private def onEmpty[F[_]](
  m: Method,	       m: Method,
  start: Long,	       start: Long,
  headerTime: Long,	       headerTime: Long,
  serviceMetrics: ServiceMetrics,	       serviceMetrics: ServiceMetrics,
  emptyResponseHandler: Option[Status]	       emptyResponseHandler: Option[Status]
)(implicit F: Sync[F]): F[Unit] =
```

it could save some implicit search in positions like

```scala
now <- Sync[F].delay(System.nanoTime)
```
?
apologies if it has been already discussed

barambani: so I guess here the preference is not to use something like 
```scala
now <- F.delay(System.nanoTime)
```
but is to use the implicit search of the `apply` on `Sync[F]`. 
Thanks 👍

ChristopherDavenport: Well it's my preference as I like to clarify which constraints I'm using. Which allows me to more easily see when I can upon removing a constrain use downgrade. Its just my personal practice.

barambani: makes sense to me. Thanks 👍

ChristopherDavenport:
Not sure if this will work. As multiple registration of the same metric into a collector generates an error. This may just allow more configuration rather than solve the per-route/service metric problem.

rossabaker:
Ugh.  Does it allow that and then fail on duplicate at scrape time?  So it's even worse than throwing when you duplicate?  That happened to me once.

ChristopherDavenport:
Well going to defer to the new metrics attribute methods.

PR #1829, opened 2018-05-07
WIP: ParseResult Alias Killing

ChristopherDavenport:
Killing off Aliases which obscure types for what I don't think is a major advantage of a shorter name. Precision in communication I think is advantageous here.

### DIFF
--- a/core/src/main/scala/org/http4s/package.scala
+++ b/core/src/main/scala/org/http4s/package.scala
@@ -15,6 +15,7 @@ package object http4s { // scalastyle:ignore
 
   type DecodeResult[F[_], A] = EitherT[F, DecodeFailure, A]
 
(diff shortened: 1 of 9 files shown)


core/src/main/scala/org/http4s/package.scala
```diff
@@ -15,6 +15,7 @@ package object http4s { // scalastyle:ignore
 
   type DecodeResult[F[_], A] = EitherT[F, DecodeFailure, A]
 
+  @deprecated("Deprecated in favor of using Either", "0.19")
```
rossabaker: That branch is called 1.0 now.

rossabaker:
Ugh.  Nope.  I got distracted by the green above.  My fault. :(

PR #1830, opened 2018-05-07
Add BlockingHttp4sServlet

Taig:
The current `Http4sServlet` [uses](https://github.com/http4s/http4s/blob/release-0.18.x/servlet/src/main/scala/org/http4s/servlet/Http4sServlet.scala#L70) the async `Servlet` feature in its implementation. Unfortunately this is [not supported](https://issuetracker.google.com/issues/65104189) in the Google App Engine (GAE) standard environment so that a blocking implementation is necessary.

Consider this PR as an initial draft. I started off by copying the `Http4sServlet` and ripping it apart. So there is currently quite a lot of code duplication which should be improved.

Please note that I have never worked with `Servlet` nor with `cats.Effect` before, so unfortunately some of the changes here were a bit of guesswork for me.

### DIFF
--- a/servlet/src/main/scala/org/http4s/servlet/BlockingHttp4sServlet.scala
+++ b/servlet/src/main/scala/org/http4s/servlet/BlockingHttp4sServlet.scala
@@ -49,8 +49,8 @@ class BlockingHttp4sServlet[F[_]](
       )
 
       F.runAsync(render) {
-          case Right(_) => Sync[IO].unit
-          case Left(t) => Sync[IO].delay(errorHandler(servletResponse)(t))
+          case Right(_) => IO.unit
+          case Left(t) => IO(errorHandler(servletResponse)(t))
         }
         .unsafeRunSync()
     } catch errorHandler(servletResponse)
--- a/servlet/src/main/scala/org/http4s/servlet/BlockingHttp4sServlet.scala
+++ b/servlet/src/main/scala/org/http4s/servlet/BlockingHttp4sServlet.scala
@@ -1,42 +1,17 @@
 package org.http4s
 package servlet
 
-import java.net.InetSocketAddress
-import javax.servlet._
-import javax.servlet.http.{HttpServlet, HttpServletRequest, HttpServletResponse, HttpSession}
+import javax.servlet.http.{HttpServletRequest, HttpServletResponse}
 
 import cats.effect._
 import cats.implicits.{catsSyntaxEither => _, _}
-import org.http4s.headers.`Transfer-Encoding`
 import org.http4s.server._
-import org.log4s.getLogger
-
-import scala.collection.JavaConverters._
 
 class BlockingHttp4sServlet[F[_]](
     service: HttpRoutes[F],
     servletIo: BlockingServletIo[F],
     serviceErrorHandler: ServiceErrorHandler[F])(implicit F: Effect[F])
-    extends HttpServlet {
-  private[this] val logger = getLogger
-
-  private[this] var serverSoftware: ServerSoftware = _
-
-  // micro-optimization: unwrap the service and call its .run directly
-  private[this] val serviceFn = service.run
-
-  object ServletRequestKeys {
-    val HttpSession: AttributeKey[Option[HttpSession]] = AttributeKey[Option[HttpSession]]
-  }
-
-  override def init(config: ServletConfig): Unit = {
-    val servletContext = config.getServletContext
-    val servletApiVersion = ServletApiVersion(servletContext)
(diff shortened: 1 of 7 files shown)


servlet/src/main/scala/org/http4s/servlet/BlockingHttp4sServlet.scala
```diff
@@ -0,0 +1,162 @@
...
+import cats.implicits.{catsSyntaxEither => _, _}
+import org.http4s.headers.`Transfer-Encoding`
+import org.http4s.server._
+import org.log4s.getLogger
+
+import scala.collection.JavaConverters._
+
+class BlockingHttp4sServlet[F[_]](
+    service: HttpRoutes[F],
+    servletIo: BlockingServletIo[F],
```
rossabaker: What's the benefit of parameterizing this?  To configure the chunk size?  I like that it mandates `BlockingServletIo` for this use case.

servlet/src/main/scala/org/http4s/servlet/BlockingHttp4sServlet.scala
```diff
@@ -0,0 +1,162 @@
...
+    try {
+      val bodyWriter = servletIo.initWriter(servletResponse)
+
+      val render = toRequest(servletRequest).fold(
+        onParseFailure(_, servletResponse, bodyWriter),
+        handleRequest(_, servletResponse, bodyWriter)
+      )
+
+      F.runAsync(render) {
+          case Right(_) => Sync[IO].unit
```
rossabaker: `IO.unit`

servlet/src/main/scala/org/http4s/servlet/BlockingHttp4sServlet.scala
```diff
@@ -0,0 +1,162 @@
...
+          case Left(t) => Sync[IO].delay(errorHandler(servletResponse)(t))
```
rossabaker: `IO(errorHandler(...))`

servlet/src/main/scala/org/http4s/servlet/BlockingHttp4sServlet.scala
```diff
@@ -0,0 +1,162 @@
...
+        }
+        .unsafeRunSync()
+    } catch errorHandler(servletResponse)
+
+  private def onParseFailure(
```
rossabaker: I think everything here below is a copy-paste and could move into an abstract class?  Make them `protected`.

servlet/src/main/scala/org/http4s/servlet/BlockingHttp4sServlet.scala
```diff
@@ -0,0 +1,162 @@
...
+import scala.collection.JavaConverters._
+
+class BlockingHttp4sServlet[F[_]](
+    service: HttpRoutes[F],
+    servletIo: BlockingServletIo[F],
+    serviceErrorHandler: ServiceErrorHandler[F])(implicit F: Effect[F])
+    extends HttpServlet {
+  private[this] val logger = getLogger
+
+  private[this] var serverSoftware: ServerSoftware = _
```
rossabaker: Could be protected and moved to abstract class.

servlet/src/main/scala/org/http4s/servlet/BlockingHttp4sServlet.scala
```diff
@@ -0,0 +1,162 @@
...
+
+  // micro-optimization: unwrap the service and call its .run directly
+  private[this] val serviceFn = service.run
```
rossabaker: Could be protected and move to abstract class.  Though maybe we save a few nanoseconds going through a field (e.g., `private[this]`) instead of an accessor.

Taig:
Any naming preferences for the abstract class? I would probably name it `Http4sServlet` and extend from `AsyncHttp4sServlet` and `BlockingHttp4sServlet`. But I assume we should not rename the current `Http4sServlet` for compatibility?

servlet/src/main/scala/org/http4s/servlet/BlockingHttp4sServlet.scala
Taig: I wanted to keep the `chunkSize` configurable but also mimic the API of `Http4sSevlet` for consistency

Taig:
I'm fine with keeping my own copy of this around until 0.19 is released, so that's not an issue for me. Let me know if you want the renaming, I don't mind waiting for 0.19.

rossabaker:
Test failure is related, fixed by #1836.

PR #1831, opened 2018-05-08
Netty server backend for http4s.

jmcardon:
# Netty support for http4s

Majority of inspiration + a chunk of netty-helping code comes from the 
[Play Framework netty server backend](https://github.com/playframework/playframework/tree/master/framework/src/play-netty-server).
Credits to them for a lot of the gluework for netty.

## Progress

- [x] Server builder 
- [x] SSL
- [X] MountService 
~~- [ ] (Maybe) Server header~~ Not gonna be supported. Can be added with middleware if needed
- [X] Builder tests
- [X] Tests similar to http1 server stage
- [X] Websocket support
~~- [ ] Websocket tests?~~ Difficult to test at the current moment
- [X] Netty Channel options (Only AUTO_READ must be set as false)


jmcardon:
Note on default values:

`miku` is being added for the sake of leveraging native transports... because speed n' stuff. Thus, I set the default transport option as `Native`, with default fallback onto `nio` in case it's not available.

build.sbt
```diff
@@ -32,7 +32,7 @@ lazy val core = libraryProject("core")
       scalaReflect(scalaOrganization.value, scalaVersion.value) % "provided",
       scodecBits,
       scalaCompiler(scalaOrganization.value, scalaVersion.value) % "provided"
-    ),
+    )
```
rossabaker: MY BEAUTIFUL TRAILING COMMAS!!!

(I like these, because you can add to the list without bringing into the diff things that don't change.  But I'm not going to die on this hill.)

build.sbt
```diff
@@ -112,6 +112,20 @@ lazy val blazeClient = libraryProject("blaze-client")
   )
   .dependsOn(blazeCore % "compile;test->test", client % "compile;test->test")
 
+lazy val mikuServer = libraryProject("miku-server")
```
rossabaker: Alright, I'll be the grump here.  Discoverability is important.  Can a user intuit what exists from naming conventions, scanning the directory structure, etc.?

- Does http4s support circe? http4s-circe, `org.http4s.circe`
- Does it support servlets? http4s-servlet, `org.http4s.servlet`  
- Does it support twirl? http4s-twirl, `org.http4s.twirl`
- Does it support netty servers? http4s-miku, `org.http4s.miku` :confused: 

Now, it would be tragic if your ASCII shrine to miku were lost.  Could we leave it as a whimsical banner in the server package so that the glory of Miku could be shared across all backends?  Use it in a tut for how to configure a server?  I'm all for having a laugh, but let's make sure it's not detracting from a high quality and oft-requested module.
 
:notes: I don't think Hank done it this way :notes:

miku-server/src/main/scala/org/http4s/miku/MikuBuilder.scala
```diff
@@ -0,0 +1,492 @@
...
+import io.netty.handler.timeout.IdleStateHandler
+import org.http4s.HttpService
+import org.http4s.server._
+import org.http4s.util.bug
+import org.log4s.getLogger
+
+import scala.collection.JavaConverters._
+import scala.collection.immutable
+import scala.concurrent.ExecutionContext
+import scala.concurrent.duration.Duration
```
rossabaker: We've been sorting our imports lexicographically.

miku-server/src/main/scala/org/http4s/miku/MikuBuilder.scala
```diff
@@ -0,0 +1,492 @@
...
+
+sealed trait NettyTransport
+case object Jdk extends NettyTransport
```
rossabaker: Doesn't Netty call this `Nio`?

For namespacing, maybe put these in a companion object of `NettyTransport`, since the names don't mean much on their own?

miku-server/src/main/scala/org/http4s/miku/MikuBuilder.scala
```diff
@@ -0,0 +1,492 @@
...
+import org.http4s.server._
+import org.http4s.util.bug
+import org.log4s.getLogger
+
+import scala.collection.JavaConverters._
+import scala.collection.immutable
+import scala.concurrent.ExecutionContext
+import scala.concurrent.duration.Duration
+
+sealed trait NettyTransport
```
rossabaker: I like `with Product with Serializable`, if only because it gets people into the habit where this inference really tends to matter.

miku-server/src/main/scala/org/http4s/miku/MikuBuilder.scala
```diff
@@ -0,0 +1,492 @@
...
+    serviceErrorHandler: ServiceErrorHandler[F],
+    transport: NettyTransport,
+    ec: ExecutionContext,
+    enableWebsockets: Boolean,
+    banner: immutable.Seq[String],
+    nettyChannelOptions: MikuBuilder.NettyChannelOptions
+)(implicit F: Effect[F])
+    extends ServerBuilder[F]
+    with IdleTimeoutSupport[F]
+    with SSLKeyStoreSupport[F]
```
rossabaker: I'm still thinking this needs to go away, but I guess it's consistent with 0.18.

miku-server/src/main/scala/org/http4s/miku/MikuBuilder.scala
```diff
@@ -0,0 +1,492 @@
...
+
+  //Todo: Router-hack thingy
+  def mountService(service: HttpService[F], prefix: String): MikuBuilder[F] =
+    copy(httpService = service)
+
+  def start: F[Server[F]] = F.delay {
+    val eventLoop: MultithreadEventLoopGroup = transport match {
+      case Jdk => new NioEventLoopGroup()
+      case Native =>
+        if (Epoll.isAvailable)
```
rossabaker: So this is how we avoid calamity when we're on a different OS?

miku-server/src/main/scala/org/http4s/miku/MikuBuilder.scala
```diff
@@ -0,0 +1,492 @@
...
+
+    val (_, boundAddress) = startNetty(eventLoop, allChannels)
+    val server = new Server[F] {
+      def shutdown: F[Unit] = F.delay {
+        // First, close all opened sockets
+        allChannels.close().awaitUninterruptibly()
+        // Now shutdown the event loop
+        eventLoop.shutdownGracefully()
+
+        logger.info("Shut down gracefully :)")
```
rossabaker: Maybe we could add the address that's shut down here, in case we have multiple servers running.

miku-server/src/main/scala/org/http4s/miku/MikuBuilder.scala
```diff
@@ -0,0 +1,492 @@
...
+    case MikuSSLContextBits(context, clientAuth) =>
+      (context, clientAuth)
+  }
+
+  /** A stream transformation that registers our channels and
+    * adds the necessary codecs and transport handlers
+    */
+  def channelPipe(
+      eventLoop: MultithreadEventLoopGroup,
+      allChannels: DefaultChannelGroup): Pipe[F, Channel, Unit] = { s =>
```
rossabaker: Does this get anything out of being a `Pipe` rather than a `Channel => F[Unit]`?

miku-server/src/main/scala/org/http4s/miku/MikuBuilder.scala
```diff
@@ -0,0 +1,492 @@
...
+        // Setup the channel for explicit reads
+        connChannel
+          .config()
+          .setOption(ChannelOption.AUTO_READ, java.lang.Boolean.FALSE)
+
+        val pipeline = connChannel.pipeline()
+        getContext() match {
+          case Some((ctx, auth)) => //Ignore client authentication for now
+            val engine = ctx.createSSLEngine()
+            engine.setUseClientMode(false)
```
rossabaker: I've been thinking in the future we might want to configure `SSL` just by offering an `F[SSLEngine]`, but I guess we do need to set `UseClientMode`.  Maybe that's okay anyway.

miku-server/src/main/scala/org/http4s/miku/MikuBuilder.scala
```diff
@@ -0,0 +1,492 @@
...
+      F.delay {
+
+        // Setup the channel for explicit reads
+        connChannel
+          .config()
+          .setOption(ChannelOption.AUTO_READ, java.lang.Boolean.FALSE)
+
+        val pipeline = connChannel.pipeline()
+        getContext() match {
+          case Some((ctx, auth)) => //Ignore client authentication for now
```
rossabaker: How are you ignoring it?

miku-server/src/main/scala/org/http4s/miku/MikuBuilder.scala
```diff
@@ -0,0 +1,492 @@
...
+        }
+
+        // Netty HTTP decoders/encoders/etc
+        pipeline.addLast(
+          "decoder",
+          new HttpRequestDecoder(maxInitialLineLength, maxHeaderSize, maxChunkSize))
+        pipeline.addLast("encoder", new HttpResponseEncoder())
+        pipeline.addLast("decompressor", new HttpContentDecompressor())
+
+        idleTimeout match {
```
rossabaker: Isn't this non-exhaustive on `Duration.NegativeInf` and `Duration.Undefined`?  (Not that anyone is likely to set those, but...) I usually use `isFinite` instead of pattern matching.  I've not seen this idiom.

miku-server/src/main/scala/org/http4s/miku/MikuBuilder.scala
```diff
@@ -0,0 +1,492 @@
...
+        .channel(classOf[NioServerSocketChannel])
+        .group(serverChannelEventLoop))
+      .option(ChannelOption.AUTO_READ, java.lang.Boolean.FALSE) // publisher does ctx.read()
+      .handler(channelPublisher)
+      .localAddress(address)
+
+  protected def getBootstrap(
+      serverChannelEventLoop: EventLoopGroup,
+      channelPublisher: HandlerPublisher[Channel],
+      address: InetSocketAddress): Bootstrap = transport match {
```
rossabaker: This logic looks tangled with the transport logic in `start`.  Is there a good way to do this transport sniffing once?  Maybe it doesn't really matter.

miku-server/src/main/scala/org/http4s/miku/MikuHandler.scala
```diff
@@ -0,0 +1,253 @@
...
+  * A note about "lastResponseSent" to help understand:
+  * By reassigning the variable with a `flatMap` (which doesn't require synchronization at all, since
+  * you can consider this handler essentially single threaded), this means that, we can run the
+  * `handle` action asynchronously by forking it into a thread in `handle`, all the while
+  * ensuring in-order writes for the handler thread by attaching the callback using `flatMap`.
+  * This ensures we can schedule more work asynchronously by streaming `lastResponseSent` like a
+  * FIFO asynchronous queue.
+  *
+  */
+sealed abstract class MikuHandler[F[_]](
```
rossabaker: If you make this package private, it's easier to change it in a binary compatible fashion when the bugs start rolling in.

miku-server/src/main/scala/org/http4s/miku/MikuHandler.scala
```diff
@@ -0,0 +1,253 @@
...
+
+  // Cache the formatter thread locally
+  private[this] val RFC7231InstantFormatter =
+    DateTimeFormatter
+      .ofPattern("EEE, dd MMM yyyy HH:mm:ss zzz")
+      .withLocale(Locale.US)
+      .withZone(ZoneId.of("GMT"))
+
+  // Compute the formatted date string only once per second, and cache the result.
+  // This should help microscopically under load.
```
rossabaker: Is this a technique that's applicable in blaze, too?

miku-server/src/main/scala/org/http4s/miku/NettyModelConversion.scala
```diff
@@ -0,0 +1,311 @@
...
+        headerBuf += Header(mapEntry.getKey, mapEntry.getValue)
+      }
+
+      val method: ParseResult[Method] =
+        Method.fromString(request.method().name())
+      val version: HV = {
+        if (request.protocolVersion() == HttpVersion.HTTP_1_1)
+          HV.`HTTP/1.1`
+        else
+          HV.`HTTP/1.0`
```
rossabaker: You've covered 99.99999999% of people not causing trouble, but, is this right?

project/Http4sPlugin.scala
```diff
@@ -249,54 +249,59 @@ object Http4sPlugin extends AutoPlugin {
...
+  lazy val jspApi                           = "javax.servlet.jsp"      %  "javax.servlet.jsp-api"       % "2.3.1" // YourKit hack
+  lazy val log4s                            = "org.log4s"              %% "log4s"                       % "1.6.1"
+  lazy val logbackClassic                   = "ch.qos.logback"         %  "logback-classic"             % "1.2.3"
+  lazy val metricsCore                      = "io.dropwizard.metrics"  %  "metrics-core"                % "4.0.2"
+  lazy val metricsJson                      = "io.dropwizard.metrics"  %  "metrics-json"                % metricsCore.revision
+  lazy val nettyHandler                     = "io.netty"               % "netty-handler"                % "4.1.24.Final"
+  lazy val nettyHttpCodec                   = "io.netty"               % "netty-codec-http"             % nettyHandler.revision
+  lazy val nettyNativeTransport             = "io.netty"               % "netty-transport-native-epoll" % nettyHandler.revision classifier "linux-x86_64"
+  lazy val nettyUnixCommon                   = "io.netty"              % "netty-transport-native-unix-common" % nettyHandler.revision
```
rossabaker: This is why I hate vertical alignment.

If we're blowing out the diff for netty-transport-native-epoll, might as well go all the way for this one, no?

build.sbt
jmcardon: Fair. i think I can rename to http4s netty and leave the glory of miku in my PR

miku-server/src/main/scala/org/http4s/miku/MikuBuilder.scala
jmcardon: Will do

miku-server/src/main/scala/org/http4s/miku/NettyModelConversion.scala
jmcardon: This codec can't even respond to http 2.0 messages if you wanted it to. It doesn't handle them properly. I can add a test though

miku-server/src/main/scala/org/http4s/miku/MikuHandler.scala
jmcardon: I was thinking if i could port over this performance hack in blaze as well actually. Maybe, if thw threading model is anything like netty, most likely, as this isnt even work lost usually, considering i think 99% of the time, people won't manually add a date header

miku-server/src/main/scala/org/http4s/miku/MikuHandler.scala
jmcardon: That one I meant to do. I forgot. will change

miku-server/src/main/scala/org/http4s/miku/MikuBuilder.scala
jmcardon: I can look to reorder the code for it, sure. I wrote that at 4am

miku-server/src/main/scala/org/http4s/miku/MikuBuilder.scala
jmcardon: Leftover comment from the days of me not using it

miku-server/src/main/scala/org/http4s/miku/MikuBuilder.scala
jmcardon: Absolutely nothing I'm sure. This is a Pipe since it was a translation of play's equivalent of this in akka streams. I didn't give it much thought after initial translation

miku-server/src/main/scala/org/http4s/miku/MikuBuilder.scala
jmcardon: I can look into it for master branch. I can create + self assign a ticket

miku-server/src/main/scala/org/http4s/miku/NettyModelConversion.scala
bryce-anderson: To @rossabaker's point, there is nothing restricting the Netty HTTP verison to 1.0 and 1.1:
```
scala> io.netty.handler.codec.http.HttpVersion.valueOf("HTTP/1.2")
res1: io.netty.handler.codec.http.HttpVersion = HTTP/1.2
```

I can't remember where, but I do recall seeing a HTTP/2.0 floating around at $work.

miku-server/src/main/scala/org/http4s/miku/MikuBuilder.scala
rossabaker: I don't think it hurts anything.  It just looks like more power than you need.

miku-server/src/main/scala/org/http4s/miku/MikuHandler.scala
rossabaker: I think the only time I've ever manually constructed a date header was in a test, trying to fool an expiring token system that I didn't control.

miku-server/src/main/scala/org/http4s/miku/NettyModelConversion.scala
rossabaker: Yes, that's what I was thinking.  I think you'd respond correctly, for free.  It's just about honesty in what we pass to the server.  (This is a really, really stupid use case.  It just caught my eye.)

More realistic, though slightly less in 2018: if the service generates an HTTP/1.0 response, I bet this server is supporting some invalid things.  Blaze is pretty careful about the distinction.

jmcardon: As a counter chess move, I will point out that http2 has [its own module](https://github.com/netty/netty/tree/4.1/codec-http2/src/main/java/io/netty/handler/codec/http2).

I don't think the `ResponseEncoder` will do valid things for the Http2 spec.

build.sbt
aeons: I support re-adding these commas! :)

netty-server/src/main/scala/org/http4s/netty/Http4sNettyHandler.scala
```diff
@@ -67,9 +72,13 @@ sealed abstract class MikuHandler[F[_]](
 
   /**
     * Handle the given request.
-    * Note: Handle implementations fork into user ExecutionContext
+    * Note: Handle implementations fork into user ExecutionContext?
```
rossabaker: Who is the question to?

netty-server/src/main/scala/org/http4s/netty/Http4sNettyHandler.scala
```diff
@@ -91,37 +101,49 @@ sealed abstract class MikuHandler[F[_]](
...
+                ctx
+                  .writeAndFlush(response)
+                  .addListener(new ChannelFutureListener {
+                    def operationComplete(future: ChannelFuture): Unit =
+                      //Run cleanup action on the user-supplied pool.
+                      //We might tweak this later on to run on the same or in a cleanup pool
+                      //Benching this would be interesting
+                      ec.execute(new Runnable {
+                        def run(): Unit =
+                          F.runAsync(cleanup(future.channel()))(_ => IO.unit).unsafeRunSync()
```
rossabaker: cats-effect has `Logger.reportFailure` for this.  We should add that to our own internals.  This `_ => IO.unit` swallows errors.

netty-server/src/main/scala/org/http4s/netty/NettyBuilder.scala
```diff
@@ -254,64 +272,73 @@ class MikuBuilder[F[_]](
   /** A stream transformation that registers our channels and
```
rossabaker: It's not really a stream transformation anymore.  It's just an effectful function.

netty-server/src/main/scala/org/http4s/netty/NettyBuilder.scala
```diff
@@ -254,64 +272,73 @@ class MikuBuilder[F[_]](
...
+
+      // Setup the channel for explicit reads
+      connChannel
+        .config()
+        .setOption(ChannelOption.AUTO_READ, java.lang.Boolean.FALSE)
+
+      val pipeline = connChannel.pipeline()
+      getContext() match {
+        case Some((ctx, auth)) =>
+          val engine = ctx.createSSLEngine()
```
rossabaker: I still think passing around an `F[SSLEngine]` might be the right API for our backends.

netty-server/src/test/scala/org/http4s/netty/NettyIntegrationTests.scala
```diff
@@ -0,0 +1,128 @@
...
+  ResourceLeakDetector.setLevel(ResourceLeakDetector.Level.PARANOID)
+
+  sequential ^ ("Http4s Netty server integration test" should {
+
+    "Properly return the ping response" in {
+      client.get(server.baseUri / "ping")(r => IO(r.status)).unsafeRunSync() must_== Ok
+    }
+
+    "Return the proper content length for a simple response" in {
+      val length = "waylonnn jenningssss".length
```
rossabaker: Are you sure Hank spelt it this way?

project/Http4sPlugin.scala
```diff
@@ -260,59 +260,60 @@ object Http4sPlugin extends AutoPlugin {
...
+  lazy val json4sCore                       = "org.json4s"             %% "json4s-core"               % "3.5.3"
+  lazy val json4sJackson                    = "org.json4s"             %% "json4s-jackson"            % json4sCore.revision
+  lazy val json4sNative                     = "org.json4s"             %% "json4s-native"             % json4sCore.revision
+  lazy val jspApi                           = "javax.servlet.jsp"      %  "javax.servlet.jsp-api"     % "2.3.1" // YourKit hack
+  lazy val log4s                            = "org.log4s"              %% "log4s"                     % "1.6.1"
+  lazy val logbackClassic                   = "ch.qos.logback"         %  "logback-classic"           % "1.2.3"
+  lazy val metricsCore                      = "io.dropwizard.metrics"  %  "metrics-core"              % "4.0.2"
+  lazy val metricsJson                      = "io.dropwizard.metrics"  %  "metrics-json"              % metricsCore.revision
+  lazy val nettyHandler                     = "io.netty"               % "netty-handler"                      % "4.0.56.Final"
```
rossabaker: Align them all or don't touch them all.

netty-server/src/main/scala/org/http4s/netty/NettyBuilder.scala
jmcardon: fair. I'll make that distinction

netty-server/src/main/scala/org/http4s/netty/NettyBuilder.scala
jmcardon: possibly. I've been considering this too and I Want to expose that on master, but for 0.18.x maybe not.

netty-server/src/main/scala/org/http4s/netty/Http4sNettyHandler.scala
jmcardon: Will add.

netty-server/src/main/scala/org/http4s/netty/Http4sNettyHandler.scala
```diff
@@ -160,6 +182,9 @@ sealed abstract class MikuHandler[F[_]](
         // https://github.com/netty/netty/blob/netty-3.9.3.Final/src/main/java/org/jboss/netty/handler/codec/http/HttpHeaders.java#L1075-L1080
         logger.debug(e)("Handling Header value error")
         sendSimpleErrorResponse(ctx, HttpResponseStatus.BAD_REQUEST); ()
+
+      case InvalidMessageException =>
+        sendSimpleErrorResponse(ctx, HttpResponseStatus.UNPROCESSABLE_ENTITY); ()
```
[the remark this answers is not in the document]
jmcardon: Fair, this is an internal server error.

netty-server/src/main/scala/org/http4s/netty/Http4sNettyHandler.scala
jmcardon: Hmm actually, reconsidering here:

The cleanup action is either just `unit` or `drainBody`, neither of which will throw an exception. 

Save a VM fatal error, `drainBody` doesn't fail/throw. I'd like to reconsider this one.

rossabaker: I think you're right, but I also think I've wasted a lot of my life thinking I was right and finding the murder weapon was hidden in an empty exception handler.  I guess I reflexively do this, if only to make it look even more jarring when people don't.

I won't die on this hill.

jmcardon: If it's any consolation whatsoever:

This part I spent a _lot_ of time on. In particular, because that cleanup action drains the buffers out of the reactive stream publisher if it wasn't used. This would be a huge memleak if it turned out to be an issue.

I spent hours debugging this and ensuring resources were released properly. I silently discard errors here because I came to the conclusion after a lot of moving around, spamming requests and observing output:
- If even one element of the stream was consumed this no-ops.
- This is primarily concerned with resource release and it only fires _after_ the channel write is complete. If it ever errors out (extremely low chance), it still means the stream was consumed somewhere else and thus released properly. It can only error if you consumed the stream and closed it and then tried to consume it again, in which case the upstream publisher is already done releasing resources.

I really hope the confidence that I gained by spending a long time on this doesn't come back to bite me, but I'm fairly confident this is right.

rossabaker:
Appveyor is unhealthy. I think this build is legitimate.

jmcardon:
Alright, so I have some performance tests here, for just a ping route, using native transports.

Blaze:
```
================================================================================
---- Global Information --------------------------------------------------------
> request count                                    6000000 (OK=5954785 KO=45215 )
> min response time                                      0 (OK=0      KO=0     )
> max response time                                  11040 (OK=11040  KO=0     )
> mean response time                                   507 (OK=511    KO=0     )
> std deviation                                        654 (OK=655    KO=0     )
> response time 50th percentile                        344 (OK=346    KO=0     )
> response time 75th percentile                        614 (OK=617    KO=0     )
> response time 95th percentile                       1580 (OK=1586   KO=0     )
> response time 99th percentile                       3091 (OK=3102   KO=0     )
> mean requests/sec                                37267.081 (OK=36986.242 KO=280.839)
---- Response Time Distribution ------------------------------------------------
> t < 800 ms                                       4978308 ( 83%)
> 800 ms < t < 1200 ms                              502200 (  8%)
> t > 1200 ms                                       474277 (  8%)
> failed                                             45215 (  1%)
---- Errors --------------------------------------------------------------------
> j.n.ConnectException: Failed to open a socket.                  41348 (91.45%)
> j.n.ConnectException: Cannot assign requested address: /52.43.   3765 ( 8.33%)
212.119:8080
> j.n.ConnectException: connection timed out: /52.43.212.119:808    102 ( 0.23%)
0
================================================================================
```

Netty (almost exactly as it is in this PR):

```
================================================================================
---- Global Information --------------------------------------------------------
> request count                                    6000000 (OK=5234048 KO=765952)
> min response time                                      0 (OK=0      KO=0     )
> max response time                                  62780 (OK=62770  KO=62780 )
> mean response time                                  3968 (OK=4548   KO=1     )
> std deviation                                       3021 (OK=2795   KO=256   )
> response time 50th percentile                       4301 (OK=4779   KO=0     )
> response time 75th percentile                       6079 (OK=6293   KO=0     )
> response time 95th percentile                       7868 (OK=8002   KO=0     )
> response time 99th percentile                      10270 (OK=10769  KO=0     )
> mean requests/sec                                12121.212 (OK=10573.834 KO=1547.378)
---- Response Time Distribution ------------------------------------------------
> t < 800 ms                                        533366 (  9%)
> 800 ms < t < 1200 ms                              218675 (  4%)
> t > 1200 ms                                      4482007 ( 75%)
> failed                                            765952 ( 13%)
---- Errors --------------------------------------------------------------------
> j.n.ConnectException: Failed to open a socket.                 548263 (71.58%)
> j.n.ConnectException: connection timed out: /52.43.212.119:808  91783 (11.98%)
0
> j.n.ConnectException: Cannot assign requested address: /52.43.  87734 (11.45%)
212.119:8080
> j.u.c.TimeoutException: Request timeout to not-connected after  29515 ( 3.85%)
 60000 ms
> j.u.c.TimeoutException: Request timeout to /52.43.212.119:8080   8657 ( 1.13%)
 after 60000 ms
================================================================================
```

So... Performance is crap. Blaze is positively obliterating netty in my load tests and it's an absurd difference.

I'd like to put this PR in the freezer as such while I perform heavy optimization on the handlers and overhaul it in general (maybe reopen it). In its current state it's not worth it to introduce into the repository.

rossabaker:
When 9% of the requests are failing on the one that's "winning," I don't really even care about the performance numbers.  Why are they both so unstable in this test?  Can you get a comparison at a load that is manageable?

aeons:
It’s just 1% failing though, right?

And since most of the errors are ‘failed to open a socket’ it seems like a benchmark problem.

rossabaker:
Excuse me.  I totally misread that report.  

I still am wary of benchmarks that return errors.  When one is clean and another is failing, that can say something.  When they're both buckling, I begin to wonder whether the client is healthy enough to be producing useful metrics.

yanns:
What is the status of this PR?

I'm willing to help here, maybe updating that for http4s 0.20?

jmcardon:
Status was the overhead of reactive streams and fs2 combined was so much I decided to stop working on it until I had the time to write a custom thing for it. 

I haven't had the time budget for open source at all, so I left it open for future "inspiration", but despite being functional I'm not a fan of the performance of this.

yanns:
Thx for the update.
Personal opinion: as long as it stays a PR, it's quite difficult for others to contribute to it.
Maybe this could be merged and marked as "experimental with bad performances"?

jmcardon:
Personally, I have a few opinions on this:

1. I don't want to merge anything I've written that I consider subpar.
2. Arguably, the part that needs to change is the meat of this PR. Anything merged will eventually be locked into binary compat and this makes fixing issues far more annoying.

yanns:
OK I understand. Thx for the update.

hamnis:
I have a version of this in a private repo on my github. Would it be OK if I publish this under my own maven groupId? @jmcardon @rossabaker ? I have dropped TLS and websockets for now, as a way of getting started. If not, I can start a new branch or move to an own repo under this org. As an experiment I think it can be valuable if not included in the core.

rossabaker:
I'd be happy to host it here in the http4s org.  (Similar to what we recently did with finagle, and have been doing with jdk-http-client).  It's more likely to be found and attract other contributors when it's next to the related projects.

hamnis:
@rossabaker allright. How do we proceed from here? We can do this on gitter if that is preferable

PR #1836, opened 2018-05-09
Remove httpbin.org dependency from client tests

rossabaker:
This is bad code but gets us back in business.  Unit tests have been failing today because httpbin.org is down.

We should look for an embedded server that's:
1. easy to set up
2. easy to set up with SSL
3. isn't ours, so we don't have circular dependencies

The Jetty scaffold only ticks boxes 1 and 3.


rossabaker:
CI failure is ongoing Travis networking issue.

PR #1838, opened 2018-05-09
Fix Http4sServlet compilation on 2.11

rossabaker:


rossabaker:
Travis is down, but master can't be any more broken. Merging.

PR #1842, opened 2018-05-09
better-monadic-for is better

rossabaker:
I consider this approaching "why wouldn't you use this?" status.

Purely compile time, so no downstream headaches.


jmcardon:
P L A G E R I S M

ChristopherDavenport:
@jmcardon takes a critical hit!
@rossabaker gains +100 experience

PR #1843, opened 2018-05-09
Enable -Ybackend-parallelism where appropriate

rossabaker:


cquiroz:
Have you measured how much faster this is?

ChristopherDavenport:
Likely not faster on travis, likely faster on our local machine though. But it would depend on the machine I imagine.

rossabaker:
67s to 63s on my machine.  Not life changing, but it's something.  Hopefully it keeps getting better.

PR #1844, opened 2018-05-09
sbt 1.1.5

sullis:


ChristopherDavenport:
Build won't even show me the whole log where the error is, restarting.

PR #1847, opened 2018-05-10
Release v0.18.11

rossabaker:


rossabaker:
bincompat was always assumed on Prometheus, I think.

jmcardon:
Fair. :+1: on green

PR #1851, opened 2018-05-10
Add an OkHttp client (with HTTP/2)

andimiller:
This adds an OkHttp implementation for Client.

This is heavily based on the one for async-http-client.

The motivation for this is to have a Client backend which is capable of HTTP/2.

Remaining questions are what the `chunkSize` should be when reading the body, if I've correctly hooked up the `dispose` on the `DisposableResponse`, and if the `writeTo` for writing bodies looks alright.

### DIFF
--- a/okhttp-client/src/main/scala/org/http4s/client/okhttp/OkHttp.scala
+++ b/okhttp-client/src/main/scala/org/http4s/client/okhttp/OkHttp.scala
@@ -28,17 +28,20 @@ import scala.concurrent.ExecutionContext
 
 object OkHttp {
 
-  val defaultconfig: OkHttpClient.Builder = new OkHttpClient.Builder()
-    .protocols(
-      List(
-        Protocol.HTTP_2,
-        Protocol.HTTP_1_1
-      ).asJava)
+  def defaultconfig[F[_]]()(implicit F: Effect[F]): F[OkHttpClient.Builder] = F.delay {
+    new OkHttpClient.Builder()
+      .protocols(
+        List(
+          Protocol.HTTP_2,
+          Protocol.HTTP_1_1
+        ).asJava)
+  }
 
-  def apply[F[_]](config: OkHttpClient.Builder = defaultconfig)(
+
+  def apply[F[_]](config: F[OkHttpClient.Builder] = null)(
       implicit F: Effect[F],
-      ec: ExecutionContext): F[Client[F]] = F.delay {
-    val client = config.build()
+      ec: ExecutionContext): F[Client[F]] = F.map(Option(config).getOrElse(defaultconfig[F]())) { c =>
+    val client = c.build()
     Client(
       Kleisli { req =>
         F.async[DisposableResponse[F]] { cb =>
@@ -56,10 +59,10 @@ object OkHttp {
     )
   }
 
-  def stream[F[_]](config: OkHttpClient.Builder = defaultconfig)(
+  def stream[F[_]](config: F[OkHttpClient.Builder] = null)(
       implicit F: Effect[F],
       ec: ExecutionContext): Stream[F, Client[F]] =
-    Stream.bracket(apply(config))(c => Stream.emit(c), _.shutdown)
+    Stream.bracket(apply(Option(config).getOrElse(defaultconfig[F]())))(c => Stream.emit(c), _.shutdown)
 
   private def handler[F[_]](cb: Either[Throwable, DisposableResponse[F]] => Unit)(
       implicit F: Effect[F],
--- a/okhttp-client/src/main/scala/org/http4s/client/okhttp/OkHttp.scala
+++ b/okhttp-client/src/main/scala/org/http4s/client/okhttp/OkHttp.scala
@@ -63,10 +63,13 @@ object OkHttp {
     )
   }
 
-  def stream[F[_]](config: F[OkHttpClient.Builder] = null)(
+  def stream[F[_]]()(implicit F: Effect[F], ec: ExecutionContext): Stream[F, Client[F]] =
+    stream(defaultConfig[F]())
+
+  def stream[F[_]](config: F[OkHttpClient.Builder])(
       implicit F: Effect[F],
       ec: ExecutionContext): Stream[F, Client[F]] =
-    Stream.bracket(apply(Option(config).getOrElse(defaultconfig[F]())))(c => Stream.emit(c), _.shutdown)
+    Stream.bracket(apply(config))(c => Stream.emit(c), _.shutdown)
 
   private def handler[F[_]](cb: Either[Throwable, DisposableResponse[F]] => Unit)(
       implicit F: Effect[F],
--- a/okhttp-client/src/main/scala/org/http4s/client/okhttp/OkHttp.scala
+++ b/okhttp-client/src/main/scala/org/http4s/client/okhttp/OkHttp.scala
@@ -1,0 +1,134 @@
+package org.http4s.client.okhttp
+
+import java.io.IOException
+
+import cats.data._
+import cats.effect._
+import cats.effect.implicits._
+import okhttp3.{
+  Call,
+  Callback,
+  OkHttpClient,
+  Protocol,
+  RequestBody,
+  Headers => OKHeaders,
+  MediaType => OKMediaType,
+  Request => OKRequest,
+  Response => OKResponse
+}
+import okio.BufferedSink
+import org.http4s.{Header, Headers, HttpVersion, Method, Request, Response, Status}
+import org.http4s.client.{Client, DisposableResponse}
+import fs2.Stream._
+import fs2._
+import fs2.io._
+
+import scala.collection.JavaConverters._
+import scala.concurrent.ExecutionContext
+
+object OkHttp {
+
+  val defaultconfig = new OkHttpClient.Builder()
+    .protocols(
+      List(
+        Protocol.HTTP_2,
+        Protocol.HTTP_1_1
+      ).asJava)
+
+  def apply[F[_]](config: OkHttpClient.Builder = defaultconfig)(
+      implicit F: Effect[F],
+      ec: ExecutionContext): Client[F] = {
+    val client = config.build()
+    Client(
+      Kleisli { req =>
+        F.async[DisposableResponse[F]] { cb =>
+          client.newCall(toOkHttpRequest(req)).enqueue(handler(cb))
+          ()
+        }
+      },
+      F.delay({
+        client.dispatcher.executorService().shutdown()
+        client.connectionPool().evictAll()
+        if (client.cache() != null) {
+          client.cache().close()
+        }
+      })
+    )
+  }
+
+  def stream[F[_]](config: OkHttpClient.Builder = defaultconfig)(
+      implicit F: Effect[F],
+      ec: ExecutionContext): Stream[F, Client[F]] =
+    Stream.bracket(F.delay(apply(config)))(c => Stream.emit(c), _.shutdown)
+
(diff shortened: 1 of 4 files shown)


okhttp-client/src/main/scala/org/http4s/client/okhttp/OkHttp.scala
```diff
@@ -0,0 +1,140 @@
...
+import fs2.Stream._
+import fs2._
+import fs2.io._
+
+import scala.collection.JavaConverters._
+import scala.concurrent.ExecutionContext
+
+object OkHttp {
+
+  def defaultconfig[F[_]]()(implicit F: Effect[F]): F[OkHttpClient.Builder] = F.delay {
```
rossabaker: A `Sync` constraint is sufficient to get `delay`.  But is this even a side effect?  Why isn't this pure?

Nit: camelcase `defaultConfig`

okhttp-client/src/main/scala/org/http4s/client/okhttp/OkHttp.scala
```diff
@@ -0,0 +1,140 @@
...
+    new OkHttpClient.Builder()
+      .protocols(
+        List(
+          Protocol.HTTP_2,
+          Protocol.HTTP_1_1
+        ).asJava)
+  }
+
+
+  def apply[F[_]](config: F[OkHttpClient.Builder] = null)(
```
rossabaker: I might prefer `default` and `apply` that takes a config instead of a null-default.  Even though I recognize what you're trying to do and it's safe, I flinched when I saw the null.

Does `build()` mutate the builder?  If not, I'm not sure why it needs to be passed in in `F`.

okhttp-client/src/main/scala/org/http4s/client/okhttp/OkHttp.scala
```diff
@@ -0,0 +1,140 @@
...
+      implicit F: Effect[F],
+      ec: ExecutionContext): F[Client[F]] = F.map(Option(config).getOrElse(defaultconfig[F]())) { c =>
```
rossabaker: Note that as soon as we're on cats-effect-1.0, we're going to want this to return a `Resource[F, Client[F]]`.  But this is the right thing for now.

okhttp-client/src/main/scala/org/http4s/client/okhttp/OkHttp.scala
```diff
@@ -0,0 +1,140 @@
...
+          client.newCall(toOkHttpRequest(req)).enqueue(handler(cb))
+          ()
+        }
+      },
+      F.delay({
+        client.dispatcher.executorService().shutdown()
+        client.connectionPool().evictAll()
+        if (client.cache() != null) {
+          client.cache().close()
+        }
```
rossabaker: Can any of these operations fail?  I'm surprised that there isn't a single cleanup method on OkHttp that handles this for you.

okhttp-client/src/main/scala/org/http4s/client/okhttp/OkHttp.scala
```diff
@@ -0,0 +1,140 @@
...
+    new Callback {
+      override def onFailure(call: Call, e: IOException): Unit =
+        ec.execute(new Runnable { override def run(): Unit = cb(Left(e)) })
+
+      override def onResponse(call: Call, response: OKResponse): Unit = {
+        val protocol = response.protocol() match {
+          case Protocol.HTTP_2 => HttpVersion.`HTTP/2.0`
+          case Protocol.HTTP_1_1 => HttpVersion.`HTTP/1.1`
+          case Protocol.HTTP_1_0 => HttpVersion.`HTTP/1.0`
+          case _ => HttpVersion.`HTTP/1.1`
```
rossabaker: Did you get around the `QUIC` / `SPDY` issue by carefully selecting the protocols above, or does this case still get hit?

okhttp-client/src/main/scala/org/http4s/client/okhttp/OkHttp.scala
```diff
@@ -0,0 +1,140 @@
...
+        }
+        val bodyStream = response.body.byteStream()
+        val dr = new DisposableResponse[F](
+          Response[F](headers = getHeaders(response), httpVersion = protocol)
+            .withStatus(
+              Status
+                .fromInt(response.code())
+                .getOrElse(Status.apply(response.code())))
+            .withBodyStream(
+              readInputStream(F.pure(bodyStream), chunkSize = 1024, closeAfterUse = true)),
```
rossabaker: There's no way to process the body asynchronously?  This is right if not, but that's sad for OkHttp.

okhttp-client/src/main/scala/org/http4s/client/okhttp/OkHttp.scala
```diff
@@ -0,0 +1,140 @@
...
+          case Protocol.HTTP_1_0 => HttpVersion.`HTTP/1.0`
+          case _ => HttpVersion.`HTTP/1.1`
+        }
+        val bodyStream = response.body.byteStream()
+        val dr = new DisposableResponse[F](
+          Response[F](headers = getHeaders(response), httpVersion = protocol)
+            .withStatus(
+              Status
+                .fromInt(response.code())
+                .getOrElse(Status.apply(response.code())))
```
rossabaker: Wow, I'm not sure what we were thinking on that `Status.apply`.  We can make a `Status(666, "This is illegal, yo.")`.

okhttp-client/src/main/scala/org/http4s/client/okhttp/OkHttp.scala
```diff
@@ -0,0 +1,140 @@
...
+                  sink.write(b); ()
+                }
+              })
+              .compile
+              .drain
+              .runAsync(_ => IO.unit)
+              .unsafeRunSync()
+        }
+      // if it's a GET or HEAD, okhttp wants us to pass null
+      case _ if req.method == Method.GET || req.method == Method.HEAD => null
```
rossabaker: What a strange rule.  What about `DELETE`?  What about a `POST` that's empty?

okhttp-client/src/main/scala/org/http4s/client/okhttp/OkHttp.scala
```diff
@@ -0,0 +1,140 @@
...
+            req.body.chunks
+              .map(_.toArray)
+              .to(Sink { b: Array[Byte] =>
+                F.delay {
+                  sink.write(b); ()
+                }
+              })
+              .compile
+              .drain
+              .runAsync(_ => IO.unit)
```
rossabaker: This swallows errors.  We should add something like cats-effect's `Logger.reportFailure`.

rossabaker:
Oh, were you trying to get this into 0.18?  The target branch is master.

okhttp-client/src/main/scala/org/http4s/client/okhttp/OkHttp.scala
andimiller: Creating the `Builder` creates all of the internals which the `OkHttpClient` is going to use, including an `Executor` and a `ConnectionPool`, so during testing when I shared the same configuration between two clients, shutting one down shut down the other.

okhttp-client/src/main/scala/org/http4s/client/okhttp/OkHttp.scala
andimiller: I originally wanted to do `config: F[OkHttpClient.Builder] = defautconfig[F]` but that's not valid, so this was my solution, I could split it into two methods to remove the null, as in the above comment, the `Builder` has a lot of nasty state in it.

okhttp-client/src/main/scala/org/http4s/client/okhttp/OkHttp.scala
andimiller: By not selecting any other protocols it should be safe, it also looks like `QUIC` and `SPDY` aren't fully implemented in OkHttp, so it's not an actual problem yet.

okhttp-client/src/main/scala/org/http4s/client/okhttp/OkHttp.scala
andimiller: https://github.com/square/okhttp/wiki/Recipes#asynchronous-get based on their wiki it seems like there's no way to actually stream the body :(

okhttp-client/src/main/scala/org/http4s/client/okhttp/OkHttp.scala
andimiller: https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/Request.java#L235-L240

This is part of their `Request` API being opinionated, `permitsRequestBody` only says no for `GET` and `HEAD`, and I should check `requiresRequestBody` to check if I'm clear there, but always passing an empty body class will work.

andimiller:
ah I guess I can re-raise it against the 0.18 branch if they're separate branches, I didn't check how far ahead master was

okhttp-client/src/main/scala/org/http4s/client/okhttp/OkHttp.scala
andimiller: https://square.github.io/okhttp/3.x/okhttp/okhttp3/OkHttpClient.html

the `Shutdown isn't necessary` section describes shutting it down this way, I should probably wrap them in case they fail

okhttp-client/src/main/scala/org/http4s/client/okhttp/OkHttp.scala
rossabaker: Oh, wow.  I thought that would be deferred until `build()`.  Then I think you're right.

okhttp-client/src/main/scala/org/http4s/client/okhttp/OkHttp.scala
rossabaker: Yeah, that implicit not being visible to default arguments is a drag.

Let's see if anyone else chimes in on the design here.

okhttp-client/src/main/scala/org/http4s/client/okhttp/OkHttp.scala
rossabaker: > The threads and connections that are held will be released automatically if they remain idle.

This seems like it would depend on the type of executor passed in?

That section makes me feel better that it's not a crisis if any of these steps fail, but yeah, wrapping for best effort is probably wise.

okhttp-client/src/main/scala/org/http4s/client/okhttp/OkHttp.scala
rossabaker: Yeah, looks like you made the best of it with this.

okhttp-client/src/main/scala/org/http4s/client/okhttp/OkHttp.scala
rossabaker: Yes, I think you got this right.

okhttp-client/src/main/scala/org/http4s/client/okhttp/OkHttp.scala
aeons: We have it split in two methods in a lot of other places. For consistency, I would like that here as well, but I'm not very opinionated on it.

rossabaker:
Yeah, I don't see any reason we couldn't support this on the 0.18 series.  We're still waiting on dependencies to stabilize before we release anything off master.

andimiller:
Please see https://github.com/http4s/http4s/pull/1864 for the continuation of this PR

PR #1855, opened 2018-05-12
[WIP/Discussion needed] Docs overhaul

asoltysik:
- Removes the `website` module, moves all stuff to `docs` module
- Changes the theme to `hugo-learn-theme` which gives us search for free
- There is also a favicon but I was too lazy to refresh my browser cache :-D
- All *.md files moved to `docs/src/main/tut` as AFAIK hugo does not support multiple content directories
- Let the screenshots speak:

Screenshot1

Screenshot2

I need some help from somebody experienced with actually deploying this stuff (@rossabaker ?) to tell me what more needs to be done. There are also some cosmetic changes left like adding "http4s" to the logo.

Also a question, this purple theme is the official one, it [can be quite simply changed](https://learn.netlify.com/en/basics/style-customization/#yours-variant) to maybe orange to match http4s theme (but i suck at css). What's your opinion?


aeons:
Does the entire hugo-learn-theme (including an example site) need to live in the repo?

asoltysik:
@aeons One solution would be to include hugo-learn-theme as a submodule, then after cloning one would have to manually download the files for it with `git submodule init` followed by `git submodule update`.

stephennancekivell:
For what its worth, I like where this is going.

Its good how the nav is on the left and in a different colour, it stands out from the content of the article. Another thing that could be added is a per article index, like the [cats-effect docs](https://typelevel.org/cats-effect/datatypes/io.html).

PS, the http4s docs are really good. They helped me convince my team to adopt it. Thanks.

rossabaker:
Glad to hear that about the docs.  I think they're one of our weaker points, but it's nice to hear that the work on them thus far is paying off!

In page anchors are discussed at https://github.com/matcornic/hugo-theme-learn/issues/68.  The [video](http://recordit.co/KTJPaCAF92) shows how they work.  I had not yet discovered this feature.  Maybe we can do (and propose) better?

asoltysik:
I am closing the issue because of the lack of webdev skills required to pull this off. I should've closed this a long time ago, but my thinking was that I would improve my (nonexistent) design skills and come back to this. I never got around to it and to be honest forgot about this PR. Sorry for the time wasted!

PR #1861, opened 2018-05-14
Backport: Log requests with empty body

rossabaker:
Backports #1852.  I didn't look carefully at the target branch before I merged.


rossabaker:
Hmm. I bet this still drops the request log when the response body is None.

ChristopherDavenport:
I think we don't have a clear answer on when request and response both have empty bodies that aren't read.

rossabaker:
This moves the finalizer to the response, which might not exist. I think we can plug that hole with an orElse case.  In the `HttpApp` world, this isn't a problem.  I think it's a bug in this backport.

rossabaker:
Hmm.  We're only logging responses on the happy path.  Nothing gets logged if the service raises an error or returns None.

PR #1862, opened 2018-05-15
Multipart Decoder Speed optimizations + File streaming multipart

jmcardon:
Note: This PR is not for the faint of heart. I knowingly abuse reference equality on `Stream.empty` to avoid method calls on `pull` as an optimization and it does fairly well

This PR addresses a conversation @rossabaker and I had on gitter w.r.t improving our multipart story for large bodies, not only as a speed optimization + Also adding a multipart decoder that, after a threshold, does not buffer parts in memory.

Benchmark Code for new, in-memory decoder: https://github.com/jmcardon/routing-bench/blob/new-multipart/src/main/scala/Bench/MultipartBench.scala

Bench results for the new decoder speed (where anything that says newMultipart is the latest one)
```
[info] Result "Bench.MultipartBench.smallOldMultipart":
[info]   690.982 ±(99.9%) 15.500 ops/s [Average]
[info]   (min, avg, max) = (658.990, 690.982, 702.756), stdev = 14.498
[info]   CI (99.9%): [675.482, 706.482] (assumes normal distribution)
[info] # Run complete. Total time: 00:04:06
[info] Benchmark                                      Mode  Cnt    Score    Error  Units
[info] Bench.MultipartBench.bigNewMultipart          thrpt   15   42.877 ±  0.872  ops/s
[info] Bench.MultipartBench.bigOldMultipart          thrpt   15   22.921 ±  0.700  ops/s
[info] Bench.MultipartBench.bigPartBodyNewMultipart  thrpt   15   54.281 ±  1.832  ops/s
[info] Bench.MultipartBench.bigPartBodyOldMultipart  thrpt   15   22.668 ±  3.414  ops/s
[info] Bench.MultipartBench.mediumNewMultipart       thrpt   15  117.357 ±  3.321  ops/s
[info] Bench.MultipartBench.mediumOldMultipart       thrpt   15   70.545 ±  1.226  ops/s
[info] Bench.MultipartBench.smallNewMultipart        thrpt   15  976.175 ± 22.306  ops/s
[info] Bench.MultipartBench.smallOldMultipart        thrpt   15  690.982 ± 15.500  ops/s
```

This PR also adds `MixedMultipart[F]`, which represents a `BasicPart` (a simple, in-memory part) and a `FilePart`, a part which has been streamed to a file (primarily useful for large bodies), which includes helpers for cleaning up files after the fact. This new multipart decoder is configurable in the number of parts, max size before writing to a file, truncation and header size Limit. 

I'm open to moving the second part to another PR, but a lot of the reason the PR is deceptively big, is that a good chnk of the testing code is just copied and adapted for `MixedMultipart`. The actual `MixedMultipart` impl is not that large.

### DIFF
--- a/tests/src/test/scala/org/http4s/multipart/MultipartParserSpec.scala
+++ b/tests/src/test/scala/org/http4s/multipart/MultipartParserSpec.scala
@@ -390,11 +390,6 @@ object MultipartParserSpec extends Specification {
     }
   }
 
-//  def multipartStreamingTests(
-//      name: String,
-//      multipartPipe: Boundary => Pipe[IO, Byte, Multipart[IO]],
-//      multipartLimitPipe: (Boundary, Int) => Pipe[IO, Byte, Multipart[IO]],
-//      partsPipe: Boundary => Pipe[IO, Byte, Part[IO]]): Fragment = {
   "form streaming parsing" should {
     Fragments.foreach(List(1, 2, 3, 5, 8, 13, 21, 987)) { chunkSize =>
       s"produce the body from a single part with chunk size ${chunkSize}" in {


core/src/main/scala/org/http4s/multipart/MultipartParser.scala
```diff
@@ -154,30 +156,35 @@ object MultipartParser {
...
+  private[this] val BoundaryBytesN: Boundary => Array[Byte] = boundary =>
+    boundary.value.getBytes("UTF-8")
   val StartLineBytesN: Boundary => Array[Byte] = BoundaryBytesN.andThen(DashDashBytesN ++ _)
 
-  private val ExpectedBytesN: Boundary => Array[Byte] =
+  private[this] val ExpectedBytesN: Boundary => Array[Byte] =
     BoundaryBytesN.andThen(CRLFBytesN ++ DashDashBytesN ++ _)
+  private[this] val dashByte: Byte = '-'.toByte
+  private[this] val streamEmpty = Stream.empty
+  private[this] val PullUnit = Pull.pure[Pure, Unit](())
```
jmcardon: Note: should probably contrib this upstream.

core/src/main/scala/org/http4s/multipart/MultipartParser.scala
```diff
@@ -388,134 +387,186 @@ object MultipartParser {
...
+      expectedBytes: Array[Byte],
+      limit: Int): Pull[F, Part[F], Unit] =
+    Pull
+      .eval(parseHeaders(headerStream))
+      .flatMap { hdrs =>
+        splitHalf(expectedBytes, rest).flatMap {
+          case (l, r) =>
+            //We hit a boundary, but the rest of the stream is empty
+            //and thus it's not a properly capped multipart body
+            if (r == streamEmpty) {
```
jmcardon: Forgive both reference equality abuses: This gets caught downstream if it happens that the reference equality check doesn't hold, but it improves the common case of bodies constructed normally from byte streams a la blaze

core/src/main/scala/org/http4s/multipart/MultipartParser.scala
```diff
@@ -388,134 +387,186 @@ object MultipartParser {
...
       values: Array[Byte],
-      stream: Stream[F, Byte]): Pull[F, Nothing, (Stream[F, Byte], Stream[F, Byte])] = {
+      stream: Stream[F, Byte],
+      limit: Int): SplitStream[F] = {
+
+    /** Check if a particular chunk a final chunk, that is,
+      * whether it's the boundary plus an extra "--", indicating it's
+      * the last boundary
+      */
+    def checkIfLast(c: Chunk[Byte], rest: Stream[F, Byte]): SplitStream[F] =
```
jmcardon: I was conflicted on moving this out of the inner body if the method, but then I realized that this needs a reference to `go` for the continuation. It's a large function but mostly a side effect of us working on chunks directly so idk how to improve this :(

core/src/main/scala/org/http4s/multipart/MultipartParser.scala
```diff
@@ -648,33 +700,215 @@ object MultipartParser {
...
+  /** Same as the other streamed parsing, except
+    * after a particular size, it buffers on a File.
+    * @param maxSizeBeforeWrite Defaults to 50 megabytes
+    * @tparam F
+    * @return
+    */
+  def parseStreamedFile[F[_]: Sync](
+      boundary: Boundary,
+      limit: Int = 1024,
+      maxSizeBeforeWrite: Int = 52428800): Pipe[F, Byte, Multipart[F]] = { st =>
```
jmcardon: So currently this helps in the common case that parts themselves are large. It doesn't help for a multipart of `N` parts which add up to the size of a large part. 

I am considering adding a configurable part limit to avoid abuse. Thoughts?

core/src/main/scala/org/http4s/multipart/MultipartParser.scala
```diff
@@ -648,33 +700,215 @@ object MultipartParser {
...
-      if (limitCTR >= limit) {
-        Pull.raiseError(
-          MalformedMessageBodyFailure(s"Part header was longer than $limit-byte limit"))
+        limitCTR: Int): SplitFileStream[F] =
+      if (limitCTR >= maxBeforeWrite) {
+        for {
+          path <- Pull.eval(F.delay(Files.createTempFile("", "")))
+          _ <- Pull.eval(lacc.through(io.file.writeAll[F](path)).compile.drain)
+          split <- streamAndWrite(s, state, Stream.empty, racc, 0, path)
+        } yield split
```
jmcardon: missing `handleErrorWith` on `streamAndWrite`

core/src/main/scala/org/http4s/multipart/MultipartParser.scala
```diff
@@ -243,17 +251,9 @@ object MultipartParser {
...
-    *
-    * If it is a fresh chunk (no carry over partial match),
-    * everything prior to the partial match is added to the accumulator,
-    * and the partial match is considered the carry over.
-    *
-    * Else, if the whole block is a partial match,
-    * add it to the carry over
+    * DO NOT USE. Was made private[http4s] because
+    * Jose messed up hard like 5 patches ago and now it breaks bincompat to
+    * remove.
```
rossabaker: If it's marked deprecated, it'll error on actual usage, and we'll know to clean it out.

core/src/main/scala/org/http4s/multipart/MultipartParser.scala
```diff
@@ -388,134 +388,185 @@ object MultipartParser {
...
+    */
+  private def parseHeaders[F[_]: Sync](strim: Stream[F, Byte]): F[Headers] = {
+    def tailrecParse(s: Stream[F, Byte], headers: Headers): Pull[F, Headers, Unit] =
+      splitHalf[F](CRLFBytesN, s).flatMap {
+        case (l, r) =>
+          l.through(fs2.text.utf8Decode[F])
+            .fold("")(_ ++ _)
+            .map { string =>
+              val ix = string.indexOf(':')
+              if (string.indexOf(':') >= 0) {
```
rossabaker: Can use your `ix` here.

core/src/main/scala/org/http4s/multipart/file/MixedPart.scala
```diff
@@ -0,0 +1,85 @@
...
+import fs2.Stream
+import fs2.io.{readInputStream, file => fs2File}
+import fs2.text.utf8Encode
+import java.io.{File, FileInputStream, InputStream}
+import java.net.URL
+import java.nio.file.{Files, Path}
+
+import org.http4s.headers.`Content-Disposition`
+
+sealed abstract class MixedPart[F[_]] {
```
rossabaker: Do we need to introduce a new type here?  Does the consumer get anything out of knowing that the part has been persisted to a temp file?  We could make the contract that you have to `drain` the part body to clean up the temp file.

core/src/main/scala/org/http4s/multipart/file/MixedPart.scala
```diff
@@ -0,0 +1,85 @@
...
+  def body(implicit F: Sync[F]): Stream[F, Byte] = bodyStream
+}
+
+final case class FilePart[F[_]](headers: Headers, fileStream: Stream[F, Byte], filePath: Path)
+    extends MixedPart[F] {
+
+  def cleanupFile(implicit F: Sync[F]): F[Unit] =
+    F.delay(Files.delete(filePath))
+
+  def body(implicit F: Sync[F]): Stream[F, Byte] = fileStream.onFinalize(cleanupFile)
```
rossabaker: `fileStream` cleans up the file and `body` does not.  Is there any reason to expose the former?

core/src/main/scala/org/http4s/multipart/file/MixedPart.scala
```diff
@@ -0,0 +1,85 @@
...
+object MixedPart {
+  private val ChunkSize = 8192
+
+  def formData[F[_]: Sync](name: String, value: String, headers: Header*): MixedPart[F] =
+    BasicPart(
+      `Content-Disposition`("form-data", Map("name" -> name)) +: headers,
+      Stream.emit(value).through(utf8Encode))
+
+  def fileData[F[_]: Sync](name: String, file: File, headers: Header*): MixedPart[F] =
+    fileData(name, file.getName, new FileInputStream(file), headers: _*)
```
rossabaker: Maybe we should use NIO here to get non-blocking.

I think we're stuck on the `URL` version.

jmcardon:
Accidentally closed, sorry. Will reopen

core/src/main/scala/org/http4s/multipart/MultipartParser.scala
jmcardon: Will mark

jmcardon: P.s, this is already cleaned out on master

PR #1864, opened 2018-05-16
Add an OkHttp client (with HTTP/2) (to 0.18)

andimiller:
This adds an OkHttp implementation for Client.

This is heavily based on the one for async-http-client.

The motivation for this is to have a Client backend which is capable of HTTP/2.

This PR continues https://github.com/http4s/http4s/pull/1851 and is against `release-0.18.x` in stead of master

### DIFF
--- a/okhttp-client/src/main/scala/org/http4s/client/okhttp/OkHttp.scala
+++ b/okhttp-client/src/main/scala/org/http4s/client/okhttp/OkHttp.scala
@@ -28,17 +28,20 @@ import scala.concurrent.ExecutionContext
 
 object OkHttp {
 
-  val defaultconfig: OkHttpClient.Builder = new OkHttpClient.Builder()
-    .protocols(
-      List(
-        Protocol.HTTP_2,
-        Protocol.HTTP_1_1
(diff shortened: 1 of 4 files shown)


okhttp-client/src/main/scala/org/http4s/client/okhttp/OkHttp.scala
```diff
@@ -0,0 +1,182 @@
...
+      val client = c.build()
+      Client(
+        Kleisli { req =>
+          F.async[DisposableResponse[F]] { cb =>
+            client.newCall(toOkHttpRequest(req)).enqueue(handler(cb))
+            ()
+          }
+        },
+        F.delay({
+          try {
```
jmcardon: Should we accumulate these errors here? Seems like we're swallowing an awful lot of stuff.

rossabaker: They're all being logged, and `OkHttp` eventually shuts down idle resources anyway. See discussion on #1851.  I think we're fine here.

PR #1865, opened 2018-05-16
Multipart default improvements, mixed file/in memory streaming decoder

jmcardon:
Continuation of #1862


jmcardon:
I can add it, sure.

jmcardon:
Do we want `max size per part`? This is the only thing that we don't have.

Alternatively, we can pass it a configuration object instead of relying purely on args.

jmcardon:
That's fair. I think we can just ask people to limit entire bodies if they're worried about a threshold instead.

jmcardon:
for the record the last commit is just a semantic change since I had a redundant assignment for some reason. I think `val bytes = chunk` used be like `val bytes = doStuffWithChunk(chunk)` or something and then I took that out.

Issue #1859, opened 2018-05-14
Add support for 128-bit Trace ID with Zipkin/X-B3 built-in header parser


sbuzzard:
Currently the Zipkin/X-B3 built-in header for Trace ID only supports 64 bit long values.  The spec allows for either 64 or 128 bit representations.  Understand that this change would not be backward compatible as it would need to have a presentation other than long.  Currently we drop the built-in parser and use our own (where we have a representation of UUID but perhaps a custom case class would be better).  I could provide a PR for this if desired.

PR #1866, opened 2018-05-16
Support both 64 and 128 bit values for Zipkin/X-B3 TraceId HTTP Heade…

sbuzzard:
…r per the spec.  This is in reference to #1859


core/src/main/scala/org/http4s/headers/package.scala
```diff
@@ -5,16 +5,35 @@ import org.http4s.util.Writer
...
-      writer
-        .append("0" * leadingHexZeroCount)
-        .append(id.toHexString)
+  private[headers] def xB3RenderValueImpl(writer: Writer, idMostSigBits: Long, idLeastSigBits: Option[Long] = None): writer.type = {
+    val bitsToHalfByteRatio = 4
+    val idLength = 16
+    idLeastSigBits match {
+      case Some(idLsb) =>
+        if (idMostSigBits == 0L && idLsb == 0L) {
+          writer.append("0" * idLength * 2)
```
rossabaker: We could do the tiniest of microoptimizations here, and below with `"0" * idLength`, by caching these strings.  Might even be faster to `.take` from the longer of these on the other zeroes.

This is not a blocker, just a thought.

Issue #1868, opened 2018-05-20
missing query parameter shouldnt return 404


stephennancekivell:
Hi, I've noticed that using `QueryParamDecoderMatcher` in a HttpService returns a 404 for a missing param. I dont think this is quite right.

I think people expect a 400 bad request saying that the param is required. If I get a 404, I think I've got the path wrong. See also [this stackoverflow that suggests using 422](https://stackoverflow.com/questions/3050518/what-http-status-response-code-should-i-use-if-the-request-is-missing-a-required)

Im not sure if we want to introduce a breaking change to change this, maybe some people like it and are depending on the behaviour. But perhaps we could add a new Matcher similar to `ValidatingQueryParamDecoderMatcher` to do this. Or even just document the behavour and a alternative.

I'd like to work on this, Im looking for advice and support on what we should do.

Consider these examples
```scala
  "gives 404 on missing query param" >> {
    object NameQueryParamMatcher
        extends QueryParamDecoderMatcher[String]("name")

    val service = HttpService[IO] {
      case GET -> Root :? NameQueryParamMatcher(name) =>
        Ok(s"hello $name")
    }

    val getRoot = Request[IO](Method.GET, uri("/"))

    val io = service.orNotFound(getRoot)

    io.unsafeRunSync().status.code must beEqualTo(404) // would be better if it was bad request
  }

  "gives 400 on invalid query param manually parsed" >> {
     // how to achieve desired result by manually parsing the request.
    val service = HttpService[IO] {
      case req @ GET -> Root =>
        req.params.get("name") match {
          case None =>
            BadRequest("missing name")
          case Some(name) =>
            Ok(s"hello $name")
        }
    }

    val responseWithName =
      service.orNotFound(Request[IO](Method.GET, uri("/?name=cat")))
    responseWithName.unsafeRunSync().status.code must beEqualTo(200)

    val responseWithoutName =
      service.orNotFound(Request[IO](Method.GET, uri("/")))
    responseWithoutName.unsafeRunSync().status.code must beEqualTo(400)
  }

  "gives 400 on present but invalid query param" >> {
    implicit val yearQueryParamDecoder: QueryParamDecoder[Year] =
      QueryParamDecoder[Int].map(Year.of)

    object YQP extends ValidatingQueryParamDecoderMatcher[Year]("year")

    val service = HttpService[IO] {
      case GET -> Root :? YQP(y) =>
        y.fold(
[...]

PR #1869, opened 2018-05-21
Extend query paramater docs #1868

stephennancekivell:
for #1868. Give a example of Optional and Validating QueryParamMater and document how missing arguments are handled.

### DIFF
--- a/docs/src/main/tut/dsl.md
+++ b/docs/src/main/tut/dsl.md
@@ -392,4 +392,54 @@ val averageTemperatureService = HttpService[IO] {
 }
 ```
 
+#### Optional query parameters
+
+To accept a optional query parameter a `OptionalQueryParamDecoderMatcher` can be used.
+
+```tut:book
+import java.time.Year
+import org.http4s.client.dsl.io._
+
+implicit val yearQueryParamDecoder: QueryParamDecoder[Year] =
+  QueryParamDecoder[Int].map(Year.of)
+
+object OptionalYearQueryParamMatcher extends OptionalQueryParamDecoderMatcher[Year]("year")
+
+def getAverageTemperatureForCurrentYear: IO[String] = ???
+def getAverageTemperatureForYear(y: Year): IO[String] = ???
+
+val service2 = HttpService[IO] {
+  case GET -> Root / "temperature" :? OptionalYearQueryParamMatcher(maybeYear) =>
+    maybeYear match {
+      case None =>
+        Ok(getAverageTemperatureForCurrentYear)
+      case Some(year) =>
+        Ok(getAverageTemperatureForYear(year))
+    }
+}
+```
+
+#### Missing required query parameters
+
+A request with a missing required query parameter will fall through to the following `case` statements and may eventually return a 404. To provide contextual error handling, optional query parameters or fallback routes can be used.
+
+#### Invalid query parameter handling
+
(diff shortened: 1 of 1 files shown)


docs/src/main/tut/dsl.md
```diff
@@ -392,4 +392,48 @@ val averageTemperatureService = HttpService[IO] {
...
 ```
 
+#### Optional query parameters
+
+To accept a optional query parameter a `OptionalQueryParamDecoderMatcher` can be used.
+
+```tut:book
+object OptionalYearQueryParamMatcher extends OptionalQueryParamDecoderMatcher[Year]("name")
+
+def getAverageTemperatureForCurrentYear(): IO[String] = ???
```
rossabaker: Nit: if it returns a pure value, Scala style is to omit the parens.

docs/src/main/tut/dsl.md
```diff
@@ -392,4 +392,48 @@ val averageTemperatureService = HttpService[IO] {
...
+def getAverageTemperatureForYear(y: Year): IO[String] = ???
+
+val service = HttpRoutes.of[IO] {
```
rossabaker: In release-0.18.x, it's just `HttpService { ... }`.  In master, it's `HttpRoutes.of`.  That's why the build failed.  To get this into the 0.18 docs, change it to the former, and we can fix it up when it rolls forward to master.

docs/src/main/tut/dsl.md
```diff
@@ -392,4 +392,48 @@ val averageTemperatureService = HttpService[IO] {
...
+        Ok(getAverageTemperatureForCurrentYear())
+      case Some(year) =>
+        Ok(getAverageTemperatureForYear(year))
+    }
+}
+```
+
+#### Missing required query parameters
+
+A request with a missing required query parameter will be fall through to the following `case` statements and may eventually return a 404. To provide contextual error handling, optional query parameters or fallback routes can be used.
```
rossabaker: s/will be fall through/will fall through/

docs/src/main/tut/dsl.md
```diff
@@ -392,4 +392,48 @@ val averageTemperatureService = HttpService[IO] {
...
+
+#### Invalid query parameter handling
+
+To validate query parsing you can use `ValidatingQueryParamDecoderMatcher` which returns a `ParseFailure` if the parameter cannot be decoded.
```
rossabaker: It may be worth mentioning that the mapping to `BadRequest` shouldn't echo any values from the request for [XSS  reasons](https://www.owasp.org/index.php/Cross-site_Scripting_(XSS)).

docs/src/main/tut/dsl.md
stephennancekivell: oh, thank you. Will do. I was pretty confused.

docs/src/main/tut/dsl.md
stephennancekivell: Its probably not pure. The evaluation of "CurrentYear" has a side effect =P

stephennancekivell:
Hey @rossabaker, Yea its not working currently. Im pretty confused as to why it says `not found: type OptionalYearQueryParamMatcher` where it is defined 4 lines above. When I do this in unit tests it works.

rossabaker:
You're invoking it as `OptionalYearQueryParamMatcher[Year]`, but you're referring to an object which has no type parameter.  What if it's just `OptionalYearQueryParamMatcher`?

stephennancekivell:
Thanks Ross, yea that was it.

PR #1873, opened 2018-05-25
Log Unmatch and Errors

ChristopherDavenport:
Seems like a little improvement to request logging.

@jmcardon Is going to love this one

### DIFF
--- a/server/src/main/scala/org/http4s/server/middleware/RequestLogger.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/RequestLogger.scala
@@ -60,7 +60,13 @@ object RequestLogger {
                     )
                   )
                 )
-            }
+            }.orElse(
+              OptionT.liftF(Logger.logMessage[F, Request[F]](req.withBodyStream(newBody))(
+                    logHeaders,
+                    logBody,
+                    redactHeadersWhen)(logger.info(_))) *>
+              OptionT.none[F, Response[F]]
+            )
           }
     }
 }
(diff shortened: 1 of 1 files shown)


ChristopherDavenport:
Scalafmt hates me!

ChristopherDavenport:
I lean towards 2. Where we explicitly log that it went through the middleware. and was an error/unmatched and that the behavior is whatever the expected behavior outside is.

server/src/main/scala/org/http4s/server/middleware/RequestLogger.scala
```diff
@@ -42,15 +42,35 @@ object RequestLogger {
             )
 
             val response = routes(changedRequest)
-            response.map { resp =>
-              resp.withBodyStream(
-                resp.body.onFinalize(
+            response.attempt
+              .flatMap {
+                case Left(e) =>
+                  OptionT.liftF(
```
jmcardon: I think this entire thing can just be one big `OptionT.liftF(logyourbleepblorp >> F.raiseError(e))`  if we want to be pedantic.

server/src/main/scala/org/http4s/server/middleware/RequestLogger.scala
```diff
@@ -42,15 +42,35 @@ object RequestLogger {
...
+                        Logger.logMessage[F, Request[F]](req.withBodyStream(newBody))(
+                          logHeaders,
+                          logBody,
+                          redactHeadersWhen)(logger.info(_))
+                      )
+                    )
+                  )
+              }
+              .orElse(
+                OptionT.liftF(
```
jmcardon: This can probable be OptionT(log >> F.pure(None)).

Issue #1872, opened 2018-05-24
Requests from Client.fromHttpService do not include Host header


bpholt:
I added a `VirtualHost` middleware to a test service (so that I can test that the client is making its request to the right server). Requests made using `Client.fromHttpService` to the service do not include a `Host` header, so the middleware rejects them as bad requests. (Blaze-client’s `Http1Connection` has a `validateRequest` method that adds a `Host` header if it’s not already on the request, so this is not a problem for normal client requests.)

See also https://tools.ietf.org/html/rfc7230#section-5.4.

This was [discussed in gitter](https://gitter.im/http4s/http4s?at=5b072c863815f70bffd74741) and I suggested we have `fromHttpService` add a `Host` header if given a request with an absolute URI, and leave the current behavior alone for other requests for now. (I'm working on a PR to implement this.)

aeons:
Closed in #1874

PR #1874, opened 2018-05-25
add Host header to requests submitted by Client.fromHttpService if the request URI is absolute

bpholt:
Addresses #1872 in the easy case where the request URI is absolute.

### DIFF
--- a/client/src/main/scala/org/http4s/client/Client.scala
+++ b/client/src/main/scala/org/http4s/client/Client.scala
@@ -9,7 +9,7 @@ import fs2._
 import java.io.IOException
 import java.util.concurrent.atomic.AtomicBoolean
 import org.http4s.Status.Successful
-import org.http4s.headers.{Accept, MediaRangeAndQValue}
+import org.http4s.headers.{Accept, Host, MediaRangeAndQValue}
 import scala.concurrent.SyncVar
 import scala.util.control.NoStackTrace
 import org.log4s.getLogger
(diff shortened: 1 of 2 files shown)


client/src/main/scala/org/http4s/client/Client.scala
```diff
@@ -319,6 +319,15 @@ object Client {
     Client(disposableService(service), F.delay(isShutdown.set(true)))
   }
 
+  private def addHostHeaderIfUriIsAbsolute[F[_]](req: Request[F]): Request[F] = {
+    val maybeHost = for {
+      host <- req.uri.host
+      if req.headers.get(Host).isEmpty
```
ChristopherDavenport: `_ <- Alternative[Option].guard(req.headers.get(Host).isEmpty)`

bpholt: @ChristopherDavenport I don't understand the difference here. Can you explain why `Alternative[Option].guard` is preferred?

ChristopherDavenport: One is implemented in terms of `flatMap` the other in terms of `withFilter`. I just see this as utilizing non-functional language features which has the possibility for issues in the future.

jmcardon: No comment on cats fp syntax (I don't care for it but I don't hate it), however, instead of the withfilter guard, you can always just

```scala
req.uri.host match {
  case Some(host) if req.headers.get(Host).isEmpty => 
     req.withHeaders(req.headers.put(host))
  case _ => req
}
```
That's microscopically more efficient (no fold call + no withFilter)

rossabaker: What is the functional objection to `withFilter` on a type that has a reasonable empty?  I agree that @jmcardon's version saves a few nanoseconds, and I agree that `withFilter` on `Future` is regrettable, but I don't have a problem with what @bpholt wrote.

PR #1876, opened 2018-05-25
Add sbt-updates to build and have it state them on project load

ChristopherDavenport:
Its a little verbose, however makes it very easy to keep up with project upgrades.


aeons:
Doesn't this add a significant delay to project reloading?

rossabaker:
I measure 4 seconds.

aeons:
Ok, that's fine then. I guess it's another command from sbt-updates that's super slow then :)

PR #1880, opened 2018-05-27
Non-trivial merge from 0.18.12

rossabaker:


tests/src/test/scala/org/http4s/multipart/MultipartSpec.scala
```diff
@@ -123,25 +118,25 @@ Content-Type: application/pdf
...
-      headers = header)
-
-    val decoded = EntityDecoder[IO, Multipart[IO]].decode(request, true)
-    val result = decoded.value.unsafeRunSync()
-
-    result must beRight
-  }
+        val header = Headers(
+          `Content-Type`(MediaType.multipart.`form-data`
+            .withExtensions(Map("boundary" -> "----WebKitFormBoundarycaZFo8IAKVROTEeD"))))
```
rossabaker: This merge was so bad I started over from what @jmcardon had, and then I summoned @cquiroz to discuss what happened to the old `multipart` convenience constructor, and now I see that it exists.  I guess I should change thes eto look like the old?

jmcardon: If possible yes plz

PR #1882, opened 2018-05-28
Add a client backend for HttpURLConnection

rossabaker:
It's not production grade.  It blocks.  But it comes with no extra dependencies, doesn't need to be shut down, and is nice for a REPL session.


client/src/main/scala/org/http4s/client/HttpURLConnectionClient.scala
```diff
@@ -0,0 +1,57 @@
...
+import cats.implicits._
+import fs2.io.{readInputStream, writeOutputStream}
+import java.net.{HttpURLConnection, URL}
+import scala.collection.JavaConverters._
+
+/** A [[Client]] based on `java.net.HttpUrlConnection` */
+object HttpURLConnectionClient {
+
+  /** Construct a new Client */
+  def apply[F[_]](implicit F: Effect[F]): Client[F] =
```
rossabaker: We probably want a configuration here for timeouts?

core/src/main/scala/org/http4s/internal/package.scala
```diff
@@ -0,0 +1,15 @@
+package org.http4s
+
+import cats.effect.Async
+import cats.implicits._
+import scala.concurrent.ExecutionContext
+
+package object internal {
+  def blocking[F[_], A](fa: F[A], blockingExecutionContext: ExecutionContext)(implicit F: Async[F], ec: ExecutionContext): F[A] =
```
rossabaker: This looks a lot like @ChristopherDavenport's [linebacker 0.1](https://github.com/ChristopherDavenport/linebacker/blob/series/0.1.x/linebacker/src/main/scala/io/chrisdavenport/linebacker/Linebacker.scala#L19-L24) and [the not ideal combinator](https://gist.github.com/SystemFw/256205f51bf0135e4c6fd95dee4590fc) in @SystemFw's gist.

Better can be done with bracket.  Better still can be done with rank-N types.  But maybe this is the best we can do now?

client/src/main/scala/org/http4s/client/JavaNetClient.scala
```diff
@@ -0,0 +1,166 @@
...
+  * back after blocking.
+  */
+sealed abstract class JavaNetClient private (
+  val connectTimeout: Duration,
+  val readTimeout: Duration,
+  val proxy: Option[Proxy],
+  val hostnameVerifier: Option[HostnameVerifier],
+  val sslSocketFactory: Option[SSLSocketFactory],
+  val blockingExecutionContext: ExecutionContext,
+)(implicit ec: ExecutionContext) {
```
rossabaker: This is a good example of what I have in mind for the client and server configurations.  This can be evolved by adding more values.  Generating the `with` methods is tedious, but I think compatibility wins here.

rossabaker: Maybe having types named `BlahClient` create a plain old `Client` is confusing.  We could all this a `JavaNetClientBuilder` or ` JavaNetClientConfig`, but I like the shape.

rossabaker:
It's still not great.  The body I/O blocks.  `readInputStreamAsync` and `writeOutputStreamAsync` shift, but don't shift back.  We could make a double shifting version of those, but not without copying in parts of `JavaInputOutputStream`.

NOTICE
```diff
@@ -0,0 +1,39 @@
+http4s
+Copyright 2013-2018 http4s.org
+Licensed under Apache License 2.0 (see LICENSE)
```
rossabaker: We're technically supposed to include the license of all the projects we copied from.  I took the liberty of doing so while adding yet another project to copy and paste from.

A former employer's lawyer once told me we need to have the copyright boilerplate on all our source files.  I told him not while we were maintaining multiple branches.  But maybe now is a good time.  But that's a discussion for another ticket.

core/src/main/scala/org/http4s/internal/package.scala
```diff
@@ -0,0 +1,95 @@
...
+      implicit F: Async[F],
+      ec: ExecutionContext): F[A] =
+    for {
+      _ <- Async.shift[F](blockingExecutionContext)
+      att <- fa.attempt
+      _ <- Async.shift(ec)
+      fa0 <- F.fromEither(att)
+    } yield fa0
+
+  /** Like fs2.io.readInputStream, but does a double shift with [[blocking]]. */
```
rossabaker: I think everything from here below disappears in fs2-1.0, because it does a double shift.  This is basically all a backport to fs2-0.10.

NOTICE
```diff
@@ -0,0 +1,39 @@
...
+
+This software contains portions of code derived from akka-http
+https://github.com/akka/akka-http
+Copyright (C) 2009-2017 Lightbend Inc. <http://www.lightbend.com>
+Licensed under Apache License 2.0 (see LICENSE)
+
+This software contains portions of code derived from finagle
+https://github.com/twitter/finagle
```
rossabaker: The version we copied from has no copyright. :confused:

rossabaker:
Bump.  I dismissed @SystemFw's earlier approval due to substantial changes.

SystemFw:
It's a bit of a shame since the main use for this is probably the repl, where you normally just import `global`, but it is what it is

rossabaker:
Yeah, it made it slightly more awkward to use in the REPL, but it'll be a bit more consistent with our other java.io builders.

You know, we could have a `repl` constructor for this one that just blocks the current thread because lol nothing matters.

PR #1884, opened 2018-05-30
Move ServerSoftware to top-level package

rossabaker:
http4s-core shouldn't have anything in the `org.http4s.server` package.


rossabaker:
Broken because #1887.

rossabaker:
If they passed appveyor, and Travis failed on the tut from #1887, I'm counting that as a pass, since it's obvious what the problem is.

PR #1885, opened 2018-05-30
Generalize AutoSlash middleware

rossabaker:


rossabaker:
Build broken because #1887.

PR #1886, opened 2018-05-30
Parameterize ChunkAggregator

rossabaker:


server/src/test/scala/org/http4s/server/middleware/ChunkAggregatorSpec.scala
```diff
@@ -47,7 +48,7 @@ class ChunkAggregatorSpec extends Http4sSpec {
 
     "handle a none" in {
       val routes: HttpRoutes[IO] = HttpRoutes.empty
-      ChunkAggregator(routes).run(Request()).value must returnValue(None)
+      ChunkAggregator(OptionT.liftK[IO])(routes).run(Request()).value must returnValue(None)
```
rossabaker: Do we need some partial application trick for this one?  Ugh.

rossabaker:
Build will be broken because of #1887.

server/src/main/scala/org/http4s/server/middleware/ChunkAggregator.scala
```diff
@@ -2,32 +2,32 @@ package org.http4s
...
 import org.http4s.headers._
 
 object ChunkAggregator {
-  def apply[F[_]](@deprecatedName('service) routes: HttpRoutes[F])(
-      implicit F: Effect[F]): HttpRoutes[F] =
-    routes.flatMapF { response =>
-      OptionT.liftF(
+  def apply[F[_], G[_], A](f: G ~> F)(@deprecatedName('service) http: Kleisli[F, A, Response[G]])(
+      implicit F: FlatMap[F],
+      G: Sync[G]): Kleisli[F, A, Response[G]] =
```
aeons: Same as the other one, could/should these implicits be context bounds instead.

aeons:
I tried the partial application thing from cats, but that still crashes the compiler.
If you add the type ascription:
```scala
val nonStreamFileHttpEndpoint: HttpRoutes[F] = ChunkAggregator(OptionT.liftK[F])(fileHttpEndpoint)
```
it works as well.

PR #1887, opened 2018-05-30
Fix deprecation in dsl tut

rossabaker:
This should unbreak master from the last merge.


rossabaker:
master is broken anyway because of deprecation, and you can see the trainwreck behind it.  This is is now broken by a latent issue captured in #1891.

I think we should merge this and start clearing out the above linked tickets.  Travis can't get any redder.

Sorry, everyone.  I moved fast and broke things.

PR #1888, opened 2018-05-30
Upgrade to async-http-client-2.4.8

rossabaker:


rossabaker:
Will need #1887 for clean build.

PR #1889, opened 2018-05-30
Generalize CORS middleware

rossabaker:


rossabaker:
Was submitted while #1887 was broken.

server/src/main/scala/org/http4s/server/middleware/CORS.scala
```diff
@@ -38,12 +38,14 @@ object CORS {
...
     * Currently, you cannot make permissions depend on request details
     */
-  def apply[F[_]](
-      @deprecatedName('service) routes: HttpRoutes[F],
-      config: CORSConfig = DefaultCORSConfig)(implicit F: Applicative[F]): HttpRoutes[F] =
+  def apply[F[_], G[_]](
+      @deprecatedName('service) http: Http[F, G],
+      config: CORSConfig = DefaultCORSConfig)(
+      implicit F: Applicative[F],
+      G: Functor[G]): Http[F, G] =
```
aeons: Could these implicits be added as context bounds instead?

Issue #1890, opened 2018-05-30
Generalize middlewares


rossabaker:
Most of our middlewares could be more parametric.

- By expanding `HttpRoutes` with `Http[F, G]`, the same middleware can be applied to either `HttpRoutes` or `HttpApp`.
- By expanding `Http[F, G]` to either `Kleisli[F, Request[G], B]` or `Kleisli[F, A, Response[G]]`, middleware that only affects one side can compose with Kleislis of different outputs or inputs.  The latter is particularly helpful for applying the same middleware to both regular and authorized HTTP functions.
- Increased parametricity makes it more clear what the middleware can do from the type alone.

This is a mostly mechanical translation, and would make a great first issue for anyone looking to better understand type classes and constraints.  If you claim a middleware and offer to submit a PR, I will help you if you get stuck.

Server:
* [x] AutoSlash.scala
* [x] ChunkAggregator.scala
* [x] CORS.scala
* [x] CSRF.scala
* [x] DefaultHead.scala
* [x] EntityLimiter.scala
* [x] GZip.scala
* [x] HSTS.scala
* [x] Jsonp.scala
* [x] Logger.scala
* [x] PushSupport.scala
* [x] RequestLogger.scala
* [x] ResponseLogger.scala
* [x] Timeout.scala
* [x] URITranslation.scala
* [x] UrlFormLifter.scala
* [x] VirtualHost.scala

Client:
* [x] FollowRedirect.scala
* [ ] Logger.scala
* [ ] package.scala
* [ ] RequestLogger.scala
* [ ] ResponseLogger.scala
* [ ] Retry.scala

rossabaker:
Wow, I should ask for help before going to sleep more often!  This is great.

To make sure we're not duplicating work, please leave a comment here if you're grabbing one.

SystemFw:
Yeah this is awesome 😄

tomwadeson:
I'll pick up `VirtualHost` and `Timeout` 👍

aworton:
Happy to have a go at Logger, RequestLogger and ResponseLogger

akilegaspi:
I'm looking to work with generalizing the HTTP stuff into Kleisli

somethingconcon:
grabbing PushSupport

tomwadeson:
I'll take `CSRF`

calvinbrown085:
I'd be happy to try `ResponseLogger.scala`

rossabaker:
[...]

PR #1892, opened 2018-05-30
Parameterize EntityLimiter

lloydmeta:
Relates to #1890

### DIFF
--- a/server/src/main/scala/org/http4s/server/middleware/EntityLimiter.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/EntityLimiter.scala
@@ -2,7 +2,9 @@ package org.http4s
 package server
 package middleware
 
+import cats.data.Kleisli
 import fs2._
+
 import scala.util.control.NoStackTrace
 
 object EntityLimiter {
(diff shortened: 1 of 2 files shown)


server/src/main/scala/org/http4s/server/middleware/EntityLimiter.scala
```diff
@@ -2,7 +2,9 @@ package org.http4s
 package server
 package middleware
 
+import cats.data.Kleisli
 import fs2._
+
```
aeons: Imports should be ordered alphabetically (I know that's true here, but just stating it :)) and should not be grouped into java, scala, scala stdlib.

jmcardon: Probably an automatic IJ thing? My commits do this too occasionally and it drives Ross mad :smile:

website/src/hugo/content/changelog.md
```diff
@@ -33,6 +33,7 @@ it.
 * Add `BlockingHttp4sServlet` for use in Google App Engine and Servlet 2.5 containers.  Rename `Http4sServlet` to `AsyncHttp4sServlet`. [#1830](https://github.com/http4s/http4s/pull/1830)
 * Generalize `Logger` middleware to log with `String => Unit` instead of `logger.info(_)` [#1839](https://github.com/http4s/http4s/pull/1839)
 * Rename `RequestLogger.apply0` and `ResponseLogger.apply0` to `RequestLogger.apply` and `ResponseLogger.apply`.  [#1837](https://github.com/http4s/http4s/pull/1837)
+* Parameterize `EntityLimiter` middleware to work on `Kleisli[F, Request[G], B]`. [#1892](https://github.com/http4s/http4s/pull/1892)
```
jmcardon: Hell yeah Commonwealth English (Paramet**e**rize) long live the Queen! :gb:

server/src/main/scala/org/http4s/server/middleware/EntityLimiter.scala
aeons: It's the default in IJ.

It can be changed in Settings > Editor > Code Style > Scala > Imports, by removing all lines from the Import Layout box, until only `all other imports` is left.

rossabaker: Is there any way we can check in a project style setting for that?

If not, I might be ready to give up and do it the IntelliJ default way.  (Which is a complicated and unappealing default, but not so bad it's worth asking every individual contributor to change it.)

aeons: I'll have a look

PR #1893, opened 2018-05-30
Parameterize HSTS middleware

lloydmeta:
Relates to #1890


server/src/main/scala/org/http4s/server/middleware/HSTS.scala
```diff
@@ -2,9 +2,10 @@ package org.http4s
 package server
 package middleware
 
-import cats._
+import cats.Functor
 import cats.data.Kleisli
 import org.http4s.headers.`Strict-Transport-Security`
+
```
aeons: Same comment as on the other one. :)

PR #1895, opened 2018-05-30
Parameterize `URITranslation` middleware

tomwadeson:

### DIFF
--- a/server/src/main/scala/org/http4s/server/middleware/URITranslation.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/URITranslation.scala
@@ -3,17 +3,18 @@ package server
 package middleware
 
 import cats.Functor
+import cats.data.Kleisli
 
 object URITranslation {
-  def translateRoot[F[_]: Functor](prefix: String)(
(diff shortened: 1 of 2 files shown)


server/src/main/scala/org/http4s/server/middleware/URITranslation.scala
```diff
@@ -3,17 +3,18 @@ package server
 package middleware
 
 import cats.Functor
+import cats.data.Kleisli
 
 object URITranslation {
-  def translateRoot[F[_]: Functor](prefix: String)(
-      @deprecatedName('service) routes: HttpRoutes[F]): HttpRoutes[F] = {
+  def translateRoot[F[_], G[_]: Functor, A](prefix: String)(
```
rossabaker: I think I called this `B` in another one, just so response was always `B`.  But I also thought about just using the first available letter like this.  They're obviously equivalent.

Does anyone have a strong preference?  My primary preference is only for consistency.  I guess maybe I like the style here better.

tomwadeson: No preference here. With you on the consistency front, so am happy to change!

PR #1898, opened 2018-05-30
Parameterize DefaultHead #1890

somethingconcon:
Parameterize DefaultHead apply to take the shape Kleisli[F, Request[G], B] returning the same.

WIP as I wanted to see if there were any spec changes requested with this.  DefaultHeadSpec is failing due to an improper signature.  Can someone point me in the right direction for the proper implicit setup? (might be a silly question sorry)


server/src/main/scala/org/http4s/server/middleware/DefaultHead.scala
```diff
@@ -14,19 +14,18 @@ import cats.implicits._
...
     Kleisli { req =>
-      req.method match {
-        case Method.HEAD =>
-          (routes <+> headAsTruncatedGet(routes))(req)
-        case _ =>
-          routes(req)
+      http(req) <+> {
+        req.method match {
+          case Method.HEAD => {
+            http(req.withMethod(Method.GET).withEmptyBody)
```
rossabaker: This is going to invoke the service with a request with an empty body.  We want to send a `GET` request, but return a `response` with an empty body.  Furthermore, we want to drain the response body, so any resources evaluated with the route are released.  So I think we want to keep the `headAsTruncatedGet`.

somethingconcon: Ah, ok. Makes sense. Fixing.

PR #1899, opened 2018-05-30
Parameterize `Timeout` middleware

tomwadeson:

I've only generalised the non-deprecated, public API so far. Should I continue?

### DIFF
--- a/examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/Module.scala
+++ b/examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/Module.scala
@@ -17,7 +17,10 @@ import org.http4s.server.middleware.{AutoSlash, ChunkAggregator, GZip, Timeout}
 
 import scala.concurrent.duration._
 
-class Module[F[_]](client: Client[F])(implicit F: ConcurrentEffect[F], S: Scheduler) {
+class Module[F[_]](client: Client[F])(
+    implicit F: ConcurrentEffect[F],
+    S: Scheduler,
+    T: Timer[OptionT[F, ?]]) {
 
   private val fileService = new FileService[F]
 
--- a/examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/Module.scala
+++ b/examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/Module.scala
@@ -17,7 +17,7 @@ import org.http4s.server.middleware.{AutoSlash, ChunkAggregator, GZip, Timeout}
 
 import scala.concurrent.duration._
 
-class Module[F[_]](client: Client[F])(implicit F: ConcurrentEffect[F], S: Scheduler, T: Timer[F]) {
+class Module[F[_]](client: Client[F])(implicit F: ConcurrentEffect[F], S: Scheduler) {
 
   private val fileService = new FileService[F]
 
--- a/examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/Server.scala
+++ b/examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/Server.scala
@@ -13,10 +13,9 @@ class HttpServer[F[_]](implicit F: ConcurrentEffect[F]) extends StreamApp[F] {
 
   override def stream(args: List[String], requestShutdown: F[Unit]): Stream[F, ExitCode] =
     Scheduler(corePoolSize = 2).flatMap { implicit scheduler =>
-      implicit val T = Timer.derive[F](F, IO.timer)
       for {
         client <- Http1Client.stream[F]()
-        ctx <- Stream(new Module[F](client)(F, scheduler, T))
+        ctx <- Stream(new Module[F](client))
         exitCode <- BlazeBuilder[F]
           .bindHttp(8080, "0.0.0.0")
           .mountService(ctx.fileHttpEndpoint, s"/${endpoints.ApiVersion}")
--- a/examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/Server.scala
+++ b/examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/Server.scala
@@ -1,10 +1,12 @@
 package com.example.http4s.blaze.demo.server
 
+import cats.data.OptionT
 import cats.effect._
 import fs2.StreamApp.ExitCode
 import fs2.{Scheduler, Stream, StreamApp}
 import org.http4s.client.blaze.Http1Client
 import org.http4s.server.blaze.BlazeBuilder
+
 import scala.concurrent.ExecutionContext.Implicits.global
 
 object Server extends HttpServer[IO]
@@ -13,6 +15,7 @@ class HttpServer[F[_]](implicit F: ConcurrentEffect[F]) extends StreamApp[F] {
 
   override def stream(args: List[String], requestShutdown: F[Unit]): Stream[F, ExitCode] =
     Scheduler(corePoolSize = 2).flatMap { implicit scheduler =>
+      implicit val T = Timer.derive[OptionT[F, ?]]
       for {
         client <- Http1Client.stream[F]()
         ctx <- Stream(new Module[F](client))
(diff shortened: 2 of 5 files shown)


server/src/main/scala/org/http4s/server/middleware/Timeout.scala
```diff
@@ -87,13 +88,15 @@ object Timeout {
...
-    val OTC = Concurrent[OptionT[F, ?]]
-    routes
-      .mapF(respF => OTC.race(respF, OptionT.liftF(T.sleep(timeout) *> timeoutResponse)))
+  def apply[F[_], G[_], A](timeout: FiniteDuration, timeoutResponse: F[Response[G]])(
+      @deprecatedName('service) http: Kleisli[F, A, Response[G]])(
+      implicit F: Concurrent[F]): Kleisli[F, A, Response[G]] = {
+
+    import scala.concurrent.ExecutionContext.Implicits.global
+    val T = Timer.derive[F]
+
```
tomwadeson: I was seeing the following error:

```
[error] /Users/tom/Projects/http4s/examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/Module.scala:48:22: Cannot find implicit value for Timer[[A]cats.data.OptionT[F,A]].
[error] Note that [A]cats.data.OptionT[F,A] needs to be a cats.effect.Async data type. You might also
[error] need a scala.concurrent.ExecutionContext in scope, or equivalent, try to
[error] import scala.concurrent.ExecutionContext.Implicits.global
[error]     Timeout(1.second)(timeoutHttpEndpoint)
```

We already require `Concurrent[F]`, which implies `LiftIO[F]`, so we're able to `Timer.derive[F]` where we need it.

If we're happy to do this, is it OK to use the global `ExecutionContext` here, or should we require a `Strategy`/`ExecutionStrategy` be passed in?

jmcardon: I would prefer you simply take a `Timer` instead of calling `derive` since it opens feeding in a different one if need be.

tomwadeson: Right! And on the issue of there not being an instance for `OptionT[F, ?]` - do you have any suggestions on how best to provide one, or would we leave this to users?

jmcardon: that's in userland/another PR I think.

tomwadeson: I get you. I'm happy to defer to your judgement on this, but would say that:

1. I think I'd argue that it's an implementation detail of the `Timeout` middleware and it's not immediately obvious to me why I might want to switch one `Timer` out for another
2. I think the ergonomics are better this way: it's cleaned up `Module` and `Server` in this PR and I think I prefer this API as a user

If that's not convincing, that's fine, I'll change it 😄

jmcardon: Counter args:

1. the global ec is often used as the general client pool in something like blaze. Where a user chooses to execute stuff is _not_ an implementation detail in my eyes. It impacts the app, even if it's true that a good chunk of the times  people do use global. Whenever you `Timer.sleep`, you force the continuation of that call onto the other pool provided by the `timer`. For advanced users that _do_ configure thread pools, this is a net loss.
2. What does it clean up aside from removing a single implicit? 

I'm in general :-1: on explicitly less powerful configuration. It's a slight win for beginner users by limiting advanced users by forcing them onto `global`. Moreover, I don't particularly like considering application configuration (which timer propagates, essentially since it does carry a thread pool reference no matter how much anyone wants to pretend it doesn't) "an implementation detail".

This is a democratic decision though, so other maintainer opinions would be nice.

tomwadeson: I don't disagree with any of that! To be clear, I'm only arguing for the `Timer` to be defined locally. The `ExecutionContext` can absolutely be passed in.

I'd also welcome other POVs. I'm acutely aware that I'm arguing from a position of ignorance on this!

jmcardon: but that's the thing though: in the ecosystem we're trying to move away from passing `ExecutonContext`s and passing `Timer[F]`s. Hence my quip.

tomwadeson: I see! I'll change it.

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/Module.scala
```diff
@@ -17,7 +17,10 @@ import org.http4s.server.middleware.{AutoSlash, ChunkAggregator, GZip, Timeout}
 
 import scala.concurrent.duration._
 
-class Module[F[_]](client: Client[F])(implicit F: ConcurrentEffect[F], S: Scheduler, T: Timer[F]) {
+class Module[F[_]](client: Client[F])(
+    implicit F: ConcurrentEffect[F],
+    S: Scheduler,
+    T: Timer[OptionT[F, ?]]) {
```
jmcardon: Sorry, is there a reason this is necessary in particular to be `OptionT`? I'm missing something not immediately clear upon inspection.

Also, `Timer` is there to replace fs2 scheduler eventually. Maybe in another PR we can proceed to clean up examples, but when you have one, the other is no longer needed

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/Server.scala
```diff
@@ -13,10 +15,10 @@ class HttpServer[F[_]](implicit F: ConcurrentEffect[F]) extends StreamApp[F] {
 
   override def stream(args: List[String], requestShutdown: F[Unit]): Stream[F, ExitCode] =
     Scheduler(corePoolSize = 2).flatMap { implicit scheduler =>
-      implicit val T = Timer.derive[F](F, IO.timer)
+      implicit val T = Timer.derive[OptionT[F, ?]]
```
jmcardon: Again, this derivation seems only because it was changed to `OptionT[F, ?]`.

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/Module.scala
tomwadeson: This is `Timeout` pre-generalisation:

```scala
  def apply[F[_]](timeout: FiniteDuration, timeoutResponse: F[Response[F]])(
      @deprecatedName('service) routes: HttpRoutes[F])(
      implicit F: Concurrent[F],
      T: Timer[F]): HttpRoutes[F] =???
```

`F` is the non-`OptionT`-adorned `F` that parameterises `HttpRoutes[F]` (which expands to `Http[OptionT[F, ?], F]`), so it doesn't matter that there's no `Timer[OptionT[F, ?]]` because we don't require evidence for one.

But by having now generalised it like so:

```scala
  def apply[F[_], G[_], A](timeout: FiniteDuration, timeoutResponse: F[Response[G]])(
      @deprecatedName('service) http: Kleisli[F, A, Response[G]])(
      implicit F: Concurrent[F],
      T: Timer[F]): Kleisli[F, A, Response[G]] = ???
```

... then when `F` is instantiated within `Module` (and I guess most places `Timeout` is used in userland too), it becomes `OptionT[F, ?]`, and we now require evidence of `Timer[OptionT[F, ?]]`:

```scala
  private val timeoutHttpEndpoint: HttpRoutes[F] =
    new TimeoutHttpEndpoint[F].service

  private val timeoutEndpoints: HttpRoutes[F] =
    Timeout(1.second)(timeoutHttpEndpoint)
```

This is quite gnarly though (and what prompted me to `Timer.derive[F]` at the use-site previously. ) Is there a simpler solution that I'm not seeing? I guess the real fix would be to have transformer instances for `Timer` defined upstream in cats.

jmcardon: Nope, 100% my bad I misread this. I just thought `Timeout` wasn't being used at all and/or it was being passed a `Http[F, F]`. You're right :+1:

tomwadeson: Cool. Happy to tackle the clean up of examples in a separate PR, as you suggest 😄

PR #1900, opened 2018-05-30
Fix MimeGenerator flags

cquiroz:
The flags `Uncompressible` and `NotBinary` were wrongly set. This PR fixes that and updates `MimeDB` accordingly. There are some changes on the upstream db as reflected on some changes in the resulting file

/cc @sderosiaux

### DIFF
--- a/core/src/main/scala/org/http4s/MimeDB.scala
+++ b/core/src/main/scala/org/http4s/MimeDB.scala
@@ -5,9 +5,9 @@ private[http4s] trait MimeDB {
   lazy val allMediaTypes
     : List[MediaType] = Nil ++ x_shader.all ++ x_conference.all ++ video.all ++ text.all ++ multipart.all ++ model.all ++ message.all ++ image.all ++ font.all ++ chemical.all ++ audio.all ++ application.all
   val Compressible: Boolean = true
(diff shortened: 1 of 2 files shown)


sderosiaux:
Glad to help! And thanks for the fix.

core/src/main/scala/org/http4s/MimeDB.scala
```diff
@@ -5,9 +5,9 @@ private[http4s] trait MimeDB {
   lazy val allMediaTypes
     : List[MediaType] = Nil ++ x_shader.all ++ x_conference.all ++ video.all ++ text.all ++ multipart.all ++ model.all ++ message.all ++ image.all ++ font.all ++ chemical.all ++ audio.all ++ application.all
   val Compressible: Boolean = true
-  val Uncompressible: Boolean = true
+  val Uncompressible: Boolean = false
```
rossabaker: I guess maybe I thought these were sum types instead of boolean constants.  Would stronger types help here?

rossabaker: Maybe?  Booleans are a smell.  I think either way, this fixes a bug and should be merged.

PR #1902, opened 2018-05-30
Parameterize `VirtualHost` middleware

tomwadeson:

### DIFF
--- a/server/src/main/scala/org/http4s/server/middleware/VirtualHost.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/VirtualHost.scala
@@ -22,7 +22,7 @@ object VirtualHost {
     * security of the underlying transport protocol.
     */
   final case class HostService[F[_], G[_]](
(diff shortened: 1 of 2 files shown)


server/src/main/scala/org/http4s/server/middleware/VirtualHost.scala
```diff
@@ -20,53 +21,51 @@ object VirtualHost {
     * filled in, if possible, using the request Uri or knowledge of the
     * security of the underlying transport protocol.
     */
-  final case class HostService[F[_]](
-      @deprecatedName('service) routes: HttpRoutes[F],
+  final case class HostService[F[_], G[_]](
+      @deprecatedName('service) http: Kleisli[F, Request[G], Response[G]],
```
rossabaker: Do we want to use the `Http[F, G]` alias instead of `Kleisli[F, Request[G], Response[G]]`?  I'm inclined toward yes.

PR #1904, opened 2018-05-31
Parameterize `jsonp` middleware

amitrai48:
relates to #1890. 

I am not sure though why an `implicit W: EntityEncoder[F, String])` was being used in the earlier apply method. Even after removing it all the tests were successfully executed.


server/src/main/scala/org/http4s/server/middleware/Jsonp.scala
```diff
@@ -26,24 +26,23 @@ object Jsonp {
...
             }
-          }
+          }.apply(req)
+
         case Some(invalidCallback) =>
           logger.warn(s"Jsonp requested with invalid callback function name $invalidCallback")
-          OptionT.pure(Response[F](Status.BadRequest).withEntity(s"Not a valid callback name."))
-        case None =>
-          routes(req)
+          http.map(response => response.withStatus(Status.BadRequest).withEntity(s"Not a valid callback name.")).apply(req)
```
rossabaker: This applies the `http` only to throw it away.  I think we want to add a `G ~> F` parameter, like `ChunkAggregator` has.  And then we can use that to lift the bad request response into `F` without calling the `http`.

rossabaker:
@amitrai48 This looks good. We just need to run `test:scalafmt` on it now, and it should be ready to merge.

PR #1909, opened 2018-06-01
Parameterize `CSRF` middleware

tomwadeson:


tomwadeson:
An issue that I'm about to bump into is bringing the tests (and examples?) inline with this new generalised API. 

To get the types to align, due to `HttpRoutes`, I'm going to have to instantiate a `CSRF` with:

```scala
  val csrf = CSRF.withGeneratedKey[OptionT[IO, ?], IO](clock = testClock).value.unsafeRunSync().get
```

... which is fairly disgusting (and I think will need to be repeated several times through the test).

Any suggestions for things I can do to mitigate this in the implementation?

tomwadeson:

rossabaker:
Instead of `HttpRoutes`, can you just use `HttpApp`?  (You can pass a total function to `HttpApp.apply`, or call `.orNotFound` on an `HttpRoutes`).

Issue #1878, opened 2018-05-26
Incosistencies Between Json Encoding Content-Type Headers


ChristopherDavenport:
Argonaut places a Charset into the header for its encoder - [Argonaut Instance](https://github.com/http4s/http4s/blob/v0.18.11/argonaut/src/main/scala/org/http4s/argonaut/ArgonautInstances.scala#L38)

While Circe/Json4s does not apply the charset - [Circe Instance](https://github.com/http4s/http4s/blob/v0.18.11/circe/src/main/scala/org/http4s/circe/CirceInstances.scala#L72) - [Json4s Instance](https://github.com/http4s/http4s/blob/v0.18.11/json4s/src/main/scala/org/http4s/json4s/Json4sInstances.scala#L57). Is there a reason for this incosistency that I am unaware of? 

As this inconsistency might surprise folks switching between encoders if we want to pick a standard or identify why those differences exist that is likely a good thing for our users.

rossabaker:
It's meaningless to add.  [Quoth the RFC](https://tools.ietf.org/html/rfc7159): 

>  No "charset" parameter is defined for this registration. Adding one really has no effect on compliant recipients.

On the other hand, if we're not explicit, `message.charset` will return `None`, which means we have to just guess when decoding to a String.  We don't want to be in the business of special casing in core every media type that internally defines its character encoding.

I strongly think we should be consistent, and weakly think we should leave it off.

ChristopherDavenport:
Easy solution is to remove the content encoding from argonaut then.

rossabaker:
I think we could target release-0.18.x with it.

calvinbrown085:
I'd be happy to try this one out @rossabaker @ChristopherDavenport

PR #1914, opened 2018-06-06
Removing charset for argonaut encoders

calvinbrown085:
fixes https://github.com/http4s/http4s/issues/1878

We wanted to keep consistency with our Content Type headers across JSON implementations.

### DIFF
--- a/argonaut/src/test/scala/org/http4s/argonaut/ArgonautSpec.scala
+++ b/argonaut/src/test/scala/org/http4s/argonaut/ArgonautSpec.scala
@@ -21,8 +21,9 @@ class ArgonautSpec extends JawnDecodeSupportSpec[Json] with Argonauts {
   "json encoder" should {
     val json = Json("test" -> jString("ArgonautSupport"))
 
-    "not have json content type" in {
-      jsonEncoder.headers.get(`Content-Type`) must_== None
+    "have json content type" in {
+      jsonEncoder.headers.get(`Content-Type`) must_== Some(
+        `Content-Type`(MediaType.application.json))
     }
 
     "write compact JSON" in {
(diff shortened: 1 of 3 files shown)


calvinbrown085:
@ChristopherDavenport @rossabaker Would you guys mind taking a look here to make sure I did the correct thing?

argonaut/src/test/scala/org/http4s/argonaut/ArgonautSpec.scala
```diff
@@ -21,8 +21,9 @@ class ArgonautSpec extends JawnDecodeSupportSpec[Json] with Argonauts {
   "json encoder" should {
     val json = Json("test" -> jString("ArgonautSupport"))
 
-    "not have json content type" in {
-      jsonEncoder.headers.get(`Content-Type`) must_== None
+    "have json content type" in {
+      jsonEncoder.headers.get(`Content-Type`) must_== Some(
+        `Content-Type`(MediaType.application.json))
```
rossabaker: Wow.  I'm trying to remember the justification for the old test.  Why hasn't it always been this way?

argonaut/src/test/scala/org/http4s/argonaut/ArgonautSpec.scala
```diff
@@ -34,21 +34,21 @@ class ArgonautSpec extends JawnDecodeSupportSpec[Json] with Argonauts {
       val custom = ArgonautInstances.withPrettyParams(PrettyParams.spaces2)
       import custom._
       writeToString(json) must_== ("""{
-          |  "test" : "ArgonautSupport"
-          |}""".stripMargin)
+                                     |  "test" : "ArgonautSupport"
```
aeons: Hmm, why is scalafmt happy with these changes?

calvinbrown085: @aeons Not sure, I did a `;test;scalafmt` and it passed.

ChristopherDavenport: I would prefer perhaps these be defined above and then checked against so the formatting could be cleaner. As that inset is not fun to read at the end of a test.

calvinbrown085:
@ChristopherDavenport Not sure what you mean there

ChristopherDavenport:
Unimportant. No worries.

PR #1916, opened 2018-06-07
Altered Logger, RequestLogger and ResponseLogger to be more parametric using Kleisli

aworton:
Couple of concerns:

1. Lost the handling of a none in F inside RequestLogger. Would this need to be handled in the passed in natural transformation?

2. Existing spec only appears to assert that there is no impact on the requests, not that the logging occurs. Also doesn't exercise the !logBody paths.

Thanks in advance for review and feedback.

### DIFF
--- a/server/src/main/scala/org/http4s/server/middleware/Logger.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/Logger.scala
@@ -4,11 +4,9 @@ package middleware
 
 import cats.data.Kleisli
 import cats.effect._
-import cats.{Monad, ~>}
 import fs2._
 import org.http4s.util.CaseInsensitiveString
 import org.log4s.getLogger
-
 import scala.concurrent.ExecutionContext
 
 /**
@@ -17,21 +15,22 @@ import scala.concurrent.ExecutionContext
 object Logger {
   private[this] val logger = getLogger
 
-  def apply[F[_] : Sync, G[_] : Effect](f: G ~> F, logAction: String => Unit = logger.info(_))(
-    logHeaders: Boolean,
-    logBody: Boolean,
-    redactHeadersWhen: CaseInsensitiveString => Boolean = Headers.SensitiveHeaders.contains
-  )(@deprecatedName('httpService) http: Kleisli[F, Request[G], Response[G]])(
-                                       implicit ec: ExecutionContext): Kleisli[F, Request[G], Response[G]] =
-    ResponseLogger(f, logAction)(logHeaders, logBody, redactHeadersWhen)(
-      RequestLogger(f, logAction)(logHeaders, logBody, redactHeadersWhen)(http)
+  def apply[F[_]: Effect](
+      logHeaders: Boolean,
+      logBody: Boolean,
+      redactHeadersWhen: CaseInsensitiveString => Boolean = Headers.SensitiveHeaders.contains,
+      logAction: String => Unit = logger.info(_)
+  )(@deprecatedName('httpService) http: Kleisli[F, Request[F], Response[F]])(
+      implicit ec: ExecutionContext): Kleisli[F, Request[F], Response[F]] =
+    ResponseLogger(logHeaders, logBody, redactHeadersWhen, logAction)(
+      RequestLogger(logHeaders, logBody, redactHeadersWhen, logAction)(http)
     )
 
-  def logMessage[F[_] : Sync, A <: Message[F]](message: A)(
-    logHeaders: Boolean,
-    logBody: Boolean,
-    redactHeadersWhen: CaseInsensitiveString => Boolean = Headers.SensitiveHeaders.contains)(
-                                                log: String => Unit): F[Unit] = {
+  def logMessage[F[_], A <: Message[F]](message: A)(
+      logHeaders: Boolean,
+      logBody: Boolean,
+      redactHeadersWhen: CaseInsensitiveString => Boolean = Headers.SensitiveHeaders.contains)(
+      log: String => Unit)(implicit F: Sync[F]): F[Unit] = {
 
     val charset = message.charset
     val isBinary = message.contentType.exists(_.mediaType.binary)
@@ -69,7 +68,7 @@ object Logger {
       Stream("").covary[F]
     }
 
-    if (!logBody && !logHeaders) Monad[F].unit
+    if (!logBody && !logHeaders) F.unit
     else {
       bodyText
         .map(body => s"$prelude $headers $body")
--- a/server/src/main/scala/org/http4s/server/middleware/Logger.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/Logger.scala
@@ -2,30 +2,36 @@ package org.http4s
 package server
 package middleware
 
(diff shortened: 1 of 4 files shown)


server/src/main/scala/org/http4s/server/middleware/RequestLogger.scala
```diff
@@ -16,65 +18,59 @@ import scala.concurrent.ExecutionContext
...
-                  OptionT.liftF(
-                    Logger.logMessage[F, Request[F]](req.withBodyStream(newBody))(
+                  f(
+                    Logger.logMessage[G, Request[G]](req.withBodyStream(newBody))(
                       logHeaders,
                       logBody,
-                      redactHeadersWhen)(logger.info(_)) *>
-                      Sync[F].raiseError[Response[F]](e)
+                      redactHeadersWhen)(logAction) *>
+                      Sync[G].raiseError[Response[G]](e)
```
rossabaker: When I need a method off the instance, we usually name the instance as a parameter.  So instead of `G[_]: Effect`, I would add an `implicit G: Effect[G]`, and then this is just `G.raiseError(e)`.

server/src/main/scala/org/http4s/server/middleware/RequestLogger.scala
```diff
@@ -16,65 +18,59 @@ import scala.concurrent.ExecutionContext
...
-  )(@deprecatedName('service) routes: HttpRoutes[F])(
-      implicit ec: ExecutionContext = ExecutionContext.global): HttpRoutes[F] =
-    Kleisli { req =>
-      if (!logBody)
-        OptionT(Logger
-          .logMessage[F, Request[F]](req)(logHeaders, logBody)(logger.info(_)) *> routes(req).value)
-      else
-        OptionT
-          .liftF(async.refOf[F, Vector[Segment[Byte, Unit]]](Vector.empty[Segment[Byte, Unit]]))
+  def apply[F[_] : Sync, G[_] : Effect, A](f: G ~> F, logAction: String => Unit = logger.info(_))(
```
rossabaker: I suspect `Effect` isn't actually needed here.  I found on #1913 that a `Sync` is sufficient.  This is still on the older version of fs2, but I bet it's also true there.  Might be worth a try to relax the constraint.

server/src/main/scala/org/http4s/server/middleware/ResponseLogger.scala
```diff
@@ -16,42 +18,44 @@ import scala.concurrent.ExecutionContext
...
 
-  def apply[F[_]](
-      logHeaders: Boolean,
-      logBody: Boolean,
-      redactHeadersWhen: CaseInsensitiveString => Boolean = Headers.SensitiveHeaders.contains
-  )(@deprecatedName('service) routes: HttpRoutes[F])(
-      implicit F: Effect[F],
-      ec: ExecutionContext = ExecutionContext.global): HttpRoutes[F] =
+  def apply[F[_]: Sync, G[_]: Effect, A](f: G ~> F, logAction: String => Unit = logger.info(_))(
+                            logHeaders: Boolean,
```
rossabaker: You'll want to run `test:scalafmt` before the final push.  CI enforces the code formatting for consistency.

server/src/main/scala/org/http4s/server/middleware/RequestLogger.scala
aworton: Will implement for consistency.

server/src/main/scala/org/http4s/server/middleware/RequestLogger.scala
aworton: 👍

server/src/main/scala/org/http4s/server/middleware/ResponseLogger.scala
aworton: Ah, thanks. Will do.

aworton:
Agree that it's better to go with more accurate usage.

aworton:
Tried relaxing Effect to Sync, but without current success.
Modified Logger signatures to be restricted to HttpApp for input and output, but kept signatures in Kleisli format.
Altered tests to use HttpApp rather than HttpRoute.
Ran scalafmt.

If I understand correctly, switching from HttpRoutes to HttpApp means the ability to selectively apply logging to particular routes only is lost. Although I'm not sure if that is that likely to be a common usage, and applying the loggers once to the HttpApp simplifies use.

rossabaker:
I think the build failure is intermittent.  Rerunning.

> If I understand correctly, switching from HttpRoutes to HttpApp means the ability to selectively apply logging to particular routes only is lost. Although I'm not sure if that is that likely to be a common usage, and applying the loggers once to the HttpApp simplifies use.

That's correct.  I think getting rid of the `G ~> F`, which itself is to support a hack, is worthwhile here.

server/src/main/scala/org/http4s/server/middleware/Logger.scala
```diff
@@ -2,30 +2,35 @@ package org.http4s
...
       logBody: Boolean,
-      redactHeadersWhen: CaseInsensitiveString => Boolean = Headers.SensitiveHeaders.contains
-  )(@deprecatedName('httpService) httpRoutes: HttpRoutes[F]): HttpRoutes[F] =
-    ResponseLogger(logHeaders, logBody, redactHeadersWhen)(
-      RequestLogger(logHeaders, logBody, redactHeadersWhen)(
-        httpRoutes
-      )
+      redactHeadersWhen: CaseInsensitiveString => Boolean = Headers.SensitiveHeaders.contains,
+      logAction: String => Unit = logger.info(_)
+  )(@deprecatedName('httpService) http: Kleisli[F, Request[F], Response[F]])(
```
jmcardon: You could use the `HttpApp` alias here, or alternatively, turn this into the parametric `F ~> G` style we have in a bunch of places now.

server/src/main/scala/org/http4s/server/middleware/RequestLogger.scala
```diff
@@ -19,16 +19,18 @@ object RequestLogger {
   def apply[F[_]: Effect](
       logHeaders: Boolean,
       logBody: Boolean,
-      redactHeadersWhen: CaseInsensitiveString => Boolean = Headers.SensitiveHeaders.contains
-  )(@deprecatedName('service) routes: HttpRoutes[F])(
-      implicit ec: ExecutionContext = ExecutionContext.global): HttpRoutes[F] =
+      redactHeadersWhen: CaseInsensitiveString => Boolean = Headers.SensitiveHeaders.contains,
+      logAction: String => Unit = logger.info(_)
+  )(@deprecatedName('service) http: Kleisli[F, Request[F], Response[F]])(
+      implicit ec: ExecutionContext = ExecutionContext.global,
```
jmcardon: can we remove the default arg implicit here? I don't like this.

rossabaker: I missed that, and I agree.

aworton:
default implicit in both request and response loggers smat/smote/smiten ;)

PR #1917, opened 2018-06-07
Implement optional implicit allowing usage of circe decoders without EntityDecoder definition

Leammas:
I've discussed the necessity of this implicit in gitter. Appeared useful.


Leammas:
The first point is already mentioned in https://github.com/http4s/http4s/pull/1917/files#diff-dcf0079785763cffc8cb901dad9f0628R205 Should I clarify it?

And I doubt that incoherent instances are possible in the same scope: they result in ambiguous implicits error.

rossabaker:
I looked at the diff, and not the wider context.  You're right, I think the first point is covered well enough.

As for the second issue, a lexically scoped instance (e.g., `import CirceEnityDecoder._`) has higher priority than an implicitly scoped instance (e.g., one found on the companion of `EntityDecoder`).  So the issue here is that `Ok("foo")` compiles with or without this new import, but the behavior varies.  It's fine as long as people know the ramifications, but it's a situation that I personally prefer to avoid.

Leammas:
I tried testing the implicit resolution and found that it does not compile at all. Somehow `EntityDecoder` from companion and generic one got the same priority.
Here's the repro: https://github.com/Leammas/http4s/blob/4f114c4ecccc3ab01a6b0491968056de527cdd5c/circe/src/test/scala/org/http4s/circe/CirceSpec.scala#L191

circe/src/main/scala/org/http4s/circe/CirceEntityDecoder.scala
```diff
@@ -0,0 +1,14 @@
+package org.http4s.circe
+
+import cats.effect.Sync
+import io.circe.Decoder
+import org.http4s.EntityDecoder
+
+/**
+  * Derive [[EntityDecoder]] if implicit [[Decoder]] is in the scope without need to explicitly call `jsonOf`
+  */
+trait CirceEntityDecoder {
```
ChristopherDavenport: Why do this as a trait and then an object which extends said trait? Why not just place this in the object?

circe/src/main/scala/org/http4s/circe/CirceEntityEncDec.scala
```diff
@@ -0,0 +1,3 @@
+package org.http4s.circe
+
+object CirceEntityEncDec extends CirceEntityDecoder with CirceEntityEncoder
```
rossabaker: I think `Codec` is a more common term than `EncDec`.  How about `CirceEntityCodec`?

docs/src/main/tut/json.md
```diff
@@ -206,6 +216,13 @@ However, be cautious when using this. Having this implicit
 in scope does mean that we would always try to decode HTTP entities
 from JSON (even if it is XML or plain text, for instance).
 
+For more convenience there is import combining both encoding 
+and decoding derivation: 
+
+```tut:book
+org.http4s.circe.CirceEntityEncDec._
```
rossabaker: This tut broke with an ambiguous implicit.  We need to sort this out.  Running `sbt doc/tutQuick` should test it locally.

circe/src/main/scala/org/http4s/circe/CirceEntityEncDec.scala
Leammas: Yeah, sounds better.

docs/src/main/tut/json.md
Leammas: Woops, the exact thing I wrote: we can't use both `CirceEntityEncoder` and plain json. I [restricted the scope](https://github.com/http4s/http4s/pull/1917/commits/4d73de742785a58cb5753f427e07e115ad99ba3f#diff-dcf0079785763cffc8cb901dad9f0628R168), but I'm not sure if it's the right way. Maybe I should not use `tut:book` there at all?

rossabaker: I think it's good to ensure that the import continues to compile, and you note the risk of ambiguity below.  I played with invisible braces around it, and it doesn't work.  I think the best we can do.

Issue #1920, opened 2018-06-14
PoolManager tries to dequeue from empty Queue 


tbrown1979:
Issue we were seeing on production:
```scala
java.util.NoSuchElementException: queue empty
    at scala.collection.mutable.Queue.dequeue(Queue.scala:66)
    at org.http4s.client.PoolManager$$anonfun$borrow$1.go$1(PoolManager.scala:173)
    at org.http4s.client.PoolManager$$anonfun$borrow$1.apply(PoolManager.scala:187)
    at org.http4s.client.PoolManager$$anonfun$borrow$1.apply(PoolManager.scala:144)
    at io.chrisdavenport.scalaz.task.instances.TaskInstances$$anon$1$$anonfun$async$1.apply(TaskInstances.scala:28)
    at io.chrisdavenport.scalaz.task.instances.TaskInstances$$anon$1$$anonfun$async$1.apply(TaskInstances.scala:26)
```

#### Scenario (a more thorough explanation is below this): 

> We `createConnection` 10 times for the _same_ `RequestKey` so we're at the `MaxTotalConnections` then 1 connection is `release`ed. 
> 
> We release that 1 connection and we have nothing in the `WaitQueue` so we add to the `IdleQueue` for the `RequestKey`. This does not decrement the `curTotal`, which is still `10` and equal to the `MaxTotal`.
> 
> We then `createConnection` again, with a _different_ `RequestKey`.
> 
> We'll match on the `case None if curTotal == maxTotal =>` case. 
> 
> Then we will evict a random value from the idle queue. 
> 
> This will dequeue from the only idle queue we have available. 
> 
> Then we'll try to `decrConnection`. This _won't_ remove the key from the `idleQueue` because we have more allocated connections with the same `RequestKey`. 
> 
> Then you `borrow` again and you can end up dequeuing from the queue that has nothing in it?

#### We can distill this it a bit:

[...]

PR #1922, opened 2018-06-14
Remove the key from idleQueue if it is empty after dequeue

tbrown1979:
Should fix https://github.com/http4s/http4s/issues/1920

### DIFF
--- a/client/src/main/scala/org/http4s/client/PoolManager.scala
+++ b/client/src/main/scala/org/http4s/client/PoolManager.scala
@@ -171,6 +171,7 @@ private final class PoolManager[F[_], A <: Connection[F]](
                     s"No connections available for the desired key. Evicting random and creating a new connection: $stats")
                   val randKey = keys.iterator.drop(Random.nextInt(keys.size)).next()
                   idleQueues(randKey).dequeue().shutdown()
+                  if (idleQueues.get(randKey).exists(_.isEmpty)) idleQueues.remove(randKey)
                   decrConnection(randKey)
                   createConnection(key, callback)
                 } else {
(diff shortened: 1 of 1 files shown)


SystemFw:
Not really asking for this specific PR, but a few more comments explaining the design of the Manager (don't have to be detailed, even a general design comment at the top) would go a long way given that the code is slightly hairier than average http4s :)

tbrown1979:
@rossabaker That's a great point. I can make that change. 

@SystemFw That would definitely be nice. Fwiw, once cats-effect 1.0 is released I can work on purifying the `PoolManager` which would make it a lot easier to reason about.

tbrown1979:
@rossabaker Let me know if that looks okay to you.

tbrown1979:
@rossabaker I look forward to seeing how I can improve things with the new cats-effect tools 😄

SystemFw:
@tbrown1979 You can start working on it right now probably, cats-effect 1.0 is not out yet but I don't foresee big changes from now until the release. Feel free to ping me should you need  a hand with the new concurrency stuff there.

@rossabaker Maybe that's a sufficient knock on the door for a milestone?

tbrown1979:
@ChristopherDavenport We still do remove an idle connection here. The [`getConnectionFromQueue`](https://github.com/http4s/http4s/pull/1922/files#diff-d2f4e3b3e2b0f3a001976f1e34fd7cb9R47) will `.dequeue` _iff_ the `Queue` for the given `key` is `.nonEmpty`, and then it returns us that connection. We'll then `.shutdown()` that connection. Let me know if that helps!

tbrown1979:
Should I make the log I added an `error` instead of `warn`? It's something that really should never happen, and if it does it might be good to have it louder than `warn`?

ChristopherDavenport:
Yet we have experience it does. So I am inclined towards the level it is at.

PR #1925, opened 2018-06-19
Test on oraclejdk10

rossabaker:
I've been watching that Travis ticket for over a month and I still don't know whether this works but they closed it so let's see.

### DIFF
--- a/website/src/hugo/content/changelog.md
+++ b/website/src/hugo/content/changelog.md
@@ -9,6 +9,7 @@ ordered chronologically, so each release contains all changes described below
 it.
 
 # v0.18.13 (unreleased)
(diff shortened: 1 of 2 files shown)


website/src/hugo/content/changelog.md
```diff
@@ -9,6 +9,7 @@ ordered chronologically, so each release contains all changes described below
 it.
 
 # v0.18.13 (unreleased)
+* Downcase type in `MediaRange` generator [#1907](https://github.com/http4s/http4s/pull/1907)
```
ChristopherDavenport: Whats with this guy?

rossabaker: I had that locally and apparently hadn't pushed it yet.

PR #1930, opened 2018-06-26
Backport #1917 (CirceEntityCodec) to 0.18.x

rossabaker:


rossabaker:
There are caveats in the docs.  I don't personally like it, but at least now we have what most people seem to be reimplementing anyway.

Open to proposals on ways to make this unnecessary for 1.0.

PR #1934, opened 2018-07-03
[PoolManager] - Synchronized -> Semaphore, wrap side-effects in Effect[F].delay

tbrown1979:
This PR is an attempt to change the use of `synchronized { ... }` to using a `Semaphore`. To make this easier I wrapped all the `Unit` functions in `F` so that `.acquire`ing a permit was easier. I opted to `.acquire` and `release` the permits instead of doing `.withPermit` just so we could be explicit in all the cases during the `.borrow` method. Doing so allowed me to push the async handling down to the specific cases that require it, instead of every case hitting the callback.

This is the first step towards purifying the `PoolManager` and hopefully making it a little more easy to reason about. I have plans for further changes after this, such as removing the `var`s. 

Thoughts are appreciated 😄If you'd like to see this done a different way I'm all ears!


blaze-client/src/main/scala/org/http4s/client/blaze/PooledHttp1Client.scala
```diff
@@ -18,10 +18,12 @@ object PooledHttp1Client {
...
-        maxConnectionsPerRequestKey = maxConnectionsPerRequestKey
-      ))
+    Effect[F].toIO {
+      Http1Client.mkClient(
+        config.copy(
+          maxTotalConnections = maxTotalConnections,
+          maxWaitQueueLimit = maxWaitQueueLimit,
+          maxConnectionsPerRequestKey = maxConnectionsPerRequestKey
+        ))
+    }.unsafeRunSync()
```
jmcardon: If we're going to force `toIO`, might as well make the whole call return `F[Client[F]]` instead. There's no need to maintain source compat on a deprecated method since M7 of 0.18.

client/src/main/scala/org/http4s/client/PoolManager.scala
```diff
@@ -317,11 +326,9 @@ private final class PoolManager[F[_], A <: Connection[F]](
     */
   private def disposeConnection(key: RequestKey, connection: Option[A]): Unit = {
     logger.debug(s"Disposing of connection: $stats")
-    synchronized {
-      decrConnection(key)
-      connection.foreach { s =>
-        if (!s.isClosed) s.shutdown()
-      }
+    decrConnection(key)
```
jmcardon: This doesn't look right. disposeConnection now lack synchronization entirely.

blaze-client/src/main/scala/org/http4s/client/blaze/PooledHttp1Client.scala
tbrown1979: @jmcardon Is it okay to just remove the `PooledHttp1Client` then? I only did this so that we could keep it around, but figured I'd get questions about it.

client/src/main/scala/org/http4s/client/PoolManager.scala
tbrown1979: @jmcardon Good catch. I will get this fixed!

tbrown1979: @jmcardon This should be fixed. Let me know if you have any suggestions about the changes I made.

client/src/main/scala/org/http4s/client/PoolManager.scala
```diff
@@ -2,11 +2,14 @@ package org.http4s
...
+import cats.effect.concurrent.Semaphore
 import java.time.Instant
 import java.util.concurrent.TimeoutException
 import org.log4s.getLogger
-import org.http4s.internal.unsafeRunAsync
-import scala.annotation.tailrec
+import cats.implicits._
+//import cats.syntax.all._
+
+//import scala.annotation.tailrec
```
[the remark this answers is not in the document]
tbrown1979: Will do!

blaze-client/src/main/scala/org/http4s/client/blaze/PooledHttp1Client.scala
```diff
@@ -18,10 +18,14 @@ object PooledHttp1Client {
...
+    Effect[F]
+      .toIO {
+        Http1Client.mkClient(
+          config.copy(
+            maxTotalConnections = maxTotalConnections,
+            maxWaitQueueLimit = maxWaitQueueLimit,
+            maxConnectionsPerRequestKey = maxConnectionsPerRequestKey
+          ))
+      }
+      .unsafeRunSync()
```
jmcardon: Honestly I am not ok with this `unsafeRunSync` call if it's litearally the same thing as http1client.

Can we remove this shit altogether? @rossabaker @ChristopherDavenport

client/src/main/scala/org/http4s/client/PoolManager.scala
```diff
@@ -90,29 +88,33 @@ private final class PoolManager[F[_], A <: Connection[F]](
...
-      unsafeRunAsync(builder(key)) {
-        case Right(conn) =>
-          IO(callback(Right(NextConnection(conn, fresh = true))))
-        case Left(error) =>
-          disposeConnection(key, None)
-          IO(callback(Left(error)))
-      }
+      incrConnection(key) *>
+        F.liftIO {
+          F.runAsync(Async.shift(executionContext) *> builder(key)) {
```
jmcardon: This doesn't seem necessary to do at all considering we're in `F[_]` now.

This should be done without runAsync and `liftIO` imo.

blaze-client/src/main/scala/org/http4s/client/blaze/PooledHttp1Client.scala
ChristopherDavenport: I mean I'm a fan of enforcing safety by construction. Any reason we should leave this?

jmcardon: https://gitter.im/http4s/http4s?at=5b3d61cef1664406611a5117

Hell yeah. Time to set this on fire.

tbrown1979: 🔥 killed!

blaze-core/src/main/scala/org/http4s/blazecore/util/ChunkWriter.scala
```diff
@@ -41,11 +40,13 @@ private[util] object ChunkWriter {
         ByteBuffer.wrap(rr.result.getBytes(ISO_8859_1))
       } else ChunkEndBuffer
     }
-    unsafeRunAsync(f) {
-      case Right(buffer) =>
-        IO { promise.completeWith(pipe.channelWrite(buffer).map(Function.const(false))); () }
-      case Left(t) =>
-        IO { promise.failure(t); () }
+    F.start {
```
jmcardon: This never runs the action in start, LOL. 

Do you mind reverting this? This isn't a necessary change and I'd rather not touch blaze core here.

edmundnoble: You need to run the outer `F` to kick it off when you use `start`; as it is, this code does nothing at all anymore.

client/src/main/scala/org/http4s/client/PoolManager.scala
```diff
@@ -90,29 +88,32 @@ private final class PoolManager[F[_], A <: Connection[F]](
...
     if (numConnectionsCheckHolds(key)) {
-      incrConnection(key)
-      unsafeRunAsync(builder(key)) {
-        case Right(conn) =>
-          IO(callback(Right(NextConnection(conn, fresh = true))))
-        case Left(error) =>
-          disposeConnection(key, None)
-          IO(callback(Left(error)))
-      }
+      F.start {
```
jmcardon: incrConnection(key) *> F.start

blaze-core/src/test/scala/org/http4s/blazecore/util/FailingWriter.scala
```diff
@@ -7,7 +7,7 @@ import fs2._
 import org.http4s.blaze.pipeline.Command.EOF
 import scala.concurrent.{ExecutionContext, Future}
 
-class FailingWriter(implicit protected val F: Effect[IO]) extends EntityBodyWriter[IO] {
+class FailingWriter(implicit protected val F: Concurrent[IO]) extends EntityBodyWriter[IO] {
```
jmcardon: Wait why???

tbrown1979:
I've been on vacation for the last few weeks. I'll get back around to this soon 😄

tbrown1979:
@rossabaker I think the last thing I had left to do was to convert `MimeLoaderApp` to an `IOApp`. Again, apologies that I haven't been able to get around to that. I will pick this up this week and get that done 😄

client/src/main/scala/org/http4s/client/PoolManager.scala
```diff
@@ -139,81 +139,83 @@ private final class PoolManager[F[_], A <: Connection[F]](
...
+                  F.asyncF(addToWaitQueue(key, _) *> semaphore.release)
+              }
+
+            case None => // we're full up. Add to waiting queue.
+              F.delay(logger.debug(s"No connections available.  Waiting on new connection: $stats")) *>
+                F.asyncF(addToWaitQueue(key, _) *> semaphore.release)
+          }
+
+      F.delay(logger.debug(s"Requesting connection: $stats")) *>
+        go()
```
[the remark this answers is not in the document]
tbrown1979: @rossabaker I originally did this, but I found I couldn't match the previous behavior with that. If we use the `semaphore.withPermit` here then we'll be waiting for the `F.asyncF` calls to finish, which may take a while - and this would be a change in behavior from what the `PoolManager` did before this PR. 

[This](https://github.com/http4s/http4s/blob/master/client/src/main/scala/org/http4s/client/PoolManager.scala#L160) executes async, which immediately finishes the `synchronized` block of code. 

I think this probably could be done with `F.asyncF` surrounding the whole thing and then hitting the callback in each `case`. However, I thought that the explicit `.release` we're doing here allows a little more clarity and preciseness than the alternative.

Please let me know what your thoughts are on that. Let me know if I'm missing something 😄

client/src/main/scala/org/http4s/client/PoolManager.scala
```diff
@@ -139,81 +139,83 @@ private final class PoolManager[F[_], A <: Connection[F]](
...
-              case Some(conn) if !conn.isClosed =>
-                logger.debug(s"Recycling connection: $stats")
-                callback(Right(NextConnection(conn, fresh = false)))
-
-              case Some(closedConn @ _) =>
-                logger.debug(s"Evicting closed connection: $stats")
-                decrConnection(key)
+    if (!isClosed) {
+      def go(): F[NextConnection] =
+        semaphore.acquire *>
```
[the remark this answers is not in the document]
tbrown1979: This does seem like it could be an issue. I'll wait for a response on the follow-up comment just to make sure I do this the correct way. Good catch!

client/src/main/scala/org/http4s/client/PoolManager.scala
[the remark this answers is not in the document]
SystemFw: if the equivalent `getConnection` is also async (with `start`), shouldn't be enough to not have `asyncF` in that case, and use `withPermit`?

PR #1936, opened 2018-07-04
Release v0.19.0-M1

rossabaker:


client/src/main/scala/org/http4s/client/JavaNetClient.scala
```diff
@@ -0,0 +1,186 @@
...
+  def withBlockingExecutionContext(blockingExecutionContext: ExecutionContext): JavaNetClient =
+    copy(blockingExecutionContext = blockingExecutionContext)
+
+  /** Creates the `JavaNetClient`.
+    *
+    * The shutdown of this client is a no-op.  Creation of the client
+    * allocates no resources, and any resources allocated while using
+    * this client are reclaimed by the JVM at its own leisure.
+    */
+  def create[F[_]](implicit F: Async[F], timer: Timer[F]): Client[F] = Client(open, F.unit)
```
rossabaker: Not sure if taking the timer here is the right thing.  In 0.18, the EC is taken on the class itself, but that would require making the whole builder polymorphic.

core/src/main/scala/org/http4s/internal/package.scala
```diff
@@ -10,4 +10,14 @@ package object internal {
...
+
+  private[http4s] def blocking[F[_], A](fa: F[A], blockingExecutionContext: ExecutionContext)(
+      implicit F: Async[F],
+      timer: Timer[F]): F[A] =
+    for {
+      _ <- Async.shift[F](blockingExecutionContext)
+      att <- fa.attempt
+      _ <- timer.shift
+      fa0 <- F.fromEither(att)
+    } yield fa0
```
rossabaker: A little different from the impl in 0.18.x.  Should look a lot like the linebacker for the corresponding version. /cc @ChristopherDavenport

client/src/main/scala/org/http4s/client/JavaNetClient.scala
```diff
@@ -0,0 +1,186 @@
...
+  * exploration in a REPL.
+  *
+  * All I/O operations in this client are blocking.
+  *
+  * @param blockingExecutionContext An `ExecutionContext` on which
+  * blocking operations will be performed.
+  * @param ec The `ExecutionContext` to which work will be shifted
+  * back after blocking.
+  */
+sealed abstract class JavaNetClient private (
```
cquiroz: This might be useful for the mythical `scala.js` client

client/src/main/scala/org/http4s/client/JavaNetClient.scala
cquiroz: `Timer` comes from `cats 1.x` right?

core/src/main/scala/org/http4s/util/threads.scala
```diff
@@ -80,4 +80,13 @@ object threads {
       timeout: Boolean = false): ExecutionContext =
     ExecutionContext.fromExecutorService(newDaemonPool(name, min, cpuFactor, timeout))
 
+  private[http4s] def newBlockingPool(name: String): ExecutorService =
+    new ThreadPoolExecutor(
+      0,
```
cquiroz: Maybe we could constants here to make it clearer what the `0` and `60` refer to

core/src/main/scala/org/http4s/internal/package.scala
ChristopherDavenport: Why is this not using bracket? Take a look at the current impl in linebacker.

PR #1939, opened 2018-07-05
Add StaticHeaders middleware

ChristopherDavenport:
Adds Static Headers to a request. Puts in place a `no-cache` middleware based on this design.

PR #1941, opened 2018-07-05
Merge 0.18.15

rossabaker:
#1939, made generic to fit the 0.19 ethos.

### DIFF
--- a/server/src/main/scala/org/http4s/server/middleware/StaticHeaders.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/StaticHeaders.scala
@@ -1,0 +1,25 @@
+package org.http4s.server.middleware
+
+import cats.Functor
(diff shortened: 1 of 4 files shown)


server/src/main/scala/org/http4s/server/middleware/StaticHeaders.scala
```diff
@@ -0,0 +1,25 @@
...
+import cats.implicits._
+import org.http4s.{CacheDirective, Header, Headers, Response}
+import org.http4s.headers.`Cache-Control`
+
+/**
+  * Simple middleware for adding a static set of headers to responses
+  * returned by a kleisli.
+  */
+object StaticHeaders {
+  def apply[F[_]: Functor, G[_]: Functor, A](headers: Headers)(
```
ChristopherDavenport: Functor on `G` is unnecessary.

PR #1943, opened 2018-07-06
Make the imports in the Client documentation silent

jcranky:


ChristopherDavenport:
I love this but I always forget to do it.

cquiroz:
What `tut:silent` does?

ChristopherDavenport:
Makes it look generally like a block of `scala` code in markdown. But actually effect the running tut file. Very useful specifically to reduce the noise of imports.

I always forget which is hidden, probably `hidden` and `silent` which suppresses the tut repl output.

PR #1945, opened 2018-07-09
Add UUID path extractor

tmfset:
Add UUIDVar, a path extractor for java.util.UUID

I also considered adding a `ValidatedUUIDVar` to extract a `Validated[Throwable, UUID]` from paths but then realized that there is already a simple way to return `BadRequest` for invalid UUIDs:

```
HttpService[IO] match {
  case GET -> Root / "user" / UUIDVar(userId) => ...
  case GET -> Root / "user" / str => BadRequest(s"Invalid identifier '$str'")
}
```

### DIFF
--- a/dsl/src/main/scala/org/http4s/dsl/Http4sDsl.scala
+++ b/dsl/src/main/scala/org/http4s/dsl/Http4sDsl.scala
@@ -22,6 +22,7 @@ trait Http4sDsl[F[_]] extends Http4s with Methods with Statuses with Responses[F
 
   val IntVar: impl.IntVar.type = impl.IntVar
   val LongVar: impl.LongVar.type = impl.LongVar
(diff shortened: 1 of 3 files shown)


tmfset:
I discussed this change with @ChristopherDavenport @rossabaker in gitter, tried to keep it simple :)

tmfset:
Binary compatibility check failed because I removed `NumericPathVar`. If necessary I can fix this change to avoid modifying that class.

tmfset:
Ha well the compatibility check still fails because I added a new class. Maybe it was inappropriate to target this for 0.18.x. @ChristopherDavenport should I target master instead?

tmfset:
@ChristopherDavenport Abandoning this PR, I will target master instead.

PR #1946, opened 2018-07-11
add play json module

julien-truffaut:


julien-truffaut:
How can I disable mima for this new module? 
```
[error] (play-json / mimaPreviousClassfiles) sbt.librarymanagement.ResolveException: unresolved dependency: org.http4s#http4s-play-json_2.12;0.18.15: not found
```

julien-truffaut:
I fixed mima and other formatting issue. However, I don't know why the last build is hanging: 

https://travis-ci.org/http4s/http4s/jobs/403105677

julien-truffaut:
I may need some help to investigate the build failure, I don't really know where to look atm

julien-truffaut:
@rossabaker probably the problem is that I can't seem to run the test locally, I tried both: `sbt ci` and `sbt test` in `2.12` and `2.11`  but I always get:

```
[error] (okhttp-client / Test / test) sbt.TestsFailedException: Tests unsuccessful
[error] (client / Test / test) sbt.TestsFailedException: Tests unsuccessful
```

should I run another command?

rossabaker:
I'm not sure what client tests you saw fail, but I used to get client failures on OSX with regard to resolving localhost.  I solved these by moving back to Linux. :penguin:

I saw a linker error when running `play/test`.  Poking around at the jawn build, it looks like that was built against play-json-3.5 for Scala 2.11 on the version of jawn we use.  Lining that up works for me locally.  If we upgrade jawn in 0.19, it looks like we can get on the latest play-json for both Scala versions.

julien-truffaut:
Thank Ross for having a look.

FYI, I am getting test failures for JavaNetClient and OkHttpClient:
```
[info] JavaNetClientSpec
[error] ! Execute GET: /simple (166 ms)
[error]  java.io.IOException: Server returned HTTP response code: 502 for URL: http://julien-HP-Z240-SFF-Workstation:37849/simple (null:-1)
```
It might be related to my work network. I will try from home.

Regarding you config change, it seems that sbt doesn't accept `playJsonVersion` being a function:
```
[error] /home/travis/build/http4s/http4s/project/Http4sPlugin.scala:320:105: missing argument list for method playJsonVersion in object Http4sPlugin
[error] Unapplied methods are only converted to functions when a function type is expected.
[error] You can make this conversion explicit by writing `playJsonVersion _` or `playJsonVersion(_)` instead of `playJsonVersion`.
[error]   def playJson(sbv: String)                 = "com.typesafe.play"      %% "play-json"                 % playJsonVersion
```
I get the same error locally.

rossabaker:
Ugh.  I had it passing, and got a couple commits tangled, and untangled them, and pushed the wrong state.  Sorry about that.  Looking again.

play-json/src/main/scala/org/http4s/play/PlayEntityDecoder.scala
```diff
@@ -0,0 +1,14 @@
...
+
+import cats.effect.Sync
+import org.http4s.EntityDecoder
+import play.api.libs.json.Reads
+
+/**
+  * Derive [[EntityDecoder]] if implicit [[Reads]] is in the scope without need to explicitly call `jsonOf`
+  */
+trait PlayEntityDecoder {
+  implicit def circeEntityDecoder[F[_]: Sync, A: Reads]: EntityDecoder[F, A] = jsonOf[F, A]
```
rossabaker: Should not call this `circe`.

play-json/src/main/scala/org/http4s/play/PlayEntityEncoder.scala
```diff
@@ -0,0 +1,15 @@
...
+
+import cats.Applicative
+import play.api.libs.json.Writes
+import org.http4s.EntityEncoder
+
+/**
+  * Derive [[EntityEncoder]] if implicit [[Writes]] is in the scope without need to explicitly call `jsonEncoderOf`
+  */
+trait PlayEntityEncoder {
+  implicit def circeEntityEncoder[F[_]: Applicative, A: Writes]: EntityEncoder[F, A] =
```
rossabaker: Should not call this `circe`.

julien-truffaut: thanks I will fix this

PR #1955, opened 2018-07-18
WIP: Basic Header Typeclass Sketch

ChristopherDavenport:
To deal with reflection of header registry.

### DIFF
--- a/core/src/main/scala/org/http4s/Header.scala
+++ b/core/src/main/scala/org/http4s/Header.scala
@@ -18,118 +18,134 @@
  */
 package org.http4s
 
-import cats._
-import cats.data.NonEmptyList
-import cats.implicits.{catsSyntaxEither => _, _}
-import org.http4s.syntax.string._
+// import cats._
+// import cats.data.NonEmptyList
+// import cats.implicits.{catsSyntaxEither => _, _}
+// import org.http4s.syntax.string._
 import org.http4s.util._
-import scala.util.hashing.MurmurHash3
+// import scala.util.hashing.MurmurHash3
 
-/**
-  * Abstract representation o the HTTP header
-  * @see org.http4s.HeaderKey
-  */
-sealed trait Header extends Renderable with Product {
-  import Header.Raw
-
+trait Header[A] {
   def name: CaseInsensitiveString
-
-  def parsed: Header
-
-  def renderValue(writer: Writer): writer.type
-
-  def value: String = {
-    val w = new StringWriter
-    renderValue(w).result
-  }
-
-  def is(key: HeaderKey): Boolean = key.matchHeader(this).isDefined
-
-  def isNot(key: HeaderKey): Boolean = !is(key)
-
-  override def toString: String = name + ": " + value
-
-  def toRaw: Raw = Raw(name, value)
-
-  final def render(writer: Writer): writer.type = {
-    writer << name << ':' << ' '
-    renderValue(writer)
-  }
-
-  final override def hashCode(): Int =
-    MurmurHash3.mixLast(name.hashCode, MurmurHash3.productHash(parsed))
-
-  final override def equals(that: Any): Boolean = that match {
-    case h: AnyRef if this eq h => true
-    case h: Header =>
-      (name == h.name) &&
-        (parsed.productArity == h.parsed.productArity) &&
-        (parsed.productIterator.sameElements(h.parsed.productIterator))
-    case _ => false
-  }
+  def parse(s: String): Option[A]
+    // some headers like Set-Cookie may emit multiple actual headers
+  def toRawHeaders(a: A): List[Header.RawHeader]
+  def isSingleton: Boolean
 }
 
 object Header {
-  def unapply(header: Header): Option[(CaseInsensitiveString, String)] =
-    Some((header.name, header.value))
+  def apply[A](implicit A: Header[A]): Header[A] = A
 
-  def apply(name: String, value: String): Raw = Raw(name.ci, value)
+  final case class RawHeader(name: CaseInsensitiveString, value: String)
 
-  /**
-    * Raw representation of the Header
-    *
-    * This can be considered the simplest representation where the header is specified as the product of
-    * a key and a value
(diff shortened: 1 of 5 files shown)


core/src/main/scala/org/http4s/Headers.scala
```diff
@@ -1,158 +1,199 @@
...
-    */
-  def removePayloadHeaders: Headers =
-    filterNot(h => Headers.PayloadHeaderKeys(h.name))
-
-  def redactSensitive(
-      redactWhen: CaseInsensitiveString => Boolean = Headers.SensitiveHeaders.contains): Headers =
-    headers.map {
-      case h if redactWhen(h.name) => Header.Raw(h.name, "<REDACTED>")
-      case h => h
+  def getRepeatable(key: CaseInsensitiveString): Option[Header.RawHeader] = {
```
rossabaker: Why wouldn't this return a `List`?  Then it would be safe to call with `"Set-Cookie".ci `?

core/src/main/scala/org/http4s/Header.scala
```diff
@@ -18,118 +18,134 @@
...
-  */
-sealed trait Header extends Renderable with Product {
-  import Header.Raw
+// import scala.util.hashing.MurmurHash3
 
+trait Header[A] {
   def name: CaseInsensitiveString
+  def parse(s: String): Option[A]
+    // some headers like Set-Cookie may emit multiple actual headers
+  def toRawHeaders(a: A): List[Header.RawHeader]
```
rossabaker: This flattening is to support `Headers#put`?  What does this do on a comma-separated header for headers that play by the rules?

core/src/main/scala/org/http4s/Sigma.scala
```diff
@@ -0,0 +1,15 @@
+package org.http4s
```
rossabaker: Maybe into util, the "things that aren't HTTP and that we haven't bothered to make into a separate library" package of shame?

core/src/main/scala/org/http4s/headers/Expires.scala
```diff
@@ -18,8 +13,14 @@ object Expires extends HeaderKey.Internal[Expires] with HeaderKey.Singleton {
...
   * @param expirationDate the date of expiration
   */
-final case class Expires(expirationDate: HttpDate) extends Header.Parsed {
-  val key = `Expires`
-  override val value = Renderer.renderString(expirationDate)
-  override def renderValue(writer: Writer): writer.type = writer.append(value)
-}
+final case class Expires(expirationDate: java.time.Instant)
+object Expires {
+  implicit val expiresHeader: Header[Expires] =
```
rossabaker: I was thinking about going full cats-convention on the arbitraries.  Maybe we should start that here?

core/src/main/scala/org/http4s/headers/Expires.scala
```diff
@@ -18,8 +13,14 @@ object Expires extends HeaderKey.Internal[Expires] with HeaderKey.Singleton {
...
+      new Header[Expires] {
+        val name = CaseInsensitiveString("Expires")
+        def parse(s: String) = HttpHeaderParser.EXPIRES(s).toOption
+        def toRawHeaders(a: Expires) = List(RawHeader(
+          name, Renderer.renderString(a.expirationDate)))
```
rossabaker: Is this a good opportunity to phase out `Renderer`?

core/src/main/scala/org/http4s/headers/Expires.scala
```diff
@@ -18,8 +13,14 @@ object Expires extends HeaderKey.Internal[Expires] with HeaderKey.Singleton {
...
-final case class Expires(expirationDate: HttpDate) extends Header.Parsed {
-  val key = `Expires`
-  override val value = Renderer.renderString(expirationDate)
-  override def renderValue(writer: Writer): writer.type = writer.append(value)
-}
+final case class Expires(expirationDate: java.time.Instant)
+object Expires {
+  implicit val expiresHeader: Header[Expires] =
+      new Header[Expires] {
+        val name = CaseInsensitiveString("Expires")
```
rossabaker: I'm not entirely sure that `CaseInsensitiveString` has been a successful experiment, and we're about to double down on it.  What do other people think?

core/src/main/scala/org/http4s/Headers.scala
```diff
@@ -1,158 +1,199 @@
...
 
-  /** Create a new Headers collection from the headers */
-  def apply(headers: List[Header]): Headers = new Headers(headers)
-
-  implicit val canBuildFrom: CanBuildFrom[Traversable[Header], Header, Headers] =
-    new CanBuildFrom[TraversableOnce[Header], Header, Headers] {
-      def apply(from: TraversableOnce[Header]): mutable.Builder[Header, Headers] = newBuilder
-
-      def apply(): mutable.Builder[Header, Headers] = newBuilder
+    def put(headers0: Sigma[Header]*): Headers = {
```
rossabaker: This is an interesting technique that would make an interesting Typelevel blog post if anyone is so motivated.

core/src/main/scala/org/http4s/headers/Expires.scala
ChristopherDavenport: Probably, same with the writer implementations.

core/src/main/scala/org/http4s/headers/Expires.scala
ChristopherDavenport: As in put them in their own package so I don't need to write them in every spots companion?

core/src/main/scala/org/http4s/Sigma.scala
ChristopherDavenport: Very reasonable.

core/src/main/scala/org/http4s/Headers.scala
ChristopherDavenport: Hmmm. Yeah.

m-sp:
this was a long time coming but finally landed! but I think we can close this one now?

PR #1961, opened 2018-07-23
Prometheus client metrics

zcox:
Client middleware that collects a few Prometheus metrics, very similar to the `prometheus-server-metrics` module. Hopefully this will be useful to the http4s community. This is my first http4s PR, feedback would be greatly appreciated.

Currently, four client metrics are collected:
- Counter for total number of responses received
- Histogram for response duration/latency
- Gauge for number of active requests
- Counter for total number of client errors

At Banno, we often use the same `Client` instance to send requests to multiple destinations (e.g. endpoints, services) and want to track metrics separately based on destination, so several of the metrics use a `destination` label. This PR uses the value of the request attribute `PrometheusClientMetrics.Destination` as the `destination` label value, if it is set. User code can set this attribute to some meaningful value when creating the `Request`. Otherwise, a fallback value can be specified when creating the middleware. This may be a bit specific to our own use cases, but hopefully is a common need for client metric collection.

A few things I'm a bit unsure about:
- Is the `destination` label sufficiently generic and useful to many users of http4s-client?
- Is the `destination` label stuff a misuse of the attribute system? Is there a better way to track metrics separately per-destination?
- Should `onClientError` also update the response counter and histogram?
- Other client metrics that should be collected?

### DIFF
--- a/prometheus-client-metrics/src/main/scala/org/http4s/client/prometheus/PrometheusClientMetrics.scala
+++ b/prometheus-client-metrics/src/main/scala/org/http4s/client/prometheus/PrometheusClientMetrics.scala
@@ -28,7 +28,8 @@ object PrometheusClientMetrics {
       destination: String,
       responseDuration: Histogram,
       activeRequests: Gauge,
-      responseCounter: Counter
+      responseCounter: Counter,
+      clientErrorsCounter: Counter
   )
 
   private def metricsClient[F[_]: Sync](
@@ -47,19 +48,22 @@ object PrometheusClientMetrics {
       end <- Sync[F].delay(System.nanoTime())
       result <- responseAttempt.fold(
         e =>
-          onClientError(request, e, start, end, metrics) *>
+          onClientError(request, metrics) *>
             Sync[F].raiseError[DisposableResponse[F]](e),
         r => onResponse(request, r.response, start, end, metrics) *> r.pure[F]
       )
     } yield result
 
-  private def onClientError[F[_]: Sync](
-      request: Request[F],
-      e: Throwable,
-      start: Long,
-      end: Long,
-      metrics: ClientMetrics): F[Unit] =
-    ??? //TODO when would this happen? how should we update metrics?
+  private def onClientError[F[_]: Sync](request: Request[F], metrics: ClientMetrics): F[Unit] =
+    Sync[F].delay {
+      //not updating responseDuration or responseCounter, since we did not receive a response
+      metrics.activeRequests
+        .labels(reportDestination(request.attributes, metrics.destination))
+        .dec()
+      metrics.clientErrorsCounter
+        .labels(reportDestination(request.attributes, metrics.destination))
+        .inc()
+    }
 
   private def onResponse[F[_]: Sync](
       request: Request[F],
@@ -109,6 +113,12 @@ object PrometheusClientMetrics {
             .name(prefix + "_" + "response_total")
             .help("Total Responses.")
             .labelNames("destination", "code")
+            .register(c),
+          clientErrorsCounter = Counter
+            .build()
+            .name(prefix + "_" + "client_errors_total")
+            .help("Total Client Errors.")
+            .labelNames("destination")
             .register(c)
         )
         client.copy(open = Kleisli(metricsClient[F](clientMetrics, client)(_)))
--- a/prometheus-client-metrics/src/main/scala/org/http4s/client/prometheus/PrometheusClientMetrics.scala
+++ b/prometheus-client-metrics/src/main/scala/org/http4s/client/prometheus/PrometheusClientMetrics.scala
@@ -18,14 +18,8 @@ object PrometheusClientMetrics {
       case _ => "5xx"
     }
 
-  /** The value of this key in the request's attributes is used as the value for the destination metric label. */
-  val Destination = AttributeKey[String]
-  val EmptyDestination = ""
-  private def reportDestination(attributes: AttributeMap, destination: String): String =
-    attributes.get(Destination).getOrElse(destination)
-
-  private case class ClientMetrics(
-      destination: String,
+  private case class ClientMetrics[F[_]](
+      destination: Request[F] => String,
       responseDuration: Histogram,
       activeRequests: Gauge,
       responseCounter: Counter,
@@ -33,7 +27,7 @@ object PrometheusClientMetrics {
   )
 
   private def metricsClient[F[_]: Sync](
-      metrics: ClientMetrics,
+      metrics: ClientMetrics[F],
       client: Client[F]
   )(
       request: Request[F]
@@ -42,7 +36,7 @@ object PrometheusClientMetrics {
       start <- Sync[F].delay(System.nanoTime())
       _ <- Sync[F].delay(
         metrics.activeRequests
-          .labels(reportDestination(request.attributes, metrics.destination))
+          .labels(metrics.destination(request))
           .inc())
       responseAttempt <- client.open(request).attempt
       end <- Sync[F].delay(System.nanoTime())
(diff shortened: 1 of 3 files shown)


prometheus-client-metrics/src/main/scala/org/http4s/client/prometheus/PrometheusClientMetrics.scala
```diff
@@ -0,0 +1,131 @@
...
+    status.code match {
+      case hundreds if hundreds < 200 => "1xx"
+      case twohundreds if twohundreds < 300 => "2xx"
+      case threehundreds if threehundreds < 400 => "3xx"
+      case fourhundreds if fourhundreds < 500 => "4xx"
+      case _ => "5xx"
+    }
+
+  /** The value of this key in the request's attributes is used as the value for the destination metric label. */
+  val Destination = AttributeKey[String]
```
rossabaker: I sometimes want to phase out attributes, but I can't get rid of all the use cases for them.  And here we have another.

Alternative: maybe a function `Request => Option[String]`  could be provided, so the destination could be calculated from the request (probably the `Uri`?)  itself.

prometheus-client-metrics/src/main/scala/org/http4s/client/prometheus/PrometheusClientMetrics.scala
```diff
@@ -0,0 +1,131 @@
...
+  )
+
+  private def metricsClient[F[_]: Sync](
+      metrics: ClientMetrics,
+      client: Client[F]
+  )(
+      request: Request[F]
+  ): F[DisposableResponse[F]] =
+    for {
+      start <- Sync[F].delay(System.nanoTime())
```
rossabaker: Would the new cats-effect `Timer` be a better fit here?

prometheus-client-metrics/src/main/scala/org/http4s/client/prometheus/PrometheusClientMetrics.scala
```diff
@@ -0,0 +1,131 @@
...
+      _ <- Sync[F].delay(
+        metrics.activeRequests
+          .labels(reportDestination(request.attributes, metrics.destination))
+          .inc())
+      responseAttempt <- client.open(request).attempt
+      end <- Sync[F].delay(System.nanoTime())
```
rossabaker: This is going to measure time-to-first-byte, and not time to process the body.  Both can be interesting, though the latter may be greatly affected by how aggressively the caller consumes the stream.  If you want to add the latter, it would be a finalizer on the response body.

prometheus-client-metrics/src/main/scala/org/http4s/client/prometheus/PrometheusClientMetrics.scala
```diff
@@ -0,0 +1,131 @@
...
+        e =>
+          onClientError(request, metrics) *>
+            Sync[F].raiseError[DisposableResponse[F]](e),
+        r => onResponse(request, r.response, start, end, metrics) *> r.pure[F]
+      )
+    } yield result
+
+  private def onClientError[F[_]: Sync](request: Request[F], metrics: ClientMetrics): F[Unit] =
+    Sync[F].delay {
+      //not updating responseDuration or responseCounter, since we did not receive a response
```
rossabaker: Maybe add an error counter?

rossabaker:
Paging Mr. Prometheus, @ChristopherDavenport.

prometheus-client-metrics/src/main/scala/org/http4s/client/prometheus/PrometheusClientMetrics.scala
zcox: I like the idea of a generic `Request => Option[String]`, will try that out. Maybe even `Request => List[String]` and let user specify a `List[String]` of label names when creating the middleware?

prometheus-client-metrics/src/main/scala/org/http4s/client/prometheus/PrometheusClientMetrics.scala
zcox: `clientErrorsCounter` is incremented a few lines below, would that be enough?

zcox:
99cc76f introduces a `Request[F] => String` function to obtain the `destination` label value, and includes the `DestinationAttribute` module with utils to store destination in request attributes.

The `destination` function could be as simple as `_.uri.host.map(_.value).getOrElse("")`.

Usage with `DestinationAttribute` would look something like this:

```scala
//somewhere things get intialized
for {
  ...
  cr = CollectorRegistry.defaultRegistry //or create your own
  client <- Http1Client[F]().flatMap(c => PrometheusClientMetrics(cr, destination = DestinationAttribute.getDestination()).run(c))
  ...
  x = ThingThatUsesClient1(DestinationAttribute.setRequestDestination(client, "d1"), )
  y = ThingThatUsesClient2(DestinationAttribute.setRequestDestination(client, "d2"), )
  z = ThingThatUsesClient3(client)
  ...
}

//ThingThatUsesClient3
for {
  req <- createRequest1(...).withAttribute(DestinationAttribute.Destination, "d3")
  resp <- client.fetch(req)
}

for {
  req <- createRequest2(...).withAttribute(DestinationAttribute.Destination, "d4")
  resp <- client.fetch(req)
}
```

prometheus-client-metrics/src/main/scala/org/http4s/client/prometheus/PrometheusClientMetrics.scala
```diff
@@ -0,0 +1,138 @@
...
+            .help("Total Client Errors.")
+            .labelNames("destination")
+            .register(c)
+        )
+        client.copy(open = Kleisli(metricsClient[F](clientMetrics, client)(_)))
+      }
+    }
+}
+
+object DestinationAttribute {
```
ChristopherDavenport: Could we get this moved to its own file please?

zcox:
I agree the attribute stuff is a bit weird/clunky, I would be totally OK with not including that in http4s (and maybe just using it internally). 

The one thing I like about it is that it puts the destination labelling close to the point of request creation. I could see the `Request[F] => String` turning into a big pattern match on hosts/paths/etc, but maybe I just need to try using that instead of attributes in our use case and see how it looks.

prometheus-client-metrics/src/main/scala/org/http4s/client/prometheus/PrometheusClientMetrics.scala
zcox: I attempted to separate response duration into "response received" and "body processed" phases in eff0654. @rossabaker @ChristopherDavenport would you mind taking a look at the `TODO` comments? Seems to work correctly in our service, but this is starting to exceed my http4s knowledge.

prometheus-client-metrics/src/main/scala/org/http4s/client/prometheus/PrometheusClientMetrics.scala
```diff
@@ -62,19 +75,34 @@ object PrometheusClientMetrics {
   private def onResponse[F[_]: Sync](
       request: Request[F],
       response: Response[F],
-      start: Long,
-      end: Long,
-      metrics: ClientMetrics[F]): F[Unit] =
+      startTime: Long,
+      responseReceivedTime: Long,
+      metrics: ClientMetrics[F]): F[Response[F]] =
+    //TODO is this correct? update most of the metrics early, in case response body is discarded
```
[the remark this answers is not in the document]
zcox: IIUC, this code will produce two metrics if the body is fully consumed, with both `response_phase=response_received` and `response_phase=body_processed`, but if the body is not fully consumed then there is only one metric with `response_phase=response_received`. But if the histogram was updated with `body_processed` on `dispose` then there would always be both metrics?

I kinda slightly lean towards not producing the `body_processed` metric if the body is not fully consumed, but I don't have a strong opinion either. I'm happy to try out the `dispose` approach if that is preferred.

zcox:
`PrometheusClientMetrics.apply` currently takes this function parameter, which allows users to compute destination however they want, with a default of just empty string:

```scala
destination: Request[F] => String = { _: Request[F] => "" }
```

All of the attribute stuff was extracted into the `DestinationAttribute` module, which I'd be totally happy to just use internally in our own code base if it's not the right thing to include in http4s. Long-term, we may not even end up using it, and may get destination from the request some other way besides attributes.

PR #1964, opened 2018-07-26
Make UriTranslation check prefix contents before matching

### LINKED ISSUE #1962: UriTranslation.translateRoot behavior is misleading
The `UriTranslation.translateRoot` middleware only checks the length rather than the contents. This should probably either be made private, changed, or documented to describe the behavior.

Context from Gitter:

> John Howard @howardjohn Jul 05 18:54
> What I am experiencing though is:
> ```scala
>   val A = HttpService[IO] { case GET -> Root => Ok("a") }
>   val B = HttpService[IO] { case GET -> Root => Ok("b") }
>   val C = HttpService[IO] { case GET -> Root => Ok("c") }
> 
>   val service = URITranslation.translateRoot("a")(A) <+>
>     URITranslation.translateRoot("b")(B) <+>
>     URITranslation.translateRoot("other")(C)
> ```
> Then doing curl on `/a, /b, /other, /, /fake, /q`. I would expect these to output `a, b, c, 404, 404, 404`. The actual output I am getting is `a, a, c, a, c a`. It seems like it only actually checks the length of the Uri not the contents?
> 

> Ross A. Baker @rossabaker Jul 06 07:51
> @howardjohn You're right, it does do a length check. I thought it was a prefix check. It's usually used in the router, where the prefix is aleady checked. So it's an optimization there, but not good for standalone. We should reconsider this.

I would be able to make this change if you indicate which direction we should take

aeons:
This PR changes the behaviour of `UriTranslation` to take the contents of the prefix into account as well.

It inlines the old implementation in `Router`, to not do the check twice.

Fixes #1962

### DIFF
--- a/server/src/main/scala/org/http4s/server/middleware/URITranslation.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/URITranslation.scala
@@ -7,26 +7,22 @@ import cats.data.{Kleisli, OptionT}
 
 object URITranslation {
   def translateRoot[F[_]: Applicative](prefix: String)(service: HttpService[F]): HttpService[F] =
-    prefix match {
-      case "" | "/" => service
-      case prefix if prefix.startsWith("/") =>
-        val newCaret = prefix.length
+    if (prefix.isEmpty || prefix == "/") service
+    else {
+      val (slashedPrefix, unslashedPrefix) =
+        if (prefix.startsWith("/")) (prefix, prefix.substring(1))
+        else (s"/$prefix", prefix)
 
-        Kleisli { req: Request[F] =>
-          if (req.pathInfo.startsWith(prefix)) service(setCaret(req, newCaret))
-          else OptionT.none
-        }
-      case prefix =>
-        val newCaret = prefix.length + 1
+      val newCaret = slashedPrefix.length
 
-        Kleisli { req: Request[F] =>
-          val pi = req.pathInfo
-          val shouldTranslate =
-            if (pi.startsWith("/")) pi.substring(1).startsWith(prefix)
-            else pi.startsWith(prefix)
-          if (shouldTranslate) service(setCaret(req, newCaret))
-          else OptionT.none
-        }
+      Kleisli { req: Request[F] =>
+        val shouldTranslate =
+          req.pathInfo.startsWith(unslashedPrefix) || req.pathInfo.startsWith(slashedPrefix)
+
+        if (shouldTranslate) service(setCaret(req, newCaret))
+        else OptionT.none
+      }
+
     }
 
   private def setCaret[F[_]: Functor](req: Request[F], newCaret: Int): Request[F] = {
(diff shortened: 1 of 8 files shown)


aeons:
I'm not completely clear on the semantics of UriTranslation.

Should it just ignore whether there's a slash at the front of the prefix and request?

server/src/main/scala/org/http4s/server/middleware/URITranslation.scala
```diff
@@ -2,21 +2,33 @@ package org.http4s
...
+import cats.data.{Kleisli, OptionT}
 
 object URITranslation {
-  def translateRoot[F[_]: Functor](prefix: String)(service: HttpService[F]): HttpService[F] = {
-    val newCaret = prefix match {
-      case "/" => 0
-      case x if x.startsWith("/") => x.length
-      case x => x.length + 1
-    }
+  def translateRoot[F[_]: Applicative](prefix: String)(service: HttpService[F]): HttpService[F] =
```
ChristopherDavenport: This changes the function signature, so isn't binary compatible

aeons: Hmm, yes, that's true.
I'm not sure if this can be solved without being to signal a not-matched via `OptionT.none`.

aeons: @rossabaker could you take a look at this? Should I retarget master and just let this be a known bug in 0.18?

rossabaker: Our type is called `Uri`, but this object is called `URI`.  That's inconsistent.  We could deprecate thid old buggy one, and use this new implementation as `UriTranslation` in release-0.18.x.  Or `TranslateUri`.

aeons: :+1:

aeons:
I rebased and changed the following:

- Readd deprecated `URITranslation`
- Add the new implementation as `TranslateUri.apply` (I got a warning about only differing in case if it was called `UriTranslation`)

server/src/main/scala/org/http4s/server/middleware/URITranslation.scala
```diff
@@ -4,6 +4,7 @@ package middleware
 
 import cats.Functor
 
+@deprecated("Use org.http4s.server.middleware.UriTranslation instead", since = "0.18.16")
```
ChristopherDavenport: Do you mean `TranslateUri`?

rossabaker: Good catch.  If only we could compile our deprecation messages.

aeons: Yeah, I tried just changing the case, but got a compiler warning.

PR #1969, opened 2018-08-01
Add AwsSigner middleware for signing AWS requests

Daenyth:


aws-signer/src/main/scala/http4s/aws/middleware/AwsSigner.scala
```diff
@@ -0,0 +1,134 @@
...
+
+import java.util.Date
+
+import cats.data.Kleisli
+import cats.effect.Effect
+import cats.implicits._
+import fs2.Stream
+import org.http4s.client.Client
+import org.http4s.{Header, Request}
+import scodec.bits.ByteVector
```
jmcardon: I'm almost sure we don't need scodec for this.

aws-signer/src/main/scala/http4s/aws/middleware/AwsSigner.scala
```diff
@@ -0,0 +1,134 @@
...
+    val kRegion = hmac(kDate, bytes(zone))
+    val kService = hmac(kRegion, bytes(service))
+    val kSigning = hmac(kService, bytes("aws4_request"))
+    hmac(kSigning, bytes(string))
+  }
+
+  def apply[F[_]: Effect](request: Request[F],
+                          getDate: F[Date]): F[Request[F]] = {
+
+    request.body.compile.to[Array].flatMap { fullBodyArray =>
```
jmcardon: It's possible to do this without holding the full body in a contiguous array in memory

aws-signer/src/main/scala/http4s/aws/middleware/AwsSigner.scala
```diff
@@ -0,0 +1,134 @@
...
+      getDate.map { date =>
+        val fullBody = ByteVector(fullBodyArray)
+
+        val amzDate = Header("x-amz-date", fullDateFormat.format(date))
+
+        val amzHost = Header("Host",
+                             request.uri.host
+                               .map(_.toString)
+                               .getOrElse(throw new Exception("need a Host")))
```
jmcardon: `getOrElse throw`

aws-signer/src/main/scala/http4s/aws/middleware/AwsSigner.scala
```diff
@@ -0,0 +1,134 @@
...
+import cats.data.Kleisli
+import cats.effect.Effect
+import cats.implicits._
+import fs2.Stream
+import org.http4s.client.Client
+import org.http4s.{Header, Request}
+import scodec.bits.ByteVector
+
+object AwsSigner {
+  def apply[F[_]: Effect](id: String,
```
rossabaker: I think this might only need `Sync`.

aws-signer/src/main/scala/http4s/aws/middleware/AwsSigner.scala
```diff
@@ -0,0 +1,134 @@
...
+        .getOrElse(throw new Exception(s"'$secret' not base64"))
+  }
+}
+
+/** Implements AWS request signing v4 as an http4s middleware.
+  * Signing described here: https://docs.aws.amazon.com/general/latest/gr/signature-version-4.html
+  */
+class AwsSigner(key: AwsSigner.Key, zone: String, service: String) {
+  val Method = "AWS4-HMAC-SHA256"
+  val Charset = java.nio.charset.Charset.forName("UTF-8")
```
rossabaker: `StandardCharsets.UTF8` is already a constant for this.

aws-signer/src/main/scala/http4s/aws/middleware/AwsSigner.scala
```diff
@@ -0,0 +1,134 @@
...
+  }
+
+  def sign(string: String, date: java.util.Date) = {
+    val kSecret = bytes(s"AWS4${key.secret}")
+    val kDate = hmac(kSecret, bytes(shortDateFormat.format(date)))
+    val kRegion = hmac(kDate, bytes(zone))
+    val kService = hmac(kRegion, bytes(service))
+    val kSigning = hmac(kService, bytes("aws4_request"))
+    hmac(kSigning, bytes(string))
+  }
```
rossabaker: I think several of the things above this point don't need to be public, which gives us more latitude to evolve the implementation without breaking the API.

aws-signer/src/main/scala/http4s/aws/middleware/AwsSigner.scala
```diff
@@ -0,0 +1,134 @@
...
+
+  def apply[F[_]: Effect](request: Request[F],
+                          getDate: F[Date]): F[Request[F]] = {
```
rossabaker: How about a `Timer[F]` instead of `F[Date]`, to use something that's probably already on hand?

aws-signer/src/main/scala/http4s/aws/middleware/AwsSigner.scala
rossabaker: But we do need to buffer the entire body, since the headers that depend on it are not trailing headers.

build.sbt
```diff
@@ -251,6 +251,13 @@ lazy val twirl = http4sProject("twirl")
   .enablePlugins(SbtTwirl)
   .dependsOn(core, testing % "test->test")
 
+lazy val awsSigner = libraryProject("aws-signer")
```
rossabaker: Would `aws` make more sense than `aws-signer`, in case we add more integrations?  Perhaps those would have to depend on the aws-sdk, whereas this does not, so maybe I'm wrong. :thinking:

aws-signer/src/main/scala/http4s/aws/middleware/AwsSigner.scala
```diff
@@ -0,0 +1,134 @@
...
+    client.copy(open = Kleisli { req =>
+      signer[F](req, Effect[F].delay(new Date)).flatMap(client.open(_))
+    })
+  }
+
+  case class Key(id: String, secret: String) {
+    def bytes =
+      ByteVector
+        .fromBase64(secret)
+        .getOrElse(throw new Exception(s"'$secret' not base64"))
```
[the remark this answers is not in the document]
rossabaker: I'm overlooking where it even gets called.  Maybe we don't need it at all?

build.sbt
Daenyth: I think in terms of modularity, I'd agree that `aws` might depend on aws-sdk, and in fact might also depend on this module.

I think it makes sense as-is

Daenyth:
Thanks for the feedback. I'll try to get to it as soon as I can

aws-signer/src/main/scala/http4s/aws/middleware/AwsSigner.scala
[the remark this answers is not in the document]
jmcardon: we could `.raiseError` or nor, if it's optionT, just return none.

aws-signer/src/main/scala/http4s/aws/middleware/AwsSigner.scala
jmcardon: I meant to say that you can do the hash calculation by traversing the whole body once, but not reading it.

rossabaker: It doesn't need a finalizer to be non-repeatable:

```scala
Stream.repeatEval(IO(random.nextInt.toByte)).take(1024)
```

aws-signer/src/main/scala/http4s/aws/middleware/AwsSigner.scala
```diff
@@ -0,0 +1,136 @@
...
+object AwsSigner {
+  def apply[F[_]: Sync](id: String, secret: String, zone: String, service: String)(
+      client: Client[F]): Client[F] = {
+    val signer = new AwsSigner(Key(id, secret), zone, service)
+    client.copy(open = Kleisli { req =>
+      signer[F](req, Sync[F].delay(new Date)).flatMap(client.open(_))
+    })
+  }
+
+  case class Key(id: String, secret: String)
```
rossabaker: Nit: I like to make all case classes `final` or `sealed abstract`.

aws-signer/src/main/scala/http4s/aws/middleware/AwsSigner.scala
```diff
@@ -0,0 +1,136 @@
...
+}
+
+/** Implements AWS request signing v4 as an http4s middleware.
+  * Signing described here: https://docs.aws.amazon.com/general/latest/gr/signature-version-4.html
+  */
+class AwsSigner(key: AwsSigner.Key, zone: String, service: String) {
```
rossabaker: Do you want this constructor to be public as an alternate way of creating one with a `Key`?  If not, you can whittle down the surface area of the API some more with a private constructor, and then even make `Key` private.

aws-signer/src/main/scala/http4s/aws/middleware/AwsSigner.scala
rossabaker: I think everything above this can be private.

aws-signer/src/main/scala/http4s/aws/middleware/AwsSigner.scala
```diff
@@ -0,0 +1,136 @@
...
+  def sign(string: String, date: java.util.Date) = {
+    val kSecret = bytes(s"AWS4${key.secret}")
+    val kDate = hmac(kSecret, bytes(shortDateFormat.format(date)))
+    val kRegion = hmac(kDate, bytes(zone))
+    val kService = hmac(kRegion, bytes(service))
+    val kSigning = hmac(kService, bytes("aws4_request"))
+    hmac(kSigning, bytes(string))
+  }
+
+  def apply[F[_]: Sync](request: Request[F], getDate: F[Date]): F[Request[F]] =
```
rossabaker: I might have called this method `sign`.

rossabaker: Continuing a conversation that GitHub has decided we're done with, where I proposed `Timer[F]` over `F[Date]`:

> I think that actually makes this harder to test, because you'd need to construct a whole timer, whereas this can be IO.pure(mydate)

> The intended "main" interface to this is via apply, where the user doesn't actually need to pick a date.

Oh, I see.  You're right, this isn't as consequential as I initially thought.

I guess I have a modest preference for `Timer` as a consistent way that time is parameterized elsewhere in cats-effect.  It's on its way to being slimmed down to `Clock`, which is two methods instead of one.  But two is still bigger than one, and I'd at least work in terms of `java.time.Instant` instead of the mutable and should-have-been-deprecated-a-decade-ago `java.util.Date`.

aws-signer/src/main/scala/org/http4s/aws/middleware/AwsSigner.scala
```diff
@@ -0,0 +1,123 @@
...
+  def apply[F[_]: Sync](request: Request[F], getDate: F[Date]): F[Request[F]] =
+    request.body.compile.to[Array].flatMap { fullBody =>
+      getDate.map { date =>
+        val amzDate = Header("x-amz-date", fullDateFormat.format(date))
+
+        val amzHost = Header(
+          "Host",
+          request.uri.host
+            .map(_.toString)
+            .getOrElse(throw new Exception("need a Host"))) // TODO fix
```
Daenyth: note to self: fix

aws-signer/src/main/scala/http4s/aws/middleware/AwsSigner.scala
Daenyth: will do

aws-signer/src/main/scala/org/http4s/aws/middleware/AwsSigner.scala
```diff
@@ -0,0 +1,123 @@
+package org.http4s.aws.middleware
+
+import java.util.Date
```
Daenyth: note to self: Instant

aws-signer/src/main/scala/org/http4s/aws/middleware/AwsSigner.scala
```diff
@@ -0,0 +1,123 @@
...
+  val Charset = StandardCharsets.UTF_8
+
+  private def dateFormat(s: String) = {
+    val f = new java.text.SimpleDateFormat(s)
+    f.setTimeZone(java.util.TimeZone.getTimeZone("UTC"))
+    f
+  }
+
+  val fullDateFormat = dateFormat("YYYYMMdd'T'HHmmss'Z'")
+  val shortDateFormat = dateFormat("YYYYMMdd")
```
ritschwumm: these are not thread safe

aws-signer/src/main/scala/org/http4s/internal/BytesToHex.scala
```diff
@@ -0,0 +1,14 @@
...
+object BytesToHex {
+  private val hexArray: Array[Char] = "0123456789ABCDEF".toCharArray();
+  def bytesToHex(bytes: Array[Byte]): String = {
+    val hexChars = new Array[Char](bytes.length * 2)
+    for (j <- 0 until bytes.length) {
+      val v: Int = bytes(j) & 0xFF
+      hexChars(j * 2) = hexArray(v >>> 4)
+      hexChars(j * 2 + 1) = hexArray(v & 0x0F)
+    }
+    new String(hexChars)
```
ritschwumm: this should include an encoding, you probably don't want to rely on the platform default encoding.

Daenyth: Well, the characters involved are all within the ASCII range, but I can still set it.

aws-signer/src/main/scala/org/http4s/aws/middleware/AwsSigner.scala
Daenyth: Will be changing to `Instant`.

In what way are they not thread safe btw?

rossabaker: > Date formats are not synchronized. It is recommended to create separate format instances for each thread. If multiple threads access a format concurrently, it must be synchronized externally.

[Reference](https://docs.oracle.com/javase/7/docs/api/java/text/SimpleDateFormat.html)

aws-signer/src/main/scala/org/http4s/internal/BytesToHex.scala
rossabaker: This is right when we construct a string from `Array[Byte]`, but this is an `Array[Char]`.  There's no encoding to specify.

rossabaker:
Still interesting, but not a blocker for M2.

Daenyth:
Closing this as it's quite stale and libraries like aws4s now provide this

PR #1974, opened 2018-08-03
Add client middleware for dropwizard metrics

luisdeltoro:


client-metrics/src/main/scala/org/http4s/client/metrics/Metrics.scala
```diff
@@ -0,0 +1,71 @@
...
+import cats.effect._
+import cats.implicits._
+import com.codahale.metrics.{Counter, MetricRegistry, Timer}
+import java.util.concurrent.TimeUnit
+import org.http4s.{Request, Status}
+import org.http4s.client.{Client, DisposableResponse}
+
+object Metrics {
+  def apply[F[_]: Sync](registry: MetricRegistry, prefix: String = "org.http4s.client")(
+      client: Client[F]): Client[F] = {
```
rossabaker: This is similar to a conversation we're having in #1969: if we added a `Timer[F]` instead of calling `System.nanoTime`, we could parameterize the clock and thus be able to test the response times.

client-metrics/src/main/scala/org/http4s/client/metrics/Metrics.scala
```diff
@@ -0,0 +1,71 @@
...
+
+    def withMetrics(metrics: MetricsCollection)(req: Request[F]): F[DisposableResponse[F]] =
+      for {
+        start <- Sync[F].delay(System.nanoTime())
+        _ <- Sync[F].delay(metrics.activeRequests.inc())
+        resp <- client.open(req)
+        _ <- Sync[F].delay(metrics.activeRequests.dec())
```
rossabaker: This will still be active through processing of the response body.  We'd want to hook into the dispose method.

client-metrics/src/main/scala/org/http4s/client/metrics/Metrics.scala
```diff
@@ -0,0 +1,71 @@
...
+        iResp <- Sync[F].delay(instrumentResponse(start, metrics, resp))
+      } yield iResp
+
+    def instrumentResponse(
+        start: Long,
+        metrics: MetricsCollection,
+        disposableResponse: DisposableResponse[F]): DisposableResponse[F] = {
+      val newDisposable = for {
+        elapsed <- Sync[F].delay(System.nanoTime() - start)
```
rossabaker: This is measuring time to first byte.  Should we be measuring that _and_ total time?  See #1961 for related discussion.

luisspringernature: I thought this will be measuring total time, as it's being recorded when disposing the response...

client-metrics/src/main/scala/org/http4s/client/metrics/Metrics.scala
luisspringernature: So, decreasing the active requests counter should be moved to the disponse handler of DisposableResponse right?

client-metrics/src/main/scala/org/http4s/client/metrics/Metrics.scala
luisspringernature: Sure, that makes total sense. Talking about tests, I have not seen any tests in the other metrics middleware. Is it desired that they come with no tests?

luisdeltoro: Sure, that makes total sense. Talking about tests, I have not seen any tests in the other metrics middleware. Is it desired that they come with no tests?

client-metrics/src/main/scala/org/http4s/client/metrics/Metrics.scala
luisdeltoro: So, decreasing the active requests counter should be moved to the disponse handler of DisposableResponse right?

client-metrics/src/main/scala/org/http4s/client/metrics/Metrics.scala
luisdeltoro: I thought this will be measuring total time, as it's being recorded when disposing the response...

luisdeltoro:
Sure. I like the destination label function concept. I'll do that here as well.

client-metrics/src/main/scala/org/http4s/client/metrics/Metrics.scala
rossabaker: Timer didn't exist, so those tests would have been harder to write.  And I think we just got lazy on the other parts.  Tests would be great, but not mandatory.

client-metrics/src/main/scala/org/http4s/client/metrics/Metrics.scala
rossabaker: Yes.  Probably with `.guarantee`, just in case something throws.

client-metrics/src/main/scala/org/http4s/client/metrics/Metrics.scala
rossabaker: You're right.  I was looking for a finalizer on the response body, like we do on the server side, but this will work and is the right way to do it.

client-metrics/src/main/scala/org/http4s/client/metrics/Metrics.scala
luisdeltoro: Could you elaborate on this please? What should we done with .guarantee? I mean updating the metrics happen before calling the original dispose method. So, as I see it, the only that go wrong here is updating the metrics itself and in that case the .guarantee would probably not help... Or what's your thinking here?

rossabaker: The `guarantee` would help if the metrics call fails, because the guaranteed effect would run, and the original error would be re-raised.

I've been bitten a few times by Prometheus throwing, but it's always on the first call, and always because I did something stupid in naming.  So it's probably not worth worrying about.

build.sbt
```diff
@@ -95,6 +95,16 @@ lazy val client = libraryProject("client")
     theDsl % "test->compile",
     scalaXml % "test->compile")
 
+lazy val clientMetrics = libraryProject("client-metrics")
+  .settings(
+    description := "Support for Dropwizard Metrics on the client",
+    libraryDependencies ++= Seq(
+      metricsCore,
+      metricsJson
```
rossabaker: Do we use metrics-json, or is this a copy and paste from the server?  I think metrics-core is enough.

client-metrics/src/main/scala/org/http4s/client/metrics/Metrics.scala
luisdeltoro: ok

build.sbt
luisdeltoro: yes, you are totally right. I've removed that dependency

rossabaker:
Hmm.  Does `Http4sSpec` not provide a `Timer[IO]`?  I thought it would.  It probably should?  How has this not come up already? :thinking:

luisdeltoro:
Apparently, it was not providing it. I've added there. I hope it's ok

luisdeltoro:
Cool. Let me know if there is anything else I can do.

rossabaker:
We usually get two committers' approvals, so we'll merge it after one more approval.

luisdeltoro:
Thanks for the explanation. I'll have a look at the tagless thing.

PR #1975, opened 2018-08-03
[WIP] Alternate websocket interface

jmcardon:
We've been asked a bunch of times for a simpler way to handle websockets. Our current websocket approach is very low level and good for people that need granular control over frames, but it does not help the case of constructing an in-out flow easily. 

This PR adds `WSFSM` (I suck at naming so suggestions if you hate it pls), which gives a simple `Msg => F[Msg]` flow to websockets, which should make constructing certain websocket programs easier.

Quips:
- Messages are no longer fragmented, for simplicity
- In [Section 5.4 of the websocket RFC](https://tools.ietf.org/html/rfc6455#section-5.4) (see the last bullet point), fragment types, if aggregated, depend on the first fragment received. As such, the type of the entire fragment aggregation depends on the first fragment, and we treat messages as such. The edge case is continuation frames. If a continuation frame is received first, we treat the whole fragment as binary. 
- I needed `compareAndSet` and `increment` semantics thus I opted for manual use of `AtomicReference` and `AtomicInteger` over `Ref`. 
- `Msg => F[Msg]` isn't great in the "ignore the message" case, so maybe I'd like to consider `Msg => F[Unit]`

### DIFF
--- a/server/src/main/scala/org/http4s/server/websocket/FSMAlgebra.scala
+++ b/server/src/main/scala/org/http4s/server/websocket/FSMAlgebra.scala
@@ -1,0 +1,121 @@
+package org.http4s.server.websocket
+
+import java.nio.charset.StandardCharsets.UTF_8
+import java.util.concurrent.atomic.{AtomicInteger, AtomicReference}
+import cats.effect.{Concurrent, Sync}
+import cats.syntax.all._
+import fs2.async.mutable.Queue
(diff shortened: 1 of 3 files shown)


rossabaker:
Was this closed intentionally?

jmcardon:
Yeah, I decided not to push through with it after chatting on gitter, unless enough people want it.

PR #1978, opened 2018-08-08
WIP: Mount HttpApp to ServerBuilders

ChristopherDavenport:


ChristopherDavenport:
The goal of this PR is to change the Blaze server to mount with a single HttpApp, and the builders of the other apps can still conceivably work with the old model but that would depreceate the builder.

ChristopherDavenport:
Closed in favor of #2054

PR #1981, opened 2018-08-08
Separate Http4sDsl's base into a minimal DSL class

andimiller:
This adds a new import under `org.http4s.dsl.minimal._` which allows you to import all of the DSL other than the `Response` related classes.

Is this a valuable addition to the DSL? does this help people who want to be careful with their `F` and avoid mixing in `Http4sDsl`? Should we also offer a syntax which lets people do `Ok[IO]` so that none of the DSL requires a type parameter to import?

### DIFF
--- a/dsl/src/main/scala/org/http4s/dsl/Http4sDsl.scala
+++ b/dsl/src/main/scala/org/http4s/dsl/Http4sDsl.scala
@@ -3,7 +3,9 @@ package org.http4s.dsl
 import org.http4s.{Http4s, Method}
 import org.http4s.dsl.impl._
 
-trait Http4sDsl[F[_]] extends Http4s with Methods with Statuses with Responses[F] with Auth {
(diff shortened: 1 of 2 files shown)


SystemFw:
Nice!

I'm not sure about the name `minimal` but no better suggestion.

> Is this a valuable addition to the DSL?
imho, yes

> does this help people who want to be careful with their F and avoid mixing in Http4sDsl?

I do think it might only help advanced users (unless we document it really well), but those users are probably going to like it

>  Should we also offer a syntax which lets people do Ok[IO] so that none of the DSL requires a type parameter to import?

Doesn't `import dsl.io._` take care of this?

Of course the counter argument is the proliferation of ways to do the same thing, which can be frustrating especially wrt syntax

andimiller:
This currently passes tests in the `dsl` module, I haven't updated dependent modules but it should be fairly simple.

rossabaker:
Still interesting, but not a blocker for M2.

PR #1983, opened 2018-08-10
Upgrade async-http-client, cats, jawn-json4s

rossabaker:
fs2-1.0.0-M3 is more complicated, so will be done separately.

### DIFF
--- a/project/Http4sPlugin.scala
+++ b/project/Http4sPlugin.scala
@@ -138,6 +138,7 @@ object Http4sPlugin extends AutoPlugin {
     },
 
     dependencyUpdatesFilter -= moduleFilter(organization = "javax.servlet"), // servlet-4.0 is not yet supported by jetty-9 or tomcat-9, so don't accidentally depend on its new features
+    dependencyUpdatesFilter -= moduleFilter(organization = "org.json4s"), // json4s-3.6 is not binary compatible with jawn-json4s-0.12
     dependencyUpdatesFilter -= moduleFilter(organization = "org.scalacheck"), // scalacheck-1.14 is incompatible with cats-laws-1.1
     dependencyUpdatesFilter -= moduleFilter(organization = "org.specs2"), // specs2-4.2 is incompatible with scalacheck-1.13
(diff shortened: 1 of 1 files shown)


project/Http4sPlugin.scala
```diff
@@ -279,10 +280,10 @@ object Http4sPlugin extends AutoPlugin {
 
   lazy val alpnBoot                         = "org.mortbay.jetty.alpn" %  "alpn-boot"                 % "8.1.12.v20180117"
   lazy val argonaut                         = "io.argonaut"            %% "argonaut"                  % "6.2.2"
-  lazy val asyncHttpClient                  = "org.asynchttpclient"    %  "async-http-client"         % "2.4.9"
+  lazy val asyncHttpClient                  = "org.asynchttpclient"    %  "async-http-client"         % "2.5.2"
   lazy val blaze                            = "org.http4s"             %% "blaze-http"                % "0.14.0-M3"
   lazy val boopickle                        = "io.suzaku"              %% "boopickle"                 % "1.3.0"
-  lazy val cats                             = "org.typelevel"          %% "cats-core"                 % "1.1.0"
+  lazy val cats                             = "org.typelevel"          %% "cats-core"                 % "1.2.0"
```
rossabaker: I did a bad merge and reverted this.  This sets it back.

project/Http4sPlugin.scala
```diff
@@ -300,7 +301,7 @@ object Http4sPlugin extends AutoPlugin {
   lazy val gatlingHighCharts                = "io.gatling.highcharts"  %  "gatling-charts-highcharts" % gatlingTest.revision
   lazy val http4sWebsocket                  = "org.http4s"             %% "http4s-websocket"          % "0.2.1"
   lazy val javaxServletApi                  = "javax.servlet"          %  "javax.servlet-api"         % "3.1.0"
-  lazy val jawnJson4s                       = "org.spire-math"         %% "jawn-json4s"               % "0.12.1"
+  lazy val jawnJson4s                       = "org.spire-math"         %% "jawn-json4s"               % "0.12.2"
```
rossabaker: There's 0.13.0 now, which looks the same as 0.12.2, but would require a jawn-fs2 bump to avoid warnings.

This compiled agianst json4s-3.5, and we'll need to bump that to get to 3.6. :weary:

Issue #1980, opened 2018-08-08
http4s banner may be displayed incorrectly


pshirshov:
Seems like the operation order is not defined:

```
I 2018-08-08T13:24:21.273+01:00[Europe/Dublin] ...nsportTest:1  (BlazeBuilder.scala:281)   _   _   _        _ _     
D 2018-08-08T13:24:21.275+01:00[Europe/Dublin] ...ector-0-0:12  (SelectorLoop.scala:122) Channel initialized.
I 2018-08-08T13:24:21.275+01:00[Europe/Dublin] ...nsportTest:1  (BlazeBuilder.scala:281)  | |_| |_| |_ _ __| | | ___
I 2018-08-08T13:24:21.277+01:00[Europe/Dublin] ...nsportTest:1  (BlazeBuilder.scala:281)  | ' \  _|  _| '_ \_  _(_-<
I 2018-08-08T13:24:21.278+01:00[Europe/Dublin] ...nsportTest:1  (BlazeBuilder.scala:281)  |_||_\__|\__| .__/ |_|/__/
I 2018-08-08T13:24:21.279+01:00[Europe/Dublin] ...nsportTest:1  (BlazeBuilder.scala:281)              |_|
```

aeons:
Does this happen consistently?
I don't think we can really control the order of log statements from different places.

pshirshov:
From time to time.
You can't control but you may print the whole banner in one message. Or just remove it.

ChristopherDavenport:
I like our banner, it looks nice and only appears once on server start. I'm very against its removal. There is the possibility of conflicting log messages, however that is just a reality about logging. 

If you set the banner in the blaze builder to an empty List or Vector this will mute the banner for you - https://github.com/http4s/http4s/blob/master/blaze-server/src/main/scala/org/http4s/server/blaze/BlazeBuilder.scala#L68

rossabaker:
What happens if we `mkString("\n")` it?  Most log formats aren't going to indent the subsequent lines properly, I suspect?  If it had a leading newline, it might work.

pshirshov:
Yup, you may use multiline output and add a leading newline. That would work.

rossabaker:
Would you be interested in submitting a PR?

pshirshov:
Okay, will do.

PR #1985, opened 2018-08-10
banner is being printed with a single logger call

pshirshov:
Fix for #1980

### DIFF
--- a/blaze-server/src/main/scala/org/http4s/server/blaze/BlazeBuilder.scala
+++ b/blaze-server/src/main/scala/org/http4s/server/blaze/BlazeBuilder.scala
@@ -278,7 +278,11 @@ class BlazeBuilder[F[_]](
         s"BlazeServer($address)"
     }
 
-    banner.foreach(logger.info(_))
+    Option(banner)
+      .filter(_.nonEmpty)
+      .map(_.mkString("\n", "\n", ""))
+      .foreach(logger.info(_))
+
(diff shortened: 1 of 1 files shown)


blaze-server/src/main/scala/org/http4s/server/blaze/BlazeBuilder.scala
```diff
@@ -278,7 +278,11 @@ class BlazeBuilder[F[_]](
         s"BlazeServer($address)"
     }
 
-    banner.foreach(logger.info(_))
+    Option(banner)
```
jmcardon: couldn't you just do this with a single `foldLeft` + an `isEmpty`?

pshirshov: I may though don't see much benefits in that. This is bit more verbose but looks more idiomatic (for my taste)

jmcardon: more of a code style thing, but if Ross doesn't mind I'll live with it.

rossabaker: If we start doing this as a single log statement, I don't think it needs to continue to be a `Seq`.  I think this works for 0.18 for bincompat, and we might simplify it on the server builders for 0.19.

PR #1988, opened 2018-08-11
 json4s-3.6 is compatible with jawn-0.12.2?

rossabaker:
Not doing on release-0.18.x, because json4s-3.5 is not compatible with 3.6.  But jawn only depends on the ast, and from my research on non/jawn#113, it seems compatible.

### DIFF
--- a/project/Http4sPlugin.scala
+++ b/project/Http4sPlugin.scala
@@ -138,7 +138,6 @@ object Http4sPlugin extends AutoPlugin {
     },
 
     dependencyUpdatesFilter -= moduleFilter(organization = "javax.servlet"), // servlet-4.0 is not yet supported by jetty-9 or tomcat-9, so don't accidentally depend on its new features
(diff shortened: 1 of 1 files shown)


project/Http4sPlugin.scala
```diff
@@ -219,14 +218,14 @@ object Http4sPlugin extends AutoPlugin {
...
-   * scalaz-7.2 "a" versions for 0.15.x and 0.16.x.
-   */
+    * @return the version we want to document, for example in tuts,
+    * given the version being built.
+    *
+    * For snapshots after a stable release, return the previous stable
+    * release.  For snapshots of 0.16.0 and 0.17.0, return the latest
+    * milestone.  Otherwise, just return the current version.  Favors
+    * scalaz-7.2 "a" versions for 0.15.x and 0.16.x.
+    */
```
aeons: wat is dis

rossabaker: An artifact of an emacs minor mode that doesn't play well with scala-mode.

aeons: aha

PR #1991, opened 2018-08-13
QueryParamEncode/Decoder for Uri

Lasering:
As discussed on Gitter
https://gitter.im/http4s/http4s?at=5b6da0b3e9b96f2c98eccd0f

### DIFF
--- a/core/src/main/scala/org/http4s/QueryParam.scala
+++ b/core/src/main/scala/org/http4s/QueryParam.scala
@@ -3,7 +3,6 @@ package org.http4s
 import cats._
 import cats.data._
 import cats.implicits._
-
 final case class QueryParameterKey(value: String) extends AnyVal
 
 final case class QueryParameterValue(value: String) extends AnyVal
@@ -93,7 +92,8 @@ object QueryParamEncoder {
       override def encode(value: String) =
         QueryParameterValue(value)
     }
-
+  implicit lazy val uriQueryParamEncoder: QueryParamEncoder[Uri] =
+    QueryParamEncoder[String].contramap(_.renderString)
 }
 
 /**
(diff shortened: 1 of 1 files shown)


core/src/main/scala/org/http4s/QueryParam.scala
```diff
@@ -195,4 +195,5 @@ object QueryParamDecoder {
       def decode(value: QueryParameterValue): ValidatedNel[ParseFailure, String] =
         value.value.validNel
     }
+  implicit lazy val uriQueryParamDecoder: QueryParamDecoder[Uri] = QueryParamDecoder[String].map(Uri.unsafeFromString)
```
rossabaker: This will throw if the String is an invalid `Uri`.  We should have a function on `QueryParamDecoder` that works through a `ParseResult`, so we can use the safe conversion from `Uri`.

core/src/main/scala/org/http4s/QueryParam.scala
```diff
@@ -3,7 +3,7 @@ package org.http4s
 import cats._
 import cats.data._
 import cats.implicits._
-
+import org.http4s.Uri
```
rossabaker: I think this is the cause of the compilation failure.

rossabaker:
Everything we're doing here should be binary compatible.  Let's retarget release-0.18.x, and we can get this out in production sooner.

core/src/main/scala/org/http4s/QueryParam.scala
Lasering: What should be done for a Parse failure?

Lasering:
No longer invoke `Uri.unsafeFromString`.

core/src/main/scala/org/http4s/QueryParam.scala
```diff
@@ -195,4 +194,8 @@ object QueryParamDecoder {
       def decode(value: QueryParameterValue): ValidatedNel[ParseFailure, String] =
         value.value.validNel
     }
+  implicit val uriQueryParamDecoder: QueryParamDecoder[Uri] = (value: QueryParameterValue) => {
+    import cats.syntax.either._
```
aeons: This shouldn't be necessary with the `import cats.implicits._` above.

Lasering:
No longer use SAM types as they are not available in Scala 2.11

rossabaker:
Cherry-picked onto release-0.18.x, will be in 0.18.17.

Issue #1990, opened 2018-08-12
fs2-1.0.0-M4 upgrade


rossabaker:
We want to upgrade to fs2-1.0.0-M4 before the next 0.19 milestone.

* [ ] jawn-fs2 (need to release again for fs2-1.0.0-M4)
* [ ] zainab-ali/fs2-reactive-streams#51
* [ ] fix compilation failures in http4s

SystemFw:
Is the fs2-reactive-streams stuff on a branch somewhere?

rossabaker:
No.  All I did was upgrade the build and hope for the best.  Do you want to take a crack at it?

SystemFw:
Yeah I can have a look :)

rossabaker:
Bumping this from M3 to M4 due to the fs2-reactive-streams issue.

rossabaker:
Ticket checklist out of date, superseded by milestone.

Issue #1956, opened 2018-07-20
blaze client: don't build an `SSLContext` if it is not required.


nigredo-tori:
At the moment `Http1Client` [falls back](https://github.com/http4s/http4s/blob/211a15fb63eb44fe509da59f32b2ee58d06e77b1/blaze-client/src/main/scala/org/http4s/client/blaze/Http1Support.scala#L32) on `SSLContext.getDefault` if another  SSL context is not provided. In theory this should be a reasonable thing to do. But there is at least one crypto provider (CryptoPro JCP) that can break `SSLContext.getDefault`, making it throw an exception unless some extra configuration happens. This can be worked around by fiddling around with system properties, or by building an `SSLContext` manually, however this gets a bit fragile and/or cubersome in the general case. This is jarring, since in my case (simple microservice interaction on a single machine) HTTPS client is not even used!

I recognize that `http4s` is not really at fault here, however it would be nice to fix this behavior. One way would be to defer the `SSLContext` initialization until it is required.

PR #1993, opened 2018-08-14
Upgrade to fs2-1.0.0-M5

### LINKED ISSUE #229: ServletContextOps should detect the servlet I/O version
Out of the box, the xsbt-web-plugin uses Servlet 3.0 containers.  ServletContextSyntax blindly uses non-blocking servlet I/O.  It should check the Servlet API version and choose an appropriate servlet I/O for war deployments.

rossabaker:
See #1990.

Fixes #229.


core/src/main/scala/org/http4s/internal/ChunkWriter.scala
```diff
@@ -0,0 +1,22 @@
+package org.http4s.internal
```
rossabaker: Moved to internal because I don't want to promise this as public API

examples/docker/src/main/scala/com/example/http4s/Example.scala
```diff
@@ -1,21 +1,22 @@
...
 
-  def stream(args: List[String], requestShutdown: IO[Unit]): Stream[IO, ExitCode] =
+  def run(args: List[String]): IO[ExitCode] =
     BlazeBuilder[IO]
       .bindHttp(8080, "0.0.0.0")
       .mountService(service)
       .serve
+      .compile
+      .drain
+      .as(ExitCode.Success)
```
rossabaker: This isn't a great idiom.  We need to figure out what this looks like in the `IOApp` world.

servlet/src/main/scala/org/http4s/servlet/AsyncHttp4sServlet.scala
```diff
@@ -54,16 +41,17 @@ class AsyncHttp4sServlet[F[_]](
...
-        toRequest(servletRequest).fold(
-          onParseFailure(_, servletResponse, bodyWriter),
-          handleRequest(ctx, _, bodyWriter)
-        )) {
-        case Right(()) =>
-          IO(ctx.complete())
-        case Left(t) =>
-          IO(errorHandler(servletRequest, servletResponse)(t))
-      }
+      F.runAsync(
```
rossabaker: This means we are no longer shifting off the servlet thread pool.

Discussion [starts about here](https://gitter.im/http4s/http4s?at=5b81987294f8164c179bf2d2).

servlet/src/main/scala/org/http4s/servlet/syntax/ServletContextSyntax.scala
```diff
@@ -14,17 +13,18 @@ trait ServletContextSyntax {
 
 final class ServletContextOps private[syntax] (val self: ServletContext) extends AnyVal {
 
-  /** Wraps an [[HttpRoutes]] and mounts it as an [[AsyncHttp4sServlet]] */
-  def mountService[F[_]: ConcurrentEffect](
+  /** Wraps an [[HttpRoutes]] and mounts it as an [[AsyncHttp4sServlet]]
+    *
+    * Assumes non-blocking servlet IO is available, and thus requires at least Servlet 3.1.
```
rossabaker: The alternative is to require a blocking `ExecutionContext` that most users aren't going to need, or to provide an `Option[ServletIO]` which defaults to `None`, which is a pattern I never really liked much.

servlet/src/main/scala/org/http4s/servlet/AsyncHttp4sServlet.scala
```diff
@@ -118,13 +106,11 @@ class AsyncHttp4sServlet[F[_]](
...
       service: HttpRoutes[F],
-      asyncTimeout: Duration = Duration.Inf,
-      executionContext: ExecutionContext = ExecutionContext.global): AsyncHttp4sServlet[F] =
+      asyncTimeout: Duration = Duration.Inf): AsyncHttp4sServlet[F] =
     new AsyncHttp4sServlet[F](
       service,
       asyncTimeout,
-      executionContext,
-      BlockingServletIo[F](DefaultChunkSize),
+      NonBlockingServletIo[F](DefaultChunkSize),
```
rossabaker: This is going to hurt benchmarks, but help in scalability.  We might gain a fair bit of it back losing that other shift.  And it's still pluggable.

examples/docker/src/main/scala/com/example/http4s/Example.scala
jmcardon: we're also missing a shutdown hook unfortunately.

server/src/main/scala/org/http4s/server/middleware/ChunkAggregator.scala
```diff
@@ -8,19 +8,20 @@ import cats.effect.Sync
...
     http.flatMapF { response =>
       f(
         response.body.chunks.compile
-          .fold((Segment.empty[Byte], 0L)) {
-            case ((seg, len), c) => (seg ++ c.toSegment, len + c.size)
-          }
-          .map {
-            case (body, len) =>
-              removeChunkedTransferEncoding[G](response.withBodyStream(Stream.segment(body)), len)
+          .to[Vector]
```
jmcardon: `.to[Array].map(vec=> Chunk.bytes(vec))`?
I'm hoping `CanBuildFrom` uses an arraybuffer or something.

servlet/src/main/scala/org/http4s/servlet/AsyncHttp4sServlet.scala
jmcardon: I don't think we bench servlets so it's all good.

servlet/src/main/scala/org/http4s/servlet/syntax/ServletContextSyntax.scala
jmcardon: How many people _want_ blocking servlets in today's world is a good question.

examples/docker/src/main/scala/com/example/http4s/Example.scala
rossabaker: Was that ever useful, rather than just terminating the server? That's why it disappeared from IOApp.

servlet/src/main/scala/org/http4s/servlet/AsyncHttp4sServlet.scala
rossabaker: We should, to monitor for regressions and let people make informed choices. We just never automated anything.

servlet/src/main/scala/org/http4s/servlet/syntax/ServletContextSyntax.scala
rossabaker: Some of those managed platforms are stuck in antiquity. We've had a few requests. Though it has been a while.

It is also higher throughput, if you can afford a thread per request.

blaze-core/src/main/scala/org/http4s/blazecore/util/CachingChunkWriter.scala
```diff
@@ -20,30 +21,38 @@ private[http4s] class CachingChunkWriter[F[_]](
...
+    bodyBuffer += chunk
+    size += chunk.size
   }
 
+  private def clear(): Unit = {
+    bodyBuffer.clear()
+    size = 0
+  }
+
+  private def toChunk: Chunk[Byte] = Chunk.concatBytes(bodyBuffer.toSeq)
```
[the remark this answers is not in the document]
rossabaker: And avoid the foldLeft for size? True.

This whole thing feels like a missing ChunkBuffer or something.

blaze-core/src/main/scala/org/http4s/blazecore/websocket/Http4sWSStage.scala
```diff
@@ -101,11 +101,6 @@ object Http4sWSStage {
...
 
-  private def unsafeRunSync[F[_], A](
-      fa: F[A])(implicit F: ConcurrentEffect[F], ec: ExecutionContext): A =
-    Deferred[IO, Either[Throwable, A]].flatMap { p =>
-      F.runAsync(Async.shift(ec) *> fa) { r =>
-        p.complete(r)
-      } *> p.get.rethrow
-    }.unsafeRunSync
+  private def unsafeRunSync[F[_], A](fa: F[A])(implicit F: ConcurrentEffect[F]): A =
+    fa.toIO.unsafeRunSync
```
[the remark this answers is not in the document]
rossabaker: That'll probably land before this. I'll deal with the conflict then.

tomcat/src/main/scala/org/http4s/server/tomcat/TomcatBuilder.scala
```diff
@@ -5,23 +5,24 @@ package tomcat
...
 import org.log4s.getLogger
 import scala.collection.JavaConverters._
 import scala.collection.immutable
-import scala.concurrent.ExecutionContext
 import scala.concurrent.duration._
 
 sealed class TomcatBuilder[F[_]] private (
     socketAddress: InetSocketAddress,
-    private val executionContext: ExecutionContext,
+    externalExecutor: Option[Executor],
```
rossabaker: As with Jetty, we want to shift blocking IO back to this.  It's going to hurt if it's internal.

jetty/src/main/scala/org/http4s/server/jetty/JettyBuilder.scala
```diff
@@ -13,17 +13,16 @@ import org.eclipse.jetty.servlet.{FilterHolder, ServletContextHandler, ServletHo
...
 import org.http4s.servlet.{AsyncHttp4sServlet, ServletContainer, ServletIo}
 import org.log4s.getLogger
-import scala.concurrent.ExecutionContext
 import scala.collection.immutable
 import scala.concurrent.duration._
 
 sealed class JettyBuilder[F[_]] private (
     socketAddress: InetSocketAddress,
-    private val executionContext: ExecutionContext,
+    threadPool: ThreadPool,
```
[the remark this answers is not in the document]
rossabaker: Actually, that's probably not true: we've relinquished that servlet thread and it's uncertain what shifting back onto it would do to the server's internal bookkeeping.

In fact, blocking IO in the blocking servlet shouldn't shift at all, because the servlet thread will be waiting anyway.  The only time this shifting is useful at all is blocking IO in the async servlet, which itself is only useful on 3.0 -- not 2.5 (blocking everywhere) or 3.1 (non-blocking servlet IO).

Greetings from 2010.

rossabaker:
On Jetty or Tomcat, requests are allocated a thread from a pool.  Blocking 1 of n threads leaves (n - 1) threads for other requests.  In Blaze, if we block a thread, requests are assigned round-robin to `SelectorLoop`s.  Blocking 1 of n threads blocks 1/n requests.  I think the sensible thing is to remove the `ExeuctorService` for the servlet backends, using their native thread pools, and keep an `ExecutorService` for blaze.

The default `ExecutoinContext.global` in the blaze builder forces the creation of the global execution context, but this PR is big enough already.

rossabaker:
@igosuki Ugh, sorry. If you notice anything you like better in yours, I'd happily take a PR here. Especially if you had a cleaner solution to combining the Chunks. I feel that's particularly sloppy in this.

rossabaker:
@Igosuki @goedelsoup was also working on examples.  Perhaps you could compare notes and see if one is cleaner than the other.

I'll cherry-pick the async-http-client here, because that one wasn't done by anyone yet.

A rho PR would be great.  That project historically takes a while to keep up with http4s.

Thanks!

goedelsoup:
oh he's way ahead of me and the idea was basically the same. i think @Igosuki should go ahead and PR his changes and we should bring in @ChristopherDavenport to iron out what we want to market as idiomatic http4s.

Igosuki:
I'm using my forks with my production build first to see if everything
migrates fine.
The only problem I have is most of my services return HttpRoutes and the
Response Logger middleware expects a Kleisli of F instead of OptionT.

Le mer. 29 août 2018 à 16:58, Cory Parent <notifications@github.com> a
écrit :

rossabaker:
See #1978 for that logger problem: we want to mount `HttpApp` to the server instead of `HttpRoutes`.  As it stands, you need to promote your `HttpRoutes` to `HttpApp` to make them total so we know what we're logging, and then demote it back to `HttpRoutes` to mount it to a server.

blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClient.scala
```diff
@@ -21,11 +21,25 @@ object BlazeClient {
...
       onShutdown: F[Unit])(implicit F: Sync[F]): Client[F] =
-    Client(
+    makeClient(
+      manager,
+      responseHeaderTimeout = config.responseHeaderTimeout,
+      idleTimeout = config.idleTimeout,
+      requestTimeout = config.requestTimeout
+    )
+
+  private[blaze] def makeClient[F[_], A <: BlazeConnection[F]](
```
rossabaker: Still useful for testing.

blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala
```diff
@@ -0,0 +1,192 @@
...
+import cats.effect._
+import cats.implicits._
+import fs2.Stream
+import java.nio.channels.AsynchronousChannelGroup
+import javax.net.ssl.SSLContext
+import org.http4s.headers.{AgentProduct, `User-Agent`}
+import scala.concurrent.ExecutionContext
+import scala.concurrent.duration._
+
+sealed abstract class BlazeClientBuilder[F[_]] private (
```
rossabaker: Brings this up to the idiom for the other clients.

blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala
```diff
@@ -0,0 +1,192 @@
...
+        requestTimeout = requestTimeout,
+        executionContext = executionContext
+      )
+  }
+}
+
+object BlazeClientBuilder {
+  def apply[F[_]](
+      executionContext: ExecutionContext,
+      sslContext: Option[SSLContext] = Some(SSLContext.getDefault)): BlazeClientBuilder[F] =
```
rossabaker: We have to treat this one delicately because it can throw in grim circumstances.  See #1957.

blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala
```diff
@@ -0,0 +1,192 @@
...
+            idleTimeout = idleTimeout,
+            requestTimeout = requestTimeout
+        )))(_.shutdown)
+
+  def stream(implicit F: ConcurrentEffect[F]): Stream[F, Client[F]] =
+    Stream.resource(resource)
+
+  private def connectionManager(
+      implicit F: ConcurrentEffect[F]): F[ConnectionManager[F, BlazeConnection[F]]] = {
+    val http1: ConnectionBuilder[F, BlazeConnection[F]] = new Http1Support(
```
rossabaker: We've made this client very HTTP/1 centric with this.  What would HTTP/2 for blaze even look like?

blaze-client/src/main/scala/org/http4s/client/blaze/ParserMode.scala
```diff
@@ -0,0 +1,9 @@
+package org.http4s.client
+package blaze
+
+sealed abstract class ParserMode extends Product with Serializable
```
rossabaker: This is a good candidate to be pushed down to blaze.

blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala
```diff
@@ -0,0 +1,192 @@
...
+    val maxWaitQueueLimit: Int,
+    val maxConnectionsPerRequestKey: RequestKey => Int,
+    val sslContext: Option[SSLContext],
+    val checkEndpointIdentification: Boolean,
+    val maxResponseLineSize: Int,
+    val maxHeaderLength: Int,
+    val maxChunkSize: Int,
+    val parserMode: ParserMode,
+    val bufferSize: Int,
+    val executionContext: ExecutionContext,
```
rossabaker: I'd like to just use a `ContextShift` here, but blaze needs an `ExecutionContext`.  This is kind of leaking its `Future` implementation in a way.

testing/src/test/scala/org/http4s/Http4sSpec.scala
```diff
@@ -120,6 +121,21 @@ trait Http4sSpec
   def beStatus(status: Status): Matcher[Response[IO]] = { resp: Response[IO] =>
     (resp.status == status) -> s" doesn't have status $status"
   }
+
+  def withResource[A](r: Resource[IO, A])(fs: A => Fragments): Fragments =
```
rossabaker: Arguably belongs in specs2-cats.

Igosuki:
@rossabaker Ok, here's the pattern I'm using after your advice : 
- promote : ```
      corsed = CORS(service.routes, corsConfig)
      logged = LoggerMiddleware[F](logHeaders = true, logBody = true)(corsed.orNotFound)
```
- demote : ```
import cats.~>
def fk : F ~> OptionT[F, ?] = new FunctionK[F, OptionT[F, ?]] {
    override def apply[A](fa: F[A]): OptionT[F, A] = OptionT.liftF(fa)
  }
logged.mapK(fk)
```

Is that correct ?

rossabaker:
I think there's an `OptionT.liftK` for your `fk`, but that looks right.

Igosuki:
Ok sounds good thanks, I just did a deployment and it works fine (using my updated rho). Will submit a PR later.

rossabaker:
I just cherry-picked your commits, and will be pushing them here, so no need for PR on that.  PR on rho would still be highly appreciated.  Thanks!

Igosuki:
Understood, will do.

Le mer. 29 août 2018 à 20:41, Ross A. Baker <notifications@github.com> a
écrit :

testing/src/main/scala/org/http4s/testing/ArbitraryInstances.scala
```diff
@@ -713,9 +713,10 @@ trait ArbitraryInstances {
       var bytes: Vector[Byte] = null
       val readBytes = IO(bytes)
       F.runAsync(stream.compile.toVector) {
-        case Right(bs) => IO { bytes = bs }
-        case Left(t) => IO.raiseError(t)
-      } *> readBytes
+          case Right(bs) => IO { bytes = bs }
+          case Left(t) => IO.raiseError(t)
+        }
+        .toIO *> readBytes
```
jmcardon: what's the extra `.toIO` for? and why not just `F.toIO`

blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala
jmcardon: Either we do this, or we rely on some config somewhere.

blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala
jmcardon: In what situation does it throw that it shouldn't be a fatal uncaught error?

okhttp-client/src/main/scala/org/http4s/client/okhttp/OkHttpBuilder.scala
```diff
@@ -93,7 +92,7 @@ sealed abstract class OkHttpBuilder[F[_]] private (
     new Callback {
       override def onFailure(call: Call, e: IOException): Unit = {
         logger.info("onFailure")
-        cb(Left(e))
+        (cs.shift *> F.delay(cb(Left(e)))).runAsync(loggingAsyncCallback(logger)).unsafeRunSync()
```
jmcardon: What on earth...

blaze-client/src/main/scala/org/http4s/client/blaze/Http1Support.scala
```diff
@@ -6,62 +6,72 @@ import cats.effect._
...
+    maxHeaderLength: Int,
+    maxChunkSize: Int,
+    parserMode: ParserMode,
+    userAgent: Option[`User-Agent`]
+)(implicit F: Effect[F]) {
 
   // SSLContext.getDefault is effectful and can fail - don't force it until we have to.
-  private lazy val sslContext = config.sslContext.getOrElse(SSLContext.getDefault)
-  private val connectionManager = new ClientChannelFactory(config.bufferSize, config.group)
+  private lazy val sslContext = sslContextOption.getOrElse(SSLContext.getDefault)
```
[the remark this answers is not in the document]
rossabaker: That's exactly what we're doing.  See https://github.com/http4s/http4s/issues/1956 for why.

If `None` meant none -- that the client doesn't speak SSL -- then we could avoid that unpleasant situation and not abuse laziness.

blaze-client/src/test/scala/org/http4s/client/blaze/BlazeClientSpec.scala
```diff
@@ -0,0 +1,173 @@
...
+import cats.implicits._
+import javax.servlet.ServletOutputStream
+import javax.servlet.http.{HttpServlet, HttpServletRequest, HttpServletResponse}
+import org.http4s._
+import org.http4s.client.testroutes.GetRoutes
+import scala.concurrent.Await
+import scala.concurrent.duration._
+import scala.util.Random
+
+class BlazeClientSpec extends Http4sSpec {
```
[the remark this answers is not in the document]
rossabaker: It used to be named "PooledHttp1ClientSpec", a name of the constructor that was deprecated a long time ago.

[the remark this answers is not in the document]
jmcardon: :+1:

async-http-client/src/main/scala/org/http4s/client/asynchttpclient/AsyncHttpClient.scala
```diff
@@ -146,6 +139,7 @@ object AsyncHttpClient {
       }
   }
 
+  import cats.syntax._
```
kubukoz: is this used? There's a `cats.implicits._` import on top of this file

rossabaker:
Tests will hang on fs2-reactive-streams and fs2-jawn incompatibilities.

async-http-client/src/main/scala/org/http4s/client/asynchttpclient/AsyncHttpClient.scala
rossabaker: I think it might be to dance around the unused import warning, but we have a different idiom.

rossabaker:
Last commit was M4, for fs2's M5

rossabaker:
Tests are now green locally, with AHC disabled pending https://github.com/zainab-ali/fs2-reactive-streams/pull/54

PR #2000, opened 2018-08-16
OptionalMultiQueryParamDecoderMatcher accumulates

### LINKED ISSUE #1825: OptionalMultiQueryParamDecoderMatcher only keeps the last error
When using `OptionalMultiQueryParamDecoderMatcher`, the extracted parameter is of type `ValidatedNel`, so I had expected that it woudl accumulate all the validation errors if there's multiple parsing errors. However, only the last one is retained. The relevant code is there:

https://github.com/http4s/http4s/blob/87eb4b0adf2b56702158907d79bb4509781ace66/dsl/src/main/scala/org/http4s/dsl/impl/Path.scala#L280

A simple fix should to use this code instead: 

```
Some(Traverse[List].traverse(values.toList)(s => QueryParamDecoder[T].decode(QueryParameterValue(s))))
```

rossabaker:
Fixes #1825.


rossabaker:
Build failure was the failure to bump version.sbt in the 0.18.16 release.

PR #2001, opened 2018-08-17
Retry request without cleaning body

### LINKED ISSUE #1999: Client's Retry with empty body
I found strange code, is this bug or feature?
Why new attempt occurs with empty body in case of failure?

https://github.com/http4s/http4s/blob/9655331c743ca91326c94bd98730f2a3f382415a/client/src/main/scala/org/http4s/client/middleware/Retry.scala#L62

maxkorolev:
fix #1999

### DIFF
--- a/client/src/main/scala/org/http4s/client/middleware/Retry.scala
+++ b/client/src/main/scala/org/http4s/client/middleware/Retry.scala
@@ -64,9 +64,7 @@ object Retry {
         }
         .getOrElse(0L)
       val sleepDuration = headerDuration.seconds.max(duration)
-      scheduler.sleep_[F](sleepDuration).compile.drain *> prepareLoop(
-        req,
-        attempts + 1)
+      scheduler.sleep_[F](sleepDuration).compile.drain *> prepareLoop(req, attempts + 1)
     }
 
     client.copy(open = Kleisli(prepareLoop(_, 1)))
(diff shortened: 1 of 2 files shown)


jcranky:
Would this deserve a test?

rossabaker:
:+1: to the fix.

:+1: to a test.  @maxkorolev, would you be able to add that?  If not, someone else can pick it up.  Thanks!

maxkorolev:
I didn't get how to fix such build failure, it doesn't depend on my commits.
It seems like regular failure for these tests, I see a lot of Boom exceptions all over the place

rossabaker:
We test some error handling cases, and we haven't muted the loggers in tests, so we end up with some stack traces during a successful run.

The build error is [here](https://travis-ci.org/http4s/http4s/jobs/417131452#L4982).  We enforce scalafmt, so running `test:scalafmt` will make it pass.  `sbt ci` will run all the tests the build runs, but take longer and requires a hugo installation for the docs.

maxkorolev:
Done)

rossabaker:
I like the change, but I'd still like a test.  If @maxkorolev doesn't have time, I'd be happy to write one and append it to this PR.

maxkorolev:
Hm, what kind of new tests do you want me to write? Maybe add some effects to request?
That would be great if you wrote one) Thank you

jmcardon:
:+1: to a test

rossabaker:
I wish scalafmt on compile weren't prohibitively slow.

Issue #2004, opened 2018-08-17
empty path followed by query cannot be matched


fommil:
in version 0.18.16, this matcher fails

```scala
  private object Code extends QueryParamDecoderMatcher[String]("code")
  private val service: HttpService[Task] = HttpService[Task] {
    case GET -> Root / "" :? Code(code) => Ok(z"WE GOT ALL YUR CODEZ $code")
  }
```

to match against queries such as `http://localhost:39833/?code=...lotsofstuff...`

The workaround is to use a non-empty path if possible.

rossabaker:
Turns out the empty path can be matched:

```scala
    case GET -> Root :? I(start) =>
      Ok(s"start: $start")
```

is tested and works:

```scala
    "GET /?{start}" in {
      val response = serve(Request(GET, uri("/?start=1")))
      response.status must_== (Ok)
      response.as[String] must returnValue("start: 1")
    }
```

fommil:
hmm, that is strange. I was getting a compile error with that kind of code 


ah :point_up: [August 17, 2018 4:26 PM](https://gitter.im/http4s/http4s?at=5b76e93a070fa234ee77a468) I was using

```scala
case GET -> Root / :? Code(code) => 
```

so the stray `/` caused the compiler error!

Should we close or should `"" / ` work?

fommil:
btw code is here https://github.com/fommil/fpmortals/blob/master/example/src/main/scala/fommil/http/oauth2/interpreters/BlazeUserInteraction.scala

rossabaker:
`//` starts a "protocol-relative" URL, which means the "//" is _not_ part of the path.  Testing it with `//example.com//?start=1` works as expected.  There's no bug here, but I will improve dsl.md to clarify.

PR #2008, opened 2018-08-19
Improve documentation of path info matching

rossabaker:
Better explanation of `Root` and `/` on the heels of #2004.

### DIFF
--- a/docs/src/main/tut/dsl.md
+++ b/docs/src/main/tut/dsl.md
@@ -284,35 +284,70 @@ values. By extension, you can also `match/case` it with different possible
 destructurings. To build these different extractors, you can make use of the
 DSL.
 
-Most often, you extract the `Request` into a HTTP `Method` (verb) and the path,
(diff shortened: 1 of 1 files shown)


docs/src/main/tut/dsl.md
```diff
@@ -284,35 +284,70 @@ values. By extension, you can also `match/case` it with different possible
 destructurings. To build these different extractors, you can make use of the
 DSL.
 
-Most often, you extract the `Request` into a HTTP `Method` (verb) and the path,
-via the `->` object. On the left side, you'll have the HTTP `Method`, on the
-other side the path. Naturally, `_` is a valid matcher too, so any call to
-`/api` can be blocked, regardless of `Method`:
+### The `~>` object
```
cquiroz: Why `~>` instead of `->`?

rossabaker: Mistake.

Issue #2015, opened 2018-08-19
Eliminate `Status.register`


rossabaker:
This was inherited from our original fork of Spray, many years ago.  It is global, mutable state.  It's fine as an internal mechanism for resolving values to a single reference, but it should not be public API.

calvinbrown085:
Hey Ross, more than happy to pick this up. https://github.com/http4s/http4s/blob/b9a28ddcd335cf5dee60a9f7d3be5bdc12523179/core/src/main/scala/org/http4s/Status.scala#L103

is this as simple as making this `private def` ?

rossabaker:
I think this solves the problem, but we might simplify the implementation a bit more.  We don't really need an `AtomicReferenceArray` anymore, since there can no longer be concurrent additions and updates.  Just an `Array` should suffice.

Either the `AtomicReferenceArray` or `Array` implementations trade space (about 50 of the 600 spots are used) for time (indexed lookup vs. hash lookup).  I think this is a fine trade, but am open to being shown I'm wrong.

ghostroad:
I'm new to http4s. I've been taking a look at this issue, and my first thought was "These should be case objects!" which I realized wouldn't work because they must be created eagerly. 

From reading the code, it seems we want to both maintain a registry of "standard" `Status`es and have the ability to create new, unregistered, ad-hoc ones (e.g., using `Status.fromInt` or `withReason`). Is this the desired behavior? If it is, then something like a private `register` method is needed, for explicitly registering the standard statuses. If not, then registration could be moved into the constructor, making it automatic.

It does seem like an `Array[Some[Status]]` would be a cleaner replacement for the existing registry.

I was also wondering about this spec:

https://github.com/http4s/http4s/blob/b9a28ddcd335cf5dee60a9f7d3be5bdc12523179/tests/src/test/scala/org/http4s/StatusSpec.scala#L43-L47

It seems like an odd thing to check for, since equality for `Status`es only cares about the status code. It would be helpful to have specs that describe this behavior more fully. 

@calvinbrown085, would it be alright if I took a look at this on a new, more up-to-date branch?

calvinbrown085:
@ghostroad Feel free!

jmcardon:
@ghostroad they don't need to be created eagerly, if the array initialization is a static block.

[...]

PR #2020, opened 2018-08-19
Making the Status.register method non-public, only for internal use

calvinbrown085:
https://github.com/http4s/http4s/issues/2015

### DIFF
--- a/core/src/main/scala/org/http4s/Status.scala
+++ b/core/src/main/scala/org/http4s/Status.scala
@@ -100,7 +100,7 @@ object Status {
       status <- Option(registry.get(code)).map(_.right.get)
     } yield status
 
-  def register(status: Status): status.type = {
+  private def register(status: Status): status.type = {
     registry.set(status.code, Right(status))
     status
   }


rossabaker:
What idiom do you have in mind to get them all cached?  You cna do it without an internal mutable registry, but the obvious ways of doing it then mention each status twice -- and risk omitting one from the cache.

jmcardon:
quips:
```scala
  private val registry =
    new AtomicReferenceArray[Right[Nothing, Status]](600)
```

- this is useless because [we call `.right.get` every time](https://github.com/http4s/http4s/blob/d9b9e54970c3bd9e46e2fb81d003b7a915d4db71/core/src/main/scala/org/http4s/Status.scala#L100). Might as well be `AtomicReferenceArray[Status]`.
- is there any particular reason why it's an `AtomicReferenceArray` in the first place over just an array, if we should never mutate it anyway outside the first initialization block? We ruled we're going to stop people from registering custom status codes anyway.
- If we had an array, the [`registered`](https://github.com/http4s/http4s/blob/d9b9e54970c3bd9e46e2fb81d003b7a915d4db71/core/src/main/scala/org/http4s/Status.scala#L97) function could just be `.filterNot(_ eq null).iterator`

calvinbrown085:
Hi guys, i'm kinda lost here :/ Seems like @jmcardon Want's to maybe get rid of the `register` ?

rossabaker:
* The `.right.get` is so we don't need to rewrap everything in `Right` in `fromInt`.
* Agreed that `AtomicReferenceArray` was questionable before and overkill now.
* I'm not worried about `val registry = { // initializer` being longer.  I'm worried about how we register every value we also store as a constant.  I think we're going to be mentioning everything twice:

```scala
val Ok = mkStatus(200, "OK")
val ImATeapot = mkStatus(419, "I'm a teapot")
val InternalServerError = mkStatus(500, "Internal Server Error")

val register = { 
  List(Ok, InternalServerError).map(/* oh no we forgot 419 */) 
}
```

rossabaker:
Oops, our comments crossed over.  I think he was concerned about the internal mutation.  I'm concerned about how we do it immutably without enumerating every variable wherever we initialize the registry.

It can also be done with reflection, and we did that once with header parsers, but I think we can all agree that's worse.

ChristopherDavenport:
I'm heavily against reflection. Although I think we all know that we should avoid that, wanted to make sure its mentioned since it was brought up.

PR #2023, opened 2018-08-20
Adding a check for existing vary header before overwriting with CORS …

### LINKED ISSUE #2003: CORS headers overwrites existing Vary headers
My application returns the Vary:Accept header, but when a client makes a request with an Origin header, the http4s CORS object overwrites with Vary:Origin,Access-Control-Request-Methods by using putHeaders:

```
def corsHeaders(origin: String, acrm: String, isPreflight: Boolean)(
          resp: Response[F]): Response[F] = {
        val methodBasedHeader =
          if (isPreflight)
            config.allowedHeaders.map(headerFromStrings("Access-Control-Allow-Headers", _))
          else config.exposedHeaders.map(headerFromStrings("Access-Control-Expose-Headers", _))
        methodBasedHeader
          .fold(resp)(h => resp.putHeaders(h))
          .putHeaders(
            Header("Vary", "Origin,Access-Control-Request-Methods"),
            Header("Access-Control-Allow-Credentials", config.allowCredentials.toString()),
            Header(
              "Access-Control-Allow-Methods",
              config.allowedMethods.fold(acrm)(_.mkString("", ", ", ""))),
            Header("Access-Control-Allow-Origin", origin),
            Header("Access-Control-Max-Age", config.maxAge.toString)
          )
      }
```

I'd like to continue to return Vary: Accept as well as Origin and Access-Control-Request-Methods - are there any issues with this?

yasuba:
…vary header

fixes #2003

### DIFF
--- a/server/src/main/scala/org/http4s/server/middleware/CORS.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/CORS.scala
@@ -30,6 +30,8 @@ final case class CORSConfig(
 object CORS {
   private[CORS] val logger = getLogger
 
+  val defaultVaryHeader = Header("Vary", "Origin,Access-Control-Request-Method")
+
   def DefaultCORSConfig =
     CORSConfig(anyOrigin = true, allowCredentials = true, maxAge = 1.day.toSeconds)
 
@@ -55,19 +57,18 @@ object CORS {
         else
           config.exposedHeaders.map(headerFromStrings("Access-Control-Expose-Headers", _))
 
-      def checkVaryHeader(response: Response[G]): Option[Header] = {
-        val maybeHeader = response.headers.get(CaseInsensitiveString("Vary"))
-        if (maybeHeader.isEmpty) Option(Header("Vary", "Origin,Access-Control-Request-Method"))
-        else maybeHeader
-      }
+      def varyHeader(response: Response[G]): Response[G] =
+        response.headers.get(CaseInsensitiveString("Vary")) match {
+          case None => response.putHeaders(defaultVaryHeader)
+          case _ => response
+        }
 
       def corsHeaders(origin: String, acrm: String, isPreflight: Boolean)(
           resp: Response[G]): Response[G] = {
         val withMethodBasedHeader = methodBasedHeader(isPreflight)
           .fold(resp)(h => resp.putHeaders(h))
 
-        checkVaryHeader(withMethodBasedHeader)
-          .fold(withMethodBasedHeader)(h => withMethodBasedHeader.putHeaders(h))
+        varyHeader(withMethodBasedHeader)
           .putHeaders(
             Header("Access-Control-Allow-Credentials", config.allowCredentials.toString()),
             Header(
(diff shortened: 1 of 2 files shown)


server/src/main/scala/org/http4s/server/middleware/CORS.scala
```diff
@@ -54,19 +55,28 @@ object CORS {
         else
           config.exposedHeaders.map(headerFromStrings("Access-Control-Expose-Headers", _))
 
+      def checkVaryHeader(response: Response[G]): Option[Header] = {
```
jmcardon: I think this function can be `Response[G] => Response[G]`, saving you a fold down the line

If you look down below:

```scala
checkVaryHeader(withMethodBasedHeader)
          .fold(withMethodBasedHeader)(h => withMethodBasedHeader.putHeaders(h))
```

you could simplify this into:

```scala
def varyHeader(response: Response[G]): Response[G] =  {
  // Match avoids a fold higher order call, but you can just do this with fold if you don't want
  response.headers.get(CaseInsensitiveString("Vary")) match {
    case None => 
      response.putHeaders(Header("Vary", "Origin,Access-Control-Request-Method"))
     case _ => 
       response
  }
}
```

and then, down the line:

```scala
varyHeader(withMethodBasedHeader)
  .putHeaders(..
```

server/src/main/scala/org/http4s/server/middleware/CORS.scala
```diff
@@ -54,19 +55,28 @@ object CORS {
...
+        val maybeHeader = response.headers.get(CaseInsensitiveString("Vary"))
+        if (maybeHeader.isEmpty) Option(Header("Vary", "Origin,Access-Control-Request-Method"))
+        else maybeHeader
```
rossabaker: This could be `maybeHeader.getOrElse(Header(...))`.  But I like @jmcardon's idea.

server/src/main/scala/org/http4s/server/middleware/CORS.scala
```diff
@@ -54,19 +55,28 @@ object CORS {
         else
           config.exposedHeaders.map(headerFromStrings("Access-Control-Expose-Headers", _))
 
+      def checkVaryHeader(response: Response[G]): Option[Header] = {
+        val maybeHeader = response.headers.get(CaseInsensitiveString("Vary"))
+        if (maybeHeader.isEmpty) Option(Header("Vary", "Origin,Access-Control-Request-Method"))
```
rossabaker: Microoptimization: this default header could be stored in a val and shared by all requests.

jmcardon: A microoptimization suggested by Ross himself?!

Are my repeated pedantic comments on microoptimizations finally making a dent?

server/src/main/scala/org/http4s/server/middleware/CORS.scala
yasuba: I'd like to polish the PR a bit - thanks for the comments

jmcardon:
Thanks for the contribution @yasuba

PR #2024, opened 2018-08-20
Add header echoing server middleware

keirlawson:
A server middleware that echoes selected headers from request to response

### DIFF
--- a/server/src/main/scala/org/http4s/server/middleware/HeaderEcho.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/HeaderEcho.scala
@@ -1,0 +1,24 @@
+package org.http4s.server.middleware
+
+import cats.Functor
+import cats.data.Kleisli
+import org.http4s._
+import cats.implicits._
+import org.http4s.util.CaseInsensitiveString
+
+object HeaderEcho {
+
+  /**
+    * Simple server middleware that adds selected headers present on the request to the response.
+    *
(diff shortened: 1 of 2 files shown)


server/src/main/scala/org/http4s/server/middleware/HeaderEcho.scala
```diff
@@ -0,0 +1,24 @@
...
+  private def isOneOf(headerKeys: Seq[HeaderKey])(header: Header): Boolean =
+    headerKeys.exists(_.matchHeader(header).isDefined)
+
+  /**
+    * Simple server middleware that adds selected headers present on the request to the response.
+    *
+    * @param headerKeys the keys corresponding to the headers to echo on the response
+    * @param http [[HttpRoutes]] to transform
+    */
+  def apply[F[_]: Functor](headerKeys: HeaderKey*)(http: HttpRoutes[F]): HttpRoutes[F] =
```
rossabaker: This requires a `HeaderKey`, while `Logger` uses a `redactHeadersWhen: CaseInsensitiveString => Boolean`.  There are pros and cons to both of those, but we should probably be consistent.

I might like the idea of a predicate.  Then you could do things like `_.startsWith("X-MyCompany-")`

I think this could be generic in `Http[F, G]` instead of `HttpRoutes[F]` and then also work for `HttpApp` and future routing DSLs.

keirlawson:
Done

Sokphen (commented):
[]()

PR #2026, opened 2018-08-21
CSRF Overhaul

### LINKED ISSUE #1989: CSRF to handle url parameter/manually fed token
So while creating a websocket service, you want to prevent [CSWSH](http://www.christian-schneider.net/CrossSiteWebSocketHijacking.html).

Solution 1 is checking origin, and that can be done in userland no problem.

Soultion 2: Our CSRF service would work for this, _except that_ clients have to use javascript to embed the locally cached token into the request... which isn't possible with the native client websocket interface.

The solution to this, of course, is that sending the token in a url parameter or as a `/<token>` part of the url itself (especially for WSS) in a similar way to CSRF, is effective enough when the tokens themselves are generated securely with enough entropy.

I already have a sketched out solution in a private repo (this is using `TSecCSRF`, but same thing for ours in this repo)

```scala
  private[this] def checkCSRF(
      token: String,
      r: Request[IO],
      f: => IO[Response[IO]]
  ): IO[Response[IO]] =
    (for {
      c    <- cookieFromRequest(csrf.cookieName, r)
      raw1 <- csrf.extractRaw(CSRFToken(c.content))
      raw2 <- csrf.extractRaw(CSRFToken(token))
      res <- if (csrf.isEqual(raw1, raw2))
        OptionT.liftF(f)
      else
        OptionT.none[IO, Response[IO]]
      newToken <- OptionT.liftF(csrf.signToken(raw1))
    } yield res.addCookie(Cookie(name = csrf.cookieName, content = newToken)))
      .getOrElse(Response[IO](Unauthorized))
```

All is left is for either me or someone that beats me to it to add it to the `CSRF` class.

jmcardon:
closes #1989 

This changes our CSRF middleware significantly. In particular:
- CSRF tokens are now newtyped... because passing around `String` everywhere was honestly giving me a goddamn headache
- CSRF token signatures are encoded in hexadecimal strings, making it url-safe. For example: In the case of websockets, clients (if using a browser) often have no way of embedding the csrf token in the header, forcing sending in a query parameter or as part of the path.
- CSRF Internal functions are now not private, as I've had to define my own combinators with them before. Just in case of a very special use case for whatever reason (i.e csrf token in multipart payload), the functions are exposed to be able to construct csrf actions from a specific endpoint.
- CSRF now takes a conditional `Request[G] => Boolean`, which is meant to be used to check `Origin` headers in particular, [as stated by OWASP](https://www.owasp.org/index.php/Cross-Site_Request_Forgery_(CSRF)_Prevention_Cheat_Sheet#Verifying_Same_Origin_with_Standard_Headers). However, given that behind a proxy, neither `Origin` nor `Referrer` may be present, the conditional is left open-ended (with a few default constructors), for the sake of custom use cases
- Default error status is `Forbidden`, given that `Unauthorized` (despite being correct in name), [for whatever silly reason is for _Authentication_ Errors](https://stackoverflow.com/questions/3297048/403-forbidden-vs-401-unauthorized-http-responses) (lol http)

### DIFF
--- a/.idea/codeStyles/Project.xml
+++ b/.idea/codeStyles/Project.xml
@@ -7,7 +7,6 @@
         </array>
       </option>
       <option name="sortAsScalastyle" value="true" />
-      <option name="USE_SCALAFMT_FORMATTER" value="true" />
     </ScalaCodeStyleSettings>
     <XML>
       <option name="XML_LEGACY_SETTINGS_IMPORTED" value="true" />
--- a/core/src/main/scala/org/http4s/internal/package.scala
+++ b/core/src/main/scala/org/http4s/internal/package.scala
@@ -4,6 +4,7 @@ import cats.effect.{Async, Effect, IO, Timer}
 import cats.implicits._
 import scala.concurrent.ExecutionContext
 import org.log4s.Logger
+import scala.util.control.NoStackTrace
 
 package object internal {
   // Like fs2.async.unsafeRunAsync before 1.0.  Convenient for when we
@@ -23,4 +24,78 @@ package object internal {
       case Left(e) => IO(logger.error(e)("Error in asynchronous callback"))
       case Right(_) => IO.unit
     }
+
+  /** Hex encoding digits. Adapted from apache commons Hex.encodeHex **/
+  private val Digits: Array[Char] =
+    Array('0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'A', 'B', 'C', 'D', 'E', 'F')
+
+  /** Encode a byte Array into a hexadecimal string
+    *
+    * @param data the array
+    * @return a hexadecimal encoded string
+    */
+  private[http4s] final def encodeHexString(data: Array[Byte]): String =
+    new String(encodeHex(data))
+
+  /** Encode a string to a Hexadecimal string representation
+    * Adapted from apache commons Hex.encodeHex
+    */
+  private[http4s] final def encodeHex(data: Array[Byte]): Array[Char] = {
+    val l = data.length
+    val out = new Array[Char](l << 1)
+    // two characters form the hex value.
+    var i = 0
+    var j = 0
+    while (i < l) {
+      out(j) = Digits((0xF0 & data(i)) >>> 4)
+      j += 1
+      out(j) = Digits(0x0F & data(i))
+      j += 1
+      i += 1
+    }
+    out
+  }
+
+  private[http4s] final def decodeHexString(data: String): Option[Array[Byte]] =
+    decodeHex(data.toCharArray)
+
+  private object HexDecodeException extends Exception with NoStackTrace
+
+  /** Dirty, optimized hex decoding based off of apache
+    * common hex decoding, ported over to scala
+    *
+    * @param data
+    * @return
+    */
+  private[http4s] final def decodeHex(data: Array[Char]): Option[Array[Byte]] = {
+    def toDigit(ch: Char): Int = {
+      val digit = Character.digit(ch, 16)
+      if (digit == -1)
+        throw HexDecodeException
+      else
+        digit
+    }
+
+    val len = data.length
+    if ((len & 0x01) != 0) None
+    val out = new Array[Byte](len >> 1)
+    var f: Int = -1
+    // two characters form the hex value.
+    try {
+      var i = 0
+      var j = 0
+      while (j < len) {
+        f = toDigit(data(j)) << 4
+        j += 1
+        f = f | toDigit(data(j))
+        j += 1
+        out(i) = (f & 0xFF).toByte
+
+        i += 1
+      }
+      Some(out)
+    } catch {
+      case HexDecodeException => None
+    }
+  }
 }
--- a/core/src/main/scala/org/http4s/util/package.scala
+++ b/core/src/main/scala/org/http4s/util/package.scala
@@ -1,11 +1,10 @@
 package org.http4s
 
 import java.nio.charset.StandardCharsets
-
 import fs2._
 import java.nio.{ByteBuffer, CharBuffer}
-
 import scala.concurrent.ExecutionContextExecutor
+import scala.util.control.NoStackTrace
 
 package object util {
   def decode[F[_]](charset: Charset): Pipe[F, Byte, String] = {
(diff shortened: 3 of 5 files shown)


.idea/codeStyles/Project.xml
```diff
@@ -7,7 +7,6 @@
         </array>
       </option>
       <option name="sortAsScalastyle" value="true" />
-      <option name="USE_SCALAFMT_FORMATTER" value="true" />
```
jmcardon: My editor does this EVERY SINGLE TIME AHHHHH

It must be my IJ version...

server/src/main/scala/org/http4s/server/middleware/CSRF.scala
```diff
@@ -161,28 +207,101 @@ final class CSRF[F[_], G[_]] private[middleware] (
...
+      port: Option[Int]): Boolean =
+    r.headers
+      .get(Origin)
+      .flatMap(o => Uri.fromString(o.value).toOption)
+      .exists(u => u.host.exists(_.value == host) && u.scheme.contains(sc) && u.port == port) || r.headers
+      .get(Referer)
+      .exists(u =>
+        u.uri.host.exists(_.value == host) && u.uri.scheme.contains(sc) && u.uri.port == port)
+
+  def proxyOriginCheck[F[_]](r: Request[F], host: Host, xff: `X-Forwarded-For`): Boolean =
```
jmcardon: Probably could use a convenience constructor?

jmcardon:
one big note on this PR: Should CSRF actually have two parameters?

In reality, I sort of don't think so. In particular, there's very liberal use of `OptionT` in `CSRF`. If `F` is `OptionT[F`, we end up with some `OptionT[OptionT[F` for no reason.

core/src/main/scala/org/http4s/util/package.scala
```diff
@@ -61,6 +60,14 @@ package object util {
   private val Digits: Array[Char] =
     Array('0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'A', 'B', 'C', 'D', 'E', 'F')
 
+  /** Encode a byte Array into a hexadecimal string
+    *
+    * @param data the array
+    * @return a hexadecimal encoded string
+    */
+  def encodeHexString(data: Array[Byte]): String =
```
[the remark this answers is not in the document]
jmcardon: Fair compromise, I forgot internal was a thing

server/src/main/scala/org/http4s/server/middleware/CSRF.scala
```diff
@@ -161,28 +207,101 @@ final class CSRF[F[_], G[_]] private[middleware] (
 
 object CSRF {
 
+  //Newtype hax
+  type CSRFToken <: String
```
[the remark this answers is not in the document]
jmcardon: IF you want the opaque encoding I can do it, I'll just do anything to stop passing around `String`

server/src/main/scala/org/http4s/server/middleware/CSRF.scala
jmcardon: I mean a way to create a `CSRF` without forcing the person to use `apply` and pass this around

server/src/main/scala/org/http4s/server/middleware/CSRF.scala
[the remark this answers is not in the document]
rossabaker: Agreed on the motivation, and I prefer the opaque solution.

server/src/main/scala/org/http4s/server/middleware/CSRF.scala
```diff
@@ -52,83 +53,128 @@ final class CSRF[F[_], G[_]] private[middleware] (
...
+    *
+    * @param token the signed csrf token
+    * @return
+    */
+  def createResponseCookie(token: CSRFToken): ResponseCookie =
+    ResponseCookie(name = cookieName, content = token, httpOnly = true, secure = self.secure)
+
+  /** Extract a csrftoken, if present, from the request,
+    * then generate a new token signature
+    * @return newly refreshed toen
```
rossabaker: token

core/src/main/scala/org/http4s/internal/package.scala
```diff
@@ -23,4 +24,78 @@ package object internal {
...
+    * @param data the array
+    * @return a hexadecimal encoded string
+    */
+  private[http4s] final def encodeHexString(data: Array[Byte]): String =
+    new String(encodeHex(data))
+
+  /** Encode a string to a Hexadecimal string representation
+    * Adapted from apache commons Hex.encodeHex
+    */
+  private[http4s] final def encodeHex(data: Array[Byte]): Array[Char] = {
```
ChristopherDavenport: I assume this is something people learn when they go to school.

server/src/main/scala/org/http4s/server/middleware/CSRF.scala
```diff
@@ -195,63 +231,67 @@ final class CSRF[F[_], G[_]] private[middleware] (
...
+    r.addCookie(createResponseCookie(token))
 
-  /** Middleware to embed a csrf token into routes that do not have one. **/
-  def withNewToken: Middleware[F, Request[G], Response[G], Request[G], Response[G]] =
-    _.andThen(embedNew _)
+  def embedInRequestCookie(r: Request[G], token: CSRFToken): Request[G] =
+    r.addCookie(createRequestCookie(token))
 
+  /** Embed a token into a response **/
+  def embedNewInCookie[M[_]: Sync](res: Response[G]): M[Response[G]] =
```
rossabaker: `embedNewInResponseCookie`, for consistency?

jmcardon: Will fix

core/src/main/scala/org/http4s/internal/package.scala
jmcardon: I've written this code more than once unfortunately to avoid pulling in a dependency just for this.

server/src/main/scala/org/http4s/server/middleware/CSRF.scala
```diff
@@ -49,86 +52,165 @@ import org.http4s.util.{CaseInsensitiveString, encodeHex}
...
     }
 
-  /** To be only used on safe methods: if the method is safe (i.e doesn't modify data),
-    * embed a new token if not present, or regenerate the current one to mitigate
-    * BREACH
+  /** To be only used on safe methods: if the method is safe (i.e doesn't modify data)
+    * and a token is present, validate and regenerate it for BREACH to be impractical
     */
-  private[middleware] def validateOrEmbed(r: Request[G], http: Http[F, G]): F[Response[G]] =
+  private[middleware] def validate(r: Request[G], response: F[Response[G]])(
```
ChristopherDavenport: Odd to take a value in a context, why does this need to be `F[Response[G]]` rather than just a `Response[G]` especially as you just evaluate it inside the comprehension in `flatMap` on both sides of the `match`

jmcardon: It can still fail. It's the reason from `F.fromEither(extractRaw(c.content))`

PR #2030, opened 2018-08-22
Restrict MethodNotAllowed constructor in Response Generator DSL.

### LINKED ISSUE #2028: Restrict the MethodNotAllowed constructor.
#### The Problem

The `http4s` DSL allows us to write an `HttpService` that, for a certain path, only accepts certain methods and replies with the status code `MethodNotAllowed` to others. For example: 
```Scala
val route = HttpService {
  case  GET -> /ping  => Ok("pong")
  case  _      -> /ping   => MethodNotAllowed()
}
```
The `MethodNotAllowed()` response would generate a response with that status code, but without any headers. However, [Section 10.4.6 of RFC 2616](https://tools.ietf.org/html/rfc2616#section-10.4.6), concerning the status `405 MethodNotAllowed`, indicates that: 

> The response MUST include an `Allow` header containing a list of valid methods for the requested   resource.

Therefore, the response in the second case is not compliant with the RFC. We can write an RFC-compliant HttpService by using the header: 
```Scala
import org.http4s.headers.Allow
val route = HttpService {
  case  GET -> /ping  => Ok("pong")
  case  _      -> /ping   => MethodNotAllowed( Allow(GET) )
}
```
However, the first code snippet gives a non-compliant route. There already are cases in the DSL, in which a response with a certain status is not allowed unless the required data is provided. For example, the `Unauthorized()` requires the `WWW-Authenticate` header.  

#### The solution

Modify the `http4s` DSL, to disallow creating a `MethodNotAllowed` response unless an `Allow` header with a non-empty list of parameters is provided.

diesalbla:
According to [Section 10.4.6 of the RFC 2616](https://tools.ietf.org/html/rfc2616#section-10.4.6), about the HTTP protocol, when using the `405 MethodNotAllowed` status code,

> The response MUST include an Allow header containing a list of valid
>  methods for the requested resource.

We thus restrict the DSL constructions for the MethodNotAllowed status. We add a trait `AllowResponseGenerator`, mostly copied and adapted from the `WwwAuthenticateResponseGenerator` trait.

Resolves #2028.

### DIFF
--- a/dsl/src/main/scala/org/http4s/dsl/impl/ResponseGenerator.scala
+++ b/dsl/src/main/scala/org/http4s/dsl/impl/ResponseGenerator.scala
@@ -120,6 +120,32 @@ trait WwwAuthenticateResponseGenerator[F[_]] extends Any with ResponseGenerator 
 }
 
 /** Helper for the generation of a [[org.http4s.Response]] which must contain
+  * an Allow header and may contain a body.
+  *
+  * A 405 status MUST contain an `Allow` header, which
+  * distinguishes this from other `ResponseGenerator`s.
+  */
+trait AllowResponseGenerator[F[_]] extends Any with ResponseGenerator {
+
+  def apply(allow: Allow, headers: Header*)(implicit F: Applicative[F]): F[Response[F]] =
+    F.pure(Response(status, headers = Headers(`Content-Length`.zero +: allow +: headers.toList)))
+
+  def apply[A](allow: Allow, body: A, headers: Header*)(
+      implicit F: Monad[F],
+      w: EntityEncoder[F, A]): F[Response[F]] = {
+    val h = w.headers ++ Headers(allow +: headers.toList)
+    val entity = w.toEntity(body)
+    val newHeaders = entity.length
+      .map { l =>
+        `Content-Length`.fromLong(l).fold(_ => h, c => h.put(c))
+      }
+      .getOrElse(h)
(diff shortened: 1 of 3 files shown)


dsl/src/main/scala/org/http4s/dsl/impl/ResponseGenerator.scala
```diff
@@ -119,6 +119,32 @@ trait WwwAuthenticateResponseGenerator[F[_]] extends Any with ResponseGenerator
...
 
+/** Helper for the generation of a [[org.http4s.Response]] which must contain
+  * an Allow header and may contain a body.
+  *
+  * A 405 status MUST contain an `Allow` header, which
+  * distinguishes this from other `ResponseGenerator`s.
+  */
+trait AllowResponseGenerator[F[_]] extends Any with ResponseGenerator {
+
+  def apply(allow: Allow, headers: Header*)(implicit F: Applicative[F]): F[Response[F]] =
```
jcranky: is an empty list/array valid here?

diesalbla: What do you mean by valid? Do you mean valid with respect to the Http RFC? 

If you meant whether the method can be applied with no `Header`, except for the `Allow` one, then you can see in the changes to `ExampleService` below that this is the case.

jcranky: nvm, I misunderstood how this is supposed to work :)

PR #2032, opened 2018-08-24
Topic/status.register 2015

ghostroad:
This is intended to address issue #2015 which was initially about making `Status.register` inaccessible, and gradually grew to encompass cleaning up the implementation of the registry and even changing the behavior of the class as described in 1) and 2) below:

1) `Status.fromInt` and `Status.fromIntAndReason` now both fail if supplied a nonstandard code.

2) `Status.fromIntAndReason` succeeds if supplied a standard code and nonstandard reason. It returns a new (uncached) Status with the standard code and the nonstandard reason. This is the same behavior as before.

3) The `Status` registry is now an `Array[Option[Status]]`. There are no `_.right.get`s any more. The methods pertaining to the registry are now collected in a private inner object called Registry. It would be nice to have been able to eliminate `Status.registered` but it is used to generate ScalaCheck instances. Perhaps it should be renamed to something like `Status.all`.

4) There was a subtle bug in `StatusSpec`:

https://github.com/http4s/http4s/blob/a53dbb1747b69cbe67e637536f73d7d369d62896/tests/src/test/scala/org/http4s/StatusSpec.scala#L9-L20

The first forAll check was completely being ignored here (not being the last expression in the block, it wasn’t returned, so it was never evaluated). This has been fixed by breaking out the assertions into separate tests.

5) Specs have been added for all the new behavior. The old behaviors are (hopefully) better highlighted as well.

### DIFF
--- a/core/src/main/scala/org/http4s/Status.scala
+++ b/core/src/main/scala/org/http4s/Status.scala
@@ -77,51 +77,40 @@ object Status {
     val ServerError = Status.Informational
   }
 
-  val MinCode = 100
-  val MaxCode = 599
+  private[http4s] val MinCode = 100
+  private[http4s] val MaxCode = 599
 
-  def fromInt(code: Int): ParseResult[Status] = lookup(code) match {
-    case None => parseAsStatus(code)
-    case Some(status) => Right(status)
-  }
+  //TODO eliminate duplication
+  def fromInt(code: Int): ParseResult[Status] = if (isInRange(code)) {
+      ParseResult.success(lookup(code).getOrElse(Status(code, "No reason provided")))
+    } else ParseResult.fail("Invalid code", s"$code is not between $MinCode and $MaxCode.")
 
-  def fromIntAndReason(code: Int, reason: String): ParseResult[Status] =
-    lookup(code, reason) match {
-      case None => parseAsStatus(code, reason)
-      case Some(status) => Right(status)
-    }
+  def fromIntAndReason(code: Int, reason: String): ParseResult[Status] =  if (isInRange(code)) {
+      ParseResult.success(lookup(code, reason).getOrElse((Status(code, reason))))
+  } else ParseResult.fail("Invalid code", s"$code is not between $MinCode and $MaxCode.")
 
-  def registered: Iterable[Status] = all
 
-  private def parseAsStatus(code: Int, reason: String = ""): ParseResult[Status] =
-    if (isStandard(code)) ParseResult.success(Status(code, reason))
-    else
-      ParseResult.fail("Invalid status", s"$code is not a valid response code.")
+  private[http4s] def registered: List[Status] = all
+
+  private def isInRange(code: Int) =
+    code >= MinCode && code <= MaxCode
+
 
   private object Registry {
-    private val registry = Array.fill[Option[Status]](MaxCode + 1)(None)
+    private val registry: Array[Either[String, Status]] = Array.fill[Either[String, Status]](MaxCode + 1){Left("unregistered")}
 
-    def lookup(code: Int): Option[Status] = if (isInRange(code)) registry(code) else None
+    def lookup(code: Int): Either[String, Status] = registry(code)
 
-    def lookup(code: Int, reason: String): Option[Status] =
-      for { status <- lookup(code) if status.reason == reason } yield status
-
-    def isStandard(code: Int): Boolean = isInRange(code) && registry(code).isDefined
+    def lookup(code: Int, reason: String): Either[String, Status] = {
+      lookup(code).filterOrElse[String]( { _.reason == reason}, "Reason did not match")
+    }
 
     def register(status: Status): Status = {
-      registry(status.code) = Some(status)
+      registry(status.code) = Right(status)
       status
     }
 
-    def all: Iterable[Status] =
-      for {
-        code <- MinCode to MaxCode
-        status <- lookup(code)
-      } yield status
-
-    private def isInRange(code: Int) =
-      code >= MinCode && code <= MaxCode
-
+    def all: List[Status] = registry.filter(_.isRight).map(_.right.get).toList
   }
 
   /**
--- a/core/src/main/scala/org/http4s/Status.scala
+++ b/core/src/main/scala/org/http4s/Status.scala
@@ -14,8 +14,10 @@ import org.http4s.util.Renderable
   * @see [http://tools.ietf.org/html/rfc7231#section-6 RFC7231, Section 6]
   * @see [http://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml IANA Status Code Registry]
   */
-sealed abstract case class Status private (code: Int)(val reason: String, val isEntityAllowed: Boolean)
-  extends Ordered[Status]
+sealed abstract case class Status private (code: Int)(
+    val reason: String,
+    val isEntityAllowed: Boolean)
+    extends Ordered[Status]
     with Renderable {
   // scalastyle:off magic.number
   val responseClass: ResponseClass =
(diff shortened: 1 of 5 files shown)


core/src/main/scala/org/http4s/Status.scala
```diff
@@ -76,102 +77,123 @@ object Status {
...
-  // scalastyle:on magic.number
+  def registered: Iterable[Status] = all
+
+  private def parseAsStatus(code: Int, reason: String = ""): ParseResult[Status] =
+    if (isStandard(code)) ParseResult.success(Status(code, reason))
+    else
+      ParseResult.fail("Invalid status", s"$code is not a valid response code.")
+
+  private object Registry {
+    private val registry = Array.fill[Option[Status]](MaxCode + 1)(None)
```
rossabaker: Could we microoptimize this by filling it with `Either[SomeDummyValue, Status]`?  Then we could return the same reference without unwrapping an `Option` and recreating a `Right`.

core/src/main/scala/org/http4s/Status.scala
```diff
@@ -76,102 +77,123 @@ object Status {
...
+    }
 
-  // scalastyle:off magic.number
-  private val registry =
-    new AtomicReferenceArray[Right[Nothing, Status]](600)
-  // scalastyle:on magic.number
+  def registered: Iterable[Status] = all
+
+  private def parseAsStatus(code: Int, reason: String = ""): ParseResult[Status] =
+    if (isStandard(code)) ParseResult.success(Status(code, reason))
```
rossabaker: Isn't `isInRange` what's important here?  We should be able to create, say, `473 I AM A PINK FLAMINGO`.

core/src/main/scala/org/http4s/Status.scala
```diff
@@ -76,102 +77,123 @@ object Status {
...
+
+    def register(status: Status): Status = {
+      registry(status.code) = Some(status)
+      status
+    }
+
+    def all: Iterable[Status] =
+      for {
+        code <- MinCode to MaxCode
+        status <- lookup(code)
```
rossabaker: Wait, what?  These monads don't line up.  Is there some implicit optionToIterable silliness happening here?

core/src/main/scala/org/http4s/Status.scala
```diff
@@ -76,102 +77,123 @@ object Status {
...
-  val BadGateway = register(Status(502, "Bad Gateway"))
-  val ServiceUnavailable = register(Status(503, "Service Unavailable"))
-  val GatewayTimeout = register(Status(504, "Gateway Timeout"))
-  val HttpVersionNotSupported = register(Status(505, "HTTP Version not supported"))
-  val VariantAlsoNegotiates = register(Status(506, "Variant Also Negotiates"))
-  val InsufficientStorage = register(Status(507, "Insufficient Storage"))
-  val LoopDetected = register(Status(508, "Loop Detected"))
-  val NotExtended = register(Status(510, "Not Extended"))
-  val NetworkAuthenticationRequired = register(Status(511, "Network Authentication Required"))
+  val Continue: Status = register(Status(100, "Continue", isEntityAllowed = false))
```
rossabaker: This doesn't hurt the singleton type tricks we use in the DSL?  I'm trying to think why these might not have been ascribed before.  If it didn't break things, :+1:

core/src/main/scala/org/http4s/Status.scala
```diff
@@ -76,102 +77,123 @@ object Status {
...
+    lookup(code, reason) match {
+      case None => parseAsStatus(code, reason)
+      case Some(status) => Right(status)
+    }
 
-  // scalastyle:off magic.number
-  private val registry =
-    new AtomicReferenceArray[Right[Nothing, Status]](600)
-  // scalastyle:on magic.number
+  def registered: Iterable[Status] = all
```
rossabaker: `Iterable` is a not very useful thing.  Maybe we should go ahead and commit to a `List` here?  At minimum, an `immutable.Seq`, so the user knows it's, like, immutable.

core/src/main/scala/org/http4s/Status.scala
ghostroad: Yeah, I'm confused about this. I just commented on this in the issue thread, because I thought @jmcardon was saying otherwise.

core/src/main/scala/org/http4s/Status.scala
ghostroad: Hmm... now that you mention it, I don't know what's happening either. I'll try to simplify it.

core/src/main/scala/org/http4s/Status.scala
ghostroad: It appears to be alright, so I guess we're lucky.

core/src/main/scala/org/http4s/Status.scala
ghostroad: Makes sense.

core/src/main/scala/org/http4s/Status.scala
jmcardon: In his defense, I did say that IMO, creating custom statuses should be disallowed.

ghostroad:
Hopefully, this incorporates the suggestions. 

1) `Status.fromInt` now behaves as it did before: it returns a custom status when you give it a nonstandard code. 
2) The registry is an `Array[Either[String, Status]]`. 
3) `Status.registered` returns a `List[Status]`. 
4) `MaxCode`, `MinCode`, and `Status.registered` are `private[http4s]`. They're used in `ArbitraryInstances`.

ghostroad:
Looks like I spoke to soon. `Either.filterOrElse` isn't available in Scala 2.11 and earlier. 

I've eliminated its use. I also took the opportunity to turn the registry into an `Array[ParseResult[Status]]`. This means that the typical uses of `Status.fromInt` and `Status.fromIntAndReason` by standard codes and reasons are now direct lookups from the registry with no intervening boxing/unboxing of data.

core/src/main/scala/org/http4s/Status.scala
```diff
@@ -77,51 +77,46 @@ object Status {
...
+  private[http4s] val MinCode = 100
+  private[http4s] val MaxCode = 599
 
-  def fromInt(code: Int): ParseResult[Status] = lookup(code) match {
-    case None => parseAsStatus(code)
-    case Some(status) => Right(status)
+  def fromInt(code: Int): ParseResult[Status] = withRangeCheck(code) {
+    val lookupResult = lookup(code)
+    if (lookupResult.isRight) lookupResult
+    else ParseResult.success(Status(code, "No reason provided"))
```
rossabaker: Nit: I'd probably leave an empty string here.  I think that's what most web libraries do.

core/src/main/scala/org/http4s/Status.scala
```diff
@@ -77,51 +77,46 @@ object Status {
...
   }
 
-  def fromIntAndReason(code: Int, reason: String): ParseResult[Status] =
-    lookup(code, reason) match {
-      case None => parseAsStatus(code, reason)
-      case Some(status) => Right(status)
-    }
+  def fromIntAndReason(code: Int, reason: String): ParseResult[Status] = withRangeCheck(code) {
+    val lookupResult = lookup(code, reason)
+    if (lookupResult.isRight) lookupResult else ParseResult.success(Status(code, reason))
```
rossabaker: The match can still be done without allocating, if you like.  (It's all about the same to me.)

```scala
lookup(code, reason) match {
  case right: Right => right
  case _: Left => ParseResult.success(Status(code, reason))
}
```

core/src/main/scala/org/http4s/Status.scala
```diff
@@ -76,102 +77,121 @@ object Status {
...
-  // scalastyle:on magic.number
+  private[http4s] def registered: List[Status] = all
 
-  def registered: Iterable[Status] =
-    for {
-      code <- 100 to 599
-      status <- Option(registry.get(code)).map(_.right.get)
-    } yield status
+  private object Registry {
+    private val registry: Array[ParseResult[Status]] =
```
jmcardon: :tada:

PR #2036, opened 2018-08-26
add jetty backend for http4s client

### LINKED ISSUE #2013: jetty-client
In addition to server, jetty provides a client.  An http4s-jetty-client backend could provide us our first asynchronous, HTTP/2 client.

This is an ambitious first issue, but there is a lot of prior art with three other client backends.

marcodippy:
closes #2013


marcodippy:
I've got an [alternative implementation](https://gist.github.com/marcodippy/eddf5905662936b92cb73b3646c1880f) for this but it requires an additional dependency ([jetty-reactive-httpclient](https://github.com/jetty-project/jetty-reactive-httpclient))

jetty-client/src/main/scala/org/http4s/client/jetty/JettyClient.scala
```diff
@@ -0,0 +1,131 @@
...
+      jReq.header(h.name.toString, h.value)
+
+    jReq.content(requestContentProvider(request))
+  }
+
+  private def requestContentProvider[F[_]](
+      request: Request[F])(implicit F: Effect[F], ec: ExecutionContext) = {
+    val dcp = new DeferredContentProvider()
+
+    ec.execute(new Runnable {
```
marcodippy: I'm not sure if running this on the ec used for running the responses is the right thing to do... ideas?

marcodippy:
Another question. Quoting the jetty client doc:

> While the request content is awaited and consequently uploaded by the client application, the server may be able to respond (at least with the response headers) completely asynchronously. In this case, Response.Listener callbacks will be invoked before the request is fully sent. This allows fine-grained control of the request/response conversation: for example the server may reject contents that are too big, send a response to the client, which in turn may stop the content upload.

I'm not entirely sure how to enable this kind of "fine-grained control of the req/resp conversation" given the current implementation

jetty-client/src/main/scala/org/http4s/client/jetty/JettyClient.scala
```diff
@@ -0,0 +1,131 @@
...
+      override def run(): Unit =
+        request.body.chunks
+          .map(_.toByteBuffer)
+          .to(Sink((b: ByteBuffer) => F.delay { dcp.offer(b); () }))
```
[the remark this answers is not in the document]
rossabaker: Oh, the Jetty Chunk's `Callback` is invoked when the content is read.  I think we probably want to use `Chunk` instead of `ByteBuffer` here, because with `Chunk`, you could implement backpressure.  As it stands, we're just going to fill up the `dcp` without any regard to how fast the server consumes it.

jetty-client/src/main/scala/org/http4s/client/jetty/JettyClient.scala
marcodippy: so, I don't think that `offer` actually blocks the thread, the chunks are written to a `java.util.ArrayDeque` that automatically resizes when needed (code [here](https://github.com/eclipse/jetty.project/blob/5b5f2fcf5f8241dd9edb998b97ad5e6cd45130a8/jetty-client/src/main/java/org/eclipse/jetty/client/util/DeferredContentProvider.java#L155)) 

the reason why I relocated it is that otherwise the request won't be sent until the entire body has been loaded into the content provider. I could move the execution of this block of code after the `request.send(...)` so that it wouldn't need to be relocated.

jetty-client/src/main/scala/org/http4s/client/jetty/JettyClient.scala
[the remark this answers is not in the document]
marcodippy: true, right now the dcp is gonna be loaded regardless of any circumstances. I'm not entirely sure how to implement the back pressure mechanism but I'll try to figure out something, I guess that each `dcp.offer` should somehow wait on the previous callback to be called with `.succeed()` or `.failed(Throwable t)` and decide either to continue or to stop processing, right? 

Btw, I'm wondering if by using the [InputStreamContentProvider](https://github.com/eclipse/jetty.project/blob/jetty-9.4.x/jetty-client/src/main/java/org/eclipse/jetty/client/util/InputStreamContentProvider.java#L54) we would get this semantic for free (but it would obviously block on `InputStream.read`)

jetty-client/src/main/scala/org/http4s/client/jetty/JettyClient.scala
```diff
@@ -0,0 +1,144 @@
...
+
+            ec.execute(new Runnable {
+              override def run(): Unit = cb(dr)
+            })
+          }
+
+          override def onContent(response: JettyResponse, content: ByteBuffer): Unit = {
+            val copy = ByteBuffer.allocate(content.remaining())
+            copy.put(content).flip()
+            queue.enqueue1(copy.some).runAsync(loggingAsyncCallback(logger)).unsafeRunSync()
```
marcodippy: I guess we lack of some kind of back pressure mechanism here too @rossabaker ; we could use the `onContent(Response response, ByteBuffer content, Callback callback)` method here and invoke the `callback` to signal that we need more content (the `callback.succeed()` doesn't control only the "demand content" semantic but also the releasing of the ByteBuffer, see [here](https://github.com/eclipse/jetty.project/issues/2429), I'm not sure how big of a problem it would be here)

jetty-client/src/main/scala/org/http4s/client/jetty/JettyClient.scala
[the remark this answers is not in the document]
rossabaker: The [PublisherContentProvider](https://github.com/jetty-project/jetty-reactive-httpclient/blob/9bf80e8c25ad10cdaa3aeb3d5af0b51a0ffe3ec3/src/main/java/org/eclipse/jetty/reactive/client/internal/PublisherContentProvider.java#L67) isn't _exactly_ the same thing, but might be an inspiration.

What about a semaphore with 1 permit?  We take a permit to emit a chunk, and return a permit on each callback?

jetty-client/src/main/scala/org/http4s/client/jetty/JettyClient.scala
rossabaker: Ooh, yeah, good point.  The queue can be a bounded queue, but we don't want to block of spin.  Yeah, I think something with the callback would be the right approach here, too.

jetty-client/src/main/scala/org/http4s/client/jetty/JettyClient.scala
```diff
@@ -4,141 +4,64 @@ package jetty
...
+import org.eclipse.jetty.client.api.{Request => JettyRequest}
+import org.eclipse.jetty.http.{HttpVersion => JHttpVersion}
+import org.log4s.{Logger, getLogger}
 import scala.concurrent.ExecutionContext
-import scala.util.control.NonFatal
 
 object JettyClient {
 
-  private val logger = getLogger
+  implicit private val logger: Logger = getLogger
```
rossabaker: Hmm. Where is this passed implicitly?  I don't think I've ever seen this done with log4s.

jetty-client/src/main/scala/org/http4s/client/jetty/JettyClient.scala
```diff
@@ -4,141 +4,64 @@ package jetty
...
     F.pure(client)
       .flatTap(client => F.delay { client.start() })
-      .map(client => {
+      .map(client =>
         Client(
           Kleisli { req =>
-            F.async[DisposableResponse[F]] { cb =>
-              toRequest(client, req).send(responseHandler(cb))
+            F.asyncF[DisposableResponse[F]] { cb =>
+              Resource
```
rossabaker: You could just use `Bracket` here instead of making a `Resource`.  `Bracket` excels at nesting more than anything.

jetty-client/src/main/scala/org/http4s/client/jetty/JettyClient.scala
```diff
@@ -4,141 +4,64 @@ package jetty
...
-          F.delay({
-            try {
-              client.stop()
-            } catch {
-              case NonFatal(t) => logger.error(t)("Unable to shut down Jetty client")
-            }
-          })
-        )
-      })
+          F.delay(client.stop()).onError {
```
rossabaker: `handleErrorWith`?  I don't think we want this to fail the task altogether.

jetty-client/src/main/scala/org/http4s/client/jetty/JettyClient.scala
```diff
@@ -4,141 +4,64 @@ package jetty
...
-          .runAsync {
-            case Left(t) =>
-              IO { logger.error(t)("Unable to write to Jetty sink") }
-            case Right(_) =>
-              IO.unit
-          }
-          .flatTap(_ => IO { dcp.close() })
-          .unsafeRunSync()
-    })
+    request.httpVersion match {
```
rossabaker: Could wrap `jReq.version` around the match and only say it once.

jetty-client/src/main/scala/org/http4s/client/jetty/JettyClient.scala
marcodippy: I have no idea how that `implicit` ended up in there :)

marcodippy:
hey @rossabaker, thank you for all the feedbacks :) 
I just pushed some changes to address your last comments

marcodippy:
great, thank you @rossabaker! I'll be happy to help again on http4s :)

PR #2040, opened 2018-08-30
Upgrade to circe-0.10.0-M2

rossabaker:


rossabaker:
This is going to fail from a bad merge train. c93601fe1f is the answer.

rossabaker:
Rebased on what I hope is the fixed master. I'm going to bed.

rossabaker:
This upgrade to circe-0.10 may have been premature: we may be ready before circe is ready, and the only real dependency conflict is circe-jawn, which we could make an internal facade for.  (We did this for argonaut once.)

PR #2042, opened 2018-09-02
Initial implementation of Throttle middleware

keirlawson:
Feedback appreciated!

There is definitely room for more features, in a subsequent PR I'd like to make taking tokens a function of the `Request` so that throttling can be based on request sizes or IPs rather than just simply the number of requests.  I'd also like to enable customisation of the response to throttled requests.  If these are considered essential then I can add them to this PR however my preference would be to get this merged before iterating further.

### DIFF
--- a/server/src/main/scala/org/http4s/server/middleware/Throttle.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/Throttle.scala
@@ -1,13 +1,12 @@
 package org.http4s.server.middleware
 
-import org.http4s.{Headers, Http, Response, Status}
+import org.http4s.{Http, Response, Status}
 import cats.data.Kleisli
 import cats.effect.{Clock, Sync}
 import cats.effect.concurrent.Ref
 import scala.concurrent.duration.FiniteDuration
 import cats.implicits._
 import java.util.concurrent.TimeUnit.NANOSECONDS
-import org.http4s.headers.`Retry-After`
 import scala.concurrent.duration._
 
 /**
@@ -59,8 +58,8 @@ object Throttle {
                     setter((newTokenTotal - 1, currentTime))
                       .map(_.guard[Option].as(TokenAvailable))
                   } else {
-                    val timeToFull = (refillEvery.toNanos * capacity) - timeDifference
-                    val successResponse = TokenUnavailable(timeToFull.nanos.some)
+                    val timeToNextToken = refillEvery.toNanos - timeDifference
+                    val successResponse = TokenUnavailable(timeToNextToken.nanos.some)
                     setter((newTokenTotal, currentTime)).map(_.guard[Option].as(successResponse))
                   }
 
@@ -96,12 +95,9 @@ object Throttle {
     createBucket.map(bucket => apply(bucket)(http))
   }
 
-  def defaultResponse[F[_]](retryAfter: Option[FiniteDuration]): Response[F] = retryAfter match {
-    case Some(retryTime) => {
-      val retryHeader = `Retry-After`.unsafeFromDuration(retryTime)
-      Response[F](Status.TooManyRequests, headers = Headers(retryHeader))
-    }
-    case None => Response[F](Status.TooManyRequests)
+  def defaultResponse[F[_]](retryAfter: Option[FiniteDuration]): Response[F] = {
+    val _ = retryAfter
+    Response[F](Status.TooManyRequests)
   }
 
   /**
--- a/server/src/main/scala/org/http4s/server/middleware/Throttle.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/Throttle.scala
@@ -1,0 +1,125 @@
+package org.http4s.server.middleware
+
+import org.http4s.{Headers, Http, Response, Status}
+import cats.data.Kleisli
+import cats.effect.{Clock, Sync}
+import cats.effect.concurrent.Ref
+import scala.concurrent.duration.FiniteDuration
+import cats.implicits._
+import java.util.concurrent.TimeUnit.NANOSECONDS
+import org.http4s.headers.`Retry-After`
+import scala.concurrent.duration._
+
+/**
+  * Transform a service to reject any calls the go over a given rate.
+  */
+object Throttle {
+
+  sealed abstract class TokenAvailability extends Product with Serializable
+  case object TokenAvailable extends TokenAvailability
+  final case class TokenUnavailable(retryAfter: Option[FiniteDuration]) extends TokenAvailability
+
+  /**
+    * A token bucket for use with the [[Throttle]] middleware.  Consumers can take tokens which will be refilled over time.
+    * Implementations are required to provide their own refill mechanism.
+    *
+    * Possible implementations include a remote TokenBucket service to coordinate between different application instances.
+    */
+  trait TokenBucket[F[_]] {
+    def takeToken: F[TokenAvailability]
+  }
+
+  object TokenBucket {
+
+    /**
+      * Creates an in-memory [[TokenBucket]].
+      *
+      * @param capacity the number of tokens the bucket can hold and starts with.
+      * @param refillEvery the frequency with which to add another token if there is capacity spare.
+      * @return A task to create the [[TokenBucket]].
+      */
+    def local[F[_]](capacity: Int, refillEvery: FiniteDuration)(
+        implicit F: Sync[F],
+        clock: Clock[F]): F[TokenBucket[F]] = {
+
+      def getTime = clock.monotonic(NANOSECONDS)
+      val bucket = getTime.flatMap(time => Ref[F].of((capacity.toDouble, time)))
+
+      bucket.map { counter =>
+        new TokenBucket[F] {
+          override def takeToken: F[TokenAvailability] = {
+            val attemptUpdate = counter.access.flatMap {
+              case ((previousTokens, previousTime), setter) =>
+                getTime.flatMap(currentTime => {
+                  val timeDifference = currentTime - previousTime
+                  val tokensToAdd = timeDifference.toDouble / refillEvery.toNanos.toDouble
+                  val newTokenTotal = Math.min(previousTokens + tokensToAdd, capacity.toDouble)
+
+                  val attemptSet: F[Option[TokenAvailability]] = if (newTokenTotal >= 1) {
+                    setter((newTokenTotal - 1, currentTime))
+                      .map(_.guard[Option].as(TokenAvailable))
+                  } else {
+                    val timeToFull = (refillEvery.toNanos * capacity) - timeDifference
+                    val successResponse = TokenUnavailable(timeToFull.nanos.some)
+                    setter((newTokenTotal, currentTime)).map(_.guard[Option].as(successResponse))
+                  }
+
+                  attemptSet
+                })
+            }
+
+            def loop: F[TokenAvailability] = attemptUpdate.flatMap { attempt =>
+              attempt.fold(loop)(token => token.pure[F])
+            }
+            loop
+          }
+
+        }
+      }
+
+    }
+
+  }
+
+  /**
+    * Limits the supplied service to a given rate of calls using an in-memory [[TokenBucket]]
+    *
+    * @param amount the number of calls to the service to permit within the given time period.
+    * @param per the time period over which a given number of calls is permitted.
+    * @param http the service to transform.
+    * @return a task containing the transformed service.
+    */
+  def apply[F[_], G[_]](amount: Int, per: FiniteDuration)(
+      http: Http[F, G])(implicit F: Sync[F], timer: Clock[F]): F[Http[F, G]] = {
+    val refillFrequency = per / amount.toLong
+    val createBucket = TokenBucket.local(amount, refillFrequency)
+    createBucket.map(bucket => apply(bucket)(http))
+  }
+
+  def defaultResponse[F[_]](retryAfter: Option[FiniteDuration]): Response[F] = retryAfter match {
+    case Some(retryTime) => {
+      val retryHeader = `Retry-After`.unsafeFromDuration(retryTime)
+      Response[F](Status.TooManyRequests, headers = Headers(retryHeader))
+    }
+    case None => Response[F](Status.TooManyRequests)
+  }
+
+  /**
+    * Limits the supplied service using a provided [[TokenBucket]]
+    *
+    * @param bucket a [[TokenBucket]] to use to track the rate of incoming requests.
+    * @param throttleResponse a function that defines the response when throttled, may be supplied a suggested retry time depending on bucket implementation.
+    * @param http the service to transform.
+    * @return a task containing the transformed service.
+    */
+  def apply[F[_], G[_]](
+      bucket: TokenBucket[F],
+      throttleResponse: Option[FiniteDuration] => Response[G] = defaultResponse[G] _)(
+      http: Http[F, G])(implicit F: Sync[F]): Http[F, G] =
+    Kleisli { req =>
+      bucket.takeToken.flatMap {
+        case TokenAvailable => http(req)
+        case TokenUnavailable(retryAfter) => throttleResponse(retryAfter).pure[F]
+      }
+    }
+}
--- a/server/src/test/scala/org/http4s/server/middleware/ThrottleSpec.scala
+++ b/server/src/test/scala/org/http4s/server/middleware/ThrottleSpec.scala
@@ -1,0 +1,155 @@
+package org.http4s.server.middleware
+
+import cats.effect.{IO, Timer}
+import cats.effect.laws.util.TestContext
+import org.http4s.{Http4sSpec, HttpApp, Request, Status}
+import cats.implicits._
+import org.http4s.dsl.io._
+import cats.effect.IO.ioEffect
+import org.http4s.headers.`Retry-After`
+import org.specs2.matcher.FutureMatchers
+import scala.concurrent.duration._
+import org.specs2.concurrent.ExecutionEnv
+import Throttle._
+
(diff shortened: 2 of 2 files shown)


server/src/main/scala/org/http4s/server/middleware/Throttle.scala
```diff
@@ -0,0 +1,90 @@
...
+
+import org.http4s.{Http, Response, Status}
+import cats.data.Kleisli
+import cats.effect.{Concurrent, Timer}
+import cats.effect.concurrent.Ref
+import scala.concurrent.duration._
+import cats.implicits._
+import fs2.Stream
+
+sealed trait TokenAvailability
```
jcranky: `Token` makes my mind think about authentication / Authorization. Is this a common terminology used for throttling as well?

SystemFw: It is. The algorithm Keir used here is commonly known as a "Token bucket". 
If you think about it _is_ similar to authentication, in both cases a token grants you the capability to do something.

jcranky: @SystemFw fair, thanks for the clarification :)

rossabaker: `sealed abstract class TokenAvailability extends Product with Serializable`: noisier, but better inference an bin compat.

This is dumping a bunch of symbols into the middleware package.  I wonder whether this ADT, and the `TokenBucket`, should be moved inside `Throttle`.  Or maybe, in the opposite direction, it could be useful on a `Client`, and we should move it into `org.http4s.util`.  :thinking:

server/src/main/scala/org/http4s/server/middleware/Throttle.scala
```diff
@@ -0,0 +1,90 @@
...
+    *
+    * @param capacity the number of tokens the bucket can hold and starts with.
+    * @param refillEvery the frequency with which to add another token if there is capacity spare.
+    * @return A singleton stream containing the [[TokenBucket]].
+    */
+  def local[F[_]](capacity: Int, refillEvery: FiniteDuration)(
+      implicit F: Concurrent[F],
+      timer: Timer[F]): Stream[F, TokenBucket[F]] =
+    Stream.eval(Ref[F].of(capacity)).flatMap { counter =>
+      val refill = Stream
```
rossabaker: I've implemented this previously by storing the last time a token was requested.  Then you can interpolate out how much has been refilled from last time before deciding whether or not to grant a token, and not need a background process.  This approach results in fractional tokens.

I think we'd also then get away with an `F[TokenBucket[F]]` instead of a `Stream[F, TokenBucket[F]]`.

server/src/main/scala/org/http4s/server/middleware/Throttle.scala
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  /**
+    * Limits the supplied service using a provided [[TokenBucket]]
+    *
+    * @param bucket a [[TokenBucket]] to use to track the rate of incoming requests.
+    * @param http the service to transform.
+    * @return a singleton stream containing the transformed service.
+    */
+  def apply[F[_], G[_]](bucket: TokenBucket[F])(http: Http[F, G])(
```
rossabaker: Should we make the 429 response customizable?

server/src/main/scala/org/http4s/server/middleware/Throttle.scala
SystemFw: > I think we'd also then get away with an F[TokenBucket[F]] instead of a Stream[F, TokenBucket[F]].

we can do this regardless btw. Returning a Resource[TokenBucket]` by `compile.drain.start` and `cancel` as the shutdown action.

You mentioned the refill approach a couple times but never with enough detail to understand _how_ you did it, so this is interesting :)

SystemFw: Just out of curiosity, it wouldn't work with a TokenBucket that had the task waiting rather than rejected, right? (if more recent requests don't arrive, old tasks are stuck)

server/src/main/scala/org/http4s/server/middleware/Throttle.scala
SystemFw: Slight preference for having it into `Throttle` and not `util` for the time being.

rossabaker: But move `Throttle` to another package or leave it here?  I agree with moving the ADT into `Throttle` regardless.

`org.http4s.utill` is a cursed place of "things that make us wish that Scala were more amenable to microlibraries," so I do acknowledge the disgust.

SystemFw: I won't oppose moving into `util` once we have the need. For now I'd rather keep it here so that the code is easier to follow. 

On a related not wrt reusability, client-side limiting imho is not a great fit for a token bucket. `upperbound` (which focuses on that) has a different design for that reason

server/src/main/scala/org/http4s/server/middleware/Throttle.scala
rossabaker: I think my rejection message included a "retry-after" like value.  I don't think I maintained a queue.  In either case, we can compute the time until full and try again.  It does get messy if we're returning the same retry-after to multiple requestors and get slammed.

server/src/main/scala/org/http4s/server/middleware/Throttle.scala
rossabaker: Now that I mention Retry-After above, hey, how about a Retry-After? :smile:

keirlawson: As I said in the PR message I want to implement this, I was just planning to do so in a follow-up PR, but if the feeling is that it is essential I can do it in this one.

SystemFw: I think customisable responses are really easy to add here, and perhaps the rest can go in another PR?

keirlawson:
Rebased on latest master, swapped to @rossabaker's solution for refilling bucket, added response customisation and retry-after set to when the bucket ought to be full.  

I'm not totally convinced with the retry-after value, you could argue it should be when the next token will be available rather than when the bucket should be full, though in the real world if clients pay attention to that things will likely blow up, seems to me neither the next-token-time nor the bucket-full-time really make sense and it would be best to leave it up to the user as I think any value is either arbitrary or app-specific.

server/src/main/scala/org/http4s/server/middleware/Throttle.scala
```diff
@@ -0,0 +1,120 @@
...
+                val timeDifference = currentTime - previousTime
+                val tokensToAdd = timeDifference.toDouble / refillEvery.toNanos.toDouble
+                val newTokenTotal = Math.min(previousTokens + tokensToAdd, capacity.toDouble)
+
+                val attemptSet: F[Option[TokenAvailability]] = if (newTokenTotal >= 1) {
+                  setter((newTokenTotal - 1, currentTime)).map(success => if (success) Some(TokenAvailable()) else None)
+                } else {
+                  val timeToFull = (refillEvery.toNanos * capacity) - timeDifference
+                  val successResponse = TokenUnavailable(Some(FiniteDuration(timeToFull, NANOSECONDS)))
+                  setter((newTokenTotal, currentTime)).map(success => if (success) Some(successResponse) else None)
```
SystemFw: `TokenUnavailable(Some(FiniteDuration(timeToFull, NANOSECONDS)))`

apologies for the bikeshedding (feel free to ignore)

`TokenUnavailable(timeToFull.nanos.some)`

server/src/main/scala/org/http4s/server/middleware/Throttle.scala
```diff
@@ -0,0 +1,120 @@
...
+import org.http4s.headers.`Retry-After`
+
+/**
+  * Transform a service to reject any calls the go over a given rate.
+  */
+object Throttle {
+
+  sealed abstract class TokenAvailability extends Product with Serializable
+  case class TokenAvailable() extends TokenAvailability
+  case class TokenUnavailable(retryAfter: Option[FiniteDuration]) extends TokenAvailability
```
SystemFw: Is this `Option` even `None`?

keirlawson: It could be with other `TokenBucket` impls, ie I envisage an "outstanding requests" version where you issue tokens when there are less than n requests currently being executed, in that scenario there is no obvious time-to-full

server/src/main/scala/org/http4s/server/middleware/Throttle.scala
keirlawson: Bikeshedding appreciated, I had the feeling there would be a more concise way to do this...

server/src/main/scala/org/http4s/server/middleware/Throttle.scala
```diff
@@ -1,90 +1,125 @@
...
+  * Transform a service to reject any calls the go over a given rate.
   */
-trait TokenBucket[F[_]] {
-  def takeToken: F[TokenAvailability]
-}
+object Throttle {
 
-object TokenBucket {
+  sealed abstract class TokenAvailability extends Product with Serializable
+  case class TokenAvailable() extends TokenAvailability
```
rossabaker: Why did this cease to be a case object?

server/src/main/scala/org/http4s/server/middleware/Throttle.scala
```diff
@@ -1,90 +1,125 @@
...
+  case class TokenUnavailable(retryAfter: Option[FiniteDuration]) extends TokenAvailability
```
rossabaker: Any reason not to make this a final case class?

server/src/main/scala/org/http4s/server/middleware/Throttle.scala
```diff
@@ -1,90 +1,125 @@
...
+    * @param throttleResponse a function that defines the response when throttled, may be supplied a suggested retry time depending on bucket implementation.
     * @param http the service to transform.
-    * @return a singleton stream containing the transformed service.
+    * @return a task containing the transformed service.
     */
-  def apply[F[_], G[_]](bucket: TokenBucket[F])(http: Http[F, G])(
-      implicit F: Concurrent[F]): Http[F, G] =
+  def apply[F[_], G[_]](
+      bucket: TokenBucket[F],
+      throttleResponse: Option[FiniteDuration] => Response[G] = defaultResponse[G] _)(
```
rossabaker: I like this, because people can alternately fill in a timestamp instead of a duration if they prefer.  (Or of course add a custom body.)

server/src/main/scala/org/http4s/server/middleware/Throttle.scala
```diff
@@ -1,90 +1,125 @@
...
     *
     * @param amount the number of calls to the service to permit within the given time period.
     * @param per the time period over which a given number of calls is permitted.
     * @param http the service to transform.
-    * @return a singleton stream containing the transformed service.
+    * @return a task containing the transformed service.
     */
   def apply[F[_], G[_]](amount: Int, per: FiniteDuration)(
-      http: Http[F, G])(implicit F: Concurrent[F], timer: Timer[F]): Stream[F, Http[F, G]] = {
+      http: Http[F, G])(implicit F: Sync[F], timer: Clock[F]): F[Http[F, G]] = {
```
rossabaker: This is kind of confusing: the types are the same as `TokenBucket.local`, but we do extra math first.

I wonder whether this overload carries its weight at all, given that it doesn't allow customization of the throttled response.  I imagine we could live without it, but I'd defer to you since you've probably tried to use this and I haven't.

keirlawson: I'm not totally attached but my thinking was that some people will not care about the token bucket impl so easier to have a default where you don't need to construct your own, and the x requests per y time was a more "natural" way that people would think about their rate limit, ie "100 per second" rather than "1 per 0.6 seconds"

SystemFw: I'm very slightly 👎 on the overload

keirlawson:
Per discussion in the Gitter channel, have changed behavior to not use the retry-after header by default, however time-to-next-token value is made available to the `throttleRepose` function should someone wish to use it, time-to-bucket-full should be derivable from this + refill rate.

server/src/main/scala/org/http4s/server/middleware/Throttle.scala
```diff
@@ -0,0 +1,121 @@
...
+      val bucket = getTime.flatMap(time => Ref[F].of((capacity.toDouble, time)))
+
+      bucket.map { counter =>
+        new TokenBucket[F] {
+          override def takeToken: F[TokenAvailability] = {
+            val attemptUpdate = counter.access.flatMap {
+              case ((previousTokens, previousTime), setter) =>
+                getTime.flatMap(currentTime => {
+                  val timeDifference = currentTime - previousTime
+                  val tokensToAdd = timeDifference.toDouble / refillEvery.toNanos.toDouble
```
jmcardon: Why are you doing a `toDouble` conversion just for the sake of using `Math.min`?

server/src/main/scala/org/http4s/server/middleware/Throttle.scala
```diff
@@ -0,0 +1,121 @@
...
+      * @param capacity the number of tokens the bucket can hold and starts with.
+      * @param refillEvery the frequency with which to add another token if there is capacity spare.
+      * @return A task to create the [[TokenBucket]].
+      */
+    def local[F[_]](capacity: Int, refillEvery: FiniteDuration)(
+        implicit F: Sync[F],
+        clock: Clock[F]): F[TokenBucket[F]] = {
+
+      def getTime = clock.monotonic(NANOSECONDS)
+      val bucket = getTime.flatMap(time => Ref[F].of((capacity.toDouble, time)))
```
jmcardon: why is this a `Double` over a `Long`? You'll essentially be working with integral amounts most of the time.

server/src/main/scala/org/http4s/server/middleware/Throttle.scala
```diff
@@ -0,0 +1,121 @@
...
+                  val tokensToAdd = timeDifference.toDouble / refillEvery.toNanos.toDouble
+                  val newTokenTotal = Math.min(previousTokens + tokensToAdd, capacity.toDouble)
+
+                  val attemptSet: F[Option[TokenAvailability]] = if (newTokenTotal >= 1) {
+                    setter((newTokenTotal - 1, currentTime))
+                      .map(_.guard[Option].as(TokenAvailable))
+                  } else {
+                    val timeToNextToken = refillEvery.toNanos - timeDifference
+                    val successResponse = TokenUnavailable(timeToNextToken.nanos.some)
+                    setter((newTokenTotal, currentTime)).map(_.guard[Option].as(successResponse))
```
jmcardon: Just to double check: Is this intended that if you keep sending requests while being throttled, it will remain throttled? I.e: If this is the case and you hit a throttle, while continuing to send requests, it will not let any through after the throttle, since it always updated the current time here.

server/src/test/scala/org/http4s/server/middleware/ThrottleSpec.scala
```diff
@@ -0,0 +1,143 @@
...
+      val testTimer: Timer[IO] = ctx.timer[IO]
+      val capacity = 5
+      val createBucket =
+        TokenBucket.local[IO](capacity, 100.milliseconds)(ioEffect, testTimer.clock)
+
+      val takeExtraToken = createBucket.flatMap(testee => {
+        val takeFiveTokens: IO[List[TokenAvailability]] = (1 to 5).toList.traverse(_ => {
+          testee.takeToken
+        })
+        testTimer.sleep(300.milliseconds) >> takeFiveTokens >> testee.takeToken
```
jmcardon: why does it sleep 300 millis, take 5 tokens then take another token? 

I might be misreading: This test aims to make sure that if trying to take a token at capacity, it won't add another token until demand goes down?

keirlawson: No, the intention of the test was to ensure the bucket never adds tokens beyond its capacity, ie we set it to capacity, wait a few refill intervals, then check that there are still only `capacity` tokens.

server/src/main/scala/org/http4s/server/middleware/Throttle.scala
keirlawson: `tokensToAdd` needs to allow for fractional tokens, if I remove both `toDouble`s from that line it will wind up a `long`

server/src/main/scala/org/http4s/server/middleware/Throttle.scala
keirlawson: If we want to allow fractional tokens, which is the basis of the algorithm suggested by @rossabaker then I think we need a double? Unless I am missing something?

server/src/main/scala/org/http4s/server/middleware/Throttle.scala
keirlawson: That is not the intention, and I don't *think* that is the behaviour here, though my reasoning could be wrong... it should keep adding fractional tokens until there is a whole token at which point the request passes, if you're unconvinced then I can add a test that with (hopefully!) prove this?

jmcardon: Yeah I was wrong, didn't take into account fractional updates.

server/src/test/scala/org/http4s/server/middleware/ThrottleSpec.scala
jmcardon: My mistake :+1:

PR #2043, opened 2018-09-02
Use Scalafmt in intellij

aeons:


jmcardon:
The fact this was removed was my bad, but I can probably reinsert it in #2027 

However, My IJ keeps editing this file no matter what I do. This is an intellij project file, meaning new contributors or anyone else using IJ may end up editing it automatically too. On top of this, running `test:scalafmt` is what people should be doing anyway.

Can we instead add some notes on setting up the right headers sorting instead of inserting project specific files? I don't think IJ files should ever be pushed, especially because the editor bugs the f out all the time, meaning I have to reimport the project by deleting the .idea file, but now it would force me to delete my whole fork and _then_ pull it again.

jmcardon:
To add to this:

I'm on the latest idea. Your formatter option is being automatically deleted by IJ. When I Select scalafmt formatter, I get
```xml
  <option name="FORMATTER" value="1" />
```

I do not expect their xml to be stable across intellij versions.

I think we should remove the idea project.xml, and instead make a note on the README about contributing with IntelliJ idea and sorting the imports correctly.

aeons:
Ugh, it has worked well for me so far, even with fresh clones of the project, but I don't have a problem with removing the IJ files.

jmcardon:
@aeons look at #2027 or #2026 . I don't edit these files myself, even after updating IJ messes with them every time :/.

jmcardon:
I mean, if you know any IJ settings so it doesn't edit that file automatically I'm all ears. I just don't know any.

rossabaker:
I would love to understand why this is happening to @jmcardon and not any of our other IntelliJ users.  I think having the project setting is nice, because before we had it, we had a lot of diff noise from imports because of IntelliJ's idiosyncratic default.

SystemFw:
IntelliJ got word that Jose got a Haskell job, now he's merely paying for his sins. I say let's keep the setting and let him suffer.

jmcardon:
@SystemFw LOL.

@rossabaker I'd love to know who else is actually PRing on intellij recently. Most of our PRs are between 4-5 people (3 of which are in this chat) and I'm privy to the information yall use emacs.

sorenbug:
My instance of IDEA is doing the same @jmcardon's is and I'm always on the latest version. It does set this option whenever you confirm using Scalafmt instead of default IJ format.

rossabaker:
:confused: Not counting Travis CI, GitHub reports 19 unique authors in the last month.

That seems like a defect in IntelliJ, given that scalafmt doesn't, and won't, organize imports.  Has anybody reported this?

In the meantime, I'm fine deleting this, but I'm frustrated by the merge conflicts it causes.  We could start calling out violations on code review, but that's just busywork.  Is there another technical solution anyone can think of?

jmcardon:
@rossabaker I think the simplest is just a page indicating how to configure your IJ, when you're creating a PR from it.

I think that moving forward this will happen more often, as IJ evolves and their xml schema is not stable. I don't know many repositories out there with ide-specific files committed for this very reason (eclipse as an example is a _nightmare_ about this).

rossabaker:
I've done this successfully on teams with a uniform environment, but if IntelliJ is really this sloppy, I agree that it's a losing battle.  I'd be in favor of removing this and adding to the CONTRIBUTING.md.

aeons:
To be fair, I haven't had time to do much scala since shortly after the scalafmt support in IJ was released, so they might have changed the format.
As I said I'd be ok with removing the config and adding some documentation on how to set up IJ.

PR #2048, opened 2018-09-04
Modify Logging within Retry

kevinmeredith:
In short, this PR updates logging when considering whether to re-try or not.

https://gitter.im/http4s/http4s?at=5b8ec8fc60f9ee7aa4d980ca prompted this change.

### DIFF
--- a/client/src/main/scala/org/http4s/client/middleware/Retry.scala
+++ b/client/src/main/scala/org/http4s/client/middleware/Retry.scala
@@ -44,6 +44,9 @@ object Retry {
                 s"Request $req threw an exception on attempt #${attempts} with message: ${e.getMessage}. Retrying after $duration.")
               nextAttempt(req, attempts, duration, None)
             case None =>
+              logger.info(
+                s"Request $req threw an exception with message: ${e.getMessage} on attempt #${attempts} attempts. Giving up."
+              )
               F.raiseError[DisposableResponse[F]](e)
           }
       }
--- a/client/src/main/scala/org/http4s/client/middleware/Retry.scala
+++ b/client/src/main/scala/org/http4s/client/middleware/Retry.scala
@@ -45,7 +45,7 @@ object Retry {
               nextAttempt(req, attempts, duration, None)
             case None =>
               logger.info(
-                s"Request $req threw an exception with message: ${e.getMessage} on attempt #${attempts} attempts. Giving up."
+                s"Request $req threw an exception on attempt #${attempts} with message: ${e.getMessage}. Giving up."
               )
               F.raiseError[DisposableResponse[F]](e)
           }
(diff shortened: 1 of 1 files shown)


client/src/main/scala/org/http4s/client/middleware/Retry.scala
```diff
@@ -41,9 +41,12 @@ object Retry {
             case Some(duration) =>
               // info instead of error(e), because e is not discarded
               logger.info(
-                s"Request $req threw an exception on attempt #$attempts attempts. Giving up.")
+                s"Request $req threw an exception on attempt #$attempts with message: ${e.getMessage}. Retrying after $duration.")
```
rossabaker: I think we should log the traces here.  It could get noisy, but if we don't, we'll never see the discarded exception.  (Line 42 should read "because we're trying again" and not "because e is discarded")

kevinmeredith: Hi @rossabaker - I don't entirely follow. Could you please say more as to what changes you're recommending? Thanks

rossabaker: `logger.info(e)(s"Request threw an exception on attempt #$attempts. Retrying after $duration")`

Without that, all we get is `${e.getMessage}`, which very often is `null`?  (What?  Where?  Who knows!)  When we pass `(e)` to `logger.info(e)`, we can diagnose the intermittent failures that resolve themselves.

kevinmeredith: Thanks for the explanation! I added this change in https://github.com/http4s/http4s/pull/2048/commits/c794efa52f2a70bf97bb9c9111bf8fe291d45707.

client/src/main/scala/org/http4s/client/middleware/Retry.scala
```diff
@@ -40,8 +40,7 @@ object Retry {
           policy(req, Left(e), attempts) match {
             case Some(duration) =>
               // info instead of error(e), because e is not discarded
-              logger.info(
-                s"Request $req threw an exception on attempt #$attempts with message: ${e.getMessage}. Retrying after $duration.")
+              logger.info(e)(s"Request threw an exception on attempt #$attempts. Retrying after $duration")
```
rossabaker: Would change the line above, too: log `e` because it's discarded, but at info level because we're trying again.

kevinmeredith: Good point. I'll change it too. Could you please explain:

> log e because it's discarded

? 

I don't understand this statement.

Thanks

jmcardon:
:+1: on CI pass

PR #2050, opened 2018-09-05
fix hang on master

jmcardon:


rossabaker:
I'd rather replace `Http4sSpec.TestExecutionContext` with global.  There's no reason to have two.

jmcardon:
This change is a tad bigger than I thought it was, one test is erroring out.

jmcardon:
I changed the pool to use the global pool purely for my test which abuses concurrency.

Issue #1270, opened 2017-07-20
Make connection pool logging easier


reactormonk:
It's already there, via `logger`. But I don't think it's possible to enable debug for a single class only in production (maybe it is), so I'd go with a logger marker for now, so it can be specifically enabled (and document it).

Issue #2005, opened 2018-08-17
Factor a common layer for client metrics


luisdeltoro:
There is two standalone implementations for client metrics on Htttp4s, one for prometeus metrics and one for dropwizard metrics.
It would be nice to try to get an abstraction that can deal with both implementations, hopefully removing duplication and enhancing long-term maintainability.

rossabaker:
As mentioned before, I tried this and failed server side.  As mentioned before, I'd love to see another attempt, because it still seems like a good idea.

luisdeltoro:
Hi @rossabaker,

I have a draft implementation of this issue. It can be found here: https://github.com/luisdeltoro/http4s/tree/CommonLayerClientMetrics_%232005/client-metrics/src

If you like the solution and think that it's worth committing I would clean up the branch, which still contains the other prometheus-client-metrics project and run it through scala format etc..

At the end, I'm using type classes to abstract the common layer that both implementations share.
From my point of view the code reads now more easily and it has the advantage of allowing other developers to easily provide typeclasses implementations for their metrics framework of choice.
I have added tests for both implementations to make sure, I did not brake anything.

Please let me if you think this is heading in the right direction.

rossabaker:
I think the approach is generally nice.  I see that a lot of the operations on `MetricsOps` are being `Sync.delay`'ed.  It would make more sense to me to keep the `MetricsOps` pure and return an `F` context.

This is tangentially related to #2053, where there's a listener interface for events that feed into metrics.  I think the right approach may end up having some similarities here.

luisdeltoro:
Hi @rossabaker,

I've changed the MetricsOps so it returns always an context F.

Regarding your second comment. I'm not sure I fully understand it. Do you mean that you would prefer that I convert MetricsOps to some kind of listener for events?

Maybe something like:
trait RequestListener[F[_]] {
  def requestInitiated(destination: Option[String]): F[Unit]
  def headersReceived(destination: Option[String]): F[Unit]
  def BodyConsumed(destination: Option[String]): F[Unit]
}

Or are you more referring to the fact of making the publishing of metrics asynchronous by getting it to run on another Fiber?

rossabaker:
I don't think it should be async publishing.  All of these calls should be very quick.  It _could_ be async publishing if the interpreter takes a `ContextShift`, but this would almost always be a bad idea.

[...]

PR #2053, opened 2018-09-06
Pool manager stats PoC

janstenpickle:
This is an attempt to address #1270 and just a PoC for now to see what people think about a possible implementation, I won't be offended if it doesn't get merged!

I've tried to maintain the same interface for the Blaze client, and made metrics publishing totally optional. The pool manager will only attempt to publish if it has been provided with a function to do so.

I've also attempted to make the process of publish metrics non-blocking by putting it in a separate fiber, with some guarantees around safety provided by a timeout and catch-all error handling. If anything goes wrong with the publish function, the error will be logged. If the publish function times out, the fact it did will also be logged.

In regard to the implementation, I considered making the stats publish `PoolManagerStats => F[Unit]`, however this would require it to be passed separately to the client configuration, or require an effect type parameter on the client configuration. So I opted for `PoolManagerStats => Unit`.

The obvious problem with `PoolManagerStats => Unit` is that it doesn't communicate the likely side effecting nature of publishing the stats. While side effects will be handled inside the pool manager, the user does not know this. It also removes the ability for the user to create an asynchronous operation.

Also it feels a bit dirty leaving an orphan fiber, which is never joined. I do attempt to ensure that threads aren't consumed, but I'd love for someone to show me a better way of doing this! Also I was considering having a `Timer` instance passed to `PoolManager` so that stats publishing could be executed on a separate thread pool.

Anyway, I'm just trying to spur discussion around this if the maintainers have time to take a look, that would be great!

### DIFF
--- a/client/src/main/scala/org/http4s/client/PoolManager.scala
+++ b/client/src/main/scala/org/http4s/client/PoolManager.scala
@@ -11,8 +11,9 @@ import scala.collection.mutable
 import scala.concurrent.ExecutionContext
 import scala.concurrent.duration._
 import scala.util.Random
+import cats.effect.implicits._
 
-private final class PoolManager[F[_], A <: Connection[F]](
+private final class PoolManager[F[_]: Timer, A <: Connection[F]](
     builder: ConnectionBuilder[F, A],
     maxTotal: Int,
     maxWaitQueueLimit: Int,
@@ -20,7 +21,9 @@ private final class PoolManager[F[_], A <: Connection[F]](
     responseHeaderTimeout: Duration,
     requestTimeout: Duration,
     semaphore: Semaphore[F],
-    implicit private val executionContext: ExecutionContext)(implicit F: Concurrent[F])
+    implicit private val executionContext: ExecutionContext,
+    listener: Option[PoolManagerListener[F]] = None,
+    listenerTimeout: FiniteDuration = 1.second)(implicit F: Concurrent[F])
     extends ConnectionManager[F, A] {
 
   private sealed case class Waiting(
@@ -36,6 +39,15 @@ private final class PoolManager[F[_], A <: Connection[F]](
   private val idleQueues = mutable.Map.empty[RequestKey, mutable.Queue[A]]
   private var waitQueue = mutable.Queue.empty[Waiting]
 
+  private def callListener(op: PoolManagerListener[F] => F[Unit]): F[Unit] =
+    listener.fold(F.unit) { li =>
+      op(li)
+        .timeoutTo(listenerTimeout, F.delay(logger.warn("Timed out calling stats listener")))
+        .handleErrorWith(th => F.delay(logger.warn(th)("Failed to call stats listener")))
+        .start
+        .void
+    }
+
   private def stats =
     s"curAllocated=$curTotal idleQueues.size=${idleQueues.size} waitQueue.size=${waitQueue.size} maxWaitQueueLimit=$maxWaitQueueLimit closed=${isClosed}"
 
@@ -49,23 +61,25 @@ private final class PoolManager[F[_], A <: Connection[F]](
     }
   }
 
-  private def incrConnection(key: RequestKey): F[Unit] = F.delay {
-    curTotal += 1
-    allocated.update(key, allocated.getOrElse(key, 0) + 1)
-  }
+  private def incrConnection(key: RequestKey): F[Unit] =
+    F.delay {
+      curTotal += 1
+      allocated.update(key, allocated.getOrElse(key, 0) + 1)
+    } *> callListener(_.onBorrow)
 
-  private def decrConnection(key: RequestKey): F[Unit] = F.delay {
-    curTotal -= 1
-    val numConnections = allocated.getOrElse(key, 0)
-    // If there are no more connections drop the key
-    if (numConnections == 1) {
-      allocated.remove(key)
-      idleQueues.remove(key)
-      ()
-    } else {
-      allocated.update(key, numConnections - 1)
-    }
-  }
+  private def decrConnection(key: RequestKey): F[Unit] =
+    F.delay {
+      curTotal -= 1
+      val numConnections = allocated.getOrElse(key, 0)
+      // If there are no more connections drop the key
+      if (numConnections == 1) {
+        allocated.remove(key)
+        idleQueues.remove(key)
+        ()
+      } else {
+        allocated.update(key, numConnections - 1)
+      }
+    } *> callListener(_.onRelease)
 
   private def numConnectionsCheckHolds(key: RequestKey): Boolean =
     curTotal < maxTotal && allocated.getOrElse(key, 0) < maxConnectionsPerRequestKey(key)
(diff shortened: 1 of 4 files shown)


blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientConfig.scala
```diff
@@ -56,7 +57,8 @@ final case class BlazeClientConfig( // HTTP properties
     // pipeline management
     bufferSize: Int,
     executionContext: ExecutionContext,
-    group: Option[AsynchronousChannelGroup]) {
+    group: Option[AsynchronousChannelGroup],
+    statsPublisher: Option[PoolManagerStats => Unit]) {
```
rossabaker: There is a `BlazeClientBuilder` now (I don't think it was there when you started), and the major rationale for that was so we could add future configuration in a binary compatible way.  These new properties should be migrated to that.  But that's where we'd want these.

client/src/main/scala/org/http4s/client/PoolManager.scala
```diff
@@ -36,6 +41,30 @@ private final class PoolManager[F[_], A <: Connection[F]](
   private val idleQueues = mutable.Map.empty[RequestKey, mutable.Queue[A]]
   private var waitQueue = mutable.Queue.empty[Waiting]
 
+  private def publishStats = statsPublisher match {
+    case None => F.unit
+    case Some(op) =>
+      F.start(
```
rossabaker: lol, not scalafmt's finest moment, is it?  If you import `cats.effect.implicits._`, you'll get syntax that makes this a little less deep.

client/src/main/scala/org/http4s/client/PoolManager.scala
```diff
@@ -20,9 +21,13 @@ private final class PoolManager[F[_], A <: Connection[F]](
     responseHeaderTimeout: Duration,
     requestTimeout: Duration,
     semaphore: Semaphore[F],
-    implicit private val executionContext: ExecutionContext)(implicit F: Concurrent[F])
+    implicit private val executionContext: ExecutionContext,
+    statsPublisher: Option[PoolManagerStats => Unit],
```
rossabaker: How about `PoolManagerStats => IO[Unit]`, with a default implementation of `_ => IO.unit`?  Just to avoid `Concurrent` overhead on something that's `F.unit`?

client/src/main/scala/org/http4s/client/PoolManager.scala
```diff
@@ -36,6 +41,30 @@ private final class PoolManager[F[_], A <: Connection[F]](
...
+                          curTotal,
+                          idleQueues.size,
+                          waitQueue.size,
+                          maxWaitQueueLimit,
+                          isClosed))))),
+              statsPublishTimeout,
+              F.delay(logger.warn("Timed out publishing stats"))
+            )
+            .handleErrorWith(th => F.delay(logger.warn(th)("Failed to process stats"))))
+        .map(_ => ())
```
rossabaker: This is equivalent to `.void`.

client/src/main/scala/org/http4s/client/PoolManager.scala
```diff
@@ -355,3 +399,10 @@ private final class PoolManager[F[_], A <: Connection[F]](
 
 case class NoConnectionAllowedException(key: RequestKey)
     extends IllegalArgumentException(s"No client connections allowed to $key")
+
+case class PoolManagerStats(
```
rossabaker: An abstract class would let us add more later without breaking compatibility.

janstenpickle:
Thanks very much for taking a look @rossabaker. I did see everything got updated with the new client builder, it all looks great, especially the use of `Resource` everywhere.

>  I'm wondering about more of an event listener

Ah yeah, this would be really nice, although it would be be able to publish gauges with the interal state of the pool as well, how about something like this?

```
trait Listener[F[_]] {
  def onClose(stats: PoolManagerStats): F[Unit]
  def onBorrow(stats: PoolManagerStats): F[Unit]
  def onRelease(stats: PoolManagerStats): F[Unit]
  def onWait(stats: PoolManagerStats): F[Unit]
}
```

Or do you think it would be better to separate the concerns, and have two separate params on the pool manager of `PoolManagerStats => F[Unit]` and `Listener`?

With the listener I suppose the ADT approach is more flexible to addition of new event types in the future, as downstream developers will only get a compilation warning on a inexhastive match rather than a compilation error on having an incomplete implementation of the listener trait.

The other thing I haven't address at all is timings, which will likely be more intrusive in the pool manager itself. It may be quite insightful recording wait times of requests in the queue.

blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientConfig.scala
janstenpickle: Ah yeah, thanks, I did see that. I notice it is now parameterised on `F` so it should be possible to add side-effecting callbacks to that quite easily.

client/src/main/scala/org/http4s/client/PoolManager.scala
janstenpickle: Haha, yeah I thought I could do with neatening this up a bit. It is a bit hadkouken.

img

client/src/main/scala/org/http4s/client/PoolManager.scala
janstenpickle: That would be nice, although I believe it would need an implict `ContextShfit[IO]` in order to schedule the work on a separate fiber (if using a fiber is a valid approach here!). We'd also then need to lift the result of starting the fiber into `F` anyway.

client/src/main/scala/org/http4s/client/PoolManager.scala
janstenpickle: I wasn't aware of that syntax, thanks!

rossabaker:
Gauge is a bit different between Prometheus and Dropwizard:
* A Prometheus Gauge, which is a counter that goes up and down, could be supported by tracking complementary events: onBorrow vs. onRelease.  A Dropwizard Counter is about the same.
* A Dropwizard Gauge would need the internal state (it's more of a pull model), though I'm not sure we need those here.

I think a `Listener` could be sufficient.

A problem with the ADT approach is that not only would the developer get an inexhaustive match on compilation, they'd get a `MatchError` until they fix it.  An addition would effectively require a major revision in strict semantic versioning.

A problem with the `Listener` approach is that we can't add defaulted methods to a trait without breaking binary compatibility until Scala 2.12.  Scala 2.11 use is dwindling, but it's still my production version and will be for a long time at a lot of Spark shops.

I do think timings would be a very useful addition to the events, however we model them.

janstenpickle:
Oh I see, I suppose that's up to the implementer of `Listener` then. If state does need to be tracked, that could be done so it can probably be done via `Ref`s. I'll do a refactor against the builder updates and create a listener trait.

janstenpickle:
I'm going to close this as things have moved on a fair bit so this code isn't really valid anymore, plus it looks like essence has been captured in #2005.

PR #2061, opened 2018-09-08
Send Disconnect event on EOF

rossabaker:
When we read an EOF, request a disconnect.  This signals outbound that we're done and allows a chance to clean up.

Before this change, when a client disconnects without either the server or client signalling the end of the connection, the HTTP stage is shut down but no command is passed out.  This delays the cleanup of the middle stages, such as `QuietTimeoutStage`.

### DIFF
--- a/blaze-server/src/main/scala/org/http4s/server/blaze/Http1ServerStage.scala
+++ b/blaze-server/src/main/scala/org/http4s/server/blaze/Http1ServerStage.scala
@@ -87,7 +87,7 @@ private[blaze] class Http1ServerStage[F[_]](
 
   private val handleReqRead: Try[ByteBuffer] => Unit = {
     case Success(buff) => reqLoopCallback(buff)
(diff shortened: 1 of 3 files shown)


blaze-server/src/main/scala/org/http4s/server/blaze/Http1ServerStage.scala
```diff
@@ -87,7 +87,7 @@ private[blaze] class Http1ServerStage[F[_]](
 
   private val handleReqRead: Try[ByteBuffer] => Unit = {
     case Success(buff) => reqLoopCallback(buff)
-    case Failure(Cmd.EOF) => stageShutdown()
+    case Failure(Cmd.EOF) => closeConnection()
```
rossabaker: Idiom question for blaze-0.14: should we be shutting down here, or waiting for a `Disconnected` message to find its way back to us?

PR #2063, opened 2018-09-08
Support cancellation in blaze-server on stage shutdown

rossabaker:
When a blaze server stage is shut down, we should be able to cancel any running request, because it's not going to render anyway.  This is similar to a servlet response being canceled when the AsyncContext timeout fires.


blaze-server/src/main/scala/org/http4s/server/blaze/Http1ServerStage.scala
```diff
@@ -253,12 +257,21 @@ private[blaze] class Http1ServerStage[F[_]](
   override protected def stageShutdown(): Unit = {
     logger.debug("Shutting down HttpPipeline")
     parser.synchronized {
+      cancel()
```
ChristopherDavenport: Does this signal cancelled even on correct shutdown?

rossabaker: I think it would.  We could blank it out, but I had to synchronize on the parser to make it reliable, and we'd have to think through the deadlock ramifications.

Alternatively: make the test pass without synchronization.

rossabaker: Detecting correct shutdown is going to be tricky here. Maybe it's more appropriate to cancel only when we receive a `Disconnected` message?

ChristopherDavenport:
Want to incorporate the Stream change?

rossabaker:
Which stream change?

rossabaker:
@ChristopherDavenport I'm still confused as to which stream change.

ChristopherDavenport:
That was from a conversation on gitter. I do not think it is relevant.

PR #2065, opened 2018-09-08
WIP: new ws interface and builder

jmcardon:


rossabaker:
Which is to say, I think this is on a good track, though I wonder whether `Close` should even be userland.

jmcardon:
I'll leave the PR open for the idea.

#2143 will wreck this but that's ok.

I do think `Close()` should be in userland. Not as a frame in itself, but as a sort of command saying "your conenction has ended", and to also close connections directly.

PR #2066, opened 2018-09-08
use more robust method of checking Scala version

SethTisue:
the old code was breaking in the Scala 2.13 community build, because
in that context we see Scala version numbers such as 2.13.0-pre-abcd123

it's not just that a "2.13" case was missing, it's also that the
`-pre-` indicates a binary incompatible Scala version, and such in a
case scalaBinaryVersion doesn't truncate to "2.13", but is the same as
scalaVersion

### DIFF
--- a/project/Http4sPlugin.scala
+++ b/project/Http4sPlugin.scala
@@ -271,10 +271,11 @@ object Http4sPlugin extends AutoPlugin {
     }
   }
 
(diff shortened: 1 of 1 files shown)


rossabaker:
#2067 passed and subtracts cruft, so I'd say that supersedes this.

We're going to need a similar scalacheck crossbuild once we can start publishing for Scala 2.13, so it is still helpful to have seen this done right.

rossabaker:
Superseded by #2066.

PR #2067, opened 2018-09-09
Use play-json-2.6.10 in all crossbuilds

rossabaker:
Solves problem of #2066, reduces complexity.

I think this is possible ever since jawn-1.13.0, but let's consult Travis CI.


rossabaker:
I will merge this if nobody else has vetoed it by morning, just because I don't want to cause the community built any more pain and it should be uncontroversial.

PR #2069, opened 2018-09-11
Renames incorrect, duplicate withIdleTimeout to withMaxTotalConnections

MarkusA380:
This fixes the name of a duplicate function `withIdleTimeout` that should actually be named `withMaxTotalConnections`. The implementation is already correct, and the real `def withIdleTimeout(idleTimeout: Duration): BlazeClientBuilder[F]` remains untouched.


aeons:
Build failure looks unrelated.

PR #2078, opened 2018-09-12
Make queue full exception specific

keirlawson:


ChristopherDavenport:
Weird place for the model though. Do we instead want it in an errors or in the companion object?

rossabaker:
`NoConnectionAllowedException` is already in the same place.  We could bikeshed that, but as with one, with the other.

SystemFw:
Is this ready for merge, or should I bring some paint? :)

rossabaker:
@keirlawson asked about 0.18 support.  I think we should cherry-pick rather than merge.

rossabaker:
Cherry picked onto release-0.18.x, merged back to master.

Issue #2071, opened 2018-09-12
Use the new `readRange` in fs2


rossabaker:
@asoltysik added a [`readRange`](https://github.com/functional-streams-for-scala/fs2/pull/1246) function to fs2.  We can replace our messy [`fileToBody`](https://github.com/http4s/http4s/blob/1c2e4c053b60c4cd978d11e87e8b60e94bede48e/core/src/main/scala/org/http4s/StaticFile.scala#L187) with it.

asarturas:
^ started working on this

aeons:
Fixed by #2079

PR #2079, opened 2018-09-12
#2071 use read range of fs2

asarturas:
Beware, this PR depends on change in fs2 (https://github.com/functional-streams-for-scala/fs2/pull/1257) and will fail the build until that one is merged


asarturas:
fixes #2071

PR #2080, opened 2018-09-12
Refactor Http4sMatchers to be generic on F-Type

### LINKED ISSUE #1430: `Http4sMatchers` isn't parameterised on Effect type
I've upgraded to 0.18.0-M1 but `Http4sMatchers` isn't parametrised on the effect type, so matchers like `haveStatus` don't work if I'm using something other than `cats.effect.IO` (eg: `cats.Id`)

diesalbla:
Resolves #1430.

As described in issue #1430, the `Http4sMatchers` trait is  not generic on the _effect_ `F[_]` type, and neither was a parent trait `IOMatchers` trait. We try to address this situation doing the following: 

- We factor out the trait `IOMatchers` into a `RunTimedMatchers` trait, which keeps most fo the IOMatchers trait methods and logic, but is generic on the effect type `F[_]`. Since IOMatchers was using some specific methods of `cats.effect.IO`, we need to extract those as abstract method, following the GoD template-method pattern. 
- We turn `IOMatchers` into an extension of `RunTimedMatchers` that specialises those abstract methods, to the methods of IO.
- We make `Http4sMatchers` generic on the F type. Luckily, this is just replacing "IO" with F

#### Other issues and related or future work. 

One thing to note is that, in the `RunTimedMatchers`, we need to declare `F`-generic abstractions for two methods in the `IO` class:
-  `unsafeRunSync`, which takes an `IO[A]` and returns an `A` (or throws an exception); and 
- `unsafeRunTimed`, which takes an `IO[A]` and a timeout, and either returns an `A` in time or throws a timeout error. 

These methods have no type-class capturing them, although an issue was opened in `cats-effect` that proposed creating a type-class that would have included the first method: https://github.com/typelevel/cats-effect/issues/230

It should be noted that, since the existing design is a lattice of inheritance and mixins, the solution only adds two more nodes to that lattice. While it may be possible to replace the design by one with less inheritance, which would avoid the inner (path-dependent) class `TimedMatcher`, that would exceed the scope of the original ticket IMHO.

### DIFF
--- a/testing/src/main/scala/org/http4s/testing/Http4sMatchers.scala
+++ b/testing/src/main/scala/org/http4s/testing/Http4sMatchers.scala
@@ -1,56 +1,56 @@
 package org.http4s
 package testing
 
+import cats.syntax.flatMap._
 import cats.data.EitherT
-import cats.effect.IO
 import org.http4s.headers._
 import org.specs2.matcher._
 
 /** This might be useful in a testkit spinoff.  Let's see what they do for us. */
 // TODO these akas might be all wrong.
-trait Http4sMatchers extends Matchers with IOMatchers {
-  def haveStatus(expected: Status): Matcher[Response[IO]] =
-    be_===(expected) ^^ { r: Response[IO] =>
+trait Http4sMatchers[F[_]] extends Matchers with RunTimedMatchers[F] {
+  def haveStatus(expected: Status): Matcher[Response[F]] =
+    be_===(expected) ^^ { r: Response[F] =>
       r.status.aka("the response status")
     }
 
-  def returnStatus(s: Status): Matcher[IO[Response[IO]]] =
-    haveStatus(s) ^^ { r: IO[Response[IO]] =>
-      r.unsafeRunSync.aka("the returned")
+  def returnStatus(s: Status): Matcher[F[Response[F]]] =
+    haveStatus(s) ^^ { r: F[Response[F]] =>
+      runAwait(r).aka("the returned")
     }
 
-  def haveBody[A: EntityDecoder[IO, ?]](a: ValueCheck[A]): Matcher[Message[IO]] =
-    returnValue(a) ^^ { m: Message[IO] =>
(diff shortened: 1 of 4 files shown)


rossabaker:
`IOMatchers` needs a `ContextShift`: https://travis-ci.org/http4s/http4s/jobs/427858664#L858

rossabaker:
I'd also like to have an http4s-laws (no specs2 dependency) and an http4s-specs2 module for the matchers.  Maybe that split could be coordinated with a scalacheck-1.14 upgrade and give us an excuse to adopt a specs2-owned trait, and we adopt these as a temporary fix.

diesalbla:
I agree that `IOMatchers`, should be removed from the `http4s` library in favour of `specs2`. Still, to achieve the goal of the issue and make `Http4sMatchers` generic, then we would also need the generic  `RunTimedMatchers` in `specs2`. I will make a PR to port these changes. 

As for the `scalacheck` dependency, in this PR we can just add an "excludeAll" in the import of `specs2`, and leave for another PR upgrading dependencies.

rossabaker:
I think the problem with `excludeAll` and upgrading to a new specs2 is that specs2's integration with scalacheck is in core, and will still be compiled against the binary incompatible 1.14.  We'd be in a much better position with cats-laws and cats-effect laws on 1.14.

diesalbla:
@rossabaker I have opened a PR in `specs2`, to port there some changes to `IOMatchers` suggested here. 

Looking through the build files of those projects, I have found a curious thing:  `cats` uses different versions of `scalacheck` [and other dependencies](https://github.com/typelevel/cats/blob/v1.4.0/build.sbt#L133-L143) , depending on which version of the Scala compiler it is building with. Unless you want to bring this kind of conditional dependencies to `http4s`, I do not see how to move forward. 

In any case, I humbly believe that upgrading test dependencies, and watching for any regression that may introduce, goes beyond the scope of the original issue.

Issue #2007, opened 2018-08-18
Origin header is a well defined triple


jmcardon:
Origin can be stated as:

```haskell
data Origin = Empty | Origin Scheme Host (Maybe Port)
```
Currently it's a `RawHeader`, which is wrong.

davegurnell:
I'll take a look at this one...

davegurnell:
I've been going through this with @hosamaly. We have a first draft of the header in this branch of my fork:

https://github.com/davegurnell/http4s/tree/topic/origin-header

Our parser breaks some tests in `CORSSpec`. These tests specify an Origin header with a trailing slash, which is technically not valid according to the spec:

http://tools.ietf.org/html/rfc6454#section-7

We *cooouuuld* take a permissive parsing approach, accept trailing slashes, and silently drop them. Or we could hold fast to the spec and edit the CORS tests. What do folks think?

(We also need to modify our code to accept a list of hosts in the Origin header, which is also technically part of the spec.)

rossabaker:
I think it would be reasonable to hold fast to the spec here and change the tests.

davegurnell:
Aces. In that case, here comes a PR...

PR #2082, opened 2018-09-13
Structured parsing of the Origin header (issue #2007)

davegurnell:
This PR improves the parsing of the `Origin` header. It's joint work with @hosamaly at Syntactic Sugar, London.

Instead of being interpreted as a string, the header is now interpreted as a list of items of type `Origin.Host`. These are similar to `Uris` except that Origin only permits a scheme, host, and port.

We commonly think of the Origin as containing a single host. According to the Mozilla Developer page, however, the header may be an empty string (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Origin). And according to the RFC, it may contain multiple hosts (https://tools.ietf.org/html/rfc6454#section-7).

There's one last complication. The RFC technically states that `Origin` may be the literal string `null`, which translates to an "opaque origin" ()

### DIFF
--- a/core/src/main/scala/org/http4s/headers/Origin.scala
+++ b/core/src/main/scala/org/http4s/headers/Origin.scala
@@ -1,4 +1,48 @@
 package org.http4s
 package headers
 
-object Origin extends HeaderKey.Default
+import cats.data.NonEmptyList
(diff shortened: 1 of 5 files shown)


davegurnell:
Ah! Thanks for the tip. Rerunning...

PR #2086, opened 2018-09-15
add Monoid instance for UrlForm

bpholt:
The tests I added are pretty rudimentary. I'm open to suggestions or even just a pointer to another Monoid instance's tests that would be good to emulate.

### DIFF
--- a/core/src/main/scala/org/http4s/UrlForm.scala
+++ b/core/src/main/scala/org/http4s/UrlForm.scala
@@ -106,6 +106,19 @@ object UrlForm {
     x.values.mapValues(_.toList).view.force === y.values.mapValues(_.toList).view.force
   }
 
+  implicit val monoidInstance: Monoid[UrlForm] = new Monoid[UrlForm] {
+    override def empty: UrlForm = UrlForm.empty
+
+    override def combine(x: UrlForm, y: UrlForm): UrlForm =
+      UrlForm(x.values.foldLeft(y.values) {
+        case (my, (k, x)) =>
+          my.updated(k, my.get(k) match {
+            case Some(y) => x ++ y
+            case None => x
+          })
+      })
+  }
+
   /** Attempt to decode the `String` to a [[UrlForm]] */
   def decodeString(charset: Charset)(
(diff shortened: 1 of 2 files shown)


bpholt:
> What are the expected semantics when both have the same key? Overwrite or combine? I would expect combine, which I think your implementation would do. Just double checking intent.

I expected the keys to be combined as well.

There is an `Arbitrary[UrlForm]`, and using cats-laws looks like it's revealed some bugs, so I'm working through those. Thanks for the suggestion.

core/src/main/scala/org/http4s/UrlForm.scala
```diff
@@ -106,6 +106,19 @@ object UrlForm {
     x.values.mapValues(_.toList).view.force === y.values.mapValues(_.toList).view.force
   }
 
+  implicit val monoidInstance: Monoid[UrlForm] = new Monoid[UrlForm] {
+    override def empty: UrlForm = UrlForm.empty
+
+    override def combine(x: UrlForm, y: UrlForm): UrlForm =
+      UrlForm(x.values.foldLeft(y.values) {
```
bpholt: I borrowed this implementation from [`MapMonoid`](https://github.com/typelevel/cats/blob/97052b8673afbcd0f8bb79dd7cc273d0e6049bdc/kernel/src/main/scala/cats/kernel/instances/map.scala#L61-L70), more or less inlining the implementation of `Semigroup.maybeCombine`, because [there's no `Monoid[Seq[?]]` instance](https://github.com/typelevel/cats/issues/1222) in cats. Open to alternatives.

bpholt:
I agree about reconsidering `Seq` later on. 

I think it would be good to backport this. I can prepare another PR for that if it's helpful—it looks like the tests pass if I cherry-pick f2af07d onto `release-0.18.x`.

rossabaker:
That sounds great, thanks.

rossabaker:
Is there anything left here we won't get from merging #2095?

bpholt:
I’m not sure if the cats wildcard import in the test file would make it via #2095, but I think that’s the only difference.

rossabaker:
Superseded by backport in #2095.

PR #2093, opened 2018-09-16
Record intermediate redirect locations

marvelm:
Hi!

I've added tracking redirect locations in the response attributes. I've also pulled the status code handling logic into a separate function.


marvelm:
@rossabaker Nice catch! Prepending the list threw me for a spin too and I only wrote the feature a few days ago 🤦‍♂️ . I've added a comment to remind the reader that function is recursive.

rossabaker:
The build is failing because we need to run `test:scalafmt`.

PR #2094, opened 2018-09-16
Common layer client metrics #2005

luisdeltoro:


dropwizard-client-metrics/src/main/scala/org/http4s/client/metrics/codahale/CodaHaleOps.scala
```diff
@@ -0,0 +1,74 @@
+package org.http4s.client.metrics.codahale
```
rossabaker: I thought the project had moved away from the "Coda Hale" branding at Coda's request.  Isn't it all in a dropwizard package these days?  I started using Prometheus more, so I haven't kept up.

project/Http4sPlugin.scala
```diff
@@ -307,8 +307,8 @@ object Http4sPlugin extends AutoPlugin {
   lazy val jspApi                           = "javax.servlet.jsp"      %  "javax.servlet.jsp-api"     % "2.3.3" // YourKit hack
   lazy val log4s                            = "org.log4s"              %% "log4s"                     % "1.6.1"
   lazy val logbackClassic                   = "ch.qos.logback"         %  "logback-classic"           % "1.2.3"
-  lazy val metricsCore                      = "io.dropwizard.metrics"  %  "metrics-core"              % "4.0.3"
-  lazy val metricsJson                      = "io.dropwizard.metrics"  %  "metrics-json"              % metricsCore.revision
+  lazy val dropwizardMetricsCore            = "io.dropwizard.metrics"  %  "metrics-core"              % "4.0.3"
```
rossabaker: These have always been alphabetical, which helps in merges.  If we're changing the name, let's move them, too.

dropwizard-client-metrics/src/main/scala/org/http4s/client/metrics/codahale/CodaHaleOps.scala
```diff
@@ -0,0 +1,74 @@
...
+        registry.counter(s"${namespace(prefix, destination)}.2xx-responses").inc()
+      case threehundreds if threehundreds < 400 =>
+        registry.counter(s"${namespace(prefix, destination)}.3xx-responses").inc()
+      case fourhundreds if fourhundreds < 500 =>
+        registry.counter(s"${namespace(prefix, destination)}.4xx-responses").inc()
+      case _ => registry.counter(s"${namespace(prefix, destination)}.5xx-responses").inc()
+    }
+}
+
+class CodaHaleOpsFactory extends MetricsOpsFactory[MetricRegistry] {
```
ChristopherDavenport: I don't like seeing the Factory pattern here, why not a function in the companion object?

dropwizard-client-metrics/src/main/scala/org/http4s/client/metrics/codahale/CodaHaleOps.scala
luisdeltoro: The code is still under codahale package, but the project is effectively called Dropwizard Metrics. I'll change that!

project/Http4sPlugin.scala
luisdeltoro: I didn't know that. I'll move them

rossabaker:
I am sure that I caused a horrible merge conflict between this and #2102.  If we can't get this one merged first, I will help with the merge conflict.

I'm going to make time to try this today.  I understand what it's trying to do and I understand @ChristopherDavenport's apprehension and I can't figure out a better way without trying it.

luisdeltoro:
@rossabaker I can take care of solving the merge conflicts if you like, so you do not need to worry about it.
But it would definitely be very helpful if you could try to find a better solution for the Factory Pattern thing

luisdeltoro:
@rossabaker I've rework the code to be compatible with the last changes in master

rossabaker:
I started playing with it a bit.

*  I'm thinking of `MetricsOps[F]` as an algebra as opposed to a typeclass.  Instead of a `DropwizardOps`, you have:

```scala
object Dropwizard {
  def apply[F[_]](reg: CollectorRegistry, prefix: String)(implicit F: Sync[F]): MetricsOps[F] = ???
}
```

Instantiation is then `Metrics(Dropwizard(reg, prefix))(client)`.  It's a little more verbose than summoning based on the type of registry, but you lose the `MetricsOpsFactory` boilerplate.  And you're doing it once per client.

* Could `MetricsOps` be generic across server and client?  The client logic is all in `Metrics` now.  The server logic could be in similar.  And then we'd only need one Prometheus and one Dropwizard module.  A lot of the `MetricsOps` would seem to be applicable two, and I can't imagine an implementation that could handle only the server or only the client.

luisdeltoro:
https://github.com/http4s/http4s/issues/2131

luisdeltoro:
#2131

client-metrics/src/main/scala/org/http4s/client/metrics/core/MetricsOps.scala
```diff
@@ -4,9 +4,8 @@ import cats.effect.Sync
 import org.http4s.Status
 
 /**
-  * Trait for objects capable of writing metrics to a metrics registry
+  * Describes an algebra capable of writing metrics to a metrics registry
```
rossabaker: @SystemFw likes algebras to have an implicit summoner with explicit instances, and I'm coming around to that view myself.

How about also:

```scala
object MetricsOps {
  def apply[F[_]](implicit ev: MetricsOps[F]) = ev
}
```

client-metrics/src/main/scala/org/http4s/client/metrics/core/MetricsOps.scala
```diff
@@ -0,0 +1,75 @@
...
+    * @param classifier the classifier to use
+    */
+  def increaseTimeouts(classifier: Option[String]): F[Unit]
+}
+
+/**
+  * Type class that allows to create a [[MetricsOps]] object for certain metrics registry R
+  * @tparam R the specific metrics registry to use
+  */
+trait MetricsOpsFactory[R] {
```
rossabaker: Is this trait still useful?

client-metrics/src/main/scala/org/http4s/client/metrics/core/MetricsOps.scala
```diff
@@ -0,0 +1,75 @@
+package org.http4s.client.metrics.core
```
rossabaker: If we're going the route of making this common, this needs to move out of client.

client-metrics/src/main/scala/org/http4s/client/metrics/core/MetricsOps.scala
```diff
@@ -0,0 +1,75 @@
...
+  *
+  */
+trait MetricsOps[F[_]] {
+
+  /**
+    * Increases the count of active requests
+    *
+    * @param classifier the request classifier
+    */
+  def increaseActiveRequests(classifier: Option[String]): F[Unit]
```
rossabaker: Should these classifiers get a newtype, or are we happy with Strings? :thinking:

client-metrics/src/main/scala/org/http4s/client/metrics/core/Metrics.scala
```diff
@@ -0,0 +1,62 @@
...
+    * @param client the [[Client]] to gather metrics from
+    * @return the metrics middleware wrapping the [[Client]]
+    */
+  def apply[F[_]](ops: MetricsOps[F], classifier: Request[F] => Option[String] = { _: Request[F] =>
+    None
+  })(client: Client[F])(implicit F: Sync[F], clock: Clock[F]): Client[F] = {
+
+    def withMetrics(req: Request[F]): Resource[F, Response[F]] =
+      (for {
+        start <- Resource.liftF(clock.monotonic(TimeUnit.NANOSECONDS))
```
rossabaker: I'm a little worried about overhead of `liftF` and `flatMap` on something that runs on every request.  It's probably insignificant, but we could probably optimize this later by fusing some of the `F`s before lifting into something heavier.  I wouldn't sweat it for this PR, because the API is at least correct.

client-metrics/src/main/scala/org/http4s/client/metrics/core/MetricsOps.scala
luisdeltoro: What is the advantage of this? Having the actual implementation of MetricsOps being defined by the implicits in scope?

client-metrics/src/main/scala/org/http4s/client/metrics/core/MetricsOps.scala
luisdeltoro: Ups, sorry that a leftover. I'll remove it

client-metrics/src/main/scala/org/http4s/client/metrics/core/MetricsOps.scala
luisdeltoro: Yes, that's true. Where should we put it? in core?

client-metrics/src/main/scala/org/http4s/client/metrics/core/Metrics.scala
luisdeltoro: Humm. I didn't about the overhead, sorry for that.
At the beginning I was lifting the Resource into the F context but then I was having problems intercepting the errors inside the Resource. But I can try to give another thought if it's important.

client-metrics/src/main/scala/org/http4s/client/metrics/core/MetricsOps.scala
luisdeltoro: I'm trying to get my head around this, but I still not understanding very well.
How would this work?
```
implicit val dropwizardMetricsOps = Dropwizard(registry, { _: Request[F] => Some("classifier")})
val metricsOps = MetricsOps()
```
I doesn't make much sense to me right now...  thinking

rossabaker: It makes sense for a lot of algebra interpreters, but I think you're right.  It doesn't make much sense here.  Ignore this one.

client-metrics/src/main/scala/org/http4s/client/metrics/core/MetricsOps.scala
rossabaker: Yeah, how about in http4s-core, in `org.http4s.metrics`?  It doesn't pull in a new dependency, so I don't see it going into its own module.

client-metrics/src/main/scala/org/http4s/client/metrics/core/MetricsOps.scala
rossabaker: Yeah, just to avoid the problem of Everything Is a String.  Maybe it's more annoying than anything.  I could go either way.  I'll trust your judgement.

client-metrics/src/main/scala/org/http4s/client/metrics/core/Metrics.scala
rossabaker: fs2-1.0.0 is almost out, and this is a breaking change.  I'd rather get this into our next milestone, and optimize later if we're so inclined.  Let's focus on the other issues.

client-metrics/src/main/scala/org/http4s/client/metrics/core/MetricsOps.scala
luisdeltoro: sure. I'll move it there

client-metrics/src/main/scala/org/http4s/client/metrics/core/Metrics.scala
luisdeltoro: ok, let's do that

client-metrics/src/main/scala/org/http4s/client/metrics/core/MetricsOps.scala
luisdeltoro: I see your point. But it doesn't look very problematic to me in this particular case, so I would be more inclined to keep it simple and leave like it is if that's fine with you.

luisdeltoro:
Should I add some documentation about the metrics middleware in the docs project?

rossabaker:
Yes, that would be great.

rossabaker:
Build failure is #2145.

luisdeltoro:
I've added the documentation, but unfortunately I had to use "tut:nofail:silent" because I couldn't get the code examples to compile, although they work correctly in my REPL. The client metrics projects seem to be missing in the classpath or something like that.
I'm not very familiar with tut, so maybe you could point to where to look to fix that...

rossabaker:
The docs project needs to [depend on](https://github.com/luisdeltoro/http4s/blob/CommonLayerClientMetrics_%232005/build.sbt#L405) all the projects being documented.

luisdeltoro:
Thanks very much for that. It was really useful, now compilation works.

build.sbt
```diff
@@ -84,8 +84,10 @@ lazy val prometheusClientMetrics = libraryProject("prometheus-client-metrics")
     description := "Support for Prometheus Metrics on the client",
     libraryDependencies += prometheusClient
   )
-  .dependsOn(client % "compile;test->test")
-
+  .dependsOn(
+    client % "compile;test->test",
+    clientMetrics % "compile"
```
rossabaker: Does this project depend on anything other than core now?  I don't think it does.  And then maybe the package name is `org.http4s.metrics.prometheus`.  Then it's a non-breaking change when the server implementation is made common.

build.sbt
```diff
@@ -99,16 +101,21 @@ lazy val client = libraryProject("client")
     scalaXml % "test->compile")
 
 lazy val clientMetrics = libraryProject("client-metrics")
+  .settings(
```
rossabaker: This one doesn't add any new dependencies.  I think this could live directly in the client project, with the implementations in their own modules.

build.sbt
luisdeltoro: Sure. I'll do that.

build.sbt
luisdeltoro: Makes sense yes.

PR #2098, opened 2018-09-19
Implicit jsonOf and jsonEncoderOf

joan38:
I'm not sure why this is not already implicit.
Anyway let me know if there is any reason.

This is also making the API naming consistent across decoder and encoder.

Thanks

### DIFF
--- a/circe/src/main/scala/org/http4s/circe/CirceInstances.scala
+++ b/circe/src/main/scala/org/http4s/circe/CirceInstances.scala
@@ -44,8 +44,12 @@ trait CirceInstances {
         case _ => jawn.jawnDecoderImpl[F, Json](msg)
       }
     }
+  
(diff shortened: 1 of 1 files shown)


ChristopherDavenport:
Thank you for the contribution! 

These already exist as implicits at [CirceEntityDecoder](https://github.com/http4s/http4s/blob/master/circe/src/main/scala/org/http4s/circe/CirceEntityDecoder.scala). The problem exists for types like string which many do not want to automatically become json. While this something that we want to have available, it is not in general a great default.

However the effort taken to make these tools that many want as a default available is very appreciated. 

General Question: Where could we put more documentation to make this option more visible?

joan38:
This is true. I felt like there was a reason :)
Thanks for the hint!

I guess the documentation on https://http4s.org/v0.19/json/ is quiet lite and I guess it would go well there.

PR #2102, opened 2018-09-22
Refactor client into algebra

rossabaker:
* `DisposableResponse` is gone
* `Client` is a trait with no concrete methods
* `DefaultClient` implements them all in terms of a `run: Request[F] => Resource[F, Response[F]]`
* Client no longer has a `shutdown`: we create clients as `Resource`s.


client/src/main/scala/org/http4s/client/DefaultClient.scala
```diff
@@ -0,0 +1,238 @@
...
+    *
+    * This is intended for use in proxy servers.  `fetch`, `fetchAs`,
+    * [[toKleisli]], and [[streaming]] are safer alternatives, as their
+    * signatures guarantee disposal of the HTTP connection.
+    */
+  def toHttpApp: HttpApp[F] = Kleisli { req =>
+    def go[A](r: Resource[F, A]): F[(A, F[Unit])] = r match {
+      case Resource.Allocate(r) =>
+        r.map { case (a, release) => (a, release(ExitCase.Completed)) }
+      case Resource.Bind(r, f) =>
```
rossabaker: I am sleepy and this feels dangerous.  Please be extra careful reviewing this.

rossabaker:
I have not tried the metrics, I have only compiled them.

This is going to cause a merge conflict with #2094, so we need to give one of them the right of way.

rossabaker:
Similar to, and supersedes, #1772.

client/src/main/scala/org/http4s/client/DefaultClient.scala
ChristopherDavenport: 😟

rossabaker: It probably pays to go back and understand why `Resource` has this ADT instead of being a wrapper around `Allocate`'s value, because we're basically reconstructing that.

(This is also a suggestion for myself.)

rossabaker: Specifically, https://github.com/typelevel/cats-effect/pull/278

client/src/main/scala/org/http4s/client/DefaultClient.scala
```diff
@@ -0,0 +1,261 @@
...
+        current: Resource[F, Any],
+        stack: List[Any => Resource[F, Any]],
+        release: F[Unit]): F[(Any, F[Unit])] =
+      current match {
+        case Resource.Allocate(resource) =>
+          F.bracketCase(resource) {
+            case (a, rel) =>
+              stack match {
+                case Nil => F.pure(a -> release.guarantee(rel(ExitCase.Completed)))
+                case f0 :: xs => continue(f0(a), xs, release.guarantee(rel(ExitCase.Completed)))
```
rossabaker: We release in reverse order, so I think these guarantees are both backward?  We don't yet have a test that exposes this.

PR #2104, opened 2018-09-23
Refactor server creation into a resource

rossabaker:
Similar to #2102, but for servers.  Removes the `shutdown` method, preferring to create `Resource`s or `Stream`s

This presents a challenge in tut, because we naughtily used the separate `.start` and `.shutdown` in that unbracketed context.  We might simulate it with `.serveWhile`.


aeons:
This and the client builder algebra look like really nice, clean solutions.

rossabaker:
Thanks.  I think it cleans up a lot now that we have `Resource`.

PR #2106, opened 2018-09-24
Don't start response until after we add async timeout listener

### LINKED ISSUE #2103: Exception on timeout listener registration
I was benchmarking http4s with Monix and Jetty and i've noticed multiple errors in logs with following exception:
```
java.lang.IllegalStateException: AsyncContext completed and/or Request lifecycle recycled
	at org.eclipse.jetty.server.AsyncContextState.state(AsyncContextState.java:54)
	at org.eclipse.jetty.server.AsyncContextState.addListener(AsyncContextState.java:93)
	at org.http4s.servlet.AsyncHttp4sServlet.$anonfun$handleRequest$1(AsyncHttp4sServlet.scala:62)
	at org.http4s.servlet.AsyncHttp4sServlet.$anonfun$handleRequest$1$adapted(AsyncHttp4sServlet.scala:61)
	at monix.eval.internal.TaskCreate$.$anonfun$async$1(TaskCreate.scala:116)
	at monix.eval.internal.TaskCreate$.$anonfun$async$1$adapted(TaskCreate.scala:113)
	at monix.eval.internal.TaskRestartCallback.start(TaskRestartCallback.scala:57)
	at monix.eval.internal.TaskRunLoop$.executeAsyncTask(TaskRunLoop.scala:550)
	at monix.eval.internal.TaskRunLoop$.startFull(TaskRunLoop.scala:114)
	at monix.eval.internal.TaskRunLoop$.$anonfun$restartAsync$1(TaskRunLoop.scala:181)
	at java.base/java.util.concurrent.ForkJoinTask$RunnableExecuteAction.exec(ForkJoinTask.java:1426)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:290)
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1603)
	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:177)
```

code:
```scala
object ApplicationHttp4s extends IOApp {
  implicit val scheduler: Scheduler = Scheduler.global

  def routes[F[_]: Sync] = new Http4sDsl[F] {
    val routes = HttpRoutes.of[F] {
      case GET -> Root / "api" / "index" =>
        Ok("Hello!")
    }
  }
  
  def stream[F[_]: ConcurrentEffect]: fs2.Stream[F, ExitCode] =
    JettyBuilder[F]
      .bindHttp(10080, "0.0.0.0")
      .mountService(routes[F].routes, "/")
      .serve

    stream[Task].compile.drain.as(ExitCode.Success).toIO
}
```

monix version - `3.0.0-RC2-d0feeba`
http4 version - `0.19.0-M2`
jdk version - `10.0.2`

it reproduces with `wrk -t1 -c1` benchmark configuration and it seems doesn't affect request itself but might slightly affect performance due throw.

rossabaker:
It is an error to add an async context listener after a context is completed.  We need to install the timeout listener before we race against the response.

The unit test passed before the fix, but shows that we don't break the basic functionality.  The error condition is rendered on an asynchronous, container owned thread, and I don't see that we have access to it.  Furthermore, it only happens under load.

Fixes #2103.

### DIFF
--- a/servlet/src/main/scala/org/http4s/servlet/AsyncHttp4sServlet.scala
+++ b/servlet/src/main/scala/org/http4s/servlet/AsyncHttp4sServlet.scala
@@ -3,6 +3,7 @@ package servlet
 
 import cats.data.OptionT
 import cats.effect._
+import cats.effect.concurrent.Deferred
 import cats.implicits.{catsSyntaxEither => _, _}
 import javax.servlet._
 import javax.servlet.http.{HttpServletRequest, HttpServletResponse}
@@ -57,22 +58,20 @@ class AsyncHttp4sServlet[F[_]](
   private def handleRequest(
       ctx: AsyncContext,
       request: Request[F],
-      bodyWriter: BodyWriter[F]): F[Unit] = {
-    val timeout = F.async[Unit] { cb =>
-      ctx.addListener(new AsyncTimeoutHandler(cb))
-    }
+      bodyWriter: BodyWriter[F]): F[Unit] = Deferred[F, Unit].flatMap { gate =>
+    // It is an error to add a listener to an async context that is
+    // already completed, so we must take care to add the listener
+    // before the response can complete.
+    val timeout =
+      F.asyncF[Response[F]](cb => gate.complete(ctx.addListener(new AsyncTimeoutHandler(cb))))
     val response =
-      optionTSync
-        .suspend(serviceFn(request))
-        .getOrElse(Response.notFound)
-        .recoverWith(serviceErrorHandler(request))
+      gate.get *>
+        optionTSync
+          .suspend(serviceFn(request))
+          .getOrElse(Response.notFound)
+          .recoverWith(serviceErrorHandler(request))
     val servletResponse = ctx.getResponse.asInstanceOf[HttpServletResponse]
-    F.race(timeout, response).flatMap {
-      case Left(()) =>
-        renderResponse(Response.timeout[F], servletResponse, bodyWriter, F.never)
-      case Right(resp) =>
-        renderResponse(resp, servletResponse, bodyWriter, timeout)
-    }
+    F.race(timeout, response).flatMap(r => renderResponse(r.merge, servletResponse, bodyWriter))
   }
 
   private def errorHandler(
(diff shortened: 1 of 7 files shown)


rossabaker:
This is a Travis-only failure.  It passes locally.

rossabaker:
Appveyor error is unrelated.

servlet/src/main/scala/org/http4s/servlet/AsyncHttp4sServlet.scala
```diff
@@ -57,22 +58,20 @@ class AsyncHttp4sServlet[F[_]](
...
+    // already completed, so we must take care to add the listener
+    // before the response can complete.
+    val timeout =
+      F.asyncF[Response[F]](cb => gate.complete(ctx.addListener(new AsyncTimeoutHandler(cb))))
     val response =
-      optionTSync
-        .suspend(serviceFn(request))
-        .getOrElse(Response.notFound)
-        .recoverWith(serviceErrorHandler(request))
+      gate.get *>
```
YulawOne: Sorry, maybe I'm missing something, but I don't see difference between race with synchronization on gate  and plain flatMap i.e. `timeout *> response` because we will wait `gate.get` completion

rossabaker: This makes `response` wait on the call to `ctx.addListener`, not the asynchronous completion of `timeout`.  The original bug was because `response` could complete `ctx` before `ctx.addListener` was called.  And `timeout *> response` would wait for the timeout to fire before attempting `response`.

YulawOne: Sorry, my brain have completely ignored part with `F.asyncF[Response[F]](cb =>...)`, and now with your explanation i've finally understood whole picture. Thanks for explanation!

rossabaker: No problem.  Thanks for the bug report!  This should be released in 0.19.0-M3 as soon as the fs2 release candidate is out.

PR #2113, opened 2018-09-25
Enable unusedCompileDependenciesTest

rossabaker:
Fails the build if we declare a dependency we don't use.  Prevents us from passing liabilities downstream.

Not enforced on private (e.g. unpublished) projects.

Bonus: stop publishing empty http4s-tests module.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -317,7 +317,8 @@ lazy val bench = http4sProject("bench")
   .enablePlugins(PrivateProjectPlugin)
   .settings(
     description := "Benchmarks for http4s",
-    libraryDependencies += circeParser
+    libraryDependencies += circeParser,
+    unusedCompileDependenciesTest := {},
   )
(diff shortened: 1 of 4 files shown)


cb372:
Looks like it fails for only the primary Travis job, where [scoverage is enabled](https://github.com/http4s/http4s/blob/e008e45ba97425239a56686b2648396fde601d01/project/Http4sPlugin.scala#L60):

```
[warn] mimedb-generator >>> The following libraries are declared in libraryDependencies but are not needed for compilation:
[warn]  - "org.scoverage" %% "scalac-scoverage-runtime" % "1.3.1"
```

Not sure how to work around that, other than adding support in the plugin for manually excluding certain dependencies as I suggested at https://github.com/cb372/sbt-explicit-dependencies/issues/13#issuecomment-424238479. I'll try and get that implemented in the next few days.

PR #2117, opened 2018-09-25
Handle unsuccessful responses in JavaNetClient

### LINKED ISSUE #2114: JavaNetClientBuilder fails with NoContent response
Calling delete on a service that returns NoContent results in an exception when using JavaNetClientBuilder.
The call is 
```
httpClient.status(Request[IO](Method.DELETE, uri))
```
The exception I get is 
```
Caused by: java.io.FileNotFoundException: http://localhost:8080/history/98613681-91f2-4f42-ac23-655de09178b3
        ...
	at sun.net.www.protocol.http.HttpURLConnection.getInputStream0(HttpURLConnection.java:1490)
	at sun.net.www.protocol.http.HttpURLConnection.getInputStream(HttpURLConnection.java:1474)
	at org.http4s.client.JavaNetClientBuilder.$anonfun$open$21(JavaNetClientBuilder.scala:120)
       ...
Caused by: java.io.FileNotFoundException: http://localhost:8080/history/98613681-91f2-4f42-ac23-655de09178b3
	at sun.net.www.protocol.http.HttpURLConnection.getInputStream0(HttpURLConnection.java:1872)
	at sun.net.www.protocol.http.HttpURLConnection.getInputStream(HttpURLConnection.java:1474)
	at java.net.HttpURLConnection.getResponseCode(HttpURLConnection.java:480)
	at org.http4s.client.JavaNetClientBuilder.$anonfun$fetchResponse$2(JavaNetClientBuilder.scala:128)
```
Note that the call succeeds using BlazeClientBuilder

rossabaker:
`HttpUrlConnection` returns the body in `getErrorStream` rather than `getInputStream` when the response code is unsuccessful.  Well, sometimes.  Sometimes it returns `null`.

1. Try to read the body from `getInputStream`
2. If that throws, try to read an error from `getErrorStream`, if it's there.
3. Otherwise, return an empty body.
4. Close whatever we can find to close on dispose.

Fixes #2114.

### DIFF
--- a/client/src/main/scala/org/http4s/client/JavaNetClient.scala
+++ b/client/src/main/scala/org/http4s/client/JavaNetClient.scala
@@ -4,6 +4,8 @@ package client
 import cats.data.Kleisli
 import cats.effect.{Effect, Sync}
 import cats.implicits._
+import fs2.Stream
+import java.io.IOException
 import java.net.{HttpURLConnection, Proxy, URL}
 import javax.net.ssl.{HostnameVerifier, HttpsURLConnection, SSLSocketFactory}
 import org.http4s.internal.{blocking, readInputStream, writeOutputStream}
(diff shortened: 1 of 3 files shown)


rossabaker:
These new tests reliably deadlock Travis.  They pass locally.  I'm still unclear where they're deadlocking Travis, but all the new unsafeRunSync`ing can't help.

rossabaker:
I think this was the cause of our intermittent client failures that have plagued us for, like, ever.

PR #2118, opened 2018-09-25
Mark case classes final

### LINKED ISSUE #2074: Finalize or seal all case classes
Look for case classes that aren't `final` or `sealed abstract` (a technique to hide undesired `apply` and `copy` methods), and make them so.  There is no reason to allow inheritance of case classes.

richashworth:
Co-authored-by: Andrea Magnorsky <roundcrisis@gmail.com>

Adds the `final` modifier to case classes that are not already `final` or `sealed abstract`. Exceptions listed below:

- `TimedMatcher` in `/testing/src/main/scala/org/http4s/testing/RunTimedMatchers.scala`
- Anything under a `src/test` path

Adding the `final` or `sealed abstract` modifiers to the following case classes fails to compile, so these have also been excluded from this refactor:
- `/core/src/main/scala/org/http4s/parser/StrictTransportSecurityParser.scala:  private case class StrictTransportSecurityParser(override val input: ParserInput)` 
- `./examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/endpoints/JsonXmlHttpEndpoint.scala:  case class Person(name: String, age: Int)`

Fixes #2074

### DIFF
--- a/client-metrics/src/main/scala/org/http4s/client/metrics/Metrics.scala
+++ b/client-metrics/src/main/scala/org/http4s/client/metrics/Metrics.scala
@@ -64,7 +64,7 @@ object Metrics {
 
 }
 
-private case class MetricsCollection(
+private final case class MetricsCollection(
     activeRequests: Counter,
     requestsHeaders: MetricTimer,
     requestsTotal: MetricTimer,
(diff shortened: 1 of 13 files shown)


rossabaker:
Wonderful, thank you!  Do you happen to remember the nature of the compile errors on the others?  If not, I can reproduce easily enough.

PR #2127, opened 2018-09-27
Remove MiMa opt-outs for recent modules

rossabaker:
When we add a new module in the middle of a series, we have to disable the `mimaPreviousArtifacts`.  We neglected to clean these up later.


rossabaker:
Oops, this was intended for release-0.18.x.

PR #2130, opened 2018-09-27
Upgrade scala and sbt

aeons:


examples/src/main/scala/com/example/http4s/ScienceExperiments.scala
```diff
@@ -110,7 +110,7 @@ class ScienceExperiments[F[_]] extends Http4sDsl[F] {
           Stream
             .eval(F.pure(Chunk.bytes(Array(' '.toByte))))
             .evalMap(_ =>
-              F.async[Byte] { cb => /* hang */
+              F.async[Byte] { _ => /* hang */
```
rossabaker: This could be `F.never` now.

aeons: Also on 0.18?

aeons:
Yea, looks like it.

PR #2138, opened 2018-09-30
WebSocketClient

rossabaker:
A very rudimentary proof of concept.


client/src/main/scala/org/http4s/client/WebSocketClient.scala
```diff
@@ -0,0 +1,33 @@
...
+}
+
+object WebSocketClient {
+  def apply[F[_]](f: Request[F] => F[WebSocket[F]]): WebSocketClient[F] =
+    new WebSocketClient[F] {
+      def connect(req: Request[F]) = f(req)
+    }
+}
+
+abstract class WebSocket[F[_]] {
```
rossabaker: This needs to be unified with the one in core.

Cue standard trait vs. abstract class debate.

client/src/main/scala/org/http4s/client/WebSocketClient.scala
```diff
@@ -0,0 +1,33 @@
+package org.http4s
+package client
+
+import fs2.{Sink, Stream}
+import java.net.SocketAddress
+import org.http4s.websocket.WebsocketBits.WebSocketFrame
+
+trait WebSocketClient[F[_]] {
```
rossabaker: I imagine we'd add more methods here for higher level (i.e., ponging, closing, aggregating fragments).

async-http-client/src/main/scala/org/http4s/client/asynchttpclient/AsyncHttpClient.scala
```diff
@@ -139,4 +148,80 @@ object AsyncHttpClient {
...
     }.toList)
+
+  def webSocketResource[F[_]](config: AsyncHttpClientConfig = defaultConfig)(
+      implicit F: ConcurrentEffect[F]): Resource[F, WebSocketClient[F]] =
+    Resource
+      .make(F.delay(new DefaultAsyncHttpClient(config)))(c => F.delay(c.close()))
+      .map(client =>
+        WebSocketClient[F] { req =>
+          for {
+            receiveQueue <- Queue.unbounded[F, WebSocketFrame]
```
rossabaker: This is not ideal but I don't see a way to apply backpressure short of blocking the listener.

async-http-client/src/main/scala/org/http4s/client/asynchttpclient/AsyncHttpClient.scala
```diff
@@ -139,4 +144,72 @@ object AsyncHttpClient {
...
+      }
+
+      def enqueue(frame: WebSocketFrame): Unit =
+        invoke(logger)(receiveQueue.enqueue1(frame))
+
+      def onClose(ahcSocket: AhcWebSocket, code: Int, reason: String) =
+        Close(code, reason).foreach(enqueue)
+
+      def onError(t: Throwable) =
+        cb(Left(t))
```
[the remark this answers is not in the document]
benhutchison: Hi @rossabaker 

I've been studying your web socket PR as I'm keen to use the client once it's ready. As well as educating me on how http4s works, I have an [existing case where I use the Akka WS client with a Http4s server](https://github.com/benhutchison/http4s-factor/blob/master/integrationTest/src/multi-jvm/scala/factor/http4s/integrationtest/FactorHttp4sWebsocketTest.scala#L21) that I'd eventually like to swap out. Generally the PR looks good and makes sense to me.

Re the point about needing to signal errors in `onError` handler, agree it is crucial. In my testing if a WS server "goes away", it shows up as an error to the client, and they need to close off the socket and signal to app.

So I wonder if the `WebSocketFrame` ADT should include a `FrameError` case? The other alternative is to return a stream of `Either[FrameError, WebSocketFrame]`. Both seem reasonable choices, I never know quite which to pick as they both make tradeoffs ...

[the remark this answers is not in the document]
benhutchison: Further thoughts on error handling: many cases in the RFC Close the WS in response to errors, eg

- [Invalid mask in frames](https://tools.ietf.org/html/rfc6455#section-5.1)
- [Underlying transport lost](https://tools.ietf.org/html/rfc6455#section-7.2.1)
- [Invalid UTF8 encoding](https://tools.ietf.org/html/rfc6455#section-8.1)

Of course, the Throwable passed to `onError` could also be a JVM/OS problem like memory or classfile problem.

But turning errors into `Close` events seems like a reasonable policy to me

benhutchison:
FYI Im planning to maintain a [Scalajs/DOM based WS client with a similar API](https://github.com/benhutchison/scalajs-websocket) to the one you're building here. Having the same "shape" API whether on JVM or JS platforms will help migrate code into the browser and generally make dealing with WSs in Scala easier

joan38:
Hey! Any news on continuing this?

amesgen:
@joan38 We have a websocket client in https://github.com/http4s/http4s-jdk-http-client based on the client in Java 11+. That's the current step towards a websocket client story in http4s, and we are looking for feedback!

rossabaker:
@amesgen I think we should proceed with a PR to the main repo based on the one in http4s-jdk-http-client, and get a second implementation, and see what can be reconciled with the server side.  Lots of people want this, and we've learned a lot since we first started discussing this.

amesgen:
@rossabaker I will try to start working on this on the weekend.

joan38:
Exciting times!

armanbilge:
Superseded by https://github.com/http4s/http4s/pull/5520.

PR #2139, opened 2018-09-30
Reinstate examples

### LINKED ISSUE #2107: Reinstate example projects
We urgently need to get those cleaned up and compiling again.

- Should build on what we learned in http4s/http4s.g8
- Should use less inheritance and look more like independent projects so people see how it fits together.

aeons:
Fixes #2107 

Updated and re-enabled all the example projects.
There's less inheritance now, and everything should use `F[_]` all the way up to `IOApp`.

The examples are not 100% consistent with each other.

I updated the push support syntax to operate on `Response[F]` instead of `F[Response[F]]`, to be in line with the syntax on response and request. I did not add a deprecated copy of the old behaviour. I can add that if you guys want it.


examples/blaze/src/main/scala/com/example/http4s/blaze/ClientMultipartPostExample.scala
```diff
@@ -37,6 +38,6 @@ object ClientMultipartPostExample extends IOApp with Http4sClientDsl[IO] {
   def run(args: List[String]): IO[ExitCode] =
     BlazeClientBuilder[IO](global).resource
       .use(go)
-      .flatMap(s => IO(println))
+      .flatMap(s => IO(println(s)))
```
rossabaker: :flushed:

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/service/GitHubService.scala
```diff
@@ -19,7 +19,7 @@ class GitHubService[F[_]: Sync](client: Client[F]) extends Http4sClientDsl[F] {
 
   private val RedirectUri = s"http://localhost:8080/$ApiVersion/login/github"
 
-  final case class AccessTokenResponse(access_token: String)
+  case class AccessTokenResponse(access_token: String)
```
rossabaker: I like the final on case classes.  It doesn't matter much here, but it's a good habit.

server/src/main/scala/org/http4s/server/middleware/PushSupport.scala
```diff
@@ -10,31 +10,33 @@ import org.log4s.getLogger
...
-            if (!url.startsWith("/")) sb.append('/')
-            sb.append(url)
-              .result()
-          } else url
-        }
+  implicit def http4sPushOps[F[_]: Functor](response: Response[F]): PushOps[F] =
+    new PushOps[F](response)
 
-        logger.trace(s"Adding push resource: $newUrl")
+  final class PushOps[F[_]: Functor](response: Response[F]) extends AnyRef {
```
rossabaker: Yes, this change makes a lot more sense.

examples/blaze/src/main/scala/com/example/http4s/blaze/demo/server/service/GitHubService.scala
aeons: Yeah, me too, but this one doesn't compile if it's final :)

PR #2140, opened 2018-10-01
Test and fix backpressure of client request bodies

rossabaker:
A client should not buffer the entire request body waiting on a slow server.

* async-http-client: fine
* blaze: fixed by starting the request rendering on a fiber. Could still be smarter about cancellation.
* jetty: fine
* okhttp: broken. The response listener doesn't wait on the `RequestBody`
* java.net client: broken. It throws if we read before we've written all the bytes.

We can't do much about socket buffering.  There is not a one-to-one correspondence between a server demanding N bytes and a client fulfilling it.  But this exposes naive buffering, and fixes it in blaze.

/cc @lbialy, who reported this on blaze

Bonus feature: honor a timeout fired while a read or write is pending


lbialy:
Cool! I'm wondering how it's possible that AHC is handling this fine though? You've set the buffer size to 64K and I'm quite sure that my backend, while it was taking 3 elements, was pulling considerably smaller package of data and AHC client obliged with just that. I've just measured and around 20 elements is ~11K (from file size via `ls -hal`).

rossabaker:
When I sleep for long enough the numbers stop going up, I'm seeing blaze, async-http-client, and jetty-client all sending a little over 3000 1KB chunks.  Note that this is POSTing to a Jetty sevlet that never reads, which is different from your scenario.  And if you respond quickly, it doesn't buffer that much.  System settings, which may be relevant:

```
/proc/sys/net/ipv4/tcp_rmem
4096	87380	6291456
/proc/sys/net/ipv4/tcp_wmem
4096	16384	4194304
```

I set the threshold to 64MB, which is ridiculously generous, but I didn't want to create flaky tests on Travis.  It appears that I failed that mission.  I don't think I even need to sleep: the broken backends were buffering the entire response in this scenario even without sleeping.

rossabaker:
This turned into a terrible mess.  Reviving in a separate PR.

PR #2143, opened 2018-10-02
Unify org.http4s.websocket package, use ByteVector

rossabaker:
The `org.http4s.websocket` package is awkwardly split across the http4s-websocket repo and http4s-core.  No package should ever live in two artifacts.

The reason for the http4s-websocket repo was to share a model between blaze and http4s without adding any dependencies to the former.  Since then, the blaze repo dropped websocket support entirely.  If we unify it here, we can use a model with all our dependencies.  This includes `ByteVector`, which is back on our classpath since fs2-core again depends on scodec-bits.  That gives us an immutable model.

There are all kinds of shady numeric truncations and widenings going on here, and it's worth looking closely at the `WebSocketFrame` class.

Additionally, standardizes on the `WebSocket` capitalization in place of `Websocket`

### DIFF
--- a/blaze-core/src/main/scala/org/http4s/blazecore/websocket/Http4sWSStage.scala
+++ b/blaze-core/src/main/scala/org/http4s/blazecore/websocket/Http4sWSStage.scala
@@ -7,16 +7,16 @@ import cats.implicits._
 import fs2._
 import fs2.concurrent.SignallingRef
 import java.util.concurrent.atomic.AtomicBoolean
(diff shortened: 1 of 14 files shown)


cquiroz:
As a concept i totally support this

rossabaker:
We should also release a blaze milestone without the old dependency to ensure that we're no longer using it.

jmcardon:
whoops, double approval was accident.

PR #2146, opened 2018-10-02
Remove outbound commands, get on blaze master

rossabaker:
This gets us working again off blaze master, so we can release that as blaze-0.14.0 before http4s-0.19.0.


rossabaker:
This build will fail, but I think it will go green after blaze-0.14.0-M5 hits Maven Central.

I'll be alseep by then, in case someone wants to restart it.

PR #2147, opened 2018-10-02
Fix 0.19 link in 0.18 docs

rossabaker:
Pointed out by @Tomczik76

Ideally, older versions of the docs would pull this from a common place, so we don't need to go back and republish old versions when new series are published.


rossabaker:
Yes, I think in master.  Which would have fixed it everywhere if we had a more sophisticated setup, but we are where we are.

PR #2149, opened 2018-10-05
Refresh the MimeDB

rossabaker:


rossabaker:
It's bigger than it looks because of how we have to split the constants across traits.  A bunch of them shifted.

PR #2150, opened 2018-10-05
Release v0.19.0-M4

rossabaker:
If anyone is passionate about a constant for `vnd.futoin+cbor`, then we can block on #2149, but I'd rather ship.


rossabaker:
Looks like a servlet change is imminent.  We'll squeeze that in.

PR #2153, opened 2018-10-06
Release v0.19.0-M4; explain v0.19.0

rossabaker:
Republishes v0.19.0 as v0.19.0-M4, as was originally intended.  Explains our mistake and our recovery plan in the changelog.

version.sbt will need to be manually bumped to 0.19.1-SNAPSHOT after this is merged.

Thanks to @bpholt and @Daenyth for their suggestions on how to deal with this.

### DIFF
--- a/version.sbt
+++ b/version.sbt
@@ -1,1 +1,1 @@
-version in ThisBuild := "0.19.1-SNAPSHOT"
+version in ThisBuild := "0.19.0-M4"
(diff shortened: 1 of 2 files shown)


version.sbt
```diff
@@ -1 +1 @@
-version in ThisBuild := "0.19.1-SNAPSHOT"
+version in ThisBuild := "0.19.0-M4"
```
ChristopherDavenport: I see it this time. :eyes:

Issue #2072, opened 2018-09-12
StaticFile.fromFile has too many overloads


rossabaker:
We abused overloading on `StaticFile.fromFile`.  Look at reducing overloads, either through default parameters or different names for the functions.  Deprecation is good if we're changing public API in a source incompatible way, but if we start now, we can get it cleaned up for 1.0.

Overlaps a bit with #2071, so communicate if you happen to be picking this up at a hack night.

calvinbrown085:
One thing i'm unsure of 
```
def fromFile[F[_]](
      f: File,
      start: Long,
      end: Long,
      buffsize: Int,
      blockingExecutionContext: ExecutionContext,
      req: Option[Request[F]],
      etagCalculator: File => F[String])
```

It seems all of the `fromFile` functions end up calling a variant of this one, having a bunch of default arguments seems kind of gross, but that's what the other functions end up doing anyway. 

What do y'all think?

rossabaker:
Hmm.  What might reasonable defaults be:

* `f` (required)
* `start = 0L`
* `end: Option[Long] = None`
* `buffSize = MagicNumberWeveNeverReallyTuned`
* `blockingExecutionContext` (required; put it second?)
* `req = None`
* `etag = calcEtag` (isn't going to compile because where will it get its `Sync` from?)

I like the approach, but the `etag` is going to be tricky.

calvinbrown085:
@rossabaker I agree, the `etag` is gross the `buffSize` looks like we use the file length?

rossabaker:
Getting the length of the file is an effect, and the file size could be huge, so I think it should be a fixed number.  A multiple of 4K makes sense due to standard block sizes.  And Java's `BufferedInputStream` is 8K, which was probably a thoughtful number.  I would go with `8192`.

If `StaticFile` were an algebra, then `fromFile` could refer to a default `etag` function, and the constraint would be on the algebra instead of the methods.  Maybe we could rename it to something like `Static` and deprecate `StaticFile` for a more gentle migration.  Usage would then look like `Static[IO].fromFile` instead of `StaticFile.fromFile[IO]`.

calvinbrown085:
@rossabaker Maybe i'm getting mixed up, i'm used to algebra's being `trait`'s so 

```
trait Static[F[_]] {
  def fromFile(inputParams*):  OptionT[F, Response[F]] = ???
```

This doesn't seem like what we would want?

rossabaker:
[...]

PR #2158, opened 2018-10-08
Cleanup StaticFile overloads by creating an algebra

calvinbrown085:
addresses https://github.com/http4s/http4s/issues/2072

StaticFile.fromFile had a lot of function overloads that really just were default values. I cleaned this up by making defaults in the function call. 

I also added the other functions present in StaticFile to the Static algebra. I'm guessing we probably want to deprecate the StaticFile function calls, I can look at doing that with appropriate messages. 

Thanks!


calvinbrown085:
I might need some help with the version we are deprecating in. Thanks!

core/src/main/scala/org/http4s/Static.scala
```diff
@@ -0,0 +1,212 @@
...
+trait Static[F[_]] {
+
+  def calcETag(f: File): F[String]
+
+  def fromString(
+    url: String,
+    blockingExecutionContext: ExecutionContext,
+    req: Option[Request[F]] = None): OptionT[F, Response[F]]
+
+  def fromResource(
```
rossabaker: I see `Resource` and think `cats.effect.Resource` now.  How about `fromClasspath` for this?

core/src/main/scala/org/http4s/Static.scala
```diff
@@ -0,0 +1,212 @@
...
+
+  def fromURL(
+     url: URL,
+     blockingExecutionContext: ExecutionContext,
+     req: Option[Request[F]] = None): OptionT[F, Response[F]]
+
+  def fromFile(f: File,
+               blockingExecutionContext: ExecutionContext,
+               start: Long = 0L,
+               end: Option[Long] = None,
```
rossabaker: Don't most IO APIs go `start` and `length` instead of `start` and `end`?  Maybe we got this wrong initially?

core/src/main/scala/org/http4s/Static.scala
```diff
@@ -0,0 +1,212 @@
...
+import org.log4s.getLogger
+
+import scala.concurrent.ExecutionContext
+
+
+trait Static[F[_]] {
+
+  def calcETag(f: File): F[String]
+
+  def fromString(
```
rossabaker: Hmm.  I don't even know what this one does without looking.

core/src/main/scala/org/http4s/Static.scala
```diff
@@ -0,0 +1,212 @@
...
+    url: String,
+    blockingExecutionContext: ExecutionContext,
+    req: Option[Request[F]] = None): OptionT[F, Response[F]]
+
+  def fromResource(
+    name: String,
+    blockingExecutionContext: ExecutionContext,
```
rossabaker: Why isn't the buffer size parameterized here, like for `fromFile`?

core/src/main/scala/org/http4s/Static.scala
```diff
@@ -0,0 +1,212 @@
...
+    req: Option[Request[F]] = None,
+    preferGzipped: Boolean = false): OptionT[F, Response[F]]
+
+
+  def fromURL(
+     url: URL,
+     blockingExecutionContext: ExecutionContext,
```
rossabaker: Again with the buffer size.

core/src/main/scala/org/http4s/Static.scala
```diff
@@ -0,0 +1,212 @@
...
+    req: Option[Request[F]] = None): OptionT[F, Response[F]]
+
+  def fromResource(
+    name: String,
+    blockingExecutionContext: ExecutionContext,
+    req: Option[Request[F]] = None,
+    preferGzipped: Boolean = false): OptionT[F, Response[F]]
+
+
+  def fromURL(
```
rossabaker: `UrlConnection` gives us enough to calculate a reasonable `Etag`.  Should there be a `UrlConnection => F[Etag]` method, and a parameter?

rossabaker: Heck, it'd even work for `fromResource`.

core/src/main/scala/org/http4s/Static.scala
calvinbrown085: I'm fine with this

core/src/main/scala/org/http4s/Static.scala
calvinbrown085: I'm not sure TBH, but length sounds fine to me

core/src/main/scala/org/http4s/Static.scala
calvinbrown085: Nvm, I see where it's used.

calvinbrown085:
@rossabaker Okay, I cleaned this up a little bit, I still have a bit more to go.

core/src/main/scala/org/http4s/Static.scala
calvinbrown085: @rossabaker I'm unsure of what you mean `fromResource` here.

As for your first comment, I have a very basic understanding of `Etag's` Is there any more info you could give on going from a `UrlConnection => Etag` ?

core/src/main/scala/org/http4s/Static.scala
rossabaker: `fromPathname` sounds good to me.

core/src/main/scala/org/http4s/Static.scala
rossabaker: `URLConnection` gives us a lastmodified time and a content length, both of which are often used in hashes of ETag.  Hashing the content makes the best `Etag`, but then you have to load the whole thing, which fights against streaming.  But anything we use to Etag a file should be, for the most part, available here.

core/src/main/scala/org/http4s/Static.scala
```diff
@@ -163,12 +167,12 @@ object Static {
...
                  blockingExecutionContext: ExecutionContext,
                  start: Long = 0L,
-                 end: Option[Long] = None,
+                 length: Option[Long] = None,
                  buffsize: Int = 8192,
                  req: Option[Request[F]] = None,
                  etagCalculator: File => F[String] = calcETag _): OptionT[F, Response[F]] =
       OptionT(for {
-        fileEndSize <- end.fold(F.delay(f.length()))(_.pure[F])
+        fileEndSize <- length.fold(F.delay(f.length()))(_.pure[F])
```
rossabaker: Since we changed this to `length`, I'd expect to see some corresponding logic that adds it to start... or else we were confusing `end` with `start` before.  But this one is not just a simple rename.

calvinbrown085:
I don't have a ton of time to work on this ATM, if someone wants to pick this up please feel free :)

Issue #2157, opened 2018-10-07
Make Http1Client default timeout finite?


kaishh:
Hi, we've recently hit a production bug whereby infinite timeout by default caused http4s server to deadlock.

Our setup was: 
1. http4s server on global execution context
2. scalaz ZIO RTS on a separate limited pool
3. http4s client on global execution context

http4s client was called at various points from ZIO pool via `.unsafeRunSync()` – you can see where this is going. http4s client has been diverging for unknown reason, either way it was hitting the infinite timeout which would eventually lead to blocking of all threads in ZIO pool – setting a timeout would fix that.

Even though async misuse was at fault here, not Http1Client, I do think that setting infinite timeout by default is a "surprising" choice and a potential footgun for unsuspecting users, given that most other http clients usually set a conservatively high, but finite timeout. How about changing it to, e.g. 5 minutes?

rossabaker:
For comparison, [here](https://github.com/AsyncHttpClient/async-http-client/blob/master/client/src/main/resources/org/asynchttpclient/config/ahc-default.properties) are the default timeouts from the async-http-client.

Yes, I think something like this could be reasonable.  We should give a sensible configuration out of the box and continue to give the knobs to tweak it.  Would you like to submit a PR?

kaishh:
@rossabaker 
> Would you like to submit a PR?

Sure https://github.com/http4s/http4s/pull/2159

PR #2159, opened 2018-10-09
#2157 Set default 1 minute timeout for all http clients

kaishh:


async-http-client/src/main/scala/org/http4s/client/asynchttpclient/AsyncHttpClient.scala
```diff
@@ -26,7 +26,7 @@ object AsyncHttpClient {
   val defaultConfig = new DefaultAsyncHttpClientConfig.Builder()
     .setMaxConnectionsPerHost(200)
     .setMaxConnections(400)
-    .setRequestTimeout(30000)
+    .setRequestTimeout(60000)
```
rossabaker: This raises an interesting philosophical point: should there be an http4s default for all client backends, that's as similar as possible, or should we default to the backends' defaults?

kaishh: I think we should set our defaults, since http4s is the principal abstraction here. - users would probably be surprised if changing the backend, while staying within http4s would change defaults.

rossabaker: I think that's a reasonable position.

Maybe we should have a defaults object in client to centralize all that, and then adapt as well as we can in each of the backends.

ChristopherDavenport: I like a second PR to do the latter. As I think there will likely be more than this that we will want to change in that PR. But I think it would be a good idea.

PR #2163, opened 2018-10-11
Add mapK to Request/Response

### LINKED ISSUE #2162: Request/Response MapK
Request and Response could both use a mapK method.

```
// added to Request
def mapK[M[_], N[_]](r: Request[M])(f: M ~> N): Request[N]

// added to Response
def responseMapK[M[_], N[_]](r: Response[M])(f: M ~> N): Response[N]
```

ChristopherDavenport:
Fixes #2162

### DIFF
--- a/core/src/main/scala/org/http4s/Message.scala
+++ b/core/src/main/scala/org/http4s/Message.scala
@@ -187,6 +187,15 @@ sealed abstract case class Request[F[_]](
       attributes = attributes
     )
 
+  def mapK[G[_]](f: F ~> G): Request[G] = Request[G](
+      method = method,
+      uri = uri,
+      httpVersion = httpVersion,
+      headers = headers,
+      body = body.translate(f),
+      attributes = attributes
+    )
+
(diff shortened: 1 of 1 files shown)


core/src/main/scala/org/http4s/Message.scala
```diff
@@ -187,6 +187,15 @@ sealed abstract case class Request[F[_]](
       attributes = attributes
     )
 
+  def mapK[G[_]](f: F ~> G): Request[G] = Request[G](
```
rossabaker: Isn't this just `withBodyStream`?

ChristopherDavenport: No, withBodyStream requires the same `F` although I might question that decision as well.

https://github.com/http4s/http4s/blob/6339d7cb3bee969502a6b50d7989853a9bf57a50/core/src/main/scala/org/http4s/Message.scala#L79

rossabaker: Oh, good point.

PR #2168, opened 2018-10-13
Add Allocate builders

### LINKED ISSUE #2167: Create allocate Method for Clients so people can opt out of lifecycle management
Opt Out For Lifecycle Management

```scala
def allocate: F[(Client[F], F[Unit]]
```
Without this we have no response for when people want to use the client is a non-referentially transparent location.

ChristopherDavenport:
Fixes #2167

### DIFF
--- a/async-http-client/src/main/scala/org/http4s/client/asynchttpclient/AsyncHttpClient.scala
+++ b/async-http-client/src/main/scala/org/http4s/client/asynchttpclient/AsyncHttpClient.scala
@@ -36,15 +36,15 @@ object AsyncHttpClient {
     * Allocates a Client and its shutdown mechanism for freeing resources.
     */
   def allocate[F[_]](config: AsyncHttpClientConfig = defaultConfig)(
-      implicit F: ConcurrentEffect[F]): F[(Client[F], F[Unit])] = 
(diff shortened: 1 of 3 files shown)


async-http-client/src/main/scala/org/http4s/client/asynchttpclient/AsyncHttpClient.scala
```diff
@@ -40,15 +54,7 @@ object AsyncHttpClient {
...
-    Resource
-      .make(F.delay(new DefaultAsyncHttpClient(config)))(c => F.delay(c.close()))
-      .map(client =>
-        Client[F] { req =>
-          Resource(F.async[(Response[F], F[Unit])] { cb =>
-            client.executeRequest(toAsyncRequest(req), asyncHandler(cb))
-            ()
-          })
-      })
+    Resource.make(allocate(config))(_._2).map(_._1)
```
rossabaker: Isn't this just `Resource.apply(allocate(config))`?

ChristopherDavenport: Yup, learned more about resource construction, thanks!

rossabaker:
Could do the same for servers for consistency.

PR #2169, opened 2018-10-13
Implement a response line timeout for blaze server

rossabaker:
If an `HttpApp` takes too long to respond, cancel it and respond with a 503.


rossabaker:
1. The servlet backend already cancels.
2. I think it's highly desirable that it does.  You can opt out, but previously, you couldn't opt in.
3. I mentioned it in the scaladoc, but we can highlight it in the changelog.

PR #2171, opened 2018-10-14
Move requestTimeout into BlazeClient

rossabaker:
Pulls the request timeout implementation out of an imperative blaze stage and into a cats-effect implementation.

The response line timeout can be pulled out next.


rossabaker:
Not in love with this implementation. Going to try again.

Issue #2131, opened 2018-09-28
Factor a common layer for client and server metrics


luisdeltoro:
The purpose of this ticket is:
* Make MetricsOps (currently in client-metrics) generic across server and client.
* Consolidate the server metrics logic for dropwizard and prometheus into one common Metrics object.
* Consolidate the server and client implementations for the different metrics providers into one unique project per metrics implementation (dropwizard-metrics and prometheus-metrics)

luisdeltoro:
@rossabaker I'm gonna start working on this issue and I was wondering how important it is right to keep the metrics "backwards-compatible" after this work.
Can I adjust the names and/or the types of some of the metrics?

luisdeltoro:
Another question. I guess we would like to use Clock here as well instead of using Sync[F].delay(System.nanoTime), right?

rossabaker:
* If you can make the API better, make it better.  Leave a deprecated API if you can.
* If you can make the types and names better, make them better.
* Yes, `Clock` is the right thing to use.

luisdeltoro:
It's not that much about making them better, but more about make them more consistent across the server and client metrics implementations.

What's your opinion about the classifier function? Should we introduce it in the server metrics too? It's probably not so useful as in the client metrics, but we could do it for consistency.
What's your opinion on that?

rossabaker:
I think it's a good idea to have it in both or neither, and people wanted it in client.  So yeah, let's add it.

luisdeltoro:
The point was not that much about make names better, but make them more consistent between client and server, so they can share the same MetricsOps implementation.

I have created a pull request with the changes: https://github.com/http4s/http4s/pull/2173
Like I already mentioned the names and types of some of the metrics needed to be adjusted.

I thought about leaving a deprecated API, but because the actual project has now disappeared, I thought it wouldn't make sense. Please let me know if you think otherwise.

PR #2173, opened 2018-10-14
Common layer server metrics #2131

luisdeltoro:


luisdeltoro:
Thanks @rossabaker Let me if there's anything else I can help with.

PR #2174, opened 2018-10-15
Extract idle timeout into its own stage

rossabaker:
Extracts the idle timeout logic out of `ClientTimeoutStage` into its own stage.  The stage takes a callback, which is lets us move more of the timeout handling out of blaze and into fs2.  This includes:

- Racing against the response
- Interrupting reading the request body
- Interrupting writing the response body

Addition and removal of the phase is now managed by an concurrent effect.


blaze-client/src/test/scala/org/http4s/client/blaze/BlazeClientSpec.scala
```diff
@@ -113,41 +111,28 @@ class BlazeClientSpec extends Http4sSpec {
             .forall(_.contains(true)) must beTrue
         }
 
-        "obey response line timeout" in {
+        "obey response header timeout" in {
```
rossabaker: I can't make much sense of what the previous test did, and it's flaky.

This is somewhat redundant with the ClientTimeoutSpec.

PR #2178, opened 2018-10-16
Redact headers in Retry logging

insdami:
It'd be nice to have a little bit more control on what's logged in `Retry`, specially in the headers.

This is not nice, but it captures the intention. Ideally I'd like to use the `Logger.logMessage` that's in master so I can format the string representation of the request. Would we possible to have in in .18?

### DIFF
--- a/client/src/main/scala/org/http4s/client/middleware/Retry.scala
+++ b/client/src/main/scala/org/http4s/client/middleware/Retry.scala
@@ -10,6 +10,7 @@ import java.time.Instant
 import java.time.temporal.ChronoUnit
 import org.http4s.Status._
 import org.http4s.headers.`Retry-After`
+import org.http4s.util.CaseInsensitiveString
 import org.log4s.getLogger
 import scala.concurrent.ExecutionContext
 import scala.concurrent.duration._
@@ -19,7 +20,7 @@ object Retry {
 
   private[this] val logger = getLogger
 
-  def apply[F[_]](policy: RetryPolicy[F])(client: Client[F])(
+  def apply[F[_]](policy: RetryPolicy[F])(client: Client[F])(redactHeaderWhen: CaseInsensitiveString => Boolean = Headers.SensitiveHeaders.contains)(
       implicit F: Effect[F],
       scheduler: Scheduler,
       executionContext: ExecutionContext): Client[F] = {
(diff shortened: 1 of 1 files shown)


client/src/main/scala/org/http4s/client/middleware/Retry.scala
```diff
@@ -19,7 +20,7 @@ object Retry {
 
   private[this] val logger = getLogger
 
-  def apply[F[_]](policy: RetryPolicy[F])(client: Client[F])(
+  def apply[F[_]](policy: RetryPolicy[F])(client: Client[F])(redactHeaderWhen: CaseInsensitiveString => Boolean = Headers.SensitiveHeaders.contains)(
```
rossabaker: I'd make this the second arg in the first list.  It's nice for middleware composition when the `client` arg stands alone in the last list.

client/src/main/scala/org/http4s/client/middleware/Retry.scala
```diff
@@ -40,14 +41,24 @@ object Retry {
...
               nextAttempt(req, attempts, duration, None)
             case None =>
+              logger.info(e)(
+                s"Request ${showRequest(req, redactHeaderWhen)} threw an exception on attempt #$attempts. Giving up."
+              ) 
               F.raiseError[DisposableResponse[F]](e)
           }
       }
 
+    def showRequest(request: Request[F], redactWhen: CaseInsensitiveString => Boolean): String = {
```
rossabaker: I'd make this private so we can refactor later without breaking bincompat.

insdami:
The need comes from redacting information that might be sensitive to keep in the logs.
I think the logging is fixed here which makes harder for the user to control what's logged and the logging level.
It'd be cool if we can somehow control that, I looked at https://github.com/http4s/http4s/blob/master/client/src/main/scala/org/http4s/client/middleware/Logger.scala#L24 which in someway enables those things.

Thanks for the feedback!

rossabaker:
I'm going to release 0.18.20 tomorrow morning (UTC-4) with the related #2182.  If you want to add a function that takes a `redactHeaders` without breaking binary compatibility, we can try to get this in, too.

insdami:
Cool, I've moved it to default argument and it takes the default sensitive headers as `redactHeadersWhen`. I think, that won't break bin compatibility

client/src/main/scala/org/http4s/client/middleware/Retry.scala
insdami: It's actually a inner function in `apply`

rossabaker: Oops. :+1:

Issue #2049, opened 2018-09-05
Silence expected stack traces in test output


rossabaker:
1. Steal [silenceSystemErr](https://github.com/typelevel/cats-effect/blob/c7af84bc2d3ae63d77336d521818cb05c66579cf/core/shared/src/test/scala/cats/effect/internals/TestUtils.scala#L31), with credit
2. Find the tests that are dumping stack traces
3. Wrap them in `silenceSystemErr`
4. Enjoy cleaner Travis runs

ashwinbhaskar:
@rossabaker can this issue be marked as resolved/close as [this](https://github.com/http4s/http4s/pull/2179) seems to fix it?

rossabaker:
We have the helper now thanks to #2179, but the CI build is still full of them.  We need to apply the helper.  A good way to find them would be to look at a [recent test run](https://github.com/http4s/http4s/runs/655202092?check_suite_focus=true), download the raw log, and search for "at org.http4s".  We should silenceSystemErr in the relevant tests.

ashwinbhaskar:
@rossabaker okay, I can take this up. Can you assign it to me?

ashwinbhaskar:
@rossabaker the trait `ErrorReportingUtils` has been deprecated. I can understand that it is being deprecated from public api usage. But marking it deprecated fails the tests too 
```
 trait ErrorReportingUtils in package testing is deprecated (since 0.21.0-RC2): Not related to HTTP, and will be removed from public interface.
```

rossabaker:
Oops, hadn't seen this yet.  Answered the question on the PR itself.

ashwinbhaskar:
@rossabaker might be a naive question. But wanted to confirm if I have got this right.

[this](https://github.com/http4s/http4s/blob/af4a5b5242685a0add29f44ed60335538b380f2b/blaze-server/src/test/scala/org/http4s/server/blaze/Http1ServerStageSpec.scala#L123) line throws an exception when the IO is evaluated. In this case the evaluation is forced by `Await` [here](https://github.com/http4s/http4s/blob/af4a5b5242685a0add29f44ed60335538b380f2b/blaze-server/src/test/scala/org/http4s/server/blaze/Http1ServerStageSpec.scala#L158). Is my understanding correct?

rossabaker:
[...]

PR #2179, opened 2018-10-16
Silence expected exceptions

entangled90:
Stack trace that print directly to stderr are suppressed, while exceptions logged at error level are still logged.
Should these logs be suppressed too?
Any advice on the "copyright" for the copied function from cats-effect?

Closes issue  #2049

### DIFF
--- a/testing/src/main/scala/org/http4s/testing/ErrorReportingUtils.scala
+++ b/testing/src/main/scala/org/http4s/testing/ErrorReportingUtils.scala
@@ -5,7 +5,6 @@ import scala.util.control.NonFatal
 
 trait ErrorReportingUtils {
 
-
   /**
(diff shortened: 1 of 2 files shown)


rossabaker:
We usually leave an [attribution like this](https://github.com/http4s/http4s/blob/a53dbb1747b69cbe67e637536f73d7d369d62896/core/src/main/scala/org/http4s/parser/AcceptCharsetHeader.scala#L1-L17).  You want to copy the copyright of whatever you copied.  And then make sure the project is mentioned in [NOTICE](https://github.com/http4s/http4s/blob/1b2a84830ab2688e85922b850408306c6ad00007/NOTICE#L10-L13), but it already is.

Thanks for this!

PR #2180, opened 2018-10-17
Change Default Response on Timeout Middleware to ServiceUnavailable

ChristopherDavenport:
Seems consistent with recent changes.


server/src/main/scala/org/http4s/server/middleware/Timeout.scala
```diff
@@ -40,5 +40,5 @@ object Timeout {
       W: EntityEncoder[G, String]): Kleisli[F, A, Response[G]] =
     apply(
       timeout,
-      Response[G](Status.InternalServerError).withEntity("The response timed out.").pure[F])(http)
+      Response[G](Status.ServiceUnavailable).withEntity("The response timed out.").pure[F])(http)
```
rossabaker: We have a `Response.timedOut`.  Any reason not to use that for consistency?

PR #2189, opened 2018-10-19
TickWheelExecutor is a resource, not an allocation

rossabaker:
Composing two `F[A, F[Unit]]`'s together defeats the purpose of the `Resource` AST: any intermediate cancellations can leak a resource.  The only safe way to combine resources is through the resource interpreter.  This makes the `BlazeClientBuilder.resource` the primitive constructor, and expresses `allocate` in terms of it.


core/src/main/scala/org/http4s/internal/package.scala
```diff
@@ -117,4 +119,47 @@ package object internal {
           }
       }
     }
+
+  /* Derived from https://github.com/typelevel/cats-effect/blob/c62c6c76b3066c500fca2f9e0897605bc12eb2a0/core/shared/src/main/scala/cats/effect/Resource.scala */
+  private[http4s] def interpretResource[F[_], A](resource: Resource[F, A])(
```
rossabaker: This is a copy of `Client.toHttpApp`, which should be expressed in terms of this.  But let's agree on whether it should return `ExitCase[Throwable] => F[Unit]` or just `F[Unit]` first.

blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala
```diff
@@ -128,33 +128,28 @@ sealed abstract class BlazeClientBuilder[F[_]] private (
...
-                idleTimeout = idleTimeout,
-                requestTimeout = requestTimeout,
-                scheduler = scheduler,
-                ec = executionContext
-              ),
-              shutdown.guarantee(shutdownS)
-            )
-        }
-    }
+  def allocate(implicit F: ConcurrentEffect[F]): F[(Client[F], ExitCase[Throwable] => F[Unit])] =
```
rossabaker: I think I saw discussion of this return type fly by on Gitter recently, but I can't find it now.  Do we want the `ExitCase` here or nah?

rossabaker:
Build failures are #2191, #2192.

blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala
ChristopherDavenport: Do we care about ExitCase ever?

core/src/main/scala/org/http4s/internal/package.scala
```diff
@@ -117,4 +119,14 @@ package object internal {
           }
       }
     }
+
+  private[http4s] def allocated[F[_], A](resource: Resource[F, A])(
```
ChristopherDavenport: This is too useful, whats the catch? The stronger constraint?

rossabaker: That's the only downside I can see.

PR #2190, opened 2018-10-19
Support setting socket options on blaze-client

rossabaker:


blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala
```diff
@@ -128,10 +133,58 @@ sealed abstract class BlazeClientBuilder[F[_]] private (
   def withoutAsynchronousChannelGroup: BlazeClientBuilder[F] =
     withAsynchronousChannelGroupOption(None)
 
+  def channelOption[A](socketOption: SocketOption[A]) =
```
rossabaker: This logic should probably move to blaze itself.

blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala
```diff
@@ -128,10 +133,58 @@ sealed abstract class BlazeClientBuilder[F[_]] private (
...
+    }
+  def withChannelOptions(channelOptions: ChannelOptions): BlazeClientBuilder[F] =
+    copy(channelOptions = channelOptions)
+  def withChannelOption[A](key: SocketOption[A], value: A): BlazeClientBuilder[F] =
+    withChannelOptions(
+      ChannelOptions(channelOptions.options.filterNot(_.key == key) :+ OptionValue(key, value)))
+  def withDefaultChannelOption[A](key: SocketOption[A]): BlazeClientBuilder[F] =
+    withChannelOptions(ChannelOptions(channelOptions.options.filterNot(_.key == key)))
+
+  def socketSendBufferSize: Option[Int] =
```
rossabaker: These are the options that are allowed on an `AsynchronousSocketChannel` and are provided as a convenience so people don't need to think about existential types.

rossabaker:
Build failure is #2191.

PR #2195, opened 2018-10-19
Merge v0.18.20 to master

rossabaker:


client/src/main/scala/org/http4s/client/middleware/Retry.scala
```diff
@@ -17,14 +18,21 @@ object Retry {
   private[this] val logger = getLogger
 
   def apply[F[_]](policy: RetryPolicy[F])(
+      client: Client[F])(implicit F: Effect[F], T: Timer[F]): Client[F] =
+    retryWithRedactedHeaders(policy, Headers.SensitiveHeaders.contains)(client)
+
+  def retryWithRedactedHeaders[F[_]](
```
rossabaker: This makes sense if and only we're going to try to preserve full bincompat with 0.19.0.  Otherwise, we should deprecate this and add the predicate to `apply`.

ChristopherDavenport: I vote deprecate

rossabaker: I abstain until we merge the rest and run MiMa and see how far away we really are.

rossabaker: `MetricsOps` in core changed incompatibly.  I think that's a dagger right there.

PR #2196, opened 2018-10-19
Require Signal instead of SignallingRef in serveWhile

rintcius:


rintcius:
@rossabaker yeah, I guess they could return an `F[ExitCode]` instead of `Stream[F, ExitCode]` now that fs2 integrates with `IOApp`? Or do you have other doubts? 

I was also thinking that `serveWhile` could be opened up a bit more to allow for an `onStart` hook. For my use-case I couldn't use `serve` and `serveWhile`, but still need to integrate with `IOApp`, so I had to reimplement that machinery

rossabaker:
Build failure is #2197.

rossabaker:
My main objection to `.serve` and `.serveWhile` is that they never give you access to the `Server` itself.  It's not interesting in a lot of cases, but it is the only way to get the bound port when you use port 0.  And it's not very orthogonal to the `Client` design, though there you _always_ need access to the client.

You're looking for a hook after the server is listening?  Just thinking out loud:

```scala
Deferred[F, Unit].flatMap { shutdownSignal =>
  val serverFiber = serverBuilder.resource.use { case server =>
    startupAction >> F.never
  }.start
  moreCodeThatEventuallyCompletesShutdownSignal
  shutdownSignal.get >> serverFiber.cancel.as(ExitCode.Success)
}
```

rintcius:
> My main objection to .serve and .serveWhile is that they never give you access to the Server itself. 

Right, this is what I meant with the onStart hook. I didn't mean to say that it is now impossible to do what I want, just that it involves more code than I think it should.

Concretely, I was thinking about something like this:

```
  final def serve: Stream[F, ExitCode] =
    serveWith(x => x)

  final def serveWith[A](onStart: Stream[F, Server[F]] => Stream[F, A]): Stream[F, ExitCode] =
    for {
      signal <- Stream.eval(SignallingRef[F, Boolean](false))
      exitCode <- Stream.eval(Ref[F].of(ExitCode.Success))
      serve <- serveWhile(onStart, signal, exitCode)
    } yield serve

  final def serveWhile[A](
      onStart: Stream[F, Server[F]] => Stream[F, A],
      terminateWhenTrue: Signal[F, Boolean],
      exitWith: Ref[F, ExitCode]): Stream[F, ExitCode] =
    onStart(stream) *> (terminateWhenTrue.discrete
      .takeWhile(_ === false)
      .drain ++ Stream.eval(exitWith.get))
```

rossabaker:
Wow.  I hadn't even connected the dots between "getting the server so you can inspect its port" and "getting the server so you know when it started."  The latter is generally much more interesting.

I can see these functions all being useful, but another question is whether they're more useful as `Stream`s or `Resource`s.  With the advent of cats-effect-1.0, we now have two common ways of composing servers.

I tend to compose applications in `Stream`, but I am uneasy with `Stream[F, ExitCode]`, because `Stream` is 0..n and an `ExitCode` should be exactly one.  It's not clear from the type whether we should take the first, the last, the first non-zero, etc.  I guess by now we're all getting used to singleton streams and maybe I'm worried about nothing.

The important thing is that we already provide `.resource`, which is the primitive from which _all_ of these other things flow.  The question is then which conveniences we want to support as public API.

(Sorry, that was a long-winded comment without any conclusion.  Just dumping some early morning thoughts.)

rintcius:
> I tend to compose applications in Stream, but I am uneasy with Stream[F, ExitCode]

yeah, me too - in fact that loops back nicely with what I mentioned above "yeah, I guess they could return an F[ExitCode] instead of Stream[F, ExitCode]"

rossabaker:
Is it more common to `.serve` in `Stream` compositions or directly in `IOApp`?  For people composing `Stream`s, is a `Stream.eval` on these reformed methods okay?

I haven't built a whole lot of apps yet in http4s-0.19, so I have more questions than answers.  I'm easily swayed in either direction here.

rintcius:
I was thinking about something like this:

```
    final def serve: F[ExitCode] = serveWith(_.pure[F])

    final def serveWith[A](onStart: Server[F] => F[A]): F[ExitCode] =
      Deferred[F, ExitCode].flatMap { d =>
        serveWhile[A, ExitCode](onStart, d, _._2.pure[F])
      }

    final def serveWhile[A, B](
        onStart: Server[F] => F[A],
        stopOnDefined: Deferred[F, ExitCode],
        onStop: ((A, ExitCode)) => F[B]): F[B] =
      resource
        .flatMap(s => Resource.liftF(onStart(s)))
        .flatMap { a =>
          (Resource.pure[F, A](a), Resource.liftF(stopOnDefined.get)).tupled
        }
        .use(onStop)
```

Does what you mentioned about graceful/hard shutdown fit with `onStop` in `serverWhile`?

rintcius:
yeah, actually I don't know how the code in `serveWhile` et al would change in case of graceful vs hard shutdown. Is there currently any code in http4s that does anything like that?

rossabaker:
No, there isn't any precedent for that.  [fs2-grpc](https://github.com/fiadliel/fs2-grpc/blob/915367b754c0be92249813bcbc329a69e479bf73/java-runtime/src/main/scala/syntax/ServerBuilderSyntax.scala#L18-L49) has a bit of prior art.

rossabaker:
Okay, #2222 is a real-world case.  With Jetty, we can't force a hard stop, but we can set a timeout and let it do a hard stop afterward, and let us know whether it was graceful or not.

rossabaker:
This got derailed by a broader discussion, but I think the original change is still worthwhile.  Merging, and we can continue discussing elsewhere.

Issue #2192, opened 2018-10-19
Flaky test: SSL Handshake WRAP produced 0 bytes.


rossabaker:
```
02:19:49.930 [InnocuousThread-4] WARN  o.h.blaze.pipeline.stages.SSLStage - SSL Handshake WRAP produced 0 bytes.
Status = OK HandshakeStatus = NEED_WRAP
bytesConsumed = 0 bytesProduced = 0
02:19:49.930 [InnocuousThread-4] WARN  o.h.blaze.pipeline.stages.SSLStage - SSL Handshake WRAP produced 0 bytes.
Status = OK HandshakeStatus = NEED_WRAP
bytesConsumed = 0 bytesProduced = 0
02:19:49.931 [InnocuousThread-4] WARN  o.h.blaze.pipeline.stages.SSLStage - SSL Handshake WRAP produced 0 bytes.
Status = OK HandshakeStatus = NEED_WRAP
bytesConsumed = 0 bytesProduced = 0
02:19:49.931 [InnocuousThread-4] WARN  o.h.blaze.pipeline.stages.SSLStage - SSL Handshake WRAP produced 0 bytes.
```

This happened on an [openjdk11 build](https://travis-ci.org/http4s/http4s/jobs/443490411#L3637), which makes me fear that http4s/blaze#233 is sticking our heads in the sand.

rossabaker:
Again: https://travis-ci.org/http4s/http4s/jobs/444236484

rossabaker:
I expect this is fixed by #2204.  Will reopen if it's not.

rossabaker:
[:fire:](https://travis-ci.org/http4s/http4s/jobs/447372206#L3562)

rossabaker:
https://travis-ci.org/http4s/http4s/jobs/447372379#L3243

ghostdogpr:
We just got hit by that in production, right after we switched to JDK11 😢

coltfred:
We also started seeing this after upgrading to jdk 11.

sderosiaux:
Same here. It happens after several minutes, client idle.
AdoptOpenJDK (build 11.0.4+11) / http4s v0.21.0-M5

rossabaker:
We're working on releases of 0.20.x and 0.21.x, but in the meantime, this should be fixed by explicitly upgrading to blaze-0.14.11.

PR #2200, opened 2018-10-20
Make "behave and not deadlock" test parallel

### LINKED ISSUE #2199: Flaky test: behave and not deadlock
https://travis-ci.org/http4s/http4s/jobs/443987859#L4035

rossabaker:
Fix "behave and not deadlock" test:
* Run in parallel, which is a more interesting test.
* Scale the size of the test to the number of processors.  We may just be asking too much of Travis.
* Remove random sleep in test servlet.  I can reproduce the failure with a long enough sleep.

Fixes #2199, I hope.


rossabaker:
The openjdk11 test is still failing due to #2192.

rossabaker:
The oraclejdk8 build has passed a few times in a row with this.

PR #2202, opened 2018-10-21
Allow body for DELETE, CONNECT, GET and HEAD

joan38:


rossabaker:
`DELETE` is just like `GET`, `HEAD`, and `CONNECT` in the spec.  The way we treat `DELETE` should be the way we treat these all:

> A payload within a GET request message has no defined semantics; sending a payload body on a GET request might cause some existing implementations to reject the request.

The one that stands out as different is `TRACE`:

> A client MUST NOT send a message body in a TRACE request.

So I think we should either proceed with this and apply it to `GET`, `HEAD`, and `CONNECT`, or else reject it and say we should use the `Request[F](method = DELETE)` constructor instead.  As discussed on Gitter, the latter steers toward more "correct" HTTP usage, but often as a client, we have to deal with bad HTTP APIs, like k8s.  I'm :+1: as long as it's applied consistently.

joan38:
PR update to do that to CONNECT, GET and HEAD too.
I leave you guys decide if that make sense to apply the change or not.

rossabaker:
Oops.  The `Retry` middleware is using `NoBody` as an indicator for requests that can be retried.  We're going to need to specify a set of retriable methods without using that marker trait.

joan38:
I guess you are speaking about:
```
  def defaultRetriable[F[_]](req: Request[F], result: Either[Throwable, Response[F]]): Boolean =
    req.method.isInstanceOf[Method.NoBody] && isErrorOrRetriableStatus(result)
```
What solution do you suggest?

rossabaker:
Looking at it again, I think idempotent is the right restriction.  I think the definition of `unsafeRetriable` should move to `defaultRetriable`, and we should deprecate `unsafeRetriable`.

joan38:
@rossabaker This CI failure is unrelated to the change right?

rossabaker:
The openjdk11 build failure is unrelated.  The oraclejdk8 build failure is related.  The first build is the only one that compiles the documentation, and we've introduced an ambiguity there.  You should be able to see it locally by running `docs/tut`.

There is something bad about the way we overloaded the apply methods in the client DSL, but I haven't had time to come up with a proposal yet.

rossabaker:
We can't merge this as is because it breaks the tuts.  We still need a solution there.

rossabaker:
Included in #2269.

joan38:
Thanks @rossabaker

rossabaker:
Nah, thanks to you for starting it and @ChristopherDavenport for finishing it. :smile:

Issue #2014, opened 2018-08-19
Look at other middleware for inspiration


rossabaker:
Let's look at other languages and see what we're missing.  Many of them are trivial, and it's a great way to make a first contribution to the project.  If you see one that you'd like, add it here.  And better yet, grab it.  We're here to help.

- [Ruby](https://github.com/rack/rack/wiki/List-of-Middleware)
- [Haskell](http://hackage.haskell.org/packages/search?terms=wai-middleware) and [more Haskell](http://hackage.haskell.org/package/wai-extra)

gstro:
Would there be interest in something like a `ResponseTimer` middleware, similar to express.js' [response-time](https://expressjs.com/en/resources/middleware/response-time.html)? Could use [cat's `Clock`](https://github.com/typelevel/cats-effect/blob/master/core/shared/src/main/scala/cats/effect/Clock.scala).

rossabaker:
Would this be different than the existing Prometheus Metrics and Dropwizard Metrics integrations?  Also see #2131, for making a common metrics implementation with library-specific interpreters.

gstro:
Oh nice! I think my intention would be another "Simple middleware" like `HeaderEcho` for basic timing info without external metrics integrations. If those metrics are sufficient, I can look into other middleware to add.

rossabaker:
You mean like a middleware that returns an `X-Response-Time` header?  I think this might be possible after #2131: instead of writing to a metrics backend, produce a header.

rossabaker:
The existing work has been on the client.  The other ticket is to do similar for the server.  Since the MetricsOps operations return `F[Unit]`, I'm not sure it would work for a response header.

rossabaker:
I'm open to it.  One thing we struggled with on the server metrics: is response time the time until the header, or time until the entire response body is created?  Both can be useful.  You'll want to make clear what's being measured.

nebtrx:
@rossabaker would you be interested in a **Http Method Overrider** middleware (kinda like the [Python's Django](https://pypi.org/project/django-method-override/))?

rossabaker:
@nebtrx Yes, that has proven useful in other languages.

nebtrx:
> @nebtrx Yes, that has proven useful in other languages.

Cool! I just pushed a PR with that purpose. Please take a look at it and let me know how it goes.

rossabaker:
We got a couple out of this.  I think it's served its purpose, but please don't hesitate to bring more useful ones.

PR #2205, opened 2018-10-25
Add simple `ResponseTiming` middleware and test

gstro:
- Uses implicit `Clock[F]` to time from headers parsed to response started
- Adds header to `Response`, named "X-Response-Time" by default

For #2014

### DIFF
--- a/server/src/main/scala/org/http4s/server/middleware/ResponseTiming.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/ResponseTiming.scala
@@ -14,25 +14,26 @@ object ResponseTiming {
   /**
     * Simple middleware for adding a custom header with timing information to a response.
     *
-    * This middleware captures the time from when the request headers are parsed and supplied
-    * to the wrapped service to when the response is started. Metrics middleware, like this one,
+    * This middleware captures the time starting from when the request headers are parsed and supplied
+    * to the wrapped service and ending when the response is started. Metrics middleware, like this one,
     * work best as the outer layer to ensure work done by other middleware is also included.
     *
     * @param http [[HttpApp]] to transform
     * @param timeUnit the units of measure for this timing
     * @param headerName the name to use for the header containing the timing info
     */
-  def apply[F[_]: Sync](
+  def apply[F[_]](
       http: HttpApp[F],
       timeUnit: TimeUnit = MILLISECONDS,
       headerName: CaseInsensitiveString = CaseInsensitiveString("X-Response-Time"))(
-      implicit clock: Clock[F]): HttpApp[F] =
+      implicit F: Sync[F],
+      clock: Clock[F]): HttpApp[F] =
     Kleisli { req =>
       for {
         before <- clock.monotonic(timeUnit)
         resp <- http(req)
         after <- clock.monotonic(timeUnit)
         header = Header(headerName.value, s"${after - before}")
-      } yield resp.copy(headers = resp.headers.put(header))
+      } yield resp.putHeaders(header)
     }
 }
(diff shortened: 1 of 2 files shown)


server/src/main/scala/org/http4s/server/middleware/ResponseTiming.scala
```diff
@@ -0,0 +1,38 @@
...
+      timeUnit: TimeUnit = MILLISECONDS,
+      headerName: CaseInsensitiveString = CaseInsensitiveString("X-Response-Time"))(
+      implicit clock: Clock[F]): HttpApp[F] =
+    Kleisli { req =>
+      for {
+        before <- clock.monotonic(timeUnit)
+        resp <- http(req)
+        after <- clock.monotonic(timeUnit)
+        header = Header(headerName.value, s"${after - before}")
+      } yield resp.copy(headers = resp.headers.put(header))
```
aeons: this could be
```suggestion
      } yield resp.withHeaders(resp.headers.put(header))
```

rossabaker:
Thanks!

* This [failure](https://travis-ci.org/http4s/http4s/jobs/446019835#L3859) is #2199.
* This [failure](https://travis-ci.org/http4s/http4s/jobs/445754003#L3411) is #2192.

I'm busy with blaze and test stability, so I'm relying others to review the code.

server/src/main/scala/org/http4s/server/middleware/ResponseTiming.scala
gstro: I don't think there is a `withHeaders` available. I can do something like this:
```scala
} yield resp.putHeaders(resp.headers.put(header).toList:_*)
```
Do you think that's better?

rossabaker: Seems like an oversight to me.

server/src/main/scala/org/http4s/server/middleware/ResponseTiming.scala
```diff
@@ -0,0 +1,38 @@
...
+    *
+    * This middleware captures the time from when the request headers are parsed and supplied
+    * to the wrapped service to when the response is started. Metrics middleware, like this one,
+    * work best as the outer layer to ensure work done by other middleware is also included.
+    *
+    * @param http [[HttpApp]] to transform
+    * @param timeUnit the units of measure for this timing
+    * @param headerName the name to use for the header containing the timing info
+    */
+  def apply[F[_]: Sync](
```
rossabaker: Consistency observation: Most of our middlewares are `apply(arg1: Blah, arg2: Foo)(http: HttpApp[F])`, so the `HttpApp` is in the final argument list.  Since most of them take implicit constraints, I also don't know how much that convention actually helps us in terms of partial application.

We also usually don't use context bounds when there are other implicits: I would move `Sync` into the list ahead of `clock`.

gstro: If I change the arg list, then application looks a bit weird because of the default arguments: 
```scala
val app = ResponseTiming()(thisService)
```
So I tried to follow [EntityLimiter](https://github.com/http4s/http4s/blob/b3849cf830aa5a1bb51dd8d594e2d8edc45eaee9/server/src/main/scala/org/http4s/server/middleware/EntityLimiter.scala#L18) in this case. But I'm happy to change it to whichever is preferred.

server/src/main/scala/org/http4s/server/middleware/ResponseTiming.scala
gstro: Looks like `putHeaders` was enough.

PR #2207, opened 2018-10-25
Halve the number of implementations of removeDotSegments

aeons:
I noticed that we had implementations of [RFC3986's Remove Dot Segments](https://tools.ietf.org/html/rfc3986#section-5.2.4) in two places (and I think we removed another implementation already?).

I made a very quick JMH benchmark:
```
[info] Benchmark                           Mode  Cnt       Score       Error  Units
[info] ThroughputBenchmark.recursive      thrpt    3   97013.737 ±  5737.676  ops/s
[info] ThroughputBenchmark.stringBuilder  thrpt    3  183276.296 ± 23578.091  ops/s
```
And replaced the recursive implementation, with the string builder based one.

It now lives in `UriFunctions` only, which means it can be found as `Uri.removeDotSegments`.


rossabaker:
I'm assuming that the `PathNormalizerSpec` didn't add any interesting tests above and beyond what's already in [`UriSpec`](https://github.com/http4s/http4s/blob/master/tests/src/test/scala/org/http4s/UriSpec.scala#L751).

aeons:
Ah wait, maybe the property test should stay.

rossabaker:
Needs a scalafmt.  openjdk11 error is yet another instance of #2192.

PR #2210, opened 2018-10-27
Streamline instances

### LINKED ISSUE #2203: Conflicting Implicits orNotFound
This leads to conflicting implicits in combination with `implicits` and the dsl for `.orNotFound`. So we should likely pick one place to be the point of reference for that method and remove it from the other

I'm inclined to favor its removal from the dsl. And have this explicitly come in from implicits.

aeons:
Continuing on from the discussions in #2203, this PR is an exploration that does the following:

- Deprecates `Http4s`, `Http4sInstances`, and `Http4sFunctions`
- Moves instances for cats typeclasses into the corresponding classes' companion objects
- Makes `Http4sDsl` extends `UriFunctions`, `EntityEncoderInstances` and `EntityDecoderInstances` instead of `Http4s`*
- Updates tests and examples

I put the single function in `QValueFunctions` into the `QValue` object. I wanted to do the same for `UriFunctions`, but `Uri.uri` is used in a lot of places and examples unqualified, so I kept it and added it to the DSL and some test traits (but moved the two other functions in the trait).
If the DSL should still bring in the `uri` function, that could be solved by adding it directly in the DSL trait (since it's a macro invocation, it can't just be called), so the `UriFunctions` trait could be removed.

`EntityEncoderInstances` and `EntityDecoderInstances` are mixed into the DSL, but when not using that, getting both of them in scope requires imports. I'm fine with that, and I think I might even prefer imports.

Since some of the traits mixed into the deprecated traits have been removed, it might cause breaking changes, but there will be a deprecation notice at least.

Closes #2203

### DIFF
--- a/blaze-server/src/test/scala/org/http4s/server/blaze/ServerTestRoutes.scala
+++ b/blaze-server/src/test/scala/org/http4s/server/blaze/ServerTestRoutes.scala
@@ -5,6 +5,7 @@ package blaze
 import cats.effect._
 import cats.implicits._
 import fs2.Stream._
+import org.http4s.implicits._
 import org.http4s.Charset._
 import org.http4s.dsl.Http4sDsl
 import org.http4s.headers._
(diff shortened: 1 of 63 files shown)


aeons:
I actually thought I folded the instances into the companion objects here, I'll push that in a bit.

PR #2213, opened 2018-10-29
middleware: CSRF

ahjohannessen:
- change `validate` and `checkCSRFToken` to add new tokens
   on request attributes.

 - Motivation for this is to be able to use CSRF with forms
   as well.

### DIFF
--- a/server/src/main/scala/org/http4s/server/middleware/CSRF.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/CSRF.scala
@@ -2,7 +2,7 @@ package org.http4s
 package server
 package middleware
 
-import cats.{Applicative}
+import cats.Applicative
 import cats.data.{EitherT, Kleisli}
 import cats.effect.Sync
 import cats.syntax.all._
@@ -13,7 +13,7 @@ import javax.crypto.spec.SecretKeySpec
 import javax.crypto.{KeyGenerator, Mac, SecretKey}
 import org.http4s.headers.{Cookie => HCookie}
 import org.http4s.headers.{Host, Origin, Referer, `X-Forwarded-For`}
-import org.http4s.util.{CaseInsensitiveString}
+import org.http4s.util.CaseInsensitiveString
 import org.http4s.internal.{decodeHexString, encodeHexString}
 import org.http4s.Uri.Scheme
 import scala.util.control.NoStackTrace
@@ -149,22 +149,26 @@ final class CSRF[F[_], G[_]] private[middleware] (
   /** To be only used on safe methods: if the method is safe (i.e doesn't modify data)
     * and a token is present, validate and regenerate it for BREACH to be impractical
     */
-  private[middleware] def validate(r: Request[G], response: F[Response[G]])(
+  private[middleware] def validate(r: Request[G], http: Http[F, G])(
       implicit F: Sync[F]): F[Response[G]] =
     CSRF.cookieFromHeaders(r, cookieName) match {
       case Some(c) =>
         (for {
-          raw <- F.fromEither(extractRaw(c.content))
-          res <- response
+          raw      <- F.fromEither(extractRaw(c.content))
           newToken <- signToken[F](raw)
+          res      <- http.run(embedInRequestAttributes(r, newToken))
         } yield res.addCookie(createResponseCookie(newToken)))
           .recover {
             case CSRFCheckFailed => Response[G](Status.Forbidden)
           }
       case None =>
         if (createIfNotFound) {
-          response.flatMap(r => embedNewInResponseCookie(r))
-        } else response
+          generateToken[F].flatMap { t =>
+            val request = embedInRequestAttributes(r, t)
+            val cookie  = createResponseCookie(t)
+            http.run(request).map(_.addCookie(cookie))
+          }
+        } else http.run(r)
     }
 
   /** Check for CSRF validity for an unsafe action.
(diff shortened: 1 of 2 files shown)


ahjohannessen:
@rossabaker @jmcardon WDYT?

ahjohannessen:
I am not sure what I was thinking wrt. changing `checkCSRFToken`, because a form only needs the `validate` change. Will change this tomorrow

ahjohannessen:
@rossabaker Ah, now I remember. It is needed for `checkCSRFToken` in scenarios where a `POST` is invalid and a form is needed, hence a token.

server/src/main/scala/org/http4s/server/middleware/CSRF.scala
```diff
@@ -181,9 +185,9 @@ final class CSRF[F[_], G[_]] private[middleware] (
         c1 <- CSRF.cookieFromHeadersF[F, G](r, cookieName)
         raw1 <- F.fromEither(extractRaw(c1.content))
         raw2 <- F.fromEither(extractRaw(rawToken))
-        response <- if (CSRF.isEqual(raw1, raw2)) respAction
-        else F.raiseError[Response[G]](CSRFCheckFailed)
         newToken <- signToken[F](raw1) //Generate a new token to guard against BREACH.
+        response <- if (CSRF.isEqual(raw1, raw2)) http.run(embedInRequestAttributes(r, newToken))
+                    else F.raiseError[Response[G]](CSRFCheckFailed)
       } yield response.addCookie(ResponseCookie(name = cookieName, content = unlift(newToken))))
```
ahjohannessen: Is there a reason why `createResponseCookie` is not used here, but used in `private[middleware] def validate` The difference is that `createResponseCookie` sets `httpOnly` to `true`, while default apply on `ResponseCookie` sets `httpOnly` to `false`. 

`TSecCSRF` where this middleware presumable originates from uses `ResponseCookie(name = cookieName, content = unlift(newToken)))` everywhere, i.e. no restriction on `httpOnly`.

@jmcardon Could you please elaborate a bit?

ahjohannessen:
```
  def createResponseCookie(token: CSRFToken): ResponseCookie =
    ResponseCookie(name = cookieName, content = token, httpOnly = true, secure = self.secure)
```

Is the client in case of `Ajax/XHR` requests not supposed to be able to get hold of tokens, or am I misunderstanding the usage of this middleware?

ahjohannessen:
The middleware type for CSRF in http4s is
```scala
Middleware[F, Request[G], Response[G], Request[G], Response[G]]
```
whereas in `TSecCSRF` it is:
```scala
Middleware[OptionT[F, ?], Request[F], Response[F], Request[F], Response[F]]
```

Why is this so?

jmcardon:
@ahjohannessen because tsec is before `0.19` changes.

server/src/main/scala/org/http4s/server/middleware/CSRF.scala
jmcardon: re first point: I think that should also use `createResponseCookie` and you just caught a bug.

> Is the client in case of Ajax/XHR requests not supposed to be able to get hold of tokens, or am I misunderstanding something obvious?

They're not supposed to be able to get hold of tokens _in the cookie_. You have to explicitly embed the token to grab it in a route where you need to do this, but you're not supposed to arbitrarily be able to read it from the cookie as well.
The owasp page on double submit cookie pattern is pretty good about describing this.

jmcardon:
What is the goal with adding CSRF tokens to request attributes? I don't think I understand.

ahjohannessen:
@jmcardon such that they are available for hidden form fields in twirl

jmcardon:
If you're using twirl to render a form server side, you can still extract the csrf token as usual. This is what this function is for:
```
   def refreshOrCreate[M[_]](r: Request[G])(
       implicit F: Sync[M]): EitherT[M, CSRFCheckFailed, CSRFToken] =	       implicit F: Sync[M]): EitherT[M, CSRFCheckFailed, CSRFToken] =
     CSRF.cookieFromHeaders(r, cookieName) match {	     CSRF.cookieFromHeaders(r, cookieName) match {
       case Some(c) =>	       case Some(c) =>
         EitherT(F.pure(extractRaw(c.content)))	         EitherT(F.pure(extractRaw(c.content)))
           .semiflatMap(signToken[M])	           .semiflatMap(signToken[M])
       case None =>	       case None =>
         EitherT.liftF(generateToken[M])	         EitherT.liftF(generateToken[M])
     }	     }
```

server/src/main/scala/org/http4s/server/middleware/CSRF.scala
ahjohannessen: >You have to explicitly embed the token to grab it in a route where you need to do this, but you're not supposed to arbitrarily be able to read it from the cookie as well.

Can you perhaps show an example how this middleware is supposed to be used. I cannot see any set x-crsf-token header in the code

ahjohannessen:
@jmcardon Reading a bit up on this I think I have misunderstood how this works. I thought that the raw token was changed on every request, it is just validated and re-signed.

So, in case of twirl forms I can do this:

 1. On login POST route after authentication I use `embedNewInResponseCookie`.
 2. Use `refreshOrCreate` for subsequent twirl forms.

I should not need to wrap login (GET/POST) routes with csrf.validate(), because the user has to authenticate anyway. 

So an example of how I imagine how to use this is:

```scala
val csrf: CSRF[OptionT[F, ?], F] = ???
...
val asset: HttpRoutes[F] = ???
val login: HttpRoutes[F] = ???

val serv1: HttpRoutes[F] = ???
val serv2: HttpRoutes[F] = ???
val serv3: HttpRoutes[F] = ???
...
val services: HttpRoutes[F] = auth.liftService(serv1 <+> serv2 <+> serv3)
val protect: HttpRoutes[F]  = csrf.validate()(services)

val app = Router("/" → protect, "/auth" → login, "/assets" → asset).orNotFound
```

Should the user of csrf not be able to specify response cookie path, to say `/`?

ahjohannessen:
Closing this as it was based on wrong assumptions from my side.

Issue #2045, opened 2018-09-03
Scalafix from 0.18 to 0.19


rossabaker:
* 0.18 to 0.19 has a lot of mechanical changes that would be awfully nice to implement via scalafix.
* 0.19 to 1.0 will likely have more, and it'd be nice to have a precedent.

Whoever does this will be Great Friend of Project. :medal_military:

Zelenya:
Idea.
1. List all the changes which need a scalafix (in this issue or separate issues)
2. Tweet a call for participation
3. Wake up to everything fixed

jmcardon:
> Wake up to everything fixed

_Someday, I wish upon a star_
_Wake up where the clouds are far behind me_

rossabaker:
I think that's a great idea.  A couple that immediately to mind:
* `withBody` => `withEntity` (tricky, because return type changes)
* `HttpService => HttpRoutes.of`
* The client builders (after #1993)

I'm sure I'll think of more as I update release notes.

What else are upgraders suffering from?

hamnis:
add typealiases `RequestCookie` and `ResponseCookie` in `org.http4s` to `org.http4s.Cookie` in 0.18.x may help a bit.

m-sp:
I think this got pretty stale, we have Scalafix 0.18 to 0.20

PR #2214, opened 2018-10-29
Scalafix 0.18 to 0.20

amarrella:
Attempts a first stab at #2045

It's not much yet but feel free to give feedback / suggestions. 
It's my first time ever writing scalafix rules.

Also, feel free to discard it completely if you have something better already


SystemFw:
This is great!

amarrella:
Thanks for the nice words :) I have a few questions:
- I think we should have only 1 rule to fix from 0.18 to 0.20 since it will be a one-off job. What do people think? We can still keep the code clean by refactoring but I think running a single command would be a nicer experience than running them in a sequence.
- For the `withBody` -> `withEntity` fix do we want to change the type or should we wrap the `foo.withEntity` in some kind of `F.pure`?
- For the client i'm working my way to it here https://github.com/amarrella/http4s/pull/1 There is a bit of complexity caused by the fact that `.apply` disappears in favor of `resource`.  I think it will be really hard to automatically refactor the user's code to use resource. Are we fine in transforming it into a `.resource` and let the user follow the compiler errors to refactor?
- For the resource/stream now we need at least a ContextShift available (if not a ConcurrentEffect, if we are not using IO). For now I think i'll just make the user do that change as well following the compile errors. We can improve later i think, at least for people that use IO. Generic code becomes a little trickier. 

I know that the code is a bit messy, but bear with me, I promise it will be much cleaner soon :)

scalafix/build.sbt
```diff
@@ -0,0 +1,60 @@
+lazy val V = _root_.scalafix.sbt.BuildInfo
+inThisBuild(
+  List(
+    organization := "com.example",
+    homepage := Some(url("https://github.com/http4s/http4s")),
+    licenses := List("Apache-2.0" -> url("http://www.apache.org/licenses/LICENSE-2.0")),
+    developers := List(
+      Developer(
```
amarrella: About this: I actually don't care about being marked as the developer or anything. So please change it to work with your release process :)

amarrella:
I published a snapshot (under my name for now) and added a readme, try it out :) 
https://github.com/http4s/http4s/pull/2214/commits/186d6cd00ff6d083c3b16ce12d3dd683464882a6

rossabaker:
We're thinking about another milestone tomorrow.  It would be neat to get this out.  I will give it a more thorough review when the workday is done.

I am ignorant on scalafix: if we try to implement rules for an 0.18.20 -> 0.20.0-M2 project, can it be run to go from 0.20.0-M1 to 0.20.0-M2?  Or does it only work between known endpoints?

amarrella:
It's totally independent from the version. scalafix just looks at your code, sees what's in there and changes it based on the rule. I called the rule Http4s018To020 because it contains the changes to go from 0.18 to 0.20, but you can run it on 0.17 or on akka-http code if you want :D 

Obviously if the api changes between 0.20-M1 and 0.20-M2 we will need to touch the rules to reflect the changes, but that's about it

At the moment the rules are a bit raw and can definitely be improved, but I think a few decisions on the points above can be made and can be handled by a separate PR :)

rossabaker:
> I think we should have only 1 rule to fix from 0.18 to 0.20 since it will be a one-off job. What do people think? We can still keep the code clean by refactoring but I think running a single command would be a nicer experience than running them in a sequence.

I agree with this, especially if it's still helpful for people migrating from milestones in between.  If not, well, we can readjust our strategy.

> For the withBody -> withEntity fix do we want to change the type or should we wrap the foo.withEntity in some kind of F.pure?

I bet `F.pure` is easier for a migration that still compiles, but it no longer _needs_ to be in `F`.  I'd like to encourage people to not keep that in `F`, but I'd also like to make it easy for people to migrate.  I'll delegate to you on which is smarter here, but that's the tradeoff.

> For the client i'm working my way to it here amarrella#1 There is a bit of complexity caused by the fact that .apply disappears in favor of resource. I think it will be really hard to automatically refactor the user's code to use resource. Are we fine in transforming it into a .resource and let the user follow the compiler errors to refactor?

This is similar to the previous question in terms of tradeoffs.  I would like to encourage `.resource`, but `.allocate` would give something that compiles.  (Replacing shutdown is trickier, though.)

> For the resource/stream now we need at least a ContextShift available (if not a ConcurrentEffect, if we are not using IO). For now I think i'll just make the user do that change as well following the compile errors. We can improve later i think, at least for people that use IO. Generic code becomes a little trickier.

Migrating `StreamApp` to `IOApp` is out of scope since those aren't our types, right?  Because if we open that door, then we're talking about cats-effect and fs2 migrations.

Sorry if these are redundant with the code.  Just trying to provide some overdue answers until I have time to look deeper.

rossabaker:
> Obviously if the api changes between 0.20-M1 and 0.20-M2 we will need to touch the rules to reflect the changes, but that's about it

There will be some modest changes between 0.20.0-M1 and 0.20.0-M2, and probably again before 0.20.0.  But I expect those to be _much_ smaller than what's already covered.

scalafix/build.sbt
rossabaker: This doesn't touch the main build.sbt at all.  Would we deploy this as a submodule, or is it just a project-in-a-project?

amarrella:
> I bet F.pure is easier for a migration that still compiles, but it no longer needs to be in F. I'd like to encourage people to not keep that in F, but I'd also like to make it easy for people to migrate. I'll delegate to you on which is smarter here, but that's the tradeoff.

I'm also for encouraging people. Following the compile errors should be enough. Worst case they can add the F.pure themselves
> This is similar to the previous question in terms of tradeoffs. I would like to encourage .resource, but .allocate would give something that compiles. (Replacing shutdown is trickier, though.)

I didn't even know about `.allocate`! Again though, I'm + for encouraging people to do the right thing,  though i expect it to be one of the most painful things to change in a codebase on migration (if they weren't using `.stream`. I think if people already used `.stream` they will be fine either way. I don't know the percentages though. Your call
> Migrating StreamApp to IOApp is out of scope since those aren't our types, right? Because if we open that door, then we're talking about cats-effect and fs2 migrations.

Yes you are right (I might actually create a separate scalafix for fs2 StreamApp to IOApp :) )

> There will be some modest changes between 0.20.0-M1 and 0.20.0-M2, and probably again before 0.20.0. But I expect those to be much smaller than what's already covered.

That's fine, we can just update the rule while we do it. This opens a question wrt versioning as well. Since the http4s versions (major & minor) themselves are in the rule name, we could just have an independent semver for the rule (since rules will evolve at a different pace than the rest of the project, i assume )

amarrella:
Also, I still have to figure out how to format/indent the generated code properly. I hope people will just use scalafmt for that or someone else takes care of it in another pr (and teaches me how to do it ^_^)

scalafix/build.sbt
amarrella: What you prefer, I didn't think about deployment at all and I don't know if it's best as a submodule or a project in a project. Your call :)

scalafix/rules/src/main/scala/fix/ClientRules.scala
```diff
@@ -0,0 +1,114 @@
...
+        Patch.addGlobalImport(importer"scala.concurrent.ExecutionContext.Implicits.global")
+      case _ =>
+        Patch.empty
+    })
+  }
+
+
+  private[this] def applyClientType(defn: Defn, client: Tree, clientType: ClientType): Patch = clientType match {
+    case ClientType.Stream => Patch.addRight(client, ".stream")
+    case ClientType.Apply =>
```
amarrella: There is a bug here that causes a matcherror in some cases, will fix in a few hours when i get home

rossabaker:
I think pointing people to scalafmt is an entirely appropriate solution to the formatting issue.

rossabaker:
> I didn't even know about .allocate! Again though, I'm + for encouraging people to do the right thing, though i expect it to be one of the most painful things to change in a codebase on migration (if they weren't using .stream. I think if people already used .stream they will be fine either way. I don't know the percentages though. Your call

The whole reason `.allocate` exists is because this refactoring is hard for people.  Can we refactor it to reference `.allocate` and leave a comment in their code about how `.resource` or `.stream` is better?  Maybe inserting "code reviews" into a scalafix is inappropriate, but it's a compromise that comes to mind.

> That's fine, we can just update the rule while we do it. This opens a question wrt versioning as well. Since the http4s versions (major & minor) themselves are in the rule name, we could just have an independent semver for the rule (since rules will evolve at a different pace than the rest of the project, i assume )

I was leaning toward it being nice to publish this along with the project.  We can be pretty aggressive about patch releases as long as they're not breaking binary compatibility with production code.  Maybe we publish this rule as part of the project, so it's easy to find, and has fixes that target the corresponding version.  But not enforce MiMa in the rules themselves.  The point of this module would be entirely to get people from previous production versions to the version that provides the fix.

amarrella:
> The whole reason .allocate exists is because this refactoring is hard for people. Can we refactor it to reference .allocate and leave a comment in their code about how .resource or .stream is better? Maybe inserting "code reviews" into a scalafix is inappropriate, but it's a compromise that comes to mind.
Ok will do that then :) We can both leave an actual comment in the code or do a `Patch.info`or `Patch.warn` which will show the message during the scalafix run (less invasive, but might be forgotten later?). Also it would be maybe nice to add this to a separate "best practices" rule set that people can run on demand or in their CI. What do you think? 

> I was leaning toward it being nice to publish this along with the project. We can be pretty aggressive about patch releases as long as they're not breaking binary compatibility with production code. Maybe we publish this rule as part of the project, so it's easy to find, and has fixes that target the 
corresponding version. But not enforce MiMa in the rules themselves. The point of this module would be entirely to get people from previous production versions to the version that provides the fix.

OK works for me :) I'll integrate it into the main `build.sbt` in a few hours when I get home.

rossabaker:
> Also it would be maybe nice to add this to a separate "best practices" rule set that people can run on demand or in their CI. What do you think?

Are people going to be running this in CI?  I expected a one-shot cleanup for most people.

amarrella:
> Are people going to be running this in CI? I expected a one-shot cleanup for most people.

The migration rule will be one shot, I was more considering adding linting/style rules to another set where people can pick and choose them and run them on their CI or something. But that could be discussed / added separately. Not in the scope of this PR / Rule

scalafix/rules/src/main/scala/fix/ClientRules.scala
amarrella: Bug was in another line actually, fixed in `6503abb4a`

amarrella:
> The whole reason .allocate exists is because this refactoring is hard for people. Can we refactor it to reference .allocate and leave a comment in their code about how .resource or .stream is better? Maybe inserting "code reviews" into a scalafix is inappropriate, but it's a compromise that comes to mind.
@rossabaker working on this. But it seems that `.allocate` returns `F[(Client[F], F[Unit])]`. Does it change in M2? Or how should I handle the case?

amarrella:
I'm stuck on the migration to the main `build.sbt`:/ `scalafix-inputs` and `scalafix-outputs` contain compile errors, but they are fine to exist in the scalafix world (the fix works even if your code doesn't compile before/after) :/

amarrella:
> I'm stuck on the migration to the main build.sbt:/ scalafix-inputs and scalafix-outputs contain compile errors, but they are fine to exist in the scalafix world (the fix works even if your code doesn't compile before/after) :/

Fixed by removing problematic code and ignoring fatal warnings in the packages

scalafix-outputs/src/main/scala/fix/Http4s018To020.scala
```diff
@@ -53,8 +53,8 @@ object Http4s018To020 {
     group = None
   )
 
-  val client: cats.effect.Resource[IO, Client[IO]] = BlazeClientBuilder[IO](ExecutionContext.fromExecutorService(Executors.newFixedThreadPool(1))).resource
+  val client: IO[Client[IO]] = BlazeClientBuilder[IO](ExecutionContext.fromExecutorService(Executors.newFixedThreadPool(1))).allocate.map{ case (client, shutdown) => client }
```
amarrella: would this be ok? I'm not sure what to do with the second value if we want to keep the same type

rossabaker: We would not be able to shut it down with that. I think we need to return the IO[Client[IO], F[Unit]]

amarrella:
For now the instructions to run this are:
1. Make sure your scala version is >= `2.11.12` or `2.12.7` ([no 2.13 yet](https://github.com/scalameta/scalameta/issues/1695))
2. Add the scalafix plugin to your `project/plugins.sbt`
`addSbtPlugin("ch.epfl.scala" % "sbt-scalafix" % "0.9.0")`
3. Add this to your `build.sbt`
```
scalafixDependencies in ThisBuild += "org.http4s" %% "http4s-scalafix" % "0.20.0-SNAPSHOT"
addCompilerPlugin(scalafixSemanticdb)
scalacOptions += "-Yrangepos"
```
4. Run
`sbt scalafix Http4s018To020`

I was hoping to find a simpler (one command to run) way that didn't make users change their build and plugins but for now that's the easiest i found.

Do you want to add this somewhere in the docs? Perhaps in the first page [https://http4s.org/v0.20/](https://http4s.org/v0.20/) under a *Migration from http4s 0.18* title so that users don't have to look for it?

I'm sorry for the many questions! Anyways besides that, I think the pr is ready now if you want to experimentally add it to M2 :)

rossabaker:
Yes, I think a page on the docs for this would be great.  Just an md in docs/src/main/tut -- even if there isn't any code to compile, that's the right way to add a new page.  Would you want to give that a try or should we leave it for a subsequent PR?

I still need to go through the code itself, but this is work is very much needed and appreciated.  Thank you for a great PR.

build.sbt
```diff
@@ -329,6 +378,10 @@ lazy val docs = http4sProject("docs")
         examplesTomcat,
         examplesWar,
         mimedbGenerator,
+        scalafixRules,
+        scalafixInputs,
+        scalafixOutputs,
+        scalafixTests
```
rossabaker: Ah, good, I was expecting this would get missed.

project/Http4sPlugin.scala
```diff
@@ -334,4 +334,6 @@ object Http4sPlugin extends AutoPlugin {
   lazy val tomcatCatalina                   = "org.apache.tomcat"      %  "tomcat-catalina"           % "9.0.12"
   lazy val tomcatCoyote                     = "org.apache.tomcat"      %  "tomcat-coyote"             % tomcatCatalina.revision
   lazy val twirlApi                         = "com.typesafe.play"      %% "twirl-api"                 % "1.3.15"
+  lazy val scalafixCore                     = "ch.epfl.scala"          %% "scalafix-core"             % "0.9.0"
+  lazy val scalafixTestKit                  = "ch.epfl.scala"          % "scalafix-testkit"           % scalafixCore.revision % Test cross CrossVersion.full
```
rossabaker: Nit: we usually keep these alphabetical.  It helps when maintaining across branches.

project/plugins.sbt
```diff
@@ -1,4 +1,7 @@
-resolvers += "jgit-repo" at "http://download.eclipse.org/jgit/maven"
+resolvers ++= Seq(
+  "jgit-repo" at "http://download.eclipse.org/jgit/maven",
+  Resolver.sonatypeRepo("releases")
```
rossabaker: Is this one not included by default?  What are we getting from this one?

project/plugins.sbt
```diff
@@ -14,7 +17,7 @@ addSbtPlugin("io.verizon.build"    %  "sbt-rig"                   % "5.0.39")
 addSbtPlugin("org.tpolecat"        %  "tut-plugin"                % "0.6.7")
 addSbtPlugin("pl.project13.scala"  %  "sbt-jmh"                   % "0.3.4")
 addSbtPlugin("com.timushev.sbt"    %  "sbt-updates"               % "0.3.4")
-
+addSbtPlugin("ch.epfl.scala"       % "sbt-scalafix"               % "0.9.0")
```
rossabaker: Another one to alphabetize, although it looks like we already messed up.

scalafix-tests/src/test/scala/fix/RuleSuite.scala
```diff
@@ -0,0 +1,7 @@
+package fix
+
+import scalafix.testkit.SemanticRuleSuite
+
+class RuleSuite extends SemanticRuleSuite() {
+  runAllTests()
+}
```
rossabaker: Are the inputs and outputs used for anything but tests?  I'm surprised they don't live in directories here.  I still need to RTFM on scalafix.

scalafix/src/main/scala/fix/MimeRules.scala
```diff
@@ -0,0 +1,20 @@
+package fix
+
+import scalafix.v1._
+import scala.meta._
+
+object MimeRules {
```
rossabaker: This one is great, because that was a really tedious change.

scalafix-tests/src/test/scala/fix/RuleSuite.scala
amarrella: Correct they are only used for tests. I think they need to be in separate packages due to how semanticdb works, but I can try (maybe after this PR) to see if they work as one package.

amarrella:
Thanks :) I really that it will help people in the migration! This library is fantastic and honestly can't wait enough for 0.20 :)

I'll address a few of your comments now and I'll push a small doc as well. That can be improved in a separate PR but I would love for people to use it in M2 so I can see what can be improved (maybe with the help of other contributors as well) before the final release, so I feel it's important to provide some docs at least.

Just fyi, I opened a [fs2](https://github.com/functional-streams-for-scala/fs2/pull/1304) pr (just for StreamApp for now) which should help in the migration as well. If this doesn't get merged to fs2, i'll publish it separately but should be available soon as well.

scalafix/src/main/scala/fix/MimeRules.scala
amarrella: I hope I covered most of the cases, waiting for real world feedback on this one :)

amarrella:
I added the doc, though I was unable to preview it. I set weight 2 so it should be just after quickstart. If that's not ok takes a second to change :) 

I did `sbt "; project docs; tut; makeSite; previewSite"` but didn't work (even though the docs compiled fine - I think I have the exact same issue as https://github.com/http4s/http4s/pull/1718 ).

scalafix/src/main/scala/fix/MimeRules.scala
amarrella: I improved these in https://github.com/http4s/http4s/pull/2214/commits/4edd9ccadcb1735d1e1f67f6e1aa483db6eb4aa3 . Now should be able to handle more use cases (and imports as well)

scalafix/src/main/scala/fix/HttpServiceRules.scala
```diff
@@ -0,0 +1,13 @@
+package fix
+
+import scalafix.v1._
+import scala.meta._
+
+object HttpServiceRules {
+  def unapply(t: Tree): Option[Patch] = t match {
+    case t @ Type.Name("HttpService") => Some(Patch.replaceTree(t, "HttpRoutes"))
```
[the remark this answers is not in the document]
amarrella: Right, awesome! I'll commit those changes now

scalafix-tests/src/test/scala/fix/RuleSuite.scala
rossabaker: Ah, that makes a lot of sense.

rossabaker:
I still kind of like the single build, so we can publish these using the machinery that we already have in place.  But as noted earlier, tweaking the scalafix rules will generally require a patch release of the main project.  It looks like the cats one is run through `publishLocal`, so maybe that could work for people between releases?

aeons:
@amarrella This is awesome, thanks a lot!

rossabaker:
Any reason we shouldn't merge this now?

amarrella:
> Any reason we shouldn't merge this now?

Not from me :)

PR #2228, opened 2018-10-31
csrf: cookie settings + builder + `createResponseCookie` usage

ahjohannessen:
- fix `checkCSRFToken` to use `createResponseCookie`.
 - make it possible to configure various relevant settings
   for csrf cookie.
 - add builder for `CSRF`
 - add ability to read token from form

### DIFF
--- a/server/src/main/scala/org/http4s/server/middleware/CSRF.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/CSRF.scala
@@ -61,7 +61,7 @@ final class CSRF[F[_], G[_]] private[middleware] (
     key: SecretKey,
     headerCheck: Request[G] => Boolean,
     csrfCheck: CSRF[F, G] => CSRF.CSRFCheck[F, G]
-  )(implicit F: Sync[F]) { self =>
+)(implicit F: Sync[F]) { self =>
   import CSRF._
 
   private val csrfChecker: CSRFCheck[F, G] = csrfCheck(self)
@@ -90,14 +90,14 @@ final class CSRF[F[_], G[_]] private[middleware] (
     */
   def createResponseCookie(token: CSRFToken): ResponseCookie =
     ResponseCookie(
-      name      = cookieSettings.cookieName,
-      content   = unlift(token),
-      expires   = None,
-      maxAge    = None,
-      domain    = cookieSettings.domain,
-      path      = cookieSettings.path,
-      secure    = cookieSettings.secure,
-      httpOnly  = cookieSettings.httpOnly,
+      name = cookieSettings.cookieName,
+      content = unlift(token),
+      expires = None,
+      maxAge = None,
+      domain = cookieSettings.domain,
+      path = cookieSettings.path,
+      secure = cookieSettings.secure,
+      httpOnly = cookieSettings.httpOnly,
       extension = cookieSettings.extension
     )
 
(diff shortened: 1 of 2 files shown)


ahjohannessen:
@jmcardon Fix for the bug you mentioned in https://github.com/http4s/http4s/pull/2213

rossabaker:
I don't know CSRF very well, but I stumbled across [this discussion](https://security.stackexchange.com/questions/175536/does-a-csrf-cookie-need-to-be-httponly).  Does that discussion ring true to people who know what they're taking about?

jmcardon:
@rossabaker that's correct.

I think better yet would be making this field optional. The csrf token being accessible from js is sort of a necessity anyway @ some point.

ahjohannessen:
@jmcardon @rossabaker Updated the PR by adding `CSRFCookieSettings`. WDYT?

jarreds:
I like this setup. Was looking to add `SameSite=strict` to the cookie and this would allow that.

server/src/main/scala/org/http4s/server/middleware/CSRF.scala
```diff
@@ -244,6 +247,24 @@ final class CSRF[F[_], G[_]] private[middleware] (
 
 object CSRF {
 
+  final case class CSRFCookieSettings(
```
rossabaker: Are we fairly certain that nothing will be added to this in the future?  I'm trying to move away from case classes for config for binary compatibility reasons.  My altnernative has been the server and client builders, which are painful to write, but can support additions in the future without breaking code.

jarreds: +1 on builders. Started on a (much less thorough) builder-like addition here https://github.com/http4s/http4s/pull/2235 until i discovered this PR.

ahjohannessen: I don’t think much more is needed wrt cookie settings. Are you thinking about a builder for the middleware itself @rossabaker ?

ahjohannessen:
According to [this site](https://scotthelme.co.uk/csrf-is-dead), apparently:
> `Same-Site` cookies completely and effectively neutralise CSRF attacks.

@rossabaker wrt nothing being added in the future, these two are missing:
```scala
expires: Option[HttpDate] = None,
maxAge: Option[Long] = None,
```

I do not see how those are relevant.

ahjohannessen:
@rossabaker Would it be useful to add this as a middleware:
```scala
/** [[Middleware]] for lifting application/x-www-form-urlencoded values
  * into the request header values. This middleware is particular useful
  * for scenarios where you use some middleware, e.g. `CSRF`, and need
  * to check for a header value that is only available as a form value.
  * */
object UrlFormToHeader {

  type FieldName  = String
  type HeaderName = String

  /*
   * Form fields listed in `fieldNames` are attempted to be lifted into headers.
   * It does not try to replace headers that already exist with same name.
   * */
  def apply[F[_]: Sync, G[_]: Sync](nt: G ~> F)(
    fieldNames: NonEmptyList[String],
    fieldToHeader: FieldName ⇒ HeaderName = identity,
    checkRequest: Request[F] ⇒ Boolean = !_.method.isSafe,
    strictDecode: Boolean = false
  ): Http[F, G] ⇒ Http[F, G] =
    http ⇒
      Kleisli { req ⇒
        def fieldsToHeaders(form: UrlForm): F[Response[G]] = {

          val relevantFormValues = form.values.filterKeys(fieldNames.contains_).toList
          val lifted = relevantFormValues.flatMap {
            case (k, vs) ⇒ vs.headOption.map(v ⇒ Header(fieldToHeader(k), v))
          }

          val original   = req.headers.toList
          val filtered   = lifted.filter(original.contains_(_))
          val newRequest = req.putHeaders(filtered: _*)

          http(newRequest)
        }

        req.headers.get(headers.`Content-Type`) match {
          case Some(headers.`Content-Type`(MediaType.application.`x-www-form-urlencoded`, _)) if checkRequest(req) ⇒
            for {
              decoded ← nt(UrlForm.entityDecoder[G].decode(req, strictDecode).value)
              resp    ← decoded.fold(mf ⇒ nt(mf.toHttpResponse[G](req.httpVersion)), fieldsToHeaders)
            } yield resp

          case _ ⇒ http(req)
        }
    }
}
```

This means that one could to this:

```scala

def csrf(routes: HttpRoutes[F]) = {
 val ufthM = UrlFormToHeader(NonEmptyList.one("X-Csrf-Token"))
 val csrfM = ???
 utfhM(csrfM.validate()(routes))
} 

val wrapped = csrf(routes1 <+> routes2)
```

server/src/main/scala/org/http4s/server/middleware/CSRF.scala
rossabaker: I was thinking a builder for the config, but I suppose it could be a builder for the middleware.  We merged configuration and builder in the server and client builders.  I haven't fully thought through the ergonomics of it for middleware.  :thinking:

* `CSRF(http)`: defaults
* `CSRF.withCookieName("x-csrf-token").withSecure(true)(http)`

So `CSRF` is a class that is `Http` => `Http`.  Object (or value) `CSRF` is  `CSRF` with default values.  And you can configure things until applying the `http`.

This pattern could apply to a few middlewares, and people are antsy for 0.20.  Maybe we roll with it as is, and introduce this refactoring before 1.0.

ahjohannessen:
To be honest @rossabaker I only have a specific need wrt. CSRF. It can be solved similar to this:

```scala

  def filter2(predicate: Request[G] ⇒ Boolean = _.method.isSafe,
              r: Request[G],
              http: Http[F, G],
              f: G ~> F): F[Response[G]] =
    if (predicate(r)) {
      validate(r, http)
    } else {
      checkCSRFDefaultV2(r, http, f)
    }

  def validate2(predicate: Request[G] ⇒ Boolean = _.method.isSafe,
                f: G ~> F): Middleware[F, Request[G], Response[G], Request[G], Response[G]] = { http ⇒
    Kleisli { r: Request[G] ⇒
      filter2(predicate, r, http, f)
    }
  }

  def checkCSRFDefaultV2(r: Request[G], http: Http[F, G], fk: G ~> F): F[Response[G]] = {
    for {
      ht ← F.pure(r.headers.get(headerName).map(_.value))
      ft ← if (ht.isDefined) F.pure(ht) else {
            fk(r.as[UrlForm].map(_.values.get("X-TSec-Csrf").flatMap(_.headOption)))
           }
      r ← ft match {
           case Some(t) ⇒ checkCSRFToken(r, http, t)
           case None    ⇒ F.pure(onFailure)
         }
    } yield r
  }


```

rossabaker:
Yeah, that could be good.

ahjohannessen:
@rossabaker I took at stab at using a builder approach, WDYT?

server/src/main/scala/org/http4s/server/middleware/CSRF.scala
```diff
@@ -244,55 +239,156 @@ final class CSRF[F[_], G[_]] private[middleware] (
...
+        r.attemptAs[UrlForm].value.map(_.fold(_ => none[String], _.values.get(fieldName).flatMap(_.headOption)))
+
+      r.headers.get(headers.`Content-Type`) match {
+        case Some(headers.`Content-Type`(MediaType.application.`x-www-form-urlencoded`, _)) => nt(extractToken)
+        case _                                                                              => F.pure(none[String])
+      }
+    }
+
+    for {
+      fst <- F.pure(csrf.getHeaderToken(r))
```
ChristopherDavenport: Why is the `pure` necessary here, seems this should just be on the next line

ahjohannessen: Not sure why I did that, perhaps because a fold resulted in `v => Some(v)` - If you have a more elegant approach please share:) on the phone now

ChristopherDavenport: You're right, all of them I know of would involve at least one thunk. Here's hoping that someone else has a clean idea.

ahjohannessen:
:+1:

PR #2231, opened 2018-11-01
Fix off-by-one error in maxWaitQueueLimit

rossabaker:
We're not counting ourselves against the `maxWaitQueueLimit` when enqueuing a waiter, which results in us going one past the `maxWaitQueueLimit`.


aeons:
So, 'procedure syntax' actually means 'trailing commas' in this case :)

rossabaker:
I wish there were a Scala 2.11 plugin to support trailing commas.  I accidentally leave them all over the place, and actually like them for diff reduction.  Thanks for the fix!

PR #2234, opened 2018-11-01
Message cleanup

### LINKED ISSUE #2224: Clean up Message and subclasses
Currently, it's hard to get an overview of the methods on `Message`, `Request` and `Response`, due, in part, to the split of some methods to the `*Ops` traits. These traits are only used for the deprecated `Effect*Syntax` stuff, as far as I can see.

For example, we have `replaceAllHeaders` and `putHeaders` on `MessageOps`, and `withHeaders` on `Request`.

I know there's some rumbling about a typeclass reimplementation of `Message`, but I think this could be done now anyway.

I would like to remove the `*Ops` traits, but I don't really see how that can be done without removing the deprecated `Effect*Syntax` traits.

aeons:
This builds on top of #2210, and should not be merged before that.

This PR does the following:
- Remove deprecated Effect{Message,Request,Response}Syntax traits and associated objects
- Remove {Message,Request,Response}Ops and put the removed methods, sans unneeded implicit parameters, directly in the classes
- Deprecate `replaceAllHeaders`, pointing to `withHeaders` instead. The other direction of this was done in 0.10, apparently. Currently, all `withX` methods replace the `X`. For headers, we have `withHeaders` that replaces and `putHeaders` that adds.
- Deprecate `withType`, which takes a `MediaType` and just wraps it in a `Content-Type`
- Deprecate `withContentTypeOption`,  recommending to just do the `Option.fold` in user code
- Add `withoutAttribute` and `withoutTrailerHeaders` to complement the `with` variants
- Change some `def`s into `lazy val`s
- `filterHeaders`' scaladoc comment said that it would remove all headers that the predicate was true for, while doing the opposite. I have changed the comment to be correct and changed `withoutContentType` which depended on the behaviour described in the comment.

Many of the methods in the `*Ops` traits were `final`. I have kept that where it made sense, but I'm not sure it's needed. Can anyone clarify?

Closes #2224


aeons:
I'll rebase onto master once #2210 is merged. The diff here will also be quite messy until that happens

aeons:
I think the appveyor failure is unrelated.

core/src/main/scala/org/http4s/Message.scala
```diff
@@ -24,27 +22,27 @@ sealed trait Message[F[_]] extends MessageOps[F] { self =>
...
-    change(headers = f(headers))
+  def withHttpVersion(httpVersion: HttpVersion): Self =
+    change(httpVersion = httpVersion)
 
-  override def withAttribute[A](key: AttributeKey[A], value: A)(implicit F: Functor[F]): Self =
-    change(attributes = attributes.put(key, value))
+  def withHeaders(headers: Headers): Self =
+    change(headers = headers)
+
+  def withHeaders(headers: Header*): Self =
```
rossabaker: Is there a better naming idiom we could adopt here and avoid the overload?  I've grown more hostile to those, but don't have a naming suggestion.

core/src/main/scala/org/http4s/Message.scala
```diff
@@ -76,44 +74,153 @@ sealed trait Message[F[_]] extends MessageOps[F] { self =>
...
+    withAttribute(Message.Keys.TrailerHeaders[F], trailerHeaders)
+
+  final def withoutTrailerHeaders: Self =
+    withoutAttribute(Message.Keys.TrailerHeaders[F])
 
   /**
-    * The trailer headers, as specified in Section 3.6.1 of RFC 2616.  The resulting
-    * task might not complete unless the entire body has been consumed.
+    * The trailer headers, as specified in Section 3.6.1 of RFC 2616. The resulting
+    * F might not complete unless the entire body has been consumed.
```
rossabaker: Maybe `s/unless/until`?

core/src/main/scala/org/http4s/Message.scala
```diff
@@ -76,44 +74,153 @@ sealed trait Message[F[_]] extends MessageOps[F] { self =>
...
+  final def withType(t: MediaType)(implicit F: Functor[F]): Self =
+    withContentType(`Content-Type`(t))
+
+  final def withContentType(contentType: `Content-Type`): Self =
+    putHeaders(contentType)
+
+  final def withoutContentType: Self =
+    filterHeaders(_.isNot(`Content-Type`))
+
+  @deprecated("Use Option.fold instead", "0.20.0-M2")
```
rossabaker: We do provide the `Option` version on our client and server builders.  I'm not so sure about the deprecation here.

core/src/main/scala/org/http4s/Message.scala
```diff
@@ -76,44 +74,153 @@ sealed trait Message[F[_]] extends MessageOps[F] { self =>
...
+  final def withContentTypeOption(contentTypeO: Option[`Content-Type`]): Self =
+    contentTypeO.fold(withoutContentType)(withContentType)
+
+  lazy val contentType: Option[`Content-Type`] =
```
rossabaker: Are we sure that laziness pays off?  It adds weight to every request and response, and only is a win if read multiple times.  We've moved away from a lot of our prior use of lazies.

Furthermore, the risk of contention is low.  We can make these faster with the null init pattern, though it's a bit more verbose.

core/src/main/scala/org/http4s/Message.scala
```diff
@@ -76,44 +74,153 @@ sealed trait Message[F[_]] extends MessageOps[F] { self =>
...
+    * Does not introspect the body for media types that define a charset
+    * internally.
+    */
+  lazy val charset: Option[Charset] =
+    contentType.flatMap(_.charset)
+
+  lazy val isChunked: Boolean =
+    headers.get(`Transfer-Encoding`).exists(_.values.contains_(TransferCoding.chunked))
+
+  // Attribute methods
```
rossabaker: There's some grouping annotation in Scaladoc that I've seen used to nice effect elsewhere and comments like this make me think of it, but I've never figured out how to use it.

core/src/main/scala/org/http4s/Message.scala
```diff
@@ -292,13 +391,32 @@ sealed abstract case class Request[F[_]](
...
   def decodeWith[A](decoder: EntityDecoder[F, A], strict: Boolean)(f: A => F[Response[F]])(
       implicit F: Monad[F]): F[Response[F]] =
     decoder.decode(this, strict = strict).fold(_.toHttpResponse[F](httpVersion), f).flatten
 
+  /** Helper method for decoding [[Request]]s
+    *
+    * Attempt to decode the [[Request]] and, if successful, execute the continuation to get a [[Response]].
+    * If decoding fails, an `UnprocessableEntity` [[Response]] is generated.
+    */
+  final def decode[A](
```
rossabaker: I'm skeptical of the value of this and `decodeStrict`, but more people use them than I realize.  I don't know that I want to fight the deprecation battle for something so common, but I dislike having multiple ways to do it.

core/src/main/scala/org/http4s/Message.scala
```diff
@@ -359,28 +477,47 @@ final case class Response[F[_]](
...
+      headers = headers,
+      attributes = attributes
+    )
 
-  override def toString: String = {
-    val newHeaders = headers.redactSensitive()
-    s"""Response(status=${status.code}, headers=$newHeaders)"""
-  }
+  /** Add a Set-Cookie header for the provided [[ResponseCookie]] */
+  def addCookie(cookie: ResponseCookie): Self =
```
rossabaker: Why is it `addCookie` when it's `putHeaders`?  Is there a reason we should be inconsistent here?

core/src/main/scala/org/http4s/Message.scala
aeons: Alright, let's keep it then.

core/src/main/scala/org/http4s/Message.scala
aeons: No clue. If we decide against the laziness, then I just think we should be consistent and have defs for all of the things. And yeah, most of these are going to be read at most once.

core/src/main/scala/org/http4s/Message.scala
aeons: I'll take a look.

core/src/main/scala/org/http4s/Message.scala
aeons: It's `headers.put` underneath, but deciding on `add` instead would be fine by me.

core/src/main/scala/org/http4s/Message.scala
rossabaker: Wow, there's a lot more laziness than I thought.  I thought you were reintroducing more.

That said, I think it's (almost?) always a mistake.  For comparison's sake, I just checked out akka-http's `HttpMessage` and found no lazies.

core/src/main/scala/org/http4s/Message.scala
rossabaker: I guess `addHeaders` would be wrong because it replaces singleton headers.  I retract the suggestion.

core/src/main/scala/org/http4s/Message.scala
aeons: I would be 👍 on deprecating it. I guess we can look at reports for it for the milestones as a gauge for how much it's used.

core/src/main/scala/org/http4s/Message.scala
aeons: The `@group` annotation needs to go before each member. It looks very nice, but unless we have scaladoc for everything I'm not sure on the value.

[Example in doobie](https://github.com/tpolecat/doobie/blob/972d27d5135415e21317903ec6308082eb8b4ef1/modules/core/src/main/scala/doobie/hi/preparedstatement.scala)
[and the scaladoc](https://static.javadoc.io/org.tpolecat/doobie-core_2.12/0.6.0/doobie/hi/preparedstatement$.html)

core/src/main/scala/org/http4s/Message.scala
aeons: I don't have a better idea.

aeons:
Is there a point to having `final def` vs. `def` in a subclass of a `sealed trait`. Does it hint anything to the compiler about inlining?

core/src/main/scala/org/http4s/Message.scala
rossabaker: Okay, future improvement, then.

core/src/main/scala/org/http4s/Message.scala
rossabaker: Deprecation + scalafix is going to be a powerful combination for us for things that can be translated directly.  It would be nice if the two start going together once the scalafix lands.  I'm comfortable taking this on in a subsequent PR.

rossabaker:
It can theoretically help with static optimizations, but the JVM is pretty smart about what's actually overridden.  I would expect it to not make a practical difference at runtime, and it's a question of signaling intent.  Here, I don't think it's a huge deal either way.

aeons:
Alright. To be consistent I've made all of them non-final. :)

PR #2235, opened 2018-11-01
Add ability to set Cookie extension in CSRF middleware

jarreds:
Looking to add `SameSite=strict` to cookie.

This should maintain BC with existing code out in the wild.


jarreds:
Looks like this might actually have a subset of the functionality in https://github.com/http4s/http4s/pull/2228

jarreds:
I do prefer the setup in #2228. Closing this.

Issue #2206, opened 2018-10-25
Parse cookies with semicolon in the end


balatbn:
If we parse cookies using Cookie.from(req.headers), it will fail if the cookie string has a ";" in the end.

Example:
cookie: S=dfs43sdf3;visited=234435786;

It is possible that as per HTTP spec, having a ";" in the end might be invalid. However, some clients (we tested with Android clients) add a semicolon in the end automatically.

So, it be would be nice, if we can ignore ending ";" during parsing.

rossabaker:
It is [illegal](https://tools.ietf.org/html/rfc6265#page-13)

```
   cookie-header = "Cookie:" OWS cookie-string OWS
   cookie-string = cookie-pair *( ";" SP cookie-pair )
```

But I don't see a reason we couldn't relax that.  The parser is [here](https://github.com/http4s/http4s/blob/b3849cf830aa5a1bb51dd8d594e2d8edc45eaee9/core/src/main/scala/org/http4s/parser/CookieHeader.scala#L43).

PR #2237, opened 2018-11-02
Allow cookie string to have ';' at the end

bhop:
fixes: #2206


rossabaker:
This is more of an enhancement than a bugfix, and 0.18.x is on its last legs.  If someone wants it, we can merge a backport PR.  Until then, I'm merging as-is.

balatbn:
Thanks for the fix. We have temporarily "fixed" the original issue(2206) by removing ";" in our code.

We can wait until next major release.

PR #2240, opened 2018-11-03
 Unify backend builders, adding allocate to server builders

rossabaker:
Builds on #2239.

* Introduces an internal `BackendBuilder` trait, which improves API cohesion through <whispering>inheritance</whispering>.
* Puts the type parameter on the builder, not the `resource`/`stream`/`allocate` methods
* Adds `allocate` to the servers, per popular request.  I still don't like it.
* Doesn't yet apply to JettyClient or AsyncHttpClient, which aren't builders


rossabaker:
I anticipate this might make @ChristopherDavenport slightly grumpy, so I'll hold off a little bit in case he wants to have an OOP fight.

core/src/main/scala/org/http4s/internal/BackendBuilder.scala
```diff
@@ -0,0 +1,32 @@
+package org.http4s.internal
+
+import cats.effect.{Bracket, Resource}
+import fs2.Stream
+
+private[http4s] trait BackendBuilder[F[_], A] {
```
ChristopherDavenport: Why extend rather than just build instances of this like a typeclass. What does this extension buy us? resource => methods? Why not just expose that rather than this inheritance scheme?

rossabaker: Pro: Traits are better for method discoverability. All of these methods will appear in the scaladoc and other primitive tooling.

Con: Typeclasses are better for ad hoc polymorphism.  Someone could get the derived functions even if they didn't call the method `resource`.  (But I'd frown at any such builder.)

I prioritized discoverability over extensibility in this context.  If I saw [scaladoc to understand extension methods](https://stackoverflow.com/questions/21406207/configure-scaladoc-to-include-extension-methods) actually work, I'd probably flip.

rossabaker: I left out two more cons.

The first is that it's more verbose.  I don't really care, because that's our pain, not users':

```scala
trait BackendBuilder[F[_], A] {
  type R
  def resource(builder: A): Resource[F, R]
  def stream(builder: A): Stream[F, R] =
    Stream.resource(resource(builder))
  def allocate(builder: A)(implicit F: Bracket[F, Throwable]): F[(R, F[Unit])] =
    allocated(resource(builder))
}

object BackendBuilder {
  def apply[F[_], A](implicit ev: BackendBuilder[F, A]): BackendBuilder[F, A] = ev

  // These probably belong in org.http4s.implicits._
  implicit def backendBuilderSyntax[F[_], A](builder: A)(implicit A: BackendBuilder[F, A]) =
    new BackendBuilderOps(builder)
  class BackendBuilderOps[F[_], A](builder: A)(implicit val A: BackendBuilder[F, A]) {
    def resource: Resource[F, A.R] = A.resource(builder)
    def stream: Stream[F, A.R] = A.stream(builder)
    def allocate(implicit F: Bracket[F, Throwable]): F[(A.R, F[Unit])] = A.allocate(builder)
  }
}
```

The second issue is that it imposes an import tax on `builder.stream` and `builder.allocate` to get the syntax into scope.  This is more pertinent to discovery than the scaladoc issue.

Issue #2197, opened 2018-10-19
Flaky test: BlazeHttp1ClientSpec GET: /simple


rossabaker:
```
[info] BlazeHttp1ClientSpec
[info] + Execute GET: /chunked (496 ms)
[info] + Execute GET: /not-found (510 ms)
[info] + Execute GET: /no-content (522 ms)
[error] ! Execute GET: /simple (10 seconds, 313 ms)
[error]  java.util.concurrent.TimeoutException: Response header timeout after 10000 ms. (ResponseHeaderTimeoutStage.scala:27)
[error] org.http4s.blazecore.ResponseHeaderTimeoutStage$$anon$1.run(ResponseHeaderTimeoutStage.scala:27)
[info] + Execute GET: /delayed (1 second, 44 ms)
```

https://travis-ci.org/http4s/http4s/jobs/443905281

rossabaker:
An [appveyor version](https://ci.appveyor.com/project/ChristopherDavenport/http4s-m0hqu/builds/19776714/job/g0dfrnmj86h3kb9u#L4801) that failed on /no-content.  This was first seen on appveyor the blaze-0.14.0-M9 upgrade, though it's not an SSL test, so it seems unlikely to be related.

rossabaker:
https://travis-ci.org/http4s/http4s/jobs/447372204#L3826

rossabaker:
I think this might be an actual problem.  I want to get this resolved before 0.20.0.

rossabaker:
https://travis-ci.org/http4s/http4s/jobs/449167552#L3758

rossabaker:
Probably related: https://travis-ci.org/http4s/http4s/jobs/449823379#L3298

rossabaker:
We haven't seen this in a while.  Optimistically closing.  I have a hunch that #2258 was helpful here.

rossabaker:
[Nuts](https://travis-ci.org/http4s/http4s/jobs/456725017#L3927).

mahe-ymc:
Happens occasionally in one of our prod apps using org.http4s:blaze-http_2.12:0.14.0-M12. Note that this happens although we set the timeout to 60 seconds, which is certainly pretty high. Additionally, the HTTP method seems irrelevant as it also occurs using PUT.

pandaforme:
I have the same issue in our production. The version we use is `0.20.0`.
Error logs:
```
Response header timeout after 10000 ms.
[...]

PR #2243, opened 2018-11-05
Remove or privatize server builder *support traits

rossabaker:
The server builder traits were a poor means of abstraction that few were using in the wild anyway.

* Eliminate marker traits for server builders
* Privatize `ServerBuilder`, which does still contribute to consistency
* Move defaults into `org.http4s.server.defaults` package


rossabaker:
Build failure is #2197.

Issue #2193, opened 2018-10-19
PoolManager waitQueue not cleaned up


ptrlaszlo:
We are using http4s 0.19.0-M1 and we observed an issue with the http client. If we call an API which is slow, the waitQueue of the PoolManager fills up and start to throw "Max wait length reached, not scheduling". This seems to be OK, but the wait queue seems to be never freed up. So once we enter this state, there is no going back.

Some details:
- `borrow` ends up in a [situation](https://github.com/http4s/http4s/blob/v0.19.0-M1/client/src/main/scala/org/http4s/client/PoolManager.scala#L177-L179) where the `idleQueues` is empty, but the `waitQueue` is full
- the test configuration we use: `maxTotal: 2, maxWaitQueueLimit: 5, responseHeaderTimeout: 18 seconds, requestTimeout: Inf`
- when the wait queue is not freed up the attributes are: `isClosed: false, curTotal: 2, size of allocated: 1, size of idleQueues: 0, size of waitQueue: 6`

I'm happy to provide more information if needed.

rossabaker:
I'm having trouble reproducing this.  Does anyone have a runnable example?

rossabaker:
Is this happening when the requests are timing out, not just slow?  I think that is the key here.

rossabaker:
Thinking this may be related to #2068.

rossabaker:
Trying to reproduce this on 0.18.x:

```scala
import cats.effect._
import fs2._
import java.util.concurrent.atomic._
import org.http4s._
import org.http4s.client.blaze._
import scala.concurrent.ExecutionContext.Implicits.global
import scala.concurrent.duration._

object Issue2193 {
  def main(args: Array[String]) = {
    val int = new AtomicInteger(0)
    (for {
      c <- Http1Client[IO](BlazeClientConfig.defaultConfig.copy(
        maxTotalConnections = 2,
        maxWaitQueueLimit = 5,
        responseHeaderTimeout = 5.seconds,
        requestTimeout = Duration.Inf
      ))
      submit = c.status(Request[IO](uri = Uri.uri("http://httpbin.org/delay/10"))).attempt
      step = submit
      s <- Stream(Stream.eval(step)).repeat.covary[IO].join(8).take(20)
        .observe1(x => IO(println(">>> " + int.incrementAndGet + " " + x))).compile.drain
      _ <- IO.sleep(10.seconds)
[...]

PR #2248, opened 2018-11-06
Release v0.18.21

rossabaker:
I haven't reproduced #2193 yet.  No reason to sit on a working bugfix in the meantime.


rossabaker:
Targeted wrong branch, now Travis won't run.

PR #2258, opened 2018-11-08
Fix deadlocks on failures with parTraverse and parSequence

ngbinh:
Both of these tests fail by timing out on my local machine:

```scala
[info] BlazeClientSpec                                                                                     
[info] Blaze Http1Client should                                     
[info]   + raise error NoConnectionAllowedException if no connections are permitted for key (789 ms)
[info]   + make simple https requests (966 ms)                                                                                                      
[info]   + behave and not deadlock (1 second, 34 ms)                                         
[error]   x behave and not deadlock on failures with parTraverse (30 seconds, 245 ms)                      
[error]    None is not Some (BlazeClientSpec.scala:151)                               
[error] org.http4s.client.blaze.BlazeClientSpec.$anonfun$new$18(BlazeClientSpec.scala:151)
[error]   x behave and not deadlock on failures with parSequence (30 seconds, 308 ms)                              
[error]    None is not Some (BlazeClientSpec.scala:184)                                                                                  
[error] org.http4s.client.blaze.BlazeClientSpec.$anonfun$new$31(BlazeClientSpec.scala:184)                                                                                                                  
[info]   + obey response header timeout (30 seconds, 374 ms)                                                                                                                                                
[info]   + unblock waiting connections (30 seconds, 417 ms)                                                                                                       
[info]   + reset request timeout (30 seconds, 462 ms)                  
[info]   + drain waiting connections after shutdown (1 second, 140 ms)                                                                   
[info]   + cancel infinite request on completion (1 second, 193 ms)                                                                                                                                         
[info] Total for specification BlazeClientSpec                  
[info] Finished in 31 seconds, 904 ms                    
[info] 10 examples, 38 expectations, 2 failures, 0 error  
```

PR #2260, opened 2018-11-08
Fix leak on a canceled connection

### LINKED ISSUE #2193: PoolManager waitQueue not cleaned up

rossabaker:
This _potentially_ fixes #2193.  It does not appear to fix #2258.

If it fixes neither, it still fixes a bug.

### DIFF
--- a/blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClient.scala
+++ b/blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClient.scala
@@ -75,26 +75,31 @@ object BlazeClient {
                 F.never[TimeoutException]
             }
 
-            next.connection.runRequest(req, idleTimeoutF).attempt.flatMap {
-              case Right(r) =>
+            next.connection
+              .runRequest(req, idleTimeoutF)
+              .flatMap { r =>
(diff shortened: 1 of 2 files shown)


orium:
Hi @rossabaker.

I think we have been hit by this bug (#2193).  Is it possible to get a new 0.18.x release with this fix?

Thanks!

Issue #2096, opened 2018-09-17
Consider moving from Seq to List in value types


bpholt:
As discussed in #2086, the new monoid instance for `UrlForm` has to duplicate some code from cats-kernel because there is no upstream `Monoid[Seq[?]]` instance. 

It would be a breaking change, but should the definition of `UrlForm` be changed (maybe along with any other type using `Seq`)?

jmcardon:
Related: https://github.com/http4s/http4s/issues/1695

rossabaker:
`List`, `Vector`, and `Chain` are the three obvious choices.  `Chain` is great if you're combining them often, but I'm not sure how common it is, even if it motivated the other ticket.  It's not random access, so I'm not too keen on `Vector`.  Maybe `List`, and be smart about how we build them?

ChristopherDavenport:
What is the downside to `Chain`?

rossabaker:
I don't have a firm grasp on its constant factors, and these are generally going to be small enough that Big O isn't critical.  It definitely could be combined effectively without any special `ListBuffer` foofery that would be tempting for a `combineAll` override.

rossabaker:
Partially done by #2262.  Demilestoning, because we shouldn't hold up the rest of the release on it.

calvinbrown085:
Just looking for other work here, would this be a candidate? https://github.com/http4s/http4s/blob/master/dsl/src/main/scala/org/http4s/dsl/impl/Path.scala#L47

rossabaker:
Expounded on a better `Http4sDsl#Path` at #3549.  `Uri.Path` was a mere `String` when Calvin commented.  (And my apologies for missing it!).

I'm struggling to find any that remain in the model representations.  Most of them became `List`s (common in headers) or `Query` and `Uri.Path` are built on a `Vector`.  There are some references to `Seq` in terms of derived properties, like `multiParams`, which let us change our mind later at the expense of the client being able to make assumptions.

rossabaker:
Duplicate of #1695.

PR #2262, opened 2018-11-10
change UrlForm to use Chain instead of Seq

bpholt:
Chain has a Monoid instance available to it, while Seq does not. Partially addresses #2096.

This was a pretty mechanical conversion and might be able to be optimized.

### DIFF
--- a/core/src/main/scala/org/http4s/UrlForm.scala
+++ b/core/src/main/scala/org/http4s/UrlForm.scala
@@ -1,6 +1,7 @@
 package org.http4s
 
 import cats._
+import cats.data._
 import cats.effect.Sync
 import cats.implicits.{catsSyntaxEither => _, _}
 import org.http4s.headers._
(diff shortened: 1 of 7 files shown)


core/src/main/scala/org/http4s/UrlForm.scala
```diff
@@ -1,30 +1,31 @@
...
 import scala.io.Codec
 
-class UrlForm private (val values: Map[String, Seq[String]]) extends AnyVal {
+class UrlForm private (val values: Map[String, Chain[String]]) extends AnyVal {
   override def toString: String = values.toString()
 
-  def get(key: String): Seq[String] =
-    this.getOrElse(key, Seq.empty[String])
+  def get(key: String): Chain[String] =
+    this.getOrElse(key, Chain.empty[String])
```
ChristopherDavenport: I have zero clue how this works calling `this` rather than `values`

rossabaker: Line 18

ChristopherDavenport: Thank you, I feel more sane now.

Issue #2254, opened 2018-11-06
Support valueless query parameters in the DSL


howyp:
I couldn't find a way to match valueless query parameters using the http4s DSL. 

These are the sort that are used as a flag ie:

`http://example.com/foo?bar`
vs 
`http://example.com/foo`

There’s no value for `bar`, but it’s presence or lack of indicates that the resource should behave differently. I tried an `OptionalQueryParamDecoderMatcher[Unit](“bar”)`, but that supplies `None` even if the parameter without a value - it returns `Some(())` only if an (potentially empty) value is supplied.

howyp:
As was very helpfully suggested by @andimiller, this is a good workaround:

```scala
case req @ GET -> Root / blah if req.params.contains("foo") => ...
```

But I think it would be nice to find a way to do this directly in the DSL

howyp:
Looking at the existing code, I guess we see the behaviour I described above because the `_.headOption` on the sequence of values occurs before the `QueryParamDecoder` is ever called:

```scala
abstract class OptionalQueryParamDecoderMatcher[T: QueryParamDecoder](name: String) {
  def unapply(params: Map[String, Seq[String]]): Option[Option[T]] =
    params
      .get(name)
      .flatMap(_.headOption)
      .traverse(s => QueryParamDecoder[T].decode(QueryParameterValue(s)))
      .toOption
}
```

So I guess if we want to fix this it may mean changing the API between the matchers and decoders ☹️

rossabaker:
Fixed by #2266.

PR #2266, opened 2018-11-11
Support flag query parameters in DSL

howyp:
Adds an new matcher to the DSL supporting option flag-style query parameters as discussed in  #2254.

### DIFF
--- a/dsl/src/test/scala/org/http4s/dsl/PathInHttpRoutesSpec.scala
+++ b/dsl/src/test/scala/org/http4s/dsl/PathInHttpRoutesSpec.scala
@@ -32,6 +32,8 @@ object PathInHttpRoutesSpec extends Http4sSpec {
 
   object MultiOptCounter extends OptionalMultiQueryParamDecoderMatcher[Int]("counter")
 
+  object Flag extends FlagQueryParamMatcher("flag")
+
   val app: HttpApp[IO] = HttpApp {
     case GET -> Root :? I(start) +& L(limit) =>
       Ok(s"start: $start, limit: ${limit.l}")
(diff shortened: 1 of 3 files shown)


howyp:
Fixes #2254

rossabaker:
I shut down the appveyor build, so don't worry about that failure.

dsl/src/test/scala/org/http4s/dsl/PathInHttpRoutesSpec.scala
```diff
@@ -235,5 +240,23 @@ object PathInHttpRoutesSpec extends Http4sSpec {
...
     }
+    "optional flag parameter when present" in {
+      val response =
+        serve(Request(GET, Uri(path = "/flagparam", query = Query.fromString("flag"))))
+      response.status must_== Ok
+      response.as[String] must returnValue("flag present")
+    }
+    "optional flag parameter when present with a value" in {
+      val response =
+        serve(Request(GET, Uri(path = "/flagparam", query = Query.fromString("flag=1"))))
```
rossabaker: There's no desire to implement "truthiness" here, correct?  This would still pass if `flag=0`?  I think it should, and does, but just double checking the intent.

howyp: Yes that was my intent. I was more trying to show that `true` is returned whether there is a value or not - the value is ignored, all that matters is if the parameter key is there or not. I guess I could express that better in this test.

I actually wavered a bit on whether `boolean` was the right return type. WDYT? The alternative was adding a little ADT with `Present` and `Absent` objects, which might be a bit clearer. 

Or ... we could have some arrangement where you get a different type if:

1. The parameter is completely absent
2. The parameter is present but with no value
3. The parameter is present but with a value (and then I guess you'd want the capabilities we have elsewhere for decoding the value)

But IMHO that seemed a bit overcomplicated? I'm not sure if anyone would actually use parameters in this way.

rossabaker: `Present` and `Absent` would make clear that it has no opinions on truthiness, but would generally be a bit more inconvenient to use.

I share your uncertainty on the ternary extractor.  It looks good on paper but I'm not sure about on practice.

I don't deal with APIs that use query parameters often, so I don't have a lot of practical experience here.  This is merged, and it made things better.  Let's try it, and if it gives us a rash, we can fix it before 1.0.

PR #2269, opened 2018-11-13
Client dsl with body first

ChristopherDavenport:
Builds on #2202


ChristopherDavenport:
Call out for help, I have zero scalafix knowledge still.

rossabaker:
I want to get this feature in place and I want to release.  Should we merge now and have a scalafix to follow?

ChristopherDavenport:
I think that seems the best idea. Help out migration as soon as we can, get the baseline code in place first.

PR #2270, opened 2018-11-13
Refresh MimeDB

rossabaker:
`application/vnd.exstream-empower+zip` users, rejoice!


core/src/main/scala/org/http4s/MimeDB.scala
```diff
@@ -2931,8 +2933,6 @@ private[http4s] trait MimeDB {
         new MediaType("application", "vnd.oma.dcdc", Compressible, NotBinary)
       lazy val `vnd.oma.dd2+xml`: MediaType =
         new MediaType("application", "vnd.oma.dd2+xml", Compressible, NotBinary, List("dd2"))
-      lazy val `vnd.oma.drm.risd+xml`: MediaType =
```
cquiroz: Don't these break MiMa?

rossabaker: They will, which is why I wanted to give it a last refresh before 0.20.

cquiroz: right, ok

PR #2276, opened 2018-11-14
Make scalafix github based

### LINKED ISSUE #2274: Make scalafix github-based
Scala steward is going to try to support opening prs with scalafixes (see https://github.com/fthomas/scala-steward/issues/149). They are initially going to support github based scalafixes. I think it would be a nice idea to convert our scalafix from binary to github based so the day 0.20 gets released and scalasteward supports it (whichever is the latest) people get flooded with nice prs and have almost a 0-step migration :) 

What do you think?

amarrella:
Fixes #2274


amarrella:
If you want to test it, just run:
```
sbt ";scalafixEnable; scalafix github:amarrella/http4s/v0_20?sha=85ef964"
```

amarrella:
@rossabaker this is done. I just moved the code around :) It's not beautiful as it's all in one file but it doesn't need to be. It's much easier to use now from a user's (and scala-steward :D ) perspective

rossabaker:
A ramification of this is that it's no longer tested by Travis, right?

Adding to 0.20 milestone since it affects what we publish.  I don't like removing published artifacts in the middle of a series.

amarrella:
> A ramification of this is that it's no longer tested by Travis, right?

I guess we could add a line to the script in `.travis.yml` if that's what we want? I'm not sure how we specify the different build.sbt, maybe we can trick travis via `cd`

amarrella:
> Adding to 0.20 milestone since it affects what we publish. I don't like removing published artifacts in the middle of a series.

Fair enough, we could also leave the binary and publish both actually. We just need to point the projects in the main build sbt to the subdirectory. This would solve the travis problem too

rossabaker:
If we want it to have CI, without polluting the main build or Maven Central, we could also add an environment variable to the build matrix.  And then the script would be `cd $ROOT_DIR && sbt ++$TRAVIS_SCALA_VERSION ci`.  We'd lose `sbt ci` from the top-level running all the tests, but very few contributors (including me) are vigilant about doing that.

If running scalafix through CI is not typical, I'd accept that, too.  I'm commenting on developer instinct, not scalafix experience.  I'm going to be highly deferential here.

amarrella:
> If running scalafix through CI is not typical, I'd accept that, too. I'm commenting on developer instinct, not scalafix experience. I'm going to be highly deferential here.

I think having ci is a good idea if possible. Though it doesn't look that e.g. cats have any ci for that. 
Your call :)

rossabaker:
Let's keep CI one way or another.  I hate to see bitrot.

amarrella:
Went the matrix route. I think it's the cleanest

Issue #2282, opened 2018-11-16
ClientMetrics will mark metrics before connection has closed


tbrown1979:
Metrics are being recorded here before we actually close the connection: https://github.com/http4s/http4s/blob/80df97e1fdb1d764059acafcd1f4331135964560/client/src/main/scala/org/http4s/client/middleware/Metrics.scala#L46

The finalizer for the client.run won't run until after the metrics have been marked. When you use the `Resource` you'll mark your metrics before you've finished consuming the body, which could potentially skew the metrics. 

A fix could be marking the metrics as a part of the `client.run`'s `Resource` `release` function. 

Conversation in gitter: https://gitter.im/http4s/http4s?at=5beef2645f5c925040699ecb

tbrown1979:
This is something I can look into fixing.

PR #2283, opened 2018-11-16
Record Metrics after `client.run` Resource has released 

tbrown1979:
Fixes a bug with the `Client` `Metrics` code where we would decrement active requests and record total time before the `Resource` returned from the `client.run` was actually `release`d. This means our metrics were not accurate. 

The fix here is to perform both of these actions _after_ the `client.run` has `.release`d. To do so, both the `recordTotalTime` and incr/decr of the active requests have been turned into `Resource`s of their own, that `.release` after the `client.run`. This results in the behavior we expect. A test has been added to demonstrate the bug, and to show that it is now fixed. 

fixes https://github.com/http4s/http4s/issues/2282

### DIFF
--- a/client/src/main/scala/org/http4s/client/middleware/Metrics.scala
+++ b/client/src/main/scala/org/http4s/client/middleware/Metrics.scala
@@ -42,19 +42,24 @@ object Metrics {
       req: Request[F])(implicit F: Sync[F], clock: Clock[F]): Resource[F, Response[F]] =
     (for {
       start <- Resource.liftF(clock.monotonic(TimeUnit.NANOSECONDS))
-      _ <- Resource.liftF(ops.increaseActiveRequests(classifierF(req)))
+      _ <- Resource.make(ops.increaseActiveRequests(classifierF(req)))(_ =>
+        ops.decreaseActiveRequests(classifierF(req)))
       resp <- client.run(req)
       end <- Resource.liftF(clock.monotonic(TimeUnit.NANOSECONDS))
       _ <- Resource.liftF(ops.recordHeadersTime(req.method, end - start, classifierF(req)))
-      _ <- Resource.liftF(ops.decreaseActiveRequests(classifierF(req)))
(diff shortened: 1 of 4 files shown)


tbrown1979:
There's a couple issues with this as present. 
1) The body's stream finalizer isn't running for non successful responses in the Dropwizard tests.
2) The decrement for active requests is happening twice now that it's cleverly a Resource.
3) The request headers are for some reason not reporting correctly for the test. Not sure what's up here yet. 

I'll look into fixing these sometime this weekend.

client/src/main/scala/org/http4s/client/middleware/Metrics.scala
```diff
@@ -41,20 +44,28 @@ object Metrics {
       classifierF: Request[F] => Option[String])(
       req: Request[F])(implicit F: Sync[F], clock: Clock[F]): Resource[F, Response[F]] =
     (for {
+      statusRef <- Resource.liftF(Ref.of[F, Option[Status]](None))
       start <- Resource.liftF(clock.monotonic(TimeUnit.NANOSECONDS))
-      _ <- Resource.liftF(ops.increaseActiveRequests(classifierF(req)))
+      _ <- Resource.make(ops.increaseActiveRequests(classifierF(req)))(_ =>
+        ops.decreaseActiveRequests(classifierF(req)))
+      _ <- Resource.make(F.delay(())) { _ =>
```
rossabaker: Can the first parameter just be an `F.unit`?

tbrown1979: Done! 😄

tbrown1979:
@rossabaker Added a test in [this](https://github.com/http4s/http4s/pull/2283/commits/55a0146017f63a6d2ce31acdd98d16b0e287209b) commit that fails with the old code but passes with the new code. It's not very pretty, but it seems to do the job. If there's any improvements you see please let me know!

PR #2303, opened 2018-12-04
Proposal: Make `orNotFound` return a `Http`

kubukoz:
This will most likely be incompatible in a variety of ways, but maybe we can make this kind of change since 0.20 is breaking anyway.

This should be fine, unless we assume that there are users who call `orNotFound` on Kleislis that don't have `Request[F]` in the input.


rossabaker:
Ugh.  I was not thinking clearly when I looked earlier.  I thought we were going the other direction, generalizing from `Http` to `Kleisli`, so it works with any input type.  I'm sorry about that.

Isn't this change going to make it not work with things like an authed service?

kubukoz:
Oh. Well.

Yeah, it probably is. I didn't realise that.

PR #2309, opened 2018-12-05
`Retry` client middleware: don't retry on `WaitQueueTimeoutException`

agustafson:
It seems odd to retry on a `TimeoutException`.

`PoolManager` throws a `TimeoutException` when a request has been in the wait queue for too long: https://github.com/http4s/http4s/blob/master/client/src/main/scala/org/http4s/client/PoolManager.scala#L209 so is something that might be a problem.

### DIFF
--- a/client/src/main/scala/org/http4s/client/PoolManager.scala
+++ b/client/src/main/scala/org/http4s/client/PoolManager.scala
@@ -16,7 +16,8 @@ final case class WaitQueueFullFailure() extends RuntimeException {
   def message: String = "Wait queue is full"
 }
 
-case object WaitQueueTimeoutException extends TimeoutException("In wait queue for too long, timing out request.")
+case object WaitQueueTimeoutException
+    extends TimeoutException("In wait queue for too long, timing out request.")
 
 private final class PoolManager[F[_], A <: Connection[F]](
     builder: ConnectionBuilder[F, A],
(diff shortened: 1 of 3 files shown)


rossabaker:
I agree on the wait queue.

What about a response header timeout or an idle timeout?  If we're willing to wait up to a minute for a response, might we set a timeout to a couple seconds, so other requests get through, and then retry 10, 20, 30 seconds later?  Or would we always set those other timeouts to a minute?  I could see it both ways.

We could distinguish these cases by subtyping TimeoutException, which might be a decent idea anyway.

agustafson:
Thanks @rossabaker - that's a good shout about sub-typing the `TimeoutException`. I'll add a specific sub-type for being in the wait queue too long (`WaitQueueTimeoutException`?)

So for the moment are we saying that if we subtype the wait queue exception, and don't retry when that exception occurs, then the PR will be ok? That certainly feels safer than catching all `TimeoutException`s.

As for the other examples I'm not quite sure I understand what you're referring to, but if I find any other examples of a `TimeoutException` in the codebase that could benefit from sub-typing then I'll create sub-types, as that seems to be a decent idea as you say. But that's prolly worthy of another PR.

Thanks for the feedback. Will amend.

agustafson:
Great, thanks Ross. I'll have a crack at that next week.

agustafson:
Hey @rossabaker I've created `WaitQueueTimeoutException`. Lemme know if there's anything else I can do to improve/change the PR.

agustafson:
Ah, ok thanks. Is the bike-shedding conversation something that's a separate conversation or can we start the flame war here?

Is the bike-shedding related to sub-typing the other `TimeoutException`s or the question around response header timeout or an idle timeout?

PR #2310, opened 2018-12-06
Set an upper limit for -Ybackend-parallelism

aeons:


aeons:
My new workstation is a 10 core xeon :)

PR #2315, opened 2018-12-15
Updated scalafix dependencies

amarrella:


rossabaker:
One thing I noticed while updating the changelog is that the version will always be one step behind.  We can't bump this before it's in Maven Central.  I think it's fine, because the scalafix can be pulled from GitHub.

Maybe we could automate this by finding the most recent tag?

PR #2316, opened 2018-12-15
Custom Json Error messages for Argonaut and Circe

zarthross:
The direction I was taking was to allow overriding of the error messages in the traits .
I'm open to using implicits instead, just need suggestions.

- [X] Jawn Custom Parse Error Messages
- [x] Argonaut Custom Decode Errors
- [x] Circe Custom Decode Errors

### DIFF
--- a/argonaut/src/main/scala/org/http4s/argonaut/ArgonautInstances.scala
+++ b/argonaut/src/main/scala/org/http4s/argonaut/ArgonautInstances.scala
@@ -7,26 +7,28 @@ import cats.Applicative
 import cats.effect.Sync
 import org.http4s.argonaut.Parser.facade
 import org.http4s.headers.`Content-Type`
-import org.http4s.jawn.JawnInstances
+import jawn.JawnInstances
+import _root_.jawn.ParseException
+import org.http4s.argonaut.ArgonautInstances.DecodeFailureMessage
 
 trait ArgonautInstances extends JawnInstances {
   implicit def jsonDecoder[F[_]: Sync]: EntityDecoder[F, Json] =
     jawnDecoder
+
+  protected def jsonDecodeError: (Json, DecodeFailureMessage, CursorHistory) => DecodeFailure =
+    ArgonautInstances.defaultJsonDecodeError
 
   def jsonOf[F[_]: Sync, A](implicit decoder: DecodeJson[A]): EntityDecoder[F, A] =
     jsonDecoder[F].flatMapR { json =>
       decoder
         .decodeJson(json)
         .fold(
-          (message, history) =>
-            DecodeResult.failure(
-              InvalidMessageBodyFailure(
-                s"Could not decode JSON: $json, error: $message, cursor: $history")),
+          (message, history) => DecodeResult.failure(jsonDecodeError(json, message, history)),
           DecodeResult.success(_)
         )
     }
 
-  protected def defaultPrettyParams: PrettyParams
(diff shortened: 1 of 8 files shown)


zarthross:
@rossabaker Switched to a builder pattern to see what it would look like.  I can continue with that if I get the :+1:  up.

argonaut/src/main/scala/org/http4s/argonaut/ArgonautInstances.scala
```diff
@@ -60,9 +63,37 @@ trait ArgonautInstances {
...
 object ArgonautInstances {
-  def withPrettyParams(pp: PrettyParams): ArgonautInstances =
-    new ArgonautInstances {
-      def defaultPrettyParams: PrettyParams = pp
-    }
+  type DecodeFailureMessage = String
+  def withPrettyParams(pp: PrettyParams): ArgonautInstancesBuilder =
+    builder.withPrettyParams(pp)
+
+  def builder: ArgonautInstancesBuilder = ArgonautInstancesBuilder()
```
rossabaker: Since the builder is immutable, this could even just be a val.

argonaut/src/main/scala/org/http4s/argonaut/ArgonautInstances.scala
```diff
@@ -60,9 +63,37 @@ trait ArgonautInstances {
...
 }
 
+sealed case class ArgonautInstancesBuilder private[argonaut] (
+    override val defaultPrettyParams: PrettyParams = PrettyParams.nospace,
+    override val jsonDecodeError: (Json, String, CursorHistory) => DecodeFailure =
+      ArgonautInstances.defaultJsonDecodeError,
+    override val jawnParseExceptionMessage: ParseException => DecodeFailure =
+      JawnInstances.defaultJawnParseExceptionMessage,
+    override val jawnEmptyBodyMessage: DecodeFailure = JawnInstances.defaultJawnEmptyBodyMessage
+) extends ArgonautInstances {
```
rossabaker: It's a little strange for a "builder" to extend what it's building instead of have a terminal method that builds it, but I don't see the harm.  If there's no terminal `build` method, maybe these `with` methods should go straight onto `ArgonautInstances` and just make that the abstract class?

argonaut/src/main/scala/org/http4s/argonaut/ArgonautInstances.scala
```diff
@@ -60,9 +63,37 @@ trait ArgonautInstances {
   }
 }
 
+sealed case class ArgonautInstancesBuilder private[argonaut] (
```
rossabaker: If we make this class abstract, we can add more fields in future revisions without breaking binary compatibility.

argonaut/src/main/scala/org/http4s/argonaut/ArgonautInstances.scala
zarthross: I wanted to keep the `ArgonautInstances` a trait so it could still be used as a mixin.  I can definitely change the semantics of the builder to include a terminator so its less weird.

rossabaker:
This seems like it's on the right track to me.

zarthross:
@rossabaker Hey, mind looking this over again?  I think I've got circe and argonaut done.

I would think if we want to do the same for the following it should probably be a separate PR.
- boopickle
- json4s-*
- play-json

zarthross:
Note: Last commit failed to build for what appears to be the same failure that's plaguing the master branch right now.

PR #2318, opened 2018-12-18
Baseline Transition for Vault Use

ChristopherDavenport:
This is a baseline transition for Vault use.

- Names left the same, do we want to change the name `attributes`? 
- UnsafeCreated the keys to global singletons. Do we want to control the control flow additionally?


ChristopherDavenport:
I'm just ecstatic that its such a small diff and it compiled and tested clean.

rossabaker:
I think i give Vault an edge over AttributeMap because, even if we do `.unsafeRunSync()`, it makes us sign the waiver.  It's a slight improvement, but it is an improvement.

ChristopherDavenport:
I don't see a need for vault to change, but could make java style methods as well which cound reduce additional noise. 

I don't mind doing the work here after the other PR's are all in with changes to attribute maps.

As far as the fact that it currently only does it this way, that was intentional. I wanted to sign the waiver to start off with, but later we could make middlewares return tuples with the key to extract the values generated by that middleware etc, which I believe would often be a better pattern. However I thought that the minimal piece was ideal.

As per the library dependency. Utilizing a shared tool means that this can and will translate with other peoples vault use. Where as an internal one means that conversions would be required to work with the library.

I think it has value, but do understand the reticence.

Issue #2273, opened 2018-11-14
Routes with `"" /:` don't match in 0.20


randm-ch:
After [chatting with](https://gitter.im/http4s/http4s?at=5be9b9c0bb06d73a99629646) @rossabaker on Gitter I'm opening this issue. Given the following code:

```import cats.implicits._
import cats.effect._
import org.http4s._
import org.http4s.dsl.Http4sDsl
import org.http4s.implicits._
import org.http4s.server.Router
import org.http4s.server.blaze._

object main extends IOApp with Http4sDsl[IO] {

  val A: Async[IO] = implicitly

  val httpApp = Router(
    "/" -> HttpRoutes.of[IO] {
      case GET -> Root => Ok("yes")
      case GET -> "" /: path => Ok(s"yes: $path")
    },
    "/other" -> HttpRoutes.of[IO] {
      case GET -> Root => Ok("yes")
      case GET -> "" /: path => Ok(s"yes: $path")
    }
  ).orNotFound

  def run(args: List[String]): IO[ExitCode] =
    BlazeServerBuilder[IO]
      .withHttpApp(httpApp)
      .serve
      .compile
      .drain
      .as(ExitCode.Success)

}```

My understanding is that this should match all routes, but only `/` and `/other` work in this example. Every other route returns a 404 error with body `Not Found` (probably from `.orNotFound`). This worked in 0.19 but stopped working in 0.20.

I also think the API is not very nice here as I would expect to be able to use `Root /: path` instead of `"" /: path`, but that's not a priority of course.

**Comment by rossabaker** [comment]
`/:` is right associative.  It behaves much like `::`.  A `Root` wouldn't make any sense here.

`/` is left associative.  Think `/` and `Root` as being like `::` and `Nil`, but in the opposite direction.

If your routes are `path` instead of `"" /: path`, I think they'll do what you want:

```scala
scala> main.httpApp(Request(uri = Uri.uri("/other/foo"))).flatMap(_.as[String]).unsafeRunSync()
res42: String = yes: /foo

scala> main.httpApp(Request(uri = Uri.uri("/foo"))).flatMap(_.as[String]).unsafeRunSync()
res43: String = yes: /foo
```

I would expect `"" /: path` to match something leading with a double slash, where the empty string represents an empty first segment.  It actually matches a triple slash.  I'm going to leave this open while we contemplate that more deeply.  But I am pretty certain that is not what you want unless you're trying to match only paths like `/other//foo` rather than `/other/foo`.

[...]

PR #2322, opened 2018-12-21
Make double slashes more reasonable in DSL

rossabaker:
Leading slashes in the DSL `Path` do not create an empty segment, because we can't distinguish `www.example.com` from `www.example.com/`.  In other words, `Path("/") == Path("") == Root`.

Currently, we drop the leading slash not once, not repeatedly, but twice:
* `Path("//a") == Root / "a"`
* `Path("///a") == Root / "" / "a"`

This change consistently drops the leading slash once:
* `Path("//a") == Root / "" / "a"`
* `Path("///a") == Root / "" / "" / "a"`

Adds a scaladoc, with doctests, and a property test about consistency between `apply(String)` and `apply(List[String])`.

This is inspired by the discussion on #2273.

### DIFF
--- a/dsl/src/main/scala/org/http4s/dsl/impl/Path.scala
+++ b/dsl/src/main/scala/org/http4s/dsl/impl/Path.scala
@@ -24,15 +24,45 @@ trait Path {
 }
 
 object Path {
+  /** Constructs a path from a single string by splitting on the `'/'`
+    * character.
+    *   
+    * Leading slashes do not create an empty path segment.  This is to
+    * reflect that there is no distinction between a request to
+    * `http://www.example.com` from `http://www.example.com/`.
+    * 
+    * Trailing slashes result in a path with an empty final segment,
+    * unless the path is `"/"`, which is `Root`.
+    * 
+    * Segments are URL decoded.
+    * 
+    * {{{
+    * scala> Path("").toList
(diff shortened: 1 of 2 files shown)


dsl/src/main/scala/org/http4s/dsl/impl/Path.scala
```diff
@@ -24,15 +24,43 @@ trait Path {
...
     if (str == "" || str == "/")
       Root
-    else if (!str.startsWith("/"))
-      Path("/" + str)
     else {
-      val slash = str.lastIndexOf('/')
-      val prefix = Path(str.substring(0, slash))
-      prefix / UrlCodingUtils.urlDecode(str.substring(slash + 1))
+      val segments = str.split("/", -1)
+      val segments0 = if (segments.head == "") segments.drop(1) else segments
```
ngbinh: will `segments.head` throw on empty `str`?

rossabaker: I think it's safe, even though the types didn't tell us that.  Line 58 guards against the empty string.  Even without that, `str.split(s, -1)` will return a non-empty Array. 

```scala
scala> "".split("/", -1).head
res23: String = ""
```

I'll add a code comment.

PR #2325, opened 2018-12-25
Use looser constraint in HttpRoutes/HttpApp

kubukoz:


rossabaker:
Thanks for this.  While `Sync.defer` is an alias for `Sync.suspend`, I don't think we should weaken this constraint.  The laws of `Defer` speak to lazy evaluation, but not side effects.  For example, we lose laws about errors.  This compiles and works, but I don't think it's conceptually equivalent.

kubukoz:
hmm, maybe we should a test that checks this? Can you give me more details of how Sync helps there?

core/src/main/scala/org/http4s/HttpRoutes.scala
```diff
@@ -59,9 +58,9 @@ object HttpRoutes {
     * @return An [[HttpRoutes]] that returns some [[Response]] in an `OptionT[F, ?]`
     * wherever `pf` is defined, an `OptionT.none` wherever it is not
     */
-  def of[F[_]](pf: PartialFunction[Request[F], F[Response[F]]])(
-      implicit F: Sync[F]): HttpRoutes[F] =
-    Kleisli(req => OptionT(F.suspend(pf.lift(req).sequence)))
+  def of[F[_]: Applicative](pf: PartialFunction[Request[F], F[Response[F]]])(
```
rossabaker: I usually don't mix context bounds and implicit parameters.  Once I name one, I name them all.  I don't know how others feel.

SystemFw: I don't mind mixing, but obviously ok with whatever

SystemFw:
We change semantics by assuming weaker laws. Your own example above is what I mean.
Now, I actually don't think it's that big of a deal because you can only observe the change if you were throwing outside of `F` before. My main argument is mostly that `Defer` is not well known, and the differences with  `Sync` are too subtle, and we will force people to care when they will see `Defer` in our signatures. But that being said, none of these arguments is that strong, honestly, so I'm happy to change my mind/be overridden, it's not going to be a big deal either way :)

rossabaker:
@kubukoz Is there code that the current constraint is preventing you from writing?

kubukoz:
Not really, my main reason for lowering this constraint was to expose a smaller set of capabilities in a Tagless Final program. For example, a controller that doesn't do any IO operations itself (but depends on some algebras to do them) wouldn't need `Sync` in its context bound. It's not a big deal though, and I don't have a strong opinion on relying on `Defer` instead :)

SystemFw:
>  For example, a controller that doesn't do any IO operations itself (but depends on some algebras to do them) wouldn't need Sync in its context bound

Actually I think this is a good argument in principle, you don't want the ability to suspend arbitrary side-effects. Otoh, the only thing preventing you from doing the same with `Defer` is discipline, at which point you might use `Sync` with the same discipline...

kubukoz:
Another thought: If we only need Sync here to have suspension and catching throwables, maybe `catchNonFatal` + `flatten` will be enough?

kaishh:
@kubukoz Said controller would still need a Defer constraint, which isn't a big step up from Sync. IMHO the correct way about it is introduce a new algebra that just wraps HttpApp.apply method, with a single Sync~>HttpAppable instance. That way the controller can only call a single method without having access to either Sync or Defer.

kubukoz:
I'll close this for now, because no matter what we decide it'll break BC with 0.20, but I'll think about what can be done for 0.20 (the main goal being not using Sync that much)

Issue #2218, opened 2018-10-30
Add request attributes for SSL requests


sebastianvoss:
When using mutual TLS authentication it would be useful to have access to a few SSL-specific request attributes (e.g. to use the clients certificate common name (CN) to do authorization):

- `javax.servlet.request.ssl_session_id`
- `javax.servlet.request.cipher_suite`
- `javax.servlet.request.key_size`
- `javax.servlet.request.X509Certificate`

See [jetty docs](https://github.com/eclipse/jetty.project/blob/6b335877d9cc8a6be0c61c5d6d2a559ab9035323/jetty-server/src/main/java/org/eclipse/jetty/server/SecureRequestCustomizer.java#L210-L213) for more details.

balatbn:
We are planning to do mTLS and we are facing the same issue: Getting client certificate(X509) from the request. Is there any other way we can get this info?

rossabaker:
We can add these to the servlet backend fairly easily.  @sebastianvoss was working on adding to blaze, but it is non-trivial.  There is no solution right now, but I'd like to support this.

sebastianvoss:
Yes, I got distracted by other stuff (it is more effort than we initially thought). I will hopefully have some time to get back to it soon. If someone else wants to jump in I'm happy to help.

aeffrig:
The PR allows for http4s to be usable for those types of applications that require client certificate information by running in a compliant servlet container. If someone wants to comment more on working with SSL session information in blaze, I might be able to help with that as well.

rossabaker:
@aeffrig The blaze branch in progress is at https://github.com/sebastianvoss/http4s/tree/add_ssl_request_attributes.  I can't remember precisely where that stalled, but @sebastianvoss may recall.

@balatbn Does this Undertow setting map map cleanly to `needClientAuth` and `wantClientAuth` on the `SSLEngine`?  An enum instead of a boolean seems appropriate.

balatbn:
Undertow seems to map correctly to `need` and `want` in `SSLEngine`.

https://github.com/undertow-io/undertow/blob/master/core/src/main/java/io/undertow/protocols/ssl/UndertowAcceptingSslChannel.java#L173

rossabaker:
Okay.  In that case, I think we should change that boolean to a sum type, and map it accordingly.

aeffrig:
[...]

PR #2328, opened 2018-12-28
Add SSL servlet attributes to attribute map

philotimos:
This addresses the first part of #2218, namely the copying of SSL servlet request attributes to the http4s request attribute map.  This allows for http4s to be usable for those types of applications that require client certificate information by running in a compliant servlet container.   If someone wants to comment more on the overall implementation of SSL session information in blaze, I might be able to help with that as well.


core/src/main/scala/org/http4s/SecureSession.scala
```diff
@@ -0,0 +1,9 @@
+package org.http4s
+
+import java.security.cert.X509Certificate
+
+final case class SecureSession(
+    sslSessionId: String,
+    cipherSuite: String,
+    keySize: Int,
+    X509Certificate: Array[X509Certificate])
```
rossabaker: Do we want to expose this as an `Array` or translate it into something immutable?  I'm worried about a case class that has a mutable element, but also don't want us to spend time eagerly converting a collection that will often be ignored.

core/src/main/scala/org/http4s/Message.scala
```diff
@@ -447,6 +447,7 @@ object Request {
     val PathTranslated: AttributeKey[File] = AttributeKey[File]
     val ConnectionInfo: AttributeKey[Connection] = AttributeKey[Connection]
     val ServerSoftware: AttributeKey[ServerSoftware] = AttributeKey[ServerSoftware]
+    val SecureSession: AttributeKey[Option[SecureSession]] = AttributeKey[Option[SecureSession]]
```
rossabaker: What does it mean if the key is present, but `None`?  That it's supported by the backend, but that the session wasn't secure?

core/src/main/scala/org/http4s/SecureSession.scala
```diff
@@ -0,0 +1,9 @@
+package org.http4s
```
rossabaker: Does this attribute ever make sense on the client, or should we consider putting this in the server package?

core/src/main/scala/org/http4s/Message.scala
aeffrig: Yes.  It wasn't clear on the preferred convention to express that secure session values are optional.  For guidance, I saw `ServletRequestKeys.HttpSession(Option(req.getSession(false)))` in Http4sServlet.scala, so I followed that lead and went with a required key and optional value.  Also the name `SecureSession` here is meant to correspond the 'secure' in the `isSecure` method of the request.

core/src/main/scala/org/http4s/SecureSession.scala
aeffrig: Here I was just following the convention for servlets since it defines these parameters that way, but yeah, it would make sense to be immutable.  Is there another lazier convention that is preferred to express these parameters or their optionality?  Or should the array just become an immutable list in this existing case class?

core/src/main/scala/org/http4s/SecureSession.scala
aeffrig: It probably makes sense to move it the server package unless anyone has other ideas.

servlet/src/main/scala/org/http4s/servlet/Http4sServlet.scala
```diff
@@ -90,7 +91,17 @@ abstract class Http4sServlet[F[_]](service: HttpRoutes[F], servletIo: ServletIo[
...
           Request.Keys.ServerSoftware(serverSoftware),
-          ServletRequestKeys.HttpSession(Option(req.getSession(false)))
+          ServletRequestKeys.HttpSession(Option(req.getSession(false))),
+          Request.Keys.SecureSession(
+            (
+              Option(req.getAttribute("javax.servlet.request.ssl_session_id").asInstanceOf[String]),
+              Option(req.getAttribute("javax.servlet.request.cipher_suite").asInstanceOf[String]),
+              Option(req.getAttribute("javax.servlet.request.key_size").asInstanceOf[Int]),
+              Option(
+                req
```
balatbn: We are getting `javax.servlet.request.X509Certificate` and converting it to an array of `java.security.cert.X509Certificate`

core/src/main/scala/org/http4s/SecureSession.scala
rossabaker: This is a case that would be nice to have an immutable array wrapper in the standard library or cats.  Just something that doesn't eagerly traverse this collection unless someone actually cares to read it.
The only standard thing I can think of that's lazy and would hide the immutability is a stdlib Stream.  Something like `Stream.range(0, array.length).map(i => array(i))`.  Whether that's actually even lighter than creating a list of just a few certificates, I don't know.

I guess be fine converting it to a `List` and not worrying about it, unless someone wants to argue that every microsecond is sacred here.

core/src/main/scala/org/http4s/Message.scala
rossabaker: Okay, yes, I agree.

core/src/main/scala/org/http4s/SecureSession.scala
rossabaker: :+1:

core/src/main/scala/org/http4s/SecureSession.scala
aeffrig: I was thinking ahead about the future Blaze integration of the code at https://github.com/sebastianvoss/http4s/tree/add_ssl_request_attributes, and I noticed a slight discrepancy how servlets return the SSL session id and how that branch returned that same information.  The servlet defines the SSL sessionId as a `String`, but the blaze branch as an `Array[Byte]`, owing to the fact that blaze retrieves it directly from the engine session as follows:
```
sessionId = engine.getSession.getId
```
So we could pick a representation and convert to it or have separate representations for blaze and for servlets, which would imply client code changes if rehosting between servlets and blaze and could mean putting a representation of `SecureSession` in both the servlet project and the blaze server projects.  I am not certain if the String servlet SSL session id is the same as the SSL session id in a different form as I have not researched the details of that conversion at the moment.  As a result it might make sense to keep the SSL Session id as a String if we want to unify the models since it's more general.

aeffrig:
Merged in PR #2341 from @balatbn

core/src/main/scala/org/http4s/SecureSession.scala
rossabaker: I think it should be the same representation.

I think the `Array[Byte]` might be the sensible representation.  The `SSLSession` seems to be the most server agnostic representation to me, and it's the servlet that's doing something weird.  Unfortunately, that brings us back to a mutable type on the case class, which we could get around with `scodec.bits.ByteVector`.  But that's Yet Another Wrapper.

If you or @balatbn have a strong opinion, I'll defer to that.  My preference is more for unification than for any specific option.

core/src/main/scala/org/http4s/SecureSession.scala
rossabaker: Like what @aeffrig did.

core/src/main/scala/org/http4s/util/SSLContextFactory.scala
```diff
@@ -0,0 +1,67 @@
...
+import java.security.cert.{CertificateFactory, X509Certificate}
+
+import javax.net.ssl.SSLSession
+
+import scala.util.Try
+
+/**
+  * Based on SSLContextFactory from jetty.
+  */
+object SSLContextFactory {
```
rossabaker: I wonder about making this internal API.  As we think about 1.0, everything we add becomes more permanent.  I think `util` is generally a sign that it belongs in another library or that it belongs in the internal package as `protected[http4s]` so we aren't on the hook for supporting it.

core/src/main/scala/org/http4s/util/SSLContextFactory.scala
```diff
@@ -0,0 +1,67 @@
...
+    *     RC4_40     * Stream    40
+    *     RC4_128      Stream   128
+    *     DES40_CBC  * Block     40
+    *     DES_CBC      Block     56
+    *     3DES_EDE_CBC Block    168
+    * </pre>
+    *
+    * @param cipherSuite String name of the TLS cipher suite.
+    * @return int indicating the effective key entropy bit-length.
+    */
```
rossabaker: Nicely documented!

server/src/main/scala/org/http4s/server/package.scala
```diff
@@ -31,6 +31,10 @@ package object server {
     val SocketAddress = InetSocketAddress.createUnresolved(Host, HttpPort)
   }
 
+  object ServerRequestKeys {
+    val SecureSession: AttributeKey[Option[SecureSession]] = AttributeKey[Option[SecureSession]]
+  }
+
```
rossabaker: This is going to need to be reconciled with #2318, which is ready to go once we come to a conclusion on that Vault dependency.  Someone is going to have a light merge to do here.

core/src/main/scala/org/http4s/SecureSession.scala
balatbn: Array[Byte] is fine. For Blaze, we can directly use the value. However, for servlet, we have to convert the hex string back to Array[Byte].

core/src/main/scala/org/http4s/util/SSLContextFactory.scala
balatbn: Copied from Jetty!

core/src/main/scala/org/http4s/util/SSLContextFactory.scala
balatbn: It was also part of "core" as I didn't find any "util" package in server repo. We can move this to blaze-server repo if this doesn't have to be in "util".

rossabaker:
Test failure is unrelated.

Issue #2334, opened 2018-12-31
Tutorial references old code


rohdef:
Hey,

reading through the tutorial I stumbled over code that no longer runs. Also found the solution at https://github.com/http4s/http4s/issues/1501

It's in the dsl-chapter where it references orNotFound in the testing part, which seems that it should be written using ```apply(getRoot).orElseF(NotFound())``` in stead (that is what works for me at least).

Would the best way to fix it be to fork the project and make a pull request on the release-0.18.x branch? Then I could probably easily update dsl.md :) I just don't have much experience with contributing, and would like to do it right

rossabaker:
@rohdef Yes, that would be exactly right.  Anything merged to release-0.18.x eventually gets merged to master, so it would correct it for both versions.  Thanks!

rohdef:
I hope I did it right :) at least there's a suggestion, and I also added a short part about getting the status and the body as a string (last one caused a lot of search and confusion, so hopefully, I'll save others from that)

rohdef:
If I did it right, I will make a bit more changes and create pull requests for those too, since I can see the problem is also present in later parts of the tutorial

Issue #1501, opened 2017-10-27
0.18 Migration Guide


m4dc4p:
I migrated a 0.16 service to 0.18-M4 recently and took notes on my experience. Here is what I found - please feel free to comment, edit and place this somewhere more appropriate.

-----------------------------

This document details changes necessary to migrate services from
http4s 0.16 (and earlier) versions to 0.18. Between those versions,
http4s dropped support for scalaz and scalaz.concurrent and migrated
to cats/cats-effect.

This isn't exhaustive but details the changes I needed to make and
challenges I faced.

Note that this upgrade requires Scala 2.12 - it is not 2.11 compatible.

build.sbt
=====

Remove any scalaz dependencies.

Add the "-Ypartial-unification" flag to scalac options:

```
scalacOptions ++= Seq(
  "-Ypartial-unification",
  ...
)
```

Scalaz imports
=====

In general, `scalaz.concurrent.Task` is always replaced with
`cats.effect._`. Other `scalaz` imports should be replaced with the
following block (I never found the need to do any finer-grained imports):

```
import cats._
import cats.data._
import cats.implicits._
```

`Task`
====

Replace all instances of `Task` with `IO`. `Task.now`/`Task.delay` becomes
`IO.pure` (if the argument is already an `IO` type use `IO.suspend`). 

`Task.fail` becomes `IO.raiseError`. 

Some `Task` instance methods replacements:

* `x.unsafePerformSync` - `x.unsafeRunSync`.
* `x.unsafePerformSyncAttempt` - `Either.catchNonFatal(x.unsafeRunSync)`.

Finally, `gatherUnordered` isn't implemented directly by `IO` and I 
didn't think it added much value, so I just replaced it with a direct
for-comprehension. For example:

```
for {
  _ <- Task.gatherUnordered(taskA, taskB)
  ...
}
```

became:

```
for {
  _ <- taskA
  _ <- taskB 
  ...
}
```

`timed` also went away (in fact, it never behaved the way I thought).

For error handling, replace `handleWith` with `handleErrorWith`.

Scalaz Types & Methods
====

With scalaz gone, the `\/` (disjunction) type and its constructors no
longer apply.

Migrate `\/` to `scala.util.Either`; `\/-` becomes `Right` and `-\/`
becomes `Left`. `\/.fromTryCatchNonFatal` becomes `Either.catchNonFatal`.

The `EitherT` type exists in cats, but the methods have different names:

[...]

PR #2335, opened 2019-01-01
Tutorial updated to newer APIs and the testing examples have been expanded a bit #2334

rohdef:
https://github.com/http4s/http4s/issues/2334 and https://github.com/http4s/http4s/issues/1501

### DIFF
--- a/docs/src/main/tut/dsl.md
+++ b/docs/src/main/tut/dsl.md
@@ -68,7 +68,7 @@ and experiment directly in the REPL.
 ```tut
 val getRoot = Request[IO](Method.GET, uri("/"))
 
-val io = service.orNotFound.run(getRoot)
+val io = service.apply(getRoot).getOrElseF(NotFound())
 ```
 
 Where is our `Response[F]`?  It hasn't been created yet.  We wrapped it
@@ -86,7 +86,17 @@ But here in the REPL, it's up to us to run it:
 val response = io.unsafeRunSync
 ```
 
-Cool.
+Cool.  Now we will be able to test things like the status:
+
+```tut
+val status = response.status
+```
+
+Or the content of our body:
+
+```tut
+val body = response.bodyAsText.compile.toList.unsafeRunSync().mkString("")
+```
 
 ## Generating responses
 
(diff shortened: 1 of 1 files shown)


docs/src/main/tut/dsl.md
```diff
@@ -86,7 +86,17 @@ But here in the REPL, it's up to us to run it:
...
+Cool.  Now we will be able to test things like the status:
+
+```tut
+val status = response.status
+```
+
+Or the content of our body:
+
+```tut
+val body = response.bodyAsText.compile.toList.unsafeRunSync().mkString("")
```
rossabaker: How about just `response.as[String].unsafeRunSync()`?

docs/src/main/tut/dsl.md
```diff
@@ -68,7 +68,7 @@ and experiment directly in the REPL.
 ```tut
 val getRoot = Request[IO](Method.GET, uri("/"))
 
-val io = service.orNotFound.run(getRoot)
+val io = service.apply(getRoot).getOrElseF(NotFound())
```
rossabaker: I think this was explained to me elsewhere, but I can't find the reference now.  What does the existing code do that's wrong?  I would think these are equivalent.

rohdef: At least in my case it seems to type check but once I run it, it cannot find orNotFound :/

docs/src/main/tut/dsl.md
rohdef: Will try that out, that looks a lot better imo

rohdef: And now that I try it again, I naturally cannot reproduce... strange, wonder if I did a typo or if I hit some kind of corner case :S

docs/src/main/tut/dsl.md
rohdef: And now that I try it again, I naturally cannot reproduce... strange, wonder if I did a typo or if I hit some kind of corner case :S

rossabaker: There needs to be an import.  In master, it's `org.http4s.implicits._`.  You might have been missing the import, or I think it also requires -Ypartial-unification.

armanbilge:
Both issues are closed, and 0.18.x is history.

PR #2336, opened 2019-01-02
Docs: Add a simple example of building server using Resource

zakpatterson:
Motivated by this example, where a user tried to lift
a server created in IO via `.stream` into a `Resource`,
then was left with a process that appeared to work
but could not exit

https://gitter.im/http4s/http4s?at=5c2550d7d945b968820580ef


zakpatterson:
Awesome thanks both!

PR #2337, opened 2019-01-02
Docs: Use :silent on import statements

zakpatterson:
Note: I just applied this change to a single file. Feel free to close this if it's not desired, or let me know if this seems better and I'll happily apply the same change to all the docs.

:silent is desirable to avoid showing the user repl responses
about each individual import. Usually we want to show the
user the types inferred from repl lines, but for import
statements there's no value. The proof the import was
found is in the docs compiling at all.

The tradeoff here is that when the markdown is rendered, there will be a break between blocks that otherwise would be in the same file

### DIFF
--- a/docs/src/main/tut/auth.md
+++ b/docs/src/main/tut/auth.md
@@ -11,7 +11,7 @@ For this section, remember that, like mentioned in the [service] section, a serv
 
 Lets start by defining all the imports we will need in the examples below:
 
-```tut:book:silent
+```tut:silent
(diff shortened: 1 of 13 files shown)


zakpatterson:
Great I'll proceed and apply this to the other files soon.

zakpatterson:
Ok, I fixed up all the import blocks, using `tut:silent` instead of `tut:book`. For consistency, I also removed all instances of `tut:book:silent` in favor of `tut:silent`, because without output, `:book` does nothing.

PR #2351, opened 2019-01-11
Change clientAuthMode to an enum

balatbn:
With this mTLS can be optional for the request.

Changes as per discussion on #2218.

Note that this is a breaking change. Existing boolean value is changed to an enum.

### DIFF
--- a/jetty/src/main/scala/org/http4s/server/jetty/JettyBuilder.scala
+++ b/jetty/src/main/scala/org/http4s/server/jetty/JettyBuilder.scala
@@ -68,14 +68,14 @@ sealed class JettyBuilder[F[_]] private (
       keyManagerPassword: String,
       protocol: String = "TLS",
       trustStore: Option[StoreInfo] = None,
-      clientAuth: SSLClientAuthMode.Value = SSLClientAuthMode.NotRequested
+      clientAuth: SSLClientAuthMode = SSLClientAuthMode.NotRequested
   ): Self =
     copy(
       sslBits = Some(KeyStoreBits(keyStore, keyManagerPassword, protocol, trustStore, clientAuth)))
 
   def withSSLContext(
       sslContext: SSLContext,
-      clientAuth: SSLClientAuthMode.Value = SSLClientAuthMode.NotRequested): Self =
+      clientAuth: SSLClientAuthMode = SSLClientAuthMode.NotRequested): Self =
     copy(sslBits = Some(SSLContextBits(sslContext, clientAuth)))
 
   override def bindSocketAddress(socketAddress: InetSocketAddress): Self =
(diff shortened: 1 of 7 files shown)


rossabaker:
Test failure is #2352.

jetty/src/main/scala/org/http4s/server/jetty/JettyBuilder.scala
```diff
@@ -173,6 +175,21 @@ sealed class JettyBuilder[F[_]] private (
...
+  private def updateClientAuth(
+      sslContextFactory: SslContextFactory,
+      clientAuthMode: SSLClientAuthMode.Value): Unit =
+    clientAuthMode match {
+      case SSLClientAuthMode.NotRequested =>
+        sslContextFactory.setWantClientAuth(false)
+        sslContextFactory.setNeedClientAuth(false)
+
+      case SSLClientAuthMode.Requested =>
+        sslContextFactory.setWantClientAuth(false)
```
rossabaker: `true`, right?

I'm inferring from [the javadoc](https://docs.oracle.com/javase/8/docs/api/javax/net/ssl/SSLEngine.html#setNeedClientAuth-boolean-) that we don't need to set both booleans.

jetty/src/main/scala/org/http4s/server/jetty/JettyBuilder.scala
```diff
@@ -173,6 +175,21 @@ sealed class JettyBuilder[F[_]] private (
...
+
+      case SSLClientAuthMode.Required =>
+        sslContextFactory.setNeedClientAuth(false)
```
rossabaker: Also `true`?

server/src/main/scala/org/http4s/server/SSLClientAuthMode.scala
```diff
@@ -0,0 +1,8 @@
+package org.http4s.server
+
+/**
+  * Client Auth mode for mTLS
+  */
+object SSLClientAuthMode extends Enumeration {
```
rossabaker: Is there some Java interop reason to use an `Enumeration` instead of a sealed trait here?  I imagine:

```
sealed abstract class SSLClientAuthMode extends Product with Serializable
object SSLClientAuthMode {
  case object NotRequested extends SSLClientAuthMode
  case object Requested extends SSLClientAuthMode
  case object Required extends SSLClientAuthMode
}
```

I vaguely remember something about Scala enums and Java interop, but not whether it's good or bad.  But I don't think this is going to be called from Java anyway.

ChristopherDavenport: I generally used `sealed trait` but otherwise the same

jetty/src/main/scala/org/http4s/server/jetty/JettyBuilder.scala
balatbn: Sorry. Will fix both.

server/src/main/scala/org/http4s/server/SSLClientAuthMode.scala
balatbn: Will convert to a trait

balatbn:
In my local machine, I get a "Connection reset" error message if mTLS fails. In CI box, it is giving a "Remote host closed connection during handshake" message.

Issue #2011, opened 2018-08-19
Identify headers with a default model, and then model them.


rossabaker:
Headers that extend `HeaderKey.Default` should have a case class that models them as something more than a String.  This task involves:

- Enumerating such headers, with link to relevant RFC.
- For each one, interpreting the RFC into a case class and writing the parser.

This, like the recent middleware reform, will be a source of good first issue tickets.

rossabaker:
Here is a list of headers we have declared a key for, but not modeled:

```scala
object `Accept-Patch` extends HeaderKey.Default
object `Access-Control-Allow-Credentials` extends HeaderKey.Default
object `Access-Control-Allow-Headers` extends HeaderKey.Default
object `Access-Control-Allow-Methods` extends HeaderKey.Default
object `Access-Control-Allow-Origin` extends HeaderKey.Default
object `Access-Control-Expose-Headers` extends HeaderKey.Default
object `Access-Control-Max-Age` extends HeaderKey.Default
object `Access-Control-Request-Headers` extends HeaderKey.Default
object `Access-Control-Request-Method` extends HeaderKey.Default
object `Content-Base` extends HeaderKey.Default
object `Content-Language` extends HeaderKey.Default
object `Content-Location` extends HeaderKey.Default
object `Content-MD5` extends HeaderKey.Default
object `Content-Security-Policy` extends HeaderKey.Default
object `Content-Transfer-Encoding` extends HeaderKey.Default
object Expect extends HeaderKey.Default
object From extends HeaderKey.Default
object `Front-End-Https` extends HeaderKey.Default
object `If-Match` extends HeaderKey.Default
object `If-Range` extends HeaderKey.Default
object `If-Unmodified-Since` extends HeaderKey.Default
object `Max-Forwards` extends HeaderKey.Default
object Origin extends HeaderKey.Default
object Pragma extends HeaderKey.Default
object `Proxy-Authentication` extends HeaderKey.Default
object `Proxy-Authorization` extends HeaderKey.Default
object `Sec-WebSocket-Accept` extends HeaderKey.Default
object `Sec-WebSocket-Key` extends HeaderKey.Default
object `Sec-WebSocket-Key1` extends HeaderKey.Default
object `Sec-WebSocket-Key2` extends HeaderKey.Default
object `Sec-WebSocket-Location` extends HeaderKey.Default
object `Sec-WebSocket-Origin` extends HeaderKey.Default
[...]

PR #2355, opened 2019-01-13
Add HttpMethodOverrider middleware

nebtrx:
This one adds an Http Method Overrider middleware(to contribute with #2014) in order to allow using  HTTP verbs such as **PUT** or **DELETE** in places where the client doesn't support it. 

It takes some ideas from [django-method-override](https://pypi.org/project/django-method-override/) and from [ExpressJS method override](https://github.com/expressjs/method-override).

Let me know what you guys think of it and if it requires some changes, specially in the test description since I'm new with the predominant testing style.

PS: Also, I just realized of the existence of the `headers` package and its particular way of representing/encoding **Headers**. 
Do you think that should be necessary for **Vary** taking into consideration how I'm using it? 

Cheers!! 😴

### DIFF
--- a/server/src/main/scala/org/http4s/server/middleware/HttpMethodOverrider.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/HttpMethodOverrider.scala
@@ -1,21 +1,13 @@
-package org.http4s.server.middleware
+package org.http4s
+package server
+package middleware
 
 import cats.data.Kleisli
 import cats.effect.Sync
 import cats.implicits._
 import cats.{Monad, ~>}
+import org.http4s.Http
 import org.http4s.util.CaseInsensitiveString
-import org.http4s.{
-  AttributeKey,
-  Header,
-  Http,
-  Method,
-  ParseResult,
-  Request,
-  Response,
-  Status,
-  UrlForm
-}
 
 import scala.reflect.runtime.universe._
 
--- a/server/src/main/scala/org/http4s/server/middleware/HttpMethodOverrider.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/HttpMethodOverrider.scala
@@ -1,10 +1,23 @@
 package org.http4s.server.middleware
 
-import cats.Applicative
 import cats.data.Kleisli
+import cats.effect.Sync
 import cats.implicits._
+import cats.{Monad, ~>}
 import org.http4s.util.CaseInsensitiveString
-import org.http4s.{AttributeKey, Header, Http, Method, ParseResult, Request, Response, Status}
+import org.http4s.{
+  AttributeKey,
+  Header,
+  Http,
+  Method,
+  ParseResult,
+  Request,
+  Response,
+  Status,
+  UrlForm
+}
+
+import scala.reflect.runtime.universe._
 
 object HttpMethodOverrider {
 
@@ -41,8 +54,10 @@ object HttpMethodOverrider {
   final case class HeaderOverrideStrategy(headerName: CaseInsensitiveString)
       extends OverrideStrategy
   final case class QueryOverrideStrategy(paramName: String) extends OverrideStrategy
-  // TODO: tory to fit this inside the main logic
-  //  final case class FunctionOverrideStrategy[G[_]](fn: Request[G] => Method) extends OverrideStrategy
+  final case class FormOverrideStrategy[G[_], F[_]](
+      fieldName: String,
+      naturalTransformation: G ~> F)
+      extends OverrideStrategy
 
   val defaultConfig = HttpMethodOverriderConfig(
     HeaderOverrideStrategy(CaseInsensitiveString("X-HTTP-Method-Override")),
@@ -57,11 +72,19 @@ object HttpMethodOverrider {
     * the desired one using a custom header or request parameter. The middleware will '''override'''
     * the original verb with the new one for you, allowing the request the be dispatched properly.
     *
-    * @param http [[HttpApp]] to transform
+    * @param http [[Http]] to transform
     * @param config http method overrider config
     */
   def apply[F[_], G[_]](http: Http[F, G], config: HttpMethodOverriderConfig)(
-      implicit F: Applicative[F]): Http[F, G] = {
+      implicit F: Monad[F],
+      S: Sync[G],
+      TT: TypeTag[G ~> F]): Http[F, G] = {
+
+    lazy val runtimeTypeNT = implicitly[TypeTag[G ~> F]].tpe
+
+    val parseMethod = (m: String) => Method.fromString(m.toUpperCase)
+
+    val processRequestWithOriginalMethod = (req: Request[G]) => http(req)
 
     def processRequestWithMethod(
         req: Request[G],
@@ -81,7 +104,7 @@ object HttpMethodOverrider {
               .getOrElse(Header(varyHeaderName.value, headerName.value))
 
           resp.withHeaders(resp.headers.put(updatedVaryHeader))
-        case QueryOverrideStrategy(_) => resp
+        case _ => resp
       }
     }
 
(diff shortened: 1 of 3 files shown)


server/src/main/scala/org/http4s/server/middleware/HttpMethodOverrider.scala
```diff
@@ -0,0 +1,87 @@
...
+    * @param http [[HttpApp]] to transform
+    * @param config http method overrider config
+    */
+  def apply[F[_]](http: HttpApp[F], config: HttpMethodOverriderConfig)(
+      implicit F: Applicative[F]): HttpApp[F] = {
+
+    def processRequestWithMethod(
+        req: Request[F],
+        parseResult: ParseResult[Method]): F[Response[F]] = parseResult match {
+      case Left(_) => F.pure(Response[F](Status.MethodNotAllowed))
```
rossabaker: I think this would be a `BadRequest`.  `MethodNotAllowed` means we recognize the method.

server/src/main/scala/org/http4s/server/middleware/HttpMethodOverrider.scala
```diff
@@ -0,0 +1,87 @@
...
+      case Right(om) => http(updateRequestWithMethod(req, om)).map(updateVaryHeader)
+    }
+
+    def updateVaryHeader(resp: Response[F]): Response[F] = {
```
rossabaker: Good call on this.  I wouldn't have thought of it.

server/src/main/scala/org/http4s/server/middleware/HttpMethodOverrider.scala
```diff
@@ -0,0 +1,87 @@
...
+        case QueryOverrideStrategy(_) => resp
+      }
+    }
+
+    def updateRequestWithMethod(req: Request[F], om: Method): Request[F] = {
+      val attrs = req.attributes ++ Seq(overriddenMethodAttrKey(req.method))
+      req.withAttributes(attrs).withMethod(om)
+    }
+
+    def ignoresOverrideIfNotAllowed(req: Request[F]): Option[Unit] =
```
rossabaker: I have a vague sense something could be done with `Alternative` here and not need this weird `Option[Unit]`, but it's internal, so I'm not going to fuss too much about it.

server/src/main/scala/org/http4s/server/middleware/HttpMethodOverrider.scala
```diff
@@ -0,0 +1,87 @@
...
+      if (config.overridableMethods.contains(req.method)) Some(()) else None
+
+    def getUnsafeOverrideMethod(req: Request[F]): Option[String] =
+      config.overrideStrategy match {
```
rossabaker: Are there other plausible strategies, such that this might be a function on the strategy and unseal the strategy?  People don't embed this in a `UrlForm`, do they?

server/src/main/scala/org/http4s/server/middleware/HttpMethodOverrider.scala
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    def updateVaryHeader(resp: Response[F]): Response[F] = {
+      val varyHeaderName = CaseInsensitiveString("Vary")
+      config.overrideStrategy match {
+        case HeaderOverrideStrategy(headerName) =>
+          val updatedVaryHeader =
+            resp.headers
+              .get(varyHeaderName)
+              .map((h: Header) => Header(h.name.value, s"${h.value}, ${headerName.value}"))
```
rossabaker: Would be nice if we had a modeled `Vary` header which we could add to, but that could be a separate enhancement.

server/src/main/scala/org/http4s/server/middleware/HttpMethodOverrider.scala
```diff
@@ -0,0 +1,87 @@
...
+import cats.implicits._
+import org.http4s.util.CaseInsensitiveString
+import org.http4s.{AttributeKey, Header, HttpApp, Method, ParseResult, Request, Response, Status}
+
+object HttpMethodOverrider {
+
+  /**
+    * HttpMethodOverrider middleware config options.
+    */
+  final case class HttpMethodOverriderConfig(
```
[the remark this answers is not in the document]
nebtrx: That makes, I wasn't aware. Done!

server/src/main/scala/org/http4s/server/middleware/HttpMethodOverrider.scala
```diff
@@ -0,0 +1,87 @@
...
+      overrideStrategy: OverrideStrategy,
+      overridableMethods: List[Method])
```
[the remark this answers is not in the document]
nebtrx: Yup, you're right. It's more idiomatic. I went for sense. I thought it simplifies things just to have a whitelist

server/src/main/scala/org/http4s/server/middleware/HttpMethodOverrider.scala
```diff
@@ -0,0 +1,87 @@
...
+    *
+    * This middleware lets you use  HTTP verbs such as PUT or DELETE in places where the client
+    * doesn't support it. Camouflage your request with another HTTP verb(usually POST) and sneak
+    * the desired one using a custom header or request parameter. The middleware will '''override'''
+    * the original verb with the new one for you, allowing the request the be dispatched properly.
+    *
+    * @param http [[HttpApp]] to transform
+    * @param config http method overrider config
+    */
+  def apply[F[_]](http: HttpApp[F], config: HttpMethodOverriderConfig)(
```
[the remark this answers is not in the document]
nebtrx: Done!

server/src/main/scala/org/http4s/server/middleware/HttpMethodOverrider.scala
nebtrx: Actually, I just was curious about the impact of this in the caching process and from there one thing led to another.

server/src/main/scala/org/http4s/server/middleware/HttpMethodOverrider.scala
nebtrx: We shared the same feeling, regarding the `Option[Unit]`. 

The thing is the code is already moving inside the `Option` context so.. it semmed natural to remain on it. 

I went for `config.overridableMethods.contains(req.method).guard[Option].as(())` to slighly improve it BTW

server/src/main/scala/org/http4s/server/middleware/HttpMethodOverrider.scala
nebtrx: Agreed, perhaps you can open an issue explaining more details. I'll be glad to help in that one.

server/src/main/scala/org/http4s/server/middleware/HttpMethodOverrider.scala
nebtrx: Ohh, I wasn't aware of `UrlForm`. I have to take a look at it since I was only thinking about `QueryParamDecoder`. The Query Param Strategy might get in the middle. If so, I'd have a to find a way to update the request and remove the param once it has been overridden.

I'm also considering a strategy like what you say. Perhaps something that receives a `Request[G] => Request[G]` or `Request[G] => Method`. The first one seems to lose some control over the request pipeline and the second one...mmm....perhaps it exposes too much information to just override the Request method.

I'll think about this.

nebtrx:
@rossabaker thanks for your feedback! 
I addressed most of the suggestions, except for a couple that requires some time to check a few things

server/src/main/scala/org/http4s/server/middleware/HttpMethodOverrider.scala
rossabaker: I like it like that.

server/src/main/scala/org/http4s/server/middleware/HttpMethodOverrider.scala
rossabaker: It's one of many mentioned in https://github.com/http4s/http4s/issues/2011.

server/src/main/scala/org/http4s/server/middleware/HttpMethodOverrider.scala
```diff
@@ -57,30 +104,44 @@ object HttpMethodOverrider {
...
-      if (config.overridableMethods.contains(req.method)) Some(()) else None
-
-    def getUnsafeOverrideMethod(req: Request[F]): Option[String] =
+    def getUnsafeOverrideMethod(req: Request[G]): F[Option[String]] =
       config.overrideStrategy match {
-        case HeaderOverrideStrategy(headerName) => req.headers.get(headerName).map(_.value)
-        case QueryOverrideStrategy(parameter) => req.params.get(parameter)
+        case HeaderOverrideStrategy(headerName) => F.pure(req.headers.get(headerName).map(_.value))
+        case QueryOverrideStrategy(parameter) => F.pure(req.params.get(parameter))
+        case FormOverrideStrategy(field, f) if runtimeTypeNT == typeOf[G ~> F] =>
```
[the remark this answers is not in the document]
nebtrx: Yeah, it scared me too :S and I hate reflection. I spend ~ two hours looking for the proper/elegant way of overcoming type erasure in this kind of scenarios and there seems to a general consensus about using the following approaches:

1. Using `ClassTag[_]`, `TypeTag[_]` or `WeakTypeTag[_]` 
2. Using `Manifest[_]` 

Both are pretty similar. I had a slightly bigger preference for the first one and since [`Headerkey`](https://github.com/http4s/http4s/blob/a53dbb1747b69cbe67e637536f73d7d369d62896/core/src/main/scala/org/http4s/HeaderKey.scala#L84) is already using that approach with `ClassTag`, I went for it.

Let me know if have a better approach. I keep looking for better ways to handle this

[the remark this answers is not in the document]
nebtrx: BTW, yes, I'd be trading `ClassCastException` for `MatchError`. It kind of feels more logical to me since we trying to match strategies. WDYT?

nebtrx:
Hey @rossabaker 

Sorry for taking too much time to address your concerns, I was off for a few days. I pushed a commit with your package suggestion and I left some comments about the reflection stuff.

server/src/main/scala/org/http4s/server/middleware/HttpMethodOverrider.scala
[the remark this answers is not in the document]
rossabaker: I can't think of a better solution, and I think it would also be difficult to use it incorrectly.  I'm inclined to roll with it.  I'd rather have the feature, and if someone can think of something more clever later, so much the better.

nebtrx:
Cool @rossabaker! Thx for walking me through this.

nebtrx:
Thanks for your review @ChristopherDavenport.

Regarding this:
> I think we'll need to address this in the future, but this gets us started.

Were you referring about introducing blockers to the ScalaJS plan? or something else?

ChristopherDavenport:
Yeah, I have a possible approach for how to improve this, but I don't want to shove this off onto you if you don't want to take it on. 

My first idea is that we can put the type parameters all the way up that coproduct hierarchy such that. `OverrideStrategy[+G[_], +F[_]]` where we can then instantiate them as `Nothing` for the other two. Or if that doesn't work, we can just have to apply the type parameter. However that would work fully in types and not needing reflection.

nebtrx:
> Yeah, I have a possible approach for how to improve this, but I don't want to shove this off onto you if you don't want to take it on.
> 
> My first idea is that we can put the type parameters all the way up that coproduct hierarchy such that. `OverrideStrategy[+G[_], +F[_]]` where we can then instantiate them as `Nothing` for the other two. Or if that doesn't work, we can just have to apply the type parameter. However that would work fully in types and not needing reflection.

@ChristopherDavenport No worries, I can explore that path, especially because that issue was really annoying for me. However, at first glance, I don't see how we can get rid of the type erasure warning when pattern matching on something like `OverrideStrategy[+G[_], +F[_]]`. I'll give it a try tho, I might be missing the elephant in the room.

rossabaker:
Extended in #2407.

PR #2359, opened 2019-01-16
Remove Coverage Steps

ChristopherDavenport:
Couldn't find where the plugin was introduced as a dependency. Removed the calls in CI that should have triggered it.


ChristopherDavenport:
Rig defaults appear to be at play at it continuing to run ever after removing all settings.

ChristopherDavenport:
Finally figured out the sbt setting override.

jgogstad:
I'm looking at introducing test coverage to my team. Out of curiosity, why did you remove it from http4s?

rossabaker:
1. Few people, if any, looked at the data.
2. The data is noisy, particularly with property tests, which introduce non-determinism.
3. If you're not careful about your build, it's easy to release instrumented code.

All that said, if taken as a reference and not as a management edict, I think coverage can still provide value.  Just beware [Goodhart's Law](https://en.wikipedia.org/wiki/Goodhart%27s_law).

PR #2366, opened 2019-01-21
bufferMaxSize parameter of CachingChunkWriter made configurable

### LINKED ISSUE #2363: Content-Length header not sent for requests larger than 10 kB
Due to [this constant](https://github.com/http4s/http4s/blob/master/blaze-core/src/main/scala/org/http4s/blazecore/util/CachingChunkWriter.scala#L17) in private class, when making a request larger than 10 kB, the `Content-Length` header is not sent (because that buffer is flushed and so headers must be also written).

It would be nice to make this constant configurable, and maybe make the default value little bit higher (e.g. 1 MB).

augi:
Also the default value changed from `10 * 1024` to `1024 * 1024`. Fixes #2363


blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientConfig.scala
```diff
@@ -31,6 +31,7 @@ import scala.concurrent.duration._
   * @param maxResponseLineSize maximum length of the request line
   * @param maxHeaderLength maximum length of headers
   * @param maxChunkSize maximum size of chunked content chunks
+  * @param chunkBufferMaxSize Size of the buffer that is used when Content-Length header is not specifiec.
```
rossabaker:
```suggestion
  * @param chunkBufferMaxSize Size of the buffer that is used when Content-Length header is not specified.
```

blaze-core/src/main/scala/org/http4s/blazecore/Http1Stage.scala
```diff
@@ -26,6 +26,8 @@ trait Http1Stage[F[_]] { self: TailStage[ByteBuffer] =>
 
   protected implicit def F: Effect[F]
 
+  protected val chunkBufferMaxSize: Int
```
rossabaker: Eager vals in a base class terrify me, but I think this one is safe because it's not referenced until it starts serving requests.  I don't think there's any harm in making this a def, though.

augi:
Thank you for the comments! I fixed both issues.

augi:
Thank you for pointing me to this test!

In TravisCI, it failed after 5 seconds (timeout) but it finishes successfully in 3 seconds on my machine. So I decided to use smaller `chunkBufferMaxSize` for this test suite.

After this change, this test finishes under one second on my machine so the 5 seconds will be hopefully suficient for TravisCI.

augi:
There were similar issue with blaze-server test so I've fixed it as well.

PR #2369, opened 2019-01-22
Logger purification

### LINKED ISSUE #2368: Logger Middleware Impure Operations
`Logger` takes argument `String => Unit` and then applies this operation in map. This operation is a logging side-effect. 

Rather than applying this in pure position we should instead modify the interface to `String => F[Unit]` such that we correctly log in an effect, and does not force people to run effects to put in place their own referentially transparent logging systems.

ChristopherDavenport:
Fixes #2368

### DIFF
--- a/server/src/main/scala/org/http4s/server/middleware/ResponseLogger.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/ResponseLogger.scala
@@ -20,16 +20,18 @@ object ResponseLogger {
       logHeaders: Boolean,
       logBody: Boolean,
       redactHeadersWhen: CaseInsensitiveString => Boolean = Headers.SensitiveHeaders.contains,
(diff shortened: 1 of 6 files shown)


server/src/main/scala/org/http4s/server/middleware/ResponseLogger.scala
```diff
@@ -45,13 +47,14 @@ object ResponseLogger {
...
                       logBody,
-                      redactHeadersWhen)(logAction)
+                      redactHeadersWhen)(log)
                   }
               )
             }
         }
         .handleErrorWith(t =>
-          F.delay(logger.info(s"service raised an error: ${t.getClass}")) *> F
+          F.delay(log(s"service raised an error: ${t.getClass}")) *> F
```
rossabaker: Getting sleepy but isn't this delaying an `F` and discarding it?

PR #2370, opened 2019-01-22
Prometheus Effect suspension

### LINKED ISSUE #2367: Prometheus - Metrics Ops Mutable State in Pure Position
https://github.com/http4s/http4s/blob/master/prometheus-metrics/src/main/scala/org/http4s/metrics/prometheus/Prometheus.scala#L155-L180

To demonstrate error
```
val cr = new CollectorRegistry()
val p1 = Prometheus[F](cr)
val p2 = Prometheus[F](cr)
```
p2 will throw already registered exception. As a result Prometheus should be created in an effect to display this referentially transparently.

ChristopherDavenport:
Fixes #2367 

Registration is not happening in an effect and can throw if it double registered for the same metric.

### DIFF
--- a/prometheus-metrics/src/main/scala/org/http4s/metrics/prometheus/Prometheus.scala
+++ b/prometheus-metrics/src/main/scala/org/http4s/metrics/prometheus/Prometheus.scala
@@ -62,7 +62,7 @@ object Prometheus {
     * * @param prefix a prefix that will be added to all metrics
     **/
   def apply[F[_]](registry: CollectorRegistry, prefix: String = "org_http4s_server")(
-      implicit F: Sync[F]): F[MetricsOps[F]] = Sync[F].delay{
+      implicit F: Sync[F]): F[MetricsOps[F]] = Sync[F].delay {
     new MetricsOps[F] {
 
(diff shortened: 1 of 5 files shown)


prometheus-metrics/src/main/scala/org/http4s/metrics/prometheus/Prometheus.scala
```diff
@@ -62,7 +62,7 @@ object Prometheus {
     * * @param prefix a prefix that will be added to all metrics
     **/
   def apply[F[_]](registry: CollectorRegistry, prefix: String = "org_http4s_server")(
-      implicit F: Sync[F]): MetricsOps[F] =
+      implicit F: Sync[F]): F[MetricsOps[F]] = Sync[F].delay {
```
rossabaker: Could just call `F.delay` here... it already has a name.

aeons:
```suggestion
      implicit F: Sync[F]): F[MetricsOps[F]] = F.delay {
```

PR #2389, opened 2019-02-06
Log Request Key

kevinmeredith:
Recently I ran into this message:

> No connections available for the desired key. Stats: ...

I wasn't sure which downstream service call was not being serviced, i.e. per the above message.

It seems useful to me to add this value to know this information, i.e. which downstream's call is not being serviced.


rossabaker:
I didn't notice this was against 0.18.  I guess we could do an 0.18.22?

kevinmeredith:
> I guess we could do an 0.18.22?

Yes, please!

PR #2393, opened 2019-02-07
Scala 2.13.0-M5 support

### LINKED ISSUE #2025: scala-2.13.0
1. Publish our internal dependencies.
2. Identify which of our external dependencies are published.
2. Once they all are, publish.

Recommend we begin this exercise with 2.13.0-M5.

Our monorepo is going to mean we're at the tail end of this, which is why we haven't even tried yet.

aeons:
So far only dependencies have been updated.

Closes #2025


SethTisue:
I don't see a 2.13 job in the CI matrix?

aeons:
This is far from building at all. There are several classes that implement collection-like behaviour, and I'm not 100% sure on how to convert those.

Do we add a separate source tree for 2.13, stop extending collections, or something completely different?

SethTisue:
> Do we add a separate source tree for 2.13, stop extending collections, or something completely different?

Very likely you will need some Scala-version-specific sources. (sbt makes this surprisingly easy to arrange.)

It is not usually necessary to copy-and-paste sources wholesale; it's usually possible to get by with just a few aliases and forwarders in version-specific source files.

an example: https://github.com/scala/scala-xml/blob/master/shared/src/main/scala-2.13/scala/xml/ScalaVersionSpecific.scala and https://github.com/scala/scala-xml/blob/master/shared/src/main/scala-2.11-2.12/scala/xml/ScalaVersionSpecific.scala , which are [wired in to the build here](https://github.com/scala/scala-xml/blob/5bb1af026ee09a4cc9b7322d46d8b7e86e88c40e/build.sbt#L29-L37) and [used here](https://github.com/scala/scala-xml/blob/master/shared/src/main/scala/scala/xml/NodeSeq.scala)

another approach is to take on scala-collection-compat as a dependency. this can require less work on your end, but has all the downsides that adding dependencies always has

aeons:
Alright. Thanks for the pointers. I'll take another look, when I have some free time.

rossabaker:
I haven't looked, but someone said fs2 has scala-collection-compat as a dependency already, so we may already have it.

core/src/main/scala/org/http4s/Query.scala
```diff
@@ -70,26 +84,15 @@ final class Query private (pairs: Vector[KeyValue])
     * If multiple values exist for a key, the first is returned. If
     * none exist, the empty `String` "" is returned.
     */
-  def params: Map[String, String] = new ParamsView(multiParams)
+  lazy val params: immutable.Map[String, String] =
+    CollectionCompat.mapValues(multiParams)(_.headOption.getOrElse(""))
```
rossabaker: This is not the most efficient thing I can imagine, but `ParamsView` is not salvageable. Look at that `+`.

rossabaker:
Missing dependencies: boopickle, scalatags, blaze

SethTisue:
> blaze

was blocked by missing spire for a long time, but spire just published a day or two ago. I just opened https://github.com/http4s/blaze/issues/274 on it

> scalatags

see https://github.com/lihaoyi/scalatags/pull/185 , I submitted support, but @lihaoyi had trouble publishing it and hasn't responded to my questions about it, I pinged just now

> boopickle 

I just opened https://github.com/suzaku-io/boopickle/issues/113

SethTisue:
@lihaoyi is asking on https://github.com/lihaoyi/scalatags/pull/185 if someone else can do the publishing

rossabaker:
The compiler is telling me the second of these cases is unreachable and generating a fatal warning.  I've got a few cases like this and I'm not sure what's going on:

```scala
  val testApp = HttpApp[IO] {
    case req @ POST -> Root / "post" =>
      Ok(req.body)
    case GET -> Root / "request" =>
      Ok("request response")
    case _ =>
      NotFound()
  }
```

rossabaker:
I'm just going to remove that warning from 2.13.0-M5 to keep things moving along.

rossabaker:
This is happening only on the new scalatags and I don't know yet if it's us or them:

```
[error] error while loading package, class file '/home/ross/src/http4s/core/target/scala-2.13.0-M5/classes/org/http4s/package.class' is broken
[error] (class java.lang.RuntimeException/error reading Scala signature of package.class: Scala signature package has wrong version
[error]  expected: 5.0
[error]  found: 5.2 in package.class)
[error] one error found
```

rossabaker:
Never mind the above.

rossabaker:
I dropped boopickle for 2.13.  It can come back once it's released.

rossabaker:
All tests are now passing for me locally.

rossabaker:
Regarding `Seq` in public API:
* blaze standardized on `collection.Seq`
* I replaced most of the ones in headers with `List`.  We could argue for `Chain` here.
* parboiled diverges between 2.12 and 2.13, so our parsers do too. That's okay, they're private.

I also had to disable the mimeDbGenerator project for Scala 2.13 for lack of a compatible treehugger.  We don't need it anyway.

I doubt this will pass, but I think I'm done making sweeping changes.  Reviews appreciated so we can get that RC prepared!

rossabaker:
So `++2.13.0-M5 test` excludes the unsupported modules appropriately, but `++2.13.0-M5 ci` doesn't respect the unsupported projects.  People used to do this with sbt-doge, but that's dead.

I think we could make this work by moving code out of the sbt-release steps and into the shell fire, but, :fire:.  Anybody know a better way?

rossabaker:
I think adding `crossScalaVersions` like I did may break our integration with the sbt-release plugin as well.  Would really like some help here.

ChristopherDavenport:
What if we set our own set of keys doing each command with a version explicitly? `publish213`, `publish212`, etc. 

Just an idea, not sure if it will work.

rossabaker:
In place of the sbt-release plugin?  It may be necessary, but it seems like a lot of work.

rossabaker:
The projects that I see doing this do not appear to be publishing from Travis: 
https://github.com/sbt/sbt-release/issues/214

SethTisue:
> The compiler is telling me the second of these cases is unreachable and generating a fatal warning. I've got a few cases like this and I'm not sure what's going on:

@rossabaker that one looks a bit alarming, would you mind opening a scala/bug ticket on it, even if you're not able to minimize it?

rossabaker:
@SethTisue I'll open a stub ticket now and try to minimize it after the conference.

ChristopherDavenport:
It lives!

project/Http4sPlugin.scala
```diff
@@ -177,16 +188,16 @@ object Http4sPlugin extends AutoPlugin {
...
+        releaseStepCommandAndRemaining("+mimaReportBinaryIssues"),
+        // releaseStepCommand("unusedCompileDependenciesTest"),
+        // releaseStepCommand("test:scalafmt::test").when(primary),
+        // releaseStepCommand("docs/makeSite").when(primary),
+        // releaseStepCommand("website/makeSite").when(primary),
         releaseStepCommandAndRemaining("+publishSigned").when(publishable),
         releaseStepCommand("sonatypeReleaseAll").when(publishable && release),
-        releaseStepCommand("docs/ghpagesPushSite").when(publishable && primary),
-        releaseStepCommand("website/ghpagesPushSite").when(publishable && primary && master),
+        // releaseStepCommand("docs/ghpagesPushSite").when(publishable && primary),
```
ChristopherDavenport: Moved to docs release stage

project/Http4sPlugin.scala
```diff
@@ -177,16 +188,16 @@ object Http4sPlugin extends AutoPlugin {
         setReleaseVersion.when(release),
         tagRelease.when(primary && release),
         runClean,
-        runTest,
-        releaseStepCommand("mimaReportBinaryIssues"),
-        releaseStepCommand("unusedCompileDependenciesTest"),
-        releaseStepCommand("test:scalafmt::test").when(primary),
-        releaseStepCommand("docs/makeSite").when(primary),
-        releaseStepCommand("website/makeSite").when(primary),
+        // runTest,
```
ChristopherDavenport: Should we add this and then suppress in the ci step?

project/Http4sPlugin.scala
```diff
@@ -177,16 +188,16 @@ object Http4sPlugin extends AutoPlugin {
...
+        releaseStepCommandAndRemaining("+mimaReportBinaryIssues"),
+        // releaseStepCommand("unusedCompileDependenciesTest"),
```
ChristopherDavenport: Checked in initial build

project/Http4sPlugin.scala
```diff
@@ -177,16 +188,16 @@ object Http4sPlugin extends AutoPlugin {
...
+        // releaseStepCommand("test:scalafmt::test").when(primary),
```
ChristopherDavenport: Checked in primary test build

project/Http4sPlugin.scala
```diff
@@ -177,16 +188,16 @@ object Http4sPlugin extends AutoPlugin {
...
+        // releaseStepCommand("docs/makeSite").when(primary),
```
ChristopherDavenport: Checked in docs test build

rossabaker:
I am desperately behind on my talk, so may be releasing this from the plane, but many thanks for jumping in.  I'll review the parts I didn't do ASAP.

aeons:
Great work, guys!

core/src/main/scala/org/http4s/Query.scala
```diff
@@ -17,28 +17,51 @@ import scala.collection.mutable.ListBuffer
...
   * by '&' while the key is separated from the value with '='
   */
-final class Query private (pairs: Vector[KeyValue])
-    extends IndexedSeq[KeyValue]
-    with IndexedSeqOptimized[KeyValue, Query]
-    with QueryOps
-    with Renderable {
-  override def apply(idx: Int): KeyValue = pairs(idx)
+final class Query private (val pairs: Vector[KeyValue]) extends QueryOps with Renderable {
+  // override def apply(idx: Int): KeyValue = pairs(idx)
```
aeons: Does this need a `@deprecated` or should it just be removed?

aeons:
Apparently I can't approve my own pull request, but I wan to do so anyway :)

:+1:

core/src/main/scala/org/http4s/Query.scala
rossabaker: I see no reason we can't keep this, though it suggests random access in the implementation. I'm ambivalent about deprecating it or not.

core/src/main/scala/org/http4s/QueryOps.scala
```diff
@@ -27,7 +28,7 @@ trait QueryOps {
     _withQueryParam(QueryParam[T].key, QueryParamEncoder[T].encode(value) :: Nil)
 
   /** alias for withQueryParam */
-  def +*?[T: QueryParam: QueryParamEncoder](values: Seq[T]): Self =
+  def +*?[T: QueryParam: QueryParamEncoder](values: List[T]): Self =
```
rossabaker: This is inconsistent with the rest, which take `collection.Seq`

PR #2396, opened 2019-02-08
Make Server Logger Implementation Generic

ChristopherDavenport:


ChristopherDavenport:
Note: This buffers on RequestLogger despite never logging anything for it on anything on it for this Route. I don't have a great way to get around this at present.

server/src/main/scala/org/http4s/server/middleware/Logger.scala
```diff
@@ -15,21 +17,37 @@ import org.log4s.getLogger
...
       Sync[F].delay(logger.info(s))
     }
-    ResponseLogger(logHeaders, logBody, redactHeadersWhen, log.pure[Option])(
-      RequestLogger(logHeaders, logBody, redactHeadersWhen, log.pure[Option])(http)
+    ResponseLogger(logHeaders, logBody, fk, redactHeadersWhen, log.pure[Option])(
+      RequestLogger(logHeaders, logBody, fk, redactHeadersWhen, log.pure[Option])(http)
     )
   }
 
+  def httpApp[F[_]: Concurrent](
```
rossabaker: We haven't specialized our others that need an `~>` like this.  I think we should do all or none.  Not strongly opinionated which.

ChristopherDavenport:
My biggest point of contention has been that there exists ping and other informational routes that we would prefer to explicitly NOT log as they are noisy.

I'm not really picky as far as our implementation of applying the `~>` or not. Just seemed easier to me to have default cases, as otherwise to use a basic logger people would need to know about `FunctionK`. Which is a high bar to set for logging http interactions. However, that also makes our api more verbose. I think perhaps we should add more of these moving forward.

ChristopherDavenport:
Plus it gives us some low hanging fruit issues which are always nice.

PR #2398, opened 2019-02-09
Allow specification of media type

### LINKED ISSUE #2397: Unable to override media type in EntityDecoder
Following discussion took place on gitter
@NAJ10
Feb 02 18:12
How do I create an EntityDecoder that when a GET request is sent sends a custom media type in the Accept Header? At the moment CirceInstances.jsonDecoderAdaptive causes the media type of application/json to be sent in the Accept header and I want to override it. Essentially I am having same problem as reported at hmrc/vat-api#212 as I get a 406 HTTP status code back when I try to call the endpoint.

Heikki Vesalainen
@hvesalai
Feb 02 22:20
Obviously I've been away too long if EntityDecoders have anything to do with request headers
But I don't think things are that bad. There must be some misunderstanding. @NAJ10, just create a GET request with Accept. There's an example of that at https://http4s.org/v1.0/client/
NAJ10
@NAJ10
Feb 02 22:41
Hi Heikki. I used that earlier today to write and run the code. However what happens is that two Accept headers are in the request. First off one for application/vnd.hmrc.1.0+json and then one for application/json. The server responds with a 406 HTTP Status as per the documentation on their site.
Heikki Vesalainen
@hvesalai
Feb 02 22:47
So who puts the wrong one in?
NAJ10
@NAJ10
Feb 02 22:49
CirceInstances.jsonDecoderAdaptive
Heikki Vesalainen
@hvesalai
Feb 02 22:51
did you check https://http4s.org/v1.0/entity/
and https://github.com/http4s/http4s/blob/master/circe/src/main/scala/org/http4s/circe/CirceInstances.scala#L53
NAJ10
@NAJ10
Feb 02 22:57
I was using the section Examples Send a GET request, treating the response as a string in https://http4s.org/v0.20/client/
And yes - line 53 is what I noticed in jsonDecoderAdaptive function
but of course I am not treating the response as a string - I have an EntityDecoder from a Circe Decoder
Heikki Vesalainen
@hvesalai
Feb 02 23:04
But in any case I do think things are bad when EntityDecoder is tampering with the request headers. Just for the record, I think that is a design failure in http4s
Ross A. Baker
@rossabaker
Feb 03 01:56
Wait, EntityEncoder or EntityDecoder?
NAJ10
@NAJ10
Feb 03 08:35
@rossabaker EntityDecoder. The HTTP operation is a GET request that returns JSON so the entity has to be decoded. I admit I have a implicit EntityEncoder also in scope for this case class but I presume the EntityEncoder is not relevant to this request
NAJ10
@NAJ10
Feb 03 08:58
This is what the RequestLogger says I am sending [ForkJoinPool-1-worker-0] INFO o.h.client.middleware.RequestLogger - HTTP/1.1 GET https://test-api.service.hmrc.gov.uk/hello/user Headers(Authorization: <REDACTED>, Accept: application/vnd.hmrc.1.0+json, Accept: application/json) body=""
I think I get the 406 because of the 2nd Accept header for application/json. I want to be able to find a way to send a request without that 2nd Accept header which my code is not adding itself.
Heikki Vesalainen
@hvesalai
Feb 03 19:15
There is this: You can construct new encoders via EntityEncoder.encodeBy or reuse an already existing one via EntityEncoder[F, T].contramap and withContentType. But there is no similar functionality for EntityDecoder whereby you could override the Accept header it sets
Ross A. Baker
@rossabaker
Feb 03 20:30
The addition of the Accept header is happening [here](https://github.com/http4s/http4s/blob/bfd2ec543552f334ff6d9dec44550d4dcb62dcb6/client/src/main/scala/org/http4s/client/DefaultClient.scala#L93.) That's a bad situation, because it sets the header after the request is passed to the client. There's not a great way to opt out of it. I agree that this is a design flaw. Just need to think about the best API to avoid it. Setting that is convenient in general.
I think @hvesalai might have a good idea with a parallel method on the decoder.

NAJ10:
fixes #2397

### DIFF
--- a/circe/src/main/scala/org/http4s/circe/CirceInstances.scala
+++ b/circe/src/main/scala/org/http4s/circe/CirceInstances.scala
@@ -1,16 +1,16 @@
 package org.http4s
 package circe
 
+import java.nio.ByteBuffer
+
 import cats._
+import cats.data.NonEmptyList
 import cats.effect._
 import cats.implicits._
 import fs2.Chunk
 import io.circe._
-import io.circe.jawn._
 import io.circe.jawn.CirceSupportParser.facade
-import java.nio.ByteBuffer
-
-import cats.data.NonEmptyList
+import io.circe.jawn._
 import org.http4s.headers.`Content-Type`
 import org.http4s.jawn.JawnInstances
 import org.typelevel.jawn.ParseException
@@ -47,10 +47,13 @@ trait CirceInstances extends JawnInstances {
 
   // default cutoff value is based on benchmarks results
   implicit def jsonDecoder[F[_]: Sync]: EntityDecoder[F, Json] =
-    jsonDecoderAdaptive(cutoff = 100000)
+    jsonDecoderAdaptive(cutoff = 100000, MediaType.application.json)
 
-  def jsonDecoderAdaptive[F[_]: Sync](cutoff: Long): EntityDecoder[F, Json] =
-    EntityDecoder.decodeBy(MediaType.application.json) { msg =>
+  def jsonDecoderAdaptive[F[_]: Sync](
+      cutoff: Long,
+      r1: MediaRange,
+      rs: MediaRange*): EntityDecoder[F, Json] =
+    EntityDecoder.decodeBy(r1, rs: _*) { msg =>
       msg.contentLength match {
         case Some(contentLength) if contentLength < cutoff =>
           jsonDecoderByteBufferImpl[F](msg)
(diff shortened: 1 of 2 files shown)


circe/src/main/scala/org/http4s/circe/CirceInstances.scala
```diff
@@ -59,7 +62,11 @@ trait CirceInstances extends JawnInstances {
     }
 
   def jsonOf[F[_]: Sync, A](implicit decoder: Decoder[A]): EntityDecoder[F, A] =
-    jsonDecoder[F].flatMapR { json =>
+    jsonOf(MediaType.application.json)
+
+  def jsonOf[F[_]: Sync, A](r1: MediaRange, rs: MediaRange*)(
```
ChristopherDavenport: This introduces an overload. Could we get this version renamed?

PR #2401, opened 2019-02-11
Remove Rig Plugin

ChristopherDavenport:


project/Http4sPlugin.scala
```diff
@@ -185,9 +183,8 @@ object Http4sPlugin extends AutoPlugin {
         releaseStepCommand("test:scalafmt::test").when(primary),
         releaseStepCommand("docs/makeSite").when(primary),
         releaseStepCommand("website/makeSite").when(primary),
-        openSonatypeRepo.when(publishable && release),
-        publishToSonatypeWithoutInstrumentation.when(publishable),
-        releaseAndClose.when(publishable && release),
+        releaseStepCommandAndRemaining("+publishSigned").when(publishable),
```
ChristopherDavenport: Any reason not to sign snapshots?

project/Http4sPlugin.scala
```diff
@@ -185,9 +183,8 @@ object Http4sPlugin extends AutoPlugin {
...
+        releaseStepCommand("sonatypeReleaseAll").when(publishable && release),
```
ChristopherDavenport: Introduces slight race conditions across repositories. I've never run into it in practice though.

project/Http4sPlugin.scala
```diff
@@ -34,12 +31,14 @@ object Http4sPlugin extends AutoPlugin {
 
   override def trigger = allRequirements
 
-  override def requires = RigPlugin && MimaPlugin && ScalafmtCorePlugin
+  override def requires = MimaPlugin && ScalafmtCorePlugin
 
   override lazy val buildSettings = Seq(
     // Many steps only run on one build. We distinguish the primary build from
     // secondary builds by the Travis build number.
+    isTravisBuild := sys.env.get("TRAVIS").isDefined,
```
ChristopherDavenport: We used this, so we introduce it.

ChristopherDavenport:
We don't really need a `CentralRequirementsPlugin` rather than an integrated set of settings anymore. but I also can just depend on it, so I'm not sure its a problem to leave it like that even without rig

ChristopherDavenport:
Should unblock as now we can depend on plugins that can interact with scala 2.13.0-M5 directly.

project/Http4sPlugin.scala
rossabaker: I don't see why not.

project/PrivateProject.scala
```diff
@@ -5,15 +5,27 @@ import com.typesafe.tools.mima.plugin.MimaPlugin
...
+      publishArtifact in Test := false,
+      publishArtifact in Compile := false,
+      publishArtifact in (Test, packageBin) := false,
+      publishArtifact in (Test, packageDoc) := false,
+      publishArtifact in (Test, packageSrc) := false,
+      publishArtifact in (Compile, packageBin) := false,
+      publishArtifact in (Compile, packageDoc) := false,
+      publishArtifact in (Compile, packageSrc) := false,
+      publishArtifact := false
+    ) ++ Seq(
```
rossabaker: Is `skip in publish := true` sufficient in modern sbt?

ChristopherDavenport: I can port in my personal anti publishing subset as well.

aeons: Should these use the newer slash-syntax instead? I like that a lot better.

Ie. `packageSrc / Compile / publishArtifact := false`

ChristopherDavenport: I could not get this to work, we can adopt it in a future iteration if you can get it running. 😄

PR #2414, opened 2019-02-13
Configure AHC to Use NoOp CookieStore by Default

### LINKED ISSUE #2413: Async-Http-Client Uses Cookie Store that Circumvents Intended Strategy
The async-http-store uses a cookie store that manages its own cookies for a session under the surface, as this is not managed by the http4s client and strategies we use in http4s this can lead to differences in behavior among backends.

My suggestion is to build a noop cookie store that will mean all the cookie management happens on the http4s side.

ChristopherDavenport:
Fixes #2413

### DIFF
--- a/async-http-client/src/main/scala/org/http4s/client/asynchttpclient/AsyncHttpClient.scala
+++ b/async-http-client/src/main/scala/org/http4s/client/asynchttpclient/AsyncHttpClient.scala
@@ -19,6 +19,9 @@ import org.http4s.util.threads._
 import org.log4s.getLogger
 import org.reactivestreams.Publisher
 import scala.collection.JavaConverters._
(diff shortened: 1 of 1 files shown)


async-http-client/src/main/scala/org/http4s/client/asynchttpclient/AsyncHttpClient.scala
```diff
@@ -145,4 +149,12 @@ object AsyncHttpClient {
     Headers(headers.asScala.map { header =>
       Header(header.getKey, header.getValue)
     }.toList)
+  private class NoOpCookieStore extends CookieStore {
+    val empty: java.util.List[Cookie] = new java.util.ArrayList()
```
rossabaker: For what it's worth, Java's constant is called `Collections.EMPTY_LIST`.

PR #2415, opened 2019-02-13
Update Changelog and Set Release for 0.18.22

ChristopherDavenport:


website/src/hugo/content/changelog.md
```diff
@@ -8,6 +8,11 @@ Maintenance branches are merged before each new release. This change log is
 ordered chronologically, so each release contains all changes described below
 it.
 
+# v0.18.22 (2019-02-13)
+
+## Bug fixes
```
rossabaker: Arguably more of an enhancement than a bugfix, but the important thing is to release it.

PR #2416, opened 2019-02-13
Port Ember Into Http4s

ChristopherDavenport:
Fully functional backend for http4s.


rossabaker:
We need to figure out our strategy here.  For documentation purposes, do we need a reference implementation in the main repo as other backends split off?  If yes, then I think Ember is the candidate for that role.  If no, then I think all the backends should split off.

ChristopherDavenport:
Closed in favor of #2696

Issue #2402, opened 2019-02-11
Create HttpApp and HttpRoutes Defaults for Server Middleware


ChristopherDavenport:
Server Middleware that is defined on Http[F, G] can sometimes be hard to reason about for folks who are trying to get a handle on the Http process. Rather than forcing individuals to learn Kleisli Composition, and FunctionK, we can create function which summon that middleware on HttpRoutes, or HttpApp respectively with `OptionT.liftK` and `FunctionK.id`. As its just a default argument being plugged in, but will make the middleware far more approachable without losing any of the composition the get Http types give us.

So each middleware would get on that take and `HttpRoutes[F]`, and an `HttpApp[F]`, preferably names `def httpRoutes` and `def httpApp`

ChristopherDavenport:
Additionally a scalafix for the middlewares that have been shifted this way would also be ideal. Especially `Logger.apply => Logger.httpApp`

voidcontext:
Can I take this one? I'd be happy if could contribute to this project... :)

ChristopherDavenport:
Absolutely, try to PR 1 at a time. However we'd love the support.

voidcontext:
@ChristopherDavenport I've opened the first PR and I am happy to discuss it.

voidcontext:
I did some investigations regarding the status of the different middlewares and this is what I found:
```
| Middleware          | httpRoutes | httpApp         | completion | notes                                                   |
|---------------------+------------+-----------------+------------+---------------------------------------------------------|
| AutoSlash           | done       | no value        | done       | MonoidK                                                 |
| CORS                | done       | done            | done       |                                                         |
| CSRF                | ?          | ?               | ?          | Mulitple special builders                               |
| Caching             | ?          | ?               | ?          | Multiple special builders                               |
| ChunkAggregator     | done       | done            | done       |                                                         |
| Date                | done       | done            | done       |                                                         |
| DefaultHead         | -          | no value        | -          | MonoidK                                                 |
| EntityLimiter       | -          | -               | -          |                                                         |
| ErrorAction         | ?          | ?               | ?          | Multiple constructors                                   |
| ErrorHandling       | -          | -               | -          |                                                         |
| GZip                | done       | done            | done       |                                                         |
| HSTS                | ?          | ?               | ?          | Multiple constructors                                   |
| HeaderEcho          | -          | -               | -          |                                                         |
| HttpMethodOverrider | -          | -               | -          | `httpRoutes` implementation might not be trivial        |
[...]

PR #2429, opened 2019-02-20
ChunkAggregator for HttpRoutes and HttpApp

voidcontext:
This PR adds convenience shortcuts to the ChunkAggregattor to apply the middleware on HttpRoutes and HttpApps.

Related issue: #2402

### DIFF
--- a/server/src/main/scala/org/http4s/server/middleware/ChunkAggregator.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/ChunkAggregator.scala
@@ -25,13 +25,11 @@ object ChunkAggregator {
           })
     }
 
-  def httpRoutes[F[_]: Sync](httpRoutes: HttpRoutes[F]): HttpRoutes[F] = {
+  def httpRoutes[F[_]: Sync](httpRoutes: HttpRoutes[F]): HttpRoutes[F] =
     apply(OptionT.liftK[F])(httpRoutes)
-  }
 
-  def httpApp[F[_]: Sync](httpApp: HttpApp[F]): HttpApp[F] = {
+  def httpApp[F[_]: Sync](httpApp: HttpApp[F]): HttpApp[F] =
     apply(FunctionK.id[F])(httpApp)
-  }
 
   private def removeChunkedTransferEncoding[G[_]: Functor](
       resp: Response[G],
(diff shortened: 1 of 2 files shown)


rossabaker:
The reason I keep bringing up traits is that I still haven't seen another maintainable technique that doesn't add an import tax to the user.  If typeclass syntax on the singleton types went into the implicits that are already imported, fine, but this would be implicits in another package, as well as a lawless typeclass.  And I fail to see how it's worse than copy and paste.

voidcontext:
@rossabaker I have a feeling I've missed some of the earlier discussions around this topic. Could you please explain what you mean when you say "but this would be implicits in another package, as well as a lawless typeclass" ?

In general I agree, it would be nice to give the middlewares a structure, where these kinds of shortcuts could be implemented in a general way. Let me dive into this and do some experiments.

Meanwhile I am happy to park this PR until we can reach consensus.

rossabaker:
I think we can merge this because we don't have a consensus on a better way forward and have been arguing about this pattern for months.  But I'd still like to solve this general problem in an agreeable way, because it keeps coming up.

voidcontext:
As far as I can see there's already a middleware type:

```
type Middleware[F[_], A, B, C, D] = Kleisli[F, A, B] => Kleisli[F, C, D]
```

Which is good, but not really used in the code. Looking at the current middleware implementations it seems in most of the cases either the request or the response is fixed, sometimes both (`Http[F, G]`). We could introduce something like `type RequestAwareMiddleware[F, G, B] = Kleisli[F, Request[G], B]` and  `ResponseAwareMiddleware` the same way, then the current implementations could return e,g, 'Reader[RequestAwareMiddleware[F, G, B], RequestAwareMiddleware[F, G, B]]', which we could extend with implicit classes to add the necessary syntax (shortcuts). Disclaimer: I didn't check if this could be implemented, or if it would cover all the cases. Just an idea. 

Probably not for the upcoming 0.20 release because of backwards compatibility, but if this is something worth to explore for future releases, I'd happily build a POC.

rossabaker:
I was thinking it couldn't be done without adding to the import tax, but I think that's wrong.  Even though the middlewares all live in server or client, the shape of middleware is core, and could be in the org.http4s.implicits._ that many people already implement.

I am not entirely comfortable adding syntax to general types to support things that aren't typeclasses, but that's exactly what we've done with `.orNotFound`.

I don't have a strong opinion.  Does anyone else?

rossabaker:
I think we should open a separate issue, because there was consensus to move forward with this.

PR #2453, opened 2019-03-16
Addressed performance issue where connection would be closed every ti…

cmcmteixeira:
…me and not only when an error ocurred.

### DIFF
--- a/blaze-client/src/main/scala/org/http4s/client/blaze/Http1Connection.scala
+++ b/blaze-client/src/main/scala/org/http4s/client/blaze/Http1Connection.scala
@@ -269,7 +269,7 @@ private final class Http1Connection[F[_]](val requestKey: RequestKey, config: Bl
                   .eval(Async.shift(executionContext) *> F.delay {
                     trailerCleanup(); cleanup(); stageShutdown()
                   })
-                  .flatMap(_ => Stream.raiseError(err))
+                  .flatMap(_ => Stream.raiseError[Byte](err))
               }
           }
         }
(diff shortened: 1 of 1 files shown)


cmcmteixeira:
I believe there could be a couple of issues with this code:

1) If the stream is cancelled, stageShutdown would never run
2) trailerCleanup(); cleanup();  will run twice if the stream terminates with an error

Not sure how relevant these constraints are unfortunately..

rossabaker:
> I believe that if the stream is cancelled, stageShutdown could never run

`onFinalizeCase` should give the hook to run this.  In practice, the client is likely to disconnect and trigger a shutdown, but we can be a bit more aggressive on our end.

> trailerCleanup(); cleanup(); will run after the stageShutdown() step

I don't think this will hurt anything, but it doesn't even need to run at all.  It could be another case in `onFinalizeCase`.

I _think_ the logic could be:

```scala
            attributes -> rawBody
              .onFinalizeCase {
                case ExitCase.Completed =>
                  Async.shift(executionContext) *> F.delay { trailerCleanup(); cleanup() }
                case _ =>
                  Async.shift(executionContext) *> F.delay(stageShutdown())
              }
```

cmcmteixeira:
Ah! `onFinalizeCase` seems to be only part of the fs2 1.x series.
I can make another PR targeting Http4s 0.20 series with the `onFinalizeCase` but for the 0.18 it seems like we're stuck with having to to `onFinalize` + `handleErrorWith`;

cmcmteixeira:
Also, because of the limitation of only having `onFinalize` & `handleErrorWith` I can't think of a way to be able to run only `stageShutdown` in case of error and `trailerCleanup+cleanup` in case of success; they will both have to run in case of error..

rossabaker:
Oh, I see.  This PR targets master, so I thought you were giving up on the 0.18.x fix.  That should target release-0.18.x.

I don't see any harm in running `trailerCleanup` and `cleanup` on error in 0.18.x.  It's cleaning up state that will be discarded, but it shouldn't negatively affect semantics.  The thing we must avoid is calling `stageShutdown()` on success.

cmcmteixeira:
Ups, my bad! Fixed now 👍 
I'll create a PR with the onFinalizeCase for 0.20x in the mean time to get this thing over with :)

PR #2462, opened 2019-03-20
WIP: 2.13 Remove collections

ChristopherDavenport:
This is an approach to a 2.13 migration by removing all dependencies internally on collections.


ChristopherDavenport:
Headers was completely successful.

I've started Query and we seem to have nothing but insane interactions.

core/src/main/scala/org/http4s/Headers.scala
```diff
@@ -5,28 +5,27 @@ import cats.implicits._
...
+  def iterator: Iterator[Header] = headers.iterator
 
-  override def foreach[B](f: Header => B): Unit = headers.foreach(f)
+  def tail: Headers = Headers(headers match {
+    case _ :: xs => xs
+    case n@Nil => n 
+  })
 
-  def iterator: Iterator[Header] = headers.iterator
+  def head: Header = headers.headOption.fold[Header](throw new NoSuchElementException)(identity)
```
rossabaker: I think the only reason to have partial methods was to be a proper Scala collection.  If we're abandoning that goal, I would call this `headOption`.

Now, for backward compatibility, we could go nuts with deprecation on all the collection methods we've left behind, or we could pick a few major ones like this.

core/src/main/scala/org/http4s/Headers.scala
```diff
@@ -111,31 +120,43 @@ final class Headers private (headers: List[Header])
...
 
 object Headers {
-  val empty = apply()
+  val empty = apply(List.empty)
+
+  def of(headers: Header*): Headers = 
+    Headers(headers.toList)
+
+  @deprecated("Use Headers.of", "0.20.0")
+  def apply(headers: Header*): Headers =
```
rossabaker: To eliminate the overload?  I like it.

dsl/src/main/scala/org/http4s/dsl/impl/ResponseGenerator.scala
```diff
@@ -76,10 +76,10 @@ trait LocationResponseGenerator[F[_], G[_]] extends Any with EntityResponseGener
   @deprecated("Use `apply(Location(location))` instead", "0.18.0-M2")
   def apply(location: Uri)(implicit F: Applicative[F]): F[Response[G]] =
     F.pure(
-      Response[G](status = status, headers = Headers(`Content-Length`.zero, Location(location))))
+      Response[G](status = status, headers = Headers(`Content-Length`.zero :: Location(location) :: Nil)))
```
rossabaker: Do you find this preferrable to `Headers.of`?

server/src/main/scala/org/http4s/server/middleware/ChunkAggregator.scala
```diff
@@ -35,7 +35,7 @@ object ChunkAggregator {
       resp: Response[G],
       len: Long): Response[G] =
     resp.transformHeaders { headers =>
-      val hs = headers.flatMap {
+      val hs = Headers(headers.toList.flatMap {
```
rossabaker: This seems like one of the stronger arguments for ceasing to be a standard collection: it's not real easy to predict what type a `flatMap` returns in the old implementation.

core/src/main/scala/org/http4s/Headers.scala
```diff
@@ -89,17 +84,31 @@ final class Headers private (headers: List[Header])
...
+  def filterNot(f: Header => Boolean): Headers =
+    Headers(headers.filterNot(f))
+
+  def filter(f: Header => Boolean): Headers =
+    Headers(headers.filter(f))
+
+  def collectFirst[B](f: PartialFunction[Header,B]): Option[B] = 
+    headers.collectFirst(f)
+
+  def foldMap[B: Monoid](f: Header => B): B =
```
rossabaker: Was there any principle to which of these forwarders was included and which were not?  The choices seem reasonable, but I can't think off the top of my head what made the cut and what didn't and why.

dsl/src/main/scala/org/http4s/dsl/impl/ResponseGenerator.scala
ChristopherDavenport: I originally was just going to deprecate, but realized it's used readily so transitioned.

core/src/main/scala/org/http4s/Headers.scala
ChristopherDavenport: This was being used for a fully tailrecursive method downstream, however we can instead convert to list in that location.

core/src/main/scala/org/http4s/Headers.scala
ChristopherDavenport: What was being used combined with what I consider the reasonable methods. I went for foldableLike but fixed on the base type.

rossabaker: Is there a `FoldableLike` somewhere?  I've thought of the concept a few times and then remembered, "duh, Foldable is F[_]`".

Whatever, I think you're on the right track.

rossabaker:
Probably too breaking of a change, but would it be simpler to eliminate `Headers` as a class, have a `Chain[Header]` on `Message`, and put the header manipulation functions on `Message`?  I wonder how hard that would ripple in practice.  Probably very.

I'm okay with the direction.  I concede it should no longer be a standard library collection.

rossabaker:
Where are we between this and #2471?

ChristopherDavenport:
Aggressively behind. 

Moving has hit my time hard. (Unsurprisingly). 

Trying to figure out a strategy to get all the necessary components in place, but I'm not sure I have a rock solid strategy to get us all the way there.

rossabaker:
I guess my question is more in terms of what's the status of this PR, given that part of it was split off?  Are we going to continue to work here?

I could probably take a chunk or two of work, but I'm not sure what the chunks are.

ChristopherDavenport:
Well the forked bit can be merged back in, I think figuring out the query component is the next big task.

rossabaker:
Forked bit being #2471?

ChristopherDavenport:
Roger that.

rossabaker:
Incorporated back into #2393.

Issue #2461, opened 2019-03-19
WaitQueueFailure defines incorrect method name


ChristopherDavenport:
https://github.com/http4s/http4s/blob/master/client/src/main/scala/org/http4s/client/PoolManager.scala#L16

defines message and does not define getMessage so throwable render with a null message. We can override that and delegate to message and maintain full compatibility.

VladKopanev:
I would like to fix this!

aeons:
Closed in #2466

PR #2466, opened 2019-03-20
Fix issue #2461

VladKopanev:


VladKopanev:
@rossabaker it seems like this method is not used at least in `htt4s` projects. I don't know if it has been used in any dependent projects outside `http4s`.  Yeah, maybe it's a good idea to have this `message` deprecated for a while. Will do.

VladKopanev:
I also want to clarify this before committing: which version should I put in `since` field of `deprecated` annotation? Is it going to be a "0.20.0-M7" release?

ChristopherDavenport:
`0.20.0` is fine presently.

PR #2469, opened 2019-03-22
Add scala-steward to adopters

fthomas:


fthomas:
I feel the same about the projects I'm depending on in my work and for which scala-steward now helps with updates. :-)

Issue #2386, opened 2019-02-05
BlazeClient: requestTimeout is not enforced


mblaze:
## Problem Description

`requestTimeout` in `BlazeClient` is not enforced on certain requests.

**Expected behavior**: A timeout exception is thrown after 1 second.

**Actual behavior**: The request seems to be stuck indefinitely.

**Related issues**: #2338 (almost certainly the same source of the problem)

## Source

> Main.scala
```scala
import cats.effect._
import org.http4s._
import org.http4s.client.blaze.BlazeClientBuilder

import scala.concurrent.duration._
import scala.concurrent.ExecutionContext

object Main extends IOApp {

  override def run(args: List[String]): IO[ExitCode] = for {
    c <- BlazeClientBuilder[IO](ExecutionContext.global)
        .withRequestTimeout(1.second)
        .resource
        .allocated
    (client, _) = c
    _ <- IO { println("Sending a request.") }
    _ <- client.fetch(Request[IO](Method.GET, Uri.unsafeFromString("http://1.2.3.4:12345/")))(_ => IO.unit)
    _ <- IO { println("Got a response.") }
  } yield ExitCode.Success
}
```

> build.sbt
```scala
name := "http4s-client-request-timeout-bug"

version := "0.1"

scalaVersion := "2.12.8"

libraryDependencies ++= Seq(
  "org.typelevel" %% "cats-core" % "1.5.0",
  "org.typelevel" %% "cats-effect" % "1.1.0",
  "org.http4s" %% "http4s-blaze-client" % "0.20.0-M5"
)
```

kevinmeredith:
```
$time(curl -vi http://1.2.3.4:12345/)
*   Trying 1.2.3.4...
* TCP_NODELAY set
* Connection failed
* connect to 1.2.3.4 port 12345 failed: Operation timed out
* Failed to connect to 1.2.3.4 port 12345: Operation timed out
* Closing connection 0
curl: (7) Failed to connect to 1.2.3.4 port 12345: Operation timed out

real	1m15.105s
user	0m0.008s
sys	0m0.008s
```

My understanding of the above output is that the TCP Connection failed to become established.

Looking at the comment for `requestTimeout`:

> requestTimeout maximum duration from the submission of a request through reading the body before a timeout.

https://github.com/http4s/http4s/blob/master/blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientConfig.scala#L17-L18

[...]

Issue #2338, opened 2019-01-04
Client hangs when firewall blocks request


tilmanrossmy:
If the client sends a tcp Syn and never gets an Syn-ack, the http4s client hangs forever and timeout settings don't have any effect. This is demonstrated in this gist:
https://gist.github.com/tilmanrossmy/cef2489230db0068b082061c9e840d22
Using a java.net Client the connectTimeout Setting will end the request for the same URL as in the gist above:
https://gist.github.com/tilmanrossmy/04421b2efc8fc934a0dfb73b03ca933e

rossabaker:
I suspect this would require a change all the way into blaze, but it's worth pursuing.  I consider the lack of a reasonable timeout, even if hardcoded, a bug.

RafalSumislawski:
@tilmanrossmy @rossabaker I think this ticket can be closed now.

rossabaker:
Fixed by #2470.

PR #2470, opened 2019-03-23
Prevent never ending wait for cancelation of a failed (timed-out) attempt to acquire a connection.

RafalSumislawski:
This is a workaround for https://github.com/typelevel/cats-effect/issues/487 causing blaze client to wait indefinitely if a `requestTimeout` occurs while `manager.borrow(key)` is running and the borrow eventually fails. This is what happens when we try to send a request to an unreachable host (incorrect IP address, firewalls, etc.).

IMO this is a solution for #2338. It literally implements "a reasonable timeout, even if hardcoded" as @rossabaker described it ;).

This is also a partial solution for #2386 . The requestTimeout still isn't enforced,  instead we wait for an OS level timeout of an attempt to establish TCP connection, but at least the request fails eventually, instead of waiting indefinitely. I want to follow up with a more complete solution, but I feel it requires making some design decisions, which should be discussed with the maintainers.

### DIFF
--- a/blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClient.scala
+++ b/blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClient.scala
@@ -60,11 +60,13 @@ object BlazeClient {
             .handleError(e => logger.error(e)("Error invalidating connection"))
 
         def borrow =
-          Resource.makeCase(manager.borrow(key)) {
-            case (_, ExitCase.Completed) =>
-              F.unit
-            case (next, ExitCase.Error(_) | ExitCase.Canceled) =>
+        // TODO: The `attempt` here is a workaround for https://github.com/typelevel/cats-effect/issues/487 .
+        //  It can be removed once the issue is resolved.
+          Resource.makeCase(manager.borrow(key).attempt) {
+            case (Right(next), ExitCase.Error(_) | ExitCase.Canceled) =>
               invalidate(next.connection)
+            case _ =>
+              F.unit
           }
 
         def idleTimeoutStage(conn: A) =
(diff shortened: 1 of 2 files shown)


blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClient.scala
```diff
@@ -60,11 +60,13 @@ object BlazeClient {
...
-          Resource.makeCase(manager.borrow(key)) {
-            case (_, ExitCase.Completed) =>
-              F.unit
-            case (next, ExitCase.Error(_) | ExitCase.Canceled) =>
+        // TODO: The `attempt` here is a workaround for https://github.com/typelevel/cats-effect/issues/487 .
+        //  It can be removed once the issue is resolved.
+          Resource.makeCase(manager.borrow(key).attempt) {
+            case (Right(next), ExitCase.Error(_) | ExitCase.Canceled) =>
               invalidate(next.connection)
+            case _ =>
```
kevinmeredith: Could you please explain, @RafalSumislawski, why it's not necessary to explicitly handle the `case (Left(error), _)) => ...` case?

In other words, what's the import of returning `IO.unit` if, as I understand, `manager.borrow(key).attempt` returns an `F(Left(RuntimeException(...)))`?

RafalSumislawski: The lines 65-70 are the `release` function. If acquiring the connection failed, there's nothing to release. There's also no need to propagate the error here, as it will already be propagated by `case Left(t) => F.raiseError(t)` on line 88. So we do nothing aka return `IO.unit`.

blaze-client/src/test/scala/org/http4s/client/blaze/ClientTimeoutSpec.scala
```diff
@@ -156,5 +157,29 @@ class ClientTimeoutSpec extends Http4sSpec {
 
       c.fetchAs[String](FooRequest).unsafeRunSync must_== "done"
     }
+
+    // Regression test for: https://github.com/http4s/http4s/issues/2386
```
kevinmeredith: Nice, well-commented test!

blaze-client/src/test/scala/org/http4s/client/blaze/ClientTimeoutSpec.scala
```diff
@@ -156,5 +157,29 @@ class ClientTimeoutSpec extends Http4sSpec {
...
+    // and https://github.com/http4s/http4s/issues/2338
+    "Eventually timeout on connect timeout" in {
+      val manager = ConnectionManager.basic[IO, BlazeConnection[IO]]({ _ =>
+        // In a real use case this timeout is under OS's control (AsynchronousSocketChannel.connect)
+        IO.sleep(2.seconds) *> IO.raiseError[BlazeConnection[IO]](new IOException())
```
rossabaker: It ate my comment.

What are the actual requirements for sleeping here?  Somewhat longer than `requestTimeout`, less than the argument to `unsafeRunTimed`?  I'm concerned about blocking a thread for two seconds, which slows down test runs and bogs other things down.  I think the way we do this contributes to our instability in blaze tests on Travis CI.

The trick is that if the timeout is too aggressive, we run into race conditions, and if it's too slow, we have slow CI and can starve other tests.  If we were smarter about `cats.effect.Timer`, we'd have less problems like this, but that would require more plumbing into blaze.  That's well beyond the scope of this ticket.

Anyway, I'd like to make these timeouts as tight as we can without running into noisy failures.  Just making sure I understand the ordering.

blaze-client/src/test/scala/org/http4s/client/blaze/ClientTimeoutSpec.scala
```diff
@@ -156,5 +157,29 @@ class ClientTimeoutSpec extends Http4sSpec {
...
+        scheduler = tickWheel,
+        ec = testExecutionContext
+      )
+
+      // if the 5.seconds timeout is hit, it's a NoSuchElementException,
+      // if the requestTimeout = 1.second is hit then it's a TimeoutException
+      // if establishing connection fails first then it's an IOException
+
+      // The expected behaviour is that the requestTimeout will happen first, but fetchAs will additionally wait for the IO.sleep(2.seconds) to complete.
+      c.fetchAs[String](FooRequest).unsafeRunTimed(5.seconds).get must throwA[TimeoutException]
```
kevinmeredith: Per your PR's description:

>  The requestTimeout still isn't enforced, instead we wait for an OS level timeout of an attempt to establish TCP connection, but at least the request fails eventually, instead of waiting indefinitely.

Why wouldn't an `IOException` be thrown since "the requestTimeout still isn't enforced?"

blaze-client/src/test/scala/org/http4s/client/blaze/ClientTimeoutSpec.scala
RafalSumislawski: The requirement is that `requestTimeout` < `IO.sleep` < `unsafeRunTimed`. For as long as everything works this tests will run for the time of `IO.sleep`, so this it the number that we should care the most about. `requestTimeout` can be near zero. `unsafeRunTimed` affects only run time of failing test, so we can be a bit more generous with it.

Maybe we could go with something like `requestTimeout = 50.millis`, `IO.sleep(1.second)`, `unsafeRunTimed(2.seconds)` ? That would make this test case the fastest in the `ClientTimeoutSpec`.

blaze-client/src/test/scala/org/http4s/client/blaze/ClientTimeoutSpec.scala
RafalSumislawski: In short: The `requestTimeout` is **why** the request fails, so that's what is returned., but the value of the `requestTimeout` is not **when** the request completes.

In detail - here's what happens:
- sending a request (including establishing connection), and the requestTimeout are started in parallel racing with each other
- after 1 second the requestTimeout is hit, it wins the race. The `racePair` completes with TimeoutException (https://github.com/http4s/http4s/blob/3bed1e6088395cbd8b291f99172d68433b9b8c31/blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClient.scala#L147) , 
- fiber.cancel is called (https://github.com/http4s/http4s/blob/3bed1e6088395cbd8b291f99172d68433b9b8c31/blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClient.scala#L159) attempting to cancel the proces of sending the request. At that time that process is still during the `acuire` part of a `Bracket` which is uninterruptable. So the `fiber.cancel` waits until the aquire completes and then interrupts the `Bracket`.
- after 2 seconds from the start of the test, the aquire completes with  an IOException. The `fiber.cancel`completes and gets flatMapped with `F.raiseError(t)`, `t` being the original TimeoutException. That becomes the final result of the request.

kevinmeredith: Thanks for that detailed explanation, Rafał!

PR #2480, opened 2019-03-29
Static content documentation

rhyskeepence:
I didn't know (until recently) about the org.http4s.server.staticcontent package object. It seems to be a much more convenient way to host static content (one line, like wai-middleware-static). I've updated the docs to include it.

### DIFF
--- a/docs/src/main/tut/static.md
+++ b/docs/src/main/tut/static.md
@@ -2,6 +2,7 @@
 menu: main
 weight: 140
 title: Static Files
+---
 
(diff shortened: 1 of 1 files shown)


docs/src/main/tut/static.md
```diff
@@ -2,31 +2,59 @@
 menu: main
 weight: 140
 title: Static Files
----
```
aeons: I think this is needed for Hugo to interpret the frontmatter in the file.

rossabaker:
Looks like we haven't flagged all the flaky tests.

PR #2486, opened 2019-04-03
Change +*?(List) to +*?(Seq) for consistency

rossabaker:


rossabaker:
The CI breakage is addressed in #2485.

PR #2490, opened 2019-04-05
[WIP] initial support for the JDK 11+ Http Client API

amesgen:
An HTTP client using the [JDK 11+ Http Client API](https://docs.oracle.com/en/java/javase/11/docs/api/java.net.http/java/net/http/package-summary.html), based on [this](https://gitter.im/http4s/http4s?at=5ca63bfca84e0c501ad728fa). Only builds on 11+ - the build has to be fixed to conditionally include this subproject only on JDK 11 (I am not sure about the right way to do this - in particular as this module also has to be published).

### DIFF
--- a/jdk-http-client/src/main/scala/org/http4s/client/jdkhttpclient/JdkHttpClient.scala
+++ b/jdk-http-client/src/main/scala/org/http4s/client/jdkhttpclient/JdkHttpClient.scala
@@ -1,17 +1,20 @@
 package org.http4s.client.jdkhttpclient
 
-import java.net.URI
 import java.net.http.HttpRequest.BodyPublishers
 import java.net.http.HttpResponse.BodyHandlers
 import java.net.http.{HttpClient, HttpRequest, HttpResponse}
+import java.net.{ProxySelector, URI}
 import java.nio.ByteBuffer
+import java.time.{Duration => JDuration}
 import java.util
-import java.util.concurrent.Flow
+import java.util.concurrent.{Executor, Flow}
 
+import cats.ApplicativeError
 import cats.effect._
 import cats.implicits._
 import fs2.interop.reactivestreams._
 import fs2.{Chunk, Stream}
+import javax.net.ssl.SSLContext
 import org.http4s.client.Client
 import org.http4s.internal.fromCompletableFuture
 import org.http4s.util.CaseInsensitiveString
@@ -19,19 +22,89 @@ import org.http4s.{Header, Headers, HttpVersion, Request, Response, Status}
 import org.reactivestreams.FlowAdapters
 
 import scala.collection.JavaConverters._
+import scala.concurrent.duration._
 
+// TODO documentation
 object JdkHttpClient {
 
-  def apply[F[_]](builder: HttpClient.Builder = HttpClient.newBuilder)(
-      implicit F: ConcurrentEffect[F]): F[Client[F]] =
-    F.delay(builder.build).map { jdkHttpClient =>
-      Client[F] { req =>
-        Resource.liftF(
-          fromCompletableFuture(
-            F.delay(jdkHttpClient.sendAsync(convertRequest(req), BodyHandlers.ofPublisher)))
-            .flatMap(convertResponse(_)))
+  def apply[F[_]](
+      jdkHttpClient: HttpClient,
+      ignoredHeaders: Set[CaseInsensitiveString] = restrictedHeaders
+  )(implicit F: ConcurrentEffect[F]): Client[F] = {
+
+    def convertRequest(req: Request[F]): F[HttpRequest] =
+      convertHttpVersionFromHttp4s[F](req.httpVersion).map { version =>
+        val rb = HttpRequest.newBuilder
+          .method(
+            req.method.name, {
+              val publisher = FlowAdapters.toFlowPublisher(
+                StreamUnicastPublisher(req.body.chunks.map(_.toByteBuffer)))
+              if (req.isChunked)
+                BodyPublishers.fromPublisher(publisher)
+              else
+                req.contentLength.fold(BodyPublishers.noBody)(
+                  BodyPublishers.fromPublisher(publisher, _))
+            }
+          )
+          .uri(URI.create(req.uri.renderString))
+          .version(version)
+        val headers = req.headers.iterator
+        // hacky workaround (see e.g. https://stackoverflow.com/questions/53979173)
+          .filterNot(h => ignoredHeaders.contains(h.name))
+          .flatMap(h => Iterator(h.name.value, h.value))
+          .toArray
+        (if (headers.isEmpty) rb else rb.headers(headers: _*)).build
       }
+
+    def convertResponse(res: HttpResponse[Flow.Publisher[util.List[ByteBuffer]]]): F[Response[F]] =
+      F.fromEither(Status.fromInt(res.statusCode)).map { status =>
+        Response(
+          status = status,
+          headers = Headers(res.headers.map.asScala.flatMap {
+            case (k, vs) => vs.asScala.map(Header(k, _))
+          }.toList),
+          httpVersion = res.version match {
+            case HttpClient.Version.HTTP_1_1 => HttpVersion.`HTTP/1.1`
+            case HttpClient.Version.HTTP_2 => HttpVersion.`HTTP/2.0`
+          },
+          body = FlowAdapters
+            .toPublisher(res.body)
+            .toStream[F]
+            .flatMap(bs =>
+              Stream.fromIterator(bs.asScala.map(Chunk.byteBuffer).iterator).flatMap(Stream.chunk))
+        )
+      }
+
+    Client[F] { req =>
+      Resource.liftF(
+        convertRequest(req)
+          .flatMap(r =>
+            fromCompletableFuture(F.delay(jdkHttpClient.sendAsync(r, BodyHandlers.ofPublisher))))
+          .flatMap(convertResponse))
     }
+  }
+
+  def createUnderlyingClient[F[_]](
+      httpVersion: HttpVersion = HttpVersion.`HTTP/1.1`,
+      connectTimeout: Duration = 10.seconds,
+      redirectPolicy: HttpClient.Redirect = HttpClient.Redirect.NEVER,
+      sslContext: => SSLContext = SSLContext.getDefault,
+      proxySelector: => ProxySelector = ProxySelector.getDefault,
+      customEC: Option[Executor] = None
+  )(implicit F: Sync[F]): F[HttpClient] =
+    convertHttpVersionFromHttp4s[F](httpVersion).flatMap(version =>
+      F.delay {
+        val builder = HttpClient.newBuilder
+          .version(version)
+          .connectTimeout(JDuration.ofNanos(connectTimeout.toNanos))
+          .followRedirects(redirectPolicy)
+          .sslContext(sslContext)
+          .proxy(proxySelector)
+        customEC.fold(builder)(builder.executor).build()
+    })
+
+  def simple[F[_]: ConcurrentEffect]: F[Client[F]] =
+    createUnderlyingClient[F]().map(apply[F](_))
 
   // see jdk.internal.net.http.common.Utils#DISALLOWED_HEADERS_SET
   private val restrictedHeaders =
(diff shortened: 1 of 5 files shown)


jdk-http-client/src/main/scala/org/http4s/client/jdkhttpclient/JdkHttpClient.scala
```diff
@@ -0,0 +1,109 @@
...
+import org.http4s.{Header, Headers, HttpVersion, Request, Response, Status}
+import org.reactivestreams.FlowAdapters
+
+import scala.collection.JavaConverters._
+
+object JdkHttpClient {
+
+  def apply[F[_]](builder: HttpClient.Builder = HttpClient.newBuilder)(
+      implicit F: ConcurrentEffect[F]): F[Client[F]] =
+    F.delay(builder.build).map { jdkHttpClient =>
```
amesgen: Building the http client is an effect (default Executor creation, SSLContext).

jdk-http-client/src/main/scala/org/http4s/client/jdkhttpclient/JdkHttpClient.scala
```diff
@@ -0,0 +1,109 @@
...
+          case _: CancellationException => ()
+          case ex: CompletionException if ex.getCause ne null => cb(Left(ex.getCause))
+          case ex => cb(Left(ex))
+        }
+      })
+      F.delay { cf.cancel(true); () }
+    })
+
+  // see jdk.internal.net.http.common.Utils#DISALLOWED_HEADERS_SET
+  private val restrictedHeaders =
```
amesgen: For some reason some headers are "restricted" (just as in `HttpUrlConnection`). In JDK11, there is no way to disable this behavior (fixed in 12+).

jdk-http-client/src/main/scala/org/http4s/client/jdkhttpclient/JdkHttpClient.scala
```diff
@@ -0,0 +1,109 @@
...
+        case HttpVersion.`HTTP/2.0` => HttpClient.Version.HTTP_2
+        case _ =>
+          throw new IllegalStateException("unsupported http version") // TODO better error handling
+      })
+    val headers = req.headers.iterator.flatMap { h =>
+      // hacky workaround (see e.g. https://stackoverflow.com/questions/53979173)
+      if (restrictedHeaders.contains(h.name)) Iterator.empty
+      else Iterator(h.name.value, h.value)
+    }.toArray
+    (if (headers.isEmpty) rb else rb.headers(headers: _*)).build
```
amesgen: For some reason, `.headers()` throws an exception if `headers` is empty...

project/Http4sPlugin.scala
```diff
@@ -341,6 +341,7 @@ object Http4sPlugin extends AutoPlugin {
   lazy val prometheusHotspot                = "io.prometheus"          %  "simpleclient_hotspot"      % prometheusClient.revision
   lazy val parboiled                        = "org.http4s"             %% "parboiled"                 % "1.0.1"
   lazy val quasiquotes                      = "org.scalamacros"        %% "quasiquotes"               % "2.1.0"
+  lazy val reactiveStreamsFlowAdapters      = "org.reactivestreams"    %  "reactive-streams-flow-adapters" % "1.0.2" // TODO formatting
```
amesgen: huge diff < aligned code?

aeons:
Nice.

Should it be JDK? Is it depending on the JDK in any way? Or just JRE?

amesgen:
No, the name can be changed. It was often called this way during its incubator period (see e.g. [here](https://openjdk.java.net/groups/net/httpclient/intro-incubating.html)). Maybe "Java 11 HTTP Client" or just "Java HTTP Client"? ([official project page](https://openjdk.java.net/groups/net/httpclient/))

amesgen:
- The "restricted header" hack is non-portable, as the set of restricted headers is implementation-defined.
 - Maybe the default config should be changed, in particular to prevent the creation of an unwanted `Executor`.
 - One step further: `JdkHttpClient#apply` could take an `HttpClient` (one could add helper methods with good defaults). This has the advantage that it is easy to add websocket support in the future without requiring to create one `HttpClient` for a `Client[F]` and another one for the websocket interface. But this is a general thing for all `Client[F]` implementations.

jdk-http-client/src/main/scala/org/http4s/client/jdkhttpclient/JdkHttpClient.scala
```diff
@@ -0,0 +1,111 @@
...
+      Client[F] { req =>
+        Resource.liftF {
+          F.delay(jdkHttpClient.sendAsync(convertRequest(req), BodyHandlers.ofPublisher))
+            .flatMap(fromJavaFuture(_))
+            .flatMap(convertResponse(_))
+        }
+      }
+    }
+
+  private def fromJavaFuture[F[_], A](cf: CompletableFuture[A])(implicit F: Concurrent[F]): F[A] =
```
rossabaker: This might be good in org.http4s.internal in core.  I can imagine other backends benefiting from this.

jdk-http-client/src/main/scala/org/http4s/client/jdkhttpclient/JdkHttpClient.scala
rossabaker: The `BlazeClientBuilder` takes an `ExecutionContext` as an argument instead of returning an `Effect`.  We might consider that pattern here.

jdk-http-client/src/main/scala/org/http4s/client/jdkhttpclient/JdkHttpClient.scala
```diff
@@ -0,0 +1,111 @@
...
+import fs2.interop.reactivestreams._
+import fs2.{Chunk, Stream}
+import org.http4s.client.Client
+import org.http4s.util.CaseInsensitiveString
+import org.http4s.{Header, Headers, HttpVersion, Request, Response, Status}
+import org.reactivestreams.FlowAdapters
+
+import scala.collection.JavaConverters._
+
+object JdkHttpClient {
```
rossabaker: I'm not thrilled with the name, but I don't have a better one.

jdk-http-client/src/main/scala/org/http4s/client/jdkhttpclient/JdkHttpClient.scala
```diff
@@ -0,0 +1,111 @@
...
+          else
+            req.contentLength.fold(BodyPublishers.noBody)(
+              BodyPublishers.fromPublisher(publisher, _))
+        }
+      )
+      .uri(URI.create(req.uri.renderString))
+      .version(req.httpVersion match {
+        case HttpVersion.`HTTP/1.1` => HttpClient.Version.HTTP_1_1
+        case HttpVersion.`HTTP/2.0` => HttpClient.Version.HTTP_2
+        case _ => HttpClient.Version.HTTP_1_1
```
rossabaker: Does this not speak 1.0?

project/Http4sPlugin.scala
rossabaker: Ugh.  Leave it for now.  Let's get 0.18 EOL'ed with the release of 0.20, and we can at least reduce the pain it will cause.

rossabaker:
I should have read the comments before making my own.  Sorry, I'm still catching up from the conference.

> The "restricted header" hack is non-portable, as the set of restricted headers is implementation-defined.

Maybe that argues for them being a configuration option, with a we-hope-it-works default?

> One step further: JdkHttpClient#apply could take an HttpClient

I'm fond of this idea, whether it's in addition to the builder or replaces it.

jdk-http-client/src/main/scala/org/http4s/client/jdkhttpclient/JdkHttpClient.scala
amesgen: This is a slight adaption of [this](https://github.com/typelevel/cats-effect/issues/160), and it doesn't seem planned to add it to cats-effect in the near future, so it makes sense to make it an http4s internal utility.

jdk-http-client/src/main/scala/org/http4s/client/jdkhttpclient/JdkHttpClient.scala
amesgen: No, [only 1.1 and 2](https://docs.oracle.com/en/java/javase/11/docs/api/java.net.http/java/net/http/HttpClient.Version.html). One could raise an error instead, but I don't think this is much of an issue (for example, the AsyncHttpClient ignores the HTTP version completely).

jdk-http-client/src/main/scala/org/http4s/client/jdkhttpclient/JdkHttpClient.scala
```diff
@@ -0,0 +1,131 @@
...
+          .flatMap(convertResponse))
+    }
+  }
+
+  def createUnderlyingClient[F[_]](
+      httpVersion: HttpVersion = HttpVersion.`HTTP/1.1`,
+      connectTimeout: Duration = 10.seconds,
+      redirectPolicy: HttpClient.Redirect = HttpClient.Redirect.NEVER,
+      sslContext: => SSLContext = SSLContext.getDefault,
+      proxySelector: => ProxySelector = ProxySelector.getDefault,
```
amesgen: There might be better names for this function. `sslContext` and `proxySelector` are by-name, as they can throw (very unlikely) - not the best solution.

amesgen:
> For building this, I think we could leave it out of the aggregate projects and then add a step to the build matrix for it. We already publish from multiple build steps, so even that wouldn't be new, though we've discussed consolidating all the publishing builds to a single step to avoid two of three cross builds making it when a test flickers. I still hope someone comes with a more appealing idea.

Another approach is building everything using JDK 11. It is not [not recommended](https://docs.scala-lang.org/overviews/jdk-compatibility/overview.html#jdk-11-compatibility-notes), but certainly possible (and for this module ofc necessary) and would greatly simplify this issue. On my PC, the entire test suite runs fine for all Scala versions on JDK 11, and rudimentary tests indicate that the produced JARs can be used in other Scala projects using JDK 8.

jdk-http-client/src/main/scala/org/http4s/client/jdkhttpclient/JdkHttpClient.scala
```diff
@@ -0,0 +1,131 @@
...
+      customEC: Option[Executor] = None
```
amesgen: IMHO the default executor created for the `HttpClient` is a good default. I decided to take an `Executor` because the main way to create an `ExecutionContext` is to use `fromExecutor` or `fromExecutorService`, and either way it is very easy to convert between them.

jdk-http-client/src/main/scala/org/http4s/client/jdkhttpclient/JdkHttpClient.scala
```diff
@@ -61,51 +120,12 @@ object JdkHttpClient {
...
-        body = FlowAdapters
-          .toPublisher(res.body)
-          .toStream[F]
-          .flatMap(bs =>
-            Stream.fromIterator(bs.asScala.map(Chunk.byteBuffer).iterator).flatMap(Stream.chunk))
-      )
-    }
+  private def convertHttpVersionFromHttp4s[F[_]](version: HttpVersion)(
+      implicit F: ApplicativeError[F, Throwable]): F[HttpClient.Version] =
+    F.fromEither(version match {
```
rossabaker: You could do `F.pure` for the `Right`s and `F.raiseError` for the `Left` and skip the `fromEither`.

jdk-http-client/src/main/scala/org/http4s/client/jdkhttpclient/JdkHttpClient.scala
rossabaker: We have seen `SSLContext` throw before in blaze.

I'm not so sure this function carries its weight over just using their builder.  I guess we get `HttpClient.Version` parsing out of it?

amesgen: What about removing `sslContext` and `proxySelector`? Even without them, I think most uses are coverd by this.

rossabaker: I've never used `ProxySelector`.  `SSLContext` is something I very often customize (typically for mutual auth), but we can always use the builder.

Maybe the function to convert our HttpVersion to their version should be public?  It's probably the most irritating part of this constructor to reinvent if someone does need a custom context or selector.

amesgen: Ok, so remove `createUnderlyingClient` and make `convertHttpVersionFromHttp4s` public? What about `simple`? I am in favor of keeping it, as it is non-obvious that `HttpClient.newBuilder().build()` (or equivalently, `HttpClient.newHttpClient()`) is side-effectful. What about this:
```scala
def simple[F[_]](implicit F: ConcurrentEffect[F]): F[Client[F]] =
  F.delay(HttpClient.newHttpClient()).map(apply(_))
```

rossabaker:
:+1: Great work.

Because of the build complexity, and to practice having separate integration modules like we want to do in the future, we'd like to spin this off into a separate repo.  @ChristopherDavenport is going to set up a side repo, and I'll move the commits so you get proper credit, and we'll release it from there.

amesgen:
Nice! But I have no problem with the commits being squashed / "losing" commit attribution.

rossabaker:
I'll probably squash them, but you'll get the glory. :smile:

amesgen:
I can also close this PR, create a new one only adding the `ConcurrentFuture` conversion, and another one in [http4s/http4s-jdk-http-client](https://github.com/http4s/http4s-jdk-http-client) with the rest. That might be easier than Git magic. ;)

rossabaker:
I didn't want to add more work for our prior indecision, but if you're offering... great! :smile:

PR #2493, opened 2019-04-06
Scala 2.13.0 support

rossabaker:
A long way from working, but we can start tracking what we need.  Many of these are blocked on their own dependencies:

* [x] better-monadic-for
* [x] cats
* [x] cats-effect
* [x] discipline
* [x] fs2
* [x] kind-projector
* [x] log4s
* [x] parboiled (our fork)
* [x] scalacheck
* [x] specs2
* [x] vault

PR #2503, opened 2019-04-11
kind-projector has moved to typelevel org

SethTisue:
perhaps there is some other in-flight PR with this change, but
regardless, it would helpful for the Scala community build to have
just this change land on master


aeons:
It looks like that version isn't released for 2.13.0-M5, which fails our build.

rossabaker:
This is included in #2493, which is likely to be stuck a while longer (the checklist there is just the first wave).  We could branch on Scala version.

SethTisue:
> We could branch on Scala version

okay I'll do that

rossabaker:
I think ddc304a will be sufficient for @SethTisue to unfork us in the community build.

PR #2506, opened 2019-04-11
Raise DecodeFailure with MonadError instead of throwing

### LINKED ISSUE #2504: `Message.as` uses throw which does not play well with ZIO
I wrote a http client using http4s and I used `Mesage[F].as[A]` to decode JSON responses. However when it fails, it uses `throw`

```scala
def as[T](implicit F: Functor[F], decoder: EntityDecoder[F, T]): F[T] =
    attemptAs.fold(throw _, identity)
```

Using `ApplicativeError.handleErrorWith` works with `cats.effect.IO` and Monix Task, but in ZIO it ends up in the `Platform.reportFailure`. I asked the folks of ZIO what to do here:
https://gitter.im/scalaz/scalaz-zio?at=5caf4e0a25686a7dc312e100

It works now wrapping the code inside `CatchDeffects` and have a `Task` instance which calls `Task.absorb` from ZIO.

@jdegoes proposes to fix this by either refactoring `as` or add a new method something like this

```scala
    def to[T](implicit F: MonadError[F, Throwable], decoder: EntityDecoder[F, T]): F[T] =
      message.attemptAs.value.flatMap(_.fold[F[T]](F.raiseError, F.pure))
```

rossabaker:
Fixes #2504. Similar to #2502.

This code was written before the effect was polymorphic, and relies on now undefined behavior.  The `MonadError` constraint lets us raise the error into `F` lawfully.  Optimization coming in cats is noted. This rippled to the oauth1 client.

### DIFF
--- a/client/src/main/scala/org/http4s/client/oauth1/oauth1.scala
+++ b/client/src/main/scala/org/http4s/client/oauth1/oauth1.scala
@@ -31,7 +31,9 @@ package object oauth1 {
       consumer: Consumer,
       callback: Option[Uri],
       verifier: Option[String],
-      token: Option[Token])(implicit F: Monad[F], W: EntityDecoder[F, UrlForm]): F[Request[F]] =
+      token: Option[Token])(
+      implicit F: MonadError[F, Throwable],
+      W: EntityDecoder[F, UrlForm]): F[Request[F]] =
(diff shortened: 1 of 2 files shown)


Fristi:
Thanks 👍 that's real quick :-)

PR #2508, opened 2019-04-12
add MediaType literal parsing

bpholt:
https://gitter.im/http4s/http4s?at=5caeb05293fb4a7dc20c2646

### DIFF
--- a/core/src/main/scala/org/http4s/Uri.scala
+++ b/core/src/main/scala/org/http4s/Uri.scala
@@ -375,6 +375,7 @@ object Uri {
     * Literal syntax for URIs.  Invalid or non-literal arguments are rejected
     * at compile time.
     */
+  @deprecated("""use uri"" string interpolation instead""", "0.20")
   def uri(s: String): Uri = macro Uri.Macros.uriLiteral
 
   implicit val http4sUriEq: Eq[Uri] = Eq.fromUniversalEquals
--- a/core/src/main/scala/org/http4s/package.scala
+++ b/core/src/main/scala/org/http4s/package.scala
@@ -100,4 +100,11 @@ package object http4s { // scalastyle:ignore
       case Left(l) => f(l)
     }
   }
+
+  final implicit class Http4sLiteralSyntax(val sc: StringContext) extends AnyVal {
+    def uri(args: Any*): Uri = macro LiteralSyntaxMacros.uriInterpolator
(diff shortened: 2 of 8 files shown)


bpholt:
Good idea to use `StringContext`. I added them for `Uri`, `MediaType`, and `QValue`. (The `QValue` one uses `qValue""` as the syntax instead of just `q""` since the latter is well-known for quasiquotes.) Let me know what you think.

One of the blaze-client tests is failing locally for me, but I'm not sure it's related to my changes. We'll see what happens in CI.

rossabaker:
Which one failed locally?  Local failures of those are my holy grail.

core/src/main/scala/org/http4s/Uri.scala
```diff
@@ -375,6 +375,7 @@ object Uri {
     * Literal syntax for URIs.  Invalid or non-literal arguments are rejected
     * at compile time.
     */
+  @deprecated("""use uri"" string interpolation instead""", "0.20")
```
rossabaker: As discussed, this one is going to cause the most migration pain, but it's like what we'll be doing in the long run.

core/src/main/scala/org/http4s/package.scala
```diff
@@ -100,4 +100,11 @@ package object http4s { // scalastyle:ignore
       case Left(l) => f(l)
     }
   }
+
+  final implicit class Http4sLiteralSyntax(val sc: StringContext) extends AnyVal {
```
rossabaker: Is this "syntax"?  Should it be imported through `org.http4s.implicits._` rather than be part of the package object?  My gut feeling is yes, but ip4s does it at the top level.

@mpilquist, is this a design you considered and rejected?

bpholt:
> Which one failed locally? Local failures of those are my holy grail.

```
[error]   x behave and not deadlock on failures with parSequence (30 seconds, 207 ms)
[error]    None is not Some (BlazeClientSpec.scala:189)
[error] org.http4s.client.blaze.BlazeClientSpec.$anonfun$new$31(BlazeClientSpec.scala:189)
```

rossabaker:
If that's consistent for you, I'd love to hear more about your machine.

bpholt:
That test failed probably 3-4 times in a row for me the other day, but it's passing now. 🤷‍♂️

rossabaker:
Any second reviewer before I merge?

Issue #2423, opened 2019-02-16
Is there a way to provide different buckets for Prometheus histogram?


mr-git:
In Prometheus, default histogram is defined as:
```java
private double[] buckets = new double[]{.005, .01, .025, .05, .075, .1, .25, .5, .75, 1, 2.5, 5, 7.5, 10};
```
It is used in `org/http4s/metrics/prometheus/Prometheus.scala:157` to measure response times.

The default bucket has good steps at start, but at is quite useless at end with resolution of 2.5s.

It would be nice, if there would be a way to set own bucket list, but I didn't find such.
Maybe it would be good idea to use custom bucket list, like:
```scala
responseDuration = Histogram
  .build()
  .buckets(0.005, 0.01, 0.02, 0.04, 0.08, 0.1, 0.2, 0.4, 0.8, 1, 1.2, 1.4, 1.6, 1.8, 2, 2.2, 2.4, 2.6, 2.8, 3, 3.2, 3.4, 3.6, 3.8, 4, 4.2, 4.4, 4.6, 4.8, 5, 5.2, 5.4, 5.6, 5.8, 6, 6.2, 6.4, 6.6, 6.8, 7, 7.2, 7.4, 7.6, 7.8, 8, 8.2, 8.4, 8.6, 8.8, 9, 9.2, 9.4, 9.6, 9.8, 10)
  //...
```

It has more than 4x buckets and has a bit better resolution on "fast" end and much better one on "slow" end, resolution is closer to linear - this can be useful in cases, when generated response depends on slow backend system(s).
It could be even useful to add `0.5` steps up to `20` seconds.

kevinmeredith:
The `Histogram` class comes from https://github.com/prometheus/client_java/blob/master/simpleclient/src/main/java/io/prometheus/client/Histogram.java.

As a result, does this question qualify as an issue? In other words, might it be better to pose this question in StackOverflow?

mr-git:
Current `http4s` API doesn't expose this ability and uses default bucket, which is quite shallow and wide...
Options:
* use the buckets as I mentioned in main
* close this as "not an issue" and forget about it
* complicate current API and expose ability to provide custom bucket list

mr-git:
Default one is not very nice, when working with external systems with SLAs around 3..4 seconds - there are 2.5s and 5s buckets and "acceptable" calls are stored in "bad" bucket :)
2.5s wide steps are not very nice, and with heat-map panels in recent Grafana they look bad too

[...]

PR #2510, opened 2019-04-13
Add optional argument to override Histogram's buckets.

kevinmeredith:
Closes https://github.com/http4s/http4s/issues/2423

Note that I chose NonEmptyList[Double] instead of List[Double] due to:

```scala
$amm
Loading...
Welcome to the Ammonite Repl 1.0.3
(Scala 2.12.4 Java 1.8.0_112)
If you like Ammonite, please support our development at www.patreon.com/lihaoyi
@ import $ivy.`io.prometheus:simpleclient:0.5.0`
import $ivy.$

@ import io.prometheus.client.{Histogram, CollectorRegistry}
import io.prometheus.client.{Histogram, CollectorRegistry}

@ val c2 = new CollectorRegistry()
c2: CollectorRegistry = io.prometheus.client.CollectorRegistry@351d93bd

@ Histogram.build().buckets(List.empty[Double] : _*).name("foo").help("help").labelNames("bar").register(c2)
java.lang.IllegalStateException: Histogram must have at least one bucket.
```

### DIFF
--- a/prometheus-server-metrics/src/main/scala/org/http4s/server/prometheus/PrometheusMetrics.scala
+++ b/prometheus-server-metrics/src/main/scala/org/http4s/server/prometheus/PrometheusMetrics.scala
@@ -184,6 +184,10 @@ object PrometheusMetrics {
       }
     } yield ()
 
+  // https://github.com/prometheus/client_java/blob/parent-0.5.0/simpleclient/src/main/java/io/prometheus/client/Histogram.java#L73
+  private val DefaultHistogramBuckets: NonEmptyList[Double] =
+    NonEmptyList(.005, List(.01, .025, .05, .075, .1, .25, .5, .75, 1, 2.5, 5, 7.5, 10))
+
   /**
     * Metrics --
     *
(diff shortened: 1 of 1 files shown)


rossabaker:
This is going to break binary compatibility.  Can we just do this on master instead?

kevinmeredith:
> This is going to break binary compatibility. 

Does the following scenario

App A uses the `foo` library, which depends on `http4s`'s `0.18.23`. App A also uses the `bar` library. `bar` depends on `http4s`'s `0.18.24`, i.e. the version, let's say, released if this PR were merged. 

Binary compatibility would be broken for any of `foo`'s calls to `PrometheusMetrics#apply` given the change in method signature?

Please note that I'm asking for my own learning, not challenging your statement, @rossabaker!

>  Can we just do this on master instead?

Sure - I'll open another PR. By opening a PR against `master`, that means that, if approved, it'll be added to `0.20.x`? I'm asking only for my own curiosity. I'm assuming that merged PRs to `master` only show up on the latest `http4s` version, i.e. `0.20.x`, as I understand.

rossabaker:
Right. Or if App A directly depends on 0.18.24. There doesn't need to be a bar. If you call the method in foo that compiled against the apply in 0.18.23, you'd get a runtime error.  We can add signatures, but not remove them. In this case, you can add this as a name other than apply.

master is 0.20.x, which is about to go final. Hurry. :smile:

PR #2515, opened 2019-04-16
Support for additional attributes on Jetty backend

CremboC:
This allows to set arbitrary parameters on the Jetty backend. Ideally, it would be a strongly typed map i.e. `Map[JettyPropertyName, AnyRef]` or perhaps even `List[JettyProperty]`, but I think this is fine for starters.

### DIFF
--- a/jetty/src/main/scala/org/http4s/server/jetty/JettyBuilder.scala
+++ b/jetty/src/main/scala/org/http4s/server/jetty/JettyBuilder.scala
@@ -14,6 +14,7 @@ import org.eclipse.jetty.servlet.{FilterHolder, ServletContextHandler, ServletHo
 import org.eclipse.jetty.util.component.{AbstractLifeCycle, LifeCycle}
 import org.eclipse.jetty.util.ssl.SslContextFactory
 import org.eclipse.jetty.util.thread.{QueuedThreadPool, ThreadPool}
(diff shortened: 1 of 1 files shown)


CremboC:
Sorry, when I said `JettyPropertyName` I meant something like the following:
```scala
sealed abstract class JettyServletContextProperty(name: String)
case object MaxFormContentSize extends JettyServletContextProperty("maxFormContentSize")
case object MaxFormKeys extends JettyServletContextProperty("maxFormKeys")
```
etc.

CremboC:
Also, I am looking into adding a test to see that this actually works.. Unfortunately there are (at least) two places where you can `setAttribute(s)` and I'm looking into the difference between them. I believe tests will help figure out which is the one we want.

Alternatively, I may expose the specific property e.g. `setMaxFormContentSize` instead of attributes - perhaps that will be clearer...

rossabaker:
The nature of these attributes is that they accrue over time.  To seal it, we'd need a list, and then to stay in sync with Jetty releases, we'd have binary compatibility pains.  I like the instinct toward avoiding String types, but I think it's the best we can reasonably do here.

armanbilge:
If interested, this PR should be retargeted to https://github.com/http4s/http4s-jetty.

PR #2516, opened 2019-04-16
Drop redundant enable web sockets

mcanlas:
The current `BlazeServerBuilder` enables web sockets by default. Should that really be the case? I feel like opting-in for features is a better interface.

If not, then the following line can be elided.


mcanlas:
Per you original comment, how costly is the extra WS stage vs having WS users add one line?

I would guess WS users are a minority.

Issue #2483, opened 2019-04-02
Client API feels inconsistent


Lasering:
There are four `expect` methods:
```
def expect[A](s: String)(implicit d: EntityDecoder[F, A]): F[A]
def expect[A](uri: Uri)(implicit d: EntityDecoder[F, A]): F[A]
def expect[A](req: F[Request[F]])(implicit d: EntityDecoder[F, A]): F[A]
def expect[A](req: Request[F])(implicit d: EntityDecoder[F, A]): F[A]
```

1) The first two work like special cases where the request is a GET. At the same time, there are two methods called `get` with the exact same parameters as the first two `expect`s. The same is happening for the `expectOr` methods.

2) There is the method:
```
def expectOption[A](req: Request[F])(implicit d: EntityDecoder[F, A]): F[Option[A]]
```
But there **isn't** the corresponding one for `F[Request[F]]`:
```
def expectOption[A](req: F[Request[F]])(implicit d: EntityDecoder[F, A]): F[Option[A]]
```

3) Tha API has a dedicated DSL for GET requests but not for other commonly used methods: DELETE, PUT, POST, HEAD.

rossabaker:
I agree.  This is an artifact of us starting with a broad set of overloads, and then backing away from that approach as we added more functions.  We should fully embrace the overloads or deprecate the other ones.

I'm open to argumentation for a smaller or larger API, but strongly in favor of consistency before 1.0.  What does everyone think?

ChristopherDavenport:
Remove/Deprecate all overloads.

kevinmeredith:
Would it be worthwhile to deprecate/remove `def expect[A](s: String)(implicit d: EntityDecoder[F, A]): F[A]`, i.e. require a valid `Uri` to be passed rather than a `String`?

`Uri` > `String` motivates this question.

kubukoz:
Strong +1 on removing all methods that take String.

While we're at it, I'd make `uri` the first argument in `Request`'s `apply` method, although that's probalby a different issue...

rossabaker:
I think moving `Uri` within `Request.apply` becomes more plausible if it ceases to be a case class, as discussed in #2537.  Would still like to do that in a non-breaking way.

Are we agreed to deprecate the overloads for 0.21 then?

kevinmeredith:
> Are we agreed to deprecate the overloads for 0.21 then?

[...]

PR #2518, opened 2019-04-18
Add status Methods Which Use `String` or `URI`

isomarcte:
These are similar to the `expect` methods which take the same arguments. They are particularly useful for test.

### DIFF
--- a/client/src/main/scala/org/http4s/client/Client.scala
+++ b/client/src/main/scala/org/http4s/client/Client.scala
@@ -141,6 +141,12 @@ trait Client[F[_]] {
   /** Submits a request and returns the response status */
   def status(req: F[Request[F]]): F[Status]
 
+  /** Submits a GET request to the URI and returns the response status */
+  def statusFromUri(uri: Uri): F[Status]
+
+  /** Submits a GET request to the URI and returns the response status */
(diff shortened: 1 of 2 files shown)


isomarcte:
I don't think the travis failure is related to my change.

rossabaker:
Thanks.  I agree on the CI and restarted Travis.

Whether we want to do this, or move in the opposite direction, is #2483.  We should agree on our strategy there, and then either approve or close this.

rossabaker:
If we do these, then also `expectOption`, `expectOptionOr`, etc., right?  I guess I'm arguing for "all-or-none" rather than "more".

rossabaker:
It's a lot easier to add methods than take them away, and we have a general aversion to overloads.  I think this is a symptom of friction in creating requests.  I'd like to think about how we can do better there.

I vote we think about this during 0.20, since 0.21 is likely just around the corner.

isomarcte:
@rossabaker any status updates on this?

rossabaker:
I hoped someone would come up with a simpler way of building requests that reduces the desire for overloads, but we haven't seen one yet.  I'm usually opposed to overloads, but I guess I'm a mild :+1: on this, since someone cared enough to add it, and nobody has cared enough to shout it down.

PR #2520, opened 2019-04-19
Parameterize selectorThreadFactory for blaze server

rossabaker:
Inspired by [Twitter conversation](https://twitter.com/djspiewak/status/1116371300765847552) with @djspiewak.  I'm not changing the default at this stage in the release cycle, but making it possible for Daniel (or others) to get empirical evidence that it's the right thing to do.  We can tweak the default in 0.20.1 or later, but this needs to go in now.

Pending release of https://github.com/http4s/blaze/pull/282 before it will compile.


rossabaker:
Build won't pass if blaze-0.14.0 isn't synced yet, but we can restart.

PR #2524, opened 2019-04-21
Fixed laws-check in RefererSpec.

heksesang:


aeons:
Looks like it needs an import to build.

rossabaker:
We discussed on Gitter, and I didn't take time to look at Travis. 😬

rossabaker:
This is actually kind of tricky.  It needs an `Arbitrary[Referer]` (which I went ahead an added), which comes from an `Arbitrary[Uri]`, which doesn't round trip, as noted in https://github.com/http4s/http4s/issues/1651.  Now it should compile but the test is going to fail until we fix that underlying issue with the Uri.

armanbilge:
Fork is gone.

PR #2527, opened 2019-04-23
Kill Kleisli

diesalbla:
This is a line of work span off a [comment](https://github.com/typelevel/cats/pull/2801#issuecomment-485488486)  to @rossabaker in a recent PR submitted to `cats`.

That PR deals about an alternative to `Kleisli`. The key question asked to Ross, in that context, was about the benefits that `Kleisli` gives for the developers of `http4s`.

This Draft PR is an exploration, intended to see how far we can get rid of `Kleisli` in `http4s` without damaging the code. It is a work in progress.

### DIFF
--- a/client/src/main/scala/org/http4s/client/Client.scala
+++ b/client/src/main/scala/org/http4s/client/Client.scala
@@ -54,7 +54,9 @@ trait Client[F[_]] {
     */
   def toKleisli[A](f: Response[F] => F[A]): Kleisli[F, Request[F], A]
 
-  @deprecated("Use toKleisli", "0.18")
+  def toFunction[A](f: Response[F] => F[A]): Request[F] => F[A]
+
+  @deprecated("Use toFunction", "0.18")
   def toService[A](f: Response[F] => F[A]): Service[F, Request[F], A]
 
   /**
--- a/client/src/main/scala/org/http4s/client/DefaultClient.scala
+++ b/client/src/main/scala/org/http4s/client/DefaultClient.scala
@@ -46,9 +46,11 @@ private[client] abstract class DefaultClient[F[_]](implicit F: Bracket[F, Throwa
   def toKleisli[A](f: Response[F] => F[A]): Kleisli[F, Request[F], A] =
     Kleisli(fetch(_)(f))
 
+  def toFunction[A](f:Response[F] => F[A]): Request[F] => F[A] = req => fetch(req)(f)
+
   @deprecated("Use toKleisli", "0.18")
   def toService[A](f: Response[F] => F[A]): Service[F, Request[F], A] =
-    toKleisli(f)
+    toFunction(f)
 
   /**
     * Returns this client as an [[HttpApp]].  It is the responsibility of
@@ -59,12 +61,11 @@ private[client] abstract class DefaultClient[F[_]](implicit F: Bracket[F, Throwa
     * [[toKleisli]], and [[streaming]] are safer alternatives, as their
     * signatures guarantee disposal of the HTTP connection.
     */
-  def toHttpApp: HttpApp[F] = Kleisli { req =>
+  def toHttpApp: HttpApp[F] = req =>
     run(req).allocated.map {
       case (resp, release) =>
         resp.withBodyStream(resp.body.onFinalize(release))
     }
-  }
 
   /**
     * Returns this client as an [[HttpService]].  It is the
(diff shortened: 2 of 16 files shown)


rossabaker:
The final Kleisli is generally paid for once, as we define the function and then perhaps wrap it in a few middlewares.  Each of these steps unwraps, composes, and rewraps the `.run` function.  But the intermediate Kleislis are discarded.  Furthermore, the backends store the mounted app's `.run` in a val as a microoptimization (we measured a long time ago: it's debatable whether it even mattered), so we don't even pay to unwrap it per request.

The question is then not one of performance, but of the richness of the methods to compose the `Http` behavior.  I think `Kleisli` wins hands down here.

diesalbla:
@rossabaker The changes so far seem to indicate that most of the code does not need those Kleisli combinators. _Services as functions_ seems to me a far simpler idea that _servers as Kleislis_.  
I am rather inclined to think that using plain functions is going to be almost just as short and probably far more readable, since some of those `Kleisli` methods can be a bit obscure.

If you want to use the methods of `Kleisli[F, A, B]`, you are very likely to get the same from an `object` with a few functions on `A => F[B]`. The `Kleisli` class, as you said, only gives you the convenience of extended syntax. This does not stop any user of the library from using Kleisli, since nothing stops her from adding the wrapping on their side.

rossabaker:
We also benefit from Kleisli's abstraction over the effect: the same middleware can apply to kleislis in `F` (`HttpApp`), `OptionT[F, ?]` (`HttpRoutes`), or other monads (I gave a talk on some of these at Typelevel this year.)

When we use a `Function`, we run into difficulties with `PartialFunction` being a subtype of it, which is challenging for something like http4s-dsl.  Many of the breaking changes up through about 0.18 were a result of iterating on functions vs. partial functions, along with various unsafe or unprincipled sentinel values for composition.  Now we can compose vertically (middleware) or horizontally (combineK), with clarity in the types.  The current design has withstood the test of time far better than several previous attempts at functions.

diesalbla:
> We also benefit from Kleisli's abstraction over the effect: the same middleware can apply to kleislis in F (HttpApp), OptionT[F, ?] (HttpRoutes), or other monads 

I am afraid I cannot see how that would be different from having a function from one function to another. Do you have an example of that?

> When we use a Function, we run into difficulties with PartialFunction being a subtype of it, which is challenging for something like `http4s-dsl`. 

Could you describe a bit more those difficulties? Like a partial function literal introducing unexpected match errors, perhaps? 

One more question: does the variance of the type-parameters causes difficulties when using the `Function`? I am asking because the one difference in the type level is that `Kleisli` is invariant on its three type parameters.

rossabaker:
> I am afraid I cannot see how that would be different from having a function from one function to another. Do you have an example of that?

I withdraw this argument: what you've done to the middleware shows that it's possible to support an `Http` alias that covers both `HttpRoutes` and `HttpApp`.  I do think you lost a lot of convenience along the way, though.

> Could you describe a bit more those difficulties? Like a partial function literal introducing unexpected match errors, perhaps?

Yes.  If we mount `A => F[B]` to our server, it's easy to get a `MatchError` from a `PartialFunction` literal, which is the heart of http4s-dsl.  If we make `PartialFunction[A, F[B]]` the common definition, we get a gentle reminder, but still get a `MatchError` if we forget to `applyOrElse`, and bias the core heavily in favor of http4s-dsl.  We could keep the `HttpRoutes.of` function to convert a partial function into a total function, but with what default response?  404?  But when we want to combine services, how then do we tell an explicit 404 from a default 404?  Reference equality with a sentinel value?  A sum type of `Pass | Response`?  A response attribute?  A `Fallthrough` typeclass so it works for multiple outputs?  Or we can lift and sequence the partial function into an `A => F[Option[B]]`, but the wrapper that does it is no less noisy than the Kleisli.   We've been through that entire mess, and `Kleisli` is the first lasting design.

> One more question: does the variance of the type-parameters causes difficulties when using the Function? I am asking because the one difference in the type level is that Kleisli is invariant on its three type parameters (the first one being invariant inside and outside).

I don't remember the variance causing any particular problems.

diesalbla:
@rossabaker Thank you for your explanation. From what you write, clearly you did tried as long as you could to stick with `Function1` and `PartialFunction`. Is it accurate and fair to infer, then, that you ended up using `Kleisli` not so much because of its benefits, but to a greater extent because of the defects of the  `FunctionN` and `PartialFunction` classes from the standard Scala library?

rossabaker:
Both.  We can make the same mistake with Klesili that we make with stdlib: you can still pass a partial function to `Kleisli.apply` or `HttpApp.apply`.  In practice, I made and saw that a lot with raw functions and not once since we introduced `HttpApp` and `HttpRoutes`.  I think the fact that these Kleisli forms don't have a subtype relationship makes people more thoughtful, even though it can still be abused in roughly the same way.

And since function output is effectful, I just find the Kleisli operations more convenient than the Function1 operations, at what should be zero cost at runtime when they're constructed once and mounted to a singular backend.

diesalbla:
From what @rossabaker comments above, using the `Function` type from the standard Scala library was tried a while ago, and did not work as well. We can close this.

PR #2530, opened 2019-04-26
Authentication: simplify code.

diesalbla:
We extract an "unauthorized" function, that generates a response with the "Unatuthorized" status with the given authentication challenge.

We edit the code of challenged, so as to avoid using the `Kleisli` and `OptionT` combinators, which can be unclear, and replace them with a slightly more transparent style.

### DIFF
--- a/server/src/main/scala/org/http4s/server/middleware/authentication/package.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/authentication/package.scala
@@ -3,7 +3,7 @@ package server
 package middleware
 
 import cats.data.{Kleisli, OptionT}
-import cats.effect._
+import cats.effect.Sync
 import org.http4s.headers._
 
 package object authentication {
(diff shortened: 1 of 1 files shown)


server/src/main/scala/org/http4s/server/middleware/authentication/package.scala
```diff
@@ -11,14 +11,14 @@ package object authentication {
...
-            OptionT.some(Response(Status.Unauthorized).putHeaders(`WWW-Authenticate`(challenge)))
+      OptionT[F, Response[F]] {
+        F.flatMap(challenge(req)) {
+          case Left(challenge) => F.pure(Some(unauthorized(challenge)))
+          case Right(authedRequest) => routes(authedRequest).value
         }
+      }
     }
+
+  def unauthorized[F[_]](challenge: Challenge): Response[F] =
```
rossabaker: I would make this private so we don't have to preserve it across future refactoring.

PR #2535, opened 2019-04-29
Add translate extension methods for HttpRoutes and HttpApp

bplommer:
Adds `imapK` extension methods to the `Kleisli` syntax module to change the effect type of `HttpRoutes` and `HttpApp`.

### DIFF
--- a/core/src/main/scala/org/http4s/syntax/AllSyntax.scala
+++ b/core/src/main/scala/org/http4s/syntax/AllSyntax.scala
@@ -1,6 +1,10 @@
 package org.http4s
 package syntax
 
+abstract class AllSyntaxBinCompat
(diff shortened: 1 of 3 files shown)


bplommer:
> Would `imapHttpApp` be too weird a name?
It feels a little wordy to me. How about `translate`, like in fs2 `Stream`?

rossabaker:
Yes, I like `translate`!

rossabaker:
After all that bincompat work, this could go on 0.20.  If it passes CI there, I'll merge there.  If it doesn't, I'll switch back and merge to matser.

rossabaker:
Ah, no need to rebuild even. :tada:

PR #2539, opened 2019-05-01
Narrow Imports from Cats in Core package

diesalbla:
- We remove every wild-card import from the `cats`, `cats.data`, or `cats.effect` packages.
- We try to narrow down imports of `cats.implicits._` to one or two specific imports from `cats.syntax` or `cats.instances`.

### DIFF
--- a/blaze-server/src/main/scala/org/http4s/server/blaze/BlazeServerBuilder.scala
+++ b/blaze-server/src/main/scala/org/http4s/server/blaze/BlazeServerBuilder.scala
@@ -2,10 +2,10 @@ package org.http4s
 package server
 package blaze
 
-import cats._
+import cats.{Alternative, Applicative}
 import cats.data.Kleisli
 import cats.implicits._
-import cats.effect._
+import cats.effect.{ConcurrentEffect, Resource, Timer}
 import java.io.FileInputStream
 import java.net.InetSocketAddress
 import java.nio.ByteBuffer
--- a/circe/src/main/scala/org/http4s/circe/CirceInstances.scala
+++ b/circe/src/main/scala/org/http4s/circe/CirceInstances.scala
@@ -3,9 +3,9 @@ package circe
 
 import java.nio.ByteBuffer
 
-import cats._
+import cats.Applicative
 import cats.data.NonEmptyList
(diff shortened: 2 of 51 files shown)


rossabaker:
Our [contributors' guide](https://http4s.org/contributing/) speaks a little about imports, but not to these particular issues.  We enforce scalafmt, but scalafmt doesn't have a lot to say about imports, which leads to this being one of the more variant areas of style.  We could tighten up the recommendations.

1) I am mildly :+1: on explicit imports from cats, cats.data, etc.  Especially as tooling has gotten better to make this easier.

2) I am more :-1: on a la carte imports for instances and syntax.  The names still aren't explicit.  Nor should they be, because they have intentionally long names to avoid accidental shadowing.  Finding the right objects is a barrier for new users and a chore for experienced users.  And despite claims of faster compilation, I've never seen a signficant difference in practice.

diesalbla:
>  Finding the right objects is a barrier for new users and a chore for experienced users.

@rossabaker I can personally understand the desire to use `cats.implicits._` import. I have worked with some Scala applications where compilation times were over a minute long, and I can relate that it is frustrating to lose a minute because of an inaccurate `cats` import. 

However, since `http4s` is not an application but a library, I would say that here the best criteria is to reduce the number of `cats` imports, which is to say, to constrain the capabilities each part of the code needs. 

The code in the PR is already looking for a compromise: at some point, where there are four or five syntax or instances imports, a `syntax.all._` or an `.instances.all._` is used.

rossabaker:
My experience is that these imports don't substantially affect compile time.  (Imports in other libraries that derive instances, for sure. These imports, no.)  I'd want to see a very striking difference on this project to change my mind.  Maintainer time is far more precious than CPU time, and I find narrowing implicit imports to be a chore.  And mixing a la carte imports with regular imports has come up a couple times on Gitter recently, and I really don't want to encourage a style that raises the barrier to new contributors, given how uninformative the compilation failures are.

I still like eliminating wild cards on things named in the code.

diesalbla:
@rossabaker Ok. I will apply the corrections later today.

rossabaker:
:+1: after conflict resolution

PR #2540, opened 2019-05-01
CSRF: Make plain

diesalbla:
Changes the code of CSRF methods to replace the use of the "EitherT" methods in favour of pattern-matching on the types from the Scala library. We extract a common code for these two methods as a private method.

### DIFF
--- a/server/src/main/scala/org/http4s/server/middleware/CSRF.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/CSRF.scala
@@ -110,13 +110,7 @@ final class CSRF[F[_], G[_]] private[middleware] (
     */
   def refreshedToken[M[_]](r: Request[G])(
       implicit F: Sync[M]): EitherT[M, CSRFCheckFailed, CSRFToken] =
-    CSRF.cookieFromHeaders(r, cookieSettings.cookieName) match {
-      case Some(c) =>
-        EitherT(F.pure(extractRaw(c.content)))
-          .semiflatMap(signToken[M])
-      case None =>
-        EitherT(F.pure(Left(CSRFCheckFailed)))
-    }
+    EitherT(refreshOpt(r).getOrElse(F.pure(Left(CSRFCheckFailed))))
 
   /** Extract a `CsrfToken`, if present, from the request,
     * then try generate a new token signature, or fail with a validation error.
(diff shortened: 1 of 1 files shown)


diesalbla:
@jmcardon The main goal is to avoid introducing the`EitherT` wrapper class until the top of the function, where it needs to. Thus, the new implementation does not rely on any library logic, other than that of the Scala standard library.

jmcardon:
@diesalbla what do you mean "avoiding library logic"? 

EitherT is literally just an `AnyVal` wrapper over `f (Either a b)`. There's no library logic there.

Moreover, you're using `Sync` in the signature of the other function you made, which is also tied to the stl.

I don't see a win between what's on master and `Option[M[Either[CSRFCheckFailed, CSRFToken]]]`, with `M` being `Sync` bound.

diesalbla:
> Being only tied to the stl isn't really a goal here, because there's no security sensitive part of the logic that cats is getting wrong at all.

It is not about security. It is fore more simple than that: it is about reducing what people need to understand just to read this code, and I prefer it if people does not need to know about  `cats.data.EitherT`, just to understand what this code is doing. The less extra stuff needed to understand it, the more people that will be able to read it quickly.

jmcardon:
>  I prefer it if people does not need to know about cats.data.EitherT, at all, just to understand what this code is doing

Knowing about `EitherT` is not too much to ask, the way say knowing about `Sync` (which is far more complicated if you consider how you start having to think about which places may suspend a side effect) isn't too much to ask for anyone working on http4s.

I don't think I agree with your reasoning behind this PR. It makes me think you're applying some personal  stylistic choice of yours and rewriting a bunch of code that _you_ believe reads better in a bunch of places, not just this PR.

The thing is the entirety of http4s relies on cats and cats-effect, meaning that cutting out a method you can easily look up the signature for in various places to swap out a pattern match or something does nothing to readability. Using `semiFlatMap` or not doesn't make a snippet more or less opaque, and it's reasonable to use such methods when, within this style of scala programming, cats is almost an stl of sorts for FP. 

I can get behind a PR if it's a performance improvement, a fix or some added functionality no problem.If http4s aimed to be cats agnostic or something, or there were some api deprecations, I'd easily get behind a rewrite too. I think where I find the issue is a repeated application of a personal criterion to already working, readable (to most others working with that code) code for no other reason than your own personal preference.

diesalbla:
> Knowing about EitherT is not too much to ask, the way say knowing about Sync (which is far more complicated if you consider how you start having to think about which places may suspend a side effect) isn't too much to ask for anyone working on http4s.

You are mixing different concerns. `Sync`, the capability to run suspended effectful computations, is **essential** for a web server. `EitherT`, an extra wrapper class for some specific mixes of `Monad` and `Either` operations, is just a programmer's tool, and should only be evaluated on how it helps the code. 

As for readability, just consider the following two pieces of code:  
```Scala
         EitherT(F.pure(extractRaw(c.content)))
          .semiflatMap(signToken[M])
```
Against the modified version: 
```Scala
       extractRaw(c.content) match {
        case Right(raw) => signToken[M](raw).map(Right(_))
```
Both pieces of code say the same: if the result of `extractRaw` is `Right`, then apply to its content the  `signToken` function. Except the second one is using the mechanisms of Scala itself (pattern matching) while the former introduces it into an `F` context and into an `EitherT`, only for the sake of using `semiflatMap`. 

Another problem is that the first one is using the `F` effect to describe what is a pure computation with in-memory values.

> I think where I find the issue is a repeated application of a personal criterion to already working, readable (to most others working with that code) code.

Readability does not mean that experts understand you. It means that non-experts do.

PR #2541, opened 2019-05-02
StaticFile / fromResource: plain code

diesalbla:
- We rewrite the code of the fromResource method, so as to rely less on
  the `OptionT` wrapper methods.
- We avoid introducing the `OptionT` wrapper until the end, and use for
  most of the operations the methods on `Option`.
- We extract a non-optionT method variant for fromUrl.

### DIFF
--- a/core/src/main/scala/org/http4s/StaticFile.scala
+++ b/core/src/main/scala/org/http4s/StaticFile.scala
@@ -36,22 +36,21 @@ object StaticFile {
           ContentCoding.`x-gzip`)
     }
 
-    val gzUrl: OptionT[F, URL] =
-      if (tryGzipped) OptionT.fromOption(Option(getClass.getResource(name + ".gz")))
-      else OptionT.none
-
-    gzUrl
-      .flatMap { url =>
+    def gzipped: Option[F[Response[F]]] =
+      Option(getClass.getResource(name + ".gz")).map { url =>
         // Guess content type from the name without ".gz"
         val contentType = nameToContentType(name)
         val headers = `Content-Encoding`(ContentCoding.gzip) :: contentType.toList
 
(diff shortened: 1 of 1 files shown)


core/src/main/scala/org/http4s/StaticFile.scala
```diff
@@ -36,22 +36,21 @@ object StaticFile {
...
     }
 
-    val gzUrl: OptionT[F, URL] =
-      if (tryGzipped) OptionT.fromOption(Option(getClass.getResource(name + ".gz")))
-      else OptionT.none
-
-    gzUrl
-      .flatMap { url =>
+    def gzipped: Option[F[Response[F]]] =
+      Option(getClass.getResource(name + ".gz")).map { url =>
```
rossabaker: `getResource` is an effect.  We didn't get this right before or now.

core/src/main/scala/org/http4s/StaticFile.scala
```diff
@@ -36,22 +36,21 @@ object StaticFile {
...
-        fromURL(url, blockingExecutionContext, req).map(
+        fromUrlF(url, blockingExecutionContext, req).map(
           _.removeHeader(`Content-Type`).putHeaders(headers: _*))
       }
-      .orElse(OptionT
-        .fromOption[F](Option(getClass.getResource(name)))
-        .flatMap(fromURL(_, blockingExecutionContext, req)))
+
+    def noGzipped: Option[F[Response[F]]] =
+      Option(getClass.getResource(name)).map(fromUrlF(_, blockingExecutionContext, req))
```
rossabaker: Another effect in pure code.

diesalbla:
Withdrawing until the code changes are tackled.

PR #2542, opened 2019-05-02
WebjarService: Simplify code

diesalbla:
- Replace OptionT methods with direct style (function calls) and
   pattern matching, using the `Option` type most of the time.
- Replace Option.map and Option.collect by pattern-matching on Option ADT.


diesalbla:
I have changed the code a little bit, to spread it vertically, and also to fix a non-exhaustive match.

PR #2546, opened 2019-05-02
Revert "Avoid closing the previously connection if no errors occurred."

michae1T:
This reverts commit 95ad428488381750cb7e9e30f1fa4811a16af90e.

===================

When sharing a client + connection pool among requests against
multiple hosts this commit causes a regression.

conditions:

- client calls A to a.com and B to b.com
- single shared connection pool size 5.

steps to reproduce:

- perform 10+ successful parallel requests A

final state:
```
org.http4s.client.PoolManager Returning idle connection to pool:
curAllocated=5 idleQueues.size=1 waitQueue.size=0 maxWaitQueueLimit=256
closed=false
```

- perform 1 request B

final state:
```
org.http4s.client.PoolManager No connections available for the desired
key, RequestKey(Scheme(https),xB.comx). Evicting random and
creating a new connection: curAllocated=5 idleQueues.size=1
waitQueue.size=0 maxWaitQueueLimit=256 closed=false
```

no further actions are then taken by the client and in our case on
a server:

```o.http4s.blazecore.IdleTimeoutStage Idle timeout after 30000 ms.```

from this point request B starts to work again (the eviction was
indeed performed).

Both with this revert and in 0.20-M6 prior the parallel connections
are released and the pool is left in a clean state for request B.

```
org.http4s.client.PoolManager Connection could not be recycled
, no pending requests. Shrinking pool: curAllocated=0 idleQueues.size=0
waitQueue.size=0 maxWaitQueueLimit=256 closed=false
````

Maintaining connections open is a good idea for performance in many
cases but it seems like the pool manager needs further work before
this behaviour can be changed.

### DIFF
--- a/blaze-client/src/main/scala/org/http4s/client/blaze/Http1Connection.scala
+++ b/blaze-client/src/main/scala/org/http4s/client/blaze/Http1Connection.scala
@@ -276,14 +276,13 @@ private final class Http1Connection[F[_]](
             cleanup()
             attributes -> rawBody
           } else {
-            attributes -> rawBody.onFinalizeCase {
-              case ExitCase.Completed =>
-                Async.shift(executionContext) *> F.delay { trailerCleanup(); cleanup(); }
-              case ExitCase.Error(_) | ExitCase.Canceled =>
-                Async.shift(executionContext) *> F.delay {
+            attributes -> rawBody.onFinalize(
+              Stream
+                .eval_(Async.shift(executionContext) *> F.delay {
                   trailerCleanup(); cleanup(); stageShutdown()
-                }
-            }
+                })
+                .compile
+                .drain)
           }
         }
         cb(


rossabaker:
Thanks.  Let's see if we can quickly fix this in a way that preserves connection reuse.  If none emerges, we'll merge this, but we do need to get this right ASAP.

aldiyen:
This isn't the intended behavior? What happens if you set `maxTotalConnections` to > 5 on your client, or reduce `maxConnectionsPerRequestKey` below the default value of 256?

aldiyen:
Also, I suspect the same behavior would occur even after this change is reverted in cases where the server isn't using `Transfer-Encoding: chunked`, as there is a different code path for whether or not to close connections in that case.

michae1T:
From the parameter names it seems like it might be possible to work around this case if the number of unique hosts is known.  

Yes, I'd say the contention is a side effect of the intended behaviour, however that the issue is that the eviction step used to recover is not carried out correctly.  

Seems like the original commit needed to go a bit farther to support connection reuse without introducing this problem.

aldiyen:
I see, yes. There's definitely a bug in PoolManager. It's putting the new request in the wait queue before closing the connection it's chosen to evict, and then it's never running the request after the connection shuts down... or the connection isn't shutting down for some reason.

I have a change that resolves the problem but I'd like to figure out if there's a deeper problem here before I make a PR to fix it.

aldiyen:
Created #2562 to address this problem.

rossabaker:
Added a unit test that I think simulates the reported reproduction and confirmed that #2562 fixes it.

PR #2548, opened 2019-05-03
Add translate method to Client [binary breaking]

bplommer:


client/src/main/scala/org/http4s/client/Client.scala
```diff
@@ -189,6 +190,15 @@ trait Client[F[_]] {
 
   @deprecated("Use expect", "0.14")
   def prepAs[T](req: F[Request[F]])(implicit d: EntityDecoder[F, T]): F[T]
+
+  /**
+    * Translates the effect type of this client from F to G
+    */
+  def imapK[G[_]: Defer: Bracket[?[_], Throwable]](fk: F ~> G)(gK: G ~> F)(
```
rossabaker: Hmm.  This shouldn't get a cats-tagless instance, which was the `imapK` concern on `Kleisli`, but I wonder whether we should just call this `translate` for consistency with the server.  Thoughts?

rossabaker: The `Defer` and `Bracket` constraint is interesting.  On server, it's `Sync`.  Is there anything that should have both these constraints that would not be `Sync`?

Some prior discussion of `Defer` in http4s: https://github.com/http4s/http4s/pull/2325

bplommer: Yeah, I think I agree - have changed it.

bplommer: Not that I know of. It seems like similar considerations apply here as in the discussion you linked, so I've changed it for consistency.

PR #2549, opened 2019-05-03
Use cats-effect 1.3.0 and remove workarounds introduced in #2470

RafalSumislawski:
- use cats-effect 1.3.0
- remove the workarounds for https://github.com/typelevel/cats-effect/issues/487 introduced in #2470 as the issue is fix in cats-effects 1.3.0 (https://github.com/typelevel/cats-effect/releases/tag/v1.3.0)

### DIFF
--- a/blaze-client/src/test/scala/org/http4s/client/blaze/ClientTimeoutSpec.scala
+++ b/blaze-client/src/test/scala/org/http4s/client/blaze/ClientTimeoutSpec.scala
@@ -178,7 +178,7 @@ class ClientTimeoutSpec extends Http4sSpec {
       // if the requestTimeout is hit then it's a TimeoutException
       // if establishing connection fails first then it's an IOException
 
-      // The expected behaviour is that the requestTimeout will happen first, but fetchAs will additionally wait for the IO.sleep(2.seconds) to complete.
+      // The expected behaviour is that the requestTimeout will happen first, but fetchAs will additionally wait for the IO.sleep(1000.millis) to complete.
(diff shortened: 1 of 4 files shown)


rossabaker:
This should go into series/0.20 instead of master.  master will be cats-2.0 and cats-effect-2.0.  But this otherwise looks good.  Thank you!

rossabaker:
Just needs a `test:scalafmt` and the build should pass.

blaze-client/src/test/scala/org/http4s/client/blaze/ClientTimeoutSpec.scala
```diff
@@ -178,7 +178,7 @@ class ClientTimeoutSpec extends Http4sSpec {
       // if the requestTimeout is hit then it's a TimeoutException
       // if establishing connection fails first then it's an IOException
 
-      // The expected behaviour is that the requestTimeout will happen first, but fetchAs will additionally wait for the IO.sleep(2.seconds) to complete.
+      // The expected behaviour is that the requestTimeout will happen first, but fetchAs will additionally wait for the IO.sleep(1000.millis) to complete.
```
kevinmeredith: So the `requestTimeout` isn't actually respected, i.e. given that `c.fetchAs[String](FooRequest).unsafeRunTimed(1500.millis)` fails after `1000 ms`, not `50 ms`, @RafalSumislawski?

kevinmeredith: > An acquire operation of cats' Bracket/Resource is uncancelable

is the answer to my question, per https://github.com/http4s/http4s/issues/2386#issuecomment-475782999?

PR #2551, opened 2019-05-04
AuthMiddleware / Apply: simplify code

diesalbla:
The `apply` method from the AuthMiddleware object bulds a service out of three elements: a function to get the auth info from a request, the authenticated service to run if it is authenticated, and the fallback service to run if authentication fails.

We simplify the code of this apply method by doing the following:
- We avoid the methods `local` and `mapF` of Kleisli.
- We avoid using the `Choice` instance for `Kleisli[OptionT]`. Instead we pattern-match right after the call to the `authUser` function. 
- We avoid calling the `AuthedRequest.apply(Kleisli)` method, a combinator. Instead, we build the `AuthedRequest` in each branch of the pattern-match. 

What is left is the code of the function in a direct style. 

This change is **BINARY-INCOMPATIBLE**, since it removes one parameter (the `Choice`).

### DIFF
--- a/server/src/main/scala/org/http4s/server/package.scala
+++ b/server/src/main/scala/org/http4s/server/package.scala
@@ -1,7 +1,6 @@
 package org.http4s
 
 import cats._
-import cats.arrow.Choice
 import cats.data.{Kleisli, OptionT}
 import cats.implicits._
 import cats.effect._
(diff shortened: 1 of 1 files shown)


server/src/main/scala/org/http4s/server/package.scala
```diff
@@ -105,17 +104,16 @@ package object server {
     def apply[F[_], Err, T](
         authUser: Kleisli[F, Request[F], Either[Err, T]],
```
diesalbla: @rossabaker Since this may change source/binary compatibility, perhaps we can reduce this to a `Request[T] => F[Either[Err, T]]`

alexknvl:
Perhaps this change can be made binary compatible by doing `val _ = C`. The unnecessary parameter could later be removed.

diesalbla:
@alexknvl Good idea. Done.

server/src/main/scala/org/http4s/server/package.scala
rossabaker: I'm kind of surprised we didn't make it `Kleisli[EitherT[F, Err, ?], Request[F], T]`, which I'm sure goes in the opposite direction that you prefer.  I also haven't used our auth and don't have a great feel for the ergonomics.

diesalbla:
@rossabaker I have reverted to the original incompatible change, without Alex suggestion, that removes the `Choice` implicit.

diesalbla:
@rossabaker I have reverted to the original incompatible change, without Alex suggestion, that removes the `Choice` implicit.

server/src/main/scala/org/http4s/server/package.scala
diesalbla: I suppose not, mostly because it is shorter and I do find it easier to read: 
 
```Scala
Request[F] => F[Either[Err, T]]

Kleisli[F, Request[F], Either[Err, T]]

Kleisli[EitherT[F, Err, ?], Request[F], T]
```

Anyway, that is a story for another time.

PR #2552, opened 2019-05-04
TimeOut Middleware apply: direct style.

diesalbla:
We replace the use of the special `Kleilsi` combinators `mapF` and `map` by
the use of a direct style, with a single Kleisli constructor, wrapping
a lambda (function literal) that contains the race logic.

### DIFF
--- a/server/src/main/scala/org/http4s/server/middleware/Timeout.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/Timeout.scala
@@ -21,9 +21,9 @@ object Timeout {
       http: Kleisli[F, A, Response[G]])(
       implicit F: Concurrent[F],
       T: Timer[F]): Kleisli[F, A, Response[G]] =
-    http
-      .mapF(respF => F.race(respF, T.sleep(timeout) *> timeoutResponse))
-      .map(_.merge)
+    Kleisli { (a: A) =>
+      F.race( http(a), T.sleep(timeout) *> timeoutResponse).map(_.merge)
+    }
 
   /** Transform the service to return a timeout response after the given
     * duration if the service has not yet responded.  If the timeout


SystemFw:
What's the motivation for this change?

diesalbla:
To make the code a bit more understandable, by using only the methods `race` and `map` of the underlying `F`, and not using any method of `Kleisli`.

jmcardon:
What about `mapF` is less understandable?

"Readability" is an incredibly arbitrary metric, especially in scala (Since it has so many styles). Within the FP style, using a few convenience methods from Kleisli is fairly normal, and I don't think using the kleisli api is significantly more complicated.

diesalbla:
> "Readability" is an incredibly arbitrary metric, especially in Scala

_Minimising unnecessary entities_ is a very good heuristic. In the current code you need to understand both `Kleisli.mapF` and `Kleisli.map`, which do nothing but mediate the parameters from one to the other; something the lambda literal does directly.

SystemFw:
> Minimising unnecessary entities is a very good heuristic

I don't think I agree. According to this metric, for example, you should always use pattern matching on `Option`, instead of `orElse`, `flatMap` and `getOrElse` and always use `foldRight` on Lists, instead of `map/filter/collect`.

Combinators actually are great at expressing intent, whereas with "direct style" in general it's easy to miss the forest for the trees.  There are certainly cases in which one can go overboard, and a simple pattern match/direct usage simplifies that. But I think a simple call to `mapF` is well below that threshold.

Furthermore, I find the `mapF` version more readable, it's easier for the me to see the typed transformation in the `F[A] => F[A]` that the treading of the parameter in the direct use of Kleisli, which I guess reinforces

> "Readability" is an incredibly arbitrary metric

diesalbla:
Closed in favour of https://github.com/http4s/http4s/pull/2553

PR #2553, opened 2019-05-05
Use Concurrent timeoutTo method.

diesalbla:
Use method [`Concurrent.timeoutTo`](https://github.com/typelevel/cats-effect/blob/master/core/shared/src/main/scala/cats/effect/Concurrent.scala#L349-L354).

### DIFF
--- a/server/src/main/scala/org/http4s/server/middleware/Timeout.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/Timeout.scala
@@ -3,8 +3,8 @@ package server
 package middleware
 
 import cats.data.Kleisli
-import cats.effect._
-import cats.implicits._
+import cats.effect.{Concurrent, Timer}
+import cats.syntax.applicative._
 import scala.concurrent.duration.FiniteDuration
 
 object Timeout {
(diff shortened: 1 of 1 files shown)


server/src/main/scala/org/http4s/server/middleware/Timeout.scala
```diff
@@ -21,9 +21,7 @@ object Timeout {
       http: Kleisli[F, A, Response[G]])(
       implicit F: Concurrent[F],
       T: Timer[F]): Kleisli[F, A, Response[G]] =
-    http
-      .mapF(respF => F.race(respF, T.sleep(timeout) *> timeoutResponse))
-      .map(_.merge)
+    http.mapF(Concurrent.timeoutTo(_, timeout, timeoutResponse))
```
SystemFw: you should be able to use syntax for the `Concurrent` op, `_.timeoutTo(timeout, response)`

diesalbla: Thought about that before the PR. IMHO, this is simple enough, and small enough, not to make it worth another `syntax` import.

server/src/main/scala/org/http4s/server/middleware/Timeout.scala
```diff
@@ -3,8 +3,8 @@ package server
 package middleware
 
 import cats.data.Kleisli
-import cats.effect._
-import cats.implicits._
+import cats.effect.{Concurrent, Timer}
+import cats.syntax.applicative._
```
rossabaker: I still don't like these imports, but I'm planning to do a passthrough to undo some of the damage IntelliJ has done, so I'll clean up then.

PR #2555, opened 2019-05-05
FileService: remove OptionT from inner methods.

diesalbla:
In the inner methods `getPartialContentFile` and `getFile`, the `OptionT` wrapper serves no purpose, since no special method is used. We remove its from those functions, and put it in the call places, where they are actually used.


server/src/main/scala/org/http4s/server/staticcontent/FileService.scala
```diff
@@ -84,11 +85,11 @@ object FileService {
   // Attempt to find a Range header and collect only the subrange of content requested
   private def getPartialContentFile[F[_]](file: File, config: Config[F], req: Request[F])(
       implicit F: Sync[F],
-      cs: ContextShift[F]): OptionT[F, Response[F]] =
-    OptionT.fromOption[F](req.headers.get(Range)).flatMap {
```
diesalbla: I found the `OptionT.fromOption` followed by `flatMap` in here particularly misleading: it suggested there could be an effectful computation where there is not.

PR #2561, opened 2019-05-09
Remove EitherT from the DecodeResult.

diesalbla:
The type of `DecodeResult[F[_], A]` was defined as `EitherT[F, DecodeFailure, A]` , which is just a wrapper for `F[Either[L, R]]`. It turns out that, for most operations, the full api of `EitherT`
is not used, and at each point only one or two operations are.

For this reason, we drop the use of `EitherT`. We reduce the type `DecodeResult[F, A]` to be just the `F[Either[DecodeResult, A]`, without the `EitherT`. 

To keep the convenience we have at certain places, we introduce a syntax extension (an implicit class), "FEither", which replicates some methods of `EitherT`, but without colliding with `map` and `flatMap`.

This breaks both source and binary compatibility.

_Note for reviewers:_ A bit over +50-50 line changes are whitespace changes, introduced by automatic formatting by `scalafmt`.

### DIFF
--- a/argonaut/src/test/scala/org/http4s/argonaut/ArgonautSpec.scala
+++ b/argonaut/src/test/scala/org/http4s/argonaut/ArgonautSpec.scala
@@ -103,7 +103,7 @@ class ArgonautSpec extends JawnDecodeSupportSpec[Json] with Argonauts {
     "decode JSON from an Argonaut decoder" in {
       val result = jsonOf[IO, Foo]
         .decode(Request[IO]().withEntity(jObjectFields("bar" -> jNumberOrNull(42))), strict = true)
-      result.value.unsafeRunSync must beRight(Foo(42))
+      result.unsafeRunSync must beRight(Foo(42))
     }
 
     // https://github.com/http4s/http4s/issues/514
@@ -114,7 +114,7 @@ class ArgonautSpec extends JawnDecodeSupportSpec[Json] with Argonauts {
         val json = Json("wort" -> jString(wort))
         val result =
           jsonOf[IO, Umlaut].decode(Request[IO]().withEntity(json), strict = true)
-        result.value.unsafeRunSync must_== Right(Umlaut(wort))
+        result.unsafeRunSync must_== Right(Umlaut(wort))
       }
     }
 
@@ -122,7 +122,7 @@ class ArgonautSpec extends JawnDecodeSupportSpec[Json] with Argonauts {
       val result = ArgonautInstancesWithCustomErrors
         .jsonOf[IO, Foo]
         .decode(Request[IO]().withEntity(jObjectFields("bar1" -> jNumberOrNull(42))), strict = true)
-      result.value.unsafeRunSync must beLeft(InvalidMessageBodyFailure(
+      result.unsafeRunSync must beLeft(InvalidMessageBodyFailure(
         "Custom Could not decode JSON: {\"bar1\":42.0}, error: Attempt to decode value on failed cursor., cursor: CursorHistory(List(El(CursorOpDownField(bar),false)))"))
     }
   }
--- a/bench/src/main/scala/org/http4s/bench/CirceJsonBench.scala
+++ b/bench/src/main/scala/org/http4s/bench/CirceJsonBench.scala
@@ -17,17 +17,16 @@ class CirceJsonBench {
 
   @Benchmark
   def decode_byte_buffer(in: BenchState): Either[DecodeFailure, Json] =
-    jsonDecoderByteBuffer[IO].decode(in.req, strict = true).value.unsafeRunSync
+    jsonDecoderByteBuffer[IO].decode(in.req, strict = true).unsafeRunSync
 
   @Benchmark
   def decode_incremental(in: BenchState): Either[DecodeFailure, Json] =
-    jsonDecoderIncremental[IO].decode(in.req, strict = true).value.unsafeRunSync
+    jsonDecoderIncremental[IO].decode(in.req, strict = true).unsafeRunSync
 
   @Benchmark
   def decode_adaptive(in: BenchState): Either[DecodeFailure, Json] =
     jsonDecoderAdaptive[IO](in.cutoff, MediaType.application.json)
       .decode(in.req, strict = true)
-      .value
       .unsafeRunSync
 }
 
--- a/boopickle/src/main/scala/org/http4s/booPickle/BooPickleInstances.scala
+++ b/boopickle/src/main/scala/org/http4s/booPickle/BooPickleInstances.scala
@@ -10,6 +10,7 @@ import java.nio.ByteBuffer
 import org.http4s._
 import org.http4s.EntityEncoder.chunkEncoder
 import org.http4s.headers.`Content-Type`
+import org.http4s.util.FEither._
 import scala.util.{Failure, Success}
 
 /**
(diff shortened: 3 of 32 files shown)


rossabaker:
`FEither`, an ad hoc concept instead of the well understood `EitherT`, is imported into most of the files that changed.  I don't see the win here.

I regret the alias, but I haven't ever really regretted the transformer.

diesalbla:
@rossabaker The problem is that `cats.data.EitherT` is a rather overkill solution for the problem at hand. You do not really need a monad transformer in this context, because you are never using the `Monad` instance for `EitherT`. Nowhere are you in the case in which you have a generic function 
`def fili[M[_]: Monad]` that you may want to invoke with the type `M[A] = F[Either[L, A]]` using the combined `F`-`Either` instance of Monad for that.
 
All you use `EitherT` for in the code is as a carrier to apply some specific transformations from an `F[Either[A, B]]` to another `F[Either[C, D]]`, most of which are neither the `map` or `flatMap` of `EitherT`. This is done at the cost of allocating the `EitherT` objects.

This Pull Requests explores an alternative way to do that: to not use a transformer where there is nothing that needs transforming, and to write those transformations as a set of extension methods (through an `implicit class`).

jmcardon:
I honestly feel like you are waging a war against `EitherT` (you raised similar concerns in the PR in the csrf methods) when it's a relatively low cost (in this case), well understood construct within FP, not just cats.

EitherT is:
- `AnyVal`, which in many cases is zero-cost.
- Well understood.
- Communicates intent.

In places related to things like decoding something with a set of typed errors, you tend to return `EitherT`s because the nature of the content of a particular operation is easy to reason about for users in terms of a fold in which they should collapse the potential error and the content into a singular case. That's it. It's not a super powerful construct, it's not "too powerful" and in the case that you absolutely hate it, you can literally just call `.value` on it and you're working with `F[Either[E, A]]` as you initially intended.

The amount of work and code this will potentially break for a fairly low cost thing is not worth it.

Moreover, you have this:
```scala
implicit class FEitherOps[F[_], L, R](val felr: F[Either[L, R]]) extends AnyVal {
```

Which is... quite literally a reinvention of `EitherT`, except you somehow avoid some huge cost of `EitherT`, ignoring that your entire code is going through multiple interpreters that are essentially based on thunks and anonymous closures which the jvm does not have a fun time inlining. 

> This shows a leaner solution for that: to write those transformations as a set of extension methods (through an implicit class), and avoid the EitherT monad transformer

No, it's not leaner, because it's literally the same thing that's already in `EitherT` (which again, is `AnyVal`), you just rewrote it to not use the transformer, with the drawback that now any code workign with `F[Either[E, A]]` requires an import within http4s for something that very clearly (and already does) belong in cats.

rossabaker:
`EitherT` is [not an AnyVal](https://github.com/typelevel/cats/issues/546), but I agree with the remainder.

jmcardon:
Sorry, was wrong, fixed.

ritschwumm:
in my experience, returning EitherT from a decoding operation is quite often not what you need, because that's not the thing you'd flatMap over - more often than not i'd have some error facility in F itself and would want to do something with the naked either to put the Left in there somehow while staying in a for-comprehension on F.

BUT: adding an implicit class to `F[Either[...]]` imho is even worse than returning an `EitherT` in the first place. if i _really_ want to flatMap on both F and Either, i can just convert to EitherT and have the full power and the well-known interface of EitherT. the implicit class, otoh, does not pull its weight: it does increase complexity and makes code harder to read, but provides almost no abstraction and no additional value over EitherT (except being a bit more performant, but i doubt this will have any measurable effect in a real-world application).

diesalbla:
> I honestly feel like you are waging a war against EitherT [...] it's a relatively low cost (in this case), well understood construct within FP, not just cats.

I do not think that Monad Transformers are an easy concept to grasp, even if they are in `cats`. The `Free` monad constructor is also in `cats`, but I would not call it easy to grasp. Even if it were, I would still avoid using it for a problem for which an easier solution exists. And I do not have a was against `EitherT`, but against its abuse when it is not needed, as was the case in `DecodeResult`; because using it when it is not needed is a sign that it is not, In fact, as well understood.

Now, there are two parts to the API of `EitherT` in `cats`. One part is the kit of special functions, like `leftFlatMap`, that are just a special combination of the `F.map`, `F.flatMap` methods with the transformations on `Either`. In Haskell, you would not use use `EitherT` for this, and just use a function: 
```Haskell
leftFlatMap:: Monad f => (a -> f (Either d b)) -> f (Either a b) -> f (Either d b)
leftFlatMap e f = e >>= (f `either` (pure . Right) ))
```
This shows that `EitherT` is, in many of those methods, just a swiss-knife of many little mixed operations on `F` and `Either`. Nothing to do with what a Monad Transformer does. All you really need in this case are functions, either isolated or as extension methods. 

It is from this perspective, and for this reason, that I have, yes, replicated the functions of `cats.data.EitherT` in the  implicit class `FEither`, since an implicit class is the only way to define extension methods in Scala.

The second part is the part in which you really use `EitherT` as a Monad Transformer, which is to say, when you want to have an instance of `Functor`, or `Monad`, or `MonadError` for the `EitherT` data type. However, **nowhere in the code of http4s** are you actually using the instances of `Monad[EitherT[F, L, ?]]`, or `MonadError[EitherT[F, L, ?]]`, to pass them to a generic function. And that is exactly what monad transformers are for. Note that using those instances is different from using the methods `map` and `flatMap` of `EitherT`, when you are tied to that type.

jmcardon:
> I do not think that Monad Transformers are an easy concept to grasp

That's your opinion. They're fairly easy once you understand everything else, and not any harder than something like `Alternative` typeclass.

> The Free monad constructor is also in cats, but I would not call it easy to grasp.

Everything in cats is easy to grasp != transformers are easy to grasp.

> And I do not have a was against EitherT, but against its abuse when it is not needed, as was the case in DecodeResult; because using it when it is not needed is a sign that it is not, In fact, as well understood.

"Not needed" _in your opinion_.

> In Haskell, you would not use use EitherT for this, and just use a function:

That's... so not true. I write haskell more than I do scala, and there's a ton of potential solutions to something like this:
- use `MonadError` for decodeerror directly.
- `EitherT` (or ExceptT/whatever stack you are using) directly.
- Classy prisms to compose errors.
- [Freer member to handle errors](http://hackage.haskell.org/package/freer-simple-1.2.1.0/docs/Control-Monad-Freer-Error.html).

And whichever you use, it's up to the author. Even with the better inference and whatnot, workign with `f (Either e a)` in haskell isn't the rule at all. In general `EitherT` (or `ErrorT`, `ExceptT`) is a choice that's fairly easy to peel if you want to peek at the underlying `f (Either a b)` form.

> This shows that EitherT is, in many of those methods, just a swiss-knife of many little mixed operations on F and Either. Nothing to do with what a Monad Transformer does. All you really need in this case are functions, either isolated or as extension methods.

What precisely does a monad transformer _do_ in your opinion that makes it so much different than something like `MaybeT` or `ContT`?

> The second part is the part in which you really use EitherT as a Monad Transformer, which is to say, when you want to have an instance of Functor, or Monad, or MonadError for the EitherT data type. However, nowhere in the code of http4s are you actually using the instances of Monad[EitherT[F, L, ?]], or MonadError[EitherT[F, L, ?]], to pass them to a generic function. And that is exactly what monad transformers are for. Note that using those instances is different from using the methods map and flatMap of EitherT, when you are tied to that type.

You are missing the forest for the trees. `EitherT` is a convenience here, which is one singular allocation more than your proposed ad-hoc solution. `EitherT` tells me "hey you should probably deal with this error before you use this value, or alternatively, compose more operatioins with errors that you know (by potentially lifting them into a "larger" error ADT". It's quite literally what `ZIO` was initially doing, before it added reader capabilities, to just bake the error into the type for performance.

jmcardon:
I seriously don't see the problem of leaving `EitherT` in here at all. As it stands, literally everything you want to do to it, in the form of `F[Either[A, B]]` can be done by simply calling `.value` on the wrapper. That's it, a minor inconvenience at best, there's no honest to god performance penalty on it that isn't significantly dominated by whatever cost the decoders have.

This is bikeshedding that's not changing anything semantically, it's simply a large stylistic change for something that hasn't been a pain point for http4s at all.

diesalbla:
Since there is no support for the PR, I am closing it.

PR #2562, opened 2019-05-09
Fixed issue in PoolManager which caused requests not to ever be executed under some circumstances

aldiyen:
I believe this properly fixes the issue described in PR #2546. I'm not terribly familiar with this code, so it's possible that there was a reason why `numConnectionsCheckHolds` was being called outside of the monadic context inside of `createConnection`, in which case I think an equivalent solution could be localized within `borrow`.

### DIFF
--- a/client/src/main/scala/org/http4s/client/PoolManager.scala
+++ b/client/src/main/scala/org/http4s/client/PoolManager.scala
@@ -98,18 +98,20 @@ private final class PoolManager[F[_], A <: Connection[F]](
     * @param callback The callback to complete with the NextConnection.
     */
   private def createConnection(key: RequestKey, callback: Callback[NextConnection]): F[Unit] =
-    if (numConnectionsCheckHolds(key)) {
-      incrConnection(key) *> F.start {
-        Async.shift(executionContext) *> builder(key).attempt.flatMap {
-          case Right(conn) =>
-            F.delay(callback(Right(NextConnection(conn, fresh = true))))
-          case Left(error) =>
-            disposeConnection(key, None) *> F.delay(callback(Left(error)))
-        }
-      }.void
-    } else {
-      addToWaitQueue(key, callback)
-    }
+    F.ifM(F.delay(numConnectionsCheckHolds(key)))(
+      {
+        incrConnection(key) *> F.start {
+          Async.shift(executionContext) *> builder(key).attempt.flatMap {
+            case Right(conn) =>
+              F.delay(callback(Right(NextConnection(conn, fresh = true))))
+            case Left(error) =>
+              disposeConnection(key, None) *> F.delay(callback(Left(error)))
+          }
+        }.void
+      }, {
+        addToWaitQueue(key, callback)
+      }
+    )
 
   private def addToWaitQueue(key: RequestKey, callback: Callback[NextConnection]): F[Unit] =
     F.delay {
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ stages:
     branches:
       only:
         - master
-        - /^release-/
+        - /^series\//
     if: type = push
 
 env:
--- a/blaze-client/src/test/scala/org/http4s/client/blaze/BlazeClientSpec.scala
+++ b/blaze-client/src/test/scala/org/http4s/client/blaze/BlazeClientSpec.scala
@@ -20,6 +20,7 @@ class BlazeClientSpec extends Http4sSpec {
 
   def mkClient(
       maxConnectionsPerRequestKey: Int,
+      maxTotalConnections: Int = 5,
       responseHeaderTimeout: Duration = 1.minute,
       requestTimeout: Duration = 1.minute,
       chunkBufferMaxSize: Int = 1024
@@ -29,6 +30,7 @@ class BlazeClientSpec extends Http4sSpec {
       .withCheckEndpointAuthentication(false)
       .withResponseHeaderTimeout(responseHeaderTimeout)
       .withRequestTimeout(requestTimeout)
+      .withMaxTotalConnections(maxTotalConnections)
       .withMaxConnectionsPerRequestKey(Function.const(maxConnectionsPerRequestKey))
       .withChunkBufferMaxSize(chunkBufferMaxSize)
       .resource
@@ -63,25 +65,12 @@ class BlazeClientSpec extends Http4sSpec {
   }
 
   "Blaze Http1Client" should {
-    // This incident is going on my permanent record.
     withResource(
       (
-        mkClient(0),
-        mkClient(1),
-        mkClient(3),
-        mkClient(3),
-        mkClient(3),
-        mkClient(1, 20.seconds),
         JettyScaffold[IO](5, false, testServlet),
         JettyScaffold[IO](1, true, testServlet)
       ).tupled) {
       case (
-          failClient,
-          successClient,
-          client,
-          seqClient,
-          parClient,
-          successTimeClient,
           jettyServer,
           jettySslServer
           ) => {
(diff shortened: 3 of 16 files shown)


client/src/main/scala/org/http4s/client/PoolManager.scala
```diff
@@ -98,18 +98,20 @@ private final class PoolManager[F[_], A <: Connection[F]](
...
-          case Right(conn) =>
-            F.delay(callback(Right(NextConnection(conn, fresh = true))))
-          case Left(error) =>
-            disposeConnection(key, None) *> F.delay(callback(Left(error)))
-        }
-      }.void
-    } else {
-      addToWaitQueue(key, callback)
-    }
+    F.ifM(F.delay(numConnectionsCheckHolds(key)))(
```
jmcardon: `numConnectionsCheckHolds` accesses a mutable map, but it does not mutate it. I don't think this requires `delay`.

Every call also within the `if` is thunked within delay as well.

aldiyen:
The problem is that `createConnection` is called after `decrConnection` is called, however `decrConnection` decrements the connection count asynchronously, while `createConnection` checks the connection count synchronously. As a result, if the connection pool is at `maxTotalConnections`, `createConnection` adds to the wait queue when the code expects it to create a new connection.

The unit tests I added fail without this change. I’m open to considering alternative solutions if someone can point me in the right direction here.

jmcardon:
You are likely right, thus this seems more of a synchronization issue than an issue with thunking a boolean check. 

`ifM` and `delay` only thunk statements, they do not guarantee any sort of atomicity in execution, thus likely your tests are only passing because the thunking forces it to go through the evaluator and execute slower (thus you observe it passing under a single fiber test), but under a bunch of concurrent calls to borrow, this will likely still fail. The real problem is with `allocated`, being a mutable map that is not thread safe.

There's a few solutions:
- `fs2.Ref` Map instead of `mutable.Map`.
- AtomicReference instead of ref (at the cost of handling low level synchro yourself, so ref is suggested)
- `@volatile` reference if you deem you don't need compareAndSet semantics.

aldiyen:
Hm, you’re probably right. Unfortunately state information is contained both in the current total connection count and the map of `RequestKey` to connection, which I think implies a more significant refactoring of `PoolManager`’s state management than even what you’ve just proposed.

aldiyen:
Seems like @rossabaker is going to resolve this by replacing `PoolManager` with a better-managed keyed pool, which seems good. Closing this PR in favor of that solution.

jmcardon:
Thanks for the attempt @aldiyen ! If it wasn't for this PR we might have not caught the small synchro issue.

rossabaker:
I should note that the keypool is non-trivial.  The current implementation only enforces limits on the idle connections, not overall.  And retrofitting it is proving to be a struggle.  If there is a quick win here, it's still welcome, even if I believe in the keypool in the long term.

aldiyen:
My best guess as to why it's never returning in some cases is that the connection count is being decremented before the callback is being added to the wait queue. Something like:
# Connection count is at exactly `maxTotal`
# `decrConnection` async operation is created
# `addToWaitQueue` async operation is created
# `decrConnection` thunks, but nothing is in wait queue
# `addToWaitQueue` thunks, but nothing ever checks wait queue again

This may not be an issue in situations with more activity going on in parallel, because I assume something will eventually check the wait queue, but under certain types of utilization I imagine it would block real applications indefinitely.

I think the change I've made here will actually alleviate this problem for that use case, but it does seem likely that some very specific parallelized utilization of the pool could still get it locked up like this if things line up just right, even with that change. The only guaranteed non-halting solution that comes to mind offhand is to make it so that the code in `borrow` for the `None if curTotal == maxTotal` case _always_ actually creates a new connection, without considering whether or not it would exceed `maxTotal` or the max requests for the specified key -- this would cause situations where the pool would exceed its configured limits, but due to the lack of synchronization on the state that is likely possible as-is anyway.

jmcardon:
See, the thing is I don't believe if fixes any order of operations, because there's no specific guarantee that `F.delay` does anything other than, by thunking, create a literal delay of hopping through the interpreter.

Now, checking a mutable data structure is actually an effect, so it's not as if `F.delay(checkMutableThing)` is wrong to do, but there's no compiler optimization applied to this snippet where this matters (it matters in the case that you have a compiler optimization that reorders pure statements, in the form of common subexpression elimination).

> I think the change I've made here will actually alleviate this problem for that use case, as it fixes the order of operations

Every call to `createConnection` is following a thunk, connected by `*>`. `F.delay` doesn't guarantee any such behavior will be fixed, at all. Like I said prior, there's no operation order fixed by this, since there's absolutely no atomicity guarantees between the check and the update. In certain cases the queue may be updated, and in others, it's likely that it wont be, it depends entirely on operation ordering between threads and how the jvm chooses to update the variable. We have no way to enforce such guarantees in the current code.

> This may not be an issue in situations with more activity going on in parallel, because I assume something will eventually check the wait queue, but under certain types of utilization I imagine it would block real applications indefinitely.

This is likely an issue no matter the type of use. We simply cannot make any guarantees about what the state of the queue is like when we're very clearly using a mutable, non-thread safe data structure in code that updates it in different threads. Without any sort of atomicity or synchronization guarantees, we have absolutely no guarantees of what the state of the `waitQueue` will be at certain points in time.

aldiyen:
Hm, interesting. Forgive my ignorance, my real-world experience with Cats is fairly limited, and so I have probably been operating under some faulty assumptions.

Are you saying that in code like `decrConnection(key) *> createConnection(k, cb)` the `decrConnection(key)` operation could be evaluated prior to the `createConnection(k, cb)` logic? Or simply that these two operations, while ordered relative to one another, are not guaranteed to execute anywhere near each other in time because the underlying threadpool may be doing many other things in between these two operations?

jmcardon:
```
Are you saying that in code like decrConnection(key) *> createConnection(k, cb) the decrConnection(key) operation could be evaluated prior to the createConnection(k, cb) operation? 
```
It for sure will be evaluated before `createConnection`.

What I'm getting at, is, in this snippet:
```scala
incrConnection(key) *> F.start {
          Async.shift(executionContext) *> builder(key).attempt.flatMap {
            case Right(conn) =>
              F.delay(callback(Right(NextConnection(conn, fresh = true))))
            case Left(error) =>
              disposeConnection(key, None) *> F.delay(callback(Left(error)))
          }
        }.void
```
The `shift` is what causes a potentially incorrect view of total allocated connections because how the operation following that shift views the state of `allocated` might not be the state that should be seen, it's hard to say how operation reordering happens without explicitly printing things and observing the state when going through different threads.

`numConnectionsCheckHolds` is:
```scala
private def numConnectionsCheckHolds(key: RequestKey): Boolean =
    curTotal < maxTotal && allocated.getOrElse(key, 0) < maxConnectionsPerRequestKey(key)
```
^ `allocated` is not being updated at all, no state is being modified. Thunking this operation doesn't guarantee anything about how the following shifted operation will view `allocated`. There's no significant semantic difference between `if (numConnectionsCheckHolds(key))` and `F.ifM(F.delay...` that will give us a nice ordering guarantee, because ultimately all you're doing is making this check hop through an additional thunk, but the happens-before guarantees are the same.

aldiyen:
I see, okay, that is what I thought. Where, though, in the shifted code does it check `allocated`, though? I'm fairly certain that the issue described in #2546 and reproduced in the tests added in this PR is occurring in the operations preceding the `incrConnection(key) *> F.start { ... }` portion, which is also the only place I see the check against `allocated` being made (inside of `numConnectionsCheckHolds`).

I do see that it could choose the "wrong" path after thunking `F.delay(numConnectionsCheckHolds(key))`, in the sense that it could lead to a situation where ther are more connections than should be allowed by `maxTotal`, but that seems preferable to a situation where one can get an effect from `borrow` which might not ever be evaluated.

rossabaker:
The more ambitous fix is taking too long, so reconsidering this.

jmcardon:
I'm alright with approving this, I just don't think it will fix anything @rossabaker.

PR #2564, opened 2019-05-10
BooPickle and Circe Decode Instances: more Cats

diesalbla:
Both the `BooPickle` and the `Circe` decoder instances call to `collectBinary` and then applied an analysis to extract either the decoded result or a failure. This analysis phase was written as a function of Chunk to a DecodeResult, flatMapped to the result of  `collectBinary`. 

This PR rewrites it as a *pure* function of `Chunk[Byte]` to `Either`, which is applied to the result of `collectBinary` using the `EitherT.subflatMap` function.

### DIFF
--- a/boopickle/src/main/scala/org/http4s/booPickle/BooPickleInstances.scala
+++ b/boopickle/src/main/scala/org/http4s/booPickle/BooPickleInstances.scala
@@ -20,18 +20,15 @@ trait BooPickleInstances {
 
   private def booDecoderByteBuffer[F[_]: Sync, A](msg: Message[F])(
       implicit pickler: Pickler[A]): DecodeResult[F, A] =
-    EntityDecoder.collectBinary(msg).flatMap { chunk =>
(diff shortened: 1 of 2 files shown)


boopickle/src/main/scala/org/http4s/booPickle/BooPickleInstances.scala
```diff
@@ -4,13 +4,13 @@ package booPickle
 import boopickle.Default._
 import boopickle.Pickler
 import cats.Applicative
+import cats.syntax.either._
```
rossabaker: I still would rather not use these specific imports.

diesalbla:
It turns out that the `Try.toEither` instance method is not available in version 2.11 of the library, so we can keep that as a `match` statement and have no need of the `Either` syntax there.

rossabaker:
Just needs a `test:scalafmt`, and the build should pass, and then :+1: from me.

PR #2567, opened 2019-05-12
Deprecate `AuthedService`, replace it with `AuthedRoutes` with updated API

bplommer:
- Add `of` and `apply` methods to `AuthedService` companion that are consistent with those for `HttpRoutes`
- Add `mapK` to AuthedRequest
- Add `mapK` extension method for AuthedService

AuthedService seems to have fallen behind `HttpRoutes` - perhaps it's worth a breaking PR to rename it to `AuthedRoutes` and change the order of the type params?

EDIT: The overloaded `apply` method on `AuthedService` caused a bunch of things to break. I've renamed the new one to `apply_` for now, but I'm not very happy with that name - thoughts?

### DIFF
--- a/core/src/main/scala/org/http4s/AuthedRequest.scala
+++ b/core/src/main/scala/org/http4s/AuthedRequest.scala
@@ -4,7 +4,10 @@ import cats._
 import cats.data._
 import cats.implicits._
 
-final case class AuthedRequest[F[_], A](authInfo: A, req: Request[F])
+final case class AuthedRequest[F[_], A](authInfo: A, req: Request[F]) {
+  def mapK[G[_]](fk: F ~> G): AuthedRequest[G, A] =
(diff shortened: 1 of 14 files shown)


rossabaker:
Thanks, I like this.  The build failure is a flake.

The order of the type parameters on `AuthedService` are weird because of #1506.  Does this reasoning no longer hold?

If we deprecate the `AuthedService` alias in favor of `AuthedRoutes`, and deprecate the `AuthedService` companion object in favor of a new `AuthedRoutes` one, we could do this in a binary compatible way, get the `apply` syntax back, and not have the awkward `apply_` method name.  What do you think?

bplommer:
Good point about deprecating `AuthedService` - I'll update the PR with `AuthedRoutes`. I'll also fix the type parameter orders.

PR #2570, opened 2019-05-15
Release v0.20.1

rossabaker:


rossabaker:
Failure is another test flicker.  I tend to think all the `unsafeRunSync` in blaze tests is the culprit here, but it's a big job to rip out.

PR #2571, opened 2019-05-15
Static content caching header fix

### LINKED ISSUE #2529: StaticFile.fromFile ignores modified ETag if unmodified since
StaticFile.fromFile will return early with a 302 if the resource has not modified since, ignoring the request ETag. If the ETag provided in the If-None-Match header is different from that of the resource, a 302 should not be returned, regardless of the not modified since check.

As far as I understand from the spec, both conditions need to be checked before returning a 302:

> If both an entity tag and a Last-Modified value have been
>         provided by the origin server, SHOULD use both validators in
>         cache-conditional requests.

This makes more sense if you consider rolling back a file to an older version, where the modification date is taken as when the version was authored, rather than when it was rolled back. Maybe this is unusual, although the above suggests it's covered in the spec and I just encountered it in the wild.

rhyskeepence:
Fixes #2529

### DIFF
--- a/core/src/main/scala/org/http4s/StaticFile.scala
+++ b/core/src/main/scala/org/http4s/StaticFile.scala
@@ -163,8 +163,9 @@ object StaticFile {
       etagCalc: ETag,
       lastModified: Option[HttpDate]): Option[Response[F]] = {
 
-    implicit val conjunction: Semigroup[Boolean] =
-      (x: Boolean, y: Boolean) => x && y
+    implicit val conjunction = new Semigroup[Boolean] {
+      def combine(x: Boolean, y: Boolean): Boolean = x && y
+    }
 
     List(etagMatch(req, etagCalc), notModifiedSince(req, lastModified)).combineAll
       .filter(identity)
--- a/core/src/main/scala/org/http4s/StaticFile.scala
+++ b/core/src/main/scala/org/http4s/StaticFile.scala
@@ -1,18 +1,21 @@
 package org.http4s
 
+import java.io._
+import java.net.URL
+
+import cats.Semigroup
 import cats.data._
 import cats.effect._
-import cats.implicits.{catsSyntaxEither => _, _}
+import cats.implicits._
 import fs2.Stream._
 import fs2.io._
 import fs2.io.file.readRange
-import java.io._
-import java.net.URL
+import io.chrisdavenport.vault._
 import org.http4s.Status.NotModified
 import org.http4s.headers._
 import org.log4s.getLogger
+
 import scala.concurrent.ExecutionContext
-import _root_.io.chrisdavenport.vault._
 
 object StaticFile {
   private[this] val logger = getLogger
@@ -132,13 +135,7 @@ object StaticFile {
 
           val lastModified = HttpDate.fromEpochSecond(f.lastModified / 1000).toOption
 
-          // See if we need to actually resend the file
-          val notModified: Option[Response[F]] = ifModifiedSince(req, lastModified)
-
-          // Check ETag
-          val etagModified: Option[Response[F]] = ifETagModified(req, etagCalc)
-
-          notModified.orElse(etagModified).orElse {
+          notModified(req, etagCalc, lastModified).orElse {
             val (body, contentLength) =
               if (f.length() < end) (empty.covary[F], 0L)
               else (fileToBody[F](f, start, end, blockingExecutionContext), end - start)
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/http4s/StaticFile.scala
```diff
@@ -1,18 +1,21 @@
 package org.http4s
 
+import java.io._
+import java.net.URL
+
+import cats.Semigroup
 import cats.data._
 import cats.effect._
-import cats.implicits.{catsSyntaxEither => _, _}
+import cats.implicits._
```
rossabaker: That catsSyntaxEither trick actually is necessary with the strict compiler flags and 2.11 support.

rhyskeepence: Got it, thanks!

core/src/main/scala/org/http4s/StaticFile.scala
```diff
@@ -1,18 +1,21 @@
 package org.http4s
 
+import java.io._
+import java.net.URL
+
+import cats.Semigroup
```
aeons: We generally have all imports in the same list, ordered alphabetically.

rossabaker: We ended up removing the IntelliJ setting that did this for people, right?  It's a pity that scalafmt can't just do this for people.

rossabaker:
I don't immediately recognize that lingering Scala 2.11 failure.  The other build failure is a blaze flake and not your fault.

core/src/main/scala/org/http4s/StaticFile.scala
```diff
@@ -163,25 +160,37 @@ object StaticFile {
...
     } yield res)
 
-  private def ifETagModified[F[_]](req: Option[Request[F]], etagCalc: ETag) =
+  private def notModified[F[_]](
+      req: Option[Request[F]],
+      etagCalc: ETag,
+      lastModified: Option[HttpDate]): Option[Response[F]] = {
+
+    implicit val conjunction: Semigroup[Boolean] =
+      (x: Boolean, y: Boolean) => x && y
```
rossabaker: This works starting in Scala 2.12 because of the single abstract method thingy, but it fails to compile in 2.12.  I think it just needs the whole `new Semigroup { ... }` ceremony.

server/src/main/scala/org/http4s/server/middleware/Metrics.scala
```diff
@@ -39,7 +39,7 @@ object Metrics {
       classifierF: Request[F] => Option[String] = { _: Request[F] =>
         None
       }
-  )(routes: HttpRoutes[F])(implicit F: Effect[F], clock: Clock[F]): HttpRoutes[F] =
+  )(routes: HttpRoutes[F])(implicit F: Sync[F], clock: Clock[F]): HttpRoutes[F] =
```
rossabaker: This will break MiMa.  I think it got picked up when you changed branches.

tests/src/test/scala/org/http4s/StaticFileSpec.scala
```diff
@@ -66,6 +86,21 @@ class StaticFileSpec extends Http4sSpec {
       response.map(_.status) must beSome(NotModified)
     }
 
+    "Send modified files when either the ETag or last modified date do not match" in {
```
rossabaker: Isn't this testing when the Etag doesn't match, and it's not tested when the If-Modified-Since doesn't match?  The "either" implies two tests to me.

core/src/main/scala/org/http4s/StaticFile.scala
rhyskeepence: Yeah, sorry, I was using the default IntelliJ settings. This should be resolved now!

tests/src/test/scala/org/http4s/StaticFileSpec.scala
rhyskeepence: Yes, you're correct, there should be two tests.

server/src/main/scala/org/http4s/server/middleware/Metrics.scala
rhyskeepence: What's the best way to resolve this? Change the target branch of this PR to master, or create a new PR with the correct source branch?

rossabaker:
We're going to have to rebase to get the two commits that aren't @rhyskeepence's off this PR, but it looks good to me after that.

server/src/main/scala/org/http4s/server/middleware/Metrics.scala
rossabaker: Target master won't get it into the 0.20 series.  Which we're about to release, but since you're still awake, we'll squeeze this in.

You can `git rebase -i` against series/0.20 and drop the two commits that aren't yours and force push, or you can open a new PR.  Whatever's easier for you.

rhyskeepence: Thank you, I have rebased and force pushed.

PR #2573, opened 2019-05-16
Release v0.20.1, take 2: fix publishable test

rossabaker:
This PR will again try to release v0.20.1 on merge.

We failed to publish 0.20.1 because it was looking for oraclejdk8 instead of openjdk8.  I think our build steps render that whole check unnecessary, but this is the easy fix shortly before midnight.

Threw in blaze-0.14.1, which I regretted missing earlier today.  Easier to do that in one PR than revert to a snapshot and then release again.


rossabaker:
It didn't work anyway.  Let's split this up more responsibly.

PR #2574, opened 2019-05-16
Simplify code in Follow Redirect: No Options

diesalbla:
Some changes to simplify the code in `FollowRedirect` by removing false options. 

- The variable `pureBody` was defined as a `Some`, so there was no   need for the `Option` in its type.
- The Method `nextRequest` was identical in both branches except for the last instruction, about the body, which depends on whether `bodyOpt` is a `Some` or a `None`
- The `bodyOpt` parameter was fixed in each call to either a `Some` or a `None`, which means we can fold last part of the `def nextRequest` method after the call. 
- The `nextRequest` was an Option only because the `pureBody` was.  Since `pureBody` is always `Some`, we can remove branching there.
- And thus remove the `nextReq.fold(dontRedirect)` at the end.

### DIFF
--- a/client/src/main/scala/org/http4s/client/middleware/FollowRedirect.scala
+++ b/client/src/main/scala/org/http4s/client/middleware/FollowRedirect.scala
@@ -4,7 +4,6 @@ package middleware
 
 import cats.effect._
 import cats.implicits._
-import fs2._
 import org.http4s.Method._
 import org.http4s.headers._
 import org.http4s.util.CaseInsensitiveString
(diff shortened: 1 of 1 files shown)


client/src/main/scala/org/http4s/client/middleware/FollowRedirect.scala
```diff
@@ -65,42 +62,26 @@ object FollowRedirect {
...
-                .withEmptyBody
-          }
-
         def doRedirect(method: Method): Resource[F, Response[F]] =
           if (redirects < maxRedirects) {
             // If we get a redirect response without a location, then there is
             // nothing to redirect.
             redirectUri.fold(dontRedirect) { nextUri =>
               // We can only redirect safely if there is no body or if we've
               // verified that the body is pure.
```
rossabaker: This comment is obsolete.  We've given up determining that the body is pure.  That's something we did on previous versions of the streaming library where we could introspect the algebra, which is where a lot of of this old complication came from in the first place.

diesalbla:
@rossabaker After digging a bit deeper, I ended up with a change a bit larger that what I started with at first. I have extracted a pure, non-effectful, non-optional function `nextRequest`, which carries the whole process of building the redirection-request from the previous request and response.

client/src/main/scala/org/http4s/client/middleware/FollowRedirect.scala
```diff
@@ -41,51 +41,41 @@ object FollowRedirect {
...
-            // If we get a redirect response without a location, then there is
-            // nothing to redirect.
-            redirectUri.fold(dontRedirect) { nextUri =>
-              // We can only redirect safely if there is no body or if we've
-              // verified that the body is pure.
-              val nextRequest: Request[F] =
-                stripSensitiveHeaders(nextUri).withMethod(method).withUri(nextUri)
+      val req2: Request[F] = req1.withMethod(method).withUri(nextUri)
+      method match {
+        case GET | HEAD => req2.withEmptyBody
```
rossabaker: I find that when I use this numbered idiom, it's easy to add another one and forget the return value.  I think I might prefer the old defs, with descriptive names, to this.  But I like extracting `nextRequest`.

diesalbla: Done. Restored some of the old methods.

rossabaker:
The test failures look related.  I think something broke in the refactoring.

diesalbla:
@rossabaker There was a mistake in composing the operations of `nextRequest`, which I have fixed.

PR #2581, opened 2019-05-18
Remove Redundant Braces.

diesalbla:
Apparently redundant braces around the body of a `case` are not  detected and removed by ScalaFmt


server/src/main/scala/org/http4s/server/middleware/DefaultHead.scala
```diff
@@ -16,14 +16,10 @@ import cats.implicits._
...
     Kleisli { req =>
-      req.method match {
-        case Method.HEAD => {
-          (http <+> headAsTruncatedGet(http))(req)
-        }
-        case _ => {
-          http(req)
-        }
-      }
+      if (req.method == Method.HEAD)
```
rossabaker: I might like the case better here.  I'm unsure whether there's a bytecode difference.

ChristopherDavenport: Definitely prefer the case statement, but agree with removing the extra bracket.

PR #2582, opened 2019-05-19
Default head no inner kleisli

diesalbla:
Removes the use and allocation of extra Kleisli in the private method.
Avoids use of instance `MonoidK[F] => MonoidK[Kleisli[F, A, B]`


rossabaker:
Trivial conflict to resolve and it's ready.

Issue #2560, opened 2019-05-08
Provide ability to return 404 insted of 401 when no route matched


matematyk60:
I have a problem with using auth middleware. I simplified my code and pasted it below. The problem is that given application works, but returns always `401 Unauthorized` when requesting non-existing path (e.g. `/organizations`). What I need is to have `404 Not found` returned (it is returned only when authorization is provided). Did any one of you ever faced such a challange in http4s? Thank you for your help

```
object Main2 extends IOApp {

  sealed trait Error

  case object AnyError extends Error

  private def authUser[F[_]: Monad]: Kleisli[F, Request[F], Either[Error, Header]] = Kleisli { request =>
    Monad[F].pure(request.headers.get(CaseInsensitiveString("Authorization")).right(AnyError))
  }

  private def onAuthFailure[F[_]: Monad]: AuthedService[Error, F] = Kleisli { req =>
    req.authInfo match {
      case AnyError => OptionT.pure(Response[F](Status.Unauthorized, body = stringEncoder[F].toEntity("AUTH_FAILED").body))
    }
  }

  private def authMiddleware[F[_]: Monad]: AuthMiddleware[F, Header] = AuthMiddleware(authUser, onAuthFailure)

  def authedService[F[_]: Monad]: AuthedService[Header, F] = {
    object dsl extends Http4sDsl[F]
    import dsl._
    AuthedService {
      case GET -> Root / "welcome" as header => Ok(s"Header is $header")
    }
  }

  def routes[F[_]: Monad: Sync] = Router(
    "/" -> authMiddleware.apply(authedService)
  )

  override def run(args: List[String]): IO[ExitCode] =
    BlazeServerBuilder[IO]
      .withHttpApp(routes[IO].orNotFound)
      .serve
      .compile
      .drain
      .map(_ => ExitCode.Success)
}
```

rossabaker:
Does [`noSpider`](https://http4s.org/v0.20/auth/) do what you want?

matematyk60:
`noSpider` is possible to use with `authUser` type `Kleisli[OptionT[F, ?], Request[F], Header]`, but I cannot use it with authUser type: `Kleisli[F, Request[F], Either[Error, Header]]`

kevinmeredith:
@matematyk60 - I've run into this scenario on a few `http4s` services. I explicitly applied the auth check only for specific routes. 

I created a minimal repo: https://github.com/kevinmeredith/http4s-auth-404-example.

[...]

PR #2585, opened 2019-05-20
Add New Function to AuthMiddleware to Use Either, not Option, for Determining Response  

kevinmeredith:
Attempts to provide a somewhat cleaner way to address https://github.com/http4s/http4s/issues/2560.

I've intended to show a use case via the unit test that I added.

### DIFF
--- a/server/src/main/scala/org/http4s/server/package.scala
+++ b/server/src/main/scala/org/http4s/server/package.scala
@@ -100,9 +100,9 @@ package object server {
     }
 
     def noSpiderWithEither[F[_]: Monad, E, T](
-                                  authUserWithError: Kleisli[EitherT[F, E, ?], Request[F], T],
-                                  onAuthFailure: (Request[F], E) => F[Response[F]]
-                                ): AuthMiddleware[F, T] = { service =>
+        authUserWithError: Kleisli[EitherT[F, E, ?], Request[F], T],
+        onAuthFailure: (Request[F], E) => F[Response[F]]
+    ): AuthMiddleware[F, T] = { service =>
       Kleisli { r: Request[F] =>
         val resp = authUserWithError(r).value.flatMap {
           case Right(authReq) =>
--- a/server/src/main/scala/org/http4s/server/package.scala
+++ b/server/src/main/scala/org/http4s/server/package.scala
@@ -2,7 +2,7 @@ package org.http4s
 
 import cats._
 import cats.arrow.Choice
-import cats.data.{Kleisli, OptionT}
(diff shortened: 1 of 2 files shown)


server/src/main/scala/org/http4s/server/package.scala
```diff
@@ -98,6 +98,20 @@ package object server {
       }
     }
 
+    def noSpiderWithEither[F[_]: Monad, E, T](
+        authUserWithError: Kleisli[EitherT[F, E, ?], Request[F], T],
```
rossabaker: This is tantalizingly close to something we can express in terms of `MonadError`, but I think we'd need a `G[_]` parameter and a natural transformation `G ~> F` and maybe even back `F ~> G`.  I can't quite make the types click in my head.

kevinmeredith: Thanks for reviewing, Ross. I'm having trouble implementing this suggestion. So I know what the `F` parameter represents, but please let me know the need for the `G`.

kevinmeredith: Well, looking at https://github.com/http4s/http4s/blob/master/server/src/main/scala/org/http4s/server/package.scala#L104-L105, should I update my PR be consistent, namely changing:

>         authUserWithError: Kleisli[EitherT[F, E, ?], Request[F], T],

to

>        authUserWithError: Kleisli[F, Request[F], Either[E, T],

?

rossabaker: I was thinking `Kleisli[G, Request[F], T]`.  I think it would require a `G ~> F` because we don't have a monad transformer typeclass to relate the `G` and `F`.  It might not be worth it.

kevinmeredith: ACK, @rossabaker. Do you think that this change is useful?

I'm asking because it still seems to be a hassle to specify the path in the `val authUserWithError: Kleisli[EitherT[IO, Error, ?], Request[IO], User]`, per my spec, https://github.com/http4s/http4s/pull/2585/commits/a60b2496c7dd8e1ffec92dfc377c24a0ff9ae175/.

It seems simplest to me to auth all requests, regardless of the path, and then return an HTTP-404/401 if that check fails.

As an aside, I'd be interested in the implementation of https://github.com/http4s/http4s/pull/2585#discussion_r287572967!

kevinmeredith:
I closed, then re-opened, to restart the build. I think it failed due to a flaky test.

kevinmeredith:
I'm figuring this change isn't worthwhile to add, so I'll close this one, @rossabaker.

PR #2590, opened 2019-05-21
GZip: map on the F response, not the `Kleisli`.

diesalbla:
A small performance tweak in GZip: instead of doing the `map` on `Kleisli.map`, which would create a new Kleisli every time, we first apply to the request, then `map` on the `F[Response]` directly.


rossabaker:
I'm not very good at counting allocations, but I think the `.map` on `F` gives one allocation back.

diesalbla:
@rossabaker Of course, but the `Kleisli.map` method also uses the `Functor[F].map`.

rossabaker:
Oh, in its implementation.  You're right.

I told you I was bad at counting allocations. :smile:

PR #2604, opened 2019-05-29
Defer creation of SSLContext.getDefault() in the client

### LINKED ISSUE #2431: BlazeClientBuilder: SslContext.getDefault can throw outside the monad
This is another appearance of the problem mentioned in #1956 . On some systems `SslContext.getDefault` can throw, which is not handled correctly. `BlazeClientBuilder` uses `SslContext.getDefault` [as a default argument](https://github.com/http4s/http4s/blob/c104d884706afc2bd73376eb84e3176acc20f4f4/blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala#L195), which leaves capturing the exception to the user.

rossabaker:
`SSLContext.getDefault()` throws on some platforms, which makes it untenable to pass as a default argument to the builders.  We use an `Option` to signify deferred creation.

Hard to test.  We'd need to do some byte code magic to intercept the creation of `SSLContext.default`, or fork a JVM configured such that it fails.  

Fixes #2431.

### DIFF
--- a/blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala
+++ b/blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala
@@ -12,6 +12,10 @@ import org.http4s.internal.BackendBuilder
 import scala.concurrent.ExecutionContext
 import scala.concurrent.duration._
 
+/**
+  * @param sslContext Some custom `SSLContext`, or `None` if the
+  * default SSL context is to be lazily instantiated.
+  */
 sealed abstract class BlazeClientBuilder[F[_]] private (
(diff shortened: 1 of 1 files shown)


kevinmeredith:
> We use an Option to signify deferred creation.

Could you please explain why `Option` indicates deferred creation? When I read "deferred creation," I think of by-name arguments.

ChristopherDavenport:
Due to how implicit resolution works, constraints are always the last arguments so can not be used in default arguments. By suspension we mean like `Sync[F].delay`

rossabaker:
`Later` might be a better encoding when we break compatibility.  Or `SyncIO`.

Issue #2603, opened 2019-05-29
Remove Uri.uri calls in the docs.


rossabaker:
`Uri.uri(...)` is deprecated in favor of a `StringContext` macro.  We should replace all those calls in the tuts with `uri"..."`.

Reported in #2598.  Preventing this in the build in #2602.  Separated as a quick win for someone looking for a first issue.

mcintdan:
May I please be assigned this issue? I am a new contributor and believe I could help fix the documentation. :-)

rossabaker:
@mcintdan :+1: thanks!  Please let us know here or on gitter if you need any help.

rossabaker:
Oops, this is already fixed.  We just neglected to close it.

PR #2607, opened 2019-05-29
replaced Uri.uri(...) with uri(...) in tuts docs

mcintdan:
I hope I interpreted issue #2603  correctly and addressed it per expectations, but please let me know if I should make changes or if there is anything I should do differently next time as I am new to contributing.


mcintdan:
Oh okay, I understand. Sorry about that! I will make the changes.

rossabaker:
No worries.  It's also okay to keep a PR open and push more fixes to it, which keeps the conversation all together.  But here or in another PR is fine for the fix... just happy to have it!

PR #2608, opened 2019-05-29
replaced Uri.uri(...) with uri(...) in tuts docs Issue 2603

mcintdan:
Does this look better for Issue #2603 ? I am sorry if I am creating more work for you, genuinely trying to help, but I do not have a lot of experience with sbt. If this isn't correct, please kindly let me know if you want me to take a step back so someone else with more sbt experience can address the issue. Otherwise, I will take a couple of days to practice and test this out before submitting another PR. Also I am not sure if I needed to actually close the previous PR and create a new one, hope what I did was okay. Thanks for your patience and appreciate all your feedback.  :-)

### DIFF
--- a/docs/src/main/tut/client.md
+++ b/docs/src/main/tut/client.md
@@ -131,7 +131,7 @@ import scala.concurrent.ExecutionContext.Implicits.global
 
 ```tut:book
 def hello(name: String): IO[String] = {
-  val target = Uri.uri("http://localhost:8080/hello/") / name
+  val target = uri("http://localhost:8080/hello/") / name
   httpClient.expect[String](target)
 }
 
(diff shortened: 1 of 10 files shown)


mcintdan:
Awesome, I am so happy it looks okay! Thanks and I look forward to contributing more to this project as I gain experience with http4s.

yoohaemin:
I have a small suggestion since this isn't merged yet: I recall it took me some time to find the correct imports during migration from `Uri.uri` to `uri""`.

Later when I found out, I was a bit surprised that the required import was not `org.http4s.syntax.all._` nor `org.http4s.implicits._` (which is the same thing), but `org.http4s._`. I think it could be worth a small mention in the docs, on where you can find the implicits for the interpolator, like 

>  If you have a literal string, you can use `uri"..."`, which is imported via `import org.http4s._`: 

let me know if you have any better suggestions :)

rossabaker:
I wondered about having these string contexts on `org.http4s.implicits._` when they were added and didn't get generate much discussion either way.  My gut feeling is that's where they belong.  I'm going to spin that off as a separate issue.

Issue #2592, opened 2019-05-21
StaticFile / fromResource: effects not handled. 


diesalbla:
As described in https://github.com/http4s/http4s/pull/2541, the `fromResource` method of the `StaticFile` has some effectual operations (`getResource`) that should be wrapped in `delay`.

rossabaker:
Fixed by cherry-picking #2611.

PR #2611, opened 2019-05-30
StaticFile / fromResource: effects not handled. #2592 

pravin-raha:
Added functional effect to `getResource`.


pravin-raha:
Ok, Thanks.

rossabaker:
Cherry-picked onto series/0.20 as 714f096a62.

PR #2612, opened 2019-05-31
Generalize AuthedRequest to ContextRequest

hamnis:
A ContextRequest is defined as a request with some data.

AuthedRequest can be defined as a ContextRequest with auth data
as the context.

AuthedRequest signals that it should only be used for auth stuff, but
ContextRequest opens up possibility to reuse it to other things.

### DIFF
--- a/core/src/main/scala/org/http4s/ContextRequest.scala
+++ b/core/src/main/scala/org/http4s/ContextRequest.scala
@@ -6,8 +6,9 @@ import cats.data.Kleisli
 
 final case class ContextRequest[F[_], A](context: A, req: Request[F]) {
   def mapK[G[_]](fk: F ~> G): ContextRequest[G, A] =
-    ContextRequest(authInfo, req.mapK(fk))
+    ContextRequest(context, req.mapK(fk))
 
+  @deprecated("Use context instead", "0.21.0")
   def authInfo: A = context
 
 }
--- a/core/src/main/scala/org/http4s/ContextRequest.scala
+++ b/core/src/main/scala/org/http4s/ContextRequest.scala
@@ -1,0 +1,19 @@
+package org.http4s
+
+import cats.syntax.functor._
+import cats.{Functor, ~>}
+import cats.data.Kleisli
+
+final case class ContextRequest[F[_], A](context: A, req: Request[F]) {
+  def mapK[G[_]](fk: F ~> G): ContextRequest[G, A] =
+    ContextRequest(authInfo, req.mapK(fk))
+
+  def authInfo: A = context
+
+}
+
+object ContextRequest {
+  def apply[F[_]: Functor, T](
+      getContext: Request[F] => F[T]): Kleisli[F, Request[F], ContextRequest[F, T]] =
+    Kleisli(request => getContext(request).map(ctx => ContextRequest(ctx, request)))
+}
--- a/core/src/main/scala/org/http4s/ContextRoutes.scala
+++ b/core/src/main/scala/org/http4s/ContextRoutes.scala
@@ -1,0 +1,44 @@
+package org.http4s
+
+import cats.data.{Kleisli, OptionT}
+import cats.Applicative
+import cats.effect.Sync
+
+object ContextRoutes {
+
+  /** Lifts a function into an [[ContextRoutes]].  The application of `run`
+    * is suspended in `F` to permit more efficient combination of
+    * routes via `SemigroupK`.
+    *
+    * @tparam F the effect of the [[ContextRoutes]]
+    * @tparam T the type of the auth info in the [[ContextRequest]] accepted by the [[ContextRoutes]]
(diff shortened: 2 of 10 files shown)


core/src/main/scala/org/http4s/ContextRequest.scala
```diff
@@ -0,0 +1,24 @@
...
+
+final case class ContextRequest[F[_], A](context: A, req: Request[F])
+
+object ContextRequest {
+  def apply[F[_]: Functor, T](
+      getContext: Request[F] => F[T]): Kleisli[F, Request[F], ContextRequest[F, T]] =
+    Kleisli(request => getContext(request).map(ctx => ContextRequest(ctx, request)))
+
+
+  implicit class AuthInfo[F[_], A](req: ContextRequest[F, A]) {
```
rossabaker: This feels awkward to me.  It's always going to be in scope for a `ContextRequest`, so why not just a member of `ContextRequest`?

core/src/main/scala/org/http4s/ContextRequest.scala
```diff
@@ -0,0 +1,24 @@
...
+    def authInfo: A = req.context
+  }
+}
+
+object AuthedRequest {
```
rossabaker: I think I'd keep this companion in its own file.

core/src/main/scala/org/http4s/ContextService.scala
```diff
@@ -0,0 +1,26 @@
+package org.http4s
+
+import cats.data.{Kleisli, OptionT}
+import cats.Applicative
+
+object ContextService {
```
rossabaker: We're moving away from the name "service".  We should consider how this interacts with #2567.

server/src/main/scala/org/http4s/server/package.scala
```diff
@@ -67,6 +67,12 @@ package object server {
   type AuthMiddleware[F[_], T] =
     Middleware[OptionT[F, ?], AuthedRequest[F, T], Response[F], Request[F], Response[F]]
 
+  /**
+    * An HTTP middleware that adds a context.
+    */
+  type ContextMiddleware[F[_], T] =
```
rossabaker: Does anybody actually use these middleware aliases?  I've come to regard them as obfuscation more than anything.  I wonder what others think.

hamnis: We are using them on our project. I find them quite useful.

core/src/main/scala/org/http4s/ContextService.scala
hamnis: I renamed that to ContextRoutes

server/src/main/scala/org/http4s/server/package.scala
RafalSumislawski: We too use them.

hamnis:
@rossabaker @ChristopherDavenport  is this something that is interesting before 0.21 final? or delay for the next bit?

hamnis:
Added a deprecation warning for use of `authInfo` and used context everywhere in the codebase.
Is this something we want to do for 0.21 ? or postpone until later?

core/src/main/scala/org/http4s/ContextRoutes.scala
```diff
@@ -0,0 +1,44 @@
...
+    * @tparam F the effect of the [[ContextRoutes]]
+    * @tparam T the type of the auth info in the [[ContextRequest]] accepted by the [[ContextRoutes]]
+    * @param run the function to lift
+    * @return an [[ContextRoutes]] that wraps `run`
+    */
+  def apply[T, F[_]](run: ContextRequest[F, T] => OptionT[F, Response[F]])(
+      implicit F: Sync[F]): ContextRoutes[T, F] =
+    Kleisli(req => OptionT(F.suspend(run(req).value)))
+
+  /** Lifts a partial function into an [[AuthedRoutes]].  The application of the
```
aeons: There's a couple of `AuthedRoutes` left in the comments.

hamnis: well spotted. will fix

core/src/main/scala/org/http4s/ContextRequest.scala
```diff
@@ -0,0 +1,20 @@
...
+import cats.syntax.functor._
+import cats.{Functor, ~>}
+import cats.data.Kleisli
+
+final case class ContextRequest[F[_], A](context: A, req: Request[F]) {
+  def mapK[G[_]](fk: F ~> G): ContextRequest[G, A] =
+    ContextRequest(context, req.mapK(fk))
+
+  @deprecated("Use context instead", "0.21.0")
+  def authInfo: A = context
```
rossabaker: Seeing how many places this changed just in this repo, it would be a great candidate for a scalafix, but that's not essential to this PR.

PR #2614, opened 2019-06-01
ResponseGenerator: DRY up and relax method implementations

diesalbla:
- DRY out a method `addEntityLenght`, which transforms a set of headers by addding the `Content-Length` of an entity, if the length of that entity is known.
- Relax type-class constraint of several methods from Monad to Applicative.

### DIFF
--- a/dsl/src/main/scala/org/http4s/dsl/impl/ResponseGenerator.scala
+++ b/dsl/src/main/scala/org/http4s/dsl/impl/ResponseGenerator.scala
@@ -2,11 +2,20 @@ package org.http4s
 package dsl
 package impl
 
-import cats.{Applicative, Monad}
+import cats.Applicative
 import org.http4s.headers._
+import ResponseGenerator.addEntityLength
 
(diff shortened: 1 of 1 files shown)


dsl/src/main/scala/org/http4s/dsl/impl/ResponseGenerator.scala
```diff
@@ -40,29 +49,16 @@ trait EntityResponseGenerator[F[_], G[_]] extends Any with ResponseGenerator {
 
   def apply[A](body: G[A])(implicit F: Monad[F], w: EntityEncoder[G, A]): F[Response[G]] = {
```
TomTriple:
```suggestion
  def apply[A](body: G[A])(implicit F: Applicative[F], w: EntityEncoder[G, A]): F[Response[G]] = {
```

diesalbla: I have added the change.

PR #2615, opened 2019-06-02
Model Server header

cquiroz:
This is a model for the Server header as for #2011. `User-Agent` and `Server` parse the same way so I unified the parsers and component classes:
`AgentProduct` -> `HeaderProduct`
`AgentToken` -> `HeaderToken`
`AgentComment` -> `HeaderComment`

I'm not sure if this is the right approach an it breaks mima. Feel free to comment

Should Blaze return a server header like
`Server: blaze-http4s/0.21.0`?

### DIFF
--- a/core/src/main/scala/org/http4s/headers/User-Agent.scala
+++ b/core/src/main/scala/org/http4s/headers/User-Agent.scala
@@ -2,38 +2,29 @@ package org.http4s
 package headers
 
 import org.http4s.parser.HttpHeaderParser
-import org.http4s.util.{Renderable, Writer}
+import org.http4s.util.Writer
 
 object `User-Agent` extends HeaderKey.Internal[`User-Agent`] with HeaderKey.Singleton {
+  def apply(id: ProductId): `User-Agent` =
+    new `User-Agent`(id, Nil)
+
   override def parse(s: String): ParseResult[`User-Agent`] =
     HttpHeaderParser.USER_AGENT(s)
 }
 
-sealed trait AgentToken extends Renderable
-
-final case class AgentProduct(name: String, version: Option[String] = None) extends AgentToken {
-  override def render(writer: Writer): writer.type = {
-    writer << name
-    version.foreach { v =>
-      writer << '/' << v
-    }
-    writer
-  }
-}
-final case class AgentComment(comment: String) extends AgentToken {
-  override def renderString: String = comment
-  override def render(writer: Writer): writer.type = writer << comment
-}
-
-final case class `User-Agent`(product: AgentProduct, other: List[AgentToken] = Nil)
+/**
+  * User-Agent header
+  * https://tools.ietf.org/html/rfc7231#section-5.5.3
+  */
+final case class `User-Agent`(product: ProductId, rest: List[ProductIdOrComment])
     extends Header.Parsed {
   def key: `User-Agent`.type = `User-Agent`
 
   override def renderValue(writer: Writer): writer.type = {
     writer << product
-    other.foreach {
-      case p: AgentProduct => writer << ' ' << p
-      case AgentComment(c) => writer << ' ' << '(' << c << ')'
+    rest.foreach {
+      case p: ProductId => writer << ' ' << p
+      case ProductComment(c) => writer << ' ' << '(' << c << ')'
     }
     writer
   }
--- a/core/src/main/scala/org/http4s/headers/package.scala
+++ b/core/src/main/scala/org/http4s/headers/package.scala
@@ -48,4 +48,13 @@ package object headers {
             .append(idMostSigBits.toHexString)
         }
     }
+
+  @deprecated("Deprecated in favor of HttpToken", "0.18")
+  type AgentToken = ProductIdOrComment
+
+  @deprecated("Deprecated in favor of HttpComment", "0.18")
+  type AgentComment = ProductComment
+
+  @deprecated("Deprecated in favor of HttpProduct", "0.18")
+  type AgentProduct = ProductId
 }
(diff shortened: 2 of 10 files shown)


rossabaker:
We're calling it `HeaderProduct` to avoid a clash with `scala.Product`?  I kind of just want to call it `Product` to maintain correspondence with the RFCs, but I'm not sure about unintended consequences.  The `Token` type appears frequently in the spec and could be used in a lot of places.

We haven't been very principled about when these model types are in `org.http4s` or `org.http4s.headers`.  For example, `MediaRange` and `TransferCoding` and `LanguageTag` are all things that appear in headers but are in the base package.  Why not these?

We could add deprecated type aliases and vals to help people move.

There should be a `Server` header option on `BlazeServerBuilder`.  Whether it defaults to `Some(http4s-blaze/0.21.0)` (accurate! branding!) or `None` (security by default!) is an open question.

cquiroz:
Yes, `Product` sounds too generic and `AgentProduct` is (I think) too close to `User-Agent`. Same with `Token` I think it is too generic

rossabaker:
Should we move them to the root package?  I think probably yes.  Most of the "values" are in the root.

`Http` might be a good prefix as well for things that are too generic without.

cquiroz:
Rebased and updated as suggested

core/src/main/scala/org/http4s/HttpToken.scala
```diff
@@ -0,0 +1,20 @@
+package org.http4s
+
+import org.http4s.util.{Renderable, Writer}
+
+sealed trait HttpToken extends Renderable
```
rossabaker: I count 27 places `token` appears in the HTTP RFCs.  I'm thinking that maybe this is a value rather than a sealed type.

cquiroz: I converted `HttpToken` to an `abstract class`. I'm not sure if that is what you have in mind

cquiroz:
Bump

core/src/main/scala/org/http4s/HttpToken.scala
```diff
@@ -0,0 +1,21 @@
...
+  override def render(writer: Writer): writer.type = {
+    writer << value
+    version.foreach { v =>
+      writer << '/' << v
+    }
+    writer
+  }
+}
+
+final case class HttpComment(override val value: String) extends HttpToken(value) {
```
rossabaker: Why is a comment a token?  It's a different thing.

```
     token          = 1*tchar
     comment        = "(" *( ctext / quoted-pair / comment ) ")"
```

I guess a sum type is needed for the `product / comment` rule.  I think `HttpToken` is maybe misleading.  `ProductOrComment` comes to mind, but is verbose.  Union types come to mind, but we're not on dotty yet!

> The User-Agent field-value consists of one or more product identifiers, each followed by zero or more comment

When thought of that way, maybe this model becomes useful:

```scala
case class Product(value: String, version: Option[String], comments: List[String])
```

cquiroz: Trying to rename it to `Product` leads to quite many conflicts on the code base with `scala.Product`. Maybe we should go back to `ProductOrComment`

rossabaker: Okay, I'm good with `ProductOrComment`.  I don't particularly like it, but the RFC doesn't give it a more sensible name.

cquiroz:
I'm converging to this encoding. WDYT?

```scala
final case class ProductIdentifier(value: String, version: Option[String] = None)

final case class ProductOrComment(value: Either[ProductIdentifier, String]) extends Renderable {

final case class `User-Agent`(product: List[ProductOrComment]) 
```

rossabaker:
I like your `ProductIdentifier` idea to disambiguate from a `scala.Product`. I think making the sum type an `Either` in a unary case class is odd, and a `User-Agent` must start with one `ProductId`.

```scala
sealed trait ProductIdOrComment
final case class ProductId(value: String, version: Option[String] = None) extends ProductIdOrComment
final case class ProductComment(value: String) extends ProductIdOrComment
final case class `User-Agent`(product: ProductId, rest: List[ProductIdOrComment])
```

That `User-Agent` could nearly be a `NonEmptyList[ProductOrComment]`, but the head _must_ be a `ProductId`.  It's just kind of an ugly spec.

cquiroz:
I rebased and squashed to make it easier to review. It seems there is a flaky test on blaze client?

core/src/main/scala/org/http4s/headers/User-Agent.scala
```diff
@@ -2,38 +2,29 @@ package org.http4s
 package headers
 
 import org.http4s.parser.HttpHeaderParser
-import org.http4s.util.{Renderable, Writer}
+import org.http4s.util.Writer
 
 object `User-Agent` extends HeaderKey.Internal[`User-Agent`] with HeaderKey.Singleton {
+  def apply(id: ProductId): `User-Agent` =
```
rossabaker: Shouldn't `Server` have this same constructor?  Or just make a default arg of `Nil`?

core/src/main/scala/org/http4s/headers/package.scala
```diff
@@ -48,4 +48,13 @@ package object headers {
             .append(idMostSigBits.toHexString)
         }
     }
+
+  @deprecated("Deprecated in favor of HttpToken", "0.18")
```
rossabaker: We haven't conclusively decided on the next version, but I'd call it 0.22 for now.

cquiroz:
I added the suggested comments

cquiroz:
Finally 😉

PR #2620, opened 2019-06-04
WIP: Add Option to have known content-length

LukaJCB:
See gitter conversation https://gitter.im/http4s/http4s?at=5cf5845b6fc5846bab5840b7

This doesn't work right now, I think my math is off somewhat, but just wanted to get some initial feedback on the code structure. :)

### DIFF
--- a/core/src/main/scala/org/http4s/multipart/Multipart.scala
+++ b/core/src/main/scala/org/http4s/multipart/Multipart.scala
@@ -17,13 +17,13 @@ final case class Multipart[F[_]](parts: Vector[Part[F]], boundary: Boundary = Bo
 
   def contentLength: Option[Long] = {
     val newlineLength = 2L
-    parts.traverse(_.contentLength).map { lengths =>
-      val bodiesLength = lengths.combineAll
+    parts.traverse(part => part.fullLength).map { lengths =>
+      val partsLength = lengths.combineAll
       val numberOfParts = lengths.size
-      val boundariesLength = (numberOfParts + 1) * boundary.length
+      val boundariesLength = (numberOfParts + 1) * Boundary.lengthInBytes
       val boundariesNewLinesLength = numberOfParts * newlineLength
 
-      bodiesLength + boundariesLength + boundariesNewLinesLength
+      partsLength + boundariesLength + boundariesNewLinesLength
     }
   }
 }
--- a/core/src/main/scala/org/http4s/multipart/Multipart.scala
+++ b/core/src/main/scala/org/http4s/multipart/Multipart.scala
@@ -1,14 +1,29 @@
 package org.http4s
 package multipart
 
+import cats.implicits._
 import org.http4s.headers._
 
 final case class Multipart[F[_]](parts: Vector[Part[F]], boundary: Boundary = Boundary.create) {
   def headers: Headers =
     Headers(
       List(
-        `Transfer-Encoding`(TransferCoding.chunked),
+        contentLength
+          .flatMap(len => `Content-Length`.fromLong(len).toOption)
+          .getOrElse(`Transfer-Encoding`(TransferCoding.chunked)),
         `Content-Type`(MediaType.multipartType("form-data", Some(boundary.value)))
       )
     )
+
+  def contentLength: Option[Long] = {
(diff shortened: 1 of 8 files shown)


core/src/main/scala/org/http4s/multipart/Part.scala
```diff
@@ -29,10 +34,41 @@ object Part {
...
+      path: Path,
+      blockingEC: ExecutionContext,
+      headers: Header*): F[Part[F]] =
+    fileDataKnownContentLength(
+      name,
+      path.toFile.getName,
+      readAll[F](path, blockingEC, ChunkSize),
+      headers: _*)
+
+  def fileDataKnownContentLength[F[_]: Sync](
```
Daenyth: I think these need a warning that they will hold all of `entityBody`/`path` (respectively) in memory; the "default" assumption is usually the opposite.

LukaJCB: Yes, 100% agree

core/src/main/scala/org/http4s/multipart/Multipart.scala
```diff
@@ -1,14 +1,29 @@
...
+        contentLength
+          .flatMap(len => `Content-Length`.fromLong(len).toOption)
+          .getOrElse(`Transfer-Encoding`(TransferCoding.chunked)),
         `Content-Type`(MediaType.multipartType("form-data", Some(boundary.value)))
       )
     )
+
+  def contentLength: Option[Long] = {
+    val newlineLength = 2L
+    parts.traverse(_.contentLength).map { lengths =>
```
rossabaker: This isn't accounting for the parts' headers' lengths, is it?

rossabaker: Where are you accounting for the parts' headers' lengths?

core/src/main/scala/org/http4s/multipart/Part.scala
```diff
@@ -29,10 +34,41 @@ object Part {
...
+
+  def fileDataKnownContentLength[F[_]: Sync: ContextShift](
+      name: String,
+      path: Path,
+      blockingEC: ExecutionContext,
+      headers: Header*): F[Part[F]] =
+    fileDataKnownContentLength(
+      name,
+      path.toFile.getName,
+      readAll[F](path, blockingEC, ChunkSize),
```
rossabaker: How about `F.delay(Files.size(path))` and flatMap that shit?  I guess there's a slight risk the file size changes between being sized and being read.

core/src/main/scala/org/http4s/multipart/Multipart.scala
LukaJCB: D'oh! Any pointers on the easiest way to get the length of a header? :)

core/src/main/scala/org/http4s/multipart/Part.scala
LukaJCB: Sounds pretty reasonable, though we can't really do that when all we have is an `entityBody`, do we?

core/src/main/scala/org/http4s/multipart/Multipart.scala
rossabaker: Yes, #2622!  Go ahead and cherry-pick that if you get back to this before it's approved.

core/src/main/scala/org/http4s/multipart/Part.scala
rossabaker: I think I wouldn't implement it in terms of the other.  I'm still a little scared of reading a growing file, where the size can become a lie.  And the javadoc is frightening:

> Returns the size of a file (in bytes). The size may differ from the actual size on the file system due to compression, support for sparse files, or other reasons. The size of files that are not regular files is implementation specific and therefore unspecified.

We could `.take(n)` bytes from the stream we read to guarantee we don't go over, but this would silently truncate things.  Maybe we have one like I'm proposing that's optimal, but with appropriate disclaimers, and then a buffering one like what you already wrote?

LukaJCB: Alright, thanks for the advice, will do :)

PR #2623, opened 2019-06-05
Propagate cookies in FollowRedirect client middleware.

rpiaggio:
Some websites set cookies when issuing a redirect, which are expected to be present when invoking the new URL. This PR takes care of these cases.


rpiaggio:
Hi @rossabaker ! Thank you for you review. Unfortunately, due to the nature of cookies that are set and sent in different headers (`Set-Cookie` and `Cookie`), special logic is necessary to propagate them.

My redirect wasn't even going across servers so the sensitive header logic wasn't the issue, it was not kicking in.

However, due to the sensitive nature of cookies, I did add now logic to prevent the propagation of cookies across servers.

Maybe `stripSensitiveHeaders` and `propagateCookies` could be combined into a single `fixHeaders`  (or similarly named) method?

rpiaggio:
Having a cookie JAR would definitely be helpful in a larger scale. It's not clear to me either the best way to implement it, since it may be needed both inside and outside the redirect middleware.

I'm using http4s to parse some pages and this PR solves for me the specific problem of parsing pages that redirect to themselves but won't render properly unless a cookie is present that the actual page sets in the redirect (this is the case at https://tickantel.com.uy/inicio/buscar_categoria?cat_id=1, for example). In order to keep the client from falling in an infinite loop, cookies have to be propagated across redirects. As long as http4s doesn't support this, I am forced to provide my own redirect middleware (the one in this PR).

I hope other people can find this functionality helpful while there is no support for a cookie jar. The logic can always be removed later from the redirect middleware when a more global solution is available.

rpiaggio:
I didn't know about the existing CookieJar functionality. In light of this, I admit my implementation to be a bit naïve.

One possible way to go forward: If there was a CookieJarClient which maintained a mutable CookieJar, we could require the client passed to FollowRedirect to be a CookieJarClient and use its jar's cookies for requests and submit to it new cookies from responses. In other words, a FollowRedirect middleware would always have to be paired to a CookieJarClient.

In the meantime, I've updated the PR. It does solve the problem for me (and it seems it would for other people too) in the absence of a CookieJarClient.

rossabaker:
So yes, if we merge this, I think it fixes the problem.  There are compile errors after the last merge to address first.  I'm inclined to merge this and sort out the issues below once we have a working cookie jar middleware, which @ChristopherDavenport [gisted](https://gist.github.com/ChristopherDavenport/048cd1eb5325af121629fb9dd5131015) today.

1. If the cookie jar middleware is outside follow redirect, it won't see cookies set on intermediate redirect responses followed by `FollowRedirect`.  We'd want to be sure to propagate those cookies to the cookie jar for subsequent requests.  To do that, `FollowRedirect` would need to be able to add to the cookie jar, or we'd need to put the cookie jar middleware inside follow redirect.

2. If the cookie jar middleware is inside follow redirect, the logic added here risks duplicating the cookie.  We would either need to assume that there is a cookie jar on the outside, or add deduplication logic to `request.addCookie`.

rpiaggio:
Compile error fixed.

ChristopherDavenport:
I worry about adding this, as if we do get a middleware and this isn't behind a flag on the middleware we will see the results you listed above, any way we could make this a param whether to enable this behavior or not?

rossabaker:
@rpiaggio had earlier proposed

> Maybe stripSensitiveHeaders and propagateCookies could be combined into a single fixHeaders (or similarly named) method?

`stripSensitiveHeaders` exists elsewhere, and `propagateCookies` is translating response headers to a request, so I think it makes sense to keep them separate.  I think an `Option[CookieJar]` would be an appropriate config as soon as we have that PR?

I am in favor of this because it fixes a problem we don't have another solution for, with the understanding that a breaking change is likely once the more general solution arrives.

PR #2624, opened 2019-06-06
Don't depend on laziness of SSLContext in blaze-client

rossabaker:
Changes the `Option[SSLContext]` to be truly optional instead of a sneaky laziness in the innards of the client.

The default remains `Some(SSLContext.getDefault())`, except on systems where this fails, in which case the error is caught and SSL is disabled.  Any client may now opt out of SSL support by calling `withoutSslContext`.  Making an https call on a client where SSL is disabled now results in an `IllegalStateException` with suggestions how to configure it, rather than throwing whatever exception may come from `SSLContext.getDefault()`.

Refines the changes in #2604.  Is binary compatible, but changes the behavior of weird cases.

### DIFF
--- a/blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala
+++ b/blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala
@@ -128,7 +128,7 @@ sealed abstract class BlazeClientBuilder[F[_]] private (
 
   /** Disable secure calls */
   def withoutSslContext: BlazeClientBuilder[F] =
-    withDefaultSslContext
+    copy(sslContext = None)
 
   def withCheckEndpointAuthentication(checkEndpointIdentification: Boolean): BlazeClientBuilder[F] =
     copy(checkEndpointIdentification = checkEndpointIdentification)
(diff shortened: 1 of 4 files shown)


rossabaker:
@nigredo-tori might also be interested in this as the reporter of the original bug, twice.

nigredo-tori:
@rossabaker, :thumbsup: from me.

rossabaker:
The [test failure](https://travis-ci.org/http4s/http4s/jobs/542376204#L1077) is real.  Something about Scala 2.13 or the new Specs2 version we use in it.

blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala
```diff
@@ -112,21 +113,20 @@ sealed abstract class BlazeClientBuilder[F[_]] private (
...
 
-  @deprecated("Use `withSslContext` or `withDefaultSslContext`", "0.20.2")
+  /** Use some provided `SSLContext` when making secure calls, or disable secure calls with `None` */
   def withSslContextOption(sslContext: Option[SSLContext]): BlazeClientBuilder[F] =
     copy(sslContext = sslContext)
 
-  @deprecated("Use `withDefaultSslContext`", "0.20.2")
+  /** Disable secure calls */
   def withoutSslContext: BlazeClientBuilder[F] =
     withDefaultSslContext
```
madanepuri: should this be `copy(sslContext = None)`?

rossabaker: You are absolutely right!  Good catch.

PR #2625, opened 2019-06-06
Change MultipartParser to drain the multipart epilogue

### LINKED ISSUE #2610: Weird multipart/chunked parsing bug
I don't know how I can shrink this further... This seems to be some race/interplay between multipart and chunked encoding parsing, but I'm not sure.

Test case:
```scala
import cats.effect._
import cats.implicits._
import fs2.{Chunk, Stream}
import java.util.concurrent.Executors
import org.http4s._
import org.http4s.client.blaze.BlazeClientBuilder
import org.http4s.implicits._
import org.http4s.multipart._
import org.http4s.server.blaze.BlazeServerBuilder
import org.specs2.execute.Result

class WeirdBugSpec extends org.specs2.mutable.Specification {

  private val log = org.slf4j.LoggerFactory.getLogger(getClass)

  val cpuEC = ExecutionContext.fromExecutorService(
    Executors.newWorkStealingPool
  )

  implicit val cs = IO.contextShift(cpuEC)
  implicit val timer = IO.timer(cpuEC)

  "Weird bug test case" >> {

    val numRequests = 100
    val responseDelay = 300.milliseconds

    val expectedResponse = Chunk.bytes("expected response".getBytes)

    val routes: HttpRoutes[IO] = {
      val dsl = Http4sDsl[IO]
      import dsl._

      HttpRoutes.of[IO] {
        case req @ POST -> Root =>
          req.decode[Multipart[IO]] { mp =>
            timer.sleep(responseDelay) *>
              // Has to be a `Stream` to trigger chunked encoding.
              Ok(Stream.chunk[IO, Byte](expectedResponse))
          }
      }
    }

    val mkServer = {
      BlazeServerBuilder[IO]
        .bindHttp(8080, "0.0.0.0")
        .withHttpApp(routes.orNotFound)
        .resource
    }

    val mkClient = BlazeClientBuilder[IO](cpuEC).resource

    val body = Multipart(
      Vector(
        Part.fileData[IO]("foo", "foo", Stream.emits("This is foo".getBytes)),
      ),
      // fewer moving parts this way
      boundary = Boundary("MwjRT0CXPTcG0AVzl/10g_uhN3Z7zk:IIJXJCvh/6")
    )

    val req =
      Request[IO](
        method = Method.POST,
        uri = Uri.unsafeFromString("http://localhost:8080")
      ).withEntity(body)
        // https://github.com/http4s/http4s/issues/146
        .withHeaders(body.headers)

    mkServer.use { _ =>
      mkClient.use { client =>
        val checkOnce = client.expect[Chunk[Byte]](req)
          .flatMap { result => IO {
            result must_=== expectedResponse
          }}.onError { case err => IO {
            log.error("Client error", err)
          }}

      (1 to numRequests).toVector
        .traverse(_ => checkOnce)
        .map(Result.forall(_)(_.toResult))
      }
    }.unsafeRunSync
  }
}
```
This fails with an error:
```
[blaze-selector-1] INFO  o.h.b.c.n.NIO1SocketServerGroup - Accepted connection from /127.0.0.1:53504 
[scala-execution-context-global-1794] INFO  o.h.s.b.Http1ServerStage$$anon$1 - HTTP body not read to completion. Dropping connection. 
[ForkJoinPool-38-worker-4] ERROR WeirdBugSpec - Client error 
org.http4s.InvalidBodyException: Received premature EOF.
```
The exchanges between the server and the client look like shown below, until the server just decides to reset after sending the headers.
```http
POST / HTTP/1.1
Host: localhost:8080
Content-Type: multipart/form-data; boundary="MwjRT0CXPTcG0AVzl/10g_uhN3Z7zk:IIJXJCvh/6"
Accept: */*
User-Agent: http4s-blaze/0.20.1
Transfer-Encoding: chunked

2d
--MwjRT0CXPTcG0AVzl/10g_uhN3Z7zk:IIJXJCvh/6

5f
Content-Disposition: form-data; name="foo"; filename="foo"
Content-Transfer-Encoding: binary

2


b
This is foo
2f

--MwjRT0CXPTcG0AVzl/10g_uhN3Z7zk:IIJXJCvh/6--
0

HTTP/1.1 200 OK
Date: Thu, 30 May 2019 11:51:25 GMT
Transfer-Encoding: chunked

11
expected response
0
```
Environment: http4s 0.20.1, Scala 2.12.8, OpenJDK 1.8.0_212 x64, Ubuntu 18.04.2 LTS.

nigredo-tori:
Fixes #2610 

This is a slightly hacky (although technically correct) way to make sure the `chunked` parser pulls the terminating zero-length "chunk" when working with the `MultipartParser`. See #2610 for more information. This also includes a test for #2610, but I only got it to work (that is, to fail on `master`) with Java 8. With Java 11 everything seems to work OK as is. Not sure what that's about.

I'm also not sure I got all the points where `MultipartParser` finishes parsing - I'd appreciate if someone familiar with it took a look. That whole file looks completely alien to me. :smile:

### DIFF
--- a/blaze-server/src/test/scala/org/http4s/server/blaze/BlazeServerSpec.scala
+++ b/blaze-server/src/test/scala/org/http4s/server/blaze/BlazeServerSpec.scala
@@ -2,6 +2,7 @@ package org.http4s
 package server
 package blaze
 
+import cats.implicits._
 import cats.effect.IO
 import java.net.{HttpURLConnection, URL}
 import java.nio.charset.StandardCharsets
@@ -9,6 +10,8 @@ import org.http4s.blaze.channel.ChannelOptions
 import org.http4s.dsl.io._
 import scala.concurrent.duration._
 import scala.io.Source
+import org.specs2.execute.Result
+import org.http4s.multipart.Multipart
 
 class BlazeServerSpec extends Http4sSpec {
 
@@ -30,6 +33,11 @@ class BlazeServerSpec extends Http4sSpec {
 
     case _ -> Root / "never" =>
       IO.never
+
+    case req @ POST -> Root / "issue2610" =>
+      req.decode[Multipart[IO]] { mp =>
+        Ok(mp.parts.foldMap(_.body))
+      }
 
     case _ => NotFound()
   }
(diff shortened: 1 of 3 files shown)


nigredo-tori:
The failure is due to AuthedRequest.scala not being properly formatted, which has nothing to do with this MR.

rossabaker:
Cherry-picked to series/0.20 as d6c503bfe71c3cfab33f70f1159698018da48ece.

PR #2638, opened 2019-06-12
Pass RedactHeadersWith from RequestLogger

ChristopherDavenport:
Correct Branch this time


ChristopherDavenport:
Suggestion for 0.21, what if we removed the default from `Logger.logMessage` that way we are passing something whenever we call this method.

PR #2645, opened 2019-06-13
Fix string literal in Router example

knuton:
The syntax error causes confusing noise in the docs.


rossabaker:
Hmm, why are we using `nofail` there?  Would that example not compile for other reasons?

ChristopherDavenport:
@rossabaker Yeah, the services it references don't exist.

PR #2648, opened 2019-06-16
Make generateMimeDb an sbt task instead of a project

rossabaker:
Run with `core/generateMimeDb`

Pros:
* It's only needed on one Scala version
* This is blocking #2493 
* We can get rid of that `thirteen` nonsense now that boopickle is published and this is no longer a module
* `core/generateMimeDb` is a fairly reasonable way to generate it.

Cons:
* Puts a lot on the sbt classpath, including an older version of ourselves

Mitigation:
* It doesn't have to use circe or http4s, but I'd rather lazily port to something we like less if it bites.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -34,7 +34,6 @@ lazy val modules: List[ProjectReference] = List(
   scalaXml,
   twirl,
   scalatags,
-  mimedbGenerator,
   bench,
   examples,
   examplesBlaze,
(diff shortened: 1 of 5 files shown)


rossabaker:
Oh, it'd be neat if the task scalafmt'ed it, too.  I didn't even think about that.  At least this unblocks 2.13.

cquiroz:
It sounds a good idea to format after generation. I could give it a shot

PR #2655, opened 2019-06-18
Tune JVM options for throughput

rossabaker:
A shameless ripoff of https://github.com/typelevel/cats/pull/2797.  Our Travis release is running out of steam after about 45 minutes.  Maybe this will help.  Maybe it'll even fix the blaze flickering!

### DIFF
--- a/.jvmopts
+++ b/.jvmopts
@@ -1,3 +1,6 @@
--Xms1536M
--Xmx1536M
--Xss6M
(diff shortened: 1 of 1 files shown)


rossabaker:
Not much of a sample size yet vs. the random noise in Travis. I think only time will tell whether this is a good idea.

PR #2661, opened 2019-06-20
Move string context macros to implicits

### LINKED ISSUE #2616: String contexts for literals: where should they live?
Regarding our string context macros (e.g., `uri`), @yoohaemin brings up on #2608:

> I was a bit surprised that the required import was not org.http4s.syntax.all._ nor org.http4s.implicits._ (which is the same thing), but org.http4s._

They're a syntactic extension, so putting them in implicits fits my intiution.  And I like our imports to fit in well with the Typelevel ecosystem, but there isn't a lot of precedent.

This issue is to discuss, before we write more: did we put them in the right place?

rossabaker:
Fixes #2616.


rossabaker:
I would like to add a scalafix, but could not quickly find an exmaple that adds an import.  Rewriting to an explicit invocation of the string context is plausible, but, ugh.  Does anybody have a model for this?

Issue #2528, opened 2019-04-25
Blaze Client: Premature EOF in 0.20.0 or later


hamnis:
Some change in the Blaze client between 0.20.0-M7 and 0.20.0 final broke our code. I am currently investigating what the problem is, but it seems like the connection to upstream is prematurely closed. The same code was working on 0.20.0-M7.

The code in question is en/decrypting an fs2 stream on the fly.

Changing to http4s-async-http-client fixed the problem. Let me know if you need more information.

rossabaker:
Do you have a reproducible example?  I'd like to see how the client is being used.

hamnis:
The code is proprietary, unfortunately.

We are doing a GET to a Openshift Swift file repository.
Then we are decrypting the stream on the fly.

Nothing major else is going on. The code works fine using the async-httpclient

hamnis:
The code below is a minimized program of what fails.
download1 works correctly, but download2 fails with EOF.
This works using async-http-client (http4s 0.20.3)
jetty-client seems to have a different problem, will get back to that later.

It seems to have something to do with file size. The dropbox link has a 400 MB file of gibberish.

```scala
package main

import java.nio.file.{Paths, StandardOpenOption}

import cats.effect.{ExitCode, IO, IOApp}
import cats.implicits._
import fs2._
import fs2.io._
import org.asynchttpclient.DefaultAsyncHttpClientConfig
import org.http4s._
import org.http4s.client.Client
import org.http4s.client.asynchttpclient.AsyncHttpClient
import org.http4s.client.blaze.BlazeClientBuilder

import scala.concurrent.ExecutionContext
import scala.concurrent.duration._

object Download extends IOApp {
  val URI =
    uri"https://www.dropbox.com/s/dl/gxu1rfogbn288gh/gibberish.txt"

  override def run(args: List[String]): IO[ExitCode] = {
    val r = for {
      client <- AsyncHttpClient.resource[IO](
                 new DefaultAsyncHttpClientConfig.Builder(AsyncHttpClient.defaultConfig)
                   .setReadTimeout(5.minutes.toMillis.toInt)
                   .setRequestTimeout(5.minutes.toMillis.toInt)
                   .build()
               )
      client2 <- BlazeClientBuilder[IO](ExecutionContext.global)
                  .withIdleTimeout(5.minutes)
                  .withRequestTimeout(5.minutes)
                  .withResponseHeaderTimeout(1.minute)
                  .resource
    } yield client
    r.use { client =>
[...]

PR #2663, opened 2019-06-21
DRAFT: HttpResource

rossabaker:
Introduces an `HttpResource`, which is an `Http[Resource[F, ?], F]` based on @nigredo-tori's idea in #2528. A resource acquired in `response.body` is too late for the response header, and a resource released in `response.body`'s finalizer may too easily be skipped.  This gives the backend a chance to manage a resource for the entire lifecycle of the response.

Implemented for the blaze-server backend.  `HttpApp` is `mapK`'ed into it, so everybody pays for this even though relatively few might use it.

Other underexplored thoughts:
* The 2-parameter DSL may come in handy here.
* Haven't begun to explore how this interacts with `HttpRoutes`
* We could maybe provide functions to treat `F[Response[F]]` more like a `Resource` instead, but it's easy to not use them.

### DIFF
--- a/blaze-server/src/main/scala/org/http4s/server/blaze/Http1ServerStage.scala
+++ b/blaze-server/src/main/scala/org/http4s/server/blaze/Http1ServerStage.scala
@@ -2,7 +2,8 @@ package org.http4s
 package server
 package blaze
 
-import cats.effect.{CancelToken, ConcurrentEffect, IO, Sync, Timer}
+import cats.effect.{CancelToken, ConcurrentEffect, IO, Timer}
+import cats.effect.implicits._
 import cats.implicits._
 import java.nio.ByteBuffer
 import java.util.concurrent.TimeoutException
@@ -26,7 +27,7 @@ import io.chrisdavenport.vault._
 private[blaze] object Http1ServerStage {
 
   def apply[F[_]](
-      routes: HttpApp[F],
+      routes: HttpResource[F],
       attributes: () => Vault,
       executionContext: ExecutionContext,
       enableWebSockets: Boolean,
@@ -66,7 +67,7 @@ private[blaze] object Http1ServerStage {
 }
 
 private[blaze] class Http1ServerStage[F[_]](
-    httpApp: HttpApp[F],
+    httpResource: HttpResource[F],
     requestAttrs: () => Vault,
     implicit protected val executionContext: ExecutionContext,
     maxRequestLineLen: Int,
@@ -79,8 +80,8 @@ private[blaze] class Http1ServerStage[F[_]](
     extends Http1Stage[F]
     with TailStage[ByteBuffer] {
 
-  // micro-optimization: unwrap the routes and call its .run directly
-  private[this] val runApp = httpApp.run
+  // micro-optimization: unwrap the HTTP resource and call its .run directly
+  private[this] val run = httpResource.run
 
   // protected by synchronization on `parser`
   private[this] val parser = new Http1ServerParser[F](logger, maxRequestLineLen, maxHeadersLen)
@@ -178,10 +179,13 @@ private[blaze] class Http1ServerStage[F[_]](
       case Right(req) =>
         executionContext.execute(new Runnable {
           def run(): Unit = {
-            val action = Sync[F]
-              .suspend(raceTimeout(req))
-              .recoverWith(serviceErrorHandler(req))
-              .flatMap(resp => F.delay(renderResponse(req, resp, cleanup)))
+            val action = raceTimeout(req).continual {
+              case Right((resp, release)) =>
+                F.delay(renderResponse(req, resp, cleanup)).guarantee(release)
+              case Left(t) =>
+                serviceErrorHandler(req)(t).flatMap(resp =>
+                  F.delay(renderResponse(req, resp, cleanup)))
+            }
 
             parser.synchronized {
               cancelToken = Some(
@@ -332,13 +336,12 @@ private[blaze] class Http1ServerStage[F[_]](
     renderResponse(req, resp, bodyCleanup) // will terminate the connection due to connection: close header
   }
 
-  private[this] val raceTimeout: Request[F] => F[Response[F]] =
+  private[this] def raceTimeout(req: Request[F]): F[(Response[F], F[Unit])] =
     responseHeaderTimeout match {
       case finite: FiniteDuration =>
-        val timeoutResponse = timer.sleep(finite).as(Response.timeout[F])
-        req =>
-          F.race(runApp(req), timeoutResponse).map(_.merge)
+        val timeoutResponse = timer.sleep(finite).as((Response.timeout[F], F.unit))
+        F.race(run(req).allocated, timeoutResponse).map(_.merge)
       case _ =>
-        runApp
+        run(req).allocated
     }
 }
--- a/blaze-server/src/main/scala/org/http4s/server/blaze/BlazeServerBuilder.scala
+++ b/blaze-server/src/main/scala/org/http4s/server/blaze/BlazeServerBuilder.scala
@@ -27,6 +27,7 @@ import org.http4s.blaze.pipeline.LeafBuilder
 import org.http4s.blaze.pipeline.stages.SSLStage
 import org.http4s.blaze.util.TickWheelExecutor
 import org.http4s.blazecore.{BlazeBackendBuilder, tickWheelResource}
+import org.http4s.internal.resourceLiftK
 import org.http4s.server.ServerRequestKeys
 import org.http4s.server.SSLKeyStoreSupport.StoreInfo
 import org.http4s.util.threads.threadFactory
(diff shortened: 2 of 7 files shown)


blaze-server/src/main/scala/org/http4s/server/blaze/Http1ServerStage.scala
```diff
@@ -332,13 +336,12 @@ private[blaze] class Http1ServerStage[F[_]](
...
 
-  private[this] val raceTimeout: Request[F] => F[Response[F]] =
+  private[this] def raceTimeout(req: Request[F]): F[(Response[F], F[Unit])] =
     responseHeaderTimeout match {
       case finite: FiniteDuration =>
-        val timeoutResponse = timer.sleep(finite).as(Response.timeout[F])
-        req =>
-          F.race(runApp(req), timeoutResponse).map(_.merge)
+        val timeoutResponse = timer.sleep(finite).as((Response.timeout[F], F.unit))
+        F.race(run(req).allocated, timeoutResponse).map(_.merge)
```
nigredo-tori: `race` doesn't guarantee that the finalizers on each of the parts "see" them being cancelled. E.g. in this example the resource is not released, even though `race` waits for allocation to finish before continuing:
```scala
val slowRes = Resource.make(IO {
  Thread.sleep(1000)
  println("Acquired")
})(_ => IO { println("Released") })

IO.race(slowRes.allocated, IO { Thread.sleep(500) })
  .flatMap(_.left.toOption.map(_._2).sequence_)
```

nigredo-tori: This gets curiouser: https://github.com/typelevel/cats-effect/issues/579.

nigredo-tori: Here's a stab at solving this issue from another angle, without diving into the `allocated`/`continual`/`uncancelable` rabbit hole:
```scala
def raceResource[F[_]: Concurrent, A](
  res: Resource[F, A],
  timeout: F[A]
): Resource[F, A] = for {
  earlyRelease <- Resource.liftF(Deferred[F, Unit])
  acquired <- Resource.liftF(Deferred[F, Either[Throwable, A]])

  _ <- res.attempt.use(
    acquired.complete(_) *> earlyRelease.get
  ).background

  a <- Resource.liftF(
    timeout.attempt.race(acquired.get)
      .flatTap { result =>
        // If the timeout has won, we can release the resource early.
        earlyRelease.complete(()).whenA(result.isLeft)
      }.map(_.merge)
  )
} yield a
```

This relies only on high-level combinators, and it shouldn't mask cancellation anywhere (there are no `Resource.make` or `uncancelable` calls). I haven't tested it, though.

rossabaker: It has been a while since I looked at this, and I can't remember whether we can add this without breaking binary compatibility.  Not as the draft currently stands, but if we are more careful in the blaze builder and grant MiMa exceptions on private API, then probably yes?

If so, this would be neat to try to get into the 0.21 series.  If not, I like your thinking above, but we'll soon have to port it to Cats Effect 3.

SystemFw: > without diving into the allocated/continual/uncancelable rabbit hole

The other thing to mention is that CE3 story on this stuff is a million times better, in general (although I haven't tried this specific example, so you never know)

SystemFw: What you wrote above is not strictly speaking a `race` I don't think, a `Resource` composed of two slow `make` calls which times out during the first `make` won't be interrupted until _both_ complete

nigredo-tori: > The other thing to mention is that CE3 story on this stuff is a million times better, in general (although I haven't tried this specific example, so you never know)

Yes, probably (I haven't really touched it yet), but I'd still rather deal with `Bracket`-like behavior over `allocated`.

> What you wrote above is not strictly speaking a race I don't think

I don't know what to call it. It encapsulates a pattern in this draft, where resource acquisition races against some timeout.

> a Resource composed of two slow make calls which times out during the first make won't be interrupted until both complete

Wait, so a `Resource` can't be interrupted between two `Allocate`s? That's surprising.

Still, I'm not sure how that applies to the code I wrote above. Unless you're saying that `res.use {... }` can't be cancelled during acquisition.

nigredo-tori: > a Resource composed of two slow make calls which times out during the first make won't be interrupted until both complete

I've made a [scastie](https://scastie.scala-lang.org/AFuMkDP5RJ2rtidKV2ZIRw) to play around. I think I now see what you mean: `earlyRelease` does not cancel the *acquisition* of `res`: note that `bar` is still acquired even after the timeout.

Here's a revised `raceResource` [without this issue](https://scastie.scala-lang.org/srfhyAb6QXKvOehzG5IwRQ) (I've changed the `res.attempt` part):
```scala
def raceResource[F[_], A](
  res: Resource[F, A],
  timeout: F[A]
)(implicit F: Concurrent[F]): Resource[F, A] = for {
  earlyRelease <- Resource.liftF(Deferred[F, Unit])
  acquired <- Resource.liftF(Deferred[F, Either[Throwable, A]])

  _ <- res.attempt.use(
    acquired.complete(_) <* F.never
  ).race(earlyRelease.get)
    .background

  a <- Resource.liftF(
    timeout.attempt.race(acquired.get)
      .flatTap { result =>
        // If the timeout has won, we can release the resource early.
        earlyRelease.complete(()).whenA(result.isLeft)
      }.map(_.merge)
      .flatMap(_.liftTo[F])
  )
} yield a
```

nigredo-tori:
Another way to approach this would be using a scope abstraction like the one defined [here](https://github.com/typelevel/cats-effect/pull/1496). Then instead of `Http[Resource[F, *], F]` as a parallel to `HttpApp`, we'll have `Http[Kleisli[F, Scope[F], *], F]` - and the parallel for `HttpRoutes` will be `ContextRoutes[Scope[F], F]`!

This *might* be easier to think about - after all, we already have `ContextMiddleware`s in the library. Speaking of which, migrating existing middlewares should be simple enough - keep the `Middleware`s and `ContextMiddleware`s the same, and provide some combinators along these lines (name and signature temporary):
```scala
def applyZipContext[F[_], A, B](
  middleware: ContextMiddleware[F, A]
)(routes: ContextRoutes[F, (B, A)]): ContextRoutes[F, B] =
  Kleisli { reqB =>
    middleware(
      routes
        .contramap(_.tupleLeft(reqB.context))
    ).run(reqB.req)
  }
```

We'll want to update the ones that currently rely on `BracketRequestResponse` from `ContextRoutes[F, T] => HttpRoutes[F]` to `ContextRoutes[F, T] => ContextRoutes[Scope[F], T]` - but that should be reasonably easy, and since the old versions would still work, we'll be able to do that as a separate issue.

rossabaker:
Let's check my understanding when I'm probably too sleepy to understand: if we unwrap all the transformers and aliases:
* The scoped HttpApp unwraps to `Request[F] => (Scope[F] => F[Response[F]])`
* The scoped HttpRoutes unwraps to `ContextRequest[Scope[F], F] => F[Option[Response[F]]`?

I'm struggling a bit with the different currying, but I am intrigued by this `Scope` idea.

nigredo-tori:
> The scoped HttpApp unwraps to Request[F] => (Scope[F] => F[Response[F]])

From what I wrote, yes. Actually, that abstraction should unwrap to `ContextRequest[F, Scope[F]] => F[Response[F]` to keep currying the same, but that didn't fit into `Http`, and I was also sleepy when I wrote that (still am). :smile:

> The scoped HttpRoutes unwraps to ContextRequest[Scope[F], F] => F[Option[Response[F]]?

Yes, but it's `ContextRequest[F, Scope[F]] => F[Option[Response[F]]` (see the order of type parameters in `ContextRequest`). 

PS: the order of type parameters in `ContextRoutes[T, F[_]]` might actually be a bug.

rossabaker:
I vaguely remember there being some inference issue in `ContextRoutes` that made it better to put the higher-kinded parameter second. It's definitely weird, and it would be interesting to revisit.

I like the idea here, especially if the currying lines up.

nigredo-tori:
> it would be interesting to revisit.

I've created #4015 for that.

blaze-server/src/main/scala/org/http4s/server/blaze/Http1ServerStage.scala
nigredo-tori: With CE3 we can just use `Resource.race` from the library instead of all this.

Issue #2583, opened 2019-05-19
Uri.fromString throws on some invalid port numbers


niij:
`org.http4s.Uri.fromString` throws when the port number is greater than 2,147,483,647  (`java.lang.Integer.MAX_VALUE`).

Versions
Present in 0.20.1 and 0.18.23

```scala
$ amm
Loading...
Welcome to the Ammonite Repl 1.1.2
(Scala 2.12.6 Java 1.8.0_181)
If you like Ammonite, please support our development at www.patreon.com/lihaoyi
@ import $ivy.`org.http4s::http4s-core:0.18.23`
import $ivy.$

@ org.http4s.Uri.fromString("https://example.com:2147483647")
res1: org.http4s.package.ParseResult[org.http4s.Uri] = Right(
  Uri(
    Some(Scheme(https)),
    Some(Authority(None, RegName(example.com), Some(2147483647))),
    "",
    Query(),
    None
  )
)

@ // Greater than java.lang.Integer.MAX_VALUE

@ org.http4s.Uri.fromString("https://example.com:2147483648")
java.lang.NumberFormatException: For input string: "2147483648"
  java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
  java.lang.Integer.parseInt(Integer.java:587)
  java.lang.Integer.parseInt(Integer.java:615)
  scala.collection.immutable.StringLike.toInt(StringLike.scala:301)
  scala.collection.immutable.StringLike.toInt$(StringLike.scala:301)
  scala.collection.immutable.StringOps.toInt(StringOps.scala:29)
  org.http4s.parser.Rfc3986Parser.Port(Rfc3986Parser.scala:85)
  org.http4s.parser.Rfc3986Parser.Port$(Rfc3986Parser.scala:83)
  org.http4s.parser.RequestUriParser.Port(RequestUriParser.scala:8)
  org.http4s.parser.Rfc3986Parser.Authority(Rfc3986Parser.scala:64)
  org.http4s.parser.Rfc3986Parser.Authority$(Rfc3986Parser.scala:63)
  org.http4s.parser.RequestUriParser.Authority(RequestUriParser.scala:8)
  org.http4s.parser.Rfc3986Parser.HierPart(Rfc3986Parser.scala:40)
  org.http4s.parser.Rfc3986Parser.HierPart$(Rfc3986Parser.scala:39)
  org.http4s.parser.RequestUriParser.HierPart(RequestUriParser.scala:8)
[...]

PR #2664, opened 2019-06-21
Address Uri#fromString Defect for Parsing the Port

kevinmeredith:
Closes https://github.com/http4s/http4s/issues/2583 by implementing one of @rossabaker's suggestions from https://github.com/http4s/http4s/issues/2583#issuecomment-496330716.

### DIFF
--- a/core/src/main/scala/org/http4s/Uri.scala
+++ b/core/src/main/scala/org/http4s/Uri.scala
@@ -13,7 +13,7 @@ import org.http4s.util._
 import scala.collection.immutable
 import scala.math.Ordered
 import scala.reflect.macros.whitebox
-import scala.util.{Try, Success, Failure}
+import scala.util.{Failure, Success, Try}
 
 /** Representation of the [[Request]] URI
   *
@@ -158,7 +158,6 @@ object Uri {
     }
 
   }
-
 
   /** Parses a String to a [[Uri]] according to RFC 3986.  If decoding
     *  fails, throws a [[ParseFailure]].
--- a/core/src/main/scala/org/http4s/Uri.scala
+++ b/core/src/main/scala/org/http4s/Uri.scala
@@ -3,7 +3,6 @@ package org.http4s
 import cats.{Eq, Order, Show}
 import cats.implicits.{catsSyntaxEither => _, _}
 import java.nio.charset.StandardCharsets
-
 import org.http4s.Uri._
 import org.http4s.internal.parboiled2.CharPredicate.{Alpha, Digit}
 import org.http4s.internal.parboiled2.{Parser => PbParser}
--- a/core/src/main/scala/org/http4s/Uri.scala
+++ b/core/src/main/scala/org/http4s/Uri.scala
@@ -3,6 +3,7 @@ package org.http4s
 import cats.{Eq, Order, Show}
 import cats.implicits.{catsSyntaxEither => _, _}
 import java.nio.charset.StandardCharsets
+
 import org.http4s.Uri._
 import org.http4s.internal.parboiled2.CharPredicate.{Alpha, Digit}
 import org.http4s.internal.parboiled2.{Parser => PbParser}
@@ -12,6 +13,7 @@ import org.http4s.util._
 import scala.collection.immutable
 import scala.math.Ordered
 import scala.reflect.macros.whitebox
+import scala.util.{Try, Success, Failure}
 
 /** Representation of the [[Request]] URI
   *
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/http4s/Uri.scala
```diff
@@ -141,10 +142,21 @@ object Uri {
...
 
   /** Decodes the String to a [[Uri]] using the RFC 3986 uri decoding specification */
-  def fromString(s: String): ParseResult[Uri] =
-    new RequestUriParser(s, StandardCharsets.UTF_8).Uri
-      .run()(PbParser.DeliveryScheme.Either)
-      .leftMap(e => ParseFailure("Invalid URI", e.format(s)))
+  def fromString(s: String): ParseResult[Uri] = {
+
+    val result: Try[ParseResult[Uri]] =
+      Try {
```
ChristopherDavenport:
```scala
Either.catchNonFatal{ block 
}.fold(Either.left[ParseFailure, Uri](_))(identity)
```

kevinmeredith: Thanks for the review, @ChristopherDavenport. I'll make that change later today!

Aside - you find `Either#catchNonFatal` stylistically cleaner, hence your review comment? I'm just asking to explore the distinction, if any, between `Try#apply` and what you noted.

ChristopherDavenport: It just brings in a new type for an equivalent type that is easier to unify. I find it easier, and cleaner in general.

SystemFw: You probably want `Either.catchOnly[ParseFailure](...)` instead

kevinmeredith: Why would it be better to catch only `ParseFailure`'s, i.e. propagate non-`ParseFailure`'s, @SystemFw, could you please say?

SystemFw: Actually my bad, it depends on what you want to do. Just go with Chris suggestion and sorry for the confusion :)

kevinmeredith: > Actually my bad

Not your bad at all! I was asking to learn (as I did talking to you at NEScala this year w/ Chris), @SystemFw, in case I was missing something.

> it depends on what you want to do.

It seemed worthwhile to me to return a `Left` for any `ParseFailure` due to a `NonFatal` being thrown. However, I've never touched this code before, so I admit I could be missing something. Please let me know if you disagree with `Either#catchNonFatal`.

tests/src/test/scala/org/http4s/UriSpec.scala
```diff
@@ -57,16 +58,49 @@ class UriSpec extends Http4sSpec with MustThrownMatchers {
...
+          {
+            val uri: ParseResult[Uri] = Uri.fromString(s"http://localhost:$notNumber/")
+            uri match {
+              case Left(ParseFailure("Invalid URI", _)) => ok
+              case unexpected => ko(unexpected.toString)
+            }
+          }
+        }.setGen(Gen.alphaNumStr.suchThat(str => str.nonEmpty && Try(str.toInt).isFailure))
+
+      }
```
rossabaker: Are we testing empty strings after a colon anywhere?  Those are irritatingly legal.

core/src/main/scala/org/http4s/Uri.scala
rossabaker: Could we just handle this directly in the Port rule of the parser?  An action function can return a rule.  Here's an example from the parboiled docs:

```scala
(foo: Rule1[Int]) ~> (i => test(i % 2 == 0) ~ push(i))
```

If we took that approach, we wouldn't have to worry about catching other funky exceptions in this much wider scope.

tests/src/test/scala/org/http4s/UriSpec.scala
kevinmeredith: Good point, Ross. I'll add such a test.

core/src/main/scala/org/http4s/Uri.scala
kevinmeredith: > Could we just handle this directly in the Port rule of the parser?

That change seems more surgical and cleaner to me.

> If we took that approach, we wouldn't have to worry about catching other funky exceptions in this much wider scope.

If I understand your comment, you recommend I remove catching and handling `NonFatal`'s given that I'll make the change in the `Port`'s parser?

kevinmeredith: I deleted my comment since I have a different question.

> Could we just handle this directly in the Port rule of the parser?

Could you please say more, @rossabaker? Looking at the `Port`'s code:

```scala
  def Port = rule {
    ":" ~ (capture(oneOrMore(Digit)) ~> { s: String =>
      (Some(s.toInt))
    } | push(None)) | push(None)
  }
```

are you suggesting that I ignore the `oneOrMoreDigit` if it does not evaluate to an `Int`, e.g. it's bigger than an `Int`'s max value?

rossabaker: Yes, I think if you create  `test` rule, you can range check the port, and have the rule fail if it's out of bounds.

rossabaker:
I think this is binary compatible if we want to target series/0.20.  I'm making a bunch of changes to URI in master, which will probably cover `Port`, but this would be nice to fix in the production version in the meantime.

core/src/main/scala/org/http4s/Uri.scala
kevinmeredith: Thanks, Ross. That helps. I'll make that change.

kevinmeredith:
To make sure I follow, @rossabaker, per https://github.com/http4s/http4s/pull/2664#issuecomment-506858839, you're suggesting that I re-open this PR against the `series/0.20.x` branch?

rossabaker:
Yes, or else I can just cherry-pick it.  Whatever is more convenient for you.

kevinmeredith:
https://github.com/http4s/http4s/pull/2687 supersedes this PR.

PR #2671, opened 2019-06-25
Implement UserInfo for URIs

rossabaker:
Model `UserInfo` as a case class, with instances and principled encoding and decoding.


core/src/main/scala/org/http4s/Uri.scala
```diff
@@ -244,6 +242,88 @@ object Uri {
...
+    *
+    * @param username The username component, decoded.
+    *
+    * @param password The password, decoded.  Passing a password in
+    * clear text in a URI is a security risk and deprecated by RFC
+    * 3986, but preserved in this model for losslessness.
+    *
+    * @see https://www.ietf.org/rfc/rfc3986.txt#section-3.21.
+    */
+  final case class UserInfo private (username: String, password: Option[String])
```
rossabaker: There are going to be more of these.  Should they continue to be nested in `Uri` or move to top level?

core/src/main/scala/org/http4s/Uri.scala
```diff
@@ -244,6 +242,88 @@ object Uri {
...
+  }
+
+  object UserInfo {
+
+    /** Parses a userInfo from a percent-encoded string. */
+    def fromString(s: String): ParseResult[UserInfo] =
+      fromStringWithCharset(s, StandardCharsets.UTF_8)
+
+    /** Parses a userInfo from a string percent-encoded in a specific charset. */
+    def fromStringWithCharset(s: String, cs: NioCharset): ParseResult[UserInfo] =
```
rossabaker: i don't know how we'd ever use this in practice without UTF-8, or why anyone would want to, but the spec allows it.  If we allow this, we should probably provide rendering to specific charsets, but that would be a partial operation.

core/src/main/scala/org/http4s/Uri.scala
```diff
@@ -244,6 +242,88 @@ object Uri {
...
+
+    private[http4s] trait Parser { self: Rfc3986Parser =>
+      def userInfo: Rule1[UserInfo] = rule {
+        capture(zeroOrMore(Unreserved | PctEncoded | SubDelims)) ~
+          (":" ~ capture(zeroOrMore(Unreserved | PctEncoded | SubDelims | ":"))).? ~>
+          (
+              (
+                  username: String,
+                  password: Option[String]) => UserInfo(decode(username), password.map(decode)))
+      }
```
rossabaker: I really don't like what scalafmt does to our parsers.  Maybe we should start exempting them.

core/src/main/scala/org/http4s/Uri.scala
```diff
@@ -244,6 +242,88 @@ object Uri {
...
+      fromStringWithCharset(s, StandardCharsets.UTF_8)
+
+    /** Parses a userInfo from a string percent-encoded in a specific charset. */
+    def fromStringWithCharset(s: String, cs: NioCharset): ParseResult[UserInfo] =
+      new Http4sParser[UserInfo](s, "Invalid user info") with Rfc3986Parser {
+        def main = userInfo
+        def charset = cs
+      }.parse
+
+    private[http4s] trait Parser { self: Rfc3986Parser =>
```
rossabaker: Following pattern of putting parsers in companion object, because they may call a private constructor after validation.  It does lead to some quirky self-typing.

core/src/main/scala/org/http4s/UriTemplate.scala
```diff
@@ -207,13 +207,16 @@ object UriTemplate {
   }
 
   protected def renderAuthority(a: Authority): String = a match {
-    case Authority(Some(u), h, None) => u + "@" + renderHost(h)
-    case Authority(Some(u), h, Some(p)) => u + "@" + renderHost(h) + ":" + p
+    case Authority(Some(u), h, None) => s"${renderUserInfo(u)}@${renderHost(h)}"
+    case Authority(Some(u), h, Some(p)) => s"${renderUserInfo(u)}@${renderHost(h)}:${p}"
```
rossabaker: I wonder whether anybody is using this feature.  I'm dragging it along for now.

laws/src/main/scala/org/http4s/laws/discipline/ArbitraryInstances.scala
```diff
@@ -627,10 +627,20 @@ private[http4s] trait ArbitraryInstances {
       genRegName)
   }
 
+  implicit val http4sTestingArbitraryForUserInfo: Arbitrary[Uri.UserInfo] =
+    Arbitrary(
+      for {
+        username <- getArbitrary[String]
+        password <- getArbitrary[Option[String]]
```
rossabaker: Maybe worth altering the mix since passwords are deprecated.

ritschwumm:
is this still useful nowadays? iirc modern browsers are removing or have removed support for this to a varying degree for security reasons... it looks like this removal is becoming standardized, too - see https://tools.ietf.org/html/rfc3986#section-7.5 .

rossabaker:
I did this one early because I'm going through the RFC top-to-bottom.  It raises some interesting questions that will pertain to other URI components, but it's a poor choice in its own relevance.  I would strongly argue against its use in any API I control, but I've also seen a lot of our users with APIs they don't control, and it remains part of the spec.  I would argue it belongs for compliance's sake, but I hope nobody uses it.

Subsequent PRs with paths and queries will be far more interesting and relevant.

PR #2673, opened 2019-06-26
Gzip middleware for client

balatbn:
Uses compress from fs2

### DIFF
--- a/client/src/main/scala/org/http4s/client/middleware/GZip.scala
+++ b/client/src/main/scala/org/http4s/client/middleware/GZip.scala
@@ -1,0 +1,47 @@
+package org.http4s
+package client
+package middleware
+
+import cats.effect.Sync
+import org.http4s.headers.{`Accept-Encoding`, `Content-Encoding`}
+
+/**
+  * Client middleware for enabling gzip.
+  */
+object GZip {
+
+  private val supportedCompressions =
+    Seq(ContentCoding.gzip.coding, ContentCoding.deflate.coding).mkString(", ")
+
+  def apply[F[_]: Sync](bufferSize: Int = 32 * 1024)(client: Client[F]): Client[F] =
+    Client[F] { req =>
+      val reqWithEncoding = addHeaders(req)
+      val responseResource = client.run(reqWithEncoding)
+
+      responseResource.map { actualResponse =>
+        actualResponse.withBodyStream(decompress(bufferSize, actualResponse))
+      }
+    }
+
+  private def addHeaders[F[_]: Sync](req: Request[F]): Request[F] =
+    req.headers.get(`Accept-Encoding`) match {
+      case Some(_) =>
+        req
+      case _ =>
+        req.withHeaders(
+          req.headers ++ Headers.of(Header(`Accept-Encoding`.name.value, supportedCompressions)))
+    }
+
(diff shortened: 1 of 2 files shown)


client/src/main/scala/org/http4s/client/middleware/GZip.scala
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+  * Client middleware for enabling gzip.
+  */
+object GZip {
+
+  private val supportedCompressions =
+    Seq(ContentCoding.gzip.coding, ContentCoding.deflate.coding).mkString(", ")
+
+  def apply[F[_]: Sync](bufferSize: Int = 32 * 1024)(client: Client[F]): Client[F] =
```
ChristopherDavenport: Is a Sync constraint necessary?

balatbn: fs2.compress.gunip requires an F that is RaiseThrowable. Since RaiseThrowable is from fs2, I added Sync here. Can I use any other type instead?

rossabaker: `ApplicativeError`?

client/src/main/scala/org/http4s/client/middleware/GZip.scala
```diff
@@ -0,0 +1,47 @@
...
+      case _ =>
+        req.withHeaders(
+          req.headers ++ Headers.of(Header(`Accept-Encoding`.name.value, supportedCompressions)))
+    }
+
+  private def decompress[F[_]: Sync](bufferSize: Int, response: Response[F]): EntityBody[F] =
+    response.headers.get(`Content-Encoding`) match {
+      case Some(header)
+          if header.contentCoding == ContentCoding.gzip || header.contentCoding == ContentCoding.`x-gzip` =>
+        response.body.through(fs2.compress.gunzip(bufferSize))
```
rossabaker: Should this remove that ContentCoding as well?  Similar on the deflate case.

balatbn: I thought it is better not to touch this header as actual code can use this to confirm whether response was actually compressed. I can remove this if this is not required.

client/src/main/scala/org/http4s/client/middleware/GZip.scala
balatbn: I tried changing it to ApplicativeError. Now, I'm getting missing Bracket error for new Client creation. I'll update PR by changing this to Bracket

balatbn: Updated PR

client/src/main/scala/org/http4s/client/middleware/GZip.scala
rossabaker: I think it's better that we do, because the `Content-Coding` no longer reflects reality.  This is important in case the client is used in a proxy.  We should also be sure `Content-Length` is gone.  I can't remember offhand whether `withBodyStream` takes care of this or not.

balatbn: Changed code to remove both the headers

PR #2679, opened 2019-06-27
Create a modeled Ipv4Address

rossabaker:
- Replace `IPv4` wrapper around `CaseInsensitiveString` with `Ipv4Address`, a case class of four bytes
- Add an `ipv4` interpolator
- Interop with `Inet4Address`


rossabaker:
Interesting, unrelated error here.  I can't reproduce it.

```
[info] decode should
[error]   x be consistent with utf8Decode (1 second, 564 ms)
[error]    Falsified after 92 passed tests.
[error]    > ARG_0: "﻿"
[error]    > ARG_0_ORIGINAL: "﻿兦냡䶏骧ⶾ䙅"
[error]    > ARG_1: 1
[error]    > ARG_1_ORIGINAL: 1445009706
[error]    The seed is e5-vbNb5f12ZKua5LXDAI4s5eyLtnY5RGUIoFXj0JEE=
[error]    
[error]    > '﻿' != '' (DecodeSpec.scala:24)
[error] org.http4s.util.DecodeSpec.$anonfun$new$4(DecodeSpec.scala:24)
```

aeons:
Is the casing change intentional? `IPv4` -> `IpV4`

rossabaker:
Dating back to my Java days, I don't capitalize acronyms, so `Ip`.  `V` begins a new word, so it's capitalized.  Hence `IpV4`.  Though we could say the `V` is part of the same acronym, and `Ipv4`.  Or we could stylize it like the spec, and `IPv4`.  I guess I can justify any of the three.  Strong opinions welcome.

aeons:
I can accept `Ipv4` :D

PR #2682, opened 2019-06-28
Update json4s-core, json4s-jackson, ... to 3.6.7

rossabaker:
Replaces #2676


rossabaker:
Nah, I don't care enough to cross version.

PR #2687, opened 2019-07-01
Address Uri#fromString Defect for Parsing the Port

### LINKED ISSUE #2583: Uri.fromString throws on some invalid port numbers

kevinmeredith:
Closes #2583 by implementing one of @rossabaker's suggestions from #2583 (comment).

Per https://github.com/http4s/http4s/pull/2664#issuecomment-506858839, I decided to close https://github.com/http4s/http4s/pull/2664 in favor of this PR.

### DIFF
--- a/core/src/main/scala/org/http4s/parser/Rfc3986Parser.scala
+++ b/core/src/main/scala/org/http4s/parser/Rfc3986Parser.scala
@@ -82,7 +82,10 @@ private[parser] trait Rfc3986Parser
 
   def Port = rule {
     ":" ~ (capture(oneOrMore(Digit)) ~> { s: String =>
-      (Some(s.toInt))
+
+      val int: Option[Int] = Either.catchOnly[NumberFormatException](s.toInt).toOption
+
+      test(int.nonEmpty) ~ push(int)
     } | push(None)) | push(None)
   }
 
--- a/core/src/main/scala/org/http4s/parser/Rfc3986Parser.scala
+++ b/core/src/main/scala/org/http4s/parser/Rfc3986Parser.scala
@@ -2,7 +2,7 @@ package org.http4s
 package parser
 
 import cats.syntax.option._
-import cats.syntax.either.catsSyntaxEitherObject 
+import cats.syntax.either.catsSyntaxEitherObject
 import java.net.URLDecoder
 import java.nio.charset.Charset
 import org.http4s.{Query => Q}
--- a/core/src/main/scala/org/http4s/parser/Rfc3986Parser.scala
+++ b/core/src/main/scala/org/http4s/parser/Rfc3986Parser.scala
@@ -82,7 +82,6 @@ private[parser] trait Rfc3986Parser
 
   def Port = rule {
     ":" ~ (capture(oneOrMore(Digit)) ~> { s: String =>
-
       val int: Option[Int] = Either.catchOnly[NumberFormatException](s.toInt).toOption
 
       test(int.nonEmpty) ~ push(int)
--- a/core/src/main/scala/org/http4s/parser/Rfc3986Parser.scala
+++ b/core/src/main/scala/org/http4s/parser/Rfc3986Parser.scala
@@ -1,7 +1,8 @@
 package org.http4s
 package parser
 
-import cats.implicits._
+import cats.syntax.option._
+import cats.syntax.either.catsSyntaxEitherObject 
 import java.net.URLDecoder
 import java.nio.charset.Charset
 import org.http4s.{Query => Q}
--- a/tests/src/test/scala/org/http4s/UriSpec.scala
+++ b/tests/src/test/scala/org/http4s/UriSpec.scala
@@ -86,6 +86,43 @@ http://example.org/a file
       }
     }
 
+    "fail to parse port" >> {
+      "if it's negative" >> prop { negative: Int =>
+      {
+        val uri: ParseResult[Uri] = Uri.fromString(s"http://localhost:$negative/")
+        uri match {
+          case Left(ParseFailure("Invalid URI", _)) => ok
+          case unexpected => ko(unexpected.toString)
+        }
+      }
+      }.setGen(Gen.choose[Int](Int.MinValue, -1))
+      "if it's larger than Int.MaxValue" >> prop { tooBig: Long =>
+      {
+        val uri: ParseResult[Uri] = Uri.fromString(s"http://localhost:$tooBig/")
+        uri match {
+          case Left(ParseFailure("Invalid URI", _)) => ok
+          case unexpected => ko(unexpected.toString)
+        }
+      }
+      }.setGen(Gen.choose[Long]((Int.MaxValue: Long) + 1, Long.MaxValue))
+      "if it's not a number or an empty String" >> prop { notNumber: String =>
+      {
+        val uri: ParseResult[Uri] = Uri.fromString(s"http://localhost:$notNumber/")
+        uri match {
+          case Left(ParseFailure("Invalid URI", _)) => ok
+          case unexpected => ko(unexpected.toString)
+        }
+      }
+      }.setGen(Gen.alphaNumStr.suchThat(str => str.nonEmpty && Either.catchOnly[NumberFormatException](str.toInt).isLeft))
+      "for an empty String" in {
+        val uri: ParseResult[Uri] = Uri.fromString("http://localhost:/")
+        uri match {
+          case Left(ParseFailure("Invalid URI", _)) => ok
+          case unexpected => ko(unexpected.toString)
+        }
+      }
+    }
+
     "support a '/' operator when original uri has trailing slash" in {
       val uri = getUri("http://localhost:8080/")
       val newUri = uri / "echo"
--- a/tests/src/test/scala/org/http4s/UriSpec.scala
+++ b/tests/src/test/scala/org/http4s/UriSpec.scala
@@ -57,8 +57,8 @@ class UriSpec extends Http4sSpec with MustThrownMatchers {
       }
 
       "parse port correctly" >> {
-        "if there is a valid (non-negative) one" >> prop {
-          nonNegative: Int => {
+        "if there is a valid (non-negative) one" >> prop { nonNegative: Int =>
+          {
             val uri = getUri(s"http://localhost:$nonNegative/")
             uri.port must_=== Some(nonNegative)
           }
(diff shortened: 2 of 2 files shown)


tests/src/test/scala/org/http4s/UriSpec.scala
```diff
@@ -84,6 +86,44 @@ http://example.org/a file
...
+        {
+          val uri: ParseResult[Uri] = Uri.fromString(s"http://localhost:$notNumber/")
+          uri match {
+            case Left(ParseFailure("Invalid URI", _)) => ok
+            case unexpected => ko(unexpected.toString)
+          }
+        }
+      }.setGen(Gen.alphaNumStr.suchThat(str =>
+        str.nonEmpty && Either.catchOnly[NumberFormatException](str.toInt).isLeft))
+      "for an empty String" in {
```
kevinmeredith: Locally, this test failed:

```
[info]   fail to parse port
[info]     + if it's negative (214 ms)
[info]     + if it's larger than Int.MaxValue (230 ms)
[info]     + if it's not a number or an empty String (256 ms)
[error]     x for an empty String (271 ms)
[error]      Right(http://localhost/) (UriSpec.scala:121)
[error] org.http4s.UriSpec.$anonfun$new$51(UriSpec.scala:121)
```

@rossabaker. 

Per https://tools.ietf.org/html/rfc3986#page-22, 

```
3.2.3.  Port

   The port subcomponent of authority is designated by an optional port
   number in decimal following the host and delimited from it by a
   single colon (":") character.

      port        = *DIGIT

   A scheme may define a default port.  For example, the "http" scheme
   defines a default port of "80", corresponding to its reserved TCP
   port number.  The type of port designated by the port number (e.g.,
   TCP, UDP, SCTP) is defined by the URI scheme.  URI producers and
   normalizers should omit the port component and its ":" delimiter if
   port is empty or if its value would be the same as that of the
   scheme's default.
```

Per https://tools.ietf.org/html/rfc3986#page-22, 

```
6.2.3.  Scheme-Based Normalization

   The syntax and semantics of URIs vary from scheme to scheme, as
   described by the defining specification for each scheme.
   Implementations may use scheme-specific rules, at further processing
   cost, to reduce the probability of false negatives.  For example,
   because the "http" scheme makes use of an authority component, has a
   default port of "80", and defines an empty path to be equivalent to
   "/", the following four URIs are equivalent:

      http://example.com
      http://example.com/
      http://example.com:/
      http://example.com:80/
```

this test, it seems clear to me, should fail, not pass. In other words, I should replace this "should fail" test with a "should parse to port 80" one.

Please let me know if you disagree.

core/src/main/scala/org/http4s/parser/Rfc3986Parser.scala
```diff
@@ -82,7 +82,9 @@ private[parser] trait Rfc3986Parser
 
   def Port = rule {
     ":" ~ (capture(oneOrMore(Digit)) ~> { s: String =>
-      (Some(s.toInt))
+      val int: Option[Int] = Either.catchOnly[NumberFormatException](s.toInt).toOption
```
kevinmeredith: I was able to use your suggestion, https://github.com/http4s/http4s/pull/2664#discussion_r296258166, @SystemFw. Thanks for it!

tests/src/test/scala/org/http4s/UriSpec.scala
ChristopherDavenport: Yup, That seems correct via the spec.

kevinmeredith:
I closed, and then re-opened, due to, what appears to be, a flaky test.

kevinmeredith:
I need to address:

```scala
[error] /home/travis/build/http4s/http4s/core/src/main/scala/org/http4s/parser/Rfc3986Parser.scala:85:69: type mismatch;
[error]  found   : Either[NumberFormatException,Int]
[error]  required: ?{def toOption: ?}
[error] Note that implicit conversions are not applicable because they are ambiguous:
[error]  both method EitherSyntax in package http4s of type [L, R](self: Either[L,R])org.http4s.package.EitherSyntax[L,R]
[error]  and method catsSyntaxEither in trait EitherSyntax of type [A, B](eab: Either[A,B])cats.syntax.EitherOps[A,B]
[error]  are possible conversion functions from Either[NumberFormatException,Int] to ?{def toOption: ?}
[error]       val int: Option[Int] = Either.catchOnly[NumberFormatException](s.toInt).toOption
[error]                                                                     ^
[error] one error found
[error] (core / Compile / compileIncremental) Compilation failed
[error] Total time: 67 s, completed Jul 2, 2019 2:27:51 PM
```

per latest build failure.

tests/src/test/scala/org/http4s/UriSpec.scala
```diff
@@ -57,14 +57,23 @@ class UriSpec extends Http4sSpec with MustThrownMatchers {
...
+            val uri = getUri(s"http://localhost:$nonNegative/")
+            uri.port must_=== Some(nonNegative)
+          }
+        }.setGen(Gen.choose[Int](0, Int.MaxValue))
         "if there is none" in {
           val uri = getUri("http://localhost/")
           uri.port must_=== None
         }
+        // See RFC 3986, section 6.2.3
+        "for an empty String" in {
```
ChristopherDavenport: Given `https://foobar:/` should this infer `443` rather than `80`? I worry as I believe some may rely on this behavior by accident and then they would go to the unsecured port by default.

rossabaker: Are we inferring anywhere in the Uri?  We do in the clients, but the Uri has an optional port.  You're correct that the inference is scheme specific, but I think this test is right.

kevinmeredith: > Given https://foobar:/ should this infer 443 rather than 80?

Is it worthwhile to open another issue, @ChristopherDavenport, if `https://foobar:/` resolves a port of 80 rather than 443? I'm only asking since, I believe, my PR addresses https://github.com/http4s/http4s/issues/2583.

niij:
Thanks @kevinmeredith !

tests/src/test/scala/org/http4s/UriSpec.scala
kevinmeredith: Hi @ChristopherDavenport -

> Given https://foobar:/ should this infer 443 rather than 80?

So the following looks right or wrong to you? I admit to not knowing how the `Port` is used if there's a supplied `Scheme`, i.e. my first `Uri#fromString` REPL example.

```scala
@ import $ivy.`org.http4s::http4s-core:0.20.6` 
import $ivy.$                               

@ Uri.fromString("https://www.google.com:80") 
res5: org.http4s.package.ParseResult[Uri] = Right(Uri(Some(Scheme(https)), Some(Authority(None, RegName(www.google.com), Some(80))), "", , None))

@ Uri.fromString("https://www.google.com:/") 
res6: org.http4s.package.ParseResult[Uri] = Right(Uri(Some(Scheme(https)), Some(Authority(None, RegName(www.google.com), None)), "/", , None))
```

Does this look right, Chris?

Issue #2660, opened 2019-06-20
Drop Scala 2.11 support


rossabaker:
Some of our integrations have dropped or are dropping scala 2.11 support.

* scalatags
* scala-xml
* circe (proposed: https://github.com/circe/circe/pull/1176)

Scala 2.11 represented approximately 5% of our downloads according to Maven Central's (dubious) stats.

Our core libraries all support 2.11, so we can continue to support it, with a bit of cross versioning in our integrations.  We have an increasing amount of complications because of it, and it's common for libraries to only support two versions of Scala concurrently.  Is it time?

Discuss.

ChristopherDavenport:
I'm mildly against this. So long as our core supports it I don't see a great reason to drop it. However most folks still on 2.11 likely aren't on the newest versions.

rossabaker:
Part of my work is on Scala 2.11, and that part is on old versions of everything.

All of my previous work was on 2.10 and 2.11, and making it work was part of my job.  That's no longer.

We help people incrementally migrate when we're generous in support, but at the expense of our own limited time.

I'm mildly in favor of keeping 2.11, but decreasigly so.  I think 0.21 might be the end.

aeons:
I am in favor of dropping 2.11, even on 0.21. I don't have any strong feelings on this either way though.

rossabaker:
Circe dropped 2.11 in the latest milestone, so we'll definitely need a strategy on master.

ChristopherDavenport:
I think it likely that those on 2.11 just aren't on the newest stuff. I don't think it's a bad idea to drop it under the condition that one of our most used modules can no longer support it.

rossabaker:
Per June Sonatype stats:
- 10% of all core downloads are 2.11
- 4% of all core downloads were 2.11 without circe.  These are the marginal users we'd leave behind.
- 63% of 2.12 users are on http4s >= 0.20, vs. 24% of 2.11 users 
- 2.11 downloads are down 50% (vs. up 13% for 2.12)

If the last three populations are independent, we're talking about roughly 1% of our users with a half-life of a month.  And we can support them with binary-compatible fixes for as long as people send them to us.  I think it's time.

rossabaker:
The result of #2694 is a small net loss in code, simpler imports, reduced build times, and more flexibility to maintain binary compatibility with abstract classes in 0.21.

ijuma:
Good call in my opinion, especially with the recent release of 2.13.0.

[...]

PR #2694, opened 2019-07-05
Upgrade to circe-0.12.0-M4, drop scala-2.11

rossabaker:
Discussion at #2660.


core/src/main/scala/org/http4s/QValue.scala
```diff
@@ -129,7 +129,7 @@ object QValue {
   @deprecated("""use qValue"" string interpolation instead""", "0.20")
   def q(d: Double): QValue = macro Macros.qValueLiteral
 
-  implicit val http4sOrderForQValue: Order[QValue] = Order.fromOrdering[QValue]
+  implicit lazy val http4sOrderForQValue: Order[QValue] = Order.fromOrdering[QValue]
```
rossabaker: It's not immediately clear to me why this became necessary.  Without it, we get:

```scala
[error] cannot create an instance for class org.http4s.testing.ResponseGeneratorSpec
[error]   caused by java.lang.ExceptionInInitializerError
[error]   caused by scala.UninitializedFieldError: Uninitialized field: /home/travis/build/http4s/http4s/core/src/main/scala/org/http4s/QValue.scala: 132
```

rossabaker: Okay, this is now causing a stack overflow error.  I'm reaching the stupid hours... what's suddenly wrong with this?

travisbrown: The issue is the new import pulling in `catsKernelOrderingForOrder`, which has higher precedence than the `Ordering`-from-`Ordered` stuff. You could either hide it, or only import `catsSyntaxEither`, or (what I'd do) not use `valueOr`.

rossabaker: Oof, thanks. I'd moved on to a different branch and was stupefied by what import was even causing it. As usual, I should have quit an hour before I did.

ChristopherDavenport:
Test failure is a flaky test on Timeout Middleware.

Issue #2651, opened 2019-06-17
Websocket close : EOF


ikempf:
I'm using http4s 0.20.3 (with blaze-server) for my websocket backend.

I tested numerous websocket clients against it and often get the following error when closing the connection :
```
http4s-eof 15:26:15.903 [scala-execution-context-global-38] ERROR org.http4s.blazecore.websocket.Http4sWSStage - Error closing Web Socket
http4s-eof org.http4s.blaze.pipeline.Command$EOF$: EOF
http4s-eof 	at org.http4s.blaze.pipeline.Command$EOF$.scala$util$control$NoStackTrace$$super$fillInStackTrace(Command.scala:23)
http4s-eof 	at scala.util.control.NoStackTrace.fillInStackTrace(NoStackTrace.scala:28)
http4s-eof 	at scala.util.control.NoStackTrace.fillInStackTrace$(NoStackTrace.scala:27)
http4s-eof 	at org.http4s.blaze.pipeline.Command$EOF$.fillInStackTrace(Command.scala:23)
http4s-eof 	at java.lang.Throwable.<init>(Throwable.java:265)
http4s-eof 	at java.lang.Exception.<init>(Exception.java:66)
http4s-eof 	at org.http4s.blaze.pipeline.Command$EOF$.<init>(Command.scala:23)
http4s-eof 	at org.http4s.blaze.pipeline.Command$EOF$.<clinit>(Command.scala)
http4s-eof 	at org.http4s.blaze.channel.nio1.NIO1HeadStage$$anon$3.run(NIO1HeadStage.scala:322)
http4s-eof 	at org.http4s.blaze.util.TaskQueue.go$1(TaskQueue.scala:82)
http4s-eof 	at org.http4s.blaze.util.TaskQueue.executeTasks(TaskQueue.scala:99)
http4s-eof 	at org.http4s.blaze.channel.nio1.SelectorLoop.org$http4s$blaze$channel$nio1$SelectorLoop$$runLoop(SelectorLoop.scala:166)
http4s-eof 	at org.http4s.blaze.channel.nio1.SelectorLoop$$anon$1.run(SelectorLoop.scala:67)
http4s-eof 	at java.lang.Thread.run(Thread.java:748)
```

This seems to only happen when the connection is closed by the server. Cleanly closing from the client side does not result in the above error log.

[...]

PR #2695, opened 2019-07-05
Handle EOF in web socket by shutting down stage

rossabaker:
As described in #2651, we're getting a noisy EOF when shutting down web sockets under certain circumstances.  Usually we call `stageShutdown()` when we receive this to do any cleanup, so added that as a special case.  This may result in `stageShutdown()` being called multiple times, but that's normal.

This feels a tad sloppy, but not as sloppy as scary errors in user logs for situations that should be harmless.

### DIFF
--- a/blaze-core/src/main/scala/org/http4s/blazecore/websocket/Http4sWSStage.scala
+++ b/blaze-core/src/main/scala/org/http4s/blazecore/websocket/Http4sWSStage.scala
@@ -8,6 +8,7 @@ import fs2._
 import fs2.concurrent.SignallingRef
 import java.util.concurrent.atomic.AtomicBoolean
 import org.http4s.blaze.pipeline.{LeafBuilder, TailStage, TrunkBuilder}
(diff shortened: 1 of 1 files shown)


rossabaker:
Even the "error closing websocket" name is misleading.  When we get a ProtocolException decoding the incoming websocket, it gets logged there.

Dennis4b:
This patch wraps `stageShutdown` in other IO:

```
128       case Left(EOF) =>                                                                                                    
129         IO(stageShutdown())
```

 but `stageShutdown` itself looks like:

```
138   override protected def stageShutdown(): Unit = {                                                                         
139     F.toIO(deadSignal.set(true)).unsafeRunSync()                                                                           
140     super.stageShutdown()                                                                                                  
141   }
```

This nesting with `unsafeRunSync`s increases the risk of deadlocks.

The following would IMHO be preferrable:
```
case Left(EOF) =>
    F.toIO(deadSignal.set(true)).map(_ => super.stageShutdown)
```

PR #2697, opened 2019-07-05
fix index.html

gekomad:
v0.21
removed `scalacOptions += "-Ypartial-unification"`
update `sbt -sbt-version 1.2.8 new http4s/http4s.g8 -b 0.20` to  `sbt -sbt-version 1.2.8 new http4s/http4s.g8 -b 0.21`

### DIFF
--- a/v0.21/index.html
+++ b/v0.21/index.html
@@ -82,16 +82,11 @@
 
           
           <h1>Quick Start</h1>
-          
-          
-          <p><strong>Note</strong>: To run examples, please make sure that the flag <code>-Ypartial-unification</code><br />
-is enabled in your compiler options (i.e <code>scalacOptions += &quot;-Ypartial-unification&quot;</code> in sbt).
-This is enabled by default in the giter8 template.</p>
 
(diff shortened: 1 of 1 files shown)


v0.21/index.html
```diff
@@ -82,16 +82,11 @@
 
           
           <h1>Quick Start</h1>
-          
-          
-          <p><strong>Note</strong>: To run examples, please make sure that the flag <code>-Ypartial-unification</code><br />
-is enabled in your compiler options (i.e <code>scalacOptions += &quot;-Ypartial-unification&quot;</code> in sbt).
-This is enabled by default in the giter8 template.</p>
```
rossabaker: It kind of stinks that this is only true for 2.13.  I wonder if we should leave a warning for our 2.12 users?

ChristopherDavenport: `To run with 2.12 `...

rossabaker: I like that idea.  Let's restore it with that language, and I think this is good to go.

rossabaker:
Oh, this PR targets gh-pages.  We need to modify the index.md.  And the original repo is gone.  I'll reopen with the proposed fix.

PR #2700, opened 2019-07-07
Improved ipv6 support in Uri

rossabaker:
* Deprecate `IPv6` in favor of `Ipv6Address`
* `Ipv6Address` is a case class of eight hextets
* Implements RFC5952 for normalized output
* Adds string interpolator
* Requires that ipv6 addresses be bracketed in authority form of request target

### DIFF
--- a/core/src/main/scala/org/http4s/Uri.scala
+++ b/core/src/main/scala/org/http4s/Uri.scala
@@ -2,11 +2,12 @@ package org.http4s
 
 import cats.{Eq, Hash, Order, Show}
 import cats.implicits._
-import java.net.{Inet4Address, InetAddress}
+import java.net.{Inet4Address, Inet6Address, InetAddress}
+import java.nio.ByteBuffer
 import java.nio.charset.{Charset => NioCharset, StandardCharsets}
 import org.http4s.Uri._
 import org.http4s.internal.parboiled2.{Parser => PbParser, _}
-import org.http4s.internal.parboiled2.CharPredicate.{Alpha, Digit}
+import org.http4s.internal.parboiled2.CharPredicate.{Alpha, Digit, HexDigit}
 import org.http4s.parser._
 import org.http4s.syntax.string._
 import org.http4s.util._
@@ -329,30 +330,15 @@ object Uri {
   }
 
   sealed trait Host extends Renderable {
-    final def value: String = this match {
-      case RegName(h) => h.toString
-      case addr: Ipv4Address =>
-        new StringBuffer()
-          .append(addr.a & 0xff)
-          .append(".")
-          .append(addr.b & 0xff)
-          .append(".")
-          .append(addr.c & 0xff)
-          .append(".")
-          .append(addr.d & 0xff)
-          .toString
-      case IPv6(a) => a.toString
-    }
+    def value: String
 
     override def render(writer: Writer): writer.type = this match {
       case RegName(n) => writer << n
       case a: Ipv4Address => writer << a.value
-      case IPv6(a) => writer << '[' << a << ']'
+      case a: Ipv6Address => writer << '[' << a << ']'
       case _ => writer
     }
   }
-
-  final case class RegName(host: CaseInsensitiveString) extends Host
 
   @deprecated("Renamed to Ipv4Address, modeled as case class of bytes", "0.21.0-M2")
   type IPv4 = Ipv4Address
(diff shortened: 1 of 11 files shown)


core/src/main/scala/org/http4s/Uri.scala
```diff
@@ -329,31 +330,16 @@ object Uri {
...
-          .toString
-      case IPv6(a) => a.toString
-    }
+    def value: String
 
     override def render(writer: Writer): writer.type = this match {
       case RegName(n) => writer << n
       case a: Ipv4Address => writer << a.value
-      case IPv6(a) => writer << '[' << a << ']'
+      case a: Ipv6Address => writer << '[' << a << ']'
```
rossabaker: We need to have a long navel-gazing over the differences between `render`, `value`, `toString`, and `Show`.  I'm only making it worse.

core/src/main/scala/org/http4s/parser/Rfc3986Parser.scala
```diff
@@ -69,9 +70,7 @@ private[http4s] trait Rfc3986Parser
   def Host: Rule1[org.http4s.Uri.Host] = rule {
     // format: off
     ipv4Address |
-    (IpLiteral | capture(IpV6Address)) ~> { s: String =>
-      org.http4s.Uri.IPv6(s.ci)
-    } |
+    "[" ~ ipv6Address ~ "]" |
```
rossabaker: I think we temporarily lost support for IPvFuture here.  There wasn't a test, and I highly doubt anyone cares, but it needs to be brought back for compliance sake.

laws/src/main/scala/org/http4s/laws/discipline/ArbitraryInstances.scala
```diff
@@ -596,29 +596,27 @@ private[http4s] trait ArbitraryInstances {
...
-      opt(atMost(1, h16colon) |+| h16) |+| :: |+| times(3, h16colon) |+| ls32,
-      opt(atMost(2, h16colon) |+| h16) |+| :: |+| times(2, h16colon) |+| ls32,
-      opt(atMost(3, h16colon) |+| h16) |+| :: |+| opt(h16colon) |+| ls32,
-      opt(atMost(4, h16colon) |+| h16) |+| :: |+| ls32,
-      opt(atMost(5, h16colon) |+| h16) |+| :: |+| h16,
-      opt(atMost(6, h16colon) |+| h16) |+| ::
-    ).map(Uri.IPv6.apply)
+  implicit val http4sTestingArbitraryForIpv6Address: Arbitrary[Uri.Ipv6Address] = Arbitrary {
+    for {
+      a <- getArbitrary[Short]
```
rossabaker: I chose not to take on the cats-scalacheck dependency to get mapN.  Since we publish the arbitraries, I thought it better to stay light.

Talk me out of it.

ChristopherDavenport: These are equivalent, so I really don't care enough to argue.

It does the job right, but its not breaking. That being said. cats-scalacheck is one of those libraries that I think shouldn't be a microlibrary but instead in cats or at least recommended as these instances should be available everywhere.

core/src/main/scala/org/http4s/Uri.scala
ChristopherDavenport: I strongly dislike the mutable renderers.

rossabaker: I was just thinking about this mutability last night, and came to the conclusion that it's a side effect I can live with.  It's absolutely necessary to have something like it, because rendering message headers is String concatenation: I don't think there's a practical way around having a `StringBuilder` under the hood.  Encoding 

* We could just target a ByteVector, but encoding to bytes on the fly has proven to be slower than building a String and encoding that.  That's why, unlike akka-http, we don't have a byte-based writer.
* We might accumulate a `Chain[String]` and flatten it for encoding, but I think that would increase GC pressure.  It would also not support the phantom writing of headers we're starting to do for multipart.
* Another answer is to suspend `Writer` operations into an effect, but these writers don't get shared.  It's the functional answer, but I'm not sure it's an appropriate engineering tradeoff here.

Maybe this merits a quick proof-of-concept benchmark.

laws/src/main/scala/org/http4s/laws/discipline/ArbitraryInstances.scala
rossabaker: I was worried about the circular dependency and wary of relying on it, but I wasn't thinking clearly.

1. Putting it into scalacheck would be a circular dependency.  That would be bad.
2. Putting it into cats-laws would not add any new dependencies.
3. It can live on its own, with the only constraint that it not upgrade scalacheck higher than what's in cats-laws for the corresponding cats.  It's just a chore, not an impossibility, to make a compatible release.

We shouldn't be so scared of the status quo (3), but I see no reason not to implement (2).

PR #2708, opened 2019-07-09
Release v0.20.5

rossabaker:
Only the blaze upgrade, but if you're affected, highly worth it.


ChristopherDavenport:
Don't we need to bump version as well?

rossabaker:
D'oh.

PR #2714, opened 2019-07-09
Release v0.20.6

rossabaker:


project/Http4sPlugin.scala
```diff
@@ -183,6 +183,7 @@ object Http4sPlugin extends AutoPlugin {
         tagRelease.when(publishable && release),
         runClean,
         releaseStepCommandAndRemaining("+mimaReportBinaryIssues"),
+        releaseStepCommandAndRemaining("""sonatypeOpen "Release via Travis""""),
```
rossabaker: This should prevent the mishap that ate the tail end of 0.20.5.

ChristopherDavenport: How so? For those of us trying to understand.

rossabaker: It creates a staging repo, gets the number, and sets it as `sonatypeStagingRepositoryProfile` and `publishTo`.  That way if concurrent builds are running in the organization, there are explicit, isolated targets, instead of publishing to whatever else happens to be open.

Issue #2670, opened 2019-06-24
Provide QueryParamMatchers for the various java.time data types


LukaJCB:

kevinmeredith:
Do you have any opinions on which [java.time.format.DateTimeFormatter](https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html) is appropriate, @LukaJCB? How about `ISO_OFFSET_DATE_TIME`?

```scala
@ import java.time._, java.time.format._

@ DateTimeFormatter.ISO_OFFSET_DATE_TIME.parse(Instant.now.toString) 
res21: temporal.TemporalAccessor = {OffsetSeconds=0, InstantSeconds=1562730675},ISO resolved to 2019-07-10T03:51:15.050

@ Instant.from 
def from(x$1: java.time.temporal.TemporalAccessor): java.time.Instant
@ Instant.from(res21) 
res22: Instant = 2019-07-10T03:51:15.050Z
```

kevinmeredith:
@LukaJCB - https://github.com/http4s/http4s/pull/2724 added support for `Instant`. What other types would you find useful - `ZonedDateTime` and `LocalDate`?

m-sp:
I think we can close this one right? feel free to reopen if this is not completely done yet

PR #2724, opened 2019-07-16
Add QueryParamEncoder[Instant] and QueryParamDecoder[Instant]

kevinmeredith:
Addresses part of https://github.com/http4s/http4s/issues/2670.

Please take a look, @rossabaker and @LukaJCB.

### DIFF
--- a/core/src/main/scala/org/http4s/QueryParam.scala
+++ b/core/src/main/scala/org/http4s/QueryParam.scala
@@ -40,6 +40,19 @@ object QueryParamKeyLike {
   }
 }
 
+trait QueryParamCodec[T] extends QueryParamEncoder[T] with QueryParamDecoder[T]
+object QueryParamCodec {
+
+  def apply[A](implicit instance: QueryParamCodec[A]): QueryParamCodec[A] = instance
+
+  def from[A](decodeA: QueryParamDecoder[A], encodeA: QueryParamEncoder[A]): QueryParamCodec[A] =
+    new QueryParamCodec[A] {
+      override def encode(value: A): QueryParameterValue = encodeA.encode(value)
+      override def decode(value: QueryParameterValue): ValidatedNel[ParseFailure, A] =
+        decodeA.decode(value)
+    }
+}
+
 /**
   * Type class defining how to encode a `T` as a [[QueryParameterValue]]s
   * @see QueryParamCodecLaws
--- a/core/src/main/scala/org/http4s/QueryParam.scala
+++ b/core/src/main/scala/org/http4s/QueryParam.scala
@@ -91,7 +91,9 @@ object QueryParamEncoder {
   implicit lazy val longQueryParamEncoder: QueryParamEncoder[Long] = fromShow[Long]
 
   def instantQueryParamEncoder(formatter: DateTimeFormatter): QueryParamEncoder[Instant] =
-    QueryParamEncoder[String].contramap[Instant]{i: Instant => formatter.format(i)}
+    QueryParamEncoder[String].contramap[Instant] { i: Instant =>
+      formatter.format(i)
+    }
 
   implicit lazy val stringQueryParamEncoder: QueryParamEncoder[String] =
     new QueryParamEncoder[String] {
@@ -205,11 +207,15 @@ object QueryParamDecoder {
   def instantQueryParamDecoder(formatter: DateTimeFormatter): QueryParamDecoder[Instant] =
     new QueryParamDecoder[Instant] {
       override def decode(value: QueryParameterValue): ValidatedNel[ParseFailure, Instant] =
-        Either.catchOnly[DateTimeParseException]{
-          val x = formatter.parse(value.value)
-          Instant.from(x)
-        }.leftMap { e => ParseFailure("Failed to read Instant", e.getMessage)}
-         .toValidatedNel
+        Either
+          .catchOnly[DateTimeParseException] {
+            val x = formatter.parse(value.value)
+            Instant.from(x)
+          }
+          .leftMap { e =>
+            ParseFailure("Failed to read Instant", e.getMessage)
+          }
+          .toValidatedNel
     }
 
   implicit val uriQueryParamDecoder: QueryParamDecoder[Uri] =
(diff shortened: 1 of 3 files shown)


kevinmeredith:
It seems slightly better to me if a single function produces both a `QueryParamEncoder` and `QueryParamDecoder`. Otherwise, a `QueryParamEncoder[Instant]` could be created with `DateTimeFormatter.A`, and then `QueryParamDecoder[Instant]` could be constructed with `DateTimeFormatter.B`.

core/src/main/scala/org/http4s/QueryParam.scala
```diff
@@ -196,6 +204,20 @@ object QueryParamDecoder {
         value.value.validNel
     }
 
+  def instantQueryParamDecoder(formatter: DateTimeFormatter): QueryParamDecoder[Instant] =
+    new QueryParamDecoder[Instant] {
+      override def decode(value: QueryParameterValue): ValidatedNel[ParseFailure, Instant] =
+        Either
```
rossabaker: `Validated` has `catchOnly` and is a `Bifunctor`.  I don't think you need the `Either`.

kevinmeredith:
> Would a function that returns both return a tuple?

That seems reasonable to me, @rossabaker.

> There's no QueryParamCodec, or is that what you're proposing?

Good question. I wasn't proposing that route. Looking at https://github.com/circe/circe/blob/v0.12.0-M4/modules/core/shared/src/main/scala/io/circe/Codec.scala#L7-L8, it shows that `Codec` is for convenience purposes only. What are your thoughts, @rossabaker?

rossabaker:
There is a good discussion of codec at https://github.com/circe/circe/issues/133.  I tend to agree with the desiderata and conclusion there.

kevinmeredith:
> There is a good discussion of codec at circe/circe#133. I tend to agree with the desiderata and conclusion there.

Thanks for pointing me to that link, @rossabaker. Could you please review the recent changes that I've made?

core/src/main/scala/org/http4s/QueryParam.scala
kevinmeredith: Done - https://github.com/http4s/http4s/pull/2724/commits/924c996094f74d947046b9b7ea12178df0fe0852

kevinmeredith:
I closed, and then opened because a flaky test appears to have been the problem: https://travis-ci.org/http4s/http4s/jobs/568317642.

kevinmeredith:
Hey @ChristopherDavenport @LukaJCB, please take a look cc @rossabaker. Thanks!

core/src/main/scala/org/http4s/QueryParam.scala
```diff
@@ -40,6 +41,43 @@ object QueryParamKeyLike {
...
+  def apply[A](implicit instance: QueryParamCodec[A]): QueryParamCodec[A] = instance
+
+  def from[A](decodeA: QueryParamDecoder[A], encodeA: QueryParamEncoder[A]): QueryParamCodec[A] =
+    new QueryParamCodec[A] {
+      override def encode(value: A): QueryParameterValue = encodeA.encode(value)
+      override def decode(value: QueryParameterValue): ValidatedNel[ParseFailure, A] =
+        decodeA.decode(value)
+    }
+
+  private def instantQueryParamDecoder(formatter: DateTimeFormatter): QueryParamDecoder[Instant] =
```
rossabaker: Why would this and `instanctQueryParamDecoder` not go on the companions of their respective typeclasses?  The `instantQueryParamCodec` to join them makes sense, as does being explicit (because there's no default format).

core/src/main/scala/org/http4s/QueryParam.scala
```diff
@@ -61,6 +99,8 @@ object QueryParamEncoder {
   /** summon an implicit [[QueryParamEncoder]] */
   def apply[T](implicit ev: QueryParamEncoder[T]): QueryParamEncoder[T] = ev
 
+  def fromCodec[T](implicit ev: QueryParamCodec[T]): QueryParamEncoder[T] = ev
```
rossabaker: I don't think circe has these.  Thinking about how these might clash.

core/src/main/scala/org/http4s/QueryParam.scala
kevinmeredith: Good point, Ross. I'll address that.

kevinmeredith: I initially thought you suggested moving `instantQueryParamDecoder` and `instantQueryParamEncoder` to `QueryParamDecoder` and `QueryParamEncoder` respectively. However, I already took that approach.

Could you please say more on your recommended change, @rossabaker?

rossabaker: Aren't they in `QueryParamCodec` still?  Maybe I need to check this out and not squint at a GitHub diff.

core/src/main/scala/org/http4s/QueryParam.scala
rossabaker: I thought this was implicit when I first read it.  I feel better about it on second look.

core/src/main/scala/org/http4s/QueryParam.scala
kevinmeredith: My mistake, Ross. You're right, `instantQueryParamDecoder` and `instantQueryParamEncoder` are defined in `object QueryParamCodec`. I'll move them to their respective companions. Sorry about that.

kevinmeredith: Done in the commits added at https://github.com/http4s/http4s/pull/2724#commits-pushed-06b6a81.

kevinmeredith:
I'm figuring the build [failed](https://travis-ci.org/http4s/http4s/jobs/570299130) due to a flaky spec, @rossabaker?

> [error] Failed: Total 65, Failed 1, Errors 0, Passed 61, Skipped 3, Pending 1
> [error] Failed tests:
> [error] 	org.http4s.client.blaze.Http1ClientStageSpec

kevinmeredith:
Maybe https://github.com/http4s/http4s/pull/2724/commits/212e3503f626f5c2f9cb5ca3feb80f5223a7a285 will fix the, I'm guessing, flaky test?

Issue #2713, opened 2019-07-09
UserInfo property failure


rossabaker:
```
[info] HTTP codec laws must hold for HttpCodec[UserInfo]
[error] x HTTP codec.roundTrip (64 ms)
[error]  Falsified after 47 passed tests.
[error]  > Labels of failing property:Expected: Right(UserInfo(㆒+笠㥰섉␪,Some(ꍋ䀎ℰ)))
[error]  Received: Right(UserInfo(㆒ 笠㥰섉␪,Some(ꍋ䀎ℰ)))> ARG_0: UserInfo(㆒+笠㥰섉␪,Some(ꍋ䀎ℰ))
[error]  The seed is zZMgwztMco-6CQeHNcLXRQUVOWnZZrnAQmaVbQfb73E=
[error] org.http4s.UserInfoSpec.check(UserInfoSpec.scala:11)
[error] org.typelevel.discipline.specs2.mutable.Discipline.$anonfun$checkAll$2(Discipline.scala:15)
```

hamnis:
This is fixed? @rossabaker ?

m-sp:
jep that one was fixed, closing

PR #2727, opened 2019-07-16
Fix UserInfo with plus sign (#2713)

valydia:


tests/src/test/scala/org/http4s/UserInfoSpec.scala
```diff
@@ -26,6 +26,14 @@ class UserInfoSpec extends Http4sSpec {
       renderString(UserInfo("hi", Some(":/?#[]@"))) must_== "hi::%2F%3F%23%5B%5D%40"
     }
 
+    "skip encoding subdelims except '+' in username" in {
+      renderString(UserInfo("!$&'()*+,;=", None)) must_== "!$&'()*%2B,;="
+    }
+
+    "skip encoding subdelims except '+' in password" in {
+      renderString(UserInfo("hi", Some("!$&'()*+,;="))) must_== "hi:!$&'()*%2B,;="
+    }
```
rossabaker: These tests will break, but maybe it's good to explicitly test that all the subdelims aren't encoded.

tests/src/test/scala/org/http4s/UserInfoSpec.scala
```diff
@@ -72,4 +80,11 @@ class UserInfoSpec extends Http4sSpec {
       math.signum(a.compareTo(b)) must_== math.signum(a.compare(b))
     }
   }
+
+  "bug2713" should {
+    "roundTrip userinfo with plus sign" in {
+      val userInfo = UserInfo("username+", Some("password+"))
+      HttpCodec[UserInfo].parse(renderString(userInfo)) must_== Right(userInfo)
+    }
+  }
```
rossabaker: Definitely keep this test.

valydia:
@rossabaker Thanks for the guidance, I wasn't entirely sure which way to fix it...

core/src/main/scala/org/http4s/parser/Rfc3986Parser.scala
```diff
@@ -144,6 +144,6 @@ private[http4s] trait Rfc3986Parser
 
   def SubDelims = rule { "!" | "$" | "&" | "'" | "(" | ")" | "*" | "+" | "," | ";" | "=" }
 
-  protected def decode(s: String) = URLDecoder.decode(s, charset.name)
+  protected def decode(s: String) = UrlCodingUtils.urlDecode(s, charset)
```
rossabaker: I'm afraid this might break tests in the query string, which _do_ encode `+`.  I was thinking of replacing the `decode` calls in `UserInfo.Parser`.  But if it works, :+1:, because this is correct in more cases.

valydia: Ok I see, I think I should correct this, even if the ci passes, because I am not sure that explicit test cases for '+' exists ...

rossabaker: I confirmed that the query is _not_ using that function.  It's path and regname.

PR #2728, opened 2019-07-17
Preserve request caret in withPathInfo

### LINKED ISSUE #2726: Request#withPathInfo deletes PathInfoCaret
In a `Request`, the `uri.path` for the request is split in to two parts, `scriptName` and `pathInfo` using the `Request.Keys.PathInfoCaret`. This is used for example by the `Router` middleware, which moves the caret such that the prefix is removed from `pathInfo`, but still accessible via `scriptName`.

When a user (e.g. another middleware) calls `withPathInfo`, my expectation is that it would only replace the `pathInfo` of the current request, and leave the `scriptName` untouched. However, this method calls `withUri`, which deletes the `PathInfoCaret`.

More formally, I would expect that:
```
val r = req.withPathInfo(x)
r.scriptName == req.scriptName
r.pathInfo = x
```

rossabaker:
`Request#withPathInfo` should preserve the position of the caret, an internal attribute that splits the URI path into `scriptName` and `pathInfo`.

Fixes #2726.  See #1183, #1213 for more historical context.

### DIFF
--- a/core/src/main/scala/org/http4s/Message.scala
+++ b/core/src/main/scala/org/http4s/Message.scala
@@ -300,10 +300,9 @@ sealed abstract case class Request[F[_]](
   private def caret =
     attributes.lookup(Request.Keys.PathInfoCaret).getOrElse(0)
 
-  def withPathInfo(pi: String): Self = {
+  def withPathInfo(pi: String): Self =
     // Don't use withUri, which clears the caret
     copy(uri = uri.withPath(scriptName + pi))
-  }
 
   def pathTranslated: Option[File] = attributes.lookup(Keys.PathTranslated)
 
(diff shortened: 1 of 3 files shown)


core/src/main/scala/org/http4s/Message.scala
```diff
@@ -294,13 +294,19 @@ sealed abstract case class Request[F[_]](
...
-    withUri(uri.withPath(scriptName + pi))
+  private def caret =
+    attributes.lookup(Request.Keys.PathInfoCaret).getOrElse(0)
+
+  private def withCaret(caret: Int) =
+    withAttribute(Request.Keys.PathInfoCaret, caret)
+
+  def withPathInfo(pi: String): Self = {
+    val oldCaret = caret
+    withUri(uri.withPath(scriptName + pi)).withCaret(oldCaret)
```
nicktelford: Instead of having to lookup the caret, couldn't you just do:
```
copy(uri = uri.withPath(scriptName + pi))
```

rossabaker: :+1: That's a good point.  `withUri` clears the caret, but this can use `copy` directly.  I changed it to this.

PR #2733, opened 2019-07-20
feat: add some simple csrf documentation

tyrantkhan:
Could probably use some fleshing out, but probably a good enough base to get feedback for and I can build upon it -- or just have someone else flesh it out a bit more!.

### DIFF
--- a/docs/src/main/tut/csrf.md
+++ b/docs/src/main/tut/csrf.md
@@ -1,0 +1,76 @@
+---
+menu: main
+title: CSRF
+weight: 123
+---
+
+Http4s provides [Middleware], named `CSRF`, to avoid Cross-site request forgery attacks.This middleware 
+is modeled after the double submit cookie pattern: https://www.owasp.org/index.php/Cross-Site_Request_Forgery_(CSRF)_Prevention_Cheat_Sheet#Double_Submit_Cookie
+
+Examples in this document have the following dependencies.
+
+```scala
+libraryDependencies ++= Seq(
+  "org.http4s" %% "http4s-dsl" % http4sVersion,
+  "org.http4s" %% "http4s-server" % http4sVersion
+)
+```
+
+And we need some imports.
+
+```tut:silent
+  import cats.effect._
+  import org.http4s._
+  import org.http4s.dsl.io._
+  import org.http4s.implicits._
+  import org.http4s.headers.Referer
+  import org.http4s.server.middleware._
+```
+
+Let's start by making a simple service.
+
+```tut:book
+val service = HttpRoutes.of[IO] {
+  case _ =>
+    Ok()
+} 
+
+val request = Request[IO](Method.GET, uri"/")
+
+service.orNotFound(request).unsafeRunSync
+```
+
+Now we can build our CSRF Middleware by creating a CSRFBuilder
+
+```tut:silent
+val cookieName = "csrf-token"
+val key  = CSRF.generateSigningKey[IO].unsafeRunSync
+val defaultOriginCheck: Request[IO] => Boolean =
+  CSRF.defaultOriginCheck[IO](_, "localhost", Uri.Scheme.http, None)
+val csrfBuilder = CSRF[IO,IO](key, defaultOriginCheck)
+```
+
+More info on what is possible here: https://http4s.org/v0.21/api/org/http4s/server/middleware/csrf$$csrfbuilder but
+we will create for a fairly simple CSRF Middleware in our example.
+
+```tut:book
+val csrf = csrfBuilder.withCookieName(cookieName).withCookieDomain(Some("localhost")).withCookiePath(Some("/")).build
+```
+
+Now we need to wrap this around our service!
+```tut:book
+val dummyRequest: Request[IO] =
+    Request[IO](method = Method.GET).putHeaders(Header("Origin", "http://localhost"))
+val resp = csrf.validate()(service.orNotFound)(dummyRequest).unsafeRunSync()
+```
+
+Notice how the response has the CSRF cookies added. How easy was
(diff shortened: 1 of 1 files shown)


docs/src/main/tut/csrf.md
```diff
@@ -0,0 +1,72 @@
+---
+menu: main
+title: CSRF
+weight: 123
+---
+
+Http4s provides [Middleware], named `CSRF`, to avoid Cross-site request forgery attacks.This middleware 
+is modeled after the double submit cookie pattern: https://www.owasp.org/index.php/Cross-Site_Request_Forgery_(CSRF)_Prevention_Cheat_Sheet#Double_Submit_Cookie
+
+Examples in this document have the following dependencies.
```
rossabaker: This is nice and we should be better about this in all our tuts.

docs/src/main/tut/csrf.md
```diff
@@ -0,0 +1,72 @@
...
+  case _ =>
+    Ok()
+} 
+
+val request = Request[IO](Method.GET, uri"/")
+
+service.orNotFound(request).unsafeRunSync
+```
+
+Now we can build our CSRF Middleware by creating a CSRFBuilder
```
rossabaker: Backtick code literals:

```suggestion
Now we can build our CSRF Middleware by creating a `CSRFBuilder`
```

docs/src/main/tut/csrf.md
```diff
@@ -0,0 +1,72 @@
...
+
+```tut:silent
+val cookieName = "csrf-token"
+val key  = CSRF.generateSigningKey[IO].unsafeRunSync
+val defaultOriginCheck: Request[IO] => Boolean =
+  CSRF.defaultOriginCheck[IO](_, "localhost", Uri.Scheme.http, None)
+val csrfBuilder = CSRF[IO,IO](key, defaultOriginCheck)
+```
+
+More info on what is possible here: https://http4s.org/v0.21/api/org/http4s/server/middleware/csrf$$csrfbuilder but
```
rossabaker: I _think_ this will expand to `0.21` so it stays current.  It would also be nice to link that instead of embedding a long URL.

`{{< version "http4s.api" >}}`

docs/src/main/tut/csrf.md
```diff
@@ -0,0 +1,72 @@
...
+```tut:book
+val dummyRequest: Request[IO] =
+    Request[IO](method = Method.GET).putHeaders(Header("Origin", "http://localhost"))
+val resp = csrf.validate()(service.orNotFound)(dummyRequest).unsafeRunSync()
+```
+
+Notice how the response has the CSRF cookies added. How easy was
+that? And, as described in [Middleware], services and middleware can be
+composed such that only some of your endpoints are CSRF enabled. By default, 
+safe methods will update the CSRF token, while unsafe methods will validate them.
```
rossabaker: It might be worth explaining what "safe" and "unsafe" methods are.

docs/src/main/tut/csrf.md
tyrantkhan: Saw other docs were doing . something like this:

[CSRFBuilder]: ../api/org/http4s/server/middleware/csrf$$csrfbuilder

tyrantkhan:
I posted this in the gitter, but I want to make note of an issue in this example documentation, I feel like  this might be a bug:

> hey y’all… I’ve been finalizing @rossabaker ’s comments on this PR, and I realized I didn’t even try to incorporate a POST in the documentation
https://github.com/http4s/http4s/blob/d39e933120ca0f20824c596046e29d099e739e1b/docs/src/main/tut/csrf.md
I’m getting a 403 and I feel like the origin check is probably failing, and I can’t seem to see why that would be the case. Unless a bug in the code? I’ve even simplified the origin check to be something simple like request.toString.length >= 0 and it still failed. So maybe another issue exists?
You can use the example in the documentation to produce the 403. Am I missing something?
Haris Khan @tyrantkhan Jul 27 21:38
Additionally applying that same predicate to validate() does seem to indicate it is passing the origin check

rossabaker:
I haven't forgotten this.  Just going to take a little extra to dig into the 403 issue and ran out of time on my lunchtime pass.  I'll come back to this ASAP.

tyrantkhan:
@rossabaker i think this is good to go,  unless you have any more feedback.

docs/src/main/tut/csrf.md
```diff
@@ -0,0 +1,90 @@
...
+
+Without getting too deep into it, safe methods are OPTIONS, GET, and HEAD. While unsafe methods are 
+POST, PUT, PATCH, DELETE, and TRACE. To put it simply, state changing methods are unsafe. For more information,
+check out this cheat sheet on [CSRF Prevention](https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html#double-submit-cookie).
+
+Unsafe requests (like POST) require us to send the CSRF token in the `X-Csrf-Token`
+header (this is the default name, but it can be changed), so we are going to get the value
+and send it up in our POST. I've also added the response cookie as a RequestCookie, normally
+the browser would send this up with our request, but I needed to do it manually for the purpose of this demo.
```
rossabaker: This is sad, but a cookie jar is coming soon. Let's make sure we revisit this then. /cc @ChristopherDavenport

ChristopherDavenport: Available in its Beta form. https://github.com/ChristopherDavenport/cookiejar

Just trying to figure out what I'm supposed to do if no domain is present in the cookie. What the constraints to either what it should be saved as, or to what servers it applies. Its not clear to me from the RFC.

Issue #2690, opened 2019-07-04
Client hangs indefinitely on connection, ignores `requestTimeout` setting


kamilkloch:
For what it's worth, both `curl` and `sttp` client cancel the same connection after the provided timeout.

rossabaker:
Which backend are you using?  In what context are you seeing this?  Do you have code that would help reproduce this?

And yes, we have cancellation as well.

kamilkloch:
> Which backend are you using? In what context are you seeing this? Do you have code that would help reproduce this?

Version `0.20.3`. I am querying a host which is filtered on the firewall. Requests hang and slowly eat up client's wait queue until it's full. 

> And yes, we have cancellation as well.
Do you mean Timeout middleware?

RafalSumislawski:
Looks very similar to #2386. #2386 is fixed in a way which makes it *eventually* (15 seconds on my machine) time out. It doesn't respect the configured timeout. May this be what you're experiencing? Or do they actually **never** time out?

kamilkloch:
@RafalSumislawski The request is stuck forever. Behaviour is reproducible with the snippet from #2386, just replace `http://1.2.3.4:nn` with any host:port so that the port is filtered on the host firewall.

RafalSumislawski:
@kamilkloch On what OS are you trying it? The behaviour of establishing TCP connections differs between OSes.

kamilkloch:
@RafalSumislawski Ubuntu.

RafalSumislawski:
One more thing: What version of cats-effect do you have on your classpath?

kamilkloch:
1.3.0.

kevinmeredith:
Hi @RafalSumislawski - do you expect https://github.com/http4s/blaze/pull/308 and/or https://github.com/http4s/http4s/pull/2736 have resolved this issue? I'm asking as an interested user of this library. Thanks

RafalSumislawski:
Hi
Yes, I expect they've solved it, although I can't be sure, as I wasn't able to reproduce it.

rossabaker:
I think it did. Please reopen if it's still a problem.

PR #2736, opened 2019-07-21
Connecting timeout

RafalSumislawski:
Follow-up to https://github.com/http4s/blaze/pull/308.

WIP: It won't compile until the aforementioned PR is merged and blaze 0.14.7 is released.

This PR:
- gives users ability to provide custom TickWheelExecutor in place of the one allocated by BlazeClientBuilder
- adds `connectingTimeout` for configuring the newly introduced `connectingTimeout` in blaze's `ClientChannelFactory`

Fixes: #2386, #2338, #2690

### DIFF
--- a/blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala
+++ b/blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala
@@ -2,6 +2,7 @@ package org.http4s
 package client
 package blaze
 
+import cats.implicits._
 import cats.effect._
 import java.nio.channels.AsynchronousChannelGroup
 
@@ -37,7 +38,7 @@ sealed abstract class BlazeClientBuilder[F[_]] private (
     val parserMode: ParserMode,
     val bufferSize: Int,
     val executionContext: ExecutionContext,
-    val scheduler: Option[TickWheelExecutor],
+    val scheduler: Resource[F, TickWheelExecutor],
     val asynchronousChannelGroup: Option[AsynchronousChannelGroup],
     val channelOptions: ChannelOptions
 )(implicit protected val F: ConcurrentEffect[F])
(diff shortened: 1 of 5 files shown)


blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala
```diff
@@ -20,6 +23,7 @@ sealed abstract class BlazeClientBuilder[F[_]] private (
     val responseHeaderTimeout: Duration,
     val idleTimeout: Duration,
     val requestTimeout: Duration,
+    val connectingTimeout: Duration,
```
rossabaker: Bikeshed: async-http-client and jetty-client both call this `connectTimeout`.  Maybe that's a better name for consistency.  Would also affect the blaze PR.

blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala
```diff
@@ -33,6 +37,7 @@ sealed abstract class BlazeClientBuilder[F[_]] private (
     val parserMode: ParserMode,
     val bufferSize: Int,
     val executionContext: ExecutionContext,
+    val scheduler: Option[TickWheelExecutor],
```
rossabaker: I think I like the [Ember idiom for defaultable resources](https://github.com/http4s/http4s/blob/05501d7f07481ead542de44e516f693f7d32a673/ember-client/src/main/scala/org/http4s/ember/client/EmberClientBuilder.scala#L56), where the builder provides a default, or an unmanaged one can be passed in.

RafalSumislawski:
The build fails due to binary compatibility issue in BlazeClientBuilder and Http1Supoport constructors. Both constructors are private. Do we really care about binary compatibility of private APIs? @rossabaker how should I approach this topic?

rossabaker:
It should be safe to add a MiMa exclusion to the build for both of those.

rossabaker:
I think both test failures are flakes.  Restarted.

PR #2740, opened 2019-07-27
Clarify Method imports for client DSL in docs

sloshy:
Clarifies the documentation to specify that `org.http4s.Methods._` is imported into scope with the default routing DSL, and that the client DSL needs an extra import in order for some of the examples to compile.

Changes could/should also be made for the current 0.20.x documentation, if it all looks good I'll write this up for that as well.


rossabaker:
Thanks, @sloshy.  I think the build is failing because it's `Method`, not `Methods`.  The clarification is helpful!

sloshy:
@rossabaker You're right - my mistake. Should be fixed up now.

Issue #2550, opened 2019-05-03
responseHeaderTimeout should be < idleTimeout


RafalSumislawski:
The `BlazeServerSpec` says ["A server should return a 503 if the server doesn't respond"](https://github.com/http4s/http4s/blob/master/blaze-server/src/test/scala/org/http4s/server/blaze/BlazeServerSpec.scala#L87) and it works like that in the test, but it does only because [response header time-out is set to 1 second](https://github.com/http4s/http4s/blob/master/blaze-server/src/test/scala/org/http4s/server/blaze/BlazeServerSpec.scala#L17). The default for `responseHeaderTimeout ` is 1 minute while the default for `idleTimeout` is 30 seconds (see: [BlazeServerBuilder](https://github.com/http4s/http4s/blob/master/blaze-server/src/main/scala/org/http4s/server/blaze/BlazeServerBuilder.scala#L402-L403) ). That means that when running with default configuration, the `idleTimeout` kicks in first, and instead of the nice behaviour of returning HTTP 503 response, we get an abrupt TCP connection termination, leaving the client with no clue what went wrong. 

IMO the defaults for the timeouts should be changed so that the `responseHeaderTimeout` is shorter than the `idleTimeout`. In addition to that, the blaze server could log a warning it is started with `responseHeaderTimeout >= idleTimeout`.

rossabaker:
I agree with all of this, but what do we propose as the new values?

RafalSumislawski:
I did some research on default timeouts used in other http client/server libraries:

- [`akka-http` ](https://github.com/akka/akka-http/blob/eb7311130021c6fa72dff05090c20568ee1319dd/akka-http-core/src/main/resources/reference.conf)
  - client-side: connecting-timeout = 10s
  - client-side: request-timeout not implemented
  - server-side: request-timeout = 20s
  - client-side: idle-timeout = 60 s
  - server-side: idle-timeout = 60 s
- [`lihaoyi/requests-scala`] (https://github.com/lihaoyi/requests-scala/blob/0c654ef3948e8e29c67d968ab4d08c325c1663ea/requests/src/requests/package.scala)
  - client-side: connectTimeout = 10s
  - client-side: readTimeout = 10s 
- [`sttp`](https://sttp.readthedocs.io/en/latest/conf/timeouts.html)
  - client-side: Connection timeout = 30s
  - client-side: Read timeout = 60s
[...]

PR #2746, opened 2019-07-30
Adjust default timeouts and add warnings about missconfiguration

RafalSumislawski:
As described in #2550

### DIFF
--- a/blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala
+++ b/blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala
@@ -185,6 +185,7 @@ sealed abstract class BlazeClientBuilder[F[_]] private (
   def resource: Resource[F, Client[F]] =
     scheduler.flatMap { scheduler =>
       verifyAllTimeoutsAccuracy(scheduler)
+      verifyTimeoutRelations()
       connectionManager(scheduler).map { manager =>
         BlazeClient.makeClient(
           manager = manager,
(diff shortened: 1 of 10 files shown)


blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala
```diff
@@ -185,6 +185,7 @@ sealed abstract class BlazeClientBuilder[F[_]] private (
   def resource: Resource[F, Client[F]] =
     scheduler.flatMap { scheduler =>
       verifyAllTimeoutsAccuracy(scheduler)
+      verifyTimeoutRelations()
```
ChristopherDavenport: These impure calls are in pure position.

RafalSumislawski: Indeed these methods write some logs. I don't see any examples of logging being handled as an effect in http4s code base, so that's the pattern I followed. Do you want me to wrap them in `F`?

rossabaker: `F.delay(logger` will show a few examples.  log4cats is only a dependency of the ember modules.

client/src/main/scala/org/http4s/client/package.scala
```diff
@@ -1,6 +1,15 @@
 package org.http4s
 
-package object client extends ClientTypes
+import scala.concurrent.duration.DurationLong
+
+package object client extends ClientTypes {
+
+  object defaults {
```
rossabaker: I would expect this to generate a compiler warning about not putting objects in a package object.  I think this wants to live at the top-level in the `client` package.

rossabaker: Although now I see we've already done this in `server`. :thinking:

RafalSumislawski:
This branch is base on 0.20 as it was my original intention to target that version, so we can merge to 0.20 and then to master, without cherry-picking.

server/src/main/scala/org/http4s/server/package.scala
```diff
@@ -16,20 +16,22 @@ import java.net.{InetAddress, InetSocketAddress}
...
          | | ' \  _|  _| '_ \_  _(_-<
          | |_||_\__|\__| .__/ |_|/__/
          |             |_|""".stripMargin.split("\n").toList
+
     val Host = InetAddress.getLoopbackAddress.getHostAddress
     val HttpPort = 8080
-    val IdleTimeout: Duration = 30.seconds
+    val SocketAddress = InetSocketAddress.createUnresolved(Host, HttpPort)
+
+    val ResponseTimeout: Duration = 30.seconds
```
rossabaker: Adding this should fix the binary compatibility issue:

```scala
    @deprecated("Renamed to ResponseTimeout", "0.21.0-M3")
    def AsyncTimeout: Duration = ResponseTimeout
```

PR #2747, opened 2019-07-30
Create .mergify.yml

ChristopherDavenport:
Starting out a mergify integration.

This automatically merges semver-patch PR's from Scala-steward that pass CI.


ChristopherDavenport:
Correct that pr ends with `labels:` which does not match the regex.

PR #2776, opened 2019-08-08
Add MaxActiveRequest middleware

ChristopherDavenport:
Initial Code for a MaxActiveRequest middleware. That will immediately return the default response if all requests allowed are currently active.

Can get test code up tommorow.

### DIFF
--- a/server/src/main/scala/org/http4s/server/middleware/MaxActiveRequests.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/MaxActiveRequests.scala
@@ -11,25 +11,24 @@ import org.http4s.{Request, Response}
 
 object MaxActiveRequests {
 
-
   def httpApp[F[_]: Concurrent](
-    maxActive: Long,
-    defaultResp: Response[F] = Response[F](status = Status.ServiceUnavailable)
-  ): F[Kleisli[F, Request[F], Response[F]] => Kleisli[F, Request[F], Response[F]]] = 
+      maxActive: Long,
+      defaultResp: Response[F] = Response[F](status = Status.ServiceUnavailable)
+  ): F[Kleisli[F, Request[F], Response[F]] => Kleisli[F, Request[F], Response[F]]] =
     inHttpApp[F, F](maxActive, defaultResp)
 
   def inHttpApp[G[_]: Sync, F[_]: Concurrent](
-    maxActive: Long,
-    defaultResp: Response[F] = Response[F](status = Status.ServiceUnavailable)
-  ): G[Kleisli[F, Request[F], Response[F]] => Kleisli[F, Request[F], Response[F]]] = {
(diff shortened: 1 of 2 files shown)


server/src/main/scala/org/http4s/server/middleware/MaxActiveRequests.scala
```diff
@@ -0,0 +1,55 @@
...
+          if (_) sem.release
+          else Sync[F].unit
+        }
+        
+      }
+    }
+  }
+
+  /**
+   * Caveats: Body Of the Response Must Be Run.
```
rossabaker: This should be guaranteed by the server, so I'm not so worried about the caveat.  I understand the difference, but I wonder whether `byHeaders` is all that useful.

This is another case where a resource model for the response would be highly useful, though we ran into other friction with that.  I can't remember why we didn't make a `bracketResponse` combinator.

ChristopherDavenport: Alright, going with the singular interface. Its cleaner by a longshot.

ChristopherDavenport: Interestingly, the byHeaders actually abstracts over all Kleislis.

server/src/main/scala/org/http4s/server/middleware/MaxActiveRequests.scala
```diff
@@ -0,0 +1,70 @@
...
+  def inHttpRoutes[G[_]: Sync, F[_]: Concurrent](
+    maxActive: Long,
+    defaultResp: Response[F] = Response[F](status = Status.ServiceUnavailable)
+  ): G[Kleisli[OptionT[F, ?], Request[F], Response[F]] => Kleisli[OptionT[F, ?], Request[F], Response[F]]] = {
+    Semaphore.in[G, F](maxActive).map{ sem => 
+      {http: Kleisli[OptionT[F, ?], Request[F], Response[F]] => 
+        Kleisli{a: Request[F] => 
+          Concurrent[OptionT[F, ?]].bracketCase(OptionT.liftF(sem.tryAcquire)){ bool => 
+            if (bool) http.run(a).map(resp => resp.copy(body = resp.body.onFinalize(sem.release)))
+              .orElseF(sem.release.as(None))
```
rossabaker: I'm getting to the stupid hours early today.  Is there not something with `~>` that DRYs up the two big implementations?  Is this `.as(None)` indicative of the problem?

ChristopherDavenport: Exactly. The problem is we need to release on the None case specifically if we know its OptionT. So if we make it generic we would need `OptionT[F, ?] ~> F` which is not a transformation we can provide.

PR #2777, opened 2019-08-08
handle invalid content range

eklavya:
FileService should send RangeNotSatisfiable when range is invalid.

### DIFF
--- a/server/src/main/scala/org/http4s/server/staticcontent/FileService.scala
+++ b/server/src/main/scala/org/http4s/server/staticcontent/FileService.scala
@@ -7,6 +7,7 @@ import cats.effect.{Blocker, ContextShift, Effect, Sync}
 import java.io.File
 import org.http4s.headers.Range.SubRange
 import org.http4s.headers._
+import cats.implicits._
 
 object FileService {
   type PathCollector[F[_]] = (File, Config[F], Request[F]) => OptionT[F, Response[F]]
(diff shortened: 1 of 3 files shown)


server/src/main/scala/org/http4s/server/staticcontent/FileService.scala
```diff
@@ -80,28 +80,31 @@ object FileService {
...
+                StaticFile.calcETag)
+              .map { resp =>
+                val hs: Headers = resp.headers
+                  .put(AcceptRangeHeader, `Content-Range`(SubRange(start, end), Some(size)))
+                resp.copy(status = Status.PartialContent, headers = hs)
+              }
+              .value
+          }
+        } else {
+          F.pure(Some(Response[F](status = Status.RangeNotSatisfiable)))
```
rossabaker: From [the spec](https://tools.ietf.org/html/rfc7233#page-15):

> When this status code is generated in response to a byte-range request, the sender SHOULD generate a Content-Range header field specifying the current length of the selected representation (Section 4.2).

We should try to add that header.

eklavya:
Sorry @rossabaker I was just sure this wasn't gonna break anything 👅

rossabaker:
I think it's okay to change that test... the 200 it was testing for is made better by this.

server/src/main/scala/org/http4s/server/staticcontent/FileService.scala
```diff
@@ -80,28 +80,39 @@ object FileService {
...
+              .map { resp =>
+                val hs: Headers = resp.headers
+                  .put(AcceptRangeHeader, `Content-Range`(SubRange(start, end), Some(size)))
+                resp.copy(status = Status.PartialContent, headers = hs)
+              }
+              .value
+          }
+        } else {
+          F.suspend {
+            val size = file.length()
```
rossabaker: We get the requisite laziness from `suspend`, but I think think this would read more naturally:

```scala
F.delay(file.length()).map { size =>
  Some(???)
}
```

server/src/test/scala/org/http4s/server/staticcontent/FileServiceSpec.scala
```diff
@@ -80,8 +80,7 @@ class FileServiceSpec extends Http4sSpec with StaticContentShared {
       )
       val reqs = ranges.map(r => Request[IO](uri = uri("testresource.txt")).withHeaders(r))
       forall(reqs) { req =>
-        routes.orNotFound(req) must returnStatus(Status.Ok)
-        routes.orNotFound(req) must returnBody(testResource)
+        routes.orNotFound(req) must returnStatus(Status.RangeNotSatisfiable)
```
rossabaker: I'd throw in a test for the header here as well.

rossabaker:
Incorporated into #2790.

PR #2778, opened 2019-08-10
fix: we are slicing keys and losing entropy.

tyrantkhan:
This resulted from a conversation with @jmcardon, we determined,
that we do not want to slice the key.

Reference: https://gitter.im/http4s/http4s?at=5d34da17e2d1aa6688e6cce4

### DIFF
--- a/server/src/main/scala/org/http4s/server/middleware/CSRF.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/CSRF.scala
@@ -500,7 +500,7 @@ object CSRF {
     * any amount less than 20 bytes will throw an exception when loaded
     * into `Mac`. Any keys larger than 64 bytes are just hashed.
     *
-    * For more information, refer refer to: https://tools.ietf.org/html/rfc2104#section-3
+    * For more information, refer to: https://tools.ietf.org/html/rfc2104#section-3
     *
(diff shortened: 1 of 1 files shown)


server/src/main/scala/org/http4s/server/middleware/CSRF.scala
```diff
@@ -446,7 +446,6 @@ object CSRF {
   ///
 
   val SigningAlgo: String = "HmacSHA1"
-  val SHA1ByteLen: Int = 20
```
rossabaker: If we deprecate this value, we could release this as a fix on series/0.20.

tyrantkhan: I think we should (deprecate it). Do I need to do anything on my side to do that?

rossabaker: I just went ahead and did it, because it's easier than taking a reviewed PR and moving it to a different base.  But deprecating a value [looks like this](https://github.com/http4s/http4s/blob/series/0.20/server/src/main/scala/org/http4s/server/middleware/CSRF.scala#L449-L450).

rossabaker:
Cherry-picked as 579c9a619a, and will be in the next 0.20 release.

PR #2781, opened 2019-08-12
AHC: work around duplicate request-body-streamer

rossabaker:
See https://github.com/AsyncHttpClient/async-http-client/issues/1660 for a deeper discussion.  This is terrible, but is something we can do without an AHC patch.


rossabaker:
Cool!  Now AHC can join the party of Tests That Fail Only On Travis CI!

rossabaker:
MaxRequestsSpec failure tracked at #2782.

rossabaker:
This causes deeper problems.

PR #2446, opened 2019-03-07
Extensible dsl

hamnis:
Add new type constructor to be able to construct this DSL for http4s-directives.

```
trait DirectivesDsl[F[_]] extends Http4sDsl2[Directive[F, ?], F]

object DirectivesDsl {
  def apply[F[_]]: DirectivesDsl[F] = new DirectivesDsl[F] {}
}


object io extends DirectivesDsl[IO]
```

PR #2785, opened 2019-08-13
Fix EntityResponseGenerator

rintcius:
We got a failing test in our test suite after trying to upgrade http4s. After some digging it appeared to be introduced in https://github.com/http4s/http4s/pull/2446

This PR fixes the failing test for us.

Take `body: F[A]` as parameter instead of `G[A]` which enables following the code to be structured as it was here https://github.com/http4s/http4s/pull/2446/files#diff-09e729dc8ff0315c384e774351a8720fL42

If 👍 I'm happy to make analogous change to other branches.

### DIFF
--- a/dsl/src/main/scala/org/http4s/dsl/impl/ResponseGenerator.scala
+++ b/dsl/src/main/scala/org/http4s/dsl/impl/ResponseGenerator.scala
@@ -38,19 +38,8 @@ trait EntityResponseGenerator[F[_], G[_]] extends Any with ResponseGenerator {
   def apply(headers: Header*)(implicit F: Applicative[F]): F[Response[G]] =
     F.pure(Response[G](status, headers = Headers(`Content-Length`.zero :: headers.toList)))
 
-  def apply[A](body: G[A])(implicit F: Monad[F], w: EntityEncoder[G, A]): F[Response[G]] = {
-    val entity = Entity(fs2.Stream.eval(body).flatMap(w.toEntity(_).body))
-    val headers = {
-      val h = w.headers
-      entity.length
-        .map { l =>
-          `Content-Length`.fromLong(l).fold(_ => h, c => h.put(c))
-        }
-        .getOrElse(h)
-    }
-
-    F.pure(Response[G](status = status, headers = headers, body = entity.body))
-  }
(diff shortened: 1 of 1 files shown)


djspiewak:
> I'm surprised that this is binary compatible, but type erasure saves us here

Unless there are some constraints on `G`, I'm pretty sure this is binary *incompatible* just in a way that mima won't catch.

rossabaker:
The constraint is `EntityEncoder[G, A]`.

For most people, `F` is `G`. And this will be mostly applications calling it. But, eh, maybe this would cause a ClassCastException in the encoder for the unlucky few? 🤔

djspiewak:
Could also cause a `ClassCastException` in the `Monad[F].flatMap`. I do agree that for most people, those two would probably be unified. I guess it depends on how fussed you want to be about it. In theory, the old function was just… broken… so even breaking that code is kind of justified? I dunno.

hamnis:
This broke the DirectivesDsl.
I suppose we will have to live with: 
`BadRequest(Directive.liftF(IO("Output")))`

hamnis:
If we instead change this to
```scala
  def apply[A](body: G[A])(implicit nat: G ~> F ,F: Monad[F], w: EntityEncoder[G, A]): F[Response[G]] =
    F.flatMap(nat[A](body))(apply[A](_))
```
then it works in both cases

rossabaker:
Sorry. I thought it was a mistake, but I remember that `Http4sDsl2` was added for the directives DSL, and should have tried that or pinged you.

Since it broke working code, and is of questionable binary compatibility, I think we should revert for 0.20.

rossabaker:
As for 0.21, I don't think of `~>` as something we usually pass around as implicit evidence.  I wonder about just giving the `F` version a different name than `apply`.

hamnis:
We might then need a different way of plugging in the lifting of `G` to `F` in 0.21 ? I can try to come up with something.

rossabaker:
Yeah, my first priority here is that we not break http4s-directives. If we can come up with something clean for both that and the use case that motivated this original PR, that's the best thing for 0.21... and if it's compatible, for 0.20.

PR #2787, opened 2019-08-13
Merge Train

ChristopherDavenport:


rossabaker:
Did this squash the commits from 0.20? That would cause conflicts the next time we merge from 0.20.

ChristopherDavenport:
Going to merge in my fix here. Will resolve before we merge. As that fix is critical.

rossabaker:
I propose we drop this for #2789. I unsquashed a9a5183 so we can keep merging, and then applied the scalafmt.

PR #2793, opened 2019-08-14
Update jetty-client, jetty-server, ... to 9.4.20.v20190813

rossabaker:
Supersedes #2791.


rossabaker:
Merging because it's the defunct 2.13.0-M5 build that failed.

PR #2811, opened 2019-08-21
Entity: added mediaType/charset fields

IndiscriminateCoding:
### The problem:
`EntityDecoder` interface is not consistent with `EntityEncoder`.

`EntityEncoder` returns an `Entity`, while `EntityDecoder` only accepts a `Message[F]`.

This is problematic because it is not possible to decode arbitrary `EntityBody` using `EntityDecoder` even if you know its Media Type.

Also it is not possible to return different media types from one `EntityEncoder`.

### Proposal:
Interface of `Entity` is changed - this change adds (optional) `mediaType` and `charset` fields, so now it is possible to specify what is inside the body.

`EntityDecoder` now accepts an `Entity` - however, there is overload for `decode` method that accepts a `Message`.

`EntityEncoder` doesn't have `contentType` and `charset` methods anymore - instead it returns an `Entity` with appropriate `mediaType` and `charset`.

**P.S.** Tests are not compiling now; I want to know what do you guys think about this change and then I'll fix it.

### DIFF
--- a/argonaut/src/test/scala/org/http4s/argonaut/ArgonautSpec.scala
+++ b/argonaut/src/test/scala/org/http4s/argonaut/ArgonautSpec.scala
@@ -8,7 +8,6 @@ import java.nio.charset.StandardCharsets
 
 import org.http4s.Status.Ok
 import org.http4s.argonaut._
-import org.http4s.headers.`Content-Type`
 import jawn.JawnDecodeSupportSpec
 import org.specs2.specification.core.Fragment
 
@@ -35,8 +34,7 @@ class ArgonautSpec extends JawnDecodeSupportSpec[Json] with Argonauts {
     val json = Json("test" -> jString("ArgonautSupport"))
 
     "have json content type" in {
-      jsonEncoder.headers.get(`Content-Type`) must_== Some(
-        `Content-Type`(MediaType.application.json))
+      jsonEncoder.toEntity(json).mediaType must_== Some(MediaType.application.json)
     }
 
     "write compact JSON" in {
@@ -60,8 +58,7 @@ class ArgonautSpec extends JawnDecodeSupportSpec[Json] with Argonauts {
 
   "jsonEncoderOf" should {
     "have json content type" in {
-      jsonEncoderOf[IO, Foo].headers.get(`Content-Type`) must_== Some(
-        `Content-Type`(MediaType.application.json))
+      jsonEncoderOf[IO, Foo].toEntity(foo).mediaType must_== Some(MediaType.application.json)
     }
 
     "write compact JSON" in {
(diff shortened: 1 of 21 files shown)


rossabaker:
Thanks for this.  I'm too tired to give the code a review tonight, but I can give some kneejerk reactions.

1. I like the symmetry.

2. `contentType` and `charset` being static properties of the `EntityEncoder` help with static inference, and (I am pretty sure) is exploited by rho.  But keeping them static is bad for multipart, where the content type depends on the boundary in the value.  This is a longstanding trade-off.  It would be nice to find a solution that's better for multipart and not harmful to rho. /cc @zarthross

3. I've been tempted to make `Entity` a first-class type on `Message`.  Doing so has pros and cons.

   a. There could be an entity implementation that knows it is sourced directly from a file, which would allow optimizations on several backends. 

   b. There could be an entity implementation that knows it's pure, and optimize some cases in things like retry and follow-redirect middleware.

   c. An entity that tracked its own content-length could be smarter about stream transformations and remove the header -- or maybe even smartly update it.  It's easy to transform the stream and forget to adjust the length.

   d. Letting an Entity manage its own payload headers could clash with the header collection on the message.  Who wins if we set a `Content-Length` in `Message` and also on the Entity?  (`Message#withEntity` already hints at this problem.)

   e. All of those optimizations are possible if we had a richer set of implementations of `Request` and `Response`, but it would double the number of implementations of Message.

4. If we proceed like this, and not the more ambitious approach above, I think `Message#withEntity` needs to be made aware of these.

IndiscriminateCoding:
>     2\. `contentType` and `charset` being static properties of the `EntityEncoder` help with static inference, and (I am pretty sure) is exploited by rho

Maybe we could add `def produces: Set[MediaRange]` to `EntityEncoder` (to be symmetric with `EntityDecoder.consumes`?

rossabaker:
As an addition to `headers`?  That would help with symmetry and not hurt static analysis, but increases the two-sources-of-truth problem and does nothing for the multipart problem.

IndiscriminateCoding:
> Letting an Entity manage its own payload headers could clash with the header collection on the message.  Who wins if we set a `Content-Length` in `Message` and also on the Entity?

It's even worse - someone could call `withEntity` twice and headers will be duplicated.
Maybe we should distinguish `Message` headers from `Entity` headers, storing it in a different (private) fields and only merge them on `Message#headers` call?

IndiscriminateCoding:
Another problem with this design is that it hard to preserve `mediaType` on parameterized `EntityEncoder`.
For example, `EntityEncoder.streamEncoder` can't add proper `mediaType` to `Entity` because it needs to run stream first.
Not sure what to do about that.

IndiscriminateCoding:
Ok, I'm giving up on this. WIll try simpler approach.

PR #2812, opened 2019-08-22
Http4sWSStage test improvements

rossabaker:
Chasing the bug that plagued the http4s-jdk-http-client tests.


rossabaker:
Highly obsolete

PR #2817, opened 2019-08-26
More useful defaults for JettyClient

IndiscriminateCoding:
This PR changes default JettyClient:

1. Enabled SSL support
2. Disabled default Content-Type. It's inconvenient to have **application/octet-stream** by default even if the body is empty. [See also](https://github.com/eclipse/jetty.project/issues/2117).


IndiscriminateCoding:
Ok, I've removed `SslContextFactory`.

PR #2821, opened 2019-08-28
Multipart decoding using EntityDecoder

IndiscriminateCoding:
This PR is an attempt to improve **multipart** decoding by allowing to use `EntityDecoder` on a `Part[F]`:

1) It introduces new interface `Media[F]` which is implemented by `Message[F]` and `Part[F]`;
2) `EntityDecoder.decode` now accepts a `Media[F]` instead of `Message[F]`.

### DIFF
--- a/boopickle/src/main/scala/org/http4s/booPickle/BooPickleInstances.scala
+++ b/boopickle/src/main/scala/org/http4s/booPickle/BooPickleInstances.scala
@@ -18,9 +18,9 @@ import scala.util.{Failure, Success}
   */
 trait BooPickleInstances {
 
-  private def booDecoderByteBuffer[F[_]: Sync, A](msg: Message[F])(
+  private def booDecoderByteBuffer[F[_]: Sync, A](m: Media[F])(
       implicit pickler: Pickler[A]): DecodeResult[F, A] =
-    EntityDecoder.collectBinary(msg).subflatMap { chunk =>
+    EntityDecoder.collectBinary(m).subflatMap { chunk =>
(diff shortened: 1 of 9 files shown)


rossabaker:
I'm so numb to the flaky tests that I didn't notice there's a tut failure here.  We would need to clean that up first.

IndiscriminateCoding:
Fixed

IndiscriminateCoding:
@ChristopherDavenport another approach (without resorting to inheritance) would be to make `Media` a `case class` and store it in a fields of `Request`/`Response`/`Part`. But that would be more invasive change (and less efficient).

rossabaker:
#1572 represents the typeclass approach.  The diff is unfortunately difficult to read after I did something catastrophically stupid while working on docs, but the comments show the mostly pros and a few unresolved cons.  I also think the typeclass approach might help with things like AuthedRequest, which I don't think was appreciated during the discussion of #1572.

I think the alternative to this PR is to have `Media` be a typeclass, like `Message` in #1572.

In either case, it would be nice if we had something that works for `Part`.  The status quo is insufficient.

IndiscriminateCoding:
Not sure if I like the (lawless) typeclass approach, tbh.
Basically you can just replace this typeclass with a bunch of lenses.
But (IMO) http4s API is already pretty complex (which is the price of MTL-like design).

rossabaker:
Would be nice to decide on this before 0.21 final, but are just looking at a milestone today.

hamnis:
I think this is reasonable and we should add this.

PR #2830, opened 2019-08-30
Move Url decoding/encoding to Uri companion object

IndiscriminateCoding:


IndiscriminateCoding:
Binary compatibility checks are failed due to missing `UrlCodingUtils` object.
Is it OK (it was package-private) or should I make a PR against **master**?

rossabaker:
It is okay to add exceptions for package privates.

core/src/main/scala/org/http4s/Uri.scala
```diff
@@ -374,6 +377,129 @@ object Uri {
...
+    *
+    * @param toEncode the string to encode
+    * @param charset the charset to use for characters that are percent encoded
+    * @param spaceIsPlus if space is not skipped, determines whether it will be
+    * rendreed as a `"+"` or a percent-encoding according to `charset`.
+    * @param toSkip a predicate of characters exempt from encoding.  In typical
+    * use, this is composed of all Unreserved URI characters and sometimes a
+    * subset of Reserved URI characters.
+    */
+  def urlEncode(
```
rossabaker: Bikeshed: why (other than private historical reasons) `url` in a class called `Uri`?  Maybe just `encode`?  Similar for `decode`.

IndiscriminateCoding: You are right! Done.

rossabaker:
Looks like this just needs MiMa exceptions.

PR #2832, opened 2019-09-01
Make PR-2785 work as intended

hamnis:
Added a Natural transformation to ReponseGenerator to allow the F and G
to work in unison.

The default implementation is the identity transformation.

@rossabaker suggested that the use of an implicit parameter for
FunctionK is not the way it is supposed to be used, I had to thread that
through the method chain. This means that we no longer can have the
Responses as value classes, as they can only have one member. This will
increase the cost of using the implicit conversions as a new object has
to be allocated.

This change is ~~most likely~~ not binary compatible, so a backport to 0.20
is probably not that useful. It is source compatible though.


hamnis:
Having value classes inside a trait or another class is not allowed by the compiler :(

Issue #2768, opened 2019-08-07
Socket Options Integrated Into Ember


ChristopherDavenport:
FS2 Just Integrated in Support for various socket options into the fs2 Socket code.

https://github.com/functional-streams-for-scala/fs2/pull/1549

So when released, we should continue this support into ember to offer the ability to configure the client and server with these changes.

vilu:
I would like to give this one a go.

vilu:
I've pushed a draft of how I imagine it would be done when the new fs2 version would be released. Would love some feedback if things aren't looking ok.

PR #2836, opened 2019-09-03
Integrate socket options into Ember

vilu:
This one shouldn't compile. I need to bump the fs2-io version but since it hasn't been published yet I can't. I open this one up just so that I can check that this is actually what's needed to be done.

Should fix: https://github.com/http4s/http4s/issues/2768

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ env:
   - LOGBACK_ROOT_LEVEL=WARN
   - LOGBACK_EXCEPTION_PATTERN=%xThrowable{3}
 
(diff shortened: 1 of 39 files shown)


ChristopherDavenport:
Remember we should merge and check this is available once #2867 is in.

vilu:
I'll revert the last commits (with some attempts to do the necessary
upgrades) and prepare the branch for a merge.

Den ons 18 sep. 2019 kl 00:27 skrev Christopher Davenport <
notifications@github.com>:

PR #2848, opened 2019-09-07
Add Error Handling Middleware

ChristopherDavenport:
This currently adds the default error handling middleware which I think is fairly non-controversial. 

Afterwards we will build into this object secondary error handlers, such as `debugJson`, `debugHtml`, etc. Which will make surfacing errors much easier when building out servers.


ChristopherDavenport:
No, this is a matter of lifting the end of the world behavior into the service layer, creating better metrics, logging, and allowing you to test what is returned for a given input.

ChristopherDavenport:
Next PR is to add other versions that are useful for debugging.

PR #2763, opened 2019-08-06
Use scala 2.12.9

aeons:

PR #2850, opened 2019-09-09
Use Scala 2.12.9 on Travis CI

rossabaker:
I think this will clean up [this error](https://travis-ci.org/http4s/http4s/jobs/582492897#L4334).


aeons:
#2763 is also doing this. There's still a scalafix issue, that I will try to get fixed.

aeons:
#2763 is for 0.20 though. This will need the same scalafix changes to work.

rossabaker:
#2763 is better, and we'll merge upward.

PR #2864, opened 2019-09-11
Use new sbt-sonatype bundle features

rossabaker:
1. Upgrades 0.20 to sbt-sonatype-3.x
2. Should stop the empty staging repos being created.
3. Publish a local bundle so Travis CI doesn't time out.


project/CentralRequirements.scala
```diff
@@ -48,7 +48,7 @@ object CentralRequirementsPlugin extends AutoPlugin {
       if (isSnapshot.value)
         Some("snapshots" at nexus + "content/repositories/snapshots")
```
rossabaker: I don't know whether it makes sense to update this as well, since we don't execute the sonatypeRelease step on snapshots.

rossabaker: I guess we can simplify further:

> If isSnapshot.value is true (e.g., if the version name contains -SNAPSHOT), publishSigned task will upload files to the Sonatype Snapshots repository without using the local bundle folder.

PR #2867, opened 2019-09-12
fs2 2.0, jawn-fs2 0.15 and circe 0.12.1

aeons:
Three tests are failing still.


project/Http4sPlugin.scala
```diff
@@ -33,7 +33,7 @@ object Http4sPlugin extends AutoPlugin {
   override def requires = MimaPlugin && ScalafmtPlugin
 
   val scala_213 = "2.13.0"
-  val scala_212 = "2.12.8"
+  val scala_212 = "2.12.10"
```
rossabaker: Metals support for this isn't released yet. Do we have to bump this now?

dropwizard-metrics/src/test/scala/org/http4s/metrics/dropwizard/DropwizardServerMetricsSpec.scala
```diff
@@ -169,7 +169,8 @@ class DropwizardServerMetricsSpec extends Http4sSpec {
       count(registry, Counter("server.default.active-requests")) must beEqualTo(0)
       count(registry, Timer("server.default.requests.total")) must beEqualTo(1)
       valuesOf(registry, Timer("server.default.requests.headers")) must beSome(Array(50000000L))
-      valuesOf(registry, Timer("server.default.get-requests")) must beSome(Array(100000000L))
+      // Not sure why this changed
```
rossabaker: Looking at the implementation, I'm surprised that it wasn't always so. But the second and third calls are close enough together to be immaterial in the real world. I don't think we should hold the release hostage figuring this out.

project/Http4sPlugin.scala
ChristopherDavenport: I thought it was, that PR was merged, I can revert. Was having some classpath issues thought this would be easy.

rossabaker: I'm good to merge this after we revert this, so we don't break tooling (more than sbt-1.3.0 already broke mine! 😆)

.travis.yml
```diff
@@ -19,7 +19,7 @@ env:
   - LOGBACK_ROOT_LEVEL=WARN
   - LOGBACK_EXCEPTION_PATTERN=%xThrowable{3}
 
-scala_version_212: &scala_version_212 2.12.9
+scala_version_212: &scala_version_212 2.12.8
```
rossabaker: We were on 2.12.9 already. Now we've overshot. But I don't care enough to keep blocking this.

PR #2869, opened 2019-09-15
Add JsonDebugErrorHandler

ChristopherDavenport:
Can be improved by #2848  and the 2 hkt message handling/decode approach

### DIFF
--- a/circe/src/main/scala/org/http4s/circe/middleware/JsonDebugErrorHandler.scala
+++ b/circe/src/main/scala/org/http4s/circe/middleware/JsonDebugErrorHandler.scala
@@ -12,68 +12,69 @@ import org.http4s.util.CaseInsensitiveString
 import org.http4s.implicits._
 import org.http4s.circe._
 
-
 object JsonDebugErrorHandler {
 
-  private[this] val messageFailureLogger = org.log4s.getLogger("org.http4s.circe.jsondebugerrorhandler.message-failures")
-  private[this] val serviceErrorLogger = org.log4s.getLogger("org.http4s.circe.jsondebugerrorhandler.service-errors")
+  private[this] val messageFailureLogger =
+    org.log4s.getLogger("org.http4s.circe.jsondebugerrorhandler.message-failures")
+  private[this] val serviceErrorLogger =
+    org.log4s.getLogger("org.http4s.circe.jsondebugerrorhandler.service-errors")
 
   // Can be parametric on my other PR is merged.
   def apply[F[_]](
-    service: Kleisli[F, Request[F], Response[F]],
-    redactWhen: CaseInsensitiveString => Boolean = Headers.SensitiveHeaders.contains
-  )(implicit F: Sync[F]) : Kleisli[F, Request[F], Response[F]] = Kleisli{
-    req =>
+      service: Kleisli[F, Request[F], Response[F]],
+      redactWhen: CaseInsensitiveString => Boolean = Headers.SensitiveHeaders.contains
+  )(implicit F: Sync[F]): Kleisli[F, Request[F], Response[F]] = Kleisli { req =>
     import cats.implicits._
     implicit def entEnc[M[_]: Applicative, N[_]] = JsonErrorHandlerResponse.entEnc[M, N](redactWhen)
 
-    service.run(req)
+    service
+      .run(req)
       .handleErrorWith {
         case mf: MessageFailure =>
           messageFailureLogger.debug(mf)(
             s"""Message failure handling request: ${req.method} ${req.pathInfo} from ${req.remoteAddr
               .getOrElse("<unknown>")}""")
-          mf.toHttpResponse[F](req.httpVersion).map{
-            firstResp =>
+          mf.toHttpResponse[F](req.httpVersion).map { firstResp =>
             Response[F](
               status = firstResp.status,
               httpVersion = firstResp.httpVersion,
-              headers= firstResp.headers.redactSensitive(redactWhen),
+              headers = firstResp.headers.redactSensitive(redactWhen)
             ).withEntity(JsonErrorHandlerResponse[F](req, mf))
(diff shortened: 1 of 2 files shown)


circe/src/main/scala/org/http4s/circe/middleware/JsonDebugErrorHandler.scala
```diff
@@ -0,0 +1,140 @@
...
+    Json.obj(
+      "method" -> req.method.name.asJson,
+      "uri" -> Json.obj(
+        "authority" -> req.uri.authority
+          .map(
+            auth =>
+              Json.obj(
+                "host" -> auth.host.toString().asJson,
+                "port" -> auth.port.asJson,
+                "user_info" -> auth.userInfo
```
rossabaker: Are we configured to drop nulls here?  I guess it doesn't really matter, but this will almost always be null.

circe/src/main/scala/org/http4s/circe/middleware/JsonDebugErrorHandler.scala
```diff
@@ -0,0 +1,140 @@
...
+      redactWhen: CaseInsensitiveString => Boolean): Json =
+    Json.obj(
+      "method" -> req.method.name.asJson,
+      "uri" -> Json.obj(
+        "authority" -> req.uri.authority
+          .map(
+            auth =>
+              Json.obj(
+                "host" -> auth.host.toString().asJson,
+                "port" -> auth.port.asJson,
```
rossabaker: I can imagine wanting an effective port here. Or some indication of the scheme.

circe/src/main/scala/org/http4s/circe/middleware/JsonDebugErrorHandler.scala
```diff
@@ -0,0 +1,140 @@
...
+          serviceErrorLogger.error(t)(
+            s"""Error servicing request: ${req.method} ${req.pathInfo} from ${req.remoteAddr
+              .getOrElse("<unknown>")}"""
+          )
+          Response[F](
+            Status.InternalServerError,
+            req.httpVersion,
+            Headers(
+              Connection("close".ci) ::
+                `Content-Length`.zero ::
```
rossabaker: Why this?

circe/src/main/scala/org/http4s/circe/middleware/JsonDebugErrorHandler.scala
```diff
@@ -0,0 +1,140 @@
...
+import org.http4s._
+import org.http4s.headers.{Connection, `Content-Length`}
+import org.http4s.util.CaseInsensitiveString
+import org.http4s.implicits._
+import org.http4s.circe._
+
+object JsonDebugErrorHandler {
+
+  private[this] val messageFailureLogger =
+    org.log4s.getLogger("org.http4s.circe.jsondebugerrorhandler.message-failures")
```
rossabaker: Nit: It's a bit weird to me that this doesn't match the package whence it comes.

circe/src/main/scala/org/http4s/circe/middleware/JsonDebugErrorHandler.scala
ChristopherDavenport: My bad, leftover.

circe/src/main/scala/org/http4s/circe/middleware/JsonDebugErrorHandler.scala
ChristopherDavenport: I can set that.

circe/src/main/scala/org/http4s/circe/middleware/JsonDebugErrorHandler.scala
```diff
@@ -0,0 +1,147 @@
...
+            s"""Message failure handling request: ${req.method} ${req.pathInfo} from ${req.remoteAddr
+              .getOrElse("<unknown>")}""")
+          mf.inHttpResponse[F, G](req.httpVersion).map { firstResp =>
+            Response[G](
+              status = firstResp.status,
+              httpVersion = firstResp.httpVersion,
+              headers = firstResp.headers.redactSensitive(redactWhen)
+            ).withEntity(JsonErrorHandlerResponse[G](req, mf))
+          }
+        case scala.util.control.NonFatal(t) =>
```
ChristopherDavenport: I don't think we need this anymore now we are off task. This had escaped my vision until this exact instance, but I bet this is in other spots as we..

```suggestion
        case t =>
```

circe/src/main/scala/org/http4s/circe/middleware/JsonDebugErrorHandler.scala
```diff
@@ -34,27 +33,26 @@ object JsonDebugErrorHandler {
...
+          mf.inHttpResponse[F, G](req.httpVersion).map { firstResp =>
+            Response[G](
               status = firstResp.status,
               httpVersion = firstResp.httpVersion,
               headers = firstResp.headers.redactSensitive(redactWhen)
-            ).withEntity(JsonErrorHandlerResponse[F](req, mf))
+            ).withEntity(JsonErrorHandlerResponse[G](req, mf))
           }
-        case scala.util.control.NonFatal(t) =>
+        case t =>
```
rossabaker: Why did we lose the `NonFatal` here?

ChristopherDavenport: Fatal errors should never be bubbled up into the cats handleErrorWith in the first place.

rossabaker: Okay, that's true. Not really awake yet.  This is just rendering what's already been raised, not catching things.  And unless someone already messed up, it won't be raised.

PR #2874, opened 2019-09-19
Release v0.21.0-M5

rossabaker:


rossabaker:
Oh, wow, so that's why it hadn't released yet.  I was refreshing Travis CI like a fool last night.

[ingests more caffeine]

PR #2884, opened 2019-09-24
Use Scala 2.13.1

aeons:


aeons:
Error is tracked here https://github.com/typelevel/kind-projector/issues/116

rossabaker:
This is now unblocked upstream if we upgrade kind-projector.

aeons:
Should be ready for merge now.

PR #2887, opened 2019-09-26
Support realm in the OAuth1 header

semenodm:
Support realm as part of Auth header


client/src/main/scala/org/http4s/client/oauth1/Header.scala
```diff
@@ -0,0 +1,77 @@
+package org.http4s.client.oauth1
+
+import cats.Show
+import cats.kernel.Order
+import cats.implicits._
+sealed trait Header {
+  val headerName: String
+  val headerValue: String
+  def toTuple: (String, String) =
+    headerName → encode(headerValue)
```
rossabaker: Unicode arrows are deprecated in Scala 2.13.  Better to use `=>`.

client/src/main/scala/org/http4s/client/oauth1/Header.scala
```diff
@@ -0,0 +1,77 @@
+package org.http4s.client.oauth1
+
+import cats.Show
+import cats.kernel.Order
+import cats.implicits._
+sealed trait Header {
```
rossabaker: Is there any reason not to model these as standard `org.http4s.Header`?  We can still have the specific types, but these won't work quite the same as other headers in http4s.

client/src/main/scala/org/http4s/client/oauth1/OAuthConfig.scala
```diff
@@ -0,0 +1,15 @@
+package org.http4s.client.oauth1
+
+import org.http4s.client.oauth1.Header.{Callback, Verifier}
+import org.http4s.client.oauth1.{Header => OAuthHeader}
+
+case class OAuthConfig(
```
rossabaker: Are we sure more config won't be necessary?  We've started to move away from case classes for configuration, so we can add more parameters later without breaking binary compatibility.  This is not an issue if this is truly a closed set, but it makes me a bit nervous.

client/src/main/scala/org/http4s/client/oauth1/OAuthConfig.scala
```diff
@@ -0,0 +1,15 @@
...
+    consumer: OAuthHeader.Consumer,
+    token: Option[OAuthHeader.Token],
+    realm: Option[OAuthHeader.Realm],
+    signatureMethod: OAuthHeader.SignatureMethod = Header.SignatureMethod(),
+    timestampGenerator: () ⇒ OAuthHeader.Timestamp,
```
rossabaker: I'd consider `F[OAuthHeader.Timestamp]` instead.  These are impure functions.

client/src/main/scala/org/http4s/client/oauth1/Header.scala
rossabaker: I see it now in `takeSigHeaders`.  Yeah, maybe make this a marker trait, but still make these standard http4s headers?

semenodm: I'm not sure these are standard headers, these are part of. So i'm not sure if we want them to be Http header.

```Authorization: OAuth realm="Example",
  oauth_consumer_key="0685bd9184jfhq22",
  oauth_token="ad180jjd733klru7",
  oauth_signature_method="HMAC-SHA1",
  oauth_signature="wOJIO9A2W5mFwDgiDvZbTSMK%2FPY%3D",
  oauth_timestamp="137131200",
  oauth_nonce="4572616e48616d6d65724c61686176",
  oauth_version="1.0"
```

client/src/main/scala/org/http4s/client/oauth1/OAuthConfig.scala
semenodm: my rule of thumb is usually if number of params > 3 i use case classes

semenodm: i never thought about your point

client/src/main/scala/org/http4s/client/oauth1/Header.scala
semenodm: Done

client/src/main/scala/org/http4s/client/oauth1/Header.scala
rossabaker: You're right.  I got fooled by the name.  These are called "parameters" in most headers, and the spec refers to them as "protocol parameters".  Maybe I'd call the trait `ProtocolParameter` to avoid ambiguity with `Header`.

client/src/main/scala/org/http4s/client/oauth1/OAuthConfig.scala
rossabaker: It's tedious to write out binary compatible case classes by hand, but it does improve the compatibility story.  If we think this will always be enough config, we can probably leave this alone.

An interesting project has emerged to help with this, but should probably be discussed separately: [data-class](https://github.com/alexarchambault/data-class).

client/src/main/scala/org/http4s/client/oauth1/Header.scala
```diff
@@ -0,0 +1,77 @@
...
+
+  case class Callback(override val headerValue: String) extends Header {
+    override val headerName: String = "oauth_callback"
+  }
+
+  case class Verifier(override val headerValue: String) extends Header {
+    override val headerName: String = "oauth_verifier"
+  }
+
+  implicit val sortHeaders: Order[Header] = new Order[Header] {
```
rossabaker: Nit we've started following the cats convention of naming with the package, type name, and typeclass name, to avoid any accidental shadowing.  So, `http4sClientOauth1SortForHeaders`.

client/src/main/scala/org/http4s/client/oauth1/Header.scala
```diff
@@ -0,0 +1,77 @@
...
+    override def compare(x: Header, y: Header): Int = x.headerName.compareTo(y.headerName)
+  }
+  val tupleShow: Show[(String, String)] = new Show[(String, String)] {
```
rossabaker: I'm a little surprised cats doesn't have an inductive instance for this one.  You could make it private so we can refactor it without a breaking change.  This shouldn't be http4s' responsibility to expose to another project.

semenodm:
@rossabaker please take a look when you have time

client/src/main/scala/org/http4s/client/oauth1/ProtocolParameter.scala
```diff
@@ -0,0 +1,81 @@
...
+
+  case class Custom(headerName: String, headerValue: String) extends ProtocolParameter
+
+  case class SignatureMethod(override val headerValue: String = "HMAC-SHA1")
+      extends ProtocolParameter {
+    override val headerName: String = "oauth_signature_method"
+  }
+
+  case class Timestamp(
+      override val headerValue: String = (System.currentTimeMillis() / 1000).toString)
```
rossabaker: Should pass in a clock and create in an effect.

client/src/main/scala/org/http4s/client/oauth1/ProtocolParameter.scala
```diff
@@ -0,0 +1,81 @@
...
+      extends ProtocolParameter {
+    override val headerName: String = "oauth_timestamp"
+  }
+
+  case class Nonce(override val headerValue: String = System.nanoTime.toString)
```
rossabaker: Should pass in a clock and create in an effect.

rossabaker:
Cherry-picked to #2992.  Proposed fixes in 579f7c752f.

PR #2891, opened 2019-09-27
Compile without fatal warnings on 2.13.0

rossabaker:
* Reimplement the case classiness of `Request` by hand to work around scala/bug#11457.
* Work around `Seq` aliasing in a spec
* Eliminate early initializer in TomcatSpec, which no longer appears to be necessary

### DIFF
--- a/core/src/main/scala/org/http4s/Message.scala
+++ b/core/src/main/scala/org/http4s/Message.scala
@@ -6,12 +6,13 @@ import cats.implicits._
 import cats.effect.IO
 import fs2.{Pure, Stream}
 import fs2.text.{utf8Decode, utf8Encode}
-import java.io.File
+import java.io.{File, Serializable}
 import java.net.{InetAddress, InetSocketAddress}
 import org.http4s.headers._
 import org.http4s.util.decode
 import org.log4s.getLogger
 import _root_.io.chrisdavenport.vault._
+import scala.util.hashing.MurmurHash3
 
 /**
   * Represents a HTTP Message. The interesting subclasses are Request and Response.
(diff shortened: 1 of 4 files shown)


core/src/main/scala/org/http4s/Message.scala
```diff
@@ -236,14 +237,14 @@ object Message {
...
-    attributes: Vault = Vault.empty
-) extends Message[F] {
+final class Request[F[_]](
+    val method: Method = Method.GET,
+    val uri: Uri = Uri(path = "/"),
+    val httpVersion: HttpVersion = HttpVersion.`HTTP/1.1`,
+    val headers: Headers = Headers.empty,
+    val body: EntityBody[F] = EmptyBody,
+    val attributes: Vault = Vault.empty
+) extends Message[F] with Product with Serializable {
```
rossabaker: I'm not a huge fan of these interfaces, but:

* `Product` makes hash code easier and was there before.
* Spark folks may be depending on `Serializable`. I don't judge.

core/src/main/scala/org/http4s/Message.scala
```diff
@@ -414,6 +415,34 @@ sealed abstract case class Request[F[_]](
...
+      (this.method == that.method) &&
+      (this.uri == that.uri) &&
+      (this.httpVersion == that.httpVersion) &&
+      (this.headers == that.headers) &&
+      (this.body == that.body) &&
+      (this.attributes == that.attributes)
+    case _ => false
+  })
+
+  def canEqual(that: Any) = that.isInstanceOf[Request[F]]
```
rossabaker: That `F` is erased and I'm surprised this didn't generate a warning but `_` didn't work.  Maybe `_[_]`?  Ugh.

aeons:
The unapply on Request seems to fail < 2.13

hamnis:
Maybe this would help: https://github.com/alexarchambault/data-class

rossabaker:
I just mentioned data-class on another ticket. :smile:  It's can also be a compile-time only dependency, which makes it more attractive for our use case.

hamnis:
I tried doing this now, and it failed with conflict on `withXXX`

hamnis:
Seems like the only change needed for this to work is changing the pattern match for `org.http4s.client.middleware.Logger` and `org.http4s.server.middleware.Logger`

```    
def prelude = message match {
      case r: Request[F] =>
        s"${r.httpVersion} ${r.method} ${r.uri}"

      case Response(status, httpVersion, _, _, _) =>
        s"$httpVersion $status"
}
```

hamnis:
@rossabaker Should I create a new PR to get this moving again?

rossabaker:
@hamnis That would be wonderful. Thank you!

rossabaker:
Superseded by #2933.

PR #2896, opened 2019-10-03
Adding the ability to filter out status codes to not include in metrics.

BusyByte:
Adding the ability to filter out status codes to not include in metrics.
For example 503's are normal because of maintenance on a third party and we don't
want them combined in with 500s which indicate an actual issue which needs fixed.
Introducing a new function and altering private functions and case classes should be backwards compatible.
The exiting apply function delegates to the new filter function as to not duplicate any code.

My immediate need is on 0.18.x but I intend on making the equivalent changes on 0.20.x and master branches as well as we'll need those as well pretty soon.

### DIFF
--- a/server-metrics/src/main/scala/org/http4s/server/metrics/Metrics.scala
+++ b/server-metrics/src/main/scala/org/http4s/server/metrics/Metrics.scala
@@ -12,7 +12,10 @@ import java.util.concurrent.TimeUnit
 object Metrics {
 
   def apply[F[_]](m: MetricRegistry, prefix: String = "org.http4s.server")(
-      implicit F: Effect[F]): HttpMiddleware[F] = { service =>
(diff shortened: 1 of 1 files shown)


BusyByte:
Binary compatibility failed but not sure why.  People shouldn't be depending on stuff marked `private`.  Let me know the recommended actions to take here.

ChristopherDavenport:
I believe the problem is including it in the case class, we can add it to the case class in the 0.20 branch, but here you'll need to pass it to the parameters rather than in the shared data structure.

BusyByte:
@ChristopherDavenport that case class is private to Metrics object but I'll trust you on this, changed in af3a79e

BusyByte:
@ChristopherDavenport I think I'm making changes to the wrong class here, I meant to make changes to the prometheus version

BusyByte:
@ChristopherDavenport I opened https://github.com/http4s/http4s/pull/2897 which replaces this one

PR #2897, opened 2019-10-03
Adding the ability to filter out status codes to not include in metrics.

BusyByte:
Adding the ability to filter out status codes to not include in metrics.
For example 503's are normal because of maintenance on a third party and we don't
want them combined in with 500s which indicate an actual issue which needs fixed.
Introducing a new function and altering private functions should be backwards compatible.
The exiting apply function delegates to the new filter function as to not duplicate any code.

My immediate need is on 0.18.x but I intend on making the equivalent changes on 0.20.x and master branches as well as we'll need those as well pretty soon.

### DIFF
--- a/prometheus-server-metrics/src/main/scala/org/http4s/server/prometheus/PrometheusMetrics.scala
+++ b/prometheus-server-metrics/src/main/scala/org/http4s/server/prometheus/PrometheusMetrics.scala
@@ -62,7 +62,8 @@ object PrometheusMetrics {
       serviceMetrics: ServiceMetrics,
       service: HttpService[F],
       emptyResponseHandler: Option[Status],
-      errorResponseHandler: Throwable => Option[Status]
+      errorResponseHandler: Throwable => Option[Status],
+      statusesToIgnore: Set[Status]
   )(
       req: Request[F]
   ): OptionT[F, Response[F]] = OptionT {
@@ -78,13 +79,21 @@ object PrometheusMetrics {
             initialTime,
             headersElapsed,
             serviceMetrics,
-            errorResponseHandler(e)) *>
+            errorResponseHandler(e),
+            statusesToIgnore) *>
             Sync[F].raiseError[Option[Response[F]]](e),
         _.fold(
-          onEmpty[F](req.method, initialTime, headersElapsed, serviceMetrics, emptyResponseHandler)
+          onEmpty[F](
+            req.method,
+            initialTime,
+            headersElapsed,
+            serviceMetrics,
+            emptyResponseHandler,
+            statusesToIgnore)
             .as(Option.empty[Response[F]])
         )(
-          onResponse(req.method, initialTime, headersElapsed, serviceMetrics)(_).some.pure[F]
+          onResponse(req.method, initialTime, headersElapsed, serviceMetrics, statusesToIgnore)(_).some
+            .pure[F]
         )
       )
     } yield result
@@ -95,7 +104,8 @@ object PrometheusMetrics {
       start: Long,
       headerTime: Long,
       serviceMetrics: ServiceMetrics,
-      emptyResponseHandler: Option[Status]
+      emptyResponseHandler: Option[Status],
+      statusesToIgnore: Set[Status]
   ): F[Unit] =
     for {
       now <- Sync[F].delay(System.nanoTime)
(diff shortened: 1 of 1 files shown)


prometheus-server-metrics/src/main/scala/org/http4s/server/prometheus/PrometheusMetrics.scala
```diff
@@ -62,7 +62,8 @@ object PrometheusMetrics {
       serviceMetrics: ServiceMetrics,
       service: HttpService[F],
       emptyResponseHandler: Option[Status],
-      errorResponseHandler: Throwable => Option[Status]
+      errorResponseHandler: Throwable => Option[Status],
+      statusesToIgnore: Set[Status]
```
rossabaker: Why just `Status`?  Might we want to ignore on headers? 

```scala
responseRule: Response[F] => Rule
```

where

```scala
sealed trait Rule
case object Record extends Rule
case object Reject extends Rule
```

And if we're filtering responses, I can also imagine filtering requests.

kevinmeredith:
I was just looking at `Prometheus.scala` regarding a similarly intentioned change, @BusyByte. 

What are your thoughts about updating [recordTotalTime](https://github.com/http4s/http4s/blob/v0.20.11/prometheus-metrics/src/main/scala/org/http4s/metrics/prometheus/Prometheus.scala#L94-L104) to include a `reportStatus: Status => String` method that's passed into `recordTotalTime`, whose default would be the existing [reportStatus](https://github.com/http4s/http4s/blob/v0.20.11/prometheus-metrics/src/main/scala/org/http4s/metrics/prometheus/Prometheus.scala#L137-L144) function?

```scala
      override def recordTotalTime(
          method: Method,
          status: Status,
          elapsed: Long,
          classifier: Option[String]): F[Unit]
```

With that approach, we could handle your HTTP-503 case with:

```scala
      private def reportStatus(status: Status): String =
        status.code match {
          case hundreds if hundreds < 200 => "1xx"
          case twohundreds if twohundreds < 300 => "2xx"
          case threehundreds if threehundreds < 400 => "3xx"
          case fourhundreds if fourhundreds < 500 => "4xx"
          case serviceUnavailable if serviceUnavailable === 503 => "503"
          case _ => "5xx"
        }
```

prometheus-server-metrics/src/main/scala/org/http4s/server/prometheus/PrometheusMetrics.scala
BusyByte: @rossabaker I like this idea because we have in the past done some shuffling of which routes gets metrics so that we don't get our `ping` route in with business route metrics.  This would make it so I could just exclude `ping`.

BusyByte:
@kevinmeredith Somebody correct me if I'm wrong but there is a few problems with this:

1. The reason that they have the buckets of 1xx, 2xx, 3xx, 4xx, 5xx is if they would have too much cardinality and prometheus performance would suffer.  Otherwise they would have just not used the `Nxx` buckets in the first place.  With my solution the cardinality does not increase.
1. Adding for a specific code meets our needs today but may not meet others and what if tomorrow we want to exclude 501 and 503?  I don't think just adding one code makes this a general solution.
1.  Having a exclude503From5xxReportStatus makes you have to duplicate all the code that ever calls the reportStatus

prometheus-server-metrics/src/main/scala/org/http4s/server/prometheus/PrometheusMetrics.scala
BusyByte: @rossabaker I'm finding this is getting complicated because there are spots where we create a status but we don't formally have a response, namely because of:
```
emptyResponseHandler: Option[Status],
errorResponseHandler: Throwable => Option[Status]
```

BusyByte: @rossabaker I'm giving it a try though, will be done shortly.

kevinmeredith:
Roger, @BusyByte. I didn't intend to hold up your progress.

prometheus-server-metrics/src/main/scala/org/http4s/server/prometheus/PrometheusMetrics.scala
BusyByte: @rossabaker changed in 6b516b6 , let me know if anything needs cleaned up

BusyByte:
closing this as it seems to not be getting uptake.  I'm wondering if the version that @ChristopherDavenport created with [epimetheus-http4s](https://github.com/ChristopherDavenport/epimetheus-http4s/pull/79) isn't a better way to go anyway.  If you want me to re-open it just let me know.

PR #2903, opened 2019-10-08
Fix Compilation Errors With New `sbt-pgp` Plugin Version

isomarcte:
WARNING: This does more than just fix the compilation errors, it also has at least one non-trivial consequence. As of [SBT 2.0.0][sbt-bc-deprecation] Bouncy Castle based PGP is deprecated. Thus this commit removes `useGpg := false`. I believe this means that `gpg` _must_ be on the `PATH` in order for `sbt-pgp` to provide PGP based operations.

[sbt-bc-deprecation]: https://github.com/sbt/sbt-pgp/blob/90465fcde014eb0688f0086fdec56db60ebfbfd6/sbt-pgp/src/main/scala/com/jsuereth/sbtpgp/SbtPgp.scala#L21 "SbtPgp.scala"

This should replace: https://github.com/http4s/http4s/pull/2870

### DIFF
--- a/project/Http4sPlugin.scala
+++ b/project/Http4sPlugin.scala
@@ -2,9 +2,9 @@ package org.http4s.build
 
 import com.timushev.sbt.updates.UpdatesPlugin.autoImport._ // autoImport vs. UpdateKeys necessary here for implicit
 import com.typesafe.sbt.SbtGit.git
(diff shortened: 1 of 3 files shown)


isomarcte:
I do not believe the current build failure is related to these changes.

PR #2905, opened 2019-10-08
Fix Bug Where The Body Of The Request Is Not Sent In Async Client

isomarcte:
The old code used the `EmptyBodyGenerator` in the event that the a request had `_.isChunked == false` and `_.contentLength.isEmpty`. This caused valid requests created in the following manner to not have their body's sent.

```scala
Request(method = POST, Stream.eval(IO.pure(1.toByte)))
```

This seems like a strange way to construct a `Request` value, and indeed it is, but since it is permissible by http4s it should be valid.

A further reason for this change is to create parity between the different client implementations. The test I added _passes_ for the other clients right now, it is only the async client which fails (before this change that is).

It should be noted that I am seeing this issue in the wild with a less contrived example. I have an http4s server using the async client to proxy requests. This uses the incoming `Request` to create the new out going `Request`. I believe that the requests themselves are invalid, that is they are missing a content-type header or a transfer encoding header, which should not be permitted by the RFC.

We could go a different route here, and categorically enforce that any `Request` value have either a `Content-Length` or a `Transfer-Encoding` header, but I was hesitant to do so because that significantly changes the design of `Message.scala` which derives many of its values from the `Headers`.

Note, this effectively reverts https://github.com/http4s/http4s/commit/13dc0e6bff0191c776ede21acd68d7434cc811ed

I'm a bit concerned that it will resurrect the issue described here, https://github.com/http4s/http4s/blob/9f7c6f5a5d3570c38aa4bc9b3b0fb518937d86ea/website/src/hugo/content/changelog.md#L1673.

Rather than the changes proposed in this PR, I'm totally open to some way of preventing the construction of invalid `Request` values. I think we'd be able to enforce being explicit about the content-length on creation and perhaps having the server reject requests which are missing these headers, but that seems like something which requires a much wider discussion.

### DIFF
--- a/async-http-client/src/main/scala/org/http4s/client/asynchttpclient/AsyncHttpClient.scala
+++ b/async-http-client/src/main/scala/org/http4s/client/asynchttpclient/AsyncHttpClient.scala
@@ -136,10 +136,33 @@ object AsyncHttpClient {
       req.body.chunks.map(chunk => Unpooled.wrappedBuffer(chunk.toArray)))
     if (req.isChunked) new ReactiveStreamsBodyGenerator(publisher, -1)
     else
-      req.contentLength match {
-        case Some(len) => new ReactiveStreamsBodyGenerator(publisher, len)
-        case None => EmptyBodyGenerator
-      }
+      // As far as I can tell there are three valid options when
+      // `req.contentLength.isEmpty`, none are particularly pleasant.
+      //
+      // 1. We can pass in -1 to the ReactiveStreamsBodyGenerator, this will
+      //    cause a 'Transfer-Encoding: chunked' header to be added to the
+      //    request headers. This is disadvantageous as this header will not
+      //    be reflected in the `req.headers` value, and thus may be
+      //    surprising to some users.
+      //
+      // 2. We can strictly resolve the body and compute the content length
+      //    here. The assumption here is that the request was intended to be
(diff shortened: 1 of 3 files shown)


isomarcte:
Looks like this is the case for the jdk based client as well...

isomarcte:
I had thought this was isolated to the async client, but it appears in `JavaNetClient` too. I've pushed a "fix" for that client as well, but I'm further second guessing if this is the right thing to do.

On the one hand dropping bodies made via the normal `Request` constructor, as opposed to the dsl, if they do not have an explicit content-length or transfer encoding header seems confusing, especially since the blaze client _doesn't_ do this. On the other hand such requests are invalid.

This makes me think more that we need to force an explicit or default behavior.

isomarcte:
My fix to the `JavaNetClient` did not quite fix it, but I think I'm going to open a different PR to tackle this from a different direction.

isomarcte:
Closing for now. There appears to be no way which does not substantially break or change the API to fix this issue in a general sense. Many of the HTTP clients generate errors if a content-length or transfer encoding chunked is set, even if the value is 0, under certain conditions. For example, okhttp's client throws an exception if a `GET` request has a transfer-encoding header, even though the RFC permits that on requests (though it should be semantically ignored) and even if the value is 0.

PR #2906, opened 2019-10-10
Weaken Http/HttpRoutes Constraint

ChristopherDavenport:
`Sync` is no longer required, we only need `Delay` for the ability to suspend the route analysis.

### DIFF
--- a/core/src/main/scala/org/http4s/Http.scala
+++ b/core/src/main/scala/org/http4s/Http.scala
@@ -1,6 +1,6 @@
 package org.http4s
 
-import cats.Applicative
+import cats._
 import cats.data.Kleisli
 import cats.effect.Sync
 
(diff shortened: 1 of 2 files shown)


rossabaker:
We had a previous discussion of this at https://github.com/http4s/http4s/pull/2325 and decided not to do it.   I'm fine with this, but curious: did something change in the library, or just in our attitude?

aeons:
I completely forgot that discussion, and thought that weaker constraints => better.
I guess I still agree with the points made there, so revert this one?

hamnis:
@ChristopherDavenport Any particular reason `AuthedRoutes` did not get the same treatment?
I can fix this as part of #2612 if this is interesting.

ChristopherDavenport:
Nope, please feel free to integrate there as well.

rossabaker:
Okay, I'll buy that, and :+1: to keeping `AuthedRoutes` consistent.

PR #2907, opened 2019-10-10
Add JsonDecoder Algebra

ChristopherDavenport:
Single Type Abstraction for separation of Sync and the locations using the `Json`, which is primarily businesss logic where we'd prefer to remove effects.

### DIFF
--- a/circe/src/main/scala/org/http4s/circe/CirceInstances.scala
+++ b/circe/src/main/scala/org/http4s/circe/CirceInstances.scala
@@ -115,6 +115,12 @@ trait CirceInstances extends JawnInstances {
       Uri.fromString(str).leftMap(_ => "Uri")
     }
 
+  implicit class JsonDecoderSyntax[F[_]: JsonDecoder](private val req: Message[F])(
+      implicit F: JsonDecoder[F]) {
+    def asJson: F[Json] = F.asJson(req)
+    def asJsonDecode[A: Decoder]: F[A] = F.asJsonDecode(req)
+  }
+
   implicit class MessageSyntax[F[_]: Sync](self: Message[F]) {
     def decodeJson[A](implicit decoder: Decoder[A]): F[A] =
       self.as(implicitly, jsonOf[F, A])
--- a/circe/src/main/scala/org/http4s/circe/CirceInstances.scala
+++ b/circe/src/main/scala/org/http4s/circe/CirceInstances.scala
@@ -115,8 +115,7 @@ trait CirceInstances extends JawnInstances {
       Uri.fromString(str).leftMap(_ => "Uri")
     }
 
-  implicit class JsonDecoderSyntax[F[_]](private val req: Message[F])(
-      implicit F: JsonDecoder[F]) {
+  implicit class JsonDecoderSyntax[F[_]](private val req: Message[F])(implicit F: JsonDecoder[F]) {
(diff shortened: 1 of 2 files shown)


circe/src/main/scala/org/http4s/circe/CirceInstances.scala
```diff
@@ -115,6 +115,12 @@ trait CirceInstances extends JawnInstances {
       Uri.fromString(str).leftMap(_ => "Uri")
     }
 
+  implicit class JsonDecoderSyntax[F[_]: JsonDecoder](private val req: Message[F])(
+      implicit F: JsonDecoder[F]) {
```
IndiscriminateCoding: Isn't it already specified as `F` context bound?
(also - don't you want to move this constraint to methods to make `implicit class` extends AnyVal to avoid allocation every time method is called?)

ChristopherDavenport: Whoops. Yes. I switched how I was invoking it and left the old one. will fix these two.

ChristopherDavenport: Can't extend AnyVal because of our position - 

`value class may not be a member of another class` - We could move these syntax's out but it would be a lot more involved move that I think I would rather do separately from this concept.

circe/src/main/scala/org/http4s/circe/CirceInstances.scala
```diff
@@ -115,6 +115,11 @@ trait CirceInstances extends JawnInstances {
       Uri.fromString(str).leftMap(_ => "Uri")
     }
 
+  implicit class JsonDecoderSyntax[F[_]](private val req: Message[F])(implicit F: JsonDecoder[F]) {
```
kevinmeredith: Is it worthwhile to add `extends AnyVal` to, as I understand, minimize allocations?

kevinmeredith: Nevermind: https://github.com/http4s/http4s/pull/2907#discussion_r333745452.

circe/src/main/scala/org/http4s/circe/JsonDecoder.scala
```diff
@@ -0,0 +1,25 @@
...
+ **/
+trait JsonDecoder[F[_]] {
+  def asJson(m: Message[F]): F[Json]
+  def asJsonDecode[A: Decoder](m: Message[F]): F[A]
+}
+
+object JsonDecoder {
+  def apply[F[_]](implicit ev: JsonDecoder[F]): JsonDecoder[F] = ev
+
+  implicit def impl[F[_]: Sync]: JsonDecoder[F] = new JsonDecoder[F] {
```
rossabaker: Thinking about what other implementations could plausibly exist.  I guess for the other `Stream.Compiler`s, though you'd want a `MonadError` for `Id` and `Pure`.  It's not a typeclass with laws, so I guess a constant decoder "works".

I get the aversion to `Sync`, but I need some help thinking about why it's good to pretend that this can't fail.

kubukoz:
I think this isn't being used in the client, do you folks think we could get that in 1.0.0?

rossabaker:
I've never used it, so I can't say how well it has aged.  But symmetry seems good.  If it has worked out well for people, I'd be :+1:.

PR #2909, opened 2019-10-11
JettyClient: properly propagate streamed errors

### LINKED ISSUE #2901: Client request never return if server respond with 401 without www-authenticate header
Given the following code:
```scala
object T extends cats.effect.IOApp {
  import org.http4s.circe._
  def run(args: List[String]): IO[ExitCode] = for {
    (c, _) <- JettyClient.allocate[IO]()
    r = Request[IO](
      method = Method.GET,
      uri = http4s.Uri(
        path = "/some/url/that/returns/401",
        scheme = Some(Scheme.http),
        authority = Some(Authority(host = RegName("hostname")))
    )
    )
    _ <- c.fetch[Unit](r) { r =>
      r.as[Json].map(println)
    }
  } yield ExitCode.Success
}
```
and a server that return **401** without **www-authenticate** header, client never returns.
In case of cats-effect IO, it prints an error (while continuing to hang - example above never terminates):
```
org.eclipse.jetty.client.HttpResponseException: HTTP protocol violation: Authentication challenge without WWW-Authenticate header
```
And when using ZIO, it doesn't prints anything and just hangs infinitely.
Reproduced using http4s 0.21.0-M4 and Jetty client.

IndiscriminateCoding:
Resolves #2901 

**The problem**: any error that occurs in a Jetty client *after* headers were read isn't propagated to the body stream.
This results in weird behavior where client just hangs and its `body` never terminates.
This PR changes implementation of a `ResponseListener` to distinguish current response state and properly propagate error whether to callback or to body stream.

### DIFF
--- a/jetty-client/src/main/scala/org/http4s/client/jetty/ResponseListener.scala
+++ b/jetty-client/src/main/scala/org/http4s/client/jetty/ResponseListener.scala
@@ -12,26 +12,38 @@ import java.nio.ByteBuffer
 import org.eclipse.jetty.client.api.{Result, Response => JettyResponse}
 import org.eclipse.jetty.http.{HttpFields, HttpVersion => JHttpVersion}
 import org.eclipse.jetty.util.{Callback => JettyCallback}
+import org.http4s.client.jetty.ResponseListener.Item
 import org.http4s.internal.{invokeCallback, loggingAsyncCallback}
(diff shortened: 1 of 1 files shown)


rossabaker:
This is cherry-picked into #2990 for release in 0.20.14.

PR #2916, opened 2019-10-16
Ensure That Metrics Only Decrements Active Requests Once

isomarcte:
Prior to this commit there was an edge case where the `ops.decreaseActiveRequests` would get run more than once if an error or cancellation occurred _after_ the `onFinalize` method was run in `onResponse`.

This change adds a `Ref` which is initialized at the start of request handling and set to `true` once the `ops.decreaseActiveRequests` is run. All attempts to set `ops.decreaseActiveRequests` are guarded by this boolean value.

### DIFF
--- a/server/src/main/scala/org/http4s/server/middleware/Metrics.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/Metrics.scala
@@ -1,8 +1,9 @@
 package org.http4s.server.middleware
 
 import cats.data.{Kleisli, OptionT}
-import cats.effect.{Clock, ExitCase, Sync}
+import cats.effect.concurrent.Ref
 import cats.effect.implicits._
+import cats.effect.{Clock, ExitCase, Sync}
 import cats.implicits._
 import fs2.Stream
 import java.util.concurrent.TimeUnit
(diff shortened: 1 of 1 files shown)


isomarcte:
This may not be a bug. I'm actually not 100% sure anymore. Leaving open until I can think on it some more.

isomarcte:
After some additional testing, I do still believe there is an edge case here and I do still think the changes I've proposed will fix it.

Here is what I know,

* I currently have an http4s 0.20.11 server running with prometheus based metrics provided by this library.
* Currently, the `org_http4s_server_active_request` value is at `-1.0`. 

Here is what I suspect.
If cancellation or error occurs in the `onEmpty` handler, it is possible that the code to decrease the active request count will run twice, once in the `onEmpty` handler and once in either of the non-success case branches of the `bracketCase` release handler. This is because there is no synchronization on decrementing that value.

So, my brain is exhausted. I'll update the commit/PR notes tomorrow with some better, and hopefully actually correct, examples if not tests cases.

rossabaker:
Oops.  This is a bugfix.  I should have backported this one.

PR #2922, opened 2019-10-23
Bugfix in the baze-core/util/IdentityWriter causing upload failure

mi-char:
Bug causes failure of a chunked upload in case when Content-Length (Long) converted to Int32 overflows to a positive number.

Test case: chunked upload of file with a content length 4,294,967,300 bytes.
Result: `IllegalArgumentException("Will not write more bytes than what was indicated by the Content-Length header 4")`

### DIFF
--- a/blaze-core/src/main/scala/org/http4s/blazecore/util/IdentityWriter.scala
+++ b/blaze-core/src/main/scala/org/http4s/blazecore/util/IdentityWriter.scala
@@ -21,9 +21,9 @@ private[http4s] class IdentityWriter[F[_]](size: Long, out: TailStage[ByteBuffer
 
   private var bodyBytesWritten = 0L
 
-  private def willOverflow(count: Long) =
-    if (size < 0) false
-    else count + bodyBytesWritten > size
+  private def willOverflow(count: Int) =
+    if (size < 0L) false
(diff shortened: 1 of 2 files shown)


blaze-core/src/main/scala/org/http4s/blazecore/util/IdentityWriter.scala
```diff
@@ -38,7 +38,7 @@ private[http4s] class IdentityWriter[F[_]](size: Int, out: TailStage[ByteBuffer]
 
       logger.warn(msg)
 
-      val reducedChunk = chunk.take(size - bodyBytesWritten)
+      val reducedChunk = chunk.take((size - bodyBytesWritten).toInt)
```
mi-char: Conversion to Int on this place is safe as long as chunk.size is also Int.

mi-char:
Thank you.
Should I change this PR to the series/0.20 branch, or leave this one to the master and create new PR to the 0.20?

rossabaker:
I can cherry pick it back tonight and get a release.going.

mi-char:
That would be perfect. So I will leave on the master branch.
Thank you in advance for the release.
And big thanks for the https4s.

rossabaker:
This is cherry-picked and released as 0.20.12.

PR #2932, opened 2019-10-28
Adding ipython notebook extension

dmvieira:


aeons:
Hi, the `MimeDB.scala` file is automatically generated from data in https://github.com/jshttp/mime-db

dmvieira:
@aeons , but it should not be maintained as code? What's the process to make this file?

rossabaker:
There are [instructions here](https://github.com/http4s/http4s/tree/063287b0bc18e402a362dfe8c4826c803314c23c/mimedb-generator). We periodically run that and submit the result as a PR.

It will only add the one you want if it's in the IANA registry.  I don't think we have a way to add custom ones, so if it's not in that registry, we should figure that out here.

dmvieira:
Ok, I'm solving this in another [issue](https://github.com/jshttp/mime-db/issues/176) and then we back here

PR #2933, opened 2019-10-30
Adapted @rossabaker's patch in #2891 to make it work for 2.12 as well

hamnis:
Repeating @rossabaker's comments here:

* Reimplement the case classiness of Request by hand to work around scala/bug#11457.
* Work around Seq aliasing in a spec
* Eliminate early initializer in TomcatSpec, which no longer appears to be necessary

Matching on `Message[F]` with this new encoding causes some obscure problem with the pattern matcher on 2.12. updated Request.unapply to first match on `Message[F]`, and `Request[F]`, this trick avoids the pattern matcher bug on 2.12.

### DIFF
--- a/core/src/main/scala/org/http4s/Message.scala
+++ b/core/src/main/scala/org/http4s/Message.scala
@@ -8,10 +8,13 @@ import fs2.{Pure, Stream}
 import fs2.text.{utf8Decode, utf8Encode}
 import java.io.File
 import java.net.{InetAddress, InetSocketAddress}
+
 import org.http4s.headers._
 import org.http4s.util.decode
 import org.log4s.getLogger
 import _root_.io.chrisdavenport.vault._
+
+import scala.util.hashing.MurmurHash3
 
 /**
   * Represents a HTTP Message. The interesting subclasses are Request and Response.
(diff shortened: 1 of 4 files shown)


hamnis:
I did a little experiment when I got to work this morning, and I tried to minimize the problem. This seems to work:
```scala
import scala.util.hashing.MurmurHash3
import java.net.URI

sealed trait Message[F[_]]

final class Request[F[_]](
    val method: String,
    val uri: URI
) extends Message[F] with Product with Serializable {
  override def hashCode(): Int = MurmurHash3.productHash(this)

  override def equals(that: Any): Boolean =
    (this eq that.asInstanceOf[Object]) || (that match {
      case that: Request[_] =>
        (this.method == that.method) &&
          (this.uri == that.uri)          
      case _ => false
    })

  override def canEqual(that: Any): Boolean = that.isInstanceOf[Request[F]]

  override def productArity: Int = 2

  override def productElement(n: Int): Any = n match {
    case 0 => method
    case 1 => uri
    case _ => throw new IndexOutOfBoundsException()
  }
}

object Request {
    def unapply[F[_]](req: Message[F]): Option[(String, URI)] = req match {
      case r: Request[F] => Some((r.method, r.uri))
      case _ => None
    }
}

final case class Response[F[_]](status: Int) extends Message[F]


object Foo {
    def onMessage[F[_]](message: Message[F]) = message match {
        case Request(method, uri) => s"$method $uri"
        case Response(status) => s"$status"
    }
    def onRequest[F[_]](req: Request[F]) = req match {
      case Request(method, uri) => s"$method, $uri"
    }
}
````

rossabaker:
I notice three interesting differences between the one that works and the one that doesn't:

* The one that works pattern matches on `F`.
* The one that doesn't work refers to `F` in its return type.
* The one that works returns `Option`. The one that doesn't returns `Some`. (This is a consequence of the first point, but I'll bet not the trigger.)

hamnis:
> I notice three interesting differences between the one that works and the one that doesn't:
> 
> * The one that works pattern matches on `F`.
The one that works actually matches on `Message[F]`

> * The one that doesn't work refers to `F` in its return type.

I'll expand my example to see if this actually matters.

> * The one that works returns `Option`. The one that doesn't returns `Some`. (This is a consequence of the first point, but I'll bet not the trigger.)

Returning Some is just a trick to making pattern matching faster, then the None case does not need to be considered.

rossabaker:
> The one that works actually matches on Message[F]

I'm guessing this is the trick that will make this work.  And of course the `Option` return type becomes necessary once the `None` case becomes necessary.

hamnis:
@rossabaker Rebased with master and changed to the trick in https://github.com/http4s/http4s/pull/2933#issuecomment-548261781

hamnis:
not quite sure what happened in the build. can someone restart that?

PR #2942, opened 2019-11-02
Fix Request.addCookie so it complies with the HTTP specification

ScalaFanatic:
[Relevant specification link](https://tools.ietf.org/html/rfc6265#section-4.2.1)

### DIFF
--- a/core/src/main/scala/org/http4s/Message.scala
+++ b/core/src/main/scala/org/http4s/Message.scala
@@ -340,7 +340,14 @@ sealed abstract case class Request[F[_]](
 
   /** Add a Cookie header for the provided [[Cookie]] */
   def addCookie(cookie: RequestCookie): Self =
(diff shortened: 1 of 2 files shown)


ScalaFanatic:
I cannot reproduce the failure in the test [Blaze Http1Client should call a second host after reusing connections on a first](https://travis-ci.org/http4s/http4s/jobs/606264904#L3525):
```
Some(Left(())) != Some(Right(simple path)) (BlazeClientSpec.scala:337)
```
Anyone know what that might have happened? The test ran for the full 5 seconds, but [someone else's pull request](https://travis-ci.org/http4s/http4s/jobs/605948937#L3198) from earlier today seems to have run fine, but those commits were rather inert...

ScalaFanatic:
Okay then, simply building again with Travis worked...

rossabaker:
Oops, I meant to cherry-pick this first. Oh well, it'll be in the history twice.

PR #2959, opened 2019-11-05
Release v0.20.13

rossabaker:


rossabaker:
v0.20.12 released directly from a PR without being approved.  I'm afraid it's going to happen again, but, we'll see.

rossabaker:
Yep, it published again before merge.  Need to figure that out.

nigredo-tori:
@rossabaker, some extra data points for that:
1. I've recently converted my travis account to the new workflow. That made it build all my forks, and I was too lazy to go through them and turn off each one. Anyway, it tried to build once I pushed to one of my PRs yesterday, and the build had the publish tasks in it.
1. https://config.travis-ci.com/explore gives me this (for the `publish` stage condition):
    ```
    [warn] on stages: unknown key "branches" ({:only=>["master", "/^series\\//"]})
    ```
1. [this](https://config.travis-ci.com/ref/job/branches) says that the `branches:` thing is a legacy.

So it looks like the `branches:` condition is either ignored or shadowed by the explicit `if:`, and `publish` runs on every push. We probably need to translate that condition to the `if:` form.

Issue #2943, opened 2019-11-02
Add a Request.cookies function


ghost:
To mirror [Response.cookies](https://github.com/http4s/http4s/blob/master/core/src/main/scala/org/http4s/Message.scala#L509), except it would return a `List[RequestCookie]`. I find myself using an extension method to accomplish this quite often, so I think it would be a good quality-of-life enhancement.

I can implement this if there are no objections. I will make the implementation compatible with the [HTTP specification](https://tools.ietf.org/html/rfc6265#section-4.2.1) (for more information, see #2938).

rossabaker:
I think this is reasonable.  Since [multiple cookie headers are allowed in HTTP/2.0](https://tools.ietf.org/html/rfc7540#section-8.1.2.5), I would still aggregate multiple headers.  You did the right thing in #2938 for what we _emit_, but we should _accept_ the split cookies as well.

ScalaFanatic:
Noted, I will get to work on this.

m-sp:
this was added in https://github.com/http4s/http4s/pull/2992/files#diff-adcd6450231fbb969572de7bdd97e36c7dd50fe3e577135b505c3a7e90320077R346

PR #2965, opened 2019-11-06
Add a Request.cookies function

ScalaFanatic:
As discussed in #2943, it works with both HTTP/1 and HTTP/2.

### DIFF
--- a/tests/src/test/scala/org/http4s/MessageSpec.scala
+++ b/tests/src/test/scala/org/http4s/MessageSpec.scala
@@ -126,6 +126,40 @@ class MessageSpec extends Http4sSpec {
       }
     }
 
+    "cookies" should {
+      val cookieList = List(
+        RequestCookie("test1", "value1"),
+        RequestCookie("test2", "value2"),
+        RequestCookie("test3", "value3"))
+
+      "be empty if there are no Cookie headers present" in {
+        Request(Method.GET).cookies mustEqual List.empty
+      }
+
+      "parse discrete HTTP/1 Cookie header(s) into corresponding RequestCookies" in {
+        val cookies = Header("Cookie", "test1=value1; test2=value2; test3=value3")
+        val request = Request(Method.GET, headers = Headers.of(cookies))
+        request.cookies mustEqual cookieList
+      }
+
+      "parse discrete HTTP/2 Cookie header(s) into corresponding RequestCookies" in {
+        val cookies = Headers.of(
+          Header("Cookie", "test1=value1"),
+          Header("Cookie", "test2=value2"),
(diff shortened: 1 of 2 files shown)


ScalaFanatic:
The scalafix error is
```
[error] [launcher] could not retrieve Scala 2.12.10: missing scala.Option
```
Should I just close and re-open the pull request to trigger another Travis build? I assume this is another one of those flaky issues with Travis.

tests/src/test/scala/org/http4s/MessageSpec.scala
```diff
@@ -126,6 +126,40 @@ class MessageSpec extends Http4sSpec {
...
+        RequestCookie("test2", "value2"),
+        RequestCookie("test3", "value3"))
+
+      "be empty if there are no Cookie headers present" in {
+        Request(Method.GET).cookies mustEqual List.empty
+      }
+
+      "parse discrete HTTP/1 Cookie header(s) into corresponding RequestCookies" in {
+        val cookies = Header("Cookie", "test1=value1; test2=value2; test3=value3")
+        val request = Request(Method.GET, headers = Headers.of(cookies))
```
kevinmeredith: Does your test require specifying the HTTP version when constructing the [Request](https://github.com/http4s/http4s/blob/master/core/src/main/scala/org/http4s/Message.scala#L242)?

tests/src/test/scala/org/http4s/MessageSpec.scala
```diff
@@ -126,6 +126,40 @@ class MessageSpec extends Http4sSpec {
...
+      "parse discrete HTTP/2 Cookie header(s) into corresponding RequestCookies" in {
+        val cookies = Headers.of(
+          Header("Cookie", "test1=value1"),
+          Header("Cookie", "test2=value2"),
+          Header("Cookie", "test3=value3"))
+        val request = Request(Method.GET, headers = cookies)
+        request.cookies mustEqual cookieList
+      }
+
+      "parse HTTP/1 and HTTP/2 Cookie headers on a single request into corresponding RequestCookies" in {
```
kevinmeredith: Same as https://github.com/http4s/http4s/pull/2965/files#r344472600

tests/src/test/scala/org/http4s/MessageSpec.scala
ScalaFanatic: No. That should not impact the way the headers are handled. If I need to add that for semantics, I can, but this specifically refers to the way the cookies are being sent.

kevinmeredith: 👍

rossabaker:
This is backported into #2992.

PR #2966, opened 2019-11-06
Middleware to redirect http traffic to https

tovbinm:
Inspects `X-Forwarded-Proto` header and if it is set to `http`, redirects to `Host` with same URL with https schema; otherwise does nothing. This middleware is useful when a service is deployed behind a load balancer which does not support such redirect feature, e.g. Heroku.


http4s/server/src/main/scala/org/http4s/server/middleware/HttpsRedirect.scala
```diff
@@ -0,0 +1,44 @@
...
+  */
+object HttpsRedirect {
+
+  private[HttpsRedirect] val logger = getLogger
+
+  def apply[F[_], G[_]](http: Http[F, G])(implicit F: Applicative[F]): Http[F, G] =
+    Kleisli { req =>
+      (req.headers.get(`X-Forwarded-Proto`), req.headers.get(Host)) match {
+        
+        case (Some(proto), Some(host)) if Scheme.fromString(proto.value).contains(Scheme.http) =>
```
kevinmeredith: I admit to not having heard of `X-Forwarded-Proto` before. However, could that header's value ever be `https`, such that `if Scheme.fromString(proto.value).contains(Scheme.http)` would, as I understand, wrongly evaluate to `true`, since `https` contains `http`?

kevinmeredith: Nevermind, I see that `Scheme.fromString(proto.value)` will return a `Either[ParseFailure, A]`.

tovbinm:
Thanks. It seems that I need to fix the code style before it can be merged.

tovbinm:
Done.

rossabaker:
Backported into #2992, to be released in 0.20.15.

PR #2968, opened 2019-11-06
Add Mechanism To Create Validated Timeout Values

isomarcte:
The impetus for this commit is the `verifyTimeoutRelations` method in `BlazeClientBuilder`. That method provides a runtime check with a warn log that the timeout values for a client pass some sanity checks.

This commit adds a new ADT `Timeout` and a product of all timeout types `Timeouts`. The latter provides a mechanism to ensure that the timeout values constructed are always valid.

### DIFF
--- a/client/src/main/scala/org/http4s/client/Timeout.scala
+++ b/client/src/main/scala/org/http4s/client/Timeout.scala
@@ -1,0 +1,50 @@
+package org.http4s
+package client
+
+import cats._
+import scala.concurrent.duration._
+
+/** ADT for the timeout types which may be used by a http4s client.
+  *
+  * Not all clients support all timeout types.
+  *
+  * See [[Timeouts]] for a mechanism to construct timeout values which always
+  * line up to expected constraints, e.g. responseHeaderTimeout < idleTimeout.
+  */
+sealed trait Timeout extends Product with Serializable {
+  def value: Duration
+}
+
+object Timeout {
+
+  implicit val showInstance: Show[Timeout] =
+    Show.fromToString
+
+  final case class ConnectTimeout(override final val value: Duration) extends Timeout
+
+  object ConnectTimeout {
+    val Inf: ConnectTimeout = ConnectTimeout(Duration.Inf)
+    val default: ConnectTimeout = ConnectTimeout(10.seconds)
+  }
+
+  final case class IdleTimeout(override final val value: Duration) extends Timeout
+
+  object IdleTimeout {
+    val Inf: IdleTimeout = IdleTimeout(Duration.Inf)
+    val default: IdleTimeout = IdleTimeout(1.minute)
+  }
+
+  final case class RequestTimeout(override final val value: Duration) extends Timeout
+
+  object RequestTimeout {
+    val Inf: RequestTimeout = RequestTimeout(Duration.Inf)
+    val default: RequestTimeout = RequestTimeout(45.seconds)
+  }
+
+  final case class ResponseHeaderTimeout(override final val value: Duration) extends Timeout
+
+  object ResponseHeaderTimeout {
+    val Inf: ResponseHeaderTimeout = ResponseHeaderTimeout(Duration.Inf)
+    val default: ResponseHeaderTimeout = this.Inf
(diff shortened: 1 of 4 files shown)


client/src/main/scala/org/http4s/client/Timeout.scala
```diff
@@ -0,0 +1,50 @@
...
+import scala.concurrent.duration._
+
+/** ADT for the timeout types which may be used by a http4s client.
+  *
+  * Not all clients support all timeout types.
+  *
+  * See [[Timeouts]] for a mechanism to construct timeout values which always
+  * line up to expected constraints, e.g. responseHeaderTimeout < idleTimeout.
+  */
+sealed trait Timeout extends Product with Serializable {
```
rossabaker: Is the value of these subtypes to prevent people calling them in the wrong order on `Timeouts`?

isomarcte:
> 1. Do we want to turn nonsensical timeouts into a runtime error at backend construction? I could argue this either way.

So the intent of this change was to leave that choice up to the user. If they want to validate that their timeout value selection is logical at construction time, they can use the `Timeouts` product and associated methods, if they don't really care then they can continue to just get the warnings in their logs.

> 2. Do we want all these wrapper types in the public API, giving a second way to do it? I'm concerned that our timeouts are already complex, and am worried about introducing a synonymous way of configuring them.

Yeah I completely agree here. If this PR was accepted I was going to propose replacing all the places we currently use raw `Duration` values to configured timeouts with the equivalent `Timeout` constructor from the ADT. This would both be to address the issue with having multiple ways to configure timeouts and make positional arguments a bit more safe since swapping places would create a type error.

This still leaves an issue with using individual `Timeout` instances vs the `Timeouts` product type. I could restructure `Timeouts` into and ADT too, with a validated and unvalidated constructors. Something like,

```scala
sealed trait Timeouts {
  def connectTimeout: ConnectTimeout
  def idleTimeout: IdleTimeout
  def requestTimeout: RequestTimeout
  def responseHeaderTimeout: ResponseHeaderTimeout
}

object Timeouts {
  final case class UnvalidatedTimeouts(
    override final val connectTimeout: ConnectTimeout,
    override final val idleTimeout: IdleTimeout,
    override final val requestTimeout: RequestTimeout,
    override final val responseHeaderTimeout: ResponseHeaderTimeout
  ) extends Timeouts

  sealed trait ValidatedTimeouts extends Timeouts

  private[this] object ValidatedTimeouts {
    private[this] final case class ValidatedTimeoutsImpl(
      override final val connectTimeout: ConnectTimeout,
      override final val idleTimeout: IdleTimeout,
      override final val requestTimeout: RequestTimeout,
      override final val responseHeaderTimeout: ResponseHeaderTimeout
    ) extends ValidatedTimeouts
  }
  /* Validation code for constructing ValidatedTimeouts elided */
}
```

Then we could have all the clients take a `Timeouts` as the exclusive mechanism for expressing timeouts. There are two downsides to this approach that I see.

* Not all of the clients are configurable with respect to all the timeouts. Forcing everyone to go through the `Timeouts` interface then adds its own level of confusion as users may logically expect that all the timeouts will be used.
* The encoding of `Timeouts` gets quite a bit uglier, though this quite subjective and is probably not a big deal.

So there seem to be downsides with all approaches...thoughts?

client/src/main/scala/org/http4s/client/Timeout.scala
isomarcte: Yes, and also it could be useful in other places in the API where we have raw `Duration` values in neighboring positions in parameters.

isomarcte:
I think the shape of things has changed a bit since I opened this PR. We may still want some sort of tighter validation on timeout construction, but things are so variable between the different implementations we probably should address from a fresh context.

rossabaker:
I'm sorry this got forgotten after you put so much work into it.

Looking at it with fresh eyes, I think I'm comfortable issuing a warning in the logs.  A misordering is not something we can catch at compile time, and we can still do what was asked, even if it's silly.  I don't think it's worth raising an error instead of starting a server.

Issue #2883, opened 2019-09-23
[feature] Expose connection pool manager's waitQueue size


anilkumarmyla:
Our use-case is to log this metric to drive decrease/increase in compute instances

https://github.com/http4s/http4s/blob/master/client/src/main/scala/org/http4s/client/PoolManager.scala#L47

kevinmeredith:
My initial thought was adding a `def waitQueueSize: F[Int]` to [trait ConnectionManager](https://github.com/http4s/http4s/blob/series/0.20/client/src/main/scala/org/http4s/client/ConnectionManager.scala#L18).

But, per https://stackoverflow.com/a/18369508/409976, I'm not sure if it would break binary compatibility.

Thoughts, please, @rossabaker and @ChristopherDavenport?

kevinmeredith:
Any thoughts, please, Chris or Ross?

rossabaker:
Basically a duplicate of #1270

PR #2971, opened 2019-11-10
[WIP] Expose wait queue 

kevinmeredith:
Closes https://github.com/http4s/http4s/issues/2883

### DIFF
--- a/blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClient.scala
+++ b/blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClient.scala
@@ -47,7 +47,7 @@ object BlazeClient {
       requestTimeout: Duration,
       scheduler: TickWheelExecutor,
       ec: ExecutionContext
-  )(implicit F: ConcurrentEffect[F]) =
+  )(implicit F: ConcurrentEffect[F]): Client[F] =
     Client[F] { req =>
       Resource.suspend {
         val key = RequestKey.fromRequest(req)
--- a/blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala
+++ b/blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala
@@ -71,7 +71,8 @@ sealed abstract class BlazeClientBuilder[F[_]] private (
       executionContext: ExecutionContext = executionContext,
       scheduler: Resource[F, TickWheelExecutor] = scheduler,
       asynchronousChannelGroup: Option[AsynchronousChannelGroup] = asynchronousChannelGroup,
-      channelOptions: ChannelOptions = channelOptions
+      channelOptions: ChannelOptions = channelOptions,
+      connectionManager: ConnectionManager[F, BlazeConnection[F]] = connectionManager
   ): BlazeClientBuilder[F] =
     new BlazeClientBuilder[F](
       responseHeaderTimeout = responseHeaderTimeout,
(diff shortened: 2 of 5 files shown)


kevinmeredith:
Hi @ChristopherDavenport and @rossabaker - in order to expose the `BlazeClient`'s current wait queue, it seems necessary to add a `def currentWaitQueueSize: F[Option[Int]]` method to `trait ConnectionManager`, per this PR. However, based on a glance at the `Client`-related code, it would seem then necessary to expose the same method on `trait Client`. 

I do question, however, whether that would make sense. As I understand, the `AsyncHttpClient` does not have the concept of a wait queue.

In short, please advise me on this question, namely how to expose the current wait queue size. Also, I appreciate any comments, please, regarding your thoughts on https://github.com/http4s/http4s/issues/2883.

rossabaker:
I would not want to add it to the `Client` interface.  I think it would make more sense to make the connection manager a parameter on relevant client builders, and then a handle could be obtained to the metrics.  Otherwise, we're into subtyping land.

kevinmeredith:
> I think it would make more sense to make the connection manager a parameter on relevant client builders, and then a handle could be obtained to the metrics. Otherwise, we're into subtyping land.

Thanks for the guidance, @rossabaker. I began, per https://github.com/http4s/http4s/pull/2971/commits/a7f491bbe4f4072f3a6a8632c74c406ff73a22af, to add `ConnectionManager[F, BlazeConnection[F]]` as a field to `BlazeClientBuilder[F]`.

However, I decided to abandon that approach since it required: supplying a `ConnectionManager[F, BlazeConnection[F]]` parameter to `BlazeClientBuilder#apply`. That, it seemed to me, made the API less straightforward since it would then require the client to construct a `ConnectionManager[F, BlazeConnection[F]]`, and then pass it. Lastly, with such an approach, it would defeat, I think, the purpose of `BlazeClientManager[F]` since, once the `ConnectionManager[F, BlazeConnection[F]]`, has been created, there's no further need for exposing `SSLContext`, connection timeout, etc, in the `BlazeClientBuilder[F]` since they've already been used in making the `ConnectionManager[F, BlazeConnection[F]]`. 

I have another idea that I'm working on now.

kevinmeredith:
> 

That thought was to change `BlazeClientBuilder`'s `scheduler` field from `Resource[F, TickWheelExecutor]` to `TickWheelExecutor`, guessing that would allow me to construct an `ConnectionManager[F, BlazeConnection[F]]` from it. However, I believe I'm wrong since creating a `ConnectionManager[F, BlazeConnection[F]]` is effectful per https://github.com/http4s/http4s/blob/v0.20.14/blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala#L241-L242. 

At this point, my only immediate thoughts are to return to passing `ConnectionManager[F, BlazeConnection[F]]` to `BlazeClientBuilder#apply`, but that includes the problems, in my observation, that I listed in https://github.com/http4s/http4s/pull/2971#issuecomment-559340859 My other thought, which you already shut down, is adding to `Client`.

Given that I'm not entirely sure how to proceed, could you please advise me, @rossabaker? I'll think about it more, but I appreciate any help. Thanks

kevinmeredith:
Any thoughts, please, @rossabaker on https://github.com/http4s/http4s/pull/2971#issuecomment-559343295?

kevinmeredith:
Any thoughts, please, @rossabaker, on https://github.com/http4s/http4s/pull/2971#issuecomment-559343295?

kevinmeredith:
I'm figuring this one won't go anywhere, so I'll close.

rossabaker:
I apologize.  I've been in a bad habit of treating my notifications as a stack instead of a queue, and this one got buried for far too long.  I think it's important that we have an idiom for exposing stats on the connection pool.  If you're still interested in the approach, I'll give it a closer look.

kevinmeredith:
Hey @rossabaker - thanks for the encouragement! 

I will, most likely, re-visit the PR,  re-read where I got stuck, re-open it, and then `@` with you a question on how to proceed.

Issue #2969, opened 2019-11-08
Create prometheus MetricsOps using Resource


iRevive:
Collectors are being registered in the collector registry upon creation. But they are never being released.   

Current implementation: https://github.com/http4s/http4s/blob/master/prometheus-metrics/src/main/scala/org/http4s/metrics/prometheus/Prometheus.scala#L167

```scala
Histogram
  .build()
  .buckets(responseDurationSecondsHistogramBuckets.toList: _*)
  .name(prefix + "_" + "response_duration_seconds")
  .help("Response Duration in seconds.")
  .labelNames("classifier", "method", "phase")
  .register(registry)
```

From my point of view, a collector should be treated as a resource.

Example:
```scala

def responseDurationHistogramResource = Resource.make {
  Sync[F].delay {
   Histogram
     .build()
     .buckets(responseDurationSecondsHistogramBuckets.toList: _*)
     .name(prefix + "_" + "response_duration_seconds")
     .help("Response Duration in seconds.")
     .labelNames("classifier", "method", "phase")
     .register(registry)
  }
} { histogram =>
  Sync[F].delay(registry.unregister(histogram)))
}

val metricsOps: Resource[F, MetricsOps[F]] = 
  for {
    responseDurationHistogram <- responseDurationHistogramResource
    ...
    metrics = MetricsCollection(responseDurationHistogram, ...)
  } yield new MetricsOps { ... }
```

If you are fine with the idea, I can take care of the implementation.

ChristopherDavenport:
I truly had no idea there was such a thing as unregistering.

zhenik:
Do you have plans to wrap up [custom collectors](https://github.com/prometheus/client_java#custom-collectors) injection or leave it via `collectorRegistry.register(new SomeCustomCollector())`?

Like [here](https://github.com/http4s/http4s/blob/853689e1334e63501392ad3cfcfc50a031833e8e/prometheus-metrics/src/main/scala/org/http4s/metrics/prometheus/PrometheusExportService.scala#L53-L60) defaults are added.

iRevive:
@zhenik 

I have changed the injection of the default collectors: https://github.com/http4s/http4s/pull/2972/files#diff-85d565b2a2f68a61c171e7922a244d9eR52. 

Perhaps it makes sense to make `Prometheus.registerCollector` public:
https://github.com/http4s/http4s/pull/2972/files#diff-16a0754058a0d75db6fd148fc7c8336fR228

PR #2972, opened 2019-11-10
Unregister Prometheus metrics

iRevive:
This PR closes https://github.com/http4s/http4s/issues/2969.

WDYT @ChristopherDavenport ?

### DIFF
--- a/docs/src/main/tut/client.md
+++ b/docs/src/main/tut/client.md
@@ -263,18 +263,20 @@ We can create a middleware that registers metrics prefixed with a
 provided prefix like this.
 
 ```tut:silent
+import cats.effect.{Resource, IO}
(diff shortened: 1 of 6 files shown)


iRevive:
One more thing. 
Can the effect constraint be removed from [PrometheusExportService](https://github.com/http4s/http4s/blob/75de389c8a81dc9b00f03614e6a55f50550cbe68/prometheus-metrics/src/main/scala/org/http4s/metrics/prometheus/PrometheusExportService.scala#L18)?  
Due to the private constructor, I don't see any reason to keep it.

```
final class PrometheusExportService[F[_]: Sync] private ( ... )
```

ChristopherDavenport:
Effect is almost certainly just a remnant.

PR #3002, opened 2019-11-29
Parametrize NotFound

CharlesD91:
- **Problem:** `NotFound` response is hardcoded

- **Solution:** add `JettyBuilder.withNotFoundResponse(notFoundResponse: Request[F] => Response[F])`

- **Unit test:** see `NotFoundSpec.scala`

### DIFF
--- a/jetty/src/main/scala/org/http4s/server/jetty/JettyBuilder.scala
+++ b/jetty/src/main/scala/org/http4s/server/jetty/JettyBuilder.scala
@@ -30,7 +30,8 @@ sealed class JettyBuilder[F[_]] private (
     sslBits: Option[SSLConfig],
     mounts: Vector[Mount[F]],
     private val serviceErrorHandler: ServiceErrorHandler[F],
-    banner: immutable.Seq[String]
+    banner: immutable.Seq[String],
+    private val notFoundResponse: Option[Request[F] => Response[F]]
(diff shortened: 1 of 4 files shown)


CharlesD91:
Hi Ross, Thank you very much for taking the time to reply. 

I looked at how `Blaze` is used and it looks like it doesn't have a default behavior for `NotFound` (while `Jetty` does). If you don't use the `.orNotFound` syntax when building the `HttpApp` then the server will throw a `scala.MatchError` on unspecified endpoints. So if `Jetty` will have the same style as `Blaze` in the future, then I can simply compose the `HttpRoutes` with a `NotFoundRoutes` that matches all the remaining requests: as in https://github.com/http4s/http4s/pull/3002/files#diff-6b8603ed2bda2b6badb1b19f63335be1 (see `case req@_ => NotFound(s"Not Found ${req.uri.path}")`).

Then there is no need for a `JettyBuilder.withNotFoundResponse(notFoundResponse: Request[F] => Response[F])` and I will delete my PR.

CharlesD91:
https://github.com/http4s/http4s/pull/3012 was merged, so now there is `JettyBuilder.mountHttpApp`. It is now convenient to handle all `NotFound` on endpoints not defined in `HttpRoutes`: simply handle them when converting `HttpRoutes` (`partial` function) into `HttpApp` (`total` function).

PR #3005, opened 2019-12-01
Add a nix shell

rossabaker:
* Totally opt-in for now, though I'd like to investigate integrating with CI.
* Run `nix-shell` to install all the dev tools necessary to run `sbt ci`.
* Pins a hugo version such that the site works.  It's clashing with my other hugos.  (Most of the heft in this PR is related to this, and can be removed when we refresh hugo.)
* Integrates with direnv so cd'ing into the project or editing a file in the project sets it up.
* Makes us cool like cats, which has had one for a while.


rossabaker:
329 days without dissent. :heavy_check_mark:

PR #3006, opened 2019-12-01
Start replacing unsafeRunSync in tests

rossabaker:
An attempt to improve the health of our CI and set a better example for others.


rossabaker:
It's not really necessary to do them all at once.  We can make Good First Issue tickets by splitting this up by module.  Let's make incremental progress here.

PR #3012, opened 2019-12-06
Use HttpApp instead of HttpRoutes in Http4sServlet

CharlesD91:
- Same `NotFound` behavior as with `HttpRoutes` for unspecified endpoints (see `NotFoundHttpAppSpec`)

- Preserve retrocompatibility in `JettyBuilder`, `ServletContextSyntax`, `TomcatBuilder`


tomcat/src/main/scala/org/http4s/server/tomcat/TomcatBuilder.scala
```diff
@@ -120,6 +121,9 @@ sealed class TomcatBuilder[F[_]] private (
     })
 
   def mountService(service: HttpRoutes[F], prefix: String): Self =
+    mountServiceHttpApp(http4s.httpRoutesToApp(service), prefix)
+
+  def mountServiceHttpApp(service: HttpApp[F], prefix: String): Self =
```
rossabaker: I think I would just call this `mountHttpApp`.

On blaze, it's called `witHttpApp`, but the servlet model is a little different: we have more than one.  So I think the different name is justified.

jetty/src/test/scala/org/http4s/server/jetty/Issue454.scala
```diff
@@ -44,7 +44,7 @@ object Issue454 {
   }
 
   val servlet = new AsyncHttp4sServlet[IO](
-    service = HttpRoutes.of[IO] {
+    service = HttpApp {
       case GET -> Root => Ok(insanelyHugeData)
     },
```
nigredo-tori: Please try and avoid inexhaustive matches. This can be done with `httpRoutesToApp`, or by adding a catch-all clause below.

core/src/main/scala/org/http4s/package.scala
```diff
@@ -77,4 +78,7 @@ package object http4s { // scalastyle:ignore
   type Http4sSyntax = syntax.AllSyntax
   @deprecated("Moved to org.http4s.syntax.all", "0.16")
   val Http4sSyntax = syntax.all
+
+  def httpRoutesToApp[F[_]: ConcurrentEffect](httpRoutes: HttpRoutes[F]): HttpApp[F] =
+    HttpApp[F](req => httpRoutes(req).getOrElse[Response[F]](Response.notFound[F]))
```
nigredo-tori: Actually, we already have a syntax for this. With `import org.http4s.syntax.kleisli._` we can do `httpRoutes.orNotFound`. So I don't think we need this as a separate function.

servlet/src/main/scala/org/http4s/servlet/AsyncHttp4sServlet.scala
```diff
@@ -66,9 +63,9 @@ class AsyncHttp4sServlet[F[_]](
       F.asyncF[Response[F]](cb => gate.complete(ctx.addListener(new AsyncTimeoutHandler(cb))))
     val response =
       gate.get *>
-        optionTSync
+        Sync[F]
           .suspend(serviceFn(request))
-          .getOrElse(Response.notFound)
+          .recover({ case _: scala.MatchError => Response.notFound[F] })
```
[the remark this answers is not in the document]
CharlesD91: It would be nice to be able to handle `NotFound` in a single place, that's why I used a `PartialFunction` and then handle the `NotFound` in the `recover`. Otherwise, when you have multiple services, you have to add `case _ => handleNotFound()` at the end of each of them. 

Furthermore, if you don't mount a service on the prefix `/`, then a request on anything caught by `/` will return an HTML that doesn't seem to be programmable.

jetty/src/test/scala/org/http4s/server/jetty/Issue454.scala
hamnis: use the `.orNotFound` syntax in as described below

servlet/src/main/scala/org/http4s/servlet/AsyncHttp4sServlet.scala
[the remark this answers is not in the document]
nigredo-tori: Sorry, I'm not quite sure what you're getting at here. I'll try and answer some of the points you are trying to make...

First, it looks like you've added this `recover` as a way to solve some issue in `JettyBuilder`, and it handles the situation that should not happen during any *normal* use of the `AsyncHttp4sServer`. This smells.

Second, in the Cats world we make a lot of assumptions, one of which is that all the functions are total except for `PartialFunction`s. We can only get a `MatchError` by breaking this rule via a non-exhaustive match - which, by the way, would lead to a compiler warning. Generally, any `MatchError` thrown signals broken code, and we should never try specifically handling it for an arbitrary function provided by the user code, never mind swallowing it without reporting. Basically, if you want to support a non-exhaustive `service` - use `HttpRoutes` instead. This is exactly what it is there for.

Third, regarding `case _ => handleNotFound()` - we can do `.orNotFound` when mounting `HttpRoutes` in `JettyBuilder`. I'm not sure why that should be a problem.

Finally, regarding `/`. As I understand the problem (if you can call it that) is that Jetty catches any requests that were not handled, and shows a standard 404 error page. I'm not sure it's our place to solve that by default, but even if it were - this could be solved by mounting a servlet to `/`. I'm still not sure what this has to do with the servlet implementation, though.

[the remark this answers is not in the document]
CharlesD91: Ok thank you for your comments. I think I understand your points. I will try to update my PR accordingly.

nigredo-tori:
Actually, if the motivation behind this PR is to just add `mountHttpApp` to the builders - can't we just add something like this to `JettyBuilder` and `TomcatBuilder`?
```scala
def mountHttpApp(service: HttpApp[F], prefix: String): Self =
    mountService(service.mapK(OptionT.liftK[F])), prefix)
```
This way we wouldn't have to break anything.

Unless, of course the change to the servlets is the main goal - in which case I don't see the point of it, since we can always get a `HttpApp` from `HttpRoutes` with `.orNotFound`.

servlet/src/main/scala/org/http4s/servlet/AsyncHttp4sServlet.scala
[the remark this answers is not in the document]
CharlesD91: @nigredo-tori I added `def orNotFound(notFoundHandler: A => Response[F]): Kleisli[F, A, Response[F]]` to `KleisliSyntax`. Also I updated `NotFoundHttpAppSpec`. It describes what I am trying to achieve here: having a simple, single way of handling custom `NotFound` responses accross the server. Do you think this would be reasonable ?

(also I agree with your latest comment about simply adding 
```
def mountHttpApp(service: HttpApp[F], prefix: String): Self =
    mountService(service.mapK(OptionT.liftK[F])), prefix)
```
to the builders)

nigredo-tori:
So, to summarize the discussion (and to get my thoughts in order):
1. You want to add a way to mount `HttpApp`s in Jetty and Tomcat builders via `mountHttpApp`. I think this is useful, and it should be done.
1. The easiest way to do this is to add `mountHttpApp` that factors through `mountService`, as I proposed before. However, in this case the `HttpApp` would be transformed to `HttpRoutes`, just to be later made back into a `HttpApp` in the servlet implementation via `.orNotFound`. This is a bit silly.
1. Looking at the servlet implementations, it makes sense for them to receive a `HttpApp` instead of `HttpRoutes` they get now, since the `Servlet` interface itself implies that every request should get a response. It makes sense to move `.orNotFound` outside the servlet implemenation, since this adds flexibility. So you've changed the constructors for servlet implementations to use `HttpApp` (so that `mountService` factors through `mountHttpApp`). Even though this is a breaking change, it shouldn't be hard to fix in the calling code (a single `.orNotFound`), and there's probably not a lot of such code out there. All of this seems completely reasonable to me.
1. The `recover { case _: MatchError =>` thing is a hack to make tests compile without updating them. It should be removed.
1. Regarding the tests for all of the above, I don't think we need any beyond updating the ones that are already there. This includes fixing any non-exhaustive matches.

servlet/src/main/scala/org/http4s/servlet/AsyncHttp4sServlet.scala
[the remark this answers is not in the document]
nigredo-tori: I've tried and summarized the important points of this below, and I suggest we continue this discussion there, since the scope of it went far beyond the single line of code I want gone. :smile:

core/src/main/scala/org/http4s/syntax/KleisliSyntax.scala
```diff
@@ -30,6 +30,9 @@ trait KleisliSyntaxBinCompat1 {
 final class KleisliResponseOps[F[_]: Functor, A](self: Kleisli[OptionT[F, ?], A, Response[F]]) {
   def orNotFound: Kleisli[F, A, Response[F]] =
     Kleisli(a => self.run(a).getOrElse(Response.notFound))
+
+  def orNotFound(notFoundHandler: A => Response[F]): Kleisli[F, A, Response[F]] =
+    Kleisli(a => self.run(a).getOrElse(notFoundHandler(a)))
```
nigredo-tori: I don't think this use case (a custom fall through handler) is a common enough use case to warrant special syntax. I can also think of a few other ways to implement similar functionality - instead of `A => Response[F]` we can use `Response[F]`, `F[Response[F]]`, `A => F[Response[F]]`, `HttpApp[F]`... All of those potential functions are equally valid, and would have their uses - I don't think the one you've chosen is any more common than the others.

If you insist, I'd like you to at least choose another name. Overloading should be avoided in general, and `orNotFound` is misleading here, since it has no relation to `NotFound`.

CharlesD91: It would be nice to be able to do:

```
JettyBuilder[IO]
      .bindAny()
      .mountServiceHttpApp(
        HttpRoutes
          .of[IO](Map.empty)
          .orNotFound(myCustomNotFoundHandler),
        "/"
      )
      .mountServiceHttpApp(
        HttpRoutes
          .of[IO] {
            case GET -> Root / "hello" => Ok("hello")
          }
          .orNotFound(myCustomNotFoundHandler),
        "/prefix1"
      )
```
So that you can see in one place how all `non-matched endpoints` are handled and that they are no leaks. The method name could be `.orHandleNonMatch`. I agree with all your other points.

nigredo-tori: > So that you can see in one place how all non-matched endpoints are handled and that there are no leaks.

You don't need a (library) syntax for that - it works just as well with a custom function:
```scala
def orMyHandler(x: HttpRoutes[F]): HttpApp[F] =
  Kleisli { req => x.run(req).getOrElse(myCustomNotFoundHandler(req)) }

//...
.mountServiceHttpApp(
  orMyHandler(httpRoutes0),
  "/"
)
// ...
```

As an aside, I've never seen someone mount a lot of services onto one builder like this. I'd rewrite your example like this instead:

```scala
JettyBuilder[IO]
  .bindAny()
  .mountServiceHttpApp(
    myOrHandler(
      Router(
        "" -> httpRoutes0,
        "prefix1" -> httpRoutes1,
        ...
        "prefixN" -> httpRoutesN
      )
    )
  )
```

CharlesD91: Thank you, this is exactly what I have been looking for !

More generally, do you think there is a need for a `JettyBuilder.mountServiceHttpApp` ? I would say yes because as `.mountServiceHttpApp` expects a `total` function then it is clearer to see how `NotFound` for instance are handled. Also it would make sense to unify the APIs between `Jetty` and `Blaze`. What do you think ?

nigredo-tori: Added clarity is not as important here; AFAICT, people tend to separate their routing logic from the server itself (http4s is great in that), so whoever wants to be sure they have matched all the possible options is probably already using `HttpApp` anyway, and just rewraps that as `HttpRoutes`. I think `mountServiceHttpApp` (or, rather, `mountHttpApp`) will be useful chiefly because it would allow us to avoid that annoying rewrapping.

> Also it would make sense to unify the APIs between Jetty and Blaze.

Not really. As it was mentioned here, Jetty and Blaze builders have fundamentally different capabilities for mounting routes, so I don't see an attempt to unify them leading to anything other than extra complexity and confusion.

CharlesD91: > I think mountServiceHttpApp (or, rather, mountHttpApp) will be useful chiefly because it would allow us to avoid that annoying rewrapping.

Useless rewrapping on the library-user side is always good, I will simplify this PR and follow your advices

servlet/src/test/scala/org/http4s/servlet/BlockingHttp4sServletSpec.scala
```diff
@@ -14,7 +14,7 @@ import scala.concurrent.duration._
 
 class BlockingHttp4sServletSpec extends Http4sSpec {
 
-  lazy val service = HttpRoutes.of[IO] {
+  lazy val service = HttpApp[IO] {
```
nigredo-tori: This match is now non-exhaustive! Please add `.orNotFound` below.

jetty/src/test/scala/org/http4s/server/jetty/NotFoundHttpAppSpec.scala
```diff
@@ -0,0 +1,75 @@
...
+import java.net.{HttpURLConnection, URL}
+
+import cats.data.Kleisli
+import cats.effect.IO
+import org.http4s.dsl.io._
+
+import scala.io.Source
+import scala.util.Try
+
+class NotFoundHttpAppSpec extends Http4sSpec {
```
nigredo-tori: We no longer need this spec - it doesn't test anything that isn't covered elsewhere.

CharlesD91:
Hello @nigredo-tori , I have fixed the PR following your advice.

I kept `NotFoundHttpAppSpec` because I think it is good to have a test on `JettyBuilder.mountHttpApp`. Also it displays how `NotFound` can be handled, while it is handled by default when using `JettyBuilder.mountService`.

---

For some reason Travis CI keeps failing on 1 of the 3 test machines:
```
[error] ## Exception when compiling 2 sources to /home/travis/build/http4s/http4s/json4s-jackson/target/scala-2.13/classes
[error] scala.reflect.internal.FatalError: Error accessing /home/travis/.cache/coursier/v1/https/repo1.maven.org/maven2/com/fasterxml/jackson/core/jackson-databind/2.9.8/jackson-databind-2.9.8.jar 
```
but it is successful on the 2 others.

Please let me know what you think and thank you for your help !

servlet/src/test/scala/org/http4s/servlet/BlockingHttp4sServletSpec.scala
CharlesD91: This was fixed in my latest commit

jetty/src/test/scala/org/http4s/server/jetty/NotFoundHttpAppSpec.scala
CharlesD91: Ok, I removed it

nigredo-tori:
> I kept NotFoundHttpAppSpec because I think it is good to have a test on JettyBuilder.mountHttpApp. 

We already test it in `JettyServerSpec` since `mountService` factors through it.

> Also it displays how NotFound can be handled, while it is handled by default when using JettyBuilder.mountService.

This by itself is not a sufficient reason for adding tests. Also, the most common case for making a `HttpApp` from `HttpRoutes` (`.orNotFound`) is already used extensively throughout the documentation. *Manually* handling this depends on the use case, and should be reasonably easy to arrive to for people familiar with `Kleisli` and `OptionT`, just by looking at `HttpRoutes` and `HttpApp` types.

nigredo-tori:
:thumbsup: from me. @hamnis, @rossabaker, please take a look again - this PR has gotten a lot lighter.

hamnis:
I cleared the travis cache and restarted the build, can be merged on build success.

PR #3017, opened 2019-12-12
Fix scaladoc

dabd:
Fix scaladoc: the timeout response is a `503 Service Unavailable` instead of `500 Internal Server Error`.

### DIFF
--- a/server/src/main/scala/org/http4s/server/middleware/Timeout.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/Timeout.scala
@@ -13,8 +13,7 @@ object Timeout {
     * duration if the service has not yet responded.  If the timeout
     * fires, the service's response is canceled.
     *
-    * @param timeout Finite duration to wait before returning a `503
-    * Service Unavailable` response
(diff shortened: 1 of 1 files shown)


server/src/main/scala/org/http4s/server/middleware/Timeout.scala
```diff
@@ -13,8 +13,8 @@ object Timeout {
     * duration if the service has not yet responded.  If the timeout
     * fires, the service's response is canceled.
     *
-    * @param timeout Finite duration to wait before returning a `500
-    * Internal Server Error` response
+    * @param timeout Finite duration to wait before returning a `503
+    * Service Unavailable` response
```
hamnis: this is still wrong, it should be more in line with `the provided response`

dabd: Changed in https://github.com/http4s/http4s/pull/3017/commits/7ea9db59afd9334f635d6c7a5ffeabf9249c162f

PR #3024, opened 2019-12-15
Implement withQueryParams and withMultiValueQueryParams

ScalaFanatic:
I could not find a more elegant way of returning `Self` (and folding on `self`) other than the proposed
```scala
replaceQuery(Query.fromVector(s.withQueryParam(k, v).query.toVector))
```
If anyone has input on this, let me know.

### DIFF
--- a/core/src/main/scala/org/http4s/QueryOps.scala
+++ b/core/src/main/scala/org/http4s/QueryOps.scala
@@ -143,7 +143,7 @@ trait QueryOps {
     * [[Query]]. If any of the given parameters' keys already exists, the
     * value(s) will be replaced. Parameters from the input map are added
     * left-to-right, so if a parameter with a given key is specified more than
-    * once, it will be self-overriding.
+    * once, it will be self-overwriting.
     */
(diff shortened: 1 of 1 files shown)


ScalaFanatic:
The test failure appears to be another one of those spotty Blaze tests.

danielvasilev:
@ScalaFanatic do you think it would be beneficial to have a couple of specs for this?

ScalaFanatic:
> @ScalaFanatic do you think it would be beneficial to have a couple of specs for this?

Actually, I considered that, but was unsure how to go about implementing them because `QueryOps` is a trait. I did not find tests for any other functions in `QueryOps`, so I just decided to skip it. I could look into adding specs if there is formal guidance on the methodology.

rossabaker:
This one fell victim to my stack-based approach to my GitHub notifications instead of queue-based.  Sorry.  Milestoned so it's properly considered for the RC cycle this weekend.

rossabaker:
Spec debt captured in #3090.

PR #3031, opened 2019-12-19
Relaxing constraints on Router.apply

gvolpe:
`Monad` is needed for `<+>`, otherwise it could work with only `Applicative`.


gvolpe:
It seems the CI build fail is unrelated to this PR?

PR #3049, opened 2020-01-08
Date Middleware

ChristopherDavenport:
Adds Date Header to all responses generated by the Server.

### DIFF
--- a/server/src/main/scala/org/http4s/server/middleware/Date.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/Date.scala
@@ -16,7 +16,7 @@ import java.time.Instant
   * by the service.
  **/
 object Date {
-  def apply[G[_]: Monad: Clock, F[_], A](
+  def apply[G[_]: MonadError[*[_], Throwable]: Clock, F[_], A](
       k: Kleisli[G, A, Response[F]]): Kleisli[G, A, Response[F]] =
     Kleisli { a =>
       for {
@@ -27,9 +27,8 @@ object Date {
             Clock[G]
               .realTime(MILLISECONDS)
               .map(Instant.ofEpochMilli(_))
-              .map(nowInstant => HDate(HttpDate.unsafeFromInstant(nowInstant)))
-            // Starting on January 1,n 10000, this will throw an exception.
-            // The author intends to leave this problem for future generations.
+              .flatMap(nowInstant => HttpDate.fromInstant(nowInstant).liftTo[G])
+              .map(nowHttpDate => HDate(nowHttpDate))
           )(_.pure[G])
 
       } yield resp.putHeaders(header)
--- a/server/src/main/scala/org/http4s/server/middleware/Date.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/Date.scala
@@ -11,23 +11,27 @@ import scala.concurrent.duration.MILLISECONDS
 
 import java.time.Instant
 
+/**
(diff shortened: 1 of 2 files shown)


server/src/test/scala/org/http4s/server/middleware/DateSpec.scala
```diff
@@ -0,0 +1,74 @@
...
+import cats.implicits._
+import cats.effect._
+import cats.data.OptionT
+import org.http4s._
+import org.http4s.headers.{Date => HDate}
+import cats.effect.testing.specs2.CatsIO
+
+class DateSpec extends Http4sSpec with CatsIO {
+
+  implicit val T = IO.timer(scala.concurrent.ExecutionContext.global)
```
rossabaker: Http4sSpec has an implicit `Timer[IO]` that doesn't use global, which leaks threads in long-running sbt sessions.

server/src/test/scala/org/http4s/server/middleware/DateSpec.scala
```diff
@@ -0,0 +1,74 @@
...
+  "Date" should {
+    "always be very shortly before the current time httpRoutes" >> {
+      for {
+        out <- testService(req).value
+      } yield {
+        out.flatMap(_.headers.get(HDate)) must beSome.like {
+          case date =>
+            val diff = date.date.epochSecond - HttpDate.now.epochSecond
+            val test = diff <= 2
+            test must beTrue
```
rossabaker: I think `test must be_<=(2)` works, and would give you the value of `test` if it fails.

server/src/main/scala/org/http4s/server/middleware/Date.scala
```diff
@@ -0,0 +1,36 @@
...
+      for {
+        resp <- k(a)
+        header <- resp.headers
+          .get(HDate)
+          .fold(
+            Clock[G]
+              .realTime(MILLISECONDS)
+              .map(Instant.ofEpochMilli(_))
+              .flatMap(nowInstant => HttpDate.fromInstant(nowInstant).liftTo[G])
+              .map(nowHttpDate => HDate(nowHttpDate))
```
rossabaker: Note for future thought: Spray found it worthwhile to cache this header for up to a second.  I'm skeptical it makes a big difference, but, if someone wants to squeeze every last nanosecond out...

server/src/test/scala/org/http4s/server/middleware/DateSpec.scala
```diff
@@ -9,7 +9,7 @@ import cats.effect.testing.specs2.CatsIO
 
 class DateSpec extends Http4sSpec with CatsIO {
 
-  implicit val T = IO.timer(scala.concurrent.ExecutionContext.global)
+  override implicit val timer: Timer[IO] = Http4sSpec.TestTimer
```
rossabaker: I'm surprised this is necessary at all, but I'm also pretty sure the compiler said it was and it has a better memory than me.

ChristopherDavenport: I think its because both CatsIO and Http4sSpec define implicit timers, so as a result we have 2 instances with the same priority so it fails to find either implicitly, by overriding in the trait we create a more specific instance of the implicit which takes priority over the inherited traits.

rossabaker: `Http4sSpec` should probably be reworked to be built on top of `CatsIO`, but that's a whole other chore.

PR #3050, opened 2020-01-09
Merge pull request #3050 from ChristopherDavenport/cookieJar

ChristopherDavenport:
Future Improvements

- More Complete Test Coverage
- Perhaps an internal counter for the cookiejar to clear all the expired cookies - Currently we clear on every request which is likely overkill.

### DIFF
--- a/client/src/main/scala/org/http4s/client/middleware/CookieJar.scala
+++ b/client/src/main/scala/org/http4s/client/middleware/CookieJar.scala
@@ -100,7 +100,7 @@ object CookieJar {
       for {
         now <- currentHttpDate
         out <- ref.update(
-          _.filterNot { t =>
+          _.filter { t =>
             now <= t._2.expiresAt
           }
         )
@@ -185,6 +185,7 @@ object CookieJar {
   private[middleware] def cookieAppliesToRequest[N[_]](
       r: Request[N],
       c: ResponseCookie): Boolean = {
+
     val domainApplies = c.domain.exists(s =>
       r.uri.host.forall { authority =>
         authority.renderString.contains(s)
--- a/client/src/main/scala/org/http4s/client/middleware/CookieJar.scala
+++ b/client/src/main/scala/org/http4s/client/middleware/CookieJar.scala
@@ -202,7 +202,14 @@ object CookieJar {
         authority.renderString.contains(s)
       })
     val pathApplies = c.path.forall(s => r.uri.path.contains(s))
-    domainApplies && pathApplies
+
+    val secureSatisfied = if (c.secure) {
+      r.uri.scheme.exists { scheme =>
+        scheme === Uri.Scheme.https
+      }
+    } else true
+
+    domainApplies && pathApplies && secureSatisfied
   }
 
   private[middleware] def cookiesForRequest[N[_]](
--- a/client/src/main/scala/org/http4s/client/middleware/CookieJar.scala
+++ b/client/src/main/scala/org/http4s/client/middleware/CookieJar.scala
@@ -26,11 +26,11 @@ trait CookieJar[F[_]] {
   /**
     * Add Cookie to the cookie jar
     */
-  def addCookie(c: ResponseCookie, uri: Uri): F[Unit] = 
+  def addCookie(c: ResponseCookie, uri: Uri): F[Unit] =
     addCookies(List((c, uri)))
 
   /**
-   * Like addCookie but puts several in at once
+    * Like addCookie but puts several in at once
    **/
   def addCookies[G[_]: Foldable](cookies: G[(ResponseCookie, Uri)]): F[Unit]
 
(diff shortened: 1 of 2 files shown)


client/src/main/scala/org/http4s/client/middleware/CookieJar.scala
```diff
@@ -0,0 +1,205 @@
...
+            now <= t._2.expiresAt
+          }
+        )
+      } yield out
+
+    override def evictAll: F[Unit] = ref.set(Map.empty)
+
+    override def addCookie(c: ResponseCookie, uri: Uri): F[Unit] =
+      for {
+        now <- currentHttpDate
```
rossabaker: We're going to retrieve the date once per cookie?  It's probably cheap enough, but I'd consider a private method into which the date is passed, to support both `addCookie` and `enrichRequest`.

client/src/main/scala/org/http4s/client/middleware/CookieJar.scala
```diff
@@ -0,0 +1,205 @@
...
+      } yield out
+
+    override def enrichRequest[N[_]](r: Request[N]): F[Request[N]] =
+      for {
+        cookies <- ref.get.map(_.map(_._2.cookie).toList)
+      } yield cookiesForRequest(r, cookies)
+        .foldLeft(r) { case (req, cookie) => req.addCookie(cookie) }
+  }
+
+  private[middleware] final case class CookieKey(
```
rossabaker: Is `CookieJar` not tight enough scope here?

client/src/main/scala/org/http4s/client/middleware/CookieJar.scala
```diff
@@ -0,0 +1,205 @@
...
+    def apply(
+        setAt: HttpDate,
+        expiresAt: HttpDate,
+        cookie: ResponseCookie
+    ): CookieValue = new CookieValue(setAt, expiresAt, cookie)
+  }
+
+  private[middleware] def currentHttpDate[F[_]: Clock: MonadError[?[_], Throwable]] =
+    Clock[F]
+      .monotonic(SECONDS)
```
rossabaker: Why not `realTime` here?

client/src/main/scala/org/http4s/client/middleware/CookieJar.scala
```diff
@@ -0,0 +1,205 @@
...
+  private[middleware] def responseCookieToRequestCookie(r: ResponseCookie): RequestCookie =
+    RequestCookie(r.name, r.content)
+
+  private[middleware] def cookieAppliesToRequest[N[_]](
+      r: Request[N],
+      c: ResponseCookie): Boolean = {
+
+    val domainApplies = c.domain.exists(s =>
+      r.uri.host.forall { authority =>
+        authority.renderString.contains(s)
```
rossabaker: This makes me embarrassed of our `Uri`.

client/src/main/scala/org/http4s/client/middleware/CookieJar.scala
ChristopherDavenport: Whoopsie.

client/src/main/scala/org/http4s/client/middleware/CookieJar.scala
ChristopherDavenport: Was future proofing for eventual testing I'm sure might become necessary.

client/src/main/scala/org/http4s/client/middleware/CookieJar.scala
ChristopherDavenport: The problem with methods like that, is that then users can't recreate what we've done here. But if we expose the private method then a user could get it wrong. What about if we exposed an `addCookies` which adds multiple cookies at the same time?

ChristopherDavenport: Added `addCookies`

ChristopherDavenport:
Failure unrelated I believe.

client/src/main/scala/org/http4s/client/middleware/CookieJar.scala
```diff
@@ -0,0 +1,216 @@
...
+        val value = CookieValue(httpDate, expires, newCookie)
+        m + (key -> value)
+      case None => // Ignore Cookies We Can't get a domain for
+        m
+    }
+
+  private[middleware] def responseCookieToRequestCookie(r: ResponseCookie): RequestCookie =
+    RequestCookie(r.name, r.content)
+
+  private[middleware] def cookieAppliesToRequest[N[_]](
```
hamnis: This should also consider Secure cookies vs unsecure cookies

ChristopherDavenport:
Question for our future selves: Now that this middleware is available should the `propagateCookies` that is present in the `FollowRedirect` be removed in favor of just using this middleware?

rossabaker:
That came up at the time this middleware was proposed, and I don't think we ever answered it adequately.  We don't have any mechanism for one middleware implying another, at correct priority, without duplication.  Such a mechanism would solve these problems, but this is the first case that really would benefit from such, and it's a lot of machinery.

hamnis:
> Question for our future selves: Now that this middleware is available should the `propagateCookies` that is present in the `FollowRedirect` be removed in favor of just using this middleware?

Lets not do this in 0.21.

PR #3051, opened 2020-01-09
Add Convenience Constructor for HttpDate

ChristopherDavenport:


rossabaker:
Do we really need both `current` and `now`?  I like this one that takes the clock, but I'm not so sure I like the redundancy.

rossabaker:
`now` is not RT and should probably be deprecated, or replaced with this if we want to keep the `now` name despite changing the type.

PR #3052, opened 2020-01-09
Merge pull request #3052 from ChristopherDavenport/cachingHeadersMiddleware

ChristopherDavenport:


server/src/main/scala/org/http4s/server/middleware/Caching.scala
```diff
@@ -0,0 +1,111 @@
...
+          Header("Pragma", "no-cache"),
+          HDate(now),
+          Expires(HttpDate.Epoch) // Expire at the epoch for no time confusion
+        )
+        resp.putHeaders(headers: _*)
+      }
+    }
+
+  object Helpers {
+    def defaultStatussToSetOn(s: Status): Boolean =
```
rossabaker: Typo: Statuss

server/src/main/scala/org/http4s/server/middleware/Caching.scala
```diff
@@ -0,0 +1,111 @@
...
+            )),
+          Header("Pragma", "no-cache"),
+          HDate(now),
+          Expires(HttpDate.Epoch) // Expire at the epoch for no time confusion
+        )
+        resp.putHeaders(headers: _*)
+      }
+    }
+
+  object Helpers {
```
rossabaker: Does this need to be public?  If it does, I might consider moving the two helpers herein to the top-level.  Things nested in `Helpers` or `Utils` or `bits` tend to not be discoverable.

server/src/main/scala/org/http4s/server/middleware/Caching.scala
```diff
@@ -0,0 +1,111 @@
...
+      isPublic: Either[CacheDirective.public.type, CacheDirective.`private`],
+      methodToSetOn: Method => Boolean,
+      statusToSetOn: Status => Boolean,
+      http: Http[G, F]
+  ): Http[G, F] = {
+    val actualLifetime = lifetime match {
+      case finite: FiniteDuration => finite
+      case _ => 315360000.seconds // 10 years
+      // Http1 caches do not respect max-age headers, so to work globally it is recommended
+      // to explicitly set an Expire which requires some time interval to work
```
rossabaker: Nice comment.  I was getting ready to throw a fit about `Duration` about 20 lines above.

server/src/main/scala/org/http4s/server/middleware/Caching.scala
```diff
@@ -0,0 +1,111 @@
...
+    def defaultMethodsToSetOn(m: Method): Boolean = methodsToSetOn.contains(m)
+
+    private lazy val methodsToSetOn: Set[Method] = Set(
+      Method.GET,
+      Method.HEAD
+    )
+  }
+
+  def publicCache[G[_]: MonadError[*[_], Throwable]: Clock, F[_]](
+      lifetime: Duration,
```
rossabaker: I feel like we more commonly put the `Http` alone in a parameter list, for partial application.  But I got that feeling a long time ago, before we had all the `G` and what not.  I'm not sure whether that's still a useful convention.

server/src/main/scala/org/http4s/server/middleware/Date.scala
```diff
@@ -7,28 +7,21 @@ import cats.effect._
...
         resp <- k(a)
         header <- resp.headers
           .get(HDate)
           .fold(
-            Clock[G]
-              .realTime(MILLISECONDS)
-              .map(Instant.ofEpochMilli(_))
-              .flatMap(nowInstant => HttpDate.fromInstant(nowInstant).liftTo[G])
-              .map(nowHttpDate => HDate(nowHttpDate))
+            HttpDate.current[G].map(HDate(_))
```
rossabaker: To our future selves: we've usually resisted putting convenience constructors directly on headers, but things like this make me wonder.

server/src/main/scala/org/http4s/server/middleware/Caching.scala
ChristopherDavenport: The thing I was hoping to achieve was to allow those who used `caching` to have a way to get the tools we use rather than recreate them, but only leave the middlewares exposed at the top level.

server/src/main/scala/org/http4s/server/middleware/Caching.scala
ChristopherDavenport: We could put it as the first argument, but if we put it in the second like we generally do, then G and F cannot infer

server/src/main/scala/org/http4s/server/middleware/Caching.scala
rossabaker: Okay, fine with me as is.

server/src/main/scala/org/http4s/server/middleware/Caching.scala
rossabaker: Drat. I think you're right.  That's sad, because I like to think of these as partially applied up to the `Http`.

ChristopherDavenport: Alternative is building it as partially applied class. So that we can say `Duration ->  (forall g, f . ((MonadError Throwable g), Clock g) => http g f)`

rossabaker: Ooh, that's an interesting idea, if the type inference still works well enough when folding them down into one.

PR #3053, opened 2020-01-09
Deprecate HttpDate.now

ChristopherDavenport:
Following Feedback from after #3051 was merged, deprecate `HttpDate.now` and instruct to use `current` instead.


ChristopherDavenport:
Other alternative: rename current -> now

Or rename now -> unsafeNow, current -> now

rossabaker:
Need to clean up the deprecation warnings.  I don't think we have much effectful header generation to this point, so it's going to introduce some maps and flatMaps.

I'm good with deprecating `now`.  `current` has some JVM familarity (`System.currentTimeMillis`), and the deprecation will let people upgrade at their own pace and explain what needs to be done.

PR #3054, opened 2020-01-10
Merge pull request #3054 from bfdes/http4s-3044

### LINKED ISSUE #3044: SameSite cookie support
Since chrome is using SameSite: Lax as default from Chrome 80, we should add this to ResponseCookie to get this right for users.

https://web.dev/samesite-cookies-explained/

bfdes:
Fixes #3044. 

Please suggest where I should add additional test cases.

According to the web.dev [article](https://web.dev/samesite-cookies-explained/) linked in the issue:

* `SameSite=Lax` by default
* Must not set `SameSite=None` without `Secure` flag

### DIFF
--- a/core/src/main/scala/org/http4s/ResponseCookie.scala
+++ b/core/src/main/scala/org/http4s/ResponseCookie.scala
@@ -98,6 +98,7 @@ final case class ResponseCookie(
     maxAge: Option[Long] = None,
     domain: Option[String] = None,
     path: Option[String] = None,
+    sameSite: SameSite = SameSite.Lax,
     secure: Boolean = false,
     httpOnly: Boolean = false,
     extension: Option[String] = None
@@ -113,7 +114,8 @@ final case class ResponseCookie(
     maxAge.foreach(writer.append("; Max-Age=").append(_))
     domain.foreach(writer.append("; Domain=").append(_))
     path.foreach(writer.append("; Path=").append(_))
-    if (secure) writer.append("; Secure")
+    writer.append("; SameSite").append(sameSite)
+    if (secure || sameSite == SameSite.None) writer.append("; Secure")
     if (httpOnly) writer.append("; HttpOnly")
     extension.foreach(writer.append("; ").append(_))
     writer
--- a/core/src/main/scala/org/http4s/ResponseCookie.scala
+++ b/core/src/main/scala/org/http4s/ResponseCookie.scala
@@ -114,7 +114,7 @@ final case class ResponseCookie(
     maxAge.foreach(writer.append("; Max-Age=").append(_))
     domain.foreach(writer.append("; Domain=").append(_))
     path.foreach(writer.append("; Path=").append(_))
-    writer.append("; SameSite").append(sameSite)
+    writer.append("; SameSite=").append(sameSite)
     if (secure || sameSite == SameSite.None) writer.append("; Secure")
     if (httpOnly) writer.append("; HttpOnly")
     extension.foreach(writer.append("; ").append(_))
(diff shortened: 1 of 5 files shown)


bfdes:
I've fixed the stack-overflow problem -- it was caused by infinite recursion in a `render` method. 

The Scala 2.13 job is still failing though, albeit in an inconsistent manner.  It passed for the build that fixed the recursion problem but failed for the subsequent commits, which were only formatting changes.

rossabaker:
Test failure is the standard irritating flake, so I restarted.

core/src/main/scala/org/http4s/ResponseCookie.scala
```diff
@@ -113,7 +114,8 @@ final case class ResponseCookie(
     maxAge.foreach(writer.append("; Max-Age=").append(_))
     domain.foreach(writer.append("; Domain=").append(_))
     path.foreach(writer.append("; Path=").append(_))
-    if (secure) writer.append("; Secure")
+    writer.append("; SameSite").append(sameSite)
```
rossabaker: Does this render the `=` somehow?

core/src/main/scala/org/http4s/ResponseCookie.scala
```diff
@@ -98,6 +98,7 @@ final case class ResponseCookie(
     maxAge: Option[Long] = None,
     domain: Option[String] = None,
     path: Option[String] = None,
+    sameSite: SameSite = SameSite.Lax,
```
rossabaker: I wonder if this should be an `Option`, so we can round trip values, or continue to send cookies without this attribute.  Semantically, `Lax` is the default, but by keeping it optional, it would remain possible to send cookies without this attribute.

core/src/main/scala/org/http4s/ResponseCookie.scala
bfdes: No, that was an oversight on my part. Addressed in 6df6614a9245b2fde1b1af1d394202f23506979a.

core/src/main/scala/org/http4s/ResponseCookie.scala
bfdes: By round trip do you mean that the cookie should be returned unchanged if nothing is done? In which case would I also need to remove this change: `if (secure || sameSite == SameSite.None) writer.append("; Secure")`.

rossabaker: Ooh, yeah.  That's unfortunate, because as is, we're trying to do the right thing by default here.  And generally we make sure that we can go [model -> string -> model](https://github.com/http4s/http4s/blob/136ab44c439bec7c532518080fc619dfb7682f8a/laws/src/main/scala/org/http4s/laws/HttpCodecLaws.scala#L11), which is different from string -> model -> string.  This is probably better how you wrote it.

server/src/main/scala/org/http4s/server/middleware/CSRF.scala
```diff
@@ -367,6 +368,7 @@ object CSRF {
       httpOnly: Boolean,
       domain: Option[String] = None,
       path: Option[String] = None,
+      sameSite: SameSite = SameSite.Lax,
```
bfdes: @rossabaker @hamnis do you mind double checking if this is appropriate? 
I wasn't sure, because the middleware is supposed to be a complementary mechanism for combating CSRF.

core/src/main/scala/org/http4s/ResponseCookie.scala
bfdes: If this behaviour is desired do you think it is worth writing an (explicit) test for it?

server/src/main/scala/org/http4s/server/middleware/CSRF.scala
hamnis: SameSite helps with CSRF, so I do think this is fine.

PR #3055, opened 2020-01-10
Ember Upgrade onto fs2.io.tls

ChristopherDavenport:


ember-client/src/main/scala/org/http4s/ember/client/EmberClientBuilder.scala
```diff
@@ -7,19 +7,21 @@ import org.http4s.client._
 import cats._
 import cats.implicits._
 import cats.effect._
-import javax.net.ssl.SSLContext
-import scala.concurrent.ExecutionContext
+// import javax.net.ssl.SSLContext
+// import scala.concurrent.ExecutionContext
```
rossabaker: Remove

ChristopherDavenport:
Yes, its currently on the snapshot that is unreleased, however its a major improvement IMO.

rossabaker:
I jumped on this to see if it's ready while I'm preparing releases. All credit for the important work to Chris.

rossabaker:
Still a compile error:

```scala
[error] /home/ross/src/http4s/ember-client/src/main/scala/org/http4s/ember/client/EmberClientBuilder.scala:86:95: type mismatch;
[error]  found   : F[Option[fs2.io.tls.TLSContext]]
[error]  required: F[Option[F[fs2.io.tls.TLSContext]]]
[error]           .fold(Sync[F].delay(TLSContext.system(blocker)).attempt.map(_.toOption))(_.some.pure[F])
[error]                                                                                               ^
```

ChristopherDavenport:
This was changed afterwards due to comments from this PR, now I think the toOption stuff is now unecessary.

ChristopherDavenport:
This is analagous to what we do in blaze. 

https://github.com/http4s/http4s/blob/master/blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala#L304

rossabaker:
Oh, right, to delay the construction for those pathological cases where the default throws.  Ugh.  Yeah, good call.

PR #3056, opened 2020-01-10
Adding a simple Stream as JSON array circe encoder

BalmungSan:
This adds some functions to create a simple encoder for Streams, which encodes it as a single JSON array. The specs show how they work.

I also added the option to specify which charset to use when encoding a JSON to all the previous functions, since the method has a default argument, I guess it is binary compatible?
I am actually pretty bad at that. If the change is dangerous I may remove it.

I am open to any feedback you may have about these changes.
_(btw, I tried to follow the style conventions that were already there)_.

### DIFF
--- a/circe/src/main/scala/org/http4s/circe/CirceInstances.scala
+++ b/circe/src/main/scala/org/http4s/circe/CirceInstances.scala
@@ -110,11 +110,6 @@ trait CirceInstances extends JawnInstances {
   implicit def streamJsonArrayEncoder[F[_]: Applicative]: EntityEncoder[F, Stream[F, Json]] =
     streamJsonArrayEncoderWithPrinter(defaultPrinter)
 
-  // Constant byte chunks.
-  private final val openBrace: Chunk[Byte] = Chunk.singleton('['.toByte)
-  private final val closeBrace: Chunk[Byte] = Chunk.singleton(']'.toByte)
-  private final val comma: Chunk[Byte] = Chunk.singleton(','.toByte)
-
   /** An [[EntityEncoder]] for a [[Stream]] of JSONs, which will encode it as a single JSON array. */
   def streamJsonArrayEncoderWithPrinter[F[_]: Applicative](
       printer: Printer): EntityEncoder[F, Stream[F, Json]] =
@@ -122,7 +117,7 @@ trait CirceInstances extends JawnInstances {
       .streamEncoder[F, Chunk[Byte]]
       .contramap[Stream[F, Json]] { stream =>
         val jsons = stream.map(fromJsonToChunk(printer))
-        Stream.emit(openBrace) ++ jsons.intersperse(comma) ++ Stream.emit(closeBrace)
+        CirceInstances.openBrace ++ jsons.intersperse(CirceInstances.comma) ++ CirceInstances.closeBrace
       }
       .withContentType(`Content-Type`(MediaType.application.json))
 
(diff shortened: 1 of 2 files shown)


BalmungSan:
As a side note, I have read somewhere that mixing **Context Bounds** with normal implicit parameters is not recommended. However, my changes do that, but just because the rest of the file also does it, and I tried to be consistent.
But, IMHO, it would be good to rewrite all that to just use implicit parameters.

BalmungSan:
It seems a test failed, but as far as I can tell, the test that failed has nothing to do with my change.
Also, I am pretty sure all test passed in my first commit _(I amened some scalafmt issues)_, so it seems the problem is non-deterministic.
Did you already know about this problem?

bfdes:
I think I have a similar problem in #3054.

circe/src/main/scala/org/http4s/circe/CirceInstances.scala
```diff
@@ -92,20 +92,52 @@ trait CirceInstances extends JawnInstances {
   implicit def jsonEncoder[F[_]: Applicative]: EntityEncoder[F, Json] =
     jsonEncoderWithPrinter(defaultPrinter)
 
-  def jsonEncoderWithPrinter[F[_]: Applicative](printer: Printer): EntityEncoder[F, Json] =
+  private def fromJsonToChunk(printer: Printer, cs: Charset)(json: Json): Chunk[Byte] =
```
rossabaker: JSON _must_ be encoded as UTF-8, UTF-16, or UTF-32.  And anything but UTF-8 is generally a bad idea: https://tools.ietf.org/html/rfc7159#section-8.1.  I'm not sure we want to parameterize this, and if we do, perhaps as a sealed type that permits only one of the three JSON encodings.

circe/src/main/scala/org/http4s/circe/CirceInstances.scala
```diff
@@ -92,20 +92,52 @@ trait CirceInstances extends JawnInstances {
...
+
+  implicit def streamJsonArrayEncoder[F[_]: Applicative]: EntityEncoder[F, Stream[F, Json]] =
+    streamJsonArrayEncoderWithPrinter(defaultPrinter)
+
+  /** An [[EntityEncoder]] for a [[Stream]] of JSONs, which will encode it as a single JSON array. */
+  def streamJsonArrayEncoderWithPrinter[F[_]: Applicative](
+      printer: Printer,
+      cs: Charset = DefaultCharset): EntityEncoder[F, Stream[F, Json]] = {
+    def fromStr(str: String): Chunk[Byte] =
+      Chunk.byteBuffer(cs.nioCharset.encode(str))
```
rossabaker: Might be able to micro-optimize this by having constant chunks for the delimiters, but I doubt it makes a huge difference.  It would be tractable if the `cs` domain were reduced to three, or one.

circe/src/main/scala/org/http4s/circe/CirceInstances.scala
BalmungSan: Didn't know about that, I guess the best would be to just remove that change and leave the default.

circe/src/main/scala/org/http4s/circe/CirceInstances.scala
BalmungSan: Given I will be removing the charset and just leave the default one, I can do this.

BalmungSan:
> I agree that mixing context bounds and implicits is bad (I wish there were a scala flag for this), and would support fixing those as well.

Would you prefer that on this PR, or in a following PR with only aesthetic changes?
Also, I am not sure if that mix happens elsewhere? Maybe it would be worth creating a new issue to track all files with such mixture and all PRs fixing those.

rossabaker:
I'd bet it does happen elsewhere, but I don't know how to find them.  I'm sure there's some scalameta wizardry that would help.

A separate PR sounds good for fixing this file.

BalmungSan:
@rossabaker I just pushed the change removing the use of **Charset** and adding the constant chunks.

Let me know if you have any other suggestions.

BalmungSan:
@rossabaker I moved the constant byte chunks into the companion object, so they aren't instantiated for each instance. And I also cached the creation of the open and close brace streams.

rossabaker:
Looks great.  The test failure is unrelated, which I'm trying to mitigate in #3058, because I don't think it describes a real issue.

PR #3057, opened 2020-01-12
Fix conflicting circe on plugin path

rossabaker:
http4s-0.20 is built on circe-0.11.  So is bloop, I think: this made it load for me on metals-0.7.6 with Emacs.


BalmungSan:
@rossabaker I can confirm that with this change metals works on vscode.

PR #3058, opened 2020-01-12
Skip reset request timeout test on Travis CI

rossabaker:
I always hate to do this, but this one is persistently irritating and I think not teaching us anything.


rossabaker:
Lately it's often this one, and usually on 2.13.  Maybe it _is_ telling us something?

ChristopherDavenport:
We should put it in an issue and then merge this.

PR #3060, opened 2020-01-12
Cleaning the CirceInstance implicits

BalmungSan:
This PR applies to changes to the CirceInstances file.

The first one, replaces the combinations of **Contex Bounds** with **implicit parameters**.
The change leaves all methods with only one of them; bounds, only when neither of the implicits was explicitly required, and implicits otherwise.

The second one was to replace the two **implicit classes** with **one value class** and one **implicit conversion**.

### DIFF
--- a/circe/src/main/scala/org/http4s/circe/CirceInstances.scala
+++ b/circe/src/main/scala/org/http4s/circe/CirceInstances.scala
@@ -59,11 +59,12 @@ trait CirceInstances extends JawnInstances {
       }
     }
 
-  def jsonOf[F[_]: Sync, A](implicit decoder: Decoder[A]): EntityDecoder[F, A] =
+  def jsonOf[F[_]: Sync, A: Decoder]: EntityDecoder[F, A] =
     jsonOfWithMedia(MediaType.application.json)
 
-  def jsonOfWithMedia[F[_]: Sync, A](r1: MediaRange, rs: MediaRange*)(
-      implicit decoder: Decoder[A]): EntityDecoder[F, A] =
+  def jsonOfWithMedia[F[_], A](r1: MediaRange, rs: MediaRange*)(
+      implicit F: Sync[F],
+      decoder: Decoder[A]): EntityDecoder[F, A] =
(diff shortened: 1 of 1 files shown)


circe/src/main/scala/org/http4s/circe/CirceInstances.scala
```diff
@@ -138,15 +139,8 @@ trait CirceInstances extends JawnInstances {
...
-  implicit class JsonDecoderSyntax[F[_]](private val req: Message[F])(implicit F: JsonDecoder[F]) {
-    def asJson: F[Json] = F.asJson(req)
-    def asJsonDecode[A: Decoder]: F[A] = F.asJsonDecode(req)
-  }
-
-  implicit class MessageSyntax[F[_]: Sync](self: Message[F]) {
-    def decodeJson[A](implicit decoder: Decoder[A]): F[A] =
-      self.as(implicitly, jsonOf[F, A])
-  }
+  implicit final def toMessageSynax[F[_]](req: Message[F]): CirceInstances.MessageSyntax[F] =
```
rossabaker: Can these functions be called outside the circe package when the syntax class is `private[circe]`?  I thought Scala emitted a warning here about something escaping into a more visible scope.

BalmungSan: @rossabaker Scala access is somewhat [_"tricky"_](https://github.com/scala/bug/issues/6794).

I believe this should just work, it should be similar to when one uses the **partially applied trick** and the returned class is `private`. I had done that on [neotypes](https://github.com/neotypes/neotypes/blob/master/core/src/main/scala/neotypes/DeferredQuery.scala#L307), and it works without problems.

Also, I just tested this on a console _(inside bloop)_.

```scala
def test[F[_]: cats.effect.Sync](m: org.http4s.Message[F]): F[io.circe.Json] = m.json
// defined function test
```

Correct me if I am wrong, but that was enough proof that the code compiled and AFAIK the Ammonite RELP is outside any package.

-----

However, if you prefer, I could make it public.

rossabaker: I'm convinced that it works, even if I am not convinced that it should.

PR #3062, opened 2020-01-13
Update Jawn and Circe

travisbrown:
We're working on finalizing Jawn 1.0.0 and this is a WIP PR that updates to the latest Jawn RC and Circe milestones. It also depends on [this unpublished PR](https://github.com/http4s/jawn-fs2/pull/86).

This probably shouldn't be merged until we release Jawn 1.0.0 (which is likely to happen in the next week), but I needed to update http4s for some local testing and thought I should go ahead and open this.

### DIFF
--- a/argonaut/src/main/scala/org/http4s/argonaut/Parser.scala
+++ b/argonaut/src/main/scala/org/http4s/argonaut/Parser.scala
@@ -12,23 +12,23 @@ import scala.collection.mutable
 /* Temporary parser until jawn-argonaut supports 6.2.x. */
 private[argonaut] object Parser extends SupportParser[Json] {
   implicit val facade: Facade[Json] =
-    new Facade[Json] {
-      override def jnull() = Json.jNull
-      override def jfalse() = Json.jFalse
-      override def jtrue() = Json.jTrue
+    new Facade.NoIndexFacade[Json] {
+      override def jnull: Json = Json.jNull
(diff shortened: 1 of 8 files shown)


rossabaker:
After [discussing on Gitter](https://gitter.im/http4s/http4s?at=5e232f52ad195a0f672e012e), we'd like to release 0.21 on jawn-1.0 and circe-0.13, so added to the milestone.  We think most jawn use is transitive through circe and/or http4s, and the circe migration itself should not be onerous.

If anybody _really_ needs circe-0.12 support, they are welcome to use our last milestone, or we can publish an http4s-jawn_0.14 and http4s-circe_0.12 in separate repos, but I'll expect volunteers.

This will require removing argonaut until there is a jawn-1.0 facade.  It could be brought back with either an internal facade or a new argonaut release.  My preference is for the latter.  (My long term preference is to remove it from this repo and leave it to the community to support.)

rossabaker:
I'm exploiting our time zone difference and bumped the newly released dependencies to see what else it uncovers.

rossabaker:
Wow.  I overlooked that you'd already patched the internal argonaut parser.  We still should spin it off, but my other comments are void.

The tests in dea77a0 are older and could all be written better, but I'm trying to keep this PR focused.

travisbrown:
Thanks, @rossabaker! Just removed the `[WIP]`, and am happy to publish Circe 0.13.0 whenever.

Issue #3059, opened 2020-01-12
Facilitate composition on ContextMiddleware


isaias-b:
To improve composability on `ContextMiddleware` and friends it would be nice to have:

* `Functor` instances for `ContextRequest` and `ContextMiddleware`. 
* Add `contramapRoutes` operator on `ContextMiddlware`, that allows for more involved mapping where errors might occur.
* Add `withoutContext` operator on `ContextMiddlware`, that basically ignores the additional context and turns the `ContextMiddleware` into a vanilla `Middleware`.

These additions would help to transform already existing `Middleware` instances more easily. Since some applications don't expose user aware routes and gathering the user context might lead to failures.

Here is a proposal how these could look like:
```
import cats.data.OptionT
import cats.{FlatMap, Functor}
import org.http4s._
import org.http4s.server._
import cats.implicits._

object Http4sAdditions {
  implicit def contextRequestFunctor[F[_]]: Functor[ContextRequest[F, *]] =
    new Functor[ContextRequest[F, *]] {
      override def map[A, B](fa: ContextRequest[F, A])(
          f: A => B): ContextRequest[F, B] =
        AuthedRequest(f(fa.context), fa.req)
    }

  implicit def contextMiddlewareFunctor[F[_]]
    : Functor[ContextMiddleware[F, *]] = new Functor[ContextMiddleware[F, *]] {
    override def map[A, B](fa: ContextMiddleware[F, A])(
        f: A => B
    ): ContextMiddleware[F, B] = (routesB: ContextRoutes[B, F]) => {
      val routesA: ContextRoutes[A, F] =
        routesB.local[ContextRequest[F, A]](_.map(f))
      fa(routesA)
    }
  }

  implicit class ContextMiddlewareOps[F[_], A](cm: ContextMiddleware[F, A]) {
    def contramapRoutes[B](f: ContextRoutes[B, F] => ContextRoutes[A, F])
      : ContextMiddleware[F, B] =
      (routesB: AuthedRoutes[B, F]) => cm(f(routesB))

    def withoutContext(implicit A: FlatMap[F]): Middleware[
      OptionT[F, *],
      Request[F],
      Response[F],
      Request[F],
      Response[F]
    ] = (routesIn: HttpRoutes[F]) => {
      val contextRoutes: ContextRoutes[A, F] =
        routesIn.local[ContextRequest[F, A]](_.req)
      cm(contextRoutes)
    }
  }
}
```

hamnis:
[...]

PR #3070, opened 2020-01-14
Merge pull request #3070 from ChristopherDavenport/contextRequestFunctor

ChristopherDavenport:
Fixes part of #3059 

Add a functor instance for `ContextRequest`

### DIFF
--- a/core/src/main/scala/org/http4s/ContextRequest.scala
+++ b/core/src/main/scala/org/http4s/ContextRequest.scala
@@ -1,7 +1,7 @@
 package org.http4s
 
-import cats.syntax.functor._
-import cats.{Applicative, Eval, Functor, Traverse, ~>}
+import cats._
+import cats.implicits._
 import cats.data.Kleisli
 
 final case class ContextRequest[F[_], A](context: A, req: Request[F]) {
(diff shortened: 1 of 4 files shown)


core/src/main/scala/org/http4s/ContextRequest.scala
```diff
@@ -17,4 +17,26 @@ object ContextRequest {
   def apply[F[_]: Functor, T](
       getContext: Request[F] => F[T]): Kleisli[F, Request[F], ContextRequest[F, T]] =
     Kleisli(request => getContext(request).map(ctx => ContextRequest(ctx, request)))
+
+  implicit def contextRequestInstances[F[_]]: NonEmptyTraverse[ContextRequest[F, *]] =
+    new NonEmptyTraverse[ContextRequest[F, *]] {
+      // Members declared in cats.Foldable
```
rossabaker: These comments seem like something tooling can tell us. I don't think they add much.

ChristopherDavenport: They were auto-generated by metals and I was too lazy to remove them. Will do.

ChristopherDavenport:
Oh goodness, looks like 2.13 can infer but 2.12 can't... Investigating.

PR #3078, opened 2020-01-18
  Wrap uninformative Java exceptions in ConnectionFailure

rossabaker:
See http4s/blaze#373.

Java's `UnresolvableAddressException` does not carry any information about the host that failed to resolve.  If a client fails to connect, we should provide context on which host we failed to connect to.  We do so by introducing a wrapper exception, `ConnectionFailure`, in the client package.

Other clients should probably use this if they can.   Fixed for blaze, as it was reported on blaze.

### DIFF
--- a/blaze-client/src/main/scala/org/http4s/client/blaze/Http1Support.scala
+++ b/blaze-client/src/main/scala/org/http4s/client/blaze/Http1Support.scala
@@ -2,12 +2,11 @@ package org.http4s
 package client
 package blaze
 
+import cats.effect._
+import cats.implicits._
 import java.net.InetSocketAddress
 import java.nio.ByteBuffer
 import java.nio.channels.AsynchronousChannelGroup
-
-import cats.effect._
-import cats.implicits._
 import javax.net.ssl.SSLContext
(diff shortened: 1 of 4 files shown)


tomasherman:
awesome, thanks!

tomasherman:
is there any chance of having this backported into the old build? Or is there an ETA on next version of http4s going stable?

rossabaker:
I'm working on 0.21-0-RC1 right now.  It does subtly change the behavior, so I'm a little reluctant to backport it into 0.20, though it's difficult to imagine anyone is depending on catching the Java exception.  I could be talked into it, because I'm also preparing an 0.20 backport release right now.

tomasherman:
If it's not too much trouble, it would be helpful for us. BTW just a
thought... Would it make sense to extend the original exception instead of
ioexception to make this backwards compatible?

rossabaker:
There are many Java exceptions that can be thrown here, and we can't dynamically mix a ConnectionFailure into all of them. But we could handle this one and throw a subtype that overrides the message in 0.20, and stick with the new wrapper in 0.21.

PR #3079, opened 2020-01-18
Release v0.21.0-RC1

rossabaker:


rossabaker:
I think we're ready after https://github.com/http4s/http4s/milestone/34.

PR #3080, opened 2020-01-18
Release v0.20.16

rossabaker:
Release v0.20.16 with eligible backports from master and compatible dependency bumps.

My expectation is that this is the final 0.20.x release, unless someone sends us a tale of woe.


rossabaker:
I continue to be confused by when we need scala-reflect and in what scope and what sbt-explicit-dependencies expects of us on this matter.  What did we do that made this start failing?

rossabaker:
Got reckless with the force push: I had already been merging this into #3079. Put it back the way it was.

client/src/test/scala/org/http4s/client/PoolManagerSpec.scala
```diff
@@ -70,6 +72,28 @@ class PoolManagerSpec(name: String) extends Http4sSpec {
...
+        conn <- pool.borrow(key)
+        waiting1 <- pool.borrow(key).start
+        waiting2 <- pool.borrow(key).start
+        _ <- IO.sleep(timeout + 20.milliseconds)
+        waiting3 <- pool.borrow(key).start
+        _ <- pool.release(conn.connection)
+        result1 <- waiting1.join.void.attempt
+        result2 <- waiting2.join.void.attempt
+        result3 <- waiting3.join.void.attempt
+      } yield (result1, result2, result3)).unsafeRunTimed(10.seconds) must beSome.like {
```
rossabaker: This, and the rest of this suite, should do better in Travis CI when we have cats-effect-testing, but this branch is cats-effect-1, so I'm just bumping the timeout higher to try to survive.

PR #3081, opened 2020-01-19
Add Config case classes

sideeffffect:
Add configuration case classes that is suitable for PureConfig

(credit goes to @jakubjanecek, the original author of [`scala-server-toolkit`](https://github.com/avast/scala-server-toolkit/) where this approach is used extensively)


rossabaker:
As mentioned, these builders are hostile to compatibility.  I'd like to support pure-config, but would prefer to explore whether we can do it without case classes.

PR #3082, opened 2020-01-19
Keep RequestKey in blaze-client UnresolvedAddressException

rossabaker:
Follow-up from #3078.  We prefer the approach in #3078, but this is a more conservative approach for the maintenance branch.

/cc @tomasherman


client/src/main/scala/org/http4s/client/ConnectionFailure.scala
```diff
@@ -0,0 +1,11 @@
+package org.http4s.client
+
+import java.io.IOException
+
+/** Indicates a failure to establish a client connection, preserving the request key
+  * that we tried to connect to.
+  */
+class ConnectionFailure(requestKey: RequestKey, cause: Throwable) extends IOException(cause) {
```
hamnis: This class is not used. Drop?

tomasherman: I think the idea is that it will be used in next release of http4s, but @rossabaker chose to use backward-compatible solution for `0.20` release

rossabaker:
Yes, I started this by cherry picking from master and failed to clean up. It will be in 0.21, but need not be here.

Issue #3087, opened 2020-01-20
Unhelpful timeout exception


tomasherman:
Hello, i found another fairly unhelpful exception coming from http4s:

```
java.util.concurrent.TimeoutException: Request timeout after 10000 ms
	at org.http4s.client.blaze.BlazeClient$$anon$1.run(BlazeClient.scala:152)
	at monix.execution.schedulers.TracingRunnable.run(TracingRunnable.scala:33)
	at monix.execution.internal.InterceptRunnable.run(InterceptRunnable.scala:27)
	at monix.execution.internal.forkJoin.AdaptedForkJoinTask.exec(AdaptedForkJoinTask.scala:27)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:290)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1020)
	at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1656)
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1594)
	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:177)
```

It would be great if we could add some context - like request that failed and maybe even IP address of the upstream that timeout out.

tomasherman:
I was hoping to help out by providing a PR but i have a hard time to find all the pieces ... i think what needs to be done is to add information about IP address to `BlazeConnection`? 

Seems like this would be a good place to put that information in: https://github.com/http4s/http4s/blob/master/blaze-client/src/main/scala/org/http4s/client/blaze/Http1Support.scala#L64 ?

tomasherman:
So there is a bunch of exceptions in `BlazeClient` ... i wonder if it would make sense to add to all of them at least the `RequestKey` ... thoughts?

rossabaker:
Yes, you could enhance the exceptions by pattern matching [Lefts]( https://github.com/http4s/http4s/blob/efa36675a8dd06aee80f5bb908899cbcbc8ce442/blaze-client/src/main/scala/org/http4s/client/blaze/Http1Support.scala#L72).  It will be the same basic tradeoffs as on the other: using the standard Java exception vs. a richer wrapper, with some opportunities to extend the standard Java exceptions like we did in 0.20.

rossabaker:
[...]

PR #3089, opened 2020-01-21
Change exception to a more generic name and wrap all clients exceptio…

tomasherman:
…ns inside it


tomasherman:
@rossabaker so this is my attempt at fixing this ... not sure if you're gonna like it but it does make sense to me. The idea is to wrap any exception that comes from http4s client inside `ClientException` with a useful context information (request key, socket address and whether it was resovled) - it should show `host:port` if it's not resolved and `ipv4/6 address:port` when it is resolved. And that way all excpetions are covered and user still can match on underlying excpetion if need be.

What do you think?

client/src/main/scala/org/http4s/client/ClientFailure.scala
```diff
@@ -0,0 +1,12 @@
+package org.http4s.client
+
+import java.io.IOException
+import java.net.InetSocketAddress
+
+/** Indicates a failure in the client when executing a request preserving the context (requestKey and upstream address)
+  */
+class ClientFailure(requestKey: RequestKey, upstream: InetSocketAddress, cause: Throwable)
```
rossabaker: Maybe we should make `requestKey` and `upstream` vals in case anyone wants to do something structurally.

I don't like making exceptions case classes, because that implies equality if the arguments are equal, and I tend not to think two exceptions are equal if their arguments are.  But we might even consider an `unapply` on a companion object.

blaze-client/src/main/scala/org/http4s/client/blaze/Http1Support.scala
```diff
@@ -69,9 +69,9 @@ final private class Http1Support[F[_]](
                 t
               }
             case Left(e) =>
-              Future.failed(e)
+              Future.failed(new ClientFailure(requestKey, addr, e))
```
rossabaker: Anything that happens here is still a failure to connect, isn't it?  I'm not sure about the rename, though if we wanted to introduce wrapping elsewhere, we could have a `ConnectionFailure extends ClientFailure`.

tomasherman: i think those exceptions might be more generic ... for example main trigger for this was timeout exception, which is not really connection failure - it's failure to process request in due time

client/src/main/scala/org/http4s/client/ClientFailure.scala
tomasherman: makes sense, i will do that later today

blaze-client/src/main/scala/org/http4s/client/blaze/Http1Support.scala
rossabaker: It wasn't a timeout in establishing a connection?  I'd be surprised to see a request processing timeout here.

tomasherman: ah, then i missed the point of what i was trying to achieve and ConnectionFailure is indeed more descriptive ... but i think this PR is still useful ... however i was trying to cover exception mentioned in #3087 which i think is about request processing

rossabaker: Yes, I think your enhancements to the exception are still valuable.  I commented on #3087 for what I think is the `TimeoutException` you're trying to clean up.

client/src/main/scala/org/http4s/client/ConnectionFailure.scala
```diff
@@ -1,11 +1,21 @@
...
 import java.io.IOException
+import java.net.InetSocketAddress
 
 /** Indicates a failure to establish a client connection, preserving the request key
   * that we tried to connect to.
   */
-class ConnectionFailure(requestKey: RequestKey, cause: Throwable) extends IOException(cause) {
+class ConnectionFailure(
+    val requestKey: RequestKey,
+    val upstream: InetSocketAddress,
```
rossabaker: @ChristopherDavenport questioned on Gitter whether we'd always have this in other backends where we might wish to wrap exceptions.  We might want to make this an `Option` for portability, though I haven't tried to thread this into other backends yet.

rossabaker:
Restarted the flaky, unrelated test.

client/src/main/scala/org/http4s/client/ConnectionFailure.scala
tomasherman: up to you, im ok with it being optional

tomasherman:
cheers

PR #3092, opened 2020-01-21
Add server side ssl support for ember

ChristopherDavenport:
Built on top of #3055


ember-client/src/main/scala/org/http4s/ember/client/EmberClientBuilder.scala
```diff
@@ -7,19 +7,19 @@ import org.http4s.client._
...
 import org.http4s.client._
 import fs2.io.tcp.SocketGroup
 import fs2.io.tcp.SocketOptionMapping
+import fs2.io.tls._
 import scala.concurrent.duration.Duration
 
 final class EmberClientBuilder[F[_]: Concurrent: Timer: ContextShift] private (
-    private val sslContextOpt: Option[(ExecutionContext, SSLContext)],
-    private val sgR: Resource[F, SocketGroup],
+    private val blockerOpt: Option[Blocker],
```
ChristopherDavenport: I can factor this out presently, as its only used to construct a SocketGroup if one is not provided, but I find it more likely that one would prefer to just apply a Blocker, and having it for the future is likely more useful for binary compatibility concerns as a lot of the fs2 operations require them. Just nothing I'm using at the moment.

PR #3097, opened 2020-01-21
Hypothetical TLSParameters Approach

ChristopherDavenport:


ember-client/src/main/scala/org/http4s/ember/client/EmberClient.scala
```diff
@@ -4,10 +4,15 @@ import cats.effect._
...
+import io.chrisdavenport.vault._
+import fs2.io.tls._
 
 final class EmberClient[F[_]: Bracket[?[_], Throwable]] private[client] (
     private val client: Client[F],
-    private val pool: KeyPool[F, RequestKey, (RequestKeySocket[F], F[Unit])]
+    private val pool: KeyPool[
+      F,
+      (RequestKey, Option[TLSParameters]),
+      (RequestKeySocket[F], F[Unit])]
```
rossabaker: I'd think about naming these tuples as case classes, even if they're private.

rossabaker: Which, looking at `state`, they can't be.

ember-client/src/main/scala/org/http4s/ember/client/EmberClient.scala
```diff
@@ -17,7 +22,11 @@ final class EmberClient[F[_]: Bracket[?[_], Throwable]] private[client] (
...
     * is a mapping between the number of connections in the pool for each requestKey.
     */
-  def state: F[(Int, Map[RequestKey, Int])] = pool.state
+  def state: F[(Int, Map[(RequestKey, Option[TLSParameters]), Int])] = pool.state
 
   def run(req: Request[F]): Resource[F, Response[F]] = client.run(req)
 }
+
+object EmberClient {
+  val TLSParameters: Key[TLSParameters] = Key.newKey[IO, TLSParameters].unsafeRunSync
```
rossabaker: At risk of being accused of Hungarian Notation, I'd put this in a `Keys` object or name it `Key`.

PR #3101, opened 2020-01-23
Tests for Request query param convenience methods

### LINKED ISSUE #3090: Add spec for withQueryParams and withMultiValueQueryParams
Would like to get #3024 in before the release, but we're taking on spec debt to do it.

bfdes:
Fixes #3090.

### DIFF
--- a/tests/src/test/scala/org/http4s/UriSpec.scala
+++ b/tests/src/test/scala/org/http4s/UriSpec.scala
@@ -632,7 +632,7 @@ http://example.org/a file
       val u = Uri().withQueryParams(params)
       u must be_==(Uri(query = Query.fromString("param1=1&param2=2")))
     }
-    "add of multiple values for same query parameter name" in {
+    "add multiple values for same query parameter name" in {
       val params = Map("param1" -> List(1), "param2" -> List(2, 3))
       val u = Uri().withMultiValueQueryParams(params)
       u must be_==(Uri(query = Query.fromString("param1=1&param2=2&param2=3")))
--- a/tests/src/test/scala/org/http4s/UriSpec.scala
+++ b/tests/src/test/scala/org/http4s/UriSpec.scala
@@ -637,6 +637,17 @@ http://example.org/a file
       val u = Uri().withMultiValueQueryParams(params)
       u must be_==(Uri(query = Query.fromString("param1=1&param2=2&param2=3")))
     }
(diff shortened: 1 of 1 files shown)


tests/src/test/scala/org/http4s/UriSpec.scala
```diff
@@ -615,7 +615,7 @@ http://example.org/a file
       val u = Uri() +*? (Ttl(2))
       u must be_==(Uri(query = Query.fromString(s"ttl=2")))
     }
-    "Work with queryParam" in {
+    "add a QueryParam instance" in {
```
bfdes: I thought this was more descriptive.

tests/src/test/scala/org/http4s/UriSpec.scala
```diff
@@ -627,6 +627,16 @@ http://example.org/a file
...
     }
+    "add multiple query parameters at once" in {
+      val params = Map("param1" -> 1, "param2" -> 2)
+      val u = Uri().withQueryParams(params)
+      u must be_==(Uri(query = Query.fromString("param1=1&param2=2")))
+    }
+    "add multiple values for same query parameter name" in {
+      val params = Map("param1" -> List(1), "param2" -> List(2, 3))
+      val u = Uri().withMultiValueQueryParams(params)
+      u must be_==(Uri(query = Query.fromString("param1=1&param2=2&param2=3")))
```
bfdes: Looks like this is the way http4s handles multi-value params. 
Another convention seems to be `param1=1&param2=2,3`.

rossabaker: The comma as a separator is arbitrary.  What if your data has a comma?  There's no single way to do it.  There's also convention in some dynamically typed languages to flag multi-values with an array syntax: `param[]=1,param[]=2`.  None of these conventions are precluded by http4s, but the repeated key has a fairly straightforward interpretation.

bfdes:
I've added a couple more tests to make sure that existing parameters are replaced. 

I think the "self-overriding" behaviour described in the docstring is a property of the `Map` implementation and `Map.apply`, rather than the convenience methods.

rossabaker:
But the `Map` is an implementation detail, so I think it's good to specify that behavior.  But yes, knowing how it works, we should be quite confident that it does.

Issue #3102, opened 2020-01-23
Cannot have more than one pending write request


rossabaker:
Reported on [Gitter](https://gitter.im/http4s/http4s?at=5e280a23364db33faa10e026) by @khrupalyk.

Potential fix [here](https://github.com/khrupalyk/http4s/commit/1cff33791445e2b3b2c1375de7e0114db7e147de).

rossabaker:
Fixed by #3105.

PR #3103, opened 2020-01-23
Fix pending write failure in blaze web socket

rossabaker:
See #3102.

### DIFF
--- a/blaze-core/src/test/scala/org/http4s/blazecore/websocket/Http4sWSStageSpec.scala
+++ b/blaze-core/src/test/scala/org/http4s/blazecore/websocket/Http4sWSStageSpec.scala
@@ -93,5 +93,18 @@ class Http4sWSStageSpec extends Http4sSpec {
       _ <- socket.pollOutbound().map(_ must_=== None)
       _ <- socket.sendInbound(Close())
     } yield ok).unsafeRunSync()
+
+    "not fail on pending write request" in (for {
+      socket <- TestWebsocketStage()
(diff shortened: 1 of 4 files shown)


blaze-core/src/test/scala/org/http4s/blazecore/websocket/Http4sWSStageSpec.scala
```diff
@@ -93,5 +95,18 @@ class Http4sWSStageSpec extends Http4sSpec {
...
+      socket <- TestWebsocketStage()
+      pings = 100L
+      in = Stream.eval(socket.sendInbound(Ping())).repeat.take(pings)
+      out = Stream.eval(socket.sendWSOutbound(Text("."))).repeat.take(200)
+      _ <- in.merge(out).compile.drain
+      _ <- socket.sendInbound(Close())
+      _ <- socket
+        .pollBatchOutputbound(500)
+        .map(_.count(_.isInstanceOf[Pong]))
+        .map(pongs => pongs must_== pings)
```
rossabaker: The test passes most of the time.

Without @khrupalyk's patch, we get an exception in the logs and a small number of pongs.  With the patch, we never get the exception, but sometimes drop a few pongs.  We could make the test reliable by looking for, say, 80 pongs, but that doesn't really describe our intent.

Sending a Close and asserting the reason for the Close response may be a more robust way, but I need to get some sleep.

rossabaker:
Superseded by #3105.

PR #3105, opened 2020-01-23
Write to blaze websockets with a permit

### LINKED ISSUE #3102: Cannot have more than one pending write request

rossabaker:
Fixes #3102.  In addition to pongs outside the normal stream, we can also send a close frame, so it seemed simplest just to guard the write in a semaphore.

/cc @khrupalyk

### DIFF
--- a/blaze-core/src/main/scala/org/http4s/blazecore/websocket/Http4sWSStage.scala
+++ b/blaze-core/src/main/scala/org/http4s/blazecore/websocket/Http4sWSStage.scala
@@ -3,6 +3,7 @@ package blazecore
 package websocket
 
 import cats.effect._
+import cats.effect.concurrent.Semaphore
 import cats.implicits._
 import fs2._
 import fs2.concurrent.SignallingRef
@@ -22,6 +23,8 @@ private[http4s] class Http4sWSStage[F[_]](
     deadSignal: SignallingRef[F, Boolean]
 )(implicit F: ConcurrentEffect[F], val ec: ExecutionContext)
     extends TailStage[WebSocketFrame] {
+
+  private[this] val writeSemaphore = F.toIO(Semaphore[F](1L)).unsafeRunSync()
 
   def name: String = "Http4s WebSocket Stage"
 
@@ -44,12 +47,12 @@ private[http4s] class Http4sWSStage[F[_]](
   }
 
   private[this] def writeFrame(frame: WebSocketFrame, ec: ExecutionContext): F[Unit] =
-    F.async[Unit] { cb =>
+    writeSemaphore.withPermit(F.async[Unit] { cb =>
       channelWrite(frame).onComplete {
         case Success(res) => cb(Right(res))
         case Failure(t) => cb(Left(t))
       }(ec)
-    }
+    })
 
   private[this] def readFrameTrampoline: F[WebSocketFrame] = F.async[WebSocketFrame] { cb =>
     channelRead().onComplete {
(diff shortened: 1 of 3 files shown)


blaze-core/src/main/scala/org/http4s/blazecore/websocket/Http4sWSStage.scala
```diff
@@ -16,10 +17,11 @@ import org.http4s.websocket.WebSocketFrame._
 import scala.concurrent.ExecutionContext
 import scala.util.{Failure, Success}
 
-private[http4s] class Http4sWSStage[F[_]](
+private[http4s] class Http4sWSStage[F[_]] private (
     ws: WebSocket[F],
     sentClose: AtomicBoolean,
-    deadSignal: SignallingRef[F, Boolean]
+    deadSignal: SignallingRef[F, Boolean],
+    writeSemaphore: Semaphore[F]
```
rossabaker: The change could be more minimal if we'd construct this semaphore in the constructor with a `.toIO.unsafeRunSync()`, an ugliness that was necessary anyway in blaze-server.

The constructor is privatized to avoid passing in a semaphore of more than one permit.

rossabaker: That broke bincompat. It's a private[http4s], and we could get away with an exception, but those have burned us before.  Going with a more conservative approach.

rossabaker: This broke bincompat.  It's `private[http4s]`, so we could get away with an exception, but I've been burned too many times by that.  1dc9664 is a more conservative approach.

rossabaker:
This is an easy enough bugfix, I don't mind cutting another 0.20 for it.

rossabaker:
This test has failed twice in a row on Travis CI in Scala 2.11 and I don't know if it's real or if it's just because I added to the weight of blocking tests for lack of cats-effect-testing in this branch.

rossabaker:
Okay, it appears I've fixed the relevant failure.

blaze-core/src/main/scala/org/http4s/blazecore/websocket/Http4sWSStage.scala
```diff
@@ -23,6 +24,8 @@ private[http4s] class Http4sWSStage[F[_]](
 )(implicit F: ConcurrentEffect[F], val ec: ExecutionContext)
     extends TailStage[WebSocketFrame] {
 
+  private[this] val writeSemaphore = F.toIO(Semaphore[F](1L)).unsafeRunSync()
```
hamnis: I'm guessing this is being created and run here to make sure you only have 1 permit?

ChristopherDavenport: This is problematic but solves the problem. :cry:

ChristopherDavenport: Hate breaking RT

rossabaker: Yes. I looked at passing it in, but we'd lose the guarantee that it's a single permit, and it would have required a MiMa exception that I didn't want to add.

rossabaker: I don't like it, but `new Http4sWSStage(...)` is already not RT, and this value can't be observed outside the instance.  This is at the far end of town where the grickle grass grows and the wind smells slow and sour when it blows.

PR #3106, opened 2020-01-23
Interrupt Body of DefaultHead Middleware

ChristopherDavenport:
So if we are processing a GET as a HEAD, we know we do not care about the body, but we still want to ensure finalizers are run. To get the finalizers we must run the original stream, but we do not want to process all the bytes, in fact any of the bytes. 

To this end we immediately interrupt the stream and make sure that is empty. 

Requires the addition of a `Concurrent` restraint on the body, however I think that is worth it for the potentially gigabytes of processing this saves the server.

### DIFF
--- a/server/src/test/scala/org/http4s/server/middleware/DefaultHeadSpec.scala
+++ b/server/src/test/scala/org/http4s/server/middleware/DefaultHeadSpec.scala
@@ -48,7 +48,7 @@ class DefaultHeadSpec extends Http4sSpec with CatsIO {
         cleanedUpRef <- Ref[IO].of(false)
         route = HttpRoutes.of[IO] {
           case GET -> _ =>
-            val body: EntityBody[IO] = Stream.empty.onFinalizeWeak(cleanedUpRef.set(true))
+            val body: EntityBody[IO] = Stream.never.onFinalizeWeak(cleanedUpRef.set(true))
             Ok(body)
         }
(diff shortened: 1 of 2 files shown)


rossabaker:
`allow GET body to clean up on fallthrough` is a naughty test: it's running a side effect in the response body.  If that side effect were a body finalizer, the test would pass reliably.

server/src/test/scala/org/http4s/server/middleware/DefaultHeadSpec.scala
```diff
@@ -39,14 +44,18 @@ class DefaultHeadSpec extends Http4sSpec {
...
-          val body: EntityBody[IO] = eval_(IO({ cleanedUp = true }))
-          Ok(body)
-      }).orNotFound
-      app(Request[IO](Method.HEAD)).flatMap(_.as[String]).unsafeRunSync()
-      cleanedUp must beTrue
+      for {
+        cleanedUpRef <- Ref[IO].of(false)
+        route = HttpRoutes.of[IO] {
+          case GET -> _ =>
+            val body: EntityBody[IO] = Stream.empty.onFinalizeWeak(cleanedUpRef.set(true))
```
rossabaker: We might even be able to test the interruption with `Stream.never.onFinalizeWeak(...)`.

rossabaker:
Just needs a little help with the type inference and I think it's good to go.

PR #3107, opened 2020-01-23
Covary `F` in Message types. 

zarthross:
Adding a `covary` function to `Message` and its subtypes that matches the `Stream.covary` function.

It's implemented as a type cast to avoid the costs of creating a new `Request` object for something that should be safe to do as a cast.

Alternative implementation without a type cast, (but including an extra allocation of `Request` is [here](https://github.com/zarthross/http4s/commit/290831ad41ccb6ef1777cecabb9e2a583f09a5b8)

### DIFF
--- a/core/src/main/scala/org/http4s/Media.scala
+++ b/core/src/main/scala/org/http4s/Media.scala
@@ -10,6 +10,7 @@ import org.http4s.util.decode
 trait Media[F[_]] {
   def body: EntityBody[F]
   def headers: Headers
+  def covary[F2[x] >: F[x]]: Media[F2]
 
   final def bodyAsText(implicit defaultCharset: Charset = DefaultCharset): Stream[F, String] =
     charset.getOrElse(defaultCharset) match {
@@ -58,5 +59,7 @@ object Media {
     def body = b
 
     def headers: Headers = h
+
+    override def covary[F2[x] >: F[x]]: Media[F2] = this.asInstanceOf[Media[F2]]
   }
 }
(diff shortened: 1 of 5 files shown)


rossabaker:
Out of curiosity, do you need this for rho, or other reasons?

zarthross:
I'll write up some spec's later, wanted to get buy in before putting in the effort.

I didn't need it for Rho.

rossabaker:
Good call, @hamnis.  It would thread through `Part` then, too.

diesalbla:
Rather than adding a `covary` method, is it feasible to make the type parameter  covariant?  
```Scala
sealed trait Message[+F[_]]
```

rossabaker:
> Rather than adding a `covary` method, can we just make the type parameter covariant?
> 
> ```scala
> sealed trait Message[+F[_]]
> ```

Maybe. It would probably want to start in `Media`.  I expect some operations that take typeclass constraints would have to move to something like fs2's `InvariantOps` (e.g., `as`, `attemptAs`).  I don't know what other fallout there would be.  Do you want to poke at it and see how sweeping the change is?

diesalbla:
Sure. Do you mind keeping this PR on hold for a little while, until I have tried those changes?

rossabaker:
We're already in an RC cycle, with hopes of the release next week that many people have been awaiting for too long.  My inclination is to proceed with @zarthross' approach in 0.21, and consider the covariance in the next release.

zarthross:
Per feedback, I've added `covary` up to `Media` so its now on `Part` as well, and I've added some specs that are hopefully sufficient.

PR #3110, opened 2020-01-24
MessageFailure: simplify toHttpResponse/inHttpResponse

IndiscriminateCoding:
`Applicative` constraints on `MessageFailure.inHttpResponse` looks redundant because one can't do anything useful with them (given polymorphic context).

### DIFF
--- a/server/src/main/scala/org/http4s/server/package.scala
+++ b/server/src/main/scala/org/http4s/server/package.scala
@@ -134,13 +134,12 @@ package object server {
     inDefaultServiceErrorHandler[F, F]
 
   def inDefaultServiceErrorHandler[F[_], G[_]](
-      implicit F: Monad[F],
-      G: Applicative[G]): Request[G] => PartialFunction[Throwable, F[Response[G]]] = req => {
+      implicit F: Monad[F]): Request[G] => PartialFunction[Throwable, F[Response[G]]] = req => {
     case mf: MessageFailure =>
       messageFailureLogger.debug(mf)(
         s"""Message failure handling request: ${req.method} ${req.pathInfo} from ${req.remoteAddr
           .getOrElse("<unknown>")}""")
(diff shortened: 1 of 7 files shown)


IndiscriminateCoding:
Not sure why it fails on jdk8/scala-2.13

rossabaker:
That failure is unrelated.  It's one I'm hoping to fix with more consistent use of cats-effect-testing.  I restarted the build.

IndiscriminateCoding:
@ChristopherDavenport the issue is, it's impossible to do RT-logging using `Applicative` alone (it requires `Sync` as the very least).

server/src/main/scala/org/http4s/server/package.scala
```diff
@@ -134,13 +134,12 @@ package object server {
     inDefaultServiceErrorHandler[F, F]
 
   def inDefaultServiceErrorHandler[F[_], G[_]](
-      implicit F: Monad[F],
-      G: Applicative[G]): Request[G] => PartialFunction[Throwable, F[Response[G]]] = req => {
+      implicit F: Monad[F]): Request[G] => PartialFunction[Throwable, F[Response[G]]] = req => {
     case mf: MessageFailure =>
       messageFailureLogger.debug(mf)(
         s"""Message failure handling request: ${req.method} ${req.pathInfo} from ${req.remoteAddr
```
rossabaker: To @ChristopherDavenport's concern, the relevant logging I see is here, and in `JsonDebugger`.  In both these places, we're returning in `F` and could compose an effect-tracked log with the response.  But I don't see any logging where we're creating these responses, so the effect is more power than we want or need.  Maybe I'm missing something, but I _think_ both of you can be made happy?

rossabaker: Excuse me.  This one is missing `Sync`.

PR #3113, opened 2020-01-25
Http4s 3095 Pretty-print an org.http4s.Request as a curl command

erewl:
Initial Implementation for the curl-representation of the request.
Maybe  --data-body for POST-Requests should be included still.  What do you think?

Also I kept it that stringified version of the Request probably should redact sensitive information (Authorization- and Cookie-Headers). Is that fine?

### DIFF
--- a/core/src/main/scala/org/http4s/Message.scala
+++ b/core/src/main/scala/org/http4s/Message.scala
@@ -267,6 +267,21 @@ final class Request[F[_]](
 
   def queryString: String = uri.query.renderString
 
+  def asCurl: String = {
+    val elements = List(
+      s"-X ${method.name}",
+      s"'${uri.renderString}'",
+      headers
+        .redactSensitive()
+        .toList
+        .map { header =>
+          s"-H '${header.toString}'"
+        }
+        .mkString(" ")
+    )
+    s"curl ${elements.filter(_.nonEmpty).mkString(" ")}"
+  }
+
   /**
     * Representation of the query string as a map
     *
--- a/core/src/main/scala/org/http4s/Message.scala
+++ b/core/src/main/scala/org/http4s/Message.scala
@@ -12,6 +12,7 @@ import java.net.{InetAddress, InetSocketAddress}
 import org.http4s.headers._
 import org.log4s.getLogger
 import _root_.io.chrisdavenport.vault._
+import org.http4s.util.CaseInsensitiveString
 
 import scala.util.hashing.MurmurHash3
 
@@ -271,18 +272,23 @@ final class Request[F[_]](
     *
     * Supported cURL-Parameters are: -X, -H
     *
-    * @param redactHeaders to determine whether sensitive information inheaders should be redacted, true per default
     */
-  def asCurl(redactHeaders: Boolean = true): String = {
-    val filteredHeaders =
-      if (redactHeaders) headers.redactSensitive(Headers.SensitiveHeaders.contains)
-      else headers
+  def asCurl(
+      redactHeadersWhen: CaseInsensitiveString => Boolean = Headers.SensitiveHeaders.contains)
+      : String = {
+    /**
+      * escapes characters that are used in the curl-command, such as '
+      */
+    def escapeQuotationMarks(s: String) = s.replaceAll("'", "'\\''")
+
     val elements = List(
       s"-X ${method.name}",
-      s"'${uri.renderString}'",
-      filteredHeaders.toList
+      s"'${escapeQuotationMarks(uri.renderString)}'",
+      headers
+        .redactSensitive(redactHeadersWhen)
+        .toList
         .map { header =>
-          s"-H '${header.toString}'"
+          s"-H '${escapeQuotationMarks(header.toString)}'"
         }
         .mkString(" ")
     )
--- a/core/src/main/scala/org/http4s/Message.scala
+++ b/core/src/main/scala/org/http4s/Message.scala
@@ -267,10 +267,12 @@ final class Request[F[_]](
 
   def queryString: String = uri.query.renderString
 
-  /**
-    * creates a cURL-command from the request, supported parameters are
-    * -X to specify the HTTP-method
-    * -H for headers, redacts sensitive information such as Authorization- or Cookie-Headers
+  /** cURL representation of the request.
+    *
+    * Supported Parameters are:
+    *
+    * -X,
+    * -H (redacts sensitive information such as Authorization- or Cookie-Headers)
     */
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/http4s/Message.scala
```diff
@@ -267,6 +267,28 @@ final class Request[F[_]](
...
+    */
+  def asCurl: String = {
+    val elements = List(
+      s"-X ${method.name}",
+      s"'${uri.renderString}'",
+      headers
+        .redactSensitive()
+        .toList
+        .map { header =>
+          s"-H '${header.toString}'"
```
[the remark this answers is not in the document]
erewl: oh you mean regular quotation marks, like " ?

core/src/main/scala/org/http4s/Message.scala
```diff
@@ -267,6 +267,28 @@ final class Request[F[_]](
...
   def queryString: String = uri.query.renderString
 
+  /** cURL representation of the request.
+    *
+    * Supported Parameters are:
+    *
+    * -X,
+    * -H (redacts sensitive information such as Authorization- or Cookie-Headers)
+    */
+  def asCurl: String = {
```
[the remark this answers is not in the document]
erewl: I have worked in a parameter for this.
You meant something like that?
Or should it get more into detail whether to log headers after all and when to redact them?

core/src/main/scala/org/http4s/Message.scala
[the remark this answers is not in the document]
rossabaker: You would need to replace every `'` in the value with with `'\''`: the first character closes it, the second character escapes it, and the third character reopens it.  I _think_ that would be sufficient.

---

If someone comes along and tells us that's wrong, Emacs' restclient takes a more paranoid approach.  Instead of quoting, it escapes anything but a handful of known safe characters:

```http
GET https://httpbin.org/get
X-escape-test: "Don't forget to escape!"
```

yields

```sh
curl -i -H X-escape-test\:\ \"Don\'t\ forget\ to\ escape\!\" -XGET https\://httpbin.org/get
```

The relevant code that does it is:

```emacs-lisp
      ;; Quote everything except POSIX filename characters.
      ;; This should be safe enough even for really weird shells.
      (replace-regexp-in-string
       "\n" "'\n'"
       (replace-regexp-in-string "[^-0-9a-zA-Z_./\n]" "\\\\\\&" argument))))
```

core/src/main/scala/org/http4s/Message.scala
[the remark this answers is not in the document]
rossabaker: The `Logger` middleware takes a [`CaseInsensitiveString => Boolean`](https://github.com/http4s/http4s/blob/28945bbb4ba0c4c9122c2bac3af2e3333f2bfb64/server/src/main/scala/org/http4s/server/middleware/Logger.scala#L25).  It would be great to stay consistent with that, since it's the same basic problem.

erewl:
> I think that's a nice method, but if that sounds like too much, we can consider that in a separate PR.

Yeah, a separate PR would be good for this I think. Still getting to know the ecosystem behind http4s

core/src/main/scala/org/http4s/Message.scala
```diff
@@ -267,6 +268,33 @@ final class Request[F[_]](
...
+    * Supported cURL-Parameters are: -X, -H
+    *
+    */
+  def asCurl(
+      redactHeadersWhen: CaseInsensitiveString => Boolean = Headers.SensitiveHeaders.contains)
+      : String = {
+    /**
+      * escapes characters that are used in the curl-command, such as '
+      */
+    def escapeQuotationMarks(s: String) = s.replaceAll("'", "'\\''")
```
rossabaker: Close! Scala's string escaping reduces the `\\` to a `\`, and `replaceAll` escapes the second `'` with the remaining `\`.

```
scala> "'".replaceAll("'", "'\\''")
res0: String = '''
```

You need to get a `\\` all the way to the regex.  You can add a second layer of escaping:

```
scala> "'".replaceAll("'", "'\\\\''")
res2: String = '\''
```

Or use triple quotes so each backslash gets to `replaceAll`:

```
scala> "'".replaceAll("'", """'\\''""")
res1: String = '\''
```

erewl: Ahhh. Okay, already got lost in the backslash-quote jungle. Will rework that in a few then!
Thanks for clarifying again :-)

rossabaker:
Oops, there's still a merge conflict. It's trivial: a test was added at the same place in `MessageSpec`.  Just keep them both.

core/src/main/scala/org/http4s/Message.scala
```diff
@@ -273,6 +274,34 @@ final class Request[F[_]](
...
+  /** cURL representation of the request.
+    *
+    * Supported cURL-Parameters are: -X, -H
+    *
+    */
+  def asCurl(
+      redactHeadersWhen: CaseInsensitiveString => Boolean = Headers.SensitiveHeaders.contains)
+      : String = {
+
+    /**
```
rossabaker: I think [this build failure](https://travis-ci.org/http4s/http4s/jobs/642122692#L354) is because it doesn't like scaladoc in an internal method. If the build fails again, a single-asterisk here should clean that up.

erewl: ah, great! was already confused why doc-build was failing, will do, when next build fails (probably it will)

rossabaker:
:tada: Thanks for persevering on this. We'll merge as soon as a second maintainer approves.

erewl:
Whoohooo!
Thanks for the guidance as well!

core/src/main/scala/org/http4s/Message.scala
```diff
@@ -273,6 +274,34 @@ final class Request[F[_]](
...
+    *
+    */
+  def asCurl(
+      redactHeadersWhen: CaseInsensitiveString => Boolean = Headers.SensitiveHeaders.contains)
+      : String = {
+
+    /*
+     * escapes characters that are used in the curl-command, such as '
+     */
+    def escapeQuotationMarks(s: String) = s.replaceAll("'", """'\\''""")
```
aeons: This adds a set of single quotation marks around the escaped one. Is that on purpose?

```scala
scala> escapeQuotationMarks("Can't do this?")
res2: String = Can'\''t do this?
```

erewl:
@rossabaker you don't squash commits when merging or did i do something different/wrong than your usual workflow? I see the commits from my branch on the master now.

PR #3115, opened 2020-01-25
Fix flaky decode

### LINKED ISSUE #2956: Flaky property: decode
```
[info] decode should
[error]   x be consistent with utf8Decode (540 ms)
[error]    Falsified after 67 passed tests.
[error]    > ARG_0: "﻿\u0008"
[error]    > ARG_0_ORIGINAL: "﻿魂Ⲷਬ纅ᗀ赕Ὴᔱ鑰跎㎯쾼ዦ㷭"
[error]    > ARG_1: 1
[error]    > ARG_1_ORIGINAL: 1
[error]    > '' != ' (DecodeSpec.scala:24)
```

mgibowski:
Fixes #2956

Test was failing when the generated string contained UTF-8 BOM character (that's why the minimum example looked like an empty String).

Last year fs2 started ignoring UTF-8 BOM characters:
https://github.com/functional-streams-for-scala/fs2/issues/1484

So flakiness of this test is a successful discovery of the difference between the behavior of fs2 and http4s decoders.

Following the example of fs2, I made changes to the `decode` function dropping BOM characters and added an example test to verify that.

### DIFF
--- a/core/src/main/scala/org/http4s/util/package.scala
+++ b/core/src/main/scala/org/http4s/util/package.scala
@@ -7,6 +7,8 @@ import java.nio.charset.StandardCharsets
 import scala.concurrent.ExecutionContextExecutor
 
 package object util {
+  private val utf8Bom: Chunk[Byte] = Chunk(0xef.toByte, 0xbb.toByte, 0xbf.toByte)
+
   def decode[F[_]](charset: Charset): Pipe[F, Byte, String] = {
     val decoder = charset.nioCharset.newDecoder
     val maxCharsPerByte = math.ceil(decoder.maxCharsPerByte().toDouble).toInt
@@ -24,7 +26,8 @@ package object util {
           else Pull.output1(outputString).as(None)
         case Some((chunk, stream)) =>
           if (chunk.nonEmpty) {
-            val bytes = chunk.toArray
+            val chunkWithoutBom = skipByteOrderMark(chunk)
+            val bytes = chunkWithoutBom.toArray
             val byteBuffer = ByteBuffer.wrap(bytes)
             val charBuffer = CharBuffer.allocate(bytes.length * maxCharsPerByte)
             decoder.decode(byteBuffer, charBuffer, false)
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/http4s/util/package.scala
```diff
@@ -24,7 +26,8 @@ package object util {
           else Pull.output1(outputString).as(None)
         case Some((chunk, stream)) =>
           if (chunk.nonEmpty) {
-            val bytes = chunk.toArray
+            val chunkWithoutBom = skipByteOrderMark(chunk)
```
rossabaker: This is probably fast enough as is, but do we only need to check this on the first chunk?  And is it incorrect if we drop it in the middle of a stream?

tests/src/test/scala/org/http4s/util/DecodeSpec.scala
```diff
@@ -55,5 +55,11 @@ class DecodeSpec extends Http4sSpec {
         decoded must_== expected
       }
     }
+
+    "drop Byte Order Mark" in {
```
rossabaker: Probably exceedingly rare, but what if the BOM spans chunks?

rossabaker:
As far as I know, this isn't a bug that has bitten any users in practice, but the change to the code goes beyond the Testing label.

rossabaker:
Do we want to merge this as is and follow up with another PR for the corner case of the corner case?

ChristopherDavenport:
Yup.

core/src/main/scala/org/http4s/util/package.scala
mgibowski: Yes, checking only the first chunk is enough.

According to [wikipedia](https://en.wikipedia.org/wiki/Byte_order_mark#UTF-8):
> its only use in UTF-8 is to signal at the start that the text stream is encoded in UTF-8, or that it was converted to UTF-8 from a stream that contained an optional BOM.

So in both implementations (here and in fs2) if UTF-8 BOM is not at the very beginning of the stream - it will not be dropped and will remain in the result.

So we could optimize it possibly introducing some var to mark the first chunk... But hopefully that is not necessary.

tests/src/test/scala/org/http4s/util/DecodeSpec.scala
mgibowski: Given, that we exclude cases of it appearing in the middle of the stream, I don't think it would ever happen as we use `unconsN`. It might be more likely in fs2 implementation, where they use `uncons1`.

mgibowski: Having written this it became clear to me that it is actually unnecessary to check `chunk.size >= 3`.
I took it from fs2, but because of `unconsN` it is redundant in our implementation.

PR #3117, opened 2020-01-26
Asynchronous tests

rossabaker:
Specs2 has a blocking execution model. `IOMatchers` and `CatsIO` both call blocking methods under the covers.  I believe this blocking is likely a cause of the flakiness that plagues our Travis CI.

This introduces an `Http4sAsyncSpec`, built on cats-effect-testing-scalatest, itself built on Scalatest's asynchronous testing.  The underlying framework accepts a `Future[Assertion]`, and the intermediaries let us not think about `Future`.

This is a radical change, because it introduces another major testing framework unless and until we migrate all the tests.  Then again, testing `IO` is a large part of what we do, and our flaky tests are highly discouraging.

Discuss.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -592,9 +592,11 @@ lazy val commonSettings = Seq(
   ),
   libraryDependencies ++= Seq(
     catsEffectTestingSpecs2,
+    catsEffectTestingScalatest,
     catsLaws,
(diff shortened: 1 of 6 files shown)


rossabaker:
I don't want to go down this road yet.

PR #3121, opened 2020-01-26
Deprecate http4s-testing public interface

rossabaker:
* Few people use http4s-testing's published artifacts
* Some of the functionality is in specs2-cats
* We should not be publishing things that don't relate to HTTP

Removing `Http4sMatchers` is a heavy lift, so an internal `Http4sLegacyMatchersIO` is introduced.  This should never be mixed with `cats.effect.testing.specs2.CatsEffect`, so both have been removed from `Http4sSpec`.  The few suites that used both were rewritten to use one.  (They both block threads, so we'd prefer to use neither in the long run.)

`RunTimedMatchers` and `IOMatchers` live in specs2-cats.  We use theirs since we deprecated ours.

`ErrorReportingUtils` is now just an internal `ErrorReporting` object, which we should be making much better use of.


build.sbt
```diff
@@ -107,6 +107,7 @@ lazy val testing = libraryProject("testing")
     description := "Instances and laws for testing http4s code",
     libraryDependencies ++= Seq(
       catsEffectLaws,
+      specs2Cats,
```
rossabaker: This is bad, because it pulls in specs2-core.  specs2-matcher is lighter weight and works in other libraries.

rossabaker:
It's hard to do this right without deprecating the entire public interface of `Http4sMatchers`.  Several of those block threads, so that might be a good idea anyway.

PR #3124, opened 2020-01-28
Change Effect to Sync in FileService

erdeszt:
The extra features provided by `Effect` are not used and relaxing the constraints makes it easier to use it with `ZIO` instead of `IO`.

Edit: I'm a bit confused about the failing spec. Seems unrelated to my change but I don't have much experience with the test suite so I could be wrong.

### DIFF
--- a/server/src/main/scala/org/http4s/server/staticcontent/NoopCacheStrategy.scala
+++ b/server/src/main/scala/org/http4s/server/staticcontent/NoopCacheStrategy.scala
@@ -6,7 +6,7 @@ import cats.effect._
 
 /** Cache strategy that doesn't cache anything, ever. */
 class NoopCacheStrategy[F[_]] extends CacheStrategy[F] {
-  override def cache(uriPath: String, resp: Response[F])(implicit F: Effect[F]): F[Response[F]] =
+  override def cache(uriPath: String, resp: Response[F])(implicit F: Sync[F]): F[Response[F]] =
     F.pure(resp)
 }
 
--- a/server/src/main/scala/org/http4s/server/staticcontent/CacheStrategy.scala
+++ b/server/src/main/scala/org/http4s/server/staticcontent/CacheStrategy.scala
@@ -2,7 +2,7 @@ package org.http4s
 package server
 package staticcontent
 
-import cats.effect.Effect
+import cats.effect.Sync
 
 /** Cache the body of a [[Response]] for future use
   *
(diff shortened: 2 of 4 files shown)


server/src/main/scala/org/http4s/server/staticcontent/NoopCacheStrategy.scala
```diff
@@ -6,7 +6,7 @@ import cats.effect._
 
 /** Cache strategy that doesn't cache anything, ever. */
 class NoopCacheStrategy[F[_]] extends CacheStrategy[F] {
-  override def cache(uriPath: String, resp: Response[F])(implicit F: Effect[F]): F[Response[F]] =
+  override def cache(uriPath: String, resp: Response[F])(implicit F: Sync[F]): F[Response[F]] =
```
rossabaker: Isn't this one just `Applicative`?

rossabaker: Ah, no, because of the trait.  Perhaps we could move this constraint to a constructor for the strategy.

erdeszt: You mean remove it from `CacheStrategy.cache`? Should I do it in this PR?

rossabaker: If it's on `CacheStrategy.cache`, it has to be on all the implementations.  I'm thinking of moving the constraints to the `MemoryCache` and `NoopCacheStrategy` constructors.  `MemoryCache` would require `Sync`, and `NoopCacheStrategy` would be fine with just `Applicative`.

... at least that's how it compiles in my head.

erdeszt: I've started on it but ran into an issue because `NoopCacheStrategy` is used as a default argument in `FileService.Config.apply`(https://github.com/http4s/http4s/blob/master/server/src/main/scala/org/http4s/server/staticcontent/FileService.scala#L38)

I could make the cache strategy a non optional argument but I'm not sure if that's the right call.

rossabaker: I prefer builder patterns for configuration, because we can add more configuration options later without breaking bincompat.  But that's still ugly with regard to this problem: we would need to supply the constraint wherever the default value is materialized.

And getting the strategy to `Applicative` doesn't even do much in practice, because the service that uses it still requires `Sync`.

I think maybe we accept this as incremental progress and rethink it for 1.0.

PR #3125, opened 2020-01-29
Configure server TLS with fs2.io.tls.TLSParameters

### LINKED ISSUE #2499: Option to set TLS protocol version and cipher suites
Is there a way we can set TLS protocol and CipherSuites while creating server.
We have a requirement to use TLSv1.2 version and remove cipher suites with 128 keys.

Or Is there a way we can pass `SSLEngine` with all the required options while creating the server?

rossabaker:
* Deprecates `SSLConfig` in the server package.
* Deprecates `SSLAuthMode`, which is good, but not part of the otherwise superior `TLSParameters`.
* Adds a `withTLSParameters` method to the server builders to replace the auth mode and allow more configuration.

Fixes #2499.

### DIFF
--- a/blaze-server/src/main/scala/org/http4s/server/blaze/BlazeServerBuilder.scala
+++ b/blaze-server/src/main/scala/org/http4s/server/blaze/BlazeServerBuilder.scala
@@ -6,6 +6,7 @@ import cats.{Alternative, Applicative}
 import cats.data.Kleisli
 import cats.implicits._
 import cats.effect.{ConcurrentEffect, Resource, Timer}
+import com.github.ghik.silencer.silent
 import fs2.io.tls.TLSParameters
 import java.io.FileInputStream
 import java.net.InetSocketAddress
@@ -83,7 +84,7 @@ class BlazeServerBuilder[F[_]](
     selectorThreadFactory: ThreadFactory,
     enableWebSockets: Boolean,
     sslContext: F[Option[SSLContext]],
-    tlsParameters: TLSParameters,
+    sslEngineConfig: SSLEngine => Unit,
     isHttp2Enabled: Boolean,
     maxRequestLineLen: Int,
     maxHeadersLen: Int,
@@ -110,7 +111,7 @@ class BlazeServerBuilder[F[_]](
       selectorThreadFactory: ThreadFactory = selectorThreadFactory,
       enableWebSockets: Boolean = enableWebSockets,
       sslContext: F[Option[SSLContext]] = sslContext,
-      tlsParameters: TLSParameters = tlsParameters,
+      sslEngineConfig: SSLEngine => Unit = sslEngineConfig,
       http2Support: Boolean = isHttp2Enabled,
       maxRequestLineLen: Int = maxRequestLineLen,
       maxHeadersLen: Int = maxHeadersLen,
@@ -131,7 +132,7 @@ class BlazeServerBuilder[F[_]](
       selectorThreadFactory,
       enableWebSockets,
       sslContext,
-      tlsParameters,
+      sslEngineConfig,
       http2Support,
       maxRequestLineLen,
       maxHeadersLen,
@@ -156,6 +157,7 @@ class BlazeServerBuilder[F[_]](
     copy(maxRequestLineLen = maxRequestLineLen, maxHeadersLen = maxHeadersLen)
 
   @deprecated("Use withSSLContext and withTLSParameters instead", "0.21.0-RC3")
+  @silent("deprecated")
   def withSSL(
       keyStore: StoreInfo,
       keyManagerPassword: String,
@@ -193,7 +195,7 @@ class BlazeServerBuilder[F[_]](
       context.some
     }
 
-    copy(sslContext = sslContext, tlsParameters = clientAuthModeToTlsParameters(clientAuth))
+    copy(sslContext = sslContext, sslEngineConfig = clientAuthModeToTlsParameters(clientAuth))
   }
 
   def withSSLContext(sslContext: SSLContext): Self =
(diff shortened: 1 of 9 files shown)


blaze-server/src/main/scala/org/http4s/server/blaze/BlazeServerBuilder.scala
```diff
@@ -106,7 +109,8 @@ class BlazeServerBuilder[F[_]](
       bufferSize: Int = bufferSize,
       selectorThreadFactory: ThreadFactory = selectorThreadFactory,
       enableWebSockets: Boolean = enableWebSockets,
-      sslBits: Option[SSLConfig] = sslBits,
+      sslContext: F[Option[SSLContext]] = sslContext,
```
rossabaker: `withSSL`'s old, quirky `KeyStoreBits` is effectful in creating the SSLContext, so we need to wrap this in an effect.

blaze-server/src/main/scala/org/http4s/server/blaze/BlazeServerBuilder.scala
```diff
@@ -150,20 +155,75 @@ class BlazeServerBuilder[F[_]](
       maxHeadersLen: Int = maxHeadersLen): Self =
     copy(maxRequestLineLen = maxRequestLineLen, maxHeadersLen = maxHeadersLen)
 
+  @deprecated("Use withSSLContext and withTLSParameters instead", "0.21.0-RC3")
```
rossabaker: Translating this call is a drag (note the heft of the implementation), but we're not trying to be a general SSL configuration library.

If people really want something like this, we should put it in a function that returns an `F[(SSLContext, SSLParameters)]`, and let them flatMap it in.  Then we can get rid of the nasty effect in this builder.

blaze-server/src/main/scala/org/http4s/server/blaze/BlazeServerBuilder.scala
```diff
@@ -150,20 +155,75 @@ class BlazeServerBuilder[F[_]](
...
+      sslContext = F.pure(sslContext.some),
+      tlsParameters = clientAuthModeToTlsParameters(clientAuth))
+
+  def withoutSSLContext: Self =
+    copy(sslContext = F.pure(None))
+
+  def withTLSParameters(tlsParameters: TLSParameters): Self =
+    copy(tlsParameters = tlsParameters)
+
+  private def clientAuthModeToTlsParameters(clientAuthMode: SSLClientAuthMode) =
```
rossabaker: This will probably move into `server` before we're done.

blaze-server/src/main/scala/org/http4s/server/blaze/BlazeServerBuilder.scala
```diff
@@ -309,88 +359,57 @@ class BlazeServerBuilder[F[_]](
...
-        .map[Server[F]] { serverChannel =>
-          new Server[F] {
-            val address: InetSocketAddress =
-              serverChannel.socketAddress
-
-            val isSecure = sslBits.isDefined
-
-            override def toString: String =
-              s"BlazeServer($address)"
+    Resource.liftF(sslContext).flatMap { sslCtxOpt =>
```
rossabaker: This is mostly whitespace change we added this.

blaze-server/src/main/scala/org/http4s/server/blaze/BlazeServerBuilder.scala
```diff
@@ -309,88 +359,57 @@ class BlazeServerBuilder[F[_]](
...
+      })(factory => F.delay { factory.closeGroup() })
+
+      def mkServerChannel(factory: ServerChannelGroup): Resource[F, ServerChannel] =
+        Resource.make(F.delay {
+          val address = resolveAddress(socketAddress)
+
+          val sslConfig = sslCtxOpt.map(ctx => (ctx, toSSLParameters(tlsParameters)))
+
+          // if we have a Failure, it will be caught by the effect
+          factory.bind(address, pipelineFactory(scheduler, sslConfig)).get
```
rossabaker: This is the important change from the resource acquired above.

core/src/main/scala/org/http4s/internal/fs2.scala
```diff
@@ -0,0 +1,27 @@
...
+import _root_.fs2.io.tls.TLSParameters
+import javax.net.ssl.SSLParameters
+import org.http4s.internal.CollectionCompat.CollectionConverters._
+
+private[http4s] object fs2 {
+  def toSSLParameters(tlsParameters: TLSParameters): SSLParameters = {
+    val p = new SSLParameters()
+    import tlsParameters._
+    algorithmConstraints.foreach(p.setAlgorithmConstraints)
+    // applicationProtocols.foreach(ap => p.setApplicationProtocols(ap.toArray))
```
rossabaker: If we leave these commented, they get ignored on JDK11 despite the `TLSParameters` type implying they'll be respected.  If we support them, we have to start building on JDK11 and risk accidentally calling its methods while claiming to be a JDK8 library.

If we can get https://github.com/functional-streams-for-scala/fs2/pull/1768 released, then on fs2's heads be it.

geoffreytchew:
> /cc @geoffreytchew

Thank you

blaze-server/src/main/scala/org/http4s/server/blaze/BlazeServerBuilder.scala
```diff
@@ -309,88 +359,57 @@ class BlazeServerBuilder[F[_]](
...
+        else
+          NIO1SocketServerGroup
+            .fixedGroup(connectorPoolSize, bufferSize, channelOptions, selectorThreadFactory)
+      })(factory => F.delay { factory.closeGroup() })
+
+      def mkServerChannel(factory: ServerChannelGroup): Resource[F, ServerChannel] =
+        Resource.make(F.delay {
+          val address = resolveAddress(socketAddress)
+
+          val sslConfig = sslCtxOpt.map(ctx => (ctx, toSSLParameters(tlsParameters)))
```
rossabaker: A troublesome point with this design is that it completely clobbers the default parameters from the `SSLContext`.  We should probably make `tlsParameters` an `Option`

blaze-server/src/main/scala/org/http4s/server/blaze/BlazeServerBuilder.scala
```diff
@@ -150,20 +155,75 @@ class BlazeServerBuilder[F[_]](
...
+          .getOrElse(KeyManagerFactory.getDefaultAlgorithm))
+
+      kmf.init(ks, keyManagerPassword.toCharArray)
+
+      val context = SSLContext.getInstance(protocol)
+      context.init(kmf.getKeyManagers, tmf.orNull, null)
+      context.some
+    }
+
+    copy(sslContext = sslContext, tlsParameters = clientAuthModeToTlsParameters(clientAuth))
```
rossabaker: This `clientAuth` flag used to update the context's default parameters, but we don't have access to the context.  We need even more state to preserve this behavior.

core/src/main/scala/org/http4s/internal/fs2.scala
rossabaker: We can live with this as is, because we're expecting an fs2 release that publicizes this.

blaze-server/src/main/scala/org/http4s/server/blaze/BlazeServerBuilder.scala
```diff
@@ -106,7 +110,8 @@ class BlazeServerBuilder[F[_]](
       bufferSize: Int = bufferSize,
       selectorThreadFactory: ThreadFactory = selectorThreadFactory,
       enableWebSockets: Boolean = enableWebSockets,
-      sslBits: Option[SSLConfig] = sslBits,
+      sslContext: F[Option[SSLContext]] = sslContext,
+      sslEngineConfig: SSLEngine => Unit = sslEngineConfig,
```
rossabaker: This signature should make people ill, but the state is 100% internal, and it saves us some garbage on every request down here where the grickle grass grows.

Being a function lets us support the old API so people can transition gracefully.  This could be tightened up to `Option[SSLParameters]` in the long run.

blaze-server/src/main/scala/org/http4s/server/blaze/BlazeServerBuilder.scala
```diff
@@ -150,20 +156,84 @@ class BlazeServerBuilder[F[_]](
...
       clientAuth: SSLClientAuthMode = SSLClientAuthMode.NotRequested): Self =
-    copy(sslBits = Some(SSLContextBits(sslContext, clientAuth)))
+    copy(
+      sslContext = F.pure(sslContext.some),
+      sslEngineConfig = clientAuthModeToTlsParameters(clientAuth))
+
+  def withoutSSLContext: Self =
+    copy(sslContext = F.pure(None))
+
+  def withSSLParameters(sslParameters: SSLParameters): Self =
```
rossabaker: This is useful over `TLSParameters` because it can be created from a default `SSLParameters` for whatever `SSLContext` is on hand.  `TLSParameters` needs to be configured from _its_ defaults, or we'd need to copy an `SSLParameters` to `TLSParameters` only to copy it back.

This effectively makes `TLSParameters` the convenience method.

rossabaker:
This turned into a mess, but I learned from it. Better to come.

PR #3132, opened 2020-01-30
Support SSLParameters on blaze-server

rossabaker:
* Deprecates `withSSLContext`, which takes a `SSLClientAuthMode`, which is only one of many parameters a user may wish to configure.

* Adds `withSslContext`, which is like `withSSLContext` without the client auth flag.

* Adds `withSslContextAndParameters`, which takes an `SSLParameters`.  This can be constructed conveniently from an `fs2.io.tls.TLSParameters`, though we provide no convenience overload.

* Adds `withoutSsl` to disable it.

Addresses #2499 for blaze.  Tomcat and Jetty implementations will be their own PRs.


blaze-server/src/main/scala/org/http4s/server/blaze/BlazeServerBuilder.scala
```diff
@@ -156,14 +158,29 @@ class BlazeServerBuilder[F[_]](
...
+    "0.21.0-RC3")
   def withSSLContext(
       sslContext: SSLContext,
       clientAuth: SSLClientAuthMode = SSLClientAuthMode.NotRequested): Self =
-    copy(sslBits = Some(SSLContextBits(sslContext, clientAuth)))
+    copy(sslConfig = new ContextWithClientAuth[F](sslContext, clientAuth))
+
+  /** Configures the server with TLS, using the provided `SSLContext` and its
+    * default `SSLParameters` */
+  def withSslContext(sslContext: SSLContext): Self =
```
rossabaker: I'd like to call this `TLS`, which would match modern usage and ember, but is confusing in terms of the raw Java API.

rossabaker:
I'd also like to deprecate `withSSL` and its incomplete set of five parameters, but it's a very long description how to migrate.

rossabaker:
Not entirely proud of what happened in the examples after deprecating `withSSL`.  I don't really want us in the business of convenience wrappers for SSL implementations, but we did make the ergonomics worse.

It also occurs to me the examples are too DRY.  They're easier for us to maintain, but less insightful to the new user.

Issue #3129, opened 2020-01-29
[blazecore] Provide ability to receive WebSocket Ping/Pong frames


khrupalyk:
Current implementation of web socket reading automatically respond to Ping and ignore Pong messages. Also `handleRead` is filtered those messages out and do not send to user stream.

So user will never receive Ping/Pong messages.

It's not a good idea to fix this method directly because all servers will start receiving those messages, and such update could break down a lot of applications.

As @rossabaker propose, we could expose new API:
> handleRead could return the frames. And we could filter them out before the send and receive on WebSocketBuilder.build see them to avoid breaking anything existing. And then introduce an interface that doesn't filter them.

```
  private[this] def handleRead(): F[WebSocketFrame] = {
    def maybeSendClose(c: Close): F[Unit] =
      F.delay(sentClose.compareAndSet(false, true)).flatMap { cond =>
        if (cond) writeFrame(c, trampoline)
        else F.unit
      } >> deadSignal.set(true)

    readFrameTrampoline.flatMap {
      case c: Close =>
        for {
          s <- F.delay(sentClose.get())
          //If we sent a close signal, we don't need to reply with one
          _ <- if (s) deadSignal.set(true) else maybeSendClose(c)
        } yield c
      case Ping(d) =>
        //Reply to ping frame immediately
        writeFrame(Pong(d), trampoline) >> handleRead()
      case _: Pong =>
        //Don't forward pong frame
        handleRead()
      case rest =>
        F.pure(rest)
    }
  }
```

rossabaker:
I'm immersed in TLS parameter enhancements, but would love to get this into the release. It should be fairly straightforward, but maybe a little tricky to keep bincompat. Willing to coach any volunteers.

khrupalyk:
Yes, it's quite easy task. One of the possible solutions is here:
https://github.com/khrupalyk/http4s/commit/b2fce31b2b45c21f55a000093d1e74f672572427
But I'm not sure about bincompat.

rossabaker:
We can break bincompat before 0.21.0-final.  I just want to be sure we're being conservative about breaking applications that are recompiled.  It needs to be a simple migration.  I'll take a look later.  (If you'd like to submit it as a PR, that would make review easier.)

khrupalyk:
@rossabaker PR created.

rossabaker:
Fixed in #3136.

PR #3136, opened 2020-01-31
Added new WS API which allows receiving Ping messages from the client

khrupalyk:
Possible fix for [Provide ability to receive WebSocket Ping/Pong frames](https://github.com/http4s/http4s/issues/3129)`


rossabaker:
Thanks.  Don't we probably want to be able to receive `Pong` frames, too?

khrupalyk:
Probably yes, we should.

khrupalyk:
@rossabaker Updated PR.

rossabaker:
Binary breaking, but not source breaking and should not be semantically breaking.

PR #3137, opened 2020-01-31
Remove Applicative constraint from ArgonautInstances

zarthross:
To match with  #3126


zarthross:
#3138 supersedes this

PR #3138, opened 2020-01-31
Purged the unnecessary Applicative constraint from EntityEncoder instances

zarthross:


LeifW:
If there's no constraint on `F`, I'm wondering why even have that parameter?

rossabaker:
> If there's no constraint on `F`, I'm wondering why even have that parameter?

Some `EntityEncoders` do have a constraint, so it needs to remain on the trait.  And we need to retain the `F` parameter on all these to create the `EntityEncoder[F, A]`.

PR #3145, opened 2020-02-04
Relax Effect to Sync in staticcontent

guymers:
All methods in the `staticcontent` package only require a `Sync`
instance.

### DIFF
--- a/server/src/main/scala/org/http4s/server/staticcontent/MemoryCache.scala
+++ b/server/src/main/scala/org/http4s/server/staticcontent/MemoryCache.scala
@@ -2,8 +2,8 @@ package org.http4s
 package server
 package staticcontent
 
-import cats.effect._
-import cats.implicits._
+import cats.effect.Sync
+import cats.syntax.functor._
 import fs2.{Chunk, Stream}
 import java.util.concurrent.ConcurrentHashMap
 import org.log4s.getLogger
--- a/server/src/main/scala/org/http4s/server/staticcontent/ResourceService.scala
+++ b/server/src/main/scala/org/http4s/server/staticcontent/ResourceService.scala
@@ -10,7 +10,7 @@ object ResourceService {
   /** [[org.http4s.server.staticcontent.ResourceService]] configuration
     *
     * @param basePath prefix of the path files will be served from
-    * @param blocker `ExecutionContext` to use when collecting content
+    * @param blocker execution context to use when collecting content
     * @param pathPrefix prefix of the Uri that content will be served from
     * @param bufferSize size hint of internal buffers to use when serving resources
     * @param cacheStrategy strategy to use for caching purposes. Default to no caching.
--- a/server/src/main/scala/org/http4s/server/staticcontent/ResourceService.scala
+++ b/server/src/main/scala/org/http4s/server/staticcontent/ResourceService.scala
@@ -3,14 +3,14 @@ package server
 package staticcontent
 
 import cats.data.{Kleisli, OptionT}
-import cats.effect._
+import cats.effect.{Blocker, ContextShift, Sync}
 
 object ResourceService {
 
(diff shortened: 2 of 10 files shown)


server/src/main/scala/org/http4s/server/staticcontent/ResourceService.scala
```diff
@@ -3,14 +3,14 @@ package server
...
-import cats.effect._
+import cats.effect.{Blocker, ContextShift, Sync}
 
 object ResourceService {
 
   /** [[org.http4s.server.staticcontent.ResourceService]] configuration
     *
     * @param basePath prefix of the path files will be served from
-    * @param blockingExecutionContext `ExecutionContext` to use when collecting content
+    * @param blocker `ExecutionContext` to use when collecting content
```
aeons: Maybe update this to refer to `cats.effect.Blocker` while you're at it.

server/src/main/scala/org/http4s/server/staticcontent/WebjarService.scala
```diff
@@ -12,7 +12,7 @@ object WebjarService {
 
   /** [[org.http4s.server.staticcontent.WebjarService]] configuration
     *
-    * @param blockingExecutionContext execution context for blocking I/O
+    * @param blocker execution context for blocking I/O
```
aeons: Same as above.

aeons:
Looks good. I've added some comments, and you would need to run scalafmt as well.

server/src/main/scala/org/http4s/server/staticcontent/WebjarService.scala
guymers: `Blocker` is just an execution context so I think the description is valid

server/src/main/scala/org/http4s/server/staticcontent/ResourceService.scala
guymers: Replaced "`ExecutionContext`" with "execution context" to be consistent with the other docblock.

server/src/main/scala/org/http4s/server/staticcontent/MemoryCache.scala
```diff
@@ -2,8 +2,8 @@ package org.http4s
 package server
 package staticcontent
 
-import cats.effect._
-import cats.implicits._
+import cats.effect.Sync
+import cats.syntax.functor._
```
rossabaker: I'd still prefer to use `cats.implicits._` here, but am open to being overruled.

server/src/main/scala/org/http4s/server/staticcontent/WebjarService.scala
rossabaker: I agree with @aeons here.  A `Blocker` is a wrapper that exists to distinguish blocking execution contexts from CPU-bound execution contexts, and steers people toward better thread pool separation through types.

rossabaker: Oops.  GitHub folded the diff, so I only saw the doc, and not the type.  I thought we were trying to use an `ExecutionContext` as a `Blocker`.  I still agree with @aeons that we should not have been quoting it as an `ExecutionContext`, but I think this phrasing is fine.

guymers: Not 100% sure I understand what you mean by refer to `Blocker`. As `Blocker` is the type of the parameter does it need to also be in the description? Something like "`Blocker` for blocking I/O"?

rossabaker: I think as is, "execution context for blocking I/O", is a good description.  

I would like us to avoid calling a blocker an `ExecutionContext`, with that capitalization, in backticks, which makes it look like a type name.

PR #3147, opened 2020-02-06
Add GitHub Action

rossabaker:
In scope:
* Validates on Scala {2.12, 2.13} x Java {8, 11}
* Builds scalafix
* Builds project website
* Builds docs site

Out of scope (for now):
* Publish snapshots
* Publish to sonatype
* Publish website

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,113 @@
+name: CI
+on: [push, pull_request]
+env:
+  HUGO_VERSION: 0.26
+  SCALA_VERSION: 2.12.10
+jobs:
+  validate:
+    name: Scala ${{ matrix.scala }}, Java ${{ matrix.java }}
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        java: [8, 11]
+        scala: [2.12.10, 2.13.1]
+    env:
+      SCALA_VERSION: ${{ matrix.scala }}
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-java@v1
+        with:
+          java-version: ${{ matrix.java }}
+      - name: Cache Coursier
+        uses: actions/cache@v1
+        with:
+          path: ~/.cache/coursier
+          key: sbt-coursier-cache
+      - name: Cache SBT
+        uses: actions/cache@v1
+        with:
+          path: ~/.sbt
+          key: sbt-${{ hashFiles('**/build.sbt') }}
+      - name: Check formatting
+        if: startsWith(matrix.scala, '2.12.') && (matrix.java == '8')
+        run: sbt ++$SCALA_VERSION scalafmtCheckAll
+      - name: Compile
+        run: sbt ++$SCALA_VERSION test:compile
+      - name: Check compatibility
+        run: sbt ++$SCALA_VERSION mimaReportBinaryIssues
+      - name: Check unused dependencies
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,113 @@
...
+      - name: Cache Coursier
+        uses: actions/cache@v1
+        with:
+          path: ~/.cache/coursier
+          key: sbt-coursier-cache
+      - name: Cache SBT
+        uses: actions/cache@v1
+        with:
+          path: ~/.sbt
+          key: sbt-${{ hashFiles('**/build.sbt') }}
```
rossabaker: Could use help from a Principal YAML Engineer to DRY these steps up.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,113 @@
...
+      - name: Compile
+        run: sbt ++$SCALA_VERSION test:compile
+      - name: Check compatibility
+        run: sbt ++$SCALA_VERSION mimaReportBinaryIssues
+      - name: Check unused dependencies
+        run: sbt ++$SCALA_VERSION unusedCompileDependenciesTest
+      - name: Test
+        run: sbt ++$SCALA_VERSION test
+      - name: Scaladoc
+        run: sbt ++$SCALA_VERSION doc
```
rossabaker: Splitting all these steps makes it clearer where it fails, but costs us ~20 seconds to start a new SBT each time.

LeifW: You mean being able to see the failure stage in the GitHub UI, as opposed to looking in the console output to see where it failed?

.github/workflows/ci.yml
LeifW: I know YAML supports references and can extend other objects.
Though perhaps putting the yaml into another format and generating it from an actual programming language would be preferable to going that route?
Though, this YAML already looks like mustache-style template variables in it... (a la that kubernetes stuff?)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,113 @@
+name: CI
+on: [push, pull_request]
+env:
+  HUGO_VERSION: 0.26
+  SCALA_VERSION: 2.12.10
```
LeifW: So this `SCALA_VERSION` is for the jobs other than `validate`, which has its defined as `SCALA_VERSION: ${{ matrix.scala }}` from the matrix?

.github/workflows/ci.yml
rossabaker: I know YAML has "anchors", which we use in .travis.yml, and I read three articles on them last night and still don't understand them. 
 Job 1 has steps A, B, C, D. Job 2 has steps A, B, C, E. How does one extract A, B, and C?

On a larger scale, I'd reach for Dhall, but we'd have to check in generated source, and it would be a new language and a new dependency for something that just won't change that often. I think pure YAML is sadly the right way here.

.github/workflows/ci.yml
rossabaker: Yes,  exactly. I guess we could time it with and without and see what it costs.

.github/workflows/ci.yml
rossabaker: Yes. And I ended up hardcoding a similar Java 8 elsewhere.

.github/workflows/ci.yml
rossabaker: I made a bunch of anchors, only to find out that they're [not supported](https://github.community/t5/GitHub-Actions/Support-for-YAML-anchors/td-p/30336).

PR #3148, opened 2020-02-06
Add HttpRoutes.strict

ChristopherDavenport:
If users don't care about efficient route composition under a large number of routes, this exposes the ability of `of` as a partial function but only requires applicative and not `of`. 

Open to different naming semantics. This change doesn't effect existing api and can be a patch if we don't want it in final, but I don't see why that's necessary for a convenience method like this.

### DIFF
--- a/core/src/main/scala/org/http4s/HttpRoutes.scala
+++ b/core/src/main/scala/org/http4s/HttpRoutes.scala
@@ -62,18 +62,16 @@ object HttpRoutes {
   def of[F[_]: Defer: Applicative](pf: PartialFunction[Request[F], F[Response[F]]]): HttpRoutes[F] =
     Kleisli(req => OptionT(Defer[F].defer(pf.lift(req).sequence)))
 
-
   /** Lifts a partial function into an [[HttpRoutes]].  The application of the
     * partial function is not suspended in `F`, unlike [[of]]. This allows for less
     * constraints when not combining many routes.
     *
-    * @tparam F the base effect of the [[HttpRoutes]] - Defer suspends evaluation
-    * of routes, so only 1 section of routes is checked at a time.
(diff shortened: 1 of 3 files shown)


core/src/main/scala/org/http4s/HttpRoutes.scala
```diff
@@ -62,6 +62,20 @@ object HttpRoutes {
   def of[F[_]: Defer: Applicative](pf: PartialFunction[Request[F], F[Response[F]]]): HttpRoutes[F] =
     Kleisli(req => OptionT(Defer[F].defer(pf.lift(req).sequence)))
 
+
+  /** Lifts a partial function into an [[HttpRoutes]].  The application of the
+    * partial function is not suspended in `F`, unlike [[of]]. This allows for less
+    * constraints when not combining many routes.
+    *
+    * @tparam F the base effect of the [[HttpRoutes]] - Defer suspends evaluation
```
rossabaker: I think this `Defer` is a copy-paste error.

core/src/main/scala/org/http4s/HttpRoutes.scala
```diff
@@ -62,6 +62,20 @@ object HttpRoutes {
...
+    * of routes, so only 1 section of routes is checked at a time.
+    * @param pf the partial function to lift
+    * @return An [[HttpRoutes]] that returns some [[Response]] in an `OptionT[F, ?]`
+    * wherever `pf` is defined, an `OptionT.none` wherever it is not
+    */
+  def simple[F[_]: Applicative](pf: PartialFunction[Request[F], F[Response[F]]]): HttpRoutes[F] =
```
rossabaker: `strict`?

rossabaker:
Failure is unrelated and #3149 should help.

Issue #3128, opened 2020-01-29
Flaky test: allow GET body to clean up on fallthrough


rossabaker:
```
[info] DefaultHead should
[info]   + honor HEAD routes (4 ms)
[info]   + return truncated body of corresponding GET on fallthrough (39 ms)
[info]   + retain all headers of corresponding GET on fallthrough (9 ms)
[error]   x allow GET body to clean up on fallthrough (27 ms)
[error]    the value is false (DefaultHeadSpec.scala:55)
```

PR #3150, opened 2020-02-06
Make DefaultHeadSpec Test Errors Clearly Apparent, Switch Finalization Case

ChristopherDavenport:
Related to #3128

### DIFF
--- a/server/src/test/scala/org/http4s/server/middleware/DefaultHeadSpec.scala
+++ b/server/src/test/scala/org/http4s/server/middleware/DefaultHeadSpec.scala
@@ -3,13 +3,17 @@ package server
 package middleware
 
 import cats.effect._
-import cats.effect.concurrent.Ref
+import cats.effect.concurrent.Deferred
 import fs2.Stream
 import org.http4s.Uri.uri
 import org.http4s.dsl.io._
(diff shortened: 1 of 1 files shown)


server/src/test/scala/org/http4s/server/middleware/DefaultHeadSpec.scala
```diff
@@ -3,13 +3,17 @@ package server
...
 import cats.effect._
-import cats.effect.concurrent.Ref
+import cats.effect.concurrent.Deferred
 import fs2.Stream
 import org.http4s.Uri.uri
 import org.http4s.dsl.io._
 import org.http4s.testing.Http4sLegacyMatchersIO
+import cats.effect.testing.specs2.CatsIO
+
+class DefaultHeadSpec extends Http4sSpec with Http4sLegacyMatchersIO with CatsIO {
```
rossabaker: If you extend `CatsEffect` instead of `CatsIO`, you don't need the two overrides below.

rossabaker:
Note: we still want to switch back to a spec that looks more like this, even after #3157.  This is not tenable for 0.21, but this is right in the long run, so let's not throw it away.

rossabaker:
Is this worth porting to MUnit?

PR #3151, opened 2020-02-06
Migrate to sbt-ci-release, remove Travis CI

rossabaker:
Adds a release workflow, based on sbt-release-ci.

- On tag, publishes a release to Sonatype. (Untested).
- On master or series/*, publishes a snapshot to Sonatype.
- On master, publishes the website project.
- On master or series/* or tag, publishes the docs project.

This removes Travis CI and its associated cruft.  Our next release should be by pushing a tag.

Plugins that are transitive through sbt-release-ci are now managed through that, to reduce trouble and Scala Steward noise.  sbt-release is removed, so we had to refactor away from its `Version` to generate the version table on the project site.  That's most of the noise in Http4sPlugin.scala.


rossabaker:
This will fail on release to master until I set up the secrets.  I've been testing this on my fork.

.github/workflows/release.yml
```diff
@@ -0,0 +1,106 @@
...
+      - name: Publish
+        run: |
+          eval "$(ssh-agent -s)"
+          echo "$SSH_PRIVATE_KEY" | ssh-add -
+          git config --global user.name "GitHub Actions CI"
+          git config --global user.email "ghactions@invalid"
+          sbt ++$SCALA_VERSION website/makeSite website/ghpagesPushSite
+        env:
+          SSH_PRIVATE_KEY: ${{ secrets.SSH_PRIVATE_KEY }}
+  publish-docs:
```
rossabaker: This should maybe not happen on tag.

.github/workflows/release.yml
```diff
@@ -0,0 +1,106 @@
...
+on:
+  push:
+    branches:
+      - master
+      - series/*
+    tags: 
+      - '*'
+env:
+  HUGO_VERSION: 0.26
+  SCALA_VERSION: 2.12.10
```
rossabaker: I wish we weren't duplicating these with ci.yml.  Ideas?  We could use secrets, but that would be an abuse.

.github/workflows/release.yml
```diff
@@ -0,0 +1,106 @@
...
+      - name: Install Hugo
+        run: scripts/install-hugo
+      - name: Configure git
+        run: |
+          eval "$(ssh-agent -s)"
+          echo "$SSH_PRIVATE_KEY" | ssh-add -
+          git config --global user.name "GitHub Actions CI"
+          git config --global user.email "ghactions@invalid"
+        env:
+          SSH_PRIVATE_KEY: ${{ secrets.SSH_PRIVATE_KEY }}
```
rossabaker: This step is vestigial.

.github/workflows/release.yml
amesgen: This duplication has already bothered me in http4s-jdk-http-client. Concerning this (and more generally, all YAML shortcomings): One could generate these config files via [Dhall](https://dhall-lang.org/) (there are already approaches for this in several languages, see e.g. [here](https://github.com/vmchale/github-actions-dhall)). This config could even be shared between multiple (http4s) repos (integrity is ensured).

Downsides:
 - Contributing to the CI scripts requires Dhall knowledge (on the other hand, translating YAML changes to the Dhall source files should be very easy, and it *should* be relatively rare).

Advantages:
 - No YAML surprises (we already had that in http4s-jdk-http-client...). I think it is safe to claim that Dhall is superior from a design perspective.
 - DRY: Sharing of variables/snippets between files or even between repos

I could try to create an example for http4s-jdk-http-client to see if this is an approach worth pursuing.

rossabaker: Ha! I mentioned Dhall on the previous ticket (#3147).  It's our strategy at work for escaping YAML Hell.

One other downside is that we'd be forced to keep generated code in our repositories.  I strongly dislike having a source of truth and a second git-managed reflection of that truth.  But maybe it's no worse than copying and pasting, and a having a Dhall utility to generate these could pay off as the integrations spin off into separate repos and as we grow more. Maybe we should see if other maintainers are enthused or horrified by the idea.

The other way we're winning at work is putting as much logic as possible into ci.nix, and the YAML remains thin and rarely changes.  The equivalent in our world would be to do what we can in sbt, or even standardly named shell scripts.  But part of the reason ours are so big here is because we're splitting into multiple jobs for greater concurrency.

.github/workflows/release.yml
rossabaker: Well, I can't say why not.  There may be version numbers to update.  This might bite us if tag and master builds are both trying to push and one gets a stale head.  We can adjust if it happens.

PR #3157, opened 2020-02-08
Alternate GET body cleanup test

rossabaker:
Mitigates #3128.

I prefer the spec in #3150 in the long run, but it exposes a problem that would be solved with #2663.  This tests that we don't leak a resource acquired in the Stream, which is about the best we can do without `HttpResource`.


rossabaker:
If this doesn't work, and nobody has thought of anything clever to make #3150 pass reliably, I'm in favor of a defeatist skipOnCi for 0.21, and fix it for real in the next release.

PR #3158, opened 2020-02-08
Release v0.21.0-RC5

rossabaker:
Will give us a chance to test the new release pipeline from #3151 before burning the eagerly anticipated 0.21.0 tag.


rossabaker:
Note, in case anyone merges this: This will not release until we push the tag. Watching some of the other merges in CI before pulling the trigger.

PR #3165, opened 2020-02-08
Set GHA cache keys

rossabaker:
The cache does not incrementally update.  We need to give it a key unique to what it's going to download.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,16 +21,22 @@ jobs:
       - uses: actions/setup-java@v1
         with:
           java-version: ${{ matrix.java }}
-      - name: Cache Coursier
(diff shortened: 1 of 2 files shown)


amesgen:
If I understand these changes correctly, the entire cache is thrown away if a single sbt setting or a single dependency got changed. Previously, the coursier cache did not change at all, which is also suboptimal.

Maybe it is better to just reset the caches every week/month (by using sth like `key: CURRENT_WEEK `)?

rossabaker:
I've seen it argued that the cache is not worthwhile in the first place. That would also be worth considering.

rossabaker:
I'm making a mess with all the merges. Going to try again elsewhere.

PR #3166, opened 2020-02-09
WIP - Classifier helper

kevinmeredith:
Adds a function, `classifierFMethodWithOptionallyExcludedPath`, to `Prometheus`.

Given an exclude function, return a 'classifier' function, i.e. for application in
`[[org.http4s.client.middleware.Metrics#apply]]`.

Let's say you want a classifier that excludes integers since your paths consist of:
```
    * GET    /users/{integer}    = get_users_*
    * POST   /users                = post_users
    * PUT    /users/{integer}    = put_users_*
    * DELETE /users/{integer} = delete_users_*
```

In such a case, we could use:

```scala
   classifierFMethodWithOptionallyExcludedPath[F] { 
         str: String => scala.util.Try(str.toInt).isSuccess 
   }
```

This idea, namely adding a classifier with an `exclude: String => Boolean` was inspired by an internal library that @ChristopherDavenport wrote.

### DIFF
--- a/prometheus-metrics/src/main/scala/org/http4s/metrics/prometheus/Prometheus.scala
+++ b/prometheus-metrics/src/main/scala/org/http4s/metrics/prometheus/Prometheus.scala
@@ -1,9 +1,12 @@
 package org.http4s.metrics.prometheus
 
+import cats.Foldable
 import cats.data.NonEmptyList
 import cats.effect.Sync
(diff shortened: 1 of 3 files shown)


kevinmeredith:
I intended to open this PR against `series/0.20`. As a result, I'll close this one and open another.

PR #3167, opened 2020-02-09
Add Helper Function to Build classifierF (to pass to client and server's Middleware)

kevinmeredith:
Adds a function, `classifierFMethodWithOptionallyExcludedPath`, to `Prometheus`.

Given an exclude function, return a 'classifier' function, i.e. for application in
`[[org.http4s.client.middleware.Metrics#apply]]` and `[[org.http4s.server.middleware.Metrics#apply]]`.

Let's say you want a classifier that excludes integers since your paths consist of:
```
    * GET    /users/{integer}    = get_users_*
    * POST   /users                = post_users
    * PUT    /users/{integer}    = put_users_*
    * DELETE /users/{integer} = delete_users_*
```

In such a case, we could use:

```scala
   classifierFMethodWithOptionallyExcludedPath[F] { 
         str: String => scala.util.Try(str.toInt).isSuccess 
   }
```

This idea, namely adding a classifier with an `exclude: String => Boolean`, was inspired by an internal library that @ChristopherDavenport wrote.


testing/src/main/scala/org/http4s/testing/ArbitraryInstances.scala
```diff
@@ -861,6 +861,9 @@ trait ArbitraryInstances {
         body <- http4sTestingGenForPureByteStream
       } yield Response(status, httpVersion, headers, body)
     }
+
+  implicit val arbUUID: Arbitrary[UUID] =
```
rossabaker: I would rather put this in another module of arbitraries, like we did for fs2.  I don't like providing instances of types we don't own for typeclasses we don't own into public API.

I'm pretty surprised this isn't core scalacheck functionality.

prometheus-metrics/src/main/scala/org/http4s/metrics/prometheus/Prometheus.scala
```diff
@@ -191,6 +194,45 @@ object Prometheus {
...
+    *
+    * classifierFMethodWithOptionallyExcludedPath { str: String => scala.util.Try(str.toInt).isSuccess }
+    *
+    * @param exclude For a given String, namely a path value, determine whether the value gets excluded.
+    * @return Request[F] => Option[String]
+    */
+  def classifierFMethodWithOptionallyExcludedPath[F[_]](
+      exclude: String => Boolean
+  ): Request[F] => Option[String] = { request: Request[F] =>
+    val dsl: Http4sDsl[F] = Http4sDsl[F]
```
rossabaker: I'm surprised http4s-dsl is available in this project.  That seems like a bad coupling.  Can we do this without it?

rossabaker:
Are you needing this in 0.20.x?

kevinmeredith:
> Are you needing this in 0.20.x?

I would not say need, but it'd be nice. If it's a pain to support it in 0.20.x, please let me know, Ross.

prometheus-metrics/src/main/scala/org/http4s/metrics/prometheus/Prometheus.scala
kevinmeredith: > I'm surprised http4s-dsl is available in this project. That seems like a bad coupling.

I only used it since I saw it available via https://github.com/http4s/http4s/blob/series/0.20/build.sbt#L153. 

> Can we do this without it?

Sure. Can you please advise if you know, off-hand, how to get the path as a `List[String]`, @rossabaker?

testing/src/main/scala/org/http4s/testing/ArbitraryInstances.scala
kevinmeredith: > I would rather put this in another module of arbitraries, like we did for fs2.

Can you please say more? I don't follow on how to implement that. My guess was just adding it to `Http4sSpec` like https://github.com/http4s/http4s/blob/series/0.20/testing/src/test/scala/org/http4s/Http4sSpec.scala#L62, but I'm not sure.

I pushed up https://github.com/http4s/http4s/pull/3167/commits/4329e58da7e6cda204f068cd9d78b0d5c0d3079b, but please let me know if it does not address your comment, Ross.

rossabaker:
I'd like to phase out 0.20.x releases in general and focus on keeping 0.21 healthy and developing the next version, but if somebody sends PRs, there's no particular reason we can't release another 0.20.

testing/src/main/scala/org/http4s/testing/ArbitraryInstances.scala
rossabaker: That's fine.  We added [fs2Arbitraries](https://github.com/http4s/http4s/blob/master/testing/src/test/scala/org/http4s/testing/fs2Arbitraries.scala) in 0.21, which I think is a good pattern for these.  But the important thing here is that they're not in `src/main`.  We don't want other people depending on ours and someone else's.  If we keep it private, and Scalacheck adds one later, we can delete ours without affecting anyone else.

prometheus-metrics/src/main/scala/org/http4s/metrics/prometheus/Prometheus.scala
rossabaker: You could copy [this code](https://github.com/http4s/http4s/blob/0801749674a83afa3e7522f4e402d1bedea76982/dsl/src/main/scala/org/http4s/dsl/impl/Path.scala#L57-L64) and build a `List` rather than a `Path`.

In the long run, turning a path into a list of segments ought to be core functionality on a URI.

rossabaker: So this dependency already exists, to support the `PrometheusExportService`.  I think we could re-express it without that, but that's a separate ticket.

kevinmeredith: Thanks, Ross, for telling me how to get a path list from a `String`. I did that, as well as removed the `Http4sDsl` usage via commits at https://github.com/http4s/http4s/pull/3167#commits-pushed-1c3a145.

testing/src/main/scala/org/http4s/testing/ArbitraryInstances.scala
kevinmeredith: Roger that, Ross. I wasn't aware that `Arbitraries`, where I originally added the `Arbitrary[UUID]` lives at `src/main`. I wrongly assumed it was `src/test`.

kevinmeredith:
> This should still pass MiMa checks and be eligible for 0.20.18 and/or 0.21.1.

Understood, @rossabaker. I ran `mimaReportBinaryIssues` from `project root` successfully. I'll check the build's result tomorrow.

I believe that I've addressed your review findings, Ross. Can you please take a look? 

> I'd like to phase out 0.20.x releases in general and focus on keeping 0.21 healthy and developing the next version, but if somebody sends PRs, there's no particular reason we can't release another 0.20.

I interpret this to mean that it's OK for me to target `series/0.20`, but let me know, please, if I understood wrongly.

kevinmeredith:
Since the other tests succeeded per https://travis-ci.org/http4s/http4s/builds/648221349?utm_source=github_status&utm_medium=notification, I'm guessing the test failure in https://travis-ci.org/http4s/http4s/jobs/648221350?utm_medium=notification&utm_source=github_status was the result of flakiness.

```
[info] BlazeClientSpec
[info] Blaze Http1Client should
[info]   + raise error NoConnectionAllowedException if no connections are permitted for key (383 ms)
[info]   + make simple https requests (940 ms)
[info]   + behave and not deadlock (994 ms)
[error]   x behave and not deadlock on failures with parTraverse (30 seconds, 186 ms)
[error]    None is not Some (BlazeClientSpec.scala:160)
```

As a result, I'll close, and then re-open, the PR to prompt another build.

kevinmeredith:
Looks like another, I'm guessing, flaky test is to blame per https://travis-ci.org/http4s/http4s/jobs/648223657?utm_medium=notification&utm_source=github_status. I will close, and then re-open. 

To make sure I follow, is it necessary to get the build green, @rossabaker, even if you have high confidence that a PR was not to blame for a failing test?

rossabaker:
And yes, if you need this on 0.20.18, it's not a problem to keep it on this branch.  I backported the GitHub Actions build and threw in a Tomcat bugfix in separate PRs to make it more worth our while.

kevinmeredith:
Hey @ChristopherDavenport. Thanks for reviewing. I will update this PR per your comments.

prometheus-metrics/src/main/scala/org/http4s/metrics/prometheus/Prometheus.scala
```diff
@@ -191,6 +193,60 @@ object Prometheus {
...
+  private def requestToPathList[F[_]](request: Request[F]): List[String] = {
+    val str: String = request.pathInfo
+
+    if (str == "" || str == "/")
+      Nil
+    else {
+      val segments = str.split("/", -1)
+      // .head is safe because split always returns non-empty array
+      val segments0 = if (segments.head == "") segments.drop(1) else segments
+      segments0.foldLeft[List[String]](Nil)((path, seg) => path ++ List(Uri.decode(seg)))
```
ChristopherDavenport: This method is super inefficient, we should fix it in both places to cons and then reverse afterwards.

kevinmeredith: ACK - https://github.com/http4s/http4s/pull/3167/commits/03517004f7b22cd5b5f9b4cb70a1e9d1239ef187.

kevinmeredith:
> Nothing of this is specific to Prometheus so could be in Metrics companion Object.

Which `Metrics`, @ChristopherDavenport, `org.http4s.server.middleware.Metrics` or `org.http4s.client.middleware.Metrics`?

kevinmeredith:
Hey Chris - I believe that I've addressed 2 and 3 of https://github.com/http4s/http4s/pull/3167#pullrequestreview-356313901. Let me know, please, if I have not.

prometheus-metrics/src/main/scala/org/http4s/metrics/prometheus/Prometheus.scala
```diff
@@ -197,6 +197,8 @@ object Prometheus {
     * Given an exclude function, return a 'classifier' function, i.e. for application in
     * [[org.http4s.server/client.middleware.Metrics#apply]].
     *
+    * Note that the output String will be lower-cased.
```
kevinmeredith: What are your thoughts on this, @ChristopherDavenport?

prometheus-metrics/src/main/scala/org/http4s/metrics/prometheus/Prometheus.scala
```diff
@@ -191,6 +193,74 @@ object Prometheus {
...
+    * )
+    *
+    * @param exclude For a given String, namely a path value, determine whether the value gets excluded.
+    * @param excludedValue Indicates the String value to be supplied for an excluded path's field.
+    * @param pathSeparator Value to use for separating the metrics fields' values
+    * @return Request[F] => Option[String]
+    */
+  def classifierFMethodWithOptionallyExcludedPath[F[_]](
+      exclude: String => Boolean,
+      excludedValue: String,
```
ChristopherDavenport: I'm inclined towards default arguments here.

prometheus-metrics/src/main/scala/org/http4s/metrics/prometheus/Prometheus.scala
ChristopherDavenport: Why do we want to lower-case it? Its case-insensitive, so its as sane a default as an in order to flatten them, but it does seem like an arbitrary choice.

prometheus-metrics/src/main/scala/org/http4s/metrics/prometheus/Prometheus.scala
```diff
@@ -191,6 +193,74 @@ object Prometheus {
...
+    *   intercalateValue = "_"
+    * )
+    *
+    * @param exclude For a given String, namely a path value, determine whether the value gets excluded.
+    * @param excludedValue Indicates the String value to be supplied for an excluded path's field.
+    * @param pathSeparator Value to use for separating the metrics fields' values
+    * @return Request[F] => Option[String]
+    */
+  def classifierFMethodWithOptionallyExcludedPath[F[_]](
+      exclude: String => Boolean,
```
ChristopherDavenport: Perhaps worth noting that this run on every segment of a path. So note that if an intermediate throwables with Stack traces is known and discarded there may be a performance penalty, like the examples you have shown with `Int` and `UUID`, both I benchmarked some approaches and regex matches should generally be preferred over throwables in this position.

prometheus-metrics/src/main/scala/org/http4s/metrics/prometheus/Prometheus.scala
```diff
@@ -191,6 +193,74 @@ object Prometheus {
...
+      excludedValue: String,
+      pathSeparator: String
+  ): Request[F] => Option[String] = { request: Request[F] =>
+    val initial: String = request.method.name.toLowerCase
```
ChristopherDavenport: Method is case sensitive. We should not change its case.

rossabaker:
@kevinmeredith Any estimate when you'll be able to circle back to this?  No hurry, just trying to plan the next release.  We have a couple bugfixes queued up that I'd like to get out there.  This can always make the next bus.

rossabaker:
Not going to block 0.20.18 for this.  We still want it, and we won't run out of numbers.

prometheus-metrics/src/main/scala/org/http4s/metrics/prometheus/Prometheus.scala
kevinmeredith: Yep - good point, Chris. I removed the `toLowerCase` in https://github.com/http4s/http4s/pull/3167#commits-pushed-badda65's commits.

prometheus-metrics/src/main/scala/org/http4s/metrics/prometheus/Prometheus.scala
kevinmeredith: I dislike default arguments, but I pushed up https://github.com/http4s/http4s/pull/3167/commits/ae745986f513ffb1693096e8a6c6e292ed0f81c9 to address your comment, Chris.

prometheus-metrics/src/main/scala/org/http4s/metrics/prometheus/Prometheus.scala
kevinmeredith: I fully agree: https://github.com/http4s/http4s/pull/3167#discussion_r378644116.

prometheus-metrics/src/main/scala/org/http4s/metrics/prometheus/Prometheus.scala
kevinmeredith: Thanks for pointing that out. I added https://github.com/http4s/http4s/pull/3167/commits/48c0661ca6a6ce800d39332685139a9b95c80541.

kevinmeredith:
I was going to update `core` to depend on `testing % "test -> test"`, but I see that `testing` depends on `core`... I'll address this tomorrow.

kevinmeredith:
As a result of https://github.com/http4s/http4s/pull/3167#issuecomment-585546292, I pushed up https://github.com/http4s/http4s/pull/3167#commits-pushed-0c6873a.

Let me know, please, your thoughts on my adding of `specs2-scalacheck` as a depedency (in test only) to `core`, @rossabaker and @ChristopherDavenport.

core/src/test/scala/org/http4s/metrics/MetricsOpsSpec.scala
```diff
@@ -1,21 +1,27 @@
...
+import org.specs2.ScalaCheck
+import org.specs2.mutable.Specification
 
-object PrometheusSpec {
+object MetricsOpsSpec {
 
   private implicit val arbUUID: Arbitrary[UUID] =
     Arbitrary(Gen.uuid)
+
+  private implicit val http4sTestingArbitraryForMethod: Arbitrary[Method] = Arbitrary(
```
rossabaker: We already have one of these in ArbitraryInstances in the testing project.

build.sbt
```diff
@@ -98,6 +98,7 @@ lazy val core = libraryProject("core")
       log4s,
       parboiled,
       vault,
+      specs2ScalacheckTest(scalaVersion.value)
```
rossabaker: Most of core is tested in the `tests` project, so it can depend on `testing`, which we also publish for third-party extensions.  That should already have this dependency, in addition to the `Arbitrary` that you had to reinvent.

We should make this clearer in the contributors guide.

kevinmeredith: Ah, I see. Thanks, Ross. I believe that I've addressed your comments via https://github.com/http4s/http4s/pull/3167#commits-pushed-ff9cae6.

core/src/test/scala/org/http4s/metrics/MetricsOpsSpec.scala
kevinmeredith: 👍 https://github.com/http4s/http4s/pull/3167/commits/59feb8bdb5585b3a0ed5bd503b79a61c67a2dcf5

kevinmeredith:
Can you please review this one, @ChristopherDavenport, again?

kevinmeredith:
The most recent [build](https://github.com/http4s/http4s/pull/3167/checks?check_run_id=448513742) appears to have run into a transient error.

```
[error] coursier.ResolutionException: 1 download error
[error]     Caught java.nio.channels.OverlappingFileLockException while downloading https://repo1.maven.org/maven2/com/fasterxml/jackson/core/jackson-databind/2.9.8/jackson-databind-2.9.8.jar
[error] (play-json / update) coursier.ResolutionException: 1 download error
[error]     Caught java.nio.channels.OverlappingFileLockException while downloading https://repo1.maven.org/maven2/com/fasterxml/jackson/core/jackson-databind/2.9.8/jackson-databind-2.9.8.jar
[error] Total time: 299 s (04:59), completed Feb 16, 2020, 4:53:50 AM
##[error]Process completed with exit code 1.
```

I'll close, and then re-open.

kevinmeredith:
I see a few tests failed:

```
[info] Blaze Http1Client should
[info]   + raise error NoConnectionAllowedException if no connections are permitted for key (1 second, 131 ms)
[info]   + make simple https requests (1 second, 158 ms)
[info]   + behave and not deadlock (1 second, 182 ms)
[error]   x behave and not deadlock on failures with parTraverse (30 seconds, 576 ms)
[error]    None is not Some (BlazeClientSpec.scala:160)
[error] org.http4s.client.blaze.BlazeClientSpec.$anonfun$new$21(BlazeClientSpec.scala:160)
[info]   + behave and not deadlock on failures with parSequence (201 ms)
[info]   + obey response header timeout (212 ms)
[info]   + unblock waiting connections (1 second, 55 ms)
[info]   + reset request timeout (2 seconds, 57 ms)
[info]   + drain waiting connections after shutdown (1 second, 139 ms)
[info]   + cancel infinite request on completion (1 second, 151 ms)
[info]   + doesn't leak connection on timeout (1 second, 161 ms)
[error]   x call a second host after reusing connections on a first (5 seconds, 68 ms)
[error]    Some(Left(())) != Some(Right(simple path)) (BlazeClientSpec.scala:324)
[error] org.http4s.client.blaze.BlazeClientSpec.$anonfun$new$82(BlazeClientSpec.scala:324)
[error] Actual:   Some(Left(()))
[error] Expected: Some(Right(simple path))
[info]   + raise a ConnectionFailure when a host can't be resolved (26 ms)
[info] Total for specification BlazeClientSpec
[info] Finished in 36 seconds, 859 ms
[info] 13 examples, 18 expectations, 2 failures, 0 error
[error] Failed: Total 66, Failed 2, Errors 0, Passed 61, Skipped 3, Pending 1
```

* https://github.com/http4s/http4s/pull/3167/checks?check_run_id=448513735
* https://github.com/http4s/http4s/pull/3167/checks?check_run_id=448513742

but I highly doubt that my changes introduced such tests' failures.

rossabaker:
Yes, three are still a few tests that are flaky.  I was hoping the move to GitHub Actions would help, and, well, maybe, but not entirely.

kevinmeredith:
Hi @rossabaker - please let me know what I can do to influence the merging of this PR. Thanks.

rossabaker:
@ChristopherDavenport, are all your comments addressed?

kevinmeredith:
Hi @rossabaker and @ChristopherDavenport - could you please cut a release per https://github.com/http4s/http4s/compare/v0.20.19...series/0.20? I don't have a rush, but I'd like to start using this function. Thanks!

kevinmeredith:
Could you please take a look at https://github.com/http4s/http4s/pull/3167#issuecomment-596875541, Ross or Chris? 

I believe I have a use case for this in prod today.

rossabaker:
I was trying to get another fix in, but life has been upside down.  Let's get this out now.  Versions are cheap.

kevinmeredith:
> I was trying to get another fix in, but life has been upside down. Let's get this out now. Versions are cheap.

Thank you, @rossabaker !

kevinmeredith:
> I was trying to get another fix in, but life has been upside down

Yes, unfortunately so. I'm glad to have spoken with you last week at NEScala. I hope you're doing well.

PR #3177, opened 2020-02-10
Update .mergify.yml for github actions

### LINKED ISSUE #3170: Fix Mergify
Mergify is looking for a Travis CI status check, which will no longer come.

We can make it wait for GitHub Actions jobs, but we need them to wait for several jobs, I think.  This seems verbose and unmaintainable?  Is there a way to look for the success of a workflow rather than a job?

https://doc.mergify.io/conditions.html#github-actions

aeons:
This requires updating when we switch java versions.

Closes #3170

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -3,8 +3,14 @@ pull_request_rules:
   - name: automatically merge scala-steward's PRs
     conditions:
       - author=scala-steward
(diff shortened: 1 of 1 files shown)


rossabaker:
I also don't see how to test it without a Scala Steward commit, but I don't think this will generate any false positives, and otherwise can't break things worse than they were.

PR #3179, opened 2020-02-10
Update to jawn-fs2-1.0.0

rossabaker:
The only change in jawn-fs2 is to bump from fs2-2.2.1 to fs2-2.2.2, which http4s already depends on.  This gets us off an RC that we neglected to finalize before http4s-0.21.0.

Thanks to @travisbrown for calling it out on Gitter.


travisbrown:
Thanks @rossabaker!

PR #3185, opened 2020-02-11
Recover EOF on bodyEncoder.write to close connection

rossabaker:
Recover EOF raised by the body encoder rather than catching it as an exception, which some implementations may wrap.  This better expresses intent and cleans up logs in such implementations.

/cc @jadlr


rossabaker:
This can be tested now in 0.21.0+12-ffbd5e7a-SNAPSHOT, published to OSS snapshots.  It will be part of the 0.21.1 release, hopefully in the next day or so.

jadlr:
Just deployed using the snapshot version. This fixes the issue :+1: 

Thanks for the quick fix!

PR #3188, opened 2020-02-12
Pagination for Client

insdami:
Pagination logic which might come handy for people querying paginated resources. 
I'm not expecting to be perfect from the get go but I wanted to put it out there to get opinions whether or not should be a client concern and if it's what'd be a good api for it.

The pagination logic was inspired by https://github.com/Dwolla/fs2-aws/blob/2.x/core/shared/src/main/scala/com/dwolla/fs2utils/Pagination.scala

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ jobs:
       SCALA_VERSION: ${{ matrix.scala }}
     steps:
       - uses: actions/checkout@v2
-+     - uses: olafurpg/setup-scala@v5
+      - uses: olafurpg/setup-scala@v5
         with:
           java-version: ${{ matrix.java }}
       - name: Cache Coursier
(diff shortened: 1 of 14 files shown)


goedelsoup:
I really would prefer this to exist outside the release cycle of http4s. I think this is going to drive a train of iterative releases that I'd prefer to exclude from my (and others) adoption of http4s. I acknowledge that this is a useful feature that I'd even use myself, but I'd prefer it to exist as a supplemental piece rather than something that drives up the 0.21 minor and confuses what people should adopt as a core pathway.

goedelsoup:
What i didn't highlight: this should probably be feature advertised from the http4s org. It's super useful IMO, but doesn't belong in the core.

hamnis:
I would suggest we add a page quite visible on the documentation site that lists derived projects. I can look into that on my next oss cycle.

I agree that this project is super useful, and can see applications for it @ work .

insdami:
I've seen on gitter that people shared some opinions around this and @ChristopherDavenport wrote a very similar thing but more succinct and using https://github.com/functional-streams-for-scala/fs2/blob/master/core/shared/src/main/scala/fs2/Stream.scala#L3668
I personally found my self writing some pagination logic in some shape or form

rossabaker:
Need to decide if pagination belongs on the client interface or off to the side.

armanbilge:
Now that we're schisming everything anyway I'd vote microlib. Feel free to open an issue/discussion if it makes sense to host this under the http4s org.

PR #3190, opened 2020-02-12
Replace runtime reflection in HttpHeaderParser init

### LINKED ISSUE #3189: Graal: remove reflective header parsers
Based on @noelwelsh's fine work, [this commit](https://github.com/rossabaker/http4s-native-image/commit/28f7d65) shows a problem with modeled headers:

```
$ curl http://localhost:8080/raw
Some(Accept: */*)
$ curl http://localhost:8080/model
None
```

It works on the JVM:

```
$ curl http://localhost:8080/raw
Some(Accept: */*)
$ curl http://localhost:8080/model
Some(Accept: */*)
```

This is almost certainly a result of bootstrapping the header parsers with runtime reflection.  It's been a hidden implementation detail for years, but fails in the world beyond the JVM.

rossabaker:
Fixes #3189.  The Graalers will appreciate it.

### DIFF
--- a/core/src/main/scala/org/http4s/parser/HttpHeaderParser.scala
+++ b/core/src/main/scala/org/http4s/parser/HttpHeaderParser.scala
@@ -112,19 +112,6 @@ object HttpHeaderParser
   }
 
   private def gatherBuiltIn(): Unit = {
-    // This used to be based on reflection, which is bad for Graal, and
-    // made everyone queasy. To regenerate:
-    //
-    // HttpHeaderParser
-    //   .getClass
-    //   .getMethods
-    //   .filter(_.getName.forall(!_.isLower))
-    //   .sortBy(_.getName)
-    //   .map { method =>
-    //     val key = method.getName.replace('_', '-');
-    //     s"""addParser_("${key}".ci, `${method.getName}`)"""
-    //   }
-    //   .foreach(println)
     addParser_("ACCEPT".ci, `ACCEPT`)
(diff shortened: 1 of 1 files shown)


core/src/main/scala/org/http4s/parser/HttpHeaderParser.scala
```diff
@@ -110,15 +111,63 @@ object HttpHeaderParser
...
-      .foreach { method =>
-        val key = method.getName.replace('_', '-').ci
-        val parser = { value: String =>
-          method.invoke(this, value)
-        }.asInstanceOf[HeaderParser]
-
-        addParser(key, parser)
-      }
+  private def gatherBuiltIn(): Unit = {
+    // This used to be based on reflection, which is bad for Graal, and
```
cquiroz: Do we want to keep this commented code?

cquiroz: I think I misunderstood, that is code to generate the `addParser_` calls

rossabaker: I kept it to show people how to regenerate it if necessary, but I don't think it has been touched in a long time, so it could probably go.

rossabaker:
I confirmed that no headers were added after 0.20, so there should be no silent loss on merge.

rossabaker:
And also confirmed that it works on http4s-native-image.

PR #3192, opened 2020-02-13
Fix case sensitivity of SameSite

rossabaker:
Like those in #3184, `SameSite` is [case insensitive](https://tools.ietf.org/html/draft-ietf-httpbis-rfc6265bis-05#page-27).


hamnis:
build failure is just a flaky test again, unrelated

PR #3195, opened 2020-02-13
Fix scalafmt in CI

rossabaker:
When we switched from setup-java to setup-scala, we lost scalafmt.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,7 +33,7 @@ jobs:
           key: sbt-${{ hashFiles('**/build.sbt') }}
       - name: Check formatting
         if: startsWith(matrix.scala, '2.12.') && (matrix.java == 'adopt@1.8')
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -32,8 +32,8 @@ jobs:
           path: ~/.sbt
           key: sbt-${{ hashFiles('**/build.sbt') }}
       - name: Check formatting
-        if: startsWith(matrix.scala, '2.12.') && (matrix.java == '8')
-        run: sbt ++$SCALA_VERSION scalafmtCheckAll
+        if: startsWith(matrix.scala, '2.12.') && (matrix.java == 'adopt@1.8')
+        run: sbt ++$SCALA_VERSION test:scalafmt::test
```
rossabaker: This line needs to be reverted when we merge to series/0.21.  We would have a hell of a merge conflict if we upgraded scalafmt on 0.20.

PR #3196, opened 2020-02-13
Add Caching Convenience Functions

ChristopherDavenport:


server/src/main/scala/org/http4s/server/middleware/Caching.scala
```diff
@@ -111,36 +141,68 @@ object Caching {
...
+        for {
+          now <- HttpDate.current[G]
+          expires <- HttpDate
+            .fromEpochSecond(now.epochSecond + actualLifetime.toSeconds)
+            .liftTo[G]
+        } yield {
+          val headers = List(
+            `Cache-Control`(
+              NonEmptyList.of(
+                isPublic.fold[CacheDirective](identity, identity),
```
rossabaker: I think this might be `isPublic.merge`?

ChristopherDavenport:
Will follow up on merge, want to squeeze this in.

rossabaker:
Oops.  This missed the bus.  There's a series/0.21.  We already have a breaking change on master.

PR #3199, opened 2020-02-13
Uri fixes

### LINKED ISSUE #3197: Add a constructor for Uri that segments the given path
Right now I'm building a Uri and I have this:
```scala
def url(tenantId: TenantId) = 
  (uri"https://manage.office.com" / s"api/v1.0/${tenantId.string}/activity/feed/subscriptions/content")
    .withQueryParam("contentType", "Audit.AzureActiveDirectory")
    .withQueryParam("startTime", "2020-02-08T11:50")
    .withQueryParam("endTime", "2020-02-08T12:00")
```
It would be nice if I could do a builder-like pattern for the whole thing like so:
```scala
def url(tenantId: TenantId) =
  uri"https://manage.office.com"
    .addPath(s"api/v1.0/${tenantId.string}/activity/feed/subscriptions/content")
    .withQueryParam("contentType", "Audit.AzureActiveDirectory")
    .withQueryParam("startTime", "2020-02-08T11:50")
    .withQueryParam("endTime", "2020-02-08T12:00")
```

There is currently `withPath`, but it doesn't URL encode and I'm told it's meant to be a wrapper around the copy constructor.

It should be very easy to add.

hamnis:
Added scaladoc to Uri.withPath to describe behaviour
Be nice and add a / if the path does not have that and is non-empty.

Fixes #3197

### DIFF
--- a/core/src/main/scala/org/http4s/Uri.scala
+++ b/core/src/main/scala/org/http4s/Uri.scala
@@ -45,15 +45,27 @@ final case class Uri(
 
   def withoutFragment: Uri = copy(fragment = Option.empty[Fragment])
 
-  def addPath(newSegment: Path): Uri = {
-    val encoded = pathEncode(newSegment)
-    val newPath =
-      if (path.isEmpty || path.last != '/') s"$path/$encoded"
-      else s"$path$encoded"
-    copy(path = newPath)
-  }
+  /**
+    * Urlencodes and adds a path segment to the Uri
+    *
+    * @param newSegment the segment to add.
+    * @return a new uri with the segment added to the path
+    */
+  def addSegment(newSegment: Path): Uri = copy(path = toSegment(path, newSegment))
 
-  def /(newSegment: Path): Uri = addPath(newSegment)
+  /**
+    * This is an alias to [[addSegment(Path)]]
+    */
+  def /(newSegment: Path): Uri = addSegment(newSegment)
+
+  /**
+    * Splits the path segments and adds each of them to the path url-encoded.
+    * A segment is delimited by /
+    * @param morePath the path to add
+    * @return a new uri with the segments added to the path
+    */
+  def addPath(morePath: Path): Uri =
+    copy(path = morePath.split("/").foldLeft(path)((p, segment) => toSegment(p, segment)))
 
   def host: Option[Host] = authority.map(_.host)
   def port: Option[Int] = authority.flatMap(_.port)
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/http4s/Uri.scala
```diff
@@ -105,7 +111,14 @@ final case class Uri(
 
       case Uri(None, None, _, _, _) =>
     }
-    writer << path
+
+    this match {
```
rossabaker: What if there is a scheme and no authority?  I guess this behavior is scheme specific?

core/src/main/scala/org/http4s/Uri.scala
```diff
@@ -35,20 +35,26 @@ final case class Uri(
...
-  def /(newSegment: Path): Uri = {
+  def addPath(newSegment: Path): Uri = {
     val encoded = pathEncode(newSegment)
     val newPath =
       if (path.isEmpty || path.last != '/') s"$path/$encoded"
       else s"$path$encoded"
     copy(path = newPath)
   }
 
+  def /(newSegment: Path): Uri = addPath(newSegment)
```
rossabaker: We could document this as an alias.

core/src/main/scala/org/http4s/Uri.scala
```diff
@@ -35,20 +35,26 @@ final case class Uri(
...
+    * @param path the path string to replace
+    */
   def withPath(path: Path): Uri = copy(path = path)
 
   def withFragment(fragment: Fragment): Uri = copy(fragment = Option(fragment))
 
   def withoutFragment: Uri = copy(fragment = Option.empty[Fragment])
 
-  def /(newSegment: Path): Uri = {
+  def addPath(newSegment: Path): Uri = {
```
rossabaker: I like having a non-symbolic name.

hamnis:
> Any reason this can't be on series/0.20? I've been trying to backport bug fixes until we officially EOL it.
> 
> We talked about a short, but indefinite, wind down, but should put a date on it.

No, no reason. i can move it there.

core/src/main/scala/org/http4s/Uri.scala
hamnis: I do believe you need an authority for the path segmenting to make sense.

core/src/main/scala/org/http4s/Uri.scala
hamnis: yes

core/src/main/scala/org/http4s/Uri.scala
hamnis: i'm renaming this as `addSegment`

core/src/main/scala/org/http4s/Uri.scala
rossabaker: `file://` is the one that jumped to my mind.  But not everything has a `/`-rooted filesystem.

PR #3206, opened 2020-02-17
Model Content-Location header

cquiroz:
Another header modeled `Content-Location` which is basically the same as `Location`

### DIFF
--- a/core/src/main/scala/org/http4s/parser/HttpHeaderParser.scala
+++ b/core/src/main/scala/org/http4s/parser/HttpHeaderParser.scala
@@ -31,6 +31,7 @@ object HttpHeaderParser
     with AcceptLanguageHeader
     with AuthorizationHeader
     with CacheControlHeader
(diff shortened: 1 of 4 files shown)


core/src/main/scala/org/http4s/parser/HttpHeaderParser.scala
```diff
@@ -125,6 +126,7 @@ object HttpHeaderParser
     addParser_("CONTENT-DISPOSITION".ci, `CONTENT_DISPOSITION`)
     addParser_("CONTENT-ENCODING".ci, `CONTENT_ENCODING`)
     addParser_("CONTENT-LENGTH".ci, `CONTENT_LENGTH`)
+    addParser_("CONTENT-LOCATION".ci, `CONTENT_LOCATION`)
```
rossabaker: Good, you were paying attention to the new thing we have to do. :smile:

PR #3210, opened 2020-02-17
Model Content-Base Header

cquiroz:
Another quick header begin modeled as by #2011

### DIFF
--- a/tests/src/test/scala/org/http4s/parser/ContentBaseHeaderSpec.scala
+++ b/tests/src/test/scala/org/http4s/parser/ContentBaseHeaderSpec.scala
@@ -1,0 +1,32 @@
+package org.http4s
+package parser
+
+import org.http4s.headers.`Content-Base`
+
+class ContentBaseHeaderSpec extends Http4sSpec {
+  "ContentBaseHeader parser".can {
+    "Parse a simple uri" in {
+      val s = "http://www.foo.com"
+      val Right(uri) = Uri.fromString(s)
+      val hs = Headers.of(Header("Content-Base", s))
+
+      hs.get(`Content-Base`) must beSome(`Content-Base`(uri))
+    }
+
+    "Parse a simple uri with a path but no authority" in {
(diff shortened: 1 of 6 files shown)


tests/src/test/scala/org/http4s/parser/ContentBaseHeaderSpec.scala
```diff
@@ -0,0 +1,32 @@
...
+
+    "Parse a simple uri with a path but no authority" in {
+      val s = "http:/foo/bar"
+      val Right(uri) = Uri.fromString(s)
+      val hs = Headers.of(Header("Content-Base", s))
+
+      hs.get(`Content-Base`) must beSome(`Content-Base`(uri))
+    }
+
+    "Parse a relative reference" in {
```
rossabaker: I think this one is actually invalid?  The spec says it has to be an absolute URI.  Which is unfortunate, because we don't have that at the type level, so it will be easy to create an invalid one even if the parser rejects this.  Any ideas?

cquiroz:
I had forgotten about this, with the switch to cats-parse I guess it should be closed

rossabaker:
In the current branching, the only place to revive it is cats-effect-3-dotty, which is a shame.  If we merged dotty to main and made that cats-parse where breaking changes can be made, this would have a good home there.

rossabaker:
No longer breaking, but needs to be mostly rewritten for 0.22.

armanbilge:
I'm confused about this PR. `Content-Base` is not in the list in https://github.com/http4s/http4s/issues/5010. Furthermore, information about it seems to be scarce.

According to [RFC2616](https://www.w3.org/Protocols/rfc2616/rfc2616-sec19.html) `Content-Base` was deleted from the spec?

> Content-Base was deleted from the specification: it was not implemented widely, and there is no simple, safe way to introduce it without a robust extension mechanism. In addition, it is used in a similar, but not identical fashion in MHTML

Closing because I'm triaging aggressively.

PR #3216, opened 2020-02-22
Security policy

hamnis:
Added a first draft of a security policy.
Please critique as much as possible. 

All maintainers that want to be added to the list should let me know, or add themselves in the PR.

We should probably say something about announcing the security release. A security advisory is not enough. Maybe someone with a bit more info on how things are done can help with this.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -15,7 +15,8 @@ We are currently providing security updates to the following versions:
 ## Reporting a Vulnerability
 
 We will use [keybase](https://keybase.io) as the vehicle for reporting security issues as that gives us a
-much better feedback loop. We want to work with the reporter to make sure we understand the reported issue.
+forum to discuss, analyze, and remediate the threat before an exploit is published.
+[Responsible disclosure](https://en.wikipedia.org/wiki/Responsible_disclosure) enhances security for the entire community.
 
 If the vulnerability is found to be true we will release a patch version of our software
 and make sure that finds it way to Maven Central before we push the patch to github.
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -18,9 +18,10 @@ We will use [keybase](https://keybase.io) as the vehicle for reporting security 
 forum to discuss, analyze, and remediate the threat before an exploit is published.
 [Responsible disclosure](https://en.wikipedia.org/wiki/Responsible_disclosure) enhances security for the entire community.
 
-If the vulnerability is found to be true we will release a patch version of our software
+If issue is deemed a vulnerability, we will release a patch version of our software
 and make sure that finds it way to Maven Central before we push the patch to github.
 After the patch is available on Maven Central, we will also provide a [security advisory](https://github.com/http4s/http4s/security/advisories) through github.
+As with every release, the source jars are published to maven central at the same time as the binaries.
 
 We strongly recommend users of our libraries to use [Scala Steward](https://github.com/fthomas/scala-steward) or something similar to 
 automatically receive updates.
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -20,7 +20,7 @@ forum to discuss, analyze, and remediate the threat before an exploit is publish
 
 If the vulnerability is found to be true we will release a patch version of our software
 and make sure that finds it way to Maven Central before we push the patch to github.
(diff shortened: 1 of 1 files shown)


SECURITY.md
```diff
@@ -0,0 +1,33 @@
...
+| ------- | ------------------ |
+| 0.21.x   | :white_check_mark: |
+| 0.20.x   | :white_check_mark: |
+| 0.19.x   | :x: |
+| 0.18.x   | :x: |
+| < 0.18   | :x: |
+
+## Reporting a Vulnerability
+
+We will use [keybase](https://keybase.io) as the vehicle for reporting security issues as that gives us a
```
kevinmeredith: Why, I genuinely ask, do you prefer to use Keybase over just opening a new Github issue?

hamnis: Discussing security severe security problems in the open allows attackers to see that as well. Once there is a security patch there is usually a race between threat actors to exploit and users to upgrade. We want users to be able to upgrade before they become attacked.

hamnis: AFAIK there is no way to limit a github issue to just be visible to the reporter and the maintainers. That would be ideal. Using the tools that we already know and use day to day.

kevinmeredith: Thanks, @hamnis, for the explanation.

rossabaker: GitHub's Security feature gives us a private fork, which is helpful for sharing code and discussing the solution among project owners and invited parties.  I'm unclear whether the security advisory exposes draft-status comments to the public.  We're learning about this tool on the fly, but I think Keybase is useful at minimum for initial disclosure.

SECURITY.md
```diff
@@ -0,0 +1,33 @@
...
+much better feedback loop. We want to work with the reporter to make sure we understand the reported issue.
+
+If the vulnerability is found to be true we will release a patch version of our software
```
rossabaker: Vulnerability as a boolean is awkward phrasing.

```suggestion
If the issue is deemed a vulnerability, will release a patch version of our software
```

SECURITY.md
```diff
@@ -0,0 +1,33 @@
...
+and make sure that finds it way to Maven Central before we push the patch to github.
```
rossabaker: Is it worth mentioning that source jars are still available for every binary on Maven Central?

SECURITY.md
```diff
@@ -0,0 +1,33 @@
...
+| 0.21.x   | :white_check_mark: |
+| 0.20.x   | :white_check_mark: |
+| 0.19.x   | :x: |
+| 0.18.x   | :x: |
+| < 0.18   | :x: |
+
+## Reporting a Vulnerability
+
+We will use [keybase](https://keybase.io) as the vehicle for reporting security issues as that gives us a
+much better feedback loop. We want to work with the reporter to make sure we understand the reported issue.
```
rossabaker:
```suggestion
forum to discuss, analyze, and remediate the threat before an exploit is published.
[Responsible disclosure](https://en.wikipedia.org/wiki/Responsible_disclosure) enhances security for the entire community.
```

SECURITY.md
```diff
@@ -0,0 +1,33 @@
...
+
+If the vulnerability is found to be true we will release a patch version of our software
+and make sure that finds it way to Maven Central before we push the patch to github.
+After the patch is available on Maven Central, we will also provide a security advisory through github.
```
rossabaker: We can link:

```suggestion
After the patch is available on Maven Central, we will also provide a [security advisory](https://github.com/http4s/http4s/security/advisories) through github.
```

SECURITY.md
```diff
@@ -0,0 +1,35 @@
...
+| 0.18.x   | :x: |
+| < 0.18   | :x: |
+
+## Reporting a Vulnerability
+
+We will use [keybase](https://keybase.io) as the vehicle for reporting security issues as that gives us a
+forum to discuss, analyze, and remediate the threat before an exploit is published.
+[Responsible disclosure](https://en.wikipedia.org/wiki/Responsible_disclosure) enhances security for the entire community.
+
+If issue is deemed a vulnerability, we will release a patch version of our software
```
aeons:
```suggestion
If the issue is deemed a vulnerability, we will release a patch version of our software
```

PR #3221, opened 2020-02-25
Change Uri.Path to be own type

hamnis:
This changes Uri.Path to be an actual class to encapsulate Path segments.
We have two boolean flags here. These can be removed by transforming
the `Path` type to an own `ADT` hierarchy.

The main motivation for this change is to represent paths as something more than a String.

I have used `scala.collection.immutable.Vector` instead of `List` as `Vector` allows O(1)
appends.

This change is binary incompatible.

Typeclasses:
* Eq
* Semigroup
* Adding a monoid is possible, but there are two zero cases. Changing the Path type to an ADT will resolve that.

We have a few convenince functions here:

* startWith to be able to use in the Routers
* indexOf and splitAt to allow us to replace the caret stuff.
* normalize: drops empty segments
* toAbsolute and toRelative
* merge: using RFC3986 rules
* concat

Unrelated, add `out` to .gitignore to be able to use bloop from intellij

Representing Encoded Strings as just Strings are not really useful, so we have added a Segment class to have encoded values in.

* Use uri and path string interpolation in tests

Request changes:
* scriptName and pathInfo changed to be Uri.Path

### DIFF
--- a/core/src/main/scala/org/http4s/Message.scala
+++ b/core/src/main/scala/org/http4s/Message.scala
@@ -209,7 +209,7 @@ object Message {
   */
 final class Request[F[_]](
     val method: Method = Method.GET,
-    val uri: Uri = Uri(path = "/"),
+    val uri: Uri = Uri(path = Uri.Path.Root),
     val httpVersion: HttpVersion = HttpVersion.`HTTP/1.1`,
     val headers: Headers = Headers.empty,
     val body: EntityBody[F] = EmptyBody,
@@ -271,11 +271,14 @@ final class Request[F[_]](
     uri.path.splitAt(caret)
 
   private def caret =
-    attributes.lookup(Request.Keys.PathInfoCaret).getOrElse(0)
+    attributes.lookup(Request.Keys.PathInfoCaret).getOrElse(-1)
 
+  @deprecated(message = "Use {withPathInfo(Uri.Path)} instead", since = "1.0.0-M1")
   def withPathInfo(pi: String): Self =
+    withPathInfo(Uri.Path.fromString(pi))
+  def withPathInfo(pi: Uri.Path): Self =
     // Don't use withUri, which clears the caret
-    copy(uri = uri.withPath(scriptName + pi))
+    copy(uri = uri.withPath(scriptName.concat(pi)))
 
   def pathTranslated: Option[File] = attributes.lookup(Keys.PathTranslated)
 
@@ -467,7 +470,7 @@ final class Request[F[_]](
 object Request {
   def apply[F[_]](
       method: Method = Method.GET,
-      uri: Uri = Uri(path = "/"),
+      uri: Uri = Uri(path = Uri.Path.Root),
       httpVersion: HttpVersion = HttpVersion.`HTTP/1.1`,
       headers: Headers = Headers.empty,
       body: EntityBody[F] = EmptyBody,
--- a/core/src/main/scala/org/http4s/Uri.scala
+++ b/core/src/main/scala/org/http4s/Uri.scala
@@ -17,10 +17,12 @@ import java.net.{Inet4Address, Inet6Address, InetAddress}
 import java.nio.{ByteBuffer, CharBuffer}
 import java.nio.charset.{Charset => JCharset, StandardCharsets}
 import org.http4s.internal.{bug, hashLower}
+import cats.kernel.Semigroup
 import org.http4s.internal.parboiled2.{Parser => PbParser, _}
 import org.http4s.internal.parboiled2.CharPredicate.{Alpha, Digit, HexDigit}
 import org.http4s.parser._
 import org.http4s.util._
+
 import scala.collection.immutable
 import scala.math.Ordered
 import scala.reflect.macros.blackbox
@@ -33,26 +35,27 @@ import scala.reflect.macros.blackbox
   * @param query      optional Query. url-encoded.
   * @param fragment   optional Uri Fragment. url-encoded.
   */
-// TODO fix Location header, add unit tests
 final case class Uri(
     scheme: Option[Uri.Scheme] = None,
     authority: Option[Uri.Authority] = None,
-    path: Uri.Path = "",
+    path: Uri.Path = Uri.Path.empty,
     query: Query = Query.empty,
     fragment: Option[Uri.Fragment] = None)
     extends QueryOps
     with Renderable {
-  import Uri._
 
   /**
     * Adds the path exactly as described. Any path element must be urlencoded ahead of time.
     * @param path the path string to replace
     */
-  def withPath(path: Path): Uri = copy(path = path)
+  @deprecated("Use {withPath(Uri.Path)} instead", "1.0.0-M1")
+  def withPath(path: String): Uri = copy(path = Uri.Path.fromString(path))
 
-  def withFragment(fragment: Fragment): Uri = copy(fragment = Option(fragment))
+  def withPath(path: Uri.Path): Uri = copy(path = path)
 
-  def withoutFragment: Uri = copy(fragment = Option.empty[Fragment])
+  def withFragment(fragment: Uri.Fragment): Uri = copy(fragment = Option(fragment))
+
+  def withoutFragment: Uri = copy(fragment = Option.empty[Uri.Fragment])
 
   /**
     * Urlencodes and adds a path segment to the Uri
@@ -60,12 +63,12 @@ final case class Uri(
     * @param newSegment the segment to add.
     * @return a new uri with the segment added to the path
     */
-  def addSegment(newSegment: Path): Uri = copy(path = toSegment(path, newSegment))
+  def addSegment(newSegment: String): Uri = copy(path = toSegment(path, newSegment))
 
   /**
     * This is an alias to [[addSegment(Path)]]
     */
-  def /(newSegment: Path): Uri = addSegment(newSegment)
+  def /(newSegment: String): Uri = addSegment(newSegment)
 
   /**
     * Splits the path segments and adds each of them to the path url-encoded.
@@ -73,12 +76,12 @@ final case class Uri(
     * @param morePath the path to add
     * @return a new uri with the segments added to the path
     */
-  def addPath(morePath: Path): Uri =
+  def addPath(morePath: String): Uri =
     copy(path = morePath.split("/").foldLeft(path)((p, segment) => toSegment(p, segment)))
 
-  def host: Option[Host] = authority.map(_.host)
+  def host: Option[Uri.Host] = authority.map(_.host)
   def port: Option[Int] = authority.flatMap(_.port)
-  def userInfo: Option[UserInfo] = authority.flatMap(_.userInfo)
+  def userInfo: Option[Uri.UserInfo] = authority.flatMap(_.userInfo)
 
   def resolve(relative: Uri): Uri = Uri.resolve(this, relative)
 
@@ -117,7 +120,7 @@ final case class Uri(
     super.renderString
 
   override def render(writer: Writer): writer.type = {
-    def renderScheme(s: Scheme): writer.type =
+    def renderScheme(s: Uri.Scheme): writer.type =
       writer << s << ':'
 
     this match {
@@ -134,7 +137,7 @@ final case class Uri(
     }
 
     this match {
-      case Uri(_, Some(_), p, _, _) if p.nonEmpty && !p.startsWith("/") =>
+      case Uri(_, Some(_), p, _, _) if p.nonEmpty && !p.absolute =>
         writer << "/" << p
       case Uri(_, _, p, _, _) =>
         writer << p
@@ -142,7 +145,7 @@ final case class Uri(
 
     if (query.nonEmpty) writer << '?' << query
     fragment.foreach { f =>
-      writer << '#' << encode(f, spaceIsPlus = false)
+      writer << '#' << Uri.encode(f, spaceIsPlus = false)
     }
     writer
   }
@@ -154,13 +157,8 @@ final case class Uri(
 
   override protected def replaceQuery(query: Query): Self = copy(query = query)
 
-  private def toSegment(path: Path, newSegment: Path): Path = {
-    val encoded = pathEncode(newSegment)
-    val newPath =
-      if (path.isEmpty || path.last != '/') s"$path/$encoded"
-      else s"$path$encoded"
-    newPath
-  }
+  private def toSegment(path: Uri.Path, newSegment: String): Uri.Path =
+    path / Uri.Path.Segment(newSegment)
 }
 
 object Uri {
@@ -273,7 +271,6 @@ object Uri {
       }
   }
 
-  type Path = String
   type Fragment = String
 
   final case class Authority(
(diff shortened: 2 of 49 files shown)


core/src/main/scala/org/http4s/Uri.scala
```diff
@@ -304,6 +305,8 @@ object Uri {
       if (endsWithSlash) writer << "/" else writer
     }
 
+    override val renderString: String = super.renderString
```
rossabaker: What is going on here?  Eager caching?

core/src/main/scala/org/http4s/Uri.scala
```diff
@@ -317,6 +320,37 @@ object Uri {
...
+    def startsWith(path: Path): Boolean = {
+      val i = segments.iterator
+      val j = path.segments.iterator
+      while (j.hasNext && i.hasNext) if (i.next() != j.next())
+        return false
+
+      !j.hasNext
+    }
+
+    def startsWith_(path: String): Boolean = startsWith(Path.fromString(path))
```
rossabaker: Underscore to avoid overloading?

core/src/main/scala/org/http4s/Uri.scala
```diff
@@ -317,6 +320,37 @@ object Uri {
...
+
+    def indexOf(path: Path): Option[Int] = {
+      val iter = segments.iterator.zipWithIndex
+      val other = path.segments.iterator
+      var idx: Option[Int] = None
+      while (iter.hasNext && other.hasNext) {
```
rossabaker: I'm getting sleepy, but wouldn't this return the last indexOf instead of the first indexOf?

core/src/main/scala/org/http4s/Uri.scala
```diff
@@ -280,6 +279,110 @@ object Uri {
...
+    override def equals(obj: Any): Boolean = obj match {
+      case s: String => doEquals(Uri.Path.fromString(s))
+      case p: Path => doEquals(p)
+      case _ => false
+    }
+
+    private def doEquals(path: Path): Boolean =
+      this.segments == path.segments && path.absolute == this.absolute && path.endsWithSlash == this.endsWithSlash
+
+    override def hashCode(): Int = 31 * segments.hashCode()
```
rossabaker: Why not acccount for absolute and endsWithSlash here, too?  You don't _have_ to, but you could.

core/src/main/scala/org/http4s/Uri.scala
```diff
@@ -280,6 +279,110 @@ object Uri {
...
+
+    def render(writer: Writer): writer.type = {
+      val start = if (absolute) "/" else ""
+      writer << start << segments.mkString_("/")
```
rossabaker: Are these segments stored decoded?  I would think they need to be encoded here.

core/src/main/scala/org/http4s/Uri.scala
```diff
@@ -280,6 +279,110 @@ object Uri {
     }
   }
 
+  final class Path(
+      val segments: Chain[String],
+      val absolute: Boolean = false,
+      val endsWithSlash: Boolean = false)
```
rossabaker: Seems to me you could get away with an empty segment at the beginning to represent absolute, and an empty segment at the end to represent endsWithSlash, and not need three variables.  But maybe that complicates other logic.

rossabaker: Your idea of the ADT would move toward being able to represent absolute URIs at the type level, and could help solve the problem in #3210.  But it would need to carry forward to `Uri`.  Probably complicates things too much.  But ... maybe?

core/src/main/scala/org/http4s/syntax/StringSyntax.scala
```diff
@@ -11,4 +11,5 @@ trait StringSyntax {
 final class StringOps(val self: String) extends AnyVal {
   def ci: CaseInsensitiveString =
     CaseInsensitiveString(self)
+  def path: Uri.Path = Uri.Path.fromString(self)
```
rossabaker: We've been drifting toward StringContexts for most things.  Maybe better to have `path"/foo"` instead?

core/src/main/scala/org/http4s/Uri.scala
hamnis: yes

core/src/main/scala/org/http4s/Uri.scala
hamnis: yes

core/src/main/scala/org/http4s/Uri.scala
hamnis: possible. I need to add some tests here

core/src/main/scala/org/http4s/Uri.scala
hamnis: forgot that, so will add it

core/src/main/scala/org/http4s/Uri.scala
rossabaker: I like the impulse, but I think of the underscore suffix as being related to `Unit` (e.g., `traverse_`).  Would `startsWithString` be too verbose?

core/src/main/scala/org/http4s/Uri.scala
hamnis: they are stored encoded.

core/src/main/scala/org/http4s/Uri.scala
hamnis: Replaced Chain with List, so we got this for free

core/src/main/scala/org/http4s/Uri.scala
hamnis: I might be lying here. I need to add some tests to make sure we do this correctly

hamnis: They are indeed stored encoded.

core/src/main/scala/org/http4s/Uri.scala
```diff
@@ -279,6 +277,104 @@ object Uri {
...
+      val segments: List[String],
+      val absolute: Boolean,
+      val endsWithSlash: Boolean)
+      extends Renderable {
+
+    def isEmpty: Boolean = segments.isEmpty
+    def nonEmpty: Boolean = segments.nonEmpty
+
+    override def equals(obj: Any): Boolean = obj match {
+      case s: String => doEquals(Uri.Path.fromString(s))
```
rossabaker, line 290: This won't be symmetric.

core/src/main/scala/org/http4s/Uri.scala
```diff
@@ -279,6 +277,104 @@ object Uri {
...
+      case p: Path => doEquals(p)
+      case _ => false
+    }
+
+    private def doEquals(path: Path): Boolean =
+      this.segments == path.segments && path.absolute == this.absolute && path.endsWithSlash == this.endsWithSlash
+
+    override def hashCode(): Int =
+      31 * segments.hashCode() + System.identityHashCode(absolute) + System.identityHashCode(
```
rossabaker, line 299: Would get a better hash not just adding absolute to endsWithSlash.

core/src/main/scala/org/http4s/Uri.scala
```diff
@@ -279,6 +277,104 @@ object Uri {
     }
   }
 
+  final class Path private (
+      val segments: List[String],
```
rossabaker, line 281: I think you fixed some tests by abandoning `Chain`, but it seems like it was a good model.  `Path` appends will be common.  Maybe the constant factors are small enough `List` is better?

core/src/main/scala/org/http4s/Uri.scala
```diff
@@ -279,6 +277,104 @@ object Uri {
...
+      Some((path.segments, path.absolute, path.endsWithSlash))
+
+    def fromString(fromPath: String): Path = fromPath match {
+      case "" => empty
+      case "/" => Root
+      case pth =>
+        Path(
+          segments = pth
+            .split("/")
+            .filterNot(_.trim.isEmpty)
```
rossabaker, line 368: What in the model depends on this?  `//` is not semantically equivalent to `/` in a Uri, even if they're often the same.

core/src/main/scala/org/http4s/Uri.scala
hamnis: will remove it

core/src/main/scala/org/http4s/Uri.scala
hamnis: not entirely sure what you mean?

core/src/main/scala/org/http4s/Uri.scala
hamnis: We could use Vector, but Scala < 2.13.2 incurs a memory penalty on small vectors. Most Paths are pretty short < 10 segments, so I dont think that the cost of append might be that expensive

core/src/main/scala/org/http4s/Uri.scala
hamnis: Maybe they are not, but I have _never_ seen a uri with empty segments that was not a user mistake. 
If we cant remove empty segments here then perhaps we need to allow the pattern matching interpreter to allow to drop empty segments.

hamnis: Maybe its enough having the `normalize` method, so users can choose to this themselves.

hamnis: I tried removing it now, and most of the tests broke. So something is dependant on this, more investigation is needed.

hamnis: Ah, i understand. never mind, will fix

core/src/main/scala/org/http4s/Uri.scala
rossabaker: In this implementation, `absolute` and `!endsWithSlash` would hash the same as `!absolute` and `endsWithSlash`.  It's common to multiply the result by the prime each step of the way:

```scala
var hash = 7
hash = 31 * hash + segments.hashCode()
hash = 31 * hash + System.identityHashCode(absolute) 
hash = 31 * hash + System.identityHashCode(endsWithSlash)
hash
```

Those `System.identityHashCode`s might box.  `java.lang.Boolean.hashCode(b)` should deal with the primitives for you.

core/src/main/scala/org/http4s/Uri.scala
rossabaker: And random access will be rare.  As long as we're not concatenating them often, sounds good to me.

tests/src/test/scala/org/http4s/MessageSpec.scala
```diff
@@ -118,7 +118,7 @@ class MessageSpec extends Http4sSpec with Http4sLegacyMatchersIO {
 
       "preserve caret in withPathInfo" in {
         val originalReq = Request(
-          uri = Uri(path = "/foo/bar"),
+          uri = Uri().withPath("/foo/bar"),
```
rossabaker, line 121: Do we want a `path` interpolator for all these?

core/src/main/scala/org/http4s/Uri.scala
hamnis: yes, of course. Will fix

core/src/main/scala/org/http4s/Uri.scala
```diff
@@ -23,52 +25,52 @@ import scala.reflect.macros.blackbox
...
     extends QueryOps
     with Renderable {
-  import Uri._
 
   /**
     * Adds the path exactly as described. Any path element must be urlencoded ahead of time.
     * @param path the path string to replace
     */
-  def withPath(path: Path): Uri = copy(path = path)
+  def withPath(path: String): Uri = copy(path = Uri.Path.fromString(path))
```
rossabaker, line 41: I think deprecating this one would be fair.

core/src/main/scala/org/http4s/Uri.scala
```diff
@@ -279,6 +275,126 @@ object Uri {
...
+      case _ => false
+    }
+
+    private def doEquals(path: Path): Boolean =
+      this.segments == path.segments && path.absolute == this.absolute && path.endsWithSlash == this.endsWithSlash
+
+    override def hashCode(): Int = {
+      var hash = segments.hashCode()
+      hash += 31 * java.lang.Boolean.hashCode(absolute)
+      hash += 31 * java.lang.Boolean.hashCode(absolute)
```
rossabaker, line 298: I think this was supposed to be `endsWithSlash`.

core/src/main/scala/org/http4s/Uri.scala
```diff
@@ -279,6 +275,126 @@ object Uri {
...
+      override def equals(obj: Any): Boolean = obj match {
+        case s: Segment => s.encoded == encoded
+      }
+
+      override def hashCode(): Int = encoded.hashCode
+
+      override val toString: String = encoded
+    }
+
+    object Segment extends (String => Segment) {
```
rossabaker, line 356: We miss this function on a lot of case-classlike-things-that-aren't-case-classes.

core/src/main/scala/org/http4s/Uri.scala
```diff
@@ -279,6 +275,126 @@ object Uri {
...
+      val (start, end) = segments.splitAt(idx)
+      Path(start, absolute = absolute) -> Path(end, endsWithSlash = endsWithSlash)
+    }
+  }
+
+  object Path {
+    val empty = Path(Nil)
+    val Root = Path(Nil, absolute = true)
+
+    class Segment private (val encoded: String) {
```
rossabaker, line 344: I think this should be final.  Alternatively, the `sealed abstract case class` trick would save needing to override the `Any` methods.

core/src/main/scala/org/http4s/Uri.scala
```diff
@@ -279,6 +275,126 @@ object Uri {
...
+          segments = relative
+            .split("/")
+            .foldLeft(List.empty[Segment])((path, segment) =>
+              path ::: List(Segment.encoded(segment))),
+          absolute = absolute,
+          endsWithSlash = relative.endsWith("/")
+        )
+    }
+
+    implicit val eq: Eq[Path] = Eq.fromUniversalEquals[Path]
```
rossabaker, line 395: A lot more are possible here.  `Hash`, `Order`.  I think `Monoid`.  `LowerBounded`, if you want to get exotic.  But it's more important to get this landed, and we can circle back to those.

core/src/main/scala/org/http4s/Uri.scala
```diff
@@ -291,6 +289,132 @@ object Uri {
...
+    def apply(
+        segments: List[Segment],
+        absolute: Boolean = false,
+        endsWithSlash: Boolean = false): Path =
+      new Path(segments, absolute, endsWithSlash)
+
+    def unapply(path: Path): Some[(List[Segment], Boolean, Boolean)] =
+      Some((path.segments, path.absolute, path.endsWithSlash))
+
+    def fromString(fromPath: String): Path =
```
hamnis, line 397: rename this to apply, then let `object Path extends (String => Path)` ?
Ref @rossabaker comment above.

core/src/main/scala/org/http4s/Message.scala
```diff
@@ -269,14 +269,14 @@ final class Request[F[_]](
     )
 
   lazy val (scriptName, pathInfo) =
```
hamnis, line 271: What should we do about these?
We can make them into Uri.Paths and use the new splitAt(0), but then we need to change all routing to make sure it used the segment indices.
If we do that, then perhaps we need to change to a different backing collection type as random access might be more interesting

core/src/main/scala/org/http4s/Uri.scala
```diff
@@ -33,52 +36,53 @@ import scala.reflect.macros.blackbox
...
     extends QueryOps
     with Renderable {
-  import Uri._
 
   /**
     * Adds the path exactly as described. Any path element must be urlencoded ahead of time.
     * @param path the path string to replace
     */
-  def withPath(path: Path): Uri = copy(path = path)
+  @deprecated("Use {withPath(Uri.Path)} instead", "1.0.0-M1")
```
hamnis, line 52: since should be checked by someone

core/src/main/scala/org/http4s/Uri.scala
hamnis: Not sure if this semigroup passes all the laws. I guess a test is in order

core/src/main/scala/org/http4s/Uri.scala
rossabaker: Does that `{}` syntax do anything in a deprecation warning?  I sometimes backtick them.  Doesn't really matter. Mostly asking for my own curiosity.

core/src/main/scala/org/http4s/Uri.scala
rossabaker: My comment was more general, for all the times we avoid a case class so we can validate the argument.

I don't really like overloading `apply`.  I think I'd probably leave it as it is in this case.

core/src/main/scala/org/http4s/Uri.scala
rossabaker: I think you've even got a `Monoid` with `empty`.

We should decide if we want this to be an `Order` or not.  It could be.  I chose not to provide an `Order[Method]` in my recent reforms there, because I didn't know that alphabetical order was compelling.  It's perhaps a bit more useful to provide a stable ordering of URIs, of which this would be a part.

hamnis:
How should we handle 991d59b ? scalafix or equals hack?

core/src/main/scala/org/http4s/Uri.scala
hamnis: I`m not sure. For some reason i was thinking scaladoc syntax here. I can change it to backticks if that is what we have used elsewhere

core/src/main/scala/org/http4s/Uri.scala
hamnis: There are two monoids, both `Uri.Path.empty` and `Uri.Path.Root` 
Although you might simulate the `Root` one with adding a `toAbsolute` after and vise versa with `toRelative`

rossabaker: Oh. Hmm. I was thinking of `empty` as being the `empty`. Maybe we can sidestep the ambiguity by just providing `Semigroup`? I would need to think through the cases a bit more deeply.

rossabaker:
Is the issue in 991d59b that code is calling `path.equals(string)`, and it worked when it was a type alias and was `string.equals(string)`?  So it'll be a silent failure?  Damned universal equality.

I think an asymmetric equals, which violates both the laws of Scala and Java, would be more problematic in the long run.  We could perhaps log a warning in the equals method, but that's still a runtime solution.  I think a scalafix lint rule (or fix) is the best way forward.

rossabaker:
If it weren't `Uri.Path`, but just `Path`, would the situation be better?  We wouldn't be changing the equality of an existing type, but people are just going to change their imports and not think about the ramifications.  That _probably_ doesn't save us?

hamnis:
that would not help us :( Universal equality are biting us here, and it still would with a new type.
Logging on String would make it visible at least, even though runtime...

I agree that we want a scalafix, and I also think that it should be a different PR for that.

hamnis:
I would like a second approval on this.

rossabaker:
Sorry, I just gave you a merge conflict.  And will probably be giving you another, as I remove that internal implementation of `hashLower` from 1.0.

Issue #3229, opened 2020-03-03
Correct comments/docs about: HTTP connection is disposed after many client functions


kovacshuni:
Often displayed scaladoc comments and [in the API](https://http4s.org/v0.21/api/org/http4s/client/client) statements like:

_The underlying HTTP connection is closed at the completion of the decoding._

and variations of it don't seem to be true for me.

Checked with tcpdump, connection is maintained between multiple calls happening to the same host.
Tested this within a stream of HTTP requests-responses even set apart by 1 second gaps.

Also such implementation detail appearing on the interface `Client` is not a good practice. Connection handling should depend on the implementation (if this is not a very specific client aimed to close after each request) and comments should appear, for example, in the `blaze-client` subproject.

Tell me if i'm wrong. Even if I'm not seeing something or I didn't understand correctly, probably more people do the same and there should be some better clarification, documentation about how connections are maintained through multiple HTTP requests.

armanbilge:
Addressed in https://github.com/http4s/http4s/pull/6313.

PR #3230, opened 2020-03-03
Removed comments/scaladocs about HTTP connection is closing

kovacshuni:
Addresses https://github.com/http4s/http4s/issues/3229

Often displayed scaladoc comments and in the API statements like:

_The underlying HTTP connection is closed at the completion of the decoding._

were removed.

### DIFF
--- a/client/src/main/scala/org/http4s/client/Client.scala
+++ b/client/src/main/scala/org/http4s/client/Client.scala
@@ -19,9 +19,7 @@ trait Client[F[_]] {
   /** Submits a request, and provides a callback to process the response.
     *
     * @param req The request to submit
-    * @param f   A callback for the response to req.  The underlying HTTP connection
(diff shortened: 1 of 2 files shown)


client/src/main/scala/org/http4s/client/Client.scala
```diff
@@ -65,9 +58,7 @@ trait Client[F[_]] {
     * responsibility of callers of this service to run the response
     * body to dispose of the underlying HTTP connection.
     *
-    * This is intended for use in proxy servers.  `fetch`, `fetchAs`,
-    * [[toKleisli]], and [[streaming]] are safer alternatives, as their
-    * signatures guarantee disposal of the HTTP connection.
+    * This is intended for use in proxy servers.
```
rossabaker: I would keep a weaker form of the deleted text.  "as they allow the client to manage resources without forcing the caller to consume the body of each returned response."

PR #3240, opened 2020-03-10
Add Client#fetchOrError to Explicitly Account for Errors 

kevinmeredith:
Typically I code an algebra and HTTP interpreter in the following way: https://github.com/kevinmeredith/testing-http4s-example/blob/master/src/main/scala/net/User.scala.

I frequently use `adaptError` to handle raised errors, namely:

* client timed out
* failed to decode HTTP Response's body as `A`

As a result, I commonly resort to `adaptError` as in https://github.com/kevinmeredith/testing-http4s-example/blob/master/src/main/scala/net/User.scala#L45-L61.

```scala
    override def get(id: UUID): F[Option[ApplicationUser]] = {
      
      val uri: Uri = base / "users" / id.toString
      
      val req: Request[F] = Request[F](method = Method.GET, uri = uri)
      
      val response: F[Option[ApplicationUser]] = 
        httpClient.fetch[Option[ApplicationUser]](req) {
          case Status.Ok(body)    => body.as[ApplicationUser].map(Some(_))
          case Status.NotFound(_) => F.pure(None)
          case unexpectedStatus   => F.raiseError(UserError(uri, id, UnexpectedStatus(unexpectedStatus.status)))
        }

      MonadError[F, Throwable].adaptError(response) {
        case ue @ UserError(_,_,_) => ue
        case t                     => UserError(uri, id, t)
      }
```

I would find it useful to have such a method available in `org.http4s.Client`. Please let me know what you think of this method.

### DIFF
--- a/client/src/main/scala/org/http4s/client/Client.scala
+++ b/client/src/main/scala/org/http4s/client/Client.scala
@@ -1,15 +1,17 @@
 package org.http4s
 package client
 
-import cats.~>
+import cats.{MonadError, ~>}
 import cats.data.Kleisli
 import cats.effect._
 import cats.effect.concurrent.Ref
 import cats.implicits._
 import fs2._
 import java.io.IOException
+
 import org.http4s.headers.Host
 import org.http4s.syntax.kleisli._
+
 import scala.util.control.NoStackTrace
 
 /** A [[Client]] submits [[Request]]s to a server and processes the [[Response]]. */
@@ -35,6 +37,20 @@ trait Client[F[_]] {
     * @return The result of applying f to the response to req
     */
   def fetch[A](req: F[Request[F]])(f: Response[F] => F[A]): F[A]
+
+  /** Submits a request, and provides a callback to process the response.
+    * Note that all errors, e.g. decoding a payload, client times out, etc.
+    *
+    * @param req An effect of the request to submit
+    * @param f A callback for the response to req.  The underlying HTTP connection
+    *          is disposed when the returned task completes.  Attempts to read the
+    *          response body afterward will result in an error.
+    * @param handle A function that maps a Throwable to an F[A]
+    * @return The result of applying f to the response to req, while handling any raised errors
+    */
+  def fetchOrError[A](req: Request[F])(f: Response[F] => F[A])(handle: Throwable => F[A])(
+    implicit ev: MonadError[F, Throwable]
+  ): F[A]
 
   /**
     * Returns this client as a [[Kleisli]].  All connections created by this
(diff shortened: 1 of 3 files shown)


kevinmeredith:
Any thoughts on this one, please, @rossabaker or @ChristopherDavenport?

hamnis:
With regards to naming, we have `expectOr`, `expectOptionOr` etc. We should at least try to keep names in the same vein with new functions.

kevinmeredith:
Hi @hamnis - I have no preference in the name. Please let me which one you think is best.

kevinmeredith:
Hi @hamnis - just following up. Please let me know if you have a preferred name and I'll update the PR.

hamnis:
Hello. I have been a bit busy @ work, so sorry for the delay in response.

Thinking about this I wonder if we should rather have a `PartialFunction[Throwable, F[A]]` .  That would allow us to only catch those errors we care about instead of having to handle _all_ possible exceptions.

With regards to name, we dont really have a similar function. The signature of the other `*or` functions deals with `Response[F] to F[Throwable]` so the current name is probably fine.

kevinmeredith:
> Hello. I have been a bit busy @ work, so sorry for the delay in response.

👍 

> Thinking about this I wonder if we should rather have a PartialFunction[Throwable, F[A]] . That would allow us to only catch those errors we care about instead of having to handle all possible exceptions.

That sounds reasonable to me, @hamnis. I'll make that change.

> Perhaps fetchAndRecoverWith is what we really mean here?

That name change sounds good to me!

kevinmeredith:
Hi @hamnis - I'm confident that I've addressed your good comments via https://github.com/http4s/http4s/pull/3240#commits-pushed-65cf137. Please take a look! Thanks

client/src/main/scala/org/http4s/client/Client.scala
```diff
@@ -36,6 +38,21 @@ trait Client[F[_]] {
     */
   def fetch[A](req: F[Request[F]])(f: Response[F] => F[A]): F[A]
 
+  /** Submits a request, and provides a callback to process the response.
+    * Note that all errors, e.g. decoding a payload, client times out, etc.
+    *
+    * @param req An effect of the request to submit
```
rossabaker: This isn't an effect.

```suggestion
    * @param req The request to submit
```

kevinmeredith:
> What is the difference between this and client.fetch(req)(f).recoverWith(handle)?

I believe nothing, @rossabaker. I can see the argument of just using `client.fetch(req)(f).recoverWith(handle)` everywhere, i.e. and thus an argument towards closing this PR.

What do you think?

client/src/main/scala/org/http4s/client/Client.scala
kevinmeredith: Thanks, Ross, for correcting me.

rossabaker:
Yeah, I think the `fetch` and `recoverWith` compose nicely.  (In fact, I believe @ChristopherDavenport has argued even against `fetch`!)  I'm not opposed if people like it, but if in doubt, it's better to leave it out.

kevinmeredith:
> I'm not opposed if people like it, but if in doubt, it's better to leave it out.

👍

PR #3241, opened 2020-03-10
Prefer blackbox macros

travisbrown:
I noticed this while porting the macros to Dotty, and found [this comment](https://github.com/http4s/http4s/pull/2508#pullrequestreview-226336649):

> Also can't remember why these had to be whitebox instead of blackbox.

…but these definitely shouldn't need to be whitebox. Maybe there was an issue on 2.11? In any case tests pass with this change now.


travisbrown:
Just noticed that `illTyped` [can also be blackbox](https://github.com/milessabin/shapeless/issues/587).

PR #3242, opened 2020-03-10
Dotty preparation

travisbrown:
I've been experimenting with cross-building http4s (just core for now) on Dotty, with the goal of turning up major blockers early if there are any. So far [porting the necessary subset of Parboiled2's macros](https://github.com/sirthias/parboiled2/issues/210) has been the hardest part, and I have those close enough to working that I can compile most of http4s-core with a few changes to the parsers on this end (changes I've not made in this PR). The log4s macros are also a problem, but I just rewrote them in a commit in my actual Dotty cross-building branch. Porting http4s's own macros to Dotty's macro system was ~trivial, but again I've not included that in this PR.

This branch just takes care of a few kinds of changes that are needed for Dotty and that are improvements (or at least neutral) on Scala 2. I'll go through the changes in roughly increasing order of complicatedness.

I didn't run formatting because when I tried running it it made a lot of unrelated changes. I'm happy to run it on master and then rebase and format this branch if people would prefer.

#### Build config

I simplified the version-specific source configuration a little, which is possible now that http4s has dropped Scala 2.11 support. Before this change we have the following for e.g. 2.13:

```
[info] * /home/travis/code/projects/http4s/core/src/main/scala-2.13
[info] * /home/travis/code/projects/http4s/core/src/main/scala
[info] * /home/travis/code/projects/http4s/core/src/main/java
[info] * /home/travis/code/projects/http4s/core/src/main/scala-2.13-2.13
[info] * /home/travis/code/projects/http4s/core/src/main/scala-2.13
[info] * /home/travis/code/projects/http4s/core/src/main/java-2.13
```
Now it looks like this:
```
[info] * /home/travis/code/projects/http4s/core/src/main/scala-2.13
[info] * /home/travis/code/projects/http4s/core/src/main/scala
[info] * /home/travis/code/projects/http4s/core/src/main/java
```
I haven't added Dotty-specific source folders yet, just made things a little cleaner for when we do.

#### Parentheses for lambda parameters with types

Dotty doesn't accept e.g. `decode { json: Json =>`, so I've added parentheses to all of these, which I think looks nicers anyway.

#### Placeholders

I replaced kind-projector's `?` with the equivalent `*` everywhere, since `?` has a different meaning for Dotty and isn't supported as a placeholder by [`-Ykind-projector`](https://github.com/lampepfl/dotty/pull/7775). We've already done this in Cats, cats-effect, etc.

#### Polymorphic function value syntax

Dotty's `-Ykind-projector` supports the `λ` syntax in types, but not the value-level `λ` for defining polymorphic function values. It only appears once in this project, so I just expanded it manually, which adds a few dozen characters to one test.

#### Higher-kinded placeholders

Dotty's `-Ykind-projector` doesn't support the `?[_]` in constraints like `F[_]: MonadError[?[_], Throwable]`. There were only 13 of these, most of them in ember-core, so I just expanded them to the equivalent implicit parameters.

#### Type annotations

Dotty requires type annotations on implicit definitions, so I added one of these that was missing. I also added a type in one place in the `Response` companion object that Scala 2 can infer but Dotty can't.

#### Cats imports

In one place Dotty was complaining about cycles because of a `cats.implicits._` import. I'm guessing this will be resolved in an upcoming release by changes I made to get Cats cross-building, but it's easy enough to work around by changing this one import to the more specific `cats.syntax.either._`

#### Companion object imports

In another place Dotty was freaking out about cycles because we're importing `org.http4s.Uri._` at the top of the file where the `Uri` case class and companion object are defined, and then using imported types in the definition of the case class. I tried quickly to minimize the issue and wasn't able to, and I'm not sure it's worth the effort since this kind of seems like a weird thing to do anyway. I resolved it by just writing out the `Uri.` in four places.

#### Mixing constraint bounds and implicit parameters with defaults

This project turned up [a Dotty bug that I minimized and reported](https://github.com/lampepfl/dotty/issues/8491) that affects methods with mixed constraint bounds and implicit parameters when the implicit parameters have default values. It's likely to get fixed, but it only happens in two places here, and I think mixing constraint bounds and implicit parameters is a bad idea anyway (and likely to be worse on Dotty, where they might desugar a little differently), so I just rewrote the constraint bounds as implicit parameters in those two places.

#### API change

I did make one breaking change, to `PlayInstances#jsonEncoderOf`, which looked like this before:

```scala
  def jsonEncoderOf[F[_]: EntityEncoder[?[_], String], A: Writes]: EntityEncoder[F, A] =
    jsonEncoder[F].contramap[A](Json.toJson(_))
```
Note that the `EntityEncoder` instance isn't used here, although the compiler doesn't tell you that. When I rewrote the `?[_]` syntax (see the "Higher-kinded placeholders" section above for details), the compiler finally noticed that the parameter wasn't used. I'm happy to keep it around for now to avoid breaking bincompat and add a TODO or open an issue to track it—just let me know.


travisbrown:
Oh, I think the formatting issue was because I also ran `scalafmtSbt`, which doesn't seem to have been run lately, and isn't checked in CI. I just pushed the result of `scalafmtAll` alone.

hamnis:
Nice stuff. Well done @travisbrown . I have planned taking a crack on replacing all the `?` with `*` next weekend, so very nice that this is done here.

With regards to functions with a a single parameter, could we add a scalafix rule that handles that for us? That would allow us to keep the compatibility without having to run through dotty.

travisbrown:
@hamnis Thanks! I agree that it would be nice to have a Scalafix rule for unparenthesized typed parameters, but I've just been hoping (for Cats, etc.) that we'll have Dotty cross-building in CI sooner rather than later and won't have to worry about it. Also Dotty has a built in rewriting tool for that specifically, so if a few slip in after this initial big change it's not that disruptive to catch them and fix them.

totallhall64:
Can u tell me who started this whole project cuz they have hacked my phone want to know starters username

travisbrown:
@rossabaker I'm definitely happy to rebase this against `series/0.21`, and move the play-json change to a follow-up PR against master. Will do that today.

The locally-published parboiled2 I've been using for Dotty has been built from the http4s fork, which is probably a little nicer than migrating upstream would be, since I can have Dotty-specific code for the Shapeless parts (although so far I haven't done that). The future of parboiled2 on Dotty / Scala 3 is still an open question—see e.g. [Mathias's response here](https://github.com/sirthias/parboiled2/issues/210#issuecomment-596623167) and [my update](https://github.com/sirthias/parboiled2/issues/210#issuecomment-597505806).

travisbrown:
Superseded by #3253.

Issue #3236, opened 2020-03-09
Thread leak in async http client


djspiewak:
```scala
          Resource(F.async[(Response[F], F[Unit])] { cb =>
            c.executeRequest(toAsyncRequest(req), asyncHandler(cb))
            ()
          })
```

This is incorrect (in `AsyncHttpClient.scala`. You need to `guarantee` a `shift` following the `async`. Unfortunately, you don't have a `ContextShift` in scope here, only a `ConcurrentEffect`, which makes this a little tricky to correct. It has real consequences though: we're seeing heap dumps with compute running on the async-http-client worker pool. An amusing example:

<img width="1211" alt="Screen Shot 2020-03-08 at 9 19 43 PM" src="https://user-images.githubusercontent.com/752/76180100-9ca62b80-6182-11ea-8e57-b30753fab06e.png">

You'll notice that we managed to get part of the http4s-server machinery scheduled onto the async-http-*client* thread pool.

rossabaker:
We have fixed these issues in the past with name change and deprecation of the broken one.  One option is to introduce the builder pattern used on many other backends, but we only have one parameter for the builder.  

A second option is to introduce `stream0`, `resource0`, etc., but then people will need to switch, and switch back next release.  We've done that before on lesser used middlewares, but this would be the highest profile case of that.

djspiewak:
Actually I realized retroactively that we can fix this just using `ConcurrentEffect` by using the same trick as `invokeCallback`. I'll write that up.

PR #3255, opened 2020-03-17
j.u.c.CompletionStage conversions

amesgen:
- Rename `fromCompletableFuture` to `fromCompletionStage`, as we don't want to `complete` or `get` the given future. I removed the call to `cancel` (therefore, we only need `Async[F]` now), as it can be seen as a special case of `completeExceptionally` and therefore does not really seem cancel anything (but I don't really know much about Java futures, so please correct me if I am wrong).
 - Guarantee a shift in `fromCompletionStage` (motivated by #3236)
 - Add `unsafeToCompletionStage` (used in http4s-jdk-http-client) while we are at it.

These changes are internal and `fromCompletableFuture` is only used in http4s-jdk-http-client, so this should be fine.

### DIFF
--- a/core/src/main/scala/org/http4s/internal/package.scala
+++ b/core/src/main/scala/org/http4s/internal/package.scala
@@ -1,9 +1,9 @@
 package org.http4s
 
-import java.util.concurrent.{CancellationException, CompletableFuture, CompletionException}
-import java.util.function.BiFunction
+import java.util.concurrent.{CancellationException, CompletionException, CompletionStage}
 
-import cats.effect.{Async, Concurrent, ConcurrentEffect, Effect, IO}
+import cats.effect.implicits._
+import cats.effect.{Async, ConcurrentEffect, ContextShift, Effect, IO}
 import cats.implicits._
 import org.http4s.util.execution.direct
 import org.log4s.Logger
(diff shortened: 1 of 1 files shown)


core/src/main/scala/org/http4s/internal/package.scala
```diff
@@ -124,19 +129,32 @@ package object internal {
...
-            case ex: CompletionException if ex.getCause ne null => cb(Left(ex.getCause))
-            case ex => cb(Left(ex))
+  private[http4s] def fromCompletionStage[F[_], CF[x] <: CompletionStage[x], A](
+      fcs: F[CF[A]])(implicit F: Async[F], CS: ContextShift[F]): F[A] =
+    fcs.flatMap { cs =>
+      F.async[A] { cb =>
+          cs.handle[Unit] { (result, err) =>
+            err match {
+              case null => cb(Right(result))
+              case _: CancellationException => ()
```
amesgen: I am not really sure if it is correct to ignore `CancellationException`s, specially as we no longer "cancel" on our own.

core/src/main/scala/org/http4s/internal/package.scala
```diff
@@ -124,19 +129,32 @@ package object internal {
...
-      implicit F: Concurrent[F]): F[A] =
-    fcf.flatMap { cf =>
-      F.cancelable { cb =>
-        cf.handle[Unit](new BiFunction[A, Throwable, Unit] {
-          override def apply(result: A, err: Throwable): Unit = err match {
-            case null => cb(Right(result))
-            case _: CancellationException => ()
-            case ex: CompletionException if ex.getCause ne null => cb(Left(ex.getCause))
-            case ex => cb(Left(ex))
+  private[http4s] def fromCompletionStage[F[_], CF[x] <: CompletionStage[x], A](
```
rossabaker: We have been burned before removing package private.  I think you're correct, but we neglected rho once.  Maybe just out of an abundance of caution, we could deprecate the old one?

core/src/main/scala/org/http4s/internal/package.scala
rossabaker: It still seems reasonable to me: if a cats effect is canceled, it doesn't terminate.  I'm not sure that raising a Java error makes sense.

core/src/main/scala/org/http4s/internal/package.scala
```diff
@@ -124,19 +129,32 @@ package object internal {
...
+              case _: CancellationException => ()
+              case ex: CompletionException if ex.getCause ne null => cb(Left(ex.getCause))
+              case ex => cb(Left(ex))
+            }
           }
-        })
-        F.delay { cf.cancel(true); () }
-      }
+          ()
+        }
```
rossabaker: From the Javadoc on `cancel`:

>  Dependent CompletableFutures that have not already completed will also complete exceptionally,

I'm not sure whether dependent futures are completed exceptionally on other exceptions.  This may be a reason to do it when we know we have a `Future`.  I don't feel strongly.

amesgen: We also could to runtime casting stuff to detect `CompletableFuture`, but it is ugly. What about requiring `Concurrent` instead of `Async`, so that potential future changes to cancellation are not breaking?

amesgen:
I set the deprecation version to 0.21.2 for now.

core/src/main/scala/org/http4s/internal/package.scala
rossabaker: That sounds good to me.  We can always add an async version later if the need arises.

rossabaker:
We'll need to cherry pick this back to series/0.21 to make that version number correct, but I think we'll want to do that anyway for the jdk-http-client's sake.

amesgen:
This PR should be ready to be merged now. I think it makes sense to squash the commits.

rossabaker:
Cherry-picked into #3263.

PR #3257, opened 2020-03-17
Make SameSite cookie attribute optional

mrdziuban:
I mentioned this in gitter and figured I'd open a PR with the change. To get the behavior of `SameSite=None` in all browsers, it's necessary to completely exclude the `SameSite` attribute for [certain clients that handle it incorrectly](https://www.chromium.org/updates/same-site/incompatible-clients). This updates the field to be optional to allow for this.

### DIFF
--- a/core/src/main/scala/org/http4s/ResponseCookie.scala
+++ b/core/src/main/scala/org/http4s/ResponseCookie.scala
@@ -102,7 +102,7 @@ final case class ResponseCookie(
     maxAge: Option[Long] = None,
     domain: Option[String] = None,
     path: Option[String] = None,
-    sameSite: Option[SameSite] = Some(SameSite.Lax),
+    sameSite: Option[SameSite] = None,
     secure: Boolean = false,
     httpOnly: Boolean = false,
     extension: Option[String] = None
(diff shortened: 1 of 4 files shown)


mrdziuban:
Now that I see the tests this raises a somewhat interesting question -- is the correct default still `Some(SomeSite.Lax)`, including when parsing a cookie that doesn't include the `SameSite` attribute, or should it be `None` in any/all cases?

hamnis:
We try to avoid breaking clients, so this seems reasonable.
This is a breaking incompatible change, so not sure if we should target 0.21 for that. Maybe @rossabaker or @ChristopherDavenport has some input on this.

Generating Some(SameSite.Lax) would keep the most forward-compatible default, but allowing to opt-out.

mrdziuban:
Thanks @rossabaker, any thoughts on the default behavior? It feels wrong to me to use `Some(SameSite.Lax)` as the default when parsing, e.g. [in this case](https://github.com/http4s/http4s/blob/master/tests/src/test/scala/org/http4s/parser/CookieHeaderSpec.scala#L25-L29) -- if a client sends a cookie without the `SameSite` attribute, http4s should respect that right?

rossabaker:
I think that's true.  Once it becomes optional, we could and should respect its absence as `None`.

rossabaker:
A thought I'm embarrassed to bring up: if this is rendering is hugely problematic for people in 0.21, we could disable the feature with a system property.  

Or this version is not too binary incompatible yet, so we could at least offer a milestone in the next series once this is merged.

mrdziuban:
Updated the default to `None`. Either of those options (sys prop or milestone) would be fine for my needs!

PR #3261, opened 2020-03-19
Fixed connection release when body isn't run, as well as thread affinity

djspiewak:
This PR fixes three issues in the async-http-client implementation:

- When the `body` of the response was never sequenced, the upstream connection would never be closed and would ultimately end up stuck in TCP `CLOSE_WAIT`
- The fiber continuation of the `Response` was run on an AHC thread, rather than on the pool governed by the relevant `ContextShift`. In order to avoid breaking bincompat, this was fixed by forking a fiber to run the `async` callback
- Under certain circumstances, when the `body` of an HTTP was empty, the prior implementation could deadlock. This is fixed by adding a callback to `onComplete`. Note that, by law, Cats Effect `async` callbacks ignore repeated invocations, so it's safe to put this here without a latch to guard it.

It's worth noting that there is still a known issue with this fix: namely, if you start running the `body` stream and then the controlling fiber is almost immediately `cancel`ed, then the connection can leak. This can happen because we cannot uninterruptibly call `subscribe` *and* start the stream (which would register the finalizer that runs `cancel`). The only way to fix this is to change fs2-reactive-streams' `.stream` function to take an `F[Unit]` representing the subscription effect, which it would then sequence and manage itself in a properly bracketed fashion.

This case is very unlikely, and at least things are better than they were.

### DIFF
--- a/async-http-client/src/main/scala/org/http4s/client/asynchttpclient/AsyncHttpClient.scala
+++ b/async-http-client/src/main/scala/org/http4s/client/asynchttpclient/AsyncHttpClient.scala
@@ -94,7 +94,10 @@ object AsyncHttpClient {
 
           responseWithBody = response.copy(body = body)
 
-          _ <- F.delay(cb(Right(responseWithBody -> (dispose >> bodyDisposal.get.flatten))))
+          // use fibers to access the ContextShift and ensure that we get off of the AHC thread pool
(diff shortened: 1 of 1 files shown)


rossabaker:
Cherry-picked onto series/0.21.

PR #3262, opened 2020-03-20
Release v0.20.20

rossabaker:
Includes #3167 and available dependency bumps.


rossabaker:
The security patch took that number. Will start fresh for 0.20.21.

PR #3263, opened 2020-03-20
Release v0.21.2

rossabaker:


website/src/hugo/content/changelog.md
```diff
@@ -8,6 +8,35 @@ Maintenance branches are merged before each new release. This change log is
...
+
+* [#3196](https://github.com/http4s/http4s/pull/3196): Add convenience functions to `Caching` middleware. 
+
+## Noteworthy
+
+* [#3253](https://github.com/http4s/http4s/pull/3253): Preparation for Dotty support. Should be invisible to end users, but calling out because it touches a lot. 
+p
+## Dependency updates
+
+* cats-2.1.1
```
hamnis: There are some missing dependency updates here.

rossabaker: I usually don't mention the test-only dependencies, but I could if you think they're insightful.

hamnis: I was thinking of the different version bumps of as mentioned in the 0.20.x docs below. Putting on "just reading the release docs" glasses,  I didnt find it entirely obvious that they are also included in the 0.21.x release as well.

rossabaker: Ah, yeah.  That's always been awkward to express given the interleaved branches.  We have that note at the top that you should always read top down until the version you're coming from.  I don't know how to make it clear without being repetitive.

hamnis: Me neither. so I guess this as good as we can get atm.

amesgen:
[This commit](https://github.com/http4s/http4s/pull/3255/commits/e4cb02c427fde5155e654a1bd0b699b4025563a9) from #3255 does not seem to have been cherry-picked, is this intentional?

rossabaker:
That missing commit from #3255 is my mistake.  I am going to close this in light of the security patch, and fix that when I reopen for v0.21.3.

PR #3267, opened 2020-03-21
Upgrade to fs2-2.3.0 on 0.21.x.

rossabaker:
`fs2.compress` is deprecated, and its replacement requires a `Sync` constraint.  We can't fix this in a binary compatible fashion, so we introduce silencer to continue using the deprecated version.  A proper fix is in #3266 

In async-http-client, use the new `.stream(F[Unit])` signature.


rossabaker:
Would like to get @djspiewak's eyes on the async-http-client change, to ensure I used the new signature correctly.

async-http-client/src/main/scala/org/http4s/client/asynchttpclient/AsyncHttpClient.scala
```diff
@@ -86,11 +86,11 @@ object AsyncHttpClient {
...
           subscribeF = F.delay(publisher.subscribe(subscriber))
           bodyDisposal <- Ref.of[F, F[Unit]] {
-            subscribeF >> subscriber.stream.take(0).compile.drain
+            subscriber.stream(subscribeF).take(0).compile.drain
           }
 
-          body = Stream.eval(F.uncancelable(bodyDisposal.set(F.unit) >> subscribeF)) >>
-            subscriber.stream.flatMap(part => chunk(Chunk.bytes(part.getBodyPartBytes)))
+          body = Stream.eval(F.uncancelable(bodyDisposal.set(F.unit))) >>
+            subscriber.stream(subscribeF).flatMap(part => chunk(Chunk.bytes(part.getBodyPartBytes)))
```
djspiewak:
```suggestion
          body = subscriber.stream(
            bodyDisposal.set(F.unit) >> subscribeF).flatMap(part => chunk(Chunk.bytes(part.getBodyPartBytes)))
```

You need the whole thing to be handled atomically, rather than just the subscription. Otherwise, looks good!

rossabaker:
Test failure is #3293. Merging.

PR #3275, opened 2020-03-25
Release v0.20.21

rossabaker:
Just the dependency bumps that were left behind in #3262.


rossabaker:
If #3268 is ready, we can include it, but it's not a major deal to end users.

PR #3276, opened 2020-03-25
Release v0.21.3

rossabaker:


core/src/main/scala/org/http4s/internal/package.scala
```diff
@@ -124,19 +129,51 @@ package object internal {
     }
 
   // Adapted from https://github.com/typelevel/cats-effect/issues/160#issue-306054982
+  @deprecated("Use `fromCompletionStage`", since = "0.21.3")
```
rossabaker: I bumped the number here compared to the cherry-pick.

rossabaker:
Would be nice to include #3267 as well.

PR #3279, opened 2020-03-25
Improve Prometheus middleware usage example

### LINKED ISSUE #3278: Prometheus Metrics middleware example in documentation incomplete.
The documentation for the Prometheus metrics middleware found at [https://http4s.org/v0.21/middleware/](https://http4s.org/v0.21/middleware/ ) is incomplete insofar as it does not mention the final step needed to expose the metrics for scraping by a Prometheus server. The PrometheusExportService does exist, but that fact is non-obvious and requires source code spelunking to discover. It would be great if the example in the documentation included usage of PrometheusExportService, perhaps along these lines?

(from @hamnis in the Gitter):
```
def server[F[_]: Sync](api: HttpRoutes[F]): Resource[F, Server[F]] =
    for {
      registry <- Prometheus.collectorRegistry[F]
      metrics <- Prometheus.metricsOps[F](registry, "server")
      prometheus <- PrometheusExportService.build[F]
      routes = Router(
        "/api" -> Metrics[F](metrics)(api),
        "/" -> prometheus.routes
      )
      server <- BlazeServerBuilder[F]
        .bindHttp("9000", "0.0.0.0")
        .withHttpApp(routes.orNotFound)
        .resource
    } yield server
```

gannicottb:
Show how the PrometheusExportService can be used to expose the metrics for scraping. I spent a long time trying to figure it out, would like to pay it forward.

### DIFF
--- a/docs/src/main/tut/middleware.md
+++ b/docs/src/main/tut/middleware.md
@@ -206,16 +206,16 @@ import org.http4s.server.middleware.Metrics
 ```tut:book
 implicit val clock = Clock.create[IO]
 
-def meteredRouter(routes: HttpRoutes[IO]): Resource[IO, HttpRoutes[IO]] =
+val meteredRouter: Resource[IO, HttpRoutes[IO]] =
   for {
-    registry <- Prometheus.collectorRegistry[IO]
-    metrics <- Prometheus.metricsOps[IO](registry, "server")
     metricsSvc <- PrometheusExportService.build[IO]
+    metrics <- Prometheus.metricsOps[IO](metricsSvc.collectorRegistry, "server")
     router = Router[IO](
-      "/api" -> Metrics[IO](metrics)(routes),
+      "/api" -> Metrics[IO](metrics)(apiService),
       "/" -> metricsSvc.routes
     )
   } yield router
+  
 ```
 
 [service]: ../service
--- a/docs/src/main/tut/middleware.md
+++ b/docs/src/main/tut/middleware.md
@@ -197,19 +197,25 @@ We can create a middleware that registers metrics prefixed with a
 provided prefix like this.
 
 ```tut:silent
-import cats.effect.{Resource, IO}
+import cats.effect.{Clock, IO, Resource}
+import org.http4s.HttpRoutes
+import org.http4s.metrics.prometheus.{Prometheus, PrometheusExportService}
+import org.http4s.server.Router
 import org.http4s.server.middleware.Metrics
-import org.http4s.metrics.prometheus.Prometheus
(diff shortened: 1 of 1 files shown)


gannicottb:
fixes #3278

docs/src/main/tut/middleware.md
```diff
@@ -197,19 +197,25 @@ We can create a middleware that registers metrics prefixed with a
...
+import org.http4s.server.Router
 import org.http4s.server.middleware.Metrics
-import org.http4s.metrics.prometheus.Prometheus
-import io.prometheus.client.CollectorRegistry
 ```
 ```tut:book
 implicit val clock = Clock.create[IO]
 
-val meteredRoutes: Resource[IO, HttpRoutes[IO]] =
+def meteredRouter(routes: HttpRoutes[IO]): Resource[IO, HttpRoutes[IO]] =
```
hamnis, line 209: This should be called somewhere with the `apiService` now its just a function that is not used in this example.

gannicottb: Hm, I guess I thought getting the reader to a Resource[IO, HttpRoutes[IO]] would be sufficient (without writing an entire sample server).

hamnis: This is better, but we should still call this method with `apiService` to show use.
or we can change this back to use the `apiService` directly

docs/src/main/tut/middleware.md
```diff
@@ -197,19 +197,25 @@ We can create a middleware that registers metrics prefixed with a
...
   for {
     registry <- Prometheus.collectorRegistry[IO]
     metrics <- Prometheus.metricsOps[IO](registry, "server")
-  } yield Metrics[IO](metrics)(apiService)
+    metricsSvc <- PrometheusExportService.build[IO]
```
ChristopherDavenport, line 213: Doesn't this require the registry?

hamnis: @ChristopherDavenport The apply method requires the registry, the `build` methods creates a new registry.

ChristopherDavenport: In which case this would need to be first and be the registry used for metering.

docs/src/main/tut/middleware.md
gannicottb: I see what you mean. I'll make that change.

docs/src/main/tut/middleware.md
gannicottb: I totally missed the difference between build and apply. I can fix that up.

rossabaker:
Thanks!  Test failures are unrelated.

PR #3280, opened 2020-03-26
Message/Media Typeclass Core Compile

ChristopherDavenport:
I started running into trouble in the Http4sMatchers, everything works and is actually much nicer to use in my opinion. 

But I'm not sure whats going on in the tests thats not letting me compile.

### DIFF
--- a/core/src/main/scala/org/http4s/EntityDecoder.scala
+++ b/core/src/main/scala/org/http4s/EntityDecoder.scala
@@ -22,7 +22,7 @@ import scala.annotation.implicitNotFound
 trait EntityDecoder[F[_], T] { self =>
 
   /** Attempt to decode the body of the [[Message]] */
-  def decode(m: Media[F], strict: Boolean): DecodeResult[F, T]
+  def decode[M[_[_]]: Media](m: M[F], strict: Boolean): DecodeResult[F, T]
 
   /** The [[MediaRange]]s this [[EntityDecoder]] knows how to handle */
   def consumes: Set[MediaRange]
@@ -31,13 +31,13 @@ trait EntityDecoder[F[_], T] { self =>
   def map[T2](f: T => T2)(implicit F: Functor[F]): EntityDecoder[F, T2] = new EntityDecoder[F, T2] {
     override def consumes: Set[MediaRange] = self.consumes
 
-    override def decode(m: Media[F], strict: Boolean): DecodeResult[F, T2] =
+    def decode[M[_[_]]: Media](m: M[F], strict: Boolean): DecodeResult[F, T2] = 
       self.decode(m, strict).map(f)
   }
 
   def flatMapR[T2](f: T => DecodeResult[F, T2])(implicit F: Monad[F]): EntityDecoder[F, T2] =
     new EntityDecoder[F, T2] {
-      override def decode(m: Media[F], strict: Boolean): DecodeResult[F, T2] =
+      override def decode[M[_[_]]: Media](m: M[F], strict: Boolean): DecodeResult[F, T2] = 
         self.decode(m, strict).flatMap(f)
 
       override def consumes: Set[MediaRange] = self.consumes
@@ -66,7 +66,7 @@ trait EntityDecoder[F[_], T] { self =>
     new EntityDecoder[F, T2] {
       override def consumes: Set[MediaRange] = self.consumes
 
-      override def decode(m: Media[F], strict: Boolean): DecodeResult[F, T2] =
+      override def decode[M[_[_]]: Media](m: M[F], strict: Boolean): DecodeResult[F, T2] = 
         self.decode(m, strict).transform(t)
     }
 
@@ -82,7 +82,7 @@ trait EntityDecoder[F[_], T] { self =>
     new EntityDecoder[F, T2] {
       override def consumes: Set[MediaRange] = self.consumes
 
-      override def decode(m: Media[F], strict: Boolean): DecodeResult[F, T2] =
+      override def decode[M[_[_]]: Media](m: M[F], strict: Boolean): DecodeResult[F, T2] = 
         DecodeResult(
           F.flatMap(self.decode(m, strict).value)(r => f(r).value)
         )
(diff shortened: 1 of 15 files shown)


ChristopherDavenport:
All of core is compiling. I have no idea what is up with tests, so if anyone can point me in the right direction it would be massively appreciated, because at this point I expect the only changes necessary are some added imports.

core/src/main/scala/org/http4s/multipart/Part.scala
```diff
@@ -10,17 +10,22 @@ import java.io.{File, InputStream}
...
     headers.get(`Content-Disposition`).flatMap(_.parameters.get("filename"))
 
-  override def covary[F2[x] >: F[x]]: Part[F2] = this.asInstanceOf[Part[F2]]
+  def covary[F2[x] >: F[x]]: Part[F2] = this.asInstanceOf[Part[F2]]
 }
 
 object Part {
   private val ChunkSize = 8192
 
+  val media: Media[Part] = new Media[Part]{
```
hamnis: missing `implicit` here?

core/src/main/scala/org/http4s/Media.scala
```diff
@@ -1,65 +1,36 @@
...
 import fs2.Stream
 import fs2.text.utf8Decode
 import org.http4s.headers._
 import org.http4s.util.decode
 
-trait Media[F[_]] {
-  def body: EntityBody[F]
-  def headers: Headers
-  def covary[F2[x] >: F[x]]: Media[F2]
+trait Media[M[_[_]]]{
```
IndiscriminateCoding, line 10: Looks like `Media` is agnostic of exact `M` structure (i.e. it doesn't use the fact that it is parameterized by `F`) - so there is no need in *too-higher-kinded* type `M[_[_]]`.
May I suggest you to consider `trait Media[F[_], -T]`, so you can have implicit instances like `Media[F, Part[F]]` and `Media[F, M <: Message[F]]`?

IndiscriminateCoding: Moreover, it is possible to hide `F[_]` to a type member, so you would get something like haskell's associated types:

```scala
trait Media[T] {
  type F[_]

  def body(of: T): EntityBody[F]
  def headers(of: T): Headers
}

implicit def mediaPartInstance[S[_]]: Media[Part[S]] = new Media[Part[S]] {
  type F[a] = S[a]

  def body(of: T): EntityBody[F] = ???
  def headers(of: T): Headers = ??
}
```

async-http-client/src/main/scala/org/http4s/client/asynchttpclient/AsyncHttpClient.scala
```diff
@@ -23,6 +23,7 @@ import org.reactivestreams.Publisher
 import _root_.io.netty.handler.codec.http.cookie.Cookie
 import org.asynchttpclient.uri.Uri
 import org.asynchttpclient.cookie.CookieStore
+import org.http4s.implicits._
```
rossabaker, line 26: This is probably about the biggest breaking change that will come from this approach.  I'm okay with it, but we've recently had discussions about http4s' high import tax.  Also some hit to discoverability, but tooling has gotten a lot better since this was first proposed a long time ago.

If we do this, and remove some imports from the server and client DSLs, it's a winning tradeoff.

core/src/main/scala/org/http4s/EntityDecoder.scala
```diff
@@ -150,44 +150,52 @@ object EntityDecoder {
     * that recoverable errors are returned as a
     * [[DecodeResult.failure]], or that system errors are raised in `F`.
     */
-  def decodeBy[F[_]: Applicative, T](r1: MediaRange, rs: MediaRange*)(
-      f: Media[F] => DecodeResult[F, T]): EntityDecoder[F, T] = new EntityDecoder[F, T] {
-    override def decode(m: Media[F], strict: Boolean): DecodeResult[F, T] =
+  // def decodeBy[F[_]: Applicative, T](r1: MediaRange, rs: MediaRange*)(
+  //     f: M[F] => DecodeResult[F, T]): EntityDecoder[F, T] = new
+      
+  abstract class DecodeByMediaRange[F[_]: Applicative, T](r1: MediaRange, rs: MediaRange*) extends EntityDecoder[F, T] {
```
rossabaker, line 156: This seems like the other big breaking change.  Anybody who wrote their own decoders with `decodeBy` is losing.  I can't think of a way out of this.

core/src/main/scala/org/http4s/Media.scala
```diff
@@ -1,65 +1,36 @@
 package org.http4s
 
-import cats.MonadError
-import cats.implicits._
+// import cats.MonadError
+// import cats.implicits._
```
rossabaker, line 4: Delete.

core/src/main/scala/org/http4s/Media.scala
rossabaker: Contravariant typeclasses are a bit unusual.  Is it just to support `Message`?  If that's all, I could see either eliminating `Message` or making it an extension of the `Media` typeclass.

core/src/main/scala/org/http4s/MessageFailure.scala
```diff
@@ -83,8 +85,10 @@ final case class MalformedMessageBodyFailure(details: String, cause: Option[Thro
     s"Malformed message body: $details"
 
   def toHttpResponse[F[_]](httpVersion: HttpVersion): Response[F] =
-    Response(Status.BadRequest, httpVersion)
-      .withEntity(s"The request body was malformed.")(EntityEncoder.stringEncoder[F])
+    Message[Response].withEntity(
```
rossabaker, line 88: Checking my own understanding: you're just avoiding syntax here, and the old version would have worked with implicits?

core/src/main/scala/org/http4s/Media.scala
IndiscriminateCoding: @rossabaker ah, you are right, no need in contravariance here. It still should be possible to write instance for all subclasses of `Message[F]` without it.

PR #3286, opened 2020-03-28
httpRoutes constructor for AutoSlash middleware

voidcontext:
This PR is part of #2402.

I am opening this PR to start some discussions about the solution, although tests are passing (after 7-8 rerun of the github actions) but I am not sure about the implementation and the value of the feature.

Please see comments below.

### DIFF
--- a/server/src/main/scala/org/http4s/server/middleware/AutoSlash.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/AutoSlash.scala
@@ -2,7 +2,7 @@ package org.http4s
 package server
 package middleware
 
-import cats.{Functor, MonoidK}
+import cats.{Functor, Monad, MonoidK}
 import cats.data.Kleisli
 import cats.implicits._
 
@@ -29,4 +29,10 @@ object AutoSlash {
       }
     }
   }
+
+  def httpRoutes[F[_]: Monad](httpRoutes: HttpRoutes[F]): HttpRoutes[F] =
+    apply(httpRoutes)
+
+  def httpApp[F[_]: MonoidK: Functor](httpApp: HttpApp[F]): HttpApp[F] =
+    apply(httpApp)
 }
--- a/server/src/main/scala/org/http4s/server/middleware/AutoSlash.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/AutoSlash.scala
@@ -32,7 +32,4 @@ object AutoSlash {
 
   def httpRoutes[F[_]: Monad](httpRoutes: HttpRoutes[F]): HttpRoutes[F] =
     apply(httpRoutes)
-
-  def httpApp[F[_]: MonoidK: Functor](httpApp: HttpApp[F]): HttpApp[F] =
-    apply(httpApp)
 }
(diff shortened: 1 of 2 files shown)


server/src/main/scala/org/http4s/server/middleware/AutoSlash.scala
```diff
@@ -29,4 +29,10 @@ object AutoSlash {
       }
     }
   }
+
+  def httpRoutes[F[_]: Monad](httpRoutes: HttpRoutes[F]): HttpRoutes[F] =
+    apply(httpRoutes)
+
+  def httpApp[F[_]: MonoidK: Functor](httpApp: HttpApp[F]): HttpApp[F] =
```
voidcontext: I am not sure how valuable this constructor is, given the fact that `F` still should have a `MonoidK` instance and in this case `F` is usually `IO` which  - at least as far as I know - doesn't have that.

server/src/test/scala/org/http4s/server/middleware/AutoSlashSpec.scala
```diff
@@ -53,5 +55,26 @@ class AutoSlashSpec extends Http4sSpec with Http4sLegacyMatchersIO {
...
+      val req = Request[IO](uri = uri("/ping/"))
+      val httpApp: HttpApp[IO] = HttpApp { r =>
+        import org.http4s.dsl.io._
+        if (r.pathInfo == "/ping") Ok()
+        else throw new Exception("Cannot handle request!")
+      }
+
+      implicit val invalidMonoidKIO: MonoidK[IO] = new MonoidK[IO] {
+        def combineK[A](x: IO[A], y: IO[A]): IO[A] = IO.ioSemigroupK.combineK[A](x, y)
+        override def empty[A]: IO[A] = IO.raiseError(new Exception("error"))
```
voidcontext: This `MonoidK[IO]` instance doesn't feel to be right, it serves the purpose of the test, but I don't think it's correct.

server/src/main/scala/org/http4s/server/middleware/AutoSlash.scala
hamnis: Agreed. this constructor is not useful

rossabaker: Does it need to be a `MonoidK`, or just a `SemigroupK`?  `IO` has one of those (as seen below), though whether it's useful is still dubious.  "Try again with the slash if I raise an error," that would say.

server/src/test/scala/org/http4s/server/middleware/AutoSlashSpec.scala
rossabaker: `Future.filter` implies something like this (it throws a `NoSuchElementException`), but I think is a wart in the standard library.  Requiring something like this is a great indication that what we're testing is not right.

server/src/main/scala/org/http4s/server/middleware/AutoSlash.scala
rossabaker: Oh, I see the `F.empty` in `apply` now.  Yeah, I think we remove this, keep `httpRoutes`, and declare victory.

rossabaker:
Sorry this one slipped through the cracks for a while.  0.20 is about to be EOL.  If we hurry, we can put it there.  Otherwise, it can be introduced in 0.21.

server/src/main/scala/org/http4s/server/middleware/AutoSlash.scala
voidcontext: Ok, I'll update this PR asap.

voidcontext:
And sorry for the delay on my side as well. The `httpApp` constructor has beeen removed.

rossabaker:
Backported onto series/0.21.

PR #3291, opened 2020-04-01
Remove type parameter from `org.http4s.server.Server`

satorg:
From `server/src/main/scala/org/http4s/server/Server.scala`:
```
abstract class Server[F[_]] {
  ...
}
```
As per [conversation](https://gitter.im/http4s/http4s-dev?at=5e7ffe79426d5b06bf92c340) with @rossabaker, this type parameter looks like a legacy and isn't actually used anymore.

### DIFF
--- a/blaze-server/src/main/scala/org/http4s/server/blaze/BlazeBuilder.scala
+++ b/blaze-server/src/main/scala/org/http4s/server/blaze/BlazeBuilder.scala
@@ -162,7 +162,7 @@ class BlazeBuilder[F[_]](
   def withBanner(banner: immutable.Seq[String]): Self =
     copy(banner = banner)
 
-  def resource: Resource[F, Server[F]] = {
+  def resource: Resource[F, Server] = {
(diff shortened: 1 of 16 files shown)


satorg:
BTW, the following tests are failing on my local machine even without my changes:
* org.http4s.client.jetty.JettyClientSpec
* org.http4s.client.okhttp.OkHttpClientSpec

rossabaker:
Hmm, it's much more common for tests to fail on CI than locally.  If you paste the errors, I might be able to make a suggestion.

That said, this is almost there.  We just need to clean up the reference in the tut.  [Link to error](https://github.com/http4s/http4s/pull/3291/checks?check_run_id=551250403#step:8:281).

satorg:
Fixed. It seems it is fine now)

satorg:
@rossabaker:
> Hmm, it's much more common for tests to fail on CI than locally. If you paste the errors, I might be able to make a suggestion.

I have created a PR (#3294) which describes the issue and suggests a fix for it:
Could you take a look please?

PR #3303, opened 2020-04-04
Add caching mechanism for the date header value

jordiolivares:
I was upgrading the http4s code in the Techempower Benchmark and got curious to see where some of the bottlenecks are.

One of them was the fact that the HTTP Date header is always calculated and rendered for each request. This means that if you receive a lot of requests in a second each time you're going to render the Date header.

To solve this, this PR adds a simple caching mechanism that only renders the date header once in a given second, reusing the same value for all subsequent requests in that same moment.

### DIFF
--- a/blaze-core/src/main/scala/org/http4s/blazecore/Http1Stage.scala
+++ b/blaze-core/src/main/scala/org/http4s/blazecore/Http1Stage.scala
@@ -7,13 +7,15 @@ import fs2._
 import fs2.Stream._
 import java.nio.ByteBuffer
 import java.time.Instant
+
 import org.http4s.blaze.http.parser.BaseExceptions.ParserException
 import org.http4s.blaze.pipeline.{Command, TailStage}
 import org.http4s.blaze.util.BufferTools
 import org.http4s.blaze.util.BufferTools.emptyBuffer
 import org.http4s.blazecore.util._
 import org.http4s.headers._
-import org.http4s.util.{StringWriter, Writer}
+import org.http4s.util.{Renderer, StringWriter, Writer}
+
 import scala.concurrent.{ExecutionContext, Future}
 import scala.util.{Failure, Success}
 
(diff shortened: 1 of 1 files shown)


jordiolivares:
Yes, I tested the improvement with a JMH benchmark and obtained these results:

```
[info] Benchmark                              Mode  Cnt        Score        Error  Units
[info] Http4sDateFormatting.http4sWay        thrpt    3  1769821.455 ±   6346.432  ops/s
[info] Http4sDateFormatting.simplyCachedWay  thrpt    3  8865904.658 ± 459379.561  ops/s
```

The benchmark code was this short snippet:

```
@State(Scope.Thread)
class Http4sDateFormatting {
  @Benchmark
  def http4sWay(): Unit = {
    val rr = new StringWriter(512)
    rr << Date.name << ": " << Instant.now << "\r\n"
  }

  val SimplyCachedRenderer: Renderer[Instant] =
    new Renderer[Instant] {
      private val dateFormat =
        DateTimeFormatter
          .ofPattern("EEE, dd MMM yyyy HH:mm:ss zzz")
          .withLocale(Locale.US)
          .withZone(ZoneId.of("GMT"))

      var cachedString: String = dateFormat.format(Instant.now())
      var currentTime: Long = Instant.now().getEpochSecond

      def format(instant: Instant): String = {
        if (instant.getEpochSecond != currentTime) {
          currentTime = instant.getEpochSecond
          cachedString = dateFormat.format(instant)
        }
        cachedString
      }

      override def render(writer: Writer, t: Instant): writer.type =
        writer << format(t)
    }

  @Benchmark
  def simplyCachedWay(): Unit = {
    val rr = new StringWriter(512)
    rr << Date.name << ": "
    rr.<<(Instant.now)(SimplyCachedRenderer)
    rr << "\r\n"
  }
}
```

rossabaker:
Cherry-picked for release into 0.21.x.

Issue #1863, opened 2018-05-15
New method for WebSocketBuilder taking Pipe[F, WebSocketFrame, WebSocketFrame]


asoltysik:
What's your opinion about another build method that looks like this:
``` scala
 def build(
        pipe: Pipe[F, WebSocketFrame, WebSocketFrame]
        headers: Headers = Headers.empty,
        onNonWebSocketRequest: F[Response[F]] =
          Response[F](Status.NotImplemented).withEntity("This is a WebSocket route.").pure[F],
        onHandshakeFailure: F[Response[F]] = Response[F](Status.BadRequest)
          .withEntity("WebSocket handshake failed.").pure[F])
```

Similarly how it's done in akka-http. I believe it is enough for many use cases of websockets, can simplify the code and it's easier to reason about for these use cases.

rossabaker:
As an additional builder or as the only builder?

What does it mean?  The input to the pipe are received frames and the output is sent frames?

SystemFw:
I'm actually thinking that a model based on a Pipe might be cleaner than what we have now, here's a small snippet discussing translating back and forth between the two:
```scala
case class Builder(pipe: Pipe[F, Frame, Frame])
case class Split(in: Sink[F, Frame], out: Stream[F, Frame])

def splitToBuilder(s: Split) = Builder(input =>  s.out.concurrently(input.to(s.in))
```
```scala
case class Builder(in: Sink[F, Frame], out: Stream[F, Frame])
case class Piped(pipe: Pipe[F, Frame, Frame])

def pipedToBuilder(p: Piped): F[Builder] = 
needs to create the queue, and also quite possibly wrap everything in Option to use `unNoneTerminate`, 
without it you get race conditions like the one in Chris example that requires `sleep`
```

jmcardon:
p.s I am :+1: on this, in that reacting to messages is often what you want

rossabaker:
This exists now.

PR #3325, opened 2020-04-14
Implement WebSocket stream with a single send-receive pipe

Q69K:
This kind of API is much more easy to use and it is kinda superset of old API - as you see I can implement old API via new, but it's very hard to do new API with an old one.
Related to http4s/http4s#1863

### DIFF
--- a/core/src/main/scala/org/http4s/websocket/WebSocket.scala
+++ b/core/src/main/scala/org/http4s/websocket/WebSocket.scala
@@ -8,14 +8,17 @@ package org.http4s.websocket
 
 import fs2._
 
-private[http4s] final case class WebSocket[F[_]](
+private[http4s] sealed trait WebSocket[F[_]] {
+  def onClose: F[Unit]
+}
+
+private[http4s] final case class WebSocketSeparatePipe[F[_]](
     send: Stream[F, WebSocketFrame],
     receive: Pipe[F, WebSocketFrame, Unit],
     onClose: F[Unit]
-) {
-  @deprecated("Parameter has been renamed to `send`", "0.18.0-M7")
-  def read: Stream[F, WebSocketFrame] = send
+) extends WebSocket[F]
 
-  @deprecated("Parameter has been renamed to `receive`", "0.18.0-M7")
-  def write: Pipe[F, WebSocketFrame, Unit] = receive
-}
+private[http4s] final case class WebSocketCombinedPipe[F[_]](
+    receiveSend: Pipe[F, WebSocketFrame, WebSocketFrame],
+    onClose: F[Unit]
+) extends WebSocket[F]
--- a/server/src/main/scala/org/http4s/server/websocket/WebSocketBuilder.scala
+++ b/server/src/main/scala/org/http4s/server/websocket/WebSocketBuilder.scala
@@ -9,83 +9,128 @@ package org.http4s.server.websocket
 import cats.Applicative
 import cats.implicits._
 import fs2.{Pipe, Stream}
-import org.http4s.websocket.{WebSocket, WebSocketContext, WebSocketFrame}
+import org.http4s.websocket.{
+  WebSocket,
+  WebSocketCombinedPipe,
+  WebSocketContext,
+  WebSocketFrame,
+  WebSocketSeparatePipe
+}
 import org.http4s.{Headers, Response, Status}
 
 /**
   * Build a response which will accept an HTTP websocket upgrade request and initiate a websocket connection using the
   * supplied exchange to process and respond to websocket messages.
-  * @param send     The send side of the Exchange represents the outgoing stream of messages that should be sent to the client
-  * @param receive  The receive side of the Exchange is a sink to which the framework will push the incoming websocket messages
-  *                 Once both streams have terminated, the server will initiate a close of the websocket connection.
-  *                 As defined in the websocket specification, this means the server
-  *                 will send a CloseFrame to the client and wait for a CloseFrame in response before closing the
-  *                 connection, this ensures that no messages are lost in flight. The server will shutdown the
-  *                 connection when it receives the `CloseFrame` message back from the client. The connection will also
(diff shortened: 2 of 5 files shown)


Q69K:
look like you have unstable test :((

rossabaker:
Yes, the test failure should be addressed by #3338.

core/src/main/scala/org/http4s/websocket/WebSocket.scala
```diff
@@ -1,15 +1,23 @@
...
     receive: Pipe[F, WebSocketFrame, Unit],
     onClose: F[Unit]
-) {
-  @deprecated("Parameter has been renamed to `send`", "0.18.0-M7")
-  def read: Stream[F, WebSocketFrame] = send
-
-  @deprecated("Parameter has been renamed to `receive`", "0.18.0-M7")
-  def write: Pipe[F, WebSocketFrame, Unit] = receive
+) extends WebSocket[F] {
+  override def receiveSend: Pipe[F, WebSocketFrame, WebSocketFrame] =
```
rossabaker, line 16: It seems like we could duplicate less of the builder functions by doing this in `build`, and then call `buildWithSinglePipe`?

server/src/main/scala/org/http4s/server/websocket/WebSocketBuilder.scala
```diff
@@ -52,22 +88,25 @@ object WebSocketBuilder {
...
+            WebSocketContext(
+              WebSocketSeparatePipe(send, finalReceive, onClose),
+              headers,
+              onHandshakeFailure))
         )
     }
   }
 
-  def apply[F[_]: Applicative]: Builder[F] = new Builder[F]
+  def apply[F[_]: Concurrent]: Builder[F] = new Builder[F]
```
rossabaker, line 109: This is probably the most controversial change.  It's why this will be a binary-breaking change, even though `F` will almost always be `Concurrent` in practice.

I think we should either upgrade the constraint on the constructor to `Concurrent`, or move the `Applicative` constraint to `buildWithSinglePipe`, the way it is on `build`.  Our builders do more of the former, though I've never convinced myself one style is superior to the other.

core/src/main/scala/org/http4s/websocket/WebSocket.scala
```diff
@@ -1,15 +1,23 @@
 package org.http4s.websocket
 
+import cats.effect.Concurrent
 import fs2._
 
-private[http4s] final case class WebSocket[F[_]](
+sealed trait WebSocket[F[_]] {
```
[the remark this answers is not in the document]
Q69K: oh, not really, will fix that

Q69K:
@rossabaker thanks for the review, will take a look in the upcoming weekend

server/src/main/scala/org/http4s/server/websocket/WebSocketBuilder.scala
Q69K: I eliminated dependency on `Concurrent` by moving `concurrently` construction back to `Http4sWSStage`

core/src/main/scala/org/http4s/websocket/WebSocket.scala
Q69K: Made something of a currying with overloaded pair of `build` methods

Q69K:
For the rationale of the convenience of such construction, I want to bring my one example.

> A pretty simple message exchange
> So a client connects to a websocket server selects a room and then receives/sends messages.

New solution:
It's very simple to build a single pipe `Pipe[F, WebSocketFrame, WebSocketFrame]` that is just a `flatMap`/`mapEval` over incoming messages.
I also can imagine any other actor-like definition of a websocket handler will naturally fit a single pipe.

Current solution:
And for now I have to define two pipes and somehow bind them with some kind of queue or other concurrent stuff. I really found this task too difficult for me and decided to build this PR.

P.S. A little design note
Before, it was two pipes `send` and `receive`, and they were just combined with `concurrently` inside of `Http4sWSStage` to a single pipe - this seems a decision that is delayed from implementation from now.

core/src/main/scala/org/http4s/websocket/WebSocket.scala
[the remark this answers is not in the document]
Q69K: fixed)

amesgen:
@Q69K I find your motivation very compelling, it is the reason why we also did not use the "send pipe, receive stream" model in [http4s-jdk-http-client's websocket interface](https://jdk-http-client.http4s.org/stable/index.html#websocket-client).

In http4s-jdk-http-client, we used a connection model, e.g. you can open a connection via `WSRequest => Resource[F, WSConnection]` and then interact with the [`WSConnection`](https://www.javadoc.io/static/org.http4s/http4s-jdk-http-client_2.12/0.3.0/org/http4s/client/jdkhttpclient/WSConnection.html). This is very similar to the [TCP interface in fs2](https://oss.sonatype.org/service/local/repositories/releases/archive/co/fs2/fs2-io_2.12/2.2.1/fs2-io_2.12-2.2.1-javadoc.jar/!/fs2/io/tcp/index.html). Both the current WS interface in http4s and your proposed model can be implemented on top of it.

I think this PR is a nice simplification for common WS interactions; I just wanted to mention this, as in the future (in particular when we move a websocket client abstraction to http4s-core), it would be nice to have a shared base abstraction for both client and server. But this would/will require some refactoring either way, so I think it is fine to introduce a new abstraction for now.

rossabaker:
Needs a scalafmt.

@amesgen, do you still like the direction here?  I think you're our resident WebSocket expert.

amesgen:
I think this is a nice goody. I think it would be a good fit for `BlazeWebSocketExample.scala`, can you update it to use your new WS builder? @Q69K @klpx

The implementation looks fine, and I don't have that much time right now for working on a unified WS client + server interface, so this PR should be good to go then.

Q69K:
@amesgen updated the example. Now looks much better, thanks for the suggestion ;-)

Issue #3304, opened 2020-04-06
WebjarService broken on Windows for assets with more than 3 path components


fthomas:
I think [this change](https://github.com/http4s/http4s/commit/464060da6c146e3b96a65843f62168e9658c4eb1#diff-be6c03bb38414d6fc44166edc99f910aR98) (which was part of the fix for GHSA-66q9-f7ff-mmx6) has partially broken the `WebjarService` on Windows. A `Path` with more than three components will result in a `WebjarAsset` where the `asset` field contains backslashes as path separator. The `asset` field is then later used to load a resource which fails because the resource name contains backslashes. For example, trying to load the webjar resource `bootstrap/3.3.7/js/bootstrap.js` results in a `WebjarAsset("bootstrap", "3.3.7", "js\bootstrap.js")` which is then later used to load the resource `"META-INF/resources/webjars/bootstrap/3.3.7/js\bootstrap.js"`.

I guess a simple fix could be to use
```scala
import scala.jdk.CollectionConverters._
asset.iterator().asScala.mkString("/")
```
instead of `asset.toString` so we're not relying on the OS-specific separators.

rossabaker:
I don't have a way to test this.  Do you want to try a PR, or would you be willing to retest a fix?  Ideally, we'd target series/0.20.  (I don't know that we want to dig up 0.18 yet again for this, since it's not a security flaw, per se.)

fthomas:
I just confirmed that the fix above works for me. I'll prepare a PR.

PR #3326, opened 2020-04-14
WebjarService: Do not use OS-specific directory seperators

fthomas:
`Path#toString` uses the OS-specific directory separator if a `Path`
contains of multiple components. On Windows this means that the
`WebjarAsset#asset` field might contain backslashes. The `assets` field
is then later used to load a resource which [fails because of the
backslashes](https://docs.oracle.com/javase/8/docs/api/java/lang/ClassLoader.html#getResources-java.lang.String-).

This change always uses `/` as directory separator for the assets path
so that loading them works on all operating systems again.

Closes: #3304

### DIFF
--- a/server/src/main/scala/org/http4s/server/staticcontent/WebjarService.scala
+++ b/server/src/main/scala/org/http4s/server/staticcontent/WebjarService.scala
@@ -88,7 +88,7 @@ object WebjarService {
   /**
     * Returns an Option(WebjarAsset) for a Request, or None if it couldn't be mapped
     *
-    * @param subPath The request path without the prefix
+    * @param p The request path without the prefix
     * @return The WebjarAsset, or None if it couldn't be mapped
     */
   private def toWebjarAsset(p: Path): Option[WebjarAsset] = {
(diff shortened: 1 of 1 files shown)


server/src/main/scala/org/http4s/server/staticcontent/WebjarService.scala
```diff
@@ -88,21 +88,33 @@ object WebjarService {
...
+    * We're not using scala.jdk.CollectionConverters (which was added in 2.13)
+    * or scala.collection.convert.ImplicitConversion (which was deprecated in 2.13)
+    * to ease cross-building against multiple Scala versions.
+    */
+  private def asScalaIterator[A](underlying: java.util.Iterator[A]): Iterator[A] =
+    new Iterator[A] {
+      override def hasNext: Boolean = underlying.hasNext
+      override def next(): A = underlying.next()
+    }
+
```
rossabaker, line 117: This could probably be moved to `CollectionCompat`, in case it's useful elsewhere, but I don't think that's introduced until 0.21.

fthomas: Oh, I didn't noticed `CollectionCompat`. It exists on the `series/0.20` branch. On `series/0.21` it also has a `CollectionConverters` that could be used here.

PR #3330, opened 2020-04-15
Added executionContext as implicit parameter to BlazeBuilder.apply

albertoadami:
Right now the BlazeBuilder.apply is using global as ExecutionContext by default.
In my opinion, is better to pass executionContext as an implicit parameter.
Please note that the official documentation is not totally clear about that, and from my point of view running an http4s server with global as the execution context in production is not correct.

### DIFF
--- a/blaze-server/src/test/scala/org/http4s/server/blaze/BlazeServerMtlsSpec.scala
+++ b/blaze-server/src/test/scala/org/http4s/server/blaze/BlazeServerMtlsSpec.scala
@@ -12,6 +12,7 @@ import org.http4s.{Http4sSpec, HttpApp}
 import scala.concurrent.duration._
 import scala.io.Source
 import scala.util.Try
+import scala.concurrent.ExecutionContext.Implicits.global
 
 /**
   * Test cases for mTLS support in blaze server
@@ -27,9 +28,8 @@ class BlazeServerMtlsSpec extends Http4sSpec {
   }
 
   def builder: BlazeServerBuilder[IO] =
-    BlazeServerBuilder[IO]
+    BlazeServerBuilder[IO](global)
       .withResponseHeaderTimeout(1.second)
-      .withExecutionContext(testExecutionContext)
 
   val service: HttpApp[IO] = HttpApp {
     case req @ GET -> Root / "dummy" =>
--- a/blaze-server/src/test/scala/org/http4s/server/blaze/BlazeServerMtlsSpec.scala
+++ b/blaze-server/src/test/scala/org/http4s/server/blaze/BlazeServerMtlsSpec.scala
@@ -12,7 +12,7 @@ import org.http4s.{Http4sSpec, HttpApp}
 import scala.concurrent.duration._
 import scala.io.Source
 import scala.util.Try
-import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.ExecutionContext.global
 
 /**
   * Test cases for mTLS support in blaze server
--- a/blaze-server/src/main/scala/org/http4s/server/blaze/BlazeBuilder.scala
+++ b/blaze-server/src/main/scala/org/http4s/server/blaze/BlazeBuilder.scala
@@ -165,7 +165,7 @@ class BlazeBuilder[F[_]](
   def resource: Resource[F, Server] = {
     val httpApp = Router(serviceMounts.map(mount => mount.prefix -> mount.service): _*).orNotFound
     var b = BlazeServerBuilder
-      .apply(F, timer, executionContext)
+      .apply[F]
       .bindSocketAddress(socketAddress)
       .withIdleTimeout(idleTimeout)
       .withNio2(isNio2)
@@ -223,13 +223,10 @@ class BlazeBuilder[F[_]](
 
 @deprecated("Use BlazeServerBuilder instead", "0.20.0-RC1")
 object BlazeBuilder {
-  def apply[F[_]](
-      implicit F: ConcurrentEffect[F],
-      timer: Timer[F],
-      executionContext: ExecutionContext): BlazeBuilder[F] =
+  def apply[F[_]](implicit F: ConcurrentEffect[F], timer: Timer[F]): BlazeBuilder[F] =
     new BlazeBuilder(
       socketAddress = ServerBuilder.DefaultSocketAddress,
-      executionContext = executionContext,
+      executionContext = ExecutionContext.global,
       idleTimeout = IdleTimeoutSupport.DefaultIdleTimeout,
       isNio2 = false,
       connectorPoolSize = channel.DefaultPoolSize,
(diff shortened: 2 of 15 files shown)


hamnis:
Being explicit with `ExecutionContext`'s is a Good Thing (tm) IMHO

albertoadami:
@hamnis from my point of view is more correct to use something like the [BlazeClientBuilder.apply](https://github.com/http4s/http4s/blob/master/blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala#L279).

Do you not agree?
Or at least the ExecutionContext should be more documented on the official documentation, from my point of view isn't totally clear that the default execution context is the global right now. 

Like a http4s user, I'd prefer to construct the server with

`BlazeServerBuilder[IO](executionContext)

than with:

```
        BlazeServerBuilder[IO]
          .withExecutionContext(executionContext)
```

rossabaker:
I think Scala style has generally evolved toward using implicits for instances and syntax, and away from using it for dependencies.  Certainly so in our FP corner of the universe.  But `ExecutionContext` has a long tradition of being passed implicitly.  So I think the tension here is about sticking to idiomatic usage of this specific type vs. correcting a historical mistake.

I wouldn't ever want to explicity call it `BlazeServerBuilder[IO](executionContext)`, where it will be tied in with instance parameters.  I think it's more of a question as to whether the default should be `global` vs. whatever implicit is in scope at the point of construction.

SystemFw:
> In my opinion, is better to pass executionContext as an implicit parameter.

I think this should be put in context. In a pure FP application, there will generally be no implicit ECs flying around, since everything is written in `IO` which manages pools differently than `Future` (for which expecting an implicit `Ec` would feel way more idiomatic than the default). 
So, often,

> whatever implicit is in scope at the point of construction.

amounts to none. In an ideal world we should be able to take an `Ec` out of `ContextShift` so that users don't have to worry about this implementation detail of `Blaze`, but since that is not possible, what do we gain by _not_ defaulting to `global`?

rossabaker:
That's a good point.  We'd break a lot of `IOApp`s unless we made it an implicit with a default argument, which neither FP idiom nor standard library idiom endorse.  The main beneficiaries of this change would be apps that create a non-default, implicit `ExecutionContext` and have incorrectly assumed it's implicitly threaded to blaze.

albertoadami:
From my point of view passing ExecutionContext as an implicit parameter is more complaint with the actual behavior of the `BlazeClientBuilder.apply`.

albertoadami:
@rossabaker if I understand correctly your comment, the changes that I need to do are the following:
1. mark the actual apply implementation as @deprecated
2. add a `BlazeServerBuilder.apply(executionContext)` method with the explicit executionContext and an other one with the implicit argument(basically what I've done in this PR)


If it's totally correct what I'm saying I could work on this changes in the next couple of days.

albertoadami:
I have some integrations about what I wrote above:

1. also `BlazeServerBuilder.withExecutionContext` should become a `@deprecated` method?
2. I think that the `BlazeServerBuilder.apply` cannot be overloaded using different kind of implicit parameters, so we need to decide if use an explicit argument or implicit.

albertoadami:
@rossabaker I refactor some code, I see that there are some error on CI on specific java versions, can you help me on this?

albertoadami:
@rossabaker I removed the `@deprecated `annotation from the `withExecutionContext` method.
I think that the implicit `executionContext` as argument is not necessary, the client is doing the same thing right now.

About the #3350 and #3338 errors, should I merge the master branch inside this one?

rossabaker:
Oops, I missed that you'd already deprecated `withExecutionContext` when I approved last night.

I'm going to need to cherry-pick this onto a maintenance branch anyway.  It will be a little easier if we don't have a merge in it.  You could rebase if you like, or we can just ignore those failures we already know to be fixed.  Whatever is easier for you.

I think this is ready as soon as we fix the docs.  I wouldn't be surprised if there are a couple more, since tut fails fast:

```
[tut] *** Error reported at /home/runner/work/http4s/http4s/docs/src/main/tut/json.md:277
<console>:44: warning: method apply in object BlazeServerBuilder is deprecated (since 0.20.22): Use BlazeServerBuilder.apply with explicit executionContext instead
       val server = BlazeServerBuilder[IO].bindHttp(8080).withHttpApp(jsonApp).resource
```

rossabaker:
Close!

```
[tut] compiling: /home/runner/work/http4s/http4s/docs/src/main/tut/service.md
[tut] *** Error reported at /home/runner/work/http4s/http4s/docs/src/main/tut/service.md:135
// <console>:37: warning: method apply in object BlazeServerBuilder is deprecated (since 0.20.22): Use BlazeServerBuilder.apply with explicit executionContext instead
//        val serverBuilder = BlazeServerBuilder[IO].bindHttp(8080, "localhost").withHttpApp(httpApp)
```

albertoadami:
@rossabaker now should be all ok.

blaze-server/src/test/scala/org/http4s/server/blaze/BlazeServerMtlsSpec.scala
```diff
@@ -12,6 +12,7 @@ import org.http4s.{Http4sSpec, HttpApp}
 import scala.concurrent.duration._
 import scala.io.Source
 import scala.util.Try
+import scala.concurrent.ExecutionContext.Implicits.global
```
satorg, line 15: If it's expected to use `global` explicitly only, it's better to
```scala
import scala.concurrent.ExecutionContext.global
```
although I'd recommend to
```scala
```
and refer to `global` in a more clear way: `BlazeServerBuilder[IO](ExecutionContext.global)`

satorg: As I can see, some code below imports the implicit `global`, some other – the explicit one. But I cannot figure out the difference. Is seems to me that `global` is referenced explicitly everywhere, isn't it?

rossabaker: That's true.  They point to the same thing, and we probably imported from `Implicits` when this ticket originally set out to make it implicit.  But it would be marginally better if it just came from `ExecutionContext.global`.  I can mop that up when I backport it to the right branch, unless @albertoadami beats me to it.

albertoadami: @rossabaker  I can do the changes directly to this PR here, it's a 5 minutes work.
 I think that's better to use the `scala.concurrent.ExecutionContext.global`.

albertoadami: I changed the global import inside the tests, but now I get a test error on **Scala 2.12.11, Java adopt@1.8**, but shouldn't be due to my commit.

rossabaker:
When this gets its second signoff, eaaf50b0434cc1808723f09a657170aff19dbcdb will apply cleanly to the series/0.20 branch so we can include it in 0.20.22.

rossabaker:
This is now on the series/0.20 branch.

PR #3333, opened 2020-04-17
Add http2 support in JettyBuilder

rolandomanrique:
Enables HTTP2CServerConnectionFactory in JettyConnector which only supports http2 over plain text connections.

For http2 over ssl this would require ALPN configuration on the SSL context.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -260,7 +260,8 @@ lazy val jetty = libraryProject("jetty")
   .settings(
     description := "Jetty implementation for http4s servers",
     libraryDependencies ++= Seq(
-      jettyServlet
+      jettyServlet,
+      jettyHttp2Server
     )
   )
   .dependsOn(servlet % "compile;test->test", theDsl % "test->test")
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -260,7 +260,8 @@ lazy val jetty = libraryProject("jetty")
   .settings(
     description := "Jetty implementation for http4s servers",
     libraryDependencies ++= Seq(
-      jettyServlet
+      jettyServlet,
+      jettyHttp2Server
```
rossabaker, line 264: This introduces a new dependency that people won't use by default, which is not great, but I don't want to create another Jetty module, and optional dependencies are frustrating.  I think this is the least of all evils.

rossabaker:
I don't think the additional dependencies here are scary, and the release cycle is the same as the extant dependencies.  I think we can cherry pick this one to a maintenance branch to get it released sooner.

rossabaker:
Backported to series/0.20.

PR #3338, opened 2020-04-19
Avoid incorrectly responding with an empty body in AHC client

### LINKED ISSUE #3293: Empty body with async-http-client
After upgrading to `0.21.2` I am seeing a lot of empty body errors while using async-http-client as the client backend.

```
Caused by: org.http4s.MalformedMessageBodyFailure: Malformed message body: Invalid JSON: empty body
	at org.http4s.jawn.JawnInstances$.defaultJawnEmptyBodyMessage(JawnInstances.scala:42)
	at org.http4s.jawn.JawnInstances.jawnEmptyBodyMessage(JawnInstances.scala:17)
	at org.http4s.jawn.JawnInstances.jawnEmptyBodyMessage$(JawnInstances.scala:16)
	at org.http4s.circe.package$.jawnEmptyBodyMessage(package.scala:3)
	at org.http4s.circe.CirceInstances.$anonfun$jsonDecoderByteBufferImpl$1(CirceInstances.scala:41)
	at scala.util.Either.flatMap(Either.scala:352)
	at cats.data.EitherT.$anonfun$subflatMap$1(EitherT.scala:365)
	at scala.Function1.$anonfun$andThen$1(Function1.scala:85)
	at cats.effect.IO$Map.apply(IO.scala:1504)
	at cats.effect.IO$Map.apply(IO.scala:1502)
	at cats.effect.internals.IORunLoop$.cats$effect$internals$IORunLoop$$loop(IORunLoop.scala:142)
```

I suspect that something in #3261 causes the body to be emptied before it is read in some cases?

wemrysi:
1. Guards the callback invocation in `onCompleted` so that it is only called if `onStream` wasn't called.
2. Fixes `bodyDisposal` to actually sequence the stream finalizer via `.pull.uncons.void`. `take(0)` turns out to [be a noop](https://github.com/functional-streams-for-scala/fs2/blob/v1.0.0/core/shared/src/main/scala/fs2/Stream.scala#L3446), so the previous version wasn't finalizing.

Fixes #3293

### DIFF
--- a/async-http-client/src/main/scala/org/http4s/client/asynchttpclient/AsyncHttpClient.scala
+++ b/async-http-client/src/main/scala/org/http4s/client/asynchttpclient/AsyncHttpClient.scala
@@ -79,9 +79,12 @@ object AsyncHttpClient {
       var state: State = State.CONTINUE
       var response: Response[F] = Response()
       val dispose = F.delay { state = State.ABORT }
+      val onStreamCalled = Ref.unsafe[F, Boolean](false)
 
       override def onStream(publisher: Publisher[HttpResponseBodyPart]): State = {
         val eff = for {
+          _ <- onStreamCalled.set(true)
+
           subscriber <- StreamSubscriber[F, HttpResponseBodyPart]
 
           subscribeF = F.delay(publisher.subscribe(subscriber))
@@ -122,9 +125,13 @@ object AsyncHttpClient {
       override def onThrowable(throwable: Throwable): Unit =
         invokeCallback(logger)(cb(Left(throwable)))
 
-      // it's okay to invoke this repeatedly since repeated async callbacks are dropped by law
       override def onCompleted(): Unit =
-        invokeCallback(logger)(cb(Right(response -> dispose)))
+        onStreamCalled.get
+          .ifM(
+            F.unit,
+            F.delay(invokeCallback(logger)(cb(Right(response -> dispose)))))
+          .runAsync(_ => IO.unit)
+          .unsafeRunSync()
     }
 
   private def toAsyncRequest[F[_]: ConcurrentEffect](request: Request[F]): AsyncRequest = {
--- a/async-http-client/src/main/scala/org/http4s/client/asynchttpclient/AsyncHttpClient.scala
+++ b/async-http-client/src/main/scala/org/http4s/client/asynchttpclient/AsyncHttpClient.scala
@@ -88,8 +88,9 @@ object AsyncHttpClient {
           subscriber <- StreamSubscriber[F, HttpResponseBodyPart]
 
           subscribeF = F.delay(publisher.subscribe(subscriber))
+
           bodyDisposal <- Ref.of[F, F[Unit]] {
-            subscriber.stream(subscribeF).take(0).compile.drain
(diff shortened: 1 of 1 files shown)


async-http-client/src/main/scala/org/http4s/client/asynchttpclient/AsyncHttpClient.scala
```diff
@@ -122,9 +126,11 @@ object AsyncHttpClient {
       override def onThrowable(throwable: Throwable): Unit =
         invokeCallback(logger)(cb(Left(throwable)))
 
-      // it's okay to invoke this repeatedly since repeated async callbacks are dropped by law
       override def onCompleted(): Unit =
-        invokeCallback(logger)(cb(Right(response -> dispose)))
+        onStreamCalled.get
+          .ifM(F.unit, F.delay(invokeCallback(logger)(cb(Right(response -> dispose)))))
```
rossabaker, line 131: `F.delay`ing something that calls `unsafeRunSync` to be called in an `unsafeRunSync` feels abusive, even for a spot we need `Effect`. I'd like to think harder about this, but more importantly, I'd like to merge the bugfix.

aeons: This is also one of the places where using the `ifTrue` and `ifFalse` labels helps readability a lot.

wemrysi:
Appreciate the feedback, I've pushed another commit addressing the comments.

async-http-client/src/main/scala/org/http4s/client/asynchttpclient/AsyncHttpClient.scala
```diff
@@ -88,8 +88,9 @@ object AsyncHttpClient {
           subscriber <- StreamSubscriber[F, HttpResponseBodyPart]
 
           subscribeF = F.delay(publisher.subscribe(subscriber))
+
           bodyDisposal <- Ref.of[F, F[Unit]] {
-            subscriber.stream(subscribeF).take(0).compile.drain
+            subscriber.stream(subscribeF).pull.uncons.void.stream.compile.drain
```
kevinmeredith, line 93: Can you please explain what "force finalization" means, @wemrysi?

wemrysi: @kevinmeredith in our usage, we noticed that if the body of responses produced by the AHC client weren't consumed, the connection was never closed. A solution to this was to add an effect to the `Resource` finalizer that would subscribe to the body stream and then immediately `cancel` the subscription, which closes the connection.

The original attempt at this used `take(0)` which turns out to be a noop (it results in a pure, empty stream) and thus the subscribe/cancel effects are never run. Switching to `pull.uncons` ensures that the first effect in the stream is sequenced and any finalizers are run when the stream ends.

rossabaker:
Oops, we merged this to master.  Cherry-picking.  I think it may fix #3354.

async-http-client/src/main/scala/org/http4s/client/asynchttpclient/AsyncHttpClient.scala
kevinmeredith: Thanks @wemrysi!

PR #3331, opened 2020-04-16
Fix #3308: serve index.html if request points to directory

danbim:

PR #3339, opened 2020-04-20
Skip call a second host test on CI

rossabaker:
Regretfully. Is causing problems, for example on https://github.com/http4s/http4s/pull/3338/.


hamnis:
this is unfortunate

rossabaker:
I just rambled about these a bit on https://github.com/http4s/http4s/pull/3331. I'm wondering about looking at MUnit and seeing if its superior async support avoids the deadlocks that plague blaze tests, but even if we ported all the blaze tests, the specs2 tests could still be deadlocking things.  Maybe we need an munit executor service as well, and could gradually port tests and see if it makes a difference?

PR #3340, opened 2020-04-20
Extract dependency versions into own object

rossabaker:
Upgrades jetty-server to 9.4.28, which we missed before, because the old way was too complicated.

This is going to be unpleasant to merge forward, but I think will be nicer once we do.


aeons:
This is the convention we use for all our internal stuff.

PR #3341, opened 2020-04-20
Extract dependency versions for 0.21

rossabaker:
Like #3340, but for 0.21.

Includes:
* scalatags-0.9.0 (closes #3337)
* specs2-4.9.3
* scala-xml-1.3.0
* okhttp-4.5.0


rossabaker:
Thanks for your persistence!  I have another PR to disable another flaky test, and there is a PR that actually fixes the AHC defect, so hopefully things will get better soon.

PR #3347, opened 2020-04-22
Use CollectionCompat in WebjarService

fthomas:
This uses `CollectionCompat.CollectionConverters.asScalaIterator` instead of our own `asScalaIterator`. Motivated by https://github.com/http4s/http4s/pull/3326#pullrequestreview-395934628.


fthomas:
The test errors on the adopt@1.14 builds seem to be unrelated.

fthomas:
Btw, feel free to squash on merge because the build errors of the first two commits are legit.

PR #3349, opened 2020-04-22
Drop semicolon to be an allowed query-part separator by default

### LINKED ISSUE #3277: Uri replaces semicolons with ampersands in query parameters
See query param value `add;fgh` is rendered as `asd&fgh`

```
scala> :paste
// Entering paste mode (ctrl-D to finish)
  val ori = "http://domain.com/path?param1=asd;fgh"
  val res = org.http4s.Uri.unsafeFromString(ori).renderString
  assert(ori == res, s"urls don't match:\n$ori\n$res")
// Exiting paste mode, now interpreting.
java.lang.AssertionError: assertion failed: urls don't match:
http://domain.com/path?param1=asd;fgh
http://domain.com/path?param1=asd&fgh
  at scala.Predef$.assert(Predef.scala:219)
  ... 36 elided
```

hamnis:
We should instead parse the ; and emit it as percent-encoded.
HTML 4 added this as a recommentation to support CGI-BIN servers.
Most sane libraries now use & for query separators.

https://www.w3.org/TR/1999/REC-html401-19991224/appendix/notes.html#h-B.2.2

If you still need to parse query-parts with ";", the code is still
there but not enabled by default.

Fixes #3277

### DIFF
--- a/core/src/main/scala/org/http4s/parser/QueryParser.scala
+++ b/core/src/main/scala/org/http4s/parser/QueryParser.scala
@@ -107,7 +107,7 @@ private[http4s] object QueryParser {
 
   def parseQueryString(queryString: String, codec: Codec = Codec.UTF8): ParseResult[Query] =
     if (queryString.isEmpty) Right(Query.empty)
-    else new QueryParser(codec, true).decode(CharBuffer.wrap(queryString), true)
+    else new QueryParser(codec, colonSeparators = false).decode(CharBuffer.wrap(queryString), true)
 
   private sealed trait State
   private case object KEY extends State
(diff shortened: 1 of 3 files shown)


hamnis:
then `Uri` should probably leave the Query params as a string and leave the interpretation up to be handled when they are needed.
That is not possible to do in a binary compatible way i think ?

rossabaker:
I think that's correct.  We don't preserve the delimiter in the current encoding, and it's a final class.  This needs to be a 1.0 fix, I believe.

hamnis:
We will handle this differently in 1.0, closing this for now.

PR #3366, opened 2020-04-28
proposal: improve client errors handling

albertoadami:
Right now every call http call that returns an unexpected result is throwing an `UnexpectedStatus` exception, in this object i get only the information about the status response.

The problem from my point of view is that to debug the exception without the stack trace is not easy ( I cannot understand which call generate the error), right now the only way that I've to handle this is to add a `recoverWith` block to handle the exception.

My proposal is to add the original call information inside the exception, to understand more easy what call generate the exception.


client/src/main/scala/org/http4s/client/Client.scala
```diff
@@ -82,7 +82,7 @@ trait Client[F[_]] {
   @deprecated("Use `Stream.eval(req).flatMap(client.stream).flatMap(f)`", "0.19.0-M4")
   def streaming[A](req: F[Request[F]])(f: Response[F] => Stream[F, A]): Stream[F, A]
 
-  def expectOr[A](req: Request[F])(onError: Response[F] => F[Throwable])(
+  def expectOr[A](req: Request[F])(onError: (Request[F], Response[F]) => F[Throwable])(
```
Daenyth, line 85: This is a pretty aggressive compatibility change.

I believe you can pull the request Uri from the Response object, no?

albertoadami: I don't see the reference to the orignal request inside the `Response`, Am I missing somehing?

rossabaker: I don't think you can, but the caller has the request.  See below for an alternate idea.

client/src/main/scala/org/http4s/client/DefaultClient.scala
```diff
@@ -230,6 +230,7 @@ private[http4s] abstract class DefaultClient[F[_]](implicit F: Bracket[F, Throwa
   def get[A](s: String)(f: Response[F] => F[A]): F[A] =
     Uri.fromString(s).fold(F.raiseError, uri => get(uri)(f))
 
-  private def defaultOnError(resp: Response[F])(implicit F: Applicative[F]): F[Throwable] =
-    F.pure(UnexpectedStatus(resp.status))
+  private def defaultOnError(req: Request[F], resp: Response[F])(
```
rossabaker, line 233: What if you curried this:

```suggestion
  private def defaultOnError(req: Request[F])(resp: Response[F])(
```

Then the `expect`s look like:

```scala
expectOr(req)(defaultOnError(req))
```

client/src/main/scala/org/http4s/client/Client.scala
```diff
@@ -236,6 +236,9 @@ object Client {
     }
 }
 
-final case class UnexpectedStatus(status: Status) extends RuntimeException with NoStackTrace {
-  override def getMessage: String = s"unexpected HTTP status: $status"
+final case class UnexpectedStatus(status: Status, originalRequest: Uri, method: Method)
```
rossabaker, line 239: I think `originalRequest` is too broad a name.  And method usually comes before uri in our API (and on the wire).  How about:

```suggestion
final case class UnexpectedStatus(status: Status, requestMethod: Method, requestUri: Uri)
```

client/src/main/scala/org/http4s/client/Client.scala
albertoadami: I've restored the old signature and used the approach suggested by @rossabaker

hamnis (approved):

rossabaker:
Ugh. We should have gotten this into 1.0.0-M5. I'll deal with the merge conflicts, and we'll get it on the next bus.

PR #3370, opened 2020-04-29
Fixes #3313 Allow passing a ClassLoader to ResourceService and Webjar…

### LINKED ISSUE #3313: Allow passing a ClassLoader to ResourceService and WebjarService
See #2834 for motivation. `StaticFile.fromResource` now has a `ClassLoader` parameter, but `ResourceService` [always uses the default value](https://github.com/http4s/http4s/blob/v0.21.3/server/src/main/scala/org/http4s/server/staticcontent/ResourceService.scala#L58-L63). Same goes for `WebjarService`.

ashwinbhaskar:
…Service

### DIFF
--- a/server/src/main/scala/org/http4s/server/staticcontent/WebjarService.scala
+++ b/server/src/main/scala/org/http4s/server/staticcontent/WebjarService.scala
@@ -13,16 +13,68 @@ import scala.util.control.NoStackTrace
   */
 object WebjarService {
 
-  /** [[org.http4s.server.staticcontent.WebjarService]] configuration
-    *
-    * @param blocker execution context for blocking I/O
-    * @param filter To filter which assets from the webjars should be served
-    * @param cacheStrategy strategy to use for caching purposes. Default to no caching.
-    */
-  final case class Config[F[_]](
+  class WebjarServiceBuilder[F[_]] private (
       blocker: Blocker,
-      filter: WebjarAssetFilter = _ => true,
-      cacheStrategy: CacheStrategy[F] = NoopCacheStrategy[F])
+      webjarAssetFiltr: WebjarAssetFilter,
+      cacheStrategy: CacheStrategy[F],
+      classLoader: Option[ClassLoader]) {
+
+    private def copy(
+        blocker: Blocker = blocker,
+        webjarAssetFiltr: WebjarAssetFilter = webjarAssetFiltr,
+        cacheStrategy: CacheStrategy[F] = cacheStrategy,
+        classLoader: Option[ClassLoader] = classLoader) =
+      new WebjarServiceBuilder[F](blocker, webjarAssetFiltr, cacheStrategy, classLoader)
+
+    def withWebjarAssetFilter(webjarAssetFilter: WebjarAssetFilter): WebjarServiceBuilder[F] =
+      copy(webjarAssetFiltr = webjarAssetFilter)
+
+    def withCacheStrategy(cacheStrategy: CacheStrategy[F]): WebjarServiceBuilder[F] =
+      copy(cacheStrategy = cacheStrategy)
+
+    def withClassLoader(classLoader: Option[ClassLoader]): WebjarServiceBuilder[F] =
+      copy(classLoader = classLoader)
+
+    def withBlocker(blocker: Blocker): WebjarServiceBuilder[F] =
+      copy(blocker = blocker)
+
+    def toRoutes(implicit F: Sync[F], cs: ContextShift[F]): HttpRoutes[F] = {
+      object BadTraversal extends Exception with NoStackTrace
+      val Root = Paths.get("")
+      Kleisli {
+        // Intercepts the routes that match webjar asset names
+        case request if request.method == Method.GET =>
+          request.pathInfo.split("/") match {
+            case Array(head, segments @ _*) if head.isEmpty =>
+              OptionT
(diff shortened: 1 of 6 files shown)


server/src/main/scala/org/http4s/server/staticcontent/WebjarService.scala
```diff
@@ -55,7 +55,9 @@ object WebjarService {
     * @param config The configuration for this service
     * @return The HttpRoutes
     */
-  def apply[F[_]](config: Config[F])(implicit F: Sync[F], cs: ContextShift[F]): HttpRoutes[F] = {
+  def apply[F[_]](config: Config[F], classLoader: Option[ClassLoader] = None)(
```
rossabaker, line 58: This is not binary compatible, so we'll have to relegate this to master. (And we need to figure out why our MiMa is not running.  This burned us on 0.20.22.)

rossabaker, line 58: Is there any reason to introduce another parameter rather than put this in `Config`?

I have a more radical idea: move this to the same kind of builder pattern we use for server builders, and unlike the case class config or default optional parameters, we'll be able to introduce more parameters without breaking compatibility in the future.

rossabaker: If this doesn't make any sense, I'd be happy to sketch one, and could follow the pattern on the other.

ashwinbhaskar: oh okay, do you want me to raise this PR against master?

ashwinbhaskar: @rossabaker A builder like this??

```
  class WebjarServiceBuilder[F[_]] {
    private var webjarAssetFilter: WebjarAssetFilter = _ => true
    private var cacheStrategy: CacheStrategy[F] = NoopCacheStrategy[F]
    private var classLoader: Option[ClassLoader] = None

    def wid(webjarAssetFilter: WebjarAssetFilter): WebjarServiceBuilder[F] = {
      this.webjarAssetFilter = webjarAssetFilter
      this
    }

    def wid(cacheStrategy: CacheStrategy[F]): WebjarServiceBuilder[F] = {
      this.cacheStrategy = cacheStrategy
      this
    }

    def wid(classLoader: ClassLoader): WebjarServiceBuilder[F] = {
      this.classLoader = Some(classLoader)
      this
    }

    def build(implicit F: Sync[F], cs: ContextShift[F]): HttpRoutes[F] = ???
  }
```

rossabaker: Yes, unless we can figure out a way to make it binary compatible.  We could if we left the original signature, and called the new one something other than `apply`.  Like `withClassLoader`.  But we'd probably want to deprecate that in 1.0 in favor of the builder.

rossabaker: That's pretty close.  This is pretty similar to a builder in an object-oriented language, but our backend "builders" are immutable.  Something like:

```scala
// private constructor: we want people to call it like `WebjarServiceBuilder[F].blah.build`
class WebjarServiceBuilder[F[_]] private (
  webjarAssetFilter: WebjarAssetFilter,
  cacheStrategy: CacheStrategy[F],
  classLoader: Option[ClassLoader]) {

  // You don't strictly need this, but it means you don't need to repeat arguments in all your `with*` methods below. 
  private def copy(
    webjarAssetFilter: WebjarAssetFilter = webjarAssetFilter,
    cacheStrategy: CacheStrategy[F] = cacheStrategy,
    classLoader: Option[ClassLoader] = classLoader) =
    new WebjarServiceBuilder(webjarAssetFilter, cacheStrategy, classLoader)

  // Each field gets its own "with" setter.
  def withWebjarAssetFilter(webjarAssetFilter: WebjarAssetFilter) = 
    copy(webjarAssetFilter = webjarAssetFilter)

  // and so on for cacheStrategy and classLoader

  // Or maybe call this `.toRoutes`?
  def build = ???
}

object WebjarServiceBuilder {
  // If you have arguments with no sensible default, pass them to apply.
  // This is also a good spot for constraints, if you have any.
  def apply[F[_]] = new WebjarServiceBuilder(
     // Supply your default arguments here
     webjarAssetFilter = _ => true,
     cacheStrategy = NoopCacheStrategy[F],
     classLoader = None)
}
```

I haven't tried to compile any of that, but look at the other files named *Builder for working examples.

ashwinbhaskar: @rossabaker cool, got it

ashwinbhaskar: @rossabaker do you want to get rid of the class `WebjarService.Config[F[_}]` all together? That will impact the way it is interfaced with the outside world. [Here](https://github.com/http4s/http4s/blob/0ec8e6a69da2bd53774bdc633eb768b336ec7b8a/server/src/main/scala/org/http4s/server/staticcontent/package.scala#L23) is the line. The users of the library will now have to change their way of calling `wejarService[F](..)` method.

ashwinbhaskar:
closing this pull request in favor of creating this pull request against master - https://github.com/http4s/http4s/pull/3373

PR #3371, opened 2020-04-29
Add SBT Missing Link

diesalbla:
SBT Missing Link is an sbt plugin, developed by the Scala Center, 
that statically analyses all library dependencies (and transitive)
looking for Binary compatibility problems. 
https://github.com/scalacenter/sbt-missinglink

Given that http4s core sits upon many Scala libraries as well as 
many Java libraries, getting a static analysis for this is needed

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -4,6 +4,7 @@ libraryDependencies += "ch.qos.logback" % "logback-classic" % "1.2.3"
 // https://github.com/coursier/coursier/issues/450
 classpathTypes += "maven-plugin"
 
+addSbtPlugin("ch.epfl.scala"              %  "sbt-missinglink"           % "0.3.1")
 addSbtPlugin("com.earldouglas"            %  "xsbt-web-plugin"           % "4.2.0")
 addSbtPlugin("com.eed3si9n"               %  "sbt-buildinfo"             % "0.9.0")
 addSbtPlugin("com.eed3si9n"               %  "sbt-unidoc"                % "0.4.3")


rossabaker:
This seems like a good idea.  I suppose we'd want to add missinglinkCheck to CI.  I tried calling it, and it's very unhappy about `StaticLoggerBinder`, which _shouldn't_ exist, so it will also require some config.

Any reason not to start running this on series/0.21 instead?

LeifW:
Is this something to make up for the fact that test coverage may not exercise all code paths, so this will proactively discover all binary incompatibilities?

rossabaker:
My interpretation is a combination of that, and libraries not strictly adhering to semantic versioning, so eviction warnings aren't failsafe.  It's an exceedingly paranoid plugin, but linker errors evoke that emotion in all of us, no?

diesalbla:
> Is this something to make up for the fact that test coverage may not exercise all code paths, so this will proactively discover all binary incompatibilities?

As far as _binary incompatibilities_, adding this plugin is better than testing.  Any static analysis is more likely to be exhaustive and complete (not miss any error) than a test suite, however exhaustive. Writing and maintaining a test suite takes more effort than just adding and integrating this plugin, and running it on every build. Furthermore, the analysis is likely to keep up with any upgrade to the libraries that happens. Thus, this plugin could be an effective way to prevent _linking errors_. 

However, it cannot replace unit tests, because we may also care what those dependency methods _do_ in its implementation. Much like the way in which a compiler can check if the types of a method call parameters match the method signature, but not what the method does.

PR #3372, opened 2020-04-30
Deprecates Client#fetch

### LINKED ISSUE #3345: Deprecate Client#fetch
@ChristopherDavenport has been [arguing for this](https://twitter.com/davenpcm/status/1244720865410822150), and I like @SystemFw's [argument about not burying `use`](https://gitter.im/http4s/http4s?at=5e9db69ca9ca186206510cfc).

Steps:
1. Add `@deprecated` annotation to both overloads of `Client.fetch`.
2. Replace `client.fetch(req)(f)` with `client.run(req).use(f)` in tests and docs.

Bonus step 1.5, but more difficult: Implement a scalafix rule. Would make step 2 easier, and will help a lot of people upgrading.

satorg:
Fixes #3345

Tasks to complete:

- [x] Add `@deprecate` annotation to these `Client` methods:
    ```scala
    def fetch[A](req: Request[F])(f: Response[F] => F[A]): F[A]
    def fetch[A](req: F[Request[F]])(f: Response[F] => F[A]): F[A]
    ```
- [x] Fix all entrances of `fetch` in the project code (including tests) by replacing it with `run` followed by `use`<sup>[1](#footnote1)</sup>.
- [x] Add **scalafix** migration rules<sup>[2](#footnote2),[3](#footnote3)</sup>.
---

<a name="footnote1">**1:**</a> There were some test cases which lost their sense after all the replacements because they turned into identical cases for the `run` method. Such cases were deleted.

<a name="footnote2">**2:**</a> (UPD: RESOLVED) Initially the **scalafix** subproject had the following dependency versions on **http4s**:
  * **input**: "0.18.21"
  * **output**: "0.20.0-M5"

To elaborate **scalafix** rules for this PR, the versions above were updated to "0.21.4" for both **input** and **output**. The old rules were removed.

If we wish to preserve the old rules, I see two ways to achieve it:
1. Create multiple pairs of **input** and **output** modules for every **http4s** version we want to handle.
2. Keep a single pair of **input** and **output**, but add versioning to **rules** which would allow to publish and then reference it by its desired version.

I'd assume that p.2 is more preferable and easy to manage, but I'm open to discussion.

<a name="footnote3">**3:**</a> Not all possible cases are handled in the rules mostly due to **scalafix** limitations. It includes cases where typeclasses like `Bracket` or `FlatMap` are not available for `F[_]` in a call site of `fetch`, but they're necessary for a new statement which replaces the old one. Unfortunately, **scalafix** doesn't allow to check whether some typeclass available for a particular type in a document tree or not. See [this](#issuecomment-630332148) comment for an example.


satorg:
I'd like to get confirmations on [1](#footnote1) and [2](#footnote2) if I can do it:
1. Remove the **TODO**-ed test cases.
2. Remove obsolete migration rules and update `http4s` versions in **scalafix** to more relevant values.

@rossabaker, WDYT?

client/src/main/scala/org/http4s/client/Client.scala
```diff
@@ -24,6 +24,7 @@ trait Client[F[_]] {
     *            response body afterward will result in an error.
     * @return The result of applying f to the response to req
     */
+  @deprecated("Use run(req).use(f)", "0.21")
```
rossabaker, line 27: We usually put patch versions  in these, so 0.21.5.

scalafix/build.sbt
```diff
@@ -27,29 +27,32 @@ lazy val rules = project.settings(
   libraryDependencies += "ch.epfl.scala" %% "scalafix-core" % V.scalafixVersion
 )
 
+val http4sInputVersion = "0.21.4"
+val http4sOutputVersion = "0.21.4"
```
rossabaker, line 31: 0.18 is ancient, but it's also a shame to lose the old rules.  Is there a way to version these, so people might run multiple iterations if catching up across multiple versions?  Probably related to these "semantic rules," but I would need to refresh my memory how these work.

satorg: The problem is that the old rules don't even compile with the newer http4s version. I'd also prefer to keep the old rules but have no idea how to achieve it easily. But I'm open to any suggestions:)

rossabaker: Perhaps @amarrella remembers how this works, as the original contributor.  If not, we can learn together. :smile:

amarrella: It's a long time since I've looked at these but can we perhaps set the old version as a dependency instead of depending on the project?

satorg: @amarrella sorry, not sure if I caught it. Could you explain your idea a bit more please?

satorg:
I think I almost done with **scalafix** rules. But there's still a question about what should we do with the old rules. I'm not sure if I understand what exactly @amarrella suggests.

BTW, once **scalafix** rules are implemented how can I apply them to the base **http4s** project's sources?

scalafix/build.sbt
amarrella: @satorg apologies, I misunderstood the problem. I see the issue now with keeping `http4sInputVersion = 0.18.x` :( 
Perhaps the cleanest version is getting rid of the rules like you did, and point users to use a older version of the jar if they need the older rule and then updating to the new one. I know it's way less ergonomic though :/

satorg: Well, I think we also could convert the **scalafix** directory into a multi-module project where each module has its own set of dependencies. But I'm not sure if it's a really good way to go.

rossabaker:
That's a great question on how to run them on our own sources.  Can you use `--tool-classpath` to point to your local build?  I confess it's been a while since I've run them, and I'm behind on the state of the art.

satorg:
@rossabaker I think I finished the scalafix rules for this task.
I wrote a bunch of rules for different cases – could you check it out please?
Let me know if you come up with some uncovered cases please.

satorg:
BTW, I realized that there are some cases which cannot be fixes properly. For example, consider the following:
```scala
trait MyIntermediateService[F[_]] {
  def client: Client[F[_]]

  final def someFuncThatCallsClientFetch[A](req: Request[F]): F[A] = {
    client.fetch(req) { res => /* do something */ }
  }
}
```
The problem with this sample is that `Client.fetch` itself doesn't require any typeclasses, but its replacement (`Client.run` followed by `Resource.use`) requires a `Bracket`. But since the enclosing trait doesn't have a `Bracket` for `F[_]`, the resulting code just won't compile. And I doubt if it is really possible to automate it somehow...

SystemFw:
ah, that's a great point. It would be nice if you could inject  a custom error for that somewhere

satorg:
> ah, that's a great point. It would be nice if we could somehow inject a custom error for that somewhere

Hmm... any ideas on that? As an alternative we could try to add a **TODO** comment to point out at the problematic line.

satorg:
@rossabaker I managed to run the new rule on the project sources in this way:
1. From the scalafix project publish the rules module locally (it gets published as `org.http4s:scalafix:0.1.0-SNAPSHOT`).
2. Temporarily add the sbt-scalafix plugin to http4s project.
3. Enable it with scalafixEnable
4. Apply the rule on main and test sources (on `master` branch) by executing compile:scalafix and test:scalafix with `dependency:v0_21@org.http4s:scalafix:0.1.0-SNAPSHOT` arg.

client/src/main/scala/org/http4s/client/Client.scala
```diff
@@ -30,6 +30,7 @@ trait Client[F[_]] {
     *            response body afterward will result in an error.
     * @return The result of applying f to the response to req
     */
+  @deprecated("Use run(req).use(f)", "0.21.5")
```
kevinmeredith, line 33: I question whether this message should indicate, "Don't use `fetch`, but, if you must, do `run(req).use(f)` per the helpful discussion in https://github.com/http4s/http4s/issues/3345#issuecomment-619598068 regarding what's wrong with `fetch`. 

Thoughts, @rossabaker and @satorg?

I admit that I don't know if "why is it deprecated" should be placed inside of the `@deprecated`'s message.

satorg: Sorry, didn't get, what is your concern with the message? Do you think it should be more descriptive?

satorg:
I think I did my best with all that jazz around Scalafix rules. Not all possible cases are covered, though, due to limitations in Scalafix itself.

client/src/main/scala/org/http4s/client/Client.scala
rossabaker: I think the message is in line with what we usually do.

client/src/main/scala/org/http4s/client/DefaultClient.scala
```diff
@@ -227,7 +230,7 @@ private[http4s] abstract class DefaultClient[F[_]](implicit F: Bracket[F, Throwa
     * @return The result of applying f to the response to req
     */
   def get[A](uri: Uri)(f: Response[F] => F[A]): F[A] =
-    fetch(Request[F](Method.GET, uri))(f)
+    run(Request[F](Method.GET, uri)).use(f)
```
rossabaker, line 233: This one is also problematic for the same reasons as `fetch`.

client/src/main/scala/org/http4s/client/DefaultClient.scala
```diff
@@ -39,18 +39,18 @@ private[http4s] abstract class DefaultClient[F[_]](implicit F: Bracket[F, Throwa
     * @return The result of applying f to the response to req
     */
   def fetch[A](req: F[Request[F]])(f: Response[F] => F[A]): F[A] =
-    req.flatMap(fetch(_)(f))
+    req.flatMap(run(_).use(f))
```
rossabaker, line 42: In a separate PR, it makes sense to deprecate this one.  It's the same hidden use.

client/src/main/scala/org/http4s/client/DefaultClient.scala
```diff
@@ -39,18 +39,18 @@ private[http4s] abstract class DefaultClient[F[_]](implicit F: Bracket[F, Throwa
...
     * service are disposed on completion of callback task f.
     *
-    * This method effectively reverses the arguments to `fetch`, and is
+    * This method effectively reverses the arguments to `run` followed by `use`, and is
     * preferred when an HTTP client is composed into a larger Kleisli function,
     * or when a common response callback is used by many call sites.
     */
   def toKleisli[A](f: Response[F] => F[A]): Kleisli[F, Request[F], A] =
-    Kleisli(fetch(_)(f))
+    Kleisli(run(_).use(f))
```
rossabaker, line 53: And maybe deprecate this one?  It could go to `Kleisli[Resource[F, ?], Request[F], Response[F]]`, but maybe it's not useful.

client/src/main/scala/org/http4s/client/Client.scala
kevinmeredith: Thanks for addressing my question, satorg and Ross.

client/src/main/scala/org/http4s/client/DefaultClient.scala
satorg: I see there're many legacy stuff to deprecate in `Client`. I think I can take on it too. But I'd prefer to work on it in a separate PR, just to keep this one maintainable.

satorg:
@rossabaker , I'm still a bit concerned about [2](#footnote2). What do you think would be the best way to overcome this? Because if we want to handle it somehow, it'd be better to do it before this PR get merged. From my prospective, the simplest approach would be to introduce versioning to the **scalafix** subproject. I can take on it, but we should agree on the versioning scheme.

*cc* @amarrella

rossabaker:
That's an excellent question.  I haven't written these rules before, and am trying to get a deeper understanding of how they're typically deployed.

I think you're proposing that the version of the rules would uniquely identify an input to output pair?  This might workout nicely if Rules Version N's input was equal to Rules Version (N-1)'s output.  The problem with that is that our releases are non-linear: typically we're deploying backports of the current production version along with milestones of the latest version.  Maybe if the rules version was equal to the http4s ouptut version, and we just focus on the production version and people on the prereleases are on their own?  Even that modest goal would be better than we've historically done.

satorg:
I think it should we okay if the scalafix module version just matches the http4s output version to which the scalafix rules should be applied to.

I can create a small PR for the previous version of scalafix module with version number "0.20.0-M5", then we can get it merged it and then I can change the version in this PR to "0.21.5". Does it work for you guys?

satorg:
@rossabaker, sorry for the late response.
So yes, pretty much. I created a PR (#3502) that should make it.
I think it should work.

_cc_ @amarrella

satorg:
I updated this PR to make it in-sync with all the most recent changes in the repository (at least I think so). Do we need to add something else to this PR or is it in a good shape already to be published with the next release?

I'm going to continue my work on deprecation of aforementioned unsafe methods in `Client` but not sure if it makes sense to stuff everything into this PR.

rossabaker:
I think we're ready to merge, but if we're saying the current http4s version is the output version, and we're going to publish to Sonatype instead of referencing straight from GitHub, then the scalafix build should just be another module in the main build.  I had to do horrible things to get 0.20 published.

I think that too can go in a separate PR.

And thanks, @satorg, for the effort on this.

rossabaker:
Oh, are we targeting 0.21.5 with this?  I thought we were, and the deprecation warnings say so, but I just noticed that it's on master.

satorg:
Yes, it is.. It was on 'master' from the beginning.. As I've realized now, I should rebase it onto '0.21.5'... I'll do it later today, sorry for misunderstanding.

rossabaker:
Okay, I've rebased this one onto 0.21 so it's part of that release.

Thanks, @satorg, for the excellent work on both the deprecation and improving our scalafix situation.  This was a long ticket, but well worth it.

fthomas:
I just learned about the new Scalafix migrations via the release notes of 0.21.5. I think it would be nice to add the new migration to [Scala Steward](https://github.com/fthomas/scala-steward/blob/master/docs/scalafix-migrations.md) although it already created a lot of PRs for http4s 0.21.5 without it. For future releases it would be nice to add new migrations to Scala Steward before a new version is released to ensure that Scala Steward can run them on all open source projects it is working on.

rossabaker:
Oh, I knew that scalafix integration had been discussed, but I didn't realize it was a reality. Very cool!

fthomas:
The rule for [http4s 0.20.0](https://github.com/fthomas/scala-steward/blob/abcf9cb6db0c0a2c791fb62add09b5ff81110cab/modules/core/src/main/resources/scalafix-migrations.conf#L45-L51) was even one of the first migrations added to Scala Steward and [was used in 30+ PRs](https://github.com/pulls?q=author%3Ascala-steward+is%3Apr+scalafix-migrations+github%3Ahttp4s%2Fhttp4s%2Fv0_20) although it was also added after the 0.20.0 release.

rossabaker:
Would the right thing to do be to add the new rule now, then?

rossabaker:
Oh, I should have looked at your linked PR.

fthomas:
I've added the new migration in https://github.com/fthomas/scala-steward/pull/1510. It will be applied to updates going from x < 0.21.5 to y >= 0.21.5. For future migrations, you can ping me when they are discussed/merged and I can help with adding them to Scala Steward.

PR #3373, opened 2020-05-01
Fixes #3313 Allow passing a ClassLoader to ResourceService and Webjar…

### LINKED ISSUE #3313: Allow passing a ClassLoader to ResourceService and WebjarService

ashwinbhaskar:
…Service


server/src/main/scala/org/http4s/server/staticcontent/ResourceService.scala
```diff
@@ -14,67 +14,112 @@ import scala.util.{Failure, Success, Try}
...
-    *
-    * @param basePath prefix of the path files will be served from
-    * @param blocker execution context to use when collecting content
-    * @param pathPrefix prefix of the Uri that content will be served from
-    * @param bufferSize size hint of internal buffers to use when serving resources
-    * @param cacheStrategy strategy to use for caching purposes. Default to no caching.
-    * @param preferGzipped whether to serve pre-gzipped files (with extension ".gz") if they exist
-    */
-  final case class Config[F[_]](
+  class ResourceServiceBuilder[F[_]] private (
```
rossabaker, line 17: I think we could move these to the top level.  Then we could leave the old interface as deprecated to assist people in migration.  What do you think?

server/src/main/scala/org/http4s/server/staticcontent/ResourceService.scala
```diff
@@ -14,67 +14,112 @@ import scala.util.{Failure, Success, Try}
...
 
-  /** [[org.http4s.server.staticcontent.ResourceService]] configuration
-    *
-    * @param basePath prefix of the path files will be served from
-    * @param blocker execution context to use when collecting content
-    * @param pathPrefix prefix of the Uri that content will be served from
-    * @param bufferSize size hint of internal buffers to use when serving resources
-    * @param cacheStrategy strategy to use for caching purposes. Default to no caching.
-    * @param preferGzipped whether to serve pre-gzipped files (with extension ".gz") if they exist
-    */
```
rossabaker, line 25: I wish we didn't lose the scaladoc.  If we make the constructor parameters vals, can we still attach it?

server/src/main/scala/org/http4s/server/staticcontent/ResourceService.scala
ashwinbhaskar: @rossabaker If this is going to go in 1.0, should we not just move away from the older interface right away? It's a breaking change anyway. Why keep the older interface?

rossabaker: I like to keep them with `@deprecated` annotations, so it tells people how to migrate to the new interface.  People are a lot more likely to read their compiler warnings than they are the release notes.

ashwinbhaskar: @rossabaker got it

ashwinbhaskar: @rossabaker I have made the changes and pushed bulder to top level.

ashwinbhaskar:
@rossabaker did you get a chance to review the PR?

server/src/main/scala/org/http4s/server/staticcontent/ResourceService.scala
```diff
@@ -32,6 +146,7 @@ object ResourceService {
       preferGzipped: Boolean = false)
 
   /** Make a new [[org.http4s.HttpRoutes]] that serves static files. */
+  @deprecated("use ResourceServiceBuilder", "1.0.0")
```
rossabaker, line 149: This will be released in a milestone:

```suggestion
  @deprecated("use ResourceServiceBuilder", "1.0.0-M1")
```

ashwinbhaskar: @rossabaker made the changes

rossabaker:
Sorry, one more thing.  Code looks ready, but I hadn't noticed the docs don't compile.  `++2.12.11 docs/tutOnly static.md` should reveal what needs to be done.

ashwinbhaskar:
@rossabaker I made the changes to `static.md ` and pushed it. But CI/Scala 2.13.1 is failing because of `SSL Handshake exception`

rossabaker:
This had some fallout from the `Uri.Path` while we were awaiting the second review. I tried to take care of the conflicts, which were non-trivial, so please check my work.

ashwinbhaskar:
@rossabaker I took a look at changes pertaining to Uri.Path. LGTM.

PR #3385, opened 2020-05-07
Finagle client and server support

jcouyang:
love Http4s DSL so much.
but also I like to have some consistency of ops and monitoring setups on http4s service and our other finagle PRC service, so it is easier to maintain if everything running on the same kind of server.
this is a very first attempt, please any feedback is welcome


jcouyang:
That sounds great!

I stared with a seperate repo already, since it is easier for me to write unit test and publish jar for testing

https://github.com/jcouyang/http4s-finagle

Would you like me to just transfer it to http4s org? @rossabaker

rossabaker:
Excellent!  I just sent a couple PRs for the license and the code of conduct so it matches the rest of the org, and then we can get it transferred over.

rossabaker:
Okay, I think the permissions are set up so you can transfer it, and remain an admin after the transfer.  If either of those are false, let us know, and we'll take another look.

jcouyang:
got this error when transferring
image
still missing permission I guess @rossabaker

rossabaker:
Sorry about that. I just tweaked it. Please try again.

jcouyang:
thanks
transferred to https://github.com/http4s/http4s-finagle
closing this one

PR #3386, opened 2020-05-08
Drop our internal jawn parser for argonaut

rossabaker:
Theirs supports 1.0 now, so we can at last dump ours.


rossabaker:
After this, we can also drop jawn from the NOTICES file in #3389.

PR #3388, opened 2020-05-08
scalafmt-2.5.1

rossabaker:
Not in love with how it wraps the implicit keyword in a few places, but the point of this is to not fight about this anymore.


rossabaker:
Will be easier just to start fresh on that.

PR #3389, opened 2020-05-08
Copyright headers

rossabaker:
A former employer's legal department recommended we do this years ago, and I said, "But I'm maintaining three branches."  Now there are only two, and they're close, so let's do it.

* This should be fully automated, except for bumping the copyright year.
* I have no love for the standard verbosity, and this SPDX looks neat.
* I standardized the way we attribute our derived sources
* In so doing, I found a few projects we hadn't acknowledged in NOTICE.


rossabaker:
Hmm. To be perfectly pedantic about this, 889cb6e should be on master. We acknowledged it just in time to remove it! :laughing: I'll revert it here, and reapply it when I merge.

rossabaker:
Merged to master, and jawn removal reapplied.

PR #3390, opened 2020-05-09
Replace with CIString from case-insensitive library

rossabaker:
Part of the effort to move non-HTTP things out of http4s.

I'd like to replace com.rossabaker.ci with org.typelevel (see https://github.com/typelevel/general/issues/102), but it will look something like this.  I don't think the `.ci` syntax carries its weight, and would like to phase it out.


website/src/hugo/content/changelog.md
```diff
@@ -1837,7 +1837,7 @@ This release is identical to v0.19.0-M4.  We mistagged it.  Please proceed to th
   implementors, with more to follow.  Added an `HttpCodecLaws` to http4s-testing.
 * `Uri.Scheme` is now its own type instead of a type alias.
 * `TransferCoding` is no longer a case class. Its `coding` member is now a
-  `String`, not a `CaseInsensitiveString`. Its companion is no longer a
+  `String`, not a `CIString`. Its companion is no longer a
```
aeons, line 1840: I guess these historical changes should not refer to the new name,

rossabaker: Oh, good catch.  You caught my overzealous search and replace.

rossabaker:
It's probably better to make only one breaking change.  There hasn't been much response on the Typelevel ticket.  If they don't want it, we could also transfer it to http4s.  I just thought I'd offer it up higher since it's not HTTP.

rossabaker:
Force-push was just a scalafmt I cleaned up after the fact.

PR #3398, opened 2020-05-09
Remove scalastyle

rossabaker:
It's a nice tool, but we don't run it or take action on it.

Alternatively: let's start running it and taking action on it.


rossabaker:
I think scalastyle still has some things to offer: it can check naming conventions, magic numbers, etc.  I'm just trying to turn off the things we don't use.

PR #3358, opened 2020-04-27
Replace tut with mdoc

iRevive:
This PR closes https://github.com/http4s/http4s/issues/3259.

There was no activity related to the ticket, thus I decided to pick it up.

PR #3399, opened 2020-05-09
Backport #3358: Replace tut with mdoc

rossabaker:
Trying to backport @iRevive's work on #3358.  I am surprised by some of these errors.  Particularly like:

```
warning: /home/ross/src/http4s/docs/src/main/mdoc/entity.md:39 (mdoc generated code) method any2stringadd in object Predef is deprecated (since 2.13.0): Implicit injection of + is deprecated. Convert to String to call +
}; $doc.binder(audioDec, 1, 4, 1, 12)
```

Have I done something silly?


rossabaker:
The silly thing I did was to try to run it on 2.13.  We can sort that out in a separate issue.

iRevive:
@rossabaker perhaps this issue is related https://github.com/scalameta/mdoc/issues/210.  
We can use `-deprecation:false` flag to hide these errors, but it's not the best way to go IMO. 

I can investigate in which way the problem can be solved.

rossabaker:
@iRevive Great! I'm going to go ahead and move forward with this, so we can get your work in before it bitrots.  tut didn't support Scala 2.13 either, so this is a strict improvement, and we can make it even better in a subsequent PR.

PR #3400, opened 2020-05-09
http4s.org SBT plugin

rossabaker:
The idea behind `Http4sPlugin` was to extract an sbt plugin that would be useful to set a baseline across our modules.  As we think about splitting up the repo, this becomes a more attractive goal.  The end state is to take a handful of these plugins, and put then in an sbt-http4s-org repo to be used by all our projects.

I am intentionally not versioning dependencies.  I did this in an org plugin at a previous job, and it was a nightmare to update the version in the plugin, release the plugin, update the version of the plugin downstream, and release.  (Made more fun when the org plugin referenced the version in one of the downstream libraries.)


rossabaker:
I'd like to set up kind-projector and better-monadic-for out of the box, but that somewhat contradicts the no-dependency-management principle.

aeons:
:+1:

I could live with bm4 and kind projector defying the rules.

rossabaker:
There could be a default plugins setting so projects can upgrade or opt out, but it mostly works.

I can imagine us having 10-12 active repositories when we start splitting this one, so my vision here is to have a "living project template" instead of something that gets cloned and then slowly falls out of date.  It worked great at one employer and horribly at the next, so this will be the tiebreaker. :laughing: If it hurts, we can always inline it, and the build is already a lot cleaner.

rossabaker:
Oops, I thought we were on the released version, and there are conflicts.

PR #3416, opened 2020-05-11
Docs non static

drocsid:


drocsid:
From https://github.com/http4s/blaze/issues/405 we discovered that macOS and windows don't support the static flag for native image

rossabaker:
Docs only, and that job already passed, so merging while I have it open.

PR #3417, opened 2020-05-11
backport deployment.md to 0.21

drocsid:


rossabaker:
The doc wasn't appearing because we moved from tut to mdoc while this was under development.  I moved it and pushed as 5eca077.

PR #3422, opened 2020-05-13
Use Sum Type instead of Option[SSLContext] for BlazeClientBuilder#apply

kevinmeredith:
Follows up on https://github.com/http4s/http4s/issues/2431#issuecomment-627474729.

Ross suggested using a sum type in https://github.com/http4s/http4s/issues/2431#issuecomment-496961469. Today, my teammates got confused at the `Option[SSLContext]`'s behavior when upgrading from 0.20.x -> 0.21.4 in `http4s`. 

Those two served as motivations to open this PR.

Hi @rossabaker - please take a look. Thanks!

### DIFF
--- a/blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala
+++ b/blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala
@@ -138,7 +138,7 @@ sealed abstract class BlazeClientBuilder[F[_]] private (
 
   /** Use the provided `SSLContext` when making secure calls */
   def withSslContext(sslContext: SSLContext): BlazeClientBuilder[F] =
-    withSslContextOption(Some(sslContext))
+    copy(sslContext = SSLContextOption.Provided(sslContext))
 
   /** Use an `SSLContext` obtained by `SSLContext.getDefault()` when making secure calls.
     *
@@ -153,7 +153,7 @@ sealed abstract class BlazeClientBuilder[F[_]] private (
 
   /** Disable secure calls */
   def withoutSslContext: BlazeClientBuilder[F] =
-    copy(sslContext = None)
+    copy(sslContext = SSLContextOption.NoSSL)
 
   def withCheckEndpointAuthentication(checkEndpointIdentification: Boolean): BlazeClientBuilder[F] =
     copy(checkEndpointIdentification = checkEndpointIdentification)
--- a/blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala
+++ b/blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala
@@ -278,15 +278,16 @@ object BlazeClientBuilder {
 
   sealed trait SSLContextOption
   object SSLContextOption {
-    case object NoSSL                                 extends SSLContextOption
-    case object TryDefaultSSLContextOrNone            extends SSLContextOption
+    case object NoSSL extends SSLContextOption
+    case object TryDefaultSSLContextOrNone extends SSLContextOption
     final case class Provided(SSLContext: SSLContext) extends SSLContextOption
 
-    def toMaybeSSLContext(sco: SSLContextOption): Option[SSLContext] = sco match {
-      case SSLContextOption.NoSSL => None
-      case SSLContextOption.TryDefaultSSLContextOrNone =>tryDefaultSslContext
-      case SSLContextOption.Provided(context) => Some(context)
-    }
+    def toMaybeSSLContext(sco: SSLContextOption): Option[SSLContext] =
+      sco match {
+        case SSLContextOption.NoSSL => None
+        case SSLContextOption.TryDefaultSSLContextOrNone => tryDefaultSslContext
+        case SSLContextOption.Provided(context) => Some(context)
+      }
   }
 
   import SSLContextOption.{TryDefaultSSLContextOrNone, toMaybeSSLContext}
--- a/blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala
+++ b/blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala
@@ -295,14 +295,9 @@ object BlazeClientBuilder {
   /** Creates a BlazeClientBuilder
     *
     * @param executionContext the ExecutionContext for blaze's internal Futures
-    * @param sslContextOption indicates how to resolve SSLContext. The sum type offers three options:
-    *                         NoSSL                = do not use SSL/HTTPS
-    *                         TryDefaultSSLContext = `SSLContext.getDefault()`, or `None` on systems where the default is unavailable
-    *                         Provided             = use the explicitly passed SSLContext
     */
   def apply[F[_]: ConcurrentEffect](
-      executionContext: ExecutionContext,
-      sslContextOption: SSLContextOption = TryDefaultSSLContext): BlazeClientBuilder[F] =
+      executionContext: ExecutionContext): BlazeClientBuilder[F] =
     new BlazeClientBuilder[F](
       responseHeaderTimeout = Duration.Inf,
       idleTimeout = 1.minute,
@@ -312,7 +307,7 @@ object BlazeClientBuilder {
       maxTotalConnections = 10,
       maxWaitQueueLimit = 256,
       maxConnectionsPerRequestKey = Function.const(256),
-      sslContext = toMaybeSSLContext(sslContextOption),
+      sslContext = toMaybeSSLContext(TryDefaultSSLContext),
       checkEndpointIdentification = true,
       maxResponseLineSize = 4096,
       maxHeaderLength = 40960,
--- a/blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala
+++ b/blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala
@@ -279,30 +279,30 @@ object BlazeClientBuilder {
   sealed trait SSLContextOption
   object SSLContextOption {
     case object NoSSL extends SSLContextOption
-    case object TryDefaultSSLContextOrNone extends SSLContextOption
+    case object TryDefaultSSLContext extends SSLContextOption
     final case class Provided(sslContext: SSLContext) extends SSLContextOption
 
     def toMaybeSSLContext(sco: SSLContextOption): Option[SSLContext] =
       sco match {
         case SSLContextOption.NoSSL => None
-        case SSLContextOption.TryDefaultSSLContextOrNone => tryDefaultSslContext
+        case SSLContextOption.TryDefaultSSLContext => tryDefaultSslContext
         case SSLContextOption.Provided(context) => Some(context)
       }
   }
 
-  import SSLContextOption.{TryDefaultSSLContextOrNone, toMaybeSSLContext}
+  import SSLContextOption.{TryDefaultSSLContext, toMaybeSSLContext}
 
   /** Creates a BlazeClientBuilder
     *
     * @param executionContext the ExecutionContext for blaze's internal Futures
     * @param sslContextOption indicates how to resolve SSLContext. The sum type offers three options:
-    *                         NoSSL                      = do not use SSL/HTTPS
-    *                         TryDefaultSSLContextOrNone = `SSLContext.getDefault()`, or `None` on systems where the default is unavailable
-    *                         Provided                   = use the explicitly passed SSLContext
+    *                         NoSSL                = do not use SSL/HTTPS
+    *                         TryDefaultSSLContext = `SSLContext.getDefault()`, or `None` on systems where the default is unavailable
+    *                         Provided             = use the explicitly passed SSLContext
     */
   def apply[F[_]: ConcurrentEffect](
       executionContext: ExecutionContext,
-      sslContextOption: SSLContextOption = TryDefaultSSLContextOrNone): BlazeClientBuilder[F] =
+      sslContextOption: SSLContextOption = TryDefaultSSLContext): BlazeClientBuilder[F] =
     new BlazeClientBuilder[F](
       responseHeaderTimeout = Duration.Inf,
       idleTimeout = 1.minute,
--- a/blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala
+++ b/blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala
@@ -148,9 +148,13 @@ sealed abstract class BlazeClientBuilder[F[_]] private (
     withSslContext(SSLContext.getDefault())
 
   /** Use some provided `SSLContext` when making secure calls, or disable secure calls with `None` */
-  @deprecated(message = "Use withDefaultSslContext, withSslContext or withoutSslContext to set the SSLContext", since = "1.0.0")
+  @deprecated(
+    message =
+      "Use withDefaultSslContext, withSslContext or withoutSslContext to set the SSLContext",
+    since = "1.0.0")
   def withSslContextOption(sslContext: Option[SSLContext]): BlazeClientBuilder[F] =
-    copy(sslContext = sslContext.fold[SSLContextOption](SSLContextOption.NoSSL)(SSLContextOption.Provided))
+    copy(sslContext =
+      sslContext.fold[SSLContextOption](SSLContextOption.NoSSL)(SSLContextOption.Provided))
 
   /** Disable secure calls */
   def withoutSslContext: BlazeClientBuilder[F] =
@@ -285,8 +289,8 @@ object BlazeClientBuilder {
     *  * TryDefaultSSLContext = `SSLContext.getDefault()`, or `None` on systems where the default is unavailable
     *  * Provided             = use the explicitly passed SSLContext
     */
-  private [blaze] sealed trait SSLContextOption extends Product with Serializable
-  private [blaze] object SSLContextOption {
+  private[blaze] sealed trait SSLContextOption extends Product with Serializable
+  private[blaze] object SSLContextOption {
     case object NoSSL extends SSLContextOption
     case object TryDefaultSSLContext extends SSLContextOption
     final case class Provided(sslContext: SSLContext) extends SSLContextOption
--- a/blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala
+++ b/blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala
@@ -276,7 +276,7 @@ sealed abstract class BlazeClientBuilder[F[_]] private (
 
 object BlazeClientBuilder {
 
-  sealed trait SSLContextOption
+  sealed trait SSLContextOption extends Product with Serializable
   object SSLContextOption {
     case object NoSSL extends SSLContextOption
     case object TryDefaultSSLContext extends SSLContextOption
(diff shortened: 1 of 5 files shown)


blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala
```diff
@@ -276,14 +276,33 @@ sealed abstract class BlazeClientBuilder[F[_]] private (
 
 object BlazeClientBuilder {
 
+  sealed trait SSLContextOption
+  object SSLContextOption {
+    case object NoSSL extends SSLContextOption
+    case object TryDefaultSSLContextOrNone extends SSLContextOption
```
Kazark, line 282: @kevinmeredith I believe `Try` already has the sense of `OrNone` (at least that it what it implies in the F# ecosystem)

blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala
```diff
@@ -276,14 +276,33 @@ sealed abstract class BlazeClientBuilder[F[_]] private (
...
-    * @param sslContext Some `SSLContext.getDefault()`, or `None` on systems where the default is unavailable
+    * @param sslContextOption indicates how to resolve SSLContext. The sum type offers three options:
+    *                         NoSSL                      = do not use SSL/HTTPS
+    *                         TryDefaultSSLContextOrNone = `SSLContext.getDefault()`, or `None` on systems where the default is unavailable
+    *                         Provided                   = use the explicitly passed SSLContext
     */
   def apply[F[_]: ConcurrentEffect](
       executionContext: ExecutionContext,
-      sslContext: Option[SSLContext] = tryDefaultSslContext): BlazeClientBuilder[F] =
+      sslContextOption: SSLContextOption = TryDefaultSSLContextOrNone): BlazeClientBuilder[F] =
```
Kazark, line 305: @kevinmeredith I think this is a good change, but it is a breaking change. Is this just getting merged to v1.0, or how does that work?

kevinmeredith: > @kevinmeredith I think this is a good change, but it is a breaking change.

Good point, Keith. Per https://github.com/http4s/http4s/issues/2431#issuecomment-627715221, my expectation is that it'll become part of `1.0`, as you noted.

blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala
kevinmeredith: That's a good point. However, based on naming alone, it's not clear to me that `TryDefaultSSLContext` would default to `None` if the `Try` failed.

What are your thoughts, Keith and others?

kevinmeredith: > at least that it what it implies in the F# ecosystem

At least from my experience in Scala, it's not clear to me that `Try` has the same implication in this language. However, I learn new things everyday, so Ross, Chris, et al, please let me know if the `OrNone` is unnecessary based on Keith's comment.

Kazark: The other reason I was thinking this was that you named this after an existing variable, which did not have `OrNone`. 🤷‍♂️

kevinmeredith: > The other reason I was thinking this was that you named this after an existing variable, which did not have OrNone

Yeah, that's a good point, Keith. I'll re-name it to what you proposed. Thanks for pressing me.

kevinmeredith: https://github.com/http4s/http4s/pull/3422/commits/7067e22da74a40fded2bbe54e7bdaae929b6b8fe

kevinmeredith:
I'm not sure why that `2.12.11` test failed:

```
[error] https://repo.scala-sbt.org/scalasbt/sbt-plugin-releases/com.typesafe.sbt/sbt-native-packager/scala_2.12/sbt_1.0/1.7.2/srcs/sbt-native-packager-sources.jar: download error: Caught java.io.IOException: Server returned HTTP response code: 504 for URL: https://dl.bintray.com/sbt/sbt-plugin-releases/com.typesafe.sbt/sbt-native-packager/scala_2.12/sbt_1.0/1.7.2/srcs/sbt-native-packager-sources.jar (Server returned HTTP response code: 504 for URL: https://dl.bintray.com/sbt/sbt-plugin-releases/com.typesafe.sbt/sbt-native-packager/scala_2.12/sbt_1.0/1.7.2/srcs/sbt-native-packager-sources.jar) while downloading https://repo.scala-sbt.org/scalasbt/sbt-plugin-releases/com.typesafe.sbt/sbt-native-packager/scala_2.12/sbt_1.0/1.7.2/srcs/sbt-native-packager-sources.jar
```

https://github.com/http4s/http4s/pull/3422/checks?check_run_id=671336299

I'm guessing it's a flaky failure?

blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala
```diff
@@ -276,14 +276,33 @@ sealed abstract class BlazeClientBuilder[F[_]] private (
...
+    case object TryDefaultSSLContext extends SSLContextOption
+    final case class Provided(sslContext: SSLContext) extends SSLContextOption
+
+    def toMaybeSSLContext(sco: SSLContextOption): Option[SSLContext] =
+      sco match {
+        case SSLContextOption.NoSSL => None
+        case SSLContextOption.TryDefaultSSLContext => tryDefaultSslContext
+        case SSLContextOption.Provided(context) => Some(context)
+      }
+  }
```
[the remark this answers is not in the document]
rossabaker: Actually, we could probably keep this internal. The original problem is referencing `SSLContext.getDefault` as a a default argument that might throw, and `NoSSL` being a lousy default.

We could model it internally as your new type, so `TryDefaultSSLContext` can be the default value of the builder.  Then we can just keep the `withSslContext` and `withoutSsl` as is to set `Provided` and `NoSSL`, respectively.

blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala
```diff
@@ -276,14 +276,33 @@ sealed abstract class BlazeClientBuilder[F[_]] private (
...
-    * @param sslContext Some `SSLContext.getDefault()`, or `None` on systems where the default is unavailable
+    * @param sslContextOption indicates how to resolve SSLContext. The sum type offers three options:
+    *                         NoSSL                = do not use SSL/HTTPS
+    *                         TryDefaultSSLContext = `SSLContext.getDefault()`, or `None` on systems where the default is unavailable
+    *                         Provided             = use the explicitly passed SSLContext
     */
   def apply[F[_]: ConcurrentEffect](
       executionContext: ExecutionContext,
-      sslContext: Option[SSLContext] = tryDefaultSslContext): BlazeClientBuilder[F] =
+      sslContextOption: SSLContextOption = TryDefaultSSLContext): BlazeClientBuilder[F] =
```
rossabaker, line 305: I think this argument could and should go away.  We usually only have the things without sensible defaults in the builder `apply` methods.

blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala
```diff
@@ -293,7 +312,7 @@ object BlazeClientBuilder {
       maxTotalConnections = 10,
       maxWaitQueueLimit = 256,
       maxConnectionsPerRequestKey = Function.const(256),
-      sslContext = sslContext,
+      sslContext = toMaybeSSLContext(sslContextOption),
```
rossabaker, line 315: And then the type of this value turns into the new `SSLContextOption`.

blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala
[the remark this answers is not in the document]
kevinmeredith: > We have this same basic problem on multiple backends, client and server, which sort of implies core.

The "same basic problem" is in terms of `SSLContext`, Ross? 

Put differently, wherever an `SSLContext` is passed, you're proposing it be replaced by `org.http4s.ssl.SSLContextOption` (or whatever it ends up being called)?

blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala
kevinmeredith: > I think this argument could and should go away.

Once I remove it, what would the default be, @rossabaker?

kevinmeredith: Nevermind. I can see that `TryDefaultSSLContext` is the sensible default.

I've addressed your comment, @rossabaker, via https://github.com/http4s/http4s/pull/3422/commits/1122f6444ad1caa3b03b77587d1ddd51841cd36a.

kevinmeredith: As a follow-up, do you think it's worthwhile to leave, but deprecate, the existing `apply(ExecutionContext, Option[SSLContext]`, @rossabaker?

blaze-client/src/main/scala/org/http4s/client/blaze/Http1Client.scala
```diff
@@ -25,7 +26,7 @@ object Http1Client {
   private def resource[F[_]](config: BlazeClientConfig)(implicit
       F: ConcurrentEffect[F]): Resource[F, Client[F]] = {
     val http1: ConnectionBuilder[F, BlazeConnection[F]] = new Http1Support(
-      sslContextOption = config.sslContext,
+      sslContextOption = config.sslContext.fold[SSLContextOption](SSLContextOption.NoSSL)(SSLContextOption.Provided),
```
kevinmeredith: Hi @rossabaker - please take a look at this one per my commit's message. Thanks

blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala
kevinmeredith: I believe that I've made the necessary changes to use `SSLContextOption`. Please take a look at https://github.com/http4s/http4s/pull/3422#commits-pushed-9c170ef, @rossabaker.

kevinmeredith:
I know that `test:compile` is broken. However, I'm awaiting Ross's thoughts on https://github.com/http4s/http4s/pull/3422#discussion_r425119221 before proceeding - to confirm/deny I'm moving in the right direction.

blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala
```diff
@@ -147,10 +147,6 @@ sealed abstract class BlazeClientBuilder[F[_]] private (
   def withDefaultSslContext: BlazeClientBuilder[F] =
     withSslContext(SSLContext.getDefault())
 
-  /** Use some provided `SSLContext` when making secure calls, or disable secure calls with `None` */
-  def withSslContextOption(sslContext: Option[SSLContext]): BlazeClientBuilder[F] =
```
kevinmeredith, line 149: Per my commit's message, if `SSLContextOption` is truly going to be internal, then I think it's worthwhile to reduce its accessibility via `private [X] ...`. Thoughts, @rossabaker?

rossabaker: I guess I'd leave this one deprecated and fold it into `NoSSL` and `Provided`, but I've been critcized for overusing deprecation. :laughing: 

But I think it's legimtiate.  This is an old implementation detail that leaked, but it doesn't cost us anything to preserve the old with a note, so anyone calling it has a gentle path to the new.

blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala
```diff
@@ -276,14 +274,33 @@ sealed abstract class BlazeClientBuilder[F[_]] private (
 
 object BlazeClientBuilder {
 
+  /**
+    * Indicates how to resolve SSLContext.
+    *  * NoSSL                = do not use SSL/HTTPS
+    *  * TryDefaultSSLContext = `SSLContext.getDefault()`, or `None` on systems where the default is unavailable
+    *  * Provided             = use the explicitly passed SSLContext
+    */
+  sealed trait SSLContextOption extends Product with Serializable
```
rossabaker, line 283: I think this type can be `private[BlazeClientBuilder]`, or moved to `org.http4s.internal` if we apply this treatment to other backends.

blaze-client/src/main/scala/org/http4s/client/blaze/Http1Client.scala
rossabaker: That looks right.  It's the same fold you'd use in preserving the old `withSslContextOption` signature.

blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala
rossabaker: Yes.  You'll probably have to eliminate the default argument in the deprecated one, and `TryDefaultSSLContext` will just be the default on the builder itself.

blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala
kevinmeredith: > I guess I'd leave this one deprecated and fold it into NoSSL and Provided

👍

> This is an old implementation detail that leaked, but it doesn't cost us anything to preserve the old with a note,

 👍

Kazark: FWIW I am very much in favor of deprecation if it leads to a discoverable migration path. 👍

rossabaker: I'm glad I'm not the only one who likes that approach. :smile:

Kazark: At my last job I was a build engineer. The thing that ended up being a big self-caused problem was that the initial way I set up the builds, while a vast overall improvement on what they'd had before, was not as discoverable as it should have been. As a result, on my second iteration I had a rule for myself that when I went to manually test things, I would start with a completely empty scenario---something where I had none of the build dependencies/files in place that were necessary for the build to succeed, and then I would walk through the whole process of setting it up by making sure that the errors that were given me were sufficient for me to fix the problem from the information provided in them alone. It actually turned out to be a very useful exercise in reflecting on what I had built and it even helped me as a user when I forgot to do something (because yes sometimes I was the "dumb" user). Since then, I've had a strong belief in discoverability. I spent too many hours explaining and re-explaining the same things to developers because of my first iteration. I decided that my time was better spent up front in improving the UX rather than down the road in doing support.

kevinmeredith: > I decided that my time was better spent up front in improving the UX rather than down the road in doing support.

💯

blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala
kevinmeredith: Hi @rossabaker - please see https://github.com/http4s/http4s/pull/3422/commits/63622ab6fe59d43ea58bda2d1440809dc2000232, namely its message. Let me know, please, if you recommend another approach.

blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala
```diff
@@ -146,12 +148,17 @@ sealed abstract class BlazeClientBuilder[F[_]] private (
     withSslContext(SSLContext.getDefault())
 
   /** Use some provided `SSLContext` when making secure calls, or disable secure calls with `None` */
+  @deprecated(
+    message =
+      "Use withDefaultSslContext, withSslContext or withoutSslContext to set the SSLContext",
+    since = "1.0.0")
```
kevinmeredith, line 154: Can you please confirm/deny that this `since` value is correct, @rossabaker?

blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala
kevinmeredith: Deprecated - https://github.com/http4s/http4s/pull/3422/commits/4215c42dee5ce53bd4b1deb4aec9a7d52c081f48

blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala
```diff
@@ -307,6 +333,20 @@ object BlazeClientBuilder {
       channelOptions = ChannelOptions(Vector.empty)
     ) {}
 
+  /** Creates a BlazeClientBuilder
+    *
+    * @param executionContext the ExecutionContext for blaze's internal Futures
+    * @param sslContext Some `SSLContext.getDefault()`, or `None` on systems where the default is unavailable
+    */
+  @deprecated(message = "Use BlazeClientBuilder#apply(ExecutionContext).", since = "1.0.0")
```
kevinmeredith, line 341: TODO - look at Ross's answer to https://github.com/http4s/http4s/pull/3422/files#r426109792 to confirm this version is accurate.

rossabaker: This will probably appear in 1.0.0-M1.  If not, we'll clean it up at release time.

kevinmeredith:
Hi @rossabaker - please take a look at https://github.com/http4s/http4s/pull/3422#discussion_r426109334. Other than that, I don't think there's any other comments for to address. Thanks

blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala
rossabaker: I think it could probably be straight up `private`, or `private[BlazeClientBuilder]`, but if the compiler disagrees, listen to it.

kevinmeredith: > I think this type can be private[BlazeClientBuilder], or moved to org.http4s.internal if we apply this treatment to other backends.

I moved `SSLContextOption` to `org.http4s.internals`.

Please take a look at https://github.com/http4s/http4s/pull/3422#commits-pushed-8643f36, @rossabaker @hamnis.

Thanks.

> but if the compiler disagrees, listen to it.

Of course! :-)

kevinmeredith:
I closed, and then re-opened, to prompt the tests to re-run. I'm guessing https://github.com/http4s/http4s/pull/3422/checks?check_run_id=687519041 was a flaky test failure since all other builds succeeded.

rossabaker:
That `TimeoutSpec` is flaky on CI.  That's not related to this.

kevinmeredith:
> That TimeoutSpec is flaky on CI. That's not related to this.

Thanks, @rossabaker. As a result, I assume that I don't need to get all tests to pass successfully, i.e. it's good as-is?

PR #3424, opened 2020-05-14
Update compatible dependencies on 0.21.x

rossabaker:
* ~~boopickle-1.3.2 (supersedes #3419)~~
* json4s-3.6.8
* log4cats-1.1.1 (supersedes #3418)
* metrics-4.1.7 
* okhttp-4.6.0
* scalatags-0.9.1
* specs2-4.9.4
* Adds filters so dependencyUpdates runs clean

### DIFF
--- a/project/Http4sPlugin.scala
+++ b/project/Http4sPlugin.scala
@@ -67,7 +67,24 @@ object Http4sPlugin extends AutoPlugin {
       IO.write(dest, buildData)
     },
 
-    dependencyUpdatesFilter -= moduleFilter(organization = "javax.servlet"), // servlet-4.0 is not yet supported by jetty-9 or tomcat-9, so don't accidentally depend on its new features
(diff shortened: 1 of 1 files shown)


rossabaker:
We have a [real problem in boopickle](https://github.com/http4s/http4s/runs/678905617#step:11:57). This looks like what happens when a library builds on a later JDK than 8 and then runs on 8, but curiously it's not affecting the 2.13.2/1.8 build.

@cquiroz, I think you've been involved on that project. Do you know what happened here?

cquiroz:
It looks like the build was done with another jdk. There  are a few troubles on scala.js as well
There is a PR to help standardize the build but it is awaiting review

rossabaker:
Okay. I think 1.3.2 was a cursed release. Let's back that out and move forward with the rest.

rossabaker:
Smaller than the one that was already approved, so merging to reduce clutter.

PR #3426, opened 2020-05-14
Fix dotty runtime error

jcouyang:
## reproduce
in [`http4s-finagle`](https://github.com/http4s/http4s-finagle)
```
sbt "++ 0.24.0-RC1 test"
```
will get runtime exception
```
May 14, 2020 1:21:20 PM com.twitter.util.RootMonitor$$anonfun$2 applyOrElse
SEVERE: Fatal exception propagated to the root monitor!
java.lang.AbstractMethodError: Method org/http4s/dsl/io$.org$http4s$dsl$impl$Methods$_setter_$GET_$eq(Lorg/http4s/Method$Semantics$Safe;)V is abstract
	at org.http4s.dsl.io$.org$http4s$dsl$impl$Methods$_setter_$GET_$eq(io.scala)
	at org.http4s.dsl.impl.Methods.$init$(Methods.scala:12)
	at org.http4s.dsl.io$.<clinit>(io.scala:11)
```
## fix
publishLocal and rerun `sbt "++ 0.24.0-RC1 test"` in `http4s-finagle`
dotty runtime is not binary compat to initialize `PermitsBody` because it extends an abstract class
well I found it is unnecessary to extends though because all methods are `new Method` extended these two trait, kind of duplicated

### DIFF
--- a/core/src/main/scala/org/http4s/Method.scala
+++ b/core/src/main/scala/org/http4s/Method.scala
@@ -44,8 +44,8 @@ object Method {
   }
 
   // Type tags for a method allowing a body or not
-  sealed trait PermitsBody
-  sealed trait NoBody
+  sealed trait PermitsBody { _: Method => }
+  sealed trait NoBody { _: Method => }
 
   def fromString(s: String): ParseResult[Method] =
     allByKey.getOrElse(
--- a/core/src/main/scala/org/http4s/Method.scala
+++ b/core/src/main/scala/org/http4s/Method.scala
@@ -44,8 +44,8 @@ object Method {
   }
 
   // Type tags for a method allowing a body or not
-  sealed trait PermitsBody extends Method
-  sealed trait NoBody extends Method
+  sealed trait PermitsBody
+  sealed trait NoBody
 
   def fromString(s: String): ParseResult[Method] =
     allByKey.getOrElse(
--- a/core/src/main/scala/org/http4s/Method.scala
+++ b/core/src/main/scala/org/http4s/Method.scala
@@ -46,6 +46,7 @@ object Method {
   // Type tags for a method allowing a body or not
   sealed trait PermitsBody { _: Method => }
   sealed trait NoBody { _: Method => }
+  sealed trait IsPermitsBody extends Method with PermitsBody
 
   def fromString(s: String): ParseResult[Method] =
     allByKey.getOrElse(
--- a/core/src/main/scala/org/http4s/Method.scala
+++ b/core/src/main/scala/org/http4s/Method.scala
@@ -46,7 +46,6 @@ object Method {
   // Type tags for a method allowing a body or not
   sealed trait PermitsBody { _: Method => }
   sealed trait NoBody { _: Method => }
-  sealed trait IsPermitsBody extends Method with PermitsBody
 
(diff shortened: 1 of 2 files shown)


jcouyang:
i don't understand, CI complain
```
[error] There are files without headers!
[error]   /home/runner/work/http4s/http4s/async-http-client/src/main/scala/org/http4s/client/asynchttpclient/NoOpCookieStore.scala
```
but I didn't change anything there... :man_shrugging:

core/src/main/scala/org/http4s/Method.scala
```diff
@@ -44,8 +44,8 @@ object Method {
   }
 
   // Type tags for a method allowing a body or not
-  sealed trait PermitsBody extends Method
-  sealed trait NoBody extends Method
+  sealed trait PermitsBody
```
rossabaker, line 47: Would dotty allow these to take a self type of `Method`?  They shouldn't be mixed into anything else.

```scala
sealed trait PermitsBody { self: Method => }
```

If not, this is fine.

rossabaker:
We opened a PR that added a file before sbt-headers, and merged it after.  CI never ran on the merge until it was already on master.  That's what broke your build.  I pushed a fix in 1f009aad74.

core/src/main/scala/org/http4s/Method.scala
jcouyang: let me try...
but this is sealed so only types in this file can extend it, is it really necessary?

jcouyang: it works

rossabaker: Yeah, arguably not necessary since it's sealed, but more intention revealing.  I'd prefer it, but am fine without it if it causes problems.  Dotty support is more important.

rossabaker:
The `IsPermitsBody` seems like it gets us back to about the same place as we were, and nothing I see is extending it.  I'm confused why that test still passes.

jcouyang:
`IsPermitsBody` is replacing the old `PermitsBody` for the unit test

dotty is fine if nothing extends `IsPermitsBody`

rossabaker:
My point is that nothing extends that `IsPermitBody` trait, so the case is never true.  The test passes if we remove that case, so we don't need to clutter `Method` with the new trait.

jcouyang:
oh, now it is weird when you point out... :thinking: should I really delete the test case?

rossabaker:
I looked closer at it while attempting to simplify the whole method-properties-by-subtyping mess, and that test is marked `pendingUntilFixed`.  This dates back to before the body was fs2, and we could examine the algebra to determine if there were any effects.  When we lost that, we lost some functionality in this middleware.

I don't think we should introduce a new trait to core to support a test that doesn't pass.  This says what we need in the test, and keeps core clean:

```scala
case (POST | PUT) if body.nonEmpty =>
```

jcouyang:
are we waiting for another review to merge?

rossabaker:
We usually look for two reviews on code, one on docs. I like to make sure people have a chance to weigh in, but I also don't like good work to be held up, and reviews come in bursts on a volunteer project.  Let's merge it, and if anyone sees something we missed, there's always the next PR.

PR #3440, opened 2020-05-18
Simplify Method model

rossabaker:
* Drops `PermitsBody` and `NoBody` mixins.  These were used as marker traits on the constants so we could create a client DSL that does not accept an entity on methods where they're disallowed. It seems like a lot of magic to prevent someone from writing `TRACE("oops this is illegal", uri"http://www.example.com/")`.  I think this supersedes #3426, but we can merge that while people debate this.

* Drops `Semantics` trait.  The name appears in no spec, and nobody was using these fields at the type level.  All we need are a pair of booleans to reflect the full IANA registry.

* `Method` has been a case class since the Spray fork, but it's not a proper product: not all Strings are a valid `name`s, and each `name` uniquely defines `isSafe` and `isIdempotent`.


core/src/main/scala/org/http4s/Method.scala
```diff
@@ -8,44 +8,45 @@ package org.http4s
...
 import cats.{Eq, Show}
-import org.http4s.Method.Semantics
 import org.http4s.parser.Rfc2616BasicRules
 import org.http4s.util.{Renderable, Writer}
+import scala.util.hashing.MurmurHash3
 
 /**
   * An HTTP method.
   *
+  * @param name The name of the method
```
rossabaker, line 18: I could be cajoled into calling this `toString`.

client/src/test/scala/org/http4s/client/middleware/FollowRedirectSpec.scala
```diff
@@ -75,7 +75,7 @@ class FollowRedirectSpec
       ) = {
         val u = uri("http://localhost") / status.code.toString
         val req: Request[IO] = method match {
-          case _: Method.PermitsBody if body.nonEmpty =>
+          case (OPTIONS | PATCH | POST | PUT) if body.nonEmpty =>
```
rossabaker, line 78: This could also be preserved as a property of `Method`, but has the same problem as cacheability: someone has to go through all the WebDAV specs.  The answer there is usually "Yes, but a specific type of XML."

rossabaker:
An alternative that would permit the old restrictions in that DSL would be to make every `Method` value an object.  Then each value would have its own type, and we could add syntax for `POST.type`, `GET.type`, etc.  This would make the DSL implementation noisier, and prevent the change of `Method` to a final class.

Another option is to phase out this DSL entirely, which I don't think has worked out well in practice.  I know @SystemFw has argued for phasing it out.  We can make it gentle if we keep code like we did here, but also deprecate it.

core/src/main/scala/org/http4s/Method.scala
```diff
@@ -8,44 +8,45 @@ package org.http4s
...
+  *
+  * @param isSafe Request methods are considered "safe" if their defined
+  * semantics are essentially read-only; i.e., the client does not request, and
+  * does not expect, any state change on the origin server as a result of
+  * applying a safe method to a target resource.
+  *
+  * @param isIdempotent A request method is considered "idempotent" if the
```
kevinmeredith, line 25: Nitpick suggestion: use MDN's definition:

> An HTTP method is idempotent if an identical request can be made once or several times in a row with the same effect while leaving the server in the same state.

https://developer.mozilla.org/en-US/docs/Glossary/Idempotent

hamnis:
needs a conflict resolution

rossabaker:
Sneaky breaking change: this removes `Method.unapply`.

client/src/main/scala/org/http4s/client/Http4sClientDsl.scala
```diff
@@ -30,13 +24,23 @@ trait Http4sClientDsl[F[_]] {
   }
 }
 
-object Http4sClientDsl {
+class MethodOps[F[_]](private val method: Method) extends AnyVal {
+
+  /** Make a [[org.http4s.Request]] using this [[Method]] */
+  final def apply(uri: Uri, headers: Header*)(implicit F: Applicative[F]): F[Request[F]] =
```
kevinmeredith, line 30: Could you please say why the type of `F[Request[F]]` rather than just `Request[F]`? Thanks

rossabaker: Uh, historical reasons?  `EntityEncoder` used to be effectful, which is why the overload is effectful, and this was effectful so the types didn't change when a body is added.  But that hasn't been true for a long time, and we'd like to eliminate the overloads in the client that take an effect of a request, at which case these would break.

I don't like this DSL anyway, but I think there's really no reason for it to continue to be effectful.

ChristopherDavenport:
If this removes unapply, how does that effect the dsl?

rossabaker:
I don't think appreciably.  Unless people were doing something like: 

```scala
case Method(name) -> Root / "blah" => 
```

hamnis (approved):

core/src/main/scala/org/http4s/Method.scala
```diff
@@ -8,44 +8,45 @@ package org.http4s
...
+  * intended effect on the server of multiple identical requests with that
+  * method is the same as the effect for a single such request.
+  *
   * @see [http://tools.ietf.org/html/rfc7231#section-4 RFC7321, Section 4]
   * @see [http://www.iana.org/assignments/http-methods/http-methods.xhtml IANA HTTP Method Registry]
   */
-sealed abstract case class Method private (name: String) extends Renderable with Semantics {
+final class Method private (val name: String, val isSafe: Boolean, val isIdempotent: Boolean)
```
kevinmeredith, line 32: Side question: technically all Safe HTTP Request's Methods are idempotent, no, @rossabaker?

Note that I'm not proposing any change, but asking this question out of curiosity (after digging through my Github notifications).

rossabaker: That's correct.  No public constructor can be safe without also being idempotent.

An alternative and perhaps better design would be to consolidate these two booleans into the old `Semantics` trait, but no longer make it a mixin to the `Method`.

PR #3450, opened 2020-05-22
Model Accept-Patch header

cquiroz:
As part of modeling headers, here's `Accept-Patch`

### DIFF
--- a/core/src/main/scala/org/http4s/Header.scala
+++ b/core/src/main/scala/org/http4s/Header.scala
@@ -120,6 +120,17 @@ object Header {
     }
   }
 
+  /** Helper trait that provides a default way of rendering the value provided a Renderer */
+  trait RecurringRenderer extends Recurring {
+    type Value
+    implicit val renderer: Renderer[Value]
+    override def renderValue(writer: Writer): writer.type = {
+      renderer.render(writer, values.head)
+      values.tail.foreach(writer << ", " << Renderer.renderString(_))
+      writer
+    }
+  }
+
   implicit val HeaderShow: Show[Header] = Show.show[Header] {
     _.toString
   }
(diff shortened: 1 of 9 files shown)


core/src/main/scala/org/http4s/Header.scala
```diff
@@ -120,6 +120,17 @@ object Header {
     }
   }
 
+  /** Helper trait that provides a default way of rendering the value provided a Renderer */
+  trait RecurringRenderer extends Recurring {
```
cquiroz, line 124: This is equivalent to `RecurringRenderable` but for `Renderer

core/src/main/scala/org/http4s/Header.scala
```diff
@@ -120,6 +120,17 @@ object Header {
...
+    type Value
+    implicit val renderer: Renderer[Value]
```
cquiroz, line 126: Any ideas on how to initialize this by default? My attempts ended poorly 😄

tests/src/test/scala/org/http4s/HeadersSpec.scala
```diff
@@ -46,7 +46,7 @@ class HeadersSpec extends Http4sSpec {
         Header("Accept-Patch", ""),
         Header("Access-Control-Allow-Credentials", "")
       )
-      headers.get(`Accept-Patch`).map(_.value) must beSome("")
+      headers.get(`Access-Control-Allow-Credentials`).map(_.value) must beSome("")
```
cquiroz, line 49: I'm not sure about the meaning of this test

core/src/main/scala/org/http4s/Header.scala
rossabaker: Eager abstract values make me nervous. How about a def?

tests/src/test/scala/org/http4s/HeadersSpec.scala
rossabaker: I think it was just making sure that we could get headers when we had a key but not a full model.  I don't think it's a great test, and what you did is fine.

core/src/main/scala/org/http4s/Header.scala
cquiroz: sure

cquiroz:
Rebased and fixed comment

PR #3451, opened 2020-05-22
JettyBuilder PrefixedServices

CharlesD91:
**What**

New function `def mountPrefixedServices(prefixed: PrefixedServices[F])` that allows to mount all combined `(service, prefix)` at once

**Why**

In our app we have the `List(service, prefix)` layer and then the `Jetty` layer. We want to split those 2 layers in a clean way (return type of the first layer: ` PrefixedServices`) so that we can then test it without starting the `Jetty`

### DIFF
--- a/jetty/src/main/scala/org/http4s/server/jetty/JettyBuilder.scala
+++ b/jetty/src/main/scala/org/http4s/server/jetty/JettyBuilder.scala
@@ -10,6 +10,7 @@ package jetty
 
 import cats.effect._
 import cats.implicits._
+import cats.Semigroup
 import java.net.InetSocketAddress
 import java.util
 import javax.net.ssl.{SSLContext, SSLParameters}
@@ -172,6 +173,10 @@ sealed class JettyBuilder[F[_]] private (
 
   def mountService(service: HttpRoutes[F], prefix: String): Self =
     mountHttpApp(service.orNotFound, prefix)
+
+  def mountPrefixedServices(prefixed: PrefixedServices[F]): Self =
+    prefixed.services.foldLeft(this)((builder, prefixedService) =>
+      builder.mountService(prefixedService.service, prefixedService.prefix))
 
   def mountHttpApp(service: HttpApp[F], prefix: String): Self =
     copy(mounts = mounts :+ Mount[F] { (context, index, builder) =>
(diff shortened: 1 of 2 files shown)


jetty/src/main/scala/org/http4s/server/jetty/JettyBuilder.scala
```diff
@@ -388,3 +393,15 @@ object JettyBuilder {
 }
 
 private final case class Mount[F[_]](f: (ServletContextHandler, Int, JettyBuilder[F]) => Unit)
+
+case class PrefixedService[F[_]](service: HttpRoutes[F], prefix: String)
```
rossabaker, line 397: Most of our case classes are final.  I think this could be.

jetty/src/main/scala/org/http4s/server/jetty/JettyBuilder.scala
```diff
@@ -388,3 +393,15 @@ object JettyBuilder {
...
+class PrefixedServices[F[_]] private (val services: List[PrefixedService[F]])
+
+object PrefixedServices {
+  def apply[F[_]](service: HttpRoutes[F], prefix: String) =
+    new PrefixedServices[F](List(PrefixedService(service, prefix)))
+
+  implicit def prefixedServicesSemigroup[F[_]]: Semigroup[PrefixedServices[F]] =
+    (x: PrefixedServices[F], y: PrefixedServices[F]) =>
+      new PrefixedServices[F](x.services ++ y.services)
+}
```
rossabaker, line 407: All of this seems as germane to Tomcat as it does to Jetty.  Would it be better in the servlet package?

jetty/src/main/scala/org/http4s/server/jetty/JettyBuilder.scala
```diff
@@ -388,3 +393,15 @@ object JettyBuilder {
...
 private final case class Mount[F[_]](f: (ServletContextHandler, Int, JettyBuilder[F]) => Unit)
+
+case class PrefixedService[F[_]](service: HttpRoutes[F], prefix: String)
+class PrefixedServices[F[_]] private (val services: List[PrefixedService[F]])
+
+object PrefixedServices {
+  def apply[F[_]](service: HttpRoutes[F], prefix: String) =
+    new PrefixedServices[F](List(PrefixedService(service, prefix)))
+
+  implicit def prefixedServicesSemigroup[F[_]]: Semigroup[PrefixedServices[F]] =
```
rossabaker, line 404: I think this could even be a `Monoid` with an empty list of services.

I like to law check our instances, but we have no notion of equality for `HttpRoutes`.  But I feel pretty good about this being lawful.

CharlesD91:
Hi @rossabaker , thank you very much for your answer. 

Doing the routing in the `HttpRoutes` using `Router` worked very well for my use case. I should have used it right away, instead of trying to do the routing at the `ServletContextHandler.addServlet(ServletHolder servlet, String pathSpec)` level. 

I don't think the performances will be any different so I close this PR.

PR #3456, opened 2020-05-25
Fixes #2049 Silence stack traces in test output

### LINKED ISSUE #2049: Silence expected stack traces in test output

ashwinbhaskar:


testing/src/test/scala/org/http4s/testing/ErrorReporting.scala
```diff
@@ -9,11 +9,81 @@
...
+      val domains = new Array[ProtectionDomain](1)
+      domains(0) = domain
+      val context = new AccessControlContext(domains)
+
+      // Redirect output to dummy stream
+      val fakeOutStream = new PrintStream(new ByteArrayOutputStream)
+      val fakeErrStream = new PrintStream(new ByteArrayOutputStream)
+      System.setOut(fakeOutStream)
+      System.setErr(fakeErrStream)
+      val result = java.security.AccessController.doPrivileged(
```
rossabaker, line 55: Can this throw?  I'd expect to see a finally block on the reset, but I'm not so familiar on this API.

testing/src/test/scala/org/http4s/testing/ErrorReporting.scala
```diff
@@ -9,11 +9,81 @@
...
+      val code: CodeSource = new CodeSource(new URL("file:/*"), certs)
+
+      val domain = new ProtectionDomain(code, perm)
+      val domains = new Array[ProtectionDomain](1)
+      domains(0) = domain
+      val context = new AccessControlContext(domains)
+
+      // Redirect output to dummy stream
+      val fakeOutStream = new PrintStream(new ByteArrayOutputStream)
+      val fakeErrStream = new PrintStream(new ByteArrayOutputStream)
```
rossabaker, line 52: This is good if we're introspecting the logged values, but it's possible to write a NullOutputStream that doesn't bother to construct the array internally.  (Here's a [Java version](https://github.com/apache/commons-io/blob/master/src/main/java/org/apache/commons/io/output/NullOutputStream.java)).  If a test logs a lot, this will create a large array to be thrown away.  It's probably not a problem in practice, but might be a nice cleanup.

ashwinbhaskar: done

testing/src/test/scala/org/http4s/testing/ErrorReporting.scala
ashwinbhaskar: This can throw a `AccessControlException` in the `doPrivileged` block if the code being executed inside it tries to redirect `System.out` or `System.err` output streams. We cannot catch it because then we would have to return some result of type `R`. If there is a code that is trying to fiddle with `System.out` (redirecting out put streams), then `silenceOutputStreams` method cannot be used to surround it with.

rossabaker: But even if it throws, the test will error, and we can keep running other tests, right?  I don't think there should be a catch here, but a finally still makes sense to me.  Or are you saying anything we throw from here is to be treated as fatal and invalidates any other tests?

ashwinbhaskar: What should go in the finally block?

rossabaker: Lines 62 and 63.  That way, even if `doPrivileged` throws, the outputs will be restored. the exception will be thrown, the test framework will record it, and we can try to run more tests in a clean state.

ashwinbhaskar: ah okay, got it. Done

PR #3460, opened 2020-05-28
Remove deprecated *Service

hamnis:
Removes *Service types and Constructors as they are redundant and have been deprecated for a few releases.


ChristopherDavenport:
I really don't mind either way. But I hate the tail of the dead code. So I'm ok with this. It has been 2 versions. If we are going to hold onto stuff perhaps we should define what an "appropriate" period to upgrade is.

rossabaker:
Well, I'm also much more open to this if we keep upping our scalafix game.

PR #3463, opened 2020-05-31
Model Access Control Allow Credentials header

cquiroz:
Another header for #2011


tests/src/test/scala/org/http4s/HeadersSpec.scala
```diff
@@ -40,15 +40,6 @@ class HeadersSpec extends Http4sSpec {
       headers.get(org.http4s.headers.Cookie).map(_.values.length) must beSome(2)
     }
 
-    "Find the headers with DefaultHeaderKey keys" in {
```
cquiroz, line 43: I removed this test in my last commit as it uses two headers now properly modeled. @rossabaker Do you think that's ok?

PR #3464, opened 2020-06-01
Fix error reporting and infinite loops in util.decode

### LINKED ISSUE #3453: util.decode doesn't terminate on malformed input
When the `decode` function transforms stream encoded in another charset than expected, it may never terminate due to `CharsetDecoder.decode` errors being ignored.
Example:
```
val source = Stream.emits(Seq(0xb3.toByte)) // 'ł' in windows-1250 encoding
decode(Charset.`UTF-8`)(source).toList // incorrect encoding provided
```
`CharsetDecoder.decode` returns an error in such a case without consuming data from `byteBuffer` thus causing an infinite loop.

This part of `decode`:
```
decoder.decode(byteBuffer, charBuffer, false)
val nextStream = stream.consChunk(Chunk.byteBuffer(byteBuffer.slice()))
```
should be probably rewritten to something like this:
```
val coderResult = decoder.decode(byteBuffer, charBuffer, false)
if(coderResult.isError)
  Pull.raiseError[F](new CharacterCodingException)
else {
  val nextStream = stream.consChunk(Chunk.byteBuffer(byteBuffer.slice()))
}
```
This will however require to have an implicit `ApplicativeError` for `F`.

rossabaker:
There are two deficiencies in our decoder.

The first is that we have not been checking the Charset decoder results.  When they are a character coding error, we need to raise that into the pull.  We have tried to size out buffers to avoid any chance of overflow, but we should also raise a bug in this case.

The second is that we are consing leftover bytes onto the stream instead of incoporating them into the next chunk.  When we have a valid, incomplete chunk, this sends us into an infinite loop, repeatedly attempting to decode the incomplete input.  We need to carry it over so it can be combined with the next chunk, or fed to the decoder as a final chunk.

Fixes #3453.

### DIFF
--- a/core/src/main/scala/org/http4s/util/package.scala
+++ b/core/src/main/scala/org/http4s/util/package.scala
@@ -10,39 +10,65 @@ import cats.ApplicativeError
 import fs2._
 import java.nio.{ByteBuffer, CharBuffer}
 import java.nio.charset.StandardCharsets
+import org.http4s.internal.{bug => bug0}
 import scala.concurrent.ExecutionContextExecutor
 
 package object util {
   private val utf8Bom: Chunk[Byte] = Chunk(0xef.toByte, 0xbb.toByte, 0xbf.toByte)
 
-  def decode[F[_]](charset: Charset): Pipe[F, Byte, String] = {
+  def decode[F[_]](charset: Charset): Pipe[F, Byte, String] = { in =>
     val decoder = charset.nioCharset.newDecoder
     val maxCharsPerByte = math.ceil(decoder.maxCharsPerByte().toDouble).toInt
     val avgBytesPerChar = math.ceil(1.0 / decoder.averageCharsPerByte().toDouble).toInt
     val charBufferSize = 128
 
-    _.repeatPull[String] {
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/http4s/util/package.scala
```diff
@@ -10,39 +10,65 @@ import cats.ApplicativeError
...
+              val byteBuffer = ByteBuffer.wrap(bytes)
+              val charBuffer = CharBuffer.allocate(bytes.length * maxCharsPerByte)
+              decoder.decode(byteBuffer, charBuffer, true) match {
+                case result if result.isError || result.isOverflow =>
+                  result.throwException()
+                  throw bug0(s"Expected to have thrown an exception from ${result}")
+                case _ =>
+                  decoder.flush(charBuffer) match {
+                    case result if result.isError || result.isOverflow =>
+                      result.throwException()
```
rossabaker, line 40: This (and again below) is bad: we want to raise the error into Pull, but can't because we don't have a `RaiseThrowable` constraint.  We can't add one to `Message#bodyAsText` or `EntityDecoder[UrlForm]` without binary breaking changes.

rossabaker:
I think we can deprecate our way out of this in a way that doesn't throw willy nilly.

PR #3466, opened 2020-06-02
Fix infinite loops in util.decode

### LINKED ISSUE #3453: util.decode doesn't terminate on malformed input

rossabaker:
* Replaces `decode` with an internal version with a `RaiseThrowable` constraint.
* Check the error codes on each operation.  Raise errors on malformed input, rather than loop infinitely.
* The new version allocates one CharBuffer and ByteBuffer, and should be more efficient.
* `Media#bodyAsText` is renamed to `Media#bodyText`, so we can add the constraint.
* `EntityDecoder.decodeString` renamed to `Entity.decodeText` for the same reason.

Fixes #3453.


core/src/main/scala/org/http4s/Media.scala
```diff
@@ -8,23 +8,32 @@ package org.http4s
...
     charset.getOrElse(defaultCharset) match {
       case Charset.`UTF-8` =>
         // suspect this one is more efficient, though this is superstition
         body.through(utf8Decode)
       case cs =>
-        body.through(decode(cs))
+        body.through(util.decode(cs))
+    }
+
+  final def bodyText(implicit RT: RaiseThrowable[F], defaultCharset: Charset = DefaultCharset): Stream[F, String] =
```
rossabaker, line 30: I don't love this implicit charset pattern that we have.  Maybe it's time to rip that out.

core/src/main/scala/org/http4s/EntityDecoder.scala
```diff
@@ -181,11 +181,18 @@ object EntityDecoder {
...
+  @deprecated(
+    "Can go into an infinite loop for charsets other than UTF-8. Replaced by decodeText",
+    "0.21.5")
   def decodeString[F[_]](
       m: Media[F])(implicit F: Sync[F], defaultCharset: Charset = DefaultCharset): F[String] =
     m.bodyAsText.compile.string
 
+  /** Decodes a message to a String */
+  def decodeText[F[_]](
+      m: Media[F])(implicit F: Sync[F], defaultCharset: Charset = DefaultCharset): F[String] =
```
rossabaker, line 193: This is a case where we could take a `RaiseThrowable[F]` and `Compiler[F, F]` as evidence, since `Sync` is offensive.  A `Compiler[F, G]` would be neat, but the `G` would ripple everywhere.

core/src/main/scala/org/http4s/internal/package.scala
```diff
@@ -203,4 +207,73 @@ package object internal {
...
+                flushLoop.as(None)
+              case _ if result.isOverflow =>
+                out.as(Some(Stream.empty))
+              case _ if result.isMalformed =>
+                Pull.raiseError(new MalformedInputException(result.length()))
+              case _ if result.isUnmappable =>
+                Pull.raiseError(new UnmappableCharacterException(result.length()))
+            }
+          case Some((chunk, stream)) =>
+            val chunkWithoutBom = skipByteOrderMark(chunk)
```
hamnis, line 254: is this run on every chunk? or am I reading this wrong?

rossabaker: It is, and that's a pre-existing defect, which we were reminded of on the original ticket.  I haven't had time to clean that up, but I wouldn't mind merging this while someone finds time to fix that.

PR #3472, opened 2020-06-05
Implement httpRoutes and httpApp constructors for EntityLimiter 

voidcontext:
This PR is part of #2402.

Note: this time I opened the PR against the series/0.21 branch, I hope this makes things easier (please let me know if it should be a different branch).


rossabaker:
That's exactly right. As long as it's binary compatible, that's the correct branch.  Thanks for resuming these!

PR #3478, opened 2020-06-08
DRY (Don't Repeat Yourself) up Logger#logMessage

kevinmeredith:
Both [org.http4s.server.middleware.Logger#logMessage](https://github.com/http4s/http4s/blob/v0.21.4/server/src/main/scala/org/http4s/server/middleware/Logger.scala#L52-L98) and [org.http4s.client.middleware.Logger#logMessage](https://github.com/http4s/http4s/blob/v0.21.4/client/src/main/scala/org/http4s/client/middleware/Logger.scala#L26-L72) are the same.

As a result, I copied their same definition to `org.http4s.internal.Logger#logMessage`, and then referenced that definition from both call sites.

Could you please review this one, @rossabaker, @ChristopherDavenport and @bpholt? 

Side note - Brian, we sat next to each other at NEScala in Philly 1-2 years ago. I'm the guy from Banno/Jack Henry. I hope you're well! Also, I reached out to you, Brian, per https://github.com/http4s/http4s/commit/801b58ce8dd2d068c64008858b5b6325ba2e14fd.

Thanks

### DIFF
--- a/client/src/main/scala/org/http4s/client/middleware/Logger.scala
+++ b/client/src/main/scala/org/http4s/client/middleware/Logger.scala
@@ -32,5 +32,6 @@ object Logger {
       logBody: Boolean,
       redactHeadersWhen: CaseInsensitiveString => Boolean = Headers.SensitiveHeaders.contains)(
       log: String => F[Unit])(implicit F: Sync[F]): F[Unit] =
-    org.http4s.internal.Logger.logMessage[F, A](message)(logHeaders, logBody, redactHeadersWhen)(log)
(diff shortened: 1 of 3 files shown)


kevinmeredith:
I restarted the build since test(s) appeared to fail unrelated to this PR.

bpholt:
Hey Kevin! Nice to hear from you! Thanks for cleaning this up—I don't remember why I needed to copy the code instead of moving to a common place originally, but I like having it in one spot instead!

Neat trick to restart the build by closing and reopening the PR, too. I haven't seen that before.

PR #3486, opened 2020-06-10
Add org.http4s.internal.Logger#logMessageWithBodyText

kevinmeredith:
My primary motivation of this PR is to, via this Logger for `client`, offer a function, Option[String => F[String]]. It supplies the option of further modifying the String body.

In my case, I'm interested to transform the String into Json, sanitize it, and then print out the sanitized JSON. By "sanitize," I mean to hide/obscure JSON keys' values that are sensitive.

I intend, in a follow-up PR, to make such a change for `org.http4s.server.middleware`'s `Request/ResponseLogger`.

### DIFF
--- a/core/src/main/scala/org/http4s/internal/Logger.scala
+++ b/core/src/main/scala/org/http4s/internal/Logger.scala
@@ -19,6 +19,12 @@ object Logger {
       logBody: Boolean,
       redactHeadersWhen: CaseInsensitiveString => Boolean = Headers.SensitiveHeaders.contains)(
       log: String => F[Unit])(implicit F: Sync[F]): F[Unit] = {
+
+  def logMessageWithBodyText[F[_], A <: Message[F]](message: A)(
+      logHeaders: Boolean,
+      logBodyText: Option[String => F[String]],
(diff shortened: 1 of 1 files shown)


kevinmeredith:
In the signature,  `logBodyText: Stream[F, Byte] => Option[F[String]]`, can you please explain the reason for `Option[F[String]]` rather than just `F[String]`, @rossabaker? I'm not sure I follow why it's optional.
`

rossabaker:
Oh, I thought you were using that Option to determine whether to log the body or not.  If not, no, i don't think it needs to be there.

kevinmeredith:
Please take a look at my recent changes, @rossabaker and @hamnis: https://github.com/http4s/http4s/pull/3486#commits-pushed-d1427a3. Thanks

kevinmeredith:
Hi Ross and Erlend - can you please take a look? Thanks

kevinmeredith:
Can you please check this one out, @hamnis? I'm asking since I intend to then request your and Ross's review on https://github.com/http4s/http4s/pull/3497. Thanks!

Issue #3485, opened 2020-06-10
X-Request-Id Middleware


Slakah:
We've got an internally used middleware, and wondering whether this would be useful to contribute back into http4s. The basic concept is, propagate the header from request to response, generating a random UUID `X-Request-Id` if it wasn't set.

Gist of the code (older version of http4s as can be seen by presence of `.ci`): https://gist.github.com/Slakah/f5eb4fbe3c6cc1ed28f8bd66482cd566

Docs:
* [Heroku docs on X-Request-Id](https://devcenter.heroku.com/articles/http-request-id)
* [Stackoverflow](https://stackoverflow.com/questions/25433258/what-is-the-x-request-id-http-header)

Does it also make sense to add official support for the header as well in the [`headers` package](https://github.com/http4s/http4s/tree/master/core/src/main/scala/org/http4s/headers), i.e.

```
package org.http4s
package headers

object `X-Requested-With` extends HeaderKey.Default
```

Let me know if this is something you'll be interested in, and I can draft a PR.

hamnis:
Option 2 is reasonable.

Slakah:
Implemented by #3487

PR #3487, opened 2020-06-10
X-Request-ID Middleware

Slakah:
Implements https://github.com/http4s/http4s/issues/3485 .

## TODO

- [x] Update docs
- [x] Deal with `UUID.randomUUID()` blocking (option 2 of https://github.com/http4s/http4s/issues/3485#issuecomment-642243296)
- [x] Include the request id as attributes in both the request and response - suggested by @ChristopherDavenport

### DIFF
--- a/server/src/main/scala/org/http4s/server/middleware/RequestId.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/RequestId.scala
@@ -56,7 +56,6 @@ object RequestId {
       } yield response.putHeaders(header)
     }
 
-
   object httpApp {
     def apply[F[_]: Sync](httpApp: HttpApp[F]): HttpApp[F] =
       RequestId.apply(requestIdHeader)(httpApp)
--- a/server/src/main/scala/org/http4s/server/middleware/RequestId.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/RequestId.scala
@@ -1,0 +1,35 @@
+/*
+ * Copyright 2013-2020 http4s.org
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.http4s
+package server
+package middleware
+
+import org.http4s.{Header, Http, Request, Response}
+import cats.data.Kleisli
+import cats.effect.Sync
+import cats.implicits._
+import org.typelevel.ci.CIString
+import java.util.UUID
+
+/** Propagate a `X-Request-Id` header to the response, generate a UUID
+  * when the `X-Request-Id` header is unset.
+  * https://devcenter.heroku.com/articles/http-request-id
+  */
+object RequestId {
+
+  def apply[G[_], F[_]](headerName: CIString = CIString("X-Request-ID"))(http: Http[G, F])(implicit
+      G: Sync[G]): Http[G, F] =
+    Kleisli[G, Request[F], Response[F]] { req =>
+      for {
+        header <- req.headers.get(headerName) match {
+          case None => G.delay[Header](Header.Raw(headerName, UUID.randomUUID().show))
+          case Some(header) => G.pure[Header](header)
+        }
+        response <- http(req.putHeaders(header))
+      } yield response.putHeaders(header)
+    }
+}
--- a/server/src/main/scala/org/http4s/server/middleware/RequestId.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/RequestId.scala
@@ -23,40 +23,64 @@ import java.util.UUID
   */
 object RequestId {
 
-  private val requestIdHeader = CIString("X-Request-ID")
+  private[this] val requestIdHeader = CIString("X-Request-ID")
+
+  def apply[G[_], F[_]](http: Http[G, F])(implicit G: Sync[G]): Http[G, F] =
+    apply(requestIdHeader)(http)
+
+  def apply[G[_], F[_]](
+      headerName: CIString
+  )(http: Http[G, F])(implicit G: Sync[G]): Http[G, F] =
+    Kleisli[G, Request[F], Response[F]] { req =>
+      for {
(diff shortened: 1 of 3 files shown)


server/src/main/scala/org/http4s/server/middleware/RequestId.scala
```diff
@@ -0,0 +1,62 @@
...
+  * https://devcenter.heroku.com/articles/http-request-id
+  */
+object RequestId {
+
+  private val requestIdHeader = CIString("X-Request-ID")
+
+  def apply[G[_], F[_]](
+      fk: F ~> G,
+      headerName: CIString = requestIdHeader,
+      genReqId: Option[F[UUID]] = None
```
rossabaker, line 31: I wonder about two constructors here instead of this optional parameter, because if we provide some generator, I don't think we need the `Sync` constraint.

server/src/main/scala/org/http4s/server/middleware/RequestId.scala
```diff
@@ -0,0 +1,62 @@
...
+        header <- fk(req.headers.get(headerName) match {
+          case None => gen.map(reqId => Header.Raw(headerName, reqId.show))
+          case Some(header) => F.pure[Header](header)
+        })
+        response <- http(req.putHeaders(header))
+      } yield response.putHeaders(header)
+    }
+  }
+
+  def httpApp[F[_]: Sync](httpApp: HttpApp[F]): HttpApp[F] =
```
rossabaker, line 45: @voidocontext was moving toward having an `httpApp` and `httpRoutes` object, and then providing overloads off of those.  Maybe that's a good idea here, particularly if we do the second constructor.

server/src/main/scala/org/http4s/server/middleware/RequestId.scala
Slakah: Okidoke, I'll mess around with the method types and see if I can get something nice working.

server/src/main/scala/org/http4s/server/middleware/RequestId.scala
Slakah: Don't suppose you have an example? Tried doing some snooping on `@ voidocontext`, but couldn't find anything.

Made the change https://github.com/http4s/http4s/pull/3487/commits/611c898f5b7024ce0ad1c881d6f893a7b2bc4970 , this "feels" more ergonomic, but might be difficult to maintain in future.

Really appreciate the help so far! However I might need a bit more instruction (maybe method signatures) to get this PR over the line.

Slakah:
Hi @ChristopherDavenport , re:
> Attribute Key and append the UUID to both Request and Response via the attribute Key

Are you referring to being able to override the `X-Request-ID` header name?

server/src/main/scala/org/http4s/server/middleware/RequestId.scala
rossabaker: He discussed it on #2402.  I don't think there are any examples yet, but I understood him to mean something like:

```scala
object httpApp {
   def apply(headerName: CIString)(httpApp: HttpApp[F]): HttpApp[F] = ???
   def apply[F[_]: Sync](
      headerName: CIString = requestIdHeader,
      genReqId: F[UUID]
  )(httpApp: HttpApp[F]): HttpApp[F] = ???
}
```

Since you're overloading `apply`, it doesn't really save much.  But if you were giving everything its own name, that technique would give you a nice namespace for the parallel methods to all be called the same thing.

rossabaker:
> Are you referring to being able to override the X-Request-ID header name?

I think he's referring to the vault on the request.  Here's an example from a client middleware:

```scala
  private val redirectUrisKey = Key.newKey[IO, List[Uri]].unsafeRunSync

  /**
    * Get the redirection URIs for a `response`.
    * Excludes the initial request URI
    */
  def getRedirectUris[F[_]](response: Response[F]): List[Uri] =
    response.attributes.lookup(redirectUrisKey).getOrElse(Nil)
```

Both the request and response have attributes, which is a `Vault`.

I'm not sure if it's that useful on the request, since it can be fetched as a header.  Although another version of this middleware that passes them along as attributes instead of adding synthetic headers could be appealing.

I don't mean to speak for Chris, but I happened to be awake.  I'll let him set us both right. :laughing:

server/src/main/scala/org/http4s/server/middleware/RequestId.scala
Slakah: Happy to apply the same pattern that we're going to use for all the other middleware.

Slakah:
Made the following changes:
* Used the object-apply pattern (from #2402) when building the `RequestId` middleware
* Added some quick docs around how to use the middleware

Not added anything to the request/response attribute, as I don't think I would use that feature, so it makes it difficult to design/implement.

ChristopherDavenport:
I feel very strongly about applying it as an attribute to all requests and responses.

1. Its immutable data so can be added at almost zero cost.
2. Adding the attribute allows you to see into the data even if it was created via the Random
3. It lets you have access within your HttpApp. Whereas otherwise this is only put into the response and is invisible to the inside kleisli.

Slakah:
> I feel very strongly about applying it as an attribute to all requests and responses.

Okidoke, added an implementation + test. Please let me know if you folks want anything else changing.

rossabaker:
I think this is another one we can backport and release in 0.21.x.

rossabaker:
Added to the 0.21.x series. Thanks!

server/src/main/scala/org/http4s/server/middleware/RequestId.scala
voidcontext: Apologies for the delay, I got swamped with other things, but indeed this is what I had in mind when I suggested this approach.

PR #3492, opened 2020-06-11
Seperate Request only Dsl

ChristopherDavenport:
I think theres some java incompatibilities here, but I'm not sure why scala would have incompatibilites with this.

Regardless I think its great. 

This allows 

```scala
import org.http4s.dsl.request._

HttpRoutes.of[F]{
   case GET -> Root / "foo" / "bar" => 
      Response(Ok).withEntity("entity").pure[F]
}
```

Possible Problems. Importing both this and Http4sDsl will cause conflicts, same as if you import `cats.implicits._` and `cats.syntax.blah._` since they instantiate two different representations of the underlying vals.


rossabaker:
I think this is a good idea.  I think @SystemFw has argued for similar.

One neat thing about it is that it sidesteps the whole `F` parameterization.  No more fighting about extending the DSL vs. extending it to import locally from it.

:+1: when we figure out the MiMa worries.

ChristopherDavenport:
How do I get a rerun on this with the new master?

rossabaker:
You mean with the corrected MiMa checks?  Just merge or rebase series/0.21 into this branch.

rossabaker:
This can perhaps catch the 0.21.5 bus.  Still needs MiMa.

rossabaker:
Yeah, MiMa didn't like this. Want to retarget for 1.0 or see if it can be salvaged on 0.21?

rossabaker:
I made it binary compatible by removing the implementations in the base trait, and then providing the forwarders in each concrete trait.  There's a little bit of duplication that could be mopped up in 1.0, but now it's MiMa-approved.

dsl/src/main/scala/org/http4s/dsl/RequestDsl.scala
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.http4s.dsl
+
+import org.http4s.Method
+import org.http4s.dsl.impl._
+
+trait RequestDsl extends Methods with Statuses with Auth {
```
rossabaker, line 12: If this a request-only DSL, why does it include `Statuses`?

PR #3497, opened 2020-06-12
Add Functions to org.http4s.[client|server].middleware.{Logger, RequestLogger, ResponseLogger} for Logging Payloads Explicitly

kevinmeredith:
Adds functions to `org.http4s.client.middleware.{Logger, RequestLogger, ResponseLogger}` and `org.http4s.client.middleware.{Logger, RequestLogger, ResponseLogger}` for specifying how to explicitly log the HTTP Request/Response payload.

My end goal is to be able to use these `Logger`'s in order to log a sanitized/redacted JSON payload on my server and client(s). So, in short, I'll be going from `Stream[F, Byte] => Stream[F, Json] => Stream[F, Json] => F[String]` or something like that where the second `Stream[F, Json]` indicates that it's sanitized.

Note that I'll be using https://github.com/circe/circe/blob/v0.14.0-M1/modules/extras/src/main/scala/io/circe/extras.scala#L7-L51 in order to sanitize the JSON.

Hi @hamnis and @rossabaker - can you please review? This one is in a follow-up to https://github.com/http4s/http4s/pull/3486. 

Thanks!

### DIFF
--- a/client/src/main/scala/org/http4s/client/middleware/RequestLogger.scala
+++ b/client/src/main/scala/org/http4s/client/middleware/RequestLogger.scala
@@ -32,8 +32,8 @@ object RequestLogger {
   private def impl[F[_]: Concurrent](
       logHeaders: Boolean,
       logBodyText: Either[Boolean, Stream[F, Byte] => Option[F[String]]],
-      redactHeadersWhen: CaseInsensitiveString => Boolean = Headers.SensitiveHeaders.contains,
-      logAction: Option[String => F[Unit]] = None
+      redactHeadersWhen: CaseInsensitiveString => Boolean,
+      logAction: Option[String => F[Unit]]
   )(client: Client[F]): Client[F] = {
     val log = logAction.getOrElse { (s: String) =>
       Sync[F].delay(logger.info(s))
--- a/client/src/main/scala/org/http4s/client/middleware/RequestLogger.scala
+++ b/client/src/main/scala/org/http4s/client/middleware/RequestLogger.scala
@@ -29,7 +29,7 @@ object RequestLogger {
   )(client: Client[F]): Client[F] =
     impl[F](logHeaders, Left(logBody), redactHeadersWhen, logAction)(client)
 
-  def impl[F[_]: Concurrent](
+  private def impl[F[_]: Concurrent](
       logHeaders: Boolean,
       logBodyText: Either[Boolean, Stream[F, Byte] => Option[F[String]]],
       redactHeadersWhen: CaseInsensitiveString => Boolean = Headers.SensitiveHeaders.contains,
--- a/client/src/main/scala/org/http4s/client/middleware/RequestLogger.scala
+++ b/client/src/main/scala/org/http4s/client/middleware/RequestLogger.scala
@@ -29,6 +29,14 @@ object RequestLogger {
   )(client: Client[F]): Client[F] =
     impl[F](logHeaders, Left(logBody), redactHeadersWhen, logAction)(client)
 
+  def logBodyText[F[_]: Concurrent](
+      logHeaders: Boolean,
+      logBody: Stream[F, Byte] => Option[F[String]],
+      redactHeadersWhen: CaseInsensitiveString => Boolean = Headers.SensitiveHeaders.contains,
+      logAction: Option[String => F[Unit]] = None
+  )(client: Client[F]): Client[F] =
+    impl[F](logHeaders, Right(logBody), redactHeadersWhen, logAction)(client)
+
   private def impl[F[_]: Concurrent](
       logHeaders: Boolean,
       logBodyText: Either[Boolean, Stream[F, Byte] => Option[F[String]]],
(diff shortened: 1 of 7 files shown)


kevinmeredith:
Hi @rossabaker and @hamnis - could you please take a look? I'll make sure to wait before pinging you again as I don't want to be bothersome. Thanks!

kevinmeredith:
Hi Team - any interest in this one, please? I think it’s a logical extension of https://github.com/http4s/http4s/pull/3486. Thanks!

client/src/main/scala/org/http4s/client/middleware/RequestLogger.scala
```diff
@@ -26,14 +26,44 @@ object RequestLogger {
       logBody: Boolean,
       redactHeadersWhen: CaseInsensitiveString => Boolean = Headers.SensitiveHeaders.contains,
       logAction: Option[String => F[Unit]] = None
+  )(client: Client[F]): Client[F] =
+    impl[F](logHeaders, Left(logBody), redactHeadersWhen, logAction)(client)
+
+  def logBodyText[F[_]: Concurrent](
```
rossabaker, line 32: Is this separate from `apply` just for binary compatibility?  I think otherwise `logBody` could have been defaulted to some value of the newer type, but I don't see a way to do that with the existing apply.

server/src/main/scala/org/http4s/server/middleware/Logger.scala
```diff
@@ -46,6 +62,14 @@ object Logger {
   )(httpApp: HttpApp[F]): HttpApp[F] =
     apply(logHeaders, logBody, FunctionK.id[F], redactHeadersWhen, logAction)(httpApp)
 
+  def httpAppLogBodyText[F[_]: Concurrent](
```
rossabaker, line 65: A lot of the middlewares with overloads are introducing an `httpApp` and `httpRoutes` object, and then putting the overloads in there.  What's currently `httpApp` would be the `apply` in that object, and then this would be `logBodyText` in that object.

client/src/main/scala/org/http4s/client/middleware/RequestLogger.scala
kevinmeredith: > Is this separate from apply just for binary compatibility?

Hi @rossabaker - yes, as well as your point in:

> I think otherwise logBody could have been defaulted to some value of the newer type, but I don't see a way to do that with the existing apply.

Also, I did not want to overload `apply` per the up-votes to https://github.com/http4s/http4s/issues/2483#issuecomment-480536534.

server/src/main/scala/org/http4s/server/middleware/Logger.scala
kevinmeredith: To make sure that I follow, @rossabaker, you're suggesting that I re-name `httpAppLogBodyText` and `httpRoutesLogBodyText` to `httpApp` and `httpRoutes`, respectively, as overloads?

rossabaker: More like

```scala
object httpApp {
  def apply(...)
  def logBodyText(...)
}
```

kevinmeredith: Thanks, Ross. That makes sense! I'll make those changes.

kevinmeredith: Please let me know, @rossabaker, if https://github.com/http4s/http4s/pull/3497#commits-pushed-a3c6f32 addresses your comment.

kevinmeredith: If I understand the build output properly, my change may not be binary compatible:

```
[error] server: Failed binary compatibility check against org.http4s:http4s-server_2.12:0.21.0! Found 36 potential problems
[error]  * method httpApp(Boolean,Boolean,scala.Function1,scala.Option,cats.data.Kleisli,cats.effect.Concurrent)cats.data.Kleisli in object org.http4s.server.middleware.ResponseLogger does not have a correspondent in current version
```

rossabaker: Ugh.  I didn't think of that.  MiMa is right.  I guess we've never added an overload since introducing this pattern.  For 0.21, let's just chop it off at 32af97d.  Sorry about that.

rossabaker:
:+1: on truncation to  32af97d.

kevinmeredith:
@rossabaker - per my recent pushed [reverts](https://github.com/http4s/http4s/pull/3497#commits-pushed-d4b83d7), it addresses https://github.com/http4s/http4s/pull/3497#issuecomment-647872572:

```
$git diff logBodyWithText-loggers 32af97d6174a099b1cde92f9b8c8f2194b11546e
$
```

kevinmeredith:
One of the builds shows a, I expect, flaky spec:

```
[error] (blaze-client / Test / test) sbt.TestsFailedException: Tests unsuccessful
[error] Total time: 144 s (02:24), completed Jun 23, 2020, 2:03:36 PM
##[error]Process completed with exit code 1.
```

I'll re-run the tests once they're all done.

kevinmeredith:
Per your earlier comment I believe this one is ready for your review, Ross. Thanks

kevinmeredith:
Thanks, Ross, for reviewing. I closed and re-opened to prompt a re-build.

kevinmeredith:
Can you please review, @hamnis ?

kevinmeredith:
Ross and Erlend - thank you for your reviews of this and my other PR!

Regarding `0.21.5`, do you have any rough date, please, of when you'll cut and release it?

rossabaker:
I milestoned a few tickets for review, but now that we're sitting on a couple bugfixes, I want it out this week whether those are done or not.

PR #3501, opened 2020-06-14
Fixes #3328 Order[ContentCoding] should depend on compareToIgnoreCase

### LINKED ISSUE #3328: Order[ContentCoding] failure
```
2020-04-13T03:25:16.4857959Z [info] order laws must hold for Order[ContentCoding]
2020-04-13T03:25:16.4858291Z [error] x order.antisymmetry (41 ms)
2020-04-13T03:25:16.4858573Z [error]  Falsified after 0 passed tests.
2020-04-13T03:25:16.4858823Z [error]  > Labels of failing property:Expected: true
2020-04-13T03:25:16.4859069Z [error]  Received: false> ARG_0: ContentCoding(t, QValue(0.0))
2020-04-13T03:25:16.4859312Z [error]  > ARG_1: ContentCoding(t, QValue(0.0))
2020-04-13T03:25:16.4859557Z [error]  > ARG_2: org.scalacheck.GenArities$$Lambda$20955/1421781070@14088a8f
2020-04-13T03:25:16.4859794Z [error]  The seed is 2kpw5tJ8tADijqPv2GG3pUWPhjzpJUnaypQufFSWHBB=
```

ashwinbhaskar:


core/src/main/scala/org/http4s/ContentCoding.scala
```diff
@@ -114,7 +114,11 @@ object ContentCoding {
   }
 
   implicit val http4sOrderForContentCoding: Order[ContentCoding] =
-    Order.by(c => (c.coding.toLowerCase, c.qValue))
+    Order.from {
+      case (a, b) if a.coding.compareToIgnoreCase(b.coding) == 0 => a.qValue.compare(b.qValue)
+      case (a, b) => a.coding.compareToIgnoreCase(b.coding)
+    }
```
rossabaker, line 120: I think this is right (I haven't rerun the test with the seed), but you could avoid repeating the comparison when codings aren't equal.  I think this is also right:

```suggestion
    val cmp = a.coding.compareToIgnoreCase(b.coding)
    if (cmp == 0) a.qValue.compare(b.qValue) else cmp
```

ashwinbhaskar: @rossabaker this does not work. The test fails with that seed value. Will change the PR to WIP and investigate.

rossabaker: Nuts. Maybe I misanalyzed it on the ticket.  If we can figure out the exact inputs to that failing case, and not just the `.toString` representation, it should be much easier to spot that error.

core/src/main/scala/org/http4s/ContentCoding.scala
```diff
@@ -114,7 +115,8 @@ object ContentCoding {
   }
 
   implicit val http4sOrderForContentCoding: Order[ContentCoding] =
-    Order.by(c => (c.coding.toLowerCase, c.qValue))
+    Order.by(c => (c.coding.toLowerCase(ju.Locale.ENGLISH), c.qValue))
```
rossabaker, line 118: Hmm. Maybe. `compareToIgnoreCase` is character-by-character, like `equalIgnoreCase`, so I'm having a hard time conceptualizing why that wouldn't work.  `toLowerCase` can change the lengths of the strings, and I fear is open to more corner cases.  And the default locale on my machine is `en_US` anyway.

If we're going to use `.toLowerCase`, I agree with this change, to not rely on the default locale.  It still doesn't feel right, but I don't have a counterexample.

ashwinbhaskar: @rossabaker checkout https://github.com/http4s/http4s/issues/3328#issuecomment-645103412

core/src/main/scala/org/http4s/ContentCoding.scala
rossabaker: I opened a PR after investigating deeper myself, but after that, I think this might still be correct.  We just don't have an example where it matters, because characters are restricted to token characters, none of which do weird things in case folding.

PR #3502, opened 2020-06-14
Assign an initial version to the scalafix subproject

satorg:
Allows the **scalafix** rules get published and then references from other projects.

**Note**: this also changes the **http4s** output version from "0.20.0-M5" to the final "0.20.0"
since there's no reason to reference a milestone version anymore.

This change is inspired by this conversation: #3372,
and ideally should be merged _before_ the former PR.

### DIFF
--- a/scalafix/build.sbt
+++ b/scalafix/build.sbt
@@ -1,7 +1,9 @@
 lazy val V = _root_.scalafix.sbt.BuildInfo
+lazy val outputVersion = "0.20.0"
 inThisBuild(
   List(
     organization := "org.http4s",
+    version := outputVersion,
     homepage := Some(url("https://github.com/http4s/http4s")),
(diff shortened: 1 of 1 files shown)


satorg:
_cc_ @rossabaker, @amarrella.

amarrella:
@satorg looks good to me 👍

rossabaker:
I'm working on publishing this.  Should we have a tag?  This is unfortunately the first thing we're publishing out-of-band from the rest of the repo.  It's going to be a Weird Tag.

satorg:
Yes agree – it looks a bit weird. I'd suggest to add tag anyway but maybe just add something quite sounding like `-scalafix-only` suffix to it? I.e. `v0.20.0-scalafix-only` or something...

PR #3504, opened 2020-06-16
Publishing http4s-scalafix-0.20.0

rossabaker:
Just enough extra *.sbt to backpublish the scalafix for 0.20.0, per #3502

If I understand our strategy correctly, we want to start publishing these with output version concurrent with http4s version, and should move scalacheck out of its isolated build and into a first-class module.

/cc @satorg

### DIFF
--- a/scalafix/build.sbt
+++ b/scalafix/build.sbt
@@ -4,6 +4,10 @@ inThisBuild(
   List(
     organization := "org.http4s",
     version := outputVersion,
+    isSnapshot := {
+      if (outputVersion == "0.20.0") false
+      else isSnapshot.value
+    },
     homepage := Some(url("https://github.com/http4s/http4s")),
     licenses := List("Apache-2.0" -> url("http://www.apache.org/licenses/LICENSE-2.0")),
     developers := List(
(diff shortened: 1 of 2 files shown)


scalafix/build.sbt
```diff
@@ -4,6 +4,10 @@ inThisBuild(
   List(
     organization := "org.http4s",
     version := outputVersion,
+    isSnapshot := {
+      if (outputVersion == "0.20.0") false
+      else isSnapshot.value
+    },
```
rossabaker, line 10: This can go away after we close the Sonatype repo.  I was fighting the dynver plugin that gives us `isSnapshot`.

scalafix/build.sbt
```diff
@@ -19,13 +23,20 @@ inThisBuild(
     scalacOptions ++= List(
       "-Yrangepos"
     ),
+    publishTo := {
+      val nexus = "https://oss.sonatype.org/"
+      if (isSnapshot.value)
+        Some("snapshots" at nexus + "content/repositories/snapshots")
+      else
+        Some("releases"  at nexus + "service/local/staging/deploy/maven2")
+    }
```
rossabaker, line 32: This can go away if we integrate into the main repo, with sbt-ci-release.

If we don't, we should replace it with sbt-ci-release.

scalafix/build.sbt
satorg: I guess it may be because dynver expects to see the tag `v0.20.0` at zero distance from HEAD...

rossabaker: I was also working on this before I committed.  I could just as well have hardcoded the publishTo and skipped the `isSnapshot` until we integrate this with the other plugins. :shrug:

rossabaker:
You are probably the scalafix expert around here by now. :laughing:

A release based on this is staged.  I'll probably look to merge this tomorrow, but give others a chance to request changes.

rossabaker:
Released as http4s-scalafix-0.20.0.

PR #3516, opened 2020-06-20
scalafmt-2.6.1 on 0.21.x

rossabaker:
The merge is not bad, and it'll be easier if we keep the branches on the same scalafmt.

### DIFF
--- a/blaze-core/src/main/scala/org/http4s/blazecore/util/CachingStaticWriter.scala
+++ b/blaze-core/src/main/scala/org/http4s/blazecore/util/CachingStaticWriter.scala
@@ -19,7 +19,9 @@ import scala.concurrent.{ExecutionContext, Future}
 private[http4s] class CachingStaticWriter[F[_]](
     out: TailStage[ByteBuffer],
     bufferSize: Int =
-      8 * 1024)(implicit protected val F: Effect[F], protected val ec: ExecutionContext)
+      8 * 1024)(implicit
+    protected val F: Effect[F],
+    protected val ec: ExecutionContext)
     extends Http1Writer[F] {
   @volatile
   private var _forceClose = false
--- a/server/src/main/scala/org/http4s/server/middleware/CSRF.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/CSRF.scala
@@ -84,7 +84,7 @@ final class CSRF[F[_], G[_]] private[middleware] (
       lift(joined + "-" + encodeHexString(out))
     }
 
-  /** Generate a new token * */
+  /** Generate a new token */
   def generateToken[M[_]](implicit F: Sync[M]): M[CSRFToken] =
     signToken[M](CSRF.genTokenString)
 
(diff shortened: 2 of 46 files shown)


blaze-core/src/main/scala/org/http4s/blazecore/util/CachingStaticWriter.scala
```diff
@@ -19,7 +19,9 @@ import scala.concurrent.{ExecutionContext, Future}
 private[http4s] class CachingStaticWriter[F[_]](
     out: TailStage[ByteBuffer],
     bufferSize: Int =
-      8 * 1024)(implicit protected val F: Effect[F], protected val ec: ExecutionContext)
+      8 * 1024)(implicit
```
satorg, line 22: This line break looks a bit weird to me. I wonder was it enforced by scalafmt or simply inherited?

server/src/main/scala/org/http4s/server/middleware/CSRF.scala
```diff
@@ -85,7 +84,7 @@ final class CSRF[F[_], G[_]] private[middleware] (
       lift(joined + "-" + encodeHexString(out))
     }
 
-  /** Generate a new token **/
+  /** Generate a new token * */
```
satorg, line 87: Looks like this gap (and a few similar below) between two 'star' symbols is due to improved scalafmt rules. Would it  make sense to remove these redundant stars?

blaze-core/src/main/scala/org/http4s/blazecore/util/CachingStaticWriter.scala
rossabaker: I think I like that it broke the second parameter list onto its own line to match the first.  I suspect this one looks weird because of the default argument, which I might prefer it didn't break.

We generally try to adhere to scalafmt defaults, with the exception of vertical alignment, which bloats diffs.  But if there's a setting that makes this look less weird without introducing a lot more noise, I'm open to seeing what it looks like.

server/src/main/scala/org/http4s/server/middleware/CSRF.scala
rossabaker: Yes, we could and probably should clean those up.  Stand back, I have emacs.

blaze-core/src/main/scala/org/http4s/blazecore/util/CachingStaticWriter.scala
satorg: well, there're some other examples below how scalafmt allows to keep parameter groups separated. I guess, for this sample it could look like:
```
  out: TailStage[ByteBuffer],
  bufferSize: Int = 8 * 1024
)(implicit
  protected val F: Effect[F],
  ...
```
Although not my favorite way of formatting, but looks pretty clear to me.

rossabaker: Yes, I like your rendering better than what scalafmt did.

PR #3522, opened 2020-06-23
Upgrade to cats-2.2 and cats-effect-2.2 RCs

rossabaker:
Let's do our part to test cats-2.2 before it goes final.  I'm ambivalent about merging this.


rossabaker:
Well, it worked. Let's discuss whether we want to base the milestone on a milestone.

rossabaker:
I think I'm mildly :-1: on publishing against this. We're not depending on any new features of it yet, but we proved that we are ready for it when it comes.  I think we should keep aggressively integrating it, but only cut over when we depend on its enhancements.

hamnis:
superseeded by #3677 and #3679

PR #3526, opened 2020-06-24
Integrate scalafix as a first-class module

### LINKED ISSUE #3523: Publish scalafix with the remainder of the project
We just did this manually for 0.20.0, but now we're trying to tie the output version to the http4s version.

rossabaker:
Scalafix is now integrated with the rest of the build.  That means:

* Code quality checks are enforced
* Tests are run
* A scalafix is published every time we publish, both snapshots and releases

Fixes #3523.


build.sbt
```diff
@@ -540,6 +544,84 @@ lazy val examplesWar = exampleProject("examples-war")
...
+lazy val scalafixInput = project
+  .in(file("scalafix/input"))
+  .settings(scalafixSettings)
+  .settings(
+    skip in publish := true,
+    libraryDependencies ++= List(
+      "http4s-blaze-client",
+      "http4s-blaze-server",
+      "http4s-dsl",
+    ).map("org.http4s" %% _ % "0.21.0"),
```
rossabaker, line 579: This is going to bite us the next time we add a crossbuild, but it's sufficient for today.

build.sbt
```diff
@@ -540,6 +544,84 @@ lazy val examplesWar = exampleProject("examples-war")
...
+      (scalafixInput / Compile / sourceDirectories).value,
+    scalafixTestkitInputClasspath :=
+      (scalafixInput / Compile / fullClasspath).value,
+  )
+  .dependsOn(scalafixRules)
+  .enablePlugins(ScalafixTestkitPlugin)
+  .enablePlugins(AutomateHeaderPlugin)
+  .disablePlugins(MimaPlugin)
+
+addCommandAlias("ci", ";clean ;test")
```
rossabaker, line 623: This one looks like cruft from the build I inlined.

scalafix/output/src/main/scala/fix/ClientFetchTests.scala
```diff
@@ -1,3 +1,8 @@
+// Copyright 2013-2020 http4s.org
+//
+// SPDX-License-Identifier: Apache-2.0
+
+
```
rossabaker, line 5: Two lines are necessary here, and I can't comment inline that they are, so I'll comment here.

scalafix/rules/src/main/scala/fix/ClientFetchPatches.scala
```diff
@@ -1,45 +1,46 @@
...
+
 package fix
 
 import scalafix.v1._
 
 import scala.meta._
 
 object ClientFetchPatches {
-  def apply(tree: Tree)(implicit doc: SemanticDocument): List[Patch] = {
+  def apply(tree: Tree)(implicit doc: SemanticDocument): List[Patch] =
```
rossabaker, line 14: All the changes here are just scalafmt.

scalafix/rules/src/main/scala/fix/ClientFetchPatches.scala
```diff
@@ -49,172 +50,151 @@ object ClientFetchPatches {
...
+              maybeCallee,
+              Some((EffectRequestValueTypeSymbol(effectTypeSym), entityType)),
+              req,
+              fun) =>
           (maybeCallee, asApplyTypesString(effectTypeSym, entityType), req, fun)
       }
   }
 
-  private def asCalleeString(maybeCalleeTerm: Option[Term]) = maybeCalleeTerm.fold("") { _ + "." }
+  private def asCalleeString(maybeCalleeTerm: Option[Term]) = maybeCalleeTerm.fold("")(_.toString + ".")
```
rossabaker, line 188: This is a non-formatting change.

scalafix/rules/src/main/scala/fix/package.scala
```diff
@@ -1,20 +1,28 @@
...
 
 import scala.meta._
 
 package object fix {
+
   /** Allows to apply multiple extractors/matchers to a single expression in a `case` block.
     */
   object & {
-    def unapply[A](a: A): Option[(A, A)] = Some(a, a)
+    def unapply[A](a: A): Option[(A, A)] = Some((a, a))
```
rossabaker, line 16: This is a non-formatting change.

rossabaker:
Why didn't CI run on this?

scalafix/input/src/main/scala/fix/ClientFetchTests.scala
```diff
@@ -19,31 +23,30 @@ object ClientFetchTests {
       this.fetch[A](req)(f)
 
     final def doFetchR2[B](req: Request[G], f: Response[G] => G[B])(implicit G: Sync[G]): G[B] =
-      fetch[B](req) { f }
+      fetch[B](req)(f)
```
satorg, line 26: oh, I guess this change is due to scalafmt rules. And it would be totally fine for production code.
But actually it will produce a different syntactic tree in scalafix so this particular case and some other below lose any sense.

We should be able to preserve these input and output files from scalafmt since they are intended to handle different ways of formatting and structuring that some other uses may have in their projects.

satorg: I think we aren't free to assume that all people who use http4s in their project have the same scalafix rules as well.

rossabaker: Oh, that's a very good point.  I'll turn off formatting and revert the inputs and outputs.  And the license headers are more pain than they're worth, here.

We can keep the usual checks in the rules and tests projects, I think.

scalafix/rules/src/main/scala/fix/ClientFetchPatches.scala
satorg: Yes, I totally fine with it here. But I would hesitate from applying scalafmt to input and output sources for scalafix. I described above why.

satorg: Looks much better after scalafmt, btw.

rossabaker: It does.  I reverted it to your last version, just to get something working.  We could tweak it, but like you say, it's the automation that's bad here.

rossabaker:
The docs job is failing, and I'm unsure how I rattled that one with this change.  mdoc doesn't run clean on 2.13 yet, but that's not the failure.

PR #3531, opened 2020-06-25
Release 1.0.0-M2

rossabaker:


rossabaker:
Philosophical question: if a signature is deprecated in a tag that fails to release, do we need to bump the versions in the annotations?

rossabaker:
I think @ChristopherDavenport found the real problem in #3533. I tested this one locally, and it probably still wasn't going to work.

But that required a code change too, so we've still burned -M1 unless we want to do a manual deploy.

ChristopherDavenport:
Milestones are cheap. No worries.

PR #3534, opened 2020-06-26
Add org.http4s.server#serviceErrorHandler

kevinmeredith:
When using the [org.http4s.server.middleware.ErrorAction](https://github.com/http4s/http4s/blob/v0.21.5/server/src/main/scala/org/http4s/server/middleware/ErrorAction.scala#L24-L45) middleware, I've observed duplicate log entries. Note that, within the `ErrorAction#log` middleware, I'm passing `{case (t, msg) => log(t, msg)` per https://github.com/http4s/http4s/blob/v0.21.5/server/src/main/scala/org/http4s/server/middleware/ErrorAction.scala#L26-L27. 

The duplicate log entry is unfavorable since it makes it hard to reason about the behavior of my http4s web service.

The duplicate occurs due to a log occurring both within the `ErrorAction` middleware, as well as [DefaultServiceErrorHandler](https://github.com/http4s/http4s/blob/v0.21.5/server/src/main/scala/org/http4s/server/package.scala#L142-L163).

As a result, to eliminate that duplicate log entry from the `DefaultServiceErrorHandler`, I'm opening this PR.

With this change in place, in my http4s web service, I will specify a new [serviceErrorHandler](https://github.com/http4s/http4s/blob/v0.21.5/blaze-server/src/main/scala/org/http4s/server/blaze/BlazeServerBuilder.scala#L97) that consist of `case (t, msg) => F.unit}`.

Note to reviewers - I think reviewing individual commits makes this one easier to review. In short, I added a new function, and then used it from an existing function for DRY purposes.

### DIFF
--- a/server/src/main/scala/org/http4s/server/package.scala
+++ b/server/src/main/scala/org/http4s/server/package.scala
@@ -140,27 +140,17 @@ package object server {
     inDefaultServiceErrorHandler[F, F]
 
   def inDefaultServiceErrorHandler[F[_], G[_]](implicit
-      F: Monad[F]): Request[G] => PartialFunction[Throwable, F[Response[G]]] =
-    req => {
-      case mf: MessageFailure =>
-        messageFailureLogger.debug(mf)(
-          s"""Message failure handling request: ${req.method} ${req.pathInfo} from ${req.remoteAddr
-            .getOrElse("<unknown>")}""")
-        mf.toHttpResponse[G](req.httpVersion).pure[F]
-      case NonFatal(t) =>
-        serviceErrorLogger.error(t)(
-          s"""Error servicing request: ${req.method} ${req.pathInfo} from ${req.remoteAddr
-            .getOrElse("<unknown>")}""")
-        F.pure(
-          Response(
-            Status.InternalServerError,
-            req.httpVersion,
-            Headers(
-              Connection("close".ci) ::
-                `Content-Length`.zero ::
-                Nil
-            )))
+      F: Monad[F]): Request[G] => PartialFunction[Throwable, F[Response[G]]] = {
+    def messageFailureLogAction: (Throwable, => String) => F[Unit] = { (t, msg) =>
+      Monad[F].pure(messageFailureLogger.debug(t)(msg))
     }
+
+    def serviceErrorLogAction: (Throwable, => String) => F[Unit] = { (t, msg) =>
+      Monad[F].pure(serviceErrorLogger.error(t)(msg))
+    }
+
+    serviceErrorHandler[F, G](messageFailureLogAction, serviceErrorLogAction)
+  }
 
   def serviceErrorHandler[F[_], G[_]](
       messageFailureLogAction: (Throwable, => String) => F[Unit],
--- a/server/src/main/scala/org/http4s/server/package.scala
+++ b/server/src/main/scala/org/http4s/server/package.scala
@@ -161,4 +161,31 @@ package object server {
                 Nil
             )))
     }
+
+  def serviceErrorHandler[F[_], G[_]](
+      messageFailureLogAction: (Throwable, => String) => F[Unit],
+      serviceErrorLogAction: (Throwable, => String) => F[Unit]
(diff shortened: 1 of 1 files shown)


server/src/main/scala/org/http4s/server/package.scala
```diff
@@ -140,25 +140,42 @@ package object server {
...
+  }
+
+  def serviceErrorHandler[F[_], G[_]](
+      messageFailureLogAction: (Throwable, => String) => F[Unit],
+      serviceErrorLogAction: (Throwable, => String) => F[Unit]
+  )(implicit F: Monad[F]): Request[G] => PartialFunction[Throwable, F[Response[G]]] =
     req => {
       case mf: MessageFailure =>
-        messageFailureLogger.debug(mf)(
+        val msg: String =
```
satorg, line 161: it seems that `msg` is evaluated eagerly here even if the following call to `messageFailureLogAction` takes no action with it. The same applies  for the `msn` below.

satorg:
If I get it correctly, I guess you could achieve pretty much the same effects without any of these changes and with the default `serviceErrorHandler` just by muting the `org.http4s.server.message-failures` and `org.http4s.server.service-errors` loggers in your logging configuration. Couldn't you?

server/src/main/scala/org/http4s/server/package.scala
satorg: I mean, if you're going to call this new `serviceErrorHandler` function directly with just a stub callback, keep in mind that these `msg` vals are going to be evaluated anyway and then just discarded.

kevinmeredith:
Hey @satorg - good question! 

Is it possible to keep logging for org.http4s.server, but turn off the two you mentioned?

server/src/main/scala/org/http4s/server/package.scala
kevinmeredith: > Is it your intention for some reason?

No - thanks for finding my mistake. Perhaps def would work better for the messages then?

satorg:
@kevinmeredith, it depends on the logging framework you're using. If you're using "logback-classic" (which is a de-facto standard I believe), then it should be possible by adding these two lines into your "logback.xml":
```xml
<logger name="org.http4s.server.message-failures" level="OFF"/>
<logger name="org.http4s.server.service-errors" level="OFF"/>
```
(or you can try `level="FATAL"` if `"OFF"` doesn't work for you for some reason).

server/src/main/scala/org/http4s/server/package.scala
satorg: In terms of lazy evaluations – yes, it should. But in terms of overall performance it may work still worse than the original code, I guess. Because originally the code leveraged macros to optimize evaluation of these message strings. In the new code there's an additional layer of indirection caused by these "by-name" parameters. I'm not sure if it is a really big deal there, though.

kevinmeredith:
Thanks, @satorg. I'll try that out, namely turning those 2 off, while still logging:

```xml
<logger name="org.http4s.server.message-failures" level="OFF"/>
<logger name="org.http4s.server.service-errors" level="OFF"/>
<logger name="org.http4s.server" level="warn">
```

I'll report back.

kevinmeredith:
Thanks, @satorg . That fixed my problem!

That being said, is there any value then, @satorg @rossabaker , for this PR? 

It seems fully practical to me to disable to log, but from a purist/principle point of view I question whether it’s reasonable to directly handle the F[Unit] logging or no-op. Thoughts?

I’m not married to this PR so I’ll happily close it if you think there’s no value to it.

rossabaker:
The log4s macros check which logging levels are enabled.  They roughly expand to the classic Java logging pattern of:

```java
if (logger.isDebugEnabled())
  logger.debug("This is created strictly and might be " + expensive() + " so the guard takes care of it");
```

I think turning off those unwanted loggers is a reasonable and performant solution.

We have some split responsibilities between those middlewares and that ad hoc handler on the server backends.  I think from a coding perspective, it's better thinking about how we can simplify and streamline that for 1.0.

PR #3538, opened 2020-06-26
Fix Chunked Parser, Encoding Matcher Was Incorrect

ChristopherDavenport:
Previosuly even if the transfer-encoding was chunked it was not seen a chunked. As a result it defaulted to the `.getOrElse(0)` which created an empty body.

This rectifies this on both Request and Response parser and adds some more trace logging for the response because it had less to debug with.


ember-core/src/main/scala/org/http4s/ember/core/Parser.scala
```diff
@@ -127,7 +127,7 @@ private[ember] object Parser {
         isChunked =
           headers
             .get(org.http4s.headers.`Transfer-Encoding`)
-            .exists(_.value.toList.contains(TransferCoding.chunked))
+            .exists(_.values.exists(_ === TransferCoding.chunked))
```
rossabaker, line 130: Wait, what? Did that convert to a `String` and then to a `Seq[Char]`?

ember-core/src/main/scala/org/http4s/ember/core/Parser.scala
```diff
@@ -209,18 +209,26 @@ private[ember] object Parser {
...
         headers <- generateHeaders(headersBV)(Headers.empty)(logger)
+        _ <- logger.trace(show"Headers: $headers")
+
         (httpV, status) <- bvToResponseTopLine[F](methodHttpUri)
 
-        contentLength = headers.get(org.http4s.headers.`Content-Length`).map(_.length).getOrElse(0L)
+        _ <- logger.trace(s"HttpVersion: $httpV - Status: $status")
+
+        contentLength = headers.get(org.http4s.headers.`Content-Length`).map(_.length)
+        transferEncoding = headers.get(org.http4s.headers.`Transfer-Encoding`)
```
rossabaker, line 221: Nit: you shouldn't have both of these and can probably defer looking for it and save a few nanoseconds per request.

ember-core/src/main/scala/org/http4s/ember/core/Parser.scala
```diff
@@ -209,18 +209,26 @@ private[ember] object Parser {
...
         isChunked =
```
rossabaker, line 222: Actually, we've got an `isChunked` on `Message` already.  Any reason not to just use that?

ChristopherDavenport: We don't have a Message yet. This is before that point.

ember-core/src/main/scala/org/http4s/ember/core/Parser.scala
ChristopherDavenport: I think so I didn't get the typing working, but I have no idea how it worked.

ember-core/src/main/scala/org/http4s/ember/core/Parser.scala
ChristopherDavenport: I think I came up a way to satisfy all of these.

rossabaker: Oops. Yeah, I made a silly comment.

PR #3544, opened 2020-06-28
Fixes #3289 Fix bug when StaticFile.fromURL is called on a directory

### LINKED ISSUE #3289: StaticFile.fromURL throws IllegalStateException when called on a directory in resources
Here's the snipplet of the stacktrace

```
[scala-execution-context-global-28] ERROR o.h.s.b.Http1ServerStage$$anon$1 - Error writing body 
java.lang.IllegalStateException: Expected `Content-Length: 96` bytes, but only 11 were written.
	at org.http4s.blazecore.util.IdentityWriter.writeEnd(IdentityWriter.scala:70)
	at org.http4s.blazecore.util.EntityBodyWriter.$anonfun$writeEntityBody$1(EntityBodyWriter.scala:53)
	at cats.effect.internals.IORunLoop$.cats$effect$internals$IORunLoop$$loop(IORunLoop.scala:87)
	at cats.effect.internals.IORunLoop$RestartCallback.signal(IORunLoop.scala:359)
```

Looks like it's using `openConnection().getContentLength` for content length header, which actually calls `File::getSize` on the directory but the bytes come from `URL::openStream` which apparently returns newline separated file listing:

```
new String(u.openStream().readAllBytes())
$23 ==> ".git\n.gitignore\n.idea\nbuild.sbt\nindex.html\nproject\nREADME.md\nsrc\ntarget\n"
```

Not sure what expected behavior should be for directories, probably either have always return `OptionT.none` like it does for `StaticFile.fromFile`. Unfortunately it looks like there's no straightforward way to determine this.

Note: this doesn't happen when the resources are packaged in the jar, since `JarURLConnection` just returns 0 for file size and no bytes

ashwinbhaskar:


core/src/main/scala/org/http4s/StaticFile.scala
```diff
@@ -66,31 +66,37 @@ object StaticFile {
...
-        )
-      } else {
-        urlConn.getInputStream.close()
-        Response(NotModified)
-      }
-    })
+      cs: ContextShift[F]): OptionT[F, Response[F]] = {
+    val fileUrl = url.getFile()
+    val file = new File(fileUrl)
+    if (file.isDirectory())
```
rossabaker, line 72: This `isDirectory()` call is a side effect.  (I _think_ the two above it are not, as they don't consult the file system yet, but I may be wrong.)  What if we do the following?

* move this into the `F.delay`
* make the `F.delay` return an `Option`. This case returns `None`, and what's there returns `Some`.
* replace the `liftF` with `apply`.

tests/src/test/scala/org/http4s/StaticFileSpec.scala
```diff
@@ -276,5 +276,14 @@ class StaticFileSpec extends Http4sSpec with Http4sLegacyMatchersIO {
           .map(_.flatMap(_.contentLength))
       len must returnValue(Some(24005L))
     }
+
+    "Read a directory from a URL" in {
```
rossabaker, line 280: Maybe say "return none from a URL that is a directory".  I was confused what "reading a directory" meant until I reread the linked issue.

rossabaker:
Also, this is binary compatible, so we can put this on series/0.21 instead.

core/src/main/scala/org/http4s/StaticFile.scala
ashwinbhaskar: Resolved this. Verified that `val fileUrl = url.getFile()` and `val file = new File(fileUrl)` don't perform any side effects.

tests/src/test/scala/org/http4s/StaticFileSpec.scala
ashwinbhaskar: resolved

ashwinbhaskar:
@rossabaker I have made the changes requested by you and raised a new PR against series/21 - https://github.com/http4s/http4s/pull/3548 . Closing this PR.

Issue #3260, opened 2020-03-18
Quick start page is outdated


anilkumarmyla:
In https://http4s.org/v0.21/ this part about HelloWorldServer & HelloWorldService and the subsequent lines don't match the current code in http4s.g8 (this was part of 0.18 series)
```
$ find src/main -name '*.scala'
./src/main/scala/com/example/http4squickstart/HelloWorldServer.scala
./src/main/scala/com/example/http4squickstart/HelloWorldService.scala
```

rossabaker:
Agreed. Any volunteers to clean this up? It would be in docs/src/main/tut/index.md.

Guisanpea:
Taking a look to it

Guisanpea:
Shouldn't this be marked as solved?

rossabaker:
Yes! Sometimes we miss these when the PR doesn't have one of the magic GitHub ticket-closing words.

PR #3545, opened 2020-06-28
Update getting started

Guisanpea:
Related to https://github.com/http4s/http4s/issues/3260


rossabaker:
Also, the test failures are unrelated.  We might need to turn on the CI skip on a few more of these, as this is happening more lately.

Guisanpea:
So should then I change something or is this Okay? I do not have clear what is going on 😄

rossabaker:
It is good. :smile:

PR #3546, opened 2020-06-28
More Efficient Ember Encoding

ChristopherDavenport:
- It was previously UTF-8, we've switched to ASCII to match the [gist](https://gist.github.com/rossabaker/31ea688c2337c40a0a0a7070cc8b4e1d), is that the correct encoding?

### DIFF
--- a/ember-core/src/main/scala/org/http4s/ember/core/Encoder.scala
+++ b/ember-core/src/main/scala/org/http4s/ember/core/Encoder.scala
@@ -9,42 +9,77 @@ package org.http4s.ember.core
 import cats.effect._
 import fs2._
 import org.http4s._
-import cats.implicits._
-import Shared._
+// import cats.implicits._
+// import Shared._
+import java.nio.charset.StandardCharsets
 
 private[ember] object Encoder {
+
+  private val SPACE = " "
+  private val CRLF = "\r\n"
   def respToBytes[F[_]: Sync](resp: Response[F]): Stream[F, Byte] = {
-    val headerStrings: List[String] =
-      resp.headers.toList.map(h => h.name.show + ": " + h.value).toList
+    val initSection = {
+      val stringBuilder = new StringBuilder()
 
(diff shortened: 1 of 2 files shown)


ember-core/src/main/scala/org/http4s/ember/core/Encoder.scala
```diff
@@ -9,42 +9,75 @@ package org.http4s.ember.core
...
+      // Apply each header followed by a CRLF
+      resp.headers.foreach { h =>
+        stringBuilder
+          .append(h.renderString)
+          .append(CRLF)
+        ()
+      }
+      // Final CRLF terminates headers and signals body to follow.
+      stringBuilder.append(CRLF)
+      stringBuilder.toString.getBytes(StandardCharsets.US_ASCII)
```
rossabaker, line 38: Note that this (and line 76) can potentially throw, if there was something naughty in the model.  Our model constructors guard against a lot of these, but they're not foolproof yet.

rossabaker: On second thought, there is an old tradition of smuggling ISO-8859-1 through this, and it achieves that 1:1 mapping between bytes and chars as ASCII. Perhaps that is the better choice.

rossabaker: The [RFC says](https://tools.ietf.org/html/rfc7230#section-3.2.4):

> Historically, HTTP has allowed field content with text in the
   ISO-8859-1 charset [ISO-8859-1], supporting other charsets only
   through use of [RFC2047] encoding.  In practice, most HTTP header
   field values use only a subset of the US-ASCII charset [USASCII].
   Newly defined header fields SHOULD limit their field values to
   US-ASCII octets.  A recipient SHOULD treat other octets in field
   content (obs-text) as opaque data.

rossabaker:
Are we keeping this off 0.21 because it changes undefined (by us or spec) behavior?  It looks binary compatible, and I wouldn't necessarily mind backporting it.

ChristopherDavenport:
Updated Charset

Issue #3452, opened 2020-05-22
GZip client middleware crashes when processing an HTTP HEAD response


nelusnegur:
When I use an HTTP client configured with the GZip client middleware to perform a HEAD request, and provided the HTTP server can return a compressed response, the GZip middleware crashes during the processing of the response with the following error:

```
fs2.compress$NonProgressiveDecompressionException: buffer size 32768 is too small; gunzip cannot make progress
  fs2.compress$.$anonfun$gunzip$3(compress.scala:239)
  fs2.compress$.$anonfun$gunzip$3$adapted(compress.scala:221)
  fs2.Pull$.$anonfun$flatMap$1(Pull.scala:51)
  fs2.internal.FreeC$$anon$1.cont(FreeC.scala:29)
  fs2.internal.FreeC$ViewL$$anon$9$$anon$10.cont(FreeC.scala:196)
  fs2.internal.FreeC$ViewL$.mk(FreeC.scala:185)
  fs2.internal.FreeC$ViewL$.apply(FreeC.scala:176)
  fs2.internal.FreeC.viewL(FreeC.scala:79)
  fs2.internal.Algebra$.go$1(Algebra.scala:178)
  fs2.internal.Algebra$.$anonfun$compileLoop$7(Algebra.scala:219)
  fs2.internal.Algebra$.$anonfun$compileLoop$1(Algebra.scala:197)
  monix.eval.internal.TaskRunLoop$.startFull(TaskRunLoop.scala:170)
  monix.eval.internal.TaskRestartCallback.syncOnSuccess(TaskRestartCallback.scala:101)
  monix.eval.internal.TaskRestartCallback$$anon$1.run(TaskRestartCallback.scala:118)
  ...
```
Here is small Scala ammonite script, which defines an HTTP client configured with GZip middleware, executes a HEAD request and crashes with the error mentioned above:

```scala
import $ivy.`org.http4s::http4s-blaze-client:0.21.4`
import $ivy.`io.monix::monix:3.2.1`
import $ivy.`co.fs2::fs2-core:2.3.0`

import monix.eval.Task
import monix.eval.Task._
import monix.execution.Scheduler
import monix.execution.Scheduler.Implicits.global
import org.http4s._
import org.http4s.client.blaze._
import org.http4s.client._
import org.http4s.client.dsl.io._
import org.http4s.client.blaze.BlazeClientBuilder
import org.http4s.client.middleware.GZip

import scala.concurrent.Await
import scala.concurrent.duration._

[...]

PR #3550, opened 2020-06-30
GZip client middleware should not decompres empty response body

nelusnegur:
This PR generlizes the solution to the ["GZip client middleware crashes when processing an HTTP HEAD response"](https://github.com/http4s/http4s/issues/3452) issue as was suggested from the [prevoius PR](https://github.com/http4s/http4s/pull/3476), by not decompresing a response body regardless of the corresponding request method.

The GZip client middleware inspects the response stream whether it's empty, by trying to peek the first byte from it without consuming it. If there is at least on byte in the stream, then the stream is considered non empty and the decompression process is applied; otherwise it just returns an empty stream. 
I only was able to signal whether the stream is empty via the error channel, because the `fs2.compression.gunzip` method before proceeding with the decompression [creates a `java.util.zip.Inflater` and encloses it using a `Stream.bracket`](https://github.com/functional-streams-for-scala/fs2/blob/80b90a6536ffdbe6e9087ae77844586d9eaa3bee/core/jvm/src/main/scala/fs2/compression.scala#L709). If I run the decompression either inside the Pull (line 63) or outside (line 66), in the process of flattening the inner stream, it seems like it gets terminated and the `Inflater` is disposed before the decompression completes and it throws a `java.lang.NullPointerException: Inflater has been closed` exception.

### DIFF
--- a/client/src/main/scala/org/http4s/client/middleware/GZip.scala
+++ b/client/src/main/scala/org/http4s/client/middleware/GZip.scala
@@ -9,8 +9,7 @@ package client
 package middleware
 
 import cats.effect.{Bracket, Sync}
-import fs2.{Pipe, Stream}
-import java.io.EOFException
+import fs2.{Pipe, Pull, Stream}
 import org.http4s.headers.{`Accept-Encoding`, `Content-Encoding`}
 
 /**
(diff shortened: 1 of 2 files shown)


nelusnegur:
One of the steps from the CI pipeline failed unrelated to the changes in this PR. It could not fetch some dependency.

rossabaker:
This looks binary compatible, but I didn't do the backport of #3476 right, so I'll PR a backport of this separately instead of messing up twice.

PR #3552, opened 2020-07-01
Add Client.liftKleisli

ChristopherDavenport:
- Getting the effectful backends into a form they can be consumed by natchez is not simple. So I figure we should make it easier.


kubukoz:
wow, I think I implemented this at least 3 times and somehow didn't think it would be useful to others... I'm glad to see it in the library!

PR #3554, opened 2020-07-01
Backport #3550: Empty bodies in client.GZip

rossabaker:
This is a backport of #3550 to the 0.21 line.  It restores the deprecated `fs2.compress` calls for binary compatibility, because `Bracket` is the only constraint we have.  When we merge this back to master, we should keep the version from #3550.

Also add a `NoStackTrace` to the internal exception.  It should never be thrown outside the private method, and it's not an error condition.  We're using an untyped exception for control flow here, which is uncomfortable, but it's a highly localized implementation detail with a test.

I think this elides the concern from my bad merge of #3476, but I wanted to get eyes on it.

/cc @nelusnegur

### DIFF
--- a/client/src/main/scala/org/http4s/client/middleware/GZip.scala
+++ b/client/src/main/scala/org/http4s/client/middleware/GZip.scala
@@ -10,8 +10,7 @@ package middleware
 
 import cats.effect.Bracket
 import com.github.ghik.silencer.silent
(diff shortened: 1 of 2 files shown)


nelusnegur:
Thanks! Looks good to me. I also tested it with the script from the issue description and looks good.

PR #3555, opened 2020-07-01
Fixes #3368 Use UTF-8 for basic authentication

### LINKED ISSUE #3368: Change basic authentication to use UTF-8
At the moment, `BasicCredentials` [use ISO 8859-1](https://github.com/http4s/http4s/blob/744a70252e7c7c1f0c8d6e0fab557f34ccd77c17/core/src/main/scala/org/http4s/Credentials.scala#L70) charset, which would break authentication if, say, a password contains a character not present in that charset.

[RFC 7617](https://tools.ietf.org/html/rfc7617) outlines support for UTF-8 in basic authentication scheme. The short of it is, that in an older RFC the charset was left unspecified by mistake, so handling of non-ASCII characters was non-standard. Now the server can choose to declare its support for UTF-8 via the `charset` auth-param, and the client can choose to use it. See [appendix B](https://tools.ietf.org/html/rfc7617#appendix-B) for ways to provide some backward compatibility.

ashwinbhaskar:


server/src/main/scala/org/http4s/server/middleware/authentication/BasicAuth.scala
```diff
@@ -46,7 +46,7 @@ object BasicAuth {
         case Some(authInfo) =>
           Right(AuthedRequest(authInfo, req))
         case None =>
-          Left(Challenge("Basic", realm, Map.empty))
+          Left(Challenge("Basic", realm, Map("charset" -> "UTF-8")))
```
rossabaker, line 49: This parameter map is going to appear a lot in servers that support this, so I'd put it in a val.

core/src/main/scala/org/http4s/Credentials.scala
```diff
@@ -65,9 +69,19 @@ final case class BasicCredentials(username: String, password: String) {
...
+    Try(decoder.decode(byteByffer).toString)
+  }
+  private def decode(bytes: Array[Byte], charset: JavaCharset): String = 
+    new String(bytes, charset)
+
   def apply(token: String): BasicCredentials = {
     val bytes = Base64.getDecoder.decode(token)
-    val userPass = new String(bytes, StandardCharsets.ISO_8859_1)
+    val userPass: String = decode(bytes, utf8CharsetDecoder)
+      .fold({ _ => decode(bytes, StandardCharsets.ISO_8859_1) }, identity)
```
rossabaker, line 84: This looks like a suggested behavior specified in Appendix B.  It would be nice to make this whole thing configurable, but I don't see a clear path to do that yet.

core/src/main/scala/org/http4s/Credentials.scala
```diff
@@ -65,9 +69,19 @@ final case class BasicCredentials(username: String, password: String) {
 }
 
 object BasicCredentials {
+  private val utf8Charset = JavaCharset.forName("UTF-8")
+  private val utf8CharsetDecoder = utf8Charset.newDecoder
+  private def decode(bytes: Array[Byte], decoder: CharsetDecoder): Try[String] = {
+    val byteByffer = ByteBuffer.wrap(bytes)
+    Try(decoder.decode(byteByffer).toString)
```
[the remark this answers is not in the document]
ashwinbhaskar: `new String(bytes, charset)` does not throw an exception when giving invalid bytes. From the docs

```
This method always replaces malformed-input and unmappable-character sequences with this charset's default replacement string. The CharsetDecoder class should be used when more control over the decoding process is required.
```
The `decode` method of `CharsetDecoder` throws a `MalformedInputException | UnmappableCharacterException` when there is malformed-input or when there is an unmappable character. 
The overloaded method 3 lines down is the fallback method which uses `new String(bytes, charset)` (which does not throw exception)

core/src/main/scala/org/http4s/Credentials.scala
ashwinbhaskar: Appendix B suggests that you fallback even when there is a user-pass mismatch
```
When processing as UTF-8 fails (due to a failure to decode as UTF-8 or a mismatch of user-id/password), a server might try a fallback...
```
But IMO -  if the decoding is successful, we don't need to fallback in case of user-pass mismatch. We only need to fall back in case of an exception when decoding with utf-8. WDYT?

core/src/main/scala/org/http4s/Credentials.scala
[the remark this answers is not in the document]
rossabaker: Oh. That sounds familiar, but I'd forgotten that.  Good catch.

core/src/main/scala/org/http4s/Credentials.scala
rossabaker: Yeah, I think falling back in a user-pass mismatch would be too aggressive a default.  Furthermore, we're just decoding.  Auth is the caller's problem.

What concerns me is that if an http4s server was interpreting them as Latin1, and a client doesn't respect the new challenge parameter, we're quietly changing behavior and don't have a good recourse.  That's why I'd like to make the challenge and this behavior configurable, even though I like your new defaults.  I just don't see a great place to thread that configuration through.

server/src/main/scala/org/http4s/server/middleware/authentication/BasicAuth.scala
ashwinbhaskar: fixed

core/src/main/scala/org/http4s/Credentials.scala
```diff
@@ -65,9 +69,19 @@ final case class BasicCredentials(username: String, password: String) {
 }
 
 object BasicCredentials {
+  private val utf8Charset = JavaCharset.forName("UTF-8")
```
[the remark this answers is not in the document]
ashwinbhaskar: fixed

core/src/main/scala/org/http4s/Credentials.scala
ashwinbhaskar: @rossabaker can we try passing an `options` or `charset` parameter to `apply` of `BasicAuth`? Since this `apply` is what returns `AuthMiddleWare`, it might be the point to start threading the configuration from.

ashwinbhaskar: @rossabaker I tried passing a charset to basic auth. But like you said I am not able to wire it up all the way. Will have to spend more time on that. Do you want that implementation in this PR? Or can we do that as a separate issue?

satorg:
My 2 cents on this... Do we really a fallback encoding here? I mean, is it even possible that some byte array fails to decode to a string with UTF-8, but succeed with ISO-8859-1?

satorg:
Just have checked this:
```scala
import java.nio.charset.StandardCharsets._
println(UTF_8 contains ISO_8859_1)
```
prints `true`, i.e. everything that can be encoded with `ISO_8859_1`, can be encoded with `UTF_8` as well (but result bytes may differ for some particular characters).

JavaDocs on `Charset.contains`:
```
<p> A charset <i>C</i> is said to <i>contain</i> a charset <i>D</i> if,
and only if, every character representable in <i>D</i> is also
representable in <i>C</i>.  If this relationship holds then it is
guaranteed that every string that can be encoded in <i>D</i> can also be
encoded in <i>C</i> without performing any replacements.

<p> That <i>C</i> contains <i>D</i> does not imply that each character
representable in <i>C</i> by a particular byte sequence is represented
in <i>D</i> by the same byte sequence, although sometimes this is the
case.
```

satorg:
Ah, ok, I see my mistake: although every char in `ISO_8859_1` can be represented in `UTF-8`, but not every byte, encoded with `ISO_8859_1` can be decoded with `UTF-8`. So the fallback approach may make sense.

core/src/main/scala/org/http4s/Credentials.scala
rossabaker: So sorry, I fell behind again and hadn't seen the updates here.  I think we can track that as a separate issue.  As it stands, this is a 1.0 change, but I think it's a good one.

PR #3562, opened 2020-07-05
[WIP] Model Access-Control-Allow-Methods 

amitrai48:
This PR aims to model `Access-Control-Allow-Methods` a header from #2011 

RFC for Access-Control-Allow-Methods can be found here - https://fetch.spec.whatwg.org/#http-access-control-allow-headers

Below are the assumptions, would request someone to please validate it

- [x] An empty `Access-Control-Allow-Methods` cannot exists

- [x] `Access-Control-Allow-Methods` cannot have an invalid method name, i.e `Access-Control-Allow-Methods: GET, xyz` is an invalid header value

- [ ] `Access-Control-Allow-Methods` can have a wildcard `*` at any place. i.e `Access-Control-Allow-Methods: GET, POST, *` is a valid header value.

Currently the PR does not solve for invalid header names.

### DIFF
--- a/core/src/main/scala/org/http4s/headers/Access-Control-Allow-Methods.scala
+++ b/core/src/main/scala/org/http4s/headers/Access-Control-Allow-Methods.scala
@@ -9,24 +9,24 @@ package headers
 
 import org.http4s.util.Writer
 import org.http4s.parser.HttpHeaderParser
+import cats.data.NonEmptyList
 
 object `Access-Control-Allow-Methods`
     extends HeaderKey.Internal[`Access-Control-Allow-Methods`]
     with HeaderKey.Singleton {
 
-  def apply(methods: Method*): `Access-Control-Allow-Methods` =
-    `Access-Control-Allow-Methods`(methods.toSet)
+  val `*` = `Access-Control-Allow-Methods`(Wildcard())
+
+  def apply(first: Method, rest: Method*): `Access-Control-Allow-Methods` =
+    `Access-Control-Allow-Methods`(Methods(NonEmptyList.of(first, rest: _*)))
 
   override def parse(s: String): ParseResult[`Access-Control-Allow-Methods`] =
     HttpHeaderParser.ACCESS_CONTROL_ALLOW_METHODS(s)
 }
 
-final case class `Access-Control-Allow-Methods`(methods: Set[Method]) extends Header.Parsed {
+final case class `Access-Control-Allow-Methods`(wildCardOrMethods: WildcardOrMethods)
+    extends Header.Parsed {
   override def key: `Access-Control-Allow-Methods`.type = `Access-Control-Allow-Methods`
 
-  override def value: String =
-    if (methods.isEmpty) "*"
-    else methods.mkString(",")
-
-  override def renderValue(writer: Writer): writer.type = writer.append(value)
+  override def renderValue(writer: Writer): writer.type = wildCardOrMethods.render(writer)
 }
--- a/core/src/main/scala/org/http4s/headers/Access-Control-Allow-Methods.scala
+++ b/core/src/main/scala/org/http4s/headers/Access-Control-Allow-Methods.scala
@@ -7,4 +7,26 @@
 package org.http4s
 package headers
 
-object `Access-Control-Allow-Methods` extends HeaderKey.Default
+import org.http4s.util.Writer
+import org.http4s.parser.HttpHeaderParser
+
+object `Access-Control-Allow-Methods`
+    extends HeaderKey.Internal[`Access-Control-Allow-Methods`]
+    with HeaderKey.Singleton {
+
+  def apply(methods: Method*): `Access-Control-Allow-Methods` =
+    `Access-Control-Allow-Methods`(methods.toSet)
+
+  override def parse(s: String): ParseResult[`Access-Control-Allow-Methods`] =
+    HttpHeaderParser.ACCESS_CONTROL_ALLOW_METHODS(s)
+}
+
+final case class `Access-Control-Allow-Methods`(methods: Set[Method]) extends Header.Parsed {
+  override def key: `Access-Control-Allow-Methods`.type = `Access-Control-Allow-Methods`
+
+  override def value: String =
+    if (methods.isEmpty) "*"
+    else methods.mkString(",")
+
+  override def renderValue(writer: Writer): writer.type = writer.append(value)
+}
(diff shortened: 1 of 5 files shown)


core/src/main/scala/org/http4s/headers/Access-Control-Allow-Methods.scala
```diff
@@ -7,4 +7,26 @@
...
+    with HeaderKey.Singleton {
+
+  def apply(methods: Method*): `Access-Control-Allow-Methods` =
+    `Access-Control-Allow-Methods`(methods.toSet)
+
+  override def parse(s: String): ParseResult[`Access-Control-Allow-Methods`] =
+    HttpHeaderParser.ACCESS_CONTROL_ALLOW_METHODS(s)
+}
+
+final case class `Access-Control-Allow-Methods`(methods: Set[Method]) extends Header.Parsed {
```
rossabaker, line 24: It looks like you're using `Set.empty` as the wildcard value.  I wonder about being more explicit in the type: `Either[Wildcard, NonEmptySet[Method]]`.  It's more verbose, but also more meaningful.  Or could even create a sum type other than `Either`, to avoid arbitrary `Left` and `Right`.

rossabaker: I'm also not sure how I feel about the `Set`.  Nothing says that they have to be unique (even if it's silly for them not to be), and we lose the ability to render them exactly the way we got them.  Maybe a `List` (or `NonEmptyList`, pending the above) instead of a `Set`?  I could go either way.

core/src/main/scala/org/http4s/parser/SimpleHeaders.scala
```diff
@@ -51,6 +51,21 @@ private[parser] trait SimpleHeaders {
...
+  def ACCESS_CONTROL_ALLOW_METHODS(value: String): ParseResult[`Access-Control-Allow-Methods`] =
+    new Http4sHeaderParser[`Access-Control-Allow-Methods`](value) {
+      def entry =
+        rule {
+          oneOrMore(Token).separatedBy(ListSep) ~ EOL ~> { (ts: Seq[String]) =>
+            val ms = ts.map(
+              Method
+                .fromString(_)
+                .toOption
+                .getOrElse(sys.error("Impossible. Please file a bug report.")))
```
rossabaker, line 63: We have a `bug` function in the internal package for this purpose.

satorg: It doesn't look like a bug actually. It can happen if someone pass invalid/unsupported method in "Access-Control-Allow-Methods" header. Cannot it?

satorg: If my assumption is correct, I'd suggest to use the `fail` or `failX` parboiled rule instead.

rossabaker: All tokens are valid methods, even if they're not IANA-registered methods with a constant, so I think it is a bug if it fails.

In `Method`, the parser is in the companion object, so we can call the private constructor.  We use this pattern on a few model classes, which means we don't have these cases, but that pattern doesn't compose well, and we're left with ugly cases like this.  I'd like to rethink how we do that, but that's probably a separate ticket.

satorg: Perhaps it makes sense to create a parser rule for Method which could produce a Method instance directly, something like `def METHOD: Rule1[Method]`...

rossabaker: That would be useful, but there's still a design issue regarding where:

1. If the rule is somewhere in the parsers package, we need to expose an unsafe constructor outside `Method`.  That can be `private[http4s]`, but we have to be careful not to call `new Method(...)` in our internals.
2. We can put that rule in a trait in a `Parser` on the `Method` companion.  I started moving in this direction, but that contributes to the spaghetti of traits we have today.
3. We could have the rule like @amitrai48 wrote it in the parsers package, and only have that awful `bug` call once.
4. It would be a tiny bit more efficient to expose a `Method.unsafeFromString` and not wrap and unwrap an `Either` every time, which probably isn't that bad since we're just throwing anyway.  It just feels like defeat.

Maybe a combination of 3 and 4 is best.

core/src/main/scala/org/http4s/headers/Access-Control-Allow-Methods.scala
amitrai48: @rossabaker thank you for the reviews and comment. I do agree going with an explicit sum type would be better and more readable. With that in mind I was hoping to create something like this

```scala
sealed trait WildCardOrMethods

final case class Wildcard() extends WildCardOrMethods
final case class Methods(methods: NonEmptyList[Method]) extends WildCardOrMethods
```

And the parser would then check if there is even one wild card in the list, if there is `Wilcard` is passed to `Access-Control-Allow-Methods` case class. If not, the parser passes the parsed methods to the case class.

Am I thinking in the right direction?

core/src/main/scala/org/http4s/parser/SimpleHeaders.scala
amitrai48: hi @rossabaker @satorg I am sorry I did not fully understand the above conversation. Ross mentioned that `All tokens are valid methods` does that mean `Access-Control-Allow-Methods: GET, xyz` is a valid header?

satorg: I guess yes even though `xyz` is not an IANA-registered method.

rossabaker: Yes, that's correct.  

I'm going to go deep into obscure spec lawyering here, and contradict one of my earlier claims.  Apologies for both.

I previously thought that `Access-Control-Allow-Methods: GET, *` is invalid, because it seems like wild cards are only intended as a lone value.  But on second thought, `*` is a token character, and thus is a technically valid method name!  It's sloppy that the spec imbues a valid value, however rare, with special semantics.  I don't see anything in the spec that tells us whether to treat that `*` as a wildcard or as a method.

The HTTP spec says that we should be able to join list headers with commas without changing the semantics.  That implies that:

```
Access-Control-Allow-Methods: GET
Access-Control-Allow-Methods: *
```

should be treated the same as:

```
Access-Control-Allow-Methods: GET, *
```

It would be highly weird that the wild card changes semantics when joined to another header.  I like to think of it as a semigroup.  I think nobody really intended `*` to be a method, and we should treat those as a wild card anywhere we see them.

In that case, instead of the sum type that we just introduced, maybe it's just a simple `NonEmptyList[Method]`.  And whoever processes this header can treat `Method("*")` as a wild card.

rossabaker:
The header model changed significantly in series/0.22, so this will need a retooling.

armanbilge:
Superseded by:
- https://github.com/http4s/http4s/pull/5376

PR #3567, opened 2020-07-07
Models Content-Language for - #2011

amitrai48:


amitrai48:
@hamnis Thank you for pointing that out. I have added the test for multi non-hyphenated language.

PR #3577, opened 2020-07-12
Models the Max-Forwards header

cquiroz:
Simple modeling of `Max-Forwards` as part of https://github.com/http4s/http4s/issues/2011

### DIFF
--- a/core/src/main/scala/org/http4s/headers/Max-Forwards.scala
+++ b/core/src/main/scala/org/http4s/headers/Max-Forwards.scala
@@ -7,4 +7,29 @@
 package org.http4s
 package headers
 
-object `Max-Forwards` extends HeaderKey.Default
+import org.http4s.parser.HttpHeaderParser
+import org.http4s.util.Writer
+
+// see https://tools.ietf.org/html/rfc7231#section-5.1.2
+sealed abstract case class `Max-Forwards`(count: Long) extends Header.Parsed {
+  override def key: `Max-Forwards`.type = `Max-Forwards`
+  override def value: String = count.toString
+  def renderValue(writer: Writer): writer.type = writer << value
+}
+
+object `Max-Forwards` extends HeaderKey.Internal[`Max-Forwards`] with HeaderKey.Singleton {
+  private class MaxForwardsImpl(length: Long) extends `Max-Forwards`(length)
+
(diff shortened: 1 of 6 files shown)


core/src/main/scala/org/http4s/headers/Max-Forwards.scala
```diff
@@ -7,4 +7,29 @@
 package org.http4s
 package headers
 
-object `Max-Forwards` extends HeaderKey.Default
+import org.http4s.parser.HttpHeaderParser
+import org.http4s.util.Writer
+
+// see https://tools.ietf.org/html/rfc7231#section-5.1.2
+sealed abstract case class `Max-Forwards`(count: Long) extends Header.Parsed {
```
rossabaker, line 14: Pedantically, there's no limit, and `Long.MaxValue + 1` would be a legal value.  It's also absurd.  I think `Long` is the pragmatic model here, even if it's strictly incorrect.

cquiroz: That's correct, i followed the same logic as for Content-length, in practice an Int probably would suffice

rossabaker: It's possible do do this (and `Content-Length`) the pedantic way. Uncompiled:

```scala
sealed abstract class `Max-Forwards` {
  def unsafeToLong: Long
  def toLong: Option[Long]
  def toBigInt: BigInt
}

object `Max-Forwards` {
  private final class AtLong(val unsafeToLong: Long) extends `Max-Forwards` {
     val toLong = Some(unsafeToLong)
     val toBigInt = unsafeToLong.toBigInt
  }

  private final class AtBigInt(val toBigInt: BigInt) extends `Max-Forwards` {
    def toLong = if (toBigInt > BigInt.valueOf(Long.MaxValue)) None else Some(unsafeToLong)
    def unsafeToLong = toBigInt.longValue
  }
}
```

What I don't like about it is most people will just want a `Long`, and we're shaming that method as `unsafe`.  We could just return the last 64 bits, which is what 99.999999999999% of messages want, and what the standard library does.

Or we could continue to just assume nobody does this in practice.  I'm just thinking out loud.

PR #3583, opened 2020-07-14
Fixes #3034 Allow configuring circe support parser

### LINKED ISSUE #3034: circe: Allow to configure parsing of the duplicate keys
The `allowDuplicateKeys` cannot be changed: https://github.com/http4s/http4s/blob/2fa25a74d31ecfc9a6e61c8076ca6ad7d02e8b4a/circe/src/main/scala/org/http4s/circe/CirceInstances.scala#L20

It would be better if it could be configured e.g. via CirceInstancesBuilder.

ashwinbhaskar:


circe/src/main/scala/org/http4s/circe/CirceInstances.scala
```diff
@@ -97,7 +101,7 @@ trait CirceInstances extends JawnInstances {
   implicit def jsonEncoder[F[_]]: EntityEncoder[F, Json] =
     jsonEncoderWithPrinter(defaultPrinter)
 
-  private def fromJsonToChunk(printer: Printer)(json: Json): Chunk[Byte] =
+  def fromJsonToChunk(printer: Printer)(json: Json): Chunk[Byte] =
```
rossabaker, line 104: What necessitated adding this to the public API?

ashwinbhaskar: @rossabaker sorry, it slipped through. Not intentional. Added back the private and pushed.

ashwinbhaskar: One more thing that slipped through was the first commit message containing a `WIP` in it. I modified the commit message to remove the `WIP`  and force pushed.

Issue #3565, opened 2020-07-06
StaticFile.fromURL() does not handle exceptions


dubinsky:
When there is no object at the `url` specified, `url.openStream` causes `StaticFile.fromUrl()` effect to throw `FileNotFoundException` instead of returning (the equivalent of) `None`: see
https://github.com/http4s/http4s/blob/master/core/src/main/scala/org/http4s/StaticFile.scala#L93

Is there anything I can do at the place of the call to work around this while waiting for the fix?
Thanks!

satorg:
As I can see from the code, the exception should be thrown not by the `fromUrl` result effect itself, but rather by the response body stream – i.e. when you try to run the response body. Is it correct?

dubinsky:
Indeed; sorry for my misuse of the terminology - I am new to this :)

One of the routes of the `HttpApp` that I configure in `BlazeServerBuilder.withHttpApp()` does 
`StaticFile.fromURL(url, blocker, Some(request)).getOrElseF(NotFound(s"Not found: $url"))`, and when the requested object does not exist, `FileNotFoundException` gets thrown - instead of `NotFound` being returned.

So yes, exception *is* thrown by the response body stream, but I think it *should not* be thrown at all...
Or did I misunderstand the contract of the `StaticFile.fromURL()`?
In which case, how should I turn this exception into `NotFound` in my code?

Thanks!

satorg:
The way how `StaticFile.fromURL` is implemented is that it only opens input stream to the URL resource when the response body stream gets executed. And despite its `OptionT` return value it never returns 'None' actually (except if the resource is a directory, which is quite a border case).

It would be possible to change `StaticFile.fromURL` implementation to first check if a resource exists and return `None` if it doesn't but it could break backward compatibility.

@rossabaker @ashwinbhaskar wdyt?

ashwinbhaskar:
@satorg I think just adding one more check to see if the file exists should suffice? 
```
if (file.isDirectory() || !file.exists)
        None
else
 ...
```

dubinsky:
@satorg  It's hard for me to imagine code that actually *relies* on exceptions being thrown when running a functional effect :)
I am playing with a local copy of the `fromURL()`, and can't figure out how to adjust the rest of the code accordingly, since `OptionT.apply(F.delay{...})` wrapped around most of it expects `Option`, not `F[Option]`...

@ashwinbhaskar local `file.exists` has no connection with the existence of the object specified by the remote URL;
[...]

PR #3584, opened 2020-07-14
Close urlConn in StaticFile.fromURL

LolHens:
The URLConnection in StaticFile.fromURL is never closed (when the resource is not expired).
This PR reuses the already opened URLConnection to read the response body which in effect causes fs2 to close the URLConnection once the body is fully read.
Or was the body intended to open a separate url connection? In that case the previous URLConnection should probably be closed before returning the response.

### DIFF
--- a/core/src/main/scala/org/http4s/StaticFile.scala
+++ b/core/src/main/scala/org/http4s/StaticFile.scala
@@ -90,7 +90,7 @@ object StaticFile {
           Some(
             Response(
               headers = headers,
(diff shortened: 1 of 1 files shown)


satorg:
@ashwinbhaskar, I guess it could be linked (indirectly though) to #3565.

LolHens:
I think it would make sense to wait with this and rebase it on top of #3565 once it gets merged.

hamnis:
marked this as draft pending rebase on #3594 to avoid merging prematurely

LolHens:
I will close this in favour of #3625

PR #3594, opened 2020-07-19
Fixes #3565 Handle FileNotFoundException in StaticFile.fromURL

### LINKED ISSUE #3565: StaticFile.fromURL() does not handle exceptions

ashwinbhaskar:


core/src/main/scala/org/http4s/StaticFile.scala
```diff
@@ -87,11 +88,18 @@ object StaticFile {
...
           val headers = Headers(lenHeader :: lastModHeader ::: contentType)
 
-          Some(
-            Response(
-              headers = headers,
-              body = readInputStream[F](F.delay(url.openStream), DefaultBufferSize, blocker)
-            ))
+          Try(url.openStream()).fold(
+            fa = {
+              case _: FileNotFoundException => None
```
satorg, line 93: I guess this may be unsafe – if it happens that `openStream` will throw some other exception, then we get `MatchError` here.

core/src/main/scala/org/http4s/StaticFile.scala
```diff
@@ -87,11 +88,18 @@ object StaticFile {
             else `Transfer-Encoding`(TransferCoding.chunked)
           val headers = Headers(lenHeader :: lastModHeader ::: contentType)
 
-          Some(
-            Response(
-              headers = headers,
-              body = readInputStream[F](F.delay(url.openStream), DefaultBufferSize, blocker)
-            ))
+          Try(url.openStream()).fold(
```
satorg, line 91: I wonder if we really need `Try` here since we already have `F: Sync` around and can suspend effects in it and do all necessary error handling inside?

dubinsky:
@ashwinbhaskar I tried out your changes in my local copy of the code, and it works great! 
I'll be tracking the changes to this pull request as it gets refined, and then switch to the official code when it gets released.
Thank you sop much!

core/src/main/scala/org/http4s/StaticFile.scala
ashwinbhaskar: @satorg I don't thing there is a way to have `utl.openStream` in `F.delay`. For example if I put `url.openStream` in `F.delay`

```
val r: F[Option[Stream[F,Byte]]] = F.delay(url.openStream)
            .redeem(recover = {
              case _: FileNotFoundException => None
            }, f = {inputStream => Some(readInputStream[F](F.pure(inputStream), DefaultBufferSize, blocker))})
```
Notice that the type of `r` is wrapped in a `F`. But we want a result of the type `Option[Response[_][F]]`.

satorg: @ashwinbhaskar it should be almost always possible :)
This implementation works for me (as well as against the test case in your PR):
```scala
  def fromURL[F[_]](url: URL, blocker: Blocker, req: Option[Request[F]] = None)(implicit
      F: Sync[F],
      cs: ContextShift[F]): OptionT[F, Response[F]] = {
    val fileUrl = url.getFile()
    val file = new File(fileUrl)
    OptionT.apply(F.suspend {
      if (file.isDirectory()) // TODO: this call should be suspended in the blocker
        F.pure(None)
      else {
        val urlConn = url.openConnection
        val lastmod = HttpDate.fromEpochSecond(urlConn.getLastModified / 1000).toOption
        val ifModifiedSince = req.flatMap(_.headers.get(`If-Modified-Since`))
        val expired = (ifModifiedSince, lastmod).mapN(_.date < _).getOrElse(true)

        if (expired) {
          val lastModHeader: List[Header] = lastmod.map(`Last-Modified`(_)).toList
          val contentType = nameToContentType(url.getPath).toList
          val len = urlConn.getContentLengthLong
          val lenHeader =
            if (len >= 0) `Content-Length`.unsafeFromLong(len)
            else `Transfer-Encoding`(TransferCoding.chunked)
          val headers = Headers(lenHeader :: lastModHeader ::: contentType)

          blocker
            .delay(url.openStream)
            .redeemWith[Option[Response[F]]](
              {
                case _: FileNotFoundException => F.pure(None)
                case other => F.raiseError(other)
              },
              stream =>
                F.pure(
                  Some(
                    Response[F](
                      headers = headers,
                      body = readInputStream[F](stream.pure[F], DefaultBufferSize, blocker))))
            )
        } else
          blocker.delay(urlConn.getInputStream.close()) >>
            F.pure(Some(Response(NotModified)))
      }
    })
  }
```
Feel free to borrow any part of this snippet if you'd like to.

tests/src/test/scala/org/http4s/StaticFileSpec.scala
```diff
@@ -285,5 +285,13 @@ class StaticFileSpec extends Http4sSpec with Http4sLegacyMatchersIO {
         .unsafeRunSync
       s must_== None
     }
+
+    "return none from a URL that points to a resource that does not exist" in {
+      val s = StaticFile
+        .fromURL[IO](new URL("https://github.com/http4s/http4s/fooz"), testBlocker)
+        .value
+        .unsafeRunSync
+      s must_== None
```
satorg, line 294: Not critical at all, but I'd suggest using `s must beNone` here.

core/src/main/scala/org/http4s/StaticFile.scala
ashwinbhaskar: ah..I see the `suspend` did the trick:)

tests/src/test/scala/org/http4s/StaticFileSpec.scala
ashwinbhaskar: done and also added one more test to check the case where an exception is raised.

core/src/main/scala/org/http4s/StaticFile.scala
ashwinbhaskar: Fixed

core/src/main/scala/org/http4s/StaticFile.scala
```diff
@@ -87,15 +87,24 @@ object StaticFile {
...
-          Some(
-            Response(
-              headers = headers,
-              body = readInputStream[F](F.delay(url.openStream), DefaultBufferSize, blocker)
-            ))
-        } else {
-          urlConn.getInputStream.close()
-          Some(Response(NotModified))
-        }
+          F.delay(url.openStream)
```
satorg, line 90: I think it makes sense to delay `url.openStream` (which is apparently a blocking operation) in the `blocker` as well as `urlConn.getInputStream.close` below, doesn't it?

tests/src/test/scala/org/http4s/StaticFileSpec.scala
```diff
@@ -283,7 +284,27 @@ class StaticFileSpec extends Http4sSpec with Http4sLegacyMatchersIO {
...
+
+    "return none from a URL that points to a resource that does not exist" in {
+      val s = StaticFile
+        .fromURL[IO](new URL("https://github.com/http4s/http4s/fooz"), testBlocker)
+        .value
+        .unsafeRunSync
+      s must beNone
+    }
+
+    "raise exception when url does not exist" in {
```
satorg, line 298: This case is a really good catch! The only thing I could mildly recommend is to try leveraging a `throwA` matcher that should make the following check as simple as
```
StaticFile
          .fromURL[IO](new URL("https://quuzgithubfoo.com/http4s/http4s/fooz"), testBlocker)
          .value
          .unsafeRunSync must throwA[UnknownHostException]
```
https://etorreborre.github.io/specs2/guide/SPECS2-4.10.0/org.specs2.guide.Matchers.html#out-of-the-box

core/src/main/scala/org/http4s/StaticFile.scala
```diff
@@ -87,15 +87,24 @@ object StaticFile {
...
+            .redeem(
+              recover = {
+                case _: FileNotFoundException => None
+                case other => throw other
+              },
+              f = { inputStream =>
```
satorg, line 96: Was struggling a bit to understand what is `f` here? Then realized that it is just the `redeem` function's parameter name :)

tests/src/test/scala/org/http4s/StaticFileSpec.scala
ashwinbhaskar: makes sense

core/src/main/scala/org/http4s/StaticFile.scala
ashwinbhaskar: true

satorg:
Fixes #3565 (just to let GitHub create a link to the original issue)

rossabaker:
Sorry, still catching up from my vacation.

My question is whether we want this on the 0.21 series?  It's a semantic change, but is anyone intentionally relying on a `FileNotFoundException` bubbling up as a 500?

satorg:
I doubt if we can know it for sure. But I guess even if someone relies on `FileNotFoundException`, they still need to handle a `None` return value as well. Therefore this change should not break someone's code completely.

rossabaker:
I think you're right.  I don't like making changes that don't cause type errors, but this is a much more sensible default. Let's cherry-pick it.

rossabaker:
Picked onto series/0.21 as c5903bd4942bfccf9288a0c32157472e62ff2c6b. Thanks!

PR #3595, opened 2020-07-20
Update dsl.md

sebastianvoss:
As `Year.of` can throw an `java.time.DateTimeException` it results in an `500` errors if an invalid value is provided. Wrapping it in `Either` should prevent this.


sebastianvoss:
If I'm not mistaken, `Doc Site` failed because `leftMap` did not exist in Scala 2.12. Let me know if I should back port this.

rossabaker:
Thanks, and apologies for the slow reply.

Is importing `cats.implicits._` enough to get the `leftMap`?  I think it comes from bifunctor syntax.

sebastianvoss:
@rossabaker You are right, the import is enough. I added it.

rossabaker:
Test failure was unrelated. Thanks!

PR #3598, opened 2020-07-22
Fixes Connection Reuse for Ember Client

### LINKED ISSUE #3592: Ember client error when polling an url
I am getting the below exception when using an **ember** client to constantly fetch data from AWS SQS using the HTTP API.

> org.http4s.ember.core.EmberException$ParseError: Incomplete Header received (sz = 0): Right()

The error is somewhat sporadic, since sometimes it happens almost immediately and sometimes it takes roughly a minute.
But, after a couple of tries, it had never lasted more than 2 minutes before dying with the error.
Whereas, if using any other client it keeps running without any problem _(expect with the **netty** client, more info below)_.

I am using [this library](https://github.com/d2a4u/sqs4s) for polling SQS.
As such, I am not sure if the problem is with the library or with **ember** _(@d2a4u you may want to take a look to this)_; but since it doesn't happen with almost all other clients, it feels it should be something with **ember**.
Also, it means I can not create a reproducible example which doesn't involve the library and SQS _(sorry!)_

Ammonite script to reproduce the problem.

```scala
// scala 2.13.3

import $repo.`https://dl.bintray.com/d2a4u/sqs4s`

import $ivy.`org.typelevel::cats-effect:2.1.4`
import $ivy.`co.fs2::fs2-core:2.4.2`
import $ivy.`co.fs2::fs2-io:2.4.2`
import $ivy.`org.http4s::http4s-core:0.21.6`
import $ivy.`org.http4s::http4s-async-http-client:0.21.6`
import $ivy.`org.http4s::http4s-blaze-client:0.21.6`
import $ivy.`org.http4s::http4s-ember-client:0.21.6`
import $ivy.`org.http4s::http4s-jdk-http-client:0.3.0`
import $ivy.`org.http4s::http4s-jetty-client:0.21.6`
import $ivy.`org.http4s::http4s-netty-client:0.1.0`
import $ivy.`org.http4s::http4s-okhttp-client:0.21.6`
import $ivy.`io.sqs4s::sqs4s-native:1.0.0`
import $ivy.`io.chrisdavenport::log4cats-slf4j:1.1.1`
import $ivy.`ch.qos.logback:logback-classic:1.2.3`

import cats.effect.{Blocker, IO, Resource}
import fs2.Stream
import fs2.io.tls.TLSContext
import io.chrisdavenport.log4cats.slf4j.Slf4jLogger
import org.http4s.client.JavaNetClientBuilder
import org.http4s.client.asynchttpclient.AsyncHttpClient
import org.http4s.client.blaze.BlazeClientBuilder
import org.http4s.client.jdkhttpclient.JdkHttpClient
import org.http4s.client.jetty.JettyClient
import org.http4s.client.okhttp.OkHttpBuilder
import org.http4s.ember.client.EmberClientBuilder
import org.http4s.netty.client.NettyClientBuilder
import org.http4s.syntax.literals._

import sqs4s.serialization.instances._
import sqs4s.api.{AwsAuth, SqsSettings}
import sqs4s.api.lo.ReceiveMessage

import org.eclipse.jetty.client.{HttpClient => JettyHttpClient}
import org.eclipse.jetty.util.ssl.SslContextFactory.{Client => JettyHttpClientSslContext}

object Main {
  implicit val cs = IO.contextShift(scala.concurrent.ExecutionContext.global)
  implicit val timer = IO.timer(scala.concurrent.ExecutionContext.global)

  // Clients ------------------------------------------------------------------
  val clientNet = for {
    blocker <- Blocker[IO]
    client <- JavaNetClientBuilder[IO](blocker).resource
  } yield client

  val clientJdk = Resource.liftF(JdkHttpClient.simple[IO])

  val clientAsync = AsyncHttpClient.resource[IO]()

  val clientOkHttp = for {
    blocker <- Blocker[IO]
    builder <- OkHttpBuilder.withDefaultClient[IO](blocker)
    client <- builder.resource
  } yield client

  val clientJetty = JettyClient.resource[IO](new JettyHttpClient(new JettyHttpClientSslContext()))

  val clientNetty = NettyClientBuilder[IO].withDefaultSSLContext.withIdleTimeout(3.seconds).resource

  val clientBlaze = BlazeClientBuilder[IO](scala.concurrent.ExecutionContext.global).resource

  val clientEmber = for {
    blocker <- Blocker[IO]
    logger <- Resource.liftF(Slf4jLogger.create[IO])
    tlsContext <- Resource.liftF(TLSContext.system[IO](blocker))
    client <- EmberClientBuilder.default[IO].withTLSContext(tlsContext).withBlocker(blocker).withLogger(logger).withTimeout(3.seconds).build
  } yield client
  // --------------------------------------------------------------------------

  // Here change the client to use.
  val events = Stream.resource(clientNetty).flatMap { client =>
    def logData(data: String): IO[Unit] = IO {
      println("-------------------- DATA --------------------")
      println(data)
      println("-------------------- DATA --------------------")
    }

    val config = SqsSettings(
      queue = uri"your/sqs/uri",
      auth = AwsAuth(
        accessKey = "YOUR_ACCESS_KEY",
        secretKey = "YOUR_SECRET_KEY",
        region = "YOUR_AWS_REGION"
      )
    )

    val request = client.expect[String](ReceiveMessage[IO, String]().mkRequest(config))

    Stream.repeatEval(request).evalMap(logData).metered(2.seconds)
  }

  val program = events.compile.drain

  def run(): Unit = {
    program.unsafeRunSync()
  }
}

Main.run()
```

----

Additional information:

+ **Scala** version: `2.13.3`
+ **cats-effect** version: `2.1.4`
+ **fs2** version: `2.4.2`
+ **http4s** version: `0.21.6`
+ **Java** version: OpenJDK GraalVM CE 20.1.0 (build 11.0.7+10-jvmci-20.1-b02)
+ OS: **WSL 2** - Ubuntu 20.04 LTS (4.19.104-microsoft-standard)

-----

Let me know if I can do something else to help debug this problem.

---

# Edit

Updated the snippet to use more clients.

### Netty

For some reason, the **netty** client fails immediately after doing the first request with a `403` _(Forbidden)_.
Apparently, because the request wasn't properly signed.

This is probably an error either with the library or with how **netty** does the request; since this doesn't happen with other clients _(not even **ember**)_. So, I am pretty sure it is not related, but I guess I lose nothing by mentioning it here too.

ChristopherDavenport:
Fixes #3592 

First Error Discovered: When we are actually able to reuse a connection it was draining the root stream of extra bytes. As we had listed to take 2060 bytes, and then with the same reference try to pull 2060 more bytes as the effect was executed in the use block. This hung waiting for new bytes. (This signals to me that the user has a hold of the root connection not only a handle of 2060 bytes worth of it as pulling twice tries to take that twice, unsure if we should do something or how to do something that separates that logic). This means as well that a partially consumed stream will trigger the finalizers for reuse but the underlying stream will have however many bytes are uncomsumed present. We could add something that measures how much is read and then read that on shutdown, but then that introduces a race condition between release and an asynchronous effect. I feel that this default is reasonable for now, but we may want to revisit it in the future.

When introducing the host logic, we created a way to write the Host header twice, as if it was present in the headers and the uri it was written twice. This is now fixed to only render a Host header from a Uri if it is not present in headers. This is how its done in blaze but feels pretty inefficient compared to my other methods here. (We were previously ok due to how headers ++ operated)

### DIFF
--- a/ember-client/src/main/scala/org/http4s/ember/client/EmberClientBuilder.scala
+++ b/ember-client/src/main/scala/org/http4s/ember/client/EmberClientBuilder.scala
@@ -14,7 +14,6 @@ import cats.implicits._
 import cats.effect._
 import scala.concurrent.duration._
 import org.http4s.headers.Connection
-import org.http4s.Response
 import org.http4s.client._
 import fs2.io.tcp.SocketGroup
 import fs2.io.tcp.SocketOptionMapping
@@ -139,6 +138,7 @@ final class EmberClientBuilder[F[_]: Concurrent: Timer: ContextShift] private (
                 timeout
               )(logger)
               .map(response =>
+              // TODO If Response Body has a take(1).compile.drain - would leave rest of bytes in root stream for next caller
                 response.copy(body = response.body.onFinalizeCaseWeak {
                   case ExitCase.Completed =>
                     val requestClose = request.headers.get(Connection).exists(_.hasClose)
@@ -151,12 +151,8 @@ final class EmberClientBuilder[F[_]: Concurrent: Timer: ContextShift] private (
                   case ExitCase.Canceled => Sync[F].unit
                   case ExitCase.Error(_) => Sync[F].unit
                 }))
-          response <- Resource.make[F, Response[F]](responseResource.pure[F])(resp =>
-            managed.canBeReused.get.flatMap {
-              case Reusable.Reuse => resp.body.compile.drain.attempt.void
-              case Reusable.DontReuse => Sync[F].unit
-            })
-        } yield response)
+        } yield responseResource
+      )
       new EmberClient[F](client, pool)
     }
 }
--- a/ember-client/src/main/scala/org/http4s/ember/client/EmberClientBuilder.scala
+++ b/ember-client/src/main/scala/org/http4s/ember/client/EmberClientBuilder.scala
@@ -138,7 +138,7 @@ final class EmberClientBuilder[F[_]: Concurrent: Timer: ContextShift] private (
                 timeout
               )(logger)
               .map(response =>
-              // TODO If Response Body has a take(1).compile.drain - would leave rest of bytes in root stream for next caller
+                // TODO If Response Body has a take(1).compile.drain - would leave rest of bytes in root stream for next caller
                 response.copy(body = response.body.onFinalizeCaseWeak {
                   case ExitCase.Completed =>
                     val requestClose = request.headers.get(Connection).exists(_.hasClose)
@@ -151,8 +151,7 @@ final class EmberClientBuilder[F[_]: Concurrent: Timer: ContextShift] private (
                   case ExitCase.Canceled => Sync[F].unit
                   case ExitCase.Error(_) => Sync[F].unit
                 }))
-        } yield responseResource
-      )
+        } yield responseResource)
       new EmberClient[F](client, pool)
     }
 }
--- a/ember-core/src/main/scala/org/http4s/ember/core/Encoder.scala
+++ b/ember-core/src/main/scala/org/http4s/ember/core/Encoder.scala
@@ -56,13 +56,14 @@ private[ember] object Encoder {
         .append(req.httpVersion.renderString)
         .append(CRLF)
 
-      // Host From Uri Becomes Header
-      req.uri.authority.foreach { auth =>
-        stringBuilder
-          .append("Host: ")
-          .append(auth.renderString)
-          .append(CRLF)
-      }
+      // Host From Uri Becomes Header if not already present in headers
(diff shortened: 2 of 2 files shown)


ChristopherDavenport:
@BalmungSan if you can confirm this fixes the issue for you that would be great. Mine now runs infinitely.

BalmungSan:
@ChristopherDavenport I can confirm that using a local publish of this branch everything works fine!
Once again, thanks for all the hard work!

_(@d2a4u you may want to know about this)_

d2a4u:
Amazing, thanks. I found this issue as well last night but was struggling to fix it by myself. Should have jumped on gitter sooner. Thanks.

PR #3601, opened 2020-07-25
Client Request/Response Modifications

ChristopherDavenport:
- Connection Keep-Alive added if not present to request
- Date add if not present to request
- Default User-Agent added, and added to request if not present along with builder options
- Added Tests For the above and for response cancelation. We have an intriguing position that streams are scoped as closing before the last element. I wonder what else this could cause elsewhere as well.

### DIFF
--- a/ember-client/src/main/scala/org/http4s/ember/client/EmberClientBuilder.scala
+++ b/ember-client/src/main/scala/org/http4s/ember/client/EmberClientBuilder.scala
@@ -35,6 +35,35 @@ final class EmberClientBuilder[F[_]: Concurrent: Timer: ContextShift] private (
     val userAgent: Option[`User-Agent`]
 ) { self =>
 
+  @deprecated("Preserved for binary compatibility", "0.21.7")
+  private[EmberClientBuilder] def this(
+      blockerOpt: Option[Blocker],
+      tlsContextOpt: Option[TLSContext],
+      sgOpt: Option[SocketGroup],
+      maxTotal: Int,
+      maxPerKey: RequestKey => Int,
+      idleTimeInPool: Duration,
+      logger: Logger[F],
+      chunkSize: Int,
+      maxResponseHeaderSize: Int,
+      timeout: Duration,
+      additionalSocketOptions: List[SocketOptionMapping[_]]
+  ) =
+    this(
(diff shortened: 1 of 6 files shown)


ember-client/src/main/scala/org/http4s/ember/client/EmberClientBuilder.scala
```diff
@@ -83,6 +85,11 @@ final class EmberClientBuilder[F[_]: Concurrent: Timer: ContextShift] private (
   def withAdditionalSocketOptions(additionalSocketOptions: List[SocketOptionMapping[_]]) =
     copy(additionalSocketOptions = additionalSocketOptions)
 
+  def withUserAgent(userAgent: `User-Agent`) =
+    copy(userAgent = userAgent.some)
+  def withoutUserAgent =
+    copy(userAgent = None)
```
rossabaker, line 91: This feature feels like a potential new middleware, though I think it's also good here, since a reasonable default depends on the backend.

ember-core/src/main/scala/org/http4s/ember/core/Parser.scala
```diff
@@ -27,7 +27,8 @@ private[ember] object Parser {
         case None =>
           Pull.raiseError[F](
             EmberException.ParseError(
-              s"Incomplete Header received (sz = ${buff.size}): ${buff.decodeUtf8}"))
+              s"""Incomplete Header received (sz = ${buff.size}): utf8: ${buff.decodeUtf8} - b64: "${buff.toBase64}"""")
```
rossabaker, line 30: This never gets into a response, correct?

The `decodeUtf8` could fail as well, and I don't see a reason to treat it as UTF-8.

ChristopherDavenport: The buff is a bytevector so if this fails its just a left. But if it is. It makes it much easier to debug.

rossabaker: Oh, right, that's an Either.  Yeah, that's fine.

I just wanted to verify that this `ParseError` message doesn't get rendered into a 400, as it can reflect bytes from a request.

PR #3603, opened 2020-07-25
Enable Ember Server Connection Reuse

ChristopherDavenport:
- Enables Server Connection Reuse For Ember
- Add Similar Processing as is done in Client, with implicit keep-alive and date.

Performance Locally

#### Blaze

```
hey -z 30s  http://localhost:8080/hello

Summary:
  Total:	30.0007 secs
  Slowest:	0.5015 secs
  Fastest:	0.0001 secs
  Average:	0.0015 secs
  Requests/sec:	75735.8589
  
  Total data:	68549378 bytes
  Size/request:	68 bytes

Response time histogram:
  0.000 [1]	|
  0.050 [999948]	|■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■
  0.100 [1]	|
  0.151 [0]	|
  0.201 [0]	|
  0.251 [0]	|
  0.301 [0]	|
  0.351 [0]	|
  0.401 [0]	|
  0.451 [0]	|
  0.502 [50]	|


Latency distribution:
  10% in 0.0003 secs
  25% in 0.0004 secs
  50% in 0.0005 secs
  75% in 0.0007 secs
  90% in 0.0011 secs
  95% in 0.0014 secs
  99% in 0.0039 secs

Details (average, fastest, slowest):
  DNS+dialup:	0.0000 secs, 0.0001 secs, 0.5015 secs
  DNS-lookup:	0.0000 secs, 0.0000 secs, 0.0030 secs
  req write:	0.0000 secs, 0.0000 secs, 0.0051 secs
  resp wait:	0.0014 secs, 0.0001 secs, 0.4993 secs
  resp read:	0.0001 secs, 0.0000 secs, 0.0089 secs

Status code distribution:
  [200]	1000000 responses
```

#### Ember

```
hey -z 30s  http://localhost:8080/hello

Summary:
  Total:	30.0409 secs
  Slowest:	0.0541 secs
  Fastest:	0.0001 secs
  Average:	0.0024 secs
  Requests/sec:	20693.2210
  
  Total data:	18754221 bytes
  Size/request:	30 bytes

Response time histogram:
  0.000 [1]	|
  0.006 [593700]	|■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■
  0.011 [73]	|
  0.016 [9]	|
  0.022 [6]	|
  0.027 [6]	|
  0.032 [8]	|
  0.038 [6]	|
  0.043 [1289]	|
[...]

PR #3605, opened 2020-07-28
Ember Parsing in N of prelude + header bytes

ChristopherDavenport:
Built on top of #3603 

This is optimized for the ideal and most probable case. Prelude and Headers are present in the first chunk of the request. It then get progressively worse from there as we concatenate arrays so we can continue in linear time. There may be a way to drop off what we have done so far, or two algorithms depending on state. Ideally then this operates in n iterations equal to exactly the number of bytes that form the prelude and headers for both requests and responses.

We generate special case information for chunked and content length, as profiling shows that going and looking for those headers afterwards becomes the most expensive operations, primarily because being rule following and checking all our parsers is slow. Presently Uri parsing is like 50% of the parsing algorithm cpu time used.

Each algorithm has throwable a variable. We exit the loop if we encounter throwables. There's no recovery on an error.

### DIFF
--- a/ember-server/src/main/scala/org/http4s/ember/server/internal/ServerHelpers.scala
+++ b/ember-server/src/main/scala/org/http4s/ember/server/internal/ServerHelpers.scala
@@ -82,8 +82,11 @@ private[server] object ServerHelpers {
 
     def runApp(socket: Socket[F], isReused: Boolean): F[(Request[F], Response[F])] =
       for {
+        id <- Sync[F].delay(java.util.UUID.randomUUID())
+        _ <- logger.info(s"Starting HttpApp Processing- isReused: $isReused - $id")
         req <- socketReadRequest(socket, requestHeaderReceiveTimeout, receiveBufferSize, isReused)
         resp <- httpApp.run(req).handleError(onError)
+        _ <- logger.info(s"Finished HttpApp Processing - $id")
       } yield (req, resp)
 
     def send(socket: Socket[F])(request: Option[Request[F]], resp: Response[F]): F[Unit] =
@@ -112,42 +115,39 @@ private[server] object ServerHelpers {
       } yield resp.putHeaders(connection, date)
     }
 
+    def withUpgradedSocket(socket: Socket[F]): Stream[F, Nothing] =
+      (Stream(false) ++ Stream(true).repeat)
+        .flatMap { isReused =>
+          Stream
+            .eval(runApp(socket, isReused).attempt)
+            .evalMap {
+              case Right((req, resp)) =>
+                postProcessResponse(req, resp).map(resp => (req, resp).asRight[Throwable])
+              case other => other.pure[F]
+            }
+            .evalTap {
+              case Right((request, response)) => send(socket)(Some(request), response)
+              case Left(err) => send(socket)(None, onError(err))
+            }
+        }
+        .takeWhile {
+          case Left(_) => false
+          case Right((req, resp)) =>
+            !(
+              req.headers.get(Connection).exists(_.hasClose) ||
+                resp.headers.get(Connection).exists(_.hasClose)
+            )
+        }
+        .drain
+
     Stream
       .eval(termSignal)
       .flatMap(terminationSignal =>
         sg.server[F](bindAddress, additionalSocketOptions = additionalSocketOptions)
           .map { connect =>
-            Stream.resource {
-              for {
-                socket <- connect.flatMap(upgradeSocket(_, tlsInfoOpt))
-                out <- Resource.liftF(
-                  (Stream(false) ++ Stream(true).repeat)
-                    .flatMap { isReused =>
-                      Stream
-                        .eval(
-                          runApp(socket, isReused).flatMap {
-                            case (req, resp) =>
-                              postProcessResponse(req, resp)
-                                .map(resp => (req, resp))
-                          }.attempt
-                        )
-                    }
-                    .evalTap {
-                      case Right((request, response)) => send(socket)(Some(request), response)
-                      case Left(err) => send(socket)(None, onError(err))
-                    }
-                    .takeWhile {
-                      case Left(_) => false
-                      case Right((req, resp)) =>
-                        req.headers.get(Connection).exists(_.hasClose) ||
-                          resp.headers.get(Connection).exists(_.hasClose)
-                    }
-                    .evalTap(_ => ContextShift[F].shift) // Yield After Each Req/Resp Pair
-                    .compile
-                    .drain // In the above Stream is how we reuse connections
-                )
-              } yield out
-            }
+            Stream
+              .resource(connect.flatMap(upgradeSocket(_, tlsInfoOpt)))
+              .flatMap(withUpgradedSocket(_))
           }
           .parJoin(maxConcurrency)
           .interruptWhen(terminationSignal)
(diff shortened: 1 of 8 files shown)


rossabaker:
Is this one ready for a review yet?  Our old friend scalafmt needs to be run.

ember-server/src/main/scala/org/http4s/ember/server/internal/ServerHelpers.scala
```diff
@@ -45,19 +56,22 @@ private[server] object ServerHelpers {
...
-        receiveBufferSize: Int
+        receiveBufferSize: Int,
+        isReused: Boolean
     ): F[Request[F]] = {
-      val (initial, readDuration) = requestHeaderReceiveTimeout match {
-        case fin: FiniteDuration => (true, fin)
+      val (initial, readDuration) = (requestHeaderReceiveTimeout, idleTimeout, isReused) match {
+        case (fin: FiniteDuration, idle: FiniteDuration, true) => (true, idle + fin)
+        case (fin: FiniteDuration, _, false) => (true, fin)
         case _ => (false, 0.millis)
```
rossabaker, line 65: There's a `Duration.Zero`. #EveryNanosecondIsSacred

tpolecat:
I'm hitting the following exception running on this branch. I'll try to make a reproduction.

```
org.http4s.ember.core.Parser$Request$ReqPrelude$ParsePreludeError: Parse Prelude Error Encountered - Partially Decoded: None None None
	at org.http4s.ember.core.Parser$Request$ReqPrelude$.$anonfun$parsePrelude$1(Parser.scala:216)
	at org.http4s.ember.core.Parser$Request$ReqPrelude$.$anonfun$parsePrelude$1$adapted(Parser.scala:189)
	at fs2.Pull$.$anonfun$flatMap$1(Pull.scala:64)
	at fs2.internal.FreeC$$anon$1.cont(Algebra.scala:33)
	at fs2.internal.FreeC$ViewL$$anon$9$$anon$10.<init>(Algebra.scala:222)
	at fs2.internal.FreeC$ViewL$$anon$9.cont(Algebra.scala:222)
	at fs2.internal.FreeC$ViewL$$anon$9$$anon$10.cont(Algebra.scala:224)
	at fs2.internal.FreeC$ViewL$.mk(Algebra.scala:213)
	at fs2.internal.FreeC$ViewL$.apply(Algebra.scala:202)
	at fs2.internal.FreeC.viewL(Algebra.scala:86)
	at fs2.internal.FreeC$.go$1(Algebra.scala:399)
	at fs2.internal.FreeC$.$anonfun$compile$7(Algebra.scala:438)
	at fs2.internal.FreeC$.$anonfun$compile$1(Algebra.scala:415)
	at cats.effect.internals.IORunLoop$.cats$effect$internals$IORunLoop$$loop(IORunLoop.scala:178)
	at cats.effect.internals.IORunLoop$RestartCallback.signal(IORunLoop.scala:414)
	at cats.effect.internals.IORunLoop$RestartCallback.apply(IORunLoop.scala:435)
	at cats.effect.internals.IORunLoop$RestartCallback.apply(IORunLoop.scala:373)
	at cats.effect.internals.IORunLoop$.cats$effect$internals$IORunLoop$$loop(IORunLoop.scala:174)
	at cats.effect.internals.IORunLoop$RestartCallback.signal(IORunLoop.scala:414)
	at cats.effect.internals.IORunLoop$RestartCallback.apply(IORunLoop.scala:435)
	at cats.effect.internals.IORunLoop$RestartCallback.apply(IORunLoop.scala:373)
	at cats.effect.internals.IORunLoop$.cats$effect$internals$IORunLoop$$loop(IORunLoop.scala:174)
	at cats.effect.internals.IORunLoop$RestartCallback.signal(IORunLoop.scala:414)
	at cats.effect.internals.IORunLoop$RestartCallback.apply(IORunLoop.scala:435)
	at cats.effect.internals.IORunLoop$RestartCallback.apply(IORunLoop.scala:373)
	at cats.effect.internals.IOShift$Tick.run(IOShift.scala:36)
	at cats.effect.internals.PoolUtils$$anon$2$$anon$3.run(PoolUtils.scala:52)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:834)
Caused by: java.lang.Throwable: Reached Ended of Stream Looking for Request Prelude
	at org.http4s.ember.core.Parser$Request$ReqPrelude$.$anonfun$parsePrelude$1(Parser.scala:213)
	... 29 more
```

tpolecat:
Actually I see this on every request from Safari (it just takes a while) because it seems that all responses have `Connection: keep-alive` by default and eventually it times out. Maybe this is unrelated to this PR.

hamnis:
@tpolecat are they also missing content-length and|or transfer-encoding: chunked ?

hamnis:
If its an empty 200 OK, try adding a `Content-Length: 0` header and see if that helps

tpolecat:
@hamnis 

```
HTTP/1.1 200 OK
Date: Wed, 19 Aug 2020 20:45:18 GMT
Connection: keep-alive
Content-Type: text/html
Content-Length: 687

<html>...
```

tpolecat:
Despite this issues above, the client works much better with this patch than without. In v21.7 the client will block until it fails with the above error, before returning the body content. With this patch the body content is returned immediately and client code continues as if all is well. The exception above happens on someone else's thread and doesn't seem to cause any ill effects. So I'm still :+1:

ChristopherDavenport:
@tpolecat  If you could retest I would be very grateful.

tpolecat:
I'm still seeing that error, but it may be due to an issue with when you're calling the `withOnError` handler. It seems to be getting called *after* the response it sent, and the alternative response that's computed is discarded. Here is an example. 

Invoking with `curl` works as expected.

```
$ curl -i localhost:8080/foo
HTTP/1.1 404 Not Found
Date: Thu, 20 Aug 2020 16:03:02 GMT
Connection: keep-alive
Content-Type: text/plain; charset=UTF-8
Content-Length: 0
$ _
```

but the log shows 

```
org.http4s.ember.core.Parser$Request$ReqPrelude$ParsePreludeError: Parse Prelude Error Encountered - Partially Decoded: None None None
	at org.http4s.ember.core.Parser$Request$ReqPrelude$.$anonfun$parsePrelude$1(Parser.scala:217)
	at org.http4s.ember.core.Parser$Request$ReqPrelude$.$anonfun$parsePrelude$1$adapted(Parser.scala:190)
	at fs2.Pull$.$anonfun$flatMap$1(Pull.scala:64)
	...
	at java.base/java.lang.Thread.run(Thread.java:834)
Caused by: java.lang.Throwable: Reached Ended of Stream Looking for Request Prelude
	at org.http4s.ember.core.Parser$Request$ReqPrelude$.$anonfun$parsePrelude$1(Parser.scala:214)
	... 29 more
```

Here is the code I'm using.

```scala
package test

import cats.effect._
import org.http4s.ember.server.EmberServerBuilder
import org.http4s.HttpApp
import java.io.StringWriter
import java.io.PrintWriter
import org.http4s.Status
import org.http4s.EntityEncoder

object EmberTest extends IOApp {

    EmberServerBuilder
      .default[IO]
      .withPort(8080)
      .withHttpApp(HttpApp.notFound)
      .withOnError { t =>
        val sw = new StringWriter
        t.printStackTrace()
        t.printStackTrace(new PrintWriter(sw))
        org.http4s.Response[IO](
          status = Status.InternalServerError,
          body   = EntityEncoder[IO, String].toEntity(sw.toString).body
        )
      }
      .build
      .use(_ => IO.never)
      .as(ExitCode.Success)

}
```

ChristopherDavenport:
Does this work 
```
org.http4s.Response[IO](
 status = Status.InternalServerError,
).withEntity(sw.toString)
```

PR #3607, opened 2020-07-29
Add connection RequestKey to logs

anilkumarmyla:
Connection requestKey is useful in logs if there are multiple downstreams


rossabaker:
Cherry-picked for next 0.21 release.

PR #3609, opened 2020-07-29
Models 'Forwarded' header

### LINKED ISSUE #3471: Support parsing the 'Forwarded' header: RFC 7239
There's a specification [1,2] for a header which does not have native parsing support in http4s. The `Forwarded` header can be included by proxies as a standards-track alternative to `X-Forwarded-For`, `X-Forwarded-Proto`, and `X-Forwarded-Host`. http4s should consider providing support for it.

[1] [Forwarded - HTTP | MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded)
[2] [RFC 7239 - Forwarded HTTP Extension](https://tools.ietf.org/html/rfc7239])

satorg:
Resolves #3471.

The `Forwarded` header model tries to be as type safe as possible and thus introduces a hierarchy of model classes for that.

In particular, `Element` enables the following syntax for building elements:
```scala
Element
  .fromFor(Node(...))
  .withBy(Node(...))
  .withHost(Host(...))
  .withProto(Proto(...))
```

### DIFF
--- a/core/src/main/scala/org/http4s/headers/Forwarded.scala
+++ b/core/src/main/scala/org/http4s/headers/Forwarded.scala
@@ -54,15 +54,12 @@ object Forwarded
         checkPortNum(num).toLeft(C(num))
 
       def unapply(port: Port): Option[Int] =
-        PartialFunction.condOpt(port) {
-          case C(num) => num
-        }
+        PartialFunction.condOpt(port) { case C(num) => num }
     }
 
     sealed trait Obfuscated extends Name with Port { _: Product =>
 
-      /**
-        * Obfuscated value must start with '_' (underscore) symbol.
+      /** Obfuscated value must start with '_' (underscore) symbol.
         */
       def value: String
     }
@@ -131,8 +128,7 @@ object Forwarded
     override def render(writer: Writer): writer.type = renderElement(writer, this)
   }
 
-  /**
-    * Enables the following construction syntax (while preserving type safety and consistency):
+  /** Enables the following construction syntax (while preserving type safety and consistency):
     * {{{
     *   Element
     *     .withBy(<by-node>)
--- a/core/src/main/scala/org/http4s/headers/Forwarded.scala
+++ b/core/src/main/scala/org/http4s/headers/Forwarded.scala
@@ -1,0 +1,203 @@
+/*
+ * Copyright 2013-2020 http4s.org
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.http4s.headers
+
+import java.net.{Inet4Address, Inet6Address}
+
+import cats.data.NonEmptyList
+import cats.syntax.either._
+import org.http4s._
+import org.http4s.util.{Renderable, Writer}
+
+object Forwarded
+    extends HeaderKey.Internal[Forwarded]
+    with HeaderKey.Recurring
+    with ForwardedRenderers
+    with parser.ForwardedModelParsing {
+
+  final case class Node(nodeName: Node.Name, nodePort: Option[Node.Port] = None)
+
+  protected[http4s] val NodeNameIpv4 = Node.Name.Ipv4
+
+  object Node {
+    def apply(nodeName: Name, nodePort: Port): Node = apply(nodeName, Some(nodePort))
+
+    sealed trait Name { _: Product => }
+
+    object Name {
+      case class Ipv4(address: Uri.Ipv4Address) extends Name
+      case class Ipv6(address: Uri.Ipv6Address) extends Name
+      case object Unknown extends Name
+
+      def apply(address: Uri.Ipv4Address): Name = Name.Ipv4(address)
+      def apply(address: Inet4Address): Name = apply(Uri.Ipv4Address.fromInet4Address(address))
+      def apply(a: Byte, b: Byte, c: Byte, d: Byte): Name = apply(Uri.Ipv4Address(a, b, c, d))
+
+      def apply(address: Uri.Ipv6Address): Name = Name.Ipv6(address)
+      def apply(address: Inet6Address): Name = apply(Uri.Ipv6Address.fromInet6Address(address))
+      def apply(a: Short, b: Short, c: Short, d: Short, e: Short, f: Short, g: Short, h: Short)
+          : Name = apply(Uri.Ipv6Address(a, b, c, d, e, f, g, h))
+    }
+
+    sealed trait Port { _: Product => }
+
+    object Port {
+      private[this] final case class C(value: Int) extends Port {
+        override def productPrefix: String = "Port"
+      }
+
+      def fromInt(num: Int): ParseResult[Port] =
+        checkPortNum(num).toLeft(C(num))
+
+      def unapply(port: Port): Option[Int] =
+        PartialFunction.condOpt(port) {
+          case C(num) => num
+        }
+    }
+
+    sealed trait Obfuscated extends Name with Port { _: Product =>
+
+      /**
+        * Obfuscated value must start with '_' (underscore) symbol.
+        */
+      def value: String
+    }
+    object Obfuscated {
+      private[this] final case class C(value: String) extends Obfuscated {
+        override def productPrefix: String = "Obfuscated"
+      }
+
+      def fromString(s: String): ParseResult[Obfuscated] =
+        new ModelNodeObfuscatedParser(s).parse
+
+      def unapply(o: Obfuscated): Option[String] = Some(o.value)
+
+      // Referenced by model parsers.
+      private[http4s] def apply(s: String): Obfuscated = C(s)
+    }
+
+    def fromString(s: String): ParseResult[Node] = new ModelNodeParser(s).parse
+  }
+
+  sealed trait Host { _: Product =>
+    def host: Uri.Host
+    def port: Option[Int]
+  }
+
+  object Host {
+    private[this] final case class C(host: Uri.Host, port: Option[Int]) extends Host {
+      override def productPrefix: String = "Host"
+    }
+
+    def apply(uriHost: Uri.Host): Host = C(uriHost, None)
+
+    def from(uriHost: Uri.Host, port: Int): ParseResult[Host] =
+      checkPortNum(port).toLeft(C(uriHost, Some(port)))
+
+    def from(uriHost: Uri.Host, port: Option[Int]): ParseResult[Host] =
+      port.fold(apply(uriHost).asRight[ParseFailure])(from(uriHost, _))
+
+    def fromUri(uri: Uri): ParseResult[Host] =
+      uri.host.toRight(Failures.missingHost(uri)).flatMap(from(_, uri.port))
+
+    def fromString(s: String): ParseResult[Host] = new ModelHostParser(s).parse
+
+    def unapply(host: Host): Option[(Uri.Host, Option[Int])] = Some((host.host, host.port))
+  }
+
+  type Proto = Uri.Scheme
+  val Proto: Uri.Scheme.type = Uri.Scheme
+
+  sealed trait Element extends Renderable { _: Product =>
+    def `by`: Option[Node]
+    def `for`: Option[Node]
+    def `host`: Option[Host]
+    def `proto`: Option[Proto]
+
+    def withBy(value: Node): Element
+    def withFor(value: Node): Element
+    def withHost(value: Host): Element
+    def withProto(value: Proto): Element
+
+    def withoutBy: Element
+    def withoutFor: Element
+    def withoutHost: Element
+    def withoutProto: Element
+
+    override def render(writer: Writer): writer.type = renderElement(writer, this)
+  }
+
+  /**
+    * Enables the following construction syntax (while preserving type safety and consistency):
+    * {{{
+    *   Element
+    *     .withBy(<by-node>)
+    *     .withFor(<for-node>)
+    *     .withHost(<host>)
+    *     .withProto(<schema>)`
+    * }}}
+    */
+  object Element {
+    // Since at least one of the fields must be set to `Some`,
+    // the `Element` trait implementation is hidden.
+    private[this] final case class C(
+        `by`: Option[Node] = None,
+        `for`: Option[Node] = None,
+        `host`: Option[Host] = None,
+        `proto`: Option[Proto] = None)
+        extends Element {
+
+      def withBy(value: Node): Element = copy(`by` = Some(value))
+      def withFor(value: Node): Element = copy(`for` = Some(value))
+      def withHost(value: Host): Element = copy(`host` = Some(value))
+      def withProto(value: Proto): Element = copy(`proto` = Some(value))
+
+      def withoutBy: Element = copy(`by` = None)
+      def withoutFor: Element = copy(`for` = None)
+      def withoutHost: Element = copy(`host` = None)
+      def withoutProto: Element = copy(`proto` = None)
+
+      override def productPrefix: String = "Element"
+    }
+
+    def withBy(value: Node): Element = C(`by` = Some(value))
+    def withFor(value: Node): Element = C(`for` = Some(value))
+    def withHost(value: Host): Element = C(`host` = Some(value))
+    def withProto(value: Proto): Element = C(`proto` = Some(value))
+
+    def unapply(elem: Element): Option[(Option[Node], Option[Node], Option[Host], Option[Proto])] =
+      Some((elem.`by`, elem.`for`, elem.`host`, elem.`proto`))
+  }
+
+  final val PortMin = 0
+  final val PortMax = 65535
+
+  private def checkPortNum(portNum: Int): Option[ParseFailure] =
+    if ((portNum >= PortMin) && (portNum <= PortMax))
+      None
+    else
+      Some(Failures.invalidPortNum(portNum))
+
+  private object Failures {
+    def invalidPortNum(num: Int) =
+      ParseFailure("invalid port number", s"port $num is not in range $PortMin..$PortMax")
+    def missingHost(uri: Uri) =
+      ParseFailure("missing host", s"no host defined in the URI '$uri'")
+  }
+
+  override def parse(s: String): ParseResult[Forwarded] = parser.HttpHeaderParser.FORWARDED(s)
+}
+
+final case class Forwarded(values: NonEmptyList[Forwarded.Element])
+    extends Header.RecurringRenderable {
+
+  override type Value = Forwarded.Element
+  override def key: Forwarded.type = Forwarded
+
+  def apply(firstElem: Forwarded.Element, otherElems: Forwarded.Element*): Forwarded =
+    Forwarded(NonEmptyList.of(firstElem, otherElems: _*))
+}
--- a/core/src/main/scala/org/http4s/headers/Forwarded.scala
+++ b/core/src/main/scala/org/http4s/headers/Forwarded.scala
@@ -57,43 +57,34 @@ object Forwarded
         PartialFunction.condOpt(port) { case C(num) => num }
     }
 
-    sealed trait Obfuscated extends Name with Port { _: Product =>
+    /** Opaque type for obfuscated identifiers.
+      *
+      * @param value obfuscated identifier with leading '_' (underscore) symbol.
+      *
+      * @see [[https://tools.ietf.org/html/rfc7239#section-6.3]]
+      */
+    sealed abstract case class Obfuscated private (value: String) extends Name with Port
 
-      /** Obfuscated value must start with '_' (underscore) symbol.
-        */
-      def value: String
-    }
     object Obfuscated {
-      private[this] final case class C(value: String) extends Obfuscated {
-        override def productPrefix: String = "Obfuscated"
-      }
-
       def fromString(s: String): ParseResult[Obfuscated] =
         new ModelNodeObfuscatedParser(s).parse
 
-      def unapply(o: Obfuscated): Option[String] = Some(o.value)
-
-      // Referenced by model parsers.
-      private[http4s] def apply(s: String): Obfuscated = C(s)
+      /** Unsafe constructor for internal use only. */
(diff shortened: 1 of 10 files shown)


hamnis:
Great stuff. =)

This should probably be targeted 0.21 ?

satorg:
I guess it could... Would you suggest to rebase it onto `series/0.21` ?

hamnis:
@satorg Yes

hamnis:
We should have some tests for this, I do believe there are some examples in the RFC

satorg:
Yes, for sure. I am working on the tests now, per se.

core/src/main/scala/org/http4s/headers/Forwarded.scala
```diff
@@ -0,0 +1,121 @@
...
+    extends HeaderKey.Internal[Forwarded]
+    with HeaderKey.Recurring
+    with ForwardedModelParsers {
+
+  final case class Node(nodeName: Node.Name, nodePort: Option[Node.Port] = None)
+
+  object Node {
+    def apply(nodeName: Name, nodePort: Port): Node = apply(nodeName, Some(nodePort))
+
+    sealed trait Name extends Any
```
rossabaker, line 26: I've never been good at predicting `AnyVal` boxing, but since it's polymorphic in `Node`, won't these always be boxed in practice?  I'm not sure the `Any` / `AnyRef` is buying anything here, but I could be wrong.

core/src/main/scala/org/http4s/headers/Forwarded.scala
```diff
@@ -0,0 +1,121 @@
...
+      final case class Num(value: Int) extends AnyVal with Port
+
+      def apply(num: Int): Port = Port.Num(num)
+      def unapply(port: Port): Option[Int] =
+        PartialFunction.condOpt(port) {
+          case Num(num) => num
+        }
+    }
+
+    final case class Obfuscated(value: String) extends AnyVal with Name with Port
```
rossabaker, line 55: Not every string is a valid `obfnode` or `obfport`.  This should probably have a private constructor and a `fromString` to prevent invalid states.

rossabaker, line 55: Reading the spec, I expected `obfnode` to be in `Node` and `obfport` to be in `Port`.  The grammar is the same for this type, but it seems weird to me for the type to do double duty.

core/src/main/scala/org/http4s/headers/Forwarded.scala
```diff
@@ -0,0 +1,121 @@
...
+      def apply(address: Uri.Ipv6Address): Name = Name.Ipv6(address)
+      def apply(address: Inet6Address): Name = apply(Uri.Ipv6Address.fromInet6Address(address))
+      def apply(a: Short, b: Short, c: Short, d: Short, e: Short, f: Short, g: Short, h: Short)
+          : Name = apply(Uri.Ipv6Address(a, b, c, d, e, f, g, h))
+    }
+
+    sealed trait Port extends Any
+
+    object Port {
+      final case class Num(value: Int) extends AnyVal with Port
```
rossabaker, line 46: If we're being real strict about the spec, only values 0 to 99999 are allowed.  (Sadly, not 0 to 65535, or we could wrap a `Short`!)

core/src/main/scala/org/http4s/headers/Forwarded.scala
```diff
@@ -0,0 +1,121 @@
...
+  sealed trait Element extends Any with Product {
+    def `by`: Option[Node]
+    def `for`: Option[Node]
+    def `host`: Option[Host]
+    def `proto`: Option[Proto]
+
+    def `by=`(value: Node): Element
+    def `for=`(value: Node): Element
+    def `host=`(value: Host): Element
+    def `proto=`(value: Proto): Element
```
rossabaker, line 79: We use `withFoo` instead of `foo=` elsewhere.  Maybe that's better here?

core/src/main/scala/org/http4s/headers/Forwarded.scala
```diff
@@ -0,0 +1,121 @@
...
+      override def productPrefix: String = "Element"
+    }
+
+    def `by=`(value: Node): Element = C(`by` = Some(value))
+    def `for=`(value: Node): Element = C(`for` = Some(value))
+    def `host=`(value: Host): Element = C(`host` = Some(value))
+    def `proto=`(value: Proto): Element = C(`proto` = Some(value))
+  }
+
+  type ElementNel = NonEmptyList[Element]
```
rossabaker, line 112: I've grown regretful over a lot of our type aliases over the years.  Is anything motivating this one?

core/src/main/scala/org/http4s/parser/ForwardedHeader.scala
```diff
@@ -0,0 +1,116 @@
...
+    override def charset: Charset = StandardCharsets.ISO_8859_1
+
+    override def entry: Rule1[Forwarded] =
+      rule {
+        oneOrMore(ELEMENT).separatedBy(ListSep) ~ EOL ~> { (elems: Seq[Forwarded.Element]) =>
+          Forwarded(elems.head, elems.tail: _*)
+        }
+      }
+
+    // TODO: remove this override once the original `ListSet` gets fixed.
```
rossabaker, line 34: I remember a conversation about this.  Are we tracking this somewhere?

core/src/main/scala/org/http4s/headers/Forwarded.scala
satorg: Agree. There's still plenty of stuff to do.

satorg: I agree that it is a controversial approach. I was just thinking – what would be advantages of making them separated?
I mean it would not only require to create two different (although identical) case classes here, but also two parser rules for creating these classes. No doubt we could share some of their implementation, but it would increase overall code complexity anyway (remember, there still should be `fromString` as well as rendering implemented for both of them). On the other side, I couldn't come up with some real disadvantages besides, yeah, some weirdness.

I'm just trying to come up with some compromise solution that would let us avoid code duplication, preserve type safety and keep code complexity on some maintainable level. WDYT?

core/src/main/scala/org/http4s/headers/Forwarded.scala
satorg: Well, I could make its ctor private and add something like `fromInt` method to `Port` if you think it might make sense.

hamnis: `fromInt(port: Int): Either[ParseError, Port]` might make sense

core/src/main/scala/org/http4s/headers/Forwarded.scala
satorg: Yeah, I anticipated you won't like it:)
Actually, I was just playing a bit with different approaches.
I agree – it's better to keep this DSL consistent with others parts.

core/src/main/scala/org/http4s/headers/Forwarded.scala
satorg: Just traces of my experiments with the model... I'll remove it.

core/src/main/scala/org/http4s/headers/Forwarded.scala
satorg: I added these at the very beginning of my work, just in case, but I agree – there's no real sense to keep them.

core/src/main/scala/org/http4s/parser/ForwardedHeader.scala
satorg: I guess nope.
The original question I raised is here: https://gitter.im/http4s/http4s-dev?at=5f056255dbf01050ab5e7a14.
I'll create a separate issue for that, if you don't mind.

satorg:
> I was surprised how complicated this is, until I read the spec. This is really nice work.

Yeah, I also didn't realize that until I dove into this task. Unfortunately, due to my personal circumstances, I just didn't have enough time to squeeze it into 0.21.7, I am sorry for that.

core/src/main/scala/org/http4s/headers/Forwarded.scala
rossabaker: This comment is about minutiae, so I don't feel too strongly.

`Obfuscated` isn't as untyped as `String`, but it still raises the question, "an obfuscated what?"  It's not a particularly meaningful type on its own.  Sharing makes the implementation more concise, but isn't semantically meaningful in an HTTP sense.

In the unlikely event that the next version of that RFC allows some other character in an obfuscated host, we've got room to change it.  If the two unrelated concepts share a type, we'd have to wait for the next breaking release to implement it.

I don't think either argument matters much in this corner case of a supplemental RFC.  I'm thinking more in general design.  And you're right that the repetition is annoying, so I'm content either way.

core/src/main/scala/org/http4s/headers/Forwarded.scala
rossabaker: It reminded me of Scala's `_=` operator syntax, but those are for side effects.  If your convention didn't require the backticks, I'd probably like it a lot! :smile:

core/src/main/scala/org/http4s/parser/ForwardedHeader.scala
rossabaker: That'd be great.  Sorry I let that one fall through the cracks.

rossabaker:
No apologies required at all.  Glad for the help, and we won't run out of patch numbers any time soon.

core/src/main/scala/org/http4s/headers/Forwarded.scala
satorg: @rossabaker I wonder how come the port range you mentioned is 0..99999? I looked through a bunch of specs but it seems that everywhere it is 0..65535, even for IPv6.

satorg: @hamnis if we would like to guarantee correctness of the port number in `Forwarded.Node.Port` instances, then perhaps it makes sense to do the same for `Forwarded.Host`. But I'm struggling to come up with some nice name for such constructor, because it takes two parameters – `Uri.Host` and `Int` simultaneously. Would something like
```
object Host {
  def parse(uriHost: Uri.Host, portNum: Int): ParseResult[Host]
}
```
be good enough or could you suggest some nicer naming?

hamnis: `parse` does not really sit well with me here.
what about just `from` 

maybe a constructor from `Uri` might make sense as well.

`def fromUri(uri: Uri): ParseResult[Host]`

core/src/main/scala/org/http4s/headers/Forwarded.scala
```diff
@@ -0,0 +1,153 @@
...
+    def withoutBy: Element
+    def withoutFor: Element
+    def withoutHost: Element
+    def withoutProto: Element
+  }
+
+  /**
+    * Enables the following construction syntax (while preserving type safety and consistency):
+    * {{{
+    *   Element.`by=`(<by-node>).`for=`(<for-node>).`host=`(<host>).`proto=`(<schema>)`
```
hamnis, line 110: this comment is not correct anymore

satorg: Good catch, thanks!

core/src/main/scala/org/http4s/headers/Forwarded.scala
```diff
@@ -0,0 +1,185 @@
...
+  }
+
+  type Proto = Uri.Scheme
+  val Proto: Uri.Scheme.type = Uri.Scheme
+
+  sealed trait Element extends Product {
+    def `by`: Option[Node]
+    def `for`: Option[Node]
+    def `host`: Option[Host]
+    def `proto`: Option[Proto]
```
satorg, line 119: Back-ticks are here only because of `for` – apparently we cannot use it as an identifier.
All other field names don't need back-ticks (apparently), but I found it would be ugly to back-tick `for` only. But it looks quite ugly anyway. So I'd appreciate any better naming suggestions here:)

core/src/main/scala/org/http4s/parser/ForwardedHeader.scala
satorg: Created the issue #3657 (for reference)

satorg:
Ok, it seems I'm getting close. I have implemented the model renderers and mostly need to add some tests for them.

satorg:
Looks like test failures are due to #3664...

jyoo980:
@satorg sorry for the hold up, I've written up my investigations into the flaky failures in #3664

satorg:
@jyoo980 no worries, I'm still working on my tests here anyway.

satorg:
I believe I'm pretty much done with this PR.
Hopefully, it reproduces the corresponding RFC as close as possible (with an exception of the controversial approach of sharing the `Obfuscated` type between `Node.Name` and `Node.Port`).
Anyway I'm open for any suggestions, feel free to point them out.

**cc** @rossabaker, @hamnis

tests/src/test/scala/org/http4s/headers/ForwardedArbitraryInstances.scala
```diff
@@ -0,0 +1,110 @@
...
+import java.nio.charset.StandardCharsets
+
+import cats.data.NonEmptyList
+import cats.syntax.all._
+import org.http4s.internal.bug
+import org.http4s.laws.discipline.ArbitraryInstances
+import org.http4s.{ParseResult, Uri}
+import org.scalacheck.{Arbitrary, Gen}
+
+private[http4s] trait ForwardedArbitraryInstances
```
satorg, line 18: I wonder, should this trait be moved to `org.http4s.laws.discipline` (where `ArbitraryInstances` sits)?

hamnis (approved):

hamnis:
A colleage of mine just hit this in production https://github.com/scala/bug/issues/8831 
Could this be a problem with the backticks for the parameters ?

satorg:
@hamnis, interesting issue, never seen it before. Thanks for pointing this out to me!

I'd guess it's unlikely to affect this code since the bug usually appears when there are some common prefixes among parameters.

But to be on the safer side, I agree, we can consider getting rid of backticks. As I wrote some time before, the backticks are only necessary for the `for` parameter, while the other parameters get backticked just to look consistently with `for`.

But it is not the only option – we could rename all the parameters somehow else. For example, since all the parameters are actually optional, they could be renamed to
```scala
sealed trait Element extends Renderable { _: Product =>
    def maybeBy: Option[Node]
    def maybeFor: Option[Node]
    def maybeHost: Option[Host]
    def maybeProto: Option[Proto]
...
}
```
..or something like that. WDYT?

hamnis:
I've never been a fan of the backticks of for instance header classes just to get the "correct" name, it makes them less discoverable by for instance IDEA. So I prefer the latter version here. Other maintainers may prefer differently. 

@rossabaker any guidelines here?

satorg:
Looks like the most recent scalafmt has broken this PR. I'll take care of it.

@rossabaker, btw, I'd appreciate your thoughts regarding the questions above. Feel free to make any suggestions please – I'm open to polish the PR as much as possible.

core/src/main/scala/org/http4s/parser/ForwardedHeader.scala
```diff
@@ -0,0 +1,117 @@
...
+    //       See also #3657.
+    override def ListSep: Rule0 = rule(OptWS ~ ',' ~ OptWS)
```
satorg, line 36: If by chance we get #3743 merged before this one, it may make sense to remove this override.

tests/src/test/scala/org/http4s/headers/ForwardedArbitraryInstances.scala
rossabaker: I think yes, but that doesn't need to block this.

core/src/main/scala/org/http4s/headers/Forwarded.scala
```diff
@@ -0,0 +1,178 @@
...
+      case object Unknown extends Name
+
+      def apply(address: Uri.Ipv4Address): Name = Name.Ipv4(address)
+      def apply(address: Inet4Address): Name = apply(Uri.Ipv4Address.fromInet4Address(address))
+      def apply(a: Byte, b: Byte, c: Byte, d: Byte): Name = apply(Uri.Ipv4Address(a, b, c, d))
+
+      def apply(address: Uri.Ipv6Address): Name = Name.Ipv6(address)
+      def apply(address: Inet6Address): Name = apply(Uri.Ipv6Address.fromInet6Address(address))
+      def apply(a: Short, b: Short, c: Short, d: Short, e: Short, f: Short, g: Short, h: Short)
+          : Name = apply(Uri.Ipv6Address(a, b, c, d, e, f, g, h))
```
rossabaker, line 43: A little late on this, but would we be better off with `from*` constructors instead of overriding `apply`?  I've soured a bit on overrides.  It's a fuzzy case: we usually `apply` if it's total, and `from*` if it's a `ParseResult`.  But we usually don't have so many options to make it total.

I could happily go either way here.

core/src/main/scala/org/http4s/headers/Forwarded.scala
```diff
@@ -0,0 +1,178 @@
...
+  }
+
+  sealed abstract case class Host private (host: Uri.Host, port: Option[Int])
+
+  object Host {
+    private[this] def apply(host: Uri.Host, port: Option[Int]) = new Host(host, port) {}
+
+    def apply(uriHost: Uri.Host): Host = apply(uriHost, None)
+
+    def from(uriHost: Uri.Host, port: Int): ParseResult[Host] =
```
rossabaker, line 86: Could avoid an override calling this `fromHost`.

core/src/main/scala/org/http4s/headers/Forwarded.scala
```diff
@@ -0,0 +1,178 @@
...
+      checkPortNum(port).toLeft(apply(uriHost, Some(port)))
+
+    def from(uriHost: Uri.Host, port: Option[Int]): ParseResult[Host] =
```
rossabaker, line 89: What's the use case for one that takes an `Option`?  This seems most useful if you have a `Uri.Authority`.

core/src/main/scala/org/http4s/parser/ForwardedHeader.scala
rossabaker: Well, it did. :smile:

rossabaker (commented):
Thanks for your patience while I dig out of my hole.  This is great work, and everything I mentioned is pretty insignificant.

core/src/main/scala/org/http4s/headers/Forwarded.scala
satorg: Just thoughts... 

First, if we prefer to reduce (or completely get rid of) overrides here, we can safely drop these two constructors:
```
def apply(address: Uri.Ipv4Address): Name
def apply(address: Uri.Ipv6Address): Name
```
Because these two can be replaced with direct calls to `Name.Ipv4` and `Name.Ipv6` – these  `apply`-s are just shortcuts and thus there's no need to add extra `from*` methods for them.

Second, I'd suggest the following names for the remaining constructors:
```
def fromInet4Address(address: Inet4Address) // reflects `Uri.Ipv4Address.fromInet4Address`
def fromIpv4Address(a: Byte, b: Byte, c: Byte, d: Byte) // reflects `Uri.Ipv4Address.apply`

def fromInet6Address(address: Inet6Address): Name // reflects `Uri.Ipv6Address.fromInet6Address`
def fromIpv6Address(a: Short, ..., h: Short): Name // reflects `Uri.Ipv6Address.apply`
```
If you don't mind, I'm up to rename them.

Third, we could consider using different names like `of*` for creating totals when we'd like to avoid `apply` overrides. Just to disambiguate them from `from*`-s, which usually return `ParseResult`. Not sure if it will look consistently across the project, though.

core/src/main/scala/org/http4s/headers/Forwarded.scala
satorg: Well, it is a bit useful in tests where `Gen[Option[Int]]` is used to generate `Host`-s with or without port numbers. But I'm inclining to agree that this override seems redundant.

core/src/main/scala/org/http4s/headers/Forwarded.scala
satorg: Just to clarify: if I remove the next `from` method, then there will be no overrides for `from` anymore. Are you ok to leave this one just as `from` then or still prefer to rename to `fromHost`?

core/src/main/scala/org/http4s/headers/Forwarded.scala
satorg: ... Also it turned out that it is used in parsers :) So some internals leverage this form of `from` a little bit.

core/src/main/scala/org/http4s/headers/Forwarded.scala
rossabaker: :+1: to all of this.  Yeah, I think I like the `of` convention.

core/src/main/scala/org/http4s/headers/Forwarded.scala
rossabaker: Great question.  I think we mostly have `fromTypeOfArgument` elsewhere, and it makes it easier to add more later.

satorg:
@rossabaker @hamnis I have worked through most of your comments, feel free to review the changes once again :)

Moreover, I voluntarily decided to get rid of backticks in `Element`'s field names – I've renamed them to `maybe*` pattern since they're options actually. I've put it into the separate last commit so if you don't like it, we can drop it.

Also please feel free to make any other suggestions, if any. There's no rush on my side, I don't mind to continue polishing if necessary.

hamnis (approved):

Issue #3588, opened 2020-07-15
Digest Auth client middleware


kuzkdmy:
Http4s provide Digest auth middleware to implement server, but luck of client middleware
This can be really not so trivial task for somebody who will just start working with http4s. 
I have this already being implemented, need to revisit and contribute this to become a part of http4s project

Daenyth:
To make sure it's clear: this is a request to add a client middleware for `Digest` auth, similar to the server middleware that exists now?

kuzkdmy:
Yes, sorry that don't find time to make a pull request. I'll have this already being implemented, just need to revisit and make a pull request, will try to make this week.

rossabaker:
@kuzkdmy Great! It sounds like this is new and wouldn't change any internals, so if you target the series/0.21 branch, we can get it into the production release series sooner.

kuzkdmy:
Migrate from what we used with a very little refactoring
https://github.com/http4s/http4s/pull/3618/files
Thanks

PR #3619, opened 2020-08-02
Digest Auth client middleware #3588

kuzkdmy:


hamnis (approved):
On green

core/src/main/scala/org/http4s/middleware/authentication/DigestUtil.scala
```diff
@@ -0,0 +1,47 @@
+/*
+ * Copyright 2013-2020 http4s.org
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.http4s.middleware.authentication
```
rossabaker, line 7: This is a new package, right?  I think we should stick this in `org.http4s.internal`.

server/src/main/scala/org/http4s/server/middleware/authentication/DigestUtil.scala
```diff
@@ -30,6 +27,7 @@ private[authentication] object DigestUtil {
     * @param qop
     * @return
     */
+  @deprecated("Use org.http4s.middleware.authentication.DigestUtil instead", "0.21")
```
rossabaker, line 30: I expect this to be part of 0.21.7.  We can be more specific in the tag.

rossabaker (commented):
This looks nice.  Thank you!

It's been a while since I've used digest auth.  The `nc` comment is the one I'm most concerned about.

client/src/main/scala/org/http4s/client/middleware/DigestClient.scala
```diff
@@ -0,0 +1,104 @@
...
+  def apply[F[_]](
+      username: String,
+      password: String,
+      qop: String
+  )(client: Client[F])(implicit F: Concurrent[F]): Client[F] = {
+    def digestAuthHeader(req: Request[F], challenge: Challenge): Header = {
+      val realm = challenge.realm
+      val method = req.method.name
+      val uri = req.uri.path.toString
+      val nc = "00000001"
```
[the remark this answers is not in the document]
hamnis: nc and cnonce are in lockstep if memory serves, if you always provide a new cnonce, then you dont need to increase nc.
This should be verified with the rfc.

client/src/test/scala/org/http4s/client/middleware/DigestClientSpec.scala
```diff
@@ -0,0 +1,60 @@
...
+import org.http4s.testing.Http4sLegacyMatchersIO
+
+class DigestClientSpec extends Http4sSpec with Http4sLegacyMatchersIO {
+
+  val realm = "Test Server Realm"
+  val login = "login"
+  val password = "password"
+
+  def authStore(u: String): IO[Option[(String, String)]] =
+    IO(if (u == login) Some(u -> password) else None)
```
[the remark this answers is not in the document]
kuzkdmy: updated, thanks

client/src/main/scala/org/http4s/client/middleware/DigestClient.scala
```diff
@@ -0,0 +1,104 @@
...
+      val cnonce = {
```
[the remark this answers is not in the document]
kuzkdmy: + , updated

core/src/main/scala/org/http4s/middleware/authentication/DigestUtil.scala
kuzkdmy: updated

server/src/main/scala/org/http4s/server/middleware/authentication/DigestUtil.scala
kuzkdmy: updated

client/src/main/scala/org/http4s/client/middleware/DigestClient.scala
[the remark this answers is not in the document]
rossabaker: > The nc parameter stands for "nonce count".  The nc value is the
      hexadecimal count of the number of requests (including the current
      request) that the client has sent with the nonce value in this
      request.  For example, in the first request sent in response to a
      given nonce value, the client sends "nc=00000001".  The purpose of
      this parameter is to allow the server to detect request replays by
      maintaining its own copy of this count -- if the same nc value is
      seen twice, then the request is a replay.

This is challenging to do right.  We could maintain a `ConcurrentMap[String, Long]` for the nonce counter, but it would grow indefinitely unless we provide an eviction method.  We could maintain a single counter, but some servers get cranky if the nonce counter skips numbers.  We could parameterize a `String => F[Long]`, but we should have a default implementation.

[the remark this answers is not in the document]
rossabaker: I propose we add the following algebra:

```scala
trait NonceCounter[F[_]] {
  /** Returns the next `nc` value for the given nonce */
  def nc(nonce: String): F[Long]
  /** Call when a conversation is over, to prevent unbounded growth. */
  def evict(nonce: String): F[Unit]
}
```

Then we can provide an implementation around that with a `Sync[F]` constraint around a `concurrent.Map`.  @ChristopherDavenport has something like this in mules.  That implementation would be the default.  People who want something more sophisticated could then easily adopt it to a more industrial cache like mules.  Or who want a more lax implementation could create a constant version.

@kuzkdny, does this make sense?  Would you like to try adding this?  I think we should at least get that algebra in place before merging, so we don't break binary compatibility.

rossabaker (changes requested):
Blocking so we can get the nonce counter strategy in place.

0.21.7 is otherwise ready, so let's slip this to 0.21.8.

rossabaker:
There's another bug to fix elsewhere, so I don't expect 0.21.8 to be far behind.  Am still excited to have this. :smile:

client/src/main/scala/org/http4s/client/middleware/DigestClient.scala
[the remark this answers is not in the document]
kuzkdmy: @rossabaker , thanks. this all makes sense, I'll take a look this days.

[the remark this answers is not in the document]
rossabaker: I suppose there also could be an `evictAll: F[Unit]` on that algebra.  I don't feel strongly about that either way, but it's another decision to consider before we merge into 0.21.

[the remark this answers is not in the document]
kuzkdmy: Just take a closer look at [rfc7616](https://tools.ietf.org/html/rfc7616) will do some more update in client middleware and DigestUtils(that one that is new and really just a copy of server DigestUtils) to be better compatible(and avoid any next pull request binary compatibility issue)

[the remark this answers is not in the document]
kuzkdmy: Hi @rossabaker I've pushed updates related to NonceCounter, have covered this with tests, but really will be happy this to be good reviewed.
`class SimpleInMemoryStoreNonceCounter[F[_]](nonceTtl: FiniteDuration, cnonceF: () => String)(
    implicit
    T: Timer[F],
    C: Concurrent[F])
    extends NonceCounter[F] {
  private val cache = TrieMap.empty[String, Ref[F, Entry]]
  override def nc(nonce: String): F[(String, Long)] =
    (for {
      now <- T.clock.realTime(TimeUnit.MILLISECONDS)
      ref <- cache.getOrElseUpdate(nonce, Ref.unsafe(Entry(now, cnonceF(), 0))).pure[F]
      next <- ref.updateAndGet(d => d.copy(usedAt = now, nc = d.nc + 1))
      res <-
        if (next.nc > Max) ref.get >>= (_ => ref.updateAndGet(_ => Entry(now, cnonceF(), 1)))
        else next.pure[F]
    } yield (res.cnonce, res.nc)) <* C.start(autoCleanIfNotUsed(nonce))

  private def autoCleanIfNotUsed(nonce: String): F[Unit] =
    for {
      time <- T.clock.realTime(TimeUnit.MILLISECONDS)
      _ <- T.sleep(nonceTtl)
      _ <- cache.get(nonce) match {
        case Some(ref) => ref.get >>= (r => if (r.usedAt > time) ().pure[F] else evict(nonce))
        case None => ().pure[F]
      }
    } yield ()

  override def evict(nonce: String): F[Unit] =
    cache.remove(nonce).map(_ => ().pure[F]).getOrElse(().pure[F])

  override def evictAll: F[Unit] =
    cache.clear.pure[F]

  case class Entry(usedAt: Long, cnonce: String, nc: Long)
}`

 I'll try to find some time this week to make client to be fully compatible with rfc, and sure to be green line

client/src/main/scala/org/http4s/client/middleware/DigestClient.scala
```diff
@@ -0,0 +1,180 @@
...
+}
+
+object NonceCounter {
+  val Max: Long = java.lang.Long.parseLong("FFFFFFFF", 16)
+  def nextCnonce(): String = {
+    val bytes = new Array[Byte](16)
+    Random.nextBytes(bytes)
+    Base64.getEncoder.encodeToString(bytes)
+  }
+  def defaultNonceCounter[F[_]: Concurrent: Timer]: NonceCounter[F] =
```
rossabaker, line 143: `nextCnonce` is impure, which is a reminder that this counter is also impure.  We need `F[NonceCounter[F]]` for referential transparency, which ripples up to the `apply` methods returning `F[Client[F]]`.  This is a bit more annoying to construct, but also a bit more honest, and dishonesty has a way of eventually biting us in FP.

client/src/main/scala/org/http4s/client/middleware/DigestClient.scala
```diff
@@ -0,0 +1,180 @@
...
+
+class SimpleInMemoryStoreNonceCounter[F[_]](nonceTtl: FiniteDuration, cnonceF: () => String)(
+    implicit
+    T: Timer[F],
+    C: Concurrent[F])
+    extends NonceCounter[F] {
+  private val cache = TrieMap.empty[String, Ref[F, Entry]]
+  override def nc(nonce: String): F[(String, Long)] =
+    (for {
+      now <- T.clock.realTime(TimeUnit.MILLISECONDS)
```
rossabaker, line 155: Maybe the monotonic clock is safer here?  The difference will almost never matter, but feels more descriptive of this use case.

client/src/main/scala/org/http4s/client/middleware/DigestClient.scala
```diff
@@ -0,0 +1,180 @@
...
+      ref <- cache.getOrElseUpdate(nonce, Ref.unsafe(Entry(now, cnonceF(), 0))).pure[F]
+      res <- ref.updateAndGet { d =>
+        val next = d.copy(usedAt = now, nc = d.nc + 1)
```
rossabaker, line 158: Really surprised the compiler isn't making you do `1L` here instead of `1`.

client/src/main/scala/org/http4s/client/middleware/DigestClient.scala
```diff
@@ -0,0 +1,180 @@
...
+        if (next.nc > Max) Entry(now, cnonceF(), 1) else next
+      }
+    } yield (res.cnonce, res.nc)) <* C.start(autoCleanIfNotUsed(nonce))
```
rossabaker, line 161: If we get interrupted between `ref` and `res`, the clean won't run.  Corner case, but a bracket is probably better here.

client/src/main/scala/org/http4s/client/middleware/DigestClient.scala
```diff
@@ -0,0 +1,180 @@
...
+
+  private def autoCleanIfNotUsed(nonce: String): F[Unit] =
+    for {
+      time <- T.clock.realTime(TimeUnit.MILLISECONDS)
+      _ <- T.sleep(nonceTtl)
+      _ <- cache.get(nonce) match {
+        case Some(ref) => ref.get >>= (r => if (r.usedAt > time) ().pure[F] else evict(nonce))
```
rossabaker, line 168: `().pure[F]` is `F.unit`.

rossabaker, line 168: Does the `get` need to be atomic with the `evict`?  We'd have to be very unlucky for it to matter, but I think it could.

client/src/main/scala/org/http4s/client/middleware/DigestClient.scala
```diff
@@ -0,0 +1,180 @@
...
+  def defaultNonceCounter[F[_]: Concurrent: Timer]: NonceCounter[F] =
+    new SimpleInMemoryStoreNonceCounter[F](1.minute, nextCnonce)
+}
+
+class SimpleInMemoryStoreNonceCounter[F[_]](nonceTtl: FiniteDuration, cnonceF: () => String)(
+    implicit
+    T: Timer[F],
+    C: Concurrent[F])
+    extends NonceCounter[F] {
+  private val cache = TrieMap.empty[String, Ref[F, Entry]]
```
rossabaker, line 152: In my mind's eye, this was more `TrieMap[String, Entry]`, looping on the `Booleans` and `Option`s to retry concurrent operations where we lost the race.  I think this would make it easier to tighten up the atomicity mentioned below, though it would be a bit more verbose.

client/src/main/scala/org/http4s/client/middleware/DigestClient.scala
```diff
@@ -0,0 +1,180 @@
...
+    (for {
+      now <- T.clock.realTime(TimeUnit.MILLISECONDS)
+      ref <- cache.getOrElseUpdate(nonce, Ref.unsafe(Entry(now, cnonceF(), 0))).pure[F]
+      res <- ref.updateAndGet { d =>
+        val next = d.copy(usedAt = now, nc = d.nc + 1)
+        if (next.nc > Max) Entry(now, cnonceF(), 1) else next
+      }
+    } yield (res.cnonce, res.nc)) <* C.start(autoCleanIfNotUsed(nonce))
+
+  private def autoCleanIfNotUsed(nonce: String): F[Unit] =
```
rossabaker, line 163: This is neat.  I wasn't thinking of doing this, but it gives a good out-of-box experience.

client/src/main/scala/org/http4s/client/middleware/DigestClient.scala
```diff
@@ -0,0 +1,180 @@
...
+  def nextCnonce(): String = {
+    val bytes = new Array[Byte](16)
+    Random.nextBytes(bytes)
+    Base64.getEncoder.encodeToString(bytes)
+  }
+  def defaultNonceCounter[F[_]: Concurrent: Timer]: NonceCounter[F] =
+    new SimpleInMemoryStoreNonceCounter[F](1.minute, nextCnonce)
+}
+
+class SimpleInMemoryStoreNonceCounter[F[_]](nonceTtl: FiniteDuration, cnonceF: () => String)(
```
rossabaker, line 147: If you take a `Duration`, could have one that never expires on this same code.  But maybe that's expert use for which people should bring their own impl.

client/src/main/scala/org/http4s/client/middleware/DigestClient.scala
```diff
@@ -0,0 +1,180 @@
...
+      time <- T.clock.realTime(TimeUnit.MILLISECONDS)
+      _ <- T.sleep(nonceTtl)
+      _ <- cache.get(nonce) match {
+        case Some(ref) => ref.get >>= (r => if (r.usedAt > time) ().pure[F] else evict(nonce))
+        case None => ().pure[F]
+      }
+    } yield ()
+
+  override def evict(nonce: String): F[Unit] =
+    cache.remove(nonce).map(_ => ().pure[F]).getOrElse(().pure[F])
```
rossabaker, line 174: Instead of `.pure[F]`, we want the cache-modifying operations inside `F.delay`, or else the side effect has already run when we return the `F[Unit]`.  Same with `evictAll`.

rossabaker (commented):
This is cool.  I like how you're taking on auto-expiration.

I'm summoning @ChristopherDavenport, since he wrote mules, and might see things in this that I didn't.

client/src/main/scala/org/http4s/client/middleware/DigestClient.scala
```diff
@@ -0,0 +1,180 @@
...
+
+  /** Call when a conversation is over, to prevent unbounded growth. */
+  def evictAll: F[Unit]
+}
+
+object NonceCounter {
+  val Max: Long = java.lang.Long.parseLong("FFFFFFFF", 16)
+  def nextCnonce(): String = {
+    val bytes = new Array[Byte](16)
+    Random.nextBytes(bytes)
```
ChristopherDavenport, line 140: `Random` captures over Global mutable state, assuming this is alright.

client/src/main/scala/org/http4s/client/middleware/DigestClient.scala
ChristopherDavenport: Agree, if we have reached this we should leverage other tools.

client/src/main/scala/org/http4s/client/middleware/DigestClient.scala
```diff
@@ -0,0 +1,180 @@
...
+class SimpleInMemoryStoreNonceCounter[F[_]](nonceTtl: FiniteDuration, cnonceF: () => String)(
+    implicit
+    T: Timer[F],
+    C: Concurrent[F])
+    extends NonceCounter[F] {
+  private val cache = TrieMap.empty[String, Ref[F, Entry]]
+  override def nc(nonce: String): F[(String, Long)] =
+    (for {
+      now <- T.clock.realTime(TimeUnit.MILLISECONDS)
+      ref <- cache.getOrElseUpdate(nonce, Ref.unsafe(Entry(now, cnonceF(), 0))).pure[F]
```
ChristopherDavenport, line 156: The cache is mutable right? In which case no operations leveraging it are pure, not to mention the unsafe ref here.

client/src/main/scala/org/http4s/client/middleware/DigestClient.scala
ChristopherDavenport: We probably need a semaphore(1) on all trie operations. :grimacing:

ChristopherDavenport (commented):
This went from simple `Ref[Map[K, V]]` territory to something a bit more complicated, not sure if thats really what we wanted here.

client/src/main/scala/org/http4s/client/middleware/DigestClient.scala
rossabaker: I don't think the `Ref.unsafe` is problematic, per se: it's not being shared.  But the `cache.getOrElseUpdate` call is a side effect that should be in `F.delay` instead of a `.pure[F]` at the end.

rossabaker:
There are two alternate representations here that I think would simplify the nonce counter a lot, with different tradeoffs:

> In my mind's eye, this was more TrieMap[String, Entry]

The `TrieMap` is mutable, and requires a lot of the `F.delay`s discussed above, but is optimized for concucurrency.  (Indeed, it's in the `concurrent` package!).  A cats-effect wrapped triemap would be a neat nanolibrary.

> simple Ref[Map[K, V]] 

This is even simpler: the `Ref` will force the scary parts into an `F`.  It won't scale as well, but for most people, it shouldn't really need to.

ChristopherDavenport:
You're welcome! https://github.com/ChristopherDavenport/mapref/blob/master/core/src/main/scala/io/chrisdavenport/mapref/MapRef.scala#L503

kuzkdmy:
Sorry, for so big pause in activity, just had to finish a lot of work and took a vacation. Will revisit all required changes this week.

kuzkdmy:
Hi, I'm still remember about this pull request. I'll take a more close look, want this test to pass, some tricks are required. Will try to take a look this weekends 
image

client/src/main/scala/org/http4s/client/digest/SimpleInMemoryStoreNonceCounter.scala
```diff
@@ -1,18 +1,23 @@
...
 
 import scala.concurrent.duration.FiniteDuration
 
-class SimpleInMemoryStoreNonceCounter[F[_]](nonceTtl: FiniteDuration, cnonceF: () => String)(
-    implicit
-    T: Timer[F],
-    C: Concurrent[F])
+class SimpleInMemoryStoreNonceCounter[F[_]](
+    nonceTtl: FiniteDuration,
+    cnonceF: () => String,
```
rossabaker, line 19: I think this should be an `F[String]` to reflect its impurity.  Sequencing that into `val current` is easy.  Getting a second one into `next`, without calculating it in advance, is a bit of a brain teaser.

client/src/main/scala/org/http4s/client/digest/SimpleInMemoryStoreNonceCounter.scala
```diff
@@ -0,0 +1,53 @@
...
+      }
+    } yield ()
+
+  override def evict(nonce: String): F[Unit] =
+    cacheRef.update(_ - nonce)
+
+  override def evictAll: F[Unit] =
+    cacheRef.set(Map.empty[String, Entry])
+
+  case class Entry(usedAt: Long, cnonce: String, nc: Long)
```
rossabaker, line 52: Can this be private so implementation can evolve?

client/src/main/scala/org/http4s/client/middleware/DigestClient.scala
```diff
@@ -0,0 +1,115 @@
...
+      password = password,
+      qop = qop,
+      nonceCounter = NonceCounter.defaultNonceCounter[F].pure[F]
+    )(client)
+
+  def apply[F[_]](
+      username: String,
+      password: String,
+      qop: String,
+      nonceCounter: F[NonceCounter[F]]
```
rossabaker, line 49: Does this have to be in `F`, or can we take a `NonceCounter[F]`?

client/src/main/scala/org/http4s/client/digest/NonceCounter.scala
```diff
@@ -0,0 +1,37 @@
...
+  /** Call when a conversation is over, to prevent unbounded growth. */
+  def evict(nonce: String): F[Unit]
+
+  /** Call when a conversation is over, to prevent unbounded growth. */
+  def evictAll: F[Unit]
+}
+
+object NonceCounter {
+  val MaxNonce: Long = java.lang.Long.parseLong("FFFFFFFF", 16)
+  def nextCnonce(): String = {
```
rossabaker, line 30: This is impure and should be an `F[String]`.  It will require a `Sync` constraint.

client/src/main/scala/org/http4s/client/digest/NonceCounter.scala
```diff
@@ -0,0 +1,37 @@
...
+    val bytes = new Array[Byte](16)
+    Random.nextBytes(bytes)
+    Base64.getEncoder.encodeToString(bytes)
+  }
+  def defaultNonceCounter[F[_]: Concurrent: Timer]: NonceCounter[F] =
```
rossabaker, line 35: And when `nextCnonce` becomes effectful, this returns in `F` as well.

rossabaker (commented):
Thank you for sticking with this, and I'm sorry it took so long for a re-review.  I'd love to get this into the release.  Just a couple more questions before we lock in the public API.

rossabaker:
De-milestoned because we're not going to block 0.21.8 on it, but I'd love to have it.

I'm back to inbox zero, so hopefully we can get an 0.21.9 done more quickly if this misses the bus.

client/src/main/scala/org/http4s/client/digest/NonceCounter.scala
kuzkdmy: Makes sense, revisiting now.

FrancescoSerra:
If there are no objections, I'll take over from here and finish off with the implementation

Issue #3606, opened 2020-07-28
Be able to create an AsyncHttpClient with a custom request timeout 


henricook:
I mentioned this briefly on Gitter and it was suggested I raise an issue.

I want to be able to create an AsyncHttpClient with a custom request timeout. At the moment I can use `AsyncHttpClient.resource[Task]()` to get the `defaultConfig` from http4s `AsyncHttpClient`. I might have missed something (please tell me if I have!) but I don't think I can tweak this to give me a custom request timeout.

`defaultConfig` is:

```scala
  val defaultConfig = new DefaultAsyncHttpClientConfig.Builder()
    .setMaxConnectionsPerHost(200)
    .setMaxConnections(400)
    .setRequestTimeout(defaults.RequestTimeout.toMillis.toInt)
    .setThreadFactory(threadFactory(name = { i =>
      s"http4s-async-http-client-worker-${i}"
    }))
    .setCookieStore(new NoOpCookieStore)
    .build()
```

I tried copying and pasting this into my own code but `NoOpCookieStore` is a package-private class and `threadFactory` was recently moved to http4s internals where I can't access it.

So my goal is to have one client with a 30 second timeout in use by my application (e.g. when talking to external services) but I want my acceptance test http4s client to have >30 seconds (e.g. 45 seconds) so it doesn't timeout whilst an external service being tested is timing out.

**TL;DR** - I just want to be able to set the request timeout for a client

Related:

@hamnis kindly pointed out that this is possible, which gives me an immediate workaround:

```scala
 AsyncHttpClient.resource({
      val b = new DefaultAsyncHttpClientConfig.Builder(AsyncHttpClient.defaultConfig)
      b.setRequestTimeout(3000)
      b.build()
    })
```

hamnis:
We could add a function to make this easier.

`AsyncHttpClient.configure(fn: DefaultAsyncHttpClientConfig.Builder => DefaultAsyncHttpClientConfig.Builder): HttpClientConfig`

jyoo980:
Could I try my hand at this?

I've use Scala at work but it'll be my first foray into Scala open-source.

hamnis:
Fixed by #3623

PR #3623, opened 2020-08-04
Adding function to configure an AsyncHttpClient with a client config builder

jyoo980:
Hello 👋 

I use Scala at work, but I'd consider myself a relative newcomer to the language. I tried implementing the work outlined in #3606 as my first foray into open-source Scala. I'm eager to learn so please let me know if I'm on the right track 🙂

### DIFF
--- a/async-http-client/src/main/scala/org/http4s/client/asynchttpclient/AsyncHttpClient.scala
+++ b/async-http-client/src/main/scala/org/http4s/client/asynchttpclient/AsyncHttpClient.scala
@@ -57,7 +57,6 @@ object AsyncHttpClient {
     * Create an HTTP client based on the AsyncHttpClient library
     *
     * @param config configuration for the client
-    * @param ec The ExecutionContext to run responses on
     */
   def resource[F[_]](config: AsyncHttpClientConfig = defaultConfig)(implicit
       F: ConcurrentEffect[F]): Resource[F, Client[F]] =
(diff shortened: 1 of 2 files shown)


async-http-client/src/main/scala/org/http4s/client/asynchttpclient/AsyncHttpClient.scala
```diff
@@ -57,7 +57,6 @@ object AsyncHttpClient {
     * Create an HTTP client based on the AsyncHttpClient library
     *
     * @param config configuration for the client
-    * @param ec The ExecutionContext to run responses on
```
jyoo980, line 60: These functions no longer consume an `ExecutionContext`, removed outdated comments.

async-http-client/src/test/scala/org/http4s/client/asynchttpclient/AsyncHttpClientSpec.scala
```diff
@@ -10,6 +10,18 @@ package asynchttpclient
 
 import cats.effect.IO
 
-class AsyncHttpClientSpec extends ClientRouteTestBattery("AsyncHttpClient") {
+class AsyncHttpClientSpec extends ClientRouteTestBattery("AsyncHttpClient") with Http4sSpec {
+  implicit val testContext = testContext()
+
   def clientResource = AsyncHttpClient.resource[IO]()
+
+  "AsyncHttpClient configure" should {
```
jyoo980, line 18: No tests existed at all for `AsyncHttpClient`, I saw what was being done in other test files and tried to replicate it here.

hamnis: I dont think the configuration has a useful equals, so this test is not working

jyoo980: I've updated the tests to assert on the fields themselves, rather than the object.

jyoo980:
@rossabaker thanks for your feedback!

Looking to contribute more in the future 😁

rossabaker:
Picked as 3e687e7, so this can be in 0.21.7. Thanks again, @jyoo980.

PR #3624, opened 2020-08-05
Use defaultHttpClient in JettyClient builder

Taig:
I believe this might have been an oversight, comparing with alternativ clients such as [AsyncHttpClient](https://github.com/http4s/http4s/blob/master/async-http-client/src/main/scala/org/http4s/client/asynchttpclient/AsyncHttpClient.scala#L43). Please feel free to close this PR if the current implementation was chosen for compatibility reasons.


hamnis:
Thanks for the PR @Taig

Good spotted, please rebase this to series/0.21 so we can get a patch out faster.

PR #3625, opened 2020-08-06
Close urlConn in StaticFile.fromURL

LolHens:
Previous discussion in #3584

The URLConnection in StaticFile.fromURL is never closed (if the resource is not expired).
This PR reuses the already opened URLConnection to read the response body which in effect causes fs2 to close the URLConnection once the body is fully read.

This PR also wraps `urlConn.getInputStream.close()` in a `Try` because it should just be a last attempt to close the resource and should return NotModified in case of errors.

Because of the discussion in #3594 I opened this against series/0.21

### DIFF
--- a/core/src/main/scala/org/http4s/StaticFile.scala
+++ b/core/src/main/scala/org/http4s/StaticFile.scala
@@ -19,6 +19,7 @@ import java.net.URL
 import org.http4s.Status.NotModified
 import org.http4s.headers._
 import org.log4s.getLogger
+import scala.util.Try
 
 object StaticFile {
   private[this] val logger = getLogger
@@ -88,7 +89,7 @@ object StaticFile {
           val headers = Headers(lenHeader :: lastModHeader ::: contentType)
 
           blocker
-            .delay(url.openStream)
+            .delay(urlConn.getInputStream)
             .redeem(
               recover = {
                 case _: FileNotFoundException => None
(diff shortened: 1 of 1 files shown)


LolHens:
Which idea?

core/src/main/scala/org/http4s/StaticFile.scala
```diff
@@ -104,7 +105,7 @@ object StaticFile {
             )
         } else
           blocker
-            .delay(urlConn.getInputStream.close)
+            .delay(Try(urlConn.getInputStream.close()))
```
rossabaker, line 108: If we used an old-fashioned try-catch block here, we'd save an allocation for a discarded `Try` value, and be able to log a warning on any exceptions.  Without seeing what's being thrown, it's hard to tell whether they'd be noise or something worth addressing.

In any case, returning the `NotModified` here is correct and a worthwhile improvement.

satorg: From my prospective something like the following snippet would look nicer here:
```
blocker
  .delay(urlConn.getInputStream.close())
  .handleError(_ => ())
  .as( ... )
```
Because it shows intent in a clearer way.

I'm mildly concerned about using `Try` just for swallowing exceptions – it is definitely not what it was designed for (even if it does the job).

LolHens: Yes this looks indeed much nicer.

PR #3634, opened 2020-08-08
implementing QueryParamEncoder[LocalDate] and QueryParamDecoder[LocalDate]

jyoo980:
Was looking through some issues when I found this conversation

> @LukaJCB - #2724 added support for `Instant`. What other types would you find useful - `ZonedDateTime` and `LocalDate`?

in #2670. I added support for `LocalDate`. I have something in my fork for `ZonedDateTime`, but maybe that'll come later since I'm having trouble testing it correctly.

I hope this is a useful contribution!

### DIFF
--- a/tests/src/test/scala/org/http4s/QueryParamCodecSpec.scala
+++ b/tests/src/test/scala/org/http4s/QueryParamCodecSpec.scala
@@ -11,7 +11,8 @@ import cats.data._
 import cats.implicits._
 import cats.laws.discipline.{arbitrary => _, _}
 import java.time.format.DateTimeFormatter
-import java.time.Instant
+import java.time.{Instant, LocalDate}
+
 import org.scalacheck.{Arbitrary, Cogen, Gen}
 import org.scalacheck.Arbitrary._
 import org.scalacheck.Prop._
@@ -25,6 +26,7 @@ class QueryParamCodecSpec extends Http4sSpec with QueryParamCodecInstances {
   checkAll("Long QueryParamCodec", QueryParamCodecLaws[Long])
   checkAll("String QueryParamCodec", QueryParamCodecLaws[String])
   checkAll("Instant QueryParamCodec", QueryParamCodecLaws[Instant])
+  checkAll("LocalDate QueryParamCodec", QueryParamCodecLaws[LocalDate])
 
   // Law checks for instances.
   checkAll(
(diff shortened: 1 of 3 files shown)


jyoo980:
Looks like there are dependency/compilation issues for adding `ZonedDateTime`, fails for anything below Scala `2.13` and anything other than JDK `1.8`, maybe we can exclude this decoder/encoder for now?

rossabaker:
Sorry, I started reviewing this, and I don't know where it ended up.

We have an `import org.http4s.internal.CollectionCompat.CollectionConverters._`.  If you need to bridge the gap between the collections API in other ways, they can be added to `CollectionCompat` in the two source directories, but I think that will fix the remaining issue.

Otherwise, looks good!

rossabaker:
It also looks binary compatible, so we can cherry-pick this back to 0.21.

jyoo980:
Thanks for the feedback, I'll work on this over my boring weekend 😉

tests/src/test/scala/org/http4s/QueryParamCodecSpec.scala
```diff
@@ -75,9 +85,31 @@ trait QueryParamCodecInstances { this: Http4sSpec =>
...
+      Gen
+        .choose[Long](LocalDate.MIN.toEpochDay, LocalDate.MAX.toEpochDay)
+        .map(LocalDate.ofEpochDay))
+
+  implicit val ArbitraryZonedDateTime: Arbitrary[ZonedDateTime] = {
+    val zoneIdStrs = ZoneId.getAvailableZoneIds.asScala.toSeq
+    Arbitrary(Gen.oneOf[String](zoneIdStrs).map { zoneIdStr =>
+      val zoneId: ZoneId = ZoneId.of(zoneIdStr)
+      val secSinceEpoch: Long =
+        Random.between(Instant.MIN.getEpochSecond, Instant.MAX.getEpochSecond)
```
jyoo980, line 111: `Random.between` is only available in 2.13, I'll write a workaround.

jyoo980:
Looks like a [Flakier than it's worth on CI](https://github.com/http4s/http4s/blob/af4a5b5242685a0add29f44ed60335538b380f2b/blaze-client/src/test/scala/org/http4s/client/blaze/Http1ClientStageSpec.scala#L195) test failed on the latest run. Other than that, looks like the tests related for my change passed.

tests/src/test/scala/org/http4s/QueryParamCodecSpec.scala
```diff
@@ -103,18 +103,12 @@ trait QueryParamCodecInstances { this: Http4sSpec =>
...
 
   implicit val ArbitraryZonedDateTime: Arbitrary[ZonedDateTime] = {
-    // Can't use Random.between because it breaks compatibility with
-    // anything below Scala 2.13
-    def randBetween(start: Long, end: Long): Long = {
-      val diff = end - start
-      (start + math.random() * diff).toLong
+    val zoneIds: Seq[String] = ZoneId.getAvailableZoneIds.asScala.toSeq
+    Arbitrary {
+      for {
```
jyoo980, line 108: Honestly embarrassed that I didn't think of this sooner

PR #3666, opened 2020-08-27
#3665 Bug in CSRF error handler for `onFailure`

### LINKED ISSUE #3665: Bug in CSRF error handler for `onFailure`
### Intention

I wanted to handle an error during CSRF validation with the `onFailure` response of the CSRF class.
The problem with some of the browsers (e.g. Firefox, Chrome) is that they do not remove session cookies. They provide settings which use the session cookies even when the browser was restarted.

The idea is to catch the CSRF validation error and provide an own `onFailure` response that sends the user to a page where a new CSRF cookie is set. (For `GET` requests.)

### Current situation

Using http4s Version 0.21

I tried the following approach by adding an own `onFailure` response to the specification of the CSRF object.

```
csrfProtect = csrfBuilder
  .withClock(csrfClock)
  .withCookieName(CSRF_COOKIE)
  .withCSRFCheck(CSRF.checkCSRFinHeaderAndForm[IO, IO](CSRF_COOKIE, FunctionK.id))
  .withOnFailure(
    Response(status = Status.SeeOther, headers = Headers(List(Header("Location", "/"))))
      .removeCookie(CSRF_COOKIE)
  )
  .build
```

### Expected result

If a user with an invalid CSRF cookie opens the site, the CSRF check will fail. The `onFailure` response should redirect the user to the frontpage and refresh the CSRF cookie.

### Actual result

The `onFailure` response is not used.

I checked the code and assume that the hard-coded `Response[G](Status.Forbidden)` in some of the check methods could be the reason. These could be replaced by the `onFailure` reponse which by default is also a `Response[G](Status.Forbidden)`.

One of these checks is:
```
  def checkCSRFToken(r: Request[G], respAction: F[Response[G]], rawToken: String)(implicit
      F: Sync[F]): F[Response[G]] =
    if (!headerCheck(r))
      F.pure(onFailure)
    else
      (for {
        c1 <- CSRF.cookieFromHeadersF[F, G](r, cookieSettings.cookieName)
        raw1 <- F.fromEither(extractRaw(c1.content))
        raw2 <- F.fromEither(extractRaw(rawToken))
        response <-
          if (CSRF.isEqual(raw1, raw2)) respAction
          else F.raiseError[Response[G]](CSRFCheckFailed)
        newToken <- signToken[F](raw1) //Generate a new token to guard against BREACH.
      } yield response.addCookie(createResponseCookie(newToken)))
        .recover {
          case CSRFCheckFailed => Response[G](Status.Forbidden)
        }
```
The recover uses the hard-coded `Forbidden`. Also in the `validate` method the `recover` does the same.

faustas:
- The `recover` cases in `validate` and `checkCSRFToken` use the defined `onFailure` response. This is by definition already `Response[G](Status.Forbidden)` and does not change the former behavior of the hard-coded responses.
- Add tests for a CSRF object that has a defined `onFailure` response.
- Fixes #3665

### DIFF
--- a/server/src/main/scala/org/http4s/server/middleware/CSRF.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/CSRF.scala
@@ -174,7 +174,7 @@ final class CSRF[F[_], G[_]] private[middleware] (
           newToken <- signToken[F](raw)
         } yield res.addCookie(createResponseCookie(newToken)))
           .recover {
-            case CSRFCheckFailed => Response[G](Status.Forbidden)
+            case CSRFCheckFailed => onFailure
           }
       case None =>
         if (createIfNotFound)
@@ -202,7 +202,7 @@ final class CSRF[F[_], G[_]] private[middleware] (
         newToken <- signToken[F](raw1) //Generate a new token to guard against BREACH.
       } yield response.addCookie(createResponseCookie(newToken)))
         .recover {
-          case CSRFCheckFailed => Response[G](Status.Forbidden)
+          case CSRFCheckFailed => onFailure
         }
 
   /** Check for CSRF validity for an unsafe action. */
--- a/server/src/test/scala/org/http4s/server/middleware/CSRFSpec.scala
+++ b/server/src/test/scala/org/http4s/server/middleware/CSRFSpec.scala
@@ -72,10 +72,14 @@ class CSRFSpec extends Http4sSpec {
 
   val csrfCatchFailure: CSRF[IO, IO] = CSRF
     .withGeneratedKey[IO, IO](defaultOriginCheck)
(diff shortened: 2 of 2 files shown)


faustas:
The failed tests seem to be out of scope of the changes. Locally I executed the tests with `+test` and everything was `green`. Maybe someone can restart the tests to see if it is just a problem of the test system.

hamnis:
needs a scalafmt

hamnis:
Test failure is unrelated

jan0sch:
Could we have this merged or does it need some work?

faustas:
`ScalafmtAll` executed succesfully

hamnis (approved):

PR #3668, opened 2020-08-31
Source-Code Reorganisation

diesalbla:
This is a reorganisation of the source modules.  This only affects the build but is source and binary compatible.

-  "modules" top-level folder to contain all source code modules.

- A "clients" subdirectory for building clients using third-party libraries: `async-http-client`, `blaze`, `ember`, `jetty`, `okhttp`.

- A "servers" subdirectory for building servers using third-partylibraries: `blaze`, `ember`, `jetty`, `servlet`, `tomcat`.

- A "codecs" subdirectory for modules that build entity encoders and decoders using third-party Json or XML libraries:   `argonaut`, `boopickle`, `circe`, `jawn`, `json4s`, `play`, `scala-xml`, `scalatags`, `twirl`.


rossabaker:
I generally like the idea, but we've also been talking for quite a while about splitting into multiple repositories on their own release cycle.  This is particularly compelling for the codecs and third-party servers, where we have to manage versions between both http4s changes and the integrated libraries.  We've already started this process with newer modules like netty, finagle, and armeria, as well as the jdk-http-client that requires a newer JDK

Maybe a modules directory for what's left is nice -- I see that in a lot of projects.  I think that's core, server, client, and whichever concrete backend we keep for documentation purposes (probably ember or a pure JDK solution).

diesalbla:
@rossabaker To clarify: is the fragmentation in separate libraries with independent release cycles a done conclusion, or an ongoing discussion? I know that synchronising an aggregation of many separate 3rd-party libraries is a pain... but it worries me that splitting modules in libraries would be moving pain on users of the library, having to manage many dependencies with different versions and checking separate source repositories.

rossabaker:
Separate release cycles are not a foregone conclusion, but as the number of integrations grows, I think the status quo becomes less tenable.  Without it, when there is a breaking change in an integrated module, our choices are:

1. Not to support that new version
2. Issue a breaking change to all of http4s
3. Support the new version in a side repo, like we used to do with argonaut

I think Scala Steward would do a great job on this in public repos and any private repos where it's integrated.  My hope is that sbt-updates would work well on private apps.  

We've managed so far by having breaking changes to our core enough that we're on 0.21, but those 21 are all effectively major versions.  Without cats and fs2 forcing our hand as often as before, I think we need to revisit our methods.

rossabaker:
If not all of our integrations publish for Dotty, our hand may be forced on splitting up the repo, and thus the release cycle.

diesalbla:
since this is ongoing discussion, PR has no landing date. Withdrawing

PR #3669, opened 2020-08-31
Merge server, client, and dsl into core

diesalbla:
As far as I have seen, people never use only the `core` module of `http4s` alone: they end up using either a `server`, or a `client`, or both. Even if they use a library that only targets this interfaces, and not one of the server or client implementations of `http4s`, they are not using the `core` alone. As for the `dsl` module, since this has the main spot in the `http4s` tutorials, and is more integrated with the `Uri` and `Request` constructors, it can be integrated further.

So, in this PR we merge the source code of the `main` folders of the `server`, `client`, and `dsl` modules into the `core` module. The `test` code, however, since it involves dependencies on upper modules, we have not merge yet in case that would cause circular dependencies.

This change could be source/binary incompatible or compatible, but in a strange way. 
- We are not modifying the source code of any class in the client, dsl, or server modules, so there should be no source compatibility problem. 
- We are also not modifying the bytecode in the `.class` files, so there should not be any binary incompatibility (class not found or method not found) problem either; and due to transitive dependencies.
- There could be some "package" incompatibility, in the sense that the contents of different versions of a same Jar (the `http4s-client` or `http4s-server`) have changed; however, since the `core` is now a transitive dependency, this should not impact any users of this library (all should still be in classpath).


IndiscriminateCoding:
>  people never use only the `core` module of `http4s` alone

I personally have [example](https://github.com/IndiscriminateCoding/api4s/blob/dev/build.sbt#L25) where a project only depends on `core` module without a `client` or `server` modules. Http4s `core` module does have HTTP-related datatypes which is useful on its own (to be more specific, my example uses `MediaType` to parse/analyze swagger media types).

As for `dsl` module, again, api4s doesn't use it because both client and server are auto-generated.

>  As for the `dsl` module, since this has the main spot in the `http4s` tutorials, and is more integrated with the `Uri` and `Request` constructors, it can be integrated further

Do you have a specific example of separate modules preventing integration or where ergonomic of `dsl` can be improved?

rossabaker:
Part of the vision of this project is to be able to run multiple ways of describing a service on multiple backends.  http4s-dsl was easy to write and Unfiltered was popular at the time, and it became the default way of writing services.  But projects like api4s, or alternative descriptors like rho and directives, offer alternatives.  I definitely think of DSL as distinct.

Another reason to keep client and server separate is that a scala.js client has long been proposed, but a scala.js server has never been in scope.  Even if the abstract server could be made to work, without any backends, bloat is a concern.

Daenyth:
Bloat is a concern on the jvm backend as well. For example, I have some low-level jar stuff in my work stack that uses core just for the uri type; I wouldn't want to bloat applications which get deployed, since deploy time is already slow. I'd actually prefer if the modules were split a bit *more*

kubukoz:
Yeah, I'll back up @daenyth here. Core is good as a dependency in libraries. If anything, I'd split it even more, so that the HTTP model itself (no entity encoders, codecs, syntax) would be separate, much like in sttp.

rossabaker:
Breaking out the model would be tricky, without splitting the root package across modules, adding an import tax, or adding confusing aliases.  That's probably its own proposal.

rossabaker:
This spurred a good discussion, but the consensus is, if anything, to move the opposite way.  If someone wants to seriously propose that, and how to divide the packages, let's spin off a separate issue.

PR #3691, opened 2020-09-12
Fixes QueryParamCodecSpec flakiness

### LINKED ISSUE #3664: ZonedDateTime QueryParamCodec flakiness
[error] x QueryParamCodec.decode . emapValidatedNel(ValidNel) . encode == successNel (40 ms)
[error]  Falsified after 86 passed tests.
[error]  > ARG_0: 2008-05-06T08:06:41Z[GMT0]
[error]  The seed is g9ZfCxo-M0krxW4NucJnoiiQa5IeWZmxorikUy8lprE=
[error] org.http4s.QueryParamCodecSpec.check(QueryParamCodecSpec.scala:22)
[error] org.http4s.Http4sSpec.$anonfun$checkAll$2(Http4sSpec.scala:77)
[error] x QueryParamCodec.decode . emapValidatedNel(InvalidNel) . encode == failedNel (22 ms)
[error]  Falsified after 35 passed tests.
[error]  > ARG_0: 1974-06-22T22:16:50Z[GMT0]
[error]  The seed is z-phGsJ7PKyG60xGk1GG0qbA8soFVDGygkZxecwPRIF=
[error] org.http4s.QueryParamCodecSpec.check(QueryParamCodecSpec.scala:22)
[error] org.http4s.Http4sSpec.$anonfun$checkAll$2(Http4sSpec.scala:77)

satorg:
Should resolve #3664

The root cause of the issue is a bug in Java 1.8 which fails parsing of zoned dates/times with "GMT0" zone id.
Please find more details on this here: https://bugs.java.com/bugdatabase/view_bug.do?bug_id=8138664

The suggested fix simply suppresses generation of `ZonedDateTime` instances with "GMT0" in `QueryParamCodecSpec` if the test runs on Java 1.8.

It doesn't fix the corresponding `QueryParamCodec` itself – I'm not sure if it would be a correct approach to try mitigating Java bugs on this level.

Also it slightly improves output for this test as well as other tests based on `Http4sSpec`.

### DIFF
--- a/testing/src/test/scala/org/http4s/Http4sSpec.scala
+++ b/testing/src/test/scala/org/http4s/Http4sSpec.scala
@@ -72,6 +72,7 @@ trait Http4sSpec
       p: Parameters,
       f: FreqMap[Set[Any]] => Pretty): Fragments = {
     addFragment(ff.text(s"$name  ${props.name} must satisfy"))
+    addBreak
     addFragments(Fragments.foreach(props.properties.toList) {
       case (name, prop) =>
         Fragments(name in check(prop, p, f))
--- a/tests/src/test/scala/org/http4s/QueryParamCodecSpec.scala
+++ b/tests/src/test/scala/org/http4s/QueryParamCodecSpec.scala
@@ -103,12 +103,23 @@ trait QueryParamCodecInstances { this: Http4sSpec =>
         .map(LocalDate.ofEpochDay))
 
   implicit val ArbitraryZonedDateTime: Arbitrary[ZonedDateTime] = {
-    val zoneIds: Seq[String] = ZoneId.getAvailableZoneIds.asScala.toSeq
+    val zoneIds = {
+      val zoneIdIter = ZoneId.getAvailableZoneIds.asScala.iterator
+      // Workaround for a bug in Java 1.8:
+      //   - https://bugs.java.com/bugdatabase/view_bug.do?bug_id=8138664
+      // See also the original issue:
+      //   - https://github.com/http4s/http4s/issues/3664
+      if (sys.props.get("java.specification.version").contains("1.8"))
+        zoneIdIter.filterNot(_ == "GMT0")
+      else
+        zoneIdIter
+    }.toSeq
+
     Arbitrary(
       for {
-        secSinceEpoch <- Gen.choose[Long](Instant.EPOCH.getEpochSecond, Instant.now.getEpochSecond)
+        instant <- Arbitrary.arbitrary[Instant] // re-use `Gen` from `Arbitrary[Instant]` here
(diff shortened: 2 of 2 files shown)


testing/src/test/scala/org/http4s/Http4sSpec.scala
```diff
@@ -72,6 +72,7 @@ trait Http4sSpec
       p: Parameters,
       f: FreqMap[Set[Any]] => Pretty): Fragments = {
     addFragment(ff.text(s"$name  ${props.name} must satisfy"))
+    addBreak
```
satorg, line 75: Inserts a break between a particular test name and all its cases. Without this break output for the first test case continues on the same line.

tests/src/test/scala/org/http4s/QueryParamCodecSpec.scala
```diff
@@ -103,12 +103,23 @@ trait QueryParamCodecInstances { this: Http4sSpec =>
...
+      if (sys.props.get("java.specification.version").contains("1.8"))
+        zoneIdIter.filterNot(_ == "GMT0")
+      else
+        zoneIdIter
+    }.toSeq
+
     Arbitrary(
       for {
-        secSinceEpoch <- Gen.choose[Long](Instant.EPOCH.getEpochSecond, Instant.now.getEpochSecond)
+        instant <- Arbitrary.arbitrary[Instant] // re-use `Gen` from `Arbitrary[Instant]` here
```
jyoo980, line 120: I'm not sure if re-using `Gen` from `Arbitrary[Instant]` here is the best idea. The reason why I chose to implement a specific generator here is because `ZonedDateTime` doesn't play nice with very large or very small values. 

That's why I clamped it from `Instant.EPOCH` to `Instant.now`. I'm not too sure if this is relevant, though.

satorg: Well, to be honest, I'm not that familiar with `ZonedDateTime` to reason about it. But at least, I haven't seen any issues with neither 'far-in-the-part` nor 'far-in-the-future' values in this test, so I decided – why not?

jyoo980: Ah I see. Well it looks like as long as the arbitrarily generated instants don't fall on `MIN` or `MAX`, it shouldn't be too much of a problem. I just came across this article as I was debugging earlier on.

https://stackoverflow.com/a/45044943

hamnis (approved):

rossabaker:
I don't think working around that JDK bug for an arcane zone ID that the Java team hasn't bothered to backport should be in scope for us.  If a user suffers from this in practice, we can revisit, but this looks good to me.

PR #3692, opened 2020-09-12
partial unification hint in Readme.md

hartmut27:
little hint for not needing partial unification switch on any more since scala 2.13, since done by default


rossabaker:
Thanks! I'm going to pull this into our series/0.21 line so it appears in those docs, too.

PR #3701, opened 2020-09-20
Parse raw headers only if its name matches

wjoel:
Otherwise the header we're matching against will be parsed even if its name doesn't match.

### DIFF
--- a/core/src/main/scala/org/http4s/HeaderKey.scala
+++ b/core/src/main/scala/org/http4s/HeaderKey.scala
@@ -91,7 +91,7 @@ object HeaderKey {
       header match {
         case h if runtimeClass.isInstance(h) =>
           Some(header.asInstanceOf[HeaderT])
-        case Header.Raw(name, _) if name == header.name && runtimeClass.isInstance(header.parsed) =>
+        case Header.Raw(_, _) if name == header.name && runtimeClass.isInstance(header.parsed) =>
           Some(header.parsed.asInstanceOf[HeaderT])
         case _ =>
           None


satorg:
I think it could be worthful to add a couple of tests for this change – some positive and negative scenario. wdyt?

hamnis:
Oh. this is a name shadowing. Well spotted.

wjoel:
A test would be nice, I agree. Seems a bit tricky but I have an idea, though I likely won't have time to try it before the weekend.

rossabaker:
I also think this can be released against series/0.21.

wjoel:
Sorry, I can't come up with a test for this.

As far as I can tell it would require mocking either `Header.Raw.parsed` or (object) `Header.unapply`, and the first isn't possible with Mockito because it's a final class and the second isn't possible because it's an object.

Using the debugger shows that this calls `Header.Raw.parsed` also for `Boom` without my fix (and that my change eliminates that call), but I don't know how to translate it into a test:

```scala
import org.http4s.headers.`Content-Location`

class DoNotParseUnnecessarily extends Http4sSpec {
  "HeaderKey matchHeader".can {
    "Skip parsing headers that don't match the name" in {
      val s = "http://www.foo.com"
      val Right(uri) = Uri.fromString(s)
      val hs = Headers.of(
        Header("Boom", "123"),
        Header("Content-Location", s)
      )

      `Content-Location`.from(hs) must beSome(`Content-Location`(uri))
    }
  }
}
```

Maybe there's a way, but I'm stumped.

satorg:
Ah, seems I got it. This check for name equality doesn't change any observable behavior but speeds `matchHeader` up by letting to avoid parsing a "raw" header. Sorry, all this header-related mechanic is a way complicated, didn't catch it from the first time. I agree, that it'd be too difficult to test this particular case appropriately and it should be safe to go without it.

wjoel:
No problem, @satorg. It was worth a shot. @rossabaker I don't think I have anything more to add, feel free to merge if it looks good to you.

rossabaker:
Cherry-picked for release in 0.21.8. Thanks again!

PR #3703, opened 2020-09-21
Update streaming.md

Grogs:
`client.stream(req).flatMap(_)` returns a partially applied function. Changing `_` to `_.body` gives the `Stream[F, Byte]` that the docs describe.


rossabaker (approved):
:+1: Thanks!  For whatever reason, this one isn't covered by the mdoc, which lets things like this slip through.

I'm going to cherry-pick this back to 0.21, so it goes in those docs.

PR #3705, opened 2020-09-22
Dependency bumps for 0.21

rossabaker:
Retargets #3704. #3695, #3698, #3696, #3688.


rossabaker:
I'm going to broaden the scope of this PR.  It looks like we've got several here to catch up on.

PR #3708, opened 2020-09-22
Drain Body in DefaultOnError

### LINKED ISSUE #3707: Improve Logging On Failed Expect
https://github.com/http4s/http4s/blob/master/client/src/main/scala/org/http4s/client/DefaultClient.scala#L228

This default response does not actually compile the stream or observe the failed bodies so they do not appear in loggers. The fix is as easy as changing constraint to sync on this method and then calling 

`resp.body.compile.drain.as(UnexpectedStatus(resp.status))`

ChristopherDavenport:
Fixes #3707

### DIFF
--- a/client/src/main/scala/org/http4s/client/DefaultClient.scala
+++ b/client/src/main/scala/org/http4s/client/DefaultClient.scala
@@ -262,6 +262,6 @@ private[http4s] abstract class DefaultClient[F[_]](implicit F: Bracket[F, Throwa
   def prepAs[T](req: F[Request[F]])(implicit d: EntityDecoder[F, T]): F[T] =
     fetchAs(req)
 
-  private def defaultOnError(resp: Response[F])(implicit F: Applicative[F]): F[Throwable] =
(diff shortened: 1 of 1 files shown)


ChristopherDavenport:
So we only have Bracket, but we require Sync. We can make this change in master, but I'm not sure we can do this binary compatible. That being said I'm unsure what our goal is here.

This really opens up the question of when we should be running streams or what the implications of leaving streams untouched is in the larger scope.

rossabaker:
We could have a `DefaultClientSync` extension and fix it there.  BlazeClient could use that new definition.  But I think `EmberClient` is the dealbreaker in 0.21, because we return the concrete subtype there.

rossabaker:
Okay, I think we can fix it the right way in 1.0, and maybe in a separate effort fix it for blaze. Looks like it does not compile as is.

armanbilge:
Superseded by https://github.com/http4s/http4s/pull/6376.

PR #3714, opened 2020-09-28
Router should use correct prefix when routing

### LINKED ISSUE #3713: Incorrect path composing with Router
**Problem:**
If the prefix used in `Router` contains more the one path fragment,  `Router` will always return `Not Found` error.

```
val ordersRoutes = HttpRoutes.of {
    case GET -> Root / "orders" => Ok("orders list")
}

val routes = Router("/api/v1/" -> ordersRoutes)
```

**Expected result:**
Router should route request `GET /api/v1/orders ` to correct handler 

**Actual result:**
Router return `Not Found` for request `GET /api/v1/orders `

Works well if the prefix contains only one path fragment:

`val routes = Router("/api/" -> ordersRoutes)`

**Http4s version**: `1.0.0-M4`

hamnis:
Added deprecations to indexOf and indexOfString and added the implementation to findSplit and findSplitOfString

Fixes #3713

### DIFF
--- a/server/src/test/scala/org/http4s/server/RouterSpec.scala
+++ b/server/src/test/scala/org/http4s/server/RouterSpec.scala
@@ -51,13 +51,23 @@ class RouterSpec extends Http4sSpec with Http4sLegacyMatchersIO {
     "/numb" -> middleware(numbers2),
     "/" -> root,
     "/shadow" -> shadow,
-    "/letters" -> letters
+    "/letters" -> letters,
+    "/numbers/v1" -> HttpRoutes.of[IO] {
+      case GET -> Root / "1" => Ok("Yeah")
+    },
+    "/numbers" -> Router[IO](
+      "/v2" -> HttpRoutes.of[IO] {
+        case GET -> Root / "1" => Ok("Indeed")
+      }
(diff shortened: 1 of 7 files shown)


ChristopherDavenport (approved):

server/src/test/scala/org/http4s/server/RouterSpec.scala
```diff
@@ -51,13 +51,23 @@ class RouterSpec extends Http4sSpec with Http4sLegacyMatchersIO {
...
+      }
+    )
   )
 
   "A router" should {
     "translate mount prefixes" in {
       service.orNotFound(Request[IO](GET, uri"/numbers/1")) must returnBody("one")
       service.orNotFound(Request[IO](GET, uri"/numb/1")) must returnBody("two")
+      service.orNotFound(Request[IO](GET, uri"/numbers/v1/1")) must returnBody("Yeah")
+      service.orNotFound(Request[IO](GET, uri"/numbers/v2/1")) must returnBody("Indeed")
```
Daenyth, line 70: 👍

Daenyth (approved):

PR #3716, opened 2020-09-29
Fix Method.fromString

### LINKED ISSUE #3680: org.http4s.Method.fromString may succeed on invalid inputs
The initial conversation is [here](https://gitter.im/http4s/http4s-dev?at=5f55d832ec534f584fea2572)

Consider the example:
```
scala> Method.fromString("FOO:BAR")
val res: org.http4s.ParseResult[org.http4s.Method] = Right(FOO)
```
In other words, although the entire input string `"FOO:BAR"` is not a valid token, but its initial part `"FOO"` gets captured since it is still a token.
Therefore, `fromString` captures everything up to the first non-token symbol but not necessarily to the end of the entire input string.

Expected behavior: `Method.fromString("FOO:BAR")` should fail to create `Method`, i.e. return `Left(...)`.

satorg:
Should resolve #3680
Fixes the `Rfc2616BasicRules.token` method which is used in `Method.fromString`.
The `Rfc2616BasicRules.isToken` method should also get fixed.

### DIFF
--- a/tests/src/test/scala/org/http4s/MethodSpec.scala
+++ b/tests/src/test/scala/org/http4s/MethodSpec.scala
@@ -6,11 +6,14 @@
 
 package org.http4s
 
-import cats.Hash
-import cats.kernel.laws.discipline._
 import java.util.Locale
+
+import cats.Hash
+import cats.implicits._
+import cats.kernel.laws.discipline._
 import org.http4s.parser.Rfc2616BasicRules
-import org.scalacheck.Prop.forAll
+import org.scalacheck.Gen
+import org.scalacheck.Prop.{forAll, forAllNoShrink}
 
 class MethodSpec extends Http4sSpec {
   import Method._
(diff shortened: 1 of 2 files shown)


tests/src/test/scala/org/http4s/MethodSpec.scala
```diff
@@ -26,10 +29,33 @@ class MethodSpec extends Http4sSpec {
 
   "only tokens are valid methods" in {
     prop { (s: String) =>
+      // TODO: this check looks meaningless: `fromString` mostly relies on `Rfc2616BasicRules.token`
+      //       and the result is compared to `isToken` which calls to `Rfc2616BasicRules.token` as well.
```
rossabaker, line 33: Yeah, it's difficult to avoid repeating implementation in some property tests.  These kind of properties guard against future optimizations that break the contract, but that's unlikely here.

rossabaker (approved):
Looks great.

PR #3726, opened 2020-10-07
add proxy support column

amesgen:
Nuances like Proxy authentication are probably not in scope for an overview table.


rossabaker (approved):
:+1:, but left organizational comment on the original ticket.

PR #3735, opened 2020-10-10
Gzip in webjar service

ashwinbhaskar:


server/src/main/scala/org/http4s/server/staticcontent/WebjarService.scala
```diff
@@ -21,21 +21,29 @@ import scala.util.control.NoStackTrace
...
   * @param classLoader optional classloader for extracting the resources
+  * @param preferGzipped prefer gzip compression format?
   */
 class WebjarServiceBuilder[F[_]] private (
     blocker: Blocker,
     webjarAssetFilter: WebjarServiceBuilder.WebjarAssetFilter,
     cacheStrategy: CacheStrategy[F],
-    classLoader: Option[ClassLoader]) {
+    classLoader: Option[ClassLoader],
+    preferGzipped: Boolean) {
```
rossabaker, line 31: If we added a secondary constructor here, this could be binary compatible and added to 0.21. What do you think?

rossabaker (commented):
Nice!

server/src/main/scala/org/http4s/server/staticcontent/WebjarService.scala
ashwinbhaskar: series/21 branch does not have `WebjarServiceBuilder`. It just has a `WebjarService` object - https://github.com/http4s/http4s/blob/series/0.21/server/src/main/scala/org/http4s/server/staticcontent/WebjarService.scala

rossabaker: Oops, bad memory/assumption on my part.

rossabaker (approved):

PR #3738, opened 2020-10-11
resolving PrometheusExportService path comparison failure

### LINKED ISSUE #3731: Prometheus metrics route never matches in 1.0.0-M4
In `1.0.0-M4` and `master` at this time, the `pathInfo` field of `Request` was changed to a `Uri`, however the Prometheus `/metrics` route still does an equality check against a string which simply never passes.

https://github.com/http4s/http4s/blob/v1.0.0-M4/prometheus-metrics/src/main/scala/org/http4s/metrics/prometheus/PrometheusExportService.scala#L49

jyoo980:
Should be work that resolves #3731

### DIFF
--- a/prometheus-metrics/src/main/scala/org/http4s/metrics/prometheus/PrometheusExportService.scala
+++ b/prometheus-metrics/src/main/scala/org/http4s/metrics/prometheus/PrometheusExportService.scala
@@ -12,7 +12,10 @@ import io.prometheus.client.CollectorRegistry
 import io.prometheus.client.exporter.common.TextFormat
 import io.prometheus.client.hotspot._
 import java.io.StringWriter
+
+import org.http4s.Uri.Path
 import org.http4s._
+import org.http4s.implicits.http4sLiteralsSyntax
 
 /*
  * PrometheusExportService Contains an HttpService
@@ -26,6 +29,9 @@ final class PrometheusExportService[F[_]: Sync] private (
 )
 
 object PrometheusExportService {
+
+  private val metricsPath: Path = path"/metrics"
+
   def apply[F[_]: Sync](collectorRegistry: CollectorRegistry): PrometheusExportService[F] =
     new PrometheusExportService(service(collectorRegistry), collectorRegistry)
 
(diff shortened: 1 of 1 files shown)


prometheus-metrics/src/main/scala/org/http4s/metrics/prometheus/PrometheusExportService.scala
```diff
@@ -46,7 +46,7 @@ object PrometheusExportService {
 
   def service[F[_]: Sync](collectorRegistry: CollectorRegistry): HttpRoutes[F] =
     HttpRoutes.of[F] {
-      case req if req.method == Method.GET && req.pathInfo == "/metrics" =>
+      case req if req.method == Method.GET && req.pathInfo.toString == "/metrics" =>
```
jyoo980, line 49: Alternatively, we could do
```Scala
req.pathInfo == path"/metrics"
```
which is something I've seen used in the codebase before.

rossabaker: I like the `path` version, and we could store that to a constant in the service.

And if we used cats' `===`, the bug wouldn't have compiled.

rossabaker:
Thanks! When this is ready, we'll cherrypick it back to series/0.21.

prometheus-metrics/src/main/scala/org/http4s/metrics/prometheus/PrometheusExportService.scala
jyoo980: Done!

jyoo980:
Looks like this was the only place where we were doing a raw string comparison for the path; other usages had the correct comparison to a `Path`

rossabaker (approved):
Thanks!

hamnis (approved):
Apparently I missed a spot.
The universal equality things are sometimes hard to find.

Good find. thanks for the fix

hamnis:
> Thanks! When this is ready, we'll cherrypick it back to series/0.21.

Why? this is only a problem for 1.0 and master

rossabaker:
You're right.  I was looking for binary compatibility without thinking.

PR #3740, opened 2020-10-12
scalafmt-2.7.4 applied to series/0.21

rossabaker:
This is the churn in defaults, and it's probably going to be a miserable merge.  But if we're going to do it, it starts here.


rossabaker:
The longer we dawdle on this, the worse it's going to be. I'm doing it and I'll deal with the merge hell.

PR #3741, opened 2020-10-12
Fix 'unused' warnings in scalafix-input module

### LINKED ISSUE #3613: Suppress compile time warnings in the scalafix module
Warnings like these below are important in general, but make no sense in the `scalafix` module (or at least in the given file), since these unused parameters were introduced intentionally to let the output match the input after applying the scalafix rule to the latter.
```
[warn] /.../http4s/scalafix/input/src/main/scala/fix/ClientFetchTests.scala:15:78: parameter value G in method doFetchR1 is never used
[warn]     final def doFetchR1[A](req: Request[G], f: Response[G] => G[A])(implicit G: Sync[G]): G[A] =
[warn]                                                                              ^
[warn] /.../http4s/scalafix/input/src/main/scala/fix/ClientFetchTests.scala:18:82: parameter value G in method doFetchFR1 is never used
[warn]     final def doFetchFR1[A](req: G[Request[G]], f: Response[G] => G[A])(implicit G: Sync[G]): G[A] =
[warn]                                                                                  ^
[warn] /.../http4s/scalafix/input/src/main/scala/fix/ClientFetchTests.scala:21:78: parameter value G in method doFetchR2 is never used
[warn]     final def doFetchR2[B](req: Request[G], f: Response[G] => G[B])(implicit G: Sync[G]): G[B] =
```

satorg:
Resolves #3613

Suppresses all "unused-whatever" warnings in the `scalafixInput` module.
Sometimes `scalafixInput` have to contain unused stuff (imports, parameters, etc) just to let the `scalafixOutput` module be semantically correct and thus compile successfully.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -614,6 +614,7 @@ lazy val scalafixInput = project
     // TODO: I think these are false positives
     unusedCompileDependenciesFilter -= moduleFilter(organization = "org.http4s"),
     scalacOptions -= "-Xfatal-warnings",
+    scalacOptions ~= { _.filterNot(_.startsWith("-Wunused:")) }
   )
   // Syntax matters as much as semantics here.
   .disablePlugins(HeaderPlugin, ScalafmtPlugin)


build.sbt
```diff
@@ -614,6 +614,7 @@ lazy val scalafixInput = project
     // TODO: I think these are false positives
     unusedCompileDependenciesFilter -= moduleFilter(organization = "org.http4s"),
     scalacOptions -= "-Xfatal-warnings",
+    scalacOptions ~= { _.filterNot(_.startsWith("-Wunused:")) }
```
satorg, line 617: Removes the following options:
```
-Wunused:imports  
-Wunused:patvars  
-Wunused:privates 
-Wunused:locals   
-Wunused:explicits
-Wunused:implicits
-Wunused:nowarn   
-Wunused:params   
-Wunused:linted   
```
These options are defined for Scala 2.13 only.

It turned out that it is not necessary to disable the corresponding options for Scala 2.12 since cross build is not set neither for `scalafixInput` nor for `scalafixOutput` modules.

rossabaker (approved):
:+1:

PR #3752, opened 2020-10-19
Ember flaky test fix

ashwinbhaskar:


core/src/main/scala/org/http4s/Uri.scala
```diff
@@ -124,14 +124,19 @@ final case class Uri(
...
 
       case Uri(None, None, _, _, _) =>
     }
 
     this match {
       case Uri(_, Some(_), p, _, _) if p.nonEmpty && !p.startsWith("/") =>
         writer << "/" << p
+      case Uri(None, None, p, _, _) =>
+        if (p.contains(":"))
+          writer << "./" << p // https://tools.ietf.org/html/rfc3986#section-4.2 last paragraph
```
rossabaker, line 137: I've probably read that RFC a dozen times and never noticed this rule. Great find.

rossabaker (approved):
Very nice work! Thanks for finding the root cause.

ChristopherDavenport (approved):
Awesome, this one really bugged me!

PR #3757, opened 2020-10-20
sbt-1.4.1 and friends

rossabaker:
Pulls in a compatible sbt-explicit-dependencies that's blocking #3755.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -85,6 +85,7 @@ lazy val core = libraryProject("core")
       scalaReflect(scalaVersion.value) % Provided,
       vault,
     ),
+    unusedCompileDependenciesFilter -= moduleFilter("org.scala-lang", "scala-reflect"),
   )
 
 lazy val laws = libraryProject("laws")
--- a/build.sbt
+++ b/build.sbt
@@ -433,6 +433,7 @@ lazy val docs = http4sProject("docs")
     ScalaUnidocPlugin,
     MdocPlugin
   )
+  .settings(docsProjectSettings)
   .settings(
     crossScalaVersions := List(scala_212),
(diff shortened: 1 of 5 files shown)


rossabaker:
The stricter MiMa rules are haunting us.

dsl/src/main/scala/org/http4s/dsl/Http4sDsl.scala
```diff
@@ -36,6 +36,16 @@ trait Http4sDsl2[F[_], G[_]] extends RequestDsl with Statuses with Responses[F,
...
+
+  // Restore binary compatibility with mixin forwarder
+  // TODO remove in 1.0
+  override implicit def http4sMethodSyntax(method: Method): Http4sDsl.MethodOps =
+    new Http4sDsl.MethodOps(method)
+
+  // Restore binary compatibility with mixin forwarder
+  // TODO remove in 1.0
+  override implicit def http4sMethodConcatSyntax(methods: MethodConcat): Http4sDsl.MethodConcatOps =
+    new Http4sDsl.MethodConcatOps(methods)
```
rossabaker, line 48: I'm not fully understanding why this works, but it appeased MiMa?

build.sbt
```diff
@@ -85,6 +85,7 @@ lazy val core = libraryProject("core")
       scalaReflect(scalaVersion.value) % Provided,
       vault,
     ),
+    unusedCompileDependenciesFilter -= moduleFilter("org.scala-lang", "scala-reflect"),
```
rossabaker, line 88: This was only an issue on Scala 2.12. It very certainly is used! I don't know what's going on here.

aeons (approved):

PR #3758, opened 2020-10-20
Refactor query param infix operators to avoid 2.13 error

### LINKED ISSUE #3719: Warning for future deprecation in Scala 2.13.3
## What's the problem?

When using [`+?` operator](https://github.com/http4s/http4s/blob/e45ca2c8703c43d7c00a55d47a35d3bd3b79dfdd/core/src/main/scala/org/http4s/QueryOps.scala#L39) to add a query param to an `Uri` value in Scala 2.13.3 the following warning is raised:

```scala
[warn] ${BASE}/Test.scala:14:37: multiarg infix syntax looks like a tuple and will be deprecated
[warn]   val uri = uri"localhost" / "test" +? ("key", "value")
[warn]                                     ^
```

The warning only dissapears if no-infix syntax is used:

```scala
(uri"localhost" / "test").+?("key", "value")
```

...which completely breaks the goal of those operators.

## Possible solution

Changing the [`+?` operator](https://github.com/http4s/http4s/blob/e45ca2c8703c43d7c00a55d47a35d3bd3b79dfdd/core/src/main/scala/org/http4s/QueryOps.scala#L39) to directly receive a tuple. It would break back-compatibility but it shouldn't break many codebases since there is no difference between the two params infix syntax and the one param tuple one.

```scala
def +?[K: QueryParamKeyLike, T: QueryParamEncoder](param: (K, T)): Self =
  +?(param._1, param._2 :: Nil)
```

> For anyone that bumps into this error and is using `-Xfatal-warnings` (or just wants to get rid of the warning) you can suppress it by adding the following to your `build.sbt`:
> 
> ```sbt
> scalacOptions += "-Wconf:cat=lint-multiarg-infix:silent"
> ```

## How to test?

Just create a simple project with these two files:

```scala
//build.sbt

scalaVersion := "2.13.3"
libraryDependencies += "org.http4s" %% "http4s-core" % "0.21.7"
```

```scala
//Test.scala

import org.http4s._
import org.http4s.syntax.literals._

object Test {

  val uri = uri"localhost" / "test" +? ("key", "value")

}
```

alejandrohdezma:
## What's the problem?

When using [`+?` operator](https://github.com/http4s/http4s/blob/e45ca2c8703c43d7c00a55d47a35d3bd3b79dfdd/core/src/main/scala/org/http4s/QueryOps.scala#L39) to add a query param to an `Uri` value in Scala 2.13.3 the following warning is raised:

```scala
[warn] ${BASE}/Test.scala:14:37: multiarg infix syntax looks like a tuple and will be deprecated
[warn]   val uri = uri"localhost" / "test" +? ("key", "value")
[warn]                                     ^
```

The warning only dissapears if no-infix syntax is used:

```scala
(uri"localhost" / "test").+?("key", "value")
```

...which completely breaks the goal of those operators.

## Changes included in this PR

- Separate usages of `+?` and `+??` in tests to be able to see the error (if someone tries to compile the first commit of this PR it should fail with `multiarg infix syntax looks like a tuple and will be deprecated`.
- Since `+?` has two two-param version methods, we need to rename `+?(T, Seq)` to `++?` to avoid clashing between them once they're changed to their "tuple" version.
- Change `+?`, `++?` and `+??` two-param operators to receive a tuple.

## Referenced PR

This fixes #3719

### DIFF
--- a/core/src/main/scala/org/http4s/QueryOps.scala
+++ b/core/src/main/scala/org/http4s/QueryOps.scala
@@ -47,7 +47,7 @@ trait QueryOps {
   def ++?[K: QueryParamKeyLike, T: QueryParamEncoder](param: (K, collection.Seq[T])): Self =
     _withQueryParam(
       QueryParamKeyLike[K].getKey(param._1),
-      param._2.map((t: T) => QueryParamEncoder[T].encode(t)))
+      param._2.map(QueryParamEncoder[T].encode))
 
   /** alias for withOptionQueryParam */
   def +??[K: QueryParamKeyLike, T: QueryParamEncoder](param: (K, Option[T])): Self =
--- a/core/src/main/scala/org/http4s/QueryOps.scala
+++ b/core/src/main/scala/org/http4s/QueryOps.scala
@@ -36,20 +36,24 @@ trait QueryOps {
     _withQueryParam(QueryParam[T].key, values.map(QueryParamEncoder[T].encode))
 
   /** alias for withQueryParam */
-  def +?[K: QueryParamKeyLike, T: QueryParamEncoder](name: K, value: T): Self =
-    ++?(name, value :: Nil)
+  def +?[K: QueryParamKeyLike, T: QueryParamEncoder](param: (K, T)): Self =
+    ++?((param._1, param._2 :: Nil))
 
   /** alias for withQueryParam */
   def +?[K: QueryParamKeyLike](name: K): Self =
     _withQueryParam(QueryParamKeyLike[K].getKey(name), Nil)
 
   /** alias for withQueryParam */
-  def ++?[K: QueryParamKeyLike, T: QueryParamEncoder](name: K, values: collection.Seq[T]): Self =
-    _withQueryParam(QueryParamKeyLike[K].getKey(name), values.map(QueryParamEncoder[T].encode))
+  def ++?[K: QueryParamKeyLike, T: QueryParamEncoder](param: (K, collection.Seq[T])): Self =
+    _withQueryParam(
+      QueryParamKeyLike[K].getKey(param._1),
+      param._2.map((t: T) => QueryParamEncoder[T].encode(t)))
 
   /** alias for withOptionQueryParam */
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/http4s/QueryOps.scala
```diff
@@ -36,20 +36,24 @@ trait QueryOps {
...
   def +?[K: QueryParamKeyLike](name: K): Self =
     _withQueryParam(QueryParamKeyLike[K].getKey(name), Nil)
 
   /** alias for withQueryParam */
-  def +?[K: QueryParamKeyLike, T: QueryParamEncoder](name: K, values: collection.Seq[T]): Self =
-    _withQueryParam(QueryParamKeyLike[K].getKey(name), values.map(QueryParamEncoder[T].encode))
+  def ++?[K: QueryParamKeyLike, T: QueryParamEncoder](param: (K, collection.Seq[T])): Self =
+    _withQueryParam(
+      QueryParamKeyLike[K].getKey(param._1),
+      param._2.map((t: T) => QueryParamEncoder[T].encode(t)))
```
satorg, line 50: This is a minutiae, but in my opinion it would be better to keep the same style between this method and the next one. I mean
```
param._2.map((t: T) => QueryParamEncoder[T].encode(t)))
```
versus
```
param._2.map(QueryParamEncoder[T].encode))
```

satorg (approved):
Looks good. I'm not that certain about introducing the new `++?` operator though, but it looks as a reasonable change to me.

core/src/main/scala/org/http4s/QueryOps.scala
alejandrohdezma: Done! https://github.com/http4s/http4s/pull/3758/commits/f92cac396bace11dd523ca3540940c405d6aab0f

alejandrohdezma:
> Looks good. I'm not that certain about introducing the new `++?` operator though, but it looks as a reasonable change to me.

Yeah, me neither, but I couldn't find other alternative given the clash after erasure 😞

rossabaker (approved):
I'm not a huge fan of the operators, but since we have them, I think the one you chose is an intuitive choice. And I love the writeup.

PR #3767, opened 2020-10-22
Clean up unbound placeholders

rossabaker:
Dotty doesn't like these.


satorg (approved):
Interesting.. I though that it should be reasonable to avoid introducing some name in a scope if there's no intent to use it... Anyway, good to know, thanks.

rossabaker:
Yeah, I am a bit disappointed that this is necessary.  And I think there are a few more of these beyond core.  These are just the ones failing to compile among all the other compile failures at the moment.

Issue #3764, opened 2020-10-21
Silencer and Dotty


rossabaker:
Silencer won't be supported by Dotty.  Figure out how to silence warnings as we cross-compile, as well as conditionally add the Silencer plugin.  

Using it is already repetitious enough that we embedded it in sbt-http4s-org, but it's tricky to reference `isDotty.value` there.

aeons:
`scala.annotation.nowarn` is only 2.13, right?

rossabaker:
Yes. scala-collection-compat includes the new annotation, so maybe if we changed the silencer annotations to that, we'd be in good shape: https://contributors.scala-lang.org/t/configurable-warnings-warning-suppression/4096/15

rossabaker:
We rolled our own collection compat traits so far.  I can't remember why we have avoided the official compat library to this point: was it not wanting the weight of an extra dependency, or was there something deeper?  It's also something worth revisiting.

Daenyth:
Silencer in combination with collections-compat allows you to use `nowarn` on 2.12 - I do this in my codebases

rossabaker:
@Daenyth Great!  In that case, I think we can adopt that strategy and target the series/0.21 branch with it.  The more we can do there, the less we'll diverge as the more difficult Dotty work is done.

PR #3770, opened 2020-10-22
Replace silent annotation with nowarn

rossabaker:
It's the basic idea @Daenyth suggested on #3764.  I'm still avoiding the (nice) scala-collection-compat library for now, just to avoid taking on more heft.

Paired with https://github.com/http4s/sbt-http4s-org/pull/33, we should be able to re-enable the `SilencerPlugin` on the dotty branch.

### DIFF
--- a/client/src/main/scala/org/http4s/client/middleware/GZip.scala
+++ b/client/src/main/scala/org/http4s/client/middleware/GZip.scala
@@ -40,7 +40,7 @@ object GZip {
           req.headers ++ Headers.of(Header(`Accept-Encoding`.name.value, supportedCompressions)))
     }
 
-  @nowarn("msg=deprecated")
+  @nowarn("cat=deprecation")
   private def decompress[F[_]](bufferSize: Int, response: Response[F])(implicit
       F: Bracket[F, Throwable]): Response[F] =
     response.headers.get(`Content-Encoding`) match {
--- a/client/src/main/scala/org/http4s/client/middleware/GZip.scala
+++ b/client/src/main/scala/org/http4s/client/middleware/GZip.scala
@@ -9,9 +9,9 @@ package client
 package middleware
 
 import cats.effect.Bracket
-import com.github.ghik.silencer.silent
 import fs2.{Pipe, Pull, Stream}
 import org.http4s.headers.{`Accept-Encoding`, `Content-Encoding`}
+import scala.annotation.nowarn
 import scala.util.control.NoStackTrace
 
 /** Client middleware for enabling gzip.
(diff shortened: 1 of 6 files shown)


client/src/main/scala/org/http4s/client/middleware/GZip.scala
```diff
@@ -40,7 +40,7 @@ object GZip {
           req.headers ++ Headers.of(Header(`Accept-Encoding`.name.value, supportedCompressions)))
     }
 
-  @silent("deprecated")
+  @nowarn("msg=deprecated")
```
satorg, line 43: I personally found that `@nowarn("cat=deprecation")` looks clearer since it disables warnings by the entire category rather than trying to figure it out from warning messages. Either way should work though, just a suggestion.

satorg: ...but I haven't tried it on Scala other than 2.13 though...

rossabaker: I'm learning about this on the fly, but I was uncomfortable with what I had to do.  Let's try that.

satorg (approved):

aeons (approved):

client/src/main/scala/org/http4s/client/middleware/GZip.scala
Daenyth: Silencer on 2.12 only supports `msg`, nothing else

satorg: Yeah, indeed. According to their docs:
> NOTE: @nowarn in Scala 2.13.2 supports various fine-grained filters (e.g. warning category, message patttern, etc.). Silencer only supports the msg=<pattern> filter - all other filters simply suppress everything, as if there were no filters specified.

But perhaps it is not that bad in our case – looks like the `cat` filter works properly in Scala 2.13 and any meaningful warning is going to be caught up anyway.

rossabaker: The 2.12 builds compiled without a fatal warning.  I think the silencer integration still did work?

satorg: Looks like if the silencer cannot recognize a filter in `@nowarn`, it will simply suppress any kind of warnings for the annotated item.

rossabaker: Oh, I see.  I think we can live with that behavior: we're talking about useful warnings that are found by 2.12, but not by 2.13 or soon Dotty, from places that already have a deprecation warning that we're ignoring.  That seems exceedingly unlikely.  The warnings are generally going to be more insightful on the newer compilers anyway.

PR #3771, opened 2020-10-22
Change default branch to main

rossabaker:
Updates some various references in comments and docs.  I already created the branch and reset all open PRs.


rossabaker:
There are five live branches right now and this gets it to four so I'm going to aggressively merge.  Ex post facto commentary still welcome.

PR #3773, opened 2020-10-22
Parboiled2 -> Parsley

rossabaker:


rossabaker:
Can somebody tell me why this source dependency is trying to resolve a POM?  We could publish a snapshot, but I just wanted to start sketching this out and seeing what help could be summoned, and this seemed the easiest way.

rossabaker:
I'm working on publishing so we can drop the source dependency.

PR #3783, opened 2020-10-22
Adjusts .gitignore for newer SBT version

satorg:
Adds  `/.bsp/` dir (which was introduced in SBT 1.4.0) to `.gitignore`


rossabaker (approved):
I like to globally ignore it.  My `~/.config/git/ignore`:

```
.bloop
.bsp/
.direnv/
.metals
metals.sbt
```

But I'm cool with adding it here.

PR #3784, opened 2020-10-22
Upgrade core to cats-effect-3

domaspoliakas:
Hello.

Took me a little bit of time to get my bearings with regards to figuring out the right artifacts to import and such. For now this is a pretty simple set of changes just to make everything compile.

- I got rid of `AsyncSyntax`; I don't think it made any sense to keep it. 
- Removed all mention of `Blocker` and `ContextShift` (naturally)
- There's a couple of instances where unsafeRunSync was done where I used the global IORuntime. I'm not sure if that's the right thing to do. If that could be avoided core could probably import just `kernel` and `std`.
- I don't know if I used the dispatcher correctly. That is, I added `Dispatcher` as an implicit in a couple of places, which I'm concerned about since it might have a knock-on effect and I don't know if that's desired.

I was primarily focused on getting this to compile and trying to understand how to migrate without changes, but now that it compiles I think that some places where `Sync` is used could be loosened due to the presence of `Files`.

Anyway, let me know how I could improve this. I will try to run tests tomorrow.


core/src/main/scala/org/http4s/EntityDecoder.scala
```diff
@@ -224,19 +225,15 @@ object EntityDecoder {
...
     EntityDecoder.decodeBy(MediaRange.`*/*`) { msg =>
-      val pipe = writeAll[F](file.toPath, blocker)
+      val pipe = Files[F].writeAll(file.toPath)
       DecodeResult.success(msg.body.through(pipe).compile.drain).map(_ => file)
     }
 
-  def textFile[F[_]](file: File, blocker: Blocker)(implicit
-      F: Sync[F],
-      cs: ContextShift[F]): EntityDecoder[F, File] =
+  def textFile[F[_]: Files](file: File)(implicit F: Sync[F]): EntityDecoder[F, File] =
```
SystemFw, line 234: why `Sync` here? Also if `Sync` is unavoidable, you can remove `Files` from there

core/src/main/scala/org/http4s/Message.scala
```diff
@@ -186,6 +186,7 @@ sealed trait Message[F[_]] extends Media[F] { self =>
 object Message {
   private[http4s] val logger = getLogger
   object Keys {
+    import cats.effect.unsafe.implicits.global
```
SystemFw, line 189: we can avoid this by using `SyncIO` for this

core/src/main/scala/org/http4s/Message.scala
```diff
@@ -495,6 +496,7 @@ object Request {
   final case class Connection(local: InetSocketAddress, remote: InetSocketAddress, secure: Boolean)
 
   object Keys {
+    import cats.effect.unsafe.implicits.global
```
SystemFw, line 499: same

core/src/main/scala/org/http4s/StaticFile.scala
```diff
@@ -25,15 +26,13 @@ object StaticFile {
 
   val DefaultBufferSize = 10240
 
-  def fromString[F[_]: Sync: ContextShift](
+  def fromString[F[_]: Files: Sync](
```
SystemFw, line 29: you don't want `Files: Sync`. It it works with just `Files` plus non effect capturing classes (e.g. `Monad` or `Concurrent`), then great, if not `Sync` is enough.

core/src/main/scala/org/http4s/StaticFile.scala
```diff
@@ -116,37 +111,31 @@ object StaticFile {
       Sync[F].delay(
         if (f.isFile) s"${f.lastModified().toHexString}-${f.length().toHexString}" else "")
 
-  def fromFile[F[_]: Sync: ContextShift](
+  def fromFile[F[_]: Files: Sync](
```
SystemFw, line 114: same

core/src/main/scala/org/http4s/StaticFile.scala
```diff
@@ -116,37 +111,31 @@ object StaticFile {
...
       f: File,
-      blocker: Blocker,
       req: Option[Request[F]],
       etagCalculator: File => F[String]): OptionT[F, Response[F]] =
-    fromFile(f, DefaultBufferSize, blocker, req, etagCalculator)
+    fromFile(f, DefaultBufferSize, req, etagCalculator)
 
-  def fromFile[F[_]: Sync: ContextShift](
+  def fromFile[F[_]: Files: Sync](
```
SystemFw, line 125: same

core/src/main/scala/org/http4s/StaticFile.scala
```diff
@@ -116,37 +111,31 @@ object StaticFile {
...
       f: File,
       buffsize: Int,
-      blocker: Blocker,
       req: Option[Request[F]],
       etagCalculator: File => F[String]): OptionT[F, Response[F]] =
-    fromFile(f, 0, f.length(), buffsize, blocker, req, etagCalculator)
+    fromFile(f, 0, f.length(), buffsize, req, etagCalculator)
 
-  def fromFile[F[_]](
+  def fromFile[F[_]: Files](
```
SystemFw, line 132: same

core/src/main/scala/org/http4s/StaticFile.scala
```diff
@@ -213,19 +202,15 @@ object StaticFile {
...
 
   private def nameToContentType(name: String): Option[`Content-Type`] =
     name.lastIndexOf('.') match {
       case -1 => None
       case i => MediaType.forExtension(name.substring(i + 1)).map(`Content-Type`(_))
     }
 
-  private[http4s] val staticFileKey = Key.newKey[IO, File].unsafeRunSync()
+  private[http4s] val staticFileKey =
+    Key.newKey[IO, File].unsafeRunSync()(cats.effect.unsafe.implicits.global)
```
SystemFw, line 215: `SyncIO`, so no `global` is needed

core/src/main/scala/org/http4s/internal/package.scala
```diff
@@ -14,37 +14,40 @@ import java.util.concurrent.{
...
   // Like fs2.async.unsafeRunAsync before 1.0.  Convenient for when we
   // have an ExecutionContext but not a Timer.
-  private[http4s] def unsafeRunAsync[F[_], A](fa: F[A])(
-      f: Either[Throwable, A] => IO[Unit])(implicit F: Effect[F], ec: ExecutionContext): Unit =
-    F.runAsync(Async.shift(ec) *> fa)(f).unsafeRunSync()
+  private[http4s] def unsafeRunAsync[F[_]: Async, A](fa: F[A])(
+      f: Either[Throwable, A] => F[Unit])(implicit
+      dispatcher: Dispatcher[F],
+      ec: ExecutionContext): Unit =
+    dispatcher.unsafeRunSync(fa.evalOn(ec).attemptTap(f).void)
```
SystemFw, line 36: you made this synchronous, your probably want to match the semantics of the existing code instead, also I'd be curious to see where this is used

core/src/main/scala/org/http4s/internal/Logger.scala
```diff
@@ -6,7 +6,7 @@
 
 package org.http4s.internal
 
-import cats.effect.Sync
+import cats.effect.kernel.Sync
```
SystemFw, line 9: not sure if we want kernel imports or not, they are re-exported in `core`, but there has been some recent activity. @djspiewak can clarify I think

core/src/main/scala/org/http4s/internal/package.scala
```diff
@@ -14,37 +14,40 @@ import java.util.concurrent.{
...
+      case Right(_) => F.unit
     }
 
   // Inspired by https://github.com/functional-streams-for-scala/fs2/blob/14d20f6f259d04df410dc3b1046bc843a19d73e5/io/src/main/scala/fs2/io/io.scala#L140-L141
-  private[http4s] def invokeCallback[F[_]](logger: Logger)(f: => Unit)(implicit
-      F: ConcurrentEffect[F]): Unit =
-    F.runAsync(F.start(F.delay(f)).flatMap(_.join))(loggingAsyncCallback(logger)).unsafeRunSync()
+  private[http4s] def invokeCallback[F[_]](logger: Logger)(
+      f: => Unit)(implicit F: Async[F], dispatcher: Dispatcher[F]): Unit =
+    dispatcher.unsafeRunSync(
```
SystemFw, line 48: you changed the semantics here, by using `unsafeRunSync`

core/src/main/scala/org/http4s/internal/package.scala
```diff
@@ -14,37 +14,40 @@ import java.util.concurrent.{
...
+      F.start(F.delay(f)).flatMap(_.join).attemptTap(loggingAsyncCallback(logger)(_)(F)).void
+    )
```
SystemFw, line 50: This whole thing is not needed anymore I don't think, it fixes a problem CE3 does not have

core/src/main/scala/org/http4s/internal/package.scala
```diff
@@ -123,68 +126,52 @@ package object internal {
...
   private[http4s] def fromCompletionStage[F[_], CF[x] <: CompletionStage[x], A](
       fcs: F[CF[A]])(implicit
       // Concurrent is intentional, see https://github.com/http4s/http4s/pull/3255#discussion_r395719880
-      F: Concurrent[F],
-      CS: ContextShift[F]): F[A] =
+      F: Async[F]): F[A] =
     fcs.flatMap { cs =>
       F.async[A] { cb =>
-        cs.handle[Unit] { (result, err) =>
+        F.delay(cs.handle[Unit] { (result, err) =>
```
SystemFw, line 156: you can use `async_` and leave the rest unchanged

core/src/main/scala/org/http4s/internal/package.scala
```diff
@@ -123,68 +126,52 @@ package object internal {
...
   private[http4s] def unsafeToCompletionStage[F[_], A](
       fa: F[A]
-  )(implicit F: Effect[F]): CompletionStage[A] = {
+  )(implicit dispatcher: Dispatcher[F], F: Sync[F]): CompletionStage[A] = {
     val cf = new CompletableFuture[A]()
-    F.runAsync(fa) {
-      case Right(a) => IO { cf.complete(a); () }
-      case Left(e) => IO { cf.completeExceptionally(e); () }
-    }.unsafeRunSync()
+    dispatcher.unsafeRunSync(fa.attemptTap {
```
SystemFw, line 171: use `unsafeRunAsync`. I need to think about threading the `Dispatcher`

core/src/main/scala/org/http4s/multipart/MultipartDecoder.scala
```diff
@@ -7,9 +7,11 @@
 package org.http4s
 package multipart
 
-import cats.effect.{Blocker, ContextShift, Sync}
+import cats.effect.kernel.Sync
```
SystemFw, line 10: do we want `kernel`?

core/src/main/scala/org/http4s/multipart/MultipartDecoder.scala
```diff
@@ -59,8 +61,7 @@ private[http4s] object MultipartDecoder {
     * @return A multipart/form-data encoded vector of parts with some part bodies held in
     *         temporary files.
     */
-  def mixedMultipart[F[_]: Sync: ContextShift](
-      blocker: Blocker,
+  def mixedMultipart[F[_]: Sync: Files](
```
SystemFw, line 64: remove `Sync` if possible, if not remove `Files`

core/src/main/scala/org/http4s/multipart/MultipartParser.scala
```diff
@@ -7,11 +7,11 @@
 package org.http4s
 package multipart
 
-import cats.effect.{Blocker, ContextShift, Sync}
+import cats.effect.kernel.Sync
```
SystemFw, line 10: same

core/src/main/scala/org/http4s/multipart/MultipartParser.scala
```diff
@@ -536,67 +536,43 @@ object MultipartParser {
   /** Same as the other streamed parsing, except
     * after a particular size, it buffers on a File.
     */
-  def parseStreamedFile[F[_]: Sync: ContextShift](
+  def parseStreamedFile[F[_]: Sync: Files](
```
SystemFw, line 539: same

core/src/main/scala/org/http4s/multipart/MultipartParser.scala
```diff
@@ -536,67 +536,43 @@ object MultipartParser {
...
-      maxParts,
-      failOnLimit,
-      blocker)
+    ignorePreludeFileStream[F](boundary, st, limit, maxSizeBeforeWrite, maxParts, failOnLimit)
       .fold(Vector.empty[Part[F]])(_ :+ _)
       .map(Multipart(_, boundary))
   }
 
-  def parseToPartsStreamedFile[F[_]: Sync: ContextShift](
+  def parseToPartsStreamedFile[F[_]: Sync: Files](
```
SystemFw, line 550: same

core/src/main/scala/org/http4s/multipart/MultipartParser.scala
```diff
@@ -536,67 +536,43 @@ object MultipartParser {
...
+    ignorePreludeFileStream[F](boundary, st, limit, maxSizeBeforeWrite, maxParts, failOnLimit)
   }
 
   /** The first part of our streaming stages:
     *
     * Ignore the prelude and remove the first boundary. Only traverses until the first
     * part
     */
-  private[this] def ignorePreludeFileStream[F[_]: Sync: ContextShift](
+  private[this] def ignorePreludeFileStream[F[_]: Sync: Files](
```
SystemFw, line 564: same

core/src/main/scala/org/http4s/multipart/MultipartParser.scala
```diff
@@ -621,15 +597,13 @@ object MultipartParser {
     * @tparam F
     * @return
     */
-  private def pullPartsFileStream[F[_]: Sync: ContextShift](
+  private def pullPartsFileStream[F[_]: Sync: Files](
```
SystemFw, line 600: same

core/src/main/scala/org/http4s/multipart/MultipartParser.scala
```diff
@@ -667,14 +639,14 @@ object MultipartParser {
...
-    F.delay(Files.delete(path))
+    F.delay(NioFiles.delete(path))
       .handleErrorWith { err =>
         logger.error(err)("Caught error during file cleanup for multipart")
         //Swallow and report io exceptions in case
         F.unit
       }
 
-  private[this] def tailrecPartsFileStream[F[_]: Sync: ContextShift](
+  private[this] def tailrecPartsFileStream[F[_]: Sync: Files](
```
SystemFw, line 649: same

core/src/main/scala/org/http4s/multipart/MultipartParser.scala
```diff
@@ -728,17 +698,16 @@ object MultipartParser {
...
     path match {
-      case Some(p) => Part(hdrs, body.onFinalizeWeak(F.delay(Files.delete(p))))
+      case Some(p) => Part(hdrs, body.onFinalizeWeak(F.delay(NioFiles.delete(p))))
       case None => Part(hdrs, body)
     }
 
   /** Split the stream on `values`, but when
     */
-  private def splitWithFileStream[F[_]](
+  private def splitWithFileStream[F[_]: Files](
```
SystemFw, line 707: same

core/src/main/scala/org/http4s/multipart/Part.scala
```diff
@@ -7,10 +7,10 @@
 package org.http4s
 package multipart
 
-import cats.effect.{Blocker, ContextShift, Sync}
+import cats.effect.kernel.Sync
```
SystemFw, line 10: same

core/src/main/scala/org/http4s/multipart/Part.scala
```diff
@@ -41,19 +41,11 @@ object Part {
       Headers(`Content-Disposition`("form-data", Map("name" -> name)) :: headers.toList),
       Stream.emit(value).through(utf8Encode))
 
-  def fileData[F[_]: Sync: ContextShift](
-      name: String,
-      file: File,
-      blocker: Blocker,
-      headers: Header*): Part[F] =
-    fileData(name, file.getName, readAll[F](file.toPath, blocker, ChunkSize), headers: _*)
+  def fileData[F[_]: Sync: Files](name: String, file: File, headers: Header*): Part[F] =
```
SystemFw, line 44: same

core/src/main/scala/org/http4s/syntax/KleisliSyntax.scala
```diff
@@ -9,7 +9,7 @@ package syntax
 
 import cats.{Functor, ~>}
 import cats.syntax.functor._
-import cats.effect.Sync
+import cats.effect.kernel.Sync
```
SystemFw, line 12: same

SystemFw:
Left a lot of comments but they are all instance of the same 4-5 things. @djspiewak two interesting points btw:
- do we want to encourage importing from `kernel`?
- look at Dispatcher, there are a couple of "mini-ends-of-the-world" cases here

djspiewak:
IMO if you have core, import from it. Kernel imports are for when you don't depend on core.

core/src/main/scala/org/http4s/internal/Logger.scala
djspiewak: More ideal to keep it as it was and import directly from core. Kernel isn't needed here since core is a dependency.

core/src/main/scala/org/http4s/internal/package.scala
djspiewak: Confirmed

core/src/main/scala/org/http4s/multipart/MultipartDecoder.scala
djspiewak: Nope

djspiewak:
I'll look at the dispatcher call sites when I get back to my laptop

core/src/main/scala/org/http4s/internal/Logger.scala
RaasAhsan: Why does http4s `core` need a dependency on CE `core`?

djspiewak: Fs2 already has one because of SyncIO

core/src/main/scala/org/http4s/internal/package.scala
```diff
@@ -14,37 +14,40 @@ import java.util.concurrent.{
...
 package object internal {
   // Like fs2.async.unsafeRunAsync before 1.0.  Convenient for when we
   // have an ExecutionContext but not a Timer.
-  private[http4s] def unsafeRunAsync[F[_], A](fa: F[A])(
-      f: Either[Throwable, A] => IO[Unit])(implicit F: Effect[F], ec: ExecutionContext): Unit =
-    F.runAsync(Async.shift(ec) *> fa)(f).unsafeRunSync()
+  private[http4s] def unsafeRunAsync[F[_]: Async, A](fa: F[A])(
+      f: Either[Throwable, A] => F[Unit])(implicit
+      dispatcher: Dispatcher[F],
+      ec: ExecutionContext): Unit =
```
djspiewak, line 35: Do you actually need to thread the `ExecutionContext` through anymore? Cats Effect contains its own EC, so in theory it shouldn't need to appear anywhere in the signatures.

core/src/main/scala/org/http4s/internal/package.scala
```diff
@@ -14,37 +14,40 @@ import java.util.concurrent.{
...
+      case Left(e) => F.delay(logger.error(e)("Error in asynchronous callback"))
+      case Right(_) => F.unit
     }
 
   // Inspired by https://github.com/functional-streams-for-scala/fs2/blob/14d20f6f259d04df410dc3b1046bc843a19d73e5/io/src/main/scala/fs2/io/io.scala#L140-L141
-  private[http4s] def invokeCallback[F[_]](logger: Logger)(f: => Unit)(implicit
-      F: ConcurrentEffect[F]): Unit =
-    F.runAsync(F.start(F.delay(f)).flatMap(_.join))(loggingAsyncCallback(logger)).unsafeRunSync()
+  private[http4s] def invokeCallback[F[_]](logger: Logger)(
+      f: => Unit)(implicit F: Async[F], dispatcher: Dispatcher[F]): Unit =
```
djspiewak, line 47: I wouldn't generally recommend making the `Dispatcher` implicit. You *can*, but it's a bit weird given that it needs to live within a `Resource`.

core/src/main/scala/org/http4s/internal/package.scala
```diff
@@ -123,68 +126,52 @@ package object internal {
   }
 
   // Adapted from https://github.com/typelevel/cats-effect/issues/199#issuecomment-401273282
-  @deprecated(
-    "Replaced by cats.effect.Async.fromFuture. You will need a ContextShift[F].",
-    "0.21.4")
+  @deprecated("Replaced by cats.effect.Async.fromFuture", "0.21.4")
```
djspiewak, line 129: I'm guessing this should just be removed entirely since this has to be a bincompat breaking release anyway.

core/src/main/scala/org/http4s/internal/package.scala
```diff
@@ -123,68 +126,52 @@ package object internal {
...
+        F.delay(cf.handle[Unit]((result, err) =>
           err match {
             case null => cb(Right(result))
             case _: CancellationException => ()
             case ex: CompletionException if ex.getCause ne null => cb(Left(ex.getCause))
             case ex => cb(Left(ex))
-          })
-        F.delay { cf.cancel(true); () }
+          })) >>
+          F.delay(Some(F.delay(cf.cancel(true)).void))
```
djspiewak, line 146: You probably want `F.pure` for the outer call here.

core/src/main/scala/org/http4s/internal/package.scala
rossabaker: All the usages are in blaze.  We could punt it to blaze-core right now, and see if we can eliminate entirely when we get to that backend.

core/src/main/scala/org/http4s/internal/package.scala
rossabaker: :+1: I like to err on the side of keeping deprecations for graceful migrations, but this is private, so we can be more aggressive here.

rossabaker (commented):
Thanks. This is an excellent start.

As you address the reviews (thanks, @SystemFw, @RaasAhsan, and @djspiewak. We're in good hands here!), please just add commits instead of force-pushing.  This will make it easier for everyone to review the incremental changes as this progresses.

djspiewak:
> > Do you actually need to thread the ExecutionContext through anymore? Cats Effect contains its own EC, so in theory it shouldn't need to appear anywhere in the signatures.
>
> All the usages are in blaze. We could punt it to blaze-core right now, and see if we can eliminate entirely when we get to that backend.

Oh is this the worker pool for blaze? Or is it blaze's `Future`-shifting pool? If it's the latter, then `Async[F].executionContext` should give you what you need (an `F[ExecutionContext]` that produces the compute pool). If it's the former, then continuing to thread it through makes sense, though it does result in some unfortunate API shenanigans higher up in the stack (my fault, I know!).

core/src/main/scala/org/http4s/EntityDecoder.scala
domaspoliakas: There's a single call that needs to be suspended downstream. I'll try to see if I can eliminate that.

core/src/main/scala/org/http4s/internal/package.scala
domaspoliakas: What would be the right action for this PR? should I move it to blaze-core but under the same package, keep it, or delete it?

core/src/main/scala/org/http4s/internal/package.scala
domaspoliakas: Can't see `unsafeRunAsync` on dispatcher, so I did `unsafeToFuture`; would that work the same?

domaspoliakas:
I _think_ I resolved most of the feedback now.

- Farewell kernel imports
- `: Sync: Files` eliminated; it is now either just `: Files` or `: Files: Concurrent`
- global IORuntime imports removed in favour of SyncIO
- couple of misc things

The things that are incomplete right now:
- the `unsafeRunAsync` that @rossabaker mentioned should go into blaze core
- One section on multipart parser (line 756). I'll leave a comment outlining my confusion there.

core/src/main/scala/org/http4s/internal/package.scala
RaasAhsan: `unsafeToFuture` will work everywhere you've got an `unsafeRunSync` but don't need backpressure. We could write an `unsafeRunAsync` that performs the `attemptTap` for you, I think that's going to be a common pattern

core/src/main/scala/org/http4s/multipart/MultipartParser.scala
```diff
@@ -776,15 +753,28 @@ object MultipartParser {
...
-        Pull
-          .eval(F.delay(Files.createTempFile("", "")))
-          .flatMap { path =>
-            (for {
-              _ <- Pull.eval(lacc.through(writeAll[F](path, blocker)).compile.drain)
-              split <- streamAndWrite(s, state, Stream.empty, racc, 0, path)
-            } yield split)
-              .handleErrorWith(e => Pull.eval(cleanupFile(path)) >> Pull.raiseError[F](e))
-          }
+        ???
```
domaspoliakas, line 756: My confusion is as thus: it seems a temp file is created but not always cleaned up. That does not sound right to me, so I tried to read through it to get a complete picture but I have to admit - I don't quite understand what's supposed to be happening here so I am left a little confused. Any help would be greatly appreciated.

domaspoliakas: Actually, I might be able to work it through. I'll give it a shot now.

core/src/main/scala/org/http4s/multipart/MultipartParser.scala
```diff
@@ -777,13 +754,14 @@ object MultipartParser {
...
       if (limitCTR >= maxBeforeWrite)
         Pull
-          .eval(F.delay(Files.createTempFile("", "")))
-          .flatMap { path =>
-            (for {
-              _ <- Pull.eval(lacc.through(writeAll[F](path, blocker)).compile.drain)
-              split <- streamAndWrite(s, state, Stream.empty, racc, 0, path)
-            } yield split)
-              .handleErrorWith(e => Pull.eval(cleanupFile(path)) >> Pull.raiseError[F](e))
+          .eval(Files[F].tempFile(???, "", "").allocated)
```
domaspoliakas, line 757: If someone could check if this logic correctly replaces previous logic that'd be very good

core/src/main/scala/org/http4s/internal/package.scala
domaspoliakas: I agree, were it not for removals I would've had to do that multiple times just in this PR alone.

core/src/main/scala/org/http4s/internal/package.scala
rossabaker: Let's move it to blaze-core, under its natural package.  That will help get core clean so other ports can commence, and give me more time to relearn and answer to @djspiewak's question. :smile:

core/src/main/scala/org/http4s/multipart/MultipartParser.scala
rossabaker: The `???` is the system property "java.io.tmpdir". The [underlying Java API](https://docs.oracle.com/javase/8/docs/api/java/io/File.html#createTempFile-java.lang.String-java.lang.String-java.io.File-) supports `null` there, and I don't know what else you'd do if that property is `None`, but I don't know what fs2 would do with that.

rossabaker: I think the logic is right.  `.allocate` is always a smell.  There's not a stream finalizer on the happy path to clean up the resource, which makes me think this function should return a `Resource`, which would bubble up and force the caller to use the temp files before they are disposed.  I don't know what ramifications that would have, but we could track that as a separate issue.

rossabaker (commented):
Great progress.

core/src/main/scala/org/http4s/multipart/MultipartParser.scala
domaspoliakas: I just checked `java.io.File` temp file creation allows nulls, but the nio temp file creation requires the dir property to be non-null and forces you to use the method which has no `dir` property for the default directory. So one way or the other the fs2 interface has to change unless we introduce a `Sync` requirement here again, but that'd be rather unfortunate. I'll see if I can get a change merged to fs2 to allow this.

domaspoliakas:
`core` now compiles; I will proceed to fixing tests

rossabaker (approved):
Fantastic work.  I'm happy going module-by-module on the PRs to keep them tractable and let others jump in.

project/Http4sPlugin.scala
```diff
@@ -218,13 +218,13 @@ object Http4sPlugin extends AutoPlugin {
...
     val cats = "2.2.0"
-    val catsEffect = "3.0-d5a2213"
+    val catsEffect = "3.0.0-M2"
     val catsEffectTesting = "0.4.1"
     val circe = "0.13.0"
     val cryptobits = "1.3"
     val disciplineSpecs2 = "1.1.0"
     val dropwizardMetrics = "4.1.13"
-    val fs2 = "3.0-5158029"
+    val fs2 = "3.0-cd73a32"
```
rossabaker, line 227: Oh, one question: Does this have anything that -M2 doesn't?

SystemFw: yeah, the `tempFile` and `tempDirectory` changes which unblock this PR, and were added today

SystemFw: https://github.com/http4s/http4s/pull/3784#discussion_r511556023

domaspoliakas:
Great! I presume I need to "undraft" this PR now, so I'll do that and then make a new branch for test upgrade.

blaze-core/src/main/scala/org/http4s/blazecore/package.scala
```diff
@@ -10,6 +10,13 @@ import cats.effect.{Resource, Sync}
 import org.http4s.blaze.util.{Cancelable, TickWheelExecutor}
 
 package object blazecore {
+
+  // Like fs2.async.unsafeRunAsync before 1.0.  Convenient for when we
+  // have an ExecutionContext but not a Timer.
+  private[http4s] def unsafeRunAsync[F[_], A](fa: F[A])(
+      f: Either[Throwable, A] => IO[Unit])(implicit F: Effect[F], ec: ExecutionContext): Unit =
+    F.runAsync(Async.shift(ec) *> fa)(f).unsafeRunSync()
```
SystemFw, line 18: Really keen to get to blaze and destroy this

core/src/main/scala/org/http4s/StaticFile.scala
```diff
@@ -98,86 +93,90 @@ object StaticFile {
...
-            .delay(urlConn.getInputStream.close())
+          F.blocking(urlConn.getInputStream.close())
             .handleError(_ => ())
             .as(Some(Response(NotModified)))
       }
     })
   }
 
-  def calcETag[F[_]: Sync]: File => F[String] =
+  def calcETag[F[_]: Files: MonadError[*[_], Throwable]]: File => F[String] =
```
SystemFw, line 108: Why not just Functor here?

domaspoliakas: I must've mistakenly replaced it along with some other `Functors` with `MonadError` while testing the waters. I'll get it changed to `Functor`.

PR #3787, opened 2020-10-27
Remove mockito from the build

rossabaker:
We use it in one test, and not to any great extent, and I'm tired of upgrading it.

Replaces #3786.


hamnis (approved):
Anytime mockito is removed from a project I rejoice.

Issue #3759, opened 2020-10-20
Port core to cats-effect-3


rossabaker:

rossabaker:
@domaspoliakas expressed an interest in this.  I'll assign it if he comments.

domaspoliakas:
Yep, already started. Should I just do ce3 or fs3 as well?

rossabaker:
They're going to have to go together.  I like @mpilquist's suggestion from Gitter:

> as we get more libraries actively porting to ce3, we’ll probably need to settle in to using milestones, but i’d start with the current fs2 hash release as it has Dispatcher

I _think_ `Dispatcher` is less critical for core, but it'll put us on a better foundation as we progress into the modules.

rossabaker:
Also, you want to push intermediate commits for feedback or see a way we might divide and conquer it, don't be shy to push works in progress.  I don't mind several failed builds along the way as long as it steers us toward a green one at the end. :smile:

domaspoliakas:
Sure thing! Do I push to my own fork and then PR for feedback?

rossabaker:
@domaspoliakas Yes, that's perfect. And you can open it as a draft to signify it's a WIP.

domaspoliakas:
I think what I'll do is get a little more done and then make a draft PR before the end of the day.

djspiewak:
I'll get a M2 up asap

domaspoliakas:
I created a draft PR now; would one normally add reviewers themselves or is it more like whoever can spare time would pick the PR up?

djspiewak:
Also just in case it helps… If there's anything I can do to help out on this, don't hesitate to ask. Happy to dive in and work on thorny bits in here as well.

SystemFw:
>  don't hesitate to ask.

I pinged you on a couple of things in the PR ;)

PR #3790, opened 2020-10-28
Roll back CI steps to what works right now

rossabaker:
Comments all the projects and steps that don't work yet.  Should let CI be meaningful on cats-effect-3 branch.

Whenever a new project is started, it needs to be uncommented in build.sbt.


rossabaker:
So #3759 doesn't compile on 2.12.12. /cc @domaspoliakas 

I think this can proceed: CI is known to be broken before this, and now it's showing our partial progress and also what's still broken.

domaspoliakas:
Ah right, I indeed never ran a cross compile. I should make sure to remember next time.

rossabaker:
Well, that's why we have the automated checks.  I forget, or take shortcuts, too!

Issue #3768, opened 2020-10-22
Add parens to lambdas for Dotty


rossabaker:
This is probably a scalafmt setting we can tweak.  It will be valid Scala 2, so we should do this on series/0.21.

```
[error] -- Error: /home/ross/src/http4s/core/src/main/scala/org/http4s/metrics/MetricsOps.scala:97:58 
[error] 97 |  ): Request[F] => Option[String] = { request: Request[F] =>
[error]    |                                                          ^
[error]    |           parentheses are required around the parameter of a lambda
[error]    |           This construct can be rewritten automatically under -rewrite.
```

jyoo980:
I can take this on!

jyoo980:
A cursory glance at the docs for `scalafmt` seems to show that there actually isn't an option to enforce this type of rule... @rossabaker do you have any other ideas on how to approach this? I guess I could compile with dotty and track down where these lambdas are. That said I'm not sure if dotty compiles the code nicely right now.

rossabaker:
Thanks for stepping forward!

I just remembered that @travisbrown [added them](https://github.com/http4s/http4s/pull/3242/commits/eab8d61b07461c25f0f0705687b930e0ad5fa2be) a while back.  Reapplying that commit might help, but it probably applies very uncleanly by now.  Even then, we need to find a switch so scalafmt doesn't eat them a second time.  A pure scalafmt solution would be better both for getting it done and keeping it done.

May also be worth checking how projects with a proper Dotty build have handled this.  Cats is one.

jyoo980:
Hey @travisbrown!

I heard you did some work to add parens around lambda parameters to prep `http4s` for dotty. Do you have any `scalafmt` rules you used to do this?

rossabaker:
@djspiewak pointed out the problem may be [this line](https://github.com/http4s/http4s/blob/3dd2340b0f6f2f78104f0c210436801afaaf0ed9/.scalafmt.conf#L15). If it were me, I'd try:

1. Remove it and rerun scalafmt.  Maybe we're done!
2. If not, try to reapply Travis' commit and see how bad the conflicts are.
3. Clean up what's left and what's new.

jyoo980:
@rossabaker those sound like great steps to run after I've added the parens around the lambda params. But before that, I'm wondering if there are any tools to point out where exactly in the code we'll need to do this. Doing this manually would probably be quite error-prone (but perhaps that's the only way to go about it right now)

djspiewak:
[...]

PR #3798, opened 2020-11-03
[WIP] adding first pass at parenthesizing lambdas with regexp

jyoo980:
Used regexp: `\{ .*:.*=>` since I had issues getting the `simulacrum-scalafix` plugin working.
 
See #3768 for more details.

I'm not _super_ confident that a regexp fix is great, since it's so fragile. That said I'll keep looking at making the scalafix rule work.


rossabaker (approved):
This is incremental progress.  I'm surprised we didn't have to remove the line from the scalafmt config, however.

rossabaker:
We'll want this on series/0.21, because Dotty needs it, and Dotty support forked off that branch.

rossabaker:
Cherry-picked this, and scalafmtAll isn't eating them anymore.

PR #3799, opened 2020-11-03
Remove incubator status

mpilquist:


SystemFw:
I think we should wait for 1.0 :trollface:

rossabaker (approved):
Incubator as a concept is on the way out, so :+1:.  

And @SystemFw's troll is acknowledged and respecte dand he knows where the cats-effect-3 tickets are. :trollface:

PR #3792, opened 2020-10-29
Add Show/Order/Hash Instances For QueryParameter{Key,Value}

isomarcte:

PR #3815, opened 2020-11-08
Add missing AbstractFunction1 inheritance

rossabaker:
We broke MiMa when we backported #3792.  This can safely be rolled back when merged to main.


rossabaker:
I'm going to merge this aggressively because it's making a mess of multiple other PRs.  Ex post facto reviews welcome.

PR #3824, opened 2020-11-10
CE3 Testing upgrade

domaspoliakas:
Hello,

Upgraded testing. Mostly involved adding ways for IO to be run unsafely.

One thing that I found peculiar was the resource usage in `Http4sSpec.withResource`. I couldn't quite understand why it wasn't just `.use` there before, but for now I just replaced it with that since `Resource` has been reworked and the old code doesn't work for it anymore anyway.


testing/src/test/scala/org/http4s/Http4sSpec.scala
```diff
@@ -99,18 +98,7 @@ trait Http4sSpec
...
-          .map { case (a, release) =>
-            fs(a).append(step(release(ExitCase.Completed).unsafeRunSync()))
-          }
-          .unsafeRunSync()
-      case Resource.Bind(r, f) =>
-        withResource(r)(a => withResource(f(a))(fs))
-      case Resource.Suspend(r) =>
-        withResource(r.unsafeRunSync() /* ouch */ )(fs)
-    }
+    r.use(a => IO(fs(a))).unsafeRunSync()
```
rossabaker, line 101: I think this is going to provide a released `a` to the fragments when they run.  Note how the old implementation was appending a step to the returned fragments.

If we can't use this allocate backdoor anymore, we're going to have a struggle in a couple modules.

rossabaker (changes requested):
Great start, but we're going to need to find another way for resourceful testing.

testing/src/test/scala/org/http4s/Http4sSpec.scala
hamnis: is the allocated method gone in cs3 completely?

hamnis: if not, could we use that to append the step?

domaspoliakas: `allocated` is still around. 

I'll go back to it and figure out a way 👍

testing/src/test/scala/org/http4s/Http4sSpec.scala
```diff
@@ -98,7 +98,8 @@ trait Http4sSpec
   }
 
   def withResource[A](r: Resource[IO, A])(fs: A => Fragments): Fragments =
-    r.use(a => IO(fs(a))).unsafeRunSync()
+    r.allocated.map { case (r, release) => fs(r).append(step(release.unsafeRunSync())) }.unsafeRunSync()
+    // r.use(a => IO(fs(a))).unsafeRunSync()
```
domaspoliakas, line 102: One day I'll remember to clean up before pushing I swear

rossabaker:
Needs a scalafmt, and I think it's ready.

domaspoliakas:
Did that. I also noticed CI was complaining about cats effect laws not necessary for laws and testing, so I removed that too.

rossabaker:
Oh, I discovered cleaned that up in #3832. Sorry, should have tagged you.

rossabaker:
Keep your version of the merge conflict, and I think we're all set.

domaspoliakas:
All done now

build.sbt
```diff
@@ -94,6 +94,10 @@ lazy val laws = libraryProject("laws")
     description := "Instances and laws for testing http4s code",
     libraryDependencies ++= Seq(
       caseInsensitiveTesting,
+<<<<<<< HEAD
+=======
+      catsLaws,
+>>>>>>> upstream/cats-effect-3
```
rossabaker, line 100: Keep `catsLaws`, lose the conflict markers.

domaspoliakas:
Apologies for the sloppy mess of a PR. I am very tired 😄

rossabaker:
We have 9.843 commits on main.  It's not the first mess!  You're doing great work, and this is the one that will open up a handful more modules.

rossabaker:
Build failure was a CI infrastructure flake. Restarting.

rossabaker (approved):
:+1: on green

PR #3826, opened 2020-11-11
Pin the Java version in the nix-shell

rossabaker:
This ensures we get a Java 8 even in SBT.

I realize I'm probably the only one who cares.


rossabaker:
Gah, that was supposed to go to 0.21.

PR #3828, opened 2020-11-11
Remove sbt-coursier

rossabaker:
Not needed in modern SBT, and is causing the dotty branch to fail to load.

Thanks to @mpilquist for finding this one.


rossabaker:
Part of #3848.

PR #3844, opened 2020-11-11
Implement a matrix path variable parser

rzeigler:
Approximates what is described at https://www.w3.org/DesignIssues/MatrixURIs.html

### DIFF
--- a/dsl/src/main/scala/org/http4s/dsl/impl/Path.scala
+++ b/dsl/src/main/scala/org/http4s/dsl/impl/Path.scala
@@ -242,7 +242,7 @@ abstract class MatrixVar[F[_]: Traverse, A](
     if (str.nonEmpty) {
       val firstSemi = str.indexOf(';')
       if (firstSemi < 0 && domain.nonEmpty) None
-      else if (firstSemi < 0) Some(Seq.empty[A])
+      else if (firstSemi < 0 && name == str) Some(Seq.empty[A])
       // Matrix segment didn't match the expected name
       else if (str.substring(0, firstSemi) != name) None
       else {
--- a/dsl/src/main/scala/org/http4s/dsl/impl/Path.scala
+++ b/dsl/src/main/scala/org/http4s/dsl/impl/Path.scala
@@ -14,7 +14,8 @@ import cats.data.Validated._
 import cats.implicits._
 import org.http4s._
 import scala.util.Try
-import cats.Traverse
+import cats.Foldable
+import cats.Monad
 
 /** Base class for path extractors. */
 trait Path {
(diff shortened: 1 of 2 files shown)


dsl/src/test/scala/org/http4s/dsl/PathSpec.scala
```diff
@@ -234,6 +235,95 @@ class PathSpec extends Http4sSpec {
...
+
+        "not enough axes" in {
+          (Path("/board/square;x=42") match {
+            case Root / "board" / BoardExtractor(x @ _, y @ _) => true
+            case _ => false
+          }) must beFalse
+        }
+
+        "to many axes" in {
+          (Path("/board/square;x=42y=0;z=39") match {
```
rossabaker, line 319: Is this supposed to be `x=42;y=0`?  And why is this invalid?  I might have expected extra axes to be ignored.

dsl/src/main/scala/org/http4s/dsl/impl/Path.scala
```diff
@@ -218,6 +219,74 @@ object LongVar extends PathVar(str => Try(str.toLong))
...
+  *    object BoardVar extends MatrixVar("square", List("x", "y"), str => Try(str.toInt).toOption)
+  *    Path("/board/square;x=5;y=3") match {
+  *      case Root / "board" / BoardVar(x, y) => ...
+  *    }
+  * }}}
+  */
+abstract class MatrixVar[F[_]: Traverse, A](
+    name: String,
+    domain: F[String],
+    range: String => Option[A]) {
```
rossabaker, line 237: In common use, are all parameters in the matrix of the same type?

```
/address;city=Indianapolis;zipCode=46268
```

Do I have to represent `zipCode` as a `String`?  If we stripped `range` and returned `String`s, could we nest extractors?

```
case Root / AddressVar(city, Zip(zipCode)) =>
```

rossabaker (commented):
Thanks.  Before I dive too deep into the implementation, a couple quick design thoughts.

dsl/src/main/scala/org/http4s/dsl/impl/Path.scala
rzeigler: I'm not really sure there is a common use. The problem I was specifically looking for was a means to solve awkward urls like /board/x/0/y/1 which implies a hierarchical structure of resources where none exists. Beyond that the goal is something that more people than just me is an ergonomic API. range is basically a mini-extractor so I will make this change.

dsl/src/test/scala/org/http4s/dsl/PathSpec.scala
rzeigler: The similarity to query parameters probably means that they should be. I just need to make the toAssocList stack safe since its currently capped at the number of parameters that are expected to be extracted.

rzeigler:
I've tried to address the comments and also make what is excepted more permissive.

rossabaker (approved):
Neat!

PR #3845, opened 2020-11-12
Use sbt-github-actions

amesgen:
Supersedes and closes #3825


amesgen:
MIMA is disabled as it fails right now, and I don't know why: [logs](https://github.com/amesgen/http4s/runs/1388170126)

Also TODO: adapt mergify

project/Http4sPlugin.scala
```diff
@@ -31,11 +31,11 @@ object Http4sPlugin extends AutoPlugin {
     ThisBuild / http4sApiVersion := (ThisBuild / version).map {
       case VersionNumber(Seq(major, minor, _*), _, _) => (major.toInt, minor.toInt)
     }.value,
-  )
+  ) ++ sbtghactionsSettings
 
   override lazy val projectSettings: Seq[Setting[_]] = Seq(
     scalaVersion := scala_213,
-    crossScalaVersions := Seq(scala_213, scala_212),
+    ThisBuild / crossScalaVersions := Seq(scala_213, scala_212),
```
rossabaker, line 38: Can a `ThisBuild` go in `projectSettings`?  I thought that needed to be in `buildSettings`?

project/Http4sPlugin.scala
```diff
@@ -227,6 +227,73 @@ object Http4sPlugin extends AutoPlugin {
...
+    inThisBuild(Http4sOrgPlugin.githubActionsSettings ++ Seq(
+      githubWorkflowBuild := Seq(
+        WorkflowStep
+          .Sbt(List("scalafmtCheckAll"), name = Some("Check formatting")),
+        WorkflowStep.Sbt(List("headerCheck", "test:headerCheck"), name = Some("Check headers")),
+        WorkflowStep.Sbt(List("test:compile"), name = Some("Compile")),
+        // TODO reenable
+        // WorkflowStep.Sbt(List("mimaReportBinaryIssues"), name = Some("Check binary compatibility")),
+        WorkflowStep.Sbt(
+          List("unusedCompileDependenciesTest"),
```
rossabaker, line 266: I'd still like to turn on the other test eventually, but that's a separate ticket.

project/Http4sPlugin.scala
```diff
@@ -227,6 +227,73 @@ object Http4sPlugin extends AutoPlugin {
...
+          name = Some("Check explicit dependencies")),
+        WorkflowStep.Sbt(List("test"), name = Some("Run tests")),
+        WorkflowStep.Sbt(List("doc"), name = Some("Build docs")),
+        setupHugoStep,
+        siteBuildStep("website"),
+        siteBuildStep("docs")
```
rossabaker, line 272: These used to be their own job, which meant they could run in parallel.  It also meant they could update if one of the safety checks failed, so maybe that was bad.

rossabaker (commented):
Nicely done. This is probably going to fail somewhere, but these things always require iteration, and it's a great start.

We should target series/0.21 with this instead.  Most work is still starting there, and I think those builds will start failing on November 16 without this.

project/Http4sPlugin.scala
amesgen: I have only a very rudimentary understanding of the differences between `projectSettings` and `buildSettings`. I had to add `ThisBuild / ` for sbt-github-actions to pick it up, and it appears to work(?). I can move it to though, it seems to continue to work locally with it being moved.

rossabaker (commented):
I thought MiMa might be related to fetch depth and not finding the tags, but it looks like the depth is 0 on both checkouts. :thinking:

amesgen:
The fail-fast behavior is not optimal, will create a PR to sbt-github-actions to allow to disable it.

I am confused about the last build error:
```
sbt.librarymanagement.ResolveException: Error downloading org.http4s:http4s-core_2.12:1.0.0-M6+9-6595f52a-SNAPSHOT
```

amesgen:
Wait, I should have not used the GitHub UI branch retarget feature... Will close and reopen

PR #3846, opened 2020-11-12
Use sbt GitHub actions new

amesgen:
Supersedes and closes #3825

- [ ] MIMA is disabled as it fails right now, and I don't know why: logs
- [ ] TODO: adapt mergify
- [ ] Put `ThisBuild` stuff in `buildSettings`? [see here](https://github.com/http4s/http4s/pull/3845#discussion_r521767157)


amesgen:
Wow I am too tired...

project/Http4sPlugin.scala
```diff
@@ -31,11 +31,11 @@ object Http4sPlugin extends AutoPlugin {
     ThisBuild / http4sApiVersion := (ThisBuild / version).map {
       case VersionNumber(Seq(major, minor, _*), _, _) => (major.toInt, minor.toInt)
     }.value,
-  )
+  ) ++ sbtghactionsSettings
 
   override lazy val projectSettings: Seq[Setting[_]] = Seq(
     scalaVersion := scala_213,
-    crossScalaVersions := Seq(scala_213, scala_212),
+    ThisBuild / crossScalaVersions := Seq(scala_213, scala_212),
```
satorg, line 38: I guess you meant to add `crossScalaVersion` to the build scope of each project. If so then perhaps you may consider overriding `def buildSettings` method rather than placing in into `projectSettings` with explicit scoping ([doc](https://www.scala-sbt.org/1.x/docs/Plugins.html#projectSettings+and+buildSettings))

PR #3847, opened 2020-11-12
Use sbt-github-actions

amesgen:
Supersedes and closes #3825

- [x] MIMA is disabled as it fails right now, and I don't know why: [logs](https://github.com/amesgen/http4s/runs/1388170126) UPDATE: pushing recent git tags to the fork did help 
- [x] adapt mergify
- [x] Put `ThisBuild` stuff in `buildSettings`? [see here](https://github.com/http4s/http4s/pull/3845#discussion_r521767157)
- [ ] Disable fail-fast strategy djspiewak/sbt-github-actions#33


project/Http4sPlugin.scala
```diff
@@ -234,6 +234,77 @@ object Http4sPlugin extends AutoPlugin {
...
+        // TODO reenable
+        // WorkflowStep.Sbt(List("mimaReportBinaryIssues"), name = Some("Check binary compatibility")),
+        WorkflowStep.Sbt(
+          List("unusedCompileDependenciesTest"),
+          name = Some("Check explicit dependencies")),
+        WorkflowStep.Sbt(List("test"), name = Some("Run tests")),
+        WorkflowStep.Sbt(List("doc"), name = Some("Build docs")),
+        setupHugoStep,
+        siteBuildStep("website"),
+        siteBuildStep("docs")
```
amesgen, line 279: > These used to be their own job, which meant they could run in parallel. It also meant they could update if one of the safety checks failed, so maybe that was bad.

Let's see how much time they take. sbt-github-actions supports adding custom jobs, so it is definitely possible to recreate the old behaviour.

amesgen:
Hmm, why is CI not running? Anyway, I am AFK for a few hours.

project/Http4sPlugin.scala
```diff
@@ -31,11 +31,11 @@ object Http4sPlugin extends AutoPlugin {
     ThisBuild / http4sApiVersion := (ThisBuild / version).map {
       case VersionNumber(Seq(major, minor, _*), _, _) => (major.toInt, minor.toInt)
     }.value,
-  )
+  ) ++ sbtghactionsSettings
 
   override lazy val projectSettings: Seq[Setting[_]] = Seq(
     scalaVersion := scala_213,
-    crossScalaVersions := Seq(scala_213, scala_212),
+    ThisBuild / crossScalaVersions := Seq(scala_213, scala_212),
```
satorg, line 38: Did you mean adding `crossScalaVersion` to the build scope of each project? If yes then perhaps it might be better to override `def buildSettings` method rather than keep it in `projectSettings` with explicit scoping ([doc](https://www.scala-sbt.org/1.x/docs/Plugins.html#projectSettings+and+buildSettings))

amesgen: Thanks for the link, will do!

amesgen:
CI is running in my fork, but still not here for some reason. See [here](https://github.com/amesgen/http4s/commits/use-sbt-github-actions-new) for the commits and CI statuses.

amesgen:
Ok, I am now stuck with a dependency resolution exception (excerpt):
```
coursier.ResolutionException: Encountered 8 error(s) in dependency resolution:
    org.http4s:http4s-dropwizard-metrics_2.12:0.21.1+614-21ba2344-SNAPSHOT:
    org.http4s:http4s-prometheus-metrics_2.12:0.21.1+614-21ba2344-SNAPSHOT:
    org.http4s:http4s-blaze-server_2.12:0.21.1+614-21ba2344-SNAPSHOT:
```
https://github.com/amesgen/http4s/runs/1390991987?check_suite_focus=true

They occur with caching enabled and disabled (contrary to [this commit message](https://github.com/http4s/http4s/pull/3847/commits/c20c8be0c3175beeb767529233990a4e4a2389bb)).

rossabaker:
I'm not sure why it isn't running here.

sbt-ci-release sets the version from the tags.  Your fork doesn't have the latest tags, which is why your version number is 0.21.1-something.  The hash lines up, so the doc step is trying to resolve what was just built, but... why?  It should just be a `dependsOn` for the docs being built.

amesgen:
Ok, MIMA now also works after I pushed the tags to my fork.

project/Http4sPlugin.scala
amesgen: Done.

amesgen:
Ok, the dependency resolution issue above seems to be caused by setting `ThisBuild / crossScalaVersions`. sbt-github-actions allows to customize the scala versions without setting `ThisBuild / crossScalaVersions` in theory, but there is a (trivial) bug right now: djspiewak/sbt-github-actions#34

amesgen:
Wooo, CI does pass now! https://github.com/amesgen/http4s/runs/1392973971

rossabaker:
Speculation: is it not running in http4s org because it's a draft?

rossabaker:
#3851 ran as a draft, and yours didn't as a ready-for-review, so my theory is wrong.

amesgen:
This PR should be ready now, only the matrix fail-fast support is missing, which will be possible with the next release of sbt-github-actions. Now closing this PR and creating a new one in order to get proper CI.

PR #3848, opened 2020-11-12
Start trying to cross build for 3.0.0-M1

rossabaker:
From here, we can run `++3.0.0-M1 core/test` (which still fails), and the Scala 2 builds should be undisturbed.

If we can make progress on Dotty without breaking Scala 2, then we can drop the long-running `dotty` branch, which is almost entirely merge commits.


rossabaker:
Never mind. I don't like the split versions on the live production branch. But I'm going to decruft the `dotty` branch.

PR #3850, opened 2020-11-12
Make scalatags encoder a bit more generic

akovbovich:
This is necessary to render fragments with `Frag` annotations.

### DIFF
--- a/scalatags/src/main/scala/org/http4s/scalatags/ScalatagsInstances.scala
+++ b/scalatags/src/main/scala/org/http4s/scalatags/ScalatagsInstances.scala
@@ -7,15 +7,15 @@
 package org.http4s
 package scalatags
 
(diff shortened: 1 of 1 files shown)


rossabaker (commented):
I don't use scalatags, so I'm trusting this is good.

Are you on the 0.21 series or tracking the 1.0 milestones?  This is binary incompatible, but we might get this released in 0.21 if we'd make  `scalatagsEncoder` not implicit, and give your corrected signature a new name.

rossabaker:
Someone can backport it as suggested above if they need it in 0.21, but there's no reason to keep sitting on it until then.

rossabaker:
I was preparing release notes and realized that actually this is binary compatible. No reason we can't do this in 0.21.

PR #3852, opened 2020-11-13
Port Boopickle to cats effect 3 and add basic munit infrastructure

cquiroz:
This PR ports boopickle to run on CE3 but perhaps more important it lays the basics to use
munit/munit-cats-effect for tests.

There is still some work to do but I wanted to verify if the idea of munit seems workable

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -30,7 +30,7 @@ lazy val modules: List[ProjectReference] = List(
   // theDsl,
    jawn,
   // argonaut,
-  // boopickle,
+  boopickle,
   // circe,
   // json4s,
   // json4sNative,
@@ -105,7 +105,11 @@ lazy val testing = libraryProject("testing")
     description := "Instances and laws for testing http4s code",
     libraryDependencies ++= Seq(
       specs2Matcher,
+      munitCatsEffect,
+      munitDiscipline
     ),
+    unusedCompileDependenciesFilter -= moduleFilter(organization = "org.typelevel", name = "discipline-munit"),
+    unusedCompileDependenciesFilter -= moduleFilter(organization = "org.typelevel", name = "munit-cats-effect-3"),
   )
   .dependsOn(laws)
 
@@ -661,6 +665,7 @@ def http4sProject(name: String) =
     .settings(
       moduleName := s"http4s-$name",
       Test / testOptions += Tests.Argument(TestFrameworks.Specs2, "showtimes", "failtrace"),
+      testFrameworks += new TestFramework("munit.Framework"),
       initCommands()
     )
     .enablePlugins(AutomateHeaderPlugin)
--- a/testing/src/test/scala/org/http4s/Http4sSuite.scala
+++ b/testing/src/test/scala/org/http4s/Http4sSuite.scala
@@ -1,0 +1,15 @@
+/*
+ * Copyright 2013-2020 http4s.org
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.http4s
+
+import munit._
+
+/** Common stack for http4s' munit based tests
+  */
+trait Http4sSuite extends CatsEffectSuite with DisciplineSuite {}
+
+object Http4sSuite {}
--- a/boopickle/src/main/scala/org/http4s/booPickle/BooPickleInstances.scala
+++ b/boopickle/src/main/scala/org/http4s/booPickle/BooPickleInstances.scala
@@ -9,7 +9,7 @@ package booPickle
 
 import boopickle.Default._
 import boopickle.Pickler
-import cats.effect.Sync
+import cats.effect.Concurrent
 import fs2.Chunk
 import java.nio.ByteBuffer
 import org.http4s._
@@ -21,7 +21,7 @@ import scala.util.{Failure, Success}
   * Note that the media type is set for application/octet-stream
   */
 trait BooPickleInstances {
-  private def booDecoderByteBuffer[F[_]: Sync, A](m: Media[F])(implicit
+  private def booDecoderByteBuffer[F[_]: Concurrent, A](m: Media[F])(implicit
       pickler: Pickler[A]): DecodeResult[F, A] =
     EntityDecoder.collectBinary(m).subflatMap { chunk =>
       val bb = ByteBuffer.wrap(chunk.toArray)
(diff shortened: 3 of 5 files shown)


boopickle/src/test/scala/org/http4s/booPickle/BoopickleSpec.scala
```diff
@@ -54,27 +53,23 @@ class BoopickleSpec extends Http4sSpec with BooPickleInstances {
...
-      result.value.unsafeRunSync() must_== Right(Banana(10.0))
-    }
+  test("decode a class from a boopickle decoder") {
+    val result = booOf[IO, Fruit]
+      .decode(Request[IO]().withEntity(Banana(10.0): Fruit), strict = true)
+    result.value.map(assertEquals(_, Right(Banana(10.0))))
   }
 
-  checkAll("EntityCodec[IO, Fruit]", EntityCodecTests[IO, Fruit].entityCodec)
+  // checkAll("EntityCodec[IO, Fruit]", EntityCodecTests[IO, Fruit].entityCodec)
```
cquiroz, line 74: `EntityCodecTests` needs a rewrite to run on munit-cats-effect, if you think munit is fine I could try to do the rewrite

testing/src/test/scala/org/http4s/Http4sSuite.scala
```diff
@@ -0,0 +1,20 @@
...
+ */
+
+package org.http4s
+
+import munit._
+import org.http4s.laws.discipline.ArbitraryInstances
+
+/** Common stack for http4s' munit based tests
+  */
+trait Http4sSuite
```
cquiroz, line 14: This is equivalent to `Http4sSpec` for munit. It will need to grow as needed

build.sbt
```diff
@@ -53,12 +53,13 @@ lazy val modules: List[ProjectReference] = List(
...
 
-lazy val root = project.in(file("."))
+lazy val root = project
+  .in(file("."))
   .enablePlugins(PrivateProjectPlugin)
   .settings(
     // Root project
     name := "http4s",
-    description := "A minimal, Scala-idiomatic library for HTTP",
+    description := "A minimal, Scala-idiomatic library for HTTP"
```
rossabaker, line 62: You don't like my trailing commas? :laughing:

boopickle/src/test/scala/org/http4s/booPickle/BoopickleSpec.scala
rossabaker: Why does it need a rewrite?  I think it's based just on Discipline?

I'm for what makes it work.  I just haven't studied it enough to understand.

boopickle/src/test/scala/org/http4s/booPickle/BoopickleSpec.scala
```diff
@@ -54,27 +53,23 @@ class BoopickleSpec extends Http4sSpec with BooPickleInstances {
...
-  "booOf" should {
-    "decode a class from a boopickle decoder" in {
-      val result = booOf[IO, Fruit]
-        .decode(Request[IO]().withEntity(Banana(10.0): Fruit), strict = true)
-      result.value.unsafeRunSync() must_== Right(Banana(10.0))
-    }
+  test("decode a class from a boopickle decoder") {
+    val result = booOf[IO, Fruit]
+      .decode(Request[IO]().withEntity(Banana(10.0): Fruit), strict = true)
+    result.value.map(assertEquals(_, Right(Banana(10.0))))
```
rossabaker, line 71: I love the removal of an `unafeRunSync()`. :heart:

testing/src/test/scala/org/http4s/Http4sSuite.scala
```diff
@@ -0,0 +1,20 @@
...
+    extends CatsEffectSuite
+    with syntax.AllSyntax
+    with ArbitraryInstances
```
rossabaker, line 17: In Cats, it was discussed to not extend syntax traits or arbitraries in the base test suite, so as to make the tests look more like the kind of code it would be testing.  We will have a few more imports in our tests without these two lines, but I think it might be better, and this is our chance at a fresh start.  What do you think?

rossabaker:
Also, if this works, we should consider backporting the munit infrastructure to `series/0.21`.  Keeping those branches aligned as much as possible works strongly to our benefit while we're finishing up 1.0.

build.sbt
cquiroz: I think they were removed by scalafmt; I can restore them

boopickle/src/test/scala/org/http4s/booPickle/BoopickleSpec.scala
cquiroz: The particular laws take an `Eq[IO[A]]` which we don't have right now

build.sbt
rossabaker: I'm maintaining four branches right now, so I'm just a little bit frantic about unnecessary diffs right now, especially when they're on the furthest branch downhill.

testing/src/test/scala/org/http4s/Http4sSuite.scala
cquiroz: I'll remove everything we don't need right away.

build.sbt
```diff
@@ -105,7 +105,11 @@ lazy val testing = libraryProject("testing")
     description := "Instances and laws for testing http4s code",
     libraryDependencies ++= Seq(
       specs2Matcher,
+      munitCatsEffect,
+      munitDiscipline
     ),
+    unusedCompileDependenciesFilter -= moduleFilter(organization = "org.typelevel", name = "discipline-munit"),
```
cquiroz, line 111: I need this to pass CI but I'm not sure why, we are using classes from both libs

cquiroz:
Maybe we can merge this with the law check commented while we figure out how to check effectful laws with discipline and munit

rossabaker (approved):
That works for me.

If we're integrating munit, I'd like to bring back the munit infrastructure to series/0.21, so it rolls downhill.  I can take care of that, unless you want to.

cquiroz:
I can do it, I'll rebase this and merge it

cquiroz:
Ok I squashed this and I will merge to move on to others, The doc task is not yet building

PR #3855, opened 2020-11-13
Introducing cats-parse

rossabaker:
`HttpVersion` is not a particularly interesting class to parse, but I'm going to bikeshed it while we're setting precedent.


core/src/main/scala/org/http4s/HttpVersion.scala
```diff
@@ -35,20 +35,21 @@ object HttpVersion {
       case "HTTP/1.1" => Either.right(`HTTP/1.1`)
       case "HTTP/1.0" => Either.right(`HTTP/1.0`)
       case _ =>
-        new Parser(s).HttpVersion.run()(Parser.DeliveryScheme.Either).leftMap { _ =>
-          ParseFailure("Invalid HTTP version", s"$s was not found to be a valid HTTP version")
+        parser.parseAll(s).leftMap { case e =>
+          ParseFailure("Invalid HTTP version", e.toString)
```
rossabaker, line 39: Is there a general pretty printer for `Error`?  Is such a thing possible?  We didn't offer much detail in the parboiled implementation, and don't have to here.  Just exploring what's possible.

core/src/main/scala/org/http4s/HttpVersion.scala
```diff
@@ -35,20 +35,21 @@ object HttpVersion {
       case "HTTP/1.1" => Either.right(`HTTP/1.1`)
       case "HTTP/1.0" => Either.right(`HTTP/1.0`)
```
rossabaker, line 36: Historically these two dominate and microbenchmarks showed it was safer to jump straight to them.  Is this still true in cats-parse?

core/src/main/scala/org/http4s/HttpVersion.scala
```diff
@@ -35,20 +35,21 @@ object HttpVersion {
...
-        "HTTP/" ~ capture(Digit) ~ "." ~ capture(Digit) ~> { (major: String, minor: String) =>
-          new HttpVersion(major.toInt, minor.toInt)
-        }
-      }
+  private val parser: Parser1[HttpVersion] = {
+    // HTTP-name = %x48.54.54.50 ; HTTP
+    val httpName = P.string1("HTTP")
+
+    // HTTP-version = HTTP-name "/" DIGIT "." DIGIT
+    val httpVersion = httpName ~ P.char('/') *> digit ~ (P.char('.') *> digit)
```
rossabaker, line 48: Does `httpName ~ P.char('/')` get optimized to `P.string("HTTP/")`?  I'm trying to mimic the RFC, but I'd rather save a few nanoseconds.  Doesn't matter here, but it will start to matter in more complicated parsers.

core/src/main/scala/org/http4s/HttpVersion.scala
```diff
@@ -35,20 +35,21 @@ object HttpVersion {
...
-  private class Parser(val input: ParserInput)
-      extends org.http4s.internal.parboiled2.Parser
-      with Rfc2616BasicRules {
-    def HttpVersion: Rule1[org.http4s.HttpVersion] =
-      rule {
-        "HTTP/" ~ capture(Digit) ~ "." ~ capture(Digit) ~> { (major: String, minor: String) =>
-          new HttpVersion(major.toInt, minor.toInt)
-        }
-      }
+  private val parser: Parser1[HttpVersion] = {
```
rossabaker, line 43: Some of our parsers have existed on the companion of what they're parsing so they can access a private constructor after validation.  It prevents a `private [http4s] unsafeFromVersion(Int, Int)`.

Do we like this pattern or no?  The lack of inheritance perhaps makes it more palatable than it was.

core/src/main/scala/org/http4s/internal/parsing/Rfc5234.scala
```diff
@@ -0,0 +1,31 @@
...
+
+/** Common rules defined in RFC5234 and imported into various HTTP
+  * specs.
+  *
+  * @see https://tools.ietf.org/html/rfc5234#appendix-B.1
+  */
+private[http4s] object Rfc5234 {
+  val alpha: P[Char] =
+    P.charIn(0x41.toChar to 0x5a.toChar).orElse1(P.charIn(0x61.toChar to 0x7a.toChar))
+  val cr: Parser1[Unit] = P.char(0x0d)
```
rossabaker, line 19: cats-parse returns `Parser1[Unit]` for the constants and `Parser1[Char]` when there is a choice.  I followed that convention here.  Is there a drawback to making these `Parser1[Char]` across the board?

core/src/main/scala/org/http4s/internal/parsing/Rfc5234.scala
```diff
@@ -0,0 +1,31 @@
...
+  val lf: Parser1[Unit] = P.char(0x0a)
+  val crlf: Parser1[Unit] = (cr ~ lf).void
+  val ctl: Parser1[Char] =
+    P.charIn(0x00.toChar to 0x1f.toChar).orElse1(P.char(0x7f).as(0x7f.toChar))
+  val digit: Parser1[Char] = P.charIn(0x30.toChar to 0x39.toChar)
```
rossabaker, line 24: This is `cats.parse.Numbers.digit`, but we directly parrot the RFC this way.

core/src/main/scala/org/http4s/internal/parsing/Rfc5234.scala
```diff
@@ -0,0 +1,31 @@
...
+import cats.parse.{Parser => P, Parser1}
+
+/** Common rules defined in RFC5234 and imported into various HTTP
+  * specs.
+  *
+  * @see https://tools.ietf.org/html/rfc5234#appendix-B.1
+  */
+private[http4s] object Rfc5234 {
+  val alpha: P[Char] =
+    P.charIn(0x41.toChar to 0x5a.toChar).orElse1(P.charIn(0x61.toChar to 0x7a.toChar))
```
rossabaker, line 18: I see some bitset magic.  Do these get optimized when stated like this, or should we express them with a predicate?

rossabaker (commented):
Spamming the cats-parse contributors, who might be interested in early usage. Thanks, and unsubscribe button is on your right. :smile:

/cc @johnynek @regadas @Slakah @stephenjudkins @zmmcoy @non @mpilquist

build.sbt
```diff
@@ -79,13 +79,17 @@ lazy val core = libraryProject("core")
...
       fs2Io,
       log4s,
       parboiled,
       scalaReflect(scalaVersion.value) % Provided,
       vault,
     ),
     unusedCompileDependenciesFilter -= moduleFilter("org.scala-lang", "scala-reflect"),
+    mimaBinaryIssueFilters ++= Seq(
+      ProblemFilters.exclude[MissingClassProblem]("org.http4s.HttpVersion$Parser"),
+    ),
```
rossabaker, line 92: We have four options here:
1. Deprecate this, and leave dead code in Dotty that would explode.
2. Deprecate and move to a scala-2 directory. We'd need an `HttpVersionCompat` mixin.
3. Remove it and grant MiMa exceptions as necessary.
4. Surrender and call this branch 0.22, which might be necessary anyway.

On Gitter, @ChristopherDavenport and I overlapped on choice 3, so I'm pursuing that here.

aeons: 3 or 4 seems like the way to go to me.

rossabaker: I was in favor of 2 or 3, and Chris liked 3 or 4.

I think the thing that may force an 0.22 is several integrations not publishing for Dotty (e.g. Play, Twirl), so we have to split up the repo, and therefore the release cycles.

core/src/main/scala/org/http4s/internal/parsing/Rfc5234.scala
johnynek: Yes. These are merged into a single BitSet.

core/src/main/scala/org/http4s/internal/parsing/Rfc5234.scala
johnynek: No we could do that. I made Parser.string return Unit and then I thought Parser.char should match string.

The downside is void comes up a lot, and then we would need a wrapper to change the type (Void(p1)).

Open an issue on cats-parse and we can discuss more the design space.

core/src/main/scala/org/http4s/HttpVersion.scala
johnynek: No those are not optimized like that because they are not the same.

If http parses then you don't find the / that is an arresting failure. Strings however peek and automatically backtrack.

And equivalent thing would be `http.soft ~ slash`, this makes a softProduct where if there are no arresting failures it will backtrack.

Lastly, that equivalent thing is not optimized but it should (though I don't know how often it really comes up).

core/src/main/scala/org/http4s/HttpVersion.scala
johnynek: There is not. The expectations are sorted but there isn't much pretty about it. I have some utilities for this in bosatsu I can pull out (mostly about pointing at the part of the string that failed). That would be good to add to cats-parse.

core/src/main/scala/org/http4s/HttpVersion.scala
johnynek: It would be interesting to see. I'd be a bit surprised this would be better than an ideal since you are checking a whole string, but if it matches 99% of the time on the first string then probably it is a win.

rossabaker: I couldn't beat this even making them specific cases in parboiled2, or using parboiled2's map facility.  No matter how little work the parser did, it couldn't compete with two string comparisons.

core/src/main/scala/org/http4s/HttpVersion.scala
rossabaker: Oh, right, that makes sense.

It's not _that_ often that the HTTP ABNF smooshes together two constants.  It's more an academic exercise in how literally we can translate the specs without sacrificing any performance.

core/src/main/scala/org/http4s/HttpVersion.scala
johnynek: It would be interesting to see.... The string parser is calling regionMatches with the offset, so there is no internal copying or anything like that.

core/src/main/scala/org/http4s/internal/parsing/Rfc5234.scala
johnynek: Also note: `someParser.string` captures the string of the whole parser by not capturing anything, then using the offsets to do one final substring at the end. If you do want the whole string at the end, that is usually not bad at all.

If you literally want Parser.string(str).string that is a waste since it if matches we wouldn't have to allocate: we already have the string, but I guess I can optimize that into a map that returns the original string without a copy.

rossabaker:
Going to retarget this for the dotty branch.  It works on series/0.21, but we don't want to introduce an 0.x core dependency there that doesn't materially improve users' lives.  This is motivated by Dotty, so let's aim for a series of small PRs to that branch and divvy the work among the interested.

core/src/main/scala/org/http4s/HttpVersion.scala
rossabaker: I realized we could do better and cache the `Right` wrapper. [Gist](https://gist.github.com/rossabaker/f97a7f340296575dc6fd90197d7b1cb5)

Results:

```
[info] Benchmark                        Mode  Cnt          Score          Error  Units
[info] ParserBench.parseFlyweight_1_0  thrpt    4   13739316.905 ± 1074228.764  ops/s
[info] ParserBench.parseFlyweight_1_1  thrpt    4   13644878.755 ±  351068.389  ops/s
[info] ParserBench.parseFlyweight_2_0  thrpt    4   13290807.918 ± 1223035.879  ops/s
[info] ParserBench.parseMatch_1_0      thrpt    4  353128170.066 ± 41155378.386  ops/s
[info] ParserBench.parseMatch_1_1      thrpt    4  346638271.637 ± 26582497.893  ops/s
[info] ParserBench.parseMatch_2_0      thrpt    4   13613696.324 ±   251148.799  ops/s
[info] ParserBench.parseOrElses_1_0    thrpt    4   26153858.757 ± 1530188.268  ops/s
[info] ParserBench.parseOrElses_1_1    thrpt    4   31797487.174 ± 7072299.041  ops/s
[info] ParserBench.parseOrElses_2_0    thrpt    4    8470594.820 ±  249313.539  ops/s
[info] ParserBench.parseSimple_1_0     thrpt    4   13589006.294 ±   880396.476  ops/s
[info] ParserBench.parseSimple_1_1     thrpt    4   13429852.948 ±  2284349.941  ops/s
[info] ParserBench.parseSimple_2_0     thrpt    4   13239167.061 ±  1578824.941  ops/s
```

It's possible I'm cheating by parsing the same string over and over again instead of a mix, but I think caching the Right makes sense.  

This is a rare opportunity for optimization.  We won't be doing similar in most of our parsers.

johnynek: Pretty conclusive. Big gap. One note: I would use string1("HTTP/") and not compose the / on with char.

johnynek: Also note you have to allocate the internal mutable State and the result for even the shortest parse. So I guess for this example, checking the likely case is worth it.

core/src/main/scala/org/http4s/internal/parsing/Rfc5234.scala
rossabaker: This object has been upstreamed to cats-parse, so it's not our worry.

laws/src/main/scala/org/http4s/laws/discipline/ArbitraryInstances.scala
```diff
@@ -84,14 +84,26 @@ private[http4s] trait ArbitraryInstances {
...
+
+  val genTchar: Gen[Char] = oneOf(tchars)
 
   val genToken: Gen[String] =
     nonEmptyListOf(genTchar).map(_.mkString)
 
+  val genNonTchar = frequency(
+    4 -> oneOf(Set(0x00.toChar to 0x7f.toChar: _*) -- tchars),
+    1 -> oneOf(0x100.toChar to Char.MaxValue)
+  )
```
rossabaker, line 99: Something like this turns out to be a much more interesting generator than extended charsets in a lot of HTTP parsing cases.  We could make some considerable improvements throughout this file.

core/src/main/scala/org/http4s/HttpVersion.scala
hamnis: I use this pattern in internal libs a lot, and have found it quite useful.

rossabaker: As long as we have no circular dependencies between parsed types, I like it, too.  No way to construct an invalid one outside this compilation unit.

rossabaker:
I think we've laid down some good patterns here, and I'd like to get this merged and keep chipping away at individual parsers before this grows out of control.

hamnis (approved):

PR #3857, opened 2020-11-13
Server/client ce3 port

RaasAhsan:
Started to port `server`/`client` over to CE3. I definitely need to revise some constraints but here are some common patterns I'm seeing:
- `Monad` + `Clock` -> `Temporal` (I think there's a proposal to have `Clock` extend `Monad`)
- `Clock.monotonic` is always converted to `Long` nanos

Still need to port the tests over which should be fairly quick


SystemFw:
> Monad + Clock -> Temporal (I think there's a proposal to have Clock extend Monad)

I reiterate, we should make `Clock` extend nothing, and avoid this breakage (of dubious value) downstream. cc @djspiewak

djspiewak:
Yeah I'm modestly convinced Clock should be be its own thing. It kinda sucks because we lose the law, but it's only one law and it doesn't matter that much in the end.

RaasAhsan:
@rossabaker sorry for losing track of this, did anyone end up getting this done? If not, I'll try to wrap things up. IIRC I got blocked because cats-effect-testing wasn't on CE3 yet

rossabaker:
No, this is still the latest and greatest.

If you're porting tests, note that a merge of #3983 will be headed this way.  A lot of what was `Http4sLegacyMatchers` is now ported to MUnit.  If the relevant tests don't extend that, it should be fine in the coming merge.

We still owe you a good review here, too.

yanns:
> @rossabaker sorry for losing track of this, did anyone end up getting this done? If not, I'll try to wrap things up. IIRC I got blocked because cats-effect-testing wasn't on CE3 yet

for info, I created a temporary trait for cats-effect-testing on https://github.com/http4s/http4s/pull/3894
And the change is likely to come quickly: https://github.com/djspiewak/cats-effect-testing/issues/117

In general, I'm ready to help here. What do you need?

RaasAhsan:
@yanns the `core`, `client` and `server` modules have been ported over, it's really just the tests that are left. I'll check out the PR and see if the trait can be used here

yanns:
@RaasAhsan great to hear. Tell me if you need help.

RaasAhsan:
Looks like we've got some dependency eviction issues on `vault` via https://github.com/ChristopherDavenport/vault/issues/159, which still depends on CE2

yanns:
For info, I've opened a PR to use version of cats-effect-testing supporting ec3: https://github.com/http4s/http4s/pull/4000

RaasAhsan:
Just a heads up, we're hitting `Sync + MonadCancel` constraint ambiguity now. I'm going to promote to `Async` until this is resolved (probably CE3-M6)

RaasAhsan:
Alright, server and client are compiling now and I've left comments where constraints need to be adjusted. A handful of tests are failing still, which I'll get to tomorrow. And also this unpleasant error:
```
$ sbt scalafmtAll
[info] Formatting 56 Scala sources...
[error] Search state exploded on '=>∙release[6710:6738]', line 124: /Users/raasahsan/Desktop/projects/http4s/server/src/main/scala/org/http4s/server/middleware/BracketRequestResponse.scala
[error] (server / Compile / scalafmt) Search state exploded on '=>∙release[6710:6738]', line 124: /Users/raasahsan/Desktop/projects/http4s/server/src/main/scala/org/http4s/server/middleware/BracketRequestResponse.scala
[error] Total time: 6 s, completed Dec 21, 2020, 2:39:46 AM
```

RaasAhsan:
All the server/client tests should be fixed now, and I also upgraded FS2 and CE3 dependencies. There's one failing test that depends on https://github.com/typelevel/cats-effect/issues/1535 so I've marked those as ignore until CE3-M6. Otherwise, there are just a bunch of constraints we need to fix wrt `Sync + MonadCancel`.

RaasAhsan:
@rossabaker I think this is ready for review. I've given up on trying to get the build to pass, I stopped at the `unusedDependencyCheck`. Let me know if you want me to undo everything before then.

client/src/main/scala/org/http4s/client/middleware/FollowRedirect.scala
```diff
@@ -94,23 +94,25 @@ object FollowRedirect {
...
-            case _ =>
-              // IF the response is missing the Location header, OR there is no method to redirect,
-              // OR we have exceeded max number of redirections, THEN we redirect no more
-              Resource.make(resp.pure[F])(_ => dispose).pure[F]
-          }
-
-        case Left(e) =>
-          F.raiseError(e)
+      F.uncancelable { _ =>
+        client.run(req).allocated.attempt.flatMap {
```
rossabaker, line 98: I never liked this `.allocated`, but I think it's still necessary.  We need to dispose of the connection before allocating another, or else we can eat up all the connections.

I wonder whether we could extract some sort of "resource loop" combinator to cats-effect.  This was a hard thing to get right.

client/src/main/scala/org/http4s/client/middleware/GZip.scala
```diff
@@ -58,15 +59,15 @@ object GZip {
         response.withBodyStream(response.body.through(decompressWith(gunzip)))
 
       case Some(header) if header.contentCoding == ContentCoding.deflate =>
-        val deflate: Pipe[F, Byte, Byte] = fs2.compression.deflate(bufferSize)
+        val deflate: Pipe[F, Byte, Byte] = fs2.compression.deflate(DeflateParams(bufferSize))
```
rossabaker, line 62: Maybe we should surface the `DeflateParams` at the top level to control more than buffer size?

server/src/main/scala/org/http4s/server/package.scala
```diff
@@ -52,7 +52,7 @@ package object server {
 
   object ServerRequestKeys {
     val SecureSession: Key[Option[SecureSession]] =
-      Key.newKey[IO, Option[SecureSession]].unsafeRunSync()
+      Key.newKey[SyncIO, Option[SecureSession]].unsafeRunSync()
```
rossabaker, line 55: We perhaps should also have used this in cats-effect-2, but probably not worth going back to correct.

rossabaker (commented):
Incredbile work. I guess my biggest concern is probably what we do about blocking.  It seems like it just got waved away, and I don't fully understand how it works in CE3.

client/src/main/scala/org/http4s/client/JavaNetClientBuilder.scala
```diff
@@ -130,17 +118,18 @@ sealed abstract class JavaNetClientBuilder[F[_]] private (
           })
           _ <- F.delay(conn.setInstanceFollowRedirects(false))
           _ <- F.delay(conn.setDoInput(true))
-          resp <- blocker.blockOn(blocking(fetchResponse(req, conn)))
+          // TODO: fix the blocking here
```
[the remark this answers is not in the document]
RaasAhsan: We've got an `Async` constraint here now, so we can just call `blocking` yeah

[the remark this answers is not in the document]
RaasAhsan: oh, the problem is that `fetchResponse` is an effect rather than a thunk. So either we wrap precisely the blocking action with `blocking`, or we have to use `evalOn` plus a separate blocking EC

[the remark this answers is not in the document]
RaasAhsan: Ok, so I think this is the blocking code:
```scala
req.body
              .through(writeOutputStream(F.delay(conn.getOutputStream), false))
              .compile
              .drain
```
fs2 will actually shift to the internal blocking pool for `writeOutputStream`, so the question is whether that's enough, or do we actually want to run the entire `writeResponse` effect on a blocking pool

client/src/main/scala/org/http4s/client/middleware/GZip.scala
RaasAhsan: The only trouble here is that the choice of zlib/gzip are also selected via `DeflateParams`, so we'd want it to be `zlib` here specifically. We could overwrite but I don't know if that's too dirty..

client/src/main/scala/org/http4s/client/middleware/FollowRedirect.scala
RaasAhsan: I'll take a closer look and see if there's an obvious refactoring now that the majority of the work is done :)

RaasAhsan: Check it out now, it uses the `Hotswap` type which does exactly what you said -- linearly allocate a sequence of resources, ensuring each one is deallocated before the next is allocated.

client/src/main/scala/org/http4s/client/JavaNetClientBuilder.scala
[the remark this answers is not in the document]
rossabaker: `conn.getResponseCode` also blocks on a `java.io.InputStream` awaiting the response line.  I can't find it from the source, but I also assume that `conn.getHeaderFields` does as well.  The JDK8 client is synchronous, making it a bit of an outlier in our ecosystem.

I didn't realize fs2 had an internal blocking pool.  I guess we could create one as well for API simplicity, or pass one in for user control?

client/src/main/scala/org/http4s/client/middleware/FollowRedirect.scala
rossabaker: Oh, yes, that's very nice!  There's a similar loop in `Retry` that could get the same treatment, I think.

client/src/main/scala/org/http4s/client/middleware/GZip.scala
rossabaker: Oh.  Maybe we deprecate the old one and just have a more general `Compression` middleware?

client/src/main/scala/org/http4s/client/JavaNetClientBuilder.scala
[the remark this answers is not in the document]
rossabaker: `conn.getResponseCode` also blocks on a `java.io.InputStream` awaiting the response line.  I can't find it from the source, but I also assume that `conn.getHeaderFields` does as well.  The JDK8 client is synchronous, making it a bit of an outlier in our ecosystem.

I didn't realize fs2 had an internal blocking pool.  I guess we could create one as well for API simplicity, or pass one in for user control?

client/src/main/scala/org/http4s/client/middleware/GZip.scala
RaasAhsan: I'll track this in another issue to think about later :)

client/src/main/scala/org/http4s/client/JavaNetClientBuilder.scala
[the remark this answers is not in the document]
RaasAhsan: ah I wasn't clear about this -- CE3 exposes an internal blocking pool that is accessed via the signature:
```
trait Sync[F[_]] {
  def blocking[A](thunk: => A): F[A]
}
```
Fs2 uses that for its various blocking ops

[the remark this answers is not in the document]
RaasAhsan: Alright I've made those two calls `blocking`, but there are a couple of other ones that I'm concerned about that we should look into more, particularly this whole method:
```scala
private def writeBody(req: Request[F], conn: HttpURLConnection): F[Unit] =
    if (req.isChunked)
      F.delay(conn.setDoOutput(true)) *>
        F.delay(conn.setChunkedStreamingMode(4096)) *>
        req.body
          .through(writeOutputStream(F.delay(conn.getOutputStream), false))
          .compile
          .drain
    else
      req.contentLength match {
        case Some(len) if len >= 0L =>
          F.delay(conn.setDoOutput(true)) *>
            F.delay(conn.setFixedLengthStreamingMode(len)) *>
            req.body
              .through(writeOutputStream(F.delay(conn.getOutputStream), false))
              .compile
              .drain
        case _ =>
          F.delay(conn.setDoOutput(false))
      }
```
going to open a follow up issue to track. (I hope you don't mind me creating all these follow ups, I think once this is merged a lot of other work will be unblocked :) )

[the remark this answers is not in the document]
rossabaker: No, I think it's great.  It's a chaotic time between the dotty branch and this branch and I'm afraid of dropping things.

As noted in #4066, I think `fs2` is taking care of the blocking calls here, buried in `writeOutputStream`.

rossabaker:
I see a bunch of compile errors in Ember.  I think perhaps an fs2 upgrade got merged in?

RaasAhsan:
Fixed the ember-core errors plus the merge conflicts

rossabaker:
I felt bad for creating another one for you as you were fixing the last one, so I took care of scala-xml. :smile:

RaasAhsan:
There was one more compile error in scala-xml so I fixed that :)

rossabaker:
That'd be great.  I merged that last one from the GitHub editor.  We get red Xs on this branch at least because of docs, and red Xs on dotty because it's not close to done, and I've been a bit sloppy checking what the build really says as I go to and fro. :grimacing:

.github/workflows/ci.yml
```diff
@@ -67,9 +67,6 @@ jobs:
       - name: Check binary compatibility
         run: sbt ++${{ matrix.scala }} mimaReportBinaryIssues
 
-      - name: Check unused dependencies
-        run: sbt ++${{ matrix.scala }} unusedCompileDependenciesTest
-
```
rossabaker, line 72: I'd like to keep this one.  It has saved us from some bugs in the past.  (In fact, I'd like to adopt `undeclaredCompileDependenciesTest`, which is verbose, but would have saved us from a cursed release recently.)

rossabaker: I'll check out this branch and make the task pass.

rossabaker: https://github.com/RaasAhsan/http4s/pull/1

RaasAhsan: Thanks, there are some flaky tests that still need to be fixed, taking a look at those

server/src/test/scala/org/http4s/server/middleware/ResponseTimingSuite.scala
```diff
@@ -31,21 +43,24 @@ class ResponseTimingSuite extends Http4sSuite {
...
-      .assertEquals(true)
+      .map(_.headers.find(_.name == CIString("X-Response-Time")).map(_.value.toInt))
+
+    header.assertEquals(Some(artificialDelay))
   }
 }
 
 object Sys {
-  private val currentTime: Ref[IO, Long] = Ref.unsafe[IO, Long](System.currentTimeMillis())
+  private val currentTime: Ref[IO, Long] = Ref.unsafe[IO, Long](0L)
```
RaasAhsan, line 53: CE timers now return a `FiniteDuration`, so we have to call `toUnit`, which returns a `Double`. Converting that to a `Long` will sometimes lose precision and round down.. so I'm just fixing it to a value that works

RaasAhsan:
All but the first two tests in `RetrySuite` is failing on CI (and locally, when running the entire test suite) with this error:
```
==> X org.http4s.client.middleware.RetrySuite.default retriable should defaultRetriable does not resubmit bodies on idempotent methods  0.008s java.lang.IllegalArgumentException: requirement failed
    at scala.Predef$.require(Predef.scala:268)
    at scala.concurrent.BatchingExecutor$Batch.run(BatchingExecutor.scala:55)
    at scala.concurrent.Future$InternalCallbackExecutor$.unbatchedExecute(Future.scala:875)
    at scala.concurrent.BatchingExecutor$Batch.blockOn(BatchingExecutor.scala:96)
    at scala.concurrent.Await$.result(package.scala:146)
    at munit.internal.PlatformCompat$.$anonfun$waitAtMost$1(PlatformCompat.scala:21)
    at scala.util.Try$.apply(Try.scala:213)
    at munit.internal.PlatformCompat$.waitAtMost(PlatformCompat.scala:21)
    at munit.FunSuite.waitForCompletion(FunSuite.scala:51)
    at munit.FunSuite.$anonfun$test$1(FunSuite.scala:37)
    at munit.GenericTest.$anonfun$withBodyMap$1(GenericTest.scala:33)
    at munit.GenericTest.$anonfun$withBodyMap$1(GenericTest.scala:33)
    at munit.MUnitRunner.$anonfun$runTestBody$1(MUnitRunner.scala:290)
```
Seems like a bug with starting the test up..? I'm afraid I don't immediately know how to fix this one

rossabaker:
Hmm. This passes for me, on bca4a43:

```
client/testOnly *RetrySuite
```

rossabaker:
The test names have gotten manged through all the merges, perhaps.

rossabaker:
If you don't like that warning becoming fatal on CI, complain to @djspiewak. :laughing:

RaasAhsan:
Yeah the test suite passes in isolation locally, seems to fail when the entire CI suite runs. Seems like the property based test might be mucking with the pool or something

rossabaker:
Hmm.  Even `test` is working for me.

My theory on the flaky CI tests has been that it's some combination of the `unsafeRunSync()` calls and the ridiculously CPU-intensive property tests, either of which can lock up threads for longer than the flaky tests wait.  Carlos has been doing heroic work cleaning up the `unsafeRunSync()`s.

Most of the flaky tests are in blaze, with async-http-client and ember making occasional appearances.  This one is not one that I recall seeing.

RaasAhsan:
Yeah, it might be some weird interaction with the new CE3 stuff. I'll open an issue for it to look at later, because the build is passing now :P

rossabaker:
It looks like my PR to this PR failed.  I'm content to merge this, and take another shot at mine after.  This is already large enough and definitely forward progress.

yanns:
> Yeah the test suite passes in isolation locally, seems to fail when the entire CI suite runs. Seems like the property based test might be mucking with the pool or something

I had a similar issue on https://github.com/http4s/http4s/pull/3894 created by blocking code.
See how I checked that https://github.com/http4s/http4s/pull/3894#issuecomment-739488014

PR #3862, opened 2020-11-14
WIP: Dotty support

rossabaker:
Dotty support is its [own project](https://github.com/http4s/http4s/projects/5), but this gives people a place to comment on it as it goes.

TBD whether this merges into series/0.21 or becomes series/0.22.


rossabaker:
This work is merged and now on series/0.22.

PR #3863, opened 2020-11-14
WIP: Cats-Effect-3

rossabaker:
The Cats-Effect-3 upgrade is its [own project](https://github.com/http4s/http4s/projects/4), but this gives people a place to comment on it as it goes.

This will merge back to main and be part of the 1.0 release.


rossabaker:
This work is being moved to main, and partially merged with main on cats-effect-3-dotty.  We will finish the merge over there.

PR #3865, opened 2020-11-14
EntityDecoder constraints loosened

domaspoliakas:
As per the discussion on gitter.

I also have a separate branch where I went over other `Sync` usages in core, but I think that might need some discussion, so for the sake of expediency I'm PRing this separately first.


rossabaker (approved):
:tada:

PR #3866, opened 2020-11-14
upgrading sbt-tpolecat to latest (v0.1.15)

### LINKED ISSUE #3830: Enable kind projector for Dotty
Dotty requires `-Ykind-projector`.  Maybe sbt-tpolecat does this, maybe we need to do it.

Target the dotty branch if you use isDotty.value, or target series/0.21 if you don't.

jyoo980:
Should resolve #3830


jyoo980:
Selected wrong branch to merge against. Closing.

PR #3868, opened 2020-11-15
use sbt-github-actions

amesgen:
Supersedes and closes #3825

See #3847 for more discussion.


amesgen:
CI does not trigger *again*, I (still) have no idea why. The CI output for this commit is here: https://github.com/amesgen/http4s/actions/runs/363811493

.mergify.yml
```diff
@@ -3,15 +3,15 @@ pull_request_rules:
...
-      - status-success=Scala 2.13.2, Java adopt@1.11
-      - status-success=Scalafix
-      - status-success=Project Site
-      - status-success=Doc Site
+      - status-success=Build and Test (ubuntu-latest, 2.13.3, adopt@1.8)
+      - status-success=Build and Test (ubuntu-latest, 2.13.3, adopt@1.11)
+      - status-success=Build and Test (ubuntu-latest, 2.13.3, adopt@1.15)
+      - status-success=Build and Test (ubuntu-latest, 2.12.12, adopt@1.8)
+      - status-success=Build and Test (ubuntu-latest, 2.12.12, adopt@1.11)
+      - status-success=Build and Test (ubuntu-latest, 2.12.12, adopt@1.15)
```
rossabaker, line 12: We haven't historically blocked mergify on 1.15, which we're testing just to get out ahead of any problems. But I don't mind adding it as a gate.

rossabaker (approved):
Very nice work!

.mergify.yml
amesgen: Ah, didn't know that. If any annoying problems occur on Java 15, we can remove disable it again. :+1:

rossabaker:
I'm curious whether subsequent PRs get tested after we merge this.

rossabaker:
I expect we'll have a couple annoying issues coming out of this, but it's definitely the way forward, and the status quo falls apart tomorrow.

PR #3870, opened 2020-11-15
Update deprecated actions in release.yml

rossabaker:
Still have some deprecated calls to add-path, which will begin failing tomorrow.


rossabaker:
This file could maybe be managed by GitHub Actions, too? 

@amesgen's PR seems to be working well on the other workflow. This one just didn't get tickled until it was merged to a release branch.

amesgen (changes requested):
Releasing is already managed by sbt-github-actions, I just forgot to delete release.yml.

 - [sbt-github-actions config for publishing](https://github.com/http4s/http4s/blob/11cdbd3095bf0034c362f985d020e85e0c13c9fd/project/Http4sPlugin.scala#L286-L305)
 - [release job](https://github.com/http4s/http4s/blob/11cdbd3095bf0034c362f985d020e85e0c13c9fd/.github/workflows/ci.yml#L103-L202)

I would suggest to simply delete release.yml in this PR, unless I am overlooking sth.

amesgen:
One thing that we might want to adapt is the set of branches where we want to publish snapshots: [config](https://github.com/http4s/http4s/blob/11cdbd3095bf0034c362f985d020e85e0c13c9fd/project/Http4sPlugin.scala#L286-L289)

rossabaker:
Ah, that's where it is!  I thought you'd set that up, but my search came up empty.

And :+1: to tweaking the snapshots targets.

rossabaker:
> One thing that we might want to adapt is the set of branches where we want to publish snapshots

We used to have an environment variable so it would trigger `publishSigned` instead of `ci-release`: `CI_SNAPSHOT_RELEASE: +publishSigned`.  I think that might be necessary, too?  We'll find out soon, because I merged to main.

amesgen:
Ah yes, one can do this by adding this env var [here](https://github.com/http4s/http4s/blob/11cdbd3095bf0034c362f985d020e85e0c13c9fd/project/Http4sPlugin.scala#L299).

rossabaker:
It might not have mattered? Looks like it succesfully published a snapshot on `main` without that environment: https://github.com/http4s/http4s/runs/1403222551?check_suite_focus=true

amesgen:
This env var seems to enable signing, see the documentation for `CI_SNAPSHOT_RELEASE` here: https://github.com/olafurpg/sbt-ci-release/blob/master/readme.md . I don't know how important that is for http4s.

amesgen:
BTW: The CI workflow you linked (https://github.com/http4s/http4s/runs/1403222551?check_suite_focus=true) is from release.yml, which has not yet been deleted. The publishing job from ci.yml (which is managed by sbt-github-actions) is here: https://github.com/http4s/http4s/runs/1403257157

rossabaker:
Oh, yeah. Signing snapshots isn't that critical (I don't even know if anybody uses them), but it's a nice thing to do if we publish them.

And you're right about the branches.  So many moving parts.  I think this is now good, and can merge to main?

amesgen:
Yes, this should be ready. We could additional branches for snapshot publishing in this PR [here](https://github.com/http4s/http4s/blob/11cdbd3095bf0034c362f985d020e85e0c13c9fd/project/Http4sPlugin.scala#L286-L289) (as discussed above), but this can wait.

amesgen (approved):

Issue #3839, opened 2020-11-11
Port jawn to cats-effect-3


rossabaker:

yanns:
I can take over

rossabaker:
Fixed by #3871.

PR #3871, opened 2020-11-15
jawn module for cats-effect 3

yanns:
Fixes https://github.com/http4s/http4s/issues/3839

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -28,7 +28,7 @@ lazy val modules: List[ProjectReference] = List(
   // jetty,
   // tomcat,
   // theDsl,
-  // jawn,
+   jawn,
   // argonaut,
   // boopickle,
   // circe,
--- a/jawn/src/main/scala/org/http4s/jawn/JawnInstances.scala
+++ b/jawn/src/main/scala/org/http4s/jawn/JawnInstances.scala
@@ -14,7 +14,7 @@ import org.typelevel.jawn.{AsyncParser, Facade, ParseException}
 import jawnfs2._
 
 trait JawnInstances {
-  def jawnDecoder[F[_]: Sync, J: Facade]: EntityDecoder[F, J] =
+  def jawnDecoder[F[_]: Concurrent, J: Facade]: EntityDecoder[F, J] =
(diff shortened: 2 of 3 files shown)


rossabaker:
Thanks! This is the right start. I think this has a couple subtleties that some similar tickets didn't:

1. It depends on jawn-fs2, which needs to be published for cats-effect-3 and fs2-3.  @aeons did the upgrade, but we never released.  I'll work on that and report back.

2. This one has decoders, which can now take a `Concurrent` effect instead of `Sync`.  This will be a big improvement for people using tracing or other kleislis as their base effect.

rossabaker:
I should be able to release typelevel/jawn-fs2 after https://github.com/typelevel/jawn-fs2/pull/151. It's bad to change the publishing organization without changing the root package.

This will require a change to the dependency in Http4sPlugin, and probably an import change in this module.

yanns:
>     1. It depends on jawn-fs2, which needs to be published for cats-effect-3 and fs2-3.  @aeons did the upgrade, but we never released.  I'll work on that and report back.

OK, I'll wait for the release.

>     2. This one has decoders, which can now take a `Concurrent` effect instead of `Sync`.  This will be a big improvement for people using tracing or other kleislis as their base effect.

https://github.com/http4s/http4s/pull/3871/commits/ed6feb48a272013f8c54497f47b4ea19c3787bf9

rossabaker:
I just released jawn-fs2-2.0.0-M2.  It's now published under `org.typelevel` instead of `org.http4s`, and all Scala imports change from `jawnfs2` to `org.typelevel.jawn.fs2`.

yanns:
The builds are failing.
My quick check if that we would need https://github.com/http4s/http4s/pull/3868 on this branch.
@rossabaker what should we do?
Should I try to cherry-pick https://github.com/http4s/http4s/pull/3868 on the `cats-effect-3` branch?
Or do we ignore the CI checks for now?
(`sbt test` is running locally)

rossabaker:
It looks right, it passes locally, we know why it failed on CI, and it's not yet a production branch.  Let's take the easy way out and merge it. There's time to fix it if we guessed wrong.

Thanks!

PR #3880, opened 2020-11-19
Declare more dependencies

rossabaker:
Original proposal:

This is the other half of the sbt-explicit-dependencies plugin.

Advantages:
1. It would have caught our last security alert.
2. It pushes us to discover certain. dependency diamonds before users do.
3. It can inform better module boundaries.

Disadvantages:
1. We'll hear slightly more from Scala Steward
2. It's one more thing for new contributors to stub their toes on

On balance, I like it.

---

Less ambitious version:

1. Doesn't repeat dependencies inherited from our own projects. This drastically reduces the noise.
2. We don't care about internals like cats-kernel or the unsuccessful json4s-ast. Trim those, too.
3. Keep the modules this did discover, where they're introduced.
4. Alas, undeclaredDependenciesTest no longer passes, and is removed from the build.


hamnis (approved):

rossabaker:
I know some people don't like this. I'm willing to be talked out of it before I merge.

rossabaker:
This is unlikely to make it into sbt-spiewak, and I'm trying to push as much of our build into that as possible, so we can benefit from that "community configuration" and so contributors to other Typelevel projects hit less surprises.  I think I'm now a sad :-1: on this one, though there are a couple learnings from it I'd like to adopt.

rossabaker:
I've updated the description to this more modest proposal.  Less enforced, but less noisy and captures the upsides.

ChristopherDavenport (approved):

PR #3881, opened 2020-11-19
Move default route to the end

ashwinbhaskar:
Currently, the default route (matches against `_`) is present on the top. People new to Scala who are trying out the example service might find it perplexing that their valid routes are being returned with `MethodNotAllowed` ? 
I moved it to the end.


ashwinbhaskar:
Closing this PR as I mistook the default match against `https method` to be the default match against path

PR #3885, opened 2020-11-19
migrate json4s, json4s-native, json4s-jackson to cats-effect 3

yanns:
I'd like to enable some tests to check https://github.com/http4s/http4s/pull/3871

Now that we are using `Concurrent` instead of `Sync`, we cannot use `F.delay` anymore.
I'm not sure that would be the right approach here.

I've started with `pure`.
As the parsing of a json can take CPU time, should we signal it that some way so that maybe a new thread should be used for it? Could this block a fiber?

Should we use something different from `F.pure()`? `F.unit.as()`?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -34,7 +34,7 @@ lazy val modules: List[ProjectReference] = List(
   // circe,
    json4s,
    json4sNative,
-  // json4sJackson,
+   json4sJackson,
   // playJson,
   // scalaXml,
   twirl,
--- a/build.sbt
+++ b/build.sbt
@@ -32,7 +32,7 @@ lazy val modules: List[ProjectReference] = List(
   // argonaut,
   // boopickle,
   // circe,
-  // json4s,
+   json4s,
    json4sNative,
   // json4sJackson,
   // playJson,
--- a/build.sbt
+++ b/build.sbt
@@ -33,7 +33,7 @@ lazy val modules: List[ProjectReference] = List(
   // boopickle,
   // circe,
   // json4s,
-  // json4sNative,
+   json4sNative,
   // json4sJackson,
   // playJson,
   // scalaXml,
--- a/json4s/src/main/scala/org/http4s/json4s/Json4sInstances.scala
+++ b/json4s/src/main/scala/org/http4s/json4s/Json4sInstances.scala
@@ -7,27 +7,28 @@
 package org.http4s
 package json4s
 
-import cats.effect.Sync
+import cats.effect.Concurrent
 import cats.implicits._
 import org.http4s.headers.`Content-Type`
 import org.json4s._
 import org.json4s.JsonAST.JValue
 import org.typelevel.jawn.support.json4s.Parser
 
+import scala.util.Try
+
 object CustomParser extends Parser(useBigDecimalForDouble = true, useBigIntForLong = true)
 
 trait Json4sInstances[J] {
-  implicit def jsonDecoder[F[_]](implicit F: Sync[F]): EntityDecoder[F, JValue] =
+  implicit def jsonDecoder[F[_]](implicit F: Concurrent[F]): EntityDecoder[F, JValue] =
     jawn.jawnDecoder(F, CustomParser.facade)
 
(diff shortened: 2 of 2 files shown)


json4s/src/main/scala/org/http4s/json4s/Json4sInstances.scala
```diff
@@ -7,27 +7,28 @@
...
       DecodeResult(
-        F.delay(reader.read(json))
-          .map[Either[DecodeFailure, A]](Right(_))
-          .recover { case e: MappingException =>
-            Left(InvalidMessageBodyFailure("Could not map JSON", Some(e)))
-          })
+        F.pure(
+          Try(reader.read(json)).toEither
+            .leftMap[DecodeFailure](e => InvalidMessageBodyFailure("Could not map JSON", Some(e)))
+        ))
```
rossabaker, line 31: Yes, `F.pure` is correct here.  I think the old `delay` was capturing exceptions.

This could be microoptimized and avoid the allocation of a `Try` with a regular old `try`/`catch` expression.

rossabaker:
I wouldn't flag things that take CPU time.  We want this executing on one of the CPU bound pools, sized to one per core.  It's blocking IO that we want to shift to blocking pools, where it's more typical to use a cached (or at least larger) thread pool for things not contending for the CPU.

json4s/src/main/scala/org/http4s/json4s/Json4sInstances.scala
yanns: Thx for the feedback.
I was not sure if a good old `try` / `catch` block was an ok-style in https.
I'll change that.

yanns: https://github.com/http4s/http4s/pull/3885/commits/1497f7d75faf92e8bb59799ca9f938e832af9d7f

build.sbt
```diff
@@ -33,7 +33,7 @@ lazy val modules: List[ProjectReference] = List(
   // boopickle,
   // circe,
   // json4s,
```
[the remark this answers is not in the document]
yanns: https://github.com/http4s/http4s/pull/3885/commits/856252f1307d7240a773145db554f9939e57cee7

yanns:
I've also activated `json4sjackson` that does not need any further changes: https://github.com/http4s/http4s/pull/3885/commits/d4f23074c302e653b2f8437fd20dba7c3df6fa57

json4s/src/main/scala/org/http4s/json4s/Json4sInstances.scala
```diff
@@ -17,17 +17,19 @@ import org.typelevel.jawn.support.json4s.Parser
...
       DecodeResult(
-        F.delay(reader.read(json))
-          .map[Either[DecodeFailure, A]](Right(_))
-          .recover { case e: MappingException =>
-            Left(InvalidMessageBodyFailure("Could not map JSON", Some(e)))
-          })
+        F.pure {
+          try Right(reader.read(json))
+          catch {
+            case e: Exception => Left(InvalidMessageBodyFailure("Could not map JSON", Some(e)))
```
rossabaker, line 29: I'd go with `NonFatal(e)` instead of `e: Exception` here.

rossabaker (commented):
I don't think try/catch is bad when wrapping a library that fails by throwing.  I think it's bad when _we_ fail by throwing.

If you're just wrapping success in `Right` and failure in `Left`, there is `Either.catchNonFatal` with `import cats.implicits._` that would hide the try and catch.  You could use that here, but then you'd need to `leftMap` it.  I think what you've done is good.

json4s/src/main/scala/org/http4s/json4s/Json4sInstances.scala
yanns: https://github.com/http4s/http4s/pull/3885/commits/44c0b8f9fcf4f161f584356a7b21a3df4c0a8590

rossabaker (approved):

PR #3886, opened 2020-11-19
Loosened some constraints in core

domaspoliakas:
Still using `Sync` given the changes
- `EntityEncoder`: `inputStreamEncoder` (uses `readInputStream`), `readerEncoder`
- `StaticFile`: `fromResource`, `fromURL`
- `org.http4s.internal.package.scala`: `CompletionStage` interop

Some notes on the changes:
- Changed `HttpApp` to `Defer` from `Sync`. My gut says that that's not right, but it's possible so I did it 🤷‍♂️ 
- `KleisliSyntax` has the sacred words `BinCompat` - not sure if that means that _no_ changes should be done there or something else in this case (my binary compatibility understanding is admittedly lackluster)

### DIFF
--- a/core/src/main/scala/org/http4s/HttpApp.scala
+++ b/core/src/main/scala/org/http4s/HttpApp.scala
@@ -8,7 +8,7 @@ package org.http4s
 
 import cats.Applicative
 import cats.data.Kleisli
-import cats.effect.Sync
+import cats.Defer
 
 /** Functions for creating [[HttpApp]] kleislis. */
 object HttpApp {
@@ -21,7 +21,7 @@ object HttpApp {
     * @param run the function to lift
     * @return an [[HttpApp]] that wraps `run`
     */
-  def apply[F[_]: Sync](run: Request[F] => F[Response[F]]): HttpApp[F] =
+  def apply[F[_]: Defer](run: Request[F] => F[Response[F]]): HttpApp[F] =
     Http(run)
 
   /** Lifts an effectful [[Response]] into an [[HttpApp]].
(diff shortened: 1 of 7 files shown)


core/src/main/scala/org/http4s/HttpApp.scala
```diff
@@ -21,7 +21,7 @@ object HttpApp {
     * @param run the function to lift
     * @return an [[HttpApp]] that wraps `run`
     */
-  def apply[F[_]: Sync](run: Request[F] => F[Response[F]]): HttpApp[F] =
+  def apply[F[_]: Defer](run: Request[F] => F[Response[F]]): HttpApp[F] =
```
rossabaker, line 24: Prior discussion on https://github.com/http4s/http4s/pull/2325. @SystemFw and @kubukoz may have evolved their thinking in a cats-effect-3 world.

core/src/main/scala/org/http4s/syntax/KleisliSyntax.scala
```diff
@@ -19,16 +19,16 @@ trait KleisliSyntax {
 }
 
 trait KleisliSyntaxBinCompat0 {
```
rossabaker, line 21: Binary compatibility resets in 1.0, so we don't need any hacks.  We should be able to just flatten this.

core/src/main/scala/org/http4s/HttpApp.scala
rossabaker: I guess we already did this on `Http`: https://github.com/http4s/http4s/pull/2906

@ChristopherDavenport, are you on board with the rest of this, then?

PR #3891, opened 2020-11-19
Fix Streamed Json Encoding

### LINKED ISSUE #3890: Json Streaming is slow
We actually encode the json in our streaming encoder using a bunch of very small chunks. Since this is going through chunked rather than Content-Length. This means backends will flush buffers on every single chunk. Resulting in buffer flushes for (N * 2) the number of elements in the stream.

Likely we will want to follow this logic via rechunking with `chunkMin` or interpolating within a chunk to keep consistent chunking as the provided stream.

https://github.com/http4s/http4s/blob/main/circe/src/main/scala/org/http4s/circe/CirceInstances.scala#L126-L129

ChristopherDavenport:
Fixes #3890 

Questions: Is this a good chunk size? should we make it configurable in other versions? how does this relate to the benches for cutoff?

This should see a reduction in buffer flushes of 2000x

### DIFF
--- a/circe/src/main/scala/org/http4s/circe/CirceInstances.scala
+++ b/circe/src/main/scala/org/http4s/circe/CirceInstances.scala
@@ -12,7 +12,7 @@ import java.nio.ByteBuffer
 import cats.data.NonEmptyList
 import cats.effect.Sync
 import cats.syntax.either._
-import fs2.{Chunk, Stream}
+import fs2.{Chunk, Pull, Stream}
 import io.circe._
 import io.circe.jawn._
 import org.http4s.headers.`Content-Type`
@@ -119,13 +119,9 @@ trait CirceInstances extends JawnInstances {
     EntityEncoder
       .streamEncoder[F, Chunk[Byte]]
       .contramap[Stream[F, Json]] { stream =>
-        val jsons = stream.map(fromJsonToChunk(printer))
-        val badlyChunked = {
-          CirceInstances.openBrace ++ 
-          jsons.intersperse(CirceInstances.comma) ++ 
-          CirceInstances.closeBrace
-        }
-        badlyChunked.chunkMin(2048) // At minimum 1024 json objects per flush
+        Stream.emit(CirceInstances.openBrace) ++
+          stream.through(CirceInstance.streamedJsonWithCommas(printer)).chunks ++
+          Stream.emit(CirceInstances.closeBrace)
       }
       .withContentType(`Content-Type`(MediaType.application.json))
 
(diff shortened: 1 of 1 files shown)


diesalbla:
> Questions: Is this a good chunk size? should we make it configurable in other versions? how does this relate to the benches for cutoff?

Could or should this turned into a parameter of the functions, even at a later PR?

rossabaker:
Part of the untyped contract of encoders is that it flushes per chunk. It's the only way the user can control when partial results become available.

Instead of an arbitrary limit, could we intersperse each `Chunk[Json]` with a comma, and then intersperse those chunks as we do now?  It would still be slow if each Json were its own chunk, but reasonable, and easy to control, if the Jsons came in larger chunks.

ChristopherDavenport:
That's what the most recent commit is. Each incoming chunk gets written, interpolated with the commas, and then rechunked. 

The first chunk gets special treatment because it has different logic than the comma interpolation. Sorry, original PR description is now wrong with what the code implements.

As currently written outputs chunks
Chunk1 `[` and first json
RestOfChunk1 , and then json in the same as the input chunk
RestofChunks The same as RestOfChunk1 except for the very last one which does not get the `,`
ChunkContaining `]`, I really don't like this one, but don't have a good way to work it into the last chunk.

rossabaker (approved):
Oh, yeah, this looks different than what I looked at.  This is good, and should obviate the need for configuration (which was a good idea by @diesalbla in the old impl).

circe/src/main/scala/org/http4s/circe/CirceInstances.scala
```diff
@@ -218,11 +214,42 @@ object CirceInstances {
...
+                val interspersed = {
+                  val bldr = Vector.newBuilder[Chunk[Byte]]
+                  bldr.sizeHint(hd.size * 2)
+                  hd.foreach { o =>
+                    bldr += CirceInstances.comma
+                    bldr += fromJsonToChunk(printer)(o)
+                  }
+                  Chunk
+                    .vector(bldr.result())
+                    .flatMap(identity) // I know there must be a more efficient weay to do this
```
peterneyens, line 240: Maybe `Chunk.concatBytes(bldr.result())`?

circe/src/main/scala/org/http4s/circe/CirceInstances.scala
```diff
@@ -218,11 +214,40 @@ object CirceInstances {
...
+                    bldr += CirceInstances.comma
+                    bldr += fromJsonToChunk(printer)(o)
+                  }
+                  Chunk.concatBytes(bldr.result())
+                }
+                Pull.output(interspersed) >> Pull.pure(Some(tl))
+            }
+          }.pull
+            .echo
+    }.stream ++ Stream.chunk(closeBrace)
```
JesusMtnez, line 244: @ChristopherDavenport @rossabaker In the case of encoding an empty list using streams, I notice the `closeBrace` is printed always at then, but the `openBrace` is only printed if it has elements. If not, it will use `Pull.done` and the encoding will only be: `]`, right?

circe/src/main/scala/org/http4s/circe/CirceInstances.scala
```diff
@@ -218,11 +214,40 @@ object CirceInstances {
...
+
+  private final val openBrace: Chunk[Byte] =
```
kevinmeredith, line 246: Out of curiosity, what purpose does 'final' serve since the val lives in an object?

rossabaker: If I remember right, a final val in an object can be inlined only if it doesn't have a type annotation.  I don't know how it matters here, if at all.

Issue #3837, opened 2020-11-11
Port blaze-core to cats-effect-3


rossabaker:
This one is for the brave.  Will require a knowledge of Cats-Effect and Future.

yanns:
I think that we can put this task into progress.
There is a PR for it: https://github.com/http4s/http4s/pull/3894

rossabaker:
I still owe that one a review and will aim to get to it this weekend.

rossabaker:
Fixed by #3894.

PR #3894, opened 2020-11-20
Port blaze-core to cats-effect-3

yanns:
https://github.com/http4s/http4s/issues/3837

It's a WIP.
I'd like to receive some early feedback before attacking the websocket part.


blaze-core/src/main/scala/org/http4s/blazecore/Http1Stage.scala
```diff
@@ -215,6 +218,8 @@ private[http4s] trait Http1Stage[F[_]] { self: TailStage[ByteBuffer] =>
           }
         go()
       } else cb(End)
+      // TODO (YaSi): I don't it's right
+      F.pure(None)
```
yanns, line 222: I have difficulties understanding the new `F.async` function.
This compiles, but it does not seem right.

blaze-core/src/main/scala/org/http4s/blazecore/util/ChunkWriter.scala
```diff
@@ -49,13 +50,10 @@ private[util] object ChunkWriter {
...
-      case Right(buffer) =>
-        IO { promise.completeWith(pipe.channelWrite(buffer).map(Function.const(false))); () }
-      case Left(t) =>
-        IO { promise.failure(t); () }
-    }
-    promise.future
+    val result = f
+      .flatMap(buffer => F.blocking(pipe.channelWrite(buffer)))
+      .as(false)
+    D.unsafeToFuture(result)
```
yanns, line 56: Is it ok to stay in `Async` and to transform to `Future` only at the end?

blaze-core/src/main/scala/org/http4s/blazecore/util/package.scala
```diff
@@ -29,23 +27,7 @@ package object util {
...
-        case None =>
-          F.async { cb =>
-            future.onComplete {
-              case Success(a) => cb(Right(a))
-              case Failure(t) => cb(Left(t))
-            }(directec)
-          }
-      }
-    }
+    F.fromFuture(f)
```
yanns, line 31: I don't know if we need a special function for this now.

blaze-core/src/test/scala/org/http4s/blazecore/util/CatsEffect.scala
```diff
@@ -0,0 +1,34 @@
...
+package org.http4s.blazecore.util
+
+import cats.effect.std.Dispatcher
+import cats.effect.{Async, Resource, Sync}
+import org.specs2.execute.{AsResult, Result}
+
+import scala.concurrent.duration.{Duration, _}
+
+/**
+  * copy of [[cats.effect.testing.specs2.CatsEffect]] adapted to cats-effect 3
```
yanns, line 16: We need a migrated https://github.com/djspiewak/cats-effect-testing.

blaze-core/src/test/scala/org/http4s/blazecore/util/Http1WriterSpec.scala
```diff
@@ -9,19 +9,23 @@ package blazecore
...
+
+import cats.effect.std.Dispatcher
 import org.http4s.blaze.pipeline.{LeafBuilder, TailStage}
 import org.http4s.util.StringWriter
+
 import scala.concurrent.Future
+import scala.concurrent.ExecutionContext.Implicits._
 
 class Http1WriterSpec extends Http4sSpec with CatsEffect {
+  implicit val d: Dispatcher[IO] = Dispatcher[IO].allocated.unsafeRunSync()._1
```
yanns, line 27: This is WIP. We need to call the `_2` function after the tests.

blaze-core/src/test/scala/org/http4s/blazecore/util/Http1WriterSpec.scala
```diff
@@ -253,13 +257,13 @@ class Http1WriterSpec extends Http4sSpec with CatsEffect {
...
-      val p = resource.through(deflate())
-      (p.compile.toVector.map(_.toArray), DumpingWriter.dump(resource.through(deflate())))
+      val p = resource.through(deflate(DeflateParams.DEFAULT))
+      (p.compile.toVector.map(_.toArray), DumpingWriter.dump(resource.through(deflate(DeflateParams.DEFAULT))))
         .mapN(_ === _)
     }
 
     "must be stack safe" in {
-      val p = repeatEval(IO.async[Byte](_(Right(0.toByte)))).take(300000)
+      val p = repeatEval(IO.pure[Byte](0.toByte)).take(300000)
```
yanns, line 266: Not sure if this is right.

blaze-core/src/main/scala/org/http4s/blazecore/Http1Stage.scala
SystemFw: Use `async_` for a drop in replacement to the old signature

blaze-core/src/main/scala/org/http4s/blazecore/util/ChunkWriter.scala
SystemFw: why is that in `blocking`? that would cause a hop and hop back to/from the blocking pool. Also, `channelWrite` is a `Future`, so that's something odd there, you want `fromFuture`

> Is it ok to stay in Async and to transform to Future only at the end?

they end up running in different thread pools (old and new diff), but I don't think that makes a huge difference. Transforming from `IO` to `Future` and back a couple times seems pointless though.

Why not
```
f.unsafeToFuture.flatMap(pipe).as(false)
```

blaze-core/src/main/scala/org/http4s/blazecore/util/package.scala
SystemFw: We do, the comment  in the old diff is still relevant, you don't want to shift back onto the IO pool

SystemFw (commented):
Really not an expert in blaze, but left some comments here and there

blaze-core/src/main/scala/org/http4s/blazecore/util/ChunkWriter.scala
yanns: Good catch. I thought `pipe.channelWrite(buffer)`  was a blocking I/O operation.

blaze-core/src/main/scala/org/http4s/blazecore/Http1Stage.scala
yanns: https://github.com/http4s/http4s/pull/3894/commits/65216331c4c301102e105680ad66ee831e5cdfba

blaze-core/src/main/scala/org/http4s/blazecore/util/ChunkWriter.scala
yanns: https://github.com/http4s/http4s/pull/3894/commits/61c196efb9dede4454d59d5e3956afc4a543a096

blaze-core/src/main/scala/org/http4s/blazecore/util/package.scala
yanns: https://github.com/http4s/http4s/pull/3894/commits/aab6fbbed0d9554453d537a3f6be906818579aa3

yanns:
Thanks for the feedback @SystemFw !

blaze-core/src/main/scala/org/http4s/blazecore/util/Http2Writer.scala
```diff
@@ -8,45 +8,45 @@ package org.http4s
...
-    else {
-      val hs = headers
-      headers = null
-      tail.channelWrite(
-        HeadersFrame(Priority.NoPriority, endStream = false, hs)
-          :: DataFrame(endStream = false, chunk.toByteBuffer)
-          :: Nil)
-    }
-}
+//import cats.effect._
```
rossabaker, line 11: If you use block comments (`/*` and `*/`), you can keep the diff much smaller while you gradually get things compiling again.

blaze-core/src/test/scala/org/http4s/blazecore/util/CatsEffect.scala
```diff
@@ -0,0 +1,36 @@
...
+
+package org.http4s.blazecore.util
+
+import cats.effect.std.Dispatcher
+import cats.effect.{Async, Resource, Sync}
+import org.specs2.execute.{AsResult, Result}
+
+import scala.concurrent.duration.{Duration, _}
+
+/** copy of [[cats.effect.testing.specs2.CatsEffect]] adapted to cats-effect 3
```
rossabaker, line 15: This would be a good issue on the cats-effect-testing repo.

Until then, I think we could put it in our testing package and maybe use it in multiple modules.

blaze-core/src/test/scala/org/http4s/blazecore/util/Http1WriterSpec.scala
rossabaker: We have a `withResource` in `Http4sSpec` that I think would work.

blaze-core/src/test/scala/org/http4s/blazecore/util/Http1WriterSpec.scala
rossabaker: I think that's fine?  It's making 300000 chunks either way.

blaze-core/src/main/scala/org/http4s/blazecore/util/Http2Writer.scala
yanns: 👍 
https://github.com/http4s/http4s/commit/9257737ecd21faf1b484b3c7ebe59714b4d68cbc

blaze-core/src/test/scala/org/http4s/blazecore/util/Http1WriterSpec.scala
yanns: https://github.com/http4s/http4s/pull/3894/commits/1976730b781568a2ef3769e2a4fa936aecc8974d

yanns:
I'm not blocked by the compilation of the tests: https://github.com/http4s/http4s/pull/3894/commits/e9eeb9511897cca94b575cb1fd69d210547514c9

The 2 specs seems to wait for the same lock. The compilation never completes.

yanns:
Update: the issue seems to be only on `Http4sWSStageSpec`. When I comment `Http4sWSStageSpec` out, the tests can compile. If I comment only `Http1WriterSpec` out, the tests' compilation never completes.

blaze-core/src/main/scala/org/http4s/blazecore/websocket/Http4sWSStage.scala
```diff
@@ -152,25 +153,26 @@ private[http4s] class Http4sWSStage[F[_]](
         .compile
         .drain
 
-    unsafeRunAsync(wsStream) {
+    val result = F.attempt(wsStream).flatMap {
```
yanns, line 156: Current situation: `F.attempt(wsStream)` does not make any progress.
In the previous implementation (`unsafeRunAsync(wsStream)`), `wsStream` was "shift" to the execution context provided in the constructor.

rossabaker: I'm very far out of the CE3 loop right now.  Do we need a `start` or something similar?

yanns: The `Dispatcher` can start it. See at line 165: `D.unsafeRunSync(result)`.
When I add a timeout at this line, the timeout exception is coming from this place.

SystemFw: `unsafeRunSync` is blocking, so the semantics have changed here. Try `unsafeToFuture` instead of `unsafeRunSync`, or add `.start` at the end of `result`

SystemFw: also the `attempt` can become an `handleErrorWith`

yanns: The previous implementation is this one:
```
  private[http4s] def unsafeRunAsync[F[_], A](fa: F[A])(
      f: Either[Throwable, A] => IO[Unit])(implicit F: Effect[F], ec: ExecutionContext): Unit =
    F.runAsync(Async.shift(ec) *> fa)(f).unsafeRunSync()
```
Does it wait for the result?
Or does it "just" start the computation?

SystemFw: `runAsync` doesn't wait, that's why you need to pass the `f` callback, instead of just receiving an `Either[Throwable, A]` as a result (and that's why it's called `*async`), but it doesn't `.start` either (the IO will shift itself as soon as an async node is encountered, which is why we add an explicit `shift` node just before). 

The closest thing to that is `unsafeToFuture`.

yanns: OK I understand about the `F.runAsync`.
I still have questions about the `.unsafeRunSync()`, if this waits for the start.
If yes `unsafeRunAsync` starts the side-effect on another thread pool right?
As `unsafeToFuture` returns a `Future`, should we wait for the `Future` with an `Await.result` or something similar? In that case, we need a timeout that we don't have in `unsafeRunAsync` .

But if `.unsafeRunSync()` does not wait for the start, then we can start the `Future` and ignore it, just returning Unit.

SystemFw: > I still have questions about the .unsafeRunSync(), if this waits for the start.

it waits for the completion of the portion of IO preceding the first async boundary. In this case, because there is a `shift` immediately, it essentially doesn't wait. (and this is in CE2 anyway)

> If yes unsafeRunAsync starts the side-effect on another thread pool right?

Strictly speaking, it just follows what the IO does, and will follow whatever async boundary the IO has. In this case there is an immediate shift.

> As unsafeToFuture returns a Future, should we wait for the Future with an Await.result or something similar? 

no, we don't need to `Await.result`, and in fact what you're doing right now with `unsafeRunSync` is essentially `Await.result`,  which I think it's causing the deadlock in the test.

>  then we can start the Future and ignore it, just returning Unit.

yes

SystemFw: In fact, call `unsafeRunAndForget` on `Dispatcher`

yanns: @SystemFw thank you a lot for your detailed answers ❤️

SystemFw: Thank you for porting this :)

yanns: https://github.com/http4s/http4s/pull/3894/commits/ed7c215810da311fdbaa8b1b911ef717fcb0cf7b
https://github.com/http4s/http4s/pull/3894/commits/932bd3a761fd173bba81f77c81ab756e24b35a4a

so that the tests could run, I had to also do the following: https://github.com/http4s/http4s/pull/3894/commits/f7d5840e441deb56c22e1bcaf7f8e382107609f4

blaze-core/src/test/scala/org/http4s/blazecore/util/CatsEffect.scala
yanns: https://github.com/http4s/http4s/pull/3894/commits/52caa170f4c4859050f7876492a112c6f40c2856

yanns:
current status: all tests are now running.

@rossabaker this PR is quite messy as it contains a lot of commits.
Should I open a new PR with one commit?

SystemFw:
> Should I open a new PR with one commit?

The decision is definitely for Ross, but I actually really dislike that. The history is valuable, if you have ever have to run a bisect on a tough bug, big commits are a pain.

In any case, if you do want to do that, there's no need to open a new PR, you can squash on merge (but don't ;))

rossabaker:
I agree with @SystemFw.  It also makes it harder to re-review things: we have to look at the whole instead of the recent changes.  I usually never force-push a PR unless I don't think anybody has looked at it yet and I think I can tell a better story with a `rebase -i`.

The more controversial thing is whether we should squash at merge time or not.  Squash merges make a cleaner history and we can still click through to see the original intermediate commits.  But even then, I think the intermediate commits have value.  In addition to bisection, sometimes we branch off WIP to explore a different direction or, like now, maintain parallel branches through big upgrades.  Squashing is better for history but worse for collaboration, and we've optimize for the latter.

All that said, I'm excited to see the progress, but it's 1:30am here, so I'll have to review it tomorrow.  Hopefully our CE3 experts can keep looking as well.

yanns:
There is one topic I'm feeling unsure about: I had to create one dispatcher per test in `Http4sWSStageSpec` to that the tests can run (https://github.com/http4s/http4s/pull/3894/commits/f7d5840e441deb56c22e1bcaf7f8e382107609f4)

But I can still share a dispatcher in `Http1WriterSpec`.

And this is a change I could eventually rollback: https://github.com/http4s/http4s/pull/3894/commits/c59e3befc5f67d6fc6be949f505d88b4224eae74
Having a timeout when running tests is good not to wait forever. But it also means that slow tests can be flaky in case the timeout is near the execution time.

rossabaker (commented):
Is multiple dispatchers a normal usage?  I'm a bit worried that having to do that is masking an issue that will prevent the web sockets from working in the real world, but my understanding of CE3 is still a bit hazy.

I like the way the rest of this has shaped up.

yanns:
> Is multiple dispatchers a normal usage? I'm a bit worried that having to do that is masking an issue that will prevent the web sockets from working in the real world, but my understanding of CE3 is still a bit hazy.

Thanks to the help from @djspiewak, I could remove blocking calls: https://github.com/http4s/http4s/pull/3894/commits/ce3ea537546a160ab24e56131b867225c6eaedd1

Now all the tests can run with the same `Dispatcher`. ✌️

testing/src/main/scala/org/http4s/testing/CatsEffect.scala
```diff
@@ -0,0 +1,35 @@
...
+
+import cats.effect.{Async, Resource, Sync}
+import cats.effect.std.Dispatcher
+import org.specs2.execute.{AsResult, Result}
+
+import scala.concurrent.duration._
+
+/** copy of [cats.effect.testing.specs2.CatsEffect](https://github.com/djspiewak/cats-effect-testing/blob/series%2F0.x/specs2/src/main/scala/cats/effect/testing/specs2/CatsEffect.scala) adapted to cats-effect 3
+  */
+trait CatsEffect {
```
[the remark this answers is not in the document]
yanns: https://github.com/djspiewak/cats-effect-testing/issues/117

rossabaker (approved):
It was just that unsafeRunSync to create that semaphore in the stage?  In CE2, that's a "I promise not to share this" thing.  I wouldn't expect it to deadlock things.  The `Dispatcher` is a different world, though.

Anyway, hooray, and great job.

yanns:
> It was just that unsafeRunSync to create that semaphore in the stage? In CE2, that's a "I promise not to share this" thing. I wouldn't expect it to deadlock things. The `Dispatcher` is a different world, though.

Yes, the `Dispatcher.unsafeRunSync` blocks.
https://github.com/typelevel/cats-effect/blob/349fc13c6a6679ebd71168901874ee332e37ff5b/std/jvm/src/main/scala/cats/effect/std/DispatcherPlatform.scala#L29
Every time we use it, we have to keep it in mind. (and we should better not use it)

> Anyway, hooray, and great job.

Thanks, that was a long way.
But I've learned a lot, thanks to all the helpful reviews. ❤️

PR #3896, opened 2020-11-20
Stop pinning scalameta

rossabaker:
This causes friction on Scala upgrades. Can we remove it yet?


rossabaker:
Supersedes #3889.

ChristopherDavenport (approved):

PR #3897, opened 2020-11-20
Update Vault For Dotty

ChristopherDavenport:


rossabaker:
It can also now rejoin its friends in the unbranched libraryDependencies for core.  It was quarantined with the non-Dotty dependencies.

PR #3899, opened 2020-11-21
WIP: cats-effect-3 on dotty

rossabaker:
The dotty branch and the cats-effect-3 branch are going to need to reunite at some point.  When dotty lands in series/0.21, this should land in cats-effect-3.  If cats-effect-3 lands in main before dotty lands in series/0.21, then this should retarget main.

I don't plan to maintain this as much as the other feature branches, but frequent integration is important.


rossabaker:
This one has gnarly, diamond-dependent branches, and won't be necessary.

PR #3900, opened 2020-11-22
Setup munit dependencies and make boopickle use munit

cquiroz:
This is a replay of #3852 on `series/0.21`

### DIFF
--- a/boopickle/src/test/scala/org/http4s/booPickle/BoopickleSpec.scala
+++ b/boopickle/src/test/scala/org/http4s/booPickle/BoopickleSpec.scala
@@ -1,0 +1,56 @@
+/*
+ * Copyright 2013-2020 http4s.org
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.http4s
+package booPickle
+
+import boopickle.Default._
+import cats.effect.IO
+import cats.implicits._
+import cats.Eq
+import cats.effect.laws.util.TestContext
+import cats.effect.laws.util.TestInstances._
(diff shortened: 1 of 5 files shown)


rossabaker:
Looks like a legit compilation error.  Maybe a cats-effect-3 thing lingering?

cquiroz:
Yes, a fix is coming soon

rossabaker:
I can't remember. What piece do we need to keep the entity codec tests?

cquiroz:
We need a `discipline-effect` type of library. I'm working on one but still have some loose ends.
It should contain a `checkAllF(...): IO[Unit]` method to check properties running in an effect

rossabaker:
This has been explained to me about four times, but I haven't seen the compilation failure with my own eyes, so it hasn't sunk in.

Do you think we should merge this before that library arrives, or should we sit on this?

rossabaker:
We could always keep a vestigiial BoopickleSpec next to the BoopickleSuite and not lose any test coverage until the integrations catch up.

cquiroz:
Hmm, I guess we should wait, otherwise there is the temptation to keep porting tests with commented checks.

rossabaker:
:+1: I don't want to lose this, though.  I think it's a better place to be long term.

cquiroz:
I guess que could keep both the Spec and Suite for a while. My interest on munit is that it should theoretically be moving faster in the dotty world

cquiroz:
Ok I restored the `Spec` part testing the `Codec` laws

rossabaker (approved):

Issue #3869, opened 2020-11-15
HttpVersion could have a cats.Hash instance


rossabaker:
Can go on `series/0.21`. See #3792 for similar work.

ammarc:
Hi @rossabaker, I've made a PR for this: https://github.com/http4s/http4s/pull/3902

PR #3902, opened 2020-11-22
Add Hash and BoundedEnumerable instances for HttpVersion

ammarc:
For https://github.com/http4s/http4s/issues/3869

Test had to be adapted in order to fix the following compile error:
```
ambiguous implicit values:
 both value http4sHttpOrderForVersion in object HttpVersion of type cats.Order[org.http4s.HttpVersion]
 and value http4sHttpHashForVersion in object HttpVersion of type cats.Hash[org.http4s.HttpVersion]
 match expected type cats.kernel.Eq[org.http4s.HttpVersion]
  checkAll("HttpVersion", OrderTests[HttpVersion].order)
```

Do let me know if there's a better way to fix this.

### DIFF
--- a/core/src/main/scala/org/http4s/HttpVersion.scala
+++ b/core/src/main/scala/org/http4s/HttpVersion.scala
@@ -85,17 +85,21 @@ object HttpVersion {
 
     // BoundedEnumerable
     override def partialNext(a: HttpVersion): Option[HttpVersion] = a match {
-      case `HTTP/1.0` => Some(`HTTP/1.1`)
-      case `HTTP/1.1` => Some(`HTTP/2.0`)
+      case HttpVersion(9, 9) => None
+      case HttpVersion(major, minor) if fromVersion(major, minor).isLeft => None
+      case HttpVersion(major, 9) => Some(HttpVersion(major + 1, 0))
+      case HttpVersion(major, minor) => Some(HttpVersion(major, minor + 1))
       case _ => None
     }
     override def partialPrevious(a: HttpVersion): Option[HttpVersion] = a match {
-      case `HTTP/1.1` => Some(`HTTP/1.0`)
-      case `HTTP/2.0` => Some(`HTTP/1.1`)
+      case HttpVersion(0, 0) => None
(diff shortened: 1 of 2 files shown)


rossabaker:
Thanks for volunteering!

The problem is that `Hash` and `Order` both extend `Eq`, so now we have two implicit `Eq` instances at the same priority, when we want exactly one.  I didn't think about this until now.  The problem we're having in the tests will also happen downstream, so we ought to fix this.

One way around it is to make one instance that has all the Cats instances.  Case-insensitive has an [example](https://github.com/typelevel/case-insensitive/blob/1e256d6d96a6812c437149c58e02b12ab96ce018/core/src/main/scala/org/typelevel/ci/CIString.scala#L71).  If you copie and paste from that to get all the methods, this can't be a `Monoid`, but it could be a `BoundedEnumerable` instead of a `LowerBounded`.

We would need to remove the `implicit` keyword from the existing instances, and should add a `@deprecated` annotation.  This is all binary compatible, so the deprecation version could be "0.21.12" and we need to target series/0.21.

Whew, that turned into a lot.  If not all of that makes sense, please ask! :smile:

ammarc:
Thanks for the explanation, Ross.

Why do we need to add `BoundedEnumerable`? Don't we only need `Order`, `Hash` and `Show`?

ammarc:
> we need to target series/0.21
Do you mean I need to make this PR to `http4s:series/0.21` iso. `http4s:main`?

rossabaker (commented):
Nicely done.  One more thing.  Whenever we add an instance, we like to check its laws.  So in `HttpVersionSpec`, we should add:

```scala
  checkAll("HttpVersion", HashTests[HttpVersion].hash)
  checkAll("HttpVersion", BoundedEnumerableTests[HttpVersion].boundedEnumerable)
```

And with that, you should see a failing test on `BoundedEnumerable`.  1.0, 1.1, and 2.0 are the common ones, but there are actually 100 valid ones.

I know this is more than you signed up for, and really appreciate the work on this!

rossabaker (approved):
Great work!

rossabaker:
CI didn't run due to #3916, but I verified everything locally.

PR #3903, opened 2020-11-22
MethodOps apply returns Request[F]

### LINKED ISSUE #2738: client EmptyRequestGenerator and EntityRequestGenerator return types
Client's `EmptyRequestGenerator` and `EntityRequestGenerator` does have `apply` methods, that returns `F[Request[F]]`.
However, both methods uses `F.pure` and there is absolutely no reason to wrap it in such way.
Shouldn't it return just `Request[F]` instead?

Lasering:
Fix #2738


rossabaker (approved):
That didn't ripple nearly as hard as I feared! Nice cleanup.

Lasering:
Now I would like to tackle #2483, which will probably ripple.

rossabaker:
Docs builds on main started hanging when this was merged.  Don't yet know why, or if this is the cause, but it has been consistent since.

PR #3905, opened 2020-11-23
Port HttpsRedirectSpec to munit

cquiroz:
This is a sample of a test using `Http4sLegacyMatchersIO` ported to use `munit.CatsEffectSuite`. IMHO it looks cleaner but I'm happy to get more feedback

### DIFF
--- a/server/src/test/scala/org/http4s/server/middleware/HttpsRedirectSpec.scala
+++ b/server/src/test/scala/org/http4s/server/middleware/HttpsRedirectSpec.scala
@@ -8,14 +8,15 @@ package org.http4s
 package server
 package middleware
 
+import cats.implicits._
 import cats.effect._
 import org.http4s._
 import org.http4s.dsl.io._
 import org.http4s.headers._
+import org.http4s.syntax.all._
 import org.http4s.Uri.{Authority, RegName, Scheme}
-import org.http4s.testing.Http4sLegacyMatchersIO
 
-class HttpsRedirectSpec extends Http4sSpec with Http4sLegacyMatchersIO {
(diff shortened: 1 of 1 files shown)


server/src/test/scala/org/http4s/server/middleware/HttpsRedirectSpec.scala
```diff
@@ -8,50 +8,65 @@ package org.http4s
...
-        HttpsRedirect.httpRoutes(innerRoutes).orNotFound,
-        HttpsRedirect.httpApp(innerRoutes.orNotFound)
-      ).map { app =>
-        val resp = app(req).unsafeRunSync()
-        val expectedAuthority = Authority(host = RegName("example.com"))
-        val expectedLocation =
-          Location(
-            Uri(path = "/", scheme = Some(Scheme.https), authority = Some(expectedAuthority)))
-        val expectedHeaders = Headers(expectedLocation :: `Content-Type`(MediaType.text.xml) :: Nil)
-        resp.status must_== Status.MovedPermanently
```
cquiroz, line 39: The original had 2 assertions and I converted them to 2 tests. It could also be done as
`IO(a).assert.. >> IO(b).assert`
WDYT?

rossabaker: I think I'm fine with the `>>` -- it's asserting over related things, and there's not insignificant setup.  But I could go either way.

rossabaker (approved):
:+1:

cquiroz:
I merged the checks, you get less code and time spent running the tests

PR #3906, opened 2020-11-23
Fix Streamed encoder for empty stream

peterneyens:
Follow up of #3891.

It seems we aren't emiting the opening `[` when the stream is empty.


JesusMtnez:
Thank you @peterneyens for the quick fix!

rossabaker (approved):
Oops. Good find. Let's cherry-pick this back to 0.21 and get a patch out ASAP.

rossabaker:
Cherry-picked onto 0.21.

PR #3909, opened 2020-11-23
Add Order Instance For Header and Headers

isomarcte:
I left the old `Eq` based symbols in place so as not to break binary compatibility.

### DIFF
--- a/core/src/main/scala/org/http4s/Header.scala
+++ b/core/src/main/scala/org/http4s/Header.scala
@@ -10,7 +10,7 @@
 
 package org.http4s
 
-import cats.{Eq, Show}
+import cats.{Eq, Order, Show}
 import cats.data.NonEmptyList
 import cats.implicits._
 import org.http4s.syntax.string._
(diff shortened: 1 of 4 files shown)


core/src/main/scala/org/http4s/Header.scala
```diff
@@ -132,7 +132,16 @@ object Header {
     _.toString
   }
 
-  implicit val HeaderEq: Eq[Header] = Eq.instance[Header] { (a, b) =>
-    a.name === b.name && a.value === b.value
-  }
+  @deprecated(message = "Please use HeaderOrder instead", since = "1.0.0-M8")
```
rossabaker, line 135: We can do this in series/0.21.  I can cherry-pick it if you don't want to rebase it, but we should change the version number.

core/src/main/scala/org/http4s/Header.scala
```diff
@@ -132,7 +132,16 @@ object Header {
...
+  lazy val HeaderEq: Eq[Header] = HeaderOrder
```
rossabaker, line 136: Nit: since it's just an alias, a `def` is faster and lighter.

rossabaker (changes requested):
Nice! Can we also add `OrderTests` for these to make sure they're lawful?  I think they are.

core/src/main/scala/org/http4s/Header.scala
isomarcte: @rossabaker I'll rebase.

core/src/main/scala/org/http4s/Header.scala
isomarcte: I'll update this.

isomarcte:
@rossabaker updated. 

* Aliases use `def` now. 
* Rebased and retargeted for `series/0.21`. 
* Updated the deprecation warning.
* Added `OrderTests` for `Header` and `Headers`.

rossabaker (approved):
Great. :+1: on green

isomarcte:
@rossabaker some tests appear to be failing, but I don't think they are related to this PR.

rossabaker:
Agreed. Unfortunately, the bigger the build gets, the more often this happens. I hope the munit work reducing unsafeRunSync()s helps.

PR #3911, opened 2020-11-24
Support raw query string

hamnis:
- Lazily parse Query by access of "pairs" variable
- Add raw for access and withRaw to replace the internal string representation

This should hopefully be binary compatible, lets see what MIMA says.

### DIFF
--- a/core/src/main/scala/org/http4s/Query.scala
+++ b/core/src/main/scala/org/http4s/Query.scala
@@ -23,29 +23,37 @@ import scala.collection.immutable
   * When rendered, the resulting `String` will have the pairs separated
   * by '&' while the key is separated from the value with '='
   */
-final class Query private (val pairs: Vector[KeyValue]) extends QueryOps with Renderable {
+final class Query private (value: Either[Vector[KeyValue], String])
+    extends QueryOps
+    with Renderable {
+  lazy val pairs: Vector[KeyValue] = value.fold(identity, Query.parse)
+
+  def raw: Option[String] = value.toOption
+
+  def withRaw(value: String) = new Query(Right(value))
+
   def apply(idx: Int): KeyValue = pairs(idx)
 
   def length: Int = pairs.length
 
-  def slice(from: Int, until: Int): Query = new Query(pairs.slice(from, until))
+  def slice(from: Int, until: Int): Query = new Query(Left(pairs.slice(from, until)))
 
   def isEmpty: Boolean = pairs.isEmpty
 
   def nonEmpty: Boolean = pairs.nonEmpty
 
-  def drop(n: Int): Query = new Query(pairs.drop(n))
+  def drop(n: Int): Query = new Query(Left(pairs.drop(n)))
 
-  def dropRight(n: Int): Query = new Query(pairs.dropRight(n))
+  def dropRight(n: Int): Query = new Query(Left(pairs.dropRight(n)))
 
   def exists(f: KeyValue => Boolean): Boolean =
     pairs.exists(f)
 
   def filterNot(f: KeyValue => Boolean): Query =
-    new Query(pairs.filterNot(f))
+    new Query(Left(pairs.filterNot(f)))
 
   def filter(f: KeyValue => Boolean): Query =
-    new Query(pairs.filter(f))
+    new Query(Left(pairs.filter(f)))
 
   def foreach(f: KeyValue => Unit): Unit =
     pairs.foreach(f)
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/http4s/Query.scala
```diff
@@ -23,29 +23,37 @@ import scala.collection.immutable
   * When rendered, the resulting `String` will have the pairs separated
   * by '&' while the key is separated from the value with '='
   */
-final class Query private (val pairs: Vector[KeyValue]) extends QueryOps with Renderable {
+final class Query private (value: Either[Vector[KeyValue], String])
```
rossabaker, line 26:
```
[error] core: Failed binary compatibility check against org.http4s:http4s-core_2.13:0.21.1! Found 1 potential problems
[error]  * method this(scala.collection.immutable.Vector)Unit in class org.http4s.Query's type is different in current version, where it is (scala.util.Either)Unit instead of (scala.collection.immutable.Vector)Unit
[error]    filter with: ProblemFilters.exclude[IncompatibleMethTypeProblem]("org.http4s.Query.this")
```

We could do it with an auxiliary constructor.  Or grant an exception, because it was private.

core/src/main/scala/org/http4s/Query.scala
```diff
@@ -23,29 +23,37 @@ import scala.collection.immutable
...
+    extends QueryOps
+    with Renderable {
+  lazy val pairs: Vector[KeyValue] = value.fold(identity, Query.parse)
```
rossabaker, line 29: I proposed a lazy val last night.  We could also consider the null init pattern used in loggers, which is a bit faster when there is no contention.

core/src/main/scala/org/http4s/Query.scala
```diff
@@ -23,29 +23,37 @@ import scala.collection.immutable
...
+
+  def raw: Option[String] = value.toOption
```
rossabaker, line 31: There's always a raw value.  I think we could lazily render a left and lose the `Option`.

core/src/main/scala/org/http4s/Query.scala
```diff
@@ -23,29 +23,37 @@ import scala.collection.immutable
...
+
+  def withRaw(value: String) = new Query(Right(value))
```
rossabaker, line 33: This feels more like a smart constructor on the companion than a method here, but I don't care that much.

rossabaker (changes requested):
This is outstanding!  Need to fix MiMa, but there's a way.

core/src/main/scala/org/http4s/Query.scala
hamnis: fixed with auxillary constructor. Finally got a use for that language feature :p

core/src/main/scala/org/http4s/Query.scala
hamnis: handled that in render and removed the method

core/src/main/scala/org/http4s/Query.scala
hamnis: there already is one, and that is fromString

core/src/main/scala/org/http4s/Query.scala
rossabaker: Yeah, I always prefer to fix it with code rather than exemptions.

rossabaker (approved):
Looks great.

/cc @drwlrsn, whose work this will enable. Look for it in http4s-0.21.12 and 1.0.0-M8.

rossabaker:
Trying to trigger GitHub Action

core/src/main/scala/org/http4s/Query.scala
```diff
@@ -160,17 +179,20 @@ object Query {
     * If parsing fails, the empty [[Query]] is returned
     */
   def fromString(query: String): Query =
-    if (query.isEmpty) new Query(Vector("" -> None))
-    else
-      QueryParser.parseQueryString(query) match {
-        case Right(query) => query
-        case Left(_) => Query.empty
-      }
+    new Query(Right(query))
```
rossabaker, line 182: Actually, this allows invalid queries.  We need to make sure it's all pchars.

We foolishly made this return a `Query` instead of a `ParseResult[Query]`.  We can't fix that now, but can in 1.0.  I guess for compatibility, we should return the empty string if it's invalid. :vomiting_face:

rossabaker (changes requested):

rossabaker (approved):
I think that's the best we can do without breaking bincompat.

PR #3914, opened 2020-11-24
Non-trivial merge of main to cats-effect-3

rossabaker:
I lost the `BoopickleSpec` somewhere along the way here.  Do we want to leave this and assume that resolving it on series/0.21 will fix it, or should we go ahead and port it?  I'm still a bit lost on CE3.


cquiroz:
We can merge BoopickeSpec later

PR #3917, opened 2020-11-24
Changelog for v0.21.12

rossabaker:


website/src/hugo/content/changelog.md
```diff
@@ -8,6 +8,28 @@ Maintenance branches are merged before each new release. This change log is
 ordered chronologically, so each release contains all changes described below
 it.
 
+# v0.21.12 (2020-11-24)
+
+## Bugfixes
+
+### http4s-core
```
rossabaker, line 15: I've been experimenting with grouping tickets by the relevant module.  Should we group by type and then module, or by module and then type?

Issue #3920, opened 2020-11-24
Uris with a colon are rendered incorrectly since 0.21.9


RafalSumislawski:
When using blaze client to send a request for an URL with a `:` in the path, the request is rendered with incorrect URL. 

For example a request for http://github.com/http4s/foo:bar gets rendered as:
```
GET .//http4s/foo:bar
```

It should be be:
```
GET /http4s/foo:bar
```

This is result of the changes introduced in #3752 (https://github.com/http4s/http4s/pull/3752/files#diff-ee2b5495b08afcff2ce456ed5f27c2867c9d7c00bcb05bd72e6d6af88b4f5c9bR137)

The "./" should be prepended only if the URI 
* is a relative-path reference
* and the first segment contains a `:`

Currently it is prepended to absolute-path references and relative-path references, and it is prepended regardless of which segment contains the colon.

## Reproduction
with blaze client 0.21.9 or newer
```
import cats.effect.{ExitCode, IO, IOApp}
import org.http4s.client.blaze.BlazeClientBuilder
import org.http4s.{Response, Uri}

import scala.concurrent.ExecutionContext

object Test extends IOApp {

  override def run(args: List[String]): IO[ExitCode] =
    BlazeClientBuilder[IO](ExecutionContext.global)
      .resource.use { client =>
      client.get[Unit](Uri.unsafeFromString("http://github.com/http4s/foo:bar"))((response: Response[IO]) => IO(println(response)))
    }.as(ExitCode.Success)
}

```

Results in the following request (captured with wireshark):

![obraz](https://user-images.githubusercontent.com/9465998/100149428-76eebd00-2e9e-11eb-9169-1c596ff07d49.png)

PR #3921, opened 2020-11-24
Fix rendering of URIs with colons

RafalSumislawski:
Fix for #3920


rossabaker:
Good find.  I need to go eat, but we'll hold 0.21.12 for this and I'll review it tonight.  Also pinged Ashwin since he was recently immersed in this spec.

ashwinbhaskar (approved):

rossabaker (approved):
Nicely done.

PR #3934, opened 2020-11-25
Fix docs for client DSL changes in #3903

### LINKED ISSUE #3933: Docs build is timing out on main
First observed after merge of #3903. We don't get any logs -- build just times out after six hours.

rossabaker:
Fixes #3933.  Some of the regular crossbuilds failed with the regular flakes, which failed fast, and we never successfully ran the doc job before we merged #3903.


rossabaker:
Post-reviews welcome, but things are horribly broken without this.

PR #3937, opened 2020-11-26
Add Order And Hash Instance For Charset

isomarcte:


rossabaker (commented):
We can cherry-pick this to series/0.21.  (If you're not making breaking changes, always target series/0.21.)  Could throw in a `Hash` instance for extra credit, but it'll need to be in the same value as the `Order` to avoid competing `Eq`s.  See #3902 for an example of that.

isomarcte:
@rossabaker oh sorry about that. I had meant to target series/0.21, but I guess I wasn't paying attention yesterday. I'll get that fixed and add the `Hash` instance.

rossabaker:
Needs a scalafmt, but otherwise looks good.

isomarcte:
@rossabaker updated.

rossabaker (approved):
:+1: Thanks!

PR #3943, opened 2020-11-27
Dependencies that go bump in the night

rossabaker:


rossabaker:
The docs project doesn't like our interpolated strings.  What's up?

rossabaker:
jawn-1.0.2 broke the docs project.

server/src/main/scala/org/http4s/server/middleware/authentication/DigestAuth.scala
```diff
@@ -77,6 +78,7 @@ object DigestAuth {
       }
     }
 
+  @nowarn("cat=other-match-analysis")
```
rossabaker, line 81: 2.13.4 wants a `case Some(Authorization(_))`, which exists?

rossabaker:
Gosh.

```
2020-11-30T22:01:07.0007751Z      final def doFetchFR2[B](req: G[Request[G]], f: Response[G] => G[B])(implicit G: Sync[G]): G[B] =
2020-11-30T22:01:07.0008745Z -      req.flatMap(this.run(_).use f)
2020-11-30T22:01:07.0009464Z +      req.flatMap(this.run(_).use { f })
```

rossabaker:
I don't think some of that scalafix output is even valid Scala?

```
  def doClientFetchR5[F[_]: Sync, A](client: Client[F], req: Request[F], f: Response[F] => F[A]): F[A] = {
-    client.run(req).use[F, A] case res@Response(_, _, _, _, _) => f(res)
+    client.run(req).use[F, A] { case res@Response(_, _, _, _, _) => f(res) }
```

rossabaker:
We're kind of pinned on this one.  Older scalafix doesn't support Scala 2.13.4, and the new one seems to introduce that bug.

rossabaker:
https://github.com/scalacenter/scalafix/issues/1299

server/src/main/scala/org/http4s/server/middleware/authentication/DigestAuth.scala
SethTisue: is `Authorization(_)` irrefutable?

rossabaker: The warning goes away if `Authorization.unapply` is declared to return `Some` instead of `Option`, but that breaks binary compatibility.  The error is confusing, because it asks to repeat the case verbatim, but I can also see why scalac sees a leak.

With fresh eyes, I think `Some(_)` would be correct, be complete, and appease scalac.

.github/workflows/ci.yml
```diff
@@ -23,7 +23,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        scala: [2.13.4, 2.12.12]
+        scala: [2.13.3, 2.12.12]
```
SethTisue: awwwww.... really?

rossabaker: There's a lot tied up in this and I have time to dive into scalafix and see why it's generating invalid code.

rossabaker: I kept the changes that make it compile on Scala 2.13.4.  Hopefully we can bring it back soon.

SethTisue: ah, the Scalafix thing. yeah, fair.

PR #3958, opened 2020-11-28
Port HttpDate-related parsers to cats-parse

rossabaker:
* Can't eliminate the parboiled2 version yet, because other rules reference it, and parboiled2 doesn't compose.
* Deserves more tests, but it's more than we had.
* Ported the test we had to MUnit.


core/src/main/scala/org/http4s/HttpDate.scala
```diff
@@ -121,4 +123,164 @@ object HttpDate {
...
+        day: Int,
+        hour: Int,
+        min: Int,
+        sec: Int): Parser[HttpDate] =
+      try {
+        val dt = ZonedDateTime.of(year, month, day, hour, min, sec, 0, ZoneOffset.UTC)
+        pure(org.http4s.HttpDate.unsafeFromZonedDateTime(dt))
+      } catch {
+        case _: DateTimeException =>
+          failWith(s"Invalid IMF-fixdate: $year-$month-$day $hour:$min:$sec")
```
rossabaker, line 144: This demonstrates a failure of our model: the RFC doesn't have anything to say about the date making sense, and being forced to flatMap is bad for parser optimization.

In 1.0, we should carry around the six fields, and have a `Try[ZonedDateTime]`.

core/src/main/scala/org/http4s/HttpDate.scala
```diff
@@ -121,4 +123,164 @@ object HttpDate {
...
+
+    def mkHttpDate(
+        year: Int,
+        month: Int,
+        day: Int,
+        hour: Int,
+        min: Int,
+        sec: Int): Parser[HttpDate] =
+      try {
+        val dt = ZonedDateTime.of(year, month, day, hour, min, sec, 0, ZoneOffset.UTC)
```
rossabaker, line 140: Note that the day of week is parsed, but never validated.  Never has been.  Probably never should be.

core/src/main/scala/org/http4s/HttpDate.scala
rossabaker: Ah, nuts:

>    A date-time specification MUST be semantically valid.  That is, the
   day-of-week (if included) MUST be the day implied by the date, the
   numeric day-of-month MUST be between 1 and the number of days allowed
   for the specified month (in the specified year), the time-of-day MUST
   be in the range 00:00:00 through 23:59:60 (the number of seconds
   allowing for a leap second; see [RFC1305]), and the last two digits
   of the zone MUST be within the range 00 through 59.

rossabaker:
I thought I could eliminate the parboiled date rules, but cookie has one, and the cookie spec is one of more complicated ones, and one whose strictness people have complained about.  That one will require more thought.

I think this one represents a nice step forward as is.

core/src/main/scala/org/http4s/headers/Expires.scala
```diff
@@ -7,12 +7,29 @@
...
+  /* `Expires = HTTP-date` */
+  private[http4s] val parser: Parser[Expires] = {
+    import Parser.anyChar
+
+    def httpDate = HttpDate.parser
+
+    // A cache recipient MUST interpret invalid date formats, especially the
+    // value "0", as representing a time in the past (i.e., "already
+    // expired").
+    def invalid = anyChar.rep.as(HttpDate.Epoch)
```
rossabaker, line 29: We used to handle  "0" and "-1" special, but the spec likes them all.

A better model in 1.0 would preserve the invalid value for round tripping instead of forcing it to the epoch.

core/src/main/scala/org/http4s/headers/Expires.scala
```diff
@@ -7,12 +7,29 @@
...
 
 object Expires extends HeaderKey.Internal[Expires] with HeaderKey.Singleton {
   override def parse(s: String): ParseResult[Expires] =
-    HttpHeaderParser.EXPIRES(s)
+    parser.parseAll(s).leftMap { e =>
+      ParseFailure("Invalid Expires header", e.toString)
+    }
+
+  /* `Expires = HTTP-date` */
+  private[http4s] val parser: Parser[Expires] = {
```
rossabaker, line 21: As a consequence of treating invalid dates as special, this becomes a `Parser` instead of a `Parser1`.  I don't _think_ that will haunt us, but, maybe?

ChristopherDavenport (approved):
I think this is a sane way to preserve the current API

Issue #3947, opened 2020-11-27
Port argonaut to cats-effect-3


rossabaker:
Probably roughly follows the patterns of #3885. Open against cats-effect-3 branch.

benderpremier:
I'll try to work on that ticket 🙏

rossabaker:
Fixed by #3961.

PR #3961, opened 2020-11-28
ISSUE-3947 use concurrent instead of sync

benderpremier:
Replaced `Sync` with `Concurrent` for #3947


rossabaker (approved):
:+1: on green

rossabaker:
Docs failure is unrelated.

PR #3962, opened 2020-11-28
Issue-3946 Activate and migrate play-json to Cats-effect 3

amiguez:
As discussed with @rossabaker. Use `Concurrent` instead of `Sync` in the play-json module. Issue: #3962


rossabaker (approved):
:+1: on green

rossabaker:
Docs failure is unrelated.

PR #3963, opened 2020-11-28
Port many tests using legacy matchers to munit

cquiroz:
This PR tries to remove usages of `Http4sLegacyMatchersIO` using munit cats effect support instead. There is a long way to go but this should remove its usage in server middleware.


project/Http4sPlugin.scala
```diff
@@ -356,6 +356,7 @@ object Http4sPlugin extends AutoPlugin {
     val reactiveStreams = "1.0.3"
     val quasiquotes = "2.1.0"
     val scalacheck = "1.15.1"
+    val scalacheckEffect = "0.6.0"
```
rossabaker, line 359: Oh, is this the missing piece on #3965?

server/src/test/scala/org/http4s/server/middleware/AutoSlashSuite.scala
```diff
@@ -0,0 +1,83 @@
...
+    val req = Request[IO](uri = uri("/ping/"))
+    route.orNotFound(req).map(_.status).assertEquals(Status.NotFound) *>
+      AutoSlash(route).orNotFound(req).map(_.status).assertEquals(Status.Ok)
+  }
+
+  test("Match a route defined with a slash") {
+    AutoSlash(route)
+      .orNotFound(Request[IO](uri = uri("/withslash")))
+      .map(_.status)
+      .assertEquals(Status.Ok) *>
```
rossabaker, line 35: What happens if you forget the `*>`?  A warning on discarded values?

rossabaker (approved):
This is great.  I admit to not reviewing every test, but I got a sense of the pattern, and I like it.

This is going to be a terrible, horrible, no good, very bad merge to main, particularly with `CaseInsensitiveString`.  But I think it's the right thing to do to start on this branch.

project/Http4sPlugin.scala
cquiroz: Not compeletly. This let's you evaluate properties running on `F` but we also need (I think) a`discipline-cats-effect` that let's you run sets of properties a.k.a `laws`. I started working on this but needs some extra work

server/src/test/scala/org/http4s/server/middleware/AutoSlashSuite.scala
cquiroz: I didn't see the warning and it is easy to miss. I thought `sbt-tpolecat` would include a flag to warn about this.

cquiroz:
I'll merge this and then port to `main`, or should it go to `dotty` first?

server/src/test/scala/org/http4s/server/middleware/AutoSlashSuite.scala
rossabaker: That's scary.  I'm worried about losing assertions.  It has been one of my problems with these tests-as-values frameworks, even though I think it's a net win.

project/Http4sPlugin.scala
cquiroz: We could go the route of one IO result per test

rossabaker:
Let's focus on `main`.  That's going to be hard, and the tests are fresher in your mind.  I can handle `dotty`, which should be easy -- no `CIString` changes.

PR #3965, opened 2020-11-28
ISSUE-3948 Port circe to cats-effect-3

benderpremier:
Still need to migrate to munit


circe/src/test/scala/org/http4s/circe/CirceSpec.scala
```diff
@@ -370,5 +364,6 @@ class CirceSpec extends JawnDecodeSupportSpec[Json] with Http4sLegacyMatchersIO
     }
   }
 
-  checkAll("EntityCodec[IO, Json]", EntityCodecTests[IO, Json].entityCodec)
+// TODO
+//  checkAll("EntityCodec[IO, Json]", EntityCodecTests[IO, Json].entityCodec)
```
rossabaker, line 368: @cquiroz: @benderpremier and I got stuck here today, because there's no `Eq[IO[A]]` anymore.  I think this is the same issue you had on boopickle where we another test integration library?  Or can we get this done now just by porting to MUnit?

build.sbt
```diff
@@ -380,6 +380,7 @@ lazy val circe = libraryProject("circe")
       circeCore,
       circeJawn,
       circeTesting % Test,
+      catsEffectLaws % Test
```
rossabaker, line 383: I think we inherit this already.  The things we used to use out of it are gone, but I believe the dependency is still here.

rossabaker (commented):
Thanks, @benderpremier!  I summoned Carlos on the test we got stuck on.

benderpremier:
I will try to do the same migration to MUnit as is #3963 , what's the right way to do it though ? Should I wait for 3963 to be merged to the `0.21` branch and then merge those in the `cats-effect-3` branch? Or should I copy the change done in the `Http4sSuite.scala` ?

circe/src/test/scala/org/http4s/circe/CirceSpec.scala
cquiroz: I think we need a `discipline-cats-effect` that provides a `checkAllF` but we are not there yet.
As a workaround I made a `BoopickleSuite` with the `munit` tests and kept `BoopickleSpec` with only the law check

rossabaker: I thought that's what you did, and I went looking for it yesterday.  I think I was just on the wrong branch.

rossabaker: Oh, is that not on cats-effect-3 yet?  I was thinking this is what needed to move to munit, but it's the thing you held _out_ of munit.  But I'm not sure how it runs on Specs2, because we're missing the old `Eq[IO[A]]` that we used to have.

rossabaker:
We want to go series/0.21 -> main -> cats-effect-3.  That first step is going to be a nightmare, but @cquiroz is working on that.

It looks like I'm wrong about munit being the (immediate) solution to that `checkAll`.  I think we might be able to bring this in for a landing just by running scalafmt, and we can get it all moved over to munit in a subsequent step.

benderpremier:
I think we still have an issue with a dependency on https://github.com/ChristopherDavenport/vault which we have in `org.http4s.Message`, the [newKey](https://github.com/http4s/http4s/blob/cats-effect-3/core/src/main/scala/org/http4s/Message.scala#L498-L501) methods need an implicit `Sync`. We probably need to publish vault against CE3 right ?

rossabaker:
Oh, drat.  You're right.  I helped with the Dotty release of that, but we need CE3:

https://github.com/ChristopherDavenport/vault/issues/159

rossabaker:
#4040 effectively gives us a CE3 vault.  If we merge in the cats-effect-3 branch, we should be able to finally get this finished up.  @benderpremier, do you still have time and interest in finishing this one up?  Thanks!

benderpremier:
Hi @rossabaker , yes I'd love to finish that one. I will have some time in the coming days to do so.

rossabaker:
I think that build error is a result of some build changes that happened while this PR was stalled.  Running `sbt headerCreateAll` should get past that.

benderpremier:
I've commented out the test in `CirceSpec`, we should be able to bring it back once https://github.com/http4s/http4s/pull/4007 is done

rossabaker (approved):
:+1: Thanks for being patient as the other pieces came together. This looks good!

PR #3966, opened 2020-11-28
Add Order Instance For Method

isomarcte:
This commit also updates the law tests for `Hash[Method]` as they were not actually testing the HashLaws, just the EqLaws.


rossabaker (approved):
:tada:

isomarcte:
@rossabaker I noticed this had a duplicate import after the implicits <-> syntax merge. I've updated to fix that.

rossabaker:
Ugh, 0-for-2 in my browser merges.  Now I remember why I don't do that.

PR #3969, opened 2020-11-29
Add Order/Hash/Show Instances For Uri And Related Types

isomarcte:
This commit adds `Order`, `Hash`, and `Show` instances for the following types.

* `org.http4s.Uri.Authority`
* `org.http4s.Query`
* `org.http4s.Uri.Host`
* `org.http4s.Uri.RegName`
* `org.http4s.Uri`

The impetus for all of them was to support instances for `org.http4s.Uri`.

In addition, two functions were added to the package object `org.http4s.internal` scoped to be private to `http4s`, `compareField`, `reduceComparisons_`, and `reduceComparisons`.

These functions substantially reduce the amount of code required to write `Order` instances for Product types.

In order to support testing the laws for these types, several new instances of `Arbitrary` and `Cogen` were added, as not all types had instances for each of these.

### DIFF
--- a/core/src/main/scala/org/http4s/package.scala
+++ b/core/src/main/scala/org/http4s/package.scala
@@ -6,7 +6,7 @@
 
 package org
 
-import cats.data.{EitherT, Kleisli, OptionT}
+import cats.data._
 import fs2.Stream
 
 package object http4s {
--- a/laws/src/main/scala/org/http4s/laws/discipline/ArbitraryInstances.scala
+++ b/laws/src/main/scala/org/http4s/laws/discipline/ArbitraryInstances.scala
@@ -570,6 +570,12 @@ private[http4s] trait ArbitraryInstances {
   val genHexDigit: Gen[Char] = oneOf(
     List('0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'A', 'B', 'C', 'D', 'E', 'F'))
 
+  val genPctEncoded: Gen[String] =
+    const("%") |+| genHexDigit.map(_.toString) |+| genHexDigit.map(_.toString)
+  val genUnreserved: Gen[Char] =
+    oneOf(alphaChar, numChar, const('-'), const('.'), const('_'), const('~'))
+  val genSubDelims: Gen[Char] = oneOf(List('!', '$', '&', '\'', '(', ')', '*', '+', ',', ';', '='))
+
   private implicit def http4sTestingSemigroupForGen[T: Semigroup]: Semigroup[Gen[T]] =
     new Semigroup[Gen[T]] {
       def combine(g1: Gen[T], g2: Gen[T]): Gen[T] = for { t1 <- g1; t2 <- g2 } yield t1 |+| t2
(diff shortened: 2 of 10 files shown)


isomarcte:
@rossabaker so this commit breaks binary compatibility in two ways.

For `org.http4s.Uri#Authority`, the addition of a companion object messes up the type hierarchy from the one generated by the case class.

```
Error:  core: Failed binary compatibility check against org.http4s:http4s-core_2.13:0.21.1! Found 1 potential problems
Error:   * the type hierarchy of object org.http4s.Uri#Authority is different in current version. Missing types {scala.runtime.AbstractFunction3}
Error:     filter with: ProblemFilters.exclude[MissingTypesProblem]("org.http4s.Uri$Authority$")
```

It's simple enough to add that back in, though a bit gross. 

```scala
  object Authority extends scala.runtime.AbstractFunction3[Option[UserInfo], Host, Option[Int], Authority] {

    override def apply(userInfo: Option[UserInfo] = None, host: Host = RegName("localhost"), port: Option[Int] = None): Authority =
      new Authority(userInfo, host, port)
```

The binary change is the addition of the new `Cogen` and `Arbitrary` instances. This is an additive change, so no symbols are changed/removed.

My intuition would be to add the `AbstractFunction3` extension now, then remove it on the rebase to the `main` branch. For the new instances for `Cogen` and `Arbitrary`, I'd just add them to the `ProblemFilters` for mima on this branch, since they are just additions, not breaking changes.

What are your thoughts?

rossabaker:
I did exactly that with an `AbstractFunction1` somewhere else.  I think your idea is exactly right.

The arbitrary changes pass MiMa?  Adding to traits was bad in 2.11, but I think is fine starting in 2.12 as long as there is a concrete implementation.

isomarcte:
@rossabaker I've updated the PR to include the `AbstractFunction3` workaround.

> The arbitrary changes pass MiMa? Adding to traits was bad in 2.11, but I think is fine starting in 2.12 as long as there is a concrete implementation.

Yeah, that is my understanding as well. To be clear, MiMa _does_ fail on them, but I don't think they are actually an issue unless http4s is still targeting 2.11.x.

Thoughts?

rossabaker:
I usually end up regretting making MiMa exceptions.  I think I understand, and then a week later, we get a bug.

We already have a couple in the companion object.  I think making them a trait was a mistake, and most people should be importing them.  I'd be fine putting them in the companion, and we can reunify them somewhere in 1.0.

isomarcte:
@rossabaker sounds good. I've updated this with the new instances only in the companion object.

isomarcte:
@rossabaker I had to make a few things `lazy` to deal with some initialization errors coming from the interplay between the static scope of the companion object and the trait, just to let you know.

isomarcte:
@rossabaker as per other PRs, the failures here do not appear related to my change.

core/src/main/scala/org/http4s/package.scala
```diff
@@ -82,4 +84,54 @@ package object http4s {
   type Http4sSyntax = syntax.AllSyntax
   @deprecated("Moved to org.http4s.syntax.all", "0.16")
   val Http4sSyntax = syntax.all
+
+  // Helper functions for writing Order instances //
```
rossabaker, line 88: We keep a lot of helpers in `org.http4s.internals`, just to reduce clutter.

core/src/main/scala/org/http4s/package.scala
```diff
@@ -82,4 +84,54 @@ package object http4s {
...
+    * type.
+    *
+    * See the `Order` instance for `Uri` for an example of this usage.
+    *
+    * @note It is important to remember that `reduceMap` and `reduceMapK`
+    *       produce different results, since `Semigroup[Option[A]].combine`
+    *       and `SemigroupK[Option].combineK` have different semantics. We
+    *       ''always'' want the `SemigroupK[Option]` instance here.
+    */
+  private[http4s] def reduceComparisons(
```
rossabaker, line 126: This runs all the comparisons, right?  Could we do a lazy one that only continues for as long as we get zeroes?

laws/src/main/scala/org/http4s/laws/discipline/ArbitraryInstances.scala
```diff
@@ -634,11 +638,12 @@ private[http4s] trait ArbitraryInstances {
     } yield Uri.Authority(maybeUserInfo, host, maybePort)
   }
 
-  val genPctEncoded: Gen[String] =
+  lazy val genPctEncoded: Gen[String] =
```
rossabaker, line 641: A long, long time ago, I saw lazy arbitraries making a difference in a profile, which caused us to use eager vals.  I don't know whether that's still a concern, but I think you could just by moving them above the new use.

Whether eager values are smart in a trait is a separate question.

I'm fine either way.

rossabaker (commented):
I agree the test failures are unrelated.  I did come up with a few more ideas.

isomarcte:
@rossabaker Updated. 

I moved the helpers into the `org.http4s.internal` package object, made `reduceComparisons` lazy for every comparison other than the first, and shuffled things around in `ArbitraryInstances` so that it could use strict values again.

rossabaker (approved):
:+1: Thanks!

isomarcte:
@rossabaker I saw this was missing an import after the merge with for the implicits <-> syntax change. I've updated the branch.

rossabaker:
I gambled and merged in the browser last night without compiling. Thanks for the mop-up!

PR #3970, opened 2020-11-29
Port many tests using legacy matchers to munit

cquiroz:
Replay of #3963 to `main`.
What would be the next branch to target?


rossabaker:
cats-effect-3.  I already did dotty.

cquiroz:
Should I merge this one?

rossabaker (approved):
It's huge, so I only reviewed a sampling, but I like what I see.  I hope it wasn't too hard.  And I hope we start to see progress in our build stability.  Did you have to rerun this, or was it green the first try?

cquiroz:
It went green on the first try, I've seen a few failures though but in other tests, like some Blaze one and another about IpV6

The port wasn't terrible, mostly mechanical changes

rossabaker:
Those blaze ones in particular I'm hoping are improved by this work.

We also have some ridiculously long running property tests from ridiculous generators that are CPU-binding the test pool for several seconds at a time.  I'm going to be taking a look at those as time permits as well.

cquiroz:
I'm trying to take this to `cats-effect-3` but it seems the `server` module isn't compiling yet, thus I can't really verify if it works

rossabaker:
If it's commented out, it already didn't work on that branch.  Take a best guess, and we'll leave it for the server port task.

PR #3972, opened 2020-11-30
Improve Link header parser conformance to RFC 8288

wemrysi:
[RFC 8288, Section 3.3](https://tools.ietf.org/html/rfc8288#section-3.3) states that any occurrences of the `rel` parameter after the first one must be ignored. This fixes the `Link` header parser to conform to this behavior.


rossabaker (approved):
Good find.  I also see that it's a required field, which I don't think we can do anything about in 0.21, but is worth reconsidering on `main`.

PR #3964, opened 2020-11-28
3944 cats2.3 switch implicits import to syntax

amiguez:
This changes addresses the https://github.com/http4s/http4s/issues/3944 
Replace cats.implicits._ with cats.syntax.all._

FIX https://github.com/http4s/http4s/issues/3944

PR #3974, opened 2020-12-02
Non-trivial merge from series/0.21

rossabaker:
Includes #3963, #3937, #3964, #3969, #3966.

### DIFF
--- a/core/src/main/scala/org/http4s/Method.scala
+++ b/core/src/main/scala/org/http4s/Method.scala
@@ -7,7 +7,7 @@
 package org.http4s
 
 import cats.syntax.all._
-import cats.{Eq, Hash, Order, Show}
+import cats.{Hash, Order, Show}
 import org.http4s.parser.Rfc2616BasicRules
 import org.http4s.util.{Renderable, Writer}
 import scala.util.hashing.MurmurHash3
@@ -158,19 +158,4 @@ object Method {
       override def compare(x: Method, y: Method): Int =
         x.name.compare(y.name)
     }
-
-  @deprecated(
-    message = "Please use catsInstancesForHttp4sMethod. Kept for binary compatibility",
-    since = "0.21.14")
-  def catsHashForHttp4sMethod: Hash[Method] = catsInstancesForHttp4sMethod
-
-  @deprecated(
-    message = "Please use catsInstancesForHttp4sMethod. Kept for binary compatibility",
-    since = "0.21.5")
-  def http4sEqForMethod: Eq[Method] = catsInstancesForHttp4sMethod
-
-  @deprecated(
-    message = "Please use catsInstancesForHttp4sMethod. Kept for binary compatibility",
-    since = "0.21.14")
-  def http4sShowForMethod: Show[Method] = catsInstancesForHttp4sMethod
 }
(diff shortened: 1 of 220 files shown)


core/src/main/scala/org/http4s/Uri.scala
```diff
@@ -438,8 +469,11 @@ object Uri {
           )
       }
 
-    implicit val eq: Eq[Path] = Eq.fromUniversalEquals[Path]
-    implicit val semigroup: Semigroup[Path] = (a: Path, b: Path) => a.concat(b)
+    implicit val http4sInstancesForPath: Order[Path] with Semigroup[Path] =
+      new Order[Path] with Semigroup[Path] {
+        def compare(x: Path, y: Path): Int = x.compare(y)
+        def combine(x: Path, y: Path): Path = x.concat(y)
+      }
```
rossabaker, line 476: This is new, but not exciting.  There probably ought to be a `Hash[Path]` here. Either @isomarcte didn't include it because it used to be a type alias, or I misplaced it because I'm juggling too damned many branches.

isomarcte: > @isomarcte didn't include it because it used to be a type alias

Yeah, since it is just `String` on series/0.21, it wasn't needed to implement the Uri instance.

core/src/main/scala/org/http4s/Method.scala
```diff
@@ -147,10 +147,30 @@ object Method {
...
 
-  implicit val http4sShowForMethod: Show[Method] = Show.fromToString
+      override def compare(x: Method, y: Method): Int =
+        x.name.compare(y.name)
+    }
+
+  @deprecated(
+    message = "Please use catsInstancesForHttp4sMethod. Kept for binary compatibility",
+    since = "0.21.14")
+  def catsHashForHttp4sMethod: Hash[Method] = catsInstancesForHttp4sMethod
```
isomarcte, line 165: @rossabaker should we drop these since (I'm assuming) this is the 1.x.x branch?

core/src/main/scala/org/http4s/Uri.scala
```diff
@@ -289,6 +291,35 @@ object Uri {
       }
   }
 
+  object Authority {
+    def apply(
+        userInfo: Option[UserInfo] = None,
+        host: Host = RegName("localhost"),
+        port: Option[Int] = None): Authority =
+      new Authority(userInfo, host, port)
```
isomarcte, line 299: @rossabaker you can drop the `new` here now that we aren't using `AbstractFunction3`.

core/src/main/scala/org/http4s/Method.scala
rossabaker: Yes.  I like to keep deprecated things around a long time to ease people's migrations, but nobody should be calling these by name.  It's a good chance to clean up.

core/src/main/scala/org/http4s/Uri.scala
rossabaker: Wait, wouldn't that make a circular definition?

isomarcte: @rossabaker yes, sorry. You should be able to drop the whole function, if I'm not mistaken.

rossabaker: Oh.  Yes, I think that's true.  Let's see if I still think so after I make the change. :laughing:

Issue #3829, opened 2020-11-11
Enable implicit conversions


rossabaker:
Dotty is stricter than Scala 2 about implicit conversions.  We need to add `-language:implicitConversions`.

Maybe sbt-tpolecat does this, maybe we need to do it.

Target the `dotty` branch if you use `isDotty.value`, or target `series/0.21` if you don't.

dwlizlo:
-`language:implicitConversions` was removed from sbt-tpolecat for Dotty

dwlizlo:
@rossabaker could you please move it to in progress/done once it's merged? I don't think I have permission to shuffle the project around. Cheers!

rossabaker:
Fixed by #3975.

PR #3975, opened 2020-12-02
 Enable implicit conversions

dwlizlo:
https://github.com/http4s/http4s/issues/3829

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -749,6 +749,10 @@ def exampleProject(name: String) =
 
 lazy val commonSettings = Seq(
   Compile / doc / scalacOptions += "-no-link-warnings",
+  scalacOptions ++= {
+    if (isDotty.value) Seq("-language:implicitConversions")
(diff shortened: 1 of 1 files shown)


rossabaker (changes requested):
Thanks!

The change looks right, but could we do it without the reformatting?  It looks like you may have run scalafmtSbt, which is usually a good thing, and something I'd like to eventually adopt.  But we're maintaining a lot of branches right now, and those diffs create a lot of work later.  (It's also why I like trailing commas!)

dwlizlo:
@rossabaker Totally missed the fact I formated the file :shrug: I can see some builds are red, some green. Is the problem with dependencies something that is known?

rossabaker (approved):
Yes.  We're still wrangling dependencies.  Not even core compiles, and won't until we replace all the parsers.  But this gets us closer.  Thanks!

PR #3976, opened 2020-12-03
Add ContextResponse Type

isomarcte:
Similar to ContextRequest, but for Response[F].

I didn't add corresponding type aliases, e.g. 

```
type ResponseContextRoutes[T, F[_]] =  Kleisli[OptionT[F, *], Request[F], ContextResponse[F, T]]
```

The reason I didn't add this is that we'd probably also want to add a `FullContextRoutes` for something using context on both the request/response. Then we'd probably also want to add `ResponseContextMiddleware` types and helpers like we have in `object ContextRoutes`. 

All that to say, we will start to have many, many, helpers just to provide aliases for functions and types already available more generically in cats, albeit this may be a bit less accessible.

Given the use cases for these various permutations are likely to only effect a small number of users, it seemed like keeping the code base lighter would be a good idea.

### DIFF
--- a/core/src/main/scala/org/http4s/ContextResponse.scala
+++ b/core/src/main/scala/org/http4s/ContextResponse.scala
@@ -16,4 +16,6 @@ final case class ContextResponse[F[_], A](context: A, response: Response[F]) {
     ContextResponse(context, response.mapK(fk))
 }
 
-object ContextResponse{}
+// Include to avoid binary compatibility issues with the apply method if/when
+// we ever need a companion object in the future.
+object ContextResponse {}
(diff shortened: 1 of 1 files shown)


isomarcte:
@rossabaker is there anyway I can poke the tests to run again?

Right now, when they fail due to something unrelated, e.g. an inconsistent Blaze client test failure, I just amend the last commit to create a new commit hash which triggers github actions to run them again, but that is less than ideal.

rossabaker:
I don't know a good way.  Closing and reopening is another way, but even noisier.

I don't have metrics, but I _think_ it has gotten a little better with Carlos' work.  The real answer here is to have less flaky tests!

rossabaker:
Would #3977 be retrofitted onto this?

isomarcte:
> Would #3977 be retrofitted onto this?

I could do that and close this PR.

isomarcte:
Merged into #3977

PR #3980, opened 2020-12-05
sbt-http4s-org-0.6.0 (with sbt-spiewak-sonatype)

rossabaker:
Adopts sbt-spiewak under the covers, which sets up a lot for us.  Look before 5b2c2c5 to see the meaningful bits.

I was curious when each module was created.  It's trivia now, but is more relevant as we split up the repo.  That's the year you see in the header.  If anyone is interested in the `-2020` or liked the SPDX format, get it into sbt-spiewak.  I'm tired of special cases.

Before we publish again, we'll need to regenerate a GPG key without a passphrase.


rossabaker:
Ran off a local copy.  We'll need to try again after it syncs.

rossabaker:
The POM for the plugin didn't publish. :tired_face:

rossabaker:
Blocked by https://github.com/djspiewak/sbt-spiewak/pull/38

rossabaker:
Now blocked by https://github.com/http4s/sbt-http4s-org/pull/53

PR #3982, opened 2020-12-05
cats-parsed cookies

rossabaker:


core/src/main/scala/org/http4s/HttpDate.scala
```diff
@@ -155,6 +155,8 @@ object HttpDate {
     val dayName =
       List("Mon", "Tue", "Wed", "Thu", "Fri", "Sat", "Sun")
         .map(string1)
+        .zipWithIndex
+        .map { case (s, i) => string1(s).as(i) }
```
rossabaker, line 159: This didn't really need to be here.  It does if we tighten up weekday validation.

core/src/main/scala/org/http4s/ResponseCookie.scala
```diff
@@ -117,4 +120,175 @@ final case class ResponseCookie(
...
+    copy(secure = secure)
+
+  private def withHttpOnly(httpOnly: Boolean): ResponseCookie =
+    copy(httpOnly = httpOnly)
+
+  private def withSameSite(sameSite: SameSite): ResponseCookie =
+    copy(sameSite = sameSite)
+
+  private def addExtension(extension: String): ResponseCookie =
+    copy(extension = self.extension.fold(Option(extension))(old => Some(old + "; " + extension)))
```
rossabaker, line 146: Some of these could be generally useful and public.  Not all of them are validated except by the parser, which would permit the construction of invalid values.  Would it be confusing to expose just the safe ones?

core/src/main/scala/org/http4s/ResponseCookie.scala
```diff
@@ -117,4 +120,175 @@ final case class ResponseCookie(
...
+     */
+    val cookieOctet = charIn(
+      Set(0x21.toChar) ++
+        Set(0x23.toChar to 0x2b.toChar: _*) ++
+        Set(0x2d.toChar to 0x3a.toChar: _*) ++
+        Set(0x3c.toChar to 0x5b.toChar: _*) ++
+        Set(0x5d.toChar to 0x7e.toChar: _*))
+
+    /* cookie-value      = *cookie-octet / ( DQUOTE *cookie-octet DQUOTE ) */
+    val cookieValue = rep(cookieOctet).orElse(dquote *> rep(cookieOctet) <* dquote).string
```
rossabaker, line 174: I think the lack of semicolon in the dquoted part is exactly where people complain that we're too strict?

core/src/main/scala/org/http4s/ResponseCookie.scala
```diff
@@ -117,4 +120,175 @@ final case class ResponseCookie(
...
+     */
+    val maxAgeAv = ignoreCase1("Max-Age=") *> (nonZeroDigit ~ rep(digit)).string.map {
+      maxAge: String => (cookie: ResponseCookie) => cookie.withMaxAge(maxAge.toLong)
+    }
+
+    /* domain-value      = <subdomain>
+     *                   ; defined in [RFC1034], Section 3.5, as
+     *                   ; *enhanced by [RFC1123], Section 2.1
+     */
+    def domainValue = Rfc1034.subdomain
```
rossabaker, line 214: This could be a less stringy type in 1.0.

core/src/main/scala/org/http4s/ResponseCookie.scala
```diff
@@ -117,4 +120,175 @@ final case class ResponseCookie(
...
+      cookie.withSecure(true)
+    }
+
+    /* httponly-av       = "HttpOnly" */
+    val httponlyAv = ignoreCase1("HttpOnly").as { (cookie: ResponseCookie) =>
+      cookie.withHttpOnly(true)
+    }
+
+    /* samesite-value    = "Strict" / "Lax" / "None"
+     * from https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-06
```
rossabaker, line 251: This is from an expired draft spec with no successor.  @hamnis, does this still exist in the real world?  I think this was your addition.

core/src/main/scala/org/http4s/internal/parsing/Rfc2616.scala
```diff
@@ -0,0 +1,95 @@
...
+  /* CHAR           = <any US-ASCII character (octets 0 - 127)>
+   * CTL            = <any US-ASCII control character
+   *                  (octets 0 - 31) and DEL (127)>
+   * separators     = "(" | ")" | "<" | ">" | "@"
+   *                | "," | ";" | ":" | "\" | <">
+   *                | "/" | "[" | "]" | "?" | "="
+   *                | "{" | "}" | SP  | HT
+   * token          = 1*<any CHAR except CTLs or separators>
+   */
+  val token: Parser1[String] = {
```
rossabaker, line 28: I haven't done the math to figure out if this is equivalent to our other token, but this way, it's expressed fairly literally compared to the referenced RFC.

core/src/main/scala/org/http4s/internal/parsing/Rfc2616.scala
```diff
@@ -0,0 +1,95 @@
...
+  final case class Rfc1123Date(
+      wkday: Int,
+      day: Int,
+      month: Int,
+      year: Int,
+      hour: Int,
+      min: Int,
+      sec: Int)
+
+  val rfc1123Date: Parser1[Rfc1123Date] = {
```
rossabaker, line 46: This whole parser is redundant with `HttpDate`, and brings up an interesting case of having two specs parsing to the same type.  We could move this onto `HttpDate` under a different name.  This is one of the first types "parser on the companion" vs. "parser with the rest of the RFC" came into tension.

rossabaker: I think it's also redundant with one of the new parsers, but we've at least traced it to the spec.

core/src/main/scala/org/http4s/ResponseCookie.scala
rossabaker: I see the draft spec is active in in [http-extensions](https://github.com/httpwg/http-extensions/issues?q=label%3A6265bis+).

tests/src/test/scala/org/http4s/headers/CookieSuite.scala
```diff
@@ -0,0 +1,24 @@
+/*
+ * Copyright 2013-2020 http4s.org
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.http4s
+package headers
+
+class CookieSuite extends Http4sSuite {
```
rossabaker, line 10: The old spec was broken:

- No longer supported is a trailing semi-colon.  Did people really need this?
- Arbitrary whitespace around the semi-colon is forbidden.  We could maybe be more forgiving here.
- The RFC is clear that quoted values can't drop the quotes.  This might break apps?

tests/src/test/scala/org/http4s/headers/SetCookieHeaderSpec.scala
```diff
@@ -5,22 +5,23 @@
...
   "Set-Cookie parser" should {
     "parse a set cookie" in {
       val cookiestr =
-        "myname=\"foo\"; Domain=example.com; Max-Age=1; Path=value; SameSite=Strict; Secure;HttpOnly"
+        "myname=\"foo\"; Domain=example.com; Max-Age=1; Path=value; SameSite=Strict; Secure; HttpOnly"
       val c = parse(cookiestr).cookie
       c.name must be_==("myname")
       c.domain must beSome("example.com")
-      c.content must be_==("foo")
+      c.content must be_==(""""foo"""")
```
rossabaker, line 24: Like request cookies, the quotes are part of the value.

tests/src/test/scala/org/http4s/headers/SetCookieHeaderSpec.scala
```diff
@@ -5,22 +5,23 @@
...
-class SetCookieHeaderSpec extends Specification with HeaderParserHelper[`Set-Cookie`] {
-  def hparse(value: String): ParseResult[`Set-Cookie`] = HttpHeaderParser.SET_COOKIE(value)
+class SetCookieHeaderSpec extends Specification {
+  def parse(value: String): `Set-Cookie` = `Set-Cookie`.parse(value).valueOr(throw _)
 
   "Set-Cookie parser" should {
     "parse a set cookie" in {
       val cookiestr =
-        "myname=\"foo\"; Domain=example.com; Max-Age=1; Path=value; SameSite=Strict; Secure;HttpOnly"
+        "myname=\"foo\"; Domain=example.com; Max-Age=1; Path=value; SameSite=Strict; Secure; HttpOnly"
```
rossabaker, line 20: Like `Cookie`, we got stricter about whitespace around semi-colons.  Maybe relax this?

rossabaker (commented):
This one is scary because the spec is a disaster, browsers are a moving target, and we were previously relying on some undefined behavior.

We now have a fairly strict view of the current RFC.  Let's make sure we're relaxing it in the right ways so the 0.21 patch isn't traumatic and we don't get complaints about it eating people's (poorly written) cookies.

core/src/main/scala/org/http4s/ResponseCookie.scala
hamnis: it is very much in the real world. Chrome requires same-site to be set for third-party cookies.

PR #3983, opened 2020-12-05
Port more tests to munit

cquiroz:
This is another batch of tests ported to munit focused on tests using `Http4sLegacyMatchers`. I hope with another batch I can remove the legacy matcher classes.


client/src/test/scala/org/http4s/client/middleware/FollowRedirectSuite.scala
```diff
@@ -0,0 +1,240 @@
...
+
+  val defaultClient = Client.fromHttpApp(app)
+  val client = FollowRedirect(3)(defaultClient)
+
+  case class RedirectResponse(
+      method: String,
+      body: String
+  )
+
+  // test("FollowRedirect should ggfollow the proper strategy") {
```
cquiroz, line 61: This test was marked as pending and it is a bit hard to port as it uses the `Tables` feature of spec. Instead I just commented it out. Perhaps it should be removed

tests/src/test/scala/org/http4s/EntityDecoderSuite.scala
```diff
@@ -0,0 +1,491 @@
...
+    EntityDecoder.decodeText(resp)(implicitly, Charset.`UTF-8`).assertEquals(str)
+  }
+
+  // we want to return a specific kind of error when there is a MessageFailure
+  sealed case class ErrorJson(value: String)
+  implicit val errorJsonEntityEncoder: EntityEncoder[IO, ErrorJson] =
+    EntityEncoder.simple[IO, ErrorJson](`Content-Type`(MediaType.application.json))(json =>
+      Chunk.bytes(json.value.getBytes()))
+
+// TODO: These won't work without an Eq for (Message[IO], Boolean) => DecodeResult[IO, A]
```
cquiroz, line 481: This was commented in the original

tests/src/test/scala/org/http4s/EntityDecoderSuite.scala
```diff
@@ -0,0 +1,491 @@
...
+
+  test("Match valid media range") {
+    assert(EntityDecoder.text[IO].matchesMediaType(MediaType.text.plain))
+  }
+
+  test("Match valid media type to a range") {
+    assert(EntityDecoder.text[IO].matchesMediaType(MediaType.text.css))
+  }
+
+  /* TODO: Parameterization
```
cquiroz, line 210: It was commented in the original

client/src/test/scala/org/http4s/client/middleware/FollowRedirectSuite.scala
rossabaker: Drop it for now.  #3986 so it can be revived later, if we can.

rossabaker (approved):
Nice.  This is a thankless task, but I think it's moving us toward a better place.

rossabaker:
Do you want to take care of the merges again?  I already did dotty, because it was trivial.

cquiroz:
Sure I'm do main and cats effect 3 which I guess will still not compile

PR #3987, opened 2020-12-06
Port more tests to munit

cquiroz:
Replay of #3983 onto main


rossabaker:
This can also go to cats-effect-3, if you've got the time.  I imagine that's easier.

cquiroz:
Yes I'll send it there, let's see if at least some can be run there

rossabaker:
I think it's going to be tough until we get a vault, but I've been mostly heads down on the Dotty branch.

PR #3993, opened 2020-12-08
Replace `Accept-Ranges` parser with cats-parse implementation

fredshonorio:
This is a preliminary PR to request feedback, it does not fully replace `RangeParser` yet.

I added the cats-parse implementation to the  `Accept-Ranges` companion object and I'd like some general feedback on my usage of cats-parse and any formatting/style stuff.

I also have a question about the previous implementation, ([here](https://github.com/http4s/http4s/blob/c8d349dbf05af3ea76284801e49759778decaea9/core/src/main/scala/org/http4s/parser/RangeParser.scala#L65)) the top level rule for `Accept-Ranges` is matching on EOL  but this caused the cats-parse version to fail tests, I assume that parboiled2 is different from cats-effect and this is not necessary but I want to confirm this.

### DIFF
--- a/core/src/main/scala/org/http4s/headers/Accept-Ranges.scala
+++ b/core/src/main/scala/org/http4s/headers/Accept-Ranges.scala
@@ -6,8 +6,9 @@
 
 package org.http4s
 package headers
-
-import org.http4s.parser.HttpHeaderParser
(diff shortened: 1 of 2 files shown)


fredshonorio:
I see there's a formatting error, I'll fix it eventually

PR #3995, opened 2020-12-08
Replace `Accept-Ranges` parser with cats-parse implementation

fredshonorio:
This is a preliminary PR to request feedback, it does not fully replace `RangeParser` yet.

I added the cats-parse implementation to the  `Accept-Ranges` companion object and I'd like some general feedback on my usage of cats-parse and any formatting/style stuff.

I also have a question about the previous implementation, ([here](https://github.com/http4s/http4s/blob/c8d349dbf05af3ea76284801e49759778decaea9/core/src/main/scala/org/http4s/parser/RangeParser.scala#L65)) the top level rule for `Accept-Ranges` is matching on EOL  but this caused the cats-parse version to fail tests, I assume that parboiled2 is different from cats-effect and this is not necessary but I want to confirm this.

### DIFF
--- a/core/src/main/scala/org/http4s/headers/Accept-Ranges.scala
+++ b/core/src/main/scala/org/http4s/headers/Accept-Ranges.scala
@@ -23,7 +23,6 @@ object `Accept-Ranges` extends HeaderKey.Internal[`Accept-Ranges`] with HeaderKe
 
   /* https://tools.ietf.org/html/rfc7233#appendix-C */
   val parser: P[`Accept-Ranges`] = {
-    val ListSep: P[Unit] = Rfc7230.ows *> P.char(',') *> Rfc7230.ows
 
     val none = P.string1("none").as(Nil)
 
@@ -38,11 +37,11 @@ object `Accept-Ranges` extends HeaderKey.Internal[`Accept-Ranges`] with HeaderKe
       P.oneOf(
         List(
           none,
-          P.repSep(rangeUnit, 1, ListSep)
+          Rfc7230.headerRep1(rangeUnit).map(_.toList)
         )
       )
 
-    acceptableRanges.map(headers.`Accept-Ranges`.apply) /* is EOL necessary? */
+    acceptableRanges.map(headers.`Accept-Ranges`.apply)
   }
 }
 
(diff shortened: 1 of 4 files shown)


core/src/main/scala/org/http4s/headers/Accept-Ranges.scala
```diff
@@ -16,7 +17,33 @@ object `Accept-Ranges` extends HeaderKey.Internal[`Accept-Ranges`] with HeaderKe
...
 
   override def parse(s: String): ParseResult[`Accept-Ranges`] =
-    HttpHeaderParser.ACCEPT_RANGES(s)
+    parser.parseAll(s).leftMap { e =>
+      ParseFailure("Invalid accept-ranges", e.toString)
+    }
+
+  /* https://tools.ietf.org/html/rfc7233#appendix-C */
+  val parser: P[`Accept-Ranges`] = {
+    val ListSep: P[Unit] = Rfc7230.ows *> P.char(',') *> Rfc7230.ows
```
rossabaker, line 26: This will do a fine job on the ows-comma-ows that reasonable actors will send, but not the empty elements between commas (including leading and trailing commas) that the spec mandates.  `Rfc7230.headerRep1` should take care of those.

core/src/main/scala/org/http4s/headers/Accept-Ranges.scala
```diff
@@ -16,7 +17,33 @@ object `Accept-Ranges` extends HeaderKey.Internal[`Accept-Ranges`] with HeaderKe
...
+
+    val none = P.string1("none").as(Nil)
```
rossabaker, line 28: Good use of `as`.

core/src/main/scala/org/http4s/headers/Accept-Ranges.scala
```diff
@@ -16,7 +17,33 @@ object `Accept-Ranges` extends HeaderKey.Internal[`Accept-Ranges`] with HeaderKe
...
+     */
+    val acceptableRanges: P[List[RangeUnit]] =
+      P.oneOf(
+        List(
+          none,
+          P.repSep(rangeUnit, 1, ListSep)
+        )
+      )
+
+    acceptableRanges.map(headers.`Accept-Ranges`.apply) /* is EOL necessary? */
```
rossabaker, line 45: I think this is correct.  Calling `parseAll` means `Accept-Ranges.parse` will have to consume the entire input, and if this is composed into other parsers, we want it to leave the leftovers.

rossabaker (changes requested):
Thanks! 

This is a great start.  Replace the repSep, but I think you're on the right track here.

fredshonorio:
Last commit should implement requested changes.

rossabaker (approved):
I made a slight tweak to the error message, but this look great. Thanks!

PR #4000, opened 2020-12-09
Cats effect testing ce3

yanns:
https://github.com/djspiewak/cats-effect-testing/issues/117#issuecomment-742034458


rossabaker (approved):
Great, thanks!

PR #4004, opened 2020-12-11
OriginHeader Parser

lewisjkl:
This PR is still a WIP. I started with implementing the Parser for OriginHeader, but realized that I needed build a parser for ipv4/6 before I could finish that parser. I started with ipv6, and have a partially working implementation here. I tried to follow the implementation for the parboiled ipv6 parser as much as possible, but it seems I must have missed something. The parser works fine for more straight-forward ipv6 addresses, but starts to struggle with ones containing double colons (e.g. `2001:db8::8a2e:370:7334`). Would love another pair of eyes on this and any feedback/tips you can offer.


core/src/main/scala/org/http4s/internal/parsing/Rfc3986.scala
```diff
@@ -0,0 +1,86 @@
...
+  val ipv6: Parser1[Ipv6Address] = {
+
+    def toIpv6(lefts: collection.Seq[Short], rights: collection.Seq[Short]): Ipv6Address =
+      (lefts ++ collection.Seq.fill(8 - lefts.size - rights.size)(0.toShort) ++ rights) match {
+        case collection.Seq(a, b, c, d, e, f, g, h) =>
+          Ipv6Address(a, b, c, d, e, f, g, h)
+      }
+
+    val h16: Parser1[Short] = {
+      (hexdig ~ hexdig.? ~ hexdig.? ~ hexdig.?).string.map { (s: String) => java.lang.Integer.parseInt(s, 16).toShort }
```
lewisjkl, line 29: The parboiled parser didn't seem to make any of these hex digits optional, but without doing so, I am unsure how IPs like `684D:1111:222:3333:4444:5555:6:77` which contain fewer than 4 chars in a single grouping would parse correctly.

core/src/main/scala/org/http4s/internal/parsing/Rfc3986.scala
```diff
@@ -0,0 +1,86 @@
...
+    def repSix[A](p: Parser1[A], sep: Parser[Unit] = Parser.unit) = (p ~ sep ~ p ~ sep ~ p ~ sep ~ p ~ sep ~ p ~ sep ~ p)
+      .map { case ((((((((((one, _), two), _), three), _), four), _), five), _), six) => List(one, two, three, four, five, six) }
+    def repFive[A](p: Parser1[A], sep: Parser[Unit] = Parser.unit) = (p ~ sep ~ p ~ sep ~ p ~ sep ~ p ~ sep ~ p)
+      .map { case ((((((((one, _), two), _), three), _), four), _), five) => List(one, two, three, four, five) }
+    def repFour[A](p: Parser1[A], sep: Parser[Unit] = Parser.unit) = (p ~ sep ~ p ~ sep ~ p ~ sep ~ p)
+      .map { case ((((((one, _), two), _), three), _), four) => List(one, two, three, four) }
+    def repThree[A](p: Parser1[A], sep: Parser[Unit] = Parser.unit) = (p ~ sep ~ p ~ sep ~ p)
+      .map { case ((((one, _), two), _), three) => List(one, two, three) }
+    def repTwo[A](p: Parser1[A], sep: Parser[Unit] = Parser.unit) = (p ~ sep ~ p)
+      .map { case ((one, _), two) => List(one, two) }
```
lewisjkl, line 63: This feels quite sub-optimal to me. Anyone know of an easier way to accomplish this same thing?

core/src/main/scala/org/http4s/internal/parsing/Rfc3986.scala
```diff
@@ -0,0 +1,86 @@
...
+
+
+    (repSix(h16Colon) ~ ls32)
```
lewisjkl, line 66: From running tests with breakpoints, it seems like this first parser `(repSix(h16Colon) ~ ls32)` is the only one that really works.. I could be wrong, but it seems like there is something in all the others that is off.

core/src/main/scala/org/http4s/internal/parsing/Rfc3986.scala
rossabaker: Parboiled did it with the `(1 to 4).times` rule.  This is something that's hard to express in cats-parse right now:
https://github.com/typelevel/cats-parse/issues/97.

```
      def h16: Rule1[Short] = rule {
        capture((1 to 4).times(HexDigit)) ~> { (s: String) => java.lang.Integer.parseInt(s, 16).toShort }
      }
```

core/src/main/scala/org/http4s/internal/parsing/Rfc3986.scala
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    val decOctet = (char('1') ~ digit ~ digit)
+      .orElse1(char('2') ~ charIn('0' to '4') ~ digit)
+      .orElse1(string1("25") ~ charIn('0' to '5'))
+      .orElse1(charIn('1' to '9') ~ digit)
+      .orElse1(digit).string.map(_.toInt.toByte)
+
+    val ls32: Parser1[(Short, Short)] = {
+      val option1 = (h16 ~ char(':').void ~ h16).map { case ((l, _), r) => (l, r) }
```
rossabaker, line 39: If you're throwing away one side, use `<*` or `>*` instead of `~`, and you won't need to map.

core/src/main/scala/org/http4s/internal/parsing/Rfc3986.scala
```diff
@@ -0,0 +1,86 @@
...
+      val dot = char('.')
+      val option2: Parser1[(Short, Short)] = (decOctet ~ dot ~ decOctet ~ dot ~ decOctet ~ dot ~ decOctet).map {
```
rossabaker, line 41: Same with `<*` or `>*`.  You'll still have to map, but you'll lose the `_`'s.

core/src/main/scala/org/http4s/internal/parsing/Rfc3986.scala
```diff
@@ -0,0 +1,86 @@
...
+        case ((((((a, _), b), _), c), _), d) =>
+          ((a << 8) | b).toShort -> ((c << 8) | d).toShort
+      }
+      option1.orElse1(option2)
+    }
+
+    val colon = char(':')
```
rossabaker, line 48: You're using the `char` expression elsewhere.  Might as well take advantage of your constant.

core/src/main/scala/org/http4s/internal/parsing/Rfc3986.scala
rossabaker: I think you can do each one with `replicateA`.

core/src/main/scala/org/http4s/internal/parsing/Rfc3986.scala
rossabaker: This is a complicated parser.  I'd add one test at a time, and one `orElse` at a time, and see if that reveals the mistake.  If it doesn't, someone can jump in and help.

core/src/main/scala/org/http4s/parser/OriginHeader.scala
```diff
@@ -9,13 +9,20 @@ package parser
...
 import org.http4s.headers.Origin
 import org.http4s.internal.parboiled2._
+import cats.parse.{Parser, Parser1, Rfc5234}
+import org.http4s.Uri.RegName
+import cats.implicits._
 
 trait OriginHeader {
-  def ORIGIN(value: String): ParseResult[Origin] =
-    new OriginParser(value).parse
+  def ORIGIN(value: String): ParseResult[Origin] = {
```
rossabaker, line 21: Most of these are moving directly to the objects.  I'd move this to `Origin.parse`.  And if nothing else references the old parboiled parser, we should be able to delete this whole file. :tada:

tests/src/test/scala/org/http4s/parser/Rfc3986ParserSpec.scala
```diff
@@ -0,0 +1,22 @@
...
+
+import org.http4s.Http4sSpec
+import org.http4s.Uri.Ipv6Address
+import org.http4s.internal.parsing.Rfc3986
+import org.scalacheck.Prop
+
+class Rfc3986ParserSpec extends Http4sSpec {
+  "Rfc3986 parser" should {
+    "parse ipv6" in {
+      Prop.forAll(http4sTestingArbitraryForIpv6Address.arbitrary) { (ipv6: Ipv6Address) =>
```
rossabaker, line 17: It's great to have a property test, but as you work through, add one example of each shape at a time.  Keep it green, and build it back up!

rossabaker (changes requested):
Nice job.  I think ipv6 is the most difficult parser we'll have anywhere in the code.  A few comments, but this is on the right track.

core/src/main/scala/org/http4s/internal/parsing/Rfc3986.scala
lewisjkl: Gotcha, I figured there had to be an easier way 😅 Thanks for the tip!

core/src/main/scala/org/http4s/internal/parsing/Rfc3986.scala
lewisjkl: That is great advice, going to give that a try along with your other suggestions. Appreciate the help!

core/src/main/scala/org/http4s/internal/parsing/Rfc3986.scala
lewisjkl: As far as I have been able to tell, I need to use `~` in situations such as this one so I don't get rid of the first `h16` value. For example `(h16 ~ char(':') *> h16)` just returns `Short` instead of `(Short, Short)` like I need here. I may be missing the point here though, so please let me know!

core/src/main/scala/org/http4s/parser/OriginHeader.scala
lewisjkl: Looks like it is being used in `HttpHeaderParser` which is used in a bunch of places. Should I go through the process of removing it from `HttpHeaderParser`? Or leave it as is?

core/src/main/scala/org/http4s/internal/parsing/Rfc3986.scala
```diff
@@ -9,14 +9,27 @@ package org.http4s.internal.parsing
...
 
+  val ipv4: Parser1[Ipv4Address] = {
+    val decOctet = (char('1') ~ digit ~ digit).backtrack
+      .orElse1(char('2') ~ charIn('0' to '4') ~ digit).backtrack
+      .orElse1(string1("25") ~ charIn('0' to '5')).backtrack
+      .orElse1(charIn('1' to '9') ~ digit).backtrack
+      .orElse1(digit).string.map(_.toInt.toByte).backtrack
+
+    val dot = char('.')
+    (decOctet ~ dot ~ decOctet ~ dot ~ decOctet ~ dot ~ decOctet).map {
```
rossabaker, line 29: This is another one that can use `<*` or `*>` around the dots.  Then your case is only four parens instead of seven.

GitHub won't let me respond to your other question about getting a `Short` instead of a `(Short, Short)`.  I think it's an operator precedence issue.  I bet this would work:

`(decOctet ~ (dot *> decOctet) ~ (dot *> decOctet) ~ (dot *> decOctet))`

core/src/main/scala/org/http4s/parser/OriginHeader.scala
rossabaker: If it can be ripped out easily, great.  If not, let's bring this in for a landing and go further in another PR.

rossabaker (commented):
I'm not sure why the workflow is out of date, but `githubWorkflowGenerate` and check in the yaml will get CI further than it goes right now.

core/src/main/scala/org/http4s/internal/parsing/Rfc3986.scala
lewisjkl: Ah, that makes sense! Didn't think about the precedence issue. I will update that where it is beneficial. Thanks!

lewisjkl:
Was able to simplify those maps by using `*>` and `<*` and get rid of the `OriginHeader` trait. I also ran `githubWorkflowGenerate` and checked that in. Let me know if there is anything else that needs addressing here.

rossabaker:
I just went looking for this as I work on URI, and see that I dropped the ball on the final review.  Sorry for that.  I created this merge conflict pain for you, so I'll take care of that now.

rossabaker (approved):
Yeah, your changes look great. :+1: if it's green.

rossabaker:
Okay, I took care of the chores at the end.  Appreciate the hard work on this!

tests/src/test/scala/org/http4s/parser/Rfc3986ParserSpec.scala
```diff
@@ -0,0 +1,179 @@
...
+    "only parse part of an ipv6 address where multiple sections are shortened (only 1 allowed)" in {
+      val invalidIp = "56FE::2159:5BBC::6594"
+      Rfc3986.ipv6.parse(invalidIp).map(_._2.value) must beRight("56fe::2159:5bbc")
+    }
+
+    "parse an ipv6 address which is too long (only parse what can be used)" in {
+      val ipv6 = "56FE:2159:5BBC:6594:1234:5678:9018:0123:2020"
+      Rfc3986.ipv6.parse(ipv6).map(_._2.value) must beRight(
+        "56fe:2159:5bbc:6594:1234:5678:9018:123")
+    }
```
rossabaker, line 131: @lewisjkl Was there a particular reason to make a best effort on these instead of rejecting them as invalid?  I see some of these used to assert `Left`, and then changed around the time the `end` was dropped.  Are these cases we intentionally want to support, or just something that accidentally works now?

lewisjkl: The reason for this is because adding `end` into the parser made it so they couldn't compose with other parsers. For example, if you compose something like:

```scala
val myParser: Parser[String] = char('a').string <* end
val otherParser: Parser[String] = char('b').string

myParser ~ otherParser
```

I am pretty sure that this parser could never succeed since you can't have an `end` followed by more input. So I figured that the job of this `ipv6` parser would *not* be to reject extra input characters at the end of an ipv6 address, but rather that would have to be the job of a different parser that was aggregating this one as a part of its greater whole.

So come to think of it, OriginHeader should probably contain an `end` in its parser since we don't want anything to come after an ipv6/ipv4/host in that type of header. But it seems like ipv6/ipv4 should allow extra characters at the end of a valid address. What are your thoughts though?

rossabaker: Oh, right.  These tests are calling `.parse`, not `.parseAll`.  So you're asserting that you're grabbing the part that's legal, and discarding the rest, which should then fail in `.parseAll`, or when composed with another parser that doesn't understand the leftovers.  I think this is correct.

I also don't think we generally want to call `.end`.  The caller uses `.parseAll` when it expect to consume the entire String.  But if we don't cap the parser with `.end`, we let a generic header parser look for optional trailing whitespace, or a headers (plural) parser parse the unconsumed input as the next header.  `.parseAll` lets the caller express "the end", when only the caller knows at what level they wish to parse.

lewisjkl: Gotcha yeah that makes sense, I didn't actually realize `parseAll` was a thing. Thanks!

Issue #3835, opened 2020-11-11
Port tests to cats-effect-3


rossabaker:
This one will be large and unpleasant. We might want to discuss a way to split it up.

yanns:
I've started working on it: https://github.com/http4s/http4s/pull/4007

jonathanschoeller:
Did this change make it into a release of `http4s-testing`? I see that it's merged into `series/0.23` but can't seem to find a `0.23.x` release of `http4s-testing`.

armanbilge:
@jonathanschoeller we don't publish the http4s-testing module, although we do publish laws. Is there something specific you are looking for?

rossabaker:
There is no longer a published http4s-testing module.  Out of curiosity, what are you missing from it?  Maybe we can propose a good alternative.

jonathanschoeller:
I inherited a project that I'm in the process of migrating from cats-effect 2.x to 3.x and it had a build dependency on `http4s-testing` declared. On looking closer though, that dependency was always unnecessary and I can just remove it and continue upgrading the other http4s modules to 0.23.x. Thanks!

PR #4007, opened 2020-12-11
port tests module to cats-effect 3

yanns:
https://github.com/http4s/http4s/issues/3835


tests/src/test/scala/org/http4s/EntityCodecSpec.scala
```diff
@@ -32,3 +32,4 @@ class EntityCodecSpec extends Http4sSpec {
 
   checkAll("EntityCodec[IO, Unit]", EntityCodecTests[IO, Unit].entityCodec)
 }
+ */
```
rossabaker, line 35: @cquiroz I've lost track... do any of these new libraries floating around solve this problem?

tests/src/test/scala/org/http4s/EntityDecoderSpec.scala
```diff
@@ -301,7 +301,7 @@ class EntityDecoderSpec extends Http4sSpec with Http4sLegacyMatchersIO with Pend
 
     "invoke the function with  the right on a success" in {
       val happyDecoder: EntityDecoder[IO, String] =
-        EntityDecoder.decodeBy(MediaRange.`*/*`)(_ => DecodeResult.success(IO.pure("hooray")))
+        EntityDecoder.decodeBy(MediaRange.`**`)(_ => DecodeResult.success(IO.pure("hooray")))
```
rossabaker, line 304: Is this change to make it compile inside the block comment?  You can put another `*/` at the end, which looks imbalanced, but works.

Or at least you could in Scala 2.9 or whatever when I first had this problem. :laughing:

tests/src/test/scala/org/http4s/EntityCodecSpec.scala
yanns: By adding an instance for `Eq[F[_]]`, this test can run: https://github.com/http4s/http4s/pull/4007/commits/c19cdd2093cd313ef9fb1d2e308af337a7dfba74

Do you think it's a suitable solution?

rossabaker: Brief discussion on [cats-effect-dev Gitter](https://gitter.im/typelevel/cats-effect-dev?at=5fd4f24628a57c581b0bfb66).

I think this is fine for now, but I think in the long run Typelevel needs a discipline-effect built around scalacheck-effect's `PropF`.

yanns:
The remaining tests to be ported are blocked by the following error:
```
==> X org.http4s.MessageSuite.initializationError  0.034s java.lang.NoSuchMethodError: 'java.lang.Object io.chrisdavenport.vault.Key$.newKey(cats.effect.kernel.Sync)'
```

https://github.com/ChristopherDavenport/vault/issues/159

core/src/main/scala/org/http4s/StaticFile.scala
```diff
@@ -154,8 +154,8 @@ object StaticFile {
       etagCalc <- etagCalculator(f).map(et => ETag(et))
       res <- Files[F].isFile(f.toPath()).flatMap[Option[Response[F]]] { isFile =>
         if (isFile) {
-
-          if (start >= 0 && end >= start && buffsize > 0) {
+          val requireCondition = start >= 0 && end >= start && buffsize > 0
+          if (!requireCondition) {
```
rossabaker, line 158: Was this broken before?  This looks correct, but inverted from the previous?

rossabaker (commented):
Thanks! Glad to see this one about ready to land.

Should we be using `Sync.blocking` where we used to use a `testBlocker`, or is it okay to drop those?

core/src/main/scala/org/http4s/StaticFile.scala
yanns: It was inverted during the ce3 migration: https://github.com/http4s/http4s/commit/31b54aa0ce46a8450cf94eb1d0ddea7aeeacf6a8#diff-ffd60bd1163f5dd37a78811adfe0b03b0462f98d4c58041c46d11089232fc336L143-L145

From:
```
          require(
            start >= 0 && end >= start && buffsize > 0,
            s"start: $start, end: $end, buffsize: $buffsize")
```
to
```
          if (start >= 0 && end >= start && buffsize > 0) {
            F.raiseError[Option[Response[F]]](
              new IllegalArgumentException(
                s"requirement failed: start: $start, end: $end, buffsize: $buffsize"))
```
which is the opposite.

yanns:
> Should we be using `Sync.blocking` where we used to use a `testBlocker`, or is it okay to drop those?

I understood that we could drop those, but I'm not entirely sure. I'm asking on the ce channel.

RaasAhsan:
I'm running into this on the #3857 PR issue as well. Short answer: `Sync.blocking` is intended to be the successor to `Blocker`, but it isn't a complete replacement. `Blocker` lets you evaluate arbitrary effects on another pool (in the sense of `evalOn`), however, `Sync.blocking` only lets you evaluate a thunk `=> A` on the pool. It seems like http4s tends to run large, composite effects on the blocking pool, but for at least the cases I've seen, the truly blocking operations do happen on the blocking pool (via fs2).

So, I think we should do the following:
1. whether `Sync.blocking` is an appropriate alternative to where `Blocker` was used before
2. do we actually need to run large effects on the blocking pool or just simple thunks? the tradeoff here is we incur a lot of overhead from switching between pools if we need to evaluate and compose a bunch of blocking actions
3. If 1 and 2 don't pan out, we'll probably need to reintroduce something like `Blocker` via a separate `ExecutionContext`. maybe worth taking back upstream

djspiewak:
> do we actually need to run large effects on the blocking pool or just simple thunks? the tradeoff here is we incur a lot of overhead from switching between pools if we need to evaluate and compose a bunch of blocking actions

This is definitely something I'm interested in understanding better. `blocking` was very prescriptively designed with the assertion that this is not a valid use-case. If http4s disproves that assertion then we need a different API.

rossabaker (approved):
No, I think these are all legitimate.  We were threading all these blockers through to `fs2.io`, and it no longer requires them because it is using either `Sync.blocking` or NIO.  The objection was just my ignorance of CE3 and FS2-3.

PR #4012, opened 2020-12-12
Add CirceSensitiveDataEntityDecoder

kevinmeredith:
```scala
/** Derive [[EntityDecoder]] if implicit [[Decoder]] is in the scope without need to explicitly call `jsonOfSensitive`
  *
  * Note that it varies from [[CirceEntityDecoder]] in that, when failing to decode [[io.circe.Json]] to an `A`,
  * the JSON will not be included in the raised [[Exception]]. In the event the JSON includes sensitive data, this trait is,
  * arguably, a better choice since it eliminates the risk of logging sensitive data, e.g. due to logging a raised [[Throwable]]
  * that includes the sensitive JSON.
  */
```

### DIFF
--- a/circe/src/main/scala/org/http4s/circe/CirceInstances.scala
+++ b/circe/src/main/scala/org/http4s/circe/CirceInstances.scala
@@ -79,11 +79,15 @@ trait CirceInstances extends JawnInstances {
   def jsonOfWithMedia[F[_], A](r1: MediaRange, rs: MediaRange*)(implicit
       F: Sync[F],
       decoder: Decoder[A]): EntityDecoder[F, A] =
+  private def jsonOfWithMediaHelper[F[_], A](
+      r1: MediaRange,
+      decodeErrorHandler: (Json, NonEmptyList[DecodingFailure]) => DecodeFailure,
+      rs: MediaRange*)(implicit F: Sync[F], decoder: Decoder[A]): EntityDecoder[F, A] =
     jsonDecoderAdaptive[F](cutoff = 100000, r1, rs: _*).flatMapR { json =>
       decoder
         .decodeJson(json)
         .fold(
-          failure => DecodeResult.failure(jsonDecodeError(json, NonEmptyList.one(failure))),
+          failure => DecodeResult.failure(decodeErrorHandler(json, NonEmptyList.one(failure))),
           DecodeResult.success(_)
         )
     }
--- a/circe/src/main/scala/org/http4s/circe/CirceInstances.scala
+++ b/circe/src/main/scala/org/http4s/circe/CirceInstances.scala
@@ -217,6 +217,8 @@ object CirceInstances {
 
   private[circe] lazy val defaultJsonDecodeError
       : (Json, NonEmptyList[DecodingFailure]) => DecodeFailure = { (json, failures) =>
+    sensitiveDataJsonDecodeError(json, _.toString, failures)
+  }
 
   private def sensitiveDataJsonDecodeError(
       json: Json,
--- a/circe/src/main/scala/org/http4s/circe/CirceInstances.scala
+++ b/circe/src/main/scala/org/http4s/circe/CirceInstances.scala
@@ -80,6 +80,17 @@ trait CirceInstances extends JawnInstances {
       F: Sync[F],
       decoder: Decoder[A]): EntityDecoder[F, A] =
     jsonOfWithMediaHelper[F, A](r1, jsonDecodeError, rs: _*)
+
+  def jsonOfWithSensitiveMedia[F[_], A](r1: MediaRange, rs: MediaRange*)(implicit
+      F: Sync[F],
+      decoder: Decoder[A]): EntityDecoder[F, A] =
+    jsonOfWithMediaHelper[F, A](
+      r1,
+      (json, nelDecodeFailures) =>
+        CirceInstances.sensitiveDataJsonDecodeError(json, _ => "<REDACTED>", nelDecodeFailures),
+      rs: _*
+    )
+
   private def jsonOfWithMediaHelper[F[_], A](
       r1: MediaRange,
       decodeErrorHandler: (Json, NonEmptyList[DecodingFailure]) => DecodeFailure,
--- a/circe/src/main/scala/org/http4s/circe/CirceInstances.scala
+++ b/circe/src/main/scala/org/http4s/circe/CirceInstances.scala
@@ -84,13 +84,14 @@ trait CirceInstances extends JawnInstances {
       decoder: Decoder[A]): EntityDecoder[F, A] =
     jsonOfWithMediaHelper[F, A](r1, jsonDecodeError, rs: _*)
 
-  def jsonOfWithSensitiveMedia[F[_], A](r1: MediaRange, rs: MediaRange*)(implicit
+  def jsonOfWithSensitiveMedia[F[_], A](redact: Json => String, r1: MediaRange, rs: MediaRange*)(
+      implicit
       F: Sync[F],
       decoder: Decoder[A]): EntityDecoder[F, A] =
     jsonOfWithMediaHelper[F, A](
       r1,
       (json, nelDecodeFailures) =>
-        CirceInstances.jsonDecodeErrorHelper(json, _ => "<REDACTED>", nelDecodeFailures),
+        CirceInstances.jsonDecodeErrorHelper(json, redact, nelDecodeFailures),
       rs: _*
     )
 
--- a/circe/src/main/scala/org/http4s/circe/CirceInstances.scala
+++ b/circe/src/main/scala/org/http4s/circe/CirceInstances.scala
@@ -76,6 +76,9 @@ trait CirceInstances extends JawnInstances {
   def jsonOf[F[_]: Sync, A: Decoder]: EntityDecoder[F, A] =
     jsonOfWithMedia(MediaType.application.json)
 
+  def jsonOfSensitive[F[_]: Sync, A: Decoder]: EntityDecoder[F, A] =
+    jsonOfWithSensitiveMedia(MediaType.application.json)
+
   def jsonOfWithMedia[F[_], A](r1: MediaRange, rs: MediaRange*)(implicit
       F: Sync[F],
       decoder: Decoder[A]): EntityDecoder[F, A] =
--- a/circe/src/main/scala/org/http4s/circe/CirceInstances.scala
+++ b/circe/src/main/scala/org/http4s/circe/CirceInstances.scala
@@ -217,8 +217,17 @@ object CirceInstances {
 
   private[circe] lazy val defaultJsonDecodeError
       : (Json, NonEmptyList[DecodingFailure]) => DecodeFailure = { (json, failures) =>
+
+  private def sensitiveDataJsonDecodeError(
+      json: Json,
+      jsonToString: Json => String,
+      failures: NonEmptyList[DecodingFailure]
+  ): DecodeFailure = {
+
+    val str: String = jsonToString(json)
+
     InvalidMessageBodyFailure(
-      s"Could not decode JSON: $json",
+      s"Could not decode JSON: $str",
       if (failures.tail.isEmpty) Some(failures.head) else Some(DecodingFailures(failures)))
   }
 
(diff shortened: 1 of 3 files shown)


kevinmeredith:
I will add specs later today. Thinking about it more, I am not sure if one of the decoding failures could include sensitive data￼.

isomarcte (commented):
I wonder if adding a subtype of `DecodeFailure` which has `sanitized` and `details` components, similar to `ParseFailure`, would make sense to handle this case. That would allow redaction for other decoding failures with other media types as well.

circe/src/test/scala/org/http4s/circe/CirceSensitiveDataEntityDecoderSpec.scala
```diff
@@ -0,0 +1,72 @@
...
+  private object Person {
+    implicit val decoder: Decoder[Person] = new Decoder[Person] {
+      override def apply(c: HCursor): Decoder.Result[Person] =
+        c.downField("ssn").as[String].map(Person(_))
+    }
+  }
+
+}
+
+class CirceSensitiveDataEntityDecoderSpec extends Http4sSpec {
```
rossabaker, line 37: Can we define new specs with MUnit instead?  We've started a migration, and it would be nice if the new ones started in the new world.  It will also save you the `unsafeRunSync()`s.

rossabaker (commented):
The general idea of this ticket seems good.  I like @isomarcte's suggestion.  It has been a general principle to not put request information in messages.  Maybe that's too sweeping a change for 0.21, or an impossible change, but it seems like a better design.  I'd like to explore what it looks like and the fallout.

circe/src/test/scala/org/http4s/circe/CirceSensitiveDataEntityDecoderSpec.scala
kevinmeredith: Sure!

kevinmeredith: Can I blame Daniel Spiewak for my usage of specs2 here re: https://www.youtube.com/watch?v=jWJ5A1irH_E&list=PLSUh6oJ5ZotUS0k5d8EcHJIN9ATVJ_SfL? 😀

rossabaker: Oh, I've always liked specs2 as well.  Still do.  MUnit is just a bit more straightforward for truly asynchronous testing, and the literate DSLs have fallen out of fashion.

kevinmeredith:
> I wonder if adding a subtype of DecodeFailure which has sanitized and details components, similar to ParseFailure, would make sense to handle this case. That would allow redaction for other decoding failures with other media types as well.

I see the following sub-types of `DecodeFailure`:

Screen Shot 2020-12-15 at 9 55 20 PM

My initial guess is that, to support your idea on `0.21.x`, @isomarcte, that would involve, for all instances of `DecodeFailure`'s sub-types, to remove any sensitive data that's passed as a `message`, i.e. required by `org.http4s.MessageFailure`? Please say more, David. Thanks.

>  It has been a general principle to not put request information in messages.

Looking at `ParseFailure`, is `ParseFailure`'s comment on `details`, namley "may freely echo a Request," at odds with your "to not put request information" statement, @rossabaker?

```scala
/** Indicates an error parsing an HTTP [[Message]].
  *
  * @param sanitized May safely be displayed to a client to describe an error
  *                  condition.  Should not echo any part of a Request.
  * @param details Contains any relevant details omitted from the sanitized
  *                version of the error.  This may freely echo a Request.
  */
final case class ParseFailure(sanitized: String, details: String)
```

circe/src/test/scala/org/http4s/circe/CirceSensitiveDataEntityDecoderSpec.scala
kevinmeredith: Please see https://github.com/http4s/http4s/pull/4012/commits/beb9ee5f48fa8c20ead7f75555a57171112d9382, Ross.

isomarcte:
@kevinmeredith I was thinking of something like this.

```scala
final case class SanitizedDecodeFailure(sanitized: String, details: String, override val cause: Option[Throwable]) extends DecodeFailure {
  override def message: String = sanitized
}

object SanitizedDecodeFailure {
  def fromDecodeFailure(value: DecodeFailure, sanitize: String => String): SanitizedDecodeFailure = 
    value match {
      case value: SanitizedDecodeFailure => value
      case otherwise: DecodeFailure => 
        SanitizedDecodeFailure(sanitize(otherwise.message), otherwise.message, otherwise.cause)
    }
}
```

Then you could opt in any EntityDecoder like this,

```scala
val decoder: EntityDecoder[F, A] = ???
val sanitizedEntityDecoder: EntityDecoder[F, A] = decoder.handleErrorWith{
  case e => EitherT.leftT(SanitizedDecodeFailure.fromDecodeFailure(e))
}
```

rossabaker:
> Looking at ParseFailure, is ParseFailure's comment on details, namely "may freely echo a Request," at odds with your "to not put request information" statement, @rossabaker?

`sanitized` can describe in broad terms what's wrong with the request, but can't repeat any information.  For example, you can say "malformed JSON", but you can't say "Malformed boolean: expected true or false, got troo".  But details could give that latter level of detail.  Echoing request information into a response is dangerous for XSS attacks, so we maintain `sanitized` as a safe thing that can go in default responses and `details` as something reserved for server-side logging.  If people want the details because it's an intranet app, they're welcome to use them, but we maintain tighter security by default.

kevinmeredith:
Thanks, @rossabaker, for writing up https://github.com/http4s/http4s/pull/4012#issuecomment-747850280. I admit that I'm still foggy on the intent of `message`, `sanitized`, and `details`. However, I expect that I'll learn more as I follow any further discussions of `SanitizedDecodeFailure`.

Thanks for sharing https://github.com/http4s/http4s/pull/4012#issuecomment-747527228, @isomarcte. 

I still admit that I don't entirely understand how `SanitizedDecodeFailure` would be threaded throughout the `0.21.x` code-base. `sanitize: String => String` seems a bit coarse to me, i.e. sanitizing a `String` rather than a, let's say an `A`, yet I don't have any better ideas.

Given that, it seems to me that a few options exist:

1. merge + cut a release of this PR since it solves a problem at-hand of logging JSON (which can include sensitive data) that fails to decode to `A`
2. @isomarcte opens a PR against this branch or a separate PR against `main` that adds `SanitizedDecodeFailure`, as well as implements an equivalent `CirceSensitiveDataEntityDecoder` using `SanitizedDecodeFailure`

Or, another option, it seems to me, is to do step 1 first, and then step 2 afterwards.

What do you think, @rossabaker @isomarcte?

kevinmeredith:
Hi @rossabaker and @isomarcte - any thoughts, please, on https://github.com/http4s/http4s/pull/4012#issuecomment-748413263? In short, I'm proposing moving forward with this PR as-is as it provides a useful `EntityDecoder[F, A]` to `import` to avoid logging sensitive data. My partial understanding is that, if `SanitizedDecodeFailure` gets added, then the internals of `CirceSensitiveDataEntityDecoder` could, perhaps, be re-factored to use such an approach.

isomarcte:
@kevinmeredith I'm fine with that, though my opinion doesn't carry much weight. I think we could add `SanitizedDecodeFailure` in series/0.21, and there may even be an argument that it should be used by default for all `EntityDecoder` instances, but if you want to go this route in the meantime I think it is still a net gain.

I doubt I'll get to making a PR for `SanitizedDecodeFailure` for a bit still if you can't/don't want to take a pass at it @kevinmeredith .

circe/src/main/scala/org/http4s/circe/CirceInstances.scala
```diff
@@ -76,14 +76,33 @@ trait CirceInstances extends JawnInstances {
...
       decoder: Decoder[A]): EntityDecoder[F, A] =
+    jsonOfWithMediaHelper[F, A](r1, jsonDecodeError, rs: _*)
+
+  def jsonOfWithSensitiveMedia[F[_], A](r1: MediaRange, rs: MediaRange*)(implicit
+      F: Sync[F],
+      decoder: Decoder[A]): EntityDecoder[F, A] =
+    jsonOfWithMediaHelper[F, A](
+      r1,
+      (json, nelDecodeFailures) =>
+        CirceInstances.jsonDecodeErrorHelper(json, _ => "<REDACTED>", nelDecodeFailures),
```
rossabaker, line 93: The `String => String` sanitization is a fair concern on the `SanitizedDecodeFailure`.  But we're just doing a constant function here.  To take advantage of having a `Json`, would it make sense to lift this function up as an argument, so we could sanitize the sensitive fields and still be able to debug the rest?

A counterargument to that is we're here due to an invalid JSON structure, and the sensitive bits might be under a JSON key with a typo.

kevinmeredith: > To take advantage of having a Json, would it make sense to lift this function up as an argument, so we could sanitize the sensitive fields and still be able to debug the rest?

If I understand properly, you're proposing re-writing `CirceSensitiveDataEntityDecoder` as:

```scala
trait CirceSensitiveDataEntityDecoder {
  def circeEntityDecoder[F[_]: Sync, A: Decoder](sanitize: Json => String): EntityDecoder[F, A] =
    jsonOfSensitive[F, A]
}
```

, @rossabaker?

I'm not convinced that's your suggestion since we'd lose the `implicit`.

> A counterargument to that is we're here due to an invalid JSON structure, and the sensitive bits might be under a JSON key with a typo.

Yep - good point.

rossabaker: I was proposing parameterizing `jsonOfWithSensitiveMedia` with that function, and fixing up the compile errors that bubble up.  You could use your `_ => "<REDACTED>"` as a default.  I might be underestimating all the places that ripples.

kevinmeredith: Hey @rossabaker - I like that idea of passing an argument of `Json => String`. 

Can you please review https://github.com/http4s/http4s/pull/4012#commits-pushed-cba03fe?

I opted against the default parameter, but, please let me know what you prefer.

circe/src/main/scala/org/http4s/circe/CirceSensitiveDataEntityDecoder.scala
```diff
@@ -0,0 +1,35 @@
...
+import org.http4s.EntityDecoder
+
+/** Derive [[EntityDecoder]] if implicit [[Decoder]] is in the scope without need to explicitly call `jsonOfSensitive`
+  *
+  * Note that it varies from [[CirceEntityDecoder]] in that, when failing to decode [[io.circe.Json]] to an `A`,
+  * the JSON will not be included in the raised [[Exception]]. In the event the JSON includes sensitive data, this trait is,
+  * arguably, a better choice since it eliminates the risk of logging sensitive data, e.g. due to logging a raised [[Throwable]]
+  * that includes the sensitive JSON.
+  */
+trait CirceSensitiveDataEntityDecoder {
```
rossabaker, line 30: I'm not convinced I like this, when it's parameterized everywhere else.  But I also don't like the implicit JSON codecs anyway, so I might be overly grumpy.  I'm curious what other people think.

rossabaker (approved):
I left one crotchety note, but I think it's more me not liking aspects of prior design than anything done wrong here, so I'm not going to let it block progress.  I'll leave it open briefly to see if anyone else has thoughts.

The rest of this looks good.

circe/src/main/scala/org/http4s/circe/CirceSensitiveDataEntityDecoder.scala
kevinmeredith: > when it's parameterized everywhere else

Could you please say more, Ross? I don't follow.

rossabaker: We've surfaced that `Json => String` everywhere else, but this one is hardcoded to the constant function.  And for it to be implicit, it must be hardcoded to some function.  You can't fix it and still retain the utility of this class.

I think implicit entity codecs from implicit circe codecs are a mistake, so I think my complaint is just a pre-existing condition here. :smile:

kevinmeredith: > I think implicit entity codecs from implicit circe codecs are a mistake, so I think my complaint is just a pre-existing condition here. 😄

Gotcha, 😀! So, philosophically, you believe the codecs should be defined in the actual application￼.

kevinmeredith:
Hey Ross - is this one good to go? Thanks!

rossabaker:
I was seeing if anyone else wanted to weigh in, but it's been open a while. :+1:

Issue #3984, opened 2020-12-05
Advent of cats-parse


rossabaker:
These files have parboiled imports that need to go.  We are moving them to cats-parse.  Some of these files are related. Some are enormous and can be broken down.  They could be tasked better, but this is a start.

All work should target the series-0.22 branch, but all of this will get it closer.

Basic steps:

1. We are putting [parsers on the companion objects](https://github.com/http4s/http4s/blob/79ed398dfe6da5988e7752b0943c16b749736f38/core/src/main/scala/org/http4s/HttpDate.scala#L127-L130) of the type they parse, as `private[http4s]`.  By being on the companion, they can invoke private constructors with validated values.  Commenting on the RFC rules each parser implements is nice, as it makes it easier to trace back to the spec.

2. Replace the `.fromString` or `.parse` method on the companion with [code like this](https://github.com/http4s/http4s/blob/79ed398dfe6da5988e7752b0943c16b749736f38/core/src/main/scala/org/http4s/HttpDate.scala#L83-L86).

3. Delete any parboiled parsers left behind.

4. Run `core/mimaReportBinaryIssues` and [add MiMa exclusions](https://github.com/http4s/http4s/blob/dotty/build.sbt#L101).  They were all private, so it's safe.

5. If you find common rules, we are [putting some RFC objects](https://github.com/http4s/http4s/tree/dotty/core/src/main/scala/org/http4s/internal/parsing) in an internal package to share.

6. Some of these are trivial and some are very hard.  If you get in over your head, [we're on Gitter](https://gitter.im/http4s/http4s-dev) or we're here.

If you're interested in one, please comment:

* [x] ./core/src/main/scala/org/http4s/ContentCoding.scala (@novakov-alexey)
* [x] ./core/src/main/scala/org/http4s/MediaType.scala (@novakov-alexey)
* [x] ./core/src/main/scala/org/http4s/QValue.scala (@novakov-alexey)
* [x] ./core/src/main/scala/org/http4s/Query.scala (@rossabaker)
* [x] ./core/src/main/scala/org/http4s/TransferCoding.scala (#4023)
* [x] ./core/src/main/scala/org/http4s/Uri.scala (@rossabaker)
* [x] ./core/src/main/scala/org/http4s/parser/AcceptCharsetHeader.scala (@novakov-alexey)
* [x] ./core/src/main/scala/org/http4s/parser/AcceptEncodingHeader.scala (@novakov-alexey)
[...]

PR #4013, opened 2020-12-12
move ContentCoding, QValue and other headers to cats-parse

novakov-alexey:
Initial part of #3984 for:
- org.http4s.ContentCoding
- org.http4s.QValue

Next step:
I am blocked by `org.http4s.parser.Http4sHeaderParser#entry` to cleanup Parboiled from `ContentCoding` and `QValue`.

I think we need to remove Parboiled from `Http4sHeaderParser` and `Http4sParser`classes and then make the final removal in the rest of the parsers, which depend on abstract `entry` method.


core/src/main/scala/org/http4s/QValue.scala
```diff
@@ -98,10 +101,29 @@ object QValue {
...
     fromString(s).fold(throw _, identity)
 
+  private[http4s] val parser: Parser[QValue] = {
+    import cats.parse.Parser.{char => ch, _}
+    import cats.parse.Rfc5234._
+
+    val optWS = (crlf.rep.with1 ~ wsp.rep1).rep
+
+    val qValue = string1(ch('0') ~ (ch('.') ~ digit.rep1).rep)
+      .map(
```
rossabaker, line 111: Can we use `.mapFilter` here and avoid throwing an exception?  It's new in cats-parse-0.2.

core/src/main/scala/org/http4s/QValue.scala
```diff
@@ -98,10 +101,29 @@ object QValue {
...
+        QValue
+          .fromString(_)
```
rossabaker, line 113: Wow.  Our `.fromString` is laxer than the spec: a qValue can only have a precision of three digits.  I guess that laxity probably isn't hurting anyone?

core/src/main/scala/org/http4s/QValue.scala
```diff
@@ -98,10 +101,29 @@ object QValue {
   def unsafeFromString(s: String): QValue =
     fromString(s).fold(throw _, identity)
 
+  private[http4s] val parser: Parser[QValue] = {
+    import cats.parse.Parser.{char => ch, _}
+    import cats.parse.Rfc5234._
+
+    val optWS = (crlf.rep.with1 ~ wsp.rep1).rep
+
+    val qValue = string1(ch('0') ~ (ch('.') ~ digit.rep1).rep)
```
rossabaker, line 110: `*>` or `<*` instead of `~` if you're not using the tuple you build.

core/src/main/scala/org/http4s/QValue.scala
```diff
@@ -98,10 +101,29 @@ object QValue {
...
+      .map(
+        QValue
+          .fromString(_)
+          .valueOr(e => throw e.copy(sanitized = "Invalid q-value"))
+      )
+      .orElse1(
+        ch('1') *> (ch('.') *> ch('0').rep).rep.as(One)
+      )
+
+    (ch(';') *> optWS *> ch('q') *> ch('=') *> qValue).orElse(pure(One))
```
rossabaker, line 120: I forget: is `q` case insensitive?

core/src/main/scala/org/http4s/QValue.scala
```diff
@@ -98,10 +101,29 @@ object QValue {
...
+    val optWS = (crlf.rep.with1 ~ wsp.rep1).rep
+
+    val qValue = string1(ch('0') ~ (ch('.') ~ digit.rep1).rep)
+      .map(
+        QValue
+          .fromString(_)
+          .valueOr(e => throw e.copy(sanitized = "Invalid q-value"))
+      )
+      .orElse1(
+        ch('1') *> (ch('.') *> ch('0').rep).rep.as(One)
```
rossabaker, line 117: What's the second `.rep` for?

rossabaker (changes requested):
Nice start!  Thanks for taking this on.

In the headers, I've been using a pattern like this:

```scala
  override def parse(s: String): ParseResult[`If-Modified-Since`] =
    parser.parseAll(s).leftMap { e =>
      ParseFailure("Invalid If-Modified-Since header", e.toString)
    }

  /* `If-Modified-Since = HTTP-date` */
  private[http4s] val parser: Parser1[`If-Modified-Since`] =
    HttpDate.parser.map(apply)
```

Then there's no reference to the parboiled header parsers, and they can be removed and granted a MiMa exception.

core/src/main/scala/org/http4s/QValue.scala
novakov-alexey: Yeah, I have not touched `def fromString(s: String): ParseResult[QValue]` since it is public API. Not sure if it is still makes sense

core/src/main/scala/org/http4s/QValue.scala
novakov-alexey: As per: https://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html `All content-coding values are case-insensitive.`

core/src/main/scala/org/http4s/QValue.scala
novakov-alexey: The second `.rep` is to make a fractional part ".000" optional. Not sure how to do this properly in cats-parse

core/src/main/scala/org/http4s/QValue.scala
novakov-alexey: Ok, I think  you mean something like that?
```
.mapFilter(
        QValue
          .fromString(_)
          .toOption
      )
```

novakov-alexey:
> Nice start! Thanks for taking this on.
> 
> In the headers, I've been using a pattern like this:
> 
> ```scala
>   override def parse(s: String): ParseResult[`If-Modified-Since`] =
>     parser.parseAll(s).leftMap { e =>
>       ParseFailure("Invalid If-Modified-Since header", e.toString)
>     }
> 
>   /* `If-Modified-Since = HTTP-date` */
>   private[http4s] val parser: Parser1[`If-Modified-Since`] =
>     HttpDate.parser.map(apply)
> ```
> 
> Then there's no reference to the parboiled header parsers, and they can be removed and granted a MiMa exception.

I mean that `ContentCoding` depends on `QValue`. The `AcceptCharsetHeader`, `AcceptLanguageHeader` depend on `QValue`. Also, `AcceptEncodingHeader`, `SimpleHeaders` depends on `ContentCoding`. Should I fix all other Headers in the same pull-request?

core/src/main/scala/org/http4s/QValue.scala
novakov-alexey: Done

core/src/main/scala/org/http4s/QValue.scala
novakov-alexey: Changed to `ignoreCaseChar('q')`

novakov-alexey:
AcceptCharsetHeader.scala, AcceptEncodingHeader.scala, AcceptLanguageHeader.scala have been migrated to cats-parse as well.

core/src/main/scala/org/http4s/parser/AdditionalRules.scala
```diff
@@ -205,3 +206,10 @@ private[http4s] trait AdditionalRules extends Rfc2616BasicRules { this: Parser =
     }
   }
 }
+
+private[http4s] object AdditionalRules {
+  import Rfc2616BasicRules._
+  def EOI = P.char('\uFFFF')
```
novakov-alexey, line 212: I put cats-parse based parsers here as temporary place, until AdditionalRules trait is fully migrated.

core/src/main/scala/org/http4s/parser/Rfc2616BasicRules.scala
```diff
@@ -80,6 +83,12 @@ private[http4s] trait Rfc2616BasicRules extends Parser {
 }
 
 private[http4s] object Rfc2616BasicRules {
+  def lws = (crlf.rep.with1 *> wsp.rep1).void
```
novakov-alexey, line 86: temporary place for cats-parse parsers.

core/src/main/scala/org/http4s/QValue.scala
rossabaker: Yes, that looks right.  We want to avoid throwing exceptions, and `mapFilter` uses a selective functor, which is more efficient than flatMap in cats-parse.

core/src/main/scala/org/http4s/QValue.scala
rossabaker: Let's leave it for this PR and we'll consider what we want in 1.0.

core/src/main/scala/org/http4s/QValue.scala
rossabaker: Oh, right. It's `.?`.

rossabaker:
That build error can be fixed with a `githubWorfklowGenerate`.

core/src/main/scala/org/http4s/QValue.scala
novakov-alexey: Ok

novakov-alexey:
> That build error can be fixed with a `githubWorfklowGenerate`.

Thanks, I was about to ask how to fix the build :-)

core/src/main/scala/org/http4s/QValue.scala
novakov-alexey: Thanks. Done

core/src/main/scala/org/http4s/QValue.scala
novakov-alexey: Done

novakov-alexey:
3.0.0-M1 compile task failed as expected.

core/src/main/scala/org/http4s/parser/AcceptCharsetHeader.scala
```diff
@@ -11,35 +11,38 @@
...
-    new AcceptCharsetParser(value).parse
-
-  private class AcceptCharsetParser(input: ParserInput)
-      extends Http4sHeaderParser[headers.`Accept-Charset`](input)
-      with QValueParser {
-    def entry: Rule1[headers.`Accept-Charset`] =
-      rule {
-        oneOrMore(CharsetRangeDecl).separatedBy(ListSep) ~ EOL ~> { (xs: Seq[CharsetRange]) =>
-          headers.`Accept-Charset`(xs.head, xs.tail: _*)
+  def ACCEPT_CHARSET(value: String): ParseResult[`Accept-Charset`] =
```
rossabaker, line 20: I've been putting the cats-parse parsers on the companion object of whatever is being parsed.  This way, it has access to private constructors to only accept valid values.  But this works for now.

core/src/main/scala/org/http4s/parser/AcceptCharsetHeader.scala
```diff
@@ -11,35 +11,38 @@
...
+    import Rfc2616BasicRules._
+    import cats.parse.Parser._
+    import org.http4s.internal.parsing.Rfc7230.token
+
+    val anyCharset = (char('*') *> QValue.parser)
+      .map(q => if (q != QValue.One) `*`.withQValue(q) else `*`)
+
+    val fromToken = (token ~ QValue.parser).mapFilter { case (s, q) =>
+      // TODO handle tokens that aren't charsets
+      Charset
```
rossabaker, line 35: I think `charset` appears in multiple places, and we could put a parser on it.

core/src/main/scala/org/http4s/parser/AcceptEncodingHeader.scala
```diff
@@ -11,22 +11,21 @@
...
-      rule {
-        oneOrMore(EncodingRangeDecl).separatedBy(ListSep) ~ EOL ~> { (xs: Seq[ContentCoding]) =>
-          `Accept-Encoding`(xs.head, xs.tail: _*)
-        }
-      }
+  private[http4s] val acceptEncodingParser: Parser1[`Accept-Encoding`] = {
+    import Rfc2616BasicRules._
+    import cats.parse.Parser._
+
+    (rep1Sep(ContentCoding.parser, 1, listSep) <* AdditionalRules.EOL).map(xs =>
```
rossabaker, line 28: I don't _think_ we want the EOL on these.  If we were parsing multiple headers at once, we would have the `\r\n` be a separator between headers, and not consumed by this.

core/src/main/scala/org/http4s/parser/AcceptCharsetHeader.scala
```diff
@@ -11,35 +11,38 @@
...
-            // TODO handle tokens that aren't charsets
-            val c = Charset.fromString(s).valueOr(throw _)
-            if (q != org.http4s.QValue.One) c.withQuality(q) else c.toRange
-          })
-      }
+    }
+
+    val charsetRange = anyCharset.orElse1(fromToken)
+
+    (rep1Sep(charsetRange, 1, listSep) <* AdditionalRules.EOL).map(xs =>
```
rossabaker, line 45: I think you want `Rfc72301#headerRep1`, not `rep1Sep` here. That's the expansion of `1#element` in the RFCs, which is surprisingly complex.

rossabaker (commented):
Looking good.  Just a couple more open items.

core/src/main/scala/org/http4s/parser/AcceptCharsetHeader.scala
novakov-alexey: can you advise what are the other places?

core/src/main/scala/org/http4s/parser/AcceptCharsetHeader.scala
novakov-alexey: Done

core/src/main/scala/org/http4s/parser/AcceptCharsetHeader.scala
novakov-alexey: Fixed

core/src/main/scala/org/http4s/parser/AcceptEncodingHeader.scala
novakov-alexey: Ok, EOL is removed

core/src/main/scala/org/http4s/parser/AcceptCharsetHeader.scala
rossabaker: I thought there were a lot more, but maybe `Content-Type` is the last of them?

I think we can address that in a follow-up after this gets merged.

core/src/main/scala/org/http4s/headers/Accept-Encoding.scala
```diff
@@ -18,12 +18,23 @@ package org.http4s
...
-    HttpHeaderParser.ACCEPT_ENCODING(s)
+    parser.parseAll(s).leftMap { e =>
+      ParseFailure("Invalid Accept Encoding header", e.toString)
+    }
+
+  private[http4s] val parser: Parser1[`Accept-Encoding`] = {
+    import Rfc2616BasicRules._
+    import cats.parse.Parser._
+
+    rep1Sep(ContentCoding.parser, 1, listSep).map(xs => `Accept-Encoding`(xs.head, xs.tail: _*))
```
rossabaker, line 36: I think this is another `headerRep1`.

core/src/main/scala/org/http4s/headers/Accept-Language.scala
```diff
@@ -18,11 +18,29 @@ package org.http4s
...
+    import Rfc2616BasicRules._
+    import cats.parse.Parser.{char => ch, _}
+    import cats.parse.Rfc5234._
+
+    val languageTag =
+      (string1(alpha.rep1(1)) ~ (ch('-') *> Rfc7230.token).rep ~ QValue.parser).map {
+        case ((main, sub), q) => LanguageTag(main, q, sub)
+      }
+
+    rep1Sep(languageTag, 1, listSep).map(tags => `Accept-Language`(tags.head, tags.tail: _*))
```
rossabaker, line 42: Another `headerRep1`.

rossabaker (commented):
This is looking good.  The build will continue to fail on 3.0.0 until this parser effort is over, plus a couple other things.  I think we just need two more tweaks.

core/src/main/scala/org/http4s/parser/AcceptCharsetHeader.scala
novakov-alexey: Wow, eveything is quite interveawed :-) 
The `Content-Type` depends on the `MediaType.scala` as well. I can take this as next parser to convert. Just let me know if it makes to continue in this PR or merge first.

core/src/main/scala/org/http4s/headers/Accept-Encoding.scala
novakov-alexey: Fixed

core/src/main/scala/org/http4s/headers/Accept-Language.scala
novakov-alexey: Fixed

core/src/main/scala/org/http4s/parser/AcceptCharsetHeader.scala
rossabaker: Yeah, the spec is complicated, and parboiled, great as it was, didn't compose well.  I'm unsurprised we're running into tangled cords here.

I'd rather get this landed and fix that in the next.

rossabaker (approved):

core/src/main/scala/org/http4s/parser/AcceptCharsetHeader.scala
novakov-alexey: ok

PR #4018, opened 2020-12-16
Use Stream.Compiler[F, F] where possible

hamnis:
We cant use this for Files or Multipart as that requires Blocker and ContextShift.

Is this useful?

### DIFF
--- a/circe/src/main/scala/org/http4s/circe/CirceInstances.scala
+++ b/circe/src/main/scala/org/http4s/circe/CirceInstances.scala
@@ -44,13 +44,19 @@ trait CirceInstances extends JawnInstances {
   protected def jsonDecodeError: (Json, NonEmptyList[DecodingFailure]) => DecodeFailure =
     CirceInstances.defaultJsonDecodeError
 
-  def jsonDecoderIncremental[F[_]](implicit C: Stream.Compiler[F, F], A: ApplicativeThrow[F]): EntityDecoder[F, Json] =
+  def jsonDecoderIncremental[F[_]](implicit
+      C: Stream.Compiler[F, F],
+      A: ApplicativeThrow[F]): EntityDecoder[F, Json] =
     this.jawnDecoder[F, Json]
 
-  def jsonDecoderByteBuffer[F[_]](implicit C: Stream.Compiler[F, F], A: ApplicativeThrow[F]): EntityDecoder[F, Json] =
+  def jsonDecoderByteBuffer[F[_]](implicit
+      C: Stream.Compiler[F, F],
+      A: ApplicativeThrow[F]): EntityDecoder[F, Json] =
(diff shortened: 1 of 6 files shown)


circe/src/main/scala/org/http4s/circe/CirceInstances.scala
```diff
@@ -43,13 +42,19 @@ trait CirceInstances extends JawnInstances {
   protected def jsonDecodeError: (Json, NonEmptyList[DecodingFailure]) => DecodeFailure =
     CirceInstances.defaultJsonDecodeError
 
-  def jsonDecoderIncremental[F[_]: Sync]: EntityDecoder[F, Json] =
+  def jsonDecoderIncremental[F[_]](implicit
+      C: Stream.Compiler[F, F],
+      A: ApplicativeThrow[F]): EntityDecoder[F, Json] =
```
Daenyth, line 47: The only `F` that satisfies both is one which has `Sync`, so I don't think this helps. Also, it would break backwards compatibility

rossabaker:
There's an extended [Gitter conversation](https://gitter.im/http4s/http4s-dev?at=5fd9f0c7b685314fe783cac6).

circe/src/main/scala/org/http4s/circe/CirceInstances.scala
rossabaker: There is also a `Stream.Compiler[Id, Id]`, which I think is what motivated @hamnis.

rossabaker:
So I don't think a `Compiler[Id, Id]` is all that compelling outside tests, because there is no error channel.  Our `IO` testing frameworks have gotten a lot better and reflect a more realistic use.  I generally preach "don't put in IO what doesn't need to be in IO," but I also like testing idiomatic usage.

The two-parameter constraint is annoying, but fs2-3.x introduces a `Compiler.Target[F]`.  This can be used as a constraint in place of the usual `Sync` (fs2-2.x) or `Concurrent` (encouraged in fs2-3.x).  It also precludes `Id`, which I think was a big motivator for this.  It might permit `Http[Either[Throwable, *], Either[Throwable, *]]`, but, to what end?  Fallible, non-side-effecting, synchronous services?  

@SystemFw argues that the conceptual burden of a new kind of constraint is offputting to newcomers, and I'm sympathetic to that argument.  These days I'm reviewing a lot more Scala than I'm writing, so I appreciate those reminders from the trenches.  He also argues that `Compiler` isn't really intended as a library constraint.  It looks like it has been made more inviting in 3.x, but I'm still dubious of the benefit.

I've tried this before, and it still resonates with my "least powerful abstraction" sense, but the concerns are real and the benefits are scarce.  I'm a mild :-1: right now, but still persuadable either way.  Since things have shifted in fs2, I'd recommend targeting cats-effect-3 branch if this proceeds.

/cc @mpilquist, because it's good to hear from all the fs2 folks

PR #4027, opened 2020-12-17
Enable Scaladoc Linking Warnings

isomarcte:
They are not fatal and we've since cleaned up _most_ of them. It is our hope that by enabling them we can catch simple typos, etc.

For context, see: https://github.com/http4s/http4s/pull/4020#issuecomment-747544065


isomarcte:
Oh that's odd. They weren't fatal when I ran them locally...I'll look into this.

rossabaker:
sbt-spiewak makes them fatal on CI, but not locally.  This is convenient when exploring and having unused imports, but it's frustrating when it comes PR time.  I'm not sure I like the behavior, and we can turn it off, but that's why, and it's a recent change.

isomarcte:
@rossabaker ah, that makes sense. I was grepping for a `-Werror` or `-Xfatal-warnings` I'd thought I'd missed.

I would be fine disabling that in CI. For just about all scopes other than docs I _do_ want `-Werror`, but docs is the exception.

rossabaker:
My attention has been split and there are a lot of exemptions floating around in a lot of plugins and I guess I need a mental reset.

So as of be0890c, we have several linking warnings.  These are fatal in CI.  Can we clean these up and leave the flags as they are?  Sometimes I try to do too many things at once and rely on CI to alert me about broken docs, so I have a mild preference for leaving them on so they're seen as they happen.  But they can be annoying -- last time I tried, we couldn't run mdoc cleanly on 2.13, which is why we just generate them from 2.12.

I'm pretty open to whatever solution you think is best here.

isomarcte:
@rossabaker the remaining warnings are all related to overloaded symbols. I would love to clean them up, but for the life of me I can't seem to get the correct incantation. You are supposed to be able to provide _part_ of the signature to distinguish between overloaded symbols, but nothing I tried seemed to work.

I'll attempt to another crack at it before we change the warning flags.

rossabaker:
An unsatisfying solution, if all else fails, is also to just unlink them.

isomarcte:
@rossabaker 

> An unsatisfying solution, if all else fails, is also to just unlink them.

I'd be hesitant to do that in this case. When there is an overloaded symbol in a Scaladoc link, Scaladoc _will_ still generate a link to one of the members, it just also emits a warning. So, from a user's perspective, they are still getting the nice linking. It's just a pain in the build.

rossabaker:
Ah, that makes sense. Could we turn off fatal warnings for doc and unidoc, but leave them on for compiling and mdoc?

armanbilge:
I'd like to land this one. It's also currently affected by https://github.com/typelevel/sbt-typelevel/issues/226 I believe.

armanbilge (approved):

PR #4032, opened 2020-12-17
Allow For Providing HttpConfiguration With Jetty

isomarcte:
This commit makes two meaningful changes,

* Allow for the user to provide their own `HttpConfiguration` value. This is important as there are many settings which are not modeled in the `JettyBuilder` that one would general expect to have access to.
* Enforces that the `HttpConfiguration` is used whether or not http2 is enabled. Formerly the `HttpConfiguration` was only used if there was not an `SslConnectionFactory` and http2 support was enabled.

### DIFF
--- a/jetty/src/main/scala/org/http4s/server/jetty/JettyBuilder.scala
+++ b/jetty/src/main/scala/org/http4s/server/jetty/JettyBuilder.scala
@@ -56,13 +56,43 @@ sealed class JettyBuilder[F[_]] private (
     mounts: Vector[Mount[F]],
     private val serviceErrorHandler: ServiceErrorHandler[F],
     supportHttp2: Boolean,
-    banner: immutable.Seq[String]
+    banner: immutable.Seq[String],
+    jettyHttpConfiguration: HttpConfiguration
 )(implicit protected val F: ConcurrentEffect[F])
     extends ServletContainer[F]
     with ServerBuilder[F] {
   type Self = JettyBuilder[F]
 
   private[this] val logger = getLogger
+
+  @deprecated(message = "Retained for binary compatibility", since = "0.21.15")
+  private[JettyBuilder] def this(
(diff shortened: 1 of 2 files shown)


jetty/src/main/scala/org/http4s/server/jetty/JettyBuilder.scala
```diff
@@ -400,6 +444,10 @@ object JettyBuilder {
       case SSLClientAuthMode.Required =>
         sslContextFactory.setNeedClientAuth(true)
     }
+
+  /** The default [[org.eclipse.jetty.server.HttpConfiguration]] to use with jetty. */
+  val defaultJettyHttpConfiguration: HttpConfiguration =
```
rossabaker, line 449: Does this need to be public?  I guess we're already committed to supporting an`HttpConfiguration` on the API, so the additional liability isn't much.  I could go either way here.

rossabaker (approved):
I like it.

I've thought it would be nice to provide an alternate builder that just took a Jetty instance and wrapped it in the typical server trappings.  This would allow the more exotic parts of the API our builder doesn't cover.  But I think this is good for what we do today.

jetty/src/main/scala/org/http4s/server/jetty/JettyBuilder.scala
isomarcte: @rossabaker I made it public so that someone who wants to make a minor adjustment to the configuration can do so and then just use the rest of the "default" http4s configuration. Of course, since our default configuration right now is just the empty constructor, that doesn't really make a lot of sense. 

I am going to make another PR which attempts to align the various configuration parameters which can be meaningfully shared between server backends, e.g. request header size etc. I'll make this private until that PR.

isomarcte: @rossabaker updated

PR #4037, opened 2020-12-18
Normalize Some Of The HTTP Server Parameters Between Backends

isomarcte:
This commit attempts to normalize some of the common server configuration parameters between backends.

When switching between different backends, one would assume that the operational semantics would be as close as possible, assuming there was no direct configuration change in user code. Prior to this commit there was a substantial delta between these parameters in Blaze/Jetty/Ember.

It is likely that _more_ parameters can be shared in the future, but some configuration options are not exposed for any give backend, and some things which are configurable on one backend may not even be expressible on another.

Thus, this commit attempts to be relatively conservative, synchronizing the more obvious parameters only.

When two backends currently differed on the configuration value for a given parameter, the more _permissive_ value was chosen. Here is a summary of the configuration changes with respect to a given backend.

* Ember
    * Default host is now derived from `org.http4s.server.defaults.Host`, though it should be the same.
    * Default Max header size, 10240B -> 40960B, which was the value used by Blaze
* Jetty
    * Default Max header size, 8192B -> 40960B, which was the value used by Blaze
    * See: https://github.com/eclipse/jetty.project/blob/jetty-9.4.x/jetty-server/src/main/java/org/eclipse/jetty/server/HttpConfiguration.java#L60

### DIFF
--- a/blaze-server/src/main/scala/org/http4s/server/blaze/BlazeServerBuilder.scala
+++ b/blaze-server/src/main/scala/org/http4s/server/blaze/BlazeServerBuilder.scala
@@ -399,26 +399,7 @@ class BlazeServerBuilder[F[_]](
 object BlazeServerBuilder {
   @deprecated("Use BlazeServerBuilder.apply with explicit executionContext instead", "0.20.22")
   def apply[F[_]](implicit F: ConcurrentEffect[F], timer: Timer[F]): BlazeServerBuilder[F] =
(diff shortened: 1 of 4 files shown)


hamnis (approved):

rossabaker (approved):
I like it.  It's possible different defaults are appropriate for different backends, but I'd rather default to making them similar, and this is a step in the right direction.

PR #4040, opened 2020-12-19
Added a copy of vault to proceed with CE3 upgrade

domaspoliakas:
As per the title

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -95,7 +95,8 @@ lazy val core = libraryProject("core")
       scalaReflect(scalaVersion.value) % Provided,
       scodecBits,
       slf4jApi, // residual dependency from macros
-      vault,
+      unique,
+      /* vault, */
     ),
     unusedCompileDependenciesFilter -= moduleFilter("org.scala-lang", "scala-reflect"),
     mimaBinaryIssueFilters ++= Seq(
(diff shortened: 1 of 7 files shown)


domaspoliakas:
Running 2.12 tests with adopt 1.8 worked locally, so I'm not sure what's going on with ci there

hamnis:
Server is not compiling, has that been converted to CE3 yet?

domaspoliakas:
The PR for that seems to be open still. It's commented out in the module list in `build.sbt`, should it not get ignored when running CI?

domaspoliakas:
It looks like someone re-ran CI and now it's good 🎊 

The DSL compiles and tests successfully in this as well without any changes to the source; should I uncomment it while I'm here, or leave it for a different PR?

hamnis:
Maybe that should be its own PR.

hamnis (approved):

hamnis:
perhaps a different take on this might be: https://eed3si9n.com/hot-source-dependencies-using-sbt-sriracha

domaspoliakas:
I think I have one reservation to pursuing that strategy (provided I understood it correctly, which I may not have). Ultimately whether we copy the source or setup a source dependency, someone will have to go back and clean it up to just use a binary dependency again once vault has a version with ce3 published. And the benefit of using the source dependency is that the cleanup work would be trivial, whereas the copy method will involve cleaning up the classes too. However, because there are so few files involved in the copying, it might take overall less time to do the copying and cleaning up the files, than fiddling with source dependencies to get them right.

I'm no library maintainer however, so if you think that the source dependency path is the right way to go I'd be happy to get that working.

rossabaker:
I think source dependencies are the right approach in theory, but I haven't had particularly good luck with them in practice, and this one is small enough for this crude approach to work.  Let's roll with this.

PR #4045, opened 2020-12-21
Port some tests using withResource

cquiroz:
This is a smaller unit port of server resources tests. Instead of `withResource` it uses munit fixtures


rossabaker (approved):
Nice. That was an awful hack in Specs2.

Issue #331, opened 2015-07-02
Upgrade to Specs2-3.x


rossabaker:
It's easier to stay current than to stagnate.  That's the upgrade we missed last time.

jedesah:
As far as sbt in concerned, http4s does not depend on specs2 per say, it is a transitive dependency of scalaz-specs2, is that correct?

I might argue that it would make sense to declare a direct dependency, but in the mean time, I gather this task is to update scalaz-specs2 to 0.4?

rossabaker:
Correct.  Those two upgrades need to happen as a pair for reasons of binary compatibility.  A direct Specs2 could conceivably get us on the latest patch revision, but risks getting out ahead of scalaz-specs2.  I think it should remain transitive until we need otherwise.

PR #4048, opened 2020-12-22
Handle CompletableFuture Correctly

isomarcte:
This commit adds a new function `fromCompletableFutureShift` which brackets the usage of a `CompletableFuture` so that if the outer effect is canceled or terminates in error, then the `CompletableFuture` is canceled or terminated as well.

This is _similar_ to the deprecated `fromCompletableFuture`, but differs in that it handles exceptional cases, not just cancellation, and properly handles the thread shifting, as `fromCompletionStage` does.

`fromCompletionStage` is used in `http4s-jdk-http-client` and under the current usage causes a resource leak if the outer effect terminates before the callback has been invoked.

### DIFF
--- a/core/src/main/scala/org/http4s/internal/package.scala
+++ b/core/src/main/scala/org/http4s/internal/package.scala
@@ -189,10 +189,8 @@ package object internal {
     }
 
   private[http4s] def fromCompletableFutureShift[F[_], A](
-      fcs: F[CompletableFuture[A]])(implicit
-      F: Concurrent[F],
-      CS: ContextShift[F]): F[A] =
-    F.bracketCase(fcs){ cs =>
+      fcs: F[CompletableFuture[A]])(implicit F: Concurrent[F], CS: ContextShift[F]): F[A] =
+    F.bracketCase(fcs) { cs =>
       F.async[A] { cb =>
         cs.handle[Unit] { (result, err) =>
           err match {
@@ -210,8 +208,7 @@ package object internal {
           F.delay(cs.completeExceptionally(e))
         case ExitCase.Canceled =>
           F.delay(cs.cancel(true))
-      }).void.guarantee(CS.shift)
-    )
+      }).void.guarantee(CS.shift))
 
   private[http4s] def unsafeToCompletionStage[F[_], A](
       fa: F[A]
--- a/core/src/main/scala/org/http4s/internal/package.scala
+++ b/core/src/main/scala/org/http4s/internal/package.scala
@@ -38,6 +38,7 @@ import scala.util.control.NoStackTrace
 import scala.util.{Failure, Success}
 import java.nio.charset.MalformedInputException
 import java.nio.charset.UnmappableCharacterException
+import cats.effect.ExitCase
 
 package object internal {
   // Like fs2.async.unsafeRunAsync before 1.0.  Convenient for when we
@@ -167,6 +168,7 @@ package object internal {
       }
     }
 
+  @deprecated(message = "Use fromCompletableFutureShift", since = "0.21.15")
   private[http4s] def fromCompletionStage[F[_], CF[x] <: CompletionStage[x], A](
       fcs: F[CF[A]])(implicit
       // Concurrent is intentional, see https://github.com/http4s/http4s/pull/3255#discussion_r395719880
@@ -185,6 +187,31 @@ package object internal {
         ()
       }.guarantee(CS.shift)
     }
+
+  private[http4s] def fromCompletableFutureShift[F[_], A](
+      fcs: F[CompletableFuture[A]])(implicit
+      F: Concurrent[F],
+      CS: ContextShift[F]): F[A] =
+    F.bracketCase(fcs){ cs =>
+      F.async[A] { cb =>
+        cs.handle[Unit] { (result, err) =>
+          err match {
+            case null => cb(Right(result))
+            case _: CancellationException => ()
+            case ex: CompletionException if ex.getCause ne null => cb(Left(ex.getCause))
+            case ex => cb(Left(ex))
+          }
+        }; ();
+      }
+    }((cs, ec) =>
+      (ec match {
+        case ExitCase.Completed => F.unit
+        case ExitCase.Error(e) =>
+          F.delay(cs.completeExceptionally(e))
+        case ExitCase.Canceled =>
+          F.delay(cs.cancel(true))
+      }).void.guarantee(CS.shift)
+    )
 
   private[http4s] def unsafeToCompletionStage[F[_], A](
       fa: F[A]


isomarcte:
This is a bit hard to test without involving the JDK 11 Http Client (which we can't do in http4s proper), but here is some toy code which I used to confirm the bug.

```scala
package org.http4s.internal

import cats._
import cats.syntax.all._
import cats.effect._
import cats.effect.syntax.all._
import java.util.concurrent.CompletableFuture
import java.net.http.HttpRequest
import java.net.http.HttpResponse
import java.net.http.HttpClient
import java.net.URI
import java.util.concurrent.TimeUnit
import scala.concurrent.duration.FiniteDuration

object Test {
  final class CF extends CompletableFuture[Unit]
  val client: HttpClient = HttpClient.newHttpClient
  implicit val cs: ContextShift[IO] = IO.contextShift(ExecutionContext.global)
  implicit val timer: Timer[IO] = IO.timer(ExecutionContext.global)


  def test[F[_]](implicit F: Concurrent[F], CS: ContextShift[F], T: Timer[F]): F[Unit] =
    fromCompletionStage(
      F.delay(
        client.sendAsync(
          HttpRequest.newBuilder(new URI("http://127.0.0.1:9090")).build
            , HttpResponse.BodyHandlers.ofString
        )
      ).guarantee(
        F.delay(println("complete"))
      )
    ).flatMap(value => F.delay(println(value))).timeout(FiniteDuration(5L, TimeUnit.SECONDS))

  val ioTest: IO[Unit] =
    test[IO]
}
```

I also added a `println` in the result of `fromCompletionStage` and `fromCompletableFutureShift`. Then I ran `nc` locally to hold the http connection open, forcing the `timeout` to trigger.

In the case of `fromCompletionStage`, if I responded to the HTTP request via `nc` _after_ the `timeout` had occurred, I still saw the printlns execute for the JDK Http client handling. In the case of `fromCompletableFutureShift`, the body handler correctly did not run.

If this all sounds a bit confusing and hard to follow, I'm sorry, I'm quite sleepy. I'm going to get a proper test up in `http4s-jdk-http-client` tomorrow to unambiguously show the bug and the fix. The short of it is that `fromCompletionStage` will eventually OOM a JVM if the outer effects fail before the `HttpResponse` is processed, and in any other case where `fromCompletionStage` is used.

diesalbla:
@isomarcte Have to look at this in depth, but, _prima facie_, is this something that covers a specific need of `http4s` and the JDK client, or is this handling of `CompletableFuture` something to bring into `cats-effect` as well?

isomarcte:
@diesalbla see this discussion: https://github.com/typelevel/cats-effect/issues/160.

TL;DR is that uses case like this are considered a bit beyond `cats-effect`'s design scope.

rossabaker:
It was decided in CE2 to put this functionality into monix-catnap, which is a nice little library, but I've been reluctant to add a dependency on Monix.  I like that project a lot, but it's a sophisticated repo that takes a while to release for new Scala and CE3 versions.

I wouldn't be opposed to such functionality in CE3, but we still need something down here on series/0.21.  This looks reasonable, but also interested in what @amesgen thinks.

amesgen:
In the #3255 (where I added this function), I wrote:

> Rename `fromCompletableFuture` to `fromCompletionStage`, as we don't want to `complete` or `get` the given future. I removed the call to `cancel` (therefore, we only need `Async[F]` now), as it can be seen as a special case of `completeExceptionally` and therefore does not really seem cancel anything (but I don't really know much about Java futures, so please correct me if I am wrong).

I got this from [here](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/CompletableFuture.html):

> Since (unlike `FutureTask`) this class has no direct control over the computation that causes it to be completed, cancellation is treated as just another form of exceptional completion. Method `cancel` has the same effect as `completeExceptionally(new CancellationException())`. Method `isCompletedExceptionally()` can be used to determine if a `CompletableFuture` completed in any exceptional fashion. 

This [stackoverflow answer](https://stackoverflow.com/a/44397321) seems to also suggest this:

> When you call `CompletableFuture#cancel`, you only stop the downstream part of the chain. Upstream part, i. e. something that will eventually call `complete(...)` or `completeExceptionally(...)`, doesn't get any signal that the result is no more needed.

As I understand it, "real" cancellation in the cats-effect sense does not really work for an arbitrary `CompletableFuture` (in particular, the argument to `cancel` is ignored).

Therefore, I am confused how your change can actually make any difference. Can you specify which printlns you mean? But you already wrote that you want to explain this in more detail in an issue/PR to http4s-jdk-http-client. Debugging these FFI-y kinds of things is pretty exhausting at times, take your time!

---

BTW you can use https://httpstat.us to get an URL which reponds after a specific amount of time, e.g. `httpstat.us/200?sleep=5000` for 5 seconds, if that simplifies things.

isomarcte:
@rossabaker @amesgen I've opened a PR to http4s-jdk-http-client (https://github.com/http4s/http4s-jdk-http-client/pull/331). It includes a test which can be used to show the error.

In `CompletableFutureTerminationTest` just change `fromCompletableFutureShift` to `org.http4s.internal.fromCompletionStage` and run the test. (https://github.com/http4s/http4s-jdk-http-client/pull/331/files#diff-a5a8777c5f8bea51fded21973d608a63c4d9da4716e7039ffc53c2dab1a6ac58R99)

I've include a bunch of comments describing everything.

isomarcte:
@amesgen 

Caveat: This is my working understanding, and it is congruent with the tests I've run. There is a lot of moving parts here, CompletableFuture/Reactive Streams/fs2/cats-effect. It very possible there are gaps in my understanding.

> This stackoverflow answer seems to also suggest this:

>     When you call CompletableFuture#cancel, you only stop the downstream part of the chain. Upstream part, i. e. something that will eventually call complete(...) or completeExceptionally(...), doesn't get any signal that the result is no more needed.

> As I understand it, "real" cancellation in the cats-effect sense does not really work for an arbitrary CompletableFuture (in particular, the argument to cancel is ignored).

Right, so you may not actually terminate _all_ computation if you invoke `CompletableFuture.cancel`. What you do get though is that the release of resources. 

That is, say, you invoke `cancel` at T0. At T1, some other thread attempts to invoke `CompletableFuture.complete`. This will now fail, because a `CompletableFuture` can only ever be completed exactly once.

Now let's examine the other case with the current implementation of `fromCompletableStage`. Let's assume the effect wrapping the `CompletableFuture` is terminated via cancellation or error, before any data is received. With the current implementation of `fromCompletableStage`, nothing will be invoked on the `CompletableFuture` API. Then at T1 the other thread invokes `CompletableFuture.complete` and no error is generated, because nothing has completed the `CompletableFuture` in anyway. In the case of the JDK client, this will create a `fs2.Stream` wrapping reactive streams API. This is where the resource leak occurs. From the JavaDocs (https://docs.oracle.com/en/java/javase/14/docs/api/java.net.http/java/net/http/HttpResponse.BodySubscriber.html)

> API Note:
>     To ensure that all resources associated with the corresponding HTTP exchange are properly released, an implementation of BodySubscriber should ensure to request more data until one of onComplete or onError are signalled, or cancel its subscription if unable or unwilling to do so. Calling cancel before exhausting the response body data may cause the underlying HTTP connection to be closed and prevent it from being reused for subsequent operations.

One of these three events _must_ occur on the `BodySubscriber` in order to release all the resources, the `BodySubscriber` must poll until `onComplete` or `onError` is signaled, _or_ something must externally call `cancel` on the subscription. In the case with the current `fromCompletionStage`, none of these methods will ever be invoked. `onComplete` and `onError` won't be invoked because no on is reading the response and `cancel` won't be invoked because there is nothing to invoke it, the async event has left scope.

I don't think this particular leak will happen in the _general_ case with `CompletableFuture`. It looks like a specific interplay of the JDK `HttpClient` API and how it is using `CompletableFuture`. That being said, I don't think there is ever any harm in invoking `cancel` or `completeExceptionally`.

isomarcte:
@rossabaker @amesgen _if_ we merge https://github.com/http4s/http4s-jdk-http-client/pull/331 I'm not sure we still need to merge this PR. I noticed that `fromCompletionStage` is actually never invoked in http4s proper, and it is package private. 

Are there other http4s projects, beyond the jdk client, which rely on this function?

amesgen:
Wow, thanks for the detailed explanation! I will look into it in the next few days. 

I am :+1: on closing this PR and moving the discussion to #331.

amesgen:
Closing as http4s/http4s-jdk-http-client#331 is now merged, and cats-effect 3 will have a method to convert from a `CompletableFuture` either way.

PR #4051, opened 2020-12-22
ResponseLogger constructor overload receiving the response

### LINKED ISSUE #4019: ResponseLogger constructor overload receiving the response
It would be helpful if `ResponseLogger` had a "constructor" where the `logAction` would have the type `Option[Response[F] => F[Unit]]` this would allow for example: successful messages to be printed in green, client errors in yellow, and server errors in red.

A similar idea could be applied for the `RequestLogger` which would allow for example printing the request uri in bold. Or read methods (GET, HEAD, etc) in a color and write methods (PUT, POST, DELETE, etc) in another color.

We could even make a default implementation something like `ResponseLogger.colored` which would make some choices right of the gate.

I can make PR.

Lasering:
Refactor Logger implementation.
Add customized loggers.
Add colored loggers.
Closes #4019

### DIFF
--- a/client/src/main/scala/org/http4s/client/middleware/Logger.scala
+++ b/client/src/main/scala/org/http4s/client/middleware/Logger.scala
@@ -56,4 +56,18 @@ object Logger {
       log: String => F[Unit])(implicit F: Sync[F]): F[Unit] =
     org.http4s.internal.Logger
       .logMessage[F, A](message)(logHeaders, logBody, redactHeadersWhen)(log)
+
+  def colored[F[_]: Concurrent](
+      logHeaders: Boolean,
+      logBody: Boolean,
+      redactHeadersWhen: CIString => Boolean = Headers.SensitiveHeaders.contains,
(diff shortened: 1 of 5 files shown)


Lasering:
In package `org.http4s.client.middleware`:
 - `Logger#logMessage` is just an alias for `org.http4s.internal.Logger#logMessage`
 - `RequestLogger#apply` invokes `Logger#logMessage`
 - but `RequestLogger#logBodyText` invokes `org.http4s.internal.Logger#logMessageWithBodyText`.
  
 Seems to me `Logger#logMessage` is only causing entropy.

rossabaker (approved):
I wonder if we could do something a bit more builder-like and get rid of some of these repeated parameters across the various constructors (e.g., `redactHeadersWhen`).  The overall API feels a lot larger to me than it ought to be.

That said, I think the new functionality is neat, and the refactoring smart. :+1: to this.

rossabaker:
I just noticed this went to main while doing release notes.  I think it was binary compatible, and could be backported if we're interested.

PR #4052, opened 2020-12-22
Winter solstice dependency bump

rossabaker:
The days are getting longer again!  Well, unless you live near @cquiroz.


cquiroz (approved):
LGTM

PR #4054, opened 2020-12-23
port scala xml to cats effect 3

### LINKED ISSUE #3841: Port scala-xml to cats-effect-3

m-sp:
fixes #3841

### DIFF
--- a/scala-xml/src/main/scala/scalaxml/ElemInstances.scala
+++ b/scala-xml/src/main/scala/scalaxml/ElemInstances.scala
@@ -17,12 +17,10 @@
 package org.http4s
 package scalaxml
 
-import cats.effect.Sync
-import cats.syntax.all._
+import cats.effect.Concurrent
 import java.io.StringReader
 import javax.xml.parsers.SAXParserFactory
 
-import cats.data.EitherT
 import org.http4s.headers.`Content-Type`
 
 import scala.util.control.NonFatal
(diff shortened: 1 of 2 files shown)


hamnis (approved):

scala-xml/src/main/scala/scalaxml/ElemInstances.scala
```diff
@@ -44,7 +44,7 @@ trait ElemInstances {
     *
     * @return an XML element
     */
-  implicit def xml[F[_]](implicit F: Sync[F]): EntityDecoder[F, Elem] = {
+  implicit def xml[F[_]](implicit F: Async[F]): EntityDecoder[F, Elem] = {
```
rossabaker, line 47: Does anything prevent this being `Concurrent`? That's the new constraint for most of the decoders.  Is it the `F.delay` on line 56?

rossabaker (commented):
Thanks! One question, out of my ignorance on cats-effect-3.

scala-xml/src/main/scala/scalaxml/ElemInstances.scala
m-sp: Exactly, I tried concurrent first but delay isn't available on concurrent

rossabaker: I don't see much of a side effect here.  It consumes the `source`, but it's our source, from a strict byte array.  We don't need to demand a tighter constraint for this local mutation.  I think we could replace that `F.delay` with a `try / catch`, sort out the errors, and lose the `Async` constraint.

m-sp: thanks for the review, makes sense. looks also a bit simpler

m-sp:
seems like the test failure is unrelated?

rossabaker (approved):
Yes, rerunning the tests. :+1: on green.

PR #4059, opened 2020-12-25
Split BlazeClientSuite

cquiroz:
I ported `BlazeClientSpec` to munit and at the same time tried to understand why some of the tests are flaky.
After quite many runs I noted that failures are manifested in 2.12 (so far).
I'm proposing to split the test into a common test suite and a few more tests running on 2.13

Unfortunately I don't have a solid theory to back it up just experimental results showing that it doesn't seem to fail on 2.13

Please let me know what do you think of this approach.

Note the tests moved to 2.13 are the ones I noted failing sometimes and 2.12 and were exactly (noted after the fact) the ones marked as `skipOnCI`


rossabaker (approved):
Okay, that's a fascinating observation.  I haven't noticed the same, but I believe you, and we can run it like this for a while and see how it goes.

rossabaker:
I'm starting to see a trend toward 2.12 failures now that you've mentioned this, but here's a [2.13 failure]( [https://github.com/http4s/http4s/pull/4068/checks?check_run_id=1611538587).  I don't know if your observation is limited just to blaze-client.

cquiroz:
I only focused on BlazeClietnSuite, there are other test suite that are flaky but I didn't try to track them

PR #4060, opened 2020-12-26
BlazeClientSuite in main

cquiroz:
This brings the client test change to main.
It carries dependency updates which I'm not sure should be included but everything compiles :+1:


rossabaker:
Yes, all of those dependency updates are good.

PR #4064, opened 2020-12-26
Dotty can't find org.http4s.util._ imports in Uri

rossabaker:
I don't understand this one, but we get 8 errors closer to Dotty support with it.


hamnis (approved):
May be some name overloads here that dotty finds, but scala 2 doesnt

PR #4067, opened 2020-12-26
scala-collection-compat

rossabaker:
# Why

* Stops publishing an annotation in the scala package, which is a naughty thing for org.http4s to do.
* Makes code look more like 2.13 and Dotty as 2.12 recedes from importance
* A bunch of projects are reimplementing this library ad hoc, and we're contributing to the problem.
* Don't have to propagate our version into Dotty.

# Why not

* We get unused import warnings on 2.13 with `scala.collection.compat._`.  It looks _too much_ like 2.13.
* Dependencies are bad.  Even well-maintained ones like this.

# Sure, but for 1.0

* Dropping the Scala dependency is a MiMa violation, even though we'd be supplying it transitively.  
* One could argue new core dependencies should wait for 1.0 if at all possible.


rossabaker:
I see @Daenyth and @SethTisue once discussed a similar unused warnings issue from this library, and am curious whether they found a satisfactory resolution.

I want to keep fatal warnings, and I want the annotation in `scala` gone, and I can't figure out how to do both.

client/src/test/scala/org/http4s/client/testroutes/GetRoutes.scala
```diff
@@ -33,19 +30,18 @@ object GetRoutes {
...
-          .pure[IO],
-        DelayedPath ->
-          timer.sleep(1.second) *>
-          Response[IO](Ok).withEntity("delayed path").pure[IO],
-        NoContentPath -> Response[IO](NoContent).pure[IO],
-        NotFoundPath -> Response[IO](NotFound).withEntity("not found").pure[IO],
-        EmptyNotFoundPath -> Response[IO](NotFound).pure[IO],
-        InternalServerErrorPath -> Response[IO](InternalServerError).pure[IO]
-      ))(_.unsafeRunSync())
+  def getPaths: Map[String, Response[IO]] =
```
rossabaker, line 33: If this PR fails, we should at least salvage this.

core/src/main/scala/org/http4s/internal/package.scala
```diff
@@ -355,4 +358,15 @@ package object internal {
       tail: Eval[Int]*
   ): Int =
     reduceComparisons_(NonEmptyChain(Eval.now(head), tail: _*))
+
+  def pairsToMultiParams[K, V](map: collection.Seq[(K, Option[V])]): IMap[K, ISeq[V]] =
```
rossabaker, line 362: Getting this out of the condemned traits.  I guess it should go right on `Query` instead.

rossabaker:
We scratched these itches with #4070 and #4072.  Closing due to lack of interest.

PR #4068, opened 2020-12-26
Update README (2019 -> 2021)

gokyo:


rossabaker (approved):
Someone else who's glad to leave 2020 behind!

The guidelines on how to date copyrights are frustratingly murky.  We ended up stripping the end dates from the source files.  Some argue that it's fraud if you haven't updated the file!  But the project as a whole has been continuously updated, so I think this is good.

PR #4070, opened 2020-12-26
Filter nowarn annotation from jar

rossabaker:
Alternative to #4067.  We just simply filter the `@nowarn` annotation out of the jar.

Pros:
- No new core dependencies
- Preserves fatal warnings without further complexity

Cons:
- Compat trait will need to be copied to a scala-3 directory on the Dotty branch.
- Doesn't steer our source toward modernity.
- MiMa doesn't catch that the annotations are missing. But we'll note it and move on, and it's 2.12 only.


rossabaker:
I'm leaning toward this being the better approach, and conservatively putting it on main instead of 0.21.

PR #4072, opened 2020-12-26
Collection compat 3

rossabaker:
If we don't go with #4067, this is necessary, and it fixes several Dotty errors now.


rossabaker:
Okay, I was trying to figure out why this had more than one commit, and accidentally merged it.  Please complain here if you didn't like it.

PR #4073, opened 2020-12-27
Restore some binary compatibility

rossabaker:
* RequestCookie and ResponseCookie companions need to extend a runtime abstract function
* `TransferCoding` lost its constructors.  (The `parseList` is inconsistent.  I think we might remove that in 1.0.)
* `Proxy-Authorization` is nice work, but needs to wait for 1.0.  We'll restore it on merge.
* A bunch of other package private things are declared.


isomarcte (approved):

rossabaker:
I brought back a `Content-Range.apply` overload when resolving the merge conflict.

PR #4076, opened 2020-12-27
Move the macros to a scala-2 directory

rossabaker:
Moves all the macro implementations into a scala-2 directory.  Preserves binary compatibility of public API, keeping the dotty-on-0.21 dream alive.

The current `LiteralMacros` are easy.

The macros deprecated in 0.20 are hard.  We don't care about them, but we can't delete them without breaking binary compatibility, and the `Macros` classes on the companion objects put `scala.reflect.macros.whitebox` in the public API.  Ouch!

Work around it thusly:
* Move the macro implementations to `DeprecatedMacros`.  In `scala-3`, these are empty classes constructed with `Nothing`. 
* Move the deprecated macro calls to a `Deprecated` trait for each affected companion.  They need not exist in `scala-3`.
* Put a type alias to `whitebox` on the `Deprecated` trait in `scala-2`.  In `scala-3`, it's `Nothing`.
* Rip it out and never speak of it again in main.


rossabaker:
Maybe the `QValue` one isn't detestable: it works on numbers instead of strings.  But you can't really do arithmetic with it as a literal anyway, except for stupid `q(0.5 + 0.1)` tricks.  So I think I'd still like to deprecate it.

rossabaker:
The way we deprecated these back in 0.20 didn't actually deprecate the usage, and there are a ton of them in the tests, particularly `Uri.uri`.  There are two ways to do all three of these literals, and we ought to unify, but: on what?

rossabaker:
See #4078

rossabaker:
This can be a lot simpler if we give up on the idea of Dotty on 0.21.

hamnis:
handled by #4145

PR #4081, opened 2020-12-27
migrate dropwizard and prometheus metrics

m-sp:
loosened constraints on client and server metrics to make the fake clock work without changes


rossabaker (commented):
This is one I could use help from a CE3 expert on.  I don't have the new typeclass relationships mastered yet.

RaasAhsan:
I think `Temporal` -> `Clock + Concurrent` looks good. I forgot to go back and replace some of those since `Clock` was still extending `Monad` at the time (it would be neat if there was a static analyzer that could mechanically check if weaker constraints fit)

djspiewak:
Why not just `Temporal` though? The only capability you're masking out is `sleep`.

RaasAhsan:
The usual answer has been testing (it's a lot easier to mock out Clock
since it's incoherent),
but otherwise that's a pretty philosophical question that touches up
against the principle of least power :) I don't care for it personally but
some people swear by it

m-sp:
> The usual answer has been testing (it's a lot easier to mock out Clock since it's incoherent)...

exactly it is testing, there is a `FakeClock` in the tests which incresease whenever `realTime` or `monotonic` is called. I didn't really see a way to do this with the testkit yet

rossabaker (approved):
Okay, that makes sense to me.

djspiewak:
> exactly it is testing, there is a FakeClock in the tests which incresease whenever realTime or monotonic is called. I didn't really see a way to do this with the testkit yet

`TimeT` is *supposed* to meet this type of use-case, but doesn't quite yet. It would be worth taking a look at it to see how far off it is. I haven't had time to work on it as much as I would like :-(

m-sp:
Thanks for the pointer I'll try to take a look and see if it can work here

djspiewak:
> Thanks for the pointer I'll try to take a look and see if it can work here

If it doesn't work, it's almost certainly my fault and you should let me know!

rossabaker:
This gets us operational, so I'm going to go ahead and merge this, and if we can make it better still, we can do it on a separate PR.

PR #4095, opened 2020-12-28
cats-parse port of URI

rossabaker:


core/src/main/scala/org/http4s/Query.scala
```diff
@@ -193,9 +191,10 @@ object Query {
     * If parsing fails, the empty [[Query]] is returned
```
rossabaker, line 191: I forget why this is, and I don't like it, but I think we're stuck with it.

core/src/main/scala/org/http4s/headers/Forwarded.scala
```diff
@@ -193,7 +193,7 @@ object Forwarded
       ParseFailure("missing host", s"no host defined in the URI '$uri'")
   }
 
-  override def parse(s: String): ParseResult[Forwarded] = parser.HttpHeaderParser.FORWARDED(s)
+  override def parse(s: String): ParseResult[Forwarded] = ???
```
rossabaker, line 196: I deleted this somewhere along the way and reverted. @fredshonhorio is looking at the Forwarded header.

core/src/main/scala/org/http4s/headers/Origin.scala
```diff
@@ -76,8 +75,8 @@ object Origin extends HeaderKey.Internal[Origin] with HeaderKey.Singleton {
       .string
       .map(Uri.Scheme.unsafeFromString)
     val stringHost = until1(char(':').orElse(`end`)).map(RegName.apply)
-    val bracketedIpv6 = char('[') *> Rfc3986.ipv6 <* char(']')
-    val host = List(bracketedIpv6, Rfc3986.ipv4, stringHost).reduceLeft(_ orElse1 _)
+    val bracketedIpv6 = char('[') *> Uri.Ipv6Address.parser <* char(']')
+    val host = List(bracketedIpv6, Uri.Ipv4Address.parser, stringHost).reduceLeft(_ orElse1 _)
```
rossabaker, line 79: This coupling suggests to me that our IP addresses shouldn't be buried in the URI object.

core/src/main/scala/org/http4s/internal/CharPredicate.scala
```diff
@@ -0,0 +1,342 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.http4s.internal
+
+import scala.annotation.tailrec
+import scala.collection.immutable.NumericRange
+
+sealed abstract class CharPredicate extends (Char => Boolean) {
```
rossabaker, line 22: We have decoders based on `CharPredicate`, and they're going to be faster than going char-by-char with cats.parse when we're decoding raw strings.  We could probably whittle this down, but the fastest way to get started was to just take it from parboiled2.

core/src/main/scala/org/http4s/internal/parsing/Rfc2234.scala
```diff
@@ -0,0 +1,35 @@
...
+package org.http4s.internal.parsing
+
+import cats.parse.Parser1
+import cats.parse.Parser.{charIn}
+
+/** Common rules defined in RFC2234.
+  *
+  * @see [[https://tools.ietf.org/html/rfc2234] RFC2234, Augmented BNF for Syntax Specifications: ABNF]
+  */
+private[http4s] object Rfc2234 {
```
rossabaker, line 26: Maybe this is a bit much: Rfc3986 points to Rfc2234 for things that are already defined a couple other places.  Alpha is alpha and digit is digit.  But I kind of like it for traceability.

core/src/main/scala/org/http4s/internal/parsing/Rfc3986.scala
```diff
@@ -14,21 +14,45 @@
...
 private[http4s] object Rfc3986 {
+  def alpha: Parser1[Char] = Rfc2234.alpha
 
-  private def ipv4Internal[A](f: (Byte, Byte, Byte, Byte) => A): Parser1[A] = {
+  def digit: Parser1[Char] = Rfc2234.digit
+
+  /* The spec references RFC2234, which is 0-9A-F, but it also
+   * explicitly permits lowercase. */
+  val hexdig: Parser1[Char] =
+    digit.orElse1(charIn("ABCDEF"))
```
rossabaker, line 36: And then I went and failed to add the lowercase. :facepalm:

core/src/main/scala/org/http4s/internal/parsing/Rfc3986.scala
```diff
@@ -14,21 +14,45 @@
...
+
+  /* unreserved    = ALPHA / DIGIT / "-" / "." / "_" / "~" */
+  val unreserved: Parser1[Char] =
+    alpha.orElse1(digit).orElse1(charIn("-._~"))
+
+  /* pct-encoded   = "%" HEXDIG HEXDIG */
+  val pctEncoded: Parser1[Char] =
+    (char('%') *> digit ~ digit).map { case (a, b) => (((a - '0') >> 4) + (b - '0')).toChar }
```
rossabaker, line 44: Decoding this per-char is likely to be suboptimal, and maybe we should discourage it by returning `Parser1[Unit]`.

core/src/main/scala/org/http4s/Uri.scala
```diff
@@ -210,11 +207,102 @@ object Uri {
   def unsafeFromString(s: String): Uri =
     fromString(s).valueOr(throw _)
 
+  /* hier-part   = "//" authority path-abempty
+   *             / path-absolute
+   *             / path-rootless
+   *             / path-empty
+   */
+  def hierPart(cs: JCharset): Parser[(Option[Authority], Path)] = {
```
rossabaker, line 215: Should be `private[http4s]`.

core/src/main/scala/org/http4s/Uri.scala
```diff
@@ -210,11 +207,102 @@ object Uri {
...
+    import Query.{parser => query}
+
+    /* origin-form    = absolute-path [ "?" query ] */
+    val originForm: Parser1[Uri] =
+      (absolutePath ~ query.?).map { case (p, q) =>
+        Uri(scheme = None, authority = None, path = p, query = q.getOrElse(Query.empty))
+      }
+
+    /* absolute-form = absolute-URI */
+    def absoluteForm: Parser1[Uri] = absoluteUri(StandardCharsets.UTF_8)
```
rossabaker, line 294: Punting on encoding!

core/src/main/scala/org/http4s/Uri.scala
```diff
@@ -210,11 +207,102 @@ object Uri {
...
+
+    /* authority-form = authority */
+    val authorityForm: Parser1[Uri] =
+      authority(StandardCharsets.UTF_8).map(a => Uri(authority = Some(a)))
```
rossabaker, line 298: Punting on encoding!

core/src/main/scala/org/http4s/Uri.scala
```diff
@@ -210,11 +207,102 @@ object Uri {
...
+
+    /* asterisk-form = "*" */
+    val asteriskForm: Parser1[Uri] =
+      char('*').as(Uri(path = "*"))
```
rossabaker, line 302: Should be a constant.

core/src/main/scala/org/http4s/Uri.scala
```diff
@@ -252,20 +340,30 @@ object Uri {
...
     def unsafeFromString(s: String): Scheme =
       fromString(s).fold(throw _, identity)
 
-    @nowarn("cat=deprecation")
+    /* scheme      = ALPHA *( ALPHA / DIGIT / "+" / "-" / "." ) */
+    private[http4s] val parser: Parser1[Scheme] = {
+      import Parser.{charIn, rep, string1}
+      import Rfc3986.{alpha, digit}
+
+      string1("https")
```
rossabaker, line 354: `fromString` going through a map would be faster here.

core/src/main/scala/org/http4s/Uri.scala
```diff
@@ -252,20 +340,30 @@ object Uri {
...
+          (alpha *> rep(alpha.orElse1(digit).orElse1(charIn("+-.")))).string.map(new Scheme(_)))
+    }
+
     private[http4s] trait Parser { self: PbParser =>
       def scheme =
         rule {
-          "https" ~ !Alpha ~ push(https) |
-            "http" ~ !Alpha ~ push(http) |
+          "https" ~ Alpha.unary_!() ~ push(https) |
+            "http" ~ Alpha.unary_!() ~ push(http) |
```
rossabaker, line 366: I hit a weird compiler warning. :shrug:

core/src/main/scala/org/http4s/Uri.scala
```diff
@@ -285,8 +383,64 @@ object Uri {
   }
 
   type Path = String
+
+  object Path {
```
rossabaker, line 387: All of these will get a lot more sophisticated on the new model in main, and the merge is going to suck.  This will be one of the first really interesting merges on cats-effect-3-dotty.

core/src/main/scala/org/http4s/Uri.scala
```diff
@@ -905,8 +1212,8 @@ object Uri {
         out.put('+')
       } else {
         out.put('%')
-        out.put(HexUpperCaseChars((c >> 4) & 0xF))
-        out.put(HexUpperCaseChars(c & 0xF))
+        out.put(HexUpperCaseChars((c >> 4) & 0xf))
+        out.put(HexUpperCaseChars(c & 0xf))
```
rossabaker, line 1216: I'm concerned that scalafmt is touching some of these things like this.

rossabaker (commented):
This code is putrescent.

The chickens are coming home to roost putting the parsers on the companion objects for non-primitive types.  We do this so we can access private constructors with parsed values.  But it scatters cats-parse code and imports everywhere, and the build is going to fail with a NullPointerException due to initialization order.

A second problem is that we are totally unprincipled about what's encoded and what's not, and what charset we're decoding to.  Most of this can only be taken on in main, but maybe we can leave ourselves in a better place here with tagged types or something.

isomarcte:
@rossabaker is your current intent to start fresh over in https://github.com/typelevel/cats-uri ?

rossabaker:
More or less.  I'd like to take a closer look at [scala-uri](https://github.com/lemonlabsuk/scala-uri) before investing too much into a cats-uri, but I think there's consensus that there's value to a URI outside http4s, and also a consensus that our 0.21 version stinks.

PR #4096, opened 2020-12-28
Remove unsafe run sync

yanns:
We should try to use `unsafeRunSync` only at the end.

### DIFF
--- a/client/src/test/scala/org/http4s/client/ClientRouteTestBattery.scala
+++ b/client/src/test/scala/org/http4s/client/ClientRouteTestBattery.scala
@@ -44,7 +44,7 @@ abstract class ClientRouteTestBattery(name: String)
     new HttpServlet {
       override def doGet(req: HttpServletRequest, srv: HttpServletResponse): Unit =
         GetRoutes.getPaths.get(req.getRequestURI) match {
-          case Some(r) => renderResponse(srv, r)
+          case Some(r) => renderResponse(srv, r).unsafeRunSync()
           case None => srv.sendError(404)
         }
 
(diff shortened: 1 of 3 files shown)


client/src/test/scala/org/http4s/client/ClientRouteTestBattery.scala
```diff
@@ -128,29 +128,31 @@ abstract class ClientRouteTestBattery(name: String)
...
-      .through(writeOutputStream[IO](IO.pure(srv.getOutputStream), closeAfterUse = false))
-      .compile
-      .drain
-      .unsafeRunSync()
-  }
+  private def renderResponse(srv: HttpServletResponse, response: IO[Response[IO]]): IO[Unit] =
+    for {
+      resp <- response
+      _ <- IO(srv.setStatus(resp.status.code))
+      _ <- IO(resp.headers.foreach { h =>
```
isomarcte, line 150: Nit: Subjectively, I prefer `traverse_` over a `foreach` call.

isomarcte (approved):

client/src/test/scala/org/http4s/client/testroutes/GetRoutes.scala
```diff
@@ -34,19 +32,18 @@ object GetRoutes {
...
-          .pure[IO],
-        DelayedPath ->
-          F.sleep(1.second) *>
-          Response[IO](Ok).withEntity("delayed path").pure[IO],
-        NoContentPath -> Response[IO](NoContent).pure[IO],
-        NotFoundPath -> Response[IO](NotFound).withEntity("not found").pure[IO],
-        EmptyNotFoundPath -> Response[IO](NotFound).pure[IO],
-        InternalServerErrorPath -> Response[IO](InternalServerError).pure[IO]
-      ))(_.unsafeRunSync())
+  def getPaths(implicit F: Temporal[IO]): Map[String, IO[Response[IO]]] =
```
rossabaker, line 35: This one is currently very dumb even on 0.21.

client/src/test/scala/org/http4s/client/ClientRouteTestBattery.scala
rossabaker: And wrap the `srv.addHeader` side effect in `IO`?  It's all local, so it doesn't make much difference to me, but that does highlight the effective bit a little better, so, :+1:.

yanns: https://github.com/http4s/http4s/pull/4096/commits/c26a6cbac341e6721ed6f47c1607e6e7e6d298e3

rossabaker (approved):

PR #4097, opened 2020-12-28
blaze-server on CE3

RaasAhsan:
This was fairly straightforward, we just need to think more about where we want to expose the `Dispatcher` arguments


RaasAhsan:
ah, we'll need to fix blaze-core up again since fs2 deprecates `Queue`, which is why the build is failing now

blaze-core/src/test/scala/org/http4s/blazecore/websocket/WSTestHead.scala
```diff
@@ -74,28 +73,37 @@ sealed abstract class WSTestHead(
...
         case Right(wsFrame) =>
           Some(wsFrame)
       }
 
-  def pollBatch(batchSize: Int, timeoutSeconds: Long): IO[List[WebSocketFrame]] =
-    outQueue
-      .dequeueChunk1(batchSize)
-      .map(_.toList)
+  def pollBatch(batchSize: Int, timeoutSeconds: Long): IO[List[WebSocketFrame]] = {
+    def batch(acc: List[WebSocketFrame]): IO[List[WebSocketFrame]] =
```
RaasAhsan, line 94: we should probably add a batched take operation to the CE queue, but it's getting mighty close to fs2 realm

rossabaker: If Queue moved to CE3, then wouldn't the operation?

blaze-server/src/main/scala/org/http4s/server/blaze/BlazeBuilder.scala
```diff
@@ -238,7 +241,7 @@ class BlazeBuilder[F[_]](
 
 @deprecated("Use BlazeServerBuilder instead", "0.20.0-RC1")
 object BlazeBuilder {
-  def apply[F[_]](implicit F: ConcurrentEffect[F], timer: Timer[F]): BlazeBuilder[F] =
+  def apply[F[_]](D: Dispatcher[F])(implicit F: Async[F]): BlazeBuilder[F] =
```
rossabaker, line 244: If we have to change the signature, I don't think there's any reason to keep the deprecation.  This is only there to help people for whom it no longer compiles.

blaze-server/src/main/scala/org/http4s/server/blaze/BlazeServerBuilder.scala
```diff
@@ -398,12 +404,11 @@ class BlazeServerBuilder[F[_]](
 
 object BlazeServerBuilder {
   @deprecated("Use BlazeServerBuilder.apply with explicit executionContext instead", "0.20.22")
-  def apply[F[_]](implicit F: ConcurrentEffect[F], timer: Timer[F]): BlazeServerBuilder[F] =
-    apply(ExecutionContext.global)
+  def apply[F[_]](D: Dispatcher[F])(implicit F: Async[F]): BlazeServerBuilder[F] =
```
rossabaker, line 407: Let's remove this (now modified) overload as well.  It's been suggested we should just default the `ExecutionContext` to global, which I might do in a separate PR.

blaze-server/src/main/scala/org/http4s/server/blaze/BlazeServerBuilder.scala
```diff
@@ -105,8 +106,9 @@ class BlazeServerBuilder[F[_]](
     httpApp: HttpApp[F],
     serviceErrorHandler: ServiceErrorHandler[F],
     banner: immutable.Seq[String],
-    val channelOptions: ChannelOptions
-)(implicit protected val F: ConcurrentEffect[F], timer: Timer[F])
+    val channelOptions: ChannelOptions,
+    D: Dispatcher[F]
```
rossabaker, line 110: How about a `withDispatcher`, too?  I think we're doing that on other backends that take one.

blaze-server/src/main/scala/org/http4s/server/blaze/Http1ServerStage.scala
```diff
@@ -189,20 +192,24 @@ private[blaze] class Http1ServerStage[F[_]](
       case Right(req) =>
         executionContext.execute(new Runnable {
           def run(): Unit = {
-            val action = Sync[F]
-              .suspend(raceTimeout(req))
+            val action = F
+              .defer(raceTimeout(req))
```
rossabaker, line 196: I can't recall why we're deferring this.  `raceTimeout` returns in `F`, no?

blaze-server/src/main/scala/org/http4s/server/blaze/Http1ServerStage.scala
```diff
@@ -189,20 +192,24 @@ private[blaze] class Http1ServerStage[F[_]](
...
 
             parser.synchronized {
-              cancelToken = Some(
-                F.runCancelable(action) {
-                  case Right(()) => IO.unit
-                  case Left(t) =>
-                    IO(logger.error(t)(s"Error running request: $req")).attempt *> IO(
-                      closeConnection())
-                }.unsafeRunSync())
+              // TODO: review blocking compared to CE2
```
rossabaker, line 201: I'm having a hard time digesting the synchronization here when it runs a background job.  I don't think you changed the semantics, but this is ... odd looking.

blaze-server/src/main/scala/org/http4s/server/blaze/Http1ServerStage.scala
```diff
@@ -311,10 +326,8 @@ private[blaze] class Http1ServerStage[F[_]](
 
   private def cancel(): Unit =
     cancelToken.foreach { token =>
-      F.runAsync(token) {
-        case Right(_) => IO(logger.debug("Canceled request"))
-        case Left(t) => IO(logger.error(t)("Error canceling request"))
-      }.unsafeRunSync()
+      Await.result(token(), Duration.Inf)
```
rossabaker, line 329: Is `Duration.Inf` safe here?

rossabaker (commented):
Impressive work.  Comments below.

blaze-core/src/test/scala/org/http4s/blazecore/websocket/WSTestHead.scala
RaasAhsan: CE3 implements a brand new `Queue`  separate from `fs2.concurrent.Queue`. CE3 doesn't have a concept of chunking (this was deemed an fs2 concern), but I think it probably makes sense to add something anyways

Originally the plan was for fs2 Queue to stick around but it seems like Mike made things work out with the CE3 one

blaze-server/src/main/scala/org/http4s/server/blaze/Http1ServerStage.scala
RaasAhsan: `fa.unsafeRunSync` is ultimately an `Await.result(fa, Duration.Inf)` in both CE2 and CE3, so it does preserve semantics

blaze-server/src/main/scala/org/http4s/server/blaze/BlazeServerBuilder.scala
RaasAhsan: Missed it!

blaze-server/src/main/scala/org/http4s/server/blaze/Http1ServerStage.scala
RaasAhsan:
```scala
private[this] val raceTimeout: Request[F] => F[Response[F]] =
    responseHeaderTimeout match {
        val timeoutResponse = F.sleep(finite).as(Response.timeout[F])
        req => F.race(runApp(req), timeoutResponse).map(_.merge)
      case _ =>
        runApp
    }
```
I don't see any strictly evaluated side effects here... so it looks safe to remove

blaze-server/src/main/scala/org/http4s/server/blaze/Http1ServerStage.scala
```diff
@@ -189,20 +192,24 @@ private[blaze] class Http1ServerStage[F[_]](
...
-                      closeConnection())
-                }.unsafeRunSync())
+              // TODO: review blocking compared to CE2
+              val fa = action.attempt.flatMap {
+                case Right(_) => F.unit
+                case Left(t) =>
+                  F.delay(logger.error(t)(s"Error running request: $req")).attempt *> F.delay(
+                    closeConnection())
+              }
+              val (_, token) = D.unsafeToFutureCancelable(fa)
```
RaasAhsan, line 208: I'll make a CE PR for `Dispatcher#unsafeRunCancelable` for convenience

blaze-server/src/main/scala/org/http4s/server/blaze/Http1ServerStage.scala
```diff
@@ -189,20 +192,24 @@ private[blaze] class Http1ServerStage[F[_]](
...
             parser.synchronized {
-              cancelToken = Some(
-                F.runCancelable(action) {
-                  case Right(()) => IO.unit
-                  case Left(t) =>
-                    IO(logger.error(t)(s"Error running request: $req")).attempt *> IO(
-                      closeConnection())
-                }.unsafeRunSync())
+              // TODO: review blocking compared to CE2
+              val fa = action.attempt.flatMap {
```
RaasAhsan, line 202: We can just chain the transformations on `action` itself, shouldn't make a difference

blaze-server/src/main/scala/org/http4s/server/blaze/Http1ServerStage.scala
RaasAhsan: The next revision will make it a lot more clearer, but it basically synchronizes on starting the effect and retrieving the cancel token, which is protected by `parser`

blaze-server/src/main/scala/org/http4s/server/blaze/BlazeServerBuilder.scala
RaasAhsan: Has a naming convention been settled on for `Dispatcher`? I don't really like `D` because, though it looks like one, it's not a typeclass instance, so I'm inclined to change everything to `dispatcher`

RaasAhsan:
@rossabaker Seems like `BlazeBuilder` is deprecated in favor of `BlazeServerBuilder`, should we go ahead and just delete?

blaze-server/src/main/scala/org/http4s/server/blaze/BlazeServerBuilder.scala
rossabaker: I like `dispatcher`.

blaze-server/src/main/scala/org/http4s/server/blaze/Http1ServerStage.scala
rossabaker: Okay, that's a good point.

rossabaker:
Yes, `BlazeBuilder` can go.  It only makes sense to keep deprecations as long as they're source compatible to help people upgrade.  That one no longer is, so it's time for it to go.

RaasAhsan:
I think everything looks good here now

rossabaker (approved):

PR #4098, opened 2020-12-28
migrate parsers for zipkin headers

m-sp:
wasn't sure about the two tests with the trailing whitespaces. `fromParser` uses `parseAll` so expects to exactly match the input. am I correct in the assumption hat the trailing whitespaces should be handled earlier in the pipeline already?


core/src/main/scala/org/http4s/headers/X-B3-Flags.scala
```diff
@@ -17,12 +17,17 @@
...
-import org.http4s.parser.HttpHeaderParser
+import cats.syntax.all._
+import cats.parse.Parser
+import cats.parse.Numbers
 import org.http4s.util.Writer
 
 object `X-B3-Flags` extends HeaderKey.Internal[`X-B3-Flags`] with HeaderKey.Singleton {
   override def parse(s: String): ParseResult[`X-B3-Flags`] =
-    HttpHeaderParser.X_B3_FLAGS(s)
+    ParseResult.fromParser(parser, "X-B3-Flags header")(s)
```
rossabaker, line 27: All the other ones say "Invalid ...".  (Which is perhaps an indication we should move `"Invalid "` into the helper.)  But let's stay consistent for now.

core/src/main/scala/org/http4s/headers/X-B3-Flags.scala
```diff
@@ -17,12 +17,17 @@
...
+
+  private[http4s] val parser: Parser[`X-B3-Flags`] =
+    (Numbers.digits <* Parser.end).map(str => fromLong(str.toLong))
```
rossabaker, line 30: I don't think this `Parser.end` is right.  It won't allow this to be composed with other parsers with remaining output.  The `parseAll` in `fromParser` should do what we want.

rossabaker, line 30: That `str.toLong` can throw an exception.  How about:

```scala
.mapFilter { str =>
  try Some(str.toLong)
  case { _: NumberFormatException => None }
}
```

I did similar in the URI parser.

build.sbt
```diff
@@ -129,6 +129,13 @@ lazy val core = libraryProject("core")
       ProblemFilters.exclude[DirectMissingMethodProblem]("org.http4s.parser.HttpHeaderParser.AUTHORIZATION"),
       ProblemFilters.exclude[DirectMissingMethodProblem]("org.http4s.parser.HttpHeaderParser.RANGE"),
       ProblemFilters.exclude[DirectMissingMethodProblem]("org.http4s.parser.HttpHeaderParser.CONTENT_RANGE"),
+      ProblemFilters.exclude[DirectMissingMethodProblem]("org.http4s.parser.HttpHeaderParser.X_B3_TRACEID"),
+      ProblemFilters.exclude[DirectMissingMethodProblem]("org.http4s.parser.HttpHeaderParser.X_B3_SPANID"),
+      ProblemFilters.exclude[DirectMissingMethodProblem]("org.http4s.parser.HttpHeaderParser.X_B3_PARENTSPANID"),
+      ProblemFilters.exclude[DirectMissingMethodProblem]("org.http4s.parser.HttpHeaderParser.X_B3_FLAGS"),
+      ProblemFilters.exclude[DirectMissingMethodProblem]("org.http4s.parser.HttpHeaderParser.X_B3_SAMPLED"),
+      ProblemFilters.exclude[IncompatibleTemplateDefProblem]("org.http4s.parser.ZipkinHeader"),
```
rossabaker, line 137: I'm trying to keep this list in alphabetical order to reduce merge conflict pain.  That ordering includes the type of exclusion, so it's just a pipe through sort in $EDITOR.

core/src/main/scala/org/http4s/headers/X-B3-Sampled.scala
```diff
@@ -17,12 +17,14 @@
...
 
-import org.http4s.parser.HttpHeaderParser
 import org.http4s.util.Writer
+import cats.parse.{Parser, Parser1, Rfc5234}
 
 object `X-B3-Sampled` extends HeaderKey.Internal[`X-B3-Sampled`] with HeaderKey.Singleton {
   override def parse(s: String): ParseResult[`X-B3-Sampled`] =
-    HttpHeaderParser.X_B3_SAMPLED(s)
+    ParseResult.fromParser(parser, "X-B3-Sampled header")(s)
+  val parser: Parser1[`X-B3-Sampled`] =
```
rossabaker, line 26: All of our parsers have been `private[http4s]` so far.  Im not committed to that, but let's keep it consistent until we decide.

rossabaker (changes requested):
Nice job.  Just a few minor comments.

core/src/main/scala/org/http4s/headers/X-B3-Flags.scala
rossabaker: It ate my earlier comment.  In response to your questoin, I think whitespace is the header's responsibility.  We're at the level of `field-value` here:

```
   header-field = field-name ":" OWS field-value OWS
```

m-sp: yeah I originally had `toLongOption` but unfortunately this is scala 2.13 only it seems 😞

m-sp: thanks for confirming, removed the `Parser.end`

build.sbt
m-sp: I hope I understood your comment correctly and the sorting now makes more sense

rossabaker (approved):
This looks great, thanks!

PR #4102, opened 2020-12-29
Port ok http client

ashwinbhaskar:


okhttp-client/src/main/scala/org/http4s/client/okhttp/OkHttpBuilder.scala
```diff
@@ -89,12 +84,11 @@ sealed abstract class OkHttpBuilder[F[_]] private (
   private def run(req: Request[F]) =
     Resource.suspend(F.async[Resource[F, Response[F]]] { cb =>
       okHttpClient.newCall(toOkHttpRequest(req)).enqueue(handler(cb))
-      ()
+      F.delay(Option(F.unit))
```
rossabaker, line 87: This is really an `F[Option[F[Unit]]`?

It looks `pure` instead of `delay`ed, at least.

rossabaker (commented):
This looks like a good start, but we'll want a CE3 expert as well.

okhttp-client/src/main/scala/org/http4s/client/okhttp/OkHttpBuilder.scala
```diff
@@ -56,27 +56,22 @@ import scala.util.control.NonFatal
...
-    val blocker: Blocker
-)(implicit protected val F: ConcurrentEffect[F], cs: ContextShift[F])
+    val okHttpClient: OkHttpClient
+)(implicit protected val F: Async[F])
     extends BackendBuilder[F, Client[F]] {
   private[this] val logger = getLogger
 
+  private def invokeCallback[F[_]](logger: Logger)(f: => Unit)(implicit
+      F: Sync[F]): Unit =
+    F.blocking(f).unsafeRunSync
```
djspiewak, line 66: Is there a Dispatcher floating around that I don't see? Also does this need to be blocking?

okhttp-client/src/main/scala/org/http4s/client/okhttp/OkHttpBuilder.scala
SystemFw: The easiest way to migrate old uncancelable `async` code, is to use `async_` , which has the same signature of the old async. The old code didn't shift back after async, whereas with ce3 it will, fyi.

ashwinbhaskar: @rossabaker yes, a pure should work. My bad. But yes, it is an `F[Option[F[Unit]]]`

okhttp-client/src/main/scala/org/http4s/client/okhttp/OkHttpBuilder.scala
ashwinbhaskar: Added the dispatcher and made a few changes.

okhttp-client/src/main/scala/org/http4s/client/okhttp/OkHttpBuilder.scala
```diff
@@ -36,46 +35,59 @@ import okhttp3.{
...
+  * we will call unsafeRunSync
   *
   * @define WHYNOSHUTDOWN It is assumed that the OkHttp client is
   * passed to us as a Resource, or that the caller will shut it down, or
   * that the caller is comfortable letting OkHttp's resources expire on
   * their own.
   *
   * @param okHttpClient the underlying OkHttp client.
-  * @param blockingExecutionContext $BLOCKINGEC
+  * @param dispatcher $BLOCKINGEC
```
rossabaker, line 56: I think the reference needs to be changed to $DISPATCHER here.

okhttp-client/src/main/scala/org/http4s/client/okhttp/OkHttpBuilder.scala
```diff
@@ -36,46 +35,59 @@ import okhttp3.{
...
 sealed abstract class OkHttpBuilder[F[_]] private (
     val okHttpClient: OkHttpClient,
-    val blocker: Blocker
-)(implicit protected val F: ConcurrentEffect[F], cs: ContextShift[F])
+    val dispatcher: Dispatcher[F]
+)(implicit protected val F: Async[F])
     extends BackendBuilder[F, Client[F]] {
   private[this] val logger = getLogger
 
+  type Result[F[_]] = Either[Throwable, Resource[F, Response[F]]]
```
rossabaker, line 65: Can this alias be private to reduce API liability?

rossabaker (commented):
This is looking good to me.  Let's see if the CE3 people see anything else.

okhttp-client/src/main/scala/org/http4s/client/okhttp/OkHttpBuilder.scala
ashwinbhaskar: done

okhttp-client/src/main/scala/org/http4s/client/okhttp/OkHttpBuilder.scala
ashwinbhaskar: done

rossabaker (approved):
This looks good to me.

PR #4105, opened 2020-12-29
Port ClientRouteTestBattery tests

cquiroz:
Port of `ClientRouteTestBattery` to munit. It is used to test each client.


async-http-client/src/test/scala/org/http4s/client/asynchttpclient/AsyncHttpClientSuite.scala
```diff
@@ -0,0 +1,107 @@
...
+  def clientResource: Resource[IO, Client[IO]] = AsyncHttpClient.resource[IO]()
+
+  test(
+    "AsyncHttpClient configure should evaluate to the defaultConfiguration given the identity function as the configuration function") {
+    val defaultConfig = AsyncHttpClient.defaultConfig
+    val customConfig = AsyncHttpClient.configure(identity)
+
+    assertEquals(defaultConfig.getMaxConnectionsPerHost, customConfig.getMaxConnectionsPerHost)
+    assertEquals(defaultConfig.getMaxConnections, customConfig.getMaxConnections)
+    assertEquals(defaultConfig.getRequestTimeout, customConfig.getRequestTimeout)
```
rossabaker, line 36: These don't need to be chained with `*>`?  It's just IO assertions that do?

rossabaker (approved):

client/src/test/scala/org/http4s/client/ClientRouteTestBattery.scala
```diff
@@ -56,93 +50,95 @@ abstract class ClientRouteTestBattery(name: String)
...
-          val body = client.run(Request[IO](uri = uri)).use(e => IO.pure(e.status))
-          body must returnValue(Ok)
-        }
+  // Need to override the context shift from munitCatsEffect
+  // This is only required for JettyClient
+  implicit val contextShift: ContextShift[IO] = Http4sSuite.TestContextShift
 
-        "Repeat a simple request" in {
-          val path = GetRoutes.SimplePath
+  clientFixture.test("Strip fragments from URI".ignore) { case (_, client) =>
```
cquiroz, line 60: I'm not totally sure why was this ignored in the first place

async-http-client/src/test/scala/org/http4s/client/asynchttpclient/AsyncHttpClientSuite.scala
cquiroz: Exactly, IO assertions not chained won't be executed

client/src/test/scala/org/http4s/client/ClientRouteTestBattery.scala
rossabaker: I don't know what's external about it.  It doesn't look like a good test.  A request-target with a fragment is illegal, so we'd want to be sure that the request line submitted by a client has no fragment.  Different servers have different tolerances for it.  Jetty allows it and Tomcat forbids it, for instance.  Reading directly from a socket to see what's sent would be the most reliable way to test it. but a pain.

cquiroz: Maybe I should remove it then

PR #4108, opened 2020-12-30
Fold dotty branch into main

rossabaker:
Ramifications:
* There will be no Dotty release of 0.21.x unless someone else backports it.
* We retire the `dotty` and `cats-effect-dotty` branches.
* We'll add a 3.0.0-Mx crossbuild when it's ready
* We can develop with `++3.0.0-M2!` in the meantime
* All changes need to keep the Scala 2 builds green
* We can fix deficiencies in the model as we continue to port parsers.
* We spend less time merging and hopefully finish these projects sooner.

Relates to #4106.


rossabaker:
5e98dfe is execrable.  I threw `lazy` at it until the NullPointers stopped, and the path parser needs refactoring, but if it's passing, I'm more interested in flattening the branches and sorting it out in its own PR.

tests/src/test/scala/org/http4s/parser/UriParserSpec.scala
```diff
@@ -138,75 +138,11 @@ class UriParserSpec extends Http4sSpec {
       check(relativeUris)
     }
 
-    "parse absolute URI with fragment" in {
```
hamnis, line 141: are these tests not valid?

hamnis: There are from #4132 i see, but not why the tests where deleted.

rossabaker: A request target can't have a fragment.  We're using the same model, so the types don't say we can't, but we're testing invalid behavior.

I am now questioning why the test on line 297 should be removed, though.

RaasAhsan: That was my mistake, all these tests are nested under the `requestTarget` label so I deleted all of the ones that mentioned fragments. That test in particular uses `fromString` though

core/src/main/scala/org/http4s/MediaType.scala
```diff
@@ -104,9 +88,34 @@ object MediaRange {
...
+
+  private[http4s] val mediaTypeExtensionParser: Parser1[(String, String)] = {
+    import Parser.char
+    import Rfc2616BasicRules.optWs
+    import org.http4s.internal.parsing.Rfc7230.{quotedString, token}
+
+    val escapedString = "\\\\"
+    val unescapedString = "\\"
+
+    (char(';') *> optWs *> token ~ (char('=') *> token.orElse(quotedString)).?).map {
```
hamnis, line 101: optWs is from `Rfc2616BasicRules.optWs` we have a `ows` in `Rfc7230` as well, I suggest we use that one instead.

hamnis (approved):

RaasAhsan (approved):

cquiroz (approved):

core/src/main/scala/org/http4s/MediaType.scala
rossabaker: This one is bloated enough, but I'll mop that up in a follow-up.

PR #4109, opened 2020-12-30
Don't render body on blaze requests of certain methods

rossabaker:
Based on [this gitter report](https://gitter.im/http4s/http4s?at=5fec4de8acd1e516f8c1bdcd).

If the method is known to have "no defined semantics" when sending a payload, don't.

### DIFF
--- a/blaze-client/src/main/scala/org/http4s/client/blaze/Http1Connection.scala
+++ b/blaze-client/src/main/scala/org/http4s/client/blaze/Http1Connection.scala
@@ -29,7 +29,7 @@ import org.http4s.{headers => H}
 import org.http4s.Uri.{Authority, RegName}
 import org.http4s.blaze.pipeline.Command.EOF
 import org.http4s.blazecore.Http1Stage
-import org.http4s.blazecore.util.Http1Writer
+import org.http4s.blazecore.util.{BodylessWriter, Http1Writer}
 import org.http4s.headers.{Connection, Host, `Content-Length`, `User-Agent`}
 import org.http4s.util.{StringWriter, Writer}
 import scala.annotation.tailrec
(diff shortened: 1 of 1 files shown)


blaze-client/src/main/scala/org/http4s/client/blaze/Http1Connection.scala
```diff
@@ -383,4 +389,7 @@ private object Http1Connection {
       writer
     } else writer
   }
+
+  private val NoPayloadMethods: Set[Method] =
+    Set(Method.GET, Method.DELETE, Method.CONNECT, Method.TRACE)
```
rossabaker, line 394: This information might be encoded on the Method itself.  (We have a `NoBody` type that's on the way out, but perhaps not even used?  And there's a difference between "MUST not" and "undefined semantics").

Anyway, this does the right thing here.

rossabaker (commented):
It's naughty that I didn't write a test.  We're starting to develop a need for a test helper that creates a socket and gathers the input.

rossabaker:
Yeah, it wasn't this simple.

RaasAhsan:
I've been digging around these parts so I'll wrap things up for you

blaze-client/src/main/scala/org/http4s/client/blaze/Http1Connection.scala
RaasAhsan: Do you think we should instead just omit the `Content-Length` header if there are no chunks to be written (the observed behavior is an implementation detail of `CachingChunkWriter`)? It might be unintuitive to users who try to submit a body with a `GET` and don't see it written. IIRC this is a poorly defined part of the spec anyways

If we really want to restrict users from passing bodies to methods which don't "support" it, doing that with type-level constraints seems like a more appropriate solution

RaasAhsan: I think a caveat here might be that, we would want to send the `Content-Length: 0` header for POST requests with empty bodies but not for GET requests

rossabaker:
Superseded by #4152.

PR #4112, opened 2020-12-30
Add explicit instance for accessing mixed multipart decoder

adschwartz:
Add an explicit instance to allow access to the mixed multipart decoder.

Referencing this discussion: https://gitter.im/http4s/http4s?at=5fecd60de7f693041f3efd9c


rossabaker (approved):
Thanks! :+1: The test failure is unrelated.  I restarted it.

I neglected to mention that we should target series/0.21 to get things into the 0.21 release series.  This is binary compatible, so we can get it out there sooner that way.  I will take care of cherry-picking this.

adschwartz:
Hi @rossabaker. I can make a new 0PR to that branch. Coming right up

PR #4116, opened 2020-12-31
Port metrics tests

cquiroz:
Signed-off-by: Carlos Quiroz <carlos.m.quiroz@gmail.com>


cquiroz:
I undid the deletion of the old Matchers files as it breaks mima compatibility.

rossabaker (approved):
Which matchers?  `IOMatchers` or `Http4sMatchers`?

cquiroz:
I deprecated `Http4sLegacyMatchers` and `Http4sLegacyMatchersIO`

testing/src/test/scala/org/http4s/testing/Http4sLegacyMatchers.scala
```diff
@@ -24,6 +24,7 @@ import org.http4s.headers._
 import org.http4s.util.CaseInsensitiveString
 import org.specs2.matcher.{RunTimedMatchers => Specs2RunTimedMatchers, _}
 
+@deprecated("Use Http4sSuite and munit-cats-effect-instead", "0.21.6")
```
rossabaker, line 27: Wait, these are in `src/test`.  We should be able to just eliminate them, no?

Also, next one is 0.21.16, not 0.21.6.

cquiroz: I'd be happy to delete them 😀

PR #4113, opened 2020-12-31
Mixed multipart decoder -> series 0.21

adschwartz:
Add an explicit instance to allow access to the mixed multipart decoder.

Referencing this discussion: https://gitter.im/http4s/http4s?at=5fecd60de7f693041f3efd9c

PR #4117, opened 2020-12-31
Restore defaults on MuliptartDecoder.mixedMultipart

rossabaker:
Necessary for binary compatibility.


rossabaker:
This was overlooked in #4113.  It's troubling that CI didn't catch that.

cquiroz (approved):
LGTM

Issue #2278, opened 2018-11-15
WritePendingException: null from client


rossabaker:
I can sometimes hit this `evalMap`ping a basic `GET` request 20,000 times.

```
2018-11-15 02:30:44 ERROR [scala-execution-context-global-346] [] [Http1Connection.scala:153] - Error rendering request
java.nio.channels.WritePendingException: null
    at sun.nio.ch.AsynchronousSocketChannelImpl.write(AsynchronousSocketChannelImpl.java:352)
    at sun.nio.ch.AsynchronousSocketChannelImpl.write(AsynchronousSocketChannelImpl.java:416)
    at org.http4s.blaze.channel.nio2.ByteBufferHead.org$http4s$blaze$channel$nio2$ByteBufferHead$$go$1(ByteBufferHead.scala:42)
    at org.http4s.blaze.channel.nio2.ByteBufferHead.writeRequest(ByteBufferHead.scala:58)
    at org.http4s.blaze.channel.nio2.ByteBufferHead.writeRequest(ByteBufferHead.scala:25)
    at org.http4s.blaze.channel.nio2.ByteBufferHead.writeRequest(ByteBufferHead.scala:15)
    at org.http4s.blaze.pipeline.Tail.channelWrite(Stages.scala:96)
    at org.http4s.blaze.pipeline.Tail.channelWrite$(Stages.scala:94)
    at org.http4s.client.blaze.ReadBufferStage.channelWrite(ReadBufferStage.scala:13)
    at org.http4s.client.blaze.ReadBufferStage.writeRequest(ReadBufferStage.scala:20)
    at org.http4s.blaze.pipeline.Tail.channelWrite(Stages.scala:96)
    at org.http4s.blaze.pipeline.Tail.channelWrite$(Stages.scala:94)
    at org.http4s.blazecore.ResponseHeaderTimeoutStage.channelWrite(ResponseHeaderTimeoutStage.scala:13)
    at org.http4s.blazecore.ResponseHeaderTimeoutStage.writeRequest(ResponseHeaderTimeoutStage.scala:39)
    at org.http4s.blaze.pipeline.Tail.channelWrite(Stages.scala:96)
    at org.http4s.blaze.pipeline.Tail.channelWrite$(Stages.scala:94)
    at org.http4s.client.blaze.Http1Connection.channelWrite(Http1Connection.scala:24)
    at org.http4s.blazecore.util.CachingChunkWriter.doWriteEnd(CachingChunkWriter.scala:74)
[...]

PR #4123, opened 2021-01-02
blaze-client to ce3

RaasAhsan:
TODO:
- [x] Port the specs to `Http4sSuite` if possible
- [x] Think about providing a `Dispatcher` directly in the suite trait
- [x] Should a `Dispatcher` be passed in to the client or be created by the client? Same question goes for the server builders. IMO it's exactly the same as passing in `UnsafeRun`. All we need is an `Async` constraint but the only `Dispatcher` constructor exposes a `Resource` so we would have to reach for `allocated`. I think this is a more general question we should discuss in CE land


RaasAhsan:
Ah yep, looks like `Http1ClientStageSpec` is one of the nondeterministic tests

rossabaker (commented):
I wonder if it's possible to run our tests that are skipOnCi, but that's purely extra credit.

RaasAhsan:
I'll give it a try. And also, I just realized that I'm porting all these specs to munit on the ce3 branch.. maybe it would've better to do it elsewhere

rossabaker:
@cquiroz has been starting them all the way down on series/0.21, which helps with merge conflicts moving forward.  If it's not too much trouble, that's a better starting point, but we can backport.

cquiroz:
It maybe better do the tests on series/0.21

RaasAhsan:
fortunately blaze-client was small enough, so I'll avoid it on the ce3 branch in the future and backport

RaasAhsan:
Is `ClientTimeoutSpec.No Response head timeout on fast header` known to be flaky on CI?

RaasAhsan:
Fixed it (some issue with pooling). I also re-enabled all the flaky ones and they seem to be doing good so far. I think this is ready for a deeper look, I'm gonna think about the two extra PR items more

blaze-client/src/test/scala/org/http4s/client/blaze/BlazeClientSuite.scala
```diff
@@ -96,7 +97,7 @@ class BlazeClientSuite extends BlazeClientBase {
           } yield r
         }
         .map(_.isRight)
-        .assert
+        .assertEquals(true)
```
cquiroz, line 100: Why revert this? Is it because `munit-cats-effect` in this branch is older?

blaze-client/src/test/scala/org/http4s/client/blaze/Http1ClientStageSuite.scala
```diff
@@ -0,0 +1,323 @@
...
+  }
+
+  test("Fail when attempting to get a second request with one in progress") {
+    val tail = mkConnection(FooRequestKey)
+    val (frag1, frag2) = resp.splitAt(resp.length - 1)
+    val h = new SeqTestHead(List(mkBuffer(frag1), mkBuffer(frag2), mkBuffer(resp)))
+    LeafBuilder(tail).base(h)
+
+    try {
+      tail.runRequest(FooRequest, IO.never).unsafeRunAsync {
```
cquiroz, line 147: Maybe you can use `intercept` and avoid the unsafe call?

blaze-client/src/test/scala/org/http4s/client/blaze/Http1ClientStageSuite.scala
```diff
@@ -0,0 +1,323 @@
...
+        case Right(_) => (); case Left(_) => ()
+      } // we remain in the body
+
+      intercept[Http1Connection.InProgressException.type] {
+        tail
+          .runRequest(FooRequest, IO.never)
+          .unsafeRunSync()
```
cquiroz, line 154: I think this can be avoided and just return `IO([A]`

blaze-client/src/test/scala/org/http4s/client/blaze/Http1ClientStageSuite.scala
```diff
@@ -0,0 +1,323 @@
...
+
+      val response = tail.runRequest(headRequest, IO.never).unsafeRunSync()
+      assertEquals(response.contentLength, Some(contentLength))
+
+      // connection reusable immediately after headers read
+      assert(tail.isRecyclable)
+
+      // body is empty due to it being HEAD request
+      val length = response.body.compile.toVector
+        .unsafeRunSync()
```
cquiroz, line 283: I think this could be avoided and do `length.assertEquals...`

cquiroz (commented):
Thanks for porting the tests, it will be great to get them running on ce3

blaze-client/src/test/scala/org/http4s/client/blaze/BlazeClientSuite.scala
RaasAhsan: Yeah there is no `assert` but `interceptMessage` works a lot better here

rossabaker (approved):
Build is failing on unused import, but this looks good to me after that.

RaasAhsan:
The build is probably going to fail because of a flaky blaze-client test. They seem to be happening because of this (https://github.com/http4s/http4s/issues/2278) which I have long heard tales about. I'm going to spend some time today figuring it out (Slack is down) because I have a really low tolerance for flaky builds, but race conditions are also super fun. :)

rossabaker:
Are you able to reproduce it?  I have struggled for years to reproduce the legendarily flaky blaze tests, and I can't make them happen but for on CI.

RaasAhsan:
My usual approach (which has worked for a great deal of CE bugs) is to first look at the code analytically and see if I can piece out a race condition. If that doesn't work after a few hours, I try to find a reproduction by running the offending code in a hot loop and then try to distill the pathological ordering of events with printlns and whatnot

rossabaker:
Well, it didn't flake here.  Should we merge this, or are you worried about destabilizing the CE3 branch that's running pretty well right now?

RaasAhsan:
We can merge, my goal is to have this figured out by EOD 🤞

PR #4124, opened 2021-01-02
CirceEntityEncoder: avoid intermediate ByteBuffer duplication

diesalbla:
In the `EntityEncoder` for Json, there is a call to the `apply` method of `fs2.Chunk.ByteBuffer` companion object, which
_duplicates_ the given byte buffer. By default, the printer from `circe` generates a new ByteBuffer for each Json, so we do not need to duplicate it. WE avoid it by calling the `view` method instead.

This should save memory allocations proportional to the size of the response.


Daenyth (approved):

cquiroz (approved):
LGTM

rossabaker:
Good find. I'm going to cherry-pick this back to 0.21.

Issue #4086, opened 2020-12-27
Port ember-server to cats-effect-3


rossabaker:

m-sp:
If no one else is working on this, I can take care of this

PR #4125, opened 2021-01-02
port ember server to ce3

m-sp:
* removed an unused typeclass constraint in encoder
* removed the deprecated method instead of adapting it
* the `EmberServerBuilder` takes now `Async` as a constraint, alternatively we could use the `Network` effect type introduced by fs2 and refined in https://github.com/typelevel/fs2/pull/2189

fixes https://github.com/http4s/http4s/issues/4086

### DIFF
--- a/ember-server/src/main/scala/org/http4s/ember/server/internal/ServerHelpers.scala
+++ b/ember-server/src/main/scala/org/http4s/ember/server/internal/ServerHelpers.scala
@@ -18,6 +18,7 @@ package org.http4s.ember.server.internal
 
 import fs2._
 import fs2.concurrent._
+import fs2.io.Network
 import fs2.io.tcp._
 import fs2.io.tls._
 import cats.effect._
@@ -40,7 +41,7 @@ private[server] object ServerHelpers {
   private val close = Connection(NonEmptyList.of(closeCi))
   private val keepAlive = Connection(NonEmptyList.one(CIString("keep-alive")))
 
-  def server[F[_]: Concurrent: ContextShift](
+  def server[F[_]: Network](
       bindAddress: InetSocketAddress,
       httpApp: HttpApp[F],
       sg: SocketGroup,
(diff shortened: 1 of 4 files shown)


m-sp:
while everything seems to compile just fine, this actually depends on https://github.com/ChristopherDavenport/log4cats which doesn't have a ce3 build yet

ember-server/src/main/scala/org/http4s/ember/server/internal/ServerHelpers.scala
```diff
@@ -58,7 +59,7 @@ private[server] object ServerHelpers {
       idleTimeout: Duration = 60.seconds,
       additionalSocketOptions: List[SocketOptionMapping[_]] = List.empty,
       logger: Logger[F]
-  )(implicit C: Clock[F]): Stream[F, Nothing] = {
+  )(implicit C: Temporal[F]): Stream[F, Nothing] = {
```
rossabaker, line 62: I usually avoid mixing context bonuds and named implicits.  Once I have to name one, I name them all.  I'd probably have an `F: Network[F]` (or maybe `N: Network[F]`) here.

I don't have a strong reason for that, and it was already that way when you found it, so I also don't mind being ignored here.

rossabaker (commented):
Looks good to me once we resolve the log4cats dependency.

m-sp:
seems log4cats was migrated to the typelevel organization, does anybody per chance know if there is a ce3 migration planed?

I think this is about to block a couple of migrations soonish, so if you prefer we could comment out the logger until log4cats is done?

rossabaker:
Yes, such a migration is very certainly planned.  Vault and Unique are my priorities, because those are core, but this and keypool will be next.

I'm fine commenting it if we get stuck, but getting unstuck on these will be my next OSS task.

rossabaker:
Incorrectly autoclosed when cats-effect-3 branch was merged and deleted.

rossabaker:
I guess we need to prioritize the log4cats release ahead of keypool?  log4cats unblocks this, and both are needed for ember-client?

m-sp:
> I guess we need to prioritize the log4cats release ahead of keypool? log4cats unblocks this, and both are needed for ember-client?

correct, thanks for all your work btw

yanns:
What is the status of this branch?
Is it blocked by https://github.com/http4s/http4s/issues/4232 and https://github.com/http4s/http4s/issues/4237?

For both, we now have a PR.
Can I help anyhow?

m-sp:
> What is the status of this branch?
> Is it blocked by #4232 and #4237?
> 
> For both, we now have a PR.
> Can I help anyhow?

it's blocked by https://github.com/typelevel/log4cats/pull/382

m-sp:
see https://github.com/http4s/http4s/pull/4256

PR #4126, opened 2021-01-02
move misc projects to ce3

m-sp:


rossabaker (approved):
Well, that moves things forward nicely!  There's perhaps a lot more to be implemented for scalafix, but I'd rather have it compiling now and add rules later.

PR #4128, opened 2021-01-03
start port of ember client to ce3

m-sp:
temporarily inlines keypool. log4catsSlf4j is not yet on ce3 so the tests are going to fail


rossabaker:

m-sp:
see https://github.com/http4s/http4s/pull/4256

PR #4132, opened 2021-01-03
Cats parse uri

RaasAhsan:
Opening up another PR (original at #4095) so I can run builds here


core/src/main/scala/org/http4s/Query.scala
```diff
@@ -212,6 +213,19 @@ object Query {
         case Left(_) => Vector.empty
       }
 
+  /* query       = *( pchar / "/" / "?" )
+   *
+   * These are illegal, but common in the wild.  We will be
+   * "conservative in our sending behavior and liberal in our
+   * receiving behavior", and encode them.
+   */
+  private[http4s] def parser: Parser[Query] = {
```
RaasAhsan, line 222: This was a `val` in the original PR, there were some initialization ordering issues that caused a NPE

core/src/main/scala/org/http4s/headers/Link.scala
```diff
@@ -29,49 +30,59 @@ object Link extends HeaderKey.Internal[Link] with HeaderKey.Recurring {
...
+              case Title(title) => lv.copy(title = Some(title))
+              case Type(tpe) => lv.copy(`type` = Some(tpe))
+            }
+          }
+      }
 
     headerRep1(linkValueWithAttr).map(links => Link(links.head, links.tail: _*))
   }
+
+  sealed trait LinkParam
```
RaasAhsan, line 81: The old parboiled2 parser is doing some magic to manipulate the `LinkValue` that was parsed earlier.. I don't think we can express that kind of parsing with cats-parse, so I introduced this intermediary type for the link parameters, and then the `LinkValue` is augmented afterwards.

core/src/main/scala/org/http4s/Query.scala
rossabaker: This is one of the dangerous of putting the parsers on the companions.  We don't get compiler help with the ordering of eager vals.  I'm not sure how heavy creating a new `Parser` instance every time is.  We possibly want a lazy val here, or to understand the cycle that caused this.  But this gets us moving forward.

core/src/main/scala/org/http4s/headers/Link.scala
rossabaker: Is this all internal to the parser?  If so, we could inline it in the parser.  (Maybe you did... I'm trying to read the indentation in the diff.)

core/src/main/scala/org/http4s/headers/Origin.scala
```diff
@@ -76,8 +75,8 @@ object Origin extends HeaderKey.Internal[Origin] with HeaderKey.Singleton {
       .string
       .map(Uri.Scheme.unsafeFromString)
     val stringHost = until1(char(':').orElse(`end`)).map(RegName.apply)
-    val bracketedIpv6 = char('[') *> Rfc3986.ipv6 <* char(']')
-    val host = List(bracketedIpv6, Rfc3986.ipv4, stringHost).reduceLeft(_ orElse1 _)
+    val bracketedIpv6 = char('[') *> Uri.Ipv6Address.parser <* char(']')
+    val host = List(bracketedIpv6, Uri.Ipv4Address.parser, stringHost).reduceLeft(_ orElse1 _)
```
rossabaker, line 79: I think we should unnest these types from the `Uri`, but that can wait until we merge this to main.

tests/src/test/scala/org/http4s/UriSpec.scala
```diff
@@ -94,9 +94,7 @@ class UriSpec extends Http4sSpec with MustThrownMatchers {
         Uri.fromString("http://example.org/a file") must beLeft(
           ParseFailure(
             "Invalid URI",
-            """Invalid input ' ', expected Pchar, '/', '?', '#' or 'EOI' (line 1, column 21):
-http://example.org/a file
-                    ^""".replace("\r", "")
+            "Error(20,NonEmptyList(EndOfString(20,25)))"
```
rossabaker, line 97: Future note: this is something we'll want to improve in `ParseResult.parseFrom`.  There's a `LocationMap` that will help with the arrow.  And bosatsu allegedly has a nicer formatter already.

rossabaker (commented):
Thank you so much for picking this up!

core/src/main/scala/org/http4s/headers/Link.scala
RaasAhsan: Nope, I need to do that :)

rossabaker:
Interesting: a few of these builds got canceled for hanging.

rossabaker:
The build hang is a pre-existing condition, and in trying to stabilize that, I hit the warning on the Link parser that this fixes.  I think we should merge this to dotty now, and I'll beat on the hanging before we merge dotty to main.

tests/src/test/scala/org/http4s/parser/Rfc3986ParserSpec.scala
```diff
@@ -36,143 +35,142 @@ class Rfc3986ParserSpec extends Http4sSpec {
...
     "not parse an invalid ipv6 address containing an invalid letter such as 'O'" in {
       val invalidIp = "1200:0000:AB00:1234:O000:2552:7777:1313"
-      Rfc3986.ipv6.parse(invalidIp).map(_._2) must beLeft
+      Ipv6Address.fromString(invalidIp) must beLeft
     }
 
     "only parse part of an invalid ipv6 address where a single section is shortened (must be 2 or more)" in {
       val invalidIp = "2001:db8::1:1:1:1:1"
-      Rfc3986.ipv6.parse(invalidIp).map(_._2.value) must beRight("2001:db8::1:1:1:1")
+      Ipv6Address.fromString(invalidIp).map(_.value) must beLeft
```
rossabaker, line 118: Clarified with @lewisjkl on #4004: they were Rights because the tests are using `parse`, and don't consume all input.  They flipped from left to right when the `Parser.end`s were removed.

`Ipv6Address.fromString` uses `parseAll`, which is a "clientside end", which correctly flips these back to lefts.  Jeff was right, and now so is this.

PR #4133, opened 2021-01-04
Finish okhttp-client for cats-effect-3

rossabaker:
It didn't compile on Scala 2.12 because that doesn't have `chaining`.  @ashwinbhaskar already did the hard work.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -30,7 +30,7 @@ lazy val modules: List[ProjectReference] = List(
   // blazeClient,
   // asyncHttpClient,
   // jettyClient,
-  // okHttpClient,
+  okHttpClient,
   // servlet,
   // jetty,
   // tomcat,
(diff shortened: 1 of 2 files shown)


ashwinbhaskar (approved):

rossabaker:
Oops, that still didn't compile.  The private logger isn't referenced.

ashwinbhaskar:
@rossabaker the logger is referenced in `logTap` method

rossabaker:
Yeah, I think it needed to move to the companion with `logTap`.  I'm on another branch right now. :sweat: 

I'm not sure how this compiled for me.  I probably missed an error.

ashwinbhaskar:
@rossabaker no worries, would you like me to fix it and push?

rossabaker:
That would be great, thanks!

ashwinbhaskar:
@rossabaker did you intend to move logTap and the type alias to the companion object? If so, we will not be able to keep them private.

rossabaker:
I did, but the builder class should be able to access the private members of the companion with the `import OkHttpBuilder._`.

ashwinbhaskar:
@rossabaker Created a new PR - https://github.com/http4s/http4s/pull/4136

PR #4138, opened 2021-01-04
Fixed boolean blindness of `EntityTag`'s weakness indicator

### LINKED ISSUE #3956: Fix boolean blindness of `EntityTag`'s weakness indicator
An entity tag can be "weak" or "strong", but we model it as `true` or `false`.  Without looking, is `true` strong or weak?  Don't know?  That's a form of "boolean blindness."  (Knowing, but having several such flags as parameters, is another.)

In the companion object of `EntityTag`:
1. Add a `sealed trait Weakness`.  
2. Extend `Product with Serializable`. This helps with type inference
3. Create a `Weak` and `Strong` that implement weakness.
4. Replace the variable.
5. Follow the compiler errors to glory.

Communicate with whoever claims #3955: whoever goes second is going to have a merge conflict if you don't!

This is a breaking change and should go on `main`.

lukoyanov:
Introduced new ADT for `ETag` weakness param.

Fixes #3956

### DIFF
--- a/core/src/main/scala/org/http4s/headers/ETag.scala
+++ b/core/src/main/scala/org/http4s/headers/ETag.scala
@@ -20,20 +20,28 @@ package headers
 import org.http4s.parser.HttpHeaderParser
 import org.http4s.util.Writer
 import cats.Show
+import org.http4s.headers.ETag.EntityTag.{Strong, Weakness}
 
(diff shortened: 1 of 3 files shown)


rossabaker (commented):
Nice, thank you!  Looks like it needs a scalafmt.

PR #4139, opened 2021-01-04
Stabilize the Dotty build

rossabaker:


rossabaker:
Still hangs in the same places after ef37e3c, though I like it better.

rossabaker:
There's a circular dependency in the MediaType and MediaRange companions that ought to be rethought, but I'm just trying to stop the bleeding right now.

rossabaker:
Passed all but the 2.13.3 / Java 15 build, which hung.

I could reproduce prior hangs locally, but not this one.

rossabaker:
Now it's 2.13.3 on 1.8 hanging.  We're going to need some thread dumps.

rossabaker:
Cleaned out the deadlock detector.

I'm afraid there are more, and running a cyclic dependency checker, particularly for statics, would be a smart move.  But this passed all the builds once.

rossabaker:
I'm merging this into #4108, but feel free to complain about it here.

rossabaker:
Included in #4108.

PR #4142, opened 2021-01-04
Poor's man discipline to run codec laws

cquiroz:
We have been using `discipline` to run laws but that is not possible when using `munit-scalacheck-effect`. `discipline` uses `Prop` which are a different entity than `PropF` than `munit-scalacheck-effect` uses.

This PR adds machinery to run a set of `PropF[IO]` inside an `HttpsSuite`

# Pros:
* It reuses the same laws as before
* no `unsafeRunXXX` calls
* No need for `Eq[F[Boolean]]` we can do it with `Eq[A]`

# Cons:
* It is a different family than `discipline`. The old ones are not deleted but switching to the new ones will force users into `scalacheck-effect`
* `EntityEncoderTest` takes a `Sync` constraint, now it needs `Effect`

This is not a full blown law testing framework but solves the problem of `Eq[F[Boolean]]`

What do you think?

If this approach seems right I should port all tests and also some of the other laws


laws/src/main/scala/org/http4s/laws/discipline/EntityCodecTests.scala
```diff
@@ -44,6 +42,14 @@ trait EntityCodecTests[F[_], A] extends EntityEncoderTests[F, A] {
         laws.entityCodecRoundTrip(a)
       }
     )
+
+  def entityCodecF(implicit
```
cquiroz, line 46: Look ma, no `Eq[F[Boolean]]`

laws/src/main/scala/org/http4s/laws/EntityEncoderLaws.scala
```diff
@@ -23,7 +23,7 @@ import cats.laws._
 import org.http4s.headers.{`Content-Length`, `Transfer-Encoding`}
 
 trait EntityEncoderLaws[F[_], A] {
-  implicit def F: Sync[F]
+  implicit def F: Effect[F]
```
cquiroz, line 26: I need `Effect` as we have to eventually evaluate on `IO` to use `munit-cats-effect` assertions

rossabaker:
I have a hunch that fs2 has some of these same problems. Summoning @mpilquist and @SystemFw for their thoughts, to see if this is happening at the right level.  Feels like there's a contribution here that might go into one of the other test libraries and then be used by us.

mpilquist:
I don't have great news unfortunately -- in fs2, we use discipline to test compliance with laws from cats and cats-effect (in `StreamLawsSpec` and `PullLawsSpec`) along with `TestContext` and otherwise use scalacheck-effect and `PropF` with "real" concurrency for testing everything else.

testing/src/test/scala/org/http4s/Http4sLawSuite.scala
```diff
@@ -15,18 +15,20 @@
...
+import org.scalacheck.effect.PropF
 
-// Originally based on ArgonautSuite
-class CirceSpec extends Http4sSpec {
-  implicit val testContext = TestContext()
-  checkAll("EntityCodec[IO, Json]", EntityCodecTests[IO, Json].entityCodec)
+/** Poor's man discipline runner to check a set of PropF[F]
+  */
+trait Http4sLawSuite extends ScalaCheckEffectSuite {
+  def checkAllPropF[F[_]](
```
rossabaker, line 25: Nit: if `checkAll` is a `Prop`, shouldn't this just be `checkAllF`?

cquiroz: Sure, I wasn't really sure how to name it.

cquiroz:
I probably need some help to solve the mima issue on the build.

On the making it a lib front, that certainly crossed my mind but I rather focused on getting it to work, `LawAdapter` may not be general enough for a lib

rossabaker:
I'm not sure I see a path to MiMa compatibility here.  Maybe this needs to just target main and be part of series/0.22.  0.21 is winding down anyway?

PR #4146, opened 2021-01-06
Fix typo

fghibellini:
See https://tools.ietf.org/html/rfc7231#section-7.1.1.1 vs https://tools.ietf.org/html/rfc7321


rossabaker (approved):
I can cite an distressing number of RFCs by number, but I learned these backward and I'll never get them right. Thanks!

isomarcte (approved):

rossabaker:
I'm going to merge #4108 before this because I'm tired of conflicts on that, but we'll get this in today.

PR #4148, opened 2021-01-06
Lifted ETag.EntityTag to org.http4s package

### LINKED ISSUE #3955: Lift ETag.EntityTag to org.http4s
Most headers are in `org.http4s.header`, and most HTTP-defined values of those headers are in the `org.http4s` root package.  

1. Lift the inner `EntityTag` class out of the `org.http4s.headers.ETag` and into `org.http4s`.
2. Leave a deprecated type alias in `ETag` pointing to the new location of the class.
3. Leave a deprecated val in `ETag` pointing to the new location of the companion object.
4. Update any references in the code.

This work is a binary-breaking change, and should be done on the main branch.

lukoyanov:
Lifted ETag.EntityTag to org.http4s which fixes #3955


rossabaker (approved):
Thanks!  This looks right to me.  It will have some merge conflicts after we merge the dotty branch, which I'd like to do first, and then we'll get this.

PR #4149, opened 2021-01-07
Port async-http-client to ce3

RaasAhsan:


async-http-client/src/main/scala/org/http4s/client/asynchttpclient/AsyncHttpClient.scala
```diff
@@ -47,28 +47,33 @@ object AsyncHttpClient {
...
-        ()
+        F.delay(
+          httpClient.executeRequest(toAsyncRequest(req, dispatcher), asyncHandler(cb, dispatcher)))
+          .as(None)
       })
     }
 
   /** Allocates a Client and its shutdown mechanism for freeing resources.
     */
   def allocate[F[_]](config: AsyncHttpClientConfig = defaultConfig)(implicit
```
RaasAhsan, line 62: Do we really need `allocate` and `stream` as constructors? I have similar sentiments on `apply`. At this point they should probably stay for compatibility, but I think having a single resource constructor is solid (I suspect most users are using the resource constructor anyways)

rossabaker: I can't remember why we did.  I was going to say to DRY up with `.stream`, but even that uses `.resource`.  I'd be happy to deprecate the superfluous ones.

rossabaker: `apply` takes one that's already configured, and thus can be an implementation other than `DefaultAsyncHttpClient` and is to my eyes more justifiable than `allocate` and `stream`.  It does feel less like an `apply` after getting the `dispatcher` argument, however.  And not providing it means lifting the result into a `Resource` that doesn't actually shut down the client.  I think you've done the best you can do here.

RaasAhsan: We could also change the `apply` signature to something like:
```scala
def apply[F[_]](httpClient: AsyncHttpClient)(implicit F: Async[F]): Resource[F, Client[F]]
```

ChristopherDavenport: Resource didn't have allocate when we initally created these.

rossabaker: I don't love still calling it `apply` when it doesn't return the companion class (or when the return type changes!), but I prefer this signature.

rossabaker:

async-http-client/src/main/scala/org/http4s/client/asynchttpclient/AsyncHttpClient.scala
RaasAhsan: I think it's worth a rename to `fromClient`, will do that

rossabaker (approved):

PR #4152, opened 2021-01-07
Conditionally omit empty content length in blaze-client

RaasAhsan:
Alternative approach to #4109. It seems kind of awkward that we have to do this but it is what it is. Though I think it's worth entertaining because it doesn't make a semantically breaking change: people may be submitting bodies on GET requests today, even though that may be frowned upon

Also, apologies for submitting this on the cats-effect-3 branch


rossabaker (approved):
I'm good with this approach and willing to backport it.  It's not so cats-effecty, so I might even just cherry-pick this one and let it merge its way through.

rossabaker:
Incorrectly autoclosed when cats-effect-3 branch was merged.  But I was going to cherry-pick it anyway.  I'll do that next.

PR #4154, opened 2021-01-07
migrate multiple simple header parser

m-sp:
migrates 
* `Age`
* `Connection`
* `Content-Length`
* `Strict-Transport-Security`


m-sp:
hmm seems the tests failed, I'll take a look

core/src/main/scala/org/http4s/headers/Connection.scala
```diff
@@ -18,16 +18,23 @@ package org.http4s
...
 
 // values should be case insensitive
 //http://stackoverflow.com/questions/10953635/are-the-http-connection-header-values-case-sensitive
 object Connection extends HeaderKey.Internal[Connection] with HeaderKey.Recurring {
   override def parse(s: String): ParseResult[Connection] =
-    HttpHeaderParser.CONNECTION(s)
+    ParseResult.fromParser(parser, "Invalid Connection header")(s)
+
+  private[http4s] val parser =
+    Parser.rep1Sep(Rfc2616.token, 1, Rfc2616BasicRules.listSep).map { (xs: NonEmptyList[String]) =>
```
rossabaker, line 35: I think this should be the headerRep1 helper.

core/src/main/scala/org/http4s/parser/AdditionalRules.scala
```diff
@@ -213,4 +214,11 @@ private[http4s] object AdditionalRules {
   def EOI = P.char('\uFFFF')
 
   def EOL = optWs *> EOI.rep
+
+  val Digits: Parser1[Long] = P.rep1(Rfc3986.digit, 1).string.mapFilter { s =>
```
rossabaker, line 218: There's a `digit` in `cats.parse.Rfc5234` that we can use.  I guess the value in having a separate rule is that yours does the work of parsing a run of them to a long, while the cats-parse one just gives you the Char.

I think I'd call this `Long`, to show that arbitrary lengths will fail. Or maybe even `NonNegativeLong`, because a leading minus won't work here.  I do think it's useful.

core/src/main/scala/org/http4s/headers/Strict-Transport-Security.scala
```diff
@@ -56,7 +57,33 @@ object `Strict-Transport-Security`
     fromLong(maxAge, includeSubDomains, preload).fold(throw _, identity)
 
   def parse(s: String): ParseResult[`Strict-Transport-Security`] =
-    HttpHeaderParser.STRICT_TRANSPORT_SECURITY(s)
+    ParseResult.fromParser(parser, "Invalid `Strict-Transport-Security` header")(s)
+
+  private[http4s] val parser: Parser[`Strict-Transport-Security`] = {
+    val maxAge: Parser1[`Strict-Transport-Security`] =
+      (Parser.string1("max-age=") *> AdditionalRules.Digits).map { (age: Long) =>
```
rossabaker, line 64: Is this case sensitive?  I bet it's not, but I'd have to find the RFC.

rossabaker (commented):
Thanks!

core/src/main/scala/org/http4s/headers/Strict-Transport-Security.scala
m-sp: https://tools.ietf.org/html/rfc6797#section-6.1

> Directive names are case-insensitive.

thanks you are right, will change

rossabaker (approved):

PR #4157, opened 2021-01-08
Move cats-effect-3 to main

rossabaker:
The port isn't quite complete, but it's close, and is the final piece of #4106.  The old main is now forked to series/0.22, freeing up this default branch for CE3 development on the 1.0 line.


rossabaker:
@cquiroz, please see changes around `LawAdapter` and the `EntityCodec` tests.  I had to boost some constraints for lack of `.toIO`, but I think I expressed it properly.

rossabaker:
These fatal warnings on CI only are a nuisance. :rage:

rossabaker:
It appears that I have again neglected to run the tests before flitting off to another branch.

rossabaker:
Ugh, that BatchingExecutor problem is Scala 2.12 only. :fire:

testing/src/test/scala/org/http4s/Http4sSuite.scala
```diff
@@ -25,6 +25,9 @@ import munit._
 /** Common stack for http4s' munit based tests
   */
 trait Http4sSuite extends CatsEffectSuite with DisciplineSuite with munit.ScalaCheckEffectSuite {
+  // The default munit EC causes an IllegalArgumentException in
+  // BatchExecutor on Scala 2.12.
+  override val munitExecutionContext = Http4sSpec.TestExecutionContext
```
cquiroz, line 30: This is interesting. I think I had to use `TestExecutionContext` in a few places too

laws/src/main/scala/org/http4s/laws/discipline/LawAdapter.scala
```diff
@@ -34,11 +34,11 @@ trait LawAdapter {
     propLabel -> PropF.boolean(prop)
 
   def isEqPropF[F[_], A: Arbitrary: Shrink, B: Eq](propLabel: String, prop: A => IsEq[F[B]])(
-      implicit F: Sync[F]): (String, PropF[F]) =
+      implicit F: MonadThrow[F]): (String, PropF[F]) =
```
cquiroz, line 37: This looks right

testing/src/test/scala/org/http4s/Http4sSuite.scala
rossabaker: I discussed with @RaasAhsan a bit last night, and I think we can fork tests and get rid of some of our specialized test thread pools.  But I'm just trying to land this plane.

Issue #4151, opened 2021-01-07
Flaky blaze-client test: Use User-Agent header provided in Request 


RaasAhsan:
I'm going to start making issues for these so I can look at them in more detail some time.

```
==> X org.http4s.client.blaze.Http1ClientStageSuite.Use User-Agent header provided in Request  0.011s munit.FailException: /home/runner/work/http4s/http4s/blaze-client/src/test/scala/org/http4s/client/blaze/Http1ClientStageSuite.scala:207 assertion failed
206:      val requestLines = request.split("\r\n").toList
207:      assert(requestLines.contains("User-Agent: myagent"))
208:      assertEquals(response, "done")
    at munit.FunSuite.assert(FunSuite.scala:11)
    at org.http4s.client.blaze.Http1ClientStageSuite.$anonfun$new$30(Http1ClientStageSuite.scala:207)
    at org.http4s.client.blaze.Http1ClientStageSuite.$anonfun$new$30$adapted(Http1ClientStageSuite.scala:205)
    at cats.effect.IOFiber.mapK(IOFiber.scala:970)
    at cats.effect.IOFiber.succeeded(IOFiber.scala:794)
    at cats.effect.IOFiber.mapK(IOFiber.scala:979)
    at cats.effect.IOFiber.succeeded(IOFiber.scala:794)
    at cats.effect.IOFiber.uncancelableSuccessK(IOFiber.scala:1080)
    at cats.effect.IOFiber.succeeded(IOFiber.scala:805)
    at cats.effect.IOFiber.unmaskSuccessK(IOFiber.scala:1097)
    at cats.effect.IOFiber.succeeded(IOFiber.scala:806)
    at cats.effect.IOFiber.onCancelSuccessK(IOFiber.scala:1065)
    at cats.effect.IOFiber.succeeded(IOFiber.scala:804)
    at cats.effect.IOFiber.asyncContinueR(IOFiber.scala:916)
    at cats.effect.IOFiber.run(IOFiber.scala:136)
    at cats.effect.unsafe.WorkerThread.runFiber(WorkerThread.scala:162)
    at cats.effect.unsafe.WorkerThread.run(WorkerThread.scala:295)
    ```

**Comment by RaasAhsan** [comment]
Another one, which I suspect is the same source of nondeterminism:
```
[...]

PR #4158, opened 2021-01-08
migrate async-http-client to ce3

yanns:
https://github.com/http4s/http4s/issues/4087

### DIFF
--- a/async-http-client/src/main/scala/org/http4s/client/asynchttpclient/AsyncHttpClient.scala
+++ b/async-http-client/src/main/scala/org/http4s/client/asynchttpclient/AsyncHttpClient.scala
@@ -19,14 +19,13 @@ package client
 package asynchttpclient
 
 import cats.effect._
-import cats.effect.concurrent._
 import cats.syntax.all._
-import cats.effect.implicits._
 import fs2.Stream._
 import fs2._
 import fs2.interop.reactivestreams.{StreamSubscriber, StreamUnicastPublisher}
 import _root_.io.netty.handler.codec.http.{DefaultHttpHeaders, HttpHeaders}
 import _root_.io.netty.buffer.Unpooled
+import cats.effect.std.Dispatcher
 import org.asynchttpclient.AsyncHandler.State
 import org.asynchttpclient.handler.StreamedAsyncHandler
 import org.asynchttpclient.request.body.generator.{BodyGenerator, ReactiveStreamsBodyGenerator}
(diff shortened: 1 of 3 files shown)


async-http-client/src/main/scala/org/http4s/client/asynchttpclient/AsyncHttpClient.scala
```diff
@@ -144,38 +145,43 @@ object AsyncHttpClient {
...
+        dispatcher.unsafeRunSync(
+          onStreamCalled.get
+            .ifM(ifTrue = F.unit, ifFalse = invokeCallbackF[F](cb(Right(response -> dispose))))
+        )
     }
 
   // use fibers to access the ContextShift and ensure that we get off of the AHC thread pool
-  private def invokeCallbackF[F[_]](invoked: => Unit)(implicit F: Concurrent[F]): F[Unit] =
-    F.start(F.delay(invoked)).flatMap(_.join)
+  private def invokeCallbackF[F[_]](invoked: => Unit)(implicit F: Async[F]): F[Unit] =
```
yanns, line 163: I'm not sure if we still need that.

yanns:
About the test failure:
```
==> X org.http4s.client.asynchttpclient.AsyncHttpClientSpec.AsyncHttpClient Execute GET  30.004s java.util.concurrent.TimeoutException: Futures timed out after [30 seconds]
```

I cannot reproduce it.
`sbt ++2.12.12 test` is running locally.
Maybe an issue due to local configuration, like the number of cores?

yanns:
OK now another test is failing: `org.http4s.client.blaze.Http1ClientStageSuite`

m-sp:
> OK now another test is failing: `org.http4s.client.blaze.Http1ClientStageSuite`:

pretty sure that this is a known flaky one, see https://github.com/http4s/http4s/issues/4151

rossabaker:

rossabaker:
Uh, wait, I think we might have two of these migrations now. See also #4149.

rossabaker:
Yeah, this is a duplicate. I'm sorry for that -- we didn't get a comment or assignment on the original ticket.  The other one is almost done, but I'll merge main into a local copy of this, look at conflict, and make sure there's nothing we missed.

rossabaker:
Yeah, it looks like algebraically equivalent differences, except for some changes to `apply` that let it manage its own `Dispatcher`.

I'm sorry again for the redundancy.  I'll try to keep a closer eye on the ticket management.

yanns:
No problem.

PR #4159, opened 2021-01-08
Fix warning for missing parens on lambdas for scala3

hamnis:


rossabaker (approved):
It would still be great to automate this one from scalafmt.

hamnis:
yes, i have asked about it on gitter, still no response.

PR #4160, opened 2021-01-08
shutdown dispatcher

yanns:


hamnis (approved):

rossabaker:

rossabaker:
Where does `Fixture` come from?  MUnit?

RaasAhsan:
BTW munit-cats-effect has fixture constructors for `Resource`s:
```scala
import cats.effect.std.Dispatcher

  val dispatcher = ResourceFixture(Dispatcher[IO])

  dispatcher.test("resources can be lifted to munit fixtures") { dsp =>
    dsp.unsafeRunAndForget(IO(42))
  }
```

yanns:
> Where does `Fixture` come from? MUnit?

Yes this is coming from [MUnit](https://scalameta.org/munit/docs/fixtures.html#reusable-suite-local-fixtures)

yanns:
> BTW munit-cats-effect has fixture constructors for `Resource`s:
> 
> ```scala
> import cats.effect.std.Dispatcher
> 
>   val dispatcher = ResourceFixture(Dispatcher[IO])
> 
>   dispatcher.test("resources can be lifted to munit fixtures") { dsp =>
>     dsp.unsafeRunAndForget(IO(42))
>   }
> ```

Yes I saw that. Do you think you should prefer that?
It's ok for me. It would mean more tests to adapt.

rossabaker:
It seems like the munit-cats-effect version requires a lambda per test?  The mixin is syntactically lighter, but too many mixins can get disorienting.  I haven't written enough MUnit tests to have a strong opinion which is nicer.  I defer to @RaasAhsan or @cquiroz.

Either way, this does look like forward progress.

RaasAhsan:
We've been using `ResourceFixture` wherever we can so far. If you decide to go the `ResourceFixture` route, see this https://github.com/typelevel/munit-cats-effect/issues/60 . Should be fixed in the next version whenever we upgrade

yanns:
Suggestion:
1. we can merge this. I'm interesting to see if this change will help to fix flaky tests like https://github.com/http4s/http4s/issues/4151
2. I can work on another version using `ResourceFixture` in another PR, and we can compare the two approaches

rossabaker:
:+1: I'm not optimistic it will help that test, but I think it is an improvement, and we can follow up with the `ResourceFixture` as you suggest.

yanns:
The variant with `ResourceFixture(Dispatcher[IO])` to compare: https://github.com/http4s/http4s/pull/4171

PR #4161, opened 2021-01-08
Use test io runtime in munit

yanns:
I'm wondering if we should use the same `Http4sSpec.TestIORuntime` in all tests / suites.

### DIFF
--- a/blaze-server/src/test/scala/org/http4s/server/blaze/BlazeServerSuite.scala
+++ b/blaze-server/src/test/scala/org/http4s/server/blaze/BlazeServerSuite.scala
@@ -119,11 +119,11 @@ class BlazeServerSuite extends Http4sSuite {
     }
 
   blazeServer.test("route requests on the service executor") { server =>
-    get(server, "/thread/routing").map(_.startsWith("io-compute-")).assertEquals(true)
+    get(server, "/thread/routing").map(_.startsWith("http4s-spec-")).assertEquals(true)
   }
 
   blazeServer.test("execute the service task on the service executor") { server =>
-    get(server, "/thread/effect").map(_.startsWith("io-compute-")).assertEquals(true)
+    get(server, "/thread/effect").map(_.startsWith("http4s-spec-")).assertEquals(true)
   }
 
   blazeServer.test("be able to echo its input") { server =>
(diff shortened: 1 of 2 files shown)


yanns:
Another flaky test:
```
X org.http4s.client.blaze.Http1ClientStageSuite.Not expect body if request was a HEAD request  0.017s munit.FailException: /home/runner/work/http4s/http4s/blaze-client/src/test/scala/org/http4s/client/blaze/Http1ClientStageSuite.scala:256 assertion failed
255:      // connection reusable immediately after headers read
256:      assert(tail.isRecyclable)
```
Different from https://github.com/http4s/http4s/issues/4151

rossabaker:

blaze-server/src/test/scala/org/http4s/server/blaze/BlazeServerSuite.scala
```diff
@@ -119,11 +119,11 @@ class BlazeServerSuite extends Http4sSuite {
     }
 
   blazeServer.test("route requests on the service executor") { server =>
-    get(server, "/thread/routing").map(_.startsWith("io-compute-")).assertEquals(true)
+    get(server, "/thread/routing").map(_.startsWith("http4s-spec-")).assertEquals(true)
```
rossabaker, line 122: I think that these were testing that they ran on a separate executor, and we might wish to make that more explicit by providing one for these tests. This feels like a test that shouldn't have had to change.

rossabaker (commented):
I think this is a right thing to do for now, though I'm interested in the fs2 model of running tests in a fork and not having specialized pools that scale to zero.  The reason we have these pools at all is to avoid a memory leak in long-running SBT sessions, which a simple fork also addresses.

blaze-server/src/test/scala/org/http4s/server/blaze/BlazeServerSuite.scala
yanns: I was also wondering if those changes have revealed an issue or not.
But I am not capable of answering that question myself.

rossabaker: I don't think you've revealed an issue with the code, but I think with how the test is written. https://github.com/http4s/http4s/issues/4166

rossabaker (approved):
:+1:

PR #4169, opened 2021-01-09
Remove uses of Uri.uri

hamnis:
As a part of removing the Uri.uri macro.

We can then remove it from Scala 3, but keep it around for Scala 2 if we want.


rossabaker (approved):
I think this is good. I'm worried about additional imports if we proceed with cats-uri (i.e., `import cats.uri.implicits._`), but I think this is still a better way to use it.

PR #4171, opened 2021-01-09
use ResourceFixture(Dispatcher[IO])

yanns:
following https://github.com/http4s/http4s/pull/4160, try if using `ResourceFixture(Dispatcher[IO])` is better than the current approach.

### DIFF
--- a/blaze-client/src/test/scala/org/http4s/client/blaze/Http1ClientStageSuite.scala
+++ b/blaze-client/src/test/scala/org/http4s/client/blaze/Http1ClientStageSuite.scala
@@ -19,7 +19,7 @@ package client
 package blaze
 
 import cats.effect._
-import cats.effect.std.Queue
+import cats.effect.std.{Dispatcher, Queue}
 import fs2.Stream
 
 import java.nio.ByteBuffer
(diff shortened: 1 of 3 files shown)


blaze-client/src/test/scala/org/http4s/client/blaze/Http1ClientStageSuite.scala
```diff
@@ -46,14 +46,23 @@ class Http1ClientStageSuite extends Http4sSuite with DispatcherIOFixture {
   // Common throw away response
   val resp = "HTTP/1.1 200 OK\r\nContent-Length: 4\r\n\r\ndone"
 
-  private val fooConnection = FunFixture[Http1Connection[IO]](
-    setup = { _ =>
-      mkConnection(FooRequestKey)
-    },
-    teardown = { tail => tail.shutdown() }
-  )
+  private def fooConnection: FunFixture[Http1Connection[IO]] =
```
yanns, line 49: Here, I'd have liked a `dispatcher.flatMap { dispatcher => FunFixture[Http1Connection[IO]](...) }`

rossabaker:
Thanks!  @RaasAhsan is our CE3 specialist and @cquiroz is our test specialist.  I'm going to ping them so I can keep working on the Davenverse, but if nobody has time to look at it, I'll get to it in the next couple days.

blaze-client/src/test/scala/org/http4s/client/blaze/Http1ClientStageSuite.scala
RaasAhsan: Check this out: https://scalameta.org/munit/docs/fixtures.html

Use FunFixture.map2 to compose multiple fixtures into a single fixture.
```
// Fixture with access to two temporary files.
val files2 = FunFixture.map2(files, files)
```

yanns: I'm speaking of `flatMap` and not of `map`.
I've used `map` where I can:
https://github.com/http4s/http4s/pull/4171/files#diff-08a7880af56e3e87db979c2e99ff16fd593789286322aca32c82d0351617aa5cR42

My issue is when one fixture needs another one:
```
  def dispatcher: FunFixture[Dispatcher[IO]] = ...
  private def fooConnection(dispachter: dispachter[IO]): FunFixture[Http1Connection[IO]] = ...
  def fixture: FunFixture[Http1Connection[IO]] = dispatcher.flatMap(d => fooConnection(d))
```
In that latter case, I'd need a `flatMap`.
I found a solution by using `flatMap` on `Resource` itself. But it could be interesting to have the same on `FunFixture`.

RaasAhsan: 👍  This looks good to me. unfortunately `FunFixture` is an invariant functor so this will have to do

RaasAhsan (approved):

Issue #4075, opened 2020-12-27
Fix RetrySuite property based test on CE3


RaasAhsan:
A property-based test in `RetrySuite` was commented out because it caused all the tests afterwards to fail, but only when running the _entire_ test suite in CI. I also got it to fail locally several times, but not deterministically. The error looks like this:
```
==> X org.http4s.client.middleware.RetrySuite.default retriable should defaultRetriable does not resubmit bodies on idempotent methods  0.008s java.lang.IllegalArgumentException: requirement failed
    at scala.Predef$.require(Predef.scala:268)
    at scala.concurrent.BatchingExecutor$Batch.run(BatchingExecutor.scala:55)
    at scala.concurrent.Future$InternalCallbackExecutor$.unbatchedExecute(Future.scala:875)
    at scala.concurrent.BatchingExecutor$Batch.blockOn(BatchingExecutor.scala:96)
    at scala.concurrent.Await$.result(package.scala:146)
    at munit.internal.PlatformCompat$.$anonfun$waitAtMost$1(PlatformCompat.scala:21)
    at scala.util.Try$.apply(Try.scala:213)
    at munit.internal.PlatformCompat$.waitAtMost(PlatformCompat.scala:21)
    at munit.FunSuite.waitForCompletion(FunSuite.scala:51)
    at munit.FunSuite.$anonfun$test$1(FunSuite.scala:37)
    at munit.GenericTest.$anonfun$withBodyMap$1(GenericTest.scala:33)
    at munit.GenericTest.$anonfun$withBodyMap$1(GenericTest.scala:33)
    at munit.MUnitRunner.$anonfun$runTestBody$1(MUnitRunner.scala:290)
```
A dumb solution might to just run it at the very end, but I don't think that's a comprehensive fix for this

PR #4172, opened 2021-01-09
enable retry suite

yanns:
fix https://github.com/http4s/http4s/issues/4075


yanns:
The test seems not to be flaky anymore?

rossabaker (approved):
Well, if it is still flaky, it's easy to revert!

Issue #4058, opened 2020-12-24
Fix CE3 TestIORuntime hanging issue


RaasAhsan:
Follow up to #3857. For some reason, all the `*Spec` test suites would hang when running on the default compute pool that's defined in `Http4sSpec`. Once I changed that to an unbounded pool, they complete successfully. However, they all pass if run in isolation. Not sure if there's some weird interaction with the spec2/munit pools or if there are fairness issues, but it's worth looking into at some point.

PR #4173, opened 2021-01-09
fix blocking AuthenticationSpec

yanns:
fix https://github.com/http4s/http4s/issues/4058

`unsafeRunSync` was used several times inside `IO { }` blocks, blocking threads.

### DIFF
--- a/server/src/test/scala/org/http4s/server/middleware/authentication/AuthenticationSpec.scala
+++ b/server/src/test/scala/org/http4s/server/middleware/authentication/AuthenticationSpec.scala
@@ -40,13 +40,13 @@ class AuthenticationSpec extends Http4sSpec {
   val username = "Test User"
   val password = "Test Password"
 
-  def authStore(u: String) =
+  def authStore(u: String): IO[Option[(String, String)]] =
     IO.pure {
       if (u === username) Some(u -> password)
       else None
     }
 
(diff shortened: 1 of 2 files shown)


rossabaker (commented):
We've been porting lots of tests to MUnit, and prioritizing ones that do `unsafeRunSync()`, so we can take advantage of MUnit's non-blocking model.  That would be an even better approach, but I think this is a good interim step.

Is there anything that prevents most of this from originating on series/0.21 and helping out there?  The `TestIORuntime` is new, but the rest looks like Cats-Effect 2?

yanns:
> We've been porting lots of tests to MUnit, and prioritizing ones that do `unsafeRunSync()`, so we can take advantage of MUnit's non-blocking model. That would be an even better approach, but I think this is a good interim step.

I can try that in another PR if you want.

> Is there anything that prevents most of this from originating on series/0.21 and helping out there? The `TestIORuntime` is new, but the rest looks like Cats-Effect 2?

I don't think so. Do you want another PR cherry-picking that on series/0.21? Or would you prefer only one PR targetting series/0.21?

rossabaker:
I can just cherry-pick from this one.  It's not hard.

If you're up for trying MUnit, that would be a good subsequent PR.

PR #4174, opened 2021-01-09
Merge pull request #4143 from RaasAhsan/blaze-client-bugfix

RaasAhsan:
I think I did this right...


rossabaker (approved):
Oh, cool, this was on my agenda after I got done pouting about the Unique build.

This one is probably worth a patch release.

PR #4175, opened 2021-01-10
Port servlet to cats-effect-3

### LINKED ISSUE #4091: Port servlet to cats-effect-3

ashwinbhaskar:


ashwinbhaskar:
@rossabaker @yanns  this is still WIP. But I wanted to know if I am on the right track. I have ported 3 files to ce3 - `AsyncHttp4sServlet`, `BlockingHttp4sServlet` and `Http4sServlet`. Would request you to review these 3 files especially the `Blocker` changes.

servlet/src/main/scala/org/http4s/servlet/AsyncHttp4sServlet.scala
```diff
@@ -103,7 +107,13 @@ class AsyncHttp4sServlet[F[_]](
           if (servletRequest.isAsyncStarted)
             servletRequest.getAsyncContext.complete()
         )
-      F.runAsync(f)(loggingAsyncCallback(logger)).unsafeRunSync()
+      val result = F
+        .attempt(f)
+        .flatMap {
+          case Right(()) => F.unit
+          case Left(e) => F.delay(logger.error(e)("Error in error handler"))
+        }
```
yanns, line 115: This should work no?

```suggestion
      val result = F
        .attempt(f)
        .flatMap(loggingAsyncCallback(logger))
```

ashwinbhaskar: @yanns `loggingAsyncCallback` requires an implicit `Sync`. But we have an implicit `Async` with us.

yanns:
```
trait Async[F[_]] extends AsyncPlatform[F] with Sync[F] with Temporal[F]
```
So it should work no?

ashwinbhaskar: @yanns I get this error
```
n/scala/org/http4s/servlet/AsyncHttp4sServlet.scala:69:38: could not find implicit value for parameter F: cats.effect.Sync[F] (Could not find an instance of Monad for F)
[error]         .flatMap(loggingAsyncCallback(logger))
```

yanns: OK thanks for trying.
I don't really understand why, but I'll check later.

ashwinbhaskar: @yanns I just noticed that the signature of `loggingAsyncCallback` has a return type of `IO[Unit]`. But our dispatcher is a `Dispatcher[F]`. We can summon a `Dispatcher[IO]`. What do you suggest?

yanns: The return type is `F[Unit]` and not `IO[Unit]` no?

ashwinbhaskar: @yanns the return type of `loggingAsyncCallback`? It is `IO[Unit]`

```
private[http4s] def loggingAsyncCallback[A](logger: Logger)(
      attempt: Either[Throwable, A]): IO[Unit] =
    attempt match {
      case Left(e) => IO(logger.error(e)("Error in asynchronous callback"))
      case Right(_) => IO.unit
    }
```

yanns: I don't have the same version as you have:
```
package object internal {

  private[http4s] def loggingAsyncCallback[F[_], A](logger: Logger)(attempt: Either[Throwable, A])(
      implicit F: Sync[F]): F[Unit] =
    attempt match {
      case Left(e) => F.delay(logger.error(e)("Error in asynchronous callback"))
      case Right(_) => F.unit
    }
```

yanns: Check the version in `main`: https://github.com/http4s/http4s/blob/main/core/src/main/scala/org/http4s/internal/package.scala#L41

yanns: It was changed by https://github.com/http4s/http4s/commit/4c5ab30eccd09b8c3b88d7941ba0eadd185ec9ce

yanns: Maybe you should rebase your change on the newest version of `main`?

ashwinbhaskar: yeah, you were right. I had to rebae with main:) But still this error persists
```
n/scala/org/http4s/servlet/AsyncHttp4sServlet.scala:69:38: could not find implicit value for parameter F: cats.effect.Sync[F] (Could not find an instance of Monad for F)
[error]         .flatMap(loggingAsyncCallback(logger))
```

yanns: ok thanks for thaving tried!

servlet/src/main/scala/org/http4s/servlet/ServletIo.scala
```diff
@@ -113,71 +111,69 @@ final case class NonBlockingServletIo[F[_]: Effect](chunkSize: Int) extends Serv
...
-                cb(Left(bug("Shouldn't have gotten here: I should be the first to set a state")))
-              else
-                in.setReadListener(
-                  new ReadListener {
-                    override def onDataAvailable(): Unit =
-                      state.getAndSet(Ready) match {
-                        case Blocked(cb) => read(cb)
-                        case _ => ()
-                      }
+          F.async_[Option[Chunk[Byte]]] { cb =>
```
rossabaker, line 114: Does this also take care of shifting?

ashwinbhaskar: @rossabaker From what I understood by reading [this](https://github.com/typelevel/cats-effect/tree/series/3.x#ergonomics), shifting happens automatically. Would be great if you could confirm that?

rossabaker: I'm trying to recall why that had to trampoline before.  I think `evalOn` would run the effect on a specified EC and then shift back.  This shifted once and never back, which I don't see a way to do without `ContextShift`.

I'm content to try without this and see if it works.

servlet/src/main/scala/org/http4s/servlet/AsyncHttp4sServlet.scala
```diff
@@ -56,16 +58,17 @@ class AsyncHttp4sServlet[F[_]](
...
-        case Left(t) =>
-          IO(errorHandler(servletRequest, servletResponse)(t))
-      }.unsafeRunSync()
+      val result = F
+        .attempt(
+          toRequest(servletRequest).fold(
+            onParseFailure(_, servletResponse, bodyWriter),
+            handleRequest(ctx, _, bodyWriter)
+          ))
+        .flatMap {
```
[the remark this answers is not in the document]
ashwinbhaskar: @rossabaker how do we cancel the request?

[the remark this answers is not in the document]
rossabaker: Losing the race in `handleRequest` would cancel it.  I'll have to think some more about what we should do about that.  If it passes the tests... I'm good with it for now.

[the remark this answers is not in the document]
ashwinbhaskar: The tests of servlet project pass for now

PR #4179, opened 2021-01-10
Ember server graceful shutdown

### LINKED ISSUE #3067: Ember Server Graceful Shutdown
Currently shutting down an ember server is not graceful, and leads to a very [large stacktrace](https://gist.github.com/BalmungSan/99978e5d699575f19b8cbcc608032a05), we can improve experience by shutting down gracefully and having reporting display that.

RaasAhsan:
Closes #3067. The goal of this PR is to give ember-server a graceful shutdown sequence, which @rossabaker captures succinctly in the original issue. I'd still like to add tests and follow up with some fs2 questions, but I think it's ready for review.


examples/ember/src/main/scala/com/example/http4s/ember/EmberServerSimpleExample.scala
```diff
@@ -26,8 +26,36 @@ import org.http4s.circe._
 import _root_.io.circe._
 import _root_.org.http4s.ember.server.EmberServerBuilder
 
+import scala.concurrent.duration._
```
RaasAhsan, line 29: Going to revert this file at the very end

ember-server/src/main/scala/org/http4s/ember/server/internal/ServerHelpers.scala
```diff
@@ -155,18 +151,22 @@ private[server] object ServerHelpers {
...
-
+    // TODO: The prefetch a really dirty hack that hurts performance a bit.
+    // We want to be able to close the server channel _before_ forcing sockets to close,
+    // so it goes without saying that we need to divorce the scopes that they ultimately live on.
+    // prefetch runs things through a queue and gives us back the consuming stream, so that
+    // when the original stream interrupts, it can finalize while still letting the current
+    // one run. I think we can write a custom combinator that achieves both
+    sg.server[F](bindAddress, additionalSocketOptions = additionalSocketOptions)
+      .interruptWhen(shutdown.signal.attempt)
+      .prefetch
```
RaasAhsan, line 162: Separately, I'm looking into improving the performance of `parJoin` and getting rid of this `prefetch` altogether, but if we're not OK with a slight performance hit we can hold this PR

ember-server/src/test/scala/org/http4s/ember/server/EmberServerSuite.scala
```diff
@@ -0,0 +1,67 @@
...
+
+  def server: FunFixture[Server[IO]] =
+    ResourceFixture(serverResource)
+
+  def fixture: FunFixture[(Server[IO], Client[IO])] =
+    FunFixture.map2(server, client)
+
+  fixture.test("server shutdown waits for outstanding responses to complete") {
+    case (server, client) =>
+      IO.sleep(3.seconds) >> client
```
RaasAhsan, line 62: For some reason the server isn't immediately connect-able to after the resource is acquired. I wonder if there's some detail in the NIO API about this, but I _have_ to wait 3 seconds or else I get a `Connection refused`, which seems to imply that the server socket channel hasn't even been opened yet?

ember-server/src/test/scala/org/http4s/ember/server/EmberServerSuite.scala
```diff
@@ -0,0 +1,67 @@
...
+import org.http4s._
+import org.http4s.client.Client
+import org.http4s.server.Server
+import org.http4s.implicits._
+import org.http4s.dsl.Http4sDsl
+import org.http4s.ember.client.EmberClientBuilder
+
+import scala.concurrent.duration._
+
+class EmberServerSuite extends Http4sSuite {
```
RaasAhsan, line 30: You'll notice that I didn't actually include any tests for the graceful shutdown mechanism... Unsurprisingly that behavior is rather hard to exercise since there's a lot of nondeterministic elements to it. In particular, we can't deterministically start a request _after_ the server socket channel has been closed.

ChristopherDavenport:
I got this feedback recently could this be integrated, @djspiewak  believe this will also speed up so may cancel out the prefetch costs.

```
.parJoin(maxConcurrency)
          .interruptWhen(terminationSignal)
          .drain)
```
to 
```
          .flatMap(s => Stream.resource(s.compile.resource.drain.background))
          .interruptWhen(terminationSignal)
          .drain)
```

RaasAhsan:
Does that have the semantics we want? It seems like it's going to start the handler for every socket in a separate fiber, lift that resource into stream, then immediately exit which cancels the fiber (since it introduced a scope). I think `resourceWeak` would work but then we are possibly leaking fibers, and we also can't control concurrency. Let me know if I'm wrong about that though

Also I realized that the `prefetch` isn't _that_ bad. It becomes a problem when you have a lot of connection cycling, but it shouldn't affect max RPS all else being equal

RaasAhsan:
Removed the comment I added about `prefetch`, because it really only limits the rate at which connections are accepted, but not max RPS. I'm going to write a special-purpose combinator that mixes the semantics of `parJoin`, `prefetch` and `drain` since they come with a lot of baggage.

ChristopherDavenport (approved):
I think this seems good.

ember-server/src/test/scala/org/http4s/ember/server/EmberServerSuite.scala
rossabaker: I think we could batter it with requests and assert that every request either completed successfully _or_ was refused a connection.  But our CI is suffering enough as it is.

rossabaker (approved):

PR #4183, opened 2021-01-11
Move to Typelevel Vault (and Unique)

rossabaker:
This removes the inlined copies as well.


rossabaker:
I have not done CE3 versions of either project yet.
* I'd rather not do Unique, pending https://github.com/typelevel/cats-effect/issues/1587
* Vault depends on that decision

When we merge this to main, if that's not sorted out yet, we'll just repackage our inhaled versions.  But it'll block a proper 1.0 milestone until we can sort that out.

ChristopherDavenport (approved):

PR #4191, opened 2021-01-12
Fixes #4084 Port jetty to cats effect 3

### LINKED ISSUE #4084: Port jetty to cats-effect-3

ashwinbhaskar:


ashwinbhaskar:
@rossabaker The tests in JettyServerSuite all time out after 30 seconds and result in error. Where did I go wrong in that?

jetty/src/test/scala/org/http4s/server/jetty/Issue454.scala
```diff
@@ -18,15 +18,16 @@ package org.http4s
...
+import cats.effect.std.Dispatcher
 import org.eclipse.jetty.server.{HttpConfiguration, HttpConnectionFactory, Server, ServerConnector}
 import org.eclipse.jetty.servlet.{ServletContextHandler, ServletHolder}
 import org.http4s.dsl.io._
 import org.http4s.servlet.AsyncHttp4sServlet
 import org.http4s.syntax.all._
 
 object Issue454 {
-  implicit val cs: ContextShift[IO] = Http4sSpec.TestContextShift
+  // implicit val cs: ContextShift[IO] = Http4sSpec.TestContextShift
```
rossabaker, line 30: This code doesn't ever run as part of the build.  If we can't figure out a way to automate it, then I think it's ready to be removed.

jetty/src/test/scala/org/http4s/server/jetty/JettyServerSuite.scala
```diff
@@ -52,25 +52,25 @@ class JettyServerSuite extends Http4sSuite {
...
+    ResourceFixture[Server](Dispatcher[IO].flatMap(serverR))
 
   def get(server: Server, path: String): IO[String] =
-    testBlocker.blockOn(
-      IO(
-        Source
-          .fromURL(new URL(s"http://127.0.0.1:${server.address.getPort}$path"))
-          .getLines()
-          .mkString))
+    IO(
```
rossabaker, line 66: This blocks, and unlike most of our blocking calls we've been porting, fs2 isn't going to handle it.

```suggestion
    IO.blocking(
```

jetty/src/test/scala/org/http4s/server/jetty/JettyServerSuite.scala
```diff
@@ -52,25 +52,25 @@ class JettyServerSuite extends Http4sSuite {
...
+      Source
+        .fromURL(new URL(s"http://127.0.0.1:${server.address.getPort}$path"))
+        .getLines()
+        .mkString)
 
   def post(server: Server, path: String, body: String): IO[String] =
-    testBlocker.blockOn(IO {
+    IO {
```
rossabaker, line 73: Same:

```suggestion
    IO.blocking {
```

jetty/src/main/scala/org/http4s/server/jetty/JettyBuilder.scala
```diff
@@ -224,7 +233,8 @@ sealed class JettyBuilder[F[_]] private (
         service = service,
         asyncTimeout = builder.asyncTimeout,
         servletIo = builder.servletIo,
-        serviceErrorHandler = builder.serviceErrorHandler
+        serviceErrorHandler = builder.serviceErrorHandler,
+        dispatcher
```
rossabaker, line 237: Some other backends are creating `dispatcher` in `Resource`, which is a pattern I like.  The challenge would be how you can thread it down to here.

jetty/src/test/scala/org/http4s/server/jetty/Issue454.scala
ashwinbhaskar: removed it

jetty/src/test/scala/org/http4s/server/jetty/JettyServerSuite.scala
ashwinbhaskar: done

jetty/src/test/scala/org/http4s/server/jetty/JettyServerSuite.scala
ashwinbhaskar: done

jetty/src/main/scala/org/http4s/server/jetty/JettyBuilder.scala
ashwinbhaskar: are you suggesting to pass `Resource[IO,Dispatcher[IO]]` instead of `Dispatcher[IO]`

rossabaker: No, I'm saying to let one of the functions in this create it.  In BlazeClientBuilder (yes, that's a client, and this is a server, but... almost the same), we're creating and using a `dispatcher` resource in the `.resource` function.  Can we do that here?

I think if the function in the `Mount` class took a `Dispatcher`, the rest of it would fall into place, and you could remove the `Dispatcher` as a parameter of the builder.

ashwinbhaskar: got it, let me try that

ashwinbhaskar: @rossabaker the resource function takes returns a resource with type parameter `F`. That means we should make a `Dispatcher[F]` in scope. We can change the signature of `def resource: Resource[F, Server]` to `def resource(dispatcher: Dispatcher[F]): Resource[F, Server]`. But this would mean changing the signature on the trait `ServerBuilder`.
If we want to avoid this, then we would have to pass a dispatcher to the builder.

rossabaker: Oh, I was thinking that resource would create a `Dispatcher` resource and flatMap it.  `BlazeClientBuilder` does it like this:

```scala
    for {
      dispatcher <- Dispatcher[F]
      ...
    } yield ...
```

rossabaker:
Are we past the timeout problem now?

jetty/src/main/scala/org/http4s/server/jetty/JettyBuilder.scala
ashwinbhaskar: @rossabaker my bad. You are right. Having an implicit `Async[F]` brings `Dispatcher[F]` in scope

ashwinbhaskar:
> Are we past the timeout problem now?

@rossabaker no, the timeout issue still persists.

rossabaker:
Oh, the module just isn't enabled yet.  I didn't see the failing test and was hoping we were through it.  Do you have any ideas yet or do you need some help?

ashwinbhaskar:
> Oh, the module just isn't enabled yet. I didn't see the failing test and was hoping we were through it. Do you have any ideas yet or do you need some help?

I am still trying to figure out the issue. Any help is welcome:)
I am also skeptical of the changes made in `Issue454.scala`. I think the dispatcher will be shutdown well `AsyncHttp4sServlet` gets a change to use it?

rossabaker:
Oh, I thought we were deleting Issue454 entirely since it isn't a test that gets run.

ashwinbhaskar:
> Oh, I thought we were deleting Issue454 entirely since it isn't a test that gets run.

okay, I understood it as deleting the implicit line 😅 
Will delete it:)

ashwinbhaskar:
Added some log statements. It confirms [this](https://github.com/http4s/http4s/blob/f5c38c1946515dc1aac2cae1039ed86e423cdace/servlet/src/main/scala/org/http4s/servlet/AsyncHttp4sServlet.scala#L71) line is causing the timeout.

rossabaker:
Try `unsafeRunAndForget` instead of `unsafeRunSync`.

ashwinbhaskar:
> Try `unsafeRunAndForget` instead of `unsafeRunSync`.

results in `IOException: Server returned HTTP response code: 503..`

ashwinbhaskar:
@rossabaker The issue was in servlet module. `async_` has signature `k: (Either[Throwable, A] => Unit) => Unit): F[A]` but `Deferred`'s complete was returning an `F[Boolean]`.

servlet/src/main/scala/org/http4s/servlet/AsyncHttp4sServlet.scala
```diff
@@ -82,7 +82,8 @@ class AsyncHttp4sServlet[F[_]](
 
       val timeout =
         F.async_[Response[F]] { cb =>
-          val _ = gate.complete(ctx.addListener(new AsyncTimeoutHandler(cb)))
+          val _ =
+            dispatcher.unsafeRunSync(gate.complete(ctx.addListener(new AsyncTimeoutHandler(cb))))
```
m-sp, line 86: no need to block here you should be able to just do

```
        F.async[Response[F]] { cb =>
          gate.complete(ctx.addListener(new AsyncTimeoutHandler(cb))).as(Option.empty[F[Unit]])
        }
```

rossabaker (approved):
:tada:

PR #4192, opened 2021-01-12
Merge macro moves and vault typelevelification

rossabaker:
This is mostly #4169, #4170, and #4183.  Unique and Vault remain inlined, but are moved to their new home.  We can release a proper Vault after https://github.com/typelevel/cats-effect/issues/1587.


hamnis (approved):
LGTM

PR #4194, opened 2021-01-12
remove blocking unsafeRunSync inside IO

yanns:


hamnis (approved):

tests/src/test/scala/org/http4s/EntityDecoderSuite.scala
```diff
@@ -330,17 +330,16 @@ class EntityDecoderSuite extends Http4sSuite {
...
-    IO.async_[String] { cb =>
+    IO.async[String] { cb =>
       request
         .decodeWith(happyDecoder, strict = false) { s =>
           cb(Right(s))
           IO.pure(Response())
         }
-        .unsafeRunSync()
-      ()
+        .as(Some(IO.unit))
```
m-sp, line 342: shouldn't this return `None`? `async` has an optional cancelation token, and this is not really canceling anything right?

yanns: 👍 
yes you're right.

yanns: https://github.com/http4s/http4s/pull/4194/commits/53e61553a68eb68aff759388503de3ba9459344f

m-sp (approved):

rossabaker:
The failing test is #4213.

PR #4195, opened 2021-01-12
use a different EC for munit

yanns:
To look for blocking code, one can setup the excecution context used by http4s to a minimal number of thread (`Http4sSpec.TestExecutionContext`).
As munit is using the same execution context, it is difficult to find out which code is blocking as munit is always blocking on one thread.
By using a different thread pool for munit, this investigation is much easier.


rossabaker (approved):
This seems reasonable.

PR #4197, opened 2021-01-12
use blocking where necessary to use the right thread pool

yanns:
Those places were found by running the tests with a compute thread pool of 2 threads.
In theory, we should be able to run all the tests on only one thread, but some blocking places are quite difficult to get rid of for now.


m-sp (approved):

rossabaker:
:+1: Thanks!

PR #4201, opened 2021-01-12
migrate content lang

m-sp:


hamnis (approved):

core/src/main/scala/org/http4s/headers/Content-Language.scala
```diff
@@ -18,11 +18,25 @@ package org.http4s
...
-    HttpHeaderParser.CONTENT_LANGUAGE(s)
+    ParseResult.fromParser(parser, "Invalid Content-Language header")(s)
+
+  private[http4s] val parser: Parser[headers.`Content-Language`] = {
+    val languageTag: Parser[LanguageTag] =
+      (Parser.string(Rfc5234.alpha.rep) ~ (Parser.string("-") *> Rfc2616.token).rep0).map {
+        case (main: String, sub: collection.Seq[String]) =>
+          LanguageTag(main, org.http4s.QValue.One, sub.toList)
+      }
+    languageTag.repSep(Rfc2616BasicRules.listSep).map { tags =>
```
rossabaker, line 35: Should this be the `headerRep1` helper?  Anything that's `1#(...)` in the RFC should be using that.  It takes care of some weird conditions with repeated commas.

m-sp: thanks, next time I won't forget about this 🤦

hamnis (approved):
On green

ChristopherDavenport (approved):

PR #4202, opened 2021-01-12
migrate cache control parser

m-sp:


core/src/main/scala/org/http4s/internal/parsing/Rfc7230.scala
```diff
@@ -55,9 +55,12 @@ private[http4s] object Rfc7230 {
   /* quoted-pair    = "\" ( HTAB / SP / VCHAR / obs-text ) */
   val quotedPair: Parser[Char] = char('\\') *> qdPairChar
 
+  private def surroundedBy[A](a: Parser0[A], b: Parser[Any]): Parser[A] =
```
m-sp, line 58: I needed a custom `surroundedBy` because as long as the quotes are required we actually do make progress and can get the `Parser` aka `Parser1`

hamnis (approved):
Looks right to me

core/src/main/scala/org/http4s/internal/parsing/Rfc7230.scala
rossabaker: Oh, is this an oversight in the cats-parse signature?

rossabaker (approved):

core/src/main/scala/org/http4s/internal/parsing/Rfc7230.scala
m-sp: the return type of the original method is a Parser0, so it makes sense that every argument is a Parser0.
so if a or b is a Parser than the return type is a Parser. otherwise the return type is a Parser0. tbh not sure if simple overloading would allow that. but yeah maybe a `surroundedBy0` would be nice

PR #4207, opened 2021-01-14
WIP: Use ip4s in Uri.Host model

rossabaker:
ip4s will imminently be a dependency of fs2-3.0, which means at least 1.0 will have it on the classpath anyway.  It is a much richer model of IP addresses than ours, and it's safer than java.net's.

Some things representable in RFC3986 are not in ip4s.  For example:
* RFC3986 ports are boundless.  ip4s constrains to 65535.
* RFC3986 hotsnames are lenient.  ip4s focuses on domain names.
* RFC3986 has syntax for "future IP" addresses.  ip4s doesn't.

So I don't think we can put a `SocketAddress` directly on `Uri.Authority`.  What we can do is eliminate some of our public type constructors, and create (some) implementations backed by ip4s.  The encoding looks a lot like Circe's `Json`, where there are operations to extract to `Option`, and a general `fold`.  We remain compliant with RFC3986, get a smarter IP representation, and don't duplicate names like `Ipv4Address` with the coming fs2.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -100,6 +100,7 @@ lazy val core = libraryProject("core")
       catsParse.exclude("org.typelevel", "cats-core_2.13"),
       fs2Core,
       fs2Io,
+      ip4sCore,
       log4s,
       scodecBits,
       slf4jApi, // residual dependency from macros
@@ -125,6 +126,7 @@ lazy val laws = libraryProject("laws")
       catsEffectLaws,
       catsLaws,
       disciplineCore,
+      ip4sTestKit,
       scalacheck,
       scalacheckEffectMunit,
       munitCatsEffect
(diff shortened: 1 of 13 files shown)


rossabaker:
Alternative designs:

1. Punt on RFC3986 compliance (weird hostnames, ipvfuture, bizarre ports), and just make a `SocketAddress` part of the `Authority`.  This can't really be done without the `Host` abstraction of ip4s-2.x, but @mpilquist was open to such a version for cats-parse-2.x.
2. Public data constructors wrapping ip4s.  So there'd be an `Uri.Host.Ipv4` wrapper around `ip4s.Ipv4Address`.  Too many similar names floating around, to my taste, but it makes it a more nuts-and-bolts ADT.
3. Public data constructors convertible to ip4s.  There'd be a `Uri.Host.Ipv4` wrapper with a `toIp4s` for people doing real network operations.

I think I like the direction in this PR, or 3.

PR #4208, opened 2021-01-14
port most simple headers to cats parse

m-sp:
well that blew up more than intended 🙈
hope this is still reviewable, let me know if I should split this up into smaller chunks


core/src/main/scala/org/http4s/headers/Host.scala
```diff
@@ -17,14 +17,31 @@
...
   override def parse(s: String): ParseResult[Host] =
-    HttpHeaderParser.HOST(s)
+    ParseResult.fromParser(parser, "Invalid Host")(s)
+
+  //  // Do not accept scoped IPv6 addresses as they should not appear in the Host header,
+  //  // see also https://issues.apache.org/bugzilla/show_bug.cgi?id=35122 (WONTFIX in Apache 2 issue) and
+  //  // https://bugzilla.mozilla.org/show_bug.cgi?id=464162 (FIXED in mozilla)
+  private[http4s] val parser = {
+    val port = Parser.string(":") *> Rfc3986.digit.rep.string.mapFilter { case s =>
+      try Some(s.toInt)
```
hamnis, line 36: Try{ s.toInt }.toOption feels like it reads better.

core/src/main/scala/org/http4s/headers/Host.scala
```diff
@@ -17,14 +17,31 @@
...
+      catch { case _: NumberFormatException => None }
+    }
+
+    val host = Rfc2616.token.orElse(Uri.Host.ipLiteral.string).backtrack <* Rfc2616BasicRules.optWs
```
hamnis, line 40: We have two versions of this token and whitespace rules now.
I  suggest we try to use `Rfc7230.token` and `Rfc7230.ows` where possible

core/src/main/scala/org/http4s/headers/Server.scala
```diff
@@ -17,15 +17,23 @@
...
 
 object Server extends HeaderKey.Internal[Server] with HeaderKey.Singleton {
   def apply(id: ProductId): Server =
     new Server(id, Nil)
 
   override def parse(s: String): ParseResult[Server] =
-    HttpHeaderParser.SERVER(s)
+    ParseResult.fromParser(parser, "invalid Server")(s)
+
+  private[http4s] val parser = {
```
hamnis, line 30: `User-Agent` and Server are very similar, maybe they should share the parser up to the last map?

hamnis:
Well done! You beat me to it =)

core/src/main/scala/org/http4s/headers/Allow.scala
```diff
@@ -17,14 +17,20 @@
...
+import org.http4s.internal.parsing.Rfc7230
 import org.http4s.util.Writer
 
 object Allow extends HeaderKey.Internal[Allow] with HeaderKey.Singleton {
   def apply(ms: Method*): Allow = Allow(ms.toSet)
 
   override def parse(s: String): ParseResult[Allow] =
-    HttpHeaderParser.ALLOW(s)
+    ParseResult.fromParser(parser, "Invalid Allow")(s)
+  private[http4s] val parser = Rfc7230
```
hamnis, line 28: alternative version: `  private[http4s] val parser = Method.parser
    .repSep0(Parser.char(',').surroundedBy(Rfc7230.ows))
    .map(list => Allow(list.toSet))`

m-sp: hmm not sure if this actually equivivalent

https://tools.ietf.org/html/rfc7230#section-7

allows e.g. `"foo , ,bar,charlie   "` as a valid value for the `#element` syntax

the header is just weirdly specified https://tools.ietf.org/html/rfc7231#section-7.4.1
the syntax is described as `Allow = #method` which would mean at least one method, but they immediately contradict the syntax: `An empty Allow field value indicates that the resource allows no methods`

core/src/main/scala/org/http4s/headers/Server.scala
m-sp: moved them to the `AdditionalParsers` if you have a better location in mind please let me know
https://github.com/http4s/http4s/pull/4208/commits/b89f9d6ebfbea75f7c51749d262d6361146b3d44

core/src/main/scala/org/http4s/headers/Host.scala
m-sp: sounds good https://github.com/http4s/http4s/pull/4208/commits/b89f9d6ebfbea75f7c51749d262d6361146b3d44

core/src/main/scala/org/http4s/headers/Host.scala
m-sp: fine for me https://github.com/http4s/http4s/pull/4208/commits/b89f9d6ebfbea75f7c51749d262d6361146b3d44

core/src/main/scala/org/http4s/headers/Allow.scala
hamnis: hm, interesting. maybe we need a test that also adds that edge-case.

core/src/main/scala/org/http4s/ProductIdOrComment.scala
```diff
@@ -32,9 +34,19 @@ final case class ProductId(value: String, version: Option[String] = None)
   }
 }
 
+object ProductId {
+  private[http4s] val parser = (Rfc2616.token ~ (Parser.string("/") *> Rfc2616.token).?).map {
```
hamnis, line 38: replace with Rfc7230.token

core/src/main/scala/org/http4s/headers/Host.scala
```diff
@@ -17,14 +17,30 @@
...
+
+  //  // Do not accept scoped IPv6 addresses as they should not appear in the Host header,
+  //  // see also https://issues.apache.org/bugzilla/show_bug.cgi?id=35122 (WONTFIX in Apache 2 issue) and
+  //  // https://bugzilla.mozilla.org/show_bug.cgi?id=464162 (FIXED in mozilla)
+  private[http4s] val parser = {
+    val port = Parser.string(":") *> Rfc3986.digit.rep.string.mapFilter { s =>
+      Try(s.toInt).toOption
+    }
+
+    val host = Rfc7230.token.orElse(Uri.Host.ipLiteral.string).backtrack <* Rfc7230.ows
```
hamnis, line 39: not sure if this is correct.

Ref comment above, Ipv6 should be disallowed, and this does not include ipv4 ?

hamnis:
``` private[http4s] val parser = {
    val host =
      (Uri.Ipv4Address.parser).orElse(Uri.RegName.parser)
    val port = Parser.string(":") *> Rfc3986.digit.rep.string.mapFilter { s =>
      Try(s.toInt).toOption
    }

    (host ~ port.?).map { case (host, port) =>
      Host(host.value, port)
    }
  }```
Possible implementation

Comment (m-sp) [reply core/src/main/scala/org/http4s/headers/Host.scala:39]: Hmm you are absolutely right. Weird thing is that I'm pretty sure this is than wrong in 0.21

Comment (m-sp) [reply core/src/main/scala/org/http4s/headers/Host.scala]: https://github.com/http4s/http4s/pull/4208/commits/69051b341fba89b474cb27fa12f83755758f458f

File: core/src/main/scala/org/http4s/ProductIdOrComment.scala (hunk 5 above)
Comment (m-sp) [reply core/src/main/scala/org/http4s/ProductIdOrComment.scala]: https://github.com/http4s/http4s/pull/4208/commits/8663fa733c3c613c41ea5b37efacabce4c1e66e8

File: core/src/main/scala/org/http4s/headers/Host.scala [hunk 7]
```diff
@@ -17,14 +17,31 @@
...
 object Host extends HeaderKey.Internal[Host] with HeaderKey.Singleton {
   def apply(host: String, port: Int): Host = apply(host, Some(port))
 
   override def parse(s: String): ParseResult[Host] =
-    HttpHeaderParser.HOST(s)
+    ParseResult.fromParser(parser, "Invalid Host")(s)
+
+  //  // Do not accept scoped IPv6 addresses as they should not appear in the Host header,
+  //  // see also https://issues.apache.org/bugzilla/show_bug.cgi?id=35122 (WONTFIX in Apache 2 issue) and
+  //  // https://bugzilla.mozilla.org/show_bug.cgi?id=464162 (FIXED in mozilla)
```
rossabaker, line 33: This was a rule in RFC2616 (which referred to the RFC2396 URI, and predates the current URI).  It's allowed in RFC7230.  I think those bugzillas issues date back to the older spec.

core/src/main/scala/org/http4s/headers/Host.scala
```diff
@@ -17,14 +17,31 @@
...
+  private[http4s] val parser = {
+    val host =
+      (Uri.Ipv4Address.parser).orElse(Uri.RegName.parser)
+    val port = Parser.string(":") *> Rfc3986.digit.rep.string.mapFilter { s =>
+      Try(s.toInt).toOption
```
rossabaker, line 38: A horrid thing about this spec is that the port can be unbounded.  I've been thinking about this in the URI redesign, as well.  I don't think it's _useful_, but it stands in the way of compliance.

I think this is fine for now, but I'm trying to think of a way out that's both practical and pedantic.

core/src/main/scala/org/http4s/headers/If-Match.scala
```diff
@@ -31,7 +32,14 @@ object `If-Match` extends HeaderKey.Internal[`If-Match`] with HeaderKey.Singleto
     `If-Match`(Some(NonEmptyList.of(first, rest: _*)))
 
   override def parse(s: String): ParseResult[`If-Match`] =
-    HttpHeaderParser.IF_MATCH(s)
+    ParseResult.fromParser(parser, "Invalid If-Match")(s)
```
rossabaker, line 35:
```suggestion
    ParseResult.fromParser(parser, "Invalid If-Match header")(s)
```

core/src/main/scala/org/http4s/headers/If-None-Match.scala
```diff
@@ -41,7 +42,14 @@ object `If-None-Match` extends HeaderKey.Internal[`If-None-Match`] with HeaderKe
     `If-None-Match`(Some(NonEmptyList.of(first, rest: _*)))
 
   override def parse(s: String): ParseResult[`If-None-Match`] =
-    HttpHeaderParser.IF_NONE_MATCH(s)
+    ParseResult.fromParser(parser, "Invalid If-None-Match")(s)
```
rossabaker, line 45:
```suggestion
    ParseResult.fromParser(parser, "Invalid If-None-Match header")(s)
```

core/src/main/scala/org/http4s/headers/Max-Forwards.scala
```diff
@@ -45,6 +45,8 @@ object `Max-Forwards` extends HeaderKey.Internal[`Max-Forwards`] with HeaderKey.
     fromLong(length).fold(throw _, identity)
 
   override def parse(s: String): ParseResult[`Max-Forwards`] =
-    HttpHeaderParser.MAX_FORWARDS(s)
+    ParseResult.fromParser(parser, "invalid Max-Forwards")(s)
```
rossabaker, line 48:
```suggestion
    ParseResult.fromParser(parser, "invalid Max-Forwards header")(s)
```

core/src/main/scala/org/http4s/headers/Server.scala
```diff
@@ -17,15 +17,21 @@
...
+import org.http4s.parser.AdditionalRules
 import org.http4s.util.Writer
 
 object Server extends HeaderKey.Internal[Server] with HeaderKey.Singleton {
   def apply(id: ProductId): Server =
     new Server(id, Nil)
 
   override def parse(s: String): ParseResult[Server] =
-    HttpHeaderParser.SERVER(s)
+    ParseResult.fromParser(parser, "invalid Server")(s)
```
rossabaker, line 28:
```suggestion
    ParseResult.fromParser(parser, "Invalid Server header")(s)
```

core/src/main/scala/org/http4s/headers/User-Agent.scala
```diff
@@ -17,15 +17,21 @@
...
+import org.http4s.parser.AdditionalRules
 import org.http4s.util.Writer
 
 object `User-Agent` extends HeaderKey.Internal[`User-Agent`] with HeaderKey.Singleton {
   def apply(id: ProductId): `User-Agent` =
     new `User-Agent`(id, Nil)
 
   override def parse(s: String): ParseResult[`User-Agent`] =
-    HttpHeaderParser.USER_AGENT(s)
+    ParseResult.fromParser(parser, "Invalid User-Agent")(s)
```
rossabaker, line 28:
```suggestion
    ParseResult.fromParser(parser, "Invalid User-Agent header")(s)
```

core/src/main/scala/org/http4s/headers/X-Forwarded-For.scala
```diff
@@ -18,13 +18,20 @@ package org.http4s
...
+import cats.parse._
 import java.net.InetAddress
-import org.http4s.parser.HttpHeaderParser
+import org.http4s.internal.parsing.Rfc7230
 import org.http4s.util.Writer
 
 object `X-Forwarded-For` extends HeaderKey.Internal[`X-Forwarded-For`] with HeaderKey.Recurring {
   override def parse(s: String): ParseResult[`X-Forwarded-For`] =
-    HttpHeaderParser.X_FORWARDED_FOR(s)
+    ParseResult.fromParser(parser, "Invalid X-Forwarded-For")(s)
```
rossabaker, line 28:
```suggestion
    ParseResult.fromParser(parser, "Invalid X-Forwarded-For header")(s)
```

core/src/main/scala/org/http4s/headers/X-Forwarded-For.scala
```diff
@@ -18,13 +18,20 @@ package org.http4s
...
+  private[http4s] val parser: Parser[`X-Forwarded-For`] =
+    Rfc7230
+      .headerRep1(
+        (Uri.Ipv4Address.parser.map(_.toInet4Address).backtrack | Uri.Ipv6Address.parser.map(
```
rossabaker, line 32: This one can't be a hostname?

core/src/main/scala/org/http4s/internal/parsing/Rfc7230.scala
```diff
@@ -57,11 +57,26 @@ private[http4s] object Rfc7230 {
...
     b *> a <* b
+  private def between[A](a: Parser[Any], b: Parser0[A], c: Parser[Any]): Parser[A] =
+    a *> b <* c
 
   /*quoted-string  = DQUOTE *( qdtext / quoted-pair ) DQUOTE*/
   val quotedString: Parser[String] =
     surroundedBy(qdText.orElse(quotedPair).rep0.string, dquote)
 
+  /* HTAB / SP / %x21-27 / %x2A-5B / %x5D-7E / obs-text */
+  val CText =
```
rossabaker, line 68: Nit: all the other parsers are lowercase names.

core/src/main/scala/org/http4s/parser/AdditionalRules.scala
```diff
@@ -221,4 +221,10 @@ private[http4s] object AdditionalRules {
       case _: NumberFormatException => None
     }
   }
+
+  val ServerAgentParser: P[(ProductId, List[ProductIdOrComment])] = {
```
rossabaker, line 225: Can we get this on the companion of `ProductIdOrComment`?  Putting models in this package gets us circular dependencies.

rossabaker (commented):
Very nice work.  The suggested edits are just me being a consistency fanatic.  A couple bigger things to think about, but this is great overall.

core/src/main/scala/org/http4s/headers/Access-Control-Allow-Credentials.scala
```diff
@@ -17,13 +17,15 @@
...
+import cats.parse.Parser
 import org.http4s.util.Writer
 
 object `Access-Control-Allow-Credentials`
     extends HeaderKey.Internal[`Access-Control-Allow-Credentials`] {
   override def parse(s: String): ParseResult[`Access-Control-Allow-Credentials`] =
-    HttpHeaderParser.ACCESS_CONTROL_ALLOW_CREDENTIALS(s)
+    ParseResult.fromParser(parser, "invalid Access-Control-Allow-Credentials")(s)
+
+  private[http4s] val parser = Parser.string("true").as(`Access-Control-Allow-Credentials`())
```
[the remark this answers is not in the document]
m-sp: yeah that one is weird. seems it also is case sensitive which was unexpected https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Credentials

core/src/main/scala/org/http4s/headers/X-Forwarded-For.scala
m-sp: it's not defined in any spec but according to the mdn it's just ip addresses: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For

core/src/main/scala/org/http4s/headers/Host.scala
m-sp: just to be sure that I under stand RFC7230. we could just delegate to `Uri.Host.parser` now? and remove the comment

core/src/main/scala/org/http4s/parser/AdditionalRules.scala
m-sp: https://github.com/http4s/http4s/pull/4208/commits/c6a368b9b9a64f5744880ec3628ed806e93e3f5a

core/src/main/scala/org/http4s/internal/parsing/Rfc7230.scala
m-sp: https://github.com/http4s/http4s/pull/4208/commits/c6a368b9b9a64f5744880ec3628ed806e93e3f5a

core/src/main/scala/org/http4s/headers/Host.scala
m-sp: https://github.com/http4s/http4s/pull/4208/commits/190d6766dc60f05351db722b8ae03ba4b15765de

hamnis (approved):

core/src/main/scala/org/http4s/headers/Access-Control-Allow-Credentials.scala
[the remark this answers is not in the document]
rossabaker: Yeah, I was about to call that out, and found my way to the same page. :laughing: 

The obvious answer on the model is to make the header a case object, but we use the companion object for the keys.  We could keep a cached value of it.

rossabaker:
Great job!

PR #4209, opened 2021-01-14
Cats parse simple headers

hamnis:


project/Http4sPlugin.scala
```diff
@@ -287,7 +287,7 @@ object Http4sPlugin extends AutoPlugin {
     val asyncHttpClient = "2.12.2"
     val blaze = "0.14.14"
     val boopickle = "1.3.3"
-    val caseInsensitive = "0.3.0"
+    val caseInsensitive = "1.0.0-RC2"
```
hamnis, line 290: whops, accidentally committed this

core/src/main/scala/org/http4s/headers/Allow.scala
```diff
@@ -17,14 +17,19 @@
...
 package headers
 
-import org.http4s.parser.HttpHeaderParser
+import cats.parse.Parser
+import org.http4s.internal.parsing.Rfc7230
 import org.http4s.util.Writer
 
 object Allow extends HeaderKey.Internal[Allow] with HeaderKey.Singleton {
+  private[http4s] val parser = Method.parser
+    .repSep0(Parser.char(',').surroundedBy(Rfc7230.ows))
```
hamnis, line 26: I thought I could use Rfc7230.headerRep1 here but tests started failing

hamnis:
Well now. Apparently @m-sp did this also in #4208  ... There I go for not looking a PRs just before pushing.

PR #4212, opened 2021-01-16
Upgrade to Scala 2.13.4 and 2.12.13

rossabaker:
Includes scalafix-0.9.25.   We were blocked with 0.9.24.


rossabaker:
This version of scalafix is only for Scala 2.12.13, so throw that in, too.

rossabaker:
Blocked by https://github.com/scalameta/mdoc/pull/445.

m-sp:
If I understand the release notes for 2.12 correctly you can now delete https://github.com/http4s/http4s/blob/main/core/src/main/scala-2.12/scala/annotation/nowarn.scala

rossabaker:
It's going to break MiMa if we clean that annotation up here, but we can do it in 0.22.

PR #4213, opened 2021-01-16
Backport munit ports and fix recyclable assertion

### LINKED ISSUE #4196: Dubious test after blaze-client WritePendingException fix
```
==> X org.http4s.client.blaze.Http1ClientStageSuite.Not expect body if request was a HEAD request  0.013s munit.FailException: /home/runner/work/http4s/http4s/blaze-client/src/test/scala/org/http4s/client/blaze/Http1ClientStageSuite.scala:267 assertion failed
266:      // connection reusable immediately after headers read
267:      assert(tail.isRecyclable)
268:
```

This seems like it is a manifestation of the warning in #4143?  I think we can no longer assert its recylability.

RaasAhsan:
Fixes #4196 and also backports some munit tests. Sorry if this causes downstream merge troubles!


blaze-client/src/test/scala/org/http4s/client/blaze/Http1ClientStageSuite.scala
```diff
@@ -0,0 +1,296 @@
...
+    val contentLength = 12345L
+    val resp = s"HTTP/1.1 200 OK\r\nContent-Length: $contentLength\r\n\r\n"
+    val headRequest = FooRequest.withMethod(Method.HEAD)
+
+    val h = new SeqTestHead(List(mkBuffer(resp)))
+    LeafBuilder(tail).base(h)
+
+    tail.runRequest(headRequest, IO.never).flatMap { response =>
+      assertEquals(response.contentLength, Some(contentLength))
+
```
RaasAhsan, line 257: The crucial change is here and might not be obvious in the merges. We deleted the assertion that the connection is immediately `isRecyclable`.

rossabaker (approved):
The earlier in the branch train we can do this work, the less problems it causes over time.  This is exactly the right way to do it.

PR #4214, opened 2021-01-17
Render request URIs in origin form

rossabaker:
Before this, Ember was sending requests in absolute-form.  This is legal, but it's also unusual unless going through a proxy, and not all servers support it. even though they "MUST".

Refactored blaze to use the same mechanism.

The absolute form is also giving an ember-server [a hard time on series/0.22](https://github.com/http4s/http4s/runs/1715619314#step:11:3860).

### DIFF
--- a/blaze-client/src/main/scala/org/http4s/client/blaze/Http1Connection.scala
+++ b/blaze-client/src/main/scala/org/http4s/client/blaze/Http1Connection.scala
@@ -390,10 +390,7 @@ private object Http1Connection {
 
   private def encodeRequestLine[F[_]](req: Request[F], writer: Writer): writer.type = {
     val uri = req.uri
-    writer << req.method << ' ' << uri.copy(
-      scheme = None,
-      authority = None,
-      fragment = None) << ' ' << req.httpVersion << "\r\n"
+    writer << req.method << ' ' << uri.toOriginForm << ' ' << req.httpVersion << "\r\n"
     if (getHttpMinor(req) == 1 && Host
         .from(req.headers)
         .isEmpty) { // need to add the host header for HTTP/1.1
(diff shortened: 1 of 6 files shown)


RaasAhsan:
> The absolute form is also giving an ember-server a hard time on series/0.22.

What's the relationship between that and the test case `org.http4s.ember.server.EmberServerSuite.server responds to requests` failing?

rossabaker:
I knew it was true, and just figured out why:

Using an absolute-form with an empty path results in an empty path info.  The `Root` extractor in series/0.21 matches the empty path in addition to the root path:

```
scala> Path("/") match { case Root => "Yep"; case _ => "Nope" }
val res1: String = Yep

scala> Path("") match { case Root => "Yep"; case _ => "Nope" }
val res2: String = Yep
```

The new Path model in series/0.22, which is unified with the DSL, doesn't conflate Root with empty:

```
scala> Uri.Path.Root match { case Root => "Yep"; case _ => "nope" }
val res3: String = Yep

scala> Uri.Path.empty match { case Root => "Yep"; case _ => "nope" }
val res4: String = nope
```

I think pathInfo will always start with `"/"` except in this corner case where a client violates the spec by sending an absolute-form when not talking to a proxy, exploiting a loophole that the path can be empty, and the server being required to handle such requests . 

We can do what we want in the DSL, but It's a behavior change.  The corrections are ugly: make the DSL root extractor not an alias to `Uri.Path.Root`, but to something that matches both `Uri.Path.Root` and `Uri.Path.empty`.  I'm inclined to leave it as is: services can match Path.empty if they want.

When we merge this, a proper URI will be sent by the ember-client used in the tests, and that failing test should clear up.

rossabaker:
Also, I again am a bad person for merging those branch merges without running the tests.

RaasAhsan (approved):

ChristopherDavenport (approved):

rossabaker:
I am afraid URI translation might surface that Path vs. Empty issue even for compliant clients, but I'll look into that separately.

Issue #4092, opened 2020-12-27
Port tomcat to cats-effect-3


rossabaker:

yanns:
depending on https://github.com/http4s/http4s/issues/4091

yanns:
I can take care of this.

PR #4216, opened 2021-01-17
port tomcat to CE3

yanns:
Fix https://github.com/http4s/http4s/issues/4092

### DIFF
--- a/tomcat/src/main/scala/org/http4s/server/tomcat/TomcatBuilder.scala
+++ b/tomcat/src/main/scala/org/http4s/server/tomcat/TomcatBuilder.scala
@@ -52,8 +52,7 @@ sealed class TomcatBuilder[F[_]] private (
     mounts: Vector[Mount[F]],
     private val serviceErrorHandler: ServiceErrorHandler[F],
     banner: immutable.Seq[String],
-    classloader: Option[ClassLoader],
-    private val dispatcher: Dispatcher[F]
+    classloader: Option[ClassLoader]
 )(implicit protected val F: Async[F])
     extends ServletContainer[F]
     with ServerBuilder[F] {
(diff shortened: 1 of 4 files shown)


tomcat/src/main/scala/org/http4s/server/tomcat/TomcatBuilder.scala
```diff
@@ -49,8 +52,9 @@ sealed class TomcatBuilder[F[_]] private (
     mounts: Vector[Mount[F]],
     private val serviceErrorHandler: ServiceErrorHandler[F],
     banner: immutable.Seq[String],
-    classloader: Option[ClassLoader]
-)(implicit protected val F: ConcurrentEffect[F])
+    classloader: Option[ClassLoader],
+    private val dispatcher: Dispatcher[F]
```
rossabaker, line 56: Customizing the dispatcher is uninteresting, so several of the backends have started provisioning their own in `resource`.  The `JettyBuilder` has an example of this, and how it gets wired through the servlet mounts.

You kind of hinted at that already with the `create` method, but creating a `Resource` of the builder rather than a resource of the server isn't something we've done elsewhere (that I remember).  What do you think of that other pattern?

rossabaker (commented):
Thanks! Consistency question below.

tomcat/src/main/scala/org/http4s/server/tomcat/TomcatBuilder.scala
yanns: That's indeed a much better approach.
Adopted in https://github.com/http4s/http4s/pull/4216/commits/8ca50b737cb9efb453d27ff3d4c6b4898ea1f211 ;)

rossabaker (approved):

PR #4217, opened 2021-01-17
remove unsafeRunSync inside IO

yanns:
Remove the blocking call to `dispatcher.unsafeRunSync` errorHandler that is used inside an `IO`.

The call to `dispatcher.unsafeRunSync` is kept only in the catch block of the `service` method.


rossabaker (approved):
:+1: Thanks!

PR #4218, opened 2021-01-17
port remaining simple headers

m-sp:
could use some feedback here; not sure how much time I have today to finish it up. will be back soon


core/src/main/scala/org/http4s/headers/Last-Event-Id.scala
```diff
@@ -30,5 +30,9 @@ final case class `Last-Event-Id`(id: EventId) extends Header.Parsed {
 
 object `Last-Event-Id` extends HeaderKey.Internal[`Last-Event-Id`] with HeaderKey.Singleton {
   def parse(s: String): ParseResult[`Last-Event-Id`] =
-    HttpHeaderParser.LAST_EVENT_ID(s)
+    ParseResult.fromParser(parser, "Invalid Last-Event-Id header")(s)
+
+  private[http4s] val parser = (Parser.anyChar.rep0).string.map { (id: String) =>
```
m-sp, line 35: this is a one to one translation of the parboiled parser, but it doesn't seem to be particularly useful

core/src/main/scala/org/http4s/headers/Content-Disposition.scala
```diff
@@ -17,14 +17,24 @@
...
 
 object `Content-Disposition`
     extends HeaderKey.Internal[`Content-Disposition`]
     with HeaderKey.Singleton {
   override def parse(s: String): ParseResult[`Content-Disposition`] =
-    HttpHeaderParser.CONTENT_DISPOSITION(s)
+    ParseResult.fromParser(parser, "Invalid Content-Disposition header")(s)
+
+  private[http4s] val parser = {
+    val value = Rfc7230.token | Rfc7230.quotedString
```
m-sp, line 31: `quotedString` only allows ascii strings but there is a test case in the `MultipartParserSpec` which e.g. uses unicode filenames and therefore now fails.
I've started to go down the RFC rabbit hole but ran out of time to come to a conclusion, if anybody has insights here I would very much appreciate it

rossabaker: The standard is repulsive: [RFC8187](https://tools.ietf.org/html/rfc8187).  It specifically cites `Content-Disposition`, which is the only place I've ever seen it in the wild.  I don't think we've ever tried to implement it, but it's probably the right thing to do.

core/src/main/scala/org/http4s/headers/Last-Event-Id.scala
rossabaker: Per the [specification](https://html.spec.whatwg.org/multipage/server-sent-events.html#parsing-an-event-stream), it's anything _but_ a LF or CR.  Also interesting is that the data must be encoded as UTF-8, but it's not clear how this would be encoded into headers.  Smuggled in as obs-text, I guess?

core/src/main/scala/org/http4s/headers/Content-Disposition.scala
m-sp: I was playing around with this a bit it probably would be the correct thing to implement it, it would actually still fail.

the example in the test is `Content-Disposition: form-data; name="http4s很棒"; filename="我老婆太漂亮.txt"` which doesn't follow `charset  "'" [ language ] "'" value-chars` at all.

but more importantly looking into https://tools.ietf.org/html/rfc5987#section-3.2


  

>  However, RFC 2231 does not specify a mandatory-to-implement character
>    set, making it hard for senders to decide which character set to use.
>    Thus, recipients implementing this specification MUST support the
>    character sets "ISO-8859-1" [ISO-8859-1] and "UTF-8" [RFC3629].

doesn't this mean that the `Rfc2231.token` should parse utf-8 characters?

core/src/main/scala/org/http4s/headers/Last-Event-Id.scala
m-sp: 7a56bcdf6e2

core/src/main/scala/org/http4s/headers/Content-Disposition.scala
hamnis: as i read this, we have two params here, `filename="token" and filename*="utf-8-token"` If we accept utf-8 tokens in the default filename="" we are actually breaking the spec.

https://tools.ietf.org/html/rfc6266#section-4.1

m-sp: well it compiles and the tests work but it's not the prettiest parser ;)

hamnis (approved):
Great work!

rossabaker:
Code looks great, but we need this on series/0.22.  I anticipate it will be an easy cherry-pick.  I'll work on it tonight unless someone rebases it.

m-sp:
So sorry about that can do the rebase in a couple if minutes

m-sp:
new pull request is at #4235

PR #4219, opened 2021-01-18
Merge #4214, #4199 to series/0.22

rossabaker:
Since I broke it last time, PRing like an adult this time.


core/src/main/scala/org/http4s/Uri.scala
```diff
@@ -162,6 +162,12 @@ final case class Uri(
 
   private def toSegment(path: Uri.Path, newSegment: String): Uri.Path =
     path / Uri.Path.Segment(newSegment)
+
+  /** Converts this request to origin-form, which is the absolute path and optional
+    * query.  If the path is relative, it is assumed to be relative to the root.
```
rossabaker, line 167: Maybe we should make this an `Option`, and reject relative paths.  A client request with a relative path is definitely wrong!  It's a slight annoyance in blaze-client and ember-client if we do.

I'd have to look up what we're doing if a client request has no Host.  That's also invalid per HTTP/1.1.

rossabaker:
Going to merge this to unbreak the build, but I'm still unhappy about making the relative URI absolute.

PR #4224, opened 2021-01-19
Ember req host test

rossabaker:
Added this test while mulling over what to do with a request that has a relative URI.  It's not that exciting, but I didn't want to throw it away.


aeons:
Does this have the same test twice?

rossabaker:
Not anymore! :flushed:

Issue #4093, opened 2020-12-27
Port examples to cats-effect-3


rossabaker:

yanns:
I can work on that.
The ember part is depending on https://github.com/http4s/http4s/issues/4086 I guess

rossabaker:
I'm in favor of massive reform here.  We tried to DRY up the services, which is great for testing behavior of various types of routes across backends, but absolutely horrible for someone trying to find a self-contained example.  I'd say if anything is difficult, don't hesitate to either refactor it or just delete it, and we'll build back better.

Issue #4085, opened 2020-12-27
Port ember-client to cats-effect-3


rossabaker:

m-sp:
I started working on this https://github.com/http4s/http4s/pull/4128 but this is blocked by https://github.com/ChristopherDavenport/log4cats not being ported to ce3 yet

PR #4226, opened 2021-01-19
migration examples to ce3

yanns:
Partially fix https://github.com/http4s/http4s/issues/4093

Ember examples are missing as ember is not migrated to ce3 yet (https://github.com/http4s/http4s/issues/4085 https://github.com/http4s/http4s/issues/4086)

I could not start the blaze examples:
```
[info] running (fork) com.example.http4s.blaze.ClientExample
[info] -Xbootclasspath/p is no longer a supported option.
[error] Error: Could not create the Java Virtual Machine.
[error] Error: A fatal exception has occurred. Program will exit.
```
I don't know if this is a known issue.


yanns:
Still an overseen warning to fix:
```
[error] /home/runner/work/http4s/http4s/examples/blaze/src/main/scala/com/example/http4s/blaze/BlazeWebSocketExample.scala:72:9: object Queue in package concurrent is deprecated (since 3.0.0): Use cats.effect.std.Queue instead
[error]         Queue
[error]         ^
[error] one error found
```

rossabaker:
That parameter is for ALPN support, which isn't necessary after Java 8 and probably should be filtered accordingly.  But we test on newer JDK, so I'm not sure why you'd be hitting that.  Which JDK did you hit that error on?

If you disable AlpnBootPlugin, that should go away... but so will HTTP/2 support.

rossabaker (approved):
:+1: for incremental progress.  I don't mind splitting this across PRs.

yanns:
> That parameter is for ALPN support, which isn't necessary after Java 8 and probably should be filtered accordingly. But we test on newer JDK, so I'm not sure why you'd be hitting that. Which JDK did you hit that error on?

```
$ javac -version
javac 15.0.1

$ java -version
openjdk version "15.0.1" 2020-10-20
OpenJDK Runtime Environment AdoptOpenJDK (build 15.0.1+9)
OpenJDK 64-Bit Server VM AdoptOpenJDK (build 15.0.1+9, mixed mode, sharing)
```


> If you disable AlpnBootPlugin, that should go away... but so will HTTP/2 support.

Thanks, I've tried that, and the examples are working:
```
 lazy val examplesBlaze = exampleProject("examples-blaze")
-  .enablePlugins(AlpnBootPlugin)
   .settings(Revolver.settings)
```

rossabaker:
Oh, and it's in `run / javaOptions`, which is why CI doesn't catch it.  We can fix it in the org plugin, and it'll work here, for upstream blaze, and for anything else that needs it for HTTP/2 support.

Issue #4232, opened 2021-01-20
Port to Typelevel log4cats


rossabaker:
This one should be mostly a search and replace, with a tiny bit of SBT work.

1. On series/0.22, update the dependency to `1.2.0-RC2`, published under `org.typelevel`.  This is all in Http4sPlugin.scala.
2. Change the imports from `io.chrisdavenport.log4cats` to `org.typelevel.log4cats`

cquiroz:
I can do this one

PR #4234, opened 2021-01-21
Switch to org.typelevel.log4cats

cquiroz:
Brought to you thanks to vim. Fixes https://github.com/http4s/http4s/issues/4232

It seems the build is flaky but this at least compiles


rossabaker (changes requested):
We only accept changes made by Emacs.

rossabaker:
On a more serious note, this looks good.

rossabaker:
Actually, let's sit on this until we get log4cats for CE3 out (I have a PR in progress), or else we won't easily be able to merge to main. But it's ready immediately after that.

rossabaker:
Okay, 2.0.0-M1 is out, so there's a path to merging to main. Let's do it.

PR #4145, opened 2021-01-06
core/laws/testing on scala 3.0.0-M2

hamnis:

PR #4241, opened 2021-01-23
Fix ipv6 parser

m-sp:
should hopefully fix https://github.com/http4s/http4s/pull/4145#pullrequestreview-574847814

but the following test fails:

```

    "only parse part of an invalid ipv6 address where a single section is shortened (must be 2 or more)" in {
      val invalidIp = "2001:db8::1:1:1:1:1"
      Ipv6Address.fromString(invalidIp).map(_.value) must beLeft
    }
```

but I don't really see why the ip address is invalid

### DIFF
--- a/core/src/main/scala/org/http4s/Uri.scala
+++ b/core/src/main/scala/org/http4s/Uri.scala
@@ -1169,12 +1169,6 @@ object Uri {
             Ipv6Address(a, b, c, d, e, f, g, h)
         }
 
-      def repN[A](n: Int, p: cats.parse.Parser0[A], sep: cats.parse.Parser0[Unit] = P.unit): cats.parse.Parser0[List[A]] =
-        ((p ~ sep).replicateA(n - 1) ~ p)
-          .map { case (head, tail) => head.map(_._1) :+ tail }
-          .backtrack
-          .orElse(if (n == 1) p.map(List(_)).backtrack else repN(n - 1, p, sep).backtrack)
-
       val h16: P[Short] =
         (hexdig ~ hexdig.? ~ hexdig.? ~ hexdig.?).string.map { (s: String) =>
           java.lang.Integer.parseInt(s, 16).toShort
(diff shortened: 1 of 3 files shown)


core/src/main/scala/org/http4s/Uri.scala
```diff
@@ -1192,39 +1186,45 @@ object Uri {
...
+        .orElse(((h16.repSep0(0, 2, colon).with1 <* doubleColon) ~ h16Colon.repExactlyAs[List[Short]](3).backtrack ~ ls32)
+            .map { case ((ls: List[Short], rs), rs2) =>
+              toIpv6(ls, rs ++ rs2)
             })
         .backtrack
-        .orElse(((repN(3, h16, colon.void).?.with1 <* doubleColon) ~ h16Colon ~ ls32)
-          .map { case ((ls: Option[List[Short]], r0: Short), rs) =>
-            toIpv6(ls.getOrElse(Seq.empty), Seq(r0) ++ rs)
-          })
+        .orElse(((h16.repSep0(0, 3, colon).with1 <* doubleColon) ~ h16Colon.repExactlyAs[List[Short]](2).backtrack ~ ls32)
```
rossabaker, line 1204: Why (0, 3) instead of (3, 3)?

rossabaker (commented):
The failing test does parse in non-http4s decoders.

core/src/main/scala/org/http4s/Uri.scala
m-sp: it would fail on e.g. `53f6::1:551b:7bd8`

the number of segments after `::` are always fixed to a number n. so before the `::` we can parse up to 8 - n segments

rossabaker: Ah, right. So this is part of the bug fix.

rossabaker:
Oh, that address was invalid because a `::` is supposed to represent multiple groups of zeroes.  Strictly, the test was correct, but I don't think it ambiguates anything to accept it.

m-sp:
> Oh, that address was invalid because a `::` is supposed to represent multiple groups of zeroes. Strictly, the test was correct, but I don't think it ambiguates anything to accept it.

I have an alternative fix here https://github.com/http4s/http4s/pull/4243 which also handles that case and IMO is a bit easier to understand. let me know which one you prefer and I can resolve the conflicts afterwards

PR #4242, opened 2021-01-23
Adopt ip4s

rossabaker:
fs2-3.x is taking on an ip4s dependency, so it will be on our classpath anyway in 1.0.  It has a richer model of addresses and is a lot harder to accidentally trigger network lookups, so it's good for us to use.

In the spirit of making 0.22 look as much like 1.0 as possible without cats-effect, we begin the adoption here.


rossabaker:
Not being as cute as we were with folds in #4207.  This will supersede it.

blaze-server/src/main/scala/org/http4s/server/blaze/BlazeServerBuilder.scala
```diff
@@ -259,8 +260,12 @@ class BlazeServerBuilder[F[_]](
...
                 Request.Keys.ConnectionInfo,
                 Request.Connection(
-                  local = local,
-                  remote = remote,
+                  local = SocketAddress(
+                    IpAddress.fromBytes(local.getAddress.getAddress).get,
+                    Port(local.getPort).get),
+                  remote = SocketAddress(
+                    IpAddress.fromBytes(remote.getAddress.getAddress).get,
+                    Port(remote.getPort).get),
```
rossabaker, line 268: This is inartful. And maybe wasteful if unreferenced!

core/src/main/scala/org/http4s/Message.scala
```diff
@@ -373,38 +373,39 @@ final class Request[F[_]](
...
-      .fold(remote.flatMap(remote => Option(remote.getAddress)))(_.values.head)
+      .fold(remote.map(_.ip))(_.values.head)
 
-  def remoteAddr: Option[String] = remote.map(_.getHostString)
+  def remoteAddr: Option[IpAddress] = remote.map(_.ip)
 
-  def remoteHost: Option[String] = remote.map(_.getHostName)
+  def remoteHost(implicit F: Sync[F]): F[Option[Hostname]] = {
+    val inetAddress = remote.map(_.ip.toInetAddress)
+    F.delay(inetAddress.map(_.getHostName)).map(_.flatMap(Hostname(_)))
```
rossabaker, line 389: I think this should have been getHostAddress all along, and then it wouldn't need the `F`?  Why are we doing reverse lookups here?

servlet/src/main/scala/org/http4s/servlet/Http4sServlet.scala
```diff
@@ -102,8 +102,8 @@ abstract class Http4sServlet[F[_]](service: HttpApp[F], servletIo: ServletIo[F])
         .insert(
           Request.Keys.ConnectionInfo,
           Request.Connection(
-            local = InetSocketAddress.createUnresolved(req.getLocalAddr, req.getLocalPort),
-            remote = InetSocketAddress.createUnresolved(req.getRemoteAddr, req.getRemotePort),
+            local = SocketAddress(IpAddress(req.getLocalAddr).get, Port(req.getLocalPort).get),
+            remote = SocketAddress(IpAddress(req.getRemoteAddr).get, Port(req.getRemotePort).get),
```
rossabaker, line 106: This is also inartful.

server/src/test/scala/org/http4s/server/middleware/ErrorActionSuite.scala
```diff
@@ -18,16 +18,15 @@ package org.http4s.server.middleware
...
-import org.http4s.syntax.all._
 import org.http4s.dsl.io._
+import org.http4s.syntax.all._
 import org.typelevel.vault.Vault
 
-import java.net.{InetAddress, InetSocketAddress}
-
 class ErrorActionSuite extends Http4sSuite {
-  val remote = "192.168.0.1"
+  val remote = IpAddress("192.168.0.1").get
```
rossabaker, line 29: We're running up against a problem here where our `ipv4` and `ipv6` interpolators conflict with ip4s', and ip4s' are imported at the root level, so they're almost always going to win.  I think we should retire ours.  Most of what @hamnis just did for ours could continue on: just remove the conflicting ones.

rossabaker (commented):
I'm surprised to not hit connection info problems in the Ember backend.

rossabaker:
An ip4s 1.4.2 release, with its `fromInetAddress` methods, would clean up a lot of `.get`.

ChristopherDavenport (approved):

hamnis (approved):

PR #4243, opened 2021-01-23
Alternative fix for ipv6

m-sp:


rossabaker (approved):
I find this one easier to understand.

rossabaker:
`UriParserSpec` is now failing. I haven't investigated why yet.

hamnis:
https://github.com/http4s/http4s/runs/1755280713#step:11:2051

m-sp:
it was the generator after all. it generates an address which shortens only a single section... exactly what is forbidden in the Rfc3986ParserSpec

rossabaker:
Nice find!  I'll bet that last commit is something we could fix in series/0.21, for whatever it's worth.  (We haven't done much to publicize our arbitraries for outside use.)

rossabaker:
I'll look into cherry-picking that commit next time I'm back on series/0.21.

PR #4244, opened 2021-01-23
Ember stream forking

RaasAhsan:
Follow up to #4179. I introduced a `prefetch` call which is a hack to get the outer stream to terminate and finalize before inner streams do, by buffering streams to a new stream. `parJoin` also shuffles the results of the inner stream back out to the new stream which isn't strictly necessary.

This PR introduces a `forking` combinator which behaves identically to `parJoin` except the outer stream can terminate and finalize before the inner stream, and there is no queueing involved.

TODO:
- [x] Concurrency control
- [ ] Possibly upstream this back into fs2, since it's useful for network servers in general


ember-server/src/main/scala/org/http4s/ember/server/internal/ServerHelpers.scala
```diff
@@ -151,18 +152,78 @@ private[server] object ServerHelpers {
...
+          inner
+            .interruptWhen(stopSignal)
+            .compile
+            .drain
+            .attempt
+            .flatMap(handleResult) >> decrementRunning
+
+      val runOuter: F[Unit] =
+        streams
+          .evalMap(inner => F.start(runInner(inner)))
```
RaasAhsan, line 209: `parJoin` messes around with scope leases a bit, which is what I'm assuming prevents finalizers for the outer stream to be run until after all inner streams close.

ember-server/src/main/scala/org/http4s/ember/server/internal/ServerHelpers.scala
```diff
@@ -151,18 +152,78 @@ private[server] object ServerHelpers {
...
-      .parJoin(maxConcurrency)
-      .drain
+      
+    forking(handler)
+  }
+
+  // Similar to parJoin with a few semantic differences:
+  // - Inner stream values are not emitted to the new stream
+  // - The outer stream can terminate and finalize early
+  def forking[F[_], O](streams: Stream[F, Stream[F, O]])(implicit F: Concurrent[F]): Stream[F, INothing] = {
```
RaasAhsan, line 170: Super similar to https://github.com/typelevel/fs2/blob/main/core/shared/src/main/scala/fs2/Stream.scala except I've ripped scope leasing and queueing out

ember-server/src/main/scala/org/http4s/ember/server/internal/ServerHelpers.scala
```diff
@@ -151,18 +152,78 @@ private[server] object ServerHelpers {
...
       }
-      .parJoin(maxConcurrency)
-      .drain
+      
+    forking(handler)
+  }
+
+  // Similar to parJoin with a few semantic differences:
+  // - Inner stream values are not emitted to the new stream
+  // - The outer stream can terminate and finalize early
```
RaasAhsan, line 169: This means that any resources bound to the scope of the outer stream should _not_ be used in the inner stream, which makes sense for these network servers, but it's a pretty huge caveat

SystemFw: cool, I was about to comment that this is the reason why `parJoin` behaves like it does (and why it needs to mess around with scope leasing to achieve that behaviour), but you got it already. I think it's worth adding it as a code comment too

ember-server/src/main/scala/org/http4s/ember/server/internal/ServerHelpers.scala
SystemFw: yes, this is what the leasing does

RaasAhsan:
@rossabaker if you encounter any issues merging downstream, definitely ping me to help, but I don't anticipate there being any trouble

ChristopherDavenport (approved):
We can contribute this to fs2?

SystemFw:
> We can contribute this to fs2?

it's extremely dangerous in the general case :)

PR #4250, opened 2021-01-24
Compile client on Dotty

rossabaker:


client/src/main/scala/org/http4s/client/middleware/FollowRedirect.scala
```diff
@@ -98,7 +98,8 @@ object FollowRedirect {
         case Right((resp, dispose)) =>
           (methodForRedirect(req, resp), resp.headers.get(Location)) match {
             case (Some(method), Some(loc)) if redirects < maxRedirects =>
-              val nextReq = nextRequest(req, loc.uri, method, resp.cookies)
+              // Dotty needs the asInstanceOf help.
+              val nextReq = nextRequest(req, (loc.asInstanceOf[Location]).uri, method, resp.cookies)
```
rossabaker, line 102: This is where the `HeaderT` projection starts to raise its ugly head.

rossabaker (commented):
I held back 1334783 from 0.21 because I think they sludge up the code and hoped someone has a better idea.

client/src/main/scala/org/http4s/client/middleware/FollowRedirect.scala
hamnis: indeed. but you can avoid that by assigning 
`resp.headers.get(Location)` to a val

its still not good, but avoids asInstanceOf

rossabaker: Oh, nice. I like that better.

rossabaker:
Most of this is on the target branch already. Will start clean on a new PR.

Issue #4237, opened 2021-01-22
Port to Typelevel keypool


rossabaker:
Like #4232, but for keypool.

PR #4251, opened 2021-01-24
port to typelevel keypool

yanns:
Fix https://github.com/http4s/http4s/issues/4237


rossabaker (approved):
Perfect, and I'm releasing 0.4.0-M1, so we have something to merge to main.

PR #4260, opened 2021-01-24
Link to mdoc and final cleanup of tut references

valencik:
This PR removes the final lingering mentions of `tut` and updates them to link to mdoc.

### DIFF
--- a/docs/src/main/mdoc/client.md
+++ b/docs/src/main/mdoc/client.md
@@ -24,7 +24,7 @@ libraryDependencies ++= Seq(
 )
 ```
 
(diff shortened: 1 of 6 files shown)


rossabaker (approved):
Thank you!

valencik:
Should I change this to target the `series/0.21` branch? It looks like the commits will cleanly apply on them.

rossabaker:
Yes, that would be great!  It's all true back to that version, and it will end up on the website corrected for all currently supported versions.

isomarcte (approved):

Issue #3957, opened 2020-11-28
Fix our @see links to be hyperlinks


rossabaker:
Several of our scaladocs have a `@see` tag that references an RFC.  These are great, but they should be wrapped in `[[` and `]]` to [be links](https://docs.scala-lang.org/overviews/scaladoc/for-library-authors.html#markup).  We can do a step better and give a description of the URL.  

Old:

```
  /* @see https://tools.ietf.org/html/rfc7231#page-65 */
```

New:

```
  /* @see [[https://tools.ietf.org/html/rfc7231#page-65 RFC7321: HTTP-date]] */
```

Please start on series/0.21, so the fixes can propagate to all maintained releases.

PR #4261, opened 2021-01-24
Fix @see links in scaladoc

valencik:
This PR addresses https://github.com/http4s/http4s/issues/3957

I standardized the RFC links to link to the html version and include the section title.


rossabaker (approved):
This is wonderful, and thanks for targeting 0.21!

PR #4266, opened 2021-01-25
Compile http4s-client and http4s-server on 3.0.0-M2

rossabaker:
These are the necessary changes I wish weren't so.

### DIFF
--- a/client/src/main/scala/org/http4s/client/oauth1/oauth1.scala
+++ b/client/src/main/scala/org/http4s/client/oauth1/oauth1.scala
@@ -80,8 +80,7 @@ package object oauth1 {
     for {
       reqParams <- getUserParams(req)
       // Working around lack of withFilter
-      req = reqParams._1
-      params = reqParams._2
+      (req, params) = reqParams
       auth <- genAuthHeader(
         req.method,
         req.uri,
--- a/client/src/main/scala/org/http4s/client/oauth1/oauth1.scala
+++ b/client/src/main/scala/org/http4s/client/oauth1/oauth1.scala
@@ -78,7 +78,10 @@ package object oauth1 {
       verifier: Option[Verifier] = None
   )(implicit F: MonadThrow[F], W: EntityDecoder[F, UrlForm]): F[Request[F]] =
     for {
-      (req, params) <- getUserParams(req)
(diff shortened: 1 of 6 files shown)


rossabaker:
http4s-server doesn't yet include tests, which reveal new pain:

```
[error] -- [E040] Syntax Error: /home/ross/src/http4s/server/src/test/scala/org/http4s/server/ContextRouterSuite.scala:27:68 
[error] 27 |  val numbers = ContextRoutes.of[Unit, IO] { case GET -> Root / "1" as _ =>
[error]    |                                                                    ^^
[error]    |                                     '=>' expected, but identifier found
```

rossabaker:
Thanks to @nafg and @ChristopherDavenport for helping me on the server change.

rossabaker:
Uh-oh, we broke FollowRedirect with this.

core/src/main/scala/org/http4s/DecodeResult.scala
```diff
@@ -30,9 +30,15 @@ object DecodeResult {
   def success[F[_], A](a: A)(implicit F: Applicative[F]): DecodeResult[F, A] =
     EitherT(F.pure(Right(a)))
 
+  def successT[F[_], A](a: A)(implicit F: Applicative[F]): DecodeResult[F, A] =
```
rossabaker, line 33: See #4268.  If we proceed, we should start a deprecation cycle in 0.21 and move these down there.

client/src/main/scala/org/http4s/client/oauth1/oauth1.scala
```diff
@@ -78,7 +78,10 @@ package object oauth1 {
       verifier: Option[Verifier] = None
   )(implicit F: MonadThrow[F], W: EntityDecoder[F, UrlForm]): F[Request[F]] =
     for {
-      (req, params) <- getUserParams(req)
+      reqParams <- getUserParams(req)
+      // Working around lack of withFilter
```
ChristopherDavenport, line 82: `(req, params) = reqParams`

rossabaker: I miss better-monadic-for already.

hamnis (approved):

PR #4273, opened 2021-01-25
Update Prometheus Client To 0.10.0 And Fix Tests

isomarcte:
Prometheus is changing the naming scheme of some of their metrics in order to be compatible with the OpenMetrics standard. Part of this means that all counters will have `_total` or `_created` as a suffix. There is no way to create a counter without a `_total` suffix.

See: https://github.com/prometheus/client_java/pull/615

### DIFF
--- a/project/Http4sPlugin.scala
+++ b/project/Http4sPlugin.scala
@@ -315,7 +315,7 @@ object Http4sPlugin extends AutoPlugin {
     val okio = "2.10.0"
     val okhttp = "4.9.0"
     val playJson = "2.9.2"
-    val prometheusClient = "0.9.0"
+    val prometheusClient = "0.10.0"
     val reactiveStreams = "1.0.3"
     val quasiquotes = "2.1.0"
     val scalacheck = "1.15.2"
--- a/prometheus-metrics/src/test/scala/org/http4s/metrics/prometheus/util.scala
+++ b/prometheus-metrics/src/test/scala/org/http4s/metrics/prometheus/util.scala
@@ -63,7 +63,7 @@ object util {
       case "2xx_responses" =>
         registry
           .getSampleValue(
-            s"${prefix}_request_count",
+            s"${prefix}_request_count_total",
             Array("classifier", "method", "status"),
             Array(classifier, method, "2xx"))
       case "2xx_headers_duration" =>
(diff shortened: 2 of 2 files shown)


isomarcte:
I'm making this a draft because I'm not 100% satisfied with this situation. 

Because of the way that the Prometheus client_java is changing the internal data model, users will silently have some of their exposed metrics changed. For example, with the default http4s metrics, the Metric name for the http4s counter defined here (https://github.com/http4s/http4s/blob/series%2F0.22/prometheus-metrics/src/main/scala/org/http4s/metrics/prometheus/Prometheus.scala#L248) will magically change from `org_http4s_server_request_count` to `org_http4s_server_request_count_total`, no matter what you explicitly set the metric name to (https://github.com/prometheus/client_java/blob/parent-0.10.0/simpleclient_common/src/main/java/io/prometheus/client/exporter/common/TextFormat.java#L70).

This means that if the user is not very closely following the upstream developments in Prometheus, they are going to have their counter related metrics just vanish. 

Because Prometheus's client doesn't offer any way to optin to either the new or old metric naming, I don't see any way to give the user a clean and safe upgrade path. The only thing I can think of is that we re-implement the old behavior from client_java < 0.10.0 and immediately deprecate it pointing the users to the new upstream changes. That wouldn't be _too_ hard, but is definitely not ideal.

Thoughts?

rossabaker:
Good questions.  If the opt-out path is to downgrade, we're taking away that option by proceeding with this upgrade.  And eyeballing the diff again, I think maybe 0.10 is backward compatible enough that people could upgrade the jar to opt in.  But the argument against that is the same as the argument against it in the ticket: at some point, users who aren't paying attention will be cut over, and it might as well be now.  Who are we to delay what the Prometheus experts have decided it's time to do?

I think we just highlight this change in the release notes and people really want to stay on 0.9 can implement metrics ops with some small copy and paste and if anyone wants to make an http4s-prometheus_0.9 nanolibrary to share, then so be it?

hamnis:
should this be retargeted to 0.22 ? 
I also agree that this needs a breaking flag.

rossabaker:
Yes, 0.22 is the right target for this.

isomarcte:
@rossabaker @hamnis gah! I meant to target 0.22. I'm never going to get that right....

isomarcte:
Okay, making non-draft.

hamnis (approved):

PR #4275, opened 2021-01-26
migrate CSRFSuite to munit

yanns:


rossabaker:
Thanks! With one import change, I brought it all the way back to 0.21 so it can start doing good there, too.

PR #4277, opened 2021-01-26
migrate MultipartParserSpec to munit

yanns:
no more `unsafeRunSync()` ;)


rossabaker:
Looks good, but would it look much different on series/0.21 or series/0.22?  The earlier we can start these, the less merge liability we have in the future.  I hope series/0.21 winds down soon, but I think it'll still be a while before the world fully runs on cats-effect-3.

yanns:
@rossabaker I've rebased on `series/0.22`.

rossabaker:
This one compiles easily on 0.21, but we've had an code change for Unicode filenames that these tests fail without.  If we ever backport that fix, we can cherry-pick this test at the same time.  This one is good for 0.22, I think.

PR #4278, opened 2021-01-26
migrate client tests to munit

m-sp:
targeting main otherwise there would probably be some nasty CE3 conflicts


rossabaker (commented):
Thanks!

@cquiroz Has been originating these on series/0.21.  The merges are not fun now, but putting them off is an ongoing merge liability in subsequent changes.  @yanns has also been working hard on the tests, though more in CE3.  They might be able to better gauge how hard that is.  I'll defer to them on how worthwhile it is to try to go back vs. just starting here.

m-sp:
retargeted to 0.21

m-sp:
question is if I should avoid the file rename? could make the merge a bit easier

rossabaker:
I'd rather try to rename.  Let me see what happens when we merge this, and then I'll tell you whether I still think so!

rossabaker:
1. Files that were still close enough that it recognized as renames, I was able to merge easily.  
2. Files that changed too much to be understood as renames, sucked, but those would have sucked anyway.
3. I got some add-add conflicts that suggested to my sleepy brain that a few of these were already migrated.  I generally kept some aspects of each.

Anyway, I'd say don't fear the renames.

PR #4281, opened 2021-01-27
Ember startup backpressure

### LINKED ISSUE #4280: No backpressure on Ember Server startup sequence
The issue can basically be summarized with the following test: https://github.com/http4s/http4s/blob/64773a0f29ba6361b150817e4abe30ac8e6155eb/ember-server/src/test/scala/org/http4s/ember/server/EmberServerSuite.scala#L58

In other words, the server may not be immediately available for connections even though we're in the `use` block. The reason is that there is no backpressure between the server starting up and the resource being allocated. This has two effects, one of which was already described. The other effect is that errors encountered in the startup sequence are effectively swallowed.

RaasAhsan:
Fixes #4280

### DIFF
--- a/ember-server/src/main/scala/org/http4s/ember/server/internal/ServerHelpers.scala
+++ b/ember-server/src/main/scala/org/http4s/ember/server/internal/ServerHelpers.scala
@@ -153,25 +153,26 @@ private[server] object ServerHelpers {
         }
         .drain
 
-    Stream.resource(sg.serverResource[F](bindAddress, additionalSocketOptions = additionalSocketOptions))
+    Stream
+      .resource(
+        sg.serverResource[F](bindAddress, additionalSocketOptions = additionalSocketOptions))
       .attempt
-      .flatMap {
-        case Right((_, sockets)) =>
-          Stream.eval(ready.complete(Right(()))) >>
-            sockets
-              .interruptWhen(shutdown.signal.attempt)
-              // Divorce the scopes of the server stream and handler streams so the
-              // former can be terminated while handlers complete.
-              .prefetch
-              .map { connect =>
-                shutdown.trackConnection >>
-                  Stream
-                    .resource(connect.flatMap(upgradeSocket(_, tlsInfoOpt)))
-                    .flatMap(withUpgradedSocket(_))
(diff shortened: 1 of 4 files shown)


ember-server/src/main/scala/org/http4s/ember/server/internal/ServerHelpers.scala
```diff
@@ -151,18 +153,26 @@ private[server] object ServerHelpers {
...
-      .prefetch
-      .map { connect =>
-        shutdown.trackConnection >>
-          Stream
-            .resource(connect.flatMap(upgradeSocket(_, tlsInfoOpt)))
-            .flatMap(withUpgradedSocket(_))
+    Stream
+      .resource(
+        sg.serverResource[F](bindAddress, additionalSocketOptions = additionalSocketOptions))
+      .attempt
```
RaasAhsan, line 159: I added a combinator called `onComplete` to Cats, which is basically attempt + evalTap + rethrow but I don't think it's come downstream yet. If it has, FS2 also has an `onComplete` so it's probably getting shadowed

ChristopherDavenport (approved):

ChristopherDavenport:
Oh, looks like build is unhappy. This makes a lot of sense to me.

ember-server/src/test/scala/org/http4s/ember/server/EmberServerSuite.scala
```diff
@@ -56,9 +57,13 @@ class EmberServerSuite extends Http4sSuite {
...
 
   fixture.test("server responds to requests") { case (server, client) =>
-    IO.sleep(3.seconds) >> client
+    client
       .get(s"http://${server.address.getHostName}:${server.address.getPort}")(_.status.pure[IO])
       .timeout(5.seconds)
       .assertEquals(Status.Ok)
   }
+
+//  server.test("server startup fails if address is already in use") { case _ =>
```
RaasAhsan, line 66: I'm leaving this commented because sometimes the new server can't bind to the same port, even though the server has already finalized. Opening a follow up issue

rossabaker (approved):

PR #4286, opened 2021-01-28
chipping away at the perf regression

m-sp:
use oneof instead of cascading, memoize uri parser construction

### DIFF
--- a/core/src/main/scala/org/http4s/Uri.scala
+++ b/core/src/main/scala/org/http4s/Uri.scala
@@ -168,10 +168,11 @@ final case class Uri(
 
 object Uri extends UriPlatform {
 
+  private val uriRefUtf8 = uriReference(StandardCharsets.UTF_8)
+
   /** Decodes the String to a [[Uri]] using the RFC 3986 uri decoding specification */
   def fromString(s: String): ParseResult[Uri] =
-    ParseResult.fromParser(utf8UriParser, "Invalid URI")(s)
-  private val utf8UriParser = uriReference(StandardCharsets.UTF_8)
+    ParseResult.fromParser(uriRefUtf8, "Invalid URI")(s)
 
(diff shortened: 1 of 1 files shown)


rossabaker (approved):
Has this been identified as a hotspot, or is it speculative?  Either way, I agree with it.

core/src/main/scala/org/http4s/Uri.scala
```diff
@@ -255,6 +255,7 @@ object Uri extends UriPlatform {
     }
   }
 
+  private[http4s] val uriReferenceUtf8: Parser0[Uri] = uriReference(StandardCharsets.UTF_8)
```
rossabaker, line 258: I bet making this lazy unhangs it.  Otherwise, catch it hanging locally and jstack it and you can figure out the circular dependency.

rossabaker (changes requested):
Nope, it hung a test, but I think I see why.

m-sp:
> Has this been identified as a hotspot, or is it speculative?  Either way, I agree with it.

It immediately showed up as a hotspot in jvisualvm. Turns put recreating parsers is expensive

core/src/main/scala/org/http4s/Uri.scala
m-sp: I guess we could get away with only `val`s but that would mean reordering all the parsers 😞

m-sp: https://github.com/http4s/http4s/pull/4286/commits/bee8655796c02cd90772bc60797bb57ddc2d5cfa

rossabaker:
Okay, that's great to learn.  I wonder if we have any more def parsers out there.

m-sp:
> Okay, that's great to learn.  I wonder if we have any more def parsers out there.

A quick search didn't reveal any but my grep foo isn't very strong ;)

PR #4287, opened 2021-01-28
Give User Access To Response to Parsing/Timeout Failures

ChristopherDavenport:


rossabaker (commented):
Is this just for binary compatibility?  Is there any reason not to have a `Throwable => F[Response[F]]`?  Is the layering to protect against `errorHandler` itself failing?  I think if that fails, we can give a hardcoded 500.

ember-server/src/main/scala/org/http4s/ember/server/EmberServerBuilder.scala
```diff
@@ -213,9 +256,23 @@ object EmberServerBuilder {
...
+    // Effectful Handler - Perhaps a Logger
+    // Will only arrive at this code if your HttpApp fails or the request receiving fails for some reason
+    def errorHandler[F[_]: Applicative]: PartialFunction[Throwable, F[Response[F]]] = {
+      case (_: Throwable) =>
+        serverFailure.covary[F].pure[F]
+    }
+
+    @deprecated(
+      "0.21.17",
+      "Use errorOfLastResort for end of the line, or error handler for expected effectful handling")
```
rossabaker, line 271: `errorOfLastResort` is gone now, right?  I think we just want to point people to `errorHandler` entirely?

ChristopherDavenport: Yep. That's gone. Will fix.

rossabaker (approved):

Issue #4131, opened 2021-01-03
Revisit Logger constraints in Cats-Effect 3


rossabaker:
In eca20a5db4b, we had a clash between `Sync` and `Concurrent` constraints.  I ignorantly promoted them all to `Async` to make things compile.  Maybe there was a better way.

rossabaker:
I think this is related to @ChristopherDavenport's concern about no longer being able to run Natchez with Ember.

rossabaker:
This was a giant merge.  I don't even know what module I was talking about.  Probably Ember, because log4s doesn't affect constraints?

m-sp:
yes this was about ember but not sure if this is really fixable, would be good to get @ChristopherDavenport feedback here

bplommer:
Is this still relevant? I think going up to `Async` is the Blessed way to get both `Sync` and `Concurrent`.

armanbilge:
Arguably, yes. I'd argue the Truly Blessed™ way is to ditch `Sync` and ask for a `LoggerFactory` instead.

PR #4288, opened 2021-01-28
Turn Date Header Into Cached Component

ChristopherDavenport:
Steals the Blaze Caching Implementation and does it with a Raw Date Header rather than a raw string

BlazeImpl - https://github.com/http4s/blaze/blob/43d9b8f379f6a54ca1a494ef555d77b03a17609c/http/src/main/scala/org/http4s/blaze/http/util/HeaderTools.scala#L29

I'm not really sure how much time this saves us over before.

### DIFF
--- a/ember-server/src/main/scala/org/http4s/ember/server/internal/DateCaching.scala
+++ b/ember-server/src/main/scala/org/http4s/ember/server/internal/DateCaching.scala
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2019 http4s.org
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package org.http4s.ember.server.internal
 
 import java.time.{Instant, ZoneId}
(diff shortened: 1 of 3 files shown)


ChristopherDavenport:
I'm not really familiar with volatiles as a functional programmer. Seems easy enough with a Ref and the Date header itself with HttpDate.current. But I'm willing to learn

ember-server/src/main/scala/org/http4s/ember/server/internal/DateCaching.scala
```diff
@@ -0,0 +1,60 @@
...
+  def impl[F[_]: Sync]: F[DateCaching[F]] = Sync[F].delay {
+    new DateCaching[F] {
+      // STOLEN From Blaze Is This good?
+      @volatile
+      var dateTime = CachedDateHeader(0L, Raw(date, dateFormat.format(Instant.now())))
+
+      def getDate: F[Header] = Sync[F].delay {
+        val cached = dateTime
+        val current = System.currentTimeMillis()
+        if (current - cached.acquired <= 1000) cached.header
```
rossabaker, line 51: If you get this at 00:00:00.999, in one millisecond, this is going to be wrong for the next 999.

rossabaker: Dividing both by 1000 would get truncation working in your favor.  Or divide acquired by 1000 and immediately multiply by 1000, so you don't need to divide every `current`.

rossabaker (approved):

ember-server/src/main/scala/org/http4s/ember/server/internal/DateCaching.scala
```diff
@@ -0,0 +1,60 @@
...
+
+  private val date = "date".ci
+
+  def impl[F[_]: Sync]: F[DateCaching[F]] = Sync[F].delay {
+    new DateCaching[F] {
+      // STOLEN From Blaze Is This good?
+      @volatile
+      var dateTime = CachedDateHeader(0L, Raw(date, dateFormat.format(Instant.now())))
+
+      def getDate: F[Header] = Sync[F].delay {
```
m-sp, line 48: the delay call is going to make it harder to fix https://github.com/http4s/http4s/issues/4131 and for ember to have any other constraint than `Async`

m-sp (approved):

PR #4290, opened 2021-01-29
Deprecate `as` for `withContext`

rossabaker:
Surprise: https://github.com/lampepfl/dotty/issues/9829.

### DIFF
--- a/docs/src/main/mdoc/auth.md
+++ b/docs/src/main/mdoc/auth.md
@@ -60,7 +60,7 @@ final `HttpRoutes` to be exposed. Notice that we now have access to the user obj
 ```scala mdoc:silent
 val authedRoutes: AuthedRoutes[User, IO] =
   AuthedRoutes.of {
-    case GET -> Root / "welcome" as user => Ok(s"Welcome, ${user.name}")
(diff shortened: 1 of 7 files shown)


hamnis:
not really a fan of `authedAs`. 
`contextAs` might be more general.
or just `context`

rossabaker:
I wondered about `ContextRequest`.  The code is for `AuthedRequest`, so I stuck to something along those lines.  It could be more general, at some loss of readability.

`withContext`?  Additionally `withAuth`?  Readability is good, aliases are bad.

rossabaker:
I don't love it, but want to get unblocked.

m-sp (approved):
mima seems to complain, otherwise lgtm

rossabaker:
Dreary. Should we just roll this out in 0.22? There's no urgent need to purge `as`. It just won't work on Dotty.

rossabaker:
Changing the base branch wasn't enough. Going to rebase.

rossabaker:
Wait, is this [allowed again anyway](https://dotty.epfl.ch/blog/2020/12/18/scala3-m3.html#drop-as-in-patterns)?

rossabaker:
This no longer appears to be necessary as of M3.

PR #4296, opened 2021-01-29
ember-client on 3.0.0-M2 (with Typelevel keypool)

rossabaker:
Includes upgrade to keypool-0.3.0-RC1.

### DIFF
--- a/ember-client/src/main/scala/org/http4s/ember/client/EmberClientBuilder.scala
+++ b/ember-client/src/main/scala/org/http4s/ember/client/EmberClientBuilder.scala
@@ -16,7 +16,7 @@
 
 package org.http4s.ember.client
 
-import io.chrisdavenport.keypool._
+import org.typelevel.keypool._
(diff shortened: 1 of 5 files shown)


ember-client/src/main/scala/org/http4s/ember/client/EmberClientBuilder.scala
```diff
@@ -112,15 +112,18 @@ final class EmberClientBuilder[F[_]: Concurrent: Timer: ContextShift] private (
       builder =
         KeyPoolBuilder
           .apply[F, RequestKey, (RequestKeySocket[F], F[Unit])](
-            (requestKey: RequestKey) =>
+            (requestKey: RequestKey) => {
+              val alloced =
```
rossabaker, line 116: It couldn't infer the `<*` without an intermediate name. :shrug:

ChristopherDavenport (approved):

PR #4299, opened 2021-01-29
Simplify okhttp builder

ashwinbhaskar:


ashwinbhaskar:
@rossabaker can we make the `create` method private? Because we are exposing `Client` as a Resource using the `resource` method.

rossabaker:
Yes, I think that's reasonable.  `create` exists because okhttp manages its own lifecycle, and there were no resources.  Now that there is a resource that the builder manages, having a raw `.create` method is less compelling.

rossabaker (approved):
:+1:

hamnis (approved):

PR #4301, opened 2021-01-29
Better Timeouts, Valid Client Connections

ChristopherDavenport:


ember-client/src/main/scala/org/http4s/ember/client/internal/ClientHelpers.scala
```diff
@@ -101,33 +101,30 @@ private[client] object ClientHelpers {
...
-          readWithTimeout(socket, start, remains, timeoutSignal.get, chunkSize)
-        )
-        _ <- Resource.liftF(timeoutSignal.set(false).void)
-      } yield resp
-
-    def writeRead(req: Request[F]) =
-      timeout match {
-        case t: FiniteDuration => onTimeout(req, requestKeySocket.socket, t)
-        case _ => onNoTimeout(req, requestKeySocket.socket)
+    def writeRead(req: Request[F]) = writeRequestToSocket(req, requestKeySocket.socket, None) >> {
```
RaasAhsan, line 104: Assuming the return type is `F[Response[F]]` here but annotation would be nice

ember-server/src/main/scala/org/http4s/ember/server/internal/ServerHelpers.scala
```diff
@@ -59,30 +59,33 @@ private[server] object ServerHelpers {
...
-                readWithTimeout[F](socket, now, readDuration, timeoutSignal.get, receiveBufferSize)
-              )
-              .flatMap { req =>
-                timeoutSignal.set(false).as(req)
-              })
+      def errorIfEmpty(
+          socket: Socket[F],
+          maxBytes: Int,
+          timeout: Option[FiniteDuration]): Stream[F, Byte] =
+        Stream.eval(socket.read(maxBytes, timeout)).flatMap {
```
RaasAhsan, line 76: Out of curiosity, what happens when the client sends two requests in a row such that they are read in as part of the same chunk?

ember-client/src/main/scala/org/http4s/ember/client/EmberClientBuilder.scala
```diff
@@ -226,7 +216,8 @@ object EmberClientBuilder {
     val acgFixedThreadPoolSize: Int = 100
     val chunkSize: Int = 32 * 1024
     val maxResponseHeaderSize: Int = 4096
-    val timeout: Duration = 60.seconds
+    val idleReadTime = org.http4s.client.defaults.RequestTimeout
```
ChristopherDavenport, line 219: Bringing attention to what these values should be.

RaasAhsan (commented):
Will the `idleTimeout` need to be applied to the response write phase as well?

ember-server/src/main/scala/org/http4s/ember/server/internal/ServerHelpers.scala
ChristopherDavenport: Presently, failure. We can easily get this by moving this stream be passed in rather than created here.

ember-client/src/main/scala/org/http4s/ember/client/internal/ClientHelpers.scala
ChristopherDavenport: `Resource[F, Response[F]]`

ember-server/src/main/scala/org/http4s/ember/server/internal/ServerHelpers.scala
ChristopherDavenport: This has been added in https://github.com/http4s/http4s/pull/4301/commits/5599b528ac3cda381d6506e108d4a6a4bd92a245

ember-client/src/main/scala/org/http4s/ember/client/EmberClientBuilder.scala
```diff
@@ -164,9 +138,22 @@ final class EmberClientBuilder[F[_]: Concurrent: Timer: ContextShift] private (
           .withOnReaperException(_ => Applicative[F].unit)
       pool <- builder.build
     } yield {
-      val client = Client[F](request =>
+      val client = Client[F] { request =>
+        def getValidManaged: Resource[F, Managed[F, (RequestKeySocket[F], F[Unit])]] =
+          pool.take(RequestKey.fromRequest(request)).flatMap { managed =>
+            Resource
+              .liftF(managed.value._1.socket.isOpen)
```
ChristopherDavenport, line 145: Clients were getting dead connections out of the pool. This doesn't fix, but is at least a little help.

ember-server/src/main/scala/org/http4s/ember/server/internal/ServerHelpers.scala
```diff
@@ -131,18 +120,29 @@ private[server] object ServerHelpers {
...
-            .evalMap {
-              case Right((req, resp)) =>
-                postProcessResponse(req, resp).map(resp => (req, resp).asRight[Throwable])
-              case other => other.pure[F]
-            }
-            .evalTap {
-              case Right((request, response)) => send(socket)(Some(request), response)
-              case Left(err) =>
+      Stream
+        .unfoldLoopEval(reachedEndError(socket))(s =>
```
RaasAhsan, line 124: This is super cool

RaasAhsan (approved):

ember-server/src/main/scala/org/http4s/ember/server/internal/ServerHelpers.scala
```diff
@@ -43,6 +43,11 @@ private[server] object ServerHelpers {
   private val serverFailure =
     Response(Status.InternalServerError).putHeaders(org.http4s.headers.`Content-Length`.zero)
 
+  private def durationToFinite(duration: Duration): Option[FiniteDuration] = duration match {
+    case f: FiniteDuration => Some(f)
+    case _ => None
+  }
```
rossabaker, line 49: I think I've seen this twice now.

ember-server/src/main/scala/org/http4s/ember/server/internal/ServerHelpers.scala
```diff
@@ -60,33 +65,12 @@ private[server] object ServerHelpers {
...
-      val action = requestHeaderReceiveTimeout match {
-        case i: FiniteDuration =>
-          Parser.Request.parser(maxHeaderSize, Some(i))(
-            errorIfEmpty(socket, receiveBufferSize, idle))
-        case _ =>
-          Parser.Request.parser(maxHeaderSize, None)(errorIfEmpty(socket, receiveBufferSize, idle))
+
+    def reachedEndError(socket: Socket[F]): Stream[F, Byte] =
+      Stream.eval(socket.read(receiveBufferSize, durationToFinite(idleTimeout))).flatMap {
+        case None => Stream.raiseError(new RuntimeException("Unexpected EOF"))
```
rossabaker, line 71: Can we give this a more informative type than `RuntimeException`?

ember-server/src/main/scala/org/http4s/ember/server/internal/ServerHelpers.scala
ChristopherDavenport: I did I can put it somewhere more common. Ideas?

ember-server/src/main/scala/org/http4s/ember/server/internal/ServerHelpers.scala
ChristopherDavenport: Certainly. Custom or something else in mind?

ember-server/src/main/scala/org/http4s/ember/server/internal/ServerHelpers.scala
rossabaker: Just in `internal/package.scala` maybe? It's not a big deal. It's just four lines.

ember-server/src/main/scala/org/http4s/ember/server/internal/ServerHelpers.scala
rossabaker: Actually, I think there's a standard one in `java.io` that would work great.

ember-client/src/main/scala/org/http4s/ember/client/EmberClientBuilder.scala
```diff
@@ -40,40 +40,12 @@ final class EmberClientBuilder[F[_]: Concurrent: Timer: ContextShift] private (
     private val logger: Logger[F],
     val chunkSize: Int,
     val maxResponseHeaderSize: Int,
+    private val idleConnectionTime: Duration,
```
rossabaker, line 43: We've never been consistent about what's a public val on a builder and what isn't.

ember-client/src/main/scala/org/http4s/ember/client/EmberClientBuilder.scala
```diff
@@ -164,42 +141,59 @@ final class EmberClientBuilder[F[_]: Concurrent: Timer: ContextShift] private (
...
+          pool.take(RequestKey.fromRequest(request)).flatMap { managed =>
+            Resource
+              .liftF(managed.value._1.socket.isOpen)
+              .ifM(
+                managed.pure[Resource[F, *]],
+                // Already Closed,
+                // The Resource Scopes Aren't doing us anything
+                // if we have max removed from pool we will need to revisit
+                Resource.liftF(managed.canBeReused.set(Reusable.DontReuse)) >>
+                  getValidManaged
```
rossabaker, line 155: This risks deadlocking the client, I think?  `managed` is still in scope when we loop and until we're done with whatever new one we get.  This has killed us on the redirect and retry middleware before.  In a loop, you need to release the old before allocating the new.  `Hotswap` gives an elegant CE3 implementation, but CE2 is much trickier.

ember-client/src/main/scala/org/http4s/ember/client/EmberClientBuilder.scala
```diff
@@ -164,42 +141,59 @@ final class EmberClientBuilder[F[_]: Concurrent: Timer: ContextShift] private (
...
+            .liftF(
+              org.http4s.ember.client.internal.ClientHelpers
+                .request[F](
+                  request,
+                  managed.value._1,
+                  managed.canBeReused,
+                  chunkSize,
+                  maxResponseHeaderSize,
+                  idleConnectionTime,
+                  timeout,
```
rossabaker, line 177: I wouldn't mind some named arguments here.  These are both durations or finite durations, right?

ember-core/src/main/scala/org/http4s/ember/core/Parser.scala
```diff
@@ -236,14 +247,16 @@ private[ember] object Parser {
...
-      final case class ParsePreludeError(
-          throwable: Throwable,
+      case class ParsePreludeError(
+          message: String,
+          caused: Option[Throwable],
           method: Option[Method],
           uri: Option[Uri],
           httpVersion: Option[HttpVersion]
-      ) extends Throwable(
+      ) extends Exception(
```
rossabaker, line 256: Why is `message` not part of this constructor?  Would help with the stack trace.

ember-server/src/main/scala/org/http4s/ember/server/EmberServerBuilder.scala
```diff
@@ -156,6 +156,7 @@ final class EmberServerBuilder[F[_]: Concurrent: Timer: ContextShift] private (
       )
       _ <- Resource.make(Applicative[F].unit)(_ => shutdown.await)
       _ <- Resource.liftF(ready.get.rethrow)
+      _ <- Resource.liftF(logger.info(s"Ember-Server service bound to address: $bindAddress"))
```
RaasAhsan, line 159: Blaze printed out a pretty crazy banner too, do we want to do something like that in Ember also?

rossabaker: I vote yes.  I did it as a joke, and people absolutely loved it.  Like, some of the most positive feedback we've ever gotten.

THE PEOPLE DEMAND THEIR FIGLETS.

rossabaker: (Unlike the potential deadlock I fretted about above, I don't want to block this on a figlet.)

ember-client/src/main/scala/org/http4s/ember/client/EmberClientBuilder.scala
rossabaker: Okay, we should only be cycling through connections that are already allocated, unless `pool.take` starts returning a series of instantly closed ones.  We could put a safety valve here, but it's not the Retry bug redux I initially feared.

rossabaker (approved):
A few unresolved nits, but I'm satisfied that this is forward progress in its current state.

ember-client/src/main/scala/org/http4s/ember/client/EmberClientBuilder.scala
ChristopherDavenport: We have not. We'd need to be consistent in our configurations. I just worry that I still have my timeouts wrong and wanted to give myself wiggle room in case I screwed up.

PR #4302, opened 2021-01-30
Upgrade to play-json-2.10.0-RC1

rossabaker:
This one has the Dotty support we'll need: https://github.com/playframework/play-json/issues/539

We can see if it works after https://github.com/typelevel/jawn-fs2/pull/178.


rossabaker:
This is probably going to need to inline the jawn facade because of `withDottyCompat`.

rossabaker:
Let's get this one off the board now.  If I'm right about inlining the facade, we'll do it when we can build Dotty enough to prove it's necessary.

Issue #3945, opened 2020-11-27
Replace our compatibility shims with scala-collection-compat


rossabaker:
We have various compatibility shims between Scala 2.12, Scala 2.13, and coming soon, Scala 3:

* The `@nowarn` annotation in core/src/main/scala-2.12
* The `CollectionCompat` objects in core/src/main/scala-2.*

We should be able to replace all of these with [scala-collection-compat](https://github.com/scala/scala-collection-compat).

We will need to make MiMa exceptions in build.sbt.  If you don't know what those are, we'll explain how and why.

This work should be performed on the series/0.21 branch.

valencik:
I can take this on :)
Working on it over on https://github.com/valencik/http4s/tree/collection-compat

I'll likely have some questions about MiMa in a bit.

rossabaker:
@valencik I'm sorry.  I failed to close this one.  I had a PR that did this, and it got no attention, and we just stuck with a lighter weight solution.

rossabaker:
Ha, the PR came in while I was typing.  Well, let's take a look. :smile:

Sorry for being a bad backlog groomer.

rossabaker:
Decided to go with the lighter approach of #4310.

PR #4304, opened 2021-01-30
WIP Collection compat

valencik:
This PR is incomplete, I'm opening a draft early to get CI runs :)

Tackles https://github.com/http4s/http4s/issues/3945

### DIFF
--- a/core/src/main/scala/org/http4s/Query.scala
+++ b/core/src/main/scala/org/http4s/Query.scala
@@ -21,13 +21,14 @@ import java.nio.charset.StandardCharsets
 import cats.syntax.all._
 import cats.{Eval, Foldable, Hash, Order, Show}
 import org.http4s.Query._
-import org.http4s.internal.CollectionCompat
 
 import org.http4s.internal.parboiled2.{CharPredicate, Parser}
 import org.http4s.parser.{QueryParser, RequestUriParser}
 import org.http4s.util.{Renderable, Writer}
 
 import scala.collection.immutable
+import scala.collection.mutable.ListBuffer
+import scala.collection.mutable
 import scala.collection.compat._
 
 /** Collection representation of a query string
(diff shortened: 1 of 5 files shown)


valencik:
I think we're definitely going to run into issues with unused imports in 2.13...

Current failures in CI:
```
[error] /home/runner/work/http4s/http4s/core/src/main/scala/org/http4s/Query.scala:32:32: Unused import
[error] import scala.collection.compat._
```

rossabaker:
I think #4310 is sufficient, and we can close #3945.

core/src/main/scala/org/http4s/Query.scala
```diff
@@ -133,14 +135,22 @@ final class Query private (value: Either[Vector[KeyValue], String])
...
 
   /** Map[String, Seq[String]] representation of the [[Query]]
     *
     * Params are represented as a `Seq[String]` and may be empty.
     */
   lazy val multiParams: Map[String, immutable.Seq[String]] =
-    CollectionCompat.pairsToMultiParams(toVector)
+    if (toVector.isEmpty) Map.empty
+    else {
+      val m = mutable.Map.empty[String, ListBuffer[String]]
```
isomarcte, line 147: I see this already existed before this PR, but does anyone know why we are using a mutable Map here?

rossabaker: It might be faster and I don't think the immutability is not observable.

isomarcte: > It might be faster and I don't think the immutability is not observable.

Yeah, I thought faster might be the reason, though I'm a bit skeptical of that unless we showed this was hot path and/or profiled it. 

Not a big deal though, it is totally hidden and easy enough to validate that it is correct.

rossabaker: I suspect the performance is similar enough, and the code local enough, it really makes little difference for performance or maintainability either way.  But I have open source YourKit licenses for anyone who wants to do Real Science.

I'm going to start beating on things with a profiler myself... after the Dotty and CE3 projects are done. :smile:

PR #4309, opened 2021-01-30
Add call of GenericServlet init during Http4sServlet init

### LINKED ISSUE #4307: Http4sServlet init method should call GenericServlet init
`GenericServlet` is doing some minimal setup in its `init` - preserving ref to `ServletConfig` in private field.

If the method is not called, then it will stay uninitialized (null) and when calling `log` method on implemented servlet it will throw exception:

```
java.lang.NullPointerException: Cannot invoke "javax.servlet.ServletConfig.getServletContext()" because the return value of "javax.servlet.GenericServlet.getServletConfig()" is null
```

arixmkii:
Fixes #4307


rossabaker (commented):
Thanks for the retarget. :+1: on green.

Issue #4297, opened 2021-01-29
Remove type projections in headers


rossabaker:
Scala 3 doesn't allow these:

```
[warn] -- Migration Warning: /home/ross/src/http4s/core/src/main/scala/org/http4s/HeaderKey.scala:58:35 
[warn] 58 |    def apply(values: NonEmptyList[HeaderT#Value]): HeaderT
[warn]    |                                   ^^^^^^^
[warn]    |                              Recurring.this.HeaderT is not a legal path
[warn]    |                              since it is not a concrete type
[warn] -- Migration Warning: /home/ross/src/http4s/core/src/main/scala/org/http4s/HeaderKey.scala:60:21 
[warn] 60 |    def apply(first: HeaderT#Value, more: HeaderT#Value*): HeaderT =
[warn]    |                     ^^^^^^^
[warn]    |                     Recurring.this.HeaderT is not a legal path
[warn]    |                     since it is not a concrete type
[warn] -- Migration Warning: /home/ross/src/http4s/core/src/main/scala/org/http4s/HeaderKey.scala:60:42 
[warn] 60 |    def apply(first: HeaderT#Value, more: HeaderT#Value*): HeaderT =
[warn]    |                                          ^^^^^^^
[warn]    |                              Recurring.this.HeaderT is not a legal path
[warn]    |                              since it is not a concrete type
[warn] -- Migration Warning: /home/ross/src/http4s/core/src/main/scala/org/http4s/HeaderKey.scala:66:28 
[warn] 66 |          acc: NonEmptyList[HeaderT#Value]): NonEmptyList[HeaderT#Value] =
[warn]    |                            ^^^^^^^
[warn]    |                            Recurring.this.HeaderT is not a legal path
[warn]    |                            since it is not a concrete type
[warn] -- Migration Warning: /home/ross/src/http4s/core/src/main/scala/org/http4s/HeaderKey.scala:66:58 
[warn] 66 |          acc: NonEmptyList[HeaderT#Value]): NonEmptyList[HeaderT#Value] =
[warn]    |                                                          ^^^^^^^
[warn]    |                              Recurring.this.HeaderT is not a legal path
[warn]    |                              since it is not a concrete type
[warn] -- Migration Warning: /home/ross/src/http4s/core/src/main/scala/org/http4s/HeaderKey.scala:70:78 
[warn] 70 |              loop(Headers(hs.toList.tail), acc.concatNel(header.values.widen[HeaderT#Value]))
[...]

PR #4319, opened 2021-01-31
Compile core on Scala 3.0.0-M3

rossabaker:
Turned off migration to work around https://github.com/lampepfl/dotty/issues/11266.

Now we are blocked by #4297.

@mpilquist points out that this is a scalafmt error, which I'm choosing to ignore for now.  We could hush it [like this](https://github.com/Comcast/ip4s/blob/b5b65a626f1abcdc4316891f999efeddc8a548b7/build.sbt#L102-L104), but I'm hoping scalafmt catches up before we acquire more build cruft.

```
/home/ross/src/http4s/core/src/main/scala-3/org/http4s/syntax/LiteralsSyntax.scala:78: error: unclosed character literal
      '{Uri.unsafeFromString(${Expr(literal)})}
```


rossabaker:
That ip4s snapshot uses withDottyCompat for cats-core, so it still doesn't work.

PR #4320, opened 2021-01-31
Migrate date spec

m-sp:


server/src/test/scala/org/http4s/server/middleware/DateSuite.scala
```diff
@@ -0,0 +1,96 @@
...
+
+import cats.data.OptionT
+import cats.syntax.all._
+import cats.effect._
+import org.http4s._
+import org.http4s.syntax.kleisli._
+import org.http4s.headers.{Date => HDate}
+
+class DateSuite extends Http4sSuite {
+  implicit val timer: Timer[IO] = Http4sSuite.TestTimer
```
m-sp, line 27: targeting 0.22 due to the TestTimer which doesn't seem available in 0.21

rossabaker: I did the Specs2 quarantine on series/0.22 because we had some deprecated things in testing that would break bincompat to get rid of.  Part of that work was to copy over some of those vals like `TestTimer`, which we probably could have done on 0.21.

m-sp: based it on top of the test-timer branch

rossabaker:
Uh-oh.  This got closed when I cleaned up my branch.

PR #4327, opened 2021-02-01
Fix parser cycles

m-sp:
seems that it is super easy to accidentally introduce an initializiation cycle in the uri parsers.
so this is a request for comments wether the uri parsers should live in there own dedicated object so that issues with hangs or nullpointers are gone

### DIFF
--- a/core/src/main/scala/org/http4s/Uri.scala
+++ b/core/src/main/scala/org/http4s/Uri.scala
@@ -170,7 +170,7 @@ object Uri extends UriPlatform {
 
   /** Decodes the String to a [[Uri]] using the RFC 3986 uri decoding specification */
   def fromString(s: String): ParseResult[Uri] =
-    ParseResult.fromParser(uriReferenceUtf8, "Invalid URI")(s)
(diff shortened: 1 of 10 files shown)


rossabaker (commented):
I think this is the way. Parsers on companions is usually a nice model for the private constructors, but we've had repeated bugs using it here, where the cyclic dependencies can be subtle.

core/src/main/scala/org/http4s/Uri.scala
```diff
@@ -302,7 +193,7 @@ object Uri extends UriPlatform {
     *
     * @see [[https://tools.ietf.org/html/rfc3986#section-3.1 RFC 3986, Section 3.1, Scheme]]
     */
-  final class Scheme private (val value: String) extends Ordered[Scheme] {
+  final class Scheme private[http4s] (val value: String) extends Ordered[Scheme] {
```
rossabaker, line 196: I made a comment that got eaten: I think this can be constrained to `Uri` because it's an inner type and be closer to the original.

rossabaker (approved):
This is a critical fix and we're revving up a release train.

PR #4331, opened 2021-02-02
Add NoPayloadMethods optimization

ChristopherDavenport:


rossabaker (approved):
I'd like a more general solution, but this looks at parity with the blaze solution.

PR #4332, opened 2021-02-02
Empty Entity dsl

ChristopherDavenport:
If an entity is allowed and none is provided apply a `Content-Length: 0`, otherwise just use the provided headers

Only does so if Entity is allowed. Where none is allowed do not add content-length

### DIFF
--- a/dsl/src/main/scala/org/http4s/dsl/impl/ResponseGenerator.scala
+++ b/dsl/src/main/scala/org/http4s/dsl/impl/ResponseGenerator.scala
@@ -46,7 +46,12 @@ private[impl] object ResponseGenerator {
   */
 trait EmptyResponseGenerator[F[_], G[_]] extends Any with ResponseGenerator {
   def apply(headers: Header*)(implicit F: Applicative[F]): F[Response[G]] =
-    F.pure(Response(status, headers = Headers(headers.toList)))
+    F.pure(
+      Response(
+        status,
+        headers =
+          if (status.isEntityAllowed) addEntityLength[F](Entity.empty, Headers(headers.toList))
+          else Headers(headers.toList)))
 }
 
 /** Helper for the generation of a [[org.http4s.Response]] which may contain a body


hamnis (approved):

rossabaker (commented):
This is interesting. This response DSL has been derided over the years, but using it this way drives behavior that the RFC says we "should" be doing.  It steers toward correctness in a way the case-class syntax does not.

Alternatively, we could implement this in the backend and be correct more often, but it would be more implicit behavior that may vary between backends.

Or, we could have a shared "default middleware" that implements the standard headers (e.g., content length, date, server) in a common layer, at the expense of higher level servers maybe doing this work twice.

:thinking:

ChristopherDavenport:
I largely think we should build a layer that does this so its visible in application space. But I think that this is more correct than before regardless.

rossabaker:
I'm kind of skittish about this one.  There's no "DSL laws", but we have advertised that these are substitutable, and that will no longer be so:

```scala
Ok() === Response(Ok).pure
```

It also makes this footgun more prominent, though that's a defect of `withEntity`:

```scala
scala> Ok().map(_.withEntity(Stream("hello".getBytes).covary[IO])).unsafeRunSync()
val res45: org.http4s.Response[cats.effect.IO] = Response(status=200, headers=Headers(Content-Length: 0, Content-Type: application/octet-stream, Transfer-Encoding: chunked))
```

This is a hack around us not preserving invariants between the headers and the body, and I think it may be a better experimental feature for 0.22 than something we should hurry into 0.21.  I think a backend solution would be preferrable for 0.21.

diesalbla:
> I think a backend solution would be preferrable for 0.21.

Since the body is a black box until compiled, perhaps we could add a process that, at the point a `Response` is encoded, that: 
- Peeks into the first "chunk" of the body, and checks if it is empty and last.
- If the first chunk is empty, or the body is empty, then use the length as the  `Content-Length` header. If there is a header already, perhaps it should override it.

This would have the downside of breaking the relation between what is sent and what is built with the `Response` case class.

rossabaker:
That strategy is roughly what I had in mind.

The relationship is already broken: `Date`, `Server`, and `User-Agent` are all added by various backends.  All of this is good for spec compliance, and all of this is to the detriment of the `Logger` middleware.  We ended up siding with the spec, but it would be nice to resolve this tension.

ScalaWilliam:
Was not a fan of akka-http deciding to mangle with the Content-Type headers - don't remember the exact example though - might be relevant to the topic.

PR #4333, opened 2021-02-02
Add Content-Length to Empty Dsl Responses 0.22

ChristopherDavenport:
Same as #4332 but targeting 0.22


hamnis:
we should rather merge from series/0.21 and then to main.

rossabaker:
Regarding git strategy, you're both right.

We should always merge from earlier branches: having the original work in the git parentage prevents us from re-merging work later.  The alternative is to cherry-pick everything, and have no assistance from git keeping feature parity.

But tickets like this make complicated merges easier: instead of someone doing several unrelated ones at once, like I was doing at 3am last night, the answer can be a simple "keep ours" from tickets like these.

If it's trivial, don't bother with these.  If it's not, these are much appreciated by me!

PR #4335, opened 2021-02-02
Ember Entity Encoding No ContentLength or TransferEncoding on NoEntityAllowed status'

ChristopherDavenport:
This previously would have encoded as chunked encoding, and then had a single empty buffer flush which is unnecessary for these.


rossabaker:
:+1: The workflow needs regeneration. Probably inherited from the CVE fork that didn't run on CI.

rossabaker:
I think this is fit for 0.21?

ChristopherDavenport:
Yes I it can be retargeted it branched of 0.21

rossabaker:
Picked into 0.21 as 31850ff.

PR #4337, opened 2021-02-02
Pass maxConnections correctly

RaasAhsan:


djspiewak (approved):

ChristopherDavenport (approved):

ChristopherDavenport:
Needs scalafmt

hamnis (approved):

rossabaker (commented):
I'm so burnt out I still don't see the mistake. Are the named arguments in a different order in the definition of `fixed`?

I trust that it's right, but if someone can describe it for the changelog, that would be swell.

RaasAhsan:
The type signature of `fixed`:
```scala
def fixed(
      workerThreads: Int = DefaultPoolSize,
      bufferSize: Int = DefaultBufferSize,
      channelOptions: ChannelOptions = ChannelOptions.DefaultOptions,
      selectorThreadFactory: ThreadFactory = defaultWorkerThreadFactory,
      acceptorThreads: Int = 1,
      acceptorThreadFactory: ThreadFactory = defaultAcceptorThreadFactory,
      maxConnections: Int = DefaultMaxConnections
  ): ServerChannelGroup
```
We were passing in `maxConnections` to `acceptorThreads` since it wasn't named explicitly, and the formal parameter `maxConnections` was taking the default. I just added the names to all the parameters for consistency

rossabaker:
Got it. Thanks for digging. I've got three hours' sleep and two barking dogs. Yeah, this is bad.

PR #4342, opened 2021-02-04
Match HttpDateSuite current test with HttpDateSpec

valencik:
This is kind of a wild guess attempt to fix https://github.com/http4s/http4s/issues/4341

All I've done is change this test to match the existing one in HttpDateSpec which does not have the "extra" `IO.map` call:
https://github.com/http4s/http4s/blob/79c78d164b1d05e812a10bc10f4160f89b2d8ef5/tests/src/test/scala/org/http4s/HttpDateSpec.scala#L27

### DIFF
--- a/tests/src/test/scala/org/http4s/HttpDateSuite.scala
+++ b/tests/src/test/scala/org/http4s/HttpDateSuite.scala
@@ -23,8 +23,8 @@ class HttpDateSuite extends Http4sSuite {
   test("current should be within a second of Instant.now") {
     for {
       current <- HttpDate.current[IO]
-      now <- IO.delay(java.time.Instant.now).map(HttpDate.unsafeFromInstant)
-      diff = current.epochSecond - now.epochSecond
+      now <- IO(HttpDate.unsafeFromInstant(java.time.Instant.now))
+      diff = now.epochSecond - current.epochSecond
     } yield assert(diff == 0 || diff == 1, "diff was " + diff)
   }
 


valencik:
First commit failed from likely another flakey test:
```
==> X org.http4s.client.blaze.BlazeClient213Suite.reset request timeout  2.312s
  java.util.concurrent.TimeoutException: Request to http://fv-az60-448.internal.cloudapp.net:34133 timed out after 2000 ms
    at org.http4s.client.blaze.BlazeClient$$anon$1.run(BlazeClient.scala:167)
```

Pushing an empty commit to run CI again...

tests/src/test/scala/org/http4s/HttpDateSuite.scala
```diff
@@ -23,7 +23,7 @@ class HttpDateSuite extends Http4sSuite {
   test("current should be within a second of Instant.now") {
     for {
       current <- HttpDate.current[IO]
-      now <- IO.delay(java.time.Instant.now).map(HttpDate.unsafeFromInstant)
+      now = HttpDate.unsafeFromInstant(java.time.Instant.now)
```
rossabaker, line 26: Sorry if this is a duplicate comment. GitHub ate my first.

`current` is sampled before `now`, so I think `current <= now`.  If the second rolls over between the samples, then `diff` would be `-1`, as in the failing test, and never `1`, as in the assertion.

What you did might help because it removes a tiny amount of overhead between the samples and makes `diff` more likely to be 0.  But I think it'll still flake and the real problem is the assertion.  I wouldn't mind +/- 1.

A difference of -2 or more is even possible if the test machine is under stupidly high load.  I don't think we've ever seen that, but this test is really just a big race condition.  If we want to make it more reliable, we'd test against a constant `Clock` instance instead of `Instant.now`, and not need a clock at all.  At that point, the test would basically echo the implementation, but it moves us back into the realm of purity.

valencik: Oh interesting, I didn't notice the ordering in `diff`.
Yeah, if `current` is sampled first we shouldn't ever see `current - now == 1`

What do you think about changing the assertion to check `0` and `-1` for now?

Changing the test to use `Clock` is not immediately obvious to me, but I can poke around more if you think it's worth it.

valencik: Also noticing that in the older Spec version the ordering for diff is `now - current`

rossabaker: Yes, I think that fixes it for good.

And actually, I'd probably leave `Instant.now` wrapped in `IO`.  It's just a small test, and it doesn't really matter for maintainability at this scale.  But people look to tests to learn unfamiliar APIs, and I think it's good that they find things used correctly.  It should look weird to us to see something impure like `Instant.now` not wrapped in IO!

This is a lot of bikeshedding for a simple test, but I appreciate the effort, and I think it will be reliable again with that sign flip.

rossabaker (approved):

valencik:
Another `Http1ClientStageSpec` test flake
I'm going to rebase and squash these messy commits and hopefully get lucky on next CI...

PR #4345, opened 2021-02-04
Remove deprecated get(`Set-Cookie`) on Headers

rossabaker:
Found this again last night. Decrufting.

/cc @ashwinbhaskar, who is working in the neighborhood of headers.


rossabaker:
Sucked into #4352.

PR #4348, opened 2021-02-05
Ember body draining

RaasAhsan:
The current ember request/response parsers break down when using keep-alive connections and bodies are only partially read or completely unread. For example, if a server route receives a request with a body but doesn't read it and sends a response back, the stream at that point is reused when reading the next request. Since the body was unread, it will try to parse the next HTTP message from where the body actually starts.

This PR has several changes to solve this issue:
1. Shallow refactor to the parsers to work in terms of `F` rather than `Stream` and `Pull`, which gives us more control around reading from sockets without losing any expressive power.
2. Implements a state machine for reading and draining fixed-length bodies.
3. Implements a state machine for reading and draining chunk-encoded bodies.

There is a simpler, appealing alternative here: instead of draining the body to continue with the next request, if we know that it hasn't been completely read by the server or client, we can preemptively close the connection after we're done with it. We still need a state machine for fixed-length and variable-length bodies, but it will be binary: was the body read to completion or not? 

I think the only tradeoff to consider here is that it would break pipelining if preceding message bodies aren't read. I'm interested in seeing what other servers do. I believe Blaze will drain the body if the connection is reused.


rossabaker:
Yes, blaze drains the body to keep the connection alive.

Thinking through pros and cons, aside from code complexity:
* If there are just a few more bytes to read, and they're immediately available, we win big keeping the connection.
* If there are just a few more bytes to read, but they're not here yet, maybe our socket handle is better used elsewhere.  I suppose request timeouts deal with this.
* If there are gigabytes left to read, and we don't want them, it's kinder to server and client to close and start over.

RaasAhsan:
So we have fixed-length bodies and variable-length bodies. We know how much we've read so far from a fixed-length body and how much is left. If there are less than N bytes left and we have them, drain them, otherwise terminate the connection. For variable-length bodies, we could do something similar: read up to N more bytes/chunks to see if we reach the end, otherwise terminate. I think it'll be a bit more complex than that, but something along those lines could work.

rossabaker:
That sounds correct.

Note that the spec just says we have to be prepared for clients to try
this, but we can pessimistically close anytime the logic gets too risky.
Deficiencies in HTTP/1 pipelining is one of the compelling cases for HTTP/2.

If this is easy, it's worth a shot. If it starts to feel rickety, just
close it.

RaasAhsan:
So I was thinking about this a bit more and I realized that there is a way to retain the parsing logic to evaluate in `Stream` or `Pull`. The idea is to type the input stream as `Stream[F, (Boolean, Chunk[Byte])]`, so we can distinguish what chunks have already been read or not. I think weaving that through the parser is going to be annoying, so I'm inclined to move forward with the current solution because we aren't really losing much expressive power by evaluating in `F`. There are probably tangible performance benefits to doing it too

RaasAhsan:
Ok, so I think this is in a good place to review now. I still have a few TODOs to look at but they are fairly minor. The other piece is what we discussed above around how many bytes we're willing to drain from the body without closing. 
1. For fixed-length bodies, if the full body hasn't been read in a chunk yet (it doesn't have to be consumed), we're going to close. We can get smarter here and maybe check if there are extra bytes that are immediately readable.
2. For variable-length bodies, if the full body hasn't been consumed, we're going to close. We can also be smarter here and look at how many bytes we've already read into a chunk and also how many bytes are immediately available on the Socket.

To keep things digestible here, I'm going to leave the above two items for a follow-up. I'm also probably going to do a lot more exhaustive testing around chunking and connection reuse in a follow up.

ember-client/src/main/scala/org/http4s/ember/client/EmberClientBuilder.scala
```diff
@@ -166,20 +172,6 @@ final class EmberClientBuilder[F[_]: Concurrent: Timer: ContextShift] private (
                   userAgent
                 )
             )
-            .map(response =>
```
RaasAhsan, line 169: `ClientHelpers` already has this logic, correct me if this actually necessary twice

ember-client/src/main/scala/org/http4s/ember/client/internal/ClientHelpers.scala
```diff
@@ -129,20 +130,27 @@ private[client] object ClientHelpers {
...
+      drain: F[Option[Array[Byte]]],
+      nextBytes: Ref[F, Array[Byte]],
+      canBeReused: Ref[F, Reusable])(implicit F: Concurrent[F]): Response[F] = {
+    // TODO If Response Body has a take(1).compile.drain - would leave rest of bytes in root stream for next caller
     val out = resp.copy(
       body = resp.body.onFinalizeCaseWeak {
         case ExitCase.Completed =>
-          val requestClose = req.headers.get(Connection).exists(_.hasClose)
-          val responseClose = resp.headers.get(Connection).exists(_.hasClose)
+          drain.flatMap {
```
RaasAhsan, line 143: i think we might be able to move this logic somewhere else so we can actually drain even if the body wasn't read at all.

ember-core/src/main/scala/org/http4s/ember/core/ChunkedEncoding.scala
```diff
@@ -57,46 +62,55 @@ private[ember] object ChunkedEncoding {
...
+    val uncons = if (head.nonEmpty) (Some(Chunk.bytes(head)): Option[Chunk[Byte]]).pure[F] else read
+    uncons.flatMap {
+      case None => (Headers.empty, Array.emptyByteArray).pure[F]
+      case Some(chunk) =>
+        // TODO: Is this state eliminated because we check for nonemptiness already? Or can we get an empty chunk from read
+        if (chunk.isEmpty) parseTrailers(maxHeaderSize)(Array.emptyByteArray, read)
         else if (chunk.toByteVector.startsWith(Shared.`\r\n`))
-          Pull.pure(Headers.empty)
+          // TODO: return remaining bytes here
+          (Headers.empty, chunk.toArray.drop(1)).pure[F]
```
RaasAhsan, line 106: This should be 2, not 1

ember-core/src/main/scala/org/http4s/ember/core/Parser.scala
```diff
@@ -30,13 +30,17 @@ private[ember] object Parser {
...
+        read: F[Option[Chunk[Byte]]],
         maxHeaderLength: Int,
-        acc: Option[ParseHeadersIncomplete])
-        : Pull[F, Nothing, (Headers, Boolean, Option[Long], Stream[F, Byte])] =
-      s.pull.uncons.flatMap {
-        case Some((chunk, tl)) =>
+        acc: Option[ParseHeadersIncomplete])(implicit
+        F: MonadThrow[F]): F[(Headers, Boolean, Option[Long], Array[Byte])] = {
+      // TODO: improve this
+      val uncons = if (head.nonEmpty) F.pure(Some(Chunk.bytes(head))) else read
```
RaasAhsan, line 40: This pattern basically shows up everywhere. I wonder if we can stick `head` and `read` together to form a more interesting type

ember-client/src/main/scala/org/http4s/ember/client/internal/ClientHelpers.scala
```diff
@@ -100,19 +100,20 @@ private[client] object ClientHelpers {
...
+              head,
+              connection.keySocket.socket.read(chunkSize, durationToFinite(idleTimeout))
+            )
+        }
 
     for {
       processedReq <- preprocessRequest(request, userAgent)
-      resp <- writeRead(processedReq)
-    } yield postProcessResponse(processedReq, resp, reuseable)
+      (resp, drain) <- writeRead(processedReq)
```
rossabaker, line 115: This won't compile in Scala 3, but we can deal with it there.

ember-core/src/test/scala/org/http4s/ember/core/ParsingSpec.scala
```diff
@@ -254,21 +295,50 @@ class ParsingSpec extends Http4sSuite {
...
+  }
+
+  test(
+    "Parser.Response.parser should parse a response with Content-Length and return the rest of the stream") {
+    val defaultMaxHeaderLength = 4096
+    val raw =
+      """HTTP/1.1 200 OK
+        |Host: localhost:8080
+        |User-Agent: curl/7.64.1
+        |Accept: */*
```
rossabaker, line 319: `Host` and `Accept` aren't typical response headers, but they're syntactically valid so I don't think it negates the test.

ember-client/src/main/scala/org/http4s/ember/client/internal/ClientHelpers.scala
rossabaker: Stick it in the vault and move it up to where we have a `Resource[F, Response[F]]`?  Is it too late to flip the `canBeReused` at that point?

ember-client/src/main/scala/org/http4s/ember/client/EmberClientBuilder.scala
rossabaker: That looks correct.

ember-core/src/main/scala/org/http4s/ember/core/ChunkedEncoding.scala
```diff
@@ -57,46 +62,55 @@ private[ember] object ChunkedEncoding {
...
-    s.pull.uncons.flatMap {
-      case None => Pull.pure(Headers.empty)
-      case Some((chunk, tl)) =>
-        if (chunk.isEmpty) parseTrailers(maxHeaderSize)(tl)
+  )(head: Array[Byte], read: F[Option[Chunk[Byte]]]): F[(Headers, Array[Byte])] = {
+    val uncons = if (head.nonEmpty) (Some(Chunk.bytes(head)): Option[Chunk[Byte]]).pure[F] else read
+    uncons.flatMap {
+      case None => (Headers.empty, Array.emptyByteArray).pure[F]
+      case Some(chunk) =>
+        // TODO: Is this state eliminated because we check for nonemptiness already? Or can we get an empty chunk from read
```
rossabaker, line 102: We just eliminated some empty chunk logic in Multipart, but only in FS2-3.  It didn't work in FS2-2.

ember-core/src/main/scala/org/http4s/ember/core/ChunkedEncoding.scala
rossabaker: This is why I sometimes write `"\r\n".length`.  I _think_ that gets inlined, but I've never inspected the bytecode to be sure.

ember-client/src/main/scala/org/http4s/ember/client/internal/ClientHelpers.scala
RaasAhsan: Yep, that's what I had in mind and it worked out perfectly

ember-client/src/main/scala/org/http4s/ember/client/internal/ClientHelpers.scala
RaasAhsan: It disappeared anyways

ember-client/src/test/scala/org/http4s/ember/client/internal/ClientHelpersSuite.scala
```diff
@@ -103,127 +101,92 @@ class ClientHelpersSuite extends Http4sSuite {
       .assertEquals(Some(name))
   }
 
-  test("Postprocess response should reuse when body is run") {
```
RaasAhsan, line 106: I deleted several tests that relied on the body being run. They can exist in some other form but they aren't applicable here anymore since the body doesn't necessarily need to be run for reuse anymore (i.e. if we already have a chunk with the body).

RaasAhsan: In other words, reuse is dependent on whether the `drain` function returns a `Some`. If it does, it means we've already buffered enough to read the next request/response. If it returns a `None`, that means we don't have enough bytes so close the connection.

Note that for variable-length bodies, if the body hasn't been read, it's going to always close, just on account of how the parser is coded.

RaasAhsan:
Just need to review error handling and I'll be happy with this.

ember-client/src/test/scala/org/http4s/ember/client/internal/ClientHelpersSuite.scala
rossabaker: Will `drain` return `None` if we've successfully consumed the entire body and are simply awaiting the next message?    I'm afraid I may have pointed in the wrong direction when I stated:

> If there are just a few more bytes to read, but they're not here yet, maybe our socket handle is better used elsewhere. I suppose request timeouts deal with this.

It's one thing to aggressively close when we know that we have to discard bytes that haven't arrived yet before we get to the next interesting one.  What I think we don't want is to aggressively close the connection when we successfully drained the previous message.  The non-pipeline case is common, where one response is wholly received before the next request begins.  If we're closing immediately in that case, I think we'll rarely have reused connections and performance will suffer.

I was trying to study prior art, and I think the relevant part of Jetty's implementation is [here](https://github.com/eclipse/jetty.project/blob/29d48b0de5fe4529c6aedb13fe96934147bcf9a8/jetty-server/src/main/java/org/eclipse/jetty/server/HttpConnection.java#L413), but I a lot more context to absorb before fully understanding it.

RaasAhsan: If we've read and consumed all available bytes and are awaiting the next message, `drain` is going to return `Some(Array.emptyByteArray)`. I'll add some tests to exercise that behavior

RaasAhsan:
Alright, I think this is in a good spot. We won't encounter any errors when draining a body because we're not actually consuming from the socket (yet, this is going to be a followup). In another followup, I'm going to add a bunch of property-based tests which should hopefully both supersede most of these handwritten tests and more exhaustively cover various chunking patterns.

I also want to make sure I get @ChristopherDavenport 's blessing on this

rossabaker (approved):

PR #4351, opened 2021-02-06
FS2 - Remove "elide empty chunks"

diesalbla:
The FS2 library, by construction, guarantees that the `uncons` does not give, in the result, a non-empty chunk.

### DIFF
--- a/core/src/main/scala/org/http4s/multipart/MultipartParser.scala
+++ b/core/src/main/scala/org/http4s/multipart/MultipartParser.scala
@@ -283,18 +283,6 @@ object MultipartParser {
     //whether it's the boundary plus an extra "--", indicating it's
     //the last boundary
     def checkIfLast(c: Chunk[Byte], rest: Stream[F, Byte]): SplitStream[F] = {
-      //Elide empty chunks until nonemptychunk is found
-      def elideEmptyChunks(str: Stream[F, Byte]): Pull[F, Nothing, (Chunk[Byte], Stream[F, Byte])] =
-        str.pull.uncons.flatMap {
-          case Some((chnk, r)) =>
-            if (chnk.size <= 0)
-              elideEmptyChunks(r)
-            else
-              Pull.pure((chnk, r))
-          case None =>
-            Pull.raiseError[F](MalformedMessageBodyFailure("Malformed Multipart ending"))
-        }
-
       //precond: both c1 and c2 are nonempty chunks
       def checkTwoNonEmpty(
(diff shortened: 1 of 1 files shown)


rossabaker (commented):
When did this become true?  FS2-3 or FS2-2?  Is this something that could also be done on an earlier release?

diesalbla:
> When did this become true? FS2-3 or FS2-2? Is this something that could also be done on an earlier release?

I found about this invariant this week, but it would be interesting to know how far it extends. 

If one would like to find out, here is what has to be done: checkout the release of `fs2` being used, look for the `Output` class, which would be a subclass of `Pull` (in the 3.x branch), or a subclass of `FreeC` (on a recent 2.x). If that `Output` class is private, and the only call to its constructor includes the guard not to create it with empty chunk, as in [this line](https://github.com/typelevel/fs2/blob/17eff83fe5f0d0068e626aa9e6590eba6d62655e/core/shared/src/main/scala/fs2/Pull.scala#L241), then the `uncons` should not be giving out an empty chunk.

rossabaker:
It appears that's true in fs2-2.5, in which case we could do the same all the way back to 0.21.

core/src/main/scala/org/http4s/multipart/MultipartParser.scala
```diff
@@ -284,17 +284,6 @@ object MultipartParser {
     //the last boundary
     def checkIfLast(c: Chunk[Byte], rest: Stream[F, Byte]): SplitStream[F] = {
       //Elide empty chunks until nonemptychunk is found
```
rossabaker, line 286: This comment can be removed.

rossabaker (approved):
This looks good after the comment is stripped.  I'll probably try to cherry-pick it to 0.21, unless you want to reopen with the same logic against that.

rossabaker:
It cherry-picked cleanly to series/0.21! :tada:

rossabaker:
Well, I should have run the tests instead of relying on reason and the compiler:

```
[error]   ! drain the epilogue with chunk size 3 (3 ms)
[error]    java.lang.RuntimeException: Chunk.empty.apply(0) (Chunk.scala:514)
[error] fs2.Chunk$EmptyChunk.apply(Chunk.scala:514)
[error] fs2.Chunk$EmptyChunk.apply(Chunk.scala:512)
[error] org.http4s.multipart.MultipartParser$.checkIfLast$1(MultipartParser.scala:312)
```

I'm going to revert the cherry-pick and merge the original.

PR #4352, opened 2021-02-06
Header cleanups

rossabaker:
Some quick wins while we get the header working on Dotty:

* `GetT` is vestigial.
* ``get(`Set-Cookie`)`` has been deprecated since 0.16!
~~* The signature of default `unapply` methods changes in Scala 3, and we're not using the one on `HeaderKey` anyway.~~

### DIFF
--- a/core/src/main/scala/org/http4s/HeaderKey.scala
+++ b/core/src/main/scala/org/http4s/HeaderKey.scala
@@ -53,7 +53,6 @@ object HeaderKey {
     */
   trait Recurring extends Extractable {
     type HeaderT <: Header.Recurring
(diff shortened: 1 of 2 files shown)


SystemFw (approved):

rossabaker:
Well, well, look who only compiled core before opening a PR.

rossabaker:
> The signature of default unapply methods changes in Scala 3, and we're not using the one on HeaderKey anyway.

Actually we do use that, and the one that changes in Scala 3 does not take a `Header` argument. Removing this change from the PR.

ashwinbhaskar (approved):

PR #4355, opened 2021-02-06
Port media type, server sent and cookie specs

m-sp:


tests/src/test/scala/org/http4s/MediaTypeSpec.scala
```diff
@@ -18,67 +18,8 @@ package org.http4s
 
 import cats.kernel.laws.discipline.EqTests
 import org.http4s.laws.discipline.HttpCodecTests
-import cats.syntax.show._
 
 class MediaTypeSpec extends Http4sSpec {
```
rossabaker, line 22: Does @cquiroz's machinery for doing these in Http4sSuite only exist in later branches?

m-sp: oh havn't checked yet

cquiroz: There is a `checkAllF` method that could be the munit discipline for cats effect

It may need to adjust the law definitions though

PR #4359, opened 2021-02-06
A few more tests

cquiroz:
Signed-off-by: Carlos Quiroz <carlos.m.quiroz@gmail.com>


tests/src/test/scala/org/http4s/parser/OriginHeaderSuite.scala
```diff
@@ -0,0 +1,85 @@
...
+ */
+
+package org.http4s
+package parser
+
+import cats.data.NonEmptyList
+import org.http4s.headers.Origin
+import org.http4s.syntax.all._
+
+class OriginHeaderSuite extends munit.FunSuite {
```
m-sp, line 24: interesting any reasons you stick with `munit.FunSuite` instead of:

```suggestion
class OriginHeaderSuite extends Http4sSuite {
```

m-sp (approved):

tests/src/test/scala/org/http4s/parser/OriginHeaderSuite.scala
cquiroz: Well I don't need anything from HttpSuite for this particular test, so it seems less fragile to use the least powerful option that will do the job

PR #4362, opened 2021-02-06
Compile laws on Scala 3.0.0-M3

rossabaker:
Extends #4319.


rossabaker:
Hmm, this fix will melt when we resolve the merge conflict that's blocking #4319. Never mind.

PR #4363, opened 2021-02-06
Advance cleanup of Dotty server warnings

rossabaker:
Dotty doesn't like this, and Dotty is right.


cquiroz (approved):
LGTM

rossabaker:
Does not look good to scalafmt!

PR #4366, opened 2021-02-06
Deprecate http4s-argonaut

rossabaker:
Argonaut support has required burdensome build hacks over the years, and is again causing problems by publishing its Dotty version using withDottyCompat on its dependencies, rendering us unable to compile with or without Dotty compat.

http4s-argonaut's usage is estimated to be less than 5% of http4s-circe's, and Circe is the recommendation of the http4s team.

If no volunteer steps up by February 13 to fix and maintain this module, we will proceed with this deprecation and removal from series/0.22 and main.


rossabaker:
If a new maintainer does emerge for this module, note that the plan is to spin it off into its own repo: https://github.com/http4s/http4s/discussions/4140

ChristopherDavenport (approved):

hamnis (approved):

PR #4367, opened 2021-02-06
Port blaze tests

m-sp:
the merge forward, especially into main will be a serious pain. not sure if it's worth it. so feel free to close!
please do let me know if I can help with the forwards merges!


blaze-server/src/test/scala/org/http4s/server/blaze/Http1ServerStageSpec.scala
```diff
@@ -31,19 +31,18 @@ import org.http4s.blaze.util.TickWheelExecutor
...
+import org.http4s.syntax.all._
 import scala.concurrent.duration._
-import scala.concurrent.Await
 import _root_.io.chrisdavenport.vault._
 import org.http4s.testing.ErrorReporting._
 
-class Http1ServerStageSpec extends Http4sSpec with AfterAll {
-  sequential
+class Http1ServerStageSpec extends Http4sSuite {
+  override def munitFlakyOK = sys.env.get("CI").isDefined
```
m-sp, line 40: this is an alternative to the `skipOnCi` we could even move it to `Http4sSuite`, wdyt?

rossabaker: :+1:

rossabaker (commented):
For merging, we can try, and if it hurts, abandon.  But it'll almost certainly be easier to fight through it than to do it twice.

rossabaker:
Okay, this one turned out to be much harder than most.  I think it's mostly related to the `dispatcher`. If you have time to help merge `series/0.22` to `main`, that would be great.  If not, I can try to take another shot tomorrow.

PR #4373, opened 2021-02-07
Upgrade to blaze-0.15.0-M1

rossabaker:
This version trims some redundant features from blaze-http and brings Dotty support.


m-sp:
shouldn't this target 0.22?

rossabaker:
Oops.  Absolutely yes.

m-sp (approved):

PR #4378, opened 2021-02-07
Upgrade to scala-xml-2.0.0-M4

rossabaker:
There's no Dotty releases for scala-xml-1.  Onward!


rossabaker:
I am in a place of poor bandwidth.

PR #4381, opened 2021-02-08
merge 0.22 to main

m-sp:
that one was painful...


m-sp:
okay the merge didn't quite work out as expected. added a terrible hack for a suite level fixture.
will comeback to and cleanup once I have time

rossabaker:
Thanks! I need to spend some time understanding munit fixtures before I can weigh in on the hack, but this gets the mission accomplished for now.

m-sp:
> Thanks! I need to spend some time understanding munit fixtures before I can weigh in on the hack, but this gets the mission accomplished for now.

well depends on your definition of hack, it is kind of one official way:
https://scalameta.org/munit/docs/fixtures.html#ad-hoc-suite-local-fixtures

but everything that has a `var`, a `null` and a `unsafeRunSync` looks very hacky to me

PR #4384, opened 2021-02-08
replace usage of illtyped

m-sp:
one less thing that we need to worry about at our push for dotty


cquiroz:
Yes, I started replacing some of these, that compileErrors should work on dotty too

cquiroz (approved):
LGTM

rossabaker (approved):
Oh, cool!

I wonder how brittle these test messages will be.  Does `compileErrors` throw if it compiles or return a non-empty string?  Just simply asserting that it failed to compile is still good enough for me.  But that's something we can relax later if these do prove themselves to be brittle.  This unblocks us now.  Great work!

rossabaker:
Well, great work on 2.13. :laughing: Looks like an issue popped up on 2.12.

Before anyone asks the inevitable question, a reminder for those who don't follow Gitter: over half our downloads are still for 2.12!!!

m-sp:
> I wonder how brittle these test messages will be. Does `compileErrors` throw if it compiles or return a non-empty string? Just simply asserting that it failed to compile is still good enough for me. But that's something we can relax later if these do prove themselves to be brittle. This unblocks us now. Great work!

pretty sure it would return an empty string. but we can always assert on non emptiness

rossabaker:
I am so tired of these Blaze flakes and not opposed to being more aggressive about skipOnCi until we push blaze into its own repo.

PR #4387, opened 2021-02-09
Merge #4384 to series/0.22

rossabaker:
The merge conflict in UriParserSpec was huge.  We had a bunch of failing tests relating to fragments in the Uri.requestTarget, which I think we realized are illegal anyway in 0.22.  I just deleted the tests.

PR'ing this one for being non-trivial and for me deleting all the failing tests instead of thinking too hard.


m-sp (approved):
Lgtm, sorry for the conflicts

hamnis:
regards the illTyped tests.

the compile messages is most likely different on dotty, so we need a Platform test.
They also have a tendency to change even between minor versions, so these tests seem a bit brittle.

rossabaker:
No need to apologize for the conflicts!

I hadn't considered the Dotty errors, but that's almost certainly correct.  It should be pretty easy to relax the assertions.  We could probably even introduce an `illTyped` function in `Http4sSuite` that asserts the non-emptiness of `compileErrors`.  I don't think we need to branch on platform to assert the precise errors per compiler version.

PR #4390, opened 2021-02-10
More core compilation fixes for 3.0.0-M3

rossabaker:
I thought I did all these already, but I can't find them now.


rossabaker:
Merging by executive order as it's blocking others, and I think these are old commits.  If you see a problem, you know where the pull request button is!

PR #4396, opened 2021-02-10
Skip Boopickle in Dotty

rossabaker:
There is no Boopickle release for Dotty (https://github.com/suzaku-io/boopickle/issues/148), and we can't yet compile it withDottyCompat because of its macro.  This can keep the project alive until the Great Schism of this repo.


cquiroz (approved):
LGTM

PR #4403, opened 2021-02-11
Remove `Content-Length` and `Content-Encoding` when decompressing

desbo:
When the gzip middleware decompresses a response, these headers are no
longer correct, so it seems best to remove them. This is crucial when
the client is used as a proxy (or, in my case, as an STTP backend) to
prevent downstream consumers attempting to decompress the response
again.

This change was previously introduced here: https://github.com/http4s/http4s/pull/2673#discussion_r298684594 but was removed in this PR: https://github.com/http4s/http4s/pull/3476

### DIFF
--- a/client/src/main/scala/org/http4s/client/middleware/GZip.scala
+++ b/client/src/main/scala/org/http4s/client/middleware/GZip.scala
@@ -20,7 +20,8 @@ package middleware
 
 import cats.effect.BracketThrow
 import fs2.{Pipe, Pull, Stream}
-import org.http4s.headers.{`Accept-Encoding`, `Content-Encoding`}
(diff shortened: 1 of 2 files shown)


bastewart:
The behaviour was changed to keep the header in a very deliberate commit: e04be63f532e197fe94e009489a0fe54e3d4a883 ("Keep Content-Encoding header after decompression")

We can't see any discussion in the PR on the motivation for this change though.

rossabaker (approved):
I agree with this.  I'm struggling to remember a good reason to preserve those headers.  The failing test is an unrelated flaky blaze test.

One thing this middleware does not do is support multiple content codings.  It's legal to layer them, in which case we should pop the one we reversed and keep the others.  But it's not a bug now, because we only decompress if we find exactly the one, and I think multiple codings is an exotic case.

PR #4405, opened 2021-02-11
Port header tests

cquiroz:
Final port of the header tests. We had two variants of `HeaderLaws` for specs2 and munit but now we can delete the former


m-sp (approved):
Nice!

hamnis (approved):

PR #4406, opened 2021-02-11
Lower constraint for EntityEncoder#fileEncoder from Effect to Sync

andreamarcolin:
Probably an oversight from https://github.com/http4s/http4s/commit/2635573b43975dd4fb4a7b66efeeca978d220381

### DIFF
--- a/core/src/main/scala/org/http4s/EntityEncoder.scala
+++ b/core/src/main/scala/org/http4s/EntityEncoder.scala
@@ -17,7 +17,7 @@
 package org.http4s
 
 import cats.{Contravariant, Show}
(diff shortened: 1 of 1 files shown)


hamnis:
Thanks for the PR!

This is not binary compatible, and we are trying to avoid adding binary incompatible changes on 0.21. 
In that spirit it should then probably be retargeted to `series/0.22`

andreamarcolin:
Hi @hamnis! Thanks for the prompt response and the suggestion. 
That was actually my bad, I did run `sbt ci` but it failed due to a flaky test I think (BlazeClient IIRC), so I committed without at least running MiMa.

Closing in favor of #4407

PR #4407, opened 2021-02-11
Lower constraint for EntityEncoder#fileEncoder from Effect to Sync

andreamarcolin:
Probably an oversight from https://github.com/http4s/http4s/commit/2635573b43975dd4fb4a7b66efeeca978d220381


hamnis (approved):

rossabaker (approved):
:tada: Thanks!

PR #4415, opened 2021-02-14
Header redesign proposal

SystemFw:
`[A: Header]` 👍 , `HeaderT#Value` 👎


core/src/main/scala/org/http4s/HeaderV2.scala
```diff
@@ -0,0 +1,292 @@
...
+  /** Parses the header from its String representation.
+    * Could be a comma separated String in case of a Header with
+    * multiple values.
+    */
+  def parse(headerValue: String): Option[A]
+}
+object Header {
+  case class Raw(name: CIString, value: String, recurring: Boolean = true)
+  object Raw {
+    implicit val showForRawHeaders: Show[Raw] =
```
rossabaker, line 50: We've never gotten consistent about our implicit names.  Some are globally unique in the cats style, some are probably unique like this, and some are just `instance`.

core/src/main/scala/org/http4s/HeaderV2.scala
```diff
@@ -0,0 +1,292 @@
...
+      h => s"${h.name.show}: ${h.value}"
+
+    implicit lazy val orderForRawHeaders: Order[Raw] =
```
rossabaker, line 53: We're missing a `Hash` instance, too.  (Does anybody actually use it as a typeclass?)

core/src/main/scala/org/http4s/HeaderV2.scala
```diff
@@ -0,0 +1,292 @@
...
+
+  /** Target for implicit conversions to Header.Raw from custom
+    * headers and key-value pairs.
+    *
+    * A method taking variadic `ToRaw` arguments will allow taking
+    * heteregenous arguments, provided they are either:
+    *
+    * - A value of type `A`  which has a `Header[A]` in scope
+    * - A (name, value) pair of `String`, which is treated as a `Recurring` header
+    * - A `Header.Raw`
```
rossabaker, line 78: This would be a great place for an inline example.

core/src/main/scala/org/http4s/HeaderV2.scala
```diff
@@ -0,0 +1,292 @@
...
+  /** Value of the header, which is represented as a String.
+    * Will be a comma separated String for headers with multiple values.
+    */
+  def value(a: A): String
+
+  /** Parses the header from its String representation.
+    * Could be a comma separated String in case of a Header with
+    * multiple values.
+    */
+  def parse(headerValue: String): Option[A]
```
rossabaker, line 45: I'm wondering about `Option` vs. `Either`.  We've never told people why their headers are bad.

core/src/main/scala/org/http4s/HeaderV2.scala
```diff
@@ -0,0 +1,292 @@
...
+      )
+  }
+
+  /** Classifies custom headers into `Single` headers, which can only
+    * appear once, and `Recurring` headers, which can appear multiple
+    * times.
+    */
+  sealed trait Type
+  case class Single() extends Type
+  case class Recurring() extends Type
```
rossabaker, line 66: This sounds ridiculous, but blame the committee: might `SetCookie` be the third?

core/src/main/scala/org/http4s/HeaderV2.scala
```diff
@@ -0,0 +1,292 @@
...
+  object ToRaw {
+    implicit def rawToRaw(h: Header.Raw): Header.ToRaw = new Header.ToRaw {
+      val value = h
+    }
+
+    implicit def keyValuesToRaw(kv: (String, String)): Header.ToRaw = new Header.ToRaw {
+      val value = Header.Raw(CIString(kv._1), kv._2)
+    }
+
+    implicit def customHeadersToRaw[H](h: H)(implicit H: Header[H, _]): Header.ToRaw =
```
rossabaker, line 94: `custom` to me sounds like a third-party addition.  akka-http calls these `"modeled"` headers, which I like.

It doesn't matter too much, since it's implicit.

core/src/main/scala/org/http4s/HeaderV2.scala
```diff
@@ -0,0 +1,292 @@
...
+
+    implicit def recurringHeaders[A](implicit
+        h: Header[A, Header.Recurring]): Select[A] { type F[B] = NonEmptyList[B] } =
+      new Select[A] {
+        type F[B] = NonEmptyList[B]
+
+        def toRaw(a: A): Header.Raw =
+          Header.Raw(h.name, h.value(a))
+
+        def from(headers: List[Header.Raw]): Option[NonEmptyList[A]] =
```
rossabaker, line 137: So we're choosing not to combine them then, but to return a List of them.  We might have `Some(NonEmptyList(Accept(NonEmptyList(...))))`?

core/src/main/scala/org/http4s/HeaderV2.scala
```diff
@@ -0,0 +1,292 @@
...
+    * Deduplicates non-recurring headers.
+    */
+  def of(headers: List[Header.Raw]): Headers =
+    if (headers.isEmpty) new Headers(List.empty)
+    else {
+      val acc = MutSet.empty[CIString]
+      val res = ListBuffer.empty[Header.Raw]
+
+      headers.foreach { h =>
+        if (h.recurring || h.name == "Set-Cookie" || acc.add(h.name))
```
rossabaker, line 217: This needs to be a `CIString("Set-Cookie")`, right?

rossabaker (commented):
This looks highly promising!

core/src/main/scala/org/http4s/HeaderV2.scala
SystemFw: Yeah that makes sense, although you then have to design an error type potentially

core/src/main/scala/org/http4s/HeaderV2.scala
SystemFw: Yeah I tried to infer a pattern but failed

core/src/main/scala/org/http4s/HeaderV2.scala
SystemFw: ah, I didn't see one, but `hashCode` is overriden, we'll need to add one then

core/src/main/scala/org/http4s/HeaderV2.scala
SystemFw: I was actually thinking this might be the case

core/src/main/scala/org/http4s/HeaderV2.scala
SystemFw: It makes sense, I've used "custom" in the scaladoc too

core/src/main/scala/org/http4s/HeaderV2.scala
SystemFw: It's hard to combine them, because we no longer have the information about the fact that there is ` values` field of type `NonEmptyList` inside our `A`.  We could introduce the semigroup idea you had, this is mostly to get the mechanics for `get` working, rather than a firm decision on how to combine them (or whether to)

core/src/main/scala/org/http4s/HeaderV2.scala
SystemFw: yeah, good catch. It compiles, which is worrying.

core/src/main/scala/org/http4s/HeaderV2.scala
SystemFw: Ok we reached a solution, I'll implement it tomorrow: if you provide an instance for `Semigroup[A]` and `Header[A, Recurring]`, we merge them, if you don't, you get a NonEmptyList

core/src/main/scala/org/http4s/HeaderV2.scala
rossabaker: `catsShowForOrgHttp4sHeadersRaw` would follow the Cats convention, but, geez.

Anyway, that should be a consistency pass before 0.22.  It shouldn't block this.  It's already a mess.

core/src/main/scala/org/http4s/HeaderV2.scala
rossabaker: We already have a `ParseFailure`, which is what all the header parsers return in a `ParseResult` alias.  I don't know that the latter alias carries its weight.  I was more pro-alias in those days.

core/src/main/scala/org/http4s/HeaderV2.scala
rossabaker: `===` would have caught it, though I've never got into the habit of using it myself.

core/src/main/scala/org/http4s/HeaderV2.scala
SystemFw: Are we definitely going for an `Either` then?

core/src/main/scala/org/http4s/HeaderV2.scala
SystemFw: yep. Although I think SetCookie doesn't need to be special-cased anymore, given the Recurring + Semigroup design

core/src/main/scala/org/http4s/HeaderV2.scala
rossabaker: I think it's a good idea to try.  If that makes the design irritating, copying back to `Option`, we can revert.

I think there's a rare opportunity for an `Ior` when combining headers when some of them are invalid, but maybe that's too cute, and I think we'd need a `Semigroup` for `ParseFailure`.

isomarcte: :+1: for `Either`

core/src/main/scala/org/http4s/HeaderV2.scala
isomarcte: I don't really mind the long names, since (hopefully) you are rarely if ever using them explicitly.

core/src/main/scala/org/http4s/HeaderV2.scala
```diff
@@ -0,0 +1,333 @@
...
+        Order.by(_.value)
+      )
+  }
+
+  /** Classifies custom headers into `Single` headers, which can only
+    * appear once, and `Recurring` headers, which can appear multiple
+    * times.
+    */
+  sealed trait Type
+  case class Single() extends Type
```
ashwinbhaskar, line 65: this could be a dumb question. But why not have the types as objects. eg: `object Single extends Type` ?

SystemFw: Because they are phantom types and only appear as type parameters, so if they were objects you'd have to write
```
implicit val yourHeader: Header[YourType, Single.type]
```
which not everyone knows, whereas 
```
implicit val yourHeader: Header[YourType, Single]
```
is more natural

cquiroz: Nice, I didn't know that

core/src/main/scala/org/http4s/HeaderV2.scala
```diff
@@ -0,0 +1,333 @@
...
+  def get[A](implicit ev: Header.Select[A]): Option[ev.F[A]] =
+    ev.from(headers)
+
+  /** TODO revise scaladoc
+    * Attempt to get a [[org.http4s.Header]] from this collection of headers
+    *
+    * @param key name of the header to find
+    * @return a scala.Option possibly containing the resulting [[org.http4s.Header]]
+    */
+  def get(key: CIString): Option[NonEmptyList[Header.Raw]] = headers.filter(_.name == key).toNel
```
cquiroz, line 180: Would it be simpler to just return `List[Header.Raw]`?

cquiroz:
This is looking great, we have had this pending for a really long time

core/src/main/scala/org/http4s/HeaderV2.scala
SystemFw: For some reason I like `Option[NonEmptyList` here, I think the emphasis is more on the fact that the header might not be present, and I'd imagine you'd want to do a bunch of `getOrElse`, whereas it's not as easy to do `onEmpty` on `List`. Certainly up for debate though. Also the typed `get` returns `Option[NonEmptyList[A]]` for recurring headers, so there's some consistency there

core/src/main/scala/org/http4s/HeaderV2.scala
SystemFw: I'm going to resolve this for now, making implicit names uniform across http4s is probably best left for a different PR :)

core/src/main/scala/org/http4s/HeaderV2.scala
```diff
@@ -25,8 +25,8 @@ import scala.collection.mutable.{ListBuffer, Set => MutSet}
 
 /** Typeclass representing an HTTP header, which all the http4s
   * default headers satisfy.
-  * You can add custom headers by providing an implicit instance of
-  * `Header[YourCustomHeader]`
+  * You can add modelled headers by providing an implicit instance of
```
rossabaker, line 28: It is Presidents' Day, yet we still have to endure the British spelling words with two L's! :disappointed:

(No, don't worry about it; let's make it compile.)

rossabaker:
What is the ultimate plan with this `v2` package?  Eliminate it when it's ready?

SystemFw:
> What is the ultimate plan with this `v2` package? Eliminate it when it's ready?

Yeah it's just to keep both things compiling for now. Once I delete the old code, `v2` is going away

laws/src/main/scala/org/http4s/laws/discipline/ArbitraryInstances.scala
```diff
@@ -571,27 +571,16 @@ private[http4s] trait ArbitraryInstances {
...
 
-  implicit val http4sTestingArbitraryForHeader: Arbitrary[Header] =
-    Arbitrary {
-      oneOf(
-        getArbitrary[`Accept-Charset`],
-        getArbitrary[Allow],
-        getArbitrary[`Content-Length`],
-        getArbitrary[Date],
-        getArbitrary[Header.Raw]
-      )
```
rossabaker, line 590: This is not living up to its full potential anyway, but it would be nice to regain something like this, that generates real headers.

core/src/main/scala/org/http4s/HeaderV2.scala
```diff
@@ -242,6 +242,8 @@ final class Headers(val headers: List[Header.Raw]) extends AnyVal {
       }
     }
 
+  def foreach(f: Header.Raw => Unit): Unit = headers.foreach(f)
```
rossabaker, line 245: Are we avoiding dump helpers like this?  I didn't bring back `nonEmpty`.

rossabaker:
Should there be an `is` still?

```
[error] /home/ross/src/http4s/ember-core/src/main/scala/org/http4s/ember/core/Encoder.scala:49:15: value is is not a member of org.http4s.v2.Header.Raw
[error]         if (h.is(`Content-Length`)) appliedContentLength = true
```

core/src/main/scala/org/http4s/HeaderV2.scala
SystemFw: I removed them all, mostly to see how painful it would be. I think adding helpers as we find the need for them is perfectly fine, I just couldn't distinguish between those that were there because they were needed, and those that were there because we wanted to create a collection type

SystemFw:
> Should there be an is still?

no there is no need for this I don't think: although an `is[HeaderType]` method on `Header.Raw` is certainly possible, I think you can rewrite the code to not require that

laws/src/main/scala/org/http4s/laws/discipline/ArbitraryInstances.scala
roundnutz14: I am totally lost

server/src/main/scala/org/http4s/server/staticcontent/FileService.scala
```diff
@@ -160,14 +169,15 @@ object FileService {
...
-            Some(
-              Response[F](
-                status = Status.RangeNotSatisfiable,
-                headers = v2.Headers
-                  .apply(AcceptRangeHeader, `Content-Range`(SubRange(0, size - 1), Some(size)))))
-          }
-      case _ => F.pure(None)
+          else nope
+      case _ =>
+        req.headers.get(CIString("Range")) match {
```
rossabaker, line 174: This is where I'm thinking the Select machinery returning an `Ior` might carry some weight.  We have to handle headers that just vanish because they're invalid.

m-sp (approved):

PR #4419, opened 2021-02-14
Dotty progress

m-sp:
misc dotty fixes

### DIFF
--- a/circe/src/test/scala/org/http4s/circe/CirceSuite.scala
+++ b/circe/src/test/scala/org/http4s/circe/CirceSuite.scala
@@ -77,9 +77,7 @@ class CirceSuite extends JawnDecodeSupportSuite[Json] with Http4sLawSuite {
 
   test("json encoder should have json content type") {
     val maybeHeaderT: Option[`Content-Type`] = jsonEncoder[IO].headers.get(`Content-Type`)
-    assertEquals(
-      maybeHeaderT,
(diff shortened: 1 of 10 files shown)


circe/src/test/scala/org/http4s/circe/CirceSuite.scala
```diff
@@ -77,9 +77,7 @@ class CirceSuite extends JawnDecodeSupportSuite[Json] with Http4sLawSuite {
 
   test("json encoder should have json content type") {
     val maybeHeaderT: Option[`Content-Type`] = jsonEncoder[IO].headers.get(`Content-Type`)
-    assertEquals(
-      maybeHeaderT,
-      Some(`Content-Type`(MediaType.application.json)))
+    assertEquals(maybeHeaderT, Some(`Content-Type`(MediaType.application.json)))
```
rossabaker, line 80: Did scalafmt want these?  I don't see the difference.

rossabaker:
Oh, I was somehow just looking at the last commit!

rossabaker:
Okay, I understand now. :+1:

circe/src/test/scala/org/http4s/circe/CirceSuite.scala
m-sp: Nah it works around inference issues. Would be solved with the header v2s aswell just wanted to make some progress

PR #4423, opened 2021-02-14
align thread name with 0.22

m-sp:
fixes the failing main build: https://github.com/http4s/http4s/runs/1898542059


rossabaker:
:+1: thanks!

PR #4424, opened 2021-02-14
Blaze-Core: Remove unused method.

diesalbla:
This fs2 helper seems to be unused.


m-sp:
Seems an import can be removed as well, rest lgtm!

rossabaker:
Thanks!  Test failure is #4423.  We don't need it even in 0.21, but I don't like MiMa exceptions, so I'll pick this to 0.22.

PR #4425, opened 2021-02-14
Blaze-Core: optimise entity body writer

diesalbla:
The method `Stream.evalMap` creates a singleton chunk for each element of the input chunk that is written.
There is no need to keep those `Chunk[Unit]`. Using the Pull interface, we can just do the action and continue iterating, without leaving any garbage behind.


rossabaker (approved):
Nice.

PR #4428, opened 2021-02-14
Update clients to be compatible with cats effect RC1

### LINKED ISSUE #4427: Blaze client throws java.lang.NoClassDefFoundError: cats/effect/implicits/package$ with cats-effect 3.0.0-RC1
Seems that cats effect RC1 introduced source and binary incompatible changes in various clients (blaze, async, ember).

arixmkii:
Fixes  #4427


arixmkii:
This fix worked for me to fix my test apps dependent on blazeClient (publishing artifacts locally), but requires more work on tests unfortunately. Will try to take a look on what is wrong soon.

arixmkii:
Needs updated fs2 (current snapshot is good enough, but will need to wait until another -M or -RC is published).

rossabaker:
Thanks! There are a bunch of binary incompatibities, so we're using #4414 to batch up the necessary changes.  We already had some of this, but I cherry-picked your commit to it as well.

PR #4429, opened 2021-02-15
Ember streaming parser tests

RaasAhsan:
By nature, streaming parsers should not make any assumptions about how input chunks are received from the source, or else it risks failing on inputs that are otherwise well-formed. I added a few property-based tests in this PR which assert that chunking is insignificant to the Ember HTTP message parsers. In the process, I found and fixed a few bugs.


ember-core/src/main/scala/org/http4s/ember/core/Parser.scala
```diff
@@ -356,83 +356,77 @@ private[ember] object Parser {
...
-                  method = method,
-                  uri = uri,
-                  httpVersion = httpVersion,
-                  headers = headers
-                )
-
-                if (chunked) {
-                  Ref.of[F, Option[Array[Byte]]](None).map { rest =>
+    )(implicit F: Concurrent[F]): F[(Request[F], F[Option[Array[Byte]]])] =
+      ReqPrelude
```
RaasAhsan, line 363: This is all indentation

ember-core/src/main/scala/org/http4s/ember/core/Parser.scala
```diff
@@ -356,83 +356,77 @@ private[ember] object Parser {
...
-              case (headers, chunked, contentLength, bytes) =>
-                val baseResp = org.http4s.Response[F](
-                  httpVersion = httpVersion,
-                  status = status,
-                  headers = headers
-                )
-
-                if (chunked) {
-                  Ref.of[F, Option[Array[Byte]]](None).map { rest =>
+      RespPrelude
```
RaasAhsan, line 401: This is all indentation

ember-core/src/test/scala/org/http4s/ember/core/StreamingParserSuite.scala
```diff
@@ -0,0 +1,198 @@
...
+  }
+
+  test("parse single request") {
+    PropF.forAllNoShrinkF(Fixtures.genRequest) { segments =>
+      (for {
+        read <- Helpers.taking[IO, Byte](segments)
+        result <- Parser.Request.parser(Int.MaxValue)(Array.emptyByteArray, read)
+        _ <- result._1.body.compile.drain
+        _ <- result._2
+      } yield true).assert
```
RaasAhsan, line 142: For now, I'm not asserting anything about the response body because we have other test suites like `TraversalSpec` that do (though that one is disabled right now). The main goal here is to assert that well-formed messages are parsed successfully regardless of their chunking.

ember-core/src/test/scala/org/http4s/ember/core/StreamingParserSuite.scala
```diff
@@ -0,0 +1,198 @@
...
+        "Mozilla\r\n",
+        "9\r\n",
+        "Developer\r\n",
+        "7\r\n",
+        "Network\r\n",
+        "0\r\n",
+        "\r\n"
+      ))
+
+    // TODO: Express in terms of Gen[Request[IO]] and Gen[Response[IO]]
```
RaasAhsan, line 125: This is a nice-to-have

ember-core/src/test/scala/org/http4s/ember/core/StreamingParserSuite.scala
```diff
@@ -0,0 +1,198 @@
...
+
+  object Helpers {
+    def taking[F[_]: Concurrent, A](segments: List[List[A]]): F[F[Option[Chunk[A]]]] =
+      for {
+        q <- Queue.unbounded[F, Option[Chunk[A]]]
+        _ <- segments.traverse(bytes => q.enqueue1(Some(Chunk.seq(bytes))))
+        _ <- q.enqueue1(None)
+      } yield q.dequeue1
+
+    def subdivided[A](as: List[A], count: Int): Gen[List[List[A]]] = {
```
RaasAhsan, line 38: Given a list, subdivide it up to `count` times such that the concatenation of all the sublists of the returned list equals the original list (there is a property-based test that asserts this at the end of the file). Maybe we can upstream this back into scalacheck?

ember-core/src/test/scala/org/http4s/ember/core/StreamingParserSuite.scala
RaasAhsan: That being said, it would be nice to ensure that we're not dropping bytes or anything in certain chunking patterns.

ember-core/src/main/scala/org/http4s/ember/core/Parser.scala
```diff
@@ -141,36 +140,38 @@ private[ember] object Parser {
...
-              idx += 1 // double advance to drop cr AND lf
+            } else if (current == lf && (idx > 0 && bv(idx - 1) == cr)) {
               complete = true // completed terminate loop
             }
           case 1 => // HeaderValue
             val current = bv(idx)
             // If crlf is next we have completed the header value
-            if (current == cr && ((bv.size >= idx + 1) && bv(idx + 1) == lf)) {
-              val hValue = new String(bv, start, idx - start) // extract header value
+            if (current == lf && (idx > 0 && bv(idx - 1) == cr)) {
```
RaasAhsan, line 156: Instead of checking whether the current index is a CR and the next index is a LF, do the opposite; we check if the current index is a LF and the previous index is a CR. This avoids an edge case where a CR is read at the end of a check so an LF isn't available yet, but the index is advanced nonetheless. This means we might do a bit more work when we see a newline, which may be more common in entities

build.sbt
```diff
@@ -219,6 +219,7 @@ lazy val emberCore = libraryProject("ember-core")
     libraryDependencies ++= Seq(
       log4catsTesting % Test,
     ),
+    fork in Test := true,
```
rossabaker, line 222: Hmm.  I thought this was already true everywhere.  Maybe it should be?  Would it make CI better or worse if it were? :thinking:

ember-core/src/test/scala/org/http4s/ember/core/StreamingParserSuite.scala
rossabaker: It may be a little esoteric for scalacheck, but it's cool.  It seems useful in fs2 applications, at least.  They don't publish a testkit, I don't think?

ember-core/src/test/scala/org/http4s/ember/core/StreamingParserSuite.scala
```diff
@@ -0,0 +1,198 @@
...
+          }
+        }
+
+      if (as.isEmpty) Gen.const(Nil) else Gen.delay(go(List(as), count))
+    }
+
+    def debug(segments: List[List[Byte]]): Unit = {
+      // handleErrorWith(e => {e.printStackTrace();IO(false)}).assert
+
+      println("----")
```
rossabaker, line 62: We just banished the `println`s from our tests.  Is this all temporary?

ember-core/src/test/scala/org/http4s/ember/core/StreamingParserSuite.scala
rossabaker: You mean to generate from arbitrary requests and responses to `List[List[Byte]]`?

RaasAhsan: Yeah exactly. Though per our Gitter conversation earlier, I'm interested in generating raw HTTP messages and then chunking that

ember-core/src/test/scala/org/http4s/ember/core/StreamingParserSuite.scala
RaasAhsan: AFAIK they don't

build.sbt
RaasAhsan: Apparently it wasn't. I need to revert this, but I had enabled it to turn off CE tracing since it was incorrectly messing with stack traces (need to fix that)

ember-core/src/test/scala/org/http4s/ember/core/StreamingParserSuite.scala
RaasAhsan: Yeah it's temporary, need to go back and remove

RaasAhsan:
BTW more forward looking, how crucial is it that the HTTP header section is parsed in a streaming fashion? I think we could eliminate complexity and even improve performance if we just buffer chunks until we hit the `\r\n\r\n`, then parse the request/status lines/headers. The entity body parsers remain the same

rossabaker:
The spec says:

> However, it is good practice to send
   header fields that contain control data first, such as Host on
   requests and Date on responses, so that implementations can decide
   when not to handle a message as early as possible.

But we're not passing a `Stream[F, Header]` to apps, and I don't think backends are implementing that logic.  The only thing I think it could do is help parse headers more in constant space, but headers shouldn't be large enough for that to be meaningful.  And we should be limiting header size before that becomes meaningful.  I think the buffering idea is fine.

RaasAhsan:
> The only thing I think it could do is help parse headers more in constant space

I'm also less concerned about it because the parser output size is linear with the input: the raw header names and values are resident in memory (at least, ATM, not sure if Fabio's PR changes anything).

rossabaker (approved):

PR #4431, opened 2021-02-15
Fixes #4294 Port prometheus to Scala 3.0.0-M3

### LINKED ISSUE #4294: Port prometheus-metrics to Scala 3.0.0-M3
Go to series/0.22 branch, run `++3.0.0-M3! prometheus-metrics/test`. There is one error.

ashwinbhaskar:


m-sp:
Seems this doesn't compile on 2.x have you tried a wildcard import instead?

m-sp (approved):

prometheus-metrics/src/main/scala/org/http4s/metrics/prometheus/PrometheusExportService.scala
```diff
@@ -25,7 +25,7 @@ import java.io.StringWriter
 
 import org.http4s.Uri.Path
 import org.http4s._
-import org.http4s.implicits.http4sLiteralsSyntax
+import org.http4s.Http4s._
```
rossabaker, line 28: Oh, wow, I forgot that exists.  That's from before Cats, and we might want to clean that up.  But, :+1: for what compiles now!

rossabaker (approved):

PR #4463, opened 2021-02-18
Merges, mostly #4429

rossabaker:
Compiles, but broken tests.


rossabaker:
Oh, hell.  Looks like they're broken on series/0.22, too.  That merge was easy.

rossabaker:
Crap.

rossabaker:
We're getting some `\r\r\n` in `hdr` coming out of this line:

```scala
                val (hdr, rem) = nh.splitAt(endOfheader + `\r\n`.size)
```

RaasAhsan:
This hunk seemed to not make it through https://github.com/http4s/http4s/blob/series/0.21/ember-core/src/main/scala/org/http4s/ember/core/ChunkedEncoding.scala#L46-L51

PR #4475, opened 2021-02-18
fix compilation

yanns:
@rossabaker this fixes the compilation but I'm really not sure. I flatMap the `Resource[F, (SocketAddress[IpAddress], Stream[F, Socket[F]])]` twice and I don't think it's good.

But maybe it will help you.


rossabaker:
This looks right to me.  I'll mop up the non-compiling examples.

PR #4476, opened 2021-02-18
update contributing

m-sp:


m-sp:
forgot it in the previous PR

m-sp:
should also fix https://github.com/http4s/http4s/issues/4469

Issue #3949, opened 2020-11-27
Hash and BoundedEnumerable instances for QValue


rossabaker:
A `QValue` can have a `Hash` and `BoundedEnumerable` instance.  The procedure is similar to #3902 and should be done from the series/0.21 branch.  This will give a tour of defining cats instances, how to avoid them clashing, and how to test for lawfulness.

rossabaker:
If we have more volunteers than tickets, we can spin off similar tickets to this.

jyoo980:
I'd like to give this a go!

rossabaker:
@jyoo980 Do you still have time for this or should I throw it back in the pool for LSUG this weekend?  Thanks!

jyoo980:
@rossabaker 

Sorry for the delay, I think I'm quite occupied at the moment, so it's better if anyone else who is interested can pick this up!

rossabaker:
Okay, no worries. And I'm sure we'll have more tickets like this in case you want one later. :smile:

Hiroki6:
I'd like to work on this issue.

Hiroki6:
Hi @rossabaker , I've made a PR for this issue: https://github.com/http4s/http4s/pull/4479

m-sp:
thanks again for your contribution!

PR #4479, opened 2021-02-18
Add Hash and BoundedEnumerable instances for QValue

Hiroki6:
For https://github.com/http4s/http4s/issues/3949

Please let me know if using `unapply` is not a good idea.

### DIFF
--- a/core/src/main/scala/org/http4s/QValue.scala
+++ b/core/src/main/scala/org/http4s/QValue.scala
@@ -21,7 +21,6 @@ import cats.{Hash, Order, Show}
 import org.http4s.internal.parboiled2.{Parser => PbParser}
 import org.http4s.parser.{AdditionalRules, Http4sParser}
 import org.http4s.util.Writer
-
 import scala.reflect.macros.whitebox
 
 /** A Quality Value.  Represented as thousandths for an exact representation rounded to three
--- a/core/src/main/scala/org/http4s/QValue.scala
+++ b/core/src/main/scala/org/http4s/QValue.scala
@@ -16,10 +16,12 @@
 
 package org.http4s
 
-import cats.{Order, Show}
+import cats.kernel.BoundedEnumerable
+import cats.{Hash, Order, Show}
 import org.http4s.internal.parboiled2.{Parser => PbParser}
 import org.http4s.parser.{AdditionalRules, Http4sParser}
 import org.http4s.util.Writer
+
 import scala.reflect.macros.whitebox
 
 /** A Quality Value.  Represented as thousandths for an exact representation rounded to three
@@ -72,11 +74,14 @@ final class QValue private (val thousandths: Int) extends AnyVal with Ordered[QV
 
       b.append(convert(mod10)) // Last digit
     }
+
 }
 
 object QValue {
   lazy val One: QValue = new QValue(1000)
   lazy val Zero: QValue = new QValue(0)
+
+  def unapply(qValue: QValue): Option[Int] = Some(qValue.thousandths)
 
   private def mkQValue(thousandths: Int, s: => String): ParseResult[QValue] =
     if (thousandths < 0 || thousandths > 1000)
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/http4s/QValue.scala
```diff
@@ -144,11 +148,44 @@ object QValue {
   @deprecated("""use qValue"" string interpolation instead""", "0.20")
   def q(d: Double): QValue = macro Macros.qValueLiteral
 
-  implicit val http4sOrderForQValue: Order[QValue] = Order.fromOrdering[QValue]
-  implicit val http4sShowForQValue: Show[QValue] = Show.fromToString[QValue]
-  implicit val http4sHttpCodecForQValue: HttpCodec[QValue] = new HttpCodec[QValue] {
-    def parse(s: String): ParseResult[QValue] = QValue.parse(s)
-    def render(writer: Writer, q: QValue): writer.type = q.render(writer)
+  implicit val catsInstancesForHttp4sQValue: Order[QValue]
```
rossabaker, line 151: Combining these is good, but it's going to break binary compatibility on version 0.21.  We should keep the old instances, make them not implicit, and deprecate them for version 0.21.20.  To avoid code duplication, I think they can just be aliases to this new combined instance.

core/src/main/scala/org/http4s/QValue.scala
```diff
@@ -72,12 +73,15 @@ final class QValue private (val thousandths: Int) extends AnyVal with Ordered[QV
...
       b.append(convert(mod10)) // Last digit
     }
+
 }
 
 object QValue {
   lazy val One: QValue = new QValue(1000)
   lazy val Zero: QValue = new QValue(0)
 
+  def unapply(qValue: QValue): Option[Int] = Some(qValue.thousandths)
```
rossabaker, line 83: I didn't think of this, but it seems good to me.

rossabaker (changes requested):
Thanks! This looks mostly on the right track.  Just a minor thing below:

rossabaker:
Oh, and it'll need a `scalafmtAll` too.  

You can test the MiMa issue locally with `mimaReportBinaryIssues`.  If it runs clean, you can disregard my other comment. :smile:

core/src/main/scala/org/http4s/QValue.scala
Hiroki6: understood. Thank you for your comment.
I fixed it.

Hiroki6:
> Oh, and it'll need a `scalafmtAll` too.
> 
> You can test the MiMa issue locally with `mimaReportBinaryIssues`. If it runs clean, you can disregard my other comment. smile

I ran `scalafmtAll` and `mimaReportBinaryIssues` locally.
Now all checks passed.

rossabaker (approved):
Looks great. Thanks!

Hiroki6:
@rossabaker 
Sorry, I noticed my mistake.
The test code of new instances has to be added `QValueSuite.scala`. 
But I added these into `QValueSpec.scala` which was an old file because my local repo was a little bit old and I took a mistake of local rebase.
What kind of process should I take?
Should I create a new PR for this?

rossabaker:
Oh, right. I noticed they were in the Spec when I did the merge to 0.22 and moved them to the Suite.  I guess we have some duplicate tests in 0.21 now between the Spec and the Suite.  If you do a PR to 0.21 to consolidate everything into QValueSuite, that'd be great.  It's not at all urgent: the tests are all there, and it's cleaned up on the later branches  Thanks!

Hiroki6:
> Oh, right. I noticed they were in the Spec when I did the merge to 0.22 and moved them to the Suite. I guess we have some duplicate tests in 0.21 now between the Spec and the Suite. If you do a PR to 0.21 to consolidate everything into QValueSuite, that'd be great. It's not at all urgent: the tests are all there, and it's cleaned up on the later branches Thanks!

@rossabaker Thank you for moving them to the Suite when you did the merge to 0.22.
I've made a PR to 0.21 to fix it.

PR #4490, opened 2021-02-19
delete QValueSpec and consolidate all into QValueSuite

Hiroki6:
For https://github.com/http4s/http4s/issues/3949

I made a mistake in the previous PR, https://github.com/http4s/http4s/pull/4479, that I added some test code into QValueSpec. But this file was already renamed into QValueSuite, so I consolidate the test code into QValueSuite and delete QValueSpec again.


m-sp (approved):
thanks for following up on this!

rossabaker (approved):

PR #4491, opened 2021-02-19
cleanup scalafix build

m-sp:
looking for some feedback, this is a bit of a change in the setup for scalafix.
it's modeled after what cats and [cats-effect](https://github.com/typelevel/cats-effect/pull/1686/files)

this disables the publishing but the scalafix can still be run via `sbt ";scalafixEnable ;scalafixAll github:http4s/http4s/series/0.22"` but would enable any dotty plans

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,7 +27,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        scala: [2.12.13, 2.13.4, 3.0.0-M3]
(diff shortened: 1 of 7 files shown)


rossabaker (commented):
This is closer to how it used to work, and I don't recall why we switched away.  

If it's no longer an aggregate, it's going to be easier to neglect to run it before CI, but CI should still keep it working.  But it solves the problem of it not keeping up with the latest Dotty versions while we're in such churn.

If this is how other successful projects are using it, I'm up for it.

PR #4492, opened 2021-02-19
enable dotty in ci

m-sp:
based on https://github.com/http4s/http4s/pull/4491

disabled twirl in examples. twirl depends on scala xml. so we have another 2.13/3.0 conflict
temporarily inlined the workflow description to work around scalafmt issues

there are some rather hacky changes which need cleanup once all dependencies are ready for dotty but this get's us real close to actually publishing a dotty version


rossabaker (approved):

rossabaker:
This is going to break main for a bit until we merge forward, but we can disable Dotty again there until #4471 is done.

PR #4505, opened 2021-02-22
Scala 3.0.0-M3 -> Scala-3.0.0-RC1

rossabaker:


m-sp:
Seems circe only supports one version of dotty...

rossabaker:
I think we should probably just give up on RC3.  It's a core dependency, and I don't want to bother with cross-versioning dependencies.

rossabaker:
Core repo dependency, that is, not http4s-core module.  It'll be here after the Great Schism.

rossabaker:
I guess this is the 3.0.0-RC1 branch now.

build.sbt
```diff
@@ -46,8 +46,7 @@ enablePlugins(SonatypeCiReleasePlugin)
 versionIntroduced := Map(
   // There is, and will hopefully not be, an 0.22.0. But this hushes
   // MiMa for now while we bootstrap Dotty support.
-  "3.0.0-M2" -> "0.22.0",
-  "3.0.0-M3" -> "0.22.0",
+  "3.0.0-RC1" -> "0.22.0",
```
m-sp, line 49:
```suggestion
  scala_3 -> "0.22.0",
```

m-sp (approved):

PR #4509, opened 2021-02-23
Entity Rethinking

ChristopherDavenport:
Makes Entity more top-class. 

Inference gets worse from time to time and I'd like to hammer that out.
Some tests are failing.

Its surprising how many things can ignore the change and just use entity.body if thats all they cared about. But anyone who does care can get extra benefits.


core/src/main/scala/org/http4s/Entity.scala
```diff
@@ -16,19 +16,58 @@
...
+  def translate[G[_]](fk: F ~> G): Entity[G]
+  def covary[G[x] >: F[x]]: Entity[G] = translate[G](cats.arrow.FunctionK.id[F].widen[G]) 
+}
 
 object Entity {
+
+  case class Strict[F[_]](chunk: Chunk[Byte]) extends Entity[F]{
+    def body: Stream[F, Byte] = Stream.chunk(chunk)
+    def length: Option[Long] = chunk.size.toLong.some
+    def translate[G[_]](fk: F ~> G): Entity[G] = Strict(chunk)
```
rossabaker, line 35: Can't we just cast this instead of allocating a new one?

core/src/main/scala/org/http4s/Entity.scala
```diff
@@ -16,19 +16,58 @@
...
+    def translate[G[_]](fk: F ~> G): Entity[G] = Chunked(body.translate(fk))
+  }
+
+  case class Empty[F[_]]() extends Entity[F]{
+    def body: Stream[fs2.Pure,Byte] = Stream.empty
+    def length: Option[Long] = Some(0L)
+    def translate[G[_]](fk: F ~> G): Entity[G] = Empty()
+  }
+
+  def empty[F[_]] = Empty[F]()
```
rossabaker: You could share an instance of this.

core/src/main/scala/org/http4s/Entity.scala
```diff
@@ -16,19 +16,58 @@
 
 package org.http4s
 
-import cats.Monoid
+import fs2.{Stream, Chunk}
+import cats._
 import cats.syntax.all._
 
-final case class Entity[+F[_]](body: EntityBody[F], length: Option[Long] = None)
+sealed trait Entity[F[_]]{
```
rossabaker, line 23: Did covariance cause a problem?

core/src/main/scala/org/http4s/EntityDecoder.scala
```diff
@@ -189,7 +189,7 @@ object EntityDecoder {
 
   /** Helper method which simply gathers the body into a single Chunk */
   def collectBinary[F[_]: Concurrent](m: Media[F]): DecodeResult[F, Chunk[Byte]] =
-    DecodeResult.success(m.body.chunks.compile.toVector.map(bytes => Chunk.concat(bytes)))
+    DecodeResult.success(m.entity.body.chunks.compile.toVector.map(bytes => Chunk.concat(bytes)))
```
rossabaker, line 192: Can you keep a `body` on `Message`?  A lot less would break.  It's `withBodyStream` that's now bad.

core/src/main/scala/org/http4s/Message.scala
```diff
@@ -77,35 +77,9 @@ sealed trait Message[F[_]] extends Media[F] { self =>
...
-    }
-    change(body = entity.body, headers = headers ++ hs)
+    change(entity = w.toEntity(b))
   }
 
-  /** Sets the entity body without affecting headers such as `Transfer-Encoding`
-    * or `Content-Length`. Most use cases are better served by [[withEntity]],
-    * which uses an [[EntityEncoder]] to maintain the headers.
-    */
-  def withBodyStream(body: EntityBody[F]): Self =
```
rossabaker, line 100: We might consider a deprecation cycle on this one to help people migrate.

core/src/main/scala/org/http4s/Message.scala
```diff
@@ -77,35 +77,9 @@ sealed trait Message[F[_]] extends Media[F] { self =>
...
-    change(body = body)
-
-  /** Set an empty entity body on this message, and remove all payload headers
-    * that make no sense with an empty body.
-    */
-  def withEmptyBody: Self =
```
rossabaker, line 106: This one is a more direct translation to the new, and could certainly survive as deprecation.

core/src/main/scala/org/http4s/StaticFile.scala
```diff
@@ -106,7 +103,11 @@ object StaticFile {
                 Some(
                   Response(
                     headers = headers,
-                    body = readInputStream[F](F.pure(inputStream), DefaultBufferSize)
+                    entity = {
+                      val body =readInputStream[F](F.pure(inputStream), DefaultBufferSize)
+                      if (len >= 0) Entity.TrustMe(body, len)
```
rossabaker, line 108: Something that has always bothered me: what if the file gets truncated while we're reading it?

tests/src/test/scala/org/http4s/EntityEncoderSpec.scala
```diff
@@ -48,20 +48,21 @@ class EntityEncoderSpec extends Http4sSuite {
...
-        EntityEncoder.encodeBy[IO, Foo](`Transfer-Encoding`(TransferCoding.gzip))(_ => Entity.empty)
+        EntityEncoder.encodeBy[IO, Foo](`Transfer-Encoding`(TransferCoding.gzip))(_ => Entity.Empty[IO]())
       implicitly[EntityEncoder[IO, Stream[IO, Foo]]].headers.get(`Transfer-Encoding`) match {
         case Some(coding: `Transfer-Encoding`) =>
           assertEquals(coding, `Transfer-Encoding`(TransferCoding.gzip, TransferCoding.chunked))
         case _ => fail("Match failed")
       }
     }
 
+/* Invalid With New Model What do we want to do about them?
```
rossabaker: Isn't the equivalent to this one a chunked body with an empty stream?

rossabaker (commented):
Wow, this holds up pretty well!

I think if you keep body, you could make this diff a lot smaller.  Maybe we don't want it anymore, but then we should at least deprecate it.

core/src/main/scala/org/http4s/Entity.scala
ChristopherDavenport: Yes. The Monoid disagreed very strongly with covariance.

core/src/main/scala/org/http4s/Entity.scala
ChristopherDavenport: Yes. I just hate writing casts on a first draft where things aren't yet compiling right.

core/src/main/scala/org/http4s/Message.scala
ChristopherDavenport: The difficulty with it is what the scala doc says. There is now no way to not effect the headers like that.

core/src/main/scala/org/http4s/Message.scala
ChristopherDavenport: Yeah. This can survive.

core/src/main/scala/org/http4s/StaticFile.scala
ChristopherDavenport: Betrayed trust.

tests/src/test/scala/org/http4s/EntityEncoderSpec.scala
ChristopherDavenport: The thing is that this is checking the tranfer-encoding header for how it interacted with chunked. Which we no longer do in the EntityEncoder.

core/src/main/scala/org/http4s/EntityDecoder.scala
ChristopherDavenport: body is back, but deprecated.

server/src/main/scala/org/http4s/server/middleware/DefaultHead.scala
```diff
@@ -45,5 +49,10 @@ object DefaultHead {
     apply(httpRoutes)
 
   private[this] def drainBody[G[_]: Concurrent](response: Response[G]): Response[G] =
-    response.copy(body = response.body.interruptWhen[G](Stream(true)).drain)
+    response.copy(entity = response.entity match {
+      case Strict(chunk) => TrustMe(Stream.empty, chunk.size.toLong)
```
rossabaker, line 53: :face_with_head_bandage:

Might `HeadBody` be a thing?  Or `DrainingBody`?  Then you've got to deal with it everywhere else, though.

core/src/main/scala/org/http4s/Message.scala
```diff
@@ -220,7 +213,7 @@ final class Request[F[_]](
     val uri: Uri = Uri(path = Uri.Path.Root),
     val httpVersion: HttpVersion = HttpVersion.`HTTP/1.1`,
     val headers: Headers = Headers.empty,
-    val body: EntityBody[F] = EmptyBody,
+    val entity: Entity[F] = Entity.Empty(),
```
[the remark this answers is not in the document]
ChristopherDavenport: All of these were from before I had introduced that syntax.

core/src/main/scala/org/http4s/Entity.scala
```diff
@@ -16,19 +16,64 @@
 
 package org.http4s
 
-import cats.Monoid
+import fs2.{Chunk, Stream}
+import cats._
 import cats.syntax.all._
 
-final case class Entity[+F[_]](body: EntityBody[F], length: Option[Long] = None)
+sealed trait Entity[F[_]] {
```
RaasAhsan, line 23: One of the things I've been thinking about lately is whether we could support more efficient file serving. The only way to send a file (or a region of a file) with the current http4s model is to read it from disk, copy it into memory, then write it back out to a socket again. `sendfile` (or `FileChannel.transferTo`) lets you completely circumvent the userspace and copying by writing the file directly to the outgoing socket in kernel space. With the proposed model, I think there is an avenue for more efficiency.

I think we can pass in a `FileChannel` or fs2 `FileHandle` and use that in different ways on different backends. For backends we control (like blaze and ember), we can use `transferTo` to write the file more efficiently. For other backends, we could do the regular pattern of streaming the file in chunks. We know the size of the file so we could do a content-length/chunked/multipart encoding. 

One issue I can see with that is that it can introduce a `Sync`/`Blocker`/`Files` constraint on certain backends (particularly Ember) for file operations. FS2 also doesn't support the `transferTo` operation so we would need to look into feasibility. General thoughts on the approach though?

RaasAhsan:
One of the things I've been thinking about lately is whether we could support more efficient file serving. The only way to send a file (or a region of a file) with the current http4s model is to read it from disk, copy it into memory, then write it back out to a socket again. sendfile (or `FileChannel.transferTo`) lets you completely circumvent the userspace and copying by writing the file directly to the outgoing socket in kernel space. With the proposed model, I think there is an avenue for more efficiency.

I think we can pass in a `FileChannel` or fs2 `FileHandle` and use that in different ways on different backends. For backends we control (like blaze and ember), we can use `transferTo` to write the file more efficiently. For other backends, we could do the regular pattern of streaming the file in chunks. We know the size of the file so we could do a content-length/chunked/multipart encoding.

One issue I can see with that is that it can introduce a `Sync`/`Blocker`/`Files` constraint on certain backends (particularly Ember) for file operations. FS2 also doesn't support the transferTo operation so we would need to look into feasibility. General thoughts on the approach though?

rossabaker:
This is exactly the zero-copy I was discussing with @ChristopherDavenport.  I couldn't remember the name of the NIO API, nor did I know whether fs2 was exploiting it.

ember-core/src/main/scala/org/http4s/ember/core/Encoder.scala
```diff
@@ -98,30 +123,52 @@ private[ember] object Encoder {
...
-        chunked = true
+      if (!NoPayloadMethods.contains(req.method)) {
+        req.entity match {
+          case Strict(chunk) =>
+            val length = `Content-Length`.unsafeFromLong(chunk.size.toLong)
+            stringBuilder
+              .append(length.renderString)
+              .append(CRLF)
+          case TrustMe(_, size) =>
+            val length = `Content-Length`.unsafeFromLong(size.toLong)
```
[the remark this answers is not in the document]
ChristopherDavenport: I am on the fence. It should be positive, and array.length will never be negative. But someone could write something pathological.

ChristopherDavenport:
I would love to see it added, but we might want to get support into fs2? I worry about its introduction until we are certain how it would work. How would a middleware transition a File based entity?

RaasAhsan:
Yup I'll take it upstream and see what an API for it could look like

> How would a middleware transition a File based entity?

Could you elaborate on that ? I figured middlewares won't/shouldn't read the streams

rossabaker:
Several middlewares transform the request or response body.  Logger observes, GZip uncompresses, ...

core/src/main/scala/org/http4s/Entity.scala
```diff
@@ -16,19 +16,64 @@
...
 object Entity {
+
+  case class Strict[F[_]](chunk: Chunk[Byte]) extends Entity[F] {
+    def body: Stream[F, Byte] = Stream.chunk(chunk)
+    def length: Option[Long] = chunk.size.toLong.some
+    def translate[G[_]](fk: F ~> G): Entity[G] = this.asInstanceOf[Strict[G]]
+  }
+
+  case class TrustMe[F[_]](body: Stream[F, Byte], size: Long) extends Entity[F] {
+    def length: Option[Long] = size.some
```
isomarcte, line 39: Should we enforce that the size lines up with the stream size by doing `body.take(size)`? I think that would make the behavior more consistent for both client and server implementations.

I suppose the question is whether or not we ever want to permit reading/writing a body which is larger than the declared size. Is that ever valuable?

isomarcte: OTOH we don't really gain much by enforcing that either I suppose.

rossabaker: Question is, what do we do about it?  Either an underflow or overflow will foul up the next message on a persistent connection.

core/src/main/scala/org/http4s/Entity.scala
RaasAhsan, line 23: Thoughts on leaving `Entity` unsealed? I think we have to do this now rather than later, but two reasons:
1. We can add new entity forms later which may have a more efficient implementation. As long as they provide a `body` it should be usable by any backend.
2. Different backends may be able to process different entities in more efficient ways, with the various subtypes serving as hints. e.g. maybe some backend can serve files more efficiently than others, but we can always fall back to `body`. Furthermore, backends could define their own entity types

ChristopherDavenport: So the reason I don't like an unsealed one is it means that this change improves the middleware layer not at all. As they don't know about them, so they can't leverage anything that is written in those types.

core/src/main/scala/org/http4s/Entity.scala
```diff
@@ -16,19 +16,64 @@
...
+sealed trait Entity[F[_]] {
+  def body: Stream[F, Byte]
+  def length: Option[Long]
+  def translate[G[_]](fk: F ~> G): Entity[G]
+  def covary[G[x] >: F[x]]: Entity[G] = translate[G](cats.arrow.FunctionK.id[F].widen[G])
+}
 
 object Entity {
+
+  case class Strict[F[_]](chunk: Chunk[Byte]) extends Entity[F] {
```
ScalaWilliam, line 32: Should these case classes be `final` or no?

core/src/main/scala/org/http4s/Entity.scala
rossabaker: You mean that with open types, middleware are going to have a hard time preserving those types?  With a self type and enough methods, this can be mitigated, but never entirely, and it would be uncharacteristically OOPy of us.

I guess I'd be interested in what might be missing besides a file and a HEAD.

core/src/main/scala/org/http4s/Entity.scala
rossabaker: I think yes.  We made a pass through these with wartremover a long time ago, and I don't think any linter catches it now, and I usually forget myself.

RafalSumislawski:
Hello folks

I'm excited about the performance improvements that the proposed redesign could bring to http4s. I'm worried by the fact that there's been no progress in this PR since March, so I was thinking, that maybe I could offer help, and take it over? At the very least I could try resolving all the conflicts and working through some of the suggestions made by the reviewers. What do you think @ChristopherDavenport?

armanbilge:
> I'm worried by the fact that there's been no progress in this PR since March

> At the very least I could try resolving all the conflicts

@RafalSumislawski At least some of the blame for that falls to me, for creating many of those conflicts during the Scala.js cross. @ChristopherDavenport paused his efforts here when I started my PR.

rossabaker:
I think the ground is mostly stabilized once the above merge conflicts are resolved.

bplommer:
Are there bits of this that can be merged incrementally, like the addition of `entity` and deprecation of `body` fields?

bplommer:
> Inference gets worse from time to time and I'd like to hammer that out.

Following #5229 I've been playing with making `Message` types covariant in `F` and that seems to hugely improve type inference - it becomes very rarely necessary to specify the effect type. Maybe that would help with the issues here too.

core/src/main/scala/org/http4s/Entity.scala
bplommer: Do you remember what the issue was? I've tried making it covariant on a local branch and there wasn't a problem.

bplommer: Oh, I'm seeing the problem now - not sure why I didn't before.

ChristopherDavenport:
This has been superseded at this point I believe.

armanbilge:
👍 let's close the PR but keep the branch in case we need to come back for anything we forgot :)

PR #4516, opened 2021-02-25
Fix Routes Constraint

ChristopherDavenport:
Defer is no longer in Sync hierarchy. This is definitely correct for any F that has Concurrent, but will work fine in most cases that are actually used without a messy api.


rossabaker (approved):
This looks overconstrained to me, but if I understand it right, it's more pragmatic in CE3.

PR #4519, opened 2021-02-25
upgrade scala 2.13.5 remove unused boundaries

m-sp:
probably should be on 21 but appeasing mama is a horrorshow


rossabaker:
I tried a cherry-pick, and it already doesn't pick cleanly.  We can just leave that branch stuck on 2.13.4.

PR #4533, opened 2021-02-27
Accept-Encoding cleanup

rossabaker:
This is the general pattern for removing deprecation from recurring headers:

1. In companion, stop extending `HeaderKey.Internal` and `HeaderKey.Recurring`
2. Add a variadic `.apply` constructor
3. `parse` is no longer interface, so remove `override`
4. In case class, stop extending `Header`
5. Remove `key` and `Value`
6. Remove any reference in `HttpHeaderParser`, which is going away.
7. Stop extending `HttpHeaderParserHelper`
8. Fix any other compilation errors.


rossabaker:
More conflicts than code.

PR #4534, opened 2021-02-27
Age cleanup

rossabaker:
1. Remove `HeaderKey.Internal` and `HeaderKey.Singleton` from the companion
2. Remove `override` declaration from `parse`
3. Remove `Header.Parsed` from the case class
4. Clean up `value`. Your goal is to return a `String`.
5. Remove `key` and `renderValue`. (`renderValue` may be useful in `value`, but it's no longer API.)
6. Remove any reference in `HttpHeaderParser`, which is going away.


rossabaker:
Handled already.

Issue #4535, opened 2021-02-27
Fix deprecations from header refactor


rossabaker:
The header refactor in #4415 was done with a temporary "v2" package.  We need to clear out references to the old.

These are the files with warnings.  For new contibutors, headers will be the easiest.  There are patterns to follow in #4533 (recurring) and #4534 (singleton).

All this work needs to happen on series/0.22.  We'll be accepting volunteers at the LSUG event on Feburary 27 to kick this off.

[March 2: resetting the list]

* [x]  HeaderEchoSuite.scala 2 server/src/test/scala/org/http4s/server/middleware/ (@rossabaker)
* [x]  HeaderLaws.scala 1 tests/src/test/scala/org/http4s/headers/ [@phongngtuan]
* [x]  HeaderParserSuite.scala 3 tests/src/test/scala/org/http4s/parser/ (@rossabaker)
* [x]  HeaderSuite.scala 15 tests/src/test/scala/org/http4s/
* [x]  HeadersSpec.scala 25 tests/src/test/scala/org/http4s/
* [x]  Http1ServerStageSpec.scala 9 blaze-server/src/test/scala/org/http4s/server/blaze/
* [x]  HttpHeaderParser.scala 55 core/src/main/scala/org/http4s/parser/ (@rossabaker)
* [x]  HttpHeadersBench.scala 9 bench/src/main/scala/org/http4s/bench/ (@rossabaker)
* [x]  Last-Modified.scala 7 core/src/main/scala/org/http4s/headers
* [x]  Location.scala 7 core/src/main/scala/org/http4s/headers/
* [x]  LocationHeaderSuite.scala 11 tests/src/test/scala/org/http4s/parser/
* [x]  Max-Forwards.scala 7 core/src/main/scala/org/http4s/headers/
* [x]  MediaTypeSuite.scala 3 tests/src/test/scala/org/http4s/ (@rossabaker)
* [x]  Origin.scala 7 core/src/main/scala/org/http4s/headers/
* [x]  OriginHeaderSuite.scala 19 tests/src/test/scala/org/http4s/parser/
* [x]  Proxy-Authorization.scala 7 core/src/main/scala/org/http4s/headers/
* [x]  Range.scala 7 core/src/main/scala/org/http4s/headers/
* [x]  Referer.scala 7 core/src/main/scala/org/http4s/headers/
* [x]  ResponseParser.scala 7 blaze-core/src/test/scala/org/http4s/blazecore/
[...]

PR #4539, opened 2021-02-27
Accept header cleanup

VlachJosef:
partially address #4535


rossabaker (approved):
:+1: Thanks for coming out to LSUG today!

PR #4540, opened 2021-02-27
Move constants to header v2 to remove deprecated warnings

phongngtuan:
partially address #4535


rossabaker (approved):
:+1: Thanks for helping out today at LSUG!

PR #4541, opened 2021-02-27
Migrate blaze-server tests

rossabaker:


rossabaker:
Tests fail on `Set` to `Headers`.  A notion of `Headers` equality that ignores order between headers, but enforces them within, would be achievable, correct, and prevent brittle tests like these.

rossabaker:
Handled better in #4555.

PR #4542, opened 2021-02-27
Access-Control-Allow-Headers header cleanup

VlachJosef:
partially address #4535

### DIFF
--- a/core/src/main/scala/org/http4s/headers/Access-Control-Allow-Headers.scala
+++ b/core/src/main/scala/org/http4s/headers/Access-Control-Allow-Headers.scala
@@ -20,14 +20,10 @@ package headers
 import org.typelevel.ci.CIString
 import org.http4s.internal.parsing.Rfc7230
 import cats.data.NonEmptyList
-import org.http4s.util.Renderer
 
-object `Access-Control-Allow-Headers`
-    extends HeaderKey.Internal[`Access-Control-Allow-Headers`]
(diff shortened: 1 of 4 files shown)


core/src/main/scala/org/http4s/headers/Access-Control-Allow-Headers.scala
```diff
@@ -20,14 +20,10 @@ package headers
...
 import org.http4s.internal.parsing.Rfc7230
 import cats.data.NonEmptyList
-import org.http4s.util.Renderer
 
-object `Access-Control-Allow-Headers`
-    extends HeaderKey.Internal[`Access-Control-Allow-Headers`]
-    with HeaderKey.Recurring {
-
-  override def parse(s: String): ParseResult[`Access-Control-Allow-Headers`] =
-    ParseResult.fromParser(parser, "Invalid Access-Control-Allow-Headers headers")(s)
```
rossabaker, line 30: Something to think about: we're losing consistency on whether this is available on the companion objects, or only via the Header instance.

rossabaker (approved):

core/src/main/scala/org/http4s/headers/Access-Control-Allow-Headers.scala
VlachJosef: Yes, removing this is perhaps not the best idea for ease of migration, so I'll put this method back.

PR #4545, opened 2021-02-28
Cache-Control header cleanup and new HeaderParserHelper trait

VlachJosef:
partially address #4535


VlachJosef:
Only two `Recurring` headers remains to be migrated:

* Connection
* Transfer-Encoding

They are both used by `ServerTestRoutes`, and since migration of `ServerTestRoutes` started in #4547 I'll wait with their migration until both branches are merged.

PR #4546, opened 2021-02-28
Content-Disposition Content-Location warns clean up
 


vder:
partially address #4535


vder:
Hi, could anyone take a look here as I'm not sure what should I do about it. As far as I understand one of the test failed, since it raised exception instead of getting actual value:
image
I've tried to switch to this specific jvm version to reproduce it locally, but it has just worked fine.

rossabaker (approved):
There are a few flaky tests in blaze on CI.  We've been more aggressively marking them as flaky, but that looks like another one.  I'll mark that as well.

core/src/main/scala/org/http4s/headers/Content-Disposition.scala
```diff
@@ -90,11 +88,10 @@ object `Content-Disposition`
...
 
 // see http://tools.ietf.org/html/rfc2183
-final case class `Content-Disposition`(dispositionType: String, parameters: Map[String, String])
-    extends Header.Parsed {
-  override def key: `Content-Disposition`.type = `Content-Disposition`
-  override lazy val value = super.value
-  override def renderValue(writer: Writer): writer.type = {
+final case class `Content-Disposition`(dispositionType: String, parameters: Map[String, String]) {
+  def key: `Content-Disposition`.type = `Content-Disposition`
+  lazy val value = v2.Header[`Content-Disposition`].value(this)
```
rossabaker, line 93: It's a little weird to depend on our type class instance here.  I think it would be best to flip that dependency.  I'm also not convinced it still needs to be lazy.  But this is forward progress as is, so, :+1:.

vder: Agree, It just I've seen this value field in several other header classes and propagated it here.  But I suppose we should rather removing these than adding ones.

PR #4547, opened 2021-02-28
Migrate content length header & co

phongngtuan:
Following up the session led by @rossabaker, partially address #4535 
- Add syntax for Header & Select
- Add implicit from `Select[A]` to `Renderer[A] `
- Migrate some of headers

Only the first commit is more or less interesting where the machinery are introduced. Really appreciate if you could take a look before I stray too far from the path. In particular
- `SimpleHeadersSpec` used to take key + value but now only works on value.
- `HeadersSpec`

todo:
- migrate the rest of the headers
- revisit the headers classes to clean up unused method/val (key, name,...) that was from inheritance. I left those as is since I don't know if anyone is using those


core/src/main/scala/org/http4s/util/Renderable.scala
```diff
@@ -93,6 +96,11 @@ object Renderer {
       override def render(writer: Writer, values: Set[T]): writer.type =
         writer.addSet(values)
     }
+
+  implicit def headerSelectRenderer[A](implicit select: v2.Header.Select[A]): Renderer[A] =
```
rossabaker, line 100: I think this is the only case where an http4s model type appears in this file.  If it works putting this on `Select` instead of `Renderer`, I think that would be preferrable.  I also just want to get this all compiling again, so I'm content to revisit this later.

PR #4548, opened 2021-02-28
Add 'Running site locally' section into Contributing guide

VlachJosef:
When trying to run website/docs site locally with latest hugo, site is broken. Users noticed this already [three years ago](https://gitter.im/http4s/http4s?at=5af21a03b37eab7d04797c45).

I'm not sure if there are any plans to migrate docs/website to the latest hugo version, but hopefully this description will save time in figuring out that latest version of hugo is not supported.

### DIFF
--- a/docs/src/hugo/config.toml
+++ b/docs/src/hugo/config.toml
@@ -1,4 +1,4 @@
-baseurl = "https://http4s.org/v0.20"
+baseurl = "https://http4s.org/v0.22"
 contentDir = "../../target/mdoc"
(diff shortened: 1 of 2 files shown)


rossabaker:
I am currently in favor of migrating the documentation to Paradox, so there are no dependencies outside the JVM.  I'm sinking my time into the Dotty and CE3 releases first, but a documentation push should be our next big effort.

The Nix shell sets up a working version of Hugo, but most of our users aren't running Nix, nor is it an expectation.  I think this a good interim note until we modernize it. :+1:

PR #4550, opened 2021-03-01
make progress on dotty warnings

m-sp:


ember-core/src/main/scala/org/http4s/ember/core/Parser.scala
```diff
@@ -133,49 +133,48 @@ private[ember] object Parser {
       var start = initStart
 
       while (!complete && idx < bv.size) {
-        (state: @switch) match {
```
m-sp, line 136: switch statements with fewer than 4 cases are never optimised, so we might as well do a boolean and an if

ember-core/src/main/scala/org/http4s/ember/core/Parser.scala
```diff
@@ -103,7 +103,7 @@ private[ember] object Parser {
         bv: Array[Byte],
         accHeaders: List[v2.Header.Raw],
         idx: Int,
-        state: Byte,
+        state: Boolean,
```
rossabaker, line 106: A "state" of boolean seems like a big smell to me, but we can sort that out later.

rossabaker (approved):

PR #4552, opened 2021-03-02
Port headerLaws to use Header typeclass

phongngtuan:
Existing header laws can be changed to use new one by replacing
parentheses, for example
```
checkAll("Expires", headerLaws(Expires))
```
to

```
checkAll("Expires", headerLaws[Expires])
```

partially addresses #4535


rossabaker (approved):
Ah, excellent.  Good to have these back.

PR #4556, opened 2021-03-02
Header migration

VlachJosef:
There was custom extractor on `Set-Cookie` header which is removed in this PR. It wasn't covered by test, so I'm not sure how much important it was.

It was removed since it depends on `matchHeader` from old `HeaderKey`.

Partially addresses #4535

### DIFF
--- a/core/src/main/scala/org/http4s/Headers.scala
+++ b/core/src/main/scala/org/http4s/Headers.scala
@@ -78,7 +78,7 @@ final class Headers private (private val headers: List[Header]) extends AnyVal {
       this.headers.foreach { orig =>
         orig.parsed match {
           case _: Header.Recurring => acc += orig
-          case _: `Set-Cookie` => acc += orig
+          // case _: `Set-Cookie` => acc += orig
           case h if !hs.exists(_.name == h.name) => acc += orig
           case _ => // NOOP, drop non recurring header that already exists
         }
--- a/core/src/main/scala/org/http4s/headers/Referer.scala
+++ b/core/src/main/scala/org/http4s/headers/Referer.scala
@@ -22,7 +22,8 @@ import org.typelevel.ci.CIString
 
 object Referer {
 
-  def parse(s: String): ParseResult[Referer] = headerInstance.parse(s)
+  def parse(s: String): ParseResult[Referer] =
+    ParseResult.fromParser(parser, "Invalid Referer header")(s)
 
   private[http4s] val parser = Uri.Parser
     .absoluteUri(StandardCharsets.ISO_8859_1)
@@ -33,7 +34,7 @@ object Referer {
     v2.Header.createRendered(
       CIString("Referer"),
       _.uri,
-      ParseResult.fromParser(parser, "Invalid Referer header")
+      parse
     )
 
 }
--- a/core/src/main/scala/org/http4s/headers/Referer.scala
+++ b/core/src/main/scala/org/http4s/headers/Referer.scala
@@ -17,13 +17,13 @@
 package org.http4s
 package headers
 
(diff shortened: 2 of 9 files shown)


core/src/main/scala/org/http4s/Headers.scala
```diff
@@ -78,7 +78,7 @@ final class Headers private (private val headers: List[Header]) extends AnyVal {
       this.headers.foreach { orig =>
         orig.parsed match {
           case _: Header.Recurring => acc += orig
-          case _: `Set-Cookie` => acc += orig
+          // case _: `Set-Cookie` => acc += orig
```
rossabaker, line 81: This is a special case in the spec and needed to be a special case in the old design, but it is covered by the new one.

core/src/main/scala/org/http4s/headers/Referer.scala
```diff
@@ -17,13 +17,13 @@
...
-import org.http4s.util.Writer
 import java.nio.charset.StandardCharsets
 import org.typelevel.ci.CIString
 
-object Referer extends HeaderKey.Internal[Referer] with HeaderKey.Singleton {
-  override def parse(s: String): ParseResult[Referer] =
-    ParseResult.fromParser(parser, "Invalid Referer")(s)
+object Referer {
+
+  def parse(s: String): ParseResult[Referer] = headerInstance.parse(s)
```
rossabaker, line 25: Stylistically, I would go the other direction here: make these the `ParseResult.fromParser` calls, and just put `parse` at the third argument in the instance.

core/src/main/scala/org/http4s/headers/Server.scala
```diff
@@ -51,21 +51,11 @@ object Server extends HeaderKey.Internal[Server] with HeaderKey.Singleton {
       ParseResult.fromParser(parser, "Invalid Server header")
     )
 
+  val name = headerInstance.name
```
rossabaker, line 54: This is another case like `parse`.

rossabaker (commented):
Looks great, though I'd invert the companion <-> instance relationship.

core/src/main/scala/org/http4s/Headers.scala
rossabaker: To be clear, no further action needed on this one.  We'll hopefully be removing this class entirely soon.

core/src/main/scala/org/http4s/headers/Referer.scala
VlachJosef: Ok, I'll change it

VlachJosef: Resolved

core/src/main/scala/org/http4s/headers/Server.scala
VlachJosef: Resolved

rossabaker (approved):
Great! :+1: after green

tests/src/test/scala/org/http4s/headers/RefererSuite.scala
```diff
@@ -18,10 +18,11 @@ package org.http4s
 package headers
 
 import cats.effect.IO
+import org.http4s.syntax.header._
 import org.http4s.laws.discipline.ArbitraryInstances._
 
 class RefererSuite extends HeaderLaws {
-  checkAll("Referer", headerLaws(`Retry-After`))
+  //checkAll("Referer", headerLaws(`Retry-After`))
```
m-sp, line 25: if you merge series/0.22 into your branch this should work again because of https://github.com/http4s/http4s/pull/4552

m-sp (approved):

PR #4558, opened 2021-03-02
more headers

vder:
Partially addresses #4535

the following headers were touched:
Last-Modified 
Location 
Max-Forwards
Proxy-Authorization 
Range    

Range header contains some additional code and I left it as it be, since I wasn't sure what to do with it.

Edit: I've added the Origin header as well, but I'm not sure if that Header instance is ok.


rossabaker (approved):
We went back and forth on whether `parse` belongs on the companion object, and ended up with a mixed verdict.  I'd like to come back and make that consistent, but I'm more interested in getting it compiling first. :+1:

rossabaker:
Oops.  I broke this handling the merge conflict.  My fault.

PR #4562, opened 2021-03-02
Check header laws for Referer

rossabaker:


tests/src/test/scala/org/http4s/headers/RefererSuite.scala
```diff
@@ -22,7 +22,7 @@ import org.http4s.syntax.header._
 import org.http4s.laws.discipline.ArbitraryInstances._
 
 class RefererSuite extends HeaderLaws {
-  //checkAll("Referer", headerLaws(`Retry-After`))
+  checkAll("Referer", headerLaws(`Retry-After`))
```
m-sp, line 25:
```suggestion
  checkAll("Referer", headerLaws[`Retry-After`])
```

rossabaker: Ugh, yes.  This was an in-browser drive-by.

PR #4563, opened 2021-03-02
Header migration cleanup

VlachJosef:
Removing temporal functionality (which was replaced by header syntax) and making header more consistent.


rossabaker (approved):
Oh, thank you! This is taking shape nicely.

PR #4573, opened 2021-03-02
Remove HttpHeaderParser

rossabaker:
Fare thee well, global mutable state.


core/src/main/scala/org/http4s/Header.scala
```diff
@@ -82,12 +82,7 @@ object Header {
     */
   @deprecated("Condemned", "LSUG")
   final case class Raw(name: CIString, override val value: String) extends Header {
-    private[this] var _parsed: Header = null
-    final override def parsed: Header = {
-      if (_parsed == null)
-        _parsed = parser.HttpHeaderParser.parseHeader(this).getOrElse(this)
-      _parsed
-    }
+    final override def parsed: Header = null // I told you it was condemned
```
ChristopherDavenport, line 85: I must be lost. What in the world happened here?

rossabaker: We ripped out the global parser registry, so there's nothing for it to do.  I guess it could have returned `this`.  It doesn't really matter, because this class is going to go away about two PRs from now.

ChristopherDavenport (approved):
Goodbye.

PR #4576, opened 2021-03-03
Remove legacy headers

rossabaker:


rossabaker:
This will depend on #4575.

ChristopherDavenport (approved):
🎉

PR #4578, opened 2021-03-03
reenable fatal warnings

m-sp:


core/src/main/scala/org/http4s/parser/QueryParser.scala
```diff
@@ -91,7 +90,7 @@ private[http4s] class QueryParser(
     // begin iterating through the chars
     while (error == null && input.hasRemaining) {
       val c = input.get()
-      (c: @switch) match {
+      c match {
```
m-sp, line 93: dotty doesn't like switch statements with guards, and if I remove the guards by hand we have too few conditions, dotty and scala only compile to a switch with more than 4 cases

build.sbt
```diff
@@ -25,7 +22,7 @@ ThisBuild / githubWorkflowBuild := Seq(
         List("unusedCompileDependenciesTest"),
         name = Some("Check unused compile dependencies"), cond = Some(s"matrix.scala != '$scala_3'")), // todo disable on dotty for now
       WorkflowStep.Sbt(List("test"), name = Some("Run tests")),
-      WorkflowStep.Sbt(List("doc"), name = Some("Build docs"), cond = Some(s"matrix.scala != '$scala_3'"))
+      WorkflowStep.Sbt(List("doc"), name = Some("Build docs"))
```
rossabaker, line 25: Oh, wow, this works now? Neat.

rossabaker (approved):

PR #4580, opened 2021-03-03
Uri path unsafe from string

hamnis:


m-sp (approved):

m-sp (approved):
on second thought, can you explain why? it doesn't seems to throw or be in any other ways dangerous

rossabaker:
Not every character is valid in a path.  Percent encoding is supported, but the spec is not prescriptive about the encoding of the URI.  (Something we're handwavy about).

- If we say this argument is an encoded value, it's not safe.
- If we say this argument is a decoded value, and we use a complete encoding like UTF-8, I think it is safe.
- If we say the URI is ISO-8859-1 and try to stuff an emoji in it, it's not safe.

m-sp:
thanks for the explanation :shipit:

PR #4581, opened 2021-03-03
Query fromString to unsafeFromString

hamnis:


m-sp (commented):
hmm in contrast to path we do not have an interpolator for the query string so I'm a bit more hesitant

core/src/main/scala/org/http4s/Query.scala
```diff
@@ -197,14 +197,18 @@ object Query {
     *
     * If parsing fails, the empty [[Query]] is returned
     */
-  def fromString(query: String): Query =
+  def unsafeFromString(query: String): Query =
```
rossabaker, line 200: If we're calling this unsafe, does it still make sense to quietly return an empty?  And might `fromString` now return an `Either`?

hamnis: I thought about doing that in two steps. first deprecating, then change the return type.
We could however do both in one step.

rossabaker: I think we'll probably want to reclaim the real name `fromString` for 0.22 and 1.0, so we should probably just go ahead and do it in the next milestone.  But the change could be part of a subsequent PR.

rossabaker:
@m-sp's point on an interpolator is interesting.  I found during the ip4s work that hiding an interpolator when it clashes is really hard!  So we should be conservative about the ones we create.  Or at least be pretty hopeful nobody else will use the same name.  `uriQuery` is probably too verbose.  Is `query` unique enough?  Or should we live without?

m-sp (approved):
> Is `query` unique enough

most likely not, I'n okay merging this but it is getting a bit hard to just construct a correct uri from scratch. especially without using any `unsafe` methods. as can be seen in our tests

hamnis:
If we had a Uri interpolator which accepted parameters, then we would be in a much better place with regards to uri construction.

Using the unsafe constructors are not bad per se, but we cant guarantee that every string is a "blessed" string without going through some parsing step. If you are sure that the string does not contain badness, then its fine.

rossabaker:
A good URI interpolator is complicated, but is one of the highest value things we could add right now.

rossabaker:
:+1:, but has some merge conflicts

Issue #4582, opened 2021-03-03
Flaky EntityEncoderSpec test


rossabaker:
```
==> X org.http4s.EntityEncoderSpec.Contravariant[EntityEncoder[F, *]]: contravariant.invariant composition  1.167s munit.FailException: Failing seed: wK0Rx3jseKC8WDwb-bfmgKE-3SiiDj_28EQkGW65JHH=
You can reproduce this failure by adding the following override to your suite:

  override val scalaCheckInitialSeed = "wK0Rx3jseKC8WDwb-bfmgKE-3SiiDj_28EQkGW65JHH="

Exception raised on property evaluation.
```

https://github.com/http4s/http4s/runs/2023774733

PR #4583, opened 2021-03-03
Stabilize entity encoder spec

m-sp:
should fix https://github.com/http4s/http4s/issues/4582


rossabaker:
Why did this fail under `IO`?  I haven't yet understood the underlying problem.

m-sp:
checking the logs the contravariant.invariant composition failed after 1.167s but the code has `.unsafeRunTimed(1.second)` so this just ran into timeouts. no more async code, no more timeouts

PR #4588, opened 2021-03-04
Make headers consistent / header cleanup

VlachJosef:
Hi, I thought it worthwhile to check all the headers once more to check they are consistent.

Core of this PR are following changes:

- Any extra `def value ...`, `def renderString` are removed form header class (in preference for syntax).
- Every header's companion defines `parse` method, and instance depends on it (to avoid two implementations).
- Couple of small tweaks


rossabaker (approved):
This is nice. Thanks for cleaning it up.

PR #4589, opened 2021-03-04
Remove SetCookieHeaderSuite (duplicate of SetCookieHeaderSpec)

VlachJosef:
Hi, I did notice there are two almost identical files:

- `tests/src/test/scala/org/http4s/parser/SetCookieHeaderSuite.scala`
- `tests/src/test/scala/org/http4s/headers/SetCookieHeaderSpec.scala`

tests they covers are 100% identical, they differ only in `Suite` they extends, so I believe they are duplicates.

Since they test parsing of `Cookie` header, I've decided to keep the one in `parser` package, so I've removed `SetCookieHeaderSpec`


rossabaker (changes requested):
Good find.

I think the `parser` package is obsolete -- it mostly held the parboiled that we just got rid of.  The `Set-Cookie` is in headers.  I like the idea, but I'd keep the other.

VlachJosef:
No problem, I'll update the PR.

PR #4590, opened 2021-03-04
removing inheritance from package object - booPickle

vder:
I'm not sure if there is a ticket for that

The other package objects with inheritance I could found are:
/play-json/src/main/scala/org/http4s/play/package.scala
/jawn/src/main/scala/org/http4s/jawn/package.scala
/circe/src/main/scala/org/http4s/circe/package.scala
/http4s/scalatags/src/main/scala/org/http4s/scalatags/package.scala
/http4s/twirl/src/main/scala/org/http4s/twirl/package.scala

If these changes are ok, I can proceed with other files as well- but probably not today ;)

### DIFF
--- a/boopickle/src/main/scala/org/http4s/booPickle/implicits.scala
+++ b/boopickle/src/main/scala/org/http4s/booPickle/implicits.scala
@@ -1,0 +1,20 @@
+/*
+ * Copyright 2018 http4s.org
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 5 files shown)


cquiroz (approved):
LGTM

rossabaker (approved):
Nicely done.

rossabaker:
circe will have some syntax.  That's the most complicated.  I'll leave it to you whether that should go next or last.  There's also a discussion to be had on implicits in dsl and client, but you found all the ones I was thinking about.

PR #4591, opened 2021-03-05
Fix flaky test where decoding could throw IllegalStateException

### LINKED ISSUE #4561: Flaky DecodeSpec example
```
==> X org.http4s.DecodeSpec.decode should either succeed or raise a CharacterCodingException  0.069s munit.FailException: /home/runner/work/http4s/http4s/tests/src/test/scala/org/http4s/DecodeSpec.scala:143
142:
143:  test("decode should either succeed or raise a CharacterCodingException") {
144:    forAll { (bs: Array[Byte], cs: Charset) =>

Failing seed: IBL0omGp6npzsX0tfYN5gMyayxqYsYysSeCU5hAY0dH=

  override val scalaCheckInitialSeed = "IBL0omGp6npzsX0tfYN5gMyayxqYsYysSeCU5hAY0dH="

Falsified after 31 passed tests.
> ARG_0: [B@69b69ad1
> ARG_1: x-COMPOUND_TEXT
> ARG_0_ORIGINAL: [B@213c7441
```

phongngtuan:
for example when we call flush at the of the stream for x-COMPOUND-TEXT charset on
jdk8 and the input stream ends in special control character (see
https://www.x.org/releases/X11R7.6/doc/xorg-docs/specs/CTEXT/ctext.html#Control_Characters)

This change relaxes the assertion on decoding result to make sure that
we only assert string content if the byte array is welformed and
accepted by nio's decoder

Fixes #4561

### DIFF
--- a/tests/src/test/scala/org/http4s/DecodeSpec.scala
+++ b/tests/src/test/scala/org/http4s/DecodeSpec.scala
@@ -17,18 +17,21 @@
 package org.http4s
 
 import cats.syntax.all._
-import java.nio.charset.{Charset => JCharset}
-import java.nio.charset.{
-  CharacterCodingException,
-  MalformedInputException,
-  UnmappableCharacterException
-}
 import fs2._
 import fs2.text.utf8Decode
 import org.http4s.internal.decode
 import org.http4s.laws.discipline.arbitrary._
-import java.nio.charset.StandardCharsets
 import org.scalacheck.Prop.{forAll, propBoolean}
+
+import java.nio.ByteBuffer
+import java.nio.charset.{
+  CodingErrorAction,
+  MalformedInputException,
+  StandardCharsets,
+  UnmappableCharacterException,
+  Charset => JCharset
+}
+import scala.util.Try
 
 class DecodeSpec extends Http4sSuite {
   test("decode should be consistent with utf8Decode") {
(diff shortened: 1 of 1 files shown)


tests/src/test/scala/org/http4s/DecodeSpec.scala
```diff
@@ -140,16 +143,21 @@ class DecodeSpec extends Http4sSuite {
...
   }
 
-  test("decode should either succeed or raise a CharacterCodingException") {
+  test("decode stream result should be consistent with nio's decode on full stream") {
     forAll { (bs: Array[Byte], cs: Charset) =>
+      val referenceDecoder = cs.nioCharset
+        .newDecoder()
+        // setting these to be consistent with our decoder's behavior
+        // note that java.nio.charset.Charset.decode and fs2's utf8Decode
+        // will replace character instead of raising exception
```
rossabaker, line 152: Maybe this is a behavior worth revisiting, if we're inconsistent with both fs2 and java defaults. :thinking:

tests/src/test/scala/org/http4s/DecodeSpec.scala
```diff
@@ -140,16 +143,21 @@ class DecodeSpec extends Http4sSuite {
...
       val source = Stream.emits(bs)
-      val decoded = source.through(decode[Fallible](cs)).compile.drain
-      decoded match {
-        case Left(_: CharacterCodingException) => true
-        case Right(_) => true
-        case _ => false
-      }
+      val decoded = source.through(decode[Fallible](cs)).compile.foldMonoid
+      // Ignoring the actual exception type
+      assertEquals(decoded.toOption, referenceResult.toOption)
```
rossabaker, line 160: Clever use of `.toOption`!

rossabaker (approved):

PR #4593, opened 2021-03-05
deinheritance-package-objects-circe

vder:
message extension methods moved out to syntax package, but earlier MethodSyntax was private and now is not.I'm not sure if this is right.
And the same remark as in #4592 - I ended up with a lot of members in CirceInstances without implicit modifier and it looks awkward, but I'm not sure how to address this.


rossabaker:
Let's sort out our strategy on #4594, and then maybe we'll have a clearer roadmap on this.

PR #4594, opened 2021-03-05
deinheritance of package objects --without circe 

vder:
I converted the files along the same pattern as the earlier boopickle package, but I ended up with some members of Instance traits without implicit modifier and I'm wondering now if this is correct.
example
trait JawnInstances { def jawnDecoder[F[_]: Sync, J: Facade]: EntityDecoder[F, J] = EntityDecoder.decodeBy(MediaType.application.json)(jawnDecoderImpl[F, J]) ...}


play-json/src/main/scala/org/http4s/play/instances/package.scala
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.http4s
+package play
+
+package object instances {
+  object all extends AllInstances
+  object play extends PlayInstances
```
rossabaker, line 22: In scala-xml, we called these `ElemInstances`, because they were for `Elem`.  Following that pattern, we could perhaps split this into `JsValue`, `Reads`, and `Writes`?  I'm not sure whether people will ever want that level of granularity, but it seems consistent with Cats.

If we don't split it up, I think we might just stick with `AllInstances` and `all`.

rossabaker, line 22: There is syntax in this: the `MessageSyntax`.  That would typically go in a `syntax` object in the Cats encoding.  Interestingly, it depends on `jsonOf`, so we need to think about where that belongs.  I've called it out on Gitter to see what the group thinks.

jawn/src/main/scala/org/http4s/jawn/implicits.scala
```diff
@@ -0,0 +1,21 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.http4s.jawn
+
+import org.http4s.jawn.instances.AllInstances
+
+object implicits extends AllInstances
```
rossabaker, line 21: If I'm not mistaken, this one doesn't have any implicits at all!

scalatags/src/main/scala/org/http4s/scalatags/instances/package.scala
```diff
@@ -0,0 +1,24 @@
...
+ * limitations under the License.
+ */
+
+package org.http4s
+package scalatags
+package instances
+
+package object instances {
+  object all extends AllInstances
+  object scalatags extends ScalatagsInstances
```
rossabaker, line 23: Okay, yay.  This one is as simple as the scala-xml one I did... and as simple as I thought the non-circe ones would all be.  But maybe we should call it `FragInstance`.

rossabaker (changes requested):
This is nice work.  Once again, I did an easy one myself and underappreciated some of the nuance we'd run into on others.  You raise an interesting point on where the "explicits" live.  I'll think about that, and see if anyone else has an opinion.

Also, great job splitting this up package by package.  It made a review much eaiser.

twirl/src/main/scala/org/http4s/twirl/instances/package.scala
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.http4s
+package twirl
+
+package object instances {
+  object all extends AllInstances
+  object twirl extends TwirlInstances
```
rossabaker, line 22: Similar to the others, we could break these down into `Html`, `Xml`, etc.  Lots of boilerplate, but I suppose nice for people who like to keep their implicits as tight as possible.

Helps your commit stats, too. :laughing:

PR #4604, opened 2021-03-07
Content-Disposition utf8 encoding fix

### LINKED ISSUE #4513: Content-Disposition header doesn't render UTF-8 encoding
Better description later, but before I forget:

```
681:  multipartParserTests(
682:    "mixed file parser",
values are not the same
=> Obtained
Headers(Content-Disposition: form-data; name*=UTF-8''http4s%20withspace; filename*="我老婆太漂亮.txt", Content-Type: application/octet-stream, Content-Transfer-Encoding: binary)
-Headers(Content-Disposition: form-data; name*=UTF-8''http4s%20withspace; filename*="我老婆太漂亮.txt", Content-Type: application/octet-stream, Content-Transfer-Encoding: binary)
+Headers(Content-Disposition: form-data; name*="http4s withspace"; filename*="我老婆太漂亮.txt", Content-Type: application/octet-stream, Content-Transfer-Encoding: binary)
```

vder:
fixes #4513


core/src/main/scala/org/http4s/headers/Content-Disposition.scala
```diff
@@ -79,13 +83,36 @@ object `Content-Disposition` {
...
+
+object PtcEncodedString {
+  implicit def fromString(s: String): PtcEncodedString = AsciiString(s)
+}
+
 // see http://tools.ietf.org/html/rfc2183
-final case class `Content-Disposition`(dispositionType: String, parameters: Map[String, String])
+final case class `Content-Disposition`(
+    dispositionType: String,
+    parameters: Map[String, PtcEncodedString])
```
rossabaker, line 118: Nothing is preventing us from creating `AsciiString("🔥")`, so I'm not sure `PctEncoding` belongs in the public API.  All regular parameters should support an ASCII encoding, and all `*` parameters are supposed to use the charset encoding.  I think we should just store everything as decoded `String`s.

Unfortunately, this simple model does not prevent us from creating invalid key names, or invalid values for keys that aren't the extended tokens.  I don't really see a way to enforce that that results in a pleasant type to use.

The spec does say `filename` and `filename*` are case insensitive, and that duplicate keys aren't supported.  It is silent on whether extension fields are case insensitive, but it's a good bet the answer is yes, so a `Map[CIString, String]` seems appropriate.

rossabaker: Very irritatingly, in [RFC7578](https://tools.ietf.org/html/rfc7578), we find the `filename*` MUST NOT be used.  So what's valid depends on whether this is a response header or a multipart part header.  I don't think we can achieve full compliance with all these specs at once without turning this into http4Idris.

core/src/main/scala/org/http4s/headers/Content-Disposition.scala
```diff
@@ -79,13 +83,36 @@ object `Content-Disposition` {
         new Renderable {
           def render(writer: Writer): writer.type = {
             writer.append(v.dispositionType)
-            v.parameters.foreach(p => writer << "; " << p._1 << "=\"" << p._2 << '"')
+            v.parameters.foreach(p => writer << "; " << p._1 << "=" << p._2)
```
rossabaker, line 86: There is a lot of nuance in rendering this correctly.  I think the [akka-http implementation](https://github.com/akka/akka-http/blob/b071bd67547714bd8bed2ccd8170fbbc6c2dbd77/akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala#L468-L492) is worth a look.  Note how it renders the extended `filename*` _after_ `filename`, which is important.  It also synthesizes a `filename*` where `filename` is unsafe, which seems like a nice idea.

rossabaker (commented):
Sorry for the slow review.  I had to reread the spec.  And what an awful spec this one is!

vder:
Thanks for input. hmm it looked easier at first, but I guess I have to do some RFC readings as well then.

vder:
Ok prepared some code based on akka-http and went back to parameters defined just as Map[String,String]. I had some doubts to introduce keys as CIString as I thought it should be consistent with other headers that are using a Map. But after quick scan through rest of these i didnt find any so I guess we could proced with this change to CIString. Just let me know if the rest of changes are all right.

core/src/main/scala/org/http4s/headers/Content-Disposition.scala
```diff
@@ -23,8 +23,11 @@ import org.http4s.internal.parsing.{Rfc2616, Rfc3986, Rfc7230}
 import org.http4s.util.{Renderable, Writer}
 import java.nio.charset.StandardCharsets
 import org.typelevel.ci._
+import scala.collection.immutable.TreeMap
 
 object `Content-Disposition` {
+  val safeChars = CharPredicate.Printable -- "%\"\\"
```
rossabaker, line 29: Could probably keep this private.

core/src/main/scala/org/http4s/headers/Content-Disposition.scala
```diff
@@ -77,9 +80,28 @@ object `Content-Disposition` {
...
             writer.append(v.dispositionType)
-            v.parameters.foreach(p => writer << "; " << p._1 << "=\"" << p._2 << '"')
+            withExtParamsSorted.foreach {
+              case (k, v) if k == "filename" =>
+                writer << "; " << k << '=' << '"'
+                writer.eligibleOnly(v, keep = safeChars, placeholder = '?') << '"'
+              case (k, v) if k.endsWith("*") =>
+                writer << "; " << k << '=' << "UTF-8''" << Uri.encode(v)
+              case (k, v) => writer << "; " << k << "=\"" << v << '"'
+            }
```
rossabaker, line 104: This is similar enough to the akka-http implementation that it's probably worth an attribution.  We already mention them in `NOTICES` since we've copied elsewhere from them.

rossabaker (commented):
This is looking good.  Some of these headers are shockingly complicated!

vder:
OK, so my understanding is I should:
- make this CharPredicate private, 
- add some comments remarking source od the change - something similar to ` '// Adapted from https://github.com/(...)' `one can see here  and there in the code
will that do?

What about this change to convert properties key to CIString? I could really use this ci interpolator here introduced lately, otherwise it could be quite annoying to use.

rossabaker:
Oh, right.  Yes, the `ci` interpolator should be available now -- it's part of case-insensitive-1.1.

Correct on the other two points.

PR #4611, opened 2021-03-12
The Debacktification

rossabaker:
We've long discussed eliminating the backticks, and I've long been in favor of it if someone brings the Scalafix.  Here's a start.


core/src/main/scala/org/http4s/CacheDirective.scala
```diff
@@ -25,51 +25,133 @@ sealed trait CacheDirective extends Product with Renderable {
...
+  val `no-transform` = noTransform
 
-  case object `only-if-cached` extends CacheDirective
+  case object onlyIfCached extends CacheDirective {
+    val name = ci"only-if-cached"
+  }
+  @deprecated("Renamed to onlyIfCached", "0.22.0-M6")
+  val `only-if-cached` = onlyIfCached
 
   final case class `private`(fieldNames: List[CIString] = Nil) extends CacheDirective {
```
rossabaker, line 101: I left this one.  Should I have?  If not, what shall we call the ones that clash with keywords?

core/src/main/scala/org/http4s/CacheDirective.scala
```diff
@@ -25,51 +25,133 @@ sealed trait CacheDirective extends Product with Renderable {
   * http://www.iana.org/assignments/http-cache-directives/http-cache-directives.xhtml
   */
 object CacheDirective {
-  final case class `max-age`(deltaSeconds: Duration) extends CacheDirective {
+  final case class maxAge(deltaSeconds: Duration) extends CacheDirective {
```
rossabaker, line 28: Should these be capitalized?  It would help with pattern matching.

rossabaker: It would also solve the `` `private` `` conundrum.

rossabaker (commented):
Some other ones coming soon:

* Do we do this for `Method`?  Huge change, but we've got a scalafix!
* What do we call `*`?  `All`?  `Any` (clashes with Predef)?  `Splat` (too cute)?

core/src/main/scala/org/http4s/HttpVersion.scala
```diff
@@ -36,12 +36,20 @@ final case class HttpVersion private[HttpVersion] (major: Int, minor: Int)
 }
 
 object HttpVersion {
-  val `HTTP/1.0` = new HttpVersion(1, 0)
-  val `HTTP/1.1` = new HttpVersion(1, 1)
-  val `HTTP/2.0` = new HttpVersion(2, 0)
+  val HTTP_1_0 = new HttpVersion(1, 0)
+  @deprecated("Rename to HTTP_1_0", "0.22.0-M6")
+  def `HTTP/1.0` = new HttpVersion(1, 0)
```
rossabaker, line 41: (Heads to the bikeshed)

Proper Scala naming convention would have been `Http1_0`.

core/src/main/scala/org/http4s/CacheDirective.scala
rossabaker: Mitigated through capitalization.

rossabaker:
Okay, `Method` is a big one, but it makes it more consistent with `Status`, and everything else.

rossabaker:
I think this is all of them except for MimeDb, which is its own hassle, and headers.  I'm inclined to take those on as separate PRs.  This one already grew far beyond its original scope.

isomarcte (commented):
Oh gosh, I'm so torn on this.

I do hate backticks, but I really do like having the constant names very unambiguously map back to the their values.

If we did go this route should we consider deprecating the backtick symbols and making them alias to the new ones? I realize that means likely 2x code and _probably_ has negative implications on anything generated (if we tackle the MimeDB with this). On the other hand, even though this would be in a 0.21.x -> 0.22.x upgrade, I still think it is really nice on users to use deprecation warnings to tell them what is going on.

I could go either way here, both on the backticks and deprecation.

ScalaWilliam:
No matter how much I dislike backticks, the advantage of having it match the canonical names is huge because you're never left guessing what the name is supposed to be.

I'm also torn like @isomarcte but thinking more about it, gear towards keeping them.
I know `HTTP/1.1` better than the back of my hand but would never guess that it would correspond to `Http1_1` (why not `Http_1_1`, for instance?)

Although it brings some duplication, would it perhaps make sense to have another DSL with aliases which gives that nicer cleaner Scala-ish look without having to do a migration - and also learn the new representations?

rossabaker:
Thanks for the thoughtful feedback.  I sincerely appreciate it, even if I'm about to disagree with most of it. :smile:
  
The HTTP versions are the least compelling.  As you note, their camel casing is debatable.  (What I chose matches a couple style guides I could find on the subject.)  But in these other types, we're intuitively removing hyphens.  Additionally, the initial capital lets them be used in extractors as the constants they are.

We're not shy about using advanced types and language features.  We hope to be approachable for new Scala users, but we're also not ashamed of our Kleislis or our higher-kinded type parameters.  As @SystemFw noted on Gitter, eventually learning these leads to a deeper understanding of the library and the language.  But the backticks are an extra short-term hurdle for limited long-term benefit.  It's another new language feature to learn out of the gate, and one that still plays poorly with language tooling after several years.  Even as both an advanced user of Scala and HTTP, I appreciate the autocompletes.

As for aliasing them via a DSL, I would rather have one way I didn't prefer than have two ways to support and teach.  I'm not looking forward to the transition, but hopefully the Scalafix makes it quick and painless.

rossabaker:
It's less about IDE support (which has gotten better), and more about not having to teach another somewhat arcane Scala feature to new Scala users.  The main beneficiaries of the status quo are people who know both Scala backticked identifiers _and_ HTTP wire representations.

jackcviers:
> HTTP wire representations

The issue I have is that if we're able to represent the language of the
domain directly, we should, because of IDE discoverability for new users.

The backticks aren't that arcane - they are in daily use in JS now, are
used in bash, and haskell. Even scalatags uses it for escaping keywords to
keep attribute names like `class` from colliding with scala reserved words.

It's not uncommon for users to need application/json as a content type, or
application/octet-stream. New library and Scala users are first going to
copy/paste a mime type in, and are going to use an IDE to then find the
name they want via autocomplete or symbol search. If they are medium level
juniors, they will grep search the library to find the type they want to
use, and will just use the identifier they find, regardless of backticks or
not. But newbies are just going to be lost without exact matches and IDE
completion, or a class or object with a method to parse mime types under a
well-named class.

I guess my thoughts are that we should make the mimes as discoverable as
possible via IDE completion. The current backticked names are very
discoverable.

Needing to specify a content type without an entity body to encode (at
which point the encoder can add the appropriate type)  is common enough
that figuring out which magic constant/class name to  use for status, Http
method names and content-types is probably the most annoying thing about
using a new http/http client lib for me -- I always end up grepping.

I understand why it is a goal for 1.0 to make the change, as without it
we're stuck with the api for another full version. I just question whether
or not backticks are that uncommon of a language feature any longer, and
want the easy discovery provided by the backticked names.

If we really want to get rid of the backticks, it might be a good idea to
make these strings and have a verification/parsing macro like we do with
the uri String context to parse into a MimeType object instead. That might
be easy to look up in the documentation and be something new users would be
accustomed to.

Jack Viers - 515.313.7838 ***@***.***

rossabaker:
People wanted this for years, until someone started doing it! 😆

Inertia is with backticks, and support for them appears to be rising.  Does anybody want to argue for proceeding with this PR?

bplommer:
I'm 👎 on this. I'm just the process now of properly learning the HTTP specs, and I'm finding it really helpful that things in http4s are named after their literal values.

rossabaker:
Public opinion has shifted, and for such a vast change, we'd want something closer to a supermajority. Withdrawing.

Issue #4612, opened 2021-03-12
CSRF checkCSRFinHeaderAndForm fails for multipart


jan0sch:
If the submitted form is a multipart form then the check does not trigger and the request is refused even if a proper csrf field is included in the form.

jan0sch:
I've started working on it and will come up with PRs soon.

rossabaker:
If the fix is binary compatible, series/0.21 would be the best place for it, so we can get it into a stable release sooner.

PR #4613, opened 2021-03-12
CSRF checkCSRFinHeaderAndForm fails for multipart

jan0sch:
Add support for multipart forms in the checkCSRFinHeaderAndForm function
to avoid getting a 403 error code on otherwise valid forms (see #4612).

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -146,7 +146,7 @@ lazy val tests = libraryProject("tests")
     description := "Tests for core project",
     startYear := Some(2013),
   )
-  .dependsOn(core, client, blazeClient, blazeServer, theDsl, testing % "test->test")
+  .dependsOn(core, client, emberServer, theDsl, testing % "test->test")
 
 lazy val server = libraryProject("server")
   .settings(
--- a/build.sbt
+++ b/build.sbt
@@ -146,7 +146,7 @@ lazy val tests = libraryProject("tests")
     description := "Tests for core project",
     startYear := Some(2013),
   )
-  .dependsOn(core, testing % "test->test")
+  .dependsOn(core, client, blazeClient, blazeServer, theDsl, testing % "test->test")
 
 lazy val server = libraryProject("server")
   .settings(
--- a/build.sbt
+++ b/build.sbt
@@ -146,7 +146,7 @@ lazy val tests = libraryProject("tests")
     description := "Tests for core project",
     startYear := Some(2013),
   )
-  .dependsOn(core, client, emberServer, theDsl, testing % "test->test")
+  .dependsOn(core, client, emberClient, emberServer, theDsl, testing % "test->test")
 
 lazy val server = libraryProject("server")
   .settings(
--- a/server/src/main/scala/org/http4s/server/middleware/CSRF.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/CSRF.scala
@@ -413,8 +413,10 @@ object CSRF {
           case Right(dec) =>
             // The `.drop(1)` removes the first empty line from the body value of the field.
             dec.parts
-              .find(_.headers.exists(h =>
-                h.is(headers.`Content-Disposition`) && h.value.contains(s"""name="$fieldName"""")))
+              .find(
+                _.headers
+                  .get(headers.`Content-Disposition`)
+                  .fold(false)(_.value.contains(fieldName)))
               .traverse(_.body.through(fs2.text.utf8Decode).drop(1).compile.fold("")(_ |+| _))
           case Left(_) => Sync[G].pure(none[String])
         }
--- a/server/src/main/scala/org/http4s/server/middleware/CSRF.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/CSRF.scala
@@ -28,6 +28,7 @@ import java.security.{MessageDigest, SecureRandom}
 import java.time.Clock
 import javax.crypto.spec.SecretKeySpec
 import javax.crypto.{KeyGenerator, Mac, SecretKey}
+import org.http4s.EntityDecoder.multipart
 import org.http4s.headers.{Cookie => HCookie}
 import org.http4s.headers.{Host, Origin, Referer, `X-Forwarded-For`}
 import org.http4s.util.CaseInsensitiveString
@@ -406,10 +407,26 @@ object CSRF {
       }
     }
 
+    def getMultipartToken: F[Option[String]] =
+      if (r.headers.get(headers.`Content-Type`).map(_.mediaType.isMultipart).getOrElse(false)) {
+        val token = multipart[G].decode(r, strict = true).value.flatMap {
+          case Right(dec) =>
+            // The `.drop(1)` removes the first empty line from the body value of the field.
+            dec.parts
+              .find(_.headers.exists(_.value.contains(s"""name="$fieldName"""")))
+              .traverse(_.body.through(fs2.text.utf8Decode).drop(1).compile.fold("")(_ |+| _))
+          case Left(_) => Sync[G].pure(none[String])
+        }
+        nt(token)
+      } else {
+        F.pure(none[String])
+      }
+
     for {
       fst <- F.pure(csrf.getHeaderToken(r))
       snd <- if (fst.isDefined) F.pure(fst) else getFormToken
-      tok <- snd.fold(csrf.onfailureF)(csrf.checkCSRFToken(r, http, _))
+      trd <- if (snd.isDefined) F.pure(snd) else getMultipartToken
+      tok <- trd.fold(csrf.onfailureF)(csrf.checkCSRFToken(r, http, _))
     } yield tok
   }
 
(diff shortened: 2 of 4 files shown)


jan0sch:
If this is feasible then I can create PRs for the other branches as well (0.22/main).

server/src/main/scala/org/http4s/server/middleware/CSRF.scala
```diff
@@ -406,10 +407,26 @@ object CSRF {
       }
     }
 
+    def getMultipartToken: F[Option[String]] =
+      if (r.headers.get(headers.`Content-Type`).map(_.mediaType.isMultipart).getOrElse(false)) {
+        val token = multipart[G].decode(r, strict = true).value.flatMap {
```
rossabaker, line 412: This seems unsafe to me: if we consume the body in a middleware, how can the application consume it a second time?  It will work for strict sources, but not for bytes streaming in off a socket.

But I see the same problem exists in `getFormToken`, and nobody has complained about that.  Is this something that definitely works in a live integration test, as opposed to just in unit tests?

server/src/main/scala/org/http4s/server/middleware/CSRF.scala
```diff
@@ -406,10 +407,26 @@ object CSRF {
...
+          case Right(dec) =>
+            // The `.drop(1)` removes the first empty line from the body value of the field.
+            dec.parts
+              .find(_.headers.exists(_.value.contains(s"""name="$fieldName"""")))
```
rossabaker, line 416: I think we should be looking for Content-Disposition headers here?  Also, without double checking the RFC, I'm not certain the name needs to be quoted.  The parser should handle that.

server/src/main/scala/org/http4s/server/middleware/CSRF.scala
jan0sch: Regarding the `getFormToken`: We use CSRF and Forms (with a token field) in several apps and never had problems. However they only are regular forms.
On the other hand a client doing fancy javascript websocket streaming could simply send the appropriate header and circumvent the issue alltogether. The form parts are only called if no header is found.

server/src/main/scala/org/http4s/server/middleware/CSRF.scala
jan0sch: Right. Regarding the quotes I found nothing specific in RFC-7578, but their example uses quotes: `Content-Disposition: form-data; name="user"`
It also mentions that the field name *could* be encoded but I guess that would have already handled here, or is it?

server/src/main/scala/org/http4s/server/middleware/CSRF.scala
jan0sch: On a second thought... I think I'll try some integration test locally to avoid involving nasty suprises. After all the body parts are streams. :thinking:

server/src/main/scala/org/http4s/server/middleware/CSRF.scala
rossabaker: Not compiled, but I think this does what you want:

```
dec.parts.find(_.headers.get(`Content-Disposition`).fold(false, _.get("name").isDefined)))
```

server/src/main/scala/org/http4s/server/middleware/CSRF.scala
```diff
@@ -413,7 +413,8 @@ object CSRF {
           case Right(dec) =>
             // The `.drop(1)` removes the first empty line from the body value of the field.
             dec.parts
-              .find(_.headers.exists(_.value.contains(s"""name="$fieldName"""")))
+              .find(_.headers.exists(h =>
+                h.is(headers.`Content-Disposition`) && h.value.contains(s"""name="$fieldName"""")))
```
rossabaker, line 417: Does it not parse the `name` into the `parameters`?

jan0sch: Sorry, was a bit in a rush yesterday. I cleaned it up a bit.
However I still have trouble with it in integration. I published locally and used in an exisiting project and there it doesn't seem to work.
After the weekend I'll dig deeper into the issue and try to reproduce that.

rossabaker: My guess is that it's related to that consuming the stream twice.  This seems like a difficult problem to handle in a streaming library: we need to parse the stream to find the token, and then buffer that stream so the app can handle it, which fights against the nature of streaming in the first place.  It's doable, but it's going to come with some nasty space tradeoffs.

jan0sch:
I tried adding an integration tests but had to mess with the dependsOn settings of the test module. Nothing which should be merged. ;-)
However I noticed that the tests bail out for every check, even header and regular form which definitely work. :-/

jan0sch:
Hmm... some debugging indicates that the `Origin` header goes missing within the test requests. So it is no wonder that these get denied.

tests/src/test/scala/org/http4s/middleware/CSRFIntegrationSuite.scala
```diff
@@ -0,0 +1,182 @@
...
+    val check = client.expect[String](req).unsafeRunSync()
+    fiber.cancel.unsafeRunSync()
+    assertEquals(check, "CSRF passed!")
+  }
+
+}
+
+object CSRFIntegrationSuite {
+  val blockingPool = Executors.newFixedThreadPool(2)
+  val blocker = Blocker.liftExecutorService(blockingPool)
```
rossabaker, line 180: There's already an `Http4sSuite.TestBlocker` that should work here.

build.sbt
```diff
@@ -145,7 +145,7 @@ lazy val tests = libraryProject("tests")
     description := "Tests for core project",
     startYear := Some(2013),
   )
-  .dependsOn(core, testing % "test->test")
+  .dependsOn(core, client, blazeClient, blazeServer, theDsl, testing % "test->test")
```
rossabaker, line 148: In the medium term, blaze is moving out of this repo, but ember is staying.  It would be better to use Ember here.

rossabaker (commented):
This is good.  Potentially, we'll want to move this out into a new integration tests module.  The heavier tests have caused problems in CI in the past, and we might want to run them in their own job, or be able to test core before we've ported backends.  But I'm happy to move around code later if it becomes a problem in practice.

jan0sch:
I've reworked it to use the ember server but the problem with the Origin header being somehow removed from the test request persists.

rossabaker:
It's a [limitation of the JDK8 client]( https://github.com/openjdk/jdk/blob/70e5173968cdc245b328b3f3cf3e91786392c5ef/jdk/src/share/classes/sun/net/www/protocol/http/HttpURLConnection.java#L194), which [appears to be fixed](https://bugs.openjdk.java.net/browse/JDK-8212692) in later JDKs.

jan0sch:
> It's a limitation of the JDK8 client...

Thanks for digging this out. I was guessing along the same lines but didn't have much time. However I'm using OpenJDK 11.0.10 and it still seems to be an issue. So I guess I'll have to use another client (ember?).

rossabaker:
Yeah.  It's a shame, because this kind of thing is exactly what that JDK client should be good for.  But Ember will remain in the core repo, so that's the other good choice.

jan0sch:
So, we're using the ember client now and everything is green.
But I will add another test case for multipart which includes also a streaming file field to model exactly the case that didn't work locally.

jan0sch:
Okay, I adjusted the tests routes to actually consume the request and as we can see the multipart routes break in that case.

So we don't have a solution yet but a working test case for the scenario. ;-)

Is it possible to "copy" the request object, consume the copy in the middleware and pass the original through upon successful check?

rossabaker:
Okay, great, I'm glad the bug is exposed.

A similar problem has come up recently, where people want to peek into a stream before committing to a status code.   Here's [an old example](https://github.com/typelevel/fs2/issues/1982).  This is tougher, because you potentially need to uncons several chunks before you find the one you're looking for.  You could theoretically write a `Pull` that would uncons enough to find the token, concatenating the chunks it has unconsed, and consing that concatenated chunk back onto a stream.  But:

1. You'd effectively be parsing twice, even if this parser didn't build the full model.
2. The multipart spec is non-trivial.  In addition to doing the work twice at runtime, this Pull will be tricky.
3. There's no limit on the chunks you uncons.  You could potentially be buffering a huge response into memory looking for that token.

This is just a nasty problem, and we might do well to see how other multipart + CSRF implementations have solved it, and then see if we can translate that back to fs2.  I'd love to find out that we're just trying to solve the wrong problem.

rossabaker:
In Spring, they have [two ways](https://docs.spring.io/spring-security/site/docs/3.2.0.CI-SNAPSHOT/reference/html/csrf.html#csrf-multipart).  It has been many years since I've done Spring, so I'm not sure precisely what they're doing with that `MultipartFilter`.  The second option, putting it in a query param, would get you out of this mess, but carries the security implications noted there.

Do you have any control over the client here?  Could you put the token in a header, where it is safe from the URL but also can be validated before you do the expensive parsing of the multipart?

jan0sch:
Nothing new, just rebased to avoid branching too far. I hope to have some time soon to revisit this problem.

PR #4614, opened 2021-03-12
Retry based on Idempotent-Key header.

vglushak:
Hi, small enhancement for consideration.

At the moment retry middleware is retrying failed requests only if method is idempotent.
Idempotency defined in rfc7231 (https://tools.ietf.org/html/rfc7231#section-4.2.2): as PUT, DELETE and safe methods.
However, it does not say that POST requests can’t be idempotent.

Enhancement is about giving user flexibility to make their requests idempotent by adding to request http header: `Idempotency-Key`. If method is idempotent by definition or if header present allow retrying failed requests.

This is in line with how some public API designed: https://stripe.com/docs/api/idempotent_requests
Moreover, there is a draft RFC to standardise it: https://tools.ietf.org/html/draft-idempotency-header-00#section-2.1

### DIFF
--- a/client/src/main/scala/org/http4s/client/middleware/Retry.scala
+++ b/client/src/main/scala/org/http4s/client/middleware/Retry.scala
@@ -20,12 +20,13 @@ package middleware
 
 import cats.effect.{Concurrent, Resource, Timer}
 import cats.syntax.all._
-import java.time.Instant
-import java.time.temporal.ChronoUnit
 import org.http4s.Status._
-import org.http4s.headers.`Retry-After`
+import org.http4s.headers.{`Idempotency-Key`, `Retry-After`}
 import org.log4s.getLogger
 import org.typelevel.ci.CIString
+
+import java.time.Instant
+import java.time.temporal.ChronoUnit
 import scala.concurrent.duration._
 import scala.math.{min, pow, random}
 
@@ -125,15 +126,16 @@ object RetryPolicy {
     GatewayTimeout
   )
 
-  /** Returns true if the request method is idempotent and the result is
-    * either a throwable or has one of the `RetriableStatuses`.
+  /** Returns true if (the request method is idempotent or request contains Idempotency-Key header)
+    * and the result is either a throwable or has one of the `RetriableStatuses`.
     *
     * Caution: if the request body is effectful, the effects will be
     * run twice.  The most common symptom of this will be resubmitting
     * an idempotent request.
     */
   def defaultRetriable[F[_]](req: Request[F], result: Either[Throwable, Response[F]]): Boolean =
-    req.method.isIdempotent && isErrorOrRetriableStatus(result)
+    (req.method.isIdempotent || req.headers.get[`Idempotency-Key`].isDefined) &&
+      isErrorOrRetriableStatus(result)
 
   @deprecated("Use defaultRetriable instead", "0.19.0")
   def unsafeRetriable[F[_]](req: Request[F], result: Either[Throwable, Response[F]]): Boolean =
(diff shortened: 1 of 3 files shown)


rossabaker (changes requested):
Thanks!  I wasn't familiar with this specification, but it seems reasonable.

It would be neat to make this a modeled header.  I wouldn't invest in that part of it for 0.21, but we could add a typeclass instance on series/0.22.

Just needs a scalafmt, and I think it's good to go.

rossabaker:
Test failure is unrelated: https://github.com/http4s/http4s/issues/4615

rossabaker:
Oh, was the original version not binary compatible?  I think we could roll that out in 0.21, but this is helpful for the new things we can do in 0.22.  I could probably find the old git hash in Actions, if you no longer have it handy.

vglushak:
I've squashed commits in one. Let me create separate branch for 0.21 with original version.

vglushak:
[Original version](https://github.com/http4s/http4s/compare/series/0.21...vglushak:retry-0.21?expand=1). Not sure what is the process. Do we need another PR?

core/src/main/scala/org/http4s/headers/`Idempotency-Key`.scala
```diff
@@ -0,0 +1,42 @@
...
+      _.key,
+      parse
+    )
+}
+
+/** Request header defines request to be idempotent used by client retry middleware.
+  *
+  *  [[https://tools.ietf.org/html/draft-idempotency-header-00#section-2.1 idempotency-header]]
+  */
+final case class `Idempotency-Key`(key: CIString)
```
rossabaker, line 42: Oh, one more thing: is there anything in the spec that suggests this key is case-insensitive?  It seems to me that it's just an opaque token.

rossabaker (commented):
Thanks, I can merge the 0.21 version from the ref you posted.  One comment below on the modeled header.

client/src/main/scala/org/http4s/client/middleware/Retry.scala
```diff
@@ -125,15 +126,16 @@ object RetryPolicy {
...
+  /** Returns true if (the request method is idempotent or request contains Idempotency-Key header)
+    * and the result is either a throwable or has one of the `RetriableStatuses`.
     *
     * Caution: if the request body is effectful, the effects will be
     * run twice.  The most common symptom of this will be resubmitting
     * an idempotent request.
     */
   def defaultRetriable[F[_]](req: Request[F], result: Either[Throwable, Response[F]]): Boolean =
-    req.method.isIdempotent && isErrorOrRetriableStatus(result)
+    (req.method.isIdempotent || req.headers.get[`Idempotency-Key`].isDefined) &&
```
kevinmeredith, line 137: Regarding the header, `idempotency-key`, must the HTTP Request's Method be a `POST`, @vglushak @rossabaker?

https://tools.ietf.org/html/draft-idempotency-header-00#section-2.2 notes:

> The idempotency key that is supplied as part of every "POST" request
MUST be unique and can not be reused with another request with a
different request payload.

In other words, should this instead be:

```scala
(req.method.isIdempotent || (req.headers.get[`Idempotency-Key`].isDefined && req.method === Method.Post)
```

?

rossabaker (approved):
Looks great. I'll also integrate the original version into 0.21.

client/src/main/scala/org/http4s/client/middleware/Retry.scala
rossabaker: Oops, I hadn't seen this before I approved.  This is a good question.  I don't see any reason that the header couldn't be applied to a `PATCH` request.  And I think if the client cared enough care to provide this header, it's with the intent to treat the request as idempotent.  The spec isn't absolute, but I think this implementation is correct as is.

Issue #4600, opened 2021-03-07
Use Literally


rossabaker:
We'll be inheriting this soon from ip4s anyway, so there's no extra dependency liability, and it clean things up for us.

See https://github.com/Comcast/ip4s/pull/261 for an example.

kevinmeredith:
OK to close this, @rossabaker? Or you only close once you cut a release! Thanks

rossabaker:
Close on merge is good.  I just am not as diligent as I ought to be about mopping up.

PR #4616, opened 2021-03-12
Add Literally[Uri]

kevinmeredith:
Replaces macro definition of `Uri` w/ `typelevel`'s `literally` library. Closes part of #4600.

```scala
$sbt
sbt:core> console
[info] compiling 1 Scala source to /Users/kevinmeredith/Workspace/http4s/core/target/scala-2.13/classes ...
[info] Starting scala interpreter...
Welcome to Scala 2.13.4 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_112).
import fs2._
import cats._
import cats.data._
import cats.effect._
import cats.implicits._

scala> import org.http4s.implicits._
import org.http4s.implicits._

scala> uri"a.b.c"
val res0: org.http4s.Uri = a.b.c

scala> uri"FOOBAR"
val res1: org.http4s.Uri = FOOBAR

scala> uri"    "
       ^
       error: invalid URI: Invalid input ' ', expected AbsoluteUri or RelativeRef (line 1, column 1):

       ^
```

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -95,6 +95,7 @@ lazy val core = libraryProject("core")
       scodecBits,
       slf4jApi, // residual dependency from macros
       vault,
+      literally
     ),
     unusedCompileDependenciesFilter -= moduleFilter("org.scala-lang", "scala-reflect"),
     mimaBinaryIssueFilters ++= Seq(
--- a/project/Http4sPlugin.scala
+++ b/project/Http4sPlugin.scala
@@ -350,6 +350,7 @@ object Http4sPlugin extends AutoPlugin {
     val treehugger = "0.4.4"
     val twirl = "1.4.2"
     val vault = "2.0.0"
+    val literally = "0.1-360f22d"
   }
 
   lazy val argonaut                         = "io.argonaut"            %% "argonaut"                  % V.argonaut
@@ -393,6 +394,7 @@ object Http4sPlugin extends AutoPlugin {
   lazy val json4sJackson                    = "org.json4s"             %% "json4s-jackson"            % V.json4s
   lazy val json4sNative                     = "org.json4s"             %% "json4s-native"             % V.json4s
   lazy val keypool                          = "io.chrisdavenport"      %% "keypool"                   % V.keypool
+  lazy val literally                        = "org.typelevel"          %% "literally"                 % V.literally
   lazy val log4catsCore                     = "io.chrisdavenport"      %% "log4cats-core"             % V.log4cats
   lazy val log4catsSlf4j                    = "io.chrisdavenport"      %% "log4cats-slf4j"            % V.log4cats
   lazy val log4catsTesting                  = "io.chrisdavenport"      %% "log4cats-testing"          % V.log4cats
--- a/tests/src/test/scala/org/http4s/LiteralSynxtaxMacrosSuite.scala
+++ b/tests/src/test/scala/org/http4s/LiteralSynxtaxMacrosSuite.scala
@@ -1,0 +1,25 @@
+/*
+ * Copyright 2013 http4s.org
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.http4s
+
+import org.http4s.implicits._
+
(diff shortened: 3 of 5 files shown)


kevinmeredith:
Can you please take a look, @rossabaker? 

It looks like it's not entirely working:

```scala
$sbt
>project core
>console
scala> import org.http4s._, org.http4s.implicits._
import org.http4s._
import org.http4s.implicits._

scala> uri"a.b.c"
       ^
       error: missing argument list for method get in trait Ops
       Unapplied methods are only converted to functions when a function type is expected.
       You can make this conversion explicit by writing `get _` or `get(_)` instead of `get`.
```

tests/src/test/scala/org/http4s/LiteralSynxtaxMacrosSuite.scala
```diff
@@ -0,0 +1,25 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.http4s
+
+import org.http4s.implicits._
+
+class LiteralSynxtaxMacrosSuite extends Http4sSuite {
```
kevinmeredith, line 21: Using munit, do you know how to test that something does not compile, @rossabaker @ChristopherDavenport?

kevinmeredith:
As of https://github.com/http4s/http4s/pull/4616/commits/60cf2cb03a6db369084643f337530fe179486857, multiple builds failed with:

```scala
[error] (core / mimaReportBinaryIssues) Failed binary compatibility check against org.http4s:http4s-core_2.13:0.21.15! Found 2 potential problems
```

It's not clear to me why that is.

build.sbt
```diff
@@ -95,6 +95,7 @@ lazy val core = libraryProject("core")
       scodecBits,
       slf4jApi, // residual dependency from macros
       vault,
+      literally
```
rossabaker, line 98: Nit: we've been alphabetizing these

tests/src/test/scala/org/http4s/LiteralSynxtaxMacrosSuite.scala
rossabaker:
```scala
    assert(
      compileErrors {
        """qValue"invalid" // doesn't compile, not parsable as a double"""
      }.nonEmpty
    )
```

project/Http4sPlugin.scala
```diff
@@ -350,6 +350,7 @@ object Http4sPlugin extends AutoPlugin {
     val treehugger = "0.4.4"
     val twirl = "1.4.2"
     val vault = "2.0.0"
+    val literally = "0.1-360f22d"
```
rossabaker, line 353: These are alphabetical too.

rossabaker (changes requested):
Thanks!  This is a library I've wanted to integrate.  We'll be getting it via ip4s soon anyway, so there's little reason to not take advantage of it.

A few observations:
* I'd rather not introduce a new dependency on the production branch. Let's target series/0.22.
* I'd also rather not merge a snapshot dependency onto a live branch, so this may sit after approval until literally has a proper release. @mpilquist, is an 0.1 imminent?
* I think it would be interesting to explore interpolating multiple parts, for say dynamic path or host components.  Ensuring that these are well-formed so the result need not be lifted into an `Either` is going to be tricky, so more likely we'll leave this as it is.  I don't know yet.  Either way, it will be good to have this library integrated so we can replace the others, which are all simpler.

mpilquist:
> @mpilquist, is an 0.1 imminent?

I just released 1.0.0-RC1

kevinmeredith:
> I'd rather not introduce a new dependency on the production branch. 

Understood, @rossabaker. I'll change it. For my own interest and learning, can you please explain your decision to not add a new library to 0.21? Thanks!

rossabaker:
I view the binary compatibility of dependencies as an extension of our own binary compatibility.  Imagine a breaking change to scodec-bits.  We might bump that on 0.21 and pass MiMa. But out in userland, we'd be evicting the scodec-bits dependency that might be used elsewhere: people using in scdoec would have to do a breaking upgrade for an allegedly non-breaking change to http4s.

The same principle holds for new libraries.  Suppose we introduced fastparse-2.  Any users of http4s-0.21 on fastparse-1 are out of luck.  The generic principle is that any dependencies that can cause breaking evictions should be postponed to breaking versions.

Now, Literally is a safer case: we'd be getting in on the earliest public release, so the risk of eviction is negligible.  We don't like production releases depending on milestones or release candidates, or else we're likely to get stuck on them per the above rules.  So we'd have to wait until literally-1.0.0 to put it on 0.21.

If we did put it on series/0.21, we'd also need to need to figure out how to do it without violating MiMa.  series/0.22 isn't MiMa locked yet, and is releasable on RC dependencies  Given where we're at and where Literally is at, it's just simpler to target 0.22.

kevinmeredith:
Thanks for your help, Ross and Michael. I'm closing this PR as I opened https://github.com/http4s/http4s/pull/4630.

PR #4618, opened 2021-03-14
uri interpolator draft

m-sp:
a first draft how a smarter uri interpolator could look like for http4s using our parsers
currently only works in dotty

putting it out there because I'm not sure when I can come back an finish this

currently there seems to be a couple of bugs, it only supports absolute Uris and the interpolating could be better. but it is a start

### DIFF
--- a/core/src/main/scala-3/org/http4s/syntax/LiteralsSyntax.scala
+++ b/core/src/main/scala-3/org/http4s/syntax/LiteralsSyntax.scala
@@ -22,7 +22,7 @@ import scala.language.`3.0`
 
 trait LiteralsSyntax {
   extension (inline ctx: StringContext) {
-    inline def uri(args: Any*): Uri = ${LiteralsSyntax.validateUri('{ctx}, '{args})}
+    inline def uri(inline args: String*): Uri = ${LiteralsSyntax.validateUri('{ctx}, '{args})}
     inline def path(args: Any*): Uri.Path = ${LiteralsSyntax.validatePath('{ctx}, '{args})}
     inline def scheme(args: Any*): Uri.Scheme = ${LiteralsSyntax.validateUriScheme('{ctx}, '{args})}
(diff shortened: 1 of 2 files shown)


rossabaker (commented):
Neat.  I'm glad to see this discussion underway.  If we figure out the semantics, I think we can figure out the crossbuild.  There are all sorts of angles to explore here:
* Should the scheme part accept a `String`, a `Scheme`, or both?
* If we put a `String` in a path, should we encode it into a path segment?  Should it accept a `Path`?
* Similar questions for queries and fragments.  And ip4s types in the hosts...

m-sp:
> Neat. I'm glad to see this discussion underway. If we figure out the semantics, I think we can figure out the crossbuild. There are all sorts of angles to explore here:
> 
> * Should the scheme part accept a `String`, a `Scheme`, or both?
> * If we put a `String` in a path, should we encode it into a path segment?  Should it accept a `Path`?
> * Similar questions for queries and fragments.  And ip4s types in the hosts...

Yep there are a lot of things to explore, if we can restrict ourselves to the concrete types `Scheme`, `Path`, `Query`, etc. we could get away with a simpler implementation.
currently the interpolator tries to progress as far as possible and at each step checks if we now need to insert the interpolation part. but if we already know which parts we have as arguments we can leave those out during parser construction.

this whole scheme breaks down again if we want to support `uri"http://http4s.org/test?hello=${"world"}"`

rossabaker:
Taking the stronger types makes a bit more sense to me, but maybe isn't as useful.  Particularly if we don't have interpolators for scheme, query, etc.

armanbilge:
Triaging aggressively, apologies. This PR looks great and I hope the work/energy can be retargeted to cats-uri. Thanks!

- https://github.com/http4s/http4s/discussions/5930
- https://github.com/typelevel/cats-uri/issues/5

PR #4619, opened 2021-03-14
Configure more secure defaults for SAXParserFactory

rossabaker:
These will be the default settings in scala-xml-2: https://github.com/scala/scala-xml/pull/177


ChristopherDavenport (approved):
I have no idea what this means, but I assume you do.

rossabaker:
On further thought, scala-xml is delaying these changes to a breaking release, and we should probably do the same.  It'll be in 0.22.

PR #4620, opened 2021-03-15
Make XML charset inference compliant with RFC7303

rossabaker:


ChristopherDavenport (approved):

rossabaker (commented):
Oops, found this comment in the drafts...

Turns out that streaming is fraught, because it has to run on a different thread pool.

PR #4621, opened 2021-03-15
Breaking improvements to scala-xml

rossabaker:
* Reverts the more complicated implicit encoding of #4380.
* Introduce safer defaults from scala-xml-2.
* Render XML declaration with charset
* Set charset in Content-Type


rossabaker:
9848eb2 could be backported.  It's a SHOULD, not a MUST.

rossabaker:
I think 7705fde was bad.  CI started hanging.

isomarcte (approved):

yanns (approved):
not an expert, but the code is looking good and seems easier to reason with.

ChristopherDavenport (approved):

ChristopherDavenport:
Should adjust description to match current status

rossabaker:
Yeah, it would be nice if we'd better make the description match the code in general.  I gave this an update.

PR #4622, opened 2021-03-15
XML encoder improvement backports

rossabaker:
The XML Declaration is more aggressive than the charset for a backport, but they're both things we SHOULD do.  It's particularly naughty to not do either.


yanns (approved):
👍

ChristopherDavenport (approved):

PR #4627, opened 2021-03-18
Deprecate the Status.apply(headers) dsl

hamnis:
The model in 0.22 and 1.0 is changed and Scala does not know how to dispatch to the overloaded apply.

This adds a deprecation to the apply method that only accepts headers and adds a `headers` constructor.


rossabaker:
:+1: to the concept.  Needs a scalafmt.  Would love to figure out a scalafix, but that can be separate.

hamnis:
This is not binary compatible, and if I make it compatible, the build fails on deprecation warnings...

rossabaker:
Sounds like we should just aim for 0.22 and skip the deprecation cycle, then?

hamnis:
Retargeting for 0.22. Closing

PR #4635, opened 2021-03-22
Migration to cats-effect-3.0.0-RC3

VlachJosef:
resolves #4634


rossabaker (commented):
Looks great.  This version of CE3 is not binary compatible with RC2 (even though the tests passed), so let's push library bumps to this branch too as they're ready, and it can all go together.

Thanks!

rossabaker:
cats-effect-3.0.0-final is immiment, and we won't release main without it.  Let's go ahead and merge this now before it bitrots.

PR #4636, opened 2021-03-23
added isErrorOrStatus function and tests for RetryPolicy

Adriani-Furtado:
As per title, adding the new function to allow more control over the response checker

### DIFF
--- a/client/src/main/scala/org/http4s/client/middleware/Retry.scala
+++ b/client/src/main/scala/org/http4s/client/middleware/Retry.scala
@@ -161,9 +161,11 @@ object RetryPolicy {
   def recklesslyRetriable[F[_]](result: Either[Throwable, Response[F]]): Boolean =
     isErrorOrRetriableStatus(result)
 
+  /** Returns true if parameter is a Left or if the response contains a retriable status(as per HTTP spec) */
   def isErrorOrRetriableStatus[F[_]](result: Either[Throwable, Response[F]]): Boolean =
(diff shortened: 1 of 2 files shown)


client/src/main/scala/org/http4s/client/middleware/Retry.scala
```diff
@@ -162,8 +162,11 @@ object RetryPolicy {
     isErrorOrRetriableStatus(result)
 
   def isErrorOrRetriableStatus[F[_]](result: Either[Throwable, Response[F]]): Boolean =
+    isErrorOrStatus(result, RetriableStatuses)
+
+  def isErrorOrStatus[F[_]](result: Either[Throwable, Response[F]], status: Set[Status]): Boolean =
```
ChristopherDavenport, line 167: If we could get a comment on this vs the other use case. As these were the only option previously. A little documentation on how and why you might use it would be useful as it was only clear to me after I looked at the unit test, which isn't in the same place as the source.

Adriani277: @ChristopherDavenport  I have added a description. Let me know if it makes sense or if there is a better wording for it

ChristopherDavenport (approved):

rossabaker:
I think this can be cherry-picked back to 0.21 and released faster. I'll try that.

rossabaker:
This will be in the next 0.21 release (as well as 0.22 and 1.0).

PR #4644, opened 2021-03-25
Update to Servlet API 4 and Jetty 10

### LINKED ISSUE #4643: Evaluate update to Servlet 4 and Jetty 10
Current builds of Http4s are using Servlet 3.0 as build dependency. It would be nice to bump this to Servlet 4.0 before full move to `jakarta` namespaces and version 5.0.

Now it should be possible to bump Servlet version.
* Tomcat 9 implements Servlet 4.0 https://tomcat.apache.org/whichversion.html
* Jetty 10 also supports Servlet 4.0 https://www.eclipse.org/jetty/

Also, Jetty 10 kept `javax.servlet` namespace. So, upgrade should not be a breaking change at least for `jetty` server part.

If considered too dangerous the change could be skipped for `series/0.21`.

arixmkii:
Fixes #4643 

Include bump of Jetty version and fixes of multiple deprecation warnings in Jetty API.

So far no API changes or use of Servlet 4.0 specific APIs added.

Haven't tested it a lot yet, but at least it passed all the tests.

### DIFF
--- a/jetty-client/src/main/scala/org/http4s/client/jetty/JettyClient.scala
+++ b/jetty-client/src/main/scala/org/http4s/client/jetty/JettyClient.scala
@@ -23,7 +23,9 @@ import cats.syntax.all._
 import fs2._
 import org.eclipse.jetty.client.HttpClient
 import org.eclipse.jetty.client.api.{Request => JettyRequest}
+import org.eclipse.jetty.client.http.HttpClientTransportOverHTTP
 import org.eclipse.jetty.http.{HttpVersion => JHttpVersion}
+import org.eclipse.jetty.io.ClientConnector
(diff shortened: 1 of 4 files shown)


arixmkii:
mima checks fail for 
```
[error] jetty-client: Failed binary compatibility check against org.http4s:http4s-jetty-client_2.13:0.21.4! Found 2 potential problems
[error]  * class org.http4s.client.jetty.StreamRequestContentProvider does not have a correspondent in current version
```

For me it looked like internal detail, but I could be wrong here.

The change was needed to reflect that it now extends different class (former was deprecated in Jetty 10).

rossabaker:
This looks good.  I incoporated it into a PR based on series/0.22: #4658

PR #4646, opened 2021-03-25
Remove private from no-cache set

### LINKED ISSUE #4645: Remove "private" directive from `no-cache` header set
It is added as part of multiple directives in `Caching` middleware. https://github.com/http4s/http4s/blob/62a4b0d8e632537794a322b992c61dc02c10b013/server/src/main/scala/org/http4s/server/middleware/Caching.scala#L50

But according to MDN it should not be used for cache prevention https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#cacheability

> If you mean to not store the response in any cache, use no-store instead.

It is a one line change. If the idea is accepted I can create PR for `series/0.21`.

arixmkii:
Fixes #4645

### DIFF
--- a/server/src/main/scala/org/http4s/server/middleware/Caching.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/Caching.scala
@@ -57,7 +57,6 @@ object Caching {
     `Cache-Control`(
       NonEmptyList.of[CacheDirective](
         CacheDirective.`no-store`,
-        CacheDirective.`private`(),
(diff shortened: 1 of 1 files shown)


rossabaker (approved):
I think you're correct here.  There's a tradition of using many directives to cover all versions of all common user agents, but `private` is a contradiction, and I can't find anything that recommends it for what we're trying to accomplish here.

PR #4650, opened 2021-03-26
Fix typo in implicit conversion

kubukoz:
This is a binary breaking change, potentially source-breaking if someone imports these by name.


rossabaker:
Thanks! I picked this back to series/0.22.

PR #4652, opened 2021-03-27
Support providing custom error handler

dozed:
With this a `Request[F] => PartialFunction[Throwable, F[Response[F]]]` can be provided to the `ErrorHandling` middleware in order to map errors to responses.

### DIFF
--- a/server/src/test/scala/org/http4s/server/middleware/ErrorHandlingSuite.scala
+++ b/server/src/test/scala/org/http4s/server/middleware/ErrorHandlingSuite.scala
@@ -16,6 +16,7 @@
 
 package org.http4s.server.middleware
 
+import cats.data.OptionT
 import cats.effect.IO
 import org.http4s._
 import org.http4s.dsl.io._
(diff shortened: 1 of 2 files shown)


server/src/test/scala/org/http4s/server/middleware/ErrorHandlingSuite.scala
```diff
@@ -54,4 +55,47 @@ class ErrorHandlingSuite extends Http4sSuite {
...
+    ).orNotFound
+      .apply(request)
+      .map(_.status)
+      .assertEquals(Status.BadRequest)
+  }
+
+  test("Handle errors with a custom error handler via httpRoutes") {
+    ErrorHandling.httpRoutes(
+      routes(ErrorCode.CouldNotParseRequest),
+      (req: Request[IO]) => {
```
dozed, line 79: A type hint needs to be supplied, the compiler seems not to be able to infer that it is `Request[IO]`

ScalaWilliam: You should be able to do what you want by having `errorHandler` in another argument group

ScalaWilliam:
What was the use case that motivated this? Would be worth documenting as I would not spot this;

I'm thinking about these cases:
- If the middleware is producing the error, rather than the underlying code, should it not be done directly in the middleware, especially with regards to leaking out exceptions and handling them at user code? 
- If it's the user code producing an error, and the error is not being handled by the middleware already, then it is not to do with middleware, and could be done with `serviceErrorHandler` at `BlazeBuilder` level.

Thanks :-)

rossabaker:
I think the original idea behind this middleware was to phase out the error handlers on the builders, since it's more of a portable concern across backends?  I thought this had a customizable error function already, so clearly my memory fails me.  @ChristopherDavenport, do you remember the motiviation?

There's something appealing about any unhandled error being a 500, but there's also something appealing about solid default behavior where we return intelligent 4xx errors from the decoders, which is what the defaults do.

PR #4654, opened 2021-03-27
FileService: Querying /file.png/anything where /file.png is a file throws

ScalaWilliam:
Exception is `java.nio.file.FileSystemException: Not a directory`.
Found bot queries in Scala Algorithms logs that are leading to an HTTP 5xx error.

To avoid this, we explicitly check if what we are querying is a file, before attempting to resolve the real path; the exception is currently raised in particular where `.toRealPath` is called.

Alternative approach would be to capture the `FileSystemException` as well, but it could hide other errors like `AccessDeniedException`. Because we already do a check, the check for `NoSuchFileException` becomes redundant.

Before changing main code:
image

### DIFF
--- a/server/src/main/scala/org/http4s/server/staticcontent/FileService.scala
+++ b/server/src/main/scala/org/http4s/server/staticcontent/FileService.scala
@@ -22,8 +22,7 @@ import cats.data.{Kleisli, NonEmptyList, OptionT}
 import cats.effect.{Blocker, ContextShift, Sync}
 import cats.syntax.all._
 import java.io.File
-import java.nio.file.NoSuchFileException
-import java.nio.file.{LinkOption, Path, Paths}
+import java.nio.file.{Files, LinkOption, NoSuchFileException, Path, Paths}
 import org.http4s.headers.Range.SubRange
 import org.http4s.headers._
 import org.http4s.server.middleware.TranslateUri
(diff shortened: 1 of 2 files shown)


server/src/main/scala/org/http4s/server/staticcontent/FileService.scala
```diff
@@ -86,13 +85,14 @@ object FileService {
               case _ => OptionT.none
             }
           resolvedPath
-            .semiflatMap(path => F.delay(path.toRealPath(LinkOption.NOFOLLOW_LINKS)))
+            .flatMapF(path =>
+              F.delay(if (Files.exists(path)) Some(path.toRealPath(LinkOption.NOFOLLOW_LINKS))
```
rossabaker, line 89: Can we add the same LinkOption to the `exists` call to make it consistent with the `toRealPath`?

rossabaker (changes requested):
Eek.  Nice find.

This could be a separate PR, but I wonder whether we have the same issue in WebJar and Resource service.

server/src/main/scala/org/http4s/server/staticcontent/FileService.scala
ScalaWilliam: Done - thank you for pointing it out.

rossabaker:
I still think this is potentially an issue in the two related services, but this is good on its own.

PR #4656, opened 2021-03-29
Upgrade to jetty-9.4.39

rossabaker:


anilkumarmyla:
FWIW, after upgrading to http4s 0.21.21, we ran into https://github.com/eclipse/jetty.project/issues/6121 when using jetty client, had to downgrade to 0.21.20

rossabaker:
Thanks for the report.  I guess let's see if they can fix this in the 9.4 series.  I'll keep an eye on the other ticket.

PR #4658, opened 2021-03-29
Servlet 5.0 and friends

rossabaker:
Supersedes #4644.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -347,7 +347,7 @@ lazy val servlet = libraryProject("servlet")
     description := "Portable servlet implementation for http4s servers",
     startYear := Some(2013),
     libraryDependencies ++= Seq(
-      javaxServletApi % Provided,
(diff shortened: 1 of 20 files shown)


rossabaker:
If we do this before we split the repository, we'll need everyone compiling on JDK11, and have to exempt the Jetty modules from JDK8 in the build matrix.

rossabaker:
And we use Jetty in tests, so the plot thickens some more.

rossabaker:
We could use the `com.sun.net.httpserver` for tests.  We'll have to spin both Jetty modules off into their own repo, which is imminent anyway with the Great Schism.  I guess we're stuck on Servlet 3 in our 0.22/0.23 timeframe, but can start removing those Jetty test dependencies anytime.

armanbilge:
Superseded by https://github.com/http4s/http4s-servlet.

PR #4663, opened 2021-03-29
Upgrade sbt-twirl to v1.5.1

kusaeva:


rossabaker (approved):
:+1: on green.  Sorry we missed this on the first pass.

PR #4679, opened 2021-03-30
Upgrade lots of things

rossabaker:
Notably, cats-2.5.0, cats-effect-2.4.1, plus a bunch of internals.


rossabaker:
mdoc-2.2.19 is stripping our backticks.  We're trying to do the same, but we're not there yet!

PR #4699, opened 2021-04-03
Add custom dns resolver to BlazeClientBuilder.

JakDar:
Hi, AFAIK there is currently no way to add custom DNS resolver for blaze-client (similar to one provided in akka-http ([this PR](https://github.com/akka/akka-http/pull/3202) and [here in docs](https://github.com/akka/akka-http/blob/master/docs/src/main/paradox/client-side/client-transport.md#custom-host-name-resolution-transport)).
I kinda need that feature, potential use cases are in linked akkka-http doc.

I added new parameter in `BlazeClientBuilder` - `customDnsResolver` - which is passed to `Http1Support` and if present - used instead of [getAddress](https://github.com/http4s/http4s/blob/e522b34ad9dbf6866a8faddb6e102503ac5822a0/blaze-client/src/main/scala/org/http4s/client/blaze/Http1Support.scala#L137).

 I'm glad to adjust/change implementation if it is not suitable.

Also, are changes from `series/0.21` eventually merged to 1.0.0 series (`main`), or should I submit separate PR for it?

### DIFF
--- a/blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala
+++ b/blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala
@@ -322,7 +322,7 @@ sealed abstract class BlazeClientBuilder[F[_]] private (
       userAgent = userAgent,
       channelOptions = channelOptions,
       connectTimeout = connectTimeout,
-      customDnsResolver = customDnsResolver
+      getAddress = customDnsResolver.getOrElse(BlazeClientBuilder.getAddress(_))
     ).makeClient
     Resource.make(
       ConnectionManager.pool(
@@ -370,6 +370,14 @@ object BlazeClientBuilder {
       channelOptions = ChannelOptions(Vector.empty),
       customDnsResolver = None
     ) {}
+
+  def getAddress(requestKey: RequestKey): Either[Throwable, InetSocketAddress] =
+    requestKey match {
+      case RequestKey(s, auth) =>
+        val port = auth.port.getOrElse(if (s == Uri.Scheme.https) 443 else 80)
+        val host = auth.host.value
+        Either.catchNonFatal(new InetSocketAddress(host, port))
+    }
 
   private def tryDefaultSslContext: Option[SSLContext] =
     try Some(SSLContext.getDefault())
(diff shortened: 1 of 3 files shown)


blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala
```diff
@@ -57,7 +58,8 @@ sealed abstract class BlazeClientBuilder[F[_]] private (
     val executionContext: ExecutionContext,
     val scheduler: Resource[F, TickWheelExecutor],
     val asynchronousChannelGroup: Option[AsynchronousChannelGroup],
-    val channelOptions: ChannelOptions
+    val channelOptions: ChannelOptions,
+    val customDnsResolver: Option[RequestKey => Either[Throwable, InetSocketAddress]]
```
rossabaker, line 62: Not sure this needs to be an `Option`: could we just have a default based on the `getAddress` that's used today?

rossabaker, line 62: Something for 1.0: we'll probably want to refine this to the ip4s address type once it becomes public API.

Well, maybe not, if blaze is just using `java.net` internally.  It's a bigger issue for things with deeper fs2 integration.  Anyway, something to think about.

rossabaker (changes requested):
Thanks!  Yes, this is eventually merged to series/0.22 and main, and would be in all three releases.  Starting it here lets you use it in a stable release ASAP.

Binary compatibility is broken because the constructor for BlazeClientBuilder changed.  We can fix that with a deprecated auxiliary constructor ([example](https://github.com/http4s/http4s/blob/6d4ce9273285f343b5acb86d77a3deed3f53e5b9/ember-server/src/main/scala/org/http4s/ember/server/EmberServerBuilder.scala#L42-L43), or I guess just add a MiMa exception since it's private.  I usually prefer the deprecated constructors, just to be 100% sure.

This might be a compelling feature in Ember, too, but we can handle that in a separate PR.

blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala
JakDar: Done. Also moved  `getAddress` to `BlazeClientBuilder`.

rossabaker (approved):
Thanks!  I expect a release train for Scala 3.0.0-RC2 support in the next few days, and we'll include 0.21 as well to pick this up.

PR #4726, opened 2021-04-07
0.21 client test munit suite fixture

hamnis:


testing/src/test/scala/org/http4s/Http4sSuite.scala
```diff
@@ -32,6 +32,19 @@ trait Http4sSuite extends CatsEffectSuite with DisciplineSuite with munit.ScalaC
   // BatchExecutor on Scala 2.12.
   override val munitExecutionContext =
     ExecutionContext.fromExecutor(newDaemonPool("http4s-munit", min = 1, timeout = true))
+
+  private[this] val suiteFixtures = List.newBuilder[Fixture[_]]
```
rossabaker, line 36: I'm a little surprised this is so DIY, but I'm not about to argue with success.

build.sbt
```diff
@@ -358,7 +358,8 @@ lazy val asyncHttpClient = libraryProject("async-http-client")
       nettyBuffer,
       nettyCodecHttp,
       reactiveStreams,
-    )
+    ),
+    Test / parallelExecution := false
```
rossabaker, line 362: Would be interesting to figure out why this must be, but that's for another time.

rossabaker (approved):
Tremendous work finding this.

PR #4729, opened 2021-04-10
Revert to async-http-client-2.12.1

rossabaker:
Based on what I understand to be the fruits of @hamnis' investigation.


rossabaker:
This ain't it.

PR #4735, opened 2021-04-11
Buffered messages in Ember

RaasAhsan:
Goals of this PR:
1. Considerably simplify how parsers work. In particular, we can avoid reentrancy by reading the HTTP message up until `\r\n\r\n` and then parse the entire message together. The memory overhead remains linear here as before.
2. Handle max header/prelude sizes more clearly.
3. Make the first step towards using more functional parsers. Since we're parsing the entire HTTP message at once, we can theoretically describe a parser for request/response messages and defer to that altogether. This is complicated by the fact that series/0.21 is still on parboiled.
4. Perhaps improve performance a bit. I haven't ran any benchmarks yet, but I would imagine that we get some wins by eliminating reentrancy.

TODO:
- [x] Tests
- [x] More error types


ember-core/src/main/scala/org/http4s/ember/core/ChunkedEncoding.scala
```diff
@@ -29,6 +29,7 @@ private[ember] object ChunkedEncoding {
   def decode[F[_]](
       head: Array[Byte],
       read: F[Option[Chunk[Byte]]],
+      maxHeaderSize: Int,
```
RaasAhsan, line 32: I think we shouldn't conflate `maxHeaderSize` (the max bytes of the trailer header section) and with what `maxChunkHeaderSize` is (the max size for chunked bodies?). I'm thinking we should probably bubble the latter out into a separate parameter. Maybe even a separate one for trailers?

ember-core/src/main/scala/org/http4s/ember/core/Parser.scala
```diff
@@ -27,57 +27,41 @@ import scala.collection.mutable
...
-            ParseHeadersError(new Throwable("Reached End of Stream Looking for Headers")))
+  object MessageP {
+
+    val emptyStreamError = new Throwable("Cannot Parse Empty Stream")
+
+    private val cr: Byte = '\r'.toByte
+    private val lf: Byte = '\n'.toByte
+    private val DoubleCrlf: Seq[Byte] = Seq(cr, lf, cr, lf)
+
+    def parseMessage[F[_]](buffer: Array[Byte], read: F[Option[Chunk[Byte]]], maxHeaderSize: Int)(
```
RaasAhsan, line 40: I think this can be made better in a followup. We don't need to linearly scan the entire array each time we get a new chunk, we can start from 3 bytes before the existing buffer.

ember-core/src/main/scala/org/http4s/ember/core/ChunkedEncoding.scala
rossabaker: A separate one for trailers sounds good.  I'm still confused by maxChunkHeaderSize.

ember-core/src/main/scala/org/http4s/ember/core/ChunkedEncoding.scala
```diff
@@ -92,24 +95,28 @@ private[ember] object ChunkedEncoding {
...
-              (headers, bytes)
-          }
+  )(buffer: Array[Byte], read: F[Option[Chunk[Byte]]]): F[Trailers] =
+    if (buffer.startsWith(Shared.`\r\n`.toArray)) {
+      Trailers(Headers.empty, buffer.drop(`\r\n`.size.toInt)).pure[F]
+    } else if (buffer.length < 2) {
+      read.flatMap {
+        case None =>
+          // TODO: end of stream?
+          Trailers(Headers.empty, Array.emptyByteArray).pure[F]
```
rossabaker, line 109: This is going to be common, right?  Is it worth caching an empty value of this?

ember-core/src/main/scala/org/http4s/ember/core/ChunkedEncoding.scala
RaasAhsan: So it looks like `maxChunkHeaderSize` is the max size of the chunk header for each chunk. so `4\r\n` has a length of 3.  Is it worth having a separate configuration parameter for that? Right now it's set to `maxHeaderSize`

RaasAhsan: I'm going to hold off on adding these for now... We're slowly reaching the limit of class parameters we can take (we're at like 15 or 16 right now) so I want to group things together a bit to make it nicer

rossabaker:
@ChristopherDavenport knows this best, but I'll take another look if he falls into a hole.

RaasAhsan:
Definitely want to get Chris' thoughts on this before merging since it's mostly an attempt to make things simpler (though it fixes several bugs)

ember-core/src/main/scala/org/http4s/ember/core/Parser.scala
```diff
@@ -179,130 +135,52 @@ private[ember] object Parser {
...
-          idx,
-          state,
-          Option(name),
-          start,
-          chunked,
-          contentLength)
+      if (throwable != null) {
+        F.raiseError(ParseHeadersError(throwable))
+      } else if (!complete) {
+        F.raiseError(IncompleteHttpMessage())
```
rossabaker, line 141: Is there any additional context we might provide here?

ember-core/src/main/scala/org/http4s/ember/core/Parser.scala
```diff
@@ -542,11 +384,22 @@ private[ember] object Parser {
...
+          F.raiseError(RespPreludeError("Encountered Error parsing", Option(throwable)))
+        else if (httpVersion == null || status == null)
+          F.raiseError(RespPreludeError("Failed to parse HTTP response prelude", None))
+        else
+          RespPrelude(httpVersion, status, idx).pure[F]
       }
+
+      private val space = ' '.toByte
+      private val cr: Byte = '\r'.toByte
+      private val lf: Byte = '\n'.toByte
```
rossabaker, line 397: I think this is maybe the third time I've seen these defined, but I guess it's not like they're big or going to change.

ember-core/src/main/scala/org/http4s/ember/core/Parser.scala
```diff
@@ -542,11 +384,22 @@ private[ember] object Parser {
...
+
+      case class RespPreludeError(message: String, cause: Option[Throwable])
+          extends Throwable(
```
rossabaker, line 400: `Throwable`?  Seems like `Exception`.

I'd make all these `RuntimeException`, but that's the old Java programmer in me.

ember-core/src/main/scala/org/http4s/ember/core/Parser.scala
RaasAhsan: Maybe they can be pulled into `Shared`

ember-core/src/main/scala/org/http4s/ember/core/Parser.scala
RaasAhsan: I added in the partially parsed headers. We could go further and produce an error that reflects what the parsing state was, but I'm more inclined to leave that to a parser library once we switch to that

rossabaker (approved):
Would still love to get @ChristopherDavenport's feedback, but it LGTM.

PR #4736, opened 2021-04-11
Add ArrayStreamJsonDecoder method to http4s-circe

rohan-sircar:
Add `streamJsonArrayDecoder` counterpart to `streamJsonArrayEncoder` methods in `CirceInstances`. ~It delegates to byteArrayParser method in circe-fs2.~

### DIFF
--- a/circe/src/main/scala/org/http4s/circe/CirceInstances.scala
+++ b/circe/src/main/scala/org/http4s/circe/CirceInstances.scala
@@ -142,10 +142,9 @@ trait CirceInstances extends JawnInstances {
     streamJsonArrayEncoderWithPrinter(defaultPrinter)
 
   implicit def streamJsonArrayDecoder[F[_]: Sync]: EntityDecoder[F, Stream[F, Json]] =
-    EntityDecoder.byteArrayDecoder[F].map { array =>
-      fs2.byteArrayParser[F].apply(Stream.fromIterator(array.iterator))
+    EntityDecoder.decodeBy(MediaType.application.json) { media =>
+      DecodeResult.successT(media.body.chunks.through(jawnfs2.unwrapJsonArray))
     }
-
   /** An [[EntityEncoder]] for a [[fs2.Stream]] of JSONs, which will encode it as a single JSON array. */
   def streamJsonArrayEncoderWithPrinter[F[_]](printer: Printer): EntityEncoder[F, Stream[F, Json]] =
     EntityEncoder
--- a/circe/src/main/scala/org/http4s/circe/CirceInstances.scala
+++ b/circe/src/main/scala/org/http4s/circe/CirceInstances.scala
@@ -141,7 +141,6 @@ trait CirceInstances extends JawnInstances {
   implicit def streamJsonArrayEncoder[F[_]]: EntityEncoder[F, Stream[F, Json]] =
     streamJsonArrayEncoderWithPrinter(defaultPrinter)
 
-
   implicit def streamJsonArrayDecoder[F[_]: Sync]: EntityDecoder[F, Stream[F, Json]] =
     EntityDecoder.byteArrayDecoder[F].map { array =>
       fs2.byteArrayParser[F].apply(Stream.fromIterator(array.iterator))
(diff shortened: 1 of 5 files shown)


circe/src/main/scala/org/http4s/circe/CirceInstances.scala
```diff
@@ -141,6 +141,11 @@ trait CirceInstances extends JawnInstances {
   implicit def streamJsonArrayEncoder[F[_]]: EntityEncoder[F, Stream[F, Json]] =
     streamJsonArrayEncoderWithPrinter(defaultPrinter)
 
+  implicit def streamJsonArrayDecoder[F[_]: Sync]: EntityDecoder[F, Stream[F, Json]] =
+    EntityDecoder.byteArrayDecoder[F].map { array =>
+      fs2.byteArrayParser[F].apply(Stream.fromIterator(array.iterator))
+    }
+
```
rossabaker, line 148: Thanks!  Two things:

1. This is going to decode the entire body to an array, and then restream it.
2. We should be able to do this without the circe-fs2 dependency.  It's a fine library, but we already have jawn-fs2, which is just a more abstract version of it.

How about this:

```suggestion
  implicit def streamJsonArrayDecoder[F[_]: Sync]: EntityDecoder[F, Stream[F, Json]] =
    EntityDecoder.decodeBy(MediaType.application.json) { media =>
      DecodeResult.successT(media.body.chunks.through(jawnfs2.unwrapJsonArray))
    }
```

circe/src/test/scala/org/http4s/circe/CirceSuite.scala
```diff
@@ -140,6 +141,26 @@ class CirceSuite extends JawnDecodeSupportSuite[Json] {
     writeToString(jsons).assertEquals("""[{"test1":"CirceSupport"},{"test2":"CirceSupport"}]""")
   }
 
+  test("stream json array decoder should decode JSON") {
```
rossabaker, line 144: This is good.

What are the semantics if we have a stream that's ill-formed JSON?  I think we would return a successful stream that fails?  Otherwise, we're back to buffering to see if it's going to fail before we start streaming.  It would be neat to make this behavior clear with a test, though.

rossabaker (changes requested):
Nice start.  Thank you for adding this!  A couple comments below.

circe/src/main/scala/org/http4s/circe/CirceInstances.scala
rohan-sircar: > This is going to decode the entire body to an array, and then restream it.

whoops I got so preoccupied with following the types that I forgot about this ^^" 
> We should be able to do this without the circe-fs2 dependency

Ah I was hesitant to add it myself. Ought to have put more attention to what you said in Gitter.

Nice of you to give me the answer outright but I think I would've learned more if you gave me hints and I had to grind it out myself haha

circe/src/test/scala/org/http4s/circe/CirceSuite.scala
rohan-sircar: > What are the semantics if we have a stream that's ill-formed JSON?

I... didn't think about that for some reason. I'll check and add a test case for it. Thanks!

rohan-sircar: Currently, the IO blows up like this on malformed json:
```
==> X org.http4s.circe.test.CirceSuite.stream json array decoder should fail on malformed JSON  0.028s org.typelevel.jawn.ParseException: expected json value got 'C...' (line 1, column 36)
// stack trace
```

rossabaker: By "successful stream that fails," I mean you get a `DecodeResult` that's a success that's a stream.  But compiling that stream will raise the error.  That's what's happening?

rohan-sircar: > you get a DecodeResult that's a success that's a stream. But compiling that stream will raise the error

I pondered upon this for a while. That seems to be the case, yes. I added a few test cases, let me know if I got it right.

rossabaker (approved):
This looks good to me.  We'll have to wait for the Bintray Brownout to finish before CI will work again.

rohan-sircar:
@rossabaker thanks for the guidance and the opportunity to contribute! I'll be on the lookout for more things I could meaningfully contribute.

Issue #2702, opened 2019-07-09
Package structure: org.http4s.$role.$backend or org.http4s.$backend.$role 


rossabaker:
This discussion [originated in the ember PR](https://github.com/http4s/http4s/pull/2696#discussion_r300818868), but we didn't have a strong consensus.  Ember is `org.http4s.ember.client`.  Blaze (and all other backends) are `org.http4s.client.blaze`.  There are good reasons for both, but I think consistency is better than either argument.  Let's standardize on one.

Quoting a fool:

> One thing that we all got right is to have both $backend and (server | client) in the package name. The Jetty server is older than the Jetty client, but we were able to add the client without moving the server later. So it's really a matter of which comes first. I'd like us to be consistent. I'm willing to consider deprecating and moving the other backends with a scalafix.
>
> 1. There's more package-private code in ember than there is in server and client, so ember benefits from this convention. The alternative is to leave ember-core in org.http4s.ember, but make everything private[http4s].
> 2. There's also quite a bit in blaze, which could benefit from backend-first. But org.http4s.blaze is the blaze project's package, so we need to be very careful about polluting that namespace. We've solved that today with a blazecore for the shared code, all private[http4s]. We could maybe move non-http4s blaze out of org.http4s, but to where? That's a lot of moving for something nobody has complained about.
> 3. There's no shared http4s-jetty code between server and client. If there were, the ember convention would be a little nicer.
> 4. No other backend has both a server and client. These can all benefit slightly from the current convention, with package-private server and client helpers. But the one and only that exists right now is DefaultClient.
>
> Points 3 and 4 are weak. 1 is a good argument for change. 2 is a good argument for the status quo.

rossabaker:
If ember stands its ground, we'll need to scalafix the heck out of a lot of other things.

rossabaker:
I can't find a single instance of us sharing private server or client components between backends, but many instances in blaze and ember of sharing backend components between server and client.  That suggests `$backend.$role`.

$backend will also more naturally reflect our repo organization, where `org.http4s.$backend` usually lives in `http4s-$backend` repo.

I think we should move in this direction, and I know @ChristopherDavenport will be happy.

rossabaker:
#4702 is a good template for this, and we could use some help.

rossabaker:
* [ ] client.asynchttpclient -> asynchttpclient.client
* [ ] client.blaze -> blaze.client
* [ ] blazecore -> blaze 
[...]

PR #4738, opened 2021-04-12
server.tomcat -> tomcat.server

rossabaker:
See #2702


rossabaker:
Needs a scalafix, and then it can be used as an example for others.

tomcat/src/main/scala/org/http4s/server/tomcat/TomcatBuilder.scala
```diff
@@ -31,10 +31,18 @@ import org.apache.catalina.util.ServerInfo
 import org.apache.coyote.AbstractProtocol
 import org.apache.tomcat.util.descriptor.web.{FilterDef, FilterMap}
 import org.http4s.internal.CollectionCompat.CollectionConverters._
+import org.http4s.server.{
+  DefaultServiceErrorHandler,
+  SSLClientAuthMode,
+  Server,
+  ServerBuilder,
+  ServiceErrorHandler
+}
```
rossabaker, line 40: So we lose these from the nested namespace, but whatever. They're public.

Issue #3233, opened 2020-03-04
Pre-signed AWS URIs expect no decoding of %2F ('/')


glammr1:
Description: Parsing URIs is correct according to RFC 3986, but AWS expects its pre-signed AWS URIs to be used as-is without URL-decoding in the query string.

Reproducible example (using ammonite with http4s 0.21.1):

```
Uri.fromString("https://bucket.s3.amazonaws.com/object?uploadId=.KFey.IJHLCmyRpxkyssU9KS0hh0TC9Z6ALXh1y.M8S44xAT1c19kIfwqYQu2kiYIXCyvFlPTHnahpag9z0VBVL790o3WnmB.qge9ly3jXazG0N1bnc5whj5c_P0DXPn&partNumber=1&X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Date=20200304T142549Z&X-Amz-SignedHeaders=content-length%3Bcontent-md5%3Bhost&X-Amz-Expires=119&X-Amz-Credential=AKXXXXXXXXXXXXXXXXXX%2F20020304%2Fus-east-1%2Fs3%2Faws4_request&X-Amz-Signature=e2437fe26dcef3f5118150b1c316ca286098335981256e421a3ff676489ea703")
res0: ParseResult[Uri] = Right(
  Uri(
    Some(Scheme(https)),
    Some(Authority(None, RegName(bucket.s3.amazonaws.com), None)),
    "/object",
    uploadId=.KFey.IJHLCmyRpxkyssU9KS0hh0TC9Z6ALXh1y.M8S44xAT1c19kIfwqYQu2kiYIXCyvFlPTHnahpag9z0VBVL790o3WnmB.qge9ly3jXazG0N1bnc5whj5c_P0DXPn&partNumber=1&X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Date=20200304T142549Z&X-Amz-SignedHeaders=content-length%3Bcontent-md5%3Bhost&X-Amz-Expires=119&X-Amz-Credential=AKXXXXXXXXXXXXXXXXXX/20020304/us-east-1/s3/aws4_request&X-Amz-Signature=e2437fe26dcef3f5118150b1c316ca286098335981256e421a3ff676489ea703,
    None
  )
)
```

Note that the `%2F` is translated to `/` in the query string.  The corresponding renderer then does not re-escape the `/` back into `%2F` and AWS rejects the URI with a 400, presumably because it cannot verify the signature.

[...]

Issue #1813, opened 2018-04-26
Query Render Innappropriate with [ and ]


ChristopherDavenport:
```
@ Uri.uri("""http://www.google.com:8085/?query[event][eq]=something""")
res10: Uri = Uri(Some(Scheme(http)), Some(Authority(None, RegName(www.google.com), Some(8085))), "/", Query(("query[event][eq]", Some("something"))), None)

@ println(res10)
http://www.google.com:8085/?query%5Bevent%5D%5Beq%5D=something
```

This appears because of our inability to render in the passing steps. 

First Uri render - 
https://github.com/http4s/http4s/blob/master/core/src/main/scala/org/http4s/Uri.scala#L108
And then in the utils
https://github.com/http4s/http4s/blob/master/core/src/main/scala/org/http4s/util/UrlCoding.scala#L39

I am inclined to believe that this should be one of the skipped characters for a query.

kevinmeredith:
Hi @ChristopherDavenport. Regarding:

```
@ Uri.uri("""http://www.google.com:8085/?query[event][eq]=something""")
res10: Uri = Uri(Some(Scheme(http)), Some(Authority(None, RegName(www.google.com), Some(8085))), "/", Query(("query[event][eq]", Some("something"))), None)

@ println(res10)
http://www.google.com:8085/?query%5Bevent%5D%5Beq%5D=something
```

what's the problem?

I'm asking since I don't fully understand this issue's reported bug per:

> I am inclined to believe that this should be one of the skipped characters for a query.

rossabaker:
Do we want to skip this in values, or just in keys?

Issue #2445, opened 2019-03-07
Slashes (`/`) are not URL-encoded in Uri string representation in query parameters


alex-iov42:
Hi all,

We have been using `http4s` in our project and we recently noticed that slashes (`/`) are not URL-encoded in query parameters.
For example, let's say we have the following URL:
```
/some/url?param1=1%2B2%2F3
```
where basically the value of `param1` is `1+2/3`.

What we get from the URI's string representation is:
```
scala> val uri = Uri.uri("/some/url?param1=1%2B2%2F3")
uri: org.http4s.Uri = /some/url?param1=1%2B2/3
```

The reason we need this is that our clients are expected to sign the URL-encoded uri.
However, most of the other libraries and tools, URL-encode slashes properly when they appear in query parameters, which causes us problems when trying to validate our clients' signatures.

Is there a reason why this is happening? Is it a bug or a conscious decision? And if it's the latter, what is the reasoning behind it?

kevinmeredith:
I'm not disputing this issue, @alex-iov42, but, it appears to be, per the following example, that `Uri#withQueryParam` works as, I believe, you expect:

```scala
$amm
Loading...
Welcome to the Ammonite Repl 1.0.3
(Scala 2.12.4 Java 1.8.0_112)
If you like Ammonite, please support our development at www.patreon.com/lihaoyi
@ import $ivy.`org.http4s::http4s-core:0.20.0` 
imimport $ivy.$                               

@ import org.http4s.Uri 
import org.http4s.Uri

@ Uri.uri("a.b.c").withQueryParam("param1", "1%2B2%2F3").renderString 
res2: String = "a.b.c?param1=1%252B2%252F3"
```

kevinmeredith:
[Uri#uri](https://github.com/http4s/http4s/blob/master/core/src/main/scala/org/http4s/Uri.scala#L128-L129) uses `Uri#fromString`.

[Uri#fromString](https://github.com/http4s/http4s/blob/master/core/src/main/scala/org/http4s/Uri.scala#L143-L147) includes a comment:

> /** Decodes the String to a [[Uri]] using the RFC 3986 uri decoding specification */


https://tools.ietf.org/html/rfc3986#section-3.4 notes:

>    The characters slash ("/") and question mark ("?") may represent data
   within the query component.  Beware that some older, erroneous
   implementations may not handle such data correctly when it is used as
   the base URI for relative references (Section 5.1), apparently
   because they fail to distinguish query data from path data when
[...]

PR #4739, opened 2021-04-12
Postpone using Query model until we need it

hamnis:
Should make #1813 #3277 and  #3233 work better on 0.22+.
I think this also was done on 0.21 a while ago, must have been lost in the parsing update.

The test failure assumes that we encode the `[]` brackets. this change would then pass them through, since we have allowed them in the parser.

### DIFF
--- a/core/src/main/scala/org/http4s/Query.scala
+++ b/core/src/main/scala/org/http4s/Query.scala
@@ -234,7 +234,7 @@ object Query {
     import cats.parse.Parser.charIn
     import Rfc3986.pchar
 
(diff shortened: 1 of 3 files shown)


hamnis:
Updating the test to just reflect that we pass on the brackets, then this will pass.

hamnis:
also should fix #3233

hamnis:
Also deals with #2445

m-sp (approved):

ChristopherDavenport (approved):

Issue #4733, opened 2021-04-11
Extract parsers into module below core


rossabaker:
Related to @bblfish's work in #4716: can we extract a sensible parser module to sit beneath http4s-core and share with others?

Why:
* akka-http uses parboiled2.  That's will be an obstacle to their Scala 3 support, and they might be able to reuse our work.
* If they can reuse our work, we'd have more eyes on it and probably catch some latent bugs in our work, and have shared maintenance moving forward.

Why not:
* Our core model is not shared with anybody else, and we parse to a model.  What exactly is it parsing to if http4s-core sits on top of it?  

Realistically:
* It's probably limited to a few RFC helpers?  Still not insubstantial, but not exactly All The Headers.

bblfish:
I had the following sketch in mind.
An HTTP parser is at base a function `parse: ByteSeq => example.lib.HttpHeader`
One could split this in two:

```scala
def syntaxParse: ByteSeq => shared.HeaderStructure
def semanticParse: shared.HeaderStructure => example.lib.HttpHeader
 ```

The syntax layer is something like [Rfc8941Types](https://github.com/co-operating-systems/Reactive-SoLiD/blob/master/src/main/scala/run/cosy/http/headers/Rfc8941.scala#L103). 

Different implementations then essentially provide functions for the second interpretation step `semanticParse`, though my guess is a lot could be re-used across implementations.
This would be maps of header names to functions, each with different range depending on the domain.  That looks like a dependent Sum.

I would love to be able to point JS and Java devs to tools they can go to for Signing Http Messages written in Scala :-) I will also need to write clients in the browser where Akka is clearly too heavy. Next I'll try using this for [Signing Http Messages](https://tools.ietf.org/html/draft-ietf-httpbis-message-signatures-03) which depends on RFC8941.

rossabaker:
Created a though experiment at #4745, so we're talking about something more concrete.

jrudolph:
@rossabaker thanks alot for evaluating the possibility of sharing parsing logic. I also had a quick look into how that would look.

Under the assumption that each user of the library would provide their own model, what would these parsers parse to? It would either have to be an intermediate model or some kind of GADT which can then be postprocessed into the final model. In #4745, you showed that for a simple product type but we will also need sum/record types.

Given that header parsing is not really difficult semantically, right now the parsers consist more of glue code than actual parsing code. A lot of glue code would still be necessary to make use of those shared parsers.

[...]

PR #4745, opened 2021-04-14
Extract parsing library

rossabaker:
An experimental take on #4733.

`http4s-parsing` has no dependencies besides cats-parse, even `http4s-core`.  If successful, this would go to live in its own repo.

### DIFF
--- a/parsing/src/main/scala/org/http4s/parsing/Rfc7230.scala
+++ b/parsing/src/main/scala/org/http4s/parsing/Rfc7230.scala
@@ -1,0 +1,33 @@
+/*
+ * Copyright 2021 http4s.org
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.http4s.parsing
+
+import cats.parse.Parser
+import cats.parse.Parser.{char, string}
+import cats.parse.Rfc5234.digit
+
+object Rfc7230 {
+  /* @see [[https://tools.ietf.org/html/rfc7230#section-2.6 RFC7230, Section 2.6]]*/
+  val httpName =
+    // HTTP-name = %x48.54.54.50 ; "HTTP", case-sensitive
+    string("HTTP")
+
+  /* @see [[https://tools.ietf.org/html/rfc7230#section-2.6 RFC7230, Section 2.6]]*/
+  val httpVersion: Parser[(Int, Int)] =
+    // HTTP-version = HTTP-name "/" DIGIT "." DIGIT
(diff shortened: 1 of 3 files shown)


parsing/src/main/scala/org/http4s/parsing/Rfc7230.scala
```diff
@@ -0,0 +1,33 @@
...
+import cats.parse.Rfc5234.digit
+
+object Rfc7230 {
+  /* @see [[https://tools.ietf.org/html/rfc7230#section-2.6 RFC7230, Section 2.6]]*/
+  val httpName =
+    // HTTP-name = %x48.54.54.50 ; "HTTP", case-sensitive
+    string("HTTP")
+
+  /* @see [[https://tools.ietf.org/html/rfc7230#section-2.6 RFC7230, Section 2.6]]*/
+  val httpVersion: Parser[(Int, Int)] =
```
rossabaker, line 30: This is where the lack of http4s-core hurts.

We can use anonymous tuples like this, which isn't really so bad for two, but is going to hit hard on more complex types.  Or we could design a set of anemic case classes, but that will create more garbage: from tuple to case class, and from case class to consuming library's richer model.

bblfish: I don't have good idea of what http4s core would do, so this is probably a silly question: Could Scala3's opaque types help?

rossabaker: Eh, maybe?  We'd have an extra map call.  I don't know if we can cast from a representation to an opaque type and avoid an allocation, so it might effectively be just a `.map(identity)`.

Bigger problem is that would be a Scala 3 solution, and at least http4s will support Scala 2 for some time to come.

bblfish: Re, Scala 2, I read on the [Scala 3 Source level page](https://scalacenter.github.io/scala-3-migration-guide/docs/compatibility/source.html) that
>  But amazingly, this new Scala 3 artifact can be consumed as a dependency in Scala 2.13.

armanbilge: @rossabaker what if we introduced something like:
```scala
trait Rfc7230[M] {
  def create(major, minor): M
}
```

And then define something like:
```scala
def httpVersionParser[M: Rfc7230]: Parser[M] = 
  httpVersion.map { case (major, minor) => Rfc7230[M].create(major, minor) }
```

Along with this in core:
```scala
implicit object rfc7230ForHttpVersion extends Rfc7230[HttpVersion] {
  def create(major: Int, minor: Int) = new HttpVersion(major, minor)
}
```

armanbilge: And btw for extra points I think this optimization even avoids the extra lambda?

```scala
trait Rfc7230[M] extends ((Int, Int) => M) {
  final def apply(x: (Int, Int)): M = parse(x._1, x._2)
  def create(major, minor): M
}

def httpVersionParser[M](implicit toModel: Rfc7230[M]): Parser[M] = 
  httpVersion.map(toModel)
```

armanbilge:
FWIW I would have loved to see this happen, but parboiled for Scala 3 landed in https://github.com/sirthias/parboiled2/pull/280 and akka-http is squeaking its way in https://github.com/akka/akka-http/pull/4079.

Without anyone to share with, there doesn't seem to be much point.

PR #4746, opened 2021-04-14
Rename tomcat and jetty modules

### LINKED ISSUE #4744: Rename modules to jetty-server and tomcat-server
`jetty` has been `jetty` since before we had a `client`, but now we have a `jetty-client`, so it really ought to be called `jetty-server`.

And everything else is `*-server` or `*-client`, so Tomcat ought to be `tomcat-server`, too.

(Reminded of this by @vasilmkd's comment on #2702.)

I don't think we can scalafix an sbt script.  Fortunately, these modules are both <10% usage.  We'll just have to scream in the release notes.

vasilmkd:
Resolves #4744.


rossabaker (approved):
:tada:

PR #4749, opened 2021-04-14
Move asynchttpclient client package

vasilmkd:
Part of #2702.

Depends on #4748 being merged first.

### DIFF
--- a/async-http-client/src/main/scala/org/http4s/asynchttpclient/client/AsyncHttpClient.scala
+++ b/async-http-client/src/main/scala/org/http4s/asynchttpclient/client/AsyncHttpClient.scala
@@ -15,8 +15,8 @@
  */
 
 package org.http4s
-package client
 package asynchttpclient
+package client
 
 import cats.effect._
 import cats.effect.concurrent._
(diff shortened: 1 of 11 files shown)


async-http-client/src/main/scala/org/http4s/asynchttpclient/client/AsyncHttpClient.scala
```diff
@@ -15,8 +15,8 @@
  */
 
 package org.http4s
-package client
 package asynchttpclient
+package client
```
rossabaker, line 19: This one's a little weird because `client` is part of the name and it appears twice, but at least it's consistent.

rossabaker (approved):
:+1:, unless anyone wants to push back on the `*client.client` bit.

async-http-client/src/main/scala/org/http4s/asynchttpclient/client/AsyncHttpClient.scala
vasilmkd: Yeah, I noticed it too. :/

PR #4754, opened 2021-04-17
cats-mtl

rossabaker:
Highly experimental.

### DIFF
--- a/server/src/main/scala/org/http4s/server/middleware/Date.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/Date.scala
@@ -17,7 +17,6 @@
 package org.http4s.server.middleware
 
 import cats._
-import cats.data.Kleisli
 import cats.syntax.all._
 import cats.effect._
 
@@ -28,18 +27,15 @@ import org.http4s.headers.{Date => HDate}
   * by the service.
   */
 object Date {
-  def apply[G[_]: Monad: Clock, F[_], A](
-      k: Kleisli[G, A, Response[F]]): Kleisli[G, A, Response[F]] =
-    Kleisli { a =>
+  def apply[G[_]: Monad: Clock, F[_], A](k: G[Response[F]]): G[Response[F]] =
+    k.flatMap { resp =>
       for {
-        resp <- k(a)
-        header <-
-          resp.headers
-            .get[HDate]
-            .fold(
-              HttpDate.current[G].map(HDate(_))
-            )(_.pure[G])
-      } yield resp.putHeaders(header)
+        date <- resp.headers
+          .get[HDate]
+          .fold(
+            HttpDate.current[G].map(HDate(_))
+          )(_.pure[G])
+      } yield resp.putHeaders(date)
     }
 
   def httpRoutes[F[_]: Monad: Clock](routes: HttpRoutes[F]): HttpRoutes[F] =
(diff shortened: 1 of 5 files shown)


server/src/main/scala/org/http4s/server/middleware/Date.scala
```diff
@@ -28,18 +27,15 @@ import org.http4s.headers.{Date => HDate}
   * by the service.
   */
 object Date {
-  def apply[G[_]: Monad: Clock, F[_], A](
-      k: Kleisli[G, A, Response[F]]): Kleisli[G, A, Response[F]] =
-    Kleisli { a =>
+  def apply[G[_]: Monad: Clock, F[_], A](k: G[Response[F]]): G[Response[F]] =
```
rossabaker, line 30: This cleans up nicely, and it's not mtl at all!

server/src/main/scala/org/http4s/server/middleware/RequestId.scala
```diff
@@ -38,38 +38,37 @@ object RequestId {
 
   val requestIdAttrKey: Key[String] = Key.newKey[IO, String].unsafeRunSync()
 
-  def apply[G[_], F[_]](http: Http[G, F])(implicit G: Sync[G]): Http[G, F] =
+  def apply[G[_], F[_]](
+      http: G[Response[F]])(implicit G: Sync[G], L: Local[G, Request[F]]): G[Response[F]] =
```
rossabaker, line 42: More abstract, less clear?  Certainly the former?

server/src/main/scala/org/http4s/server/middleware/RequestId.scala
```diff
@@ -38,38 +38,37 @@ object RequestId {
...
-    }
+  )(http: G[Response[F]])(implicit G: Sync[G], L: Local[G, Request[F]]): G[Response[F]] =
+    for {
+      req <- L.ask
+      header <- req.headers.get(headerName).map(_.head) match {
+        case None => G.delay(Header.Raw(headerName, UUID.randomUUID().toString()))
+        case Some(header) => G.pure(header)
+      }
+      reqId = header.value
+      resp <- L.local(http)(_.withAttribute(requestIdAttrKey, reqId).putHeaders(header))
```
rossabaker, line 55: Effectively asking the request a second time, because we effectfully transformed the request.  I haven't counted the flatMaps, but this feels flatMappier.

server/src/main/scala/org/http4s/server/middleware/RequestId.scala
```diff
@@ -100,6 +99,6 @@ object RequestId {
         headerName: CIString = requestIdHeader,
         genReqId: F[UUID]
     )(httpRoutes: HttpRoutes[F]): HttpRoutes[F] =
-      RequestId.apply(OptionT.liftK[F], headerName, genReqId)(httpRoutes)
+      RequestId.apply(OptionT.liftK.andThen(Kleisli.liftK[OptionT[F, *], Request[F]]), headerName, genReqId)(httpRoutes)
```
rossabaker, line 102: Inference is still garbage without this one.  The ones without the `~>` seem to do alright.

We wouldn't need it if people would just bring a `G[UUID]`.  With a bespoke `HttpRoutes` with a `Sync` instance, that's easy.  With an explicit stack, it looks like, uh, this?

server/src/test/scala/org/http4s/server/middleware/DateSuite.scala
```diff
@@ -31,7 +30,7 @@ class DateSuite extends Http4sSuite {
   }
 
   // Hack for https://github.com/typelevel/cats-effect/pull/682
-  val testService = Date(service)(Sync[OptionT[IO, *]], Clock.deriveOptionT[IO])
+  val testService = Date(service)
```
rossabaker, line 33: This is certainly neat.

PR #4755, opened 2021-04-17
Use a unique io runtime in BlazeServerSuite

### LINKED ISSUE #4166: Give blaze-server its own executor in tests
With regard to https://github.com/http4s/http4s/pull/4161/files/7c30f316ccc0e696d7b2b60e1475465859452f3b#r554253034, we shouldn't be testing against default thread names, but against a dedicated thread pool with obviously different names, so what we're testing for is explicit and not subject to the names of shared resources.  It will express intent better and be more robust to refactoring.

vasilmkd:
Resolves #4166.

### DIFF
--- a/blaze-server/src/test/scala/org/http4s/server/blaze/BlazeServerSuite.scala
+++ b/blaze-server/src/test/scala/org/http4s/server/blaze/BlazeServerSuite.scala
@@ -20,17 +20,47 @@ package blaze
 
 import cats.syntax.all._
 import cats.effect._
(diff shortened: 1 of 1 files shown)


rossabaker (approved):
I was imagining a fixture that would close the executors when done, but I think these are configured to wither to zero threads?

vasilmkd:
I can work on this some more tomorrow.

vasilmkd:
I used munit's `def afterAll(): Unit` to shut down the IO runtime. Is that satisfactory?

rossabaker (approved):
That's good enough for me.

PR #4756, opened 2021-04-17
Help the Scala compiler with default parameter in Logger.colored

### LINKED ISSUE #4647: Client `Logger.colored` can't be called with the default responseColor
When you try this:

```scala
val aa = Logger.colored[IO](
  logHeaders = true,
  logBody = false,
  logAction = Some(s => IO.println(s))
) _
```

The result is:

```
type mismatch;
 found   : org.http4s.Response[Nothing] => String
 required: org.http4s.Response[cats.effect.IO] => String
Error occurred in an application involving default arguments.
```

Providing the default `responseColor: Response[F] => String = ResponseLogger.defaultResponseColor _` makes it compile:

```scala
val aa = Logger.colored[IO](
  logHeaders = true,
  logBody = false,
  responseColor = ResponseLogger.defaultResponseColor _,
  logAction = Some(s => IO.println(s))
) _
```

Scala 2.13.5, http4s 1.0.0-M19

vasilmkd:
Resolves #4647.

### DIFF
--- a/client/src/main/scala/org/http4s/client/middleware/Logger.scala
+++ b/client/src/main/scala/org/http4s/client/middleware/Logger.scala
@@ -62,7 +62,7 @@ object Logger {
       logBody: Boolean,
       redactHeadersWhen: CIString => Boolean = Headers.SensitiveHeaders.contains,
       requestColor: String = RequestLogger.defaultRequestColor,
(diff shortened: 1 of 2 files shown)


rossabaker (approved):
That was easy.  And I think binary compatible.  I'll try a pick.

rossabaker:
Picked to 0.22.  I guess it was introduced after 0.21.

PR #4757, opened 2021-04-17
Response is not a case class anymore

### LINKED ISSUE #2537: Request (any Message actually) is not `equals`ed structurally
`Request` (and any `Message`) are not equal, even if they contain the same elements (empty `Vault`, for example)
the reason is that they contain field `attributes` of class `Vault`, which is not compared structurally

blocked on https://github.com/ChristopherDavenport/vault/pull/51 being merged and new version being published

vasilmkd:
Resolves #2537.

### DIFF
--- a/core/src/main/scala/org/http4s/Message.scala
+++ b/core/src/main/scala/org/http4s/Message.scala
@@ -651,19 +651,9 @@ object Response {
     new Response(status, httpVersion, headers, body, attributes)
 
   def unapply[F[_]](
-      message: Message[F]): Option[(Status, HttpVersion, Headers, EntityBody[F], Vault)] =
-    message match {
-      case response: Response[F] =>
-        Some(
-          (
-            response.status,
-            response.httpVersion,
-            response.headers,
-            response.body,
-            response.attributes))
-      case _ =>
-        None
-    }
+      response: Response[F]): Option[(Status, HttpVersion, Headers, EntityBody[F], Vault)] =
+    Some(
+      (response.status, response.httpVersion, response.headers, response.body, response.attributes))
 
   private[this] val pureNotFound: Response[Pure] =
     Response(
--- a/core/src/main/scala/org/http4s/Message.scala
+++ b/core/src/main/scala/org/http4s/Message.scala
@@ -522,12 +522,12 @@ object Request {
   *                   parameters which may be used by the http4s backend for
   *                   additional processing such as java.io.File object
   */
-final class Response[F[_]](
-    val status: Status = Status.Ok,
-    val httpVersion: HttpVersion = HttpVersion.`HTTP/1.1`,
-    val headers: Headers = Headers.empty,
-    val body: EntityBody[F] = EmptyBody,
-    val attributes: Vault = Vault.empty)
+final class Response[F[_]] private (
+    val status: Status,
+    val httpVersion: HttpVersion,
+    val headers: Headers,
+    val body: EntityBody[F],
+    val attributes: Vault)
     extends Message[F] {
   type SelfF[F0[_]] = Response[F0]
 
--- a/core/src/main/scala/org/http4s/Message.scala
+++ b/core/src/main/scala/org/http4s/Message.scala
@@ -528,7 +528,9 @@ final class Response[F[_]] private (
     val headers: Headers,
     val body: EntityBody[F],
     val attributes: Vault)
-    extends Message[F] {
+    extends Message[F]
+    with Product
+    with Serializable {
   type SelfF[F0[_]] = Response[F0]
 
   def copy(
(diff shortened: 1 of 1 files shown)


core/src/main/scala/org/http4s/Message.scala
```diff
@@ -522,15 +522,30 @@ object Request {
   *                   parameters which may be used by the http4s backend for
   *                   additional processing such as java.io.File object
   */
-final case class Response[F[_]](
-    status: Status = Status.Ok,
-    httpVersion: HttpVersion = HttpVersion.`HTTP/1.1`,
-    headers: Headers = Headers.empty,
-    body: EntityBody[F] = EmptyBody,
-    attributes: Vault = Vault.empty)
+final class Response[F[_]](
```
rossabaker, line 525: Is there any reason for the constructor to remain public?  We could also get rid of the redundant defaults then.

rossabaker (commented):
I'm surprised that this didn't ripple to a few people unapplying it.

This is good: we can add to it later (highly unlikely!) without breaking bincompat, and removes false expectations of equality when there is a body.  (Funny enough, with the strict entites @ChristopherDavenport is working on, we could have a reasonable `Eq` for many requests.  But still not all.)

core/src/main/scala/org/http4s/Message.scala
vasilmkd: I left it like that because I followed `Request`. Should I change them both?

core/src/main/scala/org/http4s/Message.scala
```diff
@@ -586,6 +601,17 @@ final case class Response[F[_]](
...
 object Response {
+  def apply[F[_]](
+      status: Status = Status.Ok,
+      httpVersion: HttpVersion = HttpVersion.`HTTP/1.1`,
+      headers: Headers = Headers.empty,
+      body: EntityBody[F] = EmptyBody,
+      attributes: Vault = Vault.empty): Response[F] =
+    new Response(status, httpVersion, headers, body, attributes)
+
+  def unapply[F[_]](r: Response[F]): Option[(Status, HttpVersion, Headers, EntityBody[F], Vault)] =
```
vasilmkd, line 612: Should this also take a `Message[F]` like `Request`?

vasilmkd:
I made `Response` almost identical to `Request` now in terms of "features". I have a question, why does `Request` override `equals` to an essentially structure based equality, and should `Response` do that too?

core/src/main/scala/org/http4s/Message.scala
rossabaker: I think we should.

core/src/main/scala/org/http4s/Message.scala
rossabaker: No, taking a `Response` makes sense to me.  The one on `Request` looks weird.

rossabaker:
This seems fit for 0.22.  I'll cherry-pick it when it's done, unless you want to rebase it.

core/src/main/scala/org/http4s/Message.scala
vasilmkd: I will change both.

vasilmkd:
I marked as draft until I push out the new changes. Can we get a resolution on `equals`?

rossabaker:
I missed the `equals` comment.  I think getting rid of `equals` is the motivation for why these aren't case classes: we don't have sensible equality on bodies, and we don't have sensible equality without considering bodies, so we don't have sensible equality.

vasilmkd:
I completely understand that, but `Request` at the moment _overrides_ equals like so:
```scala
  override def equals(that: Any): Boolean =
    (this eq that.asInstanceOf[Object]) || (that match {
      case that: Request[_] =>
        (this.method == that.method) &&
          (this.uri == that.uri) &&
          (this.httpVersion == that.httpVersion) &&
          (this.headers == that.headers) &&
          (this.body == that.body) &&
          (this.attributes == that.attributes)
      case _ => false
    })
```

I'm thinking this should be removed.

rossabaker:
Yes, I am agreeing with the removal.

rossabaker:
Oh, that `canEqual` can probably go, too.

vasilmkd:
`canEqual` comes from `Product`.

vasilmkd:
Should we try rebasing?

rossabaker:
Yes.  This is mostly pure code, and shouldn't hurt too badly on merge to main.

vasilmkd:
That does not work, I need to cherry pick the commits.

rossabaker:
If that cherry-picking is a pain, this would not be my first.  Let me know if you want help.

vasilmkd:
It's good for my git-fu. 😆

vasilmkd:
That was an awesome PR. Thank you @rossabaker for the support.

rossabaker (approved):
Great work!

Issue #2735, opened 2019-07-21
Reducing use of F.unsafeRunSync


Dennis4b:
(See also my last comment in #2695)

When running with a very low number of threads in order to provoke deadlocks that could manifest under load in production I run into a deadlock in `websocket/Http4sWSStage.scala`:

```
138   override protected def stageShutdown(): Unit = {                                                                         
139     F.toIO(deadSignal.set(true)).unsafeRunSync()                                                                           
140     super.stageShutdown()                                                                                                  
141   } 
```

where the `unsafeRunSync` is called, via `stageShutdown` all the way back to `sendClose` from within an effect, and can not synchronously run `deadSignal.set` as there are no more threads available:

```

"pool-1-thread-1" #15 prio=5 os_prio=0 tid=0x00007f09e6d75800 nid=0x73b0 waiting on condition [0x00007f0943ffe000]
   java.lang.Thread.State: WAITING (parking)
        at sun.misc.Unsafe.park(Native Method)
        - parking to wait for  <0x00000000fa9d0920> (a cats.effect.internals.IOPlatform$OneShotLatch)
        at java.util.concurrent.locks.LockSupport.park(LockSupport.java:175)
        at java.util.concurrent.locks.AbstractQueuedSynchronizer.parkAndCheckInterrupt(AbstractQueuedSynchronizer.java:836)
        at java.util.concurrent.locks.AbstractQueuedSynchronizer.doAcquireSharedInterruptibly(AbstractQueuedSynchronizer.java:997)
        at java.util.concurrent.locks.AbstractQueuedSynchronizer.acquireSharedInterruptibly(AbstractQueuedSynchronizer.java:1304)
        at cats.effect.internals.IOPlatform$.$anonfun$unsafeResync$2(IOPlatform.scala:50)
        at cats.effect.internals.IOPlatform$$$Lambda$2210/25411324.apply$mcV$sp(Unknown Source)
        at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:23)
        at scala.concurrent.BlockContext$DefaultBlockContext$.blockOn(BlockContext.scala:57)
        at scala.concurrent.package$.blocking(package.scala:146)
        at cats.effect.internals.IOPlatform$.unsafeResync(IOPlatform.scala:50)
        at cats.effect.IO.unsafeRunTimed(IO.scala:325)
        at cats.effect.IO.unsafeRunSync(IO.scala:240)
[...]

PR #4759, opened 2021-04-18
Remove obsolete comment about deadlock

rossabaker:
Warning is from #2735, but the underlying issue is fixed.


vasilmkd (approved):

rossabaker:
I must have forked this while main was broken.  It's just a comment?

Issue #2354, opened 2019-01-11
NonBlockingServletIo request body truncated to one chunk


dwalend:
NonBlockingServletIo request bodies are truncated to just one chunk. (With Tomcat 8.0.39, JDK 1.8.0_121-b13 , scala 2.11.12 , and http4s 0.20 M4.)

Work-around is to use BlockingServletIo, which does not have this problem.

This could be related to #1724 or #541 .

The server is 

```
package net.shrine.testapp

import cats.effect.{ConcurrentEffect, ContextShift, IO}
import javax.servlet.annotation.WebListener
import javax.servlet.{ServletContext, ServletContextEvent, ServletContextListener, ServletRegistration}
import org.http4s.server.DefaultServiceErrorHandler
import org.http4s.servlet.{AsyncHttp4sServlet, BlockingServletIo, NonBlockingServletIo}

import scala.concurrent.duration.Duration
import org.http4s.HttpRoutes
import org.http4s.dsl.impl.{->, /}
import org.http4s.dsl.io.{GET, Ok, PUT, Root, http4sOkSyntax}

import scala.language.higherKinds
import scala.concurrent.ExecutionContext.Implicits.global

/**
  * A simple servlet to bracket problems in Shrine without bringing in all of Shrine's complexity
  */

@WebListener
class TestAppBootstrap extends ServletContextListener {

  override def contextInitialized(sce: ServletContextEvent): Unit = {
    val context = sce.getServletContext

    implicit val contextShift: ContextShift[IO] = IO.contextShift(global)
    implicit val concurrentEffect: ConcurrentEffect[IO] = IO.ioConcurrentEffect(contextShift)

    mountService(
      context = context,
      name = "test http4s",
      service = service,
      mapping = "/*"
    )
  }

  def mountService[F[_]: ConcurrentEffect: ContextShift](
                    context: ServletContext,
                    name: String,
                    service: HttpRoutes[F],
                    mapping: String): ServletRegistration.Dynamic = {

    val servlet = new AsyncHttp4sServlet(
      service = service,
      asyncTimeout = Duration("30 seconds"),
      servletIo = NonBlockingServletIo(4096), //limited to 4096 bytes per request
//      servletIo = BlockingServletIo(4096,global), //works
      serviceErrorHandler = DefaultServiceErrorHandler[F]
    )
    val reg = context.addServlet(name, servlet)
    reg.setLoadOnStartup(1)
    reg.setAsyncSupported(true)
[...]

PR #4762, opened 2021-04-19
Add AsyncHttp4sServlet test suite

vasilmkd:
I cannot seem to reproduce #2354.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -365,11 +365,11 @@ lazy val servlet = libraryProject("servlet")
     libraryDependencies ++= Seq(
       javaxServletApi % Provided,
       jettyServer % Test,
-      jettyServlet % Test
+      jettyServlet % Test,
+      Http4sPlugin.asyncHttpClient % Test
     ),
   )
   .dependsOn(server % "compile;test->test")
-  .dependsOn(asyncHttpClient % Test)
 
 lazy val jetty = libraryProject("jetty")
   .settings(
--- a/build.sbt
+++ b/build.sbt
@@ -365,10 +365,11 @@ lazy val servlet = libraryProject("servlet")
     libraryDependencies ++= Seq(
       javaxServletApi % Provided,
       jettyServer % Test,
-      jettyServlet % Test,
+      jettyServlet % Test
     ),
   )
   .dependsOn(server % "compile;test->test")
+  .dependsOn(asyncHttpClient % Test)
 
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -365,10 +365,11 @@ lazy val servlet = libraryProject("servlet")
     libraryDependencies ++= Seq(
       javaxServletApi % Provided,
       jettyServer % Test,
-      jettyServlet % Test,
+      jettyServlet % Test
     ),
   )
   .dependsOn(server % "compile;test->test")
+  .dependsOn(asyncHttpClient % Test)
```
rossabaker, line 372: This dependency is going to be a pain as we test across repos after the Great Schism.  We get close to circular dependencies when we test our servers with our clients.  How hard would it be to do this with raw AHC?

servlet/src/test/scala/org/http4s/servlet/AsyncHttp4sServletSuite.scala
```diff
@@ -0,0 +1,111 @@
...
+  val servletServer = ResourceFixture[Int](serverPortR)
+
+  def get(serverPort: Int, path: String): IO[String] =
+    testBlocker.delay[IO, String](
+      Source
+        .fromURL(new URL(s"http://127.0.0.1:$serverPort/$path"))
+        .getLines()
+        .mkString)
+
+  servletServer.test("Http4sBlockingServlet handle GET requests") { server =>
```
rossabaker, line 58: The name seems off here.  Isn't it the servlet with NonBlockingIO?

rossabaker (changes requested):

build.sbt
vasilmkd: I will do it.

servlet/src/test/scala/org/http4s/servlet/AsyncHttp4sServletSuite.scala
vasilmkd: Sorry, yeah, I guess I forgot to adjust some things when I originally copied the file contents.

vasilmkd: Done.

build.sbt
vasilmkd: Done.

rossabaker (approved):

PR #4765, opened 2021-04-20
WebSocket on Ember Server

RaasAhsan:
Introduce initial support for the WebSocket API on Ember Server. The main goal here is to comply with the existing WebSocket model; refinements are going to come through in followups. The work here is pretty straightforward to the point that I have an appetite to implement a WebSocket client as well. I think that is uncharted territory though.

Something I've been thinking about is whether we want to release this immediately or wait (maybe we release on 1.x, maybe we wait for some followups to complete). I'm not personally a huge fan of long-lived development branches, but we could merge and let the code sit on mainline without exposing a flag to enable it. FWIW, I think this PR will be fairly easy to merge forward to `main`.

TODO:
- [ ] Think about how to expose websocket support
- [ ] Cover all error scenarios and ensure proper resource cleanup
- [ ] Address all the TODOs scattered around
- [ ] Rely on a custom WebSocket parser rather than the existing `FrameTranscoder`?
- [ ] Tests and benchmarks


core/src/main/scala/org/http4s/headers/Sec-WebSocket-Version.scala
```diff
@@ -17,4 +17,33 @@
 package org.http4s
 package headers
 
-object `Sec-WebSocket-Version` extends HeaderKey.Default
+import org.http4s.parser.HttpHeaderParser
+import org.http4s.util.{Renderer, Writer}
+import scala.concurrent.duration.{FiniteDuration, _}
+import scala.util.Try
+
+object `Sec-WebSocket-Version`
```
RaasAhsan, line 25: am i implementing all these headers w/ the new model correctly?

core/src/main/scala/org/http4s/parser/HttpHeaderParser.scala
```diff
@@ -139,9 +139,11 @@ object HttpHeaderParser
     addParser_("RANGE".ci, `RANGE`)
     addParser_("REFERER".ci, `REFERER`)
     addParser_("RETRY-AFTER".ci, `RETRY_AFTER`)
+    addParser_("SEC-WEBSOCKET-VERSION".ci, `SEC_WEBSOCKET_VERSION`)
```
RaasAhsan, line 142: I missed the accept header, and I think I also need to add these to a warmup phase

ember-server/src/main/scala/org/http4s/ember/server/internal/WebSocketHelpers.scala
```diff
@@ -0,0 +1,213 @@
...
+
+import scala.concurrent.duration.Duration
+import java.security.MessageDigest
+import java.util.Base64
+import java.nio.charset.StandardCharsets
+import org.http4s.headers.Connection
+import org.http4s.websocket.WebSocketFrame
+import java.nio.ByteBuffer
+
+object WebSocketHelpers {
```
RaasAhsan, line 39: I really want to make this a `WebSocketHandler` class that's parameterized over `F`. I think there's a lot more opportunity to preallocate values, and it's also easier to read since there are less arguments being passed around functions

RaasAhsan: We could do something similar with `ServerHelpers` as well -- a `ServerHandler` that's allocated once per `F`

ember-server/src/main/scala/org/http4s/ember/server/internal/WebSocketHelpers.scala
```diff
@@ -0,0 +1,213 @@
...
+        case None =>
+          // TODO: figure out what to do here
+          println("done")
+          Pull.done
+      }
+
+    go(stream, Array.emptyByteArray).void.stream
+  }
+
+  private def clientHandshake[F[_]](req: Request[F]): Either[ClientHandshakeError, String] = {
```
RaasAhsan, line 159: It made more sense to re-implement this more cleanly rather than rely on the existing handshaker. We can push it to the other package perhaps? I feel like other backends with the exception of Blaze wouldn't really use it

core/src/main/scala/org/http4s/headers/Sec-WebSocket-Version.scala
hamnis: The new model is in 0.22 and up, so this looks fine for 0.21.
I do think we should eliminate backticks for headers, but that is a different discussion.

ember-server/src/main/scala/org/http4s/ember/server/internal/WebSocketHelpers.scala
```diff
@@ -0,0 +1,217 @@
...
+import java.security.MessageDigest
+import java.util.Base64
+import java.nio.charset.StandardCharsets
+import org.http4s.headers.Connection
+import org.http4s.websocket.WebSocketFrame
+import java.nio.ByteBuffer
+
+object WebSocketHelpers {
+
+  private[this] val rfc6455Magic =
```
hamnis, line 41: add a link to the rfc?

ember-server/src/main/scala/org/http4s/ember/server/internal/ServerHelpers.scala
```diff
@@ -40,11 +40,10 @@ import scodec.bits.ByteVector
...
 
-  private val closeCi = "close".ci
-
-  private val connectionCi = "connection".ci
-  private val close = Connection(NonEmptyList.of(closeCi))
-  private val keepAlive = Connection(NonEmptyList.one("keep-alive".ci))
+  private[this] val closeCi = "close".ci
+  private[this] val connectionCi = "connection".ci
+  private[this] val close = Connection(NonEmptyList.of(closeCi))
+  private[this] val keepAlive = Connection(NonEmptyList.one("keep-alive".ci))
```
rossabaker, line 46: I got shouted down on this years ago, but I'm still of the opinion these header constants belong on the header companions.

ember-server/src/main/scala/org/http4s/ember/server/internal/ServerHelpers.scala
```diff
@@ -227,13 +225,37 @@ private[server] object ServerHelpers {
...
+            // TODO: fix the order of error handling?
             result.attempt.flatMap {
               case Right((req, resp, drain)) =>
-                send(socket)(Some(req), resp, idleTimeout, onWriteFailure) >>
-                  drain.map {
-                    case Some(nextBuffer) => Some(((req, resp), (nextBuffer, true)))
-                    case None => None
-                  }
+                // TODO: Should we pay this cost for every HTTP request?
+                // TODO: there will likely be many upgrade paths here eventually
```
rossabaker, line 232: Should we have a generic upgrade key instead, and then match on that?  It can't really be a sealed sum type, so, meh?

ember-server/src/main/scala/org/http4s/ember/server/internal/ServerHelpers.scala
```diff
@@ -227,13 +225,37 @@ private[server] object ServerHelpers {
...
 
+            // TODO: fix the order of error handling?
             result.attempt.flatMap {
               case Right((req, resp, drain)) =>
-                send(socket)(Some(req), resp, idleTimeout, onWriteFailure) >>
-                  drain.map {
-                    case Some(nextBuffer) => Some(((req, resp), (nextBuffer, true)))
-                    case None => None
-                  }
+                // TODO: Should we pay this cost for every HTTP request?
```
rossabaker, line 231: This is why blaze has an enableWebSockets flag.  But how expensive _are_ vault lookups?

ember-server/src/main/scala/org/http4s/ember/server/internal/WebSocketHelpers.scala
rossabaker: This could go somewhere in the server web socket package, I think.  It's not an Ember thing.

ember-server/src/main/scala/org/http4s/ember/server/internal/WebSocketHelpers.scala
```diff
@@ -0,0 +1,217 @@
...
+
+    reader
+      .concurrently(writer)
+      .drain
+      .compile
+      .drain
+      .attempt
+      .flatMap {
+        case Left(err) =>
+          err.printStackTrace()
```
rossabaker, line 122: We're missing some loggers here and on the println.

ember-server/src/main/scala/org/http4s/ember/server/internal/WebSocketHelpers.scala
rossabaker: I think servlets could.  The JSR web socket does not support our model, because we can't conditionally upgrade HTTP.  But the upgrade handler, now implemented in Jetty 10, gives us low-level control that might work just about like this!

ember-server/src/main/scala/org/http4s/ember/server/internal/WebSocketHelpers.scala
```diff
@@ -0,0 +1,217 @@
...
+          println("done")
+          Pull.done
+      }
+
+    go(stream, Array.emptyByteArray).void.stream
+  }
+
+  private def clientHandshake[F[_]](req: Request[F]): Either[ClientHandshakeError, String] = {
+    val connection = req.headers.get(Connection) match {
+      case Some(header) if header.values.contains_(upgradeCi) => Right(())
```
rossabaker, line 165: Cats syntax has an `Either.unit` for this, if you want to catch them.

ember-server/src/main/scala/org/http4s/ember/server/internal/WebSocketHelpers.scala
```diff
@@ -0,0 +1,217 @@
...
+      case Some(header) => Left(UnsupportedVersion(supportedWebSocketVersion, header.version))
+      case None => Left(VersionNotFound)
+    }
+
+    val key = req.headers.get(`Sec-WebSocket-Key`) match {
+      case Some(header) => Right(header.value)
+      case None => Left(KeyNotFound)
+    }
+
+    (connection, upgrade, version, key).mapN { case (_, _, _, key) => key }
```
rossabaker, line 185: Should this be a `Validated` so we can list all our complaints?

rossabaker (commented):
This is neat.

ember-server/src/main/scala/org/http4s/ember/server/internal/WebSocketHelpers.scala
RaasAhsan: I think these will just go away at the end. I haven't made a habit of adding debug log statements; is that something we want to do more of?

ember-server/src/main/scala/org/http4s/ember/server/internal/WebSocketHelpers.scala
RaasAhsan: I don't think the errors are mutually independent. For example, does it make sense to check the version or key headers if the connection and upgrade headers aren't right? The response code might also be different in the case of the former two checks (426 upgrade required)

RaasAhsan:
@rossabaker before I take the PR to the finish line, where do we think is the most appropriate target branch for this work? I'm happy to leave 0.21 behind at this point, so mostly thinking between 0.22 and main. The largest divergence is going to be the cats-parse and header changes.

I already said my piece in Gitter, but to summarize: if there's going to be a ton of feature work on 0.22, I'm highly inclined to push the work there. It also means we get to release this earlier rather than later (not sure what the timeline for main is looking like)

RaasAhsan:
Superseded by #4803

Issue #2627, opened 2019-06-07
Add a better implementation of a file-backed multipart decoder


nigredo-tori:
`MultipartDecoder.mixedMultipart` is package-private, and for a good reason: it requires the user to drain all the decoded `Part`s - and each can only be read once! This is a rather ugly restriction, which stems from the fact that `EntityDecoder` abstraction doesn't provide any resource handling.

Rather than shoehorn resource handling into an abstraction that doesn't support it, we should make it explicit. Something like this would do:
```scala
def proposedMultipart(...): Resource[F, EntityDecoder[F, Multipart[F]]]
// Usage
proposedMultipart(...).use { decoder =>
  req.decodeWith(decoder) { multipart =>
    ???
  }
}
```
This gives the user a way to explicitly specifly the multipart lifetime, and to rely on the guarantees that `Resource` provides.

As for the actual resource handling, I imagine we can keep track of created files while parsing, and delete all of them when the `Resource` is released. Or we can just create a temporary directory on acquire, create files in it while parsing, and delete the directory on release. Both these solutions promise to simplify resource handling in the `MultipartParser` (the `cleanupFile` calls littering the file).

rossabaker:
Would anything ever happen between the `.use` and the `req.decodeWith(decoder)`?  Do we gain much over `Resource[F, Multipart[F]]`?

nigredo-tori:
> Would anything ever happen between the .use and the req.decodeWith(decoder)?

Probably not. We could add some sugar for that if we expect `Resource[F, EntityDecoder[F, A]]` to become a pattern. Though the same goes for the `EntityDecoder[F, Resource[F, A]]` approach.

> Do we gain much over Resource[F, Multipart[F]]?

1. The actual decoding happens where we expect it to (in `Request.decodeWith`), rather then when acquiring the resulting resource.
1. We avoid potential pitfalls of not running the resulting `Resource[F, Multipart[F]]`, running it too late, or running it twice somehow.
1. It works better with "bracketing" functions like `Request.decodeWith`, which might be useful once we begin `Resource`ifying the server (#2663).

rossabaker:
Reading this again, I like it.

fredshonorio:
It looks like this issue could be closed given that #4915 was merged.

nigredo-tori:
> It looks like this issue could be closed given that #4915 was merged.

Absolutely. It probably was not automatically closed because `series/0.23` was not the default branch when the PR was being merged.

PR #4773, opened 2021-04-22
Refactor MultipartParser

nigredo-tori:
Yak shaving for #2627. Introduces the `Event` ADT:
```scala
  private[this] sealed trait Event
  private[this] final case class PartStart(value: Headers) extends Event
  private[this] final case class PartChunk(value: Chunk[Byte]) extends Event
  private[this] final case object PartEnd extends Event
```

The actual parsing is factored out as a `Stream[F, Event]` (see `parseEvents`), decoupling it from handling part limits (`limitParts`) and building `Part`s (e.g. `partBodyFileStream`). This should allow us to examine and modify them in isolation.

One drawback of this encoding is that the assumed order of `Event`s is not encoded in types, so the logic concerned with building `Part`s has to handle some impossible situations. However, it's easy to show that `Event`s only come in the specified order - this is apparent from the `pullPartEvents` implementation.

### DIFF
--- a/core/src/main/scala/org/http4s/multipart/MultipartParser.scala
+++ b/core/src/main/scala/org/http4s/multipart/MultipartParser.scala
@@ -36,6 +36,7 @@ object MultipartParser {
     boundary.value.getBytes("UTF-8")
   val StartLineBytesN: Boundary => Array[Byte] = BoundaryBytesN.andThen(DashDashBytesN ++ _)
 
+  /** `delimiter` in RFC 2046 */
   private[this] val ExpectedBytesN: Boundary => Array[Byte] =
     BoundaryBytesN.andThen(CRLFBytesN ++ DashDashBytesN ++ _)
   private[this] val dashByte: Byte = '-'.toByte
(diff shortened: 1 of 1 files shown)


core/src/main/scala/org/http4s/multipart/MultipartParser.scala
```diff
@@ -17,47 +17,67 @@
...
     BoundaryBytesN.andThen(CRLFBytesN ++ DashDashBytesN ++ _)
   private[this] val dashByte: Byte = '-'.toByte
   private[this] val streamEmpty = Stream.empty
-  private[this] val PullUnit = Pull.pure[Pure, Unit](())
 
   private type SplitStream[F[_]] = Pull[F, Nothing, (Stream[F, Byte], Stream[F, Byte])]
-  private type SplitFileStream[F[_]] =
-    Pull[F, Nothing, (Stream[F, Byte], Stream[F, Byte], Option[Path])]
+
+  private[this] sealed trait Event
```
rossabaker, line 45: Way back before it was called fs2, I wrote on that was a stream of `Either[Headers, ByteVector]`.  This is better than that, because Left and Right were arbitrary.  That one was public API, and I don't know how many people used it, but it was a nice foundation.

core/src/main/scala/org/http4s/multipart/MultipartParser.scala
```diff
@@ -17,47 +17,67 @@
...
+          case (None, PartStart(headers)) =>
+            (Some(Part(headers, Stream.empty)), None)
+          case (Some(acc0), PartChunk(chunk)) =>
+            (Some(acc0.copy(body = acc0.body ++ Stream.chunk(chunk))), None)
+          case (Some(_), PartEnd) =>
+            // Part done - emit it and start over.
+            (None, acc)
+          case _ =>
+            // Shouldn't happen if the `parseToEventsStream` contract holds.
+            sys.error("Unexpected state")
```
rossabaker, line 77: We have an internal.bug function that refers to our issue tracker for things like this.

rossabaker (commented):
I was postponing a review on this one until I had some time and quiet, because multipart is such a pain, but I think you made it much easier to follow.

Way back before it was called fs2, I wrote on that was a stream of `Either[Headers, ByteVector]`.  Yours is better than that, because Left and Right are arbitrary.  That one was public API, and I don't know how many people consumed it that way, but it gave great control over when to page to disk, avoided the nasty issues that arise when returning a stream of streams where one must be consumed/buffered before the next, etc.  Do you think that event stream might make a good low-level API?

core/src/main/scala/org/http4s/multipart/MultipartParser.scala
nigredo-tori: I've fleshed out these assertion errors a bit. They now go through the effect, and I've written out the actual situations to check exhaustivity. Not that it would help much - AFAICT, exhaustivity warnings don't fail the build here:
```
[warn] /home/nigredo/dev/http4s/core/src/main/scala/org/http4s/multipart/MultipartParser.scala:67:9: match may not be exhaustive.
[warn] It would fail on the following input: (Some(_), PartStart(_))
[warn]         (acc, item) match {
[warn]         ^
```

nigredo-tori:
> Way back before it was called fs2, I wrote on that was a stream of Either[Headers, ByteVector]. Yours is better than that, because Left and Right are arbitrary. That one was public API, and I don't know how many people consumed it that way, but it gave great control over when to page to disk, avoided the nasty issues that arise when returning a stream of streams where one must be consumed/buffered before the next, etc. Do you think that event stream might make a good low-level API?

It's obviously useful for implementing custom multipart caching logic. But I don't know how useful it would be compared to a properly documented "stream of streams" API. It is OK when the user needs a lot of flexibility, but I imagine that most of the time doing something like `Files[F].writeAll(path, body)` would be enough. Maybe we can expose both versions in the future, since they seem to cater to different use cases.

As an aside, I've toyed with the idea of building a `Stream[F, Part[F]]` over the stream of events, where each `Part`'s body would only be **valid** during the consuming `Stream.flatMap` iteration, but which wouldn't require the calling code to fully consume each `body`, and with proper backpressure. It should be possible with some careful parallelism. But it is definitely out of scope of the issue I've set out to fix, so it won't be in this PR, and the actual PR which would close #2627 would probably use the existing file caching logic (i.e. appending) rather than something new.

rossabaker (approved):
Sorry, this one came in about the time I fell underwater.  This looks like a good candidate for 0.23.0.

PR #4782, opened 2021-04-24
Default executionContext in blaze server (series/0.23)

kubukoz:


blaze-server/src/main/scala/org/http4s/server/blaze/BlazeServerBuilder.scala
```diff
@@ -324,22 +327,24 @@ class BlazeServerBuilder[F[_]] private (
...
-          LeafBuilder(
-            if (isHttp2Enabled) http2Stage(engine)
-            else http1Stage(secure = true, engine.some)
-          ).prepend(new SSLStage(engine))
-
-        case None =>
-          if (isHttp2Enabled)
-            logger.warn("HTTP/2 support requires TLS. Falling back to HTTP/1.")
-          LeafBuilder(http1Stage(secure = false, None))
+    dispatcher.unsafeToFuture {
```
kubukoz, line 330: Alternatively, we could construct the EC around the place where we sequence `Dispatcher` (`dispatcher <- Resource.eval(Dispatcher[F].allocated.map(_._1))`)

blaze-server/src/main/scala/org/http4s/server/blaze/BlazeServerBuilder.scala
```diff
@@ -207,7 +207,10 @@ class BlazeServerBuilder[F[_]] private (
     copy(socketAddress = socketAddress)
 
   def withExecutionContext(executionContext: ExecutionContext): BlazeServerBuilder[F] =
-    copy(executionContext = executionContext)
+    withExecutionContextF(executionContext.pure[F])
+
+  def withExecutionContextF(executionContextF: F[ExecutionContext]): BlazeServerBuilder[F] =
```
rossabaker, line 212: This seems weird to expose.  `flatMap` that shit.  It makes sense internally, since our default value comes from the constraint, but I don't know why anybody would _call_ it that way.

blaze-server/src/main/scala/org/http4s/server/blaze/BlazeServerBuilder.scala
```diff
@@ -91,7 +91,7 @@ import scodec.bits.ByteVector
   */
 class BlazeServerBuilder[F[_]] private (
     socketAddress: InetSocketAddress,
-    executionContext: ExecutionContext,
+    executionContextF: F[ExecutionContext],
```
rossabaker, line 94: If we don't want this to be an effect, we could have an internal sum type: `FromAsync` and `ExplicitExecutionContext`.  We did something similar on an `SSLContext`.  I think this model is probably fine, but I've thought about making the builder values `val`s so people can introspect, print, etc, and a sum type would be cleaner there.

It's also a bincompat refactor to get there from here, so I think this model is fine.

blaze-server/src/main/scala/org/http4s/server/blaze/BlazeServerBuilder.scala
```diff
@@ -419,9 +424,12 @@ class BlazeServerBuilder[F[_]] private (
 
 object BlazeServerBuilder {
   def apply[F[_]](executionContext: ExecutionContext)(implicit F: Async[F]): BlazeServerBuilder[F] =
+    apply[F].withExecutionContext(executionContext)
```
rossabaker, line 427: Once we have a reasonable default, there's no reason to make this overload special.  Maybe we should deprecate it.  Or better yet, scalafix it away.

rossabaker (commented):
I like the general direction here.  Some thoughts below.

blaze-server/src/main/scala/org/http4s/server/blaze/BlazeServerBuilder.scala
kubukoz: Yeah, I guess since the caller of `.resource` / `stream` is likely to be the same as the one who passes the EC, they can `flatMap` that shit themselves.

blaze-server/src/main/scala/org/http4s/server/blaze/BlazeServerBuilder.scala
kubukoz: I like this very much. Thanks!

blaze-server/src/main/scala/org/http4s/server/blaze/BlazeServerBuilder.scala
kubukoz: scalafixing it away seems reasonable, I only left it for compat right now. Will update in the next round :)

rossabaker:
I al doing like six things at once, but I also think you inherited a broken build from me doing like six things at once.

rossabaker:
I think this build is suffering from something fixed in #4781, and merging main will clean it up.

bplommer:
Is this still blocked? I think it will save a lot of bother.

rossabaker:
I _think_ it's still bincompat and appropriate for 0.23, but I delegate to MiMa.  The `withExecutionContextF` can still go, and it needs a merge conflict resolved.

@kubukoz, are you still up for getting this to completion, or should we bring in someone else?  It looks close.

rossabaker:
That `executionContextF` might require a bincompat dance, but I think it's doable.

kubukoz:
I'm up, should be doable tomorrow :)

kubukoz:
um... fix one conflict, get 17 more. Cool, cool...

I think I'll just cherry pick from scratch.

kubukoz:
The fields in `BlazeServerBuilder` are all private, so this should be compatible, but I think we agreed the client could get the same treatment and `executionContext` is already a `val` that can be accessed by callers - I don't really see a way to do it and keep it binary compatible :/ guess we'll only do the server for now, unless someone has a different idea for the client

rossabaker (approved):
Ooh, that's a bummer on the client.  I'll open a separate issue on that.

kubukoz:
3.0.1 on 1.8 has a Jetty failure, seems unrelated...

rossabaker:
I think those errors are scalafmt noise.  We probably should bite the bullet and do the scalafmt-3 upgrade.

PR #4783, opened 2021-04-24
Fix Resource Leak In Jetty

isomarcte:
We had a resource leak in Jetty due to us not calling `.destroy` in the `def resource` after the server terminated. Fixing this leak was a bit complicated.

By default, Jetty will terminate the provided `ThreadPool` as part of the resource cleanup in `.destroy`. Since we were sharing the `ThreadPool` in the builder, this would cause any new invocations of `def resource` to fail as the underlying `ThreadPool` will have now been terminated.

To fix this problem we now use a `Resource[F, ThreadPool]` which will ensure that we get a fresh `ThreadPool` on each invocation, as well as ensuring proper shutdown semantics. We have to keep around the old `ThreadPool` for binary compatibility. In order to avoid starting _two_ `ThreadPool` instances we introduce a `LazyThreadPool` which _should_ never actually create a `ThreadPool` (I had to leave some `ThreadPool` type here and I didn't want to use `null` or anything that would throw if it got accidentally accessed).

If someone explicitly sets the `ThreadPool` with the deprecated `withThreadPool` method, then we wrap it in `UndestroyableThreadPool` which hides the `join` method and thus preserves the old behavior allowing the `ThreadPool` to be shared across multiple invocations of `def resource` while still allowing us to cleanup all the other resources.

`JettyLifeCycle` provides utilities which lift the Jetty interfaces `LifeCycle` and `Destroyable` into a `Resource`. These are not public to keep the binary compatibility surface as small as possible. In order to allow users to customize the `ThreadPool` a subset of `JettyLifeCycle` is exposed in `JettyThreadPools`.

### DIFF
--- a/jetty/src/main/scala/org/http4s/server/jetty/JettyBuilder.scala
+++ b/jetty/src/main/scala/org/http4s/server/jetty/JettyBuilder.scala
@@ -65,6 +65,7 @@ sealed class JettyBuilder[F[_]] private (
 
   private[this] val logger = getLogger
 
+  @deprecated(message = "Retained for binary compatibility", since = "0.21.23")
   private[JettyBuilder] def this(
     socketAddress: InetSocketAddress,
     threadPool: ThreadPool,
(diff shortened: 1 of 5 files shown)


hamnis (approved):

hamnis:
needs a conflict resolving, otherwise good

isomarcte:
@hamnis conflicts are resolved. Thanks for the review!

PR #4786, opened 2021-04-25
Scala 3.0.0-RC3 for 0.22.x

rossabaker:
This won't work without cats-parse, but includes the first RC3-only dependency.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,7 +27,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        scala: [2.12.13, 2.13.5, 3.0.0-RC2]
+        scala: [2.12.13, 2.13.5, 3.0.0-RC3]
         java: [adopt@1.8, adopt@1.11, adopt@1.15]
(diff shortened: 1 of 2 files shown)


aeons:
`cats-parse` is here, but it looks like [blaze 0.15.0-M4](https://github.com/http4s/blaze/releases/tag/v0.15.0-M4) didn't make it to maven central.

aeons:
Whoops, looks like I pushed the non-working blaze bump as well.

rossabaker:
I've fallen into a hole.  Do we still want this PR?  There are a bunch of new bumps for 3.0.0 now.  I'm more focused on getting the last 0.21 out the door at the moment.

hamnis:
I would say no, we should focus on getting 3.0.0 final deps.

rossabaker:
Sorry, I wrote poorly.  Agree on forgetting about RC3.  Does this PR get us closer to 3.0.0 support, or should we start over?

I broke the series/0.22 branch in a merge last night because of a Cats version bump.  Maybe we should just start lining things up directly on that branch.

rossabaker:
Yeah, these versions are all obsolete, I think.

Issue #4706, opened 2021-04-04
`Header.Select` needs a method like `from` that preserves `ParseFailure`s


zarthross:
I'm working on updating [Rho](https://github.com/http4s/rho/pull/501) to support the new header redesign. One of the issues I've run into is Rho prefers to throw errors for headers that fail to parse, and the new header redesign appears to prefer simply ignoring the errors?

- Can we get a method on `Header.Select` that doesn't convert the `Either` to `Option`?
- Should this same type of method exist on `Headers`?
- Should this method preserve every parse error or just the first in the case of `Recurring`? 
- Would we prefer something like:
   - `List[Header.Raw] => Either[ParseFailure, F[A]]`
   - `List[Header.Raw] => F[Either[ParseFailure, A]`
   - `List[Header.Raw] => EitherNel[ParseFailure, F[A]]`

Happy to submit a PR, just need advice and guidance.

rossabaker:
I think this might be a rare use case for `Ior`, specifically `Ior[NonEmptyList[ParseFailure], F[A]]`.  "Tell me everything you had to throw away, but give me a value if you can."

* On the happy path, we get an `Ior.Right`.
* If it's total garbage, we get an `Ior.Left`.
* If it's a recurring header and m of n are bad, for 0 <  m < n, we'll get an `Ior.Both` with the good values and warnings of what we choked on, and the caller can decide what to do about it.

/cc @SystemFw.  This was all his brainchild.

PR #4787, opened 2021-04-25
Adding Header.Select fromSafe, toRaw multiple, and Aux

zarthross:
Some enhancements needed to help enable Rho support. 

See: https://github.com/http4s/rho/pull/501 for some usage example.

I could use feedback on these, once I get buy in I'll happily add/fixup the unit tests!

Resolves https://github.com/http4s/http4s/issues/4706

### DIFF
--- a/core/src/main/scala/org/http4s/Header.scala
+++ b/core/src/main/scala/org/http4s/Header.scala
@@ -5,6 +5,7 @@ import cats.data.NonEmptyList
 import cats.syntax.all._
 import org.typelevel.ci.CIString
 import org.http4s.util.{Renderer, Writer}
+import cats.data.Ior
 
 /** Typeclass representing an HTTP header, which all the http4s
   * default headers satisfy.
@@ -146,11 +147,16 @@ object Header {
 
     /** Transform this header into a [[Header.Raw]]
       */
-    def toRaw(a: A): Header.Raw
+    def toRawOne(a: A): Header.Raw
+
+    /** Transform this (potentially repeating) header into a [[Header.Raw]] */
+    def toRaw(a: F[A]): Header.Raw
 
     /** Selects this header from a list of [[Header.Raw]]
       */
     def from(headers: List[Header.Raw]): Option[F[A]]
+
+    def fromSafe(headers: List[Header.Raw]): Option[Ior[NonEmptyList[ParseFailure], F[A]]]
   }
   trait LowPrio {
     implicit def recurringHeadersNoMerge[A](implicit
@@ -158,27 +164,49 @@ object Header {
       new Select[A] {
         type F[B] = NonEmptyList[B]
 
-        def toRaw(a: A): Header.Raw =
+        def toRawOne(a: A): Header.Raw =
           Header.Raw(h.name, h.value(a))
+
+        def toRaw(a: F[A]): Header.Raw =
+          Header.Raw(h.name, a.map(h.value).mkString_(", "))
 
         def from(headers: List[Header.Raw]): Option[NonEmptyList[A]] =
           headers.collect(Function.unlift(Select.fromRaw(_))).toNel
+
+        def fromSafe(headers: List[Raw]): Option[Ior[NonEmptyList[ParseFailure], NonEmptyList[A]]] =
+          headers.foldLeft(Option.empty[Ior[NonEmptyList[ParseFailure], NonEmptyList[A]]]) {
+            (a, raw) =>
+              Select.fromRawSafe(raw) match {
+                case Some(aa) => a |+| aa.bimap(NonEmptyList.one, NonEmptyList.one).some
+                case None => a
+              }
+          }
       }
   }
   object Select extends LowPrio {
+    type Aux[A, G[_]] = Select[A] { type F[B] = G[B] }
+
     def fromRaw[A](h: Header.Raw)(implicit ev: Header[A, _]): Option[A] =
       (h.name == Header[A].name).guard[Option] >> Header[A].parse(h.value).toOption
+
+    def fromRawSafe[A](h: Header.Raw)(implicit ev: Header[A, _]): Option[Ior[ParseFailure, A]] =
+      (h.name == Header[A].name).guard[Option].as(Header[A].parse(h.value).toIor)
 
     implicit def singleHeaders[A](implicit
         h: Header[A, Header.Single]): Select[A] { type F[B] = B } =
       new Select[A] {
         type F[B] = B
 
-        def toRaw(a: A): Header.Raw =
+        def toRawOne(a: A): Header.Raw =
           Header.Raw(h.name, h.value(a))
+
+        def toRaw(a: A): Header.Raw = toRawOne(a)
 
         def from(headers: List[Header.Raw]): Option[A] =
           headers.collectFirst(Function.unlift(fromRaw(_)))
+
+        def fromSafe(headers: List[Raw]): Option[Ior[NonEmptyList[ParseFailure], F[A]]] =
+          headers.collectFirst(Function.unlift(fromRawSafe(_).map(_.leftMap(NonEmptyList.one))))
       }
 
     implicit def recurringHeadersWithMerge[A: Semigroup](implicit
@@ -186,13 +214,23 @@ object Header {
       new Select[A] {
         type F[B] = B
 
-        def toRaw(a: A): Header.Raw =
+        def toRawOne(a: A): Header.Raw =
           Header.Raw(h.name, h.value(a))
+
+        def toRaw(a: F[A]): Header.Raw = toRawOne(a)
 
         def from(headers: List[Header.Raw]): Option[A] =
           headers.foldLeft(Option.empty[A]) { (a, raw) =>
             fromRaw(raw) match {
               case Some(aa) => a |+| aa.some
+              case None => a
+            }
+          }
+
+        def fromSafe(headers: List[Raw]): Option[Ior[NonEmptyList[ParseFailure], F[A]]] =
+          headers.foldLeft(Option.empty[Ior[NonEmptyList[ParseFailure], F[A]]]) { (a, raw) =>
+            fromRawSafe(raw) match {
+              case Some(aa) => a |+| aa.leftMap(NonEmptyList.one).some
               case None => a
             }
           }
--- a/core/src/main/scala/org/http4s/Header.scala
+++ b/core/src/main/scala/org/http4s/Header.scala
@@ -154,9 +154,7 @@ object Header {
 
     /** Selects this header from a list of [[Header.Raw]]
       */
-    def from(headers: List[Header.Raw]): Option[F[A]]
-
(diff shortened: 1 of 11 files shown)


rossabaker:
We should probably target 0.22 with this: that's where the new header model debuts.  It shouldn't be hard to merge to main, since it's pure work.

core/src/main/scala/org/http4s/Header.scala
```diff
@@ -141,56 +142,93 @@ object Header {
...
+    def toRawOne(a: A): Header.Raw
+
+    /** Transform this (potentially repeating) header into a [[Header.Raw]] */
+    def toRaw(a: F[A]): Header.Raw
 
     /** Selects this header from a list of [[Header.Raw]]
       */
     def from(headers: List[Header.Raw]): Option[F[A]]
+
+    def fromSafe(headers: List[Header.Raw]): Option[Ior[NonEmptyList[ParseFailure], F[A]]]
```
rossabaker, line 154: Trying to remember why we have an `Option[F]` on these, resulting in an `Option[NonEmptyList]` for `Recurring` instead of `F` being `Option` or `List`.

I don't love the name `fromSafe`, but I do think the `Ior` is a correct model: "give me all you can, and warn me about the ones you couldn't."

core/src/main/scala/org/http4s/Header.scala
```diff
@@ -141,56 +142,93 @@ object Header {
 
     /** Transform this header into a [[Header.Raw]]
       */
-    def toRaw(a: A): Header.Raw
+    def toRawOne(a: A): Header.Raw
+
+    /** Transform this (potentially repeating) header into a [[Header.Raw]] */
+    def toRaw(a: F[A]): Header.Raw
```
rossabaker, line 148: I'm a bit lost on this one.  Is this something rho needed?  Can you point me to it?

core/src/main/scala/org/http4s/Header.scala
zarthross: > Trying to remember why we have an Option[F] on these, resulting in an Option[NonEmptyList] for Recurring instead of F being Option or List.

I could look into changing that, it makes more sense to me.

> I don't love the name fromSafe, but I do think the Ior is a correct model: "give me all you can, and warn me about the ones you couldn't."

I think just having 1 `from` that returns the `IOr` would be fine, and our users can `toOption` if they want to ignore the parse failures.

core/src/main/scala/org/http4s/Header.scala
zarthross: https://github.com/http4s/rho/blob/c5a90fcebaee2524d5898969f38ec183b2eb8c86/core/src/main/scala/org/http4s/rho/RhoDslHeaderExtractors.scala#L146

I debated about just changing the existing `toRaw` rather than having a separate `toRawOne`, since most of the time you have a header with a semigroup, rather than a list of something, (or at least that seems to be the http4s pattern) where you might want to just do 1 of.

zarthross:
> We should probably target 0.22 with this: that's where the new header model debuts. It shouldn't be hard to merge to main, since it's pure work.

Done.

core/src/main/scala/org/http4s/Header.scala
zarthross: Have any suggestions on this @rossabaker ?  Would it be better to have just `toRaw` and not a `toRawOne`?

core/src/main/scala/org/http4s/Header.scala
zarthross: @rossabaker thoughts?

rossabaker:
Haven't forgotten this.  0.21 sucked up all my time, but this is absolutely necessary before 0.22 so we don't leave rho hanging.  Extra opinions welcome.

core/src/main/scala/org/http4s/Header.scala
aeons: > I think just having 1 from that returns the IOr would be fine, and our users can toOption if they want to ignore the parse failures.
I think this makes sense.

rossabaker: > I think just having 1 from that returns the IOr would be fine, and our users can toOption if they want to ignore the parse failures.

I agree with this.

rossabaker:
So this is still in draft.  We want to release 0.22.0, but are blocked by a couple issues and circe.  I also want to unblock Rho.  We could merge this as is and clean it up before the final release, or we could let rho work off snapshots/source dependencies and clean up before the final release.  Thoughts?

zarthross:
I think I can make some time this weekend for the suggestions here, and hopefully any associated testing.

rossabaker:
That sounds great.  I have a stupid busy weekend, so if we can get things greened up, I may do a milestone today.  But I will commit to do another release as soon as this is ready, because we don't want to leave Rho hanging.

core/src/main/scala/org/http4s/Header.scala
zarthross: What about `Headers.get`? Should it return an `Option` or should it also return an `Option[Ior...`?

core/src/main/scala/org/http4s/Header.scala
zarthross: @rossabaker ^?

zarthross:
Just 2 questions that need answering on the above issues, and I'd be ready to move this out of Draft.

core/src/main/scala/org/http4s/Header.scala
rossabaker: `toRawOne` is equivalent to `toRaw` for single headers, and for headers with a Semigroup instance?  Only recurring headers without a Semigroup need this new signature?  And that is in practice only `Set-Cookie`?

rossabaker: I don't think `toRaw(a: F[A]): Header.Raw` is sensible for `Set-Cookie`, because a `NonEmptyList` of `Set-Cookie` can't be reduced to a single header.

zarthross: What would be correct. I'm fine with dropping the `toRawOne`, but Rho will need the new version of `toRaw` that takes a `F[A]` instead of just an `A` in order for the types to work out.

core/src/main/scala/org/http4s/Header.scala
rossabaker: Hmm.  Good question.  `My inclination is `Option[Ior[...`, but that would be a massive breaking change.  The existing signature is consistent with its map-like semantics.

I argued against the overload earlier, and I don't think the typeclass needs the overload, but `Headers` is the primary user API.  Maybe leaving `get` as is, and `getWithWarnings` that returns the `Option[Ior[...`?

core/src/main/scala/org/http4s/Header.scala
rossabaker: I guess what I'm saying is that there's no correct implementation of that signature for `Set-Cookie`, because you can't just concatenate the values of the `NonEmptyList` with a comma.  The return type would need to be a collection for that header.  It's the exception to all the other rules.

zarthross: Sooo.... are you saying we keep both signatures?  Do we just need better names?

core/src/main/scala/org/http4s/Header.scala
zarthross: Done.

core/src/main/scala/org/http4s/Header.scala
rossabaker: I'm concerned about what the need for this signature is telling us, because the one that joins with a comma is an invalid implementation.  Perhaps I need to spend more time studying that Rho link.

rossabaker: `cats-effect-3` branch, right?  I might need to get hands on with this.  The types are more than I can contemplate on GitHub.

zarthross: Correct.  The important part for that piece of code is we have an `F[A]` where `toRaw` requires and `A`.   I can probably work around this, but it just seemed logical at the time that `toRaw` should be able to convert to and from raw in the `F[A]` form.

rossabaker: Yeah, it's just that one exceptional header that makes that not work.  I'm playing type tetris with a local checkout now.

rossabaker: Okay, https://github.com/http4s/rho/pull/538 compiles after my PR to this PR.  I don't know rho nearly as well as you, so I don't know whether it's acceptable, but maybe that's our way out of this?

zarthross: I think it'll be fine.  Thanks for looking into this!

core/src/main/scala/org/http4s/syntax/HeaderSyntax.scala
```diff
@@ -17,23 +17,33 @@
...
 
 final class HeaderOps[A](val a: A, header: Header[A, _]) {
   def value: String = header.value(a)
   def name: CIString = header.name
 }
 
-final class SelectOps[A](val a: A)(implicit ev: Header.Select[A]) {
-  def toRaw: Header.Raw = ev.toRaw(a)
+final class SelectOpsOne[A](val a: A)(implicit ev: Header.Select[A]) {
+  def toRaw: Header.Raw = ev.toRaw1(a)
```
rossabaker, line 42: This seemed shady to me, but would have changed a lot to rename it to `toRaw1` where it's used.  I'm also not comfortable with how it interacts with `SelectOpsMultiple`.  We might want to take a look at this, still.

rossabaker (dismissed):
I think this is sufficient, and we can do an RC with it when circe-0.14 is ready and make sure Rho has a path forward.

core/src/main/scala/org/http4s/syntax/HeaderSyntax.scala
zarthross: Yeah... there could be instances of ambiguous implicits...  maybe we can rename `SelectOpsMultiple.toRaw` to something else to avoid that?

rossabaker: I think the syntax needs to be consistent with the typeclass: https://github.com/zarthross/http4s/pull/2

rossabaker:
Looks like we also broke this:

```
 ==> X org.http4s.client.middleware.LoggerSuite.RequestLogger should not affect a Post  0.008s org.http4s.ParseFailure: Invalid Content-Type header: Content-Type header doesn't support media ranges
```

zarthross:
> Looks like we also broke this:
> 
> ```
>  ==> X org.http4s.client.middleware.LoggerSuite.RequestLogger should not affect a Post  0.008s org.http4s.ParseFailure: Invalid Content-Type header: Content-Type header doesn't support media ranges
> ```

I can take a look.

PR #4793, opened 2021-04-26
Make Use Of IPv4 Or IPv6 In Defaults Explicit

isomarcte:
The implementation for our default `Host` was the following,

```scala
val Host = InetAddress.getLoopbackAddress.getHostAddress
```

Whether the result of this invocation is the canonical IPv4 or IPv6 loopback address is the result of this code in the JRE.

```java
class InetAddressImplFactory {

    static InetAddressImpl create() {
        return InetAddress.loadImpl(isIPv6Supported() ?
                                    "Inet6AddressImpl" : "Inet4AddressImpl");
    }

    static native boolean isIPv6Supported();
}
```

Where `isIPv6Supported` is some native method.

On standard JREs, this is governed by a set of system properties including `java.net.preferIPv6Addresses`. For example,

```shell
> scala
Welcome to Scala 2.13.4-20201218-225751-unknown (OpenJDK 64-Bit Server VM, Java 16).

scala> java.net.InetAddress.getLoopbackAddress
java.net.InetAddress.getLoopbackAddress
val res0: java.net.InetAddress = localhost/127.0.0.1

scala> :q
:q
> scala -Djava.net.preferIPv6Addresses=true
Welcome to Scala 2.13.4-20201218-225751-unknown (OpenJDK 64-Bit Server VM, Java 16).

scala> java.net.InetAddress.getLoopbackAddress
java.net.InetAddress.getLoopbackAddress
val res0: java.net.InetAddress = localhost/0:0:0:0:0:0:0:1
```

And honestly if they are running on a more exotic JRE I've no idea which one they'd get by default. Most systems these days (e.g. Linux, Windows, Web Browsers) will prefer the IPv6 Address if the system has a ipv6 address and the target has an IPv6 address, so I wouldn't be surprised if some systems used similar heuristics to give IPv6 precedence (which is of course great, IPv6 is great!).

However, I expect the dynamic and quasi-defined behavior would be surprising to most users.

This commit deprecates the default `Host` and `SocketAddress` and adds new IPv4/IPv6 specific replacements so that the behavior is unambiguous. It changes the various http4s internal uses of `Host` to use the IPv4 based defaults as that is likely what they are already doing in 99% of cases and should be the least surprising on our users.

### DIFF
--- a/blaze-server/src/main/scala/org/http4s/server/blaze/BlazeServerBuilder.scala
+++ b/blaze-server/src/main/scala/org/http4s/server/blaze/BlazeServerBuilder.scala
@@ -467,7 +467,7 @@ object BlazeServerBuilder {
       F: ConcurrentEffect[F],
       timer: Timer[F]): BlazeServerBuilder[F] =
     new BlazeServerBuilder(
-      socketAddress = defaults.SocketAddress,
+      socketAddress = defaults.IPv4SocketAddress,
       executionContext = executionContext,
       responseHeaderTimeout = defaults.ResponseTimeout,
       idleTimeout = defaults.IdleTimeout,
--- a/ember-server/src/main/scala/org/http4s/ember/server/EmberServerBuilder.scala
+++ b/ember-server/src/main/scala/org/http4s/ember/server/EmberServerBuilder.scala
@@ -185,7 +185,7 @@ object EmberServerBuilder {
     )
 
   private object Defaults {
-    val host: String = server.defaults.Host
+    val host: String = server.defaults.IPv4Host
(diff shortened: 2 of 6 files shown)


isomarcte:
I'm not sure why that test is failing. I can't reproduce it locally.

rossabaker:
@hamnis Is this what GitHub changed and broke all our localhost references a while back?

hamnis:
@rossabaker that seem like a very plausible explanation.

hamnis (approved):

isomarcte:
@rossabaker Do you want me to retarget this to 0.22.x or leave it at 0.21.x?

PR #4795, opened 2021-04-28
Initial Encoding Of ServerBuildable

isomarcte:
This commit provides an initial encoding a of `ServerBuildable` a pseudo typeclass alternative to `ServerBuilder`.

The intent here is to provide a clean minimal ad-hoc polymorphic way to define an interface for things which can configure and run servers.

Prior to this, the code required to build a server varied a great deal between the different backends. `ServerBuilder` provides somewhat of a common interface, but lacks even some very basic items, e.g. there is no way to pass in an `HttpApp` to `ServerBuilder`.

This leads to confusion among users about exactly how to configure and start a specific server, and discourages trying different http backends as users have to learn the idioms of each underlying system again. To an extent, that is unavoidable. Certain aspects of configuration will _always_ be specific a given backend, however it need not be quite a bespoke as it is.

Currently `ServerBuildable` implements 1 fundamental operation for serving a server,

```scala
  /** Returns the backend as a resource.  Resource acquire waits
    * until the backend is ready to process requests.
    */
  def resource(a: A): Resource[F, Server]
```

And 3 fundamental operations for configuring a server builder,

```scala
  /** Set the http host for the server. */
  def withHttpHost(a: A)(host: String): A

  /** Set the http port for the server. */
  def withHttpPort(a: A)(port: Int): A

  /** Set the [[HttpApp]] to serve on this server. */
  def withHttpApp(a: A)(app: HttpApp[F]): A
```

I've some thoughts on how to make this encoding a bit more bin compat friendly as well, and I'm not sold on it at all, but I thought I'd make a draft PR now so people can provide feedback. 

Currently this is targeting 0.22.x, but given the scope of this change, _if_ it gets merged 1.x.x is probably more likely.

### DIFF
--- a/examples/blaze/src/main/scala/com/example/http4s/blaze/BlazeExample.scala
+++ b/examples/blaze/src/main/scala/com/example/http4s/blaze/BlazeExample.scala
@@ -23,6 +23,7 @@ import org.http4s.server.{Router, Server}
 import org.http4s.server.blaze.BlazeServerBuilder
 import org.http4s.syntax.kleisli._
 import scala.concurrent.ExecutionContext.global
+import org.http4s.server.syntax.serverBuildable._
 
 object BlazeExample extends IOApp {
   override def run(args: List[String]): IO[ExitCode] =
@@ -39,9 +40,7 @@ object BlazeExampleApp {
     for {
       blocker <- Blocker[F]
       app = httpApp[F](blocker)
-      server <- BlazeServerBuilder[F](global)
-        .bindHttp(8080)
-        .withHttpApp(app)
-        .resource
+      builder = BlazeServerBuilder[F](global)
+      server <- builder.withHttpPort(8080).withHttpApp(app).resource
     } yield server
 }
--- a/server/src/main/scala/org/http4s/server/ServerBuildable.scala
+++ b/server/src/main/scala/org/http4s/server/ServerBuildable.scala
@@ -1,0 +1,139 @@
+/*
+ * Copyright 2014 http4s.org
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.http4s
+package server
+
+import cats.effect.concurrent._
+import cats.effect._
+import cats.syntax.all._
+import fs2._
+import fs2.concurrent._
+
+/** A pseudo-typeclass for something which can build a server.
+  *
+  * There are primary three motivations for this class.
+  *
+  * - Provide an ad-hoc polymorphic way approach to configure a server, as
+  *   opposed to the [[ServerBuilder]] subtyping approach. This is generally
+  *   more consistent with the current design idioms in use in FP Scala code
+  *   and (subjectively) is cleaner, e.g. no dealing with `Self` typing, etc.
+  * - Provide a minimal definition for what you need to provide if you are
+  *   implementing a server backend. [[ServerBuilder]] already does this to a
+  *   degree, but due to the current design of that class, not all of our
+  *   servers implement that, e.g. Ember.
+  *
+  * This class does not and ''can not'' replace backend specific
+  * builders. Those must always exist as the precise feature set of a given
+  * backend will likely never be totally shared with any other backend.
+  */
+trait ServerBuildable[F[_], A] {
+
+  /** Returns the backend as a resource.  Resource acquire waits
+    * until the backend is ready to process requests.
+    */
+  def resource(a: A): Resource[F, Server]
+
+  // Abstract Settings
+
+  /** Set the http host for the server. */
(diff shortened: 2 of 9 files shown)


isomarcte:
I updated a couple example files, just as a sanity check

server/src/main/scala/org/http4s/server/ServerBuildable.scala
```diff
@@ -0,0 +1,141 @@
...
+    * release of the backend.  This function is intended for REPL
+    * sessions, tests, and other situations where composing a
+    * [[cats.effect.Resource]] or [[fs2.Stream]] is not tenable.
+    * [[resource]] or [[stream]] is recommended wherever possible.
+    */
+  final def allocated(a: A)(implicit F: Bracket[F, Throwable]): F[(Server, F[Unit])] = resource(
+    a).allocated
+
+  /** Bind the server to a given port and host. */
+  final def bindHttp(a: A)(port: Int = defaults.HttpPort, host: String = defaults.Host): A =
```
rossabaker, line 84: This seems redundant with `withHttpPort` and `withHttpHost` if we're starting clean.

server/src/main/scala/org/http4s/server/ServerBuildable.scala
```diff
@@ -0,0 +1,141 @@
...
+    withHttpPort(withHttpHost(a)(host))(port)
+
+  /** Bind the server to localhost */
+  final def bindLocal(a: A)(port: Int): A = bindHttp(a)(port, defaults.Host)
```
rossabaker, line 88: If local is the default, I don't think we need this.

server/src/main/scala/org/http4s/server/ServerBuildable.scala
```diff
@@ -0,0 +1,141 @@
...
+
+  /** Bind the server to localhost, on some dynamically determined free port. */
+  final def bindAny(a: A)(host: String = defaults.Host): A = bindHttp(a)(0, host)
```
rossabaker, line 91: This one is kind of interesting, if people don't know the magic `0` convention.  `withAnyOpenPort`, maybe?  I don't think we should have it tied up with `host` anymore.

I think all these overloads date back to looking at Unfiltered back in the Bronze Age.

server/src/main/scala/org/http4s/server/ServerBuildable.scala
```diff
@@ -0,0 +1,141 @@
...
+      signal <- Stream.eval(SignallingRef[F, Boolean](false))
+      exitCode <- Stream.eval(Ref[F].of(ExitCode.Success))
+      serve <- serveWhile(a)(signal, exitCode)
+    } yield serve
+}
+
+object ServerBuildable {
+
+  // TODO: Generate syntax with simulacrum? It's a two parameter class so I'm
+  // not certain that is viable.
```
rossabaker, line 117: I think that would also cause Scala 3 problems.

examples/blaze/src/main/scala/com/example/http4s/blaze/BlazeExample.scala
```diff
@@ -23,6 +23,7 @@ import org.http4s.server.{Router, Server}
 import org.http4s.server.blaze.BlazeServerBuilder
 import org.http4s.syntax.kleisli._
 import scala.concurrent.ExecutionContext.global
+import org.http4s.server.syntax.serverBuildable._
```
rossabaker, line 26: We already have too many imports to get anything done, and this adds another one.

rossabaker (commented):
I like the idea of cleaning up the common interface.  I've been irritated switching between backends.

I'm a bit skeptical that a typeclass encoding is the way to do it.  I'm struggling to think of a `ServerBuilder` that can't just extend the trait.  Even though typeclasses are more in line with our style, we can see in the examples how the use requires Yet Another Import, and I think having too many imports is one of the worst features of http4s today.  We save ourselves some weird self types, but push complexity to the end user.  I think this is one of the rare cases that subtype inheritance works.

server/src/main/scala/org/http4s/server/ServerBuildable.scala
isomarcte: Yeah, it is. I was thinking it would be nice to try to keep the interface similarish to the existing one.

isomarcte: I'm happy to remove though

server/src/main/scala/org/http4s/server/ServerBuildable.scala
isomarcte: I'll remove

server/src/main/scala/org/http4s/server/ServerBuildable.scala
isomarcte: :+1:

server/src/main/scala/org/http4s/server/ServerBuildable.scala
isomarcte: If we went the typeclass route, I was thinking of using https://github.com/typelevel/simulacrum-scalafix, which I think would be fine for Scala 3, though I don't know about the higher arity class support.

isomarcte:
@rossabaker so one of the more pragmatic reasons I was considering going the typeclass route is that it would allow the existing builders to stay relatively unchanged, we can map the interfaces they provide (ideal or not) into the class and thus give an easier migration path for users. 

That said, we could do the same thing with a new super type and then just have the two compose between each other.

I'll make another branch and play with a sub-typing approach.

rossabaker:
I am probably the most pro-subtyping person here, so I invite others to shout me down here.

Issue #2277, opened 2018-11-15
IdleTimeoutStage exception on the client


cquiroz:
I'm seeing an exception sometimes when doing client calls:

```
seqexec_web_server java.lang.Exception: This stage 'IdleTimeoutStage' isn't connected!
seqexec_web_server     at org.http4s.blaze.pipeline.Tail._stageDisconnected(Stages.scala:218)
seqexec_web_server     at org.http4s.blaze.pipeline.Tail.channelRead(Stages.scala:90)
seqexec_web_server     at org.http4s.blaze.pipeline.Tail.channelRead$(Stages.scala:85)
seqexec_web_server     at org.http4s.blazecore.IdleTimeoutStage.channelRead(IdleTimeoutStage.scala:12)
seqexec_web_server     at org.http4s.blazecore.IdleTimeoutStage.readRequest(IdleTimeoutStage.scala:35)
seqexec_web_server     at org.http4s.blaze.pipeline.Tail.channelRead(Stages.scala:88)
seqexec_web_server     at org.http4s.blaze.pipeline.Tail.channelRead$(Stages.scala:85)
```

Daenyth:
> It's a known issue with unknown cause. I spent an entire flight chasing that and failed.
> A good example of "we've found the bugs in Blaze, but the ones that are left are hard to fix."

- https://gitter.im/http4s/http4s?at=5efe3854b8152d34847a4187

rossabaker:
@Daenyth Had you seen this again when you bumped this?  Trying to get a feel for how fresh this is.

PR #4810, opened 2021-05-08
Read from an idle blaze-client connection to detect when server closes it

### LINKED ISSUE #4798: Blaze Client keeps half-closed connection after server sends FIN-ACK through an idle connection
This is a problem because both sides can't release resources allocated to the connection, but also, because blaze will later try and fail to use that connection.

I prepared code that reproduces this behaviour using `BlazeServer`'s `idleTimeout` as the mechanism closing the connection on server's side. 

The scenario:
- make a request 
- wait until server closes the connection (due to 5s idle timeout)
- make another request

For comparison there's the same scenario but using `AsyncHttpClient`. The behaviour of `AsyncHttpClient` is in my opinion the expected behaviour (It's also consistent with for example Firefox's behaviour).

```
  val server = BlazeServerBuilder[IO](munitExecutionContext)
    .withIdleTimeout(5.second)
    .bindLocal(12345)
    .withHttpApp(Router("/" -> HttpRoutes.of[IO] { case GET -> Root => IO.sleep(1.seconds) >> Ok("") }).orNotFound)
    .resource

  test("Blaze Client keeps half-closed connection after server sends FIN-ACK through an idle connection") {
    (
      server,
      BlazeClientBuilder[IO](munitExecutionContext).resource
    ).tupled.use { case (_, client) =>for {
      _ <- client.expect[String](Uri.fromString(s"http://127.0.0.1:12345").yolo).attemptTap(result => IO(println(result)))
      _ <- IO.sleep(6.seconds)
      _ <- client.expect[String](Uri.fromString(s"http://127.0.0.1:12345").yolo).attemptTap(result => IO(println(result)))
    } yield ()
    }.unsafeRunSync()
  }

  test("AsyncHttpClient closes connection after server sends FIN-ACK through an idle connection") {
    (
      server,
      AsyncHttpClient.resource[IO]()
    ).tupled.use { case (_, client) => for {
      _ <- client.expect[String](Uri.fromString(s"http://127.0.0.1:12345").yolo).attemptTap(result => IO(println(result)))
      _ <- IO.sleep(6.seconds)
      _ <- client.expect[String](Uri.fromString(s"http://127.0.0.1:12345").yolo).attemptTap(result => IO(println(result)))
    } yield ()
    }.unsafeRunSync()
  }
```
(Full runnable code: https://github.com/RafalSumislawski/http4s/tree/blaze-half-closed-connection-issue)

The code itself just reproduces the behaviour, without checking if it's correct. I use wireshark to analyse to communication between the client and server.

### Blaze client:
image

In packet 1519 server closes the connection. The client ACKs it, but it doesn't send a FIN, so the connection remains half-closed (CLOSE_WAIT from server's perspective). Then packet 1521 is clients attempt to use that connection to send another request, to which the server responds with RST finally killing the connection. The client then establishes a new connection and successfully resends the request. Nonetheless precious resources have been wasted on the attempt to send request though closed connection. The retry mechanism was useful in this scenario, but it just covers the problem, not fixes it. And I see some other scenarios where it causes problems.

### AsyncHttpClient:
image

After receiving FIN-ACK, the AsyncHttpClient immediately responds with FIN-ACK releasing resource on both ends. When a new request needs to be sent, I understands that it needs to open a new connection.

I realise this may be an issue inside blaze, not http4s, nonetheless I think the perspective of http4s-client user, and the comparison with other implementation is useful for clear definition of the problem and the expected behaviour.

RafalSumislawski:
Fixes #4798

### DIFF
--- a/blaze-client/src/main/scala/org/http4s/client/blaze/Http1Connection.scala
+++ b/blaze-client/src/main/scala/org/http4s/client/blaze/Http1Connection.scala
@@ -56,7 +56,7 @@ private final class Http1Connection[F[_]](
   private val parser =
     new BlazeHttp1ClientParser(maxResponseLineSize, maxHeaderLength, maxChunkSize, parserMode)
 
-  private val stageState = new AtomicReference[State](Idle)
+  private val stageState = new AtomicReference[State](Idle(None))
 
   override def isClosed: Boolean =
     stageState.get match {
@@ -64,7 +64,11 @@ private final class Http1Connection[F[_]](
       case _ => false
     }
 
-  override def isRecyclable: Boolean = stageState.get == Idle
+  override def isRecyclable: Boolean =
+    stageState.get match {
+      case Idle(_) => true
+      case _ => false
+    }
 
   override def shutdown(): Unit = stageShutdown()
 
@@ -107,9 +111,9 @@ private final class Http1Connection[F[_]](
   def resetRead(): Unit = {
     val state = stageState.get()
     val nextState = state match {
-      case Idle => Some(Idle)
+      case i: Idle => Some(i)
       case ReadWrite => Some(Write)
-      case Read => Some(Idle)
+      case Read => Some(Idle(Some(startIdleRead())))
       case _ => None
     }
 
@@ -123,9 +127,9 @@ private final class Http1Connection[F[_]](
   def resetWrite(): Unit = {
     val state = stageState.get()
     val nextState = state match {
-      case Idle => Some(Idle)
+      case i: Idle => Some(i)
       case ReadWrite => Some(Read)
-      case Write => Some(Idle)
+      case Write => Some(Idle(Some(startIdleRead())))
       case _ => None
     }
 
@@ -135,13 +139,23 @@ private final class Http1Connection[F[_]](
     }
   }
 
+  // #4798 We read from the channel while the connection is idle, in order to receive an EOF when the connection gets closed.
+  private def startIdleRead(): Future[ByteBuffer] = {
+    val f = channelRead()
+    f.onComplete {
+      case Failure(t) => shutdownWithError(t)
+      case _ =>
+    }(executionContext)
+    f
+  }
+
   def runRequest(req: Request[F], idleTimeoutF: F[TimeoutException]): F[Response[F]] =
     F.defer[Response[F]] {
       stageState.get match {
-        case Idle =>
-          if (stageState.compareAndSet(Idle, ReadWrite)) {
+        case i @ Idle(idleRead) =>
+          if (stageState.compareAndSet(i, ReadWrite)) {
             logger.debug(s"Connection was idle. Running.")
-            executeRequest(req, idleTimeoutF)
+            executeRequest(req, idleTimeoutF, idleRead)
           } else {
             logger.debug(s"Connection changed state since checking it was idle. Looping.")
             runRequest(req, idleTimeoutF)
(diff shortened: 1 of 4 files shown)


RafalSumislawski:
Is it just me or did 0.21.11 disappear from maven repo causing these mima failures?

blaze-client/src/main/scala/org/http4s/client/blaze/Http1Connection.scala
```diff
@@ -107,9 +111,9 @@ private final class Http1Connection[F[_]](
   def resetRead(): Unit = {
     val state = stageState.get()
     val nextState = state match {
-      case Idle => Some(Idle)
+      case i: Idle => Some(i)
       case ReadWrite => Some(Write)
-      case Read => Some(Idle)
+      case Read => Some(Idle(Some(startIdleRead())))
```
RaasAhsan, line 116: So the idle read will begin immediately after a connection is set to idle after being used ? Regardless of whether it is acquired again

RafalSumislawski: Exactly. This way we keep reading while the connection sits idle in the pool. If we happen to read an EOF, the connection gets closed. Once the connection is borrower and used, the read operation that we started here is used as the first operation in the process of reading a response.

RafalSumislawski:
I should have mentioned, that this PR doesn't touch the `PoolManager`, because logic preventing borrowing closed connections is already in place: https://github.com/http4s/http4s/blob/bd9a64d95a9d280f46fa0ce3ebd80f741ef96bb0/client/src/main/scala/org/http4s/client/PoolManager.scala#L181

RafalSumislawski:
This PR makes issues similar to #2277 appear much more often. Please consider merging #4815 before this PR.

blaze-client/src/main/scala/org/http4s/client/blaze/Http1Connection.scala
```diff
@@ -380,7 +416,7 @@ private object Http1Connection {
 
   // ADT representing the state that the ClientStage can be in
   private sealed trait State
-  private case object Idle extends State
+  private final case class Idle(idleRead: Option[Future[ByteBuffer]]) extends State
```
rossabaker, line 419: Is `Idle(None)` just an init state?  I might have made a separate `State` and eliminated the `Option`, but I guess we're treating both variants of `Idle` the same in lots of places.  Maybe my idea cleans it up, maybe it makes it worse.  I trust your judgment.

rossabaker (dismissed):

blaze-client/src/main/scala/org/http4s/client/blaze/Http1Connection.scala
RafalSumislawski: Yes, it's just the state before running the first request. 

I just had a second look at that state and noticed I can remove two lines of code: ff0cc53 

I tried refactoring the code to have this separate state. This is what I came up with:
Option A: https://github.com/RafalSumislawski/http4s/commit/efb7afbb79eaeb252f0caabda0bb151921d6bc47
Option B: https://github.com/RafalSumislawski/http4s/commit/ef024654a38e0bb90cf998aab5a850897fbf129c
I like the fact that the additional state makes the model of the state machine more clear, but I don't like the additional LoC. Please let me know if I should bring one of these options into this PR. If I were to make the decision, I would choose the current shape of the PR.

rossabaker: Thanks for trying.  I think you're right.  Let's keep it as is.

rossabaker:
Build of ff0cc53 failed because git checkout was 500ing.  Restarting.

rossabaker:
The MiMa failure is legitimate, because the case object became a case class.  I don't usually like exceptions, but this is private state, and should be fine.

rossabaker:
I'm just going to do the MiMa work so we can move this and others along.

Issue #4799, opened 2021-04-30
BlazeClient's idleTimeout works differently than BlazeServer's idleTimeout


RafalSumislawski:
Both `BlazeClient` and `BlazeServer` offer an `idleTimeout` configuration option. The option has the same name in both, and shares implementation (`IdleTimeoutStage`) but from user's perspective it behaves differently. 

In `BlazeClient` the `idleTimeout` is active only when a request is in-flight. It may kick in if the server takes too long to respond (unless other timeout does it first) or when the server starts sending a response and then pauses for too long. An idle (not borrowed from the pool) connection will remain open indefinitly as the `idleTimeout` is deactivated when a connection is returned to the pool. This may contradict users assumptions about the meaning of the word "idle" in the context of an HTTP client. 

In `BlazeServer` the `idleTimeout` is always active, and it may terminate connections that have an in-flight request but nothing happened for too long, as well as connections which are kept open but unused.

I see this inconsistency as a bad developer experience. In addition to that I see the lack of mechanism for evicting unused connections from the client's connection pool, as a missing feature and potential resource leak.

rossabaker:
By way of comparison, there is a `readTimeout` and a `pooledConnectionIdleTimeout` in AHC.  Jetty client has just an `idleTimeout`, and it's not clear which definition it means.

RafalSumislawski:
#3700 seems to be related. Maybe the can be solved together.

RafalSumislawski:
#2245 also seems related

RafalSumislawski:
For reference akka-http has a similar problem, and there's a PR adding a second kind of timeout to http client: https://github.com/akka/akka-http/pull/3816

rossabaker:
So if we added a `keepAliveTimeout` and made `idleTimeout` behave like blaze-client, it seems we'd be in a good place?

RafalSumislawski:
I'm not convinced what we should do with the blaze-server `idleTimeout` (or if we should do anything with it). 

What I am currently convinced to is that the changes in akka-http-client go in the right direction and we should do the same in blaze-client. It should have two timeouts:
- an `idleTimeout` that is active while the connection is being used. It's main purpose being to prevent getting stuck in this state. The current implementation does exactly that.
- and a new `keepAliveTimeout` which would be active while the connection sits in the pool.

[...]

PR #4812, opened 2021-05-09
Remove retry from blaze client

### LINKED ISSUE #4809: Blaze Client racy write/EOF retry
Two aspects of how Blaze Client operates:
1. Blaze Client writes a request and reads a response concurrently when called.
2. If Blaze Client observes an EOF when reading a response, it will open a new connection and retry the request.

This implementation admits the following execution:
1. Client submits a request. The request write is started concurrently, and the response read begins on the current fiber.
2. The write is completed and the server processes the request.
3. For whatever reason, before returning a response, the server closes the connection. This could be because of an idle timeout, a max connection age, or a server shutdown.
4. The client initiates the read and observes that the peer closed the connection. It then establishes a new connection and retries the request.

So, what happened here is that the client has opaquely retried the request in such a way that the server processes it twice. This is really dangerous for non-idempotent operations! In practice, I think seeing this execution path is unlikely because most HTTP servers are usually well-behaved and the timing is really unlikely, but the Internet is a wilderness and our client code doesn't eliminate it either.

RafalSumislawski:
This PR fixes #4809. It depends on the #4810.

blaze-client currently has a retry logic which attempts to rerun a request on a new connection if previous attempt was done on a reused connection and it failed with EOF. 

My understanding is that the intention behind that retry was to handle the scenario in which the connection as been closed while it was sitting the the connection pool, but we only learn about it once we start reading from it. #4810 proposes a better solution to that problem (by detecting that the connection is closed, before we even borrow it from the pool). 

The current retry logic is dangerous, because it either assumes that:
1. EOF in reused connection implies the connection was closed before we sent the request
2.  or that the request can be safely sent twice

Point 1. simply isn't true. The connection may be closed at any time. Including getting closed after the server received and started processing the request.

As for point 2., not all HTTP requests are idempotent. http4s doesn't know which ones can be safely retried.

Moreover the retry logic doesn't put upper bound on the number of retries. The retries are only limited by availability of reusable connections in the pool, and the `requestTimeout` (if enabled).

IMO opinion it's better to leave retry logic up to the application, which has the knowledge if the request can be safely, and what strategy of retries is desirable in given scenario (back-off period, limited number of attempts, etc.). An awesome thing about http4s (and pure FP in general) is that it composes well. For example with `cats-retry`.

### DIFF
--- a/blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClient.scala
+++ b/blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClient.scala
@@ -24,7 +24,6 @@ import cats.effect.implicits._
 import cats.syntax.all._
 import java.nio.ByteBuffer
 import java.util.concurrent.TimeoutException
-import org.http4s.blaze.pipeline.Command
 import org.http4s.blaze.util.TickWheelExecutor
 import org.http4s.blazecore.{IdleTimeoutStage, ResponseHeaderTimeoutStage}
 import org.log4s.getLogger
@@ -114,15 +113,6 @@ object BlazeClient {
                     case _ =>
                       F.delay(stageOpt.foreach(_.removeStage()))
                         .guarantee(manager.invalidate(next.connection))
-                  }
-                }
-                .recoverWith { case Command.EOF =>
-                  invalidate(next.connection).flatMap { _ =>
-                    if (next.fresh)
-                      F.raiseError(
-                        new java.net.ConnectException(s"Failed to connect to endpoint: $key"))
-                    else
-                      loop
                   }
                 }
 


hamnis (approved):
oh, this is a dangerous behaviour. We should not have retry build in. Requests are only idempotent in certain situations. Good find @RafalSumislawski

rossabaker:
I remember this only happening on the first write, which is still dangerous behavior, but at least narrows the window considerably.  But I don't see any such limitation in the code now, and I agree with the case made above.

mr-git:
We hit error like in original report for #4809 and upgraded to 0.21.24 today.
Today, on client side, we managed to get the exception with message `EOF` and failed call, while downstream server received the call and fully executed it.

Blaze sent the request, but on attempt to get response, got `EOF`.

Did we hit some regression or edge case? Is this the expected? For us it looks like a bug.

RafalSumislawski:
@mr-git Before this change blaze would send the request, get an EOF, send it again, and you would see a successful result, but the server would have processed the request twice. Now you see the EOF. This is expected outcome of this change. 

The remaining question is why you're getting the EOF. In order to answer this question we would need to see what's happening on the network (EOF corresponds to a FIN flag in TCP). Could you make a tcpdump / wireshark dump of the communication and share (sufficiently censored) screenshot in a new issue?

Is the server side also an http4s blaze? Is it also using version 0.21.24?

mr-git:
@RafalSumislawski , sadly, I am not able to sniff the traffic, it happens rarely and I do not have ability to start tcpdump there.

If we will find the way to repeat the issue reliably, I'd be able to ask for temporary dump, but not likely that I can get it now, though I will ask tomorrow. 

Yes, server-side is also http4s blaze, on 0.21.22 currently, I will upgrade it to 0.21.24 shortly, then tomorrow I will be able to check logs if there are new `EOF` cases.

mr-git:
so far getting in load-test logs getting many entries like:
```
Critical unresolvable internal error: EOF. Closing connection abruptly.
  org.http4s.blaze.pipeline.Command$EOF$: EOF
```
while working with server at 0.21.22

mr-git:
I've upgraded both client and server to 0.21.24, but only tomorrow I will be able to check if original issue is still around

mr-git:
we are getting few times EOF failures, but rate is quite low (few hundreds over some 50M calls), which is good.
still it is a bit concerning that connections in pool can go stale, but I guess there is no way around it, is there?

RafalSumislawski:
There is no way of preventing connections from dying while in the pool, but...
- In #4810 I introduced an `idleRead` which detects when a connection siting in the pool is closed. This doesn't give 100% guarantee, but it makes borrowing stale connection much less likely. Since the EOFs you're experiencing are 1 in 10^5 occurrence, I suspect that these could be a result of the fiber borrowing the connection racing against the idleRead detecting an EOF.
- The most common reason for the connection going stale is the `idleTimeout` on the server side. (see #4799). That means that you can adjust the idleTimeout of the server to change the probability. Shorter idleTimeout => more failures => easier diagnosis. Longer idleTimeout => less failures => more stable production. If this is the case, then this can be prevented. What would be needed is some kind of timeout for connections sitting idle in the client pool. Setting such timeout to a value shorter than the `idleTimeout` of the server would eliminate the connections from the pool, before they get closed by the server. As you can see I already dug into this topic, and I may implement a solution for it if I find the time and motivation.

mr-git:
@RafalSumislawski, that sounds awesome! Thank you!

PR #4815, opened 2021-05-09
Fix "This stage 'IdleTimeoutStage' isn't connected!"

### LINKED ISSUE #2277: IdleTimeoutStage exception on the client

RafalSumislawski:
Fixes #2277 

### Reproduction of #2277 
The #227 can be reasonably reliably reproduced by running this benchmark https://github.com/RafalSumislawski/http4s/blob/blaze-bench/bench/src/main/scala/org/http4s/bench/BlazeBench.scala with `sbt "bench/jmh:run -i 2 -wi 2 -f 5 -t 1 org.http4s.bench.BlazeBench"`

It results in the following stack trace:
```
[info] 20:39:47.384 [Thread-38] ERROR o.h.client.blaze.Http1Connection - Error rendering request
[info] java.lang.Exception: This stage 'IdleTimeoutStage' isn't connected!
[info] 	at org.http4s.blaze.pipeline.Tail._stageDisconnected(Stages.scala:232)
[info] 	at org.http4s.blaze.pipeline.Tail.channelWrite(Stages.scala:113)
[info] 	at org.http4s.blaze.pipeline.Tail.channelWrite$(Stages.scala:109)
[info] 	at org.http4s.blazecore.IdleTimeoutStage.channelWrite(IdleTimeoutStage.scala:28)
[info] 	at org.http4s.blazecore.IdleTimeoutStage.writeRequest(IdleTimeoutStage.scala:57)
[info] 	at org.http4s.blaze.pipeline.Tail.channelWrite(Stages.scala:111)
[info] 	at org.http4s.blaze.pipeline.Tail.channelWrite$(Stages.scala:109)
[info] 	at org.http4s.client.blaze.Http1Connection.channelWrite(Http1Connection.scala:43)
[info] 	at org.http4s.blazecore.util.IdentityWriter.writeBodyChunk(IdentityWriter.scala:73)
[info] 	at org.http4s.blazecore.util.IdentityWriter.writeEnd(IdentityWriter.scala:80)
[info] 	at org.http4s.blazecore.util.EntityBodyWriter.$anonfun$writeEntityBody$1(EntityBodyWriter.scala:63)
[info] 	at flatMap @ org.http4s.internal.package$.fromFuture(package.scala:139)
[info] 	at map @ fs2.internal.CompileScope.$anonfun$close$9(CompileScope.scala:246)
[info] 	at flatMap @ fs2.internal.CompileScope.$anonfun$close$6(CompileScope.scala:245)
[info] 	at map @ fs2.internal.CompileScope.fs2$internal$CompileScope$$traverseError(CompileScope.scala:222)
[info] 	at flatMap @ fs2.internal.CompileScope.$anonfun$close$4(CompileScope.scala:244)
[info] 	at map @ fs2.internal.CompileScope.fs2$internal$CompileScope$$traverseError(CompileScope.scala:222)
[info] 	at flatMap @ fs2.internal.CompileScope.$anonfun$close$2(CompileScope.scala:242)
[info] 	at flatMap @ fs2.internal.CompileScope.close(CompileScope.scala:241)
[info] 	at unsafeRunSync @ org.http4s.bench.BlazeBench$BlazeBenchState.<init>(BlazeBench.scala:97)
[info] 	at flatMap @ fs2.internal.CompileScope.$anonfun$openAncestor$2(CompileScope.scala:261)
[info] 	at flatMap @ fs2.internal.FreeC$.$anonfun$compile$17(Algebra.scala:545)
[info] 	at map @ fs2.internal.CompileScope.$anonfun$close$9(CompileScope.scala:246)
[info] 	at flatMap @ fs2.internal.CompileScope.$anonfun$close$6(CompileScope.scala:245)
[info] 	at map @ fs2.internal.CompileScope.fs2$internal$CompileScope$$traverseError(CompileScope.scala:222)
[info] 	at flatMap @ fs2.internal.CompileScope.$anonfun$close$4(CompileScope.scala:244)
```
### Diagnosis

The stacktrace indicates that an `Http1Connection` attempted to write something to the channel while some of the stages were disconnected.
What I think happens:
* `Http1Connection#executeRequest` starts writing the requst and reading the response in parallel (https://github.com/http4s/http4s/blob/series/0.21/blaze-client/src/main/scala/org/http4s/client/blaze/Http1Connection.scala#L199)
* `writeBody` completes (https://github.com/http4s/http4s/blob/series/0.21/blaze-core/src/main/scala/org/http4s/blazecore/util/EntityBodyWriter.scala#L62)
* the server processes the request and sends a response
* reading the response completes
* the connection is released back to the `PoolManager`
* `isRecyclable` evaluates to false because the `Http1Connection.state == Write` (https://github.com/http4s/http4s/blob/series/0.21/client/src/main/scala/org/http4s/client/PoolManager.scala#L299)
* the connection is closed (https://github.com/http4s/http4s/blob/series/0.21/client/src/main/scala/org/http4s/client/PoolManager.scala#L261) and the stages are disconnected
* `writeBodyEnd` executes (https://github.com/http4s/http4s/blob/series/0.21/blaze-core/src/main/scala/org/http4s/blazecore/util/EntityBodyWriter.scala#L63) throwing a "java.lang.Exception: This stage 'IdleTimeoutStage' isn't connected!"

This stacktrace happens relatively rarelly, but there's another symptom that appears frequenty, is easy to observ and bad for performance. wiresharking/tcpdumping the communication between http4s-blaze-client and some server shows that the client frequenty closes connections which otherwise look perfectly fine. It happens righ after a request is completed.

### Solution

Wait for the write to finish before completing the request and returning the connection to the pool.


rossabaker:
The approach seems good to me, but it broke the "Blaze Http1Client should cancel infinite request on completion" test.  Need to figure out if that's a bad mock test or is demonstrating an issue.

Also summoning @RaasAhsan since he reformed that Http1Connection state machine in #4143.

RafalSumislawski:
That "Blaze Http1Client should cancel infinite request on completion" is an interesting case. I will have to have a deeper look into it.

RafalSumislawski:
Ok, so here's what happens in "Blaze Http1Client should cancel infinite request on completion" and why it works in 0.21.22:
- the blaze client send a request with infinitely long entity.
- the jetty server doesn't consume the entity, instead it immediately responds with 200 and closes it's end of the connection
- the blaze client process the response and returns the connection to the pool (without completing the write and without noticing that the connection is half closed)
- the pool sees that the connection isn't idle (write did not finish) so it closes the connection.

At this point it is clear why this PR broke this behavior. My intention was to prevent returning connections to the pool before the write completed, and I did it, so now we don;t close the connection ad keep waiting for the infinite write (or timeout).

There are two thing I would like to do here:
- The test scenario is quite unusual. Firstly, because in most cases, the server would first consume the request entity and then send a response. Second, because because in most cases the server would try to handle request without closing connections, because closing connections is inefficient. I' not saying the test is wrong. It's very good. But I feel a need to have a wider suite of tests for infinitely long requests, to avoid breaking some more common case while fixing this edge-case.
- I think the current state is that the blaze client does the right thing (closes the connection) for the wrong reason (because it returned it to the pool before write completed). That's why fixing one problem uncovered another. IMO blaze client should close the connection, but the reason for that should be the fact that the server closed it on it's end.

rossabaker:
I think the bugfix is more important than this test scenario, and we have a good number of other fixes ready to release.  Should we just mark this as a pending test so we can include this in 0.21.23?

RafalSumislawski:
When do you plan to do the release? I'm working on a fix right now.

RafalSumislawski:
Ok. This turns out to be harder to fix than I thought. I will remove the 053b92c  from this PR, and mark the failing test as pending. I will extend the tests and fix them in a separate PR.

PR #4823, opened 2021-05-13
Bugfix/4822

### LINKED ISSUE #4822: ab load tests not completing
User report: https://discord.com/channels/632277896739946517/632286375311573032/842082988343033907

To reproduce, start up an ember server with a route installed at `/`, and run an ab load test
```
$ ab -n 5 http://localhost:8080/
```

this should be nonterminating

RaasAhsan:
Fixes #4822 . The issue is that AB operates on HTTP/1.0 and the absence of a connection header on HTTP/1.0 implies that the client wishes the connection to be terminated after the response is written

### DIFF
--- a/ember-server/src/main/scala/org/http4s/ember/server/internal/ServerHelpers.scala
+++ b/ember-server/src/main/scala/org/http4s/ember/server/internal/ServerHelpers.scala
@@ -180,18 +180,16 @@ private[server] object ServerHelpers {
   private[internal] def isKeepAlive(httpVersion: HttpVersion, headers: Headers): Boolean = {
     // We know this is raw because we have not parsed any headers in the underlying alg.
     // If Headers are being parsed into processed for in ParseHeaders this is incorrect.
-    val hasKeepAlive = headers.exists {
-      case Header.Raw(name, values) => name == connectionCi && values.contains(keepAlive.value)
+    val connection = headers.find {
+      case Header.Raw(name, values) => name == connectionCi
       case _ => false
     }
 
-    // TODO: we only need to check this if keep-alive is absent, but is there some compiler/jit optimization that infers that dependency?
-    val hasClose = headers.exists {
-      case Header.Raw(name, values) => name == connectionCi && values.contains(closeCi.value)
-      case _ => false
+    connection match {
+      case Some(header) if header.value.contains(keepAlive.value) => true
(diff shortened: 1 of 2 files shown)


ember-server/src/main/scala/org/http4s/ember/server/internal/ServerHelpers.scala
```diff
@@ -169,20 +170,31 @@ private[server] object ServerHelpers {
   private[internal] def postProcessResponse[F[_]: Timer: Monad](
       req: Request[F],
       resp: Response[F]): F[Response[F]] = {
-    val reqHasClose = req.headers.exists {
-      // We know this is raw because we have not parsed any headers in the underlying alg.
-      // If Headers are being parsed into processed for in ParseHeaders this is incorrect.
-      case Header.Raw(name, values) => name == connectionCi && values.contains(closeCi.value)
```
RaasAhsan, line 175: `toLowerCase` was missing here on a raw header! These means that if a client made a request with `Connection: Keep-Alive` (as opposed to `keep-alive`), the server would've closed the connection!

hamnis (approved):

ember-server/src/main/scala/org/http4s/ember/server/internal/ServerHelpers.scala
```diff
@@ -169,20 +170,31 @@ private[server] object ServerHelpers {
...
     } yield resp.withHeaders(Headers.of(date, connection) ++ resp.headers)
   }
 
+  private[internal] def isKeepAlive(httpVersion: HttpVersion, headers: Headers): Boolean = {
+    // We know this is raw because we have not parsed any headers in the underlying alg.
+    // If Headers are being parsed into processed for in ParseHeaders this is incorrect.
+    def hasConnection(expected: String): Boolean =
+      headers.exists {
+        case Header.Raw(name, value) =>
+          name == connectionCi && value.toLowerCase.contains(expected)
```
rossabaker, line 187: Should always use a `Locale` on `toLowerCase`.  Most people use `Locale.US`, but I use `Locale.ROOT`.

rossabaker: This will generate false positives if `keep-alive` or `close` are substrings of a different token, but I suppose those people are just asking for trouble.

PR #4824, opened 2021-05-13
Remove multiple cookies

phiSgr:
Because `putHeaders` deduplicates, calling `removeCookie` repeatedly will remove the first one.

### DIFF
--- a/core/src/main/scala/org/http4s/ResponseCookie.scala
+++ b/core/src/main/scala/org/http4s/ResponseCookie.scala
@@ -48,8 +48,8 @@ final case class ResponseCookie(
     writer
   }
 
-  def clearCookie: headers.`Set-Cookie` =
-    headers.`Set-Cookie`(copy(content = "", expires = Some(HttpDate.Epoch)))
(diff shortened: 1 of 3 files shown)


core/src/main/scala/org/http4s/ResponseCookie.scala
```diff
@@ -48,8 +48,8 @@ final case class ResponseCookie(
     writer
   }
 
-  def clearCookie: headers.`Set-Cookie` =
-    headers.`Set-Cookie`(copy(content = "", expires = Some(HttpDate.Epoch)))
+  def clearCookie: ResponseCookie =
+    copy(content = "", expires = Some(HttpDate.Epoch))
```
rossabaker, line 52: Using `copy` leaves other attributes.  Is this desirable? I think yes?

rossabaker (approved):
Thanks! I think we should backport this to 0.22, as the new header model originates there.

rossabaker:
Argh, I meant to cherry-pick this!

PR #4827, opened 2021-05-15
Change CORSConfig to builder pattern

### LINKED ISSUE #1717: Make `CORSConfig` more typesafe
It would be nice to have a `CORSConfig` more typesafe. This is the definition at the moment ([Source]( https://github.com/http4s/http4s/blob/master/server/src/main/scala/org/http4s/server/middleware/CORS.scala#L18)):

```scala
final case class CORSConfig(
    anyOrigin: Boolean,
    allowCredentials: Boolean,
    maxAge: Long,
    anyMethod: Boolean = true,
    allowedOrigins: String => Boolean = _ => false,
    allowedMethods: Option[Set[String]] = None,
    allowedHeaders: Option[Set[String]] = Set("Content-Type", "*").some,
    exposedHeaders: Option[Set[String]] = Set("*").some
)
```

I propose to have at least `maxAge` as a `FiniteDuration` and `allowedMethods` as `Option[Set[org.http4s.Method]]`. Probably `allowedHeaders` and `exposedHeaders` could make use of the `org.http4s.Header` type as well. `org.http4s.Headers` exists but its internal representation is a `List`.

Don't mind submitting a PR with these changes.

qwbarch:
Fixes #1717. I wasn't sure how to use ``Method``'s generics so I left it as a String for now.

### DIFF
--- a/server/src/main/scala/org/http4s/server/middleware/CORS.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/CORS.scala
@@ -31,7 +31,7 @@ import scala.concurrent.duration._
   * to specify its behavior
   */
 
-final class CORSConfigBuilder private (
+final class CORSConfig private (
     val anyOrigin: Boolean,
     val allowCredentials: Boolean,
     val maxAge: FiniteDuration,
@@ -51,7 +51,7 @@ final class CORSConfigBuilder private (
       allowedMethods: Option[Set[Method]] = allowedMethods,
       allowedHeaders: Option[Set[String]] = allowedHeaders,
       exposedHeaders: Option[Set[String]] = exposedHeaders
-  ) = new CORSConfigBuilder(
+  ) = new CORSConfig(
     anyOrigin,
     allowCredentials,
     maxAge,
@@ -62,26 +62,31 @@ final class CORSConfigBuilder private (
     exposedHeaders
   )
 
-  def withAnyOrigin(anyOrigin: Boolean): CORSConfigBuilder = copy(anyOrigin = anyOrigin)
+  def withAnyOrigin(anyOrigin: Boolean): CORSConfig = copy(anyOrigin = anyOrigin)
 
-  def withAllowCredentials(allowCredentials: Boolean): CORSConfigBuilder = copy(allowCredentials = allowCredentials)
+  def withAllowCredentials(allowCredentials: Boolean): CORSConfig =
+    copy(allowCredentials = allowCredentials)
 
-  def withMaxAge(maxAge: FiniteDuration): CORSConfigBuilder = copy(maxAge = maxAge)
+  def withMaxAge(maxAge: FiniteDuration): CORSConfig = copy(maxAge = maxAge)
 
-  def withAnyMethod(anyMethod: Boolean): CORSConfigBuilder = copy(anyMethod = anyMethod)
+  def withAnyMethod(anyMethod: Boolean): CORSConfig = copy(anyMethod = anyMethod)
 
-  def withAllowedOrigins(allowedOrigins: String => Boolean): CORSConfigBuilder = copy(allowedOrigins = allowedOrigins)
+  def withAllowedOrigins(allowedOrigins: String => Boolean): CORSConfig =
+    copy(allowedOrigins = allowedOrigins)
 
-  def withAllowedMethods(allowedMethods: Option[Set[Method]]): CORSConfigBuilder = copy(allowedMethods = allowedMethods)
+  def withAllowedMethods(allowedMethods: Option[Set[Method]]): CORSConfig =
+    copy(allowedMethods = allowedMethods)
 
-  def withAllowedHeaders(allowedHeaders: Option[Set[String]]): CORSConfigBuilder = copy(allowedHeaders = allowedHeaders)
+  def withAllowedHeaders(allowedHeaders: Option[Set[String]]): CORSConfig =
+    copy(allowedHeaders = allowedHeaders)
 
-  def withExposedHeaders(exposedHeaders: Option[Set[String]]): CORSConfigBuilder = copy(exposedHeaders = exposedHeaders)
+  def withExposedHeaders(exposedHeaders: Option[Set[String]]): CORSConfig =
+    copy(exposedHeaders = exposedHeaders)
 }
 
-object CORSConfigBuilder {
+object CORSConfig {
 
-  def apply(): CORSConfigBuilder = new CORSConfigBuilder(
+  def apply(): CORSConfig = new CORSConfig(
     anyOrigin = true,
     allowCredentials = true,
     maxAge = 1.day,
@@ -103,7 +108,7 @@ object CORS {
     * based on information in CORS config.
     * Currently, you cannot make permissions depend on request details
     */
-  def apply[F[_], G[_]](http: Http[F, G], config: CORSConfigBuilder = CORSConfigBuilder())(implicit
+  def apply[F[_], G[_]](http: Http[F, G], config: CORSConfig = CORSConfig())(implicit
       F: Applicative[F]): Http[F, G] =
     Kleisli { req =>
       // In the case of an options request we want to return a simple response with the correct Headers set.
--- a/server/src/main/scala/org/http4s/server/middleware/CORS.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/CORS.scala
@@ -181,10 +181,10 @@ object CORS {
         (config.anyOrigin, config.anyMethod, origin.value, acrm.value) match {
           case (true, true, _, _) => true
           case (true, false, _, acrm) =>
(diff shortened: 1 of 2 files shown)


hamnis:
Hello, and thanks for your contribution.

Needs a `sbt scalafmtAll`

I would expect a builder to have a build, method or something that produces the final result.  Since this builder is immutable, i guess that is not really a problem.

Now we just really have the same as we did, but not able to do equality checks and other useful operations.

Perhaps we should keep the old type around, but make it a non-case class, but with equals, hashCode and toString implemented. Not sure.

qwbarch:
I wasn't sure what to do with ``allowedOrigins: String => Boolean``, so I left it out for ``equals``, and ``hashCode``.
How do I go about checking for equality and calculating a function's hash?

rossabaker:
Yeah, we call this "the builder pattern", but maybe "the [contraband](https://www.scala-sbt.org/contraband/) pattern" is more accurate.  I don't think it needs `build` or `result` method.  We're just trying to promote binary compatibility moving forward.

Function equality is hard.  We'd like two functions `A => B` to be equal if and only if for each `A` the two functions each return the same `B`.  But our `A` is `String`, with infinite inhabitants, so we can't possibly test this!  The default definition is reference equality, and it's really the best we can do.  I'd just use `==` and `##`.  The alternative is to redesign that function into some more constrained AST, but that's not really worth it to recover equality.

server/src/main/scala/org/http4s/server/middleware/CORS.scala
```diff
@@ -25,36 +25,117 @@ import org.http4s.Method.OPTIONS
...
+        maxAge === config.maxAge &&
+        anyMethod === config.anyMethod &&
+        allowedMethods === config.allowedMethods &&
+        allowedHeaders === config.allowedHeaders &&
+        exposedHeaders === config.exposedHeaders
+    case _ => false
+  }
+
+  override def hashCode(): Int =
+    Objects.hashCode(
```
rossabaker, line 100: Wow, I didn't know this existed.  `MurmurHash3` is standard in Scala, and what case classes use.  But this is a bit more concise.  I think I like it!

server/src/main/scala/org/http4s/server/middleware/CORS.scala
```diff
@@ -25,36 +25,117 @@ import org.http4s.Method.OPTIONS
...
+      anyOrigin,
+      allowCredentials,
+      maxAge,
+      anyMethod,
+      allowedMethods,
+      allowedHeaders,
+      exposedHeaders
+    )
+
+  override def toString(): String = s"CORSConfig($anyOrigin,$allowCredentials,$maxAge," +
```
rossabaker, line 110: There's really no reason to force a string concat here.  The source line gets a little long, but that's okay.

server/src/main/scala/org/http4s/server/middleware/CORS.scala
```diff
@@ -25,36 +25,117 @@ import org.http4s.Method.OPTIONS
...
+    s"$anyMethod,$allowedOrigins,$allowedMethods,$allowedHeaders,$exposedHeaders)"
+}
+
+object CORSConfig {
+
+  def apply(): CORSConfig = new CORSConfig(
```
rossabaker, line 116: This is an immutable type.  A `val default` makes sense to me.

rossabaker: Eh, most apps will only ever have one.  This really doesn't matter at all, I guess.

rossabaker:
We should cherry-pick this to series/0.22.

qwbarch:
I cherry-picked my latest commit and brought it over to ``series/0.22``, and then changed this PR's base to that branch as well.
This doesn't look quite right though, did I do it wrong?

server/src/main/scala/org/http4s/server/middleware/CORS.scala
qwbarch: I searched up how to calculate hashes in Scala and found a different solution revolving around ``ScalaRunTime``, thankfully another answer showed this simple Java ``Objects#hashCode``

qwbarch:
Whoops, I just realized this isn't even merged properly when I was cherry-picking. Going to revert it and be more careful next time

qwbarch:
I've also realized I can't switch the source branch, only the target branch (which I don't think makes a difference now that I think about it).
I've switched the target branch back to main for now since it brought in a ton of participants, I don't want to end up giving everyone a notification email 😅

rossabaker:
Don't worry about the targeting. I can cherry-pick it so it shows up as you and gets to the right place. :smile:

server/src/main/scala/org/http4s/server/middleware/CORS.scala
rossabaker: I think it needs to be `Objects.hash` (the variadic thing that hashes multiple elements), not `Objects.hashCode` (takes a single object, which the Scala compiler is trying to auto-tuple).

server/src/main/scala/org/http4s/server/middleware/CORS.scala
```diff
@@ -88,18 +171,18 @@ object CORS {
...
+            "Access-Control-Max-Age" -> config.maxAge.toSeconds.toString
           )
       }
 
       def allowCORS(origin: Header.Raw, acrm: Header.Raw): Boolean =
         (config.anyOrigin, config.anyMethod, origin.value, acrm.value) match {
           case (true, true, _, _) => true
           case (true, false, _, acrm) =>
-            config.allowedMethods.exists(_.contains(acrm))
+            config.allowedMethods.exists(_.find(_.toString() === acrm).nonEmpty)
```
rossabaker, line 182: This call, and the one below it, look off to me.  Don't we need to parse the ACRM header and compare methods?

qwbarch: For the old implementation (before my commits), doesn't ``allowedMethods`` hold the method names directly?
I might be really misunderstanding this still, but wouldn't calling ``Method#name`` (or in this case ``toString``) be the same?

server/src/main/scala/org/http4s/server/middleware/CORS.scala
qwbarch: Whoops, I completely missed that. Nice catch, sorry about that!

server/src/main/scala/org/http4s/server/middleware/CORS.scala
rossabaker: Oh, I see.  I was thinking of the ACRM header as having multiple values, but it's just one.  Yes, I think this logic is correct.

rossabaker (dismissed):

rossabaker:
Looks like that `Objects.hash` doesn't work on primitives on Scala 2.12 due to an implicit conversion.

I can dig up an example of MurmurHash3 if we want to do it the way it's generated for Scala case classes.  Or there are a few examples in our codebase.

qwbarch:
Ah that's unfortunate. I'll look into MurmurHash3 from the other examples

qwbarch:
> Looks like that `Objects.hash` doesn't work on primitives on Scala 2.12 due to an implicit conversion.

I'm curious though, where does the error show up? I want to avoid running into these issues myself but there's no compiler error nor is there a failed test.

rossabaker:
> since there's an implicit conversion, an incorrect hash is generated I think right?

It hit this error in Scala 2.12.  Something that was fine in 2.13 and 3:

```
[error] /home/runner/work/http4s/http4s/server/src/main/scala/org/http4s/server/middleware/CORS.scala:102:7: the result type of an implicit conversion must be more specific than Object
[error]       anyOrigin,
[error]       ^
[error] /home/runner/work/http4s/http4s/server/src/main/scala/org/http4s/server/middleware/CORS.scala:103:7: the result type of an implicit conversion must be more specific than Object
[error]       allowCredentials,
[error]       ^
[error] /home/runner/work/http4s/http4s/server/src/main/scala/org/http4s/server/middleware/CORS.scala:105:7: the result type of an implicit conversion must be more specific than Object
[error]       anyMethod,
[error]       
```

rossabaker (approved):
LGTM. I'm going to cherry-pick it all back to series/0.22, because there's no CE work to backport.

qwbarch:
Thanks! Just wondering since I didn't see anything about this on the contribution guide, is there a limit to how many one can work on ``Good First Issue`` labeled issues? 
The other issues are a bit advanced for my current skill-level so I'm looking to work on some easier ones until I'm more familiar with the library

rossabaker:
Nope, absolutely no limit!  Just please leave a comment when you start to look at one, so we know we're not duplicating effort.  And thanks for the great work on this one!

qwbarch:
Good to hear, thanks for helping me out with the issue. Looking forward to working on other issues!

PR #4831, opened 2021-05-16
Fix blaze client handling of early responses

RafalSumislawski:
Follow-up to infinite request body issues in #4815

It's a draft. It fixes the issue, but I need to clean up, and fix some tests, add some more tests, and probably do 100 other things.

### DIFF
--- a/blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClient.scala
+++ b/blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClient.scala
@@ -103,17 +103,18 @@ object BlazeClient {
                 case Some(stage) => F.async[TimeoutException](stage.init)
                 case None => F.never[TimeoutException]
               }
-              val res = next.connection
+              val res: F[Resource[F, Response[F]]] = next.connection
                 .runRequest(req, idleTimeoutF)
-                .map { r =>
-                  Resource.makeCase(F.pure(r)) {
-                    case (_, ExitCase.Completed) =>
-                      F.delay(stageOpt.foreach(_.removeStage()))
-                        .guarantee(manager.release(next.connection))
-                    case _ =>
-                      F.delay(stageOpt.foreach(_.removeStage()))
-                        .guarantee(manager.invalidate(next.connection))
-                  }
+                .map { response: Resource[F, Response[F]] =>
+                  response.flatMap(r =>
+                    Resource.makeCase(F.pure(r)) {
+                      case (_, ExitCase.Completed) =>
+                        F.delay(stageOpt.foreach(_.removeStage()))
+                          .guarantee(manager.release(next.connection))
+                      case _ =>
+                        F.delay(stageOpt.foreach(_.removeStage()))
+                          .guarantee(manager.invalidate(next.connection))
+                    })
                 }
 
               responseHeaderTimeout match {
--- a/blaze-client/src/main/scala/org/http4s/client/blaze/Http1Connection.scala
+++ b/blaze-client/src/main/scala/org/http4s/client/blaze/Http1Connection.scala
@@ -147,8 +147,8 @@ private final class Http1Connection[F[_]](
     f
   }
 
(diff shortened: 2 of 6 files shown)


RafalSumislawski:
#4815 introduced waiting for the request write to finish. Because this waiting happens before processing of the response (after reading the response), it behaves badly in the edge case when the server sends the response while the client has still a lot of data to send in the request.

In this PR this behaviour changes, so that the response is processed right after receiving it, and then we wait for the write to complete. Or if the server killed the connection after sending the response, then we don't wait (the EOF in idle read should close the connection interrupting the write).

@rossabaker I cleaned up the PR and commit history and it is now ready for a review.

blaze-client/src/main/scala/org/http4s/client/blaze/Http1Connection.scala
```diff
@@ -211,18 +211,18 @@ private final class Http1Connection[F[_]](
...
+            val response: F[Resource[F, Response[F]]] = for {
               writeFiber <- writeRequest.start
               response <- receiveResponse(
                 mustClose,
                 doesntHaveBody = req.method == Method.HEAD,
                 idleTimeoutS,
-                idleRead)
-              _ <- writeFiber.join
-            } yield response
+                idleRead).attemptTap(e => Sync[F].delay(println(e.toString)))
```
hamnis, line 220: forgotten println?

blaze-client/src/test/scala/org/http4s/client/blaze/BlazeClientBase.scala
```diff
@@ -73,10 +73,28 @@ trait BlazeClientBase extends Http4sSuite {
           case None => srv.sendError(404)
         }
 
-      override def doPost(req: HttpServletRequest, resp: HttpServletResponse): Unit = {
-        resp.setStatus(Status.Ok.code)
-        req.getInputStream.close()
-      }
+      override def doPost(req: HttpServletRequest, resp: HttpServletResponse): Unit =
```
rossabaker, line 76: I just ported this to the com.sun server so we have a path to Jetty 11, so we have a merge conflict looming.

blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClient.scala
```diff
@@ -103,17 +103,18 @@ object BlazeClient {
                 case Some(stage) => F.async[TimeoutException](stage.init)
                 case None => F.never[TimeoutException]
               }
-              val res = next.connection
+              val res: F[Resource[F, Response[F]]] = next.connection
```
rossabaker, line 106: This might end up being nice in 1.0, as we move toward responses being resources.

rossabaker (commented):
Looks good after the println.

Do you need an 0.21 release?  I'm not actively maintaining that now that 0.22 is in RC, but it's not hard to cut releases as long as you're motivated to keep fixing bugs there.

blaze-client/src/main/scala/org/http4s/client/blaze/Http1Connection.scala
RafalSumislawski: indeed. fixed.

RafalSumislawski:
> Do you need an 0.21 release?

Thanks. Maybe later. I may find time to address some more issues in 0.21, and then we could have a release with a bunch of fixes. I'm of course planning to move over to using and contributing to 0.22, but I'll wait until I'll be able to use it with rho.

blaze-client/src/test/scala/org/http4s/client/blaze/BlazeClientBase.scala
RafalSumislawski: I noticed it, but there wasn't much I could do if I wanted to target 0.21.

RafalSumislawski:
I noticed this change can cause some "*stage not connected" in case of request cancellation (`BlazeClient213Suite` tests with parTraver / parSequence).

hamnis:
test failure is unrelated

rossabaker (approved):

PR #4830, opened 2021-05-16
Scala 3.0.0

rossabaker:
Broken on Scala 3, but so is the main branch.

Missing:
* [x] blaze (https://github.com/http4s/blaze/pull/531)
* [x] case-insensitive (https://github.com/typelevel/case-insensitive/pull/153)
* [x] vault (https://github.com/typelevel/vault/pull/234)
* [x] unique (https://github.com/typelevel/unique/pull/191)
* [x] log4s (https://github.com/Log4s/log4s/pull/76)
* [x] keypool (https://github.com/typelevel/keypool/pull/211)
* [x] jawn-fs2 (https://github.com/typelevel/jawn-fs2/pull/204)

PR #4833, opened 2021-05-17
0.21 to 0.22

RaasAhsan:


rossabaker:
Scala 3 build is broken pending #4830.  The other failure is a flake.  I'd rather get the divergent branches synchronized.

PR #4838, opened 2021-05-17
Repackage client-blaze to blaze-client

krayoola:
Repackaging org.http4s.client.blaze -> org.http4s.blaze.client
Part of #2702

### DIFF
--- a/blaze-client/src/main/scala/org/http4s/blaze/client/BasicManager.scala
+++ b/blaze-client/src/main/scala/org/http4s/blaze/client/BasicManager.scala
@@ -15,13 +15,25 @@
  */
 
 package org.http4s
-package client
 package blaze
+package client
 
-import java.nio.ByteBuffer
-import java.util.concurrent.TimeoutException
-import org.http4s.blaze.pipeline.TailStage
+import cats.effect._
+import cats.syntax.all._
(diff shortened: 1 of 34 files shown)


blaze-client/src/main/scala/org/http4s/blaze/client/BasicManager.scala
```diff
@@ -15,11 +15,12 @@
  */
 
 package org.http4s
-package client
 package blaze
+package client
 
 import cats.effect._
 import cats.syntax.all._
+import org.http4s.client.{Connection, ConnectionBuilder, RequestKey}
```
rossabaker, line 23: `Connection` and `ConnectionBuilder` look like things that can be relocated, but that's for another ticket.

rossabaker (approved):
Nicely done!

rossabaker:
Oh, this branch is broken pending #4830. I think the CI results are what we expect here.  Merging.

blaze-client/src/main/scala/org/http4s/blaze/client/BasicManager.scala
krayoola: Sure, Seems like a chore do you have a ticket for that maybe i can also give it a try.

PR #4840, opened 2021-05-17
Fix Cats-Effect version for 1.0.x

ant8e:


aeons:
That's actually the `cats` version which is currently 2.6.x. We pull in `cats-effect` through `fs2`, which is at version 3.x.

ant8e:
Oh. I see it  now. Sorry about that

PR #4844, opened 2021-05-18
Fixed mdocs from blazeclient repackaging

krayoola:
Part of #2702, Missed changes from #4838 PR


aeons:
Thanks
There are some errors in scalafix that need the same treatment. Could you do that here as well?

krayoola:
@aeons Fixed scala fix. I do have another error related in `cors.md` seems it came from the last cherry picked changes, will try to fix that too.

krayoola:
Should be fixed now.

aeons:
Thanks!

Issue #2497, opened 2019-04-10
Client.fromHttpApp in Testing Documentation


ChristopherDavenport:
Our Testing Section of the documentation does not mention `Client.fromHttpApp`, or for that matter mention client testing, which is a significant portion of testing that individuals write. 

This is an place for more documentation under a `Client` header in on the `Testing` page. 

Where we introduce testing a client through this functionality.

Should be a good first issue, and will help to point people to existing testing approaches.

vasiliybondarenko:
@ChristopherDavenport would you mind if create PR for this?
We are talking about this doc `docs/src/main/tut/testing.md`, right?

key-eugene:
I guess it was a long tine since someone took this issue, I'll take this one.

rossabaker:
Thanks, @key-eugene.

I'm sorry we all overlooked @vasiliybondarenko's offer earlier.  If you're still interested in contributing, we have plenty of similar tickets!

aeons:
Closed by #4845

PR #4845, opened 2021-05-18
added docs for Client.fromHttpApp 

key-eugene:
#2497

### DIFF
--- a/docs/src/main/mdoc/testing.md
+++ b/docs/src/main/mdoc/testing.md
@@ -111,6 +111,31 @@ val response: IO[Response[IO]] = service[IO](doesNotMatter).orNotFound.run(
 check[String](response, Status.NotFound, Some("Not found"))
 ```
 
+### Using client
+
+Having HttpApp you can build a client for testing purposes. Following the example above we could define our HttpApp like this:
+```scala mdoc:nest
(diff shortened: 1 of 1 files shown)


rossabaker (approved):

docs/src/main/mdoc/testing.md
```diff
@@ -111,6 +111,27 @@ val response: IO[Response[IO]] = service[IO](doesNotMatter).orNotFound.run(
...
+
+```scala mdoc:nest
+val httpApp: HttpApp[IO] = service[IO](success).orNotFound
+```
+From this, we can obtain the `Client` instance using `Client.fromHttpApp` and then use it to test our sever/app.
+
+```scala mdoc:nest
+val client = Client.fromHttpApp(httpApp)
+val request: Request[IO] = Request(method = Method.GET, uri = uri"/user/not-used")
+val expectedJson = Json.obj(
```
aeons, line 126: This can also be written like this:
```scala
Json.obj(
  "name" := "johndoe",
  "age" := 42
)
```
which, imo, is a bit easier to read.

key-eugene: Sure, no problem!

aeons (approved):

PR #4847, opened 2021-05-18
Repackage server-blaze to blaze-server

krayoola:
Repackaging org.http4s.server.blaze -> org.http4s.blaze.server
Part of #2702


aeons:
Thanks
Docs and scalafix need updates as well

krayoola:
@aeons  everything should be passing now :)

aeons (approved):

PR #4848, opened 2021-05-19
Fix TimeoutException: Future timed out after [30 seconds] on BlazeClient213Suite testing

Kevin-Lee:
## Fix TimeoutException: Future timed out after [30 seconds] on BlazeClient213Suite testing
I usually get the following failures when I run the tests locally.
* `BlazeClient213Suite.behave and not deadlock on failures with parTraverse`
* `BlazeClient213Suite.Blaze Http1Client should behave and not deadlock on failures with parSequence`

fail with `TimeoutException: Future timed out after [30 seconds]`, this PR may fix it.

### DIFF
--- a/blaze-client/src/test/scala-2.13/org/http4s/client/blaze/BlazeClient213Suite.scala
+++ b/blaze-client/src/test/scala-2.13/org/http4s/client/blaze/BlazeClient213Suite.scala
@@ -23,12 +23,15 @@ import cats.syntax.all._
 import fs2.Stream
 import org.http4s._
 import org.http4s.blaze.client.BlazeClientBase
+
+import java.util.concurrent.TimeUnit
 import scala.concurrent.duration._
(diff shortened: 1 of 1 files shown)


rossabaker:
Are we sure this won't be flaky again at a 50-second timeout?  We never got to the bottom of what locks it up.

Kevin-Lee:
@rossabaker I tested about 20 times and never had any issues once I put that 50 second timeout.
Just in case, let me have more tests on it.

Kevin-Lee:
Also I'll create a PR to the `main` once the test is done and if I'm confident this solves the issue.

rossabaker:
Alright, we can give it a shot, and always reflag them as flaky if they don't work.

rossabaker:
Drat: https://github.com/http4s/http4s/pull/4864/checks?check_run_id=2642699996

Issue #4828, opened 2021-05-16
GZipSuite failures on Java 16


rossabaker:
May be coincidental, but I think this is our first run on Java 16. From https://github.com/http4s/http4s/pull/4826/checks?check_run_id=2593254945:

```
 ==> X org.http4s.server.middleware.GZipSuite.encoding  0.085s munit.FailException: /home/runner/work/http4s/http4s/server/src/test/scala/org/http4s/server/middleware/GZipSuite.scala:69
68:
69:  test("encoding") {
70:    PropF.forAllF { vector: Vector[Array[Byte]] =>
Failing seed: 0UBTmkIPcwfYn8s_NwZ50zpBlRVoK9Yh2hgmaNw5oVA=
You can reproduce this failure by adding the following override to your suite:

  override val scalaCheckInitialSeed = "0UBTmkIPcwfYn8s_NwZ50zpBlRVoK9Yh2hgmaNw5oVA="

Exception raised on property evaluation.
> ARG_0: Vector()
> Exception: munit.ComparisonFailException: /home/runner/work/http4s/http4s/server/src/test/scala/org/http4s/server/middleware/GZipSuite.scala:86
85:
86:      actual.map(Arrays.equals(_, expected)).assert
87:    }
value is not true
=> Obtained
false
=> Diff (- obtained, + expected)
-false
+true
```

```
==> X org.http4s.server.middleware.GZipSuite.encodes random content-type if given isZippable is true  0.056s munit.ComparisonFailException: /home/runner/work/http4s/http4s/server/src/test/scala/org/http4s/server/middleware/GZipSuite.scala:66
65:
66:    actual.map(Arrays.equals(_, byteStream.toByteArray)).assert
67:  }
value is not true
=> Obtained
false
=> Diff (- obtained, + expected)
-false
+true
```

aeons:
I looked at this a bit and it seems that fs2 and GZIPOutputStream disagree on what OS the test is running on. I added this comment above the ignored tests:
```scala
  // TODO: This test fails since fs2 and GZIPOutputStream disagree on the OS
  // byte in the gzip header
  // fs2: 1F 8B 08 00 00 00 00 00 00 00
  // gos: 1F 8B 08 00 00 00 00 00 00 FF
  // The last byte is the OS, 00 is FAT filesystem, while FF is unknown
```

rossabaker:
Fixed in #4871.

PR #4850, opened 2021-05-19
Fix tests

aeons:
This PR fixes the Jetty tests by reapplying #4783 and fixes the gzip tests by ignoring them on JVM 16.

See #4828 for more info.


aeons:
This will fail on the 3.0.0-RC2 builds.

PR #4851, opened 2021-05-19
Merge series/0.22 into main (again)

aeons:
This time, it will bring scala 3.0.0.

It bulds off of #4850 and still requires https://github.com/typelevel/vault/pull/233 to be merged and released.


core/src/main/scala/org/http4s/Uri.scala
```diff
@@ -654,7 +654,7 @@ object Uri extends UriPlatform {
     def toByteArray: Array[Byte] =
       address.toBytes
 
-    def toInet6Address: Inet6Address =
+    def toInetAddress: InetAddress =
```
aeons, line 657: This type change is due to https://github.com/Comcast/ip4s/pull/275

aeons:
Last dependency needed is keypool: https://github.com/typelevel/keypool/pull/213

PR #4857, opened 2021-05-21
Round of dependency bumps for 0.22

rossabaker:


aeons:
This still depends on djspiewak/sbt-spiewak#56

rossabaker:
Yep, it started working on other projects, and I forgot it didn't work here.

Issue #4865, opened 2021-05-22
ServerSentEvent support for data being a multi-line string and possible other ServerSentEvent enhancements


fizzy33:
For server sent events data can be multi-line and in that case multiple `data:` lines should be sent as outlined here https://html.spec.whatwg.org/multipage/server-sent-events.html

The current implementation only sends a single `data:` line followed by the entire data string.  So if that data string 

```
final case class ServerSentEvent(
    data: String,
    eventType: Option[String] = None,
    id: Option[EventId] = None,
    retry: Option[Long] = None
) extends Renderable {
  def render(writer: Writer): writer.type = {
    writer << "data: " << data << "\n"
    eventType.foreach(writer << "event: " << _ << "\n")
    id match {
      case None =>
      case Some(EventId.reset) => writer << "id\n"
      case Some(EventId(id)) => writer << "id: " << id << "\n"
    }
    retry.foreach(writer << "retry: " << _ << "\n")
    writer << "\n"
  }

  override def toString: String =
    s"ServerSentEvent($data,$eventType,$id,$retry)"
}
```

Here is the fix we are using in production noting it does a few more things than just fix the multi-line data issue that are worth considering i.e. comments and data is not required which allow for being a more complete sse implementation.

If these look good (or just the data fix only) I am happy to put together an MR.  Noting sending comment lines is the idiomatic ping for sse so it is critical for any robust real world implementation of SSE.

data changing from String to Option[String] is likely a breaking change but a worthwhile one to do now to get a complete ServerSentEvent implementation.  The change in http4s's internals appears trivial (only a few places) clients using it though will break so it may be best to time that change with a larger version change.  

```

  final case class ServerSentEvent(
      data: Option[String] = None,
      comment: Option[String] = None,
      eventType: Option[String] = None,
      id: Option[EventId] = None,
      retry: Option[FiniteDuration] = None
  ) extends Renderable {
    def render(writer: Writer): writer.type = {
      data.foreach ( _.linesIterator.foreach { dataLine =>
        writer << "data: " << dataLine << "\n"
      })
      comment.foreach(_.linesIterator.foreach { commentLine =>
        writer << ": " << commentLine << "\n"
      })
      eventType.foreach(writer << "event: " << _ << "\n")
      id match {
        case None =>
        case Some(EventId.reset) => writer << "id\n"
        case Some(EventId(id)) => writer << "id: " << id << "\n"
      }
      retry.foreach(writer << "retry: " << _.toMillis << "\n")
[...]

PR #4866, opened 2021-05-22
fix for ServerSentEvent.data being a multi-line string -- issue 4865

fizzy33:


core/src/main/scala/org/http4s/ServerSentEvent.scala
```diff
@@ -30,7 +30,9 @@ final case class ServerSentEvent(
     retry: Option[Long] = None
 ) extends Renderable {
   def render(writer: Writer): writer.type = {
-    writer << "data: " << data << "\n"
+    data.linesIterator.foreach { dataLine =>
```
rossabaker, line 33: I brought this up on the other ticket.  `linesIterator` will split on `\r`, `\r\n`, and `\n`.  This renderer will reassemble it with just `\n`, so it's a lossy conversion.  It's legal to separate the lines with `\r` or `\r\n`, so it doesn't _have_ to be lossy.

This is probably an exotic case, and what you've done is inarguably an improvement for real-world use.

tests/src/test/scala/org/http4s/ServerSentEventSpec.scala
```diff
@@ -117,6 +121,11 @@ class ServerSentEventSpec extends Http4sSuite {
         .through(ServerSentEvent.decoder)
         .compile
         .toVector
+      val f = roundTrip.unsafeToFuture()
+      val rt = Await.result(f, Duration(10, TimeUnit.SECONDS))
+      if ( rt != sses ) {
+        toString
+      }
```
rossabaker, line 128: I'm confused what's going on here, particularly with the blocking.  Is this for a better failure message?

rossabaker (commented):
We can cherry-pick this to series/0.22.  The newer model described in #4865 could even go into series/0.22, if it beats our final release.  But don't worry about those logistics: we can cherry-pick it when the time comes.

tests/src/test/scala/org/http4s/ServerSentEventSpec.scala
fizzy33: Ooof I will remove that is me hacking in something for debugging...

fizzy33: That code should have been removed.

fizzy33:
moving the linesIterator talk here.

So switching to using this in render
```
    data.linesWithSeparators.foreach { dataLine =>
      writer << "data: " << dataLine
    }
```

Causes a bunch of regressions because the decoder uses `fs2.text.lines`. i.e. the decoder doesn't preserve line endings.  

So happy to chase that further if we think it is worth it.  Which one way I see would be have a 
```
  def linesWithEnding[F[_]]: Pipe[F, String, (String,Option[String])] = {
```
That the decoder uses.  Noting the Option[String] ending could interestingly be an enum CR, LF, CRLF or String (liking having a special type since it makes the API clear what is what.

rossabaker:
A sum type for the line ending is fine with me, but I don't know what other "line endings" exist in exotic parts of Unicode if this is a general purpose `Pipe`.  Or might some other protocols consider a form feed character?  I think a sum type binds that pipe closely to SSE.

If that fs2 side trek is easy, let's do it.  If not, I don't think the line endings are nearly as much a real world concern as the multi-line rendering, or comments, and I'd be happy to clean that up later.  But if we punt on it, the property test should normalize carriage returns to `\n`, or it will be flaky.

rossabaker:
0.22 and 0.23 release candidates are imminent, probably as soon as Circe is released.  If we want to get comment support, we'll need to add it to the model for bincompat reasons, or else it will have to wait for 1.0.  We might want to prioritize that over the carriage return preservation.

rossabaker:
I think this is all included in #4888?

fizzy33:
Agreed #4865 covers the multi-line data (and now comment as well).

rossabaker:
I'll leave the original open to track the `\r` nuance, but we'll close this.

PR #4871, opened 2021-05-23
Scala 2.13.6

aeons:
Bumps `kind-projector` via `sbt-spiewak` via `sbt-http4s-org` which allows us to upgrade scala to 2.13.6.

Also allows us to drop `sbt-dotty`

Closes #4839

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,8 +27,8 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        scala: [2.12.13, 2.13.6, 3.0.0]
-        java: [adopt@1.8, adopt@1.11, adopt@1.15]
+        scala: [2.13.6, 2.12.13, 3.0.0]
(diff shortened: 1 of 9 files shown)


aeons:
This of course depends on 0.8.0 of sbt-http4s-org being on maven central.

rossabaker:
Broken on #4828.  [Possibly relevant change](https://github.com/AdoptOpenJDK/openjdk-jdk/commit/e5866aa7560e1a6077a90a5902ec61de76922440)?

rossabaker:
This could also refresh the mergify config.

aeons:
That change is definitely it. That byte in the header is the only difference. 

However I disabled the failing test on jvm 16. Not sure why that didn't work

rossabaker:
Because that was main (now series/0.23) and this is series/0.22.

aeons:
Gah

aeons:
I rewrote the gzip test to check the inflated bytes instead of the deflated ones, so it doesn't depend on fs2 and jvm agreeing on what OS we are supposed to be running on.

PR #4873, opened 2021-05-23
Catch Exceptions in ParseResult.fromParser, and fix Content-Type so it doesn't throw

zarthross:
Fixes an issue found in https://github.com/http4s/http4s/pull/4787#issuecomment-846633626.

### DIFF
--- a/core/src/main/scala/org/http4s/MessageFailure.scala
+++ b/core/src/main/scala/org/http4s/MessageFailure.scala
@@ -80,7 +80,8 @@ object ParseResult {
 
   private[http4s] def fromParser[A](parser: Parser0[A], errorMessage: => String)(
       s: String): ParseResult[A] =
-    parser.parseAll(s).leftMap(e => ParseFailure(errorMessage, e.toString))
+    try parser.parseAll(s).leftMap(e => ParseFailure(errorMessage, e.toString))
+    catch { case p: ParseFailure => p.asLeft[A] }
 
   implicit val parseResultMonad: MonadError[ParseResult, ParseFailure] =
     catsStdInstancesForEither[ParseFailure]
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/http4s/MessageFailure.scala
```diff
@@ -80,7 +80,8 @@ object ParseResult {
 
   private[http4s] def fromParser[A](parser: Parser0[A], errorMessage: => String)(
       s: String): ParseResult[A] =
-    parser.parseAll(s).leftMap(e => ParseFailure(errorMessage, e.toString))
+    try parser.parseAll(s).leftMap(e => ParseFailure(errorMessage, e.toString))
+    catch { case p: ParseFailure => p.asLeft[A] }
```
rossabaker, line 84: I'd say this is a bug if it happens, and we might want to screech about it in the logs.  But the defensive programming is good.

core/src/main/scala/org/http4s/headers/Content-Type.scala
```diff
@@ -30,14 +30,12 @@ object `Content-Type` {
     ParseResult.fromParser(parser, "Invalid Content-Type header")(s)
 
   private[http4s] val parser: Parser[`Content-Type`] =
-    (MediaRange.parser ~ MediaRange.mediaTypeExtensionParser.rep0).map {
+    (MediaRange.parser ~ MediaRange.mediaTypeExtensionParser.rep0).flatMap {
```
rossabaker, line 33: `mapFilter` is more efficient than `flatMap` in cats-parse, but wouldn't give as nice an error message.  I think we could do it with a variant of `select` and get that `failWith` message.

rossabaker (approved):
:+1: for finding and fixing the urgent bug, but a couple improvements that I'm happy to take on in a follow-up.

zarthross:
> +1 for finding and fixing the urgent bug, but a couple improvements that I'm happy to take on in a follow-up.

:+1: I'm out of time for the day, and it'll be a couple days before I find time again, FYI.

rossabaker:
No problem.  Appreciate the efforts today.  I'll either clean this up, or we can clean it up after the fact.  I think we got enough already to support an RC that can support Rho.

PR #4876, opened 2021-05-24
Replace Jetty Scaffold with com.sun.net.httpserver

rossabaker:
Toward removing Jetty from our test dependencies so we can eventually upgrade to whichever Jetty we want.


rossabaker:
I don't think the blaze one works yet, but it seems good for the other clients.  Let's see what CI thinks.

rossabaker:
I don't want to maintain this across CE2 and CE3.  I postponed it to series/0.23.  Sorry for all the force pushes and retargets above.

rossabaker:
This branch is a mess.

Issue #2826, opened 2019-08-29
Add etag support for StaticFile.fromUrl where supported


matthughes:
`StaticFile.fromUrl` could support etags using the same mechanism as `StaticFile.fromFile` presuming the URL provided the following headers:

`Last-Modified`
`Content-Length`

While not every URL might provide those, most will.  This would add the ability to use etags when loading resources out of the classpath, which is a very common use case.

Discussed on Gitter here: https://gitter.im/http4s/http4s?at=5d682b5606e77851a313b3b0

Will try to see if it can get in 0.20 series.

ashwinbhaskar:
@matthughes  what is the expectation if the url contains the `Etag` header?

rossabaker:
I would think if it has an `ETag`, we'd just pass that through?  And calculate from these other headers if we can.  I am not an ETag expert.

qwbarch:
Hi, I'd like to try taking on this issue please!

qwbarch:
To be completely honest, I'm very new to http libraries in general. I'm looking to solidify my understanding by working through simple issues to start with.
Despite this being simply a ``good first issue``, I'm going to need a bit of help to finish this. 😅

To clarify what's expected, do I need to calculate the ``ETag`` from the ``File`` from the provided ``Url``?  
I looked at ``StaticFile#fromFile`` and used that as an example, but I'm not sure how to even test this.
Pointing me at the right direction would be great! Hopefully this doesn't come off as me needing my hand held.

I'm a bit scared but I'll push what I've tried so far 😂

rossabaker:
Please don't worry!  We're here to help, and glad for your help!  I'll go look at and comment on the other ticket.

armanbilge:
@matthughes not entirely sure, sorry that it's been neglected. If you or @qwbarch are able to refresh it and fix the merge conflicts I can take a look :) I assume you may want to re-target it at `series/0.23`.

- https://github.com/http4s/http4s/pull/4879

qwbarch:
I haven't used Scala in a long time unfortunately. Should I close my PR?

armanbilge:
No problem, if you don't mind leave your PR open and maybe someone can pick it up :)

matthughes:
Taking a look.

matthughes:
@armanbilge I posted a new PR that brings @qwbarch's work up to date with main over at #6617.

matthughes:
Posted one for 23 branch over at #6619

PR #4879, opened 2021-05-25
Add etag support for StaticFile.fromUrl where supported

qwbarch:
Issue #2826.
This is a draft for now until it's ready since I still need help working it out.


core/src/main/scala/org/http4s/StaticFile.scala
```diff
@@ -77,7 +77,12 @@ object StaticFile {
...
 
-  def fromURL[F[_]](url: URL, req: Option[Request[F]] = None)(implicit
+  def fromURL[F[_]: Sync: Files](
+      url: URL,
+      req: Option[Request[F]] = None): OptionT[F, Response[F]] =
+    fromURL(url, req, calcETag[F])
+
+  def fromURL[F[_]](url: URL, req: Option[Request[F]], etagCalculator: File => F[String])(implicit
       F: Sync[F]): OptionT[F, Response[F]] = {
     val fileUrl = url.getFile()
```
rossabaker, line 87: This line seems buggy to me:

> Gets the file name of this URL. The returned file portion will be the same as getPath(), plus the concatenation of the value of getQuery(), if any. If there is no query portion, this method and getPath() will return identical results.

If the URL is `https://www.example.com/blah`, the result will be `/blah`, and looked up on the local file system.  I think we can only use `file` if the URL scheme is `file://`.

core/src/main/scala/org/http4s/StaticFile.scala
```diff
@@ -92,28 +97,30 @@ object StaticFile {
...
-                case other => throw other
-              },
-              f = { inputStream =>
-                Some(
-                  Response(
-                    headers = headers,
-                    body = readInputStream[F](F.pure(inputStream), DefaultBufferSize)
-                  ))
-              }
+          etagCalculator(file).map(ETag(_)).flatMap { etag =>
```
rossabaker, line 100: So if `file` is suspect (see comment above), then the `etagCalculator` that takes a `File` is not the right tool for this method.  The etag calculator in this method should probably be `URL => F[ETag]`.  A default implementation could get lastmodified and length from the `urlConn` (examples are already in this function) and then operate similarly to the file-based `calcETag`.  That default implementation should also have a `Sync` constraint on `F`... which will also complicate passing it as a default argument.

I'll leave that as an exercise (it's late here), but if that doesn't make sense, don't spin your wheels too hard on it: someone will be awake soon to give more suggestions. :smile:

rossabaker (commented):
This is a really good start, but I see a potential problem that relates to a pre-existing problem.

core/src/main/scala/org/http4s/StaticFile.scala
```diff
@@ -129,6 +136,11 @@ object StaticFile {
         .map(isFile =>
           if (isFile) s"${f.lastModified().toHexString}-${f.length().toHexString}" else "")
 
+  def calcETagURL[F[_]: Sync]: URL => F[ETag] = url => {
+    val urlConn = url.openConnection
```
qwbarch, line 140: Should I be ``F.delay``ing this?  
I only didn't suspend it within an effect because the example on line 93 doesn't do it.  
Isn't ``URL#openConnection`` side-effecting?

Also, would it make more sense to have the signature as ``URLConnection => F[ETag]`` to reuse the same connection?

qwbarch:
Since I added a new method ``calcETagURL``, would it be okay to name the old ``calcETag`` to ``calcETagFile`` to be more specific?

core/src/main/scala/org/http4s/StaticFile.scala
qwbarch: Changing it to default arguments would be for Issue #2072 right?  
Should I change ``StaticFile`` to an algebra as well, and then add in default parameters rather than overloading like I just did?

core/src/main/scala/org/http4s/StaticFile.scala
hamnis: this is blocking, so wrapping in `Sync[F].blocking{}` would make perfect sense

qwbarch: Nevermind, I took a look at the docs again and read it more carefully.
Looks like ``URLConnection`` represents a connection that doesn't happen yet.  
Line 109 confirms it for me since ``URLConnection#getInputStream`` is what reads the data.
Now I understand why an ``F.delay`` wasn't needed. It's already a pure value 😂

hamnis: it would be pure unless you where reading headers. That makes it blocking

qwbarch: Makes a lot more sense, thanks @hamnis!

hamnis: Maybe this should be private api, since we assume that we will get LastModified And ContentLength?
I assume that this is either a file:// or jar reference?

maybe also this should use `getContentLengthLong` as well.

qwbarch: Oh whoops I had it as ``getContentLengthLong`` earlier and forgot to keep it as that when I added those changes.
Good point on making it private though. I also forgot that I'm assuming the url also contains a file from where the ``etagCalculator`` is being called

core/src/main/scala/org/http4s/StaticFile.scala
hamnis: that means that it also a security issue.

qwbarch:
@rossabaker @hamnis Hey if you guys have the time, do you mind reviewing this PR?
I'd like to work on issue #2072 afterwards, but it looks like I'd have to wait for this PR to work on that one.
If you guys don't have time it's okay though, I understand maintainers lots of responsibilities to take care of

core/src/main/scala/org/http4s/StaticFile.scala
```diff
@@ -77,7 +77,12 @@ object StaticFile {
         .flatMap(fromURL(_, req)))
   }
 
-  def fromURL[F[_]](url: URL, req: Option[Request[F]] = None)(implicit
+  def fromURL[F[_]: Sync: Files](
```
hamnis, line 80: this `Files` constraint is not used, so it stopped compiling in CI
then since that is not used, perhaps we need to remove that in fromResource as well.
it should probably also be rebased on `main` or `series/0.22`

hamnis: https://github.com/http4s/http4s/pull/4879/checks?check_run_id=2680925168#step:8:96
and
https://github.com/http4s/http4s/pull/4879/checks?check_run_id=2680925109#step:8:24

qwbarch: Gotcha, thanks hamnis! I'll try and fix it in a couple hours or so 😄

PR #4880, opened 2021-05-25
Add try/catch arounds reset timeout so we dont blow up on shutdown

hamnis:
Can be enhanced by including: https://github.com/http4s/blaze/pull/535

### DIFF
--- a/blaze-core/src/main/scala/org/http4s/blazecore/IdleTimeoutStage.scala
+++ b/blaze-core/src/main/scala/org/http4s/blazecore/IdleTimeoutStage.scala
@@ -79,13 +79,15 @@ final private[http4s] class IdleTimeoutStage[A](
   }
 
   private def resetTimeout(): Unit =
-    try setAndCancel(exec.schedule(killSwitch, ec, timeout))
-    catch {
-      case r: RuntimeException if r.getMessage == "TickWheelExecutor is shutdown" =>
-        logger.warn(s"Resetting timeout after tickwheelexecutor is shutdown")
-        cancelTimeout()
-      case NonFatal(e) => throw e
-    }
+    if (exec.isAlive) {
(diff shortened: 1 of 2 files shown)


blaze-core/src/main/scala/org/http4s/blazecore/IdleTimeoutStage.scala
```diff
@@ -83,7 +84,7 @@ final private[http4s] class IdleTimeoutStage[A](
       case r: RuntimeException if r.getMessage == "TickWheelExecutor is shutdown" =>
```
rossabaker, line 84: When blaze-0.14.7 is out, we can do better here with the new Exception type.

blaze-core/src/main/scala/org/http4s/blazecore/IdleTimeoutStage.scala
```diff
@@ -79,13 +79,15 @@ final private[http4s] class IdleTimeoutStage[A](
...
-      case r: RuntimeException if r.getMessage == "TickWheelExecutor is shutdown" =>
-        logger.warn(s"Resetting timeout after tickwheelexecutor is shutdown")
-        cancelTimeout()
-      case NonFatal(e) => throw e
-    }
+    if (exec.isAlive) {
+      try setAndCancel(exec.schedule(killSwitch, ec, timeout))
+      catch {
+        case TickWheelExecutor.AlreadyShutdownException =>
+          logger.warn(s"Resetting timeout after tickwheelexecutor is shutdown")
```
hamnis, line 86: drop log?

rossabaker: How frequent is the error still? It still indicates something is broken. It seems like a good warning to me.

ChristopherDavenport (approved):
Can always lower log level/remove if its too loud/noisy.

PR #4884, opened 2021-05-25
Replace Defer[F].defer with Monad.unit.flatMap

### LINKED ISSUE #4883: Replace Defer constraints with Concurrent
In 1.0.x, `AuthedRoutes` currently requires an instance of `Defer[F]`. This causes ambiguous implicits since `Concurrent[F]` is typically in scope and as of CE3, `Concurrent` is unrelated to `Defer`.

Per @ChristopherDavenport, these `Defer` usages should be replaced with `Concurrent.unit.flatMap(_ => ...)`. We should check for all such `Defer` constraints, not just `AuthedRoutes`.

hamnis:
This is done to avoid Diverging implicits when we only have a Concurrent
constraint.

Fixes #4883

### DIFF
--- a/core/src/main/scala/org/http4s/AuthedRoutes.scala
+++ b/core/src/main/scala/org/http4s/AuthedRoutes.scala
@@ -16,7 +16,7 @@
 
 package org.http4s
 
-import cats.{Applicative, Defer}
+import cats.{Applicative, Monad}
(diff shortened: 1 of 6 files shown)


hamnis:
this would indeed be a good fit for 0.23, can retarget

rossabaker:
Picked onto 0.23.

PR #4888, opened 2021-05-26
added support for comments in ServerSentEvent

fizzy33:
A few notes.  

* unit tests had dropLast added because fs2.text.lines sees "\n" as two lines so decode sees it as two events when in fact it is one.  There is an fs2.text conversation happening now over an alternative linesWithEndings that would provide a cleaner path
* note one alternative to dropLast is we always add an extra empty event (maybe cleaner and safer since we could be dropping a non-empty buggy event)
* @rossabaker I will save the cleaner version of decode for now, it just didn't perform as well and I ran out of time for today as discussed the big thing is getting ServerSentEvents with the correct fields
* slight tweak to use an internal LineBuffer performance is on par with StringBuilder albeit a bit more memory consumed again could tweak this with something that accumulates in a StringBuilder and notes if any lines at all have been added (performance is from very informal testing)
* noting in real world I doubt the performance of this will be a gating factor

### DIFF
--- a/core/src/main/scala/org/http4s/ServerSentEvent.scala
+++ b/core/src/main/scala/org/http4s/ServerSentEvent.scala
@@ -29,11 +29,11 @@ final case class ServerSentEvent(
     eventType: Option[String] = None,
     id: Option[EventId] = None,
     retry: Option[Long] = None,
-    comments: Option[String] = None,
+    comment: Option[String] = None,
 ) extends Renderable {
   def render(writer: Writer): writer.type = {
     data.foreach (writer << "data: " << _ << "\n")
-    comments.foreach (writer << ": " << _ << "\n")
+    comment.foreach (writer << ": " << _ << "\n")
     eventType.foreach(writer << "event: " << _ << "\n")
     id match {
       case None =>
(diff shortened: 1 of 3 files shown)


core/src/main/scala/org/http4s/ServerSentEvent.scala
```diff
@@ -24,13 +25,15 @@ import org.http4s.util.{Renderable, Writer}
 import scala.util.Try
 
 final case class ServerSentEvent(
-    data: String,
+    data: Option[String] = None,
     eventType: Option[String] = None,
     id: Option[EventId] = None,
-    retry: Option[Long] = None
+    retry: Option[Long] = None,
```
rossabaker, line 31: Do you want the `FiniteDuration` before 0.22 ABI gets locked?

rossabaker (approved):
Looks like progress.  I'll cherry-pick this to 0.22 so it's in that RC today.

core/src/main/scala/org/http4s/ServerSentEvent.scala
```diff
@@ -24,13 +25,15 @@ import org.http4s.util.{Renderable, Writer}
...
+    comments: Option[String] = None,
```
rossabaker, line 32: Bikeshed: the plural name is weird to me.  It's just one comment, that might be split across multiple lines.

(`data` is wired into the spec, so I won't be a nerd and try to call that `datum`.)

fizzy33: "comment" has been pushed

core/src/main/scala/org/http4s/ServerSentEvent.scala
fizzy33: why yes I do thanks for remembering

fizzy33: pushed

rossabaker (approved):

rossabaker:
Picked to 0.22.

PR #4891, opened 2021-05-26
Avoid sleeping 1 second in tests which don't use the /delay path.

RafalSumislawski:
The `GetRoutes.getPaths` used to block for 1 second due to `unsafeRunSync` waiting for `timer.sleep(1.second)`. This PR fixes it which should make some tests faster. The difference is most noticeable in `BlazeClient213Suite`.


rossabaker (approved):
:+1:

rossabaker:
I thought this looked familiar.  Something like it was on 0.23, but we just hadn't brought it back to CE2.  Thanks!

PR #4898, opened 2021-05-27
Clean out deprecatia prior to 0.20

rossabaker:
Could be more aggressive still, but these are all two years old.


hamnis:
Lovely, I have wanted to get rid of a few of these for a while now :)

rossabaker:
If there are any that are claiming valuable real estate, we could squeeze those into 0.22, but I'm trying to minimize breaking changes.  And some from 0.20 era could almost certainly go now.  But we've had users upgrade from Way Back, and they've appreciated our deprecations.

I'll stop clinging to my deprecatia when we start getting better at scalafix. :smile:

rossabaker:
Test failure recorded as #4899.  The other one looks like an infra flake.

aeons (approved):

rossabaker:
Note: this is the first significant divergence between 0.23 and 1.0.

PR #4901, opened 2021-05-27
Fix compilation of docs module

danicheg:
The motivation for this PR was to fix docs generation via `mdoc`.
Also in this PR I've fixed some obvious things about migration to the `cats-effect` 3.0. 
Though the docs need more work on, at least now it's compile :)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -550,7 +550,7 @@ lazy val docs = http4sProject("docs")
       )
     }
   )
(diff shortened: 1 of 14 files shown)


danicheg:
Have no idea how I missed the #4894 :)

rossabaker:
Actually, parts of this are better.  I didn't know about that `global` import you used, which is more concise than what I did.  I may go back and adopt some of this. :smile:

PR #4902, opened 2021-05-27
Clean-up docs from cats-effect 2.0 things

danicheg:
Some docs clean-up.


rossabaker (approved):
Thanks!  I just ported the docs last night, but haven't done a thorough reread, so these fixes are very much appreciated!

PR #4905, opened 2021-05-28
Add ability to customize Prometheus metrics' names

danicheg:
In this PR have added the ability to customize Prometheus metrics' names.

Motivation
----------
Currently, it's possible only to specify a prefix for Prometheus metrics' names, the suffixes are hardcoded. 
It's more flexible to have the ability to specify suffixes too. Some valuable points of this:
* Some systems may have their own guidelines about metrics' naming. 
* For a microservices ecosystem with different backends will help unify metrics' names (to smoothly integrate metrics in a monitoring systems e. g.).

Using
-----
* Customizing metrics' names for server metrics:
    ```scala
    import cats.effect.{IO,Resource}
    import io.prometheus.client.CollectorRegistry
    import org.http4s._
    import org.http4s.dsl.io._
    import org.http4s.HttpRoutes
    import org.http4s.metrics.prometheus.{Prometheus,PrometheusMetricsNames,PrometheusMetricsSettings}
    import org.http4s.server.middleware.Metrics
    import org.http4s.server.Router
    
    def meteredHttpRoutes(collectorRegistry: CollectorRegistry): Resource[IO, HttpRoutes[IO]] =
        for {
          metricsNames <- Resource.eval(IO.fromEither(
            PrometheusMetricsNames(
              responseDuration = "server_api_v1_response_latency_seconds",
              activeRequests = "server_api_v1_active_request_total",
              requests = "server_api_v1_request_total",
              abnormalTerminations = "server_api_v1_failed_requests"
            )
          ))
          settings = PrometheusMetricsSettings.DefaultSettings.withMetricsNames(metricsNames)
          metricsOps <- Prometheus.metricsOps[IO](collectorRegistry, settings)
          apiService = HttpRoutes.of[IO] {
            case GET -> Root / "api" / "v1" =>
              Ok()
          }
          routes = Metrics[IO](metricsOps)(apiService)
          router = Router[IO]("/api/v1" -> routes)
        } yield router
    ```
* Customizing suffixes for a use case with metrics' prefix:
    ```scala
    import cats.effect.{IO,Resource}
    import io.prometheus.client.CollectorRegistry
    import org.http4s._
    import org.http4s.dsl.io._
    import org.http4s.HttpRoutes
    import org.http4s.metrics.prometheus.{Prometheus,PrometheusMetricsNames}
    import org.http4s.server.middleware.Metrics
    import org.http4s.server.Router
    
    def meteredHttpRoutes(collectorRegistry: CollectorRegistry): Resource[IO, HttpRoutes[IO]] =
        for {
          metricsNames <- Resource.eval(IO.fromEither(
            PrometheusMetricsNames(
              responseDuration = "response_latency_seconds",
              activeRequests = "active_request_total",
              requests = "request_total",
              abnormalTerminations = "failed_requests"
            )
          ))
          metricsOps <- Prometheus.metricsOps[IO](collectorRegistry, prefix = "server_api_v1", metricsNames = metricsNames)
          apiService = HttpRoutes.of[IO] {
            case GET -> Root / "api" / "v1" =>
              Ok()
          }
          routes = Metrics[IO](metricsOps)(apiService)
          router = Router[IO]("/api/v1" -> routes)
        } yield router
    ```

I would appreciate any feedback.


prometheus-metrics/src/main/scala/org/http4s/metrics/prometheus/PrometheusMetricsNames.scala
```diff
@@ -0,0 +1,135 @@
...
+import org.http4s.metrics.prometheus.PrometheusMetricsNames._
+
+/** Metric name may contain ASCII letters and digits, as well as underscores and colons.
+  * It must match the regex ([a-zA-Z_:][a-zA-Z0-9_:]*).
+  * Notice that only a single metric with a given name can be registered within a [[io.prometheus.client.CollectorRegistry]].
+  * Vice versa registering a metric with the same name more than once will yield an error.
+  *
+  * More details: https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels
+  */
+final class PrometheusMetricsNames private (
```
isomarcte, line 30: I think this might be a very good use case for https://github.com/typelevel/literally which we already have as a dependency. Using `literally` would lift the vast majority of metric naming errors into a compile time failure rather than a runtime `IllegalArgumentException`, which would be good as many tests setups don't invoke metrics during testing.

It also might be useful to make a single newtype for a metric name, similar to what https://github.com/davenverse/epimetheus does. This would be where `literally` would come into play.

prometheus-metrics/src/main/scala/org/http4s/metrics/prometheus/PrometheusMetricsSettings.scala
```diff
@@ -0,0 +1,44 @@
...
+
+  def withResponseDurationSecondsHistogramBuckets(
+      responseDurationSecondsHistogramBuckets: NonEmptyList[Double]
+  ): PrometheusMetricsSettings =
+    copy(responseDurationSecondsHistogramBuckets = responseDurationSecondsHistogramBuckets)
+}
+
+object PrometheusMetricsSettings {
+  // https://github.com/prometheus/client_java/blob/parent-0.6.0/simpleclient/src/main/java/io/prometheus/client/Histogram.java#L73
+  private val DefaultHistogramBuckets: NonEmptyList[Double] =
```
isomarcte, line 36: It _might_ be useful to have this public. A common use case is merely adding more buckets, or scaling all buckets by a constant scalar.

If we made it public, I'd also make this a `NonEmptySet`.

prometheus-metrics/src/main/scala/org/http4s/metrics/prometheus/Prometheus.scala
```diff
@@ -78,18 +97,44 @@ object Prometheus {
...
-    * *
-    * * @param registry a metrics collector registry
-    * * @param prefix a prefix that will be added to all metrics
+  /** Creates a [[MetricsOps]] that supports Prometheus metrics
+    *
+    * @param registry a metrics collector registry
+    * @param prefix a prefix that will be added to all metrics
+    * @param metricsNames names of the metrics that will be used as a suffix for the metrics
     */
   def metricsOps[F[_]: Sync](
```
isomarcte, line 106: Could we add a new method which takes the `metricNames` parameter and then have this method's signature remain unchanged? Something like,

```scala
def metricsOps[F[_]: Sync](
      registry: CollectorRegistry,
      prefix: String = "org_http4s_server",
      responseDurationSecondsHistogramBuckets: NonEmptyList[Double] =
        DefaultSettings.responseDurationSecondsHistogramBuckets
  ): Resource[F, MetricsOps[F]] =
    metricsOps[F](registry, prefix, responseDurationSecondsHistogramBuckets, DefaultSettings.metricsNames)
```

This would avoid a binary breaking change, which is always nice, even on a pre-releas.

isomarcte, line 106: NIT: I'd personally avoid the overloaded method name here. I realize it is more verbose to have something like,

```scala
def metricsOpsWithNames[F[_]: Sync](...)
```

but I think it is more clear to the users.

prometheus-metrics/src/main/scala/org/http4s/metrics/prometheus/PrometheusMetricsSettings.scala
danicheg: It could be used via `PrometheusMetricsSettings.DefaultSettings`, but yeah, it makes sense to have it public.

prometheus-metrics/src/main/scala/org/http4s/metrics/prometheus/Prometheus.scala
danicheg: Good idea, gonna change it.

danicheg: I've chosen the same name just for avoiding verbose name. But I'm also not sure, maybe verbosity here not so bad... We need one more opinion on this, I suppose.

prometheus-metrics/src/main/scala/org/http4s/metrics/prometheus/PrometheusMetricsNames.scala
danicheg: From my experience, it's more flexible to have metrics' names (and other configurable stuff) in config. And therefore I've chosen here to check in runtime, not in compile. The main idea is that config loads at the startup of the application and if some metrics' names invalid - the application could decide to keep on startup or not.

prometheus-metrics/src/main/scala/org/http4s/metrics/prometheus/Prometheus.scala
danicheg: Uh, I've remembered one use case - it's might be useful to have a method that specifies prefix of metrics' names and provides custom suffixes (that could be created once for a bunch of endpoints and reused).

danicheg:
@isomarcte thanks for the review! It would be great if you'll check again.

prometheus-metrics/src/main/scala/org/http4s/metrics/prometheus/PrometheusMetricsNames.scala
isomarcte: @danicheg using literally doesn't preclude runtime checking, it just permits compile time checking.

prometheus-metrics/src/main/scala/org/http4s/metrics/prometheus/Prometheus.scala
isomarcte: > Uh, I've remembered one use case - it's might be useful to have a method that specifies prefix of metrics' names and provides custom suffixes (that could be created once for a bunch of endpoints and reused).

I'm sorry, I don't quite follow how this relates to keeping the existing signature unchanged. Could you elaborate?

armanbilge:
@danicheg what do you think about landing this one? Seems like extensive changes. Currently this PR blocks schisming of this module to its own repo.

I couldn't tell if it's a breaking change or not, but after it schisms one options is to start a breaking prometheus metric 0.24.x series for http4s-core 0.23.x. WDYT, thanks!

danicheg:
This PR is pretty outdated. So the module can be safely transferred out.

PR #4915, opened 2021-06-05
Add file-based multipart decoder with better resource handling

### LINKED ISSUE #2627: Add a better implementation of a file-backed multipart decoder

nigredo-tori:
Fixes #2627.

The behavior and the interface should mostly match the existing mixed multipart parser. The difference is twofold:
1. The main point is that, instead of deleting files manually, this spawns a background fiber with the resource's lifecycle (as a `Resource.use`). Said fiber is managed by the provided `Supervisor`. When the `Supervisor` is closed, the fiber is canceled, at which point the file is deleted. See `superviseResource` for the implementation. This setup is not perfect (e.g. if the main stream fails, the fibers are not released, and the resources are not released until the supervisor is released), but it should be enough for the intended scenario (one-use `EntityDecoder`s).
2. A minor improvement is the addition of a separate parameter controlling chunk size when reading the file. In the existing parser the chunk size matches `maxSizeBeforeWrite`, which means the implementation creates 50MB arrays when reading the cached data.
    Admittedly, I didn't put much thought into the default chunk size (8192 bytes), so if there is a better default, I'd be happy to change the code.

Additional changes:
1. I've changed existing multipart tests a bit (wrapping parameters in `Resource`), so that I could reuse them for the new implementation.
2. I've factored out a common `EntityDecoder` implementation in `MultipartDecoder`.
3. I've added missing ScalaDoc to `EntityDecoder.mixedMultipart` (copied and pasted from the corresponding `MultipartDecoder` function).
4. ~~I've taken the liberty to deprecate the existing (unsafe) mixed multipart decoder.~~ CI sees the tests (!) using deprecated functions, and cries foul. Reverted this for now.

### DIFF
--- a/core/src/main/scala/org/http4s/multipart/MultipartParser.scala
+++ b/core/src/main/scala/org/http4s/multipart/MultipartParser.scala
@@ -25,6 +25,8 @@ import java.nio.file.{Path, StandardOpenOption}
 import org.typelevel.ci.CIString
 import fs2.RaiseThrowable
 import org.http4s.internal.bug
+import cats.effect.std.Supervisor
+import cats.effect.Resource
 
 /** A low-level multipart-parsing pipe.  Most end users will prefer EntityDecoder[Multipart]. */
 object MultipartParser {
@@ -46,7 +48,7 @@ object MultipartParser {
   private[this] sealed trait Event
   private[this] final case class PartStart(value: Headers) extends Event
   private[this] final case class PartChunk(value: Chunk[Byte]) extends Event
-  private[this] final case object PartEnd extends Event
+  private[this] case object PartEnd extends Event
 
   def parseStreamed[F[_]: Concurrent](
       boundary: Boundary,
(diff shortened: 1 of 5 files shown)


core/src/main/scala/org/http4s/multipart/MultipartParser.scala
```diff
@@ -591,6 +593,115 @@ object MultipartParser {
     go(stream, Stream.empty, 0)
   }
 
+  /////////////////////////////////////
+  // Resource-safe file-based parser //
+  /////////////////////////////////////
+
+  /** Like parseStreamedFile, but the produced parts' resources are managed by the supervisor.
+    */
+  def parseSupervisedFile[F[_]: Concurrent: Files](
```
rossabaker, line 602: I think we could make these `private[multipart]` to give us more bincompat wiggle room in the future?

rossabaker (commented):
Sorry for the slow review.  I wanted to get 0.22 out the door to focus on the 0.23, but this is a very welcome addition!

I think we could deprecate those old functions by either adding a nowarn annotation to the test, or just deleting the tests for the deprecated functions.  I used to just delete those tests, but that was before we had `@nowarn` at our disposal.

nigredo-tori:
> I think we could deprecate those old functions by either adding a nowarn annotation to the test, or just deleting the tests for the deprecated functions.

I've added `@nowarn`.

core/src/main/scala/org/http4s/multipart/MultipartParser.scala
nigredo-tori: Done.

rossabaker:
Not sure what went wrong on 2.13.  We've got examples of that annotation elsewhere.  I'll take a look unless you beat me to it -- I'd like to get 0.23.0 out the door later today, including this.

nigredo-tori:
> Not sure what went wrong on 2.13.

With 2.12 I had to have a `val` to add `@nowarn` to. But in one context that meant that a member of type `Unit` was added to the class, and 2.13 didn't like that. So I had to add a workaround to my workaround. Let's hope that helps.

rossabaker:
I guess I haven't done it on a test suite yet.  That's probably the novel thing here.

rossabaker:
Alright, there may be a better way, but 4180bfa works.  I'm shipping it.

Issue #4920, opened 2021-06-09
org.http4s.MessageFailure does not return detailed error message


brbrown25:
While using Rho I uncovered a case where json parsing errors are not returned in the message response for all http errors. If there is interest, I'd love to contribute a fix for this!

rossabaker:
The `toHttpResponse` is terse for a reason: anything that contains any information from the request can be used in an XSS attack.  (This design was inherited from Spray, a very long time ago.)  It's a paranoid setting: the risk is much lower when the response is plain text, and it's difficult to exploit.  But we want to err on the side of security by default.

Wanting more verbose errors in the response instead of the logs is not unreasonable.  Maybe we should consider a middleware that intercepts these messages and generates something a bit friendlier?

brbrown25:
@rossabaker thanks for looking and giving some feedback!
> The `toHttpResponse` is terse for a reason: anything that contains any information from the request can be used in an XSS attack. (This design was inherited from Spray, a very long time ago.) It's a paranoid setting: the risk is much lower when the response is plain text, and it's difficult to exploit. But we want to err on the side of security by default.
That makes a lot of sense, thanks for explaining that!
> Wanting more verbose errors in the response instead of the logs is not unreasonable. Maybe we should consider a middleware that intercepts these messages and generates something a bit friendlier?
I agree some sort of middleware would definitely be nice. The current workaround I'm using involves adding `.withServiceErrorHandler` then adding a message body to the `NonFatal` case and throwing an exception with a detailed json parsing error message. I'm not really sure where this interceptor should live so if you give me some pointers I can work on adjusting my current pr to reflect the proposed design. I'm guessing it would get enacted somewhere around here https://github.com/http4s/http4s/blob/main/blaze-server/src/main/scala/org/http4s/blaze/server/Http2NodeStage.scala#L235 ?

rossabaker:
Sorry for the slow reply.  I changed jobs and went on vacation, and things got slow.

Yes, that's where exceptions that aren't handled get converted into a message by that particular backend.  (Similar calls exist in all the backends.)  The default service error handler is [here](https://github.com/http4s/http4s/blob/7c69a4e63973caca0f1181f069d80117431ed1bf/server/src/main/scala/org/http4s/server/package.scala#L163-L183).  It should log the entire message, with underlying exception, and then invoke `.toHttpResponse` for the more paranoidly secure response over the network.

[...]

PR #4921, opened 2021-06-10
bug(MessageFailure): updating Malformed and Invalid message body fail…

brbrown25:
…ures to carry the actual failure through in the response body. #4920.


rossabaker:
Thanks for opening this.

The `toHttpResponse` is terse for a reason: anything that contains any information from the request can be used in an XSS attack.  (This design was inherited from Spray, a very long time ago.)  It's a paranoid setting: the risk is much lower when the response is plain text, and it's difficult to exploit.  But we want to err on the side of security by default.

Wanting more verbose errors in the response instead of the logs is not unreasonable.  Maybe we should consider a middleware that intercepts these messages and generates something a bit friendlier?

PR #4923, opened 2021-06-12
Add infix operator for Scala 3

qwbarch:
When using Scala 3 with ``-source:future`` compiler flag enabled, the compiler complains about a missing infix operator.

For example:
```scala
val authedRoutes: AuthedRoutes[User, IO] =
  AuthedRoutes.of {
    case GET -> Root / "welcome" as user => Ok(s"Welcome, ${user.name}")
  }
```
The compiler will complain:
```
Alphanumeric extractor as is not declared `infix`; it should not be used as infix operator.
The operation can be rewritten automatically to `as` under -deprecation -rewrite.
Or rewrite to prefix syntax as(...) manually.
```

This PR simply adds ``infix`` for the unapply method of ``as`` for Scala 3, while keeping the exact same source code for Scala 2.


rossabaker:
:+1: I didn't know about this rule.  I think it's relevant back to 0.22.  I'll cherry-pick it.

PR #4931, opened 2021-06-19
Fix "Insert a User-Agent header" test

valencik:
I *think* this PR fixes a flaky User-Agent test.

Prior to these changes this test repeatedly failed on my system.
Weirdly, metals couldn't find the definition of `BuildInfo`, and `BuildInfo.version` was also 0.15.1 in scope of this test.
(That happens to be the [blaze version number](https://github.com/http4s/http4s/blob/83df69ae4ba56d1062bff9e55722cbc9eb0bc478/project/Http4sPlugin.scala#L286), but I can't figure that out.)

Adding the import meant the test passed each time I ran it.

I've removed the `.flaky`, perhaps that is over confident...


ChristopherDavenport (approved):

rossabaker (approved):
I don't see how the change would fix an _intermittent_ failure, but given empircal evidence, I'm happy to try!

PR #4933, opened 2021-06-20
Lazy header parse

RaasAhsan:
Ran into this when while trying to debug why a header parser was throwing an exception on a header it wasn't even intended for. The way it's coded right now, if we try to parse a header from its raw form, it'll parse every preceding header searched as well, even if the name doesn't match. This is definitely a performance hit when headers are queried

### DIFF
--- a/core/src/main/scala/org/http4s/Header.scala
+++ b/core/src/main/scala/org/http4s/Header.scala
@@ -182,7 +182,7 @@ object Header {
     type Aux[A, G[_]] = Select[A] { type F[B] = G[B] }
 
     def fromRaw[A](h: Header.Raw)(implicit ev: Header[A, _]): Option[Ior[ParseFailure, A]] =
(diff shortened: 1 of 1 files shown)


core/src/main/scala/org/http4s/Header.scala
```diff
@@ -182,7 +182,7 @@ object Header {
     type Aux[A, G[_]] = Select[A] { type F[B] = G[B] }
 
     def fromRaw[A](h: Header.Raw)(implicit ev: Header[A, _]): Option[Ior[ParseFailure, A]] =
-      (h.name == Header[A].name).guard[Option].as(Header[A].parse(h.value).toIor)
+      (h.name == Header[A].name).guard[Option].map(_ => Header[A].parse(h.value).toIor)
```
RaasAhsan, line 185: this was the quickest change, but i wouldn't be against rewriting this to be an if statement so the laziness is more apparent. or we can add a comment

rossabaker (approved):
Eek.  I think of these as interchangeable.

PR #4937, opened 2021-06-23
Set version to 3.x for http4s v >= 0.23

danielkarch:


ChristopherDavenport (approved):

rossabaker (changes requested):
This column is for Cats, which is still 2.x.  Should we add a Cats-Effect column?

Note that in all modern versions, the Cats and Cats-Effect versions are implied by the FS2 version, though it doesn't hurt to be clear about the mapping for people new to the ecosystem.

armanbilge:
The site has been re-organized a couple times, apologies, and this PR needed changes anyway. Feel free to re-open.

PR #4941, opened 2021-06-27
Core - Change Message Body definition

diesalbla:
Currently, the `Message` classes (and their subclasses `Request` and `Response`) are parameterised on an `F[_]` type, of kind `Type => Type`. The reason why these classes need a type parameter of this (higher) kind is because the entity body of requests and responses are defined as streams of bytes, `Stream[F, Byte]`, using the `Stream` class from the `fs2` library. 

IIUC, the main reason for defining  why the "Entity body" is defined as a bytestream is as
- It is a natural representation for input/output from the running clients and servers.
- It allows us to represent potentially large objects (entities) that are coming "through the wire" incrementally, which is common if transmitting video files or large PDFs...
- It serves as a common target for converting different types to/from, via entity encoders and decoders.

However, this has some downsides:
- The `Stream` data type is an intrinsically opaque type, which can represent unbounded entity bodies of unknown size. However, it could be preferable to work with finite representations (like just _arrays_ of bytes), empty representations, or specific data types. 
- It forces in the clients and servers an intermediate conversions to/from streams of bytes.
- In the definition of routes, it ties the effect on which routes run with the type 
- It exposes a low-level concern, the effect on which the clients or servers are running, into the representation of the core data structures for requests and responses. From there, it propagates through to the entity encoders/decoders, the middlewares, etc.

This Draft Pull Requests starts to consider a different alternative: decouple the core definitions of Http requests and responses from the Stream datatype from fs2, remove the `F[_]` parameter, and add a type parameter `Body` for the whole type of the body.

- Requests and responses that carry no content can be represented using the type binding `Body = Unit`. For instance, we can define a plain `NotFound` response as a `Response[Unit]`, with no `Applicative` needed. 
- Entity encoders and decoders become parameterised on the type of the `Body`, but they no longer need the HKT type parameter `F[_]`. In other words, both encoders and decoders can be defined as pure functions, without necessarilly involving any effects.  s a consequence, the `DecodeResult` is no longer an `EitherT`, but just a plain `Either`. 
- The `ContextRequest` (and its alias `AuthedRequest`) are also decoupled from the `F[_]` type.

### DIFF
--- a/core/shared/src/main/scala/org/http4s/AuthedRequest.scala
+++ b/core/shared/src/main/scala/org/http4s/AuthedRequest.scala
@@ -20,10 +20,10 @@ import cats.Functor
 import cats.data.Kleisli
 
 object AuthedRequest {
-  def apply[F[_]: Functor, T](
-      getUser: Request[F] => F[T]): Kleisli[F, Request[F], AuthedRequest[F, T]] =
-    ContextRequest[F, T](getUser)
+  def apply[F[_]: Functor, Body, T](
+      getUser: Request[Body] => F[T]): Kleisli[F, Request[Body], AuthedRequest[Body, T]] =
+    ContextRequest[F, T, Body](getUser)
 
-  def apply[F[_], T](context: T, req: Request[F]): AuthedRequest[F, T] =
-    ContextRequest[F, T](context, req)
+  def apply[Body, T](context: T, req: Request[Body]): AuthedRequest[Body, T] =
+    ContextRequest[Body, T](context, req)
 }
--- a/core/shared/src/main/scala/org/http4s/AuthedRoutes.scala
+++ b/core/shared/src/main/scala/org/http4s/AuthedRoutes.scala
@@ -31,8 +31,10 @@ object AuthedRoutes {
     * @param run the function to lift
     * @return an [[AuthedRoutes]] that wraps `run`
     */
-  def apply[T, F[_]](run: AuthedRequest[F, T] => OptionT[F, Response[F]])(implicit
-      F: Monad[F]): AuthedRoutes[T, F] =
+  def apply[T, F[_]](
+    run: AuthedRequest[EntityBody[F], T] => OptionT[F, Response[EntityBody[F]]]
+  )(implicit F: Monad[F]
+  ): AuthedRoutes[T, F] =
     Kleisli(req => OptionT(F.unit >> run(req).value))
 
   /** Lifts a partial function into an [[AuthedRoutes]].  The application of the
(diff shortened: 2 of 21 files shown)


RaasAhsan:
Chris started work on an alternative model where the HTTP entity as encoded as an ADT: https://github.com/http4s/http4s/pull/4509 . There were some questions about being able to extend the model later (e.g. supporting zero-copy file transfers later), would be interesting to see if this approach solves any of those challenges

core/src/main/scala/org/http4s/MessageFailure.scala
```diff
@@ -56,9 +56,9 @@ final case class ParseFailure(sanitized: String, details: String)
 
   def cause: Option[Throwable] = None
 
-  def toHttpResponse[F[_]](httpVersion: HttpVersion): Response[F] =
+  def toHttpResponse[Body](httpVersion: HttpVersion): Response[String] =
```
nafg, line 59: Is the `Body` type parameter doing anything?

core/src/main/scala/org/http4s/Media.scala
```diff
@@ -17,25 +17,30 @@
...
 
 import cats.MonadThrow
-import fs2.{RaiseThrowable, Stream}
-import fs2.text.utf8Decode
 import org.http4s.headers._
 
-trait Media[F[_]] {
-  def body: EntityBody[F]
+trait Media[+ Body] {
+  def body: B
```
nafg, line 23:
```suggestion
  def body: Body
```

core/src/main/scala/org/http4s/EntityEncoder.scala
```diff
@@ -88,33 +88,34 @@ object EntityEncoder {
...
     }
 
   /** A stream encoder is intended for streaming, and does not calculate its
     * bodies in advance.  As such, it does not calculate the Content-Length in
     * advance.  This is for use with chunked transfer encoding.
     */
-  implicit def streamEncoder[F[_], A](implicit
-      W: EntityEncoder[F, A]): EntityEncoder[F, Stream[F, A]] =
-    new EntityEncoder[F, Stream[F, A]] {
+  implicit def streamEncoder[F[_], Body, A](implicit
```
nafg, line 115: Should there be consistency in type parameter order across different things?

nafg (changes requested):

nafg:
> Chris started work on an alternative model where the HTTP entity as encoded as an ADT: #4509 . There were some questions about being able to extend the model later (e.g. supporting zero-copy file transfers later), would be interesting to see if this approach solves any of those challenges

Maybe they can be combined, but the idea of this PR seems like a major improvement. It makes assumptions more decoupled. Orthogonal concerns are separated.

diesalbla:
> If we make messages covariant (#5251) then empty or pure messages can be Response[Nothing], with the added benefit that they covary up to Response[F]. Unit isn't part of any interesting type hierarchy, so how do you return a Response[Unit] from a route that may need to return a body?

> Doesn't this push a lot of complexity on to users in terms of defining representations for message bodies? Unless they define something like Chris's ADT themselves, they'll have a hard time composing routes that return a combination of strict/streaming/empty bodies in different formats

Well, we currently have a low-level lowest-common-denominator representation, `Stream[F, Byte]`, and then make users import the implicit translations from each type to that representation. Part of the goal of this change would be to hide that low-level representation, and let users have Json objects, or even domain objects themselves, as the type parameters for routes. Sketching it a bit, to be able to say that an http route for a "Book Library" API would be defined as a function of `Request[BookReq] => F[Option[Response[BookResponses]]]`... but this is a rather idle speculation. For now, best leave this to rest.

PR #4956, opened 2021-07-10
Catch DateTimeException for QueryParamDecoder failures, add codecs for remaining java.time types

dantb:
I had an issue where http4s was producing a 5xx with no information for an invalid `ZonedDateTime`, rather than a 400 / `Validated.Invalid`. It was due to handling a `DateTimeParseException` but not its supertype `DateTimeException` in the `QueryParamDecoder` (can be thrown when the formatter doesn't quite match up). Any objections to using `Validated.catchNonFatal`? Also added a test to reproduce the specific issue / prove it's fixed.

Other than that:
1. Added codecs for the other `java.time` types while in the area - personally there has been a need for e.g. `QueryParamDecoder[OffsetDateTime]` in the past. Tried to minimise duplication as they mostly follow a similar pattern.
2. One weirdness in the codec tests is that scalacheck's `Arbitrary[Year]` instance was incorrect so I've defined a new one - fixed in https://github.com/typelevel/scalacheck/commit/2ae1be5c8e5ee1c14abea607d631e334a56796de and TODO added to use that once released.
3. Removed the custom `Arbitrary` instances to use those in scalacheck instead. I'm hoping that zone related JDK8 bug will be protected by the if clause in the tests? Can always revert back if it breaks the CI though 😃

### DIFF
--- a/core/src/main/scala/org/http4s/QueryParam.scala
+++ b/core/src/main/scala/org/http4s/QueryParam.scala
@@ -19,8 +19,9 @@ package org.http4s
 import cats.{Contravariant, Functor, Hash, MonoidK, Order, Show}
 import cats.data.{Validated, ValidatedNel}
 import cats.syntax.all._
-import java.time.{Instant, LocalDate, ZonedDateTime}
-import java.time.format.{DateTimeFormatter, DateTimeParseException}
+
+import java.time._
+import java.time.format.DateTimeFormatter
 import java.time.temporal.TemporalAccessor
 
 final case class QueryParameterKey(value: String) extends AnyVal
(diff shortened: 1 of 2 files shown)


hamnis:
Thanks!

I tried cherry picking this to 0.22.
Most of the time it passes, but sometimes seems to be a bit flaky:

```failing seed for QueryParamCodec.decode . emapValidatedNel(InvalidNel) . encode == failedNel is XZzT9qy0cwkjt-KuOIxKdzXSCCYGRNwC_gklIWMRVQD=
==> X org.http4s.QueryParamCodecSuite.ZoneOffset QueryParamCodec: QueryParamCodec.decode . emapValidatedNel(InvalidNel) . encode == failedNel  0.024s munit.FailException: /home/maedhros/Projects/forks/http4s/tests/src/test/scala/org/http4s/QueryParamCodecSuite.scala:48
47:  checkAll("YearMonth QueryParamCodec", QueryParamCodecLaws[YearMonth])
48:  checkAll("ZoneOffset QueryParamCodec", QueryParamCodecLaws[ZoneOffset])
49:  checkAll("ZoneId QueryParamCodec", QueryParamCodecLaws[ZoneId])

Failing seed: as_Eb3PLXveHYkoKfs8mlgkHrA8SIYUEovaKDDisXQJ=

  override val scalaCheckInitialSeed = "as_Eb3PLXveHYkoKfs8mlgkHrA8SIYUEovaKDDisXQJ="
```

dantb:
Thanks @hamnis, reproduced and fixed in https://github.com/http4s/http4s/pull/4956/commits/c58d3ac314c22f339515755b113e8b0dc202544b. Ran the suite a few dozen times for different java versions to be sure

rossabaker:
This is cherry-picked for release in 0.22.1.  Thanks!

Issue #4899, opened 2021-05-27
Flaky Tomcat test


rossabaker:
```
==> X org.http4s.tomcat.server.TomcatServerSuite.server should route requests on the service executor  5.567s munit.ComparisonFailException: /home/runner/work/http4s/http4s/tomcat-server/src/test/scala/org/http4s/tomcat/server/TomcatServerSuite.scala:97
96:      .map(_.startsWith("http4s-suite-"))
97:      .assert
98:  }
value is not true
=> Obtained
false
=> Diff (- obtained, + expected)
-false
+true
```

https://github.com/http4s/http4s/pull/4898/checks?check_run_id=2687198536

rossabaker:
This keeps happening.  We should have the test at least tell us which thread it is.

rossabaker:
Got one: https://github.com/http4s/http4s/pull/5028/checks?check_run_id=3262369056 /cc @isomarcte

isomarcte:
@rossabaker do you think that is a default HTML error page? I should have re-written those better, to dump the whole output on failure.

rossabaker:
I just looked at the test name when I pinged you, and now GitHub can't find the log. 😩

isomarcte:
@rossabaker the string it was matching was something like `<doctype`

rossabaker:
Yep, sounds like.  The default Tomcat error pages are all HTML.

rossabaker:
Here's [another one](https://github.com/http4s/http4s/pull/5057/checks?check_run_id=3292905974)

```
==> X org.http4s.tomcat.server.TomcatServerSuite.server should route requests on the service executor  6.272s munit.ComparisonFailException: /home/runner/work/http4s/http4s/tomcat-server/src/test/scala/org/http4s/tomcat/server/TomcatServerSuite.scala:101
100:      .map(_.take(prefix.size))
101:      .assertEquals(prefix)
102:  }
values are not the same
=> Obtained
<!doctype htm
=> Diff (- obtained, + expected)
-<!doctype htm
+http4s-suite-
```

It's probably the response timing out because we still ask too much of these free CI boxes.

PR #4957, opened 2021-07-11
Bump dependencies before 0.22.0

rossabaker:


rossabaker:
This bumped Tomcat and we got a Tomcat error I don't remember seeing:

```
==> X org.http4s.tomcat.server.TomcatServerSuite.server should route requests on the service executor  4.111s munit.ComparisonFailException: /home/runner/work/http4s/http4s/tomcat-server/src/test/scala/org/http4s/tomcat/server/TomcatServerSuite.scala:100
99:       .map(_.startsWith("http4s-suite-"))
100:      .assert
101:  }
value is not true
=> Obtained
false
-false
+true
```

hamnis:
This apparently happened before this upgrade: https://github.com/http4s/http4s/issues/4899

Issue #4916, opened 2021-06-07
Possible Blaze Client Performance Regression in 0.21.23


djspiewak:
Seeing some anecdotal issues with performance regressions between 0.21.22 and 0.21.23. In particular, it appears to be the blaze-client (and possibly not the server) which is having problems. The symptoms are slightly higher mean response times, and *much* higher tail response times (even the p70s climb, it it's most stark in p95, p99 and p9999s). The regression appears to be on the order of magnitude of 2x in the p99s, and the instability is quite significant.

Working to see if I can get some narrower data on this or even ideally a reproduction.

djspiewak:
Can confirm that the performance regression is real. The general outcome is that the p9999 latency is just about doubled. p50 seems mostly unaffected. We're bisecting snapshots and running production performance stress tests between 0.21.22 and 0.21.23 right now to see if we can find the commit that introduced the regression.

djspiewak:
Current bisection status: the regression was likely introduced in either #4810 or #4812.

Baccata:
After applying some educated dichotomy between 0.21.22 and 0.21.23 and running performance tests against an actual service, I can confirm that https://github.com/http4s/http4s/pull/4810 introduced the regression.

djspiewak:
Additionally, @Baccata was able to confirm that #4815 appears to amplify the regression, bringing the p9999 degradation from 2x to around 3x. This pattern is seen across multiple services which are using http4s-blaze-client, so I don't believe it is in any way usage-dependent.

My recommendation would be that these semantics be made opt-in and configurable. @rossabaker @RafalSumislawski

RafalSumislawski:
In absolute terms, how high are the response times? When you say p9999 increased 3x, is it an increase from 10ms to 30ms, or 100ms to 300ms, or something even bigger? Is the system running under 100% CPU load in the stress tests?

#4810 added starting and IO operation (read), and #4815 added waiting for an IO operation (write) to complete on the execution path before the point in which the result of the request is available. I think we could try moving them to background, in the sense that the response would be made available early, and then on a different thread the connection would enter some intermediate state in which it would be prepared to be recycled back into the connection pool.

Also, in many cases we may be able to reduce the delay introduced by the waiting for the write to complete, by skipping a write if the content is an empty array.

Baccata:
In absolute terms, the 3x is from ~35ms to ~100ms. The service is not running under 100% CPU load (I signed off work so can't check for actual numbers but I can tell you that much)

[...]

PR #4958, opened 2021-07-11
Reuse idle timeout stage

RafalSumislawski:
This PR:
- addresses the #4916, in which a leak of idle timeouts / `IdleTimeoutStage`'s was observed. Now there is zero or one `IdleTimeoutStage` per `Http1Connection`. That stage is created together with the connection and keeps being reused though the life of the connection. The logic of `IdleTimeoutStage` makes sure that at most one timeout is scheduled.
- avoids the cost or creating `IdleTimoutStage` per request. In many cases that cost is minuscule. But that's not the case when using log4j2 on Java 11 (see similar issue in armeria https://github.com/line/armeria/issues/2306).
- Removes `ReadBufferStage` which was supposed to address the same issue as #4810, but apparently it didn't work. And since #4810 is merged the `ReadBufferStage` it obsolete.


djspiewak:
Can confirm that this appears to fix the observed performance regression, at least to within a moderate confidence interval. There's a decent chance that performance is still *slightly* worse across the board around 15% slower in all percentiles, but more testing would be required to say for sure. At the very least, this is a really nice fix!

ChristopherDavenport (approved):

Baccata:
We've tested it and are happy with the results :+1:

Issue #4952, opened 2021-07-08
[0.23] Error parsing address from Jetty with ipv6


perok:
Running http4s 0.23.0-RC1 with Jetty and IPv6 crashes with the following:

```
ERROR org.http4s.servlet.Http4sServlet - Error processing request
java.util.NoSuchElementException: None.get
```

The issue is that Jetty returns the following addresses:

```
req.getLocalAddr -> "[0:0:0:0:0:0:0:1]"
req.getRemoteAddr -> "[0:0:0:0:0:0:0:1]"
```

This crashes the following parsing `IpAddress.fromString(req.getRemoteAddr).get`in Http4sServlet. Is brackets on the address considered an error on the parsing side or on Jetty's side?

A temporary workaround for us is to force ipv4 with `-Djava.net.preferIPv4Stack=true`.

rossabaker:
[hangs head in shame for using `.get`]

This looks related to https://github.com/eclipse/jetty.project/issues/1503.  Rather than say Jetty or ip4s is wrong, I'd say http4s is wrong for assuming they agree.  I think we should strip brackets before that call.

Arguably, ip4s should handle both, like the Java APIs.  But that's up to @mpilquist.

mpilquist:
No strong opinions. At first glance, I'd say that bracketed v6 is part of the URI RFC, not IPv6, and hence the brackets should be removed prior to calling `IpAddress.fromString`. But I very well may be missing something and it may be more practical to handle it in ip4s regardless. Let me know what you think.

rossabaker:
I think [this Jetty comment](https://github.com/eclipse/jetty.project/issues/1503#issuecomment-760995044) summarizes the grim situation from the specs.  Since I used it as a drop-in for the Java libraries, the argument that it should be compatible with the Java libraries seems good to me.

Regardless, I think we should fix it here and not rely on the generosity of ip4s' parser.

mpilquist:
😱

rossabaker:
@san-coding I don't think anybody has taken this up, if you're still looking for one.

san-coding:
> @san-coding I don't think anybody has taken this up, if you're still looking for one.

Thanks ross, yes I am looking forward to contribute

san-coding:
[...]

PR #4967, opened 2021-07-18
Fix for #4952

Zett98:
no sure if the code in question should be 
```scala
{
                val localAddr = req.getLocalAddr()
                val ipAddr = IpAddress
                  .fromString(localAddr)
                  .orElse(IpAddress.fromString(stripBracketsFromAddr(localAddr)))
                ipAddr.get
 }
```
instead of the currently submitted 
```scala
IpAddress.fromString(stripBracketsFromAddr(req.getLocalAddr)).get
```


rossabaker:
This looks right to me.  I'll cherry-pick it onto 0.22, because I think we broke it there.

I'd love to see a test, but this gets deep enough into servlets it's probably more trouble than it's worth.

rossabaker:
Cherry-picked to series/0.22. Thanks!

PR #4983, opened 2021-07-23
FS2 - Small related rewrites

diesalbla:
- Pattern `stream.through(_.method)` is same as `stream.method`
- No need to `void` a compile-drained stream. Unlike `Stream.drain`, which returns a `Stream[F, Nothing]`, the call `stream.compile.drain` returns an `F[Unit]`
- Avoid using pipes, in favour of stream methods

### DIFF
--- a/blaze-core/src/main/scala/org/http4s/blazecore/websocket/Http4sWSStage.scala
+++ b/blaze-core/src/main/scala/org/http4s/blazecore/websocket/Http4sWSStage.scala
@@ -53,22 +53,15 @@ private[http4s] class Http4sWSStage[F[_]](
   def name: String = "Http4s WebSocket Stage"
 
   //////////////////////// Source and Sink generators ////////////////////////
-  def snk: Pipe[F, WebSocketFrame, Unit] =
-    _.evalMap { frame =>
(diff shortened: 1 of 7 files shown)


blaze-core/src/main/scala/org/http4s/blazecore/websocket/Http4sWSStage.scala
```diff
@@ -153,21 +150,18 @@ private[http4s] class Http4sWSStage[F[_]](
...
-    val receiveSend: Pipe[F, WebSocketFrame, WebSocketFrame] =
+    val receiveSent: Stream[F, WebSocketFrame] =
       ws match {
         case WebSocketSeparatePipe(send, receive, _) =>
-          incoming =>
-            send.concurrently(
-              incoming.through(receive).drain
-            ) //We don't need to terminate if the send stream terminates.
+          //We don't need to terminate if the send stream terminates.
+          send.concurrently(receive(inputstream))
```
diesalbla, line 157: No need to `drain` the background stream in `concurrently`.

rossabaker:
This looks good.  I'm going to cherry-pick it back to 0.23 to minimize branch divergence.

PR #4986, opened 2021-07-28
Merge 0.22.0

rossabaker:
The non-trivial parts of this are @RaasAhsan's Ember websocket work and @RafalSumislawski's blaze-client work, porting their effectful bits from CE2 to CE3.


blaze-client/src/main/scala/org/http4s/blaze/client/BlazeClient.scala
```diff
@@ -63,69 +62,46 @@ object BlazeClient {
...
+                        fiber.cancel >> outcome.embed(
+                          F.raiseError(new CancellationException("Response canceled")))
+                      case Right((fiber, outcome)) =>
+                        fiber.cancel >> outcome.fold(
+                          F.raiseError(new TimeoutException("Response timeout also timed out")),
+                          F.raiseError,
+                          _.flatMap(F.raiseError)
+                        )
+                    }
+                }
```
rossabaker, line 103: These `outcome` folds are novel work.  I think they're right, but perhaps not idiomatic.  The "Response timeout also timed out" situation would be particularly bizarre, but the types made me handle it.

blaze-client/src/test/scala/org/http4s/blaze/client/BlazeClientBase.scala
```diff
@@ -87,6 +87,29 @@ trait BlazeClientBase extends Http4sSuite {
...
+            exchange.sendResponseHeaders(204, 0L)
+            exchange.close()
+          case "/process-request-entity" =>
+            // We wait for the entire request to arrive before sending a response. That's how servers normally behave.
+            var result: Int = 0
+            while (result != -1)
+              result = exchange.getRequestBody.read()
+            exchange.sendResponseHeaders(204, 0L)
+            exchange.close()
+        }
```
rossabaker, line 112: This is a translation from what is still a servlet in the 0.22 branch.  I suppose the Jetty comments are no longer accurate.

ember-server/src/main/scala/org/http4s/ember/server/internal/WebSocketHelpers.scala
```diff
@@ -0,0 +1,278 @@
...
+  def upgrade[F[_]](
+      socket: Socket[F],
+      req: Request[F],
+      ctx: WebSocketContext[F],
+      buffer: Array[Byte],
+      receiveBufferSize: Int,
+      idleTimeout: Duration,
+      onWriteFailure: (Option[Request[F]], Response[F], Throwable) => F[Unit],
+      errorHandler: Throwable => F[Response[F]],
+      logger: Logger[F])(implicit F: Async[F]): F[Unit] = {
```
rossabaker, line 62: This moves from `Concurrent` to `Async` because there's a `Sync` buried within, and I didn't think too hard about how to avoid that.  It might be possible to walk that back, but we'd probably have to do assure ourselves some side effect is not observable.

ember-server/src/main/scala/org/http4s/ember/server/internal/WebSocketHelpers.scala
```diff
@@ -0,0 +1,278 @@
...
+
+    val key = req.headers.get[`Sec-WebSocket-Key`] match {
+      case Some(header) => Right(header.value)
+      case None => Left(KeyNotFound)
+    }
+
+    (connection, upgrade, version, key).mapN { case (_, _, _, key) => key }
+  }
+
+  private def serverHandshake[F[_]](value: String)(implicit F: MonadThrow[F]): F[ByteVector] =
```
rossabaker, line 240: This was `Sync` in 0.22, but all we're doing is suspending errors into `F`, I think.  I was hoping it got us out of the `Async` constraint mentioned elsewhere in this PR, but nope.

PR #4988, opened 2021-07-29
Append, not replace, Accept header in client methods

### LINKED ISSUE #4987: [Bug] Accept header not appended properly in 0.23.0-RC1
I tried the following ways to construct the `Accept` header

```
    Headers(
      Accept(
        MediaRangeAndQValue
          .withDefaultQValue(MediaRange.parse("application/vnd.docker.distribution.manifest.v2+json").toOption.get)
      ),
      Header.Raw(ci"Accept", "application/vnd.docker.distribution.manifest.v2+json"),
      Accept.parse("application/vnd.docker.distribution.manifest.v2+json").toOption.get
    )
```

But none of them actually appended the header properly.

In previous version, headers in a call looked like this
```
{"@timestamp":"2021-07-28T15:59:16.160Z","@version":"1","message":"HTTP/1.1 GET https://registry-1.docker.io/v2/rtitle/anvil-rstudio-base/blobs/sha256:aaf072362a3bfa231f444af7a05aa24dd83f6d94ba56b3d6d0b365748deac30a Headers(Accept: application/vnd.docker.distribution.manifest.v2+json, Authorization: <REDACTED>, Accept: application/json)","logger_name":"org.http4s.client.middleware.RequestLogger","thread_name":"scala-execution-context-global-1942","severity":"INFO","level_value":20000}
```

But it now looks like this with 0.23.0-RC1
```
{"@timestamp":"2021-07-28T16:07:41.141Z","@version":"1","message":"HTTP/1.1 GET https://registry-1.docker.io/v2/rtitle/anvil-rstudio-base/manifests/0.0.1 Headers(Content-Type: application/vnd.docker.distribution.manifest.v2+json, Authorization: <REDACTED>, Accept: application/json)","logger_name":"org.http4s.client.middleware.RequestLogger","thread_name":"io-compute-8","severity":"INFO","level_value":20000}
```

rossabaker:
Fixes #4987 

If an `Accept` header is present on the request, preserve it along with headers from the EntityDecoder.  This preserves the 0.21 behavior.

There's an argument to be made that we should only supply it if it was missing, because the user needs to customize the EntityDecoder (awkward) to get full control over the Accept header otherwise.  But that would be a breaking semantic change not surfaced by the types.

Bonus: add an `addHeader` to `Message`, because it was awkward without.

Bonus bonus: enable doctests in core.  Examples in the docs are good, and we should be doing more of this.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -120,6 +120,7 @@ lazy val core = libraryProject("core")
       ip4sCore,
       literally,
       log4s,
+      munit % Test,
       scodecBits,
       slf4jApi, // residual dependency from macros
       vault,
--- a/client/src/main/scala/org/http4s/client/DefaultClient.scala
+++ b/client/src/main/scala/org/http4s/client/DefaultClient.scala
@@ -88,7 +88,7 @@ private[http4s] abstract class DefaultClient[F[_]](implicit F: BracketThrow[F]) 
       d: EntityDecoder[F, A]): F[A] = {
     val r = if (d.consumes.nonEmpty) {
       val m = d.consumes.toList
-      req.putHeaders(Accept(MediaRangeAndQValue(m.head), m.tail.map(MediaRangeAndQValue(_)): _*))
+      req.addHeader(Accept(MediaRangeAndQValue(m.head), m.tail.map(MediaRangeAndQValue(_)): _*))
     } else req
 
     run(r).use {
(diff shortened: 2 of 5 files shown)


ChristopherDavenport (approved):
Build not green. Code looks solid.

PR #4990, opened 2021-07-29
Updates before 0.22.1

rossabaker:
Mostly boring, ~~but we'll get Boopickle for Scala 3 now~~.


rossabaker:
I'm going to spin off boopickle.  It doesn't work on Scala 3.

PR #4991, opened 2021-07-29
Boopickle 1.4.0

rossabaker:


rossabaker:
@cquiroz Do you know why this doesn't compile in Scala 3?  I see upstream they split off the tests between Scala 2 and Scala 3, but I don't spot the difference in the relevant part.

japgolly:
Hello! The problem is in `build.sbt:401`:

```scala
Http4sPlugin.boopickle.cross(CrossVersion.for3Use2_13)
```

should be

```scala
Http4sPlugin.boopickle
```

I tried it locally and with that change all tests compile and pass :)

rossabaker:
Oh, right!  I generally avoid that cross-compilation feature, but we used it to keep publishing the integration for Scala 2 until boopickle supported Scala 3.  Thank you, @japgolly!

japgolly:
No worries, happy to help! :relaxed:

rossabaker:
There's a slight binary incompatibility in this version of boopickle.  Our usual conservative approach would be to defer this upgrade to http4s-0.23, but these _probably_ aren't going to hurt anyone?

```
sbt:boopickle> boopickleJVM/mimaReportBinaryIssues
[error] boopickle: Failed binary compatibility check against io.suzaku:boopickle_2.13:1.3.3! Found 2 potential problems
[error]  * method bp()boopickle.BasicPicklers#BytePickler# in object boopickle.BasicPicklers#BigDecimalPickler has a different result type in current version, where it is boopickle.Pickler rather than boopickle.BasicPicklers#BytePickler#
[error]    filter with: ProblemFilters.exclude[IncompatibleResultTypeProblem]("boopickle.BasicPicklers#BigDecimalPickler.bp")
[error]  * method bp()boopickle.BasicPicklers#BytePickler# in object boopickle.BasicPicklers#BigIntPickler has a different result type in current version, where it is boopickle.Pickler rather than boopickle.BasicPicklers#BytePickler#
[error]    filter with: ProblemFilters.exclude[IncompatibleResultTypeProblem]("boopickle.BasicPicklers#BigIntPickler.bp")
[error] stack trace is suppressed; run last mimaReportBinaryIssues for the full output
```

japgolly:
@rossabaker Oh sorry about that, mima isn't enabled in the Boopickle build. Those are completely reasonable differences to ignore. Now that I see these methods, they shouldn't even be public anyway, and there's no reason for anyone would have used them. I'd say ignore away

ChristopherDavenport (approved):

PR #4997, opened 2021-07-29
Update MimeDb before 0.23.0

rossabaker:
Sneaks in any binary breaking changes before the 0.23 lockdown.


rossabaker:
I wish we'd done this before 0.22, but c'est la vie.

ChristopherDavenport (approved):

PR #4998, opened 2021-07-29
Remove deprecatia before 0.23

rossabaker:
Cleaning out things that have been deprecated for at least two years before bincompat is locked.


rossabaker:
There's more, but it's soccer time. :soccer:

ChristopherDavenport (approved):
🎉

rossabaker:
Bah, this one should have been series/0.23.

PR #5011, opened 2021-08-03
Add status 418 I'm a Teapot

bplommer:
Adds status `418 I'm a Teapot` in accordance with [RFC 2324](https://datatracker.ietf.org/doc/html/rfc2324#section-2.3.2)

### DIFF
--- a/core/src/main/scala/org/http4s/Status.scala
+++ b/core/src/main/scala/org/http4s/Status.scala
@@ -174,6 +174,7 @@ object Status {
   val UnsupportedMediaType: Status = register(Status(415, "Unsupported Media Type"))
   val RangeNotSatisfiable: Status = register(Status(416, "Range Not Satisfiable"))
   val ExpectationFailed: Status = register(Status(417, "Expectation Failed"))
+  val ImATeapot: Status = register(Status(418, "I'm A Teapot"))
   val MisdirectedRequest: Status = register(Status(421, "Misdirected Request"))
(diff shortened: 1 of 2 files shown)


ChristopherDavenport:
Relevant RFCS - 
https://datatracker.ietf.org/doc/html/rfc2324#section-2.3.2
https://datatracker.ietf.org/doc/html/rfc7168#section-2.3.3

May be scope creep, but may need appropriate media types to respond to to HTCPCP-TEA protocol.

bplommer:
> May be scope creep, but may need appropriate media types to respond to to HTCPCP-TEA protocol.

Maybe this should go in a separate module? `http4s-teapot`?

rossabaker:
Sadly the [standardization process](https://tools.ietf.org/id/draft-nottingham-thanks-larry-00.html) failed, but I'm not about to stand idly while the Python, Node, and Go folks get to use it.

> Maybe this should go in a separate module? http4s-teapot?

Surely `htcpcp4s`.

ChristopherDavenport (approved):

rossabaker:
The serious reason to merge this is that some HTTP servers in the wild do return it, even when not controlling a coffee pot.  The other serious reason is that [Emacs has a proper HTCPCP client](http://emarsden.chez.com/downloads/coffee.el), and Emacs is better than whatever you all use.

rossabaker:
Cherry-picked to 0.22.

PR #5013, opened 2021-08-03
Add DataTypes For Extracting Out The Non-Body Portion Of A Message

isomarcte:
This commit adds two data types, `RequestInfo` and `ResponseInfo`, which extract out the non-body portion of a `Request[F]` and `Response[F]`.

This data types can be used in many contexts, where we explicitly _don't_ want to deal with the body.

The only field these data types omit, other than the body, is `attributes` as that would prevent an `Order` instance from being written and it is unlikely that _most_ uses of these data types will explicitly care about the `attributes`. Of course, if they do, it is trivial to create a `(Vault, RequestInfo)` tuple.

### DIFF
--- a/core/src/main/scala/org/http4s/RequestInfo.scala
+++ b/core/src/main/scala/org/http4s/RequestInfo.scala
@@ -115,6 +115,6 @@ object RequestInfo {
   implicit val catsShowForRequestInfo: Show[RequestInfo] =
     Show.fromToString
 
-  implicit val stdLibOrdering: Ordering[RequestInfo] =
+  implicit def stdLibOrdering: Ordering[RequestInfo] =
     catsHashAndOrderForRequestInfo.toOrdering
 }
--- a/core/src/main/scala/org/http4s/RequestInfo.scala
+++ b/core/src/main/scala/org/http4s/RequestInfo.scala
@@ -1,0 +1,119 @@
+/*
+ * Copyright 2013 http4s.org
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.http4s
+
+import cats._
+import cats.syntax.all._
+import org.http4s.internal.reduceComparisons
+
+/** A projection of a [[Request]] without the body.
+  *
+  * @note The [[Request#attributes]] are omitted in this encoding because they
+  *       do not (and can not) have a [[cats.kernel.Order]] instance. If they
+  *       were included here, then we could not write a [[cats.kernel.Order]]
+  *       instance for [[RequestInfo]], limiting some of its utility, e.g. it
+  *       could not be used in a [[cats.data.NonEmptySet]].
+  */
+sealed abstract class RequestInfo extends Product with Serializable {
+  def headers: Headers
+  def httpVersion: HttpVersion
+  def method: Method
+  def uri: Uri
+
+  def withHeaders(value: Headers): RequestInfo
+  def withHttpVersion(value: HttpVersion): RequestInfo
+  def withMethod(value: Method): RequestInfo
+  def withUri(value: Uri): RequestInfo
+
+  // final //
+
+  final def mapHeaders(f: Headers => Headers): RequestInfo =
+    withHeaders(f(headers))
+
+  final def mapHttpVersion(f: HttpVersion => HttpVersion): RequestInfo =
+    withHttpVersion(f(httpVersion))
+
+  final def mapMethod(f: Method => Method): RequestInfo =
+    withMethod(f(method))
+
+  final def mapUri(f: Uri => Uri): RequestInfo =
+    withUri(f(uri))
+
+  final override def toString: String =
+    s"RequestInfo(headers = ${headers.redactSensitive()}, httpVersion = ${httpVersion}, method = ${method}, uri = ${uri})"
+}
+
+object RequestInfo {
+  private[this] final case class RequestInfoImpl(
+    override final val headers: Headers,
+    override final val httpVersion: HttpVersion,
+    override final val method: Method,
+    override final val uri: Uri
+  ) extends RequestInfo {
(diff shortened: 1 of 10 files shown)


core/src/main/scala/org/http4s/RequestInfo.scala
```diff
@@ -0,0 +1,120 @@
...
+sealed abstract class RequestInfo extends Product with Serializable {
+  def headers: Headers
+  def httpVersion: HttpVersion
+  def method: Method
+  def uri: Uri
+
+  def withHeaders(value: Headers): RequestInfo
+  def withHttpVersion(value: HttpVersion): RequestInfo
+  def withMethod(value: Method): RequestInfo
+  def withUri(value: Uri): RequestInfo
```
rossabaker, line 40: Do you actually want to manipulate these, or just project them from an existing request?

core/src/main/scala/org/http4s/RequestInfo.scala
```diff
@@ -0,0 +1,120 @@
...
+          Eval.later(x.httpVersion.compare(y.httpVersion)),
+          Eval.later(x.method.compare(y.method)),
+          Eval.later(x.uri.compare(y.uri))
+        )
+    }
+
+  implicit val catsShowForRequestInfo: Show[RequestInfo] =
+    Show.fromToString
+
+  implicit val stdLibOrdering: Ordering[RequestInfo] =
```
rossabaker, line 118: I think I usually implement `Ordered`, but I can't articulate why.  Do you have a reason for a preference for `Ordering`?

rossabaker (commented):
We've had people try to define equality on requests before several times, and this is a path to doing it safely, with known limitations.  :+1: to the idea.

History lesson: We used to at one point call this the "prelude", but I don't think the term comes from any RFC or exists in any public API.  You'll still find it in the code in a few places where the concept is hinted at.  It was a first-class type in the very early days, when we were on play-iteratee(!).

core/src/main/scala/org/http4s/RequestInfo.scala
isomarcte: I can see a lot of use cases for manipulating them. Mostly relating to normalizing them for the purposes of logging, caching, and metrics. That is, a user might only care that the uri and method are the same, and thus they may,

```scala
_.withHeaders(Headers.empty).withHttpVersion(HttpVersion.`HTTP/1.1`)
```

I also just tend to add both `withFoo` and `mapFoo` methods as a rule these days to data types which don't expose a `copy` for binary compatibility. If we wanted to add `withHeaders` later, without breaking binary compatibility, we'd have to do some gymnastics.

```scala
sealed abstract class RequestInfo {
/* elided */

// Default to make MiMa not complain
  def withHeaders(value: Headers): RequestInfo = headers
}

object RequestInfo {
    private[this] final case class RequestInfoImpl(/* elided */) extends RequestInfo {
      override def withHeaders(value: Headers): RequestInfo = this.copy(headers = value)
    }
/* elided */
}
```

core/src/main/scala/org/http4s/RequestInfo.scala
isomarcte: Using `Ordering` allows you to write more general functions. 

This is because all types which implement `Ordered` have an `Ordering` instance, but the inverse is not true. There _is_ an implicit conversion on the companion object of `Ordered` which gives you an `Ordered` from an `Ordering`, but you can't use it as a constraint. For example, given these data types, `Foo` subtyping `Ordered[Foo]` and `Bar` implementing `Ordering[Bar]`,

```scala
  final case class Foo(value: Int) extends Ordered[Foo] {
    override final def compare(that: Foo): Int =
      value.compare(that.value)
  }

  final case class Bar(value: Int)

  object Bar {
    implicit val orderingInstance: Ordering[Bar] =
      new Ordering[Bar] {
        override def compare(x: Bar, y: Bar): Int =
          x.value.compare(y.value)
      }
  }
```

We run into this problem if we try to write a generic function on types that are orderable if we used `Ordered[A]`.

```scala
def usingOrdered[A <: Ordered[A]](x: A, y: A): Int =
  x.compare(y)

scala> val foo: Foo = Foo(1)
val foo: Foo = Foo(1)

scala> val bar: Bar = Bar(1)
val bar: Bar = Bar(1)

scala> usingOrdered(foo, foo)
usingOrdered(foo, foo)
val res0: Int = 0

scala> import Ordered.orderingToOrdered

scala> usingOrdered(bar, bar)
usingOrdered(bar, bar)
       ^
       error: inferred type arguments [Test.Bar] do not conform to method usingOrdered's type parameter bounds [A <: Ordered[A]]
                    ^
       error: type mismatch;
        found   : Test.Bar
        required: A
                         ^
       error: type mismatch;
        found   : Test.Bar
        required: A
```

The `usingOrdering(bar, bar)` fails because even though we can get an `Ordered[Bar]` just fine via the implicit conversion, `Bar` still doesn't subtype `Ordered[Bar]`. The `Ordered[Bar]` we get also has the wrong arity and semantics for use as a constraint as it is pre-applied to whatever `Bar` was used for the implicit conversion.

However, this works for both `Foo` and `Bar`,

```scala
def usingOrdering[A: Ordering](x: A, y: A): Int =
  Ordering[A].compare(x, y)

scala> usingOrdering(foo, foo)
usingOrdering(foo, foo)
val res2: Int = 0

scala> usingOrdering(bar, bar)
usingOrdering(bar, bar)
val res3: Int = 0
```

So even if we did subtype `Ordered[RequestInfo]` here, we'd probably still want to have an `Ordering[RequestInfo]`.

isomarcte:
> History lesson: We used to at one point call this the "prelude", but I don't think the term comes from any RFC or exists in any public API. You'll still find it in the code in a few places where the concept is hinted at. It was a first-class type in the very early days, when we were on play-iteratee(!).

@rossabaker I kind of like "prelude" actually. I spent about 15 minutes looking through the RFCs for some term for this, but couldn't find anything and just settled on "Info", but I don't like how imprecise that is. I'm happy to change these to `ResquestPrelude` and `ResponsePrelude`. Thoughts?

core/src/main/scala/org/http4s/RequestInfo.scala
rossabaker: I can't think offhand of an '<: Ordered[A]` constraint in the standard library, vs. several `: Ordering`s, which I think is correct.  I had a misconception that most standard library types directly implemented `Ordered` to implicitly became `Ordering`s, but they are few.  (In fact, http4s has about as many as the standard library!)

I'm good with `Ordering`, but we might want to make that consistent in 1.0.

isomarcte: https://github.com/http4s/http4s/issues/5017

rossabaker (approved):
:+1: on green after scalafmt

PR #5014, opened 2021-08-03
Use `assertEquals` Rather Than `assert`

isomarcte:
This test was failing on an unrelated PR, but I could not read what the incorrect value was because we were using `assert` rather than `assertEquals`.


isomarcte:
See: https://github.com/http4s/http4s/pull/5013/checks?check_run_id=3235244509

rossabaker (approved):
Yes! I've been meaning to debug this and cursed the same thing.

isomarcte:
@rossabaker I'll get that fixed in a moment.

isomarcte:
@rossabaker Should be fixed now.

rossabaker (approved):
:+1: on green

isomarcte:
I'm actually hoping for red, so I can debug this error. ;)

PR #5015, opened 2021-08-04
Properly deprecate old literals

rossabaker:
Deprecating a macro definition doesn't generate a deprecation warning when the macro is called.  We need to deprecate the code the macro expands to.

We eliminated these entirely in 0.23, but it ate somebody at $WORK because the deprecation clock had never started.


ChristopherDavenport (approved):
Grim indeed

Kazark (approved):

PR #5032, opened 2021-08-06
Wrap blaze EOF in a SocketException

rossabaker:
Fixes secondary complaint in http4s/blaze#292, as well as a colleague's Slack lamentation.

It sucks to get an `EOF` without any context about the request that failed.  If an `EOF` bubbles up from blaze, this wraps it in a `SocketException`.  The exception mentions the request key in the message and comes with a trace.

This is a semantically breaking change: anybody who themselves was catching `EOF` directly will be in for a surprise.  We will need to trumpet this change loudly if we proceed.


blaze-client/src/main/scala/org/http4s/blaze/client/BlazeClient.scala
```diff
@@ -85,6 +86,7 @@ object BlazeClient {
           borrow.use { next =>
             val res: F[Resource[F, Response[F]]] = next.connection
               .runRequest(req)
+              .adaptError { case EOF => new ResponseException(req) }
```
rossabaker, line 89: Other exceptions can be raised by Blaze, and they're worth contemplating.  We explicitly test that a `ConnectException` percolates up.  That has request info in its message, but does not return the request for further introspection.  Should we wrap that too?  Discuss.

core/src/main/scala/org/http4s/HttpException.scala
```diff
@@ -0,0 +1,35 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.http4s
+
+import java.io.IOException
+
+/** Indicates a general failure communicating over HTTP */
+class HttpException(final val message: String = "", final val cause: Option[Throwable] = None)
```
rossabaker, line 22: This is speculative and maybe should go away.  I'm thinking of problems that happen before we have a request to operate on.

core/src/main/scala/org/http4s/HttpException.scala
```diff
@@ -0,0 +1,35 @@
...
+    extends IOException(message) {
+  cause.foreach(initCause)
+}
+
+/** Indicates a failure to obtain a response to a request */
+class ResponseException[F[_]](
+    final val request: Request[F],
```
rossabaker, line 29: Consuming this request body again is probably dangerous, which might make this an outstanding use case for our new `RequestPrelude`.

blaze-client/src/test/scala/org/http4s/blaze/client/BlazeClientSuite.scala
```diff
@@ -244,4 +246,25 @@ class BlazeClientSuite extends BlazeClientBase {
       }
       .assert
   }
+
+  test("Blaze HTTP/1 client should raise a ResponseException when it receives an unexpected EOF") {
+    SocketGroup[IO](testBlocker).use {
+      _.serverResource[IO](new InetSocketAddress(0))
+        .map { case (addr, sockets) =>
+          val uri = Uri.fromString(s"http://[${addr.getHostName}]:${addr.getPort}/eof").yolo
```
rossabaker, line 254: Our URI is a damn disgrace.  This test will be flaky if an IPv6 address isn't returned.

core/src/main/scala/org/http4s/HttpException.scala
rossabaker: Or perhaps we don't need a custom `ResponseException` at all.  Wrapping in a standard `IOException` would get a trace, let us describe the error, and anyone who wants more from the request can always create a client middleware.

core/src/main/scala/org/http4s/HttpException.scala
rossabaker: Gone.

core/src/main/scala/org/http4s/HttpException.scala
rossabaker: Went this direction with it.

blaze-client/src/main/scala/org/http4s/blaze/client/BlazeClient.scala
```diff
@@ -85,6 +87,9 @@ object BlazeClient {
           borrow.use { next =>
             val res: F[Resource[F, Response[F]]] = next.connection
               .runRequest(req)
+              .adaptError { case EOF =>
+                new SocketException(s"HTTP connection closed: ${key}")
```
rossabaker, line 91: It's unusual in the JVM to log the host info with the `IOException`s, which has me wondering if there's a reason for that.  But it's consistent with what blaze does on connection failures.

rossabaker:
The trace we get in the unit test doesn't point to the client call site.  Maybe I'm doing it wrong.

rossabaker:
If we're okay with the new exception, this would be nice to get into the next release.

ChristopherDavenport (approved):

rossabaker:
We'll do two release trains this week so I can stabilize the GHSA merges.

isomarcte (approved):

PR #5036, opened 2021-08-07
Upgrade to fs2-3.1.0

rossabaker:
This work is mostly done on main, but after the files had moved, so it's mostly redone here.  The forward merge from this to main is going to be Not Fun.


rossabaker:
Maybe we have some integrations and deprecations to do ourselves as part of this bump.  I am just making it compile for now.  @armanbilge may have thoughts already.

One area that comes to mind is `StaticFile`, for which there's a WIP to make it an algebra.

armanbilge:
>  I am just making it compile for now.

That's what I did too. Ideally, we'll replace any use of JVM APIs with the new cross-platform fs2.io.file APIs.

> One area that comes to mind is StaticFile, for which there's a WIP to make it an algebra.

Cool, what PR is this?

rossabaker:
I was thinking of #4879, which I owe a review on.  There's also #2072.

PR #5041, opened 2021-08-07
Ember Client bugfix: incorrectly assumes keep-alive for http/1.0

RaasAhsan:
there was a similar bug on the server that would interpret a keep-alive connection for HTTP/1.0 requests when `Connection` header was unspecified. same deal over here, but for the reverse flow of clients


ember-core/src/main/scala/org/http4s/ember/core/Util.scala
```diff
@@ -17,12 +17,47 @@
...
+import cats.data.NonEmptyList
+import java.util.Locale
 
 private[ember] object Util {
 
+  private[this] val closeCi = ci"close"
+  private[this] val keepAliveCi = ci"keep-alive"
+  private[this] val connectionCi = ci"connection"
+  private[this] val close = Connection(NonEmptyList.of(closeCi))
+  private[this] val keepAlive = Connection(NonEmptyList.one(keepAliveCi))
```
rossabaker, line 34: We don't have a tradition of putting constants on headers with common values and/or low cardinality, but I don't know why not.  There's nothing Ember about these.

ember-core/src/main/scala/org/http4s/ember/core/Util.scala
```diff
@@ -17,12 +17,47 @@
...
+      headers.headers.exists {
+        case Header.Raw(name, value) =>
+          name == connectionCi && value.toLowerCase(Locale.ROOT).contains(expected)
+        case _ => false
+      }
+
+    httpVersion match {
+      case HttpVersion.`HTTP/1.0` => hasConnection(keepAliveCi.toString)
+      case HttpVersion.`HTTP/1.1` => !hasConnection(closeCi.toString)
+      case _ => false
```
rossabaker, line 59: This feels like a time bomb in a future HTTP version, but I guess those probably won't go through this code path.

rossabaker (approved):
Looks good to me.

ember-core/src/main/scala/org/http4s/ember/core/Util.scala
RaasAhsan: Could go with the `sys.error()` trick so that we at least get a runtime error rather than some silent, obscure behavior

ember-core/src/main/scala/org/http4s/ember/core/Util.scala
RaasAhsan: Half the reason we're using strings here is so that we don't actually have to parse the headers (we can assume that we're looking at the raw headers since they've just been read from the socket)

RaasAhsan:
actually, i need to fix something here, the `connectionFor` method doesn't work for parsed headers which it's being used for

ember-core/src/main/scala/org/http4s/ember/core/Util.scala
rossabaker: Oh, that's true.  Only the header values would be of much use.  (I bet blaze creates those uncached....)

RaasAhsan:
> actually, i need to fix something here, the connectionFor method doesn't work for parsed headers which it's being used for

nvm it does with the new header model

PR #5042, opened 2021-08-07
Add model for `Access-Control-Request-Method` header, use typed headers in CORS middleware

bplommer:


server/src/main/scala/org/http4s/server/middleware/CORS.scala
```diff
@@ -168,44 +172,37 @@ object CORS {
...
         Header.Raw(CIString(headerName), values.mkString("", ", ", ""))
 
       (
         req.method,
-        req.headers.get(ci"Origin"),
-        req.headers.get(ci"Access-Control-Request-Method")) match {
-        case (OPTIONS, Some(NonEmptyList(origin, _)), Some(NonEmptyList(acrm, _)))
-            if allowCORS(origin, acrm) =>
+        req.headers.get[Origin],
+        req.headers.get[`Access-Control-Request-Method`]) match {
```
bplommer, line 197: Should these use `getWithWarnings` to handle edge cases involving malformed headers?

server/src/main/scala/org/http4s/server/middleware/CORS.scala
```diff
@@ -143,8 +145,10 @@ object CORS {
       F: Applicative[F]): Http[F, G] =
     Kleisli { req =>
       // In the case of an options request we want to return a simple response with the correct Headers set.
-      def createOptionsResponse(origin: Header.Raw, acrm: Header.Raw): Response[G] =
-        corsHeaders(origin.value, acrm.value, isPreflight = true)(Response())
+      def createOptionsResponse(
+          origin: Origin,
+          acrm: `Access-Control-Request-Method`): Response[G] =
```
rossabaker, line 150: I'm surprised this didn't offend MiMa.  Does it not check internal functions?  I think it's safe.

server/src/main/scala/org/http4s/server/middleware/CORS.scala
```diff
@@ -168,44 +172,37 @@ object CORS {
...
-        (config.anyOrigin, config.anyMethod, origin.value, acrm.value) match {
-          case (true, true, _, _) => true
-          case (true, false, _, acrm) =>
-            config.allowedMethods.exists(_.find(_.name === acrm).nonEmpty)
-          case (false, true, origin, _) => config.allowedOrigins(origin)
-          case (false, false, origin, acrm) =>
-            config.allowedMethods.exists(_.find(_.name === acrm).nonEmpty) &&
-              config.allowedOrigins(origin)
-        }
+      def allowCORS(origin: Origin, method: Method): Boolean = {
```
rossabaker, line 184: This is a nice cleanup.

server/src/main/scala/org/http4s/server/middleware/CORS.scala
rossabaker: Maybe.  What are we going to do about it if we get warnings?

bplommer: Probably return `400 Bad Request` if the headers are present but unparseable?

rossabaker: That sounds like a good thing.  Two tricky things generating this response:

1. We don't want to reflect any request values into the response, or we risk XSS attack vectors.
2. Creating our own entities in backends or middleware often is the only non-JSON response an API generates, and that makes certain users grumpy.  We can maybe make it customizable enough with a new type of `MessageFailure`?

bplommer: On further thought, I think we should just use `get` and ignore parsing errors. This is because:

1. CORS headers are generated by the browser, not by client-side user code. Therefore parsing failures can only result from bugs either in the browser or in our parsers, making them an extreme edge case that we expect to occur ~never.
2. When the response to a cross-origin request doesn't contain the right CORS headers, client-side JavaScript knows only that CORS has failed - not why. So it doesn't really matter what we send back anyway, provided we don't send CORS headers.
3. This behaviour is specified by the [W3C recommendation from 2014](https://www.w3.org/TR/2014/REC-cors-20140116/#resource-requests): for all kinds of CORS requests, "If the value of the Origin header is not a case-sensitive match for any of the values in list of origins, do not set any additional headers and terminate this set of steps.", and for preflight requests "If there is no Access-Control-Request-Method header or if parsing failed, do not set any additional headers and terminate this set of steps. The request is outside the scope of this specification." - that document is seemingly no longer normative, having been superseded by the [fetch standard](https://fetch.spec.whatwg.org/) which is far less specific, but it seems reasonable to use it as guidance.
4. It's simpler and therefore less error-prone.

PR #5043, opened 2021-08-08
Ember raise end of stream errors wherever appropriate

RaasAhsan:
Changes:
1. Simplified fixed-length body parsing logic
2. Fixed several bugs where a body stream silently ended if the peer closed its end of the socket without finishing writing. This raises an error now


ember-core/src/main/scala/org/http4s/ember/core/EmberException.scala
```diff
@@ -39,4 +39,16 @@ object EmberException {
   final case class ChunkedEncodingError(message: String) extends EmberException {
     override def getMessage: String = message
   }
+
+  final case class EmptyStream() extends EmberException {
+    override def getMessage: String = "Cannot Parse Empty Stream"
+  }
+
+  final case class ReachedEndOfStream() extends EmberException {
```
rossabaker, line 47: There's a standard `IOException` for this, but since `EmberException` is a `RuntimeException`, we can't extend it.

rossabaker (approved):
Looks reasonable, but let's see what @ChristopherDavenport says.

ChristopherDavenport (approved):

PR #5045, opened 2021-08-08
Ember Server: interrupt is socket closed.

diesalbla:
Sometimes, particularly in the context of server-sent events,
we are  serving in the response an infinite stream of objects.
In those cases, it may happen that the client interrupts
the connection, so we should detect that and interrupt the stream.

### DIFF
--- a/ember-server/shared/src/main/scala/org/http4s/ember/server/internal/ServerHelpers.scala
+++ b/ember-server/shared/src/main/scala/org/http4s/ember/server/internal/ServerHelpers.scala
@@ -170,6 +170,7 @@ private[server] object ServerHelpers extends ServerHelpersPlatform {
     Encoder
       .respToBytes[F](resp)
       .through(_.chunks.foreach(c => timeoutMaybe(socket.write(c), idleTimeout)))
+      .interruptWhen(Stream.repeatEval(socket.isOpen.map(x => ! x)))
       .compile
       .drain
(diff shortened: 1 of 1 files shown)


ember-server/shared/src/main/scala/org/http4s/ember/server/internal/ServerHelpers.scala
```diff
@@ -170,6 +170,7 @@ private[server] object ServerHelpers extends ServerHelpersPlatform {
     Encoder
       .respToBytes[F](resp)
       .through(_.chunks.foreach(c => timeoutMaybe(socket.write(c), idleTimeout)))
+      .interruptWhen(Stream.repeatEval(socket.isOpen.map(x => ! x)))
```
RaasAhsan, line 173: Does `Socket#isOpen` reliably reflect the current state of the channel ? IIRC I think it only told us if _we_ closed our connection or encountered some sort of error on a socket operation

Its probably worth testing out, but I think before if a client terminated its connection while the server was writing, the server would get a `Connection reset by peer` or `Broken pipe` and then just terminate that request

ChristopherDavenport: No it does not. This doesn't work since we are not reading any more. isOpen only reflects whether or not we have closed the socket.

PR #5046, opened 2021-08-08
Correctly handle the lifecycle of a fetch response

armanbilge:
Moves the lifecycle management from the `EntityBody` to the wrapping `Resource`.


dom-fetch-client/src/main/scala/org/http4s/dom/FetchClient.scala
```diff
@@ -34,11 +34,11 @@ import scala.scalajs.js.typedarray.Uint8Array
...
-    (for {
-      body <- req.body.chunkAll.filter(_.nonEmpty).compile.last
-      response <- Async[F].fromPromise {
-        Async[F].delay {
+  def apply[F[_]](implicit F: Async[F]): Client[F] = Client[F] { req: Request[F] =>
+    for {
+      body <- req.body.chunkAll.filter(_.nonEmpty).compile.last.toResource
+      response <- F.fromPromise {
+        F.delay {
           val init = new RequestInit {}
```
isomarcte, line 42: Should we create a general function to safely convert between a http4s Request and a RequestInit?

If `RequestInit` requires the body to be strict, then it'd have to be `Request[F] => Stream[F, RequestInit]` in the most general form. 

If this is the only place we use it, it may not be worth it.

dom-fetch-client/src/main/scala/org/http4s/dom/FetchClient.scala
```diff
@@ -49,27 +49,29 @@ object FetchClient {
 
           Fetch.fetch(req.uri.renderString, init)
         }
+      }.toResource
+      status <- F.fromEither(Status.fromInt(response.status)).toResource
+      body <- Resource.makeCase(response.body.pure[F]) {
```
isomarcte, line 54: I think in order for this to work they way you want, you need to have the `response` be in the acquisition of a Resource. If I'm reading this correctly, you want to ensure that if you get a response, that you always register a `_.cancel` on the body. In order to do this you need to acquire the `response` and register the finalizers to call `_.cancel` in the same `Resource`. In your current encoding, for example, if parsing the `status` fails you will never register the finalizers. 

I think you want something like this,

```scala
  def apply[F[_]](implicit F: Async[F]): Client[F] = Client[F] { req: Request[F] =>
    for {
      body <- req.body.chunkAll.filter(_.nonEmpty).compile.last.toResource
      requestInit <- F.catchNonFatal {
          val init = new RequestInit {}

          init.method = req.method.name.asInstanceOf[HttpMethod]
          init.headers = new Headers(toDomHeaders(req.headers))
          body.foreach { body =>
            init.body = arrayBuffer2BufferSource(body.toJSArrayBuffer)
          }
          init
        }.toResource
      response <- Resource.makeCase(F.fromPromise(F.delay(Fetch.fetch(req.uri.renderString, requestInit)))) {
        case (r, Resource.ExitCase.Succeeded) =>
          F.fromPromise(F.delay(r.body.cancel(null))).void
        case (r, Resource.ExitCase.Errored(ex)) =>
          F.fromPromise(F.delay(r.body.cancel(ex.getMessage()))).void
        case (r, Resource.ExitCase.Canceled) =>
          F.fromPromise(F.delay(r.body.cancel(null))).void
      }
      status <- F.fromEither(Status.fromInt(response.status)).toResource
    } yield Response[F](
      status = status,
      headers = fromDomHeaders(response.headers),
      body = readableStreamToStream(response.body)
    )
  }
```

FWIW, in cases like this where we introduce `ExitCase` based finalizers, I personally find for comps more difficult to read than a `flatMap` based approach, but that is inherently subjective.

dom-fetch-client/src/main/scala/org/http4s/dom/FetchClient.scala
```diff
@@ -49,27 +49,29 @@ object FetchClient {
...
+        case (r, Resource.ExitCase.Succeeded) => F.fromPromise(F.delay(r.cancel(null))).void
```
isomarcte, line 55: Nit: Can we provide a reason other than `null`? I know the scala doc says, 

```scala
  /**
   * The cancel method cancels the stream, signaling a loss of interest in
   * the stream by a consumer. The supplied reason argument will be given to
   * the underlying source, which may or may not use it.
   *
   * @param reason the reason <- actually not what type this is
   * @return a Promise, but not quite sure what it can contain
   */
 ```
 
 but it would be nice to not have the `null`s in the code, even just so future linters don't fuss about it.

isomarcte (changes requested):

dom-fetch-client/src/main/scala/org/http4s/dom/FetchClient.scala
armanbilge: This is the only place we're using it so far, and I don't expect any others (fetch is the modern native client for browsers).

dom-fetch-client/src/main/scala/org/http4s/dom/FetchClient.scala
armanbilge: So the official docs are here which indicate it is an optional argument: https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream/cancel

Since it's an optional argument, the idiomatic thing to do here would be use `js.undefined` but unfortunately the facade didn't account for this. Hopefully this will be fixed in a future version of scala-js-dom. OTOH, if you think we should provide a _reason_ (instead of just trying to avoid using null/linting complaints), I'm open to that :)

dom-fetch-client/src/main/scala/org/http4s/dom/FetchClient.scala
armanbilge: Oops, good catch! You are right, I'll fix this. And I agree, not crazy about the for comprehension here myself, I'll try it the other way.

isomarcte (approved):

PR #5047, opened 2021-08-08
Try G1GC to help flaky Node.js ci

armanbilge:
I think it works better, but hard to know for sure. Maybe merge and let's try it out for a while?

### DIFF
--- a/.jvmopts
+++ b/.jvmopts
@@ -1,3 +1,4 @@
 -Dfile.encoding=UTF8
 -XX:ReservedCodeCacheSize=250M
 -XX:+TieredCompilation
+-XX:MaxRAMPercentage=75.0
--- a/.jvmopts
+++ b/.jvmopts
@@ -3,4 +3,4 @@
 -Xmx6G
 -XX:ReservedCodeCacheSize=250M
 -XX:+TieredCompilation
--XX:+UseParallelGC
+-XX:+UseG1GC
--- a/.jvmopts
+++ b/.jvmopts
@@ -1,6 +1,3 @@
 -Dfile.encoding=UTF8
--Xms1G
--Xmx6G
 -XX:ReservedCodeCacheSize=250M
 -XX:+TieredCompilation
--XX:+UseG1GC
--- a/.jvmopts
+++ b/.jvmopts
@@ -1,4 +1,6 @@
 -Dfile.encoding=UTF8
+-Xms1G
+-Xmx6G
 -XX:ReservedCodeCacheSize=250M
 -XX:+TieredCompilation
--XX:MaxRAMPercentage=75.0
+-XX:+UseG1GC


isomarcte:
@armanbilge do you have a link to the flaky test?

armanbilge:
Yup, it's not flaky test specifically but flaky ci actually. At time of writing these runners have been stuck in the (extremely memory intensive) `fastOptJS` step for more than 3 hours.

https://github.com/http4s/http4s/pull/5046/checks?check_run_id=3274281956
https://github.com/http4s/http4s/pull/5046/checks?check_run_id=3274282026

isomarcte:
@armanbilge I'm not certain this is the JVM. Looking at the raw logs, I see this.

> 2021-08-08T15:52:39.9324150Z Found online and idle hosted runner(s) in the current repository's organization account that matches the required labels: 'ubuntu-latest'
> 2021-08-08T15:52:39.9324195Z Waiting for a hosted runner in 'organization' to pick this job...

That might be github actions scheduler doing something odd.

armanbilge:
Nope, that's normal, all runs start like that. Besides, why would it consistently get stuck at this step? I can point to several other examples. I'm 99% sure this is memory related.

isomarcte:
@armanbilge yeah I realize those lines are normal, but I'm curious since it is _just_ those lines. There is a bunch of stuff that is usually dumped to stdout before the JVM even starts up, but we aren't seeing any of that. If it were the JVM having issues with memory, I'd suspect we'd see other output in the logs before things get stuck.

> Besides, why would it consistently get stuck at this step?

I have no idea. I'm merely not sure why this would be the JVM if we aren't seeing any of the other non-jvm startup output from the job. It looks to me like github actions is never starting that step at all. 

FWIW, I've seen similar behavior like this from github actions before, in other code bases. I had to remove the `windows-latest` agent from one of my projects because one particular step, only for the `windows-latest` agent, would never even start up. I could never figure out why.

armanbilge:
> There is a bunch of stuff that is usually dumped to stdout before the JVM even starts up

Can you give an example of this? It's possible you are not seeing the output you expect because technically the job is still running. If you cancel it (do you have this privilege) I think the rest of the output will flush.

> It looks to me like github actions is never starting that step at all.

I'm pretty sure it started, I was watching it before and it started printing out memory-related issues. Again, I think the logs don't flush until the job stops or is cancelled.

isomarcte:
@armanbilge yeah, I'll cancel it and take a look. You may be right about it not flushing.

isomarcte:
@armanbilge 

>  [warn] In the last 13 seconds, 12.252 (95.0%) were spent in GC. [Heap: 0.84GB free of 5.45GB, max 5.45GB] Consider increasing the JVM heap using `-Xmx` or try a different collector, e.g. `-XX:+UseG1GC`, for better performance.

So we are running out of memory here. sbt is recommending G1 here because it is tuned for high throughput and low pause time (good for when you are coding locally), but it usually achieves that by actually using _more_ memory than the serial and parallel collectors, which may not help us here.

I think we may want to remove the `-Xmx` altogether. Then the JVM will dynamically resize the heap based on whatever github reports it's RAM to be on the agent. 

Current recommendations from Oracle on this are to not explicitly set the GC, Xmx, or Xms at all, unless you really have a specialized usecase.

armanbilge:
> I think we may want to remove the -Xmx altogether. Then the JVM will dynamically resize the heap based on whatever github reports it's RAM to be on the agent.

The tricky thing about this is we want to make sure there's RAM left for external processes, such as starting up browsers for browser-based tests etc. But, I'm happy with any solution that just makes it work :) so, I'll give it a try, remove Xmx and the GC flags.

isomarcte:
@armanbilge 

> The tricky thing about this is we want to make sure there's RAM left for external processes, such as starting up browsers for browser-based tests etc.

We should add in `-XX:MaxRAMPercetage=N` here (N needs to be a double for JRE 8, e.g. `75.0` not `75`). That will allow for a dynamically sizable heap, but still reserve some memory for the other processes. 

Adding `-XX:MaxHeapFreeRatio=X` and `-XX:MinHeapFreeRatio=Y` can also give you very tight memory usage, at the cost throughput, but I'd try it without them first (defaults are 70 and 40) with just `-XXMaxRAMPercentage=N`. 

I don't really know what N should be in this case, but `75.0` is probably as good a place to start as anywhere. 

Thoughts?

armanbilge:
> We should add in -XX:MaxRAMPercetage=N here (N needs to be a double for JRE 8, e.g. 75.0 not 75). That will allow for a dynamically sizable heap, but still reserve some memory for the other processes.

Cool, thanks for pointing out that subtlety about the double. We can try that next and the other flags, depending on how this current run fares.

> Thoughts?

I'm skeptical tbh :P but you seem to have much more knowledge about this than me. Regarding your point about the G1 using more memory: I think that's fine, I think the reason it could be helpful is because it's using a different strategy/algorithm to perform GC. But I have no idea, we'll see! :) Thanks for all your help btw.

armanbilge:
@isomarcte I don't think dynamic detection works, I'm getting this in the logs:
```
In the last 8 seconds, 7.781 (98.5%) were spent in GC. [Heap: 0.20GB free of 1.51GB, max 1.51GB
```

isomarcte:
@armanbilge Sorry, I think you might need to have `-XX:MaxRAMPercentage=75.0` for that to trigger. Would you mind adding that?

armanbilge:
@isomarcte no problem :) can you cancel the previous runners?

isomarcte:
@armanbilge done

isomarcte:
@armanbilge I guess github only gives us 6 GiB of memory?

> In the last 10 seconds, 6.356 (64.8%) were spent in GC. [Heap: 1.30GB free of 2.50GB, max 4.53GB] Consider increasing the JVM heap using `-Xmx` or try a different collector, e.g. `-XX:+UseG1GC`, for better performance.

armanbilge:
@isomarcte I thought it was supposed to be 7 GB:
https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners

armanbilge:
Anyway, this seems to be in trouble again (lots of GC warnings). Compare with my G1 run: https://github.com/http4s/http4s/runs/3274447300#step:11:62

isomarcte:
@armanbilge I didn't realize github only gave us 7GiB. It makes sense then that we made things worse by reducing using 75% of that. 

In any case, if your G1 with -Xmx=6G is working fine, I'm okay if we go with that. I'm sorry I didn't mean to send you on a wild goose chase for memory settings.

armanbilge:
> In any case, if your G1 with -Xmx=6G is working fine, I'm okay if we go with that. I'm sorry I didn't mean to send you on a wild goose chase for memory settings.

Ok, let me revert to that and see if it works, maybe I just got lucky the first time ;) No worries, this was a useful exercise and learning experience, at the very least to try out some of these options. You can cancel the current run.

armanbilge:
@isomarcte yup, G1 works again, don't ask me why! 🙃 It must be the algorithm and something about how `fastOptJS` creates/discards objects.

isomarcte (approved):

Issue #5044, opened 2021-08-08
Further purify the JVM/JS cross


armanbilge:
Now that I better understand the merging challenges that lie ahead (e.g., #5023), I wanted to sketch out some ideas to further improve things post-cross.

1. Crypto. My proposal: create a new library (would org.http4s like to host?) that provides a `Crypto[F]` typeclass. Implementations for JVM, Node.js, and browser platforms will be available with a Cats dependency for JVM and CE 3 dependency for JS. Starting from 0.22, re-implement all cryptography with `Crypto[F]` (its use can be kept internal) and merge this up through 1.0. If executed right, all crypto code will be identical for JVM and JS (i.e., pure cross). Second bird with this stone: remove the copy-pasted Web Crypto facade currently in core.

2. Files: coreJVM already has a dependency on fs2.io. Also add an optional dependency to fs2.io to coreJS. Re-implement StaticFiles and friends using new fs2.io cross-platform file APIs (hopefully we can run this up through 0.23) and un-platform all file-related methods. Critical assumption here (not 100% on this) is that as long as browser-based apps don't touch these file-related methods, they should be fine without the fs2.io dependency.

3. Miscellany: several of the platformed classes in core are simply to provide convenience methods to convert to/from JVM-only net classes. I say drop these in 1.0, those who really need this can always use the conversions available in ip4s (slightly less convenient).

Following 1, 2, and 3, I think it's very likely that core can actually become pure again, or nearly so. More importantly, we'll have scourged platformed classes across several modules that make merges difficult. Thoughts? :)

isomarcte:
> Crypto. My proposal: create a new library (would org.http4s like to host?) that provides a Crypto[F] typeclass. Implementations for JVM, Node.js, and browser platforms will be available with a Cats dependency for JVM and CE 3 dependency for JS. Starting from 0.22, re-implement all cryptography with Crypto[F] (its use can be kept internal) and merge this up through 1.0. If executed right, all crypto code will be identical for JVM and JS (i.e., pure cross). Second bird with this stone: remove the copy-pasted Web Crypto facade currently in core.

Are you suggesting a class `Crypto[F]` which is just a facade for all the operations needed in http4s, or something more general, e.g. classes describing encryption, decryption, hashing/digesting, etc.

[...]

PR #5049, opened 2021-08-08
Implement `fileEncoder` for DOM

armanbilge:
As inspired by the discussion in #5044, this provides an `EntityCoder` in the dom-core module for browser [`File`](https://developer.mozilla.org/en-US/docs/Web/API/File). At the moment, I'm not sure how to test this actually.

Note: this branch is based on #5046.


dom-core/src/main/scala/org/http4s/dom/package.scala
```diff
@@ -16,13 +16,19 @@
...
 import org.scalajs.dom.experimental.{Headers => DomHeaders}
 
 import scala.scalajs.js.JSConverters._
+import scala.scalajs.js.typedarray.Uint8Array
+import cats.effect.kernel.Resource
 
 package object dom {
 
-  def toDomHeaders(headers: Headers): DomHeaders =
+  private[dom] def toDomHeaders(headers: Headers): DomHeaders =
```
armanbilge, line 31: I had originally meant to keep these private, hope it's okay if I sneak it in here :(

rossabaker: Totally.  No compatibility is promised on 1.0 until 1.0.

dom-core/src/main/scala/org/http4s/dom/DomEntityEncoder.scala
```diff
@@ -0,0 +1,45 @@
...
+          // Unfortunately stream() method is missing from the File facade
+          F.delay(file.asInstanceOf[js.Dynamic].stream().asInstanceOf[ReadableStream[Uint8Array]])
+        } { case (rs, exitCase) => closeReadableStream(rs, exitCase) }
+        .flatMap(readableStreamToStream[F])
+    }
+
+  def readableStreamEncoder[F[_]: Async]: EntityEncoder[F, ReadableStream[Uint8Array]] =
+    EntityEncoder.entityBodyEncoder.contramap { rs =>
+      readableStreamToStream(rs)
+    }
```
rossabaker, line 43: Should these be implicit?

armanbilge: Not sure, should they? I modelled off of https://github.com/http4s/http4s/blob/main/core/jvm/src/main/scala/org/http4s/EntityEncoderPlatform.scala which are not implicit.

rossabaker: Oh, I think the others aren't because they used to take a `Blocker`.  I think they could be starting in 0.23.  I'll spin off a separate issue for that.

rossabaker: #5072

rossabaker: Let's go ahead and make them implicit here, and then we can close this out.

armanbilge: Cool, I'll fix it right up!

rossabaker (approved):
:+1: on green

PR #5051, opened 2021-08-09
Allow empty `Access-Control-{Allow, Expose}-Headers`

bplommer:
Per the [fetch spec](https://fetch.spec.whatwg.org/#http-new-header-syntax), these headers can be empty. Unfortunately I don't think binary compatibility can be preserved.

### DIFF
--- a/core/shared/src/main/scala/org/http4s/headers/Access-Control-Allow-Headers.scala
+++ b/core/shared/src/main/scala/org/http4s/headers/Access-Control-Allow-Headers.scala
@@ -20,17 +20,19 @@ package headers
 import org.http4s.Header
 import org.typelevel.ci._
 import org.http4s.internal.parsing.Rfc7230
-import cats.data.NonEmptyList
 
 object `Access-Control-Allow-Headers` {
-  def apply(head: CIString, tail: CIString*): `Access-Control-Allow-Headers` =
-    apply(NonEmptyList(head, tail.toList))
+  def apply(values: CIString*): `Access-Control-Allow-Headers` =
+    apply(values.toList)
+
+  val empty: `Access-Control-Allow-Headers` = `Access-Control-Allow-Headers`(Nil)
 
   def parse(s: String): ParseResult[`Access-Control-Allow-Headers`] =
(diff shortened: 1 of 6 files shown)


bplommer:
See also https://github.com/mdn/content/pull/7758

core/shared/src/main/scala/org/http4s/headers/Access-Control-Allow-Headers.scala
```diff
@@ -39,8 +41,8 @@ object `Access-Control-Allow-Headers` {
       parse
     )
 
-  implicit val headerSemigroupInstance: cats.Semigroup[`Access-Control-Allow-Headers`] =
-    (a, b) => `Access-Control-Allow-Headers`(a.values.concatNel(b.values))
+  implicit val headerSemigroupInstance: cats.Monoid[`Access-Control-Allow-Headers`] =
```
rossabaker, line 44: We could go ahead and rename this for accuracy, since we're breaking compatibility.

rossabaker (commented):
Yeah, I don't se a way to do this in a compatible fashion either, short of moving to a new header name, but this one is taking up the prime real estate.  We could also leave a scaladoc about their deficiency in 0.22 and 0.23.

core/shared/src/main/scala/org/http4s/headers/Access-Control-Allow-Headers.scala
bplommer: Yeah, I totally missed that

bplommer:
> Yeah, I don't se a way to do this in a compatible fashion either, short of moving to a new header name, but this one is taking up the prime real estate. We could also leave a scaladoc about their deficiency in 0.22 and 0.23.

We *could* make the change temporary by switching to a new name in 0.22/0.23 with the current one deprecated, then going back to the current one with the new one deprecated in 1.0 - not sure it's worth it though, at least unless/until we want to start using the modelled headers in the cors middleware.

rossabaker:
I'm reluctant to make two breaking changes to the `-Xfatal-warnings` crowd with the deprecate-and-undeprecate dance.  Especially because the CORS middleware is the most likely client of these types, and we can work around it inside those.

PR #5056, opened 2021-08-10
Check for allowed entity in GZip server middleware

ybasket:
If a response has a status that doesn't allow any entity, the GZip server middleware should not handle it because the injected Content-Encoding header is semantically questionable and not handled well by all clients.


rossabaker (commented):
Thanks!  This seems like a good change.  Could we throw in a quick unit test, too?

I'll backport this to 0.22, unless you want to reopen against that branch.

ybasket:
> Thanks! This seems like a good change. Could we throw in a quick unit test, too?
> 
> I'll backport this to 0.22, unless you want to reopen against that branch.

Just added a unit test. And opened an identical PR against the 0.22 series branch (hope that's what you meant): https://github.com/http4s/http4s/pull/5060

rossabaker:
Yes, that's perfect.  We like to originate changes on the series branches when they're compatible, so people can use them sooner on stable releases.  I can always cherry-pick back, but PRs like #5060 save me time.

PR #5062, opened 2021-08-12
Add query param matcher that returns 400 when param is missing or absent

bplommer:
The motivation for this is that properly handling missing or absent query parameters is currently cumbersome enough that people usually don't bother and just let the request fall through to a 404. On the other hand, I'm not thrilled about the extra indirection created by the solution in this PR - feedback appreciated!


rossabaker:
Confession: I have never used any of our query parameter extractors aside from code reviews!  I speak from ignorance.

It looks like you're trying to avoid this pattern:

```scala
    case GET -> Root / "valid" :? ValidatingCounter(c) =>
      c.fold(
        errors => BadRequest(errors.map(_.sanitized).mkString_("", ",", "")),
        vc => Ok(s"counter: $vc")
      )
```

1. What do you want to do if you have two missing parameters?  Are you just going to fail fast, or do you aspire to accumulate them?
2. You've decided to make a 400 with a String body.  I alluded to this problem elsehwere: what about people who want exclusively JSON errors?  Can't even satisfy them, because they might need a different envelope for them.

I like your pattern, but it feels like we need some sort of `Validated[E, A] => F[Response[F]]`, and there's not a universal solution to the error rendering.  It's the same basic reason we don't have an `Option[A] => F[Response[F]]` helper out of the box.

rossabaker:
This is another one that could go to 0.22 if it goes, unless it changes substantially.

bplommer:
> 1. What do you want to do if you have two missing parameters?  Are you just going to fail fast, or do you aspire to accumulate them?

Actually I think I have a better solution for this - will push it to this PR in a minute.

> 2. You've decided to make a 400 with a String body.  I alluded to this problem elsehwere: what about people who want exclusively JSON errors?  Can't even satisfy them, because they might need a different envelope for them.

I have a couple of thoughts on this - firstly, the nature of errors is such that it's impossible to provide a general guarantee about the format of error messages. If your server goes down and the proxy in front of it starts returning 502 errors, those are unlikely be in your preferred JSON format either. If client code contains a typo in the authority portion of the URI, it definitely won't be. This is an opt-in helper utility intended to provide a low-friction way of providing reasonable, spec-compliant error responses - if people have more specific requirements they're still welcome and encouraged to use more low-level facilities to implement them.

On a more constructive/conciliatory note - I wonder how complicated it would be to use an `EntityEncoder` to encode `BadQuery`, with a default provided but able to be overridden.

dsl/src/main/scala/org/http4s/dsl/impl/Path.scala
```diff
@@ -384,3 +384,49 @@ abstract class OptionalValidatingQueryParamDecoderMatcher[T: QueryParamDecoder](
...
+object BadQuery {
+  final case class MissingRequiredParam(param: String) extends BadQuery
+  final case class InvalidParam(param: String, errors: NonEmptyList[ParseFailure]) extends BadQuery
+
+  implicit class ops[A](self: ValidatedNel[BadQuery, A]) {
+    def respondWith[F[_]: Applicative](f: A => F[Response[F]]) = {
+      implicit val dsl: Http4sDsl[F] = org.http4s.dsl.Http4sDsl[F]
+      import dsl._
+
+      self.fold(errors => BadRequest(errors.toString), f)
```
bplommer, line 399: If we go with this solution I'll obviously replace this with proper rendering.

dsl/src/test/scala/org/http4s/dsl/PathInHttpRoutesSuite.scala
```diff
@@ -71,8 +72,8 @@ class PathInHttpRoutesSuite extends Http4sSuite {
       Ok(s"list: ${l.mkString(",")}, start: $s, limit: ${m.l}, term: $t, decimal=$d")
     case GET -> Root / "app" :? OptCounter(c) =>
       Ok(s"counter: $c")
-    case GET -> Root / "required" :? RequiredCounter(withCount) =>
-      withCount(vc => Ok(s"counter: $vc"))
+    case GET -> Root / "required" :? RequiredCounter(c) :? RequiredLimit(l) =>
+      (c, l).tupled.respondWith { case (vc, vl) => Ok(s"counter: $vc, limit: $vl") }
```
rossabaker, line 76: I'm way of ad hoc syntactic extensions, but trying to think of an alternative.  I like that you solved the composition problem.

Thinking out loud, and not compiling:

```scala
respondValid((c, l).mapN { case (vc, vl) => Ok(s"counter: $vc, limit: $vl") })
```

rossabaker (commented):
People have complained, but I think your points on generic error handling are good.

dsl/src/main/scala/org/http4s/dsl/impl/Path.scala
```diff
@@ -385,6 +385,22 @@ abstract class OptionalValidatingQueryParamDecoderMatcher[T: QueryParamDecoder](
     }
 }
 
+sealed abstract class BadQuery
+
+object BadQuery {
+  final case class MissingRequiredParam(param: String) extends BadQuery
+  final case class InvalidParam(param: String, errors: NonEmptyList[ParseFailure]) extends BadQuery
+
+  implicit class ops[A](self: ValidatedNel[BadQuery, A]) {
```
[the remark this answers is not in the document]
bplommer: > Does BadQuery being a type param get BadQuery's companion into implicit scope? 

I think so, yeah. Actually, we should be able to do something similar with `Message` to avoid needing an import for the `Kleisli` syntax.

PR #5063, opened 2021-08-12
DSL: matcher for handling all methods on a resource

bplommer:
Similar to #5062 - intended to provide an ergonomic way of returning the [correct status codes](https://datatracker.ietf.org/doc/html/rfc7231#section-4.1) when a resource is requested with an unsupported method.


bplommer:
CI failure is an unrelated Blaze flake.

dsl/src/main/scala/org/http4s/dsl/impl/Path.scala
```diff
@@ -98,6 +98,40 @@ object -> {
...
+                if (allMethods.contains(method)) {
+                  Response(
+                    status = Status.MethodNotAllowed,
+                    headers = Headers(Allow(allMethods.filter(pf.isDefinedAt))))
+                } else { Response(status = Status.NotImplemented) }
+              }
+          ),
+        req.pathInfo)
+
+    }
```
rossabaker, line 132: I like scalafmt, but if I didn't, this would be in my slide deck. 😆

dsl/src/main/scala/org/http4s/dsl/impl/Path.scala
```diff
@@ -98,6 +98,40 @@ object -> {
...
+  /** Extractor to match an http resource and then enumerate all supported methods:
+    * {{{
+    *   (request.method, Path(request.path)) match {
+    *     case withMethod ->> Root / "test.json" => withMethod {
+    *       case Method.GET => ...
+    *       case Method.POST => ...
+    * }}}
+    *
+    * Returns an error response if the method is not matched, in accordance with [[https://datatracker.ietf.org/doc/html/rfc7231#section-4.1 RFC7231]]
+    */
```
rossabaker, line 113: Would it be better to put this doc on the object?  I'm not sure what makes it more visible in tooling.

dsl/src/test/scala/org/http4s/dsl/PathInHttpRoutesSuite.scala
```diff
@@ -89,6 +90,11 @@ class PathInHttpRoutesSuite extends Http4sSuite {
     case GET -> Root / "flagparam" :? Flag(flag) =>
       if (flag) Ok("flag present")
       else Ok("flag not present")
+    case withMethod ->> Root / "resource" =>
```
rossabaker, line 93: I kind of want to call it `resource`, but this probably overloads Cats-Effect too much.

dsl/src/test/scala/org/http4s/dsl/PathInHttpRoutesSuite.scala
```diff
@@ -294,4 +300,25 @@ class PathInHttpRoutesSuite extends Http4sSuite {
     response.map(_.status).assertEquals(Ok) *>
       response.flatMap(_.as[String]).assertEquals("flag not present")
   }
+  test("Path DSL with -->> should match on GET") {
```
rossabaker, line 303: The new operator is `->>`, but the tests refer to `-->>`.

rossabaker (commented):
This special cases nesting method handlers in a way that's more generically handled by something like the akka-http DSL, but this special case _is_ prominent in RESTful development.

This DSL was only ever supposed to be a proof-of-concept, but since it's become a bit of a default, it might as well be good. :+1:

rossabaker:
I see no reason we can't cherry-pick this back to series/0.22.

dsl/src/main/scala/org/http4s/dsl/impl/Path.scala
bplommer: Not sure. I copied the pattern from the other matchers, but I can look into it. Generally I don't think extractors play great with tooling.

bplommer:
> This special cases nesting method handlers in a way that's more generically handled by something like the akka-http DSL, but this special case _is_ prominent in RESTful development.
> 
> 
> 
> This DSL was only ever supposed to be a proof-of-concept, but since it's become a bit of a default, it might as well be good. :+1:

There's a few places where the DSL is the only part of HTTP4s that encodes knowledge of how to generate spec-compliant requests and responses - I think it would be worth having that knowledge encoded in the core module, so that the DSL is purely syntactic sugar.

bplommer:
> I see no reason we can't cherry-pick this back to series/0.22.

Yeah, I tried to make this PR against 0.22 but I messed something up and got a bunch of scalafmt errors.

dsl/src/main/scala/org/http4s/dsl/impl/Path.scala
bplommer: It was never gonna be pretty but I like the fact that scalafmt makes that not be my fault.

bplommer: By the way, I'm not super happy that the Allow header has to be regenerated on every match failure but I can't see a way around that as long as the DSL is matcher-based.

dsl/src/test/scala/org/http4s/dsl/PathInHttpRoutesSuite.scala
bplommer: Oops, that's what I get for using coding ligatures.

dsl/src/test/scala/org/http4s/dsl/PathInHttpRoutesSuite.scala
bplommer: Instead of `withMethod` you mean?

Yeah, I also don't like it that `withMethod` has to be written twice, and that it feels like part of the DSL syntax but isn't.

Here's another idea - have the first item returned by the matcher be the entire request, then make `withMethod` (or maybe `respondByMethod`?) be an extension method on `Request`, so the usage looks something like
```scala
case req ->> Root / "foo" =>
  req.respondByMethod {
    case GET => ...
```
This also has the advantage of being more general, in that you can do arbitrary things with `req`. Obviously you can do that already by prefixing the matcher with `req @`, but that always feels kind of like a nasty workaround.

rossabaker:
> There's a few places where the DSL is the only part of HTTP4s that encodes knowledge of how to generate spec-compliant requests and responses - I think it would be worth having that knowledge encoded in the core module, so that the DSL is purely syntactic sugar.

Do you mean things like required headers for certain status codes, some status codes not taking a body?  It's not so hard to make constructors aware of these restrictions, but it's hard to keep enforced.  Without lifting status codes to the type level, the compiler can't catch that `response.withEntity(...)` is naughty when `response.status === Status.NoContent`.

isomarcte (approved):

rossabaker (approved):
I'll leave this one open in case @bplommer wants to advocate for his alternate definition of `->>` before it gets baked into 0.22, but I'm satisfied with this as it is.

bplommer:
> Do you mean things like required headers for certain status codes, some status codes not taking a body? 

Yes, but I was thinking of it as still being an opt-in utility rather than something that's enforced by default. That said:

> Without lifting status codes to the type level,

With Scala 2.13's singleton Integer types that doesn't seem so far-fetched. Is there a timeline for dropping Scala 2.13 support?

bplommer:
I think I'm happy to go ahead with this as is - I'll just fix the test naming and we should be good to go.

rossabaker:
> With Scala 2.13's singleton Integer types that doesn't seem so far-fetched.

I'd be interested to see a proof of concept, but I suspect that would cause a massive ripple of type parameters.

> Is there a timeline for dropping Scala 2.13 support?

If we have an opportunity for enhancements that support Cats 3, I think we should embrace those.  And I'm sure we'll get there in some yet unplanned Cats 3 / Cats-Effect 4 epic.  But I work at a large Scala shop, and getting to Scala 3 is a mountain obscured behind couple other large mountains.

bplommer:
Yee, did I say dropping Scala 2.13? I meant dropping Scala 2.12.

rossabaker:
Ah, 2.12 comes up more often.  The last time I looked at stats, Scala 2.12 was a firm majority of our downloads, and most Scala shops I've worked at were a version or two behind, and pushing too hard made people wonder "Why Scala?" more than "Let's upgrade."  I've sparred with upstream maintainers on it, but I've traditionally been willing to put a little extra effort into supporting old versions.

That said, when it really starts to hold back progress, it's time.  I don't want to drop it for 1.0 without a feature held back by it, but with reason, I'd be open.

bplommer:
> I'd be interested to see a proof of concept, but I suspect that would cause a massive ripple of type parameters.

I think there's a lot you can do in a source-compatible or even binary-compatible way with type members - I've started hacking on something because clearly I don't yet have enough open GitHub issues in my life...

> If we have an opportunity for enhancements that support Cats 3, 

Wait there's a cats 3 in development? Or did you mean Scala 3?

rossabaker:
OMG, Scala 3!!!  I've poured cold water on Cats 3 a few times!  Corrected above. 😅

rossabaker:
I took the liberty of fixing the test names to get this off the board. You're working on plenty of more interesting things!

rossabaker:
Gack.  I meant to cherry-pick this one.

PR #5064, opened 2021-08-12
Add Conservative Retry For Dead Connections

### LINKED ISSUE #4935: Ember Client sometimes failing with `org.http4s.ember.core.EmptyStreamError: Cannot Parse Empty Stream`
I discussed this on Discord with @RaasAhsan here a while back: https://discord.com/channels/632277896739946517/632286375311573032/844933274890534974

This same error was raised in another service with a different 3rd party client, using http4s-ember-client 1.0.0-M23. I've decided to revert back to using Blaze in this case for the time-being, but wanted to raise the issue here. 

It was suggested in the discord thread that this wasn't occurring in other client implementations due to some baked-in retry behaviour that ember is perhaps lacking

ChristopherDavenport:
Fixes #4935

Uses the conservative firefoxy strategy for retrying connections that are dead from the pool.

### DIFF
--- a/ember-client/src/main/scala/org/http4s/ember/client/internal/ClientHelpers.scala
+++ b/ember-client/src/main/scala/org/http4s/ember/client/internal/ClientHelpers.scala
@@ -37,6 +37,7 @@ import _root_.io.chrisdavenport.keypool._
 import javax.net.ssl.SNIHostName
 import org.http4s.headers.{Connection, Date, `User-Agent`}
 import _root_.org.http4s.ember.core.Util.durationToFinite
+import java.nio.channels.ClosedChannelException
 
 private[client] object ClientHelpers {
 
@@ -130,6 +131,10 @@ private[client] object ClientHelpers {
       processedReq <- preprocessRequest(request, userAgent)
       res <- writeRead(processedReq)
     } yield res
+  }.adaptError{
+    case e@org.http4s.ember.core.EmptyStreamError() => new ClosedChannelException(){
+      initCause(e)
+    }
   }
 
   private[internal] def preprocessRequest[F[_]: Monad: Clock](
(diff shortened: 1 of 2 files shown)


ember-client/src/main/scala/org/http4s/ember/client/internal/ClientHelpers.scala
```diff
@@ -191,4 +196,28 @@ private[client] object ClientHelpers {
               new java.net.SocketException("Fresh connection from pool was not open")))
         )
     }
+
+  private[ember] object RetryLogic {
+    import org.http4s.client.middleware._
+    import org.http4s.syntax.all._
+    // import org.http4s.headers.`Idempotency-Key`
```
isomarcte, line 203: Any reason you left this here?

ember-client/src/main/scala/org/http4s/ember/client/internal/ClientHelpers.scala
```diff
@@ -191,4 +196,31 @@ private[client] object ClientHelpers {
...
+
+    private val retryNow = 0.seconds.some
+    def retryUntilFresh[F[_]]: RetryPolicy[F] = { (req, result, retries) =>
+      if (emberDeadFromPoolPolicy(req, result) && retries <= 2) retryNow
```
isomarcte, line 207: Should this be `retries < 2`? I thought @rossabaker said only retry once. Won't this retry twice?

ember-client/src/main/scala/org/http4s/ember/client/internal/ClientHelpers.scala
```diff
@@ -130,6 +131,10 @@ private[client] object ClientHelpers {
       processedReq <- preprocessRequest(request, userAgent)
       res <- writeRead(processedReq)
     } yield res
+  }.adaptError { case e @ org.http4s.ember.core.EmptyStreamError() =>
+    new ClosedChannelException() {
+      initCause(e)
```
isomarcte, line 136: Can you provide a `getMessage` here? It would be so very nice.

isomarcte (approved):

ember-client/src/main/scala/org/http4s/ember/client/internal/ClientHelpers.scala
ChristopherDavenport: Will be used in the next version.

ember-client/src/main/scala/org/http4s/ember/client/internal/ClientHelpers.scala
ChristopherDavenport: I believe this is 1 indexed. 

https://github.com/http4s/http4s/blob/main/client/shared/src/main/scala/org/http4s/client/middleware/Retry.scala#L99

PR #5065, opened 2021-08-12
Merge from series/0.23

rossabaker:
This ~~is~~ was mostly #5023.  That got ported to JavaScript, and then a substantial rework was initiated on series/0.22.


rossabaker:
I don't have a working node environment to run the Scala.JS tests locally.  I'm getting a bunch of failures like this:

```
[info] Fast optimizing /Users/ross.baker/src/http4s/client/js/target/scala-2.12/http4s-client-test-fastopt
[error] Referring to non-existent class java.net.URLEncoder
[error]   called from private org.http4s.client.oauth1.OAuthSuite.$anonfun$signRequestWith$3(java.lang.String,java.lang.String,org.http4s.Request)void
[error]   called from org.http4s.client.oauth1.OAuthSuite.signRequestWith(org.http4s.client.oauth1.ProtocolParameter$SignatureMethod,java.lang.String,java.lang.String)cats.effect.IO
[error]   called from private org.http4s.client.oauth1.OAuthSuite.$anonfun$new$22()cats.effect.IO
[error]   called from constructor org.http4s.client.oauth1.OAuthSuite.<init>()void
[error]   called from static constructor org.http4s.client.oauth1.OAuthSuite.<stinit>()void
[error]   called from core module analyzer
[error] involving instantiated classes:
[error]   org.http4s.client.oauth1.OAuthSuite
```

armanbilge:
@rossabaker It means the code designated by the trace, which you are compiling with Scala.js is using the `java.net.URLEncoder` class which is not available on JS. So, that code either needs to be replaced with something cross-platform or split into a JVM and JS implementation.

armanbilge:
If this is #5023, this will need a second parallel implementation in JS most likely using the Web Crypto facade. I anticipated this which is why I opened #5044, if we can wait on this merge I really think that's the way to go.

BTW feel free to @ me whenever JS issues come up, so I don't miss it :)

rossabaker:
So the next step in getting this unblocked is #5044?  We've got an exciting release train coming this week, and I would hate to leave 1.0 behind.  We could conceivably leave the oauth1 out if everything else is ready.

armanbilge:
Yes, #5044 is ideal. Alternatively, we could leave oauth out, or if possible only offer it to JVM, or if someone is brave provide a parallel implementation in JS.

Re #5044, it seemed like there was general support for starting a new typelevel crypto lib?

rossabaker:
I don't understand the full scope of a new Typelevel crypto lib, and I'm stretched too thin to volunteer on it, but it sounds reasonable to me.

I doubt anyone is doing oauth1 on Scala.js, and it's just milestones.  That's probably the right way to unblock this so it can be released with everything else.

armanbilge:
> I don't understand the full scope of a new Typelevel crypto lib, and I'm stretched too thin to volunteer on it, but it sounds reasonable to me.

If someone sets up the repo I'll have a crack at it. @isomarcte seemed interested too, and maybe @ChristopherDavenport since he also has to figure out what to do with tsec.

rossabaker:
If someone gives me a name, I can create a repo.

armanbilge:
@rossabaker I propose "bobcat" (or bobcats?), as in Alice and Bob 😺

rossabaker:
Placeholder, to which you're invited: https://github.com/typelevel/bobcats

rossabaker:
It's failing now on adapting Ember.EmptyStream to a `ClosedChannelException`, which doesn't exist in Scala.js.  Is the general idea to put type aliases in a platform for these circumstances?

armanbilge:
@rossabaker taking a look now, but yes I think it almost always gets addressed via a type alias.

armanbilge:
@rossabaker I fixed it via a type alias to just an ordinary `IOException` on JS for now. The long-term goal is to migrate all these exceptions to `fs2.io` (e.g. `fs2.io.ClosedChannelException`) which aims to guarantee identical error semantics on JVM and JS. Is there a test that reproduces a `ClosedChannelException`? We need to see what the equivalent error is on Node.js so we can handle it.

ember-client/js/src/main/scala/org/http4s/ember/client/internal/ClientHelpersPlatform.scala
```diff
@@ -27,6 +27,8 @@ import com.comcast.ip4s.IDN
 
 private[internal] trait ClientHelpersPlatform {
 
+  private[internal] type ClosedChannelException = fs2.io.IOException
```
rossabaker, line 30: TIL `fs2.io.IOException`

armanbilge: 😆 FYI it's actually just an alias to `java.io.IOException` which _is_ supported on JS, but it's the parent to a bunch of `fs2.io`-only implementations of `java.io` exceptions for Scala.js.

rossabaker: Yeah, I was just wondering whether fs2 should have a `ChannelClosedException`.

rossabaker:
In 2157619, I eliminated the DSL `BinCompat` traits, which surprisingly caused a conflict.  But we don't need BinCompat baggage in 1.0 anyway.

ember-client/js/src/main/scala/org/http4s/ember/client/internal/ClientHelpersPlatform.scala
armanbilge: Hmm, why/why not?

rossabaker: If that's the parallel hierarchy for JavaScript implementations of IO failures.  Note that this one is specific about what went wrong, but the JavaScript will just be a generic IOException.

armanbilge: Ah, yes, we can definitely add it. 👍

rossabaker:
I'm having trouble finding some of those build errors.

I installed node and tried running `server/test` locally for the first time.  It brought my computer to its knees with a crapton of Node processes, and a bunch of errors like this, which I don't think are happening in CI:

```
==> X org.http4s.server.middleware.authentication.AuthenticationSuite.DigestAuthentication should avoid many concurrent replay attacks 0.45s scala.scalajs.js.JavaScriptException: TypeError: Cannot read property 'getRandomValues' of undefined
```

armanbilge:
@rossabaker What's your Node version? You need at least Node 15. Looks like you are on an old one.

rossabaker:
I was on 14.x.  I'm now running on 16.4.1, and it's still bringing my 16GB RAM Macbook to its knees.  I'm also still seeing:

```
==> X org.http4s.server.middleware.CSRFSuite.not validate different tokens 0.16s scala.scalajs.js.JavaScriptException: TypeError: Cannot read property 'subtle' of undefined
```

My GraalVM irritatingly comes with Node 12.  I had to do some tricks to get Node 16 on my $PATH, and I can't say for sure which SBT is finding.

rossabaker:
```
sbt:http4s> consoleQuick
[info] Starting scala interpreter...
Welcome to Scala 2.12.14 (OpenJDK 64-Bit Server VM GraalVM CE 21.0.0, Java 1.8.0_282).

scala> import sys.process._
import sys.process._

scala> "node --version".!
v12.20.1
res0: Int = 0

scala> 
[success] Total time: 26 s, completed Aug 30, 2021 2:55:04 PM
sbt:http4s> 

[nix-shell:~/src/http4s]$ node --version
v16.4.1
```

:hankey:

armanbilge:
Sorry for the pain 😕 I'll relax the Node 16 requirement in bobcats, there are some new APIs in Node 16 that were the quickest ways to get crypto working on both Node.js and browsers in the original cross. 

> and it's still bringing my 16GB RAM Macbook to its knees.

This I have no idea about ...

rossabaker:
I got the node 16 working on the last commit by ditching the Graal JVM.  It runs much more reasonably now, and I'm down to one error, which I'm pretty sure is not CI's problem:

```
==> X org.http4s.server.middleware.ChunkAggregatorSuite.handle chunks 0.30s java.time.zone.ZoneRulesException: Unknown time-zone ID: America/Indianapolis
```

armanbilge:
Ah, that's probably to do with the fact that Scala.js comes with a severely crippled locale database. @cquiroz knows about these things.

rossabaker:
It looks like it's coming out of constructing `ArbitraryInstances`.  Why it happens there, and eats that test, I'm not sure.

armanbilge:
Maybe one of the arbitraries is generated from your local time zone?

rossabaker:
That's definitely true.  I'm trying to figure out why `server/test` is failing on CI.  I don't see the telltale `==> x` syntax and am having trouble spotting the failure in the logs.

cquiroz:
Just a comment about graalvm, it comes with a node runtime that badly
Interacts with scala.js

I simply delete those binaries to let the system node take its place

rossabaker:
Oh, CORS I guess?

```
2021-08-30T19:22:21.9003460Z [0m[[0m[31merror[0m] [0m[0mFailed tests:[0m
2021-08-30T19:22:21.9006591Z [0m[[0m[31merror[0m] [0m[0m	org.http4s.server.middleware.CORSSuite
```

It just shows up like an empty test suite, like several others, though.  WTF.

armanbilge:
Yes, you can see it in CI here: https://github.com/http4s/http4s/pull/5065/checks?check_run_id=3465324316#step:13:336

cquiroz:
@rossabaker This may require you to fix the timezone, maybe passing as a jvm param `-Duser.timezone=UTC`
Otherwise you'd need to generate a local tz db which is probably overkill for this case

> I got the node 16 working on the last commit by ditching the Graal JVM. It runs much more reasonably now, and I'm down to one error, which I'm pretty sure is not CI's problem:
> 
> ```
> ==> X org.http4s.server.middleware.ChunkAggregatorSuite.handle chunks 0.30s java.time.zone.ZoneRulesException: Unknown time-zone ID: America/Indianapolis
> ```

server/shared/src/test/scala/org/http4s/server/middleware/CORSSuite.scala
```diff
@@ -125,8 +125,8 @@ class CORSSuite extends Http4sSuite {
     val req = buildRequest("/bar").withHeaders("Origin" -> "http://blah.com/")
     cors2
       .orNotFound(req)
-      .map(resp => resp.status.code == 403)
-      .assert
+      .map(_.status)
+      .assertEquals(Status.Forbidden)
```
armanbilge, line 129: What was this about?

rossabaker: That's a test I found failing, and I don't know what status it's actually returning, and I can't make it fail locally.

armanbilge: Humph, really weird. Maybe some weird numerics thing (all numbers in JS are represented as `Double` so...). I'll try and have a closer look at it later.

armanbilge: Oh, unless I misunderstood, you're not sure if this change actually makes the test pass?

rossabaker: As it was, it was passing on some builds and failing on others, including on a JVM, and I don't have any idea WTF is going on.

rossabaker: Something stinks in the output.

```
org.http4s.server.middleware.CORSSuite:
```

with no successes (`+`) or failues (`==> x`) reported under it.  And the only other mention is 

```
[error] Failed: Total 253, Failed 1, Errors 0, Passed 252
[error] Failed tests:
[error] 	org.http4s.server.middleware.CORSSuite
```

I wonder whether `ErrorReporting` is doing something wicked.

armanbilge: How are you running the tests? Try something like `serverJS/testOnly *CORSSuite`?

rossabaker: Yes. It's happening on the JVM, too.  I'm trying to minimize it.  It's not just CORS, either.

Output of `sbt 'tests/testOnly *Method*'` on `main`:

```
org.http4s.MethodSuite:
  + Hash[Method]: hash.antisymmetry eq 0.09s
  + Hash[Method]: hash.hash compatibility 0.01s
  + Hash[Method]: hash.reflexivity eq 0.005s
  + Hash[Method]: hash.same as scala hashing 0.007s
  + Hash[Method]: hash.same as universal hash 0.006s
  + Hash[Method]: hash.symmetry eq 0.003s
...
[info] Passed: Total 32, Failed 0, Errors 0, Passed 32
[success] Total time: 4 s, completed Aug 30, 2021 4:55:20 PM
```

Output on this branch:

```
org.http4s.MethodSuite:
[info] Passed: Total 32, Failed 0, Errors 0, Passed 32
[success] Total time: 49 s, completed Aug 30, 2021 4:57:33 PM
```

rossabaker: munit got upgraded to 0.7.28.  It's the [buffered logging](https://github.com/scalameta/munit/releases/tag/v0.7.28).

armanbilge: Ahhhhh. No wonder test output have been so weird lately.

rossabaker: Are you seeing it in other projects?  I opened https://github.com/scalameta/munit/issues/408.  I haven't had time to try it on a smaller project.

armanbilge: The buffering is very noticeable for large test suites, that's where I've noticed it most. Also, I've seen disappearing test output sometimes, but that pre-dates this change and only occurs in CI that runs both JVM and JS suites simultaneously.

rossabaker:
Okay, the test output is back.  It's failing on a CORS fallthrough.  I immediately suspected https://github.com/http4s/http4s/pull/5071, but that's not merged yet.

rossabaker:
Fails locally the same way after I merge main.  It's the Origin parser!

Issue #5069, opened 2021-08-12
Accept-Language parser fails on wildcard '*' in 0.22


shannonbarnes:
```
import org.http4s.headers.`Accept-Language`

`Accept-Language`.parse("fr-CH, fr;q=0.9, en;q=0.8, de;q=0.7, *;q=0.5")
```
Header example from [developer mozilla](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Language)

Version: `0.22.2` - Parse Failure
 `res0: org.http4s.ParseResult[org.http4s.headers.Accept-Language] = Left(org.http4s.ParseFailure: Invalid Accept-Language header: Error(0,NonEmptyList(InRange(0,A,Z), InRange(0,a,z))))`

Version `0.21.24` - Works
`res0: org.http4s.ParseResult[org.http4s.headers.Accept-Language] = Right(Accept-Language: fr-CH, fr;q=0.9, en;q=0.8, de;q=0.7)`

rossabaker:
Thank you.  Those parsers all had to be rewritten in the 0.22 migration.  We'll add this as a test case and get a fix out ASAP.

shannonbarnes:
I saw this tweeted out a few days ago https://github.com/http4s/http4s/issues/5010 and thought about giving it a try. 
I can give a stab and fixing this one.

rossabaker:
Fantastic.  Please speak up if you need any help!

PR #5070, opened 2021-08-13
Fix for #5069

shannonbarnes:
https://github.com/http4s/http4s/issues/5069
Language-Accept parser having trouble with wild-card and more than one sub-tag

### DIFF
--- a/laws/src/main/scala/org/http4s/laws/discipline/ArbitraryInstances.scala
+++ b/laws/src/main/scala/org/http4s/laws/discipline/ArbitraryInstances.scala
@@ -103,6 +103,9 @@ private[http4s] trait ArbitraryInstances {
   val genToken: Gen[String] =
     nonEmptyListOf(genTchar).map(_.mkString)
 
+  val genAlphaToken: Gen[String] =
+    nonEmptyListOf(alphaChar).map(_.mkString)
+
   val genNonTchar = frequency(
     4 -> oneOf(Set(0x00.toChar to 0x7f.toChar: _*) -- tchars),
     1 -> oneOf(0x100.toChar to Char.MaxValue)
@@ -112,7 +115,7 @@ private[http4s] trait ArbitraryInstances {
     a <- stringOf(genTchar)
     b <- nonEmptyListOf(genNonTchar).map(_.mkString)
     c <- stringOf(genChar)
-  } yield (a + b + c)
+  } yield a + b + c
 
   val genVchar: Gen[Char] =
     oneOf('\u0021' to '\u007e')
@@ -358,8 +361,8 @@ private[http4s] trait ArbitraryInstances {
   def genLanguageTagNoQuality: Gen[LanguageTag] =
     frequency(
       3 -> (for {
-        primaryTag <- genToken
-        subTags <- frequency(4 -> Nil, 1 -> listOf(genToken))
+        primaryTag <- genAlphaToken
+        subTags <- frequency(4 -> Nil, 1 -> listOf(genAlphaToken))
       } yield LanguageTag(primaryTag, subTags = subTags)),
       1 -> const(LanguageTag.`*`)
     )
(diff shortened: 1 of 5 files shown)


laws/src/main/scala/org/http4s/laws/discipline/ArbitraryInstances.scala
```diff
@@ -103,6 +103,9 @@ private[http4s] trait ArbitraryInstances {
   val genToken: Gen[String] =
     nonEmptyListOf(genTchar).map(_.mkString)
 
+  val genAlphaToken: Gen[String] =
```
rossabaker, line 106: I wish scalacheck had a `nonEmptyStringOf` generator.

rossabaker (approved):
This is nice work! :+1: on green.

rossabaker:
This also introduces some binary compatibility issues in the arbitraries.  I think we can work around it by adding the new generator to the companion object, and then we can move it back where it belongs when we eventually merge it to main.

shannonbarnes:
First PR and bit unsure what is exactly the issue is and what I need to to do. Is this a binary incompatibility between scala versions or upstream series?  I could revert the added header law check.

rossabaker:
Great question.  We are checking for binary against previous http4s releases with the same Scala version.  If this is more than you want to read, I can finish this job.  If you'd like to see it through, read on!

First, let's fetch the tags.  You probably have them already, but let's be sure.  The build uses those to calculate the previous versions:

```sh
git fetch --tags
```

From sbt, we can ask what it's going to test against.  You can leave off the `laws/` to show all modules, but I'm going to focus on the broken one to keep this tutorial concise:

```
sbt:http4s> show laws/mimaPreviousArtifacts
[info] Set(org.http4s:http4s-laws_2.12:0.22.0, org.http4s:http4s-laws_2.12:0.22.1, org.http4s:http4s-laws_2.12:0.22.2)
```

We're making sure our next 0.22 release is binary compatible with the previous 0.22 releases.  That way, if you use another library that extends http4s, as long as it was compiled with some version of 0.22, you'll be okay at runtime.  What went wrong?

```
sbt:http4s> laws/mimaReportBinaryIssues
[error] laws: Failed binary compatibility check against org.http4s:http4s-laws_2.12:0.22.0! Found 2 potential problems
[error]  * abstract synthetic method org$http4s$laws$discipline$ArbitraryInstances$_setter_$genAlphaToken_=(org.scalacheck.Gen)Unit in interface org.http4s.laws.discipline.ArbitraryInstances is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("org.http4s.laws.discipline.ArbitraryInstances.org$http4s$laws$discipline$ArbitraryInstances$_setter_$genAlphaToken_=")
[error]  * abstract method genAlphaToken()org.scalacheck.Gen in interface org.http4s.laws.discipline.ArbitraryInstances is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("org.http4s.laws.discipline.ArbitraryInstances.genAlphaToken")
[error] stack trace is suppressed; run last laws / mimaReportBinaryIssues for the full output
[error] (laws / mimaReportBinaryIssues) Failed binary compatibility check against org.http4s:http4s-laws_2.12:0.22.0! Found 2 potential problems
[error] Total time: 1 s, completed Aug 13, 2021, 10:17:00 PM
```

These can be hard to read unless you have a Java and a Scala background.  I have two decades of the former and a decade of the latter and even I still struggle!  But obviously `genAlphaToken()` is causing us problems.  The `ArbitraryInstances` interface has a new method.  Any other implementations that were compiled against the old definition of `ArbitraryInstances` might not have this and explode at runtime.  (This is highly unlikely: I bet everything that extends this is an http4s test.  But we do publish them, so we check them.)

Adding a new value to that trait in the middle of a release cycle isn't going to work.  What I propose is move the `genAlphaToken()` out of the `ArbitraryInstances` trait, and into the `ArbitraryInstances` object.  You'll find several there already with a comment about being there because of binary incompatibility in 0.21.  We should have moved those out before 0.22.0.  Oops!  Import from the companion object if the compiler can't find them.  Then if you run `mimaReportBinaryIssues` again, I think you'll find it passes.

shannonbarnes:
Thank you for the explaination.

rossabaker (approved):
Thanks for the original fix and the extra chore.  I'll merge this assuming the build passes, or has one of the known flakes.

PR #5071, opened 2021-08-13
weaken some context bounds

bplommer:


rossabaker:
That OAuthSuite compilation error snuck through CI.  Merging/rebasing series/0.22 should make it go away.

I am surprised this is binary compatible?

It also reminds me of https://github.com/http4s/http4s/pull/2325, an argument that I haven't managed to rehydrate into my brain yet this morning.

barryoneill:
@rossabaker oh wow, sorry I did not see breaking test.  Thanks for resolving it.

bplommer:
> I am surprised this is binary compatible?

I think changing a parameter type to a supertype is always binary compatible, no?

> It also reminds me of #2325, an argument that I haven't managed to rehydrate into my brain yet this morning.

Looks like this is just clearing up loose ends from https://github.com/http4s/http4s/pull/2906.

isomarcte (approved):

bplommer:
Oh, looks like this *is* failing on mima for Scala 2.

Also looks like mima isn't checking anything for Scala 3 - is this intended?

rossabaker:
> oh wow, sorry I did not see breaking OAuthSuite test

No worries.  I probably merged something that hadn't passed CI.  That's on me.

> I think changing a parameter type to a supertype is always binary compatible, no?

It's source compatible, but not binary compatible.  Adding this takes care of the "method apply" error, but not the "static method apply" error.

```scala
  @deprecated("Preserved for binary compatibility", "0.22.4")
  protected[HttpApp] def apply[F[_]](run: Request[F] => F[Response[F]])(implicit F: Sync[F]): HttpApp[F] =
    Http(run)(F: Defer[F])
```

I think that static forwarder is only used in non-Java interop.  I don't love making exceptions, but that one is probably okay?  We should be able to test the general case by compiling an app on http4s-0.22.3, and running against a snapshot built with this.  I think it will fail without the above method, and succeed with.

isomarcte (commented):
Oh derp, I guess I hadn't had enough coffee when I read this today. I'd really like to avoid breaking bin-compat.

Is there no way to work around the static method error? 🤔

rossabaker:
I don't think so.  It needs to be public, and when it becomes public, we run into an ambiguous overload.

I knew this problem was familiar: https://github.com/typelevel/cats-effect/pull/584

isomarcte:
If we're going to tackle this, we probably want to loosen constraints here too.

https://github.com/http4s/http4s/blob/series/0.22/core/src/main/scala/org/http4s/syntax/KleisliSyntax.scala#L55

I think `F` needs `Functor` and `G` needs `Defer`, neither need `Sync`.

I think those should be easier since we can just deprecate `KleisliHttpAppOps` and add a replacement class.

isomarcte:
Some of the other syntax classes might need similar treatment, I've not checked yet.

rossabaker:
If we want to be bincompat purists:

rossabaker:
If we want to be bincompat purists:
* Add `defer` and `localDefer` alongside `apply` and `local` with the constraints we want.
* Scaladoc on `apply` and `defer` why those other ones are better.
* On main, fix `apply` and `local` and deprecate `local` and `localDefer`.

It gives people a path on 0.22 and 0.23 without violating any traditional bincompat guarantees.  But it's more work for people who adopt the better ones on 0.22 and 0.23 as they move to 1.0, all to support a hypothetical Java interop that probably nobody does.

isomarcte:
>  But it's more work for people who adopt the better ones on 0.22 and 0.23 as they move to 1.0, all to support a hypothetical Java interop that probably nobody does.

Yeah I get that. I'll defer (pun intended) to whatever you think is best here.

rossabaker:
I'm truly on the fence.  My inner pragmatist is yelling "just break bincompat!" and my scars are reminding me "almost every time you break bincompat, something unexpected bites you in the ass."  I'm still hoping someone has a workaround the rest of us haven't seen.

bplommer:
How about just explicitly saying bincompat isn't guaranteed for Java interop?

rossabaker:
Let's proceed.  If a Java user stumbles into this, I'll point them to the changelog, and then feel bad, and look at bytecode postprocessing options.

rossabaker:
Looks like the only thing we have to do here is add the MiMa exceptions.

rossabaker:
I didn't realize this changes again in 0.23 until I hit a merge conflict.

PR #5073, opened 2021-08-14
Replace existing headers in withEntity

### LINKED ISSUE #5059: Request#withEntity appends Content-Type instead of replacing it since 0.22
Nice to meet you!

I've discovered a behavior change of `Request#withEntity` between 0.21.25 and 0.22.x.

If a `Request` already has a `Content-Type` header, it is replaced with a new value of `EntityEncoder` in 0.21.25, while it is appended as another header in any version of 0.22.x.

Example:

```scala
import cats.effect.IO

import org.http4s.MediaType
import org.http4s.Request
import org.http4s.headers.`Content-Type`

println {
  Request[IO]()
    .putHeaders(`Content-Type`(MediaType.text.plain))
    .withEntity[String]("foo")
}
```

Running with Coursier:

```
$ scala -cp .:"$(cs fetch --classpath org.http4s::http4s-core:0.21.25)" Test.scala
Request(method=GET, uri=/, headers=Headers(Content-Length: 3, Content-Type: text/plain; charset=UTF-8))
```

```
$ scala -cp .:"$(cs fetch --classpath org.http4s::http4s-core:0.22.2)" Test.scala
Request(method=GET, uri=/, headers=Headers(Content-Type: text/plain, Content-Length: 3, Content-Type: text/plain; charset=UTF-8))
```

(The output should be the same)

I couldn't find a relevant entry in the changelog. I understand that I may be misusing the library. However, since multiple `Content-Type`s are never allowed, it would be helpful to revert the behavior to the old one.

rossabaker:
Fixes #5059.

With the global header registry, we used to be able to append two sets of headers and detect which were single and which were recurring from their names.  In the type class model of 0.22, this is no longer possible.  We need to merge the headers from the `EntityEncoder` with the headers already on the message.

Usually the headers are `Content-Type` and `Content-Length`, both of which are single.  Using a merge operation that replaces existing headers fits well here.  We get a change in behavior if the message already has any recurring headers that the `EntityEncoder` tries to add, but it's fair to say anything the `EntityEncoder` cared to specify, it should control.

### DIFF
--- a/core/src/main/scala/org/http4s/Message.scala
+++ b/core/src/main/scala/org/http4s/Message.scala
@@ -91,7 +91,7 @@ sealed trait Message[F[_]] extends Media[F] { self =>
 
       case None => w.headers
     }
-    change(body = entity.body, headers = headers.transform(_ ++ hs.headers))
+    change(body = entity.body, headers = headers ++ hs)
   }
 
   /** Sets the entity body without affecting headers such as `Transfer-Encoding`
--- a/tests/src/test/scala/org/http4s/MessageSuite.scala
+++ b/tests/src/test/scala/org/http4s/MessageSuite.scala
@@ -18,10 +18,12 @@ package org.http4s
 
 import cats.data.NonEmptyList
 import cats.effect.IO
+import cats.syntax.all._
 import com.comcast.ip4s._
 import fs2.Pure
 import org.http4s.headers.{Authorization, Cookie, `Content-Type`, `X-Forwarded-For`}
 import org.http4s.syntax.all._
+import org.typelevel.ci._
 import org.typelevel.vault._
 
 class MessageSuite extends Http4sSuite {
@@ -291,5 +293,17 @@ class MessageSuite extends Http4sSuite {
     trait F2[A] extends F1[A]
     Response[F2]().covary[F1]
     true
+  }
+
+  test("withEntity should not duplicate Content-Type header") {
+    // https://github.com/http4s/http4s/issues/5059
+    val hdrs = Request[IO]()
+      .putHeaders(`Content-Type`(MediaType.text.html))
+      .withEntity[String]("foo")
+      .headers
+      .headers
+      .filter(_.name === Header[`Content-Type`].name)
+    // Should preserve only the EntityEncoder's Content-Type
+    assertEquals(hdrs, List(Header.Raw(ci"Content-Type", "text/plain; charset=UTF-8")))
   }
 }


rossabaker:
We could go a step further and say it should remove any payload headers.  (If a message had a gzip content encoding, and we replaced the entity without an encoder that repeated the same, shouldn't we go back to identity encoding?)  But that's a more radical change.

rossabaker:
I think the dubious fields are the payload header fields" and the [representation header fields](https://datatracker.ietf.org/doc/html/rfc7231#section-3.1) and the [payload header fields](https://datatracker.ietf.org/doc/html/rfc7231#section-3.3).  When we change the entity, either in `withEntity` or using an `EntityEncoder`, anything we inherited from those sets of fields is called into question.

Changing this is unfortunately one of those Quiet Semantic Changes that I've been trying to minimize.

ChristopherDavenport (commented):
I think this is a sensible change, this is never the right answer to duplicate single key headers.

rossabaker:
Do you think it's sensible as is, or should we clear out those other body-related headers?

isomarcte:
> Do you think it's sensible as is, or should we clear out those other body-related headers?

I don't think we should clear out the other body related headers. I've seen a good amount of code that unfortunately bootstraps the body using one `EntityEncoder`, say `EntityEncoder[String]`, and then explicitly overwrites the the `Content-Type` to the intended type, e.g. `application/xml`. Obviously this is not ideal, but I don't want to silently break that.

@rossabaker I think what you have here is good as is. I'd be curious if we could explore ways of forbidding users from directly setting the body/payload headers via the non-EntityEncoder methods, but that is a topic for another issue (and a time when my thoughts are more fully formed ;) ).

isomarcte (approved):

isomarcte:
> I don't think we should clear out the other body related headers. I've seen a good amount of code that unfortunately bootstraps the body using one EntityEncoder, say EntityEncoder[String], and then explicitly overwrites the the Content-Type to the intended type, e.g. application/xml. Obviously this is not ideal, but I don't want to silently break that.

I guess that would still work if we removed the headers here, it's the other order that is problematic. Though I'd still be hesitant to do that. There's nothing stopping users from writing `EntityEncoders` which don't set the payload headers correctly and then setting the headers before calling `withEntity`.

rossabaker:
> I'd be curious if we could explore ways of forbidding users from directly setting the body/payload headers via the non-EntityEncoder methods, but that is a topic for another issue (and a time when my thoughts are more fully formed ;) ).

I think akka-http silently ignores some of them.  It's a hard problem: you need a header name type that is "strings, with out control characters or high bit characters, that aren't in this set".  And then make it usable!

> There's nothing stopping users from writing EntityEncoder which don't set the payload headers correctly and then setting the headers before calling withEntity.

It's a natural flow, too: when we think about a message, headers come before the entity.  We like to think of a message as a product type, but the headers and the body are interdependent, and many brain teasers flow from this.

PR #5075, opened 2021-08-14
Print correct http4s version on Blaze startup

### LINKED ISSUE #5074: Blaze Server prints self version instead of http4s version on startup
Due to namespace collision Blaze server prints Blaze version twice instead of outputting Http4s + Blaze versions. Probably due to erased import.

arixmkii:
Fixes #5074 

Explicitly reference correct BuildInfo for Http4s version.

Seems that original import got lost, so, this time I reintroduced it in a more explicit way with meaningful alias.


ChristopherDavenport (approved):

rossabaker (approved):
Eek!  Good find.

PR #5082, opened 2021-08-17
Update Origin header

### LINKED ISSUE #5009: Update `Origin` header to current standard or document decision not to do so
The [model for the `Origin` header](https://github.com/http4s/http4s/blob/main/core/src/main/scala/org/http4s/headers/Origin.scala) allows a non-empty list of hosts, citing [RFC 6454](http://tools.ietf.org/html/rfc6454#section-7) but the [current standard](https://fetch.spec.whatwg.org/#origin-header) allows only one host, noting that this supplants the definition in the RFC.

I can think of the following options:

1. Update the models to only allow one host, and have the parser handle extra hosts by:
  a. ignoring them
  b. raising an error
2. Add a third subtype to the model, `SingleHost`, with a comment noting that `HostList` is not spec-compliant
3. Headers are type-classes now! Why not provide strict or permissive versions? 
4. Leave the behaviour as is, with a comment added to document the decision.

I would favour option 1a, but my view obviously doesn't deserve any special weight.

bplommer:
Resolves #5009.

I've also taken the opportunity to backtify the `null` value for consistency with the rest of the library, to stop parsing empty values as `null` (it's not allowed by the spec), and generally to simplify the parsing implementation.

### DIFF
--- a/core/shared/src/main/scala/org/http4s/headers/Origin.scala
+++ b/core/shared/src/main/scala/org/http4s/headers/Origin.scala
@@ -17,9 +17,8 @@
 package org.http4s
 package headers
 
-import cats.data.NonEmptyList
-import cats.parse.{Parser, Parser0, Rfc5234}
-import org.http4s.Uri.RegName
+import cats.parse.Parser.{`end`, char, string}
+import cats.parse.{Parser, Parser0}
 import org.http4s.util.{Renderable, Writer}
 import org.typelevel.ci._
 
@@ -28,18 +27,15 @@ sealed abstract class Origin
 object Origin {
   // An Origin header may be the string "null", representing an "opaque origin":
   // https://stackoverflow.com/questions/42239643/when-does-firefox-set-the-origin-header-to-null-in-post-requests
-  case object Null extends Origin
-
-  // If the Origin is not "null", it is a non-empty list of Hosts:
-  // http://tools.ietf.org/html/rfc6454#section-7
-  final case class HostList(hosts: NonEmptyList[Host]) extends Origin
+  case object `null` extends Origin
 
   // A host in an Origin header isn't a full URI.
   // It only contains a scheme, a host, and an optional port.
   // Hence we re-used parts of the Uri class here, but we don't use a whole Uri:
   // http://tools.ietf.org/html/rfc6454#section-7
   final case class Host(scheme: Uri.Scheme, host: Uri.Host, port: Option[Int] = None)
-      extends Renderable {
+      extends Origin
+      with Renderable {
     def toUri: Uri =
       Uri(scheme = Some(scheme), authority = Some(Uri.Authority(host = host, port = port)))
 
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/org/http4s/headers/Origin.scala
```diff
@@ -28,48 +27,32 @@ sealed abstract class Origin
 object Origin {
   // An Origin header may be the string "null", representing an "opaque origin":
   // https://stackoverflow.com/questions/42239643/when-does-firefox-set-the-origin-header-to-null-in-post-requests
-  case object Null extends Origin
-
-  // If the Origin is not "null", it is a non-empty list of Hosts:
-  // http://tools.ietf.org/html/rfc6454#section-7
-  final case class HostList(hosts: NonEmptyList[Host]) extends Origin
+  case object `null` extends Origin
```
rossabaker, line 30: I can't think of another case that we modeled the header as a sum type instead of a unary case class of a sum type, but I think I like it.  It sidesteps the `org.http4s.Origin` vs. `org.http4s.header.Origin` debate.  And it slipped through review the first time without me noticing!

tests/shared/src/test/scala/org/http4s/parser/OriginHeaderSuite.scala
```diff
@@ -30,62 +29,55 @@ class OriginHeaderSuite extends munit.FunSuite {
...
+    val origin = Origin.`null`
     val headers = Headers(("Origin", text))
     val extracted = headers.get[Origin]
     assertEquals(extracted, Some(origin))
   }
 
-  test("OriginHeader parser should Parse an empty origin") {
-    val text = ""
-    val origin = Origin.Null
+  test("OriginHeader should fail on a list of multiple hosts") {
```
rossabaker, line 70: Might be worth a comment referring to the obsolete RFC here.

tests/shared/src/test/scala/org/http4s/parser/OriginHeaderSuite.scala
```diff
@@ -30,62 +29,55 @@ class OriginHeaderSuite extends munit.FunSuite {
...
     val headers = Headers(("Origin", text))
     val extracted = headers.get[Origin]
-    assertEquals(extracted, Some(origin))
+    assertEquals(extracted, None)
   }
 
-  test("OriginHeader parser should Parse a 'null' origin") {
-    val text = "null"
-    val origin = Origin.Null
+  test("OriginHeader should fail on an empty string") {
```
rossabaker, line 77: I'm thinking about whether this a bug we should backport, or a breaking change we shouldn't backport.  But it's certainly right for this version.

bplommer: I'd say the former, but I don't think it's super important.

rossabaker: That sounds good to me.

rossabaker (approved):
This looks good to me.  I was cursing that `HostList` over the weekend on another thing I'm working on.

PR #5083, opened 2021-08-17
Make `Http*` syntax available without an import

bplommer:
As discussed in #5062 - by adding these implicits to the companion of one of the type parameters, we remove the need for an import.

### DIFF
--- a/core/src/main/scala/org/http4s/syntax/package.scala
+++ b/core/src/main/scala/org/http4s/syntax/package.scala
@@ -18,6 +18,7 @@ package org.http4s
 
 package object syntax {
   object all extends AllSyntaxBinCompat
+  @deprecated("import is no longer needed", "0.22.3")
   object kleisli extends KleisliSyntax with KleisliSyntaxBinCompat0 with KleisliSyntaxBinCompat1
   object literals extends LiteralsSyntax
   object string extends StringSyntax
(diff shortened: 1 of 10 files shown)


rossabaker:
Oops.  I backported that `->>` and it merged clean and compiled clean but failed MiMa.  Those build failures are my fault.

core/src/main/scala/org/http4s/syntax/package.scala
```diff
@@ -18,6 +18,7 @@ package org.http4s
 
 package object syntax {
   object all extends AllSyntaxBinCompat
+  @deprecated("import is no longer needed", "0.22.3")
```
rossabaker, line 21: What about the `translate` methods?  I can understand how `orNotFound` works (and it's exciting that it does!).

rossabaker:
Merging #5084 should clear the debris I left for you.

core/src/main/scala/org/http4s/syntax/package.scala
bplommer: I don't think it should be any different? They're all on types that are aliases for `Kleisli` with a `Response` parameter. I'll make sure there's test coverage though.

rossabaker: Hmm, I think you're right, if it winds its way through those type aliases.  :+1: to making sure it's tested.

bplommer:
I think this is good to go.

rossabaker:
I'm excited about this one.

bplommer:
Just realised this technique can also be used to allocate things effectfully and then define operations directly on the allocating effect. Not particularly relevant here but wanted to shout about it because it's cool: https://github.com/fd4s/fs2-kafka/pull/662

PR #5084, opened 2021-08-17
Restore binary compatibility of ->>

rossabaker:
Conjecture: none of these should be vals.

### DIFF
--- a/dsl/src/main/scala/org/http4s/dsl/Http4sDsl.scala
+++ b/dsl/src/main/scala/org/http4s/dsl/Http4sDsl.scala
@@ -28,7 +28,6 @@ trait Http4sDsl2[F[_], G[_]] extends RequestDsl with Statuses with Responses[F, 
   val :? : impl.:?.type = impl.:?
   val ~ : impl.~.type = impl.~
   val -> : impl.->.type = impl.->
-  def ->> : impl.->>.type = impl.->>
   val /: : impl./:.type = impl./:
   val +& : impl.+&.type = impl.+&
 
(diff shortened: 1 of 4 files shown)


bplommer:
What's the purpose of this indirection, by the way?

rossabaker:
That is an outstanding question.  The original DSL was lifted from Finagle, sometime early in the Mesozoic Era.  I can't remember whether this indirection was part of that.  It also might have been a binary compatibility trick from the days of 2.9.  _Maybe_ we're exploiting it to get a singleton type on `Root`, but I can't think of a reason for the rest.

rossabaker:
Well, Scala 2 apparently sees some virtue in it.

rossabaker:
I think it has to be a `val` to be used as an extractor in Scala 2, but I haven't figured out how to make a val binary compatible yet.

bplommer:
Oh, is `Http4sDsl2` a bincompat trait? Do we need another one?

rossabaker:
We can pull `->>` into `org.http4s.dsl._` and it compiles and works, but then it won't be inherited from `import org.http4s.dsl.io._`.  The way we bundle request extractors and response generators has bitten us here.

rossabaker:
`Http4sDsl2` has two type parameters, in order to support http4s-directives.  @hamnis would have the most context on that.  I don't think it had anything to do with bincompat.

bplommer:
Seems I also left these out of `RequestDsl` - that one leaves all of the extractors abstract so definitely wouldn't be binary compatible.

rossabaker:
Yeah, I just remembered that exists as I was fighting this.  We'll want to polish that up in main.  I'm still trying to figure out how we can salvage the feature for 0.22.

bplommer:
This seems to make mima happy:

```scala
trait RequestDslBinCompat extends RequestDsl {
  val ->> : impl.->>.type = impl.->>
}

object request extends RequestDslBinCompat {
  ...
}

trait Http4sDslBinCompat[F[_]] extends Http4sDsl[F] with RequestDslBinCompat

object io extends Http4sDslBinCompat[F[_]] {
  ...
}

object Http4sDsl {
  def apply[F[_]]: Http4sDsl[F] with RequestDslBinCompat = new Http4sDslBinCompat[F] {}
  ...
}
```

That last one breaks if you narrow `Http4sDsl[F] with RequestDslBinCompat` to `Http4sDslBincompat[F[_]]` - I don't have the slightest clue why one is ok and the other isn't, but mima isn't complaining.

PR #5093, opened 2021-08-18
RFC: Typed response statuses in Scala 2.13+

bplommer:
Lifting information in `Status` to type level in a backwards-compatible way is surprisingly straightforward. Trying to preserve that information in `Response` leads to a world of variance-related hurt. `ResponseBuilder` a demo of what this makes possible rather than something that's ready to be merged, but maybe something like it would be a good fit for the DSL module?

If the changes to `Status` are approved I can split them out into a standalone PR.


bplommer:
Now if only I'd thought to compile the tests

bplommer:
Apparently I messed something up when running mima locally, because this is super binary incompatible.

bplommer:
I win 😎

rossabaker:
Just a note that I haven't forgotten this but it's been buried behind more urgent looking things and I'm very sorry about it.

bplommer:
No need at all to be sorry! It's not urgent at all.

core/src/main/scala-2.13+/org.http4s/ResponseBuilder.scala
```diff
@@ -0,0 +1,97 @@
...
+  //  [error]    |                                           teapots cannot brew espresso
+
+  // scalac:
+  //  [error] http4s/core/src/main/scala-2.13+/org.http4s/ResponseBuilder.scala:85:54: ambiguous implicit values:
+  //  [error]  both method instance1 in object NotGiven of type [A]scala.util.NotGiven[A]
+  //  [error]  and method instance2 in object NotGiven of type [A](implicit ev: A): scala.util.NotGiven[A]
+  //  [error]  match expected type scala.util.NotGiven[Int(418) <:< 418]
+  //  [error]   ResponseBuilder(Status.ImATeapot).withEspressoShots(1)
+  //  [error]                                                      ^
+}
```
armanbilge, line 97: Tbh I was a bit confused by this at first :) can we move it into a test suite maybe?

core/src/main/scala-2.13+/org.http4s/ResponseBuilder.scala
```diff
@@ -0,0 +1,97 @@
...
+      @implicitNotFound(
+        s"This response status does not allow a body") @unused ev: status.IsEntityAllowed <:< true)
+      : ResponseBuilder[S, T, Some] =
+    new ResponseBuilder(status, Some(b)) {}
+
+  // TODO: implement
+  def withEspressoShots(@unused n: Int)(implicit
+      @implicitNotFound("teapots cannot brew espresso") @unused ev: NotGiven[status.Code <:< 418])
+      : ResponseBuilder[S, Body, G] =
+    this
```
armanbilge, line 40: Also confused by this 😆

core/src/main/scala-2.13+/org.http4s/ResponseBuilder.scala
bplommer: Sure, this was only supposed to be a demo!

bplommer:
I might redo this with an initial emphasis on exposing the category of status (1xx/2xx/3xx/4xx/5xx) in `Response` types, which might be more feasible with the variance changes.

PR #5100, opened 2021-08-20
Bug: Wrong PathInfoCaret set in Http4sServlet?

fthomas:
This PR is not intended to be merged as-is but a reproduction of a possible bug in `Http4sServlet` in combination with `Router`. The first commit in this PR adds a `Router` to `BlockingHttp4sServletSuite` which results in all its tests to fail. The second commit sets `PathInfoCaret` only if `getPathInfoIndex(req, uri)` returns a non-negative value which fixes the `BlockingHttp4sServletSuite` tests. I've absolutely no idea what I'm doing and if this is a correct fix.

I noticed this bug when upgrading a jetty-server in combination with `Router` and `WebjarService` from 0.21 to 0.22.

### DIFF
--- a/servlet/src/main/scala/org/http4s/servlet/Http4sServlet.scala
+++ b/servlet/src/main/scala/org/http4s/servlet/Http4sServlet.scala
@@ -98,34 +98,38 @@ abstract class Http4sServlet[F[_]](service: HttpApp[F], servletIo: ServletIo[F])
       httpVersion = version,
       headers = toHeaders(req),
       body = servletIo.reader(req),
-      attributes = Vault.empty
-        .insert(Request.Keys.PathInfoCaret, getPathInfoIndex(req, uri))
-        .insert(
-          Request.Keys.ConnectionInfo,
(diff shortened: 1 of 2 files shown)


fthomas:
I should have mentioned that changing `PathInfoCaret` has an effect on `Router` because of:
https://github.com/http4s/http4s/blob/cf77cfeb2d0bd09337da0b64a636e29bbfb77d27/server/src/main/scala/org/http4s/server/Router.scala#L53-L61
Without this change `resultCaret` is `Some(-1) |+| Some(1) = Some(0)`, and with this change it is `None |+| Some(1) = Some(1)`.

hamnis:
ouch, this is probably my fault.

The indexing is segmentbased after 0.22, which has had a few bugs since that was changed. This is a good find.
We should have a look at how the `getPathInfoIndex` function calulates the index.

armanbilge:
I believe this was fixed in https://github.com/http4s/http4s/pull/5793. Thanks for your help tracking it down!

PR #5101, opened 2021-08-20
Add fetch options+timeout to FetchClient.

rpiaggio:
Add timeout capability to `FetchClient` and allow passing all the options present in `RequestInit`.

### DIFF
--- a/dom-fetch-client/src/main/scala/org/http4s/dom/FetchClient.scala
+++ b/dom-fetch-client/src/main/scala/org/http4s/dom/FetchClient.scala
@@ -45,15 +45,14 @@ private[dom] object FetchClient {
         .makeCaseFull { (poll: Poll[F]) =>
           F.delay(new AbortController()).flatMap { abortController =>
             val requestOptions = req.attributes.lookup(FetchOptions.Key)
-            val mergedOptions = requestOptions.fold(options)(options.overrideWith)
+            val mergedOptions = requestOptions.fold(options)(options.merge)
 
             val init = new RequestInit {}
 
             init.method = req.method.name.asInstanceOf[HttpMethod]
             // Referer header is converted to a Fetch parameter below, since it's a forbidden header.
             // See https://developer.mozilla.org/en-US/docs/Glossary/Forbidden_header_name
-            init.headers = new Headers(
-              toDomHeaders(req.headers.transform(_.filterNot(_.name === Header[Referer].name))))
+            init.headers = new Headers(toDomHeaders(req.headers))
             body.foreach { body =>
               init.body = arrayBuffer2BufferSource(body.toJSArrayBuffer)
             }
--- a/dom-fetch-client/src/main/scala/org/http4s/dom/FetchClient.scala
+++ b/dom-fetch-client/src/main/scala/org/http4s/dom/FetchClient.scala
@@ -50,8 +50,6 @@ private[dom] object FetchClient {
             val init = new RequestInit {}
 
             init.method = req.method.name.asInstanceOf[HttpMethod]
-            // Referer header is converted to a Fetch parameter below, since it's a forbidden header.
-            // See https://developer.mozilla.org/en-US/docs/Glossary/Forbidden_header_name
             init.headers = new Headers(toDomHeaders(req.headers))
             body.foreach { body =>
               init.body = arrayBuffer2BufferSource(body.toJSArrayBuffer)
@@ -63,6 +61,8 @@ private[dom] object FetchClient {
             mergedOptions.keepAlive.foreach(init.keepalive = _)
             mergedOptions.mode.foreach(init.mode = _)
             mergedOptions.redirect.foreach(init.redirect = _)
+            // Referer headers are forbidden in Fetch, but we make a best effort to preserve behavior across clients.
+            // See https://developer.mozilla.org/en-US/docs/Glossary/Forbidden_header_name
             // If there's a Referer header, it will have more priority than the client's `referrer` (if present)
             // but less priority than the request's `referrer` (if present).
             requestOptions
--- a/dom-fetch-client/src/main/scala/org/http4s/dom/FetchClient.scala
+++ b/dom-fetch-client/src/main/scala/org/http4s/dom/FetchClient.scala
@@ -19,8 +19,9 @@ package dom
 
 import cats.effect.Async
 import cats.effect.Resource
-import cats.syntax.all._
+import cats.effect.Poll
 import cats.effect.syntax.all._
+import cats.syntax.all._
 import java.util.concurrent.TimeoutException
 import org.http4s.client.Client
 import org.http4s.headers.{`Referer`}
@@ -33,7 +34,7 @@ import org.scalajs.dom.experimental.RequestInit
 import org.scalajs.dom.experimental.{Response => FetchResponse}
 import scala.concurrent.duration._
 
-object FetchClient {
+private[dom] object FetchClient {
 
   private[dom] def makeClient[F[_]](
       requestTimeout: Duration,
@@ -41,7 +42,7 @@ object FetchClient {
   )(implicit F: Async[F]): Client[F] = Client[F] { (req: Request[F]) =>
     Resource.eval(req.body.chunkAll.filter(_.nonEmpty).compile.last).flatMap { body =>
       Resource
-        .makeCase {
+        .makeCaseFull { (poll: Poll[F]) =>
           F.delay(new AbortController()).flatMap { abortController =>
             val requestOptions = req.attributes.lookup(FetchOptions.Key)
             val mergedOptions = requestOptions.fold(options)(options.overrideWith)
@@ -72,8 +73,7 @@ object FetchClient {
               .foreach(referrer => init.referrer = referrer.renderString)
             mergedOptions.referrerPolicy.foreach(init.referrerPolicy = _)
 
-            val fetch = F
-              .fromPromise(F.delay(Fetch.fetch(req.uri.renderString, init)))
+            val fetch = poll(F.fromPromise(F.delay(Fetch.fetch(req.uri.renderString, init))))
               .onCancel(F.delay(abortController.abort()))
 
             requestTimeout match {
--- a/dom-fetch-client/src/main/scala/org/http4s/dom/FetchClient.scala
+++ b/dom-fetch-client/src/main/scala/org/http4s/dom/FetchClient.scala
@@ -44,7 +44,7 @@ object FetchClient {
       cache: Option[RequestCache],
       credentials: Option[RequestCredentials],
       integrity: Option[String],
-      keepalive: Option[Boolean],
+      keepAlive: Option[Boolean],
       mode: Option[RequestMode],
       redirect: Option[RequestRedirect],
       referrer: Option[String],
@@ -53,7 +53,7 @@ object FetchClient {
     Resource.eval(req.body.chunkAll.filter(_.nonEmpty).compile.last).flatMap { body =>
       Resource
         .makeCase {
-          F.delay {
+          F.delay(new AbortController()).flatMap { abortController =>
             val init = new RequestInit {}
 
             init.method = req.method.name.asInstanceOf[HttpMethod]
@@ -61,34 +61,29 @@ object FetchClient {
             body.foreach { body =>
               init.body = arrayBuffer2BufferSource(body.toJSArrayBuffer)
             }
+            init.signal = abortController.signal
             cache.foreach(init.cache = _)
             credentials.foreach(init.credentials = _)
             integrity.foreach(init.integrity = _)
-            keepalive.foreach(init.keepalive = _)
+            keepAlive.foreach(init.keepalive = _)
             mode.foreach(init.mode = _)
             redirect.foreach(init.redirect = _)
             referrer.foreach(init.referrer = _)
             referrerPolicy.foreach(init.referrerPolicy = _)
 
-            init
-          } >>= {
+            val fetch = F
+              .fromPromise(F.delay(Fetch.fetch(req.uri.renderString, init)))
+              .onCancel(F.delay(abortController.abort()))
+
             requestTimeout match {
               case d: FiniteDuration =>
-                init =>
-                  Resource
-                    .eval(F.delay(new AbortController()))
-                    .use(controller =>
-                      F.fromPromise(F.delay {
-                        init.signal = controller.signal
-                        Fetch.fetch(req.uri.renderString, init)
-                      }).timeoutTo(
-                        d,
-                        F.delay(controller.abort()) >>
-                          F.raiseError[FetchResponse](new TimeoutException(
-                            s"Request to ${req.uri.renderString} timed out after ${d.toMillis} ms"))
-                      ))
+                fetch.timeoutTo(
+                  d,
+                  F.raiseError[FetchResponse](new TimeoutException(
+                    s"Request to ${req.uri.renderString} timed out after ${d.toMillis} ms"))
+                )
               case _ =>
-                init => F.fromPromise(F.delay(Fetch.fetch(req.uri.renderString, init)))
+                fetch
             }
           }
         } { case (r, exitCase) =>
--- a/dom-fetch-client/src/main/scala/org/http4s/dom/FetchClient.scala
+++ b/dom-fetch-client/src/main/scala/org/http4s/dom/FetchClient.scala
@@ -32,7 +32,6 @@ import org.scalajs.dom.experimental.HttpMethod
 import org.scalajs.dom.experimental.RequestInit
 import org.scalajs.dom.experimental.{Response => FetchResponse}
 import scala.concurrent.duration._
-import org.typelevel.ci._
 
 object FetchClient {
 
--- a/dom-fetch-client/src/main/scala/org/http4s/dom/FetchClient.scala
+++ b/dom-fetch-client/src/main/scala/org/http4s/dom/FetchClient.scala
@@ -23,53 +23,55 @@ import cats.syntax.all._
 import cats.effect.syntax.all._
 import java.util.concurrent.TimeoutException
 import org.http4s.client.Client
+import org.http4s.headers.{`Referer`}
 import org.scalajs.dom.crypto._
 import org.scalajs.dom.experimental.AbortController
 import org.scalajs.dom.experimental.Fetch
 import org.scalajs.dom.experimental.Headers
 import org.scalajs.dom.experimental.HttpMethod
-import org.scalajs.dom.experimental.ReferrerPolicy
-import org.scalajs.dom.experimental.RequestCache
-import org.scalajs.dom.experimental.RequestCredentials
 import org.scalajs.dom.experimental.RequestInit
-import org.scalajs.dom.experimental.RequestMode
-import org.scalajs.dom.experimental.RequestRedirect
 import org.scalajs.dom.experimental.{Response => FetchResponse}
 import scala.concurrent.duration._
+import org.typelevel.ci._
 
 object FetchClient {
 
   private[dom] def makeClient[F[_]](
       requestTimeout: Duration,
-      cache: Option[RequestCache],
-      credentials: Option[RequestCredentials],
-      integrity: Option[String],
-      keepAlive: Option[Boolean],
-      mode: Option[RequestMode],
-      redirect: Option[RequestRedirect],
-      referrer: Option[String],
-      referrerPolicy: Option[ReferrerPolicy]
+      options: FetchOptions
   )(implicit F: Async[F]): Client[F] = Client[F] { (req: Request[F]) =>
     Resource.eval(req.body.chunkAll.filter(_.nonEmpty).compile.last).flatMap { body =>
       Resource
         .makeCase {
           F.delay(new AbortController()).flatMap { abortController =>
+            val requestOptions = req.attributes.lookup(FetchOptions.Key)
+            val mergedOptions = requestOptions.fold(options)(options.overrideWith)
+
             val init = new RequestInit {}
 
             init.method = req.method.name.asInstanceOf[HttpMethod]
-            init.headers = new Headers(toDomHeaders(req.headers))
+            // Referer header is converted to a Fetch parameter below, since it's a forbidden header.
+            // See https://developer.mozilla.org/en-US/docs/Glossary/Forbidden_header_name
+            init.headers = new Headers(
+              toDomHeaders(req.headers.transform(_.filterNot(_.name === Header[Referer].name))))
             body.foreach { body =>
               init.body = arrayBuffer2BufferSource(body.toJSArrayBuffer)
             }
             init.signal = abortController.signal
-            cache.foreach(init.cache = _)
-            credentials.foreach(init.credentials = _)
-            integrity.foreach(init.integrity = _)
-            keepAlive.foreach(init.keepalive = _)
-            mode.foreach(init.mode = _)
-            redirect.foreach(init.redirect = _)
-            referrer.foreach(init.referrer = _)
-            referrerPolicy.foreach(init.referrerPolicy = _)
+            mergedOptions.cache.foreach(init.cache = _)
+            mergedOptions.credentials.foreach(init.credentials = _)
+            mergedOptions.integrity.foreach(init.integrity = _)
+            mergedOptions.keepAlive.foreach(init.keepalive = _)
+            mergedOptions.mode.foreach(init.mode = _)
+            mergedOptions.redirect.foreach(init.redirect = _)
+            // If there's a Referer header, it will have more priority than the client's `referrer` (if present)
+            // but less priority than the request's `referrer` (if present).
+            requestOptions
+              .flatMap(_.referrer)
+              .orElse(req.headers.get[Referer].map(_.uri))
+              .orElse(options.referrer)
+              .foreach(referrer => init.referrer = referrer.renderString)
+            mergedOptions.referrerPolicy.foreach(init.referrerPolicy = _)
 
             val fetch = F
               .fromPromise(F.delay(Fetch.fetch(req.uri.renderString, init)))
--- a/dom-fetch-client/src/main/scala/org/http4s/dom/FetchClient.scala
+++ b/dom-fetch-client/src/main/scala/org/http4s/dom/FetchClient.scala
@@ -36,22 +36,19 @@ import org.scalajs.dom.experimental.RequestMode
 import org.scalajs.dom.experimental.RequestRedirect
 import org.scalajs.dom.experimental.{Response => FetchResponse}
 import scala.concurrent.duration._
-import scala.scalajs.js
 
 object FetchClient {
 
-  def apply[F[_]](implicit F: Async[F]): Client[F] = apply()
-
-  def apply[F[_]](
-      requestTimeout: js.UndefOr[FiniteDuration] = js.undefined,
-      cache: js.UndefOr[RequestCache] = js.undefined,
-      credentials: js.UndefOr[RequestCredentials] = js.undefined,
-      integrity: js.UndefOr[String] = js.undefined,
-      keepalive: js.UndefOr[Boolean] = js.undefined,
-      mode: js.UndefOr[RequestMode] = js.undefined,
-      redirect: js.UndefOr[RequestRedirect] = js.undefined,
-      referrer: js.UndefOr[String] = js.undefined,
-      referrerPolicy: js.UndefOr[ReferrerPolicy] = js.undefined
+  private[dom] def makeClient[F[_]](
+      requestTimeout: Duration,
+      cache: Option[RequestCache],
+      credentials: Option[RequestCredentials],
+      integrity: Option[String],
+      keepalive: Option[Boolean],
+      mode: Option[RequestMode],
+      redirect: Option[RequestRedirect],
+      referrer: Option[String],
+      referrerPolicy: Option[ReferrerPolicy]
   )(implicit F: Async[F]): Client[F] = Client[F] { (req: Request[F]) =>
     Resource.eval(req.body.chunkAll.filter(_.nonEmpty).compile.last).flatMap { body =>
       Resource
@@ -75,8 +72,8 @@ object FetchClient {
 
             init
           } >>= {
-            requestTimeout.toOption match {
-              case Some(d: FiniteDuration) =>
+            requestTimeout match {
+              case d: FiniteDuration =>
                 init =>
                   Resource
                     .eval(F.delay(new AbortController()))
--- a/dom-fetch-client/src/main/scala/org/http4s/dom/FetchClient.scala
+++ b/dom-fetch-client/src/main/scala/org/http4s/dom/FetchClient.scala
@@ -20,30 +20,78 @@ package dom
 import cats.effect.Async
 import cats.effect.Resource
 import cats.syntax.all._
+import cats.effect.syntax.all._
+import java.util.concurrent.TimeoutException
 import org.http4s.client.Client
 import org.scalajs.dom.crypto._
+import org.scalajs.dom.experimental.AbortController
 import org.scalajs.dom.experimental.Fetch
 import org.scalajs.dom.experimental.Headers
 import org.scalajs.dom.experimental.HttpMethod
+import org.scalajs.dom.experimental.ReferrerPolicy
+import org.scalajs.dom.experimental.RequestCache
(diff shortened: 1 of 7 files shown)


cquiroz (approved):
Looks good to me

armanbilge (changes requested):
Thanks for the PR, glad this module is getting attention!

But hang on—aren't many of these "options" simply headers that should be provided in the request? Apologies if I am missing something.

Additionally I'm very 👎 on using JS APIs (e..g., `js.UndefOr`). This is not meant to be a facade, but an idiomatic http4s client. So, should use standard Scala classes e.g. `Option` and maybe even implement a `FetchClientBuilder` pattern used in the other http4s modules

armanbilge:
Also, since these options are per-request instead of being used for the `FetchClient` constructor they should be provided as `attributes` on the request itself.

rpiaggio:
>Thanks for the PR, glad this module is getting attention!

Thank you for implementing this.

>aren't many of these "options" simply headers that should be provided in the request? 

Mostly they control how the Fetch client will interact with the browser. See https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/fetch#parameters

I now changed the parameters to `Option` and implemented the Builder pattern. Note that this is a breaking change for people that were using `FetchClient` directly.

>Also, since these options are per-request instead of being used for the FetchClient constructor they should be provided as attributes on the request itself.

I'm not so sure about this. In JS there's really no client instance, so it's up to us whether to interpret `Fetch.fetch` parameters as belonging to the client or the request. It seems to me it's more convenient for them to be client properties, insofar as `requestTimeout` is also a client property and not a request one. For example, we now have a 
``` scala
private val client = FetchClientBuilder[F]
    .withCredentials(RequestCredentials.include)
    .resource
```
that we use for different requests.

armanbilge:
Thanks for iterating on my feedback!

> In JS there's really no client instance, so it's up to us whether to interpret Fetch.fetch parameters as belonging to the client or the request.

Ok, this is a fair point. How about this: what if we define these as per-request attributes, and then in the fetch client builder they can also provide some default attributes? Also:

* `requestTimeout`: ✅ client parameter, this seems standard for other http4s clients
* `referrer`: [this is a header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referer)
* `referrerPolicy`: [this is a header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referrer-Policy)

dom-fetch-client/src/main/scala/org/http4s/dom/FetchClient.scala
```diff
@@ -20,30 +20,75 @@ package dom
...
+            referrerPolicy.foreach(init.referrerPolicy = _)
 
-              Fetch.fetch(req.uri.renderString, init)
+            init
+          } >>= {
+            requestTimeout match {
+              case d: FiniteDuration =>
+                init =>
+                  Resource
+                    .eval(F.delay(new AbortController()))
```
armanbilge, line 79: This `AbortController` is an important addition, thanks! Instead of `eval`ing it into the resource, maybe it should become part of the resource itself? I.e., in `Resource.makeCase` so that if the `Resource` is cancelled or errored we abort the request.

armanbilge:
[Reading about the `integrity` parameter](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity): 

> Subresource Integrity (SRI) is a security feature that enables browsers to verify that resources they fetch (for example, from a CDN) are delivered without unexpected manipulation. It works by allowing you to provide a cryptographic hash that a fetched resource must match.

This is _definitely_ a per-request option.

rpiaggio:
Ok on all comments. How does this look?

* `requestTimeout`: client (I think it would be trivial to support at request level too, but that may be inconsistent with other clients).
* `cache`: client & request.
* `credentials`: client & request.
* `integrity`: request.
* `keepalive`:  client & request.
* `mode`:  client & request.
* `redirect`:  client & request.
* `referrer`:  client & request.
* `referrerPolicy`:  client & request.

Although there are indeed `referrer` and `referrerPolicy` headers, they don't behave the same way when passed as parameters as when passed as headers. As far as I can tell:
* `referrer` allows to set a specific URL (which would be the same as setting the header), but it also allows other behaviors. Notably, passing `"client"` derives the `Referer` header from the current client URL.
* `referrerPolicy` seems to be a header usually used in responses to indicate when the browser will send the `Referer` header in future requests. I can't think of a use as a request header. In this case, the parameter is used to indicate when the `Referer` header will be used in the Fetch request, not to set the header.

https://w3c.github.io/webappsec-referrer-policy/#determine-requests-referrer

isomarcte:
@rpiaggio @armanbilge there are some request headers which are forbidden to be set in the `init.headers` section, `Referer` for example is one. 

https://developer.mozilla.org/en-US/docs/Glossary/Forbidden_header_name

dom-fetch-client/src/main/scala/org/http4s/dom/FetchClient.scala
rpiaggio: Good catch.

armanbilge:
> there are some request headers which are forbidden to be set in the init.headers section, Referer for example is one.

Thanks, that's a good point. I think the idiomatic thing to do here is let users set that as a header, and add logic in the client to extract it out.

armanbilge:
@rpiaggio 👍 to your list. The only other one I think should be request-only is `keepalive`, this seems like something that should be used conservatively.

> The keepalive option indicates that the request may “outlive” the webpage that initiated it.

Also setting it for the entire client may cause problems:
> We can’t send megabytes: the body limit for keepalive requests is 64KB ... This limit applies to all keepalive requests together. In other words, we can perform multiple keepalive requests in parallel, but the sum of their body lengths should not exceed 64KB.

Via https://javascript.info/fetch-api#keepalive.

Regarding `Referer`, thanks for explaining, lots of subtleties to those I missed. 👍 and maybe we should look for the header and extract its value into `init`, if the user specifies it that way?

rpiaggio:
>maybe we should look for the header and extract its value into init, if the user specifies it that way?

I'm on the fence about this. What would happen if it's both on the headers and in options?

armanbilge:
> What would happen if it's both on the headers and in options?

Good question. By options, I assume you mean the per-request attributes? Request-level config should always have priority. I think the priority should be (request attribute) > (request header) > (client config). The thinking is, a user should be able to set the `Referer` header idiomatically (as it would for any client) without needing to know that it's actually using the fetch client.

On the flipside, suppose we only set `Referer` either by client parameter or per-request attribute. What happens when there is a `Referer` header provided in the request? Silently ignore? Raise an error because it's forbidden?

rpiaggio:
> By options, I assume you mean the per-request attributes?

Yes.


>Request-level config should always have priority.

Agreed. That's how I implemented it.

> a user should be able to set the Referer header idiomatically (as it would for any client) without needing to know that it's actually using the fetch client.

Agreed that this is desirable.

>On the flipside, suppose we only set Referer either by client parameter or per-request attribute. What happens when there is a Referer header provided in the request? Silently ignore? Raise an error because it's forbidden?

In this case Fetch will do whatever it does when it receives a Referer header. Probably throw an exception.

>I think the priority should be (request attribute) > (request header) > (client config).

Hmm.. I implemented an override logic between the client and the request parameters. Prioritizing as you suggest would interfere with that logic, with the header getting in the way of the override. Would it be OK if it was `(request header) > (request attribute) > (client config)`; or even `(request attribute) > (client config) > (request header)`? Alternatively, we could raise an error if it's present both in headers and request options.

armanbilge:
@rpiaggio can't we just do option-chaining like this? (pseudocode)

```scala
request.attributes.lookup(RefererKey)
  .orElse(request.headers.get(`Referer`))
  .orElse(client.attributes.lookup(RefererKey))
```

rpiaggio:
> @rpiaggio can't we just do option-chaining like this? (pseudocode)

Yes, we definitely can add exceptional logic to handle this particular option. IMHO, header should have the least priority. We are setting other client and/or request option after all. But I'll go ahead and implement it as middle priority.

rpiaggio:
It dawned on me that we should also have a type for the `referrer` option, instead of just having a `String` as exposed in the facade. So I also added `FetchReferrer`.

rpiaggio:
An unrelated (`TomcatServerSuite`) seems to have failed in one of the setups. Is this sporadic? Can someone with enough privileges re-run its jobs?

armanbilge:
> An unrelated (TomcatServerSuite) seems to have failed in one of the setups. Is this sporadic? Can someone with enough privileges re-run its jobs?

Don't worry about it, just general flakiness :) the chrome tests are the important ones. Also firefox if you can run locally? Otherwise I'll do it before at some point.

rpiaggio:
I did a `publishLocal` and tested application where we use the snippet I copied above, which uses `.resource` and `.withCredentials`, in addition to `.create` and `.withCache` in other places. It seems to work fine.

rpiaggio:
BTW, I'm having trouble running locally the tests in `FetchClientSuite`. Is there something I should set up?

```
[error] Failed: Total 12, Failed 12, Errors 0, Passed 0
[error] Failed tests:.FetchClientSuite 0s
[error]         org.http4s.dom.FetchClientSuite
[error] (Test / testOnly) sbt.TestsFailedException: Tests unsuccessful
```

armanbilge:
Oh, oops. Yes. Run `./scripts/scaffold_server.js` and `./scripts/static_server.py` to start the servers. You also need to do in sbt: `set useJSEnv := JSEnv.Firefox` or `set useJSEnv := JSEnv.Chrome`. And install geckodriver for firefox or chromedriver for chrome 😅 

Is this ready for another look?

rpiaggio:
Thank you.

Yes, it's ready for review.

dom-fetch-client/src/main/scala/org/http4s/dom/FetchClient.scala
```diff
@@ -20,30 +20,71 @@ package dom
...
+  )(implicit F: Async[F]): Client[F] = Client[F] { (req: Request[F]) =>
     Resource.eval(req.body.chunkAll.filter(_.nonEmpty).compile.last).flatMap { body =>
       Resource
         .makeCase {
-          F.fromPromise {
-            F.delay {
-              val init = new RequestInit {}
+          F.delay(new AbortController()).flatMap { abortController =>
+            val requestOptions = req.attributes.lookup(FetchOptions.Key)
+            val mergedOptions = requestOptions.fold(options)(options.overrideWith)
```
armanbilge, line 47: Regarding the `AbortController`: I was imagining something more like this.
```scala
Resource.makeCase(F.delay(new AbortController)) {
  case (controller, Resource.ExitCase.Succeeded) => F.unit
  case (controller, Resource.ExitCase.Error(ex)) => F.delay(r.abort())
  case (controller, Resource.ExitCase.Cancelled) => F.delay(r.abort())
}.flatMap { abortController =>
  Resource.makeCase(/* fetch code in here */) { case (r, exitCase) =>
    closeReadableStream(r.body, exitCase)
  }
}.timeoutTo(d, F.raiseError(/* etc */))
```

armanbilge (changes requested):

dom-fetch-client/src/main/scala/org/http4s/dom/FetchClient.scala
armanbilge: Hmm, on second thought I see that this is flawed, sorry. 🤔

dom-fetch-client/src/main/scala/org/http4s/dom/FetchClient.scala
```diff
@@ -20,30 +20,71 @@ package dom
...
 import org.scalajs.dom.experimental.Headers
 import org.scalajs.dom.experimental.HttpMethod
 import org.scalajs.dom.experimental.RequestInit
+import org.scalajs.dom.experimental.{Response => FetchResponse}
+import scala.concurrent.duration._
 
 object FetchClient {
 
-  def apply[F[_]](implicit F: Async[F]): Client[F] = Client[F] { req: Request[F] =>
+  private[dom] def makeClient[F[_]](
```
armanbilge, line 38: Maybe we make `FetchClient` to `private[dom]`?

dom-fetch-client/src/main/scala/org/http4s/dom/FetchClientBuilder.scala
```diff
@@ -0,0 +1,146 @@
...
+  def withReferrerPolicyOption(referrerPolicy: Option[ReferrerPolicy]): FetchClientBuilder[F] =
+    copy(referrerPolicy = referrerPolicy)
+  def withReferrerPolicy(referrerPolicy: ReferrerPolicy): FetchClientBuilder[F] =
+    withReferrerPolicyOption(Some(referrerPolicy))
+  def withDefaultReferrerPolicy: FetchClientBuilder[F] =
+    withReferrerPolicyOption(None)
+
+  /** Creates a [[Client]].
+    */
+  def create: Client[F] = FetchClient.makeClient(
```
armanbilge, line 117: Name it `build`? Like `EmberClientBuilder`.

dom-fetch-client/src/main/scala/org/http4s/dom/FetchReferrer.scala
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+package org.http4s.dom
+
+import org.http4s.Uri
+import org.http4s.util.Writer
+import org.http4s.util.Renderable
+
+sealed trait FetchReferrer extends Renderable
+object FetchReferrer {
```
armanbilge, line 24: This was definitely the right thing to do :)
```suggestion
sealed abstract class FetchReferrer extends Renderable
object FetchReferrer {
```

armanbilge (commented):
I think this looks really great!! Huge leap forward from my initial implementation; I'm particularly happy you added the `AbortController`. Thanks for all of your effort 😀 

I'll run the tests against Firefox later today or tomorrow. Unfortunately a lot of the good stuff in here isn't being tested in the suite, one day though hopefully :)

dom-fetch-client/src/main/scala/org/http4s/dom/FetchClient.scala
armanbilge: Ok, I think there's still some trickyness here. So in `Resource.makeCase(acquire)(release)` acquire is `atomic` i.e. uncancelable. So even if the request is cancelled externally, it will still block until the request completes/times out. I think we want to use `makeCaseFull` and wrap the entirety of `acquire` inside of the `poll`, because it _is_ cancelable thanks to your addition of the `AbortController`.

https://typelevel.org/cats-effect/docs/typeclasses/monadcancel

dom-fetch-client/src/main/scala/org/http4s/dom/FetchReferrer.scala
rpiaggio: No problem with making this change, but I'm curious why an `abstract class` is preferable to a `trait` in this case?

dom-fetch-client/src/main/scala/org/http4s/dom/FetchClientBuilder.scala
rpiaggio: I followed the pattern from `JavaNetClientBuilder` and `OkHttpBuilder`, which have a `resource` method for building a `Resource` (the method is inherited from `BackendBuilder`), and `create` to return a `Client`.

I see that `EmberClientBuilder` does not inherit from `BackendBuilder` and `build` returns a `Resource`. Maybe we could make it a `BackendBuilder` subclass and rename `build` -> `resource` there? Let me know if you want me to add that to this PR.

dom-fetch-client/src/main/scala/org/http4s/dom/FetchClient.scala
rpiaggio: I followed the pattern from `BlazeClient` here, but yes, we can make it private. It just encapsulates the builder method.

dom-fetch-client/src/main/scala/org/http4s/dom/FetchClient.scala
rpiaggio: That's a very good catch. I admit I forgot about atomic acquires. I'll look into it.

dom-fetch-client/src/main/scala/org/http4s/dom/FetchClientBuilder.scala
isomarcte: Yeah, there are some discrepancies between the various implementations of both the client and server builders. There is some work to try to clean that up, but for now it is still kind of a free for all.

dom-fetch-client/src/main/scala/org/http4s/dom/FetchClientBuilder.scala
```diff
@@ -0,0 +1,146 @@
...
+    FetchOptions(
+      cache = cache,
+      credentials = credentials,
+      mode = mode,
+      redirect = redirect,
+      referrer = referrer,
+      referrerPolicy = referrerPolicy
+    ))
+
+  def resource: Resource[F, Client[F]] =
```
isomarcte, line 128: Can you add an `override` here?

dom-fetch-client/src/main/scala/org/http4s/dom/FetchClientBuilder.scala
```diff
@@ -0,0 +1,146 @@
...
+  */
+sealed abstract class FetchClientBuilder[F[_]] private (
+    val requestTimeout: Duration,
+    val cache: Option[RequestCache],
+    val credentials: Option[RequestCredentials],
+    val mode: Option[RequestMode],
+    val redirect: Option[RequestRedirect],
+    val referrer: Option[FetchReferrer],
+    val referrerPolicy: Option[ReferrerPolicy]
+)(implicit protected val F: Async[F])
```
isomarcte, line 48: Can you add an `override` here since IIRC we get this from the `BackendBuilder`?

dom-fetch-client/src/main/scala/org/http4s/dom/FetchOptions.scala
```diff
@@ -0,0 +1,64 @@
...
+    cache: Option[RequestCache] = None,
+    credentials: Option[RequestCredentials] = None,
+    integrity: Option[String] = None,
+    keepAlive: Option[Boolean] = None,
+    mode: Option[RequestMode] = None,
+    redirect: Option[RequestRedirect] = None,
+    referrer: Option[FetchReferrer] = None,
+    referrerPolicy: Option[ReferrerPolicy] = None
+) {
+  private[dom] def overrideWith(other: FetchOptions): FetchOptions =
```
isomarcte, line 49: Nit: I think a public `merge` function would be fine.

dom-fetch-client/src/main/scala/org/http4s/dom/FetchReferrer.scala
```diff
@@ -0,0 +1,34 @@
...
+  final case object NoReferrer extends FetchReferrer {
```
isomarcte: Nit: `final` is redundant here

dom-fetch-client/src/main/scala/org/http4s/dom/FetchReferrer.scala
```diff
@@ -0,0 +1,34 @@
...
+    override def render(writer: Writer): writer.type = writer << "no-referrer"
+  }
+  final case object Client extends FetchReferrer {
+    override def render(writer: Writer): writer.type = writer << "client"
+  }
+  final case class URL(uri: Uri) extends FetchReferrer {
```
isomarcte, line 31: Shouldn't this be just a path, since the docs say it must be a same origin URL?

> referrer
>   A USVString specifying the referrer of the request. This can be a same-origin URL, about:client, or an empty string.

@armanbilge do we have access to the host when we create the `Fetch` client? I'm still coming up to speed on Scalajs, but if we have the host then we can use that to set the `referrer` to `<HOST>/<PATH>`. If we take in a full `Uri` it will be ignored if it is not the same origin, which is probably not ideal.

dom-fetch-client/src/main/scala/org/http4s/dom/FetchReferrer.scala
```diff
@@ -0,0 +1,34 @@
...
+import org.http4s.util.Writer
+import org.http4s.util.Renderable
+
+sealed trait FetchReferrer extends Renderable
+object FetchReferrer {
+  final case object NoReferrer extends FetchReferrer {
+    override def render(writer: Writer): writer.type = writer << "no-referrer"
+  }
+  final case object Client extends FetchReferrer {
+    override def render(writer: Writer): writer.type = writer << "client"
```
isomarcte, line 29: This be `about:client`.

In firefox, setting this to `client` causes the `Referer` to be `<BASE_URL>/client`, but `about:client` sets it to `<BASE_URL>/fetch` which I believe is the intended semantic.

dom-fetch-client/src/main/scala/org/http4s/dom/FetchReferrer.scala
```diff
@@ -0,0 +1,34 @@
...
+package org.http4s.dom
+
+import org.http4s.Uri
+import org.http4s.util.Writer
+import org.http4s.util.Renderable
+
+sealed trait FetchReferrer extends Renderable
+object FetchReferrer {
+  final case object NoReferrer extends FetchReferrer {
+    override def render(writer: Writer): writer.type = writer << "no-referrer"
```
isomarcte, line 26: This be `""`, unless I'm mistaken.

In firefox, setting this to `no-referrer` causes the `Referer` to be `<BASE_URL>/no-referrer` while `""` causes the request to not have a referer header.

dom-fetch-client/src/main/scala/org/http4s/dom/FetchOptions.scala
```diff
@@ -0,0 +1,64 @@
...
+  * @param cache how the request will interact with the browser’s HTTP cache
+  * @param credentials what browsers do with credentials (cookies, HTTP authentication entries, and TLS client certificates)
+  * @param integrity subresource integrity value of the request
+  * @param keepAlive used to allow the request to outlive the page
+  * @param mode mode you want to use for the request, e.g., cors, no-cors, or same-origin
+  * @param redirect how to handle a redirect response
+  * @param referrer referrer of the request, this can be a same-origin URL, about:client, or an empty string.
+  * @param referrerPolicy referrer policy to use for the request
+  */
+final case class FetchOptions(
```
isomarcte, line 39: Would you be opposed to making this similar to [RequestPrelude](https://github.com/http4s/http4s/blob/a942a83e2b88bafc644fc9916fd1979eb951ced8/core/shared/src/main/scala/org/http4s/RequestPrelude.scala), e.g. not having the `case class` be public? `case class` encodings which aren't just arity 1 wrappers interact poorly with binary compatibility.

isomarcte (changes requested):

isomarcte:
@rpiaggio a lot of my feedback here is subjective, but I do think we need to address the `FetchReferrer` behavior discrepancy I'm seeing, e.g. `client` vs. `about:client`, `""` vs `no-referrer`.

dom-fetch-client/src/main/scala/org/http4s/dom/FetchReferrer.scala
armanbilge: @isomarcte yes, in the browser we do have access to `window.location.host` for the window and also `self.location.host` for web workers. For `FetchClient` to work in both of these browser environments we'll have to do this carefully 🤔 

But 👍 on using a path here instead, it narrows statically to only valid inputs.

dom-fetch-client/src/main/scala/org/http4s/dom/FetchClient.scala
```diff
@@ -20,30 +20,71 @@ package dom
...
-              body.foreach { body =>
-                init.body = arrayBuffer2BufferSource(body.toJSArrayBuffer)
-              }
+            val init = new RequestInit {}
 
-              Fetch.fetch(req.uri.renderString, init)
+            init.method = req.method.name.asInstanceOf[HttpMethod]
+            // Referer header is converted to a Fetch parameter below, since it's a forbidden header.
+            // See https://developer.mozilla.org/en-US/docs/Glossary/Forbidden_header_name
+            init.headers = new Headers(
```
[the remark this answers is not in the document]
armanbilge: This would be great for performance, but also makes me nervous about the one browser that does something annoying here!

dom-fetch-client/src/main/scala/org/http4s/dom/FetchReferrer.scala
armanbilge: Thanks. Eh, just my own superstitious voodoo 😆 the binary encoding and runtime behavior of `trait`s is/was slightly more complex than an `abstract class`. So I prefer to be conservative for `sealed` hierarchies since we have full control and only reach for `trait` if necessary.

dom-fetch-client/src/main/scala/org/http4s/dom/FetchClientBuilder.scala
armanbilge: Aha, I see. I tend to look to ember for what to do since my understanding is that it's the [blessed](https://github.com/http4s/http4s/discussions/4140#discussioncomment-261676) server/client implementation. Sounds like this hopefully will be cleaned up and made more uniform as 1.0 finalizes.

dom-fetch-client/src/main/scala/org/http4s/dom/FetchReferrer.scala
isomarcte: @armanbilge I'm of the same opinion as you, though I think on Scala >= 2.12.x if it is sealed then the differences are very minor between an `abstract class` and a `trait`. The story was different on 2.11.x when Scala wasn't using the JVM based default methods in interfaces.

dom-fetch-client/src/main/scala/org/http4s/dom/FetchClient.scala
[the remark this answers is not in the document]
armanbilge: [The spec](https://fetch.spec.whatwg.org/#headers-class) seems to indicate that forbidden headers will be silently ignored if appended. Ideally every browser implements the spec 🙃 

> Otherwise, if headers’s guard is "request" and name is a forbidden header name, return.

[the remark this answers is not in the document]
armanbilge: Hmm, it seems there is no guard when we do `new Headers`

> When a new Headers object is created using the Headers() constructor, its guard is set to none (the default).

https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Basic_concepts#guard

So, seems the headers _will_ be appended, and it's up to fetch itself to not balk.

dom-fetch-client/src/main/scala/org/http4s/dom/FetchOptions.scala
rpiaggio: I thought about naming it `merge` initially but it doesn't make it clear which side "wins" in case something is defined in both of them. We can still make it public.

dom-fetch-client/src/main/scala/org/http4s/dom/FetchReferrer.scala
rpiaggio: I definitely agree on this.

It may be tricky to support it correctly on all platforms though. As @armanbilge mentioned, we should proceed differently whether we are in window or web worker. Also, do we want to support `node-fetch`? (I have no idea what `node-fetch` does with the `referrer` option).

dom-fetch-client/src/main/scala/org/http4s/dom/FetchOptions.scala
armanbilge: I think as long as it's well-defined/documented it's fine. See also the `merge` functionality on vault, I forget what it's called.

dom-fetch-client/src/main/scala/org/http4s/dom/FetchClient.scala
```diff
@@ -19,31 +19,71 @@ package dom
...
+            mergedOptions.redirect.foreach(init.redirect = _)
+            // Referer headers are forbidden in Fetch, but we make a best effort to preserve behavior across clients.
+            // See https://developer.mozilla.org/en-US/docs/Glossary/Forbidden_header_name
+            // If there's a Referer header, it will have more priority than the client's `referrer` (if present)
+            // but less priority than the request's `referrer` (if present).
+            requestOptions
+              .flatMap(_.referrer)
+              .orElse(req.headers.get[Referer].map(_.uri))
+              .orElse(options.referrer)
+              .foreach(referrer => init.referrer = referrer.renderString)
```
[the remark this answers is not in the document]
armanbilge: @rpiaggio said this works ok in https://github.com/http4s/http4s/pull/5101#discussion_r696138540 without concatenating with host. Is this described in the spec by any chance?

> I'm not exactly sure how that is done with here, e.g. is it an effect or statically scoped in the runtime.

Err, I guess it's statically scoped 🤔 JS is weird and dynamic. The bigger problem is that `window.location.host` isn't actually available in web worker environments, which instead have `self.location.host` I think. So if we can avoid this messiness altogether that would be preferable.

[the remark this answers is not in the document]
armanbilge: I couldn't find any mention of this behavior in the spec.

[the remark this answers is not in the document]
rpiaggio: >Is this described in the spec by any chance?

Yes. Kind of. The spec mentions that the parameter can have a `URL from the current origin`. Unless I'm missing something, a `Path` is a valid relative URL (and thus a URL). And it's interpreted by the browser relative to the current origin, so I'd say we are safe.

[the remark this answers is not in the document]
rpiaggio: >So if we can avoid this messiness altogether that would be preferable.

I couldn't agree more.

dom-fetch-client/src/main/scala/org/http4s/dom/FetchClientBuilder.scala
```diff
@@ -0,0 +1,146 @@
...
+  *
+  * @param requestTimeout maximum duration from the submission of a request through reading the body before a timeout.
+  * @param cache how the request will interact with the browser’s HTTP cache
+  * @param credentials what browsers do with credentials (cookies, HTTP authentication entries, and TLS client certificates)
+  * @param mode mode you want to use for the request, e.g., cors, no-cors, or same-origin
+  * @param redirect how to handle a redirect response
+  * @param referrer referrer of the request
+  * @param referrerPolicy referrer policy to use for the request
+  */
+sealed abstract class FetchClientBuilder[F[_]] private (
```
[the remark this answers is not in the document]
rpiaggio: Sure. Just following the pattern of other parts of the codebase. See for example [BlazeClientBuilder](https://github.com/rpiaggio/http4s/blob/fetch-options/blaze-client/src/main/scala/org/http4s/blaze/client/BlazeClientBuilder.scala#L60).

dom-fetch-client/src/main/scala/org/http4s/dom/FetchClient.scala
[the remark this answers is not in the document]
armanbilge: > a Path is a valid relative URL (and thus a URL).

Ah, it is? Excellent then!

armanbilge:
@rpiaggio would you mind merging in or rebasing on main? I brought back CI for Firefox in #5123, would be good to run that here.

rpiaggio:
> @rpiaggio would you mind merging in or rebasing on main? I brought back CI for Firefox in #5123, would be good to run that here.

:+1: done

dom-fetch-client/src/main/scala/org/http4s/dom/FetchClient.scala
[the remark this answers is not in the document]
isomarcte: 👍

rpiaggio:
I just rebased again to keep up with main. @isomarcte would it be OK to merge?

armanbilge:
Sounds like we're getting a release train on Monday, unless there are any objections I'd like to merge this by then.

rossabaker:
I'm unqualified to review, but I trust those who did.  Merging.

isomarcte:
Sorry, @rpiaggio and @armanbilge for not getting back to you. I've been underwater for a bit.

PR #5105, opened 2021-08-24
Make charset parser case insensitive in Content-Disposition

LolHens:
When we updated a service from 0.21 to 0.23 we discovered that Requests with a Content-Disposition header containing a filename attribute with "utf-8" charset were suddenly rejected.
This is because the charset parser is case sensitive.


rossabaker (approved):
This can probably be backported to 0.22.

rossabaker:
This will be in the 0.22 and 0.23 patches this week.

PR #5106, opened 2021-08-24
ignoring error caused by client abruptly closing connection

Angel-O:
### Summary

When a client closes a web-socket connection abruptly an `IOException` is thrown by the underlying networking libraries.
The error handler in the `WebSocketHelpers` object (`ember-server`) logs an error for every error that occurs, even client errors.

This change aims to remove the noise caused by the logger.

#### Notes

- Unfortunately the error originates from a native method call and all we can do is check the error msg.

### DIFF
--- a/ember-server/src/main/scala/org/http4s/ember/server/internal/WebSocketHelpers.scala
+++ b/ember-server/src/main/scala/org/http4s/ember/server/internal/WebSocketHelpers.scala
@@ -87,7 +87,8 @@ object WebSocketHelpers {
     } yield ()
 
     handler.handleErrorWith {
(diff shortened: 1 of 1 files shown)


ChristopherDavenport:
Rather than removing it entirely I'm leaning towards trace here. I worry that being fully silent will eventually make someone wonder what happened if it silently closes with no trace.

rossabaker:
Note that parsing these broken pipe messages is a bad time in non-English locales: https://github.com/http4s/blaze/issues/292.  This is progress for a lot of people and I think we should merge it, but it's not a universal solution.

PR #5118, opened 2021-08-28
Don't block on cancellation in Http1ServerStage

### LINKED ISSUE #5116: TickWheelExecutor causes OOM
blaze-core:0.15.1
My application is based on http4s-blaze-server:0.23.0

Application crashed: `java.lang.OutOfMemoryError: Java heap space`

Heap size was slowly increasing during ~ 1 week and then OOM occurred

image

Looking at heap dump it seems TickWheelExecutor was keeping `Register` and `Cancel` objects infinitely

image

rossabaker:
In Cats-Effect 3, cancellation applies backpressure on completion of the finalizer.  If a response can't be canceled, the cancellation task does not complete.  Beginning in http4s-0.23, we started running this cancellation token synchronously in `Http1ServerStage`.  The result is that an uncancelable response can lock up the `TickWheelExecutor`'s scheduler thread.  Unhandled scheduled events may continue to pile up until OOM.

This runs cancellation asynchronously, as was done before http4s-0.23.  The test demonstrates that scheduling continues even after the cancellation of an uncancelable request.

Fixes #5116.

### DIFF
--- a/blaze-server/src/main/scala/org/http4s/blaze/server/Http1ServerStage.scala
+++ b/blaze-server/src/main/scala/org/http4s/blaze/server/Http1ServerStage.scala
@@ -39,7 +39,6 @@ import org.typelevel.vault._
 import scala.concurrent.duration.{Duration, FiniteDuration}
 import scala.concurrent.{ExecutionContext, Future}
 import scala.util.{Either, Failure, Left, Right, Success, Try}
-import scala.concurrent.Await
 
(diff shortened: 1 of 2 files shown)


rossabaker:
Oh good I made another flaky Blaze test.

It went red-to-green, reliably, on my laptop.

ChristopherDavenport (approved):

PR #5120, opened 2021-08-29
Undo scala-js-dom facade hacks

armanbilge:
As promised in https://github.com/http4s/http4s/pull/5111#issuecomment-906907327. Still not perfect unfortunately!


rossabaker (approved):
I don't understand this, but I trust a combination of you + CI on this one.

Issue #5090, opened 2021-08-18
Firefox regression in fetch client


armanbilge:
Unfortunately it seems #5046 introduced a serious regression to the `dom-fetch-client` in Firefox. The `ClientRouteTestBatterySuite` now fails every test with:

`TypeError: 'cancel' can't be called on a locked stream.`

Removing the `cancel` from the success case when closing a `ReadableStream` fixes the problem, but it's not the right thing to do in general. I'm investigating/working on a fix.

The bigger issue here is that currently CI doesn't test Firefox (I disabled this in the original Scala.js PR due to flakiness), but maybe this was a very bad idea.

armanbilge:
Also worth noting that Firefox CI causes pain in CE as well: https://github.com/typelevel/cats-effect/pull/2187

armanbilge:
Closing, because the Firefox CI problem was mostly fixed in #5123. Only "mostly" because that only runs the DOM test suites, and ideally we should still run the entire test suite in Firefox (particularly anything using crypto, since those are also browser-specific APIs).

PR #5123, opened 2021-08-29
Re-enable Firefox CI, only for DOM modules

armanbilge:
An attempt to solve the Firefox testing problem reported in #5090 while avoiding flakiness by only testing the DOM modules, which are deeply intertwined with browser APIs.


armanbilge:
I think this does the trick!

rossabaker:
Well, so did CI.

PR #5126, opened 2021-08-29
blaze-0.15.2

rossabaker:
Analogous to #5125.


rossabaker:
I accidentally pushed this while updating release notes.  I'll fix it tonight if it broke anything.  It probably hasn't synced to Maven Central yet, so it probably temporarily broke things.

PR #5134, opened 2021-08-30
Update g8 command to 0.22

sideeffffect:


rossabaker (approved):
Ah, good catch. Thank you!

Issue #5110, opened 2021-08-26
EmberServer SecureSession Request Attribute Does Not Contain Expected Values


geoffreytchew:
When I migrated from Blaze Server to Ember Server (http4s version 0.23.1), I encountered an issue where I could no longer retrieve the client's certificates used for TLS communication from the SecureSession request attribute. An empty list is returned instead of the expected certificates. Upon further inspection it seems that all SecureSession field values are empty/default values.
Consider the following code:
```
    val secureSessionInfo = req.attributes.lookup(ServerRequestKeys.SecureSession).flatten
    secureSessionInfo match {
      case Some(secureSession) =>
        println(s"@!@ CERT LIST SIZE   = ${secureSession.X509Certificate.size}")
        println(s"@!@ SSL Session ID   = ${secureSession.sslSessionId}")
        println(s"@!@ SSL Cipher Suite = ${secureSession.cipherSuite}")
        println(s"@!@ SSL Key Size     = ${secureSession.keySize}")
      case None =>
        println("Secure session information could not be obtained from request")
    }
```
When using Ember Server, the following values are printed out:
```
server @!@ CERT LIST SIZE   = 0
server @!@ SSL Session ID   =
server @!@ SSL Cipher Suite = SSL_NULL_WITH_NULL_NULL
server @!@ SSL Key Size     = 0
```
When using Blaze Server, the following values are printed out:
```
server @!@ CERT LIST SIZE   = 3
server @!@ SSL Session ID   = 097023bbc36f7b19bd52536aeabcf880de7f19e3cb00e2b7a942b14e1f9a6c02
server @!@ SSL Cipher Suite = TLS_AES_256_GCM_SHA384
server @!@ SSL Key Size     = 0
```

geoffreytchew:
I was able to reproduce this issue in unit tests that are based on `blaze-server/src/test/scala/org/http4s/blaze/server/BlazeServerMtlsSpec.scala`
I've attached these ember server unit tests in this comment.
To run the tests, do the following:
1. Unzip [EmberServerMtlsSuite.scala.zip](https://github.com/http4s/http4s/files/7078265/EmberServerMtlsSuite.scala.zip) and copy to `ember-server/src/test/scala/org/http4s/ember/server/EmberServerMtlsSuite.scala`
2. Copy `blaze-server/src/test/resources/keystore.jks` to `ember-server/src/test/resources/keystore.jks`
3. Run tests in ember-server
[...]

PR #5138, opened 2021-08-31
Fix for EmberServer SecureSession Request Attribute Does Not Contain Expected Values

geoffreytchew:
Fix for #5110. The request vault is now created for every request after the request is parsed, so that the SecureSession fields are populated with the correct values. Unit tests that confirm the desired behavior have been added. They were directly copied from the analogous Blaze Server MTLS unit tests with some minor adaptations. The same keystore is used for those tests, and was copied over to the ember-server test resources directory.

I will submit a PR for the main branch as well.


geoffreytchew:
Looks like there were some issues with code formatting. I'll format the files and submit an update.

geoffreytchew:
Not sure why this failed: `[error] (blaze-server / Test / test) sbt.TestsFailedException: Tests unsuccessful`
No changes were made to blaze-server

armanbilge:
No worries! There are some flaky tests in the suite, that's probably one of them. Thanks for the quick turnaround on the PR btw :)

Also, no need to submit a second PR against main, we'll merge series/0.23 into main after your PR is merged. Also, we may cherry-pick your changes down to series/0.22 although it's not obvious to me if the same issue exists there since it uses a different version of fs2. We'll find out.

ember-server/src/test/scala/org/http4s/ember/server/EmberServerMtlsSuite.scala
```diff
@@ -0,0 +1,163 @@
...
+    js.load(getClass.getResourceAsStream("/keystore.jks"), "password".toCharArray)
+
+    val tmf = TrustManagerFactory.getInstance("SunX509")
+    tmf.init(js)
+
+    val sc = SSLContext.getInstance("TLSv1.2")
+    sc.init(null, tmf.getTrustManagers, null)
+
+    sc
+  }
```
armanbilge, line 112: I think you can build the `TLSContext` directly via `Network[IO]`. See fs2: https://github.com/typelevel/fs2/blob/main/io/jvm/src/test/scala/fs2/io/net/tls/TLSSuite.scala

ember-server/src/test/scala/org/http4s/ember/server/EmberServerMtlsSuite.scala
```diff
@@ -0,0 +1,163 @@
...
+  */
+class EmberServerMtlsSuite extends Http4sSuite {
+  {
+    val hostnameVerifier: HostnameVerifier = new HostnameVerifier {
+      override def verify(s: String, sslSession: SSLSession): Boolean = true
+    }
+
+    // For test cases, don't do any host name verification. Certificates are self-signed and not available to all hosts
+    HttpsURLConnection.setDefaultHostnameVerifier(hostnameVerifier)
+  }
```
armanbilge, line 45: Is all of this really necessary? I'm no expert on this by any means but fs2 tests TLS without any of this. https://github.com/typelevel/fs2/blob/main/io/jvm/src/test/scala/fs2/io/net/tls/TLSSocketSuite.scala#L126

armanbilge (changes requested):
Thanks for adding the first TLS tests for ember, long overdue!

ember-server/src/test/scala/org/http4s/ember/server/EmberServerMtlsSuite.scala
geoffreytchew: I removed this code and the tests passed, so it looks like this isn't necessary. I'll push an update for this soon.

ember-server/src/test/scala/org/http4s/ember/server/EmberServerMtlsSuite.scala
geoffreytchew: This change requires the tests to be rewritten, because of the use of this call in get:
`conn.setSSLSocketFactory(sslContext.getSocketFactory)` (i.e. we need the SSL Context factory). 
However, we can't redefine `noAuthClientContext` to use `Network[IO]` because there's no way to independently
specify the keystore and truststore when using `Network[IO]` (same keystore is used for both). As a side note, in our 
work projects it's normal to have separate keystore and truststore files when creating a TLS context, so I think
this is an area of potential improvement for Network[IO]. (I'd be happy to submit a PR for that eventually).

Regardless, I did go down the path of rewriting the tests and hit what I think is another issue. 
In the original test code, we could specify the protocol version (i.e. TLSv1.2), but we can't when using `Network[IO]`. This resulted in TLSv1.3 being used, and this exposes an issue with `org.http4s.internal.tls deduceKeyLength`. It doesn't support the TLSv1.3 cipher suites it looks like. So we get a 0 for the key length.

```
@@@ 147ef8d27ed1a3c37e7ef734b106fd41556f3f21fe9082d5a5eb6d89009b3528
@@@ TLS_AES_128_GCM_SHA256
@@@ 0
```

Notice the cipher suite is TLS_AES_128_GCM_SHA256, but that's not handled explicitly in this code resulting in 0:
```
  def deduceKeyLength(cipherSuite: String): Int =
    if (cipherSuite == null) 0
    else if (cipherSuite.contains("WITH_AES_256_")) 256
    else if (cipherSuite.contains("WITH_RC4_128_")) 128
    else if (cipherSuite.contains("WITH_AES_128_")) 128
    else if (cipherSuite.contains("WITH_RC4_40_")) 40
    else if (cipherSuite.contains("WITH_3DES_EDE_CBC_")) 168
    else if (cipherSuite.contains("WITH_IDEA_CBC_")) 128
    else if (cipherSuite.contains("WITH_RC2_CBC_40_")) 40
    else if (cipherSuite.contains("WITH_DES40_CBC_")) 40
    else if (cipherSuite.contains("WITH_DES_CBC_")) 56
    else 0
}
```
I think it should be updated to include TLSv1.3 ciphers. I could write up an issue for that as well. 

So getting back to the review, I'd suggest going with the current implementation for now, and then I can
update the tests at a later date, after the deduceKeyLength issue has been addressed, or we're able to
specify the TLS version via a new version of Network[IO].fromKeystoreResource.

armanbilge: Hmm, thanks for the deep dive, that uncovered quite a bit. Yes please, definitely at least raise an issue in fs2 about this if you don't mind (pinging @mpilquist).

> I'd suggest going with the current implementation for now, and then I can update the tests at a later date

To give a little more context, the reason I'm trying so hard to avoid this is because I'm thinking ahead to when this PR gets merged into `main` where Ember now runs on both JVM and Node.js (Javascript via Scala.js). So the more of this test that can be written without referencing Java classes which are not available on Scala.js the easier that merge will be. Having a completely cross-platform test is unlikely though, but ideally we could share more code. Or maybe not, in fs2 the TLS tests are implemented twice for JVM and Node.js each which is unfortunate.

mpilquist: I think using `fromSSLContext` is right approach when there's a separate keystone and Truststore.

Re: protocols, use `TLSParameters#protocols` like here: https://github.com/typelevel/fs2/blob/d0b358c165e575474575dd871f8dfafafac204ee/io/jvm/src/test/scala/fs2/io/net/tls/TLSSocketSuite.scala#L47-L54

ember-server/src/test/scala/org/http4s/ember/server/EmberServerMtlsSuite.scala
geoffreytchew: I updated the comment above - not sure why things seemed to work before, but they clearly didn't when I tested just now. I got the error that was mentioned above.

ember-server/src/test/scala/org/http4s/ember/server/EmberServerMtlsSuite.scala
armanbilge: If you'll pardon my ignorance, why exactly is a seperate truststore necessary for this test?

ember-server/src/test/scala/org/http4s/ember/server/EmberServerMtlsSuite.scala
mpilquist: BTW we could add a `fromKeyStoreAndTrustStore` method to `TLSContextPlatform.scala`

ember-server/src/test/scala/org/http4s/ember/server/EmberServerMtlsSuite.scala
geoffreytchew: For the test `Server should handle mTLS request correctly without clientAuth`, we test `server = want client auth`, but the client's TLS context is created without a keystore (but with a truststore), so there's no key/cert to send. 

Update regarding rewriting the tests: I was able to rewrite one of the tests to use Network[IO].fromKeystoreResource and get it to work by passing a TLSParameters instance with the protocol set to TLSv1.2 to EmberServerBuilder. If I do this for the rest of them I can get rid of the hostnameVerifier code mentioned in the other comment. There'd still be the need to use `fromSSLContext` for the no client auth case mentioned above, but the tests would be closer to what you're looking for I think. 

One thing to point out about EmberClientBuilder in the 0.23 branch: It doesn't provide the ability to pass in TLSParameters, which means I don't think you could do what Mike was demonstrating with TLSContext client
Builder with the current EmberClientBuilder. It looks like this has been added in the main 1.0.0... branch though.

ember-server/src/test/scala/org/http4s/ember/server/EmberServerMtlsSuite.scala
geoffreytchew: If I rewrite the tests as described above, I can get rid of this code. My previous comment was made with the assumption that the old implementation was going to stay.

geoffreytchew: > BTW we could add a `fromKeyStoreAndTrustStore` method to `TLSContextPlatform.scala`
It would be useful in this case - One thing I notice about `fromKeyStoreResource` is that the various getInstance methods only allow the usage of the default types and algorithms (that is, the defaults are hard-coded):
```
KeyStore.getInstance(KeyStore.getDefaultType)
KeyManagerFactory.getInstance(KeyManagerFactory.getDefaultAlgorithm)
TrustManagerFactory.getInstance(TrustManagerFactory.getDefaultAlgorithm)
```
So I don't think I could use this with a different keystore type such as a LunaHSM provider, or whitecryption SKB provider. Not sure if that's going beyond the scope of what's intended here.

mpilquist:
Yeah, LunaHSM is a good use case for fromSSLContext

ember-server/src/test/scala/org/http4s/ember/server/EmberServerMtlsSuite.scala
```diff
@@ -0,0 +1,184 @@
...
+            .lookup(ServerRequestKeys.SecureSession)
+            .flatten
+            .foreach { session =>
+              assert(session.sslSessionId != "")
+              assert(session.cipherSuite != "")
+              assert(session.keySize != 0)
+              assert(session.X509Certificate.isEmpty)
+            }
+
+          Ok("success")
```
armanbilge, line 71:
```suggestion
          Ok(expectedNoAuthResponse)
```

ember-server/src/test/scala/org/http4s/ember/server/EmberServerMtlsSuite.scala
```diff
@@ -0,0 +1,184 @@
...
+      .get
+    val request = GET(uri)
+    for {
+      response <- client.fetchAs[String](request)
+    } yield assertEquals(clientCertCn, response)
+  }
+
+  fixture(TLSParameters(needClientAuth = true, protocols = List("TLSv1.2").some), noAuthTlsClientContext).test("Server should fail for invalid client auth") { case (server, client) =>
+    client.get(s"https://${server.address.getHostName}:${server.address.getPort}/dummy")(_.status.pure[IO])
+      .intercept[IOException]
```
armanbilge, line 153: I think you can call `client.status` or similar here?

armanbilge (approved):
Thanks for reworking the tests, they look great!! Just some minor suggestions but I think @rossabaker wants to get this on the next release train 😁

ember-server/src/test/scala/org/http4s/ember/server/EmberServerMtlsSuite.scala
rossabaker: Yes.  If I merge this before it gets cleaned up, it can be a separate PR.  I'm trying to get four releases out the door before I go coach soccer. :smile:

rossabaker (approved):
:+1: Looks good. Thanks for quick turnaround and review.  This should make today's train.

armanbilge:
@rossabaker I'm fixing

rossabaker:
Oh, ferfuckssake.  I can't win today.

rossabaker:
> Oh, ferfuckssake. I can't win today.

Clarification: I was ferfuckssaking myself.  Too many shortcuts taken by me.

PR #5139, opened 2021-08-31
Use http4s-crypto in core, server, client, ember-server

armanbilge:
This PR (binary- and source-compatibly) re-implements all crypto methods in core, server, and client to use the new cross-platform [typelevel/bobcats](https://github.com/typelevel/bobcats) library proposed in #5044. In doing so, we are able to purge all references to JVM-only classes that are not available on Scala.js

Although these changes are meaningless for 0.22 and 0.23, they will greatly simplify merges into 1.0 (such as #5023) where these implementations are currently "platformed" across multiple files. When this trickles up to 1.0, we'll be able to de-platform all those files and also remove the big ugly JS crypto facade currently sitting in core.

Why it would be excellent if this PR could make the next release train: to make things work, I had to binary-incompatibly change the signatures of the new public methods of the `SignatureAlgorithm` classes introduced in #5023. This is possible right now because those changes have not yet been released.


armanbilge:
Oops, I guess this missed the release train. I'll have to rework for bincompat.

rossabaker:
I'm sorry.  You were absolutely right on the timing.  If we waited for the PR queue to be empty, we'd never get a release out, so I focused on the GHSA and deferred reading this and it bit us in the ass.

armanbilge:
It's ok, it was the right call! The GHSA was very important and this was would have been last minute crypto-related changes which shouldn't be rushed imo. I'll be able to fix this, it just won't be as nice as it was before for both our merges and user migrations.

rossabaker:
That said, depending on a brand new library on the production branches is a pretty big risk.  A breaking bobcats would break us, or if we remain pinned, set up nasty diamond dependencies.  The impulse to unify the branches is good, but it would feel a lot safer on a bocats-1.0.

rossabaker:
[continued] whereas main might be an ideal place to make sure bobcats is mature.

armanbilge:
> That said, depending on a brand new library on the production branches is a pretty big risk. A breaking bobcats would break us, or if we remain pinned, set up nasty diamond dependencies.

I am very aware of this. I think for now we should treat it like blaze i.e. strongly dissuade other libraries from depending on bobcats while it stabilizes. Then bobcats can basically evolve in tandem with http4s.

ChristopherDavenport:
The only place I think this should target is 1.0. And we should get a very very stable component first for bobcats

armanbilge:
That's fair, but making this change only in 1.0 will make merges worse than the current situation.

> And we should get a very very stable component first for bobcats

How do we know when it is "very very stable"? Thorough review? Genuine question, it helps to have specific goals to achieve.

Based on this feedback, maybe this bobcats experiment doesn't make sense for http4s after all. The two goals were to (1) simplify forward merges and (2) clean up the JS crypto mess currently in 1.0. But it sounds like the risks outweigh these positives and so maybe 1.0 should continue as it is now.

armanbilge:
So I've been thinking about this, and still hope we can find a compromise that lets us achieve my goals (1) and (2) from above. IIUC yours (and my own) concerns stem from the fact that bobcats is an immature, external component.

In which case I propose we incubate bobcats as an internal http4s library (e.g. `org.http4s.crypto` or similar) make everything `private[http4s]`, and not expose any interfaces. This would mitigate all the risks, while still achieving goals (1) and (2) while giving time for the concept to incubate.

Then, when 1.0 rolls around, _if_ we feel confident about its stability, we could reconsider spinning it out as a separate typelevel lib. Or not! The goal here is first and foremost to simplify things for http4s.

armanbilge:
Any new thoughts on this before the next release train? Ok if it's a hard no.

rossabaker:
Private incubation seems reasonable.  As long as it doesn't leak into the public interface, using it to simplify the innards and prove a library sounds like pretty standard refactoring.

armanbilge:
Cool, thanks. That's basically what I've done here anyway due to binary compatibility constraints. So, the only remaining question is: incubate as a module inside http4s/http4s, or inside yet another repo e.g. http4s/crypto? Obviously inside http4s itself is easiest, but it would mean (unused) JS crypto stuff would be living in series/0.22.

rossabaker:
What do we gain over bobcats if it's in a separate repo?  That it's `private[http4s]` and can't introduce diamonds?  That's what we did with our parboiled2 fork for years, so I think that works.  It's basically an implicit MiMa exemption.

armanbilge:
Yes. Isn't that kind of how blaze is working right now? It warns away external dependencies in the README with no promises of bincompat.

rossabaker:
Blaze was presented as a standlone project through 0.14.  The original author left, and it had little uptake outside http4s, so I lopped off all the features that prevented an easy Scala 3 upgrade.  It's not so explicitly private to http4s as our parboiled2 fork was.  The safe way to do this is to ensure that the entire API is `private[http4s]`.

armanbilge:
I see. In any case we can keep all of this crypto stuff private and hidden through series/0.22 and series/0.23. Once we get to main, there is a more interesting conversation to be had whether it may make sense to expose some of those APIs but even there it's not strictly necessary.

rossabaker:
Okay, I set up http4s/http4s-crypto.  Alternatively, we could just transfer typelevel/bobcats if it's easier.  If I understand everything right, they're going to be the same thing modulo packaging.

armanbilge:
Thanks! I found it under `http4s/http-crypto`.

> Alternatively, we could just transfer typelevel/bobcats if it's easier. If I understand everything right, they're going to be the same thing modulo packaging.

I would have said so maybe a couple weeks ago, but bobcats is already finding some uses outside of http4s. So triangle dependencies were a real concern. So unfortunately I think we'll have to live with the fork for now.

armanbilge:
This is ready for another look. It uses the new [http4s-crypto](https://github.com/http4s/http4s-crypto) lib where everything is `private[http4s]` and thus exposes no internal APIs.

build.sbt
```diff
@@ -187,6 +188,12 @@ lazy val server = libraryProject("server")
   .settings(
     description := "Base library for building http4s servers",
     startYear := Some(2014),
+    mimaBinaryIssueFilters ++= Seq(
+      ProblemFilters.exclude[IncompatibleMethTypeProblem]("org.http4s.server.middleware.CSRF.this"), // private[middleware]
+      ProblemFilters.exclude[IncompatibleMethTypeProblem]("org.http4s.server.middleware.CSRF#CSRFBuilder.this"), // private[middleware]
+      ProblemFilters.exclude[DirectMissingMethodProblem]("org.http4s.server.middleware.authentication.DigestUtil.computeResponse"), // private[authentication]
+      ProblemFilters.exclude[DirectMissingMethodProblem]("org.http4s.server.middleware.authentication.DigestUtil.computeResponse"), // private[authentication]
```
armanbilge, line 195:
```suggestion
```

project/Http4sPlugin.scala
```diff
@@ -291,13 +291,13 @@ object Http4sPlugin extends AutoPlugin {
...
     val asyncHttpClient = "2.12.3"
     val blaze = "0.15.2"
-    val bobcats = "0.1-8c1dbba"
     val boopickle = "1.4.0"
     val caseInsensitive = "1.1.4"
     val cats = "2.6.1"
     val catsEffect = "2.5.3"
     val catsParse = "0.3.4"
     val circe = "0.14.1"
+    val crypto = "0.1-019ed6f"
```
rossabaker, line 301: It doesn't really matter in this particular case because this is the only repo that should use it, but  it freaks me out to see a hash release in production code.  Can we give this a proper monotonic number?

armanbilge: Absolutely. I was going to wait until we are happy with everything here before cutting a release but I can do that now.

project/Http4sPlugin.scala
```diff
@@ -298,6 +298,7 @@ object Http4sPlugin extends AutoPlugin {
...
```
armanbilge, line 301:
```suggestion
    val crypto = "0.1.0"
```

project/Http4sPlugin.scala
armanbilge: Done.

server/src/main/scala/org/http4s/server/middleware/CSRF.scala
```diff
@@ -340,7 +343,8 @@ object CSRF {
     def withOnFailure(onFailure: Response[G]): CSRFBuilder[F, G] = copy(onFailure = onFailure)
     def withCreateIfNotFound(createIfNotFound: Boolean): CSRFBuilder[F, G] =
       copy(createIfNotFound = createIfNotFound)
-    def withKey(key: SecretKey): CSRFBuilder[F, G] = copy(key = key)
+    def withKey(key: crypto.SecretKey): CSRFBuilder[F, G] =
```
rossabaker, line 346: I'm surprised and suspicious that this passed MiMa.

It's also introducing a `private[http4s]` into the public API.  How is anybody else going to be able to call it?

armanbilge: Ooph, this is just rough on the eyes I think. But it passes MiMa because it actually refers to `import javax.crypto`.

armanbilge: Let me write those out as FQCN so we can sleep easy.

armanbilge: There. `crypto` made more sense when the package was `bobcats` and not `crypto`.

rossabaker (approved):

hamnis (approved):

rossabaker:
We'll do two release trains this week so I can stabilize the GHSA merges.

PR #5144, opened 2021-09-02
Fix CORS preflight status

### LINKED ISSUE #5143: Behaviour change: CORSPolicy middleware returns 404 on preflight request
In the latest release, `CORSConfig` was deprecated in favour of using `CORSPolicy`. This has introduced a behaviour change where preflight requests will now return a 404 instead of a 200. I think this may cause some (all?) browsers to reject preflight requests, but I'm not sure. 

In 0.23.1 was doing the following:

```scala
val service: HttpRoutes[F] = ???

private val corsConfig = CORSConfig.default
    .withAllowedMethods(Some(Set(Method.POST, Method.GET, Method.OPTIONS)))
    .withAllowedHeaders(Some(Set("Authorization", "Content-Type", "Accept")))
    .withAllowCredentials(false)

EmberServerBuilder
      .default[F]
      .withHost(host)
      .withPort(port)
      .withHttpApp(CORS(service, corsConfig).orNotFound)
      .build
```

and then an `OPTIONS` request with headers `"Access-Control-Request-Method" ="POST"` `"Origin"="http://example.com"`,  `"Access-Control-Request-Headers"="authorization,content-type,accept"` would return `200` with the appropriate CORS headers. 

Following the deprecation of `CorsConfig` in 0.23.2, I replaced `corsConfig` like so:

```scala
private val cors = CORS.policy
    .withAllowMethodsIn(Set(Method.POST, Method.GET))
    .withAllowHeadersIn(Set(ci"Authorization", ci"Content-Type", ci"Accept"))
    .withAllowCredentials(false)

EmberServerBuilder
      .default[F]
      .withHost(host)
      .withPort(port)
      .withHttpApp(cors(service).orNotFound)
      .build
```

But now the same preflight request returns a 404. I can see that this is because in the implementation of `CorsPolicy#apply`, `service: HttpRoutes[F]` is being applied to the incoming request and the preflight headers added to the response, here: 
https://github.com/http4s/http4s/blob/5ebfb746ff98817959cbcce3ab2520342253c608/server/src/main/scala/org/http4s/server/middleware/CORS.scala#L413

So because my service doesn't have an endpoint defined for an `OPTIONS` request, it falls through and I get a 404. 

In the previous implementation an options response was created, as seen here: 

https://github.com/http4s/http4s/blob/5ebfb746ff98817959cbcce3ab2520342253c608/server/src/main/scala/org/http4s/server/middleware/CORS.scala#L251

which would always return a 200.

rossabaker:
Fixes #5143.

### DIFF
--- a/server/src/main/scala/org/http4s/server/middleware/CORS.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/CORS.scala
@@ -18,7 +18,7 @@ package org.http4s
 package server
 package middleware
 
-import cats.{Applicative, Functor, Monad}
+import cats.{Applicative, Monad}
 import cats.data.{Kleisli, NonEmptyList}
 import cats.syntax.all._
 import org.http4s.Method.OPTIONS
@@ -214,7 +214,7 @@ sealed class CORSPolicy(
 ) {
   import CORSPolicy._
 
-  def apply[F[_]: Functor, G[_]](http: Http[F, G]): Http[F, G] = {
+  def apply[F[_]: Applicative, G[_]](http: Http[F, G]): Http[F, G] = {
 
     val allowCredentialsHeader: Option[Header] =
       allowCredentials match {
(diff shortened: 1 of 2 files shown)


server/src/main/scala/org/http4s/server/middleware/CORS.scala
```diff
@@ -214,7 +214,7 @@ sealed class CORSPolicy(
 ) {
   import CORSPolicy._
 
-  def apply[F[_]: Functor, G[_]](http: Http[F, G]): Http[F, G] = {
+  def apply[F[_]: Applicative, G[_]](http: Http[F, G]): Http[F, G] = {
```
rossabaker, line 217: Tests pass, but this breaks binary compatibility.  I don't think we can implement it correctly on Functor.

rossabaker: I hacked around this.  Somewhat extreme lengths for a day-old signature, but I like to not make exceptions.

rossabaker:
Failure is unrelated.

/cc @jbwheatley @bplommer

ChristopherDavenport (approved):

PR #5147, opened 2021-09-02
munit-0.7.29, scodec-bits-1.1.28

rossabaker:
A couple dependency bumps as the release train rolls on.


rossabaker:
The Actions queue is a mile long and this error is fixed on the branch.  I'm going to gamble.

PR #5155, opened 2021-09-03
Fix merge of #5138 into 1.0

armanbilge:
JVM only for now. I'd really like to get these running Node.js, not so much for ember-server but to test TLS for ember-client. But the big yak to shave first is how to parse SSLSession data on JS.


rossabaker:
Did I break main?  I got a clean CI, but I see one of the builds has been hung for over an hour, so we haven't successfully released.

armanbilge:
No, you didn't break anything, at least not with this. It just got dumped into a folder where it was ignored. I'll check what main is doing.

armanbilge:
No idea why main is stuck, the JS runners used to get hung sometimes (although seems better now) but its a JVM runner that's stuck right now. Unfortunately can't see what it was doing.

PR #5158, opened 2021-09-03
Support monitoring of the blaze-client pool

rossabaker:
* Exposes some basic metrics on the client pool for monitoring.  Our use case is to see that the wait queue is backing up before it pops.
* Introduces a `ConnectionManager.Stateful`.  The only production implementation, `PoolManager`, is this. 
* Similar to `EmberClient`, though I didn't name the tuple. 
* Less similar to `AsyncHttpClientStats`, because we don't retain an underlying backend object to obtain stats separately


zmccoy (approved):
Thanks for doing this!  🎊

PR #5160, opened 2021-09-03
scalafmt-3.0.2 on 0.21

rossabaker:


rossabaker:
This version isn't maintained except for security fixes, but we don't want gratuitous merge conflicts when they do come up.

PR #5161, opened 2021-09-03
scalafmt-3.0.2 on 0.21

rossabaker:


rossabaker:
Breaks things on 0.22 and I'm just not going to deal with it.

Issue #5010, opened 2021-08-03
Add models for unmodelled headers


bplommer:
Quite a few headers don't yet have typed models. The following are the unmodelled headers that have their own page on [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers), excluding those marked as deprecated.

- [ ] Accept-CH
- [ ] Accept-Post
- [ ] Access-Control-Allow-Methods
- [ ] Access-Control-Allow-Origin
- [ ] Access-Control-Max-Age
- [ ] Access-Control-Request-Headers
- [ ] Access-Control-Request-Method
- [ ] Alt-Svc
- [ ] Clear-Site-Data
- [ ] Content-Security-Policy-Report-Only
- [ ] Content-Security-Policy
- [ ] Cross-Origin-Embedder-Policy
- [ ] Cross-Origin-Opener-Policy
- [ ] Cross-Origin-Resource-Policy
- [ ] Device-Memory
- [ ] Digest
- [ ] DNT
- [ ] Downlink
- [ ] Early-Data
- [ ] ECT
- [ ] Expect-CT
- [ ] Expect
- [ ] Feature-Policy
- [ ] From
- [ ] If-Range
- [ ] Keep-Alive
- [ ] Large-Allocation
- [ ] NEL
- [ ] Referrer-Policy
- [ ] RTT
- [ ] Save-Data
- [ ] Sec-Fetch-Dest
- [ ] Sec-Fetch-Mode
- [ ] Sec-Fetch-Site
- [ ] Sec-Fetch-User
- [ ] Server-Timing
- [ ] SourceMap
- [ ] TE
- [ ] Timing-Allow-Origin
- [ ] Tk
- [ ] Trailer
- [ ] Upgrade-Insecure-Requests
- [ ] Vary
- [ ] Via
- [ ] Want-Digest
- [ ] Warning
- [ ] X-Content-Type-Options
- [ ] X-DNS-Prefetch-Control
- [ ] X-Forwarded-Host
- [ ] X-Forwarded-Proto
- [ ] X-Frame-Options
- [ ] X-XSS-Protection

rossabaker:
Thanks, @bplommer.  This is a great opportunity for anybody who wants to get their hands dirty with cats-parse.  Examples abound in the [headers package](https://github.com/http4s/http4s/tree/main/core/src/main/scala/org/http4s/headers), and I will be happy to coach anyone through.

If you'd like to work on these, please:
[...]

PR #5165, opened 2021-09-05
Keep-Alive Header

zmccoy:
This adds the Keep-Alive Header model which fixes part of https://github.com/http4s/http4s/issues/5010.  

Happy to change whatever may be needed.


rossabaker:
Oh, right.  New arbitraries need to be added in the companion object for bincompat reasons.

zmccoy:
Doh, I'll get that fixed up here tonight.

core/src/main/scala/org/http4s/headers/Keep-Alive.scala
```diff
@@ -0,0 +1,172 @@
...
+keep-alive-info      =   "timeout" "=" delta-seconds
+                       / "max" "=" 1*DIGIT
+                       / keep-alive-extension
+keep-alive-extension = token [ "=" ( token / quoted-string ) ]
+   */
+
+  def apply(
+      timeoutSeconds: Option[Long],
+      max: Option[Long],
+      extension: List[(String, Option[String])]): ParseResult[`Keep-Alive`] =
```
rossabaker, line 51: It is [an error for MIME types](https://datatracker.ietf.org/doc/html/rfc6838#page-13) to repeat parameters.  Parameters must be unique in their context in [structured fields](https://datatracker.ietf.org/doc/html/rfc8941#section-3.1.2).  I can't find anything in the `Cache-Control` spec.  I can't think of any other header whose parameters take this form that speciflcally allow duplicates, but our specification didn't _disallow_ it.

The spec is also silent on whether the keys are case sensitive or not.  

I think we did it reasonably, but we're left to guess on all of this. :weary:

core/src/main/scala/org/http4s/headers/Keep-Alive.scala
```diff
@@ -0,0 +1,172 @@
...
+  def unsafeApply(
+      timeoutSeconds: Option[Long],
+      max: Option[Long],
+      extension: List[(String, Option[String])]): `Keep-Alive` =
+    apply(timeoutSeconds, max, extension).fold(throw _, identity)
+
+  def parse(s: String): ParseResult[`Keep-Alive`] =
+    ParseResult.fromParser(parser, "Invalid Keep-Alive header")(s)
+
+  private[http4s] def nonNegativeLong(l: Long, fieldName: String): ParseResult[Long] =
```
rossabaker, line 69: Why `http4s`?  You can make this private, and we'll get nice compiler warnings if it's ever unused.

core/src/main/scala/org/http4s/headers/Keep-Alive.scala
```diff
@@ -0,0 +1,172 @@
...
+import org.http4s.Header
+import org.typelevel.ci._
+import scala.concurrent.duration.FiniteDuration
+import java.util.concurrent.TimeUnit
+import scala.collection.mutable.ListBuffer
+import cats.kernel.Semigroup
+
+object `Keep-Alive` {
+
+  sealed trait KeepAlive
```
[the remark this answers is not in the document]
rossabaker: Well, wait.  You don't expose this anywhere.  Could it be private?

core/src/main/scala/org/http4s/headers/Keep-Alive.scala
rossabaker: What if I'm a big jerk and call this `Keep-Alive(None, None, List("timeout", "30".some))`?

tests/src/test/scala/org/http4s/headers/KeepAliveSuite.scala
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  test(
+    "parse keep-alive with timeout, max, and extensions with 3 keys (extKey as a token) some have values and one does not") {
+    assertEquals(
+      Header[`Keep-Alive`].parse("""timeout=3, max=33, extKey=il, nextKey="pi", thirdKey="""),
+      `Keep-Alive`(
+        Some(3),
+        Some(33),
+        List(("extKey", Some("il")), ("nextKey", Some("pi")), ("thirdKey", None)))
```
rossabaker, line 68: I think this should be `Some("")`.  Don't you want to test `thirdKey` without the equals sign?

rossabaker (commented):
We've bikeshedded the hell out of this for what is a minor header but it's full of major interesting general implications.  I thought of a couple more. :weary:

core/src/main/scala/org/http4s/headers/Keep-Alive.scala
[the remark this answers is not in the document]
zmccoy: Yeah, I'll get this to be private.  It's meant only as an internal representation.

core/src/main/scala/org/http4s/headers/Keep-Alive.scala
zmccoy: @rossabaker Think just a heads up on the decisions we've made in a comment is good on this?  

Maybe you shouldn't be a big jerk ;)

tests/src/test/scala/org/http4s/headers/KeepAliveSuite.scala
zmccoy: You found a bug!

core/src/main/scala/org/http4s/headers/Keep-Alive.scala
rossabaker: Well, without some effort, it won't round trip.  In addition to the handwaving about how extension keys shouldn't be non-tokens, they shouldn't be the two known tokens.  We can fix it in this constructor, but it's a headache.

I think we have a similar problem with CacheDirectives today: I don't remember what prevents a custom one from duplicating a modeled one.

zmccoy: That's a good point that the round trip ends up with different case class representations.  😵‍💫 .

core/src/main/scala/org/http4s/headers/Keep-Alive.scala
```diff
@@ -0,0 +1,172 @@
...
+  final case class Timeout(timeoutSeconds: Long) extends KeepAlive
+  final case class Max(max: Long) extends KeepAlive
```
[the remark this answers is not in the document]
zmccoy: This should be changed to private and renamed in https://github.com/http4s/http4s/pull/5165/commits/a36b6cd9fb80306acb8b776270d59df6562094c4

tests/src/test/scala/org/http4s/headers/KeepAliveSuite.scala
zmccoy: This should be fixed here now:  https://github.com/http4s/http4s/pull/5165/commits/a36b6cd9fb80306acb8b776270d59df6562094c4

and in https://github.com/http4s/http4s/pull/5165/commits/a0cc9c13ef0cc298919a63d588611267ffdbd7a9 to fix the tests.  Doing "nextKey=" is now invalid as it should have been before and would result in a parse failure.

core/src/main/scala/org/http4s/headers/Keep-Alive.scala
zmccoy: I've fixed this up to not allow the reserved tokens in the extension list in the apply with a applicable error.  I did not do anything with the mime types repeat params, case sensitivity etc.

Issue #5153, opened 2021-09-02
deduceKeyLength returns 0 for TLSv1.3 Cipher Suites


geoffreytchew:
While writing tests for another issue, I noticed that org.http4s.internal.tls deduceKeyLength is returning 0 for the key length when a TLSv1.3 cipher suite is specified. The function doesn't support the TLSv1.3 cipher suites, so the default value of 0 is returned for them. 

Also, the comment for the function references version 1.0 of the TLS spec, but it should also be updated to reference the 1.1, 1.2, and 1.3 versions as well. The function should be updated to handle ciphers from versions 1.0 through 1.3. A test should be added to confirm the expected behavior.

geoffreytchew:
I'll be happy to take this one if someone wants to assign it to me.

rossabaker:
:+1: I should have asked on the other ticket: is this also relevant in Cats-Effect 2?  I can try a backport if it is.  I got immersed in the CORS work.

I anticipate another release train next week.

geoffreytchew:
> 👍 I should have asked on the other ticket: is this also relevant in Cats-Effect 2? I can try a backport if it is. I got immersed in the CORS work.
> 
> I anticipate another release train next week.

Are you asking if the other issue (#5110) is relevant in cats effect 2? The bug was in EmberServer's ServerHelpers.scala at least as far back as 0.21.x.

As far as this issue, I also found it as far back as 0.21.x. It affects both the blaze and ember servers.

rossabaker:
Yes, #5011. Sorry, I meant to reference it by number.  We're generally starting on series/0.22 with anything that can be done without breaking compatibility, and merging it forward.  I'll look into backporting #5110.  series/0.21 is only getting security patches now.

series/0.22 would be the ideal place to start this one, but I'll look at (and maybe cherry-pick) a PR to any branch.

geoffreytchew:
I can submit a PR to the series/0.22 branch

geoffreytchew:
Just submitted the following PR that addresses this issue: https://github.com/http4s/http4s/pull/5166

armanbilge:
Fixed in https://github.com/http4s/http4s/pull/5166.

PR #5166, opened 2021-09-07
deduceKeyLength returns 0 for TLSv1.3 Cipher Suites #5153

geoffreytchew:
Addressed the original issue that prevented TLSv1.3 cipher suites from being properly handled and additionally added support for other cipher suites that were not previously handled, based on the list from the IANA TLS Cipher Suites registry (IANA https://www.iana.org/assignments/tls-parameters/tls-parameters.txt). Also corrected the function for "WITH_DES_CBC_40" ciphers (the key size should be 40 but previously 56 was being returned). Also added a unit test to confirm expected behavior.


tests/src/test/scala/org/http4s/internal/TlsSuite.scala
```diff
@@ -0,0 +1,381 @@
...
+    "TLS_ECDHE_PSK_WITH_AES_256_GCM_SHA384" -> 256,
+    "TLS_ECDHE_PSK_WITH_AES_128_CCM_8_SHA256" -> 128,
+    "TLS_ECDHE_PSK_WITH_AES_128_CCM_SHA256" -> 128
+  )
+
+  test("deduceKeyLength should return expected results for known cipher suites") {
+    cipherSuiteMap.foreach { case (name, expectedValue) =>
+      assertEquals(tls.deduceKeyLength(name), expectedValue)
+    }
+  }
```
armanbilge, line 380: Just a minor suggestion: what if we do something like this, so that if a test fails its easy to see which cipher it was.
```suggestion
  cipherSuiteMap.foreach { case (name, expectedValue) =>
    test(s"deduceKeyLength should return expected result for $name") {
      assertEquals(tls.deduceKeyLength(name), expectedValue)
    }
  }
```

armanbilge (changes requested):
Thanks for the fix and all your effort compiling a comprehensive test suite! LGTM, just a small suggestion.

armanbilge (approved):

PR #5171, opened 2021-09-09
Addition of the Access-Control-Max-Age header

rcardin:
Here it's the changes that introduce management of the header Access-Control-Max-Age.
I'm still missing some parsing internals, but I think it could be a good starting point.

Any feedback will be appreciated :)

The reference issue is #5010

### DIFF
--- a/core/src/main/scala/org/http4s/parser/AdditionalRules.scala
+++ b/core/src/main/scala/org/http4s/parser/AdditionalRules.scala
@@ -28,4 +28,11 @@ private[http4s] object AdditionalRules {
       case _: NumberFormatException => None
     }
   }
+
+  val Long: P[Long] = Rfc7230.token.mapFilter { s =>
+    try Some(s.toLong)
+    catch {
+      case _: NumberFormatException => None
+    }
+  }
 }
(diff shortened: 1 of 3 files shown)


core/src/main/scala/org/http4s/parser/AdditionalRules.scala
```diff
@@ -28,4 +28,11 @@ private[http4s] object AdditionalRules {
       case _: NumberFormatException => None
     }
   }
+
+  val Long: P[Long] = Rfc7230.token.mapFilter { s =>
+    try Some(s.toLong)
+    catch {
+      case _: NumberFormatException => None
+    }
+  }
```
rossabaker, line 37: Is this going to accept negative values?  If you start from `digits` instead of `token`, you guard against that, and then the try/catch takes care of (unlikely) overflows.

rossabaker: Oh, I see.  You're parsing everything as a long, and then running it through the `fromLong` constructor.  That'll work.

rossabaker (approved):

PR #5175, opened 2021-09-10
Add helper to reduce header boilerplate

bplommer:
With a random selection of usage examples


bplommer:
Converted to draft pending bincompat fix

rossabaker (approved):
I like it.  Inheritance hasn't always been popular around here, so I'll wait for a second approval, but I think it promotes consistency in this case.

PR #5179, opened 2021-09-10
Fix scala 3 compiler warnings

bplommer:


rossabaker (approved):
Is there anything preventing us from enabling fatal warnings on Scala 3 so we catch these as they come in?

armanbilge (approved):

armanbilge:
Does scala 3 have fatal warnings available already? Lots of warnings haven't been implemented so far.

rossabaker:
Oh, maybe not.  My only Scala 3 experience is maintaining libraries. :shrug:

PR #5190, opened 2021-09-13
Add effectual classifier to the Metrics middleware;

FunFunFine:
Sometimes it is required to inspect the body of the request to properly classify one.
For example, [JSON RPC](https://www.jsonrpc.org/specification) has all requests identical to each other on the HTTP layer.

Since the body is a `Stream[F, Byte]`, the inspection would be an effect (Stream has to be compiled and so on), so that classifier function has to have the type of `Request[F] => F[Option[String]]`. This PR adds the `effect` method with such a classifier function without breaking forward compatibility (I expect it to not t do so).

### DIFF
--- a/client/shared/src/main/scala/org/http4s/client/middleware/Metrics.scala
+++ b/client/shared/src/main/scala/org/http4s/client/middleware/Metrics.scala
@@ -51,12 +51,23 @@ object Metrics {
       classifierF: Request[F] => Option[String] = { (_: Request[F]) =>
         None
       })(client: Client[F])(implicit F: Clock[F], C: Concurrent[F]): Client[F] =
+    Client(withMetrics(client, ops, classifierF(_).pure[F]))
+
+  /** Wraps a [[Client]] with a middleware capable of recording metrics
+    *
+    * @param ops a algebra describing the metrics operations
+    * @param classifierF a function that allows to add a classifier that can be customized per request
+    * @param client the [[Client]] to gather metrics from
+    * @return the metrics middleware wrapping the [[Client]]
+    */
+  def effect[F[_]](ops: MetricsOps[F], classifierF: Request[F] => F[Option[String]])(
+      client: Client[F])(implicit F: Clock[F], C: Concurrent[F]): Client[F] =
     Client(withMetrics(client, ops, classifierF))
 
   private def withMetrics[F[_]](
       client: Client[F],
       ops: MetricsOps[F],
-      classifierF: Request[F] => Option[String])(
+      classifierF: Request[F] => F[Option[String]])(
       req: Request[F])(implicit F: Clock[F], C: Concurrent[F]): Resource[F, Response[F]] =
     for {
       statusRef <- Resource.eval(C.ref[Option[Status]](None))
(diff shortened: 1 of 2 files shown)


client/shared/src/main/scala/org/http4s/client/middleware/Metrics.scala
```diff
@@ -74,27 +85,29 @@ object Metrics {
...
-      classifierF: Request[F] => Option[String],
+      classifierF: Request[F] => F[Option[String]],
       req: Request[F],
       statusRef: Ref[F, Option[Status]],
       start: Long
   )(implicit F: Clock[F], C: Concurrent[F]): Resource[F, Response[F]] =
     (for {
-      _ <- Resource.make(ops.increaseActiveRequests(classifierF(req)))(_ =>
-        ops.decreaseActiveRequests(classifierF(req)))
+      classifier <- Resource.eval(classifierF(req))
```
rossabaker, line 94: What if this fails?  I guess we just won't record metrics.  Maybe this is the responsibility of the caller of `effect` to handle any errors and make a default `classifier`.  That's probably worth noting in the scaladoc.

rossabaker (commented):
Thanks for this!  I think with the care for compatibility, we could conceivably backport this to 0.22.

If you're consuming the request body, I'm concerned about recompiling that body in the wrapped client.  Compiling the body twice is one of the easiest errors to make in http4s that the compiler can't catch.  Would you envision using this with something that cached the body and passed a strict version on downstream?

FunFunFine:
> Would you envision using this with something that cached the body and passed a strict version on downstream?

Actually, in our use case requests bodies are pretty small and it's not that bad to compile the body twice, I suppose.
Nevertheless, for other cases, caching is a must IMO, but I'm afraid that would require some new machinery like `CompiledRequest[F]` with not a Stream body, but something else (I can't even imagine what type it should have).

Creating such caching mechanism is out of my league so I'll leave a note in the docs about recompiling the body, and leave the former for someone else.

rossabaker:
The reason I ask is that compiling the body twice leads to surprises when your source is effectful.  When the source is a socket, compiling doesn't rewind the socket.  Each byte only gets consumed once.  Your unit tests are likely to behave like `strictSource`, whereas your real client backed by real sockets will behave like `readOnceSource`.  Think of `A` as your classifier and `B` as your business logic:

```scala
scala> def compileTwice[F[_]: Concurrent, A](s: Stream[F, A]) = for { a <- s.compile.toVector; b <- s.compile.toVector } yield (s"A = ${a}, B = ${b}")
compileTwice: [F[_], A](s: fs2.Stream[F,A])(implicit evidence$1: cats.effect.Concurrent[F])F[String]

scala> val strictSource = Stream.emits("foo".getBytes("UTF-8").toSeq)
strictSource: fs2.Stream[[x]fs2.Pure[x],Byte] = Stream(..)

scala> val readOnceSource = Stream.fromIterator[IO]("foo".getBytes("UTF-8").toIterator, 1)
readOnceSource: fs2.Stream[cats.effect.IO,Byte] = Stream(..)

scala> compileTwice[IO, Byte](strictSource).unsafeRunSync()
res10: String = A = Vector(102, 111, 111), B = Vector(102, 111, 111)

scala> compileTwice[IO, Byte](readOnceSource).unsafeRunSync()
res11: String = A = Vector(102, 111, 111), B = Vector()
```

FunFunFine:
Ah, I see now. Should I do something about it in this PR then?

rossabaker:
I think it's fine to use as is as long as you're _not_ consuming the request body, but that's your use case and probably most other people's.  I think we can proceed if we document that it's unsafe to consume the body unless wrapped in another middleware that caches it.

rossabaker:
I keep sketching out that hypothetical middleware for people, but I don't remember that we've ever written it. :thinking:

FunFunFine:
I've also added the same effect method for server `Metrics` middleware since the PR is kind of not complete without it.

rossabaker (approved):

rossabaker:
Gack.  I was moving too fast and cherry-picked this to series/0.23 before CI finished.  I'll fix it.

This will be in the next 0.23 release.  I tried to bring it back to 0.22, but it was a mess.  I'll leave that one too any interested party if they need it.

PR #5194, opened 2021-09-15
Expand scaladoc for Query

bplommer:
I think it would be good to have optional manual control over percent-encoding, but documenting the existing behaviour is a start.

By the way, what does "modulo the identity of separators" mean?

### DIFF
--- a/core/src/main/scala/org/http4s/Query.scala
+++ b/core/src/main/scala/org/http4s/Query.scala
@@ -33,10 +33,10 @@ import scala.collection.mutable.ListBuffer
 /** Collection representation of a query string
   *
   * It is a indexed sequence of key and maybe a value pairs which maps
-  * precisely to a query string, modulo 
-  * [[https://datatracker.ietf.org/doc/html/rfc3986#section-2.1 percent-encoding]] 
+  * precisely to a query string, modulo
+  * [[https://datatracker.ietf.org/doc/html/rfc3986#section-2.1 percent-encoding]]
   * and the identity of separators.
-  * 
+  *
   * The internal representation is not percent-encoded; percent-encoding is
   * applied when rendering and decoded when parsing. For example, the strings
   * "://" and "%3A%2F%2F" will both be parsed as "://" and rendered as "%3A//".
--- a/core/src/main/scala/org/http4s/Query.scala
+++ b/core/src/main/scala/org/http4s/Query.scala
@@ -34,12 +34,7 @@ import scala.collection.mutable.ListBuffer
   *
   * It is a indexed sequence of key and maybe a value pairs which maps
   * precisely to a query string, modulo
-  * [[https://datatracker.ietf.org/doc/html/rfc3986#section-2.1 percent-encoding]]
-  * and the identity of separators.
-  *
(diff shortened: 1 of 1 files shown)


bplommer:
> Doesn't the `Right` give full control over percent encoding by storing the raw value, and then the doc is true of the left? That was the intent in making it wrap an `Either`.

Ah yes - but that only gets instantiated by `query.parser`, which is only used in the URI parser. So you can only instantiate the `Right` version via a parsed URI. All the parsing methods that are exposed by `Query` itself create the `Left` version. That makes it fiddly if you're trying to work with over-specific requirements from a third-party API about what they receive. I think `unsafeFromString` should use the same parser as the URI parser.

> This was a binary compatibility hack, and Query could probably just be a sum type in 1.0 and eliminate a wrapping class.

How about making `KeyValue` a sum type so "decoded" and "raw" segments can be mixed in a single query?

rossabaker:
> [The encoded version] only gets instantiated by query.parser, which is only used in the URI parser.

Coming back to the PR briefly :laughing:, I think the scaladoc here isn't true to all versions of the internals, but is effectively true to users, and is still valid?

> I think unsafeFromString should use the same parser as the URI parser.

Agreed.

> How about making KeyValue a sum type so "decoded" and "raw" segments can be mixed in a single query?

I think this is a great idea.  I recall the current encoding is something @hamnis pulled out of the fire when things were broken and binary compatibility limited our choices, but I think we can do better still in the context of `main`.

bplommer:
> Coming back to the PR briefly 😆, I think the scaladoc here isn't true to all versions of the internals, but is effectively true to users, and is still valid?

I don't think so - firstly because it doesn't apply for the parsing methods on `Query`, secondly because any use of `QueryOps` converts the encoded version to the decoded version:
```scala
@  uri"?%2F"
res37: Uri = Uri(scheme = None, authority = None, path = , query = %2F, fragment = None)

@ res37 +? "hi"
res38: res37.Self = Uri(scheme = None, authority = None, path = , query = /&hi, fragment = None)
```

>> I think unsafeFromString should use the same parser as the URI parser.
>
> Agreed.

Do you think that's a bugfix we can do in 0.22/0.23 or a breaking change?

rossabaker:
> Do you think that's a bugfix we can do in 0.22/0.23 or a breaking change?

Tough question.  The existing one turns parse failures into `Query.empty`.  I can't remember why that was ever a good idea, but people may be depending on that.  We could argue that the method is called `unsafe`, but a different semantic change in `Uri` caused @danicheg some significant pain recently, which makes me nervous to make any more.

Is there a different name we could give this one, and perhaps deprecate the current one?  It's a shame, because the existing name is consistent with the desired new semantics, but that's probably the kindest way to our 0.22 / 0.23 users.

danicheg:
:+1: to not change semantic on `Uri` parsing :)

bplommer:
I was thinking of still parsing an invalid query as empty (rather than throwing) but using the plain string representation for valid queries so formatting is preserved.

bplommer:
> I don't understand that modulo statement either.

Oh, I think it means '&' vs (the illegal but somethimes-used) `;`. The same caveats apply for that as for percent-encoding.

PR #5195, opened 2021-09-15
Bring back EntityDecoder[F, ByteVector]

sideeffffect:
As there used to be one https://github.com/http4s/http4s/blob/v0.16.6/core/src/main/scala/org/http4s/EntityDecoder.scala
Should I create a PR for individual `series/x` branches, or will this be propagated there from `master`?


rossabaker:
The progression is series/0.22 -> series/0.23 -> main, but I can cherry-pick.

rossabaker:
This will be in 0.22.  I'll need to adjust the constraints back to what you had when it merges forward.

PR #5200, opened 2021-09-15
New representation for Query

bplommer:
As discussed in #5194


core/shared/src/main/scala/org/http4s/Query.scala
```diff
@@ -171,24 +146,123 @@ final class Query private (value: Either[Vector[KeyValue], String])
...
+    def separator: Char = '&'
+    def key: String
+    def value: Option[String]
+
+    def normalize: Component
+
+    override def equals(that: Any): Boolean = that match {
+      case that: Component => that.renderString == renderString
+      case _ => false
+    }
```
bplommer, line 159: This can be optimized

core/shared/src/main/scala/org/http4s/Query.scala
```diff
@@ -171,24 +146,123 @@ final class Query private (value: Either[Vector[KeyValue], String])
...
+    sealed abstract class KeyOnly extends Component {
+      override val value: None.type = None
+      override def normalize: KeyOnly
+    }
+
+    object KeyOnly {
+      def apply(name: String): KeyOnly = new KeyOnlyDecoded(name)
+      def unapply(wv: KeyOnly): Some[String] = Some(wv.key)
+    }
```
bplommer, line 180: As far as I can see from [RFC 3987](https://www.rfc-editor.org/rfc/rfc3987.html), dividing query components into key and value is just a common use case rather than something built into the query spec - so it isn't really accurate to think of a non-key-value component as having a key but no value. Would it be too much of a departure to change that terminology?

core/shared/src/main/scala/org/http4s/parser/QueryParser.scala
```diff
@@ -54,34 +55,34 @@ private[http4s] class QueryParser(
...
       input: CharBuffer,
-      acc: (String, Option[String]) => Builder[Query.KeyValue, Vector[Query.KeyValue]],
+      acc: Component => Builder[Component, Vector[Component]],
       flush: Boolean): Option[String] = {
     val valAcc = new StringBuilder(InitialBufferCapactiy)
 
     var error: String = null
     var key: String = null
     var state: State = KEY
+    var sep: Char = '&'
```
bplommer, line 65: Because this starts out as `&` until a separator is encountered, prepending a component to a `;`-separated query will result in insertion of an '&'. Similarly, appending a component will also use `&`. I wonder if we should have the `Query` itself store separator information, so that a parsed semicolon-separated list also uses semicolons for new components?

laws/src/main/scala/org/http4s/laws/discipline/ArbitraryInstances.scala
```diff
@@ -166,27 +166,27 @@ private[http4s] trait ArbitraryInstances extends Ip4sArbitraryInstances {
...
           for {
             k <- getArbitrary[String]
             v <- getArbitrary[Option[String]]
-          } yield (k, v)
+          } yield Query.Component(k, v)
         },
-        2 -> const(("foo" -> Some("bar"))) // Want some repeats
+        2 -> const(Query.Component("foo", "bar")) // Want some repeats
       )
     }
```
bplommer, line 180: This only generates decoded components. It should probably generate encoded components too.

PR #5201, opened 2021-09-15
Default execution context for Blaze client

bplommer:


bplommer:
Test failure seems to be an unrelated flake.

blaze-client/src/main/scala/org/http4s/blaze/client/BlazeClientBuilder.scala
```diff
@@ -74,7 +74,7 @@ sealed abstract class BlazeClientBuilder[F[_]] private (
     val chunkBufferMaxSize: Int,
     val parserMode: ParserMode,
     val bufferSize: Int,
-    val executionContext: ExecutionContext,
+    val executionContextConfig: ExecutionContextConfig,
```
rossabaker, line 77: The param in the scaladoc no longer lines up.

I'm surprised this passed MiMa without a synthetic `def executionContext`.

rossabaker (changes requested):
A couple tweaks, and I'm probably overlooking something on the MiMa, but this will make a lot of people happy.

blaze-client/src/main/scala/org/http4s/blaze/client/BlazeClientBuilder.scala
rossabaker: Lost my other comment: since this type is `private[http4s]`, I don't think it makes sense to expose it as a val.  Just a parameter.

bplommer: > I'm surprised this passed MiMa without a synthetic `def executionContext`.

Weird, that doesn't seem right 🤔

bplommer:
Yikes, it looks like mima isn't checking anything:
```
[info] rootJVM: mimaPreviousArtifacts is empty, not analyzing binary compatibility.
[info] core: mimaPreviousArtifacts is empty, not analyzing binary compatibility.
[info] boopickle: mimaPreviousArtifacts is empty, not analyzing binary compatibility.
[info] laws: mimaPreviousArtifacts is empty, not analyzing binary compatibility.
[info] ember-core: mimaPreviousArtifacts is empty, not analyzing binary compatibility.
[info] client: mimaPreviousArtifacts is empty, not analyzing binary compatibility.
[info] dsl: mimaPreviousArtifacts is empty, not analyzing binary compatibility.
[info] jawn: mimaPreviousArtifacts is empty, not analyzing binary compatibility.
[info] scala-xml: mimaPreviousArtifacts is empty, not analyzing binary compatibility.
[info] twirl: mimaPreviousArtifacts is empty, not analyzing binary compatibility.
[info] tests: mimaPreviousArtifacts is empty, not analyzing binary compatibility.
[info] blaze-core: mimaPreviousArtifacts is empty, not analyzing binary compatibility.
[info] scalatags: mimaPreviousArtifacts is empty, not analyzing binary compatibility.
[info] server: mimaPreviousArtifacts is empty, not analyzing binary compatibility.
[info] prometheus-metrics: mimaPreviousArtifacts is empty, not analyzing binary compatibility.
[info] dropwizard-metrics: mimaPreviousArtifacts is empty, not analyzing binary compatibility.
[info] ember-client: mimaPreviousArtifacts is empty, not analyzing binary compatibility.
[info] play-json: mimaPreviousArtifacts is empty, not analyzing binary compatibility.
[info] testing: mimaPreviousArtifacts is empty, not analyzing binary compatibility.
[info] async-http-client: mimaPreviousArtifacts is empty, not analyzing binary compatibility.
[info] circe: mimaPreviousArtifacts is empty, not analyzing binary compatibility.
[info] jetty-client: mimaPreviousArtifacts is empty, not analyzing binary compatibility.
[info] okhttp-client: mimaPreviousArtifacts is empty, not analyzing binary compatibility.
[info] blaze-client: mimaPreviousArtifacts is empty, not analyzing binary compatibility.
[info] blaze-server: mimaPreviousArtifacts is empty, not analyzing binary compatibility.
[info] servlet: mimaPreviousArtifacts is empty, not analyzing binary compatibility.
[info] server-testing: mimaPreviousArtifacts is empty, not analyzing binary compatibility.
[info] bench: mimaPreviousArtifacts is empty, not analyzing binary compatibility.
```

bplommer:
I've updated the build (in a hacky way) to get mima checking again - it's caught some other things too.

bplommer:
I've had a go at doing this in a binary-compatible way by introducing `BlazeClientBuilder.V2` - but I'm not thrilled about how it doesn't extend `BlazeClientBuilder`. On the other hand, this will result in a lot of misconfigured clients getting fixed. What do you think - should we just leave this out from 0.23 and do it on main?

blaze-client/src/main/scala/org/http4s/blaze/client/BlazeClientBuilder.scala
bplommer: Found the problem: https://github.com/http4s/http4s/pull/5215

rossabaker:
Is the V2 necessary because of the `executionContext`?  On the original ticket, we discussed just deprecating it and returning `IORuntime.global.compute` if the config is the default.  That's nasty, because if we reference it, someone is responsible for shutting it down.  But if it's deprecated, it's a fair way out.  I don't imagine many people are using it as a val anyway.

bplommer:
> Is the V2 necessary because of the `executionContext`?  

Yep

>On the original ticket, we discussed just deprecating it and returning `IORuntime.global.compute` if the config is the default.  That's nasty, because if we reference it, someone is responsible for shutting it down.  But if it's deprecated, it's a fair way out.  I don't imagine many people are using it as a val anyway.

Sounds good to me - I'll try and wrap this up later.

blaze-client/src/main/scala/org/http4s/blaze/client/BlazeClientBuilder.scala
```diff
@@ -126,13 +126,19 @@ sealed abstract class BlazeClientBuilder[F[_]] private (
...
       scheduler = scheduler,
       asynchronousChannelGroup = asynchronousChannelGroup,
       channelOptions = channelOptions,
       customDnsResolver = customDnsResolver
     ) {}
 
+  @deprecated(
+    "Do not use - always returns cats.effect.unsafe.IORuntime.global.compute." +
+      "There is no direct replacement - directly use Async[F].executionContext or your custom execution context",
+    "0.23.4")
```
rossabaker, line 139:
```suggestion
    "0.23.5")
```

blaze-client/src/main/scala/org/http4s/blaze/client/BlazeClientBuilder.scala
```diff
@@ -348,13 +353,20 @@ object BlazeClientBuilder {
...
       scheduler = tickWheelResource,
       asynchronousChannelGroup = None,
       channelOptions = ChannelOptions(Vector.empty),
       customDnsResolver = None
     ) {}
 
+  @deprecated(
+    "Most users should use the default execution context provided. " +
+      "If you have a specific reason to use a custom one, use `.withExecutionContext`",
+    "0.23.4")
```
rossabaker, line 366:
```suggestion
    "0.23.5")
```

rossabaker (approved):

rossabaker:
Drat. I overlooked the deprecations when merging.

bplommer:
It also just occurred to me, I don't think there's any reason why `executionContext` can't return the actual custom execution context when there is one? Then we're not changing the behaviour of any existing code.

rossabaker:
Oh, yes.  I envisioned that when I discussed this with @kubukoz on the original ticket, and didn't catch that we're not doing that now.  :+1:

PR #5204, opened 2021-09-17
Update to Bootstrap 5

### LINKED ISSUE #5191: Update Bootstrap to 5.1.1
I find the current http4s website very cluttered, but much improved with some minor tweaks to the styling to add more white space between, e.g. above a heading to more clearly distinguish section breaks.

I notice the site currently uses bootstrap + some custom styles. A way forward might be:

- Update to current bootstrap (5.1.1 from current 4.0 beta)
- Changing some of the custom styling to add more white space. (Tbh bootstrap's defaults are pretty good, so probably deleting some of the customisation would be sufficient.)

This should be fairly self contained.

Alternatively the styling could be redeveloped using Tailwind, which I've worked with recently. Tailwind is nicer for more complex sites but I think it's overkill for http4s' needs.

noelwelsh:
- Replace Bootstrap 4 with Bootstrap 5

- Change formatting to approximate existing site. I'm not aiming to 100%
matching the existing styling as I think we'll change aspect of it
anyway. Just get something close enough without too much work.

- Close #5191


rossabaker:
* `partials/head.html` is including the bootstrap that you upgraded, and `partials/scripts.html` is including one from bootstrapcdn.com, and I'm surprised we have two.  This is a pre-existing condition, but I don't know whether it foils any of your plans.

* Something happened to the icons in the top right:
Screen Shot 2021-09-17 at 3 22 29 PM

rossabaker (approved):
Ope, those icons are broken locally on main, too.  I think that's unrelated to this.

PR #5206, opened 2021-09-17
Replace Sync and Concurrent constraints with fs2.Compiler.Target

bplommer:
There's a lot of places that need `Target` but have a context bound of `Concurrent` or `Sync` - either of which derives `Target` but neither of which implies the other. That makes the API less flexible, and client code ends up needing both and using an `Async` bound. The `Concurrent` bound also prevents use of datatypes such as `SyncIO` and `StateT` that have `Sync` but not `Concurrent` instances.

I have a more comprehensive version of this PR stashed away, but this is enough to convey the idea.

### DIFF
--- a/core/shared/src/main/scala/org/http4s/EntityDecoder.scala
+++ b/core/shared/src/main/scala/org/http4s/EntityDecoder.scala
@@ -17,7 +17,7 @@
 package org.http4s
 
 import cats.{Applicative, Functor, Monad, SemigroupK}
-import cats.effect.Concurrent
+import fs2.Compiler.Target
 import cats.syntax.all._
 import fs2._
 import scala.annotation.implicitNotFound
@@ -187,45 +187,45 @@ object EntityDecoder extends EntityDecoderCompanionPlatform {
     }
 
   /** Helper method which simply gathers the body into a single Chunk */
-  def collectBinary[F[_]: Concurrent](m: Media[F]): DecodeResult[F, Chunk[Byte]] =
+  def collectBinary[F[_]: Target](m: Media[F]): DecodeResult[F, Chunk[Byte]] =
     DecodeResult.success(m.body.chunks.compile.toVector.map(bytes => Chunk.concat(bytes)))
 
   /** Decodes a message to a String */
   def decodeText[F[_]](
-      m: Media[F])(implicit F: Concurrent[F], defaultCharset: Charset = DefaultCharset): F[String] =
+      m: Media[F])(implicit F: Target[F], defaultCharset: Charset = DefaultCharset): F[String] =
     m.bodyText.compile.string
 
   /////////////////// Instances //////////////////////////////////////////////
 
   /** Provides a mechanism to fail decoding */
-  def error[F[_], T](t: Throwable)(implicit F: Concurrent[F]): EntityDecoder[F, T] =
+  def error[F[_], T](t: Throwable)(implicit F: Target[F]): EntityDecoder[F, T] =
     new EntityDecoder[F, T] {
       override def decode(m: Media[F], strict: Boolean): DecodeResult[F, T] =
         DecodeResult(m.body.compile.drain *> F.raiseError(t))
       override def consumes: Set[MediaRange] = Set.empty
     }
 
-  implicit def binary[F[_]: Concurrent]: EntityDecoder[F, Chunk[Byte]] =
+  implicit def binary[F[_]: Target]: EntityDecoder[F, Chunk[Byte]] =
     EntityDecoder.decodeBy(MediaRange.`*/*`)(collectBinary[F])
 
-  implicit def byteArrayDecoder[F[_]: Concurrent]: EntityDecoder[F, Array[Byte]] =
+  implicit def byteArrayDecoder[F[_]: Target]: EntityDecoder[F, Array[Byte]] =
     binary.map(_.toArray)
 
   implicit def text[F[_]](implicit
-      F: Concurrent[F],
+      F: Target[F],
       defaultCharset: Charset = DefaultCharset): EntityDecoder[F, String] =
     EntityDecoder.decodeBy(MediaRange.`text/*`)(msg =>
       collectBinary(msg).map(chunk =>
         new String(chunk.toArray, msg.charset.getOrElse(defaultCharset).nioCharset)))
 
-  implicit def charArrayDecoder[F[_]: Concurrent]: EntityDecoder[F, Array[Char]] =
+  implicit def charArrayDecoder[F[_]: Target]: EntityDecoder[F, Array[Char]] =
     text.map(_.toArray)
 
-  implicit def multipart[F[_]: Concurrent]: EntityDecoder[F, Multipart[F]] =
+  implicit def multipart[F[_]: Target]: EntityDecoder[F, Multipart[F]] =
     MultipartDecoder.decoder
 
   /** An entity decoder that ignores the content and returns unit. */
-  implicit def void[F[_]: Concurrent]: EntityDecoder[F, Unit] =
+  implicit def void[F[_]: Target]: EntityDecoder[F, Unit] =
     EntityDecoder.decodeBy(MediaRange.`*/*`) { msg =>
       DecodeResult.success(msg.body.drain.compile.drain)
     }
--- a/core/shared/src/main/scala/org/http4s/UrlForm.scala
+++ b/core/shared/src/main/scala/org/http4s/UrlForm.scala
@@ -18,7 +18,7 @@ package org.http4s
 
 import cats.{Eq, Monoid}
 import cats.data.Chain
-import cats.effect.Concurrent
+import fs2.Compiler.Target
 import cats.syntax.all._
 import org.http4s.headers._
 import org.http4s.internal.CollectionCompat
@@ -105,7 +105,7 @@ object UrlForm {
       .withContentType(`Content-Type`(MediaType.application.`x-www-form-urlencoded`, charset))
 
   implicit def entityDecoder[F[_]](implicit
-      F: Concurrent[F],
+      F: Target[F],
       defaultCharset: Charset = DefaultCharset): EntityDecoder[F, UrlForm] =
     EntityDecoder.decodeBy(MediaType.application.`x-www-form-urlencoded`) { m =>
       DecodeResult(
--- a/core/shared/src/main/scala/org/http4s/multipart/MultipartDecoder.scala
+++ b/core/shared/src/main/scala/org/http4s/multipart/MultipartDecoder.scala
@@ -17,15 +17,15 @@
 package org.http4s
 package multipart
 
-import cats.effect.Concurrent
+import fs2.Compiler.Target
 import cats.syntax.all._
 import fs2.Pipe
 
 private[http4s] object MultipartDecoder extends MultipartDecoderPlatform {
-  def decoder[F[_]: Concurrent]: EntityDecoder[F, Multipart[F]] =
+  def decoder[F[_]: Target]: EntityDecoder[F, Multipart[F]] =
     makeDecoder(MultipartParser.parseToPartsStream[F](_))
 
-  private[multipart] def makeDecoder[F[_]: Concurrent](
+  private[multipart] def makeDecoder[F[_]: Target](
       impl: Boundary => Pipe[F, Byte, Part[F]]
   ): EntityDecoder[F, Multipart[F]] =
     EntityDecoder.decodeBy(MediaRange.`multipart/*`) { msg =>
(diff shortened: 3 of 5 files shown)


rossabaker:
I've tried to do this a couple times and was told by the FS2 team that `Compiler.Target` was not meant as a constraint in downstream libraries.  Their argument made sense to me at the time, but I've forgotten why not.  It was probably @SystemFw or @mpilquist who made it.

SystemFw:
I think we both made that argument, and personally I stand by it: I don't think it's a good idea to litter everyone's codebase with something as ad hoc and obscure as Compiler.Target, making UX and on boarding harder.
I think most client code will be fine just using Concurrent, and upgrading to Async when needed it's the cats-effect-idiomatic measure to take when that's not enough.
I don't think supporting StateT or SyncIO is a strong argument to the contrary, in the context of http4s (you realistically need a concurrent bound to use it anyway)

bplommer:
> I think we both made that argument, and personally I stand by it: I don't think it's a good idea to litter everyone's codebase with something as ad hoc and obscure as Compiler.Target, making UX worse and onboarding harder.

Is `Compiler.Target` really any worse from a UX or onboarding perspective than `MonadCancelThrow` or `Concurrent`? Re `Target` being ad-hoc, I not sure that's actually a downside here - `Target` is the most direct representation of the ability to compile a `Stream`, and it's far more straightforward to explain why creating an `EntityDecoder` requires being able to compile a `Stream` than why it requires being able to race fibers. Re it being obscure - that's mainly because libraries don't expose it in their APIs. If http4s does, it won't be obscure for long. I do get that there's a cognitive overhead in learning a new type class, but people who don't want to have to deal with type classes are already switching to ZIO anyway.

> I don't think supporting StateT or SyncIO is a strong argument to the contrary, in the context of http4s 

I've had to work with tests that were written against a big monad transformer stack involving `StateT`, and being able to remove `Concurrent` bounds from the code under test would have helped quite a lot. That big monad transformer stack is on my hit-list, but this kind of limitation makes it harder rather than easier to refactor it away.

> (you realistically need a concurrent bound to use it anyway)

I'm not convinced that will be the case if we replace all of the `Concurrent` bounds that we can with `Target`.

SystemFw:
> Is Compiler.Target really any worse from a UX or onboarding perspective than MonadCancelThrow or Concurrent?

Yes, on both sides, for various reasons:
- It's defined in fs2, not cats-effect. A lot of teams use the latter before getting to the former (if they ever get there)
- it's defined with a type level or-else, and I can definitely see a lot of `: Concurrent: Compiler.Target` everywhere since there is no automatic way to simplify that bound.
- All its operations are private[fs2], and a lot are scary looking and not easily relatable to user land concepts, especially `compile`

> I not sure that's actually a downside here - Target is the most direct representation of the ability to compile a Stream, and it's far more straightforward to explain why creating an EntityDecoder requires being able to compile a Stream than why it requires being able to race fibers. 

`Concurrent` can be described as "cats-effect's main control flow typeclass", and in any case any tagless user of cats-effect will know about it since it's pervasive across the ecosystem. Conversely, you can use http4s without really knowing what fs2 _is_ right now.

> Re it being obscure - that's mainly because libraries don't expose it in their APIs. 

It's obscure because it's a complex implementation detail of fs2, and its machinery is way more complex than most other typeclasses (Compiler[F, F] --> Target --> Sync OrElse Concurrent). https://github.com/typelevel/fs2/blob/main/core/shared/src/main/scala/fs2/Compiler.scala this is one of the worst files you want to land on a go-to-definition while looking to encode something. It's also essentially lawless.

> I do get that there's a cognitive overhead in learning a new type class, but people who don't want to have to deal with type classes are already switching to ZIO anyway.

I don't buy this argument (apart from the fact that you can work with `IO` without typeclasses practically in CE3) because it seems like a false dichotomy: either no typeclasses or an arbitrary number of them. On the contrary, we did try to avoid type class explosion in the cats-effect hierarchy. Also, since you will have to learn about `Concurrent` even if the the encoders in http4s didn't use it, http4s is not adding a requirement to learn about a new typeclass, which it would if it were to use Compiler.Target

> and being able to remove Concurrent bounds from the code under test

I think one of the tradeoffs of CE3's design is that it's very unrealistic that this will happen. We decided to remove reliance on `Sync/Async` everywhere, and I think we've actually succeeded (you can use those bounds in `main` and tagless implementations only), but this comes at the cost of requiring `Concurrent/Temporal` for a lot of things: if you use fs2 for control flow, you will have it, if there is some state, you will have it, not to mention the legitimate pervasiveness of the concerns it actually encodes like `timeout`s

bplommer:
Okay that makes sense - thank you for helping me so patiently through my latest bout of Dunning-Krueger 😅

I guess that also means we should replace remaining `Sync` bounds with `Concurrent` where possible.

I do still think it's worth having a capability trait in the spirit of [JsonDecoder](https://github.com/http4s/http4s/blob/main/circe/src/main/scala/org/http4s/circe/JsonDecoder.scala) for compiling entity bodies, and having instances available for both `Sync` and `Concurrent`. (If I'm wrong about this and it's better to directly require `Concurrent`, that presumably means `JsonDecoder` should be deprecated/removed too now that decoding json doesn't require `Sync`?)

rossabaker:
Currently, every `EntityEncoder` is defined in terms of `Concurrent`.  Since `Compiler.Target` is lawless, is it better to think of that as a capability trait?  If we introduce a capability for decompiling bodies as Ben proposes, how is it different from `Compiler.Target`?

bplommer:
I suppose the difference is that although the *purpose* of `Compiler.Target` is to have the functionality to able to compile a Stream, that's not how it's defined - rather, it's defined in terms of the (quite complex) functionality that turns out to be needed for that purpose.

The most general form of a capability trait would be something like the following and be straightforward to implement in terms of `Target` - though it looks like something that would belong in fs2 rather than http4s:
```scala
trait ToCompileOps[F[_]] {
  def compile[A](sfa: Stream[F, A]): Stream.CompileOps[F, F, A]
}
```

I think an http4s-specific form would essentially just be a weaker  form of this that would preserve the ability to use http4s without knowing about fs2. (I was going to say "abstract away the dependency on fs2 to keep it as an implementation detail", but that wouldn't be accurate given that `EntityBody` is just an alias for `Stream`.)

rossabaker:
There is a 1.0 proposal to make `Entity` a sum type of either strict or streamed bodies.  I'm not sure if that helps anything in this discussion, because the operation still needs to return a `CompileOps`, unless we create a passthrough capability for ops.  And even that doesn't help much, because several of those ops refer to fs2 types.  It would be a leaky capability.

SystemFw:
I would also like to point out, just for completeness, that the actual abstraction that defines the ability to compile a Stream is not `Compiler.Target`, but `Compiler[F, G]`, since the full problem requires being able to compile to `Id`, to `Either[Throwable, *]` and to `Resource[F, *]`. That's why I say `Target` specifically is an implementation detail of a more complex bit of machinery.

bplommer:
Fair point - would you see it as making a significant difference to use `Compiler[F, F]` (under a context-bound-friendly alias) rather than `Compiler.Target`?

bplommer:
> It would be a leaky capability.

Isn't that just an inescapable consequence of the fact that the http4s API is defined partly in terms of fs2 types?

rossabaker:
> > It would be a leaky capability.

> Isn't that just an inescapable consequence of the fact that the http4s API is defined partly in terms of fs2 types?

Yes  I tried to abstract over it for first-class Monix support years ago, but everything grows unsightly parameters and the abstraction is a dumbed down version that lets neither fs2 nor monix shine.  I don't think leaking fs2's API is bad: I think participating in its ecosystem is a strength.

I just think this compatibility trait would be basically equivalent to `Compiler[F, F]`.  And most decoders are going to also need a `MonadThrow[F]`, which is ... `Compile.Target`.  I suppose a few could work with `Compiler[Id, Id]`, but I bet it's not many.  `ByteVector`, `Unit`, and other total functions from `Array[Byte]`, probably.

bplommer:
> I don't think leaking fs2's API is bad: I think participating in its ecosystem is a strength.

I agree. But the current situation of exposing fs2's API while keeping it aliased and obscure enough to allow library users not to know about it feels like an awkward compromise. It just seems to perpetuate the fs2 being seen as scary and mystifying. 

(There's definitely more that could be done to make fs2 approachable. `Stream.scala` being several thousand lines long is genuinely intimidating - I've experimented with pulling out implementation details into separate files and it seems feasible but the PR would be a nightmare to review.)

(Anecdotally, confusion about naming probably doesn't help either, judging by the number of people I've heard refer to fs2-Kafka as "fs2". In an ideal world fs2 and fs2-kafka would probably be cats-stream and Kafka4cats, but renaming them would be a huge pain.)
 
> I just think this compatibility trait would be basically equivalent to `Compiler[F, F]`.  

So what's the upshot of this - decoders should either continue to use `Concurrent` or switch to using `Compiler[F, F]`, maybe aliased as something like `EntityCompiler[F]`? Both options seem to have advantages.

> And most decoders are going to also need a `MonadThrow[F]`, which is ... `Compile.Target`.  

I totally buy Fabio's argument about why `Compiler.Target` shouldn't be used - in fact it seems like ideally it would be package-private. `MonadCancel: EntityCompiler` doesn't seem like a horrible context bound to have. Alternatively `EntityCompiler` could be a thin wrapper that provides `Compiler` and `MonadThrow`/`MonadCancelThrow` - that way it can also have its own `@implicitNotFound` annotation.

bplommer:
> There is a 1.0 proposal to make `Entity` a sum type of either strict or streamed bodies.

Is there an issue for this by the way? If `Message` types were covariant in `F` then wouldn't `Message[Pure]` accomplish the same thing?

armanbilge:
@bplommer there's this PR: https://github.com/http4s/http4s/pull/4509

> If Message types were covariant in F then wouldn't Message[Pure] accomplish the same thing?

Kind of? You can be pure and still be a stream. Plus I think some of the optimizations are achieved by avoiding the streaming machinery altogether.

rossabaker:
The backend has to pay to compile and render the stream, and even if it starts as a `Message[Pure]` in the app, it will need to be covaried to a `Message[IO]` to be returned to the backend.  There are a lot of performance improvements we're leaving on the table making everything a stream, though we'll need to take stock of the complexity of having multiple kinds of bodies.  (For reference, akka-http does.)

rossabaker:
$WORK is currently suffering with the fact that the `Sync` compiler isn't as capable as the `Concurrent` compiler (sure) and fails at runtime (eek!).  Two flavors of https://github.com/typelevel/fs2/issues/2371 bit us early in our CE3 migration.  It's another thing that makes me wary of better supporting `Sync`.

bplommer:
> and fails at runtime (eek!)

Yikes. I didn't realise that - I feel like there should be a scarily named import required in order to bring the Sync version into scope. Failing that, I agree on not supporting Sync - and for the same reason, it sounds like existing Sync bounds that can't be changed to Concurrent but are also used to compile streams should be promoted to Async.

mpilquist:
> I feel like there should be a scarily named import required in order to bring the Sync version into scope.

Note the sync instance is used to compile pure streams & fallible streams as well as SyncIO streams. Also note that this isn't new with 3.x & `Compiler.Target`. In all releases since 0.9, compiling a stream has required a `Sync` instance.

mpilquist:
Sort of -- we used to capture the `Concurrent` instance used on `merge` or similar combinators and then throw it away if calling `translate`: https://github.com/typelevel/fs2/blob/cd0f23a0e3754739a6fa4f0a92ead1f6786aede0/core/shared/src/main/scala/fs2/Stream.scala#L2666-L2679

rossabaker:
Most of the operations that can trigger this require a `Concurrent`.  (I think I found two `interruptWhen` overloads that don't?)  The pattern is:

1. Build the `Stream` using `Concurrent`.
2a. Pass to something constrained to `Sync` to compile.
2b. Translate to something that only has `Sync` (e.g., Doobie's `ConnectionIO`).

Our 2a would have better been `Concurrent`, but the old code took a `Sync` and it compiled, so it wasn't called to our attention until we died at runtime.  And I don't know what we can do about 2b other than to be more careful what we pass to Doobie's large object streaming.

If we promote a `Compiler[F, F]` or `Compiler.Target[F]` over `Concurrent`, we'll be more flexible, but more likely to pass this footgun downstream.

bplommer:
> 2a. Pass to something constrained to Sync to compile.

Could the `Sync`-based `Target` constructor do a runtime check on the `Sync` instance to see if it's also a `Concurrent` instance, and use the `Concurrent` compiler if so? It wouldn't fix the issue in all cases but it would in some.

rossabaker:
That could salvage some runtime errors, but I think would introduce a whole new subtlety.  We at least understand what not to do now.

I briefly looked last night at what could be done about this.  I think the technique linked by @mpilquist above might fix 2a but not 2b, but I was sleepy and I have a lot of catching up to do on the internals.

bplommer:
> Regardless, as far as Http4S is concerned, is this Pull Request still proposed?

No - I was keeping it open as a reminder that we should prevent use of the `Sync` compiler, but I think our best answer for that may be Scalafix (#5497)

PR #5208, opened 2021-09-17
Right align the social icons in the navbar again

slice:
Bootstrap 5 renamed the `mr-auto` class to `me-auto`. This class applies the `margin-right: auto` rule which has the effect of pushing the remaining elements to the right. Since we're on Bootstrap 5 now, use the new class.

| Current | Proposed |
| - | - |
| image | image

### DIFF
--- a/website/jvm/src/hugo/themes/http4s.org/layouts/_default/single.html
+++ b/website/jvm/src/hugo/themes/http4s.org/layouts/_default/single.html
@@ -12,7 +12,7 @@
       </button>
 
       <div class="collapse navbar-collapse" id="navbarsExampleDefault">
-        <ul class="navbar-nav mr-auto">
+        <ul class="navbar-nav me-auto">
         <li class="nav-item dropdown">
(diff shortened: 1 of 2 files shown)


slice:
BTW... I changed the usage in `partials/nav.html` as well even though from what I can tell that file goes completely unused. `nav.html` is being included in `layouts/single.html` but that doesn't seem to be used, either. `_default/single.html` seems to be used instead.

rossabaker:
Thanks!

My icons are not rendering locally for some reason, but that was true before.  I can reproduce the move in placement by testing this.

slice:
@rossabaker That might have to do with the fact that the icons are implemented through a custom webfont. I believe that's not only bad for accessibility reasons, but things can probably go haywire if you have certain fonts installed locally. Maybe we should move to SVGs soon?

rossabaker:
My frontend skills ossified in the early 2000s, so I'm going to be highly deferential to people who tell me there's a better way!

PR #5209, opened 2021-09-17
Fix cards on landing page

slice:
Put `.card` and `.card-body` elements into each column so that we get a bordered look.

| Current | Proposed |
| - | - |
| image | image


rossabaker:
Nice, thank you!

PR #5212, opened 2021-09-18
Fix edit button links and positioning

slice:
There are some floating edit and navigation offcanvas buttons that have stopped floating due to Bootstrap 5 renaming some classes, so I've fixed that. I also rolled in a change that fixes the website edit links. The docs edit links are fine, though.

I also considered updating [this file](https://github.com/http4s/http4s/blob/66e322e1650f424ddd5ffd49ca211b6d9540ba8b/website/jvm/src/hugo/layouts/partials/edit-link.html) but it seems to be unused from what I can tell. Maybe ripping out all of this unused stuff can be its own issue? Or maybe it's not actually unused? (Related: https://github.com/http4s/http4s/pull/5208#issuecomment-922099474)

| Current | Proposed |
| - | - |
| image | image


rossabaker:
It's been a long time since I hacked on this, but I think the docs projects might use the theme from website.  It's possible that things that appear unused come from there.  If that's true, that's bad design and its own issue.

This looks good.  I noticed the button had drifted left.  I didn't notice the link was broken.

PR #5215, opened 2021-09-18
fix build version

bplommer:
This seems to have been preventing mima from finding any previous versions. ~There'll need to be fixes or exclusions before this can actually be merged.~ There were a couple of mima errors but they seem to be safe to exclude.

I also ran scalafmt by mistake 😐


rossabaker:
Oh, good find.

That scalafmt is going to cause problems unless we start it on an earlier branch.  I'm going to separate the meaty bits from the format before merging.

PR #5216, opened 2021-09-18
Improve Ember Parser Performance

ChristopherDavenport:
Looks to improve the expected case performance. 

Most servers will encode their prelude and headers within a single buffer. Presently we do a O(n) lookup for a doublecrlf to separate the header array from the body array. Then we do an O(n) parse over the prelude and body. 

Instead we give ourselves where we will repeat parsing over the buffers O(n) and iterate to do so again. If servers have odd chunking behavior there will be more reused work, but it the default case this will be much more performant. 

To make sure I was being scientific in this approach I created a bench suite to test our parsing timings only. 

```
[info] EmberParserBench.parseRequest     avgt   10  1658.321 ± 74.716  ns/op
[info] EmberParserBench.parseResponse    avgt   10  1080.508 ± 96.766  ns/op

[info] EmberParserBench.parseRequestNew  avgt   10  1339.713 ± 147.635  ns/op
[info] EmberParserBench.parseResponseNew avgt   10   863.105 ±  28.273  ns/op
```

All tests pass and we stop processing off as soon as we reach the maxHeadersSize.

### DIFF
--- a/bench/src/main/scala/org/http4s/ember/bench/EmberParserBench.scala
+++ b/bench/src/main/scala/org/http4s/ember/bench/EmberParserBench.scala
@@ -69,8 +69,8 @@ object EmberParserBench {
     def setup(): Unit = {
       req = Request[IO]().withEntity("Hello Bench!")
       resp = Response[IO]().withEntity("Hello Bench!")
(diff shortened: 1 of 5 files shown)


RaasAhsan (approved):

rossabaker:
Looks like it needs a MiMa exclusion.

rossabaker (approved):
:+1: on green

Issue #5218, opened 2021-09-18
UnixSocket Client/Server


ChristopherDavenport:
Ember's Parser/Encoder logic on top of the Unix Socket Implementation introduced in fs2.

I wanted to make sure we had an issue since this got mentioned recently, and its a fairly straight encoding.

armanbilge:
Without taking a look, does this actually need anything special? IIRC fs2's unix sockets also extends `SocketGroup` so can't you just pass it directly to the ember builders?

ChristopherDavenport:
I didn't realize that. I'll try that out first.

armanbilge:
Woops, I think I put my foot in my mouth. The unix sockets are not a socket group, but an instance of a unix socket is just a socket.

ChristopherDavenport:
https://github.com/typelevel/fs2/blob/main/io/shared/src/main/scala/fs2/io/net/unixsocket/UnixSockets.scala#L33 - Yeah, I am going to PR it. Its fairly straightforward, but needs a different builder.

Issue #5128, opened 2021-08-30
EmberWebSocketSuite: Unknown opcode 3


rossabaker:
https://github.com/http4s/http4s/pull/5124/checks?check_run_id=3456632732#step:13:3978

```
==> X org.http4s.ember.server.EmberServerWebSocketSuite.respond to pings  0.055s org.java_websocket.exceptions.InvalidFrameException: Unknown opcode 3
    at org.java_websocket.drafts.Draft_6455.toOpcode(Draft_6455.java:865)
    at org.java_websocket.drafts.Draft_6455.translateSingleFrame(Draft_6455.java:533)
    at org.java_websocket.drafts.Draft_6455.translateFrame(Draft_6455.java:739)
    at org.java_websocket.WebSocketImpl.decodeFrames(WebSocketImpl.java:398)
    at org.java_websocket.WebSocketImpl.decode(WebSocketImpl.java:233)
    at org.java_websocket.client.WebSocketClient.run(WebSocketClient.java:516)
    at java.lang.Thread.run(Thread.java:831)
    at apply @ org.http4s.ember.server.EmberServerBuilder.build$$anonfun$2(EmberServerBuilder.scala:136)
    at apply @ org.http4s.ember.server.EmberServerBuilder.build$$anonfun$2(EmberServerBuilder.scala:136)
    at complete @ org.http4s.ember.server.internal.Shutdown$$anon$1.<init>(Shutdown.scala:55)
    at flatMap @ org.http4s.ember.server.EmberServerWebSocketSuitePlatform$Client.close$$anonfun$2(EmberServerWebSocketSuitePlatform.scala:48)
    at apply @ org.http4s.ember.server.EmberServerBuilder.build$$anonfun$2(EmberServerBuilder.scala:136)
    at >> @ org.http4s.ember.server.EmberServerWebSocketSuitePlatform$Client.connect(EmberServerWebSocketSuitePlatform.scala:46)
    at map @ org.http4s.ember.server.EmberServerWebSocketSuite.$init$$$anonfun$16$$anonfun$3$$anonfun$3$$anonfun$3$$anonfun$3(EmberServerWebSocketSuite.scala:92)
    at apply @ org.http4s.ember.server.EmberServerBuilder.build$$anonfun$2(EmberServerBuilder.scala:136)
```

armanbilge:
An observation: this error so far only seems to occur on Java 16.

rossabaker:
I wonder if also on 17?

[...]

PR #5219, opened 2021-09-19
Add Ember Unix Socket Implementations

ChristopherDavenport:
New Features:

- Adds UnixSocket Support to both ember-client and ember-server
- On Client Uses a new Attribute Key to seemlessly switch - Other backends can use the same type to add their support in the future
- Adds Client Middleware to make all client calls redirect to the UnixSocket
- On Server adds a new configuration that when present chooses to bind to a UnixSocket rather than a net address

Concerns:

- The server interface is a bit odd actually as its reliance on InetSocketAddress makes it limited in regards to UnixSockets. We fake it unresolved.
- Testing this is a bit tricky. I intend to publishLocal and stand up a project to confirm. Ideas on easier testability here would be appreciated.


ChristopherDavenport:
Fixes #5218

ember-client/src/main/scala/org/http4s/ember/client/EmberClientBuilder.scala
```diff
@@ -186,6 +188,31 @@ final class EmberClientBuilder[F[_]: Async] private (
             }
           }
         } yield responseResource._1
+
+      def unixSocketClient(
+          request: Request[F],
+          address: UnixSocketAddress): Resource[F, Response[F]] =
+        Resource
+          .eval(EmberConnection(ClientHelpers.unixSocket(request, address, tlsContextOpt)))
```
armanbilge, line 196: More of a question, there's no advantage to using a keypool here?

ChristopherDavenport:
https://github.com/davenverse/whale-tail/pull/158

Specifically https://github.com/davenverse/whale-tail/blob/39ddb8746c7a26ba139ae4069cc1895f4ff4092f/examples/src/main/scala/StaticLocationExample.scala - Shows an example of running against a socket locally.

ember-client/src/main/scala/org/http4s/ember/client/EmberClientBuilder.scala
ChristopherDavenport: This one I'm uncertain about. You can get all the same advantages but comes with another pool of state to drag around, and I doubt this will often even be used. I think it can be added later the same as the other one if we decide we want pooling.

armanbilge: Am I crazy to wonder if there's some way to re-use the current keypool? That's what I was initially imagining.

armanbilge:
I think it would still be nice to have a test here only if it's running against itself. Note that in main where we use an external server for testing clients, we could definitely use that for a unix socket test as well. In theory this technique could be backported.

https://github.com/http4s/http4s/blob/main/scripts/scaffold_server.js

ember-server/src/main/scala/org/http4s/ember/server/internal/ServerHelpers.scala
```diff
@@ -69,7 +71,89 @@ private[server] object ServerHelpers {
...
+      maxHeaderSize: Int,
+      requestHeaderReceiveTimeout: Duration,
+      idleTimeout: Duration,
+      logger: Logger[F]
+  ) = {
+    val server =
+      // Our interface has an issue
+      Stream
+        .eval(
+          ready.complete( // Ready Complete Does Nothing - We don't have a signal
```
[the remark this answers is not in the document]
ChristopherDavenport: We aren't actually ready. There isn't any signal when the server is ready in the interface from fs2.

ember-client/src/main/scala/org/http4s/ember/client/EmberClientBuilder.scala
ChristopherDavenport: No, thats a totally different resource.

armanbilge: The KeyPool maps `RequestKey` to `EmberConnection[F]`, right? And both the ordinary sockets and unix sockets use `EmberConnection[F]`, correct? So can't we just add the unix socket path or something to the `RequestKey` to make this work? Apologies if I'm missing something.

ember-server/src/main/scala/org/http4s/ember/server/internal/ServerHelpers.scala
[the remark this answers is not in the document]
armanbilge: Gotcha. Hmm.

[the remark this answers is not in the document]
armanbilge:
```suggestion
          ready.complete( // This is a lie, note there isn't any signal from fs2 when the server is ready
```

ChristopherDavenport:
@armanbilge Added an end to end test I took from whale-tail. Shows that both rendering, and parsing via the websocket works on both the client and server. 

It doesn't give us an external test. But that's whats reliably present on other systems without additional work.

ember-client/src/main/scala/org/http4s/ember/client/EmberClientBuilder.scala
ChristopherDavenport: But why? The state is fully decoupled. Its equivalent to a fully different pool.

armanbilge: > You can get all the same advantages but comes with another pool of state to drag around

This seemed like reason enough to me. Just a thought.

ChristopherDavenport: That was my reason to not use a pool. Not to complicate the other one which is already a significant headache to think through today.

If you are insistent I use the pool I will make a new one. I just don't see the need for a first effort.

ember-server/src/test/scala/org/http4s/ember/server/EmberUnixSocketSuite.scala
```diff
@@ -0,0 +1,54 @@
...
+import cats.syntax.all._
+import cats.effect._
+import org.http4s._
+import org.http4s.ember.client.EmberClientBuilder
+import fs2.io.net.unixsocket.{UnixSocketAddress, UnixSockets}
+import org.http4s._
+import org.http4s.client.middleware.UnixSocket
+import scala.concurrent.duration._
+
+class EmberUnixSocketSuite extends Http4sSuite {
```
armanbilge, line 30: What happens to this test on non-JDK16? Do we need to add JNR to test deps?

ChristopherDavenport: Good call.

ember-server/src/test/scala/org/http4s/ember/server/EmberUnixSocketSuite.scala
```diff
@@ -0,0 +1,54 @@
...
+
+  test("Should be able to connect to a server created") {
+    def app = HttpApp.liftF(Response[IO](Status.Ok).withEntity("Hello Unix Sockets!").pure[IO])
+    val test = for {
+      id <- Resource.eval(IO(java.util.UUID.randomUUID().toString()))
+      localSocket = UnixSocketAddress("/tmp/" ++ id ++ ".sock")
```
armanbilge, line 36: Can we use fs2 temporary files method for this?

ChristopherDavenport: Why? The server creates and deletes its own unix socket by default? The likelihood of a conflict is practically non-existent?

ember-client/src/main/scala/org/http4s/ember/client/EmberClientBuilder.scala
armanbilge: > Not to complicate the other one which is already a significant headache to think through today.

Nope, not insistent at all, you seemed on the fence about this so I asked. Furthermore, I didn't appreciate the weight of messing with the current pool, I agree better let that sleeping dog lie :)

ember-server/src/test/scala/org/http4s/ember/server/EmberUnixSocketSuite.scala
armanbilge: Fair point I guess, I don't really understand about these socket files. But I remember when researching temporary files for the fs2.io JS port tricks like this were strongly discouraged in favour of using OS-provided temporary files methods for security reasons. OTOH this is just a test in CI so probably nbd :)

```suggestion
      id <- Resource.eval(cats.effect.std.UUIDGen[IO].randomUUID).map(_.toString)
      localSocket = UnixSocketAddress("/tmp/" ++ id ++ ".sock")
```

ChristopherDavenport:
```
[error] Test suite org.http4s.ember.server.EmberUnixSocketSuite failed with java.lang.NoClassDefFoundError: java/net/UnixDomainSocketAddress.
[error] This may be due to the ClassLoaderLayeringStrategy (ScalaLibrary) used by your task.
[error] To improve performance and reduce memory, sbt attempts to cache the class loaders used to load the project dependencies.
[error] The project class files are loaded in a separate class loader that is created for each test run.
[error] The test class loader accesses the project dependency classes using the cached project dependency classloader.
[error] With this approach, class loading may fail under the following conditions:
[error] 
[error]  * Dependencies use reflection to access classes in your project's classpath.
[error]    Java serialization/deserialization may cause this.
[error]  * An open package is accessed across layers. If the project's classes access or extend
[error]    jvm package private classes defined in a project dependency, it may cause an IllegalAccessError
[error]    because the jvm enforces package private at the classloader level.
[error] 
[error] These issues, along with others that were not enumerated above, may be resolved by changing the class loader layering strategy.
[error] The Flat and ScalaLibrary strategies bundle the full project classpath in the same class loader.
[error] To use one of these strategies, set the  ClassLoaderLayeringStrategy key
[error] in your configuration, for example:
[error] 
[error] set ember-server / Test / classLoaderLayeringStrategy := ClassLoaderLayeringStrategy.ScalaLibrary
[error] set ember-server / Test / classLoaderLayeringStrategy := ClassLoaderLayeringStrategy.Flat
[error] 
```

It gives multiple options. Its dang weird and I avoid these generally. Which one should I put for this?

armanbilge:
I think you need `Flat`? But I have no idea never encountered this before. BTW does this affect only Scala 3?

Also, what if we dump JNR and just enable/disable this test with `System.getProperty("java.version")` or something?

ember-server/src/main/scala/org/http4s/ember/server/internal/ServerHelpers.scala
```diff
@@ -69,7 +71,89 @@ private[server] object ServerHelpers {
...
+      maxHeaderSize: Int,
+      requestHeaderReceiveTimeout: Duration,
+      idleTimeout: Duration,
+      logger: Logger[F]
+  ) = {
+    val server =
+      // Our interface has an issue
+      Stream
+        .eval(
+          ready.complete( // This is a lie, there isn't any signal from fs2 when the server is actually ready
```
RaasAhsan, line 115: Should we call this after `unixSockets.server` has returned ?

client/src/main/scala/org/http4s/client/middleware/UnixSocket.scala
```diff
@@ -0,0 +1,28 @@
...
+
+package org.http4s.client.middleware
+
+import org.http4s.client.Client
+import org.http4s.Request
+import cats.effect.kernel.MonadCancelThrow
+
+/** Middleware to direct all requests to the provided `UnixSocketAddress` */
+object UnixSocket {
+  def apply[F[_]: MonadCancelThrow](address: fs2.io.net.unixsocket.UnixSocketAddress)(
```
RaasAhsan, line 25: Is this necessary because the current http4s model can't specify unix socket addresses as destinations ?

ember-server/src/main/scala/org/http4s/ember/server/internal/ServerHelpers.scala
armanbilge: `unixSockets.server` returns a stream of sockets so there's not much wiggle room there I think.

RaasAhsan: in the past, this kind of behavior has led to poor UX and flaky test suites since we don't know when we are cleared to make a request to the server. we could instrument retries but that isn't a silver bullet and it also pollutes the test suites

client/src/main/scala/org/http4s/client/middleware/UnixSocket.scala
ChristopherDavenport: UnixSockets doing http is basically unspecified. This was intended as a convenient way to automatically redirect all requests to the UnixSocket.

rossabaker:
Is the goal for this to make tomorrow's releases?  I've been concentrating on the bugfixes and the otherwise unreviewed.  I trust the trio who are making/reviewing this.  Let me know, because I'm trying to land the plane.

ChristopherDavenport:
I certainly think it can. This is solid with very little api changes. Trying to get the CI working, and as soon as that's in place this is good to go.

ChristopherDavenport:
Build error is unrelated

```
==> X org.http4s.ember.core.TraversalSpec.Request Encoder/Parser should preserve body with a known uri  0.317s munit.FailException: Failing seed: VXcWOhcs-bCqckiZKeLXJPo6a0GGLAgh1mRvQvbK6YC=

  override val scalaCheckInitialSeed = "VXcWOhcs-bCqckiZKeLXJPo6a0GGLAgh1mRvQvbK6YC="

Exception raised on property evaluation.
> ARG_0: Request(method=ORDERPATCH, uri=http://228.168.127.128?%EE%9A%A3=%E4%B2%AE#, headers=Headers(Y: *O))
> ARG_1: "﻿"
> Exception: munit.ComparisonFailException: /home/runner/work/http4s/http4s/ember-core/src/test/scala/org/http4s/ember/core/TraversalSpec.scala:96 Obtained empty output!
95:
96:      res.assertEquals(s)
97:    }
```

armanbilge (approved):

ember-server/src/main/scala/org/http4s/ember/server/internal/ServerHelpers.scala
ChristopherDavenport: We agree in discord we need to look into the fs2 interface at some point. But that this is alright for where we are at now.

ChristopherDavenport:
@rossabaker Since you are getting the ship ready. Pinging you to let you know this is ready, without forcing your hand on merge order.

rossabaker:
I naively expected this to be triggered by the URI of the request, but it looks like there's not yet a consensus on that.  Interesting reading: https://github.com/whatwg/url/issues/577

ember-server/src/main/scala/org/http4s/ember/server/internal/ServerHelpers.scala
```diff
@@ -69,7 +71,89 @@ private[server] object ServerHelpers {
...
+      // Defaults
+      errorHandler: Throwable => F[Response[F]],
+      onWriteFailure: (Option[Request[F]], Response[F], Throwable) => F[Unit],
+      maxConnections: Int,
+      receiveBufferSize: Int,
+      maxHeaderSize: Int,
+      requestHeaderReceiveTimeout: Duration,
+      idleTimeout: Duration,
+      logger: Logger[F]
+  ) = {
```
johnynek, line 110: Should we have a return type here?

ember-server/src/main/scala/org/http4s/ember/server/internal/ServerHelpers.scala
```diff
@@ -69,7 +71,89 @@ private[server] object ServerHelpers {
...
+      errorHandler: Throwable => F[Response[F]],
+      onWriteFailure: (Option[Request[F]], Response[F], Throwable) => F[Unit],
+      maxConnections: Int,
+      receiveBufferSize: Int,
+      maxHeaderSize: Int,
+      requestHeaderReceiveTimeout: Duration,
+      idleTimeout: Duration,
+      logger: Logger[F],
+      createRequestVault: Boolean
+  ) = {
```
johnynek, line 156: Return type?

johnynek (commented):
Thanks for doing this! I'm interested in using this for a project.

rossabaker:
Let's do two release trains this week.  I need a stable point so I can release the pending GHSA.

ember-server/src/main/scala/org/http4s/ember/server/EmberServerBuilder.scala
```diff
@@ -129,18 +135,51 @@ final class EmberServerBuilder[F[_]: Async] private (
     copy(requestHeaderReceiveTimeout = requestHeaderReceiveTimeout)
   def withLogger(l: Logger[F]) = copy(logger = l)
 
+  // If used will bind to UnixSocket
+  def withUnixSocketConfig(
+      unixSockets: UnixSockets[F],
+      unixSocketAddress: UnixSocketAddress,
+      deleteIfExists: Boolean = true,
+      deleteOnClose: Boolean = true) =
```
rossabaker, line 143: Multiple parameters with defaults may hurt for binary compatibility in the future, but we can always give it a different name.  The alternative is to wrap this up in a type, but then _it_ needs a builder.  Nested, correlated options are nasty no matter what we do.

ember-server/src/test/scala/org/http4s/ember/server/EmberUnixSocketSuite.scala
```diff
@@ -0,0 +1,55 @@
...
+import cats.effect._
+import org.http4s._
+import org.http4s.ember.client.EmberClientBuilder
+import fs2.io.net.unixsocket.{JnrUnixSockets, UnixSocketAddress}
+import org.http4s.client.middleware.UnixSocket
+import scala.concurrent.duration._
+
+class EmberUnixSocketSuite extends Http4sSuite {
+
+  test("Should be able to connect to a server created") {
```
rossabaker, line 29: Does this test need to be conditional on platform?  Does this run on Windows?  We've had a handful of contributors on Windows.  I suppose one of them can let us know if it's an issue.

ember-server/src/test/scala/org/http4s/ember/server/EmberUnixSocketSuite.scala
rossabaker: The JVM has a system property for the temp directory, `java.io.tmpdir`, which would be polite to honor.  It's not always `/tmp`.

rossabaker (approved):
None of these comments are blockers

ember-server/src/test/scala/org/http4s/ember/server/EmberUnixSocketSuite.scala
armanbilge: Ah, but that's a JVMism ;) I agree with the sentiment though, the fs2.io.files method for temporary files/directories would be the best way to go about this.

ember-server/src/test/scala/org/http4s/ember/server/EmberUnixSocketSuite.scala
armanbilge: Well, JNR doesn't suport windows. https://github.com/jnr/jnr-unixsocket/issues/83

ember-server/src/test/scala/org/http4s/ember/server/EmberUnixSocketSuite.scala
rossabaker: Does that do something that honors `$TMPDIR` on JS?  Regardless, I like the idea of outsourcing that portability concern to fs2 if they've already solved it.

ember-server/src/test/scala/org/http4s/ember/server/EmberUnixSocketSuite.scala
rossabaker: And I'll bet JDK17 fails a different way on Windows.

I don't particularly want to get into adding multiple OSes to our build matrix (though we've got another one with the resource service this week...), but I predict within six months we're going to have a contributor ask WTF on this test.  We can guard this test now or guard it then. :smile:

ember-server/src/test/scala/org/http4s/ember/server/EmberUnixSocketSuite.scala
armanbilge: That's a good question. fs2 just delegates to Node.js `os.tmpdir()` method which "Returns the operating system's default directory for temporary files" whatever that means.

rossabaker:
This one has a gruesome merge with the web socket fix.

armanbilge:
I'll try and fix this up. No promises 😅

armanbilge:
Another appearance of #5128. Also on Java 16. 🤔

PR #5226, opened 2021-09-19
Migrate to new fs2.io.file APIs

armanbilge:
Deprecates JDK file APIs in favor of the new fs2.io.file APIs introduced in 3.1.0. Also part 2 of https://github.com/http4s/http4s/issues/5044.

~Blocked~ by https://github.com/typelevel/fs2/pull/2622. Edit: not actually blocked, but that PR would certainly help. I added a TODO comment.

### DIFF
--- a/core/src/main/scala/org/http4s/StaticFile.scala
+++ b/core/src/main/scala/org/http4s/StaticFile.scala
@@ -16,7 +16,7 @@
 
 package org.http4s
 
-import cats.{Functor, MonadError, MonadThrow, Semigroup}
+import cats.{ApplicativeThrow, Functor, MonadError, MonadThrow, Semigroup}
 import cats.data.{NonEmptyList, OptionT}
 import cats.effect.{Sync, SyncIO}
 import cats.syntax.all._
@@ -130,7 +130,7 @@ object StaticFile {
         .map(isFile =>
           if (isFile) s"${f.lastModified().toHexString}-${f.length().toHexString}" else "")
 
-  def calculateETag[F[_]: Files: Functor](implicit dummy: DummyImplicit): Path => F[String] =
+  def calculateETag[F[_]: Files: ApplicativeThrow]: Path => F[String] =
     f =>
       Files[F]
         .getBasicFileAttributes(f)
@@ -138,6 +138,8 @@ object StaticFile {
           if (attr.isRegularFile)
             s"${attr.lastModifiedTime.toMillis.toHexString}-${attr.size.toHexString}"
           else "")
+        // A bit of a hack, old code relied on Java File which silently returns 0s when there is an I/O exception...
+        .handleError(_ => s"${0.toHexString}-${0.toHexString}")
 
   @deprecated("Use fromPath", "0.23.4")
   def fromFile[F[_]: Files: MonadThrow](
@@ -186,8 +188,10 @@ object StaticFile {
       buffsize: Int,
       req: Option[Request[F]],
       etagCalculator: Path => F[String]): OptionT[F, Response[F]] =
-    OptionT.liftF(Files[F].getBasicFileAttributes(f)).flatMap { attr =>
-      fromPath(f, 0, attr.size, buffsize, req, etagCalculator)
+    // A bit of a hack, old code relied on Java File which silently returns 0L if there is an I/O exception
+    OptionT.liftF(Files[F].getBasicFileAttributes(f).map(_.size).handleError(_ => 0L)).flatMap {
+      size =>
+        fromPath(f, 0, size, buffsize, req, etagCalculator)
     }
 
   @deprecated("Use fromPath", "0.23.4")
--- a/core/src/main/scala/org/http4s/StaticFile.scala
+++ b/core/src/main/scala/org/http4s/StaticFile.scala
@@ -122,7 +122,7 @@ object StaticFile {
     })
   }
 
-  @deprecated("Use overload with fs2.io.file.Path", "0.23.4")
+  @deprecated("Use calculateETag", "0.23.4")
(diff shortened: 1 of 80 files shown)


armanbilge:
I think this is good to go.

armanbilge:
I'm going to take another pass at this now that https://github.com/typelevel/fs2/pull/2622 is released in fs2 3.1.3.

server/src/main/scala/org/http4s/server/staticcontent/FileService.scala
```diff
@@ -58,69 +57,75 @@ object FileService {
...
+                          case (_, "" | "." | "..") => throw BadTraversal
+                          case (path, segment) =>
+                            path.resolve(segment)
+                        }
+                      })
+                }
+                resolvedPath
+                  .flatMapF(path =>
+                    F.ifM(Files[F].exists(path, false))(
+                      path.absolute.normalize.some.pure,
```
armanbilge, line 91: This line is replacing `path.toRealPath(LinkOption.NOFOLLOW_LINKS)`. I'm not 100% that the semantics are exactly the same, but AFAICT making a path into a "real path" _without_ following links is essentially making absolute and then normalizing. This is what the [Javadoc](https://docs.oracle.com/javase/8/docs/api/java/nio/file/Path.html#toRealPath-java.nio.file.LinkOption...-) says for `toRealPath`:
> The precise definition of this method is implementation dependent but in general it derives from this path, an absolute path that locates the same file as this path, but with name elements that represent the actual name of the directories and the file.

armanbilge:
The good news is I was able to re-work all the file stuff. The bad news is I managed to replicate whatever happened in #5268.

server/src/main/scala/org/http4s/server/staticcontent/FileService.scala
armanbilge: I took a look at the source for `path.toRealPath(LinkOption.NOFOLLOW_LINKS)` in JDK, at least for Unix:
https://github.com/openjdk/jdk/blob/5ec1cdcaf39229a7d2457313600b0dc2bf8c6453/src/java.base/unix/classes/sun/nio/fs/UnixPath.java#L814

This is what it's doing:
1. `toAbsolutePath()`
2. eliminate "." and also ".." where the previous element is not a link
3. check file exists (without following links)

So, the only part that's different is the `..` handling. _However_, in `FileService` paths that include `..` are already rejected. So, I think it is safe to make this change.

core/src/main/scala/org/http4s/StaticFile.scala
```diff
@@ -242,5 +312,5 @@ object StaticFile {
     }
 
   private[http4s] val staticFileKey =
-    Key.newKey[SyncIO, File].unsafeRunSync()
+    Key.newKey[SyncIO, Path].unsafeRunSync()
```
rossabaker, line 315: This has a slight chance of causing a problem if used in one of the backends outside this repo.  How about a `staticPathKey`?

armanbilge: Woah, scary. Isn't the great schism going to be full of bincompat landmines like this?

core/src/main/scala/org/http4s/StaticFile.scala
```diff
@@ -122,31 +122,79 @@ object StaticFile {
...
+      req,
+      etagCalculator.compose(_.toNioPath.toFile()))
 
+  def fromPath[F[_]: Files: MonadThrow](
+      f: Path,
+      buffsize: Int,
+      req: Option[Request[F]],
+      etagCalculator: Path => F[String]): OptionT[F, Response[F]] =
+    // A bit of a hack, old code relied on Java File which silently returns 0L if there is an I/O exception
+    OptionT.liftF(Files[F].getBasicFileAttributes(f).map(_.size).handleError(_ => 0L)).flatMap {
```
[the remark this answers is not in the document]
armanbilge: Good point. I'll have to adjust the tests as well.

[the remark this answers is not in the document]
armanbilge: Ah, I just remembered there's some subtlety to this. This bug only cropped up when I started testing `FileService` in the server module. Essentially, it's an assumption made by the code there ... I think to fix this, will require some non-trivial reworking of the logic in `FileService`. On it.

armanbilge:
Drafting, pending a rework of `FileService` to not rely on the aforementioned `StaticFile` bug/hack. See https://github.com/http4s/http4s/pull/5226#discussion_r716247839.

core/src/main/scala/org/http4s/StaticFile.scala
```diff
@@ -232,15 +302,18 @@ object StaticFile {
...
       case i => MediaType.forExtension(name.substring(i + 1)).map(`Content-Type`(_))
     }
 
-  private[http4s] val staticFileKey =
-    Key.newKey[SyncIO, File].unsafeRunSync()
+  private[http4s] val staticPathKey = Key.newKey[SyncIO, Path].unsafeRunSync()
+
+  @deprecated("Use staticPathKey", since = "0.23.5")
+  private[http4s] val staticFileKey: Key[File] =
+    staticPathKey.imap(_.toNioPath.toFile)(f => Path.fromNioPath(f.toPath))
```
armanbilge, line 318: @tpolecat this is absolutely brilliant.

core/src/main/scala/org/http4s/StaticFile.scala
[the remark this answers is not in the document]
armanbilge: I was wrong, fixing this didn't require any other changes. 👍

rossabaker:
I think you inherited that bincompat mess from my bad merge.

PR #5229, opened 2021-09-20
Make EntityEncoder covariant in its effect type

bplommer:
The changes to signature instances are source-breaking but IMO worthwhile with a scalafix (which I’m happy to add) - I’ve aways found it confusing that all `EntityEncoders` seemingly needed to be effectful.


armanbilge:
These PRs improving typing are really cool!

bplommer:
> Could this really make 0.22 with the breaking changes? I assumed this was targeted at 1.0.

Yep, the only changes are to type parameters, which are erased. In fact I think the mima exclusions are no longer needed either.

armanbilge:
Right, but it's still syntax breaking?

bplommer:
Right, sorry, yes - it's definitely source-breaking. I can retarget this at main if that's too heavy a price to pay.

armanbilge:
> I can retarget this at main if that's too heavy a price to pay.

Someone else would have to weigh in on this. It's a bit confusing since we're still in `0.y.z` versioning. In `x.y.z` a bump in `z` is a patch but a bump in `y` _can_ break sources since binary compact is guaranteed by `x`. Here, we rely on `y` for binary compatibility and `z` is patches.

bplommer:
One option could be to make the variance change now but leave the instance chances to 1.0 - that would be source-compatible (modulo a few type inference issues) and still allow client code to freely use `Encoder.Pure`

armanbilge:
> modulo a few type inference issues

Still sounds scary IMO 😆 we were recently burned by this in https://github.com/typelevel/cats-effect/issues/2271.

bplommer:
Hrm you're probably right. I'm usually over-eager to get shiny new stuff into production 😆

armanbilge:
I know the feeling!!

bplommer:
Updated to target `main`.

bplommer:
Some of the changes to modules seem to have been lost in the merge though - I'll resist the urge to fix those up first this morning.

bplommer:
Fairly sure the failure is unrelated.

bplommer:
Failure is a blaze flake.

rossabaker:
>> If F[_] should be covariant, should A?

> This PR doesn't make F covariant! It makes EntityEncoder covariant with respect to F. A is an input type so we could make EntityEncoder contravariant in A, though I don't know how much benefit there would be.

Mine was a bad question, but @diesalbla's PR explores the correct question.

>> If F[_] should be covariant, should it be contravariant in EntityDecoder?

> I don't think that's possible, as F appears in both the input and output types of EntityDecoder.

Correct.  The use of `F` is not symmetrical.

>  (even Argonaut's codec type classes were co- and contravariant until 6.1),

It annoyed them enough to bring it back in 6.2 milestones, and then was reverted again before 6.2.  Circe and Skunk are both invariant.  Play removed contravariance.  I think all of this was decided before the SI-2509 fix, but it's my understanding that users would need that compiler flag to make it practical even on 2.13.

> [good conversation about typeclasses]

I'm asking whether EntityEncoder and EntityDecoder are type classes because there is little tradition of variant typeclasses, and the trend has been away from them, particularly in codecs.  We pass them implicitly (unlike Skunk!), which I think matters more in this discussion than whether they are "true type classes."  

I wonder whether maintaining invariance on the second parameter has kept us sufficiently in the mainstream here that it works well.

PR #5233, opened 2021-09-20
Add override to fix JDK17 build

armanbilge:
Seems they must have added a default implementation.


rossabaker:
And I guess we get away with the override because the abstract version dates back to Java 8?

rossabaker (approved):

armanbilge:
Clock itself only seems to have been added in Java 8 :)

Issue #5236, opened 2021-09-20
dom-service-worker-tests: Reference Error: Headers is not defined


rossabaker:
Getting this locally.  I'm guessing I've failed to install another dependency.

```
==> X org.http4s.dom.ServiceWorkerSuite.ServiceWorker Execute GET /internal-server-error 0.00s scala.scalajs.js.JavaScriptException: ReferenceError: Headers is not defined
    at <jscode>.{anonymous}()(/Users/ross.baker/src/http4s/dom-service-worker-tests/.js/target/scala-2.12/http4s-dom-service-worker-tests-test-fastopt/main.js:57864)
    at scala.scalajs.runtime.AnonFunction1.apply(/Users/ross.baker/src/http4s/dom-service-worker-tests/.js/target/scala-2.12/http4s-dom-service-worker-tests-test-fastopt/main.js:131705)
    at cats.effect.IOFiber.runLoop(/Users/ross.baker/src/http4s/dom-service-worker-tests/.js/target/scala-2.12/http4s-dom-service-worker-tests-test-fastopt/main.js:221959)
    at cats.effect.IOFiber.execR(/Users/ross.baker/src/http4s/dom-service-worker-tests/.js/target/scala-2.12/http4s-dom-service-worker-tests-test-fastopt/main.js:222730)
    at cats.effect.IOFiber.run(/Users/ross.baker/src/http4s/dom-service-worker-tests/.js/target/scala-2.12/http4s-dom-service-worker-tests-test-fastopt/main.js:223185)
    at <jscode>.{anonymous}()(/Users/ross.baker/src/http4s/dom-service-worker-tests/.js/target/scala-2.12/http4s-dom-service-worker-tests-test-fastopt/main.js:88589)
    at scala.scalajs.runtime.AnonFunction0.apply(/Users/ross.baker/src/http4s/dom-service-worker-tests/.js/target/scala-2.12/http4s-dom-service-worker-tests-test-fastopt/main.js:131681)
    at <jscode>.Immediate.$c_sjs_js_Any$.fromFunction0__F0__sjs_js_Function0 [as _onImmediate](/Users/ross.baker/src/http4s/dom-service-worker-tests/.js/target/scala-2.12/http4s-dom-service-worker-tests-test-fastopt/main.js:131647)
    at <jscode>.processImmediate(node:internal/timers:464)
```

rossabaker:
And also dom-fetch-clientJS/test:

```
[...]

Issue #5230, opened 2021-09-20
Tests should start their own server scaffold


rossabaker:
Was struggling to run `client/test` on main.  Needed to manually start scripts/scaffold_server.js.  We should have our test suites manager their own scaffolds for local development.  I'd completely forgotten this was there until @ChristopherDavenport reminded me.

armanbilge:
Yes, sorry about this. The problem is we cannot rely on the test suites' ability to do this because it has a rough translation across JVM/Node.js and no translation for browsers. I think I proposed at some point we build these scaffold servers into the build project itself.

armanbilge:
Found it: https://github.com/http4s/http4s/pull/4938#discussion_r683924911

Unless there's any objections I'll try and put that together.

armanbilge:
Ok, maybe not. In discord I got some pushback on sbt build-hacking

> We should have our test suites manager their own scaffolds for local development.

If we want to go back to this, then I think we should great-schism the dom modules out of the main repo. It's only browsers where this is a problem, so if we don't have to test on browsers a lot of these hacks will vanish.

rossabaker:
I don't understand what's being proposed as "sbt build hacking".

I mostly would like to be able to run `sbt test` and have things just work.  Bonus points if we can drop the Python dependency.  More bonus points if we can write test endpoints in Scala.  I'm currently needing to add one in Node to port what I added in Scala on 0.23.

armanbilge:
So, the way these scaffold servers were previously run (and continue so on series/0.22 and 0.23) were as pure-Scala test fixtures as part of the test suites themselves.

This trick didn't work so well after the JS port because (1) it used a JDK server implementation and (2) there's no way a suite running in a browser can start its own server to test against.

My "sbt build hack" proposal is that we implement the scaffold servers in Scala as part of the sbt project. Then, we can start them automatically from within sbt whenever someone runs `test`.

rossabaker:
That proposal sounds right to me, though I don't know how people run individual tests.  I mostly use sbt inside my editor, because I still don't understand bloop.

armanbilge:
I've asked @ChristopherDavenport and @tpolecat to weigh in b/c at least on discord they didn't seem so crazy about this idea.

tpolecat:
I have no opinion, I was just being a wiseass.

armanbilge:
Fixed in #5241.

PR #5237, opened 2021-09-20
Add sbt alias to run all tests everywhere

armanbilge:
Fixes https://github.com/http4s/http4s/issues/5236. @rossabaker can you try locally? Although you will need a ton of stuff installed (Node, Firefox, Chrome, geckodriver, chromedriver).

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -142,6 +142,11 @@ addCommandAlias(
   "ciChrome",
   "; set parallelExecution := false; set Global / useJSEnv := JSEnv.Chrome; project rootBrowser")
 
+addCommandAlias(
+  "testEverything", // Runs the entire +test suite on JVM, Node.JS, Firefox, and Chrome
+  "; project rootJVM; +test; project rootNodeJS; set Global / useJSEnv := JSEnv.NodeJS; +test; project rootBrowser; set Global / useJSEnv := JSEnv.Firefox; +test; set Global / useJSEnv := JSEnv.Chrome; +test; project /; set Global / useJSEnv := JSEnv.NodeJS"
+)
+
 enablePlugins(SonatypeCiReleasePlugin)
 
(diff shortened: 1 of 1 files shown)


rossabaker:
I would like to identify all those things and get them in the nix-shell so they're there on all my machines and for any other nix folks who happen along.  I don't even know what two of them are, but I'll chip away at it.

rossabaker:
Would this be a good time to spin off the browser modules into their own repo?  Since they're not being maintained in upstream branches, it seems that those could get a head start on The Great Schism.

armanbilge:
Yep. This is what I was saying earlier today in https://github.com/http4s/http4s/issues/5230#issuecomment-923063605.

rossabaker:
Ah, I don't think I fully understood what that meant yet.  Preparing this release is the first time I've had to get all this machinery running locally, so I'm only gradually beginning to understand it.  I think that would make a lot of sense.  It'll be the first thing that has a different release cycle than 1.0, but it'll give us some runtime getting used to those problems before the modules that are widely in production hit it.  I think it will be good for everything.

armanbilge:
As I'm learning how you do things here I think so too, the dom modules are a sore thumb in the main repo. Daniel was ecstatic when we moved a lot of the browser testing in cats-effect to scala-js-macrotask-executor although we couldn't completely purge it.

rossabaker:
What's an appropriate name? http4s-browser?  http4s-dom?  Do all of these modules belong in their own?

armanbilge:
I'd say `http4s-dom`. These modules can all go together.

rossabaker:
http4s-dom exists now.

armanbilge:
Thanks. Give me a couple hours and hopefully I'll have a PR to yank it out of main 👍

armanbilge:
Superseded by #5241.

PR #5241, opened 2021-09-21
Au revoir dom

armanbilge:
Deletes the DOM modules and anything to do with browsers now that they live in https://github.com/http4s/http4s-dom.

This is only one in a series of PRs that will restore sanity to main.


armanbilge:
Just pushed a fix for https://github.com/http4s/http4s/issues/5230. No more Python or external servers.

armanbilge:
Ok, I think this should be good-to-go. I did have to disable a test route in the scaffold server due to what I suspect is an ember bug.

The JVM scaffold server returns:
```
HTTP/1.1 204 No Content Headers(Date: Tue, 21 Sep 2021 03:11:06 GMT) body=""
```

Causing ember-client (and _only_ ember-client, all other clients are happy) to fail with:
```
org.http4s.ember.client.internal.ClientHelpers$$anonfun$request$6$$anon$1: Remote Disconnect: Received zero bytes after sending request
```

rossabaker:
I've already staged 1.0.0-M27 from the GHSA, but this can take over on the next milestone.

armanbilge:
I think I need help with this merge with the latest GHSA changes.

rossabaker:
I think the conflicty bits were all in ClientRouteTestBattery already.

armanbilge:
@rossabaker yup, those new tests are failing for the ok and async clients. But not for ember client. I couldn't understand why.

rossabaker:
Well, crap. :eyes:

armanbilge:
Gah, yes it has become crazy. Ideally we'd have a _real_ sun server implementation that can just use an `HttpApp`, so then I think we'd only need this once?

armanbilge:
Nice! 👍 thanks for the assist.

rossabaker:
I've thought about a real Sun Server, but I'm generally trying to bash bugs on the backends we have before adding another.  Also, being an http4s backend is part of what gets us into circular dependencies.  That's why the old ways were to use raw Jetty, or your Node scaffold.

armanbilge:
I mean, it depends how you define "real"? I don't see how having an actual Sun Server (i.e., something built around an `HttpApp`) inside the client tests module would introduce any circular dependencies. Because we essentially have a poor-man's version of that right now anyway.

rossabaker:
A raw Sun server is fine.  An http4s-sun-server in tests would add an http4s-server test dependency in http4s-client.  Or using an http4s-client backend in server tests would introduce a test dependency the other way.

armanbilge:
> add an http4s-server test dependency in http4s-client

You mean like the one that's already there? https://github.com/http4s/http4s/blob/series/0.23/build.sbt#L237

rossabaker:
Yes, like that.  I thought we'd gotten rid of it.

armanbilge:
I still don't see why it's bad, client depends on server but not the other way around so no circle.

In any case, I don't think it would be hard to get rid of if we're only using the `Server` trait, which is really a glorified container for an IP address IIUC.

rossabaker:
It's really not _that_ harmful, which is how I forgot it was there.  It's arbitrary: why client on server, and not server on client?  It's a coupling we don't need, and it's more that needs to compile to work on client.  There was once talk of server and client being in different repos, though I don't think anyone is advocating for that anymore.  So we've probably all got better things to work on.

PR #5247, opened 2021-09-22
Match on ClosedChannelException for EmptyStream since it's being adap…

matthughes:
…ted above.

### DIFF
--- a/ember-client/src/main/scala/org/http4s/ember/client/internal/ClientHelpers.scala
+++ b/ember-client/src/main/scala/org/http4s/ember/client/internal/ClientHelpers.scala
@@ -220,9 +220,6 @@ private[client] object ClientHelpers {
     def isEmptyStreamError[F[_]](result: Either[Throwable, Response[F]]): Boolean =
       result match {
         case Right(_) => false
-        // case Left(EmberException.EmptyStream()) => true // Next version can be accessed by users
-        case Left(org.http4s.ember.core.EmptyStreamError()) =>
-          true // Note this is private in http4s in 0.21
         case Left(_: ClosedChannelException) => true
         case _ => false
       }
--- a/ember-client/src/main/scala/org/http4s/ember/client/internal/ClientHelpers.scala
+++ b/ember-client/src/main/scala/org/http4s/ember/client/internal/ClientHelpers.scala
@@ -223,6 +223,7 @@ private[client] object ClientHelpers {
         // case Left(EmberException.EmptyStream()) => true // Next version can be accessed by users
         case Left(org.http4s.ember.core.EmptyStreamError()) =>
           true // Note this is private in http4s in 0.21
+        case Left(_: ClosedChannelException) => true
         case _ => false
       }
   }


rossabaker:
I've been chasing a lot of things.  Which release did you identify the bug on?  Are you running 1.0 or something earlier?

rossabaker:
I ask because we might want to cherry-pick this back to another branch.  I'm expecting to run another release train this week.  I don't want you or anyone else to waste time on that merge conflict if this is getting backported.

rossabaker:
Looks like at least 0.23.

RaasAhsan:
The bugfix Chris made (https://github.com/http4s/http4s/pull/5064) was merged into `series/0.21`, so we may want to cherrypick to there or `series/0.22`

ember-server/src/test/scala/org/http4s/ember/server/EmberServerSuite.scala
```diff
@@ -92,4 +92,30 @@ class EmberServerSuite extends Http4sSuite {
...
+
+      val body: Stream[IO, Byte] =
+        Stream.emits(Seq("hello")).through(fs2.text.utf8Encode).covary[IO]
+
+      val uri = Uri
+        .fromString(s"http://${server.address.getHostName}:${server.address.getPort}/echo")
+        .toOption
+        .get
+      val request = GET(body, uri)
+      EmberClientBuilder.default[IO].withIdleConnectionTime(1.seconds).build.use { client =>
```
RaasAhsan, line 110: I believe this is the opposite of what we want. we are interesting in seeing what happens when the client timeout is _longer_ than the server timeout i.e. what happens if the server closes their end of the connection and the client try to make another request

RaasAhsan (changes requested):

ember-server/src/test/scala/org/http4s/ember/server/EmberServerSuite.scala
RaasAhsan: Also the right client configuration option to use here is `withIdleTimeInPool` -- idle connection time controls how much time the client will wait for a response after a request has been sent before timeout

RaasAhsan: Ugh, actually, I don't think this will test exactly what we want. At worst this will induce a connection reset by peer error which isn't what we want. I think we should just merge in the fix for now and think about how to do the test later, since none of the servers we have available let us emulate that behavior

matthughes:
I removed the test and targeted it to 0.21 branch.  This should be ok to be cherry picked forward.

ember-server/src/test/scala/org/http4s/ember/server/EmberServerSuite.scala
matthughes: @rossabaker Updated the PR.

RaasAhsan (approved):

ember-client/src/main/scala/org/http4s/ember/client/internal/ClientHelpers.scala
```diff
@@ -223,6 +223,7 @@ private[client] object ClientHelpers {
         // case Left(EmberException.EmptyStream()) => true // Next version can be accessed by users
         case Left(org.http4s.ember.core.EmptyStreamError()) =>
```
RaasAhsan, line 224: i'm guessing this match clause isn't necessary anymore but deferring to @ChristopherDavenport on that one

matthughes: Yeah I wasn't sure as to whether we should keep that in there or not.  It's not currently getting matched on.

rossabaker: It looks to me like it can only be raised where it has been adapted.  I think it can go.

ChristopherDavenport (approved):

PR #5249, opened 2021-09-22
Fix fatal warning on ClientRouteTestBattery

rossabaker:
We don't have fatal warnings locally, so I didn't catch this in the security release.


armanbilge (approved):

rossabaker:
But wait, there's more. 😦

hamnis (approved):

PR #5251, opened 2021-09-22
Make message types covariant in their effect type

bplommer:
Depends on #5229.

It's a big diff so here's a breakdown:

- ff9dca0 is the core change. 
- b3f6010 removes no-longer-needed type arguments and annotations from production code; 
ccf9851 does the same for tests. 
- 6e1827b introduces `AnyRequest`, `AnyResponse`, `AnyMessage` as aliases for respective top types and uses them where possible. (it's debatable which of these uses are actually a good idea).
- 8d1727 and 2aa5c0e make `main` and `test`, respectively, compile on Scala 2.12. (It can seemingly handle using `Response.covaryF` for the implicit conversion from `F[Response[Nothing]]` to `F[Response[F]]` but not to `F[Response[G]]`)

### DIFF
--- a/blaze-server/src/test/scala/org/http4s/blaze/server/Http1ServerStageSpec.scala
+++ b/blaze-server/src/test/scala/org/http4s/blaze/server/Http1ServerStageSpec.scala
@@ -213,7 +213,7 @@ class Http1ServerStageSpec extends Http4sSuite {
       .of[IO] { case _ =>
         val headers = Headers(H.`Transfer-Encoding`(TransferCoding.identity))
         IO.pure(
-          Response[IO](headers = headers)
+          Response(headers = headers)
             .withEntity("hello world"))
       }
       .orNotFound
@@ -238,7 +238,7 @@ class Http1ServerStageSpec extends Http4sSuite {
       val routes: HttpApp[IO] = HttpRoutes
         .of[IO] { case _ =>
           IO.pure(
-            Response[IO](status = Status.NotModified)
+            Response(status = Status.NotModified)
               .putHeaders(`Transfer-Encoding`(TransferCoding.chunked))
               .withEntity("Foo!"))
         }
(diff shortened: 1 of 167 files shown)


blaze-server/src/test/scala/org/http4s/blaze/server/Http1ServerStageSpec.scala
```diff
@@ -535,79 +535,35 @@ class Http1ServerStageSpec extends Http4sSuite {
...
-          _ <- uncancelableStarted.get
-          _ <- uncancelableCanceled.get
-          _ <- IO(head.sendInboundCommand(Disconnected))
-          head2 <- IO(runRequest(tw, reqCancelable, app))
-          _ <- cancelableStarted.get
-          _ <- IO(head2.sendInboundCommand(Disconnected))
-          _ <- cancelableCanceled.get
-        } yield ()).assert
-    }
+  fixture.test("Http1ServerStage: don't deadlock TickWheelExecutor with uncancelable request") {
```
bplommer, line 538: Merge error - fix

bplommer:
Is the principle of this controversial or is it just the implementation that needs scrutiny? I can break it into smaller PRs if that helps.

diesalbla:
@bplommer What would be the state for this Pull Request? You have proposed some of the changes in this PR in separate Pull Requests:  

- #5712 Covariant Messages - Minimum 
- #5463 Entity Encoder - Variance of Type Parameters.

That being the case, perhaps the next step with this PR would be to either:
a) Close it, to indicate that it the intended changes are already done; or
b) Either merge `main` into it or rebase the changes on top of current `master`, to discount what has already been done.

armanbilge:
@bplommer notes in Discord that
> The "make message types covariant" one can actually probably be closed as the core change was done in another pr, though there's some followup work to take full advantage of it

/cc @diesalbla who asked above :)

PR #5252, opened 2021-09-22
Nontrivial merges 2021 09 22, 1.0-edition

armanbilge:
Continuing from #5250. Lots of delayed gratification here!


rossabaker (approved):
Nice cleanup. :+1: if green.

PR #5253, opened 2021-09-22
Deprecate custom status reason phrases

rossabaker:
This was a silly thing to have a CVE on.  For precedent, servlets haven't let you set them since nineteen dickety-two, and Finagle also prohibits them.

Also removes the ability to set `isEntityAllowed` on a custom status.  I'd rather leave that decision to the IANA.


rossabaker:
Failures are https://github.com/http4s/http4s/pull/5258

Issue #5259, opened 2021-09-23
HttpVersion exposes non-validating copy constructor


rossabaker:
D'oh:

```scala
scala> HttpVersion(1, 0)
HttpVersion(1, 0)
res3: org.http4s.HttpVersion = HTTP/1.0

scala> HttpVersion(1, 1).copy(major = 10, minor = -1)
HttpVersion(1, 1).copy(major = 10, minor = -1)
res4: org.http4s.HttpVersion = HTTP/10.-1
```

armanbilge:
> Scala3 seems to be smarter in regards to case classes

Yes, but it only does this with certain configurations of compiler flags. I think our migration flags disable this behavior.

armanbilge:
See also:
- https://scala.epfl.ch/projects.html#add-support-for-binary-compatible-evolutions-to-case-classes
- https://contributors.scala-lang.org/t/pre-sip-structural-data-structures-that-can-evolve-in-a-binary-compatible-way/5684

PR #5260, opened 2021-09-23
Deprecate HttpVersion.copy

rossabaker:
See #5259.

### DIFF
--- a/core/src/main/scala/org/http4s/HttpVersion.scala
+++ b/core/src/main/scala/org/http4s/HttpVersion.scala
@@ -33,6 +33,10 @@ final case class HttpVersion private[HttpVersion] (major: Int, minor: Int)
   override def render(writer: Writer): writer.type = writer << "HTTP/" << major << '.' << minor
   override def compare(that: HttpVersion): Int =
     (this.major, this.minor).compare((that.major, that.minor))
+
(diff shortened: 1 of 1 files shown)


armanbilge (approved):

core/src/main/scala/org/http4s/HttpVersion.scala
```diff
@@ -33,6 +33,10 @@ final case class HttpVersion private[HttpVersion] (major: Int, minor: Int)
   override def render(writer: Writer): writer.type = writer << "HTTP/" << major << '.' << minor
   override def compare(that: HttpVersion): Int =
     (this.major, this.minor).compare((that.major, that.minor))
+
+  @deprecated("Does not range check parameters. Will be removed from public API in 1.0.", "0.22.6")
+  def copy(major: Int = major, minor: Int = minor): HttpVersion =
```
armanbilge, line 38: removed without replacement?

rossabaker: The traditional `copy` signature doesn't allow an error channel, and I don't think it has much of a use case.

armanbilge: Yep definitely not :)

Issue #5017, opened 2021-08-04
Be Consistent About Providing An Ordering Instance


isomarcte:
We sometimes have data types which subtype `Ordered` and other times have types which implement `Ordering`. We only _need_ to do `Ordering`. We should ensure that all types which are `Ordered` also have an `Ordering` instance.

Context: https://github.com/http4s/http4s/pull/5013#discussion_r682655331

rossabaker:
@san-coding This is a good one.  `HttpDate` is an example: it extends `Ordered`, but we'd rather have an `Ordering` instance.  Are you familiar with the two types?

san-coding:
I am not familiar yet, but will look into it, thanks @rossabaker

san-coding:
@rossabaker , I can see that some classes inherit `ordered`

isomarcte:
@san-coding if a class is already extending `Ordered`, we need to leave it that way, at least for now. This is so we don't break binary compatibility. 

In the context of this ticket, any type in http4s which is currently only extending `Ordered` should also have an implicit `Ordering` added, and they should be consistent with each other.

Hopefully, most of the types which are extending `Ordered` have a `cats.kernel.Order` instance. If they do, then adding the `Ordering` is very simple. For example, 

```scala
final case class Foo(bar: Int) extends Ordered[Foo] {
    override final def compare(that: Foo): Int = 
      // Ordered derived from cats Order instance
      Order[Foo].compare(this, that)
}

object Foo {
    implicit val catsOrderInstanceForFoo: Order[Foo] = new Order[Foo] {
      override def compare(x: Foo, y: Foo): Int = x.bar.compare(y.bar)
    }
    
    // scala Ordering derived from cats Order
    implicit val stdLibOrderingInstance: Ordering[Foo] = catsOrderInstanceForFoo.toOrdering
}
```

If they don't have a `cats.kernel.Order` instance (which you are free to add as well if you'd like, or not, either way), as in the case of `HttpDate`, then we'd do something like this.

```scala
class HttpDate private (val epochSecond: Long) extends Renderable with Ordered[HttpDate] { /* elided */ }

object HttpDate {
    implicit val stdLibOrderingInstance: Ordering[HttpDate] = /* Ordering definition */
}
```

For the sake of consistency, I'd also probably make the `compare` method from `Ordered` and the `equals` method (if it's not a case class) use the underlying `Ordering` instance. Something like,


```scala
[...]

PR #5265, opened 2021-09-23
added Ordering instance to QValue Class

san-coding:
Contributing to issue #5017 

added `Ordering` instance to `QValue` class

### DIFF
--- a/core/src/main/scala/org/http4s/QValue.scala
+++ b/core/src/main/scala/org/http4s/QValue.scala
@@ -167,6 +167,9 @@ object QValue extends QValuePlatform {
     override def minBound: QValue = QValue.Zero
     override def maxBound: QValue = QValue.One
   }
+
+  implicit val stdLibOrderingInstance: Ordering[QValue] =
+    catsInstancesForHttp4sQValue.toOrdering
 }
 
 trait HasQValue {


san-coding:
@rossabaker @isomarcte , please review this, thanks

san-coding:
what other classes require `Ordering` instance

rossabaker:
`Ordering` can be implicitly derived from both `Ordered` and `cats.kernel.Order`.  I like doing these from the `Ordered`, because I'd like `Ordered` to go away.  But when there's already a `cats.kernel.Order`, those provide this, and are here to stay.

I think we should focus on providing missing `cats.kernel.Order` instances rather than continue with more of these.  Or identify more missing non-Order instances to learn more about type classes.  I'll start looking for some of these.

san-coding:
> `Ordering` can be implicitly derived from both `Ordered` and `cats.kernel.Order`. I like doing these from the `Ordered`, because I'd like `Ordered` to go away. But when there's already a `cats.kernel.Order`, those provide this, and are here to stay.
> 
> I think we should focus on providing missing `cats.kernel.Order` instances rather than continue with more of these. Or identify more missing non-Order instances to learn more about type classes. I'll start looking for some of these.

Ohh alright, so I think we need not add it to `HttpVersion` ?
https://github.com/http4s/http4s/blob/a942a83e2b88bafc644fc9916fd1979eb951ced8/core/shared/src/main/scala/org/http4s/HttpVersion.scala#L30-L32

san-coding:
### What about `Headers.scala`
https://github.com/http4s/http4s/blob/a942a83e2b88bafc644fc9916fd1979eb951ced8/core/shared/src/main/scala/org/http4s/Headers.scala#L26

rossabaker:
I am not sure what an `Order` for `Headers` would be.  I can think of a few options, and if I can think of a few, perhaps none of them are appropriate.  I'll open a new ticket and tag you.

PR #5270, opened 2021-09-24
Fix Ember TCP Disconnect Error Handling

ChristopherDavenport:


rossabaker:
I think this is #5247, but confirms that we can eliminate that other case.

PR #5279, opened 2021-09-28
Define and link all specified HTTP versions

rossabaker:
* Add constants for `HTTP/0.9` and `HTTP/3`
* Rename `HTTP/2.0` to `HTTP/2`, updating references. This is consistent with the specification and other backends.
* Document each version, linking to their specifications.
* Adds scaladoc and doc tests


rossabaker:
Nothing in particular is motivating this.  I'm just trying to raise the bar on scaladoc as I reconcile our model with the spec before 1.0.

armanbilge (approved):

PR #5286, opened 2021-09-29
Ember Server immediately close connection on timeout

RaasAhsan:
Today, Ember Server returns a 500 Internal Server Error and closes a connection after any server-related timeout occurs (socket read timeout, idle timeout, write timeout, request header receive timeout). This means that a downstream client who writes a request after the timeout response is returned will immediately see a 500 response. This is inconsistent with several widely used HTTP servers (AWS ALB/ELB, nginx, blaze-server).

This behavior means that a retry needs to be instrumented against a 500 in userspace by downstream users. I think this is poor out-of-the-box experience. We should close the connection immediately without writing any sort of response; most mature clients (blaze, okhttp, jetty) are coded to retry immediately against a closed or reset connection. The client will either observe the connection to be closed before writing their next request, or receive a RST packet after writing their request, and retry accordingly. Ember Client actually does neither and instead raises a `Connection reset by peer` error, but I'm going to open a followup to fix that. 

I would write a test for this but it's kind of difficult with any well-behaved client, so I offer this screenshot instead:
Screen Shot 2021-09-28 at 8 16 24 PM
:

PR #5290, opened 2021-09-30
Ember Client retry reset connection and broken pipe errors

RaasAhsan:
This is a follow up for #4935 and #5286. To sum it up, there are more IO exceptions from which we can recover and retry idempotent requests.

While the particular conditions I'm trying to observe here are very well-defined within the TCP layer, the various IO syscalls and the Java IO API make it impossible to observe those conditions with any precision. I also don't want to blindly match against any `IOException` because that may catch non-retryable IO errors like connect exceptions, unresolved hosts, etc. We may be able to enumerate those out but it seems risky

So, we have to resort to match against exception messages instead. Unsurprisingly, there is precedent for doing this:
1. https://github.com/AsyncHttpClient/async-http-client/blob/master/client/src/main/java/org/asynchttpclient/netty/future/StackTraceInspector.java
2. https://github.com/reactor/reactor-netty/blob/058159592584fcc3a66c32b48a5bfe6a8fe09297/reactor-netty-core/src/main/java/reactor/netty/channel/AbortedException.java#L46

I'm pretty sure Okhttp is doing something similar but not sure where to look in their codebase. I'm also not sure how much of these condition matches we should port over, but I think it's a good start and we can expand as user reports come in.

Alternatives: some HTTP clients try reading before writing to a socket to see if the connection has closed already. This lowers the probability of getting a reset/broken pipe, but it's also not bulletproof because network shenanigans.

### DIFF
--- a/ember-client/src/main/scala/org/http4s/ember/client/internal/ClientHelpers.scala
+++ b/ember-client/src/main/scala/org/http4s/ember/client/internal/ClientHelpers.scala
@@ -219,7 +219,7 @@ private[client] object ClientHelpers {
       result match {
         case Right(_) => false
         case Left(_: ClosedChannelException) => true
-        case Left(ex: IOException) => 
+        case Left(ex: IOException) =>
           val msg = ex.getMessage()
           msg == "Connection reset by peer" || msg == "Broken pipe"
         case _ => false
--- a/ember-client/src/main/scala/org/http4s/ember/client/internal/ClientHelpers.scala
+++ b/ember-client/src/main/scala/org/http4s/ember/client/internal/ClientHelpers.scala
@@ -39,6 +39,7 @@ import javax.net.ssl.SNIHostName
 import org.http4s.headers.{Connection, Date, `Idempotency-Key`, `User-Agent`}
 import _root_.org.http4s.ember.core.Util._
 import java.nio.channels.ClosedChannelException
+import java.io.IOException
 
 private[client] object ClientHelpers {
 
@@ -212,12 +213,15 @@ private[client] object ClientHelpers {
         req: Request[F],
         result: Either[Throwable, Response[F]]): Boolean =
       (req.method.isIdempotent || req.headers.get[`Idempotency-Key`].isDefined) &&
-        isEmptyStreamError(result)
+        isRetryableError(result)
 
-    def isEmptyStreamError[F[_]](result: Either[Throwable, Response[F]]): Boolean =
+    def isRetryableError[F[_]](result: Either[Throwable, Response[F]]): Boolean =
       result match {
         case Right(_) => false
         case Left(_: ClosedChannelException) => true
+        case Left(ex: IOException) => 
+          val msg = ex.getMessage()
+          msg == "Connection reset by peer" || msg == "Broken pipe"
         case _ => false
       }
   }


RaasAhsan:
Also, I've figured out how I want to write tests for these things so I'll open up another follow up later

RaasAhsan:
Screen Shot 2021-09-29 at 11 02 32 PM
Desired behavior

ember-client/src/main/scala/org/http4s/ember/client/internal/ClientHelpers.scala
```diff
@@ -212,12 +213,15 @@ private[client] object ClientHelpers {
...
+        isRetryableError(result)
 
-    def isEmptyStreamError[F[_]](result: Either[Throwable, Response[F]]): Boolean =
+    def isRetryableError[F[_]](result: Either[Throwable, Response[F]]): Boolean =
       result match {
         case Right(_) => false
         case Left(_: ClosedChannelException) => true
+        case Left(ex: IOException) =>
+          val msg = ex.getMessage()
+          msg == "Connection reset by peer" || msg == "Broken pipe"
```
armanbilge, line 224: Are there specific exception classes we can match for, instead of relying on the messages? We've ported many of the exceptions to JS so that the semantics match, but at least right now the messages don't.

https://github.com/typelevel/fs2/blob/main/io/js/src/main/scala/fs2/io/net/NetException.scala

armanbilge: Oh, sorry, let me actually read what you wrote above 😅 apologies.

armanbilge: Hmm, frustrating. Maybe we should at least _try_ to check for those specific exception classes? And leave this clause as a final catch-all?

RaasAhsan: not in the JDK. classes like `IOException` and `SocketException` are effectively catch-alls for a large space of IO errors across various platforms, and we're only interested in a subset of those

for example, on my mac I get `java.io.IOException: Connection reset by peer`, but i've seen `SocketException` with the same message on some Linux machines

RaasAhsan: > Maybe we should at least try to check for those specific exception classes?

can you elaborate which exceptions would we be looking for ?

armanbilge: > can you elaborate which exceptions we would be looking for ?

My bad, you're totally right, there don't seem to be specific exception classes for these. The place to fix this is upstream in fs2.io.js and it's going to be a PITA.

rossabaker (approved):
It's even worse than String parsing.  It's been pointed out on Blaze by people with non-English locales that some of these messages get localized.  What's retriable in English might not be in Russian.  😩

I think the best we can do is the best out-of-box experience and configurability.

armanbilge (approved):
I'm making the appropriate changes in https://github.com/typelevel/fs2/pull/2655 to hopefully support this on ember.js as well.

PR #5291, opened 2021-09-30
0.21 changes to improve forward source compatibility

bplommer:


bplommer:
The failure seems unrelated

build.sbt
```diff
@@ -86,6 +86,7 @@ lazy val core = libraryProject("core")
     ),
     buildInfoPackage := organization.value,
     libraryDependencies ++= Seq(
+      caseInsensitive % Optional,
```
rossabaker, line 89: Optional dependencies usually end in regret for me.  I'm not sure this one is worth it.

bplommer:
@rossabaker I thought I'd left a comment on that but it seems to be in a "pending review" that I can't publish - I also can't reply to your comment for some reason.

What do you think about just making it a regular dependency?

build.sbt
bplommer: Not worth making it optional or not worth having the dependency at all?

rossabaker: If it's optional, we could accidentally start using it in our build, and leave linker errors.  I think optional is definitely not what we want.

What is the precise goal on the CI shims?  Users still on 0.21 can call `Header.Raw(ci"X-Foo", "bar")` and have it compile?  Some of those calls are changing in 0.22 anyway.  This is mostly valuable for `CIString` signatures that are source compatible in 0.22, and only on the construction side, as the read side will still come out as `CaseInsensitiveString`.

I'm more convinced the blaze forwarders are worthwhile.

bplommer: I'm trying to come up with a strong case for the CIString shim but it mostly comes down to an incoherent sense of frustration that the codebase I'm currently working on isn't straightforward to update from 0.21.

I'll take that bit out of the PR.

rossabaker: I'm trying to find time for a scalafix from 0.21 to 0.22.  0.21 to 0.23 is high on our agenda at $WORK, and we're at a scale that a scalafix would be a big win.  Maybe that would alleviate some frustration.

bplommer: I doubt the scalafix will help much until I eliminate the misguided attempts this code makes to abstract over http libraries, and once I do so the migration should be straightforward...

PR #5294, opened 2021-09-30
Add DNT header model

coltfred:
This adds the header model for the DNT header which was requested as part of #5010.


tests/src/test/scala/org/http4s/headers/DNTSuite.scala
```diff
@@ -0,0 +1,40 @@
...
+
+package org.http4s
+package headers
+
+import org.scalacheck.Gen
+import org.scalacheck.Arbitrary
+import cats.syntax.either._
+
+class DNTSuite extends HeaderLaws {
+  implicit val dntGen = Gen.option(Gen.oneOf(true, false)).map(DNT(_))
```
rossabaker: Generators usually aren't implicit.

tests/src/test/scala/org/http4s/headers/DNTSuite.scala
```diff
@@ -0,0 +1,40 @@
...
+  implicit val arbDnt = Arbitrary[DNT](dntGen)
```
rossabaker: We have an arbitraries object that includes some header arbitraries.  This (and the gen) might find a better home there.  If you put it in the `arbitraries` object instead of one of the traits, you'd avoid a bincompat mess vs. putting them in the (regretted) `ArbitraryInstances` trait.

rossabaker: Build is failing because Scala 3 wants a type on the val.

core/src/main/scala/org/http4s/headers/DNT.scala
```diff
@@ -0,0 +1,57 @@
...
+
+  implicit val headerInstance: Header[DNT, Header.Single] =
+    Header.create(
+      ci"DNT",
+      _.renderString,
+      parse
+    )
+}
+
+final case class DNT(value: Option[Boolean]) {
```
rossabaker, line 51: `false` on a heaer name that's already negative is making my head hurt.  How about a sum type?

```
sealed abstract class DNT(value: String) extends Product with Serializable

object DNT {
  case object AllowTracking("0") extends DNT
  case object DisallowTracking("1") extends DNT
  case object NoPreference("null") extends DNT
}
```

core/src/main/scala/org/http4s/headers/DNT.scala
```diff
@@ -0,0 +1,57 @@
...
+    val falseParser = char('0').as(Some(false))
+    val trueParser = char('1').as(Some(true))
+    (falseParser | trueParser | nullParser)
+  }
+
+  def parse(s: String): ParseResult[DNT] =
+    ParseResult.fromParser(parser, "Invalid DNT header")(s)
+
+  /*
+   * `DNT = 0 | 1 | null`
```
rossabaker, line 38: Where do you find `null`?  I don't see it in [the spec](https://www.w3.org/TR/tracking-dnt/#dnt-header-field).  Is that a real-world thing?

rossabaker (commented):
Thanks!

core/src/main/scala/org/http4s/headers/DNT.scala
coltfred: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/DNT#directives listed it as a value. I felt like being willing to parse it was nice to have.

core/src/main/scala/org/http4s/headers/DNT.scala
coltfred: Adjusted to use an ADT.

coltfred (commented):
@rossabaker I've addressed all your feedback I think. LMK if something seems wrong or if you disagree about parsing the null value.

tests/src/test/scala/org/http4s/headers/DNTSuite.scala
coltfred: Ya, I added the type.

rossabaker (approved):
:+1: on green

rossabaker:
Parser and the constants are flipped.  I'm not sure which is right without studying MDN again.  Which is why I didn't want booleans! 😆

coltfred:
Don't worry, I was watching. They should be fixed now. I'll verify it goes green. :laughing:

rossabaker:
Ah, we were overdue for one of these damned flaky tests.  Ignore the blaze one.

PR #5295, opened 2021-09-30
Non-trivial merge to 0.23

rossabaker:
It's mostly trivial.  Mostly dealing with the fallout of the breaking change in munit-cats-effect, and how blaze hooks into it to check thread pool shifting.


armanbilge:
Jetty needs some adjustments as well.

armanbilge (approved):

PR #5296, opened 2021-09-30
Add accept-post header model

coltfred:
Addresses part of #5010

### DIFF
--- a/core/src/main/scala/org/http4s/headers/Accept-Post.scala
+++ b/core/src/main/scala/org/http4s/headers/Accept-Post.scala
@@ -17,21 +17,19 @@
 package org.http4s
 package headers
 
-import cats.data.NonEmptyList
-import cats.parse.Parser
-import org.http4s.internal.parsing.Rfc7230.headerRep1
+import cats.parse.Parser0
+import org.http4s.internal.parsing.Rfc7230
 import org.typelevel.ci._
 //Accept-Post response header.
 //See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Post
 object `Accept-Post` {
-  def apply(head: MediaType, tail: MediaType*): `Accept-Post` = apply(
-    NonEmptyList(head, tail.toList))
+  def apply(values: MediaType*): `Accept-Post` = apply(values.toList)
 
   def parse(s: String): ParseResult[`Accept-Post`] =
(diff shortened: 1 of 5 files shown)


rossabaker (changes requested):
Looks good after the list tweak.

core/src/main/scala/org/http4s/headers/Accept-Post.scala
```diff
@@ -0,0 +1,47 @@
...
+    headerRep1(MediaType.parser).map(xs => `Accept-Post`(xs.head, xs.tail: _*))
+
+  implicit val headerInstance: Header[`Accept-Post`, Header.Recurring] =
+    Header.createRendered(
+      ci"Accept-Post",
+      _.values,
+      parse
+    )
+
+  implicit val headerSemigroupInstance: cats.Semigroup[`Accept-Post`] =
```
rossabaker, line 43: Well, heck, you probably can even have a Monoid after the other comment!

coltfred: Can I assume there should be a comment about it maybe being able to just be a List inside?

I made it a Nel to match[ the Accept header](https://github.com/http4s/http4s/blob/series/0.22/core/src/main/scala/org/http4s/headers/Accept.scala#L81)

rossabaker: The `Accept` header is wrong, too.  I think @bplommer fixed a couple of the CORS headers that made this mistake on 1.0, and you might find precedent there.

coltfred: @rossabaker I ported the changes made by @bplommer to the series/0.22. I tried to port it into the exact lines to make potential merges less painful. Let me know if there's any other changes you'd like to see.

core/src/main/scala/org/http4s/internal/parsing/Rfc7230.scala
```diff
@@ -77,6 +77,9 @@ private[http4s] object Rfc7230 {
     between(char('('), cText.orElse(quotedPair).orElse(comment).rep0.string, char(')'))
   }
 
+  def headerRep[A](element: Parser[A]): Parser0[List[A]] =
```
rossabaker, line 80: I am surprised this didn't exist already, but I can't find it elsewhere.

rossabaker (approved):

PR #5297, opened 2021-09-30
Improve Arbitrary[HttpVersion]

rossabaker:
Our arbitraries aren't as good as they ought to be.  I'll be gradually reforming them as we pass through the model.

### DIFF
--- a/laws/src/main/scala/org/http4s/laws/discipline/arbitrary.scala
+++ b/laws/src/main/scala/org/http4s/laws/discipline/arbitrary.scala
@@ -203,13 +203,14 @@ private[discipline] trait ArbitraryInstances { this: ArbitraryInstancesBinCompat
       } yield Query(vs: _*)
     }
 
-  implicit val http4sTestingArbitraryForHttpVersion: Arbitrary[HttpVersion] =
-    Arbitrary {
-      for {
-        major <- choose(0, 9)
-        minor <- choose(0, 9)
-      } yield HttpVersion.fromVersion(major, minor).yolo
-    }
+  implicit val http4sTestingArbitraryForHttpVersion: Arbitrary[HttpVersion] = {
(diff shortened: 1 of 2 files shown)


laws/src/main/scala/org/http4s/laws/discipline/arbitrary.scala
```diff
@@ -203,13 +203,14 @@ private[discipline] trait ArbitraryInstances { this: ArbitraryInstancesBinCompat
...
 
-  implicit val http4sTestingArbitraryForHttpVersion: Arbitrary[HttpVersion] =
-    Arbitrary {
-      for {
-        major <- choose(0, 9)
-        minor <- choose(0, 9)
-      } yield HttpVersion.fromVersion(major, minor).yolo
-    }
+  implicit val http4sTestingArbitraryForHttpVersion: Arbitrary[HttpVersion] = {
+    val genSpecified = Gen.oneOf(HttpVersion.specified)
```
rossabaker, line 207: Focus on creating real ones.  Most of the 100 will never be seen in the real world.

laws/src/main/scala/org/http4s/laws/discipline/arbitrary.scala
```diff
@@ -203,13 +203,14 @@ private[discipline] trait ArbitraryInstances { this: ArbitraryInstancesBinCompat
...
+    val genAll = for {
+      major <- Gen.chooseNum(0, 9)
+      minor <- Gen.chooseNum(0, 9)
```
rossabaker, line 210: `chooseNum` vs. `choose` focuses on the boundaries.

rossabaker:
This DecodeSpec is getting to be a bother.

armanbilge (approved):

PR #5298, opened 2021-09-30
Cross-build 0.23 for Scala.js (rebased)

armanbilge:
Did some rebasing of #5285 to try and better preserve history. Opening a separate PR instead of force-pushing in case of screwups.


core/shared/src/main/scala/org/http4s/util/UrlCoding.scala
```diff
@@ -0,0 +1,37 @@
+/*
```
armanbilge, line 1: Darn, somehow git still couldn't track this file.

client/shared/src/test/scala/org/http4s/client/ClientRouteTestBattery.scala
```diff
@@ -19,10 +19,7 @@ package client
```
armanbilge, line 19: Did work for `ClientRouteTestBattery` 👍

armanbilge:
@rossabaker ready when you are!

core/shared/src/main/scala/org/http4s/util/UrlCoding.scala
rossabaker: That's fine.

PR #5300, opened 2021-09-30
Update build project to use http4s 0.23

armanbilge:
This should unblock @jenshalm's plan to revamp the docs in https://github.com/http4s/http4s/discussions/5192#discussioncomment-1410456.

BTW I ran this locally and seemed to work fine 👍


rossabaker:
Weird 0.22 depend on 0.23, but there's also absolutely nothing wrong with it.  It's just a client on a different classpath!

PR #5303, opened 2021-10-01
_Highly_ non-trivial merge to main

armanbilge:
Will supersede #5282. Pending #5298.

This is already a horrendous diff (and will continue to be). But, watch the diff against series/0.23 vanish before your eyes!


rossabaker:
Should I hold off further merges from 0.22 to 0.23 until this is done?

armanbilge:
No, please go ahead :) thanks for asking!

armanbilge:
I re-copied many files straight out of 0.23. My biggest concern for this merge is that I overwrote some 1.0-only changes.

Looking through `git log ^series/0.23 main --grep='#' --format='oneline'` gives me the following list to double-check:

- [ ] #4998
- [ ] #5051
- [ ] #5063
- [ ] #5082
- [ ] #5151
- [ ] #5287

armanbilge:
Also I can't replicate the failure on my local (I cleaned and compiled everything on 2.12 before pushing). Is CI messing with me??

armanbilge:
Ok, this found its way to a good place. There's still a lot of cleanups that can be applied (removing shims, removing deprecatia, diff minimizing, further deplatforming of server module) but I think we shouldn't block on those. I'll fix these in followup PRs, some of which I might run up through 0.23 now that I can see the endgame in 1.0 better.

As for how to review this ... the strategy I used was, instead of focusing on the horrendous diff here, to look through the diff between this and series/0.23. For all the modules (except server and ember-server), the only differences you see should be the 1.0-only changes made so far. The diff for server and ember-server is still a bit messy at the moment, since they are crossed only in main.

armanbilge:
Also FYI, I'm hoping we can get a release of https://github.com/typelevel/fs2/pull/2655 before the next release train. See https://github.com/typelevel/fs2/pull/2655#issuecomment-932018375 for details.

rossabaker (approved):
Very nice work.  I added the latest series/0.23, because it needed to be done anyway, and it saved me from figuring out what your merge base was to examine the residual diff between 0.23 and 1.0.

PR #5304, opened 2021-10-01
Replace our decode with fs2's decodeWithCharset

rossabaker:


rossabaker:
I'm too tired to dive into charset esoterica for why this is failing.  I kept the old DecodeSpec just to look for changes in behavior.

rossabaker:
I think this one is because `fs2.text.utf8.decode` uses replacement characters, and  `fs2.text.decodeCWithGenericCharset`reports the errors.  Our unit tests have been testing our internal `decode`, which behaves like the latter, but our higher level (i.e. `Media`) optimizes to `fs2.text.utf8.decode`.  So we've been inconsistent on error handling between UTF-8 and all other charsets.  Oops.

tests/shared/src/test/scala/org/http4s/DecodeSpec.scala
```diff
@@ -130,7 +130,7 @@ class DecodeSpec extends Http4sSuite {
       // Found by scalachek
       val source = Stream(-36.toByte)
       val decoded =
-        source.through(decode[Fallible](Charset(JCharset.forName("x-ISCII91")))).compile.string
+        source.through(decodeWithCharset[Fallible](JCharset.forName("x-ISCII91"))).compile.string
```
rossabaker, line 133: This test is still failing.

rossabaker: This is a bug in FS2, I think.

rossabaker: https://github.com/typelevel/fs2/issues/2664

tests/shared/src/test/scala/org/http4s/DecodeSpec.scala
```diff
@@ -81,44 +81,24 @@ class DecodeSpec extends Http4sSuite {
...
-  test("decode should handle incomplete input") {
-    // Only the first byte of a two-byte UTF-8 sequence
-    val source = Stream(0xc2.toByte)
-    val decoded = source.through(decode[Fallible](Charset.`UTF-8`)).compile.string
-    assert(decoded match {
-      case Left(_: MalformedInputException) => true
-      case _ => false
-    })
-  }
-
```
armanbilge, line 115: I assume this is tested upstream?

armanbilge (approved):

tests/shared/src/test/scala/org/http4s/DecodeSpec.scala
rossabaker: We never called the deprecated `decode` publicly on UTF-8.  We optimize to `fs2.text.utf8.decode`. Unfortunately, that has different error semantics than our `decode` and the new FS2 `decodeWithCharset`.  That's another one of those semantic changes we need to wrestle with.

rossabaker: I'll open an issue in FS2, because our old inconsistent behavior is now also their inconsistent behavior.

rossabaker:
This might be hung on Java 8.  I don't think it's critical to the release train.

PR #5313, opened 2021-10-02
[WIP] Migrate Website to Laika

jenshalm:
This is not ready to be merged yet, but indented to be a better place to continue some of the discussion in #5192 while also already having some code and config to look at.

I'll break down the steps that are done or still need to be done below with relevant details and some remaining questions.
Tasks already committed to this PR are checked. I'll stop the work for now, and wait for answers to the open questions.

I'm dividing tasks into mandatory steps which are essentially mostly just ensuring you do not have less functionality than before by switching to Laika. This also includes a few optional improvements I already included as they were a quick and easy win.
The optional steps can then be done either as part of this PR, or split into small follow-up PRs or not done at all, depending on your preferences.

Mandatory Steps
---------------

- [x] Basic build setup. I am currently cramming everything into the main `build.sbt`, but no worries, I'll clean this up as the last step. 
I assume a lot of this stuff can move into the http4s sbt plugin, but testing is quicker for now with everything in one place.

- [x] Change Markdown sources to define titles as Level-1 headers instead of frontmatter properties. 
Only expceptions are a few cases where you seem to prefer a shorter headline for navigation which can be set in an explicit configuration header 

- [x] Configure syntax highlighting. These are the built-in highlighters for now. You can get back to external JS-based highlighters if you want to, but if you browse the results you might agree that they are good enough for this purpose. Styling is discussed in the optional steps below.

- [x] Configure anchors for section links. This was one of the features requested in the original ticket, and is only a matter of configuration.
Note that the anchors are to the right of the headlines for now. They can be configured to be on the left, too, 
but there seems to be a regression, they do not seem to be placed correctly anymore when on the left, so I need to have a look first.

- [x] Configure the base URL for links to Markdown sources. They appear at the bottom of the top-right page navigation box.

- [x] Internal Link Validation. This is on for all internal links by default, meaning you get a failing build in case of dead internal links, which I assume is what you want. If you add internal links to sources which are not available at build time, you need to specifically disable validation for those.
Note that currently internal links are not yet actually validated as Laika does not understand the Pretty URL feature from Hugo you are using (see discussion about this topic below).

- [x] Use yellow Laika callouts for warnings. Seems to be a good fit, instead of a regular headline. Example: CORS page on the top.

- [x] Variable substitutions (working for the `docs` project, not yet updated in the `website` project).
I ran into a little problem as Laika does not support variable substitions in code blocks at the moment (longer story).
But fortunately I was able to support your specific use case (you only use them in string literals) with a little extension of just 6 lines.
This workaround might become obsolete at the time 0.19 arrives, but that won't happen this year.
Otherwise, the build setup is simplified now, as with Laika you do not have to write those variables to a file. 
It's now a simple Scala Map exposed as an sbt setting that can be referenced from anywhere in markup or template files.

- [ ] Versioning. Didn't look into this yet, as I have many questions (see below).

- [ ] Configure top navigation bar. This is still empty at the moment, apart from a home link that 404s for now. 
I added questions about your preferences below.

- [ ] CI integration / publishing. Apply changes in `docs` project to the `website` project or merge them. Some remaining questions, see below...

- [ ] Build cleanup. Move most of the Laika configuration to the http4s plugin. Remove Hugo setup.

Optional steps
--------------

1. Bring back font and color choices of the original site
This is quite easy to do if you prefer that. This could include both, the theme colors for the site as well as the syntax highlighting.

2. Support dark mode
You can configure a second set of colors for the theme and for highlighting so that the site adapts to the user's OS/browser preferences.

3. Code formatting
Many code samples do not render well, at least not with Laika's default theme styling, as some lines are quite long, causing them to wrap. In many cases there are good opportunities for adding line breaks, I'm happy to go through them and adjust them where needed.

4. Hierarchical navigation
With Laika you do not have to stick to a flat list of links for the left navigation. Having sections might make it easier to browse. See Laika's documentation for how the default theme renders a two-layer navigation structure.

5. Favicons
You currently do not seem to have any. Happy to produce something based off your logo.

6. EPUB or PDF downloads of the site
If you'd want these, it's a mere configuration step, happy to produce a cover PNG in this case. 
EPUB would not add a noticeable amount of time to your build, but PDF would, as it involves a full layout engine. 
PDF would probably get this build step to a similar duration as the mdoc step, whereas site and EPUB rendering is really fast.

Questions for Maintainers
-------------------------

Apologies, these are quite a few. I examined the build, the Markdown sources, the templates and the Hugo configuration and short-codes, and came up with all these open questions.

1. **General Sanity Check**: Do you spot anything in my lists above that raises concerns? Am I missing any critical things that need to be done?

2. **Optional Features**: Could you briefly indicate (for the optional features listed above) whether you want them A) as part of this PR or B) as part of a later PR or C) not at all? 

3. **Versioning**: I would recommend to move to Laika's built-in support for versioning, as this would give you a smarter version drop-down that stays on the same page when switching and that always sees newer versions without re-rendering the old version.
A bunch of questions about your requirements:
a) Do you currently auto-deploy all documentation changes in the `docs` project on commits in any of the active branches?
b) Do you also auto-deploy changes to the `website` project? If yes, do you only do that from the `main` branch?
c) How many versions do you want to continue to host? I remember I saw a discussion about reducing the number of EOL versions.
d) How many versions do you want to re-render with Laika? I'm happy to do the work for a number of them.

4. **Pretty URLs**: At the moment all pages are moved into a separate sub-directory during transformation. 
As discussed in #5192, this is currently not supported by Laika, as it simply had never been requested as a feature in its 9 years of existence and is also not something that can be added easily.
I was suprised to find out that it is actually the default behaviour in Hugo (https://gohugo.io/content-management/urls/#pretty-urls).
The options for the migration are:
a) Move the Markdown sources into a matching structure of sub-directories for now. This is somewhat inconvenient for maintainers, but otherwise the simplest and most robust solution.
b) Abandon Pretty URLs. This would require setting up quite a few redirects as users would run into a lot of 404s otherwise.
c) Build a Laika extension. This is entirely possible and not too hard to do, but probably an unattractive option as it would first delay this PR for a several days at least and also mean you'd have that extension and its tests in your code base for months until Laika supports it out of the box.
If you want to see this feature in Laika, I'd like to encourage you to create an issue for it. It's much more likely to be part of 0.19 when there is an explicit user request.

5. **Automatic Triggering of `mdoc`**: Currently the `makeSite` task is defined to depend on the `mdoc` task. Do you want to keep that, so that whenever you run `laikaSite` the `mdoc` job runs first?
Note that when using Laika's built-in preview server with live-updates, you might still need to run `mdoc --watch` manually, as I don't think you can wire two long-running tasks together.

6. **Separation of `docs` and `website` projects**: Would this still be desired or necessary with Laika? I assume this separation might exist as you were using tooling that does not natively support versioning?
In any case, if you do not have any other requirements for the separation, Laika can simply distinguish between versioned and unversioned pages in a single transformation based on configuration.

7. **`sbt-site` plugin**: I assume you used this solely for triggering the Hugo job and for connecting the site generator with the ghpages plugin? In this case this dependency would no longer be necessary, as you can directly set this up with Laika.

8. **Accessibility**: Laika currently does not render ARIA tags. Can you briefly live without them? 
I'm happy to add them and this won't take as long as some of the other feature additions mentioned, as I can do this in an 0.18 patch release (most likely the one after the next).

9. **Responsiveness**: Laika's default theme is designed to be responsive, but portrait mode on phones still requires some polishing, which is one of the remaining tasks for 0.18.1 (https://github.com/planet42/Laika/issues/208). Currently the top navigation items overlap. 
From landscape mode on phones upwards it already looks decent, so it would just be a very brief UI issue for probably just a small number of users suffering by reading docs on phones in portrait mode. Would this be acceptable?

10. **Projects Menu in Top Navigation Bar**: This is a bit of an outlier and does not fit well into Laika's default theme. It is possible to keep it where it is, but this would most likely mean manual maintenance of HTML and CSS for the top bar as the placement of that menu might not work well with the default theme's responsiveness.
Could this move somewhere else? It could, for example, be a section on the left navigation with a `Related Projects` section header. It would fit particularly well when you move to 2-layer navigation anyway.

11. **Links in Top Navigation Bar**: You are currently linking to GitHub, Gitter and Twitter.
a) Do you want to keep these?
b) Don't you want to link to Discord instead of Gitter now?
c) Would you want to add a link to API docs?

12. **JavaScript**: `bootstrap.js` and `highlight.js` would no longer be necessary. However, there is also a `popper.js`. I was unable to figure out what this is used for, is this some functionality you need to retain?

13. **Content**: The `docs` project contains a page `standards.md` in the `src/hugo/content` folder.
I could not determine where this is used. Its content seems to be duplicated as part of the Contributing page in the `website` project.
Is this some old, unused artifact?


rossabaker:
Wow, this is exciting.  I'm sleepy and owe you better answers to some of these, but let me try to unblock:

> General Sanity Check: Do you spot anything in my lists above that raises concerns? Am I missing any critical things that need to be done?

Looks good.  I would like to make sure we can get off Netlify as part of this.  We moved when it was the only way to get https on a custom domain, and its redirects are far superior (jdk-http-client is using them).  But GitHub has partially closed the feature gap, and our Netlify admin bus factor is unacceptable.

> Optional Features: Could you briefly indicate (for the optional features listed above) whether you want them A) as part of this PR or B) as part of a later PR or C) not at all?

Incremental PRs are generally easier, so probably separate PRs, but whatever is less work for you.

> Bring back font and color choices of the original site

I don't care about the font.  The shades of orange are unofficially part of the branding, and I'd give that priority on appearance tweaks.  (My battery is running low and I have not previewed what you've already done, but I'll fix that tomorrow!)

> Support dark mode

Low priority for me.

> Code formatting

If the Laika is exposing this, maybe this makes sense in this PR.

> Hierarchical navigation

This is probably a conversation to include @noelwelsh in as the content gets refurbished, but it does sound nice.

> Favicons

A naive scaledown of the logo looked horrible, but someone with Inkscape/Illustrator and talent could easily do better than me.

> Do you currently auto-deploy all documentation changes in the docs project on commits in any of the active branches?

We used to on the series/* branches and main, but I think somewhere along the way it became only on releases.  There are pros and cons, and if there's one Laika likes better, I'm good with that.

> Do you also auto-deploy changes to the website project? If yes, do you only do that from the main branch?

Only from `main`.  I'd like it to be autodeployed.  It once was.

> How many versions do you want to continue to host? I remember I saw a discussion about reducing the number of EOL versions.

0.22, 0.23, 1.0 are actively maintained.  0.21 still has the most users, but only if it's not hard.  Before that can disappear unless rendering what already exists in Laika is basically free.

> How many versions do you want to re-render with Laika? I'm happy to do the work for a number of them.

Not sure how this question differs from the above?

> Pretty URLs

Are we talking `/foo/` vs. `/foo.html`?  My gut feeling is the subdirectory solution is the best, but if we're rearranging pages as the content changes, we'll be breaking a lot of links anyway, and I'd rather not fight the natural Laika way.  It's only the way it is now because it's the natural Hugo way.

> Automatic Triggering of mdoc

What experience do you like best on Laika sites?  I'm pretty open on this.

> Separation of docs and website projects

The separation is because some things pertain to the project as a whole (e.g., version matrix, community info), and some refer to a specific version of http4s.  There's no 0.22 community vs. 0.23 community, so I think the distinction makes sense.

Another thing we should consider here is that we are planning to split this repo and publish modules more independently.  With an eye toward that, "website" being the general description of http4s.org as a hole, and versioned docs for all the projects makes sense.  And so that "website" might actually belong in its own repo.  And then whether it's all tied together with subdomains, or a single gh-pages repo, or a single gh-pages repo with submodule references to individual projects (I think cats-effect does this?) is up in the air.

> sbt-site plugin

That sounds right.  I think it can go.

> Accessibility

Whatever we have with ARIA, we got for free from Hugo.  I'm glad we did, but I think we can live temporarily without.  I honestly don't know how good what we have is.

> Responsiveness

I don't have analytics, but I care most about desktop.  I expect our users skew toward a browser next to their code editor.  Mobile matters, but a temporary regression is fine.

> Projects Menu in Top Navigation Bar

I am fine moving anything around spatially.  Doing things The Laika Way is much more important than doing things The Crufty Http4s Way.

 > popper.js

I don't remember this at all. :shrug:

> standards.md

Yes, that's far out of date and can go.

jenshalm:
Great, thanks for taking the time to go through my questions, I have enough additional info now to continue with most of the work. 

Based on your answers I'll suggest to only add three things to the list of mandatory steps: adjusting the theme colours, improving code formatting and having a conversation about potential changes to navigation structure with those planning to work on content (e.g. Noel), because if we decide to break existing URLs we probably don't want to do this twice. 

I'll do everything else that is just nice-to-have in smaller follow-up PRs.

Below I'll only comment on the answers that are still kind of unresolved.

> Looks good. I would like to make sure we can get off Netlify as part of this. We moved when it was the only way to get https on a custom domain, and its redirects are far superior (jdk-http-client is using them). But GitHub has partially closed the feature gap, and our Netlify admin bus factor is unacceptable.

Happy to look into this, but I have zero knowledge of Netlify and how you use it. Regarding custom domains I had a quick look and GitHub only seems to require the presence of a `CNAME` file which you already have. However, you also have the `_redirects` file for `stable/` and `latest/` URLs, and I don't think GitHub would support this?
 
> > Do you currently auto-deploy all documentation changes in the docs project on commits in any of the active branches?
> 
> We used to on the series/* branches and main, but I think somewhere along the way it became only on releases. There are pros and cons, and if there's one Laika likes better, I'm good with that.

That's really just a matter of preference, Laika will work with both modes just fine. I only asked this to have a more complete picture on how you are rolling at the moment.

> > Do you also auto-deploy changes to the website project? If yes, do you only do that from the main branch?
> 
> Only from `main`. I'd like it to be autodeployed. It once was.

Ok, shouldn't be a problem to go back to auto-deploy.

> > How many versions do you want to continue to host? I remember I saw a discussion about reducing the number of EOL versions.
> 
> 0.22, 0.23, 1.0 are actively maintained. 0.21 still has the most users, but only if it's not hard. Before that can disappear unless rendering what already exists in Laika is basically free.

Ok, then I'd go through all branches up to and including 0.21. I guess this means a separate PR for each of those.

> > How many versions do you want to re-render with Laika? I'm happy to do the work for a number of them.
> 
> Not sure how this question differs from the above?

You could, in theory, continue to host older versions without re-rendering them with Laika. You can still configure the version drop-down to know about those older non-Laika versions. But it would potentially be confusing for users to see two different designs and messier in the deployment as you'd have the Laika theme and the old Bootstrap styling flying around in the same setup, which is probably not desirable. If you can live with no longer hosting anything pre-0.21 then it's certainly the cleanest solution to re-render everything that remains with Laika, but it requires explicit steps for each branch. 

> > Pretty URLs
> 
> Are we talking `/foo/` vs. `/foo.html`? My gut feeling is the subdirectory solution is the best, but if we're rearranging pages as the content changes, we'll be breaking a lot of links anyway, and I'd rather not fight the natural Laika way. It's only the way it is now because it's the natural Hugo way.

I'll discuss this with Noel first and see what his plans are. 
 
> > Automatic Triggering of mdoc
> 
> What experience do you like best on Laika sites? I'm pretty open on this.

Also just a matter of preference. The linking of the tasks might be potentially confusing for maintainers as it would only work for the `laikaSite` task, not for the long-running `laikaPreview` task that fires up a server with live-updates. I think I won't wire them in this PR then, it's always just a one-line change to add it back.

> > Separation of docs and website projects
> 
> The separation is because some things pertain to the project as a whole (e.g., version matrix, community info), and some refer to a specific version of http4s. There's no 0.22 community vs. 0.23 community, so I think the distinction makes sense.
> 
> Another thing we should consider here is that we are planning to split this repo and publish modules more independently. With an eye toward that, "website" being the general description of http4s.org as a hole, and versioned docs for all the projects makes sense. And so that "website" might actually belong in its own repo. And then whether it's all tied together with subdomains, or a single gh-pages repo, or a single gh-pages repo with submodule references to individual projects (I think cats-effect does this?) is up in the air.

When you start to split things into separate repos I think it would be beneficial if the various sub-sites would be more autonomous, too. Currently this is not really the case, as the website pages, while being unversioned themselves all contain the full `Documentation` dropdown with all the version info for the main http4s project. This is why, in the current setup and site structure, the split only gives you build and config duplication with no benefits. Both projects would also generate identical theme and config files for the site which might potentially be confusing for maintainers.

armanbilge:
> You could, in theory, continue to host older versions without re-rendering them with Laika.

Just a quick thought: could we just dump an archive of the site as-is now under a different URL? Idk what will happen to the internal links. But then on the new site we can just put a link to the archived old site for anyone who needs the old stuff.

An even lazier solution is just point to web.archive.org or something 😆

armanbilge:
Also, apologies for the conflicts, that was my merge into main. Basically, the docs have moved back up a directory out of the `jvm` folder. The good news is that the CE2 dependency is upgraded to CE3.

jenshalm:
> Also, apologies for the conflicts, that was my merge into main. Basically, the docs have moved back up a directory out of the `jvm` folder. The good news is that the CE2 dependency is upgraded to CE3.

Did you also change the content in some way or is it safe to just replace them with my current version in their new location?

armanbilge:
Nope, no changes to content. Just an `mv`.

jenshalm:
I rewrote the paragraph about the split of the `docs` and `website` project in my longer response above as I wasn't focusing on the info that really matters.

jenshalm:
Ok, I made some good progress, configured left + top navigation, including a manual entry for "Related Projects" in the left navigation pane, configured versioning, grey + orange theme colours and fixed some source code formatting.

I'll suspend work again now (only few steps left to do), until final decisions have been made regarding URL structure.

After some further thought I feel like it might be best to be 100% conservative with URLs in this PR even though it comes with a few downsides, which I'll briefly summarize again here. 

The advantage would be that this PR could be merged more quickly and that there is no disruption yet for users with old links getting 404s for now. The downsides are:

1. We need to fake the Pretty URLs as discussed above as Laika does not support this yet. For a few months you would have a sub-folder for each individual Markdown file. The left navigation panel might require manual setup, too, or a custom directive, as I don't feel the auto-generated navigation would work out of the box with this.

2. The project structure with two sbt projects writing to the same logical Laika site is a somewhat unusual way to use Laika. Long-term I would recommend to either merge the two projects or truly separate them. But with your idea to potentially split out modules and have separate repos with their own docs, you might want to defer the debate and decision until that point to ensure there will only be one potential point in the future where URL structures change.

What do you think? (@rossabaker, @noelwelsh).
It's also a good time to have a first look.

armanbilge:
For anyone who's curious, I tried generating the site for v1.0 and hosting on gh-pages  🤩  
https://armanbilge.github.io/http4s/v1.0/

noelwelsh:
Looks good to me!

Rewriting the documentation content is a very open ended task, ranging from minor tweaks to completely rewriting everything. It's hard to know how much time I'll have to work on this, and hence what point we'll arrive at. I agree it's best to keep the URL structure the same for now and at some point in the future we might end up breaking things.

rossabaker:
Sorry, I got distracted trying to start another release train:

> However, you also have the _redirects file for stable/ and latest/ URLs, and I don't think GitHub would support this?

That's correct.  We'll have to sacrifice that.

>> We used to on the series/* branches and main, but I think somewhere along the way it became only on releases. There are pros and cons, and if there's one Laika likes better, I'm good with that.

> That's really just a matter of preference, Laika will work with both modes just fine. I only asked this to have a more complete picture on how you are rolling at the moment.

It's nice to improve docs without a release, but the danger is when we document an enhancement that isn't released yet.  I could be persuaded either way...

> Ok, shouldn't be a problem to go back to auto-deploy [on website]

... and I think the case is very strong on website.  So for consistency, let's try it on both.

>  If you can live with no longer hosting anything pre-0.21 then it's certainly the cleanest solution to re-render everything that remains with Laika, but it requires explicit steps for each branch.

Yes, I can live with this.

> This would be different if the stuff from docs would be hosted under, say, /core/v0.23/ and not directly under /v0.23. In the latter case, for Laika these are "versioned documents belonging to /", and /contributing.html for example is an "unversioned document belonging to /" and they would share the config in /laika. In the former case, one level deeper, these could more easily be treated as independent sub-sites

I wouldn't think twice about breaking those core URIs if GitHub Pages had better rewrite support.  I suppose we could mitigate some of the damage with client-side meta-refreshes in the website project that would own the /v* namespace.

As for `/core`, are subdomains a good option?  We already have precedent with jdk-http-client.http4s.org, and a core.http4s.org wouldn't be out of the question.

armanbilge:
> It's nice to improve docs without a release, but the danger is when we document an enhancement that isn't released yet. I could be persuaded either way...

Daniel makes a strong case for this in https://github.com/typelevel/cats-effect/issues/523#issuecomment-913975573.

> In the former scenario, changes and fixes to documentation, even minor ones, take weeks or months to show up on the site. This can be quite discouraging to contributors, since their work is invisible for a long time. The advantage though is that documentation is never "ahead" of what has been released.
> 
> In the latter scenario, documentation changes make their way to the site very quickly, but you could end up with documentation which reflects something that hasn't yet been released. In practice, this hasn't really been a concern at any point (the one time it came up was with the Tracing documentation, but we solved it by just not adding it to the site index until the release).
> 
> I pretty strongly believe that the tradeoffs of continual site delivery are superior to those of on-release site updates. Documentation is just way too precious to couple its lifecycle to something as intermittent as the full binary releases.

rossabaker:
> I think you skipped question 11. :-)

Oops.

> a) Do you want to keep these?

Yes.

> b) Don't you want to link to Discord instead of Gitter now?

Yes.  Just don't expect to see me there.

c) Would you want to add a link to API docs?

Yes, definitely.  I'm starting to make a push on improving those.

rossabaker:
I'm liking it a lot!  A couple things I noticed from the preview:

1. Would be nice to get the logo back on there somewhere, but let's keep prioritizing functionality over themes.
2. The edit links don't work.

jenshalm:
> > This would be different if the stuff from docs would be hosted under, say, /core/v0.23/ and not directly under /v0.23. In the latter case, for Laika these are "versioned documents belonging to /", and /contributing.html for example is an "unversioned document belonging to /" and they would share the config in /laika. In the former case, one level deeper, these could more easily be treated as independent sub-sites
> 
> I wouldn't think twice about breaking those core URIs if GitHub Pages had better rewrite support. I suppose we could mitigate some of the damage with client-side meta-refreshes in the website project that would own the /v* namespace.
> 
> As for `/core`, are subdomains a good option? We already have precedent with jdk-http-client.http4s.org, and a core.http4s.org wouldn't be out of the question.

In light of this discussion and after I figured out how to mimic Hugo's Pretty-URL feature with just about 50 lines of extension code, I feel like my current recommendation would be this:

* Keep this PR minimally invasive, by not breaking any URLs yet. This way this PR may be ready to be merged in a few days and would already free you of all the tool integration issues you were facing. And despite of the somewhat idiosyncratic way of using Laika it will work just fine for a while. I just don't recommend to stick with the current combination of build- and URL-structure for too long as it theoretically opens the door for non-deterministic builds when the config for `docs` and `website` go out of sync for whatever reason and the two projects overwrite each other's styles. Blue site on Monday, green on Tuesday, blue again on Wednesday, would be fun. :-) Risk is low though as I centralized the config and have the two projects only reference it.

* Create a separate issue to continue the debate, specifically only focussing on project-, URL- and navigation-structure. I'm happy to create that after this is merged and summarize everything discussed so far. There are at least three areas where a decision might break existing URL patterns and you might want to bundle those into one breaking change to minimize disruption for users, and maybe take some time to make a decision. The three areas are:
 
1. Potential modularization of projects and their docs as well as disentangling umbrella site and http4s docs. You'd have lots of options here, sub-domains are probably ideal, but sub-folders would work, too, almost anything really apart from the current structure which turns the two projects into one logical site as soon as you switch on Laika's versioning support. And keeping it off is not something I'd recommend as the version menu was frankly one of the weak spots of the site, I found it was rarely complete and up-to-date on all branches, so having a tool manage this centrally is really beneficial.

2. Maybe consider breaking project navigation into several sections. I personally don't find the long, flat list easy to digest. I'm happy to do a rough proposal, but since I'm not overly familiar with the content, I feel like others might have better ideas. If you do that, the smoothest way would be to also physically organize input sources in that structure, as that would auto-generate the navigation. This would also break existing URLs then. You could, however, also keep them in the same place physically and manually configure the navigation if desired.
 
3. Decide on whether to keep the Pretty URL feature you used with Hugo. I don't have a strong recommendation either way, as it's mostly a matter of taste. But if you decide to keep this feature, I'd like to encourage you to do a feature request at the Laika project, as you probably don't want to maintain the extension code in your project indefinitely. It's also a feature that is more robust when implemented with internal APIs as opposed to extension APIs. If you decide to break URLs anyway either in 1. or 2. abandoning this feature at the same time would not have any impact on users.

As for this PR, I'm currently working on the website project and found that this requires more manual tweaking than the docs project as many pages contain verbatim HTML in the Markdown source which in turn rely on Bootstrap being around. I'd like to avoid that dependency as it's not needed by most pages, so I try to replicate the functionality with minimal manual CSS.

I might have a mergeable PR in a few days, maybe on the weekend.

Btw. sorry for the CI mess, I've never used the GitHub Actions plugin and need to figure out what it wants from me. :-)
I'm somewhat confused by the fact that similar actions appear to be declared in two places, in `build.sbt` and `Http4sPlugin` and I'm not sure whether both are active or whether one of them is dead code.

jenshalm:
> 1. Would be nice to get the logo back on there somewhere, but let's keep prioritizing functionality over themes.

The natural (intended) way to integrate a project logo would be to use it in place of the generic home icon and let if serve as the home link to. Squeezing it into the top bar in addition to the other elements would almost certainly mean you'd need to tweak and manually maintain the responsiveness of the top bar. I can do this as part of this PR if you want or defer this for a separate one. The logo would be quite small though. We could make the top bar larger, but I find it quite elegant when it's rather slim.

> 2. ~The edit links don't work.~ Oh, that's almost certainly just because it hasn't been merged yet.

Most likely because Arman moved the sources while I was working on it. :-) I just need to update the base URL.

armanbilge:
Yes, those are just flakes, don't pay them any mind!!

Thanks again for your incredible effort here 😁

jenshalm:
I added another commit and removed the WIP from the title.
Funny timing that the last commit is the first one to finally produce a green build... :-)

I rewrote most of the PR description to cover the work actually done and not a TODO list.
I only kept the list of questions to maintain context for the discussions.

I generally recommend to have a close look at all the things I mention, but in particular to one issue:
I am unable to test the new integration with the `ghpages` plugin locally as it performs an SSH-based checkout which fails without credentials. I recommend that either the auto-publishing gets temporarily disabled in this PR, or a maintainer tests this integration locally first, to avoid the risk of temporarily running a broken site.

I am offline now today, but can look at all the issues you may find from tomorrow.

PR #5316, opened 2021-10-03
0.23 JVMism cleanups

armanbilge:
Setting us up for a JVMism free 1.0, either directly or by deprecation.


core/shared/src/main/scala/org/http4s/Message.scala
```diff
@@ -408,7 +408,7 @@ final class Request[F[_]] private (
 
   def remoteHost(implicit F: Sync[F]): F[Option[Hostname]] = {
     val inetAddress = remote.map(_.host.toInetAddress)
-    F.delay(inetAddress.map(_.getHostName)).map(_.flatMap(Hostname.fromString))
+    F.blocking(inetAddress.map(_.getHostName)).map(_.flatMap(Hostname.fromString))
   }
```
armanbilge, line 412: 0.23 JS won't be getting this one since it needs an `Async`.

Actually, I will add this upstream to ip4s so in 1.0 we can just ask for an `Dns` capability here instead of `Sync`/`Async`.

https://github.com/Comcast/ip4s/blob/main/jvm/src/main/scala/com/comcast/ip4s/Dns.scala

tests/shared/src/test/scala/org/http4s/Ipv4AddressSuite.scala
```diff
@@ -36,7 +37,8 @@ class Ipv4AddressSuite extends Http4sSuite {
   if (Platform.isJvm)
     test("fromInet4Address should round trip with toInet4Address") {
       forAll { (ipv4: Ipv4Address) =>
-        assert(Ipv4Address.fromInet4Address(ipv4.toInet4Address) == ipv4)
+        assert(Ipv4Address.fromInet4Address(ipv4.toInet4Address) == ipv4): @nowarn(
+          "cat=deprecation")
```
armanbilge, line 41: Or else, we should just remove these tests. If we address the deprecation then we are just testing if ip4s works :)

armanbilge: Forgive me, I just deleted them after getting into a twist of `nowarn` and `nowarn` doesn't suppress warnings between platforms.

core/shared/src/main/scala/org/http4s/Message.scala
armanbilge: https://github.com/Comcast/ip4s/pull/325

rossabaker: I'm missing something.  How is JS not getting this?  This is in shared?

armanbilge: Scala.js is super weird :(

With Scala.js, you can pretty much _compile_ any code you want, even it's loaded with tons of JVM-only stuff. It's only when you actually try and generate a `.js` file (i.e., the _linking_ stage), that if you used one of these JVMisms it can't find a JavaScript implementation to link to and stops you in your tracks.

_However_, the first part of the linking stage is dead-code-elimination. So, if nobody uses this `remoteHost` method in their Scala.js app, then it will be dead-code-eliminated and the reference to the JVMism will disappear along with all the associated problems. In fact, the reason I didn't catch this earlier must be that our test suite never calls this method, and therefore it is always dead-code-eliminated.

This weirdness is exactly what allowed us to pull off the trick with the syntax shims: although they compile under Scala.js, as long as nobody uses them (or anything that references them) they get dead-code-eliminated before they cause problems.

So what I meant by JS "not getting this" is that because it is currently implemented using a JVMism, although this code will compile in our JS build, nobody will be able to link to it, and thus they won't "get it".

The polite thing to do here, would be to "platform"/split the file, and move this method into JVM-only sources. Because otherwise, some pour soul can accidentally call this method, compile their code (even ship a library!) but its only when somebody tries to _link_ it (presumably because they want to run it) will they find out they've been fooled all along.

IDK if that helps 😅

armanbilge: Oh also, I think I have an idea for a binary- and source-compatible way to adjust this method so that it _can_ be supported on JVM and JS, after my ip4s PR is released.

rossabaker: I guess I'm still having a hard time understanding why the Scala.js compiler sees all these signatures it won't be able to link.  If we can so easily add these dead code bombs by having unreferenced-on-Scala.js functionality, it sounds like code coverage becomes more important as a safety device.

armanbilge: > I guess I'm still having a hard time understanding why the Scala.js compiler sees all these signatures it won't be able to link.

I'll try and find a reference for this, but my understanding is that Scala.js compiles against `.class` files (such as the ones provided in the JDK) but links against `.sjsir` files. And these aren't necessarily consistent with each other (in the case of the standard Java lib: definitely not). I'm sure there are Reasons why it works this way but I don't know.

> it sounds like code coverage becomes more important as a safety device.

Yes. This. And not just in our test suite maybe ... I was only able to catch https://github.com/http4s/http4s/pull/5307 because I published locally and tried building http4s-dom against 0.23sjs. Somehow that issue was masked in our own test suite.

Tons of weird stuff can happen because of screw ups that fly under the radar. Like consider this PR which fixed Scala.js support in cats-effect-testing: https://github.com/typelevel/cats-effect-testing/pull/180. It behaved like it was working and published Scala.js artifacts and everything, although they were completely useless. I only uncovered the problem when I tried to use those artifacts downstream and even then it took me a while to realize what the underlying problem was: https://github.com/typelevel/cats-effect-testing/pull/176#issuecomment-887583137

rossabaker:
Do you want to try to squeeze the Dns in before merging this?  Do we need to wait on that to release 0.23?

armanbilge:
No, the `Dns` can come later. No need to block 0.23 on that.

armanbilge:
But I would still prefer to get https://github.com/typelevel/fs2/pull/2655 out first. What's the timeline for the release train?

core/shared/src/main/scala/org/http4s/Message.scala
armanbilge: Ok, I found the reference I was looking for:

> This will generate .sjsir and .class files for each class in your project (just like for Scala JVM). The .class files are used by the Scala.js compiler for symbol lookup during separate compilation.

https://www.scala-js.org/doc/project/building.html

rossabaker:
0.21 is ready now.  I'm gathering notes for 0.22.  Nothing says they need to go together, but it's frustrating when old versions get shiny things first.

armanbilge:
https://github.com/typelevel/fs2/pull/2655 is on the way.

async-http-client/src/main/scala/org/http4s/asynchttpclient/AsyncHttpClient.scala
```diff
@@ -116,7 +116,7 @@ object AsyncHttpClient {
         val eff = for {
           _ <- onStreamCalled.set(true)
 
-          subscriber <- StreamSubscriber[F, HttpResponseBodyPart](dispatcher)
+          subscriber <- StreamSubscriber[F, HttpResponseBodyPart]
```
armanbilge, line 119: This is a ride-along deprecation that appeared when I bumped fs2.

armanbilge:
I bumped fs2 to a hash so I could make the final changes.

armanbilge:
Seems there's something changed in fs2 since 3.1.3 that causes ember-server to fail tests. Ignoring this PR here, if on `series/0.23` I bump fs2 to `3.1-69-b2a90a1` basically every test fails with:
```
==> X org.http4s.ember.server.EmberServerSuite.server responds to requests  0.015s java.net.BindException: Address already in use
```

armanbilge:
Ok, I tracked it down to https://github.com/typelevel/fs2/pull/2644.

rossabaker:
If a release is imminent, we can probably start rolling 0.21 and 0.22 and pick up 0.23 no later than this week.

armanbilge:
Let me ping @mpilquist, when I asked yesterday it seemed possible but then I found that new regression (which btw I've fixed in https://github.com/typelevel/fs2/pull/2665).

armanbilge:
fs2 is on the way.

rossabaker (approved):
:+1: on green

Issue #5052, opened 2021-08-09
Blaze Tests Is Unreliable


isomarcte:
https://github.com/http4s/http4s/pull/5046/checks?check_run_id=3275323831

```
 ==> X org.http4s.client.blaze.BlazeClient213Suite.behave and not deadlock on failures with parTraverse  50.012s java.util.concurrent.TimeoutException: Future timed out after [50 seconds]
    at scala.concurrent.impl.Promise$DefaultPromise.tryAwait0(Promise.scala:212)
    at scala.concurrent.impl.Promise$DefaultPromise.result(Promise.scala:225)
    at scala.concurrent.Await$.$anonfun$result$1(package.scala:201)
    at scala.concurrent.BlockContext$DefaultBlockContext$.blockOn(BlockContext.scala:62)
    at scala.concurrent.Await$.result(package.scala:124)
    at munit.internal.PlatformCompat$.$anonfun$waitAtMost$1(PlatformCompat.scala:21)
    at scala.util.Try$.apply(Try.scala:210)
    at munit.internal.PlatformCompat$.waitAtMost(PlatformCompat.scala:21)
    at munit.FunSuite.waitForCompletion(FunSuite.scala:51)
    at munit.FunSuite.$anonfun$test$1(FunSuite.scala:37)
    at munit.GenericTest.$anonfun$withBodyMap$1(GenericTest.scala:33)
    at munit.MUnitRunner.$anonfun$runTestBody$1(MUnitRunner.scala:296)
```

If we can't fix this, we might consider disabling it. I'm unsure how much utility a test which is expected to fail often provides.

RafalSumislawski:
This sounds to me like a legitimate bug in the blaze-client.

rossabaker:
It might be, but sometimes we find the servers we spin up lock up in the CI environment buckle under load.  We made a lot of progress on test reliability by eliminating `unsafeRunSync()` across the test suites, but problems like this remain.  Are these clients really locking up, or are they just not performing well on an overwhelmed machine?

I'm convinced it's either a poorly written test OR a bug and that it needs attention.  I'm just not yet sure which.

RafalSumislawski:
I can reproduce the issue locally, which I think is a convincing evidence that this isn't a CI environment issue.

To reach (reasonably) reliable reproduction I commented out other tests in the suit and replaces `.replicateA(5)` with `.replicateA(200)`. It fails in ~50% attempts. But when it passes, it passes in less than 10s.


```
[...]

PR #5319, opened 2021-10-03
Introduce blaze-client connection reuse tests

RafalSumislawski:
There are some situations in which I would expect blaze-client to reuse a connection, but it doesn't. Moreover some changes I do in blaze client implementation seem to affect the reuse of connections in a way I find hard to predict. I wrote down my current knowledge and expectations regarding connection reuse in a form of tests, some of which currently fail or are flaky. This should help in preventing regressions and further improving the situation.

As next step I plan to fix the failing tests together with the #5052, which in some sense is an extreme case of not reusing a connection (because it has been leaked).


rossabaker:
Did you mean to open this against main or start on a production branch?

RafalSumislawski:
Oh. I meant to target `series/0.22`. Corrected.

blaze-client/src/test/scala/org/http4s/blaze/client/BlazeClientConnectionReuseSuite.scala
```diff
@@ -0,0 +1,197 @@
...
+import org.http4s.Method.{GET, POST}
+import org.http4s._
+
+import java.util.concurrent.TimeUnit
+import scala.concurrent.duration._
+
+class BlazeClientConnectionReuseSuite extends BlazeClientBase {
+  override def munitTimeout: Duration = new FiniteDuration(50, TimeUnit.SECONDS)
+
+  test("BlazeClient should the reuse connection after a simple successful request".flaky) {
```
rossabaker, line 32:
```suggestion
  test("BlazeClient should reuse the connection after a simple successful request".flaky) {
```

blaze-client/src/test/scala/org/http4s/blaze/client/BlazeClientConnectionReuseSuite.scala
```diff
@@ -0,0 +1,197 @@
...
+    }
+  }
+
+  //// Decoding failures ////
+
+  test("BlazeClient should reuse the connection after response decoding failed".flaky) {
+    // This will work regardless of whether we drain the entity or not,
+    // because the response is small and it is read in full in first read operation
+    val servers = makeServers()
+    val drainThenFail = EntityDecoder.error[IO, String](new Exception())
```
rossabaker, line 101: I wish we'd called this method `raiseError`.

blaze-client/src/test/scala/org/http4s/blaze/client/BlazeClientConnectionReuseSuite.scala
```diff
@@ -0,0 +1,197 @@
...
+  }
+
+  test(
+    "BlazeClient shouldn't reuse the connection after response decoding failed and the (large) entity wasn't drained") {
+    val servers = makeServers()
+    val failWithoutDraining = new EntityDecoder[IO, String] {
+      override def decode(m: Media[IO], strict: Boolean): DecodeResult[IO, String] =
+        DecodeResult[IO, String](IO.raiseError(new Exception()))
+      override def consumes: Set[MediaRange] = Set.empty
+    }
```
rossabaker, line 131: I would not have passed a quiz on how this differs from `EntityDecoder.error` if I weren't reading this!

client/src/test/scala/org/http4s/client/testroutes/GetRoutes.scala
```diff
@@ -35,6 +37,9 @@ object GetRoutes {
   def getPaths(implicit timer: Timer[IO]): Map[String, IO[Response[IO]]] =
     Map(
       SimplePath -> Response[IO](Ok).withEntity("simple path").pure[IO],
+      LargePath -> Response[IO](Ok).withEntity("a" * 8 * 1024).pure[IO],
+      HugePath -> IO( // delayed in IO to avoid allocating the huge string unless necessary
+        Response[IO](Ok).withEntity("a" * 100 * 1024 * 1024)),
```
rossabaker, line 42: Do we ever do this under load where we might be allocating several of these at once?  It could be a pure stream, so we should be able to share this response and pay for it once.

Or could we just stream it?  Is the point to be a 100MB chunk, or is the point to return 100MB?

RafalSumislawski: For now we use it in only one test and the goal is to have a response large enough that if the client tried to wait for it, it would take more than 5 seconds. The test would then hit the 5 seconds timeout and we would know that something is wrong.

I think I could rename it to "infinite" and then make it an infinite stream. For that specific test it shouldn't matter. In fact it would be better because that would remove the assumption that the test implementation currently makes: "transfering 100MB takes more than 5 seconds".

RafalSumislawski:
@rossabaker would it be ok if I added `ProblemFilters.exclude[ReversedMissingMethodProblem]("org.http4s.blaze.client.BlazeClientState.totalAllocations")`?

rossabaker:
Would it be binary compatible if you defaulted it to `F.pure(0)`?  MiMa exceptions on public interfaces have a history of biting us in the ass.

RafalSumislawski:
`F.pure(0)` works. The sad news is that it needs an `F` in the first place, so it ends up being:
```
def totalAllocations(implicit F: Applicative[F]): F[Long] = F.pure(0)
```
alternatively I could do something like
```
def totalAllocations: F[Long] = ???
```

rossabaker:
Oh, drat.  I don't think an `???` improves the situation beyond a MiMa exclusion.

An alternate encoding for this stats class would have been to make it pure, and return the stats value itself in `F`.  The disadvantage is that all of them need to be bundled and returned to access any of them.  The advantage is we have a much better story for extending it with more.  We could do that and deprecate this one if you think that's better.

Or we could do an awful `BlazeClientStats2 extends BlazeClientStats`, but I haven't thought through what impact that has on the builder.

RafalSumislawski:
Ok. I'll try out different approaches and see how they look. At first glance I like the idea of all stats as one value in F. That would make it easier to synchronise the access to the state of the `ConnectionPool` (currently there is no synchronisation, so from Java Memory Model perspective there's no guarantee that the thread asking for the stats will see any updates to the stats).

rossabaker:
If we upgrade to Servlet 4, we need Jetty 10, which requires Java 11.  This would be fine if we dropped Java 8, but Unix sockets are the only other pain point I can think of so far in our repo, and those require Java 17.

The same is true with Servlet 5, plus a breaking name change.  I'm not yet sure whether we'll ever support 4 or go straight to 5.

Tomcat has Servlet 4 and 5 implementations on Java 8.  That's a viable option if we exile Jetty to its own repo.  We plan to do that anyway in 1.0, which is where a servlet upgrade would happen.  So that should be a viable alternative.  I'm not sure that it's any lower level than the Sun server, though.

rossabaker (approved):
I leave it to @RafalSumislawski whether he wants to merge the progress here, and then add the Netty scaffold, or merge the Netty scaffold into this and build any additional tests on it.

RafalSumislawski:
I prefer merging #5601 and #5587 into their target branches and then dealing with this PR.

RafalSumislawski:
Hi @rossabaker, 
from my perspective this pull request is now ready. Feel free to merge if you don't have any remarks regarding the changes I introduced after your review.

danicheg (approved):
My two cents on this topic.

Perhaps tests of connection reuse when we sequentially send requests will always be flaky on the CI. Because between the end of one request and starting another could be a delay that we can't control at all. Only if the API will provide a mechanism for custom management of the connection for any request. But that thin ice, I guess.

Anyway, if @RafalSumislawski has some thoughts on reliable connection reuse, it would be great to see.

RafalSumislawski:
Hi @danicheg

I'm not sure if I'm getting you point. As long as the delay between sequential requests is smaller than the timeout of the connection (tens of seconds), the connection will sit and wait in the connection pool, ready to be reused.

I plan to remove all `.flaky` and `.fail` in #5385 where, according to my tests, the connection reuse happens reliably.

PR #5323, opened 2021-10-04
cors - return Access-Control-Allow-Headers if request headers non-matching

### LINKED ISSUE #5322: CORS middleware behaviour change - dropping Access-Control-Allow-Headers header
In the pre `CORSPolicy` implementation, if a preflight request included a header in `Access-Control-Request-Headers` that was not permitted by the CORS configuration, the CORS middleware would return a `Access-Control-Allow-Headers` header with the list of allowed headers. In the current implementation (so 0.23.3+, and not sure the equivalent for 1.0.0-x and 0.22.x) if the request header has more fields than are allowed, the response header gets dropped. 

Example in 0.23.4

```scala
import cats.effect.IO
import cats.effect.unsafe.implicits._
import org.http4s.server.middleware.CORS
import org.http4s._
import org.scalatest.flatspec.AnyFlatSpec
import org.scalatest.matchers.should.Matchers
import org.typelevel.ci.CIStringSyntax

class CORSTest extends AnyFlatSpec with Matchers {

  val cors = CORS.policy.withAllowHeadersIn(Set(ci"Foo", ci"Bar")).withAllowMethodsAll

  val routes = cors(HttpRoutes.pure[IO](Response[IO]())).orNotFound

  val request =
    Request[IO](
      method = Method.OPTIONS,
      headers = Headers(
        Header.Raw(ci"Origin", "http://example.com"),
        Header.Raw(ci"Access-Control-Request-Headers", "foo, faz"), //faz not allowed, foo is 
        Header.Raw(ci"Access-Control-Request-Method", "POST")
      )
    )

//FAILS
  it should "have an Access-Control-Allow-Headers header" in {
    //this header should be present and have value "Foo,Bar", but is missing
    routes(request).unsafeRunSync().headers.get(ci"Access-Control-Allow-Headers").isDefined shouldBe true
  }
}
```

and here is a worksheet using the old CORSConfig implementation where we see that the header is still returned under these circumstances https://scastie.scala-lang.org/6WjBVGOVRRy3vYGv4YZ6Qg

jbwheatley:
Fixes #5322 - full description of issue there. Changes behaviour of `allowHeadersIn` in `CORSPolicy` to match that of `allowedHeaders` in `CORSConfig`.

### DIFF
--- a/server/shared/src/main/scala/org/http4s/server/middleware/CORS.scala
+++ b/server/shared/src/main/scala/org/http4s/server/middleware/CORS.scala
@@ -361,7 +361,7 @@ sealed class CORSPolicy(
       }
       def headers = allowHeaders match {
         case AllowHeaders.All => Nil
-        case AllowHeaders.In(_) | AllowHeaders.Reflect =>
+        case AllowHeaders.In(_) | AllowHeaders.Reflect | AllowHeaders.Static(_) =>
           List(ci"Access-Control-Request-Headers")
       }
       (origin ++ methods ++ headers) match {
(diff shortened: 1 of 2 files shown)


rossabaker:
What if we added a `Static` to `AllowHeaders` that behaved this way?  We'd be able to keep our default implementation in the mainstream, but it would give a clean way to opt into this more relaxed behavior?  The sealed trait is private, so we should be able to grant the MiMa exclusion for it.  That would essentially be `Option C` from #5322.

jbwheatley:
thanks for looking into it @rossabaker, I'll modify this PR.

server/shared/src/main/scala/org/http4s/server/middleware/CORS.scala
```diff
@@ -361,7 +361,7 @@ sealed class CORSPolicy(
       }
       def headers = allowHeaders match {
         case AllowHeaders.All => Nil
-        case AllowHeaders.In(_) | AllowHeaders.Reflect =>
+        case AllowHeaders.In(_) | AllowHeaders.Reflect | AllowHeaders.Static(_) =>
```
rossabaker, line 364: In this case, I don't think the response varies by `Access-Control-Request-Headers`. It's better to have false positives than false negatives here, but it can hurt cache performance.

rossabaker (commented):
Other than the `Vary`, looks right to me.

The same could be added for methods, but that could also be an issue and a separate PR.  I think headers is your itch.

rossabaker:
Test failures unrelated.

rossabaker:
Oh, you probably want this on production.  I'll cherry-pick it back instead of merging it.

rossabaker:
Unless it's because that sealed type broke things more than I thought... I'll try after soccer practice.

jbwheatley:
> Oh, you probably want this on production. I'll cherry-pick it back instead of merging it.

Ah yes, wasn't 100% sure where to point the PR to. We're using 0.23.x in most places 😸

rossabaker:
Cherry-picked to 0.22 and merged to 0.23.

PR #5324, opened 2021-10-04
Update scalacheck-effect, ... to 1.0.3

rossabaker:


rossabaker:
Well, `GZipEncoding` doesn't like this!

Issue #5289, opened 2021-09-29
Drop Scala 2.12 for 1.0?


armanbilge:
I know this is a potential landmine, so please note the ? and that this is an issue and not a PR 😉

I wanted to spin this elephant out of https://github.com/http4s/http4s/discussions/5245#discussioncomment-1363798. IMHO, why:
* anyone who wants 2.12 support can take their pick from 0.21/0.22/0.23 etc.
* upstream pressure: https://github.com/circe/circe/pull/1818
* ecosystem catch-up: [SPARK-25075](https://issues.apache.org/jira/browse/SPARK-25075)
* singleton/literal types for https://github.com/http4s/http4s/pull/5093
* as noted by @bplommer, much better type inference, helpful for:
  * https://github.com/http4s/http4s/pull/4509
  * https://github.com/http4s/http4s/pull/5229
  * https://github.com/http4s/http4s/pull/5251

bplommer:
How about this:

1. Drop 2.12 from `main`
2. Develop against 2.13 until release candidate n.
3. Try to backport RCn to 2.12 while maintaining forwards binary (not source) compatibility.
  a. If this proves feasible, release the backported version as 1.0.0 and RCn as 1.1.0 - provide bug fixes only for 1.0.x
  b. If this doesn't prove feasible, 2.12 was holding http4s back and needs to be dropped.

rossabaker:
I mostly like Ben's idea.  It's an organic way to make the decision.  I don't love supporting two 1.x lines out of the gate along with an 0.23 we'll have to keep viable while people migrate.  I also don't think the backport will be feasible if we stop dragging 2.12 along now.

armanbilge:
I agree, it would be very organic but I'm also doubtful about the feasibility of a backport. Doesn't simply merging your PR using singleton types already put 2.12 support in a very awkward place?

I don't think it's _impossible_, but would probably require splitting into 2.12/2.13-3 sources. So then we're talking about two branches, with two separate sets of sources in the 1.0 branch (and all the ugliness involved with split files).

bplommer:
> I agree, it would be very organic but I'm also doubtful about the feasibility of a backport. Doesn't simply merging your PR using singleton types already put 2.12 support in a very awkward place?

That PR is against 0.22 and doesn't break bincompat, so yes, it puts 2.12 in an awkward place but it doesn't rule out a forward-binary-compatible backport.

[...]

PR #5328, opened 2021-10-04
Privatize HttpVersion constructors

rossabaker:
Binary breaking, but nobody should be calling these directly.


rossabaker:
I think we could eliminate the method entirely with #5289, or with the sealed abstract case class encoding.

armanbilge (approved):

PR #5329, opened 2021-10-04
Remove custom status reason phrases

rossabaker:
Follows up on the deprecations initiated in 0.22.6.

Also ports the encoding away from the weird case class with a second parameter list.  It's an exotic Scala feature that we don't need.

### DIFF
--- a/core/shared/src/main/scala/org/http4s/Status.scala
+++ b/core/shared/src/main/scala/org/http4s/Status.scala
@@ -18,9 +18,7 @@ package org.http4s
 
 import cats.{Order, Show}
 import org.http4s.Status.ResponseClass
-import org.http4s.internal.CharPredicate
 import org.http4s.util.Renderable
-import scala.annotation.nowarn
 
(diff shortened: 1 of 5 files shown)


core/shared/src/main/scala/org/http4s/Status.scala
```diff
@@ -69,30 +57,17 @@ sealed abstract case class Status private (code: Int)(
...
-    }
-
-  @nowarn("cat=deprecation")
   def apply(code: Int): Status =
-    apply(code, "", true)
-
-  private def trust(code: Int, reason: String, isEntityAllowed: Boolean = true): Status =
-    new Status(code)(reason, isEntityAllowed) {
-      override val sanitizedReason = reason
+    trust(code, "", true)
```
rossabaker, line 61: Nothing is validating the range of the code.  We've got more deprecatin' to do.

armanbilge (approved):

PR #5330, opened 2021-10-05
Merge from 0.22

rossabaker:


ember-server/src/main/scala/org/http4s/ember/server/internal/ServerHelpers.scala
```diff
@@ -32,15 +33,15 @@ import org.http4s.ember.core.{Drain, EmberException, Encoder, Parser, Read}
...
 
 import scala.concurrent.duration._
 import scodec.bits.ByteVector
-import fs2.io.net.unixsocket.UnixSocketAddress
-import fs2.io.net.unixsocket.UnixSockets
-import org.typelevel.vault.Key
-import _root_.org.http4s.websocket.WebSocketContext
+import org.http4s.headers.Connection
+import java.nio.channels.InterruptedByTimeoutException
+import java.util.concurrent.TimeoutException
```
rossabaker, line 44: I'm mainly skeptical of these two imports and what they mean in a Scala.js world.

armanbilge: ember-server is not crossed on 0.23 so it's okay here. But it will be problematic in 1.0.

`j.u.c.TimeoutException` is ported in Scala.js itself, so that should be fine:
https://github.com/scala-js/scala-js/blob/master/javalib/src/main/scala/java/util/concurrent/Throwables.scala#L30

`java.nio.channels.InterruptedByTimeoutException` is definitely not, and I don't remember adding it to fs2.io yet. That would probably be the right thing to do.

rossabaker: Testing my developing intuition: on 1.0, this would compile, but generate a linking error if EmberServer is used, because it would pull ServerHelpers into the .js?

armanbilge: Bingo 😀

armanbilge: So what is the difference between `InterruptedByTimeoutException` and `SocketTimeoutException`?

On Node.js we have `ETIMEDOUT` errors which I'm currently mapping to `SocketTimeoutException`.
> ETIMEDOUT (Operation timed out): A connect or send request failed because the connected party did not properly respond after a period of time. Usually encountered by http or net. Often a sign that a socket.end() was not properly called.

rossabaker: It's not thrown directly by http4s, fs2, or cats-effect, but it comes from an `AsynchronousSocketChannel`:

> The read and write methods defined by this class allow a timeout to be specified when initiating a read or write operation. If the timeout elapses before an operation completes then the operation completes with the exception InterruptedByTimeoutException.

armanbilge: Thanks for the pointer. Will this exception ever get thrown though? Looking at the [Javadocs for `AsynchronousSocketChannel`](https://docs.oracle.com/javase/7/docs/api/java/nio/channels/AsynchronousSocketChannel.html) I see timeouts can be passed to the read/write methods. But, looking in fs2.io I don't see any timeouts being provided:
https://github.com/typelevel/fs2/blob/main/io/jvm/src/main/scala/fs2/io/net/SocketPlatform.scala

rossabaker: I would have to delegate to @RaasAhsan on this one, since he was motivated to add it to the catch list in the first place.

armanbilge: I suppose it could if someone initializes `ember-server` with their own `fs2.io.net.SocketGroup` implementation.

In any case, I'm not sure this exception belongs in fs2.io.js, in which case we'll have to find a workaround in 1.0 (I wonder if we can type alias it to `Nothing` on JS without setting of compiler warnings).

RaasAhsan: `TimeoutException` is thrown by Ember Server in the event of a header read timeout. We can map that to an `EmberException` or something else. `InterruptedByTimeoutException` will be thrown by FS2 2.x (https://github.com/typelevel/fs2/blob/series/2.5.x/io/src/main/scala/fs2/io/tcp/SocketGroup.scala#L240).

seems like the API changes in fs2 3.x and a timeout must now be coded by the caller

armanbilge: Aha, thanks for the explanation. So sounds like it would be okay to remove this case for 0.23 and 1.x then?

RaasAhsan: we should replace `InterruptedByTimeoutException` with what the timeout mechanism in series/0.23+ raises (I always work on series/0.22 so I can't remember what it is)

armanbilge:
Did this merge with some compile failures?

rossabaker:
Yes, sorry.  I moved too fast.

PR #5337, opened 2021-10-06
Drop Scala 2.12 from 1.0

armanbilge:
Following up on https://github.com/http4s/http4s/issues/5289, where IMHO nobody made a compelling case to keep 2.12 for 1.0, at least during development. Despite the title, it's less about dropping 2.12 and more about unblocking development of 1.0.

This can sit while we get used to the idea 😁


bplommer:
Looks like this isn't happening for now - shall we close this?

armanbilge:
I won't give up so easy! I'm happy to let this sit.

core/shared/src/main/scala/org/http4s/multipart/Boundary.scala
```diff
@@ -41,9 +42,10 @@ object Boundary {
   private val rand = new Random()
 
   private def nextChar = CHARS(rand.nextInt(nchars - 1))
-  private def stream: LazyList[Char] = LazyList.continually(nextChar)
-  //Don't use filterNot it works for 2.11.4 and nothing else, it will hang.
-  private def endChar: Char = stream.filter(_ != ' ').headOption.getOrElse('X')
+  private def stream: LazyList[Char] =
+    LazyList.continually(nextChar)
+  // Don't use filterNot it works for 2.11.4 and nothing else, it will hang.
```
bplommer, line 47: Unrelated but I think this comment is out of date 😅

armanbilge: Do you happen to know what the original issue was?

bplommer: No idea, I guess it was fixed in 2.11.4 though? We could try changing it to filterNot and see if the tests hang, but that doesn’t need to happen in this PR

armanbilge: Ohh. I was reading it the other way, that _after_ 0.11.4 it's broken. Yeah I should but I was feeling lazy :P

armanbilge:
Anyone know why this is now able to compile, when it shouldn't?
```
error: json.md:110:1: Expected compile errors but program compiled successfully without errors
Hello("Alice").asJson
^^^^^^^^^^^^^^^^^^^^^
```
https://github.com/http4s/http4s/blob/c9781f51974a82759e6eec1a1eb22e73d08a876b/docs/src/main/mdoc/json.md#L109-L111

armanbilge:
Ok, I'm bamboozled. I added a `nest` modifier to the next code block and it fixes the problem above.

However, I'm still getting the `Expected compile errors but program compiled successfully without errors` problem. So, I remove the `fail` modifier and ... lo and behold, the code fails to compile for the expected reason. So idk.

armanbilge:
We've got a fix incoming in https://github.com/http4s/http4s/pull/5984.

armanbilge:
My trigger finger is starting to itch 😜

PR #5342, opened 2021-10-07
Enable http4s-circe.js to dead-code-eliminate Jawn

armanbilge:
Some minor adjusting of the circle module to allow Jawn to be completely dead-code-eliminated on Scala.js (verified manually). Addresses https://github.com/http4s/http4s-dom/issues/10, this was not at all as involved as I thought!

It is graceful, such that the default decoder bypasses Jawn, but other Jawn-based decoders will also continue to work on Scala.js if referenced directly.

This adds a new dependency to `circe-parser`, which is a _tiny_ wrapper for Jawn on JVM and the native JSON parser on Scala.js (i.e., adds no new dependencies). However, it's only used on Scala.js (there are no changes for JVM).

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -544,6 +544,7 @@ lazy val circe = libraryCrossProject("circe", CrossType.Pure)
     startYear := Some(2015),
     libraryDependencies ++= Seq(
       circeCore.value,
+      circeParser.value,
       circeTesting.value % Test
     )
   )
@@ -607,7 +608,7 @@ lazy val bench = http4sProject("bench")
   .settings(
     description := "Benchmarks for http4s",
     startYear := Some(2015),
-    libraryDependencies += circeParser,
+    libraryDependencies += circeParser.value,
     undeclaredCompileDependenciesTest := {},
     unusedCompileDependenciesTest := {},
   )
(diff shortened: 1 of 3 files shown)


armanbilge:
Oops, I didn't think carefully enough about the configuration options `maxValueSize` and `allowDuplicateKeys` options in `CirceSupportParser`. The Javascript-native JSON parser only supports `maxValueSize = None` and `allowDuplicateKeys = true`, so we could fallback to the Jawn parser for other choices. Except because they are set dynamically, this would inhibit dead-code-elimination.

I think we'll have to ignore these settings for the default decoder on Scala.js (alternate decoders could still use them, since they are opt-in). I loathe doing this, since neither the semantics (nor the defaults!) would match what the JVM is doing, but I don't think there is a workaround. Ugh. Thoughts?

rossabaker:
Yeah, let's demonstrate it's a problem in practice before we diverge from something that works uniformly.

armanbilge:
Just wanted to add some notes here after my experience working on https://github.com/satabin/fs2-data/pull/309.

In that PR, I discovered (unsurprisingly) that the JS-native JSON parser rounds very small numbers to 0. Jawn handles them beautifully. Jawn is good.

armanbilge:
One more follow-up: seems that the Jawn parser may _outperform_ the `JSON.parse` method in many situations. See
- https://github.com/circe/circe/issues/1941
- Benchmarks: https://www.armanbilge.com/jsoniter-scala/index-scalajs.html

I'm not sure if these results translate from browsers to Node.js. But it's certainly an interesting data point. Jawn is very good :)

PR #5345, opened 2021-10-07
Use `scodec.bits.crc` instead of JVMism

armanbilge:


armanbilge:
It was relying on the mutability of `TrailerGen` to work which IMO is an anti-pattern. Happy to put it back.

PR #5346, opened 2021-10-07
Relax Async constraints

bplommer:


bplommer:
Just binary incompatible. I didn't want to leave you with another of my half-baked PRs but also didn't have the energy to fix it.

PR #5349, opened 2021-10-07
Fix deadlocks in Retry and FollowRedirect middlewares

### LINKED ISSUE #5180: Retry Client Hanging
I've been experiencing an issue where the retry client seems to hang in certain cases. I was able to replicate it here:
https://github.com/tdbgamer/http4s-retry-hang

Not sure what's going on, but in the thread dump, all the IO threads are waiting on the same thing.

Maybe it could be related to [this](https://github.com/typelevel/cats-effect/issues/2050)?

rossabaker:
These should never allocate a second connection before releasing the first.  If they do, we can deadlock trying to allocate a second connection while everybody holds onto their first.

Fixes #5180.

### DIFF
--- a/client/shared/src/main/scala/org/http4s/client/middleware/FollowRedirect.scala
+++ b/client/shared/src/main/scala/org/http4s/client/middleware/FollowRedirect.scala
@@ -99,19 +99,20 @@ object FollowRedirect {
         req: Request[F],
         redirects: Int,
         hotswap: Hotswap[F, Response[F]]): F[Response[F]] =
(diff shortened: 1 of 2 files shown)


rossabaker:
The fact we had to invoke both operations of `Hotswap` suggests to me that this just isn't the right state machine, but this solution seems to work.  Maybe I can nerdsnipe @ChristopherDavenport into a better state machine for this purpose.

rossabaker:
This is a good diff to review with whitespace insignificance.

PR #5356, opened 2021-10-11
Experimental: Http1 codec reform

rossabaker:
Trying to replace `Renderable` with something both more functional and performant.  It's a riff on [this gist]( https://gist.github.com/rossabaker/6a5b4be9ddd61b8ae0ca1346f61cb06e) with [twitter discussion](https://twitter.com/rossabaker/status/1443948901665546251).

Our current approach is to render to a `StringBuilder`, and then ASCII-encode that at once into an `Array[Byte]`.  This has proven more efficient than using a `CharsetEncoder` to incrementally encode small strings to a `ByteBuffer`, and much more efficient than converting small strings to concatenated `ByteVector`s or `Chunk`s.

The new approach proposed here is to create an internal byte buffer similar to `StringBuilder` that assumes one-char-per-byte and otherwise operates just like `StringBuilder`.  The array storage from this buffer is wrapped directly in a `Chunk`, eliminating a copy.  The mutable state can be wrapped in a protected method and composed via `ContravariantSemigroupal`.  Careful use of combinators like `prefix` and `suffix` eliminates nested tuples from the semigroupal and considerably improves performance.

It is named `Http1Encoder`, because HTTP/2 is stateful and I don't even begin to understand HTTP/3.

### DIFF
--- a/bench/src/main/scala/org/http4s/bench/Http1DecoderBench.scala
+++ b/bench/src/main/scala/org/http4s/bench/Http1DecoderBench.scala
@@ -1,0 +1,101 @@
+/*
+ * Copyright 2015 http4s.org
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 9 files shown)


rossabaker:
An unresolved tension here is that our parser is character-based, and our encoding would become byte-based.  This is going to rear its head with HTTP/2, I think.

PR #5360, opened 2021-10-11
MediaType/MediaRange improvements - WIP

### LINKED ISSUE #5359: Improve spec-compliance of `MediaType`/`MediaRange`
The current implementation seems to be incorrect with respect to case-sensitivity: According to [RFC 7231, section 3.1.1.1](https://httpwg.org/specs/rfc7231.html#media.type) the main type, subtype, and parameter names are case-insensitive, but as currently implemented they're case-sensitive. Parameter values may or may not be case-sensitive (charset values are not, for example), but currently they are always compared case-sensitively.

This is partially ameliorated by the fact that the parsers convert main and subtypes to lower case, but that has two shortcomings: (1) the constructor for `MediaType` is public, so it doesn't actually enforce normalization, and (2) enforced normalization would limit user control over rendering (which becomes a problem e.g. if using [pact testing](https://github.com/pact-foundation/pact-specification/tree/version-2), which enforces a strict case-sensitive match on all headers).

On the topic of enforced normalization there are a couple of other issues: firstly, storing parameters as a `Map` means ordering isn't preserved - but the spec doesn't state that differently ordered parameter lists are equivalent (or indeed that keys can't be repeated). Secondly, the spec states that `A parameter value that matches the token production can be transmitted either as a token or within a quoted-string. The quoted and unquoted values are equivalent.` but `the first is preferred for consistency` - but http4s always renders the value as a quoted-string.

bplommer:
Will resolve #5359. Most of the diff is from manually editing the MimeDB file to add `ci` interpolators.

Still to do:
- [ ] Update MimeDB plugin
- [ ] Preserve choice of `token` vs `quoted-string` for parameter values when parsing
- [ ] Render parameter values as `token` rather than `quoted-string` by default when possible
- [x] Preserve order of parameter values
- [ ] Make charset case-insensitive
- [ ] Make parameter value case-sensitivity configurable?
- [ ] Test new behaviour

Some of this should be possible to backport to 0.22.

### DIFF
--- a/core/shared/src/main/scala/org/http4s/MediaType.scala
+++ b/core/shared/src/main/scala/org/http4s/MediaType.scala
@@ -37,7 +37,7 @@ import scala.util.hashing.MurmurHash3
 
 sealed class MediaRange private[http4s] (
     val mainType: CIString,
-    val extensions: Map[CIString, String] = Map.empty) {
+    val extensions: List[(CIString, String)] = List.empty) {
 
   /** Does that mediaRange satisfy this ranges requirements */
   def satisfiedBy(mediaType: MediaRange): Boolean =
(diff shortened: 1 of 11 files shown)


bplommer:
I don't think I'll get back to this for a while so it would be helpful if I could get some feedback on it now.

PR #5365, opened 2021-10-11
Release v0.22.7

rossabaker:


armanbilge:
As of v0.2 http4s-crypto has a CE3 dependency, so no.

PR #5385, opened 2021-10-17
Resourcify blaze client

RafalSumislawski:
This PR:
- Provides new foundations for the efforts to make `blaze-client` leak-and-deadlock-less by leveraging `Resource`
- Makes connection reuse tests introduced in #5319 pass
- Fixes connection leak on interruption leading to deadlock and flakiness of one of the tests #5052
- May accidentally fix flakiness of some other `blaze-client` tests (I'm investigating that in #5724)

This PR doesn't (a.k.a the list of future work):
- solve all the issues with cancellation of read and write operations in Http1Connection and how it interacts with connection reuse.
- improve resource management in Http1Connection
- fix CE2 cancellation model
- clean-up the logic of connection reuse (It bugs me that the decision to reuse or discard a connection is partially made by `BlazeClient` and partially by `Http1Connection#isRecyclable`)


blaze-client/src/main/scala/org/http4s/blaze/client/BlazeClient.scala
```diff
@@ -66,95 +63,82 @@ object BlazeClient {
...
 
-        def borrow: Resource[F, manager.NextConnection] =
-          Resource.makeCase(manager.borrow(key)) {
-            case (_, ExitCase.Completed) =>
-              F.unit
-            case (next, ExitCase.Error(_) | ExitCase.Canceled) =>
-              invalidate(next.connection)
-          }
+  override def run(req: Request[F]): Resource[F, Response[F]] = for {
+    _ <- Resource.pure[F, Unit](())
```
ChristopherDavenport, line 80: Why this rather than 
```
{
  val key = RequestKey.fromRequest(req)
  for {
   ...
   ```

Comment (RafalSumislawski) [reply blaze-client/src/main/scala/org/http4s/blaze/client/BlazeClient.scala:80]: There's no reason other than a stylistic choice. It results in "flatter" code. I guess that's something that is popular in my team, and looks strage to everyone outside of the team ;).  I'll change it to the `val key` version.

**Comment by RafalSumislawski** [comment]
Hi @rossabaker 
May I bring you attention to this PR?

**Comment by rossabaker** [comment]
Sorry, I missed that I'd been tagged and thought it was in draft and it fell beneath my triage radar.  I'll get this into 0.22.10 if it's cooked.

**Review by rossabaker — COMMENTED** [review]
So, it looks like I reviewed this in October, but left it pending. 🤦

File: blaze-client/src/main/scala/org/http4s/blaze/client/BlazeClient.scala [hunk 2]
```diff
@@ -21,21 +21,19 @@ package client
 import cats.effect._
 import cats.effect.concurrent._
 import cats.effect.implicits._
-import cats.syntax.all._
-import java.net.SocketException
+import cats.implicits._
```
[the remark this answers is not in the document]
rossabaker: Not a blocker, don't care.  May clean up if I do the merge.  I think it's getting late where @RafalSumislawski is.

blaze-client/src/main/scala/org/http4s/blaze/client/BlazeClient.scala
```diff
@@ -63,88 +61,78 @@ object BlazeClient {
...
+        Resource.apply(
+          Deferred[F, Either[Throwable, TimeoutException]].flatMap(timeout =>
+            F.delay {
+              val stage = new ResponseHeaderTimeoutStage[ByteBuffer](d, scheduler, ec)
+              conn.spliceBefore(stage)
+              stage.init(e => timeout.complete(e).toIO.unsafeRunSync())
+              (timeout.get.rethrow, F.delay(stage.removeStage()))
+            })
+        )
+      case _ => Resource.pure[F, F[TimeoutException]](F.never)
```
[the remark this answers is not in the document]
rossabaker: Nano-optimization, don't really care.

blaze-client/src/main/scala/org/http4s/blaze/client/Http1Connection.scala
```diff
@@ -204,49 +206,45 @@ private final class Http1Connection[F[_]](
...
+            )(writeFiber =>
+              receiveResponse(
+                mustClose,
+                doesntHaveBody = req.method == Method.HEAD,
+                cancellation.race(timeoutFiber.join).map(e => Left(e.merge)),
+                idleRead
+              ).map(response =>
+                // We need to stop writing before we attempt to recycle the connection.
+                Resource
+                  .make(F.pure(writeFiber))(_.cancel)
```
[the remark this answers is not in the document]
rossabaker: Outdated.

rossabaker (approved):
I looked at the rest.  I'll try to do the merge and see what CI thinks, but I think this is nice.

rossabaker:
So this clashes pretty hard with the retry I added in the meantime.  I think the work here is great, but the merge resolution is going to need a hard look.  We're going to need a fake Hotswap, I think.

blaze-client/src/main/scala/org/http4s/blaze/client/BlazeClient.scala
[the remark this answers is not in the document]
RafalSumislawski: I must have added the `cats.implicits._` (Force of habbit) and then removed the `cats.syntax.all._` because it was unused. 

It is getting late, but I'm still up, just got sucked in some performance profiling.

blaze-client/src/main/scala/org/http4s/blaze/client/BlazeClient.scala
```diff
@@ -63,88 +61,78 @@ object BlazeClient {
...
-          }
+  private def addResponseHeaderTimeout(conn: A): Resource[F, F[TimeoutException]] =
+    responseHeaderTimeout match {
+      case d: FiniteDuration =>
+        Resource.apply(
+          Deferred[F, Either[Throwable, TimeoutException]].flatMap(timeout =>
+            F.delay {
+              val stage = new ResponseHeaderTimeoutStage[ByteBuffer](d, scheduler, ec)
+              conn.spliceBefore(stage)
+              stage.init(e => timeout.complete(e).toIO.unsafeRunSync())
```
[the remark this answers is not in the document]
RafalSumislawski: I'm actually not sure which one is better. My intuition says `.toIO.unsafeRunSync()` is more efficient when `F = IO`. and `runAsync` is more efficient otherwise. I'll change it to runAsync.

RafalSumislawski:
I just pushed changes corresponding to you comments. I'll try understanding the merge conflict tommorow.

rossabaker:
The essence of the merge conflict is where to put https://github.com/http4s/http4s/commit/bda08ab7494e6e351654386e97b29aeea368d379#diff-a2695172d5a051556d7429b099d8ec33133ab030d2784652cdc219829ef73f50R132-R135.  I'll PR to your branch if I come up with a solution, so I don't mess up this branch if my idea is bad.

rossabaker:
If we loop in `runRequest`, the `SocketException` we need is still caught in the `Resource`.  If we loop up near `run`, I don't have the connection to invalidate, as is done in `borrowConnection`.  

I'd like to wrap it in `Retry` just like Ember does, but that depends on a `Timer`.  Maybe we just need a dumber version of that, because we don't delay.

rossabaker:
I can't PR to your fork for some reason, but my attempt at the merge is https://github.com/rossabaker/http4s/commit/b540904ebc646eb7f8534a5b35148f2e43829504.  I still have three failing tests:

```
==> X org.http4s.blaze.client.BlazeClientSuite.does not retry non-idempotent requests  0.058s munit.ComparisonFailException: /Users/ross.baker/src/http4s/blaze-client/src/test/scala/org/http4s/blaze/client/BlazeClientSuite.scala:352
351:        builder(1, retries = 3).resource
352:          .use(client => client.status(req).attempt *> attempts.get.assertEquals(1))
353:      }
values are not the same
=> Obtained
4
-4
+1
```

```
==> X org.http4s.blaze.client.BlazeClientConnectionReuseSuite.BlazeClient.status shouldn't wait for an infinite response entity and shouldn't reuse the connection  0.023s org.http4s.blaze.client.Http1Connection$InProgressException$: Stage has request in progress
```

```
==> X org.http4s.blaze.client.BlazeClientConnectionReuseSuite.BlazeClient shouldn't reuse the connection after response decoding failed and the (large) entity wasn't drained  1.783s org.http4s.blaze.client.Http1Connection$InProgressException$: Stage has request in progress
    at org.http4s.blaze.client.Http1Connection$InProgressException$.<clinit>(Http1Connection.scala:473)
    at org.http4s.blaze.client.Http1Connection.$anonfun$runRequest$1(Http1Connection.scala:165)
```

blaze-client/src/main/scala/org/http4s/blaze/client/BlazeClient.scala
```diff
@@ -172,5 +77,91 @@ object BlazeClient {
...
+    val key = RequestKey.fromRequest(req)
+    for {
+      requestTimeoutF <- scheduleRequestTimeout(key)
+      preparedConnection <- prepareConnection(key)
+      (conn, responseHeaderTimeoutF) = preparedConnection
+      timeout = responseHeaderTimeoutF.race(requestTimeoutF).map(_.merge)
+      responseResource <- Resource.eval(runRequest(conn, req, timeout)).recoverWith {
+        case _: SocketException if remaining > 0 && req.isIdempotent =>
+          Resource.eval(manager.invalidate(conn) *> F.pure(runLoop(req, remaining - 1)))
+      }
```
rossabaker, line 106: This `recoverWith` is the meaningful part of the merge.  I didn't think this would work without a `F.continual`, as in the `Retry` middleware.  I'm concerned that we're allocating a second connection before releasing the first, but the test only allows one request per key.

rossabaker (commented):
I got the tests passing, so I just pushed here.

rossabaker:
I'd like to get eyes on this, because I think we can release 0.22.10 with it.

The merge to CE3 is going to be painful.

RafalSumislawski:
I don't like the fact that the `invalidate` is called outside of the releaseing part oth the `Resource` in `borrowConnection`. But I have no idea how to improve on that, so I'm ok with the current shape of the code.

RafalSumislawski:
I can't approve this PR so I'll just say LGTM.

rossabaker:
Oh, without that invalidate, it tries to borrow a second connection before releasing the first, which is what I was [concerned about here](https://github.com/http4s/http4s/pull/5385#discussion_r792297853).  I think that out-of-band invalidation is what allowed us to do this without `Hotswap` (CE3) or `F.continual` (what `Retry` does in CE2).

rossabaker:
But invalidating a connection is not idempotent.  It calls `decrConnection`.  So when the connection resource gets cleaned up, it's going to decrement twice, de facto expanding the size of the connection pool.  I don't think this is ready yet.

rossabaker:
This fixes the `BlazeClientSuite`, but introduces the [failures on BlazeClientConnectionReuseSuite](https://github.com/http4s/http4s/pull/5385#issuecomment-1021682566).

```scala

  override def run(req: Request[F]): Resource[F, Response[F]] =
    runLoop(req, retries)

  // A light implementation of the Retry middleware.  That needs a
  // Timer, which we don't have.
  private def runLoop(req: Request[F], remaining: Int): Resource[F, Response[F]] =
    Resource.suspend[F, Response[F]](F.continual(respond(req).allocated) {
      case Right((response, dispose)) =>
        F.pure(Resource.make(F.pure(response))(_ => dispose))
      case Left(_: SocketException) if remaining > 0 && req.isIdempotent =>
        val key = RequestKey.fromRequest(req)
        logger.debug(s"Encountered a SocketException on ${key}.  Retrying up to ${remaining} more times.")
        F.pure(runLoop(req, remaining - 1))
      case Left(e) =>
        F.pure(Resource.eval(F.raiseError(e)))
    })

  private def respond(req: Request[F]): Resource[F, Response[F]] = {
    val key = RequestKey.fromRequest(req)
    for {
      requestTimeoutF <- scheduleRequestTimeout(key)
      preparedConnection <- prepareConnection(key)
      (conn, responseHeaderTimeoutF) = preparedConnection
      timeout = responseHeaderTimeoutF.race(requestTimeoutF).map(_.merge)
      responseResource <- Resource.eval(runRequest(conn, req, timeout))
      response <- responseResource
    } yield response
  }
```

rossabaker:
I built another version on fs2-2's Hotswap, which I didn't realize exists.  It has the same problem.  In the looping implementation, the connection is getting released before its state is set to `ReadActive`, which results in it being deemed recyclable.

blaze-client/src/main/scala/org/http4s/blaze/client/Connection.scala
```diff
@@ -26,7 +26,7 @@ private[client] trait Connection[F[_]] {
   def isClosed: Boolean
 
   /** Determine if the connection is in a state that it can be recycled for another request. */
-  def isRecyclable: Boolean
+  def isRecyclable: F[Boolean]
```
rossabaker, line 29: Some other methods here could get the same treatment, but I want to get this landed.

blaze-client/src/main/scala/org/http4s/blaze/client/BlazeClient.scala
```diff
@@ -90,20 +92,42 @@ private class BlazeClient[F[_], A <: BlazeConnection[F]](
...
+
   override def run(req: Request[F]): Resource[F, Response[F]] =
-    runLoop(req, retries)
+    Hotswap.create[F, Either[Throwable, Response[F]]].flatMap { hotswap =>
+      Resource.eval(retryLoop(req, retries, hotswap))
+    }
 
-  private def runLoop(req: Request[F], remaining: Int): Resource[F, Response[F]] = {
+  // A light implementation of the Retry middleware.  That needs a
+  // Timer, which we don't have.
```
rossabaker, line 103: In my desperation to get the tests passing, I found fs2 had an implementation, so this actually looks more like 0.23's Retry, and we could rebuild 0.22's on it.  It's cleaner, but I don't want to touch something on an aging branch not known to have bugs.

build.sbt
```diff
@@ -629,6 +629,29 @@ lazy val blazeClient = libraryProject("blaze-client")
...
+      ProblemFilters
+        .exclude[MissingClassProblem]("org.http4s.blaze.client.Http1Connection$ReadWrite$"),
+      ProblemFilters.exclude[MissingClassProblem]("org.http4s.blaze.client.Http1Connection$Write$"),
+      ProblemFilters.exclude[IncompatibleResultTypeProblem](
+        "org.http4s.blaze.client.Http1Connection.isRecyclable"
+      ),
+      ProblemFilters
+        .exclude[IncompatibleResultTypeProblem]("org.http4s.blaze.client.Connection.isRecyclable"),
+      ProblemFilters
+        .exclude[ReversedMissingMethodProblem]("org.http4s.blaze.client.Connection.isRecyclable"),
```
armanbilge, line 654: Are none of these Scala 3 only filters? Support for auto-filtering package-privates on Scala 3 got merged in https://github.com/lightbend/mima/pull/683 so we'll be able to remove many of these filters after the next release.

rossabaker: The ones I added are Scala 2, and I'm pretty sure the rest are, too.

I usually create crufty deprecated methods even when they're package private, because I live in constant terror.  Others don't, and they have a fine argument, so I don't push back.  I couldn't make my changes starting with 7ba8cf9 without exceptions.

rossabaker:
Any objections.  I'd love to release this today.

rossabaker:
I'm going to start merging this to 0.23, but this should be good to unblock an 0.22 release.

PR #5416, opened 2021-10-19
Expand docs on client middlewares

kubukoz:


kubukoz:
Do we want this backported to other branches? Let me know how I can help with that. Not sure if it can help but Mergify in the free tier is capable of creating backport PRs: https://mergify.io/pricing

rossabaker:
0.22 is still officially maintained, so that's a nice bonus.  It might cherry pick nicely.  I don't know how many CE3-isms it has.

I didn't know mergify did backports.  We usually merge forward, so I'm not sure how helpful that is, but it's also worth a glance.

kubukoz:
There's `MonadCancelThrow` and that's probably just it. I'll send a backport to series/0.22.

PR #5429, opened 2021-10-20
Model Vary header

bplommer:


rossabaker (commented):
That test error reminds me we need to improve our generators.

danicheg (commented):
I was thinking about a header, that encodes as a wildcard (`*`) or list of values separated by commas and accidentally found that PR.
So here are my two cents on that kind of header. Perhaps, we could encode them as a sum type `header = * | NEL[Value]` and have it as `Header.Recurring` + provide a `Semigroup` instance. In that case, if someone will use the `Headers#get[A]` method and the header was sent multiple times (for the `Vary` header it's pretty common, e. g.), then it will be parsed and folded correctly into a single header (according to the `Semigroup` instance). What behavior will be in the scenario of `Header.Single`?

server/src/main/scala/org/http4s/server/middleware/CORS.scala
```diff
@@ -360,7 +360,7 @@ sealed class CORSPolicy(
     val varyHeaderNonOptions =
       allowOrigin match {
         case AllowOrigin.Match(_) =>
-          Header.Raw(ci"Vary", Header[Origin].name.toString).some
+          Vary(Header[Origin].name).some
```
danicheg, line 363: _[compile-time micro-opt nitpick]_: we can override the `HeaderCompanion#name` to be a `val` and use it when needed

PR #5496, opened 2021-10-27
Disable logging for Ember client internal retry

RaasAhsan:
This is coming from a user issue reported in Discord: https://discord.com/channels/632277896739946517/632286375311573032/902589984391983134

> gtc_10001
 — 
Today at 11:10 AM
Hi, I'm using the Ember client (0.23.6), and I noticed that when the client automatically retries using an instance of Retry one of our custom sensitive headers that is normally redacted during logging by the middleware (via use of redactHeadersWhen) is being shown.  It seems that EmberClientBuilder allows you to specify a retry policy, but there's no way to specify a redactHeaderWhen function for the underlying instance of Retry. I ended up turning off that logger in my logging config but I ideally would like the retry logged without the sensitive header. Does anyone know of a way to do that with the current code?

The recommendation was that we disable logging for the internal retry since users are not even supposed to be aware of it.

Some alternatives:
1. Manually handcode the internal retry
2. Factor out a more reusable `Retry` interface which can invoke callback effects for the different states


ChristopherDavenport (approved):
Builders are definitely coming to middlewares functions are just hostile to bincompat. Especially those with default params. 

The inevitably leads to us coalescing a default stack as well. Because the builders will cause boiler plate. Buckle up.

ChristopherDavenport:
No, I think we still want this.

PR #5506, opened 2021-10-28
Non-trivial merge to 0.23

rossabaker:


bplommer:
Just re-triggered it and same deal. Any idea what's going on here?

armanbilge:
Ah, I know what's going on. The ci thing which segregates JVM/JS builds in the matrix went missing. So now it's trying to do both at the same time and Scala.js linking step is _extremely_ memory-intensive. So GH is killing the job.

.github/workflows/ci.yml
```diff
@@ -65,27 +77,27 @@ jobs:
 
       - name: Check formatting
         if: matrix.scala != '3.1.0'
-        run: sbt ++${{ matrix.scala }} '${{ matrix.ci }}' scalafmtCheckAll
```
armanbilge, line 68: This `${{ matrix.ci }}` is what was used to to set the root project to either `rootJVM` or `rootJS`.

bplommer: I doubt Ross would mind if you felt like pushing a fix 😄

rossabaker: Oh, something in the merge made it look like that had been intentionally removed, and I didn't understand it, so I played along.

rossabaker:
Nice green checkmark for a build so broken that didn't run at all.

bplommer (approved):

PR #5507, opened 2021-10-28
Update all the urls to new http4s-dom site

armanbilge:
I case you missed it, it has a live example of http4s in the browser! 😁 
https://http4s.github.io/http4s-dom


rossabaker:
Do you want it here, or do you want a subdomain on http4s.org?

armanbilge:
Sure, a subdomain would be cool, but this is fine too.

rossabaker:
I'm good with this until the subdomain is set up.  Is the breakage that other issue you fixed?

armanbilge:
Yeah, this inherited a bad main but should be fixed now.

PR #5514, opened 2021-10-29
Fix ci on main

armanbilge:
It's bwoken 😕

### DIFF
--- a/ember-client/js/src/main/scala/org/http4s/ember/client/internal/ClientHelpersPlatform.scala
+++ b/ember-client/js/src/main/scala/org/http4s/ember/client/internal/ClientHelpersPlatform.scala
@@ -28,7 +28,7 @@ import scala.annotation.nowarn
 
 private[internal] trait ClientHelpersPlatform {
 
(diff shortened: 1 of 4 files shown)


bplommer:
Does scala 3 support `"msg=unused"`?

bplommer (approved):

armanbilge:
Apparently not.

Still getting some JS-related failures, which frustratingly I cannot reproduce locally 😠

ember-client/js/src/main/scala/org/http4s/ember/client/internal/ClientHelpersPlatform.scala
```diff
@@ -28,7 +28,7 @@ import scala.annotation.nowarn
 
 private[internal] trait ClientHelpersPlatform {
 
-  @nowarn // ("cat=unused"), but Scala 3 doesn't implement this category
+  @nowarn("msg=never used")
```
armanbilge, line 31: We need to backport this to 0.23.

Issue #5227, opened 2021-09-19
Ember Client WebSocket support


RaasAhsan:
Now that we have ember-server WebSocket support, it shouldn't be much of a lift on the backend to build client support for it. I anticipate a bit of work around user API

armanbilge:
Besides ember-client I'd be thrilled to create e.g. a dom-websockets module implementing this as well, behind a common/shared abstraction.

https://developer.mozilla.org/en-US/docs/Web/API/WebSocket

amesgen:
http4s-jdk-http-client has a [WebSocket client](https://jdk-http-client.http4s.org/latest/index.html#websocket-client). See [`WSClient.scala`](https://github.com/http4s/http4s-jdk-http-client/blob/main/core/src/main/scala/org/http4s/jdkhttpclient/WSClient.scala) for the traits. We have our own `WSFrame` ADT there, as well es a "low level" and a "high level" interface due to differences in whether pings are automatically responded to, but these were mainly ideas/explorations.

I would be very glad to see an interface in http4s, and will happily port http4s-jdk-http-client to it!

armanbilge:
Yes, I've taken a look at it on a couple occasions, I'd love to see that (or a similar) interface in http4s proper! I'm looking at the reference for [browser web sockets](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket) and it looks like it can be used to implement your high level interface.

I'm no expert on these things, but maybe if the high-level interface is more straight forward we can start by porting that to http4s-client?

RaasAhsan:
Sounds like a plan ! @amesgen whenever you have some time, could you open a branch in the http4s repo with the relevant interfaces based off `series/0.22` (or whatever makes sense) ? That gives us a place to talk feasibility and design, but also get started on the ember implementation to use as another reference.

Some items I anticipate discussion around:
- `Client`/`WsClient` decoupling
- Parity with existing WS server machinery

pvgoran:
As far as I see this is not completed. However, on the table in https://http4s.org/v0.23/docs/integrations.html "Ember" is said to have "Websocket Client". Is this an error in the documentation?

armanbilge:
@pvgoran woops, thanks for catching that, it is a mistake. A PR fixing would be much appreciated.

armanbilge:
This is proposed as a Google Summer of Code Project! https://typelevel.org/blog/2023/02/23/gsoc.html

danghieutrung:
[...]

PR #5520, opened 2021-10-30
Create WebSocket client API

armanbilge:
Part 1 of #5227: creating a shared interface/API for websocket clients. Also linking to https://github.com/http4s/http4s-dom/issues/6.

I've literally just copy-pasted this out of the http4s-jdk-http-client, to start the conversation :) I personally don't have any prior experience with websockets.

cc @amesgen, @RaasAhsan, @rpiaggio


client/src/main/scala/org/http4s/client/websocket/WSClient.scala
```diff
@@ -0,0 +1,179 @@
...
+  def send(wsf: WSDataFrame): F[Unit]
+
+  /** Send multiple websocket frames. Equivalent to multiple `send` calls, but at least as fast. */
+  def sendMany[G[_]: Foldable, A <: WSDataFrame](wsfs: G[A]): F[Unit]
+
+  /** A `Pipe` which sends websocket frames and emits a `()` for each chunk sent. */
+  final def sendPipe: Pipe[F, WSDataFrame, Unit] = _.chunks.evalMap(sendMany(_))
+
+  /** Send a Ping frame. */
+  def sendPing(data: ByteVector = ByteVector.empty): F[Unit]
```
armanbilge, line 93: IDK if it's too restrictive, but if possible I'd like the high-level API to be fully implementable with browser websockets:
https://developer.mozilla.org/en-US/docs/Web/API/WebSocket

I think this `sendPing` is the only method in the current interface that can't be implemented in browsers. Is it necessary? :)

rpiaggio: Thanks @armanbilge for getting the ball rolling on this.

I agree that we should have an interface that's fully implementable with browser websockets.

Maybe we should split the high level and low level interfaces?

armanbilge: Yes, so the current `WSClient` looks like this:
```scala
trait WSClient[F[_]] {
  def connect(request: WSRequest): Resource[F, WSConnection[F]]
  def connectHighLevel(request: WSRequest): Resource[F, WSConnectionHighLevel[F]]
}
```

The `WSConnectionHighLevel` I think is fully-implementable with browser websockets, except for that `sendPing` method I noted above.

So the question is what to do about the non-highlevel `connect` method. We could simply throw an error if tried to use in the browser. Or for additional type-safety we could split:
```scala
trait WSHighLevelClient[F[_]] {
  def connectHighLevel(request: WSRequest): Resource[F, WSConnectionHighLevel[F]]
}

trait WSClient[F[_]] extends WSHighLevelClient[F] {
  def connect(request: WSRequest): Resource[F, WSConnection[F]]
}
```

I'm not sure how these things get used by clients, but I'd even think maybe the distinction should be `WSClient` and `WSLowLevelClient` if most people prefer the high-level interface?

armanbilge:
@amesgen @RaasAhsan when you have a moment, I'd really appreciate your thoughts/concerns about the current interface. Doesn't have to be anything polished, a brain dump would be just fine, so I can start iterating on some of your ideas :)

client/src/main/scala/org/http4s/client/websocket/WSClient.scala
ChristopherDavenport: Does the browser ping by itself. How do people prevent themselves being killed as an inactive tcp session? You can have a ping data packet, but its... repetitive.

armanbilge: That's a good question, I don't know. Probably? Can't find any information about that.

armanbilge: https://www.w3.org/Bugs/Public/show_bug.cgi?id=13104

> We don't think it's necessary to include API support for ping/pong, which is similar to TCP-level-like keepalives that can cross intermediaries. They are used to notify anyone on the path to the server and the server itself that the client is still alive.
 
> It should be the job of the stack to keep your connection alive. The application may want to specify keepalive interval, but it should not need to handle this task itself.

RaasAhsan:
Sorry for being late , will take a look this weekend

amesgen:
Thanks for your work on this @armanbilge!
 - I am not sure if having both high-level and low-level interfaces was/is worth the complexity. As one datapoint to take into consideration, sttp (whose WebSocket is newer than http4s-jdk-http-client) only has one interface[`WebSocket.scala`](https://github.com/softwaremill/sttp-shared/blob/master/ws/src/main/scala/sttp/ws/WebSocket.scala, and they also have a JavaScript implementation.
 - In the end, I think it would be a shame if `WSFrame` would not be shared between server and client. I initially went with a different (abstract) `WSFrame` for http4s-jdk-http-client because the ADT in http4s-core for the server side is very intertwined with parsing the frames from their binary representation, and there were various slight differences in how different backends handle e.g. invalid status codes. I think it works nicely in http4s-jdk-http-client, but I am not sure in what extent it is worth breaking compatibility to replace `WebSocketFrame`.

armanbilge:
Thank you @amesgen for your comments, very helpful and much appreciated! I will look into those in more detail a bit later and see what I can come up with. Regarding compatibility ...

> but I am not sure in what extent it is worth breaking compatibility to replace WebSocketFrame.

That's probably a no-go for 0.22/0.23 but sounds like a must-have for 1.0. In theory we can do a deprecation/bincompat dance like was just done for `WebSocketBuilder` and `WebSocketBuilder2` to get it into 0.22/0.23 but not sure if that's worth the effort :)

armanbilge:
Sharing a little more research.

> As one datapoint to take into consideration, sttp (whose WebSocket is newer than http4s-jdk-http-client) only has one interface WebSocket.scala, and they also have a JavaScript implementation.

So I took a look at this and it seems what they've done is provided an API closer to our low-level one here and throw an error for unsupported operations on browsers:
```scala
case _: WebSocketFrame.Ping => monad.error(new UnsupportedOperationException("Ping is not supported in browsers"))
case _: WebSocketFrame.Pong => monad.error(new UnsupportedOperationException("Pong is not supported in browsers"))
```
https://github.com/softwaremill/sttp/blob/65dbbc478c3c94cdf498a9bc48c38dc74a8e639a/core/src/main/scalajs/sttp/client3/WebSocketImpl.scala#L45-L46

I mean ... this _is_ an option, but I'm really not a fan of this, since it trades type-safety for runtime surprises. Another option would be to just no-op them on JS, but then the docs/spec can't make any guarantee about these methods. E.g. Sending a `ping` becomes hint to the implementation to complete with "best-effort" 🙃.

> I am not sure if having both high-level and low-level interfaces was/is worth the complexity.

I do see your point here. @rpiaggio's use-case only makes use of the high-level operations but @ChristopherDavenport comment points out the importance of these low-level operations as well.

client/src/main/scala/org/http4s/client/websocket/WSClient.scala
```diff
@@ -0,0 +1,179 @@
...
+  final case class Binary(data: ByteVector, last: Boolean = true) extends WSDataFrame
+}
+
+trait WSConnection[F[_]] {
+
+  /** Send a single websocket frame. The sending side of this connection has to be open. */
+  def send(wsf: WSFrame): F[Unit]
+
+  /** Send multiple websocket frames. Equivalent to multiple `send` calls, but at least as fast. */
+  def sendMany[G[_]: Foldable, A <: WSFrame](wsfs: G[A]): F[Unit]
```
RaasAhsan, line 64: are there any websocket APIs out there which offer a specialized implementation for this kind of function? I could see it being useful if vectored I/O were available, but unfortunately we are far away from that. at the very least we could provide a default implementation here if it's worth keeping around

client/src/main/scala/org/http4s/client/websocket/WSClient.scala
```diff
@@ -0,0 +1,179 @@
...
+  final case class Ping(data: ByteVector) extends WSControlFrame
+  final case class Pong(data: ByteVector) extends WSControlFrame
+  final case class Text(data: String, last: Boolean = true) extends WSDataFrame
+  final case class Binary(data: ByteVector, last: Boolean = true) extends WSDataFrame
+}
+
+trait WSConnection[F[_]] {
+
+  /** Send a single websocket frame. The sending side of this connection has to be open. */
+  def send(wsf: WSFrame): F[Unit]
```
RaasAhsan, line 61: are we demanding any particular behavior around concurrency here? i.e. safe to call `send` concurrently from multiple threads

client/src/main/scala/org/http4s/client/websocket/WSClient.scala
```diff
@@ -0,0 +1,179 @@
...
+                  case WSFrame.Ping(data) if respondToPings => conn.send(WSFrame.Pong(data))
+                  case wsf: WSFrame.Close =>
+                    recvCloseFrame.complete(wsf) *> outputOpen.get.flatMap(conn.send(wsf).whenA(_))
+                  case _ => F.unit
+                }) >> (wsf match {
+                  case wsdf: WSDataFrame => OptionT.pure[F](wsdf)
+                  case _ => receiveDataFrame
+                })
+              }
+            def defrag(text: Chain[String], binary: ByteVector): OptionT[F, WSDataFrame] =
```
RaasAhsan, line 156: IIRC the existing websocket parser in http4s already defragments messages for you (check me on this). should this API expect the underlying `WSConnection` to do that already or not? we can easily influence that with blaze/ember, but other backends may already perform it by the time we get the frames

RaasAhsan (commented):
This is looking great I'll try to find some time and get an Ember implementation going this week, will follow up with updates

client/src/main/scala/org/http4s/client/websocket/WSClient.scala
armanbilge: Yes, I think that's just good practice, unless it is complex to implement?

client/src/main/scala/org/http4s/client/websocket/WSClient.scala
armanbilge: I think it's worth keeping, IIRC the fs2.io Socket interfaces provide a similar method. Not sure about specialized implementations, but 👍 to a default one.

armanbilge:
Thanks @RaasAhsan! What are your thoughts about sharing the `WebSocketFrame` in core vs using the separate `WSFrame` ADT defined here?

client/src/main/scala/org/http4s/client/websocket/WSClient.scala
amesgen: In jdk-http-client, sending a frame is not thread-safe, so I wrapped sending a frame in a semaphore. I got the idea to add `sendMany` mainly due to this (with the tiny tiny performance optimization that we only need only acquire/release-cycle for several frames), but it is also nice to be future-proof for other optimizations like vectored IO :+1:

RaasAhsan:
> What are your thoughts about sharing the WebSocketFrame in core vs using the separate WSFrame ADT defined here?

I think there are several options here. We obviously can't break compatibility on 0.22, so we'll keep that in mind.

1. Target series/0.22, figure out a way to unify the existing server WebSocket model and the proposed client WebSocket model in a non-breaking way. This would be really nice since only one model would be available, and we could reuse websocket handlers on the client-side. I'm not sure if this is possible.
2. Target series/0.22, tread forward keeping the server/client models decoupled. This will result in duplicate code being maintained and different models depending on if we're implementing a server or client. On 1.0, we can deprecate/delete the old model and implement servers against the new model.
3. Target series/0.22, deprecate the old model, and provide the proposed unified model which will target both clients and servers. Note that it would put the responsibility on backends to support one or both models. On series/1.x, delete the old model (or keep it around).
4. Target series/1.x, implement both clients and servers against the proposed unified model, and deprecate/delete the old model. We wouldn't be able to get a stable release out for this anytime soon, and I would hate to alienate 0.22/0.23 users (which includes me) but it reduces the collective maintenance burden.
5. Target series/0.22 and try to implement clients against the existing model. I don't really know what this looks like.

So we'll probably want to choose one of these paths (or some other path I didn't think about) depending on what we care about for maintenance, release schedule, API/user experience, etc. 1 would be nice, 2 and 3 both seem reasonable, 3 also seems reasonable, 4 means this wouldn't be useful to many people for a long time, and 5 is probably at least worth gauging.

armanbilge:
Thanks for writing that all out. I'd definitely like to get something into 0.22/0.23 because otherwise it would be pretty hard to use this interface from libraries that want something a bit more stable :)

Besides unification across server/client, I think (maybe even more important) is unification across the clients, so that backends can easily be swapped out. Ideally the jdk-client would adopt whatever we do here, so it _does_ need to appeal to @amesgen 😉 I haven't really implemented these things so I defer to your judgement ...

I agree, (1) is _nice_ may be a bin-compat dance full of compromises. I think (2) can slowly become (3) (i.e., we can deprecate the old model and start migrating the servers in follow-ups). (5) depends if @amesgen would go for it :)

I can make an attempt at (1), see what's possible there.

armanbilge:
@stephenjudkins reminded me of this PR today, which I'm sad got stalled. 

I think trying to unify the websocket models was too ambitious and we ended up with nothing :) I say let's proceed with @RaasAhsan's strategy (2). The current websocket model in jdk-http-client is not a bad one. We can try harder to unify the models in 1.0 where we're allowed to break compatibility.

I'm going to freshen up this PR and get it ready for review. @RaasAhsan do you still have time to worker on an ember implementation? The tricky bit is working around @ChristopherDavenport's ember h2 PR.

client/src/main/scala/org/http4s/client/websocket/WSClient.scala
```diff
@@ -0,0 +1,181 @@
...
+  *
+  * @param uri
+  *   The URI.
+  * @param headers
+  *   The headers to send. Put your `Sec-Websocket-Protocol` headers here if needed. Some websocket
+  *   clients reject other WS-specific headers.
+  * @param method
+  *   The method of the intial HTTP request. Ignored by some clients.
+  */
+final case class WSRequest(
```
isomarcte, line 44: Should we mask the constructor for this like we do with some of our other case classes, e.g. [RequestPrelude](https://github.com/http4s/http4s/blob/2abafd2b5b265a28b27a1c8d2cbd05dd40ec4f9e/core/shared/src/main/scala/org/http4s/RequestPrelude.scala#L31)

client/src/main/scala/org/http4s/client/websocket/WSClient.scala
```diff
@@ -0,0 +1,181 @@
...
+trait WSConnection[F[_]] {
+
+  /** Send a single websocket frame. The sending side of this connection has to be open. */
+  def send(wsf: WSFrame): F[Unit]
+
+  /** Send multiple websocket frames. Equivalent to multiple `send` calls, but at least as fast. */
+  def sendMany[G[_]: Foldable, A <: WSFrame](wsfs: G[A]): F[Unit]
+
+  /** A `Pipe` which sends websocket frames and emits a `()` for each chunk sent. */
+  final def sendPipe: Pipe[F, WSFrame, Unit] = _.chunks.evalMap(sendMany(_))
```
isomarcte, line 73: Nit: I find grouping the final methods after the abstract ones makes it easier to ready

client/src/main/scala/org/http4s/client/websocket/WSClient.scala
armanbilge:
```suggestion
final case class WSRequest private (
```

client/src/main/scala/org/http4s/client/websocket/WSClient.scala
armanbilge: Like, all the way at the end of the class? Personally I prefer the proximity to `sendMany`.

armanbilge:
Apologies @isomarcte, I missed your review! I appreciate you taking a look :)

client/src/main/scala/org/http4s/client/websocket/WSClient.scala
isomarcte: That won't really help bincompat with the unapply method still leaking, right?

armanbilge: Yeah, and the `copy` methods. I can de-case-class this, that's always the safest bet.

isomarcte: I personally like `private[this]` case classes in the companion object of a `sealed trait` like is done in RequestPrelude. It still lets the library authors access the copy methods and stuff.

armanbilge: Yeah, that was a helpful pointer to `RequestPrelude`. I liked it too so I copied the same pattern here.

armanbilge:
@RaasAhsan @amesgen when you have a chance, I'd appreciate your thoughts on whether we should move forward with this. Thanks!

amesgen (approved):
Looks very good, also :+1: on the bincompatification of the case classes (also see [here](https://github.com/http4s/http4s-jdk-http-client/pull/53#discussion_r309887740)). As long as unification of the WS frame ADTs is planned for the future, I am fine with keeping them separate, but I am happy to adapt either approach in jdk-http-client.

client/src/main/scala/org/http4s/client/websocket/WSClient.scala
```diff
@@ -0,0 +1,224 @@
...
+  final def receiveStream: Stream[F, WSDataFrame] = Stream.repeatEval(receive).unNoneTerminate
+
+  /** The negotiated subprotocol, if any. */
+  def subprocotol: Option[String]
+
+  /** The close frame, if available. */
+  def closeFrame: TryableDeferred[F, WSFrame.Close]
+}
+
+trait WSClientHighLevel[F[_]] {
```
armanbilge, line 156: I know the "high level" distinction is not so popular but it's an important distinction in practice between what we can support in the browser and what we can't.

rossabaker:
Is this ready to be engraved into 0.22 or should we wait for the next bus?  I haven't been following this one, but I trust the people who have been involved.

armanbilge:
I'd be thrilled to see this on the next train. The API here has been proven in the jdk-http-client. I was hoping to hear from @RaasAhsan but I'm not sure how available they are these days.

rossabaker:
Alright.  I'm planning to work on releases this afternoon, but expect we'll have another one in a week or two for H2 and the parsers that got stuck in the holiday doldrums.  I'll check here for a go or no-go.

armanbilge:
Oh yeah, that's fine then. This can tag along with H2 in the next release.

armanbilge:
I think this is good, and would be great to get it into the next release 👍

rossabaker:
I'm not a web socket domain expert, but I'm happy if @armanbilge and @amesgen are happy.  I'll plan to include this unless another web socket expert throws a flag.

rossabaker (approved):
This API looks good to me.  The `defrag` question is interesting to me, but seems like it could be changed after the fact?

armanbilge:
IDK if we can change it, but I think we can offer two builders to support each use-case.

client/src/main/scala/org/http4s/client/websocket/WSClient.scala
isomarcte: @armanbilge we might add a brief comment to both traits. I fear it may not be obvious to a casual reader why this distinction is made.

rossabaker:
Has anybody made a concrete implementation of this yet?  We could make everything `private[http4s]` so Ember and jdk-http-client can catch up, and then make it public once the abstraction is proven.

armanbilge:
The model here is very nearly a copy-pasta of what's in the jdk-http-client. However, some breathing room might be good for me to confirm that I can make it work in the browser.

rossabaker:
I don't think it does any good to release it publicly before there's a concrete implementation, but if we make it package-private, we can probably get at least two quickly and run another train.

armanbilge:
Yeah, that was a good idea, thanks. Sounds good 👍

Issue #5497, opened 2021-10-27
ScalaFix linters


bplommer:
I raised in #5484 the possibility of using scalafix to enforce naming conventions. Other rules that could be useful to improve consistency and reduce maintenance burden:

- [x] [organize imports](https://github.com/liancheng/scalafix-organize-imports) - this will introduce a lot of merge conflicts when introduced but should subsequently reduce them
- [x] Disallow use of fs2's sync compiler - #5536
- [ ] Disallow use of `throw` in parsers not prefixed with `unsafe`
- [ ] No non-final case classes (added but not yet linting at error level)
- [x] No final case objects
- [ ] Recurring headers must have a semigroup instance (unless they're called `Set-Cookie`)
- [ ] Contents of `internal` packages must be package-private
- [ ] case classes with private constructors leaking public copy constructors
- [ ] Replace `assert(a == b)` with `assertEquals(a, b)`, `assert(a != b)` with `assertNotEquals(a, b)`

PR #5521, opened 2021-10-31
Add custom scalafix rules

bplommer:
It's possible I've become intoxicated by the power of scalafix and gone too far into bikeshedding with some of these rules.

Relates to #5497


blaze-core/src/main/scala/org/http4s/blazecore/IdleTimeoutStage.scala
```diff
@@ -147,7 +147,7 @@ object IdleTimeoutStage {
 
   sealed trait State
   case object Disabled extends State
-  case class Enabled(timeoutTask: Runnable, cancel: Cancelable) extends State
+  final case class Enabled(timeoutTask: Runnable, cancel: Cancelable) extends State
```
armanbilge, line 150: I like these sorts of changes but aren't we running into bincompat issues?

bplommer: For package-private components we can add mima exclusions. For public APIs I'm adding scalafix exclusions with a note to change them for 1.0.

armanbilge: Never played with scalafix, but it would be cool if it could detect the private status of stuff :)

bplommer: That's possible. What would it achieve though? For non-final/sealed/private case classes it's just raising a linter error, not making any changes.

armanbilge: Ah, just linting, gotcha. Save the manual exclusion?

bplommer: We still want to disallow plain `case class` for new code though, no?

armanbilge: Right, of course. Sorry to be dense 😅

bplommer: Not at all! I might disable that rule for this PR though, to avoid having manual changes mixed up with the automatic ones.

ember-server/src/test/scala/org/http4s/ember/server/ConnectionSuite.scala
```diff
@@ -64,7 +64,7 @@ class ConnectionSuite extends Http4sSuite {
       .withRequestHeaderReceiveTimeout(headerTimeout)
       .build
 
-  case class TestClient(client: Socket[IO]) {
+  sealed case class TestClient(client: Socket[IO]) {
```
danicheg, line 67: I think that case classes should be `final` rather than `sealed`.
`sealed` is needed for exhaustiveness checking on its subclasses, I guess.

danicheg: Uh oh, that's a test suite. So my comment is non-blocking at all.

bplommer: Also, final case classes that are members of another class omit a compiler warning that I don’t fully understand.

blaze-core/src/main/scala/org/http4s/blazecore/IdleTimeoutStage.scala
bplommer: What do people think about adding `sealed` to public case classes though? If I understand correctly, that would be source-breaking but not binary-breaking.

armanbilge: In general I think we should shy away from source-breaking changes on 0.22/0.23. Also a sealed `case class` as a bincompat workaround for a `final` one seems strange and probably not worth it :)

armanbilge: (Btw, I think a more idiomatic way to accomplish the same goal might be a package-private constructor. But still would be source-breaking.)

bplommer:
I'm at a loss on why the rules can only be instantiated in Scala 2.12 - I guess it's something to do with them being run from sbt-scalafix and sbt only using scala 2.12, but I can't find anything documented about this 😕

bplommer:
> I'm at a loss on why the rules can only be instantiated in Scala 2.12 - I guess it's something to do with them being run from sbt-scalafix and sbt only using scala 2.12, but I can't find anything documented about this 😕

The "optional but recommended" setting at https://scalacenter.github.io/scalafix/docs/developers/local-rules.html#prerequisite seems to have fixed it for 2.13.

blaze-core/src/main/scala/org/http4s/blazecore/IdleTimeoutStage.scala
rossabaker: I don't think we should be adding MiMa exceptions for public API that breaks bincompat, but as noted, we permit more if we don't enforce it now.  Is there a way to annotate the legacy ones with something akin to `@nowarn`?

scalafix-internal/rules/src/main/scala-2/fix/UpstreamLibs.scala
```diff
@@ -0,0 +1,34 @@
...
+ * limitations under the License.
+ */
+
+package fix
+
+import scalafix.v1._
+
+import scala.meta._
+
+class UpstreamLibs extends SemanticRule("Http4sUpstreamLibs") {
```
rossabaker, line 23: These are not always equivalent, because `as(a)` is strict and `map(_ => a)` is lazy.  We had an optimization around that recently, though I don't remember it being on `Stream`.

Why target `Stream` specifically as opposed to other functors?  Just have to start somewhere?  I think the cost is one-per-stream instead of one-per-output, which helps.

blaze-core/src/main/scala/org/http4s/blazecore/IdleTimeoutStage.scala
bplommer: > I don't think we should be adding MiMa exceptions for public API that breaks bincompat

Agreed! I’d only want to do that for package-private APIs.

> Is there a way to annotate the legacy ones with something akin to @nowarn?

yes: https://scalacenter.github.io/scalafix/docs/users/suppression.html

scalafix-internal/rules/src/main/scala-2/fix/UpstreamLibs.scala
bplommer: Yeah, it was really just a low-hanging bike shed to start hacking on. Happy to remove that one.

rossabaker: I'm inclined toward no on this one, but yes to the rest.

rossabaker (approved):

PR #5525, opened 2021-11-01
Set end-year to the copyright

danicheg:
Is it make sense to set a range instead of a year of first publication only to the copyright?

### DIFF
--- a/async-http-client/src/main/scala/org/http4s/asynchttpclient/AsyncHttpClient.scala
+++ b/async-http-client/src/main/scala/org/http4s/asynchttpclient/AsyncHttpClient.scala
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 http4s.org
+ * Copyright 2016-2021 http4s.org
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/async-http-client/src/main/scala/org/http4s/asynchttpclient/AsyncHttpClientStats.scala
+++ b/async-http-client/src/main/scala/org/http4s/asynchttpclient/AsyncHttpClientStats.scala
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 http4s.org
+ * Copyright 2016-2021 http4s.org
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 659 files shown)


rossabaker:
I did some research on this when I set up the header plugin:

* You're not supposed to bump the end date unless you make material changes.  In some jurisdictions, copyright has a particular duration, and automatically bumping the date is artificially renewing it.
* Some sources say it should contain a range of years for every "significant" modification, but we don't have automation around that, to say nothing of judging "significance".
* The term might not even be a function of the publication date, but of the author's death, if we're in the Berne Convention, which I think all contributors are.
* But which author's death?  Our copyright is declared for what isn't even a legal entity.
* I can't imagine chasing down hundreds of contributors to change that entity declaration today.
* And countries are free to extend beyond the Berne Convention.  Since we're international, then what?

Typelevel doesn't have lawyers, so there's no guidance there.  The larger software foundations don't all agree.  I honestly don't know what OSS authors are supposed to do.  The status quo has the bonus of not generating a diff every January 1, but what you've done here keeps the code looking fresh, which is nice.

Does anyone have strong opinions, or better yet, minimal expertise?

danicheg:
@rossabaker thanks for the detailed answer.
I can only confirm that there are different opinions on this, even in typelevel projects is used different approaches. So, if it is legal to have the only first publication year declared - it's even more convenient then we shouldn't update it every 1 January :)

armanbilge:
My opinion is that ... any notion of copyright here is kind of nonsensical. Some words from Daniel on this topic in the Discord a little while back:
> yeah this is complicated, and it's a complexity that all OSS projects hit
the tldr is that you can't assign copyright to us even if Typelevel as a legal entity did exist in the US
because you're not being paid
so either you own the copyright, or your employer does, but we don't
so the copyright description is deceptive. what you can do, and what is in fact directly covered by the ASLv2 terms, is assign us a non-exclusive, unrestricted, royalty-free license to redistribute
this is what the copyright header is meant to convey
this kind of setup and contribution terms are basically ubiquitous in OSS, particularly since the rise of GitHub, but to my knowledge it hasn't been tested in court

https://discord.com/channels/632277896739946517/839257782993420328/882688985510604820

rossabaker:
Our header originates from the Apache license, and the ASF [recommends a copyright header](http://www.apache.org/foundation/license-faq.html#Apply-My-Software) "in case they become detached from the LICENSE file."  Their example has one year, which is not defined.  And the ASF's own code expressly forbids a copyright declaration in the header!

Eclipse does [${firstYear,} ${lastYear}](https://www.eclipse.org/projects/handbook/#ip-copyright-headers).  

Mozilla recommends [no copyright](https://www.mozilla.org/en-US/MPL/headers/).

I'm in favor of the status quo.  Among all the options, it's the one that came from the lawyers who wrote our license, and there's not a real persuasive case made for an alternative path.

danicheg:
Good. This PR can be a reference for possible future questions about the license.

PR #5534, opened 2021-11-03
Update scalafmt

bplommer:
Updates scalafmt to 3.0.8 and uses it for Scala 3.

I've tried to optimize for minimum upfront diff size - I added a few of commented-out settings in .scalafmt.conf that I think would be an improvement in the long run but will create a big upfront diff that I don't want to inflict on @rossabaker .


.scalafmt.conf
```diff
@@ -1,23 +1,45 @@
...
 maxColumn = 100
+docstrings.wrap = false
 
 // Vertical alignment is pretty, but leads to bigger diffs
 align.preset = none
 
+# Commented-out configs are Good Ideas but will create a big initial diff if applied
+
 danglingParentheses.preset = false
+# danglingParentheses.preset = true
```
rossabaker, line 14: This reduces diffs in the long run. :+1:

.scalafmt.conf
```diff
@@ -1,23 +1,45 @@
...
   AvoidInfix
   RedundantBraces
   RedundantParens
-  AsciiSortImports
   PreferCurlyFors
+  # SortModifiers
+]
+
+rewrite.sortModifiers.order = [
+  override, implicit, private, protected, final, sealed, abstract, lazy
```
rossabaker, line 25: How does this differ from the default?

.scalafmt.conf
```diff
@@ -1,23 +1,45 @@
...
 ]
 
+# rewrite.trailingCommas.style = multiple
```
rossabaker, line 28: This reduces diffs in the long run. :+1:

.scalafmt.conf
```diff
@@ -1,23 +1,45 @@
...
    "scalafix-inputs",
    "scalafix-outputs"
 ]
+
+runner.dialect = scala212
+
+fileOverride {
+  "glob:**/scala-3/**/*.scala" {
+    runner.dialect = scala3
+    # rewrite.scala3.preset = true
```
rossabaker, line 40: I don't know what this does.

.scalafmt.conf
```diff
@@ -1,23 +1,45 @@
-version = 2.7.5
+version = 3.0.8
 
 style = default
 
 maxColumn = 100
+docstrings.wrap = false
```
rossabaker, line 6: Is this for minimization of the current diff?

I haven't configured my editor to default to the new style, so I'm not that fond of it, but I'm fond of defaults.  I could go either way on this.

rossabaker (commented):
I can apply the changes in isolation by merging, keeping "ours" on all conflicts, and reformatting.  Initial diff size matters much less than minimizing conflicts moving forward.

.scalafmt.conf
bplommer: The default (leaving out scala 3 modifiers) is implicit final sealed abstract override private protected lazy. The order I configured minimises diff and I prefer it, but I'm not hugely opposed to keeping the default.

.scalafmt.conf
bplommer: `always` would reduce them even more but is terribly, terribly ugly.

.scalafmt.conf
bplommer: It removes optional braces and uses new control flow syntax.

.scalafmt.conf
bplommer: Mostly that, yeah. Enabling wrapping also broke some doctests, but I wasn't on the latest version when I tried it so I could try again.

.scalafmt.conf
bplommer: But we can't use it because `-source:3.0-migration` is enabled.

.scalafmt.conf
bplommer: Ah yes, a couple of doctests are failing to compile now.

.scalafmt.conf
rossabaker: You had me at minimizes diff.

.scalafmt.conf
rossabaker: Oh.  Well, good.  I don't like that new syntax anyway!

.scalafmt.conf
rossabaker: Okay, let's leave it false then.  I would love to see the number of doctests grow.

rossabaker (approved):
Oh lawd he comin'

PR #5536, opened 2021-11-04
Add scalafix linter for use of fs2 Sync compiler

bplommer:


bplommer:
The test failures are blaze flakes.

server/src/main/scala/org/http4s/server/middleware/ChunkAggregator.scala
```diff
@@ -37,7 +37,7 @@ object ChunkAggregator {
       http: Kleisli[F, A, Response[G]]): Kleisli[F, A, Response[G]] =
     http.flatMapF { response =>
       f(
-        response.body.chunks.compile.toVector
+        response.body.chunks.compile.toVector // scalafix:ok Http4sFs2Linters.noFs2SyncCompiler; grandfathered until 1.0
```
rossabaker, line 40:
```suggestion
        response.body.chunks.compile.toVector // scalafix:ok Http4sFs2Linters.noFs2SyncCompiler; bincompat until 1.0
```

This is not well known inside the United States and I wouldn't expect an non-American to know it, but "grandfathering" has a [racially charged etymology]( 
https://medium.com/@nriley/words-matter-why-we-should-put-an-end-to-grandfathering-8b19efe08b6a) many find discomforting.  I've sought alternatives since learning this.

rossabaker (commented):
This is a nice solution. :+1: after the wording change.

server/src/main/scala/org/http4s/server/middleware/ChunkAggregator.scala
bplommer: Yuck, good to know

PR #5541, opened 2021-11-07
Implement Headers#parse method

### LINKED ISSUE #5538: Header typeclass requires parser loses error information
The Header typeclass implementation in 0.23.x requires a parse function yielding an `Either[ParseFailure, A]` in construction. But when using the instance on a request (e.g. `req.headers.get[A]`) it returns an `Option[A]`, losing the parse error information.

Can we add a method on `Headers` which preserves the errors?

Discord discussion: https://discord.com/channels/632277896739946517/632286375311573032/905892715194110012

nasadorian:
Closes #5538


rossabaker (commented):
Thanks!

There is already a similar `getWithWarnings`, which returns an `Ior`.  The `Ior` handles the case of a repeating header where some, but not all, of the values are invalid.  Then we can return errors _and_ a best effort at a value.

But most developers are familiar with `Either`, and `Either` compositions are more common.  Would you still rather have this signature, after learning of `getWithWarnings`?

PR #5546, opened 2021-11-08
Draft GUnzip middleware

### LINKED ISSUE #4981: Gunzip Middleware
cc @ChristopherDavenport 

http4s has a `GZip` middleware which compresses responses as `gzip`, but does not have a `gunzip` dual.

I'm proposing to implement a Gunzip middleware using the `fs2` utilities already available for such compression.

Code looks roughly like below. Let me know if this is a worthwhile contribution!

```scala
  def gunzipWrapper(routes: HttpRoutes[IO]): HttpRoutes[IO] =
    Kleisli { req =>
      req.headers.get(`Accept-Encoding`) match {
        case Some(accept)
            if accept.satisfiedBy(ContentCoding.gzip) || accept.satisfiedBy(
              ContentCoding.`x-gzip`) =>
          val newBody = req.body.through(gunzip()) >>= (_.content)
          routes(req.withBodyStream(newBody))
        case None => routes(req)
      }
    }
```

nasadorian:
Draft addressing server side. Client side remains to be implemented. Needs tests.

Closes #4981

### DIFF
--- a/server/jvm/src/main/scala/org/http4s/server/middleware/GUnzip.scala
+++ b/server/jvm/src/main/scala/org/http4s/server/middleware/GUnzip.scala
@@ -1,0 +1,61 @@
+/*
+ * Copyright 2014 http4s.org
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.http4s
+package server
+package middleware
+
+import cats.data.Kleisli
+import cats.effect.Sync
+import cats.syntax.eq._
+import fs2.compression.Compression
+import fs2.compression.InflateParams
+import org.http4s.headers.`Content-Encoding`
+import org.http4s.headers.`Content-Length`
+
+object GUnzip {
+
+  def apply[F[_], G[_]: Sync](
+      http: Http[F, G],
+      inflateParams: InflateParams = InflateParams.DEFAULT): Http[F, G] =
+    Kleisli { (req: Request[G]) =>
+      req.headers.get[`Content-Encoding`] match {
+        case Some(ce) if isGzipped(ce.contentCoding) =>
+          http(gunzipRequest(req, inflateParams))
+        case _ => http(req)
+      }
+    }
+
+  private def isGzipped(cc: ContentCoding): Boolean =
+    cc === ContentCoding.gzip || cc === ContentCoding.`x-gzip`
+
+  private def gunzipRequest[F[_]: Sync](
+      req: Request[F],
+      inflateParams: InflateParams
+  ): Request[F] = {
+    val newBody = req.body.through(Compression[F].gunzip(inflateParams)).flatMap(_.content)
+    val newHeaders = req.removeHeader[`Content-Encoding`].putHeaders(`Content-Length`(69)).headers
+    Request[F](
+      method = req.method,
+      uri = req.uri,
+      httpVersion = req.httpVersion,
+      headers = newHeaders,
+      body = newBody,
+      attributes = req.attributes
+    )
+  }
+
+}


server/jvm/src/main/scala/org/http4s/server/middleware/GUnzip.scala
```diff
@@ -0,0 +1,61 @@
...
+
+  private def isGzipped(cc: ContentCoding): Boolean =
+    cc === ContentCoding.gzip || cc === ContentCoding.`x-gzip`
+
+  private def gunzipRequest[F[_]: Sync](
+      req: Request[F],
+      inflateParams: InflateParams
+  ): Request[F] = {
+    val newBody = req.body.through(Compression[F].gunzip(inflateParams)).flatMap(_.content)
+    val newHeaders = req.removeHeader[`Content-Encoding`].putHeaders(`Content-Length`(69)).headers
```
rossabaker, line 50: `Content-Length` definitely isn't hard coded, and I don't think we can even calculate it in advance of consuming (and inflating) the stream.  Removing it is correct.

We could add a `Transfer-Encoding: chunked` header, since we usually expect either a `Content-Length` or that header.  But the request didn't actually come in that way.  I could argue that one both ways.  Is there anybody who can argue it just one way?

rossabaker (commented):
Sorry for the slow review.  This looks like a good start.

wjoel:
Did something similar a while back. Feel free to pick and choose from this as you like:

```scala
import cats.data.Kleisli
import cats.effect._
import org.http4s._
import org.http4s.headers.`Content-Encoding`

object GZipRequestMiddleware {
  def apply[F[_]: Sync](service: HttpRoutes[F], bufferSize: Int = 100 * 1024): HttpRoutes[F] =
    Kleisli { (req: Request[F]) =>
      val req2 = req.headers.get[`Content-Encoding`] match {
        case Some(header) if satisfiedByGzip(header) =>
          val decoded = req.body
            .through(fs2.compression.gunzip[F](bufferSize))
            .flatMap(_.content)
            .handleErrorWith { e =>
              throw MalformedMessageBodyFailure(
                "Failed to decode gzippped request body",
                Some(e)
              )
            }
          req
            .removeHeader[`Content-Encoding`]
            .withEntity(decoded)(
              EntityEncoder.entityBodyEncoder
            ) // resolving implicit conflict
        case _ => req
      }
      service(req2)
    }

  private def satisfiedByGzip(header: `Content-Encoding`) =
    header.contentCoding.matches(ContentCoding.gzip) || header.contentCoding
      .matches(ContentCoding.`x-gzip`)
}
```

```scala
import cats.Applicative

import java.io.ByteArrayOutputStream
import java.util.zip.GZIPOutputStream
import cats.effect.IO
import org.http4s.{EntityEncoder, HttpRoutes, Request}
import org.http4s.dsl.io._
import org.http4s.implicits._
import org.scalatest.flatspec.AnyFlatSpec
import org.scalatest.matchers.should.Matchers

class GZipRequestMiddlewareSpec extends AnyFlatSpec with Matchers {
  private val routes = HttpRoutes.of[IO] {
    case req @ POST -> Root / "echo" =>
      Ok(req.body)(implicitly(Applicative[IO]), EntityEncoder.entityBodyEncoder)
  }
  private val app = GZipRequestMiddleware(routes).orNotFound

  private val testRequest = Request[IO](uri = uri"/echo", method = POST).withEntity("hello")(EntityEncoder.stringEncoder)

  it should "leave unzipped requests alone" in {
    val result = app.run(testRequest)
      .unsafeRunSync()
      .bodyText
      .compile
      .string
      .unsafeRunSync()

    result shouldBe "hello"
  }

  private val compressedHello = {
    val out = new ByteArrayOutputStream()
    val gzip = new GZIPOutputStream(out)
    gzip.write("hello".getBytes())
    gzip.close()
    out.toByteArray
  }

  it should "decompress zipped requests" in {
    val gzipRequest = testRequest.withEntity(compressedHello)(EntityEncoder.byteArrayEncoder)
      .withHeaders(Ok("OK", "Content-Encoding" -> "gzip").unsafeRunSync().headers)
    val result = app.run(gzipRequest)
      .unsafeRunSync()
      .bodyText
      .compile
      .string
      .unsafeRunSync()

    result shouldBe "hello"
  }

  it should "remove content-encoding" in {
    val gzipRequest = testRequest.withEntity(compressedHello)(EntityEncoder.byteArrayEncoder)
      .withHeaders(Ok("OK", "Content-Encoding" -> "gzip").unsafeRunSync().headers)
    // If content-encoding isn't filtered, this will fail as we attempt to decode the
    // request body twice.
    val appWrappedTwice = GZipRequestMiddleware(GZipRequestMiddleware(routes)).orNotFound
    val result = appWrappedTwice.run(gzipRequest)
      .unsafeRunSync()
      .bodyText
      .compile
      .string
      .unsafeRunSync()

    result shouldBe "hello"
  }
}
```

PR #5547, opened 2021-11-08
Structured Field Values for HTTP (RFC8941)

### LINKED ISSUE #5432: Implement RFC 8941
There are several PRs open (#5357, #5363, #5367, #5374) for headers that use syntax defined in [RFC 8941](https://datatracker.ietf.org/doc/html/rfc8941). Parsers and renderers for this syntax should be implemented once and reused.

m3is0:
Resolves #5432


core/src/main/scala/org/http4s/StructuredField.scala
```diff
@@ -0,0 +1,278 @@
...
+      new SfToken(s) {}
+
+    def fromString(s: String): Option[SfToken] =
+      parser.parseAll(s).toOption
+
+    val parser: Parser[SfToken] =
+      Rfc8941.sfToken.map(unsafeFromString)
+  }
+
+  sealed abstract case class SfBinary(value: String) extends BareItem {
```
bplommer, line 134: Shouldn't we wrap a byte array here (or something equivalent like `scodec.ByteVector`) rather than the base64 encoded value?

If we *do* store base64, I think we should definitely validate it fully when instantiating/parsing.

core/src/main/scala/org/http4s/internal/parsing/Rfc8941.scala
```diff
@@ -0,0 +1,169 @@
...
+  }
+
+  /* sf-string = DQUOTE *chr DQUOTE
+   * chr       = unescaped / escaped
+   * unescaped = %x20-21 / %x23-5B / %x5D-7E
+   * escaped   = "\" ( DQUOTE / "\" )
+   */
+  val sfString: Parser[String] = {
+    import Parser.{charIn, string}, Rfc5234.dquote
+    val escaped = string("\\\\") | string("\\\"")
```
bplommer, line 56: Shouldn't we unescape these when parsing and re-escape them when rendering?

core/src/main/scala/org/http4s/internal/parsing/Rfc8941.scala
```diff
@@ -0,0 +1,169 @@
...
+    val base64 = (alpha | digit | charIn("+/=".toList)).rep0.string
+    char(':') *> base64 <* char(':')
+  }
+
+  /* sf-boolean = "?" boolean
+   * boolean    = "0" / "1"
+   */
+  val sfBoolean: Parser[Boolean] = {
+    import Parser.string
+    string("?0").as(false) | string("?1").as(true)
```
bplommer, line 89: This is a micro-optimization but can this just parse `?` once? Something like `char('?') *> (string('0').as(false) | string('1').as(true))`

core/src/main/scala/org/http4s/internal/parsing/Rfc8941.scala
```diff
@@ -0,0 +1,169 @@
...
+  /* bare-item = sf-integer / sf-decimal / sf-string / sf-token
+   *           / sf-binary / sf-boolean
+   */
+  def bareItem[A](
+      sfIntegerP: Parser[A],
+      sfDecimalP: Parser[A],
+      sfStringP: Parser[A],
+      sfTokenP: Parser[A],
+      sfBinaryP: Parser[A],
+      sfBooleanP: Parser[A],
```
bplommer, line 101: Does this need the type parameter and parser parameters? Can't they be hard-coded?

m3is0: This way we can keep the RFC8941 parsers separate from our data model and avoid cyclic dependencies between them.

core/src/main/scala/org/http4s/internal/parsing/Rfc8941.scala
m3is0: Done.

core/src/main/scala/org/http4s/internal/parsing/Rfc8941.scala
m3is0: Done.

core/src/main/scala/org/http4s/internal/parsing/Rfc8941.scala
m3is0: Done.

bplommer:
Sorry for taking so long to get to this! It's purely down to my own laziness - I'll definitely give this and your other PR a proper review this side of Chrsitmas...

m3is0:
Ok, @bplommer, don't worry.

PR #5549, opened 2021-11-09
Fix or exclude scalafix warnings

### LINKED ISSUE #5537: Lint is full of warnings on series/0.22
I painstakingly fixed a bunch of these, but I guess it was on a different branch, and there are bincompat concerns.  But when I run quicklint on series/0.22, [warnings abound](https://github.com/http4s/http4s/runs/4105585792?check_suite_focus=true).  Are we just living with those?  `@nowarn`ing them away some how?

bplommer:
Resolves #5537

### DIFF
--- a/core/src/main/scala/org/http4s/internal/CharPredicate.scala
+++ b/core/src/main/scala/org/http4s/internal/CharPredicate.scala
@@ -149,7 +149,7 @@ object CharPredicate {
   private def unmaskable(c: Char) = c >= 128
 
   // efficient handling of 7bit-ASCII chars
-  case class MaskBased private[CharPredicate] (lowMask: Long, highMask: Long)
+  sealed case class MaskBased private[CharPredicate] (lowMask: Long, highMask: Long)
       extends CharPredicate {
 
     def apply(c: Char): Boolean = {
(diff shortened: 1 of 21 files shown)


core/src/main/scala/org/http4s/internal/CharPredicate.scala
```diff
@@ -149,7 +149,7 @@ object CharPredicate {
   private def unmaskable(c: Char) = c >= 128
 
   // efficient handling of 7bit-ASCII chars
-  case class MaskBased private[CharPredicate] (lowMask: Long, highMask: Long)
+  sealed case class MaskBased private[CharPredicate] (lowMask: Long, highMask: Long)
```
danicheg, line 152: I'm still not convinced about having `sealed` case classes. They shouldn't be extendable at all, I believe.

danicheg (commented):
Thanks for that work!

core/src/main/scala/org/http4s/internal/CharPredicate.scala
bplommer: Right. We can make this final since the private constructor already makes it effectively final.

danicheg (approved):

rossabaker (approved):

PR #5550, opened 2021-11-09
Add further scalafix rules

bplommer:
Adds linters for leaking sealed hierarchies (i.e. sealed traits or classes with open descendants) and case classes with non-validating copy constructors.

Also includes some other refinements:
- Enabled scalafix's built in `LeakingImplicitClassVal` rule (didn't catch anything)
- Run all rules on the second scalafix pass in the lint aliases - otherwise scalafix emits spurious warnings about unneeded exclusions
- rewrite all final objects, not just final case objects


rossabaker:
I like `nonValidatingCopyConstructor` a lot!

rossabaker (approved):

bplommer:
Might be time to think about contributing some of these upstream...

PR #5552, opened 2021-11-09
Send multiple chunks when testing chunked requests

RafalSumislawski:
In it's current shape the "POST a chunked body" test sends a request with just one chunk (which may be transferred together with the head of the request). That means that the test provides poor coverage of the implementation of chunked encoding. In order to improve the coverage I changed the request body to contain multiple chunks (implemented similarly to the `/chunked` in the `GetRoutes`).


rossabaker (approved):
Good find.

PR #5556, opened 2021-11-10
Add the fast path for `Uri.decode` implementation

plokhotnyuk:
This change just adds the fast path that greatly drops CPU usage in `Uri.decode` if there no `%` characters in the path to decode.

Could it be also back ported down to 22.x version?


armanbilge:
> Could it be also back ported down to 22.x version?

Would you mind targeting your PR at 0.22? Then we'll merge it up through 0.23 into main :)

rossabaker:
I can cherry-pick it.

rossabaker:
This will be in the next 0.22 release.

PR #5557, opened 2021-11-10
Non-trivial merge into 0.23

rossabaker:


rossabaker:
Getting a bunch of warnings like this, which I don't think are true?

```
[warn] /Users/ross.baker/src/http4s/client/shared/src/main/scala/org/http4s/client/oauth1/ProtocolParameter.scala:36:19: warning: [UnusedScalafixSuppression] Unused Scalafix suppression, this can be removed
[warn]   // scalafix:off Http4sGeneralLinters.noCaseClassWithoutAccessModifier; bincompat until 1.0
[warn]                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

rossabaker:
Oh, that might be #5550?

> Run all rules on the second scalafix pass in the lint aliases - otherwise scalafix emits spurious warnings about unneeded exclusions

PR #5563, opened 2021-11-12
Cross most of server (but `Server`) for JS on 0.23

armanbilge:
Yes, yes, I know, all the crossing was supposed to be over 🙃 

I thought we didn't really need this on JS and also it would be painful due to the JVMisms at the heart of `Server`. Well, we _don't_ need `Server`, but many of those middlewares are useful for serverless. Also client depends on server for testing some stuff. Also missing this JS cross makes it annoying to cross e.g. [natchez-http4s](https://github.com/tpolecat/natchez-http4s) which depends on both server and client in the same artifact.

So this is a quickie cross of mostly the middlewares. `Server` itself and static file stuff are left JVM-only for now, and no actual server implementation is crossed.


rossabaker (approved):
What makes CSRF and GZip not cross?

armanbilge:
I've been slowly shaving the Gzip yak. Delegating to fs2 was a big improvement. In fs2 we have compression on JS, we have CRCs on JS, the only thing left is to port the methods that encode/decode the headers/trailers. It's very doable, all the necessary pieces are there. Just a yak.

CSRF is actually crossed in main. In 0.23 it's entangled with `javax.crypto.SecretKey`. We could cross it with the shim strategy/hack I used in core, but I didn't think it was worth it.

PR #5564, opened 2021-11-12
Add DefaultQueryParamDecoderMatcher class

sbly:
Hello! First-time contributor here. This PR adds a `DefaultQueryParamDecoderMatcher`, which, as the name suggests, accepts a default value for a query parameter. I followed the pattern for the other matchers. Let me know if there's anything I missed!

### DIFF
--- a/dsl/src/main/scala/org/http4s/dsl/impl/Path.scala
+++ b/dsl/src/main/scala/org/http4s/dsl/impl/Path.scala
@@ -326,6 +326,9 @@ abstract class OptionalQueryParamDecoderMatcher[T: QueryParamDecoder](name: Stri
       .toOption
 }
 
+/** A param extractor with a default value. If the query param is not present, the default value is returned
+  * If the query param is present but incorrectly formatted, will return `None`
+  */
 abstract class QueryParamDecoderMatcherWithDefault[T: QueryParamDecoder](name: String, default: T) {
   def unapply(params: Map[String, collection.Seq[String]]): Option[T] =
     params
--- a/dsl/src/main/scala/org/http4s/dsl/impl/Path.scala
+++ b/dsl/src/main/scala/org/http4s/dsl/impl/Path.scala
@@ -326,7 +326,7 @@ abstract class OptionalQueryParamDecoderMatcher[T: QueryParamDecoder](name: Stri
       .toOption
 }
 
-abstract class DefaultQueryParamDecoderMatcher[T: QueryParamDecoder](name: String, default: T) {
+abstract class QueryParamDecoderMatcherWithDefault[T: QueryParamDecoder](name: String, default: T) {
   def unapply(params: Map[String, collection.Seq[String]]): Option[T] =
     params
       .get(name)
@@ -336,8 +336,8 @@ abstract class DefaultQueryParamDecoderMatcher[T: QueryParamDecoder](name: Strin
       .map(_.getOrElse(default))
 }
 
-abstract class DefaultQueryParamMatcher[T: QueryParamDecoder: QueryParam](default: T)
-    extends DefaultQueryParamDecoderMatcher[T](QueryParam[T].key.value, default)
+abstract class QueryParamMatcherWithDefault[T: QueryParamDecoder: QueryParam](default: T)
+    extends QueryParamDecoderMatcherWithDefault[T](QueryParam[T].key.value, default)
 
 /** Flag (value-less) query param extractor
   */
--- a/dsl/src/main/scala/org/http4s/dsl/impl/Path.scala
+++ b/dsl/src/main/scala/org/http4s/dsl/impl/Path.scala
@@ -326,6 +326,19 @@ abstract class OptionalQueryParamDecoderMatcher[T: QueryParamDecoder](name: Stri
       .toOption
 }
 
+abstract class DefaultQueryParamDecoderMatcher[T: QueryParamDecoder](name: String, default: T) {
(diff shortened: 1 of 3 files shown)


armanbilge (approved):
LGTM, nice work following the existing patterns :)

sbly:
@armanbilge Thanks! Are you able to merge? It seems one CI run was unsuccessful, but looking at the error, it doesn't seem to be related to my PR. Maybe it was just a fluke?

armanbilge:
Yes, don't worry about CI, just a flake :) I was hoping to get one more pair of eyes before merging, let's give someone a nudge 😉

armanbilge:
Oh actually, @rossabaker will cherry-pick this into 0.22 at the next opportunity. Unless you want to retarget against series/0.22?

dsl/src/main/scala/org/http4s/dsl/impl/Path.scala
```diff
@@ -326,6 +326,19 @@ abstract class OptionalQueryParamDecoderMatcher[T: QueryParamDecoder](name: Stri
       .toOption
 }
 
+abstract class DefaultQueryParamDecoderMatcher[T: QueryParamDecoder](name: String, default: T) {
+  def unapply(params: Map[String, collection.Seq[String]]): Option[T] =
```
bplommer, line 330: I think this should return `Some` rather than `Option`?

armanbilge: Are you sure? What if `params.get(name)` is `None`?

dsl/src/main/scala/org/http4s/dsl/impl/Path.scala
```diff
@@ -326,6 +326,19 @@ abstract class OptionalQueryParamDecoderMatcher[T: QueryParamDecoder](name: Stri
       .toOption
 }
 
+abstract class DefaultQueryParamDecoderMatcher[T: QueryParamDecoder](name: String, default: T) {
```
bplommer, line 329: Would `QueryParamDecoderMatcherWithDefault` be a better name? We're not deciding a default query peram.

bplommer:
I like the idea, but is there a more compositional way we can do this so that defaults can be applied to arbitrary optional matchers? The answer may be no, given the requirement for stable identifiers.

dsl/src/main/scala/org/http4s/dsl/impl/Path.scala
bplommer: Isn't that what the default is for?

Sounds like this should have a doc to spell out the semantics :)

armanbilge: Haha yes I derped there 😆 good point.

bplommer: Does it also use the default if there's a value but it fails to parse? (I'm having one of those days where my code-reading comprehension is extremely minimal.) I think that should be documented either way.

armanbilge: But I remember looking at the tests and they seemed to do the right thing. So now I am confused.

bplommer: If it still compiles after changing the `Option` to `Some` then the behaviour is probably correct.

armanbilge:
> is there a more compositional way we can do this so that defaults can be applied to arbitrary optional matchers?

This sounds cool, but what would it compose with?

bplommer:
Possibly a trait `DefaultParamValue` that can be mixed into other matcher classes?

bplommer:
> > is there a more compositional way we can do this so that defaults can be applied to arbitrary optional matchers?
> 
> 
> 
> This sounds cool, but what would it compose with?

An underlying non-total matcher.

Oh, did my other comment only just send?

sbly:
@bplommer Initially I tried to extend `OptionalQueryParamDecoderMatcher` and override the `unapply` method, but this didn't work because they have different return types: `Option[Option[T]]` vs `Option[T]`. I don't think we can extend `QueryParamDecoderMatcher` either because at that point we've already swallowed the error. Or did you mean something else?

dsl/src/main/scala/org/http4s/dsl/impl/Path.scala
sbly: It's `Option` because we still return `None` in the case that the parameter is present but with an incorrect format, eg as in one of the test cases. However I can change this behavior to return the default in that case if you prefer that

sbly: Haha no problem, I'll add some docs

bplommer: I was going to say it should be changed, but then I saw the optional decoder returns `None` on the outer option in the case of a decoding failure. I disagree with that behaviour but I think for consistency we should do the same here. I'd like to reconsider that before 1.0 but I don't think we should hold this PR up on account of that.

bplommer (approved):

dsl/src/main/scala/org/http4s/dsl/impl/Path.scala
armanbilge: Interesting, look forward to hear your thoughts on that :)

rossabaker (commented):
Thanks, @sbly.  As mentioned above, I'll cherry-pick this back so it can be used starting in 0.22.

PR #5577, opened 2021-11-13
Use an instance of `Arbitrary[Year]` from ScalaCheck

danicheg:
A custom instance is not needed anymore.


armanbilge (approved):
Btw, all of these date/time instances are only in ScalaCheck for JVM. I started a PR to add them for JS as well but the problem is that it creates a circular dependency. Update: I looked back at my scalacheck PR and I think we can do it without the circular dependency :)

rossabaker:
If they're only on the JVM, what is this going to do when we merge it to CE3?

armanbilge:
These tests have been JVM-only on series/0.23 and main b/c of this. But hopefully we can fix that with the next scalacheck release.

PR #5578, opened 2021-11-14
Add sbt check for misplaced sources

armanbilge:
This mistake happens all-too-often when merging new files from 0.22 into crossed-modules in 0.23/main. A recent example:
https://github.com/http4s/http4s/tree/series/0.23/core

This adds an sbt check for these misplaced sources and will fail test/CI if they are detected. This first commit will fail as a demonstration.


armanbilge:
Example failure:
https://github.com/http4s/http4s/runs/4201570455?check_suite_focus=true#step:12:4352
```
 [error] (core / Test / test) Misplaced sources in /home/runner/work/http4s/http4s/core/src
```

rossabaker:
Nice.  I've tried to be mindful of this on every merge, and that's obviously not enough.

Issue #2012, opened 2018-08-19
Replace our URI


rossabaker:
Our `Uri` type is bad and I feel bad.

- It doesn't round trip (#1651)
- It's inconsistent which of its components are in the companion.
- The authority hierarchy is staggeringly complex.
- Only literals interpolate.  We could learn much from sttp.

I think this needs to be redesigned from the ground up, with deprecation warnings where reasonable.  This is a big job, but I think it's the single most hostile part of http4s.

ashwinbhaskar:
@rossabaker I can get started with this.

rossabaker:
This should probably be converted into a project.  Going through the backlog, I've already flagged a half dozen related issues, and I'm only about 25% of the way through.

PR #5586, opened 2021-11-16
Use `Arbitrary[Uri]` in some tests

danicheg:
A bit richer tests of the `Uri` parsing.

### DIFF
--- a/laws/src/main/scala/org/http4s/laws/discipline/arbitrary.scala
+++ b/laws/src/main/scala/org/http4s/laws/discipline/arbitrary.scala
@@ -790,9 +790,12 @@ private[discipline] trait ArbitraryInstances { this: ArbitraryInstancesBinCompat
     val genPathNoScheme = genSegmentNzNc |+| genPathAbEmpty
     val genPathAbsolute = const("/") |+| opt(genPathRootless)
 
-    oneOf(genPathAbEmpty, genPathAbsolute, genPathNoScheme, genPathRootless, genPathEmpty).map(
-      Uri.Path.unsafeFromString
-    )
+    oneOf(genPathAbEmpty, genPathAbsolute, genPathNoScheme, genPathRootless, genPathEmpty).map {
+      str =>
+        val preparedPath = if (str.startsWith("/")) str else "/" + str
+
+        Uri.Path.unsafeFromString(preparedPath)
+    }
   }
 
   implicit val http4sTestingCogenForPath: Cogen[Uri.Path] =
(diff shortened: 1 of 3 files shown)


danicheg (commented):
I've reflected a bit on this, probably it's better to wait for #2012 than try to work around the `Uri` roundtrip failure. So, I leave this on a judge of reviewers :)

armanbilge:
The arbitrary should really live here 😉 
https://github.com/typelevel/cats-uri/

laws/src/main/scala/org/http4s/laws/discipline/arbitrary.scala
```diff
@@ -815,6 +815,24 @@ private[discipline] trait ArbitraryInstances { this: ArbitraryInstancesBinCompat
     } yield Uri(scheme, authority, path, query, fragment)
   }
 
+  /** Creates an `Arbitrary[Url]` with checking that a URI is converting to String
+    * and back to URI safely.
+    * Use this Arbitrary with cautions - it may lead to tests performance degradation.
+    */
+  def createArbitraryUri: Arbitrary[Uri] =
```
rossabaker, line 822: Isn't a non-implicit `Arbitrary` just a `Gen`?

rossabaker (commented):
I'd like to eliminate the hack, but it's a big job.  The problem is well commented.  It's a better test than we have.  I lean toward moving forward with it.

laws/src/main/scala/org/http4s/laws/discipline/arbitrary.scala
armanbilge: Not necessarily. If you make it a `Gen`, it becomes annoying to use it:
https://github.com/http4s/http4s/blob/6ff6a74f877a81cd9fb59eaf2fa2d2412d4923e1/play-json/src/test/scala/org/http4s/play/PlaySuite.scala#L67

armanbilge: IIUC, the reason to keep it non-implicit is not so it doesn't get used implicitly, but so that doing so is opt-in.

danicheg: I agree with @armanbilge. If we would use a `Gen` here then we need to wrap it into an `Arbitrary` everywhere. And at the moment don't see how we can reuse it. As well as we shouldn't have this `Arbitrary` as an implicit value in the public API because of its 'hacking' nature, I believe.

rossabaker: But can't you call a `Gen` as `forAll(genUri) { uri => ... }` and not have to make it locally implicit?  

I strongly agree with not making it implicit in the `arbitrary` object because it's hacky.  Having agreed on that, I don't think the `Gen` vs. `Arbitrary` matters much either way.

rossabaker (approved):

laws/src/main/scala/org/http4s/laws/discipline/arbitrary.scala
danicheg: @rossabaker oh, sorry, I didn't get it the first time. Your suggestion is pretty nice. Then we will not deal with that tricky implicit `Arbitrary[Uri]` at all.

rossabaker:
> https://github.com/typelevel/cats-uri/

I haven't committed to making that happen yet, mainly because I'm concerned about Yet Another Import.

PR #5587, opened 2021-11-16
Netty based ServerScaffold (0.23)

RafalSumislawski:
This PR is an attempt reimplement the `ServerScaffold` as discussed in #5319. He main goals are to implement it in a way which will:
- not interfere with upgrading servlet / jetty  (this was the main drawback of jetty based scaffold, and a reason why it was replaced in 0.23)
- give us access to technical details of the communication (like jetty did, and Sun HTTP server didn't) enabling us to write assertions checking: number of connections established (actually implemented in this PR and needed in #5319), chunking of entity (just an idea), etc.
- be similar in 0.22, 0.23, 1.0
- if everything goes well, be fast and therefore usable not only for unit-tests but also for benchmarking (detection of performance regressions, etc.)

I focused on replacing the `ServerScaffold` with as little modifications to the surrounding code as possible and just enough implementation to make all the tests pass. This thing grew big nonetheless. 

I think it's worth to mention that there are three ways to create the Scaffold. This enables us to choose high-level, easy to use API for simple tests, and to peel layers of abstraction when we need access to low-level details.

When/If this gets a positive review. I will backport it to 0.22. And then adjust the #5319


client/js/src/test/scala/org/http4s/client/scaffold/ServerScaffold.scala
```diff
@@ -16,6 +16,23 @@
 
 package org.http4s
 package client
+package scaffold
+
+/*
+ * Copyright 2014 http4s.org
```
danicheg, line 22: Hmm, we already have a header in this file. Probably this came after moving file?

client/jvm/src/test/scala/org/http4s/client/scaffold/HandlersToNettyAdapter.scala
```diff
@@ -0,0 +1,158 @@
...
+      @deprecated("unused", "") request: HttpRequest,
+  ): Unit = ()
+
+  def onContent(
+      @deprecated("unused", "") ctx: ChannelHandlerContext,
+      @deprecated("unused", "") request: HttpRequest,
+      @deprecated("unused", "") content: HttpContent,
+  ): Unit = ()
+
+  def onRequestEnd(ctx: ChannelHandlerContext, request: HttpRequest): Unit
```
rossabaker, line 118: Is there any particular reason this one didn't have a default implementation, too?

rossabaker (approved):
I'm not a Netty expert, but this looks reasonable, and I am happy with the strategy.

@hamnis has done the most work integrating http4s with netty and may have some thoughts, but if he's busy we don't need to hold this: it's just test code, and whatever we got wrong can be fixed more easily than main code.

client/js/src/test/scala/org/http4s/client/scaffold/ServerScaffold.scala
RafalSumislawski: Thanks. Corrected.

client/jvm/src/test/scala/org/http4s/client/scaffold/HandlersToNettyAdapter.scala
RafalSumislawski: The idea was to make this trait Single-Abstract-Method so that in simple cases it can be implemented with the lambda expression syntax. The simple cases ended up being implemented with `RoutesToNettyAdapter` so didn't use this capability, but I still think it makes sense to have it like this. 

There is one place where I could actually use it. The '/process-request-entity'. But I opted for using the anonymous class syntax to stay consistent with `/respond-and-close-immediately` and `/respond-and-close-immediately-no-body`.

armanbilge (approved):
I'm excited about this :) I'm hopeful we can use this to create a client-testkit module. I can extract it out in a followup PR after it merges.

rossabaker (approved):

PR #5591, opened 2021-11-17
Use `assertEquals` instead of `assert`

danicheg:
This is pretty much a minor thing. However, we are using `MUnit` currently and can get nice diffs on assertions fails for free. I'm convinced `assert(foo == bar)` is almost anti-pattern and using `assertEquals(foo, bar)` is much better. If we agree on this, I'll add this point to the contributing guide. Perhaps, Scalafix-ninja may create a rewriting rule for this.
Also, we can replace `assert(boolExpr)` with `assert(boolExpr, true)`. But one may argue that this looks odd. I made it in several test classes to give you demonstrative examples. I think that nice diffs is worth the hassle anyway.

### DIFF
--- a/tests/src/test/scala/org/http4s/DecodeSpec.scala
+++ b/tests/src/test/scala/org/http4s/DecodeSpec.scala
@@ -95,20 +95,14 @@ class DecodeSpec extends Http4sSuite {
       // Not a valid first byte in UTF-8
       val source = Stream(0x80.toByte)
       val decoded = source.through(decode[Fallible](Charset.`UTF-8`)).compile.string
-      assert(decoded match {
-        case Left(_: MalformedInputException) => true
-        case _ => false
-      })
+      val Left(_: MalformedInputException) = decoded
     }
 
     test("decode should handle incomplete input") {
       // Only the first byte of a two-byte UTF-8 sequence
       val source = Stream(0xc2.toByte)
       val decoded = source.through(decode[Fallible](Charset.`UTF-8`)).compile.string
-      assert(decoded match {
-        case Left(_: MalformedInputException) => true
-        case _ => false
-      })
+      val Left(_: MalformedInputException) = decoded
     }
 
     test("decode should handle unmappable character") {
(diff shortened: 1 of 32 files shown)


rossabaker (commented):
Strong agree on `assertEquals(a, b)` over `assert(a == b)`.

I'm less convinced on the booleans.  They _must_ have returned false or thrown an exception we'll see anyway.  I guess there's a nice consistency to it, but I might be annoyed to fail a lint on it.  Here's the difference:

```
scala> assertEquals(false, true)
munit.ComparisonFailException: values are not the same
=> Obtained
false
-false
+true
  at munit.Assertions.failComparison(Assertions.scala:297)

scala> assert(false)
munit.FailException: assertion failed
```

danicheg:
@rossabaker pretty reasonable point about using `assertEquals` for checking boolean expressions, so I've reverted these changes. This PR is ready for review.

bplommer:
> Perhaps, Scalafix-ninja may create a rewriting rule for this.

Sure, add it to the list on the ticket 😄

tests/src/test/scala/org/http4s/DecodeSpec.scala
```diff
@@ -95,50 +95,38 @@ class DecodeSpec extends Http4sSuite {
       // Not a valid first byte in UTF-8
       val source = Stream(0x80.toByte)
       val decoded = source.through(decode[Fallible](Charset.`UTF-8`)).compile.string
-      assert(decoded match {
-        case Left(_: MalformedInputException) => true
-        case _ => false
-      })
+      val Left(_: MalformedInputException) = decoded
```
bplommer, line 98: This no longer has any assertion - what would a failure look like?

danicheg: If matching fails the failure is `scala.MatchError: Right(_)(of class scala.util.Right)`.

rossabaker (approved):

danicheg:
@bplommer do you have some points about this? Or can we move on?

bplommer:
> @bplommer do you have some points about this? Or can we move on?

Go for it

bplommer (approved):

PR #5595, opened 2021-11-19
Remove redundant isScala3 from build.sbt

bplommer:


danicheg (approved):

bplommer:
This is passing the scalafmt check locally but failing in CI 🤔 I'll try to fix the scalafmt update PR and see if that solves this.

PR #5601, opened 2021-11-21
 Netty based ServerScaffold (0.22) 

RafalSumislawski:
This is a backport of #5587 to `series/0.22`


rossabaker (approved):

client/src/test/scala/org/http4s/client/scaffold/package.scala
```diff
@@ -0,0 +1,53 @@
...
+package org.http4s.client
+
+import cats.effect.Async
+import cats.implicits._
+import io.netty.channel.Channel
+import io.netty.channel.ChannelFuture
+
+package object scaffold {
+
+  implicit class NettyChannelFutureSyntax[F[_]](private val fcf: F[ChannelFuture]) extends AnyVal {
```
danicheg, line 26: In the variant for `0.23` also there is manual cancellation of the `ChannelFuture` with `cancel(true)` (https://github.com/http4s/http4s/pull/5587/files#diff-23445af51d890997454c978e3ab1c2e9c7ae319fa2b49873807b311a8466c009R36). Could you please elaborate on why need cancellation for the `0.23`? IIUC, the cancellation will be synchronous, so maybe we can eliminate it in `0.23` as well? I'm not a `netty` expert, to be honest.

client/src/test/scala/org/http4s/client/scaffold/package.scala
```diff
@@ -0,0 +1,53 @@
...
+          cf.addListener { (f: ChannelFuture) =>
+            if (f.isSuccess) callback(Right(f.channel()))
+            else callback(Left(f.cause()))
+          }
+          ()
+        }
+      )
+  }
+
+  implicit class NettyFutureSyntax[F[_], A <: io.netty.util.concurrent.Future[_]](
```
danicheg, line 39: In the `0.23` one it has manual cancellation of `Future` `https://github.com/http4s/http4s/pull/5587/files#diff-23445af51d890997454c978e3ab1c2e9c7ae319fa2b49873807b311a8466c009R55`

client/src/test/scala/org/http4s/client/scaffold/package.scala
RafalSumislawski: IMO we don't really need the cancellation. The only reason why I implemented it in 0.23 is that the signature of `async` in CE3 kind of asks for it. As far as I can tell the cancellation is not synchronous. Nonetheless it increases the overall complexity, and may behave a bit differently that one would expect of a CE3 cancellation, so I'm for removing it.

rossabaker:
Nothing is blocking this, right?  It would be nice to merge it before it gets stale again.

RafalSumislawski:
I see no blockers.

PR #5611, opened 2021-11-23
`Access-Control-Allow-Origin` header

danicheg:
In the "0.22.x" and "0.23.x" series `Origin` header still supports multiple origins, so that's why the `Access-Control-Allow-Origin` header does it as well. That behavior has been changed in the 1.0 series (https://github.com/http4s/http4s/pull/5082). And if we'll agree on this one, I port the same semantic for the `Access-Control-Allow-Origin` header for "1.0" separately.

### DIFF
--- a/core/src/main/scala/org/http4s/headers/Origin.scala
+++ b/core/src/main/scala/org/http4s/headers/Origin.scala
@@ -78,11 +78,14 @@ object Origin {
     }
   }
 
-  private[http4s] val parser: Parser0[Origin] = {
-    import Parser.char
+  private[http4s] val hostListParser: Parser[NonEmptyList[Origin.Host]] = {
+    import cats.parse.Rfc5234.sp
 
-    nullHostParser.orElse(singleHostParser.repSep(char(' ')).map(hosts => Origin.HostList(hosts)))
+    singleHostParser.repSep(sp)
   }
+
+  private[http4s] val parser: Parser0[Origin] =
+    nullHostParser.orElse(hostListParser.map(hosts => Origin.HostList(hosts)))
 
   def parse(s: String): ParseResult[Origin] =
     ParseResult.fromParser(parser, "Invalid Origin header")(s)
--- a/core/src/main/scala/org/http4s/headers/Origin.scala
+++ b/core/src/main/scala/org/http4s/headers/Origin.scala
@@ -50,7 +50,13 @@ object Origin {
       toUri.render(writer)
   }
 
-  private[http4s] val parser: Parser0[Origin] = {
+  private[http4s] val nullHostParser: Parser0[Origin.Null.type] = {
+    import Parser.{`end`, string}
+
+    (string("null") *> `end`).as(Origin.Null)
+  }
+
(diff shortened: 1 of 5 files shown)


server/src/main/scala/org/http4s/server/middleware/CORS.scala
```diff
@@ -237,14 +237,9 @@ object CORS {
...
-            ),
-            // TODO model me
-            "Access-Control-Allow-Origin" -> origin.value,
-            // TODO model me
-            "Access-Control-Max-Age" -> config.maxAge.toSeconds.toString,
+            `Access-Control-Allow-Methods`(config.allowedMethods.getOrElse(Set(method))),
+            `Access-Control-Allow-Origin`.fromOrigin(origin),
+            `Access-Control-Max-Age`.unsafeFromDuration(config.maxAge),
           )
       }
```
RafalSumislawski, line 244: I can see some leftover uses of unmodeled `Header.Raw(ci"Access-Control-Allow-Origin" ...`. I think these should at least refer to `Header[`Access-Control-Allow-Origin`].name` if not just use the modeled version of the header you're introducing in this PR.

laws/src/main/scala/org/http4s/laws/discipline/arbitrary.scala
```diff
@@ -731,14 +731,17 @@ private[discipline] trait ArbitraryInstances { this: ArbitraryInstancesBinCompat
   implicit val http4sTestingCogenForUserInfo: Cogen[Uri.UserInfo] =
     Cogen.tuple2[String, Option[String]].contramap(u => (u.username, u.password))
 
+  def genMaybePort: Gen[Option[Int]] =
+    Gen.option(posNum[Int].suchThat(port => port >= 0 && port <= 65536))
```
RafalSumislawski, line 735: 0 and 65536 are not valid ports. 0 has a special meaning of "please choose a port for me". 65536 doesn't fit in two bytes.

There is more efficient way to generate a number in chosen range than via `suchThat`. In this case it could be
```
    Gen.option(chooseNum(1, 65535))
```

core/src/main/scala/org/http4s/headers/Origin.scala
```diff
@@ -50,31 +49,36 @@ object Origin {
...
+
+    val scheme = Uri.Parser.scheme
+
+    val host = Uri.Parser.host
+
     val port = char(':') *> digit.rep.string.map(_.toInt)
-    val nullHost = (string("null") *> `end`).as(Origin.Null)
 
-    val singleHost = ((scheme <* string("://")) ~ host ~ port.?).map { case ((sch, host), port) =>
+    ((scheme <* string("://")) ~ host ~ port.? ~ char('/').?).map { case (((sch, host), port), _) =>
```
RafalSumislawski, line 68: Is that `/` actually allowed by the specification? I had a look at https://www.rfc-editor.org/rfc/rfc6454#section-7 and I don't see it there.

RafalSumislawski (changes requested):

danicheg:
This is too stale

PR #5612, opened 2021-11-23
Proposal: Parse hosts with trailing slash in the `Origin` header

danicheg:
This PR is mainly a proposal. The current parser of the `Origin` header is corresponding to `RFC` well. But it fails for cases of hosts with trailing slash. Also, in 0.22.x and 0.23.x series that parser works well for hosts with a trailing slash.

### DIFF
--- a/core/shared/src/main/scala/org/http4s/headers/Origin.scala
+++ b/core/shared/src/main/scala/org/http4s/headers/Origin.scala
@@ -51,8 +51,12 @@ object Origin {
 
     val nullP = (string("null") *> `end`).as(Origin.`null`)
 
-    val hostP = ((scheme <* string("://")) ~ host ~ (char(':') *> port).?.map(_.flatten)).map {
-      case ((sch, host), port) => Origin.Host(sch, host, port)
+    val portP = (char(':') *> port).?.map(_.flatten)
+
+    val trailingSlashP = char('/').?
+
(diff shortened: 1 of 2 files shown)


bplommer:
What problem does this solve? Are there browsers that `Origin` headers with a trailing slash?

danicheg:
@bplommer I didn't research on browser's behavior (or even of every old version of browsers :)) What it's doing exactly is preserving semantic of parsing trailing slash.

bplommer:
I lean towards 👎 on this unless it solves a specific real-world problem. The Origin header is only generated by browsers, so I think we can assume it will be generated correctly. (Fwiw, I think a trailing slash on a host doesn't semantically make sense, since it's just that - a host, not uri with a path component.)

danicheg:
I ten to one agree, that browsers generate origins without trailing slash. But that header is a part of public API. And it can be used by end-users differently. So that's my point to preserve the previous behavior of parsing trailing slash. But I'm not so convinced about this one. Perhaps, I'm just overthinking :)

rossabaker:
I'm mildly :-1: on it.  Even though we used to accept it, we haven't seen anybody generating it in the wild.  It's good to be generous in what we parse, but if we start parsing hypothetical deviations from the spec instead of observed ones, where does it end?

PR #5613, opened 2021-11-25
Model If-Range header

mcarolan:
Based on #5010


core/src/main/scala/org/http4s/headers/If-Range.scala
```diff
@@ -0,0 +1,48 @@
...
+
+package org.http4s
+package headers
+
+import cats.parse.Parser
+import org.typelevel.ci._
+
+// https://datatracker.ietf.org/doc/html/rfc7233#section-3.2
+
+sealed abstract class `If-Range`
```
danicheg, line 25:
```suggestion
sealed abstract class `If-Range` extends Product with Serializable
```

core/src/main/scala/org/http4s/headers/If-Range.scala
```diff
@@ -0,0 +1,48 @@
...
+import cats.parse.Parser
+import org.typelevel.ci._
+
+// https://datatracker.ietf.org/doc/html/rfc7233#section-3.2
+
+sealed abstract class `If-Range` extends Product with Serializable
+
+object `If-Range` {
+  final case class LastModified(date: HttpDate) extends `If-Range`
+  final case class ETag(tag: EntityTag) extends `If-Range`
```
rossabaker, line 29: This is correct according to the grammar, but weak entity tags are forbidden here per the semantics.  We don't have a distinction between strong and weak at the type level.  I guess it's relatively harmless to allow them.

rossabaker (approved):

PR #5614, opened 2021-11-26
Add `Trailer` header

danicheg:
It's yet another header from the #5010 list.


rossabaker (commented):
`Trailer = 1#field-name`

Should this accept things that aren't field names?  I would say absolutely not if we had a field-name newtype.  As it stands, this is consistent with the same holes we already have.

rossabaker (approved):
I don't think it makes sense to make this stricter than header names in the same series.

PR #5618, opened 2021-11-29
(scalajs linking) StaticFile: make staticFileKey a lazy val

yurique:
This is a small tweak for fixing scalajs linking when using `StaticFile` (when using the `1.0.0` series).

Targeting this PR to series/0.23 so the changes can be synchronized with other branches more easily. 

### The problem: 
Because `staticFileKey` is not lazy, scalajs linker will fail to link a project if `StaticFiles` is referenced: 
```
[error] Referring to non-existent class java.io.File
[error]   called from constructor org.http4s.StaticFile$.<init>()void
```

### The solution:
Making it a `lazy val` fixes the linking :) (as long as `fromFile` / `fromResource` / `fromURL` are not called)

### DIFF
--- a/core/shared/src/main/scala/org/http4s/StaticFile.scala
+++ b/core/shared/src/main/scala/org/http4s/StaticFile.scala
@@ -347,6 +347,6 @@ object StaticFile {
   private[http4s] val staticPathKey = Key.newKey[SyncIO, Path].unsafeRunSync()
 
   @deprecated("Use staticPathKey", since = "0.23.5")
-  private[http4s] val staticFileKey: Key[File] =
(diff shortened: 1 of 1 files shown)


yurique:
@armanbilge: Thanks for steering me away from way more complicated ideas about how to address this :)

armanbilge (approved):
Actually this fix is applicable to SJS in 0.23 as well :)

FTR, our test suite missed this and is actually helpless to check for this kind of problem b/c of:
- https://github.com/scalameta/munit/issues/437

Unfortunately that requires a breaking change to fix so we won't see it until munit 1.0.

PR #5619, opened 2021-11-29
Update `cats-effect` version to 3.3.0

danicheg:
Steward is a bit late for now, so here is CE 3.3.0 (though it doesn't migrate changes of `TestContext` https://github.com/typelevel/fs2/runs/4344942116?check_suite_focus=true).


armanbilge (approved):
Who needs steward when we have our very own @danicheg :wink:

NIce cleanup!

danicheg:
Trying to do my best :)

armanbilge:
Btw, where is steward? It just opened https://github.com/http4s/http4s/pull/5621 but nothing for CE3.

PR #5622, opened 2021-11-29
Fix scaladoc references to io.chrisdavenport.vault

MasseGuillaume:


armanbilge (approved):
Thanks!

MasseGuillaume:
This type was actually helpful, I'm migrating http4s version, maybe it would be worth having the org change in the https://http4s.org/changelog/

rossabaker:
This should be true in 0.23 as well.  I'll cherry-pick it back.

PR #5633, opened 2021-11-30
Close some resources in tests

danicheg:
A minor thing, most for fixing warnings from linter in the $IDE. 
Introduced in this PR `AutoCloseableResource.resource` perhaps should be in the JVM module for the `0.23` series. I'll port it in case of merging this PR.

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -1,4 +1,4 @@
-version = 3.1.1
+version = 3.1.2
 
 style = default
 
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -1,4 +1,4 @@
-version = 3.1.2
+version = 3.2.0
 
 style = default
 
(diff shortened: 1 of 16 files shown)


armanbilge:
This looks very similar to `Resource.fromAutoCloseable`: https://typelevel.org/cats-effect/api/2.x/cats/effect/Resource$.html#fromAutoCloseable[F[_],A%3C:AutoCloseable](acquire:F[A])(implicitF:cats.effect.Sync[F]):cats.effect.Resource[F,A]

danicheg:
@armanbilge yep. By and large for cases not dealing with `F[_]` and CE `Resourse`. I've strived to don't refactor tests much, keeping their initial semantic.

armanbilge:
Thanks for explaining. IMHO I think it would be more idiomatic to make use of `Resource.fromAutoCloseable` and `ResourceFixture` from munit-cats-effect. But, whatever works :)

rossabaker (approved):
I'm not sure what the ResourceFixture looks like in these untracked effects.  Open to that as a subsequent refactor, but this looks good.

armanbilge:
I'm pretty sure `ResourceFixture` works the same for any kind of test, since it's wiring CE `Resource` with munit's existing fixture system (which doesn't know anything about CE).

danicheg:
@armanbilge using `ResourceFixture` or CE `Resource` is looking promising. Besides, it would mean solid refactoring of tests, which I've tried to get around. And I agree with Ross, that one may do this subsequently. I left a comment about it for hereafter investigators.

armanbilge (approved):

PR #5634, opened 2021-12-01
relax Client.translate bound on G[_] to MonadCancelThrow

bpholt:
MiMa flags this change, so I assume it needs to target 1.x?

### DIFF
--- a/client/shared/src/main/scala/org/http4s/client/Client.scala
+++ b/client/shared/src/main/scala/org/http4s/client/Client.scala
@@ -169,7 +169,7 @@ trait Client[F[_]] {
     */
   def get[A](s: String)(f: Response[F] => F[A]): F[A]
 
-  @deprecated("use public method with MonadCancelThrow instead", since = "2021-Dec")
+  @deprecated("use public method with MonadCancelThrow instead", since = "0.23.7")
   private[client] def translate[G[_]: Async](
       fk: F ~> G
   )(gK: G ~> F)(implicit F: MonadCancelThrow[F]): Client[G] = translateImpl(fk)(gK)
--- a/client/shared/src/main/scala/org/http4s/client/Client.scala
+++ b/client/shared/src/main/scala/org/http4s/client/Client.scala
@@ -170,18 +170,21 @@ trait Client[F[_]] {
   def get[A](s: String)(f: Response[F] => F[A]): F[A]
 
   @deprecated("use public method with MonadCancelThrow instead", since = "2021-Dec")
-  private[client] def translate[G[_]: Async](fk: F ~> G)(gK: G ~> F)(implicit
-      F: MonadCancelThrow[F]): Client[G] =
+  private[client] def translate[G[_]: Async](
+      fk: F ~> G
+  )(gK: G ~> F)(implicit F: MonadCancelThrow[F]): Client[G] =
     Client((req: Request[G]) =>
       run(
         req.mapK(gK)
       ).mapK(fk)
-        .map(_.mapK(fk)))
+        .map(_.mapK(fk))
+    )
 
   /** Translates the effect type of this client from F to G
     */
-  def translate[G[_]: MonadCancelThrow](fk: F ~> G)(gK: G ~> F)(implicit
-      F: MonadCancelThrow[F]): Client[G] =
+  def translate[G[_]: MonadCancelThrow](
+      fk: F ~> G
+  )(gK: G ~> F)(implicit F: MonadCancelThrow[F]): Client[G] =
     Client((req: Request[G]) =>
       run(
         req.mapK(gK)
--- a/client/shared/src/main/scala/org/http4s/client/Client.scala
+++ b/client/shared/src/main/scala/org/http4s/client/Client.scala
@@ -171,9 +171,8 @@ trait Client[F[_]] {
 
   /** Translates the effect type of this client from F to G
     */
-  def translate[G[_]: Async](
-      fk: F ~> G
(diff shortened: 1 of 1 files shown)


bplommer:
We've added mima exclusions for these in the past, since they only break bincompat if using interop from Java. I think this one can target 0.23.

armanbilge:
What's the MiMa error? I'm surprised for sure.

client/shared/src/main/scala/org/http4s/client/Client.scala
```diff
@@ -166,8 +166,8 @@ trait Client[F[_]] {
 
   /** Translates the effect type of this client from F to G
     */
-  def translate[G[_]: Async](fk: F ~> G)(gK: G ~> F)(implicit
-      F: MonadCancel[F, Throwable]): Client[G] =
+  def translate[G[_]: MonadCancelThrow](fk: F ~> G)(gK: G ~> F)(implicit
+      F: MonadCancelThrow[F]): Client[G] =
```
armanbilge, line 170: Oh, sorry, actually this change is not binary compatible. You cannot change `Async` to `MonadCancelThrow` binary compatibly. Unfortunately this is not a Java-only problem.

We may be able to work around it in 0.23 by having two versions of this method, but not sure. At the very least we can name the other one `mapK` or something ...

bplommer: > You cannot change Async to MonadCancelThrow binary compatibly. Unfortunately this is not a Java-only problem.

Wait, why is this a problem in Scala?

armanbilge: It's because binary compatibility doesn't care about the inheritance structure between `Async` and `MonadCancel`. They are different. This is a JVM level issue.

bplommer: Oh does that mean https://github.com/http4s/http4s/pull/5071 was bad and wrong too?

armanbilge: Yikes, I think it was. Two issues got conflated there. The static-forwarder problem is Java only, but any signature change applies to Scala and Java (and any JVM language).

armanbilge: In general, if MiMa complains, it is always right i.e. that change is not binary compatible. When we do exclusions we are effectively declaring that nobody on the outside is using that method (either b/c it's private/internal, or because it's only used by Java, and nobody uses http4s on Java, right?).

bplommer: > Yikes, I think it was. Two issues got conflated there. The static-forwarder problem is Java only, but any signature change applies to Scala and Java (and any JVM language).

Oops 😨 what's the fix, adding a package-private overload with the previous signature?

armanbilge: Yes, that should do it. Java is still out in the cold but I think that was the original intention :)

danicheg (commented):
That is good stuff, but I agree with @armanbilge that binary compatibility is breaking here. Probably we can have it in the next major (or epoch) release?

client/shared/src/main/scala/org/http4s/client/Client.scala
rossabaker: I think the package-private overload fixes this in classes, but not objects.  Objects still have that static forwarder problem, which we've waived a couple times for things that were causing clear and present problems, Java users be damned.

I'm +1 on this in 0.23 if it can be made MiMa clean, or if it can be reduced to a static forwarder and is causing real-world pain now.

bpholt: It looks like the package-private overload will work here. Do you want me to push it up on this branch, or should I push up a rebased version targeting series/0.23?

armanbilge: Let's try targeting 0.23 if you don't mind :)

client/shared/src/main/scala/org/http4s/client/Client.scala
```diff
@@ -169,11 +169,22 @@ trait Client[F[_]] {
     */
   def get[A](s: String)(f: Response[F] => F[A]): F[A]
 
+  @deprecated("use public method with MonadCancelThrow instead", since = "2021-Dec")
+  private[client] def translate[G[_]: Async](
+      fk: F ~> G
+  )(gK: G ~> F)(implicit F: MonadCancelThrow[F]): Client[G] =
+    Client((req: Request[G]) =>
```
armanbilge, line 176: Can this just delegate to the new method? I think you may have to pass the implicits explicitly and "cast" their types to `MonadCancelThrow` to resolve the ambiguity, but I think it should work.

bpholt: Updated

armanbilge (approved):

client/shared/src/main/scala/org/http4s/client/Client.scala
```diff
@@ -169,11 +169,20 @@ trait Client[F[_]] {
     */
   def get[A](s: String)(f: Response[F] => F[A]): F[A]
 
+  @deprecated("use public method with MonadCancelThrow instead", since = "2021-Dec")
```
rossabaker, line 172:
```suggestion
  @deprecated("use public method with MonadCancelThrow instead", since = "0.23.7")
```

rossabaker (approved):

PR #5635, opened 2021-12-01
Merge branch `series/0.22` into `series/0.23`

danicheg:
Previously, I was about to move the `AutoCloseableResource` to the JVM module. And accidentally found out that there is the `StaticFileSuite` in the shared module, which works well on JS with it. `ScalaJS` now can work with the `AutoCloseable`, or am I missing something? Anyways, tests passed locally. Let's see what'll happen on the CI.

**UPD**: that test suite contains `Platform.isJvm`, so that's why it works on JS (or accurately, not work on JS at all) 😄

### DIFF
--- a/tests/shared/src/test/scala/org/http4s/StaticFileSuite.scala
+++ b/tests/shared/src/test/scala/org/http4s/StaticFileSuite.scala
@@ -24,6 +24,7 @@ import fs2.io.file.Files
 import fs2.io.file.Path
 import org.http4s.Status._
 import org.http4s.headers._
+import org.http4s.testing.ClosableResource
 
 import java.net.URL
 import java.net.UnknownHostException
@@ -297,7 +298,8 @@ class StaticFileSuite extends Http4sSuite {
   if (Platform.isJvm)
     test("Read from a URL") {
       val url = getClass.getResource("/lorem-ipsum.txt")
-      val expected = scala.io.Source.fromURL(url, "utf-8").mkString
+      val expected =
+        ClosableResource.resource(scala.io.Source.fromURL(url, "utf-8"))(_.mkString)(_.close())
       val s = StaticFile
         .fromURL[IO](getClass.getResource("/lorem-ipsum.txt"))
         .value
--- a/tests/shared/src/test/scala/org/http4s/StaticFileSuite.scala
+++ b/tests/shared/src/test/scala/org/http4s/StaticFileSuite.scala
@@ -24,7 +24,7 @@ import fs2.io.file.Files
 import fs2.io.file.Path
 import org.http4s.Status._
 import org.http4s.headers._
-import org.http4s.testing.ClosableResource
+import org.http4s.testing.AutoCloseableResource
 
 import java.net.URL
 import java.net.UnknownHostException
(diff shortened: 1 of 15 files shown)


tests/shared/src/test/scala/org/http4s/StaticFileSuite.scala
```diff
@@ -297,7 +298,8 @@ class StaticFileSuite extends Http4sSuite {
   if (Platform.isJvm)
     test("Read from a URL") {
       val url = getClass.getResource("/lorem-ipsum.txt")
-      val expected = scala.io.Source.fromURL(url, "utf-8").mkString
+      val expected =
+        AutoCloseableResource.resource(scala.io.Source.fromURL(url, "utf-8"))(_.mkString)
```
armanbilge, line 302: This is the only `StaticFileSuite` test using `AutoCloseableResource`, and it is JVM-only :)

danicheg: Oh, shame on me. `Platform.isJvm` 😄 I didn't expect to see this. Maybe it's better to split it into two test suites?

armanbilge: If this file had been split, then this merge you just did would have been tripped up here!

armanbilge: Although I agree, when the days of merges are over we should indeed split files where appropriate :)

danicheg: What I actually don't get is why the `testingJS` module compiles. Or it'll fall in the runtime (if some code will work with `AutoCloseable`)?

armanbilge: Because `AutoCloseable` is supported on Scala.js :)
https://github.com/scala-js/scala-js/blob/main/javalanglib/src/main/scala/java/lang/AutoCloseable.scala

armanbilge: But in general, yes, Scala.js code can compile, but that doesn't mean it will link/run. For example `scala.io.Source.fromURL` is not supported on Scala.js, but because it's in the if-JVM statement, the Scala.js linker is able to remove that code before running it.

danicheg: Good to know, thanks. It seems we need to move the `AutoCloseableResource` in the JVM module as well as separate that JVM-specific test-suite (otherwise, it will not compile). What do you think?

armanbilge: I don't understand. The CI is green, and I don't see any reason it shouldn't be?

danicheg: Sorry, I'm confused a bit too. If we can leave it as is, I'll be pretty much glad :)

armanbilge: Everything is perfect :) sorry I couldn't explain it very well.

armanbilge (approved):

PR #5648, opened 2021-12-02
Publish http4s-scalatags for Scala 3

rossabaker:
This is new in 0.11.


rossabaker:
No, hardly anybody uses this module, hardly anybody uses Scala 3 yet, and a large percentage of those users are already on main.

PR #5653, opened 2021-12-03
Merge branch `series/0.22` into `series/0.23`

danicheg:
I hope I haven't lost anything.


.scala-steward.conf
```diff
@@ -8,5 +8,7 @@ updates.pin = [
   # simpleclient-0.12 breaks metric names
   { groupId = "io.prometheus", version = "0.11." },
   # okio-3 has (slight) binary incompatibilities
-  { groupId = "com.squareup.okio", artifactId = "okio", version = "2." }
+  { groupId = "com.squareup.okio", artifactId = "okio", version = "2." },
+  # scalatags-0.11 breaks bincompat
+  { groupId = "com.lihaoyi", artifactId = "scalatags", version = "0.10." },
```
rossabaker, line 13: We'll want to remember to remove this one on main.

rossabaker (approved):

.scala-steward.conf
danicheg: I've added #5654 to track this.

rossabaker:
Are you doing the main merge, or should I take care of it?

danicheg:
We didn't do that for a while. So now the `main` is behind from the `0.23` on 300+ commits. A bit scary, but I could give it a try.

rossabaker:
Those branches are much more similar than 0.22 -> 0.23, so I bet it's not as bad as the ones you've just done.

PR #5656, opened 2021-12-04
website: add favicon

### LINKED ISSUE #5451: Favicon
In #5313, @jenshalm  points out we don't have a favicon, and he volunteered to make us one.  If that offer is still good, I'd like to take him up on it.  If it's not, I'd like someone with more taste than me to do it. :smile:

jenshalm:
Uses SVG as the default icon, plus a PNG as fallback although none of
the more recent browsers require it anymore.

Closes #5451.


jenshalm:
I just noticed that this would need to be duplicated again for the `docs` project, which in turn would require the duplication of the image files (that's the downside of having the project split).

jenshalm:
Ok, it's now covering both projects, should be good to merge.

rossabaker (approved):
Nice simplification of the logo.  Looks great!

PR #5657, opened 2021-12-04
Ember H2 support

ChristopherDavenport:
### Http4s Core
- [x] Add Http Version to Request/Response

### Server
- [x]  ALPN H2
- [x] Http2 Prior Knowledge
- [x] H2c Upgrade
- [x] Push Promises

### Client
- [x] ALPN H2
- [x] Http2 Prior Knowledge
- [ ] h2c Upgrage - :stop_sign:  Wont happen in this PR - Needs a different client work-in method where we can negotiate and leak a connection from h1 into h2 
- [x] Push Promises

### ScalaJS
- [x] https://github.com/http4s/http4s/pull/5663
- [x] Hpack ScalaJS Shim
- [x] Platform Specific TLSParameters => TLSParameters for adding ALPN support
- [x] Ability to extract Application Protocol on JS (waiting on an fs2 release)
       https://github.com/typelevel/fs2/pull/2741


ChristopherDavenport:
I may need to hold off on h2c for client. Our current pooling mechanism is not compatible with the h2 connection model.

ChristopherDavenport:
Any ideas on why scaladoc would fail on just 2.12.5?

armanbilge:
Are these errors related to that?
```
[error] /home/runner/work/http4s/http4s/ember-core/shared/src/main/scala/org/http4s/ember/core/h2/H2Frame.scala:566:44: type mismatch;
[error]  found   : Int
[error]  required: Byte
[error]       val flag: Byte = if (settings.ack) 0 | (1 << 0) else 0
[error]                                            ^
[error] /home/runner/work/http4s/http4s/ember-core/shared/src/main/scala/org/http4s/ember/core/h2/H2Frame.scala:736:40: type mismatch;
[error]  found   : Int
[error]  required: Byte
[error]       val flag: Byte = if (ping.ack) 0 | (1 << 0) else 0
[error]                                        ^
[error] /home/runner/work/http4s/http4s/ember-core/shared/src/main/scala/org/http4s/ember/core/h2/H2Frame.scala:869:47: type mismatch;
[error]  found   : Int
[error]  required: Byte
[error]       val flag: Byte = if (cont.endHeaders) 0 | (1 << 2) else 0
[error]                                               ^
[error] three errors found
```
https://github.com/http4s/http4s/runs/4487053934?check_suite_focus=true#step:15:61

bplommer:
This is huge! Is there any way it can be sliced up for review?

armanbilge:
FTR there's about 20k of JavaScript lockfiles in here 😆 we can make them go away if this takes off https://github.com/davenverse/hpack.

Still big though.

ChristopherDavenport:
I'm considering how this could be managed. I don't really think it works sliced up. It really comes together as a whole. 

The way I build it so the way I  recommend slicing it is for review is.

H2Frame: This is the base level binary protocol, all the actual communication happens via this
H2Connection: How do we interpret the Receive Frames, vs how do we send frames.
H2Stream: This is the independent state machines for each of the request/response flows. The state machine supports both client and server versions because thats how the protocol is designed.

H2Server: How do we actually receive requests and turn them into responses. The impl is left commented as thats the entire component prior to its integration into the EmberServer. This works exactly how we want the server to work. So is a better place to start getting into the api level
H2Client: This is where the hacks start layering up. The pooling mechanism we are using doesn't allow connections to leak into a different infinite process. So instead we use a layered approach that uses alpn to negotiate, and if negotiates to http1, gives up and sets it to not try that trick again and delegates to a fallback http/1.1 client. This can and should be much more elegant like the server implementation but can't be due to the shortcomings of the KeyPool mechanism.(More safety actually bit us here)

Then we have the integration components into EmberServer and Ember-Client, the Server is thoroughly integrated and allows full upgrade. The client is mostly just a wrapper of the current implementation operating as the fallback to the H2Client mechanisms operation if it can work. 

If folks think it would be better I could maybe make a video or something walking through how it works, or do some live sessions.

ChristopherDavenport:
Found out the hard way that although they are case-insensittve. H2 wants them sent in lower-case. So if we switched over ours to actually be lower-case this will reduce wasted string-rewriting. I found this on the trailer header.

ChristopherDavenport:
```
 ==> X org.http4s.StatusSpec.rendering sanitizes statuses  0.041s munit.FailException: /home/runner/work/http4s/http4s/tests/shared/src/test/scala/org/http4s/StatusSpec.scala:124
123:
124:  test("rendering sanitizes statuses") {
125:    forAll { (s: Status) =>

Failing seed: FURqTqx4pvXTHRz9u5OKq81dUdrG9WO0g74vMIOVlAK=

  override val scalaCheckInitialSeed = "FURqTqx4pvXTHRz9u5OKq81dUdrG9WO0g74vMIOVlAK="

Falsified after 68 passed tests.
> ARG_0: 188 ⽣劲ᦢ蘋嚁馺硊헳Ű趭廗♒娙﻾듁⚬烁楸䠳꥞ㇿ㎦灘ា㍰䟍뗁⎴ᔷ槼쑲庇憨묒ٱ搾適鬤鶆ᡂ罊̍迺侞圥
    at munit.Assertions.fail$(Assertions.scala:277)
    at munit.FunSuite.fail(FunSuite.scala:11)
    at munit.ScalaCheckSuite.$anonfun$propToTry$4(ScalaCheckSuite.scala:104)
    at scala.util.Try$.apply(Try.scala:210)
    at munit.ScalaCheckSuite.propToTry(ScalaCheckSuite.scala:104)
    at munit.ScalaCheckSuite.$anonfun$scalaCheckPropTransform$3(ScalaCheckSuite.scala:46)
    at scala.concurrent.impl.Promise$Transformation.run(Promise.scala:475)
    at cats.effect.unsafe.WorkStealingThreadPool.$anonfun$execute$1(WorkStealingThreadPool.scala:506)
    at cats.effect.IOFiber.runLoop(IOFiber.scala:267)
    at cats.effect.IOFiber.execR(IOFiber.scala:1294)
    at cats.effect.IOFiber.run(IOFiber.scala:124)
    at cats.effect.unsafe.WorkerThread.run(WorkerThread.scala:505)
    ```

**Review by armanbilge — APPROVED** [review]
I'm approving this. Personally I don't have the time nor expertise for an in-depth review (I'm sorry) but @ChristopherDavenport does great work and in any case I'm sure he will continue iterating on this after it merges.

I think it would be valuable to get this into the next release train. As @ChristopherDavenport's grpc experiment demonstrated, actually putting it to use is a good strategy for collecting feedback.

I did pay attention to the new APIs in the ember server/client builders. These seem straightforward enough and we can designate them as experimental and promise no bincompat for them.

Finally, if the JS lockfiles are a problem, take it up with me 😉 we have a long-term strategy to make them go away, but in the meantime we could yolo without them and probably be fine.

**Comment by rossabaker** [comment]
Chris is trusted and Arman looked, so I'm okay with merging.  The downside is that binary compatibility gets locked for 0.23 once we do.  If you're confident in the public API, :+1:.  If we're not, we could at least merge it to main and then backport it later.  It's also significant enough to merit its own train, if you want to be cautious.  I'll leave it to @ChristopherDavenport's discretion when to accept the binary constraints.

I'm still digging out of the holiday rubble, but I'll focus first on the public API.

**Comment by ChristopherDavenport** [comment]
Feeling as though this is in a pretty good position. I have reviewed exposed components and almost everything is package private. What isn't I don't really expect a need to change. 

I'm going to leave this another week for folks to review before we say this is good enough for a first shot. Something is more than nothing.

File: build.sbt [hunk 1]
```diff
@@ -406,14 +406,19 @@ lazy val dropwizardMetrics = libraryProject("dropwizard-metrics")
...
     startYear := Some(2019),
     unusedCompileDependenciesFilter -= moduleFilter("io.chrisdavenport", "log4cats-core"),
     libraryDependencies ++= Seq(
       log4catsTesting.value % Test
     ),
+    scalacOptions --= Seq(
+      "-Ywarn-numeric-widen",
+      "-Werror",
+      "-Xfatal-warnings",
+    ),
```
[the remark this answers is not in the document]
rossabaker: Yeah, this is risky.  But a simple `@nowarn` might not be so simple if it spans categories.  I'd like to study what warnings are left.

ember-client/shared/src/main/scala/org/http4s/ember/client/EmberClientBuilder.scala
```diff
@@ -128,6 +141,16 @@ final class EmberClientBuilder[F[_]: Async] private (
   def withUnixSockets(unixSockets: UnixSockets[F]) =
     copy(unixSockets = Some(unixSockets))
 
+  def withHttp2 = copy(enableHttp2 = true)
+  def withoutHttp2 = copy(enableHttp2 = false)
+
+  def withPushPromiseSupport(
```
rossabaker, line 147: A scaladoc on this signature would go a long way.  I'm not sure what I'd call this with.

ember-core/js/src/main/scala/org/http4s/ember/core/h2/H2TLSPlatform.scala
```diff
@@ -0,0 +1,38 @@
...
+import cats.syntax.all._
+import fs2.io.net.tls.TLSParameters
+
+private[h2] abstract class H2TLSPlatform {
+
+  def transform(params: TLSParameters): TLSParameters =
+    TLSParameters(
+      requestCert = params.requestCert,
+      rejectUnauthorized = params.rejectUnauthorized,
+      alpnProtocols = List("http/1.1").some, // "h2", first needs support for application extraction
```
rossabaker, line 28: Should this be additive to what was in `params`, or does it need to be just this singleton?  I don't know the protocol very well.

ember-core/jvm/src/main/scala/org/http4s/ember/core/h2/HpackPlatform.scala
```diff
@@ -0,0 +1,90 @@
...
+            StandardCharsets.ISO_8859_1,
+          )
+        )
+    }
+
+    tDecoder.decode(is, listener)
+    tDecoder.endHeaderBlock()
+
+    val decoded = buffer.toList
+    NonEmptyList.fromList(decoded).toRight(new Throwable("Header List Was Empty"))
```
rossabaker, line 72: `NoSuchElementException`?

ember-core/shared/src/main/scala/org/http4s/ember/core/h2/H2Error.scala
```diff
@@ -0,0 +1,132 @@
...
+private[h2] sealed abstract class H2Error(val value: Int) {
+  def toGoAway(highest: Int): H2Frame.GoAway =
+    H2Frame.GoAway(0, highest, value, None)
+  def toRst(stream: Int): H2Frame.RstStream =
+    H2Frame.RstStream(stream, value)
+
+}
+private[h2] object H2Error {
+
+  def fromInt(int: Int): Option[H2Error] = int match {
```
rossabaker, line 29: Does this happen enough to merit caching the somes?

ember-core/shared/src/main/scala/org/http4s/ember/core/h2/H2Frame.scala
```diff
@@ -0,0 +1,885 @@
...
+      case Settings.`type` => Settings.fromRaw(rawFrame)
+      case PushPromise.`type` => PushPromise.fromRaw(rawFrame)
+      case Ping.`type` => Ping.fromRaw(rawFrame)
+      case GoAway.`type` => GoAway.fromRaw(rawFrame)
+      case WindowUpdate.`type` => WindowUpdate.fromRaw(rawFrame)
+      case Continuation.`type` => Continuation.fromRaw(rawFrame)
+      case _ => Unknown(rawFrame).asRight
+    }
+
+  def toRaw(frame: H2Frame): RawFrame = frame match {
```
rossabaker, line 131: Any reason `toRaw` isn't just a method on `H2Frame`?

ember-core/shared/src/main/scala/org/http4s/ember/core/h2/H2Frame.scala
```diff
@@ -0,0 +1,885 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.http4s.ember.core.h2
+
+import cats.syntax.all._
+import scodec.bits._
+
+private[ember] sealed trait H2Frame
```
rossabaker, line 22: Yeah, my eyes glazed over on this file.  I'm trusting there are tests to give confidence there aren't errors in all these constants.

ember-core/shared/src/main/scala/org/http4s/ember/core/h2/H2Server.scala
```diff
@@ -0,0 +1,395 @@
...
+    status = Status.SwitchingProtocols,
+    httpVersion = HttpVersion.`HTTP/1.1`,
+    headers = Headers(
+      "connection" -> "Upgrade",
+      "upgrade" -> "h2c",
+    ),
+  )
+  // Apply this, if it ever becomes Some, then rather than the next request, become an h2 connection
+  def h2cUpgradeHttpRoute[F[_]: Concurrent]: HttpRoutes[F] =
+    cats.data.Kleisli[({ type L[A] = cats.data.OptionT[F, A] })#L, Request[F], Response[F]] {
```
rossabaker, line 81: Any reason this one isn't kind projected?

ember-core/shared/src/main/scala/org/http4s/ember/core/h2/H2Server.scala
```diff
@@ -0,0 +1,395 @@
...
+  // Call on a new connection for http2-prior-knowledge
+  // If left 1.1 if right 2
+  def checkConnectionPreface[F[_]: MonadThrow](socket: Socket[F]): F[Either[ByteVector, Unit]] =
+    socket.read(Preface.clientBV.size.toInt).flatMap {
+      case Some(s) =>
+        val received = s.toByteVector
+        if (received == Preface.clientBV) Applicative[F].pure(Either.right(()))
+        else Applicative[F].pure(Either.left(received))
+      case None =>
+        new Throwable("Input Closed Before Receiving Data").raiseError
```
rossabaker, line 151: `IOException`?

ember-core/shared/src/main/scala/org/http4s/ember/core/h2/H2Server.scala
```diff
@@ -0,0 +1,395 @@
...
+        .getOrElseF(app.run(req))
+    }
+
+  // Call on a new connection for http2-prior-knowledge
+  // If left 1.1 if right 2
+  def checkConnectionPreface[F[_]: MonadThrow](socket: Socket[F]): F[Either[ByteVector, Unit]] =
+    socket.read(Preface.clientBV.size.toInt).flatMap {
+      case Some(s) =>
+        val received = s.toByteVector
+        if (received == Preface.clientBV) Applicative[F].pure(Either.right(()))
```
rossabaker, line 148: `Either.unit`?  (I think that exists... we used to have our own.)

ember-core/shared/src/main/scala/org/http4s/ember/core/h2/H2Server.scala
```diff
@@ -0,0 +1,395 @@
...
+      case None =>
+        new Throwable("Input Closed Before Receiving Data").raiseError
+    }
+
+  // For Anything that is guaranteed to only be h2 this method will fail
+  // unless the connection preface is there. For example after ALPN negotiation
+  // on an SSL connection.
+  def requireConnectionPreface[F[_]: MonadThrow](socket: Socket[F]): F[Unit] =
+    checkConnectionPreface(socket).flatMap {
+      case Left(e) => new Throwable("Invalid Connection Preface").raiseError
```
rossabaker, line 159: IllegalArgumentException?

ember-core/shared/src/main/scala/org/http4s/ember/core/h2/H2Server.scala
```diff
@@ -0,0 +1,395 @@
...
+        )
+      )
+      _ <-
+        Stream
+          .fromQueueUnterminated(closed)
+          .map(i =>
+            Stream.eval(
+              // Max Time After Close We Will Still Accept Messages
+              (Temporal[F].sleep(1.seconds) >>
+                ref.update(m => m - i)).timeout(15.seconds).attempt.start
```
rossabaker, line 258: Are these magic numbers holy or should they be parameterized/

ember-core/shared/src/main/scala/org/http4s/ember/core/h2/H2Server.scala
```diff
@@ -0,0 +1,395 @@
...
+            Pull
+              .eval(stateRef.get.map(_.closed))
+              .ifM(Pull.done, go)
+        go.stream
+      }
+      _ <- s.compile.resource.drain
+    } yield ()
+  }
+
+  // def impl[F[_]: Async: Parallel](
```
rossabaker, line 354: Do you need all this still?

ember-core/shared/src/main/scala/org/http4s/ember/core/h2/H2Stream.scala
```diff
@@ -0,0 +1,507 @@
...
+        state.get.flatMap { s =>
+          s.state match {
+            case StreamState.Idle =>
+              for {
+                h <- hpack.encodeHeaders(headers)
+                frame = H2Frame.PushPromise(originating, true, id, h, None)
+                _ <- state.update(s => s.copy(state = StreamState.ReservedLocal))
+                _ <- enqueue.offer(Chunk.singleton(frame))
+              } yield ()
+            case _ => new Throwable("Push Promises are only allowed on an idle Stream").raiseError
```
rossabaker, line 55: IllegalStateException

ember-core/shared/src/main/scala/org/http4s/ember/core/h2/H2Stream.scala
```diff
@@ -0,0 +1,507 @@
...
+          }
+        }
+      case H2Connection.ConnectionType.Client =>
+        new Throwable("Clients Are Not Allowed To Send PushPromises").raiseError
```
rossabaker, line 59: IllegalStateException

ember-core/shared/src/main/scala/org/http4s/ember/core/h2/H2Stream.scala
```diff
@@ -0,0 +1,507 @@
...
+                  case (StreamState.ReservedLocal, false) => StreamState.HalfClosedRemote
+                  case (s, _) => s // Hopefully Impossible
+                }
+                (b.copy(state = newState), newState)
+              }
+              .flatMap { state =>
+                if (state == StreamState.Closed) onClosed else Applicative[F].unit
+              }
+              .void
+        case _ => new Throwable("Stream Was Closed").raiseError
```
rossabaker, line 91: I'm out of gas and out of basketball to watch, but you see the pattern by now.

ember-client/shared/src/main/scala/org/http4s/ember/client/EmberClientBuilder.scala
```diff
@@ -54,6 +56,11 @@ final class EmberClientBuilder[F[_]: Async] private (
     val checkEndpointIdentification: Boolean,
     val retryPolicy: RetryPolicy[F],
     private val unixSockets: Option[UnixSockets[F]],
+    private val enableHttp2: Boolean,
+    private val pushPromiseSupport: Option[
+      (Request[fs2.Pure], F[Response[F]]) => F[Outcome[F, Throwable, Unit]]
+    ],
+    // private[this] val logger: Logger[F],
```
danicheg, line 63:
```suggestion
```

ember-core/shared/src/main/scala/org/http4s/ember/core/h2/Preface.scala
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+package org.http4s.ember.core.h2
+
+import scodec.bits._
+// Preface must always be sent an happens before anything else.
+private[h2] object Preface {
+  val client = "PRI * HTTP/2.0\r\n\r\nSM\r\n\r\n"
+  // This sequence MUST be followed by a SETTINGS frame
+  val clientBV = ByteVector(client.getBytes)
```
rossabaker, line 24: Needs an explicit codec.  US-ASCII or ISO-8859-1 are fine.

ember-server/shared/src/main/scala/org/http4s/ember/server/EmberServerBuilder.scala
```diff
@@ -56,6 +56,7 @@ final class EmberServerBuilder[F[_]: Async] private (
     val additionalSocketOptions: List[SocketOption],
     private val logger: Logger[F],
     private val unixSocketConfig: Option[(UnixSockets[F], UnixSocketAddress, Boolean, Boolean)],
+    private val enableHttp2: Boolean,
```
rossabaker, line 59: Is this always orthogonal to other settings?  If we opt into this, we speak 1 and 2?  Would there ever be a desire to speak 2 only?

examples/ember-server-h2/src/main/scala/com/example/http4s/ember/EmberServerH2Example.scala
```diff
@@ -0,0 +1,91 @@
...
+    //   tlsContext = Network[F].tlsContext.fromSSLContext(sslContext)
+    //   _ <- EmberServerBuilder
+    //     .default[F]
+    //     .withTLS(tlsContext, TLSParameters.Default)
+    //     .withHttp2
+    //     .withHost(ipv4"0.0.0.0")
+    //     .withPort(port"8080")
+    //     .withHttpApp(simpleApp)
+    //     .build
+    // } yield ()
```
rossabaker, line 71: Is this an illustrative example?  I would remove it, or comment on why it's commented.

project/plugins.sbt
```diff
@@ -16,6 +16,7 @@ addSbtPlugin("io.spray" % "sbt-revolver" % "0.9.1")
 addSbtPlugin("org.scalameta" % "sbt-mdoc" % "2.2.24")
 addSbtPlugin("pl.project13.scala" % "sbt-jmh" % "0.4.3")
 addSbtPlugin("org.portable-scala" % "sbt-scalajs-crossproject" % "1.1.0")
+addSbtPlugin("ch.epfl.scala" % "sbt-scalajs-bundler" % "0.20.0")
```
rossabaker, line 19: I read this readme and still don't understand it.  Do we need npm now?

rossabaker (commented):
I need a deeper understanding of the protocol before I can give a deeper review, but the two big ones are:

1. I don't like turning off warnings, but we can leave an issue to find a way to rein that back in.  I'm less averse to turning off numeric widening, particularly because I didn't notice any of the lossy bad ones.

2. I would fix all the Throwables before those end up in people's error channels.  That's a semantic change when we clean it up later.

project/plugins.sbt
armanbilge: scalajs bundler works with either npm and yarn, and we are using it with yarn. Unfortunately we do need it for Scala.js support, for [hpack.js](https://www.npmjs.com/package/hpack.js). Hopefully at some point we can replace this with a pure Scala HPACK implementation e.g. https://github.com/davenverse/hpack.

ember-core/shared/src/main/scala/org/http4s/ember/core/h2/H2Server.scala
ChristopherDavenport: You can't kind project this for scala 3. It sucks.

ember-core/shared/src/main/scala/org/http4s/ember/core/h2/H2Frame.scala
ChristopherDavenport: None really, I write things as functions initially generally.

ember-core/shared/src/main/scala/org/http4s/ember/core/h2/H2Server.scala
ChristopherDavenport: They could be parameterized. This is mostly a background system designed to eventually remove dead streams. It could be parameterized if we really wanted to but it would extend our configuration space considerably.

ember-server/shared/src/main/scala/org/http4s/ember/server/EmberServerBuilder.scala
ChristopherDavenport: We could try to introduce http2 only support. I hadn't built that presently.

ember-core/shared/src/main/scala/org/http4s/ember/core/h2/H2Error.scala
ChristopherDavenport: Not generally I don't think.

ember-core/js/src/main/scala/org/http4s/ember/core/h2/H2TLSPlatform.scala
ChristopherDavenport: This is a singeton if we want to support both protocols for h2.

examples/ember-server-h2/src/main/scala/com/example/http4s/ember/EmberServerH2Example.scala
ChristopherDavenport: Comment added.

build.sbt
```diff
@@ -559,6 +575,15 @@ lazy val emberServer = libraryCrossProject("ember-server")
...
       ),
+      ProblemFilters.exclude[DirectMissingMethodProblem](
+        "org.http4s.ember.server.internal.ServerHelpers.upgradeSocket"
+      ),
+      ProblemFilters.exclude[DirectMissingMethodProblem](
+        "org.http4s.ember.server.internal.ServerHelpers.serverInternal"
+      ),
+      ProblemFilters.exclude[DirectMissingMethodProblem](
+        "org.http4s.ember.server.internal.ServerHelpers.unixSocketServer"
+      ),
```
armanbilge, line 586: Are these MiMa issues only on Scala 3? If so they should go in the Scala 3 MiMa section below, which we'll remove in the future once MiMa is able to automatically recognize package privates in Scala 3 like it does for Scala 2.

ChristopherDavenport: Looks like yes.

ember-server/shared/src/main/scala/org/http4s/ember/server/EmberServerBuilder.scala
rossabaker: I'm mostly concerned about the future state space there.  Now there are two states, so all is well:

* HTTP/1: `enableHttp2 = false`
* HTTP/1 and HTTP/2: `enableHttp2 = true`

If there's HTTP/2 only, now what?  You could introduce an `enableHttp1` flag, but what happens when they're both false?  Or you could preemptively make this a sum type, but you can't add to that without breaking bincompat.  Or you could make all three states now, but don't support the third state yet.  I don't think there's going to be a good answer here, but worst case, the server can fail to start on senseless arguments.

rossabaker:
I have nearly fixed all the warnings. A `@nowarn` takes care of the nuisance Int => Long, which I wish never warned.  (It allows evil floating point conversions, but I don't see any.)  The rest are mostly unused names, but it caught some vars and what I believe is a bug.

It killed my battery before I was done. We can merge this and then fix the warnings, or I can PR to this tomorrow. Don't care which.

rossabaker (approved):
Let's not belabor this PR with the warnings, but let's get those restored before release.  I'll open a new PR.

PR #5663, opened 2021-12-05
Cross `Server` and ember server to JS in 0.23

armanbilge:
Here we are again :)

Merging this will fully synchronize 0.23 and main regarding JS crosses. This is valuable e.g. for https://github.com/http4s/http4s/pull/5657 since it's very difficult to anticipate a crossed-main from an uncrossed 0.23.

It's not elegant, but I finally figured out how to work around the `InetSocketAddress` issue: `Server` is (unfortunately) split/duplicated across JVM and JS, and a http4s-private `Ip4sServer` provides a cross-platform way to implement `Server` without touching `InetSocketAddress` and using only ip4s stuff. This allows us to cross ember-server.


server/shared/src/main/scala/org/http4s/server/package.scala
```diff
@@ -43,31 +44,41 @@ package object server {
...
-    val Host = InetAddress.getLoopbackAddress.getHostAddress
+    def Host = InetAddress.getLoopbackAddress.getHostAddress
     val HttpPort = 8080
 
-    val IPv4SocketAddress: InetSocketAddress =
+    def IPv4SocketAddress: InetSocketAddress =
       InetSocketAddress.createUnresolved(IPv4Host, HttpPort)
-    val IPv6SocketAddress: InetSocketAddress =
+    def IPv6SocketAddress: InetSocketAddress =
       InetSocketAddress.createUnresolved(IPv6Host, HttpPort)
```
armanbilge, line 69: I had to make these `def`s so that the Scala.js linker can elide them. It's not ideal of course, but the alternative is splitting files. Reasonable compromise?

ChristopherDavenport (approved):

rossabaker (approved):

PR #5664, opened 2021-12-05
Merge branch `series/0.23` into `main`

danicheg:
Unfortunately, that wasn't a cakewalk. I've strived to be accurate but feeling some doubts. So, dear folks, your reviews (and fixes on this branch) would be so much opportunely :)


armanbilge:
Thanks for this effort. This is a massive diff 😅 is there specific things you were not sure about?

danicheg:
@armanbilge Some files were moved in the `series/0.23` and changed in the `main` (and/or vice versa). That was difficult at most and could be error-prone. Unfortunately, I didn't take notes for all that cases, but I think introducing JS functionality was one of that scenarios.

armanbilge (approved):
I also tried this (just as large diff) to see how 0.23 will compare with main after this merges:
https://github.com/http4s/http4s/compare/series/0.23...danicheg:merge-series/0.23-into-main

CI is green, don't see anything obviously wrong, I say yolo ✅

armanbilge:
Btw, in future when doing merging or other changes and you have questions/issues about JS-related stuff please don't hesitate to ping me :)

PR #5679, opened 2021-12-07
Release v0.22.8

rossabaker:


rossabaker:
Trying out a different release notes format.  I started it from GitHub's automation, and then categorized it, because most of them are meaningless to users.

I think it's going to repeat all of these on series/0.23, which is going to suck.  It's also not going to find backports that were cherry-picked.  I might end up writing something better, but we had a user ask for releases, and it's been two months, and we can always redo them later.

website/src/hugo/content/changelog.md
```diff
@@ -8,6 +8,119 @@
...
 ordered chronologically, so each release contains all changes described below it.
 
+# v0.22.8 (2021-12-07)
+
+This is a maintenance release, binary compatible with the 0.22.x series.
+
+## http4s-core
+
+* Update case-insensitive, ... to 1.2.0 in series/0.22 by @scala-steward in https://github.com/http4s/http4s/pull/5434
+* Sec-Fetch-Dest header by @m3is0 in https://github.com/http4s/http4s/pull/5357
```
armanbilge, line 18: Were there concerns about this PR, before it get branded in the binary record?
https://github.com/http4s/http4s/pull/5357#issuecomment-951899791

rossabaker: Yeah, and it got reverted, and GitHub's automation didn't recognize that.  It also misses the cherry picks, which I guess we should turn into PRs if we're going to start doing things with this tool.

armanbilge (approved):
I think frequent releases are nice. I've seen some projects manage release notes by adding the changelog entry as part of the PR itself. Then when release day rolls around it's all there. Could that strategy be helpful here?

rossabaker:
It could be very helpful, but becomes a point of contention for merge conflicts, and most drive-by contributors aren't going to be interested in another hurdle.

More frequent releases would also help, but I've only rarely ever gotten help cutting any.  The automation on cutting them is great.  The automation on documenting them is terrible, so nobody else wants to do it.

rossabaker:
This is probably a better discussion to continue on an issue that isn't closed.

Issue #5683, opened 2021-12-08
Improve release notes process


armanbilge:
Continuing from https://github.com/http4s/http4s/pull/5679#issuecomment-988381692.

rossabaker:
Brain dump

## Naive GitHub approach

Run the GitHub automation and be done with it.

### Good

* Easy
* Credits authors.  Sort of.

### Bad

* Does not categorize meaningful PRs
* Misses cherry picks and reverts that aren't PRs.  (Mitigation: don't do that anymore.)
* Isn't branch aware.  (Mitigation: maybe we want to repeat forward merges?)
* Miscredits authors of multi-author PRs.

## Curated GitHub approach

What we did for 0.22.8 and 0.23.7.

### Good

* Credits authors

### Bad

* Labor intensive to categorize
* Manual search of cherry picks and reverts that aren't PRs.  (Mitigation: don't do that anymore.)
* Manual filtering of merges forward.  (Mitigation: maybe we want to repeat forward merges?)
* Miscredits authors of multi-author PRs.

## Ross Magit approach

What Ross has been doing for a long time

### Good

* Finds everything automatically
* Automatically filters merges forward
* Makes us think about what's noteworthy on each PR

### Bad

* Labor intensive replacing "Merge ..." with descriptions
* Labor intensive to categorize
* Nobody else knows how to do it (but the git commands aren't hard!)
* Manual credit to authors.  (I haven't bothered.)

## Add to changelog in each PR

### Good

* Finds everything if we're diligent
* Automatically filters merges forward
* Makes us think about what's noteworthy on each PR

### Bad

* Oh, $DEITY, the merge conflicts!!!
* Adds to burden of each merge
* Only works if we're diligent
* Manual credit to authors

## Release Drafter

Label each PR.  Let Release Drafter sort it out

### Good

* Automated release note drafts
* Can see at a glance if it may be worth releasing!
* Free categorization if the metadata is good

### Bad

* Requires a Principal Staff YAML Engineer to configure it
* Every PR must be labeled. (Mitigation: more automation to label Scala Steward?)

### Unknown

* Does it credit all the authors?
* Does it work across branches?

## Release more often

Works with any of the above!

### Good

* Makes them all suck less
* Features in user's hands quicker

### Bad

* Makes it suck more often
* More time spent overall doing it
* User release fatigue

rossabaker:
On the main branch, GitHub autogenerate goes all the way back to 0.23.5. 👎

[...]

PR #5690, opened 2021-12-08
Configure mergify & release notes

armanbilge:
Trying out some of the ideas in https://github.com/http4s/http4s/issues/5683.

1. A mergify config, to label PRs based on their target branch and also steward PRs. Seems like this is always read from the default branch, but probably won't hurt to repeat it.
2. A GH release config, which will have to be customized for every branch. I hope GH respects that.

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -1,18 +1,86 @@
 changelog:
-  exclude:
-    labels:
-      - series/0.22
-      - series/0.23
   categories:
+
     - title: http4s-core
       labels:
-        - module:http4s-core
+        - module:core
+    - title: http4s-dsl
+      labels:
+        - module:dsl
+    - title: http4s-laws
+      labels:
+        - module:laws
+
     - title: http4s-server
       labels:
-        - module:http4s-server
+        - module:server
     - title: http4s-client
       labels:
-        - module:http4s-client
+        - module:client
+
+    - title: http4s-ember-core
+      labels:
+        - module:ember-core
+    - title: http4s-ember-server
+      labels:
+        - module:ember-server
+    - title: http4s-ember-client
+      labels:
+        - module:ember-client
+
+    - title: http4s-blaze-core
+      labels:
+        - module:blaze-core
+    - title: http4s-blaze-server
+      labels:
+        - module:blaze-server
+    - title: http4s-blaze-client
+      labels:
+        - module:blaze-client
+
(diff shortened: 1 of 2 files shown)


.github/release.yml
```diff
@@ -0,0 +1,18 @@
...
+        - module:http4s-core
+    - title: http4s-server
+      labels:
+        - module:http4s-server
+    - title: http4s-client
+      labels:
+        - module:http4s-client
+    - title: Behind the scenes
+      labels:
+        - "*"
```
armanbilge, line 18: So apparently `*` means catch-all for anything that wasn't categorized above. Which suggests to me, that e.g. if a PR is labeled for both core and server, it'll get listed under core since that comes first. So, not ideal.

rossabaker: I think if it's core and server, labeling it core is fine.  And the modules roughly form a tree, so we can sort them.  The exception would be a few backends that descend from backend-core and (server | client).

A bigger irritant would be anything tagged with siblings: a fix on blaze-server and ember-server.  The list will always require some curation.  We're looking for an 80% solution here.

armanbilge: Fair. Anyway we can fiddle with this config at any point. The mergify one is more important to start getting the labels automatically.

.github/release.yml
```diff
@@ -1,18 +1,86 @@
...
+      labels:
+        - module:blaze-core
+    - title: http4s-blaze-server
+      labels:
+        - module:blaze-server
+    - title: http4s-blaze-client
+      labels:
+        - module:blaze-client
+
+    - title: http4s-jetty-server
```
rossabaker, line 45: I'd sort servlet above this one.

.github/release.yml
```diff
@@ -1,18 +1,86 @@
...
+      labels:
+        - module:twirl
+
+    - title: http4s-dropwizard-metrics
+      labels:
+        - module:dropwizard-metrics
+    - title: http4s-prometheus-metrics
+      labels:
+        - module:prometheus-metrics
+
```
rossabaker, line 87: There could be one for scalafix.

rossabaker (approved):

armanbilge:
So before merging PR, the new process should be:
- add a category label, if relevant
- add a series/0.22 or series/0.23 label, if relevant
- edit the PR title to make it a suitable changelog entry, if necessary

PR #5712, opened 2021-12-10
Covariant `Message`s: minimum viable PR

bplommer:
Inspired by discussion #5709, this is a minimum viable PR to make progress towards having `Request` and `Response` be covariant per #5251.

Changes to `IdleTimeoutStage.scala` and `build.sbt` are unrelated issues caught by quicklint.

UPDATE: I've also made the minimum changes to make the `Message` types covariant.


core/shared/src/main/scala/org/http4s/internal/Logger.scala
```diff
@@ -71,8 +71,8 @@ object Logger {
   )(log: String => F[Unit])(implicit F: Monad[F]): F[Unit] = {
     def prelude =
       message match {
-        case req: Request[_] => s"${req.httpVersion} ${req.method} ${req.uri}"
-        case resp: Response[_] => s"${resp.httpVersion} ${resp.status}"
+        case req: Request[F @unchecked] => s"${req.httpVersion} ${req.method} ${req.uri}"
+        case resp: Response[F @unchecked] => s"${resp.httpVersion} ${resp.status}"
```
bplommer, line 75: I'm not sure why making `Media` covariant caused scalac to start complaining about the `_`.

ChristopherDavenport (commented):
I really don't enjoy variance, and don't understand the appeal. This seems like it will just make writing stuff around these models more difficult, for a slightly nicer experience translating from Pure to F in practice which is fairly rare.

That being said there is continuing interested in making it covariant so I don't really know what to say. It works, I just find programming with variance central to the model to be unpleasant.

bplommer:
> I really don't enjoy variance, and don't understand the appeal. 

The appeal for me is that in the great majority of use cases it removes the need to think about effects at all when instantiating requests and responses. 

> for a slightly nicer experience translating from Pure to F in practice which is fairly rare.

That's just the thing though - the usual idiom is to instantiate messages in `F` even when, as in the great majority of cases, the entity is being provided as a pure value. To me this is adds significant complication not just syntactically (needing to specify an effect) but (more importantly) pedagogically - programming with effects already involves a learning curve, without needing to explain that a message with a pure body or no body is actually a pure value but we have to covary it up to `F` to make the compiler happy.  


This seems like it will just make writing stuff around these models more difficult,
> 
> 
> 
> That being said there is continuing interested in making it covariant so I don't really know what to say. It works, I just find programming with variance central to the model to be unpleasant.

One thing I'd say here is that variance is already central to the model, hence the `covary` method. The variance is just currently encoded as an unchecked cast rather than in a first-class way.

Are you taking about making development of http4s code harder, or client code, or both? Based on this and #5251, the cost of these changes seems to be to add some extra boilerplate to defining instance methods with effectful arguments and to make type inference worse in a handful of relatively advanced use cases - people can reasonably disagree on the tradeoffs involved, but I just don't see how those costs outweigh the benefits in terms of making the basic API more approachable to people new to effects / tagless final.

It's very likely there are things I've missed here, so keen to hear more on this. (I think gut reactions of people with good judgement should be taken seriously, so don't want to seem dismissive.)

bplommer:
Fun times: this seems to have hit a regression in Scala 2.13.7

.github/workflows/ci.yml
```diff
@@ -28,7 +28,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        scala: [2.13.7, 2.12.15, 3.1.0]
```
bplommer, line 31: Obviously we won't merge this with the version rollback in place.

rossabaker:
I'm too underwater to give a detailed review this morning, but if we owned both `Message` and `Stream`, would there be any argument for keeping the former invariant while making the latter covariant?  Or do they have exactly the same pros and cons?

bplommer:
@rossabaker I’d lean towards saying same pros and cons. There’s been some extended discussion on Discord - https://discord.com/channels/632277896739946517/839254871646404638/920628041158381638

rossabaker:
I'm having trouble following that link, so I'll let you all decide. I'm mildly :+1: on the concept. It works well in fs2.

bplommer:
I think the consensus on Discord was that we should go ahead with this provided it plays nicely with Chris's entity changes in #4509. (I'm going to try combining them on a new branch to see.)

Everyone seems to agree that it's an improvement for library users, but we do need to be mindful that the change creates an  extra cognitive barrier to developing `Media` and its subtypes - we should add some developer documentation on how to deal with variance-related compiler errors.

Is this a fair summary @ChristopherDavenport ?

bplommer:
I tried doing these changes against Chris's branch and it worked out fine: https://github.com/ChristopherDavenport/http4s/compare/entity...bplommer:daventity-covariant

core/shared/src/main/scala/org/http4s/Message.scala
```diff
@@ -224,6 +224,8 @@ sealed trait Message[+F[_]] extends Media[F] { self =>
 }
 
 object Message {
+  // workaround for an apparent regression in Scala 2.13.8 - Media.InvariantOps should be resolved directly.
+  implicit def toMediaInvariantOps[F[_]](self: Media[F]): Media.InvariantOps[F] = new Media.InvariantOps[F](self)
   implicit final class InvariantOps[F[_], ThisF[f[_]] <: Message[f]](val self: ThisF[F]) {
```
armanbilge, line 229:
```suggestion
  final class InvariantOps[F[_], ThisF[f[_]] <: Message[f]](val self: ThisF[F]) {
```

armanbilge: This might cause ambiguity issues maybe?

bplommer: If anything the ambiguity would be with the implicit constructor on `Media`, not this one, no?

armanbilge: Sorry, you are right!

bplommer:
I got this compiling on my machine with Scala 2.13.8-bin-b59fa66 - had to locally publish kind-projector to do it. I can't find a good workaround (the only thing that seems to work is to explicitly import the `InvariantOps` class where it's needed so let's just hold off on this until 2.13.8 is released.

rossabaker:
Will this work for apps below 2.13.8?  2.13.8 is only necessary to compile the library?

bplommer:
I assume apps using 2.13.7 will be potentially affected by the regression (the compile error is in a test that exercises the library, not in the library itself), but I think it should be fine for apps on 2.13.6 and earlier.

core/shared/src/main/scala/org/http4s/Message.scala
```diff
@@ -203,21 +198,21 @@ sealed trait Message[F[_]] extends Media[F] { self =>
   /** Generates a new message object with the specified key/value pair appended
     * to the [[attributes]].
     *
-    * @param key [[org.typelevel.vault.Key]] with which to associate the value
+    * @param key [[io.chrisdavenport.vault.Key]] with which to associate the value
```
bplommer, line 201:
```suggestion
    * @param key [[org.typelevel.vault.Key]] with which to associate the value
```

bplommer:
This needs some more comments and then should be good to go.

core/shared/src/main/scala/org/http4s/Message.scala
```diff
@@ -398,9 +395,6 @@ final class Request[F[_]] private (
 
   def remoteAddr: Option[IpAddress] = remote.map(_.host)
 
-  def remoteHost(implicit F: Monad[F], dns: Dns[F]): F[Option[Hostname]] =
-    OptionT.fromOption(remote.map(_.host)).flatMapF(dns.reverseOption).value
```
armanbilge, line 402: Can this be written like this?
```scala
def remoteHost[F2[x] >: F[x]](implicit F: Monad[F2], dns: Dns[F2]): F2[Option[Hostname]]
```

bplommer: Yep absolutely. Probably better to do this one that way.

core/shared/src/main/scala/org/http4s/Media.scala
```diff
@@ -47,36 +39,50 @@ trait Media[F[_]] {
...
 
 object Media {
+
+  implicit final class InvariantOps[F[_]](private val self: Media[F]) extends AnyVal {
+    import self._
+
+    def bodyText(implicit
+        RT: RaiseThrowable[F],
+        defaultCharset: Charset = `UTF-8`,
+    ): Stream[F, String] = {
```
armanbilge, line 52: `F2` for this one too?

core/shared/src/main/scala/org/http4s/Media.scala
```diff
@@ -47,36 +39,50 @@ trait Media[F[_]] {
...
+
+    // Decoding methods
+
+    /** Decode the [[Media]] to the specified type
+      *
+      * @param decoder [[EntityDecoder]] used to decode the [[Media]]
+      * @tparam T type of the result
+      * @return the effect which will generate the `DecodeResult[T]`
+      */
+    def attemptAs[T](implicit decoder: EntityDecoder[F, T]): DecodeResult[F, T] =
```
armanbilge, line 65: Does the `F2` thing work for this one?

core/shared/src/main/scala/org/http4s/Message.scala
armanbilge: Actually, can't many several of the methods currently in invariant ops be written like this?

core/shared/src/main/scala/org/http4s/Media.scala
bplommer: Not this one, as it’s intended to be called with exactly one explicit type argument (e.g. attemptAs[String]). But the ones with no type argument, yes.

armanbilge: Ugly, but this seems like a case for http://typelevel.org/cats/guidelines.html#-partially-applied-type.

bplommer: I don’t think that works without an explicit value argument?

armanbilge: Oh yeah. Probably not :(

core/shared/src/main/scala/org/http4s/Message.scala
```diff
@@ -490,6 +458,38 @@ final class Request[F[_]] private (
 
 object Request {
 
+  /** A [[Request]] that doesn't have a body requiring effectful evaluation.
+    *  A [[Request.Pure]] is always an instance of [[Request[F]]], regardless of `F`.
+    */
+  type Pure = Request[fs2.Pure]
+
+  implicit final class InvariantOps[F[_]](private val self: Request[F]) extends AnyVal {
```
armanbilge, line 466: What about the methods in here? 😅

bplommer: These ones could be done as partial application I think. I'm not too bothered either way but maybe we should do that as it's a more familiar pattern.

bplommer:
@rossabaker feel free to ignore the tag and go ahead with the nest milestone release - I'm being egregiously slow on this.

rossabaker:
I was egregiously slow on the milestone, but just came looking for this.  

I'm still going through tickets that piled up during the holiday malaise, so there's going to be another release train soon.

core/shared/src/main/scala/org/http4s/Message.scala
bplommer: I have this a try, but the problem is that the helper classes can’t extend `AnyVal` because they would be inner classes - and if they’re defined outside the class they would need to have an `F` parameter, which would have to occur in contravariant position at the call site where they’re instantiated, which wouldn’t compile…

let’s just leave it as-is?

armanbilge: Thanks! I gave it a try myself, this is what I came up with. Let me know what you think :)

- https://github.com/bplommer/http4s/pull/1

bplommer: I'm not wild about that change. Feels like it degrades the user experience for the sake of making implementation slightly simpler. I'm hilling to be outvoted on this though if it's what other people prefer.

Kudos for using git the way it's supposed to be used by opening a PR against my PR though

armanbilge: > Feels like it degrades the user experience for the sake of making implementation slightly simpler.

Well, there's trade-offs. IMO it makes the UX better in terms of discoverability. When those methods are hidden on implicit classes they are trickier to find.

bplommer: That’s a fair point. We could try your change and see if people complain 😅

tests/shared/src/test/scala/org/http4s/EntityDecoderSuite.scala
```diff
@@ -410,7 +410,7 @@ class EntityDecoderSuite extends Http4sSuite {
       .tempFile(None, "foo", "bar", None)
       .use { tmpFile =>
         val response = mockServe(Request()) { req =>
-          req.decodeWith(EntityDecoder.textFile(tmpFile), strict = false) { _ =>
+          req.decodeWith(EntityDecoder.textFile[IO](tmpFile), strict = false) { _ =>
```
armanbilge, line 413: I noted this in https://github.com/bplommer/http4s/pull/1#discussion_r799842833 but I'll repeat here:
> This (and the similar changes) is only necessary for 2.13. Scala 3 seems to be able to infer this just fine.

ember-server/shared/src/test/scala/org/http4s/ember/server/EmberUnixSocketSuite.scala
```diff
@@ -46,8 +46,8 @@ class EmberUnixSocketSuite extends Http4sSuite {
         .build
         .map(UnixSocket(localSocket))
       _ <- Resource.eval(IO.sleep(4.seconds))
-      resp <- client.run(Request[IO](Method.GET))
-      body <- Resource.eval(resp.bodyText.compile.string)
+      resp <- client.run(Request(Method.GET))
+      body <- Resource.eval(resp.bodyText[IO].compile.string)
```
armanbilge, line 50: This is annoying. Scala 3 wants it, but not 2.13.

bplommer: Yeah, but I'd much rather have it needed where something effectful is actually being done.

bplommer:
I think this one's finally ready.

core/shared/src/main/scala/org/http4s/Message.scala
```diff
@@ -678,6 +667,20 @@ final class Response[F[_]] private (
...
+  type Pure = Response[fs2.Pure]
+
+  /** An implicit conversion to covary pure `Response` instances up to an effect `G`.
+    * Improves type inference when constructing responses in an effectful context.
+    */
+  @nowarn("msg=never used")
+  implicit def covaryF[F[_], G[_]](fresp: F[Response[Nothing]])(implicit
+      F: Functor[F]
+  ): F[Response[G]] =
+    fresp.asInstanceOf
```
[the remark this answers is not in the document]
bplommer: > Can this be written as `.widen`?

Yep. It doesn't actually seem to be needed in the current PR though (maybe it was only ever needed for Scala 2.12 🤔), so let's leave it out for now.

armanbilge (approved):
Fantastic. Thank you Ben for championing these changes!

bplommer:
Let's leave this until after #6193.

armanbilge:
I'll merge this one tonight, if nobody has any objections and/or wants more time to review.

I think we all roughly agree this is the direction we want to move in, and I reviewed and the changes here look good. There are certainly still details to bikeshed (and some of that may depend on feedback from downstreams/users) and there's plenty of time for that before 1.0.

Most of all, let's keep this momentum going!! 🚀

danicheg (approved):
I think that bringing covariance slightly complicates reasoning for users. Particularly for newcomers to FP Scala. Though, considering `fs2` exploits that approach, probably we also can use it within `http4s`.

core/shared/src/main/scala/org/http4s/Message.scala
```diff
@@ -443,22 +430,22 @@ final class Request[F[_]] private (
...
-      f: A => F[Response[F]]
-  )(implicit F: Monad[F]): F[Response[F]] =
+  def decodeWith[F2[x] >: F[x], A](decoder: EntityDecoder[F2, A], strict: Boolean)(
+      f: A => F2[Response[F2]]
+  )(implicit F: Monad[F2]): F2[Response[F2]] =
     decoder
       .decode(this, strict = strict)
-      .fold(_.toHttpResponse[F](httpVersion).pure[F], f)
+      .fold(_.toHttpResponse(httpVersion).covary[F2].pure[F2], f)
       .flatten
```
diesalbla, line 439:
```suggestion
      .foldF( e => F.pure(e.toHttpResponse(httpVersion)), f)
```

diesalbla (commented):
A few small comments.

core/shared/src/main/scala/org/http4s/Message.scala
```diff
@@ -503,6 +490,11 @@ final class Request[F[_]] private (
 
 object Request {
 
+  /** A [[Request]] that doesn't have a body requiring effectful evaluation.
+    *  A [[Request.Pure]] is always an instance of [[Request[F]]], regardless of `F`.
+    */
+  type Pure = Request[fs2.Pure]
```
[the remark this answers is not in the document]
bplommer: No, that won't work because fs2.Pure is defined as a type inequality: `type Pure[A] <: Nothing`. Even if we define our own `Pure` type in the exact same way, the compiler won't unify them. The only way I can see to avoid referring to fs2 here is to push the definition up into Cats.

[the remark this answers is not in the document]
armanbilge: > The only way I can see to avoid referring to fs2 here is to push the definition up into Cats.

I think that's worth a PR.

core/shared/src/main/scala/org/http4s/Entity.scala
```diff
@@ -34,7 +34,7 @@ object Entity {
   def apply[F[_]](body: EntityBody[F], length: Option[Long] = None): Entity[F] =
     Default(body, length)
 
-  // The type parameter aids type inference in Message constructors. This should be unnecessary when Messages are covariant.
+  // The type parameter aids type inference in Message constructors.
   def empty[F[_]]: Entity[F] = Empty
```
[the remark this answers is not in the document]
bplommer: I tried removing the type parameter but it's still needed. An alternative would be to remove the `entity` parameter from the Message constructors, and only allow entities to be provided via `withEntity` etc - that way the constructors themselves wouldn't need an `F` parameter and would always return `Pure`.

core/shared/src/main/scala/org/http4s/Message.scala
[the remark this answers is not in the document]
armanbilge: But also, could we define our own `type Pure[A] = fs2.Pure[A]`?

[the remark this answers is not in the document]
bplommer: > I think that's worth a PR.
Sure, I'll give it a go

> But also, could we define our own `type Pure[A] = fs2.Pure[A]`?
There's nothing to stop us - I'm not wild about the extra indirection though.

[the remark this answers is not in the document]
bplommer: I opened a PR at https://github.com/typelevel/cats/pull/4158 but I don't know if I did a very good job of explaining the rationale.

core/shared/src/main/scala/org/http4s/Media.scala
```diff
@@ -46,37 +46,41 @@ trait Media[F[_]] {
...
+  }
+
   def apply[F[_]](e: Entity[F], h: Headers): Media[F] =
     new Media[F] {
       def entity: Entity[F] = e
 
       def headers: Headers = h
 
-      override def covary[F2[x] >: F[x]]: Media[F2] = this.asInstanceOf[Media[F2]]
+      override def covary[F2[x] >: F[x]]: Media[F2] = this
```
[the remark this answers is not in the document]
bplommer: Not sure about this. I see your point, but I do think it's convenient to have this chaining-friendly syntax. Willing to be outvoted on this, but I lean towards keeping it.

core/shared/src/main/scala/org/http4s/Entity.scala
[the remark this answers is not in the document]
bplommer: Let's leave that potential change out of this PR though.

core/shared/src/main/scala/org/http4s/Media.scala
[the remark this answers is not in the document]
diesalbla: In `fs2` we kept it for this reason, but it is a kind of construct I would rather people avoided.

core/shared/src/main/scala/org/http4s/Message.scala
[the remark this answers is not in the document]
SystemFw: I mean, does it matter that we expose an fs2 type? `Stream` is exposed all over the place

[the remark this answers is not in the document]
bplommer: I don't personally have strong feelings about it. I do think that as a matter of principle, cats is a better place for it to be defined - it should in principle be possible for another library that doesn't have an fs2 dependency to be able to use `Pure`/`INothing` and be fully interoperable with the fs2 types, which isn't currently possible.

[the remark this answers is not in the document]
mpilquist: This seems like a lot of change for little relative benefit.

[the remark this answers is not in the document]
armanbilge: Why a lot of change? Cats adds it, fs2 updates its aliases, nobody is none the wiser.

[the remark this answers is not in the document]
SystemFw: What's the benefit though?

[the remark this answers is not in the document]
armanbilge: We should move discussion to https://github.com/typelevel/cats/pull/4158.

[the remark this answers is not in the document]
bplommer: After a bit more fiddling, I think we may be less tied to `fs2.Pure` than I thought, rendering this moot.

core/shared/src/main/scala/org/http4s/Media.scala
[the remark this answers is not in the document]
diesalbla: Regarding I found a few places where, thanks to type covariance, the method is no longer needed, either by removing it directly or with some rewrites... https://github.com/http4s/http4s/pull/6230/commits/9b4e45c8ac6d4ea4d084cd906fa2a11ff1f74f30

Feel free to incorporate those changes to this PR.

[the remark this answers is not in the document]
bplommer: Thanks! The idea with this PR is to keep it as small as possible but we should definitely follow up with more simplicifafions.

core/shared/src/main/scala/org/http4s/package.scala
```diff
@@ -81,4 +81,5 @@ package object http4s {
   /** A stream of server-sent events */
   type EventStream[F[_]] = Stream[F, ServerSentEvent]
 
+  type PureF[A] = Nothing
```
armanbilge, line 84: Sorry to be re-litigating this 😆 but what's the purpose of `PureF` vs just using `Nothing`? Unless it's a technical thing, IIUC since it's a type alias compile errors will just show `Nothing` anyway, right?

bplommer: It's because of https://github.com/lampepfl/dotty/issues/14640 - Scala 3 gets confused with about kinds with `Nothing`. I don't like how this is currently done in this PR though - I'm tempted to revert to just using `fs2.Pure`. I'm also having second thoughts about the type aliases for `Response.Pure` etc - I'm thinking we should maybe just use `Response[fs2.Pure]. Any thoughts on that?

diesalbla: Well, since the `F[_]` is the parameter for `fs2`, it makes sense to use the type alias of `fs2`.

armanbilge: Gotcha, thanks for the pointer. Yeah, maybe let's dial back the aliases for now. This PR does a great thing at a high level (I wanna land it!) and at the moment these type issues just seem a bit in the mud.

bplommer: 👍 I'm also thinking about removing the existing `EntityEncoder.Pure` alias - what do you think?

diesalbla: @bplommer Good enough is good enough, and the changes in this PR are relevant. Bike-shedding on type aliases may wait.

bplommer:
I think we're ready to merge this 🤞

armanbilge:
Yes we are!! 🚀

PR #5723, opened 2021-12-12
Speed up shutdown of NioEventLoopGroups

RafalSumislawski:
The default we currently use takes 2 seconds.

### DIFF
--- a/client/src/test/scala/org/http4s/client/scaffold/NettyTestServer.scala
+++ b/client/src/test/scala/org/http4s/client/scaffold/NettyTestServer.scala
@@ -38,6 +38,7 @@ import org.log4s.getLogger
 
 import java.net.InetAddress
 import java.net.InetSocketAddress
+import java.util.concurrent.TimeUnit
 import javax.net.ssl.SSLContext
 
 trait TestServer[F[_]] {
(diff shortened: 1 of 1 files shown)


client/src/test/scala/org/http4s/client/scaffold/NettyTestServer.scala
```diff
@@ -93,7 +94,9 @@ object NettyTestServer {
   } yield new NettyTestServer(establishedConnections, localAddress)
 
   private def nioEventLoopGroup[F[_]](implicit F: Async[F]): Resource[F, NioEventLoopGroup] =
-    Resource.make(F.delay(new NioEventLoopGroup()))(el => F.delay(el.shutdownGracefully()).liftToF)
+    Resource.make(
+      F.delay(new NioEventLoopGroup())
+    )(el => F.delay(el.shutdownGracefully(0, 0, TimeUnit.MILLISECONDS)).liftToF)
```
danicheg, line 99: Should we care about shutdown timeout? Maybe it's not so bad to set some value?

RafalSumislawski: As far as I understand the `timeout` does nothing when the `quietPeriod` is 0.

The way this works is that when we start shutting down the `EventLoop`. It only shut down if there are no new asks submitted for the `quietPeriod`. If a new tasks is submitted after we called `shutdownGracefully` then it resets the timer. No we wait until `quietPeriod` after the last task submission. If the tasks keep coming, this may take forever. That's where the `timeout` kicks in. So the `timeout` is really a time limit for the `quietPeriod` mechanism. There is no timer running in parallel to the shutdown process and interrupting it if it takes to long or something like that.

If you're interested, the logic is in the bottom half of `io.netty.util.concurrent.SingleThreadEventExecutor#confirmShutdown`.

danicheg (approved):

PR #5724, opened 2021-12-12
Remove "flaky" from blaze-client tests and see what happens

RafalSumislawski:
Based on #5385 which hopefully fixes the root causes of the flakiness.

PR #5728, opened 2021-12-12
Mark the failing tests as flaky because they sometimes pass

RafalSumislawski:
It turns out these tests I marked with `fail` sometimes pass on CI, leading to build failure.
(https://github.com/http4s/http4s/runs/4498932845?check_suite_focus=true)


danicheg (approved):
I opine it's good to have a more reliable build; what's more, you're working on that topic in #5724.

PR #5729, opened 2021-12-12
Remove some outdated params in scaladocs

danicheg:
I initially planned to fix only `blocker` but ended up checking all outdated params in scaladocs.

### DIFF
--- a/server/shared/src/main/scala/org/http4s/server/staticcontent/WebjarService.scala
+++ b/server/shared/src/main/scala/org/http4s/server/staticcontent/WebjarService.scala
@@ -30,8 +30,6 @@ import scala.util.control.NoStackTrace
 
 /** [[org.http4s.server.staticcontent.WebjarServiceBuilder]] builder
   *
-  * @param blocker execution context for blocking I/O
-  * @param filter To filter which assets from the webjars should be served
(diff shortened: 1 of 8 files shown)


server/shared/src/main/scala/org/http4s/server/staticcontent/WebjarService.scala
```diff
@@ -30,8 +30,6 @@ import scala.util.control.NoStackTrace
 
 /** [[org.http4s.server.staticcontent.WebjarServiceBuilder]] builder
   *
-  * @param blocker execution context for blocking I/O
-  * @param filter To filter which assets from the webjars should be served
```
armanbilge, line 34: This parameter is still here, just renamed :)

danicheg: It looks so. As an excuse - I was only about removing outdated params. But yeah, some scaladocs are needed solid refactoring.

armanbilge: No excuses needed! You've been doing fantastic work cleaning all this up, let's not let it be thankless 😉

armanbilge (approved):

PR #5730, opened 2021-12-13

danicheg:
It turns out that some tests in the `BlazeClientConnectionReuseSuite` ain't failed now, though it was intended by test-suite:
`BlazeClient should reuse the connection after a successful request with large response`;
`BlazeClient should reuse the connection after response decoding failed and the (large) entity was drained`
I marked them as flaky (just because I'm not sure if they are stable enough).
cc @RafalSumislawski

### DIFF
--- a/blaze-client/src/test/scala/org/http4s/blaze/client/BlazeClientBase.scala
+++ b/blaze-client/src/test/scala/org/http4s/blaze/client/BlazeClientBase.scala
@@ -20,10 +20,13 @@ package client
 import cats.effect._
 import cats.effect.kernel.Resource
 import cats.effect.std.Dispatcher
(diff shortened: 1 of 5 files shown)


RafalSumislawski:
Chances are the tests were fixed by some changes in 0.23.0, but my skepticism makes me think that some changes in 0.23 broke the tests ;).

danicheg:
@RafalSumislawski Do you plan to dig into it? Or can we go with those changes from `series/0.22`?

RafalSumislawski:
I just dug into that topic. These tests were failing due to a race condition, so literally anything could have fixed them. I'm for marking them as `flaky` instead of the `fail`.

RafalSumislawski (approved):

PR #5742, opened 2021-12-15
Add `apply`-builder for `Http4sClientDsl`

armanbilge:
Adds an `apply`-builder for convenience. Also moves the file to match the package location.

### DIFF
--- a/client/src/main/scala/org/http4s/client/dsl/Http4sClientDsl.scala
+++ b/client/src/main/scala/org/http4s/client/dsl/Http4sClientDsl.scala
@@ -36,6 +36,10 @@ trait Http4sClientDsl[F[_]] {
   }
 }
 
+object Http4sClientDsl {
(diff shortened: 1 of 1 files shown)


client/src/main/scala/org/http4s/client/dsl/Http4sClientDsl.scala
```diff
@@ -36,6 +36,10 @@ trait Http4sClientDsl[F[_]] {
   }
 }
 
+object Http4sClientDsl {
+  def apply[F[_]]: Http4sClientDsl[F] = new Http4sClientDsl[F[_]] {}
```
bplommer, line 40:
```suggestion
  def apply[F[_]]: Http4sClientDsl[F] = new Http4sClientDsl[F] {}
```

armanbilge: Woops, thanks for the assist!

bplommer (approved):
LGTM

PR #5754, opened 2021-12-17
Minimal slice from #4509 (Entity rethinking)

bplommer:
I've taken the liberty of slicing part of #4509 into a standalone PR to try and get the ball rolling again. This slice just changes `Message` to use `Entity` instead of `EntityBody` without actually making `Entity` more opinionated - so it doesn't yet provide any benefits, but it's a more manageable diff that lays groundwork for converting `Entity` to an ADT in subsequent slices.

### DIFF
--- a/core/shared/src/main/scala/org/http4s/Entity.scala
+++ b/core/shared/src/main/scala/org/http4s/Entity.scala
@@ -20,14 +20,11 @@ import cats.Monoid
 import cats.syntax.all._
 import cats.~>
 
-final case class Entity[+F[_]](body: EntityBody[F], length: Option[Long] = None)
+final case class Entity[+F[_]](body: EntityBody[F], length: Option[Long] = None) {
+  def translate[F1[x] >: F[x], G[_]](fk: F1 ~> G): Entity[G] = Entity(body.translate(fk), length)
+}
 
 object Entity {
-  implicit class InvariantOps[F[_]](private val self: Entity[F]) extends AnyVal {
-    import self._
-
-    def translate[G[_]](fk: F ~> G): Entity[G] = Entity(body.translate(fk), length)
-  }
 
   implicit def http4sMonoidForEntity[F[_]]: Monoid[Entity[F]] =
     new Monoid[Entity[F]] {
--- a/core/shared/src/main/scala/org/http4s/Entity.scala
+++ b/core/shared/src/main/scala/org/http4s/Entity.scala
@@ -18,10 +18,17 @@ package org.http4s
 
 import cats.Monoid
 import cats.syntax.all._
+import cats.~>
 
 final case class Entity[+F[_]](body: EntityBody[F], length: Option[Long] = None)
 
 object Entity {
+  implicit class InvariantOps[F[_]](private val self: Entity[F]) extends AnyVal {
+    import self._
+
+    def translate[G[_]](fk: F ~> G): Entity[G] = Entity(body.translate(fk), length)
+  }
+
   implicit def http4sMonoidForEntity[F[_]]: Monoid[Entity[F]] =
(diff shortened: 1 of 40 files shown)


armanbilge:
This is awesome!! Very digestible slice.

I have a question, sorry if it's a dump one: will/should `Media` itself be changed to use `entity` instead of `body`? Why/why not?

bplommer:
> This is awesome!! Very digestible slice.
> 
> I have a question, sorry if it's a dumb one: will/should `Media` itself be changed to use `entity` instead of `body`? Why/why not?

Yes, it should - I didn't because I overlooked it, will add that.

core/shared/src/main/scala/org/http4s/Entity.scala
```diff
@@ -18,10 +18,17 @@ package org.http4s
 
 import cats.Monoid
 import cats.syntax.all._
+import cats.~>
 
 final case class Entity[+F[_]](body: EntityBody[F], length: Option[Long] = None)
 
 object Entity {
+  implicit class InvariantOps[F[_]](private val self: Entity[F]) extends AnyVal {
```
armanbilge, line 26: Nitpick: implciit conversion via an implicit class vs an implicit method creating an instance of that class? If I had to guess, it's probably easier to control the binary evolution of the latter. But, maybe it doesn't make a difference at all actually 🤔

armanbilge (approved):
I think this is a _very_ nice slice. IIUC, most/all of the changes are mechanical refactoring?

core/shared/src/main/scala/org/http4s/Entity.scala
bplommer: Actually, when I started on the next slice I changed this to an instance method with an extra type parameter, so I'll just make that change now instead.

bplommer:
Yeah, it's all pretty mechanical. 

I'll wait for more reviews or leave this open for a couple of days to give people a chance to object, since this is a fairly major API change.

danicheg (commented):
I exactly missed the discussion about those changes. Sorry for being late. But if you don't mind, I will point my thoughts here. And yeah, I could miss some obvious things.
I will refer to `akka-http` because it's widely adopted in the Scala ecosystem (from my observations). Their models of `Request`/`Response` also have the `Entity`, but it significantly differs from `http4s` one, I think. The data represents by static `ByteString` which length is easily computed without any manipulations with the data itself. As for `http4s`, the data is represented by `Stream`. As we know, there is a complication with getting the length of `Stream` - it's needed a materialization. 
I opine that the actual `Entity` hasn't a proper constructor right now (or `Request`/`Response` should have another model of an `Entity`). It's a bit odd to have non-empty `data` with empty `length` and is error-prone fairly speaking. From my way of thinking, the `length` should be an effectful function in the `Entity` model in the `Response`/`Request`, not a value. Having the `length` as a computed value creates false expectations.

armanbilge:
> I opine that the actual Entity hasn't a proper constructor right now (or Request/Response should have another model of an Entity)

💯 agree that `Entity` needs to be redesigned. The purpose of this PR is to do some refactoring to setup for upcoming changes, which are (at least partly) inspired by akka-http :)

bplommer:
Yep, the idea is to have a structure into which we can incrementally drop in the ADT from #4509 that divides entities into Strict (represented as a chunk), Empty, streaming but with a known size, and chunked (with an unknown size) - having size as a strict optional value should then make more sense. (I don't think it's advisable to expose it as an effectful value for chunked entities, since computing it would require running the stream.)

RafalSumislawski (approved):
I'm very happy seeing this moving forward.

hamnis (approved):

PR #5755, opened 2021-12-17
Code cleanup (mostly variance-related)

bplommer:
Removed a lot of unneeded `covary` calls, plus a couple of other minor tweaks


armanbilge (approved):
Thanks!

PR #5757, opened 2021-12-18
Remove jvm log4cats dependency in ember client js

kubukoz:
Also removes the dependency on `server` in `client`, limiting to tests only.

### DIFF
--- a/project/Http4sPlugin.scala
+++ b/project/Http4sPlugin.scala
@@ -292,7 +292,7 @@ object Http4sPlugin extends AutoPlugin {
   lazy val jnrUnixSocket = "com.github.jnr" % "jnr-unixsocket" % V.jnrUnixSocket
   lazy val keypool = Def.setting("org.typelevel" %%% "keypool" % V.keypool)
   lazy val literally = Def.setting("org.typelevel" %%% "literally" % V.literally)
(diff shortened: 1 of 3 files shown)


project/Http4sPlugin.scala
```diff
@@ -292,7 +292,7 @@ object Http4sPlugin extends AutoPlugin {
   lazy val jnrUnixSocket = "com.github.jnr" % "jnr-unixsocket" % V.jnrUnixSocket
   lazy val keypool = Def.setting("org.typelevel" %%% "keypool" % V.keypool)
   lazy val literally = Def.setting("org.typelevel" %%% "literally" % V.literally)
-  lazy val log4catsCore = "org.typelevel" %% "log4cats-core" % V.log4cats
+  lazy val log4catsCore = Def.setting("org.typelevel" %% "log4cats-core" % V.log4cats)
```
armanbilge, line 295:
```suggestion
  lazy val log4catsCore = Def.setting("org.typelevel" %%% "log4cats-core" % V.log4cats)
```

kubukoz: wtf. The dep was gone :joy: how is this still here

armanbilge (approved):

PR #5152, opened 2021-09-02
Pass WebSocketBuilder to HttpApp in builders

tpolecat:
This resolves #5108 by changing the builders to provide a compatible `WebSocketBuilder[F]` that will use the same vault `Key` as the server. The API changes are:

- The only way to get a `WebSocketBuilder` is from your server builder, by calling `.withHttpWebSocketApp(f: WebSocketBuilder[F] => HttpApp[F])`.
- `WebSocketBuilder` now has `imapK`, which allows downstream users to adapt it to contexts in which the local effect differs.

Another change is that Blaze has an annoying `webSocketEnabled` boolean, which (if it's false) will cause your app to blow up if you try to use websockets. This boolean saves a grand total of one hash lookup per request so I removed it as a simplification.

Waiting on:
- [x] https://github.com/typelevel/vault/pull/267

PR #5758, opened 2021-12-19
cross-platforming server FileService

yurique:
Updating `FileService` and its configuration to use `fs2.Path` instead of `java.io.File` allowing it to be used in the nodejs environment. 

- [x] update fs2 to a non-snapshot version 
- [ ] make sure `since` in deprecations is correct


yurique:
Not sure if changes like these preserve binary compat. 

Also, one test is failing in `serverJS`, but passes with this change: https://github.com/typelevel/fs2/pull/2759

build.sbt
```diff
@@ -342,7 +342,9 @@ lazy val server = libraryCrossProject("server")
   .settings(BuildInfoPlugin.buildInfoScopedSettings(Test))
   .settings(BuildInfoPlugin.buildInfoDefaultSettings)
   .settings(
-    buildInfoKeys := Seq[BuildInfoKey](Test / resourceDirectory),
+    buildInfoKeys := Seq[BuildInfoKey](
+      BuildInfoKey.map(Test / resourceDirectory) { case (k, v) => k -> v.toString }
```
yurique, line 346: by default it maps to `java.io.File` – doesn't work in JS

armanbilge:
Thanks @yurique!

Regarding the `FileService`, besides that split files make merges difficult, it also means that JVM/JS no longer have the same API which becomes difficult for users who want cross-build for both platforms without having to split their sources too.

Instead, I propose we do something like this:
```scala
// we'll bikeshed the name :)
type PathCollector2[F[_]] = (fs2.io.file.Path, Config[F], Request[F]) => OptionT[F, Response[F]]
```
Then we can:
1. deprecate the old `File`-based `PathCollector`.
2. re-work `Config[F[_]]` to use `PathCollector2`. For compatibility, we'll have to add an `apply` method using the old `PathCollector` and also a getter method that converts ...

2. might be too annoying. In which case, maybe we also just make a `Config2[F]` 🙃 WDYT?

yurique:
> WDYT?

I like how this sounds :) 

(and also I don't really like the split, but didn't know what to do about the `PathCollector` and wanted to quickly see if I can make it run)

While being AFK today, I also thought about something like you've suggested – in JS we won't be able to use the deprecated stuff, but it will be deprecated for a reason :)

I'm gonna give it a try!

yurique:
@armanbilge I'd rather not introduce `Config2`, but changing `Config` *will* break bin-compat, right? (if someone reads the  `pathCollector` field for some reason).

armanbilge:
It should be possible to modify `Config` without breaking bincompat, but I did say it might be annoying :P

You are right, we'd have to rename the `pathCollector` field to something else e.g. `pathCollector2`, then define a method:
```scala
def pathCollector: PathCollector = transform(pathCollector2) // changes between fs2 Path and Java File
```

We'll have to do a similar trick for the constructor.

armanbilge:
Your fs2 bugfix is in `3.2-32-1a7dede`, should be on maven shortly.

yurique:
Good idea.
I did this, but got a problem :'( – the two `Config.apply` methods (the generated one, and the one I've added with the compatible signature) – they both have the same signature after erasure: `systemPath: String, pathCollector: Function3, ...`.

As a workaround I made it a trait:
```scala
trait PathCollector2[F[_]] extends ((Path, Config[F], Request[F]) => OptionT[F, Response[F]])
```

and a constructor:

```scala
PathCollector2 {
    def apply[F[_]](
        f: (Path, Config[F], Request[F]) => OptionT[F, Response[F]]
    ): PathCollector2[F] =
      (path, config, request) => f(path, config, request)
  }
```

Which makes it less convenient:

```scala
PathCollector2 { (path: Path, config: Config[F], request: Request[F]) => ??? }
```

So I added an implicit conversion (and it works). 
Not sure, though, if this is a good thing to do.

```scala
  object PathCollector2 {
     // ...
    implicit def fromFunction[F[_]](
        f: (Path, Config[F], Request[F]) => OptionT[F, Response[F]]
    ): PathCollector2[F] = PathCollector2(f)

  }
```

armanbilge:
> the two Config.apply methods (the generated one, and the one I've added with the compatible signature) – they both have the same signature after erasure:

What if we just, re-order the arguments 😆

yurique:
> What if we just, re-order the arguments 😆

You've got a lot of tricks up your sleeve, right? xD

Pushing!

yurique:
Is there something I can do about those CI failures due to the `PathCollector` being deprecated but is still used?

armanbilge:
Would it be easy to change those to use the new APIs? Otherwise we can `@nowarn` them

yurique:
> Would it be easy to change those to use the new APIs? Otherwise we can @nowarn them

The deprecated `PathCollector` is only used in the for-the-bin-compat `apply` :) (which is also deprecated).

I'll try no-warning it.

armanbilge:
I _thought_ that as long as deprecated things are only used by other deprecated things, it should be fine.

server/shared/src/main/scala/org/http4s/server/staticcontent/FileService.scala
```diff
@@ -62,6 +63,7 @@ object FileService {
     /** For binary compatibility.
       * @return an instance of PathCollector[F] created (converted) from pathCollector2
       */
+    @nowarn
```
armanbilge, line 66: Should this be deprecated?

```suggestion
    @deprecated
```

yurique: yes, that's the one I missed

yurique:
I might have misread the compilation errors. There's one non-deprecated usage (which should be deprecated, but I missed it).

yurique:
Bin-compat got broken after all :)

armanbilge:
Now, the real fun begins :P

```
[error]  * method copy(java.lang.String,scala.Function3,java.lang.String,Int,org.http4s.server.staticcontent.CacheStrategy)org.http4s.server.staticcontent.FileService#Config in class org.http4s.server.staticcontent.FileService#Config's type is different in current version, where it is (java.lang.String,scala.Function3,java.lang.String,org.http4s.server.staticcontent.CacheStrategy,Int)org.http4s.server.staticcontent.FileService#Config instead of (java.lang.String,scala.Function3,java.lang.String,Int,org.http4s.server.staticcontent.CacheStrategy)org.http4s.server.staticcontent.FileService#Config
[error]    filter with: ProblemFilters.exclude[IncompatibleMethTypeProblem]("org.http4s.server.staticcontent.FileService#Config.copy")

[error]  * synthetic method copy$default$4()Int in class org.http4s.server.staticcontent.FileService#Config has a different result type in current version, where it is org.http4s.server.staticcontent.CacheStrategy rather than Int
[error]    filter with: ProblemFilters.exclude[IncompatibleResultTypeProblem]("org.http4s.server.staticcontent.FileService#Config.copy$default$4")

[error]  * synthetic method copy$default$5()org.http4s.server.staticcontent.CacheStrategy in class org.http4s.server.staticcontent.FileService#Config has a different result type in current version, where it is Int rather than org.http4s.server.staticcontent.CacheStrategy
[error]    filter with: ProblemFilters.exclude[IncompatibleResultTypeProblem]("org.http4s.server.staticcontent.FileService#Config.copy$default$5")

[error]  * method this(java.lang.String,scala.Function3,java.lang.String,Int,org.http4s.server.staticcontent.CacheStrategy)Unit in class org.http4s.server.staticcontent.FileService#Config's type is different in current version, where it is (java.lang.String,scala.Function3,java.lang.String,org.http4s.server.staticcontent.CacheStrategy,Int)Unit instead of (java.lang.String,scala.Function3,java.lang.String,Int,org.http4s.server.staticcontent.CacheStrategy)Unit
[error]    filter with: ProblemFilters.exclude[IncompatibleMethTypeProblem]("org.http4s.server.staticcontent.FileService#Config.this")
```

The last one, maybe we can just filter. It relates to the `new` constructor which nobody uses AFAIK.

The first three, are more annoying :) relating to the `copy` method and it's default arguments.

One idea, is that we should make `Config` no longer a `case class`, since that's bad for bincompat anyway. Then, you can manually implement a deprecated `copy` method exactly like the old one, to fix the bincompat problem.

server/shared/src/main/scala/org/http4s/server/staticcontent/FileService.scala
```diff
@@ -70,10 +96,102 @@ object FileService {
...
+          pathPrefix == that.pathPrefix &&
+          cacheStrategy == that.cacheStrategy &&
+          bufferSize == that.bufferSize
+      case _ => false
+    }
+
+    override def hashCode(): Int = {
+      val state = Seq(systemPath, pathCollector2, pathPrefix, cacheStrategy, bufferSize)
+      state.map(_.hashCode()).foldLeft(0)((a, b) => 31 * a + b)
+    }
```
armanbilge, line 153: Ooph, this was a yak. Looks like you did a fantastic job preserving behavior here, serious props.

yurique: IDEA helped me with this one :) (adding `[Any]` in the next commit (together with an extra `@nowarn`)  to make scalac happy.

armanbilge:
I think one last question is whether we can find a better name for `PathCollector2`, maybe `FS2PathCollector` 🤔 it's annoying, the old one should have been called `FileCollector` maybe ... any thoughts?

FWIW we got a `WebSocketBuilder2` in https://github.com/http4s/http4s/pull/5152#pullrequestreview-761457692 so keeping this wouldn't be unprecedented.

yurique:
> FWIW we got a `WebSocketBuilder2`

That's what allowed me to not feel overly bad about that `2` :) 

But `FS2PathCollector` sounds nicer, can change it. How about `val pathCollector2` etc?

armanbilge:
Yes, whatever we decide it should be consistent. We can run with `FS2PathCollector` for now (or maybe `Fs2PathCollector`...) and see if anyone has better proposal.

Wow, ci ✅ ! amazing work!!

yurique:
> ci ✅ !
Yay!

yurique:
Made it `fs2PathCollector: Fs2PathCollector[F]`,

server/shared/src/main/scala/org/http4s/server/staticcontent/FileService.scala
```diff
@@ -33,39 +33,206 @@ import org.log4s.getLogger
 import org.typelevel.ci._
 
 import java.io.File
+import scala.annotation.nowarn
 import scala.util.control.NoStackTrace
 
 object FileService {
   private[this] val logger = getLogger
 
+  @deprecated("use FileService.PathCollector2", "0.23.8")
```
armanbilge, line 42:
```suggestion
  @deprecated("use FileService.Fs2PathCollector", "0.23.8")
```

server/shared/src/main/scala/org/http4s/server/staticcontent/FileService.scala
```diff
@@ -33,39 +33,206 @@ import org.log4s.getLogger
...
+      val cacheStrategy: CacheStrategy[F],
+      val bufferSize: Int,
+  ) extends Product
+      with Serializable
+      with Equals {
+
+    /** For binary compatibility.
+      * @param systemPath path prefix to the folder from which content will be served
+      * @param pathPrefix prefix of Uri from which content will be served
+      * @param pathCollector function that performs the work of collecting the file or rendering the directory into a response.
```
armanbilge, line 68:
```suggestion
      * @param pathCollector function that performs the work of collecting the file or rendering the directory into a response.
      * @param pathPrefix prefix of Uri from which content will be served
```

server/shared/src/main/scala/org/http4s/server/staticcontent/FileService.scala
```diff
@@ -33,39 +33,206 @@ import org.log4s.getLogger
...
+    )
+    override def productArity: Int = 5
+
+    @deprecated(
+      "Config is no longer a case class. The productElement method is provided for binary compatibility.",
+      "0.23.8",
+    )
+    override def productElement(n: Int): Any = n match {
+      case 0 => systemPath
+      case 1 => fs2PathCollector
```
armanbilge, line 134: Since that's what the old API would return :)
```suggestion
      case 1 => pathCollector
```

server/shared/src/main/scala/org/http4s/server/staticcontent/FileService.scala
```diff
@@ -33,39 +33,206 @@ import org.log4s.getLogger
...
+      case 2 => pathPrefix
+      case 3 => cacheStrategy
+      case 4 => bufferSize
```
armanbilge, line 137: This is the old order right?
```suggestion
      case 3 => bufferSize
      case 4 => cacheStrategy
```

server/shared/src/main/scala/org/http4s/server/staticcontent/FileService.scala
```diff
@@ -33,39 +33,206 @@ import org.log4s.getLogger
...
+        systemPath == that.systemPath &&
+          fs2PathCollector == that.fs2PathCollector &&
+          pathPrefix == that.pathPrefix &&
+          cacheStrategy == that.cacheStrategy &&
+          bufferSize == that.bufferSize
+      case _ => false
+    }
+
+    override def hashCode(): Int = {
+      val state = Seq[Any](systemPath, fs2PathCollector, pathPrefix, cacheStrategy, bufferSize)
```
armanbilge, line 156: Same here?
```suggestion
      val state = Seq[Any](systemPath, fs2PathCollector, pathPrefix, bufferSize, cacheStrategy)
```

server/shared/src/main/scala/org/http4s/server/staticcontent/FileService.scala
```diff
@@ -33,39 +33,206 @@ import org.log4s.getLogger
...
+      state.map(_.hashCode()).foldLeft(0)((a, b) => 31 * a + b)
+    }
+
+    override def toString =
+      s"Config($systemPath, $fs2PathCollector, $pathPrefix, $cacheStrategy, $bufferSize)"
```
armanbilge, line 161: and here
```suggestion
      s"Config($systemPath, $fs2PathCollector, $pathPrefix, $bufferSize, $cacheStrategy)"
```

server/shared/src/main/scala/org/http4s/server/staticcontent/FileService.scala
yurique: `⌘F` failed me :'(

server/shared/src/main/scala/org/http4s/server/staticcontent/FileService.scala
yurique: Yeah, that's what I had in my mind when typing this. My typing devices decided to do something else apparently  =)

server/shared/src/main/scala/org/http4s/server/staticcontent/FileService.scala
yurique: Although I'm not sure if this exactly corresponds to what scalac does for `hashCode` for case classes (or if it is consistent across scala versions) 🤔

server/jvm/src/test/scala/org/http4s/server/staticcontent/StaticContentSharedJvm.scala
```diff
@@ -18,40 +18,15 @@ package org.http4s
...
-  }
-
-  lazy val testResourceGzipped: Chunk[Byte] = {
-    val url = getClass.getResource("/testresource.txt.gz")
-    require(url != null, "Couldn't acquire resource!")
-    val bytes = Files.readAllBytes(Paths.get(url.toURI))
-
-    Chunk.array(bytes)
-  }
+private[staticcontent] trait StaticContentSharedJvm extends StaticContentShared {
```
armanbilge, line 28: Was splitting this file necessary to avoid linking errors? It's not a huge deal, but as long as we are still merging from uncrossed series/0.22 into series/0.23 I've been trying to avoid this.

server/shared/src/main/scala/org/http4s/server/staticcontent/FileService.scala
```diff
@@ -33,39 +33,206 @@ import org.log4s.getLogger
...
+  type Fs2PathCollector[F[_]] = (Path, Config[F], Request[F]) => OptionT[F, Response[F]]
+
   /** [[org.http4s.server.staticcontent.FileService]] configuration
     *
     * @param systemPath path prefix to the folder from which content will be served
     * @param pathPrefix prefix of Uri from which content will be served
-    * @param pathCollector function that performs the work of collecting the file or rendering the directory into a response.
+    * @param fs2PathCollector function that performs the work of collecting the file or rendering the directory into a response.
     * @param bufferSize buffer size to use for internal read buffers
     * @param cacheStrategy strategy to use for caching purposes. Default to no caching.
```
armanbilge, line 53:
```suggestion
    * @param fs2PathCollector function that performs the work of collecting the file or rendering the directory into a response.
    * @param pathPrefix prefix of Uri from which content will be served
    * @param cacheStrategy strategy to use for caching purposes. Default to no caching.
    * @param bufferSize buffer size to use for internal read buffers
```

server/shared/src/main/scala/org/http4s/server/staticcontent/FileService.scala
```diff
@@ -33,39 +33,206 @@ import org.log4s.getLogger
...
+        )
+      )
+
+    /** Creates an instance of [[org.http4s.server.staticcontent.FileService]] configuration.
+      *
+      * @param systemPath path prefix to the folder from which content will be served
+      * @param pathPrefix prefix of Uri from which content will be served
+      * @param fs2PathCollector function that performs the work of collecting the file or rendering the directory into a response.
+      * @param bufferSize buffer size to use for internal read buffers
+      * @param cacheStrategy strategy to use for caching purposes. Default to no caching.
```
armanbilge, line 190:
```suggestion
      * @param systemPath path prefix to the folder from which content will be served
      * @param fs2PathCollector function that performs the work of collecting the file or rendering the directory into a response.
      * @param pathPrefix prefix of Uri from which content will be served
      * @param cacheStrategy strategy to use for caching purposes. Default to no caching.
      * @param bufferSize buffer size to use for internal read buffers
```

server/shared/src/main/scala/org/http4s/server/staticcontent/FileService.scala
```diff
@@ -33,39 +33,206 @@ import org.log4s.getLogger
...
+      new Config[F](systemPath, fs2PathCollector, pathPrefix, cacheStrategy, bufferSize)
+
+    /** Creates an instance of [[org.http4s.server.staticcontent.FileService]] configuration.
+      * A constructor that accepts PathCollector[F] for binary compatibility.
+      *
+      * @param systemPath path prefix to the folder from which content will be served
+      * @param pathPrefix prefix of Uri from which content will be served
+      * @param pathCollector function that performs the work of collecting the file or rendering the directory into a response.
+      * @param bufferSize buffer size to use for internal read buffers
+      * @param cacheStrategy strategy to use for caching purposes. Default to no caching.
```
armanbilge, line 208:
```suggestion
      * @param systemPath path prefix to the folder from which content will be served
      * @param pathCollector function that performs the work of collecting the file or rendering the directory into a response.
      * @param pathPrefix prefix of Uri from which content will be served
      * @param bufferSize buffer size to use for internal read buffers
      * @param cacheStrategy strategy to use for caching purposes. Default to no caching.
```

armanbilge (commented):
Sorry, just a couple more things I noticed!

server/jvm/src/test/scala/org/http4s/server/staticcontent/StaticContentSharedJvm.scala
yurique: Lemme take a look. I think I should be able to put it all back together.

server/shared/src/main/scala/org/http4s/server/staticcontent/FileService.scala
yurique: github won't allow me to apply this, will commit manually

server/jvm/src/test/scala/org/http4s/server/staticcontent/StaticContentSharedJvm.scala
yurique: So it worked. I moved the jvm related stuff into `shared`, but as it's not referenced from the tests in `shared` – linker didn't yell.

server/shared/src/main/scala/org/http4s/server/staticcontent/FileService.scala
armanbilge: weird

server/shared/src/test/scala/org/http4s/server/staticcontent/StaticContentShared.scala
```diff
@@ -20,38 +20,27 @@ package staticcontent
...
-          .getBytes(StandardCharsets.UTF_8)
-      )
+  protected val defaultSystemPath: String =
+    org.http4s.server.test.BuildInfo.test_resourceDirectory
+      .replace("jvm", "shared")
+      .replace("js", "shared")
 
-    Chunk.array(bytes)
-  }
+  lazy val testResource: IO[Chunk[Byte]] =
```
yurique, line 39: tried to revert these to just `Chunk[Byte]`, but there's no `unsafeRunSync` in js, right?

armanbilge: Yeah, no `unsafeRunSync`. That's fine, thanks for trying to minimize the changes :)

armanbilge (approved):
🎉 Thanks for shaving this yak and all the upstream fixes! Now we can implement http4s equivalent of `python3 -m http.server` and publish it to npm 😁

We should get another pair of eyes on this PR, since it also affects JVM side. Summary of changes for next reviewer(s):
- Deprecates `java.io.File`-based `PathCollector` for `fs2.io.file.Path`-based `Fs2PathCollector`
- Reworks `FileService.Config` to use `Fs2PathCollector`. It is no longer a `case class`, and manually implements the `case class` methods for bincompat. Credit to @yurique for being very thorough about this.
- Refactors the tests to use `fs2.io.file` APIs

All of this fully preserves bin-compat (no MiMa filters added) and source-compat.

project/Http4sPlugin.scala
```diff
@@ -208,7 +208,7 @@ object Http4sPlugin extends AutoPlugin {
     val cryptobits = "1.3"
     val disciplineCore = "1.4.0"
     val dropwizardMetrics = "4.2.7"
-    val fs2 = "3.2.3"
+    val fs2 = "3.2-32-1a7dede"
```
armanbilge, line 211:
```suggestion
    val fs2 = "3.2.4"
```

server/shared/src/main/scala/org/http4s/server/staticcontent/FileService.scala
```diff
@@ -33,39 +33,206 @@ import org.log4s.getLogger
...
+    )
+    override def productArity: Int = 5
+
+    @deprecated(
+      "Config is no longer a case class. The productElement method is provided for binary compatibility.",
+      "0.23.8",
+    )
+    override def productElement(n: Int): Any = n match {
+      case 0 => systemPath
+      case 1 => pathCollector
```
yurique, line 134: This can potentially break linking in JS.

* [ ] figure it out

yurique: Or maybe it doesn't break linking because the reference to `File` is erased?

armanbilge: That's a good point about erasure. The only way to know for sure is if we actually try it I think, because of the munit issue.

I'm not sure erasure will save us here: it seems difficult for Scala.js to prove that this `File`-based path-collector is never used, so it has to include code in case it is used. In which case it would need to link to `File`. But, I could be wrong, it's very difficult to reason about these things.

I wonder if I can convince munit to backport my fix into 0.7 series, I will ask.

yurique: 🤞

rossabaker:
Goodness, that's a clever way to work around the case class bincompat.  We can and should clean this up in main.  There is also an idea, and if memory serves, a WIP, to turn these objects into a tagless algebra.

That said, this is great for this branch and this point in time.

PR #5760, opened 2021-12-19
Convert Entity to an ADT with an empty case

bplommer:
This is the second slice of #4509.

### DIFF
--- a/core/shared/src/main/scala/org/http4s/Entity.scala
+++ b/core/shared/src/main/scala/org/http4s/Entity.scala
@@ -19,6 +19,8 @@ package org.http4s
 import cats.Monoid
 import cats.syntax.all._
 import cats.~>
+import fs2.Pure
+import fs2.Stream
 
 sealed trait Entity[+F[_]] {
   def body: EntityBody[F]
@@ -31,12 +33,26 @@ object Entity {
   def apply[F[_]](body: EntityBody[F], length: Option[Long] = None): Entity[F] =
     Default(body, length)
 
+  // The type parameter aids type inference in Message constructors. This should be unnecessary when Messages are covariant.
+  def empty[F[_]]: Entity[F] = Empty
+
   final case class Default[+F[_]](body: EntityBody[F], length: Option[Long]) extends Entity[F] {
     def ++[F1[x] >: F[x]](that: Entity[F1]): Entity[F1] = that match {
       case d: Default[F1] => Default(body ++ d.body, (length, d.length).mapN(_ + _))
+      case Empty => this
     }
 
     def translate[F1[x] >: F[x], G[_]](fk: F1 ~> G): Entity[G] = Default(body.translate(fk), length)
+  }
+
+  case object Empty extends Entity[Pure] {
+    val body: EntityBody[Pure] = Stream.empty
+
+    val length: Option[Long] = Some(0L)
+
+    def ++[F1[x] >: Pure[x]](that: Entity[F1]): Entity[F1] = that
+
+    def translate[F1[x] >: Pure[x], G[_]](fk: F1 ~> G): Entity[G] = this
   }
 
   implicit def http4sMonoidForEntity[F[_]]: Monoid[Entity[F]] =
@@ -44,6 +60,4 @@ object Entity {
       def combine(a1: Entity[F], a2: Entity[F]): Entity[F] = a1 ++ a2
       val empty: Entity[F] = Entity.empty
     }
-
-  val empty: Entity[Nothing] = Entity[Nothing](EmptyBody, Some(0L))
 }
--- a/core/shared/src/main/scala/org/http4s/Entity.scala
+++ b/core/shared/src/main/scala/org/http4s/Entity.scala
@@ -20,18 +20,29 @@ import cats.Monoid
 import cats.syntax.all._
 import cats.~>
 
-final case class Entity[+F[_]](body: EntityBody[F], length: Option[Long] = None) {
-  def translate[F1[x] >: F[x], G[_]](fk: F1 ~> G): Entity[G] = Entity(body.translate(fk), length)
+sealed trait Entity[+F[_]] {
+  def body: EntityBody[F]
+  def length: Option[Long]
+  def ++[F1[x] >: F[x]](that: Entity[F1]): Entity[F1]
+  def translate[F1[x] >: F[x], G[_]](fk: F1 ~> G): Entity[G]
 }
 
 object Entity {
+  def apply[F[_]](body: EntityBody[F], length: Option[Long] = None): Entity[F] =
+    Default(body, length)
+
+  final case class Default[+F[_]](body: EntityBody[F], length: Option[Long]) extends Entity[F] {
+    def ++[F1[x] >: F[x]](that: Entity[F1]): Entity[F1] = that match {
+      case d: Default[F1] => Default(body ++ d.body, (length, d.length).mapN(_ + _))
+    }
+
+    def translate[F1[x] >: F[x], G[_]](fk: F1 ~> G): Entity[G] = Default(body.translate(fk), length)
+  }
 
   implicit def http4sMonoidForEntity[F[_]]: Monoid[Entity[F]] =
     new Monoid[Entity[F]] {
-      def combine(a1: Entity[F], a2: Entity[F]): Entity[F] =
-        Entity(a1.body ++ a2.body, (a1.length, a2.length).mapN(_ + _))
-      val empty: Entity[F] =
-        Entity.empty
+      def combine(a1: Entity[F], a2: Entity[F]): Entity[F] = a1 ++ a2
+      val empty: Entity[F] = Entity.empty
     }
 
   val empty: Entity[Nothing] = Entity[Nothing](EmptyBody, Some(0L))


core/shared/src/main/scala/org/http4s/Entity.scala
```diff
@@ -19,20 +19,45 @@ package org.http4s
...
 }
 
 object Entity {
+  def apply[F[_]](body: EntityBody[F], length: Option[Long] = None): Entity[F] =
+    Default(body, length)
+
+  // The type parameter aids type inference in Message constructors. This should be unnecessary when Messages are covariant.
+  def empty[F[_]]: Entity[F] = Empty
+
+  final case class Default[+F[_]](body: EntityBody[F], length: Option[Long]) extends Entity[F] {
```
bplommer, line 39: Default is a placeholder for cases that haven't yet been specialized.

core/shared/src/main/scala/org/http4s/Entity.scala
```diff
@@ -19,20 +19,45 @@ package org.http4s
...
+    def ++[F1[x] >: F[x]](that: Entity[F1]): Entity[F1] = that match {
+      case d: Default[F1] => Default(body ++ d.body, (length, d.length).mapN(_ + _))
```
bplommer, line 41: The variance means Scala 2 can't handle this pattern match if you use extractors 😕

core/shared/src/main/scala/org/http4s/Entity.scala
```diff
@@ -19,20 +19,45 @@ package org.http4s
 import cats.Monoid
 import cats.syntax.all._
 import cats.~>
+import fs2.Pure
+import fs2.Stream
 
-final case class Entity[+F[_]](body: EntityBody[F], length: Option[Long] = None) {
-  def translate[F1[x] >: F[x], G[_]](fk: F1 ~> G): Entity[G] = Entity(body.translate(fk), length)
+sealed trait Entity[+F[_]] {
```
armanbilge, line 25: So, IDK when's the right time to start discussing this one (feel free to defer :) but pros/cons about `seal`ing `Entity`?

bplommer: I was vaguely thinking we should seal `Entity` but then give it an unsealed subtrait called something like `Other`, so people can extend that but we still get exhaustive matches on `Entity`. But I think we can defer this decision for now.

core/shared/src/main/scala/org/http4s/Entity.scala
armanbilge: I'll add _drop Scala 2.13_ to my PR 😉

I see similar code in fs2 `Pull` interpreter, but this sort of thing shouldn't be too user-facing I think 🤔

core/shared/src/main/scala/org/http4s/Entity.scala
```diff
@@ -19,20 +19,45 @@ package org.http4s
...
+  final case class Default[+F[_]](body: EntityBody[F], length: Option[Long]) extends Entity[F] {
+    def ++[F1[x] >: F[x]](that: Entity[F1]): Entity[F1] = that match {
+      case d: Default[F1] => Default(body ++ d.body, (length, d.length).mapN(_ + _))
+      case Empty => this
+    }
+
+    def translate[F1[x] >: F[x], G[_]](fk: F1 ~> G): Entity[G] = Default(body.translate(fk), length)
+  }
+
+  case object Empty extends Entity[Pure] {
```
armanbilge, line 48: Something quite nice about this :)

armanbilge (approved):
Thanks for keeping these coming!

core/shared/src/main/scala/org/http4s/Entity.scala
```diff
@@ -19,20 +19,45 @@ package org.http4s
...
 import cats.syntax.all._
 import cats.~>
+import fs2.Pure
+import fs2.Stream
 
-final case class Entity[+F[_]](body: EntityBody[F], length: Option[Long] = None) {
-  def translate[F1[x] >: F[x], G[_]](fk: F1 ~> G): Entity[G] = Entity(body.translate(fk), length)
+sealed trait Entity[+F[_]] {
+  def body: EntityBody[F]
+  def length: Option[Long]
```
danicheg, line 27: As I pointed in #5754 - would this better to have the `length` as an effectful function?  I'm confused a bit with a case when `Response` will have a non-empty `Stream` of data and empty `length`.

danicheg (commented):
It's nice to see your work on this! Few suggestions/questions:

core/shared/src/main/scala/org/http4s/Entity.scala
armanbilge: An empty length actually means unknown length. I.e., the only way to compute the length is by evaluating the entire `Stream`, which you might not want to do. Maybe we just need a scaladoc to explain this? :)

core/shared/src/main/scala/org/http4s/Entity.scala
armanbilge: We could do a type alias I suppose.

bplommer: The problem is that `Pure` is defined by an inequality, `Pure[A] <: Nothing`. So if we define our own one it won't be the same as fs2's one, unless it's just an alias for `fs2.Pure`.

core/shared/src/main/scala/org/http4s/Entity.scala
bplommer: Maybe we should rename `length` to `knownLength`? It's a bit awkward though

bplommer:
@danicheg are these issues blockers to merging this PR or are  you happy for us to keep iterating on this?

core/shared/src/main/scala/org/http4s/Entity.scala
danicheg: @armanbilge are we need that `length` then? I think it could be used in some surprising ways on the user site and can bring nasty bugs for those who come to `http4s` with similar to `akka-http` experience (where data is a fixed array of bytes).  I'm wondering about something like that ADT:
```
Entity = Empty | EntityWithLength | DefaulEntity
```
What do you folks think?

danicheg:
@benplommer I'd like it if we can debate a little about my concerns. But you surely can go forth on iterating.

core/shared/src/main/scala/org/http4s/Entity.scala
armanbilge: > `akka-http` experience (where data is a fixed array of bytes)

Can you explain more about this? I'm looking at akka-http scaladocs for `HttpEntity`:

https://doc.akka.io/api/akka-http/10.2.7/akka/http/scaladsl/model/HttpEntity.html

All I see is a `Stream` of bytes:
```scala
 def dataBytes: Source[ByteString, Any]
```

They also have a method similar to the one that concerns you:
```scala
/** Some(content length) if a length is defined for this entity, None otherwise.
 * A length is only defined for Strict and Default entity types.
 */
abstract def contentLengthOption: Option[Long]
```

So, I think we can provide a similar experience. What do you think?

danicheg: I mean that it's quite common in the `akka-http` land to use a strict version of the `HttpEntity` - `HttpEntity.Strict`. `contentLengthOption` is a not good design choice, I think. Even `akka-http` itself contains warnings about using it:
```
In many cases it's dangerous to rely on the (non-)existence of a content-length.
HTTP intermediaries like (transparent) proxies are allowed to change the transfer-encoding which can result in the entity being delivered as another type as expected.
```
So what I stand for - it's a safer API for end-users. Since we are trying to design a model of `HttpEntity` from scratch, we can offer to users something more robust than scaladoc with warnings, I opine.

armanbilge: Ah, thanks for explaining :)

First of all, 💯 👍 to some sort of `StrictEntity` in our `Entity` ADT. That is definitely one of the goals of these changes, that I think (hope) everyone wants 😆 

Regarding `length: Option[Long]` you make an important point there. It seems there is an important distinction whether it is derived from the `Content-Length` header or from the actual known size of a strict entity in memory. The latter should be safe, and maybe in any other case it should be `None`. Curious to hear others' thoughts on this.

bplommer: Yep, `Strict` is coming up next! That will wrap a `Chunk[byte]` and have length defined as the size of the chunk.

Streaming bodies are the harder case I think - I haven't yet properly digested how Chris's PR handles them, but there's a `Chunked` case that always has length None and a `TrustMe` case where length is set by calling code - maybe we can more discuss these issues more usefully and concretely when we get to those slices?

I still don't see how an effectful `length` field would work with an immutable data model though - computing the size would require running the body stream, leaving it in a bad state for actually running the body unless the body is memoized. (What we *could* do is to add a `prefetch` method that effectfully converts streaming entities to strict ones.)

bplommer:
> @bplommer I'd like it if we can debate a little about my concerns. But you surely can go forth on iterating.

I'll wait for your approval before merging this, don't want to force it through 🙂

core/shared/src/main/scala/org/http4s/Entity.scala
bplommer: It's hard-coded to `Some(0L)` in the empty case.

core/shared/src/main/scala/org/http4s/Entity.scala
danicheg: Sorry, I made the same error twice. I thought this design of `HttpEntity` is the final. Probably I should dig into Chris PR :) 
And yeah, if other slices are coming, we can discuss those things there.

core/shared/src/main/scala/org/http4s/Entity.scala
bplommer: Never mind, I misread what line of code you were commenting on. I guess this is superseded by the discussion below.

danicheg (approved):
👍🏻 from me to see the next slices. I'll come with my concerns about the `length` in the `HttpEntity` model though 😄 And yeah, it'd be great if other core members of the team would dig into that discussion too.

Issue #5740, opened 2021-12-14
Redesign tests based on `QueueTestHead` and `onFinalizeWeak`


RafalSumislawski:
`Http1ClientStageSuite` and `ClientTimeoutSuite`.`"Idle timeout on slow POST body"` use a combo of `QueueTestHead` to mock the network side of the connection and `request.body.onFinalizeWeak` to trigger some action in `QueueTestHead`.  In these tests `onFinalizeWeak` is meant to detect when the request has been sent. The problem is that finalisation of the request body stream, does not guarantee the request has been sent. This can be clearly seen in the code of `org.http4s.blazecore.util.EntityBodyWriter#writeEntityBody` where `writeEnd` is executed after `drain`:

```
  def writeEntityBody(p: EntityBody[F]): F[Boolean] = {
    val writeBody: F[Unit] = p.through(writePipe).compile.drain
    val writeBodyEnd: F[Boolean] = fromFutureNoShift(F.delay(writeEnd(Chunk.empty)))
    writeBody *> writeBodyEnd
  }
```

This issue leads to flaky tests like this one: 
https://github.com/http4s/http4s/runs/4498891671?check_suite_focus=true#step:13:5129

RafalSumislawski:
```
 ==> X org.http4s.blaze.client.Http1ClientStageSuite.Submit a request line with a query  0.017s munit.ComparisonFailException: /home/runner/work/http4s/http4s/blaze-client/src/test/scala/org/http4s/blaze/client/Http1ClientStageSuite.scala:153 Obtained empty output!
152:      val statusLine = request.split("\r\n").apply(0)
153:      assertEquals(statusLine, "GET " + uri + " HTTP/1.1")
154:      assertEquals(response, "done")
    at munit.Assertions.failComparison(Assertions.scala:297)
    at munit.Assertions.failComparison$(Assertions.scala:287)
    at munit.FunSuite.failComparison(FunSuite.scala:11)
    at munit.Assertions$$anon$1.handle(Assertions.scala:31)
    at munit.internal.difflib.Diffs$.assertNoDiff(Diffs.scala:42)
    at munit.Assertions.$anonfun$assertEquals$1(Assertions.scala:141)
    at get @ fs2.Stream$.$anonfun$interruptWhen$3(Stream.scala:1566)
    at map @ org.http4s.blaze.client.Http1ClientStageSuite.$anonfun$getSubmission$14(Http1ClientStageSuite.scala:124)
[...]

PR #5761, opened 2021-12-19
Improve ClientTimeoutSuite

RafalSumislawski:
This PR solves the following problems:
*  The use of `QueueTestHead` with `onFinalizeWeak` which is by it's nature racy and leads to flaky tests. (this half solves #5740)
* Incorrect construction of connections, leading to to the `IdleTimoutStage` not being included in the pipelinie (it used to be added in `mkConnection` but then removed in `MockClientBuilder`)
* All tests of idle-timeout passing despite the issue mentioned above. (This is solved by redefining the "Not timeout on only marginally slow request body" in such a way that it would fail if the `IdleTimeoutStage` wasn't included in the pipeline)

Additionally it:
* improves test coverage by dividing "Idle timeout on slow POST body" into two test cases, one of which is currently failing due to a flaw in blaze client implementation (I think #5385 will solve it).
* reduces amount of code


blaze-client/src/test/scala/org/http4s/blaze/client/ClientTimeoutSuite.scala
```diff
@@ -108,100 +95,110 @@ class ClientTimeoutSuite extends Http4sSuite {
...
+        c = mkClient(h, tickWheel)(idleTimeout = 1.second)
+        s <- c.fetchAs[String](req)
+      } yield s).intercept[TimeoutException]
+  }
 
-    c.fetchAs[String](req).assertEquals("done")
+  tickWheelFixture.test("Not timeout on only marginally slow request body".flaky) { tickWheel =>
+    // Sending request body will take 1500ms. But there will be some activity every 500ms.
+    // If the idle timeout wasn't reset every time something is sent, it would kick-in after 1 second.
+    // The chunks need to be larger than the buffer in CachingChunkWriter
```
danicheg, line 160: Could we make this more explicit in the test suite by using `chunkBufferMaxSize + N` instead of magic `2000000`?

danicheg (approved):
Thank you for adding comments for those test suites. So much appreciated.

blaze-client/src/test/scala/org/http4s/blaze/client/ClientTimeoutSuite.scala
RafalSumislawski: Done.

PR #5767, opened 2021-12-20
Collection micro-opts

danicheg:
This brings some micro-optimizations of using collections, such as:
* using the `contains` instead of `exists(_ == foo)`;
* using the `getOrElse(foo, bar)` instead of `get().getOrElse()`;
* using the `find` instead of `filter().headOption()`;
* removing the redundant collections conversions;
* and little others.

### DIFF
--- a/blaze-core/src/test/scala/org/http4s/blazecore/util/DumpingWriter.scala
+++ b/blaze-core/src/test/scala/org/http4s/blazecore/util/DumpingWriter.scala
@@ -41,7 +41,7 @@ class DumpingWriter(implicit protected val F: Effect[IO]) extends EntityBodyWrit
 
   def toArray: Array[Byte] =
     buffer.synchronized {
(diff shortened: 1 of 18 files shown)


armanbilge (approved):
Nice. Trust you + ci 👍

blaze-core/src/test/scala/org/http4s/blazecore/util/DumpingWriter.scala
```diff
@@ -41,7 +41,7 @@ class DumpingWriter(implicit protected val F: Effect[IO]) extends EntityBodyWrit
 
   def toArray: Array[Byte] =
     buffer.synchronized {
-      Chunk.concatBytes(buffer.toSeq).toArray
+      Chunk.concatBytes(buffer).toArray
```
bplommer, line 44: Is going via `Chunk` the most optimal way to do this?

bplommer (approved):

PR #5777, opened 2021-12-22
Rm `sbt-updates` plugin

danicheg:
It seems the plugin is redundant because we have Steward.


armanbilge (approved):
I wonder if some people might still use this to check for updates locally. I don't.

bplommer:
I have this in my global plugins.sbt, which is where it belongs if not part of the build IMO.

bplommer (approved):

PR #5786, opened 2021-12-25
Make the output of `asCurl` more human-readable

danicheg:
This makes the output of `asCurl` verbose and adds line breakers. The result of method calls will look like this:
```
curl \
  --request GET \
  --url 'http://localhost:1234/foo' \
  --header 'Cookie: <REDACTED>' \
  --header 'Authorization: <REDACTED>'
```

### DIFF
--- a/core/src/main/scala/org/http4s/Message.scala
+++ b/core/src/main/scala/org/http4s/Message.scala
@@ -32,6 +32,7 @@ import fs2.Pure
 import fs2.Stream
 import fs2.text.utf8Encode
 import org.http4s.headers._
+import org.http4s.internal.CurlConverter
 import org.http4s.syntax.KleisliSyntax
 import org.http4s.syntax.KleisliSyntaxBinCompat0
 import org.http4s.syntax.KleisliSyntaxBinCompat1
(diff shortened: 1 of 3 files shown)


bplommer:
I'm not sure how I feel about the inclusion of newlines. How about adding a Boolean flag (with a default value) to enable them?

danicheg:
@bplommer if we look at browsers they copy cURL with newlines (tested with Google Chrome). And from a copy-and-paste perspective (one of the most popular usage case, I guess) output with newlines is more convenient to utilize. Are we actually want an old one-line-text behavior?

armanbilge:
I definitely can't speak for other users, but last time I used this feature was to copy-paste into [curly4s](https://davenverse.github.io/curly4s/) which doesn't support newlines (although that should be fixed 😆). So I'm the guy in that comic Rafał posted: https://xkcd.com/1172/

bplommer:
I've never used `asCurl` at all so I'll leave you two to hammer this out 😅

danicheg:
@armanbilge at least in unit tests newlines works https://github.com/davenverse/curly4s/commit/870aebecc5ec11e2c48deeed936db3913c1eb758

armanbilge:
I mean specifically the web interface :) anyway, it's no big deal, just my personal thing 😆

danicheg:
@armanbilge @bplommer I would advise to use `.replace(" \\\n ", "")` for the output to get previous behaviour. But one can't so easily get the output with newlines from one-liner. Spead it through the boolean flag in the contract seems too extra for me. So I choose the option 'take it or leave it'.

bplommer:
@danicheg That sounds fine by me. The other question is whether we consider this a breaking change. I mildly lean towards targeting this to main rather than 0.22, but I don't feel at all strongly about it.

danicheg:
As one of those who are fighting against changing behavior without breaking APIs in the same minor version, I feel little doubts about having this in the 0.22 series. I can't really imagine in any way that can break somebody's workflow. This function is definitely used for debugging (or something like that) purposes. And it still will produce a valid cURL if someone is used it in some integrations.

bplommer:
Giving this my approval but I'd like to wait for @armanbilge's approval before it's merged.

armanbilge (approved):
I think this is a reasonable change for 0.22. As @bplommer suggested a `prettyPrint` flag or `asPrettyCurl` could be nice and helpful for compatibility, but as you point out perhaps that is excessive for a simple debug utiility.

It's unfortunate that it will break my "workflow" (waves hands), but as there are no open issues about this I assume it is breaking your workflow, and you did good work here cleaning up the implementation, so overall I'd say it's a net win :)

danicheg:
Yay! I've created davenverse/curly4s#17 for you @armanbilge :)

PR #5789, opened 2021-12-26
Refactor some `Http1Connection` methods

danicheg:
This refactors some methods of the `Http1Connection` to simplify reasoning. Implementation, by and large, wasn't changed. Just decomposition was made more fine-grained and nested ifs replaced by pattern-matching.

### DIFF
--- a/blaze-client/src/main/scala/org/http4s/blaze/client/Http1Connection.scala
+++ b/blaze-client/src/main/scala/org/http4s/blaze/client/Http1Connection.scala
@@ -430,28 +430,39 @@ private final class Http1Connection[F[_]](
   @tailrec private def validateRequest(req: Request[F]): Either[Exception, Request[F]] = {
     val minor: Int = getHttpMinor(req)
 
-    // If we are HTTP/1.0, make sure HTTP/1.0 has no body or a Content-Length header
-    if (minor == 0 && req.headers.get[`Content-Length`].isEmpty) {
-      logger.warn(s"Request $req is HTTP/1.0 but lacks a length header. Transforming to HTTP/1.1")
-      validateRequest(req.withHttpVersion(HttpVersion.`HTTP/1.1`))
-    }
-    // Ensure we have a host header for HTTP/1.1
-    else if (minor == 1 && req.uri.host.isEmpty) // this is unlikely if not impossible
-      if (req.headers.get[Host].isDefined) {
-        val host = req.headers.get[Host].get // TODO gross
-        val newAuth = req.uri.authority match {
-          case Some(auth) => auth.copy(host = RegName(host.host), port = host.port)
-          case None => Authority(host = RegName(host.host), port = host.port)
+    minor match {
+      // If we are HTTP/1.0, make sure HTTP/1.0 has no body or a Content-Length header
+      case 0 if req.headers.get[`Content-Length`].isEmpty =>
+        logger.warn(s"Request $req is HTTP/1.0 but lacks a length header. Transforming to HTTP/1.1")
+        validateRequest(req.withHttpVersion(HttpVersion.`HTTP/1.1`))
+
+      case 1 if req.uri.host.isEmpty => // this is unlikely if not impossible
+        // Ensure we have a host header for HTTP/1.1
+        req.headers.get[Host] match {
+          case Some(host) =>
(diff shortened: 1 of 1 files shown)


blaze-client/src/main/scala/org/http4s/blaze/client/Http1Connection.scala
```diff
@@ -310,105 +310,113 @@ private final class Http1Connection[F[_]](
         cb(Left(t))
     }(executionContext)
 
-  private def parsePrelude(
+  // it's called when the body is known
+  private def terminationConditionParsingPrelude(): Either[Throwable, Option[Chunk[Byte]]] =
```
RafalSumislawski, line 314: I would like to search for a better name for this method. It actually is relevant to parsing the body not the prelude. It's called when an EOF is received while parsing body, and the purpose of this metod it to decide whether it's ok that we recieved the EOF or not.

Maybe something like `onEofWhileReadingBody`?

blaze-client/src/main/scala/org/http4s/blaze/client/Http1Connection.scala
```diff
@@ -310,105 +310,113 @@ private final class Http1Connection[F[_]](
...
+    stageState.get match { // if we don't have a length, EOF signals the end of the body.
+      case Error(e) if e != EOF => Either.left(e)
+      case _ =>
+        if (parser.definedContentLength() || parser.isChunked())
+          Either.left(InvalidBodyException("Received premature EOF."))
+        else Either.right(None)
+    }
+
+  private def cleanupParsingPrelude(closeOnFinish: Boolean, headers: Headers): Unit =
```
RafalSumislawski, line 323: Maybe `cleanupParsingResponse`? Similarly to my comment on `terminationConditionParsingPrelude`. The `cleanupParsingPrelude` happens later than after parsing prelude. First we parse the prelude, then the body, then (optionally) the trailers and then we clean up.

IMO it should be `cleanUp` (a verb) not `cleanup` (a noun).

blaze-client/src/main/scala/org/http4s/blaze/client/Http1Connection.scala
```diff
@@ -310,105 +310,113 @@ private final class Http1Connection[F[_]](
...
+          headers = headers,
+          body = body.interruptWhen(idleTimeoutS),
+          attributes = attributes,
         )
       )
-    } catch {
+    )
+  }
+
+  private def parsePrelude(
```
RafalSumislawski, line 406: IMO this class would be easier if reading it top to bottom would be consistent with the order in which these operations happen (methods are called). The already existing methods are ordered like this. At least the most important ones.

I would like to see them in this order:
- runRequest
- executeRequest
- receiveResponse
- handleRead
- parsePrelude
- parsePreludeFinished (new)
- terminationConditionParsingPrelude (new)
- cleanupParsingPrelude (new)

RafalSumislawski (changes requested):

danicheg:
@RafalSumislawski wow, what an utter review! :) If I get it right, you argue on most about naming, formatting, and order of methods. I give or take ok with your concerns, so feel free to proceed with my work on this if you want it. I'm not so interested in pushing this further.

blaze-client/src/main/scala/org/http4s/blaze/client/Http1Connection.scala
```diff
@@ -310,105 +310,113 @@ private final class Http1Connection[F[_]](
...
+      buffer: ByteBuffer,
+      closeOnFinish: Boolean,
+      doesntHaveBody: Boolean,
+      cb: Callback[Response[F]],
+      idleTimeoutS: F[Either[Throwable, Unit]],
+  ): Unit =
+    try if (!parser.finishedResponseLine(buffer))
```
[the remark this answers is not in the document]
danicheg: Probably this should be done by Scalafmt? Though it's not looking so bizarre for me.

RafalSumislawski (approved):

RafalSumislawski:
Hi @danicheg 
I applied the changes that I suggested earlier in the review. Please check if you're ok with the changes before we merge.

danicheg:
@RafalSumislawski thanks. LGTM.

PR #5804, opened 2022-01-02
Add Strict entity

bplommer:
I've merged `Empty` into `Strict` to avoid having two legal representations for a strict empty entity - any thoughts on this?

### DIFF
--- a/core/shared/src/main/scala/org/http4s/Entity.scala
+++ b/core/shared/src/main/scala/org/http4s/Entity.scala
@@ -58,7 +58,7 @@ object Entity {
 
     def ++[F1[x] >: Pure[x]](that: Entity[F1]): Entity[F1] = that match {
       case d: Default[F1] => Default(body ++ d.body, d.length.map(chunk.size + _))
-      case Strict(chunk2) => if (chunk2.isEmpty) this else Strict(chunk ++ chunk2)
+      case Strict(chunk2) => Strict(chunk ++ chunk2)
     }
 
     def translate[F1[x] >: Pure[x], G[_]](fk: F1 ~> G): Entity[G] = this
--- a/core/shared/src/main/scala/org/http4s/Entity.scala
+++ b/core/shared/src/main/scala/org/http4s/Entity.scala
@@ -56,14 +56,10 @@ object Entity {
 
     val length: Option[Long] = Some(chunk.size.toLong)
 
-    def ++[F1[_]](that: Entity[F1]): Entity[F1] =
-      if (chunk.isEmpty) that
-      else {
-        that match {
-          case d: Default[F1] => Default(body ++ d.body, d.length.map(chunk.size + _))
-          case Strict(chunk2) => if (chunk2.isEmpty) this else Strict(chunk ++ chunk2)
-        }
-      }
+    def ++[F1[x] >: Pure[x]](that: Entity[F1]): Entity[F1] = that match {
+      case d: Default[F1] => Default(body ++ d.body, d.length.map(chunk.size + _))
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/org/http4s/Entity.scala
```diff
@@ -34,23 +35,35 @@ object Entity {
...
+        if (chunk.isEmpty) this
+        else Default(body ++ Stream.chunk(chunk), length.map(_ + chunk.size))
     }
 
     def translate[F1[x] >: F[x], G[_]](fk: F1 ~> G): Entity[G] = Default(body.translate(fk), length)
   }
 
-  case object Empty extends Entity[Pure] {
-    val body: EntityBody[Pure] = Stream.empty
+  final case class Strict(chunk: Chunk[Byte]) extends Entity[Pure] {
```
armanbilge, line 54: Bikeshedding time :) I'm of the opinion we should use a scodec `ByteVector` here rather than a `Chunk`. Is there a reason you felt drawn to `Chunk`?

armanbilge:
So, regarding the `Empty` vs `Strict` thing, I do agree it would be nice to avoid having two legal representations for this. Also it opens the question whether a distinct `Empty` entity should extend `Strict`. I wonder if we can reach for a representation like `sealed trait Strict` inhabited by `EmptyStrict` and `NonEmptyStrict`, where the non-empty one behaves like any non-empty collection in cats. That way we can have real static guarantees about this stuff, which might be useful (?) Like maybe too crazy, but I'm reminded of your PRs using singleton types to be ultra expressive, such statically requiring a 204 response to be paired with `EmptyStrict` type, and that entity only.

As always, thanks for keeping this series of PRs going!! :)

core/shared/src/main/scala/org/http4s/Entity.scala
bplommer: Nope, just copying what @ChristopherDavenport did in the original PR. I'm totally open to `ByteVector` and see the appeal of not using an fs2 type here.

armanbilge: Cool. My vote is for `ByteVector` then, curious to hear other's thoughts on this as well. My thinking is along the lines of:
1. `Chunk` is almost an fs2 internal thing here, as you point out
2. `ByteVector` is a lot more optimized for `Byte` than `Chunk[Byte]`
3. Precedent/defer to authority in https://github.com/http4s/http4s/pull/5321#discussion_r721641372

bplommer:
@armanbilge I did try something like that but then I started bikeshedding with myself so decided to just open the simplest possible PR. Maybe I'll give it another go though 😄

bplommer:
Following discussion on Discord, I'm keeping `Empty` as a separate case and leaving the representation as `Chunk` for now.

core/shared/src/main/scala/org/http4s/Entity.scala
armanbilge: Ok, the consensus on Discord is for `Chunk` primarily b/c `Chunk` can subsume `ByteVector` (but not the other way around), so this would let us avoid unnecessary copying.

Personally I still like `ByteVector` for the reasons above, but `Chunk[Byte]` is good too.

armanbilge (approved):

armanbilge:
Hum, I thought we'd wait a little more before merging. I remember @danicheg might have been deferring some discussions for this PR.

bplommer:
Good point, sorry for being so trigger-happy 😬 I'll revert it and open a new one.

PR #5805, opened 2022-01-02
Simplify logger type signatures

bplommer:


danicheg (approved):
That extra type is needed just for marking the fact that `Message` is a `Request` or `Response`, I suppose. Anyway, LGTM.

PR #5806, opened 2022-01-02
Revert "Add Strict entity" pending further review

bplommer:
Reverts http4s/http4s#5804 (I was too trigger-happy)


armanbilge (approved):
Thanks! Next time I won't be so trigger-happy with my stamp of approval 😉 we're all just very excited about these changes :)

PR #5807, opened 2022-01-02
Add Strict entity

bplommer:
Reopening #5804 for further review.

### DIFF
--- a/core/shared/src/main/scala/org/http4s/Entity.scala
+++ b/core/shared/src/main/scala/org/http4s/Entity.scala
@@ -19,6 +19,7 @@ package org.http4s
 import cats.Monoid
 import cats.syntax.all._
 import cats.~>
+import fs2.Chunk
 import fs2.Pure
 import fs2.Stream
 
@@ -35,14 +36,30 @@ object Entity {
 
   // The type parameter aids type inference in Message constructors. This should be unnecessary when Messages are covariant.
   def empty[F[_]]: Entity[F] = Empty
+  def strict(chunk: Chunk[Byte]): Entity[Pure] = Strict(chunk)
 
   final case class Default[+F[_]](body: EntityBody[F], length: Option[Long]) extends Entity[F] {
     def ++[F1[x] >: F[x]](that: Entity[F1]): Entity[F1] = that match {
       case d: Default[F1] => Default(body ++ d.body, (length, d.length).mapN(_ + _))
+      case Strict(chunk) => Default(body ++ Stream.chunk(chunk), length.map(_ + chunk.size))
       case Empty => this
     }
 
     def translate[F1[x] >: F[x], G[_]](fk: F1 ~> G): Entity[G] = Default(body.translate(fk), length)
+  }
+
+  final case class Strict(chunk: Chunk[Byte]) extends Entity[Pure] {
+    val body: EntityBody[Pure] = Stream.chunk(chunk)
+
+    val length: Option[Long] = Some(chunk.size.toLong)
+
+    def ++[F1[x] >: Pure[x]](that: Entity[F1]): Entity[F1] = that match {
+      case d: Default[F1] => Default(body ++ d.body, d.length.map(chunk.size + _))
+      case Strict(chunk2) => Strict(chunk ++ chunk2)
+      case Empty => this
+    }
+
+    def translate[F1[x] >: Pure[x], G[_]](fk: F1 ~> G): Entity[G] = this
   }
 
   case object Empty extends Entity[Pure] {
--- a/core/shared/src/main/scala/org/http4s/EntityEncoder.scala
+++ b/core/shared/src/main/scala/org/http4s/EntityEncoder.scala
@@ -101,10 +101,7 @@ object EntityEncoder {
     * This constructor is a helper for types that can be serialized synchronously, for example a String.
     */
   def simple[A](hs: Header.ToRaw*)(toChunk: A => Chunk[Byte]): EntityEncoder.Pure[A] =
-    encodeBy(hs: _*) { a =>
-      val c = toChunk(a)
-      Entity(Stream.chunk(c), Some(c.size.toLong))
-    }
+    encodeBy(hs: _*)(a => Entity.strict(toChunk(a)))
 
   /** Encodes a value from its Show instance.  Too broad to be implicit, too useful to not exist. */
   def showEncoder[A](implicit charset: Charset = `UTF-8`, show: Show[A]): EntityEncoder.Pure[A] = {


bplommer:
Failure is a blaze flake.

core/shared/src/main/scala/org/http4s/Entity.scala
```diff
@@ -35,16 +36,32 @@ object Entity {
...
     def ++[F1[x] >: F[x]](that: Entity[F1]): Entity[F1] = that match {
       case d: Default[F1] => Default(body ++ d.body, (length, d.length).mapN(_ + _))
+      case Strict(chunk) => Default(body ++ Stream.chunk(chunk), length.map(_ + chunk.size))
       case Empty => this
     }
 
     def translate[F1[x] >: F[x], G[_]](fk: F1 ~> G): Entity[G] = Default(body.translate(fk), length)
   }
 
+  final case class Strict(chunk: Chunk[Byte]) extends Entity[Pure] {
```
RafalSumislawski, line 51: In @ChristopherDavenport's version that was a `case class Strict[F[_]](chunk: Chunk[Byte]) extends Entity[F]`, could you explain the reason for going with `Entity[Pure]` instead of `Entity[F]`. My intuition says this will cause some problems as at some point because `Client[IO]` and `Entity[Pure]` will collide.

bplommer: In this version `Entity` is covariant in `F`, so an `Entity[Pure]` is also an `Entity[IO]`. I don't expect this to cause any new problems given that `Entity` is already covariant, as is `Stream` which is what we currently use for `EntityBody`. Covariant `Entity` will also be needed for covariant `Message` per #5712.

armanbilge (approved):

core/shared/src/main/scala/org/http4s/Entity.scala
danicheg, line 51: I prefer to have `body` instead of `chunk`. Because `chunk` is more likely implementation details.

core/shared/src/main/scala/org/http4s/Entity.scala
```diff
@@ -35,16 +36,32 @@ object Entity {
...
+    val body: EntityBody[Pure] = Stream.chunk(chunk)
```
danicheg, line 52: `body` then could be `stream`, maybe.

danicheg (approved):
In general, LGTM

core/shared/src/main/scala/org/http4s/Entity.scala
armanbilge: These names are currently inherited from `Entity` itself. Do you think we should rename them for everything?
```scala
sealed trait Entity[+F[_]] {
  def body: EntityBody[F]
  def length: Option[Long]
  def ++[F1[x] >: F[x]](that: Entity[F1]): Entity[F1]
  def translate[F1[x] >: F[x], G[_]](fk: F1 ~> G): Entity[G]
}
```

danicheg: We can encapsulate the type of `body` to do it, like
```scala
abstract sealed trait Entity[+F[_]] {
  type EntityBodyType
  def body: EntityBodyType
  def length: Option[Long]
  def ++[F1[x] >: F[x]](that: Entity[F1]): Entity[F1]
  def translate[F1[x] >: F[x], G[_]](fk: F1 ~> G): Entity[G]
}
```

armanbilge: Huh, that's an interesting idea. How does somebody given an `Entity` use the body if they don't know the type?

danicheg: If we'll have `def stream: EntityBody[F]` I see no problem with that :)

armanbilge: So, you can only use `body` if you know the subclass of `Entity`, otherwise you use `stream` 👍 I'm a little confused then, why do we need to define `body` as part of the shared `Entity` interface then? Since it's very difficult to use unless you know the specific subclass of `Entity`. Instead, we could just use the `body` name on each subclass. WDYT?

bplommer: I agree with renaming `body` to `stream` (or maybe `byteStream`?) - not sure about renaming `chunk` to `body`, since the latter is less informative, but I don't have a very strong opinion on this. Another idea would be to call it `toChunk`, which doesn't carry any implication about the internal representation.

bplommer: I agree with @armanbilge that there doesn't seem much benefit adding a `BodyType` type member, but I'm also not keen on having a bunch of different `body` members with different types. It seems confusing.

armanbilge: I like `byteStream` and `toChunk` 👍

armanbilge: Well, I'm not sure actually. Does it make sense to `toChunk` an _entity_? An entity body for sure can be represented as a `Chunk`, but I'm not sure about `Entity` itself 🤔

bplommer: > what about just bytes?

But the chunk is *also* the bytes.

So maybe just `stream` and `chunk`? Or `byteStream` and `byteChunk`?

armanbilge: 😆 

I kind of want to leave it the way it is. Isn't `entity-body` the "official" name for this anyway?
```
entity-body    = *OCTET
```
https://www.w3.org/Protocols/rfc2616/rfc2616-sec7.html

bplommer: Fine by me 😆

armanbilge:
Re the naming thing above, I checked to see what akka-http does here.

It seems in their `HttpEntity`, they define the body as:
```scala
def dataBytes: Source[ByteString, Any]
```

https://github.com/akka/akka-http/blob/e8c35c2bb2e201d095523581f1cee8c62c93af8f/akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala#L133

Then, they have:
```scala
final case class Strict(contentType: ContentType, data: ByteString)
```

So, not particularly helpful IMO :)

I'm still partial to `body` for the top-level entity, deferring to the spec's terminology here. However, I do think @danicheg's concern about `chunk` is valid, not only because it is an implementation detail (how would we have named it if it was a `ByteVector`?), but because "chunk" is an over-loaded term (looking at you, chunked transfer encoding).

RafalSumislawski (approved):

bplommer:
Unless anyone strongly objects, I'm going to merge this as-is. There's plenty of time before 1.0 to hammer out the naming.

armanbilge:
I made a follow up: https://github.com/http4s/http4s/issues/5812

PR #5808, opened 2022-01-03
switch from tut to mdoc for 0.18.x

niij:
1. upgraded SBT to latest (and switched from 0.13.x to 1.x syntax while I was at it)
2. switched from tut to mdoc in the 0.18 branch.
    - I removed all of the `:silent` and fixed any of the errors that were preventing it from fully compiling the documentation examples in both tut and mdoc. Mostly this was due to redefining variables with the same name, but there were also issues with conflicting implicits imports where removing some unnecessary imports was the solution.
    - I opted to avoid doing `:reset` as much as possible as it requires re-importing everything each time (`import org.http4s._, ...`) and I didn't want to clutter up the docs.


danicheg (commented):
Hey! What was your motivation to do it for the 0.18-series? At the moment, we are supporting only 0.22-, 0.23-, and 1.0-series.

niij:
@danicheg sorry, I forgot to provide that context. I'd messaged Ross Baker about it already but the tl;dr is that the _generated_ docs didn't seem to match the API in at least one place.  For example: some of the [0.18 dsl _generated_ docs](https://http4s.org/v0.18/dsl/) incorrectly refer to creating routes with `HttpRoutes.of[IO]` instead of `HttpService[IO]`. `HttpRoutes.of[IO]` wasn't available until v0.19+.

The docs _source_ for the 0.18 didn't seem to have that error in it. I suspect the 0.18 docs, when last generated using tut over a year ago, was ran against a source that may have had those errors. Re-generating with tut is difficult now so I switched it to mdoc to match the currently supported versions to make re-generating the docs easier.

My motivation to remove all of the `:silent` annotations was to prevent that going forward in the rare case any other changes need to be made to these docs.

rossabaker (approved):
Thanks, @niij.

There won't be CI on this, because 0.18 is so old, it ran on Travis CI, but I was able to republish the docs after backporting a nix-shell to get the right JDK (doesn't build on 17) and Hugo.

rossabaker:
Didn't really target series/0.22, but the label will help with our release notes automation.

Issue #5715, opened 2021-12-10
Google searches still lead to 0.18 docs


bplommer:
For example, the top goole result for "http4s cors" (and the only one leading to http4s docs) is to https://http4s.org/v0.18/cors/. Fixing this / stopping it from happening will need some SEO expertise unfortunately (I've seen similar things for Doobie).

bplommer:
Perhaps the best solution to this is to take down the 0.18 docs and have the URLs redirect to a later version?

armanbilge:
https://developers.google.com/search/docs/advanced/crawling/block-indexing

Seems it's as simple as adding this tag to the relevant pages.
```html
<meta name="robots" content="noindex">
```

bplommer:
Won't that result in no docs showing up though?

armanbilge:
Well, I was thinking we'd just put it on the v0.18 docs (and any other old ones). Can't be any worse than taking them down?

bplommer:
I was thinking that if we make the 0.18 urls redirect to newer versions then we'll keep the existing search result positions but they'll go somewhere more relevant. But maybe that's not how it works?

armanbilge:
My search results look like this. First entry is 0.18, second is 1.0. So I assume that would replace it. But I won't claim to have any idea how these work :)

![Screen Shot 2022-01-02 at 11 03 23 AM](https://user-images.githubusercontent.com/3119428/147881622-64545296-e2cc-4162-818d-c401192de802.png)

bplommer:
Ok yes, I think you're right 🙂

rossabaker:
Some projects do a nice thing where the top of obsolete docs redirects to the latest with an ugly banner at the top.  We could implement that now, except GitHub pages doesn't support a 3xx we could point.  Rewrites are why we started hosting the site on Netlify, but bus factor of administration is an issue there, so we'd like to wean ourselves from that.  I don't know how to point 0.18 at the latest and _keep_ it there, without republishing from 0.18 every time there's a new version.  I guess it could always be N hops away.

Sadly, the recent changes to 0.18 came up because we have a service on that still, and a new developer doing their job found a mistake in it.  It would be a shame for it to go away entirely.

armanbilge:
>  It would be a shame for it to go away entirely.

IMHO I think leaving them live but off Google results is reasonable compromise.

bplommer:
[...]

PR #5811, opened 2022-01-05
Merge #5808 to 0.20

rossabaker:
Merges #5808 forward, in case any more changes come from the depths.  This review is mostly for GitHub Actions, and subsequent merges can be almost entirely "keep ours".


danicheg (approved):
You're definitely doing a whacking thing, so let's make it happen.

bplommer:
@rossabaker does this mean you currently have the necessary menta l context to do #5715 without undue excess pain?

rossabaker:
Commented on 0.18 over there.  

Sorry this tedious one drew attention.  I was hoping to do it quietly and let everyone else focus on something more fulfilling!

PR #5813, opened 2022-01-05
Specialize EntityDecoder for strict entities

bplommer:

### DIFF
--- a/core/shared/src/main/scala/org/http4s/EntityDecoder.scala
+++ b/core/shared/src/main/scala/org/http4s/EntityDecoder.scala
@@ -205,7 +205,13 @@ object EntityDecoder {
 
   /** Helper method which simply gathers the body into a single Chunk */
   def collectBinary[F[_]: Concurrent](m: Media[F]): DecodeResult[F, Chunk[Byte]] =
-    DecodeResult.success(m.body.chunks.compile.toVector.map(bytes => Chunk.concat(bytes)))
+    DecodeResult.success(
+      m.entity match {
+        case Entity.Empty => Chunk.empty[Byte].pure[F]
+        case Entity.Strict(chunk) => chunk.pure[F]
+        case d: Entity.Default[F] => d.body.compile.to(Chunk)
+      }
+    )
 
   /** Helper method which simply gathers the body into a single ByteVector */
   private def collectByteVector[F[_]: Concurrent](m: Media[F]): DecodeResult[F, ByteVector] =


bplommer:
I thought about moving this method into `Entity`. The thing that put me off was deciding what to name it. `chunkF`?

bplommer:
Or, we could have something like `Stream#compile`. So `entity.compile.toChunk` etc.

armanbilge:
Here's a a related/different idea: what if we have in `Entity`:
```scala
def toStrict[F[_]: Concurrent]: F[Entity.Strict[Byte]]
```
Then, you can grab the chunk directly.

bplommer:
The thing that slightly concerns me about that is that different `Entity` types have different semantics with respect to http (in ways I don't really understand). I think converting to a different kind of `Entity` just to collect its bytes risks muddying the concepts.

core/shared/src/main/scala/org/http4s/EntityDecoder.scala
```diff
@@ -205,7 +205,13 @@ object EntityDecoder {
...
+    DecodeResult.success(
+      m.entity match {
+        case Entity.Empty => Chunk.empty[Byte].pure[F]
+        case Entity.Strict(chunk) => chunk.pure[F]
+        case d: Entity.Default[F] => d.body.compile.to(Chunk)
+      }
+    )
 
   /** Helper method which simply gathers the body into a single ByteVector */
   private def collectByteVector[F[_]: Concurrent](m: Media[F]): DecodeResult[F, ByteVector] =
```
wjoel, line 217: Should maybe change this one similarly?

wjoel, line 217: Should maybe change this one similarly?

core/shared/src/main/scala/org/http4s/EntityDecoder.scala
```diff
@@ -205,7 +205,13 @@ object EntityDecoder {
 
   /** Helper method which simply gathers the body into a single Chunk */
   def collectBinary[F[_]: Concurrent](m: Media[F]): DecodeResult[F, Chunk[Byte]] =
-    DecodeResult.success(m.body.chunks.compile.toVector.map(bytes => Chunk.concat(bytes)))
+    DecodeResult.success(
+      m.entity match {
+        case Entity.Empty => Chunk.empty[Byte].pure[F]
+        case Entity.Strict(chunk) => chunk.pure[F]
+        case d: Entity.Default[F] => d.body.compile.to(Chunk)
```
rossabaker, line 212: Is this more efficient than what's there?  This line might be a good improvement for a backport.

rossabaker:
I might like `entity.compile.toChunk`, but what other compile ops are there to justify the `compile`?  There's `toByteVector`.

bplommer:
Superseded.

PR #5818, opened 2022-01-06
Merge series/0.20 to series/0.21

rossabaker:
Intentionally boring.  I kept our docs in the conflicts, which were already on mdoc, which is the change that motivated this venture into antiquity.  This small diff should be quite easy to roll forward into the Branches That Actually Matter.


rossabaker:
I'm using executive authority to merge.  This is one EOL branch to another and we've all got more interesting things to review.

PR #5822, opened 2022-01-06
Type annotations for public API things

danicheg:
This adds type annotations to public API functions/values and `private` to test things. The first one is a rule of thumb. The second one allows recognizing dead code. Another motivation to do that is to silence the linter of my $IDE 😄 
And yeah, probably Scalafix-ninja can write a fixing rule for that.


armanbilge:
Actually I think there is a built-in scalafix rule for this one :)
https://scalacenter.github.io/scalafix/docs/rules/ExplicitResultTypes.html

armanbilge:
@bplommer can we just add that here? Is this one of the ones you considered but decided against for some reason?
https://github.com/http4s/http4s/blob/94edb56d1abc050642e0a9e3ba2412de95626489/.scalafix.conf#L1-L6

bplommer:
Yep there's no reason we can't add it there.

bplommer:
It needs to go into both lists, and I really really need to document that properly before the bus factor on it goes down from 1 to 0.

server/src/test/scala/org/http4s/server/staticcontent/FileServiceSuite.scala
```diff
@@ -30,8 +30,9 @@ import java.io.File
 import java.nio.file._
 
 class FileServiceSuite extends Http4sSuite with StaticContentShared {
-  val defaultSystemPath = org.http4s.server.test.BuildInfo.test_resourceDirectory.getAbsolutePath
-  val routes = fileService(
+  private val defaultSystemPath =
+    org.http4s.server.test.BuildInfo.test_resourceDirectory.getAbsolutePath
+  val routes: HttpRoutes[IO] = fileService(
```
danicheg, line 35: This finding blew my mind. If remove this type annotation then tests are passed well. Or I miss something stupid obvious, or it's some magic away from Hogwarts.

rossabaker: I'm still confused what was found.

rossabaker:
This is an excellent rule in `src/main`, where we don't want any accidentally inferred types getting into the binary contract.  It feels like a chore in tests, where I'm more content to let type inference do its thing.  I guess the argument in tests is that tests should mirror production-style code to uncover the surprises that hit production-style code?

If we wanted to opt out on the tests, is there a simple way to configure scalafmt in that way?

server/src/test/scala/org/http4s/server/staticcontent/FileServiceSuite.scala
danicheg: I reverted changes https://github.com/http4s/http4s/pull/5822/commits/a3cba5fa108ae68ff29d83e156b6d165a7e85c66 that were causing the test to fail.

rossabaker:
I lost track of this one after the question about rule enforcement in main but not tests.  It doesn't look like the rule is being enforced at all at this point?

I think this is strictly forward progress, even if we're deferring automated enforcement. :+1: for resolving the conflicts and getting it in.

rossabaker (approved):

PR #5833, opened 2022-01-07

danicheg:
Something has accumulated.


core/shared/src/main/scala/org/http4s/headers/If-Range.scala
```diff
@@ -34,7 +34,7 @@ object `If-Range` {
   private val parser: Parser[`If-Range`] =
     EntityTag.parser
       .eitherOr(HttpDate.parser)
-      .map(_.fold(LastModified, ETag))
+      .map(_.fold(LastModified(_), ETag(_)))
```
danicheg, line 37: > The method `apply` is inserted. The auto insertion will be deprecated, please write `If-Range.ETag.apply` explicitly.

I consider that sad.

rossabaker (approved):

PR #5835, opened 2022-01-08
HttpMethodOverrider: simplify code

diesalbla:
Using `Alternative[Option]`, `as`, and `getOrElse`, is a rather convoluted way to say "if-then-else".

### DIFF
--- a/server/shared/src/main/scala/org/http4s/server/middleware/HttpMethodOverrider.scala
+++ b/server/shared/src/main/scala/org/http4s/server/middleware/HttpMethodOverrider.scala
@@ -21,8 +21,6 @@ package middleware
 import cats.Monad
 import cats.data.Kleisli
 import cats.effect._
-import cats.instances.option._
-import cats.syntax.alternative._
 import cats.syntax.flatMap._
 import cats.syntax.functor._
 import cats.~>
@@ -95,8 +93,6 @@ object HttpMethodOverrider {
   ): Http[F, G] = {
     val parseMethod = (m: String) => Method.fromString(m.toUpperCase)
 
-    val processRequestWithOriginalMethod = (req: Request[G]) => http(req)
-
     def processRequestWithMethod(
         req: Request[G],
         parseResult: ParseResult[Method],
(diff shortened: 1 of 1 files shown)


server/shared/src/main/scala/org/http4s/server/middleware/HttpMethodOverrider.scala
```diff
@@ -147,15 +143,12 @@ object HttpMethodOverrider {
...
+        case None => http(req)
       }
 
     Kleisli { (req: Request[G]) =>
-      config.overridableMethods
-        .contains(req.method)
-        .guard[Option]
-        .as(processRequest(req))
-        .getOrElse(processRequestWithOriginalMethod(req))
+      val isOverridden = config.overridableMethods.contains(req.method)
```
danicheg, line 150: As a fan of `fold` on `Boolean`, I'd rather do it like:
```
import mouse.boolean._

config.overridableMethods.contains(req.method).fold(processRequest(req), http(req))
```
Much more concise than using `Alternative` and `map` + `getOrElse` (and better by perf). But we haven't `Mouse` for now :( Though your variant probably is the best by perf.

danicheg (approved):
Great. Could we have this for the `series/0.22`? Anyway, I can cherry-pick easily.

armanbilge:
Do you mind if I try something?

armanbilge:
@mergifyio backport series/0.22

armanbilge:
Hmm, that's not very helpful :(

server/shared/src/main/scala/org/http4s/server/middleware/HttpMethodOverrider.scala
diesalbla: In my humble opinion, it is best to stick to the language Scala itself, and to the `if-then-else` operator that is taught on week 2 of any introductory programming course. Using a `fold` for `Boolean` seems to stretch that word a lot.

rossabaker: Also not a fan of boolean fold.  Makes me pause and think whether true or false comes first, and my guess was wrong.

rossabaker:
Cherry-picked to 0.22.

PR #5836, opened 2022-01-08
BlazeServerBuilder: simplify code.

diesalbla:
- Extract local variables and definitions for Vault contents.
- Use a for comprehension for the Option combination
- Replace Alternative[Option] with an if-then-else.

### DIFF
--- a/blaze-server/src/main/scala/org/http4s/blaze/server/BlazeServerBuilder.scala
+++ b/blaze-server/src/main/scala/org/http4s/blaze/server/BlazeServerBuilder.scala
@@ -18,7 +18,6 @@ package org.http4s
 package blaze
 package server
 
-import cats.Alternative
 import cats.Applicative
 import cats.data.Kleisli
 import cats.effect.Async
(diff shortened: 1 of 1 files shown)


blaze-server/src/main/scala/org/http4s/blaze/server/BlazeServerBuilder.scala
```diff
@@ -269,39 +268,34 @@ class BlazeServerBuilder[F[_]] private (
...
+
+            // Create SSLSession object only for https requests and if current SSL session is not empty.
+            // Here, each condition is checked inside a "flatMap" to handle possible "null" values
+            def secureSession: Option[SecureSession] =
+              for {
+                engine <- optionalSslEngine
+                session <- Option(engine.getSession)
+                hex <- Option(session.getId).map(ByteVector(_).toHex)
+                cipher <- Option(session.getCipherSuite)
+              } yield SecureSession(hex, cipher, deduceKeyLength(cipher), getCertChain(session))
```
rossabaker, line 292: Does Cats do anything fancy to optimize away intermediate options in the applicative style?  I don't think so, but it certainly can't monadically.

rossabaker (approved):
Picks cleanly back to 0.23.

blaze-server/src/main/scala/org/http4s/blaze/server/BlazeServerBuilder.scala
diesalbla: I had not bothered to look at these. Since these are `Option`, each line is just a comparison to `null`, so it does not seem to be the kind of thing that applicative or monadic application would optimise or parallelise.  Furthermore, note that the last 2 lines (`hex` and `cipher` depend on the result of the second line.

rossabaker:
Cherry picked to 0.23.

Issue #5256, opened 2021-09-22
Ember client fails `/no-content` route in RouteTestBattery


armanbilge:
Initially reported in https://github.com/http4s/http4s/pull/5241#issuecomment-923594689.

armanbilge:
This is in the linked discord chat, but FTR the logged response from the scaffold server that seemed to be tripping up ember client was:
```
HTTP/1.1 204 No Content Headers(Date: Tue, 21 Sep 2021 03:11:06 GMT) body=""
```

PR #5840, opened 2022-01-09
Re-enable broken ember client test

armanbilge:
Opening for CI.

Update:
Fixes https://github.com/http4s/http4s/issues/5256.

I think the real credit for the fix goes to @RafalSumislawski in https://github.com/http4s/http4s/pull/5587.


RafalSumislawski:
That's interesting. Do you how I fixed it? Was this issue caused by some bug in the scaffold?

armanbilge:
Yes, a bug in the scaffold was a possibility.

Although IMO it's still concerning that every other client could handle whatever the old scaffold was doing and ember client could not.

RafalSumislawski (approved):

PR #5841, opened 2022-01-09
Enable fatal warnings for the `lint` command

danicheg:
I find this is not optimal to use fatal warnings in the CI and not in local development. Users should get feedback as quickly as possible. Newly, I get into that situation on my own.
Also, @ChristopherDavenport pointed about that in https://github.com/http4s/http4s/discussions/5709. Probably we should add fatal warnings for `quicklint` too? WDYT?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -911,7 +911,8 @@ addCommandAlias(
   ";scalafixAll --triggered ;scalafixAll ;scalafmtAll ;scalafmtSbt",
 )
 
+// Use this command for checking before submitting a PR
 addCommandAlias(
   "lint",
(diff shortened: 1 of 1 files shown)


armanbilge:
👍 I think this is similar to what I proposed in https://github.com/typelevel/sbt-typelevel/issues/15.

However, I think unless you change the setting back at the end, running lint once will enable fatal warnings for the rest of the session.

danicheg:
Yeah, I also think that unsetting that option at the end of the chain is a good idea. But that would take some time too. If we think about the pattern - run `lint` and then commit changes, probably it's redundant?

armanbilge:
It's only redundant, if the user restarts their sbt session after every commit.

danicheg:
Accidentally I found out that `@nowarn` does not work for Scala 3.0.2 when fatal warnings are enabled. I tried to fight with all of them, but they as hydra. I close this and will reopen it for `series/0.23` only.

PR #5843, opened 2022-01-10
Enable fatal warnings in `lint`/`quicklint` commands

danicheg:
I find this is not optimal to use fatal warnings in the CI and not in local development. Users should get feedback as quickly as possible. Newly, I get into that situation on my own.
Also, @ChristopherDavenport pointed about that in #5709.
That is for the `series/0.23` (and not `series/0.22`) for the reason I described in https://github.com/http4s/http4s/pull/5841#issuecomment-1008708190.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1080,7 +1080,6 @@ def initCommands(additionalImports: String*) =
 // This won't actually release unless on Travis.
 addCommandAlias("ci", ";clean ;release with-defaults")
 
-// OrganizeImports needs to run separately to clean up after the other rules
 addCommandAlias(
   "quicklint",
   ";scalafixAll --triggered ;scalafixAll ;scalafmtAll ;scalafmtSbt",
@@ -1088,5 +1087,5 @@ addCommandAlias(
 
 addCommandAlias(
   "lint",
-  ";clean ;+test:compile ;+scalafixAll --triggered ;+scalafixAll ;+scalafmtAll ;scalafmtSbt ;+mimaReportBinaryIssues",
+  ";clean ;+test:compile ;scalafixAll --triggered ;scalafixAll ;+scalafmtAll ;scalafmtSbt ;+mimaReportBinaryIssues",
 )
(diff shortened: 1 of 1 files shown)


danicheg:
@armanbilge I tackled the point noticed in the previous PR about unsetting the compiler option at the end of the commands chain here.

build.sbt
```diff
@@ -1080,13 +1080,36 @@ def initCommands(additionalImports: String*) =
...
   "quicklint",
-  ";scalafixAll --triggered ;scalafixAll ;scalafmtAll ;scalafmtSbt",
+  List(
+    enableFatalWarnings,
+    "scalafixAll --triggered",
+    "scalafixAll",
+    "scalafmtAll",
+    "scalafmtSbt",
+    disableFatalWarnings,
+  ).mkString(" ;"),
```
armanbilge, line 1098: Does this work like? I thought that sbt command alias need to start with `; ` but maybe I am wrong. See https://github.com/typelevel/sbt-typelevel/blob/3936eba4f2cdffafa13678517630e9880d727150/kernel/src/main/scala/org/typelevel/sbt/TypelevelKernelPlugin.scala#L47

armanbilge (commented):
If you are interested, I would love your help with the similar problem in https://github.com/typelevel/sbt-typelevel/issues/15. Hopefully soon sbt-typelevel will be running most of the builds including http4s. I know you volunteered a lot of time on many projects (every project!?) in Typelevel so your experience would be very valuable :)

build.sbt
danicheg: It's easy to check
```scala
scala> List("foo", "bar", "baz").mkString(" ;")
res1: String = foo ;bar ;baz
```
That seems like what we need. And yeah, I checked manually that `lint`/`quicklint` works as expected :)

bplommer: A leading semicolon used to be required but no longer is. Now that it's not required, I feel like it's more elegant to have whitespace after rather than before the semicolon.

bplommer (approved):

danicheg:
I'm tired to re-run CI (5 attempts lmao). We lost our stability (pretty thin but cmon) at some point. Probably tests at the local have become more stable but not at the CI.

build.sbt
armanbilge: Huh, can't keep up :) thanks!

armanbilge (approved):

PR #5846, opened 2022-01-11
Update dev shell

rossabaker:
Probably only I care, but:

* Removes Hugo, now that we're on Laika
* Sets up everything else
* Uses the new typelevel-nix, so we can easily inherit future improvements

Remains fully opt-in.


armanbilge (approved):

rossabaker:
[back from reboot]

Also fixes that Graal 8 packages a Node 12 which locks up my system on the JavaScript branches!

Issue #5844, opened 2022-01-11
CI is hanging


armanbilge:
This is getting to be a pattern.

https://github.com/http4s/http4s/pull/5839#issuecomment-1008349663
https://github.com/http4s/http4s/runs/4767682086?check_suite_focus=true
https://github.com/http4s/http4s/runs/4767690258?check_suite_focus=true

rossabaker:
Is this also related to the bind exceptions in #5843?

rossabaker:
This is the earliest timeout I see, unless maybe someone reran an earlier one. https://github.com/http4s/http4s/actions/runs/1672936425

rossabaker:
Has anybody observed this hang locally yet?  I am trying and have not.

rossabaker:
But I can reproduce the bind exception erratically:

```
==> X org.http4s.ember.server.ConnectionSuite.close connection after response when request body stream is partially read 0.15s fs2.io.net.JavaScriptBindException: Address already in use
Caused by: scala.scalajs.js.JavaScriptException: Error: listen EADDRINUSE: address already in use 127.0.0.1:8080
```

armanbilge:
Huh, I haven't seen Node.js jobs failing in CI lately, are they caught up in this too?

I think I know the story behind those bind exceptions though, multiple suites starting ember on the same port. I can probably fix that.

armanbilge:
> Has anybody observed this hang locally yet? I am trying and have not.

Our local machines all probably have more cores than a GHA runner, maybe this can only be observed with 1-2 cores.

armanbilge:
@rossabaker couple quick questions—when you are running the test suite locally, I assume you are running `test` for everything at once right? i.e. not `rootJVM/test` and `rootJS/test`, like we do in CI. Also, have you ever observed the `BindException` on JVM?

I was trying to understand why it's happening, but I didn't take into account that on your machine the JVM and Node.js test suites may be running simultaneously and could be conflicting with each other. This is why it doesn't happen in CI, and this is why the `Test / parallelExecution := false` setting doesn't save us here, since ember.jvm and ember.js are technically independent projects in our build.

bplommer:
> Our local machines all probably have more cores than a GHA runner, maybe this can only be observed with 1-2 cores.

Should be reproducible in Docker?

rossabaker:
[...]

PR #5848, opened 2022-01-12
Auto-select available port for ember suites

armanbilge:
Should fix the `BindException` reported in https://github.com/http4s/http4s/issues/5844#issuecomment-1010403086.

I tried to target this down to 0.22, but the 0 port trick doesn't work there since ember returns its address with the port number that was provided and not the port that was actually bound. Seems like that can be fixed.


armanbilge:
Actually, hold on this until I convince myself there's no underlying bug here.

armanbilge:
What I've done here aligns with https://github.com/http4s/http4s/issues/5844#issuecomment-1011144495, so good to go.

danicheg (approved):
🚀

PR #5859, opened 2022-01-13
Add Request#isIdempotent

rossabaker:
Was trying to add default retry support to Blaze, but ran into a wall for lack of a Timer.  I'll try again in 0.23, but these were pure and already done.


armanbilge:
I was sure I've seen this before. We can do some cleanup in ember.

https://github.com/http4s/http4s/blob/11d0ddddabdfb6ef6d91d9cd73b7873d5ebe48dd/ember-client/src/main/scala/org/http4s/ember/client/internal/ClientHelpers.scala#L233

rossabaker:
Yeah, I had that refactored out when I hit the Blaze wall, but we can do it here.

armanbilge (approved):

danicheg (approved):

bplommer (approved):

PR #5861, opened 2022-01-13
Stale connection mitigation in blaze-client

rossabaker:
Adds an automated retry policy to blaze-client to account for connections that closed while idling in the pool.

This is less flexible than Ember's, mainly because we don't have access to a `Timer`.  Then again, this is the bare minimum suggested by HTTP specs, and we don't need all that power.  Like Ember's, this is a conservative default that only retries idempotent requests.

This, like Ember's default strategy, may interact surprisingly with an explicit retry middleware as the retry policies compose.

### DIFF
--- a/blaze-client/src/main/scala/org/http4s/blaze/client/BlazeClient.scala
+++ b/blaze-client/src/main/scala/org/http4s/blaze/client/BlazeClient.scala
@@ -58,6 +58,7 @@ object BlazeClient {
       requestTimeout = config.requestTimeout,
       scheduler = bits.ClientTickWheel,
       ec = ec,
+      retries = 0,
     )
 
   private[blaze] def makeClient[F[_], A <: BlazeConnection[F]](
@@ -66,6 +67,7 @@ object BlazeClient {
       requestTimeout: Duration,
       scheduler: TickWheelExecutor,
       ec: ExecutionContext,
+      retries: Int,
   )(implicit F: ConcurrentEffect[F]) =
     Client[F] { req =>
       Resource.suspend {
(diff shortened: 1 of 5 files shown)


blaze-client/src/main/scala/org/http4s/blaze/client/BlazeClient.scala
```diff
@@ -86,49 +88,53 @@ object BlazeClient {
...
-                new SocketException(s"HTTP connection closed: ${key}")
-              }
-              .map { (response: Resource[F, Response[F]]) =>
-                response.flatMap(r =>
-                  Resource.make(F.pure(r))(_ => manager.release(next.connection))
-                )
-              }
+        def loop(retriesRemaining: Int): F[Resource[F, Response[F]]] =
+          borrow
+            .use { next =>
```
rossabaker, line 93: Ayo, WTF, scalafmt?

blaze-client/src/main/scala/org/http4s/blaze/client/BlazeClient.scala
```diff
@@ -86,49 +88,53 @@ object BlazeClient {
...
+                    F.racePair(gate.get *> res, responseHeaderTimeoutF)
+                      .flatMap[Resource[F, Response[F]]] {
+                        case Left((r, fiber)) => fiber.cancel.as(r)
+                        case Right((fiber, t)) => fiber.cancel >> F.raiseError(t)
+                      }
+                  }
+                case _ => res
+              }
+            }
+            .recoverWith {
```
rossabaker, line 132: Concern: are we certain this only borrows one connection at a time?  This is the golden rule of the connection pool, and previous naive flatmaps have deadlocked the connection pool.

rossabaker: I gave the client a max of one connection per key, so I think we're good.

rossabaker:
Recommend viewing this one without whitespace diffs.  It's smaller than it looks.

ChristopherDavenport (approved):

blaze-client/src/main/scala/org/http4s/blaze/client/BlazeClientBuilder.scala
```diff
@@ -368,6 +428,7 @@ object BlazeClientBuilder {
       asynchronousChannelGroup = None,
       channelOptions = ChannelOptions(Vector.empty),
       customDnsResolver = None,
+      retries = 2,
```
danicheg, line 431: Not a blocker at all, I'm just interested in why '2' was chosen here?

danicheg (approved):
LGTM.

blaze-client/src/main/scala/org/http4s/blaze/client/BlazeClientBuilder.scala
rossabaker: 1. It matches Ember.
2. I think Ember matches an RFC, but damn if I can find it in there now.

danicheg: Got it, thanks!

PR #5864, opened 2022-01-13
Fix customDnsResolver in BlazeClientBuilder

rossabaker:
The copy method should use the current value, not a fresh default value.  Without this, the custom DNS resolver will be blanked out by any other `with*` call.


danicheg (approved):
Nice finding.

PR #5871, opened 2022-01-14
Deprecate client.Connection and client.ConnectionBuilder

rossabaker:
These are blaze-specific types and clutter the Client namespace.

We keep the traits in the blaze package, because some tests use non-`BlazeConnection` implementations.

All the affected types are private, so generous MiMa exceptions are granted.  I'd rather not, but I think the alternative is an avalanche of `@nowarn`.


danicheg (approved):
LGTM.

PR #5876, opened 2022-01-14
Fix some links in the scaladoc

danicheg:


danicheg:
Unfortunately, I didn't find anything related besides https://docs.scala-lang.org/style/scaladoc.html. But it's very sketchy.

Issue #5877, opened 2022-01-14
Cookie parsing discards valid cookies


Qthunder:
When parsing a raw header with multiple cookies, a a failure is returned if any cookie can not be parse. 
So

`Headers(Header.Raw(ci"Cookie", """key1=value1; key2={"aField":{}}"""))).cookies`

will return a parse failure - because the second cookie is invalid, containing non-escaped `"` character - rather than a list of one Cookie (the second).

Perhaps, it should try and salvage valid cookies, by first splitting up the raw header by semi-colons

Gitter discussion:
https://gitter.im/http4s/http4s?at=61e16a3e526fb77b317c2735

bplommer:
I agree. Note: the gitter channel is moribund, most dicussion is on Typelevel discord (https://discord.gg/ZWmFFV73X8)

rossabaker:
This is an interesting idea.  We already have `Headers#getWithWarnings` that returns an `Ior`.  This works great when there are multiple headers of the same name, and we can only parse some of them.

This is the same problem, but the header uses the `;`-delimiter.  (Which is actually quite complicated!  See `org.http4s.internal.parsing.Rfc7230.headerRep1`.)  To make this work generally, what we'd want for recurring headers is to split the semi-colons, and then construct an `Ior` from parsing the components.  Then `getWithWarnings` could work like it does for multiple fields.  This would be slower, because the parsers aren't geared to return an Ior, so I think we'd need to feed it multiple inputs. 🤔

Ben-Woolley:
Chrome is sending cookies from the domain + subdomain in web requests (e.g. example.com + my.example.com). We happen to be the subdomain and there is an invalid cookie from example.com that we have no control over.

PR #5878, opened 2022-01-14
Parse valid cookies #5877

Qthunder:


core/shared/src/main/scala/org/http4s/headers/Cookie.scala
```diff
@@ -28,23 +28,24 @@ object Cookie {
   def apply(head: RequestCookie, tail: RequestCookie*): `Cookie` =
     apply(NonEmptyList(head, tail.toList))
 
-  def parse(s: String): ParseResult[Cookie] =
-    ParseResult.fromParser(parser, "Invalid Cookie header")(s)
+  def parse(s: String): ParseResult[Cookie] = {
+    val (errors, cookies) = s
+      .split("; |;")
```
rossabaker, line 33: I erred earlier saying this was like comma separated headers and more general.  This also means it has a simpler grammar than the standard header list separation.  I think this is right and I was wrong.  Cookies are weird.

core/shared/src/main/scala/org/http4s/headers/Cookie.scala
```diff
@@ -28,23 +28,24 @@ object Cookie {
   def apply(head: RequestCookie, tail: RequestCookie*): `Cookie` =
     apply(NonEmptyList(head, tail.toList))
 
-  def parse(s: String): ParseResult[Cookie] =
-    ParseResult.fromParser(parser, "Invalid Cookie header")(s)
+  def parse(s: String): ParseResult[Cookie] = {
```
rossabaker, line 31: The logic seems right, but It's new to have this method silently discard errors when none of our other parsers behave this way.  I think it would be more principled if we:
1. Had a `parseWithWarnings(headerValue: String): Ior[NonEmptyList[ParseFailure], A]` on the `Header` typeclass.
2. For binary compatibility, gave it a default that delegates to `parse`.
3. Integrate it into the `Headers.getWithWarnings` that already exists.
4. Your method here is close to a good implementation of that `parseWithWarnings`.

I guess the real question is what relationship is expected between `get` and `getWithWarnings` when the `Ior` returns both.  My gut feeling is warnings are "fatal" and we return a Left, and this PR suggests warnings are "non-fatal" and we should return a best effort.  I'm curious what others think.

core/shared/src/main/scala/org/http4s/headers/Cookie.scala
```diff
@@ -28,23 +28,24 @@ object Cookie {
...
+    val (errors, cookies) = s
+      .split("; |;")
+      .toList
+      .filterNot(_.isEmpty)
+      .map(parser.parseAll)
```
rossabaker, line 36: Invoking the parser multiple times is going to hurt performance a bit, but our parser isn't currently built to return an Ior in one pass.  I think we could fix this later if it became a problem in practice.

rossabaker (changes requested):
Thanks!  I want to support this, and this is mostly good, but there's a design question lurking here and I'd like to see what the other maintainers think.

Ben-Woolley:
We've had a go at implementing `parseWithWarnings` as you've suggested, it's resulted in a string of `*AllValid` defs along the path that the original `parse` takes

PR #5881, opened 2022-01-15
Support parsing response bodies without Content-Length in ember

### LINKED ISSUE #5865: ember client gets empty body for http/1.0 assume close after body with no content-length
```
* HTTP 1.0, assume close after body
< HTTP/1.0 200 OK
< Date: Thu, 13 Jan 2022 REDACTED GMT
< Server: Apache
< Set-Cookie: PHPSESSID=REDACTED; path=/
< Expires: Thu, 19 Nov 1981 REDACTED GMT
< Cache-Control: no-store, no-cache, must-revalidate, post-check=0, pre-check=0
< Pragma: no-cache
< X-Frame-Options: SAMEORIGIN
< X-XSS-Protection: 1
< Strict-Transport-Security: max-age=31536000; includeSubDomains; preload
< Content-Security-Policy: default-src 'self' *.md-hq.com *.fxmedsupport.com  *.fontawesome.com *.stripe.com *.googleapis.com *.gstatic.com 'unsafe-inline' 'unsafe-hashes' 'unsafe-eval'; img-src * data:;
< Vary: Accept-Encoding
< Connection: close
< Content-Type: application/json; charset=utf-8
```
Response header looks like this out of curl but from ember we get no data. 

I think ember assumes that no content-length means content-length is 0

wemrysi:
Adds support for parsing response bodies without a `Content-Length` header to `ember`.

Fixes #5865

### DIFF
--- a/ember-core/src/main/scala/org/http4s/ember/core/Parser.scala
+++ b/ember-core/src/main/scala/org/http4s/ember/core/Parser.scala
@@ -295,7 +295,13 @@ private[ember] object Parser {
     def parser[F[_]: Concurrent](maxHeaderSize: Int)(
         buffer: Array[Byte],
         read: Read[F],
-    ): F[(Response[F], Drain[F])] =
+    ): F[(Response[F], Drain[F])] = {
+      // per https://httpwg.org/specs/rfc7230.html#rfc.section.3.3.3
+      def expectNoBody(status: Status): Boolean =
+        status == Status.NoContent ||
+          status == Status.NotModified ||
+          status.responseClass == Status.Informational
+
       for {
         t <- MessageP.recurseFind(buffer, read, maxHeaderSize)(ibuffer =>
           EitherT(RespPrelude.parsePrelude[F](ibuffer, maxHeaderSize))
@@ -325,13 +331,19 @@ private[ember] object Parser {
                   ) ->
                   rest.get
             }
+          } else if (expectNoBody(prelude.status)) {
+            (baseResp -> (Some(finalBuffer): Option[Array[Byte]]).pure[F]).pure[F]
           } else {
-            Body.parseFixedBody(headerP.contentLength.getOrElse(0L), finalBuffer, read).map {
-              case (bodyStream, drain) =>
+            headerP.contentLength
+              .fold(Body.parseUnknownBody(finalBuffer, read))(
+                Body.parseFixedBody(_, finalBuffer, read)
+              )
+              .map { case (bodyStream, drain) =>
                 baseResp.withBodyStream(bodyStream) -> drain
-            }
+              }
           }
       } yield resp
+    }
 
     object RespPrelude {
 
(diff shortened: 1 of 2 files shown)


ChristopherDavenport:
This seems good in a lot of ways, but I think we are also exposing a massive footgun as well.

```
If this is a request message and none of the above are true, then the message body length is zero (no message body is present).

Otherwise, this is a response message without a declared message body length, so the message body length is determined by the number of octets received prior to the server closing the connection.
```

So effectively we turn the connection into an infinite tunnel, and anyone using entity decoders could be opened to reading an infinite body into a type. With chunked this was explicit-ish. But with this its the absence of content-type that makes it easy to shoot yourself in the foot. The more we do this the more I wonder if EntityLimiter should be much more present somehow, because this is way too easy to make explode.

Nothing about this is unique to this PR, but it requires some thought. I think this PR is right, but it raises some concerns to how to manage these entities.

ChristopherDavenport (approved):
I think this is correct, my issue is actually just awareness from this, not this itself.

ChristopherDavenport:
Redirected this conversation to  https://github.com/http4s/http4s/issues/5883

ember-core/src/main/scala/org/http4s/ember/core/Parser.scala
```diff
@@ -325,13 +331,19 @@ private[ember] object Parser {
                   ) ->
                   rest.get
             }
+          } else if (expectNoBody(prelude.status)) {
+            (baseResp -> (Some(finalBuffer): Option[Array[Byte]]).pure[F]).pure[F]
           } else {
-            Body.parseFixedBody(headerP.contentLength.getOrElse(0L), finalBuffer, read).map {
-              case (bodyStream, drain) =>
+            headerP.contentLength
+              .fold(Body.parseUnknownBody(finalBuffer, read))(
```
danicheg, line 338: I'm curious should be that behavior be manageable? Though I can't imagine that somebody wants this specific behavior.

danicheg (commented):
Speaking about infinite read, we have a little defense from it right now - it's a timeout on reading. Though I think that some explicit limit (via public API) in the case would be nice to have. But yeah, this is not a topic of this PR.

wemrysi:
> So effectively we turn the connection into an infinite tunnel, and anyone using entity decoders could be opened to reading an infinite body into a type

To play devil's advocate, couldn't a similarly malicious or misconfigured endpoint just return `Content-Length: ${Long.MaxValue}` also causing an effectively infinite read?

> my issue is actually just awareness from this, not this itself

Yeah, I was also unaware of this part of the spec until my colleague encountered the unexpected behavior.

rossabaker:
>> So effectively we turn the connection into an infinite tunnel, and anyone using entity decoders could be opened to reading an infinite body into a type

> To play devil's advocate, couldn't a similarly malicious or misconfigured endpoint just return Content-Length: ${Long.MaxValue} also causing an effectively infinite read?

Or a `Transfer-Encoding: chunked` without a terminal chunk.  It's a real concern, but I don't think this (inadvisable) response encoding opens any new holes.  At some point, the decoder needs to scream "Enough!"

PR #5882, opened 2022-01-15
Abstract/Generalise ContextRequest and ContextResponse

diesalbla:
- Replace "ContextRequest" and "ContextResponse" with a common class "WithContext", which does not take any parameter of Kind 1. Some existing methods, and the `apply` constructor, are replicated in syntax extension and objects for _source_-compatibility.
- Introduce a 3-parameter GenHttp type alias, that extends on the existing Http one with a Kind-1 type parameter, on the Request. An Http is just the case where the new parameter is ID.  A ContextRoutes is the case where the new parameter is "WithContext" and the response wrapper is an OptionT.


bplommer:
The build is failing on use-site exhaustivity checks - what change would be required to fix them, and is it possible to change `WithContext` so that a use-site change isn't needed?

bplommer (changes requested):
[tagging as "requested changes"]

diesalbla:
@bplommer The build seems to be working fine as of now. It was rebased to `series/0.23`.

isomarcte:
This isn't bincompat, though, right? I thought we checked that in CI, but now I'm confused.

armanbilge:
Probably CI needs to be re-run, lets' see.

armanbilge:
There are those mima failures. Note: changing the base branch on GH does not re-run CI.

isomarcte:
I'm on the fence about this change. By making `WithContext[Msg, A]` take parameters for both of its members, we've basically made it a `Tuple2`. This is more general, but if we want to go that route I'd just use `Tuple2` directly. I think _if_ `ContextRequest` and `ContextResponse` have any utility, it would be in that they are specific to `Request` and `Response`. 

FWIW, I could be convinced either way here on whether or not to keep or remove `ContextRequest`/`ContextResponse` and replace them with `Tuple2`, but I'm concerned that having `WithContext[A, B]` may be the lesser of the three options.

diesalbla:
> I'm on the fence about this change. By making WithContext[Msg, A] take parameters for both of its members, we've basically made it a Tuple2. This is more general, but if we want to go that route I'd just use Tuple2 directly

A `Tuple2` is just a case class with two "anonymous" parameters, `_1` and `_2`. Having a separate case class allows one to have more informative field names. Otherwise, users of the library would need to read the documentation to know which one is the request or context (if they get mixed up). IMHO that is an argument of favour of case classes.

PR #5884, opened 2022-01-16
Refactor some `Resource` usage

danicheg:
A little refactoring.

### DIFF
--- a/client/shared/src/main/scala/org/http4s/client/Client.scala
+++ b/client/shared/src/main/scala/org/http4s/client/Client.scala
@@ -222,9 +222,10 @@ object Client {
             }
           val req0 =
             addHostHeaderIfUriIsAbsolute(req.withBodyStream(go(req.body).stream))
+
           Resource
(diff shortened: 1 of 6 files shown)


client/shared/src/main/scala/org/http4s/client/Client.scala
```diff
@@ -222,9 +222,10 @@ object Client {
             }
           val req0 =
             addHostHeaderIfUriIsAbsolute(req.withBodyStream(go(req.body).stream))
+
           Resource
             .eval(app(req0))
-            .flatTap(_ => Resource.make(F.unit)(_ => disposed.set(true)))
```
danicheg, line 227: There is a small behavior change here. In the previous one if processing `app(req0)` leads to an exception, then `disposed.set(true)` will never happen. With that change, the finalizer will actually run. I consider it's not a bug, though semantically maybe is the more correct to run finalizer in the case.

rossabaker (approved):

PR #5887, opened 2022-01-16
Use blocking thread pool in the Prometheus metrics

danicheg:
`CollectorRegistry.clear`, `CollectorRegistry.register`, `CollectorRegistry.unregister` may block threads so it's better to use blocking thread pool.


rossabaker:
What sort of blocking do they do?  A synchronized lock?

danicheg:
@rossabaker Yep, it uses synchronized locks.

rossabaker (approved):

PR #5888, opened 2022-01-17

danicheg:


client/shared/src/main/scala/org/http4s/client/package.scala
```diff
@@ -29,6 +29,7 @@ package object client extends ClientTypes {
 trait ClientTypes {
   import org.http4s.client._
 
+  @deprecated("Is a Blaze detail.  Will be removed from public API.", "0.22.9")
```
danicheg, line 32: _Reminder_: we need to delete this when merging `series/0.23` into `main`.

rossabaker (approved):
:+1: if green

danicheg:
Uh oh, within `series/0.23` have the 'green CI' isn't so easy 😞

rossabaker:
Yeah, I am trying to fix one more thing in Blaze for $WORK, and then I want to focus on a release train, but CI is in terrible shape in 0.23, so that's going to be a blocker.

PR #5837, opened 2022-01-09
FileService Server Middleware: readability

diesalbla:
Extract logic to local methods.

PR #5889, opened 2022-01-18

danicheg:
That wasn't a pleasure cruise, so check this intently, please. Especially changes about `FileService`.


rossabaker (approved):

server/shared/src/main/scala/org/http4s/server/staticcontent/FileService.scala
```diff
@@ -32,93 +32,158 @@ import org.http4s.server.middleware.TranslateUri
...
 import scala.util.control.NoStackTrace
 
 object FileService {
   private[this] val logger = getLogger
 
-  type PathCollector[F[_]] = (Path, Config[F], Request[F]) => OptionT[F, Response[F]]
+  @deprecated("use FileService.Fs2PathCollector", "0.23.8")
+  type PathCollector[F[_]] = (File, Config[F], Request[F]) => OptionT[F, Response[F]]
+
+  type Fs2PathCollector[F[_]] = (Path, Config[F], Request[F]) => OptionT[F, Response[F]]
```
armanbilge, line 44: This was correct before.

server/shared/src/main/scala/org/http4s/server/staticcontent/FileService.scala
```diff
@@ -32,93 +32,158 @@ import org.http4s.server.middleware.TranslateUri
...
+      "Config is no longer a case class. The productElement method is provided for binary compatibility.",
+      "0.23.8",
+    )
+    override def productElement(n: Int): Any = n match {
+      case 0 => systemPath
+      case 1 => fs2PathCollector
+      case 2 => pathPrefix
+      case 3 => bufferSize
+      case 4 => cacheStrategy
+    }
```
armanbilge, line 80: These should have been removed.

armanbilge (changes requested):
I think `FileService` regressed. Maybe it would be better to keep what's on main and replay https://github.com/http4s/http4s/pull/5837.

ember-server/jvm/src/test/scala/org/http4s/ember/server/EmberServerMtlsSuite.scala
```diff
@@ -125,6 +126,7 @@ class EmberServerMtlsSuite extends Http4sSuite {
       tlsContext <- authTlsClientContext
       emberServer <- EmberServerBuilder
         .default[IO]
+        .withPort(ip4s.Port.fromInt(0).get)
```
bplommer, line 129: Not a blocker but would be nice to have a string interpolator for this

bplommer (approved):
Approving this but most of the merge conflicts aren't mine so don't give my approval too much weight.

ember-server/jvm/src/test/scala/org/http4s/ember/server/EmberServerMtlsSuite.scala
armanbilge: Ah, that's on me, sorry. We should upstream that change from 0.23 actually.

bplommer: No worries, this wasn't really the place to bring it up anyway

server/shared/src/main/scala/org/http4s/server/staticcontent/FileService.scala
danicheg: Thanks! I felt that something was wrong.

armanbilge (approved):
Thanks for fixing ❤️

PR #5895, opened 2022-01-19
Merge Series/0.23 to main

bplommer:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,7 +28,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        scala: [2.13.7, 2.12.15, 3.1.0]
+        scala: [2.13.8, 2.12.15, 3.1.0]
         java: [temurin@8, temurin@11, temurin@17]
         ci: [ciJVM, ciNodeJS]
         exclude:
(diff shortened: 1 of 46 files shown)


danicheg (approved):

bplommer:
Lots of tests failing with `BindException`.

bplommer:
Too many tests failing with `BindException` for it to be a flake. But none of the new changes here look like they would cause that. I feel like @armanbilge might have an idea about this?

armanbilge:
Huh, that's odd that so many failed. Also I can't tell for sure but are all the client suites running at the same time for some reason? It seems like the JVM scaffold server is always binding to the same port, so if multiple suites using it start in parallel it runs into trouble. Let me see if I can chase it down.

danicheg:
This seems to be for all `series/0.23` branches too #5896, #5897

armanbilge:
No, the JVM scaffold seems to be using port 0. So then I don't know what's going on anymore 😕 

https://github.com/http4s/http4s/blob/e1062470658a302d7dc5680672f6787e4ee34a39/client/jvm/src/test/scala/org/http4s/client/scaffold/ServerScaffold.scala#L65

bplommer:
Possibly just GHA playing up?

armanbilge:
I have a fix incoming.

bplommer:
I'll try to get back to this in the next couple of days but other people should feel free to take over

danicheg:
I could give it a try if you don't mind. I'm just battling for the 'merge-ninja' achievement 😆

bplommer:
Go for it 😄

bplommer:
Superseded by #5925

PR #5897, opened 2022-01-19
Use blocking thread pool in the `JettyClient#resource`

danicheg:
`org.eclipse.jetty.util.component.AbstractLifeCycle#stop` uses under the hood synchronized lock, so blocking thread pool is more desired to use.

### DIFF
--- a/jetty-client/src/main/scala/org/http4s/jetty/client/JettyClient.scala
+++ b/jetty-client/src/main/scala/org/http4s/jetty/client/JettyClient.scala
@@ -62,9 +62,10 @@ object JettyClient {
           })
         }
       )
-    val dispose = F
-      .delay(client.stop())
-      .handleErrorWith(t => F.delay(logger.error(t)("Unable to shut down Jetty client")))
+    val dispose =
+      F
+        .blocking(client.stop())
+        .handleErrorWith(t => F.delay(logger.error(t)("Unable to shut down Jetty client")))
     Resource.make(acquire)(_ => dispose)
   }
 


bplommer (approved):

jetty-client/src/main/scala/org/http4s/jetty/client/JettyClient.scala
```diff
@@ -62,9 +62,10 @@ object JettyClient {
...
         }
       )
-    val dispose = F
-      .delay(client.stop())
-      .handleErrorWith(t => F.delay(logger.error(t)("Unable to shut down Jetty client")))
+    val dispose =
+      F
+        .blocking(client.stop())
+        .handleErrorWith(t => F.delay(logger.error(t)("Unable to shut down Jetty client")))
     Resource.make(acquire)(_ => dispose)
```
bplommer, line 69: Kinda unrelated but shouldn't we dispose regardless of whether acquire succeeds?
```suggestion
    Resource.eval(acquire).onFinalize(dispose)
```

danicheg: Probably that would mean behavior change. According to the contract, `client` may come to the `resource` from outside.

rossabaker: I guess we want to stop it if and only if we start it?

bplommer: Yeah, that makes sense.

PR #5899, opened 2022-01-19
maxIdleDuration on blaze connections

rossabaker:
Discard connections that have been idle for a specified limit.  This is a means of stale connection mitigation.

We take the tack of checking them at borrow time, rather than running a background thread to evict them.  This keeps it simple in a pool that's already difficult to maintain.  If you want to do something pretty, go work on keypool.


rossabaker:
Still needs tests.  We don't have a Clock, so it's going to be gruesome.

ChristopherDavenport (approved):

blaze-client/src/main/scala/org/http4s/blaze/client/PoolManager.scala
```diff
@@ -173,14 +205,27 @@ private final class PoolManager[F[_], A <: Connection[F]](
...
+                        )
+                      ) *>
+                        decrConnection(key) *>
+                        F.delay(conn.shutdown()) *>
+                        go()
+                    case _ =>
+                      F.delay(logger.debug(s"Recycling connection for $key: $stats")) *>
+                        F.delay(callback(Right(NextConnection(conn, fresh = false))))
+                  }
+                }
```
diesalbla, line 228: Perhaps a flatter outline? 
```suggestion
              case Some(conn) if conn.isClosed =>
                F.delay(logger.debug(s"Evicting closed connection for $key: $stats")) *>
                  decrConnection(key) *>
                  go()
              case Some(conn) if conn.borrowDeadline.exists(_.isOverdue()) => 
                F.delay(logger.debug(s"Shutting down and evicting expired connection for $key: $stats")) *>
                  decrConnection(key) *>
                  F.delay(conn.shutdown()) *>
                  go()
              case Some(conn) =>
                F.delay(logger.debug(s"Recycling connection for $key: $stats")) *>
                  F.delay(callback(Right(NextConnection(conn, fresh = false))))
              }
```

diesalbla (approved):

blaze-client/src/main/scala/org/http4s/blaze/client/PoolManager.scala
rossabaker: Is scalac smart enough to do the Option match once and then the guards?

Compared to all the network stuff, checking the option two extra times really makes no difference anyway, and this is cleaner.

blaze-client/src/main/scala/org/http4s/blaze/client/Connection.scala
```diff
@@ -33,4 +35,7 @@ private[client] trait Connection[F[_]] {
 
   /** The key for requests we are able to serve */
   def requestKey: RequestKey
+
+  /** Deadline for the connection to be borrowed before we evict it */
+  @volatile var borrowDeadline: Option[Deadline] = None
```
danicheg, line 40: I'm wondering if the `borrowDeadline` is a property of the `Connection` actually. Or maybe the `PoolManager` should own the fact when a connection needs to borrow. 🤔 WDYT?

rossabaker: I think it needs to be _associated_ with a `Connection`, but it could be the `PoolManager` that maintains the association.  It would need to wrap some metadata around the connections to do so.  It could probably get rid of the mutation then. 🤔

danicheg (approved):
Thanks, that's awesome!

PR #5902, opened 2022-01-19
Fix scala-steward.conf syntax

rossabaker:

### DIFF
--- a/.scala-steward.conf
+++ b/.scala-steward.conf
@@ -4,7 +4,7 @@ updates.pin = [
   { groupId = "org.eclipse.jetty.http2", version = "9." },
   { groupId = "org.apache.tomcat", version = "9." },
   # Servlet 4 breaks Jetty 9 and Tomcat 9
-  { groupId = "javax.servlet", version = "3." },
+  { groupId = "javax.servlet", version = "3." }
 ]
 
 updates.ignore = [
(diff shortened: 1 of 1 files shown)


rossabaker:
I think this is why Tomcat appeared in #5983.

armanbilge (approved):
We need a CI validator for this ...

rossabaker:
It's HOCON.  An sbt plugin around Typesafe Config would be plausible.  I set up an Emacs linter based on pyhocon, but it's a different implementation, and we wouldn't want Python in our CI.

rossabaker:
You could go deeper with a semantic parser, but then you'd need a @scala-steward library.

armanbilge:
Yeah, an sbt plugin could be nice. HOCON is just JSON so you could define in sbt and have it generated by sbt-gh-actions and it doesn't need to be human-editable.

fthomas:
> You could go deeper with a semantic linter, but then you'd need a @scala-steward library.

This already exists: https://repo1.maven.org/maven2/org/scala-steward/scala-steward-core_2.13/0.13.0/. It is probably not too hard to write a scala-cli script that uses the above JAR to parse and decode `.scala-steward.conf` as part of the CI.

@danicheg (who is still waiting for an answer from me) proposed another solution to this problem in https://github.com/scala-steward-org/scala-steward/pull/2446.

danicheg:
Argh. Shame on me with that 😞  I am itching to fix that in some way.

PR #5907, opened 2022-01-19
Add EntityEncoder[F, ByteVector]

sideeffffect:
Continuation of https://github.com/http4s/http4s/pull/5195

/cc @ChristopherDavenport

### DIFF
--- a/core/src/main/scala/org/http4s/EntityDecoder.scala
+++ b/core/src/main/scala/org/http4s/EntityDecoder.scala
@@ -206,11 +206,11 @@ object EntityDecoder {
 
   /** Helper method which simply gathers the body into a single Chunk */
   def collectBinary[F[_]: Sync](m: Media[F]): DecodeResult[F, Chunk[Byte]] =
-    DecodeResult.success(m.body.chunks.compile.toVector.map(Chunk.concatBytes))
+    DecodeResult.success(m.body.chunks.compile.to(Chunk))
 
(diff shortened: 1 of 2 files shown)


bplommer (approved):

core/src/main/scala/org/http4s/EntityDecoder.scala
```diff
@@ -206,11 +206,11 @@ object EntityDecoder {
 
   /** Helper method which simply gathers the body into a single Chunk */
   def collectBinary[F[_]: Sync](m: Media[F]): DecodeResult[F, Chunk[Byte]] =
-    DecodeResult.success(m.body.chunks.compile.toVector.map(Chunk.concatBytes))
+    DecodeResult.success(m.body.chunks.compile.to(Chunk).map(_.flatten))
```
rossabaker, line 209: Is this as fast?  `Chunk.concatBytes` knows the total size in advance.  I'm not sure that this does.

rossabaker: Well, this should be pretty reasonable.  It iterates the list of chunks twice, but the first pass calculates the size hint so there isn't recopying.  This should be fine.

https://github.com/typelevel/fs2/blob/0f994722fb96fbe4b6f41cde5a9a59af72c76a47/core/shared/src/main/scala/fs2/Chunk.scala#L109-L119

rossabaker (approved):

PR #5908, opened 2022-01-20
Enable snapshots for 0.22

armanbilge:
I forgot to do this. We don't currently have snapshots for stable branches, do we want this? Personally I like snapshots and they help mitigate infrequent releases 🙃


rossabaker (approved):
I would like to release more often.  The release notes for the next one are giving me the sweats.  But I have no problem with this, either.

Issue #5919, opened 2022-01-21
http4s.org main page down


reactormonk:
https://http4s.org/

![Screenshot_20220121_012520](https://user-images.githubusercontent.com/49185/150443648-58a3ec4f-2a12-40a6-b552-c255c2defdc5.png)

<hr />

[Added by @rossabaker]:

* [X] restore site
* [X] recreate cert for gh-pages
* [x] restore Laika site for 0.22
* [x] restore Laika site for 0.23
* [x] fix localhost link in header of 1.0
* [X] recreate jdk-http-client cert for gh-pages
* [x] restore jdk-http-client site

armanbilge:
Thanks for reporting.

This is a combination of our site publishing pipeline already being broken + enabling snapshots on 0.22 (I forgot that publishing the site is tagged on to the end of a publish job).

rossabaker:
Do we need to roll back the snapshots and republish the website project manually?

armanbilge:
Yeah, that's probably the easiest fix. To restore the website we could probably just rollback the gh-pages branch as well.

I would like to fix the site, but it's too daunting right now.

rossabaker:
It's easy to restore the site.  I'm just mostly worried about the problem coming back.  I'll rollback #5908 and then roll back the branch.

rossabaker:
a8fa9cfa33c0a2fb2d8b0cf7a405c1a86fb3b38a should restore the site, and #5920 should stop it from happening again until we have a better idea.

rossabaker:
This is back, probably from the 0.22.9 release this morning.

armanbilge:
I actually never saw the site restored even after https://github.com/http4s/http4s/commit/a8fa9cfa33c0a2fb2d8b0cf7a405c1a86fb3b38a.

rossabaker:
The restoration job timed out, and I'm the only one who can see it, because Netlify doesn't have open source plans.  We need to get that rolled back to GitHub, but it's going to break the redirects when we do.

I reverted it to January 7, before the snapshot published, but now 0.22 and 0.23 have the Hugo theme, and the header in the 1.0 version links to 127.0.0.1, and I don't understand how any of it works anymore. :shrug:

rossabaker:
I switched back to GitHub Pages.  Things will be broken, but things were already broken, and now other people can help.

rossabaker:
It may take 30-60 minutes for that cert error to go away.

[...]

PR #5920, opened 2022-01-21
Revert "Enable snapshots for 0.22" (#5908)

rossabaker:
Temporary, to deal with #5919.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -116,7 +116,7 @@ jobs:
   publish:
     name: Publish Artifacts
     needs: [build]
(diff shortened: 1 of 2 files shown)


armanbilge (approved):
Well, at least it helped us solve the header thing once and for all.

rossabaker:
I've lost my understanding of how the site works in the Laika Era.  I don't object to these snapshots at all.  I just forget how the pieces fit.

armanbilge:
Crap, in my PR there were some mima fixes as well that had slipped through. We need those.

PR #5926, opened 2022-01-22
Merge 0.22 -> 0.23

rossabaker:


armanbilge:
I'm going to take a pass through and try and update the build.

rossabaker:
Many of the conflicts were related to the sbt-http4s-org-0.10, and I did my best, but I'm sure I missed a few things.  That and the blaze changes were the big two tumults.

armanbilge:
Yeah, the divide between 0.22 and 0.23 wrt the build is too big. But, it will be a lot nicer now b/c I was able to upstream some stuff into sbt-typelevel.

armanbilge:
And now, we fight mima.

rossabaker:
Oh, crap.  I thought quicklint ran that and we were clear.

armanbilge:
Ha ha, no way. There's never-before-reported Scala 3 and Scala.js issues to go through.

armanbilge:
Ok, so these are legit bin-compat breaks on Scala 3 that happened in https://github.com/http4s/http4s/pull/5758 when we de-case-classed `FileService#Config`. The problem is that the binary encoding of `case class`es changed between Scala 2/3 such that it is impossible to imitate it without splitting our sources across 2/3. E.g. the return type of `unapply` on Scala 2 is `Option[Config]`, on Scala 3 it is just `Config` thanks to [Option-less Pattern Matching™](https://docs.scala-lang.org/scala3/reference/changed-features/pattern-matching.html).

```
[error]  * static method unapply(org.http4s.server.staticcontent.FileService#Config)org.http4s.server.staticcontent.FileService#Config in class org.http4s.server.staticcontent.FileService#Config has a different result type in current version, where it is scala.Option rather than org.http4s.server.staticcontent.FileService#Config
[error]    filter with: ProblemFilters.exclude[IncompatibleResultTypeProblem]("org.http4s.server.staticcontent.FileService#Config.unapply")
[error]  * the type hierarchy of object org.http4s.server.staticcontent.FileService#Config is different in current version. Missing types {scala.deriving.Mirror$Product,scala.deriving.Mirror}
[error]    filter with: ProblemFilters.exclude[MissingTypesProblem]("org.http4s.server.staticcontent.FileService$Config$")
[error]  * method unapply(org.http4s.server.staticcontent.FileService#Config)org.http4s.server.staticcontent.FileService#Config in object org.http4s.server.staticcontent.FileService#Config has a different result type in current version, where it is scala.Option rather than org.http4s.server.staticcontent.FileService#Config
[error]    filter with: ProblemFilters.exclude[IncompatibleResultTypeProblem]("org.http4s.server.staticcontent.FileService#Config.unapply")
```

I'd rather revert https://github.com/http4s/http4s/pull/5758 than split our sources. But I'll also make an argument that we look the other way—this breakage would only affect Scala 3 users who are doing stuff with `FileService#Config` other than just constructing and directly passing to `FileService` which I assume is 99.9% of uses. Who in the hell is pattern matching it, or doing `Mirror`-based derivation?

rossabaker:
What would it cost to "do it right"?  A platform companion object for the differing unapplies, and a platform trait to provide the Mirror?  In a shim trait in a Scala 2 and Scala 3 directory?

I am loath to make exceptions, but this is an exotic case on an as yet little used Scala version, so I'd look the other way as long as we flag it in the notes.

armanbilge:
> What would it cost to "do it right"? A platform companion object for the differing unapplies, and a platform trait to provide the Mirror? In a shim trait in a Scala 2 and Scala 3 directory?

I think you're right. That doesn't sound too bad. I was feeling particularly pessimistic since it was a case class nested inside of an object, but maybe that doesn't matter. I will muster up the will to try and fix this.

server/shared/src/main/scala-3/org/http4s/server/staticcontent/FileServiceConfigPlatform.scala
```diff
@@ -0,0 +1,31 @@
...
+  def fromProduct(p: Product): MirroredMonoType = {
+    type F[_] = Any
+    FileService.Config.apply[F](
+      p.productElement(0).asInstanceOf[String],
+      p.productElement(1).asInstanceOf[FileService.PathCollector[F]],
+      p.productElement(2).asInstanceOf[String],
+      p.productElement(3).asInstanceOf[Int],
+      p.productElement(4).asInstanceOf[CacheStrategy[F]],
+    ): @nowarn("msg=deprecated")
+  }
```
armanbilge, line 24: Thanks for the nudge in the right direction. I hope you're happy now 😆

armanbilge:
Cool cool. We've moved on from mima to explicit dependencies which I guess don't work on Scala.js. There's an unmerged fix upstream.
- https://github.com/cb372/sbt-explicit-dependencies/pull/100

server/shared/src/main/scala-3/org/http4s/server/staticcontent/FileServiceConfigPlatform.scala
```diff
@@ -0,0 +1,31 @@
+package org.http4s
+package server
+package staticcontent
+
+import scala.annotation.nowarn
+import scala.deriving.Mirror
+
+private[staticcontent] trait FileServiceConfigCompanionPlatform extends Mirror.Product {
```
rossabaker, line 24: I usually call the version splits "Compat" instead of "Platform", which I guess only matters if there's ever a version and platform split together, but $DEITY forbid.  It's private and it's done and it can be ripped out on main so I wouldn't spend time on it.  Just sharing for next time.

blaze-client/src/main/scala/org/http4s/blaze/client/BlazeClient.scala
```diff
@@ -149,4 +155,14 @@ object BlazeClient {
         }
       }
     }
+
+  @deprecated("Preserved for binary compatibility", "0.22.9")
```
danicheg, line 159:
```suggestion
  @deprecated("Preserved for binary compatibility", "0.23.8")
```

blaze-client/src/main/scala/org/http4s/blaze/client/BlazeClientBuilder.scala
```diff
@@ -86,13 +90,65 @@ sealed abstract class BlazeClientBuilder[F[_]] private (
...
+    val retries: Int,
+    val maxIdleDuration: Duration,
 )(implicit protected val F: Async[F])
     extends BlazeBackendBuilder[Client[F]]
     with BackendBuilder[F, Client[F]] {
   type Self = BlazeClientBuilder[F]
 
   protected final val logger = getLogger(this.getClass)
 
+  @deprecated("Preserved for binary compatibility", "0.22.9")
```
danicheg, line 102:
```suggestion
  @deprecated("Preserved for binary compatibility", "0.23.8")
```

blaze-client/src/main/scala/org/http4s/blaze/client/ConnectionManager.scala
```diff
@@ -97,6 +98,28 @@ private object ConnectionManager {
         requestTimeout,
         semaphore,
         executionContext,
+        maxIdleDuration,
       )
     }
+
+  @deprecated("Preserved for binary compatibility", "0.22.9")
```
danicheg, line 105:
```suggestion
  @deprecated("Preserved for binary compatibility", "0.23.8")
```

blaze-client/src/main/scala/org/http4s/blaze/client/PoolManager.scala
```diff
@@ -45,8 +48,35 @@ private final class PoolManager[F[_], A <: Connection[F]](
     requestTimeout: Duration,
     semaphore: Semaphore[F],
     implicit private val executionContext: ExecutionContext,
+    maxIdleDuration: Duration,
 )(implicit F: Async[F])
     extends ConnectionManager.Stateful[F, A] { self =>
+
+  @deprecated("Preserved for binary compatibility", "0.22.9")
```
danicheg, line 55:
```suggestion
  @deprecated("Preserved for binary compatibility", "0.23.8")
```

server/shared/src/main/scala-3/org/http4s/server/staticcontent/FileServiceConfigPlatform.scala
armanbilge: Ah, thanks, I was wondering about this. Other projects don't distinguish and I have gotten bitten by this before at least once or twice.

rossabaker:
Looks like unused compile dependencies tests and Scala.JS are a bad fit.

rossabaker:
I tried to set the unused compile dependency checks in sbt-http4s-org to run only on the JVM, but `.jvmSettings` requires a project, and I couldn't quickly untangle the magic that the crossproject plugin is doing.  I'll take a look at this later unless someone knows the prior art for a platform-specific build setting in a plugin.

rossabaker:
I think I figured it out.

armanbilge:
> I'll take a look at this later unless someone knows the prior art for a platform-specific build setting in a plugin.

AFAIK there's no easy way to do this. Unlike Scala version which is a setting on every project, cross-projects are really just sets of (independent) projects that happen to share the same sources. I would like to find a way to make this easier and put it in sbt-typelevel but for now we'll have to do something manual and ugly.

armanbilge (approved):
Hallelujah.

PR #5927, opened 2022-01-22
Run jstack in background to troubleshoot

rossabaker:
Probably won't ever merge this one, but it's a place to chase #5844 .


armanbilge:
FYI apparently there's this thing, which may be helpful as well. Never tried it.
```scala
  val Tmate: WorkflowStep =
    Use(UseRef.Public("mxschmitt", "action-tmate", "v2"), name = Some("Setup tmate session"))
```
https://github.com/typelevel/sbt-typelevel/blob/2259d477203eac310538e9d6629affa8bfb1c9a7/github-actions/src/main/scala/org/typelevel/sbt/gha/WorkflowStep.scala#L57

rossabaker:
I would not recommend using that without https://github.com/mxschmitt/action-tmate#use-registered-public-ssh-keys.

rossabaker:
Didn't do what I want, I guess because pgrep was empty on the first run, and it broke the loop.  The tmate is probably better, once properly configured.  I'll try that tomorrow.

rossabaker:
We are hanging on line 128: https://github.com/http4s/http4s/blob/ba25a1cf1a8dc2dd81dc50cfb0731bf4c1f23f5b/servlet/src/test/scala/org/http4s/servlet/RouterInServletSuite.scala#L124-L130

rossabaker:
And there are five threads wedged in the Jetty servlet scaffold that test starts.

armanbilge:
I think f523220 has sucked up all of our CI runners.

rossabaker:
I was actively using one and burned it because I don't know how to tmux and am now using a second, but I don't need most of them.  I'll cancel a few.

rossabaker:
I can't cancel just one.  I did cancel one of the two builds, which should free up some.  I'm trying to get profiles of other blocking calls, so I'm keeping one running.

rossabaker:
I think this one has run its course, but it's a good trick to remember.

PR #5928, opened 2022-01-22
Non-deadlocking RouterInServletSuite

rossabaker:
This could be the villain in #5844.

* That `.get` in the suite should definitely be blocking
* The servlet pool is wedged, so let's use the AsyncHttp4sServlet there, too.
* I don't know if anyone has thought about what CE3 blocking means for servlets. 🤔


rossabaker:
Related conversation in #5927.

armanbilge (approved):

armanbilge:
Looks promising.

PR #5938, opened 2022-01-23
Release v0.22.9

rossabaker:


rossabaker:
bc741cd is what we got for free.  Recording it so we can think about improvements.

website/src/hugo/content/changelog.md
```diff
@@ -8,6 +8,98 @@
 Maintenance branches are merged before each new release. This change log is
 ordered chronologically, so each release contains all changes described below it.
 
+# v0.22.9 (2022-01-23)
+
+## What's Changed
+### http4s-core
```
armanbilge, line 14: Unfortunately we have to change all these `###` to bullet points, otherwise laika balks IIRC.

rossabaker: Oh, right.  Gack.

rossabaker:
I immediately see some that are appreciated but don't merit highlighting, and some that are "behind the scenes" that are worth noting.  Maybe I should adjust some labels and regenerate.

website/src/hugo/content/changelog.md
armanbilge: This should probably be fixed upstream. It should be able to handle these duplicates by appending a counter to the anchor.

rossabaker:
Some of them get labeled with a module, but don't end up being publicly interesting.  But "behind the scenes" is a fallback category, so I'm not sure how else to label them other than to unlabel them.

rossabaker:
Ah, `changelog.categories[*].exclude.labels`. :thinking:

armanbilge:
We could make behind-the-scenes a label, and collect those first (and then cut-and-paste to the bottom). Then do away with the fallback altogether. Idk.

armanbilge:
Oh, nice!

rossabaker:
We could get smarter about tagging dependencies and PRs with modules, which would help a lot.

website/src/hugo/content/changelog.md
rossabaker: Well, except the anchor's relative position is going to change over time, making it a not very useful anchor.

armanbilge: Fair point. What it really needs to respect is the nesting I suppose. Asking that to be unique seems reasonable.

rossabaker:
We get a lot of superseded PRs, which we could handle with a superseded label.  I had to go back and relabel a lot.  We need a Laika rewrite.  (I am Emacs, hear me roar.). But dd7c663 is automated and not too shabby!

rossabaker:
Nope, still bad.  My labels aren't always sticking.

website/src/hugo/content/changelog.md
```diff
@@ -8,6 +8,118 @@
 Maintenance branches are merged before each new release. This change log is
 ordered chronologically, so each release contains all changes described below it.
 
+# v0.22.9 (2022-01-23)
+
+This release is binary compatible with 0.22.x series.  The jawn upgrade mitigates [CVE-2022-21653](https://github.com/typelevel/jawn/security/advisories/GHSA-vc89-hccf-rq55) out of the box for `http4s-play-json`.
+
+## What's Changed
```
armanbilge, line 15: laika won't like this I think

website/src/hugo/content/changelog.md
```diff
@@ -8,6 +8,118 @@
...
+    * Update scalafmt-core to 3.3.2 in series/0.22 by @scala-steward in https://github.com/http4s/http4s/pull/5910
+    * Enable snapshots for 0.22 by @armanbilge in https://github.com/http4s/http4s/pull/5908
+    * Revert "Enable snapshots for 0.22" (#5908) by @rossabaker in https://github.com/http4s/http4s/pull/5920
+    * Add `scalafixInternalRules` to root project by @armanbilge in https://github.com/http4s/http4s/pull/5918
+    * Update sbt-http4s-org to 0.10.1 in series/0.22 by @scala-steward in https://github.com/http4s/http4s/pull/5922
+    * Update scalafmt-core to 3.3.3 in series/0.22 by @scala-steward in https://github.com/http4s/http4s/pull/5931
+    * Update sbt-http4s-org to 0.11.0 in series/0.22 by @scala-steward in https://github.com/http4s/http4s/pull/5935
+    * remove empty excludes from release notes config by @armanbilge in https://github.com/http4s/http4s/pull/5937
+
+## New Contributors
```
armanbilge, line 117: also this one :(

rossabaker:
The exclusion isn't doing what it says on the tin, and I can't find the source anywhere, nor anything to validate my YAML against.  The data is there in the labels, so maybe it's time to try one of the other tools.

rossabaker:
I don't know what's wrong with my category exclusions.  [release-drafter](https://github.com/release-drafter/release-drafter#categorize-pull-requests) looks promising, but I don't see how to exlcude from a category.

Maybe these behind-the-scenes ones don't belong in the changelog at all.  They're relevant to us, but they might just be noise to end users.

armanbilge:
Oh, I know what the problem is. You need to merge the release notes config first I think.

rossabaker:
Oh... yeah, that makes sense.  It would read it from the target branch, not this PR.  I did find [an example](https://github.com/equinor/ert/blob/c09bf02a17e23fa5bdfa8ab1c6ae4e65ddaa9efb/.github/release.yml) of someone doing it like us, so we're not totally mad.

danicheg (approved):

PR #5940, opened 2022-01-24
Merge from 0.22

rossabaker:
This one's pretty trivial.  It will let us start with the release notes.

PR #5941, opened 2022-01-24
Add module labeling to mergify config

armanbilge:
Part of https://github.com/http4s/http4s/pull/5940#issuecomment-1019702909. Targeting main because that's where mergify reads the config from.

Getting auto-labeling for dependencies will be a little trickier.


rossabaker (approved):
This is nice.  It looks like the build hung because it doesn't have the blocking fixes.

PR #5944, opened 2022-01-25
Replace our TestExecutionContext

rossabaker:
Ours was written in olden times to get daemon threads.  The IORuntime.compute pool is also daemon threads, and implements a fancy work stealing algorithm.  This is the one that gets used in practice, and is a more honest test.  It also touches the ClientTimeoutSuite that's acting up.  Let's see if it helps!

Also removes our `TestIORuntime` and `TestScheduler`.  If the global is good enough for munit-cats-effect, maybe it's good enough for us.

### DIFF
--- a/testing/js/src/test/scala/org/http4s/Http4sSuitePlatform.scala
+++ b/testing/js/src/test/scala/org/http4s/Http4sSuitePlatform.scala
@@ -16,8 +16,6 @@
 
 package org.http4s
 
-import cats.effect.unsafe.IORuntime
-
 import scala.scalajs.js
 import scala.util.Try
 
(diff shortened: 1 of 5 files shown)


testing/js/src/test/scala/org/http4s/Http4sSuitePlatform.scala
```diff
@@ -24,5 +24,3 @@ trait Http4sSuitePlatform { this: Http4sSuite =>
   override def munitFlakyOK =
     Try(js.Dynamic.global.process.env.CI).toOption.filterNot(js.isUndefined).isDefined
 }
```
armanbilge, line 26: Just realized, once `Env` gets released in CE 3.4 we'll be able to deplatform this :)

rossabaker:
The three commits are getting us a triple test of the first one, which I suspect is the most helpful one.  Looks promising so far!

rossabaker:
Well, 74b7671 failed on better known troublemaker. :weary:

armanbilge (approved):
Still, seems like the `ClientTimeoutSuite` flake is quashed for now.

rossabaker:
It may be luck, but the code is simpler. I'm going to bed, but I'll cut the release tomorrow.

danicheg (approved):

rossabaker:
Well, it was fun while it lasted: https://github.com/http4s/http4s/runs/4936918600?check_suite_focus=true#step:10:5059

PR #5955, opened 2022-01-26
Clean up Throwables in HTTP/2

rossabaker:
Throwables that are neither Exceptions nor Errors are awkward.  Use standard types.

### DIFF
--- a/ember-core/shared/src/main/scala/org/http4s/ember/core/h2/H2Client.scala
+++ b/ember-core/shared/src/main/scala/org/http4s/ember/core/h2/H2Client.scala
@@ -146,7 +146,7 @@ private[ember] class H2Client[F[_]: Async](
             case Some("http/1.1") => Resource.pure[F, SocketType](Http1)
             case Some(other) =>
               Resource.raiseError[F, SocketType, Throwable](
-                new ProtocolException(s"Unknown protocol")
+                new ProtocolException("Unknown protocol")
               )
             case None => Resource.pure[F, SocketType](Http1)
           }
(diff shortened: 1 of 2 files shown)


ember-core/shared/src/main/scala/org/http4s/ember/core/h2/H2Client.scala
```diff
@@ -146,7 +146,7 @@ private[ember] class H2Client[F[_]: Async](
             case Some("http/1.1") => Resource.pure[F, SocketType](Http1)
             case Some(other) =>
               Resource.raiseError[F, SocketType, Throwable](
-                new Throwable("Unknown Protocol Received")
+                new ProtocolException(s"Unknown protocol")
```
ChristopherDavenport, line 149:
```suggestion
                new ProtocolException("Unknown protocol")
```

Don't need the string interpolator without that other.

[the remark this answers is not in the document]
ChristopherDavenport: Alright, this can't happen. I just checked. Since we are the client we set the input alpn set, so it can only be one of those two in the output alpn set according to spec. 

`None` is a possibility, but this should never happen with a well-behaved server.

rossabaker: Oops.  Caught me adding it and then having second thoughts.

rossabaker:
This has the same old unrelated failure, which I hope #5385  helps with.

ChristopherDavenport (approved):

PR #5956, opened 2022-01-27
Quash Scala 3 unsafe init warnings

armanbilge:
I'm working on enrolling http4s in the Scala 3 community build in https://github.com/lampepfl/dotty/pull/14071. It seems they are compiling all projects with the [`-Ysafe-init`](https://docs.scala-lang.org/scala3/reference/other-new-features/safe-initialization.html) flag which http4s is currently failing (see errors below).

Here I've made changes to address the warnings.
1. How do we feel about these changes? It's not impossible that the warnings here are due to a bug/shortcoming in the compiler; I haven't fully wrapped my head around what the issue is in http4s.
2. Should we compile our Scala 3 sources with this (experimental) flag?
3. Should we run these changes up through 0.22?

```
[warn] -- Warning: /workspace/http4s/core/shared/src/main/scala/org/http4s/internal/CharPredicate.scala:104:33 
[warn] 104 |  val LowerAlpha = CharPredicate('a' to 'z')
[warn]     |                                 ^^^^^^^^^^
[warn]     |Cannot prove that the value is fully-initialized. May only use initialized value as method arguments.
[warn]     |
[warn]     |The unsafe promotion may cause the following problem:
[warn]     |Cannot prove that the value is fully initialized. May only use initialized value as method arguments.
[warn] -- Warning: /workspace/http4s/core/shared/src/main/scala/org/http4s/internal/CharPredicate.scala:105:33 
[warn] 105 |  val UpperAlpha = CharPredicate('A' to 'Z')
[warn]     |                                 ^^^^^^^^^^
[warn]     |Cannot prove that the value is fully-initialized. May only use initialized value as method arguments.
[warn]     |
[warn]     |The unsafe promotion may cause the following problem:
[warn]     |Cannot prove that the value is fully initialized. May only use initialized value as method arguments.
[warn] -- Warning: /workspace/http4s/core/shared/src/main/scala/org/http4s/internal/CharPredicate.scala:107:28 
[warn] 107 |  val Digit = CharPredicate('0' to '9')
[warn]     |                            ^^^^^^^^^^
[warn]     |Cannot prove that the value is fully-initialized. May only use initialized value as method arguments.
[warn]     |
[warn]     |The unsafe promotion may cause the following problem:
[warn]     |Cannot prove that the value is fully initialized. May only use initialized value as method arguments.
[warn] -- Warning: /workspace/http4s/core/shared/src/main/scala/org/http4s/internal/CharPredicate.scala:108:30 
[warn] 108 |  val Digit19 = CharPredicate('1' to '9')
[warn]     |                              ^^^^^^^^^^
[warn]     |Cannot prove that the value is fully-initialized. May only use initialized value as method arguments.
[warn]     |
[warn]     |The unsafe promotion may cause the following problem:
[warn]     |Cannot prove that the value is fully initialized. May only use initialized value as method arguments.
[warn] -- Warning: /workspace/http4s/core/shared/src/main/scala/org/http4s/internal/CharPredicate.scala:110:37 
[warn] 110 |  val LowerHexLetter = CharPredicate('a' to 'f')
[warn]     |                                     ^^^^^^^^^^
[warn]     |Cannot prove that the value is fully-initialized. May only use initialized value as method arguments.
[warn]     |
[warn]     |The unsafe promotion may cause the following problem:
[warn]     |Cannot prove that the value is fully initialized. May only use initialized value as method arguments.
[warn] -- Warning: /workspace/http4s/core/shared/src/main/scala/org/http4s/internal/CharPredicate.scala:111:37 
[warn] 111 |  val UpperHexLetter = CharPredicate('A' to 'F')
[warn]     |                                     ^^^^^^^^^^
[warn]     |Cannot prove that the value is fully-initialized. May only use initialized value as method arguments.
[warn]     |
[warn]     |The unsafe promotion may cause the following problem:
[warn]     |Cannot prove that the value is fully initialized. May only use initialized value as method arguments.
[warn] -- Warning: /workspace/http4s/core/shared/src/main/scala/org/http4s/internal/CharPredicate.scala:114:30 
[warn] 114 |  val Visible = CharPredicate('\u0021' to '\u007e')
[warn]     |                              ^^^^^^^^^^^^^^^^^^^^
[warn]     |Cannot prove that the value is fully-initialized. May only use initialized value as method arguments.
[warn]     |
[warn]     |The unsafe promotion may cause the following problem:
[warn]     |Cannot prove that the value is fully initialized. May only use initialized value as method arguments.
[warn] -- Warning: /workspace/http4s/core/shared/src/main/scala/org/http4s/headers/Server.scala:25:15 
[warn] 25 |  override val name = super.name
[warn]    |               ^
[warn]    |Access non-initialized value name. Calling trace:
[warn]    | -> object Server extends HeaderCompanion[Server]("Server") {        [ Server.scala:23 ]
[warn]    |  -> private[headers] abstract class HeaderCompanion[A](_name: String) {     [ HeaderCompanion.scala:24 ]
[warn]    |   -> private val invalidHeader = s"Invalid $name header"    [ HeaderCompanion.scala:32 ]
[warn] -- Warning: /workspace/http4s/core/shared/src/main/scala/org/http4s/util/Renderable.scala:126:51 
[warn] 126 |  def renderString: String = Renderer.renderString(this)
[warn]     |                                                   ^^^^
[warn]     |Cannot prove that the value is fully initialized. May only use initialized value as arguments. Calling trace:
[warn]     | -> override val renderString: String = super.renderString  [ Uri.scala:352 ]
[warn] 9 warnings found
```

### DIFF
--- a/core/shared/src/main/scala/org/http4s/Uri.scala
+++ b/core/shared/src/main/scala/org/http4s/Uri.scala
@@ -349,7 +349,7 @@ object Uri extends UriPlatform {
       if (endsWithSlash && segments.nonEmpty) writer << "/" else writer
     }
 
-    override val renderString: String = super.renderString
+    override lazy val renderString: String = super.renderString
     override def toString: String = renderString
 
     /** This is an alias to [[addSegment]]
--- a/core/shared/src/main/scala/org/http4s/headers/Server.scala
+++ b/core/shared/src/main/scala/org/http4s/headers/Server.scala
@@ -22,7 +22,7 @@ import org.http4s.util.Writer
 
 object Server extends HeaderCompanion[Server]("Server") {
 
-  override val name = super.name
+  override lazy val name = super.name
 
   def apply(id: ProductId, tail: ProductIdOrComment*): Server =
     apply(id, tail.toList)
--- a/core/shared/src/main/scala/org/http4s/internal/CharPredicate.scala
+++ b/core/shared/src/main/scala/org/http4s/internal/CharPredicate.scala
@@ -101,18 +101,18 @@ sealed abstract class CharPredicate extends (Char => Boolean) {
 object CharPredicate {
   val Empty: CharPredicate = MaskBased(0L, 0L)
   val All: CharPredicate = from(_ => true)
-  val LowerAlpha = CharPredicate('a' to 'z')
-  val UpperAlpha = CharPredicate('A' to 'Z')
-  val Alpha = LowerAlpha ++ UpperAlpha
-  val Digit = CharPredicate('0' to '9')
-  val Digit19 = CharPredicate('1' to '9')
-  val AlphaNum = Alpha ++ Digit
-  val LowerHexLetter = CharPredicate('a' to 'f')
-  val UpperHexLetter = CharPredicate('A' to 'F')
-  val HexLetter = LowerHexLetter ++ UpperHexLetter
-  val HexDigit = Digit ++ HexLetter
-  val Visible = CharPredicate('\u0021' to '\u007e')
-  val Printable = Visible ++ ' '
+  lazy val LowerAlpha = CharPredicate('a' to 'z')
+  lazy val UpperAlpha = CharPredicate('A' to 'Z')
+  lazy val Alpha = LowerAlpha ++ UpperAlpha
+  lazy val Digit = CharPredicate('0' to '9')
+  lazy val Digit19 = CharPredicate('1' to '9')
+  lazy val AlphaNum = Alpha ++ Digit
+  lazy val LowerHexLetter = CharPredicate('a' to 'f')
+  lazy val UpperHexLetter = CharPredicate('A' to 'F')
+  lazy val HexLetter = LowerHexLetter ++ UpperHexLetter
+  lazy val HexDigit = Digit ++ HexLetter
+  lazy val Visible = CharPredicate('\u0021' to '\u007e')
+  lazy val Printable = Visible ++ ' '
 
   def from(predicate: Char => Boolean): CharPredicate =
     predicate match {
--- a/core/shared/src/main/scala/org/http4s/internal/CharPredicate.scala
+++ b/core/shared/src/main/scala/org/http4s/internal/CharPredicate.scala
@@ -101,18 +101,18 @@ sealed abstract class CharPredicate extends (Char => Boolean) {
 object CharPredicate {
   val Empty: CharPredicate = MaskBased(0L, 0L)
   val All: CharPredicate = from(_ => true)
-  lazy val LowerAlpha = CharPredicate('a' to 'z')
-  lazy val UpperAlpha = CharPredicate('A' to 'Z')
-  lazy val Alpha = LowerAlpha ++ UpperAlpha
-  lazy val Digit = CharPredicate('0' to '9')
-  lazy val Digit19 = CharPredicate('1' to '9')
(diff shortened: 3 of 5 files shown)


armanbilge:
Ok, it seems the problem relates to the nesting of `ApplyMagnet`. Unnesting it from `CharPredicate` makes the warnings go away without resorting to lazy vals there.

core/shared/src/main/scala/org/http4s/Uri.scala
```diff
@@ -349,7 +349,7 @@ object Uri extends UriPlatform {
       if (endsWithSlash && segments.nonEmpty) writer << "/" else writer
     }
 
-    override val renderString: String = super.renderString
+    override lazy val renderString: String = super.renderString
```
rossabaker, line 352: What purpose is an override to super serving?

core/shared/src/main/scala/org/http4s/headers/Server.scala
```diff
@@ -22,7 +22,7 @@ import org.http4s.util.Writer
 
 object Server extends HeaderCompanion[Server]("Server") {
 
-  override val name = super.name
+  override lazy val name = super.name
```
rossabaker, line 25: What purpose is an override to super serving?

core/shared/src/main/scala/org/http4s/Uri.scala
armanbilge: Made a `def` a `val`? (this seems moot). And also once it's there, for binary compatibility, although that's not relevant on main.

rossabaker: It prevents it from creating a StringWriter on repeated references.  That whole thing deserves a rethink at some point.

core/shared/src/main/scala/org/http4s/internal/CharPredicate.scala
```diff
@@ -347,3 +325,26 @@ object CharPredicate {
 
   // scalafix:on
 }
+
+class ApplyMagnet(val predicate: CharPredicate)
```
rossabaker, line 329: This whole machinery was too cute anyway, in my opinion.  But this is fine... it's just private, and not used that widely.

rossabaker (commented):
Turning on stricter flags in a later series is a good way to get frustrated at merge time.  If we need to enforce this to get into the community build, it makes sense to me to start lower.

rossabaker:
So do we want to transplant this onto 0.22 then, just so these warnings don't bite us at the end of the train?  The code is not effectful, so it should be an easy merge.

armanbilge:
I'm working on this right now actually. I started with 0.22 but I think 0.23 may make more sense, because the 3.0.2 compiler on 0.22 is giving different warnings, and we don't have fatal warnings on 0.22 anyway. WDYT?

armanbilge:
Ok, I was able to get http4s green in the community build without these changes. We can close, unless we still want to pursue (some of) these changes because they're probably a Good Idea anyway?

rossabaker:
Doing it on 0.23 would add unnecessary divergences between 0.22 and 0.23, but I think it's time to resume the EOL talk about 0.22.  I'm fine if you want to start on 0.23, or cherry-pick the changesets back to 0.22 for one last hurrah if it's easy.  I trust your judgment.

armanbilge:
We got into the community build without these changes after all. They might still be good to look into, but I'm not enthusiastic to chase it further right now.

PR #4932, opened 2021-06-19
WIP Ember Docs

valencik:
This PR is just https://github.com/http4s/http4s/pull/4279 rebased on https://github.com/http4s/http4s/pull/4931, checking CI run...

PR #4279, opened 2021-01-27
Ember Use in Docs Over Blaze

ChristopherDavenport:
Working on the g8 template now

PR #5970, opened 2022-01-29
Use ember in docs

valencik:
This PR updates all the mdoc checked docs to use ember instead of blaze.

This is a continuation of https://github.com/http4s/http4s/pull/4932 and https://github.com/http4s/http4s/pull/4279.
Everything is rebased on latest `series/0.23`, but I can change the target branch if desired.

I have introduced three changes that are worth calling out as they may be controversial.

### Sleeping after ember server start
Starting the ember server in mdoc now looks like:
```scala
val startAndSleep = server.use(_ => IO.never).start <* IO.sleep(2.seconds)
val fiber = startAndSleep.unsafeRunSync()
```
Without an `IO.sleep`, any attempt to connect to the ember server would fail with a `java.net.ConnectException: Connection refused` as the request would be attempted before the ember server was actually bound to the address.

### Adding logback to docs/mdoc
This was absolutely crucial for me to figure out what was going on with the connect refusal error above.
With logging enable I could see the `ConnectException` and then a second later the ember server log:
```
[info] [io-compute-15] INFO  o.h.e.s.EmberServerBuilderCompanionPlatform - Ember-Server service bound to address: [::]:8080
```
I think having the logs is worthwhile if someone is running `mdoc`, it gives us slightly more visibility into what is going on.
I stole the [`logback.xml` from the gitter8 template](https://github.com/http4s/http4s.g8/blob/0.23/src/main/g8/src/main/resources/logback.xml).

### Forking mdoc
This is really just the workaround described in https://github.com/scalameta/mdoc/issues/613 to avoid terminating sbt when mdoc fails.

### Ditching `nest` mdoc modifier
I've removed `nest` mdoc modifiers in client.md because I found they only added confusion.
I've held off removing `nest` from the other docs as it turned out to not be necessary, but I feel like it's still worthwhile.
I don't think it buys us much.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1030,6 +1030,7 @@ lazy val docs = http4sProject("docs")
   )
   .settings(docsProjectSettings)
   .settings(
+    run / fork := true,
     libraryDependencies ++= Seq(
       circeGeneric,
       circeLiteral,
--- a/build.sbt
+++ b/build.sbt
@@ -1094,8 +1094,6 @@ lazy val docs = http4sProject("docs")
     client.jvm,
     core.jvm,
     theDsl.jvm,
-    blazeServer,
-    blazeClient,
     emberServer.jvm,
     emberClient.jvm,
     circe.jvm,
--- a/build.sbt
+++ b/build.sbt
@@ -1035,6 +1035,7 @@ lazy val docs = http4sProject("docs")
       circeGeneric,
       circeLiteral,
       cryptobits,
+      logbackClassic % Runtime,
     ),
     description := "Documentation for http4s",
     startYear := Some(2013),
--- a/build.sbt
+++ b/build.sbt
@@ -1095,6 +1095,8 @@ lazy val docs = http4sProject("docs")
     theDsl.jvm,
     blazeServer,
     blazeClient,
+    emberServer.jvm,
+    emberClient.jvm,
     circe.jvm,
     dropwizardMetrics,
     prometheusMetrics,
--- a/docs/src/main/mdoc/README.md
+++ b/docs/src/main/mdoc/README.md
@@ -10,7 +10,7 @@ Getting started with http4s is easy.  Let's materialize an http4s
 skeleton project from its [giter8 template]:
 
 ```sbt
-$ sbt -sbt-version 1.3.12 new http4s/http4s.g8 -b 0.23
+$ sbt new http4s/http4s.g8 --branch 0.23
 ```
 
 Follow the prompts.  For every step along the way, a default value is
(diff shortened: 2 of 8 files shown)


rossabaker:
Restarting, because I think it lost a race between 0.23.9 being tagged and it being available.

ChristopherDavenport (approved):

rossabaker (approved):
> Without an IO.sleep, any attempt to connect to the ember server would fail with a java.net.ConnectException: Connection refused as the request would be attempted before the ember server was actually bound to the address.

Wait, what?  I thought resource acquisition only happened when the port was bound and ready.  It's good to document things as they are, but this really ought to be cleaned up in the code.

> I've held off removing nest from the other docs as it turned out to not be necessary, but I feel like it's still worthwhile.

I agree, if we can get away with it.  I think those were added during the migration from tut, which was "implicitly nested" by default.  But that can be a separate PR.

valencik:
Thanks so much for reviewing :)

> > Without an IO.sleep, any attempt to connect to the ember server would fail with a java.net.ConnectException: Connection refused as the request would be attempted before the ember server was actually bound to the address.
> 
> Wait, what? I thought resource acquisition only happened when the port was bound and ready. It's good to document things as they are, but this really ought to be cleaned up in the code.

I'm not sure I follow "I thought resource acquisition only happened when the port was bound and ready." specifically.
But, to be clear, I don't know what *is* happening, or what *should* be happening :grimacing:.

Also, when you say this really ought to be cleaned up in the code, do you mean the EmberServer code or the mdoc documentation code in client.md? (Just trying to understand what I might be able to help with)

Either way, if we are keeping the `IO.sleep` for now, I'm happy to add in a comment explaining that it is only necessary because we are in mdoc.

Finally, regarding `nest`, yeah, I can open up a separate PR after this merges and clean up the rest.

armanbilge:
I think the reason you need the sleep here is because of the `.start`: it's a fire-and-forget, so we lose the usual backpressure signal.

Since you are not closing the server, you could use `server.allocated.void` to achieve the same effect.

armanbilge (approved):
👍 on green.

rossabaker:
Oh, I was having a dumb.  Yes, my point was:

```scala
builder.use { server =>
  // server should be listening here
}
```

But we can't `use` the builder in mdoc.  The problem is that good resource practice in an application is not amenable to a REPL tutorial.  What's here is right, but the `.allocated` is usually bad practice, and the `_2` is ugly.  We could add a `.background` function to the builders that returns an `F[F[Unit]]` to shut down, but that doesn't get rid of the `unsafeRunSync()`.

I think this is an improvement as is, so :+1:, but I regret that our tutorials are teaching people bad habits because a REPL is not an IOApp.  I'd like to think about how we can do better here.

armanbilge:
I wonder if `background` can float up to CE `Resource` itself. We already have `use_` and `useForever` so seems like it might fit in..

PR #5974, opened 2022-01-30
Improve the `WSClient` API

armanbilge:
I'm working on a browser implementation in https://github.com/http4s/http4s-dom/pull/93 and these are a couple things that tripped me up.

1. Use `DeferredSource` instead of `Deferred` for `closeFrame`. This prevents external completion of it :)
2. Remove `final` modifiers. Not sure how necessary they are and I'd like to override `receiveStream` with a more efficient implementation in the browser.

H/t Ross for making sure we kept these package-private 😅

### DIFF
--- a/client/shared/src/main/scala/org/http4s/client/websocket/WSClient.scala
+++ b/client/shared/src/main/scala/org/http4s/client/websocket/WSClient.scala
@@ -21,6 +21,7 @@ import cats.data.Chain
 import cats.data.OptionT
 import cats.effect._
 import cats.effect.kernel.Deferred
+import cats.effect.kernel.DeferredSource
 import cats.effect.kernel.Ref
 import cats.implicits._
 import fs2.Pipe
@@ -109,7 +110,7 @@ private[http4s] trait WSConnection[F[_]] {
   def sendMany[G[_]: Foldable, A <: WSFrame](wsfs: G[A]): F[Unit]
 
   /** A `Pipe` which sends websocket frames and emits a `()` for each chunk sent. */
-  final def sendPipe: Pipe[F, WSFrame, Unit] = _.chunks.evalMap(sendMany(_))
+  def sendPipe: Pipe[F, WSFrame, Unit] = _.chunks.evalMap(sendMany(_))
 
   /** Wait for a single websocket frame to be received. Returns `None` if the receiving side is
     * closed.
@@ -117,7 +118,7 @@ private[http4s] trait WSConnection[F[_]] {
   def receive: F[Option[WSFrame]]
 
   /** A stream of the incoming websocket frames. */
-  final def receiveStream: Stream[F, WSFrame] = Stream.repeatEval(receive).unNoneTerminate
+  def receiveStream: Stream[F, WSFrame] = Stream.repeatEval(receive).unNoneTerminate
 
   /** The negotiated subprotocol, if any. */
   def subprotocol: Option[String]
@@ -132,7 +133,7 @@ private[http4s] trait WSConnectionHighLevel[F[_]] {
   def sendMany[G[_]: Foldable, A <: WSDataFrame](wsfs: G[A]): F[Unit]
 
   /** A `Pipe` which sends websocket frames and emits a `()` for each chunk sent. */
-  final def sendPipe: Pipe[F, WSDataFrame, Unit] = _.chunks.evalMap(sendMany(_))
+  def sendPipe: Pipe[F, WSDataFrame, Unit] = _.chunks.evalMap(sendMany(_))
 
   /** Send a Close frame. The sending side of this connection will be closed. */
   def sendClose(reason: String = ""): F[Unit]
@@ -143,13 +144,13 @@ private[http4s] trait WSConnectionHighLevel[F[_]] {
   def receive: F[Option[WSDataFrame]]
 
   /** A stream of the incoming websocket frames. */
-  final def receiveStream: Stream[F, WSDataFrame] = Stream.repeatEval(receive).unNoneTerminate
+  def receiveStream: Stream[F, WSDataFrame] = Stream.repeatEval(receive).unNoneTerminate
 
   /** The negotiated subprotocol, if any. */
   def subprocotol: Option[String]
 
   /** The close frame, if available. */
-  def closeFrame: Deferred[F, WSFrame.Close]
+  def closeFrame: DeferredSource[F, WSFrame.Close]
 }
 
 /** A websocket client capable of establishing [[WSClientHighLevel#connectHighLevel "high level" connections]].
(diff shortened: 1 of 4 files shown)


amesgen (approved):
- I think added `final` everywhere such that the "derived" methods like `receiveStream` can't become out of sync when overriding an existing `WSClient`, but that is not a big danger, and being able to implement some of these methods more efficiently is more important.
 - TIL about `DeferredSource`, nice addition in cats-effect-3!

armanbilge:
Oh, we should also add a `translate` method.

armanbilge:
Most of these changes should be run up through 0.22 actually. Especially the typo 😬

client/shared/src/main/scala/org/http4s/client/websocket/WSClient.scala
```diff
@@ -219,8 +259,8 @@ private[http4s] object WSClient {
               }
             defrag(Chain.empty, ByteVector.empty).value
           }
-          override def subprocotol: Option[String] = conn.subprotocol
-          override def closeFrame: Deferred[F, WSFrame.Close] = recvCloseFrame
+          override def subprotocol: Option[String] = conn.subprotocol
```
amesgen, line 262: Wait, why does this not throw an error? I feel like I am missing sth here, Scala 2.13.8 rejects this (without any extra flags):

```scala
trait A { def a: Int = 1 }
new A { override def b: Int = 2 }

```

armanbilge: What error are you expecting it to throw?

armanbilge:
Ok, here's another annoying one: the `sendClose` method in the high-level API is intended to send a close-frame but not actually close the connection. I've now realized that browsers are incapable of this subtlety: your only option is to completely close the connection.

What if we remove the `sendClose` method from the high-level API, could that be ok? I hate if we keep catering to the browser as the lowest-common-denominator :(

client/shared/src/main/scala/org/http4s/client/websocket/WSClient.scala
amesgen: Ahhh, I somehow missed that "subprotocol" was misspelled in both the definition and here, somehow I assumed it was only case here. That left me very confused why it did not throw an "overrides nothing" error 😆

amesgen:
:+1: on removing `sendClose` there, I am skeptical that anyone will use this even via the low-level interface 😆

armanbilge:
Thanks! Actually I noticed it is being used here but I think it is unnecessary.
https://github.com/gemini-hlsw/clue/blob/04260d0535c8b4ed7b9b928d53fee775cd658bf1/http4s-jdk/src/main/scala/clue/http4sjdk/Http4sJDKWSBackend.scala#L55

client/shared/src/main/scala/org/http4s/client/websocket/WSClient.scala
```diff
@@ -179,21 +214,18 @@ private[http4s] object WSClient {
       override def connectHighLevel(request: WSRequest) =
         for {
           recvCloseFrame <- Resource.eval(Deferred[F, WSFrame.Close])
-          outputOpen <- Resource.eval(Ref[F].of(false))
```
armanbilge, line 182: Was this a bug? Seems that it was never set to `true`.

client/shared/src/main/scala/org/http4s/client/websocket/WSClient.scala
```diff
@@ -179,21 +214,18 @@ private[http4s] object WSClient {
...
-          override def sendClose(reason: String) =
-            conn.send(WSFrame.Close(1000, reason)) *> outputOpen.set(false)
           override def receive: F[Option[WSDataFrame]] = {
             def receiveDataFrame: OptionT[F, WSDataFrame] =
               OptionT(conn.receive).flatMap { wsf =>
                 OptionT.liftF(wsf match {
                   case WSFrame.Ping(data) if respondToPings => conn.send(WSFrame.Pong(data))
                   case wsf: WSFrame.Close =>
-                    recvCloseFrame.complete(wsf) *> outputOpen.get.flatMap(conn.send(wsf).whenA(_))
+                    recvCloseFrame.complete(wsf) *> conn.send(wsf)
```
armanbilge, line 228: Please check this one, I was confused what the right thing to do here is b/c of https://github.com/http4s/http4s/pull/5974#discussion_r795245627.

armanbilge: Actually, I wonder if the `whenA` was also a mistake, related to:
- https://github.com/typelevel/cats/pull/3893.

armanbilge:
Ok, I've managed to implement `WSClient` for the browser in https://github.com/http4s/http4s-dom/pull/93 :) still a couple funky things with tests, but it works against a live server so that's something.

client/shared/src/main/scala/org/http4s/client/websocket/WSClient.scala
amesgen: Oh, yeah, that is/was a bug. The idea was track whether we did send a close frame yet was to prevent the following situation, as it would result in an exception: We send a close frame via `sendClose`, then we receive a close frame, and then send another one in response to that. The current faulty implementation indeed does prevent this exception, but ofc simply never replies with a close frame...

client/shared/src/main/scala/org/http4s/client/websocket/WSClient.scala
amesgen: It is safe to remove `outputOpen` now that it is no longer possible to manually send a close frame before the resource handler.

armanbilge:
I created an 0.22 backport in https://github.com/http4s/http4s/pull/5975.

client/shared/src/main/scala/org/http4s/client/websocket/WSClient.scala
amesgen: Side note: I agree that `ifA` was a suboptimal idea, but what is bad about `whenA`? The issue that "both branches get executed" doesn't apply here AFAICS.

armanbilge: I've been wondering about this too. Maybe I'm confused what `whenA` does, isn't `fa.whenA(cond)` syntax sugar for `F.pure(cond).ifA(fa, F.unit)`?

amesgen: No, `cond` is used "directly", such that it is not affected by the `ifA` issue: https://github.com/typelevel/cats/blob/815db7d7f5b60643953850b555bca8b38f6aadf9/core/src/main/scala/cats/Applicative.scala#L192-L193

armanbilge: Oh, I see. That make sense. Thanks! :)

amesgen: These combinators are surprisingly trick at times 😆

rossabaker:
@amesgen, are you still happy with this?

+@armanbilge: does anything else need to be backported to #5975?

armanbilge:
👍 from me, https://github.com/http4s/http4s/pull/5975 is up-to-date and pre-merged into this PR so there won't be conflicts to deal with.

amesgen (approved):
Looking forward to adopt this in jdk-http-client :+1:

PR #5987, opened 2022-02-02
Merge 0.22 to 0.23

armanbilge:
This was easy :)


.github/workflows/ci.yml
```diff
@@ -98,28 +99,34 @@ jobs:
...
 
       - name: Upload target directories
+        if: github.event_name != 'pull_request' && (startsWith(github.ref, 'refs/tags/v') || github.ref == 'refs/heads/main')
```
danicheg, line 126: A minor but truly grouchy nitpick: reference to the `main` branch is a little bit confusing within `series/0.22`/`series/0.23`. Would CI publish snapshots for the `series/0.22`/`series/0.23` at all?

armanbilge: I'm grouchy too. We can't publish snapshot until the website is fixed.
- https://github.com/http4s/http4s/pull/5920

rossabaker: Yeah, and those patches that we're talking about are going to break things again.  Getting that fixed needs to be a priority, but I doubt I'll have time to investigate before Friday.

rossabaker (approved):

PR #5991, opened 2022-02-02
Fix regression in routing dsl

### LINKED ISSUE #5959: Changes in route matching in 0.23.7 -> 0.23.8
Hi all,

I noticed some issues after upgrading to 0.23.8. Looking into it, it seems that request matching changed slightly between versions. Here is a small example:

```scala
#!/usr/bin/env amm

import $ivy.`org.http4s::http4s-dsl:0.23.7`
import $ivy.`org.http4s::http4s-blaze-server:0.23.7`
import $ivy.`ch.qos.logback:logback-classic:1.2.3`

import cats.effect._
import fs2.Stream
import org.http4s._
import org.http4s.server.Router
import org.http4s.blaze.server.BlazeServerBuilder
import org.http4s.implicits._

object TestApp extends IOApp.Simple {
  def routes: HttpRoutes[IO] = {
    val dsl = new Http4sDsl[IO] {}
    import dsl._

    HttpRoutes.of {
      case GET -> Root / id =>
        Ok(s"GET id: $id")

      case GET -> Root =>
        Ok("GET /")
    }
  }

  def app =
    Router("tasks" -> routes).orNotFound

  def server =
    BlazeServerBuilder[IO](ExecutionContext.global)
      .bindHttp(8888, "0.0.0.0")
      .withHttpApp(app)
      .serve

  val run =
    server.compile.drain
}

@main
def main(): Unit =
  TestApp.main(Array.empty)
```

When requesting `http://localhost:888/tasks`, it will match different routes:

```
# 0.23.7
❯ curl http://localhost:8888/tasks
GET /⏎

# 0.23.8
❯ curl http://localhost:8888/tasks
GET id: ⏎
```

With http4s 0.23.7 it matches the second case and with 0.23.8 it matches the first where `id` is empty (it's the same when appending a `/`).  Is this an expected change?

Thank you!

hamnis:
Path.Root has changed to be absolute and endsWithSlash. We need to take
this into account when routing. Added a check for non-root to avoid
matching on paths we shouldn't

Fixes #5959

### DIFF
--- a/dsl/src/main/scala/org/http4s/dsl/impl/Path.scala
+++ b/dsl/src/main/scala/org/http4s/dsl/impl/Path.scala
@@ -58,7 +58,7 @@ object ~ {
 
 object / {
   def unapply(path: Path): Option[(Path, String)] =
-    if (path.endsWithSlash)
+    if (path != Root && path.endsWithSlash)
       Some(path.dropEndsWithSlash -> "")
     else
       path.segments match {
--- a/server/src/test/scala/org/http4s/server/RouterSuite.scala
+++ b/server/src/test/scala/org/http4s/server/RouterSuite.scala
@@ -123,4 +123,19 @@ class RouterSuite extends Http4sSuite {
       .map(_ == Option.empty[Response[IO]])
       .assert
   }
+
+  test("Order of variable path should not matter") {
+    val router = Router[IO]("/foo" -> HttpRoutes.of {
+      case GET -> Root / variable =>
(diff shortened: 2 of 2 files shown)


danicheg (approved):
Nice catch.

rossabaker:
This is probably worth a patch release?

rossabaker:
I don't know what's broken, but, sure!  This starts bubbling up from 0.22 anyway.

PR #5992, opened 2022-02-02
Don't force npm onto ember.js users

armanbilge:
The ember h2 support on Scala.js requires an npm package [hpack.js](https://www.npmjs.com/package/hpack.js).

Previously, this was expressed as a global `@JSImport` annotation, which forced _all_ users of ember.js to reckon with npm and/or Scala.js bundler even if they are not enabling h2. This manifests as a breaking change for existing users.

This PR changes it to a "lazy" `require`, so the npm package is only loaded if h2 support is enabled.

The long term solution is still to replace the npm package with a pure Scala implementation. This will let us remove npm from our own build.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -532,7 +532,20 @@ lazy val emberCore = libraryCrossProject("ember-core", CrossType.Full)
     },
   )
   .jsSettings(
-    jsVersionIntroduced("0.23.5")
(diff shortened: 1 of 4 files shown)


rossabaker (approved):
I'm not qualified in this area, but smells alright.

PR #5994, opened 2022-02-02
Respect URI locality in UrlFormLifter

### LINKED ISSUE #5989: UrlFormLifter should respect Message's `caret` attribute
Request stores parsing caret in `attributes` vault, but while setting new Uri whole vault is cleaned. That's cause issue in case of appending Uri' query params in UrlFormLifter as the Request lose it's parsed context.

Example (will fail):
```
  val urlForm = UrlForm("foo" -> "bar")

  val app = UrlFormLifter(OptionT.liftK[IO])(HttpRoutes.of[IO] { case r @ POST -> Root / "path" =>
    r.uri.multiParams.get("foo") match {
      case Some(ps) =>
        Ok(ps.mkString(","))
      case None =>
        BadRequest("No Foo")
    }
  }).orNotFound

  test("Respect Uri locality context") {
    val req = Request[IO](method = POST, uri = uri"/some/prefix").withEntity(urlForm)
    Router[IO]("/some/prefix" -> app.mapK(OptionT.liftK[IO]))
      .run(req)
      .map(_.status)
      .value
      .assertEquals(Some(Ok))
  }
```

rossabaker:
Fixes #5989.  Backport of #5990.  All credit to @dfahritdinov.

### DIFF
--- a/server/src/main/scala/org/http4s/server/middleware/UrlFormLifter.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/UrlFormLifter.scala
@@ -45,6 +45,7 @@ object UrlFormLifter {
 
         val newRequest = req
           .withUri(req.uri.copy(query = newQuery))
+          .withAttributes(req.attributes)
           .withEmptyBody
         http(newRequest)
       }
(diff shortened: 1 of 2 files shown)


rossabaker:
I approve #5990 and added a trivial lint commit.  I'm going to merge this soon unless someone throws a flag. I'd like to release today.

PR #5999, opened 2022-02-03
Add scaladocs for EmberClientBuilder

valencik:
This PR adds scaladocs to all the EmberClientBuilder methods.

**Disclaimer**: I am very new to ember-client. This required a fair amount of reading the code to figure out where settings actually got used and what they did. I could easily be wrong.

**General principles**:
- Tried to highlight "where" the setting was applicable, for example calling out specifically when something affected the connection pool
- Mentioned default behaviours / values
- On things I had never heard of before (hello endpoint identification in HTTPS) I tried to add a bit more detail about why we care

### DIFF
--- a/ember-client/shared/src/main/scala/org/http4s/ember/client/EmberClientBuilder.scala
+++ b/ember-client/shared/src/main/scala/org/http4s/ember/client/EmberClientBuilder.scala
@@ -138,7 +138,7 @@ final class EmberClientBuilder[F[_]: Async] private (
   /** Sets the max `chunkSize` in bytes to read from sockets at a time. */
   def withChunkSize(chunkSize: Int) = copy(chunkSize = chunkSize)
 
-  /** Sets the max size in bytes the to read from sockets. */
+  /** Sets the max size in bytes to read while parsing response headers. */
   def withMaxResponseHeaderSize(maxResponseHeaderSize: Int) =
     copy(maxResponseHeaderSize = maxResponseHeaderSize)
 
--- a/ember-client/shared/src/main/scala/org/http4s/ember/client/EmberClientBuilder.scala
+++ b/ember-client/shared/src/main/scala/org/http4s/ember/client/EmberClientBuilder.scala
@@ -123,10 +123,10 @@ final class EmberClientBuilder[F[_]: Async] private (
   /** Sets the connection pool's maximum number of pooled connections per RequestKey. */
   def withMaxPerKey(maxPerKey: RequestKey => Int) = copy(maxPerKey = maxPerKey)
 
-  /** Sets the connection pool's maximum time a connection can be idle. */
+  /** Sets the connection pool's maximum time a connection can be idle.  The timeout starts when a connection is returned the the pool, and reset when it is borrowed. */
   def withIdleTimeInPool(idleTimeInPool: Duration) = copy(idleTimeInPool = idleTimeInPool)
 
-  /** Sets the idle timeout on connections. */
+  /** Sets the idle timeout on connections.  The timeout is reset with each read or write. */
   def withIdleConnectionTime(idleConnectionTime: Duration) =
     copy(idleConnectionTime = idleConnectionTime)
 
@@ -147,21 +147,21 @@ final class EmberClientBuilder[F[_]: Async] private (
   def withAdditionalSocketOptions(additionalSocketOptions: List[SocketOption]) =
     copy(additionalSocketOptions = additionalSocketOptions)
 
-  /** Sets the User-Agent string.
-    * Manually setting a `User-Agent` header takes priority over this setting.
+  /** Sets the default User-Agent string.
+    * A `User-Agent` header on a request takes priority over this setting.
     */
   def withUserAgent(userAgent: `User-Agent`) =
     copy(userAgent = userAgent.some)
 
-  /** Sets the User-Agent string to nothing, no User-Agent header would be sent.
-    * Manually setting a `User-Agent` header takes priority over this setting.
+  /** Clears the default User-Agent string, so no User-Agent header is sent.
+    * A `User-Agent` header on a request takes priority over this setting.
     */
   def withoutUserAgent =
     copy(userAgent = None)
 
   /** Sets whether or not to force endpoint authentication/verification on the `TLSContext`.
     * Enabled by default. When enabled the server's identity will be checked against the server's
-    * certificate during SSL/TLS handshaking. This is important to avoid man in the middle attacks
+    * certificate during SSL/TLS handshaking. This is important to avoid man-in-the-middle attacks
     * by confirming server identity against their certificate.
     */
   def withCheckEndpointAuthentication(checkEndpointIdentification: Boolean) =
@@ -181,10 +181,10 @@ final class EmberClientBuilder[F[_]: Async] private (
   def withUnixSockets(unixSockets: UnixSockets[F]) =
     copy(unixSockets = Some(unixSockets))
 
-  /** Enables HTTP2 support. Disabled by default. */
+  /** Enables HTTP/2 support. Disabled by default. */
   def withHttp2 = copy(enableHttp2 = true)
 
-  /** Disables HTTP2 support. Disabled by default. */
+  /** Disables HTTP/2 support. Disabled by default. */
   def withoutHttp2 = copy(enableHttp2 = false)
 
   /** Push promises are implemented via responding with a PushPromise frame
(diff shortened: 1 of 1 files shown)


ember-client/shared/src/main/scala/org/http4s/ember/client/EmberClientBuilder.scala
```diff
@@ -103,44 +103,88 @@ final class EmberClientBuilder[F[_]: Async] private (
...
+
+  /** Unset any `TLSContext` and creates one from the system default `SSLContext`. */
   def withoutTLSContext = copy(tlsContextOpt = None)
 
+  /** Sets the `SocketGroup`, a group of TCP sockets to be used in connections. */
   def withSocketGroup(sg: SocketGroup[F]) = copy(sgOpt = sg.some)
 
+  /** Sets the connection pool's total maximum number of idle connections.
+    * Per `RequestKey` values set with `withMaxPerKey` cannot override this total maximum.
+    */
```
valencik, line 120: If folks are happy with the connection pool specific docs, I'll add something similar to upstream [keypool](https://github.com/typelevel/keypool)

ember-client/shared/src/main/scala/org/http4s/ember/client/EmberClientBuilder.scala
```diff
@@ -103,44 +103,88 @@ final class EmberClientBuilder[F[_]: Async] private (
...
+    * certificate during SSL/TLS handshaking. This is important to avoid man in the middle attacks
+    * by confirming server identity against their certificate.
+    */
   def withCheckEndpointAuthentication(checkEndpointIdentification: Boolean) =
     copy(checkEndpointIdentification = checkEndpointIdentification)
 
+  /** Disables endpoint authentication/verification. */
   def withoutCheckEndpointAuthentication = copy(checkEndpointIdentification = false)
 
+  /** Sets the `RetryPolicy`. */
```
valencik, line 173: I would have linked to mention the default behaviour here, but the `RetryPolicy` used by default is custom and defined is a private object:
https://github.com/http4s/http4s/blob/9c7c19c264d79c00a0e305555dbdae3219bf5d07/ember-client/shared/src/main/scala/org/http4s/ember/client/internal/ClientHelpers.scala#L252-L272

Perhaps I could simplify things and say something like:
> The default retries on recoverable errors up to two times

Thoughts?

ChristopherDavenport: This exists because when a tcp disconnect occurs clients can successfully write, but then reading will fail.

So we retryin order to get rid of our dead connection for a hopefully live one.

ember-client/shared/src/main/scala/org/http4s/ember/client/EmberClientBuilder.scala
```diff
@@ -103,44 +103,88 @@ final class EmberClientBuilder[F[_]: Async] private (
...
   def withIdleConnectionTime(idleConnectionTime: Duration) =
     copy(idleConnectionTime = idleConnectionTime)
 
+  /** Sets the `Logger`. */
   def withLogger(logger: Logger[F]) = copy(logger = logger)
+
+  /** Sets the max `chunkSize` in bytes to read from sockets at a time. */
   def withChunkSize(chunkSize: Int) = copy(chunkSize = chunkSize)
+
+  /** Sets the max size in bytes the to read from sockets. */
```
ChristopherDavenport, line 139: Typo here with the the, I think

ember-client/shared/src/main/scala/org/http4s/ember/client/EmberClientBuilder.scala
```diff
@@ -103,44 +103,88 @@ final class EmberClientBuilder[F[_]: Async] private (
...
     copy(additionalSocketOptions = additionalSocketOptions)
 
+  /** Sets the User-Agent string.
+    * Manually setting a `User-Agent` header takes priority over this setting.
+    */
   def withUserAgent(userAgent: `User-Agent`) =
     copy(userAgent = userAgent.some)
+
+  /** Sets the User-Agent string to nothing, no User-Agent header would be sent.
+    * Manually setting a `User-Agent` header takes priority over this setting.
```
rossabaker, line 157:
```suggestion
  /** Clears the default User-Agent string, so no User-Agent header is sent.
    * A `User-Agent` header on a request takes priority over this setting.
```

ember-client/shared/src/main/scala/org/http4s/ember/client/EmberClientBuilder.scala
```diff
@@ -103,44 +103,88 @@ final class EmberClientBuilder[F[_]: Async] private (
...
+    */
   def withoutUserAgent =
     copy(userAgent = None)
 
+  /** Sets whether or not to force endpoint authentication/verification on the `TLSContext`.
+    * Enabled by default. When enabled the server's identity will be checked against the server's
+    * certificate during SSL/TLS handshaking. This is important to avoid man in the middle attacks
```
rossabaker, line 164:
```suggestion
    * certificate during SSL/TLS handshaking. This is important to avoid man-in-the-middle attacks
```

ember-client/shared/src/main/scala/org/http4s/ember/client/EmberClientBuilder.scala
rossabaker: If we're describing it, it's also worth mentioning it only applies to idempotent requests.

ember-client/shared/src/main/scala/org/http4s/ember/client/EmberClientBuilder.scala
```diff
@@ -103,44 +103,88 @@ final class EmberClientBuilder[F[_]: Async] private (
...
     copy(retryPolicy = retryPolicy)
 
+  /** Sets underlying `UnixSockets` to use for requests with a `UnixSocketAddress`.
+    * Useful for secure and efficient inter-process communication.
+    * See also `UnixSocket` client middleware to direct all requests to a `UnixSocketAddress`.
+    */
   def withUnixSockets(unixSockets: UnixSockets[F]) =
     copy(unixSockets = Some(unixSockets))
 
+  /** Enables HTTP2 support. Disabled by default. */
```
rossabaker, line 184:
```suggestion
  /** Enables HTTP/2 support. Disabled by default. */
```

ember-client/shared/src/main/scala/org/http4s/ember/client/EmberClientBuilder.scala
```diff
@@ -103,44 +103,88 @@ final class EmberClientBuilder[F[_]: Async] private (
...
   def withHttp2 = copy(enableHttp2 = true)
+
+  /** Disables HTTP2 support. Disabled by default. */
```
rossabaker, line 187:
```suggestion
  /** Disables HTTP/2 support. Disabled by default. */
```

rossabaker (commented):
Thanks! I picked few grammar nits and suggested a couple clarifications, but I'm glad to see this.

ember-client/shared/src/main/scala/org/http4s/ember/client/EmberClientBuilder.scala
valencik: @ChristopherDavenport The successful write after a tcp disconnect occurs, is that the nature of tcp or something specific to Ember client / http4s?
I'm trying to gauge how important we think this is to mention.

I wonder what other `RetryPolicy`s are sensible. Should they mostly all consider:
- only retrying idempotent requests
- only retrying on `ClosedChannelException`, `Connection reset by peer`, or `Broken pipe`
- retry at least a minimum of two times

If so, perhaps we could introduce a simpler builder option to simply increase the number of retries?
Something like:
```scala
  /** Sets the maximum number of times to retry a `Request` with the default `RetryPolicy`.
   *  The default `RetryPolicy` retries idempotent requests on recoverable errors up to two times.
   *  For more customization see `withRetryPolicy`. 
   */
  def withMaxRetries(maxRetries: Int) =
    copy(maxRetries = maxRetries)
```

---

Independent of the above (which, if wanted, would happen in a separate PR)
How's this for the current `withRetryPolicy`:

> Sets the `RetryPolicy`
> The default retries idempotent requests on recoverable errors up to two times.

rossabaker: Blaze added a `withRetries(Int)` recently.  I initially planned to copy what Ember did, but we didn't have a `Timer`, so we couldn't use the `Retry` middleware on 0.22.  (In 0.23, we had an `Async`, and could.)  Ember's version is more flexible, but Blaze's ends up being a lot more concise to change the only value that is likely to change.  I think a convenience setting like that on Ember could be nice.

rossabaker (approved):

PR #6001, opened 2022-02-03
Add convenience methods to `WSConnectionHighLevel`

armanbilge:
I'm also sneaking in a couple convenience methods to the high-level client.

### DIFF
--- a/client/src/main/scala/org/http4s/client/websocket/WSClient.scala
+++ b/client/src/main/scala/org/http4s/client/websocket/WSClient.scala
@@ -132,6 +132,12 @@ trait WSConnection[F[_]] { outer =>
 
 trait WSConnectionHighLevel[F[_]] { outer =>
 
+  /** Send a single websocket text frame. The sending side of this connection has to be open. */
+  def send(text: String): F[Unit] = send(WSFrame.Text(text))
+
+  /** Send a single websocket binary frame. The sending side of this connection has to be open. */
+  def send(bytes: ByteVector): F[Unit] = send(WSFrame.Binary(bytes))
+
   /** Send a single websocket frame. The sending side of this connection has to be open. */
   def send(wsf: WSDataFrame): F[Unit]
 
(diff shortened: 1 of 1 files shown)


client/src/main/scala/org/http4s/client/websocket/WSClient.scala
```diff
@@ -130,7 +130,13 @@ private[http4s] trait WSConnection[F[_]] { outer =>
...
 }
 
-private[http4s] trait WSConnectionHighLevel[F[_]] { outer =>
+trait WSConnectionHighLevel[F[_]] { outer =>
+
+  /** Send a single websocket text frame. The sending side of this connection has to be open. */
+  def send(text: String): F[Unit] = send(WSFrame.Text(text))
+
+  /** Send a single websocket binary frame. The sending side of this connection has to be open. */
+  def send(bytes: ByteVector): F[Unit] = send(WSFrame.Binary(bytes))
```
armanbilge, line 139: Thoughts on overloads vs `sendText` and `sendBytes`?

armanbilge:
Looks like the jdk client is not going to implement this for 0.22, see https://github.com/http4s/http4s-jdk-http-client/issues/568.

http4s-dom also doesn't support 0.22 at all which doesn't support JS at all.

Which leads me to wonder if we should make this public at all? Since we have no concrete plans to implement it at this time.

client/src/main/scala/org/http4s/client/websocket/WSClient.scala
rossabaker: I tend to avoid overloads these days.

armanbilge: Yeah, it's good practice. But I had to try :P

rossabaker:
I would not make it public on this branch before there is at least one, preferably two, concrete implementations.  If one drops from the sky, we can make it public here, but it's probably better to unveil it in 0.23.

rossabaker (approved):

PR #6003, opened 2022-02-04
Release v1.0.0-M31

rossabaker:

### DIFF
--- a/website/src/hugo/content/changelog.md
+++ b/website/src/hugo/content/changelog.md
@@ -7,6 +7,64 @@
 Maintenance branches are merged before each new release. This change log is
 ordered chronologically, so each release contains all changes described below it.
 
+# v1.0.0-M31 (2022-02-04)
+
+This is the latest milestone release in the 1.0 series.  It breaks binary compatibility and drops support for Scala 2.12.  Scala 2.12 is still supported in the 0.22.x and 0.23.x series.
+
+This merges forward all the changes from 0.22.9, 0.22.10, 0.22.11, 0.23.8, 0.23.9, and 0.23.10.
+
+* http4s-core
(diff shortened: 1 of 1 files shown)


rossabaker:
GitHub decided to pull these release notes starting on 0.23.9, so our automation kind of sucked on this one.  I did my best.

armanbilge:
Oh yeah, I think that's the state-of-the-art for pre-releases. It always starts over from the last stable release and ignores all previous pre-releases in between. I wonder if that's configurable.

armanbilge:
Every time we cut a 1.0 milestone, we could label all the merged PRs up to that point with some "merged" label that we exclude from the release notes. Stoopid.

armanbilge (approved):

rossabaker:
It might be better next time, because this included some that predate the label.  Let's see how bad M32 is before we panic.

armanbilge:
I've done lots of pre-releases lately (sbt-typelevel and feral) and this is how it's worked for those.

rossabaker:
I tried to be smarter than it by not checking the prerelease box, but, nope.

rossabaker:
We could make mergify tag these as series/1.0, I guess.

armanbilge:
We'd have to go back and un-pre-release all the previous pre-releases as well for that to work I think. Apparently pre-releases don't really count for announcing stuff.

armanbilge:
> We could make mergify tag these as series/1.0, I guess.

I'm not sure that would solve the problem, since I think it still would include `series/1.0` PRs from previous pre-releases. That's why we need a way of marking PRs after we make a release.

rossabaker:
Oh, hell.  You're right.

PR #6005, opened 2022-02-04
Make `WSClient` and friends public

armanbilge:
We now have implementations ready in:
- https://github.com/http4s/http4s-dom/pull/93
- https://github.com/http4s/http4s-jdk-http-client/pull/571

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -411,6 +411,29 @@ lazy val client = libraryCrossProject("client")
       ProblemFilters.exclude[DirectMissingMethodProblem](
         "org.http4s.client.oauth1.*.generateHMAC"
       ), // private[oauth1]
+      // begin wsclient: initially private[http4s]
+      ProblemFilters.exclude[IncompatibleResultTypeProblem](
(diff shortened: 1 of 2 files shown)


rossabaker:
Needs the MiMa exemptions, but :+1:.

armanbilge:
Those seem less than ideal. Like they could all-too-easily filter legit issues due to future changes :(

rossabaker (approved):
I don't think there's much that can be done about the broad exemptions, other than to start MiMaing against 0.23.11, which is also not ideal.

armanbilge:
> start MiMaing against 0.23.11, which is also not ideal.

That's even more dangerous, because MiMa automatically ignore package-private stuff. So if something was public in 0.23.0 and then made package-private in 0.23.1, unless we specifically compare against 0.23.0 we could delete that method without getting any warnings.

But you are right that MiMa exemptions should be tied to specific versions e.g. these filters here should only apply when comparing to artifacts before 0.23.11. From 0.23.11 on there should be a clean slate. I do have some imagination of doing something about this at some point 😆

PR #6009, opened 2022-02-06
Replace npm hpack.js with pure Scala.js hpack

armanbilge:
I just published a snapshot of https://github.com/armanbilge/hpack.

It's a pure-Scala fork/port of the Twitter hpack Java lib, including tests, that cross-compiles to Scala.js. Unfortunately many of the tests rely on Java frameworks not available on JS so we can't run those atm, but they do pass on the JVM. I'd like to transfer the repo to http4s-org.

This PR uses that lib to replace the hpack.js npm dependency for ember.js. This enables us to remove the Scala.js Bundler plugin from the build along with nearly 20k lines of lockfiles. It also means that ember-h2.js will work for downstreams out-of-the-box.

Ember JVM continues to use the original Twitter hpack lib; there seems little reason to replace it with the Scala clone.


cquiroz:
This is really cool, we did the same for one our java dependencies

rossabaker (approved):
Is Twitter hpack "done"?  I see it hasn't changed since 2019.  My concern would be this implementation diverging from upstream.  Certainly not worse than the status quo, because we had two distinct implementations.

rossabaker:
And I'm also fine transferring that into the org.

armanbilge:
I think it's effectively done. The hpack.js npm package also hasn't been updated since 2016.

Note that my port is a fork of the Twitter repo and most of changes I made maintain a high line-by-line fidelity. So in theory if there are any upstream changes we wouldn't be completely lost with how to incorporate them.

armanbilge:
I tried transferring it but it seems I'm unable you. I'll try adding you as a collaborator and then you can transfer it?

rossabaker:
I don't have access to the settings and can't transfer it.  I just turned on repo creation for members, which I think will let you transfer it.

armanbilge:
Done, thanks. I guess we can add it to the core team? Btw the crypto repo can be added as well.

armanbilge:
Apparently Netty has its own fork of Twitter hpack with additional optimizations. Some discussion here:
- https://github.com/akka/akka-http/issues/304

armanbilge:
Ooh. Twitter HPACK is _done_ done.

> Yes, this library is not maintained, and no one uses it at Twitter.

- https://github.com/twitter/hpack/pull/32#issuecomment-247168313

Also some discussion there about why bother implementing http server/clients when you could just use Netty 🙃

rossabaker:
Ah, well, at least it's not a moving target.  Before 1.0, we should think about shading it, unless it's been replaced.  Taking away a transitive dependency is a form of breaking change.

rossabaker:
I just went ahead and created teams for crypto and hpack, since I'm not sure everyone wants the noise.  If anyone wants to join those and help, please speak up.

armanbilge:
@rossabaker can you also enable issues for the hpack fork? I can't access those settings.

rossabaker:
I just gave you admin.  I guess that didn't transfer over.

armanbilge:
Hold on this pending a release of https://github.com/http4s/hpack.

PR #6011, opened 2022-02-06
Message: add pipeBodyThrough method

diesalbla:
Add an `alterBodyBy` function to `Message`, which is just to pass a function. Use it at a few places on the code base.

_Edit_ since this is a stream to stream transformation, the method is called `pipeBodyTo`.

### DIFF
--- a/core/shared/src/main/scala/org/http4s/Message.scala
+++ b/core/shared/src/main/scala/org/http4s/Message.scala
@@ -29,6 +29,7 @@ import com.comcast.ip4s.Hostname
 import com.comcast.ip4s.IpAddress
 import com.comcast.ip4s.Port
 import com.comcast.ip4s.SocketAddress
+import fs2.Pipe
 import fs2.Pure
 import fs2.Stream
 import fs2.io.net.unixsocket.UnixSocketAddress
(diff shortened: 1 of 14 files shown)


core/shared/src/main/scala/org/http4s/Message.scala
```diff
@@ -117,6 +117,11 @@ sealed trait Message[F[_]] extends Media[F] { self =>
   def withEmptyBody: Self =
     withBodyStream(EmptyBody).transformHeaders(_.removePayloadHeaders)
 
+  /**
+    */
```
danicheg, line 121: Probably something was intended to add here?

rossabaker: Maybe worth mentioning it doesn't update any existing `Content-Length` header.  It's kind of a dangerous function, but as seen elsewhere in this PR, it is useful!

diesalbla: So, it does not update any existing `Content-Length` header, just as the existing public method `withBodyStream` does not, so I was using that precedent.

Nevertheless, and since this is intended to be an Http4S utility, I have marked it as private to http4s.

ChristopherDavenport (approved):

PR #6021, opened 2022-02-07
docs: remove workarounds for issues in Laika 0.18.0

jenshalm:
This PR removes a small number of workarounds that were added for issues in Laika 0.18.0 which have all been resolved with the 0.18.1 release in December.

It's a collection of unrelated changes, but splitting them into individual PRs felt overkill here. Happy to extract some of them into separate PRs if that is preferred.

List of changes, sorted by associated resolved issues:

* [Helium theme: better support for portrait orientation on phones](https://github.com/planet42/Laika/issues/208)
   Upgrading brought you some of the improvements, but it's only complete by also merging the corresponding template 
   additions. These are the new `${?helium.topBar.phoneLinks}` lines in three of the templates.

* [laikaDescribe task broken when used in a project with mdoc](https://github.com/planet42/Laika/pull/220)
   The changes in `build.sbt` that disabled this task are no longer necessary.
   
* [mailto links are wrongly validated as internal links](https://github.com/planet42/Laika/pull/223)
   This PR removes the workaround from `HeliumExtensions` as well as from the markup in `code-of-conduct.md`
   
* [Helium: fix CSS for anchor placement on section headers](https://github.com/planet42/Laika/pull/228)
  The entire removed block in `SiteConfig` was only necessary as left placement of anchors was broken. 
  All other values correspond to the built-in defaults. (If you have a strong preference for right placement,
  this change would need to be removed from the PR, but I think left placement is more common and more obvious
  for most users).

* [Helium: enhance version dropdown for use on unversioned pages](https://github.com/planet42/Laika/pull/225)
  There were bugs in the JavaScript for the version dropdown on unversioned pages. This PR removes the full copy
  of `laika-versions.js` which is now identical in the 0.18.1 release and also reverts the various places in templates where
  the builtin `@:heliumInitVersions` was replaced by a manual override.

After this PR the only workarounds remaining are:

1) The need for full copies of the default templates (to be addressed in 0.19)
2) The PrettyURL extension (planned as an addition for 0.19) 
3) The extension for variable substitutions in code blocks (suggested alternative: use the corresponding mdoc feature)

Regarding 3) I'd be happy to do a separate PR if you want to try this out. When I did the original work I wasn't aware that mdoc includes such a feature. It's currently quite unlikely that this feature will be part of Laika 0.19 as it's by far the most complicated feature on the 0.19 road map due to the need to rework the entire phasing system to enable this. I feel using mdoc would be a natural choice.


armanbilge:
Thanks for following up on this!

We also currently have a problem that whenever the site is published from series/0.22 or series/0.23 it removes the landing page from main. Would you be able to help us solve that?

https://github.com/http4s/http4s/issues/5919

👍 to mdoc variables, I think that's the way to go.

jenshalm:
Did this always happen or is this something that just started surfacing?

armanbilge:
I'm not sure, tbh 😕 

Here's the commit history, and the reverts we use to restore the site when it goes down.

https://github.com/http4s/http4s/commits/gh-pages

jenshalm:
Do you know a commit where you are certain it came from a branch and not from main? It's hard to follow when all commit messages are the same...
I found one (https://github.com/http4s/http4s/commit/20b1b69f921ff48295c281fa1f4032adef3fc3cd) where `index.html` was removed, but the other unversioned pages were modified instead, whereas not a single versioned file was changed. That would be strange if it would be coming from a branch.

armanbilge:
Sure, I can link https://github.com/http4s/http4s/commit/20b1b69f921ff48295c281fa1f4032adef3fc3cd to the 0.22.10 release in https://github.com/http4s/http4s/commit/bd3e078683964c550df70a8b6298ba6bbccec1d9.

Here's the CI job where that commit got created and pushed:
https://github.com/http4s/http4s/runs/4986928113?check_suite_focus=true#step:17:131

jenshalm:
That looks completely broken. I don't have more time today, but the most likely culprit seems to be the definition of the CI job. The website build should not run at all on a branch.

armanbilge (approved):

armanbilge:
Ah, thanks, that's a helpful pointer. I might be able to fix it.

armanbilge:
It looks to me like the wrong one is disabled.

https://github.com/http4s/http4s/blob/8600a2116a33cdc982ab53ec129667b2aacd9df3/project/Http4sPlugin.scala#L201-L204

Btw, what was the reason we have a separate website and docs projects?

rossabaker (approved):
@jenshalm, your writeups are a joy.  Thank you for this!  

> The extension for variable substitutions in code blocks (suggested alternative: use the corresponding mdoc feature)

This is referring to [this](https://github.com/http4s/http4s/blob/0f2295e57394d1377071960fecb7863a2c53923e/project/SiteConfig.scala#L112-L127)?  What's left there doesn't seem so bad, if it saves you from complicating laika-0.19.

> Do you know a commit where you are certain it came from a branch and not from main? It's hard to follow when all commit messages are the same...

I think we lost this when we moved to sbt-ghactions.  We used to get the "source commit" in the message.

> Btw, what was the reason we have a separate website and docs projects?

`docs` used to publish on every commit  from multiple branches, or on tags.  `website` used to publish on every commit, from main only.  And they each had their own set of assets, being logically separate deployments, so one couldn't break another.

armanbilge:
> docs used to publish on every commit from multiple branches, or on tags. website used to publish on every commit, from main only. And they each had their own set of assets, being logically separate deployments, so one couldn't break another.

Ah, I see, thanks for the explanation. If you know me then you know I'll say we should publish both website/docs on every push to every branch, and thus combine them into a single project, and avoid some of this complexity. But that's for another PR 😆

rossabaker:
`docs` is versioned content, and `website` is not.  Publishing `website` from older branches would overwrite the common areas of the site.

armanbilge:
It wouldn't if configured right via `renderUnversioned`, Laika supports version/unversioned pages. See https://github.com/typelevel/sbt-typelevel/issues/100#issuecomment-1022303454 .

jenshalm:
> It wouldn't if configured right via `renderUnversioned`, Laika supports version/unversioned pages. See [typelevel/sbt-typelevel#100 (comment)](https://github.com/typelevel/sbt-typelevel/issues/100#issuecomment-1022303454) .

As long as you have the projects separate, `website` should not run on CI at all as it does not have any unversioned pages anyway (so even if configured correctly would never publish anything ever). 

Merging the projects would obviously work, and was something I actually recommended in my original work on introducing Laika, but Ross had some objections back then, so I didn't do it. :-)

jenshalm:
> This is referring to [`variables`](https://github.com/http4s/http4s/blob/0f2295e57394d1377071960fecb7863a2c53923e/project/SiteConfig.scala#L112-L127)? What's left there doesn't seem so bad, if it saves you from complicating laika-0.19.

The variable definitions themselves can (and must) stay in some form of course. But in the long run I would not recommend to use the custom Laika extension you have to actually substitute them in code blocks which is the extension here: https://github.com/http4s/http4s/blob/main/project/HeliumExtensions.scala#L105

Those were mostly a hack and have some limitations (e.g. only work inside string literals). That's why I suggested to do a PR where this job is performed by mdoc itself, which would not have any limitations (just different syntax in the code blocks).

This way you have a more robust solution and less custom code, and I can push this feature to 0.20 in the meantime (which would mean it's not going to happen this year). The custom extension has those limitations as the syntax highlighter runs before Laika's variable substitution and making the phase model more flexible in general is a big piece of work. mdoc runs before Laika anyway, so it's a safer place for this kind of substitution for now.

rossabaker:
The mdoc solution looks good to me.  I definitely don't want to complicate the work on Laika.

As long as merging to a series branch doesn't overwrite the canonical copy on main, I don't have an ongoing objection to merging.  Though we should also be mindful of where docs live after the Great Schism: are those going to subdirectories of http4s.org, or are those going to subdomains?  That question may influence the design here.

But I think we're digressing... this is ready to merge, yes?

armanbilge:
This PR doesn't solve everything, but it's a very good one. The remaining issues are:
- use mdoc variables and remove the workaround here
- fix the website deploy problem. I think I identified the mistake in https://github.com/http4s/http4s/pull/6021#issuecomment-1031972217
Those can be follow ups.

> Though we should also be mindful of where docs live after the Great Schism: are those going to subdirectories of http4s.org, or are those going to subdomains?

I quickly glanced through: are there actually any docs that apply to stuff that will be schismed out of the core? The biggest problem is that the examples use blaze instead of ember but that was fixed for 0.23 up in https://github.com/http4s/http4s/pull/5970.

> The custom extension has those limitations as the syntax highlighter runs before Laika's variable substitution and making the phase model more flexible in general is a big piece of work. mdoc runs before Laika anyway, so it's a safer place for this kind of substitution for now.

Btw @jenshalm since using mdoc is pretty standard for any docs with code blocks in it and adding this feature seems fairly complex IMHO I would say Laika could just skip this one :)

danicheg (approved):

armanbilge:
Oh and @rossabaker I would still like to simplify the site stuff here and have it run through sbt-typelevel-site/sbt-http4s-org. I know you're flooded in review requests (among a million other things!) but when you have a chance if you could pick up from https://github.com/typelevel/sbt-typelevel/pull/116#issuecomment-1026437663 that would help unblock it. Thanks :)

jenshalm:
Ok, like last autumn, I'll leave it up to you to debate the pros and cons of merging `docs` and `website`. 😄  It really depends on your future plans. If you expect both projects will still live in the same repo in two year's time, I think it'll be more convenient to merge them. If you are not sure about this, you may want to wait with making a decision.

In any case, if you DO want to merge them, feel free to ping me with questions and/or request reviews of the relevant PRs, there are quite a few things that can go wrong. Both Laika config and the filters for the gh-pages plugin need to be set up differently for branches and `main` to make it work.

> Btw @jenshalm since using mdoc is pretty standard for any docs with code blocks in it and adding this feature seems fairly complex IMHO I would say Laika could just skip this one :)

If it would just be about variables in Scala code blocks, I'd totally agree and would not bother. The two reasons I still plan to do this bigger piece of work for 0.20 are twofold: 1) people do funny things with Laika's syntax highlighter, someone just opened a PR to add a highlighter for Dart. Those users cannot use mdoc as a workaround and 2) making the phase model more flexible has other advantages beyond variable substitution - for some extensions like AST rewrite rules the fact that you cannot chose which phase to run in limits some of the things you can do (e.g. a rule cannot insert a headline that would contribute to table of contents at the moment).

I might be able to do the mdoc PR over the weekend.

PR #6027, opened 2022-02-08
Flush the prelude in non-blocking Servlet IO

rossabaker:
We need to flush the response prelude before the first chunk when rendering a response with chunked transfer encoding.  This reaches parity with Ember and Blaze.

I thought we could just set the buffer size to zero, but this renders each byte as its own chunk.


rossabaker:
Yes, yes, I know this is from a less popular module on an old version.  Fun times at $WORK.

isomarcte (approved):

PR #6028, opened 2022-02-08
Render continually between response prelude and body

rossabaker:
When we acquire a `Response`, it is the server's responsibility to ensure that its body's finalizer runs.  Lacking a resourceful model of a response (#2663), the `.attempt.flatMap` can be interrupted and cause a leak.

We can't fix this in general, because `Http4sServlet` only requires an `Effect`.  We do a best effort by pattern matching the instance.  In `AsyncHttp4sServlet`, we can guarantee the better behavior.

Bonus: plug the same leak in blaze-server and ember-server

### DIFF
--- a/blaze-server/src/main/scala/org/http4s/server/blaze/Http2NodeStage.scala
+++ b/blaze-server/src/main/scala/org/http4s/server/blaze/Http2NodeStage.scala
@@ -18,7 +18,8 @@ package org.http4s
 package server
 package blaze
 
-import cats.effect.{ConcurrentEffect, IO, Sync, Timer}
+import cats.effect.{ConcurrentEffect, IO, Timer}
+import cats.effect.syntax.all._
 import cats.syntax.all._
 import fs2._
 import fs2.Stream._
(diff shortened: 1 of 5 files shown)


rossabaker:
I'm afraid [this line](https://github.com/http4s/http4s/blob/series/0.21/servlet/src/main/scala/org/http4s/servlet/AsyncHttp4sServlet.scala#L87) is still a leak.  We cancel `response` when it loses, but if the effect completes and buries its finalizer in `response.body` before it gets canceled for being the loser, the finalizer won't run.  This is why we need responses to be returned as a resource.

blaze-server/src/main/scala/org/http4s/server/blaze/Http2NodeStage.scala
```diff
@@ -224,18 +225,20 @@ private class Http2NodeStage[F[_]](
...
             .recoverWith(serviceErrorHandler(req))
-            .flatMap(renderResponse)
-
-          F.runAsync(action) {
-            case Right(()) => IO.unit
-            case Left(t) =>
-              IO(logger.error(t)(s"Error running request: $req")).attempt *> IO(closePipeline(None))
-          }
-        }.unsafeRunSync()
+            .continual {
```
danicheg, line 231: _[Just trying to figure things out]_: What actually could cause interruption if we are run execution via `Runnable` here? Is it some sort of JVM signals?

rossabaker: That's a good question.  In `Http1ServerStage`, we expose a cancel token outside the Runnable, but I don't see how this one might be canceled.  (Which is probably something this stage ought to do!)

isomarcte:
@rossabaker do we need to deprecate and add alternative constructors here? https://github.com/http4s/http4s/blob//servlet/src/main/scala/org/http4s/servlet/BlockingHttp4sServlet.scala#L73

isomarcte (approved):

blaze-server/src/main/scala/org/http4s/server/blaze/Http2NodeStage.scala
rossabaker: I don't think this hurts anything, and leaves us in a better place if anyone ever adds proper cancellation support.  (Though I can't imagine it happening on CE2...)

PR #6029, opened 2022-02-08
Adds async timeout as param

### LINKED ISSUE #1812: Unable to change the timeout when using mountService in ServletContextSyntax.scala
Comment from @rossabaker  
Oh, I see the problem
https://github.com/http4s/http4s/blob/master/servlet/src/main/scala/org/http4s/servlet/syntax/ServletContextSyntax.scala
We're setting the timeout there, but not passing it through as a param. That's bad design.

yuferpegom:
closes #1812 
- `mountService` method deprecated in spite of `mountRoutes` which now receives the async timeout as param

### DIFF
--- a/servlet/src/main/scala/org/http4s/servlet/syntax/ServletContextSyntax.scala
+++ b/servlet/src/main/scala/org/http4s/servlet/syntax/ServletContextSyntax.scala
@@ -39,7 +39,7 @@ final class ServletContextOps private[syntax] (val self: ServletContext) extends
     *
     * Assumes non-blocking servlet IO is available, and thus requires at least Servlet 3.1.
     */
-  @deprecated("Use mountRoutes instead")
+  @deprecated("Use mountRoutes instead", "0.23.10")
   def mountService[F[_]: Async](
       name: String,
       service: HttpRoutes[F],
--- a/servlet/src/main/scala/org/http4s/servlet/syntax/ServletContextSyntax.scala
+++ b/servlet/src/main/scala/org/http4s/servlet/syntax/ServletContextSyntax.scala
@@ -39,12 +39,21 @@ final class ServletContextOps private[syntax] (val self: ServletContext) extends
     *
     * Assumes non-blocking servlet IO is available, and thus requires at least Servlet 3.1.
     */
+  @deprecated("Use mountRoutes instead")
   def mountService[F[_]: Async](
       name: String,
       service: HttpRoutes[F],
       mapping: String = "/*",
       dispatcher: Dispatcher[F],
-      asyncTimeout: Duration = defaults.ResponseTimeout
+  ): ServletRegistration.Dynamic =
+    mountHttpApp(name, service.orNotFound, mapping, dispatcher)
+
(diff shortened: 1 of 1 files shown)


servlet/src/main/scala/org/http4s/servlet/syntax/ServletContextSyntax.scala
```diff
@@ -38,6 +39,7 @@ final class ServletContextOps private[syntax] (val self: ServletContext) extends
     *
     * Assumes non-blocking servlet IO is available, and thus requires at least Servlet 3.1.
     */
+  @deprecated("Use mountRoutes instead", "0.23.10")
```
rossabaker, line 42:
```suggestion
  @deprecated("Use mountRoutes instead", "0.23.11")
```

servlet/src/main/scala/org/http4s/servlet/syntax/ServletContextSyntax.scala
```diff
@@ -46,15 +48,25 @@ final class ServletContextOps private[syntax] (val self: ServletContext) extends
...
+      asyncTimeout: Duration = defaults.ResponseTimeout,
+  ): ServletRegistration.Dynamic =
+    mountHttpApp(name, service.orNotFound, mapping, dispatcher, asyncTimeout)
+
   def mountHttpApp[F[_]: Async](
       name: String,
       service: HttpApp[F],
       mapping: String = "/*",
       dispatcher: Dispatcher[F],
+      asyncTimeout: Duration = defaults.ResponseTimeout,
```
rossabaker, line 65: This is _source_ compatible, but not _binary_ compatible.  Default arguments are applied at compile time, but at link time, they don't exist.  So this is good, but we need to keep the old version too.  I think the old copy needs to:

1. be deprecated, so something screams if we try to call it 
2. be `private[servlet]`, so we don't get in the way of callers of this version.
3. remove the default argument to `mapping`, because two overloads can't have default arguments
4. call the new one like `mountHttpApp(name, service, mapping, defaults.ResponseTimeout)`

servlet/src/main/scala/org/http4s/servlet/syntax/ServletContextSyntax.scala
rossabaker: Or 1.0.0-M32, if we don't end up targeting 0.23.

rossabaker (changes requested):
This looks good.  If you want to keep targeting `main`, you can ignore my comments about `mountHttpApp`.  But if you want to use it in production now, we should target `series/0.23`, and my comment below shows how to keep it binary compatible.

Either way, we need to update the deprecated calls in examples-war to `mountRoutes`.

servlet/src/main/scala/org/http4s/servlet/syntax/ServletContextSyntax.scala
yuferpegom: Oh I understand now. I'll close this pr and open a new one to target `series/0.23`.

Issue #6032, opened 2022-02-09
`[warn] Multiple resolvers having different access mechanism configured with same name 'sonatype-snapshots'`


armanbilge:
Currently seeing this warning in sbt. This will interfere with publishing snapshots to s01, ~~but I think releases will be ok.~~ Actually, I'm not sure.

It is the interaction of:
- https://github.com/xerial/sbt-sonatype/pull/279
- https://github.com/scalacenter/scalafix/issues/1535

Until this is fixed in sbt-sonatype, I suppose we should hotfix it in sbt-typelevel.

armanbilge:
This got fixed in some update.

PR #6033, opened 2022-02-09
Release v0.21.32

rossabaker:
It pays the bills.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -739,6 +739,7 @@ lazy val scalafixSettings: Seq[Setting[_]] = Seq(
 lazy val scalafixRules = project
   .in(file("scalafix/rules"))
   .settings(scalafixSettings)
+  .settings(sonatypeMigrationShims)
   .settings(
     moduleName := "http4s-scalafix",
     libraryDependencies += "ch.epfl.scala" %% "scalafix-core" % V.scalafix,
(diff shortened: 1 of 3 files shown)


armanbilge (approved):
Are we going to need to adjust the build to get CI release working?

rossabaker:
Oh, crap, probably.  We just need to set the hostname to S01, right?

armanbilge:
Yeah we need to set `sonatypeCredentialHost := "s01.oss.sonatype.org"` for all the projects (`ThisBuild` doesn't work) so it would be good to plop that in the auto-plugin.

Does 0.21 have scalafixes in here? We might butt heads with https://github.com/http4s/http4s/issues/6032.

armanbilge:
Yeesh, it seems it does. So we could also copy-paste the stuff from https://github.com/typelevel/sbt-typelevel/pull/150 as well....

rossabaker:
I think that credential host needed to be on the root project because of the bundling.

armanbilge:
Oops, yes it does. That's why I suggested the auto-plugin, since it would do that automatically.

armanbilge:
Actually sbt-sonatype 3.9.11 was just released with my fix.

rossabaker:
I tried putting it there, and it didn't trigger.  I guess I'll publish this one manually, and can clean up in case there's another 0.21.  Hopefully not, but.

Issue #6035, opened 2022-02-09
Ember Server is returning a None for Socket Pings breaking `.withFilterPingPongs(false)`


CharlesAHunt:
Ember Server returns a None for all Websocket Pings in `WebSocketHelpers.scala` inside `private def handleIncomingFrames` in the following case:

`case WebSocketFrame.Ping(data) => writeFrame(WebSocketFrame.Pong(data)).as(None)`

This disregards the setting in the `WebSocketBuilder2` of `.withFilterPingPongs(false)` causing Ping frames to get filtered out, effectively breaking this functionality.

The case handling incoming frames should return a Some(PingFrame) in the case `withFilterPingPongs`is false.

PR #6036, opened 2022-02-10
Resolve broken filterPingPongs using WebSocketBuild2 with Ember

CharlesAHunt:
Pings are not passing through when using the Ember server with the WebSocketBuilder2 setting `.withFilterPingPongs(false)` due to  `WebSocketHelpers.handleIncomingFrames` always mapping them to a None regardless of the filter setting.

Fixes https://github.com/http4s/http4s/issues/6035


armanbilge:
Yeah, this is a tricky one :) the good news is that it is possible to make the change binary-compatibly.

You'll have to make `WebSocketContext` into an ordinary class and manually restore all the missing case class methods (if you are good at reading MiMa issues, they will guide you with which methods to add). Also because case classes are encoded differently on Scala 2/3 it will necessitate splitting sources.

We recently did this in https://github.com/http4s/http4s/pull/5758 and https://github.com/http4s/http4s/pull/5926/commits/566fb53711a2d745e835ea3711b4d183b3beec90 although that was complicated by type erasure problems as well.

Binary-breaking changes can go to main, but it won't help you if you are expecting a stable release anytime soon.

armanbilge:
So I looked into this a bit more, and I'm a little confused about why we need to make any changes to `WebSocketContext` at all actually.

1. The PR that added the `filterPingPongs` API also implemented support for this feature in Blaze. So it's possible to implement without this modification to `WebSocketContext`.
  https://github.com/http4s/http4s/pull/3136
2. In fact, Blaze's implementation doesn't appear to use `filterPingPongs` at all. AFAICT the parameter is used only in `WebSocketBuilder2`:
https://github.com/http4s/http4s/blob/6e13209e8b7ae6d2369ff7c32cc26c1b31d9e35e/server/shared/src/main/scala/org/http4s/server/websocket/WebSocketBuilder2.scala#L162-L166
3. The PR adding this feature made the following change to Blaze:
```diff
- case Ping(d) =>
+ case p @ Ping(d) =>
  //Reply to ping frame immediately
- writeFrame(Pong(d), trampoline) >> handleRead()
+ writeFrame(Pong(d), trampoline) >> F.pure(p)
```

So I think your PR is on the right track ...

ember-server/shared/src/main/scala/org/http4s/ember/server/internal/WebSocketHelpers.scala
```diff
@@ -164,13 +164,17 @@ object WebSocketHelpers {
...
     def writeFrame(frame: WebSocketFrame): F[Unit] =
       Stream(frame).through(encodeFrames(frameTranscoder)).through(write).compile.drain
 
     stream =>
       stream.evalMapFilter[F, WebSocketFrame] {
-        case WebSocketFrame.Ping(data) => writeFrame(WebSocketFrame.Pong(data)).as(None)
+        case WebSocketFrame.Ping(data) =>
+          val pong = writeFrame(WebSocketFrame.Pong(data))
+          if (filterPings) pong.as(None)
+          else pong.as(WebSocketFrame.Ping(data).some)
```
armanbilge, line 177: ... but the only changes we need are here, to align with the Blaze implementation.
https://github.com/http4s/http4s/blob/6e13209e8b7ae6d2369ff7c32cc26c1b31d9e35e/blaze-core/src/main/scala/org/http4s/blazecore/websocket/Http4sWSStage.scala#L126-L128

CharlesAHunt: Aha, makes sense.  Thanks for the feedback! Made the update 👍🏻

ember-server/shared/src/main/scala/org/http4s/ember/server/internal/WebSocketHelpers.scala
```diff
@@ -196,7 +198,7 @@ object WebSocketHelpers {
         }
         Stream
           .iterable(chunks)
-          .flatMap(Stream.chunk(_))
+          .flatMap(Stream.chunk)
```
armanbilge, line 201: Scala 3 requires the new syntax, I think fatal warnings will catch you here :)
```suggestion
          .flatMap(Stream.chunk(_))
```

ember-server/shared/src/main/scala/org/http4s/ember/server/internal/WebSocketHelpers.scala
armanbilge: Fantastic, thanks!

armanbilge (commented):
I'm inclined to approve. Is there an easy test we can add to lock-in this behavior?

ember-server/shared/src/main/scala/org/http4s/ember/server/internal/WebSocketHelpers.scala
armanbilge: Huh, I must be confused 🤔  Ignore me.

CharlesAHunt:
> I'm inclined to approve. Is there an easy test we can add to lock-in this behavior?

I think I should be able to whip one up.  I'll get one added this week. 👍🏻

ChristopherDavenport:
@CharlesAHunt Following up. Just don't want to lose track of this one.

armanbilge (approved):

PR #6037, opened 2022-02-10
- Adds the async timeout as a method param

### LINKED ISSUE #1812: Unable to change the timeout when using mountService in ServletContextSyntax.scala

yuferpegom:
# Adds the async timeout as param
- closes #1812 
- Deprecates `mountService`, clients should use the new `mountRoutes` method with the timeout param
- Deprecates `mountHttpApp` without the timeout param, clients should use `mountHttpApp` with the timeout param

### DIFF
--- a/servlet/src/main/scala/org/http4s/servlet/syntax/ServletContextSyntax.scala
+++ b/servlet/src/main/scala/org/http4s/servlet/syntax/ServletContextSyntax.scala
@@ -26,6 +26,7 @@ import org.http4s.syntax.all._
 
 import javax.servlet.ServletContext
 import javax.servlet.ServletRegistration
+import scala.concurrent.duration.Duration
 
 trait ServletContextSyntax {
   implicit def ToServletContextOps(self: ServletContext): ServletContextOps =
@@ -38,23 +39,43 @@ final class ServletContextOps private[syntax] (val self: ServletContext) extends
     *
     * Assumes non-blocking servlet IO is available, and thus requires at least Servlet 3.1.
     */
-  def mountService[F[_]: Async](
+  @deprecated("Use mountRoutes instead", "1.0.0-M32")
+  private[servlet] def mountService[F[_]: Async](
       name: String,
       service: HttpRoutes[F],
       mapping: String = "/*",
       dispatcher: Dispatcher[F],
   ): ServletRegistration.Dynamic =
-    mountHttpApp(name, service.orNotFound, mapping, dispatcher)
+    mountHttpApp(name, service.orNotFound, mapping, dispatcher, defaults.ResponseTimeout)
+
+  def mountRoutes[F[_]: Async](
+      name: String,
+      service: HttpRoutes[F],
+      mapping: String = "/*",
+      dispatcher: Dispatcher[F],
+      asyncTimeout: Duration = defaults.ResponseTimeout,
+  ): ServletRegistration.Dynamic =
+    mountHttpApp(name, service.orNotFound, mapping, dispatcher, asyncTimeout)
+
+  @deprecated("Use mountHttpApp with async timeout param instead", "1.0.0-M32")
+  private[servlet] def mountHttpApp[F[_]: Async](
+      name: String,
+      service: HttpApp[F],
+      mapping: String,
+      dispatcher: Dispatcher[F],
+  ): ServletRegistration.Dynamic =
+    mountHttpApp(name, service, mapping, dispatcher, defaults.ResponseTimeout)
 
   def mountHttpApp[F[_]: Async](
       name: String,
       service: HttpApp[F],
       mapping: String = "/*",
       dispatcher: Dispatcher[F],
+      asyncTimeout: Duration = defaults.ResponseTimeout,
   ): ServletRegistration.Dynamic = {
     val servlet = new AsyncHttp4sServlet(
       service = service,
-      asyncTimeout = defaults.ResponseTimeout,
+      asyncTimeout = asyncTimeout,
       servletIo = NonBlockingServletIo(DefaultChunkSize),
       serviceErrorHandler = DefaultServiceErrorHandler[F],
       dispatcher,
--- a/servlet/src/main/scala/org/http4s/servlet/syntax/ServletContextSyntax.scala
+++ b/servlet/src/main/scala/org/http4s/servlet/syntax/ServletContextSyntax.scala
@@ -40,7 +40,7 @@ final class ServletContextOps private[syntax] (val self: ServletContext) extends
     * Assumes non-blocking servlet IO is available, and thus requires at least Servlet 3.1.
     */
   @deprecated("Use mountRoutes instead", "0.23.11")
-  private[servlet] def mountService[F[_]: Async](
+  def mountService[F[_]: Async](
       name: String,
       service: HttpRoutes[F],
       mapping: String = "/*",
--- a/servlet/src/main/scala/org/http4s/servlet/syntax/ServletContextSyntax.scala
+++ b/servlet/src/main/scala/org/http4s/servlet/syntax/ServletContextSyntax.scala
@@ -88,24 +88,6 @@ final class ServletContextOps private[syntax] (val self: ServletContext) extends
   }
 }
 
-object ServletContextOps {
-
-  @deprecated("Preserved for binary compatibility", "0.23.11")
-  def `mountHttpApp$extension`[F[_]: Async](
-      servlet: ServletContext,
-      name: String,
-      service: HttpApp[F],
-      mapping: String,
-      dispatcher: Dispatcher[F],
-  ): ServletRegistration.Dynamic =
-    (new ServletContextOps(servlet)).mountHttpApp(
-      name,
-      service,
-      mapping,
-      dispatcher,
-      defaults.ResponseTimeout,
-    )
-
-}
+object ServletContextOps extends ServletContextOpsCompanionCompat
 
 object servletContext extends ServletContextSyntax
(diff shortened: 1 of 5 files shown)


yuferpegom:
Not sure how to fix the binary compatibility issue as it expect to have a `mountHttpApp` method with a `mapping` param having a default value.

servlet/src/main/scala/org/http4s/servlet/syntax/ServletContextSyntax.scala
```diff
@@ -38,23 +39,43 @@ final class ServletContextOps private[syntax] (val self: ServletContext) extends
...
+      name: String,
+      service: HttpRoutes[F],
+      mapping: String = "/*",
+      dispatcher: Dispatcher[F],
+      asyncTimeout: Duration = defaults.ResponseTimeout,
+  ): ServletRegistration.Dynamic =
+    mountHttpApp(name, service.orNotFound, mapping, dispatcher, asyncTimeout)
+
+  @deprecated("Use mountHttpApp with async timeout param instead", "1.0.0-M32")
+  private[servlet] def mountHttpApp[F[_]: Async](
```
armanbilge, line 61: I'm not familiar with this MiMa issue but I don't think it relates to the default method. Try keeping the deprecated method public?
```suggestion
  def mountHttpApp[F[_]: Async](
```

servlet/src/main/scala/org/http4s/servlet/syntax/ServletContextSyntax.scala
```diff
@@ -38,23 +39,43 @@ final class ServletContextOps private[syntax] (val self: ServletContext) extends
     *
     * Assumes non-blocking servlet IO is available, and thus requires at least Servlet 3.1.
     */
-  def mountService[F[_]: Async](
+  @deprecated("Use mountRoutes instead", "1.0.0-M32")
+  private[servlet] def mountService[F[_]: Async](
```
danicheg, line 43: I think we shouldn't make this private

servlet/src/main/scala/org/http4s/servlet/syntax/ServletContextSyntax.scala
```diff
@@ -38,23 +39,43 @@ final class ServletContextOps private[syntax] (val self: ServletContext) extends
     *
     * Assumes non-blocking servlet IO is available, and thus requires at least Servlet 3.1.
     */
-  def mountService[F[_]: Async](
+  @deprecated("Use mountRoutes instead", "1.0.0-M32")
```
danicheg, line 42: Current PR is about to merge to `series/0.23`, so the version should be `0.23.11`
```suggestion
  @deprecated("Use mountRoutes instead", "0.23.11")
```

servlet/src/main/scala/org/http4s/servlet/syntax/ServletContextSyntax.scala
```diff
@@ -38,23 +39,43 @@ final class ServletContextOps private[syntax] (val self: ServletContext) extends
...
+  def mountRoutes[F[_]: Async](
+      name: String,
+      service: HttpRoutes[F],
+      mapping: String = "/*",
+      dispatcher: Dispatcher[F],
+      asyncTimeout: Duration = defaults.ResponseTimeout,
+  ): ServletRegistration.Dynamic =
+    mountHttpApp(name, service.orNotFound, mapping, dispatcher, asyncTimeout)
+
+  @deprecated("Use mountHttpApp with async timeout param instead", "1.0.0-M32")
```
danicheg, line 60:
```suggestion
  @deprecated("Use mountHttpApp with async timeout param instead", "0.23.11")
```

servlet/src/main/scala/org/http4s/servlet/syntax/ServletContextSyntax.scala
yuferpegom: Already tried it, same error. The error is saying that it is expecting to find a method that doesn't require the `mapping` field to be sent, which I think makes sense since that is a param having a default value so before this change, users were able to call this method without including mapping param which is not the case anymore.

servlet/src/main/scala/org/http4s/servlet/syntax/ServletContextSyntax.scala
yuferpegom: Yeah, I wasn't sure about this one but it is a good practice to make deprecated methods private. It will break source compatibility but it is not an issue for binary compatibility.

servlet/src/main/scala/org/http4s/servlet/syntax/ServletContextSyntax.scala
armanbilge:
```
[error] servlet: Failed binary compatibility check against org.http4s:http4s-servlet_2.12:0.23.10 (e:info.apiURL=https://http4s.org/v0.23/api, e:info.versionScheme=early-semver)! Found 2 potential problems
[error]  * extension static method mountHttpApp$extension(javax.servlet.ServletContext,java.lang.String,cats.data.Kleisli,java.lang.String,cats.effect.std.Dispatcher,cats.effect.kernel.Async)javax.servlet.ServletRegistration#Dynamic in class org.http4s.servlet.syntax.ServletContextOps does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("org.http4s.servlet.syntax.ServletContextOps.mountHttpApp$extension")
[error]  * extension method mountHttpApp$extension(javax.servlet.ServletContext,java.lang.String,cats.data.Kleisli,java.lang.String,cats.effect.std.Dispatcher,cats.effect.kernel.Async)javax.servlet.ServletRegistration#Dynamic in object org.http4s.servlet.syntax.ServletContextOps does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("org.http4s.servlet.syntax.ServletContextOps.mountHttpApp$extension")
```

IIUC what this error is saying is that it can no longer find a method with the following signature:
```scala
def mountHttpApp$extension(
  javax.servlet.ServletContext,
  java.lang.String,
  cats.data.Kleisli,
  java.lang.String,
  cats.effect.std.Dispatcher,
  cats.effect.kernel.Async
): javax.servlet.ServletRegistration#Dynamic
```

What you have is:
```scala
  private[servlet] def mountHttpApp[F[_]: Async](
      name: String,
      service: HttpApp[F],
      mapping: String,
      dispatcher: Dispatcher[F],
  ): ServletRegistration.Dynamic
```

Which to me, look the same, assuming that the implicit `Async` is added to the end and the `ServletContext` from the `AnyVal` is added at the beginning. So personally I am baffled.

AFAICT it's not saying anything about default parameters, in my experience those are encoded differently.

armanbilge: Maybe try making the `Async` no longer a context bound but instead add it to the end of the argument list, after the dispatcher?

yuferpegom: I tried this way (let me know if that is what you meant):

```Scala
private[servlet] def mountHttpApp[F[_]](
      name: String,
      service: HttpApp[F],
      mapping: String,
      dispatcher: Dispatcher[F],
      asyncParam: Async[F]
  ): ServletRegistration.Dynamic = {
    implicit val async = asyncParam
    mountHttpApp(name, service, mapping, dispatcher, defaults.ResponseTimeout)
  }
```
Didn't work either.

armanbilge: Yep, that's exactly what I meant. I am truly baffled! I might fiddle with this myself largely out of curiosity 😆

servlet/src/main/scala/org/http4s/servlet/syntax/ServletContextSyntax.scala
```diff
@@ -88,4 +88,24 @@ final class ServletContextOps private[syntax] (val self: ServletContext) extends
   }
 }
 
+object ServletContextOps {
+
+  @deprecated("Preserved for binary compatibility", "0.23.11")
+  def `mountHttpApp$extension`[F[_]: Async](
```
armanbilge, line 94: We could consider making this `private[syntax]`. It would break bincompat for the `static` forwarder, but usually these are only used by Java users (not other Scala users).

armanbilge: Actually, we only need this for 2.12.

servlet/src/main/scala/org/http4s/servlet/syntax/ServletContextSyntax.scala
armanbilge: I'm 👎 on breaking source-compatibility in a patch release. It doesn't follow our versioning scheme; to break source-compatibility requires a minor version bump i.e. http4s 0.24.x.
> you MUST increment the MINOR version if backward source compatibility is broken

https://docs.scala-lang.org/overviews/core/binary-compatibility-for-library-authors.html#recommended-versioning-scheme

rossabaker: We have broken source compatibility in patches before, but it should never be done without urgent reason.  I agree with deprecation but not privatization.

rossabaker:
I'm happy with this as soon as `mountService` is deprivatized.

yuferpegom:
Done, thanks to everyone for your kind help!

armanbilge (approved):

ChristopherDavenport (approved):

PR #6039, opened 2022-02-14
Changes to scaladoc comments in Client

mcarolan:
Hi,

This is a draft PR to see if there is appetite for improvement to the scaladoc comments in the `org.http4s.client.Client` trait.

I've done a few changes to some of the scaladoc comments in this draft, and want a feel on whether this is the right direction of travel, any early feedback or thoughts about whether doc improvements are better done elsewhere.

The changes were based on the behaviour of `DefaultClient`

Motivation
===

As a user of `org.http4s.client.Client` it's not clear what the behaviour of some methods are without looking at the implementation.

E.g. for `expectOption`:
* Does this return a `Some` on successful response code class and a `None` for all other responses?
* Or is it the actual behaviour of `Some` for successful response code classes, None for Not Found or Gone, and errors indicated for all other responses
* Does it follow redirects on 3xx response?

Suggestion
===
Expanding/adding scaladoc comments in some places, at risk of being a bit verbose

### DIFF
--- a/client/shared/src/main/scala/org/http4s/client/Client.scala
+++ b/client/shared/src/main/scala/org/http4s/client/Client.scala
@@ -78,14 +78,30 @@ trait Client[F[_]] {
     */
   def stream(req: Request[F]): Stream[F, Response[F]]
 
+  /** Submits a request.
+   *  For response codes -
+   *  Informational (1xx):
+   *  Successful    (2xx): decodes the response body using d, indicates failure with [[DecodeFailure]] or returns the decoded body
+   *  Redirection   (3xx): same behaviour as above, N.B. does not perform additional requests to follow redirect
+   *  Client error  (4xx):
+   *  Server error  (5xx): applies onError to the response and indicates failure with the [[Throwable]]
+   *
+   *  The underlying HTTP connection is closed at the completion of the decoding.
+   */
   def expectOr[A](req: Request[F])(onError: Response[F] => F[Throwable])(implicit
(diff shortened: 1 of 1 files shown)


armanbilge:
Thanks for taking this on! Personally I'm 👍 to the direction of travel here, organizing by status code seems very good.

It would be ideal to target these changes to the series/0.23 branch so that they can find their way into a stable release sooner :)

client/shared/src/main/scala/org/http4s/client/Client.scala
```diff
@@ -78,14 +78,30 @@ trait Client[F[_]] {
...
 
+  /** Submits a request.
+   *  For response codes -
+   *  Informational (1xx):
+   *  Successful    (2xx): decodes the response body using d, indicates failure with [[DecodeFailure]] or returns the decoded body
+   *  Redirection   (3xx): same behaviour as above, N.B. does not perform additional requests to follow redirect
+   *  Client error  (4xx):
+   *  Server error  (5xx): applies onError to the response and indicates failure with the [[Throwable]]
+   *
+   *  The underlying HTTP connection is closed at the completion of the decoding.
```
rossabaker, line 89: Note: it's released to the pool, not necessarily closed.

rossabaker (commented):
I like the direction.

Are the colons "fall throughs"?  Like 4xx is treated like 5xx?  That wasn't obvious to me from the format.

PR #6047, opened 2022-02-15
Add Additional ErrorHandling Options

ChristopherDavenport:
- Custom Recovery so we can expose a direct function folks can use rather than building their own, even if its not very complex
- Recovery Options - Both for all errors and for message failures allowing people the option to mix that in.

### DIFF
--- a/server/shared/src/main/scala/org/http4s/server/middleware/ErrorHandling.scala
+++ b/server/shared/src/main/scala/org/http4s/server/middleware/ErrorHandling.scala
@@ -61,7 +61,7 @@ object ErrorHandling {
         http: Kleisli[F, Request[G], Response[G]]
     ): Kleisli[F, Request[G], Response[G]] =
       Kleisli { (a: Request[G]) =>
-        http.run(a).recover(totalRecover(a.httpVersion))
+        http.run(a).handleError(totalRecover(a.httpVersion))
       }
 
     def messageFailure[F[_]: MonadThrow, G[_], A](
@@ -79,7 +79,7 @@ object ErrorHandling {
 
     def totalRecover[F[_], G[_]](
         httpVersion: HttpVersion
-    ): PartialFunction[Throwable, Response[G]] = {
+    ): Throwable => Response[G] = {
       case m: MessageFailure => m.toHttpResponse[G](httpVersion)
       case _ =>
         Response(
--- a/server/shared/src/main/scala/org/http4s/server/middleware/ErrorHandling.scala
+++ b/server/shared/src/main/scala/org/http4s/server/middleware/ErrorHandling.scala
@@ -21,6 +21,8 @@ import cats._
 import cats.data.Kleisli
 import cats.syntax.all._
 import org.http4s._
+import org.http4s.headers._
+import org.typelevel.ci._
 
 object ErrorHandling {
   def apply[F[_], G[_]](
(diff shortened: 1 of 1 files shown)


server/shared/src/main/scala/org/http4s/server/middleware/ErrorHandling.scala
```diff
@@ -42,4 +44,53 @@ object ErrorHandling {
...
+      }
+
+    def messageFailure[F[_]: MonadThrow, G[_], A](
+        http: Kleisli[F, Request[G], Response[G]]
+    ): Kleisli[F, Request[G], Response[G]] =
+      Kleisli { (a: Request[G]) =>
+        http.run(a).recover(messageFailureRecover(a.httpVersion))
+      }
+
+    def messageFailureRecover[F[_], G[_]](
```
danicheg, line 74:
```suggestion
    def messageFailureRecover[G[_]](
```

server/shared/src/main/scala/org/http4s/server/middleware/ErrorHandling.scala
```diff
@@ -42,4 +44,53 @@ object ErrorHandling {
...
+        httpVersion: HttpVersion
+    ): PartialFunction[Throwable, Response[G]] = { case m: MessageFailure =>
+      m.toHttpResponse[G](httpVersion)
+    }
+
+    def totalRecover[F[_], G[_]](
```
danicheg, line 80:
```suggestion
    def totalRecover[G[_]](
```

server/shared/src/main/scala/org/http4s/server/middleware/ErrorHandling.scala
```diff
@@ -42,4 +44,53 @@ object ErrorHandling {
 
   def httpApp[F[_]: MonadThrow](httpApp: HttpApp[F]): HttpApp[F] =
     apply(httpApp)
+
+  object Custom {
+    def handleErrorWith[F[_]: MonadThrow, G[_], A](
+        pf: PartialFunction[Throwable, F[Response[G]]],
+        http: Kleisli[F, A, Response[G]],
```
danicheg, line 51: It would be handy, if we will have a contract like this:
```scala
def handleErrorWith[F[_]: MonadThrow, G[_], A](
  http: Kleisli[F, A, Response[G]])(pf: PartialFunction[Throwable, F[Response[G]]])
): Kleisli[F, A, Response[G]]
```
Then one may it use it like:
```scala
Custom.handleErrorWith(http) {
  case _ => ???
}

server/shared/src/main/scala/org/http4s/server/middleware/ErrorHandling.scala
danicheg: Also curious - should `messageFailureRecover` and `totalRecover` be a public API? They seem like quite specific things.

danicheg (approved):
Quite useful things 👍🏻

server/shared/src/main/scala/org/http4s/server/middleware/ErrorHandling.scala
ChristopherDavenport: Adjusted for the better ergonomics here.

rossabaker (approved):
Some scaladoc would help here.  I didn't intuit what they meant until I read the code, though it's explained well enough in the PR comment.  Not a blocker.

PR #6049, opened 2022-02-16
Make it possible to extract Path from Request.pathInfo

hamnis:
Sometimes it's useful to extract the Path directly from the request and work on that.

When using http4s-directives, we never want to handle Methods using pattern matching, since we care about correct HTTP response codes.

This makes foot-shooting possible, but that was possible anyway.


hamnis:
When thinking about it, this will not handle Path.Root, so never mind.

PR #6056, opened 2022-02-17
Merge 0.23 -> main

armanbilge:
In the process of this merge, I also removed the ember.js websocket server test. It requires Scala.js bundler, npm dependencies, and gigantic lockfiles, which we already exorcised from 0.23 in https://github.com/http4s/http4s/pull/6009. The websocket server implementation is pure Scala now anyway, so it's fairly unlikely to break on JS, and I doubt anyone would care if it did. When we get an ember websocket client, we can use that for the test.


website/src/hugo/content/default.template.html
```diff
@@ -66,8 +62,25 @@
         ]
       }
 
+<<<<<<< HEAD
```
armanbilge, line 65: Woops

ember-server/jvm/src/test/scala/org/http4s/ember/server/EmberServerWebSocketSuite.scala
```diff
@@ -30,18 +40,45 @@ import java.net.URI
...
+  val serverResource: Resource[IO, Server] =
+    EmberServerBuilder
+      .default[IO]
+      .withPort(port"0")
+      .withHttpWebSocketApp(service[IO])
+      .build
+
+  def fixture = (ResourceFixture(serverResource), dispatcher).mapN(FunFixture.map2(_, _))
+
+  sealed case class Client(
```
danicheg, line 75: Is `sealed` goes from Scalafix?

danicheg (approved):

ember-server/jvm/src/test/scala/org/http4s/ember/server/EmberServerWebSocketSuite.scala
armanbilge: Yes?

danicheg: Probably this was discussed already (remembering something), but `sealed` for `case class` just strikes the eye. It'd be nice if Scalafix could be configured to skip the rule for tests.

armanbilge: Yeah, it was scalafix in https://github.com/http4s/http4s/commit/ab458086352c18cdebd5a3345423cdc829f2b414#diff-4fc59bb173a3f7944063507512a3625fd3e6baa625b5f9544c149d13cf942c4c.

PR #6057, opened 2022-02-17
Routing on variable segments

teigen:
Router and ContextRouter are enhanced with support for routing on
variable path segments through the new methods `dynamic` and `of`.
Static and dynamic routing is unified through a new type
`{Router, ContextRouter}.Routable`.
Capturing and transforming a segment of the path is implemented as a
tiny middleware `{Router, ContextRouter}.Segment`.

As there are no changes to existing methods, this should be
backwards compatible. `of` and `dynamic` are source compatible with `apply` and `define`,
with only slightly worse type-inference for static cases), and
could replace `apply` and `define` at the next point of breaking binary compatibility.

The routing is always done static, with longest match first, and then
dynamic in the order they are defined. (In other words, same as it used to be, but
with the dynamic ones at the end)


hamnis (approved):
Very cool. This makes the Router very much more powerful and useful.

We can now build helpers to inspect the Routables for various purposes.

server/src/main/scala/org/http4s/server/ContextRouter.scala
```diff
@@ -16,15 +16,68 @@
...
+    def ->(routes: ContextRoutes[B, F]): Routable[F, A] =
+      Routable.Dynamic(this, routes)
+  }
+
+  object Routable {
+    final case class Static[F[_], A](tupled: (String, ContextRoutes[A, F])) extends Routable[F, A]
+    final case class Dynamic[F[_], A, B](segment: Segment[F, A, B], routes: ContextRoutes[B, F])
+        extends Routable[F, A]
+
+    implicit def tuple[F[_], A](tupled: (String, ContextRoutes[A, F])): Routable[F, A] =
```
danicheg, line 76: If we delete this, would be that worsen ergonomics much? I'm just trying to stay away from implicits conversions when possible.

danicheg (commented):
Nice feature 👍🏻. Few minor suggestions:

server/src/main/scala/org/http4s/server/ContextRouter.scala
```diff
@@ -16,15 +16,68 @@
...
+          )
+        } yield response
+      }
+
+    def ->(routes: ContextRoutes[B, F]): Routable[F, A] =
+      Routable.Dynamic(this, routes)
+  }
+
+  object Routable {
+    final case class Static[F[_], A](tupled: (String, ContextRoutes[A, F])) extends Routable[F, A]
```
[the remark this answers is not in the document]
teigen: I initially implemented this as 2 parameters, but given that it starts its life as a tuple from `Routable.tuple`, and is used as a tuple when passed to `define`, it seemed like the simplest thing (both for me and the cpu).
But I agree it looks a bit silly in a public api, so perhaps both Static and Dynamic should be made more private instead, as they both have sensible public constructors available in `Segment ->` and `Routable.tuple` ?

I really have no strong opinions on this, but perhaps leaning slightly towards making the implementations private the more I think about it. Either way, I'm happy to change this to either 2 parameters or making the implementations private.

server/src/main/scala/org/http4s/server/ContextRouter.scala
teigen: Yes, If we remove that, then we would have a very different user experience.
I'm not a fan of adding implicit conversions, but in this case it provides a very smooth migration path from the existing static-only routers

rossabaker:
@danicheg Are you happy with this, or do you want to hold for your concerns?  I'd like to do a release ASAP, though as I say, we won't run out of numbers if we follow up with another.

danicheg:
@rossabaker Actually, I'm happy with this PR as is. My suggestions were totally minor, as I pointed out. If no one else sees benefits from it, let's go on and merge this.

rossabaker:
Okay, we just need to get the merge conflicts fixed.  A bunch of test variables became private.

PR #6064, opened 2022-02-19
Cross origin embedder policy header model

bednam:
#5010


core/src/main/scala/org/http4s/headers/`Cross-Origin-Embedder-Policy`.scala
```diff
@@ -0,0 +1,48 @@
...
+ */
+
+package org.http4s
+package headers
+
+import cats.parse.Parser
+import cats.parse.Parser.string
+import org.typelevel.ci._
+
+sealed abstract class `Cross-Origin-Embedder-Policy`(val value: String)
```
diesalbla, line 24: Would it be possible to add some Scala-Docs with a reference (link) to the RFC, or equivalent authoritative document, where this Header is specified?

PR #6077, opened 2022-03-01
Exposing callable parameters for DigestAuth.challenge

### LINKED ISSUE #6068: Currently impossible to use DigestAuth.challenge due to private argument types
In attempting to integrate with the existing http4s authentication middleware convention, I discovered that [`DigestAuth.challenge`](https://github.com/http4s/http4s/blob/1d25fdd8755e9af604a7a45116687a479f82d8b0/server/src/main/scala/org/http4s/server/middleware/authentication/DigestAuth.scala#L80-L82) is public, but it's effectively impossible to use due to one of its arguments, [`NonceKeeper`](https://github.com/http4s/http4s/blob/1d25fdd8755e9af604a7a45116687a479f82d8b0/server/src/main/scala/org/http4s/server/middleware/authentication/NonceKeeper.scala#L22-L43), being package private.

Is this something that could be changed? I see two options:

- Make `NonceKeeper` public
- Adjust `DigestAuth.challenge`'s arguments to mirror `DigestAuth.apply`, specifically `nonceCleanupInterval`, `nonceStaleTime`, and `nonceBits`, so that `challenge` constructs the package-private `NonceKeeper` instead of `apply`

I'm happy to proceed with either of these, or another suggestion if you've got one, just looking for guidance before I start.

Thank you

blast-hardcheese:
NonceKeeper is package-private, so there's no way to construct one,
meaning there's no way to call `challenge`.

Mirror the same parameters from `apply`, and maintain the previous
`challenge` method for bincompat.

I'm happy to make changes if desired.

Thank you

### DIFF
--- a/server/src/main/scala/org/http4s/server/middleware/authentication/DigestAuth.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/authentication/DigestAuth.scala
@@ -68,15 +68,26 @@ object DigestAuth {
       nonceCleanupInterval: Duration = 1.hour,
       nonceStaleTime: Duration = 1.hour,
       nonceBits: Int = 160,
-  ): AuthMiddleware[F, A] = {
-    val nonceKeeper =
-      new NonceKeeper(nonceStaleTime.toMillis, nonceCleanupInterval.toMillis, nonceBits)
-    challenged(challenge(realm, store, nonceKeeper))
-  }
+  ): AuthMiddleware[F, A] =
+    challenged(challenge(realm, store, nonceCleanupInterval, nonceStaleTime, nonceBits))
 
   /** Side-effect of running the returned task: If req contains a valid
     * AuthorizationHeader, the corresponding nonce counter (nc) is increased.
     */
+  @annotation.nowarn("msg=Maintaining for ABI compatibility")
+  def challenge[F[_], A](
+      realm: String,
+      store: AuthenticationStore[F, A],
+      nonceCleanupInterval: Duration = 1.hour,
+      nonceStaleTime: Duration = 1.hour,
+      nonceBits: Int = 160,
+  )(implicit F: Async[F]): Kleisli[F, Request[F], Either[Challenge, AuthedRequest[F, A]]] = {
+    val nonceKeeper =
+      new NonceKeeper(nonceStaleTime.toMillis, nonceCleanupInterval.toMillis, nonceBits)
+    challenge[F, A](realm, store, nonceKeeper)
+  }
+
+  @deprecated("Maintaining for ABI compatibility", "0.23.10")
   def challenge[F[_], A](realm: String, store: AuthenticationStore[F, A], nonceKeeper: NonceKeeper)(
       implicit F: Sync[F]
   ): Kleisli[F, Request[F], Either[Challenge, AuthedRequest[F, A]]] =
--- a/server/src/main/scala/org/http4s/server/middleware/authentication/DigestAuth.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/authentication/DigestAuth.scala
@@ -102,8 +102,12 @@ object DigestAuth {
   /** Side-effect of running the returned task: If req contains a valid
     * AuthorizationHeader, the corresponding nonce counter (nc) is increased.
     */
-  private[authentication] def challenge[F[_], A](realm: String, store: AuthenticationStore[F, A], nonceKeeper: NonceKeeper)(
-      implicit F: Sync[F]
+  private[authentication] def challenge[F[_], A](
+      realm: String,
+      store: AuthenticationStore[F, A],
+      nonceKeeper: NonceKeeper,
+  )(implicit
+      F: Sync[F]
   ): Kleisli[F, Request[F], Either[Challenge, AuthedRequest[F, A]]] =
     Kleisli { req =>
       def paramsToChallenge(params: Map[String, String]) =
--- a/server/src/main/scala/org/http4s/server/middleware/authentication/DigestAuth.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/authentication/DigestAuth.scala
@@ -30,6 +30,7 @@ import org.http4s.headers._
 
 import java.math.BigInteger
 import java.util.Date
+import scala.annotation.nowarn
 import scala.concurrent.duration._
 
 /** Provides Digest Authentication from RFC 2617.
@@ -62,6 +63,7 @@ object DigestAuth {
     *                       purposes anymore).
     * @param nonceBits The number of random bits a nonce should consist of.
     */
+  @nowarn
   def apply[F[_]: Sync, A](
       realm: String,
       store: AuthenticationStore[F, A],
@@ -87,6 +89,7 @@ object DigestAuth {
     *                       purposes anymore).
     * @param nonceBits The number of random bits a nonce should consist of.
     */
+  @nowarn
   def challenge[F[_], A](
       realm: String,
       store: AuthenticationStore[F, A],
@@ -102,7 +105,11 @@ object DigestAuth {
   /** Side-effect of running the returned task: If req contains a valid
     * AuthorizationHeader, the corresponding nonce counter (nc) is increased.
     */
-  private[authentication] def challenge[F[_], A](
+  @deprecated(
+    "Calling challenge is side-effecting, please use the F[_]-suspended variant",
+    "0.22.11",
+  )
+  def challenge[F[_], A](
       realm: String,
       store: AuthenticationStore[F, A],
       nonceKeeper: NonceKeeper,
--- a/server/src/main/scala/org/http4s/server/middleware/authentication/DigestAuth.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/authentication/DigestAuth.scala
@@ -68,29 +68,41 @@ object DigestAuth {
       nonceCleanupInterval: Duration = 1.hour,
       nonceStaleTime: Duration = 1.hour,
       nonceBits: Int = 160,
-  ): AuthMiddleware[F, A] =
-    challenged(challenge(realm, store, nonceCleanupInterval, nonceStaleTime, nonceBits))
+  ): AuthMiddleware[F, A] = {
+    val nonceKeeper =
+      new NonceKeeper(nonceStaleTime.toMillis, nonceCleanupInterval.toMillis, nonceBits)
+    challenged(challenge(realm, store, nonceKeeper))
+  }
 
-  /** Side-effect of running the returned task: If req contains a valid
-    * AuthorizationHeader, the corresponding nonce counter (nc) is increased.
+  /** Similar to [[apply]], but exposing the underlying [[challenge]]
+    * [[cats.data.Kleisli]] instead of an entire [[AuthMiddleware]]
+    *
+    * @param realm The realm used for authentication purposes.
+    * @param store A partial function mapping (realm, user) to the
+    *              appropriate password.
+    * @param nonceCleanupInterval Interval (in milliseconds) at which stale
+    *                             nonces should be cleaned up.
+    * @param nonceStaleTime Amount of time (in milliseconds) after which a nonce
+    *                       is considered stale (i.e. not used for authentication
+    *                       purposes anymore).
+    * @param nonceBits The number of random bits a nonce should consist of.
     */
-  @annotation.nowarn("msg=Maintaining for ABI compatibility")
   def challenge[F[_], A](
       realm: String,
       store: AuthenticationStore[F, A],
       nonceCleanupInterval: Duration = 1.hour,
       nonceStaleTime: Duration = 1.hour,
       nonceBits: Int = 160,
-  )(implicit F: Sync[F]): Kleisli[F, Request[F], Either[Challenge, AuthedRequest[F, A]]] =
-    Kleisli { req =>
-      F.delay(new NonceKeeper(nonceStaleTime.toMillis, nonceCleanupInterval.toMillis, nonceBits))
-        .flatMap { nonceKeeper =>
-          challenge[F, A](realm, store, nonceKeeper).apply(req)
-        }
-    }
+  )(implicit F: Sync[F]): F[Kleisli[F, Request[F], Either[Challenge, AuthedRequest[F, A]]]] =
+    F.delay(new NonceKeeper(nonceStaleTime.toMillis, nonceCleanupInterval.toMillis, nonceBits))
+      .map { nonceKeeper =>
+        challenge[F, A](realm, store, nonceKeeper)
+      }
 
-  @deprecated("Maintaining for ABI compatibility", "0.22.12")
-  def challenge[F[_], A](realm: String, store: AuthenticationStore[F, A], nonceKeeper: NonceKeeper)(
+  /** Side-effect of running the returned task: If req contains a valid
+    * AuthorizationHeader, the corresponding nonce counter (nc) is increased.
+    */
+  private def challenge[F[_], A](realm: String, store: AuthenticationStore[F, A], nonceKeeper: NonceKeeper)(
       implicit F: Sync[F]
   ): Kleisli[F, Request[F], Either[Challenge, AuthedRequest[F, A]]] =
     Kleisli { req =>
--- a/server/src/main/scala/org/http4s/server/middleware/authentication/DigestAuth.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/authentication/DigestAuth.scala
@@ -81,11 +81,13 @@ object DigestAuth {
       nonceCleanupInterval: Duration = 1.hour,
       nonceStaleTime: Duration = 1.hour,
       nonceBits: Int = 160,
-  )(implicit F: Async[F]): Kleisli[F, Request[F], Either[Challenge, AuthedRequest[F, A]]] = {
-    val nonceKeeper =
-      new NonceKeeper(nonceStaleTime.toMillis, nonceCleanupInterval.toMillis, nonceBits)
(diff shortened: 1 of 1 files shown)


armanbilge:
Thanks for digging into this! Since you've done the work to preserve bincompat (appreciated!) would you be able to retarget this at a stable branch? `series/0.23` unless you are on `series/0.22`.

blast-hardcheese:
@armanbilge It would be helpful to have this on 0.22, I'm happy to open a PR against 0.23 as well if that's how things work around here 🙂

armanbilge:
No, series/0.22 is perfect. Lower branches are merged up so it'll find its way to series/0.23 and main :)

server/src/main/scala/org/http4s/server/middleware/authentication/DigestAuth.scala
```diff
@@ -68,15 +68,26 @@ object DigestAuth {
...
+      nonceCleanupInterval: Duration = 1.hour,
+      nonceStaleTime: Duration = 1.hour,
+      nonceBits: Int = 160,
+  )(implicit F: Async[F]): Kleisli[F, Request[F], Either[Challenge, AuthedRequest[F, A]]] = {
+    val nonceKeeper =
+      new NonceKeeper(nonceStaleTime.toMillis, nonceCleanupInterval.toMillis, nonceBits)
+    challenge[F, A](realm, store, nonceKeeper)
+  }
+
+  @deprecated("Maintaining for ABI compatibility", "0.23.10")
```
armanbilge, line 90:
```suggestion
  @deprecated("Maintaining for ABI compatibility", "0.22.12")
```

server/src/main/scala/org/http4s/server/middleware/authentication/DigestAuth.scala
```diff
@@ -68,15 +68,26 @@ object DigestAuth {
...
+  @annotation.nowarn("msg=Maintaining for ABI compatibility")
+  def challenge[F[_], A](
+      realm: String,
+      store: AuthenticationStore[F, A],
+      nonceCleanupInterval: Duration = 1.hour,
+      nonceStaleTime: Duration = 1.hour,
+      nonceBits: Int = 160,
+  )(implicit F: Async[F]): Kleisli[F, Request[F], Either[Challenge, AuthedRequest[F, A]]] = {
+    val nonceKeeper =
+      new NonceKeeper(nonceStaleTime.toMillis, nonceCleanupInterval.toMillis, nonceBits)
```
armanbilge, line 86: This probably should be wrapped in `F.delay`

armanbilge:
> NonceKeeper is package-private, so there's no way to construct one

Unless there's a need to share a `NonceKeeper` the approach here seems reasonable 👍  Alternatively, if `NonceKeeper` _shouldn't_ be shared, then this is definitely the right strategy :)

server/src/main/scala/org/http4s/server/middleware/authentication/DigestAuth.scala
blast-hardcheese: 🙏 Forgot to update, thank you 🙇

server/src/main/scala/org/http4s/server/middleware/authentication/DigestAuth.scala
```diff
@@ -81,11 +81,13 @@ object DigestAuth {
...
       nonceStaleTime: Duration = 1.hour,
       nonceBits: Int = 160,
-  )(implicit F: Async[F]): Kleisli[F, Request[F], Either[Challenge, AuthedRequest[F, A]]] = {
-    val nonceKeeper =
-      new NonceKeeper(nonceStaleTime.toMillis, nonceCleanupInterval.toMillis, nonceBits)
-    challenge[F, A](realm, store, nonceKeeper)
-  }
+  )(implicit F: Sync[F]): Kleisli[F, Request[F], Either[Challenge, AuthedRequest[F, A]]] =
+    Kleisli { req =>
+      F.delay(new NonceKeeper(nonceStaleTime.toMillis, nonceCleanupInterval.toMillis, nonceBits))
```
armanbilge, line 86: Hmm, sorry, this doesn't seem quite right either, since it would create a new `NonceKeeper` per-request.

So maybe the signature we want is actually `F[Kleisli[F, Request[F], Either[Challenge, AuthedRequest[F, A]]]]`?

Creating the middleware itself is a side-effect.

blast-hardcheese: do'h. Yes. Will patch shortly.

blast-hardcheese: Wait, that won't work, since `F[Kleisli[F, Request[F], Either[Challenge, AuthedRequest[F, A]]]]` will change the result type of the function

blast-hardcheese: OK. I hadn't considered the implications of allocating a new NonceKeeper every request, I'll admit -- I'd gotten a bit too focused on just getting it to compile and the linter to pass. Stepping back, I think the correct approach here is to sacrifice reusability for exposing a new, usable interface to `challenge`, for those that need it.

Consumers of this new function will appreciate the fact that `new NonceKeeper` is suspended in `F.delay` (more predictable tests), while existing users of `DigestAuth.apply` will not be impacted by the change.

server/src/main/scala/org/http4s/server/middleware/authentication/DigestAuth.scala
```diff
@@ -74,10 +74,35 @@ object DigestAuth {
...
+    F.delay(new NonceKeeper(nonceStaleTime.toMillis, nonceCleanupInterval.toMillis, nonceBits))
+      .map { nonceKeeper =>
+        challenge[F, A](realm, store, nonceKeeper)
+      }
+
   /** Side-effect of running the returned task: If req contains a valid
     * AuthorizationHeader, the corresponding nonce counter (nc) is increased.
     */
-  def challenge[F[_], A](realm: String, store: AuthenticationStore[F, A], nonceKeeper: NonceKeeper)(
+  private def challenge[F[_], A](realm: String, store: AuthenticationStore[F, A], nonceKeeper: NonceKeeper)(
```
armanbilge, line 105: Let's keep it package-private so we don't break bincompat, just in case someone was doing something hacky with packages to call it.

```suggestion
  private[authentication] def challenge[F[_], A](realm: String, store: AuthenticationStore[F, A], nonceKeeper: NonceKeeper)(
```

server/src/main/scala/org/http4s/server/middleware/authentication/DigestAuth.scala
armanbilge: Oh, woops, I hadn't noticed the `DigestAuth.apply` signature above with the same problem. I wonder if we can deprecate it and fix it too. It's not just about predictable tests, it's really about correctness.

Sorry this is unraveling into more than a small fix 😁 they always do!

blast-hardcheese:
> > NonceKeeper is package-private, so there's no way to construct one
> 
> Unless there's a need to share a `NonceKeeper` the approach here seems reasonable 👍 Alternatively, if `NonceKeeper` _shouldn't_ be shared, then this is definitely the right strategy :)

I don't believe there's any reason to share the `NonceKeeper`

server/src/main/scala/org/http4s/server/middleware/authentication/DigestAuth.scala
blast-hardcheese: Not a problem! I'm actually really happy to contribute -- this is just an area I haven't used much, so I don't want to do something haphazardly that breaks usability or ergonomics. I'm usually of the mind that if it's there, it's there for a reason, even though you're right... `F[AuthMiddleware[F, A]]` would be more correct.

blast-hardcheese: Given that, how would a backwards-incompatible, type-conflicting change be made? It would be easier if `apply` weren't at fault.

armanbilge: Thanks, glad to hear it! Yes, unfortunately I'm not an expert on this auth stuff either.

> I'm usually of the mind that if it's there, it's there for a reason

For sure, but taking a closer look at `NonceKeeper` (and the commit history) I suspect this code has just fallen behind with the times. For example the use of `synchronized` (which might hard-block) where we probably want a semantically-blocking [semaphore](https://typelevel.org/cats-effect/docs/2.x/concurrency/semaphore).
https://github.com/http4s/http4s/blob/61b48b7cfdacd49031c16b8b0d7e293ea43fa8c0/server/src/main/scala/org/http4s/server/middleware/authentication/NonceKeeper.scala#L74-L79

If you're up for it, what we could do here is re-write a modernized, safe `NonceKeeper`, and build the new API around that. But that may be more than you bargained for!

blast-hardcheese: Ha! Alright -- I'm alright, if y'all are alright keeping me in line.

To set the stage, I've been in akka-land for the past seven years, even if I was using cats I was never using cats-effect. Don't hold back, but please be patient. I'm here to learn as much as to contribute.

I'll take a crack at it and return in the next few days. Thank you for the guidance thus far!

armanbilge: > I've been in akka-land for the past seven years, even if I was using cats I was never using cats-effect.

This was me, one year ago :)

server/src/main/scala/org/http4s/server/middleware/authentication/DigestAuth.scala
```diff
@@ -74,11 +74,40 @@ object DigestAuth {
...
+    * @param store A partial function mapping (realm, user) to the
+    *              appropriate password.
+    * @param nonceCleanupInterval Interval (in milliseconds) at which stale
+    *                             nonces should be cleaned up.
+    * @param nonceStaleTime Amount of time (in milliseconds) after which a nonce
+    *                       is considered stale (i.e. not used for authentication
+    *                       purposes anymore).
+    * @param nonceBits The number of random bits a nonce should consist of.
+    */
+  def challenge[F[_], A](
```
danicheg, line 90: Probably that was addressed already, but why don't we add a public constructor for `NonceKeeper` instead?

armanbilge: That's an option too, although we have to be careful because creating a `NonceKeeper` is a side-effect. I think it depends whether a `NonceKeeper` should ever be shared between auth middlewares or expose any public methods. If not, it's an internal implementation detail.

There's already a signature that works without exposing the `NonceKeeper`, so we can imitate it for `challenge`. We can always expose the `NonceKeeper` in the future if we need to.
https://github.com/http4s/http4s/blob/548d2dd147fb895edcbe19c53af468d5bef590c9/server/src/main/scala/org/http4s/server/middleware/authentication/DigestAuth.scala#L65-L73

blast-hardcheese: My thoughts around this were that exposing `NonceKeeper` at all (currently the whole type is package private) incurred more change to the surface area of the package than just tweaking the arguments to a method of `DigestAuth`.

No need to expose complexity that doesn't benefit the users in some way.

rossabaker:
Is there anything blocking this?  I'm looking to rev up a release train, and stayed out of this one because it already had two capable reviewers.

armanbilge:
Yeah, I think this one is in a weird place. IIRC what's currently here fixes the immediate problem.

However, looking deeper the `NonceKeeper` should probably be reworked to not use `synchronized { ... }` blocks, but this will require some additional constraints.

So if we merge as-is we are basically introducing APIs that we will end up deprecating once this is fixed properly.

blast-hardcheese:
I haven't had time to rework NonceKeeper yet, but since it's private, could we expose the desired public API that NonceKeeper will eventually fill as a stopgap?

blast-hardcheese:
I'll do that early this week as to hopefully hit the release

rossabaker:
I'm queuing up a release now (there's always _something_ in flight), but there's plenty more, including this, to justify another.  I'm not opposed to two releases this week, which would test Arman's new website machinery.

armanbilge:
You're right, we can't change the return type of the existing method. However, we can deprecate it, and add a new method.

The new method can still be called `apply` as long as the signature changes, which I believe it will have to because we will need to change the typeclass constraint.

blast-hardcheese:
@armanbilge Unless I'm missing something, `[F[_]: Sync]()` to `[F[_]](implicit F: Sync[F])` has the same signature:

```scala
  def apply[F[_]: Sync, A](
      realm: String,
      store: AuthenticationStore[F, A],
      nonceCleanupInterval: Duration = 1.hour,
      nonceStaleTime: Duration = 1.hour,
      nonceBits: Int = 160,
  ): AuthMiddleware[F, A] = ???
```
...'s javap:
```
  public <F, A> scala.Function1<cats.data.Kleisli<?, org.http4s.ContextRequest<F, A>, org.http4s.Response<F>>, cats.data.Kleisli<?, org.http4s.Request<F>, org.http4s.Response<F>>> apply(java.lang.String, scala.Function1<java.lang.String, F>, scala.concurrent.duration.Duration, scala.concurrent.duration.Duration, int, cats.effect.Sync<F>);
```

vs

```scala
  def apply[F[_], A](
      realm: String,
      store: AuthenticationStore[F, A],
      nonceCleanupInterval: Duration = 1.hour,
      nonceStaleTime: Duration = 1.hour,
      nonceBits: Int = 160,
  ): F[AuthMiddleware[F, A]] = ???
```
...'s javap:
```
  public <F, A> F apply(java.lang.String, scala.Function1<java.lang.String, F>, scala.concurrent.duration.Duration, scala.concurrent.duration.Duration, int, cats.effect.Sync<F>);
```

Since the only thing that changes is in the return position, I still don't see a way to make this both binary compatible and deprecated with the same name.

Could do `applyF` as a stopgap?

blast-hardcheese:
@armanbilge Alright. `server / mimaReportBinaryIssues` is clear, `applyF` uses `F[_]: Concurrent` to permit future `Semaphore[F]` usage, I think this is good to go.

blast-hardcheese:
Oops. Was relying on `server / Test / compile`, got bit by incremental compilation fail.

blast-hardcheese:
OK. Rebased, `clean ; test` passed, `scalafmtAll`'d, `mimaReportBinaryIssues`, and we're ready to go again.

server/src/main/scala/org/http4s/server/middleware/authentication/DigestAuth.scala
```diff
@@ -62,21 +77,56 @@ object DigestAuth {
...
+    * @param nonceBits The number of random bits a nonce should consist of.
+    */
+  @nowarn
+  def challenge[F[_], A](
+      realm: String,
+      store: AuthenticationStore[F, A],
+      nonceCleanupInterval: Duration = 1.hour,
+      nonceStaleTime: Duration = 1.hour,
+      nonceBits: Int = 160,
+  )(implicit F: Sync[F]): F[Kleisli[F, Request[F], Either[Challenge, AuthedRequest[F, A]]]] =
```
armanbilge, line 117: I think this should also become a `Concurrent` constraint.

armanbilge (commented):
Nice work getting it green!

Personally I'd feel a bit funny shipping this without a matching implementation, since without one it's hard to know for sure if we got the constraints right and/or missed anything else. E.g. the note I made below.

blast-hardcheese:
> Personally I'd feel a bit funny shipping this without a matching implementation

So, given this, are we back at where we started, just exposing a new `challenge` method with callable parameters, pending the `NonceKeeper` rewrite, or are we blocked until that gets done?

armanbilge:
We should let another maintainer weigh in.

`series/0.22` is approaching EOL so maybe it's okay to just eat the caveats and go for the minimum-viable-fix. But I think we should see this fixed properly on `series/0.23` up, at least.

FWIW I don't think it should be too challenging to fix. The only change really is gating the `NonceKeeper` with a semaphore instead of relying on a `ConcurrentMap`.

blast-hardcheese:
To clarify, I actually agree with your objection. I started trying to rework `NonceKeeper`, since that needs to be done regardless. If we _could_ get a callable version of `challenge` in 0.22 that'd be grand, but not if y'all are uncomfortable with it.

blast-hardcheese:
Alright. Rolling this one back to pre `applyF`, as the future commits are in the other draft PR. Presumably the tests will all still pass and we can make some progress here.l Bincompat should still pass, and we don't have the confusion `apply` vs `applyF` to deal with, it's just the surgical exposure of a method that was previously uncallable.

armanbilge:
I'm a bit confused about the relationship between this PR and https://github.com/http4s/http4s/pull/6138. https://github.com/http4s/http4s/pull/6138 is looking promising so I expect you can fix https://github.com/http4s/http4s/issues/6068 in that PR :)

blast-hardcheese:
The intent of this PR is to just get the `challenge` method merged into 0.22 before it gets closed, and the `NonceKeeper` changes would move on to 0.23+

This might be too much to ask, but I figured since we've got direction on the future work, we could get this one moving.

armanbilge:
Gotcha! I think we should be able to land https://github.com/http4s/http4s/pull/6138 in 0.22 :)

blast-hardcheese:
> Gotcha! I think we should be able to land https://github.com/http4s/http4s/pull/6138 in 0.22 :)

Got it, I misunderstood the direction of the conversation, no worries

blast-hardcheese:
#6138 looks like it's in the home stretch, closing this one out.

Issue #6067, opened 2022-02-21
Further relax cookie parsing? 


kailuowang:
Similar to #4003, a double quote in a third-party cookie will break cookie parsing in Http4s. For example, the cookie set by Google's "Sign in with Google" button has double quotes in it and thus silently broke the cookie parsing in Http4s (`request.cookies` becomes empty even though there are other cookies. ) 

In this [discussion](https://groups.google.com/a/chromium.org/g/chromium-discuss/c/T3oHExJMr0M), Ryan from Chromium pointed out that the RFC6265 has two sections specifying the requirements for cookie value. Apparently, the requirements defined in section 5.2 are less strict than section 4.1.1 implemented by Http4s. Thus [Chromum allows](https://source.chromium.org/chromium/chromium/src/+/main:net/cookies/parsed_cookie.cc;drc=1241c4db3549f497bcaa14815cc6b232a32cf534;l=497) all ASCII chars except control characters and `;`

Based on these considerations, is it possible to further relax `RelaxedCookies` to conform to the requirements set by Chromium?

kailuowang:
If we can get some maintainer approval, I am happy to submit a PR. I don't see why http4s should be failing on cookie parsing due to unnecessary enforcement of a poorly specified cookie spec.

kailuowang:
> It's messed up people broke the spec so much they are just editing the spec.

I would say that the web stack (HTTP, HTML, CSS, JS) overall is not as principled as functional programmers are used to.

rossabaker:
There's a PR that tries to provide an ABNF for the more relaxed rules.  It still has some open problems, but worth reading: https://github.com/httpwg/http-extensions/pull/1759

rossabaker:
The Section 4 rules on what we send are SHOULD and the Section 5 rules on what we parse are MUST.  I don't think this is a case of standards vs. practice: they both say we should fix this.

kailuowang:
Thanks @rossabaker ,  
I just noticed that http4s seems to be missing the range `0x80` to `0xFF`. which is specified in both sections. (`0x7F` seems to be excluded in section 4.4.1 but very [ambiguously](https://github.com/httpwg/http-extensions/issues/1399#issuecomment-1058097695) )

rossabaker:
Fixed by #6082.

PR #6079, opened 2022-03-02
More tolerant cookie parsing - fix 6067

kailuowang:
See #6067 for detail
replaces #6078

### DIFF
--- a/core/shared/src/main/scala/org/http4s/ResponseCookie.scala
+++ b/core/shared/src/main/scala/org/http4s/ResponseCookie.scala
@@ -28,9 +28,9 @@ package org.http4s
 
 import cats.parse.Parser
 import cats.parse.Rfc5234
+import org.http4s.internal.parsing.RelaxedCookies
 import org.http4s.internal.parsing.Rfc1034
 import org.http4s.internal.parsing.Rfc2616
-import org.http4s.internal.parsing.RelaxedCookies
 import org.http4s.util.Renderable
 import org.http4s.util.Writer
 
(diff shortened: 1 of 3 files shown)


core/shared/src/main/scala/org/http4s/ResponseCookie.scala
```diff
@@ -104,7 +104,7 @@ object ResponseCookie {
     import Parser.{char, charIn, failWith, ignoreCase, pure}
     import Rfc2616.Rfc1123Date
     import Rfc5234.digit
-    import Rfc6265.{cookieName, cookieValue}
+    import RelaxedCookies.{cookieName, cookieValue}
```
kailuowang, line 107: Before this PR, `ResponseCookie` was using a different parser char set than `RequestCookie`, but sharing the same special double-quoted value parsing code. Now that `ResponseCookie` changes to the new relaxed char set which tolerates double quotes, the double-quoted value parsing code can no longer serve both char sets. 
I think the most straightforward and logical fix is what I did here, changing the `ResponseCookie` to use the same char set as `RequestCookie`.

rossabaker (approved):
I haven't rigorously compared this against the RFC to see what else may be missing, but this seems at worst a step in the right direction.

danicheg (approved):
According to https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#attributes, this looks good to me.

danicheg:
By the way, should we backport this to `0.22`? I think that behavior changes more likely should be backported.

rossabaker:
Yeah, this is a bug fix.  I'll pull it back to 0.22.

rossabaker:
Closed for #6082.

PR #6082, opened 2022-03-03
More tolerant cookie parsing

rossabaker:
Backport of #6079.  All credit to @kailuowang.

### DIFF
--- a/core/src/main/scala/org/http4s/internal/parsing/Cookies.scala
+++ b/core/src/main/scala/org/http4s/internal/parsing/Cookies.scala
@@ -63,6 +63,6 @@ private[http4s] object Rfc6265
 private[http4s] object RelaxedCookies
     extends Cookies(
       cookieOctet = charIn(
-        Set(0x20.toChar to 0x7e.toChar: _*) - 0x3b.toChar
+        Set(0x20.toChar to 0xff.toChar: _*) - 0x3b.toChar
       )
     )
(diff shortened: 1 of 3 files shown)


danicheg (approved):

core/src/main/scala/org/http4s/internal/parsing/Cookies.scala
```diff
@@ -57,15 +56,13 @@ private[http4s] object Rfc6265
...
 private[http4s] object RelaxedCookies
     extends Cookies(
-      // Compared to Rfc6265, also tolerates spaces
       cookieOctet = charIn(
-        Set(0x20.toChar, 0x21.toChar) ++
-          Set(0x23.toChar to 0x2b.toChar: _*) ++
-          Set(0x2d.toChar to 0x3a.toChar: _*) ++
-          Set(0x3c.toChar to 0x5b.toChar: _*) ++
-          Set(0x5d.toChar to 0x7e.toChar: _*)
+        Set(0x20.toChar to 0x7e.toChar: _*) - 0x3b.toChar
```
kailuowang, line 66: Just noticed that HTTP4S is missing a whole range of chars (0x80 - 0xFF) allowed in RFC
```
https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-07#section-4.1.1
```
So we may want take the opportunity to change it to 
```
Set(0x20.toChar to 0xff.toChar: _*) - 0x3b.toChar
```
note that this will include the char `?`, which is not spelled out in the section RFC6265 section 4.4.1 .

danicheg: Good finding 👍🏻

ChristopherDavenport (approved):

PR #6085, opened 2022-03-03
Fix ember client cancellation bug

### LINKED ISSUE #6058: Ember client doesn't shut down quickly when cancelled
To replicate,  set `fork := true`, in your SBT config and compile the following:

```
import cats.data.Kleisli
import cats.effect.kernel.Resource
import cats.effect.{ExitCode, IO, IOApp}
import org.http4s.ember.server.EmberServerBuilder
import org.http4s.ember.client.EmberClientBuilder
import org.http4s.implicits.http4sLiteralsSyntax
import com.comcast.ip4s._

object TestEmberClientHang extends IOApp.Simple {
  def run: IO[Unit] = EmberClientBuilder.default[IO].build.use(_.statusFromUri(uri"http://localhost:8765")).void
}

object RunServer extends IOApp.Simple {
  def run: IO[Unit] =
    EmberServerBuilder.default[IO].withPort(port"8765").withHttpApp(Kleisli(_ => IO.never)).build.useForever
}
```

1. Run `runMain RunServer`
2. In a different SBT session, run `runMain TestEmberClientHang`
3. Hit ctrl+c in the second SBT session (with `TestEmberClientHang`)

Expected: the `TestEmberClientHang` process terminates immediately, like the version using the blaze client does.
Actual: the `TestEmberClientHang` process continues to run for > 30 seconds

Tested on macOS 12.0.1 (21A559)

ChristopherDavenport:
Fixes #6058

makeCase surpresses cancellation. makeCaseFull does not. We could probably specify better what spaces are cancelable and get some better semantics, but this fixes the critical issue of all cancellation being suppressed.

Big Bug need to fix ASAP

### DIFF
--- a/ember-client/shared/src/main/scala/org/http4s/ember/client/EmberClientBuilder.scala
+++ b/ember-client/shared/src/main/scala/org/http4s/ember/client/EmberClientBuilder.scala
@@ -273,7 +273,8 @@ final class EmberClientBuilder[F[_]: Async] private (
               )
             }
           )
-          responseResource <- Resource.makeCase(
+          responseResource <- Resource.makeCaseFull(
+            (poll: Poll[F]) => poll(
             ClientHelpers
               .request[F](
                 request,
@@ -284,6 +285,7 @@ final class EmberClientBuilder[F[_]: Async] private (
                 timeout,
                 userAgent,
               )
+            )
           ) { case ((response, drain), exitCase) =>
             exitCase match {
               case Resource.ExitCase.Succeeded =>
@@ -335,7 +337,7 @@ final class EmberClientBuilder[F[_]: Async] private (
             unixSocketClient(request, address)
           }
       }
-      val stackClient = Retry.create(retryPolicy, logRetries = false)(client)
+      val stackClient = Retry.create(retryPolicy, logRetries = true)(client)
       val iClient = new EmberClient[F](stackClient, pool)
 
       optH2.fold(iClient) { h2 =>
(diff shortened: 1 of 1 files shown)


ember-client/shared/src/main/scala/org/http4s/ember/client/EmberClientBuilder.scala
```diff
@@ -273,17 +273,19 @@ final class EmberClientBuilder[F[_]: Async] private (
...
-                chunkSize,
-                maxResponseHeaderSize,
-                idleConnectionTime,
-                timeout,
-                userAgent,
-              )
+          responseResource <- Resource.makeCaseFull((poll: Poll[F]) =>
+            poll(
+              ClientHelpers
+                .request[F](
```
danicheg, line 279: FWIW: `ClientHelpers.request` should know how to clean up after itself in case it gets canceled. Though, the release step doesn't care about the cancellation at all (`Applicative[F].unit`).

danicheg (approved):
Nice catch!

ember-client/shared/src/main/scala/org/http4s/ember/client/EmberClientBuilder.scala
ChristopherDavenport: Yeah, it only does something when successful. The idea being any errors or cancellation meaning the connection is going to get scrapped.

armanbilge:
Can we add a test for this, or is it tricky?

ChristopherDavenport:
It will only reproduce in an end to end scenario. So possible but not in a clean unit test I don't think. Since it can't run into an error in the middle. Its expected behavior taking a long time and then we cancel that reproduces.

isomarcte (approved):

stephenjudkins:
👍 can confirm this fixes https://github.com/http4s/http4s/issues/6058

Issue #6026, opened 2022-02-08
v0.23 website is out of sync with the repo


bplommer:
https://http4s.org/v0.23/cors/ is not up to date with the new, fixed version of the CORS middleware, although the sources in the repo (https://github.com/http4s/http4s/blob/series%2F0.23/docs/src/main/mdoc/cors.md) are. Both the 0.22 and the 1.0 versions of the page are up to date 🤔

rossabaker:
Because of #5919, I've been reverting the website every time we publish.  It may be related to that, but I can't explain 0.22 vs. 0.23.

armanbilge:
Everything should be up-to-date now.

Issue #5674, opened 2021-12-07
Publish docs site for the 1.0


danicheg:
For now, we have links to page https://http4s.org/v1.0/downloads/ that lead to 404 (because of the recent publishing of the website). We need to publish the docs site for 1.0.

danicheg:
It seems that the landing page of the website is returning 404 now 😮 
I think that because the website was published for the `0.22` and `0.23` versions and might be because of `includeLandingPage = false` setting (for `0.22` and `0.23`). But not 100% sure.

armanbilge:
I think the problem is that series 0.22 and series 0.23 shouldn't be publishing the site at all? As I understand it, we have four laika websites:
1. the shared/unversioned pages. This includes the landing page.
2. /v0.22
3. /v0.23
4. /v1.0

(1) and (4) are published from the main branch. So series/0.22 and series/0.23 should only create their documentation pages and not the rest of the site.

danicheg:
Also, I've noticed that building the website is failing (what strange for the JDK17 only) https://github.com/http4s/http4s/runs/4457849085?check_suite_focus=true#step:3:32. This reproduces locally on the JDK17 only too. I've made a fix for it in https://github.com/http4s/http4s/pull/5689/commits/4b24833446f1d8d103ad27fd6169b872a03587c2.

armanbilge:
Site is back up for now :)

danicheg:
Great! But the publishing of the docs site is still needed.

armanbilge:
Also we still need to fix the problem which caused the site to go down.

rossabaker:
When I fixed the default JDKs, I made the site build on 17 because it's an "app" and because people were grumpy about old JDKs.  It shouldn't really matter 8, 11, or 17 for that job.

armanbilge:
Yeah, I don't think the JDK was specific to the issue. I think it's that each of our branch's builds is configured for its own docs and the shared site, when really only one of builds has that responsibility.

rossabaker:
Yep, that division is how it used to work.

armanbilge:
The 1(.0) site is now updated.

Issue #5488, opened 2021-10-26
"Edit this page" returns 404


i10416:
For example, "Edit this page" in https://http4s.org/v1.0/integrations  refers to https://github.com/http4s/http4s/edit/main/docs/src/main/mdoc/integrations/README.md, but actual file is https://github.com/http4s/http4s/edit/main/docs/src/main/mdoc/integrations.md. 

Note: According to Laika documentation, Laika assumes document is located in ${PAGE_TITLE}/README.md.

armanbilge:
The "Edit this page" functionality has been removed for now. It would be nice to bring it back as a contribution to the sbt-typelevel-site plugin so that all typelevel/http4s projects can have this feature :)

https://github.com/typelevel/sbt-typelevel/

PR #6087, opened 2022-03-05
Simplify website configuration

armanbilge:
Currently the http4s site is plagued by a number of issues. In fact, the 0.22 and 0.23 sites are currently reverted to their pre-Laika versions, and manual intervention is required after every release to keep the landing page from disappearing.

- https://github.com/http4s/http4s/issues/5488
- https://github.com/http4s/http4s/issues/5674
- https://github.com/http4s/http4s/issues/5919
- https://github.com/http4s/http4s/issues/6026

The goal of this PR is to restore the http4s website to a simple but operational configuration.

Towards that I:
1. Merged the two separate site projects into one.
2. Removed all the custom templates and extensions.
3. Reduced the configuration to the minimum necessary.
4. Integrated with the new site plugins in sbt-typelevel and sbt-http4s-org.

This PR is not ready to be merged as-is, but can be previewed locally with `site/tlSitePreview`. The unversioned pages are currently forced to be rendered from this branch, but this is purely temporary for previewing purposes: in the final configuration, they will only ever be published from main.

I recognize that a lot of effort from many contributors has gone into improving the site and I apologize for rolling back so much of that work in this PR. I hope that once the site is actually working again, we can renew those efforts. In fact, some of those changes can probably be upstreamed to sbt-typelevel so that all Typelevel and http4s-org websites may benefit.

The most pressing issue is that this PR will break most/all of our existing URLs since they are no longer "pretty" (i.e. `mypage/` vs `mypage.html`). Potential solutions, with increasing difficulty:
1. Make a very friendly 404 page.
2. Create a redirect for every broken URL with [this technique](https://github.com/http4s/http4s-jdk-http-client/blob/13d0c7ce774b1f25d6853f90b20298aac962b533/docs/index.html).
3. Try to make the URLs pretty again.

I'm aiming for (2) and am strongly against (3), at least in this PR. Laika does not natively support pretty URLs and IMHO the previous workaround made the site configuration overly-complicated, brittle to changes, and behave confusingly different from typical Laika sites.


.github/workflows/ci.yml
```diff
@@ -411,8 +442,13 @@ jobs:
...
-      - name: Scalafix tests
-        if: matrix.scala == '2.13.8'
-        run: |
-          cd scalafix
-          sbt ci
+      - name: Generate site
+        run: sbt '++${{ matrix.scala }}' site/tlSite
+
+      - name: Publish site
+        if: github.event_name != 'pull_request' && github.ref == 'refs/heads/main'
```
armanbilge, line 449: I forgot to configure this, but the plan is that the v0.22 site will publish on every push to `series/0.22`

build.sbt
```diff
@@ -871,24 +871,11 @@ lazy val bench = http4sProject("bench")
...
-  .enablePlugins(
-    GhpagesPlugin,
-    NoPublishPlugin,
-    ScalaUnidocPlugin,
-    MdocPlugin,
-    LaikaPlugin,
-  )
-  .settings(docsProjectSettings)
+lazy val unidocs = http4sProject("unidocs")
+  .enablePlugins(TypelevelUnidocPlugin)
```
armanbilge, line 875: I forgot to mention this: this is a new artifact `http4s-docs` that publishes the aggregated unidocs so they can be viewed at javadoc.io.

project/Http4sSitePlugin.scala
```diff
@@ -0,0 +1,119 @@
...
+import laika.rewrite.{Version, Versions}
+import mdoc.MdocPlugin.autoImport._
+import org.typelevel.sbt.TypelevelSitePlugin.autoImport._
+
+object Http4sSitePlugin extends AutoPlugin {
+
+  override def requires = Http4sPlugin && Http4sOrgSitePlugin
+
+  override def projectSettings = Seq(
+    mdocVariables ++= tlSiteApiUrl.value.map("API_URL" -> _.toString).toMap,
```
armanbilge, line 24: This will be upstreamed to sbt-typelevel.

project/Http4sSitePlugin.scala
```diff
@@ -0,0 +1,119 @@
...
+      latest.map { case ((major, minor), v) =>
+        s"VERSION_${major}_${minor}" -> v.toString
+      }
+    },
+    tlSiteHeliumConfig ~= {
+      _.site.versions(versions.config)
+    },
+    tlSiteHeliumConfig := {
+      val latest = Http4sPlugin.latestPerMinorVersion(baseDirectory.value)
+      if (true || version.value.startsWith("1."))
```
armanbilge, line 40: Temporarily hardcoded `true` to force rendering of landing page.

project/Http4sSitePlugin.scala
```diff
@@ -0,0 +1,119 @@
...
+    )
+
+  }
+
+  object versions {
+    def config: Versions = Versions(
+      currentVersion = current,
+      olderVersions = all.dropWhile(_ != current).drop(1),
+      newerVersions = all.takeWhile(_ != current),
+      renderUnversioned = true || current == v1_0,
```
armanbilge, line 98: Another temporary hardcode.

build.sbt
rossabaker: I have been wondering what will become of the unidocs in the Great Schism.  Since the point is to be on different release cycles, I don't think there is much we can do.  So the value of this will drop, but I think it still provides incremental value.

rossabaker (commented):
I don't understand what gets accomplished merging the sites, but I've lost all understanding of how it works, so trust that it's necessary.

The redirect of old pretty URLs seems like the right strategy to me.  I wish we could do proper HTTP redirects for that, but I have surrendered to the GitHub Pages hegemon.

build.sbt
armanbilge: We could set up an auto-repo that gets steward PRs for latest http4s modules, extracts their API docs, and publishes them to `api.http4s.org` or something. Of course, it wouldn't be able to have any reasonable versioning since it aggregates across independently-versioned modules. But, it could still be useful.

armanbilge:
> I don't understand what gets accomplished merging the sites, but I've lost all understanding of how it works, so trust that it's necessary.

Equally, I don't understand what gets accomplished by keeping them separate :) I know there's the whole concern about docs publishing features before the release but ... when was the last docs PR that would be an example of this?

armanbilge:
Btw, even with the merged sites we can still decide to deploy for 0.22 and 0.23 only on release. Personally I don't think this is necessary, but if it lets us sleep easy at night :)

The only place where it matters is `main`, where we want every push so that the unversioned stuff can be updated in real-time. If the docs on the development branch ocassionally get ahead of the latest milestone, they would technically be documenting the latest snapshot anyway :)

But, I still don't get it. Most of our docs PRs come to versioned, not unversioned material. So at that point we may as well publish the entire site only on releases anyway.

rossabaker:
>  I don't understand what gets accomplished by keeping them separate 

The things previously under `website` don't vary by version.  So for N live branches, we will have N live copies of that content.  Is it assumed that these pages are kept identical across these N branches so it doesn't matter if they're republished?  If not, from where are they published?

armanbilge:
> Is it assumed that these pages are kept identical across these N branches so it doesn't matter if they're republished? If not, from where are they published?

No, unversioned material is only published from `main`. See https://github.com/http4s/http4s/pull/6021#issuecomment-1032105261. This is how the jdk site is working too rn btw.

rossabaker:
Okay, yeah.  Since Laika has a way to discriminate versioned from not, the distinction is much less useful under Laika.  We didn't have that tech when we made the split.

project/Http4sSitePlugin.scala
```diff
@@ -0,0 +1,191 @@
...
+          )
+          .build
+    }
+
+    def html(to: Path) =
+      s"""|<!DOCTYPE html>
+        |<meta charset="utf-8">
+        |<meta http-equiv="refresh" content="0; URL=$to">
+        |<link rel="canonical" href="$to">
+        |""".stripMargin
```
armanbilge, line 144: There. That's not too bad, right?

build.sbt
```diff
@@ -10,6 +10,9 @@ ThisBuild / crossScalaVersions := Seq(scala_3, scala_212, scala_213)
 ThisBuild / tlBaseVersion := "0.22"
 ThisBuild / developers += tlGitHubDev("rossabaker", "Ross A. Baker")
 
+ThisBuild / tlCiReleaseBranches := Seq("series/0.22")
+ThisBuild / tlSitePublishBranch := Some("series/0.22")
```
armanbilge, line 14: Now that the site is fixed, we can enable snapshots :)

build.sbt
armanbilge: Until we publish a new release of v0.22, the API url on the site will be broken because the unidoc `http4s-docs` artifact does not exist yet. We could temporarily hard-code it to something else, but I'd be happy to see a release train start right after we merge this.

armanbilge (commented):
I think this is in a good place. If we're happy with the changes here, then I'll release sbt-http4s-org v0.13.0 and this PR should be ready whenever we are.

I won't claim everything is perfect, but it should be operational so it will be much much easier to push fixes and get them live when issues pop up.

armanbilge:
Actually, hold on this:
- https://github.com/planet42/Laika/issues/265

armanbilge:
Btw, while I'm in the process of redirecting stuff ...

Are we sure we want to use `http4s.org/v1.0/...` ? And not `v1.x` or just `v1`? Is the long term plan to start a new version for every minor? I think it's perfectly reasonable, but possibly overkill. Actually until recently the jdk client had a new site for every patch.

project/Http4sSitePlugin.scala
rossabaker: It's as good as GitHub is going to give us! 😆

build.sbt
rossabaker: I think a new release of 0.22 is coming in the next day or so.  There's the cookie thing and the ember client thing for 0.23.

rossabaker:
`/v1` seems fine to me.

rossabaker:
If we release before the Laika fix, we still have to do the revert dance on gh-pages and miss an opportunity to publish the unidoc?  We're sitting on a couple bug fixes, but I know there's a lot of docs debt that this fixes.

armanbilge:
Yeah, I've been thinking about this. I have a PR in to fix the Laika issue but not sure when Jens will have a chance to look at it.

If you want to start a release train, I can extract a PR that just sets up the unidoc artifact. So at least it will be published and ready for when we merge this one.

But also, I'd really like to land this and fix the site 😆 as much as I hate to say it, maybe I'll hack the darn thing for now. It's one stupid file that's not supposed to be rendered but is. There's a good chance the file is identical across the branches.

When are you thinking about starting that train? I'll try and get this together.

rossabaker:
I might have a window Friday, but I can't guarantee it.  I can do it the old way if this isn't ready whenever I get a break in the clouds.

site/src/main/mdoc/laika/versionInfo.json
```diff
@@ -0,0 +1,32 @@
...
+  "versions": [
+    { "displayValue": "1", "pathSegment": "v1", "fallbackLink": "/docs/quickstart.html", "label": "Dev" },
+    { "displayValue": "0.23", "pathSegment": "v0.23", "fallbackLink": "/docs/quickstart.html", "label": "Stable" },
+    { "displayValue": "0.22", "pathSegment": "v0.22", "fallbackLink": "/docs/quickstart.html", "label": "Stable" },
+    { "displayValue": "0.21", "pathSegment": "v0.21", "fallbackLink": "/index.html", "label": "EOL" },
+    { "displayValue": "Help me choose...", "pathSegment": "", "fallbackLink": "/versions.html" }
+  ],
+  "linkTargets": [
+    { "path": "/docs/auth.html", "versions": ["v0.22", "v0.23", "v1"] },
+    { "path": "/docs/client.html", "versions": ["v0.22", "v0.23", "v1"] },
```
armanbilge, line 11: Until the laika fix comes through, we'll need to keep this file in sync across all three branches. So long as we don't add any new http4s versions or versioned pages I think we should be okay 😆

armanbilge:
Ok, this is ready. Caveats:
1. `versionInfo.json` must be identical across all three branches. Should be fine for now.
2. Redirects on all versions are broken until this merges into main.
3. API docs are broken until the unidoc artifact is published, for each version.

rossabaker (approved):

rossabaker:
So we just need to merge this and then run quick merge trains, and then we're clear for release?  And will want to do so quickly to get the unidocs up?

armanbilge:
Yes. We can even work concurrently: release 0.22 while merging this into 0.23, etc.

rossabaker:
Ah, good.  I was just cursing that merge conflict.

armanbilge:
@rossabaker I'll start the merge into 0.23.

PR #6090, opened 2022-03-07
Collect entity from parser, support Entity.Strict and Empty

wjoel:
Together with the other PRs (#6080, #6091, #6092), this increases performance on "plaintext" from the Techempower benchmarks by around 125% (from 172k reqs/s to 392k reqs/s), and around 80% for JSON.

### DIFF
--- a/blaze-client/src/main/scala/org/http4s/blaze/client/Http1Connection.scala
+++ b/blaze-client/src/main/scala/org/http4s/blaze/client/Http1Connection.scala
@@ -351,14 +351,14 @@ private final class Http1Connection[F[_]](
     val status: Status = parser.getStatus()
     val httpVersion: HttpVersion = parser.getHttpVersion()
 
-    val (attributes, body): (Vault, EntityBody[F]) = if (doesntHaveBody) {
+    val (attributes, entity): (Vault, Entity[F]) = if (doesntHaveBody) {
       // responses to HEAD requests do not have a body
       cleanUpAfterReceivingResponse(closeOnFinish, headers)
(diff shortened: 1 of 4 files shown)


rossabaker (approved):

blaze-client/src/main/scala/org/http4s/blaze/client/Http1Connection.scala
```diff
@@ -405,7 +405,12 @@ private final class Http1Connection[F[_]](
           status = status,
           httpVersion = httpVersion,
           headers = headers,
-          entity = Entity(body.interruptWhen(idleTimeoutS)),
+          entity = entity match {
+            case Entity.Default(body, length) =>
+              Entity[F](body.interruptWhen(idleTimeoutS), length)
+            case _ =>
```
danicheg, line 411: Even it's hard to imagine we will add another subtype of `Entity`, nevertheless, I'm sure that using a wildcard on pattern matching isn't good practice. Verbosity isn't always bad.
```suggestion
            case Entity.Strict(_) | Entity.Empty =>
```

wjoel: Good point, thanks.

wjoel:
Flaky test?

armanbilge:
Yep, we have loads. Feel free to ignore :)

ChristopherDavenport (approved):

danicheg:
I think this is ready for merge. But let's try to achieve the 'green CI' (just for fun, you know).

PR #6092, opened 2022-03-07
Optimize CachingChunkWriter for Chunk.empty case

wjoel:
Together with the other PRs (#6080, #6090, #6091), this increases performance on "plaintext" from the Techempower benchmarks by around 125% (from 172k reqs/s to 392k reqs/s), and around 80% for JSON.

### DIFF
--- a/blaze-core/src/main/scala/org/http4s/blazecore/util/CachingChunkWriter.scala
+++ b/blaze-core/src/main/scala/org/http4s/blazecore/util/CachingChunkWriter.scala
@@ -42,7 +42,7 @@ private[http4s] class CachingChunkWriter[F[_]](
   import ChunkWriter._
 
   private[this] var pendingHeaders: StringWriter = _
-  private[this] var bodyBuffer: Buffer[Chunk[Byte]] = Buffer()
+  private[this] val bodyBuffer: Buffer[Chunk[Byte]] = Buffer()
   private[this] var size: Int = 0
 
   override def writeHeaders(headerWriter: StringWriter): Future[Unit] = {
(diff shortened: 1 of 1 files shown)


blaze-core/src/main/scala/org/http4s/blazecore/util/CachingChunkWriter.scala
```diff
@@ -42,37 +42,44 @@ private[http4s] class CachingChunkWriter[F[_]](
...
+      size += chunk.size
+    }
+
+  private def toChunkAndClear: Chunk[Byte] = {
+    val chunk = if (bodyBuffer.isEmpty) {
+      Chunk.empty
+    } else if (bodyBuffer.size == 1) {
+      bodyBuffer.head
+    } else {
+      Chunk.concat(bodyBuffer)
```
wjoel, line 65: The fold contained in `concat` is quite expensive, so avoid it if possible.

blaze-core/src/main/scala/org/http4s/blazecore/util/CachingChunkWriter.scala
```diff
@@ -42,37 +42,44 @@ private[http4s] class CachingChunkWriter[F[_]](
...
+    if (chunk.nonEmpty) {
+      bodyBuffer += chunk
+      size += chunk.size
+    }
+
+  private def toChunkAndClear: Chunk[Byte] = {
+    val chunk = if (bodyBuffer.isEmpty) {
+      Chunk.empty
+    } else if (bodyBuffer.size == 1) {
+      bodyBuffer.head
```
wjoel, line 63: `.headOption` was noticeably slower, so in this case I think `.head` is the right choice.

rossabaker (commented):
This looks good.  I haven't tried, but I think this one could be backported to 0.23?

wjoel:
@rossabaker thanks! That would be easy, as this file is identical on 0.23. I'm not sure about the procedure. Should I re-target the PR to `series/0.23` and you'll forward merge to `main` from there, or do something else?

armanbilge:
Yes, that's exactly what to do :) thanks!

wjoel:
Done. :)

rossabaker (approved):

blaze-core/src/main/scala/org/http4s/blazecore/util/CachingChunkWriter.scala
```diff
@@ -42,37 +42,44 @@ private[http4s] class CachingChunkWriter[F[_]](
...
-
-  private def clear(): Unit = {
+  private def addChunk(chunk: Chunk[Byte]): Unit =
+    if (chunk.nonEmpty) {
+      bodyBuffer += chunk
+      size += chunk.size
+    }
+
+  private def toChunkAndClear: Chunk[Byte] = {
+    val chunk = if (bodyBuffer.isEmpty) {
```
danicheg, line 60: This micro-opt seems legit - we won't add empty chunks to a buffer:
```suggestion
    val chunk = if (size == 0) {
```

blaze-core/src/main/scala/org/http4s/blazecore/util/CachingChunkWriter.scala
```diff
@@ -42,37 +42,44 @@ private[http4s] class CachingChunkWriter[F[_]](
...
-  }
-
-  private def clear(): Unit = {
+  private def addChunk(chunk: Chunk[Byte]): Unit =
+    if (chunk.nonEmpty) {
+      bodyBuffer += chunk
+      size += chunk.size
+    }
+
+  private def toChunkAndClear: Chunk[Byte] = {
```
danicheg, line 59: Even if `toChunk` and `clear` are coupled in use, it's better to have these atomic ("a function do one thing and is good at it"). But feel free to skip this grouchy nitpick.

danicheg (commented):
Uh-oh, seems I'm too late with my nitpicks 😿 FWIW I submit this anyway.

blaze-core/src/main/scala/org/http4s/blazecore/util/CachingChunkWriter.scala
danicheg: Actually, I think this will save a few nanoseconds, so If anyone else sees benefits from this micro opt - I'll do a PR.

blaze-core/src/main/scala/org/http4s/blazecore/util/CachingChunkWriter.scala
rossabaker: Sorry, I'm not quite following.  Are you proposing splitting this into two?

danicheg: Yep, just as it was previously. But, I'm not too sure about this take. So nevermind.

blaze-core/src/main/scala/org/http4s/blazecore/util/CachingChunkWriter.scala
wjoel: I think it's worth doing, as this is called a lot.

PR #6097, opened 2022-03-09
Docs: ember supports websockets

Daenyth:
I also changed the yes/no characters to the more colorful emoji to make it easier to visually distinguish. It looked really bad on the doc site with the plain squares

### DIFF
--- a/docs/src/main/mdoc/integrations.md
+++ b/docs/src/main/mdoc/integrations.md
@@ -8,12 +8,12 @@ Http4s provides a frontend interface compatible with cats-effect, but it support
 
 | Backend                                                                       | Jdk Support | Http4s Version(s) | 100% Nonblocking               | Http Client                | Http Server           | Websocket Client   | Websocket Server | Proxy support (Client) |
 |-------------------------------------------------------------------------------|-------------|-------------------|--------------------------------|----------------------------|-----------------------|--------------------|------------------|------------------------|
-| [Blaze](https://github.com/http4s/blaze)                                      | 8+          | 0.15+             | ☑︎ Nonblocking                  | `http4s-blaze-client`      | `http4s-blaze-server` | ☒                | ☒              | ☒                    |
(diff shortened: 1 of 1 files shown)


ChristopherDavenport:
That is massively difficult to review in a table in plaintext. 😆

armanbilge:
Yeah, thanks, this page needs some love :)

Annoyingly, I believe this page is currently on the versioned portion of the site. So actually changes on this branch won't be visible to anyone on the 0.23 version of the site for example.

Daenyth:
> That is massively difficult to review in a table in plaintext. 😆

fortunately the rich diff is actually pretty good

> Yeah, thanks, this page needs some love :)
> 
> Annoyingly, I believe this page is currently on the versioned portion of the site. So actually changes on this branch won't be visible to anyone on the 0.23 version of the site for example.

Is that a "manual fix" kind of thing, or can that be fixed by backporting the change here to 0.23's branch?

armanbilge:
Short-term, it will need to be retargeted (ideally to 0.22 actually).

I think in the long term, we should either:
- No longer make this page versioned, since it mentions specific versions on it anyway.
- If we do keep it versioned, clean up that versioning clutter. I have no idea why it's important tbh.

rossabaker:
The versioning info served as a crude measure of maturity, but not a good one.  Some of the newer ones have surpassed some of the older ones.  

If this page becomes versioned, when a backend in another repo sprouts a new feature, how does that get documented?  In the core document for whatever core first supported the feature?

armanbilge:
> If this page becomes versioned, when a backend in another repo sprouts a new feature, how does that get documented? In the core document for whatever core first supported the feature?

This page is currently versioned :)

Yeah, the great schism is a good reason to move this to an unversioned page. Then all the satellite repos and features can be listed in one place.

rossabaker:
Otherwise, this is ready with a cherry-pick back to 0.22?

Daenyth:
> The versioning info served as a crude measure of maturity, but not a good one. Some of the newer ones have surpassed some of the older ones.

For what it's worth, I only added versions to the original page as a "if you are on this http4s version, are you able to use this backend" measure.

I agree it's not particularly valuable now

PR #6098, opened 2022-03-09
Document the use of the execution context passed to the blaze builders.

zainab-ali:
Following a [discussion on discourse](https://discord.com/channels/632277896739946517/632286375311573032/950816734288412672), I'd like to improve the docs on the custom execution context provided to the `BlazeServerBuilder` and `BlazeClientBuilder`.

Let me know what you think.

Thanks!

### DIFF
--- a/blaze-server/src/main/scala/org/http4s/blaze/server/BlazeServerBuilder.scala
+++ b/blaze-server/src/main/scala/org/http4s/blaze/server/BlazeServerBuilder.scala
@@ -209,6 +209,18 @@ class BlazeServerBuilder[F[_]] private (
   override def bindSocketAddress(socketAddress: InetSocketAddress): Self =
     copy(socketAddress = socketAddress)
 
+
+  /** Configures the compute thread pool used to process requests
+    *
+    * This defaults to `cats.effect.Async[F].executionContext`. In
+    * almost all cases, it is desirable to use the default.
+    *
+    * The Blaze server has a single-threaded event loop receiver used
+    * for picking up tcp connections which is completely separate to
+    * this pool. Following picking up a tcp connection, Blaze shifts
+    * to this compute pool to process requests. Request processing
+    * logic is specified by the `HttpApp`.
+    */
   def withExecutionContext(executionContext: ExecutionContext): BlazeServerBuilder[F] =
     copy(executionContextConfig = ExecutionContextConfig.ExplicitContext(executionContext))
 
--- a/blaze-server/src/main/scala/org/http4s/blaze/server/BlazeServerBuilder.scala
+++ b/blaze-server/src/main/scala/org/http4s/blaze/server/BlazeServerBuilder.scala
@@ -209,7 +209,7 @@ class BlazeServerBuilder[F[_]] private (
   override def bindSocketAddress(socketAddress: InetSocketAddress): Self =
     copy(socketAddress = socketAddress)
 
-  /** Configures the compute thread pool used to process requests
+  /** Configures the compute thread pool used to process some async computations.
     *
     * This defaults to `cats.effect.Async[F].executionContext`. In
     * almost all cases, it is desirable to use the default.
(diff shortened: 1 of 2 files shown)


blaze-server/src/main/scala/org/http4s/blaze/server/BlazeServerBuilder.scala
```diff
@@ -209,6 +209,18 @@ class BlazeServerBuilder[F[_]] private (
   override def bindSocketAddress(socketAddress: InetSocketAddress): Self =
     copy(socketAddress = socketAddress)
 
+
+  /** Configures the compute thread pool used to process requests
```
zainab-ali, line 213: I’m not too happy with the wording here — “process requests” is pretty ambiguous. That said, I can’t think of a better term.

blaze-server/src/main/scala/org/http4s/blaze/server/BlazeServerBuilder.scala
```diff
@@ -209,6 +209,18 @@ class BlazeServerBuilder[F[_]] private (
...
+    *
+    * This defaults to `cats.effect.Async[F].executionContext`. In
+    * almost all cases, it is desirable to use the default.
+    *
+    * The Blaze server has a single-threaded event loop receiver used
+    * for picking up tcp connections which is completely separate to
+    * this pool. Following picking up a tcp connection, Blaze shifts
+    * to this compute pool to process requests. Request processing
+    * logic is specified by the `HttpApp`.
```
zainab-ali, line 222: > Request processing logic is specified by the `HttpApp`
On re-reading, I don't think this is correct. The handlers specified by `HttpApp[IO]` would run on the `Async[F].executionContext` pool instead, wouldn't they. It would just be Futures defined internally to blaze that would run on this custom execution context.

rossabaker: I understood this deeply a few effect systems ago, but had to do some experiments now!  Using `withExecutionContext(scala.concurrent.ExecutionContext.global)`, with this route:

```
      case GET -> Root / "thread" =>
        F.delay(Thread.currentThread.getName).flatMap(Ok(_))
```

* The response body is `io-compute-#`.  I'd say this is the "logic specified by the `HttpApp`".
* `reqLoopCallback` gets called on `blaze-selector-#` 
* `runRequest` gets called on `io-compute-#`
* `renderResponse` gets called on `io-compute-#`.
* The effect of writing the body with bodyEncoder gets called on `io-compute-#`.  req loop callback gets called again on the same thread.

So what actually does happen on the ExecutionContext?

* The runnable after we have parsed a request, but we were already on a compute pool.
* Some scheduled tasks, like the idle timeout
* Probably some onComplete handlers in blaze that I can't trace from this repo
* Every `EntityBodyWriter` has one, but I compiled without it!

It's a lot less useful than it used to be!  I think the less-detailed description is better for now, and we might be able to rip out more places it's used with the advances in CE3.

blaze-server/src/main/scala/org/http4s/blaze/server/BlazeServerBuilder.scala
rossabaker: I think it's good.  I came up with "responding to requests", but is that calling the database, or is that just serializing a response body?  My term is worse.

rossabaker:
The deeper I probe into this on Cats-Effect 3, the harder I'm finding it to justify making it distinct from the default.  This might be a candidate for deprecation and removal.

rossabaker (approved):
I think this is an improvement, and it also inspired some cleanups in linked tickets.  Thanks!

I'm still somewhat serious about removing the configuration, but that's probably a discussion for main.

rossabaker:
I'm going to cherry-pick this back to 0.23 before merge.  It's all true there.

PR #6100, opened 2022-03-10
Remove unused ExecutionContexts in blaze-core

rossabaker:
Discovered during #6098.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -690,6 +690,18 @@ lazy val blazeCore = libraryProject("blaze-core")
     libraryDependencies ++= Seq(
       blazeHttp
     ),
+    mimaBinaryIssueFilters := {
+      if (tlIsScala3.value)
+        Seq(
+          ProblemFilters
+            .exclude[DirectMissingMethodProblem]("org.http4s.blazecore.util.BodylessWriter.this"),
+          ProblemFilters
+            .exclude[DirectMissingMethodProblem]("org.http4s.blazecore.util.BodylessWriter.ec"),
+          ProblemFilters
+            .exclude[DirectMissingMethodProblem]("org.http4s.blazecore.util.EntityBodyWriter.ec"),
+        )
+      else Seq.empty
+    },
   )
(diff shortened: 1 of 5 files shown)


rossabaker:
I ran MiMa locally. [shakes fist]

build.sbt
```diff
@@ -726,6 +726,12 @@ lazy val blazeServer = libraryProject("blaze-server")
         .exclude[DirectMissingMethodProblem]("org.http4s.blaze.server.BlazeServerBuilder.this"),
       ProblemFilters
         .exclude[DirectMissingMethodProblem]("org.http4s.blaze.server.WebSocketDecoder.this"),
+      ProblemFilters
+        .exclude[DirectMissingMethodProblem]("org.http4s.blazecore.util.BodylessWriter.this"),
+      ProblemFilters
+        .exclude[DirectMissingMethodProblem]("org.http4s.blazecore.util.BodylessWriter.ec"),
+      ProblemFilters
+        .exclude[DirectMissingMethodProblem]("org.http4s.blazecore.util.EntityBodyWriter.ec"),
```
armanbilge, line 734: I think these are Scala 3 only?

rossabaker: They definitely got changed on every version, but they're private.  Do I have it backward?  Is it Scala 3 that enforces privates and Scala 2 lets them through?  Maybe that's why I didn't see this locally.

rossabaker:
There is more we could do here.  Some of the EC usage left here is a good candidate for `ExecutionContext.parasitic`.  We already have a `trampoline` and a `directec` in blaze that I'm reluctant to touch, but I'm also not sure they reflect modern usage.

build.sbt
armanbilge: Mima doesn't know how to recognize package-privates on Scala 3 yet*, that's why we never see these locally.

* Actually It's fixed, but not released 😕 
- https://github.com/lightbend/mima/issues/682

PR #6109, opened 2022-03-11
Rm deprecated `BlazeClientBuilder`

danicheg:
These deprecated things are an eyesore. I don't feel confident we can remove it for the `0.23` as well, but for the `1.0` definitely do.


rossabaker (approved):
I'm usually very slow to remove deprecations because I've seen a long tail of people upgrading from ancient versions over the years.  But if it's there in 1.0, it's there for a _long_ time.  An apprehensive :+1: from me.

danicheg:
Yeah, I agree with you that removing deprecations should be done carefully. Nevertheless, we are at the point of a new major release, so probably we have a good alibi to make that clean-up.

PR #6111, opened 2022-03-12
Ensure okhttp does not send body for GET and HEAD even if request body is set

### LINKED ISSUE #6108: okhttp-client: calling withEntity on GET request throws IllegalArgumentException
```
Request[IO](Method.GET, uri"https://httpbin.org/get")
        .withEntity("")
```

See https://github.com/rintcius/http4s-clients-poc/blob/02cb0943dba367e2c1644689473a18bfb42315cc/src/test/scala/com/github/rintcius/poc/http4s/client/GetReqsSuite.scala#L58-L80 for repro

Might be defendable but thought to report as an issue because this behaviour is inconsistent with other clients and a switch from any other client to okhttp can lead to breakage.

rintcius:
Resolves #6108

### DIFF
--- a/client/shared/src/test/scala/org/http4s/client/ClientRouteTestBattery.scala
+++ b/client/shared/src/test/scala/org/http4s/client/ClientRouteTestBattery.scala
@@ -129,6 +129,22 @@ abstract class ClientRouteTestBattery(name: String) extends Http4sSuite with Htt
     }
   }
 
+  GetRoutes.getPaths.get(GetRoutes.SimplePath).foreach { expected =>
+    val path = GetRoutes.SimplePath
+
+    test(s"$name Execute GET withEntity $path") {
+      serverClient().flatMap { case (server, client) =>
+        val address = server().addresses.head
+        val req =
+          Request[IO](uri = Uri.fromString(s"http://$address$path").yolo).withEntity("entity")
+        client()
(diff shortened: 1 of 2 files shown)


danicheg (commented):
It'd be great if you could add a test suite to the `ClientRouteTestBattery` to ensure that described behavior is consistent through all available clients. Also, I think that the suggested changes don't violate any specific requirements of `OkHttp`. So I see only benefits in having consistent behavior within clients.

rintcius:
If ppl like, I am also happy to PR the test separately https://github.com/http4s/http4s/compare/main...rintcius:6108-test-client-get-with-entity?expand=1 to show/verify that the test fails on `main`. After CI has run it can then be closed.

rintcius:
Will look into this failure

==> X org.http4s.client.JavaNetClientSpec.JavaNetClient Execute GET withEntity /simple

rossabaker:
The test failure just exposes that it's also broken on the JavaNetClient, right?  That backend is pretty much useless outside a REPL or a script, but it's nice to identify and fix these inconsistencies.

If fixing that is non-trivial, we can mark the test flaky, get this merged, and address the JavaNetClient in a separate PR.

rintcius:
Closing in favor of #6117

PR #6113, opened 2022-03-12
docs: use mdoc instead of laika for variable substitutions in code blocks

jenshalm:
As discussed some time ago, this PR switches from Laika to mdoc for variable substitutions in code blocks.

Using Laika for this purpose had limitations (e.g. they could only be used inside string literals) and won't be supported in a more complete form in the near future. mdoc has built-in support for this functionality without any limitations.

Changes in this PR:
* `build.sbt`: feed the variable map to both, Laika and mdoc, so that Laika can perform substitutions in text markup and mdoc in code blocks.
* `HeliumExtensions`: remove the old workaround for the limited support of this feature.
* Markdown files in the `docs` project: adjust the syntax for the existing variable substitutions to the one required for mdoc.

As probably expected, the syntax for substitutions is different:
* For mdoc the syntax is `@version_circe@`.
* For Laika the syntax is `${version.circe}`.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -906,12 +906,13 @@ lazy val docs = http4sProject("docs")
         scalafixInternalTests,
       ),
     mdocIn := (Compile / sourceDirectory).value / "mdoc",
+    mdocVariables := SiteConfig.mdocVariables.value,
     tlFatalWarningsInCi := false,
     laikaExtensions := SiteConfig.extensions,
     laikaConfig := SiteConfig.config(versioned = true).value,
     laikaTheme := SiteConfig.theme(
       currentVersion = SiteConfig.versions.current,
-      SiteConfig.variables.value,
+      SiteConfig.laikaVariables.value,
       SiteConfig.homeURL.value,
       includeLandingPage = false,
     ),
@@ -959,7 +960,7 @@ lazy val website = http4sProject("website")
     laikaConfig := SiteConfig.config(versioned = false).value,
     laikaTheme := SiteConfig.theme(
       currentVersion = SiteConfig.versions.current,
-      SiteConfig.variables.value,
+      SiteConfig.laikaVariables.value,
       SiteConfig.homeURL.value,
       includeLandingPage = false,
     ),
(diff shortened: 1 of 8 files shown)


armanbilge:
Thanks so much for following up on this 😅 this is a bit awkward, but I've actually fixed this and some other stuff over in https://github.com/http4s/http4s/pull/6087. Actually, that PR would really be helped along by Laika 0.18.2 :)

jenshalm:
Oh, I only pulled latest and did not check open PRs. 🙂 Does it mean the other PR covers everything in here and I can close it? Didn't check, that PR is big...

armanbilge:
> Does it mean the other PR covers everything in here

Yes, I think so, although I took a slightly different approach. In any case, https://github.com/http4s/http4s/pull/6087 will fix our publishing problems so if we need to make any additional changes they should be much easier ...

jenshalm:
Cool, I'll close this then. Where did the `HeliumExtensions` go? Just curious, would like to see how it looks now (assuming you still use that PrettyURL support).

armanbilge:
Hmm, the awkward part now 😅 that PR actually removes all the customizations ... I'm sorry 😕 it was just too complicated for us. You can pull my PR and preview with `site/tlSitePreview` if you wish.

jenshalm:
Ah, but that does mean you are breaking all URLs then, right? I obviously don't mind, but back then I think you preferred to keep all existing links alive.

armanbilge:
> Ah, but that does mean you are breaking all URLs then, right?

Well, yes and no. We are redirecting.

```scala
    def html(to: Path) =
      s"""|<!DOCTYPE html>
          |<meta charset="utf-8">
          |<meta http-equiv="refresh" content="0; URL=$to">
          |<link rel="canonical" href="$to">
          |""".stripMargin
```

jenshalm:
Ok, makes sense. I'll still keep the necessary extension for this on the roadmap for 0.19, as it was really an unnecessary limitation that you needed so much custom code for it. I'm just losing a real-life test case for this, but that's fine. 🙂

Issue #6107, opened 2022-03-11
async-http-client: Calling withHeaders after withEntity removes the just set entity


rintcius:
```
      Request[IO](Method.POST, uri"https://httpbin.org/post")
        .withEntity("some body")
        .withHeaders(
          org.http4s.headers.`Content-Type`(MediaType.text.plain)
        )
```

See https://github.com/rintcius/http4s-clients-poc/blob/master/src/test/scala/com/github/rintcius/poc/http4s/client/PostReqsSuite.scala#L68-L80 for repro

rintcius:
Related to #6106 - okhttp has same issue, but blaze, ember and jetty don't lose the entity

rossabaker:
There's a broader discussion to be had about this backend: upstream has been [unmaintained since June](https://github.com/AsyncHttpClient/async-http-client/commit/c959fa0483adb4a71fbccc5be94d8b6faa4f74be).

rossabaker:
`withHeaders` replaces all headers.  `putHeaders` replaces headers of the same name.  (This is a good use of "put" if we think of it like a Map, and a bad use of "put" if we think of it like HTTP.)

rintcius:
I guess the broader question here is: is the goal of http4s to make switching from one client to another as frictionless as possible? 

I.e. currently, if someone constructed a request like `req.withEntity(..).withHeaders(..)` using blaze/ember/jetty then things work as expected. However if they decide to switch to async-http-client or okhttp then things will break.

rossabaker:
> is the goal of http4s to make switching from one client to another as frictionless as possible

Yes, but "frictionless" is open to some interpretation as we get into undefined behavior.

When we call `withHeaders` after a `withEntity` of known size, we obliterate the `Content-Length` header.  `putHeaders` preserves it.  (The same happens for streaming bodies, with `Transfer-Encoding` in place of the `Content-Length` header).

```scala
scala> Request[IO]().withEntity("test").headers.foreach(println)
Content-Length: 4
Content-Type: text/plain; charset=UTF-8

scala> Request[IO]().withEntity("test").withHeaders("X-Foo" -> "bar").headers.foreach(println)
X-Foo: bar

scala> Request[IO]().withEntity("test").putHeaders("X-Foo" -> "bar").headers.foreach(println)
Content-Length: 4
Content-Type: text/plain; charset=UTF-8
X-Foo: bar
```

[...]

PR #6114, opened 2022-03-13
Deprecate async-http-client

rossabaker:
As noted on #6107, upstream has been [unmaintained for months](https://github.com/AsyncHttpClient/async-http-client/commit/c959fa0483adb4a71fbccc5be94d8b6faa4f74be).  We should start chasing people off it.

### DIFF
--- a/async-http-client/src/main/scala/org/http4s/asynchttpclient/AsyncHttpClient.scala
+++ b/async-http-client/src/main/scala/org/http4s/asynchttpclient/AsyncHttpClient.scala
@@ -41,6 +41,7 @@ import org.http4s.internal.bug
 import org.http4s.internal.threads._
 import org.reactivestreams.Publisher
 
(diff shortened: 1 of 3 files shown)


rossabaker:
I think @isomarcte is a user.  Summoning him to see if this is a hardship.

danicheg (approved):
I'm not an actual `AsycHttpClient` user, so don't rely on my :+1: much. 
Albeit, if upstream is EOL, then I opine that we should deprecate the client within `http4s`.

isomarcte:
@rossabaker I'm not a user. I tend to use Ember or the JDK client.

isomarcte (approved):

rossabaker:
Oh, I thought you had some old stuff on it.  In that case, away with it.  It's one `@nowarn` for people who can't migrate.

PR #6115, opened 2022-03-13
Deprecate internal Trampoline

rossabaker:
Blaze has its own.  Scala 2.13 has parasitic.

Only keeping this in case a satellite repo was using it.


armanbilge (approved):

rossabaker:
Well, I guess it is used.  But we can replace it in 1.0.

PR #6116, opened 2022-03-13
Remove redundant draining of response body in `Http4sServlet` 

danicheg:
Yet another micro-optimizations.

### DIFF
--- a/servlet/src/main/scala/org/http4s/servlet/Http4sServlet.scala
+++ b/servlet/src/main/scala/org/http4s/servlet/Http4sServlet.scala
@@ -91,7 +91,7 @@ abstract class Http4sServlet[F[_]](
         case Left(t) =>
           response.entity match {
             case Entity.Default(body, _) =>
-              body.drain.compile.drain.handleError { t2 =>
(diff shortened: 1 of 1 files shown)


servlet/src/main/scala/org/http4s/servlet/Http4sServlet.scala
```diff
@@ -83,15 +83,21 @@ abstract class Http4sServlet[F[_]](
...
     }.attempt
       .flatMap {
         case Right(()) => bodyWriter(response)
         case Left(t) =>
-          response.body.drain.compile.drain.handleError { t2 =>
-            logger.error(t2)("Error draining body")
-          } *> F.raiseError(t)
+          response.entity match {
+            case Entity.Default(body, _) =>
+              body.drain.compile.drain.handleError { t2 =>
```
diesalbla, line 94: By the way, we can remove the first `drain`, since it has the same meaning (discarding outputs) as the first one.

rossabaker (approved):
:+1:, and to @diesalbla's suggestion

Issue #6106, opened 2022-03-11
okhttp-client: Calling withHeaders after withEntity removes the just set entity


rintcius:
```
      Request[IO](Method.POST, uri"https://httpbin.org/post")
        .withEntity("some body")
        .withHeaders(
          org.http4s.headers.`Content-Type`(MediaType.text.plain)
        )
```

See https://github.com/rintcius/http4s-clients-poc/blob/master/src/test/scala/com/github/rintcius/poc/http4s/client/PostReqsSuite.scala#L68-L80 for repro

rossabaker:
Same comment as #6107 

> `withHeaders` replaces all headers.  `putHeaders` replaces headers of the same name.  (This is a good use of "put" if we think of it like a Map, and a bad use of "put" if we think of it like HTTP.)

PR #6117, opened 2022-03-13
Add Message.normalize and use in clients

### LINKED ISSUE #6108: okhttp-client: calling withEntity on GET request throws IllegalArgumentException

rintcius:
Resolves #6108 

I was just thinking about an alternative way to solve #6108. The advantage of this approach is that there's a central place to clean up requests, so no need to deal with internally inconsistent requests in every client.

E.g. as opposed to #6111 this approach fixes JavaNetClient automatically too. I also imagine this approach could be used to solve #6106 and #6107

This is still WIP by the way. If it's useful to have this for `Response` I can call `normalize` for responses too. 
If not, I can move `normalize` from `Message` to `Request` or even somewhere less public so that is not exposed to users of http4s.

### DIFF
--- a/client/jvm/src/main/scala/org/http4s/client/JavaNetClientBuilder.scala
+++ b/client/jvm/src/main/scala/org/http4s/client/JavaNetClientBuilder.scala
@@ -109,7 +109,8 @@ sealed abstract class JavaNetClientBuilder[F[_]] private (
     * The shutdown of this client is a no-op. $WHYNOSHUTDOWN
     */
   def create: Client[F] =
-    Client { (req: Request[F]) =>
+    Client { (request: Request[F]) =>
+      val req = request.normalize
       def respond(conn: HttpURLConnection): F[Response[F]] =
         for {
           _ <- configureSsl(conn)
--- a/client/shared/src/test/scala/org/http4s/client/ClientRouteTestBattery.scala
+++ b/client/shared/src/test/scala/org/http4s/client/ClientRouteTestBattery.scala
@@ -129,6 +129,22 @@ abstract class ClientRouteTestBattery(name: String) extends Http4sSuite with Htt
     }
   }
 
+  GetRoutes.getPaths.get(GetRoutes.SimplePath).foreach { expected =>
+    val path = GetRoutes.SimplePath
+
+    test(s"$name Execute GET withEntity $path") {
+      serverClient().flatMap { case (server, client) =>
+        val address = server().addresses.head
+        val req =
+          Request[IO](uri = Uri.fromString(s"http://$address$path").yolo).withEntity("entity")
+        client()
+          .run(req)
+          .use(resp => expected.flatMap(checkResponse(resp, _)))
+          .assert
+      }
+    }
+  }
+
   test("Mitigates request splitting attack in URI path") {
     serverClient().flatMap { case (server, client) =>
       val address = server().addresses.head
--- a/core/shared/src/main/scala/org/http4s/Message.scala
+++ b/core/shared/src/main/scala/org/http4s/Message.scala
@@ -53,6 +53,8 @@ sealed trait Message[F[_]] extends Media[F] { self =>
 
   def attributes: Vault
 
+  def normalize: Self
+
   protected def change(
       httpVersion: HttpVersion = httpVersion,
       entity: Entity[F] = entity,
@@ -297,6 +299,12 @@ final class Request[F[_]] private (
       entity = entity.translate(f),
       attributes = attributes,
     )
+
+  def normalize: Request[F] =
+    if (method === Method.GET || method === Method.HEAD) {
+      change(entity = Entity.empty).removeHeader[`Content-Length`]
+    } else
+      this
 
   def withMethod(method: Method): Request[F] =
     copy(method = method)
@@ -581,6 +589,8 @@ final class Response[F[_]] private (
     with Product
     with Serializable {
   type SelfF[F0[_]] = Response[F0]
+
+  def normalize: Response[F] = this
 
   def mapK[G[_]](f: F ~> G): Response[G] =
     Response[G](
--- a/async-http-client/src/main/scala/org/http4s/asynchttpclient/AsyncHttpClient.scala
+++ b/async-http-client/src/main/scala/org/http4s/asynchttpclient/AsyncHttpClient.scala
@@ -63,7 +63,10 @@ object AsyncHttpClient {
         Resource(F.async[(Response[F], F[Unit])] { cb =>
           F.delay(
             httpClient
-              .executeRequest(toAsyncRequest(req, dispatcher), asyncHandler(cb, dispatcher))
+              .executeRequest(
+                toAsyncRequest(req.normalize, dispatcher),
+                asyncHandler(cb, dispatcher),
+              )
           ).as(None)
         })
       }
--- a/blaze-client/src/main/scala/org/http4s/blaze/client/BlazeClient.scala
+++ b/blaze-client/src/main/scala/org/http4s/blaze/client/BlazeClient.scala
@@ -83,7 +83,8 @@ private class BlazeClient[F[_], A <: BlazeConnection[F]](
 )(implicit F: Async[F])
     extends DefaultClient[F] {
 
-  override def run(req: Request[F]): Resource[F, Response[F]] = {
+  override def run(request: Request[F]): Resource[F, Response[F]] = {
+    val req = request.normalize
     val key = RequestKey.fromRequest(req)
     for {
       requestTimeoutF <- scheduleRequestTimeout(key)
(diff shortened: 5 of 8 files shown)


client/shared/src/test/scala/org/http4s/client/ClientRouteTestBattery.scala
```diff
@@ -129,6 +129,51 @@ abstract class ClientRouteTestBattery(name: String) extends Http4sSuite with Htt
...
+        val req =
+          Request[IO](uri = Uri.fromString(s"http://$address$path").yolo).withEntity("entity")
+        client()
+          .run(req)
+          .use(resp => expected.flatMap(checkResponse(resp, _)))
+          .assert
+      }
+    }
+
+    test(s"$name Execute GET withEntity then withHeaders $path") {
```
rintcius, line 147: Note: this test confuses ember if run without `normalize` (say if it were added to #6111 ). 
The test itself succeeds but another (the next?) test in the suite starts failing.

Just thought to mention it because it may indicate an issue in ember.

armanbilge: Yes, thanks for mentioning this. It reminds me of https://github.com/http4s/http4s/issues/5256, which also had the symptom of the _next_ test failing.

rossabaker:
Seems like this could be applicable in 0.23?  This isn't just a problem in the new entity model, is it?

core/shared/src/main/scala/org/http4s/Message.scala
```diff
@@ -53,6 +53,8 @@ sealed trait Message[F[_]] extends Media[F] { self =>
 
   def attributes: Vault
 
+  def normalize: Self
```
isomarcte, line 56: Since this would be a no-op on Response, should we only define this on Request? What is the advantage of having it on the super type?

core/shared/src/main/scala/org/http4s/Message.scala
```diff
@@ -298,6 +300,12 @@ final class Request[F[_]] private (
       attributes = attributes,
     )
 
+  def normalize: Request[F] =
```
isomarcte, line 303: :thinking: A couple thoughts here.

Bodies are not forbidden with GET requests, but have undefined semantics and servers _must_ ignore them. That said, I wonder if the spec permits an encoding we should also permit it as well, but just heavily discourage the use. 

In any case, should we yield an error if the components of the message are invalid, rather than silently clean them up?

rossabaker: The original bug (#6108) says that while HTTP allows this, some backends (e.g., okhttp-client) don't.  The test guarantees that the same request is processed the same way by a server, regardless of how the client chooses to encode it.

Since a complaint HTTP server MUST ignore the body, a test against a compliant HTTP server can't further specify that a body is or is not sent.  We could check the entity with a raw TCP server, but since the distinction isn't observable through HTTP, should we?  Normalizing everything makes things more consistent, but I think maybe all we need to do is get all the backends passing this new test.  Some backends will have to normalize, and others, :shrug:.

rintcius: Yeah, the client backends seem to be too different for a single central normalization to work nicely. E.g. while reading a bit more, I came across this example https://www.elastic.co/guide/en/elasticsearch/reference/8.1/search-search.html#search-search-api-example - so maybe there are http4s users out there who have selected their client to allow this kind of request and those clients shouldn't break.

core/shared/src/main/scala/org/http4s/Message.scala
rintcius: Right, that was a fairly random decision - was mostly looking for some feedback on the approach - see last paragraph of the PR description.

client/shared/src/test/scala/org/http4s/client/ClientRouteTestBattery.scala
```diff
@@ -129,6 +129,51 @@ abstract class ClientRouteTestBattery(name: String) extends Http4sSuite with Htt
...
+        val req =
+          Request[IO](uri = Uri.fromString(s"http://$address$path").yolo).withEntity("entity")
+        client()
+          .run(req)
+          .use(resp => expected.flatMap(checkResponse(resp, _)))
+          .assert
+      }
+    }
+
+    // test(s"$name Execute GET withEntity then withHeaders $path") {
```
rintcius, line 147: This currently fails for ember.

I think there are 2 issues here:
1. `withHeaders` currently removes internally added headers (like in this test content length that's added as part of `.withEntity("entity")` ). I _think_ these internally added headers should be left alone unless they are part of `withHeaders`.
2. Requests without content length confuse ember. This looks like the request counterpart of https://github.com/http4s/http4s/pull/5881

rintcius (commented):
Now decentralised request normalization

client/jvm/src/main/scala/org/http4s/client/JavaNetClientBuilder.scala
```diff
@@ -139,6 +140,14 @@ sealed abstract class JavaNetClientBuilder[F[_]] private (
   def resource: Resource[F, Client[F]] =
     Resource.eval(F.delay(create))
 
+  private def normalizeRequest(req: Request[F]): Request[F] =
+    if (req.method === Method.GET) {
```
rintcius, line 144: I left HEAD out of scope for this PR - happy to look into this afterwards though.

client/shared/src/test/scala/org/http4s/client/ClientRouteTestBattery.scala
rintcius: Guessing that https://github.com/http4s/http4s/issues/6106 and https://github.com/http4s/http4s/issues/6107 will also be resolved if 1. was resolved.

core/shared/src/main/scala/org/http4s/Message.scala
rintcius: @isomarcte Ah yes, I removed most of the description because it was outdated - visible under "edited" still if you are interested :)

client/shared/src/test/scala/org/http4s/client/ClientRouteTestBattery.scala
```diff
@@ -129,6 +129,51 @@ abstract class ClientRouteTestBattery(name: String) extends Http4sSuite with Htt
...
+    //     val req =
+    //       Request[IO](uri = Uri.fromString(s"http://$address$path").yolo)
+    //         .withEntity("entity")
+    //         .withHeaders(headers.`Content-Type`(MediaType.text.plain))
+    //     client()
+    //       .run(req)
+    //       .use(resp => expected.flatMap(checkResponse(resp, _)))
+    //       .assert
+    //   }
+    // }
```
rossabaker, line 159: This addresses 1 [above](https://github.com/http4s/http4s/pull/6117/files/0a043b23bd9691c3e0a9ab26baba3c03236c1a93#r826395887), and now the ember-client tests pass for me:

```suggestion
    test(s"$name Execute GET withEntity then putHeaders $path") {
      serverClient().flatMap { case (server, client) =>
        val address = server().addresses.head
        val req =
          Request[IO](uri = Uri.fromString(s"http://$address$path").yolo)
            .withEntity("entity")
            .putHeaders(headers.`Content-Type`(MediaType.text.plain))
        client()
          .run(req)
          .use(resp => expected.flatMap(checkResponse(resp, _)))
          .assert
      }
    }
```

I suspect you're right on point 2, which is worth a separate issue.

rintcius: Right, the point I tried to make with these tests is that there are surprises in the use of `withHeaders`, i.e.
the request that's created via `req.withEntity(..).withHeaders(..)` is different than the request created via `req.withHeaders(...).withEntity(...)`. I think there would be less surprises if the semantics would be like I described in point 1 [here](https://github.com/http4s/http4s/pull/6117#discussion_r826395887) (but since it changes current behaviour and I'm not sure whether users rely on this behaviour somehow, I don't know if it should be changed at this stage).

Note that it's not just `withEntity(..).withHeaders(..)` that's doesn't create content length header, this test also fails for Ember also because the constructed request doesn't have content length:

```
    test(s"$name Execute GET with Request(uri, headers, entity) $path".only) {
      serverClient().flatMap { case (server, client) =>
        val address = server().addresses.head
        val req =
          Request[IO](
            uri = Uri.fromString(s"http://$address$path").yolo, 
            headers = Headers(headers.`Content-Type`(MediaType.text.plain)),
            entity = EntityEncoder.stringEncoder.toEntity("entity"))
        client()
          .run(req)
          .use(resp => expected.flatMap(checkResponse(resp, _)))
          .assert
      }
    }
  }
```

isomarcte (approved):

PR #6118, opened 2022-03-13
Further reduce ExecutionContexts in blaze

rossabaker:
Trivial mappings and flatMappings are moved to a parasitic `ExecutionContext` to avoid a thread shift.  All of these could probably use the `directec`, which is infintesimally faster, but not stack safe.  We don't have to think as hard with this one.

Because parasitic doesn't exist in 2.13, we introduce a compat trait to use the old blaze trampoline on Scala 2.12.


danicheg (approved):
A bit funny, but today I also was thinking about execution context when dug into `Http2Writer` internals (and thought about using parasitic EC). Matrix glitch, or something like that, 100% 😄

rossabaker:
I'm currently looking askance at the call to ExecutionContext in Http1ServerStage.runRequest.

rossabaker:
Also discovered as part of #6098.

PR #6126, opened 2022-03-14
Backport mergify config to 0.23

armanbilge:
Our mergify stuff stopped working, because mergify reads its config from the default branch which changed.


armanbilge:
Also: https://github.com/scala-steward-org/repos/pull/966

I guess changing the default branch also requires thoughtfulness 😅

danicheg (approved):

PR #6128, opened 2022-03-14
Remove redundant draining of request/response body

danicheg:
A tiny refactoring as a follow-up from https://github.com/http4s/http4s/pull/6116#discussion_r825437891


rossabaker (approved):
:+1:, but this is going to cause more merge conflicts on the 0.23 we're trying to get out the door.  We might want to sit on it until #6125 is done.

danicheg:
Sure, this will wait for #6125. Furthermore, I could give it a try to fix hereafter conflicts of merging.

rossabaker:
I'm merging to get off the board, but not rolling forward from 0.22 until we get what's in flight for 0.23 released.

PR #6130, opened 2022-03-14
Refactor the `EntityBodyWriter.writeEntityBody`

danicheg:
This brings changes in the `EntityBodyWriter.writeEntityBody` contract. Nevertheless, that allows adding optimizations to handle specific `Entity` subtypes.


blaze-core/src/test/scala/org/http4s/blazecore/util/DumpingWriter.scala
```diff
@@ -30,7 +30,7 @@ import scala.concurrent.Future
 object DumpingWriter {
   def dump(p: EntityBody[IO]): IO[Array[Byte]] = {
     val w = new DumpingWriter()
-    for (_ <- w.writeEntityBody(p)) yield (w.toArray)
+    for (_ <- w.writeEntityBody(Entity(p))) yield (w.toArray)
```
rossabaker, line 33: Is there any reason that `dump` wouldn't just take an `Enttiy`?

blaze-core/src/main/scala/org/http4s/blazecore/util/EntityBodyWriter.scala
```diff
@@ -51,18 +51,25 @@ private[http4s] trait EntityBodyWriter[F[_]] {
...
+  def writeEntityBody(entity: Entity[F]): F[Boolean] =
+    entity match {
+      case Entity.Default(body, _) =>
+        val writeBody: F[Unit] = writePipe(body).compile.drain
+        val writeBodyEnd: F[Boolean] = fromFutureNoShift(F.delay(writeEnd(Chunk.empty)))
+        writeBody *> writeBodyEnd
+
+      case Entity.Strict(chunk) =>
+        fromFutureNoShift(F.delay(writeBodyChunk(chunk, flush = false))) *>
+          fromFutureNoShift(F.delay(writeEnd(Chunk.empty)))
```
rossabaker, line 68: Can we just `writeEnd(chunk)` instead of doing two writes?

blaze-core/src/test/scala/org/http4s/blazecore/util/DumpingWriter.scala
danicheg: Good idea. I wasn't thinking about refactoring much because this is a test thing, but let's do it.

blaze-core/src/main/scala/org/http4s/blazecore/util/EntityBodyWriter.scala
danicheg: I feel that would mean behavior changes then. We can safely do so for the `BodylessWriter`. But I'm not sure about the cases of other writers, say `CachingChunkWriter`. Nevertheless, I could misapprehend things and would be grateful for clarification.

rossabaker: I looked, and it's not going to be a big difference on `CachingChunkWriter`.  They're both ultimately going to go through `writeBodyChunk`, and then a flatMap, and then write the trailer.  Similarly with `FushingChunkWriter`.

I would say it's better to call `writeEnd` when you know it's the last chunk.  That's why it takes a chunk.  And a writer implementation may or may not be optimized for that knowledge.  We just haven't ever known it was the last chunk since the days of scalaz-stream.

danicheg: I've tried to change `writeEntityBody` on case `Entity.Strict` and got a bunch of failed tests within `blaze-server`:
```
org.http4s.blaze.server.BlazeServerSuite
org.http4s.blaze.server.BlazeServerMtlsSpec
org.http4s.blaze.server.Http1ServerStageSpec
```
Roughly 1.5 dozen of tests. 
```scala
def writeEntityBody(entity: Entity[F]): F[Boolean] =
  entity match {
    case Entity.Default(body, _) =>
      val writeBody: F[Unit] = writePipe(body).compile.drain
      val writeBodyEnd: F[Boolean] = fromFutureNoShift(F.delay(writeEnd(Chunk.empty)))
      writeBody *> writeBodyEnd

    case Entity.Strict(_) | Entity.Empty =>
      fromFutureNoShift(F.delay(writeEnd(Chunk.empty)))
  }
```

rossabaker: I think you would need:

```scala
case Entity.Strict(chunk) =>
  fromFutureNoShift(F.delay(writeEnd(chunk)))
```

It can't share with `Entity.Empty`, because that last chunk needs to be passed to `writeEnd`.

danicheg: Darn. Now I see what did you mean. Locally works fine, all tests passed. Thanks!

rossabaker (approved):

PR #6131, opened 2022-03-14
Ember Core: avoid string allocation in method parse

diesalbla:
Current code for parsing method always creates a `String` to cover the region of the array. This string is an additional object which has to copy the positions of the byte array, and then is discarded.

Instead, we can just use existing methods, and compare their names to the bytes. This avoids the extra allocation. On the other hand, unlike the map which could compare with few, or may
be using a Hash, here we compare with all methods listed.

### DIFF
--- a/ember-core/shared/src/main/scala/org/http4s/ember/core/Parser.scala
+++ b/ember-core/shared/src/main/scala/org/http4s/ember/core/Parser.scala
@@ -160,6 +160,11 @@ private[ember] object Parser {
 
     object ReqPrelude {
 
+      // a specialised list of methods, for searching. Puts CRUD ones go on top because we
+      // assume those are the most commonly used ones.
+      private[this] val frequentMethods: List[Method] =
+        Method.GET :: Method.POST :: Method.PUT :: Method.DELETE :: Method.all
+
       // Method SP URI SP HttpVersion CRLF - REST
       def parsePrelude[F[_]](message: Array[Byte], maxHeaderSize: Int)(implicit
           F: MonadThrow[F]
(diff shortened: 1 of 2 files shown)


ember-core/shared/src/main/scala/org/http4s/ember/core/Parser.scala
```diff
@@ -174,17 +179,29 @@ private[ember] object Parser {
...
         var start = 0
         while (!complete && idx <= upperBound) {
           val value = message(idx)
           (state: @switch) match {
             case 0 =>
               if (value == space) {
-                Method.fromString(new String(message, start, idx - start)) match {
-                  case Left(e) =>
-                    throwable = e
+                frequentMethods.find(m => matchesString(start, idx, m.name)) match {
```
danicheg, line 199: I feel like linear search would be slower than getting value by key in a hash map. Wouldn't that brings some regression of speed in the parsing?

diesalbla: That is what I started to wonder. That is why I added that `frequentMethods` list, altered in favour of the frequent four methods, GET, POST, PUT, and DELETE. Not sure if allocation is more expensive than building the string and computing its hash.

ChristopherDavenport:
Have you tried benching this? I see this as additional complication that may not add any performance. But happy to consider it if this shows improved performance on a bench.

rossabaker:
I can't remember which project, but I've seen switch tables built for this, where if the first character is `P`, it looks for an `O` or a `U` next, and so on.  I wouldn't do it without a benchmark, but there's prior art.

PR #6133, opened 2022-03-15
Use blocking EC in the `BlockingServletIo`

danicheg:
Previously `BlockingServletIo` used a blocker (https://github.com/http4s/http4s/pull/4175/files#diff-476d2c035d983230b75e27a4ec07fbc8a40cb2ed1ef26c75c4dbef39a4bf1c1eL58), so I opine it should use a blocking EC accordingly.


rossabaker (approved):
Ouch.  Yes.

PR #6134, opened 2022-03-15
Add Scalafix explicit result type rule

danicheg:
This is a follow-up from https://github.com/http4s/http4s/pull/5822#issuecomment-1061659972. 
I didn't figure out ways to filter out tests/some projects for this specific rule. Probably that's not supported by Scalafix at all. Though, I'm not ten to one sure about that. Would be glad if more experienced Scalafix-folks correct me on this.
Also, one should keep in mind that there are open issues about that rule (e. g. https://github.com/scalacenter/scalafix/issues/1219).
For all that, I find that rule extra helpful. Self-described code is lovely.

### DIFF
--- a/server/src/test/scala/org/http4s/server/staticcontent/FileServiceSuite.scala
+++ b/server/src/test/scala/org/http4s/server/staticcontent/FileServiceSuite.scala
@@ -30,9 +30,9 @@ import java.io.File
 import java.nio.file._
 
 class FileServiceSuite extends Http4sSuite with StaticContentShared {
-  private val defaultSystemPath =
+  val defaultSystemPath =
     org.http4s.server.test.BuildInfo.test_resourceDirectory.getAbsolutePath
-  val routes: HttpRoutes[IO] = fileService(
+  val routes = fileService(
     FileService.Config[IO](new File(getClass.getResource("/").toURI).getPath, testBlocker)
   )
 
@@ -161,7 +161,7 @@ class FileServiceSuite extends Http4sSuite with StaticContentShared {
   }
 
   test("return files included via symlink") {
-    val relativePath = "symlink/org/http4s/server/staticcontent/StaticContentShared.scala"
+    val relativePath = "symlink/org/http4s/server/staticcontent/FileServiceSuite.scala"
     val path = Paths.get(defaultSystemPath).resolve(relativePath)
     val file = path.toFile
     val bytes = Chunk.bytes(Files.readAllBytes(path))
--- a/server/src/test/scala/org/http4s/server/staticcontent/FileServiceSuite.scala
+++ b/server/src/test/scala/org/http4s/server/staticcontent/FileServiceSuite.scala
@@ -30,7 +30,7 @@ import java.io.File
 import java.nio.file._
 
 class FileServiceSuite extends Http4sSuite with StaticContentShared {
-  val defaultSystemPath =
+  private val defaultSystemPath =
     org.http4s.server.test.BuildInfo.test_resourceDirectory.getAbsolutePath
   val routes: HttpRoutes[IO] = fileService(
     FileService.Config[IO](new File(getClass.getResource("/").toURI).getPath, testBlocker)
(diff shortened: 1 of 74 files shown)


server/src/test/scala/org/http4s/server/staticcontent/FileServiceSuite.scala
```diff
@@ -161,7 +161,7 @@ class FileServiceSuite extends Http4sSuite with StaticContentShared {
   }
 
   test("return files included via symlink") {
-    val relativePath = "symlink/org/http4s/server/staticcontent/FileServiceSuite.scala"
+    val relativePath = "symlink/org/http4s/server/staticcontent/StaticContentShared.scala"
```
danicheg, line 164: I quite misapprehend why, but that test ("return files included via symlink") is failing with added type annotations. This particular change actually fixes that test.

armanbilge: Yeah I was wondering about this too. It would be good to understand why...

danicheg: Unfortunately, I even don't have any ideas about it. I was thinking about a kinda caching problem. But the test file is consistent when SBT is starting and when the test class is running. So if someone has any ideas, it'd be nice to figure out this a bit odd behavior.

armanbilge: I'll try and take a look later, but no need to block this PR. 👍 from me, would be great to get @bplommer's eyes on it :)

bjaglin:
> I didn't figure out ways to filter out tests/some projects for this specific rule.

The `scalafixConfig` setting key from sbt-scalafix is looked up at the `project / configuration` scope, so you can use sbt scope delegation rules to target a no-op config file for a given project and/or configuration. It's not pretty but it should work.

danicheg:
> The `scalafixConfig` setting key from sbt-scalafix is looked up at the `project / configuration` scope, so you can use sbt scope delegation rules to target a no-op config file for a given project and/or configuration. It's not pretty but it should work.

Thanks! That's quite interesting. Could you please refer to some working examples of that kind of configuration?

armanbilge:
I think we need to add another file like this one:
https://github.com/http4s/http4s/blob/a0c31a6cbc1a69dcc43be6c4ad2fa126d30deecd/.scalafix.conf

Maybe call it `.scalafix.test.conf`.

Then we can probably do `scalafixConfig := file(".scalafix.test.conf")` or `Test / scalafixConfig := file(".scalafix.test.conf")`

danicheg:
@armanbilge yeah, I've asked for a demo just because I tried to do similar to your council and didn't succeed.

bjaglin:
@danicheg I could only find a test [overriding the default config for a given project](https://github.com/scalacenter/sbt-scalafix/blob/f6fffde77e5ba520db81f75984a5510444391931/src/sbt-test/sbt-scalafix/cross-build/build.sbt#L50), not for a given configuration. There might be some sbt scope delegation trick I missed... Do you have a branch with your test?

armanbilge:
I'm going to take a look a the test configuration issue and the `FileServiceSuite` weird thing. Afterwards lets merge 👍

danicheg:
I think having `ExplicitResultTypes` enabled for tests is fine. Tests are also code. But configuring Scalafix working for main/tests with different rules is high-potential though. It'd be cool if we do it separately.

armanbilge:
I think so too, which is why we shouldn't block. But let me look for 20 minutes :)

rossabaker:
I think the whole value of the rule is to avoid surprising inferred types in public API.  But it looks like it inserts them for us, so I'm not as worried as I originally was about it applying to tests.

rossabaker:
And it looks like maybe it was figured out anyway while I was feeding the dogs. 😆

armanbilge:
@mergifyio update

armanbilge:
Oh right, mergify can't actually fix it 🤦

armanbilge (approved):

danicheg:
@armanbilge thanks for putting the finishing touches on this. Now I think we have done exactly what one originally intended.

PR #6136, opened 2022-03-17
Support parsing response bodies without Content-Length in ember (0.21)

wemrysi:
Backport of #5881 to series/0.21

### DIFF
--- a/ember-core/src/main/scala/org/http4s/ember/core/Parser.scala
+++ b/ember-core/src/main/scala/org/http4s/ember/core/Parser.scala
@@ -456,12 +456,13 @@ private[ember] object Parser {
           }
 
         val body =
-          Stream
+          Pull
             .eval(consumed.get)
-            .ifM(
-              ifTrue = Stream.raiseError(BodyAlreadyConsumedError()),
-              ifFalse = Stream.chunk(Chunk.array(buffer)) ++ readAll.stream
-            )
+            .flatMap {
+              case true => Pull.raiseError(BodyAlreadyConsumedError())
+              case false => Pull.output(Chunk.array(buffer)) >> readAll
+            }
+            .stream
 
         val drain: Drain[F] = (None: Option[Array[Byte]]).pure[F]
 
--- a/ember-core/src/main/scala/org/http4s/ember/core/Parser.scala
+++ b/ember-core/src/main/scala/org/http4s/ember/core/Parser.scala
@@ -283,7 +283,13 @@ private[ember] object Parser {
     def parser[F[_]: Concurrent](maxHeaderSize: Int)(
         buffer: Array[Byte],
         read: Read[F]
-    ): F[(Response[F], Drain[F])] =
+    ): F[(Response[F], Drain[F])] = {
+      // per https://httpwg.org/specs/rfc7230.html#rfc.section.3.3.3
+      def expectNoBody(status: Status): Boolean =
+        status == Status.NoContent ||
+          status == Status.NotModified ||
+          status.responseClass == Status.Informational
+
       for {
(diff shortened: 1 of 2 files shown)


danicheg (approved):

ember-core/src/main/scala/org/http4s/ember/core/Parser.scala
```diff
@@ -432,6 +444,30 @@ private[ember] object Parser {
...
+          }
+
+        val body =
+          Stream
+            .eval(consumed.get)
+            .ifM(
+              ifTrue = Stream.raiseError(BodyAlreadyConsumedError()),
+              ifFalse = Stream.chunk(Chunk.array(buffer)) ++ readAll.stream
+            )
+
```
diesalbla, line 465: Perhaps this would also work, 

```suggestion
          Pull.eval(consumed.get).flatMap {
            case true =>  Pull.raiseError(BodyAlreadyConsumedError())
            case false => Pull.output(Chunk.array(buffer)) >> readAll
          }.stream
```

ember-core/src/main/scala/org/http4s/ember/core/Parser.scala
```diff
@@ -432,6 +444,30 @@ private[ember] object Parser {
         (EmptyBody.covary[F], (Some(buffer): Option[Array[Byte]]).pure[F]).pure[F]
       }
 
+    def parseUnknownBody[F[_]: Concurrent](
```
diesalbla, line 447: Should this be marked private, to avoid it being binary-published?

wemrysi: I'm not sure, what is the concern there? Happy to change it to whatever is preferred, I was just taking cues from `parseFixedBody`.

ember-core/src/main/scala/org/http4s/ember/core/Parser.scala
wemrysi: They appear equivalent, but if expressing via `Pull` is preferable, I will change.

rossabaker (commented):
Thanks.  0.21 is officially EOL, so I'd encourage you to upgrade ASAP.  No maintainers are aggressively backporting other bugfixes.  But I also don't mind merging and releasing when a user sends us a working PR like this.  I have legacy stuff, too! :smile:

ember-core/src/main/scala/org/http4s/ember/core/Parser.scala
rossabaker: This might be a few less allocations and is still legible.  Mild :+1:

ember-core/src/main/scala/org/http4s/ember/core/Parser.scala
wemrysi: >  But it doesn't appear to be used anywhere outside

It isn't used outside `Parser`, but it is defined in `Body` and used in `Response`. It seems appropriate to define in `Body`, so we could make it `private[Parser]` but that doesn't seem to generate the same unused compiler warnings as `private` does.

wemrysi:
> Thanks. 0.21 is officially EOL, so I'd encourage you to upgrade ASAP. No maintainers are aggressively backporting other bugfixes. But I also don't mind merging and releasing when a user sends us a working PR like this. I have legacy stuff, too! 😄

For sure, I would have respected you closing the PR with a "won't merge", tbh, but figured I'd try, so your consideration is much appreciated. Legacy stuff is the best 😅

ember-core/src/main/scala/org/http4s/ember/core/Parser.scala
rossabaker: :+1: Let's just leave it alone then.

PR #6137, opened 2022-03-17
Add Enhanced Client Translation

ChristopherDavenport:
So, lets say you want to write a middleware that targets handling with UnexpectedStatus, or DecodingFailure? Or something else that targets the behavior surfaced in `use` of the resource in on of our methods. Well there's no way to do so. 

This super-powers the mapK implementation by giving it the ability to map back and forth in the use of options like expectOr, expectOption, expect, status, etc.

These are all failures that are raised an managed by the client in the `use` of the resource, so they are not mapped today. No longer!


rossabaker (approved):
Functionality looks good.  These nits can be cleaned up later:

* If the scaladoc isn't adding anything novel, we should just inherit it so it's in one place.
* Seems like the several that don't mapK should get a default implementation from somewhere instead of copying.

Issue #4262, opened 2021-01-25
DecodeSpec property issue


rossabaker:
On https://github.com/http4s/http4s/pull/4259/checks?check_run_id=1758604037

```
[info] DecodeSpec
[info] decode should
[error]   x be consistent with utf8Decode (2 seconds, 264 ms)
[error]    Falsified after 97 passed tests.
[error]    > ARG_0: "î༗ഌൎோ๚óచઝೡ੣൶Ó౏མັघ౷ฉ౿Çூ­ଈຉ৶﻿"
[error]    > ARG_1: 1
[error]    > ARG_0_ORIGINAL: "쳍全鵤駐퐔凉왦Ӱ벮䧶㱟탂퓩벲༶솠꧓츔꘹귊᫜ᡙ얳‮흭Ó承㵠䣇㇟ㇼ̟밬ʶ塃齣﻿"
[error]    > ARG_1_ORIGINAL: 2147483647
[error]    The seed is czYXr54iEAm1ARcjHtrqOZyh4fV1BBh4CPLL2WfJu0I=
[error]    
[error]    > Right(î༗ഌൎோ๚óచઝೡ੣൶Ó౏མັघ౷ฉ౿Çூ­ଈຉ৶) != Right(î༗ഌൎோ๚óచઝೡ੣൶Ó౏མັघ౷ฉ౿Çூ­ଈຉ৶﻿) (DecodeSpec.scala:46)
```

rossabaker:
[Another](https://github.com/http4s/http4s/pull/5257/checks?check_run_id=3695932555#step:10:2690):

The array doesn't print itself nicely, but it could be chased from the seed.

```
==> X org.http4s.DecodeSpec.decode should either succeed or raise a CharacterCodingException  0.031s munit.FailException: /home/runner/work/http4s/http4s/tests/src/test/scala/org/http4s/DecodeSpec.scala:144
143:
144:    test("decode should either succeed or raise a CharacterCodingException") {
145:      forAll { (bs: Array[Byte], cs: Charset) =>

Failing seed: gPC4aM6_-CK6CCi4CrdOn3bbjcm0U0k5NRfDpb_vAeC=
You can reproduce this failure by adding the following override to your suite:

  override val scalaCheckInitialSeed = "gPC4aM6_-CK6CCi4CrdOn3bbjcm0U0k5NRfDpb_vAeC="

Falsified after 2 passed tests.
> ARG_0: [B@21e21381
[...]

PR #6138, opened 2022-03-18
Resolve #6068 digestauth challenge redux

### LINKED ISSUE #6068: Currently impossible to use DigestAuth.challenge due to private argument types

blast-hardcheese:
Resolves #6068

Attempting to address PR feedback in #6077

### Refactoring `NonceKeeper`

- Removing `synchronized` calls
- Switching `var` to `Ref`
- Making all exposed methods `F[_]`
- Splitting out `AuthStore` trait with a more secure scheme that doesn't require the user storing the user's passwords in plaintext

Some remaining tasks:

- [x] Performance testing
- [x] Push `F[_]` further: remove remaining `var`, `while`/`tailrec`/`return`

Opening this draft early on to communicate intent and direction, as well to get feedback early

### DIFF
--- a/server/src/main/scala/org/http4s/server/middleware/authentication/DigestAuth.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/authentication/DigestAuth.scala
@@ -102,7 +102,7 @@ object DigestAuth {
   /** Side-effect of running the returned task: If req contains a valid
     * AuthorizationHeader, the corresponding nonce counter (nc) is increased.
     */
-  private def challenge[F[_], A](
+  private[authentication] def challenge[F[_], A](
       realm: String,
       store: AuthenticationStore[F, A],
       nonceKeeper: NonceKeeper,
--- a/server/src/main/scala/org/http4s/server/middleware/authentication/DigestAuth.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/authentication/DigestAuth.scala
@@ -24,6 +24,7 @@ import cats.data.Kleisli
 import cats.data.NonEmptyList
 import cats.effect.Concurrent
 import cats.effect.Sync
+import cats.effect.Timer
 import cats.syntax.all._
 import org.http4s.crypto.Hash
 import org.http4s.crypto.unsafe.SecureRandom
@@ -31,7 +32,6 @@ import org.http4s.headers._
 
 import java.math.BigInteger
 import java.util.Date
-import scala.annotation.nowarn
 import scala.concurrent.duration._
 
 /** Provides Digest Authentication from RFC 2617.
@@ -54,19 +54,6 @@ object DigestAuth {
   private case object NoCredentials extends AuthReply[Nothing]
   private case object NoAuthorizationHeader extends AuthReply[Nothing]
 
-  @deprecated("Calling apply is side-effecting, please use applyF", "0.22.11")
-  def apply[F[_]: Sync, A](
-      realm: String,
-      store: AuthenticationStore[F, A],
-      nonceCleanupInterval: Duration = 1.hour,
-      nonceStaleTime: Duration = 1.hour,
-      nonceBits: Int = 160,
-  ): AuthMiddleware[F, A] = {
-    val nonceKeeper =
-      new NonceKeeper(nonceStaleTime.toMillis, nonceCleanupInterval.toMillis, nonceBits)
-    challenged(challenge(realm, store, nonceKeeper))
-  }
-
   /** @param realm The realm used for authentication purposes.
     * @param store A partial function mapping (realm, user) to the
     *              appropriate password.
@@ -77,7 +64,7 @@ object DigestAuth {
     *                       purposes anymore).
     * @param nonceBits The number of random bits a nonce should consist of.
     */
-  def applyF[F[_]: Concurrent, A](
+  def apply[F[_]: Concurrent: Timer, A](
       realm: String,
       store: AuthenticationStore[F, A],
       nonceCleanupInterval: Duration = 1.hour,
@@ -97,6 +84,9 @@ object DigestAuth {
   /** Similar to [[apply]], but exposing the underlying [[challenge]]
     * [[cats.data.Kleisli]] instead of an entire [[AuthMiddleware]]
     *
+    * Side-effect of running the returned task: If req contains a valid
+    * AuthorizationHeader, the corresponding nonce counter (nc) is increased.
+    *
     * @param realm The realm used for authentication purposes.
     * @param store A partial function mapping (realm, user) to the
     *              appropriate password.
@@ -107,44 +97,32 @@ object DigestAuth {
     *                       purposes anymore).
     * @param nonceBits The number of random bits a nonce should consist of.
     */
-  @nowarn
-  def challenge[F[_], A](
+  def challenge[F[_]: Timer, A](
       realm: String,
       store: AuthenticationStore[F, A],
       nonceCleanupInterval: Duration = 1.hour,
       nonceStaleTime: Duration = 1.hour,
       nonceBits: Int = 160,
   )(implicit F: Concurrent[F]): F[Kleisli[F, Request[F], Either[Challenge, AuthedRequest[F, A]]]] =
-    F.delay(new NonceKeeper(nonceStaleTime.toMillis, nonceCleanupInterval.toMillis, nonceBits))
+    NonceKeeper[F](nonceStaleTime.toMillis, nonceCleanupInterval.toMillis, nonceBits)
       .map { nonceKeeper =>
-        challenge[F, A](realm, store, nonceKeeper)
-      }
+        Kleisli { req =>
+          def paramsToChallenge(params: Map[String, String]) =
+            Either.left(Challenge("Digest", realm, params))
 
-  /** Side-effect of running the returned task: If req contains a valid
-    * AuthorizationHeader, the corresponding nonce counter (nc) is increased.
-    */
-  @deprecated(
-    "Calling challenge is side-effecting, please use the F[_]-suspended variant",
-    "0.22.11",
-  )
-  def challenge[F[_], A](realm: String, store: AuthenticationStore[F, A], nonceKeeper: NonceKeeper)(
-      implicit F: Sync[F]
-  ): Kleisli[F, Request[F], Either[Challenge, AuthedRequest[F, A]]] =
-    Kleisli { req =>
-      def paramsToChallenge(params: Map[String, String]) =
-        Either.left(Challenge("Digest", realm, params))
-
-      checkAuth(realm, store, nonceKeeper, req).flatMap {
-        case OK(authInfo) => F.pure(Either.right(AuthedRequest(authInfo, req)))
-        case StaleNonce => getChallengeParams(nonceKeeper, staleNonce = true).map(paramsToChallenge)
-        case _ => getChallengeParams(nonceKeeper, staleNonce = false).map(paramsToChallenge)
+          checkAuth(realm, store, nonceKeeper, req).flatMap {
+            case OK(authInfo) => F.pure(Either.right(AuthedRequest(authInfo, req)))
+            case StaleNonce =>
+              getChallengeParams(nonceKeeper, staleNonce = true).map(paramsToChallenge)
+            case _ => getChallengeParams(nonceKeeper, staleNonce = false).map(paramsToChallenge)
+          }
+        }
       }
-    }
 
   private def checkAuth[F[_]: Hash, A](
       realm: String,
       store: AuthenticationStore[F, A],
-      nonceKeeper: NonceKeeper,
+      nonceKeeper: NonceKeeper[F],
       req: Request[F],
   )(implicit F: Monad[F]): F[AuthReply[A]] =
     req.headers.get[Authorization] match {
@@ -156,22 +134,23 @@ object DigestAuth {
         F.pure(NoAuthorizationHeader)
     }
 
-  private def getChallengeParams[F[_]](nonceKeeper: NonceKeeper, staleNonce: Boolean)(implicit
+  private def getChallengeParams[F[_]](nonceKeeper: NonceKeeper[F], staleNonce: Boolean)(implicit
       F: Sync[F]
   ): F[Map[String, String]] =
-    F.delay {
-      val nonce = nonceKeeper.newNonce()
-      val m = Map("qop" -> "auth", "nonce" -> nonce)
-      if (staleNonce)
-        m + ("stale" -> "TRUE")
-      else
-        m
-    }
+    nonceKeeper
+      .newNonce()
+      .map { nonce =>
+        val m = Map("qop" -> "auth", "nonce" -> nonce)
+        if (staleNonce)
+          m + ("stale" -> "TRUE")
+        else
+          m
+      }
 
   private def checkAuthParams[F[_]: Hash, A](
       realm: String,
       store: AuthenticationStore[F, A],
-      nonceKeeper: NonceKeeper,
+      nonceKeeper: NonceKeeper[F],
       req: Request[F],
       paramsNel: NonEmptyList[(String, String)],
   )(implicit F: Monad[F]): F[AuthReply[A]] = {
@@ -187,7 +166,7 @@ object DigestAuth {
 
     val nonce = params("nonce")
     val nc = params("nc")
-    nonceKeeper.receiveNonce(nonce, Integer.parseInt(nc, 16)) match {
+    nonceKeeper.receiveNonce(nonce, Integer.parseInt(nc, 16)).flatMap {
       case NonceKeeper.StaleReply => F.pure(StaleNonce)
       case NonceKeeper.BadNCReply => F.pure(BadNC)
       case NonceKeeper.OKReply =>
--- a/server/src/main/scala/org/http4s/server/middleware/authentication/DigestAuth.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/authentication/DigestAuth.scala
@@ -22,6 +22,7 @@ package authentication
 import cats.Monad
 import cats.data.Kleisli
 import cats.data.NonEmptyList
+import cats.effect.Concurrent
 import cats.effect.Sync
 import cats.syntax.all._
 import org.http4s.crypto.Hash
@@ -30,6 +31,7 @@ import org.http4s.headers._
 
 import java.math.BigInteger
 import java.util.Date
+import scala.annotation.nowarn
 import scala.concurrent.duration._
 
 /** Provides Digest Authentication from RFC 2617.
@@ -52,16 +54,7 @@ object DigestAuth {
   private case object NoCredentials extends AuthReply[Nothing]
   private case object NoAuthorizationHeader extends AuthReply[Nothing]
 
-  /** @param realm The realm used for authentication purposes.
-    * @param store A partial function mapping (realm, user) to the
-    *              appropriate password.
-    * @param nonceCleanupInterval Interval (in milliseconds) at which stale
-    *                             nonces should be cleaned up.
-    * @param nonceStaleTime Amount of time (in milliseconds) after which a nonce
-    *                       is considered stale (i.e. not used for authentication
-    *                       purposes anymore).
-    * @param nonceBits The number of random bits a nonce should consist of.
-    */
+  @deprecated("Calling apply is side-effecting, please use applyF", "0.22.11")
   def apply[F[_]: Sync, A](
       realm: String,
       store: AuthenticationStore[F, A],
@@ -73,6 +66,33 @@ object DigestAuth {
       new NonceKeeper(nonceStaleTime.toMillis, nonceCleanupInterval.toMillis, nonceBits)
     challenged(challenge(realm, store, nonceKeeper))
   }
+
+  /** @param realm The realm used for authentication purposes.
+    * @param store A partial function mapping (realm, user) to the
+    *              appropriate password.
+    * @param nonceCleanupInterval Interval (in milliseconds) at which stale
+    *                             nonces should be cleaned up.
+    * @param nonceStaleTime Amount of time (in milliseconds) after which a nonce
+    *                       is considered stale (i.e. not used for authentication
+    *                       purposes anymore).
+    * @param nonceBits The number of random bits a nonce should consist of.
+    */
+  def applyF[F[_]: Concurrent, A](
+      realm: String,
+      store: AuthenticationStore[F, A],
+      nonceCleanupInterval: Duration = 1.hour,
+      nonceStaleTime: Duration = 1.hour,
+      nonceBits: Int = 160,
+  ): F[AuthMiddleware[F, A]] =
+    challenge[F, A](
+      realm = realm,
+      store = store,
+      nonceCleanupInterval = nonceCleanupInterval,
+      nonceStaleTime = nonceStaleTime,
+      nonceBits = nonceBits,
+    ).map { runChallenge =>
+      challenged(runChallenge)
+    }
 
   /** Similar to [[apply]], but exposing the underlying [[challenge]]
     * [[cats.data.Kleisli]] instead of an entire [[AuthMiddleware]]
@@ -87,13 +107,14 @@ object DigestAuth {
     *                       purposes anymore).
     * @param nonceBits The number of random bits a nonce should consist of.
     */
+  @nowarn
   def challenge[F[_], A](
       realm: String,
       store: AuthenticationStore[F, A],
       nonceCleanupInterval: Duration = 1.hour,
       nonceStaleTime: Duration = 1.hour,
       nonceBits: Int = 160,
-  )(implicit F: Sync[F]): F[Kleisli[F, Request[F], Either[Challenge, AuthedRequest[F, A]]]] =
+  )(implicit F: Concurrent[F]): F[Kleisli[F, Request[F], Either[Challenge, AuthedRequest[F, A]]]] =
     F.delay(new NonceKeeper(nonceStaleTime.toMillis, nonceCleanupInterval.toMillis, nonceBits))
       .map { nonceKeeper =>
         challenge[F, A](realm, store, nonceKeeper)
@@ -102,12 +123,12 @@ object DigestAuth {
   /** Side-effect of running the returned task: If req contains a valid
     * AuthorizationHeader, the corresponding nonce counter (nc) is increased.
     */
-  private[authentication] def challenge[F[_], A](
-      realm: String,
-      store: AuthenticationStore[F, A],
-      nonceKeeper: NonceKeeper,
-  )(implicit
-      F: Sync[F]
+  @deprecated(
+    "Calling challenge is side-effecting, please use the F[_]-suspended variant",
+    "0.22.11",
+  )
+  def challenge[F[_], A](realm: String, store: AuthenticationStore[F, A], nonceKeeper: NonceKeeper)(
+      implicit F: Sync[F]
   ): Kleisli[F, Request[F], Either[Challenge, AuthedRequest[F, A]]] =
     Kleisli { req =>
       def paramsToChallenge(params: Map[String, String]) =
--- a/server/src/main/scala/org/http4s/server/middleware/authentication/DigestAuth.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/authentication/DigestAuth.scala
@@ -44,7 +44,11 @@ object DigestAuth {
     * which is _strongly_ discouraged. Please use Md5HashedAuthenticationStore
     * if you can.
     */
-  final case class PlainTextAuthenticationStore[F[_], A](func: String => F[Option[(A, String)]])
+  object PlainTextAuthenticationStore {
+    def apply[F[_], A](func: String => F[Option[(A, String)]]): AuthenticationStore[F, A] =
+      new PlainTextAuthenticationStore(func)
+  }
+  final class PlainTextAuthenticationStore[F[_], A](val func: String => F[Option[(A, String)]])
       extends AuthenticationStore[F, A]
 
   /** A function mapping username to a user object and precomputed md5
@@ -53,7 +57,11 @@ object DigestAuth {
     * More secure than PlainTextAuthenticationStore due to only needing to
     * store the digested hash instead of the password in plain text.
     */
-  final case class Md5HashedAuthenticationStore[F[_], A](func: String => F[Option[(A, String)]])
+  object Md5HashedAuthenticationStore {
+    def apply[F[_], A](func: String => F[Option[(A, String)]]): AuthenticationStore[F, A] =
+      new Md5HashedAuthenticationStore(func)
+  }
+  final class Md5HashedAuthenticationStore[F[_], A](val func: String => F[Option[(A, String)]])
       extends AuthenticationStore[F, A]
 
   private trait AuthReply[+A]
@@ -219,8 +227,8 @@ object DigestAuth {
           case NonceKeeper.BadNCReply => F.pure(BadNC)
           case NonceKeeper.OKReply =>
             (store match {
-              case PlainTextAuthenticationStore(func) =>
-                func(params("username")).flatMap {
+              case authStore: PlainTextAuthenticationStore[F, A] =>
+                authStore.func(params("username")).flatMap {
                   case None => F.pure(UserUnknown)
                   case Some((authInfo, password)) =>
                     DigestUtil
@@ -240,8 +248,8 @@ object DigestAuth {
                         else WrongResponse
                       }
                 }
-              case Md5HashedAuthenticationStore(func) =>
-                func(params("username")).flatMap {
+              case authStore: Md5HashedAuthenticationStore[F, A] =>
+                authStore.func(params("username")).flatMap {
                   case None => F.pure(UserUnknown)
                   case Some((authInfo, ha1Hash)) =>
                     DigestUtil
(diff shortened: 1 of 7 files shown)


armanbilge:
Took a quick pass through—excellent work, this is looking very nice 👍

armanbilge:
A couple other notes:
1. Annoying, but I think we have to keep the old `NonceKeeper` as well, because we need it to implement the deprecated methods in `DigestAuth` that are not using `F[_]` to preserve backwards-compatibility. Just like you did in your other PR.
2. Don't worry about performance testing, but it would be nice to clean up those remaining imperative stuff.

blast-hardcheese:
Performance analysis, I ran [this script](https://gist.github.com/blast-hardcheese/5bda592624b5842373997f71e2622df8) against efadfc99d0 (origin/series/0.22) and c666a7fc2a (this branch):

```diff
-BasicAuthentication should respond to a request with correct credentials: 15ms
+BasicAuthentication should respond to a request with correct credentials: 19ms
-BasicAuthentication should respond to a request with unknown username with 401: 4ms
+BasicAuthentication should respond to a request with unknown username with 401: 6ms
-BasicAuthentication should respond to a request with wrong password with 401: 1ms
+BasicAuthentication should respond to a request with wrong password with 401: 1ms
-BasicAuthentication should respond to a request without authentication with 401: 20ms
+BasicAuthentication should respond to a request without authentication with 401: 24ms
-DigestAuthentication should avoid many concurrent replay attacks: 16ms
+DigestAuthentication should avoid many concurrent replay attacks: 31ms
-DigestAuthentication should respond to a request with correct credentials: 29ms
+DigestAuthentication should respond to a request with correct credentials: 40ms
-DigestAuthentication should respond to a request without authentication with 401: 8ms
+DigestAuthentication should respond to a request without authentication with 401: 49ms
-DigestAuthentication should respond to invalid requests with 401: 17ms
+DigestAuthentication should respond to invalid requests with 401: 7ms
-DigestAuthentication should respond to many concurrent requests while cleaning up nonces: 85ms
+DigestAuthentication should respond to many concurrent requests while cleaning up nonces: 140ms
```

(the script just runs `server / testOnly org.http4s.server.middleware.authentication.AuthenticationSuite` multiple times in a loop, then averages the resulting times) 

I'm somewhat confused by the directive of "don't worry about performance", but I'm heartened by my largely amateurish attempt not having _too_ negative of a performance impact. I'm available and interested in idioms or suggestions to improve performance, if necessary. `mimaReportBinaryIssues` reports success as well, which is also great.

server/src/main/scala/org/http4s/server/middleware/authentication/DigestAuth.scala
```diff
@@ -52,6 +51,19 @@ object DigestAuth {
   private case object NoCredentials extends AuthReply[Nothing]
   private case object NoAuthorizationHeader extends AuthReply[Nothing]
 
+  @deprecated("Calling apply is side-effecting, please use applyF", "0.22.11")
```
rossabaker, line 54:
```suggestion
  @deprecated("Calling apply is side-effecting, please use applyF", "0.22.13")
```

server/src/main/scala/org/http4s/server/middleware/authentication/NonceF.scala
```diff
@@ -0,0 +1,36 @@
...
+
+import java.math.BigInteger
+import java.util.Date
+
+private[authentication] class NonceF[F[_]](val created: Date, val nc: Ref[F, Int], val data: String)
+
+private[authentication] object NonceF {
+  val random = new SecureRandom()
+
+  private def getRandomData(bits: Int): String = new BigInteger(bits, random).toString(16)
```
armanbilge, line 32: Yikes, this is another icky one, see discussion in https://github.com/typelevel/cats-effect/issues/2865#issuecomment-1061950448. Note that generating (secure) random numbers is a side-effect and potentially blocking.

I'm not sure what's the best way to fix this, between the global state and my lack of familiarity with CE2. @rossabaker do you have any recommendations?

rossabaker, line 32: This is potentially a blocking call.  It's going to be fine for most cases, but have the same problems on exotic platforms as as https://github.com/typelevel/cats-effect/issues/2882.

Probably not urgent: it was that way before, and nobody noticed enough to report it.

server/src/main/scala/org/http4s/server/middleware/authentication/NonceKeeperF.scala
```diff
@@ -0,0 +1,124 @@
...
+  */
+private[authentication] class NonceKeeperF[F[_]: Timer](
+    staleTimeout: Long,
+    nonceCleanupInterval: Long,
+    bits: Int,
+    semaphore: Semaphore[F],
+    lastCleanup: Ref[F, Long],
+)(implicit F: Concurrent[F]) {
+  require(bits > 0, "Please supply a positive integer for bits.")
+  private val nonces = new LinkedHashMap[String, NonceF[F]]
```
armanbilge, line 57: I wonder if we should move this into the constructor. Creating a new instance of a class is easily assumed to be side-effect free, and this violates that.

server/src/main/scala/org/http4s/server/middleware/authentication/NonceKeeperF.scala
```diff
@@ -0,0 +1,124 @@
...
+
+  val clock = Clock[F]
+
+  /** Removes nonces that are older than staleTimeout
+    * Note: this _MUST_ be executed with the singleton permit from the semaphore
+    */
+  private def checkStale(): F[Unit] =
```
armanbilge, line 64: Scaladoc is good but this really gets it across!

```suggestion
  private def unsafeCheckStale(): F[Unit] =
```

server/src/main/scala/org/http4s/server/middleware/authentication/DigestAuth.scala
```diff
@@ -62,115 +74,153 @@ object DigestAuth {
...
+      store,
+      F.delay(nonceKeeper.newNonce()),
+      (data, nc) => F.delay(nonceKeeper.receiveNonce(data, nc)),
+    )
+
+  /** Similar to [[apply]], but exposing the underlying [[challenge]]
+    * [[cats.data.Kleisli]] instead of an entire [[AuthMiddleware]]
+    *
+    * Side-effect of running the returned task: If req contains a valid
+    * AuthorizationHeader, the corresponding nonce counter (nc) is increased.
```
armanbilge, line 108:
```suggestion
    * Side-effect of the middleware: If req contains a valid
    * AuthorizationHeader, the corresponding nonce counter (nc) is increased.
```

rossabaker (commented):
You've been in good hands with others here, so I only skimmed it.  Looking good, though.  I imagine we'll do another round of releases this upcoming week, and I'd love to get this in.

On the subject of "don't worry about performance", I haven't followed the conversation closely enough to see where it came up.  I'd say we do strive for performance, but correctness comes first, and we don't generally trade legibility or purity for performance without a profiled hotspot.  Benchmarking and optimizing is fun, but when we're all mostly writing services that listen on a socket and then make a remote database query and then enrich that with a client call to another web service, an extra flatMap here or there is a drop in the ocean.

armanbilge (commented):
Ross beat my review by a few minutes, so apologies for the duplicated comments 😆 

I made the comment about performance in https://github.com/http4s/http4s/pull/6138#issuecomment-1072343000 (emphasis added):
> Don't worry about performance _testing_

What I meant was, we wouldn't block your PR due to lack of benchmarks :) for sure, we don't want you to _tank_ performance! Also, meaningful benchmarking is not easy: you have to account for JVM warm-up and JIT-optimizations, etc.

server/src/main/scala/org/http4s/server/middleware/authentication/NonceF.scala
armanbilge: Yeah, this caught my eye too. I guess if it's okay we can just leave it, although @blast-hardcheese make sure to wrap this in an `F.delay` b/c generating random numbers is a side-effect.

blast-hardcheese:
Thanks for the reviews, I've got my work cut out for me!

In taking the next step in verifying my work here, I ran across [`Requires that the server can recover the password in clear text, which is _strongly_ discouraged.`](https://github.com/http4s/http4s/blob/series/0.22/server/src/main/scala/org/http4s/server/middleware/authentication/DigestAuth.scala#L39-L43), which would give the impression that this whole class is not intended for even limited production use in real systems, as it's impossible to use `DigestAuth` without supplying an instance of that function, but the function itself is documented as being strongly discouraged.

In an attempt to resolve this, I've pushed up `Flesh out AuthenticationStore`, a bincompat change that promotes a type alias to a sealed trait, preserving the previous functionality but offering a more secure variant.

server/src/main/scala/org/http4s/server/middleware/authentication/NonceF.scala
blast-hardcheese: Resolved in the other comment

server/src/main/scala/org/http4s/server/middleware/authentication/DigestAuth.scala
blast-hardcheese: Critically, this _isn't_ an `AuthMiddleware`, it's just a `Kleisli` -- this is more of a proto-middleware, do you still feel strongly about this change?

server/src/main/scala/org/http4s/server/middleware/authentication/NonceKeeperF.scala
blast-hardcheese: I've always had `unsafe` mapped to "side-effecting without F[_]", but I guess the only way to make this safe is to enrich the semaphore with a value, like `Resource`. Good suggestion, that comment was bugging me as well.

blast-hardcheese:
> I made the comment about performance in [#6138 (comment)](https://github.com/http4s/http4s/pull/6138#issuecomment-1072343000) (emphasis added):
> 
> > Don't worry about performance _testing_
> 
> What I meant was, we wouldn't block your PR due to lack of benchmarks :) for sure, we don't want you to _tank_ performance! Also, meaningful benchmarking is not easy: you have to account for JVM warm-up and JIT-optimizations, etc.

That makes more sense 😅

I figured enough runs of the auth suite would be sufficient ballpark, which is why I included them, since this is (from my vantage point) a significant enough change to warrant them, though Ross' point about an extra `flatMap` here or there was useful for keeping from spiraling off into the weeds.

I've updated the PR with feedback, and will update again shortly once I've finished this test for the `Md5HashedAuthenticationStore`.

server/src/main/scala/org/http4s/server/middleware/authentication/NonceF.scala
rossabaker: Arman brought up how it's difficult to do "right" in CE2, and that's true.  `F.delay` is appropriate for the vast majority of users, but `F.blocking` is required in more exotic cases.  In CE3, the interfaces are the same, so we can detect the exotic cases and block only where needed.  In CE2, we can also detect, but we'd need to lug around a `Blocker` just in case.

I'll see if I can draft a quick helper.  The logic will be mostly what we need for that CE3 ticket anyway.

server/src/main/scala/org/http4s/server/middleware/authentication/DigestAuth.scala
armanbilge: I'm sorry, you're completely right, I was mixing up my terminology. A middleware is not a `Kleisli` but a `Kleisli => Kleisli`. Ignore me :)

server/src/main/scala/org/http4s/server/middleware/authentication/DigestAuth.scala
```diff
@@ -35,11 +35,22 @@ import scala.concurrent.duration._
   */
 object DigestAuth {
 
+  type PlainAuthenticationStore[F[_], A] = String => F[Option[(A, String)]]
+  sealed trait AuthenticationStore[F[_], A]
```
armanbilge, line 39: A `type` alias has no binary signature, so promoting to a `trait` is binary-compatible. However, it would be source-breaking ... but maybe that's okay, since the benefit is tangible.

I'm not sure if we need the `PlainAuthenticationStore` alias, we can just use `String => F[Option[(A, String)]]` and deprecate anything using that anyway.

server/src/main/scala/org/http4s/server/middleware/authentication/DigestAuth.scala
```diff
@@ -35,11 +35,22 @@ import scala.concurrent.duration._
...
     * None if no user exists.  Requires that the server can recover
     * the password in clear text, which is _strongly_ discouraged.
     */
-  type AuthenticationStore[F[_], A] = String => F[Option[(A, String)]]
+  final case class PlainTextAuthenticationStore[F[_], A](func: String => F[Option[(A, String)]])
+      extends AuthenticationStore[F, A]
+
+  /** A function mapping username to a user object and precomputed md5
+    * hash of the username, realm, and password, or None if no user exists.
+    * More secure than PlainTextAuthenticationStore, due to .
```
armanbilge, line 50: due to? :)

server/src/main/scala/org/http4s/server/middleware/authentication/DigestAuth.scala
```diff
@@ -35,11 +35,22 @@ import scala.concurrent.duration._
...
+  type PlainAuthenticationStore[F[_], A] = String => F[Option[(A, String)]]
+  sealed trait AuthenticationStore[F[_], A]
+
   /** A function mapping username to a user object and password, or
     * None if no user exists.  Requires that the server can recover
     * the password in clear text, which is _strongly_ discouraged.
     */
-  type AuthenticationStore[F[_], A] = String => F[Option[(A, String)]]
+  final case class PlainTextAuthenticationStore[F[_], A](func: String => F[Option[(A, String)]])
+      extends AuthenticationStore[F, A]
```
armanbilge, line 46: Annoying, but if we ever need to change the `case class` signature it is very difficult to retain bincompat across all 3 scala versions.

What if we do like:
```scala
trait AuthenticationStore[F[_], A] {
  def get(username: String): F[Option[(A, String)]]
}

sealed trait PlainTextAuthenticationStore[F[_], A] extends AuthenticationStore[F, A]
object PlainTextAuthenticationStore {
  def apply[F[_], A](func: String => F[Option[(A, String)]]) = ???
}

// etc.
```

blast-hardcheese: Ah, yeah, that's a good point. Will update.

server/src/main/scala/org/http4s/server/middleware/authentication/DigestAuth.scala
blast-hardcheese: Yeah, I think the self-documenting nature of the compiler error is worth the source compat breakage. I was thinking of doing something complex like

```scala
sealed class AuthenticationStore[F[_], A] extends Function1[String, F[Option[(A, String)]]] { ... }
```

in order to maintain some sort of compatibility, but it just made everything way too complicated for negligible benefit.

armanbilge: Could we call the new one `AuthStore` instead, and deprecate the `AuthenticationStore` alias?

blast-hardcheese:
Alright, this is good to go, barring one final thought:

Without knowing _how_ to precompute the `ha1`, I perceive this sort of like an open loop; we're expecting the users to just know how to do that, reimplementing a part of the spec in their own service.

I added a `precomputeHash(username: String, realm: String, password: String): F[String]` convenience function attached to the `Md5HashedAuthStore` companion object to resolve this usability gap, though I'm open to removing it if it's deemed unnecessary.

server/src/main/scala/org/http4s/server/middleware/authentication/DigestAuth.scala
```diff
@@ -22,25 +22,60 @@ package authentication
...
   */
 object DigestAuth {
 
+  @deprecated(
+    "AuthenticationStore is going away, in favor of explicit subclasses of AuthStore. PlainTextAuthStore maintains the previous, insecure behaviour, whereas Md5HashedAuthStore is the new advised implementation going forward.",
+    "0.22.13",
+  )
+  type AuthenticationStore[F[_], A] = String => F[Option[(A, String)]]
+
+  sealed trait AuthStore[F[_], A]
```
rossabaker, line 44: Does this need to be sealed?  I see we match on them in `checkAuthParams`.  Would it be better if those cases were a method on this, to permit more pluggable implementations?  Or are plaintext and MD5 the two that are specified?

server/src/main/scala/org/http4s/server/middleware/authentication/DigestAuth.scala
```diff
@@ -62,115 +110,179 @@ object DigestAuth {
     *                       purposes anymore).
     * @param nonceBits The number of random bits a nonce should consist of.
     */
-  def apply[F[_]: Sync, A](
+  def applyF[F[_]: Concurrent: Timer, A](
```
rossabaker, line 113: Do you want this to be `applyF` or `apply` in the long run?  I don't generally like overloads, but I think they're compatible.

server/src/main/scala/org/http4s/server/middleware/authentication/DigestUtil.scala
```diff
@@ -30,6 +30,43 @@ private[authentication] object DigestUtil {
...
+    * @param nonce
+    * @param nc
+    * @param cnonce
+    * @param qop
+    * @return
+    */
+  def computeHashedResponse[F[_]: Monad: Hash](
+      method: String,
+      ha1: String,
+      uri: String,
```
rossabaker, line 46: Is this the path or the full URI?  I haven't connected the dots.  There are case-insensitive bits to a full URI.  I'm wondering what the spec requires here.

server/src/main/scala/org/http4s/server/middleware/authentication/Nonce.scala
```diff
@@ -0,0 +1,32 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.http4s.server.middleware.authentication
+
+import org.http4s.crypto.unsafe.SecureRandom
+
+import java.math.BigInteger
+import java.util.Date
```
rossabaker, line 22: Is this better than a `java.time.Instant`?  That's what we use most places.  It's not public API, so not a huge deal.

server/src/main/scala/org/http4s/server/middleware/authentication/NonceKeeperF.scala
```diff
@@ -0,0 +1,126 @@
...
+
+private[authentication] object NonceKeeperF {
+  def apply[F[_]: Timer](
+      staleTimeout: Long,
+      nonceCleanupInterval: Long,
+      bits: Int,
+  )(implicit F: Concurrent[F]): F[NonceKeeperF[F]] = for {
+    // This semaphore controls who has access to `nonces` during stale nonce eviction. This must never be set above one.
+    semaphore <- Semaphore[F](1)
+    currentTime <- Clock[F].realTime(MILLISECONDS)
```
rossabaker, line 38: `realTime` is prone to non-monotonicity.  If this is about intervals and not about wall time, we might want the monotonic time.

server/src/main/scala/org/http4s/server/middleware/authentication/NonceKeeperF.scala
```diff
@@ -0,0 +1,126 @@
...
+import cats.syntax.all._
+
+import java.util.LinkedHashMap
+import java.{util => ju}
+import scala.concurrent.duration.MILLISECONDS
+
+private[authentication] object NonceKeeperF {
+  def apply[F[_]: Timer](
+      staleTimeout: Long,
+      nonceCleanupInterval: Long,
```
rossabaker, line 33: A `Duration` would be good here, just to make it clearer what that `Long` represents.

server/src/main/scala/org/http4s/server/middleware/authentication/Nonce.scala
```diff
@@ -0,0 +1,32 @@
...
+
+private[authentication] class Nonce(val created: Date, var nc: Int, val data: String)
```
rossabaker, line 24: I couldn't find in GitHub where this one still gets used.

server/src/main/scala/org/http4s/server/middleware/authentication/NonceF.scala
```diff
@@ -0,0 +1,39 @@
...
+
+private[authentication] class NonceF[F[_]](val created: Date, val nc: Ref[F, Int], val data: String)
+
+private[authentication] object NonceF {
+  val random = new SecureRandom()
+
+  private def getRandomData[F[_]](bits: Int)(implicit F: Sync[F]): F[String] =
+    F.delay(new BigInteger(bits, random).toString(16))
+
+  def gen[F[_]: Sync](bits: Int): F[NonceF[F]] =
```
rossabaker, line 35: This is going to need a `Blocker` and a `ContextShift` in the long run.  This can merge before #6165, but it would be best to get that done before we have a release and need to do a MiMa exemption.

The nice thing is, in Cats-Effect 3, this becomes more transparent in the API.

rossabaker (commented):
This is shaping up well.  I gave it a more thorough review.  Lots of comments, but hopefully none of them stir the pot too much.

server/src/main/scala/org/http4s/server/middleware/authentication/DigestAuth.scala
blast-hardcheese: Based on how I understand the spec, there are only two sensible implementations:

#### Plain-text passwords:

    md5($un:$realm:$pw):$nonce:$nc:$cnonce:$qop:md5($method:$uri)

#### Pre-hashed passwords:

    md5($un:$realm:$pw) -> ha1 in the db
                    |
             retrieve from db
                    v
                   $ha1:$nonce:$nc:$cnonce:$qop:md5($method:$uri)

The way I see it, by pushing more into subclasses of `AuthStore`, we'd need to expose all these parameters for dubious benefit. `ha1` is the only secret that needs to be exposed to user code

server/src/main/scala/org/http4s/server/middleware/authentication/DigestAuth.scala
blast-hardcheese: They're compatible, save for the default arguments. I'd expect this needs to revert back to `apply`, once the non-`F[_]`'d `apply` is deleted, but that'll be a breaking change (possibly with a deprecation cycle, duplicating the signature that just passes through to the underlying `apply`)

server/src/main/scala/org/http4s/server/middleware/authentication/DigestAuth.scala
blast-hardcheese: Reviewing the spec again for another comment, this is actually called out explicitly [by the spec](https://datatracker.ietf.org/doc/html/rfc2617#section-3.3):

> Note that the HTTP server does not actually need to know the user's
> cleartext password. As long as H(A1) is available to the server, the
> validity of an Authorization header may be verified.

server/src/main/scala/org/http4s/server/middleware/authentication/DigestUtil.scala
blast-hardcheese: This is `req.uri.toString()` -- I don't see anything in the spec that explicitly clarifies uri being case sensitive or insensitive.

That being said, considering that http permits different content to be at different uris which could be equal when compared case-insensitively, I think it follows that this should also be case-sensitive.

I could push `Uri` down into the `uri` parameter, if desired.

server/src/main/scala/org/http4s/server/middleware/authentication/Nonce.scala
blast-hardcheese: Swapping this out to `Instant` also let me swap out for `Clock[F].instantNow`, good call

server/src/main/scala/org/http4s/server/middleware/authentication/Nonce.scala
blast-hardcheese: `Nonce` and `NonceKeeper` are the bincompat deprecated variants of `NonceF` and `NonceKeeperF`

server/src/main/scala/org/http4s/server/middleware/authentication/DigestUtil.scala
rossabaker: That might give us flexibility in case we get it wrong.  This just feels like the kind of issue that bites us in the ass later.

> The URI from Request-URI of the Request-Line; duplicated here because proxies are allowed to change the Request-Line in transit.

I don't see anything about normalizing it, so what you've done is _probably_ right.  But we could call `toString` here just as surely as we called `toString` wherever you called it.

server/src/main/scala/org/http4s/server/middleware/authentication/DigestAuth.scala
rossabaker: I was just thinking, if you want to call it `apply` now, you probably could.  Although they both take default arguments, so, actually, I think you can't.  Suggestion withdrawn.

server/src/main/scala/org/http4s/server/middleware/authentication/DigestAuth.scala
rossabaker: > An optional header allows the server to specify the algorithm used to
   create the checksum or digest. By default the MD5 algorithm is used
   and that is the only algorithm described in this document.

It looks like it can be `md5`, `md5-sess`, or extensions that probably nobody understands.  I think it would be unpleasant to register and understand more for little gain.

server/src/main/scala/org/http4s/server/middleware/authentication/NonceF.scala
blast-hardcheese: Sort of interesting -- `NonceKeeperF[F].newNonce()` already requires a permit for the singleton semaphore, so I think acquiring an instance of `Blocker` at that time makes sense. I can keep moving it further up if desired, I really don't have a good intuition around where that instance needs to come from.

server/src/main/scala/org/http4s/server/middleware/authentication/NonceKeeperF.scala
blast-hardcheese: Turns out they were already `Duration` one level up, I presume they were converted to `Long` to avoid having to do the same work again and again.

server/src/main/scala/org/http4s/server/middleware/authentication/DigestUtil.scala
blast-hardcheese: > But we could call toString here just as surely as we called toString wherever you called it.

I'll not take credit for what already existed before, I just broke up a single function that took all the parameters at the same time to a few that take different combinations of the parameters.

I'll make the change.

server/src/main/scala/org/http4s/server/middleware/authentication/DigestAuth.scala
blast-hardcheese: The need to add support for this stuff may motivate new contributors 😉

server/src/main/scala/org/http4s/server/middleware/authentication/NonceF.scala
blast-hardcheese: Does `Blocking[F].use` actually acquire a new thread/threadpool? Should that parameter be bubbled all the way out to the caller's code?

blast-hardcheese:
Seems like an unrelated failure, I spent a few minutes trying to understand what's going on but I'm completely out of my depth.

While the failure message is

```
==> X org.http4s.DecodeSpec.decode should either succeed or raise a CharacterCodingException  0.083s munit.FailException: /Users/dstewart/Projects/deps/http4s/tests/src/test/scala/org/http4s/DecodeSpec.scala:134
133:
134:    test("decode should either succeed or raise a CharacterCodingException") {
135:      forAll { (bs: Array[Byte], cs: Charset) =>

Failing seed: ul4X_OcjscjNgHDzfcislTYeeRBho4GD7b-pOuxHoPJ=

  override val scalaCheckInitialSeed = "ul4X_OcjscjNgHDzfcislTYeeRBho4GD7b-pOuxHoPJ="

Falsified after 91 passed tests.
> ARG_0: [B@7f12f19
> ARG_1: x-COMPOUND_TEXT
> ARG_0_ORIGINAL: [B@55d47e2
[error] Failed: Total 10, Failed 1, Errors 0, Passed 9
[error] Failed tests:
[error]         org.http4s.DecodeSpec
[error] (tests / Test / testOnly) sbt.TestsFailedException: Tests unsuccessful
[error] Total time: 8 s, completed Mar 21, 2022 2:31:32 PM
```

I actually found that `val` caused a NPE, but `override def` reproduces the error:
```diff
-  override val scalaCheckInitialSeed = "ul4X_OcjscjNgHDzfcislTYeeRBho4GD7b-pOuxHoPJ=";
+  override def scalaCheckInitialSeed = "ul4X_OcjscjNgHDzfcislTYeeRBho4GD7b-pOuxHoPJ=";
```

server/src/main/scala/org/http4s/server/middleware/authentication/NonceF.scala
rossabaker: Cats-Effect 2's `Blocker` does require a thread pool.  One `Blocker` should suffice for most apps, but any app that does blocking needs one.  It's part of the runtime in Cats-Effect 3, and available from the `Sync` constraint, so this will melt away in 0.23.

Engineering tradeoff: if we don't thread it through, this could suck real bad for a small number of users.  If we do thread it through, we're going to all that work for, like, Windows users, and other people who do Weird Shit.  http4s-0.22 is basically EOL.  Personally, I'm only fixing bugs on it.  And this blocking is not new, and nobody has complained.  So if you'd prefer, we could put a fat disclaimer on it and know it will be better in CE3.  But this is "safer".

rossabaker:
Yeah, that test flake looks like another https://github.com/http4s/http4s/issues/4262.  Don't worry about that here.

server/src/main/scala/org/http4s/server/middleware/authentication/NonceF.scala
rossabaker: Oh, I overlooked the `Blocker.use`.  Yeah, that creates a new cached thread pool on every invocation.  What we want is one cached blocker for the entire app that gets threaded through.  This is where the API is going to get uglier in CE2.

blast-hardcheese: Swapped it out to a single allocated `Blocker` on init, backed by the `global` threadpool, as this is going away anyway.

blast-hardcheese: Alright -- threaded it through, as this is a new integration, and the exposure of it will better match the functionality in CE3 for users through the upgrade process.

rossabaker (approved):
Thanks for your perseverance on this.  I think it's in a good place.

rossabaker:
@armanbilge, are you good with this as is?  You had opinions earlier.

server/src/main/scala/org/http4s/server/middleware/authentication/Nonce.scala
```diff
@@ -0,0 +1,32 @@
...
+
+private[authentication] object Nonce {
+  val random = new SecureRandom()
```
armanbilge, line 27: When this becomes an effect, it can no longer be a global. So where will it be initialized?

armanbilge (commented):
Yeah, the scope of this PR got away from me 😆 really fantastic work by @blast-hardcheese. Both CE2 and the RFC are a bit out of my comfort zone.

I glanced through and have one question, but not a blocker.

server/src/main/scala/org/http4s/server/middleware/authentication/Nonce.scala
rossabaker: The usage is deprecated.  I don't think we need to make it an effect.  Though we could maybe add a deprecation notice to this type, to assure it's only called through deprecated code.

armanbilge: Oh, is this not going to use the stuff from https://github.com/http4s/http4s/pull/6165?

rossabaker: I was planning on welding it together after this.  @blast-hardcheese has suffered enough.  But I could merge it into this just to see what it looks like.

armanbilge: Cool, sounds good to me 🚀

armanbilge (approved):

PR #6139, opened 2022-03-18
Fix `Client#translate` and relax constraints

armanbilge:
Alternative to https://github.com/http4s/http4s/pull/6137 that also manages to relax the constraints on `Client#translate`.


client/shared/src/main/scala/org/http4s/client/Client.scala
```diff
@@ -176,19 +178,66 @@ trait Client[F[_]] {
...
+    implicit val G: MonadCancelThrow[G] = liftMonadCancel(F)(fk)(gK)
+    Client[G]((req: Request[G]) =>
       run(
         req.mapK(gK)
       ).mapK(fk)
         .map(_.mapK(fk))
     )
+  }
+
+  private def liftMonadCancel[G[_]](
```
danicheg, line 204: It seems that the `MonadCancel` can have the `mapK` method or even this `lift` method. Did that discuss on discord or somewhere? I will be grateful for the link 🙏🏻 
Anyway, I feel that this `lift` method is super-specific to have it within the `client` package. Maybe there is a more appropriate `fp.utils` package?

armanbilge: Yes, `MonadCancel` can have an `imapK` method :)

Yeah, bit of long discussion 😅 
I propose my idea for this PR here:
https://discord.com/channels/632277896739946517/839254871646404638/954226420601716736
Then we discuss it here:
https://discord.com/channels/632277896739946517/839254871646404638/954238156767625266

I do agree this is weird one for `Client` 😆 @ChristopherDavenport is very suspicious about it though, which suggests we should keep it private-scoped to a small area. I can ask if CE will accept `imapK` addition.

rossabaker: Can anyone summarize the objection?  Neither PR has a unit test: it would be great to see evidence that they satisfy the aim.  Or if this approach doesn't work, a test that shows how it fails.

This one better fits the [golden rule of software quality](https://www.haskellforall.com/2020/07/the-golden-rule-of-software-quality.html) if it indeed works.

armanbilge: @ChristopherDavenport on Discord:
> So I agree that will work, but I have been told and so practice that it is bad to override a typeclass like that.
> As canonically a typeclass should have a single implementation for a type.

https://discord.com/channels/632277896739946517/839254871646404638/954238688676704328

He said he will throw up a unit test today.

ChristopherDavenport: Tests are push to my branch, feel free to steal them over here.

armanbilge: My implementation passes @ChristopherDavenport's tests.

rossabaker: On the theoretical question, is it any worse than:

* [Parallel.applicativeError](https://github.com/typelevel/cats/blob/219fcfbc27a354ef2ace11b5a86e2f2590bded79/core/src/main/scala/cats/Parallel.scala#L81)?
* [Applicative.monoid](https://github.com/typelevel/cats/blob/219fcfbc27a354ef2ace11b5a86e2f2590bded79/core/src/main/scala/cats/Applicative.scala#L198-L199)?

ChristopherDavenport: I believe private[this] is more private than private. As mentioned before I'm not theory person and this version is much cleaner than mine.

armanbilge:
```suggestion
  private[this] def liftMonadCancel[G[_]](
```

rossabaker: `private` is also accessible from other instances of `Client`. `private[this]` is accessible only from the same instance of `Client`.  Furthermore, `private[this]` is direct field access instead of through an accessor method.  I think that change is conservative, and therefore good, until everyone is comfortable with it.

On theory, there's a nice [Haskell discussion](https://www.haskellforall.com/2022/03/applicatives-should-usually-implement.html) of `Applicative.monoid` by Gabriella Gonzalez.

I think there's nothing wrong with providing explicit derived instances like this, and let users promote them to implicit if they are judged to be the canonical instance for the type, or tuck them away as done here when they do the right thing.

client/shared/src/main/scala/org/http4s/client/Client.scala
```diff
@@ -170,36 +172,83 @@ trait Client[F[_]] {
...
+
   @deprecated("use public method with MonadCancelThrow instead", since = "0.23.7")
   private[client] def translate[G[_]: Async](
       fk: F ~> G
   )(gK: G ~> F)(implicit F: MonadCancelThrow[F]): Client[G] = translateImpl(fk)(gK)
 
   /** Translates the effect type of this client from F to G
     */
-  def translate[G[_]: MonadCancelThrow](
+  @deprecated("use public method without MonadCancelThrow[G] constraint instead", since = "0.23.11")
```
rossabaker, line 192:
```suggestion
  @deprecated("use public method without MonadCancelThrow[G] constraint instead", since = "0.23.12")
```

rossabaker: It missed the bus

rossabaker (approved):
This looks good to me.  It does what it says on the tin, and the controversial bits are private in case the theoretical objection manifests in a practical way.

ChristopherDavenport (approved):

Issue #2794, opened 2019-08-14
Flaky test: TraversalSpec


rossabaker:
Probably a URI bug.

```
[info] TraversalSpec
[info] Request Encoder/Parser should
[info]   * preserve headers (396 ms) Pending until fixed
[info]   + preserve method with known uri (47 seconds, 955 ms)
[info]   * preserve uri.scheme (589 ms) Pending until fixed
[error]   x preserve body with a known uri (1 minute 2 seconds, 621 ms)
[error]    Falsified after 98 passed tests.
[...]

PR #6141, opened 2022-03-18
Merge 0.23 -> main

armanbilge:


armanbilge:
Crap, I botched this bad.

rossabaker:
Anything particularly interesting in this one for review?

armanbilge:
No, this one was fine. Let me add the unidoc fix then should be good.

blaze-core/src/main/scala-2.13/org/http4s/blazecore/util/ParasiticExecutionContextCompat.scala
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+package org.http4s.blazecore.util
+
+import org.http4s.blaze.util.Execution
+
+import scala.concurrent.ExecutionContext
+
+private[util] trait ParasiticExecutionContextCompat {
+  final def parasitic: ExecutionContext = Execution.trampoline
```
armanbilge, line 24: Huh, shouldn't this be `ExecutionContext.parasitic` on Scala 2.13?

core/shared/src/main/scala/org/http4s/internal/Trampoline.scala
```diff
@@ -21,6 +21,7 @@ import scala.annotation.tailrec
 import scala.concurrent.ExecutionContext
 import scala.concurrent.ExecutionContextExecutor
 
+@deprecated("Unused.  Will be removed in 1.0", "0.23.11")
```
danicheg: That was supposed to be removed. So, I did it in 6cfa1e4.

build.sbt
```diff
@@ -335,6 +339,37 @@ lazy val blazeCore = libraryProject("blaze-core")
     libraryDependencies ++= Seq(
       blazeHttp
     ),
+    mimaBinaryIssueFilters := {
+      if (tlIsScala3.value)
+        Seq(
+          ProblemFilters
```
danicheg, line 345: Could we remove these MiMa filters?

armanbilge: Ah yes, all MiMa filters can go 👍

danicheg: I almost committed this too. But thanks, @armanbilge.

async-http-client/src/main/scala/org/http4s/asynchttpclient/AsyncHttpClient.scala
```diff
@@ -43,6 +43,7 @@ import org.reactivestreams.Publisher
 
 import scala.jdk.CollectionConverters._
 
+@deprecated("Upstream is unmaintained. Recommend choosing another backend.", "0.22.12")
```
danicheg, line 46: Also, a good question is will we support `AsyncHttpClient` within `1.0`? I tend to say yes gingerly. But let's then fix the deprecated-since version.
```suggestion
@deprecated("Upstream is unmaintained. Recommend choosing another backend.", "1.0")
```

async-http-client/src/main/scala/org/http4s/asynchttpclient/AsyncHttpClientStats.scala
```diff
@@ -25,6 +25,7 @@ import org.asynchttpclient.{Response => _}
 
 import scala.jdk.CollectionConverters._
 
+@deprecated("Upstream is unmaintained. Recommend choosing another backend.", "0.22.12")
```
danicheg, line 28:
```suggestion
@deprecated("Upstream is unmaintained. Recommend choosing another backend.", "1.0")
```

async-http-client/src/test/scala/org/http4s/asynchttpclient/AsyncHttpClientSuite.scala
```diff
@@ -27,6 +27,7 @@ import org.http4s.client.ClientRouteTestBattery
 import org.http4s.client.DefaultClient
 import org.http4s.client.defaults
 
+@deprecated("Upstream client is deprecated", "0.23.11")
```
danicheg, line 30:
```suggestion
@deprecated("Upstream client is deprecated", "1.0")
```

async-http-client/src/main/scala/org/http4s/asynchttpclient/AsyncHttpClient.scala
armanbilge: I'm an extremist, but honestly I say let's toss it. I love any excuse to slim down the build.

blaze-core/src/main/scala-2.13/org/http4s/blazecore/util/ParasiticExecutionContextCompat.scala
danicheg: Hmm, seems like I saw the `Execution.parasitic` indeed for 2.13 in Ross's PR.

danicheg: Uh-oh, this was for the Scala 3 https://github.com/http4s/http4s/pull/6118/files#diff-132f6ad9670a22754e1983195601c792e110429f8b1e70f1b3f992fbed456409R22. Probably, it worth fixing for downstream too.

armanbilge: @danicheg yes. While you fix that, the async deprecation versions need to be fixed too 😅

async-http-client/src/main/scala/org/http4s/asynchttpclient/AsyncHttpClient.scala
armanbilge: Btw, I think we should say deprecated since 0.23.11. It was not deprecated in 1.0, it was deprecated _before_ 1.0 :)

site/src/main/mdoc/landing.template.html
```diff
@@ -26,7 +26,7 @@
   <link rel="stylesheet" type="text/css" href="helium/laika-helium.css" />
   <link rel="stylesheet" type="text/css" href="helium/landing.page.css" />
   <link rel="stylesheet" type="text/css" href="styles/landing.page.css" />
-  <script>initVersions("", "", "");</script>
+  <script>initVersions("", "", " ");</script>
```
armanbilge, line 29: Don't even ask ...

async-http-client/src/main/scala/org/http4s/asynchttpclient/AsyncHttpClient.scala
danicheg: Very likely that download stats (of artifact) could help in the decision-making of removing. However, this removing is a real act of brave heart, you know :)

site/src/main/mdoc/landing.template.html
armanbilge: ### Before
Screen Shot 2022-03-18 at 9 54 11

### After
Screen Shot 2022-03-18 at 9 54 37

.github/workflows/ci.yml
```diff
@@ -426,8 +457,13 @@ jobs:
...
-      - name: Scalafix tests
-        if: matrix.scala == '2.13.8'
-        run: |
-          cd scalafix
-          sbt ci
+      - name: Generate site
+        run: sbt '++${{ matrix.scala }}' site/tlSite
+
+      - name: Publish site
+        if: github.event_name != 'pull_request' && github.ref == 'refs/heads/series/0.23'
```
danicheg, line 464: I wonder, wouldn't this break the publishing of the landing page? Or maybe I misapprehend how we publish it.

armanbilge: Crappp this needs to be fixed :) should be publishing if main.

Anyway, the landing page is only published from main branch. The lower branches never publish versioned material.

project/Http4sSitePlugin.scala
```diff
@@ -0,0 +1,205 @@
...
+
+  }
+
+  object versions {
+    def config(isCi: Boolean): Versions = Versions(
+      currentVersion = current,
+      olderVersions = all.dropWhile(_ != current).drop(1),
+      newerVersions = all.takeWhile(_ != current),
+      // helpful to render unversioned pages when previewing locally
+      renderUnversioned = current == v1_0 || !isCi,
```
armanbilge, line 114: @danicheg this `renderUnversioned` setting disables the landing page from being generated by the 0.22/0.23 branches.

blaze-core/src/main/scala-2.13/org/http4s/blazecore/util/ParasiticExecutionContextCompat.scala
danicheg: Let's wait a bit for @rossabaker opinion on the topic.

project/Http4sSitePlugin.scala
danicheg: Got it, thanks ❤️

site/src/main/mdoc/landing.template.html
danicheg: Sorry for being late on this, but why did we choose the `1` over `1.0`? Is the point that we will have pages for major versions only (`1`, `2` and not `1.1`, `1.2`, etc)?

armanbilge: No worries, I'd like to hear your opinion on this too :) But yes, exactly. Discussion was in https://github.com/http4s/http4s/pull/6087#issuecomment-1060749888.

armanbilge:
Failure is https://github.com/http4s/http4s/issues/2794.

blaze-core/src/main/scala-2.13/org/http4s/blazecore/util/ParasiticExecutionContextCompat.scala
rossabaker: Fuck.  Well, they behave about the same, but let's get it fixed as part of the next train.

danicheg: Done #6145

async-http-client/src/main/scala/org/http4s/asynchttpclient/AsyncHttpClient.scala
rossabaker: It's would move to a new repo in the Great Schism anyway.

It's 9.4% of our client downloads in February.  On 1.0.0-M31, it's 4.0%.  Let's leave it in this milestone just to announce its demise to users who aren't watching every merge PR, and cut it in the next.

```scala
@deprecated("Upstream is unmaintained. This backend will be removed in the next milestone. If anyone wants to adopt it, please contact the http4s team.", "1.0.0-M32")
```

rossabaker (approved):

PR #6140, opened 2022-03-18
Release v0.23.11

rossabaker:
Need to land the plane sometime.  I expect another release train next week.

PR #6142, opened 2022-03-18
Fix the unidoc artifact

armanbilge:
Fixes the problem mentioned in https://github.com/http4s/http4s/pull/6140#issuecomment-1072501417.

We could have avoided this if we fixed:
- https://github.com/typelevel/sbt-typelevel/issues/88
- https://github.com/typelevel/sbt-typelevel/issues/214


rossabaker (approved):
Approving because it looks superficially reasonable and you understand it and there are still plenty of things out there nobody understands.

PR #6145, opened 2022-03-18
Use parasitic EC in the blaze for Scala 2.13

danicheg:
This is a follow-up from https://github.com/http4s/http4s/pull/6141#pullrequestreview-914574674. I think this is legit to use parasitic EC from std for Scala 2.13.


rossabaker (approved):
Yes, that was the intent.  I whiffed on this one.

Issue #6150, opened 2022-03-18
Releases from v0.21 completely destroy the site


armanbilge:
I didn't see this one coming 😓 https://github.com/http4s/http4s/commit/1e67876ff4113d7c143ba6c8d174c9f3139eaef7

armanbilge:
Manually reverted.

rossabaker:
I had a hunch.  Thanks for checking.

rossabaker:
Fixed by #6151.

PR #6151, opened 2022-03-18
Don't publish website from 0.21

armanbilge:
Fixes https://github.com/http4s/http4s/issues/6150.

When it comes to the http4s site, you just can't win 😩


rossabaker (approved):
Too many incompatible methods that worked in their day.

PR #6152, opened 2022-03-18
Remove deprecated `CSRF.SHA1ByteLen`

danicheg:
That's a too old thing, so let's clean up.


armanbilge:
Mild 👎 no matter how old or deprecated, it's not nice to break bincompat on a stable branch IMO. I guess a source-breaking package-private could be ok.

danicheg:
@armanbilge I tend to agree with your objection in the general case, but I believe that `val` is totally unused. It doesn't hold any functionality. Consequently, I really don't see evil in this clean-up. But the decision of accepting PR is up to you 👍🏻

rossabaker:
We probably should have cleaned this one up before 0.22 or 0.23, but I'd leave it for main.

danicheg:
Ok then, I'll reopen this with targeting to main.

PR #6159, opened 2022-03-20
Ember-Core micro-optimisation: avoid middle lists.

diesalbla:
Introduce an auxiliary function `mapRefForEach` to run an action `F[Unit]` on all the items in the `mapRef` map of streams. It uses an iterator and loop rather than throw-away lists.

### DIFF
--- a/ember-core/shared/src/main/scala/org/http4s/ember/core/h2/H2Connection.scala
+++ b/ember-core/shared/src/main/scala/org/http4s/ember/core/h2/H2Connection.scala
@@ -23,6 +23,7 @@ import fs2._
 import fs2.io.net.Socket
 import fs2.io.net.SocketException
 import org.typelevel.log4cats.Logger
+import scala.annotation.tailrec
 import scodec.bits._
 
 private[h2] class H2Connection[F[_]](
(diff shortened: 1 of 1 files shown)


ChristopherDavenport (commented):
This one is an unpleasant indirection for me. I don't really understand why I want to figure out this function in the middle of understanding the flow here.

ember-core/shared/src/main/scala/org/http4s/ember/core/h2/H2Connection.scala
```diff
@@ -140,14 +141,21 @@ private[h2] class H2Connection[F[_]](
     } >>
       H2Connection.KillWithoutMessage().raiseError
 
+  private[this] def mapRefForEach(f: H2Stream[F] => F[Unit]): F[Unit] = {
```
ChristopherDavenport, line 144: Isn't this just `traverse_`

ChristopherDavenport: m.valuesIterator traverse_ seems like we could do somehow. That would be simpler then methods with iterator.

diesalbla: Ah, well... the problem is that there is no `traverse_` for `Iterator`, there is only a `Foldable` for `Iterable`, and then implementation of `traverse_` of `cats` also incurs memory cost. 

```Scala
  def traverse_[G[_], A, B](fa: F[A])(f: A => G[B])(implicit G: Applicative[G]): G[Unit] =
    foldRight(fa, Always(G.pure(()))) { (a, acc) =>
      G.map2Eval(f(a), acc) { (_, _) =>
        ()
      }
    }.value
```

ember-core/shared/src/main/scala/org/http4s/ember/core/h2/H2Connection.scala
```diff
@@ -140,14 +141,21 @@ private[h2] class H2Connection[F[_]](
...
+    @tailrec
+    def go(iter: Iterator[H2Stream[F]], acc: F[Unit]): F[Unit] =
+      if (iter.hasNext) go(iter, acc *> f(iter.next())) else acc
```
ChristopherDavenport, line 147: Isn't iter.hasNext an effect?

diesalbla: An iterator is a mutable structure, and each `next()` pops the next element. Since `*>` is strict, this would force the evaluation of `f(iter.next())`, which would itself force the evaluation of `iter.next()`, all before doing the recursive call to `go`. Mutability is contained within the method, and different calls to the method should create different iterators.

PR #6160, opened 2022-03-20
Remove unused `EntityBodyWriter.wroteHeader`

danicheg:
Zero usages found. So, let's burn it?


rossabaker (approved):
I thought I remembered this getting removed somewhere else, but if not, :+1:.

PR #6163, opened 2022-03-20
Ember Core: write readLoop without Streams.

diesalbla:
The `readLoop` in the `ember-core` module made a use of fs2 Streams that was not necessary for the use-case at hand. In this use case: 

- There is no emitted outputs that the consumer may use to cut the stream processing.
- There were no interruptions that connected the flow of the stream to an outside event. 
- There were no resources acquired during the stream that had a limited nested scope.
- There was no chunking or re-chunking of outputs or intermediate values. The way the stream was constructed, it was just generating one output (in a singleton chunk) and immediately processing it, one by one. 

In this circumstances, fs2 Streams do not add any functionality, and takes some cognitive toll and a bit of memory cost. Instead, writing it as a tail-recursive `F[Unit]` operation fits the problem better.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -509,6 +509,8 @@ lazy val emberCore = libraryCrossProject("ember-core", CrossType.Full)
     ),
     mimaBinaryIssueFilters ++= Seq(
       ProblemFilters
+        .exclude[IncompatibleResultTypeProblem]("org.http4s.ember.core.h2.H2Connection.readLoop"),
+      ProblemFilters
         .exclude[DirectMissingMethodProblem]("org.http4s.ember.core.Encoder.reqToBytes"),
       ProblemFilters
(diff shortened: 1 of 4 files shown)


ChristopherDavenport (approved):

build.sbt
```diff
@@ -508,6 +508,8 @@ lazy val emberCore = libraryCrossProject("ember-core", CrossType.Full)
       log4catsTesting.value % Test,
     ),
     mimaBinaryIssueFilters ++= Seq(
+      ProblemFilters
+        .exclude[IncompatibleResultTypeProblem]("org.http4s.ember.core.h2.H2Connection.readLoop"),
```
armanbilge, line 512: Probably we don't need this. Let's see.
```suggestion
```

PR #6164, opened 2022-03-20
Relax ContextRouter's constraints

### LINKED ISSUE #6105: ContextRouter has the wrong constraints in 0.22, 0.23 and main
It requires Sync, but only needs Monad.

danicheg:
Closes #6105. Fixing this for `0.22`/`0.23` would mean breaking MiMa. I'm not confident it's worth the hassle. The fix for `1.0` is pretty straightforward and clear.

### DIFF
--- a/server/shared/src/main/scala/org/http4s/server/ContextRouter.scala
+++ b/server/shared/src/main/scala/org/http4s/server/ContextRouter.scala
@@ -84,6 +84,13 @@ object ContextRouter {
   def apply[F[_]: Monad, A](mappings: (String, ContextRoutes[A, F])*): ContextRoutes[A, F] =
     define(mappings: _*)(ContextRoutes.empty[A, F])
 
+  @deprecated("Kept for binary compatiblity.  Use the apply with the Monad constraint.", "0.23.13")
+  def apply[F[_], A](
+      mappings: Seq[(String, ContextRoutes[A, F])],
+      sync: Sync[F],
+  ): ContextRoutes[A, F] =
+    defineHelper(mappings, ContextRoutes.empty[A, F](sync))(sync)
+
   /** Defines an [[ContextRoutes]] based on list of mappings and
     * a default Service to be used when none in the list match incoming requests.
     *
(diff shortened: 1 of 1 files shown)


armanbilge:
It can be done on the lower branches, without breaking MiMa. The trick is to add a new public method with a matching binary signature:
```scala
def apply[F[_], A](mappings: Seq[(String, ContextRoutes[A, F])], sync: Sync[F])
//                           ^^^ need to double-check what the collection used for varargs is
```

danicheg:
@armanbilge Looks not so pretty, to be honest :( Are we really want this at all costs?

armanbilge:
Sure, bincompat never is! The http4s codebase is full of not-so-pretties 😅 

I think `Sync` -> `Monad` is a pretty big change 😆 but, I didn't open the issue :)

armanbilge:
Not if we keep it public. We only lose static forwarders if we do the package-private thing, in which case my bizarre rewrite isn't necessary. The idea is to add something source-breaking but not binary-breaking, so the new method is used instead.

rossabaker (approved):
I'm ambivalent to backporting, but this definitely seems good for main.

danicheg:
Would be good to get an opinion about that of the reporter @hamnis.

hamnis (approved):
I can live with the Sync constraint in 0.22 / 0.23. 

Its just annoying that I have to have Async constrains deep into the callstack because of this constraint.

rossabaker:
We've added static forwarder exceptions for similar before.  I'm fine merging this one to take the victory, but if @hamnis needs it, we can do another PR with @armanbilge's suggestion for 0.23.

armanbilge:
I'm a broken record, but we won't need any static forwarder exception! :)

danicheg:
I lost the train of what we want from this. Is it ok to get this merged? Or does someone want to try another approach?

armanbilge:
I'm confident that we can fix this on the lower branches without breaking MiMa at all (no exceptions). But seems nobody is interested to do the work, which is fine. Let's merge and we can leave the issue open if anybody wants to do it.

rossabaker:
I think I tried that and concluded it wasn't possible, but I found a way based on @armanbilge's idea.

armanbilge:
Yay! Re-triggering CI since the branch change confused it.

server/shared/src/main/scala/org/http4s/server/ContextRouter.scala
```diff
@@ -81,17 +81,38 @@ object ContextRouter {
   /** Defines an [[ContextRoutes]] based on list of mappings.
     * @see define
     */
-  def apply[F[_]: Sync, A](mappings: (String, ContextRoutes[A, F])*): ContextRoutes[A, F] =
+  def apply[F[_]: Monad, A](mappings: (String, ContextRoutes[A, F])*): ContextRoutes[A, F] =
     define(mappings: _*)(ContextRoutes.empty[A, F])
 
+  @deprecated("Kept for binary compatiblity.  Use the apply with the Monad constraint.", "0.23.13")
```
armanbilge, line 87: Right? So hard to keep track 😆 

```suggestion
  @deprecated("Kept for binary compatiblity.  Use the apply with the Monad constraint.", "0.23.12")
```

server/shared/src/main/scala/org/http4s/server/ContextRouter.scala
```diff
@@ -81,17 +81,38 @@ object ContextRouter {
...
     *
     * The mappings are processed in descending order (longest first) of prefix length.
     */
-  def define[F[_]: Sync, A](
+  def define[F[_]: Monad, A](
       mappings: (String, ContextRoutes[A, F])*
   )(default: ContextRoutes[A, F]): ContextRoutes[A, F] =
+    defineHelper(mappings, default)
+
+  @deprecated("Kept for binary compatiblity.  Use the define with the Monad constraint.", "0.23.13")
```
armanbilge, line 104:
```suggestion
  @deprecated("Kept for binary compatiblity.  Use the define with the Monad constraint.", "0.23.12")
```

armanbilge (approved):

PR #6165, opened 2022-03-20
Add Random shim

rossabaker:
This is neither a superset nor subset of the Cats-Effect 3 random, but it shows how to detect a blocking instance, and supports #6138.

### DIFF
--- a/core/src/main/scala/org/http4s/internal/Random.scala
+++ b/core/src/main/scala/org/http4s/internal/Random.scala
@@ -5,11 +5,15 @@ import cats.effect.ContextShift
 import cats.effect.Sync
 import cats.syntax.all._
 
+import java.math.{BigInteger => JBigInteger}
 import java.security.SecureRandom
 import java.util.{Random => JRandom}
 
 /** A partial backport of Cats-Effect 3's `Random`. */
 private[http4s] trait Random[F[_]] {
+
+  /** Generates a random BigInt between 0 and (2^bits-1) inclusive. */
+  def nextBigInt(bits: Int): F[BigInt]
 
   /** Generates `n` random bytes into a new array */
   def nextBytes(n: Int): F[Array[Byte]]
@@ -28,6 +32,9 @@ object Random {
     */
   def javaUtilRandomNonBlocking[F[_]](random: JRandom)(implicit F: Sync[F]): Random[F] =
     new Random[F] {
+      def nextBigInt(bits: Int): F[BigInt] =
+        F.delay(BigInt(new JBigInteger(bits, random)))
+
       def nextBytes(n: Int): F[Array[Byte]] =
         F.delay {
           val arr = new Array[Byte](n)
@@ -46,6 +53,9 @@ object Random {
       blocker: Blocker
   )(random: JRandom)(implicit F: Sync[F], cs: ContextShift[F]): Random[F] =
     new Random[F] {
+      def nextBigInt(bits: Int): F[BigInt] =
+        blocker.blockOn(F.delay(BigInt(new JBigInteger(bits, random))))
+
       def nextBytes(n: Int): F[Array[Byte]] =
         F.delay(new Array[Byte](n)).flatMap { arr =>
           blocker.blockOn(
--- a/core/src/main/scala/org/http4s/internal/Random.scala
+++ b/core/src/main/scala/org/http4s/internal/Random.scala
@@ -1,0 +1,81 @@
+package org.http4s.internal
+
+import cats.effect.Blocker
+import cats.effect.ContextShift
+import cats.effect.Sync
+import cats.syntax.all._
+
+import java.security.SecureRandom
+import java.util.{Random => JRandom}
+
+/** A partial backport of Cats-Effect 3's `Random`. */
+private[http4s] trait Random[F[_]] {
+
+  /** Generates `n` random bytes into a new array */
+  def nextBytes(n: Int): F[Array[Byte]]
+}
+
+object Random {
+
+  /** Create a non-blocking Random instance.  This is more efficient
+    * than [[javaSecurityRandom]], but requires a more thoughtful
+    * instance of random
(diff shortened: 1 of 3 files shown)


tests/src/test/scala/org/http4s/internal/RandomSuite.scala
```diff
@@ -0,0 +1,68 @@
...
+import java.util.{Random => JRandom}
+import scala.util.Try
+
+class RandomSuite extends Http4sSuite {
+  test("javaUtilRandomBlocking.nextBigInt works") {
+    val rnd = new JRandom()
+    Random
+      .javaUtilRandomBlocking[IO](testBlocker)(rnd)
+      .nextBigInt(10)
+      .map(_ > 0)
```
armanbilge, line 16: Why `_ > 0`?

https://docs.oracle.com/javase/8/docs/api/java/math/BigInteger.html#BigInteger-int-java.util.Random-

core/src/main/scala/org/http4s/internal/package.scala
```diff
@@ -346,4 +348,12 @@ package object internal {
...
     }
   }
+
+  private[http4s] def javaMajorVersion[F[_]](implicit F: Sync[F]): F[Option[Int]] =
+    F.delay(sys.props.get("java.version")).map {
+      case Some(v) =>
+        Try(v.split("\\.").head.toInt).toOption
+      case None =>
+        None
+    }
```
armanbilge, line 358: I think on Java 8 `java.version` is formatted like `1.8...`.

tests/src/test/scala/org/http4s/internal/RandomSuite.scala
rossabaker: Because my son's hockey practice was almost over and I forgot what I was doing.  The tests could be property tests.

core/src/main/scala/org/http4s/internal/package.scala
rossabaker: Yes, that's true.  It behaves correctly for the versions I cared about, but it's not yet right in general.

tests/src/test/scala/org/http4s/internal/RandomSuite.scala
rossabaker: It's now a property test.

core/src/main/scala/org/http4s/internal/Random.scala
```diff
@@ -67,21 +68,43 @@ private[http4s] object Random {
     * non-blocking.  If a non-blocking instance can't be guaranteed,
     * falls back to a blocking implementation.
     */
-  def javaSecuritySecureRandom[F[_]](blocker: Blocker)(implicit F: Sync[F], cs: ContextShift[F]) =
+  def javaSecuritySecureRandom[F[_]](blocker: Blocker)(implicit F: Sync[F], cs: ContextShift[F]) = {
+    // This is a known, non-blocking, threadsafe algorithm
+    def happyRandom = F.delay(SecureRandom.getInstance("NativePRNGNonBlocking"))
```
armanbilge, line 73: So in what situations will this algorithm not be available? Are any of the other implementations applicable? E.g. `NativePRNG` seems non-blocking after it has been seeded.
```
SHA1PRNG (Initial seeding is currently done via a combination of system attributes and the java.security entropy gathering device)
NativePRNG (nextBytes() uses /dev/urandom, generateSeed() uses /dev/random)
NativePRNGBlocking (nextBytes() and generateSeed() use /dev/random)
NativePRNGNonBlocking (nextBytes() and generateSeed() use /dev/urandom)
```
https://docs.oracle.com/javase/8/docs/technotes/guides/security/SunProviders.html#SUNProvider

rossabaker: I don't think it will be available on Windows.  The NativePRNG on Windows

core/src/main/scala/org/http4s/internal/Random.scala
```diff
@@ -0,0 +1,110 @@
...
+        fallbackPool
+    }
+  }
+
+  def pool[F[_]](n: Int)(random: F[Random[F]])(implicit F: Sync[F]): F[Random[F]] =
+    for {
+      ref <- Ref[F].of(0)
+      array <- Vector.fill(n)(random).sequence
+    } yield {
+      val incrGet = ref.modify(i => (if (i < (n - 1)) i + 1 else 0, i))
```
armanbilge, line 103: This is just `(i + 1) % n`, right? 😅

core/src/main/scala/org/http4s/internal/Random.scala
rossabaker: I don't think `NativePRNGNonBlocking` will be available on Windows, whose `NativePRNG` is a [dummy implementation](https://github.com/openjdk/jdk/blob/master/src/java.base/windows/classes/sun/security/provider/NativePRNG.java).  It also won't be available if `/dev/urandom` is unreadable, which would be highly weird.

`NativePRNG` should be available in mostly the same circumstances as `NativePRNGNonBlocking`.

I can't find any guarantees about `SHA1PRNG`.  Sounds like the best case is that we'd have to block on seeding.

core/src/main/scala/org/http4s/internal/Random.scala
rossabaker: Uh, probably.  I stole it from CE3 Random and neglected to attribute it.

armanbilge (dismissed):
Looks good 👍 I hope we can land something similar in CE 3.4 before it locks down on source-breaking changes for another 6 months 😆

rossabaker:
Alas and alack, the method http4s needs is not on the CE3 Random interface, so we're going to need to figure out how to make our own BigInt on merge.  I think we can do it from `nextBytes`.

danicheg (approved):

PR #6166, opened 2022-03-21
Fix comment in org.http4s.Message.scala

takapi327:
@ChristopherDavenport thanks!


rossabaker (approved):
I'm going to pull this back onto an earlier branch, but looks good.  Thanks!

PR #6169, opened 2022-03-22
Improvements to multipart boundaries

rossabaker:
Not all servers gracefully handle multipart quoted boundaries.  We remove all non-token characters.  This new alphabet is consistent with the generators in both Apache HTTP and Spring.  It's also a built-in Base64 decoder, so we can use the platform's optimized encoder from an array of bytes vs. our (ugh) random access into a List!

`Boundary.create` is impure and deprecated.

Introduces a `Multiparts` algebra to house all this.

### DIFF
--- a/core/src/main/scala/org/http4s/multipart/Boundary.scala
+++ b/core/src/main/scala/org/http4s/multipart/Boundary.scala
@@ -32,17 +32,23 @@ object Boundary {
   private val BoundaryLength = 41
   val CRLF = "\r\n"
 
-  private val DIGIT = ('0' to '9').toList
-  private val ALPHA = ('a' to 'z').toList ++ ('A' to 'Z').toList
-  // Many more chars are allowed (see bchars definition in
-  // https://www.rfc-editor.org/rfc/rfc2046), but these are known to
-  // be robust in implementation.
-  private val OTHER = "-_".toList
-  private val CHARS = DIGIT ++ ALPHA ++ OTHER
-  private val nchars = CHARS.length
+  private val alphabet = {
+    val arr = Array.newBuilder[Char]
+    arr ++= ('0' to '9')
+    arr ++= ('a' to 'z')
+    arr ++= ('A' to 'Z')
+    // Many more chars are allowed (see bchars definition in
+    // https://www.rfc-editor.org/rfc/rfc2046), but these are known to
+    // be robust in implementation.  Specifically, they don't trigger
+    // quotation in the Content-Type parameter.
+    arr += '-'
+    arr += '_'
+    arr.result()
+  }
+  private val nchars = alphabet.length
   private val rand = new Random()
 
-  private def nextChar = CHARS(rand.nextInt(nchars - 1))
+  private def nextChar = alphabet(rand.nextInt(nchars - 1))
   private def stream: CollectionCompat.LazyList[Char] =
     CollectionCompat.LazyList.continually(nextChar)
   private def value(l: Int): String = stream.take(l).mkString
--- a/core/src/main/scala/org/http4s/multipart/Boundary.scala
+++ b/core/src/main/scala/org/http4s/multipart/Boundary.scala
@@ -17,6 +17,7 @@
 package org.http4s.multipart
 
 import cats.Eq
+import cats.effect.Sync
 import fs2.Chunk
 import org.http4s.internal.CollectionCompat
 
@@ -46,14 +47,26 @@ object Boundary {
     arr.result()
   }
   private val nchars = alphabet.length
-  private val rand = new Random()
+  private val defaultRandom = new Random()
 
-  private def nextChar = alphabet(rand.nextInt(nchars - 1))
-  private def stream: CollectionCompat.LazyList[Char] =
-    CollectionCompat.LazyList.continually(nextChar)
-  private def value(l: Int): String = stream.take(l).mkString
+  private def nextChar(random: Random) = alphabet(random.nextInt(nchars - 1))
+  private def stream(random: Random): CollectionCompat.LazyList[Char] =
+    CollectionCompat.LazyList.continually(nextChar(random))
+  private def value(random: Random, l: Int): String =
+    stream(random).take(l).mkString
 
-  def create: Boundary = Boundary(value(BoundaryLength))
+  @deprecated("Impure. Use fromScalaRandom", "0.22.14")
+  def create: Boundary = unsafeCreate()
+
+  /** Create a new MIME boundary. */
+  def fromScalaRandom[F[_]](random: Random)(implicit F: Sync[F]): F[Boundary] =
+    F.delay(unsafeCreateFromScalaRandom(random))
+
+  private[multipart] def unsafeCreate(): Boundary =
+    unsafeCreateFromScalaRandom(defaultRandom)
+
+  private def unsafeCreateFromScalaRandom(random: Random): Boundary =
+    Boundary(value(random, BoundaryLength))
 
   implicit val boundaryEq: Eq[Boundary] = Eq.by(_.value)
 }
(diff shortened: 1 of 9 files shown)


core/src/main/scala/org/http4s/multipart/Boundary.scala
```diff
@@ -29,27 +30,43 @@ final case class Boundary(value: String) extends AnyVal {
...
+  private def stream(random: Random): CollectionCompat.LazyList[Char] =
+    CollectionCompat.LazyList.continually(nextChar(random))
+  private def value(random: Random, l: Int): String =
+    stream(random).take(l).mkString
+
+  @deprecated("Impure. Use fromScalaRandom", "0.22.14")
+  def create: Boundary = unsafeCreate()
+
+  /** Create a new MIME boundary. */
+  def fromScalaRandom[F[_]](random: Random)(implicit F: Sync[F]): F[Boundary] =
```
rossabaker, line 62: I started in 0.22 because this is arguably a bug.  In 0.23, this would better take a Cats-Effect `Random`.

core/src/main/scala/org/http4s/multipart/Multipart.scala
```diff
@@ -19,7 +19,16 @@ package multipart
...
-final case class Multipart[F[_]](parts: Vector[Part[F]], boundary: Boundary = Boundary.create) {
+/** Create a new multipart from a vector of parts and a boundary.
+  *
+  * Caution: the default argument for boundary is impure.  There is no
+  * binary-compatible fix, but the default parameter will be dropped
+  * in 1.0.
+  */
+final case class Multipart[F[_]](
+    parts: Vector[Part[F]],
+    boundary: Boundary = Boundary.unsafeCreate(),
```
rossabaker, line 30: I can't figure out a binary compatible way to deprecate just the default parameter.  The signature is otherwise what we want.

tests/src/test/scala/org/http4s/multipart/MultipartSuite.scala
```diff
@@ -55,29 +57,34 @@ class MultipartSuite extends Http4sSuite {
...
-      val body = entity.body
-      val request =
-        Request(method = Method.POST, uri = url, body = body, headers = multipart.headers)
-      val decoded = EntityDecoder[IO, Multipart[IO]].decode(request, true)
-      val result = decoded.value
-
-      assertIOBoolean(result.map(_ === Right(multipart)))
+      Boundary
+        .fromScalaRandom[IO](random)
+        .flatMap { boundary =>
```
rossabaker, line 62: The purer API is a bit more awkward to call.  I'm open to putting a convenience on `Multipart`

rossabaker: It's a little better if we make a Multiparts algebra.

core/src/main/scala/org/http4s/multipart/Boundary.scala
rossabaker: [Moved to `Multiparts`]

core/src/main/scala/org/http4s/multipart/Multipart.scala
armanbilge: Ooh, I can give this a try.

rossabaker: I tried adding a `apply$default$2()` or whatever it is, and was still in synthetic hell.  Might be able to figure it out with javap, but I gave up.

armanbilge: Btw, if we take away the default parameter it will definitely be source-breaking. Is that ok?

A source-preserving strategy would be to deprecate this `apply` and add a new one with the arguments in reverse order.

rossabaker: That's an intriguing idea, but inconsistent with the ordering of the fields of the case class.  I'm not sure that's a good deal.

core/src/main/scala/org/http4s/multipart/Multipart.scala
```diff
@@ -19,10 +19,26 @@ package multipart
...
       `Transfer-Encoding`(TransferCoding.chunked),
       `Content-Type`(MediaType.multipartType("form-data", Some(boundary.value))),
     )
 }
+
+object Multipart {
+  @deprecated("Retaining for binary-compatibility", "0.22.14")
+  def `<init>$default$2`: String = apply$default$2
+  @deprecated("Retaining for binary-compatibility", "0.22.14")
+  def apply$default$2: String = Boundary.unsafeCreate().value
```
armanbilge, line 43: If we leave them public, then we don't have to fight mima.

danicheg: What if we just create the `apply` method with the default parameter for `boundaries`? That will break the binary compatibility, but not source compatibility. `<init>$default$2`, `apply$default$2` looks pretty bizarre IMO..

rossabaker: Clever, but I don't love that this ends up being a breaking change with no compile time advice how to fix it.  I'd say we could publish a scalafix, but most users aren't running it, and lifting multipart creation into IO is a pretty big change.  It's naughty and impure, but it's probably not an area people are doing a lot of reasoning about.  I think we hang our heads in shame now and fix it for 1.0.

armanbilge: FTR, we can make this package-private. It will remove static forwarders, which we usually deem as "ok".

rossabaker: So the real question is still whether this impurity is so egregious that we should break apps for people who don't read the docs or release notes in a patch version.  I'm still leaning no, but I could be dissuaded.

danicheg: > Clever, but I don't love that this ends up being a breaking change with no compile time advice how to fix it.

My initial intention was to don't break compile time. Probably I'm missing something dully obvious, but providing that `apply` method:
```scala
@deprecated
def apply[F[_]](parts: Vector[Part[F]], boundary: Boundary = Boundary.unsafeCreate()) = new Multipart(parts, boundary)
```
it should work well for goals of preserving source compatibility (stable compilation).

danicheg: But yeah, that will steal the default case class constructor. And if someone used a fully specified constructor (simply put - didn't use a default parameter of the constructor), it will mean some inconvenience. Not so user-friendly indeed.

rossabaker: Arman's technique is binary compatible, and with a deprecated unary apply and deprecated auxiliary constructor, we can make it source compatible.

core/src/main/scala/org/http4s/multipart/Multiparts.scala
```diff
@@ -0,0 +1,59 @@
...
+}
+
+object Multiparts {
+
+  /** Creates a `scala.util.Random` and provides Multiparts around it.
+    * This instance can be shared, or is cheap enough to create closer
+    * to where Multipart values are generated.
+    */
+  def forSync[F[_]](implicit F: Sync[F]): F[Multiparts[F]] =
+    F.delay(new Random()).map(fromScalaRandom[F])
```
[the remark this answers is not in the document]
armanbilge: I see your point from a practical perspective. Theoretically it makes me think:
> If a tree falls in a forest and no one is around to hear it, does it make a sound?

It seems like a slippery slope :)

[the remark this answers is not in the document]
rossabaker: What I've done here is consistent with how CE3's `Random` creates one, but I'm having a hard time justifying it.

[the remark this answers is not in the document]
ChristopherDavenport: Randoms can be created from seeds which are then the same. This is definitely the right way to do this.

armanbilge (approved):
A review by an amateur, but I understand the changes and 👍

The RFC doesn't seem to say anything about the minimum boundary length, so a comment about how/why that was chosen would be helpful.

rossabaker:
I should have taken notes.  There was a viral idea in the Java community to generate randomly between 30 and 40.  I noticed that the safe characters were the modified Base64 alphabet and exploited that.

PR #6170, opened 2022-03-22
Render a trailing newline on multipart close-delimiter

rossabaker:
Adds a `"\r\n"` after the close-delimiter when encoding multipart bodies.  This is not required per [specification](https://datatracker.ietf.org/doc/html/rfc1521#section-7.2), but is more robust with certain server implementations.


danicheg (commented):
As far I can tell, ending CRLF for close-delimiter is fine in the case of the presence of epilogue. And according to the RFC (https://www.w3.org/Protocols/rfc1341/7_2_Multipart.html), it would be a part of the epilogue then. So, since this actually doesn't violate the RFC, I haven't much against it. But If I'm not wrong, we don't encode any epilogue at the moment (or maybe I poorly searched)?

rossabaker:
That's correct.  The epilogue is currently empty.  We don't provide a way to control it, because the recipient is supposed to ignore it.  This just helps it work with bad parsers in the wild.

danicheg (approved):
I get that, thanks for the clarification.

PR #6172, opened 2022-03-22
Add mapK to MetricsOps

hamnis:
Sometimes it is useful to transform the MetricsOps effect.
For instance when using Natchez.


rossabaker (commented):
Does this need a full `translate` like `Client`, or is the `mapK` sufficient?

rossabaker (approved):
:+1: on import cleanup

PR #6175, opened 2022-03-22
Remove http4s-async-http-client

rossabaker:
This stops short of removing it entirely from the build.  We still use it in servlet tests.  I don't want it in production anymore, but it's not going to bitrot, so I'm okay with that as a test dependency.


danicheg (approved):
🔥

PR #6176, opened 2022-03-22
Add withContentLength and toStrict helpers to Message

rossabaker:
Similar to the logic in `BodyCache`, but applies to any Message.  Takes care of setting an appropriate `Content-Length` header and stripping `chunked` from any `Transfer-Encoding` header.

Motivation: there are many servers that can't handle chunked transfer encoding.  This is particularly common in multipart implementations.  Calling `toStrict` defeats streaming, but enables communication with such servers.

### DIFF
--- a/core/src/main/scala/org/http4s/Message.scala
+++ b/core/src/main/scala/org/http4s/Message.scala
@@ -34,10 +34,10 @@ import fs2.Stream
 import fs2.text.utf8Encode
 import org.http4s.headers._
 import org.http4s.internal.CurlConverter
-import org.http4s.syntax.header._
 import org.http4s.syntax.KleisliSyntax
 import org.http4s.syntax.KleisliSyntaxBinCompat0
 import org.http4s.syntax.KleisliSyntaxBinCompat1
+import org.http4s.syntax.header._
 import org.log4s.getLogger
 import org.typelevel.ci.CIString
 import org.typelevel.vault._
--- a/core/src/main/scala/org/http4s/Message.scala
+++ b/core/src/main/scala/org/http4s/Message.scala
@@ -28,11 +28,13 @@ import com.comcast.ip4s.Hostname
 import com.comcast.ip4s.IpAddress
 import com.comcast.ip4s.Port
 import com.comcast.ip4s.SocketAddress
+import fs2.Chunk
 import fs2.Pure
 import fs2.Stream
 import fs2.text.utf8Encode
 import org.http4s.headers._
 import org.http4s.internal.CurlConverter
+import org.http4s.syntax.header._
 import org.http4s.syntax.KleisliSyntax
 import org.http4s.syntax.KleisliSyntaxBinCompat0
 import org.http4s.syntax.KleisliSyntaxBinCompat1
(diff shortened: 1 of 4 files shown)


rossabaker:
This will have interesting interactions with the new Entity model in 1.0.

core/src/main/scala/org/http4s/headers/Transfer-Encoding.scala
```diff
@@ -46,9 +46,23 @@ object `Transfer-Encoding` {
...
+    * return none
+    *
+    * {{{
+    * scala> import org.http4s._
+    * scala> val te = `Transfer-Encoding`(TransferCoding.chunked, TransferCoding.gzip)
+    * scala> te.filter(_ != TransferCoding.chunked)
+    * res0: Option[`Transfer-Encoding`] = Some(Transfer-Encoding(NonEmptyList(TransferCoding(gzip))))
+    * scala> te.filter(_ => false)
+    * res0: Option[`Transfer-Encoding`] = None
+    * }}}
```
rossabaker, line 64: This doctest isn't running yet, but I'd already written it. 🤷  It's covered by the withContentLength test in Message.

core/src/main/scala/org/http4s/headers/Transfer-Encoding.scala
```diff
@@ -46,9 +46,23 @@ object `Transfer-Encoding` {
...
+    */
+  def filter(f: TransferCoding => Boolean): Option[`Transfer-Encoding`] =
```
rossabaker, line 66: Functions like this are plausible for every recurring header, I think?

core/src/main/scala/org/http4s/Message.scala
```diff
@@ -211,6 +213,55 @@ sealed trait Message[F[_]] extends Media[F] { self =>
...
+    * res1: Headers = Headers(Content-Type: text/plain, Content-Length: 1024)
+    * }}}
+    */
+  def withContentLength(contentLength: `Content-Length`): Self =
+    transformHeaders(_.transform { hs =>
+      val b = List.newBuilder[Header.Raw]
+      hs.foreach { h =>
+        h.name match {
+          case `Transfer-Encoding`.name =>
+            `Transfer-Encoding`
```
danicheg, line 248: I feel that we could do here impure but faster things like: 
```scala
if (h.value.contains(","))
  TransferCoding
    .parseList(h.value)
    .fold(
      _ => b += h,
      _.toList.foreach {
        case TransferCoding.chunked =>
          ()
        case tc =>
          b += Header.Raw(`Transfer-Encoding`.name, tc.coding)
      },
    )
else
  TransferCoding
    .parse(h.value)
    .fold(
      _ => b += h,
      {
        case TransferCoding.chunked =>
          ()
        case tc =>
          b += Header.Raw(`Transfer-Encoding`.name, tc.coding)
      },
    )
```
On one hand, this exposes the internals of the `Transfer-Encoding` header. One could fix that by placing that function within `Transfer-Encoding` or  `TransferCoding` and making it private. On another, in the case of separated `Transfer-Encoding` headers, we will have a performance boost. But this could be a kinda premature optimization. I haven't numbers for proof.

danicheg: Also, I wonder - are we really should parse `TransferCoding` values. Or we could rely on existing header values. If so, then we can remove the parsing step in case of a single `TransferCoding` value, I guess.

rossabaker: Yeah, this strikes me as a premature optimization, but maybe we could do a separate PR if it's thought to be worthwhile.

danicheg (approved):

PR #6177, opened 2022-03-24
Integrate Random into DigestAuth

rossabaker:
Integrates `Random` into `DigestAuth` to avoid blocking.

### DIFF
--- a/server/src/main/scala/org/http4s/server/middleware/authentication/DigestAuth.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/authentication/DigestAuth.scala
@@ -131,6 +131,10 @@ object DigestAuth {
       challenged(runChallenge)
     }
 
+  @deprecated(
+    "Uses a side-effecting NonceKeeper. Use challenge(String, AuthStore, Blocker, Duration, Int, Int).",
+    "0.22.13",
+  )
   def challenge[F[_], A](
       realm: String,
       store: String => F[Option[(A, String)]],
--- a/server/src/main/scala/org/http4s/server/middleware/authentication/Nonce.scala
+++ b/server/src/main/scala/org/http4s/server/middleware/authentication/Nonce.scala
@@ -21,8 +21,10 @@ import org.http4s.crypto.unsafe.SecureRandom
 import java.math.BigInteger
 import java.util.Date
 
+@deprecated("Contains mutable java.util.Date. Use Nonce.F", "0.22.13")
 private[authentication] class Nonce(val created: Date, var nc: Int, val data: String)
 
+@deprecated("Untracked side effects. Use NonceF.", "0.22.13")
 private[authentication] object Nonce {
   val random = new SecureRandom()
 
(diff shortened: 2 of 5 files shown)


server/src/main/scala/org/http4s/server/middleware/authentication/DigestAuth.scala
```diff
@@ -131,6 +131,10 @@ object DigestAuth {
       challenged(runChallenge)
     }
 
+  @deprecated(
+    "Uses a side-effecting NonceKeeper. Use challenge(String, AuthStore, Blocker, Duration, Int, Int).",
+    "0.22.13",
+  )
   def challenge[F[_], A](
```
rossabaker, line 138: I'm not sure whether this signature is still valid, but I thought `NonceKeeper` was considered dirty. 

/cc @blast-hardcheese

server/src/main/scala/org/http4s/server/middleware/authentication/DigestAuth.scala
```diff
@@ -131,6 +131,10 @@ object DigestAuth {
       challenged(runChallenge)
     }
 
+  @deprecated(
+    "Uses a side-effecting NonceKeeper. Use challenge(String, AuthStore, Blocker, Duration, Int, Int).",
+    "0.22.13",
+  )
```
rossabaker, line 137: I'm not sure whether this signature is still valid, but I thought `NonceKeeper` was considered dirty. 

/cc @blast-hardcheese

server/src/main/scala/org/http4s/server/middleware/authentication/NonceKeeperF.scala
```diff
@@ -42,14 +43,15 @@ private[authentication] object NonceKeeperF {
     currentMillis <- Clock[F].monotonic(MILLISECONDS)
     lastCleanupMillis <- Ref[F].of(currentMillis)
     nonces = new LinkedHashMap[String, NonceF[F]]
+    random <- Random.javaSecuritySecureRandom[F](blocker)
```
rossabaker, line 46: This is as high up as we can pull `Random` without making it public API.  In Cats-Effect 3, we can consider surfacing it all the way to the top.

server/src/main/scala/org/http4s/server/middleware/authentication/Nonce.scala
```diff
@@ -21,8 +21,10 @@ import org.http4s.crypto.unsafe.SecureRandom
 import java.math.BigInteger
 import java.util.Date
 
+@deprecated("Contains mutable java.util.Date. Use Nonce.F", "0.22.13")
```
danicheg, line 24:
```suggestion
@deprecated("Contains mutable java.util.Date. Use NonceF.", "0.22.13")
```

danicheg (approved):

rossabaker:
I'll hold just to confirm that new deprecation was desired.

server/src/main/scala/org/http4s/server/middleware/authentication/DigestAuth.scala
blast-hardcheese: It's impossible to be called publicly, since `NonceKeeper` is package private -- I had a `@deprecated` on it but it must have gotten lost. Thanks for putting it back

blast-hardcheese (approved):
Excellent, thanks for the assistance here!

rossabaker:
:+1: Okay, yeah, I thought that was maybe the signature that motivated this whole thing.

PR #6178, opened 2022-03-24
Merge 0.22 -> 0.23

rossabaker:
The heavy lift on this one is getting the new DigestAuth to CE3.

It still has the blocking Random embedded.  We'll need to adjust this further in light of #6177.


rossabaker:
I preemptively merged #6177.  I did not yet surface `Random` in the public API.

server/shared/src/main/scala/org/http4s/server/middleware/authentication/NonceF.scala
```diff
@@ -32,16 +32,22 @@ private[authentication] class NonceF[F[_]](
...
+  private def getRandomData[F[_]: Functor](random: Random[F], bits: Int): F[String] = {
+    val bytes = (bits + 7) / 8
+    random.nextBytes(bytes).map { arr =>
+      if (arr.nonEmpty) {
+        val extraBits = 8 * bytes - bits
+        arr(0) = (arr(0) & ((1 << (8 - extraBits)) - 1)).toByte
+      }
+      new BigInteger(arr).toString(16)
+    }
+  }
```
rossabaker, line 44: I think this is about the same as `Random.nextBigInt` from our CE2 shim.  We can't add it to `cats.effect.std.Random`, but we could create a derived method for it.

rossabaker:
Whoopsies: https://github.com/typelevel/cats-effect/issues/2902

server/shared/src/main/scala/org/http4s/server/middleware/authentication/NonceF.scala
rossabaker: To hell with your BigInt.

server/shared/src/main/scala/org/http4s/server/middleware/authentication/Nonce.scala
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+package org.http4s.server.middleware.authentication
+
+import org.http4s.crypto.unsafe.SecureRandom
+
+import java.math.BigInteger
+import java.util.Date
+
+@deprecated("Contains mutable java.util.Date. Use Nonce.F", "0.22.13")
```
danicheg, line 24:
```suggestion
@deprecated("Contains mutable java.util.Date. Use NonceF.", "0.22.13")
```

server/shared/src/main/scala/org/http4s/server/middleware/authentication/NonceKeeperF.scala
```diff
@@ -0,0 +1,134 @@
...
+      staleTimeout: Duration,
+      nonceCleanupInterval: Duration,
+      bits: Int,
+  )(implicit F: Async[F]): F[NonceKeeperF[F]] = for {
+    // This semaphore controls who has access to `nonces` during stale nonce eviction. This must never be set above one.
+    semaphore <- Semaphore[F](1)
+    current <- F.monotonic
+    lastCleanupMillis <- Ref[F].of(current)
+    nonces = new LinkedHashMap[String, NonceF[F]]
+    random <- Random.javaSecuritySecureRandom[F]
```
armanbilge, line 41: I see in CE3 this is implemented as:
```scala
  def javaSecuritySecureRandom[F[_]: Sync]: F[Random[F]] =
    Sync[F].delay(new java.security.SecureRandom).flatMap(r => javaUtilRandom(r))
```

We can do the same thing and unblock JS:
```scala
  def javaSecuritySecureRandom[F[_]: Sync]: F[Random[F]] =
    Sync[F].delay(new org.http4s.crypto.unsafe.SecureRandom).flatMap(r => javaUtilRandom(r))
```

core/shared/src/main/scala/org/http4s/internal/package.scala
```diff
@@ -329,4 +330,13 @@ package object internal {
       }
     }
   }
+
+  private[http4s] def javaMajorVersion[F[_]](implicit F: Sync[F]): F[Option[Int]] =
```
danicheg, line 334: This seems unused. Do we want this within `0.23`?

danicheg (approved):

core/shared/src/main/scala/org/http4s/internal/package.scala
rossabaker: it's used in 0.22.  Something like it will need to be in Cats-Effect if we fix its Random.  I left it mainly because I don't want to write it again if we ever need it, but it wasn't hard.  I guess we can prune it.

server/shared/src/main/scala/org/http4s/server/middleware/authentication/NonceKeeperF.scala
rossabaker: Oops.  I didn't remember making [6753dd0](https://github.com/http4s/http4s/pull/6178/commits/6753dd0265c41a20b6f4e01dedef1a1580974508) part of this.  But it's a tight spot either way:

* If we do it, we want a consensus on `cats.effect.std.SecureRandom` before we release.
* If we don't do it, we can't easily introduce `Random` to the API later without another crap overload.

rossabaker: I'm going to prune that commit because there's a discussion of it on another draft.

PR #6179, opened 2022-03-24
Blaze server enhancements

danicheg:
This PR brings several proposals to the blaze server:
1. Tries to reduce lock contention in case of saving cancel token.
2. Avoid canceling the previous request in the case of failing the current one.

### DIFF
--- a/blaze-server/src/main/scala/org/http4s/blaze/server/Http1ServerStage.scala
+++ b/blaze-server/src/main/scala/org/http4s/blaze/server/Http1ServerStage.scala
@@ -30,10 +30,9 @@ import org.http4s.blaze.http.parser.BaseExceptions.ParserException
 import org.http4s.blaze.pipeline.Command.EOF
 import org.http4s.blaze.pipeline.TailStage
 import org.http4s.blaze.pipeline.{Command => Cmd}
-import org.http4s.blaze.util.BufferTools
+import org.http4s.blaze.util.{BufferTools, TickWheelExecutor}
 import org.http4s.blaze.util.BufferTools.emptyBuffer
 import org.http4s.blaze.util.Execution._
-import org.http4s.blaze.util.TickWheelExecutor
 import org.http4s.blazecore.Http1Stage
 import org.http4s.blazecore.IdleTimeoutStage
 import org.http4s.blazecore.util.BodylessWriter
@@ -48,6 +47,7 @@ import org.typelevel.vault._
 
 import java.nio.ByteBuffer
 import java.util.concurrent.TimeoutException
+import java.util.concurrent.atomic.AtomicReference
 import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
 import scala.concurrent.duration.Duration
@@ -124,7 +124,7 @@ private[blaze] class Http1ServerStage[F[_]](
   // protected by synchronization on `parser`
   private[this] val parser = new Http1ServerParser[F](logger, maxRequestLineLen, maxHeadersLen)
   private[this] var isClosed = false
-  private[this] var cancelToken: Option[CancelToken[F]] = None
+  private[this] val cancelToken = new AtomicReference[Option[CancelToken[F]]](None)
 
   val name = "Http4sServerStage"
 
@@ -235,9 +235,7 @@ private[blaze] class Http1ServerStage[F[_]](
               }.unsafeRunSync()
             )
 
-            parser.synchronized {
-              cancelToken = theCancelToken
-            }
+            cancelToken.set(theCancelToken)
           }
         })
       case Left((e, protocol)) =>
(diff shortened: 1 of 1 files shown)


blaze-server/src/main/scala/org/http4s/blaze/server/Http1ServerStage.scala
```diff
@@ -229,15 +229,13 @@ private[blaze] class Http1ServerStage[F[_]](
               F.runCancelable(action) {
                 case Right(()) => IO.unit
                 case Left(t) =>
-                  IO(logger.error(t)(s"Error running request: $req")).attempt *> IO(
-                    closeConnection()
-                  )
+                  IO(logger.error(t)(s"Error running request: $req")).attempt *>
+                    IO.delay(cancelToken.set(None)) *>
```
danicheg, line 233: The `closeConnection()` is cancel previous request. Probably after successful request processing, nothing is happening in the case of that canceling. But I think it's better to avoid that canceling at all.

blaze-server/src/main/scala/org/http4s/blaze/server/Http1ServerStage.scala
```diff
@@ -229,15 +229,13 @@ private[blaze] class Http1ServerStage[F[_]](
...
+                    IO(closeConnection())
               }.unsafeRunSync()
             )
 
-            parser.synchronized {
```
danicheg, line 238: I actually don't get why we need that synchronization here. If anyone can explain this, I'd be very grateful.

rossabaker: `cancel()` is also synchronized on `parser`, so it linearizes those.  I'm not awake enough yet to be 100% sure to what end.

wjoel:
Cool. A port to of this to the 1.x branch improves performance by 3% on plaintext, and 13% on JSON, in the TechEmpower framework benchmarks that I'm playing around with. Nice work!

danicheg:
@wjoel wow, thank you so much for benchmarking this!

rossabaker (approved):
That is good news.  Thanks, @wjoel!

Unfortunately, my other comment got stuck in pending...

blaze-server/src/main/scala/org/http4s/blaze/server/Http1ServerStage.scala
```diff
@@ -124,7 +124,7 @@ private[blaze] class Http1ServerStage[F[_]](
   // protected by synchronization on `parser`
   private[this] val parser = new Http1ServerParser[F](logger, maxRequestLineLen, maxHeadersLen)
   private[this] var isClosed = false
-  private[this] var cancelToken: Option[CancelToken[F]] = None
+  private[this] val cancelToken = new AtomicReference[Option[CancelToken[F]]](None)
```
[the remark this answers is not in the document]
danicheg: And, yeah, sorry. I didn't address your question fully. As I said above, I didn't get why that synchronization was needed. And so my original intent was to reduce lock contentions by changing `synchronized` to non-blocking atomic.

danicheg:
@wjoel If you have resources would be super-helpful if you run performance tests with recent changes one more time. 🙏🏻 I bet to no changes in results.

wjoel:
@danicheg Sure, happy to run them again tomorrow, but I won't have time today.

wjoel:
@danicheg I managed to find the time after all, and the news are good, the wins might be within the margin of error but it looks like it's doing better at lower levels of concurrency, and maybe absolutely better on plaintext.

AtomicReference: https://www.techempower.com/benchmarks/#section=test&shareid=32961ba7-8804-4348-8928-5997525ed9c5&hw=ph&test=json&a=2
volatile: https://www.techempower.com/benchmarks/#section=test&shareid=a75aa08c-0529-43c1-b0f7-15fd76b67d39&hw=ph&test=json&a=2

blaze-server/src/main/scala/org/http4s/blaze/server/Http1ServerStage.scala
rossabaker: @RafalSumislawski moved the running of the action out of this block in #4761, but left the synchronization.  I'd be interested to hear if he can think of any reason it needs to remain.

danicheg:
Perfect, thanks again @wjoel

blaze-server/src/main/scala/org/http4s/blaze/server/Http1ServerStage.scala
RafalSumislawski: We need this synchronised for it's Java Memory Model effects. We need a `cancelToken = theCancelToken` -happens-before-> `cancelToken.foreach {...}`. The syncronized provides that thanks to the fact that the `cancelToken.foreach {...}` sits in the `cancel` methood which is called only from `parser.syncrionized{...}` block inside `stageShutdown`. 

`volatile` will do the job as well.

RafalSumislawski (approved):

PR #6180, opened 2022-03-24
Clean up `BlazeServerBuilder` scaladoc

danicheg:
A tiny clean-up.


rossabaker (approved):
For all the linters we have in modern times, it's surprising that nothing complains about these.

PR #6181, opened 2022-03-24
Surface Random

rossabaker:
Demonstration of what `Random` looks like in the public API.  We want to decide before the next 0.23.x release, and also decide on the fate of `SecureRandom` in Cats-Effect 3 (https://github.com/typelevel/cats-effect/pull/2905).

### DIFF
--- a/blaze-server/src/main/scala/org/http4s/blaze/server/BlazeServerBuilder.scala
+++ b/blaze-server/src/main/scala/org/http4s/blaze/server/BlazeServerBuilder.scala
@@ -77,9 +77,12 @@ import scala.concurrent.duration._
   * @param connectorPoolSize: Number of worker threads for the new Socket Server Group
   * @param bufferSize: Buffer size to use for IO operations
   * @param isHttp2Enabled: Whether or not to enable Http2 Server Features
+  * @param maxRequestLineLen: Maximum request line to parse
+  *    If exceeded returns a 400 Bad Request.
   * @param maxHeadersLen: Maximum data that composes the headers.
   *    If exceeded returns a 400 Bad Request.
   * @param chunkBufferMaxSize Size of the buffer that is used when Content-Length header is not specified.
+  * @param httpApp: The services that are mounted on this server to serve..
   * @param serviceErrorHandler: The last resort to recover and generate a response
   *    this is necessary to recover totality from the error condition.
   * @param banner: Pretty log to display on server start. An empty sequence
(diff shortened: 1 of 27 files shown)


rossabaker:
This PR really begins at 6753dd0265c41a20b6f4e01dedef1a1580974508.

rossabaker:
There will be one raw `Random` left, in [`PoolManager`](https://github.com/http4s/http4s/blob/ef00f5180eb51a66b88cd9ae7b419d6737407e86/blaze-client/src/main/scala/org/http4s/blaze/client/PoolManager.scala#L244).  We could surface that, but it doesn't save us from an `Async` constraint, and I can't think of a compelling reason anyone would want to control it.  I'm inclined to let that one be.

rossabaker:
I'm going to rebase this one after #6178 lands.  But here's the discussion before the next 0.23 release:

* If we release with `Random` in the API, we can't take advantage of the `SecureRandom` algebra if it materializes.
* If we don't release with `Random` in the API, we can't until 1.0, unless we overload some more.
    * That code already needs an `Async` constraint.
    * But there's no way to customize the nonce generation without it.

armanbilge:
> If we release with Random in the API, we can't take advantage of the SecureRandom algebra if it materializes.

Why not, can't we add overloads here too?

rossabaker:
Oh, yeah, I guess we can.  It's just going to be ugly and involve another deprecation for something that was around for a couple weeks.

armanbilge:
Right, so the third road here is that we keep all of this new stuff private for another cycle(s) of releases until a decision is made on `SecureRandom` and/or it's shipped. Personally I think it's a great idea.

rossabaker:
I'm out Monday-Wednesday of next week.  Unless someone else does a release, there's a chance it has all come together by the time I can turn the crank again anyway.

@blast-hardcheese is the one who needs the new feature, and he submitted it for 0.22.  We could even do an 0.22 release out a little ahead of 0.23.

blast-hardcheese:
I'm still fine with my workarounds, if it makes sense to delay for a bit I'm not impacted if it makes things easier for the maintainers, I'm still just grateful for the help getting my changes in

rossabaker:
Eek.  I don't remember merging this.  :facepalm:

rossabaker:
Oh, it's because this commit was included and reverted in the other one.  I guess I should have made an issue instead of a draft.

armanbilge:
Btw @blast-hardcheese your changes are included in the 0.22 snapshots if that's helpful.

PR #6182, opened 2022-03-24
Deprecate blaze TrustingSslContext

rossabaker:
It's unused publicly.  A safer version is ported to the test that uses it.

The unfitness for production extends beyond the blocking call, for the record.

### DIFF
--- a/blaze-client/src/test/scala/org/http4s/blaze/client/BlazeClientBase.scala
+++ b/blaze-client/src/test/scala/org/http4s/blaze/client/BlazeClientBase.scala
@@ -42,14 +42,16 @@ import scala.concurrent.duration._
 trait BlazeClientBase extends Http4sSuite {
   val tickWheel: TickWheelExecutor = new TickWheelExecutor(tick = 50.millis)
 
-  val TrustingSslContext: IO[SSLContext] = IO {
+  val TrustingSslContext: IO[SSLContext] = IO.blocking {
     val trustManager = new X509TrustManager {
       def getAcceptedIssuers(): Array[X509Certificate] = Array.empty
       def checkClientTrusted(certs: Array[X509Certificate], authType: String): Unit = {}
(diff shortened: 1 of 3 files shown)


blaze-client/src/test/scala/org/http4s/blaze/client/BlazeClientBase.scala
```diff
@@ -33,19 +33,31 @@ import org.http4s.client.scaffold._
...
   val tickWheel: TickWheelExecutor = new TickWheelExecutor(tick = 50.millis)
 
+  val TrustingSslContext: IO[SSLContext] = IO {
+    val trustManager = new X509TrustManager {
+      def getAcceptedIssuers(): Array[X509Certificate] = Array.empty
+      def checkClientTrusted(certs: Array[X509Certificate], authType: String): Unit = {}
+      def checkServerTrusted(certs: Array[X509Certificate], authType: String): Unit = {}
+    }
+    (trustManager, SSLContext.getInstance("TLS"))
+  }.flatMap { case (mgr, ctx) => IO.blocking(ctx.init(null, Array(mgr), new SecureRandom)).as(ctx) }
```
RafalSumislawski, line 52: Just a minor nitpick:
I would prefer putting the whole thing in `IO.blocking`. It seems strange (arbitrary?) that the allocations of the `trustManager` and the `SSLContext` are are pushed to a separate `IO{}` while the allocation of the `Array` and the `SecureRandom` are inside the `IO.blocking`.

RafalSumislawski (approved):

blaze-client/src/test/scala/org/http4s/blaze/client/BlazeClientBase.scala
rossabaker: I like to highlight the call that blocks and keep the CPU work on the CPU pool, but here, the CPU work is trivial.  I switched to something simpler.

PR #6183, opened 2022-03-25
Move WebSocketHandshake to blaze-core

rossabaker:
Found this while looking for hidden `Random`s while that's a hot topic in Cats-Effect.  After I toiled to make it compile, it turns out the client handshake still isn't used, but it does provide a round-trip test to ensure the server side works.

The original version can go poof on merge to main.


danicheg (approved):
Is `WebSocketHandshake` from `core` supposed to be removed in the main?

rossabaker:
Yes, it should be.  I don't think anybody is using it other than blaze.  I only left it because it was scoped broadly enough it might have been used in another repo that I wasn't thinking of.

PR #6186, opened 2022-03-25
Default redactHeadersWhen to include any header with "token"

### LINKED ISSUE #6104: Change Server and Client Logger default redactHeadersWhen to include any header with "token"
As the title says. For example for Server Logger the apply would be:
```scala
def apply[F[_]: Async](
    logHeaders: Boolean,
    logBody: Boolean,
    redactHeadersWhen: CIString => Boolean = (name: String) => Headers.SensitiveHeaders.contains(name) || name.toString.toLowerCase.contains("token"),
    logAction: Option[String => F[Unit]] = None,
)
```

Or even better create a
```scala
def defaultRedactHeadersWhen(name: CIString): Boolean =
    Headers.SensitiveHeaders.contains(name) || name.toString.toLowerCase.contains("token")
```
and then
```scala
def apply[F[_]: Async](
    logHeaders: Boolean,
    logBody: Boolean,
    redactHeadersWhen: CIString => Boolean = defaultRedactHeadersWhen,
    logAction: Option[String => F[Unit]] = None,
 )
```
I can make a PR.

Lasering:
Fixes #6104

### DIFF
--- a/client/shared/src/main/scala/org/http4s/client/middleware/Logger.scala
+++ b/client/shared/src/main/scala/org/http4s/client/middleware/Logger.scala
@@ -25,10 +25,13 @@ import org.typelevel.ci.CIString
 /** Simple Middleware for Logging All Requests and Responses
   */
 object Logger {
+  def defaultRedactHeadersWhen(name: CIString): Boolean =
+    Headers.SensitiveHeaders.contains(name) || name.toString.toLowerCase.contains("token")
+
   def apply[F[_]: Async](
       logHeaders: Boolean,
       logBody: Boolean,
-      redactHeadersWhen: CIString => Boolean = Headers.SensitiveHeaders.contains,
+      redactHeadersWhen: CIString => Boolean = defaultRedactHeadersWhen,
       logAction: Option[String => F[Unit]] = None,
   )(client: Client[F]): Client[F] =
     ResponseLogger.apply(logHeaders, logBody, redactHeadersWhen, logAction)(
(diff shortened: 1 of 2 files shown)


ChristopherDavenport (approved):

ChristopherDavenport:
Do we want to target this to 0.22 so we can forward it?

danicheg (commented):
Since it's possible to configure which headers need to be redacted on the user side, I think it's not good to change the default behavior. Other users (who are happy with the current behavior) will need to do changes. So it seems odd a bit. But if the majority are ok with these changes, I won't make waves.

rossabaker (approved):
I don't generally like semantic changes that don't surface themselves via the compiler, but this is a more secure default.  I am in favor.

bplommer (approved):

armanbilge:
Is it just me or seems like this should be retargeted to 0.23?

armanbilge:
@Lasering I see you 👍 https://github.com/http4s/http4s/pull/6186#issuecomment-1222503758 ... if you don't mind re-targeting this PR at series/0.23, I will merge it and it can land in the next release :)

Lasering:
@armanbilge I'm so sorry. I've never re-targeted a PR, and since my commits were on the main branch of my repo (which was out-of-date and not in sync with series/0.23) I thought the easiest way has to delete the main branch of my repo and push a new clean branch of series/0.23 with my commit on top. But as soon as I deleted my main branch on the upstream Github closed this PR and now it doesn't allow me to edit it to change the base branch. So I had to create a new PR. Next time I'll simply force push to the main of my repo.

armanbilge:
@Lasering it's not a problem at all :) thanks for retargeting!

PR #6187, opened 2022-03-26
Remove deprecated `Client.fetch`, `DefaultClient.fetch`

danicheg:
Those methods were deprecated since `0.21`, so I feel we can do that clean-up. If not now, then when? `2.0` is probably too far from nowadays.


armanbilge (approved):

rossabaker (approved):
If we had kept these instead of exposing the resource, we'd be able to trace the use of responses.  I don't think we should go back, but I'm going to be sad in this comment.

[is sad]

[approves]

danicheg:
There is "sad but true". Now we have "sad but approve" too 🙈

PR #6193, opened 2022-03-27
Merge 0.23 -> main

rossabaker:
Deletes the shims left for binary compatibility.  Preserves a few source-breaking deprecations to give users time to catch up.


build.sbt
```diff
@@ -117,7 +117,6 @@ lazy val core = libraryCrossProject("core")
       catsCore.value,
       catsEffectStd.value,
       catsParse.value.exclude("org.typelevel", "cats-core_2.13"),
-      crypto.value,
```
armanbilge, line 120: Oh, do we not need it in core anymore?

build.sbt
```diff
@@ -336,7 +337,8 @@ lazy val blazeCore = libraryProject("blaze-core")
     description := "Base library for binding blaze to http4s clients and servers",
     startYear := Some(2014),
     libraryDependencies ++= Seq(
-      blazeHttp
+      blazeHttp,
+      crypto.value,
```
armanbilge, line 341: Blaze should be able to direclty use JDK crypto APIs since it doesn't need to cross-compile to JS.

build.sbt
rossabaker: WebSocketHandshake was the only use in core.  We could move the improved blaze copy back to core, but nobody else uses it.

build.sbt
rossabaker: Good point, but I'm not going to do that until the prior comment is resolved.

build.sbt
armanbilge: Ah, I remember that change but didn't realize the implications. 👍

client/shared/src/main/scala/org/http4s/client/Client.scala
```diff
@@ -148,18 +150,7 @@ trait Client[F[_]] {
...
-  )(gK: G ~> F)(implicit F: MonadCancelThrow[F]): Client[G] = translateImpl(fk)(gK)
-
-  /** Translates the effect type of this client from F to G
-    */
-  def translate[G[_]: MonadCancelThrow](
-      fk: F ~> G
-  )(gK: G ~> F)(implicit F: MonadCancelThrow[F]): Client[G] = translateImpl(fk)(gK)
-
-  /** As [[expectOptionOr]], but defined in terms of [[cats.data.OptionT]]. */
+  /** As [[#expectOptionOr]], but defined in terms of [[cats.data.OptionT]]. */
```
danicheg, line 153: A little fix of scaladoc
```suggestion
  /** As [[expectOptionOr]], but defined in terms of [[cats.data.OptionT]]. */
```

server/shared/src/main/scala/org/http4s/server/middleware/authentication/DigestAuth.scala
```diff
@@ -19,29 +19,63 @@ package server
...
-import java.util.Date
 import scala.concurrent.duration._
 
 /** Provides Digest Authentication from RFC 2617.
   */
 object DigestAuth {
 
+  @deprecated(
+    "AuthenticationStore is going away, in favor of explicit subclasses of AuthStore. PlainTextAuthStore maintains the previous, insecure behaviour, whereas Md5HashedAuthStore is the new advised implementation going forward.",
+    "0.22.13",
```
danicheg, line 40: Keeping deprecated versions consistent within `0.22.`, `0.23`, `1.0` it's not a cakewalk. It seems like there should be `0.23`. Though, from my way of thinking, it's ok to set `1.0`.

blaze-core/src/main/scala-2.13/org/http4s/blazecore/util/ParasiticExecutionContextCompat.scala
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.http4s.blazecore.util
+
+import scala.concurrent.ExecutionContext
+
+private[util] trait ParasiticExecutionContextCompat {
```
danicheg, line 21: Unused. Shall we burn it?

build.sbt
rossabaker: Neither did I until the plugin caught it. Slimming down unused dependencies is a great thing, even though in this case it's reintroduced in both client and server.

client/shared/src/main/scala/org/http4s/client/Client.scala
rossabaker: Oops, I thought that changed the other direction.

server/shared/src/main/scala/org/http4s/server/middleware/authentication/DigestAuth.scala
rossabaker: I used to just leave them, but it assumes too much familiarity with the linear order of releases. I think this one was a miss in the merge from 0.22.

armanbilge: Feels like using dates instead of versions would save us a lot of trouble, but anyway 😆

rossabaker: Commit date?  Merge date?  Release date?  Date it hits that release series?  The first is useless, and everything else is still some sort of chore.

bplommer:
Looks like an obsolete test snuck in here - I've opened a PR against this branch at https://github.com/rossabaker/http4s/pull/5

armanbilge (approved):

PR #6196, opened 2022-03-27
Merge 0.22 -> 0.23

rossabaker:


blaze-server/src/main/scala/org/http4s/blaze/server/Http1ServerStage.scala
```diff
@@ -210,18 +210,12 @@ private[blaze] class Http1ServerStage[F[_]](
               .flatMap {
                 case Right(_) => F.unit
                 case Left(t) =>
-                  F.delay(logger.error(t)(s"Error running request: $req")).attempt *> F.delay(
-                    closeConnection()
-                  )
+                  F.delay(logger.error(t)(s"Error running request: $req")).attempt *>
+                    F.delay { cancelToken = None } *>
+                    F.delay(closeConnection())
```
rossabaker, line 215: Maybe there should be some `guarantee`s in here instead of the ad hoc `attempt`.  This is just a fairly literal merge.

danicheg: Good point. If nobody minds, I would undertake this.

danicheg (approved):

PR #6198, opened 2022-03-28
Use tryScheduling within IdleTimeoutStage

hamnis:
To avoid nasty stacktraces when shutting down blaze.


danicheg (approved):
Is that would mean that `TickWheelExecutor.AlreadyShutdownException` would gone!? That's tremendous!

hamnis:
@danicheg that is my hope at least. I think this should cover most cases.

PR #6201, opened 2022-03-28
Optimize Headers put and ++ using ListBuffer

wjoel:
I'm opening this as a draft for now, as I have some questions.

1. There's some repetition here, which is done to optimize for performance. Is it worth the added lines? I'd like to think so, as this file hasn't been changed much over time, and the functionality is fairly basic, but if you disagree, I can remove some of it.
2. Is changing "appended to "prepended" in `Message.addHeader` actually a non-breaking change? As far as I can tell, the order doesn't actually matter, and prepending that is more efficient (the number of header values we already have are likely to be larger than the number of header values we're adding), but...
3. We could keep track of header names in a mutable set internally, which would remove the need for an extra traversal in `!headers.exists(_.name == thatHeaderName)`, but would that be worth doing, in terms of added code, memory usage, and mutable state? I haven't done any benchmarks on this, yet.
4. Not directly related to this PR, but I'm wondering if there's any interest in considering something different than a `List` for `val headers: List[Header.Raw]`, as it's not a great choice if we want efficient append while also replacing previous values. Maybe a multimap or other data structure would be more appropriate, as the order doesn't seem to actually matter?

As for the benefit, this improves plaintext performance on Techempower benchmarks from 308k req/s to 314k req/s, which isn't huge but not insignificant for that benchmark, as it's using almost no headers.

The result matches my expectations based on profiling, where this headers stuff take a small, yet surprisingly large amount of time (about 2% of total CPU time).

EDIT: Test failures may have answered question 2. Might still be worth considering for 1.x if we don't go for 4.

### DIFF
--- a/core/shared/src/main/scala/org/http4s/Headers.scala
+++ b/core/shared/src/main/scala/org/http4s/Headers.scala
@@ -72,7 +72,11 @@ final class Headers(val headers: List[Header.Raw]) extends AnyVal {
   def put(in: Header.ToRaw*): Headers =
     in match {
       case thatHeader :: Nil =>
-        putOne(thatHeader)
+        if (this.headers.isEmpty) {
+          Headers(thatHeader)
+        } else {
+          putOne(thatHeader)
+        }
       case Nil =>
         this
       case _ =>
@@ -80,28 +84,19 @@ final class Headers(val headers: List[Header.Raw]) extends AnyVal {
     }
 
   private def putOne(header: Header.ToRaw): Headers =
-    if (this.headers.isEmpty) {
-      Headers(header)
-    } else {
-      header.values match {
-        case values @ head :: _ =>
-          val thatHeaderName = head.name
-          if (!headers.exists(_.name == thatHeaderName)) {
-            Headers(this.headers ++ values)
-          } else {
-            val newHeaders = mutable.ListBuffer.empty[Header.Raw]
-            headers.foreach { h =>
-              if (h.name != thatHeaderName) {
-                newHeaders += h
-              }
-            }
-            newHeaders ++= values
-            Headers(newHeaders.toList)
-          }
-        // Shouldn't really happen, but if it does, there's nothing to be done.
-        case Nil =>
-          this
-      }
+    header.values match {
+      case values @ head :: _ =>
+        val thatHeaderName = head.name
+        if (!headers.exists(_.name == thatHeaderName)) {
+          Headers(this.headers ++ values)
+        } else {
+          val newHeaders = mutable.ListBuffer.from(headers.filterNot(_.name == thatHeaderName))
+          newHeaders ++= values
+          Headers(newHeaders.toList)
+        }
+      // Shouldn't really happen, but if it does, there's nothing to be done.
+      case Nil =>
+        this
     }
 
   def ++(those: Headers): Headers =
@@ -120,7 +115,7 @@ final class Headers(val headers: List[Header.Raw]) extends AnyVal {
           newHeaders += thatHeader
           Headers(newHeaders.toList)
         case thoseHeaders =>
-          val thoseNames = mutable.Set.empty[CIString]
+          val thoseNames = mutable.Set.from(thoseHeaders.map(_.name))
           val newHeaders = mutable.ListBuffer.empty[Header.Raw]
           headers.foreach { h =>
             if (!thoseNames.contains(h.name)) {
@@ -128,7 +123,6 @@ final class Headers(val headers: List[Header.Raw]) extends AnyVal {
             }
           }
           thoseHeaders.foreach { h =>
-            thoseNames += h.name
             newHeaders += h
           }
           Headers(newHeaders.toList)
--- a/core/shared/src/main/scala/org/http4s/Headers.scala
+++ b/core/shared/src/main/scala/org/http4s/Headers.scala
@@ -70,44 +70,47 @@ final class Headers(val headers: List[Header.Raw]) extends AnyVal {
     * @return a new [[Headers]] containing the sum of the initial and input headers
     */
   def put(in: Header.ToRaw*): Headers =
-    in match {
-      case thatHeader :: Nil =>
-        if (this.headers.isEmpty) {
-          Headers(thatHeader)
-        } else {
+    if (this.headers.isEmpty) {
+      Headers(in)
+    } else {
+      in match {
+        case thatHeader :: Nil =>
           putOne(thatHeader)
-        }
-      case Nil =>
-        this
-      case _ =>
-        this ++ Headers(in.values)
+        case Nil =>
+          this
+        case _ =>
+          this ++ Headers(in.values)
+      }
     }
 
   private def putOne(header: Header.ToRaw): Headers =
     header.values match {
       case values @ head :: _ =>
         val thatHeaderName = head.name
-        if (!headers.exists(_.name == thatHeaderName)) {
-          Headers(this.headers ++ values)
-        } else {
-          val newHeaders = mutable.ListBuffer.from(headers.filterNot(_.name == thatHeaderName))
-          newHeaders ++= values
-          Headers(newHeaders.toList)
+        val newHeaders = mutable.ListBuffer.empty[Header.Raw]
+        this.headers.foreach { h =>
+          if (h.name != thatHeaderName) {
+            newHeaders += h
+          }
         }
+        values.foreach { h =>
+          newHeaders += h
+        }
+        Headers(newHeaders.toList)
       // Shouldn't really happen, but if it does, there's nothing to be done.
       case Nil =>
         this
     }
 
   def ++(those: Headers): Headers =
-    if (those.headers.isEmpty) this
-    else if (this.headers.isEmpty) those
+    if (this.headers.isEmpty) those
+    else if (those.headers.isEmpty) this
     else {
       those.headers match {
         case thatHeader :: Nil =>
           val thatHeaderName = thatHeader.name
           val newHeaders = mutable.ListBuffer.empty[Header.Raw]
-          headers.foreach { h =>
+          this.headers.foreach { h =>
             if (h.name != thatHeaderName) {
               newHeaders += h
             }
@@ -115,9 +118,10 @@ final class Headers(val headers: List[Header.Raw]) extends AnyVal {
           newHeaders += thatHeader
           Headers(newHeaders.toList)
         case thoseHeaders =>
-          val thoseNames = mutable.Set.from(thoseHeaders.map(_.name))
+          val thoseNames = mutable.Set.empty[CIString]
+          thoseHeaders.foreach(h => thoseNames += h.name)
           val newHeaders = mutable.ListBuffer.empty[Header.Raw]
-          headers.foreach { h =>
+          this.headers.foreach { h =>
             if (!thoseNames.contains(h.name)) {
               newHeaders += h
             }
--- a/core/shared/src/main/scala/org/http4s/Headers.scala
+++ b/core/shared/src/main/scala/org/http4s/Headers.scala
@@ -111,18 +111,14 @@ final class Headers(val headers: List[Header.Raw]) extends AnyVal {
       those.headers match {
         case thatHeader :: Nil =>
           val thatHeaderName = thatHeader.name
-          if (!headers.exists(_.name == thatHeader.name)) {
-            Headers(this.headers :+ thatHeader)
-          } else {
-            val newHeaders = mutable.ListBuffer.empty[Header.Raw]
-            headers.foreach { h =>
-              if (h.name != thatHeaderName) {
-                newHeaders += h
-              }
+          val newHeaders = mutable.ListBuffer.empty[Header.Raw]
+          headers.foreach { h =>
+            if (h.name != thatHeaderName) {
+              newHeaders += h
             }
-            newHeaders += thatHeader
-            Headers(newHeaders.toList)
           }
+          newHeaders += thatHeader
+          Headers(newHeaders.toList)
         case thoseHeaders =>
           val thoseNames = mutable.Set.empty[CIString]
           val newHeaders = mutable.ListBuffer.empty[Header.Raw]
--- a/core/shared/src/main/scala/org/http4s/Headers.scala
+++ b/core/shared/src/main/scala/org/http4s/Headers.scala
@@ -70,15 +70,73 @@ final class Headers(val headers: List[Header.Raw]) extends AnyVal {
     * @return a new [[Headers]] containing the sum of the initial and input headers
     */
   def put(in: Header.ToRaw*): Headers =
-    this ++ Headers(in.values)
+    in match {
+      case thatHeader :: Nil =>
+        putOne(thatHeader)
+      case Nil =>
+        this
+      case _ =>
+        this ++ Headers(in.values)
+    }
+
+  private def putOne(header: Header.ToRaw): Headers =
+    if (this.headers.isEmpty) {
+      Headers(header)
+    } else {
+      header.values match {
+        case values @ head :: _ =>
+          val thatHeaderName = head.name
+          if (!headers.exists(_.name == thatHeaderName)) {
+            Headers(this.headers ++ values)
+          } else {
+            val newHeaders = mutable.ListBuffer.empty[Header.Raw]
+            headers.foreach { h =>
+              if (h.name != thatHeaderName) {
+                newHeaders += h
+              }
+            }
+            newHeaders ++= values
+            Headers(newHeaders.toList)
+          }
+        // Shouldn't really happen, but if it does, there's nothing to be done.
+        case Nil =>
+          this
+      }
+    }
 
   def ++(those: Headers): Headers =
     if (those.headers.isEmpty) this
     else if (this.headers.isEmpty) those
     else {
-      val thoseNames = mutable.Set.empty[CIString]
-      those.headers.foreach(h => thoseNames.add(h.name))
-      Headers(headers.filterNot(h => thoseNames.contains(h.name)) ++ those.headers)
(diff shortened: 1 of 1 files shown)


wjoel:
Looks like some tests rely on the order of the headers, but maybe they shouldn't, so I'm awaiting feedback before making any further changes.

Daenyth:
> Is changing "appended to "prepended" in Message.addHeader actually a non-breaking change?

I'm not as familiar with the HTTP specs as other folks here, but I'm pretty sure this is a breaking change and the order matters

Daenyth:
>  The order in which header fields with the same field-name are received is therefore significant to the interpretation of the combined field value, and thus a proxy MUST NOT change the order of these field values when a message is forwarded.

https://www.rfc-editor.org/rfc/rfc2616#section-4.2

wjoel:
> > The order in which header fields with the same field-name are received is therefore significant to the interpretation of the combined field value, and thus a proxy MUST NOT change the order of these field values when a message is forwarded.
> 
> https://www.rfc-editor.org/rfc/rfc2616#section-4.2

That's with the same field name, though. That could be done with a `Map[String, List[String]]`, for example, right? vert.x, Netty HTTP, Undertow, and many other frameworks use maps to store headers, so I think http4s should be able to do so as well.

armanbilge:
@wjoel in light of https://github.com/http4s/http4s/pull/6578#discussion_r942508403 seems the preference is to preserve header order. Can we still do some optimizations here with that restriction?

wjoel:
@armanbilge it might be possible to rework this PR to preserve order while still using `ListBuffer`, at least.

But for 1.0 I'd really, really like to move away from using a `List` to represent headers, and instead use something like `Map[String, List[String]]` (or at least something map-like), and not give any guarantees about the order. I don't think lists are particularly nice for this purpose, and searching through it over and over again to find certain headers is really inefficient. And it's really inefficient to search through it to the end to append headers as well.

At work we have a lot of headers on requests that have passed through an API gateway and auth service, and I think that's only going to be more common in the future (as API gateways and microservice architectures become more popular), so I think efficiency does matter here.

sideeffffect:
Please excuse a humble question from a random person on the internet:

> move away from using a `List` to represent headers, and instead use something like `Map[String, List[String]]`

Is it really _either/or_? Could we have
 * `List[Header.Raw]` as the underlying, order preserving representation of headers (or `Vector` or something else sequence-like, whatever comes up good in benchmarks)
 * `Map[String, List[String]]` (or `Map[String, Vector[String]]` or ...) as a `lazy val`, which will get computed only at the first time somebody want to search for particular header names(s)
?

wjoel:
There's nothing to excuse, I'm just a random person as well. :)

The problem with keeping `List` as the underlying representation is that http4s searches for headers like `Keep-Alive` on every request, and any non-trivial use of http4s will use middlewares like CORS which searches for `Vary` on every response, so in reality that `lazy val` would practically always be computed.

In addition, if the `List` was the underlying representation, all the "add this header" operations would still be `O(n)` (and there are many of those in our middlewares).

Some of the searches could perhaps be removed, and some of the appends be aggregated in middlewares like CORS, but that's just working around the use of `List` and it would require every middleware author to be aware of the implementation of headers and the cost or various operations. The need for such work arounds is, in my opinion, an indication that `List` isn't a good choice here. If `Map` (or something with similar operation costs) was used instead, there would be less to worry about.

sideeffffect:
OK, so if I understand the issue correctly, we often need to append to headers (at the end), which is O(n) with `List`. So why not then use `Vector[Header.Raw]`? And have the `Map` as a `lazy val` as I've described above.

sideeffffect:
Or, as per https://www.rfc-editor.org/rfc/rfc2616#section-4.2

> _The order in which header fields with differing field names are
   received is not significant._ However, it is "good practice" to send
   general-header fields first, followed by request-header or response-
   header fields, and ending with the entity-header fields.

have just `Map[String, Vector[String]]`...
:shrug:

wjoel:
> Or, as per https://www.rfc-editor.org/rfc/rfc2616#section-4.2
> 
> > _The order in which header fields with differing field names are
> > received is not significant._ However, it is "good practice" to send
> > general-header fields first, followed by request-header or response-
> > header fields, and ending with the entity-header fields.
> 
> have just `Map[String, Vector[String]]`... shrug

Indeed, why not just use `Map` so we get efficient lookups and adds, is what I'm saying. :)

sideeffffect:
Yep, you and the RFC have convinced me :smile_cat: So `Map[String, Vector[String]]` in the end?

wjoel:
That makes two random people on the Internet who thinks that would be a good idea (at least for 1.0, where we can still make breaking changes until it's released), but I don't know if there's any consensus among http4s maintainers on this. Doesn't seem like it, but I don't know.

armanbilge:
FWIW I don't see why it's _bad_ idea and it does match the spec, so I would approve such a PR targeted at `main`. I wouldn't feel confident enough to merge it though 😅

CremboC:
Hey, is there any way we can progress this and merge into 0.23.x? I've done some profiling and indeed it seems like adding headers is obscenely expensive 😅 

We have this basic middleware that simply replaces the `Server` header that Jetty appends by default:
```scala
object ResponseHeaderCleanerMiddleware {

  final private[this] val ServerHeader = Server(ProductId("Permutive"))

  /**
    * Overwrites `Server` header to avoid Jetty filling it with its version.
    * This is to reduce risk of exploit based on known vulnerabilities of the HTTP server.
    */
  def apply[F[_]](app: HttpApp[F])(implicit f: Functor[F]): HttpApp[F] =
    app.map { response =>
      response.putHeaders(ServerHeader)
    }

}
```

And this one tiny `putHeaders` call results in the following stack trace and approx 16% of memory allocations in the span of 5 minutes:
image

sideeffffect:
Btw, maybe it should be `Map[String, NonEmptyVector[String]]`, because if the header name is in the map, the values are guaranteed to be there. We do have `NonEmptyVector` available, right?

wjoel:
> Hey @wjoel, is there any way we can progress this and merge into 0.23.x? I've done some profiling and indeed it seems like adding headers is obscenely expensive sweat_smile
>
> ...
> 
> And this one tiny `putHeaders` call results in the following stack trace and approx 16% of memory allocations in the span of 5 minutes:

Yes, it's pretty wild, isn't it? Thanks for confirming what I saw all those months ago, and for reminding me of just how bad this is. :)

I'll resume work on this PR today and/or tomorrow, and try to make it pass the tests as they are, perhaps creating another PR that removes the order preservation of the headers (which is controversial to some, but not me).

wjoel:
@CremboC did you try running that test with the changes in this PR, btw?

wjoel:
Since this no longer includes the potentially controversial reordering of headers, and passes all (non-flaky) tests, I believe it is ready for review.

I still believe the order in which headers are added shouldn't be respected by http4s (as it's not respected by the HTTP specification), and I don't think `List` is a good choice to model headers, but that can be discussed/changed in other PRs.

core/shared/src/main/scala/org/http4s/Headers.scala
```diff
@@ -70,15 +70,73 @@ final class Headers(val headers: List[Header.Raw]) extends AnyVal {
...
+          } else {
+            val newHeaders = mutable.ListBuffer.empty[Header.Raw]
+            headers.foreach { h =>
+              if (h.name != thatHeaderName) {
+                newHeaders += h
+              }
+            }
+            newHeaders += thatHeader
+            Headers(newHeaders.toList)
+          }
```
armanbilge, line 125: Is it just me or is the `!headers.exists(...)` doing an extra traversal? It seems like no matter whether the result is `true` or `false` we end up creating a new list and appending to the end.

wjoel: It does, that's 3. from my original comment:

3. We could keep track of header names in a mutable set internally, which would remove the need for an extra traversal in `!headers.exists(_.name == thatHeaderName)`, but would that be worth doing, in terms of added code, memory usage, and mutable state? I haven't done any benchmarks on this, yet.

wjoel: My gut feeling is that it would be worth it if we insist on staying with `List`, but not worth it if we want change the underlying data structure to something more appropriate, and I'm strongly in favor of the latter.

armanbilge: Hum, sorry, I'm confused. Can't we write this as equivalently like this? Apologies if I am being stupid 😅 
```suggestion
        case thatHeader :: Nil =>
          val thatHeaderName = thatHeader.name
          val newHeaders = mutable.ListBuffer.empty[Header.Raw]
          headers.foreach { h =>
            if (h.name != thatHeaderName) {
              newHeaders += h
            }
          }
          newHeaders += thatHeader
          Headers(newHeaders.toList)
```

wjoel: Although it's late over here, you're definitely not stupid, and it looks to me like that would work. The tests agree. It might have been more reasonable when doing preprends. :thinking:

armanbilge: Yes, when `prepend`ing you wouldn't have to copy the list. But in this case we're copying no matter what.

CremboC:
hey @wjoel unfortunately I cannot run the code in the same environment I ran earlier (where I provided the flamegraph). Best second thing I could do is running the benchmark I added in https://github.com/http4s/http4s/pull/6718, together with JFR enabled `-prof jfr`. It seems like in that particular test there isn't any noticable difference in performance or memory allocations unfortunately! It's perfectly plausible there is a difference in other tests, however.

wjoel:
> hey @wjoel unfortunately I cannot run the code in the same environment I ran earlier (where I provided the flamegraph). Best second thing I could do is running the benchmark I added in #6718, together with JFR enabled `-prof jfr`. It seems like in that particular test there isn't any noticable difference in performance or memory allocations unfortunately! It's perfectly plausible there is a difference in other tests, however.

No worries, your benchmarks were still somewhat useful, I found a mistake on my part and made some minor improvements with their help. It's not surprising that you didn't see much of a difference on them as they are, but it gets slightly more interesting if you add more than one `Content-Length` (then the improved check to see if it must be replaced is useful), or even just moving that one to the end seems to make a difference.

I see quite a bit of variance on those benchmarks (in part, I believe, because it makes a difference depending on which Ryzen core ends up being used), even after I upped it to 10 warmups and 10 iterations, but:
```
main, single content-length (first):
[info] Benchmark                Mode  Cnt     Score    Error  Units
[info] HeadersBench.putHeaders  avgt   10  1032.155 ± 90.169  ns/op
main, single content-length (last):
[info] Benchmark                Mode  Cnt    Score    Error  Units
[info] HeadersBench.putHeaders  avgt   10  908.060 ± 11.742  ns/op
main, two content-length (first, and last):
[info] Benchmark                Mode  Cnt     Score     Error  Units
[info] HeadersBench.putHeaders  avgt   10  1329.752 ± 107.671  ns/op

This PR, single content-length (first):
[info] Benchmark                Mode  Cnt    Score    Error  Units
[info] HeadersBench.putHeaders  avgt   10  752.709 ± 16.718  ns/op
This PR, single content-length (last):
[info] Benchmark                Mode  Cnt    Score     Error  Units
[info] HeadersBench.putHeaders  avgt   10  872.468 ± 179.501  ns/op
This PR, two content-length (first, and last):
[info] Benchmark                Mode  Cnt     Score   Error  Units
[info] HeadersBench.putHeaders  avgt   10  1185.210 ± 4.789  ns/op
```

... and test failures tell me that `.from` isn't available on Scala 2.12, so I'll have to continue this another day. I'd like to re-run a full TFB test for plaintext, as well.

wjoel:
Going to have to run some more real-world(ish) benchmarks tomorrow, currently this isn't looking too great, although the lowered gc isn't bad, and might make a bigger difference on end-to-end benchmarks.
Using `++2.13 bench/jmh:run -i 10 -wi 5 -f 1 -t 1 -prof gc org.http4s.bench.HeadersBench` in sbt shell with two content-length headers (first, and last):
```
This PR
Benchmark                                             Mode  Cnt     Score     Error   Units
HeadersBench.putHeaders                               avgt   10  1511.238 ± 224.235   ns/op
HeadersBench.putHeaders:·gc.alloc.rate                avgt   10  3942.249 ± 574.347  MB/sec
HeadersBench.putHeaders:·gc.alloc.rate.norm           avgt   10  6504.000 ±   0.001    B/op
HeadersBench.putHeaders:·gc.churn.G1_Eden_Space       avgt   10  3944.447 ± 574.498  MB/sec
HeadersBench.putHeaders:·gc.churn.G1_Eden_Space.norm  avgt   10  6507.717 ±  44.131    B/op
HeadersBench.putHeaders:·gc.churn.G1_Old_Gen          avgt   10     0.011 ±   0.003  MB/sec
HeadersBench.putHeaders:·gc.churn.G1_Old_Gen.norm     avgt   10     0.019 ±   0.004    B/op
HeadersBench.putHeaders:·gc.count                     avgt   10   475.000            counts
HeadersBench.putHeaders:·gc.time                      avgt   10   313.000                ms

main:
Benchmark                                             Mode  Cnt     Score      Error   Units
HeadersBench.putHeaders                               avgt   10  1215.887 ±  246.980   ns/op
HeadersBench.putHeaders:·gc.alloc.rate                avgt   10  4957.511 ± 1022.279  MB/sec
HeadersBench.putHeaders:·gc.alloc.rate.norm           avgt   10  6528.000 ±    0.001    B/op
HeadersBench.putHeaders:·gc.churn.G1_Eden_Space       avgt   10  4967.840 ± 1008.995  MB/sec
HeadersBench.putHeaders:·gc.churn.G1_Eden_Space.norm  avgt   10  6543.204 ±   93.237    B/op
HeadersBench.putHeaders:·gc.churn.G1_Old_Gen          avgt   10     0.011 ±    0.005  MB/sec
HeadersBench.putHeaders:·gc.churn.G1_Old_Gen.norm     avgt   10     0.015 ±    0.005    B/op
HeadersBench.putHeaders:·gc.count                     avgt   10   494.000             counts
HeadersBench.putHeaders:·gc.time                      avgt   10   356.000                 ms
```

Where "first, and last" means:
```scala
  @Benchmark
  def putHeaders: Response[IO] =
    baseMessage
      .putHeaders(`Content-Length`(1000L))
      .putHeaders(`Content-Encoding`(ContentCoding.gzip))
      .putHeaders(Server(ProductId("Benchmark")))
      .putHeaders(`Content-Length`(100L))
```

wjoel:
Some numbers from (sorry) TechEmpower Framework Benchmarks. Starting with 0.23 (main, this PR, 0.23.15), the results are:
```
main 0.23 0.23.16-26-6b0a05e-SNAPSHOT
512 json Requests/sec: 156496.53
4096 plaintext Requests/sec: 187207.02

faster-headers 0.23.16-30-c93c7a9-SNAPSHOT
512 json Requests/sec: 158389.16
4096 plaintext Requests/sec: 189174.36

0.23.15
256 json Requests/sec: 154661.20
4096 plaintext Requests/sec: 190817.51
```
Practically no difference, I'd say it's within the margin of error. Possibly a slight improvement.

Moving on to 1.0.0, and this pretty much https://github.com/wjoel/FrameworkBenchmarks/tree/actually-faster-http4s but adjusted to use ByteVector instead of Chunk.array:
```
1.0.0-M37
512 json Requests/sec: 231400.15
4096 plaintext Requests/sec: 263929.75
1.0.0-M37 faster-headers 1.0-23151b7-20220928T183554Z-SNAPSHOT
512 json Requests/sec: 234754.53
4096 plaintext Requests/sec: 261429.34
```
Basically a tie, running those benchmarks more than once could well change the numbers by a few thousands one way or the other.

In case we're limited by other things, I ran the benchmarks with http4s-netty-server (compiled from main, as it's not yet released for 1.0.0-M37):
```
1.0.0-M37 Netty faster-headers
512 json Requests/sec: 427374.49
1024 plaintext Requests/sec: 582823.48
4096 plaintext Requests/sec: 503303.61

1.0.0-M37 Netty
512 json Requests/sec: 435923.12
1024 plaintext Requests/sec: 581189.77
4096 plaintext Requests/sec: 507361.20
```
In this case, plaintext gets its best result on 1024 concurrent connections, but I'm including the result for 4096 for reference. Those results may hint at this PR making things slightly slower, but bigger numbers in general gives a bigger margin of error, so I'm not confident in calling a winner here, either.
Now, let's take a moment and appreciate how far we've come from 0.23, and the incredible raw potential of http4s-netty-server.

I think the next step would be to create a benchmark that uses more headers, by which I mean a realistic amount, as might be sent from a web browser or an API gateway, perhaps something like 10-30. The benchmarks above use only 2-3, so it's not too surprising that this PR doesn't make a big difference. It would also be interesting to see if combining this with #6718 changes the results in a significant way.

On the other hand, before reverting the change to use prepend for headers, this PR *did* make a big difference, so it might also be interesting to look at what bringing that back could do. If prepend was used, it would also allow for better optimizations if we had a `Set` of header names already added, as in those case we can completely avoid looping through the current list of headers (we just prepend the new one if its header name is not in that `Set`).

The current conclusion with regards to this PR, in my opinion, is that although it doesn't make performance *worse*, it doesn't make it demonstrably *better* either, so it's not worth the added complexity as it stands right now.

armanbilge:
> The current conclusion with regards to this PR, in my opinion, is that although it doesn't make performance worse, it doesn't make it demonstrably better either, so it's not worth the added complexity as it stands right now.

I appreciate the straightforward take. Basically we are going to need more drastic changes which have controversial compatibility implications 😕

wjoel:
Closing this as the proposed optimization didn't do much, if anything. More radical changes should be considered for 1.0.0.

PR #6206, opened 2022-03-30
Should Metrics classifierF use Generic vs String?

Ophirr33:
This is a proof of concept for whether or not the `Metrics` middleware should use a generic type instead of a `String` for `classifierF`. When implementing a `MetricsOps`, I found myself wanting to pass along more information than is easily included in a string. My specific use case is reporting metrics to datadog, and wanting to synthesize both a metric name and a set of metric tags from the request path, e.g.,

```scala
case class Context(name: String, tags: Seq[(String, String)])
```

I was able to solve this by converting a `Request[F]` into a `Context` and then serializing to json within classifierF and deserializing within my `MetricsOps` implementation, but I had the thought that this round trip could be avoided if the classifier worked with a type besides just `String`.

This PR is mostly a proof-of-concept for how this could look, happy to close this and move into an issue if that would be better for discussion.

### DIFF
--- a/client/shared/src/main/scala/org/http4s/client/middleware/Metrics.scala
+++ b/client/shared/src/main/scala/org/http4s/client/middleware/Metrics.scala
@@ -51,9 +51,9 @@ object Metrics {
     * @param client the [[Client]] to gather metrics from
     * @return the metrics middleware wrapping the [[Client]]
     */
-  def apply[F[_]](
-      ops: MetricsOps[F],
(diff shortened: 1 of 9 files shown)


Ophirr33:
@rossabaker I see that you've worked on this middleware a fair bit, any thoughts on updating the classifier?

brianmowen:
Do you have any thoughts on this @ChristopherDavenport?

PR #6208, opened 2022-03-30
Tweak `async` usage in `Http1Connection`

danicheg:
An itsy-bitsy tweaking.

### DIFF
--- a/blaze-client/src/main/scala/org/http4s/blaze/client/Http1Connection.scala
+++ b/blaze-client/src/main/scala/org/http4s/blaze/client/Http1Connection.scala
@@ -261,19 +261,8 @@ private final class Http1Connection[F[_]](
       idleRead: Option[Future[ByteBuffer]],
   ): F[Response[F]] =
     F.async_[Response[F]] { cb =>
-      idleRead match {
-        case Some(read) =>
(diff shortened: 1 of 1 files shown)


blaze-client/src/main/scala/org/http4s/blaze/client/Http1Connection.scala
```diff
@@ -260,22 +260,19 @@ private final class Http1Connection[F[_]](
...
+        case None =>
+          handleRead(
+            channelRead(),
+            cb,
+            closeOnFinish,
+            doesntHaveBody,
+            "Initial Read",
+            idleTimeoutS,
+          )
       }
```
diesalbla, line 276:
```suggestion
    F.async_[Response[F]] { cb =>
      val rr = idleRead.getOrElse(channelRead())
      handleRead(rr, cb, closeOnFinish, doesntHaveBody, "InitialRead", idleTimeoutS)
```

rossabaker (approved):
:+1:, and @diesalbla's refactor also looks right

blaze-client/src/main/scala/org/http4s/blaze/client/Http1Connection.scala
danicheg: Good one, thanks!

PR #6212, opened 2022-04-01
Use separate `Dispatcher` instances in the blaze client builder

danicheg:
[It's proposed by CE](https://github.com/typelevel/cats-effect/blob/27307cfe6cd9ea3cb230a8f9be5e16f624a914cd/std/shared/src/main/scala/cats/effect/std/Dispatcher.scala#L40-L42) to use separate instances of Dispatcher when possible to reduce contentions on submitting effects.

### DIFF
--- a/blaze-client/src/main/scala/org/http4s/blaze/client/BlazeClientBuilder.scala
+++ b/blaze-client/src/main/scala/org/http4s/blaze/client/BlazeClientBuilder.scala
@@ -408,7 +408,6 @@ sealed abstract class BlazeClientBuilder[F[_]] private (
             sslContextOption = sslContext,
             bufferSize = bufferSize,
             asynchronousChannelGroup = asynchronousChannelGroup,
-            executionContextConfig = executionContextConfig,
             scheduler = scheduler,
             checkEndpointIdentification = checkEndpointIdentification,
             maxResponseLineSize = maxResponseLineSize,
@@ -422,7 +421,7 @@ sealed abstract class BlazeClientBuilder[F[_]] private (
             dispatcher = dispatcher,
             idleTimeout = idleTimeout,
             getAddress = customDnsResolver.getOrElse(BlazeClientBuilder.getAddress(_)),
-          ).makeClient(requestKey)
+          ).makeClient(requestKey, executionContext)
 
     for {
       dispatcher <- Dispatcher[F]
--- a/blaze-client/src/main/scala/org/http4s/blaze/client/BlazeClientBuilder.scala
+++ b/blaze-client/src/main/scala/org/http4s/blaze/client/BlazeClientBuilder.scala
@@ -341,8 +341,8 @@ sealed abstract class BlazeClientBuilder[F[_]] private (
       scheduler <- scheduler
       _ <- Resource.eval(verifyAllTimeoutsAccuracy(scheduler))
       _ <- Resource.eval(verifyTimeoutRelations())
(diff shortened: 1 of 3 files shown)


rossabaker:
I'm uncertain where that advice begins and ends.  This separates the connection pool manager from the rest of the client, but only moves us from one to two per client, no?  If more is better, is it excessive to create one per request?

armanbilge:
BTW @wjoel recently made some big strides for shared dispatchers in:
- https://github.com/typelevel/cats-effect/pull/2876

danicheg:
> I'm uncertain where that advice begins and ends. This separates the connection pool manager from the rest of the client, but only moves us from one to two per client, no? If more is better, is it excessive to create one per request?

I'm confident creating `Dispatcher` per request has no benefits because that creation isn't free. To be honest, I haven't numbers to speak for sure. I just intended to separate the running effects of the pool manager and the rest of the client, because, following the CE spec, it could reduce contention. I guess that contention strikes only within the high load. But creating two instances of Dispatcher in the 'wiring' of the client seems cheap, so that was my point.

danicheg:
@armanbilge Good to know, thanks. Is that mean the advice of creating `Dispatcher` as often as possible has been gone?

armanbilge:
Good question. I don't know if that advice is "gone" but maybe less important after those changes. I think contention is always an issue for any sort of concurrent state. That PR just improves the performance under contention, so it's less of a big deal. But would be good to get some expert opinion on that too :)

I think the only definitive answer is benchmarks really :)

rossabaker:
I don't have the automated benchmarking that I used to.  I could easily be persuaded either way on this one, but I could be most persuaded by evidence.

wjoel:
It seems to only be used for `dispatcher.unsafeRunAndForget(closed.complete(()))` in `Http1Stage`, so this may not matter in this particular case, but note that changing the "level" of where the dispatcher is created will subtly change what happens to such running but "forgotten" actions when cancellations occur. The actions will be canceled when the dispatcher is closed, because the supervisor in the dispatcher has not forgotten them and will cancel them.
https://github.com/typelevel/cats-effect/issues/1881

However, I don't think the comment is accurate anymore, to be honest. This is what `Dispatcher` looked like when it was added: https://github.com/typelevel/cats-effect/blob/4201c8e3ceca811645730cd8ccd81b16f3402da0/std/shared/src/main/scala/cats/effect/std/Dispatcher.scala

Even then I find the "minimal overhead" a bit questionable in terms of resource usage, but it's true that the overhead in terms of fibers was just one. However, since https://github.com/typelevel/cats-effect/pull/1760 and https://github.com/typelevel/cats-effect/pull/2876 the contention just isn't an issue, but on the other hand the dispatcher uses more resources to create its data structures and is now even less cheap in that way.

I don't believe it's still good advice, more than two years later.

armanbilge:
> It seems to only be used for `dispatcher.unsafeRunAndForget(closed.complete(()))` in `Http1Stage`, so this may not matter in this particular case

Ah. Actually this might be a good candidate for the new sequential `Dispatcher` being introduced in https://github.com/typelevel/cats-effect/pull/2854.

armanbilge:
@danicheg do you want to move forward with this? Trying to cleanup PRs before schisming blaze.

danicheg:
I think we had a good discussion here and in https://github.com/typelevel/cats-effect/issues/2933. So, I close this.

PR #6230, opened 2022-04-03
Reduce use of `covary` method.

diesalbla:
After #5712 the `covary` method is less necessary. In some places we can just remove it. In the case of calls to `.pure[F]`, it happens that the syntax extensions fixes the type of the receiver object, so we need to use `F.pure` instead.


danicheg (approved):

rossabaker:
:+1:, but we got a conflict.

PR #6237, opened 2022-04-04
Publish testing module internally

rossabaker:
Publish the testing module internally so spun off modules that depend on it can compile.  No binary compatibility will be guaranteed.


rossabaker:
Having sketched this, I'm not confident:
1. It belongs in this repo at all.  It's a totally different release cycle.
2. It's a dependency worth taking on.  It doesn't really do a lot!

armanbilge:
Another thing that's been on my mind and particularly with the client/(server?) testkits is whether we can distribute something useful without enforcing munit. Like munit is a very reasonable choice nowadays, and it might not be worth the effort, but it could be nice if it wasn't completely baked in.

rossabaker:
I think the testkit aspect is much more compelling than this testing module.  I'm rebuilding the servlet to not need this at all, and it wasn't difficult.

rossabaker:
Looks like this failed anyway.  Let's see how far we can get without it.

PR #6238, opened 2022-04-04
Remove war example

rossabaker:
1. It bitrotted
2. A working version is in now in http4s-servlet


armanbilge (approved):
Bye bye!

rossabaker:
Eek.  One build hung.

armanbilge:
Seems it's not the first.
https://github.com/http4s/http4s/actions/runs/2086187444
https://github.com/http4s/http4s/actions/runs/2086187444

... looking back, first one seems to be
https://github.com/http4s/http4s/actions/runs/2081893375

This one might be on me, sorry. I wonder if adding some `lazy val`s caused some initialization locks somehow?

rossabaker:
Maybe.  There's a description of a few things that can go wrong [here](https://docs.scala-lang.org/sips/improved-lazy-val-initialization.html.

armanbilge:
Ok, gotcha.

https://gist.github.com/armanbilge/a1759c829bb579484a9e1542dc0443c1

PR #6246, opened 2022-04-05
Server Metrics Middleware

diesalbla:
- Extract local methods for main operations: starting the metric "timer", ending the metrics "timer". 
- Rename `MetricsRequestContext` to `MetricsEntry`, because they are not directly related to requests, and "context" is also used with "ContextRequest" and "ContextResponse", which can add confusion.


rossabaker:
This has MiMa issues.

diesalbla:
> This has MiMa issues.

Fixed.

rossabaker (approved):

Issue #6242, opened 2022-04-04
MimeDB-initialization deadlock


armanbilge:
This has been plaguing our builds since https://github.com/http4s/http4s/pull/6211, but Ross noticed it first in https://github.com/http4s/http4s/pull/6238#issuecomment-1088035957.

I was able to get a `jstack` dump that reveals the deadlock relating to `lazy val` intialization. Would appreciate some help solving it :)

https://gist.github.com/armanbilge/a1759c829bb579484a9e1542dc0443c1

armanbilge:
In https://github.com/http4s/http4s/pull/6211 I made a `val` into a `lazy val`. One fix would be to split that file across JVM/JS, revert the change on JVM and keep it on JS.

Note that previously all these `lazy val`s were kind of pointless on JVM: because of that `val` which I changed, the entire MimeDB was being initialized at startup. This may or may not have been a big deal.

So the change in https://github.com/http4s/http4s/pull/6211 was kind of a fix for that, except now since things are properly lazily initialized we get into deadlock territory.

armanbilge:
Zeroing in, this deadlock seems to occur when one thread access `MediaType.application.json` while another accesses `MediaType.all`. Dare I point out, this was a pre-existing problem that we weren't observing since the MimeDB was being forced to initialize in its entirety when loading first loading the `MediaType` class anyway?

```
"io-compute-2":
	at org.http4s.MediaType$.application_parts$lzycompute$1(MediaType.scala:252)
	- waiting to lock <0x0000000754d4de80> (a org.http4s.MediaType$)
	at org.http4s.MediaType$.application_parts(MediaType.scala:252)
	at org.http4s.MimeDB$application$.org$http4s$MimeDB$application_parts$application_0$$$outer(MimeDB.scala:6207)
	at org.http4s.MimeDB$application_parts$application_0.json(MimeDB.scala:376)
	at org.http4s.MimeDB$application_parts$application_0.json$(MimeDB.scala:375)
	at org.http4s.MimeDB$application$.json$lzycompute(MimeDB.scala:6207)
	- locked <0x000000075400e878> (a org.http4s.MimeDB$application$)
	at org.http4s.MimeDB$application$.json(MimeDB.scala:6207)
```

```
"io-compute-4":
	at org.http4s.MimeDB$application$.all$lzycompute(MimeDB.scala:6212)
	- waiting to lock <0x000000075400e878> (a org.http4s.MimeDB$application$)
[...]

PR #6248, opened 2022-04-05
Use thread unsafe "lazy val"s to avoid deadlocks initializing MimeDB

armanbilge:
Hopefully fixes https://github.com/http4s/http4s/issues/6242.

This uses the null-init pattern suggested by Ross in https://github.com/http4s/http4s/issues/6242#issuecomment-1088187008.

IIUC we actually need to use this pattern for any of the lazily-initialized collections, since attempting to initialize one of those concurrently with a `MediaType` within that collection creates a deadlock-prone situation. Unfortunately said collections are all public, so albeit unlikely it could happen to any of them.


armanbilge:
Oh yeah, I've be rerunning MimeLoader loads :) that's why I made all that fuss in the last PR about reproducibility

https://github.com/http4s/http4s/blob/17bd355936ab45bb0ab9d8c82fe4407ab6c43cd6/project/MimeLoader.scala#L52-L53

rossabaker:
Oh, right, I forgot we pinned it.

PR #6249, opened 2022-04-05
Remove synchronizations in `TestHead`, `QueueTestHead`, `SlowTestHead`

danicheg:
These synchronizations seem redundant. Unfortunately, we didn't achieve any significant performance boost. Nevertheless, it could be considered just like refactoring.


rossabaker (approved):
I'm a little squeamish removing synchronization, but it seems right, it's just tests, and they still passed.  Let's just keep this one in mind if we see an uptick in flakes.

PR #6250, opened 2022-04-05
Merge 0.23 -> main

armanbilge:
Seems main was using a different MimeDB version than 0.23. Rather than try and hunt down the specific version, I just bumped to the latest and re-generated.


rossabaker:
Yeah, main gets bumped periodically because it's where we can break bincompat.

rossabaker (approved):

PR #6252, opened 2022-04-06
Use CE `Random.javaSecuritySecureRandom` instead of Java `SecureRandom`

armanbilge:
The Java `SecureRandom` blocks. The JS `SecureRandom` was first shimmed in http4s-crypto, then in CE.

CE's `Random.javaSecuritySecureRandom` doesn't solve the blocking problem yet, but hopefully it will soon. In any case, this change lets us iterate in CE and reap the benefits everywhere.

### DIFF
--- a/server/jvm/src/main/scala/org/http4s/server/middleware/CSRF.scala
+++ b/server/jvm/src/main/scala/org/http4s/server/middleware/CSRF.scala
@@ -106,7 +106,7 @@ final class CSRF[F[_], G[_]] private[middleware] (
 
   /** Generate a new token */
   def generateToken[M[_]](implicit F: Sync[M]): M[CSRFToken] =
-    CSRF.genTokenString.to[M].flatMap(signToken[M])
+    CSRF.genTokenString[M].flatMap(signToken[M])
 
   /** Create a Response cookie from a signed CSRF token
     *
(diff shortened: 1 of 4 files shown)


server/jvm/src/main/scala/org/http4s/server/middleware/CSRF.scala
```diff
@@ -526,11 +526,8 @@ object CSRF {
     )
 
   /** Generate an unsigned CSRF token from a `SecureRandom` */
-  private[middleware] def genTokenString: String = {
-    val bytes = new Array[Byte](CSRFTokenLength)
-    CachedRandom.nextBytes(bytes)
-    encodeHexString(bytes)
-  }
+  private[middleware] def genTokenString: SyncIO[String] =
```
danicheg, line 529: I wonder, can we move to `F[_]: Sync` instead of `SyncIO`? It seems `genTokenString` is used only when `Sync` is in-game.

danicheg (approved):
Lovely, despite my safely skippable nitpick.

server/jvm/src/main/scala/org/http4s/server/middleware/CSRF.scala
armanbilge: Ah, yes I think so :) let's do that 👍

armanbilge:
The JS builds are getting flaky because of the artifact size test I added :(

PR #6253, opened 2022-04-06
Relax `authentication#challenged` constraints

danicheg:


armanbilge (approved):
Nice!

server/shared/src/main/scala/org/http4s/server/middleware/authentication/package.scala
```diff
@@ -18,16 +18,15 @@ package org.http4s
...
 import cats.data.OptionT
-import cats.effect.Sync
 import org.http4s.headers._
 
 package object authentication {
-  // TODO Could be reduced to a Monad[F]
   def challenged[F[_], A](
       challenge: Kleisli[F, Request[F], Either[Challenge, AuthedRequest[F, A]]]
-  )(routes: AuthedRoutes[A, F])(implicit F: Sync[F]): HttpRoutes[F] =
+  )(routes: AuthedRoutes[A, F])(implicit F: Monad[F]): HttpRoutes[F] =
```
satorg, line 29: I'm wondering – is it binary compatible? I remember that once I attempted to do a similar thing in Cats, but that made Mima unhappy so I had to add some tricks to calm Mima down.

armanbilge: It's not, but this is the development branch so anything goes :)

satorg: Ok got it, cool then, thank you.

satorg (approved):

PR #6254, opened 2022-04-06
Fix counting of current allocated connections in the blaze client pool manager

danicheg:
`-=`/`+=` operators aren't thread-safe. So this means we almost certainly* have issues in counting current allocated connections.

\* I haven't proof, just speaking from common sense.

### DIFF
--- a/blaze-client/src/main/scala/org/http4s/blaze/client/PoolManager.scala
+++ b/blaze-client/src/main/scala/org/http4s/blaze/client/PoolManager.scala
@@ -26,6 +26,7 @@ import org.http4s.internal.CollectionCompat
 import org.log4s.getLogger
 
 import java.time.Instant
+import java.util.concurrent.atomic.AtomicInteger
 import scala.collection.mutable
 import scala.concurrent.ExecutionContext
 import scala.concurrent.duration._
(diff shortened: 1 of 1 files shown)


blaze-client/src/main/scala/org/http4s/blaze/client/PoolManager.scala
```diff
@@ -109,13 +110,13 @@ private final class PoolManager[F[_], A <: Connection[F]](
 
   private def incrConnection(key: RequestKey): F[Unit] =
     F.delay {
-      curTotal += 1
+      curTotal.getAndIncrement()
       allocated.update(key, allocated.getOrElse(key, 0) + 1)
```
satorg, line 114: I would suppose that `allocated.update(key, allocated.getOrElse(key, 0) + 1)` is also not thread safe, is it?
Because it is just
```
private val allocated = mutable.Map.empty[RequestKey, Int]
```
Shouldn't `scala.collection.concurrent.TrieMap` be used here instead?

danicheg: Probably you're right, but I feel that changing the collection for sure will need some performance checking. I'd rather leave this PR with only the proposed change. Feel free to open a new one, or I may work on it after a while.

satorg (approved):
LGTM – I cannot find anything wrong with the suggested changes.

blaze-client/src/main/scala/org/http4s/blaze/client/PoolManager.scala
rossabaker: Yes, that's certainly incorrect.

rossabaker: The semaphore should protect us from concurrent updates, but I think cached reads are still a problem.

PR #6255, opened 2022-04-06
Move JS size-test-app to Test scope, fix flaky CI OOMs?

armanbilge:
Started noticing some flaky CI and OOMs for JS. My experience is doing the test _linking_ and test _running_ in parallel can cause this to happen.

Our CI already links everything in `Test` scope as a separate step, precisely to avoid this issue. So, by moving this size-test-app into `Test` scope it will get linked as part of that step as well.


danicheg (approved):
I believe you know what you do.

PR #6257, opened 2022-04-07
Message - remoteHost: define without OptionT.

diesalbla:
Takes a few more lines, but should be evident.

It could be an Option `fold`, but that would need type arguments


rossabaker (approved):
Yeah, this one's cleaner.

PR #6258, opened 2022-04-07
Server - Chunk Aggregator Middleware: code and docs

diesalbla:
Add a Scaladoc section to the object, with a link to the RFC standard
and a description of functionality.

Also, extract the post-processing logic to a private `aggregate`
function to leave `apply` as a mere `flatMapF`.


diesalbla:
This is the error of the build. I do not think it is related.
```
[error] Failed tests:
[error] 	org.http4s.blaze.client.ClientTimeoutSuite
```

rossabaker (approved):
I could take or leave the refactor, but the scaladoc is a great addition.

PR #6259, opened 2022-04-07
Avoid Alternative-Guard

diesalbla:
The `guard` method of the `Alternative` typeclass is less known that the Scala `if-then-else`

### DIFF
--- a/server/shared/src/main/scala/org/http4s/server/middleware/Throttle.scala
+++ b/server/shared/src/main/scala/org/http4s/server/middleware/Throttle.scala
@@ -74,11 +74,12 @@ object Throttle {
                   val attemptSet: F[Option[TokenAvailability]] =
                     if (newTokenTotal >= 1)
                       setter((newTokenTotal - 1, currentTime))
-                        .map(_.guard[Option].as(TokenAvailable))
+                        .map(if (_) Some(TokenAvailable) else None)
                     else {
                       val timeToNextToken = refillEvery.toNanos - timeDifference
                       val successResponse = TokenUnavailable(timeToNextToken.nanos.some)
-                      setter((newTokenTotal, currentTime)).map(_.guard[Option].as(successResponse))
+                      setter((newTokenTotal, currentTime))
+                        .map(if (_) Some(successResponse) else None)
                     }
 
                   attemptSet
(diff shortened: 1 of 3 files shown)


server/shared/src/main/scala/org/http4s/server/middleware/Throttle.scala
```diff
@@ -74,11 +74,12 @@ object Throttle {
                   val attemptSet: F[Option[TokenAvailability]] =
                     if (newTokenTotal >= 1)
                       setter((newTokenTotal - 1, currentTime))
-                        .map(_.guard[Option].as(TokenAvailable))
+                        .map(if (_) Some(TokenAvailable) else None)
```
danicheg, line 77:
```suggestion
                        .map(Option.when(_)(TokenAvailable))
```

server/shared/src/main/scala/org/http4s/server/middleware/Throttle.scala
```diff
@@ -74,11 +74,12 @@ object Throttle {
...
                     else {
                       val timeToNextToken = refillEvery.toNanos - timeDifference
                       val successResponse = TokenUnavailable(timeToNextToken.nanos.some)
-                      setter((newTokenTotal, currentTime)).map(_.guard[Option].as(successResponse))
+                      setter((newTokenTotal, currentTime))
+                        .map(if (_) Some(successResponse) else None)
```
danicheg, line 82:
```suggestion
                        .map(Option.when(_)(successResponse))
```

danicheg (commented):
A few suggestions:

server/shared/src/main/scala/org/http4s/server/middleware/Throttle.scala
```diff
@@ -74,11 +74,12 @@ object Throttle {
                   val attemptSet: F[Option[TokenAvailability]] =
                     if (newTokenTotal >= 1)
                       setter((newTokenTotal - 1, currentTime))
-                        .map(_.guard[Option].as(TokenAvailable))
+                        .map(Option.when(_)(TokenAvailable))
```
satorg, line 77: Yeah, I like `Option.when` and `Option.unless` too. Unfortunately they don't exist for Scala 2.15.

diesalbla: Oh, that's a shame... Well, removing...

server/shared/src/main/scala/org/http4s/server/middleware/Throttle.scala
diesalbla: It was nice, but not for 2.12

server/shared/src/main/scala/org/http4s/server/middleware/Throttle.scala
danicheg: Darn. Sorry for interrupting.

danicheg (approved):

PR #6266, opened 2022-04-08
Fix unsafe concurrency in allocated map in blaze `PoolManager`

### LINKED ISSUE #6261: Unsafe concurrency in allocated map in blaze PoolManager
Noted by @satorg in https://github.com/http4s/http4s/pull/6254#discussion_r844401735

danicheg:
FIxes #6261

### DIFF
--- a/blaze-client/src/main/scala/org/http4s/blaze/client/PoolManager.scala
+++ b/blaze-client/src/main/scala/org/http4s/blaze/client/PoolManager.scala
@@ -23,10 +23,13 @@ import cats.effect.concurrent.Semaphore
 import cats.syntax.all._
 import org.http4s.client.RequestKey
 import org.http4s.internal.CollectionCompat
+import org.http4s.internal.CollectionCompat.CollectionConverters._
 import org.log4s.getLogger
 
 import java.time.Instant
+import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.atomic.AtomicInteger
+import java.util.function.BiFunction
 import scala.collection.mutable
 import scala.concurrent.ExecutionContext
 import scala.concurrent.duration._
@@ -90,7 +93,7 @@ private final class PoolManager[F[_], A <: Connection[F]](
 
   private var isClosed = false
   private val curTotal = new AtomicInteger(0)
-  private val allocated = mutable.Map.empty[RequestKey, Int]
+  private val allocated = new ConcurrentHashMap[RequestKey, Int]()
   private val idleQueues = mutable.Map.empty[RequestKey, mutable.Queue[PooledConnection]]
   private var waitQueue = mutable.Queue.empty[Waiting]
 
@@ -111,24 +114,30 @@ private final class PoolManager[F[_], A <: Connection[F]](
   private def incrConnection(key: RequestKey): F[Unit] =
     F.delay {
       curTotal.getAndIncrement()
-      allocated.update(key, allocated.getOrElse(key, 0) + 1)
+      val update: BiFunction[RequestKey, Int, Int] = (_: RequestKey, v: Int) => v + 1
+      allocated.compute(key, update)
+      ()
     }
 
   private def decrConnection(key: RequestKey): F[Unit] =
     F.delay {
       curTotal.getAndDecrement()
-      val numConnections = allocated.getOrElse(key, 0)
+      val numConnections = allocated.getOrDefault(key, 0)
       // If there are no more connections drop the key
       if (numConnections == 1) {
         allocated.remove(key)
         idleQueues.remove(key)
         ()
-      } else
-        allocated.update(key, numConnections - 1)
+      } else {
+        allocated.put(key, numConnections - 1)
+        ()
+      }
     }
 
   private def numConnectionsCheckHolds(key: RequestKey): Boolean =
-    curTotal.intValue() < maxTotal && allocated.getOrElse(key, 0) < maxConnectionsPerRequestKey(key)
+    curTotal.intValue() < maxTotal && allocated.getOrDefault(key, 0) < maxConnectionsPerRequestKey(
+      key
+    )
 
   private def isRequestExpired(t: Instant): Boolean = {
     val elapsed = Instant.now().toEpochMilli - t.toEpochMilli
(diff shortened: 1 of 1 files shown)


blaze-client/src/main/scala/org/http4s/blaze/client/PoolManager.scala
```diff
@@ -111,24 +114,30 @@ private final class PoolManager[F[_], A <: Connection[F]](
   private def incrConnection(key: RequestKey): F[Unit] =
     F.delay {
       curTotal.getAndIncrement()
-      allocated.update(key, allocated.getOrElse(key, 0) + 1)
+      val update: BiFunction[RequestKey, Int, Int] = (_: RequestKey, v: Int) => v + 1
+      allocated.compute(key, update)
```
danicheg, line 118: I find this fascinating - the default behavior of `ConcurrentHashMap` on getting values on empty keys is about returning NPE. But if we try it in Scala with `Int` as a type of values, the '0' value will be returned then. So we don't need to check a possible NPE here.

blaze-client/src/main/scala/org/http4s/blaze/client/PoolManager.scala
```diff
@@ -90,7 +93,7 @@ private final class PoolManager[F[_], A <: Connection[F]](
 
   private var isClosed = false
   private val curTotal = new AtomicInteger(0)
-  private val allocated = mutable.Map.empty[RequestKey, Int]
+  private val allocated = new ConcurrentHashMap[RequestKey, Int]()
```
danicheg, line 96: Initially, @satorg proposed to use `TrieMap`, but I think `ConcurrentHashMap` will be fine here. Besides the fact, `ConcurrentHashMap` [has a bit more performance](https://stackoverflow.com/a/38700133).

blaze-client/src/main/scala/org/http4s/blaze/client/PoolManager.scala
danicheg: If one is interested in why this is so: https://github.com/scala/bug/issues/602#issuecomment-292349468

blaze-client/src/main/scala/org/http4s/blaze/client/PoolManager.scala
satorg: Personally I am totally fine with the Java's `ConcurrentHashMap` too :)

blaze-client/src/main/scala/org/http4s/blaze/client/PoolManager.scala
```diff
@@ -90,7 +93,7 @@ private final class PoolManager[F[_], A <: Connection[F]](
...
   private val idleQueues = mutable.Map.empty[RequestKey, mutable.Queue[PooledConnection]]
   private var waitQueue = mutable.Queue.empty[Waiting]
```
satorg, line 98: I mean, if we're worried about thread safety of this code in general, then I'd guess we should protect these two guys too.

blaze-client/src/main/scala/org/http4s/blaze/client/PoolManager.scala
```diff
@@ -111,24 +114,30 @@ private final class PoolManager[F[_], A <: Connection[F]](
...
       if (numConnections == 1) {
         allocated.remove(key)
         idleQueues.remove(key)
         ()
-      } else
-        allocated.update(key, numConnections - 1)
+      } else {
+        allocated.put(key, numConnections - 1)
+        ()
+      }
```
satorg, line 134: If it is possible that this piece of code can be accessed from multiple threads simultaneously, then (I guess) it is not going to be thread-safe even if change each collection to a thread-safe version.

Consider the scenario:
1. Two threads (A and B) get to line 125 and both receive `numConnections == 2` for the same key X.
2. Consequently they both pass to line 132 directly and each of them decrease numConnection by 1 therefore it becomes 0 (skipping 1)
3. The key X never gets removed.

Wdyt?

danicheg:
@RafalSumislawski I'm not so sure, because every thread may have its copy of `curTotal` and `allocated`. Especially the `curTotal` wasn't volatile. And `scala.collection.mutable.HashMap` isn't seem the best choice for multithreaded usage.

danicheg:
@RafalSumislawski By the way, it's good to catch you in this PR. I wanna ask you, as one of the investigators of the blaze, why did we use a single `Semaphore` for all connections? Why not one `Semaphore` per connection?

RafalSumislawski:
> I'm not so sure, because every thread may have its copy of curTotal and allocated. Especially the curTotal wasn't volatile. And scala.collection.mutable.HashMap isn't seem the best choice for multithreaded usage.

We should either discuss this on JMM level, or CPU level. Threads don't have copies of memory. 

If we go the CPU way then:
- the semaphore let's us ensure that at most a single core is processing the state of the `PoolManager` at given time. 
- release followed by aquisition of the semaphore, on hardware level, result in execution of memory bariers, which though the means of cache coherence protocols ensure that the core that will now process the state of the `PoolManager` will have it's cache in a state consistent with the one left behind by the core that last processed the state

I'm simplyfing here a bit, by assuming that the entirety of the `semaphore.withPermit` is executed on one core, but even if it isn't this will work because shifting fibers to different threads also results in memory bariers on hadware level. Or to put it differently: cats-effect works. This will brake if we introduce some parallelism inside `semaphore.withPermit`, so naturally we don't do it.

If we go the JMM way:
modifications of `curTotal`/`allocated`/... -happens_before-> release of semaphore -happens_before-> aquisition of semaphore -> hapens before read access to `curTotal`/`allocated`/... 
All `volatile` does is create additinal happens-before, which isn't really needed because we already have enough. All synchronisation in ConcurrenHashMap does is add some more mutual exclusion and happens-before which we don't need.

It's concurrency, with some nontrivial code inside, so prooving in detail that the code is correct is hard. But this code uses a simple old pattern: aquire a lock, do stuff, release the lock. I have no doubts about soundness of that pattern.

> why did we use a single Semaphore for all connections? 

The `Semaphore` is there to synchronise access to the mutable state of `PoolManager` that is the: `isClosed`, `curTotal`, `allocated`, `idleQueues`, `waitQueue`. 

> Why not one Semaphore per connection? And maybe it's possible to rewrite PoolManager to get rid of Semaphore at all then?

It's not clear to me what a `Semaphore` per connection would do. It probably wouldn't be needed, as you would need to synchronise access to the state of the `PoolManager` in some other way. It's possible to replace the `Semaphore` with som other synchronisation mechanism. The question is: what would be the goal of such rewrite?

danicheg:
> It's not clear to me what a Semaphore per connection would do. It probably wouldn't be needed, as you would need to synchronise access to the state of the PoolManager in some other way

If the main goal of `PoolManager` is to give a connection to a request (and reuse connections), then it's odd why different requests are synchronized. The question is can we remove synchronization (or decrease its quantity) and handle state in a non-blocking way.

PR #6267, opened 2022-04-09
Throttle Server middleware: use recursion

diesalbla:
The `takeToken` method uses the Ref.access method, which yields a setter. The setter yields `true` if it can set the value, or `false` if it could not, perhaps because of a concurrent write. In the latter case, the method has to re-try.

Currently, that recursion goes through emitting an Option and `flatMap`. This commit changes that to a direct recursive call.


rossabaker (approved):
Nice cleanup.

Issue #6239, opened 2022-04-04
Publish http4s scalafix rules


rossabaker:
We're losing the scalafix rules as we split the repo.  We need to figure out how to refer to them here and get them into module.g8, or publish them as a jar.

armanbilge:
Right. Then ideally we get scalafix integrated into sbt-typelevel.
- https://github.com/typelevel/sbt-typelevel/issues/236

armanbilge:
I can work on this one.

PR #6268, opened 2022-04-10
Publish internal scalafixes

armanbilge:
Testing out some ideas for https://github.com/http4s/http4s/issues/6239.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1307,6 +1307,7 @@ lazy val scalafixInternalRules = project
   .disablePlugins(ScalafixPlugin)
   .settings(
     name := "http4s-scalafix-internal",
+    mimaPreviousArtifacts := Set.empty,
     startYear := Some(2021),
     libraryDependencies ++= Seq(
       "ch.epfl.scala" %% "scalafix-core" % _root_.scalafix.sbt.BuildInfo.scalafixVersion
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -1303,9 +1304,9 @@ lazy val examplesTomcat = exampleProject("examples-tomcat")
 
 lazy val scalafixInternalRules = project
   .in(file("scalafix-internal/rules"))
-  .enablePlugins(NoPublishPlugin)
   .disablePlugins(ScalafixPlugin)
   .settings(
+    name := "http4s-scalafix-internal",
```
armanbilge, line 1309: Hoping that it's as simple as publishing them. We shouldn't guarantee bincompat though. Bikeshed the name.

armanbilge:
Ok, gave this a try in https://github.com/http4s/http4s-dom/pull/131 and seems to work!

https://github.com/http4s/http4s-dom/runs/5958750538?check_suite_focus=true#step:11:37

Also :100: to @bplommer for building us such nice scalafixes 🥰 the long-term goal should definitely be to upstream some of these.

Unsurprisingly most of the yak falls to integrating this into sbt-typelevel 🙃 the plan is for 0.5.0, which is on the horizon 😅 maybe we can try it out in sbt-http4s-org first ...

PR #6279, opened 2022-04-10
Fix deprecated since versions

danicheg:
`0.22.13` isn't released, so I think it's better to fix this.


armanbilge (approved):
Yeah, thanks. Sorry we're so bad at these ...

PR #6280, opened 2022-04-10
Tweak badges in the readme

danicheg:
Two proposals:
1. if the actual 'main' branch is `series/0.23` then we should refer to the right CI build status
2. let's promote the latest major release version


armanbilge:
WDYT about a scaladex badge instead of a maven central badge?

[http4s-core Scala version support](https://index.scala-lang.org/http4s/http4s/http4s-core)

danicheg:
@armanbilge I have no preference between badges 😆 Feel free to change it within that branch if you like that one more.

danicheg:
@armanbilge btw there is a difference between them. With the maven central one - the title is 'maven-central', with the scaladex one - the title is the name of the particular module 'http4s-core'.

armanbilge (approved):
Yeah, I think the Maven one is more configurable. Scaladex doesn't have the `?versionPrefix=0.23` option.

PR #6281, opened 2022-04-10
Kleisli Response Ops: add an "orElse" method.

diesalbla:
Although `orNotFound` is frequent and quick, `orElse` could also be useful?

### DIFF
--- a/core/shared/src/main/scala/org/http4s/syntax/KleisliSyntax.scala
+++ b/core/shared/src/main/scala/org/http4s/syntax/KleisliSyntax.scala
@@ -45,6 +45,11 @@ trait KleisliSyntax {
 final class KleisliResponseOps[F[_]: Functor, A](self: Kleisli[OptionT[F, *], A, Response[F]]) {
   def orNotFound: Kleisli[F, A, Response[F]] =
     Kleisli(a => self.run(a).getOrElse(Response.notFound))
+
+  /** Closes the `self` HttpRoutes into an HttpApp by giving the response
+    * to be given if the `self` routes yield `None`. */
+  def orElse(default: => Response[F]): HttpApp[F] =
+    Kleisli(a => self.run(a).getOrElse(response))
 }
 
 final class KleisliHttpRoutesOps[F[_]](self: HttpRoutes[F]) {


core/shared/src/main/scala/org/http4s/syntax/KleisliSyntax.scala
```diff
@@ -45,6 +45,11 @@ trait KleisliSyntax {
 final class KleisliResponseOps[F[_]: Functor, A](self: Kleisli[OptionT[F, *], A, Response[F]]) {
   def orNotFound: Kleisli[F, A, Response[F]] =
     Kleisli(a => self.run(a).getOrElse(Response.notFound))
+
+  /** Closes the `self` HttpRoutes into an HttpApp by giving the response
+    * to be given if the `self` routes yield `None`. */
+  def orElse(default: => Response[F]): HttpApp[F] =
+    Kleisli(a => self.run(a).getOrElse(response))
```
satorg, line 52: I think, it should be named `getOrElse`, because by a general convention `orElse` usually takes the same type as the one it is called on, e.g. (simplified):
```scala
class Option[A] {
  def orElse(alternative: => Option[A]): Option[A]
}
class OptionT[F[_], A] {
  def orElse(default: => OptionT[F, A]): OptionT[F, A]
}
class Either[A, B] {
  def orElse(or: => Either[A, B]): Either[A, B]
}
```
etc.

Therefore I suppose, `orElse` itself in this case should look like
```scala
def orElse(default: Kleisli[OptionT[F, *], A, Response[F]): Kleisli[OptionT[F, *], A, Response[F]
```
or something like that.
I.e. it is supposed to be designed for combining routes rather than returning a default response.

satorg: In fact, there's `orElse` in cats-core available already via `ApplicativeError` syntax available for both `Kleisli` and `OptionT`. Check this out:
```scala
import cats.syntax.all._

type Testee[F[_], A, B] = Kleisli[OptionT[F, *], A, B]

val testee1: Testee[List, String, Int] = Kleisli.liftF(OptionT.none)
val testee2: Testee[List, String, Int] = Kleisli.pure[OptionT[List, *], String, Int](456)

println(testee1.orElse(testee2).run("whatever"))
```

satorg: Also, the suggested syntax has nothing to deal with `Response` type itself. I mean, it can be even more generic, something like this:
```scala
implicit final class KleisliOptionTSyntax[F[_], A, B](val self: Kleisli[OptionT[F, *], A, B]) extends AnyVal {
  def getOrElse(fallback: B)(implicit F: Functor[F]): Kleisli[F, A, B] = {
    Kleisli(a => self.run(a).getOrElse(fallback))
  }
}

type Testee[F[_], A, B] = Kleisli[OptionT[F, *], A, B]

val testee1: Testee[List, String, Int] = Kleisli.liftF(OptionT.none)
println(testee1.getOrElse(456).run("whatever"))
```
So I'd suppose that such `getOrElse` semantic is not specific to http4s but rather could be promoted to `cats-core` (which could be useful IMO).

diesalbla: >  I.e. it is supposed to be designed for combining routes rather than returning a default response.

That would be a useful synonym for the `<+>` operator. Some Scala developers have reservations about using operators rather than words. 

> So I'd suppose that such getOrElse semantic is not specific to http4s but rather could be promoted to cats-core (which could be really useful there IMO).

Can leave that to others. In what other libraries are `OptionT` and `Kleisli` used as much as in Http4S?

satorg: Once it is added and released it would be way more difficult to remove:) Therefore it may lead to quite annoying conflicts in the future.

So personally I'm kinda hesitant about this change because (just to summarize):
1. The suggested `orElse` is not following the convention that is used for other `orElse` methods found in other places like standard library and cats-core. The `getOrElse` name would look more appropriate in this case.
2. Even though it is not dealing with http4s specific code somehow – therefore looks like more generic syntax to me.

rossabaker: > That would be a useful synonym for the `<+>` operator. Some Scala developers have reservations about using operators rather than words.

`<+>` is a synonym for `combineK`, which is equivalent to `orElse` for some `SemigroupK`, including `IO`.

> Once it is added and released it would be way more difficult to remove:) Therefore it may lead to quite annoying conflicts in the future.

:100: We've seen this pattern repeatedly over the years.  We want to be careful that our syntax is HTTP-specific.  I agree that this would be better in Cats.

PR #6282, opened 2022-04-10

danicheg:


core/shared/src/main/scala/org/http4s/Message.scala
```diff
@@ -232,6 +284,25 @@ sealed trait Message[+F[_]] extends Media[F] { self =>
...
+    * body.  Replaces any `Transfer-Encoding: chunked` with a
+    * `Content-Length` header.  It is the caller's responsibility to
+    * assure there is enough memory to materialize the body.
+    */
+  def toStrict[F1[x] >: F[x]](implicit F: Concurrent[F1]): F1[SelfF[F1]] =
+    entity match {
+      case Entity.Empty =>
+        F.pure(self.covary[F1])
+      case Entity.Strict(chunk) =>
+        F.pure(self.withContentLength(`Content-Length`.unsafeFromLong(chunk.size.toLong)))
```
armanbilge, line 298: 👍

core/shared/src/main/scala/org/http4s/Message.scala
```diff
@@ -232,6 +284,25 @@ sealed trait Message[+F[_]] extends Media[F] { self =>
 
   def mapK[F2[x] >: F[x], G[_]](f: F2 ~> G): SelfF[G] =
     self.change(entity = entity.translate(f))
+
+  /** Compiles the body stream to a single chunk and sets it as the
+    * body.  Replaces any `Transfer-Encoding: chunked` with a
+    * `Content-Length` header.  It is the caller's responsibility to
+    * assure there is enough memory to materialize the body.
+    */
+  def toStrict[F1[x] >: F[x]](implicit F: Concurrent[F1]): F1[SelfF[F1]] =
```
danicheg, line 293: This takes into account Ross's comment https://github.com/http4s/http4s/pull/6176#issuecomment-1075607483

core/shared/src/main/scala/org/http4s/multipart/Multiparts.scala
```diff
@@ -0,0 +1,59 @@
...
+object Multiparts {
+
+  /** Creates a `scala.util.Random` and provides Multiparts around it.
+    * This instance can be shared, or is cheap enough to create closer
+    * to where Multipart values are generated.
+    */
+  def forSync[F[_]](implicit F: Sync[F]): F[Multiparts[F]] =
+    F.delay(new Random()).map(fromScalaRandom[F])
+
+  def fromScalaRandom[F[_]](random: Random)(implicit F: Sync[F]): Multiparts[F] =
```
armanbilge, line 46: I think I forgot this on 0.23.x branch, we should switch to use `Random` typeclass here instead.

armanbilge (approved):

core/shared/src/main/scala/org/http4s/Message.scala
danicheg, line 298: One thing I'm not sure about - should we replace the `Content-Length` header here. I'm leaning that it's better to do.

core/shared/src/main/scala/org/http4s/multipart/Multiparts.scala
danicheg: Will you port this to the 0.23 branch? Or do you say that we should do it only for the main?

armanbilge: We should definitely fix it on the 0.23.x branch. I can do it a bit later 👍

core/shared/src/main/scala/org/http4s/Message.scala
```diff
@@ -204,6 +207,55 @@ sealed trait Message[+F[_]] extends Media[F] { self =>
...
+    * res1: Headers = Headers(Transfer-Encoding: gzip, Content-Type: text/plain, Content-Length: 1024)
+    *
+    * scala> val const = Request().withHeaders(
+    *      |   `Content-Length`(2048),
+    *      |   `Content-Type`(MediaType.text.plain))
+    * scala> const.withContentLength(`Content-Length`.unsafeFromLong(1024)).headers
+    * res1: Headers = Headers(Content-Type: text/plain, Content-Length: 1024)
+    * }}}
+    */
+  def withContentLength(contentLength: `Content-Length`): SelfF[F] =
```
bplommer, line 236: I don't think this fits well with the new entity model. It's the job of `Entity` to know about content length, not `Message` - anything in `Message` relating to content length should just be a thin wrapper around `Entity` - and it shouldn't be possible to set an incorrect length on a strict entity, so trying to do so should fail either at compile time or run time.

At the very least, I think this method should have `unsafe` or `unchecked` in its name.

Assuming the change in 0.23 hasn't been released yet, could we consider adjusting this in 0.23 to be more forward-compatible with 1.0's model?

Sorry to be late to the party on this 😬

core/shared/src/main/scala/org/http4s/Message.scala
bplommer: I think this could be `F[SelfF[Pure]]`.

core/shared/src/main/scala/org/http4s/Message.scala
armanbilge: Good point, that makes sense. Pretty sure this hasn't been released yet, I'm curious what you have in mind :)

bplommer: I don't have an immediate good idea. Does this need to be public, or could it be made package-private and released as-is, so we can think later on about what a public version should look like?

danicheg: > At the very least, I think this method should have unsafe or unchecked in its name.

I opine that prefixes 'unsafe' don't work. But yeah, it'd be cool if Scala has a `@unsafe` annotation to mark unsafe code and special compiler flags to configure behavior - compile unsafe code, emit a warning, or a compilation error.

bplommer: That could definitely be done as a scalafix or wartremover rule!

bplommer: I think in `main` what we should have is `def transformEntity(f: Entity[F] => Entity[F]): SelfF[F]` (might need an extra type parameter to compile) - then the user can change the size where it makes sense to do so.

bplommer: I drafted a PR with a slightly different idea - https://github.com/http4s/http4s/pull/6285

core/shared/src/main/scala/org/http4s/Message.scala
danicheg: Yeah, I think so. Though this would mean we should add extra copying in cases of `Entity.{Empty|Strict}`, because using `covary` isn't an option (`[fs2.Pure] do not conform to method covary's type parameter bounds [F2[x] >: F[x]]`). But still, it's good to highlight the fact of effects evaluating.

danicheg:
I wonder can we merge this as is and continue the work of @bplommer in other PRs? Just because current PR is propagating changes from upstream mostly (with tweaking `toStrict` regarding Entity model).

armanbilge:
Yeah right, the mission of this PR is to merge 0.23 -> main. I think would be good to have answer about your question https://github.com/http4s/http4s/pull/6282#discussion_r846862526 (I'm also not sure btw).

danicheg:
@armanbilge tbh, it's pretty minor. It addresses a super-micro-optimization. And it could be done whenever it becomes significant.

armanbilge:
Oh, I misunderstood. I thought there were some concerns about correctness, not performance there.

armanbilge (approved):

bplommer (approved):

PR #6283, opened 2022-04-11
Use `ce.std.Random` for `Multiparts`

armanbilge:
0.23 follow-up to https://github.com/http4s/http4s/pull/6169.

### DIFF
--- a/core/shared/src/main/scala/org/http4s/multipart/Multiparts.scala
+++ b/core/shared/src/main/scala/org/http4s/multipart/Multiparts.scala
@@ -16,10 +16,10 @@
 
 package org.http4s.multipart
 
+import cats.Functor
 import cats.effect.Sync
+import cats.effect.std.Random
 import cats.syntax.all._
-
-import scala.util.Random
 
 /** An algebra for creating multipart values and boundaries.
   *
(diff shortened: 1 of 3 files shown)


core/shared/src/main/scala/org/http4s/multipart/Multiparts.scala
```diff
@@ -36,22 +36,17 @@ trait Multiparts[F[_]] {
...
 object Multiparts {
 
-  /** Creates a `scala.util.Random` and provides Multiparts around it.
+  /** Creates a `cats.effect.std.Random` and provides Multiparts around it.
     * This instance can be shared, or is cheap enough to create closer
     * to where Multipart values are generated.
     */
   def forSync[F[_]](implicit F: Sync[F]): F[Multiparts[F]] =
-    F.delay(new Random()).map(fromScalaRandom[F])
+    Random.scalaUtilRandom[F].map(fromRandom[F])
```
armanbilge, line 44: FTR I feel weird about this methods. `Multiparts` doesn't need `Sync`. It so happens that you can get a `Random` with a `Sync`, but to me that seems to be no business of `Multiparts`.

core/shared/src/main/scala/org/http4s/multipart/Multiparts.scala
```diff
@@ -36,22 +36,17 @@ trait Multiparts[F[_]] {
...
 
-  def fromScalaRandom[F[_]](random: Random)(implicit F: Sync[F]): Multiparts[F] =
```
danicheg, line 46: That's nice refactoring, but there is a nuance. This means we will end up with non-binary compatible `0.22.13` and `0.23.12` versions. Not so good. Probably making it deprecated and adding new methods is a better option.

armanbilge: Sorry, I don't understand 😕 
1. This change is only on 0.23 series
2. This method has not been publish yet
3. MiMa in CI passes

What am I missing?

danicheg: > This change is only on 0.23 series

I mean that  `fromScalaRandom` and friends there are in 0.22 https://github.com/http4s/http4s/blob/series/0.22/core/src/main/scala/org/http4s/multipart/Multiparts.scala#L46

> MiMa in CI passes

I guess because we haven't released fresh 0.22 still.

armanbilge: But, we don't compare MiMa/bincompat between 0.22 and 0.23. They are _supposed_ to be binary-breaking, because of CE2/CE3 :)

danicheg: Oops. But initially, I just addressed user experience.

danicheg: But I agree, on migration from CE2 to CE3 users will have so many issues, so probably this is a tiny one :)

core/shared/src/main/scala/org/http4s/multipart/Multiparts.scala
rossabaker: Yeah, It's just a convenience method.  Since it already existed, I'd be inclined to keep it.  The important thing is we can create Multiparts without that constraint now.

core/shared/src/main/scala/org/http4s/multipart/Multiparts.scala
rossabaker: I probably would have deprecated this one, but I'm a notorious API conservative.

danicheg (approved):
This is waiting for a long time but actually is pretty good to get merged.

PR #6285, opened 2022-04-11
Move `withContentLength` from Message to  Headers

bplommer:
Following discussion on #6282 - I think this is too low-level to expose on `Message`.

### DIFF
--- a/core/shared/src/main/scala/org/http4s/Message.scala
+++ b/core/shared/src/main/scala/org/http4s/Message.scala
@@ -39,7 +39,6 @@ import fs2.text.utf8
 import org.http4s.headers._
 import org.http4s.internal.CurlConverter
 import org.http4s.syntax.KleisliSyntax
-import org.http4s.syntax.header._
 import org.log4s.getLogger
 import org.typelevel.ci.CIString
 import org.typelevel.vault._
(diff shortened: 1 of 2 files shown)


armanbilge:
Definitely agree this is a low-level and kind of unsafe method. I guess I'm confused in what context this is used (it seems we don't use it internally?). In those cases, aren't you more likely interacting with a `Message`?

bplommer:
It's used in `toStrict` - I think that's the only place.

core/shared/src/main/scala/org/http4s/Message.scala
```diff
@@ -301,7 +253,7 @@ sealed trait Message[F[_]] extends Media[F] { self =>
   def toStrict(implicit F: Concurrent[F]): F[Self] =
     body.compile.to(Chunk).map { chunk =>
       withBodyStream(Stream.chunk(chunk))
-        .withContentLength(`Content-Length`.unsafeFromLong(chunk.size.toLong))
+        .transformHeaders(_.withContentLength(`Content-Length`.unsafeFromLong(chunk.size.toLong)))
     }
```
armanbilge, line 257: Ah sorry, missed this. Hm. Also now I realize I'm not sure how this interacts with https://github.com/http4s/http4s/pull/6286 which places the `Content-Length` header on the entity itself. Shouldn't that be the canonical?

rossabaker (approved):
@armanbilge's concern is legitimate, but I am currently leaning that we ought to have a `Length` value type.  If people agree with that, then I think this is good as is.

core/shared/src/main/scala/org/http4s/Message.scala
rossabaker: Well, that would be introduced on main, and doesn't change this for 0.23 purposes, right?  I don't think #6286 is a reason to hold this after all?

armanbilge: Ah, right. Didn't mean to hold this one up.

armanbilge (approved):

Issue #2371, opened 2019-01-23
Add StreamedJsonArray wrapper


chuwy:
Default `EntityEncoder` for `Stream[F, A]` works that all `A`s are streamed as invalid JSON:

```
{"name": "Bob"}{"name": "Fred"}
```

instead of

```
[{"name": "Bob"},{"name": "Fred"}]
```

I believe that this behavior is slightly confusing, but nevertheless valid. However, we can add auxiliary wrapper class and corresponding encoder to it. Example implementation: https://gist.github.com/andimiller/671d5bfdfbf9ef1411522ca9c88a369c

rossabaker:
This relates to #407.  For arbitrary `A`, we don't necessarily have a way to concatenate `A` in a reasonable way.  I'd say it should depend on a `Monoid[A]`, but that would defeat streaming.  Maybe the streaming encoder should be explicit with start, middle, and end delimiters, and then we could provide implicit ones that do the right thing for `Byte`, `String`, `Json`, and others that come to mind.

rossabaker:
Catching up on backlog, and I stand by my previous comment.

SoTeKie:
Is there anybody working on this?
If possible I'd love to get my hands on this, I will probably be unavailable until next week though.

armanbilge:
@SoTeKie sure, it's yours if you want :)

SoTeKie:
Hello again, sadly for personal reasons I won't actually have the time to work on this :( 
Sorry for any inconvenience caused by this.

gringrape:
If there is no one working on this, I would love to get my hands on this.

armanbilge:
@gringrape awesome!

NothingADSR123:
Hey, Is the Issue resolved ? If not I'd love to contribute to this.

PR #6298, opened 2022-04-16
EntityEncoder - Avoid implicit encoders for streams.

diesalbla:
Streams of items can be encoded in many ways. The Core defines a simple definition, which is to append without separators the encoding of each item into the body. This may cause some problems. On the other hand, an encoding to Json may choose to wrap items in a Json array, as if the stream was just a list sent in parts.

Implicits should only be used when the mapping from type to implementation is _almost_ unique. Since this is not the case for encoding streams, it may be best to avoid implicit defs, and let users of library explicitly choose. Otherwise the implicit resolution can blindside developers. 

_Note_: this change is not source-compatible, but should be binary compatible.

### DIFF
--- a/core/shared/src/main/scala/org/http4s/EntityEncoder.scala
+++ b/core/shared/src/main/scala/org/http4s/EntityEncoder.scala
@@ -122,7 +122,7 @@ object EntityEncoder {
     * bodies in advance.  As such, it does not calculate the Content-Length in
     * advance.  This is for use with chunked transfer encoding.
     */
-  implicit def streamEncoder[F[_], A](implicit
+  def streamEncoder[F[_], A](implicit
       W: EntityEncoder[F, A]
   ): EntityEncoder[F, Stream[F, A]] =
     new EntityEncoder[F, Stream[F, A]] {
--- a/blaze-client/src/test/scala/org/http4s/blaze/client/BlazeClientBase.scala
+++ b/blaze-client/src/test/scala/org/http4s/blaze/client/BlazeClientBase.scala
@@ -83,7 +83,8 @@ trait BlazeClientBase extends Http4sSuite {
         RoutesToHandlerAdapter(
           HttpRoutes.of[IO] {
             case Method.GET -> Root / "infinite" =>
-              Response[IO](Ok).withEntity(Stream.emit[IO, String]("a" * 8 * 1024).repeat).pure[IO]
+              val entity: Stream[IO, String] = Stream.emit("a" * 8 * 1024).repeat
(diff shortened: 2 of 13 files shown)


core/shared/src/main/scala/org/http4s/EntityEncoder.scala
```diff
@@ -122,7 +122,7 @@ object EntityEncoder {
     * bodies in advance.  As such, it does not calculate the Content-Length in
     * advance.  This is for use with chunked transfer encoding.
     */
-  implicit def streamEncoder[F[_], A](implicit
+  def streamEncoder[F[_], A](implicit
       W: EntityEncoder[F, A]
   ): EntityEncoder[F, Stream[F, A]] =
```
armanbilge, line 127: This will make the change source-compatible. Users relying on the `implicit` will get a deprecation.

```suggestion
  @deprecated(...)
  implicit def implicitStreamEncoder[F[_], A](implicit
      W: EntityEncoder[F, A]
  ): EntityEncoder[F, Stream[F, A]] = streamEncoder[F, A]

  def streamEncoder[F[_], A](implicit
      W: EntityEncoder[F, A]
  ): EntityEncoder[F, Stream[F, A]] =
```

rossabaker:
I wrote this encoder many years ago, so I'm the last one to ask if it's surprising, but I can't remember a single user complaint.  I can see that it could be surprising for a `Stream[F, Json]` (which has a natural delimiter), but it seems obvious for a `Stream[F, Byte]`.  I'm inclined to leave this alone, but we'll see what others think.

satorg:
I am not that certain about this. Sometimes it may make sense, sometimes may not. For example, in one of my work projects we have a quite big service that has about a hundred of routes to serve and almost all of them use same encoder and decoder. So currently they all get passed implicitly. But if we had to pass them explicitly, then it would lead us to repeat these parameters literally everywhere which would make the code quite less clear.

Also I'd refer to the Akka HTTP library which (if I'm not mistaken) still utilizes implicit encoders. And seems they're fine about that.

Also worth to mention that it is always possible to override the default implicit encoder by declaring (or importing) another one in an inner scope. It should be done once in the scope in contrast to repeating explicit parameters in each call.

rossabaker:
There's a school of thought that codecs are data types, not instances.  That's the approach that Skunk has taken.  There are pros and cons to it, but it would be a fairly radical idea for us.  

What I don't think makes sense is targeting just these.  If we continue down the path of implicit codecs (like the JSON libraries) instead of data types (like Skunk), then I think our instances are reasonable, and also have precedent on their side.

armanbilge:
Related discussion:
- https://github.com/http4s/http4s/issues/2371

PR #6299, opened 2022-04-16
Message - Use Strict Entity for string responses

diesalbla:


core/shared/src/main/scala/org/http4s/Message.scala
```diff
@@ -764,7 +763,7 @@ object Response extends KleisliSyntax {
   val notFound: Response[Pure] =
     Response(
       Status.NotFound,
-      entity = Entity(Stream("Not found").through(utf8.encode)),
+      entity = Entity.Strict(Chunk.array("Not found".getBytes(StandardCharsets.UTF_8))),
```
armanbilge, line 766: A strict UTF8 string seems a common pattern, I wonder if it deserves a special helper constructor.

rossabaker (approved):

ChristopherDavenport (approved):

core/shared/src/main/scala/org/http4s/Message.scala
diesalbla: Adding in another PR.

PR #6300, opened 2022-04-16
Tweak unused args suppressing

danicheg:
A little clean-up.


server/shared/src/main/scala/org/http4s/server/middleware/Throttle.scala
```diff
@@ -112,10 +113,9 @@ object Throttle {
     createBucket.map(bucket => apply(bucket, defaultResponse[G] _)(http))
   }
 
-  def defaultResponse[F[_]](retryAfter: Option[FiniteDuration]): Response[F] = {
-    val _ = retryAfter
+  @nowarn
```
rossabaker, line 116: Can this have a category?

server/shared/src/main/scala/org/http4s/server/middleware/Throttle.scala
```diff
@@ -112,10 +113,9 @@ object Throttle {
...
+  def defaultResponse[F[_]](retryAfter: Option[FiniteDuration]): Response[F] =
     Response[F](Status.TooManyRequests)
```
satorg, line 118: I'd suggest instead
```suggestion
  def defaultResponse[F[_]](@nowarn("cat=unused-params") retryAfter: Option[FiniteDuration]): Response[F] =
    Response[F](Status.TooManyRequests)
```

rossabaker: I like the first one, if all the supported Scala versions like it.

armanbilge: `cat` doesn't always work with Scala 3, but they all seem to agree on `msg`.

satorg: Unfortunately you're right:
- Scala 3.0.2 does not support `@nowarn` (i.e. simply ignores it).
- Scala 3.1.2 supports `@nowarn` but recognizes certain categories only:  "cat=deprecation", "cat=feature", "cat=unchecked" and raises an error for any other category name.

It is so annoying...

satorg: After a small investigation I've figured out that perhaps the best all-compilers compatible (i.e. assuming 2.12.15, 2.13.8, 3.0.2 and 3.1.2) but still quite precise way could be
```scala
def defaultResponse[F[_]](@nowarn retryAfter: Option[FiniteDuration]): Response[F]
```
But on Scala 3.1.2 it still may issue a warning
> @nowarn annotation does not suppress any warnings

😄  🤷

armanbilge: Ah that's right, because Scala 3 doesn't know about unused at all actually ...

rossabaker: Perhaps the `val _` is actually the simplest for this one after all.

danicheg: That was my intent to get rid of `val _` :)
Is plain `nowarn` hurting really?

satorg: Well, `@nowarn` without any additional details may suppress some unexpected warnings which can be introduced in the future. I believe, this is the only harm, but I would not under-estimate it. Therefore `@nowarn` applied to a whole method will suppress all potential warnings for that methods. Whereas `@nowarn` applied to a single parameter only is less harmful, apparently.

I am kind of hesitant what is really better in this case – the bare `@nowarn` applied to the parameter, or that `val _ = paramName` work-around. The latter would target the unused param warning more precisely, but the former would be easier to spot and fix eventually once the full support for Scala3 will be implemented.

satorg: Just realized... In fact, we can try utilizing the `@unused` annotation (i.e. `scala.annotation.unused`).

The only obstacle – it is defined in Scala library since 2.13 only. Whereas Scala library for 2.12.x just does not have it declared.

But we can work-around it in a cross-compile compatible way. For example, we can create a `main/scala-2.12` source directory, then create our own `unused` annotation version:
```scala
package scala.annotation

@nowarn("msg=subclassing ClassfileAnnotation does not\nmake your annotation visible at runtime")
final class unused extends nowarn("cat=unused-params")
```

Therefore, the fixed version of the fragment being discussing:
```suggestion
  def defaultResponse[F[_]](@unused retryAfter: Option[FiniteDuration]): Response[F] =
    Response[F](Status.TooManyRequests)
```
@danicheg @armanbilge @rossabaker – wdyt?

satorg: More thoughts on this...

Turns out that `scala-collection-compat` already defines `@unused` for Scala 2.12 but does it in a different way:
https://github.com/scala/scala-collection-compat/blob/e35e330fca28c78c39fa2d4812fa33ce892f19e3/compat/src/main/scala-2.11_2.12/scala/annotation/unused.scala#L15

Both definitions work for Scala 2.12.15, but AFAIK the `http4s` library does not depend on `scala-collection-compat` by default (does it?). Therefore to avoid further collisions for downstream projects that have both libraries among their dependencies, it's better to create a compat annotation somewhere inside an `http4s` internal scope.

- for Scala 2.12:
  ```scala
  package org.http4s.internal.compat

  final class unused extends deprecated("unused", "unused") // as defined in scala-collection-compat

  // OR altrernatively:
  //
  // @nowarn("msg=subclassing ClassfileAnnotation does not\nmake your annotation visible at runtime")
  // final class unused extends nowarn("cat=unused-params")
  ```

- for Scala 2.13+:
  ```scala
  package org.http4s.internal.compat

  final class unused extends scala.annotation.unused
  ```

And then import and use the compat `@unused` instead of that from Scala library.

satorg: Now I'm wondering if I should take the same approach for my work in typelevel/cats#4187 🤔

rossabaker: We briefly provided something in the `scala` package for 2.12 -- it might have been that annotation -- and it made a mess of people's assemblies.

I'm content with the blank `@nowarn` applied to the parameter, knowing Scala 2.12 is going away in http4s-1.0, and that it currently happens in exactly one place.  The compat annotation is very smart, but seems like a lot of work for one use site in the twilight of its life.

rossabaker (approved):

nifledcar7-blip:
http4s/http4s#6300

PR #6305, opened 2022-04-17
Deprecate Entity Apply

diesalbla:
Since `Entity.Stream` is not a good default, because of the overhead of dealing with streams, we deprecate the Entity.apply method in favour of a explicit `stream` constructor.

We also add a explicit `string` constructor for strict entities.

### DIFF
--- a/core/shared/src/main/scala/org/http4s/Entity.scala
+++ b/core/shared/src/main/scala/org/http4s/Entity.scala
@@ -32,28 +32,42 @@ sealed trait Entity[+F[_]] {
 }
 
 object Entity {
-  def apply[F[_]](body: EntityBody[F], length: Option[Long] = None): Entity[F] =
-    Default(body, length)
+
+  @deprecated("Use Entity.stream instead", "1.0.0-M38")
+  def apply[F[_]](body: Stream[F, Byte], length: Option[Long] = None): Entity[F] =
+    stream(body, length)
+
+  def stream[F[_]](body: Stream[F, Byte], length: Option[Long] = None): Entity[F] =
+    Streamed(body, length)
 
   // The type parameter aids type inference in Message constructors.
   def empty[F[_]]: Entity[F] = Empty
   def strict(bytes: ByteVector): Entity[Pure] = Strict(bytes)
 
-  final case class Default[+F[_]](body: EntityBody[F], length: Option[Long]) extends Entity[F] {
+  /** Generates a Strict entity with the string encoded the given charset. */
+  def string(str: String, charset: Charset): Entity[Pure] =
+    Entity.Strict(ByteVector(str.getBytes(charset.nioCharset)))
+
+  /** Generates a Strict entity with the string encoded the given charset. */
+  def utf8String(str: String): Entity[Pure] =
+    string(str, Charset.`UTF-8`)
+
+  final case class Streamed[+F[_]](body: EntityBody[F], length: Option[Long]) extends Entity[F] {
     def ++[F1[x] >: F[x]](that: Entity[F1]): Entity[F1] = that match {
-      case d: Default[F1] => Default(body ++ d.body, (length, d.length).mapN(_ + _))
-      case strict @ Strict(bytes) => Default(body ++ strict.body, length.map(_ + bytes.size))
+      case d: Streamed[F1] => Streamed(body ++ d.body, (length, d.length).mapN(_ + _))
+      case strict @ Strict(bytes) => Streamed(body ++ strict.body, length.map(_ + bytes.size))
       case Empty => this
     }
 
-    def translate[F1[x] >: F[x], G[_]](fk: F1 ~> G): Entity[G] = Default(body.translate(fk), length)
+    def translate[F1[x] >: F[x], G[_]](fk: F1 ~> G): Entity[G] =
+      Streamed(body.translate(fk), length)
 
     override def toString: String = length match {
       case None =>
-        "Entity.Default"
+        "Entity.Streamed"
 
       case Some(dataSize) =>
-        s"Entity.Default($dataSize bytes total)"
+        s"Entity.Streamed($dataSize bytes total)"
     }
   }
 
@@ -63,7 +77,7 @@ object Entity {
     val length: Option[Long] = Some(bytes.size)
 
     def ++[F1[x] >: Pure[x]](that: Entity[F1]): Entity[F1] = that match {
-      case d: Default[F1] => Default(body ++ d.body, d.length.map(bytes.size + _))
+      case d: Streamed[F1] => Streamed(body ++ d.body, d.length.map(bytes.size + _))
       case Strict(bytes2) => Strict(bytes ++ bytes2)
       case Empty => this
     }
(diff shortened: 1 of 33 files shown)


satorg:
Agree. I am kind of 👍 for getting rid of over-use of `apply` methods in favor of explicitly named static ones.

core/shared/src/main/scala/org/http4s/Entity.scala
```diff
@@ -31,7 +31,12 @@ sealed trait Entity[+F[_]] {
 }
 
 object Entity {
-  def apply[F[_]](body: EntityBody[F], length: Option[Long] = None): Entity[F] =
+
+//  @deprecated("Use Entity.stream instead", "1.0.0-M32")
+//  def apply[F[_]](body: Stream[F, Byte], length: Option[Long] = None): Entity[F] =
+//    stream(body, length)
```
rossabaker, line 37: I would keep the deprecation to give people a nudge in the right direction.  I don't see a lot of Entity creation directly in user code, but still, we need to consider the experience of people moving to 1.0.

rossabaker (commented):
:+1: in principle, but I'm pro-deprecation.

core/shared/src/main/scala/org/http4s/Entity.scala
diesalbla: I was commenting it in order to reveal compile errors... kept.

danicheg (approved):
Despite a few suggestions below, this is great :+1:

core/shared/src/main/scala/org/http4s/Entity.scala
```diff
@@ -31,13 +33,23 @@ sealed trait Entity[+F[_]] {
 }
 
 object Entity {
-  def apply[F[_]](body: EntityBody[F], length: Option[Long] = None): Entity[F] =
+
+  @deprecated("Use Entity.stream instead", "1.0.0-M32")
+  def apply[F[_]](body: Stream[F, Byte], length: Option[Long] = None): Entity[F] =
+    stream(body, length)
+
+  def stream[F[_]](body: Stream[F, Byte], length: Option[Long] = None): Entity[F] =
```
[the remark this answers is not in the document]
danicheg: Then probably we should rename `Default` -> `Streamed`? :) Though I have nothing against `Default`. From the perspective of extending the `Entity` model by adding new subtypes, say for `Transfer-Encoding: chunked`, then `stream` will be ambiguous. But if someone feels this is a premature thing, this safely could be skipped.

core/shared/src/main/scala/org/http4s/Entity.scala
```diff
@@ -32,13 +35,24 @@ sealed trait Entity[+F[_]] {
 }
 
 object Entity {
-  def apply[F[_]](body: EntityBody[F], length: Option[Long] = None): Entity[F] =
+
+  @deprecated("Use Entity.stream instead", "1.0.0-M32")
```
armanbilge, line 39: We need to update this.

diesalbla: done!

core/shared/src/main/scala/org/http4s/Entity.scala
```diff
@@ -32,13 +35,24 @@ sealed trait Entity[+F[_]] {
...
     Default(body, length)
 
   // The type parameter aids type inference in Message constructors.
   def empty[F[_]]: Entity[F] = Empty
   def strict(bytes: ByteVector): Entity[Pure] = Strict(bytes)
 
+  /** Generates a Strict entity with the string encoded the given charset,
+    * which is UTF_8 by default.
+    */
+  def string(str: String, charset: JavaCharset = StandardCharsets.UTF_8): Entity[Pure] =
```
danicheg, line 53: I have somehow misseen that you used Java's `Charset` here. We have a dedicated type `org.http4s.Charset`. Let's use it instead, wdyt?

diesalbla: This was just replacing what we had in, e.g., the `Message` file, where it is currently using the Java one. Perhaps that may be another PR?

danicheg: Although I'm not against doing it in separate PR, we shouldn't release this version of `Entity.string` basically. I'm in favor of this refactoring overall.

diesalbla: Done

core/shared/src/main/scala/org/http4s/multipart/Part.scala
```diff
@@ -44,16 +42,13 @@ object Part {
   def formData(name: String, value: String, headers: Header.ToRaw*): Part[fs2.Pure] =
     Part(
       Headers(`Content-Disposition`("form-data", Map(ci"name" -> name))).put(headers: _*),
-      Entity.Strict(ByteVector.view(value.getBytes(UTF_8))),
+      Entity.string(value),
```
danicheg, line 45: Previously charset was set explicitly. Shouldn't we do the same? Though it's hard to imagine changing the default charset (and the function's argument default value consequently).

danicheg (approved):
I want to drop out one bikeshedding about renaming `Entity.Default` -> `Entity.Streamed` (or any other that would highlight the streaming nature of that entity). Besides that, this PR makes sense to me!

core/shared/src/main/scala/org/http4s/multipart/Part.scala
diesalbla: Done.

danicheg:
I advise getting one more like before getting this merged.

core/shared/src/main/scala/org/http4s/Entity.scala
```diff
@@ -32,13 +32,24 @@ sealed trait Entity[+F[_]] {
...
     Default(body, length)
 
   // The type parameter aids type inference in Message constructors.
   def empty[F[_]]: Entity[F] = Empty
   def strict(bytes: ByteVector): Entity[Pure] = Strict(bytes)
 
+  /** Generates a Strict entity with the string encoded the given charset,
+    * which is UTF_8 by default.
+    */
+  def string(str: String, charset: Charset = Charset.`UTF-8`): Entity[Pure] =
```
rossabaker, line 50: The default charset argument in the Java standard library has been a significant source of bugs for a couple decades.  I would make a `utf8String`.  It could delegate to this without the default argument.

core/shared/src/main/scala/org/http4s/Entity.scala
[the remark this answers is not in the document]
rossabaker: I agree that `Streamed` is a better name for that type.

rossabaker (commented):
I like it.  But if this merges without a scalafix, providing one should be a separate ticket.

core/shared/src/main/scala/org/http4s/Entity.scala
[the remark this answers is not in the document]
diesalbla: renamed.

ChristopherDavenport (approved):

armanbilge (approved):

PR #6306, opened 2022-04-17
Use Strict entities where possible

diesalbla:


rossabaker (approved):

rossabaker:
:+1: in principle, but there are build failures.

rossabaker (approved):

PR #6307, opened 2022-04-17
Fix `SelectOpsMultiple#renderString`

danicheg:


rossabaker (approved):
Cleaning up the `Renderer` is something I've been thinking about for 1.0, but it isn't going away on 0.22 or 0.23, so this seems fine.

PR #6308, opened 2022-04-17
Ember-Core - HeaderToRequestNoBody - Tweak

diesalbla:
- Use exists / forall instead of filter/find/map
- Avoid the call to `String.toUpperCase`, since that would allocate an entire new string that is immediately discarded.

### DIFF
--- a/ember-core/shared/src/main/scala/org/http4s/ember/core/h2/PseudoHeaders.scala
+++ b/ember-core/shared/src/main/scala/org/http4s/ember/core/h2/PseudoHeaders.scala
@@ -59,6 +59,12 @@ private[h2] object PseudoHeaders {
   }
 
   def headersToRequestNoBody(hI: NonEmptyList[(String, String)]): Option[Request[fs2.Pure]] = {
+    def toRawHeader(k: String, v: String): Header.ToRaw =
+      Header.ToRaw.rawToRaw(Header.Raw(org.typelevel.ci.CIString(k), v))
+
+    def isLowerCase(str: String): Boolean =
+      str.forall(c => c.toUpper === c)
+
     // TODO This can be a 1 pass operation. This is not...
     val headers: List[(String, String)] = hI.toList
     val pseudos = headers.takeWhile(_._1.startsWith(":"))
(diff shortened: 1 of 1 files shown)


ember-core/shared/src/main/scala/org/http4s/ember/core/h2/PseudoHeaders.scala
```diff
@@ -70,34 +76,33 @@ private[h2] object PseudoHeaders {
...
     val authority = extractAuthority(pseudos)
 
-    val noOtherPseudo = pseudos
-      .filterNot(t => requestPsedo.contains(t._1))
-      .isEmpty
+    val noOtherPseudo = pseudos.forall(t => requestPsedo.contains(t._1))
 
-    val teIsCorrect = headers.find(_._1 === "te").headOption.fold(true)(_._2 === "trailers")
-    val connectionIsAbsent = headers.find(_._1 === "connection").isEmpty
+    val teIsCorrectAndNoConnection = !headers.exists { case (k, v) =>
```
danicheg, line 81: `!exists(_ == foo)` is equal to `forall(_ != foo)`, isn't?

diesalbla: Yeah... it's one of this cases in which one may want to think this in English... Which one would be easier to say?

> All headers must not be neither 1. a connection, nor 2. a "te" without "trailers" 

> No header can be either 1. a connection, or 2. a "te" without "trailers"

The former bends easier to `.forall( case (k, v) => ! (---) && ! (---) ) `, whereas the latter look closer to the code as is.

ember-core/shared/src/main/scala/org/http4s/ember/core/h2/PseudoHeaders.scala
```diff
@@ -59,6 +59,12 @@ private[h2] object PseudoHeaders {
   }
 
   def headersToRequestNoBody(hI: NonEmptyList[(String, String)]): Option[Request[fs2.Pure]] = {
+    def toRawHeader(k: String, v: String): Header.ToRaw =
+      Header.ToRaw.rawToRaw(Header.Raw(org.typelevel.ci.CIString(k), v))
+
+    def isLowerCase(str: String): Boolean =
```
ChristopherDavenport, line 65: This one I think may create more allocated objects than it saves us.

diesalbla: Are you referring to the `String.forall` boxing the Char characters?

This is the implementation in the library: 

```Scala
  def forall(@deprecatedName("f", "2.13.3") p: Char => Boolean): Boolean = {
    var i = 0
    val len = s.length
    while(i < len) {
      if(!p(s.charAt(i))) return false
      i += 1
    }
    true
  }
```

diesalbla: The `StringOps` is just a syntax-extension "decorator" class, and it does not inherit from any collection-generic trait that may require a `forall` with an `AnyRef` on it... so it should be unboxed... but only the decompiler can tell, to be sure.

PR #6311, opened 2022-04-18

danicheg:


server/shared/src/main/scala/org/http4s/server/middleware/RequestId.scala
```diff
@@ -46,13 +47,17 @@ object RequestId {
   def apply[G[_], F[_]](http: Http[G, F])(implicit G: Sync[G]): Http[G, F] =
     apply(requestIdHeader)(http)
 
+  @deprecated("Preserved for bincompat", "0.23.12")
```
danicheg, line 50: I'm feeling some doubts about changing this deprecated-since version to something 1.0-based here. But if somebody is confident about the necessity of this change, we could do it.

rossabaker: I've usually just left them alone.  There's a whole philsophical discussion to be had there about first linearized version, first version in the same series, date, etc., and I decided it doesn't matter enough I just roll with it.

rossabaker (approved):

server/shared/src/main/scala/org/http4s/server/middleware/RequestId.scala
armanbilge: Actually, can't we just remove it? It's not preserving source-compatibility for sure. Is preserving bincompat on main worth anything?

danicheg: @armanbilge Hmmm. I think you're right. I didn't look over what this deprecation was about.

armanbilge (approved):
Thank you :)

PR #6313, opened 2022-04-19
Clarify connection lifecycle

### LINKED ISSUE #6312: Question: Client persistent connection
In [`Client` ScalaDoc](https://www.javadoc.io/doc/org.http4s/http4s-docs_3/latest/org/http4s/client/Client.html), all requesting methods (`expect`, `fetchAs`, `get`) say that:
> The underlying HTTP connection is closed at the completion of the decoding.

How to keep the connection open across multiple requests?

More generally, libraries such as Python's `requests` have the notion of [`Session`s](https://docs.python-requests.org/en/latest/user/advanced/#session-objects), which preserve cookies etc. Does `http4s` support anything like that?

rossabaker:
Fixes #6312.

### DIFF
--- a/client/src/main/scala/org/http4s/client/Client.scala
+++ b/client/src/main/scala/org/http4s/client/Client.scala
@@ -28,7 +28,14 @@ import org.http4s.headers.Host
 import java.io.IOException
 import scala.util.control.NoStackTrace
 
-/** A [[Client]] submits [[Request]]s to a server and processes the [[Response]]. */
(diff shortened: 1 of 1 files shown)


armanbilge:
👍 I see a couple more references to "dispose" on the page, should those be changed as well?

https://github.com/http4s/http4s/blob/c27f709a2660d49f2918e30767c4c5e83cc51f4f/client/src/main/scala/org/http4s/client/Client.scala#L64-L74

danicheg (approved):

armanbilge (approved):

PR #6316, opened 2022-04-20
Add timeout and max bytes handling to `Message#toStrict`

danicheg:
We didn't release 0.23 after introducing `Message#toStrict`, so MiMa shouldn't bit us on these changes.

### DIFF
--- a/core/shared/src/main/scala/org/http4s/Message.scala
+++ b/core/shared/src/main/scala/org/http4s/Message.scala
@@ -31,7 +31,11 @@ import com.comcast.ip4s.Hostname
 import com.comcast.ip4s.IpAddress
 import com.comcast.ip4s.Port
 import com.comcast.ip4s.SocketAddress
-import fs2.{Chunk, Pipe, Pull, Pure, Stream}
+import fs2.Chunk
+import fs2.Pipe
+import fs2.Pull
+import fs2.Pure
+import fs2.Stream
 import fs2.io.net.unixsocket.UnixSocketAddress
 import fs2.text.utf8
 import org.http4s.Message.EntityStreamException
@@ -43,7 +47,8 @@ import org.typelevel.ci.CIString
 import org.typelevel.vault._
 
 import java.io.File
-import scala.concurrent.duration.{Duration, FiniteDuration}
+import scala.concurrent.duration.Duration
+import scala.concurrent.duration.FiniteDuration
 import scala.util.control.NoStackTrace
 import scala.util.hashing.MurmurHash3
 
--- a/core/shared/src/main/scala/org/http4s/Message.scala
+++ b/core/shared/src/main/scala/org/http4s/Message.scala
@@ -23,6 +23,7 @@ import cats.data.OptionT
 import cats.effect.Concurrent
 import cats.effect.Sync
 import cats.effect.SyncIO
+import cats.effect.kernel.Temporal
 import cats.syntax.all._
 import cats.~>
 import com.comcast.ip4s.Dns
@@ -30,12 +31,10 @@ import com.comcast.ip4s.Hostname
 import com.comcast.ip4s.IpAddress
 import com.comcast.ip4s.Port
 import com.comcast.ip4s.SocketAddress
-import fs2.Chunk
-import fs2.Pipe
-import fs2.Pure
-import fs2.Stream
+import fs2.{Chunk, Pipe, Pull, Pure, Stream}
 import fs2.io.net.unixsocket.UnixSocketAddress
 import fs2.text.utf8
+import org.http4s.Message.EntityStreamException
 import org.http4s.headers._
 import org.http4s.internal.CurlConverter
 import org.http4s.syntax.KleisliSyntax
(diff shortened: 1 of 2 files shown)


server/shared/src/main/scala/org/http4s/server/middleware/BodyCache.scala
```diff
@@ -66,7 +66,7 @@ object BodyCache {
     apply(app)(identity, _ => identity)(FunctionK.id)
 
   private def compileBody[F[_]: Concurrent](req: Request[F]): F[Request[F]] =
-    req.toStrict
+    req.unsafeToStrict
```
danicheg, line 69: I wonder, shouldn't the `BodyCache` take into account body length and timeout also? Probably, it's worth creating a 'safe' version. If someone else agreed with that, I might work on it too. But that is out of the scope of this PR for sure.

armanbilge: See some discussion about that here:
- https://github.com/http4s/http4s/pull/5321#issuecomment-935245046

core/shared/src/main/scala/org/http4s/Message.scala
```diff
@@ -246,18 +252,77 @@ sealed trait Message[F[_]] extends Media[F] { self =>
...
-    * `Content-Length` header.  It is the caller's responsibility to
-    * assure there is enough memory to materialize the body.
+    * body. Replaces any `Transfer-Encoding: chunked` with a
+    * `Content-Length` header. It is the caller's responsibility to
+    * assure there is enough memory to materialize the body. Also,
+    * potentially this method could run infinitely.
+    * Consider to use [[org.http4s.Message.toStrict]] instead.
     */
-  def toStrict(implicit F: Concurrent[F]): F[Self] =
+  def unsafeToStrict(implicit F: Concurrent[F]): F[Self] =
```
rossabaker, line 261: I'm confused what makes this unsafe.  We deal with effects that never complete all the time.  There is a risk of an infinite stream, but we're using methods fs2 doesn't consider unsafe.  And I think this is equivalent to the "safe" `unsafeToStrict(Duration.Inf, None)`.

core/shared/src/main/scala/org/http4s/Message.scala
```diff
@@ -246,18 +252,77 @@ sealed trait Message[F[_]] extends Media[F] { self =>
...
+    * set an appropriate timeout.
+    *
+    * @param timeout duration to wait for the entity stream completion.
+    *                Exceeding finite timeout duration
+    *                (an instance of [[FiniteDuration]]) will lead to
+    *                failing processing with the `TimeoutException`
+    * @param maxBytes maximum length of the entity stream. If the stream
+    *                 exceeds the limit then processing fails with the [[EntityStreamException]]
+    */
+  def toStrict(timeout: Duration, maxBytes: Option[Long])(implicit F: Temporal[F]): F[Self] = {
```
rossabaker, line 280: Timeout is orthogonal: we can just apply it to the result of this.  I guess maxBytes is handy, but that too could be an orthogonal operation.  And then we're back to the original `toStrict`, which was fine by me.

core/shared/src/main/scala/org/http4s/Message.scala
danicheg: I supposed when implementing that the main problem we defining the public method, it's better if users will pay more attention when using this method. Also, `akka-http` have the same [method](https://doc.akka.io/docs/akka-http/current/routing-dsl/directives/basic-directives/toStrictEntity.html) with time/size limitations.

core/shared/src/main/scala/org/http4s/Message.scala
danicheg: I agree that timeout is orthogonal, fewer about the size limiting (because it goes first when transforming the entity's body). But my whole point is that it's better to slightly force users to make a deliberate decision when using `toStrict`.

rossabaker: I think I'd be happy with a `toStrict(maxBytes: Option[Long])`.  The alternative would be an `limitEntity(maxBytes: Long)` that composes with `.toStrict`.

core/shared/src/main/scala/org/http4s/Message.scala
```diff
@@ -246,18 +252,77 @@ sealed trait Message[F[_]] extends Media[F] { self =>
...
+            .take(limit)
+            .flatMap {
+              case Some(_) =>
+                Pull.raiseError[F](EntityStreamException.createWithDefaultMsg(limit))
+              case None =>
+                Pull.done
+            }
+            .stream
+
+        limitingPipe(entityBody)
```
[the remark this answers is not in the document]
danicheg: I can't agree with this much, `Option[Long]` encodes three invariants and `ToPull#take` takes a `Long` parameter. I believe `None` isn't equal 0 here in behavior. Though, I could be wrong.

core/shared/src/main/scala/org/http4s/Message.scala
danicheg: Touche 🤝  Let's make it so.

core/shared/src/main/scala/org/http4s/Message.scala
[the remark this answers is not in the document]
danicheg: Oh, sorry, I just now get your point. And yes, of course, we could remove that Pipe.

PR #6317, opened 2022-04-20
ember: Add support for strict entities, look for content-length only once

wjoel:


ember-core/shared/src/main/scala/org/http4s/ember/core/Parser.scala
```diff
@@ -430,13 +429,14 @@ private[ember] object Parser {
...
         read: Read[F],
-    )(implicit F: Concurrent[F]): F[(EntityBody[F], Drain[F])] =
+    )(implicit F: Concurrent[F]): F[(Entity[F], Drain[F])] =
       if (contentLength > 0) {
-        if (buffer.length >= contentLength) {
-          val (body, rest) = buffer.splitAt(contentLength.toInt)
-          val eb: EntityBody[F] = Stream.chunk(Chunk.byteVector(ByteVector(body)))
-          val drain: Drain[F] = F.pure(Some(rest))
-          F.pure(eb -> drain)
+        if (buffer.length == contentLength) {
```
wjoel, line 434: Possibly not worth the extra condition.

diesalbla:
@wjoel How are you running those "techempower" benchmarks, so we may reproduce them? We would like to try out a few things.

wjoel:
@diesalbla it's a bit of a mess (I'm still just trying things out), but basically the branch over at https://github.com/wjoel/FrameworkBenchmarks/tree/ember-tests, `cp -r ~/.ivy2/local/org.http4s/ /home/wjoel/github/FrameworkBenchmarks/frameworks/Scala/http4s/` and then `./tfb --mode benchmark --type plaintext json --test http4s` (or just `./tfb --mode benchmark --test http4s` for all tests).

You can also use their test containers for profiling when running the WebServer in an IDE by running `docker run --network=host techempower/tfb.wrk wrk -H 'Host: tfb-server' -H 'Accept: text/plain,text/html;q=0.9,application/xhtml+xml;q=0.9,application/xml;q=0.8,*/*;q=0.7' -H 'Connection: keep-alive' --latency -d 15 -c 256 --timeout 8 -t 24 http://localhost:8080/plaintext -s pipeline.lua -- 16`

ember-core/shared/src/main/scala/org/http4s/ember/core/Parser.scala
```diff
@@ -430,13 +429,14 @@ private[ember] object Parser {
...
+          val drain: Drain[F] = F.pure(Some(Array.emptyByteArray): Option[Array[Byte]])
```
danicheg, line 435:
```suggestion
          val drain: Drain[F] = F.pure(Option(Array.emptyByteArray))
```

ember-core/shared/src/main/scala/org/http4s/ember/core/Parser.scala
```diff
@@ -430,13 +429,14 @@ private[ember] object Parser {
...
+          F.pure(Entity.strict(Chunk.array(buffer)) -> drain)
+        } else if (buffer.length > contentLength) {
+          val drain: Drain[F] = F.pure(Some(buffer.drop(contentLength.toInt)): Option[Array[Byte]])
```
danicheg, line 438:
```suggestion
          val drain: Drain[F] = F.pure(Option(buffer.drop(contentLength.toInt)))
```

danicheg (approved):
Don't rely much on my :+1:, because I'm not an actual ember user, but this seems legit to me.

ChristopherDavenport (approved):

PR #6319, opened 2022-04-21
Take into account the `Entity` model in the `ResponseLogger`

danicheg:


danicheg:
Gently asking for a review 🙏🏻  I don't want this to become stale.

rossabaker:
Sorry, I am always a bit behind, and a few like this get caught between my newest-first and oldest-first thrashings. :weary:

rossabaker (approved):

PR #6323, opened 2022-04-23
Address the entity model in the server's `RequestLogger`

danicheg:


server/shared/src/main/scala/org/http4s/server/middleware/RequestLogger.scala
```diff
@@ -79,35 +79,45 @@ object RequestLogger {
...
+        } <* fk(logAct)
+    }
+
+    Kleisli {
+      case req if !logBody =>
+        logRequest(req)
+      case req =>
+        req.entity match {
+          case Entity.Empty | Entity.Strict(_) =>
+            logRequest(req)
```
wjoel, line 100: `Entity.Strict` has a body that should be logged if requested, but it's a lot more straightforward than what we need to do for `Entity.Default`.

server/shared/src/main/scala/org/http4s/server/middleware/RequestLogger.scala
```diff
@@ -79,35 +79,45 @@ object RequestLogger {
...
+                val newBody = Stream.eval(vec.get).flatMap(v => Stream.emits(v)).unchunks
+                val logRequest: F[Unit] = logMessage(req.withBodyStream(newBody))
+
+                http(changedRequest)
+                  .guaranteeCase {
+                    case Outcome.Succeeded(_) => G.unit
+                    case _ => fk(logRequest)
+                  }
+                  .map(_.pipeBodyThrough(_.onFinalizeWeak(logRequest)))
               }
```
diesalbla, line 119: Could we extract the body of this lambda as a function? 
```Scala
def bufferRequest(vv: Vector[Chunk[Byte]]): F[Response[F]]
```

danicheg: Where do you think we should place this function? I'm not too sure about the `Message` as a placeholder. It discloses implementation details of the particular subtype of the `Entity`.

diesalbla: Oh, I meant as a nested function of `impl`. Perhaps before the `Kleisli` at line 94, in which case it would also need the request parameter.

rossabaker (approved):

PR #6325, opened 2022-04-24
Address the entity model in the `Client#fromHttpApp`

danicheg:
Sorry for annoying with that kind of PRs, but I believe they bring some performance enhancements. Though I don't give the numbers to support my points, it just goes from common sense and looks natural.


rossabaker (approved):
I don't think this needs the Ref unless the body is a Stream, but this is progress.

danicheg:
@rossabaker Yeah, I was thinking about this too. The current implementation is using a single Ref within checking/processing bodies of request and response. Basically, it's a matrix 2x2, and there is only one case where we won't create Ref. Probably, it's worth pushing forward, but I'm happy if we can do it iteratively.

Issue #6331, opened 2022-04-27
EOL for the `0.22` series


danicheg:
It has been discussed multiple times in different PRs, so I think it's worth opening the issue and making this process transparent. What stops us from 'archiving' the 0.22 series at this point? I guess we have to stop updating 0.22 through Scala Steward. And we have to remove some update ignorings in the config (that we are getting from the 0.22 series now). Anything else?

armanbilge:
Thank you very much for opening the issue to do this properly 👍

We should also release one final 0.22.x and update the website.
https://github.com/http4s/http4s/blob/9ca373147fedf9486d42290bdceb976bc3096f42/project/Http4sSitePlugin.scala#L115-L118

danicheg:
So, roughly roadmap looks like this:
1. Merging of https://github.com/http4s/http4s/pull/6332
2. Fixing external Scala Steward config to don't update dependencies for the 0.22 branch.
3. Fixing internal Scala Steward config in the 0.23 branch.
4. Releasing the 'final' (at the moment) 0.22 version.
5. Updating website (0.22 becomes EOL).

If I missed something, please, don't be silent.

armanbilge:
There are 9 header-related PRs targeting 0.22:
https://github.com/http4s/http4s/pulls?q=is%3Apr+is%3Aopen+base%3Aseries%2F0.22+

Unfortunately I don't think anyone has time to review them, so I will retarget them to 0.23.

danicheg:
https://github.com/http4s/http4s/releases/tag/v0.22.13

PR #6332, opened 2022-04-27
Roll back the #6254

danicheg:
This is a follow-up from https://github.com/http4s/http4s/pull/6266#issuecomment-1104014154
Also, this addresses the #6331.


armanbilge (approved):
I don't know blaze but there seemed to be consensus on this revert 👍

danicheg:
This is actually a straightforward revert. But, yeah, I'm crossing my fingers that I did it correctly 🤞🏻

PR #6240, opened 2022-04-04
Spin off servlet, jetty-server, and tomcat modules

rossabaker:
They have moved as part of The Great Schism:

* [http4s-servlet](https://github.com/http4s/http4s-servlet)
* [http4s-jetty](https://github.com/http4s/http4s-jetty)
* [http4s-tomcat](https://github.com/http4s/http4s-tomcat)

jetty-client will eventually move to http4s-jetty, but we have work to do with the TCK.

PR #6354, opened 2022-05-02
Delete servlet and jetty-server srcs

armanbilge:
I'm confused why they weren't deleted as part of https://github.com/http4s/http4s/pull/6240. These sources aren't compiled as part of any project.


rossabaker (approved):
I think the scope of that ticket expanded and I just missed it.

PR #6355, opened 2022-05-02
Bye bye play-json

armanbilge:
Now lives at https://github.com/http4s/http4s-play-json


danicheg (approved):
Also, are we need to publish the separated module then?

armanbilge:
> Also, are we need to publish the separated module then?

Snapshots are already being published:
https://s01.oss.sonatype.org/content/repositories/snapshots/org/http4s/http4s-play-json_2.13/0.23-85af5cf-SNAPSHOT/

So, the question remains, when we release http4s 0.23.12, we can either:
1. Also publish v0.23.12 of all the satellite modules against http4s-core v0.23.12, even if they have no changes; or
2. Explain in the release notes that these modules will now be published independently.

@danicheg what are your thoughts on this?

Issue #6360, opened 2022-05-02
CI cache step isn't caching anything ???


armanbilge:
Super weird. Seems to be working fine in other repos, no idea what's up here.

```
Received 22 of 22 (100.0%), 0.0 MBs/sec
Cache Size: ~0 MB (22 B)
/usr/bin/tar --use-compress-program zstd -d -xf /home/runner/work/_temp/e546e7c6-89ba-4600-8ede-d14dad582952/cache.tzst -P -C /home/runner/work/http4s/http4s
Cache restored successfully

Cache restored from key: Linux-sbt-cache-v2-5b06b3838e8b17ed7ae1fec5ec43b05b24c02fe7dd7850e3d0edd2268a878ccd-cd22cb1893f8f8bde18b4535632a62a46a4bd34ee04559465378212d0a1497a7
```

PR #6361, opened 2022-05-02
Only create scalafix job for 2.13

armanbilge:
First of all the extra jobs were worthless, because they weren't doing anything.

Secondly, because they ran so quickly (i.e. without starting sbt and downloading stuff) they were the first in line to upload the cache ... of nothing 🙃 

Fixes https://github.com/http4s/http4s/issues/6360.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -39,10 +39,9 @@ ThisBuild / githubWorkflowAddedJobs ++= Seq(
       WorkflowStep.Run(
         List("cd scalafix", "sbt ci"),
         name = Some("Scalafix tests"),
-        cond = Some(s"matrix.scala == '$scala_213'"),
       )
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -39,10 +39,9 @@ ThisBuild / githubWorkflowAddedJobs ++= Seq(
       WorkflowStep.Run(
         List("cd scalafix", "sbt ci"),
```
armanbilge, line 40: Still, this fast-running job is not particularly helpful for our cache, since it doesn't use the primary `build.sbt`. This is something to fix in sbt-typelevel.
- https://github.com/typelevel/sbt-typelevel/issues/268

danicheg (approved):

PR #6367, opened 2022-05-06
Add RetryCountKey - Allow other middlewares access to what retry count we are on.

ChristopherDavenport:
This will allow other middlewares the ability to see retry count on either the request or response depending on where the middlewares are applied.

### DIFF
--- a/client/shared/src/main/scala/org/http4s/client/middleware/Retry.scala
+++ b/client/shared/src/main/scala/org/http4s/client/middleware/Retry.scala
@@ -27,6 +27,7 @@ import org.http4s.headers.`Idempotency-Key`
 import org.http4s.headers.`Retry-After`
 import org.log4s.getLogger
 import org.typelevel.ci.CIString
+import org.typelevel.vault.Key
 
 import java.time.Instant
 import java.time.temporal.ChronoUnit
@@ -34,12 +35,11 @@ import scala.concurrent.duration._
 import scala.math.min
 import scala.math.pow
 import scala.math.random
-import org.typelevel.vault.Key
 
 object Retry {
   private[this] val logger = getLogger
 
-  val RetryCountKey = Key.newKey[cats.effect.SyncIO, Int].unsafeRunSync()
+  val RetryCountKey: Key[Int] = Key.newKey[cats.effect.SyncIO, Int].unsafeRunSync()
 
   def apply[F[_]](
       policy: RetryPolicy[F],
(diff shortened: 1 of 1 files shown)


ChristopherDavenport:
Recommend hiding whitespace as it reformatted a bunch more than what I changed.

client/shared/src/main/scala/org/http4s/client/middleware/Retry.scala
```diff
@@ -38,6 +39,8 @@ import scala.math.random
 object Retry {
   private[this] val logger = getLogger
 
+  val RetryCountKey: Key[Int] = Key.newKey[cats.effect.SyncIO, Int].unsafeRunSync()
```
armanbilge, line 42: Maybe we need a brief doc here. Is this the attempt we are on (e.g. 1 for first, 2 for second) or the number of previous attempts (0 for first, 1 for second)?

armanbilge: So, the initial request should set this key to `0`? Because I'm not sure if that's what it currently does, I think it starts from 1.
https://github.com/http4s/http4s/blob/331a588be0c4bbdf5672420e06329ce6e23757db/client/shared/src/main/scala/org/http4s/client/middleware/Retry.scala#L128

ChristopherDavenport: In my case, which is the open telemetry standard. I think this should just be set to the number http4s uses and other should adjust based on that.

armanbilge: Sounds good to me. Sorry, one last bikeshed: would `AttemptCount` make more sense?

ChristopherDavenport: I like it, more clear. Will switch.

armanbilge (approved):

isomarcte (approved):

client/shared/src/main/scala/org/http4s/client/middleware/Retry.scala
```diff
@@ -38,6 +39,14 @@ import scala.math.random
 object Retry {
   private[this] val logger = getLogger
 
+  /** This key tracks the attempt count, the retry middleware starts the very
+    * first request with 1 as the first request, even with no retries
+    * (aka the first attempt). If one wants to monitor retries
+    * explicitly and not the attempts one may want to subtract the
+    * value of this key by 1.
+    */
+  val AttemptCountKey: Key[Int] = Key.newKey[cats.effect.SyncIO, Int].unsafeRunSync()
```
danicheg, line 48: Just wondering, can we mark this lazy?

armanbilge: Hmm, do we have a precedent for making keys lazy in http4s? I don't think it's very expensive, and it's already lazy depending on whether the relevant classes are loaded or not anyway.

danicheg: Yeah, that's correct. No key is defined as 'lazy' at the moment (ok, most of those I've seen). But is it intended?

danicheg: Though, my comment isn't about this particular PR, and don't hesitate on merging.

armanbilge: > But is it intended?

IMHO I think it's fine. Allocating a key is cheap, and it will only happen if the middleware class is loaded into the JVM anyway because of lazy class initialization. Meanwhile, any `lazy val` adds an additional synchronization point which probably has some performance cost over vanilla `val`.

Issue #6363, opened 2022-05-05
StaticFile on symbolic link


gekomad:
On Linux and http4s 1.0.0-M32 this line works `StaticFile.fromPath(fs2.io.file.Path(s"/home/foo/index.html"), Some(req)).getOrElseF(NotFound())` but if the file is a symbolik link I see only 2 rows of html page on chrome, fireforx and curl. This problem there isn't on http4s 0.21.33

armanbilge:
Thanks for reporting! Would you mind checking if the issue exists in 1.0.0-M27 and 1.0.0-M28? It may have been introduced in https://github.com/http4s/http4s/pull/5226.

armanbilge:
Actually, if you don't mind checking v0.23.0 and v0.23.1 as well that would be really helpful.

gekomad:
1.0.0-M28 doesn't work
1.0.0-M27 with `StaticFile.fromFile` works

armanbilge:
@gekomad thank you for trying it, that is very helpful. I'd like to fix this but I'll need your help to reproduce it first. I wrote a test in https://github.com/http4s/http4s/commit/85b4e622a91d5f515a0d76d23cdd6421b14ff366 but it seems to work ok.

gekomad:
this my project:

```scala
val Http4sVersion = "1.0.0-M32"

version := "0.0.1-SNAPSHOT"
scalaVersion := "2.13.8"
libraryDependencies += "org.http4s"                 %% "http4s-blaze-server"    % Http4sVersion
libraryDependencies += "org.http4s"                 %% "http4s-dsl"    % Http4sVersion
```

```scala
package com.example.quickstart

import cats.effect._
import org.http4s.blaze.server.BlazeServerBuilder
import org.http4s.{HttpRoutes, StaticFile}
import cats.effect.unsafe.implicits.global
import org.http4s.dsl.io._

object Main extends IOApp {

  val routes = HttpRoutes.of[IO] {
    case request@GET -> Root =>
          StaticFile.fromPath(fs2.io.file.Path("/home/foo/index.html"), Some(request)).getOrElseF(NotFound())
  }

  override def run(args: List[String]): IO[ExitCode] =
    BlazeServerBuilder[IO]
      .bindHttp(9000, "0.0.0.0").withHttpApp(routes.orNotFound).resource.use(_ => IO.never).start.unsafeRunSync().joinWithNever
}

```

armanbilge:
[...]

PR #6368, opened 2022-05-06
Follow links when accessing file attributes in `StaticFile`

armanbilge:
Fixes https://github.com/http4s/http4s/issues/6363.

### DIFF
--- a/tests/shared/src/test/scala/org/http4s/StaticFileSuite.scala
+++ b/tests/shared/src/test/scala/org/http4s/StaticFileSuite.scala
@@ -114,6 +114,14 @@ class StaticFileSuite extends Http4sSuite {
     emptyFile.flatMap(StaticFile.fromPath[IO](_).value).map(_.isDefined).assert
   }
 
+  test("handle a symlink") {
+    StaticFile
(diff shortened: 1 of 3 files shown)


tests/shared/src/test/scala/org/http4s/StaticFileSuite.scala
```diff
@@ -114,6 +114,14 @@ class StaticFileSuite extends Http4sSuite {
     emptyFile.flatMap(StaticFile.fromPath[IO](_).value).map(_.isDefined).assert
   }
 
+  test("handle a symlink") {
+    StaticFile
+      .fromPath[IO](Path("tests/shared/src/test/resources/lipsum-symlink.txt"))
+      .semiflatMap(_.body.compile.count)
+      .getOrElse(0)
+      .assertEquals(24005)
```
danicheg, line 122: What's the magic number? Could we replace it with computation?

armanbilge: Yes we can, we can just directly read the size of the file `lorem-ipsum.txt` instead of via the symlink.

danicheg (approved):

PR #6369, opened 2022-05-09
Address the `Entity` model in `Arbitrary[Entity[*]]`

danicheg:


laws/src/main/scala/org/http4s/laws/discipline/arbitrary.scala
```diff
@@ -860,7 +860,15 @@ private[discipline] trait ArbitraryInstances { this: ArbitraryInstancesBinCompat
...
-      } yield Entity(body, length)
+        entity = length match {
+          case Some(0) =>
+            Entity.Empty
+          case Some(_) =>
+            Entity.strict(body.compile.to(fs2.Chunk))
+          case None =>
+            Entity(body)
+        }
+      } yield entity
```
armanbilge, line 871: What do you think about having separate generators for the different types of entities, and then choosing between them?

danicheg: In fact, they'll serve only one generator and won't be reused. So it's a question about tastes, I think.

armanbilge: I think it also helps for reasoning about the generator. For example, this generator doesn't included a chunked streaming entity with a known length, but I didn't realize it until now.

danicheg: Touche 🤔 Applied your advice in a recent patch

armanbilge (approved):
Looks good, thanks :)

PR #6370, opened 2022-05-10
WIP: Instrumentation of blaze-server with otel4s

rossabaker:


rossabaker:
This is just a fit check for the otel4s API.  Don't take it too seriously... yet.

rossabaker:
I think the core library can add zero additional dependencies, and be a candidate for instrumenting backends on 0.23.

wjoel:
This is very interesting. @rossabaker, do you think it would be possible to get "request parent" trace spans that start when the http4s backend starts reading a request, and which end when the backend finishes writing the response? I don't think it's currently possible to do with a middleware, but perhaps I'm mistaken?

rossabaker:
Yes, getting all the way to the edges of the backend is important. It will also be able to see the times backends "fill in the blanks", such as recovering a failed error as a 500 response.

TimWSpence:
@rossabaker I've been having some problems trying to switch from http4s-jetty to ember (timeouts leading to kubernetes healthcheck failures). I came to discord asking if there was any instrumentation for ember to help me diagnose it and he pointed me to here. Is there anything I can do to help move this forwards? :)

TimWSpence:
Aaand I just realized this is for blaze 🤦 But @armanbilge suggested I open a similar one for Ember so I'll have a go at that

PR #6371, opened 2022-05-10
Fix regression of `Router#define`

danicheg:
The current implementation of `Router#define` from 0.23 differs from that we have at main. I think this was done accidentally in https://github.com/http4s/http4s/commit/8b148f48ae2c167f36d27f2b64a92bb638ca24ec#diff-7b0e93ead3858152f51e0f8ceb5dd1aec7c0eee85dd58a812e1192d204e7a059L42.

### DIFF
--- a/server/shared/src/main/scala/org/http4s/server/Router.scala
+++ b/server/shared/src/main/scala/org/http4s/server/Router.scala
@@ -76,12 +76,14 @@ object Router {
     mappings.sortBy(_._1.length).foldLeft(default) { case (acc, (prefix, routes)) =>
       val prefixPath = Uri.Path.unsafeFromString(prefix)
 
-      Kleisli { req =>
-        if (req.pathInfo.startsWith(prefixPath))
-          routes(translate(prefixPath)(req)).orElse(acc(req))
-        else
-          acc(req)
-      }
+      if (prefixPath.isEmpty) routes <+> acc
+      else
+        Kleisli { req =>
+          if (req.pathInfo.startsWith(prefixPath))
+            routes(translate(prefixPath)(req)).orElse(acc(req))
+          else
+            acc(req)
+        }
     }
 
   def of[F[_]: Monad](mappings: Routable[F]*): HttpRoutes[F] =
--- a/server/shared/src/test/scala/org/http4s/server/RouterSuite.scala
+++ b/server/shared/src/test/scala/org/http4s/server/RouterSuite.scala
@@ -80,6 +80,14 @@ class RouterSuite extends Http4sSuite {
     "/routable" -> routable,
   )
 
+  private val emptyAttributeTestRoutes = {
+    import org.http4s.dsl.io._
+    HttpRoutes.of[IO] { case r @ GET -> Root / "foo" =>
+      assert(r.attributes.isEmpty, "Requests attributes should be empty")
+      Ok()
+    }
+  }
+
   test("routable") {
     service
       .orNotFound(Request[IO](GET, uri"/routable/1"))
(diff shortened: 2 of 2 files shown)


rossabaker:
Are we missing a test?

danicheg:
It looks so. Though I didn't dig deep in that, just found this on merging 0.23 -> main.

danicheg:
@rossabaker returning to this one, did you hint that we need to add a test? Or maybe if this is good enough, we can merge?

rossabaker (approved):
I think a test would be good, because it wasn't obvious to me that the original was broken, but we can leave an issue and declare victory here.

armanbilge:
> I think a test would be good, because it wasn't obvious to me that the original was broken, but we can leave an issue

Apologies for hijacking this PR, but I do feel that (lack of) tests is a recurring issue. For example, reviewing Diego's H2 PRs is terrifying because I'm just not sure whether there's test coverage on those changes or not. In general, improving test coverage seems a good thing to do for 1.0.

I think scoverage used to be a thing here, but not anymore. Was there a reason, and could it make sense to bring it back?

danicheg:
I agree that tests are a must-have thing. Here I just fixed the accidental regression. But ok, I'll add a test for this case for avoidance precedents.

server/jvm/src/test/scala/org/http4s/server/middleware/RouterSuite.scala
```diff
@@ -0,0 +1,27 @@
...
+import org.http4s.Request
+import org.http4s.Status
+import org.http4s.server.Router
+import org.http4s.syntax.all._
+
+class RouterSuite extends Http4sSuite {
+  private val routes = {
+    import org.http4s.dsl.io._
+    HttpRoutes.of[IO] { case r @ GET -> Root / "foo" =>
+      assert(r.attributes.isEmpty, "Requests attributes should be empty")
```
danicheg, line 15: This isn't fancy but works. On the previous `Router#define` assert would fail. Also, if everybody would like this, I'll do a backport to 0.23.

armanbilge: Thanks! Yes, running this up through 0.23 seems to be the way 👍 if you don't mind 😅

armanbilge: Btw, why is it in JVM sources, does it not work on JS?

danicheg: mehhhh, I'll fix this

danicheg: Thanks for pointing that out, fixed.

armanbilge: I was kind of surprised we didn't have a `RouterSuite` yet!

rossabaker:
I have a bad attitude toward scoverage because of a boss who turned it into an OKR.  Once it becomes a metric, it becomes a gamified metric.  

It also used to discourage people on Cats, who would see test coverage drop infintesimally, when the non-determinism of property tests introduced more noise than the signal of a small PR to a large project.

Then again, I manually added a bunch of printlns to the servlet last night and piped through sort and uniq to give myself a really shitty scoverage clone.

So maybe yes if we're all adult enough to use it as a guide, and absolutely not if it becomes a scorecard.

armanbilge:
Ok, thanks, that sounds like a lot more no than yes 😅

FrancescoSerra:
I'd agree with @rossabaker. Scoverage metrics in my experience tend to introduce a lot of grief, as they bear little value, apart from the semantics of "this method has been called in this test", which doesn't really prove that some logic has been tested. On the other hand, I think that an optimized goal for the team could be to strive for making as much as possible illegal states unrepresentable, so to use the compiler's support full on, and document where this doesn't/can't happen.
Caveat: I'm new to the codebase and I don't know what's the state of the art of the test suite, so if I stated the obvious or the impractical I apologize

PR #6372, opened 2022-05-10
Fix `Request#productElement` and `Response#productElement`

danicheg:
I think these methods should contemplate `Entity[F]` rather than `Stream[F, Byte]`.


armanbilge (approved):
Good catch.

PR #6373, opened 2022-05-11
Add info about the entity into the `toString` of `Request` and `Response`

danicheg:
Since we have overridden the `toString` of `Entity`, we can enhance the `toString` of Request/Response.


rossabaker (approved):
I think this is fine, but does the size of the entity ever reveal sensitive information?  (Probably not, or the Content-Length is also likely to be a problem.)

:+1: if I didn't frighten anybody else with that.

danicheg:
Yeah, my intuition also is that we already expose the entity size via headers (perhaps, not for all cases, but still). Though is this a problem - it's an open question. If I'm not wrong, akka-http also exposes the body size by default.

Because it's easy to revert, I take the responsibility and merge this 🤞🏻

PR #6374, opened 2022-05-13
Ember-Core: reimplement write loop without streams.

diesalbla:
The `writeLoop` was using FS2 streams without strict need. The streams in question are generated, by pulling all available data from a queue, and immediately processed. This is the kind of processing that can be done with a monadically recursive loop. 

For this use case, we can make use of a newer Queue operation, tryTakeN, which if given a `None` parameter just dumps the entire queue on each iteration.

_Note:_ The new operation `tryTakeN` is not being added to `cats-effect` until the next minor release, so this PR should be kept as draft.

### DIFF
--- a/ember-core/shared/src/main/scala/org/http4s/ember/core/h2/H2Connection.scala
+++ b/ember-core/shared/src/main/scala/org/http4s/ember/core/h2/H2Connection.scala
@@ -192,12 +192,12 @@ private[h2] class H2Connection[F[_]](
     firstGoAway.getOrElse(F.unit) >> go(chunk)
   }
 
-  def writeLoop: Stream[F, Nothing] =
-    Stream
-      .fromQueueUnterminatedChunk[F, H2Frame](outgoing, Int.MaxValue)
-      .chunks
-      .foreach(writeChunk)
-      .handleErrorWith(ex => Stream.exec(logger.debug(ex)("writeLoop terminated")))
+  def writeLoop: F[Unit] =
+    outgoing.tryTakeN(None).flatMap { (chs: List[Chunk[H2Frame]]) =>
+      val ch = Chunk.concat(chs)
+      writeChunk(ch).handleErrorWith(ex => logger.debug(ex)("writeLoop terminated")) >> writeLoop
+    }
+
   // TODO Split Frames between Data and Others Hold Data If we are at cap
   //  Currently will backpressure at the data frame till its cleared
 
--- a/ember-core/shared/src/main/scala/org/http4s/ember/core/h2/H2Client.scala
+++ b/ember-core/shared/src/main/scala/org/http4s/ember/core/h2/H2Client.scala
@@ -272,7 +272,7 @@ private[ember] class H2Client[F[_]](
     for {
       h2 <- Resource.eval(createH2Connection)
       _ <- h2.readLoop.background
-      _ <- h2.writeLoop.compile.drain.background
+      _ <- h2.writeLoop.background
       _ <- clearClosed(h2).background
       _ <- pullCreatedStreams(h2).background
       _ <- Resource.eval(processSettings(h2))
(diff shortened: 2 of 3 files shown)


project/Http4sPlugin.scala
```diff
@@ -100,7 +100,7 @@ object Http4sPlugin extends AutoPlugin {
     val blaze = "0.15.3"
     val caseInsensitive = "1.3.0"
     val cats = "2.8.0"
-    val catsEffect = "3.3.14"
+    val catsEffect = "3.4.0-RC1"
```
armanbilge, line 103:
```suggestion
    val catsEffect = "3.4.0-RC2"
```

armanbilge:
@diesalbla I've opened https://github.com/http4s/http4s/pull/6747 which fixes the CE 3.4.0 deprecations.

ember-core/shared/src/main/scala/org/http4s/ember/core/h2/H2Connection.scala
```diff
@@ -141,53 +141,59 @@ private[h2] class H2Connection[F[_]](
...
+      ch.collectFirst { case g: H2Frame.GoAway =>
+        mapRef.get.flatMap { m =>
+          m.values.toList.traverse_(connection => connection.receiveGoAway(g))
+        } >> state.update(s => s.copy(closed = true))
+      }.sequence_
+
+    def loop: F[Unit] =
+      outgoing.tryTakeN(None).flatMap { (chs: List[Chunk[H2Frame]]) =>
+        val chunk = Chunk.concat(chs)
+        firstGoAway(chunk) >> go(chunk) >> loop
```
danicheg, line 191: Sorry for perhaps a silly question, but in the current implementation, all elements available in the queue `outgoing` will be dequeued and emitted as a single chunk once.  Although `QueueSource#tryTakeN(None)` also dequeues and emits all available elements, you're doing it in a loop. Could you please elaborate on why we need a loop here?

diesalbla: Because the queue is constantly filling up, so after each "dequeue" operation is completed, the queue is still receiving new elements. 

A Stream intrinsically encodes that "pulling" loop, whereby each pull from the stream repeats the dequeue operation. What the PR is doing is cutting off that middle-man stream.

ember-core/shared/src/main/scala/org/http4s/ember/core/h2/H2Connection.scala
```diff
@@ -192,12 +192,12 @@ private[h2] class H2Connection[F[_]](
...
-    Stream
-      .fromQueueUnterminatedChunk[F, H2Frame](outgoing, Int.MaxValue)
-      .chunks
-      .foreach(writeChunk)
-      .handleErrorWith(ex => Stream.exec(logger.debug(ex)("writeLoop terminated")))
+  def writeLoop: F[Unit] =
+    outgoing.tryTakeN(None).flatMap { (chs: List[Chunk[H2Frame]]) =>
+      val ch = Chunk.concat(chs)
+      writeChunk(ch).handleErrorWith(ex => logger.debug(ex)("writeLoop terminated")) >> writeLoop
+    }
```
armanbilge, line 199: Isn't this a busy-loop?

diesalbla: Busy Loop?

armanbilge: What I mean is that `tryTakeN(None)` always returns immediately i.e. it will not semantically block waiting for elements in the queue. So the program will always be busy looping.

https://en.wikipedia.org/wiki/Busy_loop

diesalbla: Oh, I see. I wonder, could we make `outgoing` into a `Channel`, as was done for the `body` below?

PR #6376, opened 2022-05-13
Drain response body in `DefaultClient#defaultOnError`

### LINKED ISSUE #3707: Improve Logging On Failed Expect

danicheg:
This fixes #3707.
Credit to @ChristopherDavenport and @rossabaker for their work/discussion in #3708.

### DIFF
--- a/blaze-client/src/main/scala/org/http4s/blaze/client/BlazeClient.scala
+++ b/blaze-client/src/main/scala/org/http4s/blaze/client/BlazeClient.scala
@@ -18,6 +18,7 @@ package org.http4s
 package blaze
 package client
 
+import cats.Applicative
 import cats.effect.implicits._
 import cats.effect.kernel.Async
 import cats.effect.kernel.Deferred
@@ -30,6 +31,7 @@ import org.http4s.blazecore.ResponseHeaderTimeoutStage
 import org.http4s.client.Client
 import org.http4s.client.DefaultClient
 import org.http4s.client.RequestKey
+import org.http4s.client.UnexpectedStatus
 import org.http4s.client.middleware.Retry
 import org.http4s.client.middleware.RetryPolicy
 
@@ -94,6 +96,11 @@ private class BlazeClient[F[_], A <: BlazeConnection[F]](
       response <- responseResource
     } yield response
   }
+
+  override def defaultOnError(req: Request[F])(resp: Response[F])(implicit
+      G: Applicative[F]
+  ): F[Throwable] =
+    resp.body.compile.drain.as(UnexpectedStatus(resp.status, req.method, req.uri))
 
   private def prepareConnection(key: RequestKey): Resource[F, (A, F[TimeoutException])] = for {
     conn <- borrowConnection(key)
--- a/ember-client/shared/src/main/scala/org/http4s/ember/client/EmberClient.scala
+++ b/ember-client/shared/src/main/scala/org/http4s/ember/client/EmberClient.scala
@@ -16,7 +16,9 @@
 
 package org.http4s.ember.client
 
+import cats.Applicative
 import cats.effect._
+import cats.syntax.functor._
 import org.http4s._
 import org.http4s.client._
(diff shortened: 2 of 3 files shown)


armanbilge:
> Unfortunately, the same fix couldn't be applied for ember-client, because it has a weaker `MonadCancelThrow` constraint.

Can we pattern match on the constraint, to see if it is in fact `Concurrent` and then do the right thing in that case? This is an idea that started in Cats Effect, e.g. see https://github.com/typelevel/cats-effect/issues/2772

blaze-client/src/main/scala/org/http4s/blaze/client/BlazeClient.scala
```diff
@@ -95,6 +97,11 @@ private class BlazeClient[F[_], A <: BlazeConnection[F]](
     } yield response
   }
 
+  override def defaultOnError(req: Request[F])(resp: Response[F])(implicit
+      G: Applicative[F]
+  ): F[Throwable] =
+    resp.body.compile.drain.as(UnexpectedStatus(resp.status, req.method, req.uri))
```
danicheg, line 103: Reminder: in `main` this should be like
```scala
  override def defaultOnError(req: Request[F])(resp: Response[F])(implicit
      G: Applicative[F]
  ): F[Throwable] =
    resp.entity match {
      case Entity.Empty | Entity.Strict(_) =>
        F.pure(UnexpectedStatus(resp.status, req.method, req.uri))
      case Entity.Default(body, _) =>
        body.compile.drain.as(UnexpectedStatus(resp.status, req.method, req.uri))
    }
```

danicheg:
@armanbilge oh, indeed, thanks. Though I don't that much like this approach, however it works and uses many times in Typelevel libraries.

danicheg:
Pushing this pattern forward, we could apply it even in `DefaultClient` and have one version for every client. Though this pattern is kinda tricky. So maybe it'd be better to use it in particular implementations, not interfaces. I dunno, left it as-is for comments.

ember-client/shared/src/main/scala/org/http4s/ember/client/EmberClient.scala
```diff
@@ -36,4 +38,15 @@ final class EmberClient[F[_]] private[client] (
...
+
+  override def defaultOnError(req: Request[F])(resp: Response[F])(implicit
+      G: Applicative[F]
+  ): F[Throwable] = F match {
+    case concurrentF: Concurrent[F @unchecked] =>
+      implicit val C = concurrentF
+      resp.body.compile.drain.as(UnexpectedStatus(resp.status, req.method, req.uri))
+
+    case _ =>
+      F.pure(UnexpectedStatus(resp.status, req.method, req.uri))
```
rossabaker, line 50: Is this case worth logging a warning that the behavior is buggy?

danicheg: Basically, I believe this case shouldn't appear at runtime, because 99% of effect usage is CE.IO/monix.Task/etc effect that is supposed to have a `Concurrent` instance. But `defaultOnError` is used for serving requests overall, and end-users could even don't use logging, but this method actually will be used. So this case preserves the current behavior of request processing mostly, not just logging behavior.

rossabaker (approved):
Yeah, it should improve things in most practice, and isn't worse anywhere.

danicheg:
@armanbilge Is your point strict for merging? Don't you mind if we go ahead with this as is?

armanbilge (approved):
I agree with Ross, the only "strict" here is strictly improving :)

PR #714, opened 2016-09-18
Improve Retry client middleware

rossabaker:
* Only retry idempotent requests with pure bodies
* Only retry when status code is appropriate
* Retry failed tasks from client

Partially addresses #712.  Further fixes (e.g., Retry-After) are binary incompatible and would come in a separate PR.

PR #6379, opened 2022-05-15
Resubmit idempotent requests with only pure entities in `Retry` middleware

### LINKED ISSUE #712: Retry should only resubmit idempotent methods with pure bodies
1. Retry middleware should never resubmit a POST or other non-idempotent methods.
2. Retry middleware should only resubmit if it can verify that the body is pure.  Code for this already exists in FollowRedirect, which we can abstract out.  Right now, it just blanks out the body, which is bad.

I imagine this is all binary compatible and suitable for release-0.14.x.

danicheg:
Closes #712. I've crossed my fingers that I correctly followed Ross's reasoning on the issue. Nevertheless, I'm open to taking rotten tomatoes.

### DIFF
--- a/client/shared/src/main/scala/org/http4s/client/middleware/Retry.scala
+++ b/client/shared/src/main/scala/org/http4s/client/middleware/Retry.scala
@@ -184,7 +184,7 @@ object RetryPolicy {
     */
   def defaultRetriable[F[_]](req: Request[F], result: Either[Throwable, Response[F]]): Boolean = {
     def isReqIdempotentAndEntityPure =
-      req.method.isIdempotent && req.isPureEntity
+      req.method.isIdempotent && req.hasKnownPureEntity
 
     def isHeadersContainsIdempotencyHeader =
       req.headers.contains[`Idempotency-Key`]
--- a/client/shared/src/main/scala/org/http4s/client/middleware/Retry.scala
+++ b/client/shared/src/main/scala/org/http4s/client/middleware/Retry.scala
@@ -169,7 +169,7 @@ object RetryPolicy {
     GatewayTimeout,
   )
 
-  /** Returns true if (the request method is idempotent and its entity isn't streamed;
+  /** Returns true if (the request method is idempotent and the request's entity is pure (isn't effectful);
     * or request contains [[`Idempotency-Key`]] header) and the result is either a throwable or
     * has one of the `RetriableStatuses`.
     *
--- a/client/shared/src/main/scala/org/http4s/client/middleware/Retry.scala
+++ b/client/shared/src/main/scala/org/http4s/client/middleware/Retry.scala
@@ -169,9 +169,15 @@ object RetryPolicy {
     GatewayTimeout,
   )
 
-  /** Returns true if (the request method is idempotent and the request's entity is pure (isn't effectful);
-    * or request contains [[`Idempotency-Key`]] header) and the result is either a throwable or
-    * has one of the `RetriableStatuses`.
+  /** Returns true if both the request and response effect are retriable.
+    *
+    * The request is retriable if either:
+    * - its method is idempotent and its request entity is pure
+    * - it has an [[`Idempotency-Key`]] header
+    *
+    * The response effect is retriable if:
+    * - it raised an error
+    * - it is a response whose status code is a member of `RetriableStatuses`
     *
     * Caution: a request wouldn't be resubmitted if the idempotent request contains
     * streamed (effectful) [[Entity]] (nor empty, nor fully loaded into memory).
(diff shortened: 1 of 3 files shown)


core/shared/src/main/scala/org/http4s/Message.scala
```diff
@@ -465,6 +465,22 @@ final class Request[+F[_]] private (
...
+   *         'true' if entity is empty or already loaded into memory;
+   *         'false' if entity is streamed.
+   */
+  def isPureEntity: Boolean =
+    entity match {
+      case Entity.Empty | Entity.Strict(_) =>
+        true
+      case Entity.Default(_, _) =>
+        false
+    }
```
armanbilge, line 482: Note that a `Stream`ing entity can also be pure, it's just harder to check this :) something to think about in method naming and/or scaladoc e.g. `isKnownPureEntity` (as in Scala's `knownSize` methods).

client/shared/src/main/scala/org/http4s/client/middleware/Retry.scala
```diff
@@ -169,16 +169,23 @@ object RetryPolicy {
...
-    * and the result is either a throwable or has one of the `RetriableStatuses`.
+  /** Returns true if (the request method is idempotent and the request's entity is pure (isn't effectful);
+    * or request contains [[`Idempotency-Key`]] header) and the result is either a throwable or
+    * has one of the `RetriableStatuses`.
     *
-    * Caution: if the request body is effectful, the effects will be
-    * run twice.  The most common symptom of this will be resubmitting
-    * an idempotent request.
+    * Caution: a request wouldn't be resubmitted if the idempotent request contains
+    * streamed (effectful) [[Entity]] (nor empty, nor fully loaded into memory).
```
armanbilge, line 177: I'm not sure if we need this "caution" anymore, I think these changes make this method safer.

core/shared/src/main/scala/org/http4s/Message.scala
```diff
@@ -465,6 +465,22 @@ final class Request[+F[_]] private (
...
 
+  /**
+   * Checks whether the [[Entity]] of the [[Message]] is pure in the sense of
+   * indicating that the entity body can be re-run without side effects.
+   *
+   * @return the [[Boolean]] value:
+   *         'true' if entity is empty or already loaded into memory;
+   *         'false' if entity is streamed.
+   */
+  def isPureEntity: Boolean =
```
danicheg, line 476: Probably, we could add this to the `isIdempotent` method directly. But I feel it's too risky because it may hurt people's runtime in bizarre ways. Changing the behavior of the `Retry` middleware seems a bit less risky. I could be wrong here, so your comments are appreciated so much.

core/shared/src/main/scala/org/http4s/Message.scala
armanbilge: Btw, should this be `hasPureEntity`, since it's defined on `Message`?

client/shared/src/main/scala/org/http4s/client/middleware/Retry.scala
danicheg: Yep, but this changes behavior without changes in API. So maybe it's worth leaving it, I dunno..

core/shared/src/main/scala/org/http4s/Message.scala
armanbilge: I could be wrong about this, but I think there's an interesting distinction here between an idempotent request (i.e., has no side-effect on the server-side) and a request with a pure entity (i.e., has no side-effect in the client-side).

core/shared/src/main/scala/org/http4s/Message.scala
danicheg: If we do some archeological work, in the past http4s contained something you are talking about https://github.com/http4s/http4s/pull/714/files#diff-adcd6450231fbb969572de7bdd97e36c7dd50fe3e577135b505c3a7e90320077R39, but it was removed. But yeah, you're right that we need to rename this method.

danicheg: Renamed to `hasPureEntity`. Let it sound more native.

rossabaker: I don't think there's any way to determine whether a stream is pure or not at runtime.  The algebra isn't introspectable like it was in antiquity.  Something like "hasKnownPureEntity" works for me.

rossabaker (approved):

core/shared/src/main/scala/org/http4s/Message.scala
danicheg: My two cents about the 'known' suffix - it gives not so many insights. Users still have to check scaladoc/impl to get what it actually means. But If you insist, I'm not that much against it.

PR #6387, opened 2022-05-17
Add max wait for connection borrowing in pool manager

### LINKED ISSUE #836: Consider a max wait in pool manager
#796 was a bug, but it's too easy to deadlock a server.  We should consider a timer so borrows don't hang indefinitely for the next time we or our an app does something stupid.

danicheg:
Closes http4s/blaze#654. It's better to review with hidden whitespaces. The impl is super straightforward, and I could be missing something stupidly obvious.

### DIFF
--- a/blaze-client/src/main/scala/org/http4s/blaze/client/PoolManager.scala
+++ b/blaze-client/src/main/scala/org/http4s/blaze/client/PoolManager.scala
@@ -210,17 +210,18 @@ private final class PoolManager[F[_], A <: Connection[F]](
       F.defer(
         Ref[F]
           .of(false)
-          .flatMap(ref =>
+          .flatMap(cancelingRef =>
             F.async[NextConnection] { callback =>
               semaphore.permit.use { _ =>
                 if (!isClosed) {
                   def go(): F[Unit] =
-                    ref.get.flatMap {
+                    cancelingRef.get.flatMap {
                       case true =>
-                        F.delay(
-                          logger.warn(s"Requesting connection for $key has exceed timeout")
-                        ) *>
-                          F.delay(callback(Left(ConnectionBorrowingException(key))))
+                        val errorMsg =
+                          s"Requesting connection for $key was canceled"
+
+                        F.delay(logger.warn(errorMsg)) *>
+                          F.delay(callback(Left(new IllegalStateException(errorMsg))))
 
                       case false =>
                         getConnectionFromQueue(key).flatMap {
@@ -300,7 +301,7 @@ private final class PoolManager[F[_], A <: Connection[F]](
                   F.delay(callback(Left(new IllegalStateException("Connection pool is closed"))))
                     .as(None)
               }
-            }.onCancel(ref.set(true))
+            }.onCancel(cancelingRef.set(true))
           )
       )
 
@@ -494,4 +495,4 @@ final case class NoConnectionAllowedException(key: RequestKey)
     extends IllegalArgumentException(s"No client connections allowed to $key")
 
 final case class ConnectionBorrowingException(key: RequestKey)
-    extends IllegalStateException(s"Requesting connection for $key has exceed timeout")
+    extends IllegalStateException(s"Requesting connection for $key has exceeded the timeout")
--- a/blaze-client/src/main/scala/org/http4s/blaze/client/PoolManager.scala
+++ b/blaze-client/src/main/scala/org/http4s/blaze/client/PoolManager.scala
@@ -37,7 +37,8 @@ final case class WaitQueueFullFailure() extends RuntimeException {
 }
 
 /** @param maxIdleDuration the maximum time a connection can be idle
-  * and still be borrowed
+  *                        and still be borrowed
+  * @param maxBorrowDuration the maximum time for connection borrowing
   */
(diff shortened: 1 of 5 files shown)


blaze-client/src/test/scala/org/http4s/blaze/client/PoolManagerSuite.scala
```diff
@@ -261,4 +263,38 @@ class PoolManagerSuite extends Http4sSuite with AllSyntax {
...
+        .assert
+    } yield ()
+  }
+
+  test("Should borrow connection if it doesn't exceed the maxBorrowDuration") {
+    for {
+      pool <- mkPool(
+        maxTotal = 1,
+        maxIdleDuration = Duration.Inf,
+        maxBorrowDuration = 2.seconds,
```
danicheg, line 289: Anecdotally says, but at local, this test-suite takes about 0.001s on execution. However, at CI it may become flaky. But still, I didn't mark it as 'flaky' before receiving any facts about its flakiness.

danicheg: Lovely. It turned out that the other test suite became flaky. mehh

blaze-client/src/main/scala/org/http4s/blaze/client/PoolManager.scala
```diff
@@ -201,78 +205,115 @@ private final class PoolManager[F[_], A <: Connection[F]](
...
-                        getConnectionFromQueue(randKey).map(
-                          _.fold(
-                            logger.warn(s"No connection to evict from the idleQueue for $randKey")
-                          )(_.conn.shutdown())
+  def borrow(key: RequestKey): F[NextConnection] = {
+    val connectionBorrowF =
+      F.defer(
+        Ref[F]
+          .of(false)
+          .flatMap(ref =>
```
rossabaker, line 213: `ref` could get a more descriptive name here.  I had to keep reading a while to understand what this was.

blaze-client/src/main/scala/org/http4s/blaze/client/PoolManager.scala
```diff
@@ -201,78 +205,115 @@ private final class PoolManager[F[_], A <: Connection[F]](
...
-                ) *>
-                  addToWaitQueue(key, callback)
-            }
-
-          F.delay(logger.debug(s"Requesting connection for $key: $stats")).productR(go()).as(None)
-        } else
-          F.delay(callback(Left(new IllegalStateException("Connection pool is closed")))).as(None)
-      }
+                          F.delay(callback(Left(ConnectionBorrowingException(key))))
+
```
rossabaker, line 224: Is this the only reason this can get canceled?  I'm trying to reason about when we'd see this message instead of the timeout on line 311, and whether the message (and callback invocation) is accurate.

danicheg: Uh-oh, you're 100% right. Thanks for pointing that out. For sure, canceling could be caused by various reasons. But do you hint that we shouldn't invoke the callback here? I feel we can't be too sure if CE will resolve the asynchronous deadlock via the standard canceling mechanism smoothly. According to `timeoutTo` scaladoc:
> "If the source is uncancelable, the resulting effect will wait for it to complete before evaluating the fallback."

So I've tried to soften the blow.

rossabaker: I'm not sure what would make that source uncancelable.  And if it is, instead of never completing, won't it just ignore the borrow?  It seems you're partially reinventing `timeoutTo`.

armanbilge:
No rush, FYI this is a blocker for blaze schism :)

rossabaker:
I can recreate this in blaze if we have to.

blaze-client/src/main/scala/org/http4s/blaze/client/PoolManager.scala
danicheg: After the 'cancelation in CE3' tutorial by Daniel, I'm sure I was wrong and actually was reinventing the wheel with `Ref` here, as you pointed out (if someone is interested, link to the discussion https://discord.com/channels/632277896739946517/632278585700384799/977634027651608626).
I think a few hot-takes should be highlighted. Also, this answers the question 'what makes the source uncancelable':
1. `async` is uncancelable during registration;
2. if one defines `async` without cancel token (i. e. using `None` instead of actual effect), then `async` will be uncancellable;
3. things created via `delay` is uncancellable too.

So summing up, I'm not sure that introducing timeout via `timeoutTo` will work as expected all the time. I think that for sure could be scenarios when 
1. the whole `async` effect will be considered uncancelable, 
2. and `timeoutTo` will wait for the natural execution of `async` then produce an exception (for timeout limit exceeding case). 

And I don't think that what we actually want is that behavior.

rossabaker: Right.  This never returns:

```scala
IO.uncancelable(_ => IO.never)
  .timeoutTo(1.second, IO.unit)
  .unsafeRunSync()
```

danicheg:
Superseded by https://github.com/http4s/blaze/pull/681.

PR #6388, opened 2022-05-18
Ember-Server: WebSocketHelper rewrite / optimisation

diesalbla:
- Reduce some `Pipe` methods to extract the logic as functions on the effect `F`. 
- Replace some calls to `through` with calls to Stream methods.
- Fuse the `encodeFrame` with the `write` method into a `writeFrame` function, that avoids building a new finite in-memory stream for each frame.  

_Note:_ half the diff is whitespace.

### DIFF
--- a/ember-server/shared/src/main/scala/org/http4s/ember/server/internal/WebSocketHelpers.scala
+++ b/ember-server/shared/src/main/scala/org/http4s/ember/server/internal/WebSocketHelpers.scala
@@ -125,7 +125,7 @@ object WebSocketHelpers {
             .through(decodeFrames[F])
             .through(handleIncomingFrames(write, close))
             .through(receiveSend)
-            .through(encodeFrames[F])
+            .flatMap(encodeFrame[F])
             .through(write) -> onClose
         case WebSocketSeparatePipe(send, receive, onClose) =>
           val closeFrame: F[Stream[F, WebSocketFrame]] = close.get.flatMap {
@@ -142,7 +142,7 @@ object WebSocketHelpers {
           val sendWithClose = send ++ Stream.eval(closeFrame).flatten
 
           val writer = sendWithClose
-            .through(encodeFrames[F])
+            .flatMap(encodeFrame[F])
             .through(write)
 
           val reader = incoming
@@ -166,7 +166,7 @@ object WebSocketHelpers {
       closeState: Ref[F, Close],
   )(implicit F: Concurrent[F]): Pipe[F, WebSocketFrame, WebSocketFrame] = {
     def writeFrame(frame: WebSocketFrame): F[Unit] =
-      Stream(frame).through(encodeFrames[F]).through(write).compile.drain
+      encodeFrame[F](frame).through(write).compile.drain
 
     stream =>
       stream.evalMapFilter[F, WebSocketFrame] {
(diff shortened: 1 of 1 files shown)


FrancescoSerra (approved):

ember-server/shared/src/main/scala/org/http4s/ember/server/internal/WebSocketHelpers.scala
```diff
@@ -121,32 +122,28 @@ object WebSocketHelpers {
...
-            .through(handleIncomingFrames(write, frameTranscoder, close))
+            .through(decodeFrames[F])
+            .evalMapFilter(handleIncomingFrame[F](writeFrame, close))
             .through(receiveSend)
-            .through(encodeFrames(frameTranscoder))
-            .through(write) -> onClose
+            .foreach(writeFrame) -> onClose
         case WebSocketSeparatePipe(send, receive, onClose) =>
-          val closeFrame: F[Stream[F, WebSocketFrame]] = close.get.flatMap {
+          val closeFrame: F[List[WebSocketFrame]] = close.get.flatMap {
```
armanbilge, line 130: Can we use an `Option` here?

diesalbla: I am afraid not, because the `evalSeq` below needs an `Iterable`, and an Option is not.

armanbilge: What if we just do `eval(...).unNone` or similar? `evalSeq` doesn't seem to be doing anything special.

```scala
  def evalSeq[F[_], S[A] <: Seq[A], O](fo: F[S[O]]): Stream[F, O] =
    eval(fo).flatMap(Stream.emits)
```

https://github.com/typelevel/fs2/blob/9e924ca24ef72d15b09825905ea29b344e4bd67b/core/shared/src/main/scala/fs2/Stream.scala#L3165-L3166

diesalbla: Just pushed a change to wrap the `sendFrame` into the method instead, and leave a `Strea.eval` for it.

ember-server/shared/src/main/scala/org/http4s/ember/server/internal/WebSocketHelpers.scala
```diff
@@ -121,32 +122,33 @@ object WebSocketHelpers {
...
+                _ <- writeFrame(endFrame)
+              } yield ()
+            case _ => F.unit
           }
-          val sendWithClose = send ++ Stream.eval(closeFrame).flatten
 
-          val writer = sendWithClose
-            .through(encodeFrames(frameTranscoder))
-            .through(write)
+          val writer = send.foreach(writeFrame) ++ Stream.eval(sendClosingFrame)
```
armanbilge, line 147:
```suggestion
          val writer = send.foreach(writeFrame) ++ Stream.exec(sendClosingFrame)
```

diesalbla: Fixed!

armanbilge (approved):
Thanks!

PR #6393, opened 2022-05-19
Use `parTraverse` in tests

danicheg:
A teeny-weeny speed-up of tests running.


rossabaker (approved):
Did you notice any speedup, or does this just seem like the right thing to do?  Either way, I like it.

danicheg:
Honestly, it's likely the same within the infelicity at local, probably 1-2s at most. But it could be more valuable at the CI runs.

Issue #5320, opened 2021-10-04
Create a published `testkit` module


armanbilge:
cc @amesgen. Linking to:
https://github.com/http4s/http4s-jdk-http-client/issues/398
https://github.com/http4s/http4s/discussions/4140#discussioncomment-265556

Now that we have two satellite modules `http4s-jdk-http-client` and `http4s-dom` and with the ever impending great schism this is only all the more important. 

AFAICT mainly `ClientRouteTestBattery` has been discussed ... I think it would be valuable to publish a testkit module for _both_ clients and servers. As I understand it, a "TestBattery" is a (`HttpApp`, `Suite`) pair (where a suite is a collection of tests) , with a pluggable server and pluggable client. Depending on what you want to test, ideally one of those should be a "control" 😁 so it seems to me without much additional work, a solid battery could be used for both?

Besides a battery, I wonder what else might be useful in a `testkit` module?

rossabaker:
Yes, I would imagine both a client and server version.  I suppose it would generally need to be blackbox: everything over a TCP socket.  Also is the suite small for the minimal things every backend must do, or large to expose what they can and cannot do?  The behavior will never be 100% aligned.
 
I think this is great and needed, but, whew, it's daunting.

armanbilge:
Thank you both for the feedback 😅 yes this is  _very_ daunting and I hope we can start small by publishing the `ClientRouteTestBattery` and the server scaffolds so that we can use them in `jdk-http-client` and the `FetchClient` in `http4s-dom`. Hope that @amesgen can weigh in too when they have a chance.

> In it's current shape, the assertions in these tests look into blaze, which makes the tests impossible to reuse for ember.

Yes, I did see that. I was initially thinking we could just abstract an interface by which the client backend could the necessary information (I'm not sure how hard it would be to implement this for ember). But, if this could be done with a special test server that sounds cool idea, tests relying on sniffing sounds complicated but I wouldn't know 😅

armanbilge:
The `ServerScaffold` introduced in the following two PRs can be the start of our test kit, it just needs to be extracted out of the `client` module into a new `client-testkit` module, if anyone would like to work on that? :)
- https://github.com/http4s/http4s/pull/5601
- https://github.com/http4s/http4s/pull/5587

armanbilge:
This issue blocks schisming of the various client backends.

PR #6394, opened 2022-05-19
Publish `http4s-client-testkit` module

armanbilge:
Closes https://github.com/http4s/http4s/issues/5320.

It makes sense to focus this only on clients, right? Or should it just be a generic `testkit` module that will maybe do servers too in the future?


build.sbt
```diff
@@ -13,6 +13,8 @@ ThisBuild / developers += tlGitHubDev("rossabaker", "Ross A. Baker")
 ThisBuild / tlCiReleaseBranches := Seq("series/0.23")
 ThisBuild / tlSitePublishBranch := Some("series/0.23")
 
+ThisBuild / tlCiReleaseBranches += "feature/client-testkit"
```
armanbilge, line 16: Temporarily snapshotting this branch for downstream experimentation.

build.sbt
```diff
@@ -404,6 +406,24 @@ lazy val client = libraryCrossProject("client")
...
 
+lazy val clientTestkit = libraryCrossProject("client-testkit")
+  .settings(
+    description := "Client testkit for building http4s clients",
+    startYear := Some(2014),
+    libraryDependencies ++= Seq(
+      munit.value,
+      munitCatsEffect.value,
+    ),
+    mimaPreviousArtifacts := Set.empty,
```
armanbilge, line 417: No bincompat guarantees.

client-testkit/shared/src/main/scala/org/http4s/client/testkit/ClientRouteTestBattery.scala
```diff
@@ -205,4 +209,23 @@ abstract class ClientRouteTestBattery(name: String) extends Http4sSuite with Htt
...
+  ): Fixture[IO[A]] =
+    registerSuiteFixture(UnsafeResourceSuiteLocalDeferredFixture(name, resource))
+
+  implicit private class ParseResultSyntax[A](self: ParseResult[A]) {
+    def yolo: A = self.valueOr(e => sys.error(e.toString))
+  }
+
+  private[this] lazy val suiteFixtures = List.newBuilder[Fixture[_]]
+
+  override def munitFixtures: Seq[Fixture[_]] = suiteFixtures.result()
```
armanbilge, line 229: Had to inline these to drop the http4s testing dependency.

build.sbt
rossabaker: That's what we usually do.  We can revisit if someone is testing a backend outside `org.http4s`, but I've never heard of one.

rossabaker:
A separate server-testkit in due time makes sense.  Server and client are themselves already separate.

armanbilge:
👍 working in the schismed repos.

client-testkit/shared/src/main/scala/org/http4s/client/testkit/ClientRouteTestBattery.scala
rossabaker: I don't love this.  It seems like most of them are munit concerns, and either are missing from munit or us using munit inappropriately.  But I think this is sensible for now.

rossabaker (approved):

armanbilge:
I'm fixing them now. The conflicts are because of the testkit package rename and you just merged my other PR deleting the various client backends :)

client-testkit/shared/src/main/scala/org/http4s/client/testkit/ClientRouteTestBattery.scala
armanbilge: Yep, agree. This weirdness gave me a heck of a time with with null pointer exceptions and initialization order issues.

PR #6395, opened 2022-05-19
Update to circe 0.14.2

armanbilge:
And a cherry on top!

### DIFF
--- a/project/Http4sPlugin.scala
+++ b/project/Http4sPlugin.scala
@@ -144,7 +144,7 @@ object Http4sPlugin extends AutoPlugin {
     val cats = "2.7.0"
     val catsEffect = "2.5.5"
     val catsParse = "0.3.7"
-    val circe = "0.14.1"
+    val circe = "0.14.2"
     val crypto = "0.1.0"
     val cryptobits = "1.3"
     val disciplineCore = "1.4.0"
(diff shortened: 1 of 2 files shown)


armanbilge:
Oh no, it needs Scala 3.1 :(

armanbilge:
Is this a failure to be concerned about 🤔 
```
==> X org.http4s.circe.CirceSensitiveDataEntityDecoderSpec.should not include the JSON when failing to decode due to wrong data type of JSON key's value  0.008s munit.ComparisonFailException: /home/runner/work/http4s/http4s/circe/src/test/scala/org/http4s/circe/CirceSensitiveDataEntityDecoderSpec.scala:59
58:          assertEquals(details, "Could not decode JSON: <REDACTED>")
59:          assertEquals(cause.getMessage, "String: DownField(ssn)")
60:        case other => fail(other.toString)
values are not the same
=> Obtained
Got value '123456789' with wrong type, expecting string: DownField(ssn)
-Got value '123456789' with wrong type, expecting string: DownField(ssn)
```

rossabaker:
Failure is probably this: https://github.com/circe/circe/pull/1617.  That was called out as a semantic change when they were debating calling the next release 0.14 vs. 0.15.

It would be weird to do a rugpull on Scala 3.0 users on the last release of the series.  They can bump to 0.14.2 themselves.

armanbilge:
Right, I was just concerned if the error is leaking information it should. Agree that we should not bump to 3.1.

rossabaker:
It's fine as long as no request information is in the sanitized field.  It can be in details.

PR #6396, opened 2022-05-20
Bye-bye async, okhttp, jetty clients

armanbilge:
- https://github.com/http4s/http4s-async-http-client/
- https://github.com/http4s/http4s-okhttp-client/
- https://github.com/http4s/http4s-jetty-client

This leaves only the blaze modules to be schismed I believe. What is the plan for those, will they get their own repo http4s/http4s-blaze or somehow get integrated into the existing http4s/blaze repo?


rossabaker:
Those sources should be migrated into the blaze repo.  Blaze as a standalone project hasn't had a champion in several years.

rossabaker (approved):

PR #6401, opened 2022-05-20
Update to circe 0.14.2

armanbilge:
Finishes what https://github.com/http4s/http4s/pull/6398 started :)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -845,14 +845,13 @@ lazy val circe = libraryCrossProject("circe", CrossType.Pure)
     startYear := Some(2015),
     libraryDependencies ++= Seq(
       circeCore.value,
+      circeJawn.value,
       circeTesting.value % Test,
     ),
   )
   .jsSettings(
     jsVersionIntroduced("0.23.5")
   )
-  .jvmSettings(libraryDependencies += circeJawn.value)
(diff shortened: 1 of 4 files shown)


build.sbt
```diff
@@ -845,14 +845,13 @@ lazy val circe = libraryCrossProject("circe", CrossType.Pure)
...
       circeCore.value,
+      circeJawn.value,
       circeTesting.value % Test,
     ),
   )
   .jsSettings(
     jsVersionIntroduced("0.23.5")
   )
-  .jvmSettings(libraryDependencies += circeJawn.value)
-  .jsSettings(libraryDependencies += circeJawn15.value)
```
armanbilge, line 855: So this is a bit weird ... http4s-circe.js is going from 0.15.0-M1 to 0.14.2. The good news is that AFAICT 0.14.2 is binary-compatible with 0.15.0-M1 and I also don't think we are relying on any new APIs in 0.14.2, so hopefully there shouldn't be any issues. Evictioning might get wonky though :(

rossabaker: That's life on a milestone.  The Circe team figured out that everything was bincompat after 0.15 had already started.

rossabaker (approved):

PR #6404, opened 2022-05-20
Make Node.js interop APIs private

armanbilge:
I initially intended to make these APIs public since they are helpful for integrating with serverless frameworks. After thinking more, I changed my mind.

Besides these interop APIs, there is essentially nothing else JS-specific in http4s, and I think that's great.

Furthermore, the only reason we need these interop APIs in http4s at all is to implement the Node.js scaffold server for the client testkit. I'd like to reserve the option to remove the need for them completely in the future: for example, by running the scaffold server as an external process via sbt (i.e., a testkit distributed as an sbt plugin).


rossabaker (approved):
I'm completely deferential on js, but the rationale sounds right.

PR #6403, opened 2022-05-20
Release v0.23.12

rossabaker:

PR #6405, opened 2022-05-20
Bye-bye blaze

armanbilge:
This is a bit premature, but will help unblock https://github.com/http4s/http4s/pull/6403.

The open PRs seem like they might need another cycle or two, so I'll start preparing a PR for the blaze repo with what's currently on series/0.23.

Edit: also, a blaze maintainer will have to transfer the issues.


rossabaker:
I can make you a blaze maintainer, but some would view that as more of a threat than an offer.

armanbilge:
Yep definitely a threat 😆 hmm yeah it's a lot of issues, surely somebody else would like to transfer them? Otherwise I will click as needed, and then unsubscribe.

armanbilge:
Voila:
- https://github.com/http4s/blaze/pull/628

rossabaker (approved):

PR #6407, opened 2022-05-23
Highlight the uncancelable behavior in `Timeout` middleware scaladoc

danicheg:
I think the uncancelable behavior is noteworthy because it could lead to some surprising effects on the user site. It's indisputable that it's the user's responsibility to check how their services produce effects. But still, I'm sure it's better to highlight that behavior. 
Also, I've added the test that proves my point about `Timeout` middleware behavior when dealing with uncancelable effects.

### DIFF
--- a/server/shared/src/main/scala/org/http4s/server/middleware/Timeout.scala
+++ b/server/shared/src/main/scala/org/http4s/server/middleware/Timeout.scala
@@ -32,13 +32,13 @@ object Timeout {
     * fires, the service's response is canceled.
     *
     * @note if the service runs uncancelable effects while responding
-    * (e.g. if service uses `MonadCancel#uncancelable` under the hood) and has exceeded
+    * (e.g. if service uses [[cats.effect.kernel.MonadCancel#uncancelable MonadCancel#uncancelable]] under the hood) and has exceeded
     * the timeout, then the expected behavior is:
     * <ul>
     *   <li> uncancelable effects will be completed naturally (regardless of how long it takes), </li>
     *   <li> after that, the timeout response will be returned. </li>
     * </ul>
-    * To get more insights on effect cancelation, dig into the Cats-Effect documentation.
+    * To get more insights on effect cancelation, dig into the [[cats.effect.kernel.MonadCancel MonadCancel]] documentation.
     *
     * @param timeout Finite duration to wait before returning the provided response
     */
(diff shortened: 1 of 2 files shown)


server/shared/src/main/scala/org/http4s/server/middleware/Timeout.scala
```diff
@@ -31,6 +31,15 @@ object Timeout {
...
     * fires, the service's response is canceled.
     *
+    * Note, if effects of service responding is uncancelable
+    * (e.g. if service uses `MonadCancel#uncancelable` under the hood) and has exceeded
+    * the timeout, then an expected behavior is:
+    * <ul>
+    *   <li> effects will be completed naturally (regardless of how long it takes), </li>
+    *   <li> after that, the timeout response will be returned. </li>
+    * </ul>
+    * To get more insights on effect canceling, dig into the Cats-Effect documentation.
```
armanbilge, line 41: It would be nice if we had a more specific link for this, but I suspect it doesn't exist! 😬

server/shared/src/main/scala/org/http4s/server/middleware/Timeout.scala
```diff
@@ -31,6 +31,15 @@ object Timeout {
     * duration if the service has not yet responded.  If the timeout
     * fires, the service's response is canceled.
     *
+    * Note, if effects of service responding is uncancelable
+    * (e.g. if service uses `MonadCancel#uncancelable` under the hood) and has exceeded
+    * the timeout, then an expected behavior is:
+    * <ul>
+    *   <li> effects will be completed naturally (regardless of how long it takes), </li>
```
armanbilge, line 38: I think there is some nuance here, because cancelable and uncancelable effects can be interleaved.

```suggestion
    *   <li> uncancelable effects will be completed naturally (regardless of how long it takes), </li>
```

server/shared/src/main/scala/org/http4s/server/middleware/Timeout.scala
armanbilge:
```suggestion
    * To get more insights on effect cancellation, dig into the Cats-Effect documentation.
```

armanbilge:
Thanks, sorry for the nitpicks. I think this is a really important topic to cover.

server/shared/src/test/scala/org/http4s/server/middleware/TimeoutSuite.scala
```diff
@@ -76,6 +81,18 @@ class TimeoutSuite extends Http4sSuite {
...
+    "return a 503 error if the result takes too long and execute underlying uncancelable effect"
+  ) {
+    testMiddleware(5.milliseconds) { app =>
+      for {
+        _ <- app(uncancelableReq).map(_.status).assertEquals(Status.ServiceUnavailable)
+        _ <- checkStatus(app(uncancelableReq), Status.ServiceUnavailable)
+          .intercept[TimeoutException]
+      } yield ()
+    }
+  }
```
armanbilge, line 94: Sorry, I'm a little confused what this test is demonstrating?

armanbilge: Ah, thank you for explaining, I think I finally understood. The test for (1) is the `.intercept[TimeoutException]` because `1.seconds < IO.sleep(1100.milliseconds)` IIUC.

server/shared/src/main/scala/org/http4s/server/middleware/Timeout.scala
danicheg: So the whole thing will look like this:

> Note, if the service runs uncancelable effects while responding (e.g. if service uses `MonadCancel#uncancelable` under the hood) and has exceeded the timeout, then the expected behavior is:
> - uncancelable effects will be completed naturally (regardless of how long it takes),
> - after that, the timeout response will be returned.

I think I like it.

server/shared/src/test/scala/org/http4s/server/middleware/TimeoutSuite.scala
danicheg: Yeah, it so, you're correct.

armanbilge (approved):
Looks good! I hope you don't mind, I added links to the CE API docs and confirmed that they work as expected.

PR #6409, opened 2022-05-24
Tweak `Timeout` middleware scaladoc

danicheg:
These references in scaladoc look really monstrous. And their inlining doesn't work, at least in my $IDE.


armanbilge:
Note that I'd like to re-enable linking warnings in https://github.com/http4s/http4s/pull/4027. So please let's make sure not to get a regression in that 🙏

armanbilge (approved):
Checked locally and miraculously it still links!

PR #6410, opened 2022-05-24
Cleanup unnecessary projects

armanbilge:
An attempt to make loading http4s in sbt a bit faster by cutting down the number of projects. This PR gets us from 52320 settings to 47283.

1. Move `testing` project into `Compile` scope of `tests` project.
2. Delete the ember.js H2 examples. They were mostly to prove it actually worked, in lieu of tests...


rossabaker (approved):
Looks good, but I'd like to sit on it until 0.23.12 is tagged.

Issue #619, opened 2016-04-30
Server Sent Events


rossabaker:
We should have a codec for text/event-stream.

rossabaker:
We don't currently a decoder to a `Process`.  We can decode the entity to a `Task[Process[Task,_]]`, but, eh?

agourlay:
I am really interested in receiving `ServerSentEvent` on the client side.
My use case is to connect to an `SSE` endpoint an gather all events for a given duration.
What is missing to achieve that?
I am willing to offer help to push this forward 😃

rossabaker:
First off, since the above comment is old, `Process` is `Stream` and `Task[Process[Task, _]]` is `F[Stream[F, _]]`.  But there still exists no `EntityDecoder` to a `Stream`.  And maybe there doesn't need to be: we can probably build on `response.body.through(ServerSentEvent.decoder)` to get our `Stream[F, ServerSentEvent]`.  At a simple level, that might just work.

At a more complete level, it's worth looking at the [Processing model](https://www.w3.org/TR/2009/WD-eventsource-20091029/#processing-model).  Some servers might be strict in checking things like the `Accept` header.  Maybe a `def streamServerSentEvents(uri: Uri, lastEventId: Option[LastEventId]): Stream[F, ServerSentEvent]` method on the client that set all those headers and checked status codes appropriately would be a good idea?

kalexmills:
I'd like to use comment lines to provide a keep-alive notification and keep the browser from shutting down existing streams. 

Per [the spec](https://www.w3.org/TR/2009/WD-eventsource-20090421/#event-stream-interpretation) lines starting with `:` are 'comments' and are ignored by the client, but should serve to keep the socket open.

Are folks open to adding functionality for sending comments to the implementation? My alternative right now is to send empty data, which is just 5 bytes too many. ;-)

rossabaker:
Setting aside the comments issue we overlooked (:disappointed:), I think the client would be neat.  It's probably too esoteric for the main `Client` interface, but can't go on core.  A new `SseClient` that wraps a `Client` would be a good home for implementing this spec.

PR #6413, opened 2022-05-25
Add `EntityDecoder[EventStream]`

armanbilge:
All the pieces were already here, including a test, so not sure why this wasn't already exposed.

### DIFF
--- a/core/shared/src/main/scala/org/http4s/EntityDecoder.scala
+++ b/core/shared/src/main/scala/org/http4s/EntityDecoder.scala
@@ -355,6 +355,11 @@ object EntityDecoder {
   ): EntityDecoder[F, Multipart[F]] =
     MultipartDecoder.mixedMultipart(headerLimit, maxSizeBeforeWrite, maxParts, failOnLimit)
 
+  implicit def serverSentEventDecoder[F[_]: Applicative]: EntityDecoder[F, EventStream[F]] =
+    EntityDecoder.decodeBy(MediaType.`text/event-stream`) { msg =>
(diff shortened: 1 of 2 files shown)


rossabaker (approved):
I thought there was a reason, but I didn't document it and don't remember it.

armanbilge:
@rossabaker I found some discussion on this in https://github.com/http4s/http4s/issues/619, I still need to digest it but does anything there jog your memory?

rossabaker:
I guess it was having a decoder that produced the prehistoric equivalent of `F[Stream[F, Event]]` was distasteful?  You probably want to `Stream.eval` the decoded result from this?

armanbilge:
Or [`Stream.force`](https://github.com/typelevel/fs2/blob/ff9d758ae61cb19ad4d6f5cde4d0b236efab8157/core/shared/src/main/scala/fs2/Stream.scala#L3499) maybe? I did notice that, and it's slightly awkward, but I can't see why it's that big a deal.

In theory we could add a special syntax for it roundabouts here:
https://github.com/http4s/http4s/blob/e94c9d48fec00758d70afa59107408463d89e818/core/shared/src/main/scala/org/http4s/Media.scala#L67

e.g.
```scala
final def asStream[A](implicit F: MonadThrow[F], decoder: EntityDecoder[F, Stream[F, A]]): Stream[F, A] =
  Stream.force(as[A])
```

Otherwise not really sure what we could do about it.

armanbilge:
> I thought there was a reason, but I didn't document it and don't remember it.

Well, we figured out why 😕 
- https://github.com/http4s/http4s/issues/6694

PR #6422, opened 2022-05-28
Ember-Core - H2Client - Rewrite without alternative.

diesalbla:
In this code, the use of `Alternative[Option]` can be replaced with an if-then-else statement. Furthermore, we can replace the for-comprehension with a simple flatMap and less inlining.

_Note_: most changes are whitespace.

### DIFF
--- a/ember-core/shared/src/main/scala/org/http4s/ember/core/h2/H2Client.scala
+++ b/ember-core/shared/src/main/scala/org/http4s/ember/core/h2/H2Client.scala
@@ -361,26 +361,21 @@ private[ember] object H2Client {
     } yield (http1Client: TinyClient[F]) => { (req: Request[F]) =>
       val key = H2Client.RequestKey.fromRequest(req)
       val priorKnowledge = req.attributes.lookup(H2Keys.Http2PriorKnowledge).isDefined
-      val h2Prior = Alternative[Option].guard(priorKnowledge).as(Http2)
-      for {
-        socketType <- Resource.eval(
-          socketMap.get.map(_.get(key))
-        )
-        resp <- h2Prior.orElse(socketType) match {
-          case Some(Http2) => h2.runHttp2Only(req)
-          case Some(Http1) => http1Client(req)
-          case None =>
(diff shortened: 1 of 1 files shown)


armanbilge (approved):

ember-core/shared/src/main/scala/org/http4s/ember/core/h2/H2Client.scala
```diff
@@ -361,26 +361,21 @@ private[ember] object H2Client {
     } yield (http1Client: TinyClient[F]) => { (req: Request[F]) =>
       val key = H2Client.RequestKey.fromRequest(req)
       val priorKnowledge = req.attributes.lookup(H2Keys.Http2PriorKnowledge).isDefined
-      val h2Prior = Alternative[Option].guard(priorKnowledge).as(Http2)
-      for {
-        socketType <- Resource.eval(
```
danicheg, line 366: Whoa, it seems `socketType` has been computing regardless of the guard's result. This refactoring could also bring saving a few processor ticks. 👏🏻

danicheg (approved):

ember-core/shared/src/main/scala/org/http4s/ember/core/h2/H2Client.scala
```diff
@@ -361,26 +361,21 @@ private[ember] object H2Client {
...
+        case Some(Http2) => h2.runHttp2Only(req)
+        case Some(Http1) => http1Client(req)
+        case None =>
+          (
+            h2.runHttp2Only(req) <*
+              Resource.eval(socketMap.update(s => s + (key -> Http2)))
+          ).handleErrorWith[org.http4s.Response[F], Throwable] {
+            case InvalidSocketType() | MissingHost() | MissingPort() =>
+              Resource.eval(socketMap.update(s => s + (key -> Http1))) >>
+                http1Client(req)
```
diesalbla, line 375: @armanbilge Thanks for fixing this, just pushed a full change.

PR #6424, opened 2022-05-29
Ember-Core H2Server: replace pull with a recursive function

diesalbla:
The last part of the `fromSocket` is just a loop to keep the "connection" open until the state in `stateRef` is closed. This was implemented with a Pull and streams, but it can be written as a one-line self-recursive loop.


danicheg (approved):
LGTM.

armanbilge (approved):

PR #6426, opened 2022-05-29
Ember-Client: no Alternative

diesalbla:
The use of `>>` and guard from Alternative type-class can be replaced with basic if-then-else Scala construct.

### DIFF
--- a/ember-client/shared/src/main/scala/org/http4s/ember/client/EmberClientBuilder.scala
+++ b/ember-client/shared/src/main/scala/org/http4s/ember/client/EmberClientBuilder.scala
@@ -32,6 +32,7 @@ import org.http4s.client.middleware.Retry
 import org.http4s.client.middleware.RetryPolicy
 import org.http4s.ember.client.internal.ClientHelpers
 import org.http4s.ember.core.h2.H2Client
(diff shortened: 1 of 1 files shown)


ember-client/shared/src/main/scala/org/http4s/ember/client/EmberClientBuilder.scala
```diff
@@ -253,20 +254,16 @@ final class EmberClientBuilder[F[_]: Async] private (
...
-            logger,
-            if (pushPromiseSupport.isDefined) default
-            else {
-              default.copy(enablePush =
-                org.http4s.ember.core.h2.H2Frame.Settings.SettingsEnablePush(false)
-              )
-            },
-          )
-      )
+      optH2 <- (if (enableHttp2) tlsContextOptWithDefault else None).traverse { context =>
```
danicheg, line 257: Is it the case when `Applicative#when` fits well?

diesalbla: No, because this yields an `Option`, and the `when` is for `F[Unit]`. In any case, that is still less straightforward than `if-then-else`.

armanbilge (approved):

PR #6431, opened 2022-06-03
Update to Vault 3.2.1

armanbilge:
Not sure what the long-term plans are, but FTR it would be very easy to whip up an http4s-steward. See: https://github.com/armanbilge/diy-steward

### DIFF
--- a/project/Http4sPlugin.scala
+++ b/project/Http4sPlugin.scala
@@ -131,7 +131,7 @@ object Http4sPlugin extends AutoPlugin {
     val slf4j = "1.7.36"
     val treehugger = "0.4.4"
     val twitterHpack = "1.0.2"
-    val vault = "3.1.0"
+    val vault = "3.2.1"
   }
 
   lazy val blazeCore = "org.http4s" %% "blaze-core" % V.blaze


rossabaker:
There is a [typelevel-steward](https://github.com/settings/apps/typelevel-steward) for Nix flakes, which is nice because it's from @typelevel-bot and not a person.  Is there a good reason we couldn't just install that here?

armanbilge:
Is that a "true" bot (e.g. managed by the org admins) or just a user account made for botlike purposes?

by "true bot" I mean something like this:
https://github.com/apps/lucuma-steward
https://github.com/gemini-hlsw/lucuma-svgdotjs/pull/105

rossabaker:
@typelevel-bot is a GitHub account with a shared login.  It created the typelevel-steward app [corrected link](https://github.com/apps/typelevel-steward).  I originally called it typelevel-flake-steward.

I guess we'd want separate apps in case flake updates and Scala updates ever needed distinct permissions, but it seems unlikely: they just wake up and open PRs.

armanbilge:
> @typelevel-bot is a GitHub account with a shared login. It created the typelevel-steward app [corrected link](https://github.com/apps/typelevel-steward).

IMHO this seems a bit arcane. Seems that typelevel-bot should transfer the typelevel-steward app to the org and retire.

rossabaker:
I don't remember org apps being an option when I set that up, but it sounds fine now.  I'll work on it after lunch for Typelevel and http4s.

PR #6432, opened 2022-06-03
Add a Scala Steward workflow

rossabaker:
Uses organization secrets and the [typelevel-steward](https://github.com/apps/typelevel-steward) app.

### DIFF
--- a/.github/workflows/steward.yml
+++ b/.github/workflows/steward.yml
@@ -1,0 +1,30 @@
+name: Scala Steward
+
+on:
+  schedule:
+    - cron: '0 */8 * * *'
+  workflow_dispatch:
+
+jobs:
+  steward:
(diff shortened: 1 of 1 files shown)


rossabaker:
I think I'm going to have to merge this to run it.

.github/workflows/steward.yml
```diff
@@ -0,0 +1,30 @@
...
+    runs-on: ubuntu-latest
+    name: DIY Steward
+    steps:
+      - uses: actions/checkout@v3
+          
+      - name: Generate token
+        id: generate-token
+        uses: tibdex/github-app-token@v1
+        with:
+          app_id: 123456
```
armanbilge, line 19: This should be `${{ secrets.STEWARD_APP_ID }}` I think

PR #6433, opened 2022-06-03
Make Scala Steward workflow work

rossabaker:
#6432 was necessary to start it running, now can iterate here until it works...

### DIFF
--- a/.github/workflows/steward.yml
+++ b/.github/workflows/steward.yml
@@ -16,7 +16,7 @@ jobs:
         id: generate-token
         uses: tibdex/github-app-token@v1
         with:
-          app_id: 123456
+          app_id: ${{ secrets.STEWARD_APP_ID }}
           private_key: ${{ secrets.STEWARD_PRIVATE_KEY }}
 
       - name: Launch Scala Steward
--- a/.github/workflows/steward.yml
+++ b/.github/workflows/steward.yml
@@ -24,7 +24,7 @@ jobs:
         with:
           github-token: ${{ steps.generate-token.outputs.token }}
           branches: series/0.23,main
-          author-email: ${{ secrets.STEWARD_APP_ID }}+typelevel-steward[bot]@users.noreply.github.com
+          author-email: 106827141+typelevel-steward[bot]@users.noreply.github.com
           author-name: typelevel-steward[bot]
           github-app-id: ${{ secrets.STEWARD_APP_ID }}
           github-app-key: ${{ secrets.STEWARD_PRIVATE_KEY }}


armanbilge:
Pros/cons of running it out of this repo vs giving it its own http4s-steward repo, that we can coordinate the entire org from?

armanbilge:
Aha, I see it's running for the whole org :)

rossabaker:
I thought this action only ran per repo until I started watching the output.  (Which is really surprising to me, given the branches configuration option.  I need to RTFM.)

I think it's weird to run anything orgwide out of this repo.  This probably ought to move if that's how it works.

armanbilge:
It runs on every repo the app is installed on.

rossabaker:
Yep, and respects a repos.md if anyone wants to opt out.  But then I think we'd all have to opt in, bootstrapped from the old scala-steward-org config or something.

I think we should move this, but I'm afraid it's going to generate duplicate PRs.

armanbilge:
> Yep, and respects a repos.md if anyone wants to opt out. But then I think we'd all have to opt in, bootstrapped from the old scala-steward-org config or something.

Right. Also I don't know how to do multiple branches without a `repos.md`. But maybe that can be fixed in upstream steward.

> I think we should move this, but I'm afraid it's going to generate duplicate PRs.

Possibly, but if we merge them first that will help :)

armanbilge:
@rossabaker also note that with the current configuration, anyone can install https://github.com/apps/typelevel-steward on their repo and mooch our CI for their updates!

rossabaker:
Can they mooch without the private key?  🤔

armanbilge:
Yes, because it's only needed in the repo where the workflow is actually running. Every other repo is just getting PRs, nothing is actually running there.

rossabaker:
I'm still processing the threat model here.

-- 

1. http4s installs `typelevel-steward`
2. http4s has connections at Typelevel and got a private key.
3. Workflow runs as http4s and opens a bunch of PRs against http4s as http4s.  

Cool.

--

1. Adversary installs `typelevel-steward`.
2. Adversary creates workflow but doesn't have Typelevel private key.
3. Workflow runs in adversary's repo under adversary's quota and fails.  Womp, womp.

Cool?

--

1. Adversary creates `adversary-steward`
2. Adversary configures a repos.md full of typelevel repos 
3. Workflow runs as adversary
4. `adversary-steward` isn't installed on typelevel and has no permissions to DoS us?

Cool?

rossabaker:
I removed it from http4s so I could make it private while we sort this out... but I _think_ it's okay?  Worst case, we make an http4s-steward, too.  I just thought it would be nice to have one point of config for the same purpose for affiliated orgs.

.github/workflows/steward.yml
```diff
@@ -16,15 +16,15 @@ jobs:
...
+          app_id: ${{ secrets.STEWARD_APP_ID }}
           private_key: ${{ secrets.STEWARD_PRIVATE_KEY }}
 
       - name: Launch Scala Steward
         uses: scala-steward-org/scala-steward-action@v2
         with:
           github-token: ${{ steps.generate-token.outputs.token }}
           branches: series/0.23,main
-          author-email: ${{ secrets.STEWARD_APP_ID }}+typelevel-steward[bot]@users.noreply.github.com
+          author-email: 106827141+typelevel-steward[bot]@users.noreply.github.com
```
rossabaker, line 27: That's annoying.  I had to do that in my Nix stuff, too.  And it opens dozens of PRs this way before we correct it.

.github/workflows/steward.yml
```diff
@@ -16,15 +16,15 @@ jobs:
...
         with:
-          app_id: 123456
+          app_id: ${{ secrets.STEWARD_APP_ID }}
           private_key: ${{ secrets.STEWARD_PRIVATE_KEY }}
 
       - name: Launch Scala Steward
         uses: scala-steward-org/scala-steward-action@v2
         with:
           github-token: ${{ steps.generate-token.outputs.token }}
           branches: series/0.23,main
```
rossabaker, line 26: I don't know what this means if it's not a per-repo setting.

armanbilge:
I'm not sure I understand your scenarios. But here's what happens.
1. Adversary installs typelevel-steward app.
2. The workflow runs, not in any repo except the steward repo.
3. typelevel-steward app gets list of repos to update by checking where it's installed. This includes adversary repo.
4. typelevel-steward starts updating the adversary.

.github/workflows/steward.yml
```diff
@@ -16,15 +16,15 @@ jobs:
...
-          author-email: ${{ secrets.STEWARD_APP_ID }}+typelevel-steward[bot]@users.noreply.github.com
+          author-email: 106827141+typelevel-steward[bot]@users.noreply.github.com
           author-name: typelevel-steward[bot]
           github-app-id: ${{ secrets.STEWARD_APP_ID }}
           github-app-key: ${{ secrets.STEWARD_PRIVATE_KEY }}
```
armanbilge, line 30: This is the configuration that enables steward to automatically get a list of repositories to update. If we remove it, then we have to use the `repos.md`. This has the advantage that it supports branches and is adversary-proof, since we just won't add them to the list. Any repo that wants updates needs to install the app and add itself to the list.

rossabaker:
Oh, I see.  Without a repos.md, it's going to pick up that it's installed in adversary's repos and clone those projects and send them PRs under our quota (not sure if it can open them...) ... or, worse, initialize malicious plugins in their builds as us.

armanbilge:
> it's going to pick up that it's installed in adversary's repos and clone those projects and send them PRs under our quota (not sure if it can open them...)

When you install the app in a repo, you are doing two things: (1) you are signing up for updates and (2) you are giving the app permission to create branches, PRs, commits, etc. in your repo.

armanbilge:
Btw, one thing to consider about having an org-level app/bot/steward is we don't need to arbitrate who gets on the update list: either the repo is in the org, or its not.

rossabaker:
Yeah, I was thinking about offering up a Typelevel one with a repos.md that is open to affiliate projects, but that's a human-curated list.  It lowers the barrier to entry a bit for friendly repos, but it's something more for Typelevel admins to deal with.

Should we hurry up and close today's PRs before anyone merges them with its incorrect contributor ID?  I guess it doesn't matter that much, but it's untidy.

armanbilge:
> Should we hurry up and close today's PRs before anyone merges them with its incorrect contributor ID? I guess it doesn't matter that much, but it's untidy.

Oh no, sorry, I already merged a bunch. I just didn't think it's that big of a deal 😬

rossabaker:
No worries.  We're not squash committers, either.  It was already a mess. :smile:

rossabaker:
I'm having a dumb on https://github.com/http4s/steward-playground.

The first build tries to update another org's repo.  That's a test for the idea of a Typelevel action taking care of affiliates.

The second build tries simply to update one of ours.  I'm getting a 404 on both.  I probably configured the app wrong, but I don't see it.

rossabaker:
My 404 was because I had an incorrect app ID. :facepalm:

1. It works on an in-org repo. 👍 
2. It gets a 403 trying to PR another org's repo before the app installed on that repo. 👍 
3. It gets a 403 trying to PR another org's repo after the app is installed. 😕

rossabaker:
Okay, I think the 403 happens going cross org because, even if the app is installed in both places, each org has a different "installation id" to which the GITHUB_TOKEN is scoped.  I don't think Typelevel can solve this for its affiliate projects, but we have a reasonable path forward at the org level.

PR #6437, opened 2022-06-04
Workflow to update flake weekly

rossabaker:


.github/workflows/update-flake.yml
```diff
@@ -0,0 +1,26 @@
+name: update-flake
+on:
+  workflow_dispatch: # allows manual triggering
+  schedule:
+    - cron: '45 7 * * 2' # runs weekly on Tuesday at 07:45
```
armanbilge, line 5: Question: I assume this should only run on series/0.23, and we merge changes into `main`? Should we restrict the branch this workflow runs on then?

rossabaker: Yeah, merging it forward makes the most sense.  If the flake ever had different sources, that would be awkward, but that's also unlikely to happen.

Will it not run by default on the default branch?  I'm not sure what the action is here.

rossabaker:
This is something that ideally ought to be in all our repositories with a nix flake.  Unlike Scala Steward, we don't yet have the tech to run the PR from a central "steward" repo.  That would require some sort of mashup of the Scala Steward repo iterator and the Flake Lock Update flake lock updater.  Kind of outside my wheelhouse.

.github/workflows/update-flake.yml
armanbilge: I kinda suspect the workflow may run at the scheduled time on every branch it exists, unless told otherwise. But gotta check the docs

rossabaker: Weekly might be aggressive when we think about a couple dozen repos with a flake.  It hasn't run here in five months and I haven't been annoyed.  I care very much that it's here and only a little that it's fresh.

armanbilge:
Linking to:
- https://github.com/typelevel/sbt-typelevel/issues/109

.github/workflows/update-flake.yml
armanbilge: > Scheduled workflows run on the latest commit on the default or base branch.

Nope, I was wrong. 👍 I don't mind it running weekly, it's just a drop in the bucket /shrug

armanbilge (approved):

danicheg (approved):

rossabaker:
The linked ticket reminds me this really ought to have CI for the dev shell...

Issue #6427, opened 2022-05-30
AuthMiddleware with Ember backend returning chunked http response


christiankjaer:
The default `AuthMiddleware` (returning an empty 401) in combination with the Ember backend will return a chunked response that will leave the socket open.

Minimal example with version 1.0.0M33, scala 3.1.2 and the ember backend:
```scala
import cats._, cats.effect._, cats.implicits._, cats.data._
import cats.effect._
import com.comcast.ip4s._
import org.http4s.implicits._
import org.http4s.ember.server._
import org.http4s._
import org.http4s.dsl.io._
import org.http4s.server._

val helloWorldService = AuthedRoutes.of[Unit, IO] {
  case GET -> Root as _ =>
    Ok()
}

val authUser: Kleisli[[X] =>> OptionT[IO, X], Request[IO], Unit] =
  Kleisli(_ => OptionT(IO(None)))

val middleware: AuthMiddleware[IO, Unit] =
  AuthMiddleware(authUser)

object Main extends IOApp {
    def run(args: List[String]): IO[ExitCode] =
      EmberServerBuilder
      .default[IO]
      .withHost(ipv4"0.0.0.0")
      .withPort(port"8080")
      .withHttpApp(middleware(helloWorldService).orNotFound)
      .build
      .use(_ => IO.never)
      .as(ExitCode.Success)
}
```
Testing with cURL:
```
$ curl -v localhost:8080
*   Trying 127.0.0.1:8080...
* Connected to localhost (127.0.0.1) port 8080 (#0)
> GET / HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.81.0
> Accept: */*
>
* Mark bundle as not supporting multiuse
< HTTP/1.1 401 Unauthorized
< Date: Mon, 30 May 2022 13:36:16 GMT
< Connection: keep-alive
< Transfer-Encoding: chunked
<
```
And then the connection is just left open.

christiankjaer:
Just by inlining what the auth middleware is doing, then the following handler will have the same behavior:
```scala
val service = HttpRoutes.of[IO] { case GET -> Root =>
  IO.pure(Response[IO](Status.Unauthorized))
}
```

danicheg:
[...]

PR #6444, opened 2022-06-04
Encoding of response with empty body

christiankjaer:
Fixes https://github.com/http4s/http4s/issues/6427

Around the repository there are different ways that empty responses are encoded. In some places the content length is set to 0, and in other places like the auth middleware, an empty response is returned with no headers set.

Here is a potential fix where an empty body with `content-length: 0` is always set when the status code allows a non-empty response and the header is not already set.

Something similar could probably also be done in the request encoding as well.

### DIFF
--- a/ember-core/shared/src/main/scala/org/http4s/ember/core/Encoder.scala
+++ b/ember-core/shared/src/main/scala/org/http4s/ember/core/Encoder.scala
@@ -30,7 +30,8 @@ private[ember] object Encoder {
 
   private val SPACE = " "
   private val CRLF = "\r\n"
-  val chunkedTansferEncodingHeaderRaw = "Transfer-Encoding: chunked"
+  val chunkedTransferEncodingHeaderRaw = "Transfer-Encoding: chunked"
+  val zeroContentLengthRaw = "Content-Length: 0"
 
   def respToBytes[F[_]](resp: Response[F], writeBufferSize: Int = 32 * 1024): Stream[F, Byte] = {
     var chunked = resp.isChunked
@@ -63,8 +64,12 @@ private[ember] object Encoder {
           ()
         }
       }
-      if (!chunked && !appliedContentLength && resp.status.isEntityAllowed) {
-        stringBuilder.append(chunkedTansferEncodingHeaderRaw).append(CRLF)
+      if (!appliedContentLength && resp.body == EmptyBody && resp.status.isEntityAllowed) {
+        stringBuilder.append(zeroContentLengthRaw).append(CRLF)
+        chunked = false
+        ()
+      } else if (!chunked && !appliedContentLength && resp.status.isEntityAllowed) {
+        stringBuilder.append(chunkedTransferEncodingHeaderRaw).append(CRLF)
         chunked = true
         ()
       }
@@ -135,7 +140,7 @@ private[ember] object Encoder {
         }
 
         if (!chunked && !appliedContentLength && !NoPayloadMethods.contains(req.method)) {
-          stringBuilder.append(chunkedTansferEncodingHeaderRaw).append(CRLF)
+          stringBuilder.append(chunkedTransferEncodingHeaderRaw).append(CRLF)
           chunked = true
           ()
         }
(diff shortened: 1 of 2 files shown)


ember-core/shared/src/main/scala/org/http4s/ember/core/Encoder.scala
```diff
@@ -63,8 +64,12 @@ private[ember] object Encoder {
           ()
         }
       }
-      if (!chunked && !appliedContentLength && resp.status.isEntityAllowed) {
-        stringBuilder.append(chunkedTansferEncodingHeaderRaw).append(CRLF)
+      if (!appliedContentLength && resp.body == EmptyBody && resp.status.isEntityAllowed) {
```
rossabaker, line 67: The logic here looks correct, but it's observable behavior that's dependent on reference equality of a stream.  They're both "correct" HTTP, and this is a better outcome for a common case.  I think I'm okay with this, but calling it out.

In main, with the new entity model, we can target the Empty entity and not have this concern.

rossabaker (approved):
:+1:, but I'd like to hear someone else's thought on my worry below.

ember-core/shared/src/main/scala/org/http4s/ember/core/Encoder.scala
christiankjaer: I looked at it, and I could not see any nice way to check whether a fs2 stream is empty. Maybe somebody can enlighten me.

christiankjaer: I am still confused why it was broken with an empty body and the chunked encoding.

rossabaker: There really isn't a good way.  We can "peek" a stream, but that semantically blocks until the first chunk of the body (if any) is produced, and would delay rendering the headers.  In scalaz-stream, we used to be able to introspect the algebra and find a wider class of empty streams, but still not all off them.  Effects are, well, effectful.

christiankjaer: Ah ok. Makes sense :+1:

ember-core/shared/src/test/scala/org/http4s/ember/core/EncoderSuite.scala
```diff
@@ -125,6 +126,32 @@ class EncoderSuite extends Http4sSuite {
...
+    val expected =
+      """HTTP/1.1 200 OK
+      |Transfer-Encoding: chunked
+      |
+      |0
+      |
+      |""".stripMargin
+
+    Helpers.encodeResponseRig(resp).assertEquals(expected)
+  }
```
christiankjaer, line 153: So is this intended behavior? I guess it's the only implementation that makes sense.

rossabaker: Yeah, I don't think we can do better than this.  And even though it's arguably overspecified, I like the test to alert us to behavior changes.

rossabaker:
This will need an import sort.  I think `ember-core/scalafix --all` will do it.

rossabaker (approved):
:+1: if green

christiankjaer:
Only question is whether to actually have `main` as the base branch, since I observe the bug only on the 1.0 pre release.

rossabaker:
Do those tests pass on this branch without your change?  I think only one of them would?

We will want to tweak this on the merge to main, where we can detect empty entities (it's a sum type, so no FS2 introspection is involved).

christiankjaer:
Yes. Only one will pass.

rossabaker:
I'd say this is correctly targeted for 0.23 then.

ember-core/shared/src/main/scala/org/http4s/ember/core/Encoder.scala
armanbilge: > I think I'm okay with this, but calling it out.

Seems fine to me.

armanbilge (approved):
Nice work!

PR #6450, opened 2022-06-05
Added note on possible circe import error in docs

cgoldammer:
Added a note that helped resolve a `circe` import error I encountered.

This was the solution proposed by armanbilge on Discord. In my case, I set up the project with `sbt new http4s/http4s.g8 --branch 0.23-scala3` and the import `import io.circe._` will work with Scala 3.1.0, but not 3.1.2. The proposed change allows the import to run.

### DIFF
--- a/docs/docs/json.md
+++ b/docs/docs/json.md
@@ -44,7 +44,7 @@ import org.http4s.dsl.io._
 import org.http4s.implicits._
 ```
 
-Note: Under some setup options (we saw this in Scala 3.1.2 but not in 3.1.0), running `import io.circe._` might give you an error of `-- [E008] Not Found Error: value circe is not a member of io`. This means that `io` is set from some other package, and you can run `import _root_.io.circe._`.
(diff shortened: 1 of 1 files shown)


docs/docs/json.md
```diff
@@ -44,6 +44,8 @@ import org.http4s.dsl.io._
 import org.http4s.implicits._
 ```
 
+Note: Under some setup options (we saw this in Scala 3.1.2 but not in 3.1.0), running `import io.circe._` might give you an error of `-- [E008] Not Found Error: value circe is not a member of io`. This means that `io` is set from some other package. In that case, you should run `import _root_.io.circe._`, analogously for the other `circe` imports.
+
```
armanbilge, line 48: This will make it render in an "info box" :)

```suggestion
@:callout(info)

Under some setup options (we saw this in Scala 3.1.2 but not in 3.1.0), running `import io.circe._` might give you an error of `-- [E008] Not Found Error: value circe is not a member of io`. This means that `io` is shadowed from some other package. In that case, you should run `import _root_.io.circe._`, analogously for the other `circe` imports.

@:@
```

armanbilge (approved):
Thanks for contributing to http4s!

PR #6451, opened 2022-06-05
Backport of `Router#define` test

danicheg:
[As we agreed](https://github.com/http4s/http4s/pull/6371#pullrequestreview-996021371), this is a backport of the Router's empty path case test.


armanbilge (approved):
Thank you!

PR #6456, opened 2022-06-06
Setup scoverage

armanbilge:
Following from https://github.com/http4s/http4s/discussions/6453.


armanbilge:
Results available at https://app.codecov.io/gh/http4s/http4s/branch/topic%2Fscoverage

rossabaker (approved):
Nice work.  That coverage job might also be a neat flag in sbt-http4s-org.  The one project I've really wanted this on in the last year was servlet, and it was just over the weekend!

PR #6457, opened 2022-06-06
Run coverage job for PRs, but don't upload results

armanbilge:
Sorry for the follow-up, realized I overlooked this subtlety. Occasionally the scoverage plugin will break for some otherwise valid code (like the `\r\n` thing). A PR could do this by mistake and we wouldn't find out till merge.

So actually, we _do_ want to run coverage for PRs, just to make sure it still works. We just don't want to upload the results!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -408,7 +408,6 @@ jobs:
 
   coverage:
     name: Generate coverage report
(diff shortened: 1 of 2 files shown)


armanbilge:
Cool. This PR from my fork also confirms that we won't get a coverage report here.

rossabaker (approved):

PR #6464, opened 2022-06-11
Point upgrade docs at scalafix published in series/0.22

armanbilge:
As discussed in https://github.com/http4s/http4s/pull/6460#issuecomment-1152525769. Confirmed that this works locally.

See also: https://scalacenter.github.io/scalafix/docs/developers/tutorial.html#using-github


FrancescoSerra (approved):
:+1:

rossabaker (approved):

Issue #6467, opened 2022-06-14
Use `TestControl.executeEmbed` for time-based tests


armanbilge:
I noticed a flake in the server `TimeoutSuite` yesterday and realized that many/all of our time-based tests are actually running against real time, I believe.

In Cats Effect 3.3 `TestControl.executeEmbed` was introduced as a way of running an `IO` against "simulated" time instead of real time. This allows the test to run much more quickly (since it can simulate a sleep instantly, instead of actually sleeping) and more reliably.

Cats Effect docs: https://typelevel.org/cats-effect/docs/core/test-runtime#full-execution

I recently used `executeEmbed` in https://github.com/typelevel/fs2/pull/2922, as an example :) 

So `executeEmbed` should be sprinkled in the appropriate places in our test suite. The `Timeout` suite is a good place to start :)

https://github.com/http4s/http4s/blob/08d219ba64f08a3741ebf2816f97a81b9561269a/server/shared/src/test/scala/org/http4s/server/middleware/TimeoutSuite.scala#L30

danicheg:
FWIW: unfortunately, `TestControl.executeEmbed` doesn't work with 'layered' `IO`s and `IO`s produced by `Async#async`. So it could be kinda tricky to make this actually work.

danicheg:
I've tried to apply `TestControl` in tests and many times faced with [this](https://github.com/typelevel/cats-effect/blob/series/3.x/testkit/shared/src/main/scala/cats/effect/testkit/TestControl.scala#L374-L379). But sure, maybe I was doing something wrong, just wanted to add some nota bene :)

fserra-mdsol:
I'd like to give it a crack, if it's fine with everyone

FrancescoSerra:
OK, this is the right account on which I can look into this :-)

FrancescoSerra:
thanks @armanbilge and sorry for all the trouble!

PR #6469, opened 2022-06-15
Test control on time based tests

FrancescoSerra:
Use TestControl for time based test suites
    
    Test suites impacted:
    * TimeoutSuite
    * ThrottleSuite


FrancescoSerra:
This addresses [#6467](https://github.com/http4s/http4s/issues/6467)

server/shared/src/test/scala/org/http4s/server/middleware/ThrottleSuite.scala
```diff
@@ -27,107 +29,142 @@ import org.http4s.Request
...
+        for {
+          _ <- control.results.assertEquals(None)
+          _ <- control.tick
+          _ <- control.results.assertEquals(None)
+          interval <- control.nextInterval
+          _ <- control.advanceAndTick(interval)
+          _ <- control.results.assertEquals(
+            Option(Outcome.succeeded[Id, Throwable, TokenAvailability](TokenAvailable.pure[Id]))
+          )
+        } yield ()
```
armanbilge, line 82: Can we rewrite this to record a timestamp when `takeTokenAfterRefill` completes and assert that it's what we expect? And then we'd only need `executeEmbed` instead of the full control machinery.

server/shared/src/test/scala/org/http4s/server/middleware/TimeoutSuite.scala
```diff
@@ -47,7 +49,7 @@ class TimeoutSuite extends Http4sSuite {
   private val uncancelableReq = Request[IO](GET, uri"/uncancelable")
 
   def checkStatus(resp: IO[Response[IO]], status: Status): IO[Unit] =
-    resp.map(_.status).timeout(1.seconds).assertEquals(status)
+    resp.map(_.status).assertEquals(status)
```
armanbilge, line 52: Why can the `timeout` go away here, I guess it was just acting as a test suite timeout with no real purpose in the tests?

armanbilge (commented):
This is looking very very nice. Fixed some flakes and upgraded test to properties to boot!

server/shared/src/test/scala/org/http4s/server/middleware/ThrottleSuite.scala
FrancescoSerra: Wouldn't that be more imprecise than this implementation? If we take a timestamp we're saying that the new token was taken sometime up to that moment in time - with what we got now we're establishing that up to the interval it couldn't happen, and only after that the fiber was able to pick up the token. Thoughts?

server/shared/src/test/scala/org/http4s/server/middleware/TimeoutSuite.scala
FrancescoSerra: Precisely - the timeout was a spurious time control over the execution of the test, rather than the actual timeout that would come from the middleware.

server/shared/src/test/scala/org/http4s/server/middleware/ThrottleSuite.scala
armanbilge: Hmm, not sure. For the record I could be misunderstanding what this test control is actually testing :)

> If we take a timestamp we're saying that the new token was taken sometime up to that moment in time

The timestamp would record the exact moment that the fiber continued, not before nor after.

> with what we got now we're establishing that up to the interval it couldn't happen, and only after that the fiber was able to pick up the token

The fact that the timestamp is not less than that interval would prove that up until then it couldn't happen.

FrancescoSerra: Same caveats apply to me: I've just learned about `TestControl`!

> The timestamp would record the exact moment that the fiber continued, not before nor after.
So you mean emitting a timestamp out of the test code execution flow? Hmmm, I think I'm seeing what you're getting at. OK, let's give it a spin

server/shared/src/test/scala/org/http4s/server/middleware/ThrottleSuite.scala
```diff
@@ -66,21 +66,10 @@ class ThrottleSuite extends Http4sSuite {
 
       val takeTokenAfterRefill = createBucket.flatMap { testee =>
         testee.takeToken *> IO.sleep(exceeded) *>
-          testee.takeToken
+          (IO.realTime, testee.takeToken).mapN((_, _))
```
armanbilge, line 69: Nice! But we should swap the order here, so that the timestamp is recorded after `takeToken` completes :)

FrancescoSerra:
Now the question is: would there be more suites that need such refactoring?

server/shared/src/test/scala/org/http4s/server/middleware/ThrottleSuite.scala
FrancescoSerra: Or maybe use `parMapN` and record it in parallel? All in all here we're already past the point of the refill time

armanbilge: > All in all here we're already past the point of the refill time.

Ah, that's a fair point. I'm actually not sure what the value of checking the timestamp here is? Given that `testee.takeToken` is non-blocking, it seems to me it is testing if `IO.sleep` works :)

server/shared/src/test/scala/org/http4s/server/middleware/ThrottleSuite.scala
```diff
@@ -27,107 +28,131 @@ import org.http4s.Request
...
+        takeFiveTokens >> IO.sleep(300.milliseconds) >> testee.takeToken
       }
-      IO.sleep(300.milliseconds) >> takeFiveTokens >> testee.takeToken
-    }
 
-    takeExtraToken
-      .map { result =>
-        ctx.advanceAndTick(300.milliseconds)
-        result
+      TestControl.execute(takeExtraToken).flatMap { control =>
```
armanbilge, line 88: Would you mind seeing if more of these can be written `executeEmbed`? Thanks!

server/shared/src/test/scala/org/http4s/server/middleware/ThrottleSuite.scala
FrancescoSerra: Yeah, I'm actually thinking: here we want to test that a particular function return some result only *after* some threshold, hence what we should do is:
1. test that before that threshold the outcome is not the expected value
2. test that after instead it is
So, that's what I'm going to implement - perhaps this was the missing bit in the first iteration!

server/shared/src/test/scala/org/http4s/server/middleware/ThrottleSuite.scala
FrancescoSerra: checking

FrancescoSerra: found another one

FrancescoSerra:
Unless we can find another Suite that needs refactoring, I think this PR is complete

server/shared/src/test/scala/org/http4s/server/middleware/ThrottleSuite.scala
```diff
@@ -27,107 +28,136 @@ import org.http4s.Request
...
+          _ <- control.tick
+          _ <- control.advanceAndTick(fd2)
+          _ <- control.results
+            .map(_.collect { case Outcome.Succeeded(TokenUnavailable(Some(interval))) =>
+              interval
+            })
+            .map(_.exists(_ == (fd1 - fd2)))
+            .assert
+        } yield ()
       }
```
armanbilge, line 159: Hmm, is it not possible to rewrite this with `executeEmbed`? At a glance it seems plausible :)

server/shared/src/test/scala/org/http4s/server/middleware/ThrottleSuite.scala
```diff
@@ -27,107 +28,124 @@ import org.http4s.Request
...
+          _ <- control.results.assertEquals(None)
+          interval <- control.nextInterval
+          _ <- control.advanceAndTick(interval)
+          _ <- control.results
+            .map(_.exists {
+              case Outcome.Succeeded(TokenUnavailable(Some(_))) => true
+              case _ => false
+            })
+            .assert
+        } yield ()
```
armanbilge, line 106: Sorry to be annoying, I think this is the last one! :)

FrancescoSerra: I must be blind :-) I thought they were all gone - will fix

FrancescoSerra: Done

armanbilge: Will make one more passthrough review a bit later but I reckon things are pretty good now :) Thanks for all your work on this!

FrancescoSerra: Sure thing! And actually thanks for being my watchdog in this PR! 😊

server/shared/src/test/scala/org/http4s/server/middleware/ResponseTimingSuite.scala
```diff
@@ -45,7 +46,7 @@ class ResponseTimingSuite extends Http4sSuite {
 
     val header = res
       .map(_.headers.headers.find(_.name == ci"X-Response-Time"))
-    header.map(_.map(_.value.toInt) === Some(artificialDelay)).assert
+    TestControl.executeEmbed(header.map(_.map(_.value.toInt) === Some(artificialDelay))).assert
```
armanbilge, line 49: I can't comment on it directly because it's above your changes 👆 but what is this `Sys.tick()` stuff? Where is it coming from. 

Aha, it's down below 👇 I think we can toss that. We don't need it now that we have `TestControl` 😂

FrancescoSerra: Yep, I also found out about [this guy](https://github.com/http4s/http4s/blob/series/0.23/client/shared/src/test/scala/org/http4s/client/middleware/RetrySuite.scala), and I'm going to use `TestControl` there too - hold on tight ;-)

FrancescoSerra: While I'm fixing this, I thought: shouldn't we put in some contributions guidelines that it would be recommended to use `TestControl` anywhere in tests the passing of time is a factor? If so, I'd happily make a PR to update the documentation and explain the pattern

armanbilge: Yep, that would be great!

FrancescoSerra: Would you mind pointing me to where the contributing guidelines sit, so that I can make the change? :-)

armanbilge: https://github.com/http4s/http4s/blob/c63372cc886bef753a62a90009e9dd8df8330a51/docs/contributing.md

FrancescoSerra: done

docs/contributing.md
```diff
@@ -194,10 +194,15 @@ Apache License 2.0.
   testing.
 * For assertions in tests using `assertEquals(a, b)` is preferable to `assert(a == b)`. 
   It brings nice diffs on assertions failures. For more details, see the [MUnit docs].
+* For time sensitive tests, where the passing of time is a factor in the logic that the
+  test is checking on, and that operates in the `IO` context,  consider using the mock 
+  runtime provided by `cats.effect.testkit.TestControl`. Fore mode details check out the
```
armanbilge, line 199:
```suggestion
  runtime provided by `cats.effect.testkit.TestControl`. For more details check out the
```

server/shared/src/test/scala/org/http4s/server/middleware/ResponseTimingSuite.scala
```diff
@@ -16,25 +16,21 @@
...
 import org.http4s.syntax.all._
 import org.typelevel.ci._
 
-import java.util.concurrent.TimeUnit
 import scala.concurrent.duration._
 
 class ResponseTimingSuite extends Http4sSuite {
-  import Sys.clock
-
   private val artificialDelay = 10
```
armanbilge, line 30: Just a nit! :)

```suggestion
  private val artificialDelay = 10.milliseconds
```

server/shared/src/test/scala/org/http4s/server/middleware/ThrottleSuite.scala
```diff
@@ -27,107 +27,116 @@ import org.http4s.Request
...
       }
-      .assertEquals(1)
+
+    TestControl.executeEmbed(prog).assertEquals(1)
   }
 
+  val localGen = for {
+    fd1 <- genFiniteDuration
+    fd2 <- genFiniteDuration.suchThat(_ < fd1)
+  } yield fd1 -> fd2
```
armanbilge, line 124: Would it be more efficient to generate two finite durations and sum them to get a monotonically increasing sequence?

server/shared/src/test/scala/org/http4s/server/middleware/TimeoutSuite.scala
```diff
@@ -70,47 +72,67 @@ class TimeoutSuite extends Http4sSuite {
...
-    testMiddleware(5.milliseconds) { app =>
-      for {
-        _ <- app(uncancelableReq).map(_.status).assertEquals(Status.ServiceUnavailable)
-        _ <- checkStatus(app(uncancelableReq), Status.ServiceUnavailable)
-          .intercept[TimeoutException]
-      } yield ()
+    forAllF(genFiniteDuration) { (timeOut: FiniteDuration) =>
+      val fixed = timeOut.min(
+        1099.milliseconds
+      ) // 1100 millis is the hard coded response time of uncancelableReq
```
armanbilge, line 100:
```suggestion
      // 1100 millis is the hard coded response time of uncancelableReq
      val fixed = timeOut.min(1099.milliseconds) 
```

armanbilge (commented):
This is awesome!! So many cleanups and improvements. Just a some tiny tiny nitpicks, but LGTM!

server/shared/src/test/scala/org/http4s/server/middleware/ResponseTimingSuite.scala
FrancescoSerra: Ah, this makes sense

server/shared/src/test/scala/org/http4s/server/middleware/ThrottleSuite.scala
FrancescoSerra: I'm not sure I'm seeing the improved efficiency - may be the Saturday morning brain slowness :grinning: care explaining how you'd do it and why it'd be more efficient with an increasing sequence rather than working out the difference between the 2 FDs?

armanbilge: IIUC the problem is that `genFiniteDuration.suchThat(_ < fd1)` is a rejection sampler: it will keep generating new samples until it satisfies the criterion that `fd2 < fd1`.

Instead, what if we just do:
```scala
val localGen = for {
  fd2 <- genFiniteDuration
  delta <- genFiniteDuration
  fd1 = fd2 + delta
} yield fd1 -> fd2
```

armanbilge: Another alternative could be:
```scala
val localGen = for {
  fd1 <- genFiniteDuration
  fd2 <- genFiniteDuration
} yield fd1.max(fd2) -> fd1.min(fd2)
```

FrancescoSerra: Ohh, I see what you mean by "adding" the FDs! 
Yes `suchThat` is indeed a rejection sampler, and in some scenarios it could lead to the exhaustion of the max allowed attempts and cause a spurious test failure. The problem that I see with both your suggestions is that the sampler could pick zero as value for both the samples, resulting in `0 -> 0` as result, which would invalidate the semantics of the test, I believe: in fact, if we allow a wait time of 0 how do we quantify that the fiber _has to wait_ in order to pick up the next available token? Now that I'm thinking of it, perhaps we should also enforce that the refill time (a.k.a: `fd1`) should be != 0?

armanbilge: Good point about `0`. One lazy solution is to just `.map(_ + 1.millis))` as needed 😉 (`TestControl` has millisecond-resolution).

> Now that I'm thinking of it, perhaps we should also enforce that the refill time (a.k.a: `fd1`) should be != 0?

Yep, might be good to add a `require` in `TokenBucket.local`. Currently seems like it will lead to an `NaN` via divide-by-zero.
https://github.com/http4s/http4s/blob/32d0c4644144234ccc307b5902e91d5015e22284/server/shared/src/main/scala/org/http4s/server/middleware/Throttle.scala#L71

FrancescoSerra: yeah, I'll go for the lazy solution on the generator :-)

FrancescoSerra: I also took the chance to revisit a bit the code for `TokenBucket.local`

server/shared/src/main/scala/org/http4s/server/middleware/Throttle.scala
```diff
@@ -80,7 +80,7 @@ object Throttle {
       override def takeToken: F[TokenAvailability] =
         for {
           values <- counter.access
-          previousTokens -> previousTime -> setter = values
+          ((previousTokens, previousTime), setter) = values
```
armanbilge, line 83: Huh, how is this related to JS? The compiler frontends should be the same, otherwise this is a bug 😂

FrancescoSerra: Nahh, ignore the log message: it's an issue with both scala 2.12 and JS!

armanbilge: If it's specific to JS, that's still a bug :) does it happen on 2.12 JVM?

server/shared/src/test/scala/org/http4s/server/middleware/ThrottleSuite.scala
```diff
@@ -27,107 +27,118 @@ import org.http4s.Request
...
+
+    TestControl.executeEmbed(prog).assertEquals(1)
   }
 
+  val localGen = for {
+    // guarantee a refill time > 0
+    fd1 <- genFiniteDuration
+    // guarantee some wait time < refill time
+    fd2 <- genFiniteDuration.map(i => if (i < fd1) i else fd1 - 1.millisecond)
+  } yield fd1 -> fd2
```
armanbilge, line 126: Sorry, just a nit, if `fd1` and `fd2` are the refill and wait times, can we just name them that way? :)

server/shared/src/main/scala/org/http4s/server/middleware/Throttle.scala
```diff
@@ -53,40 +56,50 @@ object Throttle {
       def takeToken: G[TokenAvailability] = fk(t.takeToken)
     }
 
+    private def getTime[F[_]: Temporal] =
+      Temporal[F].monotonic.map(_.toNanos)
```
armanbilge, line 60: Maybe `getNanoTime`?

server/shared/src/main/scala/org/http4s/server/middleware/Throttle.scala
```diff
@@ -53,40 +56,50 @@ object Throttle {
...
       *
-      * @param capacity the number of tokens the bucket can hold and starts with.
+      * @param capacity    the number of tokens the bucket can hold and starts with.
       * @param refillEvery the frequency with which to add another token if there is capacity spare.
       * @return A task to create the [[TokenBucket]].
       */
     def local[F[_]](capacity: Int, refillEvery: FiniteDuration)(implicit
         F: Temporal[F]
-    ): F[TokenBucket[F]] = {
-      def getTime = F.monotonic.map(_.toNanos)
```
armanbilge, line 65: Ah, I see, you moved it from here. Actually if this became a `val` then it would be slightly more efficient than as a `private def`.

server/shared/src/main/scala/org/http4s/server/middleware/Throttle.scala
```diff
@@ -53,40 +56,50 @@ object Throttle {
...
-                }
-
-                // If setter fails (yields the value false), then retry with recursive call
-                if (newTokenTotal >= 1)
-                  setter((newTokenTotal - 1, currentTime)).ifM(F.pure(TokenAvailable), takeToken)
-                else
-                  setter((newTokenTotal, currentTime)).ifM(F.pure(unavailable), takeToken)
+    ): F[TokenBucket[F]] = for {
+      // toNanos throws if called on an infinite duration
+      nanoValue <- F.catchNonFatal(refillEvery.toNanos)
```
armanbilge, line 72: `refillEvery: FiniteDuration` so that shouldn't happen :)

server/shared/src/main/scala/org/http4s/server/middleware/Throttle.scala
```diff
@@ -53,40 +56,50 @@ object Throttle {
...
+            val timeDifference = currentTime - previousTime
+            val tokensToAdd = timeDifference.toDouble / refillEvery.toNanos.toDouble
+            val newTokenTotal = Math.min(previousTokens + tokensToAdd, capacity.toDouble)
+
+            // If setter fails (yields the value false), then retry with recursive call
+            if (newTokenTotal >= 1)
+              setter((newTokenTotal - 1, currentTime))
+                .ifM(F.pure(TokenAvailable: TokenAvailability), takeToken)
+            else {
+              lazy val unavailable: TokenAvailability = {
```
armanbilge, line 95: Nit but this was fine as a `def`.`lazy val` can have hidden performance costs due to synchronization.

server/shared/src/main/scala/org/http4s/server/middleware/Throttle.scala
FrancescoSerra: Yep, good shout

server/shared/src/main/scala/org/http4s/server/middleware/Throttle.scala
FrancescoSerra: Where did I see it as `Duration`? 😂
Will remove

server/shared/src/test/scala/org/http4s/server/middleware/ThrottleSuite.scala
FrancescoSerra: Yep, of course

server/shared/src/main/scala/org/http4s/server/middleware/Throttle.scala
FrancescoSerra: I tend to prefer `lazy val` to `def` whenever the evaluation of the expression can happen _at most once_. Also, in this case I don't see the potential synchronization issue, as it's a local (helper) variable that could have well been inline'd at call site, hence my proposal to change it as is. I don't mind leaving it as a `def` anyway 😊 will fix

server/shared/src/main/scala/org/http4s/server/middleware/Throttle.scala
FrancescoSerra: Apparently so, as it was failing on the `rootJVM` job compiling against `2.12.x`

server/shared/src/main/scala/org/http4s/server/middleware/Throttle.scala
armanbilge: > I tend to prefer lazy val to def whenever the evaluation of the expression can happen at most once. 

Actually, when I know this for sure, I would definitely prefer a `def` because there would be no benefit to caching a `lazy val`, but only potential performance issues due to the overhead.

Just to illustrate that, here's how the code decompiles with `lazy val`

```java
    private final /* synthetic */ Throttle.TokenAvailability unavailable$lzycompute$1(LazyRef unavailable$lzy$1, long timeDifference$1) {
        Throttle.TokenAvailability tokenAvailability;
        LazyRef lazyRef = unavailable$lzy$1;
        synchronized (lazyRef) {
            Throttle.TokenAvailability tokenAvailability2;
            if (unavailable$lzy$1.initialized()) {
                tokenAvailability2 = (Throttle.TokenAvailability)unavailable$lzy$1.value();
            } else {
                long timeToNextToken = this.refillEvery$1.toNanos() - timeDifference$1;
                tokenAvailability2 = (Throttle.TokenAvailability)unavailable$lzy$1.initialize((Object)new Throttle.TokenUnavailable((Option<FiniteDuration>)OptionIdOps$.MODULE$.some$extension(implicits$.MODULE$.catsSyntaxOptionId((Object)new package.DurationLong(package$.MODULE$.DurationLong(timeToNextToken)).nanos()))));
            }
            tokenAvailability = tokenAvailability2;
        }
        return tokenAvailability;
    }
```

Here's the code if we make it a `def` instead.

```java
    private final Throttle.TokenAvailability unavailable$1(long timeDifference$1) {
        long timeToNextToken = this.refillEvery$1.toNanos() - timeDifference$1;
        return new Throttle.TokenUnavailable((Option<FiniteDuration>)OptionIdOps$.MODULE$.some$extension(implicits$.MODULE$.catsSyntaxOptionId((Object)new package.DurationLong(package$.MODULE$.DurationLong(timeToNextToken)).nanos())));
    }
```

For sure I'm no expert, but I think the second one has a better chance at being JITed and inlined by the JVM.

FrancescoSerra: I'm no expert either, and I hear what you're saying. I've reverted it to a def :+1:

server/shared/src/main/scala/org/http4s/server/middleware/Throttle.scala
FrancescoSerra: This is the offending job for 2.12 JVM: https://github.com/http4s/http4s/runs/6951228394?check_suite_focus=true

server/shared/src/main/scala/org/http4s/server/middleware/Throttle.scala
```diff
@@ -55,37 +58,47 @@ object Throttle {
...
-                val timeDifference = currentTime - previousTime
-                val tokensToAdd = timeDifference.toDouble / refillEvery.toNanos.toDouble
-                val newTokenTotal = Math.min(previousTokens + tokensToAdd, capacity.toDouble)
-                def unavailable = {
-                  val timeToNextToken = refillEvery.toNanos - timeDifference
+      val getNanoTime = Temporal[F].monotonic.map(_.toNanos)
+      for {
+        nanoValue <- refillEvery.toNanos.pure[F]
+        // make sure that refillEvery is positive
+        _ <- F.raiseUnless(nanoValue > 0L)(
```
armanbilge, line 72: I promise one last nit 😬 

```suggestion
      val getNanoTime = Temporal[F].monotonic.map(_.toNanos)
      val refillEveryNanos = refillEvery.toNanos
      for {
        // make sure that refillEvery is positive
        _ <- F.raiseUnless(nanoValue > 0L)(
```

PR #6473, opened 2022-06-17
Fix Content-Disposition filename encoding

leoniv:
Hi! It was easy to fix it, instead of submitting an issue. I faced the problem at 0.22 I suspect it's present at 0.23 too.

### DIFF
--- a/core/src/main/scala/org/http4s/headers/Content-Disposition.scala
+++ b/core/src/main/scala/org/http4s/headers/Content-Disposition.scala
@@ -36,6 +36,10 @@ object `Content-Disposition` {
   def parse(s: String): ParseResult[`Content-Disposition`] =
     ParseResult.fromParser(parser, "Invalid Content-Disposition header")(s)
 
+  val extraSafeChars = List(
+    '!', '#', '$', '&', '+', '-', '.', '^', '_', '`', '|', '~',
+  )
+
   private[http4s] val parser = {
     sealed trait ValueChar
     case class AsciiChar(c: Char) extends ValueChar
@@ -43,7 +47,7 @@ object `Content-Disposition` {
 
     val attrChar = Rfc3986.alpha
       .orElse(Rfc3986.digit)
-      .orElse(Parser.charIn('!', '#', '$', '&', '+', '-', '.', '^', '_', '`', '|', '~'))
+      .orElse(Parser.charIn(extraSafeChars))
       .map { (a: Char) =>
         AsciiChar(a)
       }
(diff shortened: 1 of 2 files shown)


FrancescoSerra (approved):
LGTM

core/src/main/scala/org/http4s/headers/Content-Disposition.scala
```diff
@@ -106,7 +108,10 @@ object `Content-Disposition` {
                 writer << "; " << k << '=' << '"'
                 writer.eligibleOnly(v, keep = safeChars, placeholder = '?') << '"'
               case (k @ ci"${_}*", v) =>
-                writer << "; " << k << '=' << "UTF-8''" << Uri.encode(v)
+                writer << "; " << k << '=' << "UTF-8''" << Uri.encode(
+                  toEncode = v,
+                  toSkip = CharPredicate.AlphaNum ++ extraSafeChars,
```
danicheg, line 113: Probably it'd be good to move this to predefined `val`, also add referring to [RFC's `attr-char`](https://datatracker.ietf.org/doc/html/rfc8187#section-3.2.1). WDYT?

tests/src/test/scala/org/http4s/parser/ContentDispositionHeaderSuite.scala
```diff
@@ -41,4 +44,29 @@ class ContentDispositionHeaderSuite extends Http4sSuite {
     )
   }
 
+  property("ContentDisposition filename encoding roundtrip") {
```
danicheg, line 47: It'd be good to add a few particular test cases when the previous parser fails and the new one doesn't.

danicheg (approved):
This is great. Thank you for your work! 
A bit more context for readers - [RFC declares](https://datatracker.ietf.org/doc/html/rfc8187#section-3.2.1) that:

> Inside the value part, characters not contained in attr-char are
>    encoded into an octet sequence using the specified character
>    encoding.  That octet sequence is then percent-encoded as specified
>    in Section 2.1 of [RFC3986].

The whole PR is excellent a few minor suggestions:

core/src/main/scala/org/http4s/headers/Content-Disposition.scala
leoniv: Done

tests/src/test/scala/org/http4s/parser/ContentDispositionHeaderSuite.scala
leoniv: Done

leoniv:
@danicheg what about fix it in 0.23? I can submit another PR. WDYT?

armanbilge:
@leoniv thanks! No need to worry about that, after we merge this PR we will merge the series/0.22 branch into series/0.23 :)

rossabaker:
Thanks!  0.22 is officially EOL, but if it bothered you enough to fix there, it's easy enough for us to do a courtesy release there.  I'll get that started.

leoniv:
Thanks, we a little bit behind the progress :)

rossabaker:
v0.22.14 is releasing from CI now with this.  I'll be asleep before it finishes. :smile:

PR #6475, opened 2022-06-19
Pick 'filename*' and ignore 'filename' in `Content-Disposition` header

danicheg:
According to [RFC6266](https://datatracker.ietf.org/doc/html/rfc6266#section-4.3):
>  Many user agent implementations predating this specification do not
>    understand the "filename*" parameter.  Therefore, when both
>    "filename" and "filename*" are present in a single header field
>    value, recipients SHOULD pick "filename*" and ignore "filename".

I think it should be fine to fix it for 0.22, considering another [`Content-Disposition` fix](https://github.com/http4s/http4s/pull/6473).

### DIFF
--- a/tests/src/test/scala/org/http4s/parser/ContentDispositionHeaderSuite.scala
+++ b/tests/src/test/scala/org/http4s/parser/ContentDispositionHeaderSuite.scala
@@ -25,8 +25,7 @@ class ContentDispositionHeaderSuite extends Http4sSuite {
   def parse(value: String): ParseResult[`Content-Disposition`] =
     `Content-Disposition`.parse(value)
 
-  test("ContentDisposition Header should render the correct values") {
-
+  test("Content-Disposition header should render the correct values") {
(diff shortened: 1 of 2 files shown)


tests/src/test/scala/org/http4s/parser/ContentDispositionHeaderSuite.scala
```diff
@@ -41,4 +40,11 @@ class ContentDispositionHeaderSuite extends Http4sSuite {
     )
   }
 
+  test("Content-Disposition header should pick 'filename*' parameter and ignore 'filename'") {
+    val headerValue = """form-data; filename="filename"; filename*=UTF-8''http4s%20filename"""
```
danicheg, line 44: Actually, 'filename*' should contain non-ISO-8859-1 characters in terms of fairness, but [that fix](https://github.com/http4s/http4s/pull/6473) must be merged first.

rossabaker:
Undoing my sloppy web merge to do it right... force push coming...

rossabaker:
I think it might be better to make that a synthetic property, rather than hide access.  I'm going to push a counterproposal.

PR #6481, opened 2022-06-21
Make Uri.Path.merge compliant

### LINKED ISSUE #5099: `Uri.resolve` behaviour very surprising, seems to discard last segment of `base`
This is a REPL with http4s 0.23.1 (and Scala 2.13):

```scala
scala> uri"http://example.com/foo".resolve(uri"bar")
val res4: org.http4s.Uri = http://example.com/bar

scala> uri"http://example.com/foo/bar".resolve(uri"baz")
val res9: org.http4s.Uri = http://example.com/foo/baz
```

As I understand [RFC 3986 5.2.3](https://datatracker.ietf.org/doc/html/rfc3986#section-5.2.3), which the http4s source mentions, the above is intended behaviour.

However, adding a / at the end of the base uri literal has no effect

```scala
scala> uri"http://example.com/foo/".resolve(uri"bar")
val res12: org.http4s.Uri = http://example.com/bar

scala> uri"http://example.com/foo/bar/".resolve(uri"baz")
val res13: org.http4s.Uri = http://example.com/foo/baz
```

I would expect the result to be http://example.com/foo/bar and http://example.com/foo/bar/baz, respectively.

As a workaround, calling `/ ""` on the base before resolve produces the expected result, but I would hope that this is not necessary.

```scala
scala> (uri"http://example.com/foo" / "").resolve(uri"bar")
val res2: org.http4s.Uri = http://example.com/foo/bar
```

I think that in order to implement RFC 3986 5.2.3 correctly, `Uri.Path.merge` should check whether the base path ends with a /, for example

```scala
    /* Merge paths per RFC 3986 5.2.3 */
    def merge(path: Path): Path = {
      val merge = if (isEmpty || endsWithSlash) segments else segments.init
      Path(merge ++ path.segments, absolute = absolute, endsWithSlash = path.endsWithSlash)
    }
```

Thank you for consideration

FrancescoSerra:
Fixes #5099 

This PR aims at making Uri.Path.merge compliant with RFC 3986 5.2.3

### DIFF
--- a/tests/shared/src/test/scala/org/http4s/PathSuite.scala
+++ b/tests/shared/src/test/scala/org/http4s/PathSuite.scala
@@ -27,14 +27,16 @@ class PathSuite extends Http4sSuite {
 
   test("equals should be consistent with equality") {
     forAll { (a: Path, b: Path) =>
-      if (a == b) (a.segments == b.segments) && (a.absolute == b.absolute) && (a.endsWithSlash == b.endsWithSlash)
-      else (a.segments != b.segments) || (a.absolute != b.absolute) || (a.endsWithSlash != b.endsWithSlash)
+      if (a == b)
+        (a.segments == b.segments) && (a.absolute == b.absolute) && (a.endsWithSlash == b.endsWithSlash)
+      else
+        (a.segments != b.segments) || (a.absolute != b.absolute) || (a.endsWithSlash != b.endsWithSlash)
     }
   }
 
   test("hashcode should be consistent with equality") {
     forAll { (a: Path, b: Path) =>
-      if (a == b) (a.## == b.##)
+      if (a == b) a.## == b.##
       else a.## != b.##
     }
   }
--- a/tests/shared/src/test/scala/org/http4s/PathSuite.scala
+++ b/tests/shared/src/test/scala/org/http4s/PathSuite.scala
@@ -19,8 +19,21 @@ package org.http4s
 import cats.kernel.laws.discipline._
 import org.http4s.Uri.Path
 import org.http4s.laws.discipline.arbitrary._
+import org.scalacheck.Prop._
 
 class PathSuite extends Http4sSuite {
   checkAll("Order[Path]", OrderTests[Path].order)
   checkAll("Semigroup[Path]", SemigroupTests[Path].semigroup)
+
+  test("equals should be consistent with equals") {
+    forAll { (a: Path, b: Path) =>
+      (a == b) == (a.segments == b.segments) == (a.absolute == b.absolute) == (a.endsWithSlash == b.endsWithSlash)
+    }
+  }
+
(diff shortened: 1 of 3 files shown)


FrancescoSerra:
Interestingly, if I just remove the property testing scaffoldings and statically provide to the test the values generated by `Scalacheck` on the failure, the test passes - hence I'm puzzled.

tests/shared/src/test/scala/org/http4s/PathSuite.scala
```diff
@@ -19,8 +19,21 @@ package org.http4s
...
 import org.http4s.laws.discipline.arbitrary._
+import org.scalacheck.Prop._
 
 class PathSuite extends Http4sSuite {
   checkAll("Order[Path]", OrderTests[Path].order)
   checkAll("Semigroup[Path]", SemigroupTests[Path].semigroup)
+
+  test("equals should be consistent with equals") {
+    forAll { (a: Path, b: Path) =>
+      (a == b) == (a.segments == b.segments) == (a.absolute == b.absolute) == (a.endsWithSlash == b.endsWithSlash)
```
armanbilge, line 30: Should those be `&&` ?

```suggestion
      (a == b) == ((a.segments == b.segments) && (a.absolute == b.absolute) && (a.endsWithSlash == b.endsWithSlash))
```

FrancescoSerra:
Like [here](https://github.com/http4s/http4s/runs/6981833275?check_suite_focus=true): with 
```
> ARG_0: *
> ARG_1: %4E/
```
I'm now wondering though if I'm getting deceived by the string representation of the `Path`s when scalacheck reports the failure...!

tests/shared/src/test/scala/org/http4s/PathSuite.scala
armanbilge: And maybe an implies actually?

```suggestion
      (a == b) ==> (a.segments == b.segments) && (a.absolute == b.absolute) && (a.endsWithSlash == b.endsWithSlash)
```

FrancescoSerra:
```
  /** Returns a new property that holds if and only if both this
   *  and the given property generates a result with the exact
   *  same status. Note that this means that if one of the properties is
   *  proved, and the other one passed, then the resulting property
   *  will fail. */
  def ==(p: => Prop) = this.flatMap { r1 =>
    p.map { r2 =>
      mergeRes(r1, r2, if(r1.status == r2.status) True else False)
    }
  }
```
:facepalm: 
Ignore me - thanks for pointing that out!

FrancescoSerra: Unfortunately it still fails :-/

rossabaker (approved):
Looks good to me.  I like that it also does not interfere with the workaround people have found.

Issue #5809, opened 2022-01-04
Multipart part's Content-Disposition with non-ascii filename fails to parse


adamw:
Using http4s 0.23.7.

Given the a simple service which parses the body as a multipart, which:

```scala
import cats.effect._
import org.http4s.blaze.server.BlazeServerBuilder
import org.http4s.dsl.io._
import org.http4s.headers.`Content-Disposition`
import org.http4s.server.Router
import org.http4s.{EntityDecoder, HttpRoutes, multipart}

import scala.concurrent.ExecutionContext

object Test2 extends IOApp {
  val service: HttpRoutes[IO] = HttpRoutes.of[IO] {
    case r @ POST -> Root / "file" =>
      implicitly[EntityDecoder[IO, multipart.Multipart[IO]]]
        .decode(r, strict = false)
        .value
        .flatMap {
          case Left(_) => BadRequest()
          case Right(mp) =>
            println("Got parts: " + mp.parts)
            mp.parts.foreach { part =>
              println("Part name: " + part.name)
              println("Part content disposition: " + part.headers.getWithWarnings[`Content-Disposition`])
            }
            Ok()
        }
    case _ =>
      BadRequest()
  }

  implicit val ec: ExecutionContext = scala.concurrent.ExecutionContext.Implicits.global

  override def run(args: List[String]): IO[ExitCode] = {
    BlazeServerBuilder[IO]
      .withExecutionContext(ec)
      .bindHttp(8080, "localhost")
      .withHttpApp(Router("/" -> service).orNotFound)
      .resource
      .use { _ => IO.never }
      .as(ExitCode.Success)
  }
}
```

And invocation:

```
curl -v -X 'POST' 'http://localhost:8080/file' -H 'accept: text/plain' -H 'Content-Type: multipart/form-data' -F 'file=@中文文件名.json;type=application/json'
```

The result on stdout is:

```
Got parts: Vector(Part(Headers(Content-Disposition: form-data; name="file"; filename="中文文件名.json", Content-Type: application/json),Stream(..)))
Part name: None
Part content disposition: Some(Left(NonEmptyList(org.http4s.ParseFailure: Invalid Content-Disposition header: Error(34,NonEmptyList(InRange(34,","))))))
```

aeons:
According to [the spec](https://httpwg.org/specs/rfc6266.html#disposition.parameter.filename), the `filename` parameter must be encoded using ISO-8859-1. If characters outside that are required, it should be set in the `filename*` parameter.

[...]

PR #6485, opened 2022-06-21
Add filename property to Content-Disposition

rossabaker:
As per RFC6266, Section 4.3, if `filename*` is defined, prefer that.  If not, try `filename`.

Kind of like #6475, but lossless.

### DIFF
--- a/core/src/main/scala/org/http4s/headers/Content-Disposition.scala
+++ b/core/src/main/scala/org/http4s/headers/Content-Disposition.scala
@@ -127,4 +127,13 @@ object `Content-Disposition` {
 }
 
 // see https://datatracker.ietf.org/doc/html/rfc2183
-final case class `Content-Disposition`(dispositionType: String, parameters: Map[CIString, String])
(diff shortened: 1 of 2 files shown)


danicheg (approved):
The notable differences between the two approaches:
1. Users that parse `filename` from the `parameters` manually will get new behavior without changes in their code (if we will do things in parser*).
2. Doing 'magic' within the parser will allow us to soft parsing rules for `filename` in case of the presence of `filename*`. To fix this issue #5809, for example. I was planning to dig into it further. But probably we are doing well in strict behavior. It's needed to clarify.

Nevertheless, I like this PR. It seems more transparent and less tricky, to be honest. So it's up to you to decide where we go with this.

rossabaker:
Yeah, I like preserving both parameters as they came to us, but giving a method that gives an interpretation that steers toward proper interpretation of the spec.  What to do about parsing non-Latin-1 in `filename` is one of those irksome strictness questions that I don't think changes what we do here.

PR #6486, opened 2022-06-22
Add Ordering for MediaRange

### LINKED ISSUE #5017: Be Consistent About Providing An Ordering Instance

FrancescoSerra:
This is the only missing `Ordering` I could find - all the other data types extend `Ordered`, which then allows deriving `Ordering` with implicit extension from `cats`

Closes #5017

### DIFF
--- a/core/shared/src/main/scala/org/http4s/ContentCoding.scala
+++ b/core/shared/src/main/scala/org/http4s/ContentCoding.scala
@@ -15,7 +15,7 @@
  */
 
 /*
- * Copyright 2013-2020 http4s.org
+ * Copyright 2013-2022 http4s.org
  *
  * SPDX-License-Identifier: Apache-2.0
  *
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/org/http4s/ContentCoding.scala
```diff
@@ -15,7 +15,7 @@
  */
 
 /*
- * Copyright 2013-2020 http4s.org
+ * Copyright 2013-2022 http4s.org
```
armanbilge, line 18: Spurious change? :) see also https://github.com/http4s/http4s/pull/5525

FrancescoSerra: Ahhh, thanks! It's good to know. Will rectify :-)

armanbilge (approved):
Thanks for checking these!

PR #6489, opened 2022-06-23
Add `Host#fromIp4sHost` method

### LINKED ISSUE #6421: Add Host.fromIp4sHost
`Uri.Host` already has a `fromIpAddress` that uses the IP4S IpAddress type.  Since Ip4s also has a `Host` type, can we support that as well?  I believe http4s and ip4s Host models are isomorphic.

danicheg:
Closes #6421.

### DIFF
--- a/core/shared/src/main/scala/org/http4s/Uri.scala
+++ b/core/shared/src/main/scala/org/http4s/Uri.scala
@@ -657,7 +657,7 @@ object Uri extends UriPlatform {
       * the underlying IP protocol version of the given address, either IPv4
       * or IPv6.
       */
-    @inline def fromIpAddress(value: ip4s.IpAddress): Host =
+    def fromIpAddress(value: ip4s.IpAddress): Host =
       value match {
         case value: ip4s.Ipv4Address =>
           Ipv4Address(value)
--- a/core/shared/src/main/scala/org/http4s/Uri.scala
+++ b/core/shared/src/main/scala/org/http4s/Uri.scala
@@ -657,13 +657,24 @@ object Uri extends UriPlatform {
       * the underlying IP protocol version of the given address, either IPv4
       * or IPv6.
       */
-    def fromIpAddress(value: ip4s.IpAddress): Host =
+    @inline def fromIpAddress(value: ip4s.IpAddress): Host =
       value match {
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/org/http4s/Uri.scala
```diff
@@ -657,14 +657,25 @@ object Uri extends UriPlatform {
       * the underlying IP protocol version of the given address, either IPv4
       * or IPv6.
       */
-    def fromIpAddress(value: ip4s.IpAddress): Host =
+    @inline def fromIpAddress(value: ip4s.IpAddress): Host =
```
rossabaker, line 660: Are we sure that `@inline` annotation will never generate a warning at the callsite?  Not opposed to it, but I've never been confident enough its ramifications to use it myself.  The rest looks good.

armanbilge: What kind of warning?

FWIW I'm leery about `@inline` myself, esp. setting precedent to decide whether we should `@inline` a method or not. Happy to leave it to the JIT.

rossabaker: Quoth [the docs](https://www.scala-lang.org/api/2.13.4/scala/inline.html):

> When inlining is enabled, the inliner will always try to inline methods or callsites annotated @inline (under the condition that inlining from the defining class is allowed, see -opt-inline-from:help). If inlining is not possible, for example because the method is not final, an optimizer warning will be issued. See -opt-warnings:help for details.

They give one example.  Are there more?  I live a life of fear.

danicheg: Can't prove that users won't face any warnings. So let's remove it. In my defense, I can say that I never faced inline-based warnings but faced using `@inline` for patmat-based methods.

rossabaker: Yeah, sorry to bring paranoia instead of evidence.  I got excited about `@inline` several Scalae ago and was warned it was for special occasions.  I've just started trusting the defaults until I have a smoking benchmark.

rossabaker (approved):

PR #6490, opened 2022-06-23
Relax `server.middleware.Caching` constraints

danicheg:
This is a pretty minor thing, just some TF-guys would be happy.

PR #6491, opened 2022-06-24
Spring cleaning: Remove unneeded temporal constraints and make some cosmetic changes

FrancescoSerra:
In the spirit of @danicheg 's [#6490](https://github.com/http4s/http4s/pull/6490) I took a stab at cleaning up some of the `Temporal` constraints and use weaker ones where appropriate.


FrancescoSerra:
Of course, MiMa :facepalm:

FrancescoSerra:
Obviously not applicabile !

server/shared/src/main/scala/org/http4s/server/middleware/Throttle.scala
```diff
@@ -60,77 +61,77 @@ object Throttle {
...
-    * @param http the service to transform.
+    * @param per    the time period over which a given number of calls is permitted.
+    * @param http   the service to transform.
     * @return a task containing the transformed service.
     */
-  def apply[F[_], G[_]](amount: Int, per: FiniteDuration)(
+  def apply[F[_]: Concurrent: Clock, G[_]](amount: Int, per: FiniteDuration)(
       http: Http[F, G]
-  )(implicit F: Temporal[F]): F[Http[F, G]] =
+  ): F[Http[F, G]] =
```
armanbilge, line 119: Serializing my Discord question:
> if you have `Concurrent` and `Clock`, should you just reach for `Temporal`? `Temporal` is strictly stronger, but you could say the same when rewriting `Concurrent` and `Sync` to `Async`

Fabio responds:
> I would argue that it isn't _strictly_ stronger
> and therefore just switch to Temporal
> I can build a Scheduler with Concurrent and Clock 😛

SystemFw: More seriously, I would stick to `Temporal` here. Upgrading from `Concurrent: Clock` to `Temporal` adds convenience and I'd argue doesn't significantly weaken parametricity. `Concurrent` already lets you add semantic blocking, so the fact that you have a new source of that with `Temporal` is not that significant.

armanbilge: > `Concurrent` already lets you add semantic blocking

Because I had to think about it for a second: via waiting on a `Deferred`.

FrancescoSerra: Thanks both for chiming in. This seems to be one of those cases where the whole is more than the sum of its parts 🙂 I'll revert the constraints here

rossabaker: How do you complete the Deferred without a dispatcher or spinning on the clock?

server/shared/src/main/scala/org/http4s/server/middleware/Throttle.scala
```diff
@@ -62,75 +62,74 @@ object Throttle {
...
-                        .map(if (_) Some(successResponse) else None)
-                    }
-
-                  attemptSet
+      val getNanoTime = F.monotonic.map(_.toNanos)
+      val refillEveryNanos = refillEvery.toNanos
+      for {
+        // make sure that refillEvery is positive
+        _ <- F.raiseUnless(refillEveryNanos > 0L)(
+          new IllegalArgumentException("refillEvery should be > 0 nano")
```
armanbilge, line 70: These are repeat of changes in https://github.com/http4s/http4s/pull/6469 right? We should probably merge `series/0.23` into `main` :)

FrancescoSerra: Yep, they turned up in here because I had originally based this branch on 0.23

armanbilge: Ah, gotcha. I opened https://github.com/http4s/http4s/pull/6499 btw.

armanbilge: Apologies, merging in 0.23 created some conflicts here 😅

FrancescoSerra: that's fine, I was expecting that :-D

armanbilge (approved):
LGTM, thanks! Always good to reduce the constraints before bincompat makes it painful 😅

rossabaker:
I like it, but I'm still not seeing how to implement `Temporal` from `Concurrent` + `Clock` unless we're allowing a `Thread.sleep` or a busy wait.  We don't have a callback to pass nor a dispatcher to complete a deferred.

armanbilge:
Depending on the answer to that question, do you think we _should_ reduce some of those `Temporal` constraints to `Concurrent+Clock` as originally proposed?

server/shared/src/main/scala/org/http4s/server/middleware/Throttle.scala
SystemFw: Ok, to clarify. The first comment was really meant as a joke, because you can only build a Scheduler without Temporal by spinning. The second comment is serious though: the increase in power is from the ability to wait, it doesn't really matter imho if you're waiting because you're `sleep`ing or because you haven't received an http call yet. `Concurrent` lets you create `Deferred`s, and therefore gives you the ability to wait

rossabaker:
I can think of two flimsy reasons not to unify the constraints:

* We are forgoing the parametricity of knowing it's not sleeping.  I've never reasoned this way.
* We are excluding the types that have `Concurrent` and `Clock` but not `Temporal`.  I've never seen one in the wild.

Did we draw the line at `MonadThrow` + `Clock` because parametricity rules out asynchrony?  The argument for this split is stronger, but we've drawn an arbitrary line.  I'm hoping someone can show me how to derive `sleep`, so I can stop worrying about it being arbitrary!

armanbilge:
This has accumulated conflicts 😢 unless there's any objections I am going to land this one. `Temporal` is stronger than `Concurrent` + `Clock` so we can always make that change binary- and source-compatibly in the future if we regret our choice here.

@FrancescoSerra would you have a moment to resolve the merge conflict? Sorry about that and thank you for your work here 🙏

FrancescoSerra:
> 

Nothing to be sorry about :-) Yes, will fix

armanbilge (approved):
Thanks again!

FrancescoSerra:
Thanks to you!

PR #6500, opened 2022-06-28
fix typo in Throttle middleware httpAapp->httpApp

jbwheatley:
Deprecate method with typo and add new method with correction


armanbilge:
Ouch, good catch and thanks for the PR. Since this is an issue in 0.23.x would you mind retargeting your PR to `series/0.23`? You will have to rebase and force-push.

PR #6501, opened 2022-06-28
fix typo in `Throttle` middleware `httpAapp->httpApp`

jbwheatley:
replaces #6500 now targeting series/0.23 rather than main


armanbilge (approved):
Thanks!

PR #6511, opened 2022-07-05
Don't use `Stream` in JSON client example

armanbilge:
Basically the same issue as https://github.com/http4s/http4s.g8/issues/278, which leads to questions in Discord :) the combination of `Resource` and `IO` works just fine here.


FrancescoSerra (approved):
LGTM

PR #6514, opened 2022-07-06
Remove redundant collection converting in `Query#multiParams`

danicheg:


core/shared/src/main/scala/org/http4s/Query.scala
```diff
@@ -146,7 +146,7 @@ final class Query private (value: Either[Vector[KeyValue], String])
   lazy val params: Map[String, String] =
     multiParams.view.mapValues(_.headOption.getOrElse("")).toMap
 
-  /** Map[String, Seq[String]] representation of the [[Query]]
+  /** `Map[String, Seq[String]` representation of the [[Query]]
```
rossabaker, line 149: I think that second bracket was right (not that we're compiling it):

```suggestion
  /** `Map[String, Seq[String]]` representation of the [[Query]]
```

PR #6518, opened 2022-07-07
Ember-Core: reimplement "combineArrays"

diesalbla:
Within the parser of `ember-core`, rewrites the `combineArrays` to
something that avoids some work.

- Not allocate, fill, and grow, an intermediaryArrayBuffer.
- Not allocate an array, immediately discarded, that is a copy of the Chunk.
- Maybe? avoid boxing brought by the use of a type-parametric API of `ArrayBuffer`, and deal directly with arrays of `Byte`.

### DIFF
--- a/ember-core/shared/src/main/scala/org/http4s/ember/core/Parser.scala
+++ b/ember-core/shared/src/main/scala/org/http4s/ember/core/Parser.scala
@@ -27,7 +27,6 @@ import org.typelevel.ci.CIString
 import scodec.bits.ByteVector
 
 import scala.annotation.switch
-import scala.collection.mutable
 
 private[ember] object Parser {
 
@@ -43,7 +42,7 @@ private[ember] object Parser {
           else {
             read.flatMap {
               case Some(chunk) =>
-                val nextBuffer = combineArrays(currentBuffer, chunk.toArray)
+                val nextBuffer = appendByteArrays(currentBuffer, chunk)
                 recurseFind(nextBuffer, read, maxHeaderSize)(f)(idx)
               case None if currentBuffer.length > 0 =>
                 F.raiseError(EmberException.ReachedEndOfStream())
@@ -512,11 +511,11 @@ private[ember] object Parser {
     }
   }
 
-  private def combineArrays[A: scala.reflect.ClassTag](a1: Array[A], a2: Array[A]): Array[A] = {
-    val buff = mutable.ArrayBuffer[A]()
-    buff.++=(a1)
-    buff.++=(a2)
-    buff.toArray
+  private[this] def appendByteArrays(a1: Array[Byte], a2: Chunk[Byte]): Array[Byte] = {
+    val res = new Array[Byte](a1.length + a2.size)
+    System.arraycopy(a1, 0, res, 0, a1.length)
+    a2.copyToArray(res, a1.length)
+    res
   }
 
   private[this] val space = ' '.toByte


ember-core/shared/src/main/scala/org/http4s/ember/core/Parser.scala
```diff
@@ -512,11 +511,18 @@ private[ember] object Parser {
...
+    val res = new Array[Byte](a1.length + a2.size)
+    var ix = 0
+    while (ix < a1.length) {
+      res(ix) = a1(ix)
+      ix += 1
+    }
+    while (ix < a2.size) {
+      res(ix) = a2(ix)
+      ix += 1
+    }
```
armanbilge, line 524: Is the math here right, shouldn't indices for `a2` start again from zero?

diesalbla: Thanks! Fixed.

ChristopherDavenport (commented):
Have we done anything to confirm this will improve performance?

ember-core/shared/src/main/scala/org/http4s/ember/core/Parser.scala
```diff
@@ -512,11 +511,19 @@ private[ember] object Parser {
...
+  private[this] def appendByteArrays(a1: Array[Byte], a2: Chunk[Byte]): Array[Byte] = {
+    val res = new Array[Byte](a1.length + a2.size)
+    var ix = 0
+    while (ix < a1.length) {
+      res(ix) = a1(ix)
+      ix += 1
+    }
+    ix = 0
+    while (ix < a2.size) {
+      res(a1.length + ix) = a2(ix)
```
armanbilge, line 523: @diesalbla actually we might want to check if `a2` is a `Chunk.ArraySlice` or `Chunk.ByteBuffer` and use specialized implementations for those. I think this `a2(ix)` call will be boxed.

diesalbla: So, in the call from `recurseFind`, the `a2` chunk comes from the call to `read`, which would come from the implementations in FS2 for Socket. For some it seems to be using ByteBuffer,  so that would help.

ember-core/shared/src/main/scala/org/http4s/ember/core/Parser.scala
```diff
@@ -512,11 +511,15 @@ private[ember] object Parser {
...
-    buff.++=(a1)
-    buff.++=(a2)
-    buff.toArray
+  private[this] def appendByteArrays(a1: Array[Byte], a2: Chunk[Byte]): Array[Byte] = {
+    val res = new Array[Byte](a1.length + a2.size)
+    var ix = 0
+    while (ix < a1.length) {
+      res(ix) = a1(ix)
+      ix += 1
+    }
```
armanbilge, line 520: I feel stupid now, but we should use `System.arraycopy` for this one too 😆

diesalbla: Well, isn't that one JVM-specific?

armanbilge: It's not! :)

https://github.com/scala-js/scala-js/blob/33ed7588ad80eaff5802aaf3b6f620b408a200cd/javalanglib/src/main/scala/java/lang/System.scala#L97

https://github.com/scala-native/scala-native/blob/024a46917b5b5b835c200aa99227915b99fdbb34/javalib/src/main/scala/java/lang/System.scala#L23

diesalbla: Oh, that's nice. Is it also "specialised", so it deals with arrays of bytes without boxing?

armanbilge: I mean, it's a `System` intrinsic call. I think it just copies the underlying memory directly.

diesalbla:
> Have we done anything to confirm this will improve performance?

We can run some of the benchmarks we have for that cross-language, and cross-stack, benchmarks. However, we have an intuition why it cannot be slower, and the code is not longer now (with the use of special methods).

ember-core/shared/src/main/scala/org/http4s/ember/core/Parser.scala
armanbilge: Actually, personally I prefer the development style of adding commits and never rebasing or force-pushing. It's much easier to review and I find the history helpful when looking at old PRs.

But, I appreciate your contributions, so I won't complain :)

ember-core/shared/src/main/scala/org/http4s/ember/core/Parser.scala
```diff
@@ -512,11 +511,11 @@ private[ember] object Parser {
...
-    val buff = mutable.ArrayBuffer[A]()
-    buff.++=(a1)
-    buff.++=(a2)
-    buff.toArray
+  private[this] def appendByteArrays(a1: Array[Byte], a2: Chunk[Byte]): Array[Byte] = {
+    val res = new Array[Byte](a1.length + a2.size)
+    System.arraycopy(a1, 0, res, 0, a1.length)
+    a2.copyToArray(res, a1.length)
+    res
   }
```
armanbilge, line 519: This is fantastic. @ChristopherDavenport what do you think?

armanbilge (approved):

ember-core/shared/src/main/scala/org/http4s/ember/core/Parser.scala
ChristopherDavenport: This looks good. I _think_ this code will get hit in a pass with end to end tests we have.

ChristopherDavenport (approved):
I believe this looks good, anytime Im thinking about arrays I'm paranoid about an off by one. But I think this is good and should get tested.

PR #6523, opened 2022-07-08
Ember-Core H2Client: split method fromSocket into procedures.

diesalbla:
Small refactor of the code in the `fromSocket` method of the `H2Client` class, for creating and starting up an `H2Connection`. Splits long code of the `for` comprehension (in Resource) into a set of auxiliary methods in `F`. 

Each of the auxiliary methods operates at the level of the "F" effect, and they are only lift to a `Resource` for the purposes of background running.

Initialising the `H2Connection` object is done in a separate method. Instead of accessing the components from the variables in the for loop, the other auxiliary methods access them
through the `H2Connection` fields themselves.

### DIFF
--- a/ember-core/shared/src/main/scala/org/http4s/ember/core/h2/H2Client.scala
+++ b/ember-core/shared/src/main/scala/org/http4s/ember/core/h2/H2Client.scala
@@ -64,7 +64,7 @@ Connection
     PushPromise - Special Case only after Open or Half-closed(remote)
 
  */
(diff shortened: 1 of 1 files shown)


armanbilge:
Thanks for the PR. Note that at present none of this code is tested, which makes reviewing this PR terrifying 😅

https://app.codecov.io/gh/http4s/http4s/blobs/series%2F0.23/ember-core/shared/src/main/scala/org/http4s/ember/core/h2/H2Client.scala

Issue #1695, opened 2018-03-09
Smite Seq wherever possible


jmcardon:
#1694 adds [this](https://github.com/http4s/http4s/pull/1694/files#diff-fc471fcf59af6fc124e6a51987dc7fa3R59) as a temp fix for not throwing an exception in multipart/form-data, but a more principled fix is simply acknowledging that you can't transmit a `multipart/form-data` body in a sane way without at least one part.

That said, there's more than one Fishy `Seq`, so we can probably take the opportunity to smite them at the same time.

rossabaker:
What is your genreal replacement for `Seq`?  `List`?  `Vector`?  `immutable.Seq`?  There are the `NonEmpty*` types in certain contexts.

rossabaker:
I still agree with this.  The first step is to identify them, and then we can identify correct replacements.

danicheg:
From my way of thinking, we have smitten `Seq` where it was reasonable and doable. I'm closing this, but feel free to reopen it.

PR #6524, opened 2022-07-09
Smite `Seq` in `Message#multiParams`, `Uri#multiParams`, `Query#multiParams`

danicheg:
Addresses the #1695.
This might be very controversial. So share what you think.


hamnis:
Since we are targeting > 2.12 we can use Vector instead of List.
That has better performance characteristics even for small collections after the changes in 2.13

danicheg:
> That has better performance characteristics even for small collections after the changes in 2.13

@hamnis Can you please share some links to the benchmark or where this is declared? I've attempted to find it multiple times and didn't succeed every time. Though I believed for sequence collections of <100 elems size `List` is about the same perf as `Vector`. At least for sure, I remember a small discussion with @RafalSumislawski on Twitter about it :)

hamnis:
One is here:

https://august.nagro.us/list-vs-vector.html

official docs:

https://docs.scala-lang.org/overviews/collections-2.13/performance-characteristics.html

armanbilge:
So IIUC:
1. Internally we are already using `List`, since you only had to change signatures.
2. The signature changes are all in covariant positions (i.e., we are giving the user a `List`, not asking for one).

If both those things are true, then actually we can even make this change compatibly in `series/0.23` if we wanted, since we are not breaking any contract.

danicheg:
@armanbilge I'm 100% sure this change should go in 1.0. Since users can expect in their code that returning type is exactly `Seq`. So we can break it very easily. Hence even if MiMa will allow us to do this change in 0.23, it's still not binary-compatible.

armanbilge:
> Since users can expect in their code that returning type is exactly `Seq`

A `List` is a `Seq` :)

> Hence even if MiMa will allow us to do this change in 0.23, it's still not binary-compatible.

It is binary-compatible :) binary-compatibility is about the signatures as represented in the bytecode. Indeed, if you inspect the bytecode you will find that it is unchanged, due to type erasure of generics.

danicheg:
@armanbilge Hmm, tend to agree with you. I just was worried about using `scala.collection.Seq` within `Message` in 2.12 (not immutable one as we do almost everywhere when using `Seq`). So it seems because we already use `List` in internals, we can't hurt users somehow.

armanbilge:
FWIW I'm not advocating we make this change in `series/0.23` ... just saying we could, if we wanted to, since it is actually compatible. So I think this is not such a controversial change for 1.0 at all.

Changing to `Vector` would be more controversial IMHO.

armanbilge (approved):

rossabaker:
Did those Vector improvements apply to 2.13?  The benchmarks were run on 3.1.1.  The last time I baked off `List` and `Vector`, `List` was a winner on traversals.

There was [this 2018 argument for Chain](https://typelevel.org/blog/2018/09/04/chain-replacing-the-list-monoid.html), but I don't imagine we use it semigroupally enough to justify the exotic Cats type.

armanbilge:
Yes, the Scala 2.13 standard lib is shared with Scala 3. So the `Vector` benchmark should apply to both.

danicheg:
So, with the power vested in me, I merge this. If someone wants to bring `Vector` somewhere, I believe we can do this iteratively.

PR #6526, opened 2022-07-10
Expose `Network` constraint in ember builders

armanbilge:
This change is motivated by the I/O Integrated Runtime Concept in https://github.com/typelevel/cats-effect/discussions/3070 which proposes an alternate implementation of `Network[IO]`. There have also been discussions of a `Network` implementation based on [fs2-netty](https://github.com/typelevel/fs2-netty).

For Ember to be able to take advantage of such advances, we must expose the `Network[F]` constraint in the builder. Fortunately this is source-compatible, because an implicit `Network[F]` is always in scope for `Async[F]` which was already required.

See also how Skunk does it:
https://github.com/tpolecat/skunk/blob/14c756d68ba88da4415699bea388149c43e4e962/modules/core/shared/src/main/scala/Session.scala#L262

~~While looking at that, I notice Skunk actually only takes a `Concurrent` constraint there. Seems to me we may be able to achieve a similar relaxation, let me give that a try.~~ No, this is not possible. We need `Sync` to interop with the HPACK implementation.

### DIFF
--- a/ember-client/shared/src/main/scala/org/http4s/ember/client/EmberClientBuilder.scala
+++ b/ember-client/shared/src/main/scala/org/http4s/ember/client/EmberClientBuilder.scala
@@ -41,7 +41,7 @@ import org.typelevel.log4cats.Logger
 import scala.concurrent.duration.Duration
 import scala.concurrent.duration._
 
-final class EmberClientBuilder[F[_]: Async] private (
+final class EmberClientBuilder[F[_]: Async: Network] private (
     private val tlsContextOpt: Option[TLSContext[F]],
     private val sgOpt: Option[SocketGroup[F]],
     val maxTotal: Int,
@@ -351,7 +351,7 @@ final class EmberClientBuilder[F[_]: Async] private (
 
 object EmberClientBuilder extends EmberClientBuilderCompanionPlatform {
 
-  def default[F[_]: Async] =
+  def default[F[_]: Async: Network] =
     new EmberClientBuilder[F](
       tlsContextOpt = None,
       sgOpt = None,
@@ -371,6 +371,10 @@ object EmberClientBuilder extends EmberClientBuilderCompanionPlatform {
       enableHttp2 = false,
       pushPromiseSupport = None,
     )
+
+  @deprecated("Use the overload which accepts a Network", "0.23.14")
+  def default[F[_]](async: Async[F]): EmberClientBuilder[F] =
+    default(async, Network.forAsync(async))
 
   private object Defaults {
     val acgFixedThreadPoolSize: Int = 100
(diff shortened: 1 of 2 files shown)


ChristopherDavenport:
I think this should be a `with` method rather than an additional implicit. If we are replacing it, it should be explicit rather than an implicit. Because it implies more than 1 implementation and only single implementations should be implicit.

armanbilge:
@ChristopherDavenport forgetting what I said about multiple `Network` implementations (I'm actually not sure about this anymore) does this change still make sense? There's sort of a design/philosophical question here: should we be broad in our constraints, or only ask for the minimum constraint to satisfy our requirements.

For example, does it make sense to ask for `Async` + `Console`, since you can build a `Console` from an `Async`?

ChristopherDavenport:
I'm not following. The question is a matter of defaults. So is the idea that it will infer automatically at the site either way, but that way the users can override?

armanbilge:
Sorry, I muddied the issue with talk of multiple `Network` implementations. I'm no longer sure if that's a good idea or what will end up happening.

The question here is whether the constraint should be `F[_]: Async: Network` or just `F[_]: Async` since you can derive a `Network` from an `Async`.

ChristopherDavenport:
I think we can leave it for the time being. Since the things this instance is used for is actually replaced via the builder. Unless I'm missing something.

armanbilge:
Fair enough.

ChristopherDavenport:
I've changed my mind. This is better.

ChristopherDavenport (approved):

ember-client/shared/src/main/scala/org/http4s/ember/client/EmberClientBuilder.scala
```diff
@@ -372,6 +372,10 @@ object EmberClientBuilder extends EmberClientBuilderCompanionPlatform {
       pushPromiseSupport = None,
     )
 
+  @deprecated("Use the overload which accepts a Network", "0.23.14")
```
armanbilge, line 375:
```suggestion
  @deprecated("Use the overload which accepts a Network", "0.23.16")
```

ember-server/shared/src/main/scala/org/http4s/ember/server/EmberServerBuilder.scala
```diff
@@ -262,6 +262,10 @@ object EmberServerBuilder extends EmberServerBuilderCompanionPlatform {
       false,
     )
 
+  @deprecated("Use the overload which accepts a Network", "0.23.14")
```
armanbilge, line 265:
```suggestion
  @deprecated("Use the overload which accepts a Network", "0.23.16")
```

Issue #5812, opened 2022-01-05
Bikeshed `StrictEntity`


armanbilge:
See discussion in https://github.com/http4s/http4s/pull/5807.

armanbilge:
So new datapoint on the `Chunk[Byte]` vs `ByteVector` question for `Entity.Strict`. In https://github.com/http4s/http4s/pull/5804#discussion_r777230747 there was a consensus for `Chunk`, since `Chunk` can subsume `ByteVector`.

However, while working on the http4s AWS lambda integration in https://github.com/typelevel/feral/pull/214#discussion_r882320272 I wanted to do base 64 encoding and UTF8 decoding of `Entity.Strict` bytes. AFAICT while it is easy to do these ops directly on a `ByteVector`, it's not possible to do them on `Chunk` without lifting it into a `Stream`, which defeats the whole purpose of the `Strict` optimization.

Maybe this is a niche case, but I wonder if there is friction between the fact that `Chunk` is ultimately intended to be used with `Stream` and has less standalone support for byte manipulation. And meanwhile `Entity.Strict` is intended to avoid materializing a `Stream`.

armanbilge:
... and, I ran into this same thing when optimizing the jdk-http-client, where again `ByteVector` is the more helpful datatype. See:
- https://github.com/http4s/http4s-jdk-http-client/pull/644#discussion_r882347645

danicheg:
It turns out that we changed the data type to `ByteVector` in `Entity.Strict`. Do we still want to discuss something in this issue, or can we close it? cc @armanbilge

armanbilge:
In https://github.com/http4s/http4s/pull/5807 there was some discourse about naming:
- https://github.com/http4s/http4s/pull/5807#discussion_r777726528
- https://github.com/http4s/http4s/pull/5807#issuecomment-1004822561

Personally I'm happy with `Entity.Strict` naming following my PR https://github.com/http4s/http4s/pull/6528. Unless someone else would still like to revisit this, we can close.

PR #6528, opened 2022-07-11
Use `scodec.bits.ByteVector` for `Entity.Strict`

armanbilge:
**tl;dr** Because `ByteVector` is specialized for `Byte`, it has idiomatic APIs for byte manipulation and copy-free interop that `Chunk` lacks.

This is a follow-up to my comments in https://github.com/http4s/http4s/issues/5812. See there for more details, summary below:

1. IMHO a `ByteVector` the more idiomatic data structure for, well, bytes. `Chunk[Byte]` is kind of an FS2 thing.
2. `ByteVector` exposes useful ops specifically for byte manipulation, not available on `Chunk[Byte]`. I wanted this in:
    https://github.com/typelevel/feral/pull/214#discussion_r882320272
3. `ByteVector` exposes better APIs for copy-free interop. I wanted this in:
    https://github.com/http4s/http4s-jdk-http-client/pull/644#discussion_r882347645
4. Even though `ByteVector` does not subsume `Chunk`, many common `Chunk`s can become a `ByteVector` without copying.
    https://github.com/typelevel/fs2/blob/79344baae70b6e3d2a3731f5e7f8e3d81ce9eedb/core/shared/src/main/scala/fs2/Chunk.scala#L364-L377
5. Does anyone have an example of a backend where `Chunk` has an advantage over `ByteVector`?

### DIFF
--- a/core/shared/src/main/scala/org/http4s/EntityEncoder.scala
+++ b/core/shared/src/main/scala/org/http4s/EntityEncoder.scala
@@ -147,14 +147,14 @@ object EntityEncoder {
   ): EntityEncoder.Pure[Array[Char]] =
     stringEncoder.contramap(new String(_))
 
+  implicit val byteVectorEncoder: EntityEncoder.Pure[ByteVector] =
+    simple(`Content-Type`(MediaType.application.`octet-stream`))(identity)
+
   implicit val chunkEncoder: EntityEncoder.Pure[Chunk[Byte]] =
     byteVectorEncoder.contramap(_.toByteVector)
 
   implicit val byteArrayEncoder: EntityEncoder.Pure[Array[Byte]] =
     byteVectorEncoder.contramap(ByteVector.view)
-
-  implicit def byteVectorEncoder[F[_]]: EntityEncoder[F, ByteVector] =
-    simple(`Content-Type`(MediaType.application.`octet-stream`))(identity)
 
   /** Encodes an entity body.  Chunking of the stream is preserved.  A
     * `Transfer-Encoding: chunked` header is set, as we cannot know
--- a/core/shared/src/main/scala/org/http4s/EntityEncoder.scala
+++ b/core/shared/src/main/scala/org/http4s/EntityEncoder.scala
@@ -101,14 +101,12 @@ object EntityEncoder {
     *
     * This constructor is a helper for types that can be serialized synchronously, for example a String.
     */
-  def simple[A](hs: Header.ToRaw*)(toChunk: A => Chunk[Byte]): EntityEncoder.Pure[A] =
-    encodeBy(hs: _*)(a => Entity.strict(toChunk(a)))
+  def simple[A](hs: Header.ToRaw*)(toByteVector: A => ByteVector): EntityEncoder.Pure[A] =
+    encodeBy(hs: _*)(a => Entity.strict(toByteVector(a)))
 
   /** Encodes a value from its Show instance.  Too broad to be implicit, too useful to not exist. */
-  def showEncoder[A](implicit charset: Charset = `UTF-8`, show: Show[A]): EntityEncoder.Pure[A] = {
-    val hdr = `Content-Type`(MediaType.text.plain).withCharset(charset)
-    simple[A](hdr)(a => Chunk.array(show.show(a).getBytes(charset.nioCharset)))
-  }
+  def showEncoder[A](implicit charset: Charset = `UTF-8`, show: Show[A]): EntityEncoder.Pure[A] =
+    stringEncoder.contramap(show.show)
 
   def emptyEncoder[A]: EntityEncoder.Pure[A] =
     new EntityEncoder[fs2.Pure, A] {
@@ -141,7 +139,7 @@ object EntityEncoder {
 
   implicit def stringEncoder(implicit charset: Charset = `UTF-8`): EntityEncoder.Pure[String] = {
     val hdr = `Content-Type`(MediaType.text.plain).withCharset(charset)
-    simple(hdr)(s => Chunk.array(s.getBytes(charset.nioCharset)))
+    simple(hdr)(ByteVector.encodeString(_)(charset.nioCharset).toOption.get)
   }
 
   implicit def charArrayEncoder(implicit
@@ -150,13 +148,13 @@ object EntityEncoder {
     stringEncoder.contramap(new String(_))
 
   implicit val chunkEncoder: EntityEncoder.Pure[Chunk[Byte]] =
-    simple(`Content-Type`(MediaType.application.`octet-stream`))(identity)
+    byteVectorEncoder.contramap(_.toByteVector)
 
   implicit val byteArrayEncoder: EntityEncoder.Pure[Array[Byte]] =
-    chunkEncoder.contramap(Chunk.array[Byte])
+    byteVectorEncoder.contramap(ByteVector.view)
 
   implicit def byteVectorEncoder[F[_]]: EntityEncoder[F, ByteVector] =
-    chunkEncoder.contramap(Chunk.byteVector)
+    simple(`Content-Type`(MediaType.application.`octet-stream`))(identity)
 
   /** Encodes an entity body.  Chunking of the stream is preserved.  A
     * `Transfer-Encoding: chunked` header is set, as we cannot know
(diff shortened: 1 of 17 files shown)


core/shared/src/main/scala/org/http4s/EntityEncoder.scala
```diff
@@ -141,7 +139,7 @@ object EntityEncoder {
 
   implicit def stringEncoder(implicit charset: Charset = `UTF-8`): EntityEncoder.Pure[String] = {
     val hdr = `Content-Type`(MediaType.text.plain).withCharset(charset)
-    simple(hdr)(s => Chunk.array(s.getBytes(charset.nioCharset)))
+    simple(hdr)(ByteVector.encodeString(_)(charset.nioCharset).toOption.get)
```
armanbilge, line 142: So, this is a bit weird. Am I missing something, should this be a `Pure` encoder?

core/shared/src/main/scala/org/http4s/Message.scala
```diff
@@ -768,7 +768,7 @@ object Response extends KleisliSyntax {
   val notFound: Response[Pure] =
     Response(
       Status.NotFound,
-      entity = Entity.Strict(Chunk.array("Not found".getBytes(StandardCharsets.UTF_8))),
+      entity = Entity.Strict(ByteVector.view("Not found".getBytes(StandardCharsets.UTF_8))), // TODO
```
armanbilge, line 771: I opened a PR to add literals e.g. `utf8Bytes"Not found"`.

- https://github.com/scodec/scodec-bits/pull/381

core/shared/src/main/scala/org/http4s/EntityEncoder.scala
```diff
@@ -150,13 +148,13 @@ object EntityEncoder {
...
 
   implicit val chunkEncoder: EntityEncoder.Pure[Chunk[Byte]] =
-    simple(`Content-Type`(MediaType.application.`octet-stream`))(identity)
+    byteVectorEncoder.contramap(_.toByteVector)
 
   implicit val byteArrayEncoder: EntityEncoder.Pure[Array[Byte]] =
-    chunkEncoder.contramap(Chunk.array[Byte])
+    byteVectorEncoder.contramap(ByteVector.view)
 
   implicit def byteVectorEncoder[F[_]]: EntityEncoder[F, ByteVector] =
```
RafalSumislawski, line 156: Why is this `EntityEncoder[F, X]` while the others  are `EntityEncoder.Pure[X]`?

RafalSumislawski (approved):

core/shared/src/main/scala/org/http4s/EntityEncoder.scala
RafalSumislawski: I'm slightly inclined towards making it an impure encoder to get rid of this `.toOption.get`.

core/shared/src/main/scala/org/http4s/EntityEncoder.scala
armanbilge: Good catch, probably an oversight

core/shared/src/main/scala/org/http4s/EntityEncoder.scala
```diff
@@ -101,14 +101,12 @@ object EntityEncoder {
...
-    encodeBy(hs: _*)(a => Entity.strict(toChunk(a)))
+  def simple[A](hs: Header.ToRaw*)(toByteVector: A => ByteVector): EntityEncoder.Pure[A] =
+    encodeBy(hs: _*)(a => Entity.strict(toByteVector(a)))
 
   /** Encodes a value from its Show instance.  Too broad to be implicit, too useful to not exist. */
-  def showEncoder[A](implicit charset: Charset = `UTF-8`, show: Show[A]): EntityEncoder.Pure[A] = {
-    val hdr = `Content-Type`(MediaType.text.plain).withCharset(charset)
-    simple[A](hdr)(a => Chunk.array(show.show(a).getBytes(charset.nioCharset)))
-  }
+  def showEncoder[A](implicit charset: Charset = `UTF-8`, show: Show[A]): EntityEncoder.Pure[A] =
```
danicheg, line 108: This is irrelevant to current PR, but having `Charset` in implicit scope isn't that a good design, I guess. It isn't a typeclass. Should we discuss this before we release 1.0?

core/shared/src/main/scala/org/http4s/multipart/Part.scala
```diff
@@ -44,7 +43,7 @@ object Part {
   def formData(name: String, value: String, headers: Header.ToRaw*): Part[fs2.Pure] =
     Part(
       Headers(`Content-Disposition`("form-data", Map(ci"name" -> name))).put(headers: _*),
-      Entity.Strict(Chunk.array(value.getBytes(UTF_8))),
+      Entity.Strict(ByteVector.encodeUtf8(value).toOption.get),
```
danicheg, line 46: To work around `.toOption.get` we can do this probably
```suggestion
      Entity.Strict(ByteVector.view(value.getBytes(UTF_8))),
```

danicheg (commented):
`scodec` is great and I think we can rely on it. But it's always scary to depend on third-party libraries in the core models :) If `scodec` will decide to break something we will have a classic dilemma.

armanbilge:
> But it's always scary to depend on third-party libraries in the core models :)

scodec-bits is already our dependency in the core models, via fs2 :) itself doesn't have any dependencies and is very stable.

core/shared/src/main/scala/org/http4s/EntityEncoder.scala
armanbilge: Yeah definitely worth a discussion. I agree that anything `implicit`-based should be well-justified.

core/shared/src/main/scala/org/http4s/multipart/Part.scala
armanbilge: @danicheg thanks, I'm pretty curious what the difference is, so some stupid questions: is `value.getBytes(UTF_8)` guaranteed to never throw an error, why is that? What makes it different from `ByteVector.encodeUtf8`? Should this encoder be `Pure` if there can be error?

ChristopherDavenport:
Without digging in too much. Working with Bytevector is awesome, we will be able to do a lot of things with this without getting either into mutable arrays, but will make transformations much simpler.

I'm in favor of this design generally.

core/shared/src/main/scala/org/http4s/multipart/Part.scala
danicheg: Honestly, I'm not a `scodec` expert, I can just suppose things. 
Speaking from the perspective of the types, we already had the version on `value.getBytes(UTF_8)` and didn't check possible exceptions. Though, I think that `scodec` is much safer if we wouldn't lose on the purpose of this safety through `.toOption.get`. 
Basically, your variant is as safe as the earlier one, I guess. Just users might face a "None.get" exception, instead of what Java API could throw. Will that exception be more clear to users? I dunno.

armanbilge: > Basically, your variant is as safe as the earlier one

Thanks, yes, exactly. And the error message is a regression :) since this is 1.0 and we can be breaking things, maybe this should not be `Pure` then.

armanbilge: Ah, `getBytes` is total and never throws an error.
> This method always replaces malformed-input and unmappable-character sequences with this charset's default replacement byte array.

armanbilge:
Thanks everyone for the feedback! I think the consensus is overall positive so I will merge this in the next few days unless someone has some new thoughts.

armanbilge:
I've got a PR up optimizing all the `Chunk`-to-`ByteVector` conversions in https://github.com/typelevel/fs2/pull/2956 so they are all copy-free. So in case anybody really needs `Chunk` instead of `ByteVector` it is possible to go back and forth with minimal overhead.

rossabaker:
This is good.

History lesson: scodec-bits was a core dependency all the way back to scalaz-stream days.  fs2-0.9 dropped it, and fs2-0.10 brought it back as a non-core integration.  We never fully got rid of it, but were distancing ourselves from it following fs2's lead.  Once it came back in fs2-core-1.0, there became no particular reason to avoid it, and still lots of reasons above to like it.

PR #6533, opened 2022-07-13
Feature/digest client

### LINKED ISSUE #3588: Digest Auth client middleware

FrancescoSerra:
Closes #3588

Follows up on PR #3619

I'm not overly fond of this implementation, so I'd like to get feedback - I kept the changes to the minimum


client/shared/src/main/scala/org/http4s/client/digest/NonceCounter.scala
```diff
@@ -0,0 +1,48 @@
...
+
+object NonceCounter {
+  val MaxNonce: Long = java.lang.Long.parseLong("FFFFFFFF", 16)
+
+  def make[F[_]: Async](
+      ttl: FiniteDuration = 1.minute,
+      maxNonce: Long = MaxNonce,
+  ): F[NonceCounter[F]] =
+    Random
+      .javaSecuritySecureRandom[F]
```
FrancescoSerra, line 45: According to the relative [RFC](https://datatracker.ietf.org/doc/html/rfc7616), there's no strict need for a cryptographically secure cnonce - however I think this won't hurt.

armanbilge: Is this the relevant section?
https://datatracker.ietf.org/doc/html/rfc7616#section-5.12

> The security of this protocol is critically dependent on the
   randomness of the randomly chosen parameters, such as client and
   server nonces.  These should be generated by a strong random or
   properly seeded pseudorandom source (see [[RFC4086](https://datatracker.ietf.org/doc/html/rfc4086)]).

FrancescoSerra: yes, I think the `properly seeded pseudorandom` seems covered by the standard `Random` typeclass.

FrancescoSerra: Same! I'm only interpreting the RFC text, but I'm no expert and I might be blatantly wrong 😊

Issue #5543, opened 2021-11-08
Deprecate `BackendBuilder#allocate`


bplommer:
Using `.resource.allocated` is just the right level of awkwardness for something people shouldn't normally do, and makes it easy to search/lint for leaky resource usages.

PR #6540, opened 2022-07-17
Remove `BackendBuilder#allocate`

danicheg:
This addresses #5543. Though @bplommer has suggested deprecating it, I feel that for 1.0 we should remove it. Users could still do this on their side via `serverBuilder.resource.allocated`.


armanbilge:
👍 but it is kind to put a `@deprecated` in `series/0.23` for anything we will remove in 1.0.

bplommer (approved):

ChristopherDavenport (approved):

Issue #5280, opened 2021-09-28
Parse errors are difficult to read


rossabaker:
We should be able to do better:

```
scala> HttpVersion.fromString("blech")
res1: org.http4s.ParseResult[org.http4s.HttpVersion] = Left(org.http4s.ParseFailure: HTTP version: Error(0,NonEmptyList(OneOfStr(0,List(HTTP/)))))
```

armanbilge:
Is this something that can be improved here, or in cats-parse?

rossabaker:
Relevant: https://github.com/typelevel/cats-parse/issues/239

rossabaker:
We can do it here, but if it's not HTTP specific, it might be contributable back to cats-parse.

johnynek:
It would be nice to have a prettier rendering, but for programming language applications you probably want to control that.

I would make sure you look at LocationMap which at least will compute line/col for you, and can fetch lines so you can show a context to users.

johnynek:
https://github.com/johnynek/bosatsu/pull/808

There is a PR with code you can copy if you want.

I don't know how easy it is to make it general (e.g. bosatsu is showing errors to users using paiges Doc, but you might be serializing errors and sending them over the network).

PR #6541, opened 2022-07-17
Prettify error messages when parsing

danicheg:
This addresses #5280. 
How it was:
```scala
scala> HttpVersion.fromString("blech")
res1: org.http4s.ParseResult[org.http4s.HttpVersion] = Left(org.http4s.ParseFailure: HTTP version: Error(0,NonEmptyList(OneOfStr(0,List(HTTP/)))))
```
How it will be:
```scala
scala> HttpVersion.fromString("blech")
val res0: org.http4s.ParseResult[org.http4s.HttpVersion] =
Left(org.http4s.ParseFailure: HTTP version: blech
^
expectation:
* must match string: "HTTP/")
```

All credits to @MasseGuillaume and [cats-parse](https://github.com/typelevel/cats-parse) team.

Though it's hard to imagine how users might use these error messages in their logic part of programs (except for just error printing), these changes alter behavior for sure. But I like this prettifying so much that I want to have it in 0.23. What do you think?


armanbilge (commented):
I think this is a good change. I agree, it's hard to see how the old errors could be used meaningfully.

One thing I'm surprised about is how few changes this needed, these are really the only places?

danicheg:
> One thing I'm surprised about is how few changes this needed, these are really the only places?

Can't be confident on 100%, but thanks to the design of how http4s parse things we needed to change `ParseResult#fromParser` only.

core/shared/src/main/scala/org/http4s/headers/Forwarded.scala
```diff
@@ -203,7 +204,7 @@ object Forwarded extends ForwardedRenderers {
       */
     def fromString(s: String): ParseResult[Host] =
       parser.parseAll(s).left.map { e =>
-        ParseFailure(s"invalid host '$s'", e.toString)
+        ParseFailure(s"invalid host '$s'", e.show)
```
armanbilge, line 207: Thanks, so then my question is why aren't we using `fromParser` here? Sorry if it's dumb one 😅

danicheg: 🤷🏻  probably this could be refactored

ChristopherDavenport (approved):

danicheg:
Yay, thanks for the review @ChristopherDavenport. I happily merge this to get it in a new release.

PR #6544, opened 2022-07-18
Retire unlawful copy method of `HttpVersion`

danicheg:
This tackles https://github.com/http4s/http4s/issues/5259

### DIFF
--- a/core/shared/src/main/scala/org/http4s/HttpVersion.scala
+++ b/core/shared/src/main/scala/org/http4s/HttpVersion.scala
@@ -25,6 +25,8 @@ import cats.parse.{Parser => P}
 import cats.syntax.all._
 import org.http4s.util._
 
+import scala.util.hashing.MurmurHash3
+
 /** HTTP's version number consists of two decimal digits separated by
   * a "." (period or decimal point). The first digit ("major version")
   * indicates the messaging syntax, whereas the second digit ("minor
@@ -41,8 +43,10 @@ import org.http4s.util._
   * HTTP Semantics, Protocol Versioning]]
   */
 // scalafix:off Http4sGeneralLinters.nonValidatingCopyConstructor; bincompat until 1.0
-final case class HttpVersion private (major: Int, minor: Int)
-    extends Renderable
+final class HttpVersion private (val major: Int, val minor: Int)
+    extends Product
+    with Serializable
+    with Renderable
     with Ordered[HttpVersion] {
   // scalafix:on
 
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/org/http4s/HttpVersion.scala
```diff
@@ -64,10 +68,51 @@ final case class HttpVersion private (major: Int, minor: Int)
...
+      case _ => false
+    }
 }
 
 object HttpVersion {
 
+  /** Unsafe constructor for [[HttpVersion]], consider
+    * using its safe version [[HttpVersion.fromVersion]] instead.
+    */
+  def apply(major: Int, minor: Int): HttpVersion = new HttpVersion(major, minor)
```
danicheg, line 107: In 0.23 this constructor is available in public API, so I did the same here. But if we want, we can easily hide it from the public. Thoughts? cc @rossabaker

danicheg:
Hmm, site generating is failing at CI. @armanbilge you might be interested in it.

armanbilge:
Thanks for the ping. It looks like your changes have trigger a pprint bug, since `HttpVersion` is include in the mdoc output on many pages. e.g. https://http4s.org/v0.23/docs/dsl.html#testing-the-service

We should try updating to latest mdoc in http4s. I couldn't ship it in sbt-typelevel b/c other projects are having bugs with the new one 🙃

satorg:
Just to clarify: the main purpose of this PR is to make `HttpVersion` pretending it is a case class but without `copy` – is that it or am I missing something?

core/shared/src/main/scala/org/http4s/HttpVersion.scala
satorg: I think, since this PR targets `main` and we don't worry about bin-compat issues in there, such a method just should not exist. Because it makes `HttpVersion` unsafe regardless of a bunch of its `from*` methods.

satorg: ..however, we could consider adding an `unsafeFromVersion` method which would resemble `apply` but apparently would make the fact it is unsafe more sounding.

armanbilge:
```
abstract sealed case class Foo private (...)
```

Note that this still generates an `unapply` method.

satorg:
> Note that this still generates an unapply method.

Correct. And this is absolutely safe in contrast to the `apply` one.

armanbilge:
Well, it's safe in terms of validation. But if we ever (for some reason) want to add another field to this class, then it become a problem. It's not excellent for binary-compatible evolution.

satorg:
> But if we ever (for some reason) want to add another field to this class, then it become a problem. It's not excellent for binary-compatible evolution.

Oh, good point. But is it really applicable to HttpVersion? And what is the goal then – to completely get rid of `unapply` or keep it while making it explicitly managed?

armanbilge:
> Oh, good point. But is it really applicable to HttpVersion?

_Probably_ not. IDK, what happens if HTTP starts introducing patch versions 😂 but it's a good thing to keep in mind in general.

> to completely get rid of `unapply` or keep it while making it explicitly managed?

We need to completely get rid of it. We can't make it explicitly managed, because the only thing that changes is the _return_ value, and there is no way to overload that.

I.e. there is no way to make:
```scala
def unapply(foo: Foo): Option[(A, B)] = ???
```
into
```scala
def unapply(foo: Foo): Option[(A, B, C)] = ???
```
without breaking everyone 😅

satorg:
> We need to completely get rid of it. We can't make it explicitly managed, because the only thing that changes is the return value, and there is no way to overload that.

Ok, that's fair. Although I don't feel the issue you described is going to happen.. But I wouldn't bet for it whatsoever :) So for the sake of safety let's go with that.

Nevertheless, I could suggest one more approach that allows to avoid re-implementing a case class from scratch, while providing additional safety in terms of `unapply`:
```scala
sealed trait Woo
  extends Product
  with Serializable
  /* with AllOtherMixIns */ {

  def value1: Int
  def value2: Int

  // no `copy` will be generated

  // Woo will still get `equal`, `hashCode`, `productArity`, etc
}

object Woo {
  // neither `apply` nor `unapply` will be generated here

  private final case class Impl(override val value1: Int, override val value2: Int) extends Woo {
    override def productPrefix = "Woo" // do not expose "Impl"
  }

  def fromInt(v1: Int, v2: Int): Option[Woo] =
    if (n1 > 0 && n2 > 0)
      Some(Impl(n1, n2))
    else
      None
}
```

PR #6551, opened 2022-07-20
0.23 -> main

armanbilge:
No conflicts.


server/shared/src/main/scala/org/http4s/server/middleware/Throttle.scala
```diff
@@ -129,9 +129,14 @@ object Throttle {
       Throttle.httpRoutes(bucket, defaultResponse[F] _)(httpRoutes)
     )
 
+  @deprecated("Use httpApp instead.", "0.23.14")
+  def httpAapp[F[_]](amount: Int, per: FiniteDuration)(httpApp: HttpApp[F])(implicit
```
danicheg, line 133: I think we can remove this

danicheg (approved):

server/shared/src/main/scala/org/http4s/server/middleware/Throttle.scala
armanbilge: Maybe we should give once cycle of deprecation, to help migration?

danicheg: 🤷🏻 Can't suggest how many users might be affected by just removing this, but this was introduced weeks ago

danicheg: I don't mind if we release this with deprecation and then remove it. Just we definitely have to remove it for 1.0 :)

ChristopherDavenport (approved):

PR #6554, opened 2022-07-21
Freshen up `Middleware` docs

### LINKED ISSUE #5086: middleware.md still refers to HttpService
Thanks to @nafg for pointing that out [here](https://github.com/http4s/http4s/pull/1693#issuecomment-900737068).

danicheg:
Fixes #5086.
I did it very straightforwardly. If you have something to add, let me know.

### DIFF
--- a/docs/docs/middleware.md
+++ b/docs/docs/middleware.md
@@ -111,7 +111,7 @@ newService.orNotFound(badRequest).unsafeRunSync()
 Let's consider Authentication middleware as an example. Authentication
 middleware is a function that takes `AuthedRoutes[F]` 
 (that translates to `AuthedRequest[F, T] => F[Option[Response[F]]]`) 
-and returns `HttpRoutes[F]` (that translates to `Request[F, T] => F[Option[Response[F]]]`). 
+and returns `HttpRoutes[F]` (that translates to `Request[F] => F[Option[Response[F]]]`). 
 There is a type defined for this in the `http4s.server` package:
 
 ```scala
--- a/docs/docs/middleware.md
+++ b/docs/docs/middleware.md
@@ -5,7 +5,7 @@ the `Request` sent to service, and/or the `Response` returned by the service. In
 some cases, such as [Authentication], middleware may even prevent the service
 from being called.
 
-At its most basic, middleware is simply a function that takes one service
+At its most basic, middleware is a function that takes one service
 and returns another. The middleware function can take any additional parameters 
 it needs to perform its task. Let's look at a simple example.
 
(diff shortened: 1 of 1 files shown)


docs/docs/middleware.md
```diff
@@ -1,14 +1,13 @@
 # Middleware
 
-A middleware is a wrapper around a [service] that provides a means of manipulating
```
danicheg, line 3: I can be biased, but explaining things using 'wrapper' things doesn't help much.

docs/docs/middleware.md
```diff
@@ -1,14 +1,13 @@
...
+A middleware is an abstraction around a [service] that provides a means of manipulating
 the `Request` sent to service, and/or the `Response` returned by the service. In
 some cases, such as [Authentication], middleware may even prevent the service
 from being called.
 
-At its most basic, middleware is simply a function that takes one `Service` as a
-parameter and returns another `Service`. In addition to the `Service`, the middleware
-function can take any additional parameters it needs to perform its task. Let's look
-at a simple example.
+At its most basic, middleware is simply a function that takes one service
```
ChristopherDavenport, line 8: The word simply here has always bugged me.

docs/docs/middleware.md
```diff
@@ -107,29 +108,35 @@ newService.orNotFound(goodRequest).unsafeRunSync()
...
 ```
 
-It is possible for the wrapped `Service` to have different `Request` and `Response`
-types than the middleware. Authentication is, again, a good example. Authentication
-middleware is an `HttpService` (an alias for `Service[Request, Response]`) that wraps an `
-AuthedService` (an alias for `Service[AuthedRequest[T], Response]`. There is a type
-defined for this in the `http4s.server` package:
+Let's consider Authentication middleware as an example. Authentication
+middleware is a function that takes `AuthedRoutes[F]` 
+(an alias for `Kleisli[OptionT[F, *], AuthedRequest[F, T], Response[F]]`)
```
ChristopherDavenport, line 113: Instead of kleisli what about explaining it in terms of `AuthedRequest[F, T] => F[Option[Response[F]]]`

danicheg: good call, will fix this

docs/docs/middleware.md
```diff
@@ -107,29 +108,35 @@ newService.orNotFound(goodRequest).unsafeRunSync()
...
 
-It is possible for the wrapped `Service` to have different `Request` and `Response`
-types than the middleware. Authentication is, again, a good example. Authentication
-middleware is an `HttpService` (an alias for `Service[Request, Response]`) that wraps an `
-AuthedService` (an alias for `Service[AuthedRequest[T], Response]`. There is a type
-defined for this in the `http4s.server` package:
+Let's consider Authentication middleware as an example. Authentication
+middleware is a function that takes `AuthedRoutes[F]` 
+(that translates to `AuthedRequest[F, T] => F[Option[Response[F]]]`) 
+and returns `HttpRoutes[F]` (that translates to `Request[F, T] => F[Option[Response[F]]]`).
```
ChristopherDavenport, line 114:
```suggestion
and returns `HttpRoutes[F]` (that translates to `Request[F] => F[Option[Response[F]]]`). 
```

ChristopherDavenport (commented):
I love it, theres a fix thats still needed but I like it a lot.

ChristopherDavenport:
My wonder I've always had is what if we generally omitted the `Request[F]` to just call in `Request` docs. This is more a discussion point than something I want to see addressed here. That way when explaining it would look like `Request => F[Option[Response]]` which hides the noise that has to do with the internal stream bodies.

docs/docs/middleware.md
danicheg: 🤦🏻  thank you for pointing this out

danicheg:
I think that hiding the `F` type parameter is absolutely legit in docs, especially when we're explaining things using pseudo-code. And vice versa, every code snippet should be compiled and ready for copy-pasting by users.

ChristopherDavenport (approved):

PR #6555, opened 2022-07-22
Rewrite hashcode computation for `Uri.Path`

### LINKED ISSUE #6538: `PathSuite.hashcode should be consistent with equality` flake
/cc @FrancescoSerra 

```
==> X org.http4s.PathSuite.hashcode should be consistent with equality0.016s munit.FailException: /home/runner/work/http4s/http4s/tests/shared/src/test/scala/org/http4s/PathSuite.scala:37
36:
37:  test("hashcode should be consistent with equality") {
38:    forAll { (a: Path, b: Path) =>
Failing seed: KgTzLMfGMbGjFL7Ip90jnCi2YhdvVuUa2fKVcDQc_UJ=
  override val scalaCheckInitialSeed = "KgTzLMfGMbGjFL7Ip90jnCi2YhdvVuUa2fKVcDQc_UJ="
Falsified after 1 passed tests.
> ARG_0: /@
> ARG_1: @/
    at munit.Assertions.fail$(Assertions.scala:277)
    at munit.FunSuite.fail(FunSuite.scala:11)
    at munit.ScalaCheckSuite.$anonfun$propToTry$4(ScalaCheckSuite.scala:104)
    at scala.util.Try$.apply(Try.scala:210)
    at munit.ScalaCheckSuite.propToTry(ScalaCheckSuite.scala:104)
    at munit.ScalaCheckSuite.$anonfun$scalaCheckPropTransform$3(ScalaCheckSuite.scala:46)
    at scala.concurrent.impl.Promise$Transformation.run(Promise.scala:475)
    at cats.effect.IOFiber.executeRunnableR(IOFiber.scala:1400)
    at cats.effect.IOFiber.run(IOFiber.scala:147)
    at cats.effect.unsafe.WorkerThread.run(WorkerThread.scala:523)
```

https://github.com/http4s/http4s/runs/7374681276?check_suite_focus=true#step:13:913

FrancescoSerra:
Closes #6538

The actual fix would be along the lines of what was discussed [here](https://github.com/http4s/http4s/discussions/5930#discussioncomment-2997587), but we'll have to wait for [cats-uri](https://github.com/typelevel/cats-uri) for a definitive solution

### DIFF
--- a/core/shared/src/main/scala/org/http4s/Uri.scala
+++ b/core/shared/src/main/scala/org/http4s/Uri.scala
@@ -553,7 +553,8 @@ object Uri extends UriPlatform {
           )
       }
 
-    implicit val http4sInstancesForPath: Order[Path] with Semigroup[Path] with Hash[Path] =
+    def http4sInstancesForPath: : Order[Path] with Semigroup[Path] = http4sInstancesForPathBinCompat
+    implicit val http4sInstancesForPathBinCompat: Order[Path] with Semigroup[Path] with Hash[Path] =
       new Order[Path] with Semigroup[Path] with Hash[Path] {
         def compare(x: Path, y: Path): Int = {
           def comparePaths[A: Order](focus: Path => A): Int =
--- a/core/shared/src/main/scala/org/http4s/Uri.scala
+++ b/core/shared/src/main/scala/org/http4s/Uri.scala
@@ -339,12 +339,14 @@ object Uri extends UriPlatform {
     private def doEquals(path: Path): Boolean =
       this.segments == path.segments && path.absolute == this.absolute && path.endsWithSlash == this.endsWithSlash
 
-    override def hashCode(): Int = {
-      var hash = segments.hashCode()
-      hash += 31 * absolute.hashCode()
-      hash += 31 * endsWithSlash.hashCode()
-      hash
-    }
+    override def hashCode(): Int =
+      // this prevents hashcode clashing when two Paths have `absolute` and `endWithSlash`
+      // asymmetric values, i.e. path1 (absolute = true, endWithSlash = false), path2 (absolute = false, endWithSlash = true)
+      31 * (
+        31 * (
+          31 * endsWithSlash.##
+        ) + absolute.##
+      ) + segments.##
 
     def render(writer: Writer): writer.type = {
       val start = if (absolute) "/" else ""
--- a/core/shared/src/main/scala/org/http4s/Uri.scala
+++ b/core/shared/src/main/scala/org/http4s/Uri.scala
@@ -55,6 +55,7 @@ import java.nio.charset.StandardCharsets
 import java.nio.charset.{Charset => JCharset}
 import scala.collection.immutable
 import scala.math.Ordered
+import scala.util.hashing.MurmurHash3
 
 /** Representation of the [[Request]] URI
   *
@@ -342,11 +343,15 @@ object Uri extends UriPlatform {
     override def hashCode(): Int =
       // this prevents hashcode clashing when two Paths have `absolute` and `endWithSlash`
       // asymmetric values, i.e. path1 (absolute = true, endWithSlash = false), path2 (absolute = false, endWithSlash = true)
-      31 * (
-        31 * (
-          31 * endsWithSlash.##
-        ) + absolute.##
-      ) + segments.##
+      {
+        var hash =
+          MurmurHash3.mix(MurmurHash3.productSeed, "Uri.Path".##)
+
+        hash = MurmurHash3.mix(hash, segments.##)
+        hash = MurmurHash3.mix(hash, absolute.##)
+        hash = MurmurHash3.mix(hash, endsWithSlash.##)
+        MurmurHash3.finalizeHash(hash, 3)
+      }
 
     def render(writer: Writer): writer.type = {
       val start = if (absolute) "/" else ""
--- a/core/shared/src/main/scala/org/http4s/Uri.scala
+++ b/core/shared/src/main/scala/org/http4s/Uri.scala
@@ -340,16 +340,13 @@ object Uri extends UriPlatform {
     private def doEquals(path: Path): Boolean =
       this.segments == path.segments && path.absolute == this.absolute && path.endsWithSlash == this.endsWithSlash
 
-    override def hashCode(): Int =
-      // this prevents hashcode clashing when two Paths have `absolute` and `endWithSlash`
-      // asymmetric values, i.e. path1 (absolute = true, endWithSlash = false), path2 (absolute = false, endWithSlash = true)
-      {
-        var hash = Path.hashSeed
-        hash = MurmurHash3.mix(hash, segments.##)
-        hash = MurmurHash3.mix(hash, absolute.##)
-        hash = MurmurHash3.mix(hash, endsWithSlash.##)
-        MurmurHash3.finalizeHash(hash, 3)
-      }
+    override def hashCode(): Int = {
+      var hash = Path.hashSeed
+      hash = MurmurHash3.mix(hash, segments.##)
+      hash = MurmurHash3.mix(hash, absolute.##)
+      hash = MurmurHash3.mix(hash, endsWithSlash.##)
+      MurmurHash3.finalizeHash(hash, 3)
+    }
 
     def render(writer: Writer): writer.type = {
       val start = if (absolute) "/" else ""
--- a/core/shared/src/main/scala/org/http4s/Uri.scala
+++ b/core/shared/src/main/scala/org/http4s/Uri.scala
@@ -54,7 +54,6 @@ import java.nio.CharBuffer
 import java.nio.charset.StandardCharsets
 import java.nio.charset.{Charset => JCharset}
 import scala.collection.immutable
-import scala.math.Ordered
 import scala.util.hashing.MurmurHash3
 
 /** Representation of the [[Request]] URI
--- a/core/shared/src/main/scala/org/http4s/Uri.scala
+++ b/core/shared/src/main/scala/org/http4s/Uri.scala
@@ -428,14 +428,7 @@ object Uri extends UriPlatform {
     def toRelative: Path = copy(absolute = false)
   }
 
-  private[Uri] trait LowPriorityImplicitsForPath {
-    implicit val pathHash: Hash[Path] = new Hash[Path] {
-      override def hash(x: Path): Int = x.##
-      override def eqv(x: Path, y: Path): Boolean = x == y
-    }
-  }
-
-  object Path extends LowPriorityImplicitsForPath {
+  object Path {
     val empty: Path = new Path(Vector.empty, absolute = false, endsWithSlash = false)
     val Root: Path = new Path(Vector.empty, absolute = true, endsWithSlash = true)
     lazy val Asterisk: Path =
@@ -560,8 +553,8 @@ object Uri extends UriPlatform {
           )
       }
 
-    implicit val http4sInstancesForPath: Order[Path] with Semigroup[Path] =
-      new Order[Path] with Semigroup[Path] {
+    implicit val http4sInstancesForPath: Order[Path] with Semigroup[Path] with Hash[Path] =
+      new Order[Path] with Semigroup[Path] with Hash[Path] {
         def compare(x: Path, y: Path): Int = {
           def comparePaths[A: Order](focus: Path => A): Int =
             compareField(x, y, focus)
@@ -573,6 +566,8 @@ object Uri extends UriPlatform {
         }
 
         def combine(x: Path, y: Path): Path = x.concat(y)
+
+        def hash(x: Path): Int = x.##
       }
 
   }
--- a/core/shared/src/main/scala/org/http4s/Uri.scala
+++ b/core/shared/src/main/scala/org/http4s/Uri.scala
@@ -428,7 +428,14 @@ object Uri extends UriPlatform {
     def toRelative: Path = copy(absolute = false)
   }
 
-  object Path {
+  private[Uri] trait LowPriorityImplicitsForPath {
+    implicit val pathHash: Hash[Path] = new Hash[Path] {
+      override def hash(x: Path): Int = x.##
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/org/http4s/Uri.scala
```diff
@@ -339,12 +339,14 @@ object Uri extends UriPlatform {
     private def doEquals(path: Path): Boolean =
       this.segments == path.segments && path.absolute == this.absolute && path.endsWithSlash == this.endsWithSlash
 
-    override def hashCode(): Int = {
-      var hash = segments.hashCode()
-      hash += 31 * absolute.hashCode()
-      hash += 31 * endsWithSlash.hashCode()
-      hash
-    }
+    override def hashCode(): Int =
```
danicheg, line 342: Since we're breaking how we calculate hashCode, I suggest calculating it in a manner as Scala 2.13 does for case classes. It should bring much fewer collisions. WDYT?

danicheg: I recently did the same for [`HttpVersion` hashCode()](https://github.com/http4s/http4s/pull/6544/files#diff-d2517b0e3afb33d3674f2249a71dd41ec8a966644a9930d334169e3753728c8cR79-R84), I can say it produces an identical result with Scala 2.13.

FrancescoSerra: cool! Do you have a preference over whether to implement it exactly as you did for `HttpVersion` or is the snippet I posted above fine?

armanbilge: I don't think `System.identityHashCode` is right, it's based on the underlying reference. E.g.
```
val x = Some(1)
val y = Some(1)
```

`x` and `y` would not have the same `identityHashCode` because they are different references.

FrancescoSerra: Gotcha. Let me fix that

armanbilge:
I'm very sorry this PR missed the release, it's a good fix and I think it could have landed but unfortunately nobody had reviewed and I was focused on https://github.com/typelevel/fs2/security/advisories/GHSA-2cpx-6pqp-wf35.

@danicheg would you be able to look again when you have a chance? Thanks!

danicheg (commented):
To be honest, I'm not sure that we're free to do that change in 0.23 and not in 1.0. But I like the technical solution in general :+1:

armanbilge:
> To be honest, I'm not sure that we're free to do that change in 0.23 and not in 1.0.

I mean, it's a bug fix, if the equality and hashcode are not consistent with each other then one of them has to change :)

danicheg:
> I mean, it's a bug fix

Sure, but we change the whole mechanism of `hashCode` calculating. So this might affect users somehow (in a pretty strange way though). But maybe I'm overthinking, so feel free to decide on your own about this one.

core/shared/src/main/scala/org/http4s/Uri.scala
```diff
@@ -339,12 +340,18 @@ object Uri extends UriPlatform {
...
+      // this prevents hashcode clashing when two Paths have `absolute` and `endWithSlash`
+      // asymmetric values, i.e. path1 (absolute = true, endWithSlash = false), path2 (absolute = false, endWithSlash = true)
+      {
+        var hash =
+          MurmurHash3.mix(MurmurHash3.productSeed, "Uri.Path".##)
```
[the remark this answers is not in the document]
FrancescoSerra: Yep, makes sense

core/shared/src/main/scala/org/http4s/Uri.scala
```diff
@@ -339,12 +340,16 @@ object Uri extends UriPlatform {
...
-    override def hashCode(): Int = {
-      var hash = segments.hashCode()
-      hash += 31 * absolute.hashCode()
-      hash += 31 * endsWithSlash.hashCode()
-      hash
-    }
+    override def hashCode(): Int =
+      // this prevents hashcode clashing when two Paths have `absolute` and `endWithSlash`
+      // asymmetric values, i.e. path1 (absolute = true, endWithSlash = false), path2 (absolute = false, endWithSlash = true)
+      {
```
armanbilge, line 346: Sorry, what is "this"? Also, can we add a test case for exactly this clashing scenario you describe? Thanks! :)

FrancescoSerra: Good point on the wording - I'll rephrase. Yeah, and I'll add the example based test for the scenario described :+1:

tests/shared/src/test/scala/org/http4s/PathSuite.scala
```diff
@@ -41,6 +45,21 @@ class PathSuite extends Http4sSuite {
...
+    val nonEmptySegmentsPathGen =
+      http4sTestingAbitraryForPath.arbitrary.suchThat(_.segments.nonEmpty)
+
+    forAll(nonEmptySegmentsPathGen) { (path0: Path) =>
+      val a = Path(path0.segments, true, false)
+      val b = Path(path0.segments, false, true)
+
+      assert(a.## != b.##)
+    }
+  }
```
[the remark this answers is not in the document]
FrancescoSerra: > I'm actually confused if this is a bug vs a poor implementation of hashCode

I'm failing to see the difference: a poor implementation is it not a bug? I mean, I can see why it's not _strictly_ a bug, but definitely an enabler of potential bugs, right?
I still think it's worth changing the generation of the hashcode, as this is definitely a sounder way of doing it compared to what it was before, and I concur with you that it's an improvement (albeit not the best possible solution, because - me being a broken record - I think that the whole implementation of `Uri.Path` as it stands is broken and we'll need cats-uri to reach us for the proper solution).

> which makes me suspicious that the other test is testing something that isn't actually true

I'm confused by your confusion :grinning: the other test is testing that with 2 different Paths we should also get 2 different hashcodes. Why is that not actually true?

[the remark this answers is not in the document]
armanbilge: > I'm failing to see the difference: a poor implementation is not a bug? I mean, I can see why it's not strictly a bug, but definitely an enabler of potential bugs, right?

It's not an enabler of bugs, it would be a bug to depend on hashcodes in this way. But certainly it's an enabler of performance degradation. For example it would lead to collisions with `HashMap` (but if the `HashMap` breaks because of this, then the `HashMap` is broken.

> I'm confused by your confusion 😀 the other test is testing that with 2 different Paths we should also get 2 different hashcodes. Why is that not actually true?

The pigeonhole principle :) 🐦 there are an infinite number of possible paths, but only 2^32 possible hashcodes. Therefore, there must exist different paths that have the same hashcode.

https://en.wikipedia.org/wiki/Pigeonhole_principle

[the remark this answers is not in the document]
FrancescoSerra: > it would be a bug to depend on hashcodes in this way

I agree with that

> The pigeonhole principle

well, yes, but what's the solution then? Not test the custom implementation of `Path.hashcode`? Arguably it's still useful, considering that we found out about the weaker implementation by doing it.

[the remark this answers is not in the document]
armanbilge: > Arguably it's still useful, considering that we found out about the weaker implementation by doing it.

Great question, see https://github.com/typelevel/cats/pull/4118#issuecomment-1023517001 for some discussion.

> Not test the custom implementation of `Path.hashcode`?

We should definitely test it! But only test things that are actually supposed to hold. For example, if the hashCodes are different, then they should be different by equality. Also, if they are equal, then the hashcodes should be the same. But asserting that the hashCodes are different just because the `Path`s are different does not hold, nor does asserting that the `Path`s are equal because the hashCodes are.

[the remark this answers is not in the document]
FrancescoSerra: > see https://github.com/typelevel/cats/pull/4118#issuecomment-1023517001 for some discussion.

I couldn't agree more: in spite of being naive, the PB test that generated the error reported originally, surfaced that tiny chance of collision which we're now investigating :smile: 

> For example, if the hashCodes are different, then the Paths should be different by equality. Also, if they are equal, then the hashcodes should be the same.

Yes, this is the right idea here - I'll change the test accordingly. Thanks for elaborating!

tests/shared/src/test/scala/org/http4s/PathSuite.scala
```diff
@@ -28,16 +28,24 @@ class PathSuite extends Http4sSuite {
...
   test("hashcode should be consistent with equality") {
     forAll { (a: Path, b: Path) =>
-      if (a == b) a.## == b.##
-      else a.## != b.##
+      assert(
+        if (a.## != b.##)
+          a != b // if the hashCodes are different, then the Paths should be different by equality
+        else
+          (a == b) && (a.## == b.##) // if the Paths are equal, then the hashcodes should be the same
+      )
```
armanbilge, line 48: I wonder if this law is expressing the same idea, the expression is so concise I'm not exactly sure 😂 in any case, I think it would be good to run the `HashTests` for `Path`.
https://github.com/typelevel/cats/blob/771c6c802f59c72dbc1be1898081c9c882ddfeb0/kernel-laws/shared/src/main/scala/cats/kernel/laws/HashLaws.scala#L30-L31

FrancescoSerra: And then, that's where you learn about something you didn't know existed :-D
Well, yes, that looks like it does!
While I'm on that, I'm thinking that maybe also the test above (`equals should be consistent with equality`) could be evicted by running `EqTests` for `Path`? https://github.com/typelevel/cats/blob/771c6c802f59c72dbc1be1898081c9c882ddfeb0/kernel-laws/shared/src/main/scala/cats/kernel/laws/EqLaws.scala
wdyt?

armanbilge: Ah, yes, sounds good! Delegating to law tests is always good.

tests/shared/src/test/scala/org/http4s/PathSuite.scala
```diff
@@ -24,30 +25,8 @@ import org.scalacheck.Prop._
...
-      assert(
-        if (a.## != b.##)
-          a != b // if the hashCodes are different, then the Paths should be different by equality
-        else
-          (a == b) && (a.## == b.##) // if the Paths are equal, then the hashcodes should be the same
-      )
-    }
-  }
+  checkAll("Hash[Path]", HashTests[Path](Hash.fromHashing).hash)
+  checkAll("Eq[Path]", EqTests[Path].eqv)
```
armanbilge, line 29: Are we missing a `Hash` instance? Maybe we should add that.

core/shared/src/main/scala/org/http4s/Uri.scala
```diff
@@ -428,7 +428,14 @@ object Uri extends UriPlatform {
     def toRelative: Path = copy(absolute = false)
   }
 
-  object Path {
+  private[Uri] trait LowPriorityImplicitsForPath {
+    implicit val pathHash: Hash[Path] = new Hash[Path] {
+      override def hash(x: Path): Int = x.##
+      override def eqv(x: Path, y: Path): Boolean = x == y
+    }
+  }
```
armanbilge, line 436: Nice! Sorry, on second thought I think we can just mixin `with Hash` to the existing instances.
https://github.com/http4s/http4s/blob/8e31e5f69b1e7b92d7bb3176a8c0719d5b4b7cc2/core/shared/src/main/scala/org/http4s/Uri.scala#L563-L564

FrancescoSerra: Right!

core/shared/src/main/scala/org/http4s/Uri.scala
```diff
@@ -560,8 +553,8 @@ object Uri extends UriPlatform {
           )
       }
 
-    implicit val http4sInstancesForPath: Order[Path] with Semigroup[Path] =
-      new Order[Path] with Semigroup[Path] {
+    implicit val http4sInstancesForPath: Order[Path] with Semigroup[Path] with Hash[Path] =
+      new Order[Path] with Semigroup[Path] with Hash[Path] {
```
armanbilge, line 557: I think something like this will fix the bincompat:

```suggestion
    def http4sInstancesForPath: : Order[Path] with Semigroup[Path] = http4sInstancesForPathBinCompat
    implicit val http4sInstancesForPathBinCompat: Order[Path] with Semigroup[Path] with Hash[Path] =
      new Order[Path] with Semigroup[Path] with Hash[Path] {
```

FrancescoSerra: Nice trick!

tests/shared/src/test/scala/org/http4s/PathSuite.scala
```diff
@@ -24,22 +24,8 @@ import org.scalacheck.Prop._
 class PathSuite extends Http4sSuite {
   checkAll("Order[Path]", OrderTests[Path].order)
   checkAll("Semigroup[Path]", SemigroupTests[Path].semigroup)
```
armanbilge, line 26: Ah, we are already running tests for `Order[Path]`, then tests for `Eq[Path]` are redundant. Actually tests for `Hash[Path]` also run the `Eq` tests as well :)

armanbilge (commented):
LGTM, just realized we are running some redundant tests :)

tests/shared/src/test/scala/org/http4s/PathSuite.scala
FrancescoSerra: well spotted! Thanks :-)

armanbilge (approved):
The JavaDoc says

> The general contract of hashCode is: ... This integer need not remain consistent from one execution of an application to another execution of the same application.

https://docs.oracle.com/javase/8/docs/api/java/lang/Object.html#hashCode()

So I'd say it's safe to change. Anyone who is using this is much more likely to care about collisions than this being consistent across http4s versions.

PR #6556, opened 2022-07-23
Deprecate `BackendBuilder#allocate`

### LINKED ISSUE #5543: Deprecate `BackendBuilder#allocate`

danicheg:
This is a follow up to #6540.
Finally fixes #5543.


armanbilge (approved):
Thank you for following-up.

PR #6557, opened 2022-07-23
Don't instantiate redundant `List` in `Path#addSegment`

danicheg:
A micro-optimization, but `addSegment` can be used on the users' side heavily, so this could bring a positive impact (yes, a tiny one) to memory consumption.


FrancescoSerra (approved):
Nice

ChristopherDavenport (approved):

PR #6564, opened 2022-07-25
Don't override `endsWithSlash` when adding an empty `Path` segment

### LINKED ISSUE #6562: Adding an empty segment to the `Path` overrides `endsWithSlash`
It turns out that
```scala
uri"/test//".path.endsWithSlash === true // segments are Vector(test, )
uri"/test/".path.addSegment("").endsWithSlash === false // segments are Vector(test, )
```
Yeah, this is yet another annoying URI thing. Sorry for that. But I think we can fix this for 0.23.
What's more, `Path#addSegments` is doing similar on adding even empty collection.
```scala
uri"/test/".path.endsWithSlash === true
uri"/test/".path.addSegments(Seq.empty).path.endsWithSlash === false
```
One like and I'll dig into fixing this.

danicheg:
Fixes #6562.
Read this only with the hiding of whitespace changes in UI!

### DIFF
--- a/core/shared/src/main/scala/org/http4s/Uri.scala
+++ b/core/shared/src/main/scala/org/http4s/Uri.scala
@@ -377,7 +377,22 @@ object Uri extends UriPlatform {
       addSegment(encoder.toSegment(segment))
 
     def addSegments(value: Seq[Path.Segment]): Path =
-      Path(this.segments ++ value, absolute = absolute || this.segments.isEmpty)
+      if (value.isEmpty) this
+      else {
+        val segments = this.segments ++ value
+        val endsWithSlash = value match {
+          case Nil | Seq(Path.Segment.empty) =>
+            this.endsWithSlash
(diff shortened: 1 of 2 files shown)


rossabaker (approved):

armanbilge (approved):
Quick fix! After merging, would you mind opening a merge 0.23->main? Thanks 🙏

danicheg:
> After merging, would you mind opening a merge 0.23->main?

hehe, I'm feeling that you're sick of merging after my PRs to 0.23 😄  I apologize for this and sure will do a merge. 🙇🏻‍♂️

armanbilge:
@danicheg not sick at all, I promise, we all do our fair share of merges here :) I was working quietly on the https://github.com/typelevel/fs2/security/advisories/GHSA-2cpx-6pqp-wf35 release train today and needed to keep the branches updated so we could release in sync.

PR #6575, opened 2022-07-27
WebSocketBuilder2: make constructor public

### LINKED ISSUE #6543: It's not possible to unit test a WebSocket application
Hi, 
I’m migrating my `Http4s` WebSocket application from `1.0.0-M23` to `1.0.0-M34`.

Now, as I shouldn’t create a `WebSocketBuilder` myself, but it is provided by the server, how can I unit test my routes if they require passing a `WebSocketBuilder` which I can’t create myself?

gvolpe:
closes #6543

### DIFF
--- a/ember-server/shared/src/main/scala/org/http4s/ember/server/EmberServerBuilder.scala
+++ b/ember-server/shared/src/main/scala/org/http4s/ember/server/EmberServerBuilder.scala
@@ -197,7 +197,7 @@ final class EmberServerBuilder[F[_]: Async] private (
               requestHeaderReceiveTimeout,
               idleTimeout,
               logger,
-              wsBuilder.wsKey,
+              wsBuilder.webSocketKey,
               enableHttp2,
             )
             .compile
@@ -222,7 +222,7 @@ final class EmberServerBuilder[F[_]: Async] private (
             requestHeaderReceiveTimeout,
             idleTimeout,
             logger,
-            wsBuilder.wsKey,
+            wsBuilder.webSocketKey,
             enableHttp2,
           )
           .compile
--- a/ember-server/shared/src/main/scala/org/http4s/ember/server/EmberServerBuilder.scala
+++ b/ember-server/shared/src/main/scala/org/http4s/ember/server/EmberServerBuilder.scala
@@ -34,8 +34,6 @@ import org.http4s.ember.server.internal.Shutdown
 import org.http4s.server.Ip4sServer
 import org.http4s.server.Server
 import org.http4s.server.websocket.WebSocketBuilder2
-import org.http4s.websocket.WebSocketContext
-import org.typelevel.vault.Key
 
 import scala.concurrent.duration._
 
@@ -178,7 +176,7 @@ final class EmberServerBuilder[F[_]: Async] private (
       sg <- sgOpt.getOrElse(Network[F]).pure[Resource[F, *]]
       ready <- Resource.eval(Deferred[F, Either[Throwable, SocketAddress[IpAddress]]])
       shutdown <- Resource.eval(Shutdown[F](shutdownTimeout))
-      wsKey <- Resource.eval(Key.newKey[F, WebSocketContext[F]])
+      wsBuilder <- Resource.eval(WebSocketBuilder2[F])
       _ <- unixSocketConfig.fold(
         Concurrent[F].background(
           ServerHelpers
@@ -187,7 +185,7 @@ final class EmberServerBuilder[F[_]: Async] private (
               port,
               additionalSocketOptions,
               sg,
-              httpApp(WebSocketBuilder2(wsKey)),
+              httpApp(wsBuilder),
               tlsInfoOpt,
               ready,
               shutdown,
@@ -199,7 +197,7 @@ final class EmberServerBuilder[F[_]: Async] private (
               requestHeaderReceiveTimeout,
               idleTimeout,
               logger,
-              wsKey,
+              wsBuilder.wsKey,
               enableHttp2,
             )
             .compile
@@ -212,7 +210,7 @@ final class EmberServerBuilder[F[_]: Async] private (
             unixSocketAddress,
             deleteIfExists,
             deleteOnClose,
-            httpApp(WebSocketBuilder2(wsKey)),
+            httpApp(wsBuilder),
             tlsInfoOpt,
             ready,
             shutdown,
@@ -224,7 +222,7 @@ final class EmberServerBuilder[F[_]: Async] private (
             requestHeaderReceiveTimeout,
             idleTimeout,
             logger,
-            wsKey,
+            wsBuilder.wsKey,
             enableHttp2,
           )
           .compile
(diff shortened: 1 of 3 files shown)


server/shared/src/main/scala/org/http4s/server/websocket/WebSocketBuilder2.scala
```diff
@@ -47,6 +48,8 @@ sealed abstract class WebSocketBuilder2[F[_]: Applicative] private (
 ) {
   import WebSocketBuilder2.impl
 
+  private[http4s] val wsKey: Key[WebSocketContext[F]] = webSocketKey
```
gvolpe, line 51: Don't know what you guys think about this, could have made it `private[http4s] val webSocketKey` directly in the `sealed abstract class` arguments, but that could probably affect binary compatibility? 

The idea is to have this accessible from `EmberServerBuilder` as there is a key per WS builder.

server/shared/src/main/scala/org/http4s/server/websocket/WebSocketBuilder2.scala
```diff
@@ -191,6 +194,9 @@ object WebSocketBuilder2 {
       webSocketKey = webSocketKey,
     )
 
+  def apply[F[_]: Applicative: Unique]: F[WebSocketBuilder2[F]] =
+    Key.newKey[F, WebSocketContext[F]].map(apply[F])
```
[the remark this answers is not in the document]
armanbilge: Aha, I take this back, I see how you are using it.

ember-server/shared/src/main/scala/org/http4s/ember/server/EmberServerBuilder.scala
```diff
@@ -178,7 +176,7 @@ final class EmberServerBuilder[F[_]: Async] private (
       sg <- sgOpt.getOrElse(Network[F]).pure[Resource[F, *]]
       ready <- Resource.eval(Deferred[F, Either[Throwable, SocketAddress[IpAddress]]])
       shutdown <- Resource.eval(Shutdown[F](shutdownTimeout))
-      wsKey <- Resource.eval(Key.newKey[F, WebSocketContext[F]])
+      wsBuilder <- Resource.eval(WebSocketBuilder2[F])
```
armanbilge, line 179: Oh, I see now. This is a nice change indeed 👍

server/shared/src/main/scala/org/http4s/server/websocket/WebSocketBuilder2.scala
```diff
@@ -177,7 +180,7 @@ sealed abstract class WebSocketBuilder2[F[_]: Applicative] private (
 }
 
 object WebSocketBuilder2 {
-  private[http4s] def apply[F[_]: Applicative](
+  def apply[F[_]: Applicative](
```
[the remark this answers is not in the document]
armanbilge: It almost seems like we should deprecate this constructor in favor of the other.

armanbilge (commented):
Thanks, I really like the changes here! I'll need to take a second passthrough to fully appreciate it :)

server/shared/src/main/scala/org/http4s/server/websocket/WebSocketBuilder2.scala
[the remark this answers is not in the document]
gvolpe: I think we could deprecate it, didn't remove it because the old `WebSocketBuilder` still uses it, but that's deprecated so... What should I set as the version on the deprecated message? 0.23.6?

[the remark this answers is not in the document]
armanbilge: I think we're onto 0.23.15?

gvolpe:
@armanbilge added a new commit with the deprecation of the old `apply`, should I amend and force-push? Or are we happy to squash the commits when merging? Forgot what the norm is here, it's been a while 😄

armanbilge:
Thanks! We just merge the raw history here :) if you don't mind, personally I don't like force-pushes, they make reviewing very difficult.

server/shared/src/main/scala/org/http4s/server/websocket/WebSocketBuilder2.scala
armanbilge: @gvolpe would you mind adjusting this one as well?

gvolpe:
Perfect, I don't mind any strategy, just to adjust to the contributing guidelines :)

PS. Apologies for the previous force-push then, did it immediately after I found out there is a check for headers that didn't pass 🥲

server/shared/src/main/scala/org/http4s/server/websocket/WebSocketBuilder2.scala
gvolpe: Sure, no problems. Make it a `private[http4s] val webSocketKey` then?

armanbilge: Actually, I wonder if it should be completely public. Otherwise you can't implement a websocket server outside of `org.http4s`?

gvolpe: You cannot do it at the moment AFAIU, but that's deeply rooted in the code: [WebSocket](https://github.com/http4s/http4s/blob/series/0.23/core/shared/src/main/scala/org/http4s/websocket/WebSocket.scala#L22)

I don't think making `sealed abstract class WebSocketBuilder2[F[_]: Applicative] private (...)` public makes any difference at the moment, but I'm not sure how I feel about making this argument a public val... No preferences, I guess.

gvolpe: I landed at that wall when I was trying implement one myself 🥲

armanbilge: Ah, didn't realize that. `private[http4s]` makes sense then.

server/shared/src/main/scala/org/http4s/server/websocket/WebSocketBuilder2.scala
```diff
@@ -177,7 +178,19 @@ sealed abstract class WebSocketBuilder2[F[_]: Applicative] private (
 }
 
 object WebSocketBuilder2 {
-  private[http4s] def apply[F[_]: Applicative](
+  @deprecated(
+    "Only used by the deprecated WebSocketBuilder. Use the other constructor if you need an instance for testing.",
+    "0.23.15",
+  )
+  def apply[F[_]: Applicative](
```
armanbilge, line 185:
```suggestion
  @deprecated(
    "Use the arg-less constructor to create a `WebSocketBuilder2` and access its key with the webSocketKey method",
    "0.23.15",
  )
  private[http4s] def apply[F[_]: Applicative](
```

armanbilge (approved):
I think this is good. Moving the key construction as part of the `WebSocketBuilder2` constructor makes a lot of sense to me, and guarantees that each builder has a unique key, which I think is right.

Thanks for tackling this!

danicheg (approved):
Nice!

armanbilge:
I had another thought about this PR:
- https://github.com/http4s/http4s/issues/6590

PR #6578, opened 2022-07-28
Don't backtrack in request/response prelude parsing

TimWSpence:
Don't backtrack in request/response prelude parsing for better performance.

Also resolves https://github.com/http4s/http4s/issues/6580 for 1.0.x

### DIFF
--- a/ember-core/shared/src/main/scala/org/http4s/ember/core/Parser.scala
+++ b/ember-core/shared/src/main/scala/org/http4s/ember/core/Parser.scala
@@ -159,7 +159,7 @@ private[ember] object Parser {
 
     object ReqPrelude {
 
-      class ReqPreludeParser[F[_]] {
+      class Parser[F[_]] {
 
         var idx = 0
         var state: Byte = 0
@@ -236,79 +236,6 @@ private[ember] object Parser {
         }
       }
 
-      // // Method SP URI SP HttpVersion CRLF - REST
-      // def parsePrelude[F[_]](message: Array[Byte], maxHeaderSize: Int)(implicit
-      //     F: MonadThrow[F]
-      // ): F[Either[Unit, ReqPrelude]] = {
-      //   var idx = 0
-      //   var state: Byte = 0
-      //   var complete = false
-
-      //   var throwable: Throwable = null
-      //   var method: Method = null
-      //   var uri: Uri = null
-      //   var httpVersion: HttpVersion = null
-      //   val upperBound = Math.min(message.size - 1, maxHeaderSize)
-
-      //   var start = 0
-      //   while (!complete && idx <= upperBound) {
-      //     val value = message(idx)
-      //     (state: @switch) match {
-      //       case 0 =>
-      //         if (value == space) {
-      //           Method.fromString(new String(message, start, idx - start)) match {
-      //             case Left(e) =>
-      //               throwable = e
-      //               complete = true
-      //             case Right(m) =>
-      //               method = m
-      //           }
-      //           start = idx + 1
-      //           state = 1
-      //         }
-      //       case 1 =>
-      //         if (value == space) {
-      //           Uri.fromString(new String(message, start, idx - start)) match {
-      //             case Left(e) =>
-      //               throwable = e
-      //               complete = true
-      //             case Right(u) =>
-      //               uri = u
-      //           }
-      //           start = idx + 1
-      //           state = 2
-      //         }
-      //       case 2 =>
-      //         if (value == lf && (idx > 0 && message(idx - 1) == cr)) {
-      //           HttpVersion.fromString(new String(message, start, idx - start - 1)) match {
-      //             case Left(e) =>
-      //               throwable = e
-      //               complete = true
-      //             case Right(h) =>
-      //               httpVersion = h
-      //           }
-      //           complete = true
-      //         }
-      //     }
-      //     idx += 1
-      //   }
-
-      //   if (throwable != null)
-      //     F.raiseError(
-      //       ParsePreludeError(
-      //         throwable.getMessage(),
-      //         Option(throwable),
-      //         Option(method),
-      //         Option(uri),
-      //         Option(httpVersion),
-      //       )
-      //     )
-      //   else if (method == null || uri == null || httpVersion == null)
-      //     ().asLeft.pure[F]
-      //   else
-      //     ReqPrelude(method, uri, httpVersion, idx).asRight.pure[F]
-      // }
-
       final case class ParsePreludeError(
           message: String,
           caused: Option[Throwable],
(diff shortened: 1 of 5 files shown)


danicheg (commented):
Sorry for grumbling, but if this aimed to improve performance, could we get some numbers ensuring that?

ChristopherDavenport:
> Sorry for grumbling, but if this aimed to improve performance, could we get some numbers ensuring that?

This addresses a problem where by not having reentrant behavior we go into recursive behavior. If this can work as it is, it will be a better long-term approach to the problem. Single-pass parsing is ideal and ideally this will go through in 1 pass if the buffer has enough info, so we can for the most part consider it from that perspective. But other parties can send their buffers in any shape they want and be spec compliant.

TimWSpence:
I think TCP send windows also mean we can see arbitrary chunking, regardless of what the other parties sent at the application layer. Theoretically at least - I have no idea how often that's a real problem 😅

ember-core/shared/src/main/scala/org/http4s/ember/core/Parser.scala
```diff
@@ -62,26 +66,52 @@ private[ember] object Parser {
...
 
-    def parseHeaders[F[_]](message: Array[Byte], initIndex: Int, maxHeaderSize: Int)(implicit
+    final case class ParserState(
+        idx: Int,
+        state: Boolean,
+        throwable: Throwable,
+        complete: Boolean,
+        chunked: Boolean,
+        contentLength: Option[Long],
+        headers: ListBuffer[Header.Raw],
```
ChristopherDavenport, line 83: This is a mutable variable. We do not have so many headers, this cant be passed in its immutable variant.

ember-core/shared/src/main/scala/org/http4s/ember/core/Parser.scala
```diff
@@ -62,26 +66,52 @@ private[ember] object Parser {
...
+        name: String,
+        start: Int,
+    )
+
+    object ParserState {
+      def initial: ParserState = ParserState(
```
ChristopherDavenport, line 89: Why not just type them to options rather than introduce nulls?

ember-core/shared/src/main/scala/org/http4s/ember/core/Parser.scala
TimWSpence: Yeah I'd wondered about that. Will fix 👍

ember-core/shared/src/main/scala/org/http4s/ember/core/Parser.scala
TimWSpence: Mainly to make the change as non-invasive as possible. I can change it if you prefer?

ember-core/shared/src/main/scala/org/http4s/ember/core/Parser.scala
TimWSpence: Did you just mean `List` sorry? Currently the tests fail because the order of the headers is now reversed. Do we guarantee that we  preserve the request order of headers?

wjoel: If we agree that we don't guarantee anything about the order of headers, I could pick up https://github.com/http4s/http4s/pull/6201 again. :thinking:

armanbilge: Glad you saw this @wjoel, I was exactly thinking about that PR as well :)

ChristopherDavenport: Reverse when you complete. Shouldn't make a noticeable difference.

ember-core/shared/src/main/scala/org/http4s/ember/core/Parser.scala
ChristopherDavenport: Yeah, its probably smoother if the state is that its Null or a value. There are combinators on Option to do that, which is useful as you can just convert it to the same types inside the function body, but be with good layers outside.
.orNull I think its called.

ChristopherDavenport (approved):

danicheg (approved):

armanbilge:
Sorry if it's a dumb question, why is this PR targeted at 1.x instead of 0.23.x?

armanbilge:
I made a hast attempt rebase it onto series/0.23 in 2d75971d95b7853f36786bd95c798c7d2a0d6fb5 and it passes all tests. Unfortunately I somehow rebased it onto a stale series/0.23 so there are some merge conflicts 🤦 😅 

@TimWSpence would you be able to do it properly and force-push / retarget this PR? 🙏 thanks for all your work on this!

TimWSpence:
> I made a hasty attempt to rebase it onto series/0.23 in 2d75971d95b7853f36786bd95c798c7d2a0d6fb5 and it passes all tests. Unfortunately I somehow rebased it onto a stale series/0.23 so there are some merge conflicts 🤦 😅 
> 
> @TimWSpence would you be able to do it properly and force-push / retarget this PR? 🙏 thanks for all your work on this!

Haha sure! I had wondered about 0.23 - how are we keeping changes in sync between 1.0 and 0.23?

TimWSpence:
@armanbilge I _think_ that's it done 😅

armanbilge:
Woops we may need to add some bincompat filters :P those are all private implementation things right?

```
[error] ember-core: Failed binary compatibility check against org.http4s:http4s-ember-core_sjs1_3:0.23.8 (e:info.apiURL=https://s01.oss.sonatype.org/service/local/repositories/snapshots/archive/org/http4s/http4s-docs_3/0.23.14-110-0cdfef6-SNAPSHOT/http4s-docs_3-0.23.14-110-0cdfef6-SNAPSHOT-javadoc.jar/!/index.html, e:info.versionScheme=early-semver)! Found 2 potential problems (filtered 105)
[error]  * method recurseFind(Array[Byte],java.lang.Object,Int,scala.Function1,scala.Function1,cats.MonadError)java.lang.Object in object org.http4s.ember.core.Parser#MessageP does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("org.http4s.ember.core.Parser#MessageP.recurseFind")
[error]  * class org.http4s.ember.core.Parser#Response#RespPrelude#RespPreludeError is declared final in current version
[error]    filter with: ProblemFilters.exclude[FinalClassProblem]("org.http4s.ember.core.Parser$Response$RespPrelude$RespPreludeError")
[info] tests: mimaPreviousArtifacts is empty, not analyzing binary compatibility.
[info] client-testkit: mimaPreviousArtifacts is empty, not analyzing binary compatibility.
[info] js-artifact-size-test: mimaPreviousArtifacts is empty, not analyzing binary compatibility.
[error] java.lang.RuntimeException: Failed binary compatibility check against org.http4s:http4s-ember-core_sjs1_3:0.23.8 (e:info.apiURL=https://s01.oss.sonatype.org/service/local/repositories/snapshots/archive/org/http4s/http4s-docs_3/0.23.14-110-0cdfef6-SNAPSHOT/http4s-docs_3-0.23.14-110-0cdfef6-SNAPSHOT-javadoc.jar/!/index.html, e:info.versionScheme=early-semver)! Found 2 potential problems (filtered 105)
```

armanbilge (approved):
Thanks for getting this over the finish line!! That's one more ember enhancement on 0.23 🚀

Issue #6561, opened 2022-07-24
Proposal: Redesign of `EntityEncoders` on using `Charset` in implicit scope


danicheg:
This is a follow-up to a started discussion in https://github.com/http4s/http4s/pull/6528#discussion_r928091805.
Since `Charset` isn't a typeclass, using it in implicit scope isn't that accurate. Though I haven't any ideas besides moving it to the explicit scope and creating an object with predefined UTF-8-based encoders (so we won't ask for `Charset` through an implicit scope within there). But yeah, this definitely will break a few existing implicit encoders. And sure, we can do such things only for 1.0. What are your thoughts about it? I can be super wrong here, so don't hesitate to point this out.

PR #6585, opened 2022-07-31
Proposal: Retire `Charset` from the implicit scope in `EntityEncoder`

danicheg:
This addresses #6561.
The whole point is that `Charset` isn't typeclass, so it shouldn't appear in implicit scope. Though this breaks binary compatibility, this was aimed to don't fail compilation for users who use default `EntityEncoder` with `UTF-8`. 
Also, this wasn't discussed at all, so don't hesitate to point out that I'm wrong.

### DIFF
--- a/core/shared/src/main/scala/org/http4s/EntityEncoder.scala
+++ b/core/shared/src/main/scala/org/http4s/EntityEncoder.scala
@@ -25,7 +25,6 @@ import fs2.Chunk
 import fs2.Stream
 import fs2.io.file.Files
 import fs2.io.readInputStream
-import org.http4s.Charset.`UTF-8`
 import org.http4s.headers._
 import org.http4s.multipart.Multipart
 import org.http4s.multipart.MultipartEncoder
@@ -73,7 +72,7 @@ trait EntityEncoder[+F[_], A] { self =>
     }
 }
 
-object EntityEncoder {
+object EntityEncoder extends Utf8EntityEncoders {
   type Pure[A] = EntityEncoder[fs2.Pure, A]
   object Pure {
     def apply[A](implicit ev: EntityEncoder.Pure[A]): EntityEncoder.Pure[A] = ev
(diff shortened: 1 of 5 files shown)


core/shared/src/main/scala/org/http4s/EntityEncoder.scala
```diff
@@ -231,3 +229,16 @@ object EntityEncoder {
...
+    EntityEncoder.stringEncoder(Charset.`UTF-8`)
+
+  implicit val utf8CharArrayEncoder: EntityEncoder.Pure[Array[Char]] =
+    utf8StringEncoder.contramap(new String(_))
+
+  implicit def utf8ReaderEncoder[F[_]: Sync, R <: Reader]: EntityEncoder[F, F[R]] =
+    EntityEncoder.readerEncoder[F, R](Charset.`UTF-8`)
+}
+
+object Utf8EntityEncoders
```
armanbilge, line 244: What is the reasoning to keep these separate?

armanbilge (commented):
Absolutely agree, the use of `implicit` is not appropriate for `Charset`. Keeping UTF8-based implicits as the default seems reasonable to me.

Note that we can actually make this change binary-compatibly. But it would be very subtly source-breaking for anyone using a non-UTF8 implicit charset, so better for 1.x.

core/shared/src/main/scala/org/http4s/UrlForm.scala
```diff
@@ -100,14 +99,15 @@ object UrlForm {
...
+  def entityEncoder(charset: Charset = Charset.`UTF-8`): EntityEncoder.Pure[UrlForm] =
+    EntityEncoder
+      .stringEncoder(charset)
       .contramap[UrlForm](encodeString(charset))
       .withContentType(`Content-Type`(MediaType.application.`x-www-form-urlencoded`, charset))
 
   implicit def entityDecoder[F[_]](implicit
       F: Concurrent[F],
-      defaultCharset: Charset = `UTF-8`,
+      defaultCharset: Charset = Charset.`UTF-8`,
```
bplommer, line 110: This shouldn't be an implicit any more right?

danicheg: Yeah, 💯 . I've just tried to fix `EntityEncoder` in this PR.

core/shared/src/main/scala/org/http4s/EntityEncoder.scala
danicheg: Hmm, I haven't a strong opinion of this. Just this seemed to be more modular for me.

armanbilge: Do you have an idea of how it would be used? If it's just for keeping our sources tidy then I would prefer we keep `Utf8EntityEncoders` package-private to avoid expanding our API surface unnecessarily.

PR #6586, opened 2022-07-31
Fix some scaladocs

danicheg:


core/shared/src/main/scala/org/http4s/EntityEncoder.scala
```diff
@@ -58,7 +58,7 @@ trait EntityEncoder[F[_], A] { self =>
       override def headers: Headers = self.headers
     }
 
-  /** Get the [[org.http4s.headers.Content-Type]] of the body encoded by this [[EntityEncoder]],
+  /** Get the [[org.http4s.headers.`Content-Type`]] of the body encoded by this [[EntityEncoder]],
```
armanbilge, line 61: Actually this breaks it, see https://github.com/http4s/http4s/pull/4027.

danicheg: Somewhere is missing backtick support. ``` org.http4s.headers.`Content-Type` ``` is a legit reference from my way of thinking. I've opened the https://github.com/sbt/sbt-unidoc/issues/124. Let's see if we can fix this somewhere.

armanbilge: You are right about that. The good news is that only the scaladoc parser has an issue here. The generate scaladoc website renders the link correctly.
https://www.javadoc.io/static/org.http4s/http4s-docs_2.13/0.23.14/org/http4s/EntityEncoder.html#contentType:Option[org.http4s.headers.Content-Type]

danicheg: So, by a twist of fate, I've created a PR to `scalac` https://github.com/scala/scala/pull/10101. If someone likes it, please give it some reactions. Maybe this will bring some attention to maintainers :)

armanbilge: Ha ha, incredible follow-through! Dropped my reactions :)

danicheg:
@armanbilge aaand it's alive. Can you have a look at this once again?

armanbilge (approved):
This is lovely!! Nice job 😃

Issue #6558, opened 2022-07-23
Ember doesn't process websocket messages on time


grzegorz-bielski:
Discussed on Discord: https://discord.com/channels/632277896739946517/632286375311573032/1000496748004921506

When more than 1 message is send to the http4s backed by Ember at once, the subsequent messages seems to be stuck in some internal queue. 

Given an echo server like this:
```Scala
object WSExample extends IOApp.Simple with Http4sDsl[IO]:
  def run =
    for
      _ <- EmberServerBuilder
        .default[IO]
        .withHostOption(Host.fromString("0.0.0.0"))
        .withPort(Port.fromInt(3000).get)
        .withHttpWebSocketApp { builder =>
          HttpRoutes.of[IO] { case GET -> Root / "endpoint" =>
              val inOut: Pipe[IO, WebSocketFrame, WebSocketFrame] =
                _.map { m =>
                  println(m.toString)
                  m
                }
              builder.build(inOut)
            }
            .orNotFound
        }
        .withErrorHandler { e =>
          IO.println("Could not handle a request" -> e) *> InternalServerError()
        }
        .build
        .use(_ => IO.never)
        .void
    yield ()
```
and `index.html` with a script like this, that runs in Chrome:
```JavaScript
const socket = new WebSocket("ws://localhost:3000/endpoint");
socket.addEventListener('open', event => {
    socket.send("1")
    socket.send("2")
    socket.send("3")
})

socket.addEventListener('message', event => {
  console.log('Message from server ', event.data)
})
```
Even though all messages: `"1"`, `"2"` and `"3"` are correctly send to the server, only the `"1"` reaches the  `inOut` pipe.
But after executing `socket.send("4")`  in the JS console, the `"2"` suddenly appears in the `inOut`'s `println`.

Tested on `0.23.13` `http4s-ember-server`. 
Everything seems to be working fine on `0.23.12` `http4s-blaze-server`.

armanbilge:
I had similar problems in https://github.com/http4s/http4s-dom/pull/93#issuecomment-1025237877 (although only in CI, I was unable to reproduce locally) and I also heard a similar report from @buntec.

armanbilge:
Also linking to https://github.com/http4s/http4s/issues/6420#issuecomment-1138864091, possibly related?

Issue #6420, opened 2022-05-26
ember-server hanging on cancellation


armanbilge:
Running into this in https://github.com/http4s/http4s-jdk-http-client/pull/645.

I extracted the test into a scala-cli runnable `IOApp` and did a fiber dump. They are both here:
https://gist.github.com/armanbilge/ec958260fc4bd0fed34e86c8c5091035

The fiber dump seems to indicate a hang here:
```
cats.effect.IOFiber@5f03cbbf WAITING
 ├ delay @ fs2.io.net.SocketCompanionPlatform$AsyncSocket.$anonfun$readChunk$1(SocketPlatform.scala:123)
 ├ async @ fs2.io.net.SocketCompanionPlatform$AsyncSocket.readChunk(SocketPlatform.scala:117)
 ├ flatMap @ fs2.io.net.SocketCompanionPlatform$BufferedReads.$anonfun$read$1(SocketPlatform.scala:85)
 ├ delay @ fs2.io.net.SocketCompanionPlatform$BufferedReads.$anonfun$withReadBuffer$1(SocketPlatform.scala:54)
 ├ release @ org.http4s.jdkhttpclient.CompletableFutureTerminationApp$.stallingServerR(bug.scala:182)
 ├ release @ org.http4s.jdkhttpclient.CompletableFutureTerminationApp$.stallingServerR(bug.scala:182)
 ├ release @ org.http4s.jdkhttpclient.CompletableFutureTerminationApp$.stallingServerR(bug.scala:182)
 ╰ apply @ org.http4s.jdkhttpclient.CompletableFutureTerminationApp$.run(bug.scala:70)
```

However, in an ember-client cancellation bug the fiber dump also pointed us to that exact same line, but it was a red herring. See:
- https://github.com/http4s/http4s/issues/6058#issuecomment-1043299750

amesgen:
Another comletely different, but more minimal example of ember-server hanging up, minimized from a test in the linked issue above (https://github.com/http4s/http4s-jdk-http-client/pull/645): Add

```scala
  fixture.test("send and receive messages") { case (server, dispatcher) =>
    for {
      client <- createClient(
        URI.create(s"ws://${server.address.getHostName}:${server.address.getPort}/ws-echo"),
        dispatcher,
      )
      _ <- client.connect
      _ <- (client.send("foo") *> IO.println("sent sth")).replicateA(10).void
      msgs <- (client.messages.take <* IO.println("received sth")).replicateA(10)
[...]

PR #6587, opened 2022-07-31
Parse all WebSocket frames in a `Chunk` in ember-server

buntec:
This is a fix for https://github.com/http4s/http4s/issues/6558.

The bug can be observed whenever a single `Chunk` of bytes contains more than one frame. In this case the current version decodes the first frame using `bufferToFrame`, emits it and recurses via `Pull.output1(value) >> go(next, remaining)`. But `go` will only continue when more data is read from the socket. This explains why sending messages "1", "2", "3" will only show "1", because the first chunk contains all three message. Messages "2" and "3" will only be decoded when further messages arrive, which might be never. The solution is simple: decode as many frames as possible from every chunk before recursing.

### DIFF
--- a/ember-server/shared/src/main/scala/org/http4s/ember/server/internal/WebSocketHelpers.scala
+++ b/ember-server/shared/src/main/scala/org/http4s/ember/server/internal/WebSocketHelpers.scala
@@ -48,8 +48,8 @@ import scodec.bits.ByteVector
 
 import java.io.IOException
 import java.nio.ByteBuffer
-import scala.concurrent.duration.Duration
 import scala.collection.mutable.ArrayBuffer
+import scala.concurrent.duration.Duration
 
 object WebSocketHelpers {
 
--- a/ember-server/shared/src/main/scala/org/http4s/ember/server/internal/WebSocketHelpers.scala
+++ b/ember-server/shared/src/main/scala/org/http4s/ember/server/internal/WebSocketHelpers.scala
@@ -204,7 +204,7 @@ object WebSocketHelpers {
               val frames = ArrayBuffer.empty[WebSocketFrame]
               var frame = nonClientTranscoder.bufferToFrame(byteBuffer)
               while (frame != null) {
-                frames.addOne(frame)
+                frames += frame
                 // We need to slice b/c `bufferToFrame` does absolute reads.
                 byteBuffer = byteBuffer.slice()
                 frame = nonClientTranscoder.bufferToFrame(byteBuffer)
--- a/ember-server/shared/src/main/scala/org/http4s/ember/server/internal/WebSocketHelpers.scala
+++ b/ember-server/shared/src/main/scala/org/http4s/ember/server/internal/WebSocketHelpers.scala
@@ -195,26 +195,35 @@ private[internal] object WebSocketHelpers {
       def go(rest: Stream[F, Byte], acc: Array[Byte]): Pull[F, WebSocketFrame, Unit] =
         rest.pull.uncons.flatMap {
           case Some((chunk, next)) =>
-            val buffer = acc ++ chunk.toArray[Byte]
-            // A single chunk might contain multiple frames
-            // but `bufferToFrame` decodes at most one, so we
-            // call it repeatedly until all frames in the buffer are decoded.
-            val frames = ArrayBuffer.empty[WebSocketFrame]
-            var byteBuffer = ByteBuffer.wrap(buffer)
-            var frame = nonClientTranscoder.bufferToFrame(byteBuffer)
-            while (frame != null) {
-              frames += frame
-              // We need to slice b/c `bufferToFrame` does absolute reads.
-              byteBuffer = byteBuffer.slice()
-              frame = nonClientTranscoder.bufferToFrame(byteBuffer)
-            }
-            if (frames.nonEmpty) {
-              val remaining = new Array[Byte](byteBuffer.remaining())
-              byteBuffer.get(remaining)
-              Pull.output(Chunk.array(frames.toArray)) >> go(next, remaining)
-            } else {
-              go(next, buffer)
-            }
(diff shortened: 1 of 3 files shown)


buntec:
Added unit test as suggested here https://github.com/http4s/http4s/issues/6420#issuecomment-1138864091

ember-server/shared/src/main/scala/org/http4s/ember/server/internal/WebSocketHelpers.scala
```diff
@@ -193,15 +194,29 @@ object WebSocketHelpers {
...
+              val frames = ArrayBuffer.empty[WebSocketFrame]
+              var frame = nonClientTranscoder.bufferToFrame(byteBuffer)
+              while (frame != null) {
+                frames += frame
+                // We need to slice b/c `bufferToFrame` does absolute reads.
+                byteBuffer = byteBuffer.slice()
+                frame = nonClientTranscoder.bufferToFrame(byteBuffer)
+              }
+              Chunk.array(frames.toArray)
+            })
```
armanbilge, line 213: I wonder if we should write this as a tail-recursion, particularly in `F[_]`. This way if there is a big chunk containing multiple frames the runtime can auto-cede during parsing and avoid hogging the thread, which is better for fairness.

Of course there is a performance trade-off here, so ideally someone else should weigh in :) mostly I just wanted to point this out.

armanbilge (approved):
This is fantastic!! Thank you so much for chasing this down and the excellent explanation :)

ember-server/shared/src/main/scala/org/http4s/ember/server/internal/WebSocketHelpers.scala
ChristopherDavenport: This should be tail recursion. This offers relatively low performance benefit by comparison to putting a var in a position like this. 

Functional closures. But this closure is in fact cheating with a mutable in a non-pure position. 

So, this can move to a function that is pure OR this should be tail-recursive. But mutability should not be in place like this.

ChristopherDavenport (changes requested):
Comment above - The fix concept is good, but mutability introduction in a pure and functional context is not what we want for this.

buntec:
Thank you for the feedback! After taking a second look, I don't see the point of suspending `bufferToFrame` in `F.delay`. We can even get rid of the `Async` constraint and replace it with `Temporal`.  The `Some((chunk, next))` case is now a pure function (despite `byteBuffer` being a `var`).

ChristopherDavenport:
If this is pure, what happens in bufferToFrame if the bytes are not a frame?

buntec:
You are right, `bufferToFrame` might throw. However, `flatMap` on `Pull` explicitly handles this case for us:
```
    * - If evaluating `f(r)` to build the pull throws an exception, the result
    *   is a pull that fails with that exception.
```
Is this not the behavior we want here? I guess we could use `F.catchOnly` and be more explicit about the types of exceptions that can occur, but this wasn't the case before when it was simply wrapped in `F.delay`.

armanbilge:
Wow, I missed the fact that `bufferToFrame` may throw. Ideally, we should refactor _that_ API...

Saving that for another day, what we should use here is `F.catchNonFatal`. Like `F.delay` it will capture exceptions but it will not suspend side-effects.

buntec:
Yep, except `catchNonFatal` is conveniently built into `flatMap` of `Pull` so it would be redundant, I think.

armanbilge:
`IO` also catches in `flatMap`, but I'm of the opinion we should avoid relying on that.  Here's a quote from Fabio on the matter.
> I'm not blindly about principles, and I do support, for example, the fact that `IO` catches in `map` and `flatMap` as a safety net, but I see it more as an alert for a problem you'd then need to fix by wrapping properly, rather than something you should expect to be able to actively exploit

https://discord.com/channels/632277896739946517/632310980449402880/943562847517028444

buntec:
I've prepared the change but before I push it, I have to say that it does feel a bit strange to introduce the overhead of `Pull.eval(F.catchNonFatal{...}).flatMap{...}` just because we don't want to rely on a documented feature of a stable API. I would fully agree with the advice in a context where someone working in `F` relies on it because they know they will later plug in `IO`. That being said, I trust your better judgement here and my goal is to get this bug fixed so I'm happy to comply :blush:.

ChristopherDavenport:
To better explain let's take a different example.

I know the implementation of Functor catches since I'll use it with IO in my use case.

```
def with functor[F[_]: Functor, A, B](fa: F[A]) : F[B] = fa.map(_ => ???)
```

Now used with IO this will work as intended, but will fail with Id, Option, List, etc.

The point here is that the laws that govern the types we are leveraging are what we should respect not any given implementation we may be leveraging. In this case exceptional cases are masked by an implementation and not conveyed in the types.

The types should be able to communicate everything about what is being done. Getting as broad as arbitrary asynchronous  side-effects.

buntec:
I agree 100%. In fact, your example is exactly what I had in mind in my previous comment. I still think this case is different because the `flatMap` in question is not on `F` but on `Pull`, which is a concrete type that happens to offer the guarantee that `flatMap` catches. Consequently, we don't rely on anything not conveyed by the types involved. If you still prefer to wrap in `F.catchNonFatal` just let me know and I will push the required changes.

armanbilge:
FWIW actually I think we can do `ApplicativeError[Pull[...]].catchNonFatal(...)`. That will delegate to `pure` which should be more efficient.

ChristopherDavenport:
Either with F or Pull please catch from ApplicativeThrow.

I think overall any difference should be minimal while long term that clarity will help the next folks reading this. We can always revisit later.

ChristopherDavenport (approved):

armanbilge (approved):
Thanks so much for fixing this!

Issue #5988, opened 2022-02-02
Unify server/client websocket models


armanbilge:
Some discussion in https://github.com/http4s/http4s/pull/5520.

PR #6594, opened 2022-08-06
Always respond to client close frame with 1000 "normal closure"

yurique:
We've been hunting this down in Discord with @armanbilge and this seems to be the right thing to do :)

The problem: when connecting to a websocket endpoint served by ember, and then disconnecting from it (`ws.close()` in the browser), there is an exception thrown in the browser: 

```
WebSocket connection to 'ws://...' failed: Received a broken close frame containing a reserved status code.
```

Another client (Dark WebSocket terminal) also reports an abnormal closure: `Close status: 1006 (Abnormal Closure)`

After some digging, we realized we are sending back the close code that the client sends in originally while initiating the connection termination, which happens to be `1005` and is not allowed to be used by the endpoint:

```
1005 is a reserved value and MUST NOT be set as a status code in a
      Close control frame by an endpoint.  It is designated for use in
      applications expecting a status code to indicate that no status
      code was actually present.
```
(@armanbilge found this, I don't have a link :) )
edit: that's the spec actually 😜 
https://www.rfc-editor.org/rfc/rfc6455.html#section-7.4.1

So in this PR we're changing the close code to always be `1000` - `CLOSE NORMAL`. 

This resolves the issue and the clients are now happy with how ember terminates the connection.

### DIFF
--- a/ember-server/jvm/src/test/scala/org/http4s/ember/server/EmberServerWebSocketSuite.scala
+++ b/ember-server/jvm/src/test/scala/org/http4s/ember/server/EmberServerWebSocketSuite.scala
@@ -91,7 +91,7 @@ class EmberServerWebSocketSuite extends Http4sSuite with DispatcherIOFixture {
       messages: Queue[IO, String],
       pongs: Queue[IO, String],
       remoteClosed: Deferred[IO, Unit],
-      remoteCloseCode: Deferred[IO, Int],
+      closeCode: Deferred[IO, Int],
       client: WebSocketClient,
   ) {
     def connect: IO[Unit] =
@@ -113,7 +113,7 @@ class EmberServerWebSocketSuite extends Http4sSuite with DispatcherIOFixture {
       queue <- Queue.unbounded[IO, String]
       pongQueue <- Queue.unbounded[IO, String]
       remoteClosed <- Deferred[IO, Unit]
-      remoteCloseCode <- Deferred[IO, Int]
+      closeCode <- Deferred[IO, Int]
       client = new WebSocketClient(target) {
         override def onOpen(handshakedata: ServerHandshake): Unit = {
           val fa = waitOpen.complete(None)
@@ -123,7 +123,7 @@ class EmberServerWebSocketSuite extends Http4sSuite with DispatcherIOFixture {
         override def onClose(code: Int, reason: String, remote: Boolean): Unit = {
           val fa = waitOpen
             .complete(Some(new Throwable(s"closed: code: $code, reason: $reason")))
-            .attempt >> remoteCloseCode.complete(code) >> waitClose.complete(None)
+            .attempt >> closeCode.complete(code) >> waitClose.complete(None)
           dispatcher.unsafeRunSync(fa)
           ()
         }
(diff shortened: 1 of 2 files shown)


yurique:
Yeah I looked at the suite, but it seemed to me to be on a higher level of abstraction. But I might have missed something.

yurique:
I think I have an idea.

yurique:
And I think it worked.

armanbilge:
Note that the test passes on `series/0.23` without the fix in this PR. I'm trying to see how we can write a test that will fail before the change. Seems that sending `1005` doesn't really make sense, this is something that the decoder sets if there is no data in the close frame.

https://github.com/http4s/http4s/blob/f2bb429d8fadbcc6ec76ed72c6be8d98cc6c2701/core/shared/src/main/scala/org/http4s/websocket/WebSocketFrame.scala#L105-L108

yurique:
Maybe in the tests, the java websocket client is also sending the `1000` code when closing the connection (Unlike Chrome, which sends `1005`), thus the server also sends it back?

yurique:
Oh, re-read your message. So you think the browser might not be sending the code at all and it gets decoded as `1005`?

yurique:
I tried changing the test to make the client send a non-1000 code, and the test failed. 
I think the java ws client might be giving us the code **it** sent, not the one it received from the server 🤔 

image
image

armanbilge:
> So you think the browser might not be sending the code at all and it gets decoded as 1005?

Right, exactly. I tried sending 1005 with the Java Websocket client used for testing ember and it throws saying 1005 isn't transmitted over the wire. This corresponds with the linked implementation of the decoder in http4s.

That's a good idea to send a code besides 1000 :)

yurique:
> That's a good idea to send a code besides 1005 :)

Well it yelled at me, told me I can't send `1005` either xD

ember-server/jvm/src/test/scala/org/http4s/ember/server/EmberServerWebSocketSuite.scala
```diff
@@ -90,6 +91,7 @@ class EmberServerWebSocketSuite extends Http4sSuite with DispatcherIOFixture {
       messages: Queue[IO, String],
       pongs: Queue[IO, String],
       remoteClosed: Deferred[IO, Unit],
+      remoteCloseCode: Deferred[IO, Int],
```
armanbilge, line 94: I guess this should just be `closeCode` /shrug

armanbilge:
Eh, I'm inclined to approve as-is. I don't think we really test what we want here. AFAICT we don't even have access to the close code on the server side so we can't assert against that either.

Reading some of the scaladoc didn't exactly instill confidence 😅 I think all of this needs a serious revisit before 1.0 Linking to https://github.com/http4s/http4s/issues/5988.
https://github.com/http4s/http4s/blob/f2bb429d8fadbcc6ec76ed72c6be8d98cc6c2701/server/shared/src/main/scala/org/http4s/server/websocket/WebSocketBuilder2.scala#L149-L155

yurique:
I looked at the sources a bit. 
I think we're get the close code from this:
image

the call to `onWebsocketClosing` endup updating our `Deferred`. And all sites that call `flushAndClose` seem to pass along the provided `code`. Hard to say 100%, though.

I've renamed `remoteCloseCode` -> `closeCode` – kind of neither local nor remote ¯\_(ツ)_/¯

armanbilge (approved):

ChristopherDavenport (approved):

PR #6598, opened 2022-08-08
Add doc for mixing left- and right-associative

jan0sch:
Add a small note at the paragraph that introduces right-associative path
matchers (`/:`) that they cannot be mixed with the left-associative
ones.

### DIFF
--- a/docs/docs/dsl.md
+++ b/docs/docs/dsl.md
@@ -375,7 +375,7 @@ HttpRoutes.of[IO] {
 }
 ```
 
-**Please note:** You cannot mix left- and right-associative matchers in a path! So something like `case GET -> "hello" / "world" /: rest => ???` will **not compile**.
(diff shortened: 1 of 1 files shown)


docs/docs/dsl.md
```diff
@@ -375,6 +375,16 @@ HttpRoutes.of[IO] {
 }
 ```
 
+**Please note:** You cannot mix left- and right-associative matchers in a path! So something like `case GET -> "hello" / "world" /: rest => ???` will **not compile**.
```
danicheg, line 378: Actually, I'm not that much UI designer, but if something doesn't compile, I'd argue we shouldn't highlight this much 😄 Also, perhaps, we can mark that this behavior is _expected_ rather than point out _failing compilation_. WDYT?

jan0sch: I removed the bold one on `not compile`
For the rest I'd say we stick with `cannot` because at least from my perspective I expected to be able to mix both or have it working by throwing around parentheses. ;-)

jan0sch:
Can this be merged?

hamnis (approved):

Issue #6596, opened 2022-08-07
Ember: Server: HTTP2 error when using a custom TLSContext and a Stream response


BalmungSan:
# Reproducing steps

Using the latest `0.23` **SNAPSHOT** version _(`0.23.14-39-b99efba-SNAPSHOT`)_ and the following code:

```scala
import cats.effect.{IO, IOApp, Resource}
import fs2.Stream
import fs2.io.net.tls.TLSContext
import org.http4s._
import org.http4s.dsl.Http4sDsl
import com.comcast.ip4s._
import org.http4s.server.{Server, Router}
import org.http4s.ember.server.EmberServerBuilder
import org.typelevel.log4cats.Logger
import org.typelevel.log4cats.slf4j.Slf4jLogger
import java.nio.file.Paths

object Main extends IOApp.Simple {
  override final val run: IO[Unit] =
    Slf4jLogger.fromName[IO]("api").flatMap { logger =>
      logger.info("Starting the server") >>
      HttpServer.server(
        "changeit",
        logger
      ).evalTap { server =>
        logger.info(s"Server has started at ${server.address}")
      }.useForever.void
    }
}

object HttpServer {
  private def buildTlsContext(keystorePassword: Array[Char]): IO[TLSContext[IO]] =
    TLSContext.Builder.forAsync[IO].fromKeyStoreFile(
      file = Paths.get("./custom.jks"),
      keyPassword = keystorePassword,
      storePassword = keystorePassword
    )

  private def buildServer(app: HttpApp[IO], tlsContext: TLSContext[IO], logger: Logger[IO]): Resource[IO, Server] =
    EmberServerBuilder
      .default[IO]
      .withHost(host"0.0.0.0")
      .withPort(port"8080")
      .withHttpApp(app)
      .withLogger(logger)
      .withHttp2
      .withTLS(tlsContext)
      .withErrorHandler {
        case error =>
          logger.error(error)("Unexpected error:").as(Response(status = Status.InternalServerError))
      }.build

  def server(keystorePassword: String, logger: Logger[IO]): Resource[IO, Server] =
    Resource.eval(buildTlsContext(keystorePassword.toCharArray)).flatMap { tlsContext =>
      buildServer(
        app =Router(
          "/" -> Routes.instance,
        ).orNotFound,
        tlsContext,
        logger
      )
    }
}

object Routes extends Http4sDsl[IO] {
  val instance = HttpRoutes.of[IO] {
    case GET -> Root / "test" / "io" =>
[...]

PR #6600, opened 2022-08-08
Ember H2 Connection Header Compliance

ChristopherDavenport:
So questions: 
1. How does Blaze deal with this?
2. Removing in the backend means it appears to be using chunked when logging, but is secretly not if h2 is used.
3. Overall it seems a bit like we have queued up nice helpers for a primarily http/1.1 world, but those helpers seem wrong with http2.

### DIFF
--- a/ember-core/shared/src/main/scala/org/http4s/ember/core/h2/PseudoHeaders.scala
+++ b/ember-core/shared/src/main/scala/org/http4s/ember/core/h2/PseudoHeaders.scala
@@ -20,6 +20,7 @@ import cats._
 import cats.data._
 import cats.syntax.all._
 import org.http4s._
+import org.typelevel.ci.CIString
 
 /** HTTP/2 pseudo headers */
 private[h2] object PseudoHeaders {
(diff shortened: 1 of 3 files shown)


ChristopherDavenport:
Fixes #6596

ChristopherDavenport:
Blaze seems to remove it, like this PR does, though I'm having difficulty locating the source.

ChristopherDavenport:
Alright found in the next section 

> This means that an intermediary transforming an HTTP/1.x message to HTTP/2 will need to remove any header fields nominated by the Connection header field, along with the Connection header field itself. Such intermediaries SHOULD also remove other connection-specific header fields, such as Keep-Alive, Proxy-Connection, Transfer-Encoding, and Upgrade, even if they are not nominated by the Connection header field.

https://httpwg.org/specs/rfc7540.html#rfc.section.8.1.2

So rather than what I do here I think I should just Filter the above values.

BalmungSan:
I can confirm that it fixes #6596

ember-core/shared/src/main/scala/org/http4s/ember/core/h2/PseudoHeaders.scala
```diff
@@ -162,4 +166,11 @@ private[h2] object PseudoHeaders {
     }.toOption
       .flatten
 
+  val connectionHeadersFields: Set[CIString] = Set(
+    org.http4s.headers.`Transfer-Encoding`.headerInstance.name,
+    org.http4s.headers.Connection.headerInstance.name,
+    org.http4s.headers.Upgrade.headerInstance.name,
+    org.http4s.headers.`Keep-Alive`.headerInstance.name,
+    CIString("Proxy-Connection"),
+  )
```
armanbilge, line 175: Maybe add a comment linking to the RFC here?
https://httpwg.org/specs/rfc7540.html#rfc.section.8.1.2.2

armanbilge (approved):
Thanks for adding a test suite!

ember-core/shared/src/main/scala/org/http4s/ember/core/h2/PseudoHeaders.scala
ChristopherDavenport: Added

PR #6606, opened 2022-08-09
Cross-compile `GZip` middlewares for JS

armanbilge:
Finally shaved this yak!

### DIFF
--- a/core/shared/src/main/scala/org/http4s/MediaType.scala
+++ b/core/shared/src/main/scala/org/http4s/MediaType.scala
@@ -317,7 +317,12 @@ object MediaType extends MimeDB {
         new MediaType(mainType, subType),
       )
     else // don't link the MimeDB!!!!
-      new MediaType(mainType, subType, binary = isBinary(mainType, subType))
+      new MediaType(
+        mainType,
+        subType,
+        compressible = isCompressible(mainType, subType),
+        binary = isBinary(mainType, subType),
+      )
   }
 
   implicit val http4sEqForMediaType: Eq[MediaType] =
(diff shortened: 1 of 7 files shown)


server/shared/src/main/scala/org/http4s/server/middleware/GZip.scala
```diff
@@ -20,18 +20,17 @@ package middleware
...
 import org.log4s.getLogger
 
-object GZip {
+object GZip extends GZipPlatform {
   private[this] val logger = getLogger
 
   // TODO: It could be possible to look for F.pure type bodies, and change the Content-Length header after
   // TODO      zipping and buffering all the input. Just a thought.
-  def apply[F[_]: Functor, G[_]: Sync](
+  def apply[F[_]: Functor, G[_]: Compression](
```
armanbilge, line 33: This is source-compatible for JVM, because anyone who already had `Sync` automatically has a `Compression` in scope too. Binary-compatibility is preserved via a deprecated overload in `GZipPlatform`.

core/shared/src/main/scala/org/http4s/MediaType.scala
```diff
@@ -410,4 +415,10 @@ object MediaType extends MimeDB {
         }
       case _ => false
     }
+
+  // lazy, incomplete predictor, used only on JS
+  private[this] def isCompressible(mainType: String, subType: String): Boolean = {
+    val _ = subType
+    mainType == "text"
+  }
```
armanbilge, line 423: AFAICT all `text` types are compressible. We can add more cases as needed, but I don't expect this to get hit much outside of our tests since this is used only for the server middleware.

ChristopherDavenport (approved):

PR #6607, opened 2022-08-10
Fix 1.0 version table compatibility with 2.12 & 2.13

JackTreble:
The version table currently suggests that 2.12 is supported and 2.13 is not for version 1.0
image


armanbilge (approved):
Woops, thank you!

PR #6610, opened 2022-08-12
Make StreamForking resource safe in the face of cancelation

TimWSpence:
Make `StreamForking` resource safe in the face of cancelation.

`StreamForking` isn't actually used at the moment but it looks as if the intention is to use it again.


TimWSpence:
The current tests fail because https://github.com/typelevel/fs2/pull/2962 isn't released yet. If anybody has a suggestion of an alternative testing approach then I'm all ears!

ChristopherDavenport (approved):

ChristopherDavenport:
If this is working, can we use this again to get safe shutdowns again?

TimWSpence:
> If this is working, can we use this again to get safe shutdowns again?

Sorry, I have no idea 😅 All I saw was [this comment](https://github.com/http4s/http4s/blob/series/0.23/ember-server/shared/src/main/scala/org/http4s/ember/server/internal/ServerHelpers.scala#L277) which said we need to fix something else. I haven't done the archaeology yet to find out what that is

armanbilge:
Ah, I believe that comment is related to this gnarly fs2 issue:
- https://github.com/typelevel/fs2/issues/2289
- https://github.com/typelevel/fs2/issues/2343

RafalSumislawski:
Given that https://github.com/http4s/http4s/pull/6656 (fs2 3.2.14) is now merged. Is this PR ready to be merged?

It looks to me like it would make sense to retarget/backport this to 0.23. What do you think?

TimWSpence:
> Given that #6656 (fs2 3.2.14) is now merged. Is this PR ready to be merged?
> 
> It looks to me like it would make sense to retarget/backport this to 0.23. What do you think?

I'm in favour 😂

PR #6611, opened 2022-08-13
In docs, added note on kind projector flag

cgoldammer:
In the docs, added clarification that some code requires the `kind-projector` scalac flag. Thanks arminbilge (in Discord) for the hint!

Tests:
- Ran `ack "Ykind-projector"` in the `docs` folder to ensure that this isn't already mentioned. The only match are the comments in this commit
- Ran `sbt quicklint`

### DIFF
--- a/docs/docs/auth.md
+++ b/docs/docs/auth.md
@@ -14,6 +14,13 @@ import org.http4s.dsl.io._
 import org.http4s.server._
 ```
 
+We will also require the `Ykind-projector` scalac flag, which you can add into your `build.sbt` using
+```scala mdoc:silent
+scalacOptions ++= Seq(
+  "-Ykind-projector"
+)
+```
+
 To add authentication to a service, we need some kind of `User` object which identifies the user
 who sent the request. We represent that with `AuthedRequest[F, User]`, which allows you to reference
 such object, and is the equivalent to `(User, Request[F])`. _http4s_ provides you with `AuthedRequest`,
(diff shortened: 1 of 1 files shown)


docs/docs/auth.md
```diff
@@ -14,6 +14,13 @@ import org.http4s.dsl.io._
 import org.http4s.server._
 ```
 
+We will also require the `Ykind-projector` scalac flag, which you can add into your `build.sbt` using
+```scala mdoc:silent
+scalacOptions ++= Seq(
+  "-Ykind-projector"
+)
+```
```
armanbilge, line 22: Thanks for this! We have to take into account a couple considerations.
1. This scalac flag only works for Scala 3. For Scala 2 a compiler plugin is required.
  https://github.com/typelevel/kind-projector
2. Actually, this is blanket advice that may be required for many examples, not just auth. So I wonder if we should put it in one place, and add a reminder for every page? Not sure.

cgoldammer: Yes that's a great point - I simply didn't know this. I can't find a section in the docs that quite fits this, maybe I should create a new section, after "Quick Start", and call it something like "General setup"? And that can combine this type of info? Maybe it's only this extension, but I'd guess there's more to come.

armanbilge: It's required for any page that makes use of the `*` syntax as a type parameter (a lot of them, if not most of them, AFAICT).

BTW, maybe we can add this information to the quick start page. That page mentions the partial unification compiler flag for example.
https://http4s.org/v0.23/docs/quickstart.html

That page also points to the g8 template for http4s, which actually sets this up:
https://github.com/http4s/http4s.g8/blob/e5b55668c29836201e4a554ee1f338a897dfff22/src/main/g8/build.sbt#L26

PR #6618, opened 2022-08-17
Ember-Core / H2Client /  fromSocket: split for comprehension.

diesalbla:
Redo of #6523, in a series of small-step commits. 

__________ 

Small refactor of the code in the `fromSocket` method of the `H2Client` class, for creating and starting up an `H2Connection`. Splits long code of the `for` comprehension (in Resource) into a set of auxiliary methods in `F`. 

Each of the auxiliary methods operates at the level of the "F" effect, and they are only lift to a `Resource` for the purposes of background running.

Initialising the `H2Connection` object is done in a separate method. Instead of accessing the components from the variables in the for loop, the other auxiliary methods access them
through the `H2Connection` fields themselves.


armanbilge (approved):
Thank you for breaking the original PR into small step-by-step commits. I appreciate your work on this code.

PR #6621, opened 2022-08-18
Etag take3

### LINKED ISSUE #2826: Add etag support for StaticFile.fromUrl where supported

matthughes:
Bringing PR #4879 up to date with series/0.23.  Fixes #2826. This preserves @qwbarch original changes.


matthughes:
cc @armanbilge

core/shared/src/main/scala/org/http4s/StaticFile.scala
```diff
@@ -137,6 +142,14 @@ object StaticFile {
     })
   }
 
+  private def calcETagURL[F[_]](implicit F: Sync[F]): URL => F[ETag] = url => {
+    val urlConn = url.openConnection
```
armanbilge, line 146: This method can throw an `IOException`. I wonder if it should be wrapped in `F.blocking` too.

https://docs.oracle.com/javase/8/docs/api/java/net/URL.html#openConnection--

core/shared/src/main/scala/org/http4s/StaticFile.scala
```diff
@@ -137,6 +142,14 @@ object StaticFile {
...
+    for {
+      lastModified <- F.blocking(urlConn.getLastModified.toHexString)
+      contentLength <- F.blocking(urlConn.getContentLengthLong.toHexString)
```
armanbilge, line 149: `getLastModified` returns `0` if unknown and `getContentLengthLong` returns `-1`. Should we be doing something special in these situations? Otherwise seems like the calculated etag could be misleading.

core/shared/src/main/scala/org/http4s/StaticFile.scala
```diff
@@ -156,7 +169,6 @@ object StaticFile {
           else ""
         )
 
-  @deprecated("Use fromPath", "0.23.5")
```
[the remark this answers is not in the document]
matthughes: yes

core/shared/src/main/scala/org/http4s/StaticFile.scala
danicheg: @armanbilge good call 👍🏻

core/shared/src/main/scala/org/http4s/StaticFile.scala
matthughes: Note that in existing code we don't use blocking around lastModified/getContentLengthLong or openConnection.  Should those be changed?

matthughes: Changed to return Option if both are empty.

core/shared/src/main/scala/org/http4s/StaticFile.scala
matthughes: Fixed.

core/shared/src/main/scala/org/http4s/StaticFile.scala
```diff
@@ -90,53 +90,70 @@ object StaticFile {
...
     })
   }
 
+  private def calcETagURL[F[_]](implicit F: Sync[F]): URL => F[Option[ETag]] = url =>
+    for {
+      urlConn <- F.blocking(url.openConnection)
+      lastModified <- F.blocking(urlConn.getLastModified)
+      contentLength <- F.blocking(urlConn.getContentLengthLong)
+    } yield
+      if (lastModified == 0 && contentLength == 0) None
```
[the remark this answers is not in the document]
matthughes: Fixed.  Also changed logic to be OR.  Unless both are defined we won't return an etag for this calculation.  Folks are free to override.

matthughes:
Know there has been a lot of back and forth here but I think everything discussed thus far has been addressed.

armanbilge:
Thanks! Can we add a test(s) for this?

matthughes:
Added three tests.  I'm not thrilled with them but it's not clear to me how to improve for testing the buit-in etag generator.  I can't mock out URLConnection.

I hard-coded the desired outcome.  I tried to use the Last-Modified/Content-Length headers but that doesn't work because Last-Modified header encodes epochSeconds and the etag calculator uses the more precise epochMilliseconds.  I could change the etag calculator to use seconds instead but that seems like going backwards.  Open to any suggestions if you think this isn't enough.

core/shared/src/main/scala/org/http4s/StaticFile.scala
```diff
@@ -86,57 +95,74 @@ object StaticFile {
...
+                  .as(Some(Response(NotModified)))
+            }
+        }
       }
     })
   }
 
+  private def calcETagURL[F[_]](implicit F: Sync[F]): URL => F[Option[ETag]] = url =>
+    for {
+      urlConn <- F.blocking(url.openConnection)
```
armanbilge, line 159: Is it weird that the connection is being opened twice? Both here, and in `fromURL`. Maybe it's not that big of a deal.

armanbilge (commented):
I think the tests are good. Thanks for adding them!

core/shared/src/main/scala/org/http4s/StaticFile.scala
matthughes: I don't know.  I thought about just passing contentLength/lastModified time directly to the etag function instead of URL but if they wanted to do more (hash the actual contents) that wouldn't be that helpful.  We could pass the actual `URLConnection` but not sure that improves much.

> Returns a [URLConnection](https://docs.oracle.com/javase/7/docs/api/java/net/URLConnection.html) instance that represents a connection to the remote object referred to by the URL.

> A new instance of [URLConnection](https://docs.oracle.com/javase/7/docs/api/java/net/URLConnection.html) is created every time when invoking the [URLStreamHandler.openConnection(URL)](https://docs.oracle.com/javase/7/docs/api/java/net/URLStreamHandler.html#openConnection(java.net.URL)) method of the protocol handler for this URL.

> It should be noted that a URLConnection instance does not establish the actual network connection on creation. This will happen only when calling [URLConnection.connect()](https://docs.oracle.com/javase/7/docs/api/java/net/URLConnection.html#connect()).

I dug a little unto the gnarly, mutable, impure mess of JarURLConnection and can't find anything too bad.  We could change etagCalculator to take some sort of `UrlContext` trait:

```scala
trait UrlContext {
  def contentLength: Option[Long]
  def lastModified: Option[Long]
  def urlConnection: UrlConnection
  def url: Url
}
```

This would let the default implentation reuse connection and contentLength/lastModified without re-opening the connection.  By making it a trait it allows us to add members in the future without breaking API.  Might be a bit over-engineering though.

armanbilge: The fact that `openConnection` may throw an `IOException` suggests to me that it may be doing something non-trivial. So maybe the etag calculator should be in terms of `URLConnection` so it can be opened once and shared.

Would be great to get another opinion on this.

matthughes: @armanbilge So one option that could move this PR forward is we just remove the ability to customize the etag.  Do we really need to support that?  If we remove that, the internal, canonical implementation can just use the opened connection, parameters listed above.  Seems to me the only thing keeping this moving forward is cause we don't want to commit to an API change that we would be locked into.

armanbilge: @matthughes yes, that seems like a good solution to me :) we can always look into adding this API in the future if someone wants to customize it.

diogocanut: Hey! What's the status of this PR? Let me know if I can help somehow : )

armanbilge: @diogocanut if you want to open a new PR based on this branch and making the changes discussed here. I think we decided to hide this API in the end?

Btw thanks for all your work on Typelevel projects lately, you've been awesome 🚀

diogocanut: Moved it to #7155 . I implement this method in terms of `URLConnection` so we don't have to use `openConnection` twice, it was the final decision here right?

PR #6622, opened 2022-08-18
add SourceMap header

cobr123:
It's yet another header from the https://github.com/http4s/http4s/issues/5010 list.

### DIFF
--- a/tests/shared/src/test/scala/org/http4s/headers/SourceMapSuite.scala
+++ b/tests/shared/src/test/scala/org/http4s/headers/SourceMapSuite.scala
@@ -1,0 +1,61 @@
+/*
+ * Copyright 2013 http4s.org
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


danicheg (commented):
Thanks! This looks good to me, one suggestion:

tests/shared/src/test/scala/org/http4s/headers/SourceMapSuite.scala
```diff
@@ -0,0 +1,74 @@
...
+package org.http4s
+package headers
+
+import cats.effect.IO
+import org.http4s.syntax.header._
+
+class SourceMapSuite extends HeaderLaws {
+
+  def getUri(uri: String): Uri =
+    Uri.fromString(uri).fold(_ => sys.error(s"Failure on uri: $uri"), identity)
```
danicheg, line 26: `sys.error` looks too strict, wouldn't it be fine just to throw an exception here?

cobr123: done

danicheg: Oh snap! 😞 Didn't wake up properly and didn't suggest using URI literal syntax. Fixed on my own, hope you don't mind!

danicheg (approved):

danicheg:
I think this PR is good to get into the forthcoming 0.23.15 release. Otherwise, it could get stuck for a month waiting for a new release.

PR #6623, opened 2022-08-18
RFC - Compositional builder for server response logger

bplommer:
Found myself down this rabbit hole after looking into adding more options around logging behaviour and reeling at the combinatorial explosion of logging options that would result.

This is an attempt to get that under control - all forms of feedback and bikeshedding welcome. If people agree with the approach I would apply it to (at least) the other server loggers before considering it ready to merge.

I strongly recommend hiding whitespace when looking at the diff.

### DIFF
--- a/server/shared/src/main/scala/org/http4s/server/middleware/ResponseLogger.scala
+++ b/server/shared/src/main/scala/org/http4s/server/middleware/ResponseLogger.scala
@@ -27,10 +27,11 @@ import cats.effect.kernel.Outcome
 import cats.effect.kernel.Sync
 import cats.effect.syntax.all._
 import cats.syntax.all._
-import cats.~>
+import cats.{Functor, ~>}
 import fs2.Chunk
 import fs2.Pipe
 import fs2.Stream
+import org.http4s.server.middleware.ResponseLogger.Lift
 import org.log4s.getLogger
 import org.typelevel.ci.CIString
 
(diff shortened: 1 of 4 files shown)


server/shared/src/main/scala/org/http4s/server/middleware/ResponseLogger.scala
```diff
@@ -31,121 +32,165 @@ import cats.~>
...
       logAction: Option[String => F[Unit]] = None,
-  )(
-      http: Kleisli[G, A, Response[F]]
-  )(implicit G: MonadCancelThrow[G], F: Async[F]): Kleisli[G, A, Response[F]] =
-    impl[G, F, A](logHeaders, Left(logBody), fk, redactHeadersWhen, logAction)(http)
+  )(implicit F: Async[F])
+      extends ResponseLogger[F] {
+    override def apply[G[_], A](fk: F ~> G)(
+        http: Kleisli[G, A, Response[F]]
+    )(implicit G: MonadCancelThrow[G]): Kleisli[G, A, Response[F]] = {
```
bplommer, line 69: The implementation is completely unchanged. In retrospect maybe I should have left it as a method on the object to reduce the diff.

server/shared/src/main/scala/org/http4s/server/middleware/ResponseLogger.scala
```diff
@@ -31,121 +32,165 @@ import cats.~>
 import fs2.Chunk
 import fs2.Pipe
 import fs2.Stream
+import org.http4s.server.middleware.ResponseLogger.Lift
 import org.log4s.getLogger
 import org.typelevel.ci.CIString
 
+sealed abstract class ResponseLogger[F[_]] {
```
danicheg, line 39: Probably we can name it builder or factory (I prefer builder if someone would wonder 😆)? Also, shouldn't we hide it from public API?

danicheg (commented):
```scala
Foo[F]
  .withBar(bar)
  .withQux(qux1, qux2)
  .withQuux(quux)
```
is exactly nicer to me than
```scala
Foo[F](a = bar, b = qux1, c = qux2, d = quux)
```
so just a few naming things from me:

server/shared/src/main/scala/org/http4s/server/middleware/ResponseLogger.scala
```diff
@@ -31,121 +32,165 @@ import cats.~>
...
-      Right(logBody),
-      OptionT.liftK[F],
-      redactHeadersWhen,
-      logAction,
-    )(httpRoutes)
+    apply(logHeaders, logBody)
+      .withRedactHeadersWhen(redactHeadersWhen)
+      .withLogActionOpt(logAction)(httpRoutes)
+
+  sealed abstract class Lift[F[_], G[_]] {
```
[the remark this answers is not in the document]
bplommer: > Does this more likely belong to some common util module rather than `server.middleware?`

Yes possibly, but it should probably have proper laws if we expose it as something more general-purporse.

> Also, I just wonder, would it be more intuitive if we call this `HasFunctionK`? And by and large, it's a good question if we can add it to `cats` straightforwardly...

If `HasFunctionK` is just encoding that there's a canonical natural transformation from `F` to `G` does that need its own type or could it just be expressed by making certain `FunctionK` instances implicit?

server/shared/src/main/scala/org/http4s/server/middleware/ResponseLogger.scala
bplommer: > Probably we can name it builder or factory (I prefer builder if someone would wonder 😆)? 

I think `ResponseLoggerBuilder` implies that the build action generates a `ResponseLogger`, but it doesn't - it transforms an `Http` instance to add logging. Given this class already has an `apply` method that adds logging to a client, I think it's fair to describe it as a logging middleware rather than a builder for one.

> Also, shouldn't we hide it from public API?

I don't think so - the idea is that users will call the `.withXXX` methods directly, so that we don't need to keep adding more factory methods to handle different configuration possibilities

server/shared/src/main/scala/org/http4s/server/middleware/ResponseLogger.scala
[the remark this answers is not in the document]
bplommer: I don't think so, `FunctorK` expresses the ability to transform an algebra with a `FunctionK` instance but it doesn't provide the instance.

bplommer:
Superseded by #6629

PR #6628, opened 2022-08-22
Simplify type signature for internal logger

bplommer:
Simplifies type signatures for `org.http4s.internal.Logger`. For whatever reason it isn't package-private even though it's in the `internal` package, but it's binary-compatible though source-breaking so I think it's ok?

### DIFF
--- a/client/shared/src/main/scala/org/http4s/client/middleware/Logger.scala
+++ b/client/shared/src/main/scala/org/http4s/client/middleware/Logger.scala
@@ -55,7 +55,7 @@ object Logger {
       redactHeadersWhen: CIString => Boolean = Headers.SensitiveHeaders.contains,
   )(log: String => F[Unit])(implicit F: Async[F]): F[Unit] =
     org.http4s.internal.Logger
-      .logMessage[F, A](message)(logHeaders, logBody, redactHeadersWhen)(log)
+      .logMessage(message)(logHeaders, logBody, redactHeadersWhen)(log)
 
   def colored[F[_]: Async](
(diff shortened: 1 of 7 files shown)


armanbilge (approved):
We do have to be careful with these bincompat-wise as they may be used in some of the satellite repos, but these changes seem fine to me.

While we're here should we make it package-private?

bplommer:
> While we're here should we make it package-private?

In `main` I think so yes, though I do want to expose this stuff in a thoughtful way (I'm working on a PR to do that but it got a bit rabbit-holey). IMO that's too big a source breakage to do on 0.23 though.

armanbilge:
Sounds good! We can keep the status quo on 0.23.

FYI I've also been poking at the logging stuff, see https://github.com/http4s/http4s/pull/6614. In case your redesign could benefit from a more direct log4cats integration :)

bplommer:
> FYI I've also been poking at the logging stuff, see #6614. In case your redesign could benefit from a more direct log4cats integration :)

Oh cool I'll take a look!

bplommer:
Sorry, I may have been too trigger-happy again with merging this - I can revert if anyone has any issues with it?

danicheg:
Don't want to be a grumpy old geezer, but since this hurts (though a little) users, why didn't we target 1.0?

bplommer:
My thinking is that it's a pretty obscure API (even marked `internal`) so it won't have a widespread effect. And if people are maintaining code that touches it, they'll probably be happy to have this simplification.

PR #6629, opened 2022-08-22
Add builder for server logger middleware

bplommer:
An attempt to impose some order on the combinatorial explosion of logging options. I've re-targeted these changes at 0.23.

### DIFF
--- a/server/shared/src/main/scala/org/http4s/server/middleware/Logger.scala
+++ b/server/shared/src/main/scala/org/http4s/server/middleware/Logger.scala
@@ -29,7 +29,7 @@ import fs2.Stream
 import org.log4s.getLogger
 import org.typelevel.ci.CIString
 
-sealed abstract class Logger[F[_]] extends internal.Logger[F, Logger[F]] {
+sealed abstract class LoggerBuilder[F[_]] extends internal.Logger[F, LoggerBuilder[F]] {
   def apply[G[_]](fk: F ~> G)(
       http: Http[G, F]
   )(implicit G: MonadCancelThrow[G]): Http[G, F]
@@ -49,21 +49,21 @@ object Logger {
     Sync[F].delay(logger.info(s))
 
   private[middleware] final case class Impl[F[_]](
-      responseLogger: ResponseLogger[F],
-      requestLogger: RequestLogger[F],
-  ) extends Logger[F] {
+      responseLogger: ResponseLoggerBuilder[F],
+      requestLogger: RequestLoggerBuilder[F],
+  ) extends LoggerBuilder[F] {
 
     override def apply[G[_]](fk: F ~> G)(
         http: Http[G, F]
     )(implicit G: MonadCancelThrow[G]): Http[G, F] = responseLogger(fk)(requestLogger(fk)(http))
 
-    override def withRedactHeadersWhen(f: CIString => Boolean): Logger[F] =
+    override def withRedactHeadersWhen(f: CIString => Boolean): LoggerBuilder[F] =
       copy(
         responseLogger = responseLogger.withRedactHeadersWhen(f),
         requestLogger = requestLogger.withRedactHeadersWhen(f),
       )
 
-    override def withLogAction(f: String => F[Unit]): Logger[F] = copy(
+    override def withLogAction(f: String => F[Unit]): LoggerBuilder[F] = copy(
       responseLogger = responseLogger.withLogAction(f),
       requestLogger = requestLogger.withLogAction(f),
     )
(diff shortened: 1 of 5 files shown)


server/shared/src/main/scala/org/http4s/server/middleware/Logger.scala
```diff
@@ -18,82 +18,141 @@ package org.http4s
...
+    override def withLogAction(f: String => F[Unit]): LoggerBuilder[F] = copy(
+      responseLogger = responseLogger.withLogAction(f),
+      requestLogger = requestLogger.withLogAction(f),
+    )
+  }
+
+  def builder[F[_]: Async](
+      logHeaders: Boolean,
+      logBody: Boolean,
+  ): LoggerBuilder[F] =
```
armanbilge, line 75: Since we're re-doing these builders, this is a chance to do it right and ask for a `LoggerFactory` instead of `Sync` and friends :) feel free to base on https://github.com/http4s/http4s/pull/6614, I never force-push.

bplommer: Good thinking - does that require extra ceremony at the call site to provide a `LoggerFactory` though? If so I'm reluctant to force that on users while deprecating the existing constructors - could we provide a default `LoggerFactory` instance?

armanbilge: Good point. It will indeed require some change on the user-end. If we take the `LoggerFactory` implicitly, most of them will probably want to do:

```scala
import org.typelevel.log4cats.slf4j._
```
https://github.com/typelevel/log4cats/blob/4f02ef39fbee3568c9eaa52fee42904247988c62/slf4j/src/main/scala/org/typelevel/log4cats/slf4j/package.scala#L23

Ideally there should be an equivalent import for the no-op logger, but I don't see it. I can PR that now.

Thoughts? log4cats-slf4j will already be on the classpath for JVM users so adding the import seems trivial enough.

The only `LoggerFactory` we could provide by default without asking for `Sync` is the no-op one, but I don't think that's what we want to do.

bplommer: So one question here is whether we want to keep allowing users to provide a `String => F[Unit]` for the log action, in which case a `LoggerFactory` wouldn't be used. I'd strongly say yes, in which case it seems like a bad UX to require a `LoggerFactory` to be provided.

This is really a more acute version of a problem we already have, which is that we always ask for `Async` rather than `Concurrent` (because the default logger requires it) even if the user provides a custom log action. I've been experimenting with how to change that - the latest changes on this PR move the `Async` constraint from the `builder` method (before customisation takes place) to the `apply` method (after), and I'm hoping it's possible (without using too much heavy type-level machinery) to weaken the requirement to `Concurrent` when a custom log action is required. If so we should be able to do something analogous with `LoggerFactory`.

As for the specific behaviour in choosing loggers, here's what I'd suggest if it's not too messy to implement:
- By default, use a prioritised implicit search for either `LoggerFactory` or `Sync` - so if a `LoggerFactory` isn't in scope we silently fall back to slf4j (on the JVM at least)
- Add a builder option to specify a `LoggerFactory`
- Don't require a `LoggerFactory` if a `String => F[Unit]` log action is specified

What do you think?

(As an aside, I'd like to also allow a few other signatures for custom log actions - `(SelfAwareStructuredLogger[F], String) => F[Unit]` but also `(SelfAwareStructuredLogger, Message[Any], String) => F[Unit]` so that logging levels and context can be customised based on things like methods/status codes/headers, but I think that discussion can be postponed)

armanbilge: Yeesh, that got complicated real fast 😅 

> So one question here is whether we want to keep allowing users to provide a `String => F[Unit]` for the log action, in which case a `LoggerFactory` wouldn't be used. I'd strongly say yes, in which case it seems like a bad UX to require a `LoggerFactory` to be provided.

Ugh. Agree about the bad UX. Putting aside 0.23 and compatibility, I'd actually lean toward no. Seems kind of crazy to complicate our API with this "branch" between a `LoggerFactory` vs a `String => F[Unit]`, esp. if this is a recurring pattern in http4s. Instead what if we had something that could lift a `String => F[Unit]` into `LoggerFactory` (or is this too ambitious 😂 ).

Overall, everything you say seems reasonable, but I'm worried about the complexity of the the implementation. Rather than complicating our builders to be "flexible" can we improve the abstractions they ask for?

bplommer: > Instead what if we had something that could lift a `String => F[Unit]` into `LoggerFactory` (or is this too ambitious 😂 ).

Interesting thought - it would be nice in any case for it to be easier to instantiate simple implementations of log4cats types without having to implement a zillion different methods. Did you have something like [this](https://github.com/typelevel/log4cats/compare/main...bplommer:log4cats:lift-function-to-logger) in mind?

armanbilge: Gosh that's a lot of boilerplate for a "simple" logger 😂 but yes pretty much that! Please ping me when you open a PR :)

PR #6630, opened 2022-08-22
Release v0.23.15

armanbilge:

PR #6633, opened 2022-08-22
Revert "Simplify type signature for internal logger"

ChristopherDavenport:
This reverts commit a8bee5e25ed18be48ac2ef070879f6dca71bcb7d.

Unblocks Release For More Debate Per #6630


armanbilge:
FWIW I was also fine with this change, I don't think we need to revert it.

ChristopherDavenport:
Cool, then lets roll forward with the release. I'm cool with it as well, but didn't want things to be blocked.

PR #6638, opened 2022-08-25
Ember-Core H2Server: get streamCreationLock from H2Connection

diesalbla:
Since `streamCreationLock.permit` is itself a functional value, it should not be a problem to share it from the connection.


armanbilge:
@diesalbla this has conflicts btw :)

PR #6645, opened 2022-08-27
Expunge all unsafe logging

armanbilge:
This starts as a forward merge of https://github.com/http4s/http4s/pull/6614.

Then 8ee829563fc1bd60e5826171bc0d8453cceb46aa goes crazy and tries to fixup all the unsafe logging by adding `LoggerFactory` constraints as needed. This ... was a bit messy.

Inline comments incoming.


core/shared/src/main/scala/org/http4s/Message.scala
```diff
@@ -86,19 +85,11 @@ sealed trait Message[+F[_]] extends Media[F] { self =>
...
-            },
-            Some(_),
-          )
-      case None => None
-    }
+    val cl: Option[`Content-Length`] = entity.length.flatMap(
+      `Content-Length`
+        .fromLong(_)
+        .toOption
+    )
```
armanbilge, line 92: We can't log in this method unless we make `.withEntity` return a result in `F[_]` ... which seems like terrible ergonomics. So I just removed it.

server/shared/src/main/scala/org/http4s/server/middleware/PushSupport.scala
```diff
@@ -1,123 +0,0 @@
...
-import cats.Monad
-import cats.data.Kleisli
-import cats.effect.SyncIO
-import cats.syntax.all._
-import org.log4s.getLogger
-import org.typelevel.vault._
-
-@deprecated("Obsolete. Not implemented by any backends.", "0.23.12")
-object PushSupport {
-  private[this] val logger = getLogger
```
armanbilge, line 30: Rather than deal with this, since it was deprecated I just tossed it.

server/shared/src/main/scala/org/http4s/server/middleware/CORS.scala
```diff
@@ -22,125 +22,12 @@ import cats.Applicative
...
-
-/** CORS middleware config options.
-  * You can give an instance of this class to the CORS middleware,
-  * to specify its behavior
-  */
-
-@deprecated(
-  """Deficient. See https://github.com/http4s/http4s/security/advisories/GHSA-52cf-226f-rhr6.""",
-  "0.21.27",
-)
```
armanbilge, line 43: I removed all the deprecated CORS stuff so I wouldn't have to rework its logging.

server/shared/src/main/scala/org/http4s/server/middleware/CORS.scala
```diff
@@ -320,7 +84,9 @@ sealed class CORSPolicy(
 ) {
   import CORSPolicy._
 
-  def apply[F[_]: Applicative, G[_]](http: Http[F, G]): Http[F, G] = {
+  def apply[F[_]: Applicative, G[_]: Applicative: LoggerFactory](
+      http: Http[F, G]
+  ): G[Http[F, G]] = {
```
armanbilge, line 89: This is where we get into trouble. To log in this method (see below) we have to return the result in an effect now. Maybe this is ok, but it caused extensive refactoring of the CORS test suite.

server/shared/src/main/scala/org/http4s/server/staticcontent/ResourceService.scala
```diff
@@ -82,13 +82,13 @@ class ResourceServiceBuilder[F[_]] private (
 
   def withBufferSize(bufferSize: Int): ResourceServiceBuilder[F] = copy(bufferSize = bufferSize)
 
-  def toRoutes(implicit F: Async[F]): HttpRoutes[F] = {
+  def toRoutes(implicit F: Async[F]): F[HttpRoutes[F]] = {
```
armanbilge, line 85: Another signature change here, that caused the test suite to need major refactoring. We could, also, just not log ...

armanbilge:
Ouch 😂

armanbilge:
Yes, exactly. No more slf4j dependency.

TonioGela:
Any updates? That's a definitively useful feature to merge.

armanbilge:
I can't merge my own PR, but if someone new champions this effort and fix the conflicts I can approve and merge 😜

Issue #6650, opened 2022-09-04
Deprecate methods that take `F[Request[F]]` as a parameter


danicheg:
As a further step in #3345 @satorg [has suggested](https://github.com/http4s/http4s/issues/3345#issuecomment-817362069) deprecating methods that take `F[Request[F]]` as a parameter. I propose to discuss this suggestion here.

satorg:
I just think that methods taking `F[Request[F]]` as a parameter are although harmless but quite useless.
Since `Request[F]` is supposed to be pure and safe then there are no real need to keep those method as first-class citizens of `Client`. Having same methods that take just `Request[F]` it is unlikely that they can be somehow optimized for `F[Request[F]]`. They can be implemented in a form of extension methods though, if necessary.

danicheg:
This makes sense to me. So I've opened #6651.

PR #6651, opened 2022-09-05
Deprecate some methods in `Client`

danicheg:
Addresses #6650.

### DIFF
--- a/client/shared/src/main/scala/org/http4s/client/Client.scala
+++ b/client/shared/src/main/scala/org/http4s/client/Client.scala
@@ -101,6 +101,7 @@ trait Client[F[_]] {
       d: EntityDecoder[F, A]
   ): F[A]
 
+  @deprecated("Use req.flatMap(expect(_))", "0.23.16")
   def expect[A](req: F[Request[F]])(implicit d: EntityDecoder[F, A]): F[A]
 
   def expectOr[A](uri: Uri)(onError: Response[F] => F[Throwable])(implicit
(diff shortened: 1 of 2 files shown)


client/shared/src/main/scala/org/http4s/client/Client.scala
```diff
@@ -101,6 +101,7 @@ trait Client[F[_]] {
       d: EntityDecoder[F, A]
   ): F[A]
 
+  @deprecated("Use req.flatMap(expect(_))", "0.23.16")
```
danicheg, line 104: If we decide to accept this, in the `1.x` series perhaps we should remove these deprecations then. WDYT?

armanbilge: Yes. In theory it is Nice:tm: to do at least a single deprecation cycle in 1.x as well. But it also means more work for us.

satorg: I wonder how many projects are affected by these deprecations. In a case if there are quite a lot of them then perhaps we could consider preserving such methods in a form of `Client` extension syntax.

armanbilge: I disagree. What is the point of putting it into extension syntax? If we intend to keep it, we should just keep it on `Client`.

danicheg: Agree. I thought that was our intent to clean up API.

danicheg:
So, this got two likes, I think it's legit to move onward. Hope to don't receive a ton of dislikes from users 🤞🏻

client/shared/src/main/scala/org/http4s/client/Client.scala
satorg: I would not say I think we should. Personally I would prefer just to get rid of those methods. But wondering if there could be a demand for them from dependent projects.

Issue #6326, opened 2022-04-25
Don't create redundant `Ref` in the `Client#fromHttpApp`


danicheg:
This is a follow-up from https://github.com/http4s/http4s/pull/6325.
Ross made a [proposal](https://github.com/http4s/http4s/pull/6325#pullrequestreview-951279467) that we can don't create `Ref` in a case when the request/response body isn't streaming. 
One should keep in mind, that a single `Ref` instance is reusing.

omaximov:
I'd like to take a crack at this, if that's okay!

armanbilge:
That's wonderful, thank you!

omaximov:
What's the ideal timeline on getting this implemented? I've been rather swamped at work lately, and I'm not *super* familiar with cats-effect or http4s. But I'd like to be! So it may be two weeks before I get a chance to really dig in and familiarize myself with the tech to come up with a solution. Maybe it'll be a quick fix but I don't want to half-ass it.

danicheg:
Currently, no one is working on this issue, so it's up to you exactly 👍🏻

geoffjohn11:
@omaximov I took a look at this issue and thought I would try it.  Don't want to step on your toes, though.  If you'd like you can use my PR (if it's the right approach).

omaximov:
Please do! I admittedly let this issue get away from me

geoffjohn11:
Addressed in https://github.com/http4s/http4s/pull/6654

PR #6653, opened 2022-09-06
removed un-needed Ref from Client

geoffjohn11:
PR for https://github.com/http4s/http4s/issues/6326
I'm not entirely sure this is the right direction, but though I would take a try at it.

### DIFF
--- a/.scala-steward.conf
+++ b/.scala-steward.conf
@@ -1,3 +1,4 @@
 updates.pin = [
-  { groupId = "org.slf4j", artifactId = "slf4j-api", version = "1." }
+  { groupId = "org.slf4j", artifactId = "slf4j-api", version = "1." },
+  { groupId = "ch.qos.logback", artifactId = "logback-classic", version = "1.2." }
 ]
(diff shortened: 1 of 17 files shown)


armanbilge:
Thanks for the PR! I think the original proposal was in https://github.com/http4s/http4s/pull/6325#pullrequestreview-951279467.
> I don't think this needs the Ref unless the body is a Stream, but this is progress.

In `series/0.23` the body is always a `Stream` so I am not sure if we can make this change here. However in `main` we have redefined the `Entity` model so that the body is not always a `Stream`. In those cases we do not need the `Ref`.

geoffjohn11:
> Thanks for the PR! I think the original proposal was in [#6325 (review)](https://github.com/http4s/http4s/pull/6325#pullrequestreview-951279467).
> 
> > I don't think this needs the Ref unless the body is a Stream, but this is progress.
> 
> In `series/0.23` the body is always a `Stream` so I am not sure if we can make this change here. However in `main` we have redefined the `Entity` model so that the body is not always a `Stream`. In those cases we do not need the `Ref`.

Thanks!  I'll take a look at the main branch and see if I can come up with the right idea.

geoffjohn11:
Committed to wrong branch

PR #6654, opened 2022-09-06
removed the Ref boolean

geoffjohn11:
PR to address Ref comment in https://github.com/http4s/http4s/pull/6325#pullrequestreview-951279467

### DIFF
--- a/client/shared/src/main/scala/org/http4s/client/Client.scala
+++ b/client/shared/src/main/scala/org/http4s/client/Client.scala
@@ -237,7 +237,7 @@ object Client {
     *
     * @param app the [[HttpApp]] to respond to requests to this client
     */
-  def fromHttpApp[F[_]: MonadCancelThrow: Concurrent](
+  def fromHttpApp[F[_]](
       app: HttpApp[F]
   )(implicit F: Async[F]): Client[F] = {
     def until[A](disposed: Ref[F, Boolean])(source: Stream[F, A]): Stream[F, A] = {
--- a/client/shared/src/main/scala/org/http4s/client/Client.scala
+++ b/client/shared/src/main/scala/org/http4s/client/Client.scala
@@ -255,8 +255,7 @@ object Client {
     }
 
     def run(
-        message: Message[F],
-        disposedOpt: Option[Ref[F, Boolean]] = None,
+        message: Message[F]
     ): Resource[F, Response[F]] = message.entity match {
       case Entity.Empty | Entity.Strict(_) =>
         message match {
@@ -273,15 +272,11 @@ object Client {
               Resource
                 .eval(app(reqAugmented))
                 .onFinalize(disposed.set(true))
-                .flatMap(run(_, Option(disposed)))
+                .flatMap(run(_))
             case resp @ Response(_, _, _, _, _) =>
-              disposedOpt.fold(
-                Resource
-                  .eval(F.pure(resp.pipeBodyThrough(until(disposed))))
-                  .onFinalize(disposed.set(true))
-              ) { reqDisposed =>
-                Resource.eval(F.pure(resp.pipeBodyThrough(until(reqDisposed))))
-              }
+              Resource
+                .eval(F.pure(resp.pipeBodyThrough(until(disposed))))
+                .onFinalize(disposed.set(true))
           }
         }
         Resource.suspend(refResource)
--- a/client/shared/src/main/scala/org/http4s/client/Client.scala
+++ b/client/shared/src/main/scala/org/http4s/client/Client.scala
@@ -237,8 +237,10 @@ object Client {
     *
     * @param app the [[HttpApp]] to respond to requests to this client
     */
-  def fromHttpApp[F[_]](app: HttpApp[F])(implicit F: Async[F]): Client[F] = {
-    def untilDisposed[A](disposed: Ref[F, Boolean])(source: Stream[F, A]): Stream[F, A] = {
+  def fromHttpApp[F[_]: MonadCancelThrow: Concurrent](
+      app: HttpApp[F]
+  )(implicit F: Async[F]): Client[F] = {
+    def until[A](disposed: Ref[F, Boolean])(source: Stream[F, A]): Stream[F, A] = {
       def go(stream: Stream[F, A]): Pull[F, A, Unit] =
         stream.pull.uncons.flatMap {
           case Some((chunk, stream)) =>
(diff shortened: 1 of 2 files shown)


armanbilge:
Thanks for retargeting! At least according to https://github.com/http4s/http4s/pull/6325#pullrequestreview-951279467 we do still need the `Ref` if the body is a `Stream`.

client/shared/src/main/scala/org/http4s/client/Client.scala
```diff
@@ -248,33 +248,27 @@ object Client {
...
-          addHostHeaderIfUriIsAbsolute(entity)
+    def runDefault(req: Request[F]) =
+      Resource.suspend(
+        Ref[F].of(false).map { disposed =>
+          val req0 =
+            addHostHeaderIfUriIsAbsolute(req.pipeBodyThrough(untilDisposed(disposed)))
+          Resource
+            .eval(app(req0))
+            .onFinalize(disposed.set(true))
+            .map(_.pipeBodyThrough(untilDisposed(disposed)))
```
geoffjohn11, line 259: @danicheg is it OK that I removed the entity check on the Response here?  I'm not sure if a Request has an Entity type of Default if the Response must also always be Default.

client/shared/src/main/scala/org/http4s/client/Client.scala
```diff
@@ -234,7 +234,7 @@ object Client {
     * @param app the [[HttpApp]] to respond to requests to this client
     */
   def fromHttpApp[F[_]](app: HttpApp[F])(implicit F: Async[F]): Client[F] = {
```
kubukoz, line 236: Any chance we can change F to `F[_]: MonadCancelThrow: Ref.Make` here?

armanbilge: Not crazy about `Ref.Make` (what's the motivation?) but maybe this could become `Concurrent`.

kubukoz: Motivation in general: `Ref.Make` works for `F`s that are Sync but not Concurrent, but doesn't require the full powers of `Sync`. In this case `Concurrent` probably works better

client/shared/src/main/scala/org/http4s/client/Client.scala
geoffjohn11: This code has problems.  If the request body isn't a Stream the response could be a Stream.

client/shared/src/main/scala/org/http4s/client/Client.scala
```diff
@@ -234,47 +231,42 @@ object Client {
...
-              req.pipeBodyThrough(until(disposed))
-          }
-
-          addHostHeaderIfUriIsAbsolute(entity)
+    def run(req: Request[F]): F[Resource[F, Response[F]]] = {
+      val req0 = {
+        val entity = req.entity match {
+          case Entity.Empty | Entity.Strict(_) =>
+            req
+          case Entity.Default(_, _) =>
```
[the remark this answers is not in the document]
geoffjohn11: this should be fixed

client/shared/src/main/scala/org/http4s/client/Client.scala
geoffjohn11: added Concurrent and MonadCancelThrow

client/shared/src/main/scala/org/http4s/client/Client.scala
```diff
@@ -237,7 +237,9 @@ object Client {
     *
     * @param app the [[HttpApp]] to respond to requests to this client
     */
-  def fromHttpApp[F[_]](app: HttpApp[F])(implicit F: Async[F]): Client[F] = {
+  def fromHttpApp[F[_]: Concurrent](
+      app: HttpApp[F]
+  )(implicit F: MonadCancelThrow[F]): Client[F] = {
```
armanbilge, line 242: `Concurrent` extends `MonadCancelThrow`, so we can just ask for `Concurrent`.

```suggestion
  def fromHttpApp[F[_]](
      app: HttpApp[F]
  )(implicit F: Concurrent[F]): Client[F] = {
```

geoffjohn11:
@armanbilge @danicheg @kubukoz I think the ClientSuite test is missing the case for Stream Request -> Strict Response, and Stream Request -> Stream Response.  I'd like to add these tests, and then will this PR be good to go?

armanbilge:
@geoffjohn11 thanks for all your work on this! Yes please, adding tests is always great, good catch.

geoffjohn11:
@armanbilge Thanks for your feedback and reviews!  Hopefully this is good to go.

client/shared/src/main/scala/org/http4s/client/Client.scala
```diff
@@ -252,33 +254,46 @@ object Client {
...
+            refOp.fold {
+              Resource.suspend(Ref[F].of(false).map { disposed =>
+                Resource
+                  .eval(F.pure(resp.pipeBodyThrough(until(disposed))))
+                  .onFinalize(disposed.set(true))
+              })
+            } { disp =>
+              Resource
+                .eval(F.pure(resp.pipeBodyThrough(until(disp))))
+                .onFinalize(disp.set(true))
```
armanbilge, line 292: I'm a little confused by these lines.
1. `Resource.eval(F.pure(...))` is equivalent to `Resource.pure(...)`
2. Since the code is the same, and there is just the issue of whether the `Ref` is provided or not, I think you can do `refOp.fold(Ref[F].of(false))(F.pure).flatMap { disp => ... }` and then share the code.

geoffjohn11: @armanbilge I think I fixed the duplication, let me know if you'd like additional changes.  Thanks for the review!

client/shared/src/test/scala/org/http4s/client/ClientSuite.scala
```diff
@@ -53,6 +55,30 @@ class ClientSpec extends Http4sSuite with Http4sDsl[IO] {
...
+    Request[IO](POST)
+      .withEntity(stream)
+      .pure[IO]
+      .flatMap { req =>
+        // This is bad. Don't do this.
+        client.run(req).use(IO.pure).flatMap(_.as[String])
+      }
+      .attempt
+      .map(_.left.toOption.get.getMessage)
+      .assertEquals("response was disposed")
```
armanbilge, line 72:
```suggestion
      .interceptMessage[IOException]("response was disposed")
```

armanbilge (commented):
Sorry for the very slow review cycle :( I appreciate your continued work on this.

armanbilge (approved):
I think this is good. Thanks for all your work on it.

General comment (not related to your changes): personally I found `run` difficult to reason about (and thus this PR difficult to review), since it handles both requests/responses and calls itself recursively. I wonder if anyone else feels the same.

diesalbla (approved):

PR #6657, opened 2022-09-08
Update to munit 1.0.0-M6

armanbilge:
At this point it is quite stable and brings many improvements we could benefit from, such as async resource fixtures (thanks to @danicheg for working on that in munit core :)

### DIFF
--- a/client-testkit/shared/src/main/scala/org/http4s/client/testkit/ClientRouteTestBattery.scala
+++ b/client-testkit/shared/src/main/scala/org/http4s/client/testkit/ClientRouteTestBattery.scala
@@ -22,6 +22,7 @@ import cats.effect._
 import cats.syntax.all._
 import fs2._
 import munit._
+import munit.catseffect._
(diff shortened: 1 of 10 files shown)


danicheg (approved):
It's fine with me to adopt the new MUnit version if we would benefit from it.

rossabaker:
This is a published dependency, via laws.  The laws are underdeveloped, rarely depended on, and would not typically be used in applications, so it's not tragic.  But this was a breaking change in a patch, and now production depends on a milestone.

armanbilge:
Sorry, I was completely unaware of that. Why do our laws depend on munit? Typically laws depend on Discipline and not a specific test framework.

rossabaker:
Good question.  Looks like `LawAdapter`.

armanbilge:
Hmm. At a glance I think the Right:tm: thing to do there would have been to use the Cats Effect testkit. Not ruling out that effectful laws are tricky.

We can fix this on 1.x.

PR #6691, opened 2022-09-16
Setup linking paths for OpenSSL on macOS

armanbilge:
Yeah, ok, not a great look 😅 teething pains until I figure out how to simplify some of this stuff.


ChristopherDavenport (approved):
Very necessary if you are on a mac.

Issue #5684, opened 2021-12-08
Node.js ember-core TraversalSpec failure


armanbilge:
```
==> X org.http4s.ember.core.TraversalSpec.Request Encoder/Parser should preserve body with a known uri 0.63s munit.FailException: Failing seed: HpLPaM8JBl1x8n6MOA6WPN6oGhsCMVzbJcB5mmFWb4E=
You can reproduce this failure by adding the following override to your suite:

  override def scalaCheckInitialSeed = "HpLPaM8JBl1x8n6MOA6WPN6oGhsCMVzbJcB5mmFWb4E="
```
https://github.com/http4s/http4s/runs/4444713273?check_suite_focus=true#step:15:3387

armanbilge:
Oh, is this a symptom of https://github.com/http4s/http4s/issues/3935?

armanbilge:
Dupe of https://github.com/http4s/http4s/issues/2794.

rossabaker:
[Another one](https://github.com/http4s/http4s/actions/runs/9522260793/job/26251453037#step:15:239), on 7c15478e7871eb19256e54ebd08140a673c124dd:

==> X org.http4s.ember.core.TraversalSpec.Request Encoder/Parser should preserve body with a known uri  0.863s munit.FailException: Failing seed: V7IhcNXi8qaV9zm8Bxm3vV7hUOb6FQwBf60MFFsfu_J=
You can reproduce this failure by adding the following override to your suite:

  override def scalaCheckInitialSeed = "V7IhcNXi8qaV9zm8Bxm3vV7hUOb6FQwBf60MFFsfu_J="

Issue #3935, opened 2020-11-25
ember-core test flake


rossabaker:
1. Our generators are creating ... weird things
2. Our URI is broken
3. Looks like a redux of #3874, perhaps after #3920

```
2020-11-25T20:35:03.5324818Z [error]   ! preserve existing headers (1 second, 845 ms)
[...]

PR #6709, opened 2022-09-21
`Cross-Origin-Resource-Policy` header model

samspills:
- adds a model for the `Cross-Origin-Resource-Policy` header, following the [mozilla reference](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cross-Origin-Resource-Policy)
- tests as well
  - I looked at some other examples and they were adding the `checkAll` header laws test as well as explicit checks, but it kind of looked to me like those header laws would have already covered these explicit checks? Could anyone explain why we do both? 🙏🏻 
https://github.com/http4s/http4s/blob/d6e5f6ab0db674e857a0f4f4331655dddfc0f35a/tests/shared/src/test/scala/org/http4s/headers/HeaderLaws.scala#L33-L35

related to #5010

### DIFF
--- a/core/shared/src/main/scala/org/http4s/headers/Cross-Origin-Resource-Policy.scala
+++ b/core/shared/src/main/scala/org/http4s/headers/Cross-Origin-Resource-Policy.scala
@@ -1,0 +1,54 @@
+/*
+ * Copyright 2013 http4s.org
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.http4s
+package headers
+
+import cats.parse.Parser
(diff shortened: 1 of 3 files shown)


samspills:
Hmm the test failure is 
`org.http4s.ember.core.TraversalSpec.Request Encoder/Parser should preserve body with a known uri`

and that test is here:
https://github.com/http4s/http4s/blob/d6e5f6ab0db674e857a0f4f4331655dddfc0f35a/ember-core/shared/src/test/scala/org/http4s/ember/core/TraversalSpec.scala#L82-L98

A different test in the suite `Request Encoder/Parser should preserve method with known uri` passes, so I don't think I've broken request parsing entirely 😅

I'm not sure what's up though. Is this a flaky test, or have I broken something I don't understand?

armanbilge:
Yup, sorry, that's a known flake 😕 
- https://github.com/http4s/http4s/issues/3935
- https://github.com/http4s/http4s/issues/5684

armanbilge:
> I looked at some other examples and they were adding the checkAll header laws test as well as explicit checks, but it kind of looked to me like those header laws would have already covered these explicit checks?

It's very possible that this is just a mistake, and the redundant checks are not necessary 😅 good catch!

armanbilge:
Honestly I haven't looked too carefully at it yet :) since it seems to affect many existing tests perhaps it deserves an issue. I'd lean towards removing them from the standpoint of test-time optimization.

RafalSumislawski (approved):
Looks good to me. 

Regarding the tests, I see no problem keeping both the explicit ones and the laws.

armanbilge (approved):
Thank you! 😊

PR #6712, opened 2022-09-22
Cache monoid instance

TimWSpence:
Cache `Monoid[List[Raw]]` instance. On a loaded server running http4s with Jetty, these instances accounted for 14% of our total allocations 😱

### DIFF
--- a/core/shared/src/main/scala/org/http4s/Header.scala
+++ b/core/shared/src/main/scala/org/http4s/Header.scala
@@ -18,6 +18,7 @@ package org.http4s
 
 import cats.Foldable
 import cats.Hash
+import cats.Monoid
 import cats.Order
 import cats.Semigroup
 import cats.Show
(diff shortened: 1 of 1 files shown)


armanbilge (approved):
Cool! Did you benchmark again with these changes?

TimWSpence:
> Cool! Did you benchmark again with these changes?

I'm afraid that I haven't yet. This was spotted as part of a CE3 upgrade that got rolled back for various reasons 😭

bplommer:
Could we change this and similar things in cats instead to have a single `Monoid[List[Any]]` instance that just gets cast to the desired type when summoned?

danicheg:
> Could we change this and similar things in cats instead to have a single Monoid[List[Any]] instance that just gets cast to the desired type when summoned?

Sounds awesome. But will it actually work? I bet the `catsKernelMonoidForList` will be called anyway.

armanbilge:
@bplommer is right, good call 👍

TimWSpence:
Thanks @bplommer and @armanbilge, good call! Sorry just saw this, will do it now!

TimWSpence:
Closing in favour of https://github.com/typelevel/cats/pull/4314

PR #6715, opened 2022-09-23
Use ubuntu-22.04 in ci

armanbilge:
This fixes the Scala Native CI (sorry). See https://github.com/typelevel/fs2/pull/2992 for additional detail.


danicheg (approved):
Since 22.04 is an LTS, I see no evil to pin the ubuntu version in CI if it helps us :+1:

PR #6716, opened 2022-09-23
Call Chunk.toByteBuffer directly

CremboC:
Once https://github.com/typelevel/fs2/pull/2990 is released and we upgrade fs2 version, this operation should be significantly cheaper in many cases (or at worst: the same as before).

See also https://github.com/typelevel/jawn-fs2/pull/376

### DIFF
--- a/circe/src/main/scala/org/http4s/circe/CirceInstances.scala
+++ b/circe/src/main/scala/org/http4s/circe/CirceInstances.scala
@@ -30,8 +30,6 @@ import org.http4s.jawn.JawnInstances
 import org.typelevel.jawn.ParseException
 import org.typelevel.jawn.fs2.unwrapJsonArray
 
-import java.nio.ByteBuffer
-
 trait CirceInstances extends JawnInstances {
   protected val circeSupportParser =
     new CirceSupportParser(maxValueSize = None, allowDuplicateKeys = false)
@@ -54,7 +52,7 @@ trait CirceInstances extends JawnInstances {
 
   private def jsonDecoderByteBufferImpl[F[_]: Concurrent](m: Media[F]): DecodeResult[F, Json] =
     EntityDecoder.collectBinary(m).subflatMap { chunk =>
-      val bb = ByteBuffer.wrap(chunk.toArray)
+      val bb = chunk.toByteBuffer
       if (bb.hasRemaining)
         circeSupportParser
           .parseFromByteBuffer(bb)


diesalbla:
@CremboC Until such time as the FS2 - Jawn releases come through, would it help here to replace the definition of 
```Scala
  DecodeResult.success(m.body.chunks.compile.to(Chunk).map(_.flatten))
```

With something like: 

```Scala
def collectBinary[F[_]: Concurrent](m: Media[F]): DecodeResult[F, Chunk[Byte]] = {
  def flattenChunk[A](ch: Chunk[Chunk[A]]): Chunk[A] = 
    if (ch.size == 1) ch(0) else ch.flatten
  DecodeResult.success(m.body.chunks.compile.to(Chunk).map(flattenChunk))
}
```

So as to get this out of the way.

armanbilge:
Btw ... it's still unclear to me why need to compile to a `Chunk` at all. Can't we just feed it streaming into the Jawn parser?

CremboC:
@armanbilge yes and no...

http4s defaults to using this decoder:
https://github.com/http4s/http4s/blob/b786493904ee905f17bbba13eb38161535093887/circe/src/main/scala/org/http4s/circe/CirceInstances.scala#L71-L82

As you can see, it decides which exact parser to use based on content length. If the content is large enough (>100000), then jawnDecoderImplis used, which looks like
https://github.com/http4s/http4s/blob/b786493904ee905f17bbba13eb38161535093887/jawn/src/main/scala/org/http4s/jawn/JawnInstances.scala#L38-L54

You can see this implementation uses streaming (async) parser of Jawn. However, there's one issue with `AsyncParser` in Jawn: it only exposes three ways to use it [ByteBuffer](https://github.com/typelevel/jawn/blob/6b328ba589cbcb9d2d13ca47679ae02294f42190/parser/shared/src/main/scala/jawn/AsyncParser.scala#L103), [Array[Byte]](https://github.com/typelevel/jawn/blob/6b328ba589cbcb9d2d13ca47679ae02294f42190/parser/shared/src/main/scala/jawn/AsyncParser.scala#L113) and [String](https://github.com/typelevel/jawn/blob/6b328ba589cbcb9d2d13ca47679ae02294f42190/parser/shared/src/main/scala/jawn/AsyncParser.scala#L116). Unfortunately all of these are not good for us as they all require a conversion from `Chunk[Byte]`.

This is a downside of lack of deep integration unfortunately, unless we have own our custom version of `AsyncParser` in `jawn-fs2` or similar.

armanbilge:
> Unfortunately all of these are not good for us as they all require a conversion from `Chunk[Byte]`.

That's not entirely true. A `Chunk[Byte]` is often backed by an `Array[Byte]` or a `ByteBuffer` and a conversion can be cheap/free.

CremboC:
> > Unfortunately all of these are not good for us as they all require a conversion from `Chunk[Byte]`.
> 
> That's not entirely true. A `Chunk[Byte]` is often backed by an `Array[Byte]` or a `ByteBuffer` and a conversion can be cheap/free.

Yes exactly, hence this PR: https://github.com/typelevel/jawn-fs2/pull/376. Due to the smart nature of `toByteBuffer`, it should be on average more efficient (compared to `toArray`, which is always copying the underlying data, it seems).

armanbilge:
Right, so can we use that same technique to interface with the Jawn async parser?

> (compared to `toArray`, which is always copying the underlying data, it seems).

Hm, I see. You can also match for `Chunk.ArraySlice` I believe as a way to get raw access to the underlying data. Or I think there may be a method `.toArraySlice`.

CremboC:
@armanbilge that's basically already happening. If you follow through the code by starting in `JawnInstances.jawnDecoderImpl`.

> Hm, I see. You can also match for Chunk.ArraySlice I believe as a way to get raw access to the underlying data. Or I think there may be a method .toArraySlice.

Yes that's already happening in `Chunk.toByteBuffer`:
```scala
  def toByteBuffer[B >: O](implicit ev: B =:= Byte): JByteBuffer =
    this match {
      case c: Chunk.ArraySlice[_] if c.values.isInstanceOf[Array[Byte]] =>
        JByteBuffer.wrap(c.values.asInstanceOf[Array[Byte]], c.offset, c.length)
      case c: Chunk.ByteBuffer =>
        val b = c.buf.asReadOnlyBuffer
        if (c.offset == 0 && b.position() == 0 && c.size == b.limit()) b
        else {
          (b: JBuffer).position(c.offset.toInt)
          (b: JBuffer).limit(c.offset.toInt + c.size)
          b
        }
      case _ =>
        JByteBuffer.wrap(this.asInstanceOf[Chunk[Byte]].toArray, 0, size)
    }
```

The only additional potential optimisation is pattern match on `Chunk.ArraySlice` and `Chunk.ByteBuffer` ourselves in http4s, but I'm not sure that's a good idea. We will only be saving 1 allocation of a ByteBuffer, which isn't much in this case because it doesn't actually allocate any memory: it's simply a view on top of the array from the `ArraySlice`.

armanbilge:
Ah, ok, sorry for the dumb questions. After reading through more carefully I see now. Thanks for walking me through!

According to the code comment, the `Content-Length` of `100 000` bytes is chosen based on benchmark results. The implication being that it must be more performant to collect and flatten the `Chunk` instead of using the async parser. Do we have a benchmark that can still verify this is the case?

armanbilge:
> Until such time as the FS2 - Jawn releases come through

I just merged this last night. We're good to go, thanks for chasing this across 3 deps!

PR #6718, opened 2022-09-23
Optimise putting headers, add interface to allow caching static headers

CremboC:
I'm opening this PR in draft mode because I don't really understand why the existing logic of *putting* a header is so complicated and perhaps we're missing something.

Note I haven't updated all headers, although probably all headers could be updated to extend the `AsRaw1` interface which allows automatic caching of the value.

**Benchmarks**
Performance of `putHeaders` in `master`, using the same benchmark added in this PR:
```
[info] Benchmark                                                   Mode  Cnt     Score     Error   Units
[info] HeadersBench.putHeaders                                     avgt   10   891.876 ±  15.552   ns/op
[info] HeadersBench.putHeaders:·gc.alloc.rate                      avgt   10  4635.263 ±  81.338  MB/sec
[info] HeadersBench.putHeaders:·gc.alloc.rate.norm                 avgt   10  4552.203 ±   0.003    B/op
[info] HeadersBench.putHeaders:·gc.churn.G1_Eden_Space             avgt   10  4636.559 ±  90.078  MB/sec
[info] HeadersBench.putHeaders:·gc.churn.G1_Eden_Space.norm        avgt   10  4553.459 ±  32.059    B/op
[info] HeadersBench.putHeaders:·gc.churn.G1_Survivor_Space         avgt   10     0.017 ±   0.007  MB/sec
[info] HeadersBench.putHeaders:·gc.churn.G1_Survivor_Space.norm    avgt   10     0.017 ±   0.006    B/op
[info] HeadersBench.putHeaders:·gc.count                           avgt   10   990.000            counts
[info] HeadersBench.putHeaders:·gc.time                            avgt   10  1054.000                ms
```

Performance of `putHeaders` with minor changes in this PR:
```
[info] Benchmark                                                 Mode  Cnt     Score     Error   Units
[info] HeadersBench.putHeaders                                   avgt   10   822.643 ±  80.883   ns/op
[info] HeadersBench.putHeaders:·gc.alloc.rate                    avgt   10  4998.685 ± 472.242  MB/sec
[info] HeadersBench.putHeaders:·gc.alloc.rate.norm               avgt   10  4512.201 ±   0.002    B/op
[info] HeadersBench.putHeaders:·gc.churn.G1_Eden_Space           avgt   10  5001.228 ± 483.802  MB/sec
[info] HeadersBench.putHeaders:·gc.churn.G1_Eden_Space.norm      avgt   10  4514.091 ±  21.311    B/op
[info] HeadersBench.putHeaders:·gc.churn.G1_Survivor_Space       avgt   10     0.014 ±   0.006  MB/sec
[info] HeadersBench.putHeaders:·gc.churn.G1_Survivor_Space.norm  avgt   10     0.013 ±   0.005    B/op
[info] HeadersBench.putHeaders:·gc.count                         avgt   10  1068.000            counts
[info] HeadersBench.putHeaders:·gc.time                          avgt   10  1087.000                ms
```

However, the new `putHeader` (note singular) together with the caching and `AsRaw1` trait, we get the following:
```
[info] Benchmark                                                   Mode  Cnt     Score     Error   Units
[info] HeadersBench.putHeaderRaw                                   avgt   10   100.361 ±  15.176   ns/op
[info] HeadersBench.putHeaderRaw:·gc.alloc.rate                    avgt   10  4597.130 ± 624.751  MB/sec
[info] HeadersBench.putHeaderRaw:·gc.alloc.rate.norm               avgt   10   504.022 ±   0.001    B/op
[info] HeadersBench.putHeaderRaw:·gc.churn.G1_Eden_Space           avgt   10  4603.441 ± 615.393  MB/sec
[info] HeadersBench.putHeaderRaw:·gc.churn.G1_Eden_Space.norm      avgt   10   504.796 ±   4.592    B/op
[info] HeadersBench.putHeaderRaw:·gc.churn.G1_Survivor_Space       avgt   10     0.012 ±   0.005  MB/sec
[info] HeadersBench.putHeaderRaw:·gc.churn.G1_Survivor_Space.norm  avgt   10     0.001 ±   0.001    B/op
[info] HeadersBench.putHeaderRaw:·gc.count                         avgt   10   975.000            counts
[info] HeadersBench.putHeaderRaw:·gc.time                          avgt   10   966.000                ms
```

As you can see, approximately 9x the performance while maintaining the same GC. In order to further improve performance, we'd need to get rid of the `List` data structure here which doesn't seem compatible at all with the idea of `Headers`, given that we need to always `append` rather than `prepend`, so every time we add a header to the list, we have to traverse the whole existing list of headers and add it to the end. Sounds like we should be using `Chain` or something.

### DIFF
--- a/bench/src/main/scala/org/http4s/bench/HeadersBench.scala
+++ b/bench/src/main/scala/org/http4s/bench/HeadersBench.scala
@@ -43,7 +43,7 @@ class HeadersBench {
   @Benchmark
   def putHeaderRaw: Response[IO] =
     baseMessage
-      .putHeader(`Content-Length`(1000L))
-      .putHeader(cachedEncoding)
-      .putHeader(cachedServer)
+      .putRawHeader(`Content-Length`(1000L))
+      .putRawHeader(cachedEncoding)
+      .putRawHeader(cachedServer)
 }
--- a/bench/src/main/scala/org/http4s/bench/HeadersBench.scala
+++ b/bench/src/main/scala/org/http4s/bench/HeadersBench.scala
@@ -1,0 +1,49 @@
+/*
+ * Copyright 2015 http4s.org
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.http4s
+package bench
+
+import cats.effect.IO
+import org.http4s.headers._
+import org.openjdk.jmh.annotations._
+
+import java.util.concurrent.TimeUnit
+
+// sbt "bench/jmh:run -i 5 -wi 5 -f 1 -t 1 org.http4s.bench.HeadersBench"
+@BenchmarkMode(Array(Mode.AverageTime))
+@OutputTimeUnit(TimeUnit.NANOSECONDS)
+@State(Scope.Benchmark)
+class HeadersBench {
+  val baseMessage: Response[IO] = Response[IO]()
+
+  val cachedEncoding: `Content-Encoding` = `Content-Encoding`(ContentCoding.gzip)
+  val cachedServer: Server = Server(ProductId("Benchmark"))
+
+  @Benchmark
+  def putHeaders: Response[IO] =
+    baseMessage
+      .putHeaders(`Content-Length`(1000L))
+      .putHeaders(`Content-Encoding`(ContentCoding.gzip))
+      .putHeaders(Server(ProductId("Benchmark")))
+
+  @Benchmark
+  def putHeaderRaw: Response[IO] =
+    baseMessage
+      .putHeader(`Content-Length`(1000L))
+      .putHeader(cachedEncoding)
+      .putHeader(cachedServer)
+}
--- a/core/shared/src/main/scala/org/http4s/Header.scala
+++ b/core/shared/src/main/scala/org/http4s/Header.scala
@@ -195,6 +195,12 @@ object Header {
     }
   }
 
+  /** Abstracts over headers that can be cached. The most common use case would be
+    * extending a modelled header (e.g. Accept-Encoding), extending the class
+    * and overriding asRaw1 using lazy val.
+    * This enables better efficiency when the same 'static' header is added to the headers
+    * of many responses, as it's not re-created on every single response.
+    */
   trait AsRaw1 {
     def asRaw1: Header.Raw
   }
--- a/core/shared/src/main/scala/org/http4s/Header.scala
+++ b/core/shared/src/main/scala/org/http4s/Header.scala
@@ -179,9 +179,24 @@ object Header {
     // Required for 2.12 to convert variadic args.
     implicit def scalaCollectionSeqToRaw[H](
         h: collection.Seq[H]
+    )(implicit convert: H => ToRaw with Primitive): Header.ToRaw =
+      scalaCollectionListToRaw(h.toList)
+
+    implicit def scalaCollectionListToRaw[H](
+        h: List[H]
     )(implicit convert: H => ToRaw with Primitive): Header.ToRaw = new Header.ToRaw {
-      val values: List[Raw] = h.toList.foldMap(v => convert(v).values)
+      val values: List[Raw] = {
+        val result = List.newBuilder[Raw]
+        h.foreach { v =>
+          result ++= convert(v).values
+        }
+        result.result()
+      }
     }
+  }
+
+  trait AsRaw1 {
+    def asRaw1: Header.Raw
   }
 
   /** Abstracts over Single and Recurring Headers
--- a/core/shared/src/main/scala/org/http4s/Headers.scala
+++ b/core/shared/src/main/scala/org/http4s/Headers.scala
@@ -72,7 +72,7 @@ final class Headers(val headers: List[Header.Raw]) extends AnyVal {
   def put(in: Header.ToRaw*): Headers =
     this ++ Headers(in.values)
 
-  def putOne(in: Header.Raw): Headers =
+  def putOneRaw(in: Header.Raw): Headers =
     if (this.headers.isEmpty) new Headers(List(in))
     else {
       new Headers(this.headers.filterNot(_.name == in.name) :+ in)
--- a/core/shared/src/main/scala/org/http4s/Headers.scala
+++ b/core/shared/src/main/scala/org/http4s/Headers.scala
@@ -78,6 +78,17 @@ final class Headers(val headers: List[Header.Raw]) extends AnyVal {
       new Headers(this.headers.filterNot(_.name == in.name) :+ in)
     }
 
+  def putManyRaw(in: List[Header.Raw]): Headers =
+    if (this.headers.isEmpty) new Headers(in)
+    else {
+      val inNames = {
+        val builder = Set.newBuilder[CIString]
+        in.foreach(h => builder += h.name)
+        builder.result()
+      }
+      new Headers(this.headers.filterNot(h => inNames.contains(h.name)) ::: in)
+    }
+
   def ++(those: Headers): Headers =
     if (those.headers.isEmpty) this
     else if (this.headers.isEmpty) those
--- a/core/shared/src/main/scala/org/http4s/Headers.scala
+++ b/core/shared/src/main/scala/org/http4s/Headers.scala
@@ -72,6 +72,12 @@ final class Headers(val headers: List[Header.Raw]) extends AnyVal {
   def put(in: Header.ToRaw*): Headers =
     this ++ Headers(in.values)
 
+  def putOne(in: Header.Raw): Headers =
+    if (this.headers.isEmpty) new Headers(List(in))
+    else {
+      new Headers(this.headers.filterNot(_.name == in.name) :+ in)
+    }
+
   def ++(those: Headers): Headers =
     if (those.headers.isEmpty) this
     else if (this.headers.isEmpty) those
--- a/core/shared/src/main/scala/org/http4s/Message.scala
+++ b/core/shared/src/main/scala/org/http4s/Message.scala
@@ -176,11 +176,11 @@ sealed trait Message[F[_]] extends Media[F] { self =>
   def putHeaders(headers: Header.ToRaw*): Self =
     transformHeaders(_.put(headers: _*))
 
-  def putHeader(header: Header.Raw): Self =
-    transformHeaders(_.putOne(header))
+  def putRawHeader(header: Header.Raw): Self =
+    transformHeaders(_.putOneRaw(header))
 
-  def putHeader(header: Header.AsRaw1): Self =
-    transformHeaders(_.putOne(header.asRaw1))
+  def putRawHeader(header: Header.AsRaw1): Self =
+    transformHeaders(_.putOneRaw(header.asRaw1))
 
   /** Add a header to these headers.  The header should be a type with a
     * recurring `Header` instance to ensure that the new value can be
--- a/core/shared/src/main/scala/org/http4s/Message.scala
+++ b/core/shared/src/main/scala/org/http4s/Message.scala
@@ -176,6 +176,12 @@ sealed trait Message[F[_]] extends Media[F] { self =>
   def putHeaders(headers: Header.ToRaw*): Self =
     transformHeaders(_.put(headers: _*))
 
+  def putHeader(header: Header.Raw): Self =
+    transformHeaders(_.putOne(header))
+
+  def putHeader(header: Header.AsRaw1): Self =
+    transformHeaders(_.putOne(header.asRaw1))
+
   /** Add a header to these headers.  The header should be a type with a
     * recurring `Header` instance to ensure that the new value can be
     * appended to any existing values.
--- a/core/shared/src/main/scala/org/http4s/Message.scala
+++ b/core/shared/src/main/scala/org/http4s/Message.scala
@@ -176,12 +176,32 @@ sealed trait Message[F[_]] extends Media[F] { self =>
   def putHeaders(headers: Header.ToRaw*): Self =
     transformHeaders(_.put(headers: _*))
 
-  def putRawHeader(header: Header.Raw): Self =
-    transformHeaders(_.putOneRaw(header))
-
+  /** Add the providers headers to the existing headers, replacing those
+    * of the same header name.
+    *
+    * Note this method is different to [[putHeaders]] by being efficient, but slightly
+    * less flexible, as no implicits conversions are provided into a Header.Raw
+    *
+    * Consider using [[putRawHeader]] if you only have a single header to add, as it
+    * avoids creating a list.
+    */
   def putRawHeaders(headers: List[Header.Raw]): Self =
     transformHeaders(_.putManyRaw(headers))
 
+  /** Add the providers headers to the existing headers, replacing those
+    * of the same header name.
+    *
+    * This is a more efficient version of [[putHeaders]]
+    */
+  def putRawHeader(header: Header.Raw): Self =
+    transformHeaders(_.putOneRaw(header))
+
+  /** Add the providers headers to the existing headers, replacing those
+    * of the same header name.
+    *
+    * This is a more efficient version of [[putHeaders]] and similar to [[putRawHeader]],
+    * but accept a modelled header.
+    */
   def putRawHeader(header: Header.AsRaw1): Self =
     transformHeaders(_.putOneRaw(header.asRaw1))
 
--- a/core/shared/src/main/scala/org/http4s/Message.scala
+++ b/core/shared/src/main/scala/org/http4s/Message.scala
@@ -179,6 +179,9 @@ sealed trait Message[F[_]] extends Media[F] { self =>
   def putRawHeader(header: Header.Raw): Self =
     transformHeaders(_.putOneRaw(header))
 
+  def putRawHeaders(headers: List[Header.Raw]): Self =
+    transformHeaders(_.putManyRaw(headers))
+
   def putRawHeader(header: Header.AsRaw1): Self =
     transformHeaders(_.putOneRaw(header.asRaw1))
 
--- a/core/shared/src/main/scala/org/http4s/headers/Content-Encoding.scala
+++ b/core/shared/src/main/scala/org/http4s/headers/Content-Encoding.scala
@@ -29,7 +29,7 @@ object `Content-Encoding` {
 
   implicit val headerInstance: Header[`Content-Encoding`, Header.Single] =
     Header.createRendered(
-      ci"Content-Encoding",
+      name,
       _.contentCoding,
       parse,
     )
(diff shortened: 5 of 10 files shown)


bench/src/main/scala/org/http4s/bench/HeadersBench.scala
```diff
@@ -0,0 +1,35 @@
...
+
+// sbt "bench/jmh:run -i 5 -wi 5 -f 1 -t 1 org.http4s.bench.HeadersBench"
+@BenchmarkMode(Array(Mode.AverageTime))
+@OutputTimeUnit(TimeUnit.NANOSECONDS)
+@State(Scope.Benchmark)
+class HeadersBench {
+  val baseMessage: Response[IO] = Response[IO]()
+
+  val cachedEncoding: `Content-Encoding` = `Content-Encoding`(ContentCoding.gzip)
+  val cachedServer: Server = Server(ProductId("Benchmark"))
```
CremboC, line 18: Some headers can be created once and 'cached'. This is especially relevant for these two headers, and probably for many more.

core/shared/src/main/scala/org/http4s/Header.scala
```diff
@@ -179,11 +179,25 @@ object Header {
     // Required for 2.12 to convert variadic args.
     implicit def scalaCollectionSeqToRaw[H](
         h: collection.Seq[H]
+    )(implicit convert: H => ToRaw with Primitive): Header.ToRaw = scalaCollectionListToRaw(h.toList)
+
+    implicit def scalaCollectionListToRaw[H](
+      h: List[H]
     )(implicit convert: H => ToRaw with Primitive): Header.ToRaw = new Header.ToRaw {
-      val values: List[Raw] = h.toList.foldMap(v => convert(v).values)
```
CremboC, line 183: `foldMap` is really inefficient according to profiling, for whatever reason. I've replaced it with an inlined simple version below.

core/shared/src/main/scala/org/http4s/Message.scala
```diff
@@ -176,6 +176,12 @@ sealed trait Message[F[_]] extends Media[F] { self =>
   def putHeaders(headers: Header.ToRaw*): Self =
     transformHeaders(_.put(headers: _*))
 
+  def putHeader(header: Header.Raw): Self =
+    transformHeaders(_.putOne(header))
```
CremboC, line 180: I've added this interface, which is not used currently, so that middleware like CORS (https://github.com/http4s/http4s/blob/8efa92281ea90e625b9b767dd06052277e78c818/server/shared/src/main/scala/org/http4s/server/middleware/CORS.scala#L223) can be updated to be more efficient via `putHeader`

core/shared/src/main/scala/org/http4s/Header.scala
armanbilge: Possibly related to https://github.com/http4s/http4s/pull/6712?

core/shared/src/main/scala/org/http4s/headers/Content-Encoding.scala
```diff
@@ -25,6 +25,8 @@ object `Content-Encoding` {
 
   private[http4s] val parser = ContentCoding.parser.map(`Content-Encoding`(_))
 
+  val name = ci"Content-Encoding"
+
   implicit val headerInstance: Header[`Content-Encoding`, Header.Single] =
     Header.createRendered(
       ci"Content-Encoding",
```
armanbilge, line 32:
```suggestion
      name,
```

core/shared/src/main/scala/org/http4s/Headers.scala
```diff
@@ -72,6 +72,12 @@ final class Headers(val headers: List[Header.Raw]) extends AnyVal {
   def put(in: Header.ToRaw*): Headers =
     this ++ Headers(in.values)
 
+  def putOne(in: Header.Raw): Headers =
+    if (this.headers.isEmpty) new Headers(List(in))
+    else {
+      new Headers(this.headers.filterNot(_.name == in.name) :+ in)
+    }
```
armanbilge, line 79: This could be more efficient by doing the filtering and the appending at the same time. But it's just a band-aid.

CremboC:
What confuses me the most is why `putHeaders` is currently so indirect.

Say we want to add a `Content-Length: 1000` header.
This is the process (I'll be tracking approx total allocations in parenthesis):
1. Create a `Content-Length` case class (1 alloc).
2. Call `putHeaders`: this accepts a vararg, hence `Seq` (2 allocs) and it must be a `Header.ToRaw`
3. We convert `Content-Length` into `Header.ToRaw with Primitive` (3 allocs)
4. `Header.ToRaw with Primitive` we need a `List[Header.Raw]` so we create a list (4 allocs)
5. `Message.putHeaders` calls `Headers.put`. Let's assume we started with no headers, so we calling `this ++ Headers(in)`.
6. `Headers.apply(in)` is actually more conversion steps: `Seq[Header.Raw]` is converted into `Header.Raw` using `scalaCollectionSeqToRaw`, which converts the `Seq` into `List` (5 allocs). We also call `foldMap` which does many more allocations (to keep it simple let's say we're at 6 allocations), but it depends on the amount of headers.
7. `scalaCollectionSeqToRaw` also calls `convert` implicit, which is `identityToRaw`. This creates another `Header.ToRaw with Primitive` instance with a `List` inside (8 allocations now).
8. `scalaCollectionSeqToRaw` gets called a second time inside when doing `Headers.apply(in)`, but this time with `rawToRaw` implicit (12 allocations now)

And we're done!

@SystemFw @rossabaker @bplommer your names are in the Header code. I was hoping one of you would know if there's a reason why we can't just go `Content-Length() -> Header.Raw` and simply replace/add it to the `headers` in `final class Headers(val headers: List[Header.Raw])`. This PR basically does that (or rather, enables doing that).

core/shared/src/main/scala/org/http4s/Message.scala
wjoel: I understand that we want this to use `HeaderRaw` for a reason (we want the `.name`?), but it seems a bit off that `putHeaders` works on `Header.ToRaw` while `putHeader` works on `Header.Raw`. Maybe `putRawHeader` would be a more appropriate name, if we must have this?

But maybe we wouldn't need it, if `name` was made available on all `Header.ToRaw` in a more efficient way?

bench/src/main/scala/org/http4s/bench/HeadersBench.scala
```diff
@@ -0,0 +1,49 @@
+/*
+ * Copyright 2015 http4s.org
```
wjoel, line 2: 2022? :)

armanbilge: - https://github.com/http4s/http4s/pull/5525

wjoel: But this is a new file, so it shouldn't say 2015, afaict? Ember files are mostly "Copyright 2019 http4s.org", for example. Oh well, not terribly important.

armanbilge: - https://github.com/typelevel/sbt-typelevel/issues/353

core/shared/src/main/scala/org/http4s/Message.scala
CremboC: Honestly I'm not entirely sure why `Header.ToRaw` exists. It's just a `List[Header.Raw]`. What we actually need, and all that `Headers` contains is `List[Header.Raw]`, so I don't really see why we need `ToRaw`.

bench/src/main/scala/org/http4s/bench/HeadersBench.scala
CremboC: Sorry to be honest I just copied over the copyright from another file, didn't think about it much!

CremboC:
I updated `CORS` middleware to use my new `putRawHeader` and `putRawHeaders` methods (pushed in the last commit). I then ran the following benchmark:
```scala
@BenchmarkMode(Array(Mode.AverageTime))
@OutputTimeUnit(TimeUnit.NANOSECONDS)
@State(Scope.Benchmark)
class CORSHeadersBench {
  val CorsMiddleware = CORS.policy
    .withAllowOriginHeader(_ => true)
    .withAllowCredentials(true)
    .withAllowHeadersIn(Set("Content-Type", "Authorization", "*").map(CIString(_)))
    .withMaxAge(1.day)

  private val routes = HttpRoutes.of[IO] {
    case req if req.pathInfo === path"/foo" =>
      Response[IO](Ok).withEntity("foo").pure[IO]
  }

  private val app = CorsMiddleware(routes).orNotFound

  private val exampleOrigin = Origin.Host(Uri.Scheme.https, Uri.RegName("example.com"), None)
  private val exampleOriginHeader = Origin.HostList(NonEmptyList.of(exampleOrigin))
  private val request = Request[IO](uri = uri"/foo").putHeaders(exampleOriginHeader: Origin)

  @Benchmark
  def corsHandling(bh: Blackhole): Unit = {
    bh.consume(app.run(request).unsafeRunSync())
  }
}
```

Using old `putHeaders`:
```
[info] Benchmark                                                       Mode  Cnt      Score     Error   Units
[info] CORSHeadersBench.corsHandling                                   avgt   10  16523.222 ± 667.241   ns/op
[info] CORSHeadersBench.corsHandling:·gc.alloc.rate                    avgt   10    748.393 ±  29.224  MB/sec
[info] CORSHeadersBench.corsHandling:·gc.alloc.rate.norm               avgt   10  13609.781 ±   0.070    B/op
[info] CORSHeadersBench.corsHandling:·gc.churn.G1_Eden_Space           avgt   10    754.788 ±  24.535  MB/sec
[info] CORSHeadersBench.corsHandling:·gc.churn.G1_Eden_Space.norm      avgt   10  13728.009 ± 206.675    B/op
[info] CORSHeadersBench.corsHandling:·gc.churn.G1_Survivor_Space       avgt   10      0.003 ±   0.001  MB/sec
[info] CORSHeadersBench.corsHandling:·gc.churn.G1_Survivor_Space.norm  avgt   10      0.048 ±   0.020    B/op
[info] CORSHeadersBench.corsHandling:·gc.count                         avgt   10    461.000            counts
[info] CORSHeadersBench.corsHandling:·gc.time                          avgt   10    312.000                ms

```

Using new `putRawHeaders/putRawHeader`:
```
[info] Benchmark                                                       Mode  Cnt      Score      Error   Units
[info] CORSHeadersBench.corsHandling                                   avgt   10  14390.159 ± 1065.625   ns/op
[info] CORSHeadersBench.corsHandling:·gc.alloc.rate                    avgt   10    710.729 ±   49.141  MB/sec
[info] CORSHeadersBench.corsHandling:·gc.alloc.rate.norm               avgt   10  11241.467 ±    0.037    B/op
[info] CORSHeadersBench.corsHandling:·gc.churn.G1_Eden_Space           avgt   10    717.102 ±   51.975  MB/sec
[info] CORSHeadersBench.corsHandling:·gc.churn.G1_Eden_Space.norm      avgt   10  11341.475 ±  114.635    B/op
[info] CORSHeadersBench.corsHandling:·gc.churn.G1_Survivor_Space       avgt   10      0.003 ±    0.001  MB/sec
[info] CORSHeadersBench.corsHandling:·gc.churn.G1_Survivor_Space.norm  avgt   10      0.045 ±    0.010    B/op
[info] CORSHeadersBench.corsHandling:·gc.count                         avgt   10    438.000             counts
[info] CORSHeadersBench.corsHandling:·gc.time                          avgt   10    291.000                 ms

```

Note all CORSSuite tests are passing so I _believe_ I didn't break anything.

You can see the improvement is quite significant (around 13% more performant).

core/shared/src/main/scala/org/http4s/Message.scala
SystemFw: I've tried to answer in the other comment, `Header.ToRaw` exists to support an api that works smoothly with typed headers, key-value pairs, Header.Raw, foldable of Header.Raw, etc.
The typed header is especially relevant since it lets you do `putHeaders(ContentType(...), Accept(..))`, which is the same api that existed in the old Scala 2 encoding. In turn, typed headers correctly handle whether they two occurrences of the same header should replace each other, merged, or accumulated in a List

SystemFw: in this particular case, you basically aren't allowing users to work smoothly with typed headers

bench/src/main/scala/org/http4s/bench/HeadersBench.scala
rossabaker: A former committer disagreed heavily with caching values on the header companions, and I never understood, and still don't.  There is a a lot of low-hanging fruit here, probably.

bench/src/main/scala/org/http4s/bench/HeadersBench.scala
rossabaker: Everything defaults to `startYear` for the project.  We _could_ parse git, but this would be prohibitively slow on a project of this size.  If you delete the header and run `headerCreateAll`, it should create one that's consistent with the project configuration.

core/shared/src/main/scala/org/http4s/headers/Content-Encoding.scala
```diff
@@ -25,12 +25,17 @@ object `Content-Encoding` {
...
-      ci"Content-Encoding",
+      name,
       _.contentCoding,
       parse,
     )
 }
 
-final case class `Content-Encoding`(contentCoding: ContentCoding)
+final case class `Content-Encoding`(contentCoding: ContentCoding) extends Header.AsRaw1 {
+  lazy val asRaw1: Header.Raw =
```
rossabaker, line 39: What's preventing us from a generic implementation of this?  Access to the companion and the instance?

If we do this on the case classes, we'll be adding some bytes of overhead to every modeled header.  That's probably okay, but it's not free.  I was expecting to see this on cached values of particularly common header values.

I think rolling this out on some modeled headers, but not all, could be quite confusing.

server/shared/src/main/scala/org/http4s/server/middleware/CORS.scala
```diff
@@ -178,6 +178,9 @@ object CORS {
   )
   val defaultVaryHeader: Header.Raw = Header.Raw(ci"Vary", "Origin,Access-Control-Request-Method")
 
+  private[this] final val accessControlAllowCredentialsHeaderTrue: Header.Raw =
+    Header.Raw(ci"Access-Control-Allow-Credentials", "true")
```
rossabaker, line 182: This is kind of fighting the idea of having modeled headers in the first place.  Something like a `Access-Control-Allow-Credentials.true.toRaw` would achieve the same thing?

Line 179 is "cheating" in the same way, so it's not a new problem.

rossabaker (commented):
Interesting ideas here.  I like where you're going with it, but there's just a tension between what exists and what's added that I'd like to think about during a jog.

wjoel:
Although caching might have a value of its own, I think a lot of the problems stem from `name` not being part of `ToRaw`
(hence my comment in https://github.com/http4s/http4s/pull/6718#discussion_r980415544). Looking at what `ToRaw` means, it obviously couldn't be: `values` of `ToRaw` is a `List[Header.Raw]`, and they can all have different names, as (for example) a list of ("header-name", "value") is supported by the `ToRaw` implicit conversions.

That's all very well, except that a lot of the `put` calls are for just a single header value, and pretty much every call to `put` needs the header name, it becomes very inefficient to have to (first) convert that header to a list of raw headers (even if it's only going to be a single element list, and we know it, and this conversion may happen many times without the caching) and then not have any guarantee (unless we check that it's a singleton list, as in #6201) that all the headers in this list have the same name.

If `ToRaw` had a `values: Map[CIString, List[Header.Value]]` (or `Chain`, or `NonEmptyVector`, or whatever) instead, this would be less problematic. Caching might still be useful, and removing implicits and implementing the `ToRaw` trait directly where possible might also reduce the number of allocations (although the JVM might be able to make them practically free), but at least `Headers` could access the names of headers directly in an efficient manner.

That brings us to https://github.com/http4s/http4s/issues/6720, but that's mostly for 1.x, so this PR or some version of it may still be very useful for 0.23.

It may seem like this is fighting modeled headers, but perhaps it's fighting `ToRaw` even more, or in addition?

core/shared/src/main/scala/org/http4s/headers/Content-Encoding.scala
SystemFw: > Access to the companion and the instance?

Note that going in that direction somewhat gets you back to the old design which required path projections. I'm not at all opposed to experimentation here ofc, but worth calling that out

CremboC: @rossabaker
> I think rolling this out on some modeled headers, but not all, could be quite confusing.

Absolutely, I agree. All modelled headers need to be updated with this, hence the draft mode for this PR.

> If we do this on the case classes, we'll be adding some bytes of overhead to every modeled header. That's probably okay, but it's not free. I was expecting to see this on cached values of particularly common header values.

I'm not sure that's true. Every modelled header is converted into a `Header.Raw` anyway, this does does it pre-emptively and avoids the 12 allocations described in https://github.com/http4s/http4s/pull/6718#issuecomment-1257675532.

> What's preventing us from a generic implementation of this? Access to the companion and the instance?

Sorry not sure I understand what you mean by that. The point of having it as an interface and a lazy val is mostly useful for headers that are re-used often. I'd actually argue it's not very for headers like `Content-Header` as that changes for every request, but `Content-Encoding` probably doesn't change between requests (or ever, as in GZip middleware).

server/shared/src/main/scala/org/http4s/server/middleware/CORS.scala
CremboC: Yeah I don't think much is changing here as you said: the codebase already supports both modelled and non-modelled headers. They're both used heavily in different locations in the code (see CORS header: it's a complete mix of both)

CremboC:
@wjoel so what you're describing is basically what I'm trying to do in this PR. `Header.Raw` is the actual case class being stored in `Headers`. As you said, we need the `name` and the `value` of each header really, and that case class (`Header.Raw`) already has everything we need. For whatever reason (historical, probably), `Message.putHeaders` actually accepts `Header.ToRaw`. For another, currently unclear to me reason, the way we convert headers into `Header.Raw` is rather complicated and requires many intermediate steps.

This is what the key idea in this PR is to change that: simply allow ~adding~ **putting** `Header.Raw` into `Headers` as that doesn't require any conversions.

We can probably separate this into two PRs:
1. Just adding method on `Message` which is something like `putRawHeader(header: Header.Raw)`
2. Update middleware to use `putRawHeader` for better performance

SystemFw:
> For whatever reason (historical, probably), Message.putHeaders actually accepts Header.ToRaw. For another, currently unclear to me reason, the way we convert headers into Header.Raw is rather complicated and requires many intermediate steps.

@CremboC I've tried to give an answer to both of these questions in two comments already, not quite sure how else to get the point across :)

> Message.putHeaders actually accepts Header.ToRaw. 

The reason `putHeaders` accepts `Header.ToRaw` is to support _both_ modelled headers and raw headers (and in fact it also supports key value pairs, and foldables of headers).
This is actually in the scaladoc for `ToRaw`:
```scala
  /** Target for implicit conversions to Header.Raw from modelled
    * headers and key-value pairs.
    *
    * A method taking variadic `ToRaw` arguments will allow taking
    * heteregenous arguments, provided they are either:
    *
    * - A value of type `A`  which has a `Header[A]` in scope
    * - A (name, value) pair of `String`, which is treated as a `Recurring` header
    * - A `Header.Raw`
    * - A `Foldable` (`List`, `Option`, etc) of the above.
    *
    * @see [[org.http4s.Headers.apply]]
    */
    ```
In particular, right now you can do 
```scala
putHeaders(Accept(..), Allow(..), "myHeader" -> "myValue")
```
> another, currently unclear to me reason, the way we convert headers into Header.Raw is rather complicated and requires many intermediate steps.

The reason for that is that modelled headers guarantee correct treatment of duplicate headers, which depending on the header can mean "replace old one", "merge with old one", or "have a list with both"

PR #6719, opened 2022-09-23
Fix Keep-Alive header flaky test

### LINKED ISSUE #6676: `KeepAliveSuite` flake
I saw this on a JS run, but at a glance I have no reason to believe it's JS specific.

```
org.http4s.headers.KeepAliveSuite:
failing seed for header.header does not match arbitrary name is hJyBbXICp5OwOcoOzsaBHYTzIF9y5zo3pHoSs3oX0gO=
==> X org.http4s.headers.KeepAliveSuite.Keep-Alive: header.header does not match arbitrary name 0.09s munit.FailException: Failing seed: UjCaqEjILdSnoSjRhoICQxtFpTAoxvYsTBQfsngpQhG=

  override val scalaCheckInitialSeed = "UjCaqEjILdSnoSjRhoICQxtFpTAoxvYsTBQfsngpQhG="

Exception raised on property evaluation.
> Exception: org.http4s.ParseFailure: Invalid Keep-Alive header: Reserved token of list List(token, max) was found in the extensions.
```

FrancescoSerra:
Closes #6676
Remove 'max' deprecated token.
Fix generator so that it doesn't generates reserved token strings.
Add test that proved that parser fails if reserved 'token' is present in the extensions

### DIFF
--- a/core/shared/src/main/scala/org/http4s/headers/Keep-Alive.scala
+++ b/core/shared/src/main/scala/org/http4s/headers/Keep-Alive.scala
@@ -52,11 +52,11 @@ keep-alive-extension = token [ "=" ( token / quoted-string ) ]
       extension: List[(String, Option[String])],
   ): ParseResult[`Keep-Alive`] =
     if (timeoutSeconds.isDefined || max.isDefined || extension.nonEmpty) {
-      val reservedTokens = List("token", "max")
-      if (extension.exists(p => reservedTokens.contains(p._1))) {
+      val reservedToken ="token"
+      if (extension.exists(p => p._1 == reservedToken)) {
         ParseResult.fail(
           "Invalid Keep-Alive header",
-          s"Reserved token of list $reservedTokens was found in the extensions.",
+          s"Reserved token '$reservedToken' was found in the extensions.",
         )
       } else {
         val validatedTimeoutSeconds = timeoutSeconds.traverse(t => nonNegativeLong(t, "timeout"))
--- a/core/shared/src/main/scala/org/http4s/headers/Keep-Alive.scala
+++ b/core/shared/src/main/scala/org/http4s/headers/Keep-Alive.scala
@@ -52,7 +52,7 @@ keep-alive-extension = token [ "=" ( token / quoted-string ) ]
       extension: List[(String, Option[String])],
   ): ParseResult[`Keep-Alive`] =
     if (timeoutSeconds.isDefined || max.isDefined || extension.nonEmpty) {
-      val reservedToken ="token"
+      val reservedToken = "token"
       if (extension.exists(p => p._1 == reservedToken)) {
         ParseResult.fail(
           "Invalid Keep-Alive header",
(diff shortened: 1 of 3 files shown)


armanbilge:
Thanks for the investigation and fix! Sorry for the uninformed (i.e. me 😜 ), can you give a brief tl;dr on the story with this `max` token? Is there an RFC?

FrancescoSerra:
> can you give a brief tl;dr on the story with this `max` token? Is there an RFC?

yes: as reported on [the relative RFC](https://datatracker.ietf.org/doc/html/draft-thomson-hybi-http-timeout-03#section-2.2.1), the parameter has been deprecated (and in fact, it's even reported as such[ in a comment](https://github.com/http4s/http4s/blob/series/0.23/core/shared/src/main/scala/org/http4s/headers/Keep-Alive.scala#L41) on the relative file in http4s.

core/shared/src/main/scala/org/http4s/headers/Keep-Alive.scala
```diff
@@ -52,11 +52,11 @@ keep-alive-extension = token [ "=" ( token / quoted-string ) ]
       extension: List[(String, Option[String])],
   ): ParseResult[`Keep-Alive`] =
     if (timeoutSeconds.isDefined || max.isDefined || extension.nonEmpty) {
-      val reservedTokens = List("token", "max")
-      if (extension.exists(p => reservedTokens.contains(p._1))) {
+      val reservedToken = "token"
```
RafalSumislawski, line 55: Isn't that supposed to be `val reservedToken = "timeout"`? There's nothing in the RFC about "token". And it would make sense that the predefined names ("timeout", "max") cannot be used as extension names.

RafalSumislawski, line 55: If the intention of this PR is to drop support for the deprecated `max` then IMO it should be dropped completely (including the `case class Max` the `max` parameter in `apply` etc), not just from the extension list. But that breaks bin-compat. So maybe it would be better to not remove the max at all? If we remove it only from the list of extensions then we end up with the strange situation where max can be define in two different places.

RafalSumislawski (changes requested):

core/shared/src/main/scala/org/http4s/headers/Keep-Alive.scala
FrancescoSerra: I think you're very much on point. Now thinking of it, I couldn't really work out why the original code mentioned "token". Thanks!

FrancescoSerra: Good shout. I wasn't aware of the places where it exists. Thoughts @armanbilge ?

armanbilge: Right, we can fix this on 1.x however.

The part I still haven't grokked is how this change relates to the flaky test?

PR #6728, opened 2022-10-03
Provide a dev shell per Java version

rossabaker:

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -3,7 +3,7 @@
 
   inputs = {
     typelevel-nix.url = "github:typelevel/typelevel-nix";
-    nixpkgs.follows = "typelevel-nix/nixpkgs";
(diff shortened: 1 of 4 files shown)


flake.nix
```diff
@@ -3,7 +3,7 @@
 
   inputs = {
     typelevel-nix.url = "github:typelevel/typelevel-nix";
-    nixpkgs.follows = "typelevel-nix/nixpkgs";
+    # nixpkgs.follows = "typelevel-nix/nixpkgs";
```
rossabaker, line 6: Temurin was recently added to nixpkgs.  We can roll this back after the next typelevel-nix update, which should kick off tonight.

armanbilge:
Merging because CI doesn't run on PRs against PR branches 🤔

PR #6738, opened 2022-10-08
Server Auth Middleware: no Kleisli nor OptionT

diesalbla:
The constructors of auth middlewares need to take a function for extracting the authentication "token" or data from the request. These are passed as functions, but we can make our API easier by just taking them as functions.

### DIFF
--- a/server/shared/src/main/scala/org/http4s/server/package.scala
+++ b/server/shared/src/main/scala/org/http4s/server/package.scala
@@ -113,11 +113,13 @@ package object server {
   /** An HTTP middleware converts an [[HttpRoutes]] to another.
     */
   type HttpMiddleware[F[_]] =
+    // HttpRoutes[F] => HttpRoutes[F]
     Middleware[OptionT[F, *], Request[F], Response[F], Request[F], Response[F]]
 
   /** An HTTP middleware that authenticates users.
     */
   type AuthMiddleware[F[_], T] =
+    // AuthedRoutes[T, F] => HttpRoutes[F]
     Middleware[OptionT[F, *], AuthedRequest[F, T], Response[F], Request[F], Response[F]]
 
   /** An HTTP middleware that adds a context.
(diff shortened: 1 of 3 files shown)


wjoel:
> These are passed as functions, but we can make our API easier by just taking them as functions.
*are passed as Kleisli, I suppose? Looks like a solid improvement to me, at least.

satorg:
Just to clarify: what is the simplification expected here?
Do you mean that instead of
```AuthMiddleware(Kleisli { req => myAuth(req) })```
with this change
```AuthMiddleware( req => myAuth(req).value )```
can be called?
Assuming that `myAuth: Request[F] => OptionT[F, Response]`

diesalbla:
@satorg Almost. It means that you can apply it to `req => myAuth(req)`, where `myAuth` is of type `Request[F] => F[Option[Response]]`, so neither Kleisli nor `OptionT`.

satorg:
I see. I think it might be better approach that instead of replacing Klesli[OptionT[F, *] ... (and thus getting rid of it) completely
provide an alternate syntax in the form you're suggesting.

Just because some people may prefer using monad transformers while some other prefer avoiding it.

wjoel:
> I see. I think it might be better approach that instead of replacing Klesli[OptionT[F, *] ... (and thus getting rid of it) completely provide an alternate syntax in the form you're suggesting.
> 
> Just because some people may prefer using monad transformers while some other prefer avoiding it.

Some `AuthMiddleware.applyK`, `AuthMiddleware.withFallThroughK` etc. wouldn't hurt too much, then (although honestly, who does? ... and there doesn't seem to be any precedence of those in http4s at the moment), but the major point of those changes is to make new users of htt4ps not have to use `Kleisli` directly, so `Kleisli` shouldn't be in `apply` and other "default" methods.

diesalbla:
@wjoel @satorg We could also keep the original methods as "deprecated", since they may still be needed for bincompat. But yeah, the goal is that people should, hence forth, switch to plain ones.

wjoel:
Sure, makes sense for 0.23 etc., but less so for 1.x imo. I just assumed this PR was for 1.x as it's targeted for `main`.

satorg:
> the goal is that people should, hence forth, switch to plain ones.

I personally feel not that certain about "forcing" people to do something.

What if their auth methods were created as Kleisli/OptionT compositions from the very beginning and they are pretty comfortable with monad transformers otherwise? For such people the deprecation or the removal would be not a simplification but rather an annoying change.

Not sure though on how vast is the majority of such people...

wjoel:
> > the goal is that people should, hence forth, switch to plain ones.
> 
> I personally feel not that certain about "forcing" people to do something.

"force" is a strong word, but the current situation "forces" people to use `Kleisli` which is unfamiliar to many, especially users new to pure FP (and arguably, even users not so new to FP). I believe this PR is related to https://github.com/http4s/http4s/discussions/4846, which explains the reasons for moving away from `Kleisli` in public APIs, at least for the default. It's about helping new users of http4s have a good experience.

> What if their auth methods were created as Kleisli/OptionT compositions from the very beginning and they are pretty comfortable with monad transformers otherwise? For such people the deprecation or the removal would be not a simplification but rather an annoying change.
> 
> Not sure though on how vast is the majority of such people...

On the other hand, such advanced users would have no problems with making the changes necessary. They changes required are really small, as shown in this PR. An annoyance, perhaps, but a small price to pay if it helps improve the http4s experience for new users, increasing adoption, no?

satorg:
I would say I was more concerned  about getting rid of `OptionT` rather that `Kleisli` because the former is way more common in regular functions... But on the second thought.. un-lifting from monad transformers usually is more straightforward than lifting into them. So I think you're right that it may not be worth to worry about them.

danicheg (commented):
Probably, we can have both variants of those functions – with `Kleisli` + `OptionT` and unwrapped ones. In that case, we wouldn't force users to change their code. WDYT? 

Overall, `Kleisli` is a base abstraction of http4s for composition. Its use is coherent in the whole library. Perhaps, breaking that coherency should be done for the library conjunctly (for the API, at least).

diesalbla:
> Probably, we can have both variants of those functions – with Kleisli + OptionT and unwrapped ones. In that case, we wouldn't force users to change their code. WDYT?

I have changed the PR to 1. rebase on `series/0.23` and target there; 2. kept the Kleisli versions with a deprecation notice and add the other ones.  

> Overall, Kleisli is a base abstraction of http4s for composition. Its use is coherent in the whole library. Perhaps, breaking that coherency should be done for the library conjunctly (for the API, at least).

The goal is to reduce use of Kleisli, and this is an incremental step on that direction.

rossabaker (commented):
The additions seem fine, but until and unless Kleisli and OptionT get consistently removed from the more fundamental interface, it seems pretty arbitrary to deprecate them here.

diesalbla:
> The additions seem fine, but until and unless Kleisli and OptionT get consistently removed from the more fundamental interface, it seems pretty arbitrary to deprecate them here.

I would not call them arbitrary. There are places and parameters where we also just pass an effectful function `A => F[B]` instead of the `Kleisli`. 

The key question would be: what does the use of `Kleisli` convery to the reader of an API, to anybody approaching the method, that a mere function would not? 

This was partially prompted by this discussion https://github.com/http4s/http4s/discussions/4846 about `Kleisli` and the phrase 

> mostly everyone using Scala understands functions and Option, but not everyone coming to the language will know what Kleisli and OptionT is.

The other reason was that, while drafting this pull request, https://github.com/http4s/http4s/pull/6740, I realised that these `Kleisli` in particular could be removed.

PR #6740, opened 2022-10-10
HTTP without Kleislis / transformers

diesalbla:
An old idea prompted back after seeing https://github.com/http4s/http4s/discussions/4846. 

The idea is to look for approaches to define the core `Http`, `HttpApp`, `HttpRoutes`, etc types, not as types aliases for "Kleisli" but as traits with an `apply` method. This allows to use single-abstract-method syntax in several definitions...


rossabaker (changes requested):
~~Sorry, I don't like this at all.~~  [Edit: that reads much harsher than I meant.  Thank you for making this concrete to discuss.  Seeing it laid out, I don't agree with the direction.]

1. It still aliases concrete types.  The way these get expanded into something else in error messages is confusing to newcomers.
2. It still requires syntax for basic operations like `orNotFound`.  A concrete `HttpRoutes` could do this with a plain old type.
3. Calling those "map" and "contramap" when the output and input parameters are fixed is confusing.
4. There can no longer be Cats instances. 
5. Instead of the server middlewares consistently being `Kleisli => Kleisli`, we now have an awkward mix of `Http => Http` and `Function => Function`.  If we kept them all in `Http`, we'd lose the parametricity that allows Auth and Context to participate.

Now, the Client is a single-abstract method with fixed request input and response output.  Why has this mostly worked?

1. There are more compelling operations derived from the client's SAM than the server's.  The entire interface, except run. 
2. There are more obvious ways to stack effects on the server than the client.

This doesn't align client with server more closely (a `type HttpResource[F] = Http[Resource[F, *], F]` might, but brings back the problem of fallthrough routing), nor does it share more middleware between server and client.

I think this trades Kleisli confusion for bespoke confusion.

diesalbla:
> 1. It still aliases concrete types.  The way these get expanded into something else in error messages is confusing to newcomers.
> 2. It still requires syntax for basic operations like `orNotFound`.  A concrete `HttpRoutes` could do this with a plain old type.

The solution to these would be to have both `HttpApp`, `HttpRoutes`, as well as `ContextRoutes`, each as their own separate trait. 

```Scala
trait HttpApp[F[_]] { def apply(r: Request[F]) : F[Response[F]] }
trait HttpRoutes[F[_]] { def apply(r: Request[F]) : F[Option[Response[F]]] }
```

> 3. Calling those "map" and "contramap" when the output and input parameters are fixed is confusing.

Those were added because they are called in other places, but other names may be easier..

rossabaker:
With the separate traits, there is no longer a relationship between HttpApp and HttpRoutes.  Everything being Kleisli, and the middleware parametric on `F[_]`, is what makes the middleware generally applicable.

cats-mtl could achieve the same ends without the Kleisli dependency, and I think is still the correct approach for middleware.

diesalbla:
> cats-mtl could achieve the same ends without the Kleisli dependency, and I think is still the correct approach for middleware

Would that mean basing the middleware constructors, not on them working on a fixed data type, but on them working on a type-class over "Kleisli"-like data-types, one for which both `HttpApp` or `HttpRoutes` would have a defined instance?

rossabaker:
Here is the cats-mtl draft: https://github.com/http4s/http4s/pull/4758.

> Does that mean that a desirable change would be for HttpApp and HttpRoutes not to be aliases of a same datatype, Kleisli or otherwise, but separate types? So as to make it easier for people to recognise types in error messages as types in the code?

Under cats-mtl, yes, but I don't know that the `Http` prefix is still appropriate once `Request` and `Response` are no longer hardcoded.  They might be newtypes around a `run` value that encompasses all the appropriate effects.

PR #6741, opened 2022-10-10
Add Uri.Host.fromString method

CraigHammondDexcom:
Adds a `fromString` method to `org.http4s.Uri.Host`. While it is possible to get a `org.http4s.Uri.Host` from a `String` using `org.http4s.headers.Forwarded.Host.fromString` to get a `org.http4s.headers.Forwarded.Host` which itself has a `host` field of type `org.http4s.Uri.Host`, it felt less discoverable. This also bring `Host` in line with other objects in `org.http4s.Uri` which themselves have a convenient `fromString` method.

### DIFF
--- a/core/shared/src/main/scala/org/http4s/Uri.scala
+++ b/core/shared/src/main/scala/org/http4s/Uri.scala
@@ -682,6 +682,9 @@ object Uri extends UriPlatform {
 
   object Host {
 
+    def fromString(s: String): ParseResult[Host] =
+      ParseResult.fromParser(Parser.host, "Invalid host")(s)
+
     /** Create a [[Host]] value from an [[com.comcast.ip4s.IpAddress]].
       *
       * This is a convenience method for creating the correct host based on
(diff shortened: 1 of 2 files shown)


armanbilge:
> While it is possible to get a `org.http4s.Uri.Host` from a `String` using `org.http4s.headers.Forwarded.Host.fromString` to get a `org.http4s.headers.Forwarded.Host` which itself has a `host` field of type `org.http4s.Uri.Host`, it felt less discoverable.

You could also parse your string to an `ip4s.Host`, and then convert that to an http4s `Host`. Or if you have a constant string then simply `host"..."`. But agree on discoverability :)

armanbilge (approved):
Thanks!

satorg:
Would it make sense to add `Host.unsafeFromString(s: String): Host`? Most of the model parsers that implement `fromString` also implement `unsafeFromString` usually.

satorg (approved):

PR #6749, opened 2022-10-13
docs: remove `Future[_]` from presupplied encoders

yoshinorin:
follow up: #1694
Presupplied future has been deleted in 13527d6c320efd25800680a8fd841365dec7965e. Isn't it?

 

Thank you :)

### DIFF
--- a/docs/docs/entity.md
+++ b/docs/docs/entity.md
@@ -81,8 +81,7 @@ println(response.flatMap(_.as[Resp]).unsafeRunSync())
 The `EntityEncoder`/`EntityDecoder`s shipped with http4s.
 
 ### Raw Data Types
-These are already in implicit scope by default, e.g. `String`, `File`,
-`Future[_]`, and `InputStream`. Consult [EntityEncoder] and [EntityDecoder] for
+These are already in implicit scope by default, e.g. `String`, `File`, and `InputStream`. Consult [EntityEncoder] and [EntityDecoder] for
 a full list.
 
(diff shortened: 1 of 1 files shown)


docs/docs/entity.md
```diff
@@ -81,8 +81,7 @@ println(response.flatMap(_.as[Resp]).unsafeRunSync())
 The `EntityEncoder`/`EntityDecoder`s shipped with http4s.
 
 ### Raw Data Types
-These are already in implicit scope by default, e.g. `String`, `File`,
-`Future[_]`, and `InputStream`. Consult [EntityEncoder] and [EntityDecoder] for
+These are already in implicit scope by default, e.g. `String`, `File`, and `InputStream`. Consult [EntityEncoder] and [EntityDecoder] for
```
armanbilge, line 84: I didn't check do we still have it for `InputStream` as well? Personally I would be happy to see those go away 😂

yoshinorin: Yes. `inputStreamEncoeder` still exists.

https://github.com/http4s/http4s/blob/c73f42ef557f0181c43cb3431f7b5b780d99b288/core/shared/src/main/scala/org/http4s/EntityEncoder.scala#L190

armanbilge: Ah, it is for `F[InputStream]`. I suppose that is okay. Thanks!

armanbilge (approved):
Thanks!

PR #6751, opened 2022-10-14
Scala specific nowarnX annotations

satorg:
Introduces a bunch of Scala-version specific `@nowarnX` annotation which map to a real `scala.annotation.nowarn` according to the table:
| Scala | 2.12 | 2.13 | 3 |
| -- | -- | -- | -- |
| `@nowarn2` | ✅ | ✅ | ☠️ |
| `@nowarn212` | ✅ | ☠️ | ☠️ |
| `@nowarn213` | ☠️ | ✅ | ☠️ |
| `@nowarn3` | ☠️ | ☠️ | ✅ |

Also this PR re-fixes https://github.com/http4s/http4s/pull/6750#discussion_r994977997 in a nicer and more generic way (hopefully). As well as the proposed annotations can be used in the future to address similar issues related to differences in Scalac warning detection/generation.

Basically it should allow to avoid inventing ad-hoc solutions every time and let us simply use a set of `nowarn` options that works for all current cross-build Scala versions in the project.

Also, Scala3 may use a set of filters for compiler messages that can be different from one used by Scala2. Therefore, these annotations could allow adjusting the filters in a finer and more precise way.

---
In fact, this approach can be generalized to make these custom annotations available for other projects. Perhaps we could consider extracting them into a separate <sub>(micro-)</sub>module for that 🤔

### DIFF
--- a/project/Http4sPlugin.scala
+++ b/project/Http4sPlugin.scala
@@ -50,6 +50,7 @@ object Http4sPlugin extends AutoPlugin {
     headerSources / excludeFilter := HiddenFileFilter,
     doctestTestFramework := DoctestTestFramework.Munit,
     libraryDependencies += scalacCompatAnnotation,
+    unusedCompileDependenciesFilter ~= { _ & (_ == scalacCompatAnnotation) },
     ivyConfigurations += CompileTime,
     Compile / unmanagedClasspath ++= update.value.select(configurationFilter(CompileTime.name)),
   )
--- a/project/Http4sPlugin.scala
+++ b/project/Http4sPlugin.scala
@@ -47,6 +47,7 @@ object Http4sPlugin extends AutoPlugin {
   override lazy val projectSettings: Seq[Setting[_]] = Seq(
     headerSources / excludeFilter := HiddenFileFilter,
     doctestTestFramework := DoctestTestFramework.Munit,
+    libraryDependencies += scalacCompatAnnotation,
   )
 
   def extractApiVersion(version: String) = {
@@ -124,6 +125,7 @@ object Http4sPlugin extends AutoPlugin {
     val munitDiscipline = "2.0.0-M3"
     val netty = "4.1.84.Final"
     val quasiquotes = "2.1.0"
+    val scalacCompat = "0.1.0"
     val scalacheck = "1.17.0"
     val scalacheckEffect = "2.0.0-M2"
     val scalaJavaLocales = "1.4.1"
@@ -181,6 +183,8 @@ object Http4sPlugin extends AutoPlugin {
   lazy val nettyBuffer = "io.netty" % "netty-buffer" % V.netty
   lazy val nettyCodecHttp = "io.netty" % "netty-codec-http" % V.netty
   lazy val quasiquotes = "org.scalamacros" %% "quasiquotes" % V.quasiquotes
+  lazy val scalacCompatAnnotation =
+    "org.typelevel" %% "scalac-compat-annotation" % V.scalacCompat % CompileTime
   lazy val scalacheck = Def.setting("org.scalacheck" %%% "scalacheck" % V.scalacheck)
   lazy val scalacheckEffect =
     Def.setting("org.typelevel" %%% "scalacheck-effect" % V.scalacheckEffect)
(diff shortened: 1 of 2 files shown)


danicheg:
> I'd really love to see this in a microlib instead of buried inside http4s.

this reminds me of days when typelevel has its own fork of scalac. what a wild time it was!

satorg:
@armanbilge ,
> If I understand correctly, this seems like a solution for https://github.com/typelevel/sbt-typelevel/issues/378 as well?

Yeah, although I wouldn't bet it will cover everything, but it should help with the cases where different Scala versions can treat `@nowarn` differently.

> I'd really love to see this in a microlib instead of buried inside http4s. I have a bad feeling we will be need more of it

Agree. I anticipate that many of warnings in Cats could be addressed with help of these annotations we well.

satorg:
Oh, if we decide to extract it into a separate lib, then we could add even more convenient stuff there. For example, there's `@unused` annotation in Scala v2.13 and v3 which is pretty handy, but does not exist in v2.12.

armanbilge:
Right, like we add kind-projector and better-monadic-for automatically.

danicheg:
so, who will drop typelevel/better-nowarn?

satorg:
Right, thank you @armanbilge. I'm going to move this code to the new repo shortly. 
But I wonder what would be a correct roadmap for rolling it out? I guest it might be
1. Release `scalac-compat` first with these annotations included
2. Add this <sub>micro-</sub>lib to the `sbt-typelevel` as a default (but configurable) dependency for projects that use the plugin.
3. Release a new `sbt-typelevel` version.
4. Update `http4s` to make it using the new `sbt-typelevel`.
5. Apply the fix to this file (as well as others where applicable).

Would it work, wdyt?

armanbilge:
@satorg great, thanks for your work on this! That sounds good to me. Probably I would skip steps (2) and (3) while we try this out in http4s, and maybe another project that needs it. Then when we are more confident we can add it to sbt-typelevel, definitely in time for 0.5.x.

satorg:
The annotation classes are copied to here: typelevel/scalac-compat#1

project/Http4sPlugin.scala
```diff
@@ -29,6 +29,8 @@ object Http4sPlugin extends AutoPlugin {
   val scala_212 = "2.12.17"
   val scala_3 = "3.2.0"
 
+  private val CompileTime = config("compile-time").hide
```
armanbilge, line 32: ... another thing to upstream 😂 

Awesome though, thanks for figuring this out!!

armanbilge (approved):

satorg:
@armanbilge @danicheg I re-worked this PR to make use of the recently released `scalac-compat-annotation` library.
Also I made that dependency compile-time only.

project/Http4sPlugin.scala
```diff
@@ -47,6 +49,9 @@ object Http4sPlugin extends AutoPlugin {
   override lazy val projectSettings: Seq[Setting[_]] = Seq(
     headerSources / excludeFilter := HiddenFileFilter,
     doctestTestFramework := DoctestTestFramework.Munit,
+    libraryDependencies += scalacCompat,
+    ivyConfigurations += CompileTime,
+    Compile / unmanagedClasspath ++= update.value.select(configurationFilter(CompileTime.name)),
```
satorg, line 54: The magic is borrowed from typelevel/cats#1889.
But perhaps it would be better to put it into `sbt-typelevel`.

satorg: See also sbt/sbt#2503

project/Http4sPlugin.scala
satorg: Yes, I think it could be added to sbt-typelevel.
It does not look invasive – if not used, it should not cause any effect.

satorg:
Omg... CI failed due to `unusedCompileDependenciesTest`.. 🤦  Guess, which one...

danicheg (approved):
I can't speak about SBT magic, but I like this work overall!

satorg:
Just for a record: `unusedCompileDependenciesTest` is from [sbt-explicit-dependencies](https://github.com/cb372/sbt-explicit-dependencies), which in turn is brought in by [sbt-http4s-org](https://github.com/http4s/sbt-http4s-org). So it looks specific to **http4s**, not for all **sbt-typelevel**-dependent projects.

armanbilge:
Thanks, yes. We have a number of workarounds in place for it (it also doesn't work with Scala.js and Scala Native).

PR #6752, opened 2022-10-15
Special-case `EmptyBody` in request encoder

armanbilge:
This PR replays https://github.com/http4s/http4s/pull/6444 for the ember request encoder. Specifically, when the body is known (by reference equality) to be the `EmptyBody` i.e. an empty `Stream`, then a `Content-Length: 0` header is added and chunked transfer encoding is disabled. This relates to https://github.com/http4s/http4s/issues/4935, and possibly even fixes it.

### DIFF
--- a/ember-core/shared/src/main/scala/org/http4s/ember/core/Encoder.scala
+++ b/ember-core/shared/src/main/scala/org/http4s/ember/core/Encoder.scala
@@ -140,9 +140,13 @@ private[ember] object Encoder {
         }
 
         if (!chunked && !appliedContentLength && !NoPayloadMethods.contains(req.method)) {
-          stringBuilder.append(chunkedTransferEncodingHeaderRaw).append(CRLF)
-          chunked = true
-          ()
+          if (req.body == EmptyBody) {
+            stringBuilder.append(zeroContentLengthRaw).append(CRLF)
+            chunked = false
+          } else {
+            stringBuilder.append(chunkedTransferEncodingHeaderRaw).append(CRLF)
+            chunked = true
+          }
         }
 
         // Final CRLF terminates headers and signals body to follow.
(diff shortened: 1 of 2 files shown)


ember-core/shared/src/main/scala/org/http4s/ember/core/Encoder.scala
```diff
@@ -139,7 +139,12 @@ private[ember] object Encoder {
           }
         }
 
-        if (!chunked && !appliedContentLength && !NoPayloadMethods.contains(req.method)) {
+        if (
+          !appliedContentLength && req.body == EmptyBody && !NoPayloadMethods.contains(req.method)
```
hamnis, line 143: should one maybe use reference equality `eq` directly instead of `equals(==)` ?

danicheg, line 143: It'd be good to define `def`s for those boolean expressions for better readability.

danicheg (approved):
Should we do the same fix for blaze?

armanbilge:
> Should we do the same fix for blaze?

I don't think blaze has this problem, based on the experiments in https://github.com/http4s/http4s/issues/4935.

ember-core/shared/src/main/scala/org/http4s/ember/core/Encoder.scala
armanbilge: Good point, I'll make this change in https://github.com/http4s/http4s/pull/6753. We need to fix it in at least one other place.

armanbilge: I agree about readability but I feel superstitious about performance. Probably it will be inlined by JIT 🤔

armanbilge: It will be nice when in scala 3 we write `inline` and sleep easy.

PR #6753, opened 2022-10-15
Low-hanging Ember optimizations

armanbilge:
Just some micro-optimizations I noticed while working on https://github.com/http4s/http4s/pull/6752.


armanbilge:
Actually the optimization I really wanted to make is to this:
https://github.com/http4s/http4s/blob/04b2bbbd1787ac7084f589240ed8a062f38ed584/ember-core/shared/src/main/scala/org/http4s/ember/core/Encoder.scala#L78

I was hoping we could go directly from `StringBuilder` to `Array[Byte]` without making an intermediate `String`. One way to do this is to use `CharBuffer.wrap(sb)` with the ISO_8859_1 `CharsetEncoder`. But there is an encoding fast-path for `String` that is not available for a generic `CharBuffer` so I'm skeptical if this would improve performance.

ember-core/shared/src/main/scala/org/http4s/ember/core/Encoder.scala
```diff
@@ -47,31 +47,24 @@ private[ember] object Encoder {
...
-      resp.headers
-        .get[`Content-Length`]
-        .foreach { _ =>
-          appliedContentLength = true
-        }
-
       // Apply each header followed by a CRLF
       resp.headers.foreach { h =>
         if (h.isNameValid) {
+          appliedContentLength |= h.name == `Content-Length`.name
```
ChristopherDavenport, line 53: I do not recognize this operator, is this optimized over an if with assignment. If so great, but specialized mutability stuff I don't really know.

armanbilge: Good point. Maybe I should write it like this.
```scala
appliedContentLength = appliedContentLength || h.name == `Content-Length`.name
```

ember-core/shared/src/main/scala/org/http4s/ember/core/ChunkedEncoding.scala
```diff
@@ -139,21 +139,23 @@ private[ember] object ChunkedEncoding {
         .map { case (headerP, rest) => Trailers(headerP.headers, rest) }
     }
 
-  private val lastChunk: Chunk[Byte] =
-    Chunk.byteVector((ByteVector('0') ++ crlf ++ crlf).compact)
+  private[this] val lastChunk = {
```
diesalbla, line 142:
```suggestion
  private[this] val lastChunk: Stream[Nothing, Byte] = {
```

diesalbla (approved):

ember-core/shared/src/main/scala/org/http4s/ember/core/Encoder.scala
```diff
@@ -121,33 +115,25 @@ private[ember] object Encoder {
...
-        if (
-          !appliedContentLength && req.body == EmptyBody && !NoPayloadMethods.contains(req.method)
-        ) {
+        def isEmptyBody = req.body eq EmptyBody
+        def isNoPayloadMethod = NoPayloadMethods.contains(req.method)
+        if (!appliedContentLength && isEmptyBody && !isNoPayloadMethod) {
           stringBuilder.append(zeroContentLengthRaw).append(CRLF)
           chunked = false
-        } else if (!chunked && !appliedContentLength && !NoPayloadMethods.contains(req.method)) {
+        } else if (!chunked && !appliedContentLength && !isNoPayloadMethod) {
```
armanbilge, line 134: @danicheg wdyt? I studied the bytecode and looks like these are emitted as private static methods, which I think it's safe to assume that the JIT will inline.

ember-core/shared/src/main/scala/org/http4s/ember/core/Encoder.scala
```diff
@@ -121,33 +115,25 @@ private[ember] object Encoder {
...
-            ()
           }
         }
 
-        if (
-          !appliedContentLength && req.body == EmptyBody && !NoPayloadMethods.contains(req.method)
-        ) {
+        def isEmptyBody = req.body eq EmptyBody
+        def isNoPayloadMethod = NoPayloadMethods.contains(req.method)
+        if (!appliedContentLength && isEmptyBody && !isNoPayloadMethod) {
```
danicheg, line 131: Continuing bike-shedding, speaking in terms of saving a few ns, shouldn't we place `!isNoPayloadMethod` ahead of `isEmptyBody`? I mean, `!isNoPayloadMethod` might be `false` a bit frequently then `isEmptyBody`.

ember-core/shared/src/main/scala/org/http4s/ember/core/Encoder.scala
danicheg: Can confirm – readability is better now :)

ember-core/shared/src/main/scala/org/http4s/ember/core/Encoder.scala
armanbilge: That might be, but I think reference equality `eq` is faster than set membership `.contains`.

armanbilge:
Apologies, since this PR is sitting I started committing additional optimizations :) each commit should be individual reviewable.

valencik (approved):
I think this looks good. Lots of `ByteVector.view` to avoid copying arrays and some nice `unchunks`.
It does make me think we could benefit from some more ember benchmarks. I'm going to look into how hard those might be to add.

ember-core/shared/src/main/scala/org/http4s/ember/core/ChunkedEncoding.scala
```diff
@@ -139,21 +139,23 @@ private[ember] object ChunkedEncoding {
         .map { case (headerP, rest) => Trailers(headerP.headers, rest) }
     }
 
-  private val lastChunk: Chunk[Byte] =
-    Chunk.byteVector((ByteVector('0') ++ crlf ++ crlf).compact)
+  private[this] val lastChunk: Stream[fs2.Pure, Byte] = {
+    val bytes = new Array[Byte](5)
+    bytes(0) = '0'.toByte
+    `\r\n\r\n`.copyToArray(bytes, 1)
+    Stream.chunk(Chunk.array(bytes))
```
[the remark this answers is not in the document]
armanbilge: Thanks yeah, that's a lot nicer 😂

rossabaker:
Anything holding this up from merge?

armanbilge:
Oh, when Andrew first reviewed they weren't a maintainer so I couldn't merge. Now I can!

PR #6755, opened 2022-10-15
Ember-Core H2Client impl initialiser: extract local methods.

diesalbla:
- Extract a type for the Connection Pool maintained.
- Extract an extension (implicit) class around the mutable variable with methods to clean closed connections, schedule a periodic clean, and stop (run shutdown) of all connections.

### DIFF
--- a/ember-core/shared/src/main/scala/org/http4s/ember/core/h2/H2Client.scala
+++ b/ember-core/shared/src/main/scala/org/http4s/ember/core/h2/H2Client.scala
@@ -297,6 +297,7 @@ private[ember] class H2Client[F[_]](
 
 private[ember] object H2Client {
   private type TinyClient[F[_]] = Request[F] => Resource[F, Response[F]]
+
   def impl[F[_]: Async](
       onPushPromise: (
           org.http4s.Request[fs2.Pure],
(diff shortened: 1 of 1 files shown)


ember-core/shared/src/main/scala/org/http4s/ember/core/h2/H2Client.scala
```diff
@@ -313,6 +313,27 @@ private[ember] class H2Client[F[_]](
 
 private[ember] object H2Client {
   private type TinyClient[F[_]] = Request[F] => Resource[F, Response[F]]
+
+  private type ConnectionPool[F[_]] = Map[H2Client.RequestKey, (H2Connection[F], F[Unit])]
+
+  implicit private class PoolOps[F[_]](private val pool: Ref[F, ConnectionPool[F]]) extends AnyVal {
```
danicheg, line 319: I'm in favor of doing regular private functions instead of syntax. Because I'm almost sure that `AnyVal` wouldn't help, and the class will be instantiated.

ember-core/shared/src/main/scala/org/http4s/ember/core/h2/H2Client.scala
```diff
@@ -313,6 +313,27 @@ private[ember] class H2Client[F[_]](
 
 private[ember] object H2Client {
   private type TinyClient[F[_]] = Request[F] => Resource[F, Response[F]]
+
+  private type ConnectionPool[F[_]] = Map[H2Client.RequestKey, (H2Connection[F], F[Unit])]
```
danicheg, line 317: It looks too specialized and won't be reused, so I would make it private with the most restricted scope.

ember-core/shared/src/main/scala/org/http4s/ember/core/h2/H2Client.scala
diesalbla: Ok. Turned into internal functions of the main method.

ember-core/shared/src/main/scala/org/http4s/ember/core/h2/H2Client.scala
diesalbla: Written into the function.

PR #6756, opened 2022-10-15
Encode trailer headers in Ember

armanbilge:
Fixes another issue discovered while working on https://github.com/http4s/http4s/pull/6753.


ember-core/shared/src/main/scala/org/http4s/ember/core/ChunkedEncoding.scala
```diff
@@ -140,20 +142,37 @@ private[ember] object ChunkedEncoding {
...
         Chunk.byteVector(
           ByteVector.view(bv.size.toHexString.toUpperCase.getBytes) ++ crlf ++ bv ++ crlf
         )
-    _.mapChunks { ch =>
-      encodeChunk(ch.toByteVector)
-    } ++ Stream.chunk(lastChunk)
+    _.mapChunks(ch => encodeChunk(ch.toByteVector)) ++
+      Stream.chunk(lastChunk) ++
+      Stream.evalUnChunk(trailers.map(encodeTrailers(_))) ++
+      Stream.chunk(Chunk.byteVector(crlf))
```
armanbilge, line 159: We can make some additional optimizations here. Just waiting to see whether this PR or https://github.com/http4s/http4s/pull/6753 will merge first.

ember-core/shared/src/main/scala/org/http4s/ember/core/ChunkedEncoding.scala
```diff
@@ -140,20 +142,37 @@ private[ember] object ChunkedEncoding {
...
+  }
+
+  private def encodeTrailers(trailers: Headers): Chunk[Byte] = {
+    val stringBuilder = new StringBuilder()
```
diesalbla, line 163: As a small suggestion: a StringBuilder starts with a capacity of 16 characters, but it duplicates every time. https://docs.oracle.com/javase/7/docs/api/java/lang/StringBuilder.html#StringBuilder()
https://github.com/AdoptOpenJDK/openjdk-jdk11/blob/master/src/java.base/share/classes/java/lang/AbstractStringBuilder.java

Perhaps it could be more efficient to first calculate the total size of the trailers names, and values, with the colons, and initialise the StringBuilder to that size.

It could also help to keep the same instance of `StringBuilder`, with its underlying array, from use to use.

armanbilge: @diesalbla I agree. We encode headers in at least three places, ideally these could share a single, optimized method.

> Perhaps it could be more efficient to first calculate the total size of the trailers names, and values, with the colons, and initialise the StringBuilder to that size.

In that case, I think if we could work directly with a `Array[Char]` instead of a `StringBuilder`. See my comment in  https://github.com/http4s/http4s/pull/6753#issuecomment-1279670163.



> It could also help to keep the same instance of `StringBuilder`, with its underlying array, from use to use.

I thought about this as well, but this would have problems with thread-safety.

ember-core/shared/src/main/scala/org/http4s/ember/core/ChunkedEncoding.scala
```diff
@@ -140,20 +142,37 @@ private[ember] object ChunkedEncoding {
     }
 
   private val lastChunk: Chunk[Byte] =
-    Chunk.byteVector((ByteVector('0') ++ crlf ++ crlf).compact)
+    Chunk.byteVector((ByteVector('0') ++ crlf).compact)
 
   /** Encodes chunk of bytes to http chunked encoding.
     */
-  def encode[F[_]]: Pipe[F, Byte, Byte] = {
+  def encode[F[_]: Functor](trailers: F[Headers]): Pipe[F, Byte, Byte] = {
```
ChristopherDavenport, line 149: Just making sure we actually set this key, for every entity we are writing?

ChristopherDavenport (approved):

ember-core/shared/src/main/scala/org/http4s/ember/core/ChunkedEncoding.scala
ChristopherDavenport: https://github.com/http4s/http4s/blob/series/0.23/core/shared/src/main/scala/org/http4s/Message.scala#L210 - Resolving my own question

PR #6762, opened 2022-10-18
Add a hasPayload method to Method

daddykotex:
Discussed here, https://github.com/disneystreaming/smithy4s/pull/522, we think it's a good idea to know whether a Method is expected to have a payload or not. @armanbilge suggested to open a PR so here it is.

### DIFF
--- a/core/shared/src/main/scala/org/http4s/Method.scala
+++ b/core/shared/src/main/scala/org/http4s/Method.scala
@@ -52,6 +52,8 @@ final class Method private (val name: String, val isSafe: Boolean, val isIdempot
       case _ => false
     }
 
+  val hasPayload: Boolean = Method.hasPayload(this)
+
   override def hashCode(): Int = MurmurHash3.stringHash(name, Method.HashSeed)
 
   override def toString(): String = name
@@ -74,6 +76,11 @@ object Method {
     new Method(name, isSafe = false, isIdempotent = true)
   private def safe(name: String) =
     new Method(name, isSafe = true, isIdempotent = true)
+
+  private val NoPayloadMethods: Set[Method] =
+    Set(Method.GET, Method.DELETE, Method.CONNECT, Method.TRACE)
+
+  def hasPayload(method: Method): Boolean = !NoPayloadMethods(method)
 
   val ACL: Method = idempotent("ACL")
   val `BASELINE-CONTROL`: Method = idempotent("BASELINE-CONTROL")


core/shared/src/main/scala/org/http4s/Method.scala
```diff
@@ -75,6 +77,11 @@ object Method {
   private def safe(name: String) =
     new Method(name, isSafe = true, isIdempotent = true)
 
+  private val NoPayloadMethods: Set[Method] =
+    Set(Method.GET, Method.DELETE, Method.CONNECT, Method.TRACE)
```
danicheg, line 81: What about `HEAD` and `OPTIONS`?

core/shared/src/main/scala/org/http4s/Method.scala
```diff
@@ -52,6 +52,8 @@ final class Method private (val name: String, val isSafe: Boolean, val isIdempot
       case _ => false
     }
 
+  val hasPayload: Boolean = Method.hasPayload(this)
```
danicheg, line 55: Since it wouldn't be used frequently, shouldn't we define this as a `def`?

core/shared/src/main/scala/org/http4s/Method.scala
armanbilge: There is a table of methods in the wiki page, describing the payload body as either "yes", "optional", or "no".
https://en.wikipedia.org/wiki/Hypertext_Transfer_Protocol#Request_methods

armanbilge: That is to say, perhaps the name of the method should be `payloadRequired` ?

armanbilge:
I've been staring at the RFC and actually I don't see anything about "required" payloads.
https://datatracker.ietf.org/doc/html/rfc7231#section-4.3

For POST and PUT it seems to say nothing.

For the "optional" methods it says this.
```
   A payload within a GET request message has no defined semantics;
   sending a payload body on a GET request might cause some existing
   implementations to reject the request.
```

And for TRACE it says this.
```
A client MUST NOT send a message body in a TRACE request.
```

rossabaker:
Bodies (at least a Content-Length: 0 header) are [required at the SHOULD level](https://httpwg.org/specs/rfc9110.html#field.content-length) for some methods.

> A user agent SHOULD send Content-Length in a request when the method defines a meaning for enclosed content and it is not sending [Transfer-Encoding](https://httpwg.org/specs/rfc9112.html#field.transfer-encoding). For example, a user agent normally sends Content-Length in a POST request even when the value is 0 (indicating empty content). A user agent SHOULD NOT send a Content-Length header field when the request message does not contain content and the method semantics do not anticipate such data.

There are at least three states here: SHOULD, SHOULD NOT, and MUST NOT.  My inclination is to have a sum type, but there's so much weird shit in HTTP, the proposed boolean encoding is probably the safest.

daddykotex:
The more I read about this, the more I think this is not a good idea and that such a helper method will only lead to bugs in very specific scenario:

> I have a server X and I can curl it with `curl -XGET 'http://some-site.com/endpoint' -d 'my-data'` and it works fine but I can't do it in http4s-ember. The body is never sent

I feel like the spec is so ambiguous and written after existing implementation were already in the wild that it's a per-use case scenario. Write your code, test it and if it works, move on. If it does not work, add a `content-length` header (if you did not have one), or use an explicit empty body stream and try again.

PR #6765, opened 2022-10-21
A bit more efficient size checking for `Seq` collections

danicheg:
```
[info] Benchmark                 (listSize)  Mode  Cnt  Score   Error  Units
[info] SizeCheckingBench.size            10  avgt    5  2.106 ± 0.008  ns/op
[info] SizeCheckingBench.size           100  avgt    5  2.105 ± 0.010  ns/op
[info] SizeCheckingBench.size          1000  avgt    5  2.103 ± 0.006  ns/op
[info] SizeCheckingBench.size         10000  avgt    5  2.101 ± 0.021  ns/op
[info] SizeCheckingBench.sizeIs          10  avgt    5  2.003 ± 0.016  ns/op
[info] SizeCheckingBench.sizeIs         100  avgt    5  2.003 ± 0.006  ns/op
[info] SizeCheckingBench.sizeIs        1000  avgt    5  2.014 ± 0.016  ns/op
[info] SizeCheckingBench.sizeIs       10000  avgt    5  2.012 ± 0.014  ns/op
```
So, yeah, this is a 5% improvement at max (tested on `List` by equality check with '1').

### DIFF
--- a/core/shared/src/main/scala/org/http4s/Uri.scala
+++ b/core/shared/src/main/scala/org/http4s/Uri.scala
@@ -425,13 +425,13 @@ object Uri extends UriPlatform {
     def splitAt(idx: Int): (Path, Path) =
       if (idx <= 0) {
         (Path.empty, this)
-      } else if (idx < segments.size) {
+      } else if (segments.sizeIs > idx) {
         val (start, end) = segments.splitAt(idx)
         (
           Path(start, absolute = absolute),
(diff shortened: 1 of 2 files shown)


armanbilge (approved):

armanbilge:
Actually, can't we do this in 0.23? Or is `sizeIs` not available in 2.12 maybe?

danicheg:
> Or is sizeIs not available in 2.12 maybe?

yeah, it's from 2.13

PR #6766, opened 2022-10-22
A bit more efficient string building from `Seq` collections

danicheg:
```
[info] Benchmark                      (listSize)  Mode  Cnt  Score   Error  Units
[info] stringBuildingWithIterator             10  avgt    5  6.249 ± 0.014  ns/op
[info] stringBuildingWithIterator            100  avgt    5  6.252 ± 0.056  ns/op
[info] stringBuildingWithIterator           1000  avgt    5  6.241 ± 0.006  ns/op
[info] stringBuildingWithoutIterator          10  avgt    5  6.576 ± 0.103  ns/op
[info] stringBuildingWithoutIterator         100  avgt    5  6.599 ± 0.006  ns/op
[info] stringBuildingWithoutIterator        1000  avgt    5  6.584 ± 0.044  ns/op
```
A little tricky (not that tricky tbh) thing that brings a 5% performance improvement for string building.


armanbilge (approved):
Local mutability ftw 😂

PR #6767, opened 2022-10-22
Utilize `Using` from std instead of the self-written `AutoCloseableResource`

danicheg:
@armanbilge suggested some time ago using `munit.CatsEffectFixtures` or `cats.effect.Resource.fromAutoCloseable` instead. But it seems like a more solid refactoring of tests. Probably, we can create a low-hanging fruit issue for that.


armanbilge (approved):
I tried removing these in https://github.com/http4s/http4s/pull/6768 but it becomes a cross-compile mess. I think it's fine.

PR #6769, opened 2022-10-22
[wip] Service monad sketch

rossabaker:
Spinoff of https://github.com/http4s/http4s/discussions/4846.

### DIFF
--- a/core/shared/src/main/scala/org/http4s/Service.scala
+++ b/core/shared/src/main/scala/org/http4s/Service.scala
@@ -2,6 +2,7 @@ package org.http4s
 
 import cats.Monad
 import cats.StackSafeMonad
+import cats.SemigroupK
 import cats.effect.kernel.Resource
 
 import Service._
@@ -19,9 +20,17 @@ sealed abstract class Service[F[_], -A, B] {
         case None => Resource.pure[F, Option[C]](None)
       })
     )
+
+  def orElse[AA <: A](default: => Service[F, AA, B]): Service[F, AA, B] =
+    apply(a =>
+      run(a).flatMap {
+        case some: Some[_] => Resource.pure(some)
+        case None => default.run(a)
+      }
+    )
 }
 
-object Service extends ServiceKleisliMonadInstance {
+object Service extends ServiceInstances {
   private[Service] final case class Run[F[_], -A, B](run: A => Resource[F, Option[B]])
       extends Service[F, A, B]
 
@@ -32,9 +41,10 @@ object Service extends ServiceKleisliMonadInstance {
     apply(Function.const(Resource.pure(Some(b))))
 }
 
-private[http4s] sealed abstract class ServiceMonad[F[_], A]
+private[http4s] sealed abstract class ServiceInstance[F[_], A]
     extends Monad[Service[F, A, *]]
-    with StackSafeMonad[Service[F, A, *]] {
+    with StackSafeMonad[Service[F, A, *]]
+    with SemigroupK[Service[F, A, *]] {
   override def map[B, C](fb: Service[F, A, B])(f: B => C): Service[F, A, C] =
     fb.map(f)
 
@@ -43,9 +53,13 @@ private[http4s] sealed abstract class ServiceMonad[F[_], A]
 
   def flatMap[B, C](fb: Service[F, A, B])(f: B => Service[F, A, C]): Service[F, A, C] =
     fb.flatMap(f)
+
+  def combineK[B](x: Service[F, A, B], y: Service[F, A, B]): Service[F, A, B] =
+    x.orElse(y)
 }
 
-private[http4s] trait ServiceKleisliMonadInstance {
-  implicit def catsMonadForHttp4sService[F[_], A, B]: Monad[Service[F, A, *]] =
-    new ServiceMonad[F, A] {}
+private[http4s] trait ServiceInstances {
+  implicit def catsInstancesForHttp4sService[F[_], A, B]
+      : Monad[Service[F, A, *]] with SemigroupK[Service[F, A, *]] =
+    new ServiceInstance[F, A] {}
 }
--- a/core/shared/src/main/scala/org/http4s/Service.scala
+++ b/core/shared/src/main/scala/org/http4s/Service.scala
@@ -1,6 +1,7 @@
 package org.http4s
 
-import cats.Functor
+import cats.Monad
+import cats.StackSafeMonad
 import cats.effect.kernel.Resource
 
 import Service._
@@ -9,24 +10,42 @@ sealed abstract class Service[F[_], -A, B] {
   def run: A => Resource[F, Option[B]]
 
   def map[C](f: B => C): Service[F, A, C] =
-    Run((a: A) => run(a).map(_.map(f)))
+    apply((a: A) => run(a).map(_.map(f)))
+
+  def flatMap[AA <: A, C](f: B => Service[F, AA, C]) =
+    apply((a: AA) =>
+      run(a).flatMap(_ match {
+        case Some(b) => f(b).run(a)
+        case None => Resource.pure[F, Option[C]](None)
+      })
+    )
 }
 
-object Service extends ServiceKleisliFunctorInstance {
+object Service extends ServiceKleisliMonadInstance {
   private[Service] final case class Run[F[_], -A, B](run: A => Resource[F, Option[B]])
       extends Service[F, A, B]
 
   def apply[F[_], A, B](run: A => Resource[F, Option[B]]): Run[F, A, B] =
     Run(run)
+
+  def pure[F[_], A, B](b: B): Service[F, A, B] =
+    apply(Function.const(Resource.pure(Some(b))))
 }
 
-private[http4s] sealed abstract class ServiceFunctor[F[_], A] extends Functor[Service[F, A, *]] {
-
-  def map[B, C](fb: Service[F, A, B])(f: B => C): Service[F, A, C] =
+private[http4s] sealed abstract class ServiceMonad[F[_], A]
+    extends Monad[Service[F, A, *]]
+    with StackSafeMonad[Service[F, A, *]] {
+  override def map[B, C](fb: Service[F, A, B])(f: B => C): Service[F, A, C] =
     fb.map(f)
+
+  def pure[B](b: B): Service[F, A, B] =
+    Service.pure(b)
+
+  def flatMap[B, C](fb: Service[F, A, B])(f: B => Service[F, A, C]): Service[F, A, C] =
+    fb.flatMap(f)
 }
 
-private[http4s] trait ServiceKleisliFunctorInstance {
-  implicit def catsFunctorForHttp4sService[F[_], A, B]: Functor[Service[F, A, *]] =
-    new ServiceFunctor[F, A] {}
+private[http4s] trait ServiceKleisliMonadInstance {
+  implicit def catsMonadForHttp4sService[F[_], A, B]: Monad[Service[F, A, *]] =
+    new ServiceMonad[F, A] {}
 }
--- a/core/shared/src/main/scala/org/http4s/Service.scala
+++ b/core/shared/src/main/scala/org/http4s/Service.scala
@@ -1,0 +1,14 @@
+package org.http4s
+
+import cats.effect.kernel.Resource
+
+sealed abstract class Service[F[_], -A, B] { self =>
+  def run: A => Resource[F, Option[B]]
+}
+
+object Service {
+  private[Service] final case class Run[F[_], -A, B](run: A => Resource[F, Option[B]])
+
+  def apply[F[_], A, B](run: A => Resource[F, Option[B]]) =
+    Run(run)
+}
--- a/core/shared/src/main/scala/org/http4s/Service.scala
+++ b/core/shared/src/main/scala/org/http4s/Service.scala
@@ -1,8 +1,8 @@
 package org.http4s
 
 import cats.Monad
+import cats.MonoidK
 import cats.StackSafeMonad
-import cats.SemigroupK
 import cats.effect.kernel.Resource
 
 import Service._
@@ -39,12 +39,15 @@ object Service extends ServiceInstances {
 
   def pure[F[_], A, B](b: B): Service[F, A, B] =
     apply(Function.const(Resource.pure(Some(b))))
+
+  def pass[F[_], A, B]: Service[F, A, B] =
+    apply(Function.const(Resource.pure(None)))
 }
 
 private[http4s] sealed abstract class ServiceInstance[F[_], A]
     extends Monad[Service[F, A, *]]
     with StackSafeMonad[Service[F, A, *]]
-    with SemigroupK[Service[F, A, *]] {
+    with MonoidK[Service[F, A, *]] {
   override def map[B, C](fb: Service[F, A, B])(f: B => C): Service[F, A, C] =
     fb.map(f)
 
@@ -56,10 +59,13 @@ private[http4s] sealed abstract class ServiceInstance[F[_], A]
 
   def combineK[B](x: Service[F, A, B], y: Service[F, A, B]): Service[F, A, B] =
     x.orElse(y)
+
+  def empty[B]: Service[F, A, B] =
+    Service.pass
 }
 
 private[http4s] trait ServiceInstances {
   implicit def catsInstancesForHttp4sService[F[_], A, B]
-      : Monad[Service[F, A, *]] with SemigroupK[Service[F, A, *]] =
+      : Monad[Service[F, A, *]] with MonoidK[Service[F, A, *]] =
     new ServiceInstance[F, A] {}
 }
--- a/core/shared/src/main/scala/org/http4s/Service.scala
+++ b/core/shared/src/main/scala/org/http4s/Service.scala
@@ -1,14 +1,32 @@
 package org.http4s
 
+import cats.Functor
 import cats.effect.kernel.Resource
 
-sealed abstract class Service[F[_], -A, B] { self =>
+import Service._
+
+sealed abstract class Service[F[_], -A, B] {
   def run: A => Resource[F, Option[B]]
+
+  def map[C](f: B => C): Service[F, A, C] =
+    Run((a: A) => run(a).map(_.map(f)))
 }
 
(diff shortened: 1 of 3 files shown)


core/shared/src/main/scala/org/http4s/Service.scala
```diff
@@ -0,0 +1,88 @@
...
+        case None => default.run(a)
+      }
+    )
+
+  def local[C](f: C => A): Service[F, C, B] =
+    apply(c => run(f(c)))
+}
+
+object Service extends ServiceInstances {
+  private[Service] final case class Run[F[_], -A, B](run: A => Resource[F, Option[B]])
```
rossabaker, line 39: We can optimize this with a PartialFunction-like encoding.

core/shared/src/main/scala/org/http4s/Service.scala
```diff
@@ -0,0 +1,88 @@
...
+import cats.Monad
+import cats.MonoidK
+import cats.StackSafeMonad
+import cats.effect.kernel.Resource
+import cats.mtl.Local
+
+import Service._
+
+sealed abstract class Service[F[_], -A, B] {
+  def run: A => Resource[F, Option[B]]
```
rossabaker, line 13: Probably a `runOrElse` or `seal` is a better public interface.

core/shared/src/main/scala/org/http4s/Service.scala
armanbilge: How would that work? `PartialFunction` has:
```
def isDefinedAt(x: A): Boolean
```
which seems problematic in an effectual setting. Even if it's `F[Boolean]`.

project/Http4sPlugin.scala
```diff
@@ -97,6 +97,7 @@ object Http4sPlugin extends AutoPlugin {
     val caseInsensitive = "1.3.0"
     val cats = "2.8.0"
     val catsEffect = "3.3.14"
+    val catsMtl = "1.2.1"
```
armanbilge, line 100:
```suggestion
    val catsMtl = "1.3.0"
```

core/shared/src/main/scala/org/http4s/Service.scala
```diff
@@ -0,0 +1,107 @@
...
+      extends Service[F, A, B]
+
+  def apply[F[_], A, B](run: A => Resource[F, Option[B]]): Run[F, A, B] =
+    Run(run)
+
+  def pure[F[_], A, B](b: B): Service[F, A, B] =
+    apply(Function.const(Resource.pure(Some(b))))
+
+  def pass[F[_], A, B]: Service[F, A, B] =
+    apply(Function.const(Resource.pure(None)))
```
armanbilge, line 68: We can cache this and cast 😅

core/shared/src/main/scala/org/http4s/Service.scala
```diff
@@ -0,0 +1,107 @@
...
+
+  def ask[F[_], A]: Service[F, A, A] =
+    Service(a => Resource.pure(Some(a)))
```
armanbilge, line 71: Same with this one I reckon.

core/shared/src/main/scala/org/http4s/Service.scala
rossabaker: See [01275e4](https://github.com/http4s/http4s/pull/6769/commits/01275e4896c8dd5b9b16c30d6c99f4615aa798e1).  I only compiled it, with a lot of shady casts, so it might not work at all.

core/shared/src/main/scala/org/http4s/Service.scala
rossabaker: Yes, if the applyOrElse encoding falls on its face.

core/shared/src/main/scala/org/http4s/Service.scala
```diff
@@ -0,0 +1,128 @@
...
+private[http4s] trait ServiceInstances {
+  implicit def catsInstancesForHttp4sService[F[_], A]
+      : Monad[Service[F, A, *]] with Alternative[Service[F, A, *]] =
+    new ServiceInstance[F, A] {}
+
+  implicit def catsMtlLocalForHttp4sService[F[_], A]: Local[Service[F, A, *], A] =
+    new Local[Service[F, A, *], A] {
+      val applicative = Applicative[Service[F, A, *]]
+      def ask[E >: A]: Service[F, A, E] =
+        Service.ask.asInstanceOf[Service[F, A, E]]
```
rossabaker, line 124: Shady.  I got rid of the covariance of A.  I can try to bring it back, but it has to be on applyOrElse, and that ripples a lot of other places.

core/shared/src/main/scala/org/http4s/Service.scala
```diff
@@ -0,0 +1,128 @@
...
+sealed abstract class Service[F[_], A, B] { self =>
+  def applyOrElse(a: A, default: A => Resource[F, B]): Resource[F, B]
+
+  def map[C](f: B => C): Service[F, A, C] =
+    new Service[F, A, C] {
+      def applyOrElse(a: A, default: A => Resource[F, C]): Resource[F, C] = {
+        val rb = self.applyOrElse(a, checkFallback)
+        if (isFallback(rb)) default(a) else rb.map(f)
+      }
+    }
```
rossabaker, line 37: Maybe I can lambda-lift these because they're SAMs and make them less noisy? :shrug:

core/shared/src/main/scala/org/http4s/Service.scala
```diff
@@ -0,0 +1,128 @@
...
+import cats.Alternative
+import cats.Applicative
+import cats.Monad
+import cats.StackSafeMonad
+import cats.effect.kernel.Resource
+import cats.mtl.Local
+
+import Service._
+
+sealed abstract class Service[F[_], A, B] { self =>
```
diesalbla, line 28: I think the following could improve readability: 

```suggestion
sealed abstract class Service[A, F[_], B] { self =>
```

This puts the parameters in the same order as they appear in the underlying function `A => F[B]` or `A => Resource[F, B]` type.

This is something I would have also wished for Kleisli.

SystemFw: That messes partial unification up

diesalbla: > That messes partial unification up

Is this a known problem in Scala 2.13, that partial unification would fail selectively over the first vs the second parameter? Does it still happens in Scala 3?

Speaking about the type system, would it be possible to make `A` contravariant and `B` covariant?

rossabaker: A was contravariant until the last commit and could be again with some additional work.

Covariance of B runs into https://typelevel.org/blog/2018/09/29/monad-transformer-variance.html, though fs2 somehow made it work with similar signatures, so, maybe.

rossabaker:
This ought to be benchmarked before and after [01275e4](https://github.com/http4s/http4s/pull/6769/commits/01275e4896c8dd5b9b16c30d6c99f4615aa798e1) so we can tell whether the more complicated applyOrElse encoding makes a difference.  I'm pretty sure it doesn't make a difference in the real world, but at least it would be something ready to point to in the next FUDdening.

core/shared/src/main/scala/org/http4s/Service.scala
SystemFw: > s this a known problem in Scala 2.13, that partial unification would fail selectively over the first vs the second parameter? Does it still happens in Scala 3?

it's not a "problem", it's literally the definition of partial unification. It's like expecting that you can have `foo: String => Int => Boolean` and do `foo(3)`

core/shared/src/main/scala/org/http4s/Service.scala
diesalbla: Using SAM is nice, at lease to remove the `new Service[F, A, C]` constructor line.

rossabaker: I tried and couldn't make it compile.  Does it work with type parameters?

armanbilge: Hm, it should work 🤔 let me take a look.

armanbilge: Seems like SAMs don't work for `sealed`.

diesalbla: Perhaps we can drop `sealed` in favour of a `private[http4s]` constructor?

rossabaker: The encoding doesn't work yet either.  That PartialFunction trick may not be viable here.

armanbilge: I'm trying that right now, but it also doesn't seem to work 🤔 

Would it be so bad if it became a public trait?

rossabaker: If what did?  The Service?

I suspect we can optimize it with an algebra and an interpreter that "compiles" it to a function.  But whatever makes it actually work now.

rossabaker:
Hasty benchmark from b141aa1.  Those error bars are suspicious.

AltService is the run-based encoding from [dfc44db](https://github.com/http4s/http4s/pull/6769/commits/dfc44dbbca015b778e919e877b4fdbdc0d893e66).  It should beat routes, and should eventually lose to service.  If it doesn't lose soundly to service, altService is much easier to maintain.

```
[info] Benchmark                Mode  Cnt     Score      Error  Units
[info] ServiceBench.altService  avgt    3  9796.304 ± 2060.851  ns/op
[info] ServiceBench.routes      avgt    3  9521.738 ± 8432.776  ns/op
```

core/shared/src/main/scala/org/http4s/Service.scala
armanbilge: Right, let's not worry about SAMs until we decide the final encoding :)

rossabaker:
2afc896716 benchmarks:
* run as: `bench/jmh:run -i 3 -wi 3 -f 1 -t 8 org.http4s.bench.ServiceBench`.  (No, that's not many iterations or forks.)
* switched to Throughput mode
* run on multiple threads to be more realistic

```
[info] ServiceBench.altService  thrpt    3  430516.988 ± 208694.617  ops/s
[info] ServiceBench.routes      thrpt    3  326203.417 ± 363084.568  ops/s
[info] ServiceBench.service     thrpt    3  392654.888 ± 260234.873  ops/s
```

Initial thoughts:
* Strictly on performance, we're firmly in the "won't show up except for contrived benchmarks" zone.
* `AltService` does improve the error message experience and could have some extra combinators for manipulating the deeper effects.
* Variations on `routes` let you assemble your own stack and I still think that's beautiful even if it's secondary.
* The `Service` approach that elides the `Option` _should_ win.  By how much, and with how much effort, and how many people would understand its innards?
* If done right, we should be able to transparently switch from `AltService` to an improved `Service`.

rossabaker:
Some normalized memory pressure stats.  Here we see results like we'd expect.

```
[info] ServiceBench.altService:·gc.alloc.rate.norm               thrpt    3    1817.602 ±      0.168    B/op
[info] ServiceBench.altService:·gc.churn.PS_Eden_Space.norm      thrpt    3    1838.042 ±     72.360    B/op
[info] ServiceBench.altService:·gc.churn.PS_Survivor_Space.norm  thrpt    3       0.924 ±      0.492    B/op
[info] ServiceBench.routes:·gc.alloc.rate.norm                   thrpt    3    2233.586 ±      0.148    B/op
[info] ServiceBench.routes:·gc.churn.PS_Eden_Space.norm          thrpt    3    2257.726 ±    246.869    B/op
[info] ServiceBench.routes:·gc.churn.PS_Survivor_Space.norm      thrpt    3       1.037 ±      0.527    B/op
[info] ServiceBench.service:·gc.alloc.rate.norm                  thrpt    3    1377.629 ±      0.140    B/op
[info] ServiceBench.service:·gc.churn.PS_Eden_Space.norm         thrpt    3    1391.298 ±     75.091    B/op
[info] ServiceBench.service:·gc.churn.PS_Survivor_Space.norm     thrpt    3       0.928 ±      2.278    B/op
```

core/shared/src/main/scala/org/http4s/Service.scala
rossabaker: Fixed in [e8e380f](https://github.com/http4s/http4s/pull/6769/commits/e8e380fdc46fdd750db66a9b7b68a2f001854763)

core/shared/src/main/scala/org/http4s/Service.scala
rossabaker: I was curious whether a SAM would affect performance.  Doesn't seem to.

rossabaker:
I'll run the benchmarks overnight and report back in the morning.  `Service` is getting pretty easy to read, and competitive on throughput and allocations.

core/shared/src/main/scala/org/http4s/Service.scala
armanbilge: Right, SAMs should not affect performance. Lambdas are ultimately implemented as classes; for SAMs this class is synthesized at runtime and thus bring a small saving in bytecode.

rossabaker:
Apparently my Mac goes to sleep even when it's plugged in.  I didn't get the full numbers I wanted.  But:

The argument for `Service` over `AltService` seems to be allocations.  It might be possible to squeeze a few more cycles out with a custom algebra and tricks like map fusion, but it really doesn't matter.

It might also be possible to delay the creation of Resource, but at this point, I'm more concerned with the API than the runtime.

```console
sbt:http4s> bench/jmh:run -i 10 -wi 3 -f 1 -t 8 -prof gc org.http4s.bench.ServiceBench.*Fallback
[info] Benchmark                                                          Mode  Cnt       Score       Error   Units
[info] ServiceBench.altServiceFallback                                   thrpt   10  433212.686 ± 16443.973   ops/s
[info] ServiceBench.altServiceFallback:·gc.alloc.rate                    thrpt   10     777.925 ±    29.545  MB/sec
[info] ServiceBench.altServiceFallback:·gc.alloc.rate.norm               thrpt   10    1977.584 ±     0.017    B/op
[info] ServiceBench.altServiceFallback:·gc.churn.PS_Eden_Space           thrpt   10     785.614 ±    29.640  MB/sec
[info] ServiceBench.altServiceFallback:·gc.churn.PS_Eden_Space.norm      thrpt   10    1997.151 ±    12.259    B/op
[info] ServiceBench.altServiceFallback:·gc.churn.PS_Survivor_Space       thrpt   10       0.345 ±     0.047  MB/sec
[info] ServiceBench.altServiceFallback:·gc.churn.PS_Survivor_Space.norm  thrpt   10       0.878 ±     0.132    B/op
[info] ServiceBench.altServiceFallback:·gc.count                         thrpt   10    1197.000              counts
[info] ServiceBench.altServiceFallback:·gc.time                          thrpt   10     949.000                  ms
[info] ServiceBench.functionFallback                                     thrpt   10  440788.339 ±  4010.608   ops/s
[info] ServiceBench.functionFallback:·gc.alloc.rate                      thrpt   10     519.370 ±     4.756  MB/sec
[info] ServiceBench.functionFallback:·gc.alloc.rate.norm                 thrpt   10    1297.641 ±     0.007    B/op
[info] ServiceBench.functionFallback:·gc.churn.PS_Eden_Space             thrpt   10     525.027 ±     6.479  MB/sec
[info] ServiceBench.functionFallback:·gc.churn.PS_Eden_Space.norm        thrpt   10    1311.768 ±     8.392    B/op
[info] ServiceBench.functionFallback:·gc.churn.PS_Survivor_Space         thrpt   10       0.344 ±     0.032  MB/sec
[info] ServiceBench.functionFallback:·gc.churn.PS_Survivor_Space.norm    thrpt   10       0.859 ±     0.083    B/op
[info] ServiceBench.functionFallback:·gc.count                           thrpt   10    1236.000              counts
[info] ServiceBench.functionFallback:·gc.time                            thrpt   10     912.000                  ms
[info] ServiceBench.routesFallback                                       thrpt   10  416448.413 ± 16917.391   ops/s
[info] ServiceBench.routesFallback:·gc.alloc.rate                        thrpt   10     853.416 ±    34.553  MB/sec
[info] ServiceBench.routesFallback:·gc.alloc.rate.norm                   thrpt   10    2257.583 ±     0.037    B/op
[info] ServiceBench.routesFallback:·gc.churn.PS_Eden_Space               thrpt   10     861.984 ±    37.524  MB/sec
[info] ServiceBench.routesFallback:·gc.churn.PS_Eden_Space.norm          thrpt   10    2280.142 ±    11.081    B/op
[info] ServiceBench.routesFallback:·gc.churn.PS_Survivor_Space           thrpt   10       0.346 ±     0.076  MB/sec
[info] ServiceBench.routesFallback:·gc.churn.PS_Survivor_Space.norm      thrpt   10       0.913 ±     0.185    B/op
[info] ServiceBench.routesFallback:·gc.count                             thrpt   10    1144.000              counts
[info] ServiceBench.routesFallback:·gc.time                              thrpt   10     926.000                  ms
[info] ServiceBench.serviceFallback                                      thrpt   10  435881.448 ±  7898.478   ops/s
[info] ServiceBench.serviceFallback:·gc.alloc.rate                       thrpt   10     516.571 ±     9.444  MB/sec
[info] ServiceBench.serviceFallback:·gc.alloc.rate.norm                  thrpt   10    1305.635 ±     0.021    B/op
[info] ServiceBench.serviceFallback:·gc.churn.PS_Eden_Space              thrpt   10     521.956 ±    10.255  MB/sec
[info] ServiceBench.serviceFallback:·gc.churn.PS_Eden_Space.norm         thrpt   10    1319.237 ±     6.274    B/op
[info] ServiceBench.serviceFallback:·gc.churn.PS_Survivor_Space          thrpt   10       0.340 ±     0.028  MB/sec
[info] ServiceBench.serviceFallback:·gc.churn.PS_Survivor_Space.norm     thrpt   10       0.859 ±     0.067    B/op
[info] ServiceBench.serviceFallback:·gc.count                            thrpt   10    1247.000              counts
[info] ServiceBench.serviceFallback:·gc.time                             thrpt   10     927.000                  ms
[success] Total time: 549 s (09:09), completed Oct 24, 2022 9:34:16 AM
```

PR #6772, opened 2022-10-24
Ember-Core: simplify logic for push promises.

diesalbla:
Compact the calls to `traverse` and `parTraverse` on a single branch, the one with push enabled and any push promises in the response.


danicheg (approved):

danicheg:
I think this simplification is absolutely legit 👍🏻

PR #6774, opened 2022-10-26
Use `map` instead of `mapValues` on `Map`s

danicheg:
```
[info] Benchmark           (size)  Mode  Cnt      Score      Error  Units
[info] MapBench.map            10  avgt    5    238.153 ±   40.219  ns/op
[info] MapBench.map           100  avgt    5   6024.884 ±  409.198  ns/op
[info] MapBench.map          1000  avgt    5  59488.263 ±  243.845  ns/op
[info] MapBench.mapValues      10  avgt    5    318.889 ±    2.540  ns/op
[info] MapBench.mapValues     100  avgt    5   6222.427 ±  104.340  ns/op
[info] MapBench.mapValues    1000  avgt    5  64835.303 ± 2168.535  ns/op
```
Test results are bouncing a little but nonetheless show that we should get a 5-10% performance boost at least.

### DIFF
--- a/core/shared/src/main/scala/org/http4s/Query.scala
+++ b/core/shared/src/main/scala/org/http4s/Query.scala
@@ -143,7 +143,9 @@ final class Query private (value: Either[Vector[KeyValue], String])
     * none exist, the empty `String` "" is returned.
     */
   lazy val params: Map[String, String] =
-    multiParams.view.mapValues(_.headOption.getOrElse("")).toMap
+    multiParams.map { case (k, v) =>
+      k -> v.headOption.getOrElse("")
+    }
 
   /** `Map[String, List[String]]` representation of the [[Query]]
     *
--- a/core/shared/src/main/scala/org/http4s/UrlForm.scala
+++ b/core/shared/src/main/scala/org/http4s/UrlForm.scala
@@ -134,7 +134,7 @@ object UrlForm {
   )(urlForm: String): Either[MalformedMessageBodyFailure, UrlForm] =
     QueryParser
       .parseQueryString(urlForm.replace("+", "%20"), new Codec(charset.nioCharset))
-      .map(q => UrlForm(q.multiParams.view.mapValues(Chain.fromSeq).toMap))
+      .map(q => UrlForm(q.multiParams.map(x => x._1 -> Chain.fromSeq(x._2))))
       .leftMap { parseFailure =>
         MalformedMessageBodyFailure(parseFailure.message, None)
       }


armanbilge (approved):

rossabaker:
Does this not apply in 0.23?

danicheg:
Basically, we can use `map` for `Map` in 2.12, but its implementation differs from 2.13. I bet that we will get a worse performance for 2.12. But let me check.

danicheg:
These results make me whaaaaat
```
[info] Benchmark           (size)  Mode  Cnt      Score      Error  Units
[info] MapBench.map            10  avgt    5    631.974 ±   18.464  ns/op
[info] MapBench.map           100  avgt    5   5596.358 ±  529.657  ns/op
[info] MapBench.map          1000  avgt    5  52479.851 ± 2228.271  ns/op
[info] MapBench.mapValues      10  avgt    5      3.380 ±    0.074  ns/op
[info] MapBench.mapValues     100  avgt    5      3.340 ±    0.064  ns/op
[info] MapBench.mapValues    1000  avgt    5      3.379 ±    0.232  ns/op
```
I don't know what occurred with `mapValues` in 2.13, but it blasts the `Map#map` entirely.

danicheg:
Yeah, I mean that `map.view.mapValues(f).toMap` is hugely worse than `map.mapValues(f)`.

satorg:
Just for a record: it is not enough simply to call `mapValues` in Scala 2.12. Because it produces a view instead of real `Map`. I guess this is why you got such weird benchmarks for 2.12. To make a fair comparison, you have to call `.mapValues(...).force` which will re-create a new concrete `Map` out of the view.

satorg:
In other words, in Scala 2.12 `.mapValues` is similar to just `.view.mapValues` (without `.toMap` following) in 2.13+ although is less explicit.

danicheg:
I have tested on `map.view.mapValues(f).toMap` doesn't that enough for fairness?

danicheg:
By the way, here is the code for the benchmark I used. Feel free to run it and share the results.

```scala
import org.openjdk.jmh.annotations._
import java.util.concurrent.TimeUnit
import scala.util.Random

@BenchmarkMode(Array(Mode.AverageTime))
@OutputTimeUnit(TimeUnit.NANOSECONDS)
@State(Scope.Benchmark)
class MapBench {
  @Param(Array("10", "100", "1000"))
  var size: Int = _

  var queryParams: Map[String, List[String]] = _

  private def genString(): String =
    Random.nextString(Random.nextInt(50))

  @Setup(Level.Trial)
  def setup(): Unit =
    queryParams = (1 to size).map(_ => genString() -> List.fill(10)(genString())).toMap

  @Benchmark
  def mapValues: Map[String, String] =
    queryParams.mapValues(_.headOption.getOrElse("")).toMap

  @Benchmark
  def map: Map[String, String] =
    queryParams.map(x => x._1 -> x._2.headOption.getOrElse(""))
}
```

satorg:
> I have tested on `map.mapValues(f).toMap` (for 2.12) doesn't that enough for fairness?

I'm afraid, nope, because it simply return `this` there. Check this out:
```
Welcome to Scala 2.12.17 (OpenJDK 64-Bit Server VM, Java 11.0.16.1).

scala> val map1 = (1 to 10).map(n => n -> n.toString).toMap
map1: scala.collection.immutable.Map[Int,String] = Map(5 -> 5, 10 -> 10, 1 -> 1, 6 -> 6, 9 -> 9, 2 -> 2, 7 -> 7, 3 -> 3, 8 -> 8, 4 -> 4)

scala> val map2 = map1.mapValues("[" + _ + "]")
map2: scala.collection.immutable.Map[Int,String] = Map(5 -> [5], 10 -> [10], 1 -> [1], 6 -> [6], 9 -> [9], 2 -> [2], 7 -> [7], 3 -> [3], 8 -> [8], 4 -> [4])

scala> val map3 = map2.toMap
map3: scala.collection.immutable.Map[Int,String] = Map(5 -> [5], 10 -> [10], 1 -> [1], 6 -> [6], 9 -> [9], 2 -> [2], 7 -> [7], 3 -> [3], 8 -> [8], 4 -> [4])

scala> map2 eq map3
res0: Boolean = true

scala> val map4 = map2.iterator.toMap
map4: scala.collection.immutable.Map[Int,String] = Map(5 -> [5], 10 -> [10], 1 -> [1], 6 -> [6], 9 -> [9], 2 -> [2], 7 -> [7], 3 -> [3], 8 -> [8], 4 -> [4])

scala> map2 eq map4
res1: Boolean = false
```
Sorry, I forgot that there's no `Map#force` in Scala 2.12. The enforcement has to be conducted with something like `.iterator.toMap` instead.

satorg:
Oh, I remembered why there could be `.force` in Scala 2.12:
```
scala> val map5 = map2.view.force
map5: scala.collection.immutable.Map[Int,String] = Map(5 -> [5], 10 -> [10], 1 -> [1], 6 -> [6], 9 -> [9], 2 -> [2], 7 -> [7], 3 -> [3], 8 -> [8], 4 -> [4])

scala> map2 eq map5
res2: Boolean = false
```
Which is pretty similar to `.iterator.toMap`, I guess.

danicheg:
But calling `.iterator.toMap` is too extra in the real world, I believe. I have tested the real [code usage](https://github.com/http4s/http4s/blob/4b928e0dc0ba6edbdbe7461204663e13a7013f8c/core/shared/src/main/scala-2.12/org/http4s/internal/CollectionCompat.scala#L23-L24) from http4s. At least that was aimed.

satorg:
> But calling .iterator.toMap is too extra in the real world, I believe.

It was pretty common in Scala prior to v2.13 to see either `.mapValues.view.force` or `.mapValues.iterator.toMap` if one wanted to get a concrete collection, not just view. 🤷  And there was no other way to achieve that.

And it is pretty much corresponds to `.view.mapValues.force` in Scala v2.13+ (but the latter is more explicit and sounding).

So if we're comparing `.mapValues` in Scala 2.12 to `.view.mapValues.force` in Scala 2.13 then we are sort of comparing apples and oranges, I think.

danicheg:
I mean there is no reason to do `map.mapValues(f).iterator.toMap`. Also using `map.mapValues(f).toMap` is pretty common in 2.12, doubtless. Apologize for my dumb question, but what are you arguing about? :)

satorg:
Oh, no, I'm not arguing actually. Sorry if it looks like I do. I just made a note, that results retrieved on Scala 2.12 for `.mapValues` are not that surprising. But I have no objections regarding the PR itself.

bwiercinski:
how would perform fmap from cats? see cats.instances.MapInstances#catsStdInstancesForMap

PR #6776, opened 2022-10-28
More efficient string building from `Headers`

danicheg:
```
[info] Benchmark                            (size)  Mode    Cnt      Score      Error  Units
[info] currentMkString                        10     avgt    5      653.514 ±   37.092  ns/op
[info] currentMkString                        100    avgt    5     6576.612 ±  907.011  ns/op
[info] currentMkString                        1000   avgt    5    58623.540 ± 1841.507  ns/op
[info] newMkString                            10     avgt    5      422.176 ±   12.110  ns/op
[info] newMkString                            100    avgt    5     4197.781 ±   15.698  ns/op
[info] newMkString                            1000   avgt    5    34725.574 ±  234.841  ns/op
```
This adds new methods for `Headers` for string building. Also, this brings 55-70% performance improvement and definitely should reduce memory consumption quite much.

```scala
@Benchmark
def currentMkString: String =
  headers.redactSensitive().headers.mkString("Headers(", ", ", ")")

@Benchmark
def newMkString: String =
  headers.mkString("Headers(", ", ", ")")
```

### DIFF
--- a/core/shared/src/main/scala/org/http4s/Headers.scala
+++ b/core/shared/src/main/scala/org/http4s/Headers.scala
@@ -180,6 +180,27 @@ final class Headers(val headers: List[Header.Raw]) extends AnyVal {
       }
       .mkString(start, separator, end)
 
+  /** Creates a string representation for a list of headers
+    * and redacts sensitive headers' values.
+    *
+    *  @param separator   the separator string
+    *  @param redactWhen  the function for filtering out header values of sensitive headers
+    *  @return            a string representation of the list of headers.
+    *                      The resulting string is constructed from the string representations
+    *                      of all headers separated by the string `separator`. Sensitive headers'
+    *                      values are redacted with the `redactWhen` function.
+    */
+  def mkStringSeparatedBy(
+      separator: String,
+      redactWhen: CIString => Boolean = Headers.SensitiveHeaders.contains,
+  ): String =
+    headers.iterator
+      .map {
+        case h if redactWhen(h.name) => Header.Raw.toString(h.name, "<REDACTED>")
+        case h => Header.Raw.toString(h.name, h.value)
+      }
+      .mkString(separator)
+
   override def toString: String =
     this.show
 }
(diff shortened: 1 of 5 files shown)


core/shared/src/main/scala/org/http4s/Headers.scala
```diff
@@ -153,6 +153,54 @@ final class Headers(val headers: List[Header.Raw]) extends AnyVal {
...
+    * and redacts sensitive headers' values.
+    *
+    *  @param separator   the separator string
+    *  @param redactWhen  the function for filtering out header values of sensitive headers
+    *  @return            a string representation of the list of headers.
+    *                      The resulting string is constructed from the string representations
+    *                      of all headers separated by the string `separator`. Sensitive headers'
+    *                      values are redacted with the `redactWhen` function.
+    */
+  def mkStringSeparatedBy(
```
danicheg, line 193: Unfortunately, scalac 2 has [limitations](https://stackoverflow.com/questions/4652095/why-does-the-scala-compiler-disallow-overloaded-methods-with-default-arguments) for having overloaded methods with default values. And I didn't manage to come up with some better name.

armanbilge: Instead of default values we could just add more overloads without those arguments 😅

danicheg:
Hey @http4s/core, your review of this would be so much appreciated. It seems to me that this brings quite a significant improvement to one hot path (logging).

hamnis:
this looks very reasonable to do. I agree with @armanbilge that overloading without default parameters is better. default parameters may have binary compatibility problems as well

danicheg:
Truth be told, I'm not a fan of default values (especially for constructors) they hurt me many times. But almost everywhere in http4s API, that function (for detecting headers for redacting) is passed with a default value. So I preserved the same behavior here.

danicheg:
But if the majority of maintainers agree with functions without default parameter values, then that's fine with me too :+1:

danicheg:
Folks, let's finish this together. Should I fix something else?

armanbilge (approved):
Nope, LGTM 😃 thanks for iterating!

PR #6777, opened 2022-10-29
Don't instantiate redundant `NonEmptyList` in `Writer#addList`

danicheg:
The performance of `Writer#addList` shouldn't change, but fewer allocations is also fine.

### DIFF
--- a/core/shared/src/main/scala/org/http4s/util/Renderable.scala
+++ b/core/shared/src/main/scala/org/http4s/util/Renderable.scala
@@ -206,10 +206,14 @@ trait Writer { self =>
       start: String = "",
       end: String = "",
   ): this.type =
-    NonEmptyList.fromList(s) match {
-      case Some(s) => addNel(s, sep, start, end)
-      case None =>
+    s match {
+      case Nil =>
         append(start)
+        append(end)
+      case x :: list =>
+        append(start)
+        append(x)
+        list.foreach(s => append(sep).append(s))
         append(end)
     }
 


core/shared/src/main/scala/org/http4s/util/Renderable.scala
```diff
@@ -206,11 +206,15 @@ trait Writer { self =>
...
-      case None =>
+    s match {
+      case Nil =>
         append(start)
         append(end)
+      case x :: list =>
+        append(start)
+        append(x)
+        list.foreach(s => append(sep).append(s))
+        append(end)
```
satorg, line 217: So this basically repeats `addNel` implementation but with `NonEmptyList` decomposed into head and tail parts. Would it make sense to extract that implementation into a private method and then re-use it from both places, wdyt?
```suggestion
      case x :: list =>
        addListHT(x, list)
```
then below
```scala
        // just a random name
        // we can consider using `@inline` here
        private def addListHT[T: Render](head:  T, tail: List[T], sep, start, end) = {
          // the implementation itself
        }

        def addNel[T: Render](...) = addListHT(...)
```

danicheg: DRY is okay but for a few liners like this, I don't that much like it. Because why don't we unify implementation also for `Seq` then?

satorg: Why not? If we made the input collection just `Iterable` we could unify it even for `Set`:
```
@inline private def addEmpty(...) = {
  append(start)
  append(end)
}
@inline private def addNonEmpty[T](head: T, tail: Iterable[T], ...) = {
  // all that stuff
}
```
then we could use the combination of those above in a bunch of methods like `addList`, `addNel`, `addSet` and even `addStrings`, e.g.:
```scala
def addSet[T](s: Set[T], ...) = {
  if (s.isEmpty) addEmpty(...)
  else addNonEmpty(s.head, s.tail, ...)
}
```

satorg: On the second thought... Looks like there's no need for so much hustle even like that.
It seems that all the cases can be covered with just a single (and higher-performance) method like this:
```scala
private def addIterator[T: Renderer](it: Iterator[T], sep: String, start: String, end: String) = {
    append(start)

    var hasNext = it.hasNext
    while (hasNext) {
      append(it.next())

      hasNext = it.hasNext
      if (hasNext) append(sep)
    }

    append(end)
  }

```
Besides it can cover `Seq`, `List`, `NonEmptyList` and everything that can provide `Iterator`, it also gets rid of `.foreach` and thus a closure passed to it.

Therefore all other methods can simply call this one under the hood.

diesalbla: Just a warning, but if we want to reduce allocations, take into account that an iterator is a separate object from the collection.

diesalbla (approved):
@danicheg As a further suggestion for reducing allocations, on another PR: a `StringBuilder` is implemented using a resizing array, which means there is a duplication (build a new array and move all items there) every time the array gets full. If we can first add up the total size of the final string (which for headers is feasible), or even a generous overestimate, and then initialise the builder to that size, that would save those duplications.

core/shared/src/main/scala/org/http4s/util/Renderable.scala
satorg: That's fair, agree. On the other hand, the iterator would allow to avoid another indirection – a call to a lamda method in `foreach` which occurs for _each_ element rather than just once. But of course, it is hard to say, what is better without benchmarking of both. And perhaps, it is not that important here.

satorg (approved):

danicheg:
I like the suggestions of @diesalbla  and @satorg for further refactoring of `Writer` and other things. The only thing I wanted to achieve there – is only don't instantiate redundant collection. If we can go with this as it is – I'm happy. But don't hesitate and create PRs with your ideas.

Issue #6780, opened 2022-10-31
Ember server: Empty shutdown timeout causes "connection reset by peer" on all requests


kubukoz:
Expectation: `withShutdownTimeout(0)` is only used when the server is shutting down

Reality: the value of 0 seems to have a special effect in which all client requests get rejected.

To reproduce, start the Server app here (scala-cli):

```scala
//> using scala "2.13.10"
//> using lib "org.typelevel::cats-effect:3.3.4"
//> using lib "org.http4s::http4s-ember-server:0.23.16"
//> using lib "org.http4s::http4s-ember-client:0.23.16"
import cats.effect.IOApp
import cats.effect.IO
import cats.effect.std.Console
import com.comcast.ip4s._
import org.http4s.ember.server.EmberServerBuilder
import org.http4s.HttpApp
import org.http4s.ember.client.EmberClientBuilder
import org.http4s.client.dsl.io._
import org.http4s.Method
import org.http4s.implicits._
import scala.concurrent.duration.Duration

object Server extends IOApp.Simple {

  def run: IO[Unit] =
    EmberServerBuilder
      .default[IO]
      .withHttpApp(HttpApp.notFound)
      .withPort(port"4000")
      .withHost(host"0.0.0.0")
      .withShutdownTimeout(Duration.Zero)
      .build
      .useForever

}

object Client extends IOApp.Simple {

  def run: IO[Unit] = EmberClientBuilder
    .default[IO]
    .build
    .use { client =>
      client.run(Method.GET.apply(uri"http://localhost:4000")).use(_.bodyText.compile.string)
    }
    .flatMap(IO.println(_))

}
```

For clients, you can use anything (curl, httpie, the `Client` app above all surface the issue).

kubukoz:
I got as far as `Shutdown.immediateShutdown` which seems to very much be a special case, but I didn't dig deeper.

PR #6781, opened 2022-10-31
`Shutdown#trackConnection` should never be empty

armanbilge:
Fixes https://github.com/http4s/http4s/issues/6780.


ember-server/shared/src/main/scala/org/http4s/ember/server/internal/Shutdown.scala
```diff
@@ -100,7 +100,7 @@ private[server] object Shutdown {
         override val signal: F[Unit] = unblock.get
         override val newConnection: F[Unit] = F.unit
         override val removeConnection: F[Unit] = F.unit
-        override val trackConnection: Stream[F, Unit] = Stream.empty
+        override val trackConnection: Stream[F, Unit] = Stream.emit(())
```
armanbilge, line 103: In the next FS2 release this can be `Stream.unit`.

kubukoz (approved):

ChristopherDavenport (approved):

PR #6787, opened 2022-11-03

danicheg:


ember-core/shared/src/main/scala/org/http4s/ember/core/Encoder.scala
```diff
@@ -140,7 +140,9 @@ private[ember] object Encoder {
         }
 
         if (
-          !appliedContentLength && req.body == EmptyBody && !NoPayloadMethods.contains(req.method)
+          !appliedContentLength && req.entity == Entity.Empty && !NoPayloadMethods.contains(
+            req.method
+          )
```
armanbilge, line 145: Nice :)

danicheg: Not really ._.

armanbilge: Yeah the formatting got a bit ugly. But I think if we can land https://github.com/http4s/http4s/pull/6762 then we can make this a lot nicer and also more efficient.

armanbilge (approved):

PR #6788, opened 2022-11-05
Fix the wrong `Http4sClientDsl` location

danicheg:
`Http4sClientDsl` belongs to the `org.http4s.client.dsl` package, but is located in the `./org/http4s/client` folder.


satorg:
I agree with the motivation: since `import org.http4s.client.dsl._` does not bring any DSL into a scope (we have to use `import org.http4s.client.dsl.io._` for the IO-specific DSL), then to me `org.http4s.client.dsl` also looks like a better place for all client DSL related stuff.

But seems that this PR simply moves a file around folders, but does not change its package directives.
```
package org.http4s
package client
package dsl
```
Therefore the resulting location of the trait has not beet changed – is that correct?

armanbilge (approved):

armanbilge:
Yup, this has no impact on compiled artifacts. It's just a code style thing.

danicheg:
That's right, thanks @armanbilge for explaining things.

satorg:
Ah, yes, I see now: the trait was in the correct package already but not in a correct folder. So this PR has fixed the discrepancy. Sorry, I just didn't connect two points in my brain initially.

PR #6789, opened 2022-11-05
Ember-Server - ServerHelpers.scala - Refactors

diesalbla:
1. Make "upgradeSocket" into a nested method, fewer parameters.
2. Extract serverFromSocket shared local method
3. Extract a runConnectionAux method.
4. replace match with an if-then-else
5. Extract body of "map" as an "onConnection" function.

### DIFF
--- a/ember-server/shared/src/main/scala/org/http4s/ember/server/internal/ServerHelpers.scala
+++ b/ember-server/shared/src/main/scala/org/http4s/ember/server/internal/ServerHelpers.scala
@@ -214,41 +214,43 @@ private[server] object ServerHelpers extends ServerHelpersPlatform {
         enableHttp2,
       )
 
+    def onConnection(connect: Socket[F]): Stream[F, Nothing] = {
+      val handler: Stream[F, Nothing] = shutdown.trackConnection >>
+        Stream
+          .resource(upgradeSocket(connect))
+          .flatMap {
+            case (socket, Some("h2")) =>
+              // ALPN H2 Strategy
+              Stream.exec(H2Server.requireConnectionPreface(socket)) ++
+                Stream.resource(serverFromSocket(connect)).drain
+            case (socket, Some(_)) =>
+              // SSL Connection, not h2, will be http/1.1 but thats not how types align
+              // Prior Knowledge is only allowed over clear where application
+              // protocol has not been agreed via handshake
+              runConnectionAux(socket, ByteVector.empty)
+            case (socket, None) => // Cleartext Protocol
+              if (enableHttp2) {
+                // Http2 Prior Knowledge Check, if prelude is first bytes received tread as http2
+                // Otherwise this is now http1
+                Stream.eval(H2Server.checkConnectionPreface(socket)).flatMap {
+                  // Pass read bytes we thought might be the prelude
+                  case Left(bv) => runConnectionAux(socket, bv)
+                  case Right(_) => Stream.resource(serverFromSocket(connect)).drain
+                }
+              } else {
+                // Since its not enabled, run connection normally.
+                runConnectionAux(socket, ByteVector.empty)
+              }
+          }
+
+      handler.handleErrorWith { t =>
+        Stream.eval(logger.error(t)("Request handler failed with exception")).drain
+      }
+    }
+
     val streams: Stream[F, Stream[F, Nothing]] = server
       .interruptWhen(shutdown.signal.attempt)
-      .map { connect =>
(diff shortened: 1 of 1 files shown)


ember-server/shared/src/main/scala/org/http4s/ember/server/internal/ServerHelpers.scala
```diff
@@ -173,109 +173,84 @@ private[server] object ServerHelpers extends ServerHelpersPlatform {
...
+                // Since its not enabled, run connection normally.
+                runConnectionAux(socket, ByteVector.empty)
+              }
+          }
 
-        handler.handleErrorWith { t =>
-          Stream.eval(logger.error(t)("Request handler failed with exception")).drain
-        }
+      handler.handleErrorWith { t =>
+        Stream.eval(logger.error(t)("Request handler failed with exception")).drain
```
valencik, line 247:
```suggestion
        Stream.exec(logger.error(t)("Request handler failed with exception"))
```

ember-server/shared/src/main/scala/org/http4s/ember/server/internal/ServerHelpers.scala
```diff
@@ -173,109 +173,84 @@ private[server] object ServerHelpers extends ServerHelpersPlatform {
...
+              Stream.exec(H2Server.requireConnectionPreface(socket)) ++
+                Stream.resource(serverFromSocket(connect)).drain
+            case (socket, Some(_)) =>
+              // SSL Connection, not h2, will be http/1.1 but thats not how types align
+              // Prior Knowledge is only allowed over clear where application
+              // protocol has not been agreed via handshake
+              runConnectionAux(socket, ByteVector.empty)
+            case (socket, None) => // Cleartext Protocol
+              if (enableHttp2) {
+                // Http2 Prior Knowledge Check, if prelude is first bytes received tread as http2
```
valencik, line 233:
```suggestion
                // Http2 Prior Knowledge Check, if prelude is first bytes received treat as http2
```

ember-server/shared/src/main/scala/org/http4s/ember/server/internal/ServerHelpers.scala
```diff
@@ -173,109 +173,84 @@ private[server] object ServerHelpers extends ServerHelpersPlatform {
...
-                      onWriteFailure,
-                      createRequestVault,
-                      webSocketKey,
-                      ByteVector.empty,
-                      enableHttp2,
-                    ).drain
+
+    def upgradeSocket(socketInit: Socket[F]): Resource[F, (Socket[F], Option[String])] =
+      tlsInfoOpt match {
+        case None => (socketInit, Option.empty[String]).pure[Resource[F, *]]
```
valencik, line 179:
```suggestion
        case None => Resource.pure(socketInit, None)
```

Maybe the `.pure` style is preferred? I don't know.
But `Resource.pure(socketInit, None)` feels more straight forward to me.

valencik (approved):
_Unofficial review_

I like this. I think breaking things up is nice and there's some good de-duplication / simplification here. 👍

ember-server/shared/src/main/scala/org/http4s/ember/server/internal/ServerHelpers.scala
diesalbla: Me too. Do not much like dealing with asterisks.

ember-server/shared/src/main/scala/org/http4s/ember/server/internal/ServerHelpers.scala
```diff
@@ -173,109 +173,84 @@ private[server] object ServerHelpers extends ServerHelpersPlatform {
...
-                      onWriteFailure,
-                      createRequestVault,
-                      webSocketKey,
-                      ByteVector.empty,
-                      enableHttp2,
-                    ).drain
+
+    def upgradeSocket(socketInit: Socket[F]): Resource[F, (Socket[F], Option[String])] =
+      tlsInfoOpt match {
+        case None => Resource.pure(socketInit, None)
```
hamnis, line 179: fails on 2.12: https://github.com/http4s/http4s/actions/runs/3826329960/jobs/6510029546#step:10:105

valencik: Woops, sorry, should have had more parens:
```suggestion
        case None => Resource.pure((socketInit, None))
```
tested locally, worked with 2.12.17

PR #6790, opened 2022-11-06
Use modeled headers in CORS middleware

danicheg:
This resolves some `// TODO`s.


core/shared/src/main/scala/org/http4s/headers/`Access-Control-Max-Age`.scala
```diff
@@ -18,6 +18,7 @@ package org.http4s
 package headers
 
 import org.http4s.parser.AdditionalRules
+import org.typelevel.ci.CIString
 import org.typelevel.ci.CIStringSyntax
```
satorg, line 22: Just a minor nit: although there's nothing besides CIString and its syntax in the package, then these two imports could be replaced with just a single `import org.typelevel.ci._`

I think it is a suggested way in any case...

satorg (approved):
Thank you!

satorg:
One more thing to note (not directly related to this PR though, but still).

I noticed that both `Access-Control-Allow-Methods` and `Access-Control-Max-Age` header models are located in files with quite unusual back-ticked names: `` `Access-Control-Allow-Methods`.scala `` and `` `Access-Control-Max-Age`.scala `` correspondingly. There are only three those files with such names (along with another one `` `Idempotency-Key`.scala ``).

I think it would be better to get rid of backticks in filenames – I personally had been struggling to find those files until I realized that there are backticks in their names. Not necessary to be addressed in this PR, of course...

core/shared/src/main/scala/org/http4s/headers/`Access-Control-Max-Age`.scala
danicheg: There is a crowd that believes using wildcard imports makes compilation slower. But I don't mind this change at all. Let's do it.

danicheg:
> There are only three those files with such names (along with another one `Idempotency-Key`.scala).

I'm not sure about that 🤔 
Screenshot 2022-11-07 at 21 12 08

satorg:
Oh, I'm talking about file names, not class names. Backticked class names look fine to me. But their corresponding file names look this way currently:
```
.../http4s/core/shared/src/main/scala/org/http4s/headers (series/0.23)$ ls -1
Accept-Charset.scala
Accept-Encoding.scala
Accept-Language.scala
Accept-Patch.scala
Accept-Post.scala
Accept-Ranges.scala
Accept.scala
Access-Control-Allow-Credentials.scala
Access-Control-Allow-Headers.scala
Access-Control-Expose-Headers.scala
Access-Control-Request-Method.scala
Age.scala
Allow.scala
Authorization.scala
Cache-Control.scala
Connection.scala
Content-Disposition.scala
Content-Encoding.scala
Content-Language.scala
Content-Length.scala
Content-Location.scala
Content-Range.scala
Content-Type.scala
Cookie.scala
Cross-Origin-Resource-Policy.scala
DNT.scala
Date.scala
ETag.scala
Expires.scala
Forwarded.scala
ForwardedRenderers.scala
HeaderCompanion.scala
Host.scala
If-Match.scala
If-Modified-Since.scala
If-None-Match.scala
If-Range.scala
If-Unmodified-Since.scala
Keep-Alive.scala
Last-Event-Id.scala
Last-Modified.scala
Link.scala
LinkValue.scala
Location.scala
Max-Forwards.scala
Origin.scala
Proxy-Authenticate.scala
Proxy-Authorization.scala
Range.scala
Referer.scala
Retry-After.scala
Sec-WebSocket-Accept.scala
Sec-WebSocket-Key.scala
Sec-WebSocket-Version.scala
Server.scala
Set-Cookie.scala
SourceMap.scala
Strict-Transport-Security.scala
Trailer.scala
Transfer-Encoding.scala
Upgrade.scala
User-Agent.scala
WWW-Authenticate.scala
X-B3-Flags.scala
X-B3-ParentSpanId.scala
X-B3-Sampled.scala
X-B3-SpanId.scala
X-B3-TraceId.scala
X-Forwarded-For.scala
X-Forwarded-Proto.scala
`Access-Control-Allow-Methods`.scala
`Access-Control-Max-Age`.scala
`Idempotency-Key`.scala
package.scala
```
So there are only three files in the whole directory that have backticked file names. And such namings break their sorting order in particular. And I think that backticked file names look weird no matter what.

danicheg:
Ah, now I see. Ok, let's fix that too.

satorg (approved):
lgtm, thank you!

